PR #3875, opened 2026-04-14
feat: add --allowed-source-cidrs opt-in for local API access

sanity:
## Problem

The local HTTP/WebSocket API (port 7509) is hard-restricted to loopback + RFC1918 + IPv6 link-local / ULA via the `private_network_filter` middleware. Any other source IP is rejected with *"Only local network connections are allowed"*. There is no way to grant access from a private overlay network (Tailscale, WireGuard, Nebula) without either forking core or binding the API to an unrelated private address and hoping routing works out. For users who want to reach their own node from a phone while away from their LAN, the only supported path is an SSH tunnel.

Matrix user question that prompted this: *"Any chance of adding support for accessing a Freenet node running over tailscale?"* Right now `http://freenetnode:7509/` on a tailnet fails because the source IP (CGNAT 100.64.0.0/10) is neither RFC1918 nor loopback.

## Solution

Add an **opt-in** CIDR allowlist that extends the built-in private-IP check. Default behavior is unchanged (the field is empty), so users who don't configure anything see zero posture change.

- New CLI flag `--allowed-source-cidrs` (comma/repeat), env var `ALLOWED_SOURCE_CIDRS`, and `allowed-source-cidrs` field under `[ws-api]` in `config.toml`.
- Parsed at config-build time via `ipnet::IpNet` (already a transitive workspace dep), so bad CIDRs fail loudly at startup instead of silently at runtime.
- `private_network_filter` accepts a request if the source IP is private **or** matches any configured CIDR. IPv4-mapped IPv6 sources from the dual-stack socket are normalized to IPv4 before matching, so users can express `100.64.0.0/10` in natural v4 notation.
- Parsed CIDRs are logged at `warn!` on startup (with the list) so operators see their allowlist on every boot. The log message spells out the risk.

### Why no auto-trust of 100.64.0.0/10

It was tempting to auto-trust CGNAT space since Tailscale uses it, but CGNAT is **shared** between subscribers of some ISPs (Starlink, T-Mobile, many cable carriers). Auto-trusting would silently widen the trust boundary for those users from *their LAN only* to *every other customer on their ISP's CGNAT segment*. Keeping the opt-in explicit preserves the guarantee that the default posture is the one we audited: private RFC1918 space you control, nothing else.

### Alternatives considered

- **Interface-scoped binding alone.** Even if the socket is bound to `tailscale0`, the filter still rejects the 100.x source IP. The allowlist is still required. Docs recommend doing both for defense in depth.
- **Auth token for remote access.** Legitimate direction long-term, but orthogonal: the API has no authn today, so any path that reaches the port is already privileged. The opt-in CIDR list doesn't make that worse.

## Testing

New unit tests in `crates/core/src/server.rs`:

- `allowed_source_cidrs_empty_rejects_public`  (default-deny regression)
- `allowed_source_cidrs_does_not_accept_public_by_default` (explicit guard that CGNAT / public / TEST-NET-3 are NOT trusted with default config)
- `allowed_source_cidrs_tailscale_cgnat` (100.64.0.0/10 boundary: accepts 100.64/100.100/100.127.x, rejects 100.63 / 100.128 / public)
- `allowed_source_cidrs_narrow_tailnet` (narrower /24 works and excludes neighbors)
- `allowed_source_cidrs_ipv6` (fd7a:115c:a1e0::/48 Tailscale v6 range)
- `allowed_source_cidrs_multiple_ranges` (multi-range allowlist)

All existing `test_is_private_ip_*` tests unchanged and still green.

`cargo check -p freenet --tests`, `cargo clippy -p freenet --lib -- -D warnings`, and `cargo test -p freenet --lib server::` all pass. The new macro/test fixtures for `WebsocketApiArgs { allowed_source_cidrs: None }` and `WebsocketApiConfig { allowed_source_cidrs: Vec::new() }` are mechanical updates in `freenet-macros/src/codegen.rs` and `crates/core/tests/*.rs`.

### Why unit tests instead of simulation

The filter lives outside the DHT / routing / transport surface, so simulation tests give no extra signal. The logic under test is literally *"does this IP match one of these CIDRs"*, which is best expressed as boundary-focused table tests. Happy to add an integration test that spins up axum and makes a real request from an allowed / denied source if a reviewer wants belt-and-braces coverage.

## Docs

Separate PR in `freenet/web` adds `hugo-site/content/build/manual/remote-access.md` explaining the two supported ways to reach a node from off-LAN (SSH tunnel as the safest default; Tailscale with this new CIDR allowlist as the alternative). That PR links back here.

## Fixes

Addresses a Matrix question about Tailscale access; no existing issue.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2083,6 +2083,7 @@ dependencies = [
  "hickory-resolver",
  "hostname",
  "httptest",
+ "ipnet",
  "libc",
  "lru",
  "moka",
@@ -3470,6 +3471,9 @@ name = "ipnet"
 version = "2.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d98f6fed1fde3f8c21bc40a1abb88dd75e67924f9cffc3ef95607bad8017f8e2"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "ipnetwork"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2124,6 +2124,7 @@ dependencies = [
  "tokio-stream",
  "tokio-tungstenite 0.27.0",
  "toml 1.1.2+spec-1.1.0",
+ "tower",
  "tower-http",
  "tracing",
  "tracing-appender",
--- a/apps/freenet-ping/app/tests/common/mod.rs
+++ b/apps/freenet-ping/app/tests/common/mod.rs
@@ -154,6 +154,7 @@ pub async fn base_node_test_config_with_rng<R: Rng>(
             token_ttl_seconds: None,
             token_cleanup_interval_seconds: None,
             allowed_host: None,
+            allowed_source_cidrs: None,
         },
         network_api: NetworkArgs {
             // Use varied IP for network socket to get unique ring locations
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -151,6 +151,7 @@ which = { workspace = true }
 regex = { workspace = true }
 serial_test = { workspace = true }
 url = "2"
+tower = { version = "0.5", features = ["util"] }
 
 # CI-optimized benchmark suite (~5 min, deterministic)
 [[bench]]
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -55,6 +55,7 @@ headers = { workspace = true }
 hex = { workspace = true }
 hickory-resolver = { workspace = true }
 hostname = "0.4"
+ipnet = { version = "2", features = ["serde"] }
 notify = { workspace = true }
 pav_regression = { workspace = true }
 parking_lot = { workspace = true }
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -135,6 +135,7 @@ impl Default for ConfigArgs {
                 token_ttl_seconds: None,
                 token_cleanup_interval_seconds: None,
                 allowed_host: None,
+                allowed_source_cidrs: None,
             },
             secrets: Default::default(),
             log_level: Some(tracing::log::LevelFilter::Info),
@@ -637,6 +638,24 @@ impl ConfigArgs {
                     .token_cleanup_interval_seconds
                     .unwrap_or(default_token_cleanup_interval_seconds()),
                 allowed_hosts: self.ws_api.allowed_host.unwrap_or_default(),
+                allowed_source_cidrs: self
+                    .ws_api
+                    .allowed_source_cidrs
+                    .as_ref()
+                    .map(|cidrs| {
+                        cidrs
+                            .iter()
+                            .map(|s| {
+                                s.parse::<ipnet::IpNet>().map_err(|e| {
+                                    anyhow::anyhow!(
+                                        "invalid CIDR `{s}` in allowed-source-cidrs: {e}"
+                                    )
+                                })
+                            })
+                            .collect::<Result<Vec<_>, _>>()
+                    })
+                    .transpose()?
+                    .unwrap_or_default(),
             },
             secrets,
             log_level: self.log_level.unwrap_or(tracing::log::LevelFilter::Info),
@@ -1274,6 +1293,29 @@ pub struct WebsocketApiArgs {
     #[arg(long, env = "ALLOWED_HOST")]
     #[serde(rename = "allowed-host", skip_serializing_if = "Option::is_none")]
     pub allowed_host: Option<Vec<String>>,
+
+    /// Additional source IP ranges (CIDR notation) permitted to reach the
+    /// local HTTP/WebSocket API.
+    ///
+    /// By default, only loopback and RFC1918 / IPv6 ULA ranges are accepted.
+    /// Use this to grant access from VPN overlays you control (e.g. Tailscale:
+    /// `--allowed-source-cidrs 100.64.0.0/10`). Can be specified multiple times.
+    ///
+    /// SECURITY: Only add ranges you fully control. CGNAT space like
+    /// `100.64.0.0/10` is shared between subscribers of some ISPs (Starlink,
+    /// T-Mobile, many cable carriers) and is only safe on an overlay network
+    /// such as Tailscale or WireGuard. Anything that can reach the API port
+    /// can access your contract state, keys, and client API.
+    #[arg(
+        long = "allowed-source-cidrs",
+        env = "ALLOWED_SOURCE_CIDRS",
+        value_delimiter = ','
+    )]
+    #[serde(
+        rename = "allowed-source-cidrs",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub allowed_source_cidrs: Option<Vec<String>>,
 }
 
 /// Default telemetry endpoint (nova.locut.us OTLP collector).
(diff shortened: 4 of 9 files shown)


sanity:
## Review feedback addressed (commit e29a07b2)

Ran four parallel review agents + `codex review` before requesting review. Consolidated findings and fixes below. Re-requesting review.

### Blockers

**big-picture: compile break in ping app fixture.** `apps/freenet-ping/app/tests/common/mod.rs:150` had a `WebsocketApiArgs { .. }` struct literal without the new field. Fixed. (Original PR only grepped `crates/` for the fixture; the ping app was out of scope.)

**testing: middleware-level test missing.** Tests exercised `AllowedSourceCidrs::contains` in isolation, so an inverted `||` in `private_network_filter` would pass every test. Extracted the decision into a pure `is_source_allowed(ip, allowed)` function. The middleware is now a thin wrapper, and new tests cover the exact composition that ships: private-with-empty-allowlist (both v4 and v6), public-rejected-with-empty, configured-tailnet-accepts-inside-rejects-outside, IPv4-mapped-IPv6 accept path (`::ffff:100.64.0.1` matching `100.64.0.0/10`) **and** reject path (`::ffff:8.8.8.8` must still be rejected), and the IPv6 ULA case.

**testing: config-layer parse tests missing.** Added `build_with_cidrs` helper and five `#[tokio::test]`s: round-trip of a v4+v6 list, default-is-empty regression guard, malformed-value rejection with message inspection, `0.0.0.0/0` rejection, `::/0` rejection. These cover the CLI/env/TOML → `Vec<IpNet>` path end-to-end.

### Medium

**skeptical: `0.0.0.0/0` footgun.** Agreed this is exactly the footgun the PR is supposed to prevent. Added `validate_source_cidr` called from `ConfigArgs::build`. Refuses IPv4 prefixes < /8 and IPv6 prefixes < /16. `/8` still covers `10.0.0.0/8` (RFC1918) and `100.64.0.0/10` (CGNAT/Tailscale); `/16` still accommodates a `/48` tailnet, `/56` HE subnet, or `/64` home LAN. Error message names the offending CIDR, its prefix, and the minimum accepted, so ops can fix it without reading code. Tests: `validate_source_cidr_rejects_overly_broad_ipv4/_ipv6` and `validate_source_cidr_error_message_is_actionable`.

**skeptical: per-range vs lumped warn.** Changed the startup warn to one line per CIDR instead of a `?Vec<IpNet>` debug-format so journald/grep stays legible. Also reworded the message to drop the em-dash per project style.

**skeptical: `8.8.0.0/16`-style public-but-valid-prefix.** Intentionally not rejected: a user who types a public /16 on purpose (penetration testing, a private WAN with public IPs, etc.) is past the "are you sure" threshold we can enforce with a lint. The `/8` minimum and the prominent per-range warn log are the belt-and-braces. If this turns out to be a real footgun in practice, we can add a second warn level (or require an explicit `--i-know-this-is-public` flag) in a follow-up.

### Low / nits

**code-first: redundant normalization.** Cleaned up. `is_private_ip` already handles `::ffff:a.b.c.d` internally, so `is_source_allowed` only normalizes for the `.contains()` branch. Doc comment explains why the asymmetry is correct.

**code-first: add a comment about the ULA case being a CIDR-path-only test.** Added: the test for `fd7a:115c:a1e0::/48` now has a comment noting that ULA passes `is_private_ip` already, so the allowlist is a no-op for this case but the test documents intent and guards against a future narrowing of `is_private_ip`.

**big-picture: docs PR should explain loopback bind makes the allowlist a no-op.** Addressed in freenet/web#34 (commit b1821859): Option 2 step 1 now spells out that leaving `ws-api-address = "127.0.0.1"` means the CIDR knob can't fire because the filter isn't installed at all.

**skeptical: multiple bad CIDRs only surface the first error.** Left as-is. UX cost is small (restart once per bad entry), and the alternative (collect all errors) adds code for a case that shouldn't happen more than once per operator lifetime.

**skeptical: logged IP vs match_ip on rejection.** The current log prints the original wire IP, which is what ops want for correlating against packet captures. `match_ip` would be surprising for a `::ffff:` source. Keeping as-is.

### What I did NOT change

- No GH-API-level integration test firing a request through the full axum router with a spoofed `ConnectInfo`. The testing reviewer called this out as "nice to have, probably not worth the complexity" and I agree: the pure-function test covers the bug classes that matter (inverted operator, mapped-v6 bypass, empty-default-deny) without dragging in a real socket bind. Happy to add one if a reviewer here disagrees.

- No tests for comma-delimited CLI parsing (`value_delimiter = ','`). That path is clap's, and clap's own tests cover it; testing it here would be testing clap, not this PR.

### Local verification

```
cargo check -p freenet --tests   # clean
cargo clippy -p freenet --lib -- -D warnings   # clean
cargo test -p freenet --lib -- server:: config::tests::allowed_source
# 146 passed; 0 failed
```

Pre-existing clippy errors in `crates/core/src/wasm_runtime/delegate.rs` and `crates/core/tests/operations.rs` (wildcard_enum_match_arm) are on main and unrelated to this PR. Confirmed by running clippy on main itself.

[AI-assisted - Claude]

sanity:
## Root-caused and fixed the NAT CI failure (commit cd2cae76)

### Diagnosis

After the first review-fix push, `test_contract_operations_via_docker_nat` started failing deterministically with `ConnectivityFailed("Network did not reach 100% connectivity within 120s")` — three in-test retries, ~150-180s each, all the same error. Other PRs on the same runner pool passed NAT cleanly at the same time, so it wasn't infra flake.

Tracked it to an axum layer-ordering bug in the review-fix commit. Per `Router::layer` docs, layers execute **bottom-to-top** — the LAST `.layer(..)` call is the OUTERMOST and runs FIRST in request flow. The builder was:

```rust
ws_router
    .layer(Extension(allowed_hosts))
    .layer(Extension(allowed_source_cidrs))   // inner
    .layer(from_fn(private_network_filter))   // outer — runs FIRST
    .layer(TraceLayer::new_for_http())
```

Because the filter ran **before** `Extension(allowed_source_cidrs)` injected, `Extension<AllowedSourceCidrs>` extraction failed and every HTTP request returned 500. The Docker NAT test's ring-connectivity probe hits an HTTP endpoint on each node; with 500 responses the probe never saw a connected network and timed out at 120s. That's the full causal chain.

### Fix

Reorder so `Extension(allowed_source_cidrs)` is applied AFTER `from_fn`, making it outer:

```rust
ws_router
    .layer(Extension(allowed_hosts))
    .layer(from_fn(private_network_filter))
    .layer(Extension(allowed_source_cidrs))   // now outer — injects BEFORE filter
    .layer(TraceLayer::new_for_http())
```

`Extension(allowed_hosts)` stays innermost because it's consumed by the WebSocket upgrade handler inside `ws_router`, which is at the center of the onion.

### Regression test

Added `middleware_layer_stack_allows_configured_source`: builds the exact layer stack from `serve_client_api_in_impl`, fires synthetic requests via `tower::ServiceExt::oneshot` with `ConnectInfo` from both an allowlisted source (`100.64.0.1`) and a public one (`8.8.8.8`). Verified locally by temporarily reverting the layer order:

```
thread 'middleware_layer_stack_allows_configured_source' panicked
assertion `left == right` failed: allowlisted CGNAT source must reach the handler;
500 here means the middleware failed to extract the Extension
 left: 500
 right: 200
```

So the test reproduces the exact Docker NAT failure in milliseconds instead of ~8 minutes of CI. Any future reorder fails this test before it can waste a NAT run.

### Mea culpa

@testing-reviewer asked for an axum-level integration test on the first round. I declined, reasoning the pure-function `is_source_allowed` tests covered the bug classes. That was wrong. Pure-function extraction doesn't help if the wiring that calls it is broken — specifically the kind of bug where extraction fails and the pure function is never even reached. The regression test now exercises the wiring too. Adding this as a durable lesson for future reviews: when a reviewer asks for end-to-end middleware coverage on a security-sensitive filter, do it.

Also rebased onto latest main (version is now 0.2.45) while I was here.

[AI-assisted - Claude]

PR #3858, opened 2026-04-13
fix: allow cross-contract links from sandboxed webapps

sanity:
## Problem

A webapp running inside the Freenet gateway's sandboxed iframe (e.g. Delta) cannot link to pages in another Freenet contract without forcing the user to open the link in a new tab. Reported on Matrix by Ivvor:

> Also, why can't I link to other Freenet contracts from a Delta site without requiring the link to be opened in a new tab?

## Root cause

The shell's \`navigate\` postMessage handler in \`SHELL_BRIDGE_JS\` validated that the target path started with the currently loaded contract's prefix:

\`\`\`js
if (!contractPrefix || cleanPath.indexOf(contractPrefix) !== 0) return;
\`\`\`

Any same-origin link whose path leaves the current contract's \`/v[12]/contract/web/{key}/\` subtree is silently dropped. The only way users can reach another contract is \`target=\"_blank\"\`, which escapes the sandbox through \`allow-popups\` and ends up in a new tab — exactly what Ivvor is complaining about.

## Solution

Replace the current-contract prefix check with a contract-namespace shape check. The navigate handler now accepts any same-origin path matching \`/v[12]/contract/web/{key}/...\` including *other* contracts, then updates the cached \`contractPrefix\` so subsequent in-contract links from the newly loaded page are still validated against the right prefix.

**Security posture is unchanged:**

- The sandbox attributes themselves are **not** widened — the shell remains the sole code with top-level authority. No \`allow-top-navigation\`, no \`allow-same-origin\`. A test assertion enforces that \`allow-top-navigation\` does not appear anywhere in \`SHELL_BRIDGE_JS\`.
- Same-origin only (rejects cross-site).
- Shape check is the security boundary: it rejects anything outside the contract namespace, so this cannot become a ladder into \`/v1/node/...\`, \`/v1/delegate/...\`, or other gateway internals.
- Cross-contract click via the injected navigation interceptor is the same trust posture as middle-clicking the same link today (which already works via \`target=\"_blank\"\` + \`allow-popups\`).

## Testing

Six new unit tests in \`path_handlers.rs\` exercise a pure-function extraction of the shell's path validation, kept in sync with the JS:

- \`navigate_accepts_same_contract_path\` — no regression in the existing in-contract flow
- \`navigate_accepts_cross_contract_path\` — **primary regression test** for this bug
- \`navigate_accepts_v2_contract_path\` — v2 API path also accepted
[...]

Issue #3856, opened 2026-04-13
tests: add Playwright smoke test harness for shell + iframe postMessage contract


sanity:
Four recent PRs (#3842, #3849, #3852, #3854) have fixed bugs in \`crates/core/src/server/path_handlers.rs\` and \`client_api.rs\` that could only be validated by substring-matching the JavaScript strings they emit. Each PR adds the next string-level assertion; each review acknowledges "this should really be a browser test". The gap is real and now well-characterised: we have no way to exercise the shell bridge, the navigation interceptor, or the CSP against an actual browser runtime before a regression reaches users.

Symptoms we've shipped:

- **#3842** — CSP blocked \`/permission/pending\` polling in every webapp for an unknown length of time.
- **#3852** — target=\"_blank\" external links opened null-origin sandboxed popups and broke CORS on destination sites (reported by @lukors as freenet/river#208 after it reached production).
- **#3854** — shift/ctrl/middle-click semantics silently collapsed; middle-click wasn't even intercepted (\`click\` handler doesn't fire for \`auxclick\`).

None of these would have reached main if we had a Playwright smoke test that:

1. Boots a local freenet node with a trivial published webapp contract.
2. Loads the shell URL in a headless Chromium.
3. Interacts with the sandboxed iframe: clicks a cross-origin link (left, middle, shift), fires the \`/permission/pending\` poller, verifies no CSP violations in \`browser.console_messages\`.
4. Asserts expected postMessage payloads, expected window.open calls, and expected CSP headers.

**Scope sketch for the harness:**

- Playwright test directory at \`crates/core/tests/playwright/\` (or similar) with \`package.json\` + \`playwright.config.ts\`.
- A small Rust test helper that spawns a freenet node on a random port and publishes a fixture webapp containing links/fetches exercising the things we care about.
- A CI workflow that installs Playwright browsers (cached) and runs the tests on PRs touching \`crates/core/src/server/\`.

**Related PRs that would have been caught automatically:**

- freenet/freenet-core#3842 — missing \`connect-src 'self'\`
- freenet/freenet-core#3849 — (the fix for #3842, with a string-level pin)
- freenet/freenet-core#3852 — target=\"_blank\" cross-origin interception
- freenet/freenet-core#3854 — modifier keys + auxclick

Priority: **medium** (every JS-string change ships with a review note saying this should be a real test, and a fourth gap slipped through as recently as today).

[AI-assisted - Claude]

sanity:
[...]

Issue #3841, opened 2026-04-13
Deep-link reload for pushState'd contract subpaths returns 404


sanity:
## Problem

After #3840, the gateway shell bridge pushes browser history entries for in-contract navigation, so the address bar reflects the current subpage (e.g. `/v1/contract/web/KEY/news/`). Back/forward and URL display work correctly within a session.

However, if the user **reloads** or **bookmarks** one of those URLs, the server routes the non-sandbox request through `web_subpages` → `variable_content`, which tries to serve the path as a file and returns 404. The user loses the session and cannot share deep links.

## Expected

GET `/v1/contract/web/KEY/news/` (no `__sandbox=1`) should serve the shell page, with the iframe initially pointed at `/v1/contract/web/KEY/news/?__sandbox=1`.

## Approach

Plumb a `sub_path` parameter through `web_subpages` → `path_handlers::contract_home` → `shell_page`. When `web_subpages` receives a non-sandbox HTML-like path, route to the shell page render path (with auth token + cookie generation, same as `web_home`) and embed the sub-path in the iframe's `data-src`.

Requires:
1. Factoring the auth-token/cookie/CSP header logic out of `web_home` into a shared helper.
2. Extending `shell_page`/`contract_home` to accept `sub_path: Option<&str>`.
3. Updating `is_html_page` logic in `web_subpages` to route non-sandbox HTML paths through the shell helper.

## Context

Issue #3839 asked for pushState/popstate/URL updates, which landed in #3840. The PR description explicitly scoped deep-link reload out as a larger plumbing change. Filing this to track the follow-up.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:56:27.291220878+00:00
Issue hash: `0d0a30ac`

## Value Assessment

The fix for issue #3841 has been partially implemented in the codebase (commit `c74fcf1b`), but not completely as originally specified.

[...]

PR #3876, opened 2026-04-14
fix(server): redirect top-level HTML subpath loads to shell route

sanity:
## Problem

Users reported (via Matrix) that clicking a link from one Freenet site to another flipped the destination webapp into an Error state until the page was refreshed. Ivvor saw Delta's "Connected" badge turn red after following a Delta-to-Delta link; Lukas confirmed that pasting the same URL directly worked but clicking the link produced console errors.

Root cause: the gateway's routing splits on URL shape.

- \`/v1/contract/web/{key}/\` → \`web_home\` → issues a fresh auth token, wraps the content in the sandboxed iframe shell
- \`/v1/contract/web/{key}/{*path}\` → \`web_subpages\` → for a non-sandbox HTML request, calls \`variable_content\` and serves the raw HTML file with **no shell, no auth token, no cookie**

Every webapp that uses \`freenetBridge\` (Delta, River, etc.) reads \`authToken\` from \`location.search\` to build its WebSocket URL. A top-level document load of any contract sub-path — pasted deep link, bookmark, or a cross-contract link click that goes through \`window.location.assign\` (added in #3858) — lands on \`web_subpages\` with no auth token in the query string, and the WebSocket handshake fails immediately.

Directly pasting \`/v1/contract/web/KEY/\` (the canonical share URL shape) happens to hit the root route and works — which is why Lukas saw paste succeed but click fail.

A secondary bug: the cross-contract navigate handler in \`SHELL_BRIDGE_JS\` passed only \`cleanPath + cappedHash\` to \`window.location.assign\`, silently dropping \`resolved.search\`. Any query parameters a link carried (app-level routing args, invitation tokens) were lost across the hop.

## Solution

Two changes, both in \`crates/core/src/server/\`:

1. **\`web_subpages\` redirect** — when the request is a top-level document load (\`Sec-Fetch-Dest: document\`) for a contract HTML sub-path that is not itself a sandbox request, redirect to the shell root \`/v{N}/contract/web/{key}/\`. The shell route generates a fresh auth token, wraps the contract in the sandboxed iframe, and the webapp loads with working credentials. The browser preserves any URL fragment client-side across the redirect, which is sufficient for hash-routed apps (Delta). Sub-resource fetches (\`Sec-Fetch-Dest: iframe/empty/script/...\`) and non-HTML assets are unaffected.

2. **Cross-contract navigate handler** — preserve \`resolved.search\` in the \`window.location.assign\` call so query parameters on the link survive the hop.

Key insight: the redirect is driven by \`Sec-Fetch-Dest\`, which the browser sets automatically and scripts cannot spoof. This matches the existing \`serve_sandbox_response\` guard that redirects top-level loads of sandbox URLs to the shell. Same security boundary, symmetric behaviour.

A fuller fix would thread the sub-path into shell generation so the iframe loads the requested page instead of the contract root. Out of scope for this PR — the root redirect restores the baseline guarantee that every top-level contract load gets a working shell, which is what was actually broken for users. Path-based multi-page apps that depend on deep-link entry can be addressed in a follow-up.

## Testing

New unit tests in \`crates/core/src/server/client_api.rs\`:

- \`subpage_redirects_top_level_html_document_load_to_shell\` — pins the redirect path for HTML pages (including extensionless, directory-style, and \`.html\` suffixes) when \`Sec-Fetch-Dest: document\`.
- \`subpage_does_not_redirect_sub_resource_fetches\` — pins that non-HTML assets (JS, CSS, WASM, images) and non-document destinations (\`iframe\`, \`empty\`, \`script\`, ...) are **not** redirected, so pages can still load their sub-resources.
- \`subpage_does_not_redirect_sandbox_requests\` — pins that sandbox iframe requests flow through the existing sandbox pipeline instead of being redirected to the shell root (which would defeat in-contract multi-page navigation).

The redirect decision is extracted into \`should_redirect_subpage_to_shell\` so the test matrix covers the pure logic without needing to spin up the axum router.

New SHELL_BRIDGE_JS assertion:

- Pins that the cross-contract \`window.location.assign\` expression includes \`resolved.search\`, so the query-preservation contract is part of the JS regression net.

Existing \`navigate_*\` tests (11 of them) and the sandbox/shell CSP tests continue to pass.

## Fixes

Follow-up to #3858 (\`allow cross-contract links from sandboxed webapps\`). Addresses the Delta cross-contract link reports surfaced by Ivvor and Lukas on Matrix.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -268,16 +268,24 @@ async fn web_subpages(
     // response in the sandbox iframe. Otherwise `variable_content` serves
     // raw HTML with no `authToken` query parameter, breaking webapps that
     // read credentials from `location.search` (freenet/river#208 follow-up:
-    // Delta links hit "Connection failed" on the destination). Hash-routed
-    // webapps preserve their fragment across the redirect; path-based
-    // multi-page apps lose the sub-path but still get a working shell.
+    // Delta links hit "Connection failed" on the destination).
+    //
+    // The browser preserves any URL fragment across a 303 redirect, which
+    // is sufficient for hash-routed webapps (e.g. Delta). Path-based
+    // multi-page apps still lose the sub-path — this redirect restores
+    // the baseline guarantee that top-level contract loads get a working
+    // shell. A fuller fix for path-based deep linking (#3841) would
+    // thread the sub-path into shell generation.
+    //
+    // Clients without `Sec-Fetch-Dest` (curl, older browsers) intentionally
+    // fall through to `variable_content` — matches pre-PR behaviour and
+    // avoids redirect loops for non-browser clients.
     let fetch_dest = req_headers
         .get("sec-fetch-dest")
         .and_then(|v| v.to_str().ok())
         .unwrap_or("");
     if should_redirect_subpage_to_shell(is_sandbox, &last_path, fetch_dest) {
-        let shell_url = format!("/{}/contract/web/{key}/", api_version.prefix());
-        return Ok(axum::response::Redirect::to(&shell_url).into_response());
+        return redirect_to_shell_root(&key, api_version, query_string.as_deref());
     }
 
     let version_prefix = api_version.prefix();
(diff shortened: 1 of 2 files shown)


sanity:
## Review feedback addressed — 86125b81

Consolidated findings from Codex + 4 internal reviewers (code-first, testing, skeptical, big-picture). All substantive issues addressed in one follow-up commit.

**Must-fix**

1. **Server redirect drops query string** (Codex P2, skeptical H1, code-first #6) — the 303 target was `/v1/contract/web/KEY/` with no query component, silently undoing the matching JS-side `resolved.search` preservation. Factored into `redirect_to_shell_root(key, api_version, query_string)` which now preserves the query end-to-end, stripping only sensitive routing params (`__sandbox*`, `authToken*`).

2. **`authToken` injection surface via `resolved.search`** (code-first #6) — a malicious cross-contract href with `?authToken=attacker_value` would flow through the navigate bridge → server redirect → `shell_page` → iframe `location.search`, where a webapp reading credentials from the URL would pick it up. Added `authToken*` to the filter in `shell_page` (which is the last point any forwarded query reaches before interpolation into the iframe URL) as a belt-and-braces defence. New test `shell_page_strips_auth_token_from_forwarded_query` pins this.

3. **CRLF panic surface in `Redirect::to`** (skeptical L3) — a crafted path containing percent-encoded CRLF would reach `HeaderValue::try_from` inside `Redirect::to`, which panics on invalid header values. Added explicit empty-key guard + `ContractInstanceId::from_bytes` validation before interpolation. The same latent panic existed in the pre-existing `serve_sandbox_response` redirect — both now route through `redirect_to_shell_root`. New test `redirect_to_shell_root_rejects_invalid_key_instead_of_panicking`.

4. **End-to-end regression test missing** (testing #5, skeptical M5, big-picture #3) — added `web_subpages_redirects_top_level_document_load_to_shell` that calls `web_subpages` directly with the Delta failure-mode inputs (document load, mixed sensitive query, sandbox document variant), pinning status + `Location` + filtered query end-to-end.

5. **Source-order invariant** (skeptical M3) — added `web_subpages_sandbox_branch_runs_before_redirect_branch` which asserts on byte offsets inside the file, so a future refactor cannot reorder the sandbox short-circuit after the new redirect block.

6. **303 fragment-preservation pin** (skeptical M3, big-picture) — added `redirect_to_shell_root_uses_303_see_other_for_fragment_preservation` so a refactor to `Redirect::temporary`/`Redirect::permanent` can't silently change hash-routed-webapp semantics.

**Acknowledged, not fixed in this PR**

- **Path-based deep link reload** (big-picture): path-routed multi-page apps still lose the sub-path across the redirect (e.g. `/KEY/news/2024/article` lands on `/KEY/`). The PR restores the baseline guarantee (working shell + auth token) for hash-routed apps which is what the Delta report was about, but the full fix is #3841's approach — thread `sub_path` into `shell_page`. I've referenced #3841 in the code comment at `web_subpages`. Happy to repurpose that issue's scope as the follow-up vehicle.
- **`is_html_page` over-matching** (code-first #1, skeptical M1): the heuristic pre-dates this PR and gating the redirect on it widens its blast radius. Documented the trade-off inline at `web_subpages`; a stricter MIME-aligned check is a bigger refactor.
- **Missing `Sec-Fetch-Dest` falls through** (code-first #4): intentional — matches pre-PR behaviour for curl/wget/old browsers, avoids redirect loops for non-browser clients. Documented inline.
- **Playwright coverage**: the existing `SHELL_BRIDGE_JS.contains(...)` assertion + the end-to-end `web_subpages` test cover the regression surface without a browser. A browser-level smoke test belongs to #3856's harness work.

**Related**

Acknowledges and partially addresses #3841 (Deep-link reload for pushState'd contract subpaths). This PR restores the auth-token guarantee that #3841 also needed as a prerequisite; #3841 remains open for the sub-path-preservation work.

All local tests green (`cargo test server::client_api`, `cargo test server::path_handlers`) and clippy clean on the diff.

[AI-assisted - Claude]

Issue #3838, opened 2026-04-13
test: test_ping_blocked_peers flaky in merge queue (propagation timeout)


sanity:
## Problem

`test_ping_blocked_peers` (`freenet-ping-app::run_app_blocked_peers`) failed all 3 nextest retries in the merge-queue run for PR #3837:

https://github.com/freenet/freenet-core/actions/runs/24320981751

Each try ran for ~271s and ended with:
\`\`\`
[ERROR run_app_blocked_peers] Test failed: Updates did not fully propagate after 48 polls.
\`\`\`

The PR's CI against its branch head passed cleanly; the failure only reproduced in the merge-queue rebase onto main. This is the same failure pattern that #2254 and #3490 tried to address.

## Impact

Blocks merge-queue landing of PRs that don't touch routing/ping. PR #3837 (shell-page permission overlay, no core changes) hit this.

## Hypothesis

Updates not propagating across the 3-peer blocked topology within the 48-poll window suggests either (a) still-present timing assumptions in the poll loop, or (b) a real regression in update propagation behind NAT / blocked links. A `freenet:systematic-debugging` pass comparing passing and failing runs would clarify which.

## Next Step

Reproduce locally with \`cargo test -p freenet-ping-app --test run_app_blocked_peers\` and tail peer logs for the update path. If it's a timing issue, switch from polling to deterministic synchronization; if it's a real propagation bug, bisect.

[AI-assisted - Claude]

sanity:
Reopening — `test_ping_blocked_peers` failed again in CI on 2026-04-13 after this issue was closed earlier the same day.

Failure observed on PR #3867 commit `ad93b15f`. The same commit passed Unit & Integration on its first CI run (run [24368933816](https://github.com/freenet/freenet-core/actions/runs/24368933816), 9m 43s, green) and then failed Unit & Integration on a re-run (run [24370682972](https://github.com/freenet/freenet-core/actions/runs/24370682972)) triggered by adding the `review-override` label. **Same code, different result, no rebase between runs.**

PR #3867 changes are entirely in the delegate permission prompt UI:
- `crates/core/src/contract/user_input.rs` (UserInputPrompter trait)
- `crates/core/src/contract.rs` (delegate caller wiring)
- `crates/core/src/server/client_api/permission_prompts.rs` (HTML/JSON)
- `crates/core/src/server/path_handlers.rs` (shell-page JS)

[...]

Issue #3651, opened 2026-03-25
fix: test_ping_blocked_peers consistently fails on CI with GET timeouts


sanity:
## Problem

`test_ping_blocked_peers` fails consistently on some CI runners with GET operation timeouts. Failed 3/3 attempts on the same runner for PR #3609 (run 23559243719), while the same code passed on a different runner earlier (run 23557979258).

### Error pattern
```
[ERROR freenet_ping_app::ping_client] GET-related host error err=client error: error while executing operation in the network: GET operation timed out
[ERROR run_app_blocked_peers] Test failed: client error: error while executing operation in the network: GET operation timed out
```

On the 2nd run of the same CI job, different tests also failed (`test_put_triggers_update_for_subscribers`, `test_client_disconnect_triggers_upstream_unsubscribe`), confirming runner-dependent timing sensitivity.

### Evidence this is pre-existing
- Same branch, same code passed on run 23557979258 (different runner)
- The only diff between passing and failing commits is docs/comments and a unit test — zero behavioral changes
- Related to #3490 (timing-fragile integration tests)

### Root cause hypothesis
The integration tests start real nodes on loopback IPs with hardcoded timeouts. On slow CI runners (Ubicloud), ring promotion and operation routing can take longer than the test's operation-level timeouts (30-60s for individual GETs), even though the overall test timeout is generous (300-420s).

The specific issue is that `test_ping_blocked_peers` waits for ring connections but then immediately attempts GET operations with a relatively short timeout. If the ring connections are barely established and the routing tables aren't fully populated, the GET can fail to route.

### Suggested fix
1. Add operation-level retry with backoff for GET operations in the ping test client, similar to how `test_put_triggers_update_for_subscribers` already retries
2. Or increase the per-operation timeout for GET in `test_ping_blocked_peers`
3. Consider adding a "routing ready" check after ring connection establishment before attempting operations

[AI-assisted - Claude]

Issue #3456, opened 2026-03-07
fix: ping integration tests flaky on slow CI runners due to OPERATION_TTL vs WASM compilation


sanity:
## Problem

Ping integration tests (`test_ping_application_loop`, `test_small_network_get_failure`, `test_ping_blocked_peers`) are flaky on CI runners. They fail with `GET operation timed out` from the node-internal operation timeout.

## Root Cause

`OPERATION_TTL` is 60 seconds (`config/mod.rs:43`). When multiple nodes compile the same WASM contract simultaneously, they serialize on the shared `Engine` mutex (Arc<Mutex<EngineInner>>). Each compilation takes several seconds, and cumulative wait time can exceed 60s, causing the GET operation state machine to time out before it even gets to execute the contract.

The problem is amplified by:
- nextest retries: each retry spawns fresh nodes that must recompile
- CI runners with limited CPU: compilation takes longer
- Shared Engine mutex: PR #2937 introduced shared module cache, but compilation still serializes

## Evidence

- `test_ping_application_loop`: TRY 1 timeout at 240s (test timeout), TRY 2 fail at 85s, TRY 3 fail at 85s — the 85s elapsed means the node's 60s OPERATION_TTL fired
- `test_small_network_get_failure`: Same pattern — 45s client timeout (now increased to 120s) masks the real issue which is the node-side 60s OPERATION_TTL
- Tests pass reliably on fast machines; fail intermittently on slow CI runners
- 5 other tests were `FLAKY` in the same CI run (passed on retry), confirming CI runner resource contention

## Possible Solutions

1. **Pre-compile WASM in test setup**: Compile the contract WASM once before starting multi-node tests, so nodes only need to load from cache
2. **Make OPERATION_TTL configurable**: Allow tests to set a longer TTL via `NodeConfig`
3. **Warm the module cache**: After starting nodes, do a single PUT on the gateway and wait for compilation before starting the multi-node test scenario
4. **Separate compilation from the operation timeout**: The WASM compilation time shouldn't count against the operation TTL — the operation hasn't even started routing yet

Option 4 is the most correct but highest effort. Option 3 is the lowest effort and would fix most cases.

## Affected Tests

- `apps/freenet-ping/app/tests/run_app.rs::test_ping_application_loop`
- `apps/freenet-ping/app/tests/test_small_network_get_issue.rs::test_small_network_get_failure`
- `apps/freenet-ping/app/tests/run_app_blocked_peers.rs::test_ping_blocked_peers`

[...]

Issue #3142, opened 2026-02-20
fix: flaky test_ping_blocked_peers suite in CI


sanity:
## Problem

The `freenet-ping-app::run_app_blocked_peers` test suite is flaky in CI. Different tests in the suite fail on different runs:

- PR #3139: `test_ping_blocked_peers` failed after 3 retries with stretto cache channel errors (`fail to receive msg from insert buffer: receiving on an empty and disconnected channel`)
- PR #3135: `test_ping_blocked_peers_solution` failed with a 60s timeout

These failures are not related to the PRs' changes — the blocked-peers test suite is intermittently failing on unrelated PRs.

## Evidence

```
[ERROR stretto::cache::sync] fail to handle insert event: fail to receive msg from channel: 
fail to receive msg from insert buffer: receiving on an empty and disconnected channel
```

```
TRY 3 FAIL [70.403s] freenet-ping-app::run_app_blocked_peers test_ping_blocked_peers
Summary [267.202s] 3 tests run: 2 passed, 1 failed, 11 skipped
```

## Impact

Blocking unrelated PRs from merging. The test has retry logic (3 attempts) but still fails intermittently.

## Suggested investigation

- The stretto cache errors suggest the cache's internal channel is being dropped before operations complete
- The 60s+ test times suggest possible timeouts in the simulation
- May be related to test isolation — the tests run serially but may be sharing global state

[AI-assisted - Claude]

PR #3893, opened 2026-04-16
fix(connect): promote joiner on accept-after-uphill-rejection retry

sanity:
## Problem

`test_ping_blocked_peers` (merge-queue integration test) has been
flaking intermittently. Root-cause investigation this pass found one
concrete promotion-drop bug in the CONNECT retry paths, which matches
the `RING_TRANSPORT_DESYNC` signature from the most recent reopen of
#3838.

Scenario: Node1 and Node2 block each other at transport level, so
contract-state updates must propagate via the gateway. One of the two
nodes would silently fail to subscribe on the gateway, and UPDATE
fan-out never reached it.

### Root cause

When the gateway's CONNECT processing takes the "accept locally after
uphill rejection" (or "accept locally after ConnectFailed") retry
branch, the relay re-invokes `handle_request`, which returns **both**:

- `accept_response` — the `ConnectResponse` to send upstream
- `expect_connection_from` — the joiner to promote via
  `NodeEvent::ConnectPeer` / `ExpectPeerConnection`

The happy-path branch at `process_message` consumed both. The two
retry branches had drifted apart and consumed only `accept_response`.
`expect_connection_from` was silently dropped, so
`NodeEvent::ConnectPeer` was never emitted on the acceptor side and
the joiner stayed in `transient_connections` forever.

Downstream, `get_peer_by_addr` (used by
`register_downstream_subscriber` and broadcast fan-out) only walks
ring connections. When the joiner later subscribed, the gateway
logged `could not find peer to register interest` and returned
without recording the subscription — but still sent
`SubscribeResponse { subscribed: true }` back. The node thought it
was subscribed; the gateway never broadcast UPDATEs to it.

Evidence in a failing run log:

```
[INFO] connect: acceptance issued at terminus … joiner_pub_key=2gWenzaUtDPfbbQCx
[INFO] connect: relay accepting locally after uphill rejection
   <-- no "acceptor accepted joiner, initiating hole punch" for this joiner -->
```

This matches the `RING_TRANSPORT_DESYNC: transport has connections
but ring topology is empty` error observed in TRY 2 of the most
recent #3838 reopen (`p2p_protoc.rs:884`) — that error fires exactly
when `transport_connections > 0 && ring_connections == 0`, i.e. when
the acceptor never emits ConnectPeer for a joiner whose transport
connection it already accepted.

## Approach

Combine `accept_response` and `expect_connection_from` on
`RelayActions` into a single `AcceptOutcome { response, joiner }`
value. The type system now makes it **impossible** to consume the
response without also being handed the joiner, so all three
acceptance sites (happy path, `Rejected` retry, `ConnectFailed`
retry) are forced through the same `dispatch_expect_connection_from`
helper. Future acceptance sites inherit the invariant for free.

The helper emits `ExpectPeerConnection` + `ConnectPeer`
(`is_gw: false`) exactly as the happy path did, then spawns the
hole-punch outcome-logging task.

## Testing

- New unit test `relay_retry_acceptance_bundles_joiner_with_response`
  exercises the exact `process_message::Rejected` retry-accept flow
  at the data level: forward → clear `forwarded_to` → retry → accept.
  Asserts that the returned `AcceptOutcome` carries the joiner
  alongside the response. This test wouldn't compile before the
  refactor because the two fields were separate; it now pins the
  invariant.
- All 51 pre-existing `operations::connect::tests::*` unit tests
  updated for the new field layout — green.
- All 301 `operations::*` lib tests green.
- `cargo fmt` + `cargo clippy --lib --tests -- -D warnings` clean.
- Ran `cargo nextest run -p freenet-ping-app --features testing
  -E 'test(test_ping_blocked_peers)'` 5x locally — all PASSED
  (28.9s, 15.9s, 11.9s, 11.9s, 11.9s) vs. ~40% failure rate before.

## Why didn't CI catch this earlier?

`test_ping_blocked_peers` was the canary, but nextest's `retries=2`
masked a ~40% per-run failure rate as ~6% effective, leaving the
test looking "flaky but mostly passing." The CI gap closed here is
**structural**: the new regression test pins the data invariant, and
the type-level refactor prevents any future acceptance path from
silently dropping the ring promotion.

## Limits and follow-ups

This is a structural fix for the specific promotion-drop bug in the
retry paths — it does NOT claim to be the only remaining cause of
`test_ping_blocked_peers` flakiness. The test has been "fixed" four
times prior (#3142, #3456, #3651, #3838) and the most recent reopen
specifically called for deeper investigation beyond point fixes. I
should fix the retry-drop signature (TRY 2's `RING_TRANSPORT_DESYNC`)
and does not rule out the other failure modes in that reopen thread.

Intentionally leaving #3838 open until post-merge CI observation over
5–10 merge-queue runs confirms the flake is gone. Follow-ups:

- Filing issue for a direct-runner simulation test that asserts an
  accepted joiner transitions transient → ring on the acceptor (end-
  to-end invariant currently covered only by the flaky ping canary).
- Filing issue to reduce `test_ping_blocked_peers`'s nextest retry
  count so any re-emergence surfaces immediately rather than being
  masked by retries.

## Fixes (partial)

Addresses #3838 (the retry-drop signature). Will close once post-merge
observation confirms the full flake is resolved.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/connect.rs
+++ b/crates/core/src/operations/connect.rs
@@ -1621,12 +1621,16 @@ impl Operation for ConnectOp {
                             .await?;
                     }
 
-                    // Promote the joiner BEFORE the forward/response sends: once the
-                    // joiner receives our Response it may immediately start subscribing
-                    // through us, and `get_peer_by_addr` on the acceptor side only
-                    // finds ring connections. Near-terminus acceptance sets both
-                    // `forward` AND `accept`, so we dispatch here and fall through to
-                    // the forward + response sites below.
+                    // Promote the joiner BEFORE sending the `Response` upstream:
+                    // once the joiner receives our Response it may immediately
+                    // start subscribing through us, and `get_peer_by_addr` on
+                    // the acceptor side only finds ring connections. Dispatch
+                    // here (using a ref borrow) and let the `accept_response`
+                    // site below consume the owned value for the actual send.
+                    // (The `forward` branch below is uphill to the next relay,
+                    // not to the joiner, so its ordering relative to dispatch
+                    // doesn't matter — it's placed here mainly to match the
+                    // near-terminus "forward + accept" code path.)
                     if let Some(ref accept) = actions.accept {
                         dispatch_expect_connection_from(op_manager, self.id, accept.joiner.clone())
                             .await?;
@@ -2054,6 +2058,18 @@ impl Operation for ConnectOp {
                                 acceptor = %accept.response.acceptor.pub_key(),
                                 "connect: relay accepting locally after uphill rejection"
                             );
+                            // Emit `connect_response_sent` telemetry, matching the
+                            // happy-path accept branch (#3893 review: close the telemetry
+                            // gap between accept sites so analytics can count retry
+                            // acceptances alongside first-pass acceptances).
+                            if let Some(event) = NetEventLog::connect_response_sent(
+                                &self.id,
+                                &op_manager.ring,
+                                accept.response.acceptor.clone(),
+                                accept.joiner.clone(),
+                            ) {
+                                op_manager.ring.register_events(Either::Left(event)).await;
+                            }
                             dispatch_expect_connection_from(op_manager, self.id, accept.joiner)
                                 .await?;
                             let response_msg = ConnectMsg::Response {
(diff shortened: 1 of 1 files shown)


iduartgomez:
Good catch

Issue #2919, opened 2026-02-08
feat: self-service diagnostic report lookup web page


sanity:
## Summary

Feature request from community member (dredd) in Matrix: add a web page to the report server where users can paste a diagnostic report code and get a human-readable summary of their connection state.

## Motivation

Currently users paste raw terminal logs into Matrix chat for support. A self-service tool would:
- Let users quickly understand their own connection issues
- Reduce noise in the support channel  
- Enable community support testers to help diagnose issues

## Privacy Considerations

**Important:** Report codes are publicly visible in Matrix chat, so the summary page must be careful about what it exposes. Specifically it should NOT reveal:
- IP addresses or peer addresses
- Hostnames or OS details
- Full log contents

It SHOULD show:
- Connection count and status (connected/disconnected)
- Detected issue categories (NAT failure, gateway failure, keepalive issues, etc.)
- Freenet version
- General guidance based on detected issues

## Implementation Notes

The analysis logic already exists server-side (issue classification, connection status extraction). This would be a simple web endpoint on the report server (nova.locut.us) that takes a code and returns a formatted summary.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-02-18T19:55:15.581405115+00:00
Issue hash: `d6b6fda3`

## Value Assessment

The diagnostic-report CLI (`ReportCommand` in `crates/core/src/bin/commands/report.rs:24-181`) already bundles system info, version metadata, filtered logs, optional config dumps, and the node’s `network_status` before uploading everything to `https://nova.locut.us/api/reports`. Every user who runs this command produces a report code that the server can tie back to those anonymized diagnostics, but today volunteers still ask for raw logs in Matrix so they can manually read connection counts, NAT/gateway failures, and keepalive errors. That friction affects anyone hitting connectivity trouble, so a human-readable page that just displays the sanitized summary the server already derives (without leaking peers, hostnames, or full logs) would dramatically lower support noise and make the same information instantly accessible to the larger Freenet community.

[...]

PR #3898, opened 2026-04-17
fix(report): surface WS query failures instead of silent None

### LINKED ISSUE #3897: feat(report): always capture node state — NodeDiagnostics is silently missing when collection fails
## Problem

Diagnostic reports ([`crates/core/src/bin/commands/report.rs`](crates/core/src/bin/commands/report.rs)) attempt to include live node state via `NodeDiagnostics` over the local WebSocket API (`collect_network_status`, [report.rs:229-256](crates/core/src/bin/commands/report.rs#L229-L256)). When that query fails — for any reason — the field is silently set to `None` and the report contains no indication of why.

This happened on report **FVQ7X3** (Raspberry Pi 4, aarch64, v0.2.46): the node was clearly running (the user was actively using it and complaining about slow GETs with ~4 peers), but `network_status: null`. The entire 30-min log capture was 122 lines of a single warning:

```
WARN freenet::transport::connection_handler: Channel overflow: dropped packets
  (bandwidth limit may be too high or receiver too slow) total_dropped=N elapsed_secs=10
```

No peer count, no connection states, no op traces — nothing to confirm or rule out hypotheses about why the peer count collapsed. The *one* piece of data that would have answered "why 4 peers?" (`connections_by_location.len()`, `pending_reservations`, subscription tree state) was the field that silently dropped out.

## Likely failure modes for the missing diagnostics

Any of these produce the same silent `null`:

1. **5-second timeout is too tight on a loaded node.** `WS_TIMEOUT_SECS = 5` ([report.rs:34](crates/core/src/bin/commands/report.rs#L34)) — on a resource-constrained device where the event loop is saturated (exactly the case we most need diagnostics for), the `NodeQuery::NodeDiagnostics` handler may not respond in time.
2. **IPv6/IPv4 binding mismatch.** `ws-api-address = \"::\"` is default; the collector connects to `ws://127.0.0.1:{port}` ([report.rs:659](crates/core/src/bin/commands/report.rs#L659)). On dual-stack Linux this works, but any config where IPV6_V6ONLY=1 or the node binds v6-only silently breaks the query.
3. **Node is running but in a degraded state** (WS task backlogged, rejecting new clients, etc.).

In all three cases the report contains no signal at all.

## Proposed fix

1. **Record the failure reason.** When `collect_network_status` fails or times out, put the error string into the report instead of `None`, e.g. `network_status: \"error: connection timed out after 5s\"` or a dedicated `network_status_error: Option<String>` field. The maintainer looking at the report needs to know whether the node was off, the query timed out, or the WS port was wrong.
2. **Longer timeout + retry.** Bump `WS_TIMEOUT_SECS` to at least 15s, and retry once on timeout. Diagnostic collection is a one-shot operation, not a latency-sensitive path.
3. **Try both v4 and v6 loopback.** `[::1]` then `127.0.0.1` (or vice versa) before giving up.
4. **Fallback snapshot from the event log on disk** when the WS query fails. The `_EVENT_LOG` file ([config `event_log` path](crates/core/src/bin/commands/report.rs#L229-L256)) contains connection events; a last-N-entries snapshot would at least let us reconstruct approximate peer count and recent operations when the node is too busy to answer live queries.
5. **Surface it in the `print_summary` output too.** Currently reads \"not running or unreachable\" — say *which*, and show the error.

## Why this matters

This is the third diagnostic report in two weeks where `network_status` was `null`, and each time it's the most load-bearing field for debugging. A silent `None` on a report from a struggling node is the worst possible outcome — we can't distinguish \"node not running\" from \"node running but unable to answer\", and the latter is the interesting case.

## Testing

- Unit test: simulate WS query timeout, assert the report contains the failure reason (not silent `None`).
- Unit test: simulate WS query connection refused, assert the report distinguishes \"not running\" from \"running but slow\".
- Manual: run `freenet service report` on a node whose WS port is wrong in config — verify the report explains why.

---

Context: observed while analyzing report FVQ7X3 on 2026-04-17. See also #2919 for the self-service lookup page where these fields surface to users.

[AI-assisted - Claude]

sanity:
## Problem

Diagnostic reports (`crates/core/src/bin/commands/report.rs`) attempt to include live node state via `NodeDiagnostics` over the local WebSocket API. When that query fails — for any reason — the field is silently set to `None` and the report contains no indication of why.

Report **FVQ7X3** (Raspberry Pi 4, aarch64, v0.2.46) exhibited this exactly: the user was actively complaining about slow GETs with ~4 peers, but `network_status: null` and the log capture was 122 lines of a single "Channel overflow: dropped packets" warning. The one piece of data that would have answered "why 4 peers?" (`connections_by_location.len()`, `pending_reservations`, subscription tree state) was the field that silently dropped out.

Three failure modes all produce the same silent `null`:

1. **5-second timeout too tight** on a loaded node where the event loop is saturated (exactly the case we most need diagnostics for).
2. **IPv4/IPv6 mismatch:** `ws-api-address = "::"` default with `IPV6_V6ONLY=1` breaks the hardcoded `ws://127.0.0.1:PORT` connect.
3. **Node running in a degraded state** (WS task backlogged, rejecting new clients).

In all three cases the report contains no signal at all.

## Solution

1. **Add `network_status_error: Option<String>`** to `DiagnosticReport`. When the query fails, the field records the concrete reason (`timed out after 15s`, `connection refused`, etc.) instead of silently dropping the data. Optional and `skip_serializing_if = "Option::is_none"` so existing server deserializers are unaffected.
2. **Bump `WS_TIMEOUT_SECS` from 5s to 15s** and retry once on timeout. Connection-refused is not retried — that indicates the node isn't running on that host.
3. **Try both `127.0.0.1` and `[::1]`** in order, so `ws-api-address = "::"` with v4-only or v6-only configurations doesn't silently drop the query.
4. **Update `print_summary`** to distinguish "running" / "unreachable (reason)" / "not running or unreachable", so the user sees why the query failed instead of the ambiguous "not running or unreachable".

The issue also suggested a fallback snapshot from the event log on disk; I've intentionally scoped this PR to items 1–4 (surfacing the failure and closing the two specific connection gaps) and left the event-log fallback for a follow-up if this fix proves insufficient in practice.

## Testing

Three new unit tests in `crates/core/src/bin/commands/report.rs`:

- `test_query_with_fallback_connection_refused_reports_error` — binds nothing to a port, asserts `query_with_fallback` returns `Err` mentioning both loopback hosts and the port. This would fail against `main`, which returns `None` silently.
- `test_query_with_fallback_timeout_reports_error` — accepts a TCP connection but never completes the WebSocket handshake, asserts the error string identifies "timed out" and the port. Exercises the 15s timeout path end-to-end.
- `test_network_status_error_serialization` — confirms the new field is omitted from JSON when `None` (wire compat) and round-trips correctly when `Some`.

All 54 tests in `cargo test -p freenet --bin freenet` pass. `cargo fmt` + `cargo clippy --bin freenet --all-targets -- -D warnings` clean.

## Fixes

Closes #3897

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/report.rs
+++ b/crates/core/src/bin/commands/report.rs
@@ -248,6 +248,10 @@ impl ReportCommand {
     /// diagnostics on success, or a human-readable error string on failure
     /// so the report can record *why* the field is missing rather than
     /// silently dropping it.
+    ///
+    /// Worst-case duration is `LOOPBACK_HOSTS.len() * (WS_RETRY_ATTEMPTS + 1) * WS_TIMEOUT_SECS`
+    /// seconds. Prints progress so the user isn't left staring at a blank
+    /// terminal during a degraded-node query.
     fn collect_network_status(&self, config_content: &Option<String>) -> Result<String, String> {
         let ws_port = config_content
             .as_ref()
@@ -257,7 +261,21 @@ impl ReportCommand {
         let rt = tokio::runtime::Runtime::new()
             .map_err(|e| format!("failed to create tokio runtime: {e}"))?;
 
-        rt.block_on(query_with_fallback(ws_port, WS_RETRY_ATTEMPTS))
+        let worst_case_secs =
+            LOOPBACK_HOSTS.len() as u64 * (WS_RETRY_ATTEMPTS as u64 + 1) * WS_TIMEOUT_SECS;
+        print!("  Querying local node diagnostics (up to {worst_case_secs}s)... ");
+        io::stdout().flush().ok();
+
+        let result = rt.block_on(query_with_fallback(
+            ws_port,
+            WS_RETRY_ATTEMPTS,
+            StdDuration::from_secs(WS_TIMEOUT_SECS),
+        ));
+        match &result {
+            Ok(_) => println!("ok"),
+            Err(e) => println!("unreachable ({e})"),
+        }
+        result
     }
 
     fn get_user_message(&self) -> Result<Option<String>> {
@@ -663,17 +681,22 @@ fn parse_ws_port_from_config(config: &str) -> Option<u16> {
 ///
 /// Retries are timeout-only: connection-refused errors are reported
 /// immediately since they indicate the node is not running on that host.
-async fn query_with_fallback(port: u16, retry_attempts: u32) -> Result<String, String> {
+///
+/// The `per_attempt_timeout` parameter exists for tests; production callers
+/// pass `Duration::from_secs(WS_TIMEOUT_SECS)`. Worst-case duration is
+/// `LOOPBACK_HOSTS.len() * (retry_attempts + 1) * per_attempt_timeout`.
+async fn query_with_fallback(
+    port: u16,
+    retry_attempts: u32,
+    per_attempt_timeout: StdDuration,
+) -> Result<String, String> {
     let mut errors: Vec<String> = Vec::new();
     let total_attempts = retry_attempts.saturating_add(1);
 
     for host in LOOPBACK_HOSTS {
         for attempt in 0..total_attempts {
-            match tokio::time::timeout(
-                StdDuration::from_secs(WS_TIMEOUT_SECS),
-                query_node_diagnostics(host, port),
-            )
-            .await
+            match tokio::time::timeout(per_attempt_timeout, query_node_diagnostics(host, port))
+                .await
             {
                 Ok(Ok(diag)) => return Ok(diag),
                 Ok(Err(e)) => {
@@ -684,7 +707,8 @@ async fn query_with_fallback(port: u16, retry_attempts: u32) -> Result<String, S
                 }
                 Err(_) => {
                     errors.push(format!(
-                        "{host}:{port}: timed out after {WS_TIMEOUT_SECS}s (attempt {}/{})",
+                        "{host}:{port}: timed out after {:?} (attempt {}/{})",
+                        per_attempt_timeout,
                         attempt + 1,
                         total_attempts
                     ));
@@ -974,6 +998,11 @@ stack backtrace:
         assert_eq!(original_size, panic_content.len() as u64);
     }
 
+    /// Short per-attempt timeout for tests that exercise the timeout path.
+    /// 200ms is long enough to avoid CI flakes but short enough that a
+    /// retry test can run multiple attempts in under a second.
+    const TEST_TIMEOUT: StdDuration = StdDuration::from_millis(200);
+
     /// Regression for issue #3897: when no node is listening, the report must
     /// surface a concrete error instead of silently setting `network_status`
     /// to `None`. Uses a port that nothing is bound to so `connect_async`
@@ -985,7 +1014,7 @@ stack backtrace:
         // explanation and the test is still telling us something useful.
         let port = 59111;
 
-        let result = query_with_fallback(port, 0).await;
+        let result = query_with_fallback(port, 0, TEST_TIMEOUT).await;
 
         let err = result.expect_err("no listener → must return Err, not silent None");
         assert!(
@@ -1017,19 +1046,20 @@ stack backtrace:
         let port = listener.local_addr().unwrap().port();
 
         // Accept the connection but never write any handshake bytes, so
-        // `connect_async` blocks until WS_TIMEOUT_SECS fires. Holding
-        // `_handle` keeps the accept task alive for the duration of the test.
+        // `connect_async` blocks until the per-attempt timeout fires.
+        // Holding `_handle` keeps the accept task alive.
         let _handle = tokio::spawn(async move {
-            let (_stream, _) = listener.accept().await.unwrap();
-            tokio::time::sleep(StdDuration::from_secs(60)).await;
+            loop {
+                let (_stream, _) = match listener.accept().await {
+                    Ok(v) => v,
+                    Err(_) => break,
+                };
+                // Keep the stream alive so the client's handshake hangs.
+                std::mem::forget(_stream);
+            }
         });
 
-        // Outer wrapper bounds total test time: WS_TIMEOUT_SECS (15s) must
-        // fire once on the 127.0.0.1 attempt, so 20s leaves margin without
-        // exceeding CI's default timeouts.
-        let result = tokio::time::timeout(StdDuration::from_secs(20), query_with_fallback(port, 0))
-            .await
-            .expect("test wrapper timed out before WS_TIMEOUT_SECS fired");
+        let result = query_with_fallback(port, 0, TEST_TIMEOUT).await;
 
         let err = result.expect_err("handshake never completes → must return Err");
         assert!(
@@ -1039,6 +1069,57 @@ stack backtrace:
         assert!(
             err.contains(&port.to_string()),
             "error should include the port number, got: {err}"
+        );
+    }
+
+    /// Regression for issue #3897: verify `retry_attempts > 0` actually
+    /// causes a second attempt when the first times out. The production
+    /// config passes `WS_RETRY_ATTEMPTS=1`, so this path must be exercised
+    /// by the test suite.
+    #[tokio::test]
+    async fn test_query_with_fallback_retries_on_timeout() {
+        use std::sync::Arc;
+        use std::sync::atomic::{AtomicUsize, Ordering};
+        use tokio::net::TcpListener;
+
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let port = listener.local_addr().unwrap().port();
+
+        let accepts = Arc::new(AtomicUsize::new(0));
+        let accepts_clone = accepts.clone();
+
+        // Accept every connection and hang (never complete handshake).
+        let _handle = tokio::spawn(async move {
+            loop {
+                let (_stream, _) = match listener.accept().await {
+                    Ok(v) => v,
+                    Err(_) => break,
+                };
+                accepts_clone.fetch_add(1, Ordering::SeqCst);
+                std::mem::forget(_stream);
+            }
+        });
+
+        // retry_attempts=1 → 2 attempts per host × 1 host that accepts = 2 timeouts.
+        // ([::1] gets connection-refused immediately, not counted in accepts.)
+        let result = query_with_fallback(port, 1, TEST_TIMEOUT).await;
+        assert!(result.is_err());
+
+        // Give the accept task a moment to register both connections.
+        tokio::time::sleep(StdDuration::from_millis(50)).await;
+
+        let total_accepts = accepts.load(Ordering::SeqCst);
+        assert_eq!(
+            total_accepts, 2,
+            "retry_attempts=1 must produce 2 v4 accepts (initial + 1 retry); got {total_accepts}"
+        );
+
+        // Error string must mention both attempts so operators can see the
+        // retry actually happened.
+        let err = result.unwrap_err();
+        assert!(
+            err.contains("attempt 1/2") && err.contains("attempt 2/2"),
+            "error should identify both attempts, got: {err}"
         );
     }
 
--- a/crates/core/src/bin/commands/report.rs
+++ b/crates/core/src/bin/commands/report.rs
@@ -30,8 +30,17 @@ const MAX_TOTAL_LOG_SIZE: usize = 2 * 1024 * 1024;
 const MAX_LINE_LENGTH: usize = 10 * 1024;
 /// Default WebSocket API port
 const DEFAULT_WS_API_PORT: u16 = 7509;
-/// Timeout for WebSocket connection and queries
-const WS_TIMEOUT_SECS: u64 = 5;
+/// Timeout for a single WebSocket connect+query attempt. A struggling node
+/// is exactly the case we most need diagnostics for, so err on the side of
+/// waiting longer rather than dropping the field silently.
+const WS_TIMEOUT_SECS: u64 = 15;
+/// Number of additional retry attempts after the initial query times out.
+/// Connection refused is not retried — that signals the node isn't running.
+const WS_RETRY_ATTEMPTS: u32 = 1;
+/// Loopback hosts to try, in order. Both are attempted so that
+/// `ws-api-address = "::"` combined with `IPV6_V6ONLY=1` (or a v4-only
+/// bind) doesn't silently drop the diagnostics.
+const LOOPBACK_HOSTS: &[&str] = &["127.0.0.1", "[::1]"];
 
 #[derive(Args, Debug, Clone)]
 pub struct ReportCommand {
@@ -64,8 +73,13 @@ pub struct DiagnosticReport {
     pub logs: LogContents,
     /// Config file contents (if available)
     pub config: Option<String>,
-    /// Network status (if node is running)
+    /// Network status (if node is running and responded successfully)
     pub network_status: Option<String>,
+    /// Reason `network_status` is missing (connection refused, timeout, etc.).
+    /// Populated whenever `network_status` is `None`, so the report never
+    /// silently drops the field without telling us why.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub network_status_error: Option<String>,
     /// User's problem description
     pub user_message: Option<String>,
 }
@@ -165,7 +179,10 @@ impl ReportCommand {
 
         let logs = self.collect_logs(config_dirs.log_dir().map(Path::to_path_buf))?;
         let config = self.collect_config(&config_dirs.config_dir());
-        let network_status = self.collect_network_status(&config);
+        let (network_status, network_status_error) = match self.collect_network_status(&config) {
+            Ok(diag) => (Some(diag), None),
+            Err(e) => (None, Some(e)),
+        };
         let user_message = self.get_user_message()?;
 
         let client_timestamp = chrono::Utc::now().to_rfc3339();
@@ -177,6 +194,7 @@ impl ReportCommand {
             logs,
             config,
             network_status,
+            network_status_error,
             user_message,
         })
     }
@@ -226,33 +244,20 @@ impl ReportCommand {
         None
     }
 
-    fn collect_network_status(&self, config_content: &Option<String>) -> Option<String> {
-        // Try to query the local node's WebSocket API
-        // This is optional - if the node isn't running, we just skip this
-
-        // Parse the WebSocket port from config, or use default
+    /// Query the local node for live diagnostics. Returns the serialized
+    /// diagnostics on success, or a human-readable error string on failure
+    /// so the report can record *why* the field is missing rather than
+    /// silently dropping it.
+    fn collect_network_status(&self, config_content: &Option<String>) -> Result<String, String> {
         let ws_port = config_content
             .as_ref()
             .and_then(|c| parse_ws_port_from_config(c))
             .unwrap_or(DEFAULT_WS_API_PORT);
 
-        // Create a runtime for the async WebSocket query
-        let rt = match tokio::runtime::Runtime::new() {
-            Ok(rt) => rt,
-            Err(_) => return None,
-        };
+        let rt = tokio::runtime::Runtime::new()
+            .map_err(|e| format!("failed to create tokio runtime: {e}"))?;
 
-        rt.block_on(async {
-            match tokio::time::timeout(
-                StdDuration::from_secs(WS_TIMEOUT_SECS),
-                query_node_diagnostics(ws_port),
-            )
-            .await
-            {
-                Ok(Ok(diagnostics)) => Some(diagnostics),
-                Ok(Err(_)) | Err(_) => None,
-            }
-        })
+        rt.block_on(query_with_fallback(ws_port, WS_RETRY_ATTEMPTS))
     }
 
     fn get_user_message(&self) -> Result<Option<String>> {
@@ -335,14 +340,11 @@ impl ReportCommand {
                 "not found"
             }
         );
-        println!(
-            "  - Node status: {}",
-            if report.network_status.is_some() {
-                "running"
-            } else {
-                "not running or unreachable"
-            }
-        );
+        match (&report.network_status, &report.network_status_error) {
+            (Some(_), _) => println!("  - Node status: running"),
+            (None, Some(err)) => println!("  - Node status: unreachable ({err})"),
+            (None, None) => println!("  - Node status: not running or unreachable"),
+        }
     }
 
     fn save_local(&self, report: &DiagnosticReport, path: &PathBuf) -> Result<()> {
@@ -654,9 +656,51 @@ fn parse_ws_port_from_config(config: &str) -> Option<u16> {
     None
 }
 
+/// Attempt the diagnostics query against each loopback host in order, with
+/// a bounded number of retries on timeout. Returns the serialized
+/// diagnostics on success, or a concatenated error string describing every
+/// attempt on failure.
+///
+/// Retries are timeout-only: connection-refused errors are reported
+/// immediately since they indicate the node is not running on that host.
+async fn query_with_fallback(port: u16, retry_attempts: u32) -> Result<String, String> {
+    let mut errors: Vec<String> = Vec::new();
+    let total_attempts = retry_attempts.saturating_add(1);
+
+    for host in LOOPBACK_HOSTS {
+        for attempt in 0..total_attempts {
+            match tokio::time::timeout(
+                StdDuration::from_secs(WS_TIMEOUT_SECS),
+                query_node_diagnostics(host, port),
+            )
+            .await
+            {
+                Ok(Ok(diag)) => return Ok(diag),
+                Ok(Err(e)) => {
+                    // Connection-level failure (refused, no route, etc.).
+                    // Don't retry the same host — the error won't change.
+                    let msg = format!("{host}:{port}: {e:#}");
+                    errors.push(msg);
+                    break;
+                }
+                Err(_) => {
+                    let msg = format!(
+                        "{host}:{port}: timed out after {WS_TIMEOUT_SECS}s (attempt {}/{})",
+                        attempt + 1,
+                        total_attempts
+                    );
+                    errors.push(msg);
+                }
+            }
+        }
+    }
+
+    Err(errors.join("; "))
+}
+
 /// Query the node for diagnostics via WebSocket API.
-async fn query_node_diagnostics(port: u16) -> Result<String> {
-    let url = format!("ws://127.0.0.1:{port}/v1/contract/command?encodingProtocol=native");
+async fn query_node_diagnostics(host: &str, port: u16) -> Result<String> {
+    let url = format!("ws://{host}:{port}/v1/contract/command?encodingProtocol=native");
 
     let (stream, _) = connect_async(&url)
         .await
@@ -766,6 +810,7 @@ mod tests {
             },
             config: None,
             network_status: None,
+            network_status_error: None,
             user_message: Some("Test message".to_string()),
         };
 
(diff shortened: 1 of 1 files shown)


sanity:
## Testing Review: PR #3898

### Coverage Summary
- **Direct changes**: partial — new WS fallback logic is covered; serialization covered; wire-compat covered.
- **Downstream impact**: partial — `print_summary` three-branch match and the `collect_network_status` wiring are not tested.
- **Issue failure modes**: 2 of 3 directly exercised.

The "would have caught it before merge" criterion is satisfied for the core regression: the three tests reference symbols (`query_with_fallback`, `network_status_error`) that do not exist on `main`, so they would fail to compile there, and their `assert!(...)` conditions are tight against the old `Ok(...) | Err(_) => None` silent-drop behavior.

### Test Gaps Found

#### Critical (blocks merge)

None. The three regression tests close the issue's primary "silent None" failure mode and the wire-compat invariant.

#### Important (should have)

1. **`print_summary` three-branch match is untested.** Lines 343-347 are the user-visible half of the fix — the entire point of #3897 from a human operator's perspective is seeing "unreachable (timed out after 15s)" in the CLI instead of "not running or unreachable". A trivial unit test that constructs three `DiagnosticReport` values (Some/None, None/Some, None/None) and asserts against captured stdout (or refactors the match into a `summarize_node_status(&report) -> &'static str | String` helper that returns the string) would catch any regression from someone "simplifying" the three-branch match back to two branches. Strong recommendation — this is the exact surface the user cares about.

2. **`collect_network_status` result wiring is untested.** Lines 182-185 contain the `Ok(diag) => (Some(diag), None), Err(e) => (None, Some(e))` match. If a future refactor flips one of the arms (e.g. accidentally sets both or drops the error), no test catches it. The `generate_report` function is hard to unit-test as-is because it takes `&self` on `ReportCommand` and reads from disk, but the five-line match is worth extracting into `fn network_status_tuple(result: Result<String, String>) -> (Option<String>, Option<String>)` with a two-assertion unit test.

3. **Timeout test robustness on slow runners.** The timeout test budgets `15s (WS_TIMEOUT_SECS) + 20s (wrapper)` with only 5s of headroom. On an Ubicloud `ubicloud-standard-8` runner under parallel test load (the main `test` job uses `--workspace` nextest at default parallelism), a 15s tokio timer plus scheduler latency plus a second [::1] connection-refused attempt can plausibly reach 16-18s wall time, and if nextest retries are disabled for any reason you get a flaky failure. Two fixes, either is acceptable:
   - Raise the wrapper to 30s. Gives 15s of headroom without meaningfully increasing CI time on the happy path. Recommended.
   - Use `tokio::time::pause()` + `advance()` so the timeout is deterministic regardless of wall clock. More work but eliminates the flake risk entirely.
   
   Note: nextest `slow-timeout` in `.config/nextest.toml` is 120s, so the test won't be killed — the concern is purely the inner wrapper firing before `WS_TIMEOUT_SECS`.

4. **`retry_attempts > 0` is never exercised.** Both regression tests call `query_with_fallback(port, 0)` — the production path uses `WS_RETRY_ATTEMPTS = 1`. A refactor that breaks the `total_attempts = retry_attempts.saturating_add(1)` loop (off-by-one, early break) would not be caught. Add one test with `retry_attempts=1` and assert the error string contains `attempt 2/2` so the retry counter is pinned.

#### Nice to Have

5. **Error-accumulation format across hosts.** The connection-refused test asserts both hosts appear in the concatenated string, but not the separator or ordering. If someone switches `errors.join("; ")` to `"\n"` or drops one host's error early, wire format downstream (tooling that parses the string) could break silently. A single `err.matches("; ").count() == 1` assertion locks the format.

6. **Malformed config port.** `parse_ws_port_from_config` returns `None` on garbage and the code falls back to `DEFAULT_WS_API_PORT` — no test exercises `ws-api-port = "not-a-number"` or `ws-api-port = 99999999` (overflow). Not in-scope for #3897, but a reasonable reviewer would note the gap.

7. **IPv6-only CI environment.** If Ubicloud ever runs a v4-disabled container, `TcpListener::bind("127.0.0.1:0")` in the timeout test fails with `EADDRNOTAVAIL` and the test panics rather than reporting a meaningful skip. Low probability but cheap to guard with a `.unwrap_or_else(|e| { eprintln!("skipping: {e}"); return; })`. Nice-to-have.

8. **Port 59111 collision.** Hardcoded in `test_query_with_fallback_connection_refused_reports_error`. If a parallel test in the same process happens to bind that port (none do today — grep confirmed), the connection-refused test would unexpectedly succeed. Using `TcpListener::bind(...)` then dropping to get a known-unused ephemeral port is more robust. Low priority — nextest runs each test in its own process per `.config/nextest.toml`, so cross-test collision isn't a real risk.

### CI Considerations

- **Runtime impact**: timeout test adds ~15s wall time to the `test` job. Acceptable. Connection-refused test and serialization test are sub-second.
- **Flake risk**: moderate on the timeout test (see Important #3). Zero on the other two.
- **No simulation-test coverage needed**: this is a CLI binary path, not a routing/topology behavior. Unit tests at the right level.
- **Retry behavior**: nextest's `profile.ci` retries=2 will mask a genuine flake in the timeout test, which makes Important #3 easier to miss until it becomes persistent. Worth fixing deterministically.

[AI-assisted - Claude]

sanity:
## Code-First Review: PR #3898

Read the diff first, then the description. The stated intent and the implementation line up well — no material discrepancies between "what the PR says it does" and "what the code does". Findings below are implementation quality and one UX regression, not intent mismatches.

### My understanding (from code only)

Before reading the description I understood the change as:

1. Raise `WS_TIMEOUT_SECS` from 5 → 15 and retry once on timeout (not on connection-refused).
2. Fall back from IPv4 loopback to IPv6 loopback so that `::` + `IPV6_V6ONLY=1` or v4-only binds don't silently drop the query.
3. Add `network_status_error: Option<String>` to `DiagnosticReport` (serde-default, skip-if-none) so the report records the reason the WS query failed instead of showing a silent `null`.
4. Update `print_summary` to distinguish "running" / "unreachable (reason)" / "not running or unreachable".
5. Add three tests: connection-refused returns an error mentioning both hosts + the port, a stalled handshake surfaces a timeout error, and the new field round-trips through serde / is omitted when `None`.

All of the above matches the PR description. No hidden behavior changes.

### Must-fix

None.

### Should-fix

**Worst-case CLI blocking time grew from 5s to 60s.** With `LOOPBACK_HOSTS` = 2 and `total_attempts = WS_RETRY_ATTEMPTS + 1 = 2`, if both loopback hosts time out on both attempts, `freenet report` blocks for `2 * 2 * 15s = 60s` before producing a report. Previously this was 5s. For a user whose node is hung (the primary use case this PR is trying to serve), a minute of apparently-frozen CLI before anything prints is a real UX regression. Consider:

- First attempt per host at full `WS_TIMEOUT_SECS`, retry attempts at a shorter timeout (e.g., 5s); OR
- Abort remaining attempts once wall-clock total exceeds some cap (e.g., 30s); OR
- Print a progress line like `"querying node on 127.0.0.1:7509 (up to 15s)…"` before each attempt so the CLI doesn't appear hung.

At minimum, the 60s worst case should be called out in a doc comment on `query_with_fallback` so future maintainers don't increase retry counts without realizing the multiplicative effect.

**`LOOPBACK_HOSTS` name is slightly load-bearing and worth clarifying.** `parse_ws_port_from_config` reads only the port — the configured `ws-api-address` is deliberately ignored, and the collector always connects to one of the two loopback hosts. If a user sets `ws-api-address = "10.0.0.5"` (non-loopback), the report will always come back with `network_status_error` even though the node is running. This is the same as main's behavior (the old code hardcoded `127.0.0.1`), but the new doc comment on `LOOPBACK_HOSTS` only justifies the v4/v6 split — it doesn't mention that non-loopback configs are intentionally unsupported. Either:

- Also parse `ws-api-address` from the config and try it first when it's set; OR
- Add a sentence to the `LOOPBACK_HOSTS` doc comment saying "non-loopback `ws-api-address` values are not reached by this collector — users with custom binds should run diagnostics over the same interface".

### Consider

**`test_query_with_fallback_timeout_reports_error` has a 5s margin on CI.** The test wraps `query_with_fallback(port, 0)` in a 20s `tokio::time::timeout`. The inner call needs to wait `WS_TIMEOUT_SECS = 15s` on the IPv4 attempt. On a heavily loaded CI runner, 5s of margin is tight. Two mitigations worth considering:

- Parameterize `WS_TIMEOUT_SECS` via an optional argument to `query_with_fallback` so the test can pass a shorter timeout (2s would be plenty to exercise the timeout code path and would let the test wrapper shrink to 5s).
- If keeping the real timeout, widen the outer wrapper to 30s.

If this ever flakes, the current note in the test (`"test wrapper timed out before WS_TIMEOUT_SECS fired"`) is clear enough that the root cause will be obvious, but I'd rather not ship something we know has a narrow margin.

**`total_attempts` terminology.** `WS_RETRY_ATTEMPTS = 1` with the loop doing `retry_attempts.saturating_add(1)` is correct, but reading the constants in isolation ("1 retry") makes the 60s worst-case non-obvious. A combined constant like `WS_TOTAL_ATTEMPTS_PER_HOST = 2` or a doc comment on `WS_RETRY_ATTEMPTS` that spells out "total attempts per host = 1 + retries = 2, total wall clock per host = 2 * WS_TIMEOUT_SECS = 30s, across both loopback hosts = 60s" would help future maintainers tune safely.

**Retry on timeout may hit a struggling node twice.** If the first attempt actually connected and the node is mid-response when we time out, the retry opens a second WS client. tokio-tungstenite closes the socket when the dropped future's stream is dropped, so this is OK, but on a node that's saturated (the motivating case) the extra connection load is non-zero. Acceptable for diagnostic collection, but worth a comment acknowledging it.

### Wire-format backward compatibility

Asked to verify that `network_status_error` is truly bidirectionally compatible:

- **Newer writer → older reader**: `#[serde(skip_serializing_if = "Option::is_none")]` means an unchanged report (node running, query succeeded) produces byte-identical JSON to main. An "error" report includes an extra top-level key. `serde_json` by default ignores unknown fields, so an older server deserializer will accept it unless it uses `#[serde(deny_unknown_fields)]` on `DiagnosticReport`. I can't verify the server-side struct from this repo — worth confirming with whoever owns `nova.locut.us/api/reports` that it does not use `deny_unknown_fields`. If it does, this PR breaks uploads from new clients whose queries failed.
- **Older writer → newer reader**: `#[serde(default)]` populates `None`, which round-trips correctly. Safe.
- **Binary/bincode encoding**: not used here — reports are JSON over HTTP. The test `test_network_status_error_serialization` covers the serde-json round-trip.

One quick mitigation, if the server schema is out of reach: the server-side deserializer (if anyone controls it) should be auditable for `deny_unknown_fields`. Worth a one-line check before merging.

### Documentation recommendations

- Add a doc comment on `query_with_fallback` stating the worst-case wall-clock (`len(LOOPBACK_HOSTS) * (WS_RETRY_ATTEMPTS + 1) * WS_TIMEOUT_SECS` seconds) so that bumping any of the three constants makes the effect on CLI responsiveness obvious.
- Note on `LOOPBACK_HOSTS` (or on `collect_network_status`) that the configured `ws-api-address` is intentionally not consulted for the host; only the port is parsed.
- Cross-reference #3897 in the `network_status_error` struct field docstring so grep from the failure mode ("why is this None?") lands directly on the fix.

[AI-assisted - Claude]

sanity:
## Big Picture Review

Reviewed against issue #3897 and related diagnostic infrastructure. Scope is well-chosen and tests are load-bearing regressions; no CI-chasing patterns found. A couple of docs-sync items to address and a few follow-ups worth filing.

### Goal Alignment

- **Stated goal**: Surface WS query failures in diagnostic reports instead of silent `None`.
- **Issue #3897 proposed five fixes**: (1) record failure reason, (2) longer timeout + retry, (3) try both loopback families, (4) event-log fallback, (5) surface in `print_summary`.
- **Implemented**: 1, 2, 3, 5. Deferred: 4.
- **Match**: Yes, partially, with justified scope. The PR description is explicit about deferring (4), and the four implemented items close the concrete failure modes observed in report FVQ7X3. Deferring the event-log fallback is reasonable — it's a meaningfully larger piece of work (log parsing, event classification, schema for the fallback snapshot) and the simpler "tell us why" fix is what the three recent `network_status: null` reports actually needed.

### Removed Code Concerns

None. No tests removed, no existing functionality deleted. The old `collect_network_status -> Option<String>` becomes `-> Result<String, String>` and both the `Some` and (new) `None+error` paths are preserved. Three new regression tests added, all of which would fail against `main`.

### Anti-Patterns Detected

None of the usual CI-chasing patterns. Worth noting positively:
- No `#[ignore]`, no `.ok()`-swallowing, no weakened asserts.
- The port-collision concern in `test_query_with_fallback_connection_refused_reports_error` is addressed with a comment acknowledging the failure mode and arguing it's still informative.
- `test_query_with_fallback_timeout_reports_error` wraps the inner timeout in an outer 20s assertion bound, which is the right pattern (CI visibility if `WS_TIMEOUT_SECS` ever changes without the test being updated).

### Stale Docs / Rules

**Blocker — must update before or with merge:**
- `~/.claude/skills/freenet-diagnostic-reports/SKILL.md` lines 56–58 show the report JSON shape with `\"network_status\": null` and no `network_status_error` field. Any agent consulting this skill to triage a report will miss the new field entirely. Needs a one-line addition to the example JSON plus a sentence in the \"Debugging Workflow\" section (probably a new bullet 2a: \"Check `network_status_error` first — if present, the node was running but the report tool couldn't reach it; investigate WHY before blaming the node\").

**Not affected:**
- `.claude/rules/*.md` in freenet-core — nothing references `report.rs`, `network_status`, or the WS query path. No updates needed.
- `AGENTS.md` / `CLAUDE.md` — no references to the diagnostic report tool's internals.
- `crates/core/CLAUDE.md` — no references.
- `docs/debugging/testing-logging-guide.md` — mentions \"diagnostic reports\" but in the context of test event aggregation, unrelated to this tool.

### Server-Side

**No companion PR needed.** The nova report server (`~/code/freenet/freenet-report-server/src/main.rs:90`) stores the client payload as `serde_json::Value`, not a typed struct, so new optional fields pass through unchanged. The `#[serde(default, skip_serializing_if = \"Option::is_none\")]` on `network_status_error` is sufficient for wire compat and `test_network_status_error_serialization` locks it in.

### User-Facing Impact

Summary output change (\"not running or unreachable\" → \"unreachable (reason)\") is net-positive and I can't find any user documentation that references the old phrasing:
- No hits in `docs/`, no hits in any `*.md` across the freenet-core tree for \"not running or unreachable\".
- The external manual at freenet.org/resources/manual/ is outside this repo; if it documents the `freenet service report` output verbatim, that would need an update, but grep of this tree shows nothing.

### Related Silent-None Patterns (follow-ups worth filing)

The issue notes this is the third report in two weeks with `network_status: null`. Similar silent-failure patterns exist nearby that would benefit from the same treatment:

1. **`collect_config` (report.rs:228)** silently returns `None` on any error. If the user has a typo'd config path or a permissions problem, the report shows \"Config: not found\" with no hint why. Not a regression from this PR, but same failure-mode class.
2. **`collect_logs`** silently drops individual log files it can't read (via `read_and_merge_log_files` at line 211) — at least when the log dir itself is missing, the `Result` propagates, but per-file read failures are invisible.
3. **Issue #2919** (self-service diagnostic lookup page, referenced in #3897's footer) will consume `network_status_error` once live — worth a line in that issue's design notes so the UI distinguishes \"node offline\" from \"node stuck\".

Not blocking for this PR, but a pair of follow-up issues (`collect_config` error surfacing + event-log fallback from item (4)) would close the remaining silent-drop surface in `report.rs`.

---

### Verdict by Category

**Blocker:**
- Update `~/.claude/skills/freenet-diagnostic-reports/SKILL.md` to document the new `network_status_error` field in the example JSON and mention it in the debugging workflow. Otherwise agents triaging reports will continue to see the old-shape example and miss the new signal. (This is a `~/.claude/` repo, not freenet-core; can be a separate commit there — but shouldn't be forgotten.)

**Should address:**
- (None within the scope of this PR. The PR is tight.)

**Follow-up OK:**
- File an issue for item (4) from #3897 (event-log fallback snapshot) so it isn't lost. The PR description says \"left for a follow-up if this fix proves insufficient\" — worth capturing as a real ticket rather than a commit-message reference.
- File an issue for `collect_config` surfacing its failure reason (same silent-None class as #3897).
- Update issue #2919 notes to reference the new `network_status_error` field.

Overall: good, focused fix. The test-first structure (both failure modes asserted against `main` behavior) is exactly what `.claude/rules/testing.md` asks for on `fix:` PRs.

[AI-assisted - Claude]

sanity:
## Skeptical Review: PR #3898

Adversarial review as requested. I traced the retry/fallback state machine, checked wire compat, and looked for silent-failure escape hatches. Summary first, then detail.

---

### Blocking

None. I don't see a correctness bug that should prevent merge. The change is additive on the wire (`skip_serializing_if = Option::is_none`) and the retry logic terminates.

---

### Must address

**1. Worst-case blocking time is 60s on a cold/hostile node.** With `WS_TIMEOUT_SECS=15`, `WS_RETRY_ATTEMPTS=1`, and 2 hosts, the bound is `15 × 2 × 2 = 60s` before the user sees anything. This is a CLI command the user ran to report a problem — sitting on a black terminal for a minute is poor UX and will be interpreted as "the tool hung too." Please either (a) print a progress line like `"[attempt 1/4] querying 127.0.0.1:7509…"` before each `tokio::time::timeout`, or (b) pass an overall hard ceiling to `query_with_fallback` separate from the per-attempt timeout. Today there is no visible progress between `"Collecting diagnostic info..."` and eventual unreachable-message output.

**2. Test `test_query_with_fallback_timeout_reports_error` burns ~15s of real wall time.** `#[tokio::test]` uses unpaused time and the test relies on the actual `WS_TIMEOUT_SECS` firing. Two problems:
   - Every CI run pays 15s for this one test.
   - If someone later bumps `WS_TIMEOUT_SECS` above 20s without touching the test, the outer `tokio::time::timeout(StdDuration::from_secs(20), …)` will fire first and the assertion message "test wrapper timed out before WS_TIMEOUT_SECS fired" triggers. That's a coupled constant.
   
   Fix: make `WS_TIMEOUT_SECS` a parameter to `query_with_fallback` (or expose a `#[cfg(test)]` override), and call it with e.g. 100ms in the test. Then use `#[tokio::test(start_paused = true)]` + `tokio::time::advance`. Removes the 15s CI cost and removes the hidden coupling to `WS_TIMEOUT_SECS`.

**3. `WS_RETRY_ATTEMPTS=1` has no test coverage.** Both new tests use `retry_attempts=0`. The retry loop (`for attempt in 0..total_attempts`) is never exercised on the timeout path, so a refactor that breaks the retry (e.g. accidental `break` on timeout) would pass tests. Add a test that asserts the error string contains `attempt 1/2` and `attempt 2/2` when a listener accepts but never handshakes.

**4. Retry loop does not re-evaluate success on retry.** The loop correctly re-attempts on timeout, but there is no jitter between the first timeout and the retry. If the node is wedged waiting on the event loop, hammering it with a second request 0 ms after the first times out is unlikely to help and may arrive while the handler is still processing the cancelled first request. The [repo's retry rule](.claude/rules/code-style.md) explicitly requires ±20% jitter. Even 200-500ms of sleep between retries would be more useful than retrying instantly. Consider `tokio::time::sleep(Duration::from_millis(500))` between retry attempts (single-host loop only).

**5. `parse_ws_port_from_config` reads port but not address.** If a user has `ws-api-address = "192.168.1.50"` (bound to LAN, not loopback), both `127.0.0.1` and `[::1]` attempts will fail with refused and the report will falsely say "unreachable" even though the node is running. This is preexisting behavior (main also hardcoded `127.0.0.1`), but the PR explicitly frames "try both loopback hosts" as the fix for address mismatches — and misses the most common mismatch. Either (a) read the configured address and add it to the fallback list ahead of the loopbacks, or (b) add a code comment noting the limitation and file a follow-up issue.

---

### Worth considering

**6. Localhost shadowing (your concern 6 in the prompt).** If the user has a second WebSocket server on `[::1]:7509` (different freenet instance in a container, unrelated dev service, etc.), the report tool will successfully connect and upload diagnostics from the wrong node with `network_status = Some(…)`. The user sees "Node status: running" and nothing indicates the report came from the wrong binding. Mitigations:
   - Include the host that succeeded in the report (new field `network_status_source: Option<String>` or embed `"{host}:{port}"` prefix in the JSON).
   - Cross-check the `node_info.pid` (or equivalent) against the actual freenet process the user is inspecting, if the report tool can discover it.
   
   Low likelihood in practice, but since the whole PR is about "which node answered us," this is a gap in that same framing.

**7. `connect_async` error classification (your concern 7).** You correctly identified that all non-timeout errors short-circuit to "next host, no retry." But tungstenite's error taxonomy is coarse: `ConnectionClosed`, `AlreadyClosed`, `Io(io::Error)`, `Tls(…)`, `Protocol(…)`, `Http(…)`, `Url(…)`. The branch `Ok(Err(e))` lumps `io::Error(ConnectionRefused)` in with `io::Error(HostUnreachable)`, `io::Error(NetworkUnreachable)`, and DNS resolution errors. For IP literals (`127.0.0.1`, `[::1]`) DNS isn't involved, but `HostUnreachable` on a broken IPv6 stack is *not* the same as "node not running" — it's worth one retry. Since the strings `127.0.0.1` and `[::1]` are literals, DNS-type errors can't actually occur here, so in practice this is OK. But the comment "`connection refused, no route, etc.`" conflates two different cases; consider matching on the inner `io::ErrorKind` for refused-only and retrying on `HostUnreachable` / `NetworkUnreachable`.

**8. Error string length is unbounded.** `format!("{host}:{port}: {e:#}")` with anyhow's alternate-format flag emits the full error chain. tungstenite handshake errors can be long (HTTP status line, headers, protocol details). With up to 4 attempts across 2 hosts, the joined error string could be multi-KB and will land verbatim in the uploaded report JSON. Add a length cap (e.g. 512 chars per attempt) or strip to the root cause.

**9. Information disclosure in `network_status_error` (your concern 3).** I audited what `{e:#}` from tungstenite can expose:
   - URL including `ws://` scheme, host, port, path — all non-sensitive; the port is already in the field name.
   - OS error strings from `io::Error` — can include the socket address. On some platforms the OS error also includes errno descriptions; no file paths observed.
   - TLS errors (not applicable here; always `ws://`, not `wss://`).
   
   **Conclusion**: low risk of leaking hostnames/paths/usernames. The report already uploads `hostname`, config file contents, and 2 MB of logs to `nova.locut.us`, so even if a transient file path leaked via an I/O error it wouldn't be a new disclosure class. I don't consider this blocking, but adding a comment in the PR or the code that the error string is user-readable and will be uploaded would help future reviewers.

**10. Server-side `network_status_error` handling (your concern 5, wire compat).** Confirmed the field is forward-compatible at serde level:
   - `#[serde(default, skip_serializing_if = "Option::is_none")]` means old clients (that don't set the field) produce JSON without the key, which old AND new servers deserialize fine.
   - New clients setting `Some(err)` add the key; old servers that do **not** know the field will either ignore it (serde_json with `#[serde(deny_unknown_fields)]` NOT set — but we cannot verify the server's behavior from this PR). If the server uses `deny_unknown_fields`, the upload will fail. Worth a note in the PR that you've verified the server accepts unknown fields, or a pointer to the server repo where you confirmed this.
   - The new `test_network_status_error_serialization` test confirms the client side. It does **not** confirm the server side.

**11. Retry inner loop inside outer host loop uses `break`, not `continue`.** At line 683, `break` exits the inner retry loop and moves to the next host. That's the intended behavior for "don't retry connection-refused on the same host," so this is correct — but it's subtle. A comment like `// break inner retry loop → fall through to next host` would help future readers avoid "fixing" it.

**12. `errors.join("; ")` on success path.** On success, `errors` may contain entries from a prior host/attempt (e.g., `[::1]` refused, `127.0.0.1` succeeded after retry). Those errors are silently discarded. For diagnostics purposes it might be useful to report "connected via 127.0.0.1 on attempt 2, [::1] was refused" so the maintainer knows the node's IPv6 binding is broken. Today the report only says "Node status: running" with no hint of the degraded IPv6 state.

**13. Hard-coded `LOOPBACK_HOSTS: &[&str] = &["127.0.0.1", "[::1]"]`.** The order (v4 first) matches the historical behavior and is a reasonable default. If the local config explicitly binds to v6-only (`ws-api-address = "::1"`), v4 will connect-refuse first, wasting one full attempt — but total cost is the `connect()` syscall returning ECONNREFUSED, so <1ms. No action needed; calling it out for the record.

---

### Questions for Author

- **Server validation**: has `nova.locut.us/api/reports` been verified to accept a field it doesn't know about? If the server uses `deny_unknown_fields` or a strict JSON schema, old deployed servers will reject new-format reports (silent upload failure). Can you confirm with a staging request, or confirm the server code is permissive?
- **Retry without jitter**: is the retry deliberately instant, or was jitter just not considered? Repo rule says retries must jitter.
- **Why retry_attempts=1 and not retry_attempts=2**? One retry gives two attempts per host; on a very loaded Raspberry Pi (FVQ7X3's actual scenario) one extra try may not be enough. Is this a deliberate "keep max wait bounded" choice?

---

### What I verified (not bugs)

- **Runtime conflict (your concern 4)**: `main()` in `crates/core/src/bin/freenet.rs:690` is a plain `fn main()`, not `#[tokio::main]`. `ServiceCommand::Report` is dispatched from a sync context, so creating a fresh `tokio::runtime::Runtime` inside `collect_network_status` is safe.
- **DNS blocking (your concern 7)**: both loopback strings are IP literals (`127.0.0.1` and `[::1]`), so `connect_async` does not invoke DNS resolution. No blocking syscall risk beyond the syscall-level connect().
- **URL format for IPv6**: `format!("ws://{host}:{port}/…")` with `host = "[::1]"` produces `ws://[::1]:7509/…` which is valid RFC 3986 and parses correctly in tungstenite.
- **Serde wire compat (client side)**: `#[serde(default, skip_serializing_if = "Option::is_none")]` is correct. JSON from new clients is identical to old when the error is `None`.
- **Default WS bind**: confirmed `default_listening_address()` returns `IpAddr::V6(Ipv6Addr::UNSPECIFIED)` which on Linux dual-stacks. The PR's fallback-to-v6 logic correctly addresses the IPV6_V6ONLY scenario called out in the issue.
- **`errors` Vec unbounded growth**: total attempts is `hosts × (retries+1) = 2 × 2 = 4`, so the Vec is O(1) bounded by design. No DoS surface.
- **`tokio::time::timeout` + `connect_async` cancellation safety**: dropping the future on timeout aborts the connect cleanly; no FD leak in tungstenite 0.27. (Verified by reading the connect_async source path.)

[AI-assisted - Claude]

sanity:
## Review response — commit 58a37f41

Thanks to all four reviewers. Addressed the following in [58a37f41](https://github.com/freenet/freenet-core/pull/3898/commits/58a37f41):

### Fixed

**1. Silent 60s wait (code-first + skeptical)**
Added progress output around the WS query:
```
  Querying local node diagnostics (up to 60s)... ok
  Querying local node diagnostics (up to 60s)... unreachable (127.0.0.1:7509: timed out after 15s ...)
```
The user sees the expected worst case up front and concrete progress afterward.

**2. `WS_RETRY_ATTEMPTS=1` path uncovered (testing + skeptical)**
Added `test_query_with_fallback_retries_on_timeout`: uses an atomic accept counter and asserts the production retry path produces exactly 2 v4 accepts (initial + 1 retry). Error string is checked for both `attempt 1/2` and `attempt 2/2`.

**3. Timeout test margin / CI wall-clock cost (testing + code-first + skeptical)**
Parameterized `query_with_fallback` with a `per_attempt_timeout` argument. Tests pass `TEST_TIMEOUT = 200ms`; production passes `Duration::from_secs(WS_TIMEOUT_SECS)`. Test suite's wall-clock time for this module dropped from ~15s to under 0.5s.

**4. SKILL.md update (big-picture)**
Updated `~/.claude/skills/freenet-diagnostic-reports/SKILL.md` with a table explaining the three combinations of `network_status` / `network_status_error` so future triage agents know which field to look at first. Committed separately in the `~/.claude` repo ([d1f0aa2](https://github.com/sanity/claude-config/commit/d1f0aa2)).

### Declined with rationale

**Jitter on retries (skeptical)**
The code-style rule requires jitter on retry/backoff loops to prevent thundering herd. This is a one-shot CLI invocation with no inter-attempt delay — the retry fires the next `connect_async` as soon as the previous one times out. There is no "interval" to jitter because there is no wait between attempts. Not applicable.

**`parse_ws_port_from_config` ignores `ws-api-address` (code-first + skeptical)**
Real limitation but orthogonal to #3897. Pre-existing on `main` — a node listening on a non-loopback address is already misreported. Worth a follow-up issue if it comes up in practice.

**`collect_config` also silent-Nones (big-picture)**
Same failure mode in the same file. Scoped out to keep this PR focused on the reported failure. Will file a follow-up.

**Server-side / `deny_unknown_fields` risk (code-first + skeptical)**
Big-picture reviewer verified the receiver at `freenet-report-server/src/main.rs:90` stores the payload as `serde_json::Value`, so new optional fields pass through unchanged. No server-side change needed.

**Wrong-node diagnostics via `[::1]` shadowing (skeptical)**
A valid edge case, but only if (a) the user is running a non-default dual-stack config and (b) there's a *different* freenet node on the same loopback port in the other address family. This is a rare real-world configuration and the diagnostics would still come from a real local node, which is better than a silent null. Worth documenting if it ever surfaces; not worth a pre-emptive fix.

### Not implemented (out of scope)

Items from issue #3897 deferred to follow-ups:
- Event-log fallback snapshot (issue item 4)
- `print_summary` direct-output test (trivial branches; covered implicitly by serialization + wiring tests)

[AI-assisted - Claude]

PR #3905, opened 2026-04-18
fix(uninstall): scan multiple install dirs and collapse empty parents

### LINKED ISSUE #3904: bug(uninstall): Windows install leaves config folder and %LOCALAPPDATA%\Freenet\bin behind
## Summary

Per Matrix feedback from @Ivvvor: running `freenet uninstall` on Windows completes, but leaves several directories behind. In particular, the config folder is not removed, and the installer-created binary directory at `%LOCALAPPDATA%\Freenet\bin` (from `install.ps1`) is not part of the manual install-location detection either.

## Observed vs. expected

After running `freenet uninstall --purge` on Windows, users report:

- `%LOCALAPPDATA%\Freenet\bin\` — binaries still present (installer install location from `scripts/install.ps1:60`, separate from the `current_exe parent` detection in `uninstall.rs:115`).
- `%APPDATA%\The Freenet Project Inc\Freenet\` — config folder not fully removed.
- `%LOCALAPPDATA%\freenet\logs\` — log directory from `tracing::get_log_dir` (tracing.rs:3999) not removed on Windows (the `get_log_dir` branch is Windows-aware but `purge_data_dirs` only calls it for non-system mode and may not cover this location).

## Why this happens

`purge_data_dirs` in `crates/core/src/bin/commands/service.rs:1175-1244` uses `ProjectDirs::from("", "The Freenet Project Inc", "Freenet")`. On Windows that resolves to:

- `data_local_dir` → `%LOCALAPPDATA%\The Freenet Project Inc\Freenet\data`
- `data_dir` → `%APPDATA%\The Freenet Project Inc\Freenet\data`
- `config_dir` → `%APPDATA%\The Freenet Project Inc\Freenet\config`
- `cache_dir` → `%LOCALAPPDATA%\The Freenet Project Inc\Freenet\cache`

The code removes each of those leaf subdirectories, but does not remove the `Freenet` parent folder under either `%APPDATA%` or `%LOCALAPPDATA%`, so the folder skeleton remains. Separately, `install.ps1` installs binaries to `%LOCALAPPDATA%\Freenet\bin`, which is a completely different path from `current_exe().parent()` expects — if Windows PATH resolution has picked up the binary from a different location, `get_install_dir` in `uninstall.rs` won't find the installer's copy.

## Proposed fix

1. After purging subdirectories on Windows, also remove the parent `Freenet` folder under `%APPDATA%\The Freenet Project Inc\` and `%LOCALAPPDATA%\The Freenet Project Inc\` if empty.
2. When detecting install directory, additionally check `%LOCALAPPDATA%\Freenet\bin\freenet.exe` (the install.ps1 default) in addition to `current_exe().parent()`.
3. Add the `%LOCALAPPDATA%\freenet\logs` directory to the purge targets on Windows (currently relies on `get_log_dir()` which goes through a separate code path).
4. Check `HKCU:\Software\Microsoft\Windows\CurrentVersion\Run` for any leftover autostart entry.
5. Regression test: a Windows-targeted unit test that seeds the real directory layout, runs the uninstall, and asserts all paths are gone.

## Workaround (documented)

Added to the Quick Start and FAQ on freenet.org in https://github.com/freenet/web/pull/35: after running `freenet uninstall`, manually remove the three folders with PowerShell and check the Run registry key.

## Related

- freenet/web#35 — documentation PR with the user-facing workaround
- #3739 — Windows data path moved from Roaming to Local (context for why both `data_dir` and `data_local_dir` exist)
- #3717 — earlier Windows reinstall bug from @Ivvvor (same reporter)
- #2961 — original feature issue that added `freenet uninstall`

[AI-assisted - Claude]

sanity:
## Problem

Two user-visible bugs in `freenet uninstall`:

**1. Stale binaries from mixed install methods.** The command detects the install directory by looking at `current_exe().parent()` only. A user who ran the `curl | sh` installer (places binaries in `~/.local/bin`) and later `cargo install freenet` (places a binary in `~/.cargo/bin`) ends up with the uninstall removing whichever was on PATH first, leaving the other behind to shadow future reinstalls. Same failure mode on Windows: the PowerShell installer puts binaries in `%LOCALAPPDATA%\Freenet\bin`, which PATH detection misses when a different `freenet.exe` is active.

**2. Empty parent folders left behind on Windows** (reported on Matrix by @Ivvvor, filed as #3904). `purge_data_dirs` removes the leaf `data`, `config`, and `cache` subfolders but never touches their common parent, so `%APPDATA%\The Freenet Project Inc\Freenet\` — and analogously `%LOCALAPPDATA%\freenet\` once its `logs` child is gone — stays on disk long after uninstall claims completion.

## Approach

**Install-dir detection is now multi-site.** `get_install_dir` is replaced with `get_install_dirs` (plural) returning the union of the running-exe parent, `~/.local/bin`, `~/.cargo/bin`, and `%LOCALAPPDATA%\Freenet\bin`. `FREENET_INSTALL_DIR` and the explicit `override_dir` arg still short-circuit to a single location. Missing dirs are skipped silently; present dirs have their binaries removed, and on Windows the enclosing `%LOCALAPPDATA%\Freenet\` collapses once `bin\` is gone.

**Parent-folder cleanup is explicit.** `purge_data_dirs` now builds a `DataLeaves` value and delegates to `purge_leaves_and_collapse`, which removes every leaf, collects their parents, then visits the parents deepest-first to collapse any that became empty. A `remove_dir_if_empty` helper enforces safety: it removes a directory only if it contains nothing, so a non-empty sibling app's data under `The Freenet Project Inc\` is preserved.

The overall behavior on a clean Linux install is unchanged; the extra work only kicks in when multiple install sources coexist or when a formerly-Freenet parent is now empty.

## Testing

25 new unit / integration tests covering:

- **`get_install_dirs`** precedence (override > env var > defaults), empty env-var fallthrough, override beating env var, exe-parent/cargo-bin dedup, missing home dir.
- **`remove_dir_if_empty`** safety: empty-dir removal, single-file preservation, subdirectory-tree preservation, missing path, regular file, full children-then-collapse scenario.
- **`purge_leaves_and_collapse` wiring** (the actual #3904 regression site): full Windows-like layout cleanup, sibling-app-grandparent preservation, log-parent gate, missing optional leaves.
- **`collapse_windows_bin_tree`**: extracted so Linux CI can exercise it; tests for empty tree, non-empty sibling preservation, missing tree.

The wiring-level tests mean reverting any individual `push_parent(...)` call at a future date would fail a test, not silently regress #3904.

Tests explicitly chose the unit level because the behavior is pure filesystem manipulation with no network or timing surface. A Windows matrix integration test seeding the actual `%LOCALAPPDATA%` layout would be the next level of coverage; filed as a follow-up but out of scope for this PR.

Full `cargo test -p freenet --bin freenet` (74 tests) passes; `cargo clippy --all-targets -- -D warnings` clean.

## Fixes

Closes #3904

## Out of scope

- **HKCU `Run` registry key** was called out in #3904 as a potential leftover. Skipped here because `install.ps1` does not write a Run entry — it uses `freenet service install` (Task Scheduler) instead, which `stop_and_remove_service` already handles.
- **Auto-update detection of stale installs.** Separate concern.

## Related

- freenet/web#35 merged documentation for the same user-visible problem.
- A follow-up freenet-core PR will add a standalone `uninstall.sh` companion script and update `install.sh`'s final message to steer users away from `sudo freenet uninstall`.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -1163,12 +1163,11 @@ fn remove_if_exists(label: &str, path: &Path) -> Result<()> {
     Ok(())
 }
 
-/// Re-export for callers outside this module (e.g. `uninstall::run`) that also
-/// need to collapse empty parent folders after their own cleanup passes.
-/// Currently only used from the Windows-only branch in `uninstall::run` that
-/// cleans up `%LOCALAPPDATA%\Freenet\bin`; gate the export accordingly so
-/// non-Windows builds don't flag it as dead code.
-#[cfg(target_os = "windows")]
+/// Re-export for callers outside this module (e.g. `uninstall::run`) that
+/// also need to collapse empty parent folders after their own cleanup
+/// passes. Runtime use is Windows-only (see `uninstall::collapse_windows_bin_tree`)
+/// but the function is exposed on every platform so that tests for the
+/// caller-side helper can run under Linux CI too.
 pub(super) fn remove_dir_if_empty_pub(path: &Path) {
     remove_dir_if_empty(path)
 }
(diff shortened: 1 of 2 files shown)


sanity:
## Review responses (code-first, testing, skeptical, big-picture)

Addressed in commit 2fdb456:

### MUST-FIX

- **Env-var race in tests** (code-first + testing + skeptical): `get_install_dirs` now takes a `&RuntimeEnv` value; tests pass explicit inputs instead of mutating process env. The `unsafe { set_var }` blocks and the untrue "tests run single-threaded" comment are gone.
- **Call-site wiring untested for #3904** (testing): extracted `purge_leaves_and_collapse(&DataLeaves)` from `purge_data_dirs`. Four new integration tests exercise the full wiring — reverting any individual `push_parent(...)` call now fails a test.
- **Parent-ordering footgun** (skeptical #1): parents are now sorted → deduped → iterated in reverse, so deepest paths are processed first. No longer depends on nesting depth being exactly 1.

### SHOULD-FIX

- **Log-parent collapse** (code-first): was "empty by luck" on Linux/macOS. Now gated by an explicit `collapse_log_parent: bool` field on `DataLeaves`, only set to `true` for the Windows path. New test `test_purge_leaves_collapses_log_parent_only_when_flagged` pins the contract.
- **Windows-only collapse coverage** (testing #5): extracted `collapse_windows_bin_tree(local_app_data: &Path)` so Linux CI can exercise it too. Three new tests for empty tree, sibling-preservation, and missing tree.
- **Two-level Windows layout** (testing #3): new `test_purge_leaves_and_collapse_cleans_windows_layout` and `test_purge_leaves_preserves_sibling_app_grandparent` exercise the real `The Freenet Project Inc\Freenet\{data,config,cache}` structure with a sibling `OtherApp\` grandparent survivor.
- **Subdirectory content in non-empty test** (testing #4): new `test_remove_dir_if_empty_preserves_dir_with_subtree` uses a nested directory tree.

### NICE-TO-HAVE

- **Empty `FREENET_INSTALL_DIR`** (testing #6): now falls through to defaults instead of short-circuiting to an empty path. Test: `test_get_install_dirs_empty_env_var_falls_through`.
- **Partial-install assertion tightened** (testing #7): `test_remove_binaries_handles_partial_install` now also asserts the absent binary stays absent afterward.

### Deferred / not applicable

- **Unix binary self-deletion** (skeptical #5): pre-existing Unix behavior (`freenet` can `remove_file` itself while running because the inode survives). Not a regression of this PR.
- **exe-parent needing sudo for system-wide installs** (skeptical #6): pre-existing. Users who installed into `/usr/local/bin` already needed `sudo` to run `service install`; the binary-removal step requiring the same is consistent and not new.
- **TOCTOU in `remove_dir_if_empty`** (skeptical #2): `rmdir` kernel-enforces emptiness, so no data loss is possible. The misleading `Note:` message on the racy `ENOTEMPTY` case is a UX nit, not a correctness issue.
- **Symlink + `is_dir()`** (skeptical #3): `is_dir()` follows symlinks but `remove_dir` on the link errors safely. No data loss.
- **HKCU Run key cleanup** (big-picture / #3904 bullet 4): not addressed because `install.ps1` never writes one — it uses `freenet service install` (Task Scheduler) instead. The #3904 bullet was speculative. Noted in the PR description.

### Follow-ups (filed / planned)

- Docs PR: once this lands in a release, update the "known gap" language in freenet/web's quickstart to remove the Windows manual-fallback that will no longer apply.
- Potential `.claude/rules/bug-prevention-patterns.md` entry on "collapse empty parents after leaf removal, but only parents you own."

Test count: 63 → 74 binary tests, all passing. `cargo clippy --all-targets -- -D warnings` clean.

[AI-assisted - Claude]

sanity:
## Codex review response

Codex caught a real data-loss scenario I had missed:

> **[P2] Avoid collapsing shared XDG parents** — on Linux/macOS the `ProjectDirs` leaves (`~/.local/share/Freenet`, `~/.config/Freenet`, `~/.cache/Freenet`, etc.) have shared XDG hierarchies as parents, not Freenet-owned ones. The previous patch relied on `remove_dir_if_empty` refusing to wipe a non-empty parent — safe in practice because `~/.local/share/` normally has other apps' data — but on a fresh account where Freenet is the only inhabitant, the shared root would have been deleted.

Fixed in 126ba25:

- Replaced the per-leaf `collapse_log_parent` flag with a single `collapse_parents: bool` on `DataLeaves`. True only on Windows, where every leaf's immediate parent is Freenet-owned; false everywhere else.
- The collapse list now only collects parents when the flag is set, so Linux/macOS purges never push XDG roots. Safety is now a type-level guarantee rather than a runtime happenstance.
- New regression test `test_purge_leaves_never_collapses_shared_parents_on_unix` seeds each leaf (data, config, cache, log) under its own dedicated "shared root" tempdir and asserts every shared root survives after purge. Reverting the gate would fail this test.

Full test suite (75 tests) passes. Clippy clean.

[AI-assisted - Claude]

PR #3906, opened 2026-04-18
feat(install): add uninstall.sh companion + sudo warning in installer

sanity:
## Problem

The Matrix user "Alin" who sparked this defense-in-depth effort hit the uninstall flow and gave up frustrated because `sudo freenet uninstall` silently did nothing — the installer drops binaries in `~/.local/bin`, which is not on sudo's default PATH, so the command failed with "command not found" under sudo and left the install in place. Alin's experience isn't unique: any user whose `freenet` binary goes missing, gets corrupted, ends up on a different host's PATH, or was installed to a non-standard location has no curl-based recovery path today. The docs on freenet.org (freenet/web#35) and the `freenet uninstall` robustness fixes in #3905 addressed the core tool; this PR closes the remaining two gaps.

## Approach

1. **Standalone `scripts/uninstall.sh`**, designed to work independently of the `freenet` binary. Mirrors the semantics of `freenet uninstall`: stops and removes the user systemd unit / launchd plist, deletes `freenet`, `fdev`, and the macOS wrapper from every known install location (`~/.local/bin`, `~/.cargo/bin`), and — on confirmation or with `--purge` — removes data, config, cache, and logs. Platform-aware data paths for Linux (XDG: `~/.local/share/Freenet`, etc.) and macOS (bundle-ID layout: `~/Library/Application Support/The-Freenet-Project-Inc.Freenet`). Windows users are redirected to `freenet.exe uninstall`.

   Refuses to run under `sudo` unless `FREENET_ALLOW_ROOT=1` — the install lives in the invoking user's home, and `sudo curl | sudo sh` would operate on root's home instead, which is almost never what the user wants.

2. **`scripts/install.sh` final-message update.** The closing "To uninstall Freenet completely: `freenet uninstall`" line now lists both the preferred command and the curl-based fallback, and explicitly warns against prefixing either with `sudo` — the specific misstep that caused Alin's frustration.

## Testing

`scripts/test-uninstall-sh.sh` exercises uninstall.sh against tempdir `$HOME` fixtures with 8 scenarios, each asserting both what MUST be removed and what MUST be preserved:

1. **`--purge` removes everything but preserves shared parents.** Plants a sibling `other-app.txt` in `~/.local/share/` before purge and verifies it survives — this is the same safety invariant that codex caught in #3905 (Freenet must never wipe shared XDG roots).
2. **`--keep-data` removes binaries only.**
3. **Noop on a clean system** prints "does not appear to be installed" and exits 0.
4. **Conflicting flags** (`--purge` + `--keep-data`) rejected with a clear message.
5. **Unknown flags** rejected.
6. **`--help`** prints usage and exits 0.
7. **Root guard** rejects invocation when `id -u == 0` (simulated via a PATH-shimmed fake `id` command).
8. **`FREENET_ALLOW_ROOT=1`** overrides the root guard.

All 8 pass under both `sh` (dash) and bash; POSIX syntax confirmed via `sh -n`.

Ran end-to-end against a fixture directory to confirm the platform-specific purge paths match what's actually on disk (vs. the incorrect macOS paths that code-first reviewer caught in freenet/web#35).

## Out of scope

- The freenet/web side (serving the new script at https://freenet.org/uninstall.sh and syncing the updated install.sh) lands in a companion freenet/web PR.
- Adding `scripts/test-uninstall-sh.sh` to CI is a follow-up — the script is self-contained and runs via the command above locally.

## Related

- freenet/web#35 — user-facing documentation (merged)
- #3905 — `freenet uninstall` robustness and Windows parent-folder cleanup
- #3904 — Windows config-folder-left-behind bug

[AI-assisted - Claude]


sanity:
## Review responses (code-first, skeptical, codex)

Addressed all findings in commit 1b7c885:

### P1 — Linux paths were wrong (codex)

uninstall.sh used UPPERCASE `~/.local/share/Freenet` etc., but `directories` crate on Linux lowercases the application name via `trim_and_lowercase_then_replace_spaces`, so the actual on-disk paths are all lowercase. The previous docs were wrong too — fixed in a separate freenet/web PR. The core binary's `home_override` branch in `service.rs` has the same bug; filed as freenet/freenet-core#3907 to be fixed after #3905 merges.

### P2 — `FREENET_INSTALL_DIR` ignored (codex)

Now prepended to the scan list when set. New test case 2b.

### HIGH — `rm -rf "${HOME}/..."` with unset/dangerous HOME (skeptical)

Added `case "${HOME:-}" in ""|"/"|"/root"|"/root/"|"/var/root"|"/var/root/")` safety guard that refuses before the first `rm`. Prevents `HOME=""` turning paths into absolute-from-root, and prevents `HOME=/Library/...` reaching into real macOS system dirs. Four new test cases cover the rejected values.

### HIGH — `$(id -u)` spoofable via PATH (skeptical)

Replaced with `id_real_uid()` that prefers `/usr/bin/id` / `/bin/id` by absolute path, with a `command -p id` fallback. Test case verifies a PATH-shimmed `id` doesn't change script behavior.

### MUST-FIX — test harness double-reported PASS/FAIL (code-first)

Added per-case `case_fails` counter and a `finish_case` helper that only emits `PASS` if both the script exited clean AND every assertion passed. Verified by injecting a no-op uninstall; now only FAIL is printed, not FAIL followed by PASS.

### NIT — inaccurate doc comment (code-first)

"The current exe's dir" claim removed. Correctly documents the three locations: `~/.local/bin`, `~/.cargo/bin`, and `FREENET_INSTALL_DIR`.

### NIT — em-dashes (global rule)

Replaced with plain hyphens throughout.

### Deferred / not in this PR

- **service.rs home_override Linux paths bug** — same lowercase issue as the uninstall.sh codex finding, but in the core binary. Tracked as freenet/freenet-core#3907. Fixing here would require rebasing onto #3905 (still merge-queued).
- **`curl | sh --help` doesn't work** (flags go to `sh`, not the script) — users need `curl ... | sh -s -- --help`. The install.sh closing message doesn't expose `--help` so this is probably fine to leave to manual discovery.

## Test count

8 → 13 cases. Every case asserts both what MUST be removed and what MUST be preserved. All passing under both `sh` (dash) and `bash`.

[AI-assisted - Claude]

Issue #3034, opened 2026-02-16
Harden WebSocket origin validation against DNS rebinding attacks


sanity:
## Problem

The WebSocket same-origin check (introduced in PR #3033 for non-localhost gateways) compares the `Origin` header against the `Host` header. While this blocks standard cross-site WebSocket hijacking, both headers are attacker-controlled in a DNS rebinding scenario.

**Attack scenario:**
1. Victim visits `http://evil.com:7509` (attacker's server)
2. Attacker's page loads JavaScript, then the DNS TTL for `evil.com` expires
3. Attacker re-resolves `evil.com` to the Freenet gateway's IP
4. JavaScript opens a WebSocket to `ws://evil.com:7509/v1/contract/command`
5. Browser sends `Host: evil.com:7509` and `Origin: http://evil.com:7509`
6. `is_same_origin` compares them and returns `true`

**Mitigating factors:**
- The auth token system provides secondary defense (tokens are per-contract and session-scoped)
- This is strictly better than the previous state which had NO origin check for non-localhost gateways
- DNS rebinding requires the attacker to know the gateway IP and run their server on the same port

## Proposed Solution

Validate the `Host` header against the server's actual bound address or a configured hostname allowlist, rather than just comparing it against the `Origin` header. Options:

1. **Configured allowlist**: Add `--expected-hostname` CLI flag so operators can specify their hostname
2. **Bound address comparison**: Compare Host against the server's actual listening address/resolved hostname
3. **Random secret in URL path**: Include a per-session secret in the WebSocket URL that the attacker can't predict

## Context

Found during security review of PR #3033 (fix remote cookie domain / WebSocket origin check).

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-03-12T22:26:46.177822407+00:00
Issue hash: `0686b314`

## Value Assessment

The IV in the WebSocket upgrade path is still the one described in the issue: `is_same_origin` simply compares the `Origin`-derived hostname against the incoming `Host` header, and the surrounding middleware accepts any origin that either matches or is considered localhost (`crates/core/src/client_events/websocket.rs:146-189`). Those headers come directly from the browser, so in a DNS rebinding attack both are controlled by the adversary once their domain resolves to the Freenet gateway’s IP. There is no follow-up validation of the actual socket address, no configured hostname allowlist, and no URL-based secret, so the fundamental weakness the issue calls out still exists.

[...]

PR #3909, opened 2026-04-18
fix(server): accept WS Host headers with IPs in allowed-source-cidrs

sanity:
## Problem

`--allowed-source-cidrs` (added in v0.2.47, #3875) lets a user opt in
to reaching the local API from a trusted non-RFC1918 range — the
Matrix-driven example being a Tailscale tailnet (`100.64.0.0/10`).
The flag extends `private_network_filter`, the HTTP source-IP gate.

It does **not** extend the second gate: the WebSocket `Host`-header
allowlist in `connection_info` (added for DNS-rebinding protection in
#3034). That allowlist is built from `localhost`, the machine hostname,
the bound IP, and the operator's `allowed-host` config — it never
sees the CIDR list.

Net effect for a user who sets `allowed-source-cidrs = ["100.64.0.0/10"]`
and visits `http://<their-tailnet-ip>:7509/`:

- ✅ Web UI and static dApp assets load (HTTP path passes
  `private_network_filter`)
- ❌ Every WebSocket upgrade from a dApp (River, freenet-ping, etc.) is
  rejected with `403 "WebSocket connections from this origin are not
  allowed"` because the browser's `Host: <tailnet-ip>:7509` is not in
  `allowed_hosts`.

Reported on Matrix by Ivvor (2026-04-18) against v0.2.47.

## Solution

When the operator opts in to `allowed-source-cidrs`, accept a
WebSocket `Host` header that is an **IP literal inside one of those
CIDRs** as a third allowed case — in addition to localhost-origin and
the existing `allowed_hosts` match.

Key property: **IPs only**. Hostnames never match this new path. So
the DNS-rebinding defence the Host-header check exists for is
preserved — an attacker at `evil.com` cannot smuggle through by
resolving their domain to a tailnet IP, because `connection_info`
never resolves DNS. It only accepts `Host` values that parse as
IP literals and then tests CIDR membership.

### Why IP-only (not also DNS)

A rebinding attack sets the attacker-controlled hostname to resolve
to a local IP after TTL expiry. The `Host` header the browser sends
in that attack is the **hostname** (`evil.com`), not the IP — that's
the whole point of the rebind: the browser thinks it's still talking
to `evil.com`. So rejecting hostnames on this path is sufficient to
defeat rebinding without requiring the operator to duplicate every
peer's tailnet IP into `allowed-host`.

### Alternative considered

Documenting that `--allowed-source-cidrs` must be paired with
`--allowed-host <my-tailnet-ip>` is one line of docs. But a tailnet
may assign many IPs over its lifetime (new device, reassignment, IPv6)
and every peer's browser sends a different Host. Making the operator
keep two flags in sync is a footgun; accepting IPs in the already-
trusted CIDR range has no additional security cost and zero user
friction.

### Layer wire-up

`AllowedSourceCidrs` is now layered as an `Extension` on the WS
router in **both** the loopback and LAN branches of
`serve_client_api_in_impl`, so `connection_info` can extract it
unconditionally. The loopback branch always gets an empty CIDR list
(matching config), so there is **zero behaviour change** for users
who don't opt in.

## Testing

New unit tests in `crates/core/src/client_events/websocket.rs`:

- `parse_host_header_ip_variants` — parse of `host:port`, `[v6]`,
  `[v6]:port`, bare v6, hostnames (must return None), empty input
- `host_header_ip_in_tailscale_cidr_allowed` — regression for the
  Ivvor/v0.2.47 scenario: `100.64.x.x:7509` accepted under
  `100.64.0.0/10`
- `host_header_ip_outside_cidr_rejected` — IPs just outside the CIDR
  (`100.128.0.1`, `100.63.255.255`, `8.8.8.8`) rejected
- `host_header_hostname_never_accepted_via_cidr` — DNS-rebinding
  guard: `evil.com`, `evil.com:7509`, `100-64-1-5.evil.com` all
  rejected even with a non-empty CIDR list
- `host_header_empty_cidrs_matches_nothing` — default-deny: empty
  CIDR list matches zero IPs (no posture drift for non-opt-in users)
- `host_header_ipv6_in_cidr_allowed` — `[fd7a:115c:a1e0::1]:7509`
  accepted under `fd7a:115c:a1e0::/48`; `[fd7a:115c:a1e1::1]` rejected
- `host_header_missing_rejected` — no Host header → rejected (no
  panic, no accept)

Existing coverage retained:
- `is_allowed_host` tests unchanged
- `middleware_layer_stack_allows_configured_source` in `server.rs`
  still validates the source-IP branch of the layer stack

Local validation:
- `cargo fmt --all` clean
- `cargo clippy --locked -- -D warnings` clean (same invocation CI uses)
- `cargo test -p freenet --lib` — **2382 passed, 0 failed, 16
  ignored** (no pre-existing ignores touched)

### Why unit tests instead of integration

`host_header_ip_in_cidrs` is a pure function (HeaderMap +
`AllowedSourceCidrs` → bool). Boundary-focused table tests give
stronger signal than a full axum+tower e2e would. The
regression that triggered this PR is a pure Host-header parsing +
CIDR-membership decision; if this function is correct, the middleware
is correct. The layer-ordering regression test in `server.rs`
(`middleware_layer_stack_allows_configured_source`) already covers
the case where someone reorders the Extension stack and breaks
extractor injection; extending that test to the WS router would add
complexity without catching a new failure mode.

## Fixes

Reported on Matrix by Ivvor (2026-04-18) against v0.2.47. No GitHub
issue; filing one would duplicate this PR.

[AI-assisted - Claude]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,4 @@ test-results/
 worktrees/
 logs/
 .grepai/
+.codex
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -172,6 +172,61 @@ fn is_allowed_host(headers: &axum::http::HeaderMap, allowed_hosts: &HashSet<Stri
     allowed_hosts.contains(&host_header.to_lowercase())
 }
 
+/// Returns `true` if the request's `Host` header is an IP literal that falls
+/// within one of the operator-allowed source CIDRs.
+///
+/// When a user opts in to `--allowed-source-cidrs` (e.g. for Tailscale access),
+/// they will reach the node via an IP in that range. The browser's `Host`
+/// header for such a request is that IP literal, so we treat it as a
+/// first-class allowed host — saving operators from having to duplicate every
+/// address on every peer in a separate `allowed-host` list.
+///
+/// Hostnames (including DNS-rebinding bait like `evil.com`) are ignored here;
+/// only string-parseable IP literals are considered. This preserves the
+/// DNS-rebinding protection `is_allowed_host` exists for: the attacker cannot
+/// smuggle a hostname through by resolving it locally, because we never
+/// resolve — we only accept IPs that are already in the operator's trusted
+/// range.
+fn host_header_ip_in_cidrs(
+    headers: &axum::http::HeaderMap,
+    allowed_cidrs: &crate::server::AllowedSourceCidrs,
+) -> bool {
+    if allowed_cidrs.is_empty() {
+        return false;
+    }
+    let Some(host_header) = headers
+        .get(axum::http::header::HOST)
+        .and_then(|h| h.to_str().ok())
+    else {
+        return false;
+    };
+    let Some(ip) = parse_host_header_ip(host_header) else {
+        return false;
+    };
+    allowed_cidrs.contains_ip(&ip)
+}
+
+/// Extracts the IP part of a `Host` header value, handling `host:port`,
+/// `[v6]:port`, and bare IPv6 forms. Returns `None` if the host part isn't a
+/// parseable IP literal (i.e. it's a hostname).
+fn parse_host_header_ip(host_header: &str) -> Option<std::net::IpAddr> {
+    let host = host_header.trim();
+    // Bracketed IPv6 form: `[::1]` or `[::1]:7509`.
+    if let Some(rest) = host.strip_prefix('[') {
+        let end = rest.find(']')?;
+        return rest[..end].parse().ok();
+    }
+    // For non-bracketed forms, treat a single colon as `host:port`; anything
+    // with more than one colon is a bare IPv6 literal without brackets
+    // (technically non-conformant, but browsers never produce it — we still
+    // try to parse it for safety).
+    let candidate = match host.as_bytes().iter().filter(|&&b| b == b':').count() {
+        0 | 1 => host.rsplit_once(':').map(|(h, _)| h).unwrap_or(host),
+        _ => host,
+    };
+    candidate.parse().ok()
+}
+
 /// Checks if an error represents a client-side disconnect rather than a server error.
 ///
 /// Client disconnects are expected behavior when clients timeout, crash, or close
@@ -420,6 +475,7 @@ async fn connection_info(
         streaming: _,
     }): Query<ConnectionInfo>,
     Extension(allowed_hosts): Extension<crate::server::AllowedHosts>,
+    Extension(allowed_source_cidrs): Extension<crate::server::AllowedSourceCidrs>,
     mut req: axum::extract::Request,
     next: axum::middleware::Next,
 ) -> Response {
@@ -481,8 +537,23 @@ async fn connection_info(
     if is_ws_upgrade {
         if let Some(origin) = req.headers().get(axum::http::header::ORIGIN) {
             if let Ok(origin_str) = origin.to_str() {
+                // Accept the upgrade if any of:
+                // 1. The origin is a localhost form (browser page loaded from
+                //    the same machine), OR
+                // 2. The Host header matches an auto-detected or operator-
+                //    configured hostname (DNS-rebinding protection), OR
+                // 3. The Host header is an IP literal inside one of the
+                //    operator-allowed source CIDRs. This is the path that
+                //    unblocks Tailscale-style access: the node is reached
+                //    at e.g. `http://100.64.1.5:7509`, so the Host header is
+                //    `100.64.1.5:7509` — never in `allowed_hosts` unless the
+                //    operator duplicated every peer address into it.
+                //    Matching by IP (not hostname) preserves the rebinding
+                //    defense: a malicious `evil.com` can't smuggle through
+                //    because we never resolve DNS here.
                 if !is_localhost_origin(origin_str)
                     && !is_allowed_host(req.headers(), &allowed_hosts)
+                    && !host_header_ip_in_cidrs(req.headers(), &allowed_source_cidrs)
                 {
                     let host_header = req
                         .headers()
@@ -1647,15 +1718,6 @@ mod tests {
         .map(|s| s.to_string())
         .collect();
 
-        fn headers_with_host(host: &str) -> axum::http::HeaderMap {
-            let mut map = axum::http::HeaderMap::new();
-            map.insert(
-                axum::http::header::HOST,
-                axum::http::HeaderValue::from_str(host).unwrap(),
-            );
-            map
-        }
-
         // Allowed: host with port
         assert!(is_allowed_host(
             &headers_with_host("192.168.1.50:7509"),
@@ -1692,6 +1754,166 @@ mod tests {
         assert!(is_allowed_host(
             &headers_with_host("LOCALHOST:7509"),
             &allowed
+        ));
+    }
+
+    fn headers_with_host(host: &str) -> axum::http::HeaderMap {
+        let mut map = axum::http::HeaderMap::new();
+        map.insert(
+            axum::http::header::HOST,
+            axum::http::HeaderValue::from_str(host).unwrap(),
+        );
+        map
+    }
+
+    fn cidrs(list: &[&str]) -> crate::server::AllowedSourceCidrs {
+        crate::server::AllowedSourceCidrs(Arc::new(
+            list.iter().map(|s| s.parse().unwrap()).collect(),
+        ))
+    }
+
+    /// Parsing: `host:port`, `[v6]`, `[v6]:port`, bare v6, and hostnames.
+    #[test]
+    fn parse_host_header_ip_variants() {
+        assert_eq!(
+            parse_host_header_ip("100.64.1.5:7509"),
+            Some("100.64.1.5".parse().unwrap())
+        );
+        assert_eq!(
+            parse_host_header_ip("100.64.1.5"),
+            Some("100.64.1.5".parse().unwrap())
+        );
+        assert_eq!(
+            parse_host_header_ip("[fd7a:115c:a1e0::1]:7509"),
+            Some("fd7a:115c:a1e0::1".parse().unwrap())
+        );
+        assert_eq!(
+            parse_host_header_ip("[fd7a:115c:a1e0::1]"),
+            Some("fd7a:115c:a1e0::1".parse().unwrap())
+        );
+        // Bare IPv6 without brackets (non-conformant but parseable).
+        assert_eq!(
+            parse_host_header_ip("fd7a:115c:a1e0::1"),
+            Some("fd7a:115c:a1e0::1".parse().unwrap())
+        );
+        // Hostnames must not parse as IPs.
+        assert_eq!(parse_host_header_ip("evil.com"), None);
+        assert_eq!(parse_host_header_ip("evil.com:7509"), None);
+        assert_eq!(parse_host_header_ip("localhost:7509"), None);
+        assert_eq!(parse_host_header_ip(""), None);
+    }
+
+    /// Regression for the v0.2.47 gap: a user configuring
+    /// `allowed-source-cidrs = ["100.64.0.0/10"]` to reach their node via
+    /// Tailscale must be able to complete the WebSocket upgrade. The Host
+    /// header they send is their own tailnet IP, which is not in
+    /// `allowed_hosts` unless duplicated manually. Before the fix, `is_ws_upgrade`
+    /// rejected these with 403.
+    #[test]
+    fn host_header_ip_in_tailscale_cidr_allowed() {
+        let allowed = cidrs(&["100.64.0.0/10"]);
+        assert!(host_header_ip_in_cidrs(
+            &headers_with_host("100.64.1.5:7509"),
+            &allowed,
+        ));
+        assert!(host_header_ip_in_cidrs(
+            &headers_with_host("100.64.1.5"),
+            &allowed,
+        ));
+        // A neighbouring user's tailnet IP is still inside 100.64.0.0/10 —
+        // if the operator widened the CIDR, that's the operator's choice.
+        assert!(host_header_ip_in_cidrs(
+            &headers_with_host("100.127.9.9:7509"),
+            &allowed,
+        ));
+    }
+
+    /// IPs that fall OUTSIDE the operator's CIDR must still be rejected,
+    /// even when the CIDR list is non-empty. This is the regression guard
+    /// against "CIDR match accidentally widened to any IP".
+    #[test]
+    fn host_header_ip_outside_cidr_rejected() {
+        let allowed = cidrs(&["100.64.0.0/10"]);
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("8.8.8.8:7509"),
+            &allowed,
+        ));
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("100.128.0.1:7509"),
+            &allowed,
+        ));
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("100.63.255.255:7509"),
+            &allowed,
+        ));
+    }
+
+    /// DNS-rebinding defense: hostnames must NEVER be accepted via the CIDR
+    /// path, even if the attacker sets up a DNS record that resolves to an
+    /// IP inside the allowed CIDR. We intentionally do not resolve DNS in
+    /// this middleware; only string-parseable IP literals are considered.
+    #[test]
+    fn host_header_hostname_never_accepted_via_cidr() {
+        let allowed = cidrs(&["100.64.0.0/10"]);
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("evil.com"),
+            &allowed,
+        ));
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("evil.com:7509"),
+            &allowed,
+        ));
+        // A hostname that happens to look like an IP at a glance — still a
+        // hostname to `IpAddr::from_str`, so still rejected.
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("100-64-1-5.evil.com"),
+            &allowed,
+        ));
+    }
+
+    /// Empty CIDR list is the default posture. Zero IPs must match, so a
+    /// user who never opted in gets no behaviour change. This is load-bearing
+    /// for the default-deny guarantee.
+    #[test]
+    fn host_header_empty_cidrs_matches_nothing() {
+        let empty = crate::server::AllowedSourceCidrs::default();
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("100.64.1.5:7509"),
+            &empty,
+        ));
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("127.0.0.1:7509"),
+            &empty,
+        ));
+    }
+
+    /// IPv6 CIDRs must work end-to-end, including the bracketed `[v6]:port`
+    /// form browsers send in the Host header.
+    #[test]
+    fn host_header_ipv6_in_cidr_allowed() {
+        let allowed = cidrs(&["fd7a:115c:a1e0::/48"]);
+        assert!(host_header_ip_in_cidrs(
+            &headers_with_host("[fd7a:115c:a1e0::1]:7509"),
+            &allowed,
+        ));
+        assert!(host_header_ip_in_cidrs(
+            &headers_with_host("[fd7a:115c:a1e0::1]"),
+            &allowed,
+        ));
+        // Outside the /48.
+        assert!(!host_header_ip_in_cidrs(
+            &headers_with_host("[fd7a:115c:a1e1::1]:7509"),
+            &allowed,
+        ));
+    }
+
+    /// Missing Host header must not panic and must not accept.
+    #[test]
+    fn host_header_missing_rejected() {
+        let allowed = cidrs(&["100.64.0.0/10"]);
+        assert!(!host_header_ip_in_cidrs(
+            &axum::http::HeaderMap::new(),
+            &allowed,
         ));
     }
 
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -175,18 +175,11 @@ fn is_allowed_host(headers: &axum::http::HeaderMap, allowed_hosts: &HashSet<Stri
 /// Returns `true` if the request's `Host` header is an IP literal that falls
 /// within one of the operator-allowed source CIDRs.
 ///
-/// When a user opts in to `--allowed-source-cidrs` (e.g. for Tailscale access),
-/// they will reach the node via an IP in that range. The browser's `Host`
-/// header for such a request is that IP literal, so we treat it as a
-/// first-class allowed host — saving operators from having to duplicate every
-/// address on every peer in a separate `allowed-host` list.
-///
-/// Hostnames (including DNS-rebinding bait like `evil.com`) are ignored here;
-/// only string-parseable IP literals are considered. This preserves the
-/// DNS-rebinding protection `is_allowed_host` exists for: the attacker cannot
-/// smuggle a hostname through by resolving it locally, because we never
-/// resolve — we only accept IPs that are already in the operator's trusted
-/// range.
+/// Only string-parseable IP literals are considered; hostnames are never
+/// matched. This preserves the DNS-rebinding defense provided by
+/// `is_allowed_host`: an attacker at `evil.com` cannot smuggle through by
+/// resolving their hostname to an IP in the allowed range, because we never
+/// resolve DNS here.
 fn host_header_ip_in_cidrs(
     headers: &axum::http::HeaderMap,
     allowed_cidrs: &crate::server::AllowedSourceCidrs,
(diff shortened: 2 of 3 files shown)


sanity:
## Code-First Review

Formed an understanding from the diff before reading the description, then compared.

### What the code actually does
- Adds `host_header_ip_in_cidrs` + `parse_host_header_ip` in `client_events/websocket.rs`.
- Extends the WS-upgrade gate in `connection_info` with a third accept branch: localhost-origin OR `allowed_hosts` match OR Host-header IP literal inside an operator CIDR.
- Layers `Extension<AllowedSourceCidrs>` on the WS router in both the LAN branch (already had it for `private_network_filter`) and the loopback branch (newly added, so the extractor always resolves).
- Renames `AllowedSourceCidrs::contains` to `contains_ip` and exposes `is_empty` for early-exit.
- Normalizes IPv4-mapped IPv6 before CIDR membership (mirrors `is_source_allowed`).
- Tests cover `host:port`, `[v6]`, `[v6]:port`, bare v6, hostnames, empty Host, IPv4-mapped v6, IPv6 CIDR boundaries, default-deny with empty list.

Accept ordering (`is_localhost_origin || is_allowed_host || host_header_ip_in_cidrs`) is safe: all three are independent pure checks, and the CIDR arm rejects hostnames by design (it never parses a non-literal as an IP).

### Discrepancies vs description
None material. Two minor doc nits:

1. The loopback-branch comment in `server.rs:606-608` says "Loopback binds still inject an (empty) `AllowedSourceCidrs`". Strictly speaking, the list is whatever the operator configured in `config.allowed_source_cidrs` — it just happens to be functionally unused on loopback (since loopback traffic already passes the source-IP gate trivially). Harmless, but the parenthetical "(empty)" is slightly off. Matches the description's own wording ("always gets an empty CIDR list (matching config)"), so the description is wrong in the same way. The real invariant is "whatever the operator set, but nothing depends on it being empty".

### Concerns

1. **Bare IPv6 Host headers with trailing `:port` are parsed ambiguously.** `parse_host_header_ip("fd7a::1:7509")` first tries `host.parse::<IpAddr>()`, which **succeeds** — bincode-style: `fd7a::1:7509` is a valid IPv6 literal (equivalent to `fd7a:0:0:0:0:0:1:7509`). So the code does NOT strip the port; it treats `7509` as the last v6 group. In practice RFC 7230 requires brackets for IPv6 in Host headers, so well-behaved clients send `[fd7a::1]:7509`. An attacker who sends a hand-crafted `Host: fd7a::1:7509` would not gain anything (still needs the resulting IP to be in the operator's CIDR), but the `parse_host_header_ip_variants` test only asserts the `host.parse()` success path for bare v6 without a trailing colon group — consider adding a test for `fd7a::1:7509` to lock in that behavior (either "rejects because not in CIDR" or "accepts because `fd7a::0:0:0:1:7509` is in the operator's subnet", whichever you prefer). This is behavior-documenting, not a bug fix.

2. **Host header with userinfo (`user@host:port`).** `parse_host_header_ip` does not strip userinfo. Host headers should never carry userinfo (RFC 7230 forbids it), so this is almost certainly irrelevant, but a malicious `Host: user:pass@100.64.1.5:7509` would `rsplit_once(':')` → `("user:pass@100.64.1.5", "7509")`, then `"user:pass@100.64.1.5".parse::<IpAddr>()` → Err, and the helper returns `None`. Correct behavior (reject), so this is fine — worth a one-line test if you want to pin it.

3. **`AllowedSourceCidrs` debug/logging path on LAN branch.** `allowed_source_cidrs` is cloned cheaply (`Arc<Vec<...>>`) for both `.layer(Extension(..))` and the earlier `for cidr in allowed_source_cidrs.0.iter()` loop. No issue — noting it's not double-consumed.

4. **`is_empty()` short-circuit.** `host_header_ip_in_cidrs` returns false early when the CIDR list is empty, before even looking at the Host header. Good — preserves "zero behaviour change for non-opt-in users" claim, and avoids the hostname-resolution cost for the common case.

### Test coverage — strong
The suite covers the Ivvor scenario, DNS-rebinding defense, boundary IPs on both sides of `/10`, IPv6 `/48` inside/outside, empty list, missing Host header, and hostname-that-looks-IP-ish (`100-64-1-5.evil.com`). No gaps on the regression this PR targets.

The description's justification for unit-over-integration ("if this function is correct the middleware is correct") is accurate given the middleware just `&&`s the three predicates — the existing `middleware_layer_stack_allows_configured_source` test in `server.rs` already pins the Extension wire-up for the LAN branch. Adding an equivalent pin for the loopback branch would be cheap and would catch a future regression where someone drops the `.layer(Extension(allowed_source_cidrs))` from the loopback arm. Consider it a nice-to-have, not a blocker.

### Documentation recommendations
- Minor: reword the "(empty)" comment in `server.rs:606` to "Loopback binds still inject `AllowedSourceCidrs` so `connection_info` always finds the extractor; in practice the CIDR check is moot here because loopback traffic already trivially passes the source-IP gate." Same idea in the PR description.
- Consider adding one line to `host_header_ip_in_cidrs`'s doc about IPv4-mapped-v6 normalization — it's subtle and matches `is_source_allowed`, so the invariant is worth calling out.

### Bottom line
Implementation matches the stated intent. The CIDR branch is gated on IP literal parsing so DNS rebinding stays blocked. Extension wire-up is correct on both branches. Tests are thorough for the target bug. No blockers.

[AI-assisted - Claude]

sanity:
## Big Picture Review

### 1. Goal alignment: yes, tightly scoped

Stated goal (unblock WS upgrades for users who set `--allowed-source-cidrs` for Tailscale/overlays) matches the implementation precisely:

- A new `host_header_ip_in_cidrs` predicate gates a *third* accept-path in `connection_info` alongside the existing localhost-origin and `allowed_hosts` paths.
- `AllowedSourceCidrs` is added as an `Extension` on both branches of `serve_client_api_in_impl` (loopback and non-loopback). The loopback branch injects an empty `AllowedSourceCidrs` so the extractor always finds the value — missing it would 500 every WS upgrade, and the comment at server.rs:606-608 calls this out explicitly.
- IPs only (hostnames never match). The DNS-rebinding defense from #3576 is preserved — an attacker at `evil.com` who rebinds to a tailnet IP still sends `Host: evil.com`, which fails IP parsing and falls through to the existing `is_allowed_host` path.

No scope creep. The only non-WS change is `AllowedSourceCidrs::contains` → `contains_ip` (visibility + rename), which is a trivial refactor forced by the new call site in `websocket.rs`. The rename is reasonable because the original name would collide with `AllowedSourceCidrs::contains` on a hypothetical `Vec`-ish API surface; `contains_ip` is less ambiguous.

### 2. Docs: no update needed

I checked `~/code/freenet/web/main/hugo-site/content/build/manual/remote-access.md` and grepped the entire web repo: **the public docs never mention `--allowed-host`.** The manual only documents `allowed-source-cidrs`. So the premise in the review prompt (that the docs currently recommend pairing the two) is false — no doc update is required for this PR.

One minor observation for later: the remote-access.md warning about CGNAT ranges (`100.64.0.0/10` being shared between ISP subscribers) is now **more** relevant because this PR extends trust to any IP in those CIDRs over the Host-header path too. The doc already warns strongly enough; no change needed, but worth noting that users who set an over-broad CIDR now have two gates widened by it rather than one.

### 3. Agent/skill instructions: no drift

- No agent or skill file references `AllowedHosts`, `AllowedSourceCidrs`, `connection_info`, `is_allowed_host`, `private_network_filter`, or the `allowed-host` / `allowed-source-cidrs` config keys. Grepped `~/.claude/skills/`, `~/.claude/rules/`, `~/code/freenet/*/CLAUDE.md`, and `~/code/freenet/*/AGENTS.md` — no hits.
- `crates/core/src/server.rs:596` references `middleware_has_extension_injected`, but that comment was already in the file and the referenced test still exists. The new comment about `connection_info` extracting `Extension<AllowedSourceCidrs>` is correctly added.
- No hooks changed.

### 4. Tests removed/weakened/ignored: none

- `is_allowed_host` tests retained verbatim.
- `middleware_layer_stack_allows_configured_source` retained.
- Renamed `contains` → `contains_ip` in six existing tests — trivial mechanical change.
- `headers_with_host` helper was hoisted out of the `is_allowed_host` test fn to module scope so the new tests can share it. No logic change; this is standard refactoring.
- 7 new tests added covering: host-header IP parsing variants (v4/v6, bracketed, bare, port-stripped), CGNAT CIDR acceptance, outside-CIDR rejection, hostname DNS-rebinding guard (`evil.com`, `evil.com:7509`, `100-64-1-5.evil.com`), empty-CIDR default-deny, IPv6 end-to-end, missing Host header rejection.
- No `#[ignore]` added. No assertions weakened.

The test coverage is stronger than the code it guards. In particular the `100-64-1-5.evil.com` rebinding test and the empty-CIDR default-deny test are the most valuable — they lock in the two properties a future refactor could silently violate.

### 5. CI gap

There is no integration test that actually exercises a WS upgrade against a non-loopback bind with `allowed-source-cidrs` configured. This PR does not add one; the justification in the description (pure-function unit tests + existing layer-stack regression test) is reasonable. **But note:** the existing `middleware_layer_stack_allows_configured_source` test only covers the HTTP source-IP branch, not the WS upgrade path. A future regression where the loopback branch at server.rs:609-612 accidentally drops `.layer(Extension(allowed_source_cidrs))` would not be caught by any existing test — `connection_info` would 500 on every WS upgrade from loopback, and unit tests alone wouldn't detect it.

Recommendation (not blocking): consider extending `middleware_layer_stack_allows_configured_source` or adding a sibling test that drives the *WS router's* layer stack with a synthetic upgrade request and both allowed and disallowed Host headers. This would close the gap for ~30 lines of test code and match the precedent the PR description cites.

### 6. Symptom vs root cause: fix matches report

The Matrix report from Ivvor described the exact symptom now covered by `host_header_ip_in_tailscale_cidr_allowed`: Tailscale IP in allowed-source-cidrs, HTTP assets load, WS upgrade 403s with "WebSocket connections from this origin are not allowed". Root cause is correctly identified: `allowed-source-cidrs` extends `private_network_filter` (source-IP gate) but not `connection_info`'s Host-header gate. The fix addresses the root cause in the correct location.

Related in-flight work: PR #3890 (merged 2026-04-16) fixed a separate bug where `allowed-source-cidrs` from config.toml was silently ignored because the config merge block missed it. That was the first gap Ivvor hit; this PR is the second. Both originated from the same user report. No conflict or overlap between the two PRs — they fix different layers.

### Verdict

Ship it. Clean, tightly-scoped, well-tested, no landmines for future work. The only optional follow-up is a layer-stack test for the WS router; not a blocker.

[AI-assisted - Claude]

sanity:
## Testing Review: PR #3909

### Coverage Summary
- Direct changes: **covered** (7 new unit tests + 1 uncommitted IPv4-mapped v6 test)
- Downstream impact on `connection_info` middleware wiring: **partial** — no axum-level test drives the full WS router with `Extension<AllowedSourceCidrs>` attached/detached

### Test Gaps Found

#### Critical (blocks merge)
None. The pure function is well-covered at the unit level and the Ivvor/v0.2.47 regression scenario (`host_header_ip_in_tailscale_cidr_allowed`) reproduces the exact failing case.

#### Important (should have)
1. **Uncommitted IPv4-mapped v6 test is in the working tree** (`host_header_ipv4_mapped_v6_normalized_to_v4`, plus the normalization block in `host_header_ip_in_cidrs`). Currently `git status` shows these as unstaged — they need to be committed before merge or they're lost. That test is genuinely important: without the normalization, an operator writing an IPv4 CIDR (`100.64.0.0/10`) would reject a browser that sent `Host: [::ffff:100.64.1.5]:7509`, which some dual-stack stacks will produce. I confirmed behavior by probing `IpAddr::from_str` directly. Strongly recommend keeping both the code and the test.

2. **Loopback-branch Extension-injection regression**. The loopback branch of `serve_client_api_in_impl` now layers `Extension(allowed_source_cidrs)` too (previously it didn't need to). If a future refactor removes it, `connection_info` will 500 every WS upgrade on `127.0.0.1` — which is the dev-default config and would be caught by nobody's local run until a user hit it. The existing `middleware_layer_stack_allows_configured_source` test only covers `private_network_filter` on the LAN branch; it doesn't drive `connection_info` and wouldn't catch the missing loopback extension.

   Concrete suggestion: add a `tokio::test` that builds a router with `axum::middleware::from_fn(connection_info)` + only `Extension(AllowedHosts)` (no `AllowedSourceCidrs`), sends a WS upgrade with an Origin+Host, and asserts it does NOT 500. Then the same test with both Extensions asserts 101/200. Total: ~30 lines, directly mirrors `middleware_layer_stack_allows_configured_source`.

#### Nice to Have
3. **Host headers with user-info prefix** (`user@100.64.1.5:7509`). RFC 7230 says the Host header MUST NOT include userinfo, but a misbehaving client could still send it. Current behavior: `parse_host_header_ip` returns `None` (rejected), which is the right outcome. A one-line test documenting this would lock the behavior: `assert_eq!(parse_host_header_ip("user@100.64.1.5:7509"), None);`.

4. **Mixed-case hex IPv6 in Host** (`[FD7A:115C:A1E0::1]:7509`). Works today via `IpAddr::from_str`, but not asserted. One-line addition to `parse_host_header_ip_variants`.

5. **Trailing whitespace in Host header**. `host.trim()` handles it, but not tested.

6. **Comment drift in `server.rs`**. The layer-ordering comment (around line 596) references a test called `middleware_has_extension_injected`, but the actual test is `middleware_layer_stack_allows_configured_source`. Cosmetic but will mislead the next reader.

### Tests at Risk of Being Weakened
None identified. The change is additive — it widens the accept path via a separate OR branch. None of the existing `is_allowed_host` tests were modified, and the DNS-rebinding defense is preserved by the IP-only gate (verified by `host_header_hostname_never_accepted_via_cidr`).

### CI Considerations
All new tests are pure-function unit tests; total runtime negligible. No flakiness surface area.

### Verdict
Unit-level coverage is strong. Two follow-ups recommended before merge:
1. Commit the IPv4-mapped v6 normalization + test (currently uncommitted).
2. Add a middleware-level test for `connection_info` that asserts the loopback branch doesn't 500 due to a missing `AllowedSourceCidrs` extension. This closes the one regression mode that the current unit tests can't catch.

[AI-assisted - Claude]

sanity:
## Skeptical Review: PR #3909

Thorough adversarial review focused on the seven investigation points. One high-severity finding; the rest hold up.

### Potential Bugs Found

#### High severity

**1. Cross-site WebSocket hijacking (CSWSH) is now possible from a foreign Origin when the victim is on the allowed CIDR**

The PR description frames the change as preserving the DNS-rebinding defense because hostnames aren't matched. That argument is correct for *DNS rebinding* in the narrow sense (attacker hostname briefly resolves to a private IP; browser still sends `Host: evil.com` which fails the IP-only path). **But the Host-header allowlist from #3576 also guarded a second attack class that this PR weakens: cross-site WebSocket hijacking against a known IP.**

Concrete trace:
- Victim has `allowed-source-cidrs = ["100.64.0.0/10"]`, bound to `100.64.1.5:7509`.
- Victim's browser is on the tailnet (source IP `100.64.x.x`), so any connection it initiates passes `private_network_filter`.
- Victim visits `http://evil.com` (or any page the attacker controls).
- `evil.com` JS calls `new WebSocket("ws://100.64.1.5:7509/v1/contract/command")`.
- Browser sends `Host: 100.64.1.5:7509`, `Origin: http://evil.com`.
- `is_localhost_origin("http://evil.com")` = false.
- `is_allowed_host` = false (operator did not duplicate the IP into `allowed-host`).
- **`host_header_ip_in_cidrs` = true** — `100.64.1.5` is inside `100.64.0.0/10`.
- Upgrade accepted. `evil.com` now has an unauthenticated WS session with `websocket_commands` and can issue contract GET/PUT/UPDATE/SUBSCRIBE requests against the victim's node (and exfiltrate any responses back to the attacker's server).

This is exactly the attack class #3576 was written to block. The prior requirement ("operator must explicitly add IPs to `allowed-host`") was load-bearing: it forced a deliberate acknowledgement that a specific IP is a legitimate Host the browser should present. The PR turns that into automatic acceptance for any IP in the CIDR.

The PR body's argument that "accepting IPs in the already-trusted CIDR range has no additional security cost" conflates source-IP trust (who can open the TCP socket) with Host-header trust (what the browser thinks it's connecting to — a CSRF-class check). They're not the same trust decision; the original allowlist split them on purpose.

Suggested fixes (any one):
- **Preferred:** revert to the documented-alternative approach (tell operators to pair `--allowed-source-cidrs` with `--allowed-host <tailnet-ip>`). The PR rejects this as a "footgun" but that friction is the security feature — it forces the operator to enumerate *which IPs* are valid Host values, not *any IP in the CIDR*.
- Accept IP-in-CIDR only when `Origin` is also an IP in the CIDR (or `Origin: null`-with-qualifier). A foreign `Origin: http://evil.com` should still fail.
- Require a separate opt-in flag (`--trust-cidr-as-host` or similar) so operators who understand the CSWSH tradeoff can enable it, rather than coupling it to `--allowed-source-cidrs`.

None of the unit tests in the PR cover this — they test `host_header_ip_in_cidrs` in isolation, not the full `connection_info` decision tree with a foreign `Origin` header. A regression test that exercises the full `if is_ws_upgrade` block with `Origin: http://evil.com` + `Host: <tailnet-ip>` would pin the current (problematic) behavior and make the trade-off explicit.

#### Medium severity

**2. Asymmetry between source-IP CIDR match and Host-header CIDR match for IPv4-mapped IPv6**

`is_source_allowed` (`server.rs:449-461`) normalizes `::ffff:a.b.c.d` → `a.b.c.d` before calling `contains_ip`, so a dual-stack v4 client matches the operator's v4 CIDR.

`host_header_ip_in_cidrs` (`websocket.rs:183-200`) does NOT normalize. Verified empirically with `ipnet::IpNet::V4(100.64.0.0/10).contains(&IpAddr::V6(::ffff:100.64.1.5))` = `false`.

In practice this is unlikely to bite (browsers don't typically emit `Host: [::ffff:a.b.c.d]` for IPv4 destinations), so this is probably cosmetic — but it's an unintentional asymmetry with the sibling function and should be aligned. If a client ever sends the mapped form, the PR will reject a WS upgrade that the source-IP filter just accepted, and the 403 message will be misleading.

Fix: in `host_header_ip_in_cidrs`, apply the same `v6.to_ipv4_mapped()` normalization as `is_source_allowed`, then call `contains_ip`. Add a test case. (Alternative: push the normalization into `AllowedSourceCidrs::contains_ip` itself so both call sites are correct by construction.)

#### Low severity / code smell

**3. `parse_host_header_ip` silently accepts trailing garbage after `]`**

`parse_host_header_ip("[fd7a::1]garbage")` returns `Some(fd7a::1)`. Not a security bug — the IP is still correct for CIDR matching — but it's a quiet departure from RFC3986. A strict parser would return `None`. Worth either documenting as intentional or tightening to require `end == rest.len() - 1` or `rest[end+1..]` starts with `:`.

**4. Bare-IPv6 handling is RFC-non-conformant but the code comment acknowledges it**

`parse_host_header_ip("fd7a::1:7509")` parses as a v6 address `fd7a::1:7509` (because v6 has multiple colons, `host.parse()` succeeds before the port-stripping fallback runs). A browser will never send an unbracketed v6 Host per RFC7230, but a malicious client can. No exploit path I can see; just a landmine if someone later changes behavior.

### What I verified and found to be safe

- **Q3 — IPv4-mapped-in-v6 bypass:** Tested `contains_ip(&IpAddr::V6(::ffff:8.8.8.8))` against `100.64.0.0/10` — returns false. A crafted `[::ffff:100.64.1.5]:7509` Host header is **rejected**, not accepted as a v4 CIDR match, so it's not a bypass. (It's the medium-severity asymmetry above, not a bypass.)

- **Q4 — port in host matches `0.0.0.0/0`:** `validate_source_cidr` (`server.rs:422-435`) rejects any IPv4 CIDR with prefix `<8` at config-load time with a hard error, and IPv6 `<16`. Covered by `validate_source_cidr_rejects_overly_broad_ipv4`. A misconfiguration attempt to set `0.0.0.0/0` fails boot; the prefix-length gate is enforced at the right layer.

- **Q5 — layer ordering on loopback:** The loopback branch adds `Extension(allowed_source_cidrs)` (always empty for loopback) but does NOT add `private_network_filter`. Confirmed `private_network_filter` is only wired in the `needs_lan_filter` arm. `connection_info` extracts the `AllowedSourceCidrs` extractor from the extension stack regardless, so the empty-CIDR path correctly short-circuits at `allowed_cidrs.is_empty()`. Test `middleware_layer_stack_allows_configured_source` (now in server.rs) exercises the LAN arm but there's no analogous "loopback arm extension is present" test for `connection_info` specifically. Adding one would pin the invariant, but it's not actively broken.

- **Q5b — `needs_lan_filter` on `0.0.0.0`/`::`:** Verified `Ipv4Addr::UNSPECIFIED.is_loopback() == false` and `Ipv6Addr::UNSPECIFIED.is_loopback() == false`, so unspecified binds take the LAN path (correct — default-bind nodes get the filter).

- **Q6 — TOCTOU:** No. `HeaderMap` is owned by the request and not mutated between the CIDR check and any subsequent Host parsing. Single-threaded within the request lifecycle.

- **Q7 — panic potential in `host_header_ip_in_cidrs` / `parse_host_header_ip`:** Exercised against `""`, `"   "`, `"["`, `"[::1]garbage"`, `":7509"`, `"[localhost]"`, various malformed forms. All return `None`. Slicing `rest[..end]` where `end = rest.find(']')` uses a valid UTF-8 byte offset (ripgrep-style find returns None or a valid boundary). `headers.get(...).to_str().ok()` rejects non-ASCII control chars before reaching the parser. No panic path found.

### Questions for author

1. Did you consider the CSWSH class (known-IP + foreign Origin) when analyzing the security posture change? The PR body's "no additional security cost" claim holds only for the DNS-rebinding subclass, and I'd like to confirm the CSWSH widening is intentional and judged acceptable (with, e.g., a docs warning), or is an oversight.

2. If CSWSH is an accepted tradeoff for the UX win, can we gate it behind a second explicit opt-in (`--trust-cidr-hosts` or similar) so users who want the source-IP filter but not the Host-IP auto-trust can keep them separate? The two flags today *are* separate; this PR conflates them.

3. The v4-mapped-v6 asymmetry between `is_source_allowed` and `host_header_ip_in_cidrs` — intentional or oversight? If intentional, worth a doc comment; if oversight, worth aligning and adding a test.

[AI-assisted - Claude]

sanity:
## Review response — addressed all findings

Posting one consolidated response to all four review agents + Codex rather than replying four times.

### 🔴 Skeptical (high): CSWSH via foreign Origin + Host-IP-in-CIDR

**Accepted and fixed in 54fb4d39.** The reviewer is right that my initial framing conflated DNS-rebinding (narrowly defeated by IP-only matching) with the broader CSWSH class. Concrete attack from the review:

```
Origin: https://evil.com
Host:   <victim-tailnet-ip>:7509   ← in CIDR
```

My first pass accepted this. Now the CIDR path additionally requires Origin to be one of:
- `null` — Freenet's dApp sandbox iframe (no `allow-same-origin`)
- localhost form — handled by the existing `is_localhost_origin` branch
- IP literal matching the same Host IP — direct browser access

Attacker's `Origin: https://evil.com` parses to the hostname `evil.com`, not an IP, so `origin_compatible_with_host_ip` returns false. New test `cswsh_cross_origin_real_origin_rejected` pins this. The remaining risk (a malicious sandboxed iframe sending `Origin: null`) is the same risk profile as an operator-configured `--allowed-host` entry, explicitly documented in the code comment.

Did not pursue the reviewer's \"extra opt-in flag\" suggestion — the tightened Origin check closes the specific attack vector without requiring a second flag.

### 🟡 Skeptical (medium) + Codex: IPv4-mapped IPv6 asymmetry

**Fixed in 277290f0 and 54fb4d39.** `host_header_ip_in_cidrs` now normalizes `::ffff:a.b.c.d` → IPv4 before the CIDR check, mirroring `is_source_allowed`. New test `host_header_ipv4_mapped_v6_normalized_to_v4` covers both in-range and out-of-range mapped inputs.

### 🟠 Codex: `parse_host_header_ip` too permissive

**Fixed in 54fb4d39.** The parser now rejects:
- trailing garbage after `]` (`[100.64.1.5]evil.com`)
- bracketed IPv4 (`[100.64.1.5]:7509` — RFC 3986 forbids this)
- non-numeric / empty ports (`[::1]:notaport`, `100.64.1.5:`)
- unbracketed bare IPv6 (`fd7a::1` — ambiguous with `host:port`)

Policy is \"fail closed on any ambiguity\" since this function gates a security decision. Expanded `parse_host_header_ip_variants` with 8 new rejection cases.

### 🟢 Big-picture: docs

Confirmed: `hugo-site/content/build/manual/remote-access.md` never mentioned `--allowed-host` pairing, so there's no stale doc to update.

### 🟢 Testing: integration-level test for loopback-branch Extension wire-up

Not adding. The existing `middleware_layer_stack_allows_configured_source` in `server.rs` pins the LAN branch's layer stack. A sibling test for the loopback branch's `connection_info` stack would duplicate axum wiring with ~6 Extensions and add ~100 lines for a regression class (someone dropping `.layer(Extension(allowed_source_cidrs))`) that already produces immediate loud 500s on every WS upgrade — not subtle failure. The unit tests + rule-lint agent (`<!-- claude-rule-review -->`) give enough signal.

### 🟢 Code-first: doc nit about \"empty\" CIDRs on loopback

Fixed the framing in the code comment — the loopback branch receives whatever the operator configured (possibly non-empty), it's just functionally unused because loopback traffic passes the source-IP gate trivially. The `is_empty()` short-circuit in `host_header_ip_in_cidrs` is what preserves the zero-behaviour-change property for non-opt-in users.

### Test summary after all changes

- 44 tests in `client_events::websocket::tests` (was 30 pre-PR, +14 new)
- 12 new assertions in `parse_host_header_ip_variants` covering malformed inputs
- New CSWSH regression test (`cswsh_cross_origin_real_origin_rejected`)
- Direct-browser-access happy path (`direct_browser_access_same_ip_accepted`)
- `cargo fmt` clean, `cargo clippy --locked -- -D warnings` clean,
  `cargo test -p freenet --lib` all pass

[AI-assisted - Claude]

PR #3915, opened 2026-04-20
fix(operations/update): downgrade benign WASM merge rejections to INFO

### LINKED ISSUE #3914: Downgrade misleading ERROR/WARN logs when contract correctly rejects stale broadcast UPDATE
## Problem

On production gateways (nova, vega) we see ~80–130 `ERROR` log lines per hour matching:

```
ERROR network_event_listener{...}:process_network_message{... tx_type=update}: 
  freenet::operations::update: Failed to update contract value 
  contract=<key> 
  reason: execution error: invalid contract update, 
          reason: New state version <N> must be higher than current version <N>
```

These are **not failures**. The contract WASM is correctly rejecting an incoming broadcast UPDATE whose state is at the same version we already hold (typically a re-broadcast that the 60s `BroadcastDedupCache` TTL missed). The runtime layer one level down already understands this is benign:

```rust
// crates/core/src/contract/executor/runtime.rs:1335-1343
if local_valid {
    tracing::info!(
        contract = %key,
        error = %merge_err,
        local_state_size = current_state.size(),
        incoming_state_size = valid_incoming.size(),
        event = "merge_rejected_valid_local",
        "Merge rejected incoming state but local state is valid - \
         not replacing (incoming state may be stale)"
    );
    return Err(merge_err);
}
```

It logs at **INFO** with `event="merge_rejected_valid_local"`. The op layer above then logs the exact same event at **ERROR** (`update.rs:1701`) and the streaming variant at **WARN** with a misleading message (`update.rs:1315-1319`).

A representative burst (one tx, three log lines, microseconds apart):

```
22:03:58.708616 INFO   contract_handling:upsert_contract_state: 
                       event="merge_rejected_valid_local" 
                       local_state_size=612752 incoming_state_size=612752
22:03:58.708901 ERROR  freenet::operations::update: Failed to update contract value
22:03:58.709059 WARN   freenet::operations::update: 
                       BroadcastToStreaming update skipped — contract not ready locally
```

## User Impact

- Anyone alerting on ERROR-level logs from `freenet::operations::update` gets a constant ~80–130/hr false-positive baseline per gateway.
- The WARN message `"BroadcastToStreaming update skipped — contract not ready locally"` is actively misleading. The contract IS ready locally; the merge ran and the WASM rejected the stale incoming state. The line right below already distinguishes the cases for the self-heal path:
  ```rust
  // update.rs:1321
  if !err.is_contract_exec_rejection() {
      // Self-heal: trigger a background GET ...
      op_manager.try_auto_fetch_contract(key, sender_addr);
  }
  ```
  but the log message above it doesn't.

## Proposed Fix

Two log-level / message changes in `crates/core/src/operations/update.rs`:

1. **Line 1701** (`process_message` Update reply handler): when the failure is a contract-exec rejection (`err.is_contract_exec_rejection()`), log at `info!` with `event="merge_rejected_stale"` instead of `error!`. Reserve `error!` for actual failures (storage errors, missing contract code, internal bugs).

2. **Lines 1315-1319** (`BroadcastToStreaming` handler): split the log into two cases.
   - When `err.is_contract_exec_rejection()` → `info!` with a message like `"BroadcastToStreaming merge rejected — incoming state stale, keeping local"`.
   - Otherwise → keep the existing `warn!` and `"contract not ready locally"` message (the self-heal path).

Same shape as the runtime layer's existing handling.

## Out of Scope (separate investigation)

The **upstream cause** — why peers periodically broadcast a full state that's identical (in version, often in size) to what the receiver already has — is worth understanding separately. Bursts arrive every 1–25 min for the same contract/version on vega; the 60s `BroadcastDedupCache` TTL misses them every time. Each rejected merge runs a WASM `update_state` + a WASM `validate_state` (the `local_valid` check) on a 600 KB state, so there's a real (small) CPU cost. Tracking that down is a separate issue; this one is purely a logging-noise fix.

## Testing

- Unit test in `operations::update::tests` that exercises the `process_message` Update branch with a `ContractExecError` and asserts the log severity/event field. (`tracing-test` already used elsewhere in the crate.)
- Same for the `BroadcastToStreaming` branch.

## References

- `crates/core/src/operations/update.rs:1701` — ERROR site
- `crates/core/src/operations/update.rs:1315-1319` — WARN site (misleading message)
- `crates/core/src/operations/update.rs:1321` — `is_contract_exec_rejection()` is already detected one line below the misleading log
- `crates/core/src/contract/executor/runtime.rs:1327-1345` — the layer that already gets this right

[AI-assisted - Claude]

sanity:
## Problem

On production gateways (nova, vega) we logged ~80–130 ERROR-level lines per hour per gateway from `freenet::operations::update`:

```
ERROR network_event_listener{...}:process_network_message{... tx_type=update}:
  freenet::operations::update: Failed to update contract value
  contract=<key>
  reason: execution error: invalid contract update,
          reason: New state version <N> must be higher than current version <N>
```

These are **not failures**. The contract WASM is correctly rejecting an incoming broadcast UPDATE whose state is at the same version we already hold (typically a re-broadcast that the 60s `BroadcastDedupCache` TTL missed). The runtime layer one level down already understands this is benign and logs it at INFO with `event="merge_rejected_valid_local"`.

The op layer above logged the same event at ERROR (`update.rs:1701`), and the streaming relay path added a WARN with the misleading message `"BroadcastToStreaming update skipped — contract not ready locally"` (`update.rs:1315–1319`). The contract IS ready locally — the merge ran and the WASM rejected the stale state. The line right below already used `err.is_contract_exec_rejection()` to skip auto-fetch in this case, but the log message above didn't follow the same distinction.

A representative burst (one tx, three log lines, microseconds apart):

```
22:03:58.708616 INFO   contract_handling:upsert_contract_state:
                       event="merge_rejected_valid_local"
                       local_state_size=612752 incoming_state_size=612752
22:03:58.708901 ERROR  freenet::operations::update: Failed to update contract value
22:03:58.709059 WARN   freenet::operations::update:
                       BroadcastToStreaming update skipped — contract not ready locally
```

Anyone alerting on ERROR-level logs from `freenet::operations::update` got a constant false-positive baseline that drowns real failures.

## Solution

Both log sites now branch on `err.is_contract_exec_rejection()` (the classifier already used to skip auto-fetch for these errors) and emit INFO with `event=\"merge_rejected_stale\"` when the rejection is benign. Real failures (missing contract code, missing parameters, internal bugs) keep their ERROR/WARN levels and unchanged messages.

The decision was extracted into two small helpers — `log_update_contract_failure` and `log_broadcast_to_streaming_failure` — so the level dispatch is unit-testable in isolation:

- `log_update_contract_failure(&key, &err)` — INFO if contract-exec rejection, otherwise ERROR.
- `log_broadcast_to_streaming_failure(&tx, key, &err) -> bool` — INFO + return `false` (no auto-fetch) if rejection, otherwise WARN + return `true`. The boolean replaces the inline `if !err.is_contract_exec_rejection()` guard at the auto-fetch call site.

`ExecutorError::request` was widened from private to `pub(crate)` so the test helpers can construct the specific `RequestError::ContractError::Update` variants `is_contract_exec_rejection()` matches against. No external API surface change.

## Testing

Four new regression tests in `operations::update::tests::log_severity`, each capturing actual `tracing` output via `TestLogger`:

- `update_contract_failure_logs_info_for_stale_state_rejection` — stale-state rejection produces INFO with `event=\"merge_rejected_stale\"` and zero ERROR lines.
- `update_contract_failure_logs_error_for_real_failure` — \"missing contract parameters\" still produces ERROR with the original message.
- `broadcast_to_streaming_failure_logs_info_and_skips_auto_fetch_for_stale` — stale rejection produces INFO, returns `false` (no self-heal), and the misleading \"contract not ready locally\" line does NOT appear.
- `broadcast_to_streaming_failure_logs_warn_and_triggers_auto_fetch_for_real_failure` — real failure still produces WARN with the original message and returns `true` to trigger self-heal.

Each test would have failed before this PR — they assert both the captured log level (INFO vs ERROR/WARN) and the `event` field, so the regression cannot reappear silently.

All 36 `operations::update` tests pass; all 138 `contract::executor` tests pass; `cargo clippy --all-targets -- -D warnings` clean.

## Out of Scope

The upstream cause — why peers periodically broadcast a full state that's identical (in version, often in size) to what the receiver already has — is worth understanding separately. Each rejected merge runs a WASM `update_state` + a WASM `validate_state` (the `local_valid` check) on a 600 KB state, so there's a real (small) CPU cost. Tracking that down is a separate investigation; this PR is purely the logging-noise fix.

## Fixes

Closes #3914

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -177,7 +177,7 @@ impl ExecutorError {
         }
     }
 
-    fn request(error: impl Into<RequestError>) -> Self {
+    pub(crate) fn request(error: impl Into<RequestError>) -> Self {
         Self {
             inner: Either::Left(Box::new(error.into())),
             fatal: false,
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -177,7 +177,7 @@ impl ExecutorError {
         }
     }
 
-    pub(crate) fn request(error: impl Into<RequestError>) -> Self {
+    fn request(error: impl Into<RequestError>) -> Self {
         Self {
             inner: Either::Left(Box::new(error.into())),
             fatal: false,
@@ -236,7 +236,15 @@ impl ExecutorError {
 
     /// Returns true if this error indicates the contract's WASM merge function
     /// ran and rejected the update (e.g., stale version). This means the contract
-    /// code IS present locally — no auto-fetch is needed.
+    /// code IS present locally, so no auto-fetch is needed.
+    ///
+    /// This is BROADER than `is_invalid_update_rejection`: it ALSO returns true
+    /// for runtime failures like out-of-gas, max-compute-time, traps, etc.,
+    /// because those still mean the contract code is present : only the
+    /// execution itself failed. Use this for auto-fetch decisions, NOT for log
+    /// severity (a contract that runs out of gas is a real bug operators must
+    /// see at ERROR level : see `is_invalid_update_rejection` for the
+    /// log-severity discriminator).
     ///
     /// Only matches errors created via `StdContractError::update_exec_error()`
     /// (cause starts with "execution error:"), NOT other `Update` variants like
@@ -247,6 +255,36 @@ impl ExecutorError {
                 req_err.as_ref(),
                 RequestError::ContractError(StdContractError::Update { cause, .. })
                     if cause.starts_with("execution error")
+            ),
+            Either::Right(_) => false,
+        }
+    }
+
+    /// Returns true ONLY when the contract WASM merge function ran to completion
+    /// and returned a typed `InvalidUpdate` / `InvalidUpdateWithInfo` rejection
+    /// (e.g., "New state version N must be higher than current version N"). This
+    /// is the precise case that production gateways hit on every re-broadcast
+    /// missed by the dedup cache (issue #3914) and the only case where ERROR-
+    /// level logging is operationally noise.
+    ///
+    /// Excluded by design (these remain real failures and keep their ERROR/WARN
+    /// log levels):
+    /// - Out-of-gas / max-compute-time-exceeded
+    /// - WASM traps (stack overflow, division by zero, etc.)
+    /// - Compilation errors, instantiation errors, internal runtime errors
+    /// - Other contract-side `ContractError` variants (`Deser`, `InvalidState`,
+    ///   `InvalidDelta`, `Other`, `DoublePut`, `InvalidArrayLength`, etc.)
+    ///
+    /// Discriminator: stdlib's `ContractError::InvalidUpdate{,WithInfo}` Display
+    /// impls produce strings beginning with "invalid contract update", which
+    /// `update_exec_error` then prefixes with "execution error: ". Any other
+    /// flavor of execution error has a different prefix and falls through.
+    pub fn is_invalid_update_rejection(&self) -> bool {
+        match &self.inner {
+            Either::Left(req_err) => matches!(
+                req_err.as_ref(),
+                RequestError::ContractError(StdContractError::Update { cause, .. })
+                    if cause.starts_with("execution error: invalid contract update")
             ),
             Either::Right(_) => false,
         }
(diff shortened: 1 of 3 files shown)


sanity:
## Review feedback addressed (commit 15a82eb)

Both the skeptical reviewer and Codex independently flagged the same critical issue: `is_contract_exec_rejection()` is too broad for log-severity decisions. It matches any error that flowed through `StdContractError::update_exec_error()`, which includes real WASM faults (OOG, MaxComputeTimeExceeded, DoublePut, traps from `WasmError`). After v1, those would have been silenced from ERROR to INFO with the misleading "incoming state stale" message — operators would lose visibility into buggy/hostile contracts. That's the opposite of what #3914 wants.

### Fix
Added a tighter predicate `ExecutorError::is_invalid_update_rejection()` that matches ONLY the contract WASM's typed `InvalidUpdate{,WithInfo}` rejection (cause string `"execution error: invalid contract update..."`).

- Log severity uses the tight predicate (only the precise stale-state case becomes INFO).
- Auto-fetch keeps using the broad `is_contract_exec_rejection` (correct: OOG/traps still mean contract code is present, no fetch needed).
- The two decisions are now deliberately decoupled in `log_broadcast_to_streaming_failure`.

### Test additions
- `executor_error_tests::test_invalid_update_rejection_*` — predicate matrix: stale rejection ✓, OOG ✗, MaxComputeTime ✗, DoublePut ✗, missing-params ✗.
- `log_severity::update_contract_failure_logs_error_for_out_of_gas` — pins that OOG stays at ERROR.
- `log_severity::broadcast_to_streaming_failure_logs_warn_and_skips_auto_fetch_for_out_of_gas` — asserts WARN level AND skipped auto-fetch independently (validates the decoupling).

### Other reviewer items addressed
- Event field renamed `merge_rejected_stale` → `merge_rejected_invalid_update` to match what the predicate actually matches (per big-picture review on naming consistency).
- Added a comment in `log_broadcast_to_streaming_failure` explaining the `&ExecutorError` vs `&OpError` signature asymmetry (per code-first review).
- Reverted the `pub(crate)` widening of `ExecutorError::request` — tests now use `From<RequestError>` (already public). Per testing reviewer.
- Fixed em-dashes in new code (project style rule). Restored em-dashes in pre-existing surrounding code I had inadvertently mangled.
- Tightened docstrings to NOT overstate the runtime-layer mirror (the runtime checks `local_valid` first; the upper layer can't, but the tighter predicate gets close).

### Items deliberately NOT addressed in this PR
- **`put.rs:1858`** has the same `tracing::error!(...phase = "error", "Failed to update contract value")` template. Filing as a follow-up: PUT-path noise is a separate failure mode (PUT is contract creation, not stale rebroadcast) and the production evidence in #3914 is purely UPDATE.
- **`handle_op_result` ERROR for legacy non-streaming `BroadcastTo`** (`operations.rs:160-167`): the legacy non-streaming relay path returns `Err(err)` after a benign rejection, which `handle_op_result` then logs at ERROR + sends an `Aborted`. The production logs in #3914 don't include a `handle_op_result` line in the burst (only the streaming WARN + the now-INFO update_contract line), suggesting the streaming variant dominates traffic. Worth tracking as a follow-up if the gateway noise persists after this PR ships.

CI status pending — will report when checks complete.

[AI-assisted - Claude]

PR #3910, opened 2026-04-19
refactor(ops): relay UPDATE task-per-tx (slice A) — #1454 phase 5 follow-up

iduartgomez:
## Problem

Phase 5 of #1454 migrated relay GET to the task-per-tx pattern (PR #3896 / `bc8ee9d`). Relay UPDATE still ran on the legacy `OpManager.ops.update` DashMap state machine. UPDATE relays inherit the same amplifier classes that broke GET (per-tx GC blind spots, dedup-set leakage, virtual-time TTL collapse) — keeping UPDATE on legacy blocks the eventual deletion of `OpManager.ops` + `process_message` + `OperationResult` planned for Phase 5 close-out.

## Solution

Slice A (this PR) ports the non-streaming UPDATE relay variants — `RequestUpdate` and `BroadcastTo` — to the task-per-tx driver pattern. Slice B (streaming variants) and slice C (cleanup of unreachable arms in `update.rs::process_message`) follow.

Key insight: UPDATE is fire-and-forget end-to-end (no upstream reply, no retry loop), so the relay driver is structurally simpler than GET — no `ForwardingAck`, no loop-back to preserve. Still adds the per-node dedup gate (`active_relay_update_txs` DashSet + `RelayUpdateInflightGuard` RAII cleanup) to suppress GC-spawned re-entries for the same tx.

Dispatch in `node.rs::handle_pure_network_message_v1` UPDATE branch is gated on:
- `source_addr.is_some()` (true relay; internal callers fall through to legacy)
- `!op_manager.has_update_op(id)` (no pre-registered op; GC retries / `start_targeted_op` fall through)
- non-streaming variants only (`RequestUpdate`, `BroadcastTo`)

`drive_relay_request_update`: have-contract → `update_contract` + telemetry; no-contract → single `send_fire_and_forget` forward; no peers → error + telemetry.

`drive_relay_broadcast_to`: sender summary update → telemetry → `BroadcastDedupCache` BEFORE `update_contract` (to prevent amplification) → on delta failure send `InterestMessage::ResyncRequest` → on full-state failure with missing code call `try_auto_fetch_contract` → on success spawn `send_proactive_summary_notification`.

Streaming variants (`RequestUpdateStreaming`, `BroadcastToStreaming`) and the deprecated `Broadcasting` wire variant stay on the legacy path in slice A — see `docs/port-plans/relay-update-task-per-tx.md` §3.

## Testing

[...]

Issue #3335, opened 2026-02-28
Investigate Network Amplification via UPDATE Broadcast


iduartgomez:
<html><head></head><body>
<h2>Investigate Network Amplification via UPDATE Broadcast</h2>
<h3>Problem Statement (from audit)</h3>
<blockquote>
<p>No rate limiting on broadcast frequency. A contract that always returns <code>changed=true</code> triggers unlimited broadcast cascades.</p>
</blockquote>
<h3>Important context: This finding requires careful analysis</h3>
<p>Unlike M4/M5 which are straightforward missing limits, M7 sits at the intersection of the state synchronization algorithm and security. The codebase already has <strong>multiple layers of write-amplification protection</strong> that the audit may not have fully accounted for. Your primary task is to investigate whether the remaining gap — after all existing protections — is actually exploitable, and if so, what the minimum effective fix is that doesn't break the convergence protocol.</p>
<p><strong>Do not write any code until you have completed the full investigation and reached a clear conclusion.</strong></p>
<h3>What to investigate</h3>
<h4>1. Existing protections (understand before proposing changes)</h4>
<p>The broadcast path has at least <strong>seven</strong> layers of amplification protection already. Read and understand each one before assessing the residual risk.</p>
<p><strong>A. CRDT merge idempotency — <code>UpsertResult::NoChange</code></strong></p>
<p>Read <code>crates/core/src/contract/executor/runtime.rs</code> (~line 1183-1189):</p>
<pre><code class="language-rust">if updated_state.as_ref() == current_state.as_ref() {
    Ok(UpsertResult::NoChange)
} else {
    self.commit_state_update(&amp;key, &amp;params, &amp;updated_state).await?;
    Ok(UpsertResult::Updated(updated_state))
}
</code></pre>
<p>The contract's merge function (WASM <code>update_state</code>) determines the output. If a node applies the same update twice, a correct merge produces the same state → <code>NoChange</code> → no <code>commit_state_update</code> → no <code>BroadcastStateChange</code>. This is the <strong>primary protection</strong> against broadcast storms for well-behaved contracts.</p>
<p>Read <code>crates/core/src/contract/mod.rs</code> (~line 872-882): <code>UpsertResult::CurrentWon</code> also sets <code>state_changed: false</code>.</p>
<p>Read <code>crates/core/src/operations/update.rs</code> (~line 347-352):</p>
<pre><code class="language-rust">if !changed {
[...]

PR #3916, opened 2026-04-20
fix(ops): send summary back to sender when WASM rejects broadcast as stale

sanity:
## Problem

Production gateways (nova, vega) emit ~80–130 ERROR/hr of the form:

```
ERROR freenet::operations::update: Failed to update contract value
      reason: "New state version N must be higher than current version N"
```

Three sites produce the same underlying event at mismatched severity:
- `update.rs:1701` — ERROR (legacy `BroadcastTo`)
- `update.rs:1315-1319` — WARN with misleading `"contract not ready locally"` (legacy `BroadcastToStreaming`)
- `op_ctx_task.rs:949-954` — ERROR (task-per-tx driver from #3910)

The runtime one layer down already knows this is benign and logs INFO `event="merge_rejected_valid_local"` at `runtime.rs:1327-1343`. But beyond the log severity, the underlying behaviour is a real amplifier worth fixing.

## Root cause

The `broadcast_queue.rs` send path falls back to sending full state whenever `get_peer_summary(contract, peer)` returns `None` (`:352-356`), and **only caches the peer's summary after a delta send** (`:511-518`). If the cache starts empty (new peer, or cleared by a prior `ResyncRequest`), it stays empty forever — full-state sends never populate it.

The 5-min `InterestMessage::Summaries` heartbeat is supposed to converge, but production data on vega shows bursts of same-version re-broadcasts every 1–25 min per stale contract, missing the 60s `BroadcastDedupCache` TTL every single time. Each rejection burns a WASM `update_state` + `validate_state` call on a 600 KB state.

The three rejection sites spawn `send_proactive_summary_notification` on the **success** path (which excludes the sender by design: the sender already has what they sent us) but do nothing on the **rejection** path — even though the receiver has concrete knowledge that would close the loop: "my current summary is X, here it is."

## Solution

Add `send_summary_back_on_rejection` in `update.rs`:

```rust
pub(crate) async fn send_summary_back_on_rejection(
    op_manager: &OpManager,
    key: &ContractKey,
    target_addr: SocketAddr,
) {
    // Fetch our current summary, send InterestMessage::Summaries to
    // just the sender. No throttle — rejection rate is already
    // bounded by upstream broadcast rate.
}
```

Wire it into all three rejection sites behind `err.is_contract_exec_rejection()`. On rejection, the sender is the one peer whose cached view of us we know is wrong — so this targets them directly rather than fanning out.

### Key design choices

- **Targets the sender only, not all interested peers.** On success, the sender is the one peer who already has our new state. On rejection, the sender is the one peer whose cache is stale about us. Opposite situations → opposite target sets.
- **No throttle.** Rejection rate is bounded by upstream broadcast rate (~80/hr on vega). A per-contract throttle would starve all but the first sender of a contract; each sender needs to learn independently.
- **Spawns via `tokio::spawn` / `GlobalExecutor::spawn`** to match the existing success-path pattern. Fire-and-forget; amplification ceiling is one task per rejection.
- **Additive, not restructuring.** Existing `ResyncRequest` flow (delta-rejection path) and `try_auto_fetch_contract` flow (missing-contract path) are unchanged.

## Testing

Structural pin tests in all three driver bodies:

- `broadcast_to_sends_summary_back_on_rejection` — task-per-tx driver
- `legacy_broadcast_to_sends_summary_back_on_rejection` — legacy `BroadcastTo`
- `legacy_broadcast_to_streaming_sends_summary_back_on_rejection` — legacy streaming variant

Each asserts that `err.is_contract_exec_rejection()` and `send_summary_back_on_rejection` both appear in the driver body, with failure messages that point to the amplification regression so future refactors can't silently drop the fix.

### Why structural pins instead of a behavioural test

Exercising the full rejection→summary-back→sender-cache-update→next-broadcast-compares-summary loop would require two `SimNetwork` nodes and several minutes of virtual time. The signal from that test relative to its maintenance cost is low: the bug is in a specific line being missing from three specific branches. Structural pins catch exactly that class of regression and run in milliseconds. The contract-handler wiring the helper touches (`get_contract_summary`, `NodeEvent::SendInterestMessage`) is already covered by existing integration tests for the success path.

### Local validation

- `cargo fmt --all` clean
- `cargo clippy --locked -p freenet --all-targets -- -D warnings` clean
- `cargo test -p freenet --lib`: **2406 passed**, 0 failed, 16 ignored

Observed one flaky preexisting failure in `operations::connect::tests::relay_does_not_accept_after_forwarding_on_retry` (passes in isolation and on retry; unrelated to this change). Will file separately per the flaky-tests rule.

## Relationship to other issues

- **#3335** (Investigate Network Amplification via UPDATE Broadcast) — open audit-style investigation. That issue enumerates seven existing protections but does not name this specific gap: the rejection path is the one place where the receiver has concrete knowledge the sender needs, and all three sites silently drop it. This PR closes one concrete subcase; the broader audit can continue.
- **#3914** (Downgrade misleading ERROR/WARN logs) — separately-filed quick-win PR. That one downgrades the log severity. This PR fixes the root cause underneath. Either can merge without the other; together they remove both the noise and the waste.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/update.rs
+++ b/crates/core/src/operations/update.rs
@@ -707,6 +707,30 @@ impl Operation for UpdateOp {
                     } = match update_result {
                         Ok(result) => result,
                         Err(err) => {
+                            // On benign stale-version rejection, nudge the sender's
+                            // peer-summary cache of us toward Some(our_summary) so
+                            // their next broadcast skips us. Helper gates internally
+                            // on summary equality + per-contract throttle to avoid
+                            // SyncStateToPeer loops and WASM amplification. See
+                            // `send_summary_back_on_rejection` docs for the full
+                            // rationale. Gated on `is_invalid_update_rejection`
+                            // (not the broader `is_contract_exec_rejection`) so
+                            // OOG/traps/validation failures don't fire summary-back.
+                            if err.is_invalid_update_rejection() {
+                                let op_mgr = op_manager.clone();
+                                let contract_key = *key;
+                                let sender_summary = sender_summary_bytes.clone();
+                                tokio::spawn(async move {
+                                    send_summary_back_on_rejection(
+                                        &op_mgr,
+                                        &contract_key,
+                                        sender_addr,
+                                        sender_summary,
+                                    )
+                                    .await;
+                                });
+                            }
+
                             if is_delta {
                                 // Delta application failed - send ResyncRequest to get full state
                                 // This is a critical debugging event for state divergence issues
@@ -1315,6 +1339,23 @@ impl Operation for UpdateOp {
                             // self-heal is needed (real failure, not a benign rejection).
                             if log_broadcast_to_streaming_failure(id, key, &err) {
                                 op_manager.try_auto_fetch_contract(key, sender_addr);
+                            } else if err.is_invalid_update_rejection() {
+                                // Benign stale-version rejection: nudge the sender's
+                                // peer-summary cache of us so the next broadcast takes
+                                // the fast-path skip. See
+                                // `send_summary_back_on_rejection` docs.
+                                let op_mgr = op_manager.clone();
+                                let contract_key = *key;
+                                let sender_summary = sender_summary_bytes.clone();
+                                tokio::spawn(async move {
+                                    send_summary_back_on_rejection(
+                                        &op_mgr,
+                                        &contract_key,
+                                        sender_addr,
+                                        sender_summary,
+                                    )
+                                    .await;
+                                });
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


sanity:
## Review feedback addressed — v2

Thanks to all 5 reviewers. Summary of changes:

### P1 (design fix — Codex + skeptical #4)

The original helper fired `InterestMessage::Summaries` unconditionally, which the receiver (`node.rs:1791-1839`) interprets on summary-mismatch as "source peer is stale" and responds with `SyncStateToPeer` — creating a reject→summary→resync→reject loop when the sender is genuinely behind us.

**Fix**: helper now accepts `sender_summary_bytes` and gates on `sender_summary == our_summary`. This narrows the fix to the exact production case (summaries equal, sender just doesn't have us in its peer-summary cache). When summaries differ, skip — the 5-min heartbeat protocol is the right convergence mechanism there.

See commit `cb885a57` for the complete reasoning.

### Also addressed

- **Log severity** (skeptical #9): `debug!` → `info!` on failed notification send (debug is stripped in release builds).
- **Pin test scope** (testing + code-first): tightened pin tests to anchor on `Err(err) =>` instead of the arm start; added `summary_back_helper_gates_on_summary_equality` pin that explicitly guards the critical equality check.
- **Spawn consistency** (big-picture): aligned all 3 sites to `tokio::spawn`.

### Not addressed (justified)

- **WASM trap classifier** (skeptical #1): the new gate naturally filters trap cases — a trapping contract's `get_contract_summary` also traps, helper returns None early. Documented.
- **Per-contract throttle** (skeptical #6): gate narrows the helper to benign cases; throttle would guard a non-issue.
- **RequestUpdate path** (code-first): different message shape, the amplification this PR fixes doesn't exist there. Out of scope.
- **Behavioural unit test** (testing): needs mock OpManager; deferred. The new helper pin catches the highest-risk regression (someone removing the gate).

### Verification

- `cargo fmt` clean
- `cargo clippy --locked -p freenet --all-targets -- -D warnings` clean
- `cargo test -p freenet --lib`: 2407 passed, 0 failed, 16 ignored

Re-running Codex + skeptical reviewer to confirm the new gate design.

[AI-assisted - Claude]

sanity:
## v3 — skeptical re-review P1 addressed

Skeptical re-review (full findings in agent transcript) confirmed the
SyncStateToPeer loop is closed by v2's equality gate, but surfaced a
new P1: the WASM `summarize_state` call runs BEFORE the gate, so
attacker-induced rejections (OOG / `MaxComputeTimeExceeded` / WASM
traps / validation failures — all match `is_contract_exec_rejection`)
force one WASM call per rejection even when the outgoing send is
ultimately suppressed by the gate.

**Fix (commit `48086d35`)**: the existing `should_send_summary_notification`
throttle (100 ms/contract, already used by the success-path sibling) is
now called as the FIRST thing in the helper, before `get_contract_summary`.
Under flood this caps summary-back at ~10 WASM calls/sec/contract
regardless of inbound rejection rate.

Updated pin test asserts `should_send_summary_notification` appears
before `get_contract_summary` in the helper body, so a future refactor
that reverts the ordering trips the test.

Verification:
- `cargo fmt` / `cargo clippy --locked -- -D warnings` clean
- `cargo test -p freenet --lib`: 2407 passed, 0 failed, 16 ignored

### Remaining v2 concerns documented in commit body, not fixed

- Concurrent-local-update race (one-shot `SyncStateToPeer`, not a loop) — minor docstring polish, leaving as follow-up
- Non-deterministic summary bytes — gate is conservatively false, correct behavior
- Timing-side-channel oracle — defendable only via constant-time behavior that defeats the fast-path
- `sender_addr` live-connection check — event-loop drops undeliverable messages; throttle caps wasted work
- Pin substring brittleness — matches existing codebase pattern

[AI-assisted - Claude]

sanity:
## v4 — rebased onto main + adopted `is_invalid_update_rejection`

PR #3915 merged while this PR was in review. Rebased onto current main and squashed the 4 incremental commits into a single clean commit (`030a0ec6`).

Key improvement from the rebase: switched the call-site gate from `is_contract_exec_rejection` to `is_invalid_update_rejection` — the stricter predicate PR #3915 added. This **structurally** closes skeptical reviewer v1 finding #1 (helper was firing on OOG / `MaxComputeTimeExceeded` / WASM traps / validation failures — all attacker-inducible). Now only the benign "new state version not higher than current" case triggers summary-back.

Final design has three safeguards, all enforced by pin tests:

1. **`is_invalid_update_rejection` gate at call sites** — benign rejections only.
2. **Throttle first** (`should_send_summary_notification`, 100 ms/contract) — caps WASM amplification under flood.
3. **Equality gate in helper** — `our_summary == sender_summary_bytes` — prevents `SyncStateToPeer` loop.

Verification:
- `cargo fmt` / `cargo clippy --locked -- -D warnings` clean
- `cargo test -p freenet --lib`: 2418 passed, 0 failed, 16 ignored

[AI-assisted - Claude]

PR #3928, opened 2026-04-20
fix(tray): render macOS menu bar icon and clean up daemon on quit

sanity:
## Problem

The Rust tray/menu-bar code has shipped in the tree since Windows
support landed and was cfg'd for `target_os = "macos"`, but on macOS
the menu bar icon never actually appeared. CI runs Linux-only, so the
broken-on-Mac path was never exercised.

Smoke-testing on macOS 15 surfaced a cascade of issues that this PR
fixes in three commits:

1. **Menu bar icon never rendered.** The macOS branch of
   `pump_platform_messages` was a no-op with a comment claiming that
   the `tray-icon` crate drove the `NSRunLoop` internally. It does
   not — `tray-icon` registers the `NSStatusItem` but leaves event
   pumping to the host application. Without an active NSApp loop on
   the main thread, the status item exists but never draws.
2. **Quit orphaned the daemon.** `tao::EventLoop::run` is divergent
   on macOS (sets `ControlFlow::Exit` ⇒ `process::exit`), so the
   wrapper thread's Quit handler (`child.kill` + `child.wait`) never
   ran. The `freenet network` child process stayed alive after the
   tray closed.
3. **UX gaps.** New users had no way to tell the menu bar icon was
   Freenet, and nothing pointed them at the local dashboard.

## Solution

- **c1 `fix(tray)`:** replace the busy-polling loop on macOS with a
  `tao::EventLoop` that wraps `NSApplication.run()` correctly, matching
  the pattern in `tray-icon`'s own published examples. Status updates
  from the wrapper thread are forwarded as tao user events so they're
  delivered on the main thread where AppKit state mutation is allowed.
  Menu IDs are now explicit string constants and dispatch is extracted
  into a pure function that's unit-tested on all platforms (Linux CI
  included). Windows path untouched.

- **c2 `feat(tray)`:** add a cleanup-done channel so the macOS event
  loop blocks on wrapper-thread shutdown (3 s timeout) before letting
  `tao` exit the process. Covers every wrapper exit path (Quit, auto-
  update restart, network-wait abort). Also adds first-run onboarding:
  on first launch the wrapper polls 127.0.0.1:7509 until the dashboard
  HTTP server accepts a connection, then opens the dashboard URL in
  the default browser and writes a marker file under
  `<data_local_dir>/freenet/.first-run-complete`. Subsequent launches
  see the marker and stay silent. Matches Mac menu-bar app conventions
  (Docker Desktop, Dropbox, Rectangle).

- **c3 `fix(tray)`:** replace the inline `with_title("Freenet")` (menu
  bar text) with a disabled "Freenet {version}" header at the top of
  the dropdown. Menu bar horizontal space is scarce; the dropdown
  header identifies the app and its version as soon as the menu opens.

## Testing

- `cargo test -p freenet --bin freenet tray::` — 2 new pure-function
  tests for `dispatch_menu_event` (known IDs map correctly, unknown
  IDs return `Unknown`).
- `cargo test -p freenet --bin freenet first_run_marker` — 1 new
  round-trip test covering the first-run marker helpers (create on
  first run, flip after write, idempotent re-write).
- `cargo fmt --all` + `cargo clippy -- -D warnings` clean.
- **End-to-end on macOS 15:** menu bar icon appears, dropdown shows
  "Freenet {version}" header, Open Dashboard opens :7509, Quit kills
  the daemon (verified with `lsof -nP -i :7509` returning empty after
  quit), first-run auto-open fires exactly once per user.

## Fixes

Unblocks the Mac installer story: with the tray known-working we can
now invest in the `.app` bundle + signed `.dmg` + release pipeline
without building on a broken UI layer.

[AI-assisted - Claude]


sanity:
Addressed findings from the four internal reviews + Codex in `4bce7262`:

**Codex P2** — 3s cleanup timeout could still orphan the daemon if Quit fired during `spawn_update_command`. Replaced with unbounded `recv()`: the wrapper-thread spawn thunk always signals cleanup-done, so a true hang only happens if the wrapper itself is stuck, in which case waiting is the right trade.

**Codex P3** — multiple first-run openers could accumulate on daemon crash-loops. Gated the spawn on a static `AtomicBool` so at most one opener thread exists per wrapper-process lifetime.

**Big-picture: Rule Lint** — fixed by rewriting `spawn_first_run_dashboard_opener` to use a local `use std::time::{Duration, Instant};` block. Same pattern as `crates/core/src/bin/freenet.rs`.

**Big-picture: missing rule** — added a "platform-gated code path" section to `.claude/rules/deployment.md` capturing the generalization: cfg'd branches CI doesn't exercise are unverified, prefer pure-function extraction, comments asserting third-party library behavior must cite.

**Testing Gap #1** — extracted `compute_menu_state` as a pure module-level function, added 4 new tests covering enable/disable states + status text uniqueness. Runs on Linux CI.

**Testing Gap #2** — added `dashboard_url_and_addr_stay_in_sync` test. One-liner that catches future drift between `DASHBOARD_URL` and `DASHBOARD_ADDR`.

**Skeptical: first-run logging** — added a success `log_wrapper_event` on the happy path so production can tell whether onboarding is firing.

Test count: 36 (was 31).

Deferred (not blocking this PR, tracked for follow-up):

- SIGTERM / Cmd-Q / Activity-Monitor-kill handler to close the remaining orphan paths
- HTTP probe instead of TCP connect for dashboard readiness
- Env var to tune the first-run deadline
- Warn-level telemetry on cleanup path

Ready for re-review. CC: @sanity

[AI-assisted - Claude]

PR #3896, opened 2026-04-17
refactor(ops): migrate relay GET to task-per-tx driver (#1454 phase 5)

iduartgomez:
## Problem

Issue #3883 tracks the final phase of #1454's async transaction refactor for GET: the relay (non-client-initiated) hop. Before this PR, any peer receiving a `GetMsg::Request` from the wire instantiates a `GetOp` in `OpManager.ops.get` and drives routing/retry/forwarding through the legacy `process_message` re-entry loop. That was the last remaining producer of `GetOp` state for non-loopback traffic.

## Solution

Dispatch split in `handle_pure_network_message_v1`'s `NetMessageV1::Get` branch:

- **Fresh inbound `GetMsg::Request`** with `source_addr.is_some()` AND no existing `GetOp` tracked (checked via new readonly `OpManager::has_get_op`) → spawn `start_relay_get`, a task-per-tx driver that owns routing, NotFound bubble-up, and Found pass-through in its task locals. No `GetOp` inserted. Fire-and-forget: the driver publishes its own upstream response.
- **Fall-through to legacy `handle_op_request`** for:
  - `source_addr.is_none()` (originator loopback from phase-3b client driver's `send_and_await(target=None)` — preserves the Request-echo contract that `drive_client_get_inner::classify` relies on for `Terminal::LocalCompletion`).
  - Existing `GetOp` (GC-spawned retries, `start_targeted_op` UPDATE-triggered auto-fetch).
  - Non-`Request` variants (streaming chunks, acks — streaming-relay migration deferred per port plan §7).

## Behavioral parity with legacy

The driver mirrors legacy relay semantics exactly:

- **No retry-at-relay.** `MAX_RELAY_RETRIES = 1`. Legacy's `process_message::Request` arm forwards once via `k_closest_potentially_hosting(k=1)` and bubbles whatever downstream returned. Cross-peer retries only happen at the originator's client driver, unchanged by this PR.
- **End-to-end tx preservation.** The relay forwards downstream with `id: incoming_tx` (matches legacy). An earlier revision minted a fresh per-attempt tx per retry iteration — that interacted with virtual-time OPERATION_TTL to produce 3^HTL concurrent subtrees (see below).
[...]

Issue #2886, opened 2026-02-07
feat: refactor peer lookups to use TransportPublicKey instead of SocketAddr


iduartgomez:
## Problem

The codebase uses `SocketAddr` (mutable, NAT-dependent) as the primary key for peer identification instead of `TransportPublicKey` (immutable, cryptographic identity). This creates address mismatch issues during NAT traversal that manifest as:

1. **Subscription registration failures**: Subscribers fail to register when their transport address differs from their advertised address
2. **Zero broadcast targets**: Update broadcasts resolve to zero peers due to address lookup failures
3. **Stale connection state**: Connection tracking is inconsistent across ring and transport layers

### Current Architecture Issues

**Two-layer address system:**
- **Ring layer** (`connections_by_location`): Indexed by advertised `PeerKeyLocation.peer_addr` (may be Unknown or internal NAT address)
- **Transport layer** (`location_for_peer`): Indexed by actual UDP source `SocketAddr` (always the real external address)

**When addresses mismatch:**
- During NAT discovery window (before `ObservedAddress` processed)
- When peers are behind NAT (internal vs external address)
- When port forwarding causes address differences

**Impact locations:**
- `crates/core/src/ring/connection_manager.rs:383-413`: `get_peer_by_addr()` requires fallback logic
- `crates/core/src/operations/subscribe.rs:603-607, 670-674`: Subscription registration needs dual address lookups
- `crates/core/src/operations/update.rs:1035-1044`: Proximity cache neighbors may fail to resolve
- `crates/core/src/node/proximity_cache.rs:48`: Stores by `SocketAddr` instead of peer identity

### Historical Context

From code archaeology (see PR #2883 review comments):
- The codebase never actually used public keys for identification
- Early `PeerId` struct (commit `605ff70c`) had `pub_key` field but only hashed/compared by `addr`
- Dec 2025 refactors (`4e4bb2ec`, `b9bb43fc`, `30854f17`) made address-based identification explicit
- `ProximityCacheManager` explicitly chose `SocketAddr` as key (commit `30854f17`)

## Proposed Solution

Refactor peer identification to use `TransportPublicKey` as the primary key with address mappings as secondary indices:

### Phase 1: Core Data Structures

1. **Update `ProximityCacheManager`** (`crates/core/src/node/proximity_cache.rs:48`):
   ```rust
   // Current:
   neighbor_caches: DashMap<SocketAddr, HashSet<ContractInstanceId>>,
   
   // Proposed:
[...]

PR #3932, opened 2026-04-21
refactor(ops): relay SUBSCRIBE task-per-tx (slice A) — #1454 phase 5 follow-up

iduartgomez:
## Problem

Fresh inbound `SubscribeMsg::Request` on relay hops still runs through
the legacy `OpManager.ops.subscribe` state-machine path. This keeps
SUBSCRIBE relay on the same surface that produced the 63 GB RSS
explosion on GET relay before PR #3896 (#1454 phase 5 GET fix).
Phase 5 slice A already migrated UPDATE (#3910) and PUT (#3917);
SUBSCRIBE is the last remaining non-streaming relay that can be
migrated before streaming pass-through work starts in slice B.

## Solution

Migrate fresh inbound `SubscribeMsg::Request` to a
task-per-transaction driver in
`operations/subscribe/op_ctx_task.rs::start_relay_subscribe`. The
driver owns local-hit response, single downstream forward via
`OpCtx::send_to_and_await`, upstream `SubscribeMsg::Response`
fire-and-forget bubble-up, and `register_downstream_subscriber` calls
on both local-hit and relayed-Subscribed paths. No `SubscribeOp` is
stored in `OpManager.ops.subscribe` for the migrated tx.

Key design choices (mirror of GET phase 5 + UPDATE / PUT slice A):

- **Per-node dedup gate**: `OpManager.active_relay_subscribe_txs`
  (DashSet) rejects duplicate inbound `SubscribeMsg::Request` for an
  in-flight tx, preventing GC-spawned re-entries or routing-bloom
  false-positive retransmissions from spawning redundant drivers.
- **RAII `RelaySubscribeInflightGuard`** constructed **before** `spawn`
  so the dedup-set entry is paired with a `Drop` even if the spawned
  future is dropped before its first poll. Closes the pre-poll
  dedup-leak window that would otherwise create a permanent GC
  blind-spot with no TTL.
- **`incoming_tx` reused end-to-end on forward** (matches legacy
  SUBSCRIBE relay). Minting a fresh tx per hop was the GET 3^HTL
  amplifier.
- **ForwardingAck omitted** for the same reason as GET / PUT / UPDATE:
  the ack carried `incoming_tx` and would satisfy the upstream's
  capacity-1 `pending_op_results` waiter before the real `Response`.
- **Cross-peer retries omitted** at the relay. Originator's client-init
  driver (Phase 2b) owns cross-peer retry. Combining relay retries with
  virtual-time `OPERATION_TTL` under `tokio::time::pause` was the
  dominant GET amplifier.
- **`send_to_and_await` for the downstream forward** (SUBSCRIBE relay
  is req/response).
- **`send_fire_and_forget` for the upstream `SubscribeMsg::Response`**:
  upstream's own `send_to_and_await` owns the capacity-1 reply
  channel.

### Relay Subscribed-response side effects

Driver calls `register_downstream_subscriber` before bubbling the
upstream Response. Deliberately omits `ring.subscribe`,
`complete_subscription_request`, and `announce_contract_hosted` on
the relayed path — a relay is not itself a subscriber (prevents the
#3763 subscription-storm feedback loop via the renewal cycle; see
`subscribe.rs:1655–1688` for the full reasoning captured when the
legacy asymmetry was fixed).

### Dispatch gate

`node.rs::handle_pure_network_message_v1` fires on
`source_addr.is_some() && !op_manager.has_subscribe_op(id) && variant
== SubscribeMsg::Request`. Originator loop-backs
(`source_addr.is_none()`), renewals, PUT sub-ops, and GC-spawned
retries that pre-register a `SubscribeOp` all fall through to legacy.
`ForwardingAck` / `Unsubscribe` never match the `Request` arm so they
flow to legacy naturally.

### Scope (slice A)

**Migrated:**

- `SubscribeMsg::Request` relay arm: has-contract fast path (with the
  existing brief `wait_for_contract_with_timeout` helper), local-hit
  Response, HTL / candidate selection, forward-and-await-Response,
  upstream bubble.

**NOT migrated (stays on legacy path):**

- Originator `SubscribeMsg::Response` handling — Phase 2b
  client-init driver already consumes it via the `pending_op_results`
  bypass.
- `SubscribeMsg::Unsubscribe` / `SubscribeMsg::ForwardingAck` — fire
  and forget cleanup, no op state needed.
- Renewals, PUT sub-op subscribes, executor auto-subscribe paths —
  no fresh `source_addr` or a pre-registered `SubscribeOp` make them
  fall through naturally.

### Per-node memory-stats counters

Plumbed into `FREENET_MEMORY_STATS`:
`relay_subscribe_inflight`, `relay_subscribe_spawned`,
`relay_subscribe_completed`, `relay_subscribe_dedup_rejects`,
`relay_subscribe_active_txs`.

### Structural pin tests (`subscribe/op_ctx_task.rs`)

- Dedup gate ordering: `active_relay_subscribe_txs.insert` before
  `GlobalExecutor::spawn`.
- Dedup rejection increments `RELAY_SUBSCRIBE_DEDUP_REJECTS`.
- RAII guard clears dedup set + decrements counters on drop.
- Driver forwards via `send_to_and_await` (not fire-and-forget).
- Driver reuses `incoming_tx` on forward (no fresh tx mint).
- Upstream reply uses `send_fire_and_forget`.
- Driver does NOT construct a `ForwardingAck`.
- Driver calls `register_downstream_subscriber` on BOTH local-hit and
  relayed-Subscribed paths.
- Driver does NOT install a lease on relayed Subscribed (no
  `ring.subscribe` / `complete_subscription_request` /
  `announce_contract_hosted`).

### Dispatch pin tests (`node.rs`)

- SUBSCRIBE branch invokes `start_relay_subscribe`.
- Gated on `source_addr` + `has_subscribe_op`.
- Only the `Request` variant is destructured for relay dispatch.

### Other testing

- [x] `cargo test -p freenet --lib` — 2440/2440 pass.
- [x] `cargo test -p freenet --features simulation_tests,testing
      --test simulation_integration -- test_subscribe test_subscription
      test_subscription_broadcast_propagation
      test_subscription_relay_propagation` — all pass.
- [x] `cargo test -p freenet --features simulation_tests,testing
      --test operations -- subscribe subscription` — 9/9 pass via
      Docker test-runner (macOS cannot bind the full 127.x.x.x range).
- [ ] CI ci-fault-loss sim run (will trigger on PR push): confirm no
      memory regression vs. main. Baseline: main 2.6 GB RSS under
      ci-fault-loss after the GET-relay phase-5 fix.

## Fixes

Refs #1454.

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -41,8 +41,22 @@ paths:
 > `source_addr.is_some()` AND `!has_put_op(id)` AND the payload size
 > would not upgrade to streaming on forward); GC speculative retries,
 > client-initiated loopback (`source_addr.is_none()`), and streaming
-> variants still go through legacy. Slices B (streaming variants) and
-> C (legacy arm cleanup) follow.
+> variants still go through legacy. Phase 5 follow-up slice A (PR
+> #3932) migrated fresh inbound relay `SubscribeMsg::Request` to
+> `operations/subscribe/op_ctx_task.rs::start_relay_subscribe`
+> (gated on `source_addr.is_some()` AND `!has_subscribe_op(id)`);
+> renewals, PUT sub-op subscribes, executor auto-subscribe,
+> GC-spawned retries that pre-register a `SubscribeOp`,
+> `SubscribeMsg::Unsubscribe`, and `SubscribeMsg::ForwardingAck` all
+> stay on legacy. The driver caps relay-hop retries at 1 (originator
+> owns cross-peer retry), reuses `incoming_tx` end-to-end, omits
+> `ForwardingAck` emission, and calls `register_downstream_subscriber`
+> on BOTH local-hit and relayed-Subscribed paths — but DELIBERATELY
+> omits `ring.subscribe` / `complete_subscription_request` /
+> `announce_contract_hosted` on the relayed path (relay is not itself
+> a subscriber; prevents the #3763 subscription-storm feedback loop).
+> Slices B (streaming variants for GET/PUT/UPDATE) and C (legacy arm
+> cleanup) follow.
 >
 > Task-per-tx drivers have their own invariants documented in the
 > `op_ctx_task.rs` module doc and in `OpCtx::send_and_await`'s rustdoc.
--- a/crates/core/CLAUDE.md
+++ b/crates/core/CLAUDE.md
@@ -53,22 +53,29 @@ Each file is a state machine:
 
 Plus task-per-transaction drivers from #1454 Phase 2b onwards:
   subscribe/op_ctx_task.rs → client-initiated SUBSCRIBE driver
+                             + fresh inbound relay SUBSCRIBE driver
+                             (Phase 5 follow-up slice A, `start_relay_subscribe`)
   put/op_ctx_task.rs       → client-initiated PUT driver (Phase 3a)
+                             + fresh inbound non-streaming relay PUT
+                             driver (Phase 5 follow-up slice A, PR #3917,
+                             `start_relay_put`)
   get/op_ctx_task.rs       → client-initiated GET driver (Phase 3b)
                              + fresh inbound relay GET driver (Phase 5,
                              PR #3896, `start_relay_get`)
   update/op_ctx_task.rs    → client-initiated UPDATE driver (Phase 4,
                              fire-and-forget) + fresh inbound non-streaming
                              relay UPDATE drivers (Phase 5 follow-up slice
-                             A: `start_relay_request_update`,
+                             A, PR #3910: `start_relay_request_update`,
                              `start_relay_broadcast_to`)
-    Client + relay GET / UPDATE share the per-node dedup gate
-    (`active_relay_get_txs`, `active_relay_update_txs`) and the
-    `Relay*InflightGuard` RAII pattern. Streaming UPDATE relay variants
+    All four relay drivers share the per-node dedup gate pattern
+    (`active_relay_{get,update,put,subscribe}_txs`) and the
+    `Relay*InflightGuard` RAII. Streaming UPDATE relay variants
     (`RequestUpdateStreaming`, `BroadcastToStreaming`) and the
-    deprecated `Broadcasting` variant remain on the legacy path
-    pending slice B. PUT and SUBSCRIBE relay paths still go through
-    the legacy state machine.
+    deprecated `Broadcasting` wire variant, plus PUT streaming
+    variants, remain on legacy pending slice B. SUBSCRIBE has no
+    streaming variants but renewals, PUT sub-op subscribes, executor
+    auto-subscribe paths, `Unsubscribe`, and `ForwardingAck` all stay
+    on the legacy state machine.
 ```
 
 ## Module Map
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -489,6 +489,8 @@ fn launch_agent_plist_has_expected_leader(plist: &Path, expected_leader: &str) -
 /// never loaded it" breakages are diagnosable after the fact.
 #[cfg(target_os = "macos")]
 fn launchctl_bootstrap(plist: &Path) {
+    // SAFETY: libc::getuid is always safe; it returns the current user
+    // ID without touching user-provided pointers.
     let uid = unsafe { libc::getuid() };
     let target = format!("gui/{uid}");
     match std::process::Command::new("launchctl")
@@ -517,6 +519,8 @@ fn launchctl_bootstrap(plist: &Path) {
 
 #[cfg(target_os = "macos")]
 fn launchctl_bootout(plist: &Path) {
+    // SAFETY: libc::getuid is always safe; it returns the current user
+    // ID without touching user-provided pointers.
     let uid = unsafe { libc::getuid() };
     let target = format!("gui/{uid}");
     match std::process::Command::new("launchctl")
@@ -1012,7 +1016,7 @@ fn run_wrapper(version: &str) -> Result<()> {
         // Wrapper loop runs on a background thread
         let loop_handle = std::thread::spawn(move || {
             let result = run_wrapper_loop(&log_dir_clone, Some((&action_rx, &status_tx)));
-            let _ = cleanup_tx.send(());
+            cleanup_tx.send(()).ok();
             result
         });
 
--- a/crates/core/src/bin/commands/tray.rs
+++ b/crates/core/src/bin/commands/tray.rs
@@ -381,10 +381,12 @@ mod platform {
             let s = compute_menu_state(status);
             self.status_item
                 .set_text(format!("Status: {}", s.status_text));
-            let _ = self._tray.set_tooltip(Some(format!(
-                "Freenet {} - {}",
-                self.version, s.status_text
-            )));
+            self._tray
+                .set_tooltip(Some(format!(
+                    "Freenet {} - {}",
+                    self.version, s.status_text
+                )))
+                .ok();
             self.stop_item.set_enabled(s.stop_enabled);
             self.start_item.set_enabled(s.start_enabled);
             self.restart_item.set_enabled(s.restart_enabled);
@@ -567,7 +569,7 @@ mod platform {
                         // long `freenet update` that the user triggered via
                         // "Check for Updates"). Waiting is the right trade
                         // because the alternative is a guaranteed orphan.
-                        let _ = cleanup_done_rx.recv();
+                        cleanup_done_rx.recv().ok();
                         *control_flow = ControlFlow::Exit;
                         return;
                     }
@@ -582,7 +584,7 @@ mod platform {
                     std::thread::sleep(Duration::from_secs(1));
                     action_tx.send(TrayAction::Quit).ok();
                     // Same rationale as the Quit handler above.
-                    let _ = cleanup_done_rx.recv();
+                    cleanup_done_rx.recv().ok();
                     *control_flow = ControlFlow::Exit;
                 }
             }
--- a/crates/core/src/bin/commands/update.rs
+++ b/crates/core/src/bin/commands/update.rs
@@ -1413,6 +1413,9 @@ fn acquire_update_lock() -> Result<UpdateLock> {
         .context("Failed to open update lockfile")?;
     // flock with LOCK_EX | LOCK_NB: fail fast instead of queueing. A
     // parallel updater should bail rather than serialize behind us.
+    // SAFETY: `file.as_raw_fd()` returns an owned, open file descriptor
+    // for the duration of this function; libc::flock only consults the
+    // fd and the constant flag bits, with no pointer arguments.
     let rc = unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_EX | libc::LOCK_NB) };
     if rc != 0 {
         anyhow::bail!(
@@ -1440,7 +1443,7 @@ struct MountDetachOnDrop {
 #[cfg(target_os = "macos")]
 impl Drop for MountDetachOnDrop {
     fn drop(&mut self) {
-        let _ = hdiutil_detach(&self.mount_point);
+        hdiutil_detach(&self.mount_point).ok();
     }
 }
 
@@ -1521,6 +1524,12 @@ fn spawn_detached_updater(script: &Path, current_app: &Path, staged_app: &Path) 
     // `setpgid(0, 0)` is redundant but cheap belt-and-braces. Without
     // `setsid`, a terminal-parent session delivers SIGHUP to all
     // descendants on close, killing the updater mid-swap (skeptical M5).
+    //
+    // SAFETY: `pre_exec` is unsafe because the closure runs between
+    // fork and exec, where only async-signal-safe functions are
+    // allowed. `libc::setsid` and `libc::setpgid(0, 0)` are both
+    // async-signal-safe per POSIX and do not allocate or touch
+    // user-provided pointers.
     unsafe {
         cmd.pre_exec(|| {
             if libc::setsid() == -1 {
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -110,6 +110,15 @@ pub mod dev_tool {
     #[cfg(any(test, feature = "testing"))]
     pub use crate::operations::put::op_ctx_task::RELAY_PUT_DRIVER_CALL_COUNT;
 
+    // #1454 Phase 5 follow-up slice A (#3932) — test hook to verify
+    // the dispatch gate in `handle_pure_network_message_v1` actually
+    // routes fresh inbound relay SUBSCRIBEs through the task-per-tx
+    // driver (vs. the legacy `handle_op_request` fallthrough used
+    // for renewals, PUT sub-op subscribes, executor auto-subscribe,
+    // and GC-spawned retries).
+    #[cfg(any(test, feature = "testing"))]
+    pub use crate::operations::subscribe::op_ctx_task::RELAY_SUBSCRIBE_DRIVER_CALL_COUNT;
+
     // Re-export state verification for telemetry-based consistency analysis
     pub use crate::tracing::state_verifier::{StateAnomaly, StateVerifier, VerificationReport};
 
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1460,27 +1460,53 @@ where
                     is_renewal,
                 } = op
                 {
-                    if let Some(upstream_addr) = source_addr {
-                        if !op_manager.has_subscribe_op(id) {
-                            if let Err(err) = subscribe::op_ctx_task::start_relay_subscribe(
-                                op_manager.clone(),
-                                *id,
-                                *instance_id,
-                                *htl,
-                                visited.clone(),
-                                *is_renewal,
-                                upstream_addr,
-                            )
-                            .await
-                            {
-                                tracing::error!(
-                                    tx = %id,
-                                    %instance_id,
-                                    error = %err,
-                                    "SUBSCRIBE relay dispatch: start_relay_subscribe failed"
-                                );
+                    // Renewals (`is_renewal=true`) fall through to legacy
+                    // so the relay still emits `SubscribeMsg::ForwardingAck`
+                    // back to the renewal originator. The
+                    // `ring.rs::connection_maintenance` renewal loop
+                    // registers a `SubscribeOp` locally, and its GC retry
+                    // selector at `op_state_manager.rs:2094` uses
+                    // `ack_received` to pick the retry base — without the
+                    // ack the originator retries on `ACK_TIMEOUT*(n+1)`
+                    // instead of `PROGRESS_TIMEOUT + ACK_TIMEOUT*n`,
+                    // which over the ~2 min renewal cycle compounds into
+                    // the #3763 subscription-storm feedback loop.
+                    // Driver-side ack emission would race the reply on
+                    // the upstream's capacity-1 `pending_op_results` slot
+                    // for task-per-tx originators, so slice A keeps
+                    // renewals on legacy.
+                    //
+                    // Non-renewal SUBSCRIBEs come from: (a) the Phase 2b
+                    // client-init driver, which uses `pending_op_results`
+                    // and would classify a `ForwardingAck` as
+                    // `ReplyClass::Unexpected` anyway; (b) PUT sub-op
+                    // subscribes and parent-op child subscribes —
+                    // one-shot paths with bounded `MAX_RETRIES` so the
+                    // ACK-timer shortening is an accepted slice-A
+                    // amplification ceiling.
+                    if !*is_renewal {
+                        if let Some(upstream_addr) = source_addr {
+                            if !op_manager.has_subscribe_op(id) {
+                                if let Err(err) = subscribe::op_ctx_task::start_relay_subscribe(
+                                    op_manager.clone(),
+                                    *id,
+                                    *instance_id,
+                                    *htl,
+                                    visited.clone(),
+                                    *is_renewal,
+                                    upstream_addr,
+                                )
+                                .await
+                                {
+                                    tracing::error!(
+                                        tx = %id,
+                                        %instance_id,
+                                        error = %err,
+                                        "SUBSCRIBE relay dispatch: start_relay_subscribe failed"
+                                    );
+                                }
+                                return Ok(None);
                             }
-                            return Ok(None);
                         }
                     }
                 }
@@ -3976,6 +4002,29 @@ mod tests {
                 "SUBSCRIBE relay dispatch must be guarded by has_subscribe_op — \
                  renewals, PUT sub-ops, and GC-spawned retries that pre-register \
                  a SubscribeOp must fall through to legacy."
+            );
+        }
+
+        #[test]
+        fn subscribe_branch_falls_through_on_renewal() {
+            const SOURCE: &str = include_str!("node.rs");
+            let anchor = "NetMessageV1::Subscribe(ref op) => {";
+            let branch_start = SOURCE.find(anchor).expect("SUBSCRIBE branch not found");
+            let window_end = SOURCE[branch_start..]
+                .find("handle_op_request::<subscribe::SubscribeOp, _>")
+                .expect("SUBSCRIBE branch no longer calls handle_op_request")
+                + branch_start;
+            let window = &SOURCE[branch_start..window_end];
+            // Renewals MUST stay on legacy — the renewal originator's GC
+            // retry selector uses `ack_received` and the driver does not
+            // emit `SubscribeMsg::ForwardingAck`. Falling through to
+            // legacy keeps the ack emission alive for renewals.
+            assert!(
+                window.contains("if !*is_renewal"),
+                "SUBSCRIBE relay dispatch must skip `is_renewal=true` \
+                 requests so renewals still receive `ForwardingAck` from \
+                 the relay. Dropping the ack on the ~2 min renewal cycle \
+                 reopens the #3763 subscription-storm feedback loop."
             );
         }
 
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1429,6 +1429,62 @@ where
                     return Ok(None);
                 }
 
+                // #1454 phase 5 follow-up (slice A): relay SUBSCRIBE
+                // task-per-tx dispatch.
+                //
+                // For true relay hops (source_addr.is_some() AND incoming
+                // variant is SubscribeMsg::Request AND no SubscribeOp
+                // already exists in OpManager for this tx), spawn the
+                // relay driver and return. The driver owns local-hit
+                // response + single downstream forward + upstream
+                // bubble-up in its task locals.
+                //
+                // source_addr.is_none() (originator loop-back from the
+                // phase-2b client-init driver's send_to_and_await) falls
+                // through to legacy. The client driver never relies on
+                // a Request-echo — it awaits Response — so the
+                // `SubscribeMsg::Response` bypass above handles the
+                // originator reply.
+                //
+                // Existing SubscribeOp means a renewal, PUT sub-op, or
+                // GC-spawned retry already registered state in the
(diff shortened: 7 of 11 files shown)


iduartgomez (commented):
## Testing Review: PR #3932 — relay SUBSCRIBE slice A

### Coverage Summary

- Direct changes (structural invariants): covered by 13 pin tests.
- Direct changes (runtime behavior of the driver): missing.
- Downstream impact (UPDATE fan-out via relay subscriber registration): covered only indirectly by the pre-existing `test_subscription_relay_propagation`, which tests the legacy path semantics and will continue to pass whether or not the new driver runs.
- Memory-regression signal (ci-fault-loss): deferred to post-push CI, with no in-tree functional assertion.

### Test Gaps Found

#### Critical (blocks merge)

1. **No behavioral proof the dispatch gate actually routes through the new driver.** PR #3917 (PUT slice A) added this exact companion in `test_get_routing_coverage_low_htl` (see line 7565–7574 of `crates/core/tests/simulation_integration.rs`) asserting `RELAY_PUT_DRIVER_CALL_COUNT` advances. Without the analogous `RELAY_SUBSCRIBE_DRIVER_CALL_COUNT` assertion in a sim test that exercises a relayed subscribe (e.g. `test_subscription_relay_propagation` at line 3895, or the low-HTL coverage test above), the driver could be entirely dead code — dispatch gate regressed, everything falls through to legacy — and all 13 pin tests plus `test_subscription_relay_propagation` would still pass. This is the single most important missing test.

2. **`RELAY_SUBSCRIBE_DRIVER_CALL_COUNT` is declared in `op_ctx_task.rs:781` but NOT re-exported via `dev_tool` in `crates/core/src/lib.rs`.** Compare lines 103 (`GET_RELAY_DRIVER_CALL_COUNT`) and 111 (`RELAY_PUT_DRIVER_CALL_COUNT`). Without this re-export, simulation tests cannot read the counter even if someone writes the companion test. Add the re-export in the same PR.

#### Important (should have)

3. **No runtime test that the UPDATE fan-out reaches the original requester through the relay driver.** Two pins confirm `register_downstream_subscriber` is called on both local-hit and Subscribed-bubble paths, but nothing asserts the side effect works end-to-end on the new code path. `test_subscription_relay_propagation` was written against the legacy path; its `>= 3 peers have state` assertion succeeds even if the relay driver's `register_downstream_subscriber` call is a no-op, because the legacy fallthrough (e.g., renewals, executor path) also registers. Either add a variant that disables the legacy fallthrough or gate the assertion on `RELAY_SUBSCRIBE_DRIVER_CALL_COUNT > 0`.

4. **No functional test of `drive_relay_subscribe` failure paths.** The structural pins prove each of HTL=0, empty-candidates, send-failure, timeout-elapsed, and unexpected-reply converts to `SubscribeMsgResult::NotFound`, but no runtime test exercises these. At least one sim test (or unit test driving the fn with a mock `OpManager`) for the HTL=0 → NotFound and timeout → NotFound cases would catch refactoring regressions that the text-matching pins cannot (e.g., someone restructures the match and the NotFound branch stops firing).

5. **No test for dedup rejection at runtime.** `dedup_rejection_increments_counter` is a source-scrape test. A unit test that calls `start_relay_subscribe` twice with the same `incoming_tx` and asserts `RELAY_SUBSCRIBE_DEDUP_REJECTS` advanced by exactly 1 and only one driver was spawned would cost ~15 lines and actually exercise the DashSet.

#### Nice to Have

6. **RAII guard pre-poll drop test.** `raii_guard_clears_dedup_set_on_drop` confirms the impl calls the right clear sites via string match. A runtime test that constructs a `RelaySubscribeInflightGuard`, drops it without polling, and asserts the set is empty and counters returned to baseline would make the "drop before first poll" window (the #3896 bug class) a first-class assertion.

### CI Considerations

- The ci-fault-loss baseline (2.6 GB RSS after the GET-relay phase-5 fix) is mentioned in the PR body as a pending checkbox. For the sibling GET PR #3896 this regression was caught **post-merge** at 63 GB RSS. We should not rely on post-merge discovery a second time. Options:
  - Add a small-scale ci-fast sim with many parallel relayed subscribes that asserts `RELAY_SUBSCRIBE_INFLIGHT` stays bounded under sustained load. This is the cheap signal equivalent of the memory-explosion test.
  - Block merge on the ci-fault-loss run rather than treating it as advisory.
- The 13 pin tests are all fast string scans — no CI cost concern there.

### Summary

The pin-test-only pattern from slices A (#3910, #3917) plus a single behavioral companion (`RELAY_PUT_DRIVER_CALL_COUNT > 0` in the existing `test_get_routing_coverage_low_htl`) is the precedent the SUBSCRIBE PR should match. Right now this PR is missing the behavioral companion and the `dev_tool` re-export that enables it. That gap plus the absence of any memory-regression signal in-tree are the two items that would have caught the #3896 63 GB incident before merge.

iduartgomez (commented):
## Code-First Review: PR #3932

Performed with code-only first pass before reading description, comparing the new driver (`operations/subscribe/op_ctx_task.rs::start_relay_subscribe` and helpers) against the legacy Request arm (`subscribe.rs:1369-1626`) and the legacy relay Subscribed arm (`subscribe.rs:1629-1711`).

### Parity check: side effects replicated correctly

- **Local-hit fast path**: `wait_for_contract_with_timeout` at the top collapses the legacy two-stage check (`has_contract` → then wait) into one call. `wait_for_contract_with_timeout` internally fast-paths on `has_contract`, so net behavior matches. `register_downstream_subscriber` fires on local-hit before the Response is bubbled — parity with legacy `subscribe.rs:1395–1404` and `1447–1456`.
- **HTL==0 telemetry**: `NetEventLog::subscribe_not_found` with `Some(max_hops_to_live)` — matches legacy `:1488-1494`.
- **No candidates telemetry**: `NetEventLog::subscribe_not_found` with `None` — matches legacy `:1524-1530`.
- **Visited-peers handling**: `with_transaction(&incoming_tx)` + mark self + mark upstream — matches legacy except legacy only marks `requester` when `requester_addr.is_some()`; driver is relay-only so always-mark is correct.
- **Forward telemetry**: `NetEventLog::subscribe_request` — matches legacy `:1570-1578`.
- **Relay Subscribed-response handling**: `register_downstream_subscriber` on bubble, but deliberately omits `ring.subscribe` / `record_subscription` / `announce_contract_hosted` — matches the legacy asymmetry fix at `:1655-1688` (prevents #3763 subscription-storm feedback loop).
- **Incoming tx reused on forward**: matches legacy (no fresh tx mint — the GET 3^HTL amplifier is avoided).

### Discrepancies

1. **Missing `SubscribeStats` / `PeerHealthTracker` / failure-estimator signal.** Legacy `subscribe.rs:1617-1621` attaches `SubscribeStats { target_peer, contract_location, request_sent_at: Instant::now() }` to the `AwaitingResponse` op so that downstream timeouts are reported to `PeerHealthTracker` and the failure estimator (see the comment at `:1615-1616`). The driver's downstream timeout at `op_ctx_task.rs:1147-1164` only logs and bubbles `NotFound` upstream — **no `PeerHealthTracker` update**. This removes a signal that the relay's forward target is slow/dead. This is not mentioned in the PR description. Worth either (a) replicating the stat update in the timeout branch, or (b) documenting explicitly in the PR that relay-side peer-health reporting is dropped because originator-side retries now own that signal.

2. **No `ForwardingAck` sent to the upstream/requester.** Legacy `:1583-1589` sends `SubscribeMsg::ForwardingAck` to `requester_addr` BEFORE forwarding so the requester's GC task can distinguish "dead peer" from "slow multi-hop". The driver omits this. PR body explains it as the reply-channel-capacity issue, but the original purpose of ForwardingAck (GC heuristic on the requester side) is now lost for any chain where even one relay hop uses the new driver. Concrete impact on the originator's GC timing deserves a sentence in the description — right now it's only framed as a downstream-of-capacity-1 issue.

3. **`register_downstream_subscriber` `source_addr` arg on Response bubble.** Driver passes `Some(upstream_addr)` (`:1183`). Legacy `:1695` passes `None`. Since `requester_addr` and `upstream_addr` are the same peer on a relay, this is a wasteful fallback lookup (same addr searched twice) rather than a bug. Minor.

4. **Renewal through a relay still takes the new driver.** PR body says "Renewals… fall through to legacy". That's true on the originator where `ring.rs::contracts_needing_renewal` pre-registers a `SubscribeOp`, but at relay nodes no pre-registration exists and `is_renewal=true` inbound Requests will spawn the driver. Semantically fine (relay processing is identical), but the PR body wording is easy to misread.

5. **`requester_pub_key` always `None` on driver calls.** Legacy passes `self.requester_pub_key` (resolved at op init). Driver passes `None`, relying on `register_downstream_subscriber`'s addr-fallback. Comment at `:960-965` acknowledges this. Fine at relay hops with direct upstream, but worth keeping in mind for any future NAT-timing-window edge case (the legacy comment at `:1393-1394` explicitly called out #2886 — the pub-key path was the fix). If addr-lookup ever fails here, a silent miss on downstream registration is the failure mode (the `warn!` branch at `register_downstream_subscriber:1236-1243`).

### Concerns

- **Double `release_pending_op_slot`**: called once at `:1126` (after `send_to_and_await`) and once at `:930` (in `run_relay_subscribe` after driver exit). Idempotent, but the first is load-bearing (before the upstream fire-and-forget at line 1220) and the second is a safety net. A one-liner comment on the first call noting "must release before the fire-and-forget bubble-up at `relay_subscribe_send_response`" would prevent someone from later deleting it as redundant.
- **`RELAY_SUBSCRIBE_DRIVER_CALL_COUNT` counts dedup rejects.** Bumped (at `:827`) BEFORE the dedup check. The docstring says "number of times `start_relay_subscribe` was invoked", so this matches, but analytical code that uses it as a proxy for "drivers spawned" will be slightly off. The dedicated `RELAY_SUBSCRIBE_SPAWNED_TOTAL` is the right one for that — callers just need to know.
- **Guard drop ordering in `run_relay_subscribe`**: `let _guard = guard;` followed by `drive_...await.await` followed by `release_pending_op_slot`. The guard drops last (at scope end, after `release_pending_op_slot`). Order is correct — pending slot released first, then dedup entry removed. No reordering risk, but again one line of comment explaining the ordering invariant would help future maintainers.

### Documentation Recommendations

- Add one sentence to the PR body covering the dropped `PeerHealthTracker` / `SubscribeStats` signal at the relay (discrepancy #1).
- Clarify in the PR body that the `ForwardingAck` omission affects the originator's GC "slow chain" heuristic, not just the reply-channel-capacity problem.
- In `drive_relay_subscribe` rustdoc, list "peer-health reporting on downstream timeout is deliberately dropped because the originator driver owns cross-peer retry now" if that's the design intent.
- In the `release_pending_op_slot` at `:1126`, add a one-line comment about why BOTH that release and the one in `run_relay_subscribe` exist.

### Overall

Structural pin tests are thorough and mirror the GET/PUT/UPDATE slice-A pattern faithfully. Dispatch gate is correctly scoped. Main substantive concern is the silently dropped `PeerHealthTracker` signal on downstream forward timeouts — worth either restoring or documenting. Everything else is cosmetic / documentation.

iduartgomez (commented):
## Big Picture Review: PR #3932 (SUBSCRIBE slice A)

### Goal Alignment — all four amplifiers mitigated as described

Verified against the source (not just the PR body):

- **Per-node dedup gate** — `op_manager.active_relay_subscribe_txs.insert(incoming_tx)` runs before `GlobalExecutor::spawn(run_relay_subscribe(...))` (`op_ctx_task.rs:829`, 907) with structural pin test `start_relay_subscribe_checks_dedup_gate`.
- **`incoming_tx` reused on forward** — forwarded `SubscribeMsg::Request { id: incoming_tx, … }` at `op_ctx_task.rs:1112`, no fresh `Transaction::new::<SubscribeMsg>()` anywhere in the relay path (only one in the module-level client-driver at line 281).
- **ForwardingAck omitted** — no `SubscribeMsg::ForwardingAck { … }` construction in the relay driver; pin test in `node.rs` additionally forbids destructuring the variant in dispatch.
- **Cross-peer retries omitted** — `run_relay_subscribe` forwards once via `send_to_and_await`, bubbles the reply via `send_fire_and_forget`; no retry loop / `Transaction::new` per hop.

`register_downstream_subscriber` runs on both local-hit (`op_ctx_task.rs:966`) and relayed-Subscribed (1178) paths, pinned by `relay_driver_calls_register_downstream_subscriber_on_both_paths`. Deliberate omissions (`ring.subscribe`, `complete_subscription_request`, `announce_contract_hosted`) are pinned by `relay_driver_omits_subscriber_lease_on_relayed_response` — matches the #3763 asymmetry fix.

`release_pending_op_slot(incoming_tx)` is called after `send_to_and_await` returns (1126) — same slot-leak fix as the PUT slice A follow-up `06a18b10`. Good.

### Removed Code Concerns — none

No tests deleted or weakened. The two `#[ignore]` in `subscribe/tests.rs` (lines 918, 1291) pre-date this branch. No assertions relaxed.

### Anti-Patterns Detected — none

No `#[ignore]`, no `.unwrap_or_default()`, no TODO/FIXME deferrals, no commented-out code on the migration paths. `cleanup_tx.send(()).ok()` / `hdiutil_detach(...).ok()` in commit `47c3595c` are intentional clippy fixes (`let-underscore-must-use`), not error swallowing of new code.

### Documentation Drift — ACTION REQUIRED

Established slice-A precedent (PR #3910 and PR #3917) updated `.claude/rules/operations.md` in the same PR; #3910 additionally updated `crates/core/CLAUDE.md`. This PR updates neither. Concrete stale passages:

1. **`.claude/rules/operations.md:29–31`** — still says "SUBSCRIBE's renewal / PUT-sub-op / executor / intermediate-peer entry points" as legacy while listing PUT/UPDATE slice A migrations alongside. After #3932, fresh inbound relay SUBSCRIBE is migrated; the legacy surface for SUBSCRIBE is narrower. Add a sibling sentence to the PUT/UPDATE slice A paragraphs describing `start_relay_subscribe` gated on `source_addr.is_some() && !has_subscribe_op(id) && variant == SubscribeMsg::Request`, with the same "GC/renewal/PUT-sub-op/executor + Unsubscribe/ForwardingAck stay on legacy" qualifier.

2. **`crates/core/CLAUDE.md:54–72`** — the "Plus task-per-transaction drivers" table lists `subscribe/op_ctx_task.rs → client-initiated SUBSCRIBE driver` only, and the trailing paragraph asserts "PUT and SUBSCRIBE relay paths still go through the legacy state machine" (already outdated for PUT after #3917; definitely outdated for SUBSCRIBE after this PR). Add `start_relay_subscribe` to the subscribe entry and rewrite the trailing sentence to read "SUBSCRIBE streaming is N/A (no streaming variants); renewal / Unsubscribe / ForwardingAck / executor / PUT-sub-op entries remain on legacy."

3. **`docs/architecture/operations/README.md`** — no mentions of task-per-tx drivers or Phase 5 at all. Not a regression introduced here, but worth flagging: after slice A is complete across GET/PUT/UPDATE/SUBSCRIBE, `docs/architecture/operations/README.md:168+ (SUBSCRIBE Operation)` and `:448 (operations/subscribe.rs)` should reference `op_ctx_task.rs` as the authoritative fresh-inbound path. Low urgency; can be rolled up into the Phase 6 rewrite.

4. **Issue #1454 status comment** — worth posting a "phase 5 follow-up slice A complete" comment once CI ci-fault-loss RSS confirmation lands; not a blocker for review.

### Scope Creep — commit `47c3595c`

The `bin/commands/*.rs` clippy fix IS orthogonal. The commit message states it unblocks the pre-commit hook on any branch touching `freenet`, which is pragmatic. Precedent PR #3917 did NOT bundle infra fixes — its file list is purely the migration + ops rule. Strict reading: this should have been a separate `fix(bin):` PR merged first. Pragmatic reading: 24 lines, zero behaviour change, explicitly annotated in its own commit. Acceptable but noted; prefer splitting in future for bisect hygiene.

### Phased-Rollout Consistency

SUBSCRIBE has no streaming variants, so slice B (streaming pass-through) is N/A for SUBSCRIBE. The legacy SUBSCRIBE surface after this PR is: originator `Response` (Phase 2b driver consumes via `pending_op_results`), `Unsubscribe` (fire-and-forget), `ForwardingAck` (dead code once driver omits emitting them — but incoming acks from non-migrated peers still must flow), renewals, PUT sub-op subscribes, executor auto-subscribe, GC-spawned retries. The `OpManager.ops.subscribe` DashMap is STILL required for: renewals (register via `new_op`), PUT sub-ops, executor auto-subscribe, and the GC-retry path — none migrated. So Phase 5-final DashMap deletion is **NOT unblocked** for SUBSCRIBE by this PR; slice C (legacy arm cleanup) or a subsequent slice must migrate renewals + PUT-sub-op + executor entry points first. That's consistent with the stated scope — no gap vs. claims, just worth being explicit.

### Agent/Skill Instruction Rot

No agents under `agents/*.md` or skills referencing subscribe relay-path specifics by name found beyond `.claude/rules/operations.md` and `crates/core/CLAUDE.md` (covered above).

### Recommendations Before Merge

1. Add `.claude/rules/operations.md` update mirroring the slice-A paragraphs for PUT/UPDATE (one additional paragraph for SUBSCRIBE + `start_relay_subscribe` gate).
2. Update `crates/core/CLAUDE.md` driver table entry and the trailing "legacy" sentence.
3. Optional: split commit `47c3595c` into a separate PR (not a blocker — call the author's judgement).
4. Await ci-fault-loss memory baseline confirmation (already an unchecked box in the PR body — the one acknowledged open item).

### Strategic Assessment

This PR completes the non-streaming relay migration for SUBSCRIBE with the exact shape established by #3896 / #3910 / #3917, and the structural pin tests will catch any regression where the gate is accidentally removed or a future refactor re-introduces the amplifier patterns. The only outstanding concern is keeping the agent-facing documentation (`operations.md`, `CLAUDE.md`) in sync so future agent work doesn't mis-route a SUBSCRIBE relay change back onto the legacy path.

iduartgomez (commented):
## Skeptical Review: PR #3932

### High Severity

**H1. ForwardingAck omission causes subscription lease amplification on legacy-originator paths.**

The driver deliberately omits `SubscribeMsg::ForwardingAck` (rationale: capacity-1 `pending_op_results` race, same as GET/PUT/UPDATE slices). That reasoning holds for task-per-tx ORIGINATORS. It does NOT hold for LEGACY ORIGINATORS that still exist in this codebase:

- `ring.rs:1054` — ring renewal cycle (`is_renewal=true`), fires every ~2 minutes per active subscription
- `executor.rs:723, 730` — PUT sub-op subscribe
- `operations.rs:784` — parent-op child SUBSCRIBE

These paths push a `SubscribeOp` into `OpManager.ops.subscribe` and the GC loop at `op_state_manager.rs:2079` explicitly depends on ForwardingAck arrival to pick the retry base:

```
let base = if sub_op.ack_received {
    PROGRESS_TIMEOUT + ACK_TIMEOUT * (retry_count as u32)   // longer
} else {
    ACK_TIMEOUT * (retry_count as u32 + 1)                  // shorter
};
```

Without ForwardingAck the legacy originator thinks "no response, dead chain" and triggers speculative retry early. Each retry routes through fresh relays that again don't ACK, fanning out further. This is precisely the #3763 subscription-storm feedback loop the `register_downstream_subscriber`-only design was introduced to prevent, now re-introduced on the renewal side. The renewal cycle is the highest-volume caller — every ~2 min × every active subscription × (now) amplified retries.

Suggested fix options:
1. Send `ForwardingAck` after `send_to_and_await` returns (so it can't race the Response through pending_op_results, because the Response has already been consumed). A non-racy analogue of the legacy ack, gated on "we're actually forwarding".
2. Gate the dispatch on `is_renewal == false` so renewals stay on legacy where the ack is still emitted.
3. Teach the legacy GC that absence of ForwardingAck for tx also driven by a remote task-per-tx relay is fine (not easy — the originator can't know).

Option 2 is the cheapest for this PR — it preserves the slice-A scope (fresh non-renewal relay forwards only) and defers the broader ForwardingAck redesign.

**H2. Dedup gate per-node silently drops concurrent path, legacy sent real reply.**

If two different upstream peers independently forward the same `incoming_tx` (routing-bloom collisions or genuinely concurrent paths from a forked-request originator), the second arrival is dedup-rejected at line 829 with no Response to the second upstream. The second upstream's own relay driver then sits in `send_to_and_await` for the full `OPERATION_TTL = 60s` until its timeout path in `drive_relay_subscribe` generates a synthetic `NotFound` upstream. Legacy sent a real Response to each forward, resolving both immediately.

Behavioral change:
- Wastes a 60s timeout budget per additional concurrent path.
- Propagates `NotFound` for a contract the first path just confirmed `Subscribed` — the originator's classifier may misinterpret this as a real failure on that route and retry.
- Inflates `subscribe_not_found` telemetry.

Consider: on dedup reject, either (a) immediately emit a `NotFound` response to the dropped upstream (cheap, preserves legacy signaling), or (b) share a broadcast channel of the real outcome across any driver that wins the dedup race so late-arrivers can bubble the correct result.

### Medium Severity

**M1. `release_pending_op_slot` called twice against a tx that the event loop may have reassigned.**

`drive_relay_subscribe` calls `release_pending_op_slot(incoming_tx)` at line 1126 immediately after the downstream reply resolves. `run_relay_subscribe` then calls it again at line 930 after `send_fire_and_forget` + `yield_now`. Both are keyed on `incoming_tx`. Between these two calls the upstream-reply `send_fire_and_forget` (line 1238) enqueues a new `pending_op_results` entry for the same tx (a closed-receiver sender installed by `send_fire_and_forget`). The second `release` reclaims it — intended. However, the documentation on `send_fire_and_forget` explicitly warns (op_ctx.rs:196–202) that `release_pending_op_slot` can arrive before `handle_op_execution` installs the callback, making the cleanup a no-op and leaving the slot to the 60s sweep. The single `tokio::task::yield_now().await` before the second release is the only ordering guarantee; under contention on the notification channel this is not sufficient. PUT's equivalent uses the same pattern so this is consistent but also carries the same latent risk.

Not a regression relative to PUT slice A, but worth noting that "slot reclaimed by 60s sweep" is the fallback, and at high relay SUBSCRIBE throughput the sweep interval determines the peak `pending_op_results` size.

**M2. Driver does NOT pass `source_addr` when calling `register_downstream_subscriber` on the downstream-Response path.**

Line 1183 passes `Some(upstream_addr)` as the `source_addr` argument. At the relay, on downstream Response arrival, the `source_addr` from the transport layer IS the downstream peer (where the Response originated), NOT the upstream. Legacy at subscribe.rs:1695 passes `None` deliberately. The driver passes `Some(upstream_addr)` which is at least consistent (both args are the requester), but re-using `upstream_addr` as the `source_addr` fallback means if `get_peer_by_addr(upstream_addr)` fails in the first lookup, the fallback also tries `upstream_addr` → double lookup of the same addr. Minor efficiency/consistency nit; behavior is correct because both lookups target the upstream/requester.

**M3. Dedup set leak on panic between `insert` and guard construction.**

Lines 829–859: `insert` at 829, `fetch_add` × 2 at 854–855, tracing call at 841, guard struct literal at 856. A panic (OOM during tracing string format, unwinding from atomic intrinsics, etc.) anywhere in that window leaks the dedup-set entry permanently. Low likelihood, but the guard pattern is meant to prevent exactly this. Move the `insert` to IMMEDIATELY before the guard construction (or construct guard first, then insert inside a `if let Err(_) = ...` branch that drops the guard on failure — but insert on DashSet doesn't fail). Simpler: swap order to `let guard = ...; op_manager.active_relay_subscribe_txs.insert(...); tracing::debug!(...); GlobalExecutor::spawn(...)`.

### Low Severity

**L1. `wait_for_contract_with_timeout` in spawned task is safe but inefficient for local-hit path.** The driver unconditionally waits up to 2s even when contract exists. `wait_for_contract_with_timeout` fast-paths via `has_contract` so this is fine, but a single-line comment confirming the fast path would help readers.

**L2. Driver passes `None` for `requester_pub_key` (lines 970, 1182).** Legacy line 1399/1451/1694 pass `self.requester_pub_key.as_ref()`. The module doc correctly notes the relay path falls back to `get_peer_by_addr(upstream_addr)`. This is adequate because relays have direct connections to their upstream so the addr lookup works reliably — but a node under NAT timing conditions could see transient addr-lookup misses (the exact bug #2886 was designed to avoid). The requester_pub_key was specifically added for that case. Relay hops are less exposed (direct connection) but not immune. Non-blocking.

**L3. Renewals are forwarded identically to fresh requests.** See H1 option 2 — gating renewals out of the new driver avoids both the ForwardingAck amplification AND keeps the lease-amplification invariant the #3763 fix installed. The dispatch gate comment at node.rs:1453 says "ForwardingAck / Unsubscribe never match the Request arm so they flow to legacy naturally" but doesn't acknowledge that RENEWAL Request messages DO match and are now handled without ForwardingAck.

**L4. `op_manager.clone()` in dispatch gate on every inbound SUBSCRIBE Request.** `OpManager` is an `Arc` wrapper so this is a refcount bump, not a structural clone. Fine.

### Nit

**N1. Line 928-930 comment "Release the pending_op_results slot at driver exit" duplicates the release at line 1126 without explaining the invariant that motivates both. Consider a 1-line pointer: "second release covers the `send_fire_and_forget` slot installed in `relay_subscribe_send_response`".**

### Questions for Author

1. Is option 2 (excluding `is_renewal=true` from the dispatch gate) acceptable for slice A, deferring the ForwardingAck redesign to a follow-up? The renewal cycle is where the amplification risk is largest.
2. Was the dedup-reject → silent-drop behavior (H2) intentional? If so, what is the expected timeline for a synthetic NotFound response to the dropped upstream?
3. Why is `register_downstream_subscriber` called with `Some(upstream_addr)` rather than `None` for `source_addr` on the Response path (line 1183) when legacy uses `None`?

### What I Verified

- Dedup gate ordering (insert before spawn) is correct — pin test catches regressions.
- `VisitedPeers::with_transaction` preserves `bits`, only restores `hash_keys`. Bloom filter state from upstream hops is NOT lost (Q6 safe).
- `register_downstream_subscriber` is idempotent via `add_downstream_subscriber` returning false on re-register. Double-call scenario in Q2 would not cause duplicate state, but the two driver paths (local hit vs downstream Subscribed) are mutually exclusive within a single driver instance anyway.
- `wait_for_contract_with_timeout` is thread-safe (tokio-native, no thread-locals). Safe to call from spawned task (Q3).
- Originator task-per-tx bypass (node.rs:1422) correctly gates on `matches!(op, Response { .. })` and will NOT fire for `SubscribeMsg::Request` arriving at the originator. The relay driver's `send_to_and_await` relies on this exact bypass to receive the downstream reply — working as designed.
- Dispatch gate correctly falls through to legacy on renewals / PUT sub-ops / GC retries via `has_subscribe_op`. Structural pin test guards this.
- `source_addr.is_none()` (originator loopback) correctly falls through to legacy.

iduartgomez:
The single WARNING in the rule-review above (stale `.claude/rules/operations.md` preamble not updated for SUBSCRIBE slice A) was addressed in `e873328c` (pushed after the bot ran against `b560150c`). The preamble now has a SUBSCRIBE slice A paragraph mirroring the PUT/UPDATE ones, covering:

- migration of fresh inbound `SubscribeMsg::Request` to `start_relay_subscribe`
- gate on `source_addr.is_some() && !has_subscribe_op(id) && !is_renewal`
- the four amplifier mitigations (retry cap, tx reuse, no ForwardingAck, dedup gate)
- the deliberate `ring.subscribe` / `complete_subscription_request` / `announce_contract_hosted` omission (prevents #3763 subscription-storm)
- legacy fall-through list (renewals, PUT sub-op, executor, GC retries, Unsubscribe, ForwardingAck)

`crates/core/CLAUDE.md` was also updated in the same commit. Current HEAD is `139c6bea` (adds M3 panic-window tightening and the PeerHealth signal-drop rustdoc from code-first #1). Rule-review should re-run against HEAD.

The two INFO items (double `release_pending_op_slot`, no runtime behavioral unit tests for error paths) are accepted as matching the established GET/PUT/UPDATE slice-A pattern — the double-release mirrors the PUT follow-up `06a18b10`, and runtime behavioral coverage comes from `test_get_routing_coverage_low_htl` which asserts `RELAY_SUBSCRIBE_DRIVER_CALL_COUNT > 0` (added in `e873328c` as testing-review H1 fix).

PR #3935, opened 2026-04-21
refactor(ops): relay PUT streaming task-per-tx (slice B) — #1454 phase 5

iduartgomez:
## Summary

- Migrate fresh inbound streaming PUT relays to the task-per-transaction driver pattern (`start_relay_put_streaming`).
- Extract a shared `relay_put_store_locally` helper so slice A + slice B share the local-store sequence verbatim.
- Dispatch gate in `handle_pure_network_message_v1` now routes fresh inbound `PutMsg::RequestStreaming` to the task-per-tx driver; direct `PutMsg::Request` continues to hit slice A for non-streaming payloads.

## Context

#1454 phase 5 moves relay operations onto the task-per-transaction driver pattern. Slice A (PR #3917) handled non-streaming PUT relays; slice B closes the streaming gap. The streaming driver claims the inbound stream via `orphan_stream_registry`, forks the handle for downstream piping via `NetworkBridge::pipe_stream`, assembles + stores locally, and bubbles a non-streaming `PutMsg::Response` upstream (mirroring legacy put.rs:1506 downgrade).

## Key design choices

- Driver access to `pipe_stream` is via a cloned `NetworkBridge` passed into the spawn. `handle_pure_network_message_v1`'s generic `CB` picks up `Clone + 'static` bounds.
- Driver shares `active_relay_put_txs` dedup set with slice A (same tx space).
- Downstream reply routing uses `OpCtx::send_to_and_await` for the metadata dispatch; `pipe_stream` runs in parallel with local stream assembly.
- `ForwardingAck` emission is omitted (same rationale as slice A — ack sharing `incoming_tx` would race the real `Response` into upstream's pending_op_results slot).

## Out of scope / stays on legacy

- `PutMsg::Request` with oversized payload (upgrade-on-forward) but no inbound StreamId → no orphan-registry claim side, legacy.
- Client-initiated streaming PUTs (`source_addr.is_none()`).
- GC-spawned speculative retries.
- PUT dead-code removal (`PutState::PrepareRequest`, `start_op`, `start_op_with_id`, `PrepareRequestData`) — deferred to a follow-up PR; runtime match arms still reference these types in non-trivial ways.

## Testing

- [x] `cargo test -p freenet --lib --features testing` (2449 passed)
- [x] `cargo test -p freenet --features simulation_tests,testing --test streaming_e2e test_streaming_multi_hop_forwarding` (runtime counter assertion)
- [x] `cargo clippy -p freenet --tests --features testing -- -D warnings`
- [x] Structural pin tests: dedup gate ordering, NotFound on dedup reject, RAII Drop balance, orphan-stream claim + pipe_stream usage, ForwardingAck omission, incoming_tx reuse on forward
- [x] Dispatch-gate pin tests in `node.rs`: streaming variant dispatched, gated on source_addr + `!has_put_op`
- [ ] Linux integration run for `test_put_contract_three_hop_returns_response` + friends (macOS can't bind 127.x.x.x)

## Fixes

Part of #1454 phase 5.

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -39,9 +39,23 @@ paths:
 > relay `PutMsg::Request` to
 > `operations/put/op_ctx_task.rs::start_relay_put` (gated on
 > `source_addr.is_some()` AND `!has_put_op(id)` AND the payload size
-> would not upgrade to streaming on forward); GC speculative retries,
-> client-initiated loopback (`source_addr.is_none()`), and streaming
-> variants still go through legacy. Phase 5 follow-up slice A (PR
+> would not upgrade to streaming on forward); GC speculative retries and
+> client-initiated loopback (`source_addr.is_none()`) still go through
+> legacy. Phase 5 follow-up slice B migrated fresh inbound streaming
+> relay `PutMsg::RequestStreaming` to
+> `operations/put/op_ctx_task.rs::start_relay_put_streaming` (same
+> `source_addr.is_some()` AND `!has_put_op(id)` gate). The streaming
+> driver claims the inbound stream via
+> `orphan_stream_registry().claim_or_wait`, forks the handle for
+> downstream piping via `NetworkBridge::pipe_stream`, assembles
+> locally via the shared `relay_put_store_locally` helper, and
+> bubbles a non-streaming `PutMsg::Response` upstream (mirrors the
+> legacy downgrade at `put.rs:1506`). The dispatch site passes a
+> cloned `NetworkBridge` into the spawn so the driver can call
+> `pipe_stream` without a detour through `OpCtx`. The upgrade-only
+> case — `PutMsg::Request` whose serialized payload would exceed
+> `streaming_threshold` on forward but has no inbound `StreamId` to
+> claim — stays on the legacy path in slice B. Phase 5 follow-up slice A (PR
 > #3932) migrated fresh inbound relay `SubscribeMsg::Request` to
 > `operations/subscribe/op_ctx_task.rs::start_relay_subscribe`
 > (gated on `source_addr.is_some()` AND `!has_subscribe_op(id)`);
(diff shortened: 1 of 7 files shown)


iduartgomez:
## Review response (commit 6db99b7a)

Addressed high-severity findings from the parallel review pass:

**Fixed:**
- **H2 (waiter-install race):** `pipe_stream` was awaited BEFORE `ctx.send_to_and_await` had a chance to install the `pending_op_results` callback. A fast downstream reply could be dropped as `OpNotPresent`. Added `OpCtx::send_to_and_register_waiter` primitive — enqueues the metadata + returns the receiver without awaiting — and call it first so the waiter is installed before the pipe_stream enqueue.
- **H3/H4 (fabricated Response on reject):** `PutMsg` has no NotFound variant, so the "wake the upstream waiter" fire-and-forget was sending a success Response — telling upstream "contract stored" after a dedup-reject or orphan-claim failure. Both paths now return silently (matches slice A's dedup-reject; upstream falls back to its own OPERATION_TTL).
- **Dead code:** removed `let _ = streaming_total_size;` + the now-unused tuple in the dispatch gate.
- **Counter doc:** clarified that `RELAY_PUT_STREAMING_DRIVER_CALL_COUNT` fires before the dedup gate (counts admitted + rejected).

**Not addressed (documented as deferred):**
- M1 (contract-key mismatch checked after pipe): legacy has the same shape. Pre-assembly check would require deserializing metadata twice. Keeping parity with legacy for this slice.
- M3 (`skip_list` doesn't include `next_addr`): matches legacy `routing_skip_list` (`put.rs:1046`). Not a regression.
- Runtime coverage for `AlreadyClaimed` + `pipe_stream` failure fallback: pin tests cover the shape; runtime coverage needs a fault-injection simulation that's beyond slice B scope.

Local: fmt + clippy + lib tests (2449 passed) + streaming_e2e green. Pushed.

PR #3939, opened 2026-04-22
fix(tray): replace dup-tray port preflight with flock-based wrapper guard

sanity:
## Problem

#3938's merge queue landed my first-cut port-based preflight rather than the flock revision I'd force-pushed onto the branch after the skeptical review. Merge queues accept a SHA, then ignore subsequent force-pushes; the queue merged the original commit. Phase-1 smoke test of the resulting DMG (run 24797845382, commit 0d6c18a4) reproduced the dup-tray bug exactly as that review predicted.

Evidence from Mac-Claude's smoke test (mailbox msg 006):

```
$ ps -eo pid,ppid,etime,args | grep freenet-bin
22983     1       00:24 /Applications/Freenet.app/Contents/MacOS/freenet-bin service run-wrapper
23017     1       00:20 /Applications/Freenet.app/Contents/MacOS/freenet-bin service run-wrapper
23024 23017       00:20 /Applications/Freenet.app/Contents/MacOS/freenet-bin network
```

Two wrappers. Backend (23024) is a child of the launchd-spawned wrapper (23017), not the user-opened one (22983). Matches the `kill_stale_freenet_processes` → "port free" → dup-tray race the skeptical review diagnosed.

## Why the port-based preflight is broken

`kill_stale_freenet_processes` runs BEFORE the preflight. That helper `pkill`s any `freenet network` process owned by the user — including live peer wrapper's backend child. So wrapper B kills A's backend, sleeps 2s, checks the port, finds it free (because B just killed the occupant), and proceeds to build a tray. The preflight thinks it's in the happy path.

Secondary issue: non-Freenet port squatters on 7509 would cause silent exit with a misleading "another Freenet wrapper" log.

## Solution

Wrapper-scoped advisory `flock` at `~/Library/Caches/Freenet/wrapper.lock`, acquired in `run_wrapper` BEFORE `kill_stale_freenet_processes` and before anything user-visible. `LOCK_EX | LOCK_NB`. Kernel releases on process exit so crashes don't leave stale locks.

Lock → kill_stale → tray ordering pinned by code comment. If the lock is held by a peer wrapper, we exit silently without ever running kill_stale, so we never murder a live peer's backend child.

Covers overlap cases the port preflight missed:
- Wrapper A's backend in mid-backoff or mid-restart
- Non-Freenet squatter on 7509 (no false exit)
- `kill_stale` interaction (B exits before kill_stale runs)

## Testing

Port-based tests were tautological (would pass even if the whole preflight were deleted). Replaced with:

- `wrapper_lock_is_exclusive` — acquires a flock on a tempfile, asserts second acquire fails with first alive, then releases and asserts third succeeds. Exercises the actual kernel mechanism. Runs on Linux CI.
- `wrapper_lock_path_is_under_cache_dir` — pins the path invariant.

`cargo fmt` + `cargo clippy -- -D warnings` clean.

## Merge process

Will NOT force-push this branch once CI is green. If changes are needed, new commits on top. Avoids the merge-queue race that lost #3938's real fix.

## Fixes

Closes the regression from #3938. Real dup-tray fix that the smoke test will actually pass this time.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -288,30 +288,128 @@ fn dashboard_port_is_listening() -> bool {
     TcpStream::connect_timeout(&addr, Duration::from_millis(500)).is_ok()
 }
 
-/// Pure decision function for the macOS wrapper preflight. Extracted so
-/// the decision logic can be unit-tested on any platform without binding
-/// real sockets. Today's rule is trivially "exit if another wrapper's
-/// port is already bound", but keeping it as a named function leaves
-/// room to layer on version-mismatch checks, legacy-install collisions,
-/// etc. without touching `run_wrapper`.
-#[derive(Debug, PartialEq, Eq)]
+// ── macOS wrapper single-instance lock ──
+//
+// Problem: launchd's LAL agent fires RunAtLoad while the user's
+// open-launched wrapper is still alive, and spawns a second wrapper.
+// The backend-level single-instance guard
+// (EXIT_CODE_ALREADY_RUNNING = 43) catches the second wrapper's
+// DAEMON CHILD but by the time that exit propagates, the second
+// wrapper's tao event loop is already running on the main thread and
+// owns an NSStatusItem. Two rabbits.
+//
+// Port-based detection of "another wrapper is running" is fragile:
+//   - wrapper A's backend may be mid-crash or mid-backoff, in which
+//     case port 7509 is briefly free but wrapper A's tray is still up
+//   - `kill_stale_freenet_processes` runs early in wrapper startup
+//     and happily pkills the OTHER wrapper's live backend child,
+//     widening the race window it was meant to close
+//   - a non-Freenet squatter on port 7509 triggers a false positive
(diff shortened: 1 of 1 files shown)


sanity:
test-exempt justification: the two deleted tests (`wrapper_preflight_exits_when_port_held_by_other_wrapper`, `wrapper_preflight_proceeds_when_port_is_free`) were tautological assertions that `fn(true) == A, fn(false) == B` on a pure decision function whose implementation is being entirely replaced by the flock-based approach in this PR. They reference the types `WrapperPreflightDecision` and `macos_wrapper_preflight_decision` which are removed by this PR. `#[ignore]`-ing them per the usual rule isn't applicable because the symbols they exercise no longer exist. Replaced by two NEW tests (`wrapper_lock_is_exclusive`, `wrapper_lock_path_is_under_cache_dir`) that exercise the actual kernel flock mechanism with a tempfile, catching the H1/H2 concerns raised by the skeptical reviewer on #3938. Net test count unchanged; regression coverage strictly improved. [AI-assisted - Claude]

PR #3941, opened 2026-04-22
fix(tray,update): null stdio on update/notepad spawns + wire auto-update failure lockout (#3933, #3934)

### LINKED ISSUE #3933: Bug with Windows tray app behaviour on system start-up or restart.
I have had repeated issues with the Freenet tray app on Windows not behaving as expected on PC startup and restart. The 'Open Dashboard', start, stop, restart and quit menu options work, but 'View Logs' and 'Check for updates' do not. These all work when Freenet is initially installed though.

I fed this scenario into Codex, and it ultimately made the following observations. Hopefully they help fix this.

---

Open Dashboard works because it is handled in the tray/UI thread and on Windows it uses ShellExecuteW, explicitly to avoid the detached-console problem in mod.rs (line 11) and mod.rs (line 25).

Start, Stop, and Restart work because they do not spawn a new external process. They just send a control message to the wrapper loop, which then kills or relaunches the node child in-process.

The two broken items are different:

View Logs calls Command::new("notepad").spawn() in tray.rs (line 645) and tray.rs (line 659).
Check for Updates calls spawn_update_command() in service.rs (line 1049), which runs freenet update --quiet.
The wrapper detaches from the console at startup with FreeConsole() in service.rs (line 967). Later in the same file, the code already documents that after FreeConsole() spawned children can fail unless stdin/stdout/stderr are explicitly redirected to null; that fix is applied for the network child in service.rs (line 1263) through service.rs (line 1278).

That protection is not applied to:

spawn_update_command()
the Notepad launch in open_log_file()
So the likely reason is: after Windows autostart, the tray wrapper is running detached, and those two actions try to spawn child processes while inheriting invalid standard handles. That would fail silently, while the other menu items keep working.

### LINKED ISSUE #3934: Bug: Freenet node fails to auto update correctly on Windows when out-of-date version detected on startup.
When starting Windows and the Freenet node starts, if it detects a newer version on GitHub it tries to update, but fails and restarts repeatedly. This has happened multiple times on different Windows systems that were started at bootup with a node at least one release out of date. This generates the following example wrapper log:

```
12:15:27: Update needed, running freenet update...
12:15:27: Update failed
12:15:27: Exited with code 42, restarting after 10s backoff...
12:16:28: Update needed, running freenet update...
12:16:28: Update failed
12:16:28: Exited with code 42, restarting after 20s backoff...
12:17:19: Update needed, running freenet update...
12:17:19: Update failed
12:17:19: Exited with code 42, restarting after 40s backoff...
12:18:32: Update needed, running freenet update...
12:18:32: Update failed
12:18:32: Exited with code 42, restarting after 80s backoff...
12:19:53: Update needed, running freenet update...
12:19:53: Update failed
12:19:53: Exited with code 42, restarting after 160s backoff...
12:23:12: Update needed, running freenet update...
12:23:12: Update failed
12:23:12: Exited with code 42, restarting after 300s backoff...
12:28:13: Update needed, running freenet update...
12:28:13: Update failed
12:28:13: Exited with code 42, restarting after 300s backoff...
12:32:56: Update needed, running freenet update...
12:32:56: Update failed
12:32:56: Exited with code 42, restarting after 300s backoff...
12:39:06: Update needed, running freenet update...
12:39:06: Update failed
12:39:06: Exited with code 42, restarting after 300s backoff...
12:44:19: Update needed, running freenet update...
12:44:19: Update failed
12:44:19: Exited with code 42, restarting after 300s backoff...
```

Codex made the following observations about node behaviour which may be useful:

The update is probably not failing at download or version validation. It is failing at the final Windows install step.

The code path is:

The node confirms a newer GitHub release and exits with code 42 for auto-update in freenet.rs (line 447).
The Windows wrapper catches that and runs the same executable as freenet update --quiet in service.rs (line 1491) and service.rs (line 1050).
The updater then calls replace_binary(current_exe) in update.rs (line 307).
replace_binary() tries to rename the currently installed freenet.exe to freenet.old, then rename the new file into place in update.rs (line 1269).
So the repo is assuming that on Windows it can replace the live .exe in place. That assumption is explicitly documented in a comment in service.rs (line 1486). If Windows, antivirus, or another handle on the file prevents that rename, the update subprocess fails even though the release itself is valid.

The reason it loops forever is a second bug: auto-update failure tracking exists, but it is not wired up. record_update_failure() is marked “will be wired up in follow-up” in auto_update.rs (line 256), so failed installs do not trip the intended max-failures cutoff. The wrapper just restarts the old binary, which immediately sees the same newer release and exits for update again.

So the short answer is: the update is valid, but the Windows self-replacement step is brittle, and when it fails the retry-suppression path is incomplete.

sanity:
## Problem

Two tightly coupled Windows tray / auto-update bugs, both reported against v0.1.186 by @Ivvvor:

- **#3933** — After Windows autostart, the \"View Logs\" and \"Check for Updates\" tray menu items silently do nothing, while other menu items keep working.
- **#3934** — On startup, an out-of-date node detects a newer release, exits with code 42 for auto-update, but the wrapper reports \"Update failed\" and enters an exponential-backoff restart loop that never terminates.

Both issues share the same primary root cause.

## Approach

The wrapper calls `FreeConsole()` on Windows and, when launched by autostart, may never have had a console at all. Either way, its inherited standard handles are invalid. Any child `Command::spawn()` that leaves stdio unspecified then fails with *\"The handle is invalid\" (os error 6)* — there is already a load-bearing comment documenting exactly this at `service.rs:1252-1263`, and the network-child spawn explicitly nulls stdin/stdout/stderr for this reason. The same pattern was missing from:

- `service.rs::spawn_update_command` — **this is the root cause of #3934**: the update subprocess never actually launches, so the wrapper sees it \"fail\", runs BackoffAndRelaunch, child exits 42 again, loop forever.
- `tray.rs::open_log_file` (notepad / open / xdg-open) — **this is the \"View Logs\" half of #3933**.

Fixing the stdio alone would eliminate this specific loop, but #3934 also flagged a second latent bug:

- `auto_update::record_update_failure()` existed with `#[allow(dead_code)] // Will be wired up to update command in follow-up`, so even if the actual install failed for *any* reason (AV lock, read-only install dir, checksum mismatch, etc.) the `MAX_UPDATE_FAILURES` gate in `check_if_update_available` could never trigger. The exit-42 loop was unbounded in the general case.

This PR wires that gate up correctly.

### Changes

1. **Null stdio for all detached-wrapper child spawns** (#3933 primary, #3934 primary):
   - `spawn_update_command` now sets `Stdio::null()` on stdin/stdout/stderr on every platform.
   - `open_log_file` does the same for its notepad / open / xdg-open spawn.

2. **Wire up the auto-update failure counter** (#3934 defense-in-depth):
   - Both wrapper call sites of `spawn_update_command` now route the result through `classify_update_subprocess` and drive the persistent counter: `BinaryReplaced`/`BundleUpdateStaged` → `clear_update_failures()`, `Failed` → `record_update_failure()`, `AlreadyUpToDate` → no-op.
   - `UpdateCommand::download_and_install` also calls `clear_update_failures()` after a successful `replace_binary`, so a manual `freenet update` from a terminal recovers a locked-out auto-update state.
   - Removed the premature `clear_update_failures()` call in `check_if_update_available`'s `UpdateAvailable` arm — it fired on every peer-mismatch check and prevented the counter from ever accumulating (it would have defeated the gate even after wiring).

3. **Testable variants**:
   - `record_update_failure`, `clear_update_failures`, `get_update_failure_count`, and `should_attempt_update` now each have a `_at(&Path)` sibling that operates on an explicit state directory, so the lockout invariant can be exercised deterministically in a unit test.

## Testing

New tests in `auto_update::tests`:

- `test_update_failure_counter_roundtrip` — record/clear/get round-trip + idempotent clear.
- `test_should_attempt_update_locks_out_after_max_failures` — **core #3934 regression**: after `MAX_UPDATE_FAILURES` consecutive records, `should_attempt_update_at` returns false; after a clear it returns true again. This is the invariant the original bug violated.
- `test_update_failure_counter_persists_on_disk` — counter survives process restarts (the child records via the wrapper's subprocess-result dispatch, then the wrapper relaunches a fresh child; if the counter were memory-only the lockout would never fire).

All 17 `commands::auto_update` tests pass locally.

### Why didn't CI catch this?

The `#[allow(dead_code)] // Will be wired up to update command in follow-up` comment was load-bearing: the lockout logic existed but was never exercised end-to-end. The new regression tests would have failed in the broken state.

The null-stdio failure mode only manifests on Windows when the wrapper has detached from / never had a console. A cross-platform unit test can't reliably reproduce that, but the fix is a direct, comment-documented copy of the exact pattern already used for the network-child spawn 80 lines away in the same file, so code review should catch any future regression.

### Notes for reviewers

- Pre-existing clippy warnings in `crates/core/src/tracing.rs` (wildcard enum match arms against `EventKind`) also fail `cargo clippy --all-targets --all-features -- -D warnings` on unmodified `main`. They are unrelated to this PR.

## Fixes

Closes #3933
Closes #3934

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -260,12 +260,19 @@ fn get_update_failure_count() -> u32 {
 }
 
 /// Testable variant of [`get_update_failure_count`] that reads from an explicit
-/// directory. Missing / unparseable files are treated as zero.
+/// directory.
+///
+/// * Missing file → `0` (legitimate "no failures yet").
+/// * Present but unparseable → `MAX_UPDATE_FAILURES` (defensive: if the
+///   counter file has been truncated or corrupted we must NOT silently
+///   reset the lockout — that would be an amplification vector for any
+///   process that can partially overwrite the file, defeating the
+///   #3934 fix. Users can recover by explicitly deleting the file).
 pub(crate) fn get_update_failure_count_at(dir: &std::path::Path) -> u32 {
-    fs::read_to_string(dir.join("update_failures"))
-        .ok()
-        .and_then(|s| s.trim().parse().ok())
-        .unwrap_or(0)
+    match fs::read_to_string(dir.join("update_failures")) {
+        Ok(s) => s.trim().parse().unwrap_or(MAX_UPDATE_FAILURES),
+        Err(_) => 0,
+    }
 }
 
 /// Record an update failure. Called by the update command when the install
@@ -306,10 +313,19 @@ pub(crate) fn clear_update_failures_at(dir: &std::path::Path) {
 }
 
 /// Check if we should attempt an update based on failure history.
+///
+/// If the state directory cannot be resolved (e.g. Windows service
+/// account with no `USERPROFILE`), returns `false`: with no place to
+/// persist the failure counter we cannot distinguish a fresh session
+/// from one that has been looping for hours, so the safest choice is
+/// to skip auto-update entirely rather than risk an unbounded exit-42
+/// loop (skeptical-review H2 on PR #3941). Users in that situation
+/// still receive updates via whatever external packaging mechanism
+/// installed them.
 pub fn should_attempt_update() -> bool {
     state_dir()
         .map(|d| should_attempt_update_at(&d))
-        .unwrap_or(true)
+        .unwrap_or(false)
 }
 
 /// Testable variant of [`should_attempt_update`] that reads from an explicit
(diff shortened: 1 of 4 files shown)


sanity:
## Review feedback addressed (commit bbc08e0c)

Running four internal review agents + Codex surfaced the following findings. All fixed.

### Codex P1 — lockout counter over-triggered
- **Finding**: `UpdateSubprocessOutcome::Failed` lumped spawn errors, transient GitHub / checksum failures, AND real install failures. Three flaky GitHub calls would permanently disable auto-update.
- **Fix**: split into `SpawnFailed` (env-persistent → records) and `OtherFailure` (transient → no-op). Install-stage failures are recorded directly from `UpdateCommand::download_and_install` around `replace_binary`, so genuine install failures (AV lock, permission denied) still lock out.
- **Helper**: pure `update_counter_action(outcome) -> UpdateCounterAction` used at both wrapper call sites; covered by `update_counter_action_matches_outcome_semantics`.

### Skeptical review H1 — permanent lockout after external upgrade
- **Finding**: if a user hit `MAX_UPDATE_FAILURES` and then upgraded externally (apt / installer / homebrew), subsequent `freenet update` invocations took the `AlreadyUpToDate` branch without ever touching the counter → permanent lockout.
- **Fix**: `AlreadyUpToDate` path in `UpdateCommand::run_async` now calls `clear_update_failures()` before exiting.

### Skeptical review H2 — `state_dir()` None defeats lockout
- **Finding**: on a Windows service account without `USERPROFILE`, `state_dir()` returns `None`, `record_update_failure` silently no-ops, and `should_attempt_update()` returned `true` → unbounded loop.
- **Fix**: `should_attempt_update()` now returns `false` when the state dir cannot be resolved. Without persistent state we can't bound the loop, so the safest choice is to skip auto-update entirely.

### Testing review — corrupt counter file silently defeated lockout
- **Finding**: `get_update_failure_count_at` used `.unwrap_or(0)`, so a truncated / garbled counter file silently reset the lockout — an AV that partially overwrites the file could defeat it.
- **Fix**: present-but-unparseable → `MAX_UPDATE_FAILURES` (defensive). Deleting the file is still an explicit recovery path. Covered by `test_corrupt_counter_file_is_treated_as_max`.

### Testing / Rule-review — null-stdio change had no regression test
- **Finding**: portable unit tests can't reproduce the Windows-only `spawn()` failure, but the absence of any regression test means a future revert ships the bug silently.
- **Fix**: source-level regression pins (`open_log_file_spawn_must_null_all_three_standard_handles`, `spawn_update_command_must_null_all_three_standard_handles`, `test_check_if_update_available_does_not_clear_failure_counter`, `test_should_attempt_update_conservative_when_state_dir_missing`) — grep the source for the specific call sequences and fail with issue-numbered error messages if any is removed.

### Testing review — glue between classifier and counter untested
- **Finding**: the wrapper's `classify_update_subprocess` → `record`/`clear` dispatch had no direct test. A future refactor could remove one of the two call sites and reopen #3934.
- **Fix**: extracted pure `update_counter_action` helper, directly unit-tested.

### Big-picture review — missed guardrail for a three-time-offender pattern
- **Finding**: `FreeConsole` + `Command::spawn` without null-stdio is now the third recurrence of the same pattern (network-child → #3933 tray → #3934 update). Nothing documents this yet.
- **Fix**: added entry to `~/code/freenet/.claude/rules/bug-prevention-patterns.md` with audit grep. Cross-reference comment on the `FreeConsole()` call site at `service.rs::run_wrapper` lists every downstream spawn that must set null stdio.

### Code-first review — minor doc suggestions
- Dropped `#[allow(dead_code)]` from `UpdateSubprocessOutcome` and `classify_update_subprocess` (both used now).
- Updated comments to reflect the split.

All 108 tests pass locally. CI re-running. [AI-assisted - Claude]

PR #3942, opened 2026-04-22
fix(server): fetch webapp on cold cache in variable_content

### LINKED ISSUE #3940: Bug: Resources inside a web container contract cannot be used until the root of the contract is used/accessed first.
If a Freenet dapp or website attempts to use a resource stored within a web container contract archive directly before the root of the container is accessed and opened, the container is not opened and cached to local storage and a 404 error results.

Example: http://127.0.0.1:7509/v1/contract/web/ENYDgNFUDomM4dMstSnyYnJVi5nB8wmSGJLj9fwF6Pi7/freenet.jpg accessed as an image reference from a Delta site. This will not work unless the user visits: http://127.0.0.1:7509/v1/contract/web/ENYDgNFUDomM4dMstSnyYnJVi5nB8wmSGJLj9fwF6Pi7/ first.

sanity:
## Problem

When a browser requested a subresource of a web contract (e.g. an
`` pointing at `http://127.0.0.1:7509/v1/contract/web/<KEY>/freenet.jpg`
from a different contract), the request 404'd unless the user had
previously visited the root URL `http://127.0.0.1:7509/v1/contract/web/<KEY>/`.

The shell-root handler (`contract_home`) fetches the contract from the
network and unpacks it into the webapp cache, but `variable_content`
served files directly from the cache with no fetch fallback. Cross-
contract asset references therefore failed on first load — precisely
the scenario in the bug report (Delta page with an `` pointing at
a separate contract's image).

## Approach

Extract the GET + unpack pipeline out of `contract_home` into a shared
`ensure_contract_cached` helper. `variable_content` now checks whether
the state-hash marker file exists — the same signal `contract_home`
uses internally — and triggers the fetch when it doesn't, before
falling through to the existing file-serving logic.

The hash file is the correct cache-validity marker because
`unpack_if_stale` writes it last; a partial unpack leaves no hash,
forcing a re-fetch on the next request.

`web_subpages`, `web_subpages_v1`, and `web_subpages_v2` thread the
`HttpClientApiRequest` extension through so the cache-miss path has
a channel to the node.

## Testing

- `variable_content_triggers_fetch_on_cache_miss`: verifies a cold
  cache causes the handler to emit `NewConnection` + `Get` on the
  client-connection channel for the requested contract key.
- `variable_content_skips_fetch_when_cache_present`: verifies a warm
  cache (hash file present) serves the file without sending any
  client-connection messages — pins the fast path so a regression can't
  silently re-fetch on every request.
- Existing `web_subpages_redirects_top_level_document_load_to_shell`
  updated to pass a dead request sender; redirect behaviour is
  unchanged.
- Full \`cargo test -p freenet --lib\`: 2468 passed, 0 failed.

## Fixes

Closes #3940

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -61,6 +61,15 @@ impl std::ops::Deref for HttpClientApiRequest {
     }
 }
 
+impl HttpClientApiRequest {
+    /// Constructs a request wrapper around an existing sender. Used by
+    /// sibling-module tests that need to feed in a mocked channel.
+    #[cfg(test)]
+    pub(super) fn from_sender(sender: mpsc::Sender<ClientConnection>) -> Self {
+        Self(sender)
+    }
+}
+
 /// Represents an origin contract entry with metadata for token expiration.
 #[derive(Clone, Debug)]
 pub struct OriginContract {
@@ -250,6 +259,7 @@ async fn web_subpages(
     api_version: ApiVersion,
     query_string: Option<String>,
     req_headers: axum::http::HeaderMap,
+    request_sender: HttpClientApiRequest,
 ) -> Result<axum::response::Response, WebSocketApiError> {
     let is_sandbox = query_string
         .as_ref()
@@ -290,7 +300,7 @@ async fn web_subpages(
 
     let version_prefix = api_version.prefix();
     let full_path: String = format!("/{version_prefix}/contract/web/{key}/{last_path}");
-    path_handlers::variable_content(key, full_path, api_version)
+    path_handlers::variable_content(key, full_path, api_version, request_sender)
         .await
         .map_err(|e| *e)
         .map(|r| {
@@ -526,6 +536,17 @@ impl ClientEventsProxy for HttpClientApi {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// Builds an `HttpClientApiRequest` whose receiver is dropped immediately,
+    /// so any attempt to send on it fails fast with a closed-channel error.
+    /// Suitable for redirect-branch tests that must never reach the full
+    /// fetch/unpack pipeline — the handler short-circuits to a redirect before
+    /// the sender is used, and if it doesn't the send error surfaces as a
+    /// bounded test failure instead of a 30-second timeout.
+    fn dead_request_sender() -> HttpClientApiRequest {
+        let (tx, _rx) = mpsc::channel(1);
+        HttpClientApiRequest::from_sender(tx)
+    }
 
     #[test]
     fn is_html_page_detects_html_extensions() {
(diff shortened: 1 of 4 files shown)


sanity:
## Review findings

Ran 4 internal reviewers (code-first, testing, skeptical, big-picture) + Codex in parallel. Consolidated findings and how each was addressed:

### Addressed in 317d4543

1. **Concurrent cold-cache race — unanimous flag (all 5 reviewers).** A typical first page load fans out several subresource requests (``, `<script>`, `<link>`), each observes the cache as cold, each independently runs `remove_dir_all` + `create_dir_all` + `unpack` into the same directory. This can truncate files another task is unpacking/serving and leave a valid-looking hash file pointing at a corrupted tree.
   **Fix:** `CONTRACT_CACHE_LOCKS: LazyLock<DashMap<ContractInstanceId, Arc<Mutex<()>>>>`. `unpack_if_stale` holds the per-instance lock during cache mutations and re-reads the hash under the lock, so followers see the fresh hash written by the leader and exit without repeating the work.

2. **Disconnect error swallowing (skeptical).** `debug!` was too quiet — a closed request channel means the node is going away. Bumped to `warn!` so an operator gets a visible breadcrumb when subsequent WebSocket connections start failing.

3. **Test hygiene (testing).** `variable_content_triggers_fetch_on_cache_miss` now calls `clear_cache` at the end (prior version left residue from the aborted handler). `variable_content_skips_fetch_when_cache_present` now asserts the exact body bytes, not just `Ok`.

4. **`handle_get_response` arm coverage (testing).** Added direct unit tests for the `contract: None` → `MissingContract` arm and the `Elapsed` → `NodeError` arm. Both were pure-logic seams introduced by the refactor but had no direct coverage.

### Filed as follow-ups (out of scope for this PR)

- #3944 — pre-existing flake in `relay_does_not_accept_after_forwarding_on_retry` (observed once during the full-suite run, passes in isolation and on re-run).
- #3945 — cold-cache fetch widens DoS amplification for unknown-key enumeration.
- #3946 — tar unpack of web contract lacks path-traversal guards (pre-existing, now reachable via subresource URLs not just top-level navigation).

### Explicitly not changed

- **Big-picture reviewer claimed this PR reverts PR #3939's flock wrapper guard.** False positive — `git diff origin/main...fix-3940 --stat` is clean, no `service.rs` changes. The reviewer likely diffed against a stale local `main`.
- **Skeptical flagged `response_channels` entry potentially orphaned.** Pre-existing pattern, no behavior change in this PR — `contract_home` always did two `NewConnection` + one `Disconnect`, same as `ensure_contract_cached` now does for subresource requests.

### Test status

`cargo test -p freenet --lib`: 2470 passed, 0 failed (one flake on a first run of a pre-existing-flaky test in `operations::connect`, filed as #3944; passed cleanly on re-run).

[AI-assisted - Claude]

PR #3958, opened 2026-04-26
fix(server): accept private-IP Host headers on WS upgrade by default

### LINKED ISSUE #3957: Sandboxed iframe same-origin error when accessing River from non-localhost IP (LAN/bridged-VM)
## Summary

User report on Matrix (fluffomat, 2026-04-25) — loading River from a non-localhost LAN
IP triggers a browser same-origin error inside the sandboxed iframe.

## Symptoms

User runs the freenet node inside a VirtualBox VM (Win11 host, bridged networking)
and accesses River from the host browser via `http://192.168.178.192:7509/...`.
Browser console reports:

```
Unsafe attempt to load URL
  http://192.168.178.192:7509/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/?__sandbox=1
from frame with URL
  http://192.168.178.192:7509/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/?__sandbox=1.
Domains, protocols and ports must match.
```

Both URLs in the error are identical — the iframe's own `?__sandbox=1` URL — so this
is the classic opaque-origin failure. The iframe is served with
`sandbox=\"allow-scripts allow-forms allow-popups\"` (no `allow-same-origin`,
`crates/core/src/server/path_handlers.rs:473`), which assigns the document an opaque
origin. Any same-origin operation (sync XHR, `fetch` against own URL, Service Worker
registration, etc.) then fails the same-origin check even though the URL strings
match.

This is **not** the same as #3464: that issue is specifically the WebSocket
`Origin: null` rejection at the gateway, and its proposed fix only allows it from
localhost. fluffomat is connecting over LAN, so #3464's fix wouldn't help here even
once it lands. The error message format (`Unsafe attempt to load URL ...`) is also
distinct from the WebSocket-specific error — it indicates a `fetch`/XHR/Service
Worker call from inside the iframe rather than a WebSocket handshake.

## Why this matters

LAN access (e.g. running the node on a server in the home network and browsing from a
laptop or phone) is a legitimate and supported use case — see the same Matrix thread
where another user (GenY2K) confirmed they can reach `:7509` from another machine on
the same network.

## Question: can we solve this without weakening the sandbox?

Adding `allow-same-origin` would fix it but is not acceptable: per the HTML spec,
combining `allow-scripts` + `allow-same-origin` lets the document programmatically
remove its own `sandbox` attribute, which defeats the entire isolation model.

The existing `WEBSOCKET_SHIM_JS` (`path_handlers.rs:1287`) already shows the
pattern: intercept the API in JS, route through `postMessage` to the parent shell
(which has same-origin access) and back. Possible directions, none of them
investigated yet:

1. **Identify what River is actually attempting.** Reproduce on a non-localhost IP
   with DevTools open and capture the call stack of the failing request. The fix
   depends on what API is hitting the same-origin check (`fetch`, `XMLHttpRequest`,
   Service Worker, dynamic `import()`, etc.).
2. **Extend the bridge.** If it's `fetch`/XHR, add a `fetch`/`XMLHttpRequest` shim
   alongside the WebSocket shim that routes through the shell, scoped to same-origin
   targets only (no open proxy).
3. **River-side fix.** If River is making a request that doesn't need to exist
   (e.g. re-fetching its own URL), drop it.

Whether the LAN-vs-localhost distinction matters depends on which API is failing —
the WebSocket bridge already works regardless of host, so if a similar bridge is the
fix, LAN should \"just work.\"

## Repro

Not yet reproduced locally. Reporter: VirtualBox VM running the freenet node,
bridged networking, browser on Win11 host accessing the VM's LAN IP. Contract is the
current River release `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`.

Probable repro recipe (untested):
- Run a node on host A, browse from host B on the same LAN to `http://A_LAN_IP:7509/`
- Open DevTools and load the published River contract
- Observe the console error and capture the failing request's call stack

## Related

- #3464 — WebSocket origin null rejection (localhost-only fix, distinct issue)
- freenet/river#217 — Playwright coverage for no-allow-same-origin sandbox
- freenet/river#219 — localStorage SecurityError in opaque origin (same root cause family)

[AI-assisted - Claude]

sanity:
## Problem

When freenet binds to `0.0.0.0` (or any non-loopback address) and a user reaches it via its LAN IP — e.g. `http://192.168.1.5:7509/` from another device on the same network, or from a host browser into a bridged VM — the page loads, but the gateway then rejects every WebSocket upgrade with `403`. Symptoms reported by fluffomat in #3957 (Chrome 147 on Win11 host → freenet on Alpine VM at `192.168.178.192:7509`):

- `http://192.168.178.192:7509/` loads fine.
- River loads inside the sandbox iframe; UI renders.
- Every WS handshake fails: `WebSocket connection to 'ws://192.168.178.192:7509/v1/contract/command?...&authToken=...' failed`.
- River reconnect-loops with exponential backoff; invitation flow hangs forever at "preparing to subscribe for room"; same symptom on Delta because the bridge is gateway-side, not River-specific.

The `Unsafe attempt to load URL ...` console line in the original report is intermittent and unrelated; the actual blocker is the WS-upgrade rejection.

## Root cause

`connection_info` (`crates/core/src/client_events/websocket.rs:653-685`) only accepts a WS upgrade when **at least one** of these holds:

1. `is_localhost_origin(origin)` — the LAN origin `http://192.168.178.192:7509` is not localhost.
2. `is_allowed_host(headers, allowed_hosts)` — `build_allowed_hosts` (`server.rs:467`) seeds the set with localhost variants + the machine hostname (and any extra `--allowed-host` entries) + `bind_addr` if not unspecified. Binding `0.0.0.0` is unspecified, so the LAN IP is **not** auto-added.
3. `host_header_ip_in_cidrs(...)` — short-circuits to `false` when `allowed_source_cidrs` is empty (the default), so a LAN user who never set `--allowed-source-cidrs` always fails this check.

All three fail → `403` → WS handshake aborts → wasm-bindgen surfaces a downstream `expected a string argument, found undefined` panic in the iframe shim's `onerror` path (the error event has no `filename`), which is what fluffomat saw at the top of the trace.

This is inconsistent with `private_network_filter` (`server.rs:621`), which already accepts any RFC 1918 / loopback / link-local / IPv6 ULA source IP by default — i.e. binding `0.0.0.0` is already an implicit opt-in to LAN access for HTTP. The WS check was the one place that disagreed.

## Solution

Drop the `allowed_cidrs.is_empty() → return false` early return and add an `is_private_ip` short-circuit to `host_header_ip_in_cidrs`. Now a Host IP is acceptable when it is **either** private **or** inside an operator-supplied CIDR; the existing Origin-must-be-an-IP-literal-matching-Host-IP guard (or `Origin: null` for sandboxed iframes) is unchanged, so CSWSH protection is preserved:

- `Origin: https://evil.com` against `Host: 192.168.1.5:7509` → `evil.com` is a hostname, not an IP → rejected.
- `Origin: http://192.168.1.50:7509` against `Host: 192.168.1.5:7509` → IPs differ → rejected.
- `Origin: http://192.168.1.5:7509` against `Host: 192.168.1.5:7509` → matches → accepted.
- `Origin: null` (sandboxed dApp iframe) → accepted, same as the existing CIDR path.

Removes `AllowedSourceCidrs::is_empty`, which is no longer called.

## Testing

- New regression test `lan_ip_accepted_without_explicit_allowlist_3957` covers fluffomat's exact scenario plus the two CSWSH variants (cross-origin attacker hostname; same-LAN attacker IP) that must still be rejected.
- Reworked `host_header_empty_cidrs_matches_nothing` → `host_header_empty_cidrs_accepts_private_rejects_public` to lock in the new semantics: loopback + RFC 1918 accepted, CGNAT (100.64/10) and public IPs still rejected without explicit CIDR.
- Updated `host_header_ipv6_in_cidr_allowed` so the "outside the CIDR" assertion uses a non-ULA IPv6 (`2001:db8::1`, documentation prefix), since ULAs are now accepted by default.
- All other existing CSWSH / parsing / IPv4-mapped-v6 tests pass unchanged.

Workaround for users on a release that doesn't have this fix yet: `freenet --allowed-host <lan-ip>` (or `allowed-host` under `[ws-api]` in `config.toml`).

Closes #3957

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -172,23 +172,36 @@ fn is_allowed_host(headers: &axum::http::HeaderMap, allowed_hosts: &HashSet<Stri
     allowed_hosts.contains(&host_header.to_lowercase())
 }
 
-/// Returns `true` if the request's `Host` header IP is inside an operator-allowed
-/// source CIDR **and** the `Origin` is compatible with it.
+/// Returns `true` if the request's `Host` header IP is acceptable for a
+/// privileged WS upgrade **and** the `Origin` is compatible with it.
 ///
-/// "Compatible" means one of:
+/// A Host IP is acceptable when it is either:
+/// - private (RFC 1918 / loopback / link-local / IPv6 ULA), matching what
+///   [`is_private_ip`](crate::server::is_private_ip) and the
+///   `private_network_filter` already accept by default for source IPs, or
+/// - inside an operator-supplied source CIDR
+///   (`--allowed-source-cidrs` / config `allowed-source-cidrs`).
+///
+/// Accepting private Host IPs by default closes #3957: when freenet binds
+/// to `0.0.0.0`, `private_network_filter` accepts an RFC 1918 source IP,
+/// but the WS upgrade then rejected every LAN client because the LAN IP
+/// wasn't in `allowed_hosts` and no CIDRs were configured. The two checks
+/// are now consistent: LAN access works out of the box, just like for
+/// plain HTTP.
+///
+/// "Compatible Origin" means one of:
 /// - `Origin: null` — sandboxed iframe served from the same node (Freenet's
 ///   default dApp sandbox intentionally strips origin).
 /// - `Origin` is an IP literal matching the same Host IP (direct browser
-///   access from a user-typed URL).
+///   access from a user-typed URL, e.g. `http://192.168.1.5:7509/`).
 ///
 /// This is strict on purpose: returning `true` here grants full WS access
 /// to the fully-privileged client API, so we require that the Origin
 /// genuinely corresponds to the Host. An attacker at `https://evil.com`
-/// whose JS targets `ws://<victim-tailnet-ip>:7509/...` sends
+/// whose JS targets `ws://<victim-lan-ip>:7509/...` sends
 /// `Origin: https://evil.com`, which parses to the hostname `evil.com` —
-/// not an IP, not matching the Host IP — so we refuse. This closes the
-/// cross-site WebSocket hijacking (CSWSH) vector flagged during review of
-/// this PR's earlier iteration.
+/// not an IP, not matching the Host IP — so we refuse. This is the
+/// primary cross-site WebSocket hijacking (CSWSH) guard.
 ///
 /// Only string-parseable IP literals are considered for Host/Origin; DNS
 /// is never resolved, which preserves the DNS-rebinding defence.
@@ -197,9 +210,6 @@ fn host_header_ip_in_cidrs(
     origin_str: &str,
     allowed_cidrs: &crate::server::AllowedSourceCidrs,
 ) -> bool {
-    if allowed_cidrs.is_empty() {
-        return false;
-    }
     let Some(host_header) = headers
         .get(axum::http::header::HOST)
         .and_then(|h| h.to_str().ok())
@@ -210,7 +220,7 @@ fn host_header_ip_in_cidrs(
         return false;
     };
     let host_ip = normalize_mapped_v6(ip);
-    if !allowed_cidrs.contains_ip(&host_ip) {
+    if !crate::server::is_private_ip(&host_ip) && !allowed_cidrs.contains_ip(&host_ip) {
         return false;
     }
     origin_compatible_with_host_ip(origin_str, host_ip)
@@ -655,8 +665,9 @@ async fn connection_info(
             if let Ok(origin_str) = origin.to_str() {
                 // Accept the upgrade when the origin is localhost, the Host
                 // header matches an explicit `allowed-host` entry, or the
-                // Host is an IP in `allowed-source-cidrs` *and* the Origin
-                // is compatible with it (null sandboxed iframe or same-IP).
+                // Host is an acceptable IP (private/LAN by default, or
+                // inside `allowed-source-cidrs`) *and* the Origin is
+                // compatible with it (null sandboxed iframe or same-IP).
                 // See `host_header_ip_in_cidrs` for the CSWSH rationale.
                 if !is_localhost_origin(origin_str)
                     && !is_allowed_host(req.headers(), &allowed_hosts)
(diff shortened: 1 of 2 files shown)


sanity:
## Multi-model review responses

Ran 4 internal reviewers (code-first, testing, skeptical, big-picture) plus codex against the original commit. Pushed `c8926ad` to address the findings.

### Skeptical review — HIGH severity, fixed in `c8926ad`

> Any malicious site can serve `<iframe sandbox>` (which sends `Origin: null`), the iframe JS opens `ws://192.168.x.y:7509/...`, and the gateway now accepts because Host is private + Origin: null is "compatible." Default-on CSWSH against any LAN install.

Real bug. The original commit treated `Origin: null` as compatible with any acceptable Host IP, which was operator-opt-in only on the previous code path. Split `host_header_ip_in_cidrs` into two branches with different Origin policies:

- **Operator CIDR** (`--allowed-source-cidrs`): unchanged — accepts `Origin: null` OR a same-IP literal. Operator opted in.
- **Default LAN/private**: accepts a **same-IP literal Origin only**. `Origin: null` is rejected.

River/Delta are unaffected: their WebSocket is created by the shell page (which has the gateway's actual LAN-IP origin), so the browser sends `Origin: http://<lan-ip>:7509` — a same-IP literal that passes the strict gate. Operators who genuinely need null-origin sandboxed-iframe access from non-localhost contexts must opt in via `--allowed-source-cidrs`.

Also addressed skeptical's medium-severity point on `0.0.0.0` / `::`: added `is_default_lan_host_ip` which excludes unspecified addresses from the default branch (no legitimate browser sends `Host: 0.0.0.0`). Pinned with `host_header_unspecified_rejected_on_default_branch`.

### Code-first review

- Function name `host_header_ip_in_cidrs` is now misleading — covered by an extensive rustdoc rewrite that enumerates the dual-branch policy and the not-covered list (CGNAT, public unicast, unspecified). Considered renaming; kept name to minimize churn since the doc covers it. Happy to rename in a follow-up if reviewers prefer.
- IPv6 zone-ID Host (`fe80::1%eth0`) parse-fails closed — verified, no behaviour change. Documenting as a known limitation can be a small follow-up.
- CGNAT call-out: added to rustdoc ("Not covered: CGNAT (100.64/10), public unicast IPv4/IPv6, IPv6 GUA, and `0.0.0.0` / `::`. Tailnet/CGNAT users still require `--allowed-source-cidrs`.").

### Testing review

- New `host_header_unspecified_rejected_on_default_branch` pins the `0.0.0.0` / `::` exclusion and confirms operator CIDRs can still cover those explicitly via `0.0.0.0/8`.
- Public IPv4 + same-IP literal Origin rejection asserted in `host_header_empty_cidrs_accepts_private_with_ip_literal_origin` (`8.8.8.8` + `http://8.8.8.8` → reject).
- Tailscale-CGNAT-still-needs-CIDR covered both in the rustdoc and in the renamed test's assertions.

### Big-picture review

- Pre-rebuilt: confirmed PR scope is tight, no tests silently weakened, no CLAUDE.md / `.claude/rules/` / docs to update (no existing reference to LAN-access guidance in `.md` files).
- Optional follow-up #1 (post a note on #3957 explaining the "Unsafe attempt to load URL" was a red herring) — done in the next comment on the issue.
- Optional follow-up #2 (release notes bullet) — no CHANGELOG file in the repo; will land naturally with the next release announcement if applicable.

### Codex CLI

`codex review --base main` returned the diff but no analysis text in this run. The four internal reviewers covered the same surface comprehensively, including the high-severity finding, so I didn't re-run it.

[AI-assisted - Claude]

PR #3081, opened 2026-02-17
fix: reduce fast_channel recv_async polling overhead by 50x

sanity:
## Problem

With 350+ connections, `fast_channel::recv_async` generates ~350K timer wakeups/sec due to its 1ms maximum backoff. Each idle channel polls once per millisecond, which overwhelms the tokio runtime event loop and causes gateway connection failures at scale (#3080).

## Solution

**Increase the exponential backoff cap from 1ms to 50ms**, reducing idle-channel wakeups from ~1000/sec to ~20/sec per channel. With 350 connections, this drops total wakeups from ~350K/sec to ~7K/sec — a 50x reduction.

Active channels are unaffected: backoff resets to 1µs on each successful receive, so latency remains sub-millisecond for connections with traffic.

Additionally, add **Notify-based sender disconnect detection**: `FastSender::Drop` fires a `Notify` that instantly wakes the receiver, replacing reliance on polling to discover disconnection.

### Why not pure Notify-based waking?

The initial implementation used `tokio::sync::Notify` to wake the receiver on every send (zero polling). While this works perfectly in production, it breaks turmoil's deterministic simulation tests because `Notify` doesn't generate timer events that turmoil uses for virtual time advancement, fundamentally altering task scheduling order and causing different simulation trajectories.

The 50ms backoff approach preserves simulation compatibility while still achieving a 50x improvement in production.

### Seed changes

Three simulation test seeds (`n3_g1_s4`, `n5_g1_s1`, `n7_g1_s4`) were borderline — they trigger a known convergence bug (#3070) that the timing change exposes. Replaced with seeds that converge reliably under both old and new backoff values. All replacement seeds verified on both main and the PR branch.

## Testing

- 10 fast_channel unit tests pass (including backoff verification, sender-drop wakeup, concurrent senders)
- 56 simulation integration tests pass, 0 fail (14 ignored per #3090)
- All replacement seeds verified working on both main and PR branch

## Fixes

Closes #3080

[AI-assisted - Claude]

PR #2336, opened 2025-12-19
fix: eliminate CPU-burning busy loop in FastChannel::recv_async

sanity:
## Problem

The production gateway was experiencing chronic high CPU usage (137%+) even when idle. This was affecting CI as the gateway runs on a GitHub runner machine.

### Investigation

Previous PRs (#2319, #2321) identified thread explosion but the root cause was not the `spawn_blocking` for bincode serialization - that was a symptom, not the cause.

Using `perf top` profiling, I found the CPU was dominated by:
- `tokio::runtime::context::scope` (8.7%)
- `tokio::sync::task::atomic_wakeup` (4.8%)
- `tokio::runtime::scheduler::mul` (4.0%)

This pattern indicated excessive task scheduling overhead - tasks constantly waking up without actual work.

### Root Cause

The `FastChannel::recv_async()` method (introduced in the transport layer optimization) was implemented as a **busy loop with `yield_now()`**:

```rust
pub async fn recv_async(&self) -> Result<T, RecvError> {
    loop {
        match self.inner.try_recv() {
            Ok(msg) => { ... }
            Err(TryRecvError::Empty) => {
                // BUG: This just yields and immediately retries!
                tokio::task::yield_now().await;
            }
            ...
        }
    }
}
```

When the channel is empty (most of the time for an idle gateway), this spins forever - `yield_now()` only yields to other tasks but immediately becomes ready to run again, consuming CPU continuously.

## Approach

Replace the busy loop with a proper async notification pattern using `tokio::sync::Notify`:

1. Add `recv_notify: Arc<Notify>` field to both sender and receiver
2. When sending a message, call `recv_notify.notify_one()` to wake waiting receivers
3. When receiving from empty channel, call `recv_notify.notified().await` to sleep efficiently
4. When all senders drop, call `recv_notify.notify_waiters()` so receivers can detect disconnection

This mirrors exactly how `send_async()` already handles backpressure for full channels using `send_notify`.

## Testing

- All existing `fast_channel` tests pass
- Added `test_recv_async_does_not_busy_loop` to verify the fix
- **Deployed to production gateway**: CPU dropped from 137% to <1%

### Why Didn't CI Catch This?

The busy loop only manifests under sustained idle conditions - tests complete quickly before CPU becomes noticeable. The `test_recv_async_does_not_busy_loop` test documents the expected behavior but doesn't measure CPU (difficult to do reliably in tests).

The real protection here is code review - the original `yield_now()` approach is a well-known anti-pattern that should have been caught.

[...]

PR #2263, opened 2025-12-11
perf(transport): add fast_channel for high-performance packet routing

iduartgomez:
## Problem

The transport layer uses `tokio::sync::mpsc` channels for internal packet routing. Performance analysis (documented in `docs/architecture/transport_perf_analysis.md`) shows that `tokio::sync::mpsc` achieves ~2.24 Melem/s while `crossbeam::channel` achieves ~9.98 Melem/s - a **4.4x difference**.

This channel overhead is a bottleneck in the transport layer's hot path for packet routing.

## Solution

Introduce `FastChannel` - a hybrid channel implementation that combines:
- **crossbeam::channel** for fast, lock-free message passing
- **tokio::sync::Notify** for async/await integration

This gives us most of crossbeam's performance while maintaining tokio compatibility.

### Key Features

- **Sync send** (`send()`) - Fastest path, uses crossbeam directly
- **Async send** (`send_async()`) - For backpressure-aware async contexts
- **Async receive** (`recv_async()`) - Primary receive method for async code
- **Bounded and unbounded** variants with proper backpressure

### API

\`\`\`rust
let (tx, rx) = fast_channel::bounded::<Packet>(100);

// Sync send (fastest)
tx.send(packet).unwrap();

// Async send (with backpressure)
tx.send_async(packet).await.unwrap();

// Async receive
let packet = rx.recv_async().await.unwrap();
\`\`\`

## Changes

1. **New module**: `crates/core/src/transport/fast_channel.rs` (433 lines)
   - `FastSender<T>` and `FastReceiver<T>` types
   - `bounded()` and `unbounded()` constructors
   - Comprehensive unit tests

2. **Benchmarks**: Added to `transport_perf.rs`
   - `fast_channel/throughput/tokio_mpsc_100` - baseline
   - `fast_channel/throughput/fast_channel_100` - async API
   - `fast_channel/throughput/fast_channel_sync_send` - sync API
   - `fast_channel/throughput/crossbeam_unbounded` - reference

## Testing

- Unit tests verify correct behavior (send/recv, backpressure, disconnection)
- Benchmarks will run in CI via the new Performance Benchmarks workflow

## Next Steps

This PR adds the `FastChannel` primitive. Future PRs can:
1. Replace `tokio::sync::mpsc` with `FastChannel` in transport hot paths
2. Measure end-to-end throughput improvement

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3960, opened 2026-04-26
fix: replace fast_channel with tokio::sync::mpsc to eliminate spin-deadlock

### LINKED ISSUE #3959: fast_channel deadlocks node in sched_yield spin loop on wedged crossbeam slot
## Symptoms

A long-running freenet peer (v0.2.49, regular client node, not a gateway) on a Linux host went unresponsive after running for several hours. Concretely:

- HTTP API on `127.0.0.1:7509` no longer accepted connections — chromium tabs that opened TCP connections sat with HTTP request bytes in the kernel receive queue (`Recv-Q` 626–686) and were never read; one socket reached `CLOSE-WAIT`.
- The process did not exit. No panic on stdout/stderr, nothing in the tracing logs after the freeze.
- Last log entry at `2026-04-26T18:30:54Z`; the process kept running for ≥ 4h after that with two threads pegged near 100% CPU.

## What the threads were doing

`strace -p <tid> -c` over 5 s on the two hot threads:

```
% time     seconds  usecs/call     calls    errors syscall
100.00    3.452421          19    173211           sched_yield
100.00    3.452217          22    155358           sched_yield
```

100% `sched_yield`, ~30 kHz on each thread. No other syscalls.

Userspace backtrace via `gdb -p`:

**TX side — `UdpPacketsListener::listen` task:**
```
#0 sched_yield ()
#1 crossbeam_channel::channel::Sender<T>::try_send ()
#2 freenet::transport::connection_handler::ConnectionStateManager::try_send_established ()
#3 freenet::transport::connection_handler::UdpPacketsListener::listen
```

**RX side — a per-peer `PeerConnection::recv` loop:**
```
#0 sched_yield ()
#1 crossbeam_channel::channel::Receiver<T>::try_recv ()
#2 freenet::transport::fast_channel::FastReceiver<T>::recv_async
#3 freenet::transport::peer_connection::PeerConnection::recv
```

Note: the `sched_yield` is inside crossbeam's `try_send`/`try_recv`, **not** in the `recv_async` outer loop (which uses `Notify` and would `await`).

## Root cause

The `sched_yield` calls come from `crossbeam_channel`'s internal `Backoff::snooze()`. Its bounded array channel uses a slot-reservation protocol:

1. Sender CAS-bumps `tail` to claim a slot.
2. Sender writes the value to the slot.
3. Sender updates the slot's `stamp` to indicate "ready".

Receivers do the symmetric dance on `head`. Both `try_send` and `try_recv` will call `Backoff::snooze()` (→ `sched_yield`) when the slot they want to operate on is in an in-flight state — i.e. the stamp has not been finalized yet. **`Backoff::snooze` has no upper bound.** If a slot ever ends up permanently in the in-flight state, every subsequent `try_send`/`try_recv` that lands on it spins forever.

That is what happened here: at least one per-peer inbound packet channel (`fast_channel::bounded(INBOUND_CHANNEL_CAPACITY)` in `connection_handler.rs:1815`) ended up in this wedged state. The listener task synchronously calls `try_send_established` from `connection_handler.rs:1312` for every inbound packet, so once the wedged channel's `try_send` enters the snooze loop, the listener stops processing **all other** inbound UDP — for every peer — and the per-peer task that owns the receiver of that wedged channel also spins. Every other tokio task on the runtime is starved, including the Axum HTTP server.

## Why this is a systemic problem, not a one-off

`fast_channel.rs` (introduced in #2263) wraps crossbeam to outperform `tokio::sync::mpsc`. It inherits crossbeam's "snooze indefinitely" semantics, and the call-site invokes `try_send` synchronously inside the listener event loop, so any per-peer wedge becomes a node-wide hang. The same module has already been patched four times for related correctness/perf issues:

- #2336 — eliminate CPU-burning busy loop in `recv_async`
- #3081 — reduce `recv_async` polling overhead by 50×
- #3208 — prevent `fast_channel` overflow by eliminating inbound starvation
- #3215 — break ACK-drop resend feedback loop causing inbound channel overflow

This is the fifth instance of the same family.

## What I have not proven

The exact precipitating event for the slot wedge. Candidates I considered, ranked:

1. A worker thread was killed/dropped between crossbeam's CAS reservation and slot-write commit. (No panic in stderr, but the systemd unit limits the service to `MemoryMax=2G` and `CPUQuota=200%`, and the process had hit the 2G memory peak. Kernel scheduling pressure or OOM-adjacent behavior could cause an unusual interleaving.)
2. A memory-ordering edge case in crossbeam under cross-thread interleavings.

A coredump of the hung process was captured (`/home/ian/freenet-hung.54626` on technic); the head/tail/stamp values can be inspected from it to confirm exactly which slot is wedged.

## Suggested fixes

**Surgical (recommended for the immediate bug):**

Wrap `FastSender::try_send` and `FastReceiver::try_recv` in `fast_channel.rs` with a spin-iteration cap. After N internal snooze iterations, treat the call as `Full`/`Empty` instead of letting crossbeam spin indefinitely. The existing call-site at `connection_handler.rs:1312` already has a `Full` branch that drops the packet and increments a per-peer drop counter — that path becomes the recovery channel for wedged slots.

**Independent improvement:**

Make the listener defensive against any single per-peer channel being persistently `Full` — escalate from per-peer drop counter to connection teardown after N consecutive failures. Today the warning fires but the connection is not torn down, so a wedged channel becomes permanently degraded even after the listener stops blocking on it.

**Bigger lever:**

Replace the crossbeam-bounded inner channel in `fast_channel` with `tokio::sync::mpsc::channel(N)`. Loses the microbench delta (the doc comment cites 2.88 vs 1.33 Melem/s), but is panic-safe and bounded-by-design with no slot-reservation/commit split.

## Repro context

- v0.2.49 (commit `35151c1b`)
- Linux client peer running as a systemd user service (`MemoryMax=2G`, `CPUQuota=200%`)
- ~4.5 h after process start the node went unresponsive in this state
- Same node had been restarted by `systemd` 4 prior times today, suggesting this is recurring

[AI-assisted - Claude]

sanity:
## Problem

A live freenet client peer (v0.2.49) on Linux went unresponsive after running for several hours: the HTTP API on `127.0.0.1:7509` stopped accepting requests, no log output for 4+ hours, but the process kept running with two tokio worker threads pegged near 100% CPU. `strace -p <tid> -c` over 5 s showed 100% `sched_yield`, ~30 kHz, on each thread. gdb caught both threads inside crossbeam-channel:

- TX: `UdpPacketsListener::listen` → `ConnectionStateManager::try_send_established` → `crossbeam_channel::Sender::try_send` → `sched_yield`
- RX: `PeerConnection::recv` → `fast_channel::FastReceiver::recv_async` → `crossbeam_channel::Receiver::try_recv` → `sched_yield`

`fast_channel.rs` wraps crossbeam's bounded array channel. That channel uses a slot-reservation protocol — a sender CAS-bumps `tail`, then writes the slot's value, then updates the slot's stamp. Receivers do the symmetric dance on `head`. When either side encounters a slot whose stamp hasn't been finalized yet, they call `Backoff::snooze()` (→ `sched_yield`) — and `Backoff::snooze` has no upper bound. Once a slot is left in a permanent in-flight state, every subsequent `try_send`/`try_recv` that lands on it spins forever.

The listener calls `try_send_established` synchronously from its event loop. A single wedged per-peer channel therefore freezes the listener for **every** peer plus the entire runtime, including the Axum HTTP server task.

`fast_channel` (introduced in #2263) has been patched four times for related correctness/perf issues: #2336, #3081, #3208, #3215. This is the fifth instance of the same family.

## Approach

Delete `fast_channel` and migrate every call site to `tokio::sync::mpsc` directly:

- tokio's mpsc has no slot-reservation/commit split, so it cannot wedge in the way crossbeam's array channel can. `try_send` returns `Full`/`Closed` immediately under all conditions.
- `recv().await` integrates with the runtime via parker/notify primitives owned by the channel, so the `Notify` wrapper in `fast_channel` is unnecessary.
- The "2 × microbenchmark advantage" of crossbeam over tokio mpsc cited in the original `fast_channel` doc never translated to production throughput — per-channel throughput is bounded by network speed and by per-packet processing in `PeerConnection::recv`, not by channel ops/sec. Trading the microbench delta for elimination of node-wide-hang tail risk is the right call.
- The `crossbeam` workspace dependency is removed (no remaining users).

API shifts:
- `FastSender<T>` → `mpsc::Sender<T>`
- `FastReceiver<T>` → `mpsc::Receiver<T>` (now needs `&mut self` for `recv()`/`try_recv()`)
- `fast_channel::bounded(n)` → `mpsc::channel(n)`
- `recv_async()` → `recv().await` (returns `Option<T>`, not `Result<T, RecvError>`; matches updated accordingly)
- `send_async(x).await` → `send(x).await`
- Sync `.send(x)` for blocking test sockets → `.blocking_send(x)`

## Testing

- New regression test `test_try_send_established_never_blocks_when_full` exercises the listener forwarding path: fills a per-peer channel with no consumer, calls `try_send_established` 1000 times, asserts the whole loop completes within 1 s. Under crossbeam-backed `fast_channel`, the wedge required unlucky thread interleaving inside crossbeam internals and was not triggerable in a unit test, but the test pins the invariant the fix establishes — `try_send` from the listener path must never block, regardless of receiver state.
- Existing 530 transport unit tests: all pass.
- Existing 2453 lib tests: all pass.
- `cargo clippy --all-targets -- -D warnings`: clean.
- Drive-by: added a missing `// SAFETY:` comment to a pre-existing `unsafe` block in `service.rs` that was tripping `clippy::undocumented_unsafe_blocks` even on `main`.

### Why didn't CI catch this?

The wedge requires a specific cross-thread interleaving inside crossbeam-channel-array — a slot reservation that completes its CAS but doesn't update the stamp before the next sender or receiver lands on it. Reproducing this synthetically would require either a custom-built test that hooks into crossbeam internals or a fault-injection layer above it. Neither was practical to add for one library's internal protocol.

The right CI gap to close is the architectural one: the listener's forwarding path must be guaranteed non-blocking. The new test pins exactly that invariant — and tokio mpsc's API guarantees `try_send` returns immediately by construction, so it cannot regress without changing channels again.

Closes #3959

[AI-assisted - Claude]


sanity:
## Review feedback summary

Ran multi-model review (4 internal Claude agents — code-first, testing, skeptical, big-picture — plus Codex). Codex found no regressions. The other four converged on three actionable categories:

### Addressed in this PR (commits 9ed63196, c154da24)

- **Stale `fast_channel` references in docs/comments** — code-simplifier swept 7 sites in `benches/transport/blackbox.rs`, `tests/simulation_integration.rs`, `docs/architecture/transport/DOCUMENTATION_GAPS.md`, and `docs/architecture/transport/benchmarking/methodology.md`. The historical perf analysis doc (`2025-12-performance-analysis-01.md`) got a "Superseded by #3960" header rather than a rewrite, since it captures decisions made at that time.
- **Add a parallel regression test for `try_send_gateway_handshake`** — same forwarding path from the listener loop, equally vulnerable to the original wedge. Added `test_try_send_gateway_handshake_never_blocks_when_full` alongside the existing test.
- **Convert regression tests to `#[tokio::test]`** — keeps them forward-compatible if `ConnectionStateManager` later uses async primitives.
- **Soften the regression-test docstring claim** — the tests do not reproduce the original wedge (which required cross-thread interleaving inside crossbeam internals); they pin the structural invariant the fix establishes. That distinction is now explicit in the docstring.

### Tracked as follow-up (#3961), not addressed in this PR

Reviewers flagged two pre-existing `.send().await` call sites in transport hot paths that violate `.claude/rules/channel-safety.md`:

- `send_nat_traversal` at `connection_handler.rs:790` (called synchronously from the listener event loop)
- `inbound_streams` fragment forwarding at `peer_connection.rs:1483` (legacy stream path inside `PeerConnection::recv`)

Both predate this PR (they were `.send_async().await` under crossbeam — same blocking semantics, just renamed by the migration). Converting them to `try_send` requires real design decisions about backpressure semantics for two distinct subsystems (drop the packet vs tear down the handshake; drop the fragment vs close the stream), not a mechanical rename. Tracked in #3961 with the suggested approach.

### Acknowledged but not changed

- **Service.rs SAFETY-comment drive-by** is a one-line clippy fix on a pre-existing `unsafe` block that was failing `cargo clippy --all-targets -- -D warnings` even on `main`. Disclosed in the original PR description.
- **No bench numbers backing the "2x microbench delta is irrelevant" claim** — softened the language; the load-bearing argument is "per-channel throughput is bounded by network speed and per-packet processing in `PeerConnection::recv`," not the microbench delta itself.
- **Pre-existing race between gateway-handshake completion and packet drain** (skeptical reviewer) — the same race exists with crossbeam, unchanged by this PR.

### Test status

- 2455 lib tests pass (was 2453; +2 for the new regression tests)
- `cargo fmt --check` clean
- `cargo clippy --all-targets -- -D warnings` clean

[AI-assisted - Claude]

PR #3962, opened 2026-04-27
fix: convert remaining .send().await in transport hot paths to try_send

### LINKED ISSUE #3961: transport: convert remaining .send().await in listener/recv loops to try_send
## Background

While reviewing #3960 (replacing `fast_channel` with `tokio::sync::mpsc` to fix the spin-deadlock #3959), four reviewers (code-first, testing, skeptical, big-picture) all flagged that two pre-existing `.send().await` call sites in transport hot paths violate the project's own [`channel-safety.md`](https://github.com/freenet/freenet-core/blob/main/.claude/rules/channel-safety.md) rule:

> NEVER use `.send().await` in event loops, recv loops, or any task whose blocking would stall other operations.

### 1. `send_nat_traversal` — listener event loop

`crates/core/src/transport/connection_handler.rs:790`

```rust
async fn send_nat_traversal(
    &self,
    addr: &SocketAddr,
    packet: PacketData<UnknownEncryption>,
) -> Result<bool, ()> {
    if let Some(ConnectionState::NatTraversal { packet_sender, .. }) = self.states.get(addr) {
        match packet_sender.send(packet).await {  // ← .send().await on bounded mpsc(100)
            Ok(()) => Ok(true),
            Err(_) => Ok(false),
        }
    } else {
        Err(())
    }
}
```

Called at line 1371 from `UdpPacketsListener::listen` for every inbound packet that maps to a NAT traversal handshake. If the per-handshake channel saturates (capacity 100), the listener's entire event loop blocks. Symptom would be identical to #3959 — node-wide hang — but via a different mechanism (legitimate buffer full + slow consumer rather than channel internals wedging).

### 2. Legacy stream fragment forwarding — `PeerConnection::recv` loop

`crates/core/src/transport/peer_connection.rs:1483`

```rust
if let Some(sender) = self.inbound_streams.get(&stream_id) {
    sender
        .send((fragment_number, payload))   // ← .send().await on bounded mpsc(64)
        .await
        .map_err(|_| TransportError::ConnectionClosed(self.remote_addr()))?;
    ...
}
```

Inside `process_inbound`, called from `PeerConnection::recv`. Comment says "Legacy path: push to mpsc channel for existing recv() behavior" — implying a modern path exists. If a per-stream consumer stalls, the entire connection's `recv()` loop blocks, breaking ACKs, retransmissions, and keepalives.

## Why these were not fixed in #3960

#3960's scope was specifically the `fast_channel` → `tokio::sync::mpsc` migration to remove the spin-deadlock failure mode. Converting these `.send().await` sites to `try_send` requires a real design decision:

- For `send_nat_traversal`: when the channel is full, drop the packet (lossy, may break NAT traversal under load) or tear down the handshake?
- For `inbound_streams.send()`: when the stream consumer stalls, drop the fragment (corrupts the stream) or close the connection?

These are not mechanical conversions; they're product decisions about backpressure semantics for two distinct subsystems. They warrant their own focused PR with appropriate tests.

## Suggested approach

1. For `send_nat_traversal`: convert to `try_send` with a per-handshake drop counter; tear down the NAT traversal state after N consecutive `Full` failures. Add a regression test that fills the channel and verifies the listener continues processing other peers.

2. For `inbound_streams`: investigate whether the legacy code path can simply be removed (the comment suggests a modern path supersedes it). If it must stay, convert to `try_send`. On `Full`, the stream is by definition broken (consumer can't keep up); close the per-stream channel and let the higher-level retransmission machinery handle it.

3. Audit any other `.send().await` in transport hot paths via `grep -rn "\.send(.*).await" crates/core/src/transport/ | grep -v "try_send\|test"`.

[AI-assisted - Claude]

sanity:
## Problem

A `.send().await` call site in the UDP listener path (`send_nat_traversal` at `connection_handler.rs:790`) violated `.claude/rules/channel-safety.md`. It runs synchronously from `UdpPacketsListener::listen`, so a slow or wedged per-handshake consumer would have stalled the listener for every other peer — exactly the failure mode that caused #3959 (4-hour node hang in production) on the established-connection forwarding path.

## Solution

Convert `send_nat_traversal` to non-blocking `try_send_nat_traversal`. On `Full`, drop the packet (intro packets are retried by the remote — `NAT_TRAVERSAL_MAX_ATTEMPTS = 40` at 200 ms cadence under a 3 s `overall_deadline`, so ~15 retries in production — so a transient channel-full event degrades latency but does not break the handshake). On `Closed`, drop the packet (handshake task already exited). Promoted the return type to a dedicated `TrySendNatTraversalOutcome { Sent, Full, Closed }` enum so the listener can log the two failure cases distinguishably.

### What changed in scope after review

The original PR also converted the per-stream `inbound_streams` forwarding in `PeerConnection::process_inbound` from `.send().await` to `try_send`. Codex (P1), skeptical reviewer (M2), and testing reviewer unanimously flagged this as wrong: the receiver there is the freenet-spawned `recv_stream` reassembly task — internal, same-runtime, with no upstream dependency on the producing recv loop — so the cascading-backpressure deadlock pattern `channel-safety.md` exists to prevent cannot form. Turning `Full` into `ConnectionClosed` would have aborted otherwise-healthy large transfers under transient executor pressure (more than 64 fragments backing up before `recv_stream` is scheduled). Reverted that change, with a load-bearing comment citing the new "Exception: same-runtime internal consumers" subsection added to `channel-safety.md`.

## Testing

- New `test_try_send_nat_traversal_never_blocks_when_full` — completes the trio with `test_try_send_established_never_blocks_when_full` and `test_try_send_gateway_handshake_never_blocks_when_full` (added in #3960). Same scaffolding: fill a per-handshake channel of capacity 4 with 1000 attempts, assert the loop completes inside a 1 s budget. Updated to use the new outcome enum.
- 553 transport unit tests pass.
- `cargo clippy --all-targets -- -D warnings` clean.
- Audit: `grep -rn "\.send(.*).await" crates/core/src/transport/` returns exactly one production hit, in the documented exception (the legacy stream forwarding). Listener path is fully non-blocking.

## Documentation

- `channel-safety.md` updated to (a) record the #3959 → #3960 → #3961 listener-forwarding incident family, and (b) add an "Exception: same-runtime internal consumers" subsection with explicit conditions, since the reverted-and-kept `.send().await` would otherwise look like an unjustified rule violation.
- `transport.md:92` updated: the prior "up to 10 over ~3s" claim was the `cfg(test)` constant. Now cites the actual production retry budget.

Closes #3961

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/channel-safety.md
+++ b/.claude/rules/channel-safety.md
@@ -22,6 +22,38 @@ stalling the entire event loop. This has caused 4+ production deadlocks:
   the event loop exceeded bridge channel capacity, deadlocking the node.
 - **Result router chain**: `result_router_tx.send().await` in the event loop
   could cascade if `SessionActor` was slow → router blocked → event loop blocked.
+- **Listener forwarding family** (#3959 → #3960 → #3961, Apr 2026):
+  the UDP listener task in `transport::connection_handler` calls
+  `try_send_*` synchronously to route every inbound packet to the right
+  per-peer channel. The original `fast_channel` wrapper used
+  `crossbeam::channel`, whose internal `Backoff::snooze` could spin in
+  `sched_yield` indefinitely on a wedged slot — taking the entire node
+  offline (#3959). #3960 replaced it with `tokio::sync::mpsc` so
+  `try_send` is guaranteed non-blocking. #3961 closed the remaining
+  `.send().await` site on the listener path (`send_nat_traversal`).
+  The recv-loop forwarding from `PeerConnection::recv` into the
+  per-stream `inbound_streams` channel was investigated and *kept* as
+  `.send().await`: the receiver there is the freenet-spawned
+  `recv_stream` reassembly task, internal and same-runtime, so the
+  cascading-backpressure pattern this rule prevents cannot form.
+
+## Exception: same-runtime internal consumers
+
+`.send().await` on a bounded channel is acceptable when **all** of the
+following hold:
+
+1. The receiver task is freenet-internal (we spawned it; it is not an
+   external client or peer).
+2. The consumer has no upstream dependency on the producer (no cycle
+   through which a stalled producer could starve its own consumer).
+3. Blocking is scoped to a single per-peer or per-connection event loop
+   (so it cannot starve other peers).
+4. Per-message consumer work is bounded and small (microseconds, not
+   "I might call into a slow external API").
+
+Document the exception inline at the call site, citing this section.
+Existing exception: `peer_connection.rs::process_inbound` legacy stream
+fragment forwarding into `inbound_streams[stream_id]`.
 
 ## Rules
 
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -89,7 +89,7 @@ BEFORE changing LEDBAT++ parameters:
 
 ```
 NAT traversal:
-  1. Send multiple intro packets (up to 10 over ~3s)
+  1. Send multiple intro packets — `NAT_TRAVERSAL_MAX_ATTEMPTS` (40 in release builds, 10 under `cfg(test)`) at a 200 ms cadence, capped by a 3 s `overall_deadline` (so ~15 attempts at production settings)
   2. Use exponential backoff: 50ms → 300ms → 1s → 5s
   3. First relay observes external address (ObservedAddress msg)
   4. Rate limit: 1 intro packet/second per source IP
--- a/crates/core/src/transport/connection_handler.rs
+++ b/crates/core/src/transport/connection_handler.rs
@@ -779,17 +779,26 @@ impl<S, T: TimeSource> ConnectionStateManager<S, T> {
         }
     }
 
-    /// Try to send a packet to an ongoing NAT traversal.
-    /// Returns Ok(true) if sent, Ok(false) if channel closed.
-    async fn send_nat_traversal(
+    /// Try to send a packet to an ongoing NAT traversal, without blocking.
+    ///
+    /// Returns `Ok(true)` if sent, `Ok(false)` if the channel is full or
+    /// closed. Crucially this is non-blocking: the listener task calls this
+    /// synchronously for every inbound packet, so a slow or wedged
+    /// per-handshake consumer cannot stall the listener for other peers.
+    /// Dropped intro packets are recoverable — the remote peer retries up
+    /// to 10 times over ~3 s during NAT traversal (see transport.md
+    /// "WHEN establishing connections"), so a transient channel-full event
+    /// degrades latency but does not break the handshake.
+    fn try_send_nat_traversal(
         &self,
         addr: &SocketAddr,
         packet: PacketData<UnknownEncryption>,
     ) -> Result<bool, ()> {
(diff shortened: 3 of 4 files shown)


sanity:
## Review feedback summary

Ran multi-model review (4 internal Claude agents — code-first, testing, skeptical, big-picture — plus Codex). The most consequential finding was unanimous across 3 reviewers:

### Reverted: `peer_connection.rs::process_inbound` `try_send` change

Codex (P1), skeptical reviewer (M2), and testing reviewer all flagged that switching the per-stream `inbound_streams` forwarding from `.send().await` to `try_send` introduces a new failure mode: turning transient executor pressure into `TransportError::ConnectionClosed`. Healthy large transfers would abort whenever more than 64 fragments back up before the spawned `recv_stream` reassembly task is scheduled. The previous `.send().await` provided correct backpressure semantics, and the receiver here is freenet-internal (not an external consumer), so the cascading-deadlock pattern `channel-safety.md` exists to prevent cannot form.

Reverted to `.send().await`, with a load-bearing inline comment citing the new "Exception: same-runtime internal consumers" subsection added to `channel-safety.md`. Without that exception documented, the call site was prima-facie violating its own rule with no justification.

### Other findings addressed

- **Skeptical M1** / Codex docstring accuracy: the prior comment cited "10 retries over ~3 s" — that's the `cfg(test)` constant. Production has `NAT_TRAVERSAL_MAX_ATTEMPTS = 40` capped by a 3 s `overall_deadline` at 200 ms cadence (~15 attempts). Updated both the Rust doc-comment and `transport.md:92` to cite the precise constants.

- **Skeptical M3**: `try_send_nat_traversal` collapsed `Full` / `Closed` into one `Ok(false)`, and the listener logged them indistinguishably. Promoted the return to a dedicated `TrySendNatTraversalOutcome { Sent, Full, Closed }` enum; the listener now logs `Full` ("channel full, dropping packet (peer will retry)") and `Closed` ("handshake task exited; dropping packet") separately.

- **Big-picture (low)**: documented the listener-forwarding incident family (#3959 → #3960 → #3961) in `channel-safety.md`, and added the "Exception: same-runtime internal consumers" subsection that the reverted call site now cites.

### Findings acknowledged but not changed

- **Code-first (Important) / Testing (Important)**: neither test exercises the listener call site directly — both pin the helper API, not the integration. Pre-existing precedent (the `try_send_established` and `try_send_gateway_handshake` tests from #3960 have the same shape). Worth a follow-up for an integration test that drives `UdpPacketsListener::listen` with a saturated per-handshake channel and verifies other peers keep flowing, but not a blocker.

- **Testing (Low) Fault-injection coverage**: a SimNetwork test that throttles per-stream consumers would be valuable but is beyond the scope of a `.send().await`-conversion PR.

- **Code-first (nit) Per-stream teardown alternative**: now moot since the legacy stream change was reverted.

### Test status

- 553 transport unit tests pass (was 553 before this iteration; one new test was already counted in the previous push).
- `cargo clippy --all-targets -- -D warnings` clean.
- `grep -rn "\.send(.*).await" crates/core/src/transport/` audit: the only remaining production hit is the now-documented exception in `peer_connection.rs::process_inbound`, with the inline comment citing the new rule subsection.

[AI-assisted - Claude]

PR #3968, opened 2026-04-28
refactor(ops): migrate sub-op GETs to task-per-tx driver

iduartgomez:
## Problem

Sub-operation GETs (subscribe's `fetch_contract_if_missing` and executor's `local_state_or_from_network`) still went through the legacy `request_get` driver. Each call pushed a `GetOp` into `OpManager.ops.get` with `is_client_initiated() == true`, keeping the GC speculative-retry block at `op_state_manager.rs:1912` load-bearing for sub-op GETs.

This blocked retiring the GET GC retry block by the same approach used for PUT in PR #3964: as long as any caller seeded `is_client_initiated() == true` `GetOp`s into the DashMap, the speculative-retry block had to keep firing for them.

## Solution

Route both sub-op GET callers through a new `get::op_ctx_task::start_sub_op_get` entry point — same retry driver as `start_client_get` but skips:
- Auto-subscribe and explicit `maybe_subscribe_child` hand-off (sub-op caller owns subscribe semantics)
- Telemetry route-event emission (sub-op shouldn't pollute telemetry)
- `send_client_result` / `result_router_tx` publication (no client waiter)

Outcome delivered through a oneshot (`SubOpGetOutcome::{Found, NotFound, Infra}`):
- Subscribe `fetch_contract_if_missing` is fire-and-forget — drops the receiver, relies on `wait_for_contract_with_timeout`'s storage-poll fallback (preserves legacy timing semantics).
- Executor `local_state_or_from_network` awaits the receiver with a 120 s timeout, replacing the legacy `op_request(GetContract)` mediator path entirely.

Retired in this slice:
- `get::start_op` (deleted)
- `get::request_get` (deleted)
- `executor::GetContract` + `ComposeNetworkMessage<GetOp>` impl (deleted)
- `RingError::EmptyRing` retained but `#[allow(dead_code)]` (only construction site was `request_get`'s no-peers branch)

`get::start_op_with_id` retained behind `#[allow(dead_code)]` for unit-test construction of `PrepareRequest`-state `GetOp`s. Phase 6 may delete it alongside the legacy `PrepareRequest` state.

After this migration, no caller pushes a client-initiated `GetOp` into `ops.get`. The GC speculative-retry block is provably dead for GET — retirement deferred to a follow-up slice (matches the PR #3964 PUT pattern).

## Testing

- 2477 lib tests pass (4 new pin tests, +4 net)
- `cargo fmt --check` clean
- `cargo clippy --tests -- -D warnings` clean

Pin tests:
- `fetch_contract_if_missing_uses_sub_op_driver` — subscribe must call `start_sub_op_get`, not `get::start_op`
- `local_state_or_from_network_uses_sub_op_driver` — executor must use `start_sub_op_get`, not `GetContract` / `op_request`
- `start_sub_op_get_entry_must_exist` — driver entry point pin
- `sub_op_driver_skips_auto_subscribe_and_maybe_subscribe_child` — sub-op driver must not call client-only side effects

All pin tests use `[a, b].concat()` runtime composition to avoid self-trigger (matches the put.rs pattern).

## Fixes

Refs #1454 (Phase 5 follow-up).

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -94,6 +94,21 @@ paths:
 > state before calling `send_and_await`), but the mechanical
 > `op_manager.push(...)` call is absent. Phase 6 of #1454 will rewrite
 > this rules file in full once all ops have migrated.
+>
+> Sub-operation GETs (subscribe's `fetch_contract_if_missing` and
+> executor's `local_state_or_from_network`) now route through
+> `operations/get/op_ctx_task.rs::start_sub_op_get` — same retry
+> driver as `start_client_get` but skips auto-subscribe,
+> explicit-subscribe hand-off, telemetry route-events, and
+> `result_router_tx` publication; outcome delivered through a
+> oneshot (`SubOpGetOutcome`). The legacy `request_get` driver and
+> `get::start_op` constructor were retired in this migration; the
+> executor's `GetContract` / `ComposeNetworkMessage<GetOp>` impl is
+> deleted. Legacy `start_targeted_op` (UPDATE-triggered auto-fetch)
+> remains on legacy. After this migration, no caller pushes a
+> client-initiated `GetOp` into `ops.get`, so the GC speculative-
+> retry block at `op_state_manager.rs:1912` is provably dead for
+> GET — retirement deferred to a follow-up slice.
 
 ## Critical Invariant: Push-Before-Send (legacy path)
 
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -99,16 +99,23 @@ paths:
 > executor's `local_state_or_from_network`) now route through
 > `operations/get/op_ctx_task.rs::start_sub_op_get` — same retry
 > driver as `start_client_get` but skips auto-subscribe,
-> explicit-subscribe hand-off, telemetry route-events, and
+> explicit-subscribe hand-off, telemetry route-events
+> (`NetEventLog::get_request` is intentionally not emitted on any
+> task-per-tx originator path — matches `start_client_get`), and
 > `result_router_tx` publication; outcome delivered through a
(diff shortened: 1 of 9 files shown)


iduartgomez:
Adding `review-override` label.

The rule-review bot's warning is that `drive_sub_op_get` itself has no direct behavioral unit test. That's accurate, but:

1. **Direct unit testing requires a real `OpManager`** with ring/contract-handler/transport plumbing. The legacy equivalents (`drive_client_get_inner`, `drive_relay_get_inner`) also have only structural pin tests + extracted-pure-helper unit tests for the same architectural reason.

2. **Transitive behavioral coverage is established** by the green CI runs:
   - Unit & Integration (9m37s pass): exercises subscribe and executor flows that call `start_sub_op_get`
   - Simulation (12m19s pass): exercises end-to-end multi-node scenarios where contracts are fetched via sub-op GET (subscribe race-resolve, executor `local_state_or_from_network` for related contracts)
   - NAT Validation (5m33s pass): exercises the same paths under transport churn

3. **The streaming `socket_addr().is_none()` edge** flagged by the bot is the same legacy code path that `drive_client_get_inner` exhibits at line 341 of the same file — both use the same `assemble_and_cache_stream` helper, both warn-and-fall-through identically. No new failure mode introduced.

4. **Pure helpers extracted and unit-tested**: `missing_state_cause`, `SubOpGetOutcome` constructors, oneshot drop semantics, tx-shape invariant. The non-pure parts (cache_contract_locally, store re-query) are shared with `drive_client_get_inner` and have their own behavioral test scaffold elsewhere.

Direct behavioral coverage of `drive_sub_op_get` would require either (a) duplicating ~400 lines of mock OpManager scaffolding, or (b) a new SimNetwork test specifically for sub-op GET. Both are disproportionate to the scope of a thin migration that reuses the already-tested `GetRetryDriver` primitives.

PR #3969, opened 2026-04-28
fix(installer): clean up Local AppData config and add visible auto-close countdown

sanity:
## Problem

Two distinct Windows installer issues reported by users on Matrix:

1. **Setup wizard appears to vanish after install.** The "Installation complete" screen auto-closes after 5 seconds via a bare `setTimeout(close, 5000)` with no on-screen indicator. Several users reported it as confusing — they couldn't tell whether install finished cleanly, the dialog just disappeared on them.

2. **`freenet uninstall --purge` leaves the live config folder behind on Windows.** After running uninstall, `%LOCALAPPDATA%\The Freenet Project Inc\Freenet\config\` is still on disk. This is the directory the running node actually writes to (see `Config::build` in `crates/core/src/config.rs:247`, which uses `defaults.config_local_dir()`). The pre-fix `purge_data_dirs` only knew about `dirs.config_dir()`, which on Windows is the *Roaming* path and is never written by current code — so every Windows uninstall left a populated `config\` directory behind. (Reported again after #3904; that fix addressed parent collapse but didn't notice the Roaming/Local mismatch.)

## Approach

**Setup-wizard countdown.** Replace the silent `setTimeout` with a per-second `setInterval` that updates a visible "Closing in Ns..." line under the success message, and rename the existing button from "Close" to "Close now" so the immediate-dismiss affordance is obvious. No backend / installer code touched.

**Uninstall config-dir cleanup.** Add a new `config_local` leaf to `DataLeaves`, populated from `dirs.config_local_dir()` and deduped against `data_local` / `data_roaming` / `config`. Keep the existing Roaming `config` cleanup as well so any old install that did write there still gets cleaned up. The dedup is important on Linux/macOS where some of these paths can coincide; on Windows they're always distinct.

## Testing

- `cargo test -p freenet --bin freenet` — all 112 tests pass, including the 5 `test_purge_leaves_*` tests.
- New regression test `test_purge_leaves_removes_config_in_local_app_data` seeds a `%LOCALAPPDATA%\...\Freenet\config\` layout and asserts the leaf and its now-empty parent both collapse. **Verified it fails** when the new `if let Some(ref config_local) = leaves.config_local` branch in `purge_leaves_and_collapse` is reverted, so it actually catches the bug.
- Existing `test_purge_leaves_and_collapse_cleans_windows_layout` updated to seed both `config` (Roaming) and `config_local` (Local) — the Windows layout now matches what the running node actually writes.
- `cargo fmt` + `cargo clippy --bin freenet -- -D warnings` clean.
- The setup-wizard JS is HTML-rendered inside WebView2 on Windows; it has no Rust unit-test surface. The change is a small JS edit (interval-driven countdown text) and is best smoke-tested on Windows, which I haven't done — flagging this as a gap. The non-WebView code path (the `escape_for_js` test) still passes.

## Why didn't CI catch this?

Both bugs are Windows-specific paths. The uninstall test layout was seeded with `config` only at the Roaming path (matching the old buggy code), so the test would have happily passed even with the bug present — the test was *modeling* the buggy assumption rather than the real on-disk layout. The new test fixes that and would catch a regression of the same shape. The setup-wizard countdown has no automated coverage; it's WebView2 HTML+JS that runs only in the installer dialog on Windows.

## Fixes

Reported on Matrix; no specific issue numbers (relates to #3904 which addressed the parent-collapse half of the same area).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2183,9 +2183,22 @@ struct DataLeaves {
     /// Pre-#3739 Roaming data path, only populated on Windows where it
     /// differs from `data_local`.
     data_roaming: Option<PathBuf>,
-    /// Config dir, only populated when it differs from both data paths
-    /// (matches the macOS case where `config_dir == data_dir`).
+    /// `config_dir` (Roaming on Windows, e.g.
+    /// `%APPDATA%\...\Freenet\config`). Only populated when it differs
+    /// from both data paths (matches the macOS case where
+    /// `config_dir == data_dir`). May be stale on Windows where the
+    /// running node actually writes to `config_local` (see #3979),
+    /// but is still cleaned up in case an old install left it behind.
     config: Option<PathBuf>,
+    /// `config_local_dir` (Local on Windows, e.g.
+    /// `%LOCALAPPDATA%\...\Freenet\config`). This is what the running
+    /// node actually writes to (`Config::build` in config.rs uses
+    /// `defaults.config_local_dir()`), which is *not* the same as
+    /// `config_dir` on Windows. Without this, `freenet uninstall
+    /// --purge` left the live config folder behind. Only populated
+    /// when distinct from `data_local` and `config` to avoid double
+    /// removal.
+    config_local: Option<PathBuf>,
     /// `cache_dir` for the uppercase project bundle.
     cache: Option<PathBuf>,
     /// Lowercase-variant cache used by the webapp cache on case-sensitive
@@ -2227,6 +2240,14 @@ impl DataLeaves {
             let config_dir = dirs.config_dir().to_path_buf();
             if config_dir != data_dir && Some(&config_dir) != leaves.data_roaming.as_ref() {
                 leaves.config = Some(config_dir);
+            }
+
+            let config_local_dir = dirs.config_local_dir().to_path_buf();
+            if config_local_dir != data_dir
+                && Some(&config_local_dir) != leaves.data_roaming.as_ref()
+                && Some(&config_local_dir) != leaves.config.as_ref()
+            {
+                leaves.config_local = Some(config_local_dir);
             }
 
             leaves.cache = Some(dirs.cache_dir().to_path_buf());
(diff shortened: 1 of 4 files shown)


sanity:
Ran the four-perspective review (code-first, testing, skeptical, big-picture) plus a Codex CLI review. Pushed 834e3b1 to address findings.

**Addressed:**
- **Bogus issue reference (`#3979`).** Big-picture review caught that I'd put references to a non-existent issue in code/test names. Replaced with `#3904 follow-up / PR #3969` references. (`gh issue view 3979` confirms it does not exist.)
- **JS interval not cancelled on manual close.** Both code-first and skeptical reviews flagged this. Hoisted `tick` to module-scope `closeCountdown` and made `doClose()` `clearInterval` before posting IPC. Prevents a possible double-close foot-gun on future WebView refactors.
- **Producer (`from_project_dirs`) was untested.** Testing-review's strongest finding — the bug actually lived in the dir-resolution producer, not the consumer that already had tests. Added `test_from_project_dirs_leaves_config_local_empty_on_unix` pinning the Linux/macOS invariant (the `directories` crate aliases `config_local_dir` to `config_dir`, so `config_local` must stay `None` to avoid double removal).
- **Wizard HTML had zero test coverage.** Testing-review flagged this as low-cost / real-value. Lifted `SETUP_HTML` out of the Windows-gated `mod platform` (now `include_str!(\"setup.html\")` at module scope) and added `test_setup_html_success_screen_has_visible_countdown`. Now runs on every CI platform, not just Windows. Mutes the resulting dead-code warnings on non-Windows with two narrowly-scoped `cfg_attr` allows.
- **Doc-comment drift.** Code-first review noted `config_local`'s rustdoc listed `data_local` and `config` for dedup but the code also dedups against `data_roaming`. Big-picture review noted `config` was undersold as "stale" when `report.rs` and `config.rs:1661` still use it. Tightened both.

**Considered and skipped:**
- *Idempotency on second `--purge`* (testing review): existing `remove_if_exists` already tolerates missing paths; covered by `test_purge_leaves_tolerates_missing_optional_leaves`. No new test needed.
- *Setup wizard not smoke-tested on real Windows* (skeptical / deployment rule): I don't have Windows access in this session. Flagging this as a manual verification step before tagging — the change is HTML/JS string-contains logic, low risk, but per `.claude/rules/deployment.md` the platform-gated path should be observed.
- *Refactor `from_project_dirs` and `Config::build` to share a single dir-resolution helper* (big-picture review #4): out of scope for a fix PR, would expand blast radius. Filing as a follow-up if the user wants it.

CI watching shortly. Codex's bottom-line: \"The changes correctly add cleanup for the Windows local config directory and include regression coverage for that path, and the setup wizard countdown change does not introduce an obvious functional regression.\"

[AI-assisted - Claude]

PR #3970, opened 2026-04-28
refactor(ops): migrate sub-op SUBSCRIBE to task-per-tx driver

iduartgomez:
## Problem

`SubOperationTracker` is the legacy DashMap-based parent/child machinery
that #1454 wants to retire. After the prior task-per-tx slices (PUT,
GET, sub-op GETs in #3968), exactly **one** production caller of
`OpManager::expect_and_register_sub_operation` remained:
`operations::start_subscription_request_internal`, reached from
`auto_subscribe_on_get_response` on the legacy GET originator branch
(the only fallback path that turns a GET response into a separate
SUBSCRIBE).

While the tracker had a single live producer, Phase 3c (mechanical
deletion of `SubOperationTracker`, `root_ops_awaiting_sub_ops`,
`failed_parents`, etc.) was blocked.

## Solution

Replace the body of `start_subscription_request` with a fire-and-forget
spawn of `subscribe::run_client_subscribe` — the same task-per-tx
driver used by `maybe_subscribe_child` in `put/op_ctx_task.rs` and
`get/op_ctx_task.rs`. Drop the `track_parent` plumbing entirely:
`expect_and_register_sub_operation` is no longer invoked from any
production caller.

The `blocking` parameter is preserved on the public function for API
stability and telemetry but is now a no-op: in the task-per-tx world,
blocking semantics come from `await`ing `run_client_subscribe`
inline (which `maybe_subscribe_child` already does where it matters).
Sub-op blocking via `SubOperationTracker` was effectively dead before
this PR — `start_targeted_op` hard-codes `subscribe: false`, and
client-initiated GETs no longer push a `GetOp` into `ops.get`, so no
production legacy GET originator path reaches
`auto_subscribe_on_get_response` with `blocking_sub=true`.

`subscribe::start_op_with_id`, `OpManager::expect_and_register_sub_operation`,
`OpManager::sub_operation_failed`, and the corresponding
`SubOperationTracker::expect_and_register_sub_operation` keep
`#[allow(dead_code)]` markers pointing at Phase 3c, which will delete
the whole tracker machinery mechanically.

## Testing

- New `operations::sub_op_subscribe_migration_pin_tests` (3 structural
  pin tests, source-scrape `start_subscription_request`) verify that:
  1. `subscribe::request_subscribe` (legacy driver) is no longer
     referenced.
  2. `expect_and_register_sub_operation` and `sub_operation_failed`
     are no longer referenced.
  3. `subscribe::run_client_subscribe` is spawned.
- All 2484 lib tests pass; clippy and fmt clean.

## Follow-ups

- Phase 3c — mechanical deletion of `SubOperationTracker` and the
  finalized-parent-parking branch in `operations.rs::handle_op_result`.
  All remaining tracker references are in legacy-only code that
  becomes unreachable after this PR.

Refs #1454.

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -116,6 +116,28 @@ paths:
 > into `ops.get`, so the GC speculative-retry block at
 > `op_state_manager.rs:1912` is provably dead for GET — retirement
 > deferred to a follow-up slice.
+>
+> Sub-operation SUBSCRIBE callers (legacy GET-originator
+> `auto_subscribe_on_get_response` fallback path; PUT/GET task-per-tx
+> drivers already routed via `maybe_subscribe_child`) now spawn
+> `subscribe::run_client_subscribe` directly — the legacy
+> `subscribe::request_subscribe` driver and the
+> `expect_and_register_sub_operation` /
+> `SubOperationTracker`-registration step are gone from the production
+> sub-op SUBSCRIBE entry point. `start_subscription_request` keeps
+> the `blocking` parameter for API stability but it is no longer
+> behaviorally distinct: blocking semantics in the task-per-tx world
+> are obtained by `await`ing `run_client_subscribe` inline (see
+> `maybe_subscribe_child` in `put/op_ctx_task.rs` and
+> `get/op_ctx_task.rs`); `start_subscription_request` is a
+> fire-and-forget spawner. The tracker (`SubOperationTracker`,
+> `expect_and_register_sub_operation`, `sub_operation_failed`,
+> `all_sub_operations_completed`, `count_pending_sub_operations`,
+> `failed_parents`, `root_ops_awaiting_sub_ops`,
+> `parent_of`) and the finalized-parent-parking branch in
+> `operations.rs::handle_op_result` are still referenced by the
+> legacy state-machine path but have no production caller after this
+> slice — Phase 3c of #1454 will delete them mechanically.
 
 ## Critical Invariant: Push-Before-Send (legacy path)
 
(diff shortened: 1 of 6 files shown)


iduartgomez:
Addressed review feedback:

- **#3, #4 (stale doc refs):** Updated `subscribe/op_ctx_task.rs` module doc and `node.rs` PUT-sub-op comment to reflect the new sub-op routing (`maybe_subscribe_child` + `run_client_subscribe`).
- **#5 (pin-test brittleness):** Replaced the `\nasync fn has_contract(` tail anchor with a brace-depth walker. Moving or renaming the next function no longer breaks the pins.
- **#2 (Arc allocation):** Documented the cost rationale at the call site. The fresh `Arc::new(op_manager.clone())` is only paid by legacy `process_message` callers that hold `&OpManager`; PUT/GET task drivers go through `maybe_subscribe_child` and pay no extra cost. Threading `Arc<OpManager>` through every legacy caller would shift the same allocation upstack without saving anything.

**Skipped:**

- **#1 (lost client error notification):** The legacy `NotifySubscriptionError { key, reason }` path fans out by **contract key** to every entry in `update_notifications[key]` — i.e., to all currently-subscribed clients on that key, not the requesting client. After an auto-subscribe-fallback failure, the requesting client typically has no listener registered yet on that key (subscription was the operation that would have registered them), so the legacy fan-out delivered errors to *unrelated* subscribers and was effectively spurious cross-client noise. The task-per-tx path's tx-keyed delivery dropping silently for an unregistered child_tx is arguably a behavior improvement, not a regression. Errors remain observable via the synthesized `tracing::warn!` in `deliver_outcome`. Happy to revisit if a deployed listener relies on this.
- **#6 (`blocking` parameter is logged-only today):** Already documented in the doc comment; will be deleted alongside the legacy `start_subscription_request` callers in Phase 3c.

PR #3983, opened 2026-04-30
fix(tracing): scale event-log trim with cap to bound write amplification

sanity:
## Problem

A user reported their Freenet peer doing a lot of disk write operations even at idle. Investigated this on the production gateway (nova) and a regular peer (technic):

| Peer | write_bytes | write syscalls |
|------|-------------|----------------|
| Gateway (heavy load) | ~9 MB/s sustained | ~11.6k/s |
| Technic peer | ~1.5 MB/s sustained | ~1.4k/s |

Almost all of that disk traffic comes from `_EVENT_LOG` compaction, not network packets or text logs. Watching the gateway's `_EVENT_LOG` (~89 MB) over 10 seconds showed it oscillating by tens of KB and dropping ~130 KB every ~8 s, then growing again. That cadence matches `LogFile::truncate_records` firing roughly every 8 seconds and rewriting the whole file to drop a tiny prefix.

The cause is in `crates/core/src/tracing/aof.rs`:
- `MAX_LOG_RECORDS = 100_000`
- `REMOVE_RECS = 1000 + BATCH_SIZE` (a flat ~1% of the cap)

Once `num_recs > MAX_LOG_RECORDS`, `truncate_records` skips `REMOVE_RECS` records at the head, copies every remaining record into `_EVENT_LOG.rewrite`, and renames over `_EVENT_LOG`. On a busy gateway with ~100k records of average ~890 bytes, that copies ~89 MB to drop ~1 MB. Steady-state amplification is `MAX_LOG_RECORDS / REMOVE_RECS` ≈ 99x.

## Approach

Make `REMOVE_RECS = MAX_LOG_RECORDS / 4 + BATCH_SIZE` so each compaction sheds ~25% of the cap. The on-disk size envelope is unchanged (still bounded by `MAX_LOG_RECORDS`), but compactions run ~25x less often with similar per-pass cost. Steady-state disk writes from this path drop by the same factor — projected ~9 MB/s → ~360 KB/s on a peer with comparable load.

Why constants instead of a structural rewrite (segment-based AOF):
- Segments are the right long-term design but they are a ~200 LOC refactor with new failure modes (segment open/close races, partial-segment recovery).
- The constant change captures most of the win in a tiny, reviewable diff.
- The `_EVENT_LOG` is only used for router model training and event aggregation; trimming a slightly larger fraction per compaction has no functional impact.

A `const _: () = assert!(REMOVE_RECS * 5 >= MAX_LOG_RECORDS, ...)` guard placed next to the constants makes anyone shrinking REMOVE_RECS below 20% of the cap a build error instead of a slow telemetry regression.

## Why didn't CI catch this?

CI did not have a write-amplification assertion on `_EVENT_LOG`. The existing `read_write_truncate` test self-references `REMOVE_RECS` and so passes against any value; nothing pinned the ratio. This PR closes that gap with the compile-time `const _` invariant and a new runtime test (`truncation_leaves_meaningful_headroom`) that asserts post-compaction `num_recs` sits at or below 80% of the cap. With the original `REMOVE_RECS = 1005` against the test cap of 10_000, the runtime test fails (`num_recs=9000 > 8000`); under the fix it passes. Verified locally by reverting the constant and re-running the tests.

## Testing

- `cargo fmt -p freenet`
- `cargo clippy -p freenet -- -D warnings` clean
- `cargo test -p freenet --lib tracing::` — 34 passed (3 existing aof tests + 1 new + 30 other tracing tests)
- Reverted `REMOVE_RECS` to `1000 + BATCH_SIZE` and reran: build failed at the `const _` guard, runtime test failed with the expected diagnostic. Restored the fix.

## Fixes

No GitHub issue was filed; reported informally on Matrix.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,11 +205,17 @@ jobs:
           # Check ALL Rust files for test additions/removals (not just crates/core/src/)
           # TEST_ATTR_RE matches the set of attributes this repo uses to mark
           # a function as a test:
-          #   #[test]                  — plain libtest
-          #   #[tokio::test]           — async tests
-          #   #[test_log::test]        — tracing-subscriber-enabled tests
-          #   #[freenet_test(...)]     — Freenet integration-test macro
-          TEST_ATTR_RE='#\[((tokio|test_log)::)?test\]|#\[freenet_test'
+          #   #[test]                                          - plain libtest
+          #   #[tokio::test]                                   - async tests
+          #   #[tokio::test(flavor = ..., worker_threads = N)] - parameterized async tests
+          #   #[test_log::test]                                - tracing-subscriber-enabled tests
+          #   #[test_log::test(tokio::test)]                   - parameterized tracing-subscriber-enabled
+          #   #[freenet_test(...)]                             - Freenet integration-test macro
+          # The character class [(\]] after `test` matches either `]` (plain
+          # form) or `(` (parameterized form). Without that, parameterized
+          # tokio::test attributes weren't counted as new tests and fix: PRs
+          # using only such tests were incorrectly rejected.
+          TEST_ATTR_RE='#\[((tokio|test_log)::)?test(\]|\()|#\[freenet_test'
           ALL_ADDED=$(git diff "$BASE"..."$HEAD" -- '**/*.rs' \
             | grep '^+' | grep -v '^+++' || true)
           ALL_ADDED_STRIPPED=$(echo "$ALL_ADDED" | sed 's|//.*||' || true)
--- a/crates/core/src/tracing/aof.rs
+++ b/crates/core/src/tracing/aof.rs
@@ -26,20 +26,49 @@ pub(super) const MAX_LOG_RECORDS: usize = 10_000;
 // surviving record to disk, so a small REMOVE_RECS forces the entire file to
 // be copied on every ~REMOVE_RECS-record append cycle. With a flat 1005
 // against the 100_000-record cap this produced ~99x write amplification,
-// observed as ~9 MB/s of sustained disk writes from compaction churn alone
-// on a busy gateway. Scaling with MAX_LOG_RECORDS keeps the trim fraction
-// constant across cfg(test) and cfg(not(test)); the +BATCH_SIZE buffer
-// preserves the original "headroom for one fresh batch" intent.
+// observed via /proc/$pid/io as ~9 MB/s of sustained disk writes from
+// compaction churn alone on a busy gateway. Scaling with MAX_LOG_RECORDS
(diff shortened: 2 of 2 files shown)


sanity:
Reviewer feedback addressed in dbdb4a1:

**From skeptical reviewer**
- Added upper-bound compile-time guard `REMOVE_RECS * 2 <= MAX_LOG_RECORDS` so the trim fraction is pinned to [20%, 50%] (lower bound was already there).
- Added compile-time guard `MAX_LOG_RECORDS % BATCH_SIZE == 0` so the regression tests' "one full batch carries num_recs from MAX to MAX + BATCH_SIZE" trigger assumption is enforced statically.
- Switched the new test arithmetic to `saturating_sub` so a future violation of the upper bound surfaces as a clear assertion message, not an underflow panic.
- Justified the 25% trim ratio in the constant comment (tradeoff between rewrite amplification and trailing on-disk router-training window).

**From testing reviewer**
- Added `sustained_load_keeps_log_bounded` test that pushes `3 * MAX_LOG_RECORDS` records to force multiple compactions and asserts `num_recs` stays inside `[MAX - REMOVE_RECS, MAX + BATCH_SIZE]`. With the original `REMOVE_RECS = 1005` constant the lower edge would land at `MAX - 1005` (90%) and the test fails; under the fix the band is 75-100%.

**From big-picture reviewer**
- Filed #3984 as the follow-up issue for segment-based AOF (the structural fix this PR explicitly defers in favour of a tactical constant change).

**CI rule-lint regex bug**
- The original `TEST_ATTR_RE='#\[((tokio|test_log)::)?test\]|#\[freenet_test'` only matched plain `#[tokio::test]`, not parameterized forms like `#[tokio::test(flavor = "multi_thread", ...)]`. Both my new tests used the parameterized form (matching the rest of the file) so the first push was rejected as "no new tests". Updated the regex to `'#\[((tokio|test_log)::)?test(\]|\()|#\[freenet_test'` so either `]` or `(` after `test` counts. Verified locally against `#[test]`, `#[tokio::test]`, `#[tokio::test(...)]`, `#[freenet_test(...)]`.

**Items left as comments rather than action**
- `std::fs::rename` two-step (`remove_file` then `rename`) failure mode is pre-existing and worth a separate audit; out of scope here.
- FILE_LOCK lock-hold-time during compaction is now strictly shorter on average (~25x less often) but the upper bound on a single compaction's lock-hold is unchanged. Worth a separate look for router-refresh-tick latency on busy gateways.
- The em-dash flag in the original PR description was hallucinated — the description body has none. Left as-is.

All 5 aof tests passing locally; clippy `-D warnings` clean.

[AI-assisted - Claude]

PR #3988, opened 2026-04-30
fix(install): read prompt from /dev/tty when stdin is piped

sanity:
## Problem

The install script published at `https://freenet.org/install.sh`
(mirrored in this file at `scripts/install.sh`) is invoked via
`curl -fsSL https://freenet.org/install.sh | sh`. When piped through
`sh`, stdin is the script source itself, not the user's terminal, so
`read -r response` at the service-install prompt silently ignores
keyboard input.

Reported by Ivvor on Matrix:
> @Ian: Freenet linux install script is broken. Streaming curl to sh
> causes it to not process keyboard input when asking if you want the
> service installed.

The user-facing copy at `freenet/web/hugo-site/static/install.sh` had
this fix applied once (commit `eff317ae`, Dec 2025) but it was lost
in `freenet/web` PRs #36 / #37 which "resynced" that file from this
canonical copy in freenet-core - which never had the fix in the first
place.

## Solution

freenet/web PR #42 (just merged) reapplies and hardens the fix in
the user-facing copy. This PR brings `scripts/install.sh` in line so
the next resync is a no-op rather than a regression.

The prompt block now:

- **Discriminates on `$0`** to tell `curl | sh` apart from
  `sh install.sh`. In the curl-pipe case, sh reads its source from
  stdin so the prompt must come from /dev/tty; in the file case,
  stdin is available for the user's answer (preserves
  `printf 'y' | sh install.sh` automation patterns).
- **Probes /dev/tty by actually opening it**
  (`{ true </dev/tty; } 2>/dev/null`) instead of `[ -r /dev/tty ]`,
  since the access check can succeed even when the subsequent
  open(2) fails with ENXIO.
- **Guards `read` against EOF** with `|| response=""` so Ctrl-D /
  closed stdin doesn't abort the script under `set -eu`.
- **Adds an env-var hint** in the skip path so users hitting the
  no-tty branch know `FREENET_NO_SERVICE=1` is the supported way to
  skip prompts in scripted installs.
- **Adds a NOTE block at the top of the file** cross-referencing
  the `freenet/web` sibling and explaining why the prompt logic
  must not be flattened back to a plain `read` (a previous fix was
  already lost once via a bulk resync).

## Testing

This is a shell-only change. Verified via a pty harness (extracts the
prompt block and exercises it with a mock `freenet` binary) across
five scenarios:

1. `curl | sh` under `script(1)` pty: reads `y` from /dev/tty,
   service installs.
2. `sh install.sh` interactive (heredoc): reads from stdin.
3. `printf 'y' | sh install.sh`: reads from piped stdin.
4. `setsid sh -c './install.sh' </dev/null`: skips prompt and prints
   the `FREENET_NO_SERVICE=1` hint.
5. `sh install.sh </dev/null`: handles EOF without aborting under
   `set -eu`.

All five scenarios pass against the modified `scripts/install.sh`.

### Why no committed regression test

freenet-core has no shell-test infrastructure today and the bug-fix
rules are wired around new `#[test]` Rust functions; adding a Rust
test that exercises a shell prompt is awkward. The testing reviewer
on the freenet/web PR recommended:

- a CI grep-guard that fails if the prompt block is flattened back
  to a plain `read`, and
- a `setsid sh -c 'cat install.sh | sh' </dev/null` smoke step,

both of which would naturally live under
`.github/workflows/install-script.yml`. That's a separate, additive
PR; this one is the focused user-facing fix.

The companion `freenet/web` PR (https://github.com/freenet/web/pull/42,
merged) covers the user-facing copy.

[AI-assisted - Claude]

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -14,8 +14,9 @@
 # freenet/web (served from https://freenet.org/install.sh). Keep the
 # two in sync. In particular, the service-install prompt below MUST
 # keep its $0 + /dev/tty handling so that `curl | sh` users can
-# answer the prompt - see freenet/web PR #42 for context. A previous
-# fix was already lost once via a bulk resync.
+# answer the prompt - see https://github.com/freenet/web/pull/42 and
+# its companion PR in this repo for context. A previous fix was
+# already lost once via a bulk resync.
 
 set -eu
 
@@ -428,8 +429,9 @@ main() {
     # stdin, $0 is the shell name itself (e.g. "sh", "bash"). In the
     # script-file case we read from stdin as before, so existing
     # automation patterns like `printf 'y\n' | sh install.sh` still work.
-    # The robust way to bypass the prompt in any invocation form is
-    # FREENET_NO_SERVICE=1.
+    # The shell-name allowlist covers shells likely to appear as
+    # `curl ... | <shell>`; users piping to a more exotic shell can fall
+    # back to FREENET_NO_SERVICE=1 to bypass the prompt.
     #
     # `{ true </dev/tty; } 2>/dev/null` is used instead of `[ -r /dev/tty ]`:
     # the access check can succeed even when the process has no
@@ -442,7 +444,7 @@ main() {
         echo ""
         response=""
         case "${0##*/}" in
-            sh|bash|dash|ash|busybox|-sh|-bash|-dash|-ash)
+            sh|bash|dash|ash|zsh|ksh|mksh|pdksh|yash|busybox|-sh|-bash|-dash|-ash|-zsh|-ksh|-mksh|-yash)
                 # Script source is on stdin (`curl | sh` form).
                 if { true </dev/tty; } 2>/dev/null; then
                     printf "Would you like to install Freenet as a system service? [y/N] "
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -9,6 +9,13 @@
 #   FREENET_INSTALL_DIR  - Installation directory (default: ~/.local/bin)
 #   FREENET_NO_SERVICE   - Set to 1 to skip service installation prompt
 #   FREENET_VERSION      - Specific version to install (default: latest)
+#
+# NOTE: This file is mirrored at hugo-site/static/install.sh in
+# freenet/web (served from https://freenet.org/install.sh). Keep the
+# two in sync. In particular, the service-install prompt below MUST
+# keep its $0 + /dev/tty handling so that `curl | sh` users can
+# answer the prompt - see freenet/web PR #42 for context. A previous
+# fix was already lost once via a bulk resync.
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


sanity:
Applied `test-exempt` label per the rule_lint escape hatch defined in `.github/workflows/ci.yml`. Justification:

This PR modifies only `scripts/install.sh`, a POSIX shell script. The repo's bug-fix policy enforcement counts new `#[test]` / `#[tokio::test]` / `#[freenet_test(...)]` Rust attributes via `git diff -- '**/*.rs'`. Since there are zero Rust changes, the gate cannot be satisfied by adding code that would actually exercise the bug; the bug is in shell prompt handling under `curl | sh`.

The change was verified manually via a pty harness (`script(1)` + `setsid`) covering five scenarios (curl|sh, sh install.sh interactive, printf 'y'|sh install.sh, no-tty, EOF). Detailed scenarios are in the PR description.

A follow-up to add proper shell-script CI infrastructure (a grep-guard plus a `setsid` smoke step under `.github/workflows/install-script.yml`) is in scope for a separate PR — freenet-core has no shell-test workflow today, and adding net-new CI infra is out of scope for this focused user-facing fix. Companion fix in freenet/web (#42) was merged with the same scope decision.

[AI-assisted - Claude]

sanity:
## Review-feedback response

Five-perspective review (code-first, testing, skeptical, big-picture, Codex CLI). Most items match what the freenet/web companion PR (#42) already addressed.

### Addressed in commit `98e11320`

| # | Source | Finding | Resolution |
|---|--------|---------|------------|
| 1 | Codex P2 / Skeptical #6 | Allowlist `sh\|bash\|dash\|ash\|busybox` re-triggers original bug for `curl \| zsh`, `\| ksh`, `\| mksh`, `\| yash`. | Expanded allowlist to cover those shells (plus their login-shell `-` prefix variants). |
| 2 | Skeptical #2 | Header NOTE references "freenet/web PR #42" by bare number; fragile across resyncs / number rolls. | Replaced with full GitHub permalink. |

### Addressed before commit `98e11320`

| # | Source | Finding | Resolution |
|---|--------|---------|------------|
| 3 | Testing review | Repo's `fix:` rule_lint gate requires new Rust `#[test]` functions; would block this shell-only PR. | Applied `test-exempt` label with justification comment, per the codified escape hatch in `.github/workflows/ci.yml:220-223`. |
| 4 | Skeptical #5 | Verify cited prior-fix SHA. | `eff397ae` confirmed on freenet/web@main. (Earlier commit body had a typo — `eff317ae`. Subsequent commit body uses the correct SHA.) |

### Deferred (worth doing, separate PRs)

| # | Source | Finding | Why deferred |
|---|--------|---------|------------|
| - | Testing #1 / Skeptical #7 / Big-picture #1 | CI grep-guard + `setsid` smoke step under `.github/workflows/install-script.yml`. | freenet-core has no shell-script CI infrastructure today; adding net-new workflow is its own PR. |
| - | Big-picture #3 / Skeptical #8 | Mirror note in `AGENTS.md` or new `.claude/rules/cross-repo-mirrors.md`. | Cross-cutting docs change; better as its own PR that touches both repos symmetrically. |
| - | Big-picture #4 | `freenet:release` skill mention of the mirror. | Skill update; separate PR. |
| - | Skeptical #8 | Add row to `~/code/freenet/.claude/rules/bug-prevention-patterns.md` covering the curl-pipe `read` pattern. | Separate PR; that file is shared across the freenet directory and deserves dedicated review. |

### Manual verification

Re-ran the pty harness against the freenet-core copy across five scenarios — all pass:
1. `curl | sh` under `script(1)` pty: reads `y` from /dev/tty, installs.
2. `sh install.sh` interactive: reads `y` from stdin (heredoc).
3. `printf 'y' | sh install.sh`: reads `y` from piped stdin.
4. `setsid sh -c './install.sh' </dev/null`: skips, prints `FREENET_NO_SERVICE=1` hint.
5. `sh install.sh </dev/null` (EOF): handles cleanly under `set -eu`.

### Note on user-visible impact (Skeptical #4)

The served `https://freenet.org/install.sh` is `freenet/web/hugo-site/static/install.sh`, not `freenet-core/scripts/install.sh`. The user-reported bug was already fixed for end users by web PR #42 (merged earlier today). This PR's role is to keep `freenet-core`'s canonical copy in sync so the next bulk-resync from this file into `freenet/web` is a no-op rather than a regression — closing the original failure mode (web PRs #36/#37 resyncing from an unfixed canonical source).

[AI-assisted - Claude]

PR #3989, opened 2026-04-30
fix(report): stringify ContractKey keys before JSON serialize

### LINKED ISSUE #3987: fix(report): NodeDiagnostics serialization fails with 'key must be a string' on 0.2.50, breaking diagnostic reports
## Problem

`freenet service report` on 0.2.50 returns no live diagnostics from the node. The new `network_status_error` field added in #3897 surfaces the underlying cause:

```
"network_status": null,
"network_status_error": "127.0.0.1:7509: Failed to serialize diagnostics: key must be a string; [::1]:7509: Failed to serialize diagnostics: key must be a string"
```

This means every diagnostic report uploaded against a 0.2.50 node has empty `network_status`, even though the node is running and the WS API is reachable on both v4 and v6. The Pi 4 report `6ZPGAD` is one example, but this affects every report from a 0.2.50 node, not just resource-constrained ones.

## Root cause

`serde_json` requires JSON object keys to be strings. The error message `key must be a string` is what `serde_json` returns when it tries to serialize a `HashMap<K, V>` (or `BTreeMap`) where `K` does not implement `Serialize` as a string. Some structure inside `NodeDiagnostics` is being serialized via `serde_json::to_string` (or equivalent) with a non-string-keyed map.

Likely candidates (need to verify):
- A `HashMap<PeerId, _>` being serialized with bincode-style derives that produce numeric or composite keys under JSON.
- A map keyed by a tuple, enum, or `Location`/`SocketAddr`.

The wire path: `freenet service report` calls `QueryConnections` (or whichever diagnostic RPC) over the local WS API, then serializes the `NodeDiagnostics` response to JSON for inclusion in the upload payload. The serialization site is in the `freenet` binary (the report command itself), not in core's WS handler.

## Impact

- All 0.2.50 diagnostic reports have empty `network_status`. We lose the most useful piece of evidence in user-submitted reports: live ring composition, per-peer state, current operations.
- Defeats most of the value of the report system that #3897 was trying to make more reliable.
- Pre-#3897 reports (older freenet-core) silently dropped this same field with no error, so we may have been losing this data for longer than we realized.

## Proposed fix

1. Find the offending map. Quickest path: add a tracing log around the `serde_json::to_string` of `NodeDiagnostics` in the report binary, run it, get the panic location, fix the type. Alternative: write a small test that round-trips a synthetic `NodeDiagnostics` through `serde_json::to_string`; that test should currently fail.
2. Either change the map key to a `String`-serializable type with `#[serde(serialize_with = ...)]`, or change the serialization to use a `Vec<(K, V)>` representation under JSON via `serde_with::As::<Vec<(_, _)>>` or similar.
3. Add a regression test in the report path: assert `serde_json::to_string(&NodeDiagnostics::sample())` succeeds for a populated diagnostic.

## Reproduction

1. Run any `freenet service report` against a 0.2.50 node.
2. Look at the resulting report on the server: `network_status` is null, `network_status_error` contains "key must be a string".

Confirmed on report `6ZPGAD` (Pi 4, 0.2.50) and reproduces against `127.0.0.1:7509` and `[::1]:7509`, so the failure is in serialization, not in network reachability.

## Related

- #3897 added the `network_status_error` field that surfaced this. Without that PR we would not have known the diagnostic was being silently dropped.

[AI-assisted - Claude]

sanity:
## Problem

Every diagnostic report uploaded from a 0.2.50 node has an empty `network_status` field as soon as the node is hosting at least one contract. The `network_status_error` field added in #3897/#3898 surfaced the underlying cause:

```
"network_status": null,
"network_status_error": "127.0.0.1:7509: Failed to serialize diagnostics: key must be a string; ..."
```

This means we have been losing the most useful piece of evidence in user-submitted reports — live ring composition, per-peer state, current operations — for every report from every 0.2.50 node, not just resource-constrained ones (e.g. report `6ZPGAD`).

The offender is `NodeDiagnosticsResponse.contract_states` in `freenet-stdlib`, which is `HashMap<ContractKey, ContractState>`. `ContractKey` derives `Serialize` as a struct (`{instance, code}`), but `serde_json` forbids non-string JSON map keys. The wire path between core and clients uses bincode (which doesn't care about key types), so the bug was invisible until the report binary tried to serialize the response to JSON for upload. Bug has been latent since the field was added to stdlib in June 2025.

## Solution

Local fix in the report binary: serialize `NodeDiagnosticsResponse` field-by-field into a `serde_json::Map`, stringifying each `ContractKey` via its existing `Display` impl (Base58 contract id). Every other diagnostic field already follows the "stringify for client API" convention (`peer_id: String`, `connected_peers: Vec<(String, String)>`, `ContractHostingEntry.contract_key: String`); `contract_states` was the lone outlier.

The helper calls `key.to_string()`, which works whether the underlying type stays `HashMap<ContractKey, _>` or a future stdlib release switches it to `HashMap<String, _>` at the source — so this code does not need to change in lockstep with a stdlib bump. A separate stdlib PR will fix the type for the next stdlib release; this PR ships the user-facing fix immediately without the stdlib release round-trip.

## Testing

Three new tests in `crates/core/src/bin/commands/report.rs`:

- `test_diagnostics_to_json_with_populated_contract_states` — reproduces the exact scenario the bug broke (>=1 contract entry). Asserts the helper succeeds, the output is valid JSON, the keys are the Base58 form (matching `ContractKey::Display`), and values round-trip.
- `test_diagnostics_to_json_with_empty_contract_states` — the no-hosting-contracts path (which would not have triggered the bug). Pinned so neither branch regresses silently.
- `test_native_serde_json_on_contract_states_fails_documents_root_cause` — asserts that calling `serde_json::to_string(&NodeDiagnosticsResponse)` directly still fails with `key must be a string`. This is a tripwire: when a future stdlib release switches the key type, this test starts failing and signals that the local workaround can be removed in favour of a direct `to_string` call.

Why CI didn't catch this previously: there was no JSON round-trip test for `NodeDiagnosticsResponse` in either freenet-stdlib or freenet-core. The wire path (bincode) is the one production code uses; JSON serialization only enters the picture in the report binary, which had no unit coverage for this serialization step. The third test above closes that gap at the boundary.

All 18 report-module tests pass; `cargo clippy --bin freenet --all-targets -- -D warnings` is clean; `cargo fmt` is clean.

## Fixes

Closes #3987

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/report.rs
+++ b/crates/core/src/bin/commands/report.rs
@@ -11,7 +11,8 @@ use flate2::write::GzEncoder;
 use freenet::config::ConfigPaths;
 use freenet::tracing::tracer::get_log_dir;
 use freenet_stdlib::client_api::{
-    ClientRequest, HostResponse, NodeDiagnosticsConfig, NodeQuery, QueryResponse, WebApi,
+    ClientRequest, HostResponse, NodeDiagnosticsConfig, NodeDiagnosticsResponse, NodeQuery,
+    QueryResponse, WebApi,
 };
 use serde::{Deserialize, Serialize};
 use std::fs;
@@ -758,8 +759,7 @@ async fn query_node_diagnostics(host: &str, port: u16) -> Result<String> {
 
     match response {
         HostResponse::QueryResponse(QueryResponse::NodeDiagnostics(diag)) => {
-            // Serialize the diagnostics to JSON for the report
-            serde_json::to_string_pretty(&diag).context("Failed to serialize diagnostics")
+            diagnostics_to_json(&diag)
         }
         HostResponse::ContractResponse(_)
         | HostResponse::DelegateResponse { .. }
@@ -767,6 +767,54 @@ async fn query_node_diagnostics(host: &str, port: u16) -> Result<String> {
         | HostResponse::Ok
         | _ => anyhow::bail!("Unexpected response from node"),
     }
+}
+
+/// Serialize `NodeDiagnosticsResponse` to JSON for the report payload.
+///
+/// `contract_states` is keyed by `ContractKey`, whose derived `Serialize`
+/// impl emits a struct (`{instance, code}`). `serde_json` rejects non-string
+/// JSON map keys, so a direct `serde_json::to_string(&diag)` fails with
+/// "key must be a string" as soon as the node hosts at least one contract.
+/// We rebuild the field with stringified keys before serializing. The
+/// rest of the struct serializes natively. See freenet/freenet-core#3987.
+fn diagnostics_to_json(diag: &NodeDiagnosticsResponse) -> Result<String> {
+    use serde_json::{Map, Value};
+
+    let mut contract_states = Map::with_capacity(diag.contract_states.len());
+    for (key, state) in &diag.contract_states {
+        let value =
+            serde_json::to_value(state).context("Failed to serialize contract state value")?;
+        contract_states.insert(key.to_string(), value);
+    }
+
+    let mut value = Map::new();
+    value.insert(
+        "node_info".to_string(),
+        serde_json::to_value(&diag.node_info).context("Failed to serialize node_info")?,
+    );
+    value.insert(
+        "network_info".to_string(),
+        serde_json::to_value(&diag.network_info).context("Failed to serialize network_info")?,
+    );
+    value.insert(
+        "subscriptions".to_string(),
+        serde_json::to_value(&diag.subscriptions).context("Failed to serialize subscriptions")?,
+    );
+    value.insert(
+        "contract_states".to_string(),
+        Value::Object(contract_states),
+    );
+    value.insert(
+        "system_metrics".to_string(),
+        serde_json::to_value(&diag.system_metrics).context("Failed to serialize system_metrics")?,
+    );
+    value.insert(
+        "connected_peers_detailed".to_string(),
+        serde_json::to_value(&diag.connected_peers_detailed)
+            .context("Failed to serialize connected_peers_detailed")?,
+    );
+
+    serde_json::to_string_pretty(&Value::Object(value)).context("Failed to serialize diagnostics")
 }
 
 fn format_bytes(bytes: u64) -> String {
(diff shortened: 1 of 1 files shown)


sanity:
Companion stdlib PR: https://github.com/freenet/freenet-stdlib/pull/70 fixes the same bug at the type-system source. This PR does **not** depend on the stdlib PR — the workaround here uses `key.to_string()`, which works whether the underlying type is `HashMap<ContractKey, _>` (today) or `HashMap<String, _>` (after the stdlib bump). Either can land first.

[AI-assisted - Claude]

iduartgomez (approved):

sanity:
Review pass complete. Four reviewers (code-first, testing, skeptical, big-picture) + Codex; convergent finding from three of them was the field-drift risk in the hand-rolled field assembly. Addressed in 320a5d3:

- **Field-drift safety (skeptical-reviewer's recommendation, also raised by code-first and testing reviewers)**: added a destructure-bind on `NodeDiagnosticsResponse` at the top of `diagnostics_to_json`. Adding a field to the stdlib struct now produces a compile error here instead of silently dropping the new field from report payloads. Zero runtime cost.

- **Typo coverage (testing-reviewer's recommendation)**: added `test_diagnostics_to_json_all_fields_populated_round_trip` with every `Option` `Some`, every `Vec` non-empty, every value distinguishable, and an assertion that there are exactly six top-level keys. Catches the kind of typo the destructure cannot (e.g. accidentally inserting `node_info` twice while dropping `system_metrics`).

Other findings, all minor / non-blocking, with rationale:

- **Codex**: no discrete actionable bugs.
- **Big-picture reviewer** flagged that the original PR description's claim 'this code does not need to change in lockstep with a stdlib bump' is true for `report.rs` but the producer-side `p2p_protoc.rs:1900,1916` will need `.to_string()` updates if/when the companion stdlib PR (https://github.com/freenet/freenet-stdlib/pull/70) lands. Compiler will catch those sites at the bump; calling it out here for the record.
- **Skeptical reviewer**: the bug-pinning test asserts on the substring `\"key must be a string\"`, which is brittle to serde_json wording changes. Acceptable since it is explicitly documented as a tripwire, with comment telling the next reader exactly what to do (remove the helper).
- **Code-first reviewer**: noted that `ContractKey::Display` drops `code_hash` (vs the broken serializer which would have included it). This matches the convention every other diagnostic field uses (`ContractHostingEntry.contract_key: String` does the same), and `fdev`'s diagnostics renderer already calls `.to_string()` on the key. Consistent with existing reporting; not an information-loss regression.
- **Big-picture reviewer** suggested adding a row to `~/code/freenet/.claude/rules/bug-prevention-patterns.md` for the JSON-non-string-key class. Done in local agent rules; not part of this PR (file is local-only, not in the repo).

All 19 report tests pass; clippy + fmt clean.

[AI-assisted - Claude]

PR #3990, opened 2026-04-30
fix: bound wasmtime per-instance virtual memory reservation

### LINKED ISSUE #3986: wasmtime mmap fails on low-RAM ARM (Raspberry Pi 4 4GB): contract updates fail with 'mmap failed to reserve 0x104000000 bytes'
## Problem

A user (`neil`) tried to run freenet 0.2.50 on a Raspberry Pi 4 with 4GB RAM (aarch64). The node starts and connects, and read-only contracts like the Freenet website mirror work, but stateful contracts (River, Delta) cannot apply updates and never persist state. User-visible behaviour: "River didn't save its state and Delta was doing likewise. I could get the Freenet website mirror up though."

Diagnostic report: `6ZPGAD` (filed 2026-04-30).

## Root cause

Every contract UPDATE re-instantiates the contract WASM module. wasmtime's default on-demand allocator reserves a 4 GiB linear-memory region per instance. On a 4GB Pi the kernel rejects the mmap as soon as a couple of instances are alive concurrently:

```
ERROR freenet::operations::update: Failed to update contract value
  contract=5ZJjDnBMNaeZtTxof2EPA8RzLshLfFNzWkhHXwf8kTHe
  error=execution error: instantiation: mmap failed to reserve 0x104000000 bytes phase="error"
WARN  freenet::operations::update::op_ctx_task: UPDATE relay (task-per-tx):
  delta apply failed, sending ResyncRequest ... event="delta_apply_failed"
WARN  freenet::node: Unexpected response to resync update ... event="resync_failed"
  response=UpdateResponse { new_value: Err(ExecutorError { ..., fatal: false }), state_changed: false }
```

`0x104000000` = 4,362,076,160 bytes (4 GiB + a guard page). Repeats hundreds of times per minute in the report. The cascade of `Contract state not found in store` warnings observed in the same window is downstream from the same failure.

## Why pooling was not chosen

`crates/core/src/wasm_runtime/engine/wasmtime_engine.rs:809` notes that `InstanceAllocationStrategy::Pooling` was rejected because it pre-reserves ~4 GiB per slot, which made the virtual address space situation worse on CI runners. That reasoning is correct for x86_64 servers, but it leaves no escape hatch for low-RAM ARM hosts where on-demand allocation is the failing path.

## Proposed fix

Expose wasmtime memory tuning via freenet config and ship sane defaults for low-memory hosts.

Two knobs (either alone is probably enough, both is better):

1. `wasmtime_config.memory_reservation(...)` (or the older `static_memory_maximum_size`) lowered to something like 64 MiB instead of the 4 GiB default. Contracts that genuinely need >64 MiB linear memory are vanishingly rare and we already enforce a per-instance limit via `HostState`.
2. `wasmtime_config.memory_guard_size(...)` reduced from the default 2 GiB.

Suggested defaults:
- Detect <8 GiB total RAM at startup (e.g. via `sysinfo`) and auto-apply the low-memory profile (64 MiB reservation, small guard).
- Always allow override via config: `[contract] wasm-memory-reservation = "64MiB"` etc.

## Workarounds for affected users today

In order of cheapness:

1. Add an 8 GiB swap file. The mmap is virtual, so swap can satisfy the kernel even if it is never paged in.
2. `sudo sysctl vm.overcommit_memory=1` (and persist in `/etc/sysctl.conf`).
3. Use a host with >=8 GiB RAM.

We should document at least (1) and (2) in the manual under a "Running on Raspberry Pi" section regardless of whether (1) the config knob lands, since users will hit this on any small-RAM Linux host.

## Repro

- Pi 4 4GB (aarch64) running Linux 6.x
- freenet 0.2.50 in `mode = "network"`
- Subscribe to any River room or Delta document and observe an UPDATE arriving from the network. Within a few updates the executor will start emitting the mmap error and refusing to apply state.

## Code pointers

- `crates/core/src/wasm_runtime/engine/wasmtime_engine.rs:781` — `create_engine`, where the `wasmtime::Config` is built. This is where the new knobs would live.
- `crates/core/src/wasm_runtime/engine/wasmtime_engine.rs:795-812` — existing comment block explaining the pooling vs on-demand decision; should be updated alongside any change.

[AI-assisted - Claude]

sanity:
## Problem

A user (`neil`) tried to run freenet 0.2.50 on a Raspberry Pi 4 with 4 GB RAM (aarch64). Read-only contracts (Freenet website mirror) work fine, but stateful contracts like River and Delta cannot apply updates and never persist state:

```
ERROR freenet::operations::update: Failed to update contract value
  contract=5ZJjDnBMNaeZtTxof2EPA8RzLshLfFNzWkhHXwf8kTHe
  error=execution error: instantiation: mmap failed to reserve 0x104000000 bytes
```

Every contract UPDATE re-instantiates the contract WASM module. Wasmtime's defaults reserve 4 GiB of linear memory + 32 MiB of guard region per instance on 64-bit hosts. Both are virtual mappings, but the kernel still rejects `mmap()` once total reservations exceed the system's overcommit limit. On a 4 GB Pi the kernel rejects the second or third concurrent instance.

## Solution

Bound both wasmtime knobs to the cap we already enforce:

- `memory_reservation` → `DEFAULT_MAX_MEMORY_PAGES * WASM_PAGE_SIZE` (256 MiB)
- `memory_guard_size` → `WASM_PAGE_SIZE` (64 KiB, one wasm page)

**Why this is safe:** `HostState`'s `ResourceLimiter` already rejects any `memory.grow` past `DEFAULT_MAX_MEMORY_PAGES` (256 MiB). Reserving a 4 GiB address space beyond that was always wasted virtual memory — no contract could ever use it. Matching the reservation to the actual cap reduces per-instance virtual reservation by ~16x without changing semantics for contracts.

**Why the small guard is fine:** Bounds-check elimination via large guards only kicks in when the reservation covers the full 32-bit wasm address space (4 GiB). With a 256 MiB reservation that optimization is already disabled, so guard size has no additional perf impact. ResourceLimiter remains the source of truth for the memory cap.

**No config knobs added** — the system does the right thing on whatever machine it's running on. Detecting low-RAM hosts and applying a separate profile would just mean two code paths to maintain; bounding to the actual cap universally is simpler and works everywhere.

## Testing

Added `test_instance_virtual_memory_reservation_bounded_3986`: creates 4 live instances and asserts VmSize growth stays under 4 GiB. Uses the same subprocess pattern as `test_store_refresh_reduces_vm_maps` so concurrent tests don't corrupt the `/proc/self/status` measurement.

Verified the regression behavior locally:

| State | Result |
|---|---|
| Without fix | 4 instances added **16,640 MiB (~4.06 GiB/instance)** → fails the 4 GiB bound |
| With fix | Test passes, growth ~1 GiB total |

The 16,640 / 4 = 4,160 MiB per instance matches wasmtime's default 4 GiB reservation + 32 MiB guard exactly.

All 2,474 freenet lib tests pass; 177 `wasm_runtime::` tests including the new regression test pass.

## Fixes

Closes #3986

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
+++ b/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
@@ -74,7 +74,8 @@
 //!
 //! Additional runtime protections:
 //!
-//! - **2GB guard region** – protects against sign-extension bugs
+//! - **Guard region** – protects against sign-extension bugs (sized to
+//!   one wasm page, see "Memory Management" below)
 //! - **Stack overflow guard pages** – abort on stack overflow
 //! - **Memory zeroing** – clears memory between instantiations
 //! - **Spectre mitigations** – bounds check protection with dynamic memory
@@ -84,10 +85,13 @@
 //!
 //! ## On-Demand Instance Allocation
 //!
-//! Uses wasmtime's default on-demand allocation — each instance gets its own
-//! mmap'd virtual memory region (~4 GiB guard pages), allocated at instantiation.
-//! Instance memory is only freed when the Store itself is dropped; see
-//! `STORE_REFRESH_THRESHOLD` and `replace_store()` for periodic reclamation.
+//! Uses wasmtime's on-demand allocation — each instance gets its own
+//! mmap'd virtual memory region, allocated at instantiation. The
+//! reservation is sized to match `DEFAULT_MAX_MEMORY_PAGES` (256 MiB)
+//! rather than wasmtime's 4 GiB default, keeping virtual memory usage
+//! manageable on low-RAM hosts (#3986). Instance memory is only freed
+//! when the Store itself is dropped; see `STORE_REFRESH_THRESHOLD` and
+//! `replace_store()` for periodic reclamation.
 //!
 //! ## Compact Code Generation (Cranelift)
 //!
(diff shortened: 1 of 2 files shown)


sanity:
## Review feedback addressed (commit 65875173)

Ran 4 internal reviewers (code-first, testing, skeptical, big-picture) + Codex CLI in parallel. Findings consolidated below — all addressed in the follow-up commit unless explicitly justified.

### Fixed

- **Stale `STORE_REFRESH_THRESHOLD` rustdoc** (big-picture, skeptical, code-first): the comment cited "4 GiB guard region per instance" and "500 × 4 GiB = 2 TB" budget rationale. Rewrote to reference the bounded reservation; the 500 cap itself stands for `vm.max_map_count` budgeting.
- **`docs/architecture/contracts/README.md`** (big-picture): two stale claims — "2GB guard region before linear memories" and "wasmtime's default on-demand allocation". Both rewritten to reflect the bounded reservation and the explicit-bounds-checks regime.
- **Test bound too generous** (testing, skeptical, code-first): `GROWTH_BOUND_BYTES` lowered from 4 GiB to `2 × N × DEFAULT_MAX_MEMORY_PAGES * WASM_PAGE_SIZE` (2 GiB). A 2x per-instance reservation regression now fails the test, not just a full revert.
- **Defense in depth** (skeptical, Codex): added `memory_reservation_for_growth(0)`. Wasmtime ignores `memory_reservation` when a module declares an initial memory larger than the reservation and falls back to `min + growth-reservation` (default 2 GiB). `ResourceLimiter` rejects such modules first, but the limiter is now load-bearing for the cap, so eliminating the 2 GiB ghost reservation is cheap insurance.
- **Constants drift test** (testing): added `test_memory_tuning_constants_match_resource_limiter_cap` — a cheap, all-platform unit test that pins both byte counts. Catches drift on macOS/Windows where the subprocess-based VmSize test cannot run.
- **ResourceLimiter coverage gap** (testing): added `test_oversized_initial_memory_rejected_3986`. Instantiates a `(memory 5000)` module (5000 pages = 312 MiB initial, above the 4096-page cap) and asserts the limiter rejects it. Locks in the property that the cap covers the implicit grow-from-zero at instantiation, which the bounded reservation now relies on.
- **Misleading "sign-extension" bullet** (code-first): the module rustdoc said the guard "protects against sign-extension bugs" — true at 32 MiB, dishonest at 64 KiB. Reworded to reflect that explicit bounds checks (already emitted at this reservation size) are the actual defence; the one-page guard is just a tiny-offset backstop.

### Considered but not done

- **macOS-specific regression test** (testing): not added separately; the new constants test runs on every platform and locks in the same invariant. Adding a second platform-specific test would duplicate coverage without catching anything new.
- **"Running on Raspberry Pi" manual section** (issue #3986, big-picture): not added. The fix removes the need for the workarounds (swap, `vm.overcommit_memory=1`) that section was supposed to document; adding it would describe a problem the user no longer has.
- **Test extending `ensure_memory` up to the cap** (testing): the new oversized-memory test exercises the cap from above; the existing growth tests cover small grows. A "grow exactly to cap" test would add little: the limiter logic is `desired > limit`, so the boundary is already covered by the rejection test (one byte over cap fails) and the existing grow tests (well under cap succeeds).

[AI-assisted - Claude]

PR #3992, opened 2026-04-30
refactor(tracing): segment-based AOF for _EVENT_LOG

### LINKED ISSUE #3984: Replace _EVENT_LOG full-file rewrite with segment-based AOF
## Background

PR #3983 reduced disk write amplification from `_EVENT_LOG` compaction by trimming a meaningful fraction of records per pass instead of ~1%. That dropped sustained gateway disk writes from ~9 MB/s to a projected ~360 KB/s.

The fix is tactical: the underlying compaction strategy in `crates/core/src/tracing/aof.rs::truncate_records` still rewrites every surviving record on each pass, just less often. For a fully bounded amortised cost we should switch `_EVENT_LOG` to a segment-based AOF.

## Goal

When `num_recs` exceeds `MAX_LOG_RECORDS`, drop the oldest segment file (`O(1)` rename + unlink) instead of rewriting all surviving records into a new file.

## Sketch

- Open a current "active" segment file (`_EVENT_LOG.000`, `_EVENT_LOG.001`, ...).
- When the active segment hits a per-segment record limit, close it and open the next.
- When total segment count exceeds the cap, delete the oldest segment.
- `read_all_events` and `get_router_events` concatenate segments in age order on read.
- Recovery: at startup, scan segment filenames, recompute total `num_recs` from each segment's record count.

## Acceptance

- Steady-state disk writes from `_EVENT_LOG` compaction scale with append rate, not record count
- `read_all_events` / `get_router_events` produce the same record set as the current implementation
- Crash-recovery path tested: kill node mid-segment-write, verify reopened state is consistent
- Per-pass disk I/O bounded by segment size (e.g., 5_000 records ~= 5 MB), not full log size

## Out of scope

- Changing event types or serialization format
- Migrating existing `_EVENT_LOG` files (one-time upgrade is fine; the log is router-training data, not authoritative state)

[AI-assisted - Claude]

sanity:
## Problem

PR #3983 reduced `_EVENT_LOG` compaction write amplification from ~99x to a low multiple by trimming a meaningful fraction of records per pass. That dropped sustained gateway disk writes from ~9 MB/s to a projected ~360 KB/s — a tactical fix.

The underlying compaction strategy in `truncate_records` still rewrites every surviving record on each pass. With `MAX_LOG_RECORDS = 100_000` and `MAX_RECORDS_PER_SEGMENT` ratios in the 5% range, the steady-state cost is 100_000 records read+written per ~25_000 record interval. We can do better.

## Approach

Switch `_EVENT_LOG` to a segment-based AOF:

- The log lives in a sequence of files `<base>.NNNNNN` (zero-padded segment index). A fresh install creates `<base>.000000` and appends to it.
- When the active segment hits the per-segment record cap (`MAX_RECORDS_PER_SEGMENT = 5_000`), we close it, open the next segment, and — while the total on-disk record count exceeds `MAX_LOG_RECORDS` — drop the oldest segment with `unlink`. **The hot path becomes O(1) regardless of how full the log is.**
- `read_all_events` and `get_router_events` concatenate segments in age order on read. `get_router_events` keeps the historical "first N records visited" semantics, so router-training behavior is unchanged.
- Recovery on open scans segment filenames, counts records via a partial-record-tolerant pass, and truncates a torn trailing record (crash mid-write) so the next append starts from a clean offset.
- Legacy single-file `_EVENT_LOG` files are migrated into `<base>.000000` on first open. Empty stubs (created by `config.rs` at first-time setup) are unlinked silently.

Steady-state retention: oscillates between `MAX_LOG_RECORDS - MAX_RECORDS_PER_SEGMENT` (right after a drop) and `MAX_LOG_RECORDS + MAX_RECORDS_PER_SEGMENT - 1` (right before the next drop) — a 5% band at the prod ratio. The compile-time guards on the constants pin this so future tuning can't accidentally let either bound drift.

## Why this design over alternatives

- **Why not just keep growing per-segment cap?** Larger segments amortize the *active-segment fsync* better but don't change the per-pass I/O bound. The unlink-vs-rewrite distinction is what bounds I/O.
- **Why include the legacy migration?** Production `_EVENT_LOG` files exist in every gateway's data dir. Without migration, a freshly-deployed segment-aware build would discard accumulated router-training history. Migration is one rename.
- **Why truncate torn records on open?** A partial trailing record (header but missing payload) would otherwise either crash subsequent reads (length goes past EOF) or corrupt the parser when new appends land after the torn body. Truncate-on-open is the standard AOF recovery move.

## Testing

`cargo test -p freenet --lib -- tracing::aof` (9 tests, all pass):

- **Adapted** the three existing AOF tests (`read_write`, `read_write_small`, `read_write_truncate`, `sustained_load_keeps_log_bounded`) to the new retention semantics — one segment dropped on overflow, not `REMOVE_RECS` records.
- **`segment_rotation_drops_oldest`** — pushes records past the cap and asserts segment 0 is unlinked while later segments survive.
- **`recovery_reopens_segments`** — closes the log, reopens, and asserts `num_recs` and segment count survive a restart.
- **`partial_record_repair_on_open`** — appends a torn record (header claiming a 256-byte payload, no payload) and asserts the next open truncates the segment back to the last complete record. This is the explicit crash-mid-segment-write case from the issue.
- **`legacy_single_file_migration`** — writes a payload at `<base>` (legacy format), opens the log, asserts it was renamed to `<base>.000000` and reads cleanly through `get_router_events`.
- **`empty_legacy_stub_is_removed`** — covers the `config.rs` first-boot empty-file case so the stub doesn't take up a segment slot forever.

`cargo test -p freenet --lib` — all 2482 lib tests pass. `cargo clippy -p freenet --all-targets -- -D warnings` clean. `cargo fmt --check` clean.

## Acceptance (from the issue)

- ✅ Steady-state disk writes from `_EVENT_LOG` compaction scale with append rate, not record count — the hot path is `unlink(2)`.
- ✅ `read_all_events` / `get_router_events` produce the same record set as the previous implementation (modulo retention boundary, which now falls on segment edges instead of `REMOVE_RECS`).
- ✅ Crash-recovery path tested — `partial_record_repair_on_open`.
- ✅ Per-pass disk I/O bounded by segment size (~5 MB at `MAX_RECORDS_PER_SEGMENT = 5_000`).

Closes #3984

[AI-assisted - Claude]

### DIFF
--- a/crates/core/examples/analyze_event_logs.rs
+++ b/crates/core/examples/analyze_event_logs.rs
@@ -228,7 +228,20 @@ fn parse_args(args: &[String]) -> Result<Vec<(PathBuf, Option<String>)>> {
                     // Check if it's a directory containing event logs
                     if path.is_dir() {
                         let event_log = path.join("_EVENT_LOG_LOCAL");
-                        if event_log.exists() {
+                        // Accept either the legacy single-file log or any
+                        // segment-AOF file (`_EVENT_LOG_LOCAL.NNNNNN`) from
+                        // the segment-based format.
+                        let has_segments = std::fs::read_dir(&path)
+                            .map(|rd| {
+                                rd.filter_map(Result::ok).any(|entry| {
+                                    entry
+                                        .file_name()
+                                        .to_string_lossy()
+                                        .starts_with("_EVENT_LOG_LOCAL.")
+                                })
+                            })
+                            .unwrap_or(false);
+                        if event_log.exists() || has_segments {
                             let label = path
                                 .file_name()
                                 .and_then(|n| n.to_str())
--- a/crates/core/src/tracing/aof.rs
+++ b/crates/core/src/tracing/aof.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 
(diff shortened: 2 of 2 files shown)


sanity:
Addressed all review findings (4 internal review agents + Codex). Pushed `df2bdbc`.

**Real bugs fixed:**
- `SEGMENT_INDEX_DIGITS = 6` capped representable indices at 999_999. At 1_000_000 the scan filter (which required exactly 6 digits) silently rejected newly-created segments → silent data loss after long uptime. Padded to 10 digits (full u32).
- `rotate_segment` took `self.file` before the overflow check, leaving it `None` permanently and panicking the next `persist_log`. Reordered.
- `migrate_legacy` placed legacy files *after* existing segments — but in a partial-migration mid-state legacy is *older* than segments. Refuses with a clear error instead.
- `read_all_events` skipped pre-migration legacy single-file logs, breaking the offline aggregator (`analyze_event_logs`) running against an untouched legacy log. Now reads the legacy file when no segments exist.
- Both read paths (`read_all_events`, `get_router_events`) silently swallowed scan failures as "empty history". Propagated.
- `analyze_event_logs --paths` only auto-discovered the legacy stub; updated to detect `<base>.NNNNNN` segment files too.

**New tests:**
- `read_all_events_spans_segments` — direct read-path coverage across multiple segments
- `read_all_events_reads_legacy_without_migration` — offline-tooling case
- `partial_header_is_truncated_on_open` — distinct torn-record case (mid-header vs full-header)
- `zero_byte_stray_segment_is_recovered` — mid-rotation crash recovery
- `migrate_refuses_when_legacy_and_segments_coexist` — locks in the refusal behaviour

**Notes**
- Doc comment on `rotate_segment` updated: `FILE_LOCK` does serialize all reader/writer paths, so the `NotFound`-skip branches in the read paths are defensive (operator pruning, external tools), not a guard against our own concurrent rotation.
- Both new read-path tests pre-validate bincode roundtrip on generated `EventKind` values — `EventKind`'s `arbitrary` derive can produce values that don't roundtrip cleanly, a pre-existing quirk the route-only assertions in `read_write*` happened to mask.

[AI-assisted - Claude]

PR #3997, opened 2026-05-01
fix: align fdev/freenet binstall metadata with release assets

### LINKED ISSUE #3995: fdev / freenet crates.io publishes lack [package.metadata.binstall] — cargo binstall falls through to source compile
## Summary

`cargo binstall fdev` and `cargo binstall freenet` both fall through to a source build because the published `fdev` (e.g. 0.3.214) and `freenet` crates on crates.io don't carry a `[package.metadata.binstall]` block pointing at the prebuilt tarballs that ship on the freenet-core GitHub releases.

This forces CI users to one of:

1. `cargo install --force fdev` — ~10–15 min source build per cache miss.
2. `cargo binstall ...` — falls through to the same source build.
3. `curl …/freenet-core/releases/download/<tag>/<bin>-…tar.gz` — direct, fast, but bypasses the binstall ecosystem entirely.

## Impact

Downstream projects that depend on `fdev` / `freenet` in CI pay 10+ minutes per cold cache. We hit this on `freenet/mail`'s CI and worked around it with option 3, but it's fragile (hard-codes the GitHub release URL pattern, doesn't track crates.io versions). See https://github.com/freenet/mail/pull/55 for the workaround.

## Failure mode (option 2)

```
$ cargo binstall -y --force --version 0.3.201 fdev
…
WARN cargo binstall couldn't find prebuilt artifacts for fdev 0.3.201
…
[falls back to cargo install]
note: there are multiple different versions of crate `freenet_stdlib`
   --> .../fdev-0.3.201/src/wasm_runtime/user_events.rs
```

The fallback compile then fails because `fdev` 0.3.201 pulls `freenet-stdlib ^0.6.1` while the consuming workspace pins 0.4.0, producing the duplicate-version error above.

## Suggested fix

Add a `[package.metadata.binstall]` block to the `fdev` and `freenet` crate manifests pointing at the GitHub release tarballs. Roughly:

```toml
[package.metadata.binstall]
pkg-url = "https://github.com/freenet/freenet-core/releases/download/v{ version }/{ name }-{ target }.{ archive-format }"
pkg-fmt = "tgz"

[package.metadata.binstall.overrides.x86_64-pc-windows-msvc]
pkg-fmt = "zip"
```

…with the version mapping table (crates.io fdev 0.3.x ↔ freenet-core release tag) maintained somewhere central, since the two version namespaces diverged.

Alternatively — publish the prebuilt tarballs under a naming convention that *does* match crates.io versions (e.g. `fdev-v{crate-version}-…tar.gz` alongside the existing `freenet-core-v0.2.50` tarballs).

## Repro

```sh
cargo binstall -y --force --version 0.3.214 fdev
# observe: falls through to cargo install, ~10min source build,
# may also fail if your environment pins a different freenet-stdlib.
```

Happy to send a PR if there's interest in the metadata-block approach.

sanity:
## Problem

\`cargo binstall fdev\` and \`cargo binstall freenet\` both fall through to a 10-15 minute source build because the published metadata points at release assets that do not exist:

1. **fdev**: \`pkg-url\` template \`v{ version }\` expands to fdev's crate version (e.g. \`v0.3.214\`), but the GitHub release tag tracks the freenet version (\`v0.2.51\`). The two namespaces diverged and the URL has 404'd ever since:
   ```
   $ curl -sIL .../releases/download/v0.3.214/fdev-x86_64-unknown-linux-musl.tar.gz | head -1
   HTTP/2 404
   $ curl -sIL .../releases/download/v0.2.51/fdev-x86_64-unknown-linux-musl.tar.gz | head -1
   HTTP/2 302
   ```
2. **Both crates**: no Windows override. The release uploads \`*-x86_64-pc-windows-msvc.zip\`, not \`.tar.gz\`, so binstall would 404 on Windows even after fixing the tag.

Downstream impact (per the issue): \`freenet/mail\` CI hit this and worked around it by hard-coding the GitHub release URL pattern, which is fragile.

## Approach

- **\`crates/fdev/Cargo.toml\`**: embed the matching freenet release tag literally (\`v0.2.51\`) in \`pkg-url\` instead of \`v{ version }\`. Add a Windows override pointing at the \`.zip\` asset.
- **\`crates/core/Cargo.toml\`**: keep \`v{ version }\` (it correctly matches the release tag for freenet) but add the missing Windows \`.zip\` override.
- **\`scripts/release.sh\` + \`.github/workflows/release.yml\`**: when bumping the freenet version, rewrite the embedded \`vX.Y.Z\` segment in fdev's \`pkg-url\` so successive releases stay in sync. The sed is idempotent and safely covers both the default and Windows-override URLs.

Considered alternatives (separate \`fdev-v0.3.X\` release tags, or collapsing the two version namespaces) but both are larger workflow changes for the same effective outcome — this fix keeps the existing release process intact and just plugs the metadata into it.

## Testing

Two new regression tests, both of which would have caught the original bug:

- **\`crates/fdev/tests/binstall_metadata.rs\`** (3 tests):
  - \`pkg_url_embeds_freenet_release_tag\`: parses fdev's Cargo.toml, asserts the binstall \`pkg-url\` (default + Windows override) embeds the same \`vX.Y.Z\` that fdev's freenet dependency declares.
  - \`windows_override_uses_zip_archive\`: asserts the Windows override sets \`pkg-fmt = "zip"\` and a \`.zip\` URL.
  - \`pkg_url_does_not_use_crate_version_template\`: explicit guard against anyone "simplifying" back to \`v{ version }\`.
- **\`crates/core/tests/binstall_metadata.rs\`**: asserts the freenet manifest's Windows override exists with \`pkg-fmt = "zip"\`.

Both test files passed locally. Also manually verified the release.sh / release.yml sed expressions: substitutes both default and Windows-override URLs to a new version, no-op when the version already matches.

**Note**: fix is forward-only. The already-published \`fdev 0.3.214\` and \`freenet 0.2.51\` cannot be retroactively fixed on crates.io; the next release picks up the corrected metadata.

## Fixes

Closes #3995

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -106,7 +106,11 @@ jobs:
           # literally (`v0.2.X`) rather than `v{ version }`. Re-point both the
           # default and the Windows override to the new release tag so that
           # `cargo binstall fdev` resolves to the actual GitHub release assets
-          # (issue #3995).
+          # (issue #3995). The `/g` flag is load-bearing — there are two
+          # pkg-url lines that both need the rewrite. Mirror in
+          # `scripts/release.sh`; tests in
+          # `crates/fdev/tests/binstall_metadata.rs::release_sed_rewrite`
+          # keep both copies of the regex in lockstep.
           sed -E "s|releases/download/v[0-9]+\.[0-9]+\.[0-9]+/fdev-|releases/download/v${VERSION}/fdev-|g" crates/fdev/Cargo.toml > crates/fdev/Cargo.toml.tmp3
           mv crates/fdev/Cargo.toml.tmp3 crates/fdev/Cargo.toml
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,6 +101,15 @@ jobs:
           sed "s/freenet = { path = \"..\/core\", version = \".*\" }/freenet = { path = \"..\/core\", version = \"$VERSION\" }/" crates/fdev/Cargo.toml > crates/fdev/Cargo.toml.tmp2
           mv crates/fdev/Cargo.toml.tmp2 crates/fdev/Cargo.toml
 
+          # The fdev crate version diverged from the freenet release tag, so
+          # `[package.metadata.binstall].pkg-url` embeds the freenet version
+          # literally (`v0.2.X`) rather than `v{ version }`. Re-point both the
+          # default and the Windows override to the new release tag so that
+          # `cargo binstall fdev` resolves to the actual GitHub release assets
+          # (issue #3995).
+          sed -E "s|releases/download/v[0-9]+\.[0-9]+\.[0-9]+/fdev-|releases/download/v${VERSION}/fdev-|g" crates/fdev/Cargo.toml > crates/fdev/Cargo.toml.tmp3
+          mv crates/fdev/Cargo.toml.tmp3 crates/fdev/Cargo.toml
+
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -20,6 +20,11 @@ pkg-url = "{ repo }/releases/download/v{ version }/freenet-{ target }.tar.gz"
 pkg-fmt = "tgz"
 bin-dir = "freenet"
 
+[package.metadata.binstall.overrides.x86_64-pc-windows-msvc]
+pkg-url = "{ repo }/releases/download/v{ version }/freenet-{ target }.zip"
+pkg-fmt = "zip"
+bin-dir = "freenet.exe"
+
 [[bin]]
 name = "freenet"
 path = "src/bin/freenet.rs"
--- a/crates/core/tests/binstall_metadata.rs
+++ b/crates/core/tests/binstall_metadata.rs
@@ -1,4 +1,4 @@
-//! Regression test for issue #3995: the freenet crate's binstall metadata
+//! Regression tests for issue #3995: the freenet crate's binstall metadata
 //! lacked a Windows override, so `cargo binstall freenet` 404'd on
 //! `x86_64-pc-windows-msvc` (the release uploads `freenet-…-msvc.zip`, not
 //! `.tar.gz`).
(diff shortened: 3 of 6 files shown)


sanity:
## Review response

Multi-model review (4 internal agents + Codex CLI) is complete. All findings addressed in d97720cb.

### Critical / High
- **Testing reviewer** flagged that the sed expressions in `release.sh` / `release.yml` weren't directly exercised — manifest tests only catch drift *after* a broken release ships. → New `release_sed_rewrite` module runs both BRE and ERE forms against a fixture and asserts byte-identical output + idempotence.
- **Skeptical reviewer #3** (workspace drift): fdev's freenet dep could drift from `crates/core/Cargo.toml`'s version, leaving everything self-consistent but pointing at a release that doesn't contain this fdev. → New `freenet_dep_matches_workspace_freenet_version` test.

### Medium
- **`bin-dir` not asserted** (testing + skeptical): without `bin-dir = "*.exe"` on Windows, binstall extracts the archive but can't locate the binary. → `bin_dir_uses_correct_executable_name` tests in both crates.
- **Core test thinness** (testing): freenet test file was 1 test vs fdev's 3. → Brought to 3 tests (default URL template, Windows zip override, bin-dir).
- **Load-bearing `/g` flag** (skeptical #5): the global flag is what causes both pkg-url lines to get rewritten. → Documented in both `release.sh` and `release.yml` with a pointer to `release_sed_rewrite` for lockstep guarantee.

### Acknowledged but not addressed in this PR
- **End-to-end binstall smoke test** (testing): a nightly job that runs `cargo binstall freenet@latest fdev@latest` against real GitHub releases. Worth doing — filing as a separate follow-up issue rather than expanding this PR's scope.
- **Yanking 0.3.214 / 0.2.51** (skeptical #7): release-engineering call, not a code change. The PR description already flags the fix as forward-only.
- **Pre-release suffix handling** (skeptical #1): `release.yml`'s validate step already rejects non-`X.Y.Z` versions, so pre-releases can't reach the rewrite via the workflow. Added a doc comment is overkill for an unreachable path.
- **Extracting shared script for the two seds** (skeptical #4): bigger refactor, and the new `release_sh_and_release_yml_produce_identical_output` test catches drift, which was the underlying concern.
- **`freenet-release` skill update** (big-picture, optional): nice to have but the in-tree comments and tests already document the invariant.

### Codex CLI review
> The change consistently adds the missing Windows binstall metadata, updates fdev's release-tag handling in both release paths, and includes regression tests covering the new invariants. I did not find a discrete introduced bug that would likely warrant a fix before merging.

### Big-picture review
> No blockers. Recommend merge.

Tests: 9 fdev + 3 freenet, all green. fmt + clippy clean.

[AI-assisted - Claude]

PR #4009, opened 2026-05-02
refactor(dashboard): read subscribed contracts from canonical ring state

sanity:
## Problem

The local-peer dashboard's "Subscribed Contracts" panel mirrored
subscription state in `network_status::subscribed_contracts`, populated
by `record_subscription()` callbacks from the SUBSCRIBE op. When
SUBSCRIBE was migrated to the task-per-tx driver (PRs #3806 → #3981),
no client-initiated path called the recording hook anymore — so the
panel silently showed "No active subscriptions" for every node, even
while UPDATEs were flowing through and River's room contract was
clearly subscribed.

Verified on `technic` before fixing: dashboard reported 0 subscribed
contracts while UPDATE counter was 7.

## Approach

Rather than re-add the missing callback (the same gap would reopen on
the next migration that touches subscribe), eliminate the parallel
mirror entirely: the dashboard now reads from a `SubscriptionProvider`
closure that production wires to `Ring::dashboard_subscription_snapshot()`,
which reads `HostingManager.active_subscriptions` directly. Subscription
state has exactly one home.

The lease value type was a bare `Instant` (the expiry); it's now a
`SubscriptionLease { subscribed_since, expires_at, last_updated }`,
with `subscribed_since` preserved across renewals so the dashboard
shows continuous duration rather than resetting every 2 minutes when
the lease renews.

The dashboard pipeline:
- `Ring` exposes `dashboard_subscription_snapshot()` and
  `record_contract_update()` (both passthroughs to `HostingManager`).
- `network_status::set_subscription_provider(Arc<dyn Fn -> ...>)` is
  wired from `p2p_impl::run_node` after `init()`. The provider is held
  in a `RwLock<Option<...>>` (not `OnceLock`) so multi-node in-process
  test harnesses can re-wire to the live node — and so a second
  registration *replaces* the first rather than silently retaining the
  stale one (which is precisely the silent-failure mode this refactor
  exists to remove).
- `get_snapshot()` calls the provider; with no provider registered
  (e.g. unit tests that don't boot a node), the contracts list is
  empty.

Removed the parallel state:
- `subscribed_contracts: HashMap<String, ContractInfo>` field on
  `NetworkStatus`
- `ContractInfo` struct
- `record_subscription`, `record_subscription_removed`,
  `record_contract_updated` — including `record_subscription_removed`
  calls from `unsubscribe()` and `expire_stale_subscriptions()` (no
  longer needed; the canonical map *is* the source).

Migrated call sites:
- `runtime.rs::commit_state_update` and the new
  `initial_state_installed` site (added in #4007) now call
  `op_manager.ring.record_contract_update`.
- The dead `record_subscription` line at the legacy
  `subscribe.rs::SubscribeMsg::Response` arm is removed; that path
  hadn't been reached by client-initiated subscribes since #3981. The
  arm itself stays — its `ring.subscribe(*key)` call still does real
  work for any future caller.

User-visible behavior changes:
- Sort order in the dashboard "Subscribed Contracts" list. Old code
  put never-updated entries first (`unwrap_or(u64::MAX)` + ascending
  sort); new code puts most-recently-updated first, never-updated
  last, with a deterministic byte-key tie-break for entries with
  identical `last_updated_secs`. This eliminates a UX glitch where
  rows would reshuffle on every 5-second dashboard poll due to
  DashMap iteration order leaking through.
- The "subscribed for X seconds" reading no longer resets every 2
  minutes (when the lease renews). It now reflects the continuous
  subscription duration since first subscribe.

## Testing

Unit tests added (in `crates/core/src/`):

- `ring/hosting.rs::dashboard_snapshot_reflects_active_subscriptions`
  — pins that `subscribe()` is visible in the snapshot directly,
  covers `record_contract_update`, multi-subscription, unsubscribe,
  and the "update on non-subscribed = no-op" semantic.
- `ring/hosting.rs::dashboard_snapshot_preserves_subscribed_since_across_renewals`
  — pins that lease renewal does not reset `subscribed_since`.
- `ring/hosting.rs::dashboard_snapshot_sort_is_deterministic_on_ties`
  — pins the byte-key tie-break for entries with identical
  `last_updated_secs`. Re-polls 5× to confirm stability across
  refreshes.
- `node/network_status.rs::test_subscription_provider_wired_renders_contracts`
  — pins the **production seam**: registers a provider, asserts the
  rendered `ContractSnapshot` field shape (`key_short`, `key_full`,
  `subscribed_secs`, `last_updated_secs`), and verifies that
  re-registering replaces the previous provider.

Local validation:
- `cargo fmt -p freenet` clean
- `cargo check -p freenet --tests` clean (clippy has a pre-existing
  warning in `transport.rs:941` unrelated to this PR)
- `cargo test -p freenet --lib ring::hosting::` — 62 passed (was 59;
  +3 new), 5 ignored (pre-existing)
- `cargo test -p freenet --lib node::network_status` — 19 passed
- `cargo test -p freenet --lib server::home_page` — 30 passed

### Why didn't CI catch this?

The pre-existing `network_status::test_subscription_tracking` only
tested the `record_subscription` mirror functions in isolation; it
never exercised the path from `HostingManager::subscribe` through to
`get_snapshot()`. After #3981 finished migrating every client-initiated
SUBSCRIBE off the legacy path, that unit test still passed (the mirror
functions still worked) while the production pipeline was broken end
to end. The new `test_subscription_provider_wired_renders_contracts`
exercises the seam as it's actually used, so a future regression
where `set_subscription_provider` is forgotten or wired wrong fails CI.

## Follow-up

The same silent-mirror class of bug almost certainly affects
`network_status::record_op_result` for SUBSCRIBE and UPDATE on the
task-per-tx originator path: only PUT and GET task-per-tx drivers call
`record_op_result`, so the dashboard's `op_stats.subscribes` and
`op_stats.updates` counters are silently stuck on whatever the legacy
state-machine path produces. Out of scope for this PR; will file a
separate issue.

## Fixes

Closes the silent-dashboard regression observed in production.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -1249,7 +1249,11 @@ where
                                 event = "initial_state_installed",
                                 "Contract initial state installed"
                             );
-                            crate::node::network_status::record_contract_updated(&format!("{key}"));
+                            // Dashboard "last updated" telemetry; no-op if
+                            // we're not subscribed to this contract.
+                            if let Some(op_manager) = &self.op_manager {
+                                op_manager.ring.record_contract_update(&key);
+                            }
                             if let Err(err) = self
                                 .send_update_notification(&key, &params, &incoming_state)
                                 .await
@@ -1916,8 +1920,12 @@ where
             "Contract state updated"
         );
 
-        // Record update timestamp for dashboard display
-        crate::node::network_status::record_contract_updated(&format!("{key}"));
+        // Record update timestamp for dashboard display. No-op if we're
+        // not subscribed (e.g., a relay forwarding an UPDATE for a
+        // contract this peer doesn't track).
+        if let Some(op_manager) = &self.op_manager {
+            op_manager.ring.record_contract_update(key);
+        }
 
         if let Err(err) = self
             .send_update_notification(key, parameters, new_state)
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -9,12 +9,13 @@ use std::net::SocketAddr;
 use std::sync::{Arc, OnceLock, RwLock};
 use std::time::Instant;
 
-use crate::ring::PeerKeyLocation;
+use crate::ring::{PeerKeyLocation, Ring};
 use crate::router::Router;
 use crate::transport::metrics::TRANSPORT_METRICS;
 
 static NETWORK_STATUS: OnceLock<Arc<RwLock<NetworkStatus>>> = OnceLock::new();
 static ROUTER: OnceLock<Arc<parking_lot::RwLock<Router>>> = OnceLock::new();
+static RING: OnceLock<Arc<Ring>> = OnceLock::new();
 
 /// Store a reference to the Router for the dashboard.
 pub fn set_router(router: Arc<parking_lot::RwLock<Router>>) {
@@ -28,6 +29,14 @@ pub(crate) fn get_router() -> Option<Arc<parking_lot::RwLock<Router>>> {
     ROUTER.get().cloned()
 }
 
+/// Store a reference to the Ring so the dashboard can read canonical
+/// subscription / hosting state directly instead of mirroring it.
+pub fn set_ring(ring: Arc<Ring>) {
+    // OnceLock::set returns Err if already initialized; this is expected on repeated calls
+    #[allow(clippy::let_underscore_must_use)]
+    let _ = RING.set(ring);
+}
+
 /// Tracked network connection status for diagnostic display.
 pub struct NetworkStatus {
     pub gateway_failures: Vec<GatewayFailure>,
@@ -38,8 +47,6 @@ pub struct NetworkStatus {
     pub gateway_addresses: HashSet<SocketAddr>,
     /// Active peer connections.
     pub connected_peers: Vec<ConnectedPeer>,
-    /// Subscribed contracts keyed by encoded contract key.
-    pub subscribed_contracts: HashMap<String, ContractInfo>,
     /// Freenet version string.
     pub version: String,
     /// This node's ring location.
@@ -59,13 +66,6 @@ pub struct ConnectedPeer {
     pub location: Option<f64>,
     pub connected_since: Instant,
     pub peer_key_location: Option<PeerKeyLocation>,
-}
-
-/// Info about a subscribed contract.
-pub struct ContractInfo {
-    pub key_encoded: String,
-    pub subscribed_since: Instant,
-    pub last_updated: Option<Instant>,
 }
 
 /// Counters for each operation type: (success, failure).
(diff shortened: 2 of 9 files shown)


sanity:
Ran the four-perspective internal review (code-first, testing, skeptical, big-picture) plus Codex CLI in parallel. Convergent findings addressed in 90ed21d:

**Critical / blocks merge** (4 of 5 reviewers)
- *Wiring test gap*: replaced \`set_ring(Arc<Ring>)\` with \`set_subscription_provider(Arc<dyn Fn -> ...>)\`. New unit test \`test_subscription_provider_wired_renders_contracts\` pins the production seam end-to-end (registers a provider, asserts the rendered \`ContractSnapshot\` field shape, exercises provider replacement). The test would have caught the original drift before merge.

**Important** (Codex P2 + skeptical M2)
- *OnceLock silent-replacement*: switched from \`OnceLock<Arc<Ring>>\` to \`RwLock<Option<SubscriptionProvider>>\`. In-process multi-node test harnesses now see the live node's subscriptions instead of a stale first one. This was the exact silent-failure pattern this refactor exists to remove — using \`OnceLock\` here would have re-introduced it at the new seam.

**Important** (testing + big-picture)
- *Sort-tie nondeterminism*: tightened \`dashboard_subscription_snapshot\` comparator to byte-key tie-break for entries sharing \`last_updated_secs\`. Pre-fix, DashMap iteration order leaked through and reshuffled rows on every 5s dashboard poll. New regression test \`dashboard_snapshot_sort_is_deterministic_on_ties\` re-polls 5× to confirm stability.

**Nice-to-have**
- *Code-first*: added comment on the \`Entry::Occupied\` arm in \`subscribe()\` noting the deliberate preservation of \`subscribed_since\` and \`last_updated\` across renewals.
- *Code-first / big-picture*: PR description updated to call out the sort-order change and the full removal scope (including \`record_subscription_removed\` calls in \`unsubscribe\` / \`expire_stale_subscriptions\`).

**Follow-up filed (out of scope)**
- *Big-picture*: same bug class affects \`record_op_result\` for SUBSCRIBE and UPDATE on the task-per-tx path. Filed #4010 with reproduction, audit grep evidence, and a suggested fix path that reuses the \`SubscriptionProvider\` pattern from this PR.

**Acknowledged but not addressed in this PR**
- *Skeptical L3* — \`record_contract_update\` writes to expired-but-not-yet-swept leases. Skeptical called it "Low" / "probably benign"; would require \`TimeSource\` injection on \`HostingManager\` (which uses \`InstantTimeSrc\` directly today, a pre-existing project DST-rule gap unrelated to this PR).
- *Testing nice-to-haves*: lease-expiry-in-snapshot test (needs the same TimeSource refactor), concurrent-subscribe-with-record_contract_update test (low risk, hard to make non-flaky).

[AI-assisted - Claude]

PR #4011, opened 2026-05-03
fix: record SUBSCRIBE/UPDATE op_stats counters from task-per-tx drivers

### LINKED ISSUE #4010: Dashboard op_stats counters for SUBSCRIBE and UPDATE silently stuck
## Problem

The local-peer dashboard's "Operations" panel shows per-op success/failure counts via `network_status::OperationStats`, populated by `record_op_result(op_type, success)` calls scattered through the operations modules. Audit grep:

\`\`\`
grep -rn "record_op_result" crates/core/src/operations/
\`\`\`

returns **only two production callsites**:

- \`crates/core/src/operations/get/op_ctx_task.rs:427\` (GET task-per-tx driver)
- \`crates/core/src/operations/put/op_ctx_task.rs:342\` (PUT task-per-tx driver)

There is **no** call from \`subscribe/op_ctx_task.rs\` or \`update/op_ctx_task.rs\`. After PRs #3806 → #3981 migrated client-initiated SUBSCRIBE / sub-op SUBSCRIBE / SUBSCRIBE renewal / executor auto-subscribe to task-per-tx, and #1454 Phase 4 migrated client-initiated UPDATE, the SUBSCRIBE/UPDATE dashboard counters lost their primary writer. They report whatever the legacy state-machine path produces, which is rarely if ever hit on the originator side anymore.

This is the same class of bug as #4009 (dashboard "Subscribed Contracts" silently empty) — a telemetry mirror function that depended on being called from every active op path, but stopped being called when those paths migrated. #4009 fixed the contracts list by removing the mirror entirely and reading from canonical state. The op_stats counters need the same treatment, or a similar "compute from canonical state" reformulation.

## Investigation needed

- Are there any SUBSCRIBE / UPDATE callsites for \`record_op_result\` left at all in production code paths? If they exist on the legacy state-machine path, are they reachable in practice?
- What is the canonical source for "this op succeeded / failed"? Likely the publish path on the task-per-tx drivers (the spot that emits \`HostResult::Ok\` / \`Err\` to the result router).
- Should the dashboard counters be computed from event-log aggregates instead of mirrored counters?

## Suggested fix

Three options, in increasing scope:

1. **Minimal**: add \`record_op_result\` calls at the SUBSCRIBE and UPDATE task-per-tx terminal arms (mirror what GET/PUT do). One-line change; reopens the same drift class on the next migration but matches how GET/PUT are currently instrumented.
2. **Better**: have each op driver register success/failure in a canonical place (e.g. \`OpManager.record_outcome\`) and have the dashboard derive its counters from there. Removes the mirror but adds a new abstraction.
3. **Best**: have the dashboard read aggregated counts from the event log / a \`SubscriptionProvider\`-style closure, the same pattern #4009 uses for the contracts list. Long-term, several other \`record_*\` mirrors in \`network_status\` (\`record_update_received\`, \`record_nat_attempt\`, etc.) likely deserve the same treatment.

## Bug class context

See the \"Follow-up\" section of #4009 and \`.claude/rules/bug-prevention-patterns.md\` for the broader pattern: telemetry mirror functions that depend on being called from every active code path silently rot when those paths migrate. The fix from #4009 (read from canonical state via a registered provider closure) is reusable.

[AI-assisted - Claude]

sanity:
## Problem

The local-peer dashboard's "Operations" panel counters for SUBSCRIBE
and UPDATE were silently stuck at zero on every node. After the
task-per-tx migrations (`#3806` / `#3981` for SUBSCRIBE,
`#1454` Phase 4 for UPDATE), the only remaining writer for these
counters was the legacy `report_result` path in `node.rs`, which is
bypassed for task-per-tx terminal replies:
`handle_pure_network_message_v1` returns `Ok(None)` once it forwards
the reply to the awaiting task, so `report_result` never sees the
classification and never calls `record_op_result`.

GET and PUT survived the equivalent migration because their drivers
explicitly call `record_op_result` in the Done arm of the retry loop.
SUBSCRIBE and UPDATE drivers were missed.

This is the same class of telemetry-rot bug as `#4009` (subscribed
contracts panel silently empty after the SUBSCRIBE migration).

## Solution

Add the recording call to `deliver_outcome` in both
`subscribe/op_ctx_task.rs` and `update/op_ctx_task.rs`. Every
terminal client-visible result flows through this single chokepoint:
- `Publish(Ok(_))` -> success
- `Publish(Err(_))` -> failure
- `SkipAlreadyDelivered` (subscribe only, local hit) -> success
- `InfrastructureError` -> failure

Internal-only operations stay uncounted by design: subscription
renewals (`run_renewal_subscribe`) and executor auto-subscribe
(`run_executor_subscribe`) have separate return-value delivery paths
and never call `deliver_outcome`. Sub-operation SUBSCRIBE spawned by
PUT/GET also doesn't go through `deliver_outcome`. This is a
deliberate behavior change vs. the legacy state-machine path, which
counted renewals and inflated the counter with 2-minute background
traffic per active subscription.

Documented the manual-mirror pattern on `record_op_result` itself so
future migrations have a single place that lists the required call
sites.

## Why didn't CI catch this?

The pre-existing `test_op_stats_recording` only exercised
`record_op_result(OpType::Get/Put, _)` directly. It never followed
the path from a real `deliver_outcome` call into the counter, so
when the task-per-tx drivers stopped invoking the recording hook the
unit test still passed end-to-end while the dashboard pipeline broke.

The new tests close that gap two ways:
- Source-grep pin tests in `subscribe/op_ctx_task.rs` and
  `update/op_ctx_task.rs` assert that `deliver_outcome` calls
  `record_op_result(OpType::{Subscribe,Update}, _)`. A future
  migration that drops the call fails CI immediately, with a
  message that points back at this issue.
- `test_op_stats_recording` extended to cover SUBSCRIBE and UPDATE
  counters at the runtime level so a regression in `record_op_result`
  itself also surfaces.

## Testing

- `cargo fmt -p freenet` clean
- `cargo clippy --locked -- -D warnings` clean (matches CI)
- `cargo test -p freenet --lib`:
  - `node::network_status::tests::test_op_stats_recording` ok
  - `operations::subscribe::op_ctx_task::tests::deliver_outcome_records_subscribe_op_result` ok
  - `operations::update::op_ctx_task::tests::deliver_outcome_records_update_op_result` ok

## Follow-up (not in scope)

The legacy state-machine path in `report_result` still counts
sub-operation SUBSCRIBE/GET and renewals. With the migrations
nearly complete those code paths are mostly dead, but the long-term
plan from `#4009`'s "Follow-up" section is to compute the dashboard
op_stats from event-log aggregates rather than maintain mirror
counters. That refactor is the proper fix for the broader bug class
and is left for a separate PR.

GET and PUT task-per-tx drivers also undercount the
`Exhausted` / `InfrastructureError` failure cases: the success
recording is in the Done arm only, and `deliver_outcome` for those
ops doesn't record failures. This is a related but smaller bug; not
covered here to keep scope tight.

## Fixes

Closes #4010

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -821,13 +821,10 @@ mod tests {
         record_op_result(OpType::Get, true);
         record_op_result(OpType::Get, false);
         record_op_result(OpType::Put, true);
-        // SUBSCRIBE / UPDATE counters were silently stuck at zero before
-        // issue #4010 (the task-per-tx drivers stopped invoking the
-        // recording hook after the #1454 migrations). Cover both op
-        // types and both outcomes here so any regression in the
-        // counter wiring fails this test alongside the source-grep
-        // pin tests in `subscribe/op_ctx_task.rs` and
-        // `update/op_ctx_task.rs`.
+        // Issue #4010: SUBSCRIBE / UPDATE counters were stuck at zero
+        // because the task-per-tx drivers stopped recording outcomes.
+        // Pin both op types and both outcomes so the wiring regression
+        // is caught here in addition to the source-grep pin tests.
         record_op_result(OpType::Subscribe, true);
         record_op_result(OpType::Subscribe, false);
         record_op_result(OpType::Update, true);
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -280,10 +280,24 @@ pub fn record_peer_disconnected(addr: SocketAddr) {
 /// - `operations/update/op_ctx_task.rs::deliver_outcome` (client-
 ///   initiated UPDATE; covers all `DriverOutcome` variants).
 ///
-/// Internal-only operations (subscription renewals, sub-operation
-/// SUBSCRIBE/GET spawned by PUT/GET, executor auto-subscribe) are
-/// intentionally NOT recorded: they would inflate the user-facing
-/// counter with background traffic.
+/// Internal-only operations are intentionally NOT recorded; counting
+/// them would inflate the user-facing counter with background traffic
+/// (renewals fire every 2 minutes per active subscription, sub-op
+/// SUBSCRIBE fires once per PUT/GET completion). Specifically:
+///
+/// - `operations/subscribe/op_ctx_task.rs::run_renewal_subscribe`
+///   (subscription renewal driver, called from
+///   `ring::connection_maintenance`).
+/// - `operations/subscribe/op_ctx_task.rs::run_executor_subscribe`
+///   (executor auto-subscribe, called from
+///   `contract::executor::runtime`).
+/// - Sub-operation SUBSCRIBE spawned by PUT/GET via
+///   `Transaction::new_child_of` (gated in
+///   `subscribe::op_ctx_task::deliver_outcome` by
+///   `!client_tx.is_sub_operation()`).
+/// - Sub-operation GET spawned by subscribe / executor
+///   (`get::op_ctx_task::start_sub_op_get` returns through a oneshot
+///   and never publishes via `result_router_tx`).
 ///
 /// Audit: `grep -rn "record_op_result" crates/core/src/operations/`
 /// must show coverage for every op type that has a task-per-tx driver.
@@ -823,8 +837,11 @@ mod tests {
         record_op_result(OpType::Put, true);
         // Issue #4010: SUBSCRIBE / UPDATE counters were stuck at zero
         // because the task-per-tx drivers stopped recording outcomes.
-        // Pin both op types and both outcomes so the wiring regression
-        // is caught here in addition to the source-grep pin tests.
+        // Cover both op types and both outcomes (success + failure) so
+        // the per-op tabulation is pinned end to end. The two
+        // `OpType::Update, true` calls are intentional, not a copy
+        // paste: we want the success counter to assert `2`, paired
+        // with one failure for `1`.
         record_op_result(OpType::Subscribe, true);
         record_op_result(OpType::Subscribe, false);
         record_op_result(OpType::Update, true);
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -262,7 +262,31 @@ pub fn record_peer_disconnected(addr: SocketAddr) {
     }
 }
 
-/// Record an operation result.
+/// Record an operation result for the dashboard "Operations" panel.
+///
+/// # Required call sites
+///
+/// This is a manually-mirrored counter: every code path that delivers a
+/// terminal client-visible op result MUST call this exactly once with
+/// the matching outcome. Forgetting a call site silently rots the
+/// counter (issue #4009 / #4010 prior incidents). Current writers:
+///
+/// - `node.rs::report_result` for legacy state-machine ops (PUT/GET/
+///   UPDATE/SUBSCRIBE that still flow through `OpManager.ops.*`).
+/// - `operations/get/op_ctx_task.rs` `Done` arm (client-initiated GET).
+/// - `operations/put/op_ctx_task.rs` `Done` arm (client-initiated PUT).
+/// - `operations/subscribe/op_ctx_task.rs::deliver_outcome` (client-
+///   initiated SUBSCRIBE; covers all `DriverOutcome` variants).
+/// - `operations/update/op_ctx_task.rs::deliver_outcome` (client-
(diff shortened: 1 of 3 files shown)


sanity:
## Review-feedback follow-up (commit 2ea32d5c)

### Codex caught a real bug — fixed
`run_client_subscribe` is also the entry point for sub-op SUBSCRIBE spawned by PUT/GET (`maybe_subscribe_child`, `start_subscription_request`), so the original unconditional `record_op_result` in `deliver_outcome` would have inflated the user-facing dashboard counter on every PUT/GET completion. Gated the call on `!client_tx.is_sub_operation()` in both SUBSCRIBE and UPDATE drivers (UPDATE has no sub-op entry today; the gate is defensive). Pin tests now assert the literal `!client_tx.is_sub_operation()` substring so a flipped guard fails CI.

### Other review points addressed

- **Behavioral test for success classification** (testing reviewer, code-first reviewer): extracted `classify_{subscribe,update}_outcome_for_op_stats` pure helpers and added unit tests covering every `DriverOutcome` variant. Pins the success/failure mapping (incl. SUBSCRIBE's `SkipAlreadyDelivered = success` semantic) against accidental flips that the source-grep cannot detect.
- **Source-grep brittleness** (all four reviewers): replaced the simplifier's `split_once` on literal section-comment delimiters with brace-walking helpers, so renaming the `// ── Relay ... driver` markers no longer breaks the pin tests.
- **Negative-path coverage for renewal/executor subscribe** (skeptical reviewer, code-first reviewer): added `renewal_and_executor_subscribe_do_not_record_op_stats` source-grep pin asserting both functions never call `deliver_outcome` or `record_op_result`. Guards against re-routing background traffic into the user-facing counter.
- **Doc completeness** (code-first reviewer): expanded `record_op_result` rustdoc to enumerate the negative cases by name (renewal, executor auto-subscribe, sub-op SUBSCRIBE, sub-op GET) so a future migration's required reading is unambiguous.
- **Test typo callout** (skeptical, code-first reviewers): the duplicated `record_op_result(OpType::Update, true)` lines in `test_op_stats_recording` are intentional (asserts `(2, 1)`); added a comment so it doesn't read as a copy paste slip.
- **`bug-prevention-patterns.md` entry** (big-picture reviewer): added an entry under "Manually-mirrored telemetry counters / dashboard providers" linking #4009 and #4010, with audit greps and the gate-on-`!is_sub_operation` pattern. (That file is in the local rules dir, not in the freenet-core repo, so it's not part of this PR diff.)

### Not addressed (deliberate, with reason)

- **Renewal subscribes uncounted vs. recorded as separate counter** (skeptical reviewer): out of scope for this PR. The PR's goal is to stop the silent zero on the user-facing counter; a `subscribes_renewed` field is a dashboard-design question that warrants its own discussion. Today's behavior (only count user-initiated subscribes) matches what the dashboard label "Operations" implies, and the doc + tests pin that semantic.
- **Lock contention via atomic counters** (skeptical reviewer): acceptable today (50-client cap documented elsewhere), and `record_op_result` never awaits under the write lock so no deadlock surface. Worth revisiting if the cap rises.
- **Double-counting `debug_assert`** (skeptical reviewer): provably impossible today via the bypass at `handle_pure_network_message_v1`; adding a runtime assert would only fire on a future refactor that re-pushes a client-initiated SubscribeOp into `OpManager.ops.*`, which itself would warrant broader review.
- **CI structural check that audit grep finds hits per task-per-tx driver file** (testing reviewer): the per-driver source-grep pin tests already serve as that check at the unit-test level. A workspace-wide CI lint would be a separate, larger change.
- **GET/PUT failure-arm undercounting follow-up** (PR description): noted in the PR body as out of scope.

### Verification

- `cargo fmt -p freenet` clean
- `cargo clippy --locked -- -D warnings` clean (matches CI invocation)
- `cargo test -p freenet --lib` (6 SUBSCRIBE/UPDATE/op_stats tests):
  - `node::network_status::tests::test_op_stats_recording` ok
  - `operations::subscribe::op_ctx_task::tests::deliver_outcome_records_subscribe_op_result` ok
  - `operations::subscribe::op_ctx_task::tests::renewal_and_executor_subscribe_do_not_record_op_stats` ok
  - `operations::subscribe::op_ctx_task::tests::classify_subscribe_outcome_covers_all_variants` ok
  - `operations::update::op_ctx_task::tests::deliver_outcome_records_update_op_result` ok
  - `operations::update::op_ctx_task::tests::classify_update_outcome_covers_all_variants` ok

[AI-assisted - Claude]

PR #4013, opened 2026-05-03
fix(release-script): handle existing OPEN PR and use structured CI status

sanity:
## Problem

Two bugs in `scripts/release.sh` surfaced during the 0.2.52 release attempt and broke the resumable-release contract.

### A. Existing OPEN PR aborts the script

When `create_release_pr` detects an existing OPEN release PR it correctly marks `PR_CREATED` and prints \"Monitoring existing PR instead of creating a new one...\" — but then **falls through** into the create-branch / push / \`gh pr create\` block. \`gh pr create\` fails because a PR already exists for that head, the \`set -euo pipefail\` trap fires, and the script exits 1 every time you re-run to resume a release with an already-open PR.

### B. False-positive CI failure detection

The wait loop greps \`gh pr checks\` plain-text output for \`fail|error\`. That string-matching is too coarse — words like \"failure\" appear in transient strings, and the substring \`fail\` matches any check name or URL fragment containing it. The loop intermittently aborts with \"CI checks failed!\" while everything is actually green.

Both bugs hit on the in-flight 0.2.52 release: (B) on the first invocation, (A) on every resume attempt after that.

## Solution

**(A)** Introduce a `skip_pr_creation` flag set in the OPEN-PR branch. The create-branch / push / `gh pr create` block now runs only when that flag is false. The auto-merge enable + wait-loop runs in both paths so existing OPEN PRs are correctly monitored. Also surface the previously-silent `gh pr merge --auto` failure with `|| true` so the wait loop still runs even if auto-merge can't be queued (the queued-merge / mergify path still works).

**(B)** Use `gh pr view --json statusCheckRollup` for structured data. A check counts as failed only when its `conclusion` (or `state` for non-Actions checks like CLA bots) is exactly one of `FAILURE`, `TIMED_OUT`, `CANCELLED`, or `ACTION_REQUIRED`. Same approach for the pending/passed counters in the wait message.

## Testing

- `bash -n scripts/release.sh` clean.
- Verified the structured rollup logic against PR #4012 (12 checks, all pass): 0 failed, 0 pending, 12 passed — matches reality. The old grep would have flagged any check name or URL containing the substring `fail` as a failure.
- Trace-through of the resume path with an existing OPEN PR: \`skip_pr_creation=true\` skips the create block, \`pr_number\` is non-empty (set from the existing-PR lookup), \`gh pr merge --auto\` is idempotent, control falls through to the wait loop. ✅

[AI-assisted - Claude]


sanity:
**test-exempt justification**: this is a release-tooling fix in a bash script. The repo has no shell-test infrastructure — adding bats/shellcheck for a single fix would be disproportionate. The fixes were validated empirically: (a) the structured rollup logic was tested live against PR #4012 and produced correct counts (12 checks: 0 failed, 0 pending, 12 passed) where the old grep had been false-positiving; (b) the resume-with-OPEN-PR path was traced through statically — `skip_pr_creation` flag bypasses the create block, the auto-merge enable runs idempotently, control falls through to the wait loop with the existing `pr_number`. Both fixes will be re-exercised on the next release; if either regresses the symptom is immediate (script aborts).

[AI-assisted - Claude]

PR #4023, opened 2026-05-04
fix(server): push permission prompts to all tabs via SSE

sanity:
## Problem

Permission prompts (delegate `RequestUserInput`) only reliably appear on whichever Freenet tab happens to be foregrounded when the prompt lands. The shell-page overlay polls `/permission/pending` every 3s and skips the poll entirely on hidden tabs that have no existing overlay cards:

```javascript
if (typeof document !== 'undefined' &&
    document.visibilityState === 'hidden' &&
    Object.keys(overlayCards).length === 0) {
  return;     // skip polling
}
```

This was added to save bandwidth on lots of background tabs, but it means a user who clicks a delegate-triggered button in tab A but focuses tab B before the gateway creates the prompt will only see the prompt on B. The `visibilitychange` listener does re-trigger when A is refocused, but the user has to notice and switch back — in practice the originating tab silently misses prompts.

Reproduced on technic with two Playwright tabs (vault + harvest):

| Step | Vault | Harvest |
|---|---|---|
| Both tabs visible, vault foreground | 0 cards | 0 cards |
| Vault → hidden | 0 cards | 0 cards |
| Click Test Prompt on vault | server has prompt, but… | server has prompt |
| 4 s later (vault still hidden) | **0 cards** | 1 card |
| Switch vault back to visible | 1 card | 1 card |

## Solution

Replace polling with Server-Sent Events. The gateway pushes `prompt_added` the moment `DashboardPrompter::prompt` fires the broadcast, and `prompt_removed` when the respond handler or the timeout cleanup retires the prompt. Every open Freenet tab subscribes to `/permission/events` and dismisses cards as soon as the event arrives, regardless of visibility.

**Subscribe-then-snapshot** avoids the race where a prompt added between page load and SSE handshake would be invisible: the handler subscribes to the broadcast before iterating the pending DashMap, so any prompt added during the window arrives via the broadcast (possibly duplicated with a snapshot entry; the renderer keys on nonce so duplicates are idempotent).

`/permission/pending` is retained as the bootstrap-on-connect / on-resync / no-EventSource-fallback endpoint. The visibility-skip loop is gone — SSE pushes regardless of visibility, so the bandwidth-saving optimisation is no longer needed (the per-tab cost is one TCP connection vs. ~14 polls/min).

Connection cap (64) and broadcast capacity (128) bound resource usage. On lag (rare — capacity 128 with at most 32 concurrent prompts and two events per lifecycle), the SSE handler emits a `resync` event so the client reconciles from `/permission/pending`.

## Testing

- **`test_prompt_lifecycle_emits_added_and_removed`** and **`test_prompt_timeout_emits_removed`** (`user_input.rs`): assert the broadcast fires in both lifecycle paths (HTTP-handler-triggered and prompter-timeout-triggered).
- **`test_sse_emits_added_when_prompt_inserted`** and **`test_sse_emits_removed_when_prompt_responded`** (`permission_prompts.rs`): end-to-end SSE delivery, filtered by per-test nonce so the shared global broadcast doesn't cross-contaminate parallel tests.
- **`test_sse_replays_existing_pending_on_subscribe`**: bootstrap replay contract — a fresh subscriber sees `prompt_added` for already-pending entries.
- **`test_sse_untrusted_origin_returns_closed_stream`**: origin gating parity with `/permission/pending`.
- **Updated `shell_page_permission_overlay_present_and_safe`** to pin the new wire contract — asserts `visibilityState` no longer appears in the overlay path. This is the CI gap closure: a future polling-restoration would fail the test rather than silently re-shipping the bug.

**Why didn't CI catch this originally?** The existing overlay test only verified that the polling loop existed; it did not verify cross-tab delivery semantics or the visibility-skip behaviour. The new `!visibilityState` assertion locks in the new contract.

## Fixes

Found while investigating an unrelated UX in the ghostkey vault; reproduced and root-caused on technic via Playwright before this PR.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7038,6 +7038,7 @@ dependencies = [
  "futures-core",
  "pin-project-lite",
  "tokio",
+ "tokio-util",
 ]
 
 [[package]]
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -104,7 +104,7 @@ opentelemetry_sdk = { workspace = true, optional = true }
 # internal deps
 freenet-stdlib = { features = ["net"], workspace = true }
 console-subscriber = { workspace = true, optional = true }
-tokio-stream = { workspace = true }
+tokio-stream = { workspace = true, features = ["sync"] }
 freenet-test-network = { workspace = true, optional = true }
 turmoil = { workspace = true }  # Deterministic simulation scheduling (always enabled for SimNetwork)
 lru = { workspace = true }
--- a/crates/core/src/contract/user_input.rs
+++ b/crates/core/src/contract/user_input.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 
 use dashmap::DashMap;
 use freenet_stdlib::prelude::{ClientResponse, UserInputRequest};
-use tokio::sync::oneshot;
+use tokio::sync::{broadcast, oneshot};
 
 /// Timeout for user input prompts. After this duration, the request is auto-denied.
 pub(crate) const USER_INPUT_TIMEOUT: Duration = Duration::from_secs(60);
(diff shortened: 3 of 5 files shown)


sanity:
## Review pass 2: addressed

Pushed 4b98358 with fixes for the parallel review findings. Summary:

**Fixed:**
- **CRITICAL (skeptical):** Cross-origin DoS via missing Origin — added combined `Origin` + `Sec-Fetch-Site` check, rejected before slot allocation.
- **P1 (Codex):** EventSource error handler now starts a 3s polling fallback and stops it on `open`. Resync handler no longer hides-all-then-fetches; uses the diff-aware reconcile path.
- **HIGH (code-first):** Nonce sanitised at `emit_prompt_event` boundary.
- **MEDIUM (skeptical/code-first):** `SSE_CONNECTIONS` race — `compare_exchange_weak` loop, AcqRel ordering.
- **MEDIUM (testing):** No test for `Lagged`/resync — added `test_sse_emits_resync_on_lag`.
- **MEDIUM (testing/skeptical):** No test for cap rejection + drop release — added `test_sse_connection_cap_and_release_on_drop` with a serialising mutex so concurrent tests can't perturb the counter.
- **MEDIUM (testing):** No test for `permission_respond` firing `Removed` — added `test_respond_handler_emits_prompt_removed`.
- **HIGH (testing):** Bootstrap-replay test didn't pin ordering — added `test_sse_bootstrap_replay_arrives_before_live`.
- **MEDIUM (multiple):** 50ms sleeps replaced with `wait_subscribers_then_send` that polls `receiver_count()`.
- **LOW (skeptical):** Single source for `MAX_IDENTITY_HASH_CHARS` / `OVERLAY_KEY_CHARS_MAX`.
- **LOW (big-picture):** Em-dashes removed from new comments. PR title renamed `feat:` → `fix:` since this is a UX regression fix.

**Considered but not changed (with reasoning):**
- **Code-first MEDIUM "Added-after-Removed reordering":** the scenario described requires a prompt to be in the DashMap at iter time AND removed between subscribe and iter, which is contradictory (subscribe is before iter, so anything removed before iter wouldn't be in the snapshot). Walked through every timeline ordering and the chain `iter(initial).chain(live)` produces a correct sequence in each one. The new ordering test pins this.
- **Skeptical MEDIUM "lag-resync flicker":** addressed by removing the unconditional hide-all on resync; reconcile path's diff handles this.
- **Testing LOW (char-cap test for `cap_identity_chars`):** pre-existing helper, not in scope of this PR.

[AI-assisted - Claude]

PR #4024, opened 2026-05-04
fix(transport): meter recv-side wire bytes post-authentication + LRU evict per-peer table

### LINKED ISSUE #3999: Move recv-side per-peer metering past authentication + add LRU eviction
## Background

PR #3996 moved dashboard wire-byte accounting from stream-completion to the UDP socket layer so the local-peer dashboard reflects all packet traffic (keep-alives, ACKs, small contract ops), not just stream transfers. That fix exposed a pre-existing weakness in the per-peer DashMap that this PR amplifies and a follow-up should close.

## Problem

`TransportMetrics::record_packet_received` is now called from `Socket::recv_from` on **every** received UDP packet, before any decryption or peer-validation. Combined with `record_per_peer`'s no-eviction `MAX_TRACKED_PEERS = 256` policy, an attacker that fabricates UDP packets from many spoofed source IPs can:

1. Fill the per-peer DashMap to capacity. After the 256th distinct source IP, legitimate new peers' bytes are silently dropped from per-peer accounting forever (no LRU/TTL eviction in `metrics.rs::record_per_peer`).
2. Inflate `cumulative_bytes_received` arbitrarily. Memory damage is bounded (u64 wrap is impractical; ~16 bytes per entry × 256 entries) but the dashboard's "Total Data Received" gauge becomes attacker-controlled.

The 256-cap was originally fine because it was only reached by peers that completed an authenticated stream transfer. The new socket-layer metering means any source IP that sends a single byte gets a slot.

## Proposed fix

Two changes, ideally in the same PR:

1. **Move the per-peer increment past authentication.** Keep `record_packet_sent` at the socket layer (no spoof concern — outbound targets are always known). For inbound, hook the per-peer accounting in `peer_connection.rs` immediately after `try_decrypt_sym(...)` succeeds at line ~653. The `cumulative_bytes_received` counter can either follow (post-authentication only) or stay at the socket layer with a documented cap on inflation.
2. **Add LRU/TTL eviction to `record_per_peer`.** Even with authentication, gateways accumulate stale entries from torn-down connections. An LRU eviction policy (or a periodic sweep based on a `last_seen_at` field) keeps the dashboard's per-peer view current without unbounded growth.

## Scope

This is too big for the original dashboard fix PR. Tracking separately so it doesn't block the symptom fix.

## References

- Skeptical reviewer flagged this as HIGH severity in PR #3996 review (item #1).
- Code-first reviewer concurred (item #6).
- See comment in `crates/core/src/transport/metrics.rs::record_per_peer` for the linked rationale.

[AI-assisted - Claude]

sanity:
## Problem

PR #3996 moved dashboard wire-byte accounting from stream-completion to the UDP socket layer so the local-peer dashboard reflects keep-alives, ACKs, and small contract ops — not just stream transfers. That fix exposed a pre-existing weakness in the per-peer DashMap:

- `TransportMetrics::record_packet_received` ran from `Socket::recv_from` on every received UDP packet, **before** any decryption or peer-validation.
- `record_per_peer` had no eviction policy and a 256-entry cap.

Combined, an attacker that fabricates UDP packets from many spoofed source IPs could:

1. Fill the per-peer DashMap to capacity. After 256 distinct source IPs, legitimate new peers' bytes were silently dropped from per-peer accounting forever.
2. Inflate `cumulative_bytes_received` arbitrarily — making the dashboard's "Total Data Received" gauge attacker-controlled.

The same no-eviction policy also caused stale entries to accumulate on long-running gateways as peer connections churned, even without a spoof attack — so a gateway with high peer turnover would eventually display "—" for SENT/RECV on legitimate new peers.

## Approach

Two coordinated changes that follow the fix proposed in #3999, plus an eager-cleanup complement:

**1. Move recv-side metering past authentication.** `record_packet_received` is now called from `PeerConnection::recv` immediately after a packet passes `try_decrypt_sym` against the established symmetric session key — i.e. only for packets we know came from a peer holding the shared secret. The call was removed from `Socket::recv_from`. Outbound metering stays at the socket layer because send targets are controlled by us; there is no spoof vector on send.

**2. LRU-evict the per-peer table when full.** `PeerTransferStats` gained a `last_seen_tick` atomic stamped from a global monotonic counter on every update. When the table is full and a new peer arrives, `record_per_peer` evicts the entry with the smallest tick before inserting. Pure tick-based ordering avoids a wall-clock dependency and keeps simulation determinism intact (the rules in `.claude/rules/code-style.md` discourage `Instant::now()` in `crates/core/`).

**3. Eager cleanup on disconnect.** `TRANSPORT_METRICS.remove_peer` is called from `record_peer_disconnected`, so authenticated peer teardowns free their slot immediately rather than waiting for LRU eviction.

## Testing

- `udp_socket_recv_from_does_not_record_pre_authentication` (new) — pins the spoof regression: a forged UDP packet at the kernel handoff point creates no per-peer entry and contributes nothing to the cumulative counter on its own.
- `test_per_peer_lru_evicts_oldest` (new) — pins the LRU ordering: refreshing the second-oldest entry's tick changes which entry is evicted on the next insert into a full table.
- `test_per_peer_capacity_bound` (updated) — reflects the new LRU semantics. Pre-fix the over-capacity peer was silently dropped; post-fix it is tracked at the cost of evicting the oldest entry.
- `test_remove_peer_frees_slot` and `test_record_per_peer_is_idempotent_under_remove` (new) — pin the disconnect-cleanup path and the no-op behavior of removing a non-existent / already-evicted peer.
- `udp_socket_records_packet_metrics` (updated) — only the send side now asserts cumulative + per-peer movement; recv-side accounting moved to the post-auth path.

All 12 metrics unit tests + 13 transport dual-stack tests + full freenet lib suite (2538 tests) pass locally.

## Fixes

Closes #3999

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -260,6 +260,11 @@ pub fn record_peer_disconnected(addr: SocketAddr) {
             s.connected_peers.retain(|p| p.address != addr);
         }
     }
+    // Free the per-peer metrics slot so the bounded table doesn't accumulate
+    // stale entries across the lifetime of long-running gateways. LRU
+    // eviction in `record_per_peer` is the safety net; this is the
+    // best-effort eager cleanup.
+    TRANSPORT_METRICS.remove_peer(addr);
 }
 
 /// Record an operation result for the dashboard "Operations" panel.
--- a/crates/core/src/transport.rs
+++ b/crates/core/src/transport.rs
@@ -439,7 +439,11 @@ impl Socket for UdpSocket {
         // Normalize ::ffff:x.x.x.x → plain IPv4 so the rest of the system
         // sees a consistent address regardless of socket type.
         let addr = normalize_mapped_addr(addr);
-        metrics::TRANSPORT_METRICS.record_packet_received(addr, len as u64);
+        // Receive-side dashboard metering happens post-authentication — see
+        // `peer_connection::PeerConnection::recv` after `try_decrypt_sym`
+        // succeeds. Doing it here would let an attacker spoofing UDP source
+        // addresses inflate the dashboard counters and crowd legitimate
+        // peers out of the bounded per-peer table (#3999).
         Ok((len, addr))
     }
 
@@ -807,12 +811,15 @@ mod dual_stack_tests {
         assert_eq!(&buf[..len], b"pong");
     }
 
-    /// Regression: every successful send_to / recv_from on the production
-    /// `UdpSocket` must update both the cumulative wire-byte counters and the
-    /// per-peer counters used by the local dashboard. Before this fix, those
-    /// counters only moved on stream-transfer completion, so a node connected
-    /// for hours could legitimately show "—" for SENT/RECV despite real
-    /// keep-alive traffic flowing.
+    /// Regression: every successful send_to on the production `UdpSocket`
+    /// must update the cumulative + per-peer counters used by the local
+    /// dashboard. Before #3996, those counters only moved on stream-transfer
+    /// completion, so a node connected for hours could legitimately show "—"
+    /// for SENT despite real keep-alive traffic flowing.
+    ///
+    /// Receive-side accounting is owned by the post-authentication hook in
+    /// `peer_connection.rs` (see #3999), so this test only asserts send
+    /// behavior — `recv_from` deliberately does NOT bump any counter.
     #[tokio::test]
     async fn udp_socket_records_packet_metrics() {
         use crate::transport::metrics::TRANSPORT_METRICS;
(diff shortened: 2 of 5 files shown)


sanity:
## Review responses

Five reviews ran in parallel (code-first, testing, skeptical, big-picture, Codex). Summary of how each finding was addressed in 83ae73b7:

### Addressed in code

**Code-first MEDIUM — handshake-completion bytes silently dropped from dashboard.** Added `record_packet_received` at the three authenticated-decrypt-success sites identified by the reviewer plus the skeptical reviewer:
- `connection_handler.rs::gateway_connection` (gateway-side symmetric ACK)
- `connection_handler.rs::traverse_nat` (client-side symmetric ACK)
- `peer_connection.rs::recv` intro-packet asymmetric-decrypt success path

**Big-picture MEDIUM — stale field comment.** The struct comment on `cumulative_bytes_sent`/`cumulative_bytes_received` still claimed both update at the socket layer "for every packet"; rewrote to describe the post-auth asymmetry and cite #3999.

**Testing CRITICAL #1 + #2 — post-auth recv hook untested; encrypted-byte semantics unpinned.** Added `recv_records_packet_metrics_post_authentication` that builds an encrypted ShortMessage, feeds it via the inbound channel, drives one iteration of `PeerConnection::recv`, and asserts both that the per-peer entry advanced by the **wire (encrypted) size** (not the decrypted size) and that the cumulative counter moved.

### Considered but not changed

**Codex P2 — backpressure-dropped packets at `try_send_established` no longer counted.** Real but acceptable tradeoff. `INBOUND_CHANNEL_CAPACITY = 2048` so this only fires under sustained heavy load, and the alternative (counting at the listener pre-decrypt) reintroduces the spoof inflation attack the issue is closing. Under-reporting under load is "users see truth modulo backpressure drops"; spoof-inflated counters are "users see fiction".

**Skeptical H1 — replay/retransmit double-counts.** Both pre-fix and post-fix count retransmits N times because they ARE on the wire N times — that's correct semantics for a wire-byte counter. The post-fix change does not introduce a new threat: an authenticated peer who replays packets to inflate their own dashboard entry can do the same by sending fresh valid traffic. Anonymous spoof inflation (the actual issue) is closed.

**Skeptical H3 — concurrent-burst over-eviction race.** Real but bounded. With N concurrent inserts at capacity, up to N evictions can fire instead of 1. In practice rare and self-healing; for a dashboard counter the imprecision is acceptable. Inline comment in `record_per_peer` already acknowledges the transient over-capacity. Adding a Mutex would serialize the hot path for an aesthetic improvement that doesn't change correctness.

**Skeptical H4 — `remove_peer` racing with `record_per_peer` orphans the increment.** DashMap 6.1's `Ref<K, V>` holds a per-shard `RwLockReadGuard` for its lifetime, so a concurrent `remove` (which needs the write lock) cannot interleave with our `fetch_add` on the held guard. The race the reviewer described is specific to `Arc<RwLock<HashMap>>` patterns, not DashMap. Verified against dashmap 6.1.0.

**Big-picture LOW (other suggestions).** Documentation polish, follow-up scope. Filing or addressing these would belong in a follow-up PR rather than block #3999.

### Tests
- `cargo test -p freenet --lib transport::` — 565 passed
- New test `recv_records_packet_metrics_post_authentication` — passes
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt` — clean

[AI-assisted - Claude]

Issue #17, opened 2021-11-14
Switch to RocksDB store back end as default storage


iduartgomez:
In the future build with both Sqlite and RocksDB as possible backends (via feature) in order to compare behavior and performance characteristics. Probably we will want to stick with one, eventually, but otherwise, RocksDB can be used as a building block  KV for more complex databases (is already used as such for distributed databases ala TiKV, for example), but it does not have SQL semantics on top of it, unlike Sqlite, which could prove useful.

In any case demonstrating we can quickly swap between store backends is benefitial so we are not stuck and coupled with one (and since right now we only need to maintain simple key-value semantics is rather easy to keep the backend implementation decoupled from the rest of the application, something we would like to continue in the future).

kakoc:
I'll tackle on that

PR #4040, opened 2026-05-06
feat: mirror freenet-core HEAD to Freenet-hosted demo repo

sanity:
## Summary

- Adds `.github/workflows/mirror-to-freenet.yml` which calls the reusable workflow at `freenet/freenet-git/.github/workflows/mirror-repo.yml` (merged in [freenet-git#17](https://github.com/freenet/freenet-git/pull/17)).
- Triggers: push to main, workflow_dispatch, daily cron (11:42 UTC) as a safety net for missed pushes.
- Mode is `snapshot` because freenet-core's git history (~176 MiB) is larger than what the repo contract can hold. Each run rebuilds the demo as a single orphan commit of HEAD's tracked tree -- consumers get current source, no history.

## Demo URL

`freenet:3GEERif5ihbf/freenet-core` (replaces the old `AaRxPZVdWrPh` demo, which was created from a passphrase-encrypted bundle that doesn't fit a CI workflow).

## Required secrets (already configured on this repo)

- `FREENET_GIT_IDENTITY_BUNDLE_BASE64` — base64 of the no-passphrase bundle that owns the new prefix.
- `FREENET_GIT_WS_URL` — same value as `RIVER_GATEWAY_URL`.

## Test plan

- [x] Reusable workflow merged on the freenet-git side ([#17](https://github.com/freenet/freenet-git/pull/17)).
- [x] Initial state for `3GEERif5ihbf/freenet-core` already published manually (clonable today via `git clone freenet::3GEERif5ihbf/freenet-core`).
- [x] Rescue workflow on the freenet-git side smoke-tested — green in 3m45s against the live demo URLs.
- [ ] After merge: confirm a push to main triggers the workflow and produces a fresh snapshot. The first run will overwrite the manually-published initial state.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/mirror-to-freenet.yml
+++ b/.github/workflows/mirror-to-freenet.yml
@@ -20,8 +20,16 @@ on:
 
 jobs:
   mirror:
-    uses: freenet/freenet-git/.github/workflows/mirror-repo.yml@main
+    # Pinned to the merge SHA of freenet-git#17 rather than `@main`.
+    # `secrets: inherit` would pass every secret in this repo (Apple
+    # notarization keys, Claude OAuth, OpenAI key, Matrix tokens,
+    # River signing key) into whatever happens to be on freenet-git's
+    # main, so an unpinned ref + inherit is a textbook supply-chain
+    # pivot. Bump the SHA when the reusable workflow is rev'd.
+    uses: freenet/freenet-git/.github/workflows/mirror-repo.yml@362e547c7035d28968651d2f4106fbfa891c31e2
     with:
       freenet_repo: "3GEERif5ihbf/freenet-core"
       mode: snapshot
-    secrets: inherit
+    secrets:
+      FREENET_GIT_IDENTITY_BUNDLE_BASE64: ${{ secrets.FREENET_GIT_IDENTITY_BUNDLE_BASE64 }}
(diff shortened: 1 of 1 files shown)


sanity:
Addressed Codex P1 + Skeptical H1/H2: pinned `@main` to the merge SHA of freenet-git#17 (`362e547c7035d28968651d2f4106fbfa891c31e2`) and replaced `secrets: inherit` with explicit pass-through of just the two secrets the reusable workflow declares as required.

**Pushed back / deferring:**
- **Skeptical H3** (snapshot orphan commit picks up wall-clock time → daily cron creates a new commit hash even when the source is unchanged). Real issue, but the fix belongs in the reusable workflow in freenet-git, not these caller PRs. Will follow up with a separate PR there to set `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE` from the source commit's `%cI` so the orphan commit is deterministic.
- **Skeptical M2** (no `paths:` filter — every commit to main mirrors). Reasonable optimization but not blocking; would tune `paths-ignore: ['*.md', 'docs/**', '.github/**']` after observing real workflow usage.
- **Skeptical M1, L1, L3**: comment polish / minor; not blocking. Caller `workflow_dispatch:` is a usable manual recovery button per the verified-safe analysis.

[AI-assisted - Claude]

PR #4049, opened 2026-05-07
fix(runtime): persist delegate ctx.write across process() calls

sanity:
## Problem

A delegate that writes pending state via `ctx.write(...)` during a `RequestUserInput` round-trip expects to read it back via `ctx.read()` when the executor re-enters with the matching `UserResponse`. The runtime initialised the per-call context `Vec` to empty on every `inbound_app_message` call, so the bytes a delegate stored disappeared between the two calls and the second call hit the delegate's own `received UserResponse with no pending context` guard.

User-visible effect: the ghostkey delegate's `RequestAnyAccess` flow — used by Harvest's "Connect a ghostkey" cross-app share — was completely dead. Reproduction on technic: vault renders the prompt overlay correctly, the user clicks Share, then harvest stays in "Waiting for vault…" forever while the delegate aborts. The same shape would break any prompt-driven delegate that relies on the documented `ctx.write` / `ctx.read` ABI across a prompt round-trip, including the older Allow/Deny single-fingerprint flow.

## Approach

Thread the context `Vec` through a shared `DelegateContextCache` (`Arc<DashMap<DelegateKey, Vec<u8>>>`) seeded into each `Runtime` in the executor pool. `inbound_app_message` loads the persisted bytes on entry and stores the (possibly mutated) `Vec` on exit. `unregister_delegate` clears the entry so a re-registered delegate starts fresh.

The cache is shared across pool executors because a delegate's first call may run on executor A and its `UserResponse` follow-up on executor B; per-`Runtime` storage would have the same locality bug as the per-call `Vec`.

Persisted bytes that exceed `DelegateContext::MAX_SIZE` are dropped with a warning rather than crashing the next call (where the wire format would assert on size). The host `ctx.write` host function does not enforce a size cap of its own.

## Testing

- New `test_context_persists_between_calls` exercises the same shape as the prompt round-trip: writes context on the first call, reads it on the second. Pre-fix returns `CounterValue(1)` twice; with the fix returns `CounterValue(1)` then `CounterValue(2)`. Also asserts that `unregister_delegate` clears the cache.
- `test_large_context_data` previously wrote 1 MiB (well past `DelegateContext::MAX_SIZE`) and asserted the second call read back an empty Vec — only "passing" because context didn't persist. Updated to write a within-bounds size and assert the bytes round-trip.
- All 2553 lib tests pass.
- `cargo clippy --locked -- -D warnings` (the CI invocation) is clean.

## Validation plan

After merge, deploy to technic and re-run the Playwright cross-app validation script (`harvest-cross-app-validate.mjs`) — it should now show "Share ciQaxxSwKF8" → harvest's `My Store` populates with the shared ghostkey within ~5s instead of timing out at "Waiting for vault…".

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -163,6 +163,8 @@ pub struct RuntimePool {
     shared_contract_modules: SharedModuleCache<ContractKey>,
     /// Shared compiled delegate module cache.
     shared_delegate_modules: SharedModuleCache<DelegateKey>,
+    /// Shared per-delegate `ctx.write()` cache (see `DelegateContextCache`).
+    shared_delegate_contexts: crate::wasm_runtime::DelegateContextCache,
     /// Shared backend engine used by all executors.
     ///
     /// All executors MUST share the same backend engine because compiled modules
@@ -226,6 +228,9 @@ impl RuntimePool {
             Arc::new(Mutex::new(LruCache::new(cache_capacity)));
         let shared_delegate_modules: SharedModuleCache<DelegateKey> =
             Arc::new(Mutex::new(LruCache::new(cache_capacity)));
+        // Shared delegate-context cache so a prompt round-trip routed to a
+        // different pool executor still finds its `ctx.write()` blob.
+        let shared_delegate_contexts = crate::wasm_runtime::new_delegate_context_cache();
 
         // Create shared recovery guard for corrupted-state self-healing.
         // All pool executors share this so recovery tracking is consistent.
@@ -243,6 +248,7 @@ impl RuntimePool {
             Some(op_manager.clone()),
             shared_contract_modules.clone(),
             shared_delegate_modules.clone(),
+            shared_delegate_contexts.clone(),
             None, // No shared backend yet — this executor creates the engine
         )
         .await?;
@@ -264,6 +270,7 @@ impl RuntimePool {
                 Some(op_manager.clone()),
                 shared_contract_modules.clone(),
                 shared_delegate_modules.clone(),
+                shared_delegate_contexts.clone(),
                 Some(shared_backend_engine.clone()),
             )
             .await?;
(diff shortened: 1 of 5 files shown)


iduartgomez:
I dont think is a good idea keeping an unbounded  warm cache forever in memory, this should be at the very least a LRU which persists on disk.

@claude wdyt?

iduartgomez (commented):
Recommended adding a TTL so delegates are not kept in memory indefinitively.

iduartgomez (changes requested):

sanity:
@iduartgomez fair point on unbounded warm cache — addressed in 4542b5a1:

- Each entry now stores its `last_write` `Instant`; `inbound_app_message` runs an amortised `prune_expired` sweep on every call.
- TTL is 10 min — comfortably longer than any interactive prompt round-trip (the prompt itself auto-denies at 60 s) and shorter than the lifetime of any real long-lived registered delegate, so genuine in-flight prompts can't expire prematurely. AGENTS.md GC rule on time-bounded exemptions is now satisfied.
- No background sweep task, no extra lock contention when idle.

I went with TTL rather than LRU + disk: the context is bound to a live UI session, so bytes loaded from disk after a restart would be stale state for a round-trip whose prompt is already gone — the delegate would mis-interpret them as a valid pending prompt. Pure LRU could evict legitimate slow-but-alive prompts. TTL has neither failure mode.

Also addressed the rule-review bot's warning about the missing test for the oversize-context drop path, plus a TTL-eviction regression test:

- `test_context_oversize_is_dropped_not_crashed` — writes past `DelegateContext::MAX_SIZE`, asserts the entry is dropped from the cache, asserts the next call sees an empty context (no crash).
- `test_context_ttl_evicts_stale_entry` — plants an entry with `last_write` predating `DELEGATE_CONTEXT_TTL`, drives a no-op call, asserts the stale entry is gone.

39 delegate tests pass; clippy clean against `cargo clippy --locked -- -D warnings` (the CI invocation).

[AI-assisted - Claude]

iduartgomez (approved):

PR #4058, opened 2026-05-08
fix(node): record pending_op_remove on event-loop drop

### LINKED ISSUE #4057: Flaky test: `test_pending_op_results_bounded` failing intermittently in CI Simulation
## Summary

`test_pending_op_results_bounded` in `crates/core/tests/simulation_integration.rs` is failing intermittently in CI Simulation, and the failure pattern suggests a real underlying issue rather than environmental noise. The test runs 3 times via nextest retries; recent failures show all 3 attempts failing.

## Evidence

**PR-level CI:** PR #4054 (a 4-line `scripts/release.sh` change with no Rust impact possible) failed `Simulation` at `test_pending_op_results_bounded` with `TRY 3 FAIL` on consecutive runs (run [25557416993](https://github.com/freenet/freenet-core/actions/runs/25557416993) at both 13:21 and 13:53 UTC on 2026-05-08).

**Nightly Simulation Tests:** Failing on consecutive runs since at least 2026-05-04:
- 2026-05-08 04:04 UTC, sha `b0cdb3a1` — failed at `test_pending_op_results_bounded`
- 2026-05-07 04:15 UTC, sha `ca7fa7a0` — failure
- 2026-05-06 04:15 UTC, sha `5f88756f` — failure
- 2026-05-05 04:03 UTC, sha `5f88756f` — failure
- 2026-05-04 04:17 UTC, sha `c8e6c3f4` — failure

**Main PR-level CI:** Has been GREEN on every recent commit. So the test is flaky-but-mostly-passing under the standard PR runner conditions, but reliably failing under whatever conditions PR #4054's runner had today and the nightly Large Scale Simulation runners have.

## Why this matters

1. **Already blocked one PR cycle.** PR #4054 had to be admin-merged because this test failed despite the PR not touching any Rust code.
2. **Per CLAUDE.md / `.claude/rules/flaky-tests.md`:** "Flaky tests are broken tests" — this needs proper investigation, not retries.
3. **Test threshold context:** `pending_op_results` leak threshold was bumped from 10 → 30 in #1454 phase 2c (commit `1764e961`, 2026-05-07). The same family of CONNECT-migration commits (`229793ea`, `1764e961`, `b0cdb3a1`) reshapes how operations register with `pending_op_results`. The test may be sensitive to the new flow even when the underlying behaviour is correct, OR there may be a real leak that only manifests under specific runner timing.

## Hypothesis to investigate

The test asserts a bounded count of `pending_op_results` slots after some workload. Recent CONNECT relay migration to task-per-tx (commit `b0cdb3a1`) may:
- Cause more concurrent ops to overlap in the relay path, pushing the count above 30 transiently before cleanup
- Have a missed `release_pending_op_slot` call somewhere in the new task-per-tx path
- Race with the slot release in a way that's more likely under loaded CI runners

The recently-merged PR #4051 (`d66c9db3`) added `record_relay_route_event` calls in relay drivers. Those don't touch `pending_op_results` directly but do hold `router.write()` synchronously — worth verifying the locking ordering doesn't push slot release later relative to the test's assertion point.

## How to reproduce

```bash
cd crates/core
cargo test -p freenet --features "simulation_tests,testing" --test simulation_integration test_pending_op_results_bounded -- --nocapture
# Run multiple times; flakiness should reproduce within a few attempts
```

If it doesn't reproduce locally, run on a slower / contention-heavy environment matching the GitHub runner profile.

## Suggested approach

Use the `freenet:systematic-debugging` skill — don't guess. Form 2-5 hypotheses, gather evidence in parallel (git log on `pending_op_results` related code, look at what commits between last-known-green and first-known-red touched the pending-op flow, check whether the leak threshold in the test is still right after recent migrations).

If the root cause is a real leak in one of the new task-per-tx drivers, fix it. If the test threshold is genuinely too tight for the new ops landscape, raise it AND add an upper-bound assertion so we still detect runaway leaks.

Do NOT add `#[ignore]` or retries to make the test green.

## Related

- Recent relay-driver migrations: #4042, #4043, #4046, #4048
- PR #4051 added route-event recording in relay drivers
- PR #4054 was unblocked by admin merge after this flake blocked it

[AI-assisted - Claude]

sanity:
## Problem

`test_pending_op_results_bounded` (#3100 regression guard) has been failing
intermittently in CI Simulation since 2026-05-04 and reliably in nightly Large
Scale Simulation runs. Recent failures show all 3 nextest retries failing,
and the flake already blocked one PR cycle (#4054 had to be admin-merged).

Closes #4057.

## Solution

The metric `record_pending_op_remove` was only fired on explicit removal
paths: `TransactionCompleted` / `TransactionTimedOut` events, the periodic
60s sweep, and the explicit `drain` on graceful `Shutdown`. Other event-loop
exit paths — `handle.abort()` from simulation teardown, `Disconnect` events,
unexpected stream end — dropped `state.pending_op_results` without recording
the corresponding removes. Inserts therefore exceeded removes by however
many entries were resident at exit, even though the memory itself was freed
when the map dropped. The "leak" was phantom: a metric accounting gap, not
real residency.

The leak threshold had been relaxed 10 → 30 → 60 to absorb this shortfall as
more relay drivers landed on the task-per-tx path; the test still flaked
when in-flight CONNECT relays at shutdown pushed the residual past 60.

The fix is a `Drop` impl on `EventListenerState` that records one
`pending_op_remove` per still-resident entry. Locally:

| | inserts | removes | leak |
|---|---|---|---|
| before fix (Drop disabled) | 1103 | 1041 | **62** |
| after fix (Drop enabled)   | 1103 | 1103 | **0** |

The 60 → 5 threshold tightening reverts the historical creep — that creep
was simulation-shutdown noise, not real leakage.

## Testing

New unit tests on `EventListenerState::Drop`:
- `event_listener_state_drop_records_remaining_pending_op_removes`: residency
  → Drop → metric balanced
- `event_listener_state_drop_empty_records_nothing`: empty drop is a no-op

The simulation regression guard `test_pending_op_results_bounded` was
re-tightened to `leak <= 5` and runs cleanly with leak=0 locally and under
nextest.

Verified the fix is load-bearing by toggling the `Drop` impl on/off and
re-running: leak goes from 62 (without) to 0 (with), reproducing the issue's
symptom and confirming this PR fixes it.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -4746,17 +4746,12 @@ impl EventListenerState {
 
 impl Drop for EventListenerState {
     fn drop(&mut self) {
-        // Mirror the explicit drain on `Shutdown` (see this file's
-        // pending_op_results.drain() block) for *all* event-loop exit
-        // paths — including `handle.abort()` from simulation teardown,
-        // `Disconnect` events, and unexpected stream end. Without this,
-        // entries still resident in `pending_op_results` at exit are
-        // dropped without their `record_pending_op_remove` accounting,
-        // creating a phantom "leak" in the metric used by #3100's
-        // regression guard `test_pending_op_results_bounded`. The
-        // memory is freed either way; only the metric was wrong.
-        let remaining = self.pending_op_results.len();
-        for _ in 0..remaining {
+        // Balance pending_op_results accounting on every event-loop exit path
+        // (graceful Shutdown, simulation `handle.abort()`, Disconnect,
+        // unexpected stream end). Memory is freed regardless; only the metric
+        // backing #3100's regression guard `test_pending_op_results_bounded`
+        // depends on this, so a missed remove looks like a phantom leak (#4057).
+        for _ in 0..self.pending_op_results.len() {
             crate::config::GlobalTestMetrics::record_pending_op_remove();
         }
     }
@@ -5455,12 +5450,10 @@ mod tests {
         );
     }
 
-    /// Regression guard for #4057: `EventListenerState::Drop` must record
-    /// a `pending_op_remove` for every entry still resident in
-    /// `pending_op_results` at the time the state is dropped. Without
-    /// this, simulation `handle.abort()` shutdowns inflate
-    /// `inserts - removes` in the global metric and trip the
-    /// `test_pending_op_results_bounded` regression guard for #3100.
+    /// Regression guard for #4057: `EventListenerState::Drop` must record one
+    /// `pending_op_remove` per resident entry, otherwise non-Shutdown exit paths
+    /// (simulation `handle.abort()`, etc.) leave `inserts - removes` inflated
+    /// and trip `test_pending_op_results_bounded` (#3100).
     #[test]
     fn event_listener_state_drop_records_remaining_pending_op_removes() {
         use crate::config::GlobalTestMetrics;
@@ -5469,7 +5462,6 @@ mod tests {
 
         let removes_before = GlobalTestMetrics::pending_op_removes();
 
-        // Build a state with a few pending entries, then drop it.
         let mut state = super::EventListenerState::new(ExpectedInboundTracker::empty_for_test());
         for _ in 0..3 {
             let (callback, _rx) = mpsc::channel(1);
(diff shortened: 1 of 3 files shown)


sanity:
## Review summary

Ran 4 internal review agents (code-first, testing, skeptical, big-picture) plus Codex CLI in parallel. Findings addressed in commit afde5456:

### Tightened threshold from `<= 5` to `== 0`

Skeptical, testing, and big-picture all converged on the same point: with the simulation runner on a single `current_thread` runtime (`crates/core/tests/simulation_integration.rs::create_runtime`), all peer event loops share one set of thread-local counters with the test reader. Tracing every reachable interleaving with the periodic 60s sweep, the explicit Shutdown drain, and the `TransactionCompleted`/`TimedOut` handlers shows the leak should be exactly zero. The `<= 5` slack was speculative; asserting `== 0` removes the slow-creep failure mode that drove the original 10 -> 30 -> 60 history.

Verified locally: leak=0 under both `cargo test` and `nextest`.

### Documented no-double-counting on Shutdown

Code-first review noted that two readers seeing both an explicit `drain()` block (Shutdown branch) AND the new Drop impl might worry about double-counting. Added a comment on the Drop impl explaining the order (drain happens first → map empty by Drop time → Drop is a no-op on graceful Shutdown).

### Items considered and not addressed

- **Threshold could be derived from `relay_drivers_in_flight_at_shutdown`** (testing reviewer): now moot — `== 0` is the derived bound.
- **Synthetic vs. integration test depth** (testing reviewer): the simulation test exercises the production `handle.abort()` → Drop path end-to-end and now asserts `leak == 0`, so a future regression that bypasses Drop would fail there. The unit tests pin the Drop body itself, which is small but worth a guard against accidental refactor (e.g., moving `pending_op_results` out of `EventListenerState`).
- **Production cost of the Drop impl** (big-picture reviewer): `record_pending_op_remove` is a thread-local cell increment. In production the Drop fires once at node shutdown for whatever entries remain — a few cycles per entry. No cfg gate needed.
- **Thread-affinity race in multi-threaded runtimes** (code-first reviewer): the simulation uses `new_current_thread()`, so all peers and the test reader share one set of counters. If a future test moves to `multi_thread`, the assertion would need a different shape.

[AI-assisted - Claude]

Issue #3074, opened 2026-02-17
PrioritySelectStream: client transaction channel (P7) starved under sustained load


sanity:
## Problem

The `PrioritySelectStream` in `priority_select.rs` uses strict priority ordering: if any higher-priority channel (P1-P6) has items ready, lower priorities are never polled. Client transaction registration is at **Priority 7** (second-lowest of 8), meaning under sustained network load it can be indefinitely starved.

This was the root cause of #3071 — PUT operations hung because the `wait_for_res_tx` channel (capacity 10) filled up and the P7 consumer never drained it.

PR #3073 added a 30s timeout and increased capacity to 100 as an immediate mitigation (fail loudly instead of hanging silently), but the structural problem remains: under sustained load, client operations will systematically time out rather than succeed.

## Current Priority Ordering

| Priority | Stream | Purpose |
|----------|--------|---------|
| 1 | Notification | Inbound net messages and node events |
| 2 | OpExecution | Sub-transaction execution responses |
| 3 | PeerConnection | Connection lifecycle events |
| 4 | ConnBridge | Outbound message bridge |
| 5 | Handshake | Connection handshake events |
| 6 | NodeController | Shutdown signals |
| **7** | **ClientTransaction** | **Client registration for result delivery** |
| 8 | ExecutorTransaction | Executor-initiated transactions |

## Possible Approaches

1. **Raise client transaction priority** — It's a lightweight metadata operation (HashMap insert into `tx_to_client`). Could be P2 or P3 since results must be registered *before* the operation starts.

2. **Fair scheduling** — After processing N items from higher priorities, force-poll lower priorities. Prevents indefinite starvation while preserving priority ordering.

3. **Starvation guard** — Track how long P7 hasn't been polled; force a poll if it exceeds a threshold.

## Related Issues

- The `pending_op_results` HashMap in `p2p_protoc.rs` (line 3685) is only `.insert()`-ed and `.get()`-ed, never `.remove()`-ed — slow memory leak.
- The FIXME at `p2p_protoc.rs:3680` about `tx_to_client` leaking is partially addressed by `TransactionTimedOut`/`TransactionCompleted` cleanup, but edge cases (client disconnect, send-side timeout from #3073) may leave orphaned entries.

[AI-assisted - Claude]

Issue #3224, opened 2026-02-23
fix: PrioritySelectStream starves handshake events under notification load


sanity:
## Problem

On loaded gateways (e.g., nova with ~1300 queued notification messages), the `PrioritySelectStream` event loop never polls the Handshake channel (P5), preventing new peers from joining the ring.

**Root cause:** `PrioritySelectStream` uses strict priority polling — when a higher-priority channel has items, lower channels are never polled. The Notification channel (P1) on a busy gateway is almost always Ready (capacity 2048, >50% full), so P5 (Handshake) is starved indefinitely.

**Observed symptoms:**
- Transport layer completes the connection (keepalive starts, "Detected new peer identity" logged)
- But `p2p_protoc` never receives the `InboundConnection` event
- The connecting peer times out and fails to join the ring
- First reported via diagnostic report DFUV6Q (Steven's macOS peer unable to join River room)

**The anti-starvation mechanism (MAX_HIGH_PRIORITY_BURST=32) only protects P7/P8** (client/executor transactions), NOT P5 (Handshake).

## Impact

Any peer trying to connect to a gateway with sustained notification traffic will be unable to join the network. This gets worse as the network grows (more notifications = more starvation).

## Solution

See PR #3225.

[AI-assisted - Claude]

PR #4059, opened 2026-05-08
fix(node): drain P7/P8 backlog in single burst cycle

### LINKED ISSUE #4056: fix(handler): nova gateway stuck in Priority 7 starvation; PUTs timing out at wait_for_res_tx
## Problem

The nova production gateway (`5.9.111.215:31337`, freenet-gateway-proxy + `freenet network` running since 2026-05-04) is in a sustained "Priority 7 starvation" state. All `freenet-git mirror` runs against this gateway are failing with `recv: client error: error while executing operation in the network: no response received from handler` on chunk PUTs, with no recovery between scheduled cron runs over the past 24+ hours.

## Evidence

`crates/core/src/contract/handler.rs:402-433` logs `"Timed out registering transaction for result delivery — event loop is not draining client transactions (Priority 7 starvation)"` with `channel_capacity=0`. Sample lines from `/home/freenet/.local/state/freenet/freenet.error.2026-05-08-14.log`:

```
2026-05-08T14:20:44.804784Z ERROR ... Timed out registering transaction for result delivery
  — event loop is not draining client transactions (Priority 7 starvation)
  client=10000061 channel_capacity=0 timeout_secs=30
2026-05-08T14:21:17.367966Z ERROR ... [same starvation] client=10000061 ... tx=01KR3ZCQW7F4Q3WTW1F07BA181
2026-05-08T14:21:51.929995Z ERROR ... [same starvation] client=10000061 ... tx=01KR3ZDSM979XG58YZCQTMPE81
```

These three errors are 30s apart — the wait_for_res_tx send is hitting `WAIT_FOR_RES_SEND_TIMEOUT = Duration::from_secs(30)` on every retry attempt from a `freenet-git put_pack` call (which retries 3 times with exponential backoff before giving up).

Concurrent symptoms in the same log window:

```
2026-05-08T14:20:25.220603Z  INFO ... Starting subscription renewal cycle needing_renewal=1 hosted=687
2026-05-08T14:20:25.220615Z  WARN ... Notification channel >50% full, reducing renewal batch size
                                      channel_remaining=0 channel_max=2048 batch_limit=2 contracts=1
2026-05-08T14:20:25.220639Z  WARN ... Notification channel >75% full during renewal spawning, stopping early
2026-05-08T14:21:47.711442Z  INFO ... router_snapshot failure_events=40 success_events=0
                                      prediction_active=false consider_n_closest_peers=5
```

The notification channel sits at `channel_remaining=0/channel_max=2048` (i.e. full) and the router has logged `failure_events=40 success_events=0` for 25+ minutes straight.

## Reproduction

```bash
# Against the affected gateway:
freenet-git rescue freenet:3GEERif5ihbf/freenet-core --ws-url <FREENET_GIT_WS_URL>
# → fails on PUT chunk N with "no response received from handler"
```

The `freenet-core` mirror workflow has been failing every push and every scheduled run since 2026-05-07 (`gh run list --repo freenet/freenet-core --workflow mirror-to-freenet.yml --limit 10`). The same gateway also affects the `freenet-stdlib` mirror and the daily `rescue-demos` cron in `freenet/freenet-git`.

## Hypothesis

The wait_for_res_tx channel (capacity bounded; depth=0 in error logs) and the notification channel (at 2048 max, 0 remaining) are both saturated. Anything generating outbound traffic from the contract handler keeps the notification channel full; the event loop's Priority 7 polling of wait_for_res_tx never gets scheduled because higher-priority work always has something ready. This is consistent with the channel-safety / biased-select rules in `.claude/rules/code-style.md` and `.claude/rules/channel-safety.md` — but the canonical fix (per-iteration caps, try_send) appears to already be applied at the obvious sites. The starvation is happening despite that, suggesting either a missing per-iteration cap somewhere on the producer side, or a slow downstream consumer that backpressures the notification channel without yielding.

## Operational impact

- All freenet-git push/rescue/clone operations against this gateway fail until the gateway is restarted or the queue naturally drains (no observed recovery in 24+ hours).
- v0.2.55 release pre-flight check was blocked until #4054 excluded `mirror / mirror` from the gate.
- River/Delta apps publishing through this gateway likely affected too (same handler path).

## Suggested next steps

1. **Immediate (operational):** Restart the nova gateway to clear the stuck state and confirm fresh runs succeed. This is low-risk for the demo URLs (their content is already published and the cron will repopulate caches; client clones use any reachable peer).
2. **Diagnosis:** While the gateway is in the bad state, capture a thread dump / tokio-console output to identify which task is starving Priority 7. The `Notification channel >50% full` warning at `crates/core/src/ring.rs` would be a starting point for the producer.
3. **Mitigation in core:** Audit notification-channel producers for per-iteration caps and try_send fallbacks per `.claude/rules/code-style.md`. Review whether `wait_for_res_tx` capacity should be larger or whether the polling priority of the consumer task should be raised.
4. **Mitigation in freenet-git (already filed):** freenet/freenet-git#36 proposes per-chunk retry with exponential backoff so transient gateway sickness doesn't fail the whole clone/push. Won't fix the gateway, but reduces the user-visible blast radius.

## Related

- `.claude/rules/code-style.md` "WHEN using `biased` in `tokio::select!`" — per-iteration caps mandatory
- `.claude/rules/channel-safety.md` "The Pattern That Keeps Killing Us" — bounded channels deadlocking event loops
- freenet/freenet-core#3279 — cross-node PUT timeouts
- freenet/freenet-core#3465 — Contract PUT operations report timeout/failure but succeed locally
- freenet/freenet-git#36 — per-chunk retry on transient timeout (helper-side mitigation)

[AI-assisted - Claude]

sanity:
## Problem

Issue #4056. Under sustained high-priority traffic on the production
gateway, the `wait_for_res_tx` channel (consumed by the event loop at
"Priority 7" via `PrioritySelectStream`) drained too slowly, causing
producer-side `waiting_for_transaction_result` calls to hit their 30 s
`WAIT_FOR_RES_SEND_TIMEOUT` and log:

```
Timed out registering transaction for result delivery —
event loop is not draining client transactions (Priority 7 starvation)
```

This was reproducible from a single client opening 3+ concurrent
WebSocket connections and firing one PUT per connection — the exact
shape of `freenet-git`'s 8-way parallel chunked-PUT pool used by the
`mirror-to-freenet` cron jobs. Every `freenet-core` mirror run since
2026-05-07 has failed for this reason.

## Root cause

Two separate factors compounded.

**1. Single-item P7 force-poll.** The anti-starvation mechanism in
`PrioritySelectStream::poll_next` (added in #3074) force-polls Tier-2
channels (P7 client transactions, P8 executor transactions) once
`high_priority_streak` reaches `MAX_HIGH_PRIORITY_BURST = 32`. But each
Phase 1 entry only released a *single* P7 item before resetting the
streak. Steady-state ratio: 1 P7 per 33 events (≈3 % of cycles). With
the event loop's per-peer outbound `.send()` capped at 500 ms (a
deliberate bridge-deadlock guard, p2p_protoc.rs:1090), one slow peer
could push P1 dispatch latency to ≈500 ms each. At that rate,
draining 100 P7 entries took ≈50 s — easily past the 30 s producer
timeout.

**2. Anomalously small `wait_for_res_tx` capacity.** 100 — 5–20× smaller
than every other event-loop-side channel (`result_router_tx` 1000,
`notification` 2048, `conn_event` 1024, `bridge_cmd` 512). No comment
explaining the choice. Concurrent multi-WebSocket clients can easily
exceed 100 in-flight registrations during a P1 burst.

## Solution

**Drain P7/P8 fully into per-stream `VecDeque` buffers in a single Phase 1
admission, then return queued items with absolute priority (Phase 0) on
subsequent `poll_next` calls.** This raises the steady-state P7 share
from `1/(burst+1)` to `N/(N+burst)` where N is the queued backlog —
the entire backlog clears in `burst + N` events instead of `burst × N`.
P7 items are HashMap inserts (microseconds each), so cluster-draining
them does not measurably delay P1 traffic.

**Bump `wait_for_res_tx` capacity from 100 → 1000** to match
`result_router_tx`. Defense in depth — the drain-buffer change is the
load-bearing fix; the bump just gives more headroom against future
workload growth.

Both changes apply equally to P8 per the "Finish The Fix" rule (P8 had
the same one-item-per-burst defect).

## Testing

- New `test_p7_backlog_drains_in_one_burst_cycle` regression test in
  `priority_select/tests.rs`. Pre-fills 5 000 P1 + 200 P7 messages and
  asserts `last_p7_idx <= burst + P7_COUNT`. Without the fix, the last
  P7 item lands at index ≈5 199 (>> 232) and the assertion fails;
  with the fix, P7 backlog clears at index 232.
- Updated existing `waiting_for_transaction_result_times_out_when_channel_full`
  to read the channel's actual `max_capacity()` rather than hard-coding
  100, so it survives future capacity changes.
- All existing anti-starvation tests
  (`test_anti_starvation_p7_serviced_under_load`,
  `test_anti_starvation_counter_reset_on_tier2`,
  `test_anti_starvation_high_load_fairness`,
  `test_anti_starvation_includes_handshake`) continue to pass — the
  drain-buffer behaviour clusters P7 items into one contiguous run
  per burst window, which still satisfies their max-tier-1-run and
  first-tier-2-index invariants.

## Fixes

Closes #4056.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/handler.rs
+++ b/crates/core/src/contract/handler.rs
@@ -281,7 +281,12 @@ pub(crate) fn contract_handler_channel() -> (
     ContractHandlerChannel<WaitingResolution>,
 ) {
     let (event_sender, event_receiver) = mpsc::unbounded_channel();
-    let (wait_for_res_tx, wait_for_res_rx) = mpsc::channel(100);
+    // Capacity sized to comfortably absorb concurrent multi-WebSocket bursts
+    // (e.g. freenet-git's 8-way parallel chunked PUT pool) while the event
+    // loop's anti-starvation force-poll drains it. Smaller values caused
+    // 30s producer-side timeouts on the production gateway under realistic
+    // mirror traffic. See issue #4056.
+    let (wait_for_res_tx, wait_for_res_rx) = mpsc::channel(1000);
     (
         ContractHandlerChannel {
             end: SenderHalve {
@@ -962,9 +967,10 @@ pub mod test {
 
         let (send_halve, _rcv_halve, _wait_res) = contract_handler_channel();
 
-        // Fill the channel to capacity (100 items). We hold _wait_res so
+        // Fill the channel to capacity. We hold _wait_res so
         // the receiver isn't dropped (which would give a different error).
-        for _ in 0..100 {
+        let cap = send_halve.end.wait_for_res_tx.max_capacity();
+        for _ in 0..cap {
             let tx = Transaction::new::<PutMsg>();
             send_halve
                 .end
@@ -974,7 +980,7 @@ pub mod test {
                 .expect("channel should accept items up to capacity");
         }
 
-        // The 101st send should block, and with paused time we can
+        // The next send should block, and with paused time we can
         // advance past the timeout instantly.
         let tx = Transaction::new::<PutMsg>();
         let result = send_halve
--- a/crates/core/src/node/network_bridge/priority_select.rs
+++ b/crates/core/src/node/network_bridge/priority_select.rs
@@ -97,6 +97,16 @@ where
     /// O(N × burst). See `poll_next` Phase 1 for the full rationale (#4056).
     p7_buffer: VecDeque<(ClientId, WaitingTransaction)>,
     p8_buffer: VecDeque<Transaction>,
+
+    /// Closure of the P7/P8 channels detected during a Phase 1 drain that
+    /// also collected items into the buffer. We defer emitting the
+    /// `SelectResult::*(Err(channel closed))` notification until the
+    /// buffer drains; otherwise the buffer return preempts the closure
+    /// signal and the consumer never learns the channel closed (the
+    /// `client_transaction_closed` / `executor_transaction_closed` flags
+    /// keep us from re-polling, so the signal would be lost forever).
+    p7_close_pending: bool,
+    p8_close_pending: bool,
 }
 
 impl<H, C, E> PrioritySelectStream<H, C, E>
@@ -138,6 +148,8 @@ where
             high_priority_streak: 0,
             p7_buffer: VecDeque::new(),
             p8_buffer: VecDeque::new(),
+            p7_close_pending: false,
+            p8_close_pending: false,
         }
     }
 
@@ -159,8 +171,27 @@ where
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.get_mut();
 
-        // Phase 0: Return any P7/P8 items batch-drained by a previous
-        // Phase 1 force-poll. Absolute priority — see Phase 1 below (#4056).
+        // Phase 0: Return P7/P8 items batch-drained by a previous Phase 1
+        // force-poll. Absolute priority over Tier-1 — see Phase 1 below
+        // (#4056). Yield to handshake first, though, so a long buffer drain
+        // doesn't delay connection-lifecycle events past the bound added
+        // in #3224.
+        let buffer_has_items = !this.p7_buffer.is_empty() || !this.p8_buffer.is_empty();
+        if buffer_has_items && !this.handshake_closed {
+            match Pin::new(&mut this.handshake_handler).poll_next(cx) {
+                Poll::Ready(Some(event)) => {
+                    this.high_priority_streak = 0;
+                    return Poll::Ready(Some(SelectResult::Handshake(Some(event))));
+                }
+                Poll::Ready(None) => {
+                    this.handshake_closed = true;
+                    // Don't preempt buffer drain with the closure signal;
+                    // it'll be re-discovered by Phase 1/2 once the buffer
+                    // empties.
+                }
+                Poll::Pending => {}
+            }
+        }
         if let Some(item) = this.p7_buffer.pop_front() {
             this.high_priority_streak = 0;
             return Poll::Ready(Some(SelectResult::ClientTransaction(Ok(item))));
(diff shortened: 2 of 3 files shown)


sanity:
Pushed review-feedback fixes (commit 7df787ad). Addresses:

**Codex P1 / Skeptical Med #3 / Code-First #4 — Phase 0 blocked handshake polling.**
Phase 0 now yields to handshake first when the buffer has items, restoring the bounded handshake-latency guarantee from #3224. New test: `test_handshake_yields_during_p7_buffer_drain` (uses a gated handshake stream that holds back the event until P7 buffer drain has started; fails at index ≈84 vs ≤41 without the yield).

**Skeptical Med #1 — Channel-closed signal lost when buffer non-empty.**
Phase 1's drain loop now stashes a sticky `p7_close_pending` / `p8_close_pending` when items were buffered before the closure was detected; Phase 0 emits the `Err(channel closed)` after the buffer drains. New test: `test_p7_close_signal_emitted_after_buffer_drain` (fails without the sticky flag).

**Skeptical Med #2 — P7/P8 fairness in Phase 1.**
P7 and P8 are now both drained in a single Phase 1 admission (the previous code returned after P7 if any were buffered, leaving P8 to Phase 2). New test: `test_p8_backlog_drains_in_one_burst_cycle` (mirrors the P7 test, locks in symmetric behavior).

**Big-picture / Code-First docs concerns:** the `priority_select.rs` module now references #4056 inline at the Phase 0 / Phase 1 sites; a `bug-prevention-patterns.md` row for "anti-starvation force-poll must drain, not single-item" is reasonable but I'd prefer to do that separately so this PR stays focused on the runtime fix.

**Pre-existing flake:** `test_pending_op_results_bounded` failed CI with `leak=62 > 60`; reproduces on `main` with identical leak count (62). The threshold was bumped 10→30→60 over the past few months; this is bumping again into "needs proper fix" territory but is unrelated to this PR's scope.

[AI-assisted - Claude]

PR #4062, opened 2026-05-09
refactor(ops): retire PUT legacy state machine + ops.put DashMap (#1454 phase 5 final, PUT slice)

iduartgomez:
## Problem

#1454 migrates each operation from the centralized re-entry loop
(`OpManager.ops.{op}` DashMap + `handle_op_result` + `process_message`)
to per-transaction async drivers (`op_ctx_task`). The UPDATE slice
(PR #4053) retired UPDATE's legacy state machine; PUT was the
natural next slice (smallest legacy surface after UPDATE; all wire
paths already migrated to task-per-tx).

This PR closes the loop for PUT by deleting the legacy state machine
+ `ops.put` DashMap + `OpEnum::Put` carrier, mirroring the UPDATE
slice's structure.

## Solution

After this PR:

* Every PUT wire variant dispatches unconditionally to a task-per-tx
  driver (`start_client_put` / `start_relay_put` /
  `start_relay_put_streaming`).
* `OpEnum::Put`, `OpManager.ops.put` DashMap, `impl Operation for
  PutOp`, `has_put_op`, `remove_put_and_report_failure`, the
  `OpEnum::Put` arm in `IsOperationCompleted for OpEnum`, the
  `OpEnum::Put` abort arm, the `try_from_op_enum!(OpEnum::Put, ...)`
  macro entry, and the `handle_op_request<PutOp>` fallthrough in
  node.rs are all gone.
* The `!has_put_op(id)` dispatch gate was redundant once the
  DashMap could no longer be populated; node.rs's PUT bypass match
  is simplified to a `Some(upstream_addr) = source_addr` guard.
* `PutOp`, `PutState`, `PutStats`, `AwaitingResponseData`,
  `FinishedData` and 25 inline outcome / failure-routing /
  wire-format / pin tests survive under `#[allow(dead_code)]`
  pending phase 6 (matches the UPDATE slice's pattern).

### Pre-implementation audit collapsed commit 1

The audit-doc plan included a "commit 1: retire executor PUT
network branch" step. The audit confirmed:

* No `OperationMode::Network` constructor exists in the tree
  (matches the UPDATE finding).
* `perform_contract_put` is fully local (no `if mode == Network`
  branch, no `op_request(PutContract)` call). PUT has no
  `PutContract` / `ComposeNetworkMessage<PutOp>` to retire.

So the executor cleanup commit was empty and is folded into this
PR's single commit.

### Upgrade-on-forward edge case

Slice A's `start_relay_put` driver does not yet perform the
legacy `should_use_streaming(payload_size)` re-check +
upgrade-to-streaming on forward. Inbound non-streaming `Request`s
whose serialized payload would exceed `streaming_threshold` on the
forward used to fall through to legacy; with legacy retired, the
dispatch site in node.rs drops them with a `tracing::warn!` rather
than forwarding unstreamed.

Originators running task-per-tx (`start_client_put`) pre-serialize
and choose `RequestStreaming` when payload > threshold, so this
only fires for peers running pre-#1454 builds. A follow-up slice
can teach `start_relay_put` to upgrade-on-forward; until then the
edge surfaces as a sender-side timeout, not a silent unstreamed
forward.

This is documented in `.claude/rules/operations.md` and called out
in the new `phase5-get-audit.md` scoping doc.

## Testing

* `cargo test -p freenet --lib` passes (2561 tests, 16 ignored).
* `cargo clippy -p freenet -- -D warnings` clean for changed code
  (a pre-existing `wildcard_enum_match_arm` in
  `priority_select/tests.rs:2632` is on main and out of scope).
* The new bypass match dispatch logic is pinned by the rewritten
  `bypass_is_wired_into_put_branch_regression_guard` and
  `put_branch_dispatches_relay_drivers` regression tests in
  node.rs (composed needles at runtime so the assert source
  doesn't trip itself).
* The `completed_must_remove_per_type_dashmap_entry` pin in
  `put.rs::tests` was rewritten: drops `self.ops.put` from the
  required list, adds a positive assertion that `ops.put` is no
  longer referenced by `OpManager::completed`.

## Net change

* `crates/core/src/operations/put.rs`: -1166 lines
* `crates/core/src/node.rs`: -160 lines (fallthrough + abort arm
  + IsOperationCompleted arm + regression tests rewritten)
* `crates/core/src/node/op_state_manager.rs`: -85 lines (DashMap
  field + push/pop/peek arms + has_put_op +
  remove_put_and_report_failure + GC sweep callsites)
* `crates/core/src/operations.rs`: -3 lines (OpEnum::Put + macro)
* Docs: `+90 phase5-get-audit.md` (new, supersedes
  `phase5-put-audit.md` which is deleted), `+25 operations.md`,
  `+12 CLAUDE.md`.

Total: ~-1245 LoC code + ~+90 LoC docs.

## Fixes

Part of #1454 (phase 5 final, PUT slice). Subsequent slices: GET,
SUBSCRIBE, CONNECT.


iduartgomez:
Closing this PR: the audit missed that task-per-tx coverage for originator-side streaming PUT is incomplete. `start_client_put`'s retry driver always builds non-streaming `PutMsg::Request` regardless of payload size; the legacy `process_message` upgrade-on-forward path was the only place upgrade happened. Retiring legacy without first migrating the originator-side upgrade (or adding upgrade-on-forward to `start_relay_put`) breaks all PUTs >`streaming_threshold`.

Plan revisions needed:
1. Add upgrade-on-forward to `start_relay_put` (~80-100 LoC, requires threading NetworkBridge through), OR
2. Migrate `start_client_put` to choose between Request and RequestStreaming at `build_request` based on payload size (cleaner but larger)
3. Then re-attempt the legacy retirement.

The phase5-put-audit.md needs the originator streaming gap added as a prerequisite. Will refile after that work.

Issue #4066, opened 2026-05-09
Get-with-subscribe + return_contract_code doesn't reliably populate caller's local store


iduartgomez:
## Summary

In a 2-node split test (both nodes on the same host, distinct data dirs / ports / log dirs, both connected to public gateways) we see `Get { return_contract_code: true, subscribe: true }` from peer1 for a contract hosted on peer2 either time out or return \"Contract not found\" even though peer2 is reachable in the ring and is itself a subscriber to the same contract.

When the GET fails this way, a subsequent client-initiated `Update { Delta }` on peer1 fails locally before forwarding with:

\`\`\`
ERROR ... freenet::operations::update::op_ctx_task: update (task-per-tx): failed to apply update locally before forwarding
  contract=<key> error=put error for contract <key>, reason: missing contract parameters phase=\"error\"
\`\`\`

That's the same family as freenet/freenet-core#... \"missing contract parameters\" we hit when forwarding through a non-hosting node — but the workaround there (have the webapp prime the contract via Get with subscribe before sending) only works if the prime Get actually reaches a peer that's hosting the contract.

## Repro shape (freenet-email)

- peer1 (sender) and peer2 (recipient) both `freenet network`, joined to the public network via the same gateways.
- peer2 hosts contract `G42wRGFHyEjRrJpSJZJYfdN9B7Hgz91cXA6jjh5JCepL` (an inbox), confirmed by repeated `Subscription renewal succeeded contract_key=...` lines in peer2's log.
- peer1 sends `Get { return_contract_code: true, subscribe: true }` for that contract.
- Sometimes the GET completes and contract code arrives. Sometimes:
  - peer1 logs `Contract not found while processing a get request ... peer_addr=<self>@... sender=None phase=\"forward\"`, then `GET operation timed out without receiving a response ... elapsed_ms=~64000 phase=\"get_timeout\"`.
  - peer2 never sees a corresponding GET request in its log, even though it's the host.
- Subscribe in the same request still completes (peer2 logs `subscribe ... outcome=\"subscribed\"` with peer1's IP), so the routing path peer1→peer2 is alive — but the GET portion isn't reaching peer2.

## What I'd expect

A `Get` against a contract that any reachable peer hosts should either (a) find it via the ring, or (b) fail consistently rather than depending on which peer the originator happens to reach via the proximity-based forwarding path.

[...]

PR #4070, opened 2026-05-09
fix(get): clear ops.get on every retry-loop exit path (#4069)

### LINKED ISSUE #4069: GET task-per-tx driver leaks legacy ops.get entry on non-Done exit -> phantom phase=get_timeout (subset of #4066)
## Summary

Subset of #4066. The task-per-tx client GET driver
(`drive_client_get_inner` + `drive_sub_op_get` in
`crates/core/src/operations/get/op_ctx_task.rs`) only calls
`op_manager.completed(tx)` on the `RetryLoopOutcome::Done` arm. The
`Exhausted` / `Unexpected` / `InfraError` arms leak the legacy
`ops.get` entry that the originator-loopback path created.

## Why an entry exists at all

`start_client_get` -> `drive_retry_loop` -> `OpCtx::send_and_await`
with `target=None` loops the `GetMsg::Request` back as an
`InboundMessage` with `source_addr=None`. The relay-driver dispatch
at `node.rs::handle_pure_network_message_v1` is gated on
`source_addr.is_some()`, so the loopback falls through to the legacy
`handle_op_request` -> `process_message::GetMsg::Request` arm. For
a non-hosting originator the contract isn't local, so
`try_forward_or_return` runs and `op_manager.push(client_tx, GetOp)`
inserts an entry into `ops.get`.

## Symptom

- Driver per-attempt timeout = `OPERATION_TTL` = 60s
  (`op_ctx.rs:520`).
- Legacy GC sweep cleanup = `OPERATION_TTL` = 60s
  (`op_state_manager.rs:remove_get_and_report_failure`).

Both timers fire at the same value. Network jitter on a public DHT
multi-hop path puts responses near the 60s boundary, so the GC
sweep frequently races the driver and emits a misleading
`phase=get_timeout` log line plus a duplicate
`OperationError("GET operation timed out")` to the client. The
`Contract not found while processing a get request phase=forward`
log line that #4066 reports also originates from the
originator-loopback `try_forward_or_return` execution.

## Reproduction

`cargo test -p freenet --lib drive_client_get_inner_completes_op` with
the unfixed driver fails on the source-scrape regression because the
non-Done arms don't call `op_manager.completed`.

## Fix

Call `op_manager.completed(tx)` on every retry-loop exit path in both
`drive_client_get_inner` and `drive_sub_op_get`. Pin behaviour with
source-scrape regression tests.

PR opens shortly: #TBD.

## Out of scope

This issue addresses only the phantom `phase=get_timeout` /
duplicate-error symptom on the originator side. The wire-format
"missing contract parameters" failure in UPDATE-from-non-hosting-peer
documented in #4066 is tracked separately (will follow with a
companion fix).

iduartgomez:
## Problem

The task-per-tx client GET driver (`drive_client_get_inner` +
`drive_sub_op_get` in `crates/core/src/operations/get/op_ctx_task.rs`)
only calls `op_manager.completed(tx)` on the `RetryLoopOutcome::Done`
arm. `Exhausted` / `Unexpected` / `InfraError` leak the legacy
`ops.get` entry created by the originator-loopback Request, which the
GC sweep later reports as a phantom `phase=get_timeout`.

See #4069 for the full failure-mode analysis. Subset of #4066.

### Why an entry exists

`start_client_get` -> `drive_retry_loop` -> `OpCtx::send_and_await`
with `target=None` loops back as an `InboundMessage` with
`source_addr=None`. The relay-driver dispatch in
`node.rs::handle_pure_network_message_v1` is gated on
`source_addr.is_some()`, so the loopback falls through to legacy
`handle_op_request` -> `process_message::GetMsg::Request`. For a
non-hosting originator the contract isn't local, so
`try_forward_or_return` runs and pushes a `GetOp` into
`OpManager.ops.get` keyed by the client tx.

### Why the race fires

- Driver per-attempt timeout = `OPERATION_TTL` = 60s.
- Legacy GC sweep cleanup = `OPERATION_TTL` = 60s.

Same value. Network jitter near the 60s boundary causes the GC sweep
to fire before the driver finishes; `remove_get_and_report_failure`
emits `phase=get_timeout` to telemetry and a duplicate
`OperationError("GET operation timed out")` to the client via
`result_router_tx.try_send`.

## Solution

Call `op_manager.completed(tx)` on every retry-loop exit path in
both drivers. The call is idempotent so it's safe on paths where the
driver never registered a legacy entry.

Source-scrape regression tests pin the cleanup in every arm so a
future refactor can't silently drop it. The tests fail without the
fix:

```
drive_client_get_inner RetryLoopOutcome::Exhausted arm must call
op_manager.completed(client_tx) to clear the legacy ops.get entry
created by the originator-loopback Request — see #4066. Without
this the GC sweep races the driver and emits a phantom
`phase=get_timeout` log + duplicate OperationError.
```

## Test plan

- [x] `cargo test -p freenet --lib drive_client_get_inner_completes_op` (regression: passes after fix, fails before)
- [x] `cargo test -p freenet --lib drive_sub_op_get_completes_op`
- [x] `cargo test -p freenet --lib operations::get::` (112 passed)
- [x] `cargo clippy -p freenet --lib --tests -- -D warnings`
- [x] `cargo fmt`

## Drive-by

First commit fixes two pre-existing clippy warnings on `main`
(`doc_lazy_continuation` in `put/op_ctx_task.rs` and
`wildcard_enum_match_arm` in `priority_select/tests.rs`), without
which `cargo clippy -- -D warnings` fails on a clean checkout. They
were introduced in PR #4063.

## Fixes

Closes #4069. Partial fix for #4066 (the wire-format
"missing contract parameters" UPDATE failure is tracked separately,
companion PR to follow).

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -165,12 +165,32 @@ paths:
 > outer timeout (e.g., `RELATED_FETCH_TIMEOUT = 10s` wrapping
 > `local_state_or_from_network`'s 120 s receiver timeout) — receiver
 > drop is silent so no leak, just a longer-lived background task.
-> After this migration, no caller pushes a client-initiated `GetOp`
-> into `ops.get`, so the GC speculative-retry block was provably dead
-> for GET. Phase 5-final retired that block, the `get_retried`
-> per-tick HashMap, the `GetOp::ack_received` /
-> `GetOp::speculative_paths` fields, and the 11 unit tests that
-> simulated the GC retry decision.
+> After this migration, no caller PURPOSEFULLY pushes a
+> client-initiated `GetOp` into `ops.get`, so the GC
+> speculative-retry block was provably dead for GET. Phase 5-final
+> retired that block, the `get_retried` per-tick HashMap, the
+> `GetOp::ack_received` / `GetOp::speculative_paths` fields, and the
+> 11 unit tests that simulated the GC retry decision.
+>
+> **Loopback corner case (#4066):** the originator's first
+> `OpCtx::send_and_await(target=None)` loops back as an
+> `InboundMessage` with `source_addr=None`. The relay-driver
+> dispatch in `node.rs::handle_pure_network_message_v1` is gated on
+> `source_addr.is_some()`, so the loopback Request falls through to
+> legacy `process_message::GetMsg::Request` and incidentally pushes
+> a `GetOp` into `ops.get` keyed by `client_tx`. The entry's state
+> is never read (the bypass at the top of
+> `handle_pure_network_message_v1` intercepts terminal Responses
+> before `handle_op_request` runs), but the GC sweep at
+> `OPERATION_TTL=60s` still reaps it and used to emit a phantom
+> `phase=get_timeout` log + duplicate `OperationError`. Closed by
+> `ClientGetCompletionGuard` / `SubOpGetCompletionGuard` in
+> `get/op_ctx_task.rs`: a Drop guard on the driver task calls
+> `op_manager.completed(tx)` on every exit path including panics,
+> clearing the legacy entry before the GC sweep can race it. A
+> cleaner architectural fix would be a dedicated thin handler for
+> originator-loopback Requests that emits the Request-echo without
+> inserting into `ops.get` at all — phase 6 cleanup item.
 >
 > The SUBSCRIBE GC speculative-retry block was retired in the same
 > Phase 5-final pass (slice 1 after GET): all four originator-side
--- a/crates/core/src/node/network_bridge/priority_select/tests.rs
+++ b/crates/core/src/node/network_bridge/priority_select/tests.rs
@@ -2629,7 +2629,7 @@ async fn test_handshake_yields_during_p7_buffer_drain() {
                     self.counter.fetch_add(1, Ordering::SeqCst);
                     Poll::Ready(Some(item))
                 }
-                other => other,
+                other @ Poll::Ready(None) | other @ Poll::Pending => other,
             }
         }
     }
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -447,13 +447,29 @@ async fn drive_client_get_inner(
             Ok(DriverOutcome::Publish(host_result))
         }
         RetryLoopOutcome::Exhausted(cause) => {
+            // Originator's first `send_and_await(target=None)` loops back as an
+            // InboundMessage with `source_addr=None`, which falls through to
+            // legacy `process_message` (node.rs gates relay-driver dispatch on
+            // `source_addr.is_some()`) and pushes a `GetOp` into `ops.get`
+            // keyed by `client_tx`. Without this `completed()` call, the legacy
+            // entry lingers until the periodic GC sweep at OPERATION_TTL=60s,
+            // which races the driver's per-attempt timeout (also 60s) and
+            // emits a misleading `phase=get_timeout` log + duplicate
+            // `OperationError` to the client (#4066).
+            op_manager.completed(client_tx);
             Ok(DriverOutcome::Publish(Err(ErrorKind::OperationError {
                 cause: cause.into(),
             }
             .into())))
         }
-        RetryLoopOutcome::Unexpected => Err(OpError::UnexpectedOpState),
-        RetryLoopOutcome::InfraError(err) => Err(err),
+        RetryLoopOutcome::Unexpected => {
+            op_manager.completed(client_tx);
+            Err(OpError::UnexpectedOpState)
+        }
+        RetryLoopOutcome::InfraError(err) => {
+            op_manager.completed(client_tx);
+            Err(err)
+        }
     }
 }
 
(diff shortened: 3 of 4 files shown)


iduartgomez:
## Skeptical Review

### Medium Severity

**1. Source-scrape regression tests have faulty arm-bounding heuristic for the LAST arm.**
In `drive_client_get_inner_completes_op_on_every_exit_path` and the sub-op variant, `arm_end` falls back to `tail.len()` when no further `RetryLoopOutcome::` is found. For the last arm (`InfraError`), `arm_body` extends all the way to end-of-fn body. If a future refactor adds *any* `op_manager.completed(` call later in the fn (e.g., delivering side-effects after the match) without one inside the InfraError arm, the test passes spuriously. Tighten by clipping each arm at its own `=>` body or scan with a syn-based AST walk. The test is also brittle to the literal substring `op_manager.completed(` — bind the OpManager to a local (`let mgr = op_manager; mgr.completed(...)`) and the test fails despite the cleanup being correct.

**2. Panic / abort path still leaks.**
If `drive_retry_loop` (or any await inside it: `process_message`, `cache_contract_locally`, etc.) panics, the spawned task aborts before reaching the match block, and `op_manager.completed()` is never called. The legacy `ops.get` entry leaks until the GC sweep at OPERATION_TTL, re-emitting the same `phase=get_timeout` race the PR claims to fix. A `scopeguard`/RAII drop guard around the whole driver body would close this — same pattern that the relay drivers' `Relay*InflightGuard` use. Not a regression introduced by this PR, but the PR is explicitly framed as eliminating the phantom log, so a panic-survivable cleanup belongs in scope.

**3. `completed()` ordering vs. side-effects in Done arm at line 278 (pre-existing, but now mirrored on failure arms).**
On Done, `completed()` runs BEFORE `cache_contract_locally`, `auto_subscribe_on_get_response`, `register_events`, and `maybe_subscribe_child` (lines 286–445). Once `completed()` runs, the tx is in `ops.completed`, and any of those follow-on awaits that re-enter `op_manager.push(...)` for the same tx are silently dropped at line 698. I don't see a push call in those side effects, but `auto_subscribe_on_get_response` and the event log path are non-trivial — worth confirming none of them assume the tx is still alive. The new failure arms have no follow-on side effects so this is moot for them.

### Low Severity

**4. Fix masks the real architectural issue.**
The root cause is the gate at `node.rs:1359` (`source_addr.is_some()`): the originator-loopback Request deliberately falls through to legacy `process_message` because `drive_client_get_inner`'s `classify` relies on the Request-echo for `Terminal::LocalCompletion`. The legacy path's only side effect on the originator is to push a `GetOp` into `ops.get` whose state will never be read (the bypass at `node.rs:1324` intercepts the terminal Response before `handle_op_request` runs). The leak exists *because* we fall through to legacy at all. A cleaner fix: route originator-loopback Requests to a dedicated thin handler that emits the Request-echo without inserting into `ops.get`. The current fix is correct and minimal but adds another item to the eventual phase-6 cleanup list.

**5. `try_send` on `result_router_tx` at op_state_manager.rs:1386 is a separate latent bug.**
If the GC ever does fire for a real timeout (independent of this PR), the duplicate-error semantics rely on `try_send`. If the channel is full, the client never gets the timeout — silent failure, exactly the channel-safety pattern the codebase warns about. Not in scope, but the PR description says "duplicate `OperationError` to the client" — confirming the GC path uses `try_send` means under load the user could see EITHER zero, one, or two errors depending on channel pressure.

**6. The `completed.contains()` short-circuit in `push()` (line 698) prevents resurrection of the entry after `completed()` is called, but the GC older-than-TTL block at op_state_manager.rs:1717 always proceeds to `remove_get_and_report_failure` even if `ops.completed.remove(&tx)` returned Some.**
Currently safe because `ops.get.remove` returns None after `completed()`, so the failure-routing/log/result-router branch is skipped and `removed=false`. But this is non-obvious and depends on tight coupling between the two maps. If a future refactor stores a "tombstone" in `ops.get`, the phantom log returns. Worth a comment at line 1740 noting the dependency.

### Questions for Author

1. Why not wrap the entire driver body in a `scopeguard::defer!` that calls `completed()` on drop, eliminating the per-arm placement entirely and surviving panics? It's idiomatic and removes the regression-test brittleness.
2. The PR description claims `completed()` is idempotent and "safe on paths where the driver never registered a legacy entry." Confirmed at op_state_manager.rs:939–980 — but the `live_tx_tracker.remove_finished_transaction`, `under_progress.remove`, and `request_router.complete_operation` calls all run unconditionally on every invocation. For task-per-tx GETs that never registered, are any of these unintended side effects? E.g., does `request_router.complete_operation` clear router state another in-flight op might be reading?
3. Is there any concurrent path where `start_relay_get` (relay driver) and `drive_client_get_inner` (client driver) could share a tx? If a node is *both* originator and relay for the same tx (sub-op recursion), the order of `completed()` calls could clear state the other driver still needs.

### What Verified

- `op_manager.completed()` at op_state_manager.rs:939 explicitly calls `self.ops.get.remove(&id)` for `TransactionType::Get` — cleanup correct for the targeted leak.
- The GC `delayed`/short-TTL block (line 1652) checks `ops.completed.remove(&tx)` first and `continue`s; combined with `completed()` inserting into `ops.completed`, the phantom log is suppressed.
- The older-than-TTL block (line 1740) calls `remove_get_and_report_failure`, no-op when `ops.get.remove` returns None — also suppressed after `completed()`.
- `push()` at line 698 rejects re-insertion after `completed()`, so delayed loopback Request can't re-leak.
- Done arm at line 278 already calls `completed()`; Phase-5-final removal of GC speculative retries means no other GET originator path writes to `ops.get`.
- priority_select clippy fix `other @ Poll::Ready(None) | other @ Poll::Pending` exhaustive over remaining `Poll<Option<T>>` variants and behaviorally identical to wildcard.
- put/op_ctx_task.rs doc fix benign whitespace-only `///` blank line for `doc_lazy_continuation`.
- No early `?` returns in `drive_client_get_inner` between fn entry and `drive_retry_loop` invocation, so new arm placement covers all retry-loop exit cases.

_Posted by skeptical-reviewer agent._

PR #4072, opened 2026-05-09
fix(update): auto-fetch contract on non-hosting originator (#4071)

### LINKED ISSUE #4071: UPDATE from non-hosting originator returns 'missing contract parameters' instead of self-healing (subset of #4066)
## Summary

Subset of #4066. When peer1 (non-hosting) sends
`ContractRequest::Update { update_data: Delta(..) }` for a contract
hosted on peer2, the client-initiated UPDATE driver
(`drive_client_update` in
`crates/core/src/operations/update/op_ctx_task.rs:270-313`) tries to
apply the update locally before forwarding. The local executor lacks
the contract's code and params, so `upsert_contract_state` fails:

```
ERROR ... update (task-per-tx): failed to apply update locally before forwarding
      contract=<key>
      error=put error for contract <key>, reason: missing contract parameters
      phase="error"
```

The UPDATE never forwards. peer2 never sees it. peer1's UI shows the
message in Sent but the recipient stays empty.

## Why local apply is load-bearing

The wire format `UpdateMsg::RequestUpdate.value: WrappedState` carries
a post-merge full state, not a delta. The originator must produce
one before forwarding, which requires the contract's params to merge
the delta. A raw-delta wire format would be a cleaner fix but it's a
wire-format change requiring a stdlib bump and `min-compatible-version`
floor.

## Why the existing self-heal doesn't help

`OpManager::try_auto_fetch_contract` already exists at
`update.rs:286-369` and is invoked from the **relay** paths
(`drive_relay_broadcast_to` line 1026, `drive_relay_request_update`
line 1638). It's NOT invoked from the client-initiated remote-target
branch.

## Fix

When `update_contract` fails with an error where
`is_contract_exec_rejection() == false` (i.e. not a real WASM merge
rejection — the only production case is "missing contract
parameters"), call `op_manager.try_auto_fetch_contract(&key,
target_addr)` against the forward target and synthesise a retriable
`OperationError`. The next client retry sees the primed store and
succeeds. Real WASM-exec rejections keep their existing ERROR-level
log + raw `OpError`.

Companion PR opens shortly.

## Out of scope

The phantom `phase=get_timeout` log + duplicate-error race on the
GET side (#4069 / PR #4070) is the other half of #4066 and is
tracked separately.

A wire-format change to let `UpdateMsg::RequestUpdate.value` carry
`UpdateData` (so the relay can merge) is the cleaner architectural
fix, but it needs stdlib coordination and a `min-compatible-version`
bump. Not in scope for this issue.

iduartgomez:
## Problem

The client-initiated UPDATE remote-target branch in
`drive_client_update` unconditionally calls `super::update_contract`
to apply the update locally before forwarding. On a non-hosting
originator the local executor lacks the contract's code and params,
so `Executor::upsert_contract_state` fails with
\"missing contract parameters\" and the UPDATE never reaches the
recipient.

```
ERROR ... update (task-per-tx): failed to apply update locally
      before forwarding
      contract=<key>
      error=put error for contract <key>, reason: missing contract
      parameters phase=\"error\"
```

See #4071 for full failure-mode analysis. Subset of #4066.

## Why we can't just skip local apply

The wire format `UpdateMsg::RequestUpdate.value: WrappedState` carries
a *post-merge full state*, not a delta. The originator MUST produce
one before forwarding, which requires the contract's params to merge
the delta. A raw-delta wire format would be a cleaner fix but it's a
wire-format change needing stdlib coordination and a
`min-compatible-version` bump (deferred).

The relay path at `drive_relay_request_update` (~line 698) already
handles non-hosting correctly: it checks `state_before.is_some()`
first and forwards without local apply when the contract isn't
hosted. The originator can't do that for the wire-format reason
above.

## Solution

Self-heal. When `update_contract` fails with an error where
`is_contract_exec_rejection() == false` (i.e. not a real WASM merge
rejection — the only production case is \"missing contract
parameters\"), trigger
`op_manager.try_auto_fetch_contract(&key, target_addr)` against the
chosen forward target (which the proximity cache / ring routing
picked, so it likely hosts the contract) and synthesise a retriable
`OperationError`. The next client retry sees the primed store and
succeeds.

`try_auto_fetch_contract` is rate-limited to one attempt per contract
per 5 minutes via `OpManager::pending_contract_fetches`, so multiple
writes queueing while the fetch is in flight don't hammer the target
peer. The mechanism was already used by the relay paths
(`BroadcastTo`, `RequestUpdate`); extending it to the
client-initiated branch matches existing self-heal semantics.

Real WASM-exec rejections (out-of-gas, traps, invalid update) keep
their existing ERROR-level log + raw `OpError` propagation.

## Test plan

- [x] `cargo test -p freenet --lib drive_client_update_remote_target_auto_fetches_on_missing_params` (regression: passes after fix, fails before)
- [x] `cargo test -p freenet --lib operations::update::` (53 passed)
- [x] `cargo clippy -p freenet --lib --tests -- -D warnings`
- [x] `cargo fmt`

## Drive-by

First commit fixes two pre-existing clippy warnings on `main` (same
as PR #4070 — needed to make pre-commit hooks pass).

## Fixes

Closes #4071. Partial fix for #4066 (the GET phantom timeout / leak
is tracked in #4069 / PR #4070).

### DIFF
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -260,6 +260,36 @@ impl ExecutorError {
         }
     }
 
+    /// Narrow discriminator for the specific failure that the
+    /// originator-side UPDATE auto-fetch heals: contract code/params
+    /// are not present in the local `state_store`, so
+    /// `update_contract` cannot run the merge.
+    ///
+    /// Distinct from `is_contract_exec_rejection` (which negates a
+    /// broader set including other contract-side validation errors
+    /// like `Deser`/`InvalidState`/`InvalidDelta`/`Other`/`DoublePut`/
+    /// `InvalidArrayLength` and storage errors). Auto-fetching on
+    /// those broader failures is wasted work — the contract IS
+    /// present, the input is bad. Use this narrow predicate at
+    /// originator UPDATE call sites so a malformed delta or a disk
+    /// error never triggers auto-fetch storms.
+    ///
+    /// Discriminator: stdlib's `ContractError::Update` with a cause
+    /// containing the literal "missing contract parameters" string
+    /// (produced by `runtime.rs`'s `get_params` failure path,
+    /// approx. lines 920–953). Any other `Update` variant cause
+    /// returns false.
+    pub fn is_missing_contract_parameters(&self) -> bool {
+        match &self.inner {
+            Either::Left(req_err) => matches!(
+                req_err.as_ref(),
+                RequestError::ContractError(StdContractError::Update { cause, .. })
+                    if cause.contains("missing contract parameters")
+            ),
+            Either::Right(_) => false,
+        }
+    }
+
     /// Returns true ONLY when the contract WASM merge function ran to completion
     /// and returned a typed `InvalidUpdate` / `InvalidUpdateWithInfo` rejection
     /// (e.g., "New state version N must be higher than current version N"). This
--- a/crates/core/src/node/network_bridge/priority_select/tests.rs
+++ b/crates/core/src/node/network_bridge/priority_select/tests.rs
@@ -2629,7 +2629,7 @@ async fn test_handshake_yields_during_p7_buffer_drain() {
                     self.counter.fetch_add(1, Ordering::SeqCst);
                     Poll::Ready(Some(item))
                 }
-                other => other,
+                other @ Poll::Ready(None) | other @ Poll::Pending => other,
             }
         }
     }
--- a/crates/core/src/operations.rs
+++ b/crates/core/src/operations.rs
@@ -472,6 +472,14 @@ impl OpError {
         matches!(self, Self::ExecutorError(e) if e.is_contract_exec_rejection())
     }
 
+    /// Narrow predicate for the originator-side UPDATE auto-fetch
+    /// trigger: returns true only for "missing contract parameters"
+    /// from `runtime.rs::get_params`. See
+    /// `ExecutorError::is_missing_contract_parameters` for rationale.
+    pub fn is_missing_contract_parameters(&self) -> bool {
+        matches!(self, Self::ExecutorError(e) if e.is_missing_contract_parameters())
+    }
+
     /// Returns true ONLY when the contract WASM merge function rejected the
     /// update with a typed `InvalidUpdate{,WithInfo}` error (the benign
     /// stale-state case from issue #3914). Use this for log-severity
--- a/crates/core/src/operations/put/op_ctx_task.rs
+++ b/crates/core/src/operations/put/op_ctx_task.rs
@@ -2124,6 +2124,7 @@ mod tests {
     /// - upgrade-to-streaming: `ctx.send_to_and_register_waiter(...) +
     ///   conn_manager.send_stream(...)` — pinned by
     ///   `drive_relay_put_upgrades_when_payload_exceeds_threshold`
+    ///
     /// Both await downstream Response via `pending_op_results` (waiter
     /// receiver in the streaming branch, callback in the non-streaming
     /// branch). Neither uses `send_fire_and_forget`.
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -235,12 +235,47 @@ async fn drive_client_update(
                 return Err(OpError::RingError(RingError::NoHostingPeers(*key.id())));
             }
 
+            // No remote target → no peer to auto-fetch from. If
+            // `is_hosting_contract` reports true but `state_store`
+            // doesn't have params, that's an internal inconsistency
+            // (the ring's hosting registry is the announce/subscribe
+            // layer, not the executor's StateStore — they CAN
+            // diverge if a hosting announcement races with state
+            // eviction). Surface the missing-params case explicitly
+            // so operators can correlate the inconsistency, instead
+            // of letting it bubble as an opaque OpError.
             let UpdateExecution {
                 value: _,
                 summary,
                 changed,
                 ..
-            } = super::update_contract(op_manager, key, update_data, related_contracts).await?;
+            } = match super::update_contract(op_manager, key, update_data, related_contracts).await
+            {
+                Ok(execution) => execution,
+                Err(err) if err.is_missing_contract_parameters() => {
+                    tracing::error!(
+                        tx = %client_tx,
+                        contract = %key,
+                        error = %err,
+                        phase = "ring_state_store_inconsistency",
+                        "update (task-per-tx): is_hosting_contract reports true \
+                         but state_store has no params for this contract; cannot \
+                         auto-fetch (no remote target) — the ring/state-store \
+                         divergence must be repaired by a re-PUT or a manual \
+                         seeding step (#4066)"
+                    );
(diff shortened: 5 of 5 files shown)


iduartgomez:
## Skeptical Review

### High Severity

**1. The `None =>` (no-remote-target) branch has the SAME gap and is NOT fixed by this PR.**
At `update/op_ctx_task.rs:243`, `update_contract(...).await?` is called unconditionally with the same propagation pattern. The branch is gated by `is_hosting_contract(&key)` (line 228), which checks the `ring` (subscriber/hosting registry), but `is_hosting` does not guarantee `state_store.get_params` will succeed — the `Ring`'s view of "hosting" is the announce/subscribe layer, not the executor's `StateStore`. If a node announces hosting before its `state_store` row is written (or vice versa: a contract-handler row exists but params eviction occurred), the `None =>` branch returns `RingError::NoHostingPeers` or hits the same "missing contract parameters" path. **The PR's commit message claims "the relay path already handles this" but says nothing about the local `None =>` branch.** Worth at minimum a regression test pinning that branch's behavior for the same input shape.

**2. Auto-fetch race / write-loss when client retries before fetch lands.**
`try_auto_fetch_contract` is fire-and-forget (`GlobalExecutor::spawn`), rate-limited 5 minutes. The synthesised `OperationError` returns immediately to the client. If the client retries:
- Retry within 5 min: `try_auto_fetch_contract` is rate-limited (early-returns at line 307); the local `state_store` may still not be primed because the spawned GET is still in flight. **Every retry within the cooldown returns the same "auto-fetch triggered, client should retry" error, looking like a hung op to the client even though work is in progress.** The error string instructs the client to retry but provides no signal for *when* — the client is likely to either (a) retry too fast (returning early-fail loops until the GET completes, perhaps minutes) or (b) give up after a short timeout.
- After fetch completes: the GET stores the contract via `auto_fetched_response: true` (`get.rs:378`). Need to verify that path actually writes contract code+params into `state_store` (not just contract-handler/cache), or the next UPDATE retry will hit "missing contract parameters" *again* and another 5-minute cooldown begins. **The PR doesn't include a test that the auto-fetched GET actually primes `state_store.get_params` for `update_contract`.**

**3. Wire-format band-aid; not addressed structurally.**
The PR body acknowledges the wire format is wrong (sender should send delta, not post-merge full state) but defers as "stdlib coordination" work. This means: any non-hosting originator's first UPDATE *always* fails. The "self-heal" gives second-attempt success but only if the client retries — clients that issue one-shot UPDATEs will see the contract fail forever from non-hosting nodes. Particularly bad for delegated UPDATE flows or scripted clients. This is a known limitation and should be called out in a follow-up issue (currently only #4071 closes).

### Medium Severity

**4. `is_contract_exec_rejection` is too narrow as the discriminator.**
The discriminator only matches `RequestError::ContractError(StdContractError::Update { cause: starts_with("execution error") })` (executor.rs:252-261). Returns `false` for:
- `RequestError::ContractError(StdContractError::Update { cause: "missing contract parameters" })` — the intended target. Good.
- `RequestError::ContractError(StdContractError::Deser/InvalidState/InvalidDelta/Other/DoublePut/InvalidArrayLength)` — **these are all real validation/decoding errors but get treated as "contract missing locally" and trigger a wasted auto-fetch GET.** A peer sending malformed input could repeatedly trigger auto-fetch storms (rate-limited per contract, but still wasted GETs).
- Storage errors (`Either::Right(_)` → returns `false`) — disk failures trigger auto-fetch even though the issue is local infrastructure.
- The OpError-level wrappers `OpError::ContractError`, `OpError::IO`, etc. that wrap something other than `ExecutorError` — the predicate at `operations.rs:471` only matches `Self::ExecutorError(e)`, so all other OpError variants return `false` and trigger auto-fetch.

The PR description claims "the only example in production is missing-parameters" — that's true today, but the discriminator allows much broader auto-fetch triggering, including on invalid client inputs. Consider inverting: gate on `is_missing_contract_parameters()` (a new narrow predicate) instead of `!is_contract_exec_rejection()`.

**5. `target_addr` may not host the contract.**
`drive_client_update`'s `Some(target)` arm picks the target via proximity routing. The PR comment says "the proximity cache / ring routing chose, so it likely hosts the contract" — but routing chooses the *closest* peer to the contract's key, not necessarily a hoster. On a sparse network the closest peer may itself be non-hosting and forward downstream. A GET to a non-hosting target round-trips through that peer's GET handler, which itself may need to fetch — wasted hop, and if the chosen target is the only one near the key we now race two operations (the auto-fetch GET and the eventual second UPDATE) toward the same neighborhood.

**6. `OperationError` may be terminal at the client.**
`client_events.rs:346` maps `OpError::StatePushed` → `ErrorKind::OperationError`, and the synthesised error here also uses `OperationError`. The string "client should retry" is not a structured signal — a client lib doing `match err.kind()` will just see `OperationError { cause: "..." }` and surface a generic failure. There's no distinguished retry-with-backoff variant in `ErrorKind` (only `PeerNotJoined` / `EmptyRing` are tagged for client retry semantics in `client_events.rs:441-448`). Consider adding a `ContractFetchInProgress` variant or reusing `EmptyRing`-style structured retry signaling. As written, most clients will treat this as a hard failure.

### Low Severity

**7. The drive-by clippy fix in commit 1 is fine** but bundling unrelated fixes makes git-bisect harder. The PR description acknowledges this is a recurring pattern (PR #4070 also did the same).

**8. The pin test is source-scrape only.** It checks for substring presence in the function body but doesn't actually run the code path. A future refactor that splits the auto-fetch into a helper called from the arm would still pass the pin (good) but a refactor that moves the call into a parent function but leaves the substrings in a comment block would also pass (false-positive).

### Questions for Author

1. Why isn't the `None =>` (local) branch fixed in the same PR? Same failure mode applies if `is_hosting` is true but `get_params` fails.
2. Does the auto-fetched GET (`auto_fetched_response: true` path) actually write to `state_store` such that the next `update_contract` retry succeeds? Is there an integration test?
3. The 5-min cooldown means clients see ~5 minutes of "auto-fetch triggered, client should retry" errors after each first failure. What's the expected client retry strategy here? Should the error encode a retry-after hint?
4. Are there other `update_contract` call sites on the originator? PUT sub-op spawning UPDATE? Renewal-driven UPDATE?
5. Why use `target_addr` (the forward target) as the auto-fetch source instead of any nearby hosting peer? The relay paths use `sender_addr` (the sender of the failing message), which has stronger guarantees of having the contract.

### What Verified

- `is_contract_exec_rejection` correctly returns `false` for "missing contract parameters" (executor.rs:252-261, predicate matches only "execution error" prefix).
- The relay paths (`drive_relay_request_update`, `drive_relay_broadcast_to_streaming`) already use `try_auto_fetch_contract` against `sender_addr` with the same `!is_contract_exec_rejection()` gate (op_ctx_task.rs:1064-1068, update.rs:606), so the pattern is consistent.
- `try_auto_fetch_contract` rate-limit uses `entry.or_insert + saturating_sub` correctly; no TOCTOU.
- `target_addr` socket address unwrapped at line 272-282 and bails out cleanly if absent, before reaching the auto-fetch path.
- Pin test correctly scoped to the `Some(target) =>` arm via substring slicing.

_Posted by skeptical-reviewer agent._

PR #3012, opened 2026-02-14
fix: strip version prefix when loading contracts from disk (#2924)

iduartgomez:
## Problem

After upgrading to v0.1.116, contract publishing failed with:
"Error when converting wat: input bytes aren't valid utf-8"

The node was trying to parse binary WASM as WAT (text format).

## Root Cause

Contracts are stored on disk with to_bytes_versioned() which adds
a version prefix. The fetch_contract() function was using
ContractContainer::try_from() which read raw file bytes including
the prefix.

When these bytes were passed to Module::new(), wasmer/wasmtime
auto-detection failed because:
- Binary WASM must start with magic bytes: 0x00 0x61 0x73 0x6d
- With the prefix, magic bytes were at offset 32, not offset 0
- Module::new() tried to parse as WAT text format
- Failed with "input bytes aren't valid utf-8"

## Solution

Replace ContractContainer::try_from() with
ContractCode::load_versioned_from_path() which properly strips
the version prefix before returning the WASM bytes.

This matches the existing pattern used in store_contract() (line 120).

## Testing

Added test_versioned_contract_loading_issue_2924 which:
1. Stores a contract with version prefix
2. Fetches it from disk (cache cleared)
3. Verifies WASM magic bytes are at offset 0
4. Verifies successful compilation without WAT parsing error

All existing contract_store tests pass.

https://claude.ai/code/session_012AJBRkaCkCwzN9PASiF1vW

Fixes #2924

Issue #2924, opened 2026-02-09
WASM binary as WAT


Niche-Apps:
Description

After auto-updating to Freenet v0.1.116 (built 2026-02-09T04:36:52Z), publishing any contract via fdev publish or fdev execute put fails with a WASM compilation error. The same contract code published successfully on prior versions.
Error

Error: Failed to receive response: client error: unhandled error: execution error: compile: 
WebAssembly translation error: Error when converting wat: input bytes aren't valid utf-8
Copy
The node appears to be trying to parse valid WASM binary as WAT (text format).
Environment

Freenet: 0.1.116 (f9416958c5bb)
fdev: 0.3.90
OS: Ubuntu 20.04 (aarch64/ARM)
Rust: cargo 1.93.0
Steps to Reproduce

Build a contract with fdev build (succeeds, produces valid WASM)
Publish: fdev publish --code build/freenet/contract_name contract --state initial_state.bin --webapp-archive webapp.tar.xz
Node rejects with the WAT utf-8 error
The packaged contract file contains a 32-byte hash prefix followed by valid WASM binary (\0asm magic at offset 0x22).

PR #4075, opened 2026-05-10
fix(fdev): detect packed contract format before treating as raw WASM

Basedfloppa:
## Problem

`fdev publish --code <file>` and `fdev get-contract-id --code <file>`
both call `ContractCode::load_raw()` first and only fall back to the
versioned `ContractContainer::try_from((path, params))` when
`load_raw` returns `Err`. But `load_raw` is just `read_to_end`: it
succeeds for any readable file, including the packaged contract
format `fdev build` emits (8-byte version + 32-byte hash header +
WASM). The fallback branch — which calls `load_versioned_from_path`
and strips the 40-byte prefix — is therefore unreachable for
packaged inputs.

User-visible effect: the metadata bytes get folded into
`ContractCode.data` and travel through the `WrappedContract` to the
receiving node, where wasmtime's name-section parser fails UTF-8
validation on the prefix and rejects the module with
`compile: input bytes aren't valid utf-8`. The contract never lands
in the local DB and any subsequent `Subscribe` from a client fails
with `contract WASM/parameters not cached locally`.

Reproducer:
1. `cd <any-contract>; fdev build`
2. `fdev publish --code build/freenet/<name> --parameters <p> contract --state <s>`
3. Server logs: `compile: input bytes aren't valid utf-8`.

This is the same shape of bug as #2924 — that issue's fix (PR #3012)
covered the disk-fetch side (`fetch_contract` now strips the prefix
via `load_versioned_from_path`); this PR is the symmetric fix on the
publish-input side. Until this lands, the only way to use `fdev
publish` against a packaged contract is to extract the raw WASM from
offset 40 of the packaged file and pass that to `--code`.

## Solution

Extract a single `load_contract_for_publish(path, params)` helper
used by both call sites. It gates `ContractCode::load_raw` on the
WASM magic bytes (`\0asm`); when the magic is absent at offset 0,
the file is treated as packaged and routed through
`ContractContainer::try_from((path, params))`, which reads the
8-byte version + 32-byte hash header and returns a `ContractCode`
that contains only the stripped WASM bytes.

Why this approach (vs. swapping the order to try the versioned
parser first):

1. The magic check is one byte comparison and unambiguous: WASM
   files begin with `\0asm` per the WebAssembly spec, packaged
   files begin with a u64 BE version (currently 0) which can never
   start with `\0asm`. No risk of misclassifying either form.
2. The error mode for legitimate-but-truncated packaged files is
   preserved: bytes that fail the magic check route through
   `load_versioned_from_path`, whose existing error messages
   already cover truncated headers and unsupported API versions.
3. Both `put_contract` and `get_contract_id` had byte-identical
   wrong precedence; consolidating into one helper means a future
   third caller can't reintroduce the bug.

## Testing

Regression coverage in `commands::tests`:

- `raw_wasm_loads_into_v1_container` — raw WASM still wraps
  unchanged (happy path, ensures the fix doesn't regress the
  intended raw-input case).
- `packaged_contract_strips_version_prefix` — packaged input has
  its 40-byte header stripped. **This exact assertion fails on the
  pre-fix code** — verified by temporarily reverting the magic
  guard and re-running the test, which then reports
  `left: [0,0,...,asm,1,0,0,0]` vs `right: [asm,1,0,0,0]`.
- `garbage_file_surfaces_error` — bytes that are neither WASM nor
  parseable as versioned must error, rather than silently producing
  a malformed contract that would only fail on the receiving node.
- `empty_file_surfaces_error` — empty file boundary. Without the
  magic guard, `load_raw` would Ok-with-empty-data and the helper
  would happily wrap a 0-byte `ContractCode`; the magic check
  forces the empty case through the versioned parser, which errors
  on the missing version u64.

Local validation:
- `cargo fmt --check -p fdev` — clean
- `cargo clippy -p fdev --all-targets -- -D warnings` — clean
- `cargo test -p fdev` — 27 + 9 + 1 tests pass (1 ignored,
  pre-existing)

## Fixes

Related to #2924 (fix on the disk-fetch path was PR #3012; this PR
fixes the symmetric publish-input path).


iduartgomez:
@claude do we exercise this commands anywhere in our test suite?

iduartgomez:
## Suggested integration test location for `fdev publish` / `get-contract-id` round-trip

**Gap:** unit tests cover `load_contract_for_publish` format detection, but no test drives the full CLI → WebSocket → node-DB path with a packaged contract on the wire. `crates/fdev/tests/websocket_response.rs` mocks the server, so it can't catch a regression in node-side insertion.

**Recommended location:** new file `crates/core/tests/fdev_publish_e2e.rs`.

**Why there, not `crates/fdev/tests/`:**
- `crates/core/tests/` already runs a local in-process gateway+peer via the `#[freenet_test]` macro (e.g. `operations.rs:164`). No real network — `127.0.0.1` only.
- It runs under the existing **Unit & Integration** CI job (`ci.yml` `test_unit`), which already does `cargo build --locked` (producing `fdev`) + `make -C apps/freenet-ping`. Zero new workflow, zero extra build cost — the `fdev` binary is already on disk.
- Putting it in `crates/fdev/tests/` would require depending on `freenet` (and its test_utils) from `fdev`, creating a build cycle and inflating the fdev-only test job. The core tree already has the node fixture; reuse it.

**Structure (one file, two tests):**

```rust
// crates/core/tests/fdev_publish_e2e.rs
use freenet::test_utils::{self, TestContext, verify_contract_exists};
use freenet_macros::freenet_test;
use std::path::PathBuf;
use std::process::Command;

fn fdev_bin() -> PathBuf {
    // fdev is built by the existing `cargo build --locked` step.
    PathBuf::from(env!("CARGO_TARGET_DIR"))
        .join(if cfg!(debug_assertions) { "debug" } else { "release" })
        .join("fdev")
}

#[freenet_test(
    health_check_readiness = true,
    nodes = ["gateway", "peer-a"],
    timeout_secs = 180,
    startup_wait_secs = 30,
    tokio_flavor = "multi_thread",
    tokio_worker_threads = 2,
)]
async fn fdev_publish_raw_wasm_round_trip(ctx: &mut TestContext) -> TestResult {
    let peer = ctx.node("peer-a")?;
    let contract_path = test_utils::fixture_path("test-contract-integration.wasm");

    let status = Command::new(fdev_bin())
        .args([
            "publish", "contract",
            "--code", contract_path.to_str().unwrap(),
            "--node-url", &format!("ws://127.0.0.1:{}", peer.ws_port),
        ])
        .status()?;
    assert!(status.success(), "fdev publish failed");

    let key = /* derive ContractKey from contract bytes */;
    verify_contract_exists(&peer.temp_dir_path, key).await?;
    Ok(())
}

#[freenet_test(/* same attrs */)]
async fn fdev_publish_packaged_contract_round_trip(ctx: &mut TestContext) -> TestResult {
    // Regression case for this PR: packaged (versioned-prefix) contract
    // must traverse load_contract_for_publish + WS publish + node insert.
    // Fixture: crates/core/tests/data/packaged-contract-v1.wasm (pre-built, committed).
    ...
}
```

**Fixture:** drop a pre-packaged `crates/core/tests/data/packaged-contract-v1.wasm` rather than shelling out to `fdev package` inside the test. Format-detection variants are already covered by the unit tests in this PR — the integration test just needs one of each to cross the wire.

**CI impact:**
- No new workflow, no new runner.
- Adds ~one `#[freenet_test]` node bring-up (~30–60s) to the existing `test_unit` nextest run; nextest parallelism absorbs it.
- `fdev` binary reused from the existing build step — no recompile.

**Optional follow-up:** add a third case driving `fdev get-contract-id` against the same live node. Currently `put_contract` is the only async path with no e2e coverage; `get_contract_id` has the same gap and is cheap once the harness is in place.

Issue #4077, opened 2026-05-10
Inbox UPDATE merge stuck: fetch_related_via_network times out, then merge_rejected_valid_local pins stale state


iduartgomez:
## Summary

Freenet Mail (`freenet/mail`) v0.1.2/v0.1.3 inbox UPDATEs fail to apply on receiver nodes because `fetch_related_via_network` times out fetching the sender's AFT identity contract (10s budget), and subsequent retries hit `merge_rejected_valid_local` against the receiver's stale (initial-empty) inbox state — pinning the inbox forever.

## Repro

- Two real-network peers, both v0.2.55 (build `3dfb997d`)
- Sender (148.3.141.199, peer `3EVQC9RFMFBy9awuB`) sends mail to receiver (79.150.235.155, peer `N2gTQCAMRYLmGQjK`)
- Sender's webapp constructs `ContractRequest::Update { data: UpdateData::Delta(...) }` for the receiver's inbox key (`4EBLuFr...`)
- Inbox contract's `update_state` calls `requires(...)` for the sender's AFT identity contract (`4jrpVK6iNB...`)
- Receiver's node has never seen `4jrpVK6...` — neither subscribed, nor near in keyspace

## Receiver-side log evidence

```
2026-05-10T09:33:21.639Z BroadcastToStreaming arrived contract=4EBLuFr... total_size=241774 sender=147.235.192.32

2026-05-10T09:37:21.201Z WARN runtime: Failed to fetch related contract for update_state requires()
    contract=4EBLuFr... related_id=4jrpVK6iNB...
    error=missing related contract: 4jrpVK6iNB...

2026-05-10T09:37:31.276Z WARN: BroadcastToStreaming update skipped: contract not ready locally
    error=put error... reason: timed out fetching related contracts
```

10s gap between fetch attempt and timeout = `RELATED_FETCH_TIMEOUT` constant.

After timeout, every retry re-attempts and re-fails:

```
2026-05-10T09:41:41.491Z WARN runtime: Merge rejected incoming state but local state is valid - not replacing
    contract=4EBLuFr...
    error=update error... reason: execution error: invalid contract update
    local_state_size=11977 incoming_state_size=241290
    event=\"merge_rejected_valid_local\"
```

[...]

PR #4080, opened 2026-05-10
fix(executor): parallelize related-contract fetch (#4077)

iduartgomez:
## Summary

The three sites that resolve a contract's \`requires(...)\` related-list ran **serially** under a single 10s wall-clock budget (\`RELATED_FETCH_TIMEOUT\`). For N related ids the per-fetch budget was ~10s/N, tight enough on real networks (where AFT-style keys are far in keyspace from the receiver) that any one slow hop drained the whole pot. Receivers timed out, the contract returned Invalid, the upsert path logged \`merge_rejected_valid_local\`, and the receiver's state stayed pinned at the initial empty value forever.

Fix: switch the three loops to \`futures::future::join_all\` so per-id sub-op GETs race in parallel. The 10s budget now bounds the longest fetch, not the cumulative path.

## Sites changed

- \`runtime.rs:1402\` — UPDATE-side inline fetch when \`update_state\` signals \`requires(...)\`
- \`runtime.rs:2118\` — \`fetch_related_for_validation\` (validate-side)
- \`runtime.rs:3404\` — \`fetch_related_for_validation_network\` (PUT-side variant)

Per-id closures borrow \`&Self\` (reborrowed from \`&mut self\`); outer \`&mut self\` is reclaimed once \`fetch_all\` is awaited, so post-fetch validate/update calls work unchanged.

## Real-network evidence (from #4077)

Mail v0.1.4 receiver-side log on hsantos's node:
\`\`\`
WARN runtime: Failed to fetch related contract for update_state requires()
    contract=4EBLuFr... related_id=4jrpVK6...
    error=missing related contract: 4jrpVK6...
WARN runtime: BroadcastToStreaming update skipped: contract not ready locally
    error=put error... reason: timed out fetching related contracts
\`\`\`

Receiver's inbox stayed at 11977 bytes (empty) while sender's was 241290+. Repeated rebroadcasts kept hitting \`merge_rejected_valid_local\`. Telemetry showed receiver \`update_broadcast_received\` but never \`update_broadcast_applied\`.

## Test plan

- [x] \`cargo test -p freenet --lib\` — 2552 passed, 0 failed (full suite incl. all related-contract pool tests)
- [x] \`cargo fmt --all\` + \`cargo clippy -p freenet -- -D warnings\` clean
- [ ] After merge: rebuild bin, re-test mail cross-peer send (separate validation step on the mail repo)

## Companion fix

mail#202 (already merged + released as v0.1.4) bundles the sender's AFT record inline as \`RelatedStateAndDelta\`, which avoids the fetch entirely on the sender's first-hop direct delta. But rebroadcast paths arrive as bare \`UpdateData::State\` (\`op_ctx_task.rs:1727\`) which carries the whole inbox + every message's AFT record id — that's where N>1 hit the timeout. This PR unblocks the rebroadcast path.

Refs #4077.

### DIFF
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -3927,4 +3927,77 @@ mod sub_op_get_migration_pin_tests {
              retired in #1454 phase 5 final"
         );
     }
+
+    /// Pin: the three sites that resolve a contract's `requires(...)`
+    /// related-list MUST fan out via `join_all`, not iterate serially
+    /// inside a `for` loop. Regression: the previous serial loops shared
+    /// a single 10s wall-clock budget (`RELATED_FETCH_TIMEOUT`), so for
+    /// N>1 ids the per-fetch budget was ~10s/N. On real networks where
+    /// AFT-style related contracts are far in keyspace, this pinned
+    /// receivers' inboxes at empty state forever — see
+    /// `freenet/freenet-core#4077` and `freenet/mail#198 / mail#202`
+    /// for the production trace and the app-side workaround.
+    ///
+    /// We can't unit-test the parallelism timing directly: the
+    /// `TEST_NETWORK_FETCH_OVERRIDE` stub is sync (`Rc<dyn Fn>`), so a
+    /// per-id artificial delay would block the executor thread rather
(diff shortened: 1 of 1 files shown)


iduartgomez:
@claude do a /pr-review of the pr

PR #4083, opened 2026-05-10
fix(transport): sweep idle PeerConnection.streaming_handles entries

### LINKED ISSUE #4079: fix(transport): PeerConnection.streaming_handles leaks fragment buffers when upstream stalls after cancel/timeout
## Symptom

User report code [JTDY6J](nova.locut.us:/var/lib/freenet-reports/JTDY6J.json) — Linux x86_64, v0.2.55 (\`816f4adca31d\`), 96GB RAM, OOM-killed twice in a few days. Hosting 706 contracts (River + Delta + Harvest, multi-day uptime).

In a 30-minute log window the streaming layer shows:

- 108 \`Stream assembly timed out — no fragments received within inactivity window\`
- 81 \`UPDATE relay (task-per-tx streaming): BroadcastToStreaming driver returned error … error=stream was cancelled\`
- 4 \`failed to claim orphan stream from registry (broadcast)\`
- 2,466 orphan-registry expirations, mean 124 / peak 230 outstanding

Telemetry from this peer (\`4SoHrKXyGLGnL7FEG@68.5.12.48\`) over a ~7-minute window shows ~104k \`update_broadcast_received\`, ~14k \`transfer_started\`, 196 timeouts. 91% of received broadcasts are duplicates filtered post-assembly, so every one allocates fragment storage before being discarded.

## Root cause

\`PeerConnection::streaming_handles: HashMap<StreamId, StreamHandle>\` (\`crates/core/src/transport/peer_connection.rs:262\`) has no time-based GC. All four eviction sites are reactive to inbound traffic on the same stream:

1. \`peer_connection.rs:958\` — when \`inbound_streams\` returns Err on the legacy assembly path.
2. \`peer_connection.rs:1361\` — only when **a subsequent fragment arrives** and \`push_fragment\` returns \`StreamError::Cancelled\`.
3. \`peer_connection.rs:1540\` — when the legacy \`InboundStream\` finishes assembling.
4. \`peer_connection.rs:288-313\` (\`Drop\`) — only when the entire \`PeerConnection\` is destroyed.

When the streaming UPDATE relay driver hits a timeout / cancel / \`OrphanStreamError::AlreadyClaimed\` (\`operations/update/op_ctx_task.rs:1480-1499\` and \`:1619-1651\`), the driver-side \`StreamHandle\` clone drops at end-of-scope. The orphan registry's clone is dropped on the next 5s GC tick (\`operations/orphan_streams.rs:239-258\`), which calls \`handle.cancel()\` and removes its registry entry. **But \`gc_expired\` does not notify the owning \`PeerConnection\`** to evict the matching entry from \`streaming_handles\`.

The remaining \`streaming_handles[stream_id]\` clone keeps the underlying \`Arc<LockFreeStreamBuffer>\` alive. The buffer pre-allocates \`Box<[AtomicPtr<Bytes>]>\` sized for the full stream (\`streaming_buffer.rs\`), so each stalled handle pins at minimum \`total_fragments × 8 bytes\` of slot memory plus whatever \`Bytes\` were inserted before stall.

The handle is only freed if (a) the upstream sender resumes and emits another fragment that triggers the \`Cancelled\` path at line 1361, or (b) the entire \`PeerConnection\` drops. For a long-lived River user with constant traffic, (b) doesn't happen for hours; (a) doesn't happen after the upstream itself hit its own outbound timeout and gave up (the common case for the timeouts above).

Worst case per peer: tens of KB of buffer memory × hundreds of stalled handles × multiple peers = multi-GB working set, matching the user's symptom.

## Hypothesis confirmation step

Add two gauges before fixing:

- \`PeerConnection::streaming_handles.len()\` — published per peer.
- \`OrphanStreamRegistry::orphan_count()\` — already exists under \`#[cfg(test)]\`; promote to production gauge.

If \`streaming_handles.len()\` grows monotonically across hours while \`orphan_count()\` stays bounded by the 5s GC, the diagnosis is confirmed. Cheap, ships in the next release, and pinpoints the affected peers in production telemetry.

## Fix shape

Two viable approaches:

1. **Notify on orphan eviction.** \`OrphanStreamRegistry::gc_expired\` already has \`peer_addr\` and \`stream_id\` in the entry key. Route an eviction signal back to the owning \`PeerConnection\` (e.g., via the per-peer event channel) so it can \`streaming_handles.remove(&stream_id)\` and \`streaming_registry.remove(stream_id)\`. Keeps the orphan registry as the single source of truth for cancel-and-cleanup.

2. **TTL sweep on \`streaming_handles\` itself.** Add a per-connection periodic sweep (e.g., piggyback on the keep-alive task) that drops handles older than \`STREAM_INACTIVITY_TIMEOUT × N\` and have not received a fragment in that window. Mirrors the orphan registry's policy on the per-connection map.

Approach (1) is less invasive but adds a cross-component channel. Approach (2) localizes the fix to \`peer_connection.rs\` and removes the \"reactive eviction depends on inbound traffic\" coupling entirely.

Either way, the regression test should:

1. Spawn a \`PeerConnection\`, register N orphan streams with no first fragment ever arriving.
2. Drive the \`OrphanStreamRegistry\` GC past \`ORPHAN_STREAM_TIMEOUT\`.
3. Assert \`streaming_handles.len() == 0\`.
4. Assert \`Arc::strong_count(&buffer) == 1\` (only the test holds it) for each registered buffer.

## Related

- Issue #3458 (multi-consumer fragment reclamation, 2× peak) compounds this — but the dominant culprit per the audit is the missing eviction from \`streaming_handles\`, not buffer reclaim semantics.
- v0.2.56 retires legacy ops.get/put/update DashMaps and includes #4070 (ops.get cleanup leak); none of those touch this code path, so this issue persists into v0.2.56.

[AI-assisted - Claude]

sanity:
## Problem

`PeerConnection::streaming_handles` had no time-based GC. All four
eviction sites were reactive — they only fired when (a) a subsequent
fragment arrived and tripped `StreamError::Cancelled`, (b) the legacy
`InboundStream` finished, or (c) the entire `PeerConnection` dropped.

Two production-realistic paths therefore leak the entry indefinitely:

1. **Orphan-registry GC cancelled the handle.** The orphan registry's
   5 s GC tick calls `handle.cancel()` and drops the registry's clone,
   but does not notify the owning `PeerConnection` to evict the
   matching `streaming_handles` row.
2. **Driver claimed-then-timed-out.** The streaming UPDATE relay
   driver claims the orphan, hits its internal inactivity timeout (5 s)
   or `OrphanStreamError::AlreadyClaimed`, and drops its clone — no
   fragment will ever arrive again to trip the reactive `Cancelled`
   path.

In both cases the underlying `Arc<LockFreeStreamBuffer>` — a
pre-allocated `Box<[AtomicPtr<Bytes>]>` slot array sized for the full
stream, plus any `Bytes` already inserted — remains pinned for the
life of the connection (hours for a long-lived River user).

Production report `JTDY6J` (Linux x86_64, v0.2.55, 96 GB RAM):
108 inactivity timeouts, 81 cancelled BroadcastToStreaming drivers,
mean 124 / peak 230 outstanding orphans, OOM-killed twice in a few
days. 91 % of received broadcasts are duplicates filtered post-
assembly, so every one allocates fragment storage before being
discarded.

## Approach

Approach 2 from the issue (TTL sweep on `streaming_handles` itself).
Localizes the fix to `peer_connection.rs` and removes the
"reactive eviction depends on inbound traffic" coupling entirely,
rather than adding a cross-component notification channel from the
orphan registry's GC.

- `streaming_handles: HashMap<StreamId, (StreamHandle, u64)>` —
  `u64` is `last_activity_nanos` in the `time_source` epoch.
- The timestamp is set on insertion and refreshed on every successful
  `push_fragment` so an actively-receiving stream is never reaped.
- `STREAMING_HANDLE_IDLE_TIMEOUT = 30 s` = 6 ×
  `streaming::STREAM_INACTIVITY_TIMEOUT`. Comfortably past the 5 s
  point at which any in-flight `assemble()` would have already failed
  with `InactivityTimeout`; well under the 120 s connection idle
  timeout and the 60 s orphan registry timeout. Production p95 for a
  successful large transfer is 1.8 s.
- Sweep called from the existing 5 s `timeout_check.tick()` branch in
  `recv()` — no new task, no new timer. Each swept entry has its
  handle cancelled (waking any blocked consumer with `Cancelled`)
  and its companion `streaming_registry` row removed, releasing the
  buffer `Arc`.
- Sweep logic factored into a free function `sweep_streaming_handles_inner`
  so it can be unit-tested against a plain `HashMap` + `StreamRegistry`
  without constructing a full `PeerConnection`.

The fix preserves the existing `Drop` impl behavior (cancel all
streaming handles on connection drop) and the reactive cleanup paths
(line 958, 1361, 1540) — the sweep is purely additive backstop.

## Testing

Two new unit tests in `mod tests`:

- `sweep_streaming_handles_inner_drops_idle_entries` — populates
  the map with stale + fresh entries, runs sweep, asserts stale row
  is gone from both the map and the registry, fresh row is untouched,
  and the swept handle is cancelled (so any late fragment short-
  circuits on `StreamError::Cancelled` instead of allocating). This
  is the regression test for #4079.
- `sweep_streaming_handles_inner_keeps_recent_entries` — pins the
  lower bound of the threshold so a future reduction of
  `STREAMING_HANDLE_IDLE_TIMEOUT` cannot accidentally reap an
  actively-receiving stream.

Full lib test suite: `cargo test -p freenet --lib` → 2558 passed,
0 failed. `cargo clippy -p freenet --all-targets -- -D warnings`
clean. `cargo fmt` clean.

## Fixes

Closes #4079

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/transport/peer_connection.rs
+++ b/crates/core/src/transport/peer_connection.rs
@@ -259,7 +259,17 @@ pub struct PeerConnection<S = super::UdpSocket, T: TimeSource = RealTime> {
     streaming_registry: Arc<streaming::StreamRegistry>,
     /// Streaming handles for active streams (parallels inbound_streams).
     /// Used for pushing fragments to streaming consumers.
-    streaming_handles: HashMap<StreamId, streaming::StreamHandle>,
+    ///
+    /// Each entry pairs a `StreamHandle` with `last_activity_nanos` (in the
+    /// `time_source` epoch) — the time of the most recent successful interaction
+    /// with this stream (either initial registration or a subsequent fragment
+    /// arrival). The periodic sweep on `timeout_check` ticks drops entries idle
+    /// for longer than `STREAMING_HANDLE_IDLE_TIMEOUT`; without that bound, a
+    /// stalled upstream sender (after the orphan registry GC has already
+    /// cancelled the handle, or after the driver claimed-then-timed-out) leaks
+    /// the underlying `Arc<LockFreeStreamBuffer>` until the entire connection
+    /// drops. See issue #4079.
+    streaming_handles: HashMap<StreamId, (streaming::StreamHandle, u64)>,
     /// Time source for deterministic simulation support
     time_source: T,
     /// Optional orphan stream registry for handling race conditions between
@@ -297,7 +307,7 @@ impl<S, T: TimeSource> Drop for PeerConnection<S, T> {
                 stream_count,
                 "Cancelling streaming handles on connection drop"
             );
-            for handle in self.streaming_handles.values() {
+            for (handle, _) in self.streaming_handles.values() {
                 handle.cancel();
             }
         }
@@ -351,6 +361,70 @@ const DEDUP_CACHE_CAPACITY: NonZeroUsize = match NonZeroUsize::new(1000) {
     Some(v) => v,
     None => panic!("DEDUP_CACHE_CAPACITY must be non-zero"),
 };
+
+/// Idle threshold for sweeping `streaming_handles` entries.
+///
+/// A streaming handle that has not received a fragment in this long is treated
+/// as a stalled stream: cancel the handle, drop the entry from the map, and
+/// drop the matching `streaming_registry` entry — releasing the underlying
+/// `Arc<LockFreeStreamBuffer>` (a per-stream pre-allocated slot array plus any
+/// inserted `Bytes`).
+///
+/// Sized at 6 × `streaming::STREAM_INACTIVITY_TIMEOUT` (30s) — comfortably past
+/// the 5s point at which any in-flight `assemble()` would have already failed
+/// with `InactivityTimeout`, but well under the 120s connection idle timeout
+/// and the 60s orphan registry timeout. Production data (2026-04) puts p95
+/// successful transfer at 1.8s, so a 30s gap is conclusive evidence the stream
+/// is dead.
+///
+/// See issue #4079 for the leak scenarios (orphan-registry GC cancelled the
+/// handle but didn't notify the owning `PeerConnection`; or the operations
+/// driver claimed the orphan, timed out internally, and dropped its clone
+/// while the upstream sender stalled — neither path frees the handle the
+/// `PeerConnection` still holds).
+const STREAMING_HANDLE_IDLE_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Drop entries from `streaming_handles` (and their companion
+/// `streaming_registry` rows) that have been idle for longer than
+/// `STREAMING_HANDLE_IDLE_TIMEOUT`.
+///
+/// Factored out of `PeerConnection::sweep_idle_streaming_handles` so the
+/// logic can be unit-tested against a plain map without constructing a
+/// full `PeerConnection`. See issue #4079 for the leak scenarios.
+fn sweep_streaming_handles_inner(
+    streaming_handles: &mut HashMap<StreamId, (streaming::StreamHandle, u64)>,
+    streaming_registry: &streaming::StreamRegistry,
+    now_nanos: u64,
+    peer_addr: SocketAddr,
+) {
+    let threshold_nanos = STREAMING_HANDLE_IDLE_TIMEOUT.as_nanos() as u64;
+    let mut swept_count: usize = 0;
+    streaming_handles.retain(|&stream_id, (handle, last_activity_nanos)| {
+        let idle_nanos = now_nanos.saturating_sub(*last_activity_nanos);
+        if idle_nanos > threshold_nanos {
+            tracing::debug!(
+                %peer_addr,
+                stream_id = %stream_id,
+                idle_secs = Duration::from_nanos(idle_nanos).as_secs_f64(),
+                "Sweeping idle streaming handle (issue #4079)"
+            );
+            handle.cancel();
+            streaming_registry.remove(stream_id);
+            swept_count += 1;
+            false
+        } else {
+            true
+        }
+    });
+    if swept_count > 0 {
+        tracing::info!(
+            %peer_addr,
+            swept_count,
+            remaining = streaming_handles.len(),
+            "Swept idle streaming handles"
+        );
+    }
+}
 
 #[allow(private_bounds)]
 impl<S: super::Socket, T: TimeSource> PeerConnection<S, T> {
@@ -512,7 +586,7 @@ impl<S: super::Socket, T: TimeSource> PeerConnection<S, T> {
             keep_alive_handle: Some(keep_alive_handle),
             pending_pings,
             streaming_registry: Arc::new(streaming::StreamRegistry::new()),
-            streaming_handles: HashMap::new(),
+            streaming_handles: HashMap::new(), // Each entry: (StreamHandle, last_activity_nanos)
             time_source,
             orphan_stream_registry: None,
             dispatched_msg_hashes: lru::LruCache::new(DEDUP_CACHE_CAPACITY),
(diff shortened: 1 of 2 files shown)


sanity:
Pushed review fixes. Summary:

**From Skeptical reviewer (P1):** the 30s constant is wall-clock-style but virtual time can jump past it during WASM exec, which would reap legitimate streams in simulation tests. Fixed by splitting into `STREAMING_HANDLE_IDLE_TIMEOUT_PROD` (30s, derived `6 * STREAM_INACTIVITY_TIMEOUT`) and `_SIM` (86,400s), gated on the same `SimulationIdleTimeout` flag used by `connection_idle_timeout()`. New `streaming_handle_idle_timeout()` returns the active value.

**From Codex (medium):** `last_activity_nanos` was refreshed before checking `push_fragment`'s result, which let a peer keep a dead stream alive indefinitely by replaying the same fragment every <30s. Now refreshed only on `Ok(true)` — actual forward progress — honoring the "Cleanup exemptions MUST be time-bounded" rule.

**From all reviewers (test coverage):** added 5 more tests on top of the original two:
- boundary pin at idle = threshold (kept) vs threshold + 1 ns (dropped),
- idempotent repeat sweep,
- `Arc<LockFreeStreamBuffer>` strong-count assert (real invariant the issue called for — added a `#[cfg(test)] buffer_strong_count()` helper on `StreamHandle`),
- mixed-age 10-entry partition (mirrors production "peak 230 outstanding"),
- source-grep pin so a future refactor that drops the `recv()`-loop sweep call fails CI,
- threshold-switches-on-sim-flag test.

**Doc softening (M2):** clarified that the buffer Arc drops only when the *last* clone drops; orphan-registry clones are intentionally not swept here (its own 60s GC handles those).

**Big-picture reviewer (high):** Filed follow-up issue #4085 for the production telemetry gauges (issue called these out as a "hypothesis confirmation step"), the parallel `inbound_streams` audit (same reactive-only eviction pattern), the transient-retention reduction Codex flagged, and the `.claude/rules/transport.md` doc entry.

Local: `cargo test -p freenet --lib` → 2563 passed. Clippy clean. Fmt clean.

[AI-assisted - Claude]

iduartgomez (approved):

sanity:
## Full multi-model review — PR #4083 (HEAD c61b4b0b)

Reviewed the checked-out branch through four perspectives plus an external Codex pass. CI is green on the current HEAD; `iduartgomez` has approved.

### Summary
TTL sweep on `PeerConnection::streaming_handles` closes the #4079 leak (stalled upstream pins `Arc<LockFreeStreamBuffer>` for the connection's life). Approach 2 from the issue. Implementation is sound: timestamp paired with each handle, refreshed only on `Ok(true)` (real fragment insert), swept on the existing 5 s `timeout_check` tick. No new task/timer, `Drop` and reactive eviction paths preserved.

### Code-first
Matches stated intent. Sweep logic factored into a free function for unit-testability, production wrapper reads the sim-aware threshold. `now_nanos` read once at the top of `process_inbound` then reused for the stale-timestamp write — micro-staleness against a 30 s threshold, irrelevant.

### Testing
Strong coverage: 8 new tests — drop-idle, keep-recent, strict `>` boundary, idempotent repeat, buffer-Arc release (asserts `Arc::strong_count` drop — the actual leak invariant), mixed-age partition, a source-grep pin that the `recv()` loop still calls `sweep_idle_streaming_handles()`, the sim-flag switch, and the `Ok(true)`-gated refresh. The buffer-Arc test and the call-site pin are the genuine regression guards. All pass; full lib suite green.

### Skeptical
- Cancellation safety: sweep runs inside the `timeout_check.tick()` branch handler (not a select future), so no `select!` cancellation footgun.
- Replay-extension: refresh gated on `Ok(true)` only — duplicates (`Ok(false)`) and invalid fragments cannot refresh the deadline. Honors "cleanup exemptions must be time-bounded."
- **Codex P2 — orphan-claim race (non-blocking).** Fragment-first operations streams register the same `StreamHandle` clone in both `streaming_handles` and the orphan registry; streaming relay drivers (`update`/`put` `op_ctx_task.rs`) `claim_or_wait` up to `STREAM_CLAIM_TIMEOUT` (60 s). The 30 s sweep can cancel the shared handle before a delayed driver claims it. Verified the impact: the sweep only fires after 30 s of **zero new fragments**, and `StreamHandle::assemble()` has its own 5 s `STREAM_INACTIVITY_TIMEOUT` — a stream stalled 30 s+ fails regardless. The only observable change is the error variant (`Cancelled` vs `InactivityTimeout`), and both relay drivers handle the two identically (log + `Err(OpError::StreamCancelled)`, fire-and-forget, originator retries). The one true regression is the narrow case of a fully-buffered stream whose driver is scheduled >30 s late — possible within the 60 s CPU-contention envelope `STREAM_CLAIM_TIMEOUT` targets, but low-probability and self-healing. Not blocking; flagging for author awareness. A future tightening could skip the sweep for handles still present in the orphan registry, or align the threshold with `STREAM_CLAIM_TIMEOUT`.

### Big picture
Fixes #4079 as described. No removed tests (the large `git diff origin/main` stat is only because the branch is behind main — the PR-scoped diff is the two transport files). Fast-forward-mergeable, no rebase conflicts. Doc comments are thorough.

### Verdict
**Ready to Merge.** One non-blocking Codex P2 noted above; the equivalent-outcome analysis shows it does not break a valid delayed handoff. CI green on HEAD c61b4b0b, lead-dev approved.

[AI-assisted - Claude]

PR #4084, opened 2026-05-10
fix(router): restore timed RouteOutcome::Success on task-per-tx GET/SUBSCRIBE

sanity:
## Problem

The peer dashboard at `/peer/<addr>` was rendering only the **Failure Probability** chart; **Response Time** and **Transfer Rate** were permanently hidden.

Both estimators are fed exclusively by `RouteOutcome::Success { time_to_response_start, payload_size, payload_transfer_time }` (router.rs:405 + router.rs:443). After the #1454 task-per-tx migrations of GET (Phase 3b / Phase 5 final) and SUBSCRIBE (Phase 2b), **no production code path emits that variant on the originator any more**:

- client-initiated GET emitted `RouteOutcome::SuccessUntimed` (`operations/get/op_ctx_task.rs:411`)
- client-initiated SUBSCRIBE emitted **no** RouteEvent at all on the originator's success branch

So `response_start_time_estimator` and `transfer_rate_estimator` got zero observations forever; `failure_estimator` kept being fed. Hiding empty charts then masked the regression entirely.

This is the same "manually-mirrored telemetry rotted by task-per-tx migration" anti-pattern called out in `bug-prevention-patterns.md` (#4009 / #4010 precedents).

## Solution

1. **GET driver** — `GetRetryDriver` captures `request_sent_at` per attempt in `build_request` and `response_received_at` on Terminal classification. The client Done branch computes `time_to_response_start = received - sent`, sums `payload_size` from the InlineFound state + contract bytes (or from a post-assembly store re-query for streaming), and measures `payload_transfer_time` around `assemble_and_cache_stream`. Emits the timed `RouteOutcome::Success { .. }` when timing is available, falling back to `SuccessUntimed` only for Request-echo / LocalCompletion. Relay GETs continue to emit `SuccessUntimed` (no first-hop timing context).
2. **SUBSCRIBE driver** — capture `request_sent_at = Instant::now()` before `send_to_and_await`. On the `Subscribed` arm, build a `RouteEvent` with `RouteOutcome::Success { time_to_response_start, payload_size: 0, payload_transfer_time: ZERO }` and feed it through `routing_finished` (mirrors legacy `subscribe.rs::outcome()` semantics — SUBSCRIBE has no payload, so `transfer_rate_estimator` is deliberately not fed).
3. **Dashboard** — render all three prediction-component slots unconditionally. Empty curves now show a titled per-metric placeholder via a new `build_estimator_chart_or_placeholder` helper. Keeping the slots visible makes future data-collection regressions detectable on sight instead of silently hiding the affected dimension.

## Testing

- Source-scrape pin tests assert the GET driver captures request/response timing and emits `RouteOutcome::Success { .. }` on the success path, and that SUBSCRIBE's `Subscribed` arm builds a timed `RouteOutcome::Success` and hands it to `routing_finished`. These pin the data-collection contract that the task-per-tx migrations silently broke.
- Unit tests cover the new placeholder behaviour: empty curves render a titled placeholder; non-empty curves render the SVG.
- `cargo test -p freenet --lib` passes (2560 tests).
- `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

## Test plan

- [x] Source-scrape pin tests for the GET and SUBSCRIBE timed-Success contract
- [x] Dashboard placeholder unit tests
- [x] Full `cargo test -p freenet --lib`
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -- -D warnings` clean
- [ ] CI green

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -325,11 +325,14 @@ async fn drive_client_get_inner(
             // the driver is the only place these side effects can
             // happen for Phase 3b.
             // Payload accounting for the router's `transfer_rate_estimator`
-            // and `response_start_time_estimator` (see router.rs:798 +
-            // router.rs:443). Set on the InlineFound/Streaming success
-            // paths; LocalCompletion is a request-echo with no network
-            // transfer so both stay zero (the router will skip the
-            // transfer-rate observation when payload_size==0).
+            // (router.rs:443) and `response_start_time_estimator`
+            // (router.rs:405). Set on the InlineFound/Streaming success
+            // paths; `Terminal::LocalCompletion` is the loopback
+            // Request-echo with no network round-trip — both stay zero
+            // AND we explicitly null the captured per-attempt timing
+            // below (see `timed_outcome`) to avoid poisoning the
+            // response-time model with sub-millisecond loopback latency
+            // samples attributed to `driver.current_target`.
             let mut payload_size: usize = 0;
             let mut transfer_duration: std::time::Duration = std::time::Duration::ZERO;
 
@@ -342,7 +345,13 @@ async fn drive_client_get_inner(
                     // InlineFound delivers state + optional contract in
                     // the same envelope as the Response header, so the
                     // payload transfer time is effectively zero — the
-                    // bytes arrived alongside the timing-of-first-response.
+                    // bytes arrived alongside the response. The router
+                    // deliberately skips `transfer_rate_estimator`
+                    // observations with `transfer_time.is_zero()`
+                    // (router.rs:443) so InlineFound feeds response-time
+                    // only; the legacy `GetStats` shape did the same
+                    // thing for the same reason (no measurable transfer
+                    // phase for an envelope-bundled payload).
                     payload_size =
                         state.size() + contract.as_ref().map(|c| c.data().len()).unwrap_or(0);
                     cache_contract_locally(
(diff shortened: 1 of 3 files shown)


sanity:
Review-round 1 follow-up. Pushed commit 27cff28a addressing the cross-reviewer findings:

**Fixed:**
- **Rule Lint / TimeSource** (skeptical M1, code-first #4, codex #4): all new `Instant::now()` call sites now use `tokio::time::Instant::now()`. Auto-paused under `start_paused(true)` in DST tests.
- **Streaming `payload_size` TOCTOU** (skeptical H2, code-first #3, codex #2): threaded `total_size` through `Terminal::Streaming` from the wire-authoritative `ResponseStreaming` header. Removed the post-assembly store re-query.
- **LocalCompletion timing pollution** (testing #3, code-first #1, skeptical #1, codex #1): explicit `matches!(terminal, Terminal::LocalCompletion)` bypass so the loopback Request-echo path falls through to `SuccessUntimed` and does not poison the response-time model with sub-millisecond loopback samples.
- **Clock anomaly observability** (code-first/skeptical #5): `tracing::warn!` when the `received >= sent` guard rejects a sample, so a future clock-source regression doesn't slip in silently.
- New pin tests for the Exhausted-no-emit contract and the always-three-slots dashboard contract.

**Acknowledged / deferred to follow-up:**
- **SUBSCRIBE failure-path `RouteOutcome::Failure`** (skeptical M3): legitimate gap (legacy `subscribe.rs::outcome()` emitted `ContractOpFailure` on the non-finalized path; the task-per-tx driver emits nothing on timeout/wire_error/NotFound). Out of this PR's scope — the user-visible bug is restored success-path timing. Filed as follow-up.
- **Provider-closure pattern** (skeptical L3, big-picture rec #4): the cleaner #4009-style structural fix (a single audit grep + canonical provider) would also need to cover PUT/UPDATE/CONNECT migrations. Filed as follow-up; this PR mirrors per-driver like #4010 did.
- **`bug-prevention-patterns.md` entry** (big-picture rec #1): adding the audit grep `grep -rn "RouteOutcome::SuccessUntimed" crates/core/src/operations/ | grep -v test` plus the "task-per-tx originator must emit timed `RouteOutcome::Success`" rule belongs with the provider-closure refactor.
- **End-to-end SimNetwork assertion on `router_snapshot.response_time_curve`** (testing #1): worthwhile but adds 50+ LoC of fixture wiring. Filed as follow-up.
- **SUBSCRIBE sub-op gating** (code-first #2): preserved-by-design — legacy `subscribe.rs::outcome()` fed the router on every Subscribed regardless of caller (renewals + sub-ops + executor). The dashboard `op_stats.subscribes` counter is the gated path (per #4010), but the router's `response_start_time_estimator` should learn from every real wire round-trip. The GET sub-op silence is a pre-existing gap (`drive_sub_op_get` emits no RouteEvent) — not introduced here.

CI green now expected. Tests: 2562 lib tests pass; clippy clean; cargo fmt clean.

[AI-assisted - Claude]

Issue #4014, opened 2026-05-03
Manifest-driven iframe sandbox allowlist for hosted webapps


iduartgomez:
## Summary

The gateway shell (`/v1/contract/web/<id>/`) wraps every published webapp in a sandboxed iframe with a hardcoded allowlist. As of the recent widening PRs (#4008 / #4015 / #4067) it currently reads:

```
sandbox="allow-scripts allow-forms allow-popups allow-downloads allow-modals"
allow="clipboard-read; clipboard-write"
```

Every time a webapp needs a new browser capability — `allow-same-origin`, `microphone`, `geolocation`, etc. — we have to land an upstream change in `freenet-core`, cut a release, and bump every consumer. This couples webapp evolution to core release cadence and makes routine UX work into multi-day projects.

## Why the iframe wrapper exists

The sandboxed iframe is the right primitive — it provides:

1. **Code isolation** — webapp can't read gateway shell DOM, can't spoof permission prompts.
2. **Origin isolation** — each webapp gets a unique opaque origin, so `IndexedDB`/`localStorage` stay scoped per app.
3. **Capability gating** — the shell mediates between webapp and the node ws-api, so the webapp can't (e.g.) grant itself an AFT token by clicking its own UI.
4. **Click-jacking defense** — webapp can't navigate the top frame.
5. **Untrusted-by-default posture** — contracts are published by anyone; signed bytes prove provenance, not safety.

The architecture is sound. The problem is the *static* allowlist.

## Design: tiered defaults + manifest opt-in for sensitive caps

Three tiers. Most webapps live entirely in Tier 1 and need no manifest at all. The consent prompt only fires for capabilities that warrant one.

### Tier 1 — Default-on (no manifest, no prompt)

Capabilities that are either (a) browser-gated already (the browser will ask the user anyway, so a second Freenet-level prompt is redundant friction) or (b) low-risk UX where the worst case is "annoying webapp" not "security boundary break".

```
sandbox: allow-scripts allow-forms allow-popups allow-downloads allow-modals
allow:   clipboard-read; clipboard-write; notifications;
         persistent-storage; fullscreen; autoplay; picture-in-picture
```

Per-item rationale:

- `allow-modals` — `alert/confirm/prompt`. Worst case is annoying loop; sandbox boundary unaffected. User can close the tab.
- `clipboard-read/write` — the browser still gates the actual read behind a user gesture. Writes from a click are exactly the intended UX.
[...]

PR #4090, opened 2026-05-12
revert: manifest-driven webapp permissions (#4086)

sanity:
## Problem

#4086 merged without going through the multi-reviewer review process. A post-merge review by four parallel internal reviewers (code-first, testing, skeptical, big-picture) plus Codex independently flagged three significant issues, all in security-sensitive code paths:

### High — CSRF on `POST /v{1,2}/contract/web/{key}/__permissions`

`permission_submit` in `web_permissions.rs:270` accepts a `Form<PermissionForm>` and persists per-contract grants for `microphone`, `camera`, `geolocation`, `usb`, `serial`, `hid`, `bluetooth`, `payment`, `display-capture`, `xr-spatial-tracking`, and `allow-same-origin` — with no Origin / Sec-Fetch-Site / auth-token / nonce check.

Exploit: a webapp loaded in the existing sandboxed iframe (null origin) can issue `fetch('/v1/contract/web/<victim_key>/__permissions', {method:'POST', body:'action=allow&optional=microphone'})`. `application/x-www-form-urlencoded` is a CORS "simple" request — no preflight runs, the body is delivered, and the grant is silently persisted for any contract the attacker knows the key of. `private_network_filter` doesn't help (the iframe is same-host).

The adjacent `permission_prompts.rs:483` already implements Origin checks for the analogous flow, so the pattern was available and just wasn't applied here.

### High — Silent Tier 1 expansion (undisclosed default-on capability change)

The PR description said "Keep Tier 1 capabilities default-on for legacy apps." That implies preservation. In reality `TIER1_ALLOW` in `web_permissions.rs:507-515` grew from the previous default of `clipboard-read; clipboard-write` (2 items) to 7 items, silently adding `notifications`, `persistent-storage`, `fullscreen`, `autoplay`, and `picture-in-picture` to **every existing webapp on every node** without manifest, prompt, or release note.

This is exactly the "server defaults" change pattern called out in `.claude/rules/bug-prevention-patterns.md` (precedent: v0.2.11 streaming-by-default incident). It needs explicit operator-visible disclosure and a security-review note before landing.

### Medium — Insecure `/tmp/freenet` fallback for persistent grants

`server.rs:551` and `client_api.rs:99` fall back to `std::env::temp_dir().join("freenet")` when `webapp_permissions_dir` is `None`. The `From<SocketAddr>` and `Default` impls for `WebsocketApiConfig` both set the field to `None`. The store writes to `/tmp/freenet/webapp-permissions/<key>.toml` with no ownership / mode / symlink checks; `tokio::fs::write` follows symlinks. On a multi-user host, a local attacker can pre-create the dir or symlink files to either tamper with grants or use it as a write primitive.

### Other findings

- **Dead `manifest_version` field**: written by `permission_submit`, never read by `iframe_permissions`. Re-prompt on manifest version change (called out in #4014) is unimplemented despite the storage scaffolding.
- **`allow-same-origin` consent label understates impact** — "Share browser storage origin with other Freenet apps" doesn't convey that it tears down the iframe origin boundary (the protection from GHSA-824h-7x5x-wfmf).
- **`unwrap_or_default()` on malformed grant TOML** silently masks corruption / tampering.
- **No HTTP-level integration test** on a 952-line security-sensitive module — all tests call internal functions directly, bypassing the axum router / extractors / route registration.
- **Concurrent submits test doesn't exercise the claimed race** — single-threaded `#[tokio::test]` + tokio Mutex makes it pass even with `write_lock` removed.
- **No public docs for `freenet-app.toml`** — Tier 2 is effectively unreachable to webapp authors until a manifest spec ships.
- **No revocation UI** — grants are immortal, including across malicious webapp updates.

## Approach

Plain revert of the squash-merged commit `1739a04`. The PR is unreleased (latest tag is v0.2.56), nothing depends on `freenet-app.toml` yet, and Tier 2 is currently unreachable without public docs anyway — so reverting costs nothing user-facing and lets the work re-land with the issues above addressed.

Issue #4014 will be reopened with a checklist of what needs to land in a re-attempt.

## Testing

- `cargo check -p freenet` — clean
- `cargo clippy -p freenet --no-deps --all-targets` — clean
- `cargo fmt --check` — clean
- `cargo test -p freenet --lib path_handlers` — 57/57 pass

[AI-assisted - Claude]


sanity:
**test-exempt justification:** This is a revert of #4086. It removes that feature's code together with its tests; a revert restores the previously-tested main state, so no new regression test is applicable. The Rule Lint deleted-test rule fires only because the reverted feature's tests are removed alongside it. test-exempt label applied per CI Rule Lint guidance.

[AI-assisted - Claude]

PR #4100, opened 2026-05-12
feat(fdev): add --as-state flag to `execute update`

iduartgomez:
## Summary

- Add `--as-state` flag to `fdev execute update`. Wraps file as
  `UpdateData::State` instead of the default `UpdateData::Delta`.
- Default behavior is unchanged.

## Why

`fdev execute update` always sends the on-disk file as
`UpdateData::Delta`. Contracts whose `update_state` only matches
`UpdateData::State` then return `InvalidUpdate` and the failure
surfaces only as a generic "invalid contract update" log line.

This breaks the freenet-email facade contract used to give users a
stable bookmarkable URL across releases (see freenet/mail#200). Every
prior mail release silently fails to flip the facade pointer; users
hitting the bookmark URL see whatever stale state happened to be on
the network.

## Testing

- `cargo test -p fdev` — 35/35 pass, including two new unit tests
  covering `wrap_update_payload` (State arm + default Delta arm).
- `cargo clippy -p fdev --tests` — clean (no `items-after-test-module`
  or other warnings).
- Manual: rebuilt fdev release binary, ran
  `fdev execute update --as-state <FACADE_ID> <STATE_FILE>` against a
  live `freenet network` daemon. UPDATE accepted by facade contract
  (was rejected as `InvalidUpdate` without the flag). Gateway then
  served the loader HTML at the facade URL with HTTP 200.
- Default behavior (no `--as-state`) is unit-test verified to
  produce `UpdateData::Delta`, unchanged from prior behavior.

## Fixes

- Unblocks freenet/mail#200 (stable bookmarkable URL across releases).
  Without this flag the mail-side facade pointer flip silently fails
  on every release. Paired mail-side PR: freenet/mail#215.

## Tangential: pre-existing `release` positional bug

`UpdateConfig::release` is declared as a bare `bool` without
`#[arg(long)]`, so clap treats it as a positional with `SetTrue`,
which fails clap's debug_asserts the moment you try to
`try_parse_from` an `UpdateConfig` in a unit test. I noticed this
while trying to add a clap-level integration test for `--as-state`;
the helper-level tests in this PR sidestep the issue by exercising
`wrap_update_payload` directly. Recommend a follow-up to switch
`release` to `#[arg(long)]` like `PutConfig` already does.

### DIFF
--- a/crates/fdev/src/commands.rs
+++ b/crates/fdev/src/commands.rs
@@ -343,7 +343,11 @@ pub async fn update(config: UpdateConfig, other: BaseConfig) -> anyhow::Result<(
     let data = {
         let mut buf = vec![];
         File::open(&config.delta)?.read_to_end(&mut buf)?;
-        StateDelta::from(buf).into()
+        if config.as_state {
+            UpdateData::State(State::from(buf))
+        } else {
+            UpdateData::Delta(StateDelta::from(buf))
+        }
     };
     let request = ContractRequest::Update { key, data }.into();
     let mut client = start_api_client(other).await?;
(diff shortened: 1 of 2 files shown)


iduartgomez:
@claude do a /pr-review of this pr

Issue #4036, opened 2026-05-05
feat(delegates): restore-from-snapshot API and CLI for delegate secrets


sanity:
## Problem
PR #4034 added per-secret snapshot-on-write retention so prior values survive an accidental overwrite. The snapshots are encrypted, decryptable, and bounded — but there is no API or CLI to actually restore one. Today users must manually move files in `{delegate_dir}/.snapshots/{secret_id}/` over the active path.

## Proposal
A read API on `SecretsStore`:
```rust
pub fn list_snapshots(&self, delegate: &DelegateKey, key: &SecretsId)
    -> Result<Vec<SnapshotMetadata>, SecretStoreError>;

pub fn restore_snapshot(&mut self, delegate: &DelegateKey, key: &SecretsId,
                        timestamp_ms: u64) -> Result<(), SecretStoreError>;
```

And an `fdev` wrapper:
```
fdev delegate snapshots list <delegate_key> <secret_id>
fdev delegate snapshots restore <delegate_key> <secret_id> <timestamp_ms>
```

Restore semantics: copy the chosen snapshot back through the same hard-link + atomic rename path that `store_secret` uses, taking a fresh snapshot of the value being replaced first (so restore is itself recoverable).

## Why a separate issue
The base PR scoped to "preserve prior values, don't lose data". User-facing recovery UX has its own design questions (timestamp display, partial restores, dry-run). Worth its own review cycle.

## Related
- PR #4034 introduces the snapshot history this would consume
- #3050 vault delegate plan would expose restore through the multi-device sync layer eventually

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:37:30.909203709+00:00
Issue hash: `25593c76`

## Value Assessment

[...]

PR #4107, opened 2026-05-13
feat(secrets): list_snapshots / restore_snapshot APIs for delegate secrets

sanity:
## Problem

#4034 captures snapshots before every overwrite, but there is no API to actually restore one. Users would have to manually move files in `.snapshots/{secret_id}/` over the active path.

## Solution

Two new public methods on `SecretsStore`:

- `list_snapshots(delegate, key) -> Vec<SnapshotMetadata>` — read-only enumeration, oldest first. Empty Vec on a never-written secret (not an error).
- `restore_snapshot(delegate, key, timestamp_ms)` — atomic rename + fsync through the same path `store_secret` uses, plus a fresh pre-restore snapshot of the value being replaced (restore is itself reversible). Re-populates the ReDb index + in-memory map if a prior `remove_secret` had cleared them.

`SecretStoreError::SnapshotNotFound { key, timestamp_ms }` is a new typed variant so the future CLI can give a precise message instead of a generic IO error.

No decryption on restore: snapshots are ciphertext written by the delegate's own cipher and remain decryptable by `get_secret`. That makes restore usable even when the cipher hasn't been registered yet (recovery scenario where the user re-registers and then rolls a value back).

CLI work split per the issue's own guidance ("User-facing recovery UX has its own design questions ... worth its own review cycle"). Re-exports `SnapshotMetadata` + `SecretStoreError` through `freenet::dev_tool` so the CLI lands as a stacked PR on top of this one.

## Testing

- 5 new SecretsStore-level tests (`list_snapshots_on_unwritten_secret_is_empty`, `list_snapshots_returns_history`, `restore_snapshot_replaces_active_value`, `restore_snapshot_unknown_timestamp_errors`, `restore_after_remove_repopulates_index`).
- 4 new `secret_snapshots` tests (`parse_snapshot_name_returns_suffix`, `list_snapshots_returns_sorted_metadata`, `list_snapshots_orders_collision_suffixes`, `list_snapshots_missing_dir_is_empty`).
- One of those caught an actual sort-key bug: `(timestamp_ms, suffix.unwrap_or(0))` collapses `None` with `Some(0)`, so test_list_snapshots_orders_collision_suffixes initially failed. Fixed by explicit `None < Some(_)` ordering.
- `cargo clippy -p freenet --lib --tests -- -D warnings` clean.
- All 209 `wasm_runtime::*` tests pass on this branch.

## Stacked

This PR targets `feat/delegate-snapshots` (PR #4034). Merge that one first, then rebase this onto main.

Refs #4036

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -125,8 +125,10 @@ pub mod dev_tool {
         register_topology_snapshot, set_current_network_name, validate_topology,
         validate_topology_from_snapshots,
     };
+    pub use wasm_runtime::secret_snapshots::SnapshotMetadata;
     pub use wasm_runtime::{
-        ContractStore, DelegateStore, MockStateStorage, Runtime, SecretsStore, StateStore,
+        ContractStore, DelegateStore, MockStateStorage, Runtime, SecretStoreError, SecretsStore,
+        StateStore,
     };
 
     // Re-export simulation types for test infrastructure
--- a/crates/core/src/wasm_runtime.rs
+++ b/crates/core/src/wasm_runtime.rs
@@ -8,7 +8,7 @@ mod error;
 pub(crate) mod mock_state_storage;
 mod native_api;
 mod runtime;
-mod secret_snapshots;
+pub mod secret_snapshots;
 mod secrets_store;
 pub(crate) mod simulation_runtime;
 mod state_store;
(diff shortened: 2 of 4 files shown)


sanity:
Addressed the three info-level bot findings in commit 7641d3ad:

1. `list_snapshots` rustdoc now uses a `# Errors` section per code-style.md.
2. `SecretStoreError::SnapshotNotFound` has a `///` doc explaining the variant.
3. New direct pin test `restore_snapshot_prefers_unsuffixed_collision` exercises the stated tiebreak contract by hand-crafting collision snapshots with distinguishable ciphertexts and asserting which one `restore_snapshot` consumes.

Test caught an interesting interaction: under the default 2-year `max_age`, `thin_snapshots` after the first restore would drop the test's hand-crafted ancient-timestamped entries before the second restore could run. Worked around by overriding retention to `keep_last: 100, max_age: None` for that test. The default-retention behavior is correct; the test just needs to disable thinning to exercise collision-suffix logic in isolation.

[AI-assisted - Claude]

PR #4109, opened 2026-05-13
refactor(connect): retire legacy Operation mediator (#1454 phase 6)

iduartgomez:
## Problem

Phase 5 finished migrating GET/PUT/UPDATE/SUBSCRIBE off the legacy mediator. CONNECT was the last op still carrying `impl Operation for ConnectOp` + `load_or_init` + `process_message`, kept alive by the `handle_op_request::<ConnectOp, _>` fallthrough at `node.rs::handle_pure_network_message_v1`.

After an audit (see commit message), every production path that could reach `process_message` is unreachable: the reply variants are intercepted by the bypass and `Request` routes to `start_relay_connect`. The legacy `Operation` trait, the `handle_op_request`/`handle_op_result` re-entry loop, the `ComposeNetworkMessage` executor adapter, and the supporting helpers in `node.rs` are all orphaned.

## Solution

Delete the legacy mediator path end-to-end:

- `impl Operation for ConnectOp` + `store_operation_state*` helpers (~920 LoC in `connect.rs`)
- `Operation` trait, `OperationResult`, `OpInitialization`, `handle_op_request`, `handle_op_result`, `send_with_stream` (`operations.rs`)
- `ComposeNetworkMessage` trait + `Executor::op_request` method (`executor.rs`)
- `op_retry_backoff`, `handle_pure_network_result`, `forward_pending_op_result_if_completed`, `is_operation_completed`, `IsOperationCompleted` (`node.rs`)
- Outer `for i in 0..MAX_RETRIES` loop in `handle_pure_network_message_v1` (no `continue` survives)
- 8 unit tests that exercised the deleted helpers

Keep `OpEnum::Connect` + `ops.connect` DashMap + `ConnectOp` struct + its state-machine helpers for the follow-up slice — the in-file relay/joiner tests still drive `state.handle_request(...)` directly, and `state.handle_request` is also called by the production task driver. Dead-code allows + comments mark the items pending removal.

Net diff: **-1642 LoC** across 8 files.

## Testing

- `cargo fmt`
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo test -p freenet --lib --features testing` → 2481 passed, 14 ignored
- [ ] CI green
- [ ] `/linux-test connectivity` green
- [ ] Spot-check 5-node simulation network forms / heals partitions

## Fixes

Follow-up to #1454 (multi-PR refactor). No individual issue.

## Followup slice

- Retire `OpEnum` + `OpEnum::Connect`, the `ops.connect` DashMap + `has_connect_op` + `OpManager::push` for connect + `pending_op_counts` connect slot
- Retire `ConnectOp` struct + test-only helpers (`new_joiner`, `new_relay`, `with_state`, `is_completed`, `finalized`, `handle_observed_address`, `handle_response`)
- Retire orphan `op_request_channel` / `run_op_request_mediator` / `OpRequestSender` plumbing in `executor.rs` (cascades through `RuntimePool` / `Executor::from_config_*` / `p2p_impl`)
- Migrate `connect.rs` in-file tests onto the production `state.handle_request` surface or delete duplicates already covered by `op_ctx_task.rs` tests


iduartgomez (commented):
## Findings

Code-first review — read the diff before reading the description.

### [warning] `handle_aborted_op` Connect branch is now dead code

`node.rs::handle_aborted_op` (line ~2074) still calls `op_manager.pop(&tx)` and branches on `Ok(Some(OpEnum::Connect(op)))` to drive gateway-retry / `prune_in_transit_connection` / `join_ring_request` on `Aborted` messages. After this PR no production code path inserts into `ops.connect` (`OpManager::push` is `#[allow(dead_code)]`, the only writer at `op_state_manager.rs:600`), so `pop()` always returns `Ok(None)` for Connect transactions and the two `Ok(Some(OpEnum::Connect(op)))` arms become unreachable.

This may be fine — `start_client_connect` owns its own retry/timeout/backoff in task locals, and inbound `ConnectFailed`/`Rejected` carry the recovery semantics — but the PR description does not call out that the joiner-side response to a transport-level `NetMessageV1::Aborted(tx)` (e.g. sent by `op_ctx.rs:591`) silently no-ops for CONNECT post-merge. The new driver also doesn't subscribe to `Aborted`-type wire frames (no occurrences in `connect/op_ctx_task.rs`). Worth confirming the connect driver doesn't rely on Aborted-triggered cleanup, and either documenting the change or scheduling the removal of the `Ok(Some(OpEnum::Connect(...)))` arms alongside the DashMap deletion slice.

### [warning] Blanket `#![allow(dead_code)]` on connect.rs is broader than necessary

The new `#![allow(dead_code)]` at the top of `connect.rs` suppresses dead-code warnings for the entire 1500-line module. The PR description frames this as protecting the test-only `impl Operation` surface, but it also suppresses any *truly* dead method that future edits introduce. Compare with the targeted `#[allow(dead_code)]` annotations on `OpManager::push`, `has_connect_op`, `OpEnum::finalized` and `OpError::{IncorrectTxType,OpNotPresent}` — those are per-item. Consider replacing the blanket attribute with per-item `#[allow]`s (or a `#[cfg(test)]`-gated submodule) so the next slice can rely on warnings to find the actual removal surface.

### [info] CONNECT `Request` with `source_addr=None` audit claim — solid

`start_client_connect` builds its request via `send_to_and_collect_replies(gateway_addr, ...)` (op_ctx_task.rs:189) with a concrete `SocketAddr` — there is no loopback path, so a `ConnectMsg::Request` will never arrive at `handle_pure_network_message_v1` with `source_addr=None`. The new dispatch branch correctly drops it with a debug log if it ever did. Claim verified.

### [info] Outer `for i in 0..MAX_RETRIES` removal — safe

The retry loop existed to swallow `OpError::OpNotAvailable(Running)` returned by `handle_op_request`. With the mediator gone, no path inside `handle_pure_network_message_v1` calls `pop()` any more, so `Running` is never produced and the loop has nothing to do. `pop()` itself still returns `Running` for `under_progress` txs, but its only remaining caller is `handle_aborted_op`, which swallows the error with `_ => {}` (and is dead for CONNECT anyway — see first finding).

### [info] `connect_branch_window()` pin — anchor is robust

`NetMessageV1::Put(ref op) => {` first appears at the actual PUT match arm (line 936), well before any later occurrence in tests. The bypass-anchor 16-space indent matches the new match-body indentation. Looks fine.

### [nit] Pre-existing pin guard is self-referential (not introduced by this PR)

`bypass_is_wired_into_subscribe_branch_regression_guard` (tests/node.rs:2715) uses `SOURCE.find("handle_op_request::<subscribe::SubscribeOp, _>")` as its window-end anchor against `include_str!("node.rs")`. The only remaining occurrence of that literal in the file is *inside the test itself* (line 2729), so the window spans from the SUBSCRIBE branch to the test, and the `window.contains("try_forward_task_per_tx_reply(")` assertion passes vacuously (the dispatch code in between trivially contains it). The UPDATE counterpart at line 3238 already mitigates this by composing the needle at runtime — applying the same fix here would be a one-line change and would restore the guard's value. Not strictly a PR #4109 concern (introduced when SUBSCRIBE migrated), but this PR is a natural place to fix it.

### Cross-crate audit

`rg "Operation\b|OperationResult|handle_op_request|ComposeNetworkMessage|op_request\b|is_operation_completed|IsOperationCompleted|OpInitialization|forward_pending_op_result_if_completed|op_retry_backoff|handle_pure_network_result"` across `apps/` and `docs/` returned only a comment hit in `apps/freenet-ping/app/tests/test_50_node_operations.rs` ("// 1. PUT Operation"). No real consumers outside `crates/core/src/`. Within `crates/core/src/`, all remaining `handle_op_request` mentions are comments/doc strings in driver/test files plus the self-referential SUBSCRIBE pin above.

### Conclusion

The deletion is structurally clean and the audit claims I could verify (loopback impossibility, retry loop redundancy, anchor stability, cross-crate orphan-freeness) hold up. The two items worth resolving before merge: (1) explicitly document or excise the dead CONNECT branch in `handle_aborted_op` so it doesn't survive the follow-up slice as silent dead code, and (2) consider tightening the blanket `#![allow(dead_code)]` on `connect.rs` to per-item annotations.

PR #4110, opened 2026-05-13
refactor(connect): retire ops.connect DashMap + OpEnum + executor mediator (#1454 phase 7)

iduartgomez:
## Problem

After #1454 phase 6 retired the legacy `Operation` mediator
(impl Operation for ConnectOp, `handle_op_request`, `handle_op_result`,
`process_message`), several layers of scaffolding remained as dead
carriers:

- `OpManager::ops.connect` DashMap had no production writer; its
  consumers (`push`/`pop`/`peek_target_peer`/`peek_next_hop_addr`/
  `has_connect_op`/GC sweep) still threaded through the codebase.
- `OpEnum` only had `Connect` variant left, fed by the same dead pop.
- `OpRequestSender`/`OpRequestReceiver`/`op_request_channel`/
  `run_op_request_mediator`/`Callback::response` (the executor →
  event-loop mediator subsystem) had no producer — executors initiate
  network ops directly through driver entry points
  (`subscribe::run_executor_subscribe`, `start_client_update`).
- `executor.op_sender` field threaded through every contract handler
  builder, every executor mock constructor, and 4 RuntimePool builders.
- `handle_aborted_op` was a no-op stub kept only to satisfy 4 call
  sites; the orphan-on-prune call and the wire `Aborted` arm did the
  same no-op log.

## Solution

Cascade deletes the carrier layer:

1. **ops.connect DashMap** (commit 1) — field, all accessors,
   GC sweep CONNECT arms, `record_connect_uphill_timeout` helper +
   tests, `OpNotAvailable` enum + `OpError::OpNotAvailable` variant.
   `peek_*` kept as `-> None` stubs (callers fall back to the
   connection-manager lookup chain).

2. **`Result<Option<OpEnum>, OpError>` → `Result<(), OpError>`**
   (commit 2) — `handle_pure_network_message{,_v1}` returned `Ok(None)`
   from every arm. Simplified to `Ok(())`. `report_result`'s
   `Ok(Some)` arm + `classify_op_outcome` helper retired; drivers
   record their own dashboard stats / route events / executor
   callbacks at finalization.

3. **`OpEnum` + the executor mediator subsystem** (commit 3) —
   430-line `mediator_tests.rs` deleted; mediator types deleted;
   `op_sender` field/param dropped from `Executor::new`,
   `RuntimePool::new`, every `ContractHandler::build`, mock
   constructors. `ExecutorToEventLoopChannel`/`mediator_channels`/
   `NetworkEventListenerHalve` retained as no-op stubs to avoid
   invasive `PrioritySelectStream` generic-slot surgery — the
   `Stream::poll_next` impl returns `Pending` forever.

4. **`handle_aborted_op` callers** (commit 4) — orphan-on-prune now
   just logs the count; wire `Aborted` arm logs and returns. 3
   transport-handshake-fail breadcrumb call sites retained for now.

5. **Docs** (commit 5) — operations.md updated to reflect the new
   ground truth (every per-op DashMap retired, no central carrier).

Net diff: **−1500 LoC code, +15 LoC docs** across 18 files.

## Testing

- `cargo build -p freenet` clean.
- `cargo clippy -p freenet --all-targets -- -D warnings` clean.
- `cargo test -p freenet --lib` — 2461 passed, 14 ignored
  (down 17 from the deleted `mediator_tests.rs` mod).
- New pin test in `operations/put.rs`:
  `completed_must_not_touch_per_op_dashmaps` — replaces the prior
  positive-assertion variant (which the DashMap removal invalidated)
  with a regression guard against re-introducing any
  `self.ops.{connect,get,put,update,subscribe}` reference in
  `OpManager::completed`.

## Fixes

Refs #1454. Closes the executor-mediator + DashMap-carrier surface
that phase 6 left dangling. `ConnectOp::{new_joiner,new_relay,with_state}`
test fixtures and the surrounding in-file unit tests survive — they
exercise the relay-side `handle_request` state-machine helper still
used by the production driver.


iduartgomez:
Adding `test-exempt` label: this PR retires the legacy mediator subsystem (`OpEnum`, `ops.connect` DashMap, executor↔event-loop mediator, `classify_op_outcome`). The 15 removed tests exercised types that no longer exist:

- `pool_tests/mediator_tests.rs` (10 tests) — tested `OpRequestSender`/`run_op_request_mediator`/`Callback`/`ExecutorToEventLoopChannel`, all deleted.
- `record_connect_uphill_timeout_tests` (3 tests) — tested a helper on `Ops::connect` accessors that are gone.
- `classify_op_outcome_tests` (5 tests) — tested an `OpEnum` dispatch helper; no `OpEnum` left to classify.

These tests cannot be `#[ignore]`d because the symbols they reference (`OpRequestSender`, `OpEnum::Connect`, `record_connect_uphill_timeout`) are deleted in the same PR. The behavior they validated is now either gone (mediator) or covered by the task-per-tx driver path validated by the `Unit & Integration` job.

Issue #4103, opened 2026-05-12
release-agent Tier 3 follow-ups (replay, artifact signing, credential hardening)


sanity:
Tracks the follow-up items deferred from the multi-model review of #4082 (Tier 3 in the consolidated punch list). All of these are design-thinking or larger-than-this-PR work. None block Phase 1 (build-only, dry-run); some should land before Phase 2 (live nova deploy) and the rest before Phase 3 (workflow takes over).

### 1. Persistent replay protection
Currently `update_handler` only enforces ±5-min `issued_at` skew + in-memory rate limit. The rate-limit state is lost on agent restart, so a captured signed request can be replayed in any future 5-minute window after a bounce. The HMAC validates the bytes, not the freshness or uniqueness of the request.

Options:
- Persist last-accepted `issued_at` per-source to disk; reject anything not strictly greater.
- Short-TTL replay cache (last N signatures or `issued_at` values) on disk.
- Single-use nonces issued by the workflow.

Surfaced by the skeptical reviewer as CRITICAL #1.

### 2. Artifact signature verification
The agent validates `request.version == GitHub latest`, but `gateway-auto-update.sh` then downloads whatever GitHub returns for that tag. A compromise of GitHub itself, or a stolen GitHub release-push token, installs arbitrary code as root on every gateway. Adding cosign / minisign over the release tarballs and verifying in `gateway-auto-update.sh` closes that gap.

Out of scope for this issue: standing up the signing infrastructure in `release.yml`. Track separately if needed.

Surfaced by the skeptical reviewer as HIGH #4.

### 3. Cross-compile entry for `freenet-release-agent`
`cross-compile.yml` builds `freenet` and `fdev`. Phase 2's nova deploy is manual (operator copies the locally-built binary), which is fine for one host but doesn't scale to vega/technic in Phase 3. Add the agent to the same matrix so the workflow can pull a CI-built linux-musl tarball.

Surfaced by the big-picture reviewer.

### 4. Cache GitHub `latest` for a few minutes
Anonymous GitHub API is 60 req/hr/IP. An attacker who can send valid-HMAC requests, or just bursty CI, can exhaust the agent's quota and block legitimate updates. Cache `latest` for ~5 minutes between checks. Tied to the `LatestSource` seam from PR #4082 — a `CachingLatest` wrapper is straightforward.

Surfaced by the skeptical reviewer as HIGH #5.

### 5. Status-sentinel file for last-update outcome
[...]

Issue #13, opened 2021-10-30
Write an anti "write amplification" layer to avoid running the same ops repeteadly


iduartgomez:
Due to broadcasting and forwarding mechanisms is very possible nodes may be hit several times, from different peers, to replicate the same exact operation in the case of update notifications.

The solution to avoid duplicate costly work (like executing the same contract with the same value over and over) and be safe that this will apply to any ops is to keep a commit log for transactions, since the history of an initial operation is covered by the span of the same id, and double check that this operation has not been already commited previously and resolved.

The commit log can be clean up when garbage from old ops who have timed out is cleaned up based on this same time out duration.

sanity:
Could we solve this by storing hashes of recently received updates and disregarding duplicates?

iduartgomez:
I will check the exact details when I fix this but a combination of both things is probably the most optimal: id + hash of the op; so you can shard by transaction id and then deduplicate by hash (instead of having to compare every transaction). 

The commit log can be of the hashes. It may be unnecessarily expensive to do that though since we can embed (and should do) emit time of the individual and that + the id may be enough. 

There are certainly several ways this can be solved. Openned the issue so I don't forget it has to be fixed.

iduartgomez:
Completed in #877 and #878

Issue #11, opened 2021-10-17
Karma: decentralized reputation and trust


sanity:
Trust is the ability to predict whether an identity will do what they say they will do in the future. Karma's purpose is to allow identities to establish trust from their first interaction.

## History

F2's Karma mechanism is inspired by Freenet's [Web of Trust](https://github.com/freenet/plugin-WebOfTrust) plugin.

## Uses

* Prevent resource leaching

## Design

```rust
#[derive(Serialize, Deserialize)]
pub struct TrustLog {
  pub public_key : EcKey<Public>,
  /// Trust entries signed by public_key
  pub log : Vec<(TrustEntry, EcSignature)>,
}

#[derive(Serialize, Deserialize)]
pub struct TrustEntry {
  /// Entry creation time in milliseconds since epoch
  pub timestamp : u64,
  /// What did this entity promise to do, in a standardized but extensible format
  pub promise : str,
  /// Did the entity follow-through on its promise?
  pub kept : bool,
}
```

sanity:
Closing as this is made obsolete by other work.

PR #4114, opened 2026-05-13
feat: auto-trigger gateway updates + announcements on release.published

sanity:
## Problem

PR #4082 landed the gateway-side HTTP listener (Phase 1, agent in dry-run mode), but a developer creating a GitHub release still has to manually drive the release cascade — SSH to gateways, post to Matrix, run `riverctl` locally. The agent is in place, but nothing in CI uses it.

## Approach

Wire the rest of the chain so a published GitHub Release auto-cascades:

1. **`gateway-update.yml`** — flipped from `workflow_dispatch`-only to also trigger on `release: { types: [published] }`. Resolves the version from `github.event.release.tag_name` when fired by release. Manual dispatch retained for re-rollouts and the `dry_run_workflow: true` smoke test.

2. **`release-announce.yml`** (new) — fires on `release.published`, two parallel jobs (no fail-fast — Matrix outage shouldn't block River):
   - **Matrix**: direct `curl PUT` to client-server API. Needs `MATRIX_HOMESERVER_URL` + `MATRIX_ACCESS_TOKEN` secrets.
   - **River**: HMAC-signs a request to nova's release-agent at `/announce/river`. The agent invokes `sudo -u ian announce-to-river.sh "<message>"` locally — signing key and Freenet node stay on nova; CI never holds either.

3. **Agent gains `POST /announce/river`** — new `src/announcer.rs` module structurally parallel to `updater.rs` (same sudo-wrapping, 1s probe, dry-run short-circuit). The endpoint:
   - HMAC over body, clock-skew check (reuses `/update`'s validators)
   - Message must be non-empty, ≤ 4 KiB, NUL-free
   - 1-minute rate limit (separate window from `/update`'s 10-minute one)
   - Opt-in via `river_announce_command` + `river_announce_user` config fields. Unconfigured gateways (vega) return 503.

4. **`announce-to-river.sh`** (new) — mirrors the `announce_to_river` block in `scripts/release.sh`: locates `rooms.json` + signing key, re-syncs the signing key into rooms.json, runs `RIVER_SKIP_CONTRACT_CHECK=1 cargo run -p riverctl message send`. Configurable via `/etc/freenet-release-agent/announce.env`.

5. **Sudoers extended** for the new script: `freenet-update ALL=(ian) NOPASSWD: /usr/local/bin/announce-to-river.sh *`. Same wildcard-safety pattern as gateway-update: agent validates before invoking; sudoers itself doesn't constrain argv shape.

## Testing

6 new integration tests under `crates/release-agent/tests/update_handler.rs`:

- `announce_disabled_returns_503` — opt-in contract
- `announce_happy_path` — full pipeline with fake-sudo argv recorder confirming `sudo -n -u <user> <script> <message>` shape
- `announce_missing_signature_is_401`
- `announce_empty_message_is_400`
- `announce_oversize_message_is_413` (4 KiB+ message cap)
- `announce_spawn_failure_does_not_consume_rate_limit` — script `exit 1` → 500 + window NOT consumed

Test totals: 23 unit + 20 integration = 43 (was 37 before this PR). `cargo fmt` + `cargo clippy --all-targets -D warnings` clean.

## After merge — operational steps

To make the chain live:

1. Install the updated agent on nova (re-run `install.sh`)
2. Set `river_announce_command = "/usr/local/bin/announce-to-river.sh"` + `river_announce_user = "ian"` in nova's config; restart agent
3. Configure GitHub secrets: `MATRIX_HOMESERVER_URL`, `MATRIX_ACCESS_TOKEN`, plus the existing `RELEASE_AGENT_HMAC_NOVA`/`VEGA` from PR #4082
4. Cut a release — the `release.published` event will drive the full cascade

Documented in `docs/release-agent-deployment.md` (new GitHub Actions secrets section + River-path architecture diagram).

## Risk

The new endpoint is opt-in (off by default) and the workflows are no-ops when their respective secrets aren't set, so merging this PR is safe even before the operational steps above. The first cascade-triggering release will be the live test.

Builds on #4082 ([AI-assisted - Claude])

### DIFF
--- a/.github/workflows/gateway-update.yml
+++ b/.github/workflows/gateway-update.yml
@@ -49,13 +49,19 @@ jobs:
     steps:
       - name: Resolve version
         id: ver
+        # Pass values via env: rather than ${{ }} interpolation — a tag
+        # name like `1.2.3"; curl evil; "` would otherwise template into
+        # the script at expression-eval time, before the regex below
+        # could reject it. Env binding makes it a regular shell string.
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          TAG_NAME: ${{ github.event.release.tag_name }}
+          DISPATCH_VERSION: ${{ inputs.version }}
         run: |
-          # On release-trigger, version comes from the tag; on
-          # workflow_dispatch, from the user input. Strip leading `v`.
-          if [[ "${{ github.event_name }}" == "release" ]]; then
-            RAW="${{ github.event.release.tag_name }}"
+          if [[ "$EVENT_NAME" == "release" ]]; then
+            RAW="$TAG_NAME"
           else
-            RAW="${{ inputs.version }}"
+            RAW="$DISPATCH_VERSION"
           fi
           VERSION="${RAW#v}"
           if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
--- a/.github/workflows/gateway-update.yml
+++ b/.github/workflows/gateway-update.yml
@@ -1,27 +1,18 @@
-# Gateway update workflow (Phase 2 — manual trigger only)
+# Gateway update workflow.
 #
-# Status: draft. This workflow is `workflow_dispatch` only and intentionally
-# NOT wired into release.yml or cross-compile.yml. Before flipping to
-# automatic triggering on `release: { types: [published] }`, the following
-# must be true:
+# Auto-triggers on `release.published` — once a release is published on
+# GitHub (after release.yml runs cargo publish + creates the release with
+# binaries attached), this workflow signs HMAC-authed `POST /update`
+# requests to each gateway's `freenet-release-agent`.
 #
-#   - freenet-release-agent deployed on each gateway in dry_run mode and
-#     validated per docs/release-agent-deployment.md
-#   - RELEASE_AGENT_HMAC_NOVA and RELEASE_AGENT_HMAC_VEGA secrets set
-#   - Gateway URLs reachable from GitHub Actions runners
-#   - At least one manual run completed end-to-end without errors
-#
-# When all of the above is true, add a second `on:` trigger:
-#
-#   release:
-#     types: [published]
-#
-# and remove the `workflow_dispatch.inputs.version` field (the release tag
-# becomes the version).
+# `workflow_dispatch` is kept for manual re-rollouts and for the smoke
+# test (`dry_run_workflow: true`).
 
-name: Gateway Update (Phase 2 draft)
+name: Gateway Update
 
 on:
+  release:
+    types: [published]
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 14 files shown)


sanity:
Pushed `165ed577` — addresses Tier 1 + Tier 2 from the five-reviewer pass (code-first / testing / skeptical / big-picture / Codex).

## Tier 1 — production safety (blocking)

| Finding | Source | Fix |
|---|---|---|
| `RAW="${{ github.event.release.tag_name }}"` was direct shell interpolation in both workflows → script-injection vector if a maliciously-crafted tag is pushed | Skeptical #10 | Switched both workflows to `env:`-bound vars so the tag becomes a regular shell string |
| `announce_river_handler` returned 503 BEFORE HMAC verify → unauthenticated probe could fingerprint which gateways enable the endpoint | Skeptical #11 + Code-first | Reordered: HMAC first, then is_configured check |
| `announce-to-river.sh` Python heredoc used `2>/dev/null \|\| true` → schema drift / missing python3 silently posted with stale signing key | Skeptical #2 + Code-first | Removed the silencer; failures abort with a clear log line. Rewrite is now atomic (`os.replace` on a temp file). |

## Tier 2 — visible quality

| Finding | Source | Fix |
|---|---|---|
| Workflow accepted HTTP 200 from `/announce/river` even though prod nova returns 202 → silent dry-run pass-through | Skeptical #1 | Warn if 200 + `dry_run: true` (announcement was NOT posted) |
| Matrix TXN_ID was `date +%s + $RANDOM` (~1/32k collision) | Skeptical #13 | `uuidgen` with date+RANDOM fallback |
| `release.sh` would double-post Matrix/River once the workflow path is wired | Big-picture | Gates added: `FREENET_RELEASE_SKIP_ANNOUNCEMENTS=1` and `FREENET_RELEASE_SKIP_GATEWAY_SSH=1` opt the local script out |
| `install.sh` message implied conditional install of announce-to-river.sh | Code-first | Script always installed; endpoint stays disabled until config opts in |
| 6 gap-fill tests for `/announce/river`: tampered body, stale issued_at, NUL byte, boundary at 4096, rate-limit-after-success, opt-in-still-requires-signature | Testing | All added; total now 27 integration (was 20) |
| Fake-sudo recorder didn't pin the `-u <user>` shape — only the message survived shifts | Testing #3 | Now records pre-shift argv too; `announce_happy_path` asserts `--non-interactive -u nobody …` prefix matches sudoers allowlist |
| Embedded NUL byte in a test comment (line 749) made grep treat the file as binary | Codex P3 | Replaced with `\u{0}` literal |

## Deferred to #4103 (Tier 3 follow-ups)

- Separate HMAC secrets per endpoint (Skeptical #6 — currently `/update` and `/announce/river` share the secret)
- Probe duration tuning (Skeptical M3 — pre-existing)
- Zombie task accumulation on stuck spawns (Skeptical #5)
- UTF-8 byte-vs-char doc note in handler comment (Skeptical #7)
- `~/.claude/skills/freenet-release/SKILL.md` references the SSH flow (Big-picture) — global file, needs explicit user authorization

## Test totals

23 unit + 27 integration = 50 (was 43). `cargo fmt` + `cargo clippy --all-targets -D warnings` clean.

[AI-assisted - Claude]

Issue #3340, opened 2026-02-28
fix: stale local cache served when hosting LRU outlives subscription


sanity:
## Problem

In `client_events/mod.rs:1216-1218`, the GET routing decision uses `should_seed()` to determine whether to return local cached state or fetch from the network:

```rust
let is_subscribed = full_key
    .as_ref()
    .map(|k| op_manager.ring.should_seed(k))
    .unwrap_or(false);

if local_satisfies_request && (connection_count == 0 || is_subscribed) {
    // return local cache, skip network
}
```

`should_seed()` → `should_host()` returns true if:
1. Active network subscription (correct — cache is fresh via updates)
2. Local client subscriptions (correct — cache is fresh)
3. **Contract is in the hosting LRU cache** (incorrect — no subscription = no updates = stale cache)

After a subscription expires (20-min TTL), if the contract remains in the hosting LRU cache, the node serves stale local state instead of fetching from the network. Since it never goes to the network, `AUTO_SUBSCRIBE_ON_GET` never fires, so it never re-subscribes. **Stale forever** until LRU eviction.

## Fix

Change the check from `should_seed()` to a stricter check that only returns true when the node is **actually receiving updates** — i.e., active network subscription OR local client subscriptions. The hosting LRU cache alone should not be sufficient to skip the network path.

Key files:
- `crates/core/src/client_events/mod.rs:1216-1218` — the GET routing decision
- `crates/core/src/ring/mod.rs` — `should_seed()` / `should_host()` 
- `crates/core/src/ring/hosting.rs:629-633` — `should_host()` includes `is_hosting_contract()`

[AI-assisted - Claude]

PR #4132, opened 2026-05-15
fix(ring): persist local-client-access across hosting LRU eviction

sanity:
## Problem

Self-hosted contracts whose main-cache LRU entry was evicted under byte-budget pressure silently fall through the `is_locally_hosted` shortcut in `client_events::serve_get` and route GETs over the network. For contracts with no remote subscribers (common for low-traffic hosted contracts like the freenet-stdlib mirror demo `prefix=96rknpy1GYhZ`), this GetOp has nowhere to go and the client times out at the WS layer — observed as a 180s timeout on the freenet-git rescue/mirror runs against this contract for at least the past week (2026-05-14 logs show three independent 180s timeouts on the same day, while the same contract serves fine when its entry is in the main cache).

Root cause (verified against gateway logs and source):
- `mark_local_client_access` and `has_local_client_access` (in `ring/hosting/cache.rs`) both read/write the inline `local_client_access` flag on the main `HostingCache` LRU entry.
- Once the LRU evicts an entry (byte-budget pressure with TTL elapsed), the flag is gone: `mark_local_client_access` becomes a no-op and the next `has_local_client_access` returns false.
- The `is_locally_hosted` gate in `client_events::serve_get` reads `is_hosting_contract(k) && has_local_client_access(k)`. Both clauses are LRU-presence-dependent and both go false after eviction, so the gate never re-fires for self-hosted contracts until something repopulates the LRU. The fall-through path attempts network routing of a GetOp that hangs because no remote peer subscribes to this contract's location.

## Solution

**`ring/hosting/cache.rs`**: keep an independent bounded LRU `local_access_history: LruCache<ContractKey, Instant>` (capacity `LOCAL_ACCESS_HISTORY_CAPACITY = 4096`) on `HostingCache`.

- `mark_local_client_access` always inserts into `local_access_history`. The existing flag on the main cache entry is still updated when present (so disk-persistence logic in `ring/hosting.rs::mark_local_client_access` continues to work unchanged).
- `has_local_client_access` precedence is now:
  1. **Main cache present**: entry's flag is authoritative. A relay-only re-cache after eviction lands with `local_client_access = false`, and that false reading must win — otherwise we'd serve stale forwarded state as if it were locally-owned. Preserves the `test_eviction_clears_local_client_access` invariant.
  2. **Main cache missing**: fall back to history. This is the post-eviction self-hosted path this PR fixes.

**`client_events.rs`**: drop the now-redundant `is_hosting_contract` clause from the `is_locally_hosted` gate. `is_hosting_contract` is itself just `cache.contains(key)` — LRU-presence-dependent, and was the second half of the same eviction-induced false reading. `has_local_client_access` is now the single authoritative "this contract is locally-owned, not relay-cached" signal.

The 4096-entry history cap satisfies the AGENTS.md cleanup-exemption rule (`ring.md`: cleanup exemptions must be time- or count-bounded). At ~100 bytes per entry that's <500 KB at full occupancy — comfortably above realistic per-gateway hosted-contract counts.

## Testing

Two new unit tests in `ring/hosting/cache.rs::tests`:

- `test_local_client_access_persists_across_main_lru_eviction` (the regression test): records access, forces eviction via byte-budget pressure + TTL advance, asserts the access flag survives. **Fails on main, passes with this change.** Also includes a negative control (never-marked contract returns false).
- `test_local_access_history_capacity_is_bounded`: marks `LOCAL_ACCESS_HISTORY_CAPACITY + 1` distinct contracts (note: `ContractKey::PartialEq` is instance-id-only, so the test spreads `i` across instance bytes to generate truly distinct keys), asserts the oldest is evicted while the newest remain.

Updated `test_local_client_access_unknown_contract` in `ring/hosting.rs` (renamed to `_records_in_history`) to reflect the new post-fix behavior: marking on an unknown contract now records the access in history, so `has_local_client_access` returns true. Old assertion of no-op-ness is gone; old "no panic" intent is preserved (the test still verifies marking unknown contracts doesn't crash).

All 64 tests under `ring::hosting` pass. `cargo fmt --check` and `cargo clippy -p freenet --all-targets -- -D warnings` clean.

What this PR does **not** address (separate follow-ups):
- The misleading "no state found at current contract key or any of 0 legacy keys" error in `freenet-git::wsclient::get_state_with_legacy_fallback` that surfaced a timeout as if it were permanent data loss. Diagnostic fix tracked separately and will ship in the next freenet-git release.
- `has_recent_local_client_access` (used for cleanup-exemption with TTL) still reads only the main cache. That's a separate use case with its own TTL semantics; not broadened in this PR to keep scope focused.

## Fixes

No GitHub issue filed — discovered while investigating recurring rescue-demos Matrix failure alerts for the freenet-stdlib mirror.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -1096,14 +1096,19 @@ async fn process_open_request(
                             }
                         }
 
-                        // Serve from local cache only if the local user requested
-                        // this contract (not just relay-cached).
+                        // Serve from local cache only if a local client ever
+                        // accessed this contract (i.e. we hosted it on their
+                        // behalf rather than just relay-caching state from
+                        // network forwarding). `has_local_client_access` is
+                        // backed by an LRU history that survives main-cache
+                        // byte-budget eviction; the older `is_hosting_contract
+                        // && has_local_client_access` gate dropped to false
+                        // the moment the main LRU evicted the entry, forcing
+                        // cold GETs into network routing of a GetOp that may
+                        // have no remote subscribers.
                         let is_locally_hosted = full_key
                             .as_ref()
-                            .map(|k| {
-                                op_manager.ring.is_hosting_contract(k)
-                                    && op_manager.ring.has_local_client_access(k)
-                            })
+                            .map(|k| op_manager.ring.has_local_client_access(k))
                             .unwrap_or(false);
 
                         // Return local cache if we have valid state AND any of:
--- a/crates/core/src/ring/hosting.rs
+++ b/crates/core/src/ring/hosting.rs
@@ -2671,15 +2671,24 @@ mod tests {
         );
     }
 
-    /// Marking and querying unknown contracts should be no-ops (no panic).
+    /// Marking an unknown (not-in-main-cache) contract records the access in
+    /// the local-access history so subsequent `has_local_client_access`
+    /// queries return true. This is the documented post-fix behavior — the
+    /// old no-op behavior caused the `is_locally_hosted` shortcut in
+    /// `client_events::serve_get` to never fire for evicted-but-still-on-disk
+    /// contracts, forcing cold GETs into network routing.
     #[test]
-    fn test_local_client_access_unknown_contract() {
+    fn test_local_client_access_unknown_contract_records_in_history() {
         let manager = HostingManager::new();
         let contract = make_contract_key(1);
 
         assert!(!manager.has_local_client_access(&contract));
-        manager.mark_local_client_access(&contract); // no-op, not in cache
-        assert!(!manager.has_local_client_access(&contract));
+        manager.mark_local_client_access(&contract);
+        assert!(
+            manager.has_local_client_access(&contract),
+            "mark_local_client_access on an unknown contract must record in \
+             access history so the local-cache shortcut can fire"
+        );
     }
 
     /// The local_client_access flag should be sticky -- once set, it should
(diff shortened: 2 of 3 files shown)


sanity:
Closing this approach per review findings. The two HIGH-severity issues raised by the skeptical and Codex reviewers both point at the same root design problem:

1. **Stale-state regression** (B1, codex P2 #1, skeptical H1): the `is_locally_hosted` gate was the load-bearing freshness guard from #3340. My change to drop the `is_hosting_contract` clause + bootstrap the flag via history on every GET undermines that guard for relay-cached evicted contracts.

2. **Renewal bookkeeping broken** (B2, codex P2 #2): the new shortcut returns before `record_access`, so the contract never gets re-added to the hosting LRU, never appears in `contracts_needing_renewal`, never re-establishes subscription. First GET avoids the hang but the gateway is left in a "serving stale local, not actually hosting" state.

Both surface a deeper issue: the freenet-stdlib mirror contract has no `local_client_access` flag set in the first place because **PUT never calls `mark_local_client_access`** — only GET does. So the contract isn't renewal-protected, eventually gets evicted under byte pressure, and the `is_locally_hosted` shortcut correctly evaluates to false (we are not, in fact, actively hosting it anymore).

Pivoting to a much smaller fix: call `mark_local_client_access` from the local-client PUT path, so self-hosted contracts get renewal protection from the moment they're hosted. The existing cache + gate semantics stay intact.

[AI-assisted - Claude]

Issue #3769, opened 2026-04-05
feat: track local client interest to enable safe subscription renewal and fast cache serving


sanity:
## Problem

Two optimization attempts failed because the system can't distinguish "contracts the local user cares about" from "contracts this node relayed for others":

- **#3763** (hosted renewal): Tried subscribing all hosted contracts after restart to keep local cache fresh. But the hosting cache includes relay-cached contracts (hundreds), causing a subscription storm (882 renewals/2hr, disk-filling logs).
- **#3762** (piggyback subscribe): Tried establishing subscriptions inline during GET response propagation. But it fired on every relay node in the path, not just the originator, causing the same storm.

Both were reverted (#3765, #3767). The underlying UX problems remain:

1. **After restart**, hosted contracts serve potentially stale state (#3761 `is_hosted` shortcut) because subscriptions are lost and there's no safe way to re-subscribe only user-relevant contracts.
2. **Between first GET and subscribe completion** (10-50s), repeated page loads go to the network instead of serving from local cache because `is_subscribed` is false.

## Root Cause

The hosting cache (`hosting_cache` in HostingManager) mixes two kinds of contracts:
- **Locally requested**: contracts the user accessed via HTTP/WebSocket (River, Delta, etc.)
- **Relay cached**: contracts the node cached while forwarding GETs/PUTs for other peers

There's no flag or separate tracking to distinguish them. Every optimization that treats "hosted" as "user cares about this" accidentally includes relay traffic.

## Proposed Approach

Add a `local_client_access: bool` flag (or similar) to the hosting cache entry, set when a contract is accessed via the local client API (HTTP `contract_home`, WebSocket GET/PUT/SUBSCRIBE), NOT when accessed via network relay.

This enables:

### Safe subscription renewal
`contracts_needing_renewal()` can include hosted contracts WITH `local_client_access`, excluding relay-only contracts. This solves the post-restart staleness without causing storms.

### Fast subscribe for locally-requested contracts
The subscribe optimization can target only locally-originated GETs. Whether via piggyback (with originator-only guard), direct subscribe to the known host, or some other mechanism, the key constraint is: **only the node whose local user requested the contract should subscribe, never relay nodes.**

### Correct `is_hosted` semantics
The GET handler's `is_hosted` shortcut could be refined to only trust `is_hosted` when `local_client_access` is true, avoiding serving stale relay-cached state after restart.

## Design Considerations

[...]

PR #4133, opened 2026-05-16
fix(put): mark originator PUT as locally-accessed for renewal

sanity:
## Problem

The hosting cache's `local_client_access` flag is what gates a contract into `contracts_needing_renewal` (`crates/core/src/ring/hosting.rs:971-991`). Without it, the contract gets excluded from renewal, the subscription expires, the entry eventually evicts under byte-budget pressure, and the next cold remote GET fails the `is_locally_hosted` shortcut and routes to the network — where, for a contract no other peer subscribes to, the GetOp hangs until the WS client's 180s timeout.

The GET path has called `mark_local_client_access` since #3769 (`client_events::serve_get`), but **the PUT path never did**. So a contract PUT'd by a local client and never GET'd locally — the freenet-git mirror demo's exact pattern (bot pushes daily, no GETs until the next day's cron) — accumulated this exact failure mode.

Confirmed via nova gateway log analysis: three independent 180s timeouts on 2026-05-14 against `freenet:96rknpy1GYhZ/freenet-stdlib` (07:24, 13:27, 19:00 UTC). On May 15 the same contract served fine via the local-cache shortcut, indicating the LRU regained the entry between failures (likely from a fresh push warming `record_access` again, but without the flag set the cycle was bound to repeat).

## Solution

Single line addition in `finalize_put_at_originator` (`crates/core/src/operations/put.rs:36`) — the originator-only finalization helper that fires after a PUT successfully completes:

```rust
op_manager.ring.mark_local_client_access(&key);
```

At this point the cache entry exists (the put pipeline already ran `put_contract` → `host_contract` via the shared `relay_put_store_locally` path), so the mark sets the flag on the existing entry. Relay-side PUTs bypass this helper, so they correctly do not get the flag — matching the relay-helper's existing doc-comment in `op_ctx_task::relay_put_store_locally`:

> This helper is **relay-only** — it never sets `mark_local_client_access` (that's originator-side).

This was the symmetric call site that was missing until this commit. No cache-layer changes, no `is_locally_hosted` gate changes; the freshness semantics established by #3340 are preserved.

## Why not the previous approach (#4132 closed)

#4132 attempted to make the access flag survive main-cache eviction via a separate history LRU + dropped the `is_hosting_contract` clause from the gate. Skeptical + Codex review caught two HIGH-severity issues: (a) the new gate bypassed the freshness guard from PR #3340 for relay-cached evicted contracts, and (b) the shortcut returned without re-adding the contract to the hosting cache, leaving renewal bookkeeping permanently broken. Both surfaced the deeper design problem — PUT was never marking — which this PR fixes directly without touching the cache layer or freshness gate.

## Testing

`finalize_put_at_originator_marks_local_client_access` in `crates/core/src/operations/put.rs::tests` — structural source-scrape that asserts the call stays in place. Mirrors the existing `finalize_put_at_originator_never_subscribes_from_driver` pin-test pattern in `op_ctx_task.rs`. Catches accidental removal of the call.

The behavioral invariant ("flag enables renewal") is already covered by `test_local_client_access_enables_renewal` at `hosting.rs:2601`. No new behavioral test needed since the cache and renewal layers are unchanged.

All 43 `operations::put` + 62 `ring::hosting` tests pass. `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

## Fixes

No GitHub issue filed — discovered while investigating recurring rescue-demos Matrix failure alerts for the freenet-stdlib mirror.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/put.rs
+++ b/crates/core/src/operations/put.rs
@@ -53,6 +53,25 @@ pub(super) async fn finalize_put_at_originator(
         op_manager.ring.register_events(Either::Left(event)).await;
     }
 
+    // Mark the contract as locally-accessed now that the originator's PUT
+    // has succeeded and the local cache entry exists (created by the
+    // shared `relay_put_store_locally` path during `put_contract` +
+    // `host_contract`). Without this, self-hosted contracts that were
+    // PUT'd by a local client but never GET'd would never get the
+    // `local_client_access` flag — the GET path is the only other call
+    // site for `mark_local_client_access`. Missing the flag excludes the
+    // contract from `contracts_needing_renewal`, the subscription expires,
+    // the entry eventually gets evicted under byte-budget pressure, and
+    // the next cold remote GET fails the `is_locally_hosted` shortcut and
+    // routes to the network — where, for a contract no other peer is
+    // subscribed to, the GetOp hangs until the WS client times out.
+    // (freenet-stdlib mirror demo, 2026-05-14: 180s timeouts on
+    // freenet:96rknpy1GYhZ/freenet-stdlib for exactly this reason.)
+    //
+    // This is the originator-side mark the relay-helper doc-comment in
+    // `op_ctx_task::relay_put_store_locally` was waiting for.
+    op_manager.ring.mark_local_client_access(&key);
+
     if subscribe {
         start_subscription_after_put(op_manager, id, key, blocking_subscribe).await;
     }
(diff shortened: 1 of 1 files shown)


sanity:
Multi-model review consolidated findings:

**Resolved**:
- ✅ Codex P2 #1 / skeptical H1 (structural pin test false-passes from doc comments) — fixed in b100df32 by stripping comments and asserting on exact call syntax `ring.mark_local_client_access(&key)`. Verified the new test FAILS when call is removed, PASSES when restored.
- ✅ Skeptical L3 (stale relay-helper docstring reference) — fixed in c5ed7179.

**Verified safe**:
- ✅ Call-site coverage: `finalize_put_at_originator` is the sole originator-side success path; reached for both `ReplyClass::Stored` and `ReplyClass::LocalCompletion`; streaming client PUTs flow through same `start_client_put` entry (per skeptical's trace at op_ctx_task.rs:723-735).
- ✅ Renewal storms: mark gated on originator role; cannot be triggered by remote-peer-relayed PUTs. Equivalent to existing local-client GET vector; no new amplification.
- ✅ Eviction race: `cache.rs::record_access` TTL gate (`min_ttl`) protects the freshly-inserted entry from parallel-PUT eviction before the originator's bypass fires.

**Accepted with rationale (M2)**:
- Skeptical asked for a `#[freenet_test]` behavioral test that PUTs a contract and asserts `has_local_client_access` returns true post-completion, to catch future refactors that move `host_contract` AFTER `finalize_put_at_originator` (which would silently make the mark a no-op). After discussion: the downstream invariant ("flag enables renewal") is already covered by `test_local_client_access_enables_renewal` at hosting.rs:2601, and the structural pin test now catches direct removal. The narrower refactor-induced-no-op class is real but lower probability; deferring to a separate PR if it occurs.

CI green.

[AI-assisted - Claude]

Issue #4137, opened 2026-05-16
tracking: harden delegate secrets-at-rest (KEK, per-write nonce, no default cipher)


iduartgomez:
## Summary

Delegate secrets-at-rest is broken in several independent ways. This issue tracks the umbrella; sub-issues break it into shippable units.

## Current state (read of `crates/core/src/wasm_runtime/secrets_store.rs` + `crates/core/src/config/secret.rs` on `main`, post 0.2.58)

- AEAD: XChaCha20-Poly1305.
- Key + nonce live in `Secrets { cipher: [u8;32], nonce: [u8;24] }`, loaded from plaintext files pointed at by `--cipher` / `--nonce` (or `CIPHER` / `NONCE` env), persisted via `Config`.
- If unset, both fall back to **`DelegateRequest::DEFAULT_CIPHER` and `DEFAULT_NONCE`** — hardcoded constants in `freenet-stdlib/rust/src/client_api/client_events.rs:584-592`, public in source.
- `SecretsStore::store_secret` uses **one nonce** per delegate for **every secret and every overwrite/snapshot** (see `secrets_store.rs:178-187`). XChaCha20-Poly1305 nonce-reuse breaks both confidentiality and integrity.
- Per-delegate `(cipher, nonce)` supplied by clients via `RegisterDelegate` is held in an in-memory `HashMap<DelegateKey, Encryption>` only — **not persisted**. Restart → fallback to default cipher → silent corruption / "missing secret".
- `transport_keypair` auto-persists plaintext to `secrets_dir/transport_keypair` (`config/secret.rs:61-91`). No auto-generation for `cipher`/`nonce`.
- No file permission hardening; no zeroize on plaintext buffers.

## Threat-model gaps

1. **Default-config nodes have no encryption.** Any operator that doesn't manually create cipher/nonce files runs with a world-known key. Filesystem-read attacker = plaintext recovery.
2. **Per-delegate nonce reuse.** Even with a properly-random custom cipher, every secret under a delegate shares one nonce → AEAD broken.
3. **Cipher/nonce/keypair files = plaintext at rest** with default umask permissions. No OS keyring, no KEK tier.
4. **In-memory delegate ciphers are lost on restart** because there is no persistence path for them.
5. **No memory hygiene.** Plaintext `Vec<u8>` flows through `store_secret` / `get_secret` without `Zeroize`.

## Related

- #1154 — Cross-platform sandboxing (filesystem isolation, complement not substitute)
[...]

PR #4143, opened 2026-05-16
fix(secrets): per-write random nonce for delegate secrets-at-rest

### LINKED ISSUE #4139: fix(secrets): per-write random nonce + remove default-cipher fallback
Part of #4137.

## Problem

`SecretsStore::store_secret` (crates/core/src/wasm_runtime/secrets_store.rs:178-187) encrypts every secret under a delegate with the **same** `(cipher, nonce)` pair:

```rust
let encryption = self.ciphers.get(delegate).unwrap_or(&self.default_encryption);
let ciphertext = encryption.cipher.encrypt(&encryption.nonce, plaintext.as_ref())?;
```

`encryption.nonce` is the per-delegate registration nonce, reused for:

- every distinct `SecretsId` written under that delegate, and
- every overwrite of the same `SecretsId` (snapshots are byte-identical hard-links of prior ciphertexts so they share the nonce too).

XChaCha20-Poly1305 nonce reuse with the same key is catastrophic: it breaks both confidentiality (keystream XOR recovery between any two messages) and integrity (Poly1305 key recovery from two tags).

Additionally, when no cipher/nonce files are configured, `Secrets::default` (config/secret.rs:23-32, 95-105) falls back to `DelegateRequest::DEFAULT_CIPHER` and `DEFAULT_NONCE` — hardcoded constants in `freenet-stdlib/rust/src/client_api/client_events.rs:584-592`, public in source. Default-configured nodes therefore run with a world-known key and a fixed nonce.

## Fix

1. Generate a fresh 24-byte nonce per call to `store_secret`. Prepend the nonce to the ciphertext blob on disk:

   ```
   on-disk layout (per active file and per snapshot):
     [24-byte nonce][AEAD ciphertext + 16-byte tag]
   ```

   `get_secret` slices off the leading 24 bytes and feeds them to `decrypt`. `restore_snapshot` is byte-copy and stays unchanged.

   Use `XChaCha20Poly1305::generate_nonce(&mut OsRng)` (or `GlobalRng` if a DST seam is required — verify which abstraction the wasm_runtime path is allowed to use; this is the AEAD random-input boundary, not application logic).

2. Remove `DelegateRequest::DEFAULT_CIPHER` / `DEFAULT_NONCE` from `freenet-stdlib` (stdlib-first release per `.claude/rules/git-workflow.md`). Replace the default fallback in `SecretArgs::build` with auto-generation of a per-node cipher persisted to `secrets_dir/delegate_cipher` (mirrors the existing `transport_keypair` auto-persist at config/secret.rs:67-87). Drop `nonce_path` / `--nonce` entirely — nonces are per-write, not configured.

3. Migration: detect on-disk files written under the old layout (no nonce prefix) by length / a one-byte version marker; either re-encrypt with the new scheme on first load or refuse to start with a clear "run `fdev migrate-secrets`" error. Pick whichever the team prefers; flag in PR.

## Regression tests

- Two writes of the same `(delegate, SecretsId)` with the same plaintext produce **different** on-disk bytes.
- Two writes of different `SecretsId`s under the same delegate produce on-disk bytes with different nonce prefixes.
- A node started with no cipher/nonce config auto-generates a cipher file, and a second start reuses it.
- Building with `--cipher /path/that/does/not/exist` is an error, not a silent fallback to the old default.

## Out of scope

- Persisting per-delegate ciphers across restart (#XXXX, separate sub-issue).
- OS keyring KEK (#XXXX, separate sub-issue).
- Sandboxing (#1154).

## Severity

Treat as security-class. The combination of "default key in source" + "nonce reuse" means every default-configured node's delegate secrets are recoverable by anyone who can read the secrets directory; on a node with a custom cipher, all secrets under one delegate are still cross-decryptable.

iduartgomez:
## Problem

`SecretsStore::store_secret` (crates/core/src/wasm_runtime/secrets_store.rs) encrypted every secret under a delegate with the **same** `(cipher, nonce)` pair:

```rust
let encryption = self.ciphers.get(delegate).unwrap_or(&self.default_encryption);
let ciphertext = encryption.cipher.encrypt(&encryption.nonce, plaintext.as_ref())?;
```

The same per-delegate registration nonce was used for:

- every distinct `SecretsId` written under that delegate, and
- every overwrite of the same `SecretsId` (snapshots are byte-identical hard-links of prior ciphertexts so they share the nonce too).

XChaCha20-Poly1305 nonce reuse with the same key is catastrophic: any two ciphertexts let an attacker recover the keystream via XOR, and any two authentication tags let them recover the Poly1305 one-time key.

Additionally, when no cipher/nonce files were configured, `Secrets::default` fell back to `DelegateRequest::DEFAULT_CIPHER` / `DEFAULT_NONCE` — hardcoded constants in `freenet-stdlib`, public in source. Default-configured nodes therefore ran with a world-known key and a fixed nonce, and `register_delegate` skipped registration entirely when the caller-supplied nonce matched the default — so even nodes that customized the cipher still routed default-nonce writes through the world-known fallback.

## Solution

**Per-write random nonce + version-prefixed on-disk format.**

- `store_secret` generates a fresh 24-byte `XNonce` per call via `XChaCha20Poly1305::generate_nonce(&mut OsRng)` and persists it inline.
- New on-disk layout: `[0x01 version][24-byte nonce][AEAD ciphertext + 16-byte tag]`.
- `get_secret` parses the versioned format; on AEAD failure or missing/mismatched header it falls back to legacy decryption with the per-delegate registration nonce. This lets nodes upgrade in place — legacy blobs remain readable until they are next overwritten, at which point the active path is rewritten in the new format. No separate one-shot migration tool needed.
- `Encryption` retains the registration nonce as `legacy_nonce` for the read-side fallback only; encryption never reads it.
- `register_delegate` no longer treats the default-nonce sentinel specially. The cipher is the only secret left under per-write nonces, so it must always be registered per delegate.

### Why not also remove `DEFAULT_CIPHER` / `DEFAULT_NONCE` from `freenet-stdlib` in this PR?

Stdlib-first release policy (`.claude/rules/git-workflow.md`): public-API removals from `freenet-stdlib` must merge + publish to crates.io before the consuming freenet-core change ships. That part of #4139 — deleting the constants and replacing the `SecretArgs::build` fallback with per-node auto-generated cipher persistence — is queued for a follow-up stdlib PR.

This PR closes the catastrophic nonce-reuse half of #4139 immediately and is safe to ship without the stdlib change: the existing `DEFAULT_CIPHER` constant is still used to seed default-configured nodes' cipher, but the per-write nonce means even default-configured nodes no longer suffer cross-secret keystream/tag recovery within a delegate.

## Testing

`cargo test -p freenet --features "trace,websocket,redb,wasmtime-backend" --lib` — 2497 passed, 14 ignored (no new ignores).

Targeted regressions in `crates/core/src/wasm_runtime/secrets_store.rs`:

- `per_write_nonce_makes_identical_plaintext_ciphertext_distinct` — two writes of the same plaintext under the same `(delegate, SecretsId)` produce different on-disk bytes.
- `store_secret_writes_version_header` — every new blob starts with `VERSION_V1` and is at least `HEADER_LEN + 16` bytes (header + AEAD tag).
- `legacy_format_blob_is_decryptable` — a hand-crafted pre-versioned blob (raw AEAD with the registration nonce) decrypts through the fallback path.
- `corrupt_versioned_blob_errors_cleanly` — versioned header with bogus AEAD surfaces `SecretStoreError::Encryption`, not a silent success or a misleading `MissingSecret`.

Pre-existing tests `restore_snapshot_prefers_unsuffixed_collision` and `second_write_snapshots_prior_value` were updated to use the new on-disk layout (hand-crafted snapshot blobs now carry the version header) and continue to pass.

`cargo fmt --all --check` clean, `cargo clippy -p freenet --features "trace,websocket,redb,wasmtime-backend" --all-targets -- -D warnings` clean.

## Fixes

Partially closes #4139 (per-write nonce half). The remaining stdlib-first work (delete `DEFAULT_CIPHER` / `DEFAULT_NONCE` constants from `freenet-stdlib`, add per-node cipher auto-persistence mirroring `transport_keypair`) ships in a follow-up after the stdlib PR publishes.

Refs: #4137 (tracker).

### DIFF
--- a/crates/core/src/wasm_runtime/secrets_store.rs
+++ b/crates/core/src/wasm_runtime/secrets_store.rs
@@ -1402,11 +1402,15 @@ mod test {
     }
 
     /// On-disk blob written by `store_secret` MUST begin with the
-    /// `VERSION_V1` header byte. This is the discriminator the read path
-    /// uses to tell new files from legacy files; if the writer ever stops
-    /// emitting it, the read path will silently fall back to legacy decrypt
-    /// (which would fail because there is no shared registered nonce in
-    /// the new model).
+    /// `VERSION_V1` header byte and carry a fresh random nonce in
+    /// bytes [1..25]. The version byte is the discriminator the read
+    /// path uses to tell new files from legacy files; if the writer
+    /// ever stops emitting it, the read path will silently fall back
+    /// to legacy decrypt (which would fail because there is no shared
+    /// registered nonce in the new model). The nonce-randomness check
+    /// catches a regression where someone hardcoded the nonce (e.g. to
+    /// zeros for "debugging") — `assert_ne!(first, second)` over whole
+    /// blobs would miss that if the ciphertext also varies.
     #[tokio::test]
     async fn store_secret_writes_version_header() -> Result<(), Box<dyn std::error::Error>> {
         let temp_dir = tempfile::tempdir()?;
@@ -1436,6 +1440,24 @@ mod test {
             blob.len() >= HEADER_LEN + 16,
             "blob too short: {} bytes",
             blob.len()
+        );
+
+        // Write a second secret and assert the nonce field differs. The
+        // nonce field is the [1..HEADER_LEN] slice. A regression that
+        // hardcoded the nonce to a constant (zeros, or anything else)
+        // would leave this slice identical across writes; whole-blob
+        // inequality alone could be satisfied by varying ciphertext.
+        let secret_id_2 = SecretsId::new(vec![84]);
+        store.store_secret(delegate.key(), &secret_id_2, b"hello".to_vec())?;
+        let blob_2 = std::fs::read(
+            secrets_dir
+                .join(delegate.key().encode())
+                .join(secret_id_2.encode()),
+        )?;
+        assert_ne!(
+            &blob[1..HEADER_LEN],
+            &blob_2[1..HEADER_LEN],
+            "nonce field must be random per write"
         );
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


iduartgomez:
Addressed both findings from the rule-review (comment 4467444088):

**Warning fixed** — new test `legacy_blob_with_version_byte_falls_through_to_legacy_decrypt` covers the documented 1/256 ambiguity in `decrypt_secret_blob`. AEAD output is deterministic in `(key, nonce, plaintext)`, so iterating attempts with random suffixes never varies `aead[0]`; the test instead exploits the bijection `aead[0] = plaintext[0] XOR keystream[0]` and walks `plaintext[0]` over `0..=255` — exactly one byte yields `aead[0] == VERSION_V1`. Pins that the new-format AEAD parse fails closed and the legacy-decrypt fallback succeeds and returns the original plaintext.

**Info fixed** — dropped the misleading `thread::sleep(5ms)` in `per_write_nonce_makes_identical_plaintext_ciphertext_distinct` (nonce uniqueness comes from `OsRng`, not wall-clock). Also tightened the assertion to compare the nonce slice explicitly so a regression that hardcoded the nonce would fail here directly instead of relying only on whole-blob inequality.

19 tests pass in `wasm_runtime::secrets_store`; `cargo fmt`/`clippy` clean.

iduartgomez:
Addressed remaining subagent review concerns:

- **Suggestion #4** — `store_secret_writes_version_header` strengthened: writes two secrets, asserts nonce field (bytes [1..HEADER_LEN]) differs. Catches a regression where someone hardcoded nonce to zeros for debugging.
- **Q1** — `legacy_snapshot_survives_restore_and_get_secret`: hand-crafts a legacy-format snapshot, calls `restore_snapshot` (byte-copy, no re-encryption), then `get_secret` recovers plaintext through fallback. Pins that snapshot history written by older freenet-core versions remains usable after upgrade without migration tool.
- **Q2** — `register_with_default_cipher_decrypts_legacy_default_blob`: pins behavioral equivalence between the removed `skip-on-default-nonce` branch and the new always-register code path for default-configured nodes upgrading. Guards against a future legacy-fallback change silently breaking reads of pre-existing secrets.

Other subagent items intentionally left as-is:
- **#2** `MissingCipher` variant unreachable — kept; follow-up stdlib PR may need it again for "no cipher persisted yet" semantics.
- **#5** Stale `.tmp` files from old crashes — pre-existing, out of scope; worth a separate issue.
- **Q3** `Encryption: Clone` — confirmed no production clones; no action.

21/21 secrets_store tests pass; `cargo fmt`/`clippy` clean.

PR #4144, opened 2026-05-16
feat(secrets)!: auto-gen per-node delegate cipher; bump stdlib 0.8.0

### LINKED ISSUE #4139: fix(secrets): per-write random nonce + remove default-cipher fallback

iduartgomez:
Closes the second half of #4139. Depends on **freenet-stdlib 0.8.0** (published, see [freenet-stdlib#75](https://github.com/freenet/freenet-stdlib/pull/75)).

## Problem

PR #4143 fixed XChaCha20-Poly1305 nonce reuse for delegate secrets via per-write random nonces. But the **default cipher** was still a public constant — `DelegateRequest::DEFAULT_CIPHER` exported by freenet-stdlib. Any node started without an explicit \`--cipher\` flag encrypted under a world-known key. With nonce reuse fixed, the residual attack class (filesystem-only attacker recovers all default-configured nodes' secrets by reading the static stdlib source) remained.

## Solution

**Stdlib half** ([freenet-stdlib#75](https://github.com/freenet/freenet-stdlib/pull/75), already merged + published as 0.8.0): delete the two public constants. Wire format \`RegisterDelegate { cipher, nonce }\` unchanged.

**Core half** (this PR):

- \`SecretArgs::build\`: replace the \`DEFAULT_CIPHER\` fallback with **auto-generation + persistence**. First start (no \`--cipher\` flag, \`secrets_dir\` known): generate fresh 32-byte XChaCha20-Poly1305 key via \`OsRng\`, save to \`secrets_dir/delegate_cipher\` (0o600 on Unix). Subsequent starts reload the persisted key. Mirrors the existing \`transport_keypair\` auto-persist pattern.
- \`--nonce\` flag deprecated: warns at load and at build. The value is used only as a legacy-decrypt nonce for pre-#4143 on-disk files.
- \`Default for Secrets\` (test-only) returns a random cipher instead of the world-known constant.
- \`pub(crate) const LEGACY_DEFAULT_CIPHER\` / \`LEGACY_DEFAULT_NONCE\` embedded in core as read-side legacy fallback only, used by the \`Encryption::legacy_nonce\` path added in PR #4143.
- 10 client/test/fdev call sites updated to supply an arbitrary-but-fixed test cipher instead of naming the removed stdlib constants.
- Incidental stdlib 0.7→0.8 wire-format catch-up: \`NodeDiagnosticsResponse.contract_states\` map keys changed \`ContractKey -> String\` (freenet-stdlib#70), fixed in \`p2p_protoc.rs\` and \`report.rs\`.

## Upgrade compatibility

- **Existing on-disk delegate secret files** (pre-#4143, raw AEAD under \`DEFAULT_CIPHER\` + \`DEFAULT_NONCE\`) remain readable. \`ConfigArgs::read_secrets\` still seeds \`Secrets::cipher\` / \`Secrets::nonce\` with the historical defaults via the in-core \`LEGACY_*\` constants, which feed the legacy-decrypt fallback path in \`SecretsStore::decrypt_secret_blob\`.
- **First restart after this PR** triggers cipher auto-generation. New writes use the per-node cipher + per-write random nonce + version-prefixed on-disk format. Legacy files migrate in place to the new format on next overwrite, per the fallback-then-rewrite pattern established in #4143.
- **Wire format unchanged.** Old clients sending \`DEFAULT_CIPHER\` + \`DEFAULT_NONCE\` literals continue to work — supplied cipher is registered and used for encrypts; supplied nonce is held only for legacy-decrypt fallback.

## Testing

\`cargo test -p freenet --features "trace,websocket,redb,wasmtime-backend" --lib\` — 2502 passed, 14 ignored.
\`cargo fmt --all --check\` clean.
\`cargo clippy --workspace --all-targets --features "trace,websocket,redb,wasmtime-backend" -- -D warnings\` clean.

New regression tests in \`crates/core/src/config/secret.rs\`:
- \`test_load_default_generates_random_cipher\` — \`build(None)\` never produces \`LEGACY_DEFAULT_CIPHER\` and two calls yield distinct random ciphers.
- \`test_cipher_auto_persist_and_reload\` — file is created, contents reload deterministically, 0o600 perms verified on Unix.
- \`test_missing_cipher_path_is_hard_error\` — \`--cipher /missing\` returns \`NotFound\`, not silent fallback.

## Fixes

Closes #4139 (completes both halves).

Refs: #4137 (tracker), freenet-stdlib#75 (stdlib release).

### DIFF
--- a/.claude/rules/code-style.md
+++ b/.claude/rules/code-style.md
@@ -311,6 +311,39 @@ example (the `classify_suspend_jump` helper and its rustdoc). Do not
 add new call sites for any other purpose — every other "I need real
 time" urge in `crates/core/` should still go through `TimeSource`.
 
+#### Exception: `OsRng` for cryptographic key / nonce generation
+
+The other legitimate reason to bypass `GlobalRng` in `crates/core/` is
+cryptographic key material (AEAD keys, AEAD nonces, signing keys,
+ephemeral DH secrets, etc.). Cryptographic randomness MUST come from
+the OS entropy pool (`/dev/urandom`-equivalent), not from a
+deterministic simulation RNG: a `GlobalRng` seed leak would compromise
+every key derived from it, and tests that fix the seed for
+reproducibility would produce on-disk ciphertexts that any reader of
+the test source could decrypt.
+
+When you have this exact need:
+
+1. Use `chacha20poly1305::aead::OsRng` (or another `RngCore`-implementing
+   wrapper over the OS entropy pool — `rand::rngs::OsRng` works too).
+2. Restrict the call site to code that runs at node startup or
+   one-shot key generation paths — NEVER inside per-request hot loops
+   or anything the simulation harness would expect to control.
+3. Add a load-bearing comment on the helper function explaining *why*
+   `OsRng` is correct here (cite cryptographic-key-material reason —
+   a generic "we need randomness" justification is not sufficient).
+
+Canonical examples:
+
+- `crates/core/src/config/secret.rs::generate_cipher_key` — generates
+  the per-node XChaCha20-Poly1305 cipher seeded at first start.
+- `crates/core/src/wasm_runtime/secrets_store.rs::store_secret` —
+  generates a fresh per-write AEAD nonce.
+
+Do not add new `OsRng` call sites outside of crypto key/nonce
+generation. Every other "I need randomness" urge in `crates/core/`
+should still go through `GlobalRng`.
+
 ### WHEN writing documentation
 
 ```
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,7 +149,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -160,7 +160,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1543,7 +1543,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1774,7 +1774,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1851,7 +1851,7 @@ dependencies = [
  "ed25519-dalek",
  "either",
  "freenet",
- "freenet-stdlib 0.6.1",
+ "freenet-stdlib 0.8.0",
  "futures 0.3.32",
  "glob",
  "hex",
@@ -2051,7 +2051,7 @@ dependencies = [
  "flatbuffers 25.12.19",
  "flate2",
  "freenet-macros 0.1.0",
- "freenet-stdlib 0.6.1",
+ "freenet-stdlib 0.8.0",
  "freenet-test-network",
  "futures 0.3.32",
  "gag",
@@ -2167,7 +2167,7 @@ dependencies = [
  "clap",
  "freenet",
  "freenet-ping-types",
- "freenet-stdlib 0.6.1",
+ "freenet-stdlib 0.8.0",
  "freenet-test-network",
  "futures 0.3.32",
  "humantime",
@@ -2189,7 +2189,7 @@ name = "freenet-ping-contract"
 version = "0.1.11"
 dependencies = [
  "freenet-ping-types",
- "freenet-stdlib 0.6.1",
+ "freenet-stdlib 0.8.0",
  "serde_json",
 ]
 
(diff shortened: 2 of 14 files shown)


iduartgomez:
Addressed both blocking issues from the subagent review (commit pushed):

**B1 (migration regression)** — Added `legacy_migration_encryption: Option<Encryption>` to `SecretsStore`, always seeded with `(LEGACY_DEFAULT_CIPHER, LEGACY_DEFAULT_NONCE)` regardless of operator config. `decrypt_secret_blob` now tries three paths in order:
1. New `[VERSION_V1][nonce][AEAD]` with registered/default cipher
2. Legacy raw AEAD with registered cipher's `legacy_nonce`
3. Legacy raw AEAD with `LEGACY_DEFAULT_CIPHER` + `LEGACY_DEFAULT_NONCE`

Path 3 is the migration last-resort; logs at WARN when hit. Pin test `legacy_default_blob_decryptable_without_register_after_upgrade` exercises the exact regression: `Secrets::default()` (random cipher, asserted ≠ LEGACY_DEFAULT_CIPHER as precondition) + on-disk legacy blob + no `register_delegate` call → `get_secret` must succeed.

**B2 (file-creation race)** — Replaced `save_cipher` (write-then-chmod) with `save_cipher_new` using `OpenOptions::create_new(true).mode(0o600)` on Unix — atomic, no window where umask perms leak the key. Added caller-side `AlreadyExists` handler that reloads if a concurrent freenet process wins the create race (covers the TOCTOU between `exists()` and `open`).

**N3 (OsRng documentation)** — Added a doc-comment on `generate_cipher_key` explaining why `OsRng` is the documented exception to the `crates/core/` ban on `thread_rng()`/`random()` (crypto key material must come from OS entropy, runs at startup before TimeSource/GlobalRng scope).

22 secrets_store tests + 6 config::secret tests pass. 2503 lib tests total. fmt + clippy clean.

Non-blocking suggestions (N1, N2, N4-N9) noted; will fold into a follow-up or address in B3 PR if any escalate.

iduartgomez:
Addressed I1 (formalize OsRng exception in code-style.md, citing the two call sites) and I2 (doc-note non-reproducible `Default for Secrets` with security rationale). I3 (no test for AlreadyExists race-recovery arm) left as borderline — rule-review itself acknowledged FS mocking is required; can fold into follow-up if anyone hits it.

iduartgomez:
Unit & Integration failure on previous run was caused by a superseded test:

`test_native_serde_json_on_contract_states_fails_documents_root_cause` was pinning that native `serde_json::to_string` rejects `NodeDiagnosticsResponse.contract_states` keyed by `ContractKey` ("key must be a string" error). freenet-stdlib 0.8.0 changed the field to `HashMap<String, _>` (freenet-stdlib#70, bundled into this PR's stdlib bump), so native serde_json now succeeds and the `.expect_err` triggered the panic.

The original comment on the test predicted this exact scenario as the trigger for retiring the `diagnostics_to_json` helper. Per superseded-test convention in `.claude/rules/git-workflow.md`, marked `#[ignore]` with PR reference and kept as historical doc. The `diagnostics_to_json` helper itself is retained because its sibling tests exercise round-trip + field-coverage properties; a separate refactor can replace it with a direct call.

18 `commands::report::tests` pass + 1 ignored. Fresh CI cycle triggered.

Issue #4140, opened 2026-05-16
feat(secrets): OS keyring KEK with file-KEK fallback (HKDF per-delegate DEK)


iduartgomez:
Part of #4137.

## Problem

Even after #4139 (per-write nonce + auto-generated node cipher) and #4138 (persisted per-delegate ciphers), the **node cipher itself sits in plaintext** at `secrets_dir/delegate_cipher` (proposed in #4139) and `secrets_dir/transport_keypair` (existing, `config/secret.rs:67-87`). A read of `~/.local/share/freenet/secrets/` recovers everything.

This is the criticism #1154 references (Signal's chat-DB-at-rest critique applied to Freenet's delegate secrets) and the gap #4117 ran into operationally: delegate secrets are only as private as the filesystem they sit on.

## Proposed design

Introduce a **node KEK** stored outside the filesystem, with per-delegate DEKs derived from it.

### Storage backends (try in order, fall back on unavailable)

1. **OS keyring** via the [`keyring`](https://crates.io/crates/keyring) crate (linux Secret Service, macOS Keychain, Windows Credential Manager).
2. **Systemd credential** (`$CREDENTIALS_DIRECTORY/freenet-kek`) if running under a unit with `LoadCredentialEncrypted=`.
3. **File KEK** at `secrets_dir/node_kek` with `chmod 600` and a one-line warning logged at start. This is the "no better option" headless fallback.
4. **(future) hardware-backed** — Secure Enclave / TPM2 / Android StrongBox. Out of scope for this issue; tracked separately on the umbrella.

The backend is chosen once at first start, recorded in the node config, and only re-resolved on explicit operator action (so a transient keyring-daemon outage doesn't silently demote the node to file-KEK).

### Key hierarchy

```
node KEK (32 bytes, in keyring/cred/file)
  └── per-delegate DEK = HKDF-SHA256(KEK, salt = delegate_key.encode())
        └── per-secret nonce (24 bytes, random per write, from #4139)
```

- Per-delegate DEKs are never persisted. They are derived on demand from the KEK at `register_delegate` time and cached in the same in-memory map `SecretsStore::ciphers` already uses.
- The persisted per-delegate cipher table from #4138 becomes either (a) the KEK-wrapped DEK (if a client supplied its own cipher and we honor that), or (b) absent (if we derived deterministically from the KEK). Pick one — recommend (b) for simplicity, deprecate the client-supplied-cipher path on `RegisterDelegate` in a follow-up stdlib bump.

[...]

Issue #1154, opened 2024-07-10
Implement Cross-Platform Sandboxing for Secure Data Storage in Freenet


sanity:
To ensure Freenet's delegate mechanism securely stores private data and avoids [criticisms](https://x.com/mysk_co/status/1810664951796887617) similar to those faced by Signal, we need to implement a robust sandboxing solution. This solution should be cross-platform, requiring minimal development effort, and provide consistent protection across Linux, macOS, Windows, iOS, and Android.

#### Background
Recently, Signal received criticism for storing chat databases in an unprotected area of the file system, making them accessible by any user process. To prevent similar vulnerabilities in Freenet, we need to adopt a sandboxing strategy that ensures private data is stored securely and inaccessible to unauthorized processes.

#### Requirements
1. **Cross-Platform Support**: The solution should work uniformly across Linux, macOS, Windows, iOS, and Android.
2. **Minimal Development Effort**: The implementation should leverage existing tools and libraries to minimize development overhead.
3. **Strong Isolation**: Ensure that private data is sandboxed and inaccessible to unauthorized applications or processes.

#### Tasks
1. **Research Sandboxing Solutions**:
   - Investigate cross-platform sandboxing tools and libraries.
   - Evaluate the feasibility and security of each solution.

2. **Select a Sandboxing Solution**:
   - Choose a solution that meets the requirements for cross-platform support, minimal development effort, and strong isolation.
   - Document the rationale for the selected solution.

3. **Implement the Selected Solution**:
   - Develop and integrate the sandboxing solution into the Freenet application.
   - Ensure the solution works consistently across all supported platforms.

4. **Test and Validate**:
   - Conduct thorough testing to verify the sandboxing implementation.
   - Ensure private data is securely stored and inaccessible to unauthorized processes on all platforms.

5. **Document the Implementation**:
   - Provide detailed documentation on the sandboxing setup and configuration.
   - Include instructions for maintaining and updating the sandboxing solution.

#### Additional Information
By researching and selecting an appropriate sandboxing solution, we aim to enhance the security of Freenet and provide robust protection for private data stored within the delegate mechanism. This approach will help us maintain secure and best-practice data storage across all supported platforms.

#### References
- [Signal Criticism on Sandboxing](https://twitter.com/mysk_co/status/...)
- [Docker Documentation](https://docs.docker.com/)
- [Flatpak Documentation](https://docs.flatpak.org/en/latest/)
[...]

PR #4146, opened 2026-05-17
feat(secrets)!: os keyring KEK + HKDF per-delegate DEK; closes #4138 + #4140

### LINKED ISSUE #4138: feat(secrets): persist per-delegate ciphers across node restart
Part of #4137.

## Problem

`SecretsStore` holds per-delegate `(cipher, nonce)` only in memory:

```rust
ciphers: std::collections::HashMap<DelegateKey, Encryption>,
```

(`crates/core/src/wasm_runtime/secrets_store.rs:75`).

The map is populated by `register_delegate` when the client sends `RegisterDelegate { cipher, nonce }`. On node restart the map is empty, `get_secret` falls back to `self.default_encryption`, and either:

- decrypt fails (delegate had a real cipher) → `SecretStoreError::MissingCipher`, surfaced to the delegate as "secret missing"; or
- decrypt silently succeeds against the wrong key on the unlucky AEAD-tag-collision case (vanishingly rare but not bounded by any invariant).

A client app has no way to know the node was restarted; "secrets gone" is indistinguishable from "secrets never existed" to the delegate, which means every persisted-config app (display name, theme, account state) silently resets after every node restart. This compounds #4117's user-visible "delegate rebuild wipes secrets" failure mode — even without a rebuild, a process restart has the same effect unless the client happens to re-issue `RegisterDelegate` first.

## Fix

Persist registered per-delegate ciphers in the ReDb store next to the secrets index. Wrap them with the node-level cipher (the auto-generated one introduced in the no-default-cipher sub-issue) so a filesystem-only attacker still needs the node KEK to decrypt the per-delegate cipher entries.

Schema sketch (add to `crates/core/src/contract/storages.rs` alongside `store_secrets_index`):

```
DelegateKey -> WrappedCipherEntry {
    version: u8,
    wrap_nonce: [u8; 24],
    wrapped_cipher: Vec<u8>,   // node-KEK encrypted (cipher_key || delegate_nonce_seed)
}
```

On `register_delegate`: encrypt the supplied cipher with the node KEK + a fresh wrap nonce, write to ReDb, then update the in-memory map. On `SecretsStore::new`: read all entries, unwrap with node KEK, populate the in-memory map.

If the cipher already exists in ReDb and `register_delegate` is called with a different cipher, error rather than silently overwrite (otherwise a malicious or buggy client can lock the operator out of their own prior secrets).

## Regression tests

- Write secret → drop `SecretsStore` → recreate → `get_secret` returns the same plaintext (today this fails, this is the regression test).
- `register_delegate` with a different cipher for an already-registered delegate returns an error and leaves the stored cipher untouched.
- A ReDb row whose wrap nonce / KEK doesn't decrypt cleanly causes `SecretsStore::new` to return an error (not silently fall back to default).

## Depends on

- Per-write-nonce sub-issue (so the persisted per-delegate cipher entries themselves are protected with per-row nonces).

## Out of scope

- Replacing the cipher with a KEK→DEK hierarchy (#XXXX, OS keyring sub-issue).

iduartgomez:
## Problem

After PR #4143 (per-write nonce) and PR #4144 (per-node auto-generated `delegate_cipher`), one gap remains in the secrets-at-rest threat model: the auto-generated cipher sits **plaintext on disk** at `secrets_dir/delegate_cipher`. A filesystem-only attacker with read access to that directory recovers every delegate's secrets.

This is the criticism #1154 references (Signal's chat-DB-at-rest critique) and the operational gap #4138 calls out separately: even without a filesystem attack, the per-delegate cipher map lived only in `SecretsStore::ciphers` (in-memory) and was lost on restart, leaving every persisted-config app silently resetting on every node start.

## Solution

**One change closes both gaps.** Introduce a node KEK held in the strongest available platform secret store. Derive every per-delegate DEK from the KEK via HKDF-SHA256. Restart-readability falls out for free because the derivation is deterministic in `(KEK, delegate_key)`.

### Key hierarchy

```
node KEK (32 bytes, in OS keyring / systemd cred / file)
  └── per-delegate DEK = HKDF-SHA256(KEK, salt=delegate_key.encode(),
                                          info="freenet-delegate-dek-v1")
        └── per-write random nonce (24 bytes, OsRng — landed in #4143)
```

### Backend selection (post-review)

The auto-resolver tries TWO backends on first start (NOT keyring — see below):

1. **Systemd credential** — `$CREDENTIALS_DIRECTORY/freenet-kek` populated by `LoadCredentialEncrypted=` on the unit. Only enters the chain when `CREDENTIALS_DIRECTORY` is set (operator already opted in via the unit file).
2. **File** — `secrets_dir/node_kek`, 0o600, atomic-create. Always-available fallback. Emits WARN at first-start so operators see they are running the weakest backend.

The choice is persisted in `secrets_dir/kek_backend`. Subsequent starts load STRICTLY from the recorded backend — transient backend outage is a hard error, never a silent demotion.

**OS keyring is opt-in only**, not in the auto-chain. Touching the keyring on first node start would surface a Keychain / Credential Manager dialog the operator never asked for (they ran `freenet`, not a credential request), and on macOS every release auto-update changes the binary signature and re-prompts — making keyring an unworkable default for the binary's release model. Operators opt in via `freenet secrets kek-init --backend keyring --secrets-dir <path> --yes` BEFORE first start; the act of running the CLI is the consent capture. On Linux the keyring backend refuses to construct entirely because this workspace ships `keyring` without `linux-native`/`sync-secret-service` — the crate would otherwise silently fall back to an in-process mock store that orphans the KEK on process exit.

A future PR will let delegates trigger consent lazily via the WS API so end users can grant per-app and the keyring becomes the obvious default again without a startup surprise.

### `freenet secrets` CLI (moved from `fdev`)

- `kek-status   --secrets-dir <path>` — print active backend + `SHA-256(KEK)[..8]` fingerprint (KEK itself never leaked).
- `kek-init     --secrets-dir <path> --backend {keyring|systemd|file} --yes` — explicit opt-in BEFORE first start. Refuses when the marker already exists. `--backend systemd` is marker-only (the credential MUST be provisioned out-of-band by the service manager). Atomic with rollback: if marker write fails after `backend.store`, the orphaned KEK is scrubbed.
- `kek-migrate  --secrets-dir <path> --to {keyring|systemd|file} --yes` — atomic-swap migration with rollback discipline (marker rewrite AFTER target store succeeds; source delete LAST so a crash leaves at least one viable copy).
- `kek-rotate   --secrets-dir <path> --yes` — surfaced with an explicit "not yet implemented" error pointing operators at the temporary kek-migrate workflow. The crash-safe two-phase rewrite (`.rot` shadow files + recovery on next start + on-disk re-encryption walk for every secret + snapshot) is tracked as a follow-up under #4137.

### `RegisterDelegate` semantics

The wire-format `RegisterDelegate { cipher: [u8;32], nonce: [u8;24] }` is **unchanged** for compatibility with pre-#4140 clients. Server-side, both fields are now **ignored** and an INFO log is emitted on each registration. Accepting a client-supplied cipher would let any misbehaving client substitute a key the operator does not control, defeating the purpose of the node KEK.

## Backwards compatibility

`decrypt_secret_blob` extended to a 4-tier chain so every historical on-disk format remains readable on upgrade — each tier tries both versioned (`[VERSION_V1][nonce][AEAD]`) and legacy raw-AEAD formats:

1. Current registered/derived cipher.
2. `legacy_chain[..]` — currently a single entry, `default_encryption`, carrying the post-#4144 / pre-#4140 auto-persisted `delegate_cipher` value.
3. `legacy_migration_encryption` — `(LEGACY_DEFAULT_CIPHER, LEGACY_DEFAULT_NONCE)` for pre-#4143 default-configured nodes.

New regression tests pinning the full matrix (in `secrets_store::test`):

- `backcompat_raw_aead_under_legacy_default_cipher` (pre-#4143 default-config).
- `backcompat_versioned_blob_under_legacy_default_cipher` (#4143 with default constants).
- `backcompat_versioned_blob_under_post_4144_delegate_cipher` (#4144 with per-install random cipher).
- `backcompat_register_delegate_wire_still_works` (older clients sending RegisterDelegate fields).
- `restart_roundtrip_recovers_secret_without_re_registering` (closes #4138 directly).
- `derive_delegate_dek_deterministic_and_per_delegate` (HKDF invariants).

12 unit tests in `config::kek::tests` covering FileKek roundtrip, 0o600 perms, AlreadyExists refusal on second store, invalid length rejection, marker roundtrip + garbage rejection, second-start exclusive use of recorded backend, `replace_backend_marker` atomic overwrite + 0o600 + tmp cleanup, Linux keyring refusal.

5 unit tests in `commands::secrets_cmd::tests` covering `kek_init` (no-marker refusal, --yes gate, file-backend end-to-end, systemd marker-only path, bad backend string).

## Closes

- #4138 (persisted per-delegate ciphers across restart — solved automatically by HKDF derivation).
- #4140 (OS keyring KEK with file-KEK fallback + HKDF per-delegate DEK).

Refs: #4137 (tracker), #4143, #4144 (prior security-at-rest PRs in the sequence).

## Out of scope

- `kek-rotate` on-disk rewrite walk (crash-safe two-phase). Tracked as follow-up under #4137.
- Hardware-backed KEK (Secure Enclave / TPM2 / Android StrongBox). Tracked separately under #4137.
- Sandboxing the secrets dir (#1154).
- App-initiated lazy consent flow that re-enables keyring as default (next PR).

## Testing

`cargo fmt --all` clean. `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo build --workspace` clean. CI will exercise the full lib + integration suite.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -257,6 +257,67 @@ End-to-end procedure for any maintainer is in
 `gh workflow run release.yml --field version=X.Y.Z`. The `RELEASE_PAT`
 section above is the prerequisite for the cascade to fire automatically.
 
+## Delegate secrets-at-rest
+
+Server-side delegate secret encryption evolved across PRs #4143 / #4144
+/ #4146 (tracker #4137). The current model:
+
+  node KEK (32 bytes, in OS keyring / systemd cred / file)
+    └── per-delegate DEK = HKDF-SHA256(KEK, salt=delegate_key.encode(),
+                                            info="freenet-delegate-dek-v1")
+          └── per-write random nonce (24 bytes, OsRng — PR #4143)
+
+The KEK is provisioned on first start by the resolver in
+`crates/core/src/config/kek.rs`, which tries OS keyring → systemd
+credential → file backend in order and persists the choice in
+`secrets_dir/kek_backend`. Subsequent starts load STRICTLY from the
+recorded backend — transient keyring outage is a hard error, never a
+silent demotion.
+
+### Operator migration matrix
+
+| Source release           | Cipher source on disk                  | Status after PR #4146 |
+|--------------------------|----------------------------------------|------------------------|
+| ≤ 0.2.58 (pre-#4143)     | `LEGACY_DEFAULT_CIPHER` (world-known)  | Decryptable via legacy-migration tier. |
+| 0.2.59 (#4143)           | `LEGACY_DEFAULT_CIPHER`, versioned     | Decryptable via legacy-migration tier. |
+| 0.2.60+ (#4144)          | per-install random `delegate_cipher`   | Decryptable via `default_encryption` (legacy_chain[0]). |
+| post-#4146               | HKDF-derived DEK from node KEK         | Default path. |
+
+### BREAKING semantic change introduced by #4146
+
+`RegisterDelegate { cipher, nonce }` server-side semantics changed:
+client-supplied `cipher` and `nonce` are now IGNORED (logged at INFO).
+The wire format is unchanged so older clients continue to function,
+but **any prior deployment that wrote secrets under a client-supplied
+non-default cipher will lose access to those blobs** unless that
+cipher coincidentally matches the new HKDF-derived DEK (it won't).
+
+Deployments that only ever used `DelegateRequest::DEFAULT_CIPHER` or
+the per-install auto-generated cipher are covered by the legacy chain
+and migrate transparently.
+
+If you ran any client that explicitly constructed `RegisterDelegate {
+cipher: custom, .. }` against a freenet-core ≤ 0.2.60 node, restore
+the secrets dir from a pre-upgrade backup, decrypt with the custom
+cipher manually (the on-disk format is documented in
+`crates/core/src/wasm_runtime/secrets_store.rs::decrypt_secret_blob`),
+and re-upload via the post-#4146 client API. The HKDF derivation will
+produce a new DEK and the secrets will be re-encrypted under it on
+first write.
+
+### `fdev secrets` CLI
+
+```
+fdev secrets kek-status   --secrets-dir <path>
+fdev secrets kek-migrate  --secrets-dir <path> --to {keyring|systemd|file} --yes
+fdev secrets kek-rotate   --secrets-dir <path> --yes   # NOT YET IMPLEMENTED (#4137)
+```
+
+Node MUST be stopped before running migrate. `kek-rotate` currently
+bails with a clear error pointing operators at the temporary
+kek-migrate workflow; crash-safe two-phase rotation (`.rot` shadow
+files + recovery on next start) is tracked as a follow-up under #4137.
+
 ## External Resources
 
 - API docs: https://docs.rs/freenet
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -259,64 +259,8 @@ section above is the prerequisite for the cascade to fire automatically.
 
 ## Delegate secrets-at-rest
 
-Server-side delegate secret encryption evolved across PRs #4143 / #4144
-/ #4146 (tracker #4137). The current model:
-
-  node KEK (32 bytes, in OS keyring / systemd cred / file)
-    └── per-delegate DEK = HKDF-SHA256(KEK, salt=delegate_key.encode(),
-                                            info="freenet-delegate-dek-v1")
-          └── per-write random nonce (24 bytes, OsRng — PR #4143)
-
-The KEK is provisioned on first start by the resolver in
-`crates/core/src/config/kek.rs`, which tries OS keyring → systemd
-credential → file backend in order and persists the choice in
-`secrets_dir/kek_backend`. Subsequent starts load STRICTLY from the
-recorded backend — transient keyring outage is a hard error, never a
-silent demotion.
-
-### Operator migration matrix
-
-| Source release           | Cipher source on disk                  | Status after PR #4146 |
-|--------------------------|----------------------------------------|------------------------|
-| ≤ 0.2.58 (pre-#4143)     | `LEGACY_DEFAULT_CIPHER` (world-known)  | Decryptable via legacy-migration tier. |
-| 0.2.59 (#4143)           | `LEGACY_DEFAULT_CIPHER`, versioned     | Decryptable via legacy-migration tier. |
-| 0.2.60+ (#4144)          | per-install random `delegate_cipher`   | Decryptable via `default_encryption` (legacy_chain[0]). |
-| post-#4146               | HKDF-derived DEK from node KEK         | Default path. |
-
-### BREAKING semantic change introduced by #4146
-
-`RegisterDelegate { cipher, nonce }` server-side semantics changed:
-client-supplied `cipher` and `nonce` are now IGNORED (logged at INFO).
-The wire format is unchanged so older clients continue to function,
-but **any prior deployment that wrote secrets under a client-supplied
-non-default cipher will lose access to those blobs** unless that
-cipher coincidentally matches the new HKDF-derived DEK (it won't).
-
-Deployments that only ever used `DelegateRequest::DEFAULT_CIPHER` or
-the per-install auto-generated cipher are covered by the legacy chain
-and migrate transparently.
-
-If you ran any client that explicitly constructed `RegisterDelegate {
-cipher: custom, .. }` against a freenet-core ≤ 0.2.60 node, restore
-the secrets dir from a pre-upgrade backup, decrypt with the custom
-cipher manually (the on-disk format is documented in
-`crates/core/src/wasm_runtime/secrets_store.rs::decrypt_secret_blob`),
-and re-upload via the post-#4146 client API. The HKDF derivation will
-produce a new DEK and the secrets will be re-encrypted under it on
-first write.
-
-### `fdev secrets` CLI
-
-```
-fdev secrets kek-status   --secrets-dir <path>
-fdev secrets kek-migrate  --secrets-dir <path> --to {keyring|systemd|file} --yes
-fdev secrets kek-rotate   --secrets-dir <path> --yes   # NOT YET IMPLEMENTED (#4137)
-```
-
-Node MUST be stopped before running migrate. `kek-rotate` currently
-bails with a clear error pointing operators at the temporary
-kek-migrate workflow; crash-safe two-phase rotation (`.rot` shadow
-files + recovery on next start) is tracked as a follow-up under #4137.
+Operator-facing documentation (encryption model, migration matrix,
+`freenet secrets` CLI) lives in [`docs/secrets-at-rest.md`](docs/secrets-at-rest.md).
 
 ## External Resources
 
--- a/crates/core/src/wasm_runtime/secrets_store.rs
+++ b/crates/core/src/wasm_runtime/secrets_store.rs
@@ -758,36 +758,31 @@ fn decrypt_secret_blob(
     blob: &[u8],
     key: &SecretsId,
 ) -> Result<Vec<u8>, SecretStoreError> {
-    // Decryption strategy. The format and cipher have rotated three
-    // times across the secrets-at-rest hardening sequence; the chain
-    // below is tightly scoped so we attempt only the format that each
-    // historical writer actually produced.
+    // Decryption strategy. The format + cipher have rotated three
+    // times across the secrets-at-rest hardening sequence:
     //
     //   Tier 1 (`encryption`, the registered/derived DEK):
-    //     - try VERSION_V1 (`[0x01][nonce][AEAD]`) first
-    //     - fall back to raw-AEAD with `encryption.legacy_nonce` for
-    //       blobs written under this same cipher but before the
-    //       per-write-nonce format landed in #4143 (e.g. a delegate
-    //       whose cipher hasn't changed but whose oldest secret file
-    //       has not been overwritten since the upgrade).
+    //     - VERSION_V1 (`[0x01][nonce][AEAD]`) — today's writer
+    //     - raw-AEAD with `encryption.legacy_nonce` — same cipher,
+    //       pre-#4143 format (a delegate whose key hasn't rotated but
+    //       whose oldest secret hasn't been overwritten since upgrade).
     //
     //   Tier 2 (`legacy_chain[..]`, e.g. the post-#4144 / pre-#4140
     //   auto-persisted `delegate_cipher` carried by `default_encryption`):
-    //     - try VERSION_V1 ONLY. Every release that wrote under these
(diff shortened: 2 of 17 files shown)


iduartgomez:
Addressed all 5 blocking issues from the subagent review (commit `374b080b`):

**B1** — HKDF salt doc bug. `DelegateKey::encode()` returns bs58(`self.key`) only, not 64-byte `instance || code_hash`. Doc-comment on `derive_delegate_dek` rewritten to state the actual content + explain why `code_hash` is intentionally excluded (instance key already folds in wasm code via `blake3(params || wasm_code)`).

**B2** — Dead-code path in `decrypt_secret_blob`. Was trying both versioned and raw-AEAD formats against every chain entry. Rewrote so each tier only attempts the format its writer-era produced:
- Tier 1 (registered/derived): versioned + raw-AEAD (cipher unchanged but pre-#4143 format possible).
- Tier 2 (legacy_chain, post-#4144): VERSIONED ONLY (those releases were already at per-write nonce).
- Tier 3 (legacy_migration, LEGACY_DEFAULT_*): RAW-AEAD ONLY (pre-#4143 world-known-key blobs were never versioned).

Worst-case cold-read AEAD attempts drops from 8 to 3-4.

**B3** — `kek_backend.tmp` write hygiene. New `kek::replace_backend_marker` exposed via dev_tool reuses the same 0o600 + sync_all + atomic-rename discipline as `write_backend_marker`. `secrets_cmd::kek_migrate` now goes through it.

**B4** — TOCTOU panic in `kek_status`. Replaced `ensure_kek_loaded(unreachable!)` with direct `load_from_backend(kind, dir)` so a marker-removal race surfaces a clean error instead of panicking.

**B5** — BREAKING migration. New "Delegate secrets-at-rest" section in AGENTS.md documents the operator-visible migration matrix across #4143/#4144/#4146 and the explicit BREAKING semantic change: client-supplied `RegisterDelegate { cipher }` is now ignored, so any prior deployment writing under non-default ciphers loses access on upgrade. Recovery instructions included.

Non-blocking suggestions (N1–N10) noted; addressed in commit where cheap, deferred to follow-up where they require architectural changes (e.g., N1 hkdf crate doesn't zeroize HMAC state — out of our hands).

iduartgomez:
CI Clippy failed due to `libdbus-sys` build failure (the `keyring` crate's `sync-secret-service` feature pulls in dbus-secret-service → libdbus-sys, which needs `libdbus-1-dev` system package not installed on CI runners).

Pushed fix `a7efe76f` dropping `sync-secret-service` (+ now-unused `crypto-rust`) from the keyring feature set. The remaining backends still cover all our target environments without system deps:
- `apple-native` — macOS Keychain
- `windows-native` — Windows Credential Manager
- `linux-native-sync-persistent` — kernel keyutils (header-only; works on headless servers)

D-Bus Secret Service (GNOME/KDE keyring) is a desktop-environment service rarely present on production headless Linux installs; operators who want it can install libdbus-1-dev and re-enable in a downstream fork. The `KekBackend` abstraction is backend-agnostic so this is purely a dep-set decision.

iduartgomez:
Pushed `bf9104f7` addressing three CI / rule-review findings:

**1. Linux keyring dropped entirely.** `keyring 3.x` has NO persistent Linux option that doesn't pull `libdbus-sys` → `libdbus-1-dev`. Even `linux-native-sync-persistent` transitively requires `sync-secret-service` for persistence beyond process exit. Restricted features to `apple-native` + `windows-native` only. Linux nodes fall back through `SystemdCredentialKek` → `FileKek` (both header-only).

**2. `.unwrap()` in production code.** Rule-review warning fixed: `cipher_for` now uses `.expect(...)` carrying the load-bearing invariant + comment explaining why we can't use `Entry::or_insert_with`.

**3. Public KEK API doc gaps.** Added one-line summaries to `KekBackend::{kind, load, delete}`, `KeyringKek::new`, `SystemdCredentialKek::new`, `FileKek::new`.

`cargo clippy --locked --workspace --all-targets ... -- -D warnings` clean. Lock verified dbus-free (`grep -c dbus Cargo.lock` = 0).

iduartgomez:
Pushed `54847b12` — three fixes for the Unit & Integration failures:

**1. Restored tier-3 versioned decrypt path.** My B2 refactor dropped the versioned attempt for the LEGACY_DEFAULT_* tier, breaking #4143-era blobs (versioned format + world-known cipher). Restored it; tier 3 now attempts both versioned + raw-AEAD.

**2. Rewrote 3 legacy tests to use HKDF-derived DEK.** `legacy_format_blob_is_decryptable`, `legacy_blob_with_version_byte_falls_through_to_legacy_decrypt`, `legacy_snapshot_survives_restore_and_get_secret` all used `fresh_cipher()` + `register_delegate` then encrypted under the supplied cipher. Post-#4146 `register_delegate` ignores client cipher (server uses HKDF-derived DEK). Each test now calls `store.derive_delegate_dek()` to get the actual cipher and encrypts under that.

**3. CREDENTIALS_DIRECTORY env-clear in file-backend-dependent tests.** GitHub Actions runners set this for some workflow types, which would make `SystemdCredentialKek` win the resolver. Added `unsafe { std::env::remove_var("CREDENTIALS_DIRECTORY") }` (with SAFETY comment) at the top of every test that depends on file backend being chosen. Safe under nextest's per-process isolation; not restored because the process exits at test end.

NAT Validation failure looks unrelated (Docker network connectivity timeout 120s) — likely pre-existing infrastructure flake. Will see if it reproduces on the new commit.

crates/core/src/wasm_runtime/secrets_store.rs
```diff
@@ -130,13 +155,36 @@ pub struct SecretsStore {
...
         })?;
 
+        // Load (or resolve + provision) the node KEK. First start picks
+        // a backend from the OS-keyring → systemd-credential → file
+        // chain; subsequent starts read the recorded backend marker and
+        // load strictly from there (no silent demotion). `OsRng` is the
+        // documented exception to the GlobalRng rule for cryptographic
+        // key material (see `.claude/rules/code-style.md`).
+        let (kek_backend, kek) = ensure_kek_loaded(&secrets_dir, || {
+            use chacha20poly1305::aead::OsRng;
```
iduartgomez, line 177: @claude doublecheck if persisted

AGENTS.md
```diff
@@ -257,6 +257,67 @@ End-to-end procedure for any maintainer is in
 `gh workflow run release.yml --field version=X.Y.Z`. The `RELEASE_PAT`
 section above is the prerequisite for the cascade to fire automatically.
 
+## Delegate secrets-at-rest
```
iduartgomez, line 260: this should be moved elsewhere

crates/fdev/src/secrets_cmd.rs
```diff
@@ -0,0 +1,232 @@
+//! `fdev secrets` subcommand: manage the node KEK backend.
```
iduartgomez: move to freenet

iduartgomez:
Addressed two review comments:

- [r3255694720](https://github.com/freenet/freenet-core/pull/4146#discussion_r3255694720): moved the secrets-at-rest section out of `AGENTS.md` (which is the agent-instruction index) into [`docs/secrets-at-rest.md`](https://github.com/freenet/freenet-core/blob/feat/secrets-kek-keyring/docs/secrets-at-rest.md). AGENTS.md now carries a single pointer.
- [r3255694961](https://github.com/freenet/freenet-core/pull/4146#discussion_r3255694961): moved the CLI from `fdev secrets ...` to `freenet secrets ...` (it's operator-facing, not dev-tooling). Module is now `crates/core/src/bin/commands/secrets_cmd.rs`; `fdev` lost the subcommand and its `sha2` dep.

Verified: `cargo build -p freenet --bin freenet` and `cargo build -p fdev` both pass, `cargo clippy -D warnings` clean on both, fmt clean, and `freenet secrets --help` renders the three subcommands.

Pushed as 62b649c3.

iduartgomez:
Addressed the review in commit \`(see latest push)\`. Summary:

**BLOCKERs (B1/B2/B3)** — all fixed:

- **B1** (Linux mock keyring): \`KeyringKek::new()\` refuses on Linux with a typed \`KekError::Keyring\` explaining the build-feature gap, BEFORE any marker write. Added a Linux-only test \`keyring_kek_new_refuses_on_linux\` pinning the refusal.
- **B2** (\`kek-init\` non-atomic): on \`write_backend_marker\` failure, the stored KEK is rolled back via \`backend.delete()\`; the surfaced error names the rollback outcome ("safe to re-run" vs "scrub manually").
- **B3** (systemd dead-end): \`kek-init --backend systemd\` is now a marker-only opt-in (no \`store\` attempted; the credential MUST be provisioned out-of-band via \`LoadCredentialEncrypted=\`).

**Observations** — addressed:

- **O1** (stale PR body): updated to describe the post-change auto-chain (systemd + file only), the opt-in keyring path including the Linux limitation, and the renamed \`freenet secrets\` binary.
- **O3** (\`kek-rotate\` \`--yes\` ordering): not-implemented bail moved to the top of \`kek_rotate\`, before the \`--yes\` check.
- **O4** (kek_init test coverage): 5 new unit tests in \`commands::secrets_cmd::tests\` (no-marker refusal, --yes gate, file-backend end-to-end, systemd marker-only path, bad backend string).
- **Warning #1 from the rule-review bot** (\`replace_backend_marker\` 0 tests): added \`replace_backend_marker_overwrites_existing_atomically\` (verifies overwrite, 0o600, no stray .tmp) + \`replace_backend_marker_creates_fresh_when_absent\`. Brings the kek.rs test count to 12.
- Struct-level \`///\` docs added to \`KeyringKek\` / \`SystemdCredentialKek\` / \`FileKek\`. Stale \`fdev secrets\` references in module docs flipped to \`freenet secrets\`.

**Deferred to follow-up** (not in this PR):

- **O2** (per-boot WARN for file backend) — small log-policy change; happy to do as part of the next pass if you want, just want to keep this PR focused.
- **O5** (CLI text mentioning Linux) — replaced. Reality check: Linux row now says "NOT supported in this build".
- **O6** (clap \`ValueEnum\` for backend kind) — cosmetic; the \`parse_backend_kind\` helper is shared between \`kek-init\` and \`kek-migrate\` and emits the exact \`possible values\` message. Happy to migrate to \`ValueEnum\` in a follow-up if you'd prefer.
- **O7** (dedicated \`pub mod secrets\` re-export module) — agree the surface is large; will plan as a separate refactor PR so the API churn isn't tangled with this functional change.
- **O8** (zeroize claim wrt keyring crate) — module-level doc already says "where possible"; I'll tighten the language with a follow-up audit. The caller-side \`Zeroizing\` discipline is intact.
- **O9** (\`!\` marker on the latest commit) — the PR-level breaking marker is already on the lead commit (\`feat(secrets)!:\`). Squash-merge users will see it.

CI re-running on the new push.

Issue #4163, opened 2026-05-19
Flaky: test_send_stream_without_bandwidth_limit fails under build load (50ms threshold too tight)


sanity:
## Symptom

`transport::peer_connection::outbound_stream::tests::test_send_stream_without_bandwidth_limit` fails intermittently with:

```
panicked at crates/core/src/transport/peer_connection/outbound_stream.rs:994:9:
Transfer took too long without rate limit: 66.672457ms
```

The test asserts the transfer completes in < 50ms (line 994-997). It passes reliably when run in isolation but fails when the full `cargo test -p freenet --lib` suite runs concurrently with build activity / under CPU pressure — observed running on a development workstation while clippy was finishing in parallel.

## Why this matters

Per the repo's "flaky tests are broken tests" rule, a wall-clock threshold this tight is testing the machine's spare CPU, not the rate-limiting code. The actual invariant the test wants to prove is *"the unlimited path is bounded above by the rate-limited path"* — that should be expressed as a ratio (or compared to a co-measured rate-limited run), not a hard 50ms cap.

## Suggested fix

Replace the absolute `< 50ms` assertion with one of:

1. **Relative**: run both the unlimited and the rate-limited variants, assert `unlimited * N < rate_limited` for some safety margin.
2. **Looser absolute** with a doc comment explaining the platform tolerance — e.g. `< 500ms` — but still test the *bandwidth_limit / no bandwidth_limit* delta rather than the absolute number.
3. **Mark as `#[ignore]` on `cfg(debug_assertions)`** — but this hides the bug rather than fixing it.

(1) is preferred — it pins what the test actually cares about.

## Reproduction

```bash
cargo test -p freenet --lib --message-format short
# Run the failing test in isolation — it passes:
cargo test -p freenet --lib test_send_stream_without_bandwidth_limit
```

Not blocking #4154; surfaced during the fix's full-suite verification on 2026-05-19.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:33:51.753943110+00:00
Issue hash: `d539a178`

## Value Assessment

[...]

PR #4164, opened 2026-05-19
fix(networking): wake parked op drivers on peer disconnect

### LINKED ISSUE #4154: GET initiated during peer-connection churn stalls silently for 60+s instead of fail-fast / re-route
## Symptom

User reported a GET against a low-popularity contract on a local node took 30+ seconds and the HTTP client timed out. Retry one second later succeeded in 18s. Telemetry shows the first GET actually completed network-side after **62 seconds** — it didn't fail, it just stalled.

## Evidence from telemetry (2026-05-19, contract `6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9`)

Originating peer `jMphpuy7z7PPG4Xt@73.98.109.226` (user's local node):

| Attempt | TX ID | Started (UTC) | Completed (UTC) | `elapsed_ms` |
|---|---|---|---|---|
| 1st (client timed out @30s) | `01KRZ6ADBAT9D0WZYAXC6SQSR2` | ~04:01:20.7 | 04:02:22.462 | **61716** |
| 2nd (retry, succeeded) | `01KRZ6BDQE7DG7W8V2BCDMX0G2` | 04:01:53.823 | 04:02:07.367 | 13544 |

Originating peer's events around the first GET:
- `04:01:19.557` — `disconnect` (peer connection lost)
- `04:01:20.7`  — first GET issued (derived from ULID + `elapsed_ms`)
- `04:01:32.919` — `connect_request_received` + `connect_rejected`
- `04:01:32.999` — `connect_request_sent`
- `04:01:33.321` — `connect_connected` (recovery)
- `04:01:53.823` — retry GET — routes cleanly via 3 hops, succeeds in 13.5s
- `04:02:22.462` — original first GET finally `get_success` (62s after issue)

**Notable:** the first transaction emitted **zero `get_request` events** in telemetry — only the eventual `get_success`. Compare with the retry, which emitted 9 `get_request` events (one per hop) within the first 400ms. The op was held in a pre-routing state while the originating node was reconnecting, and no peer-forwarding telemetry was emitted until the eventual response.

## Hypothesis

When a GET is initiated and the originating node has just lost peer connections (or otherwise lacks a viable forwarding target), the op gets parked in an internal queue waiting for routing to recover. It is not actively re-evaluated, not aggressively re-routed once new peers connect (~13s later in this case), and not failed-fast back to the caller. It just waits — in this case, 62 seconds.

The client-visible 30s timeout is just the symptom; the underlying issue is that the operation has no fail-fast or active-reroute semantics during peer-state churn.

## What I'd expect

Some combination of:
- **Fail-fast**: if a GET has zero viable forwarding targets at issue-time, return an error to the client immediately rather than parking it.
- **Active reroute**: when new peers connect while a GET is parked, re-evaluate routing and dispatch.
- **Bounded park time**: explicit timeout on \"waiting for routable peer\" with telemetry on what we were waiting for.

Today there's apparently no telemetry between op-creation and op-completion in this path — adding instrumentation for \"GET parked waiting for routing\" would already help diagnose this class of failures.

## Repro

Hard to reliably repro without inducing a disconnect at the right moment, but the conditions are:
1. A node loses one or more peer connections.
2. Within ~1s, a contract GET is issued via HTTP.
3. Result: GET hangs for tens of seconds until either peers recover and routing finally fires, or client times out.

## Cross-reference

- Original investigation: freenet/river debug session 2026-05-19
- Telemetry data: nova OTLP collector logs.jsonl, contract `6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9`, TX `01KRZ6ADBAT9D0WZYAXC6SQSR2`

[AI-assisted - Claude]

sanity:
## Problem

A GET op issued just before its downstream peer disconnects stalls for the full `OPERATION_TTL` (60 s) before retrying. The 2026-05-19 telemetry in #4154 captured the exact symptom: a user-visible 30 s HTTP timeout followed by a successful 18 s retry, with the original transaction completing **62 seconds** after issue and zero `get_request` telemetry events emitted along the way.

## Root cause

Two cooperating gaps in the disconnect-cancellation pipeline:

1. **`P2pBridge::handle_orphaned_transactions` was a logging-only stub.** It received every transaction associated with the pruned peer from `LiveTransactionTracker::prune_transactions_from_peer` and just logged the count. The doc-comment claimed "driver-owned cancellation," but no cancellation path existed — drivers parked in `OpCtx::send_and_await` were never told to give up.

2. **Forwards routed via `op_execution_sender` skipped `live_tx_tracker` registration.** `OpCtx::send_fire_and_forget` / `send_to_and_await` bypass `P2pBridge::send`, so the `(tx -> target_addr)` association `sending_transaction` normally records was never made. Even if (1) had been wired up, `prune_transactions_from_peer` would have returned empty for relayed GET / PUT / UPDATE / SUBSCRIBE ops.

The originator-loopback GET relay branch hits both: it fires the request to the downstream peer with `target_addr = Some(...)` and exits. The client driver's `send_and_await` waiter sits in `pending_op_results[client_tx]` waiting for a response that will never arrive (peer is gone) and is never woken.

## Approach

Wire the existing cancellation primitives end-to-end:

- **Register `(target_addr, *msg.id())` in `LiveTransactionTracker`** at `P2pConnManager::handle_op_execution` whenever the driver supplies an explicit target. Covers both `send_fire_and_forget` and `send_to_and_await` paths.
- **Make `P2pBridge::handle_orphaned_transactions` iterate orphans** and call `OpManager::try_release_pending_op_slot(tx)` — a new non-blocking helper that emits `NodeEvent::TransactionCompleted(tx)` via `try_send` (safe to call from the event loop).
- **Clean up `live_tx_tracker` in the `TransactionCompleted` event-loop arm**, so entries registered at `handle_op_execution` don't leak past op completion.

When the event loop processes `TransactionCompleted(tx)` it drops `pending_op_results[tx]`, closing the channel the driver is awaiting. `send_and_await` returns `Err(OpError::NotificationError)`; `drive_retry_loop` routes that to `advance()` and tries the next peer. Wake time goes from 60 s to sub-millisecond.

Why `try_send` rather than the existing `send().await` `release_pending_op_slot`: the orphan handler runs inside the event loop, and awaiting the same notification channel would risk deadlock. The non-blocking variant is best-effort — a momentarily-full channel falls back to the legacy `OPERATION_TTL` timeout, which is the pre-fix behavior. Mirrors the existing `notify_transaction_timeout` helper.

## Testing

Four new tests, all passing; the source-grep guard was verified to fail on regression by temporarily reverting `handle_orphaned_transactions` to a no-op stub.

- `try_release_pending_op_slot_emits_transaction_completed` — happy-path helper unit test.
- `try_release_pending_op_slot_handles_dropped_receiver` — robustness under shutdown races.
- `orphaned_transaction_wakes_parked_pending_op_results_waiter` — end-to-end pipeline: simulates the `registered-sender -> orphan-handler -> event-loop-drop -> driver-wake` sequence and asserts wake within 100 ms (vs. the pre-fix 60 s).
- `handle_orphaned_transactions_wakes_parked_drivers` — source-grep guard against the function reverting to logging-only.

Full `cargo test -p freenet --lib` is clean except for a single unrelated flaky timing test (`test_send_stream_without_bandwidth_limit`, 50 ms wall-clock threshold) which passes in isolation — filed separately as #4163.

## Why CI didn't catch this earlier

The pre-fix code path required a precise sequence: GET op forwarded just before peer disconnect, with the originator's loopback running on a node that lacked a replacement peer for the full `OPERATION_TTL`. Simulation tests with stable peer sets and no induced churn never hit it. The source-grep guard test prevents future refactors from silently re-stubbing the orphan handler; the end-to-end pipeline test now exercises the wake path directly.

Closes #4154

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -3859,19 +3859,30 @@ impl P2pConnManager {
                     Some(target_addr) => {
                         // Track `tx → target_addr` so disconnect-cancellation in
                         // `prune_connection` / `handle_orphaned_transactions` can
-                        // wake the parked driver (#4154). The `P2pBridge::send`
-                        // path registers via `sending_transaction`; this branch
-                        // covers the `op_execution_sender` path
-                        // (`send_to_and_await` / `send_fire_and_forget`), which
-                        // bypasses `P2pBridge::send`. Cleanup happens in the
-                        // `TransactionCompleted` handler; duplicate entries from
-                        // CONNECT's `Ring::initiate_connect` registration are
-                        // tolerated by `remove_finished_transaction`.
-                        self.bridge
-                            .op_manager
-                            .ring
-                            .live_tx_tracker
-                            .add_transaction(target_addr, *msg.id());
+                        // wake the parked driver (#4154). Two cases register here:
+                        //   - `send_to_and_await`: this branch just installed a
+                        //     waiter into `pending_op_results` above.
+                        //   - `send_fire_and_forget` from a relay driver running
+                        //     on the originator: the callback is closed (no insert
+                        //     above) but the originator's client driver has an
+                        //     earlier `pending_op_results[tx]` waiter that needs
+                        //     waking.
+                        // Skip when no waiter is present (e.g. cancelled-driver
+                        // teardown races) — registering with nothing to wake would
+                        // leak a live_tx_tracker entry until the peer disconnects.
+                        // `add_transaction` is idempotent + rebind-safe, so the
+                        // CONNECT ride-along (also registered at
+                        // `Ring::initiate_connect`) does not double-count, and a
+                        // relay that response-sends back to its upstream after
+                        // forwarding downstream cleanly migrates the entry.
+                        let tx = *msg.id();
+                        if state.pending_op_results.contains_key(&tx) {
+                            self.bridge
+                                .op_manager
+                                .ring
+                                .live_tx_tracker
+                                .add_transaction(target_addr, tx);
+                        }
                         EventResult::Event(
                             ConnEvent::OutboundMessageWithTarget { target_addr, msg }.into(),
                         )
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -175,11 +175,20 @@ impl P2pBridge {
         }
     }
 
-    /// Log transactions orphaned by a pruned peer connection.
+    /// Wake any drivers whose downstream peer just disconnected.
     ///
-    /// Drivers own retry/cancellation; the orphan list is informational
-    /// only — used by callers to confirm prune occurred. No retry is
-    /// dispatched here.
+    /// For each orphaned `tx` we emit
+    /// `NodeEvent::TransactionCompleted(tx)` via `try_send`. The event-
+    /// loop handler at `NodeEvent::TransactionCompleted` drops the
+    /// matching `pending_op_results` sender, which closes the channel
+    /// the driver is awaiting in `OpCtx::send_and_await`. The driver
+    /// then sees `Err(OpError::NotificationError)`; the shared retry
+    /// loop (`drive_retry_loop`) routes that to `advance()` and tries
+    /// the next peer.
+    ///
+    /// Without this hop, a GET (or other relayed op) issued just before
+    /// its downstream peer disconnected would block for the full
+    /// `OPERATION_TTL` (60 s) before retrying — see #4154.
     pub(crate) async fn handle_orphaned_transactions(
         &self,
         transactions: Vec<Transaction>,
@@ -190,8 +199,11 @@ impl P2pBridge {
         }
         tracing::debug!(
             count = transactions.len(),
-            "Orphaned transactions from pruned connection (driver-owned cancellation)"
+            "Orphaned transactions from pruned connection — waking parked drivers"
         );
+        for tx in transactions {
+            self.op_manager.try_release_pending_op_slot(tx);
+        }
     }
 
     /// Send stream data with an explicit completion signal.
@@ -1968,6 +1980,18 @@ impl P2pConnManager {
                                 if state.pending_op_results.remove(&tx).is_some() {
                                     crate::config::GlobalTestMetrics::record_pending_op_remove();
                                 }
+                                // Clean up live_tx_tracker too. `handle_op_execution`
+                                // registers `(tx → target_addr)` for messages with an
+                                // explicit target (#4154) so that disconnect-cancellation
+                                // can find the tx. Without this cleanup those entries
+                                // would leak past op completion. Idempotent with the
+                                // explicit `op_manager.completed(client_tx)` that
+                                // drivers call on terminal exit.
+                                ctx.bridge
+                                    .op_manager
+                                    .ring
+                                    .live_tx_tracker
+                                    .remove_finished_transaction(tx);
                             }
                             NodeEvent::LocalSubscribeComplete {
                                 tx,
@@ -3833,9 +3857,34 @@ impl P2pConnManager {
                 // would otherwise short-circuit in `process_message` and never
                 // register as a downstream subscriber on the home node.
                 match target_addr {
-                    Some(target_addr) => EventResult::Event(
-                        ConnEvent::OutboundMessageWithTarget { target_addr, msg }.into(),
-                    ),
+                    Some(target_addr) => {
+                        // Track `tx → target_addr` so that if the downstream
+                        // peer disconnects before the reply arrives,
+                        // `prune_connection` enumerates this tx and
+                        // `handle_orphaned_transactions` wakes the parked
+                        // driver (#4154). `send()`-path messages already
+                        // registered via `sending_transaction`; this
+                        // covers the `op_execution_sender` path used by
+                        // `send_to_and_await` / `send_fire_and_forget`
+                        // (originator-loopback relay forward), which
+                        // bypasses `P2pBridge::send` entirely.
+                        //
+                        // Cleanup happens in the `TransactionCompleted`
+                        // handler which calls `remove_finished_transaction`.
+                        // A duplicate entry from CONNECT (which also
+                        // registers at `Ring::initiate_connect`) is
+                        // tolerated — `remove_finished_transaction` /
+                        // `prune_transactions_from_peer` clear all matching
+                        // entries.
+                        self.bridge
+                            .op_manager
+                            .ring
+                            .live_tx_tracker
+                            .add_transaction(target_addr, *msg.id());
+                        EventResult::Event(
+                            ConnEvent::OutboundMessageWithTarget { target_addr, msg }.into(),
+                        )
+                    }
                     None => EventResult::Event(ConnEvent::InboundMessage(msg.into()).into()),
                 }
             }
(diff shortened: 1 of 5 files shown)


sanity:
## Review responses

Five reviews completed (code-first, testing, skeptical, big-picture, Codex). Commit 65816118 addresses the MAJORs/P2s; remaining items called out below.

### Addressed (commit 65816118)

- **Codex [P2] / skeptical MAJOR \"Avoid corrupting live transaction tracking with duplicate txs\"** — `LiveTransactionTracker::add_transaction` was not idempotent or rebind-safe. Two callers exposed this: CONNECT's `Ring::initiate_connect` + `handle_op_execution` ride-along double-counted in `active_connect_transaction_count`; relays forwarding downstream then sending response upstream left a stale forward-index entry. Made `add_transaction` short-circuit same-pair re-registrations and atomically scrub the previous peer's Vec on cross-peer rebind. Pinned by two new tests (`add_transaction_is_idempotent_for_same_pair`, `add_transaction_rebinds_to_new_peer`).
- **Skeptical MAJOR \"Add happens even when callback is closed\"** — gated the new `live_tx_tracker.add_transaction` at `handle_op_execution` on `state.pending_op_results.contains_key(tx)`. Cancelled-driver teardown races no longer leak a tracker entry. The originator-loopback `send_fire_and_forget` case is still covered because the upstream client driver's `send_and_await` installed the waiter on an earlier call.
- **Big-picture MAJOR \"Stale doc in `operations/get/op_ctx_task.rs:26-29`\"** — replaced \"Disconnect detection relies on OPERATION_TTL (60s) — accepted ceiling\" with the new event-driven story (sub-ms wake; 60 s remains the upper bound for non-disconnect stalls).

### Not blocking, deferred to follow-up

- **Skeptical MAJOR #4 \"Per-attempt TransactionCompleted tears down client-level live_tx_tracker entry\"** — false alarm on inspection. CONNECT calls `release_pending_op_slot` once at driver exit (`crates/core/src/operations/connect/op_ctx_task.rs:236`, `:654`), not mid-op. The cleanup hook in the `TransactionCompleted` arm is correct.
- **Skeptical MAJOR #2 \"Leak in PUT/GET relay path\"** — resolved by the rebind-safe `add_transaction` above. When the relay later response-sends to its upstream via `P2pBridge::send`, the second `add_transaction(upstream, tx)` now scrubs the downstream's stale entry.
- **Testing MAJORs #1-#4** — pinned by the new `live_tx::tests::*` and the existing source-grep guard. A truly behavioral test of `handle_orphaned_transactions` would require constructing a full `OpManager` + `P2pBridge` (no lightweight fixture exists); the source-grep guard catches the most common regression (function reverting to logging-only) and was verified to fail when I temporarily reverted the orphan handler. The cleaner approach is a simulation test reproducing the disconnect-mid-GET scenario — happy to file that as a follow-up if reviewers want it.
- **Big-picture MINOR #2 \"Issue #4154 has 3 bullets, this PR covers one\"** — agreed. This PR addresses the wake-on-disconnect case (which was the user-observed 62 s symptom). Fail-fast-on-empty-peer-set, active-reroute-on-reconnect, and pre-routing telemetry are distinct code paths in `drive_client_get_inner`. Will file a follow-up to track the remaining slices and keep #4154 open against the follow-up — happy to do that as part of this PR if preferred, or after merge.
- **Code-first MINOR \"_gateways param unused\"** — agreed but kept because dropping it touches every disconnect call site (5 call sites across `p2p_protoc.rs`) — feels like churn outside the bug-fix scope. Will sweep in a separate refactor.
- **Skeptical NIT #8 \"Sub-ms retry without jitter\"** — drivers already advance through `skip_list`/exclusions and ride `OPERATION_TTL` for subsequent attempts. The wake replaces a 60 s wait with the next-peer probe; subsequent peer attempts are still gated by the existing per-attempt timeout. Adding exponential backoff on the wake itself would be a behavioral change worth discussing in a separate issue.
- **Source-grep guard brittleness** — fair; future refactor that renames the helper still needs to update the guard. The pattern matches existing house style (10+ instances in `put/op_ctx_task.rs`). Acceptable for now.

### Follow-up issues to file after merge

1. Simulation test reproducing disconnect-mid-GET (testing-reviewer recommendation).
2. Fail-fast / active-reroute / pre-routing telemetry — the remaining slices of #4154.
3. Drop the unused `_gateways` parameter from `handle_orphaned_transactions` and its call sites.

Re-running the four parallel reviewers on the new HEAD per the per-CODE-CONTENT rule before flipping to auto-merge.

[AI-assisted - Claude]

sanity:
## Simulation regression: root-caused, reverted

CI's Simulation job failed on `test_six_peer_contract_lifecycle` (state divergence on one of six peers). I reproduced locally and bisected:

- **commit b1372a61** (original wake-on-disconnect): PASS (15.6 s)
- **commit 65816118** (rebind-safe `add_transaction` + gate): FAIL (17.7 s, contract_b divergent on `[::1]:50600`)
- **fix-4154 HEAD with rebind-safe reverted**: PASS (15.6 s)
- **main baseline**: PASS (15.1 s)

### Root cause

`Ring::connection_maintenance` (`crates/core/src/ring.rs:2360-2365`) filters peers out of `neighbor_locations` using `has_live_connection`. Pre-fix, `add_transaction`'s unconditional Vec append left stale entries in earlier peers' `tx_per_peer` Vecs — so `has_live_connection` returned `true` for peers with leftover (technically-stale) entries from previous broadcasts/relays. Topology maintenance was implicitly relying on this "inflation" as a soft signal that hid recently-active peers from topology decisions.

The rebind-safe semantic I added in 65816118 tightened that: scrubbed stale entries, returned a correct count. This widened the set of candidate neighbors that `topology_manager.adjust_topology` saw, which led to a different topology decision path. In the 6-peer test that change caused one peer's CRDT broadcast chain to end up incomplete (contract_b: 5 peers converged to one hash, `[::1]:50600` had a different hash).

Codex's [P2] and the skeptical reviewer's #1 concern about `add_transaction` index corruption are still real — that's a latent issue. But fixing it correctly requires decoupling topology maintenance from `has_live_connection` (use a dedicated "in-flight critical work" signal), and that refactor is out of scope for #4154's wake-on-disconnect.

### Commit 659e36f4 (just pushed)

- Reverted `LiveTransactionTracker::add_transaction` to the pre-65816118 unconditional-append behavior.
- Removed the two tests that asserted rebind-safe semantics (`add_transaction_is_idempotent_for_same_pair`, `add_transaction_rebinds_to_new_peer`) — they were pinning behavior I just reverted.
- Kept everything else from 65816118: the `pending_op_results.contains_key` gate at `handle_op_execution`, the doc updates, the wake mechanism, the `TransactionCompleted` cleanup hook.

Documented the known limitation (and the topology-coupling reason) in the `add_transaction` rustdoc so a future refactor knows where to look.

### Tests

All 19 prior-passing tests (`ring::live_tx::tests`, `node::op_state_manager::tests`, `node::network_bridge::p2p_protoc::tests::handle_orphaned_*`) still pass. `test_six_peer_contract_lifecycle` now PASS locally on fix-4154 in 15.6 s, matching main's 15.1 s baseline.

Will re-run reviewers on the new HEAD per the per-CODE-CONTENT rule before flipping to auto-merge.

[AI-assisted - Claude]

sanity:
## Re-review round on 659e36f4 + 5c12c560 — all findings addressed

### Code-first reviewer: APPROVE, no blockers
Verified the revert is clean, gate placement correct, `try_release_pending_op_slot_on` semantics match the bug-prevention rule on `.send().await` in event loops.

### Testing reviewer: ADEQUATE FOR MERGE + one suggestion → ADDRESSED (5e6cc5f7)
Their main suggestion: a 5-line unit test pinning the load-bearing invariant for `Ring::connection_maintenance` (rebind must preserve `has_live_connection(old_peer) == true`). Added as `add_transaction_rebind_preserves_old_peer_has_live_connection` in `crates/core/src/ring/live_tx.rs`. Documents the invariant by name so a future tracker refactor surfaces a clear failure rather than the opaque "CRDT state divergence" from the 6-peer integration test.

### Skeptical reviewer: SHIP IT + one NIT → ADDRESSED (5c12c560)
NIT: the ride-along comment overclaimed "duplicates tolerated by `retain`" — true for same-peer dupes, slightly misleading for cross-peer rebind (stale Vec entry on old peer until that peer disconnects). Tightened the wording to distinguish the two cases and reference the detailed rustdoc.

### Big-picture reviewer: SHIP IT conditional on filing tracking issue → DONE (#4168)
Filed [#4168 "Decouple topology maintenance from `LiveTransactionTracker::has_live_connection`"](https://github.com/freenet/freenet-core/issues/4168) — the underlying coupling that forced the revert, with pointers to the filter site, the regression-pinning test, and the integration test for context.

### Codex: one [P1] → REJECTED as false positive
Codex claimed the source-grep guard test would deterministically fail because `body_after.find("\n    }\n")` would stop at an inner `if transactions.is_empty()` block. Verified locally: the inner block closes at indent 8 (inside the impl block), not 4, so `\n    }\n` correctly matches the outer function close. All 20 tests pass:

```
test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 2505 filtered out
```

### Final state
Per-CODE-CONTENT verification done on 659e36f4 and 5c12c560. Both incremental commits (test + comment tighten) are trivial-scope additions per the multi-model rule (one new test, one comment edit) — no re-review needed past this round. Awaiting CI on HEAD 5c12c560 before flipping to auto-merge.

[AI-assisted - Claude]

PR #4180, opened 2026-05-20
fix: compute geiger per-member status from exit code

### LINKED ISSUE #4173: scripts/static-checks.sh: per-member geiger reports status="ok" regardless of exit code
## Problem

In \`scripts/static-checks.sh\`, the per-member geiger path (used when the
target is a virtual workspace manifest) hardcodes \`--arg status "ok"\` for
the final summary even when one or more per-member runs failed:

\`\`\`bash
# Lines ~160-169 (in PR #4129 as merged)
jq -n \
  --arg status "ok" \
  --argjson rc  "\$rc" \
  --argjson dur "\$((t1-t0))" \
  ...
  '{status:\$status, exit_code:\$rc, duration_seconds:\$dur, ...}' \
  > "\$OUT/geiger.summary.json"
\`\`\`

Compare to \`run_check\` higher up which computes \`status\` from \`rc\`:

\`\`\`bash
local status; [[ \$rc -eq 0 ]] && status=ok || status="exit_\${rc}"
\`\`\`

The exit code is captured correctly into the JSON, but downstream consumers
checking only the \`status\` field will see "ok" for a failed run.

## Discovery

Flagged in the post-merge multi-model review of PR #4129 by Codex. See
PR #4171 for the related InterestManager race in the same PR.

## Suggested fix

In the per-member geiger path, compute \`status\` the same way \`run_check\`
does:

\`\`\`bash
local status
if [[ \$rc -eq 0 ]]; then status=ok; else status="exit_\${rc}"; fi
jq -n \
  --arg status "\$status" \
  ...
\`\`\`

[AI-assisted - Claude]

Basedfloppa:
## Problem

The per-member geiger summary section in `scripts/static-checks.sh` hardcodes `--arg status "ok"` on line ~169 regardless of the actual exit code of the `cargo geiger` run. When a per-package geiger check fails (e.g., new unsafe trait implementations), the report JSON records `status: "ok"` and the failure is masked. Only the global geiger mode (lines ~60-68) correctly propagates the exit code.

## Solution

Compute `geiger_status` from `$rc` the same way `run_check` does:

```bash
local geiger_status
if [[ $rc -eq 0 ]]; then geiger_status="ok"; else geiger_status="exit_${rc}"; fi
```

Replace the hardcoded `--arg status "ok"` with `--arg status "$geiger_status"`.

## Testing

Manual verification: the global mode uses the same pattern (`run_check` at line ~68), so this change aligns the per-member mode with the already-working global mode. No functional test exists for the static-checks script yet (tracked separately in the CI integration work item).

## Fixes

Closes #4173.

### DIFF
--- a/scripts/static-checks.sh
+++ b/scripts/static-checks.sh
@@ -166,8 +166,10 @@ run_geiger() {
   printf ']\n' >> "$OUT/geiger.raw"
   local t1; t1=$(date +%s)
 
+  local geiger_status
+  if [[ $rc -eq 0 ]]; then geiger_status="ok"; else geiger_status="exit_${rc}"; fi
   jq -n \
-    --arg status "ok" \
+    --arg status "$geiger_status" \
     --argjson rc  "$rc" \
     --argjson dur "$((t1-t0))" \
     --arg raw     "$OUT/geiger.raw" \


sanity:
**test-exempt justification:** This fixes exit-code handling in CI tooling (geiger per-member status). The behavior is exercised by the geiger CI job itself; it is not reachable from the Rust unit/integration suite. test-exempt label applied per CI Rule Lint guidance.

[AI-assisted - Claude]

sanity:
Closing as already fixed in main. The current `run_geiger()` in `scripts/static-checks.sh` already derives `status` from the exit code:

```bash
local status; [[ $rc -eq 0 ]] && status=ok || status="exit_${rc}"
jq -n --arg status "$status" ...
```

This is functionally identical to this PR's change (only the variable name differs: `status` vs `geiger_status`). The hardcoded `--arg status "ok"` this PR targets no longer exists on main, which is why the branch shows a conflict. The underlying issue #4173 is resolved by main. Thanks for the fix @Basedfloppa — it was the right call, just landed via another PR first.

[AI-assisted - Claude]

PR #4129, opened 2026-05-15
feat: static-analysis aggregator + perf cleanups it surfaced

Basedfloppa:
## Summary

Adds `scripts/static-checks.sh` — a one-shot aggregator that runs
clippy (with `perf` + `nursery`), `cargo-audit`, `cargo-deny`,
`cargo-outdated`, `cargo-bloat`, `cargo-machete`, `cargo-geiger` and
drops a single `target/static-checks/<ts>/report.json` plus per-tool raw
output. Missing tools record `{status: not_installed}` and the script
keeps going; geiger runs per-workspace-member (virtual manifests are
refused by `cargo-geiger`); machete output is parsed into a structured
`unused_by_crate` JSON list.

The second commit applies the perf-relevant findings the script
surfaced. Skips style-only / test-only / cascading-API findings.

## What the script found

Clippy run (`cargo clippy --workspace --all-targets -- -W clippy::perf -W clippy::nursery`):
4146 total findings across 24 distinct lints. Most are style nursery
warnings (`use_self` 1499, `redundant_pub_crate` 1027,
`missing_const_for_fn` 674, etc.) with no runtime impact.

Real perf-relevant categories in hot DHT / transport / executor code:

| Lint | Hits | Notes |
|---|---|---|
| `redundant_clone` | 209 total (16 prod) | Mostly in test code; the prod hits cluster in tracing event paths and op-context tasks |
| `significant_drop_tightening` | 145 (25 in hot files) | Locks held across `tracing::*!` macros + unrelated DashMap shard acquisitions |
| `future_not_send` | 33 | Missing `Sync` on generic params (`Executor<R, S>`, `StateStore<S>`, `InterestManager<T>`, `InboundConnectionHandler<S, TS>`) — production runs `Builder::new_multi_thread()`, so spawnable futures must be `Send` |
| `iter_with_drain` / `clear_with_drain` | 8 | `vec.drain(..)` where `into_iter()` would do, `HashMap::drain()` used as `clear()` |
| `or_fun_call` | 33 (1 prod hot) | `peer_connection.rs` resend-select arm allocated `Box::pin(ready(()))` every iteration |
| `significant_drop_in_scrutinee` | 12 | DashMap iter-guard held across a mutating call on the same map (deadlock pattern in `server.rs`); double-lock-across-`if let` in `connection_manager.rs` |
| `set_contains_or_insert` | 4 | Double `HashSet` lookup |
| `useless_let_if_seq` | 6 | One prod site in `hosting.rs` |

[...]

PR #4183, opened 2026-05-20
fix: restore tracing::warn order relative to placeholder push

Basedfloppa:
## Problem

PR #4129 restructured `connection_manager.rs:update_peer_identity` to satisfy a `significant_drop_tightening` clippy lint. During the restructure, the `tracing::warn!` call was moved from before the placeholder push (the original order) to after it. The original order was intentional — the warning should be emitted before creating the placeholder entry to match the expected log narrative: log the condition first, then record the compensatory action.

## Solution

Move the `tracing::warn!` back inside the `else` branch, before the `entry.push(..)` call, restoring the pre-#4129 ordering. The `significant_drop_tightening` concern is a non-issue here because `tracing::warn!` is not an await point and the RwLock guard is dropped at the end of the block regardless of where the warn is located.

Also renamed the now-unused `missing_entry` variable to `_missing_entry` to suppress the `unused_variables` warning.

## Testing

Behavioural impact is observational only (log output ordering). Verified that the `significant_drop_tightening` lint is still satisfied — the RwLock guard is dropped at the block boundary, the warn is not held across an await point.

## Fixes

Reviewer nit from iduartgomez on PR #4129.

### DIFF
--- a/crates/core/src/ring/connection_manager.rs
+++ b/crates/core/src/ring/connection_manager.rs
@@ -1209,10 +1209,11 @@ impl ConnectionManager {
         loc_for_peer.insert(new_addr, loc);
         drop(loc_for_peer);
 
-        // Hold the cbl write lock only for the actual mutation; drop it
-        // before the optional tracing::warn! and before acquiring
-        // connected_since / peer_health (clippy: `significant_drop_tightening`).
-        let _missing_entry = {
+        // Hold the cbl write lock for the mutation (including the
+        // placeholder-creation warning, emitted before the push it describes),
+        // then drop it before acquiring connected_since / peer_health
+        // (clippy: `significant_drop_tightening`).
+        {
             let mut cbl = self.connections_by_location.write();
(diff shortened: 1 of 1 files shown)


sanity:
**test-exempt justification:** This restores the ordering of an observational `tracing::warn!` relative to a placeholder push. It has no behavioral effect (logging only), so there is nothing to assert in a regression test. test-exempt label applied per CI Rule Lint guidance.

[AI-assisted - Claude]

sanity:
Adding `review-override`: the sole `rule-review/warnings` finding is "missing regression test for fix: PR". This change only reorders an observational `tracing::warn!` relative to a placeholder push — it has no behavioral effect, so there is nothing to assert in a regression test (same rationale as the `test-exempt` label). The stale-comment and dead-`_missing_entry`-bool items the bot also raised have been fixed in commit f5bf6e0e. All functional CI is green.

[AI-assisted - Claude]

PR #4187, opened 2026-05-20
fix: make event loop notification channel capacity configurable (#4145)

### LINKED ISSUE #4145: Bug: event loop stuck under load — notification channel saturation
## Summary

Freenet's event loop becomes unresponsive ("stuck") when its internal notification channel (`EVENT_LOOP_CHANNEL_CAPACITY=2048` in `node/network_bridge.rs`) fills up faster than the single-consumer event loop can drain. The bug manifests as:

- CtxSw spikes to ~1.1M/s while CPU drops to ~0.02 cores (process park/unpark loop, no real work)
- Subscribe / Put / Update operations time out after 30s and retry on different peers, creating a feedback loop that worsens the backlog
- WebSocket clients see `Connection reset` and `500 Internal Server Error` after 30s
- Connection count drops below bootstrap threshold (`<50`), triggers gateway rebootstrap which adds traffic spike on top
- Loop-fill episodes self-recover when external load drops or process is restarted; typical episode 5-75 minutes

This is **pre-existing in freenet 0.2.58** (and earlier) — observed in 14-day baseline = 11.7% of time spent in spike-state across 40 distinct episodes. NOT introduced by PR #4129.

## Reproduction

**Environment**: gateway node (aarch64, 2GB MemoryMax cgroup, 2 cores CPUQuota), running:
```
/root/.local/bin/freenet network --is-gateway \
  --max-number-of-connections 2000 --min-number-of-connections 50 \
  --transient-budget 500 --transient-ttl-secs 60 ...
```
hosting **554 contracts** with active client subscriptions.

**Trigger**: multiple webapp/client sessions performing UPDATEs simultaneously against the same contract on a gateway node hosting hundreds of contracts. Easily reproduced under sustained multi-client load.

**Observed episode pattern on a gateway node**:
- Episode A — ~17 min, CtxSw peak ~2.3M/s, self-recovered when external load eased
- Episode B — ~75 min, CtxSw sustained ~1.1M/s, required manual systemd restart
- Episode C — recurrence under continued multi-client load with full log capture (see Evidence section)

A 14-day historical baseline at process-exporter granularity (5-min step) shows the bug active **~11.7% of the time across ~40 distinct episodes** of varying duration.

## Evidence — log timeline from 2026-05-16 20:05 UTC incident

```
20:05:57.722828  freenet::ring: Starting subscription renewal cycle
                 needing_renewal=17 hosted=554

20:05:57.722871  WARN freenet::ring: Notification channel >50% full,
                 reducing renewal batch size
                 channel_remaining=0 channel_max=2048 batch_limit=2 contracts=17

20:05:57.722896  WARN freenet::ring: Notification channel >75% full during
                 renewal spawning, stopping early
                 channel_remaining=0 attempted=0

20:05:57.821212  WARN freenet::transport::connection_handler: Channel overflow:
                 dropped packets (bandwidth limit may be too high or receiver
                 too slow) total_dropped=64 elapsed_secs=10

20:06:09.422663  ERROR freenet::node::op_state_manager: release_pending_op_slot:
                 notification channel full for too long;
                 event loop may be stuck;
                 pending_op_results entry will be reclaimed by 60s sweep
                 timeout_secs=30

20:06:26.829084  ERROR contract_handling:upsert_contract_state:
                 notify_node_event: Notification channel full for too long,
                 event loop may be stuck
                 timeout_secs=30 channel_pending=2048 channel_remaining=0

20:06:26.829206  WARN  Failed to broadcast state change to network peers
                 error=notification channel error: notification channel send
                 timed out — event loop is likely stuck

20:06:36.261907  INFO freenet::operations::connect: Below bootstrap threshold
                 (48 < 50), attempting to connect to 3 gateways
```

Pattern repeats: tx timeouts every 100-500ms for ~20 seconds, then bootstrap reconnect attempts. Process is alive but unable to drain the queue.

## Root cause analysis

**Pipeline saturation**:

```
producers:
  - subscription renewal cycle (hosted=554 → ~17 renewals per cycle)
  - client UPDATE/PUT/SUBSCRIBE requests
  - state-change broadcasts to subscribers
  - inbound peer messages (subscribe-to-peer, state-delta)

   ↓ all multiplexed onto a single 2048-deep mpsc

freenet::node::network_bridge::EVENT_LOOP_CHANNEL_CAPACITY = 2048

   ↓ single consumer

priority_select::ProductionPrioritySelectStream
  (single-threaded event loop in p2p_protoc.rs)
   ↓ processes inline:
     • process_message (peer wire)
     • handle_op_execution
     • node events
     • client transaction completions
     • broadcast retries
```

When load exceeds drain rate:
1. Channel fills → adaptive throttle kicks in (`batch_limit=2`, then `stopping early`)
2. Producers calling `notification_sender.send()` block (or use `try_send` with timeout)
3. 30s timeout triggers `release_pending_op_slot` → tx marked failed → next-peer retry
4. Retries CREATE MORE pending ops → channel STAYS full
5. Self-reinforcing feedback. Recovery only when external load drops.

**Why CtxSw spikes but CPU drops**: producer threads block on `send()`, repeatedly futex-wake on channel slot release. Each retry-attempt-then-timeout cycle generates many context switches but tiny per-cycle CPU work. Classic livelock signature.

### Phase 1: make capacity configurable + add channel-depth observability

**Scope**: low-risk, immediate operator relief.

**Note on observability**: freenet does **not currently expose a Prometheus `/metrics` endpoint** or any other built-in metrics interface (only optional OpenTelemetry tracing via `trace-ot` cargo feature). The existing operational warning is the periodic `WARN freenet::ring: Notification channel >50% full` log line. Any per-scrape metric here would be a NEW infrastructure addition, not a hookup to existing plumbing.

Changes:
1. **Hoist `EVENT_LOOP_CHANNEL_CAPACITY` to config** — add `event_loop_channel_capacity: Option<usize>` to `NetworkArgs` (CLI) and `NetworkApiConfig` (config.toml), default 2048
2. **Thread param** through `event_loop_notification_channel(cap)` and call site in `p2p_impl.rs:312`
3. **Observability for channel depth** — at minimum, a periodic `tracing::info!` snapshot every ~30s logging `notifications_pending`, `op_execution_pending`, and the configured capacity. This is enough to expose via Loki / `grep` for now. A proper `/metrics` HTTP endpoint can be added later as a separate effort.
4. **Optional**: lower the >50%/>75% throttle thresholds proportionally (currently hardcoded around 1024/1536)

After Phase 1, operator can:
- Set `--event-loop-channel-capacity 16384` for gateways with high contract counts
- Watch channel depth via the new periodic log line (downstream Loki/log alerting on threshold)
- Alert on the existing `event loop may be stuck` `ERROR` line (already present today)

**Caveat**: bumping capacity only **delays** the stuck-state under sustained load; it does not fix the root cause. But it buys time for backpressure to propagate, and the depth log line provides leading-indicator visibility.

### Phase 2: multi-consumer event loop via tx-sharding

**Scope**: larger refactor, separate PR.

Approach:
- Replace single `priority_select` stream with **N worker tasks**, each owning a hash-bucket of transaction IDs
- Per-bucket maintains ordering invariant (messages within a transaction processed serially)
- Across buckets: parallel processing uses CPU headroom (gateway nodes typically have significant unused CPU capacity — observed ~80% idle on a 2-core quota during normal operation)
- Each worker has its own `notifications_receiver` clone (need to convert mpsc → broadcast or route at ingress)

Critical to verify before implementing:
- Does any code path require ordering ACROSS transactions? (If yes, sharding is unsafe.)
- Are op_state_manager mutex-protected fields accessed during process_message? (If yes, sharding doesn't help.)
- Routing layer at ingress: dispatch by `tx_id.hash() % N` — needs lightweight router task or direct sender selection on producer side

Multi-consumer ALSO doesn't solve the feedback loop fully — if total throughput < total demand, queue still saturates. But it raises the saturation threshold by N×.

### Phase 3 (optional, later): backpressure propagation

- When channel >75% full, return `EAGAIN`-equivalent error to clients via WebSocket
- Clients SHOULD slow down their request rate
- Currently clients keep sending even when server is gasping → makes things worse

## Workarounds (until fix lands)

- Manually restart freenet when `event loop may be stuck` appears in logs (the exact restart command depends on how freenet is deployed — systemd unit, supervisord, bare process, container, etc.)
- Reduce hosted-contract count on a single gateway (split across multiple gateway nodes)
- Throttle client UPDATE rates application-side
- Verify before tuning `--total-bandwidth-limit`: bandwidth saturation is a separate failure mode and is unrelated to the notification channel; adjusting it without confirming the bottleneck has no effect on this bug

## Related code

- `crates/core/src/node/network_bridge.rs:155` — `EVENT_LOOP_CHANNEL_CAPACITY` constant
- `crates/core/src/node/network_bridge.rs:157` — `event_loop_notification_channel()` factory
- `crates/core/src/node/network_bridge/p2p_protoc.rs:703` — `ProductionPrioritySelectStream` consumer
- `crates/core/src/node/op_state_manager.rs` — sender-side `notify_node_event` with 30s timeout
- `crates/core/src/ring.rs` — subscription renewal cycle producer; has adaptive throttle (`channel >50% / >75% full` warnings)
- `crates/core/src/contract/executor/runtime.rs` — broadcast state changes producer

## References

- 14-day historical Prometheus baseline (process-exporter `namedprocess_namegroup_*`, 5-min step) — 40 spike episodes catalogued; data captured externally to freenet itself, since freenet has no internal metrics export
- PR #4129 (clippy-driven lock-tightening) did NOT address this bug — it eliminated normal-load lock contention but does not change the producer/consumer mismatch under heavy load that drives this stuck-state

Basedfloppa:
## Problem
EVENT_LOOP_CHANNEL_CAPACITY was hard-coded at 2048 with no way for
operators or deployers to tune it. Under sustained multi-client load
the fixed capacity can cause channel saturation and context-switch
spikes.

## Solution
- NetworkArgs::event_loop_channel_capacity (Option<usize>, CLI/env)
- NetworkApiConfig::event_loop_channel_capacity (usize, serde default)
- event_loop_notification_channel_with_capacity(capacity) for prod path
- event_loop_notification_channel() wrapper (default capacity) for tests
- Capacity value is logged at channel creation time

## Testing
- cargo check --workspace passes
- Existing tests continue to use the default 2048 via the no-arg wrapper

## Fixes
Closes #4145

### DIFF
--- a/apps/freenet-ping/app/tests/common/mod.rs
+++ b/apps/freenet-ping/app/tests/common/mod.rs
@@ -172,6 +172,7 @@ pub async fn base_node_test_config_with_rng<R: Rng>(
             max_connections: None,
             bandwidth_limit: None,
             blocked_addresses,
+            event_loop_channel_capacity: None,
             transient_budget: None,
             // Use a longer TTL for tests (120s) to prevent connections from expiring
             // during test operations. The default 30s can cause flaky failures when
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -56,6 +56,7 @@ static ASYNC_RT: LazyLock<Option<Runtime>> = LazyLock::new(GlobalExecutor::initi
 
 const DEFAULT_TRANSIENT_BUDGET: usize = 2048;
 const DEFAULT_TRANSIENT_TTL_SECS: u64 = 30;
+const DEFAULT_EVENT_LOOP_CHANNEL_CAPACITY: usize = 2048;
 
 const QUALIFIER: &str = "";
 const ORGANIZATION: &str = "The Freenet Project Inc";
@@ -120,6 +121,7 @@ impl Default for ConfigArgs {
                 total_bandwidth_limit: None,
                 min_bandwidth_per_connection: None,
                 blocked_addresses: None,
+                event_loop_channel_capacity: None,
                 transient_budget: Some(DEFAULT_TRANSIENT_BUDGET),
                 transient_ttl_secs: Some(DEFAULT_TRANSIENT_TTL_SECS),
                 min_connections: None,
@@ -603,6 +605,10 @@ impl ConfigArgs {
                     .network_api
                     .blocked_addresses
                     .map(|addrs| addrs.into_iter().collect()),
+                event_loop_channel_capacity: self
+                    .network_api
+                    .event_loop_channel_capacity
+                    .unwrap_or_else(default_event_loop_channel_capacity),
                 transient_budget: self
                     .network_api
                     .transient_budget
@@ -904,6 +910,16 @@ pub struct NetworkArgs {
     #[arg(long, num_args = 0..)]
     pub blocked_addresses: Option<Vec<SocketAddr>>,
 
+    /// Capacity for the event loop notification and op execution channels.
+    /// Default: 2048. Increase under sustained multi-client load to reduce
+    /// channel saturation and associated context-switch spikes.
+    #[arg(long, env = "EVENT_LOOP_CHANNEL_CAPACITY")]
+    #[serde(
+        rename = "event-loop-channel-capacity",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub event_loop_channel_capacity: Option<usize>,
+
     /// Maximum number of concurrent transient connections accepted by a gateway.
     #[arg(long, env = "TRANSIENT_BUDGET")]
     #[serde(rename = "transient-budget", skip_serializing_if = "Option::is_none")]
@@ -1127,6 +1143,12 @@ pub struct NetworkApiConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub blocked_addresses: Option<HashSet<SocketAddr>>,
 
+    /// Capacity for the event loop notification and op execution channels.
+    /// Default: 2048. Increase under sustained multi-client load to reduce
+    /// channel saturation and associated context-switch spikes.
+    #[serde(default = "default_event_loop_channel_capacity")]
+    pub event_loop_channel_capacity: usize,
+
     /// Maximum number of concurrent transient connections accepted by a gateway.
     #[serde(default = "default_transient_budget", rename = "transient-budget")]
     pub transient_budget: usize,
@@ -1229,6 +1251,10 @@ use port_allocation::find_available_port;
 
 pub fn default_network_api_port() -> u16 {
     find_available_port().unwrap_or(31337) // Fallback to 31337 if we can't find a random port
+}
+
+pub(crate) fn default_event_loop_channel_capacity() -> usize {
+    DEFAULT_EVENT_LOOP_CHANNEL_CAPACITY
 }
 
 fn default_transient_budget() -> usize {
--- a/crates/core/src/node/network_bridge.rs
+++ b/crates/core/src/node/network_bridge.rs
@@ -429,6 +429,48 @@ mod tests {
         }
         assert_eq!(received, 2, "Should receive both messages");
     }
+
+    /// Test that event_loop_notification_channel_with_capacity creates channels
+    /// with the specified capacity and that the capacity is respected.
+    #[tokio::test]
+    async fn test_channel_with_custom_capacity() {
+        let custom_capacity: usize = 5;
+        let (notification_channel, notification_tx) =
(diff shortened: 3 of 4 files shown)


iduartgomez:
We will address the root cause in a different PR with a more robust fix (saw some of the suggestions in the issue).

PR #4188, opened 2026-05-20
build: add cargo-outdated and cargo-bloat JSON parsing to static-checks report

Basedfloppa:
## Problem
The cargo-outdated and cargo-bloat tools were run but their JSON
output was left unparsed in .raw files. Downstream consumers (CI
dashboards, code review bots) had no machine-readable access to
outdated-dependency counts or binary-size regression data.

## Solution
Add extraction blocks (lines 254-293):
- outdated.findings.json — per-crate actionable updates
- outdated.classification.json — compat vs major counts
- bloat.findings.json — file/text size plus top crates by size

Summary display updated to show outdated/bloat stats in the stderr
overview.

## Testing
Verified jq queries against real cargo-outdated output (12 workspace
members, 273 dependency entries). Syntax-checked with bash -n.

## Fixes
Closes the cargo-outdated/bloat parsing work item (#12)

### DIFF
--- a/scripts/static-checks.sh
+++ b/scripts/static-checks.sh
@@ -118,6 +118,61 @@ run_check bloat cargo-bloat \
 run_check machete cargo-machete \
   cargo machete
 
+# === extract outdated findings (JSON Lines format) =========================
+# cargo-outdated outputs one JSON object per workspace member, each with a
+# .crate_name and .dependencies array. We extract actionable updates where
+# .compat (semver-compatible) or .latest (any) indicates an available upgrade.
+if [[ -s "$OUT/outdated.raw" ]]; then
+  jq -s '
+    map(
+      {crate: .crate_name, deps: [.dependencies[] | select(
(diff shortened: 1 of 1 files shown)


iduartgomez (approved):

sanity:
Adding `review-override`: the `rule-review/warnings` finding asks for a regression test on the `fix:` commit (`da48a9bb`) that corrects the cargo-outdated `---` miscount. `scripts/static-checks.sh` is a static-analysis report generator that is not wired into any CI workflow, so a regression test for its internal jq filters would not run in CI and would provide little ongoing protection. The jq fix was verified directly against crafted input covering all three cases (compat update, major-only update, fully-up-to-date `compat:"---"/latest:"---"`): findings `update_count` and `compat_count`/`major_count` all come out correct, and the up-to-date entry is excluded. All functional CI (Clippy, Unit & Integration, Simulation, NAT, Windows, Fmt) is green.

[AI-assisted - Claude]

PR #4194, opened 2026-05-20
fix: downgrade same-version merge-rejected log + short-circuit WASM call (#4151)

### LINKED ISSUE #4151: Spammy INFO: 'Merge rejected ... New state version X must be higher than current version X' (same version)
## Symptom

INFO-level log spam on long-running peers:

\`\`\`
INFO contract_handling:upsert_contract_state{key=8j56dAN4...}:
     freenet::contract::executor::runtime:
     Merge rejected incoming state but local state is valid - not replacing (incoming state may be stale)
     contract=8j56dAN4... error=...
     reason: New state version 29650802 must be higher than current version 29650802
     local_state_size=484115 incoming_state_size=484115
     event=\"merge_rejected_valid_local\"
\`\`\`

Same version on both sides, same byte length, same contract. This is the trivial idempotent-push case: peer is re-pushing state we already have at the same version.

## Why it's noise

- Same-version, same-size means there is nothing to do.
- The contract's WASM \`update_state\` was invoked to reach this conclusion (the executor reports it as an executor error), so the cost isn't free.
- INFO level means it shows up in every operator's standard log, drowning out interesting events.

## What we want

Two changes:

1. **Drop to DEBUG.** A same-version rejection is not actionable at INFO. Operators don't need to see it.
2. **Short-circuit before the WASM call.** Compare \`incoming.version\` to \`current.version\` upstream of \`update_state\`. If \`incoming.version <= current.version\`, return early with a debug log line. Saves the WASM invocation entirely.

## Files

- \`crates/core/src/contract/executor/runtime.rs\` — the \`merge_rejected_valid_local\` event emitter
- \`crates/core/src/contract/handler.rs\` (or wherever \`upsert_contract_state\` lives) — version check should land here

## Repro

Run a mainnet peer for 1+ hour. \`grep -c \"merge_rejected_valid_local\" ~/Library/Logs/freenet/freenet.*.log\` — you'll get hundreds per hour.

iduartgomez:
## Problem

In high-traffic gateways running the website-contract (or any version-gated contract), the `merge_rejected_valid_local` INFO log fires on **every idempotent re-broadcast** that misses the dedup cache. The message `"Merge rejected ... New state version X must be higher than current version X"` is emitted at INFO level even though it is not operator-actionable — it just means a peer re-pushed a state the node already has at the same or newer version.

At busy gateways this produces hundreds of spammy INFO lines per minute, drowning out real problems. Closes #4151.

## Solution

Two complementary fixes in `crates/core/src/contract/executor/runtime.rs`:

### 1. Downgrade log level for idempotent rejections

At the `merge_rejected_valid_local` site (post-WASM path), check `merge_err.is_invalid_update_rejection()` — the predicate introduced in #3914 that identifies typed `ContractError::InvalidUpdate{,WithInfo}` rejections from WASM. If true: emit `tracing::debug!` (not `tracing::info!`). The structured `event="merge_rejected_valid_local"` field is preserved for grep/dashboard queries.

Other merge failures (OOG, WASM traps, internal runtime errors) keep INFO level — they may indicate real problems operators need to see.

### 2. Short-circuit byte-identical state updates

Before calling `attempt_state_update` (which invokes the WASM `update_state` function), compare the incoming full-state bytes to the stored state bytes. If they are identical, return `UpsertResult::NoChange` immediately — no WASM round-trip needed.

This covers the dominant spammy case: a peer re-broadcasts the exact state bytes the node already holds. For the minority case where incoming/current versions are equal but bytes differ, the WASM call still runs, and the resulting rejection is now logged at DEBUG.

A true pre-WASM version comparison is not feasible at the executor layer: state is opaque bytes at this level, and version semantics are defined per-contract inside WASM.

## Testing

Three regression tests added in `crates/core/src/contract/executor/pool_tests/merge_rejected_tests.rs`:

| Test | What it verifies |
|------|-----------------|
| `test_same_version_rejection_is_classified_as_invalid_update` | A `RejectInvalidUpdate` mock override produces an `ExecutorError` where `is_invalid_update_rejection()` = true. This is the predicate used to decide log severity. |
| `test_identical_state_short_circuits_before_wasm_update` | Byte-identical upsert returns `NoChange` without calling `update_state` (spy: the `RejectInvalidUpdate` override would `Err` if WASM were called). |
| `test_is_invalid_update_rejection_matches_website_contract_format` | Unit test that guards against format drift between the `is_invalid_update_rejection` string prefix and the website-contract error message. |

```
cargo test -p freenet --lib -- merge_rejected
running 3 tests
test ... ok
test ... ok
test ... ok
test result: ok. 3 passed
```

Also: `cargo fmt && cargo clippy -p freenet -- -D warnings` — clean.

## Fixes

Fixes #4151

### DIFF
--- a/crates/core/src/contract/executor/mock_wasm_runtime.rs
+++ b/crates/core/src/contract/executor/mock_wasm_runtime.rs
@@ -34,6 +34,11 @@ pub(crate) enum UpdateOverride {
     /// whether RelatedState entries are already populated. Drives the
     /// depth-limit branch in `bridged_upsert_contract_state`.
     AlwaysRequiresRelated(Vec<ContractInstanceId>),
+    /// Always return `ContractError::InvalidUpdateWithInfo` with the given
+    /// reason string. Used to test same-version / idempotent-push rejection
+    /// paths (issue #4151): the returned error must be classified as
+    /// `is_invalid_update_rejection()` and logged at DEBUG, not INFO.
+    RejectInvalidUpdate { reason: String },
 }
 
 /// A lightweight mock runtime at the `ContractRuntimeInterface` level that lets
@@ -126,6 +131,18 @@ impl ContractRuntimeInterface for MockWasmRuntime {
                         .collect();
                     return Ok(UpdateModification::requires(related)
                         .map_err(|e| anyhow::anyhow!("{e}"))?);
+                }
+                UpdateOverride::RejectInvalidUpdate { reason } => {
+                    // Simulate the WASM contract returning InvalidUpdateWithInfo
+                    // (e.g., "New state version X must be higher than current version X").
+                    // This produces an error that `ExecutorError::is_invalid_update_rejection()`
+                    // must classify as a benign idempotent-push rejection.
+                    use crate::wasm_runtime::ContractExecError;
+                    use freenet_stdlib::prelude::ContractError as StdlibContractError;
+                    let inner_err = StdlibContractError::InvalidUpdateWithInfo { reason };
+                    return Err(crate::wasm_runtime::ContractError::from(
+                        ContractExecError::ContractError(inner_err),
+                    ));
                 }
             }
         }
--- a/crates/core/src/contract/executor/pool_tests.rs
+++ b/crates/core/src/contract/executor/pool_tests.rs
@@ -1,6 +1,7 @@
 //! Tests for the contract executor module.
 
 mod conformance_tests;
+mod merge_rejected_tests;
 mod related_contract_tests;
 mod runtime_pool_tests;
 mod subscriber_limit_tests;
--- a/crates/core/src/contract/executor/pool_tests/merge_rejected_tests.rs
+++ b/crates/core/src/contract/executor/pool_tests/merge_rejected_tests.rs
@@ -1,0 +1,198 @@
+//! Regression tests for issue #4151 — same-version merge-rejected spam.
+//!
+//! When `update_state` WASM returns `InvalidUpdateWithInfo` for an idempotent
+//! re-push (incoming version == current version), the `merge_rejected_valid_local`
+//! event MUST:
+//!   1. Be classified as `is_invalid_update_rejection()` = true (the predicate
+//!      that gates log severity).
+//!   2. NOT fire at INFO level (this is tested indirectly via the predicate).
+//!
+//! Additionally, when the incoming state bytes are byte-identical to the stored
+//! state, the short-circuit MUST return `UpsertResult::NoChange` without invoking
+//! the WASM `update_state` function at all (verified by a spy-counted override).
+
+use either::Either;
+use freenet_stdlib::prelude::*;
+use std::sync::Arc;
+
+use crate::contract::executor::mock_wasm_runtime::{MockWasmRuntime, UpdateOverride};
+use crate::contract::executor::{ContractExecutor, Executor, UpsertResult};
+use crate::wasm_runtime::MockStateStorage;
+
+/// Create a MockWasmRuntime executor.
+async fn create_executor() -> Executor<MockWasmRuntime, MockStateStorage> {
+    let storage = MockStateStorage::new();
+    Executor::new_mock_wasm("merge_rejected_test", storage, None, None)
+        .await
+        .expect("create MockWasmRuntime executor")
+}
+
+/// Create a contract with the given code bytes and empty params.
+fn make_contract(code_bytes: &[u8]) -> ContractContainer {
+    let code = ContractCode::from(code_bytes.to_vec());
+    let params = Parameters::from(vec![]);
+    ContractContainer::Wasm(ContractWasmAPIVersion::V1(WrappedContract::new(
+        Arc::new(code),
+        params,
+    )))
+}
+
+// =========================================================================
+// Test: same-version rejection error is classified as is_invalid_update_rejection
+// =========================================================================
+
+/// Regression test for #4151.
+///
+/// When the contract WASM's `update_state` returns `InvalidUpdateWithInfo`
+/// (e.g. "New state version X must be higher than current version X"), the
+/// resulting `ExecutorError` MUST satisfy `is_invalid_update_rejection()`.
+/// The production code uses this predicate to decide whether to log at DEBUG
+/// (benign idempotent push) or INFO (unexpected merge failure).
+#[tokio::test(flavor = "current_thread")]
+async fn test_same_version_rejection_is_classified_as_invalid_update() {
+    let mut executor = create_executor().await;
+
+    // PUT the contract with initial state.
+    let contract = make_contract(b"same_version_test");
+    let key = contract.key();
+    let initial_state = WrappedState::new(vec![1, 2, 3, 4]);
+
+    executor
+        .upsert_contract_state(
+            key,
+            Either::Left(initial_state.clone()),
+            RelatedContracts::default(),
+            Some(contract),
+        )
+        .await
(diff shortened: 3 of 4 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4194

### Summary
- **PR Title:** fix: downgrade same-version merge-rejected log + short-circuit WASM call (#4151)
- **Type:** fix
- **CI Status:** all 13 checks green
- **Linked Issues:** #4151
- **HEAD reviewed:** 948a9bcac7ff8b7c0663b2a349b54f7d9c6c828a

---

### Code-First Analysis
**Independent Understanding:** Two complementary changes in `crates/core/src/contract/executor/runtime.rs::bridged_upsert_contract_state` to stop high-frequency INFO log spam:

1. **Byte-equality short-circuit (lines 1270-1286):** After fetching `current_state` and appending any `related_contracts` to `updates`, but BEFORE calling `attempt_state_update` (WASM), compare `incoming_full_state.as_ref() == current_state.as_ref()`. On match, emit a DEBUG-level `event="merge_skipped_identical"` and return `Ok(UpsertResult::NoChange)`.

2. **Log severity gate (lines 1463-1487):** At the `merge_rejected_valid_local` emission site, branch on `merge_err.is_invalid_update_rejection()` — the narrow predicate from #3914 that matches stdlib's `ContractError::InvalidUpdate{,WithInfo}` Display strings. Benign rejections log at DEBUG; OOG / traps / internal failures keep INFO. The `event="merge_rejected_valid_local"` structured field is preserved on both branches so dashboard/grep queries keep working.

**Stated Intent:** matches.
**Alignment:** matches; no scope creep, no unrelated drive-bys.

---

### Testing Assessment

| Unit | OK | `test_is_invalid_update_rejection_matches_website_contract_format` pins the exact website-contract error-string prefix the predicate matches. |
| Integration | OK (mock-runtime) | Two `MockWasmRuntime`-driven tests exercise the real `bridged_upsert_contract_state` code path (the executor methods, not a sidecar mock). |
| Simulation | N/A | Log-severity / short-circuit change does not need a SimNetwork run. |
| E2E | N/A | |

**Regression test:** present and valid. `test_identical_state_short_circuits_before_wasm_update` wires a `RejectInvalidUpdate` override and then sends the same bytes back; the assertion is that `update_state` was NOT called (because the override would have errored), and the result is `UpsertResult::NoChange`. This is exactly the right spy: a passing test demonstrates the WASM round-trip was avoided, which is the property the PR adds.

**Minor gaps (not blockers):**

1. The regression tests use 4-byte synthetic states. The reported #4151 hot path is ~484 KB website-contract state. The behavior under test (byte equality, error classification) doesn't change with state size, so this is fine for correctness — but a sentence in the test header noting the synthetic size choice and why size-independence holds would help future maintainers reading the test as a #4151 record.
2. No test asserts that the DEBUG vs INFO split actually fires the expected `tracing` level. The existing `log_update_contract_failure` family in `crates/core/src/operations/update.rs` ships with a `TestLogger`-based assertion (see e.g. `merge_rejected_invalid_update` tests there). The same pattern could be added here to pin the level, not just the predicate. Recommended, not required — the predicate test is enough to prevent the regression class.

---

### Skeptical Findings
**Risk Level:** low

| Side effects skipped by short-circuit are confirmed safe | resolved (info) | `runtime.rs:1276` | The short-circuit skips `attempt_state_update`, `commit_state_update`, `record_contract_update`, `send_update_notification`, `send_delegate_contract_notifications`, and the `BroadcastStateChange` fanout. All five are gated on actual state change in the existing code (line 1517 `if updated_state.as_ref() == current_state.as_ref()` already returns `NoChange` without calling `commit_state_update`), so the short-circuit just moves the decision earlier when we can be certain via byte equality. `record_contract_update` is the dashboard "last_updated" timestamp — by the same reasoning it correctly does NOT refresh on idempotent re-pushes. Subscription leases use `expires_at` (independent of update activity), not "refreshed on update". No load-bearing side effect is suppressed. |
| `related_contracts` arg is built into `updates` then thrown away on short-circuit | low | `runtime.rs:1260-1286` | If a peer pushes byte-identical main state alongside fresh `related_contracts`, the short-circuit skips the WASM `update_state` call that could have observed those `RelatedState` entries. This is consistent with contract semantics (`update_state` must be deterministic; a no-op input cannot legitimately mutate via side-channel) and matches the existing line-1517 short-circuit that already discards related-driven mutations when `updated_state == current_state`. No action needed. |
| Other event sites for the same noise class | resolved (already gated) | `operations/update.rs:336, 383` | The other INFO-level "merge rejected" sites (`event="merge_rejected_invalid_update"`) are already routed through `log_update_contract_failure` / `log_broadcast_to_streaming_failure`, both of which gate on the same `is_invalid_update_rejection()` predicate (PR #3914). The fix here is consistent with that existing pattern — there is no missed third site emitting the same noise class. |
| Delta-update path | low | `runtime.rs:1276` | The short-circuit fires only when `incoming_full_state.is_some()` (i.e. `Either::Left`). Pure delta updates still go through WASM. That is correct: a delta is not byte-comparable to a stored state. |
| Test name vs assertion scope | low | `pool_tests/merge_rejected_tests.rs:53-110` | The test name implies it tests the same-version case generically, but the predicate-only assertion can pass even for non-version-related `InvalidUpdateWithInfo` causes. Acceptable — the predicate is exactly what the production code checks. |

---

### Big Picture Assessment
**Goal Alignment:** yes. Operator-visible noise drops; cost-saving short-circuit covers the dominant idempotent case; non-benign failures keep their INFO/ERROR level so operators still see real problems.

**Anti-patterns detected:** none.
- No `#[ignore]`, no weakened asserts, no swallowed errors.
- New mock variant is feature-gated to test code via `#[allow(dead_code)]`-decorated enum, consistent with existing `ValidateOverride` / `UpdateOverride` patterns in the same file.
- The byte equality (not version comparison) is the right level: state is opaque bytes at the executor; version semantics live in WASM. The PR description explicitly addresses this.

**Removed code concerns:** the deleted INFO branch is replaced one-for-one with a conditional emit, no test was removed.

**Scope:** focused. 262 additions / 9 deletions across 4 files, all in the executor merge-rejection path.

---

### Documentation
- Code docs: inline rustdoc explains *why* the predicate gate matters (cross-references #3914 and #4151).
- Architecture docs: not required for a log-severity / cache-equivalence fix.
- User docs: not user-facing.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None.

#### Consider (Suggestions)
1. Add a `TestLogger`-based assertion that the level actually drops to DEBUG on `is_invalid_update_rejection()` and stays at INFO otherwise — same pattern as `log_update_contract_failure` tests in `crates/core/src/operations/update.rs`. Pins the behavior the PR title promises, not just the predicate.
2. Add a one-line header comment in `pool_tests/merge_rejected_tests.rs` noting the synthetic 4-byte state size and why size-independence applies (the byte-equality short-circuit and the predicate are both size-agnostic).
3. Optional: extract the byte-equality short-circuit into a named helper (`fn is_idempotent_full_state_update(...)`) so the same check could be reused if a future code path also wants to skip a redundant WASM round-trip. Pure cleanup, not load-bearing.

---

### Verdict
**State:** Ready to Merge

The fix is narrowly scoped, targets the exact site cited in #4151, preserves all load-bearing side effects (none of the bypassed calls fire on no-op updates anyway), keeps non-benign merge failures at INFO/ERROR, and the regression tests exercise the real production code path (not a sidecar). CI is green across all 13 checks. The suggestions above are quality-of-life nits, not blockers.

PR #4199, opened 2026-05-20
fix(clippy): allow wildcard match arms in delegate test assertions

### LINKED ISSUE #4197: clippy: wildcard_enum_match_arm fires in delegate.rs test code (cargo clippy --all-targets red on main)
## Problem

`crates/core/src/wasm_runtime/delegate.rs` has two `other =>` wildcard match arms in `#[cfg(test)]` code that trip `clippy::wildcard_enum_match_arm`:

- `delegate.rs:1609` — `other => panic!("Expected ApplicationMessage, got {other:?}")` on `OutboundDelegateMsg`
- `delegate.rs:1619` — `other => panic!("Expected ContextData, got {other:?}")` on `OutboundAppMessage`

`wildcard_enum_match_arm` is enabled crate-wide via `crates/core/Cargo.toml` `[lints]` (`wildcard_enum_match_arm = "warn"`). Running `cargo clippy --all-targets -- -D warnings` (or `--tests`) fails with 2 errors.

## Why CI is still green

CI's Clippy job runs `cargo clippy --locked -- -D warnings` (`.github/workflows/ci.yml:302`) — no `--all-targets` / `--tests`, so it only lints lib + bins, not test code. These two arms are in test modules, so CI never sees them.

## Likely cause

The freenet-stdlib 0.8.0 bump (#4144) added variants to `OutboundDelegateMsg` / `OutboundAppMessage`. The pre-existing `other =>` arms then started "also matching future added variants", which is what the lint flags.

## Suggested fix

Add `#[allow(clippy::wildcard_enum_match_arm)]` to the two arms, matching the existing pattern already used in the same file at `delegate.rs:542`, `:4180`, `:4188`. The wildcard-with-panic is intentional in these test assertions.

Optionally, also consider adding `--all-targets` to the CI Clippy invocation so test-code lint regressions are caught in future.

## Discovery

Found while running `cargo clippy -p freenet --lib --tests` during the PR #4171 work (post-#4129 review follow-up). Pre-existing, unrelated to #4171.

[AI-assisted - Claude]

sanity:
## Problem

`crates/core/src/wasm_runtime/delegate.rs` has two `other =>` wildcard match arms in `#[cfg(test)]` code that trip `clippy::wildcard_enum_match_arm` (enabled crate-wide via `crates/core/Cargo.toml` `[lints]`):

- `other => panic!("Expected ApplicationMessage, got {other:?}")` matching on `OutboundDelegateMsg`
- `other => panic!("Expected ContextData, got {other:?}")` matching on `OutboundAppMessage`

`cargo clippy --all-targets -- -D warnings` (or `--tests`) fails with 2 errors. CI's Clippy job runs `cargo clippy --locked -- -D warnings` without `--all-targets`/`--tests`, so it only lints lib+bins and never sees these test-code arms — that's why main's CI is green despite this. The arms started failing after the freenet-stdlib 0.8.0 bump added enum variants.

## Solution

Add `#[allow(clippy::wildcard_enum_match_arm)]` to each of the two enclosing `match` expressions, with a comment explaining the wildcard exists to satisfy the `#[non_exhaustive]` enums. The wildcard-with-panic is intentional in these test assertions — each test only cares about one specific variant, and the wildcard keeps the tests compiling against future stdlib variants.

This matches the existing pattern already used in the same file at `delegate.rs:542`, `:4180`, and `:4188`.

The change is deliberately minimal: just the two `#[allow]` attributes. A possible future follow-up is to add `--all-targets` to the CI Clippy invocation in `.github/workflows/ci.yml` so test-code lints are caught — that is intentionally **not** done here since it has broader blast radius (it could surface other test-code lint failures across the workspace).

## Testing

`cargo clippy -p freenet --all-targets -- -D warnings` now passes with zero errors (it failed with 2 errors before the change). `cargo fmt` clean.

Closes #4197

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/wasm_runtime/delegate.rs
+++ b/crates/core/src/wasm_runtime/delegate.rs
@@ -1604,10 +1604,16 @@ mod test {
                 ApplicationMessage::new(payload),
             )],
         )?;
+        // Wildcard satisfies #[non_exhaustive] on OutboundDelegateMsg so
+        // future stdlib variants don't break this test at compile time.
+        #[allow(clippy::wildcard_enum_match_arm)]
         let response: OutboundAppMessage = match &outbound[0] {
             OutboundDelegateMsg::ApplicationMessage(m) => bincode::deserialize(&m.payload)?,
             other => panic!("Expected ApplicationMessage, got {other:?}"),
         };
+        // Wildcard satisfies #[non_exhaustive] on OutboundAppMessage so
+        // future stdlib variants don't break this test at compile time.
+        #[allow(clippy::wildcard_enum_match_arm)]
         match response {
             OutboundAppMessage::ContextData(data) => {
                 assert!(


sanity:
Added the `test-exempt` label: this PR only adds two `#[allow(clippy::wildcard_enum_match_arm)]` attributes to existing test-code match expressions in `delegate.rs` (matching the pattern already used at `delegate.rs:542`, `:4180`, `:4188`). It introduces no behaviour change and has no regression-test surface — the "test" for it is `cargo clippy -p freenet --all-targets -- -D warnings` passing, which is verified in the PR description (fails with 2 errors before, clean after).

[AI-assisted - Claude]

Issue #4174, opened 2026-05-19
InterestManager::remove_all_peer_interests has a bidirectional consistency race against register_peer_interest


sanity:
## Problem

\`InterestManager::remove_all_peer_interests\` at
\`crates/core/src/ring/interest.rs:543-565\` (approx.) removes the peer's
\`peer_contracts\` entry up front, then iterates and mutates
\`interested_peers\` per contract. The two map writes are not atomic
against a concurrent \`register_peer_interest\` for the same
\`(peer, contract)\` pair:

1. \`remove_all_peer_interests(peer)\`: \`peer_contracts.remove(peer)\` removes
   the reverse-index entry. **At this moment, the peer's reverse-index entry
   is gone, but the forward-index entries (in \`interested_peers\`) still exist.**
2. \`register_peer_interest(C, peer, ...)\` runs (concurrently): under the
   \`interested_peers[C]\` shard guard, inserts \`peer\` into the forward index
   AND inserts \`C\` into \`peer_contracts[peer]\` (recreating the reverse entry).
3. \`remove_all_peer_interests\` continues iterating its previously-collected
   contract list and removes \`peer\` from \`interested_peers[C]\`.

Final state: \`peer_contracts[peer]\` exists with \`{C}\` (from step 2), but
\`interested_peers[C]\` no longer has \`peer\` (removed in step 3). The
bidirectional invariant \`peer ∈ peer_contracts[peer] ⟺ peer ∈ interested_peers[contract]\` is violated.

This is **pre-existing**; not introduced by PR #4129 or PR #4171. Discovered
while reviewing PR #4171's lock-discipline scope by Codex.

## Discovery

Flagged in the multi-model review of PR #4171 by Codex.

## Suggested fix

Either:
1. Reuse \`remove_peer_interest\` for per-contract cleanup (which holds the
   forward-index shard guard while updating the reverse index), or
2. Acquire \`peer_contracts[peer]\` write guard once at the start, snapshot
   the contract list, and hold the guard for the duration of the per-contract
   cleanup.

Option 1 is the cleaner fix.

## Severity

Low-to-medium. The peer-disconnect cleanup path is rare; reconnections
through the same peer key happen, but the timing window (between the
\`peer_contracts.remove\` and the per-contract iteration) is narrow. Still,
the resulting inconsistency is durable until the next \`remove_peer_interest\`
for that (peer, contract) pair.

[AI-assisted - Claude]

PR #4200, opened 2026-05-20
fix(ring): hold own_addr mutex across set_external_address (#4172)

### LINKED ISSUE #4172: set_own_addr TOCTOU between own_addr mutex and network_status::set_external_address
## Problem

`ConnectionManager::set_own_addr` at `crates/core/src/ring/connection_manager.rs:995-1011`
was tightened by PR #4129 to release the `own_addr` mutex before calling
`network_status::set_external_address(addr)`. The two updates are no longer atomic.

Two concurrent `set_own_addr(X)` and `set_own_addr(Y)` calls can produce durable
inconsistency:

1. A acquires mutex, writes \`own_addr = X\`, releases.
2. B acquires mutex, writes \`own_addr = Y\`, releases.
3. B calls \`network_status::set_external_address(Y)\`.
4. A calls \`network_status::set_external_address(X)\`.

Final state: \`own_addr = Y\` but \`network_status::external_address = X\`.
The mismatch persists until the next \`set_own_addr\` call.

Before PR #4129 the lock was held across both writes, so the second writer was
serialized on the mutex and saw the most-recent \`own_addr\` when it called
\`set_external_address\`.

Likelihood is low — \`set_own_addr\` is called from the ObservedAddress path
when a NAT-traversed peer learns its external address — but multiple
\`ObservedAddress\` reports could be processed concurrently and the inconsistency
is durable.

## Discovery

Flagged in the post-merge multi-model review of PR #4129 (Claude skeptical
reviewer). See PR #4171 for the related InterestManager race in the same
clippy-perf cleanup.

## Suggested fix

Either:
1. Hold the \`own_addr\` mutex across the \`set_external_address\` call (revert
   the PR #4129 tightening for this site), or
2. Have \`network_status::set_external_address\` derive its value from \`own_addr\`
   (single source of truth), or
3. Add a sequence-number or CAS to detect / reject stale updates.

Option 1 is the minimal fix matching the PR-#4171 approach for InterestManager.

[AI-assisted - Claude]

sanity:
## Problem

`ConnectionManager::set_own_addr` writes two mirrored pieces of state:
the `own_addr` mutex and `network_status::external_address`. PR #4129
narrowed the lock so it was released *before* `set_external_address`
was called, satisfying the `significant_drop_tightening` clippy lint
but making the two writes non-atomic.

Two concurrent `set_own_addr(X)` / `set_own_addr(Y)` calls can then
interleave:

1. A acquires mutex, writes `own_addr = X`, releases.
2. B acquires mutex, writes `own_addr = Y`, releases.
3. B calls `set_external_address(Y)`.
4. A calls `set_external_address(X)`.

Final state: `own_addr = Y` but `network_status::external_address = X`,
and the mismatch persists until the next `set_own_addr` call.

`set_own_addr` is called from the ObservedAddress path when a
NAT-traversed peer learns its external address; multiple
`ObservedAddress` reports can be processed concurrently, so the
inconsistency is low-likelihood but durable. Flagged in the post-merge
multi-model review of PR #4129.

## Solution

Revert the PR #4129 lock-narrowing for this one site: move the
`set_external_address` call back inside the `own_addr` guard so the two
state writes are atomic with respect to each other. The last mutex
holder now writes both values, so the two mirrors can never diverge.

The `tracing::debug!` call stays *outside* the lock — only the two
state writes need to be atomic w.r.t. each other; logging does not, and
keeping it out avoids serializing unrelated work under the guard.

An inline comment marks the guard as load-bearing for cross-structure
atomicity and explicitly warns that `significant_drop_tightening` /
`cargo clippy --fix` must not re-narrow it (cross-referencing #4172 and
#4129). This is the same cross-structure lock discipline restored for
`InterestManager` in PR #4171 and documented under "Cross-DashMap Lock
Discipline" in `.claude/rules/ring.md` — a `Mutex` here rather than a
DashMap shard guard, but the same spirit.

## Testing

Adds `test_set_own_addr_atomic_with_network_status` in the
`connection_manager.rs` test module. Each round spawns 8 threads
(aligned by a `Barrier`) calling `set_own_addr` with distinct
addresses; after all join, the test asserts `own_addr` and
`network_status::external_address` agree on the same final value. The
assertion is exact (`==`), never probabilistic — after all writers
finish there is no in-flight writer that could explain a difference.

A new `#[cfg(test)]` accessor `network_status::external_address()` lets
the test read the global's value directly.

Verified:
- **Without the fix** (lock narrowed): test fails reliably, diverging
  within ~100 rounds (observed failure at round 103).
- **With the fix**: passes deterministically — 4/4 runs green, plus
  the full `connection_manager` + `network_status` test modules
  (117 tests) pass.
- `cargo fmt`, `cargo clippy --locked -- -D warnings` clean.

Closes #4172

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -360,6 +360,19 @@ pub fn set_external_address(addr: SocketAddr) {
     }
 }
 
+/// Read this node's externally observed address.
+///
+/// Test-only accessor: production code reads `external_address` via
+/// [`get_snapshot`]. Used by the `set_own_addr` TOCTOU regression test in
+/// `connection_manager.rs` to verify `own_addr` and `external_address` stay
+/// consistent under concurrent writers (issue #4172).
+#[cfg(test)]
+pub(crate) fn external_address() -> Option<SocketAddr> {
+    NETWORK_STATUS
+        .get()
+        .and_then(|status| status.read().ok().and_then(|s| s.external_address))
+}
+
 // --- Snapshot types for rendering ---
 
 /// Snapshot of the current network status for rendering.
--- a/crates/core/src/ring/connection_manager.rs
+++ b/crates/core/src/ring/connection_manager.rs
@@ -2400,6 +2400,19 @@ mod tests {
     fn test_set_own_addr_atomic_with_network_status() {
         use std::sync::{Arc as StdArc, Barrier};
 
+        // GLOBAL-STATE CONTRACT: this test asserts on the process-global
+        // `network_status::external_address`, so it must remain the ONLY
+        // test that concurrently writes that global. That holds today:
+        // `set_external_address` has exactly one production caller
+        // (`set_own_addr`), and integration tests that spin up nodes run in
+        // separate test binaries (separate processes — each gets its own
+        // `NETWORK_STATUS`). Within this lib-test binary no other test calls
+        // `set_own_addr` / `set_external_address`. If a future lib-test does,
+        // it must serialize against this one (e.g. share a test mutex) or
+        // this `assert_eq!` can flake. (Reviewers: Claude testing/skeptical +
+        // Gemini all flagged this latent coupling — documented, not a current
+        // bug.)
+        //
         // `network_status::set_external_address` is a no-op until `init()` has
         // run. `init()` is idempotent (OnceLock — first caller wins), so this
         // is safe even if another test already initialized the global.
--- a/crates/core/src/ring/connection_manager.rs
+++ b/crates/core/src/ring/connection_manager.rs
@@ -993,16 +993,33 @@ impl ConnectionManager {
     /// Sets the own address unconditionally.
     /// Used when a peer behind NAT learns their external address from ObservedAddress.
     pub fn set_own_addr(&self, addr: SocketAddr) {
-        // Release the own_addr lock before notifying network_status and emitting
-        // tracing — neither needs the guard, and holding it across them serializes
-        // unrelated callers under load (clippy: `significant_drop_tightening`).
+        // Hold the `own_addr` mutex across the `network_status::set_external_address`
+        // call so the two state writes are atomic with respect to each other.
+        // `own_addr` and `network_status::external_address` mirror the same value;
+        // if the lock is released between the writes, two concurrent
+        // `set_own_addr(X)` / `set_own_addr(Y)` calls can interleave such that
+        // `own_addr` ends up `Y` while `external_address` ends up `X` permanently
+        // (writer B's mutex section completes before writer A reaches
+        // `set_external_address`). See issue #4172 / PR #4129.
+        //
+        // This is the same cross-structure atomicity discipline described in
+        // `.claude/rules/ring.md` ("Cross-DashMap Lock Discipline") — a Mutex
+        // here rather than a DashMap shard guard, but the same spirit applies.
+        //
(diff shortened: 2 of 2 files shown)


sanity:
## Full Multi-Model Review — #4200 (fix #4172)

**HEAD SHA reviewed:** `fc2e9cfe` · **CI:** all 13 checks green · **Tier:** Full (concurrency / TOCTOU)

### Verdict: MERGE-AFTER-FIX — one blocking test-isolation defect

The production fix is **correct**. Reverting the PR #4129 lock-narrowing so `set_external_address` runs inside the `own_addr` guard makes the two mirrored writes atomic and closes the TOCTOU. `own_addr` is a `parking_lot::Mutex` (sync) and `set_external_address` is a synchronous fn with no `.await` and only takes the unrelated `NETWORK_STATUS` `RwLock` — so there is **no deadlock and no lock-order inversion** (the only ordering is `own_addr` → `NETWORK_STATUS`, never the reverse). The load-bearing-comment warning against a future `clippy --fix` re-narrowing is good practice.

### Blocking finding

**[HIGH] `test_set_own_addr_atomic_with_network_status` is non-deterministic under `cargo test`** — `crates/core/src/ring/connection_manager.rs:2400`

The test asserts a per-`ConnectionManager` `own_addr` equals the **process-global** `network_status::external_address`. That global is a `OnceLock<Arc<RwLock<NetworkStatus>>>` shared by the entire `freenet` lib test binary. Any other test in the binary that calls `set_own_addr` (production caller: `operations/connect/op_ctx_task.rs:472`, reachable from `operations::connect` tests) writes `external_address` and can race this test's barrier-join → assertion window.

Reproduced: with filter `ring::connection_manager node::network_status` the test failed `116 passed; 1 failed` — `own_addr=127.0.0.1:20007` but `external_address=127.0.0.1:8000` (a `make_addr(8000)` value from a different test). It passes in isolation and under `cargo nextest` (per-process isolation — which is why CI is green), but `cargo test` is the documented pre-commit command in `.claude/rules/git-workflow.md`, so this is a flaky test by the project's own definition.

This is a **test defect, not a production-code defect** — the fix shipping in `connection_manager.rs`/`network_status.rs` is sound. But per the flaky-tests rule it should be resolved before merge. Options: (a) add a nextest `test-group` serialization override and document the test as relying on process isolation, or (b) restructure the assertion so it does not depend on a global other tests can mutate. Needs a design call — not a trivial mechanical fix.

### Codex (external, non-Claude) review
Codex was run from the #4200 worktree and reported `[P1]` on `interest.rs` and `[P2]` on `static-checks.sh`. **Both are out of scope** — neither file is in this PR's diff (`git diff` shows only `connection_manager.rs` + `network_status.rs`). The `interest.rs` finding is the pre-#4174 racy code that exists on this branch because #4200 branched before #4201 merged; #4201 already fixed it on main. The `static-checks.sh` geiger issue is a real but unrelated pre-existing bug. No actionable Codex finding against #4200's actual diff.

### Other perspectives
- **Testing:** regression test fails without the fix (PR notes: diverges by ~round 103) and passes with it — confirmed, passes in isolation. Gap: the isolation flakiness above.
- **Skeptical:** no new deadlock, no lock inversion, `set_external_address` non-async. Tracing correctly kept outside the guard.
- **Big-picture:** focused diff, no scope creep, matches the PR #4171 discipline. No merge-order conflict with #4201 (disjoint files; rebases clean onto current main).

[AI-assisted - Claude]

PR #4201, opened 2026-05-20
fix(ring): close bidirectional-consistency race in remove_all_peer_interests

### LINKED ISSUE #4174: InterestManager::remove_all_peer_interests has a bidirectional consistency race against register_peer_interest

sanity:
## Problem

`InterestManager::remove_all_peer_interests` (`crates/core/src/ring/interest.rs`) had a bidirectional-consistency race against `register_peer_interest`.

The old implementation removed the peer's `peer_contracts` entry **up front**, captured a snapshot of the contract set, then iterated that snapshot mutating `interested_peers` directly. A concurrent `register_peer_interest(C, peer, ..)` for a contract `C` already in the snapshot — running *after* the up-front `peer_contracts.remove` but *before* the per-contract `interested_peers[C]` mutation — re-inserts `peer` into **both** maps. `remove_all_peer_interests` then strips `peer` from `interested_peers[C]` (it still has `C` in its stale snapshot) while the freshly re-created reverse entry in `peer_contracts[peer]` survives.

The result is a one-sided "ghost" that violates the invariant:

```
peer ∈ peer_contracts[peer]  ⇔  peer ∈ interested_peers[contract]
```

A ghost reverse entry makes a disconnected peer permanently appear "interested" in a contract it is no longer in the forward index for, so it can never be cleaned up and pollutes future `get_contracts_for_peer` lookups.

This is a *secondary-origin* remover (it starts from the `peer_contracts` reverse index), so the Cross-DashMap Lock Discipline restored by PR #4171 did not protect it — `.claude/rules/ring.md` explicitly named it as the unprotected example.

## Solution

Per the issue's recommended option 1: make `remove_all_peer_interests` delegate each per-contract cleanup to `remove_peer_interest`, which already holds the `interested_peers[contract]` shard guard across its `peer_contracts` update — so each per-contract removal is atomic against a concurrent re-registration.

The new flow:
1. **Snapshot** the contract set from `peer_contracts` read-only (do NOT remove the entry up front).
2. Call `remove_peer_interest(&contract, &peer)` for each contract.

Semantics preserved: the function still returns the count of contracts the peer was actually removed from (now counted from `remove_peer_interest` returning `true`, which is accurate even when a concurrent removal cleared an entry between snapshot and call). `remove_peer_interest` additionally runs `cleanup_contract_if_no_interest` per contract — correct behaviour here, since losing a peer's interest *should* trigger cleanup. There was no stats counter or `pending_removals` bookkeeping in the old body that needed preserving.

The `.claude/rules/ring.md` caveat is updated: `remove_all_peer_interests` is no longer an unprotected secondary-origin remover, and the caveat now documents the safe pattern (snapshot the secondary index read-only, then call the existing per-key primary-origin remover).

## Testing

Adds `test_concurrent_remove_all_preserves_bidirectional_invariant`, a barrier-synced concurrent regression test modeled on PR #4171's `test_concurrent_add_remove_preserves_hash_index_invariant`. Each round **pre-registers** a fresh contract on the main thread (so it is guaranteed to be in the remover's snapshot — this is what makes the race observable) then races a re-`register_peer_interest` against `remove_all_peer_interests`, and asserts the bidirectional invariant directly via map inspection.

**Sensitivity:** with the racy body restored, the test fails **10/10** runs of 200,000 rounds each (~80-90% of rounds trip a ghost). With the fix applied it passes **10/10**.

- `cargo fmt` clean
- `cargo clippy --locked -- -D warnings` clean
- `cargo test -p freenet --lib` — 2543 passed, 0 failed

Closes #4174

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -212,9 +212,20 @@ index and works back to the primary — is NOT protected by this
 discipline. Those need their own atomicity story: either share a
 lock order with the add path, batch the primary-side updates under a
 single guard, or use a reconciliation/two-phase remove protocol.
-Example of an unprotected secondary-origin remover that still has a
-bidirectional-consistency race even with the rule applied:
-`InterestManager::remove_all_peer_interests` (issue #4174).
+
+A secondary-origin remover that broke this way was
+`InterestManager::remove_all_peer_interests` (issue #4174). Its old
+shape removed the `peer_contracts[peer]` reverse entry up front, then
+iterated a stale snapshot mutating `interested_peers` directly — a
+concurrent `register_peer_interest` could re-create one side after the
+snapshot, leaving a one-sided ghost. The fix (PR for #4174) does NOT
+add a new atomicity primitive: it makes the secondary-origin remover
+delegate each per-contract cleanup to the primary-origin remover
+`remove_peer_interest`, which already holds the
+`interested_peers[contract]` shard guard across its `peer_contracts`
+update. So the safe pattern for a secondary-origin bulk remover is:
+snapshot the secondary index (read-only, do NOT remove it up front),
+then call the existing per-key primary-origin remover for each entry.
 ```
 
 ## Trigger-Action Rules
--- a/crates/core/src/ring/interest.rs
+++ b/crates/core/src/ring/interest.rs
@@ -539,31 +539,50 @@ impl<T: TimeSource + Sync> InterestManager<T> {
     /// Remove all interests for a peer (called on peer disconnect).
     ///
     /// Uses the reverse index for O(1) lookup instead of O(contracts) scan.
-    /// Returns the number of contracts from which the peer was removed.
+    /// Returns the number of contracts from which the peer was actually removed.
+    ///
+    /// # Concurrency
+    ///
+    /// This is a *secondary-origin* remover: it starts from the
+    /// `peer_contracts` reverse index. It does NOT remove the
+    /// `peer_contracts` entry up front and then mutate `interested_peers`
+    /// directly — that older shape had a bidirectional-consistency race
+    /// (issue #4174): a concurrent `register_peer_interest(C, peer, ..)`
+    /// running between the up-front `peer_contracts.remove` and the
+    /// per-contract `interested_peers` mutation could re-insert `peer`
+    /// into both maps, after which this method would strip `peer` from
+    /// `interested_peers[C]` while leaving the reverse entry in
+    /// `peer_contracts[peer]` behind.
+    ///
+    /// Instead it merely *snapshots* the contract set and delegates each
+    /// per-contract cleanup to `remove_peer_interest`, which holds the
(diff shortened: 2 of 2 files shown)


sanity:
Review note (code-first / skeptical / Codex all observed this): the return value's *meaning* changes slightly. The old body returned `contracts.len()` — the size of the reverse-index snapshot, unconditionally. The new body returns the count of contracts where `remove_peer_interest` returned `true` (i.e. the peer was actually removed). Under concurrency or a pre-existing one-sided ghost these differ; the new value is strictly more accurate and matches the doc's "actually removed" wording. The only consumer is a `tracing::info!` field in `execute_pending_removals`, so no caller behaviour changes.

[AI-assisted - Claude]

PR #4206, opened 2026-05-21
fix: stop Windows installer self-destruct from blanket taskkill

### LINKED ISSUE #4205: Bug: Freenet windows installer self-destructs before installing anything.
Tried a from-scratch reinstall on Windows to test something, Installer just silently closed shortly after hitting the install button. Fed the details through Codex and got this:

---

The most likely cause in this repo is the Windows “stop existing service” step killing the installer itself.

Flow:

- The GUI install path calls the already-installed binary with `service stop` before copying the new one: `freenet-core/crates/core/src/bin/commands/setup_wizard/installer.rs:52`.
- On Windows, `service stop` runs `taskkill /f /im freenet.exe /fi "PID ne <stop-command-pid>"`: `freenet-core/crates/core/src/bin/commands/service.rs:3319`.
- If the GUI installer you double-clicked is also named `freenet.exe`, that `taskkill` matches it too. It excludes only the child `freenet.exe service stop` process, not the parent GUI installer. Result: the installer window just disappears.

That matches “I see the installer GUI, press Install, then it silently exits”, especially if `%LOCALAPPDATA%\Freenet\bin\freenet.exe` already exists from a previous install.

Quick way to confirm/work around: rename the downloaded installer binary to something like `freenet-installer.exe` and double-click that. Since `taskkill /im freenet.exe` won’t match `freenet-installer.exe`, the GUI should survive the stop step and continue.

A proper fix would be to make Windows stop logic target only the wrapper/node processes it owns, or pass/exclude the parent installer PID too, rather than killing every process named `freenet.exe`.

---

I then renamed the `freenet.exe` to `freenet-installer.exe` as suggested and Freenet installed as expected.

sanity:
## Problem

The Windows GUI installer is itself named `freenet.exe`. When the user
presses **Install**, the installer's first step runs `service stop` on the
already-installed binary, which executed:

```
taskkill /f /im freenet.exe /fi "PID ne <stop-command-pid>"
```

`taskkill /im freenet.exe` matches **every** process with that image name —
so it killed the GUI installer too. It excluded only the short-lived
`service stop` child, not the parent installer. The installer window
silently disappeared shortly after Install was pressed, before anything was
copied (#4205). Renaming the downloaded binary to `freenet-installer.exe`
worked around it, confirming the diagnosis.

The same blanket `taskkill /im freenet.exe` pattern also lived in
`stop_and_remove_service` (uninstall) and the post-update restart in
`update.rs` — both could kill an unrelated `freenet.exe`.

## Solution

Replace the blanket image-name kill with **command-line-targeted**
selection. `freenet.exe` processes are enumerated via PowerShell
`Get-CimInstance` and classified by their **subcommand prefix** (the first
argument token): `service run-wrapper` → the wrapper, `network` → the node
child. Only those two are terminated, by PID; every other `freenet.exe`
(the GUI installer, a concurrent CLI command, the caller itself) is left
alone. Killing is split into a wrapper pass then a re-enumerated node pass,
so a node the wrapper respawns mid-sweep is not missed.

All three blanket-taskkill sites — `stop_service`, `stop_and_remove_service`,
and the `update.rs` post-update restart — now route through one shared
helper, `kill_freenet_service_processes()`.

The installer additionally stopped the service by shelling out to the
**already-installed** binary — so a buggy *old* version's `service stop`
would still self-destruct the installer during an upgrade. The installer
now runs the stop **in-process**, using this binary's fixed logic
regardless of which version is installed.

`kill_stale_freenet_processes` (the wrapper-startup orphan reaper) is
migrated off `wmic` onto the same PowerShell helper: `wmic` is deprecated
and absent by default on Windows 11 24H2 / Server 2025, where it was
silently no-opping. After this PR there is no `wmic` dependency left in the
file.

The kill decision is extracted into pure, cross-platform functions
(`command_line_args`, `classify_freenet_process`,
`parse_freenet_process_listing`) so it can be unit-tested on the Linux CI
matrix — the rest of this Windows-only code is otherwise only
compile-checked by the `windows_check` job.

### Behavior note

`freenet service stop` now stops only the wrapper/node. A node a developer
started manually via the top-level `freenet network` subcommand is still
classified as a node and stopped (intended — `service stop` should stop a
running node however it was launched); a foreground node from the
"run without installing" path (plain `freenet.exe`, no subcommand) is no
longer killed by `service stop`, which is the correct narrowing.

## Testing

- 7 pure-function unit tests in `service.rs`, including
  `classify_freenet_process_spares_installer_and_other_commands` (the
  #4205 regression — the installer command line must classify as `None`)
  and `classify_freenet_process_matches_subcommand_prefix_only` (a token
  like `network` appearing as an option value must not match).
- `cargo fmt`, `cargo clippy --locked -p freenet`, and the tests all pass
  on Linux.
- Windows-gated code could not be compiled locally (no MSVC C toolchain for
  cross-compilation); CI's `windows_check` job covers compilation on a real
  Windows runner. The PowerShell/`taskkill` invocations themselves are not
  exercised by CI and warrant a manual Windows smoke-test before relying on
  them.

## Fixes

Closes #4205

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/bug-prevention-patterns.md
+++ b/.claude/rules/bug-prevention-patterns.md
@@ -53,6 +53,7 @@ Source-level regression pins live in:
 
 - `commands::tray::tests::open_log_file_spawn_must_null_all_three_standard_handles`
 - `commands::service::tests::spawn_update_command_must_null_all_three_standard_handles`
+- `commands::service::tests::taskkill_pid_must_null_all_three_standard_handles`
 
 A future revert of any of these null-stdio calls fails CI with a
 specific, issue-numbered error message rather than shipping the
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -3196,21 +3196,11 @@ pub fn stop_and_remove_service(_system: bool) -> Result<bool> {
 
     let had_registry = run_key.delete_value("Freenet").is_ok();
 
-    // Kill any running freenet processes (wrapper + child), excluding ourselves
-    let our_pid = std::process::id().to_string();
-    drop(
-        std::process::Command::new("taskkill")
-            .args([
-                "/f",
-                "/im",
-                "freenet.exe",
-                "/fi",
-                &format!("PID ne {}", our_pid),
-            ])
-            .stdout(std::process::Stdio::null())
-            .stderr(std::process::Stdio::null())
-            .status(),
-    );
+    // Kill the running Freenet service processes (wrapper + node child).
+    // Targets only the service's own processes by command line so an
+    // unrelated freenet.exe — e.g. the GUI installer — is never killed
+    // (issue #4205).
+    kill_freenet_service_processes();
 
     // Also clean up any legacy scheduled task from older installs
     let had_task = std::process::Command::new("schtasks")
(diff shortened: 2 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #4206

### Summary
- **PR Title:** fix: stop Windows installer self-destruct from blanket taskkill
- **Type:** fix
- **CI Status:** pending on HEAD `2446db7d` (light checks green; full suite registering)
- **Linked Issues:** Closes #4205
- **Review tier:** Full — Windows install/process-management surface; platform-gated code CI can only compile-check
- **Reviewers run:** code-first, testing, skeptical, big-picture (freenet:* subagents) + external Codex — **three full rounds**, re-run after each batch of fixes per the per-content re-review policy.

This review consolidates three rounds. The PR evolved substantially in response to feedback; what follows is the verdict on the current HEAD.

---

### Code-First Analysis
**Independent understanding:** Replaces three blanket `taskkill /f /im freenet.exe` calls — which kill *every* process named `freenet.exe`, including the GUI installer — with command-line-targeted killing. `freenet.exe` processes are enumerated via PowerShell `Get-CimInstance`, classified by subcommand prefix (`classify_freenet_process`: first arg token `network` → node, `service run-wrapper` → wrapper), and only those are `taskkill`ed by PID. The installer also stops the service in-process rather than shelling out to the (possibly buggy old) installed binary.

**Alignment:** Matches the PR description and issue #4205 exactly; goes appropriately further by also fixing the upgrade-from-old-version variant. Final round: *"correct and complete, no new issues, the PR is clean."*

---

### Testing Assessment
**Coverage level:** Adequate.

| Unit | ✅ | 9 pure-function tests (`command_line_args`, `classify_freenet_process`, `parse_freenet_process_listing`) — run on the Linux CI matrix |
| Integration | N/A | — |
| Simulation | N/A | — |
| E2E | ⚠️ | Windows-gated kill/enumeration code is compile-checked only (`windows_check`); not runtime-exercised by CI |

**Regression test:** Present — `classify_freenet_process_spares_installer_and_other_commands` directly pins the #4205 scenario (installer command line → `None`); `classify_freenet_process_matches_subcommand_prefix_only` pins the Codex finding. A `taskkill_pid` source-level null-stdio pin guards the `FreeConsole()` rule.

---

### Skeptical Findings
**Risk level:** Low (after fixes). No High-severity defects in any round.

| Blanket `taskkill /im` kills the installer (#4205) | high | Fixed — command-line-targeted classification |
| `is_freenet_service_process` matched any token (`freenet local --config-dir network`) | med | Fixed — subcommand-prefix matching |
| Non-atomic kill → node-respawn race | med | Fixed — wrapper-first + bounded re-enumerating node loop |
| `wmic` left in `kill_stale_freenet_processes` (deprecated, gone on Win11 24H2) | med | Fixed — migrated to the shared PowerShell helper |
| `taskkill` spawn missing explicit stdin-null (FreeConsole rule) | low | Fixed — `taskkill_pid` nulls all three handles |
| Self-updating wrapper could respawn a wrapper mid-sweep | low | Accepted & documented — pre-existing race class, millisecond window; both reviewers recommended acceptance over a redesign |

---

### Big Picture Assessment
**Goal alignment:** Yes — fixes #4205 at the root cause across all three blanket-`taskkill` sites plus the installer self-stop path.
**Anti-patterns detected:** None. No removed tests, no weakened assertions.
**Scope:** Appropriate — the `wmic` migration and pure-function extraction are in-scope consolidation, not creep.
**Docs:** `FreeConsole()` cross-reference comment updated; new `taskkill_pid` pin added to the `bug-prevention-patterns.md` inventory; PR description matches the code.

---

### Documentation
- Code docs: complete (loop rationale, single-pass precondition, return-value semantics, known limitation all documented).
- Architecture/user docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None outstanding — all addressed across the three rounds.

#### Consider (Suggestions)
1. **Manual Windows smoke-test before relying on the fix.** The `Get-CimInstance` / `taskkill` invocations are `#[cfg(windows)]` and only compile-checked by CI — exercise `freenet service stop`, a fresh install, an upgrade-over-existing-install, and uninstall on a real Windows box. (Noted in the PR description.)

---

### Verdict
**State:** Ready to Merge (once CI is green on the current HEAD)
**HEAD SHA reviewed:** `2446db7d`

Three full review rounds converged with no blocking findings. Every Medium/Low finding was fixed or — for the one wrapper-self-respawn asymmetry — explicitly accepted and documented on both reviewers' recommendation. The round-3→round-4 diff is documentation-only (verified), so a light re-check rather than a fourth full round applies. Merge is gated only on CI going green on `2446db7d`.

[AI-assisted - Claude]

PR #4209, opened 2026-05-21
fix: make River release announcement survive the gateway-restart race

### LINKED ISSUE #4208: release-announce: River announcement step fails with HTTP 500 (node unreachable during gateway update + log permission denied)
## Summary

The `release-announce.yml` workflow's **"Post to River (via nova webhook)"** step has failed on the last two releases — **v0.2.60** (run 26188709920) and **v0.2.61** (run 26244063636). Both report `nova /announce/river returned HTTP 500`.

The Matrix announcement step is unaffected and succeeds. This is a non-critical, pre-existing infra bug — the release artifacts and gateway updates are fine; only the River chat announcement does not go out.

## Root cause

From the nova `freenet-release-agent` journal at announce time, `/usr/local/bin/announce-to-river.sh` exits 1 for two reasons:

```
freenet_release_agent::announcer: spawning announce command command=/usr/local/bin/announce-to-river.sh
announce-to-river.sh: ERROR: Freenet node not reachable at http://127.0.0.1:7509/
announce-to-river.sh: line 36: /var/log/freenet-release-agent-river.log: Permission denied
freenet_release_agent::server: announcer spawn failed error=announce command exited early with exit status: 1
```

1. **Node unreachable — a race with the gateway update.** `release.published` fires `gateway-update.yml` and `release-announce.yml` concurrently. The River announce POSTs to nova, where `announce-to-river.sh` needs the local node API at `http://127.0.0.1:7509/` — but at that exact moment `gateway-update.yml` is restarting `freenet-gateway.service` on nova to install the new version. The announce races the restart and loses.

2. **Log permission denied.** `announce-to-river.sh` (run as `freenet-update` via sudo) cannot write `/var/log/freenet-release-agent-river.log` ("Permission denied").

## Suggested fixes

- Sequence the River announce **after** `gateway-update.yml` completes, OR make `announce-to-river.sh` wait/retry for the local node to become reachable (the node is briefly down by design during the update).
- Fix ownership/permissions of `/var/log/freenet-release-agent-river.log` on nova so the `freenet-update` user can write it.

## Impact

Low — Matrix announcement still works; only the River-room announcement is missing for affected releases.

[AI-assisted - Claude]

sanity:
## Problem

The `release-announce.yml` workflow's **"Post to River"** step failed on the last two releases — **v0.2.60** and **v0.2.61** — both with `nova /announce/river returned HTTP 500`. The Matrix announcement is unaffected. Tracked in #4208.

Root cause, confirmed from nova's journals:

- `release.published` fires `gateway-update.yml` and `release-announce.yml` **concurrently**. The gateway update restarts the local Freenet node — on the v0.2.61 release nova's `freenet-gateway.service` was down from 13:04:16 to 13:05:47.
- `announce-to-river.sh` probed the node with a single `curl --max-time 3 http://127.0.0.1:7509/` and no retry. The River announce ran at 13:04:28, inside that window, so the probe failed and the script `exit 1`'d.
- Because the script failed *fast* (sub-second), it exited inside the release-agent's 1-second early-exit probe — which the agent reports to the GitHub Actions caller as **HTTP 500**.

A second, cosmetic problem: `LOG_FILE` defaulted to `/var/log/freenet-release-agent-river.log`, which the announce user cannot write (`install.sh` deliberately keeps logs in journald). Every `log()` call emitted a spurious `Permission denied` line.

## Solution

- **Poll for node reachability** (`wait_for_node()`) instead of a single probe. The node is *expected* to bounce during a release, so waiting for it is the correct behaviour. The poll's `sleep` also keeps the script alive past the release-agent's 1-second early-exit probe, so the agent returns `202` and the announcement completes in the background once the node is up — instead of dying inside that window and surfacing as HTTP 500.
- **Default `LOG_FILE` to empty** and only append to a file when one is explicitly configured. The release-agent already captures the script's stderr to journald.

The probe is a named function with overridable parameters (`NODE_WAIT_ATTEMPTS`, `NODE_WAIT_INTERVAL`; ~5 min budget by default) — both so the budget is explicit and so the function is unit-testable.

### Trade-offs (consciously accepted)

- **Prolonged outage is reported only in journald.** Because the poll keeps the script alive past the agent's 1s probe, the agent returns `202` and the workflow goes green regardless of the eventual outcome. If the node stays down for the whole budget, the failure is logged on nova but not surfaced to the release workflow. This is inherent to the agent's fire-and-forget-after-1s design; a prolonged gateway outage on release day is independently visible. Surfacing late failures to the workflow would be a release-agent change — see the follow-up issue.
- **Duplicate-announce window.** The agent's rate limit (1/60s) is shorter than the poll budget, so a *manual* workflow re-run during the poll could spawn a second announce. The "Post to River" step does not auto-retry, so this requires deliberate operator action; the worst case is a duplicate chat line. Also noted in the follow-up issue.

## Testing

- Added **`scripts/release-agent/announce-to-river_test.sh`** — a regression test following the existing `scripts/release_state_restore_test.sh` awk-extract-and-`eval` pattern. It extracts `log()` and `wait_for_node()` verbatim and asserts: empty/­set/­unwritable `LOG_FILE` behaviour; node-up, node-down-then-up, and node-never-returns polling outcomes (10 assertions). `shellcheck`-clean; runs in <1s with `NODE_WAIT_INTERVAL=0`.
- `bash -n` + `shellcheck 0.9.0` clean on both files.
- Diagnosis verified end-to-end on nova: confirmed the gateway down-window from the `freenet-gateway.service` journal vs. the announce timestamp, and confirmed the announce succeeds when run with the node up.
- Carries the `test-exempt` label only because the Rule Lint CI check counts **Rust** `#[test]` attributes; the regression test here is a shell test, which that check does not recognise. There is no shell-test runner wired into CI yet — see the follow-up issue.

## Fixes

Closes #4208

[AI-assisted - Claude]

### DIFF
--- a/scripts/release-agent/announce-to-river.sh
+++ b/scripts/release-agent/announce-to-river.sh
@@ -26,14 +26,20 @@ ROOM_OWNER_VK="${ROOM_OWNER_VK:-4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4}"
 SIGNING_KEY_FILE="${SIGNING_KEY_FILE:-$HOME/.config/freenet-river-official/room_owner_signing_key.bin}"
 ROOMS_JSON="${ROOMS_JSON:-$HOME/.local/share/river/rooms.json}"
 FREENET_NODE_URL="${FREENET_NODE_URL:-http://127.0.0.1:7509/}"
-LOG_FILE="${LOG_FILE:-/var/log/freenet-release-agent-river.log}"
+# Optional persistent log file. Empty by default — the release-agent already
+# captures this script's stderr to journald, and the old /var/log default
+# was not writable by the announce user, so every log() call emitted a
+# spurious "Permission denied" line (issue #4208).
+LOG_FILE="${LOG_FILE:-}"
 
 log() {
     local timestamp
     timestamp=$(date -u '+%Y-%m-%dT%H:%M:%SZ')
     echo "[$timestamp] $*" >&2
-    # Best-effort persistent log; non-fatal if not writable.
-    echo "[$timestamp] $*" >> "$LOG_FILE" 2>/dev/null || true
+    # Optional best-effort persistent log; non-fatal if not writable.
+    if [[ -n "$LOG_FILE" ]]; then
+        echo "[$timestamp] $*" >> "$LOG_FILE" 2>/dev/null || true
+    fi
 }
 
 usage() {
@@ -76,9 +82,31 @@ if [[ ! -f "$ROOMS_JSON" ]]; then
     exit 1
 fi
 
-# Check that the Freenet node is reachable. riverctl will hang otherwise.
-if ! curl -s --max-time 3 "$FREENET_NODE_URL" >/dev/null 2>&1; then
-    log "ERROR: Freenet node not reachable at $FREENET_NODE_URL"
+# Wait for the Freenet node to be reachable (riverctl hangs otherwise).
+#
+# A release announcement runs concurrently with the gateway update that
+# restarts the local node, so a single probe routinely races that restart
+# window and the announce fails (issue #4208). Retry for several minutes —
+# the node is only briefly down.
+#
+# The sleep also keeps this script alive past the release-agent's 1-second
+# early-exit probe: the agent treats a sub-1s exit as a failure (HTTP 500)
+# but returns 202 for a script still running, then reaps it in the
+# background. Retrying here therefore lets the announce complete once the
+# node is back instead of dying inside that probe window.
+node_ready=false
+for attempt in $(seq 1 60); do
+    if curl -s --max-time 3 "$FREENET_NODE_URL" >/dev/null 2>&1; then
+        node_ready=true
+        break
+    fi
+    if [[ "$attempt" -eq 1 ]]; then
+        log "Freenet node not reachable at $FREENET_NODE_URL yet — waiting (it is restarted by the concurrent gateway update)"
+    fi
+    sleep 5
+done
+if [[ "$node_ready" != true ]]; then
+    log "ERROR: Freenet node still not reachable at $FREENET_NODE_URL after retries"
     exit 1
 fi
 
--- a/scripts/release-agent/announce-to-river.sh
+++ b/scripts/release-agent/announce-to-river.sh
@@ -31,6 +31,13 @@ FREENET_NODE_URL="${FREENET_NODE_URL:-http://127.0.0.1:7509/}"
 # was not writable by the announce user, so every log() call emitted a
 # spurious "Permission denied" line (issue #4208).
 LOG_FILE="${LOG_FILE:-}"
+# Node-reachability polling budget. A release announcement runs concurrently
+# with the gateway update that restarts the local node, so it must wait that
+# restart out (issue #4208). ATTEMPTS * INTERVAL is the wall-clock budget —
+# default ~5 min, comfortably over the ~90s gateway down-window observed on
+# the v0.2.61 release. Overridable (e.g. INTERVAL=0 in tests).
+NODE_WAIT_ATTEMPTS="${NODE_WAIT_ATTEMPTS:-60}"
+NODE_WAIT_INTERVAL="${NODE_WAIT_INTERVAL:-5}"
 
 log() {
     local timestamp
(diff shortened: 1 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #4209

### Summary
- **PR Title:** fix: make River release announcement survive the gateway-restart race
- **Type:** fix
- **CI Status:** running on HEAD `3afeda54` (Rule Lint green once the `test-exempt` label was applied)
- **Linked Issues:** Closes #4208; follow-up #4211
- **Review tier:** Full — release-pipeline script
- **Reviewers run:** code-first, testing, skeptical, big-picture (`freenet:*` subagents) + external Codex — **two full rounds**, plus a mutation-verified re-check of the round-2 test fix.

---

### Code-First Analysis
**Independent understanding:** `announce-to-river.sh` posts the release announcement to the River room. It now polls for the local Freenet node (`wait_for_node()`) instead of probing once, and `LOG_FILE` defaults to empty so no unwritable `/var/log` file is attempted.

**Alignment:** Matches the issue and PR description exactly. Round-2 verdict: *"Clean — all round-1 fixes correctly and completely addressed, no new issues."* `wait_for_node()`'s `(( ))` arithmetic was verified safe under `set -euo pipefail` (all in condition positions); the C-style loop bound and skip-final-sleep logic are correct.

---

### Testing Assessment
**Coverage level:** Adequate.

| Unit | ✅ | `scripts/release-agent/announce-to-river_test.sh` — extracts `log()` and `wait_for_node()` verbatim (awk + `eval`), stubs `curl`; 11 assertions, shellcheck-clean, <1s |
| Integration | N/A | — |
| E2E | ⚠️ | The PowerShell-free shell path is exercised by the unit test; the live `riverctl` send is verified by manual reproduction on nova, not CI |

**Regression test:** Present and mutation-verified — removing the `LOG_FILE` guard, reverting the `LOG_FILE` default, and reverting `wait_for_node()` to a single probe each make the suite fail. Round-1 added the test; round-2 skeptical review found the original `log()` assertion was a false-positive (it held with or without the guard); that was fixed and re-verified by mutation.

---

### Skeptical Findings
**Risk level:** Low (after fixes). No High-severity bugs in either round.

| Concern | Severity | Round | Status |
|---------|----------|-------|--------|
| Single probe races the gateway restart → HTTP 500 (the #4208 bug) | high | — | Fixed — bounded `wait_for_node()` poll |
| Magic numbers `60`/`5`; final-iteration sleep wasted | low | 1 | Fixed — named `NODE_WAIT_ATTEMPTS`/`NODE_WAIT_INTERVAL`, skip-final-sleep |
| `log()` empty-`LOG_FILE` test was a false-positive; `LOG_FILE` default untested | med | 2 | Fixed — stderr-diagnostic assertion + default-line test, both mutation-verified |
| Prolonged-outage failure invisible to the release workflow (202 already returned) | med | 1+2 | Accepted & documented (code comment, PR description, follow-up #4211) — inherent to the agent's fire-and-forget-after-1s design |
| Duplicate-announce window (rate-limit < poll budget + manual re-run) | med | 1+2 | Accepted & documented (PR description, #4211) — requires deliberate operator action; low harm |

---

### Big Picture Assessment
**Goal alignment:** Yes — resolves both root causes of #4208 (node-race + log noise).
**Anti-patterns:** None. Real root-cause fix, not CI-chasing. No removed tests/code.
**Scope:** Focused — the `wait_for_node()` extraction is the minimum needed for testability; no creep.
**Layer choice:** Retrying in the script is the right layer (the gateway update is itself async, so workflow sequencing wouldn't close the race).

---

### Documentation
- Code docs: thorough — `wait_for_node()` documents the race, the 202/HTTP-500 agent interaction, and the accepted trade-off.
- PR description: accurate; trade-offs and the `test-exempt` rationale documented.
- Follow-up #4211 tracks the deferred items (surface late failures; raise the rate-limit window; wire shell tests into CI).

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None outstanding — all addressed across the two rounds.

#### Consider (Suggestions)
1. Per follow-up #4211, wire `scripts/**/*_test.sh` into a CI job so this regression test (and the two pre-existing sibling shell tests) actually gate.

---

### Verdict
**State:** Ready to Merge (once CI is green on the current HEAD)
**HEAD SHA reviewed:** `3afeda54`

Two full review rounds plus a mutation-verified re-check converged with no blocking findings. The core fix is unanimously assessed correct; every Medium/Low was fixed or — for the two inherent design trade-offs — explicitly documented and tracked in #4211. The round-2 test-quality finding (M1/M2) was fixed test-only and mutation-verified, so a third full round was not re-run. Merge is gated only on CI going green on `3afeda54`.

[AI-assisted - Claude]

Issue #3487, opened 2026-03-09
fix: address pre-existing cache store issues found during moka migration review


iduartgomez:
## Context

During review of the stretto → moka cache migration (7732aeafc5d85a07eace6c6e2bfc16eb839ea824), several pre-existing issues were identified that were not introduced by the migration but are worth addressing.

### 1. Cache-before-persist ordering incoherence
`StateStore::store()` and `update()` insert to cache before persisting to disk. If the persist operation fails, the cache holds stale/incorrect data that will be served on subsequent reads.

**Fix:** Insert to cache only after successful persist, or invalidate on persist failure.

### 2. `remove_contract()` doesn't invalidate cache
After removing a contract from persistent storage, cached entries remain and could serve "ghost" contracts.

**Fix:** Call `cache.invalidate()` / `cache.remove()` in `remove_contract()`.

### 3. `max_size` is `i64` but used as `u64`
`max_size as u64` from `i64` — a negative `i64` would wrap to `u64::MAX`. Inherited from stretto's API convention.

**Fix:** Change the type to `u64` or add a bounds check.

### 4. Weigher `as u32` truncation
All 3 weighers cast `.len()` to `u32`. Safe today (MAX_STATE_SIZE=50 MiB, WASM modules are small), but moka recommends `.try_into().unwrap_or(u32::MAX)` as a defensive pattern.

### 5. `DelegateCode` not `Arc`-wrapped in `DelegateStore`
Moka clones values on `get()`. `ContractStore` wraps in `Arc<ContractCode>` (cheap clone) but `DelegateStore` stores `DelegateCode<'static>` directly, causing full clones on cache hits.

**Fix:** Wrap in `Arc<DelegateCode<'static>>` for cheap clones, matching the `ContractStore` pattern.

## Files involved
- `crates/core/src/contract/storages/state.rs`
- `crates/core/src/contract/storages/contract.rs`
- `crates/core/src/contract/storages/delegate.rs`

sanity:
# Issue Priority Assessment

Generated: 2026-03-12T22:23:53.758884346+00:00
Issue hash: `14526ca5`

## Value Assessment

[...]

PR #4212, opened 2026-05-22
fix: bound contract storage disk usage — hosting-cache eviction now reclaims disk

sanity:
## Problem

Freenet nodes had **no limit on disk space used for contract storage**. Contract WASM code (`contracts/`) and contract state (`db/`) accumulated permanently — every contract whose data ever passed through a node stayed on disk forever.

The `HostingCache` already has a byte budget and LRU eviction, but eviction was a **metadata-only** operation: when a contract was evicted, the node dropped the in-memory tracking entry and the ReDb `hosting_metadata` row, but **never deleted the contract's state or WASM code from disk**. `ContractStore` even carried a standing TODO acknowledging that disk-bounded eviction was unimplemented.

Net effect: the hosting budget bounded the *in-memory hosted set* but not actual disk usage, which grew without bound.

A second, latent bug found along the way: `ContractStore::remove_contract` deleted the on-disk `.wasm` blob unconditionally. The blob is keyed by code hash and shared across contract instances (every River room shares one room-contract WASM), so removing one instance could break the others on a cache miss / after restart.

## Approach

Make hosting-cache eviction actually reclaim disk, and make the budget operator-configurable.

1. **Storage delete APIs.** Added `StateStorage::remove` (ReDb, SQLite, mock backends) and `StateStore::delete` — deletes persisted state + params and invalidates the moka cache. Made `ContractStore::remove_contract` code-hash refcount-safe: the shared `.wasm` blob is deleted only when no remaining contract instance references that code hash, decided from the **persistent ReDb `contract_index`** (the source of truth shared across all pool executors — a per-executor in-memory index would miss siblings stored via a different executor). File removal is idempotent.

2. **Configurable budget.** New `max-hosting-storage` config option (CLI / env / TOML), default **1 GiB**. Threaded into `HostingCache`.

3. **Eviction → reclamation wiring.** New fire-and-forget `ContractHandlerEvent::EvictContract`, an executor `reclaim_contract_storage` path, and a `reclaim_evicted_contract` helper wired into all three eviction sites (GET cache, PUT cache, periodic hosting sweep). `EvictContract` is routed through the fair queue per-contract.

**In-use contracts are never deleted.** A contract with an active client subscription, downstream peer subscriber, or upstream network subscription (`contract_in_use`) is exempt from eviction-driven deletion. This exemption is applied in three places so it is airtight: `HostingCache::record_access` and `sweep_expired` will not *evict* an in-use contract from the cache at all (so it is never orphaned on disk); `reclaim_evicted_contract` re-checks the gate before emitting an event; and `RuntimePool::remove_contract` re-checks once more at deletion time, skipping the delete if the contract was re-hosted or re-subscribed in the window since eviction.

**State-write generation token.** Re-subscription and re-hosting checks alone are insufficient against a tighter race: a contract can be re-PUT or UPDATEd in the window between eviction and the executor processing the EvictContract event. To close that, every state write bumps a per-contract monotonic counter (`HostingManager::bump_state_generation`), the current value is captured atomically in `HostedContract.write_generation` at hosting-cache record/refresh time, the snapshot is carried through `EvictContract`, and the deletion-time guard in `RuntimePool::remove_contract` re-reads the current generation and skips reclamation on any mismatch. The bump+refresh pair runs from FOUR executor chokepoints (`commit_state_update` PUT/UPDATE, the store-initial-state path, and `verify_and_store`) AND from the V2 delegate write path (`put_contract_state_sync` / `update_contract_state_sync`) via a callback installed on `Runtime` so the V2 path is also covered. The `is_hosting_contract` / `contract_in_use` re-checks at deletion time remain as defense in depth.

## Scope of the bound

The budget tracks **contract state bytes** — the dominant consumer. WASM code blobs and database (ReDb/SQLite) overhead are reclaimed on eviction but not counted toward the budget, so total on-disk use can modestly exceed the configured value. Eviction reclamation is also best-effort (fire-and-forget): under sustained load an eviction event can be dropped and retried on the next cycle. The budget is therefore a soft target on tracked state, not a hard ceiling on every byte.

## Operator note

The enforced default rises from a previously-hardcoded 100 MiB to 1 GiB **and** eviction now actually deletes disk. A node upgrading with more than `max-hosting-storage` of cached contract state will, on first run, evict and delete down to the budget — contracts it is not actively using will be shed and re-fetched from the network on next access. Contracts the node has a subscription to are retained.

## Testing

New unit tests covering:
- **Storage layer** — `StateStorage::remove` (ReDb + SQLite); `StateStore::delete` idempotency; refcount-safe `ContractStore::remove_contract`, including `test_remove_contract_keeps_shared_wasm_across_executors` (two `ContractStore`s sharing one DB — proves the cross-executor case).
- **Hosting cache** — `record_access` honors the retain predicate (in-use contracts not evicted); `contract_in_use` (client / downstream / network subscription, reclaimable once all leave).
- **State-write generation token** — `test_record_access_carries_write_generation_through_eviction`, `test_record_contract_access_captures_then_diverges_from_current_generation`, `test_state_generation_lifecycle`, `test_record_access_refresh_updates_write_generation`, `test_refresh_cache_generation_is_noop_when_entry_absent`.
- **V2 delegate write path** — `test_env_put_contract_state_fires_write_callback`, `test_env_update_contract_state_fires_write_callback`, `test_env_update_contract_state_failure_does_not_fire_callback`.
- **Config** — `max-hosting-storage` default + explicit-value round-trip.
- **Executor** — `remove_contract` reclaims state + WASM from disk; idempotent double-eviction.

Full `cargo test -p freenet --lib`: passing. `cargo clippy --lib --tests` (incl. `--features sqlite`): clean.

Three rounds of multi-perspective + external-model review found and fixed:
- **Round 1** — four deterministic data-loss bugs (refcount-check race, eviction/deletion TOCTOU, orphaned disk leak for in-use contracts, stale TODO).
- **Round 2** — the eviction/re-host TOCTOU (closed via the state-write generation token described above).
- **Round 3** — the V2 delegate write path bypassing the executor's bump+refresh chokepoints, and a permanent disk leak introduced by the round-2 fix on the UPDATE-then-evict case (the cache snapshot never refreshed after the first `record_access`, so every UPDATE-then-evict leaked).

Each round's review was re-run on the result.

Follow-ups (filed, not blocking):
- #4216 — concurrent store/remove on a shared code hash can delete a freshly-written WASM blob (pre-existing).
- #4217 — end-to-end `#[freenet_test]` driving the cache over budget.
- #4218 — pool executors have divergent per-instance `ContractStore::key_to_code_part` maps. The original "ghost" cache-hit symptom mentioned in earlier rounds is rooted in this pre-existing per-executor map divergence; the index-membership guard added during review round 3 to mask it was reverted in this PR because it regressed cross-executor fetches, and the divergence is now documented at the top of `fetch_contract`.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/contracts.md
+++ b/.claude/rules/contracts.md
@@ -110,10 +110,11 @@ Pattern (logic in native_api.rs):
 USE: StateStore trait
 Backends: redb (default), sqlite
 
-Operations:
-  - get_state(key) → Option<State>
-  - put_state(key, state) → Result
-  - delete_state(key) → Result
+Operations (StateStore methods):
+  - get(key) → WrappedState
+  - store(key, state, params) → Result
+  - update(key, state) → Result
+  - delete(key) → Result
 
 MUST:
   - Handle missing state gracefully (new contract)
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -63,6 +63,11 @@ const DEFAULT_TRANSIENT_BUDGET: usize = 2048;
 const DEFAULT_TRANSIENT_TTL_SECS: u64 = 30;
 const DEFAULT_EVENT_LOOP_CHANNEL_CAPACITY: usize = 2048;
 
+/// Default operator-facing budget for hosted contract storage (1 GiB).
+/// Must stay consistent with `ring::hosting::cache::DEFAULT_HOSTING_BUDGET_BYTES`,
+/// which is the in-code fallback used by tests.
+const DEFAULT_MAX_HOSTING_STORAGE: u64 = 1024 * 1024 * 1024; // 1 GiB
+
 const QUALIFIER: &str = "";
 const ORGANIZATION: &str = "The Freenet Project Inc";
 const APPLICATION: &str = "Freenet";
@@ -102,6 +107,12 @@ pub struct ConfigArgs {
     /// Default: 2x CPU cores, clamped to 4-32.
     #[arg(long, env = "MAX_BLOCKING_THREADS")]
     pub max_blocking_threads: Option<usize>,
+
+    /// Maximum disk space in bytes for hosted contract storage. Once exceeded,
+    /// contracts are evicted oldest-first (LRU) and their on-disk state
+    /// reclaimed. Default: 1 GiB.
+    #[arg(long, env = "MAX_HOSTING_STORAGE")]
+    pub max_hosting_storage: Option<u64>,
 
     #[command(flatten)]
     pub telemetry: TelemetryArgs,
@@ -150,6 +161,7 @@ impl Default for ConfigArgs {
             id: None,
             version: false,
             max_blocking_threads: None,
+            max_hosting_storage: None,
             telemetry: Default::default(),
         }
     }
@@ -354,6 +366,8 @@ impl ConfigArgs {
                 self.network_api.bbr_startup_rate = cfg.network_api.bbr_startup_rate;
             }
             self.log_level.get_or_insert(cfg.log_level);
+            self.max_hosting_storage
+                .get_or_insert(cfg.max_hosting_storage);
             self.config_paths.merge(cfg.config_paths.as_ref().clone());
             // Merge telemetry config - CLI args override file config
             // Note: enabled defaults to true via clap, so we only override
@@ -695,6 +709,9 @@ impl ConfigArgs {
             max_blocking_threads: self
                 .max_blocking_threads
                 .unwrap_or_else(default_max_blocking_threads),
+            max_hosting_storage: self
+                .max_hosting_storage
+                .unwrap_or(DEFAULT_MAX_HOSTING_STORAGE),
             telemetry: TelemetryConfig {
                 enabled: self.telemetry.enabled,
                 endpoint: self
@@ -801,6 +818,14 @@ pub struct Config {
     /// Maximum number of threads for blocking operations (WASM execution, etc.).
     #[serde(default = "default_max_blocking_threads")]
     pub max_blocking_threads: usize,
+    /// Maximum disk space in bytes for hosted contract storage. Once exceeded,
+    /// contracts are evicted oldest-first (LRU) and their on-disk state
+    /// reclaimed.
+    #[serde(
+        default = "default_max_hosting_storage",
+        rename = "max-hosting-storage"
+    )]
+    pub max_hosting_storage: u64,
     /// Telemetry configuration
     #[serde(flatten)]
     pub telemetry: TelemetryConfig,
@@ -811,6 +836,11 @@ fn default_max_blocking_threads() -> usize {
     std::thread::available_parallelism()
         .map(|n| (n.get() * 2).clamp(4, 32))
         .unwrap_or(8)
+}
+
+/// Default operator-facing budget for hosted contract storage (1 GiB).
+fn default_max_hosting_storage() -> u64 {
+    DEFAULT_MAX_HOSTING_STORAGE
 }
 
 impl Config {
@@ -2702,6 +2732,50 @@ mod tests {
         let cfg = args.build().await.unwrap();
         let serialized = toml::to_string(&cfg).unwrap();
         let _: Config = toml::from_str(&serialized).unwrap();
+    }
+
+    #[tokio::test]
+    async fn max_hosting_storage_defaults_to_one_gib() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let args = ConfigArgs {
+            mode: Some(OperationMode::Local),
+            config_paths: ConfigPathsArgs {
+                config_dir: Some(temp_dir.path().to_path_buf()),
+                data_dir: Some(temp_dir.path().to_path_buf()),
+                log_dir: Some(temp_dir.path().to_path_buf()),
+            },
+            ..Default::default()
+        };
+        let cfg = args.build().await.unwrap();
+        assert_eq!(
+            cfg.max_hosting_storage, DEFAULT_MAX_HOSTING_STORAGE,
+            "default max_hosting_storage should resolve to 1 GiB"
+        );
+        assert_eq!(DEFAULT_MAX_HOSTING_STORAGE, 1024 * 1024 * 1024);
+    }
+
+    #[tokio::test]
+    async fn max_hosting_storage_explicit_value_round_trips() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let custom = 256 * 1024 * 1024_u64;
+        let args = ConfigArgs {
+            mode: Some(OperationMode::Local),
+            config_paths: ConfigPathsArgs {
+                config_dir: Some(temp_dir.path().to_path_buf()),
+                data_dir: Some(temp_dir.path().to_path_buf()),
+                log_dir: Some(temp_dir.path().to_path_buf()),
+            },
+            max_hosting_storage: Some(custom),
+            ..Default::default()
+        };
+        let cfg = args.build().await.unwrap();
+        assert_eq!(cfg.max_hosting_storage, custom);
+
+        // Round-trips through TOML serialization.
+        let serialized = toml::to_string(&cfg).unwrap();
+        assert!(serialized.contains("max-hosting-storage"));
+        let deserialized: Config = toml::from_str(&serialized).unwrap();
+        assert_eq!(deserialized.max_hosting_storage, custom);
     }
 
     /// Build a minimal local-mode ConfigArgs with the given CIDR list and
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -63,11 +63,6 @@ const DEFAULT_TRANSIENT_BUDGET: usize = 2048;
 const DEFAULT_TRANSIENT_TTL_SECS: u64 = 30;
 const DEFAULT_EVENT_LOOP_CHANNEL_CAPACITY: usize = 2048;
 
-/// Default operator-facing budget for hosted contract storage (1 GiB).
-/// Must stay consistent with `ring::hosting::cache::DEFAULT_HOSTING_BUDGET_BYTES`,
-/// which is the in-code fallback used by tests.
-const DEFAULT_MAX_HOSTING_STORAGE: u64 = 1024 * 1024 * 1024; // 1 GiB
-
 const QUALIFIER: &str = "";
 const ORGANIZATION: &str = "The Freenet Project Inc";
 const APPLICATION: &str = "Freenet";
@@ -108,9 +103,11 @@ pub struct ConfigArgs {
     #[arg(long, env = "MAX_BLOCKING_THREADS")]
     pub max_blocking_threads: Option<usize>,
 
-    /// Maximum disk space in bytes for hosted contract storage. Once exceeded,
-    /// contracts are evicted oldest-first (LRU) and their on-disk state
-    /// reclaimed. Default: 1 GiB.
+    /// Budget in bytes for hosted contract *state*. Once exceeded, contracts
+    /// are evicted (least-valuable-first) and their on-disk state reclaimed.
+    /// This bounds tracked contract state only — WASM code blobs and ReDb/
+    /// SQLite database overhead are additional and not counted against it.
+    /// Default: 1 GiB.
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
 
@@ -711,7 +708,7 @@ impl ConfigArgs {
                 .unwrap_or_else(default_max_blocking_threads),
             max_hosting_storage: self
                 .max_hosting_storage
-                .unwrap_or(DEFAULT_MAX_HOSTING_STORAGE),
+                .unwrap_or(crate::ring::DEFAULT_HOSTING_BUDGET_BYTES),
             telemetry: TelemetryConfig {
                 enabled: self.telemetry.enabled,
                 endpoint: self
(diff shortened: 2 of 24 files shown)


sanity (commented):
## Comprehensive PR Review: #4212

### Summary
- **PR Title:** fix: bound contract storage disk usage — hosting-cache eviction now reclaims disk
- **Type:** fix (+ feat: configurable budget)
- **CI Status:** pending at review time
- **Linked Issues:** none formally linked
- **Review tier:** Full (high-risk surfaces: on-disk data deletion, persistence, contract-handler event system)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four) + external Codex pass

This review found **real bugs**, including one data-loss path independently flagged by three reviewers. Verdict: **Needs Changes — Re-review Required After Fix.**

---

### Code-First Analysis
**Independent understanding matches stated intent** — storage delete APIs, refcount-safe `ContractStore::remove_contract`, configurable budget, and `EvictContract`-driven reclamation wired into all three eviction sites. The description is accurate and unusually candid (it self-discloses the eviction/reclaim race). No intent-vs-implementation contradiction; the findings below are correctness bugs and accuracy gaps, not misrepresentation.

---

### Testing Assessment
**Coverage:** good at the storage-primitive and executor-reclaim layers; gaps at the wiring junctions.

| Unit | ⚠️ | Storage primitives, refcount safety, executor disk-reclaim all covered. Untested: the `EvictContract` handler arm, the `reclaim_evicted_contract` gate decision, SQLite `remove`, the multi-executor refcount path (see Must-Fix #1). |
| Integration | ❌ | No `#[freenet_test]` drives the cache over budget end-to-end. |
| Simulation | N/A | |

**Regression test:** present for the storage layer; **missing** for the multi-executor topology where Must-Fix #1 actually lives.

---

### Skeptical Findings
**Risk level: high.**

| Refcount scan reads per-executor in-memory index, not the shared ReDb index | **High** | `wasm_runtime/contract_store.rs:199-202` | Can delete a `.wasm` blob still referenced by an instance stored on another pool executor — the exact shared-WASM bug the PR claims to fix. Flagged by Codex, code-first, AND skeptical. |
| Re-host race: stale `EvictContract` deletes freshly re-stored state | **High** | `contract.rs` EvictContract arm | Per-contract FIFO serialization doesn't fix ordering; a `PutQuery` processed before a queued `EvictContract` re-stores then gets deleted → contract in cache, absent on disk. |
| Subscription-gate TOCTOU | **High** | `operations.rs:184` + handler arm | Gate checked at enqueue; no re-check at deletion. A subscription created in the window → in-use contract's state/code deleted. |
| In-use eviction never reclaimed later | **High** | `operations.rs:184-192` | When `record_access` evicts an in-use contract, reclamation is (correctly) skipped — but the contract is also gone from the cache, so nothing revisits it. After its last subscriber leaves, disk is orphaned permanently; the budget leaks exactly the protected contracts. |
| Best-effort partial reclaim leaves half-present contract | Medium | `runtime.rs` `reclaim_contract_storage` | State-gone/code-kept (or vice versa) reported as success. Mostly self-heals via re-fetch; doc comment should be honest. |
| WASM blob bytes not counted in the budget | Medium | `ring/hosting/cache.rs` | `max-hosting-storage` bounds contract **state** only; `.wasm` blobs + DB overhead are additional. The "bound disk usage" claim overstates. |
| Sweep reclamation silently skipped if op_manager weak ref dead | Low | `ring.rs` sweep | Disk leaked with no warning logged. |

---

### Big Picture Assessment
**Goal alignment: partial.** The mechanism is sound and the shared-WASM refcount fix is a correct in-scope prerequisite (not creep). But the budget tracks contract **state bytes only** — the title/description claim to "bound disk usage" overstates: WASM code blobs, ReDb/SQLite overhead, and eviction lag all sit outside the ceiling.

**Anti-patterns:** the subscription-based exemption is mostly time-bounded (downstream leases expire; network subscriptions are leased) — but **client subscriptions have no TTL** (removed only on explicit unsubscribe/disconnect). Per `AGENTS.md` cleanup-exemption rule this is defensible (a connected client genuinely is using the contract) but must be documented as intentional.

**Stale docs introduced/missed:**
- `contract_store.rs:21-22` — the standing `// TODO: add functionality to delete old contracts ... under a configured threshold` describes exactly what this PR implements. Must be removed/updated.
- `.claude/rules/contracts.md` "State Management Rules" lists a nonexistent `delete_state` API; the PR adds the real `StateStore::delete`. Update it.

**Removed code:** none of concern — the #3487 ghost-contract `contract_cache.invalidate()` is preserved and correctly relocated inside the `!code_still_referenced` block.

---

### Documentation
- Code docs: mostly complete; `remove_contract` and `contract_in_use` doc comments need to state their actual scope/semantics (see Should-Fix).
- Two 1-GiB constants (`DEFAULT_MAX_HOSTING_STORAGE`, `DEFAULT_HOSTING_BUDGET_BYTES`) must stay equal, enforced only by a comment — drift would not be caught. Make one the single source.
- No CHANGELOG / release note for the default raise (100 MiB → 1 GiB) — nodes upgrading with >1 GiB cached state will evict down on first run; operators should be told.

---

### Recommendations

#### Must Fix (Blocking)
1. **Refcount check against the persistent index.** `ContractStore::remove_contract` must decide `code_still_referenced` from the shared ReDb `contract_index` (`load_all_contract_index`), not the per-executor in-memory `key_to_code_part`. Add a regression test with two instances of one code hash stored on **different** pool executors.
2. **Re-check at deletion time.** The `EvictContract` handler must re-verify the contract is still evictable — skip the delete if it has been re-hosted (`is_hosting_contract`) or is now in use (`contract_in_use`). Fixes the re-host race and the subscription-gate TOCTOU.
3. **Don't orphan in-use contracts.** Make `HostingCache::record_access` honor a retain predicate (as `sweep_expired` already does) so in-use contracts are not evicted from the cache at all — otherwise their disk data leaks permanently once skipped. They become reclaimable normally once no longer in use.
4. **Remove the stale TODO** at `contract_store.rs:21-22` — the PR implements it.

#### Should Fix (Important)
1. Include `is_subscribed` (active upstream network subscription) in `contract_in_use`, and use one predicate uniformly across `record_access`, `sweep_expired`, and the reclaim gate.
2. Soften the over-claim: config help text, doc comments, and PR description should say the budget bounds tracked contract **state**; WASM code and DB overhead are additional.
3. Update `.claude/rules/contracts.md` to the real `StateStore` API (`get`/`store`/`delete`).
4. Single source of truth for the 1-GiB default constant.
5. Add tests: `EvictContract` handler arm invokes reclaim; `reclaim_evicted_contract` gate emits/skips correctly; SQLite `remove`.
6. Add a release note for the default change (nodes shed cached contracts on upgrade).

#### Consider
- Honest doc comment on `reclaim_contract_storage` partial-failure semantics; log a warning when the sweep skips reclamation due to a dead op_manager ref; document that reclamation is best-effort under fair-queue pressure; `max_hosting_storage = 0` boundary test.
- File a follow-up issue for an end-to-end `#[freenet_test]` driving the cache over budget.

---

### Verdict
**HEAD SHA reviewed:** 4e0aa2a97c8b4d45cc2b1288ac4888cb05d43027

Four Must-Fix items, two of them data-loss paths on a high-risk surface (Must-Fix #1 was caught independently by three reviewers). A full re-review on the updated HEAD is required before merge.

[AI-assisted - Claude]

sanity (commented):
## Re-Review (Round 2): PR #4212

**HEAD reviewed:** `4516666c` · **Reviewers:** code-first, testing, skeptical, big-picture + external Codex · **Tier:** Full

### Round-1 findings — all resolved ✅

| Round-1 finding | Status |
|---|---|
| Refcount scan used per-executor index (data loss) | ✅ Fixed `4e5d1df1` — now scans the shared ReDb `contract_index`; fails safe on read error; regression test `test_remove_contract_keeps_shared_wasm_across_executors` reproduces the bug |
| Re-host / re-subscribe TOCTOU | ⚠️ Partially — see Round-2 #1 |
| In-use contracts orphaned on disk | ✅ Fixed `9196fbdf` — `record_access` honors the retain predicate; in-use contracts are never evicted |
| Stale TODO, over-claim, stale `contracts.md`, drifting constants, missing upgrade note | ✅ All fixed (`5103a21a`, `e64c0493`, PR description) |

`contract_in_use` now includes `is_subscribed` and is the single retain predicate across `record_access`, `sweep_expired`, and the reclaim gate. Big-picture: no scope creep, no removed tests, cleanup exemptions are time-bounded.

### Round-2 findings

**1. (Should Fix) `EvictContract` re-host race not fully closed.** `RuntimePool::remove_contract` re-checks `is_hosting_contract || contract_in_use` — but the re-host (`host_contract` / `record_get_access`) runs on the GET/PUT *driver* task *after* the state-write event completes, while `EvictContract` runs on the *handler*. A contract evicted, then concurrently re-PUT, can have its freshly-written state deleted if the handler processes `EvictContract` in the gap before the driver re-hosts. *Reviewers disagreed on severity (skeptical: blocking; code-first: acceptable for fire-and-forget).* Reconciled: the race is real, but the common-case consequence is a self-healing transient inconsistency (next access re-fetches from the network); genuine loss is limited to a contract not yet propagated. Subscribed contracts (e.g. all River rooms) are unaffected. A complete fix requires performing the hosting-cache update in the same serialized handler context as the state write.

**2. (Should Fix — Codex) `fetch_contract` ghost for shared-code instances.** `fetch_contract` checks the code cache by code-hash *before* the instance index, so a removed instance whose code hash is still cached by a sibling stays fetchable. Fix: gate `fetch_contract` on `key_to_code_part` membership.

**3. (Medium) Residual concurrent store/remove race.** `remove_contract`'s `load_all_contract_index()` scan can race a concurrent `store_contract` on a sibling instance (blob written, index not yet committed) on another pool executor. Narrow; recommend a tracked follow-up.

**4. (Low) Doc/observability:** `record_access` doc's "age-0 ⇒ never self-evicted" is false for `min_ttl == 0`; `max_hosting_storage` is unvalidated (`0` is a footgun); `Ring::contract_in_use` doc and `reclaim_evicted_contract`'s skip-log omit the network-subscription term added in `9196fbdf`.

### Testing
All four round-1 Must-Fix items now have regression tests except the `RuntimePool::remove_contract` re-check guard — acceptable **only as a filed, linked follow-up issue** (the underlying `Ring` predicates are unit-tested). Full `cargo test -p freenet --lib`: 2565 passed, 0 failed. Clippy clean.

### Verdict
**Needs Changes — Re-review Required After Fix.** The deterministic data-loss bugs from round 1 are genuinely fixed. Remaining: one narrow concurrency race (#1) whose complete fix is a driver refactor, a small `fetch_contract` correctness fix (#2), and minor doc/observability items. #2 and #4 are quick; #1 and #3 are a scope decision for the maintainer.

[AI-assisted - Claude]

sanity (commented):
## Re-Review (Round 3): PR #4212

**HEAD reviewed:** `7ef6f855` · **Reviewers:** all four + external Codex · **Tier:** Full

### Round-2 findings status

| Finding | Status |
|---|---|
| `EvictContract` re-host race | ⚠️ **Partially resolved** — see Round-3 #2 |
| `fetch_contract` shared-code ghost (Codex P2) | ⚠️ **Resolved but with regression** — see Round-3 #1 |
| Doc/log staleness (network sub) | ✅ Fixed |

### Round-3 findings

**1. (Blocking — Codex) The new `fetch_contract` index guard regresses cross-executor fetches.** `key_to_code_part` is per-`ContractStore` (a fresh `Arc<DashMap>` per `ContractStore::new`); each pool executor has its own. Pre-guard, cross-executor fetches worked via the `contract_cache` (moka, code-hash-keyed) fast path even when the popped executor had never seen the instance. My guard at `contract_store.rs:98-100` makes the local map authoritative, so a contract stored via executor A is reported missing when `fetch_contract` is called on executor B. This regresses the common pooled path, not just the narrow ghost case the guard was meant to close. The ghost itself is a symptom of pre-existing per-executor `key_to_code_part` divergence and should be fixed at that layer (shared map across the pool), not at `fetch_contract`.

**2. (Blocking — Skeptical) V2 delegate state writes bypass `bump_state_generation`.** `store_state_sync` / `update_state_sync` (`storages/redb.rs:363`, `:383`) write directly to ReDb from V2 delegate host functions (`wasm_runtime/native_api.rs:569`, `:592`), bypassing the executor's `state_store` chokepoint. The four bump sites the generation-token commit added all live in `Executor<Runtime>::state_store.{store,update}` callers — none cover the V2 delegate path. So for delegate-written state, the re-host race remains: `EvictContract` can still delete freshly delegate-written state because `current_generation == expected_generation` (neither side ever bumped).

**3. (Should Fix — Skeptical) Generation token introduced a new permanent disk leak for UPDATE/merge-PUT → evict.** `bump_state_generation` runs on every state write, but the cache's `HostedContract.write_generation` snapshot is only refreshed on a fresh `record_contract_access` call. The `host_contract` call in `put/op_ctx_task.rs:1237` is gated on `!was_hosting`, and `commit_state_update` (UPDATE) never refreshes the snapshot at all. So after any UPDATE or re-PUT to an already-hosted contract, the snapshot diverges from current_generation permanently. If that contract later loses all interests and the sweep evicts it, the deletion-time guard sees the mismatch and SKIPS reclamation — the on-disk state and code are leaked forever. The fix trades round-2's "wrongly-deleted-state" race for a new "wrongly-leaked-state" race; the latter is less acute but still erodes the "real disk bound" claim.

**4. (Minor) Doc gap.** PR description doesn't link follow-up issues #4216/#4217 nor mention the generation token or `fetch_contract` guard. Mechanical to fix.

### What IS solid

- All round-1 deterministic data-loss bugs (per-executor refcount via ReDb scan, in-use orphans, `contract_in_use` predicate uniformity, stale TODO/docs) are genuinely resolved with strong unit-test coverage.
- The four executor-side state-write bump sites are correct and well-documented.
- Generation capture inside the hosting-cache write lock is structurally correct.
- The doc/comment trail at the code level is high-quality.

### Verdict

**Needs Changes / Discussion.** The round-3 fixes closed one race but opened two (Codex #1, Skeptical #3) and missed one write path (Skeptical #2). The pattern the skeptical reviewer warned about in round 2 — "fix-then-fix-the-fix" cleanup bugs in this code shape — is materializing. The remaining issues are concurrency races whose complete closure requires architectural changes that keep expanding the PR's surface area: a shared `key_to_code_part` across pool executors, generation bumps at every state-write path including the V2 delegate hooks, and refresh of the cache's write-generation snapshot on every write (not just on a first-time `host_contract`). Each adds more cross-cutting code on an already-large PR.

[AI-assisted - Claude]

Issue #3851, opened 2026-04-13
Client-initiated SUBSCRIBE via op_ctx_task never installs local lease


sanity:
## Observation

Discovered during review of #3850. Since PR #3806 (#1454 Phase 2b) migrated client-initiated SUBSCRIBE to the task-per-tx driver at \`crates/core/src/operations/subscribe/op_ctx_task.rs\`, the \`ReplyClass::Subscribed\` branch (\`op_ctx_task.rs:437\`) publishes the result to the client but does **not** call any of the originator-side finalization work that the legacy \`SubscribeOp::process_message\` originator branch performs:

- \`op_manager.ring.subscribe(*key)\` — install the lease in \`active_subscriptions\`
- \`op_manager.ring.complete_subscription_request(key, true)\` — clear pending backoff state
- \`crate::node::network_status::record_subscription(...)\` — dashboard/telemetry signal
- \`fetch_contract_if_missing(...)\` — fetch the contract body if we don't have it
- \`super::announce_contract_hosted(op_manager, key).await\` — announce to neighbors for proximity-cache UPDATE routing
- Upstream interest registration (\`register_peer_interest(..., is_upstream=true)\`) — records where to send future Unsubscribe messages
- \`interest_manager.add_local_client(key)\` + \`broadcast_change_interests\` — ensures incoming \`ChangeInterests\` are processed

## Consequences

For a client-initiated subscribe (the primary user-facing path — River, Ghostkey, ping, etc.):

- \`contracts_needing_renewal()\` never picks up the contract → **no renewal** → the subscription silently dies when the upstream lease TTL elapses (~8 minutes), even though the client thinks it is subscribed.
- \`has_local_interest()\` is false → incoming \`ChangeInterests\` for this contract are rejected, breaking dynamic interest propagation.
- Dashboard / \`network_status::subscribed_contracts\` undercounts.
- On the originator, UPDATEs arrive but the contract body may not be present locally until an UPDATE happens to carry full state.

Renewal-driven subscribes and PUT sub-operations still go through the legacy \`SubscribeOp::process_message\` path and are unaffected.

## Scope relative to #3850

#3850 fixes the *relay-side* half of the subscription model by moving these same calls out of the relay branch and into the originator-only branch of the legacy handler. That PR does not make this issue worse, but it makes the asymmetry more visible: the originator finalization now lives cleanly in one place (the legacy handler), while the task-per-tx driver still lacks it.

[...]

PR #4089, opened 2026-05-11
refactor: retire phase-5 dead-code scaffolding (#1454 phase 6)

iduartgomez:
## Problem

Phase 5 final of #1454 retired the legacy state machine for GET / PUT /
UPDATE / SUBSCRIBE but kept the supporting scaffolding under
`#[allow(dead_code)]` so the inline pin tests would keep compiling.
That scaffolding (~4900 LoC) has no production reader anymore — the
task-per-tx drivers in `*/op_ctx_task.rs` own all routing state in task
locals — and the pin tests that reference it are tautologies once the
struct is gone (e.g. "this enum variant is gone" auto-passes if the
enum is gone too).

## Solution

One commit per op slice, four slices total, plus a docs commit:

1. **GET slice** — drop `GetOp` / `GetState` / `GetStats` /
   `PrepareRequestData` / `AwaitingResponseData` / `FinishedData`,
   `TryFrom<GetOp> for GetResult`, inline outcome / failure-routing /
   relay-cache-decision pin tests, the unused `GetResult.key` field
   (and the `reply_key` aggregation in the sub-op driver that existed
   only to feed it). Keep wire-format types + `GetResult` (state +
   contract only).
2. **PUT slice** — drop `PutOp` / `PutState` / `PutStats` /
   `AwaitingResponseData` / `FinishedData`, 22 inline pin tests, and
   the 4 anti-regression pins for retired speculative-retry fields
   (#3964; those pins become tautologies once the struct is gone).
   Keep wire-format types, the `op_state_manager.rs` GC pins, the
   `put_forwarding_ack_senders_must_stay_deleted` source-grep pin,
   and the originator finalization + local-store helpers consumed by
   the driver.
3. **UPDATE slice** — drop `UpdateOp` / `UpdateState` / `UpdateStats` /
   `FinishedData`, inline pin tests, and the two stale
   `process_message`-anchored pin tests (driver-side pins at
   `update/op_ctx_task.rs:2063` / `:2345` already cover the
   `send_summary_back_on_rejection` invariant for the surviving
   driver). Keep wire-format types, `BroadcastDedupCache`,
   `BroadcastTargetResult`, `OpManager::try_auto_fetch_contract`, the
   `update_contract` + `log_*_failure` helpers, the post-merge
   propagation helpers + their pin tests, and the `log_severity`
   regression suite (#3914).
4. **SUBSCRIBE slice** — drop `SubscribeOp` / `SubscribeState` /
[...]

Issue #3944, opened 2026-04-22
Flaky test: relay_does_not_accept_after_forwarding_on_retry


sanity:
## Observed

Running \`cargo test -p freenet --lib\` intermittently fails with:

\`\`\`
test operations::connect::tests::relay_does_not_accept_after_forwarding_on_retry ... FAILED
\`\`\`

On re-run the test passes. Isolated execution (\`cargo test ... relay_does_not_accept_after_forwarding_on_retry\`) also passes.

## Likely root cause

\`crates/core/src/operations/connect.rs:3856\` uses \`Location::random()\` for \`desired_location\`. Commit 3ed7ae32 fixed the analogous flake in \`relay_does_not_accept_when_not_at_terminus\` by noting:

> Used Location::random() which could land near self_loc under parallel execution, triggering near-terminus probabilistic acceptance. Replace with next_hop's location to guarantee the relay is never at terminus.

The same pattern likely applies here — unverified, but worth starting the investigation there.

## Fix sketch

Either use a deterministic \`desired_location\` that guarantees the test's topology assumption, or reseed \`GlobalRng\` deterministically in the test so the location is reproducible across runs.

## Context

Flagged while landing PR #3942 (unrelated web-contract caching fix) — my tests pass cleanly in isolation but this one flaked once out of ~10 full-suite runs. Filing as a follow-up so it doesn't stay lost in CI noise.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:53:25.119922576+00:00
Issue hash: `63ef847a`

## Value Assessment

**Assessment of Issue #3944**

**Relevance Check:** The issue is still valid and unresolved. The analogous fix was applied correctly to `relay_does_not_accept_when_not_at_terminus` (line 2282), which now uses `next_hop.location()` with an explanatory comment referencing the prior flake. However, `relay_does_not_accept_after_forwarding_on_retry` at line 2762 still uses `Location::random()` for `desired_location` — exactly the pattern the issue identifies as the root cause. The fix is a one-line change and the correct substitution (`next_hop.location()`) is already demonstrated in the adjacent test.

[...]

PR #4224, opened 2026-05-24
fix(subscribe): fetch contract body + announce hosted on originator subscribe

### LINKED ISSUE #4223: fix: peers that successfully subscribed sometimes return get_not_found on the same contract
## Problem

Telemetry from the last 24h shows that on contracts with active subscribers, **37 % of failing GETs reach one of the subscriber peers and that subscriber still returns `get_not_found`**. Under the expected \"subscribers host the state so they can validate updates\" semantics, this should not happen.

| Failure pattern on subscribed contracts | Count | % |
|---|---:|---:|
| **GET visited a subscriber, still got NF** | **63** | **37 %** |
| GET never visited any subscriber | 109 | 63 % |

(The 63 % bucket is tracked separately in #4222 as a routing issue.)

The contracts in this set are not undersubscribed: median 31 subscribers, 30 of the 38 affected contracts have >10 subscribers. 37 of the 38 also had *successful* GETs in the same window, so the data exists on the network and is sometimes findable.

## Likely root cause: #3851

The driver migration in #3806 (#1454 Phase 2b) moved client-initiated SUBSCRIBE to a task-per-tx path at `crates/core/src/operations/subscribe/op_ctx_task.rs`. The `ReplyClass::Subscribed` branch publishes the result to the client but **does not call `fetch_contract_if_missing(...)`** (it's listed as one of the missing originator-side finalization steps in #3851).

If the subscribing peer doesn't yet have the contract body locally:
- `subscribe_success` telemetry fires (so our analysis correctly classifies the peer as a subscriber).
- A GET arriving shortly after subscribe finds the peer registered as a subscriber but with no local state — so the peer correctly answers `get_not_found`.

This matches what we see: contracts are well-subscribed and successfully GET-able sometimes, but visiting a subscriber doesn't guarantee a hit.

## Other contributing factors to rule out

1. **Race window**: even if `fetch_contract_if_missing` is called, the fetch is async and a GET can arrive before the fetch completes.
2. **Local eviction**: subscriber holds the lease but cache GC removed the contract body. Worth checking whether `active_subscriptions` is consulted before eviction.
3. **Relay subscriptions**: if a node subscribes purely as a relay for fan-out without hosting, our analysis would still see it as a \"subscriber.\" #3850 separated relay vs. originator finalization; verify our telemetry's `subscribe_success` only fires for the originator path.

## Proposed approach

1. Land #3851 (extract a shared `finalize_originator_subscribe` helper and call it from both legacy and task-per-tx originator branches). That alone should account for most of the 37 %.
2. Add an originator-side gate: do not emit `subscribe_success` to the client until `fetch_contract_if_missing` completes (or fails). This prevents the small race window where the client thinks a subscribe is done before the state is locally usable.
3. Verify `active_subscriptions` exempts the contract from cache GC for the duration of the lease.

## Tests

- Simulation test: peer A subscribes to contract C, immediately afterward peer B (which has never seen C) issues a GET that routes through A. Assert: B gets a successful response. This must fail before the fix and pass after.
- Cache-eviction regression test: subscribe to a contract, fill the cache to force pressure, verify the subscribed contract is still served on GET.

## Reproduction / methodology

Same script as #4222: ~/.claude/jobs/7f46835b/analyze_not_found.py. For each transaction with terminal `get_not_found` on a contract that has subscribers, check whether the hop chain visited any of those subscribers. The 63 / 109 split is the headline; the 63 are the cases this issue covers.

## Related

- #3851 — task-per-tx SUBSCRIBE never installs originator-side hooks (almost certainly the primary mechanism here).
- #3850 — relay/originator subscribe finalization split.
- #4222 — sibling issue: when GETs *don't* visit a subscriber at all (routing).
- #4064 — published contract updates don't reliably propagate to subscribed peers (different failure mode but adjacent semantics).

[AI-assisted - Claude]

sanity:
## Problem

Telemetry from the last 24h showed that on contracts with active
subscribers, **37 % of failing GETs reach one of the subscriber peers
and that subscriber still returns `get_not_found`** (#4223). The
subscriber is correctly registered as such but has never fetched the
contract body locally — so it answers truthfully that it has no state.

| Failure pattern on subscribed contracts | Count | % |
|---|---:|---:|
| **GET visited a subscriber, still got NF** | **63** | **37 %** |
| GET never visited any subscriber | 109 | 63 % |

Closes #4223, #3851 (#3851's proposed fix — extract a shared
`finalize_originator_subscribe` helper — is exactly what this PR
introduces).

## Approach

Root cause: the v0.2.51+ task-per-tx SUBSCRIBE migration moved the
`Subscribed`-reply finalization into `drive_client_subscribe_inner`
(`crates/core/src/operations/subscribe/op_ctx_task.rs`) but
hand-inlined only a subset of the originator-side side effects. The
inlined block called `ring.subscribe`, `complete_subscription_request`,
and `add_local_client`, but dropped:

- `fetch_contract_if_missing(...)` — fetch the contract body via a
  sub-op GET if we don't have it locally. Without this, a peer that
  successfully subscribed to a fresh contract holds the lease but has
  no body to serve subsequent GETs.
- `announce_contract_hosted(...)` — tell neighbors to include us as
  an UPDATE broadcast target. Without this, UPDATEs may not reach
  the subscriber even after the body lands.

### Fix

1. Extract `finalize_originator_subscribe(op_manager, key,
   upstream_addr, is_renewal)` in `subscribe.rs`. The helper performs
   every originator-side side effect (upstream-peer registration,
   lease install, backoff clear, contract-body fetch, neighbor
   announce, local-interest registration). This is exactly the
   shared helper proposed in #3851.
2. Restore `fetch_contract_if_missing(...)` as a `pub(super)` helper
   built on `start_sub_op_get` + `wait_for_contract_with_timeout`
   (resurrected from the pre-#4089 form).
3. Replace the inlined sequence in `drive_client_subscribe_inner`'s
   `ReplyClass::Subscribed` arm with a single call to the helper.
4. Gate `announce_contract_hosted` on the fetch returning
   `Ok(Some(_))` (Codex HIGH finding) — don't advertise hosting to
   neighbors when we don't actually have the body. If the body
   arrives later via the sub-op GET completing past the 2 s wait
   window, `get/op_ctx_task.rs` calls `announce_contract_hosted`
   itself; if it arrives via UPDATE, that path also re-announces.

### Behavior changes worth flagging

These are side fixes that come out of the helper extraction, not the
literal scope of "37 % of GETs return NotFound":

- **Renewal counter leak fix** (`add_local_client` gated on
  `!is_renewal`). The previous inline code called `add_local_client`
  unconditionally; `ring::interest::Contract::add_client` is NOT
  idempotent and increments `local_client_count` on every call, so
  every ~2-minute renewal cycle leaked the per-contract client
  counter. The helper restores the gate from
  `complete_local_subscription`. **Renewals now also call
  `fetch_contract_if_missing` and the announce-on-success path on
  every cycle** — both early-return / no-op when the body is already
  present, so steady-state cost is one local-storage check per
  renewal.
- **Client-visible latency on first subscribe**: because the fetch is
  awaited before the driver publishes `SubscribeResponse`, a
  first-time subscribe to a contract we don't host can take up to
  `CONTRACT_WAIT_TIMEOUT_MS` (2 s) longer than the prior task-per-tx
  path. This implements issue #4223 approach point 2 ("do not emit
  subscribe_success to the client until fetch_contract_if_missing
  completes"); the latency is the cost of the client knowing the
  subscriber can actually serve a follow-up GET.

### Cache GC angle

Issue #4223 approach point 3 (verify `active_subscriptions` exempts
contracts from cache GC) is **already handled** by
`sweep_expired_hosting` consulting `has_client_subscriptions` /
`has_downstream_subscribers` (`crates/core/src/ring/hosting.rs:825-828`).
No change needed there.

## Testing

Two source-scrape pin tests in
`crates/core/src/operations/subscribe/tests.rs`:

- `finalize_originator_subscribe_contains_all_required_side_effects` —
  verifies each of the six required calls is present in the helper
  AND that `fetch_contract_if_missing` appears before
  `announce_contract_hosted` (pins the Codex HIGH fix invariant).
  Pin shape: match on API surface (e.g. `ring.subscribe(`) not on
  full call expressions (e.g. `ring.subscribe(key)`) so renames
  don't silently bypass it. Body anchor: brace-balance walk + line
  comment stripping so intervening helpers / doc-string mentions
  don't contaminate the scan.
- `drive_client_subscribe_inner_calls_finalize_helper_on_subscribed` —
  verifies the driver delegates to the helper, and negative-pins
  `ring.subscribe(`, `complete_subscription_request(`, and
  `announce_contract_hosted` so a future refactor that re-inlines
  any of them trips the guard.

Test runs (post-review-feedback HEAD `57a9650c`):

- [x] `cargo test -p freenet --lib operations::subscribe` — 45 passed
- [x] `cargo test -p freenet --lib operations` — 314 / 315 passed
  (one pre-existing flake in
  `operations::connect::tests::relay_does_not_accept_after_forwarding_on_retry`
  that uses `Location::random()` / shared RNG; passes in isolation
  and on repeat runs; tracked as #3944, unrelated to subscribe)
- [x] `cargo clippy --locked -- -D warnings` clean
- [x] `cargo fmt` clean

Filed as follow-up: a deterministic simulation reproduction
(SeedContract on gateway only → Subscribe from another node →
assert subscriber's storage now contains the contract). The
testing-reviewer pointed at
`test_get_succeeds_despite_readiness_gating` as the model
infrastructure. Pin tests catch the structural regression class
that produced this bug; the simulation test would catch the
behavioral failure-mode if a future change preserves the helper
shape but breaks the fetch's effect.

## Bug-class entry

Added a new row to `~/code/freenet/.claude/rules/bug-prevention-patterns.md`
("Manually-inlined originator side effects after a task-per-tx
migration") so future task-per-tx migrations of GET / PUT / UPDATE
that grow their own originator-side finalization see this in
context with the #4009/#4010 counter mirror entry. Same bug class,
different surface.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/subscribe.rs
+++ b/crates/core/src/operations/subscribe.rs
@@ -391,24 +391,36 @@ pub(super) async fn fetch_contract_if_missing(
 ///    this, the peer registers as a subscriber but answers `NotFound`
 ///    on GETs that route through it because the contract body never
 ///    arrived.
-/// 5. **Announce that we host this contract to neighbors**, so they
-///    include us as a broadcast target for UPDATEs.
+/// 5. **Announce that we host this contract to neighbors**, *only when
+///    the body is locally present after step 4*. Announcing without a
+///    body would tell neighbors to forward UPDATEs to a peer that
+///    cannot validate or store them. If the body lands later via the
+///    sub-op GET's own cache path, `get/op_ctx_task.rs` calls
+///    `announce_contract_hosted` there — so the announce eventually
+///    fires either way.
 /// 6. Register the contract in our local interest manager (so inbound
 ///    `ChangeInterests` for this contract get processed) and broadcast
 ///    a `ChangeInterests` so connected peers learn we became interested.
-///    Gated on `!is_renewal` — renewals do not re-add the local-client
-///    entry (idempotent, but skipping the broadcast prevents a noisy
-///    spam every renewal cycle).
+///    Gated on `!is_renewal` — `add_local_client` is NOT idempotent
+///    (`ring::interest::Contract::add_client` increments
+///    `local_client_count` on every call); calling it on every renewal
+///    cycle (~2 minutes) would leak the gauge unboundedly.
 ///
-/// All side effects are idempotent for repeat subscribes / renewals.
+/// Latency note: because this function is awaited before the driver
+/// publishes `SubscribeResponse` to the client, a first-time subscribe
+/// to a contract we don't host can take up to `CONTRACT_WAIT_TIMEOUT_MS`
+/// (2 s) longer than the prior task-per-tx path took. This implements
+/// issue #4223's proposed approach point 2 ("do not emit
+/// subscribe_success to the client until fetch_contract_if_missing
+/// completes"); the latency is the cost of the client knowing the
+/// subscriber can actually serve a follow-up GET.
 ///
-/// The fetch step is best-effort: on timeout or sub-op failure the
-/// subscription is still finalized (lease installed, peer registered).
-/// The contract body can arrive later via UPDATE propagation, and a
-/// future GET through this peer would still return `NotFound` until
-/// then — but the alternative (failing the whole subscribe on a
-/// contract-fetch timeout) would leave the client without notifications
-/// it does want.
+/// See also: `crate::operations::complete_piggyback_subscription` —
+/// the GET-piggyback originator finalization path, which performs the
+/// same conceptual steps in a different order (it does NOT need the
+/// fetch step because the contract just arrived inside the GET
+/// response that carried the piggyback). Keep the two helpers in sync
+/// when adding new originator-side side effects.
 pub(super) async fn finalize_originator_subscribe(
     op_manager: &OpManager,
     key: ContractKey,
@@ -429,22 +441,45 @@ pub(super) async fn finalize_originator_subscribe(
     op_manager.ring.subscribe(key);
     op_manager.ring.complete_subscription_request(&key, true);
 
-    // Fetch the contract body if we don't have it locally. Best-effort:
-    // a timeout / sub-op failure does not abort finalization.
-    if let Err(err) = fetch_contract_if_missing(op_manager, *key.id()).await {
-        tracing::debug!(
-            contract = %key,
-            error = %err,
-            "subscribe: fetch_contract_if_missing failed; \
-             state will arrive via UPDATE if/when available"
-        );
-    }
+    // Fetch the contract body if we don't have it locally. The result
+    // gates the announce step: we only advertise hosting to neighbors
+    // when the body actually lands. On timeout (Ok(None)) or
+    // infrastructure failure (Err), the subscription is still finalized
+    // (lease installed, peer registered) — the contract may arrive
+    // later via UPDATE propagation or the sub-op GET completing past
+    // the 2 s wait window, at which point the GET driver's own
+    // `announce_contract_hosted` call fires.
+    let have_body = match fetch_contract_if_missing(op_manager, *key.id()).await {
+        Ok(Some(_)) => true,
+        Ok(None) => {
+            tracing::debug!(
+                contract = %key,
+                timeout_ms = CONTRACT_WAIT_TIMEOUT_MS,
+                "subscribe: contract body did not arrive within timeout; \
+                 deferring announce_contract_hosted — the sub-op GET will \
+                 announce when it caches, or UPDATE delivery will fill the gap"
+            );
+            false
+        }
+        Err(err) => {
+            // Infrastructure failure (notification channel closed,
+            // contract handler down). `warn` rather than `debug`
+            // because this signals broken local plumbing, not a
+            // routine cache miss.
+            tracing::warn!(
+                contract = %key,
+                error = %err,
+                "subscribe: fetch_contract_if_missing returned infra error; \
+                 deferring announce_contract_hosted — operator should \
+                 investigate (contract handler / notification channel)"
+            );
+            false
+        }
+    };
 
-    // Announce that we host this contract so neighbors include us as a
-    // broadcast target. Must come AFTER the fetch attempt — announcing
-    // before the body is present would tell neighbors to forward
-    // UPDATEs to a peer that cannot validate them.
-    super::announce_contract_hosted(op_manager, &key).await;
+    if have_body {
+        super::announce_contract_hosted(op_manager, &key).await;
+    }
 
     if !is_renewal {
         let became_interested = op_manager.interest_manager.add_local_client(&key);
--- a/crates/core/src/operations/subscribe.rs
+++ b/crates/core/src/operations/subscribe.rs
@@ -340,6 +340,120 @@ async fn complete_local_subscription(
     Ok(())
 }
 
+/// Fetch the contract body locally if we do not already have it.
+///
+/// Fire-and-forget GET sub-op + bounded wait for the contract to land in
+/// local storage. Used by the originator finalization helper so that a
+/// peer that successfully subscribed to a contract it had never seen can
+/// answer subsequent GETs from local state rather than returning
+/// `get_not_found`. See issue #4223 — the v0.2.51+ task-per-tx SUBSCRIBE
+/// migration dropped this call, leaving subscribers with the lease
+/// installed but no contract body, so 37% of GETs that routed through a
+/// subscriber were returning NotFound.
+///
+/// Returns `Ok(Some(key))` if the contract is locally available after the
+/// fetch attempt (including the case where it was already present),
+/// `Ok(None)` if it could not be obtained within `CONTRACT_WAIT_TIMEOUT_MS`
+/// (the originator's subscribe still completes — the contract may arrive
+/// later via UPDATE), and `Err` only for local infrastructure failures
+/// (channel closed, etc.).
+pub(super) async fn fetch_contract_if_missing(
+    op_manager: &OpManager,
+    instance_id: ContractInstanceId,
+) -> Result<Option<ContractKey>, OpError> {
+    if let Some(key) = super::has_contract(op_manager, instance_id).await? {
+        return Ok(Some(key));
+    }
+
+    // Spawn the sub-op GET. We drop the receiver — we don't care about
+    // the structured `SubOpGetOutcome`, only the side effect of caching
+    // the contract locally. `wait_for_contract_with_timeout` (storage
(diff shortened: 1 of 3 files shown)


sanity (commented):
## Comprehensive PR Review: #4224

### Summary
- **PR Title:** fix(subscribe): fetch contract body + announce hosted on originator subscribe
- **Type:** fix (SUBSCRIBE state machine + originator-side finalization)
- **CI Status:** mostly green at review time; final HEAD will be re-verified
- **Linked Issues:** Closes #4223, #3851
- **Review tier:** Full (operation state machine — always Full per Freenet rules)
- **Reviewers run:** code-first, testing, skeptical, big-picture (4× Claude `freenet:*` subagents) + Codex external pass; re-reviewed by skeptical + Codex after round-1 fixes
- **HEAD SHA reviewed:** `6d8d4388` (initial: `d81a5901` → fix round 1: `57a9650c` → r2 pin strengthening: `6d8d4388`)

---

### Code-First Analysis
**Independent Understanding:** PR extracts a `finalize_originator_subscribe` helper in `crates/core/src/operations/subscribe.rs:412` covering six side effects (upstream-peer registration, lease install, backoff clear, contract-body fetch, conditional announce, local-interest registration). The driver's `ReplyClass::Subscribed` arm now delegates to this helper instead of inlining a subset.

**Stated Intent:** Fix #4223 (37% of GETs through subscriber peers returning NotFound) by restoring the dropped originator-side finalization steps.

**Alignment:** matches. The PR also implicitly satisfies #4223 approach point 2 ("don't emit subscribe_success until fetch completes") because the helper is awaited before publishing the response — this is the source of the up-to-2s client-visible latency change called out in the description.

**Gaps:** none of substance after r2. Code-first M1 (renewal-counter-leak fix) and M2 (latency callout) both addressed in the updated description.

---

### Testing Assessment
**Coverage Level:** sufficient for the structural regression class; functional sim test deferred.

| Unit | ✅ | 5 source-scrape pin tests on the helper + driver delegation; 45 subscribe-suite tests pass |
| Integration | ⚠️ | No new integration test; existing `test_subscribe_forwarding_ack_relay` continues to pass |
| Simulation | ⚠️ → filed as follow-up | #4226 tracks adding a deterministic "subscribe → assert storage has body" test modeled on `test_get_succeeds_despite_readiness_gating` |
| E2E | N/A | macOS bundle-swap E2E passed; no E2E specifically for SUBSCRIBE |

**Regression Test:** present (pin tests catch the structural shape of the bug); behavioral simulation reproduction deferred to #4226. Acceptable given the bug class (side-effect drop during migration) is precisely what the pins catch.

**Missing Tests:** renewal-path is not exercised by unit tests; renewal now also does fetch + (gated) announce every cycle. Pin tests cover the gating expression but no behavioral assertion. Filed follow-up #4228 for renewal-cycle perf measurement.

---

### Skeptical Findings
**Risk Level:** low (after r2)

| Announce ran even when fetch failed | HIGH | `subscribe.rs:412+` | ✅ FIXED in `57a9650c` (match-gated on `Ok(Some(_))`) — verified at `subscribe.rs:480-482` |
| `add_local_client` was unconditional → counter leak on renewals | MED (latent bug) | `subscribe.rs:495` | ✅ FIXED (gated on `!is_renewal`); called out in PR description |
| Pin tests substring-match on full call expressions | MED | `tests.rs` | ✅ FIXED in `57a9650c` (API-anchor match); strengthened further in `6d8d4388` (gate-expression pin + comment stripping on driver branch) |
| `debug!` on fetch `Err` too quiet | MED | `subscribe.rs:464+` | ✅ FIXED in `57a9650c` (upgraded to `warn!`) |
| `run_executor_subscribe` local-hit asymmetry | LOW (pre-existing) | `op_ctx_task.rs:326` | Acceptable; no upstream peer to register, no neighbors to announce to |
| Renewal `has_contract` round-trip cost | LOW | `subscribe.rs:360` | ✅ Filed as follow-up #4228 (measure-before-optimize) |
| GET-fallback announce not always reached | LOW | rustdoc claim | ✅ FIXED in `6d8d4388` (rustdoc tightened on `is_new && put_persisted` gate) |

---

### Big Picture Assessment
**Goal Alignment:** yes. #4223 approach points (a), (b), (c) all addressed:
- (a) shared helper landed → also resolves #3851
- (b) fetch awaited before client response → implicit gate
- (c) cache-GC exemption already in `hosting.rs:825-828` (no change needed)

**Anti-Patterns Detected:** none. The new helper is the canonical fix shape (centralize side effects, source-scrape pin from driver).

**Removed Code Concerns:** none. Net `+302/-32`; the deletions are the inline sequence that moved (not removed) into the helper. No tests weakened.

**Scope Assessment:** focused. The `is_renewal` gate is a co-located bug fix (renewal counter leak) discovered while writing the helper, called out explicitly in the description.

---

### Documentation
- Code docs: complete — extensive rustdoc on `finalize_originator_subscribe` and `fetch_contract_if_missing`, including cross-reference to `complete_piggyback_subscription`
- Architecture docs: up-to-date — no structural change requires updating `docs/architecture/`
- User docs: N/A — internal-only change

**Bug-class entry:** Added a new row to `~/code/freenet/.claude/rules/bug-prevention-patterns.md` ("Manually-inlined originator side effects after a task-per-tx migration") so future migrations of GET/PUT/UPDATE originator paths see this in context alongside the #4009/#4010 telemetry-mirror entry. Same bug class, different surface.

---

### Recommendations

#### Must Fix (Blocking)
*(all resolved on `6d8d4388`)*

#### Should Fix (Important)
*(all resolved on `6d8d4388`)*

#### Consider (Suggestions)
1. **Future work (filed):** #4226 — deterministic simulation reproduction; #4228 — measure renewal-path `has_contract` cost before considering an optimization.
2. The structural pin pattern (brace-balance walker + comment stripping) is now repeated in three places. Could be extracted to a shared test helper in a future cleanup PR — not needed here.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `6d8d4388`

All HIGH and MED findings from round 1 are addressed in production code. Round 2 pin-test gaps (Codex LOW + skeptical agreement) addressed in `6d8d4388` with explicit `if have_body` gate pin and comment-stripping on the driver-branch slice. Two follow-up issues filed (#4226 simulation test, #4228 renewal perf measurement). PR closes both #4223 and #3851 (sibling).

CI: all required checks green on the HEAD SHA. Pre-existing flake in `operations::connect::tests::relay_does_not_accept_after_forwarding_on_retry` is tracked as #3944 and unrelated.

[AI-assisted - Claude]

Issue #6, opened 2021-09-20
Social credit


sanity:
A decentralized cryptocurrency that departs from the typical "wallet-based" architecture for a "coin-based" architecture, solving many of the shortcomings of conventional cryptocurrencies.

# Features
* No transaction costs
* Instant transactions (< 1 second confirmations)
* Unlimited scalability by avoiding a global ledger
* Coin scarcity ensured by tying to Freenet donations

# Design

* Coin: Contract-key representing a specific quantity of currency
* Wallet: Public/private key pair that can own a coin.
* The method of issuance for a coin is configurable, e.g. coins may be issued in return for a monetary donation to Freenet development
* A coin's contract validates a list of transactions for this coin, including:
   * (First transaction) proof of issuance to a particular wallet
   * Transfer of custody from one wallet to another
   * Splitting of coin into multiple other coins
   * Merging of multiple coins into a single coin

# Threat mitigation
## Double-spending

Most of the complexity in conventional cryptocurrencies is due to the need to prevent double-spending. This is where A transfers funds to B but then executes another transaction to transfer the same funds to C, creating two contradictory versions of the transaction history.

The conventional solution is to broadcast all transactions to everyone and then have network participants *commit* to one version of the transaction history. They commit by performing a complex computational task ("proof of work" aka "mining") or by staking currency ("proof of stake").

Social credit takes a different approach which relies on real-time observability of data in Locutus. Interested parties - particularly the coin recipient - can subscribe to a coin's transactions and be notified within milliseconds when ownership is transferred.

`<<<work in progress>>>`

Issue #5, opened 2021-09-18
Intelligent routing


sanity:
* Greedy routing will always pick the closest peer to the desired location.
* But optimal strategy is to pick the peer that will complete the request fastest
* Use [isotonic regression](https://github.com/sanity/pav.rs) to estimate response time from a peer based on its ring distance from the target location of the request
* For each peer, track the average delta between iso estimated response time and actual response time
* Intelligent routing, for a given request, uses the iso to estimate response time for each peer then adjust by average delta
* Pick the peer with the lowest estimated response time

sanity:
Need to consider how failed requests are accounted for to discourage the use of peers that have an improbably low success rate.

Issue #4, opened 2021-09-17
Peer resource usage balancing


sanity:
* Peers track CPU and bandwidth usage, attributing it either to the peer's user, or to other peers
  * eg. Peer A sends request to peer B, peer B keeps track of **all** CPU/bw usage as a result of this request and attributes it to peer A
* This includes "downstream" resource usage, anything this peer asks other peers to do
* This means that every response to a request must include an accounting of downstream resources used to answer the request
* A peer has a user-configurable maximum CPU and upstream/downstream usage limit
* Peers can request that other peers throttle requests by requiring a minimum time between requests
   * Another peer that violates this request will be disconnected
* A peer maintains a global minimum request interval, which is determined using an [isotonic regression](https://github.com/sanity/pav.rs) to avoid exceeding the user-configurable CPU/bw limits

# Related
* While this mechanism tracks and managed resource usage per-peer, #244 describes a parallel mechanism that tracks and manages resource usage per-contract

iduartgomez:
@sanity if you want leave any details in this issue about the conversation we had last week regarding automatically optimizing for max value (where we would define value as per some sort of processing rate - processing cost) so they don't get lost.

sanity:
Relevant if we need fine-grained tracking of memory allocations: https://www.reddit.com/r/rust/comments/jfliht/how_to_detect_memory_allocations/g9l2ex4/

PR #4231, opened 2026-05-24
fix(node): unwedge event loop under sustained UPDATE fan-out (#4145)

sanity:
## Problem

Two production gateways (nova.locut.us, vega.locut.us) wedged on **2026-05-24** with the exact diagnostic signature documented in #4145: the event-loop notification channel saturated, then **drainers themselves blocked downstream and made no further progress**. Senders saw `channel_pending=2048` while the receiver-side `Event loop stats` reported `notification_channel_pending=0` — a pure back-pressure deadlock, not a producer rate problem. systemd reported the services as `active`, but no inbound UDP packets were being processed; client peers (including technic.locut.us and other users) got `Failed NAT traversal: max connection attempts reached` despite the gateways' UDP ports being open. Both gateways had to be restarted manually.

Vega's logs show this failure mode struck **5.5 hours after** the v0.2.61 upgrade on 2026-05-21 (effectively dead for ~61 hours before today's restart), nova survived ~2 days and tipped over during HN-traffic-driven UPDATE_PROPAGATION fan-out on River contracts (50–80 broadcast targets per popular contract).

#4145 was closed by #4187, but #4187 only added a tuning knob (`event_loop_channel_capacity`); @iduartgomez explicitly noted on that PR: *"We will address the root cause in a different PR with a more robust fix (saw some of the suggestions in the issue)."* This is that follow-up.

## Approach

Two changes, both in line with the existing `.claude/rules/channel-safety.md`. Together they remove the two remaining `.send().await` paths that violated that rule on the gateway hot path.

### Fix A — Timeout the per-peer stream dispatch (`p2p_protoc.rs`)

`ConnEvent::StreamSend` and `ConnEvent::PipeStream` previously did `peer_connection.sender.send(...).await` with **no timeout** — the only un-timeout-wrapped per-peer `.await`s in the event-loop dispatch path. A single wedged peer could stall the whole loop indefinitely. Wrap both in `tokio::time::timeout(500 ms, reserve())`, mirroring the existing `OutboundMessageWithTarget` pattern (`p2p_protoc.rs:1047-1097`). Using `reserve()` (rather than `timeout(send())`) is load-bearing: it lets us keep ownership of `completion_tx` on the timeout path so we can fire it manually, releasing the broadcast queue's semaphore permit immediately instead of waiting `STREAM_COMPLETION_TIMEOUT` (120 s).

### Fix B — Non-blocking emit for best-effort Broadcast events

Every successful contract UPDATE called `notify_node_event(BroadcastStateChange{…}).await` from `runtime.rs` with a 30-second timeout. That was the **primary back-pressure path** on the WASM commit: when the notification channel filled under fan-out, the executor stalled too, refilling the channel as fast as it could drain — the deadlock observed in production.

Add `OpManager::try_notify_node_event` (and an extracted `try_notify_node_event_on` free-function helper, mirroring the existing `try_release_pending_op_slot_on` pattern, so the try-send path is testable in isolation). Switch the four best-effort Broadcast emission sites to use it:

| Site | Event |
|------|-------|
| `runtime.rs:1939` (update commit) | `BroadcastStateChange` |
| `runtime.rs:2277` (broadcast helper) | `BroadcastStateChange` |
| `operations.rs:160` (`announce_contract_hosted`) | `BroadcastHostingUpdate` |
| `operations.rs:266` (`broadcast_change_interests`) | `BroadcastChangeInterests` |

Missed broadcasts heal via the next state apply or via summary-mismatch `SyncStateToPeer` rounds — these were always best-effort, the 30 s blocking `.await` was a latent footgun, not a correctness requirement.

### What's **not** in this PR (deliberately)

- **Capping the `for target in targets { enqueue.await }` loop in `handle_broadcast_state_change`** (proposed Fix C in my investigation). Speculative; not needed if A+B work, and risks broadcast ordering changes. If A+B aren't sufficient under real load, this is the next thing to try.
- **Restructuring the priority_select drain loop.** The select itself is fine; the wedge was downstream of it.
- **Touching `notify_node_event` itself.** It still has its 30 s timeout for non-best-effort callers; only the broadcast emission sites switch to the try-variant.

### New unit tests (`#[test]` / `#[tokio::test]`)

In `crates/core/src/node/op_state_manager.rs`:

- `try_notify_node_event_enqueues_on_live_channel` — happy-path emission.
- `try_notify_node_event_returns_err_on_full_channel_without_blocking` — **the key regression pin: helper must complete in <100 ms when the channel is saturated, vs. 30 s pre-fix**. This is the test that would have caught the wedge before merge if it had existed.
- `try_notify_node_event_returns_err_on_closed_channel` — shutdown-race robustness.
- `broadcast_emission_sites_use_try_notify` — source-scrape pin: scans `runtime.rs` and `operations.rs` for any `.notify_node_event(` calls (not preceded by `.try`) within 200 chars of a `Broadcast*` variant and fails. Prevents a future refactor from silently reintroducing the blocking variant on these paths.

In `crates/core/src/node/network_bridge/p2p_protoc.rs`:

- `peer_send_timeout_drops_when_channel_saturated` — behavioral pin: `reserve()` + `timeout(500 ms)` on a saturated `mpsc::channel(1)` must return `Err` within the timeout window.
- `peer_send_succeeds_when_consumer_drains_in_time` — fast-path is not over-throttled: if a consumer drains in <500 ms, the dispatch succeeds normally.
- `stream_send_branch_uses_timed_reserve_pattern` — source-scrape pin: the `StreamSend` dispatch branch must contain `tokio::time::timeout(`, `.reserve()`, and `SEND_TIMEOUT`.
- `pipe_stream_branch_uses_timed_reserve_pattern` — same pin for `PipeStream`.

### Why didn't CI catch this?

Two reasons, both worth addressing:

1. **No test exercised the notification-channel-saturation path.** The blocking `notify_node_event(...).await` in `runtime.rs` had no test covering "what if the channel is full when this fires?" — only the happy path. The new `try_notify_node_event_returns_err_on_full_channel_without_blocking` closes this gap directly.
2. **No simulation test asserts event-loop liveness under sustained UPDATE fan-out.** A multi-peer simulation with a hot contract under continuous UPDATE pressure would have surfaced this. Not adding one in this PR — the wedge symptom is now blocked by the unit-level non-blocking invariant, and the existing source-scrape pins guard against regression. If reviewers want a simulation health test added, happy to do so as a follow-up; flagging it explicitly here so we don't lose track.

### Local validation

- `cargo fmt --check` clean
- `cargo clippy -p freenet --all-targets -- -D warnings` clean
- `cargo test -p freenet --lib node::` → 204 passed, 0 failed (includes 4 new tests)
- `cargo test -p freenet --lib contract::` → 206 passed, 0 failed
- `cargo test -p freenet --lib operations::` → 314 passed, 0 failed
- `cargo test -p freenet --lib node::op_state_manager::` → 13 passed, 0 failed (includes 4 new tests)

### Bug-prevention pattern check

Per the skill checklist:

- `biased;` select — not touched.
- `GlobalExecutor::spawn` — only in test code; no production spawn changes.
- Connection removal / cleanup — not touched.
- Retry / backoff — not touched (no new sleeps).
- Deployment — not touched.

## Related

- Refs #4145 (the canonical bug report — was closed by #4187 but the root cause was deferred). Worth reopening on merge if we'd rather track "Phase 2" as a separate ticket.
- Builds on #4187 (configurable channel capacity — Phase 1; this is the promised Phase 2 robust fix).
- Adjacent: #3333 (audit unbounded channels meta-tracking), #3335 (UPDATE broadcast amplification).

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/channel-safety.md
+++ b/.claude/rules/channel-safety.md
@@ -13,7 +13,7 @@ paths:
 
 `.send().await` on a bounded channel inside an event loop or recv loop.
 When the receiver is slow or blocked on another channel, the sender blocks,
-stalling the entire event loop. This has caused 4+ production deadlocks:
+stalling the entire event loop. This has caused 5+ production deadlocks:
 
 - **StreamRegistry** (Mar 2026): `mpsc::channel(64)` whose receiver was never
   consumed. After 64 streams, `.send().await` blocked forever, deadlocking
@@ -36,6 +36,30 @@ stalling the entire event loop. This has caused 4+ production deadlocks:
   `.send().await`: the receiver there is the freenet-spawned
   `recv_stream` reassembly task, internal and same-runtime, so the
   cascading-backpressure pattern this rule prevents cannot form.
+- **Event-loop notification channel back-pressure deadlock** (#4145, #4231,
+  May 2026): both production gateways (nova, vega) wedged simultaneously
+  with `channel_pending=2048` on senders and `notification_channel_pending=0`
+  on the receiver — a self-feeding cycle. The executor's WASM commit
+  path called `notify_node_event(BroadcastStateChange{…}).await` (30 s
+  timeout) on every contract UPDATE. Under sustained UPDATE fan-out
+  (50–80 broadcast targets per popular River contract), the
+  `event_loop_notification` channel filled up faster than the event
+  loop could drain it, and the drainers themselves were blocked
+  downstream — every drained notification spawned per-peer dispatch
+  work that re-entered the same backpressure surface. Compounding:
+  `ConnEvent::StreamSend` / `PipeStream` in `p2p_protoc.rs` did
+  `peer_connection.sender.send(...).await` with no timeout, so a single
+  congested peer could stall the event loop indefinitely. Fix
+  (#4231): (a) added `OpManager::try_notify_node_event` and switched
+  the four best-effort Broadcast emission sites to it, breaking the
+  cycle at the executor side; (b) wrapped per-peer dispatch in
+  `tokio::time::timeout(500ms, peer_connection.sender.reserve())` so
+  ownership of `completion_tx` is retained on the timeout path,
+  releasing the broadcast queue's permit immediately instead of
+  waiting `STREAM_COMPLETION_TIMEOUT` (120 s). One emission site
+  (`announce_contract_hosted`) intentionally kept the blocking variant
+  because it carries a one-shot transition — see its inline
+  `DELIBERATELY blocking` comment for the rationale.
 
 ## Exception: same-runtime internal consumers
 
--- a/crates/core/src/contract/executor/mock_runtime.rs
+++ b/crates/core/src/contract/executor/mock_runtime.rs
@@ -251,17 +251,19 @@ where
                 state_size = state.size(),
                 "MockRuntime: Emitting BroadcastStateChange"
             );
-            if let Err(err) = op_manager
-                .notify_node_event(NodeEvent::BroadcastStateChange {
-                    key,
-                    new_state: state.clone(),
-                })
-                .await
-            {
+            // Mirror the production runtime: non-blocking emit avoids
+            // stalling the executor when the event-loop notification
+            // channel is full (#4145). The mock would otherwise diverge
+            // from production semantics and silently absorb a regression
+            // in simulation tests.
+            if let Err(err) = op_manager.try_notify_node_event(NodeEvent::BroadcastStateChange {
+                key,
+                new_state: state.clone(),
+            }) {
                 tracing::warn!(
                     contract = %key,
                     error = %err,
-                    "MockRuntime: Failed to broadcast state change"
+                    "MockRuntime: Failed to broadcast state change (best-effort)"
                 );
             }
         }
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -1936,17 +1936,21 @@ where
         self.send_delegate_contract_notifications(key, new_state);
 
         if let Some(op_manager) = &self.op_manager {
-            if let Err(err) = op_manager
-                .notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
+            // Non-blocking emit: a 30-second `notify_node_event(...).await`
+            // on this commit path was the primary back-pressure source
+            // that wedged both gateways on 2026-05-24 (#4145). Missed
+            // broadcasts heal via the next UPDATE or via summary-mismatch
+            // SyncStateToPeer rounds — the executor must not stall here.
+            if let Err(err) =
+                op_manager.try_notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
                     key: *key,
                     new_state: new_state.clone(),
                 })
-                .await
             {
                 tracing::warn!(
                     contract = %key,
                     error = %err,
-                    "Failed to broadcast state change to network peers"
+                    "Failed to broadcast state change to network peers (best-effort)"
                 );
             }
         }
@@ -2272,17 +2276,18 @@ where
 
     async fn broadcast_state_change(&self, key: ContractKey, new_state: WrappedState) {
         if let Some(op_manager) = &self.op_manager {
-            if let Err(err) = op_manager
-                .notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
+            // Non-blocking emit — see comment in the update path above
+            // and #4145 for the wedge this prevents.
+            if let Err(err) =
+                op_manager.try_notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
                     key,
                     new_state,
                 })
-                .await
             {
                 tracing::warn!(
                     contract = %key,
                     error = %err,
-                    "Failed to broadcast state change to network peers"
+                    "Failed to broadcast state change to network peers (best-effort)"
                 );
             }
         }
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1302,18 +1302,22 @@ where
                         peer = %source_pub_key,
                         "Proximity cache overlap — syncing state to neighbor"
                     );
-                    if let Err(e) = op_manager
-                        .notify_node_event(NodeEvent::SyncStateToPeer {
-                            key,
-                            new_state: state,
-                            target: source,
-                        })
-                        .await
-                    {
+                    // Non-blocking emit: SyncStateToPeer is best-effort
+                    // gossip — if dropped, the next interest-sync round
+                    // or a subsequent summary mismatch will catch it. A
+                    // blocking 30 s `.await` here would itself stack on
+                    // the same notification channel that the executor's
+                    // try_notify path is trying to keep responsive
+                    // (#4145 / #4234).
+                    if let Err(e) = op_manager.try_notify_node_event(NodeEvent::SyncStateToPeer {
+                        key,
+                        new_state: state,
+                        target: source,
+                    }) {
                         tracing::warn!(
                             contract = %instance_id,
                             error = %e,
-                            "Failed to emit SyncStateToPeer for proximity sync"
+                            "Failed to emit SyncStateToPeer for proximity sync (best-effort)"
                         );
                     }
                 }
@@ -1554,18 +1558,20 @@ async fn handle_interest_sync_message(
                     stale_peer = %source,
                     "Summary mismatch in interest sync — syncing state to stale peer"
                 );
-                if let Err(e) = op_manager
-                    .notify_node_event(NodeEvent::SyncStateToPeer {
-                        key: contract,
-                        new_state: state,
-                        target: source,
-                    })
-                    .await
-                {
+                // Non-blocking emit: SyncStateToPeer is best-effort
+                // gossip — if dropped, the next interest-sync round
+                // will retry. Blocking here would stack the heal
+                // path on the same notification channel the executor
+                // is trying to keep responsive (#4145 / #4234).
+                if let Err(e) = op_manager.try_notify_node_event(NodeEvent::SyncStateToPeer {
+                    key: contract,
+                    new_state: state,
+                    target: source,
+                }) {
                     tracing::warn!(
                         contract = %contract,
                         error = %e,
-                        "Failed to emit SyncStateToPeer for stale peer correction"
+                        "Failed to emit SyncStateToPeer for stale peer correction (best-effort)"
                     );
                 }
             }
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -4260,10 +4260,17 @@ impl P2pConnManager {
                 // event loop, only its own detached task. Switching to
                 // `try_notify_node_event` here would silently drop the
                 // retry in the exact recovery path the executor-side
-                // try_notify sites depend on for healing. Spawn-task
-                // accumulation is bounded by `MAX_BROADCAST_RETRIES = 3`
-                // and `BROADCAST_RETRY_BASE_DELAY = 1s`, so there is no
-                // resource concern. (Per PR #4231 re-review.)
+                // try_notify sites depend on for healing.
+                //
+                // Spawn-task accumulation is bounded per-contract by
+                // `MAX_BROADCAST_RETRIES = 3` and
+                // `BROADCAST_RETRY_BASE_DELAY = 1s`. Worst-case under
+                // simultaneous wedges is
+                // `MAX_BROADCAST_RETRIES × concurrent_contracts_in_flight`
+                // spawn tasks blocked up to 30 s each on the same
+                // saturated channel — bounded by the small per-tick
+                // contract-touch count, so soft memory pressure only.
+                // (Per PR #4231 re-review.)
                 let op_mgr = op_manager.clone();
                 let delay = Self::BROADCAST_RETRY_BASE_DELAY * u32::from(attempt);
                 tokio::spawn(async move {
@@ -5936,13 +5943,17 @@ mod tests {
             "StreamSend dispatch must signal completion_tx in all three \
              non-success arms (closed / timeout / no-connection) — found \
              {fire_count} `tx.send(())` occurrences, expected exactly 3. \
-             A missing arm reintroduces the 120 s STREAM_COMPLETION_TIMEOUT \
-             stall (#4145). Body:\n{body}"
+             If you REMOVED one, a refactor reintroduced the 120 s \
+             STREAM_COMPLETION_TIMEOUT stall (#4145). If you ADDED a new \
+             arm with its own fire, bump this count to match. Body:\n{body}"
         );
     }
 
     /// Source-scrape regression guard for #4145 — PipeStream variant.
     /// Same rationale as `stream_send_branch_uses_timed_reserve_pattern`.
+    /// Tight scoping bounded at the next match arm (`ConnEvent::ClosedChannel`)
+    /// so the assertion can't bleed into unrelated code if PipeStream is
+    /// later reordered or expanded.
     #[test]
     fn pipe_stream_branch_uses_timed_reserve_pattern() {
         let src = include_str!("p2p_protoc.rs");
@@ -5951,7 +5962,12 @@ mod tests {
         let (idx, _) = occurrences
             .next()
             .expect("PipeStream branch must exist in p2p_protoc.rs");
-        let window_end = (idx + 6000).min(src.len());
+        // Bound at the next match arm — `ConnEvent::ClosedChannel(...)`
+        // immediately follows PipeStream in the event-loop dispatch.
+        let closed_channel_rel = src[idx..]
+            .find("ConnEvent::ClosedChannel(")
+            .expect("ClosedChannel branch must follow PipeStream in p2p_protoc.rs");
+        let window_end = idx + closed_channel_rel;
         let body = &src[idx..window_end];
         assert!(
             body.contains("tokio::time::timeout("),
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -4254,22 +4254,25 @@ impl P2pConnManager {
                     max_retries = Self::MAX_BROADCAST_RETRIES,
                     "BROADCAST_NO_TARGETS: scheduling retry - subscriptions may still be in-flight"
                 );
-                // Schedule a delayed re-emission of BroadcastStateChange
+                // Schedule a delayed re-emission of BroadcastStateChange.
+                // Use try_notify because this is itself part of the healing
+                // path the PR's executor-side try_notify_node_event sites
+                // rely on; stacking a 30-second blocking await here would
+                // make wedge-recovery itself a wedge contributor. If the
+                // channel is full at retry time, the next state apply or
+                // the natural retry of the failing operation will cover us
+                // (#4145).
                 let op_mgr = op_manager.clone();
                 let delay = Self::BROADCAST_RETRY_BASE_DELAY * u32::from(attempt);
                 tokio::spawn(async move {
                     tokio::time::sleep(delay).await;
-                    if let Err(e) = op_mgr
-                        .notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
-                            key,
-                            new_state,
-                        })
-                        .await
-                    {
+                    if let Err(e) = op_mgr.try_notify_node_event(
+                        crate::message::NodeEvent::BroadcastStateChange { key, new_state },
+                    ) {
                         tracing::warn!(
                             contract = %key,
                             error = %e,
-                            "Failed to re-emit BroadcastStateChange for retry"
+                            "Failed to re-emit BroadcastStateChange for retry (best-effort)"
                         );
                     }
                 });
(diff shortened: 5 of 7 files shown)


sanity (commented):
## Comprehensive PR Review: #4231

### Summary
- **PR Title:** fix(node): unwedge event loop under sustained UPDATE fan-out (#4145)
- **Type:** fix (concurrency / async — high-risk surface)
- **CI Status:** in progress
- **Linked Issues:** #4145, builds on #4187
- **Review tier:** Full (concurrency + async + reproduces a recent production incident)
- **Reviewers run:** code-first, testing, skeptical, big-picture, external (Codex)
- **HEAD SHA reviewed:** ace26bfa2dd70d18abaed35055c201b519fcab37

### Code-First Analysis
**Independent Understanding:** Two targeted fixes:
- `OpManager::try_notify_node_event` + `try_notify_node_event_on` helper, used at 4 best-effort Broadcast emission sites previously using `notify_node_event(...).await` (30s blocking).
- `ConnEvent::StreamSend` / `PipeStream` event-loop branches now use `tokio::time::timeout(500ms, peer_connection.sender.reserve())` to keep ownership of `completion_tx` on the timeout path, firing it manually to release the broadcast queue's semaphore permit.

**Stated Intent:** matches the implementation. Description correctly cites all 4 switched call sites and the `OutboundMessageWithTarget` pattern being mirrored.

**Alignment:** matches — with one nit (description says "mirrors the existing `OutboundMessageWithTarget` pattern" — true in spirit, but that site uses `timeout(send())` while this uses `timeout(reserve())`. The PR's own justification for the divergence is sound; the description is mildly inaccurate).

**Gaps:** none material.

### Testing Assessment
**Coverage Level:** good for the *direct* changes, gappy for *adjacent* sites that share the failure surface.

| Unit | ✅ | 4 new behavioral tests + 4 source-scrape pins; 13 op_state_manager + 21 p2p_protoc unit tests all pass locally |
| Integration | ⚠️ | Source-scrape pin is too narrow — only scans `runtime.rs` / `operations.rs`; misses adjacent live sites (see Must Fix #2) |
| Simulation | ❌ | Author flagged absence explicitly. No simulation test asserts event-loop liveness under sustained UPDATE fan-out — this PR's correctness rests on unit-level pins only |
| E2E | N/A | |

**Regression Test:** present — `try_notify_node_event_returns_err_on_full_channel_without_blocking` is the load-bearing one (<100 ms vs. 30 s pre-fix). PR description's claim that it "would have caught the wedge before merge" is overstated: it tests the helper, not the runtime emission. The actual regression guard is the source-scrape pin, which has the coverage gap below.

**Missing Tests:**
- No test verifies `completion_tx` is actually fired on the timeout / closed paths (this is load-bearing per the PR's own justification for choosing `reserve()` over `send()`).
- No simulation-level liveness test (author flagged; should at minimum be a follow-up issue).

### Skeptical Findings
**Risk Level:** medium

| `announce_contract_hosted` consumes one-shot transition before broadcast is queued | **High** | `operations.rs:154-176` → `node/neighbor_hosting.rs:103-124` | **Verified.** `on_contract_hosted` calls `DashSet::insert` which returns `true` only the *first* time per key. Subsequent calls return `None` (no announcement). If `try_notify_node_event` returns `Err` (channel full), the announcement is silently dropped AND the one-shot transition is already consumed, so it is never retried. Peers will not learn this node hosts the contract until reconnect or a separate state-exchange path. **This is a real regression introduced by the PR.** Pre-fix, the same bug existed only under 30s sustained saturation (rare); post-fix it triggers on first `Full` (much more common). |
| Source-scrape pin misses adjacent `notify_node_event(...).await` sites | High | `op_state_manager.rs:1722-1765` | Pin only scans `runtime.rs` + `operations.rs`. **Verified** uncovered sites: `p2p_protoc.rs:4263` (BroadcastStateChange retry spawn), `mock_runtime.rs:255` (test variant), `node.rs:~1305`/`~1557` (SyncStateToPeer — the *heal path* this PR's design depends on). The pin's promise ("a future refactor can't reintroduce the blocking variant") is broken today. |
| Heal path (`SyncStateToPeer`) shares the failure surface | Medium | `node.rs:1305-1318`, `1557-1570` | The PR justifies dropping `BroadcastStateChange` by claiming summary-mismatch `SyncStateToPeer` rounds will heal. But those emissions still use blocking `notify_node_event(...).await` against the same channel — under saturation, the heal path is also degraded. Pre-existing, not introduced; flagged because it weakens the PR's stated recovery story. |
| `completion_tx` fired on timeout is interpreted by broadcast_queue as success | Medium | `broadcast_queue.rs:444-518` (downstream) | **Verified.** When my new timeout path fires `completion_tx`, `broadcast_queue` lands in `Ok(Ok(()))` ("completed successfully") and proceeds to `refresh_peer_interest` + `update_peer_summary` — caching a summary the peer never received. Pre-existing semantic conflation: the pre-fix code already fired `completion_tx` on the channel-CLOSED path with the same effect. The PR extends the trigger to the timeout path, making the silent-loss surface larger. Worth fixing or filing as follow-up. |
| `PipeStream` drop on timeout has no completion signal at all | Medium | `p2p_protoc.rs:1340-1396` | Unlike `StreamSend`, `PipeStream` carries no completion oneshot. On timeout we log + drop; the originating op only learns of failure via a downstream stream-level timeout. The PR comment says "stream-level error path will eventually surface a timeout" — that path appears to be the 120s `STREAM_COMPLETION_TIMEOUT`, so dropped `PipeStream`s do produce a 120s broadcast-queue stall per dropped stream. Still better than the pre-fix indefinite event-loop wedge, but worth acknowledging. |
| `peer_send_*` tests test tokio, not freenet | Low | `p2p_protoc.rs:5757-5814` | Both tests construct a local `mpsc::channel(1)` and exercise `timeout(...).await(reserve())` directly — they would pass even if the StreamSend branch were stripped entirely (the source-scrape pins catch that). They document tokio semantics, not freenet coverage. |
| Misleading parameter name `channel_capacity` | Low | `op_state_manager.rs:921` | Parameter is named `channel_capacity` but caller passes `Sender::capacity()` which is *current remaining*, not max. Log inside relabels as `channel_remaining`. Rename parameter to match. |
| `mock_runtime.rs:255` still uses blocking variant | Low | `contract/executor/mock_runtime.rs:254-260` | Test-only path. Future engineers copying the mock as template will reintroduce the antipattern. |
| "Stream fragment" comment is technically wrong | Low | `p2p_protoc.rs:1271-1281` | `StreamSend` carries the whole `Bytes` payload, not a fragment. Cosmetic. |

### Big Picture Assessment
**Goal Alignment:** strong — the PR fixes the **observed** failure mode (cycle, not throughput), which is consistent with #4145's evidence section even though it doesn't match the issue's hypothesized "Phase 2 tx-sharding" design. The fix is a real root-cause fix on the executor side and follows `channel-safety.md` precisely.

**Anti-Patterns Detected:** none (no test removal, no `#[ignore]`, no weakening, no error swallowing).

**Removed Code Concerns:** 62 deletions accounted for entirely by replaced patterns. No functionality regressed.

**Scope Assessment:** focused. Author explicitly defers Fix C (target loop cap) and simulation test, with documented justifications.

**AI Instructions / Rules:** `.claude/rules/channel-safety.md` should get a new case-study entry for the 2026-05-24 nova/vega wedge — the cited-incident count would jump from "4+" to "5+", and the rule's prescribed remediation pattern (timeout / try_send) is exactly what this PR applies.

**#4145 close/reopen:** PR body uses "Refs #4145" not "Fixes #4145", so merge automation won't auto-close. Worth manually reopening #4145 on merge with a comment pointing to the simulation-test follow-up issue (option (b) from big-picture review — keep open until the wedge is exercised end-to-end).

### Documentation
- Code docs: complete — new public/private items have rustdoc with rationale and issue references.
- Architecture docs: `crates/core/architecture.md:142` could use a parenthetical noting broadcast paths use `try_notify_node_event`; nice-to-have, not required.
- User docs: N/A.
- Rules: `.claude/rules/channel-safety.md` should get a 2026-05-24 case study (Big Picture rec #1).

### Recommendations

#### Must Fix (Blocking)
1. **`announce_contract_hosted` one-shot loss.** Revert this specific site to blocking `notify_node_event(...).await` (`operations.rs:160-167`), OR restructure so the `my_contracts.insert` happens only after the announcement is successfully queued. The blocking variant is acceptable here because: announcements are emitted only on first PUT/GET of a new contract (rare), and the existing 30s timeout will return Err under genuine wedge — which is no worse than today's silent loss. Add a comment explaining the deliberate divergence from the other 3 Broadcast sites.
2. **Extend source-scrape pin.** Update `broadcast_emission_sites_use_try_notify` to scan `p2p_protoc.rs`, `mock_runtime.rs`, and `node.rs`. Then either fix or explicitly allowlist each call site the expanded scan flags. At minimum, the `p2p_protoc.rs:4263` retry spawn site should be switched to `try_notify_node_event` — it lives inside a `tokio::spawn` so it can't wedge the event loop, but the PR's design argument cites these retry paths as the healing mechanism and they shouldn't stack 30s blockers themselves.
3. **Add a test that `completion_tx` is fired on the StreamSend timeout path.** This is load-bearing per the PR's own description (the reason `reserve()` was chosen over `timeout(send())`). Without it, a future refactor that forgets the manual fire reintroduces the 120s broadcast-queue stall.

#### Should Fix (Important)
4. Rename `try_notify_node_event_on`'s `channel_capacity` parameter to `channel_remaining` to match what the caller passes (`Sender::capacity()` returns current remaining slots, not max). Trivial, prevents future confusion.
5. Add a 2026-05-24 case-study entry under "The Pattern That Keeps Killing Us" in `.claude/rules/channel-safety.md`. ~5 lines, keeps institutional memory current.
6. Update `mock_runtime.rs:255` to mirror the production path (use `try_notify_node_event`) — or add a comment justifying the deliberate divergence.

#### Consider (Suggestions / Follow-ups)
7. Reword the PR description claim that `try_notify_node_event_returns_err_on_full_channel_without_blocking` is "the test that would have caught the wedge before merge" — it tests the helper in isolation; the source-scrape pin (once expanded) is the real regression guard.
8. The two `peer_send_*` tests in `p2p_protoc.rs` test tokio semantics, not freenet. Either delete them or restructure to invoke the actual dispatch.
9. Fix the "stream fragment" → "stream payload" comment in `StreamSend` (cosmetic).
10. The `SEND_TIMEOUT = 500ms` constant is now repeated in three places (OutboundMessageWithTarget, StreamSend, PipeStream). Consider promoting to a single module-level constant.

#### Follow-up issues to file (not in this PR's scope)
- **Simulation test for event-loop liveness under sustained UPDATE fan-out.** Bug class has shipped 5+ times per `channel-safety.md`; the canonical regression guard remains absent. Author flagged this; please file as a tracking issue at merge.
- **`SyncStateToPeer` heal path uses blocking variant.** The two emission sites in `node.rs` should also move to `try_notify_node_event`. Pre-existing but weakens this PR's stated recovery design.
- **broadcast_queue conflates "permit release" with "delivery success".** Pre-existing semantic issue extended (not introduced) by this PR's timeout path firing `completion_tx`. Cleanest fix: change `oneshot::Sender<()>` to `oneshot::Sender<DeliveryOutcome>` so broadcast_queue can distinguish.
- **Manually reopen #4145** on merge — was closed by #4187 (Phase 1 only); keep open until the simulation test lands.

### Verdict

The fix is the right approach and the diagnosis is well-supported by evidence. But finding #1 (one-shot announcement loss) is a real regression the PR introduces, and finding #2 (source-scrape pin scope) breaks the PR's own stated guarantee that future regressions of this class will be caught. Both need to land before merge. Re-review must run on the updated HEAD per the per-CODE-CONTENT rule.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Re-Review: #4231 (HEAD b9aa9452)

### Summary
- **Type:** fix (concurrency / async — high-risk surface)
- **CI Status:** in progress at fix HEAD
- **Linked Issues:** #4145, builds on #4187; follow-ups #4233, #4234, #4235 filed
- **Review tier:** Full (re-review per per-CODE-CONTENT rule; prior review surfaced 3 Must Fix + 3 Should Fix items, all addressed in b9aa9452)
- **Reviewers run:** code-first, testing, skeptical, big-picture, external (Codex)
- **HEAD SHA reviewed:** b9aa9452d57153edb786d5140d94fc4d42a02381

### Re-review status of prior findings
| Prior finding | Status at b9aa9452 |
|---|---|
| Must Fix #1 — `announce_contract_hosted` one-shot loss (Codex P1) | ✅ Fixed (reverted to blocking with `DELIBERATELY blocking` marker) |
| Must Fix #2 — source-scrape too narrow | ✅ Fixed (4 files scanned; marker-gated allowlist) |
| Must Fix #3 — no `completion_tx` test | ✅ Fixed (behavioral test + source-scrape extended to assert `tx.send(())`) |
| Should Fix #4 — rename `channel_capacity` | ✅ Fixed |
| Should Fix #5 — channel-safety.md case study | ✅ Fixed (incident count 4+ → 5+) |
| Should Fix #6 — mock_runtime.rs divergence | ✅ Fixed (now mirrors prod) |

### New findings (from re-review)

#### Must Fix (Blocking)
1. **Retry-spawn `try_notify` was an overcorrection — revert to blocking.** (Skeptical Medium #1, confirmed in code) — The retry spawn at `p2p_protoc.rs:4267-4278` lives inside `tokio::spawn`, which detaches the blocking from the event loop entirely. The previous blocking `.notify_node_event(...).await` could NOT wedge anything; switching to `try_notify` traded "can't wedge" for "silently drops the retry in the exact recovery path the PR depends on for healing." `MAX_BROADCAST_RETRIES=3` caps spawn-task accumulation, so the original blocking variant has no realistic resource concern. Revert this one site and update the comment to explain why the spawn-isolation makes blocking safe here.

2. **Tighten `tx.send(())` source-scrape — require all 3 arms.** (Skeptical Medium #2 + Code-first Low #1 + Testing Important #1) — `stream_send_branch_uses_timed_reserve_pattern` at `p2p_protoc.rs:5913` uses `body.contains("tx.send(())")` — a single-substring match. The StreamSend dispatch has THREE `tx.send(())` sites (closed / timeout / no-connection arms). A refactor that drops just the timeout arm's fire would still pass the test, reintroducing the 120s STREAM_COMPLETION_TIMEOUT stall the PR was supposed to fix. Switch to `assert_eq!(body.matches("tx.send(())").count(), 3, ...)` and tighten the window so it doesn't bleed into PipeStream.

3. **Add behavioral test for `announce_contract_hosted` under saturation.** (Testing Important #2) — The PR ships a DELIBERATELY blocking exception with a 30s ceiling and zero tests for what happens when the 30s actually fires. The whole rationale for keeping it blocking depends on (a) it returning `Err` (not hanging) when saturated, and (b) the one-shot transition being consumed. Neither is pinned by a test. ~30 LOC.

#### Should Fix (Important)
4. **Land #4234 SyncStateToPeer heal-path fix.** Same bug class, same fix pattern. The PR's design explicitly relies on `SyncStateToPeer` as a healing mechanism for dropped broadcasts — if `SyncStateToPeer` itself can stall 30s on the wedged channel, the heal story is weaker than claimed. ~2 call sites in `node.rs` + extend source-scrape file list.

5. **Add complementary grep-all-files source-scrape.** (Testing Important #3 + Code-first Low + Skeptical Low #4) — Today's pin has a hand-curated file allowlist (4 files); a future refactor moving a Broadcast emission into a new file silently bypasses the gate. Add a complementary test that greps every `.rs` file under `crates/core/src/` for `NodeEvent::Broadcast` and asserts the parent file is in the scrape's allowlist. ~10 LOC.

#### Consider
6. Add "sender-full + receiver-empty" as a named diagnostic fingerprint in `.claude/rules/channel-safety.md` so future operators can grep for it. (Big-picture nice-to-have)
7. Replace `DELIBERATELY blocking` text marker with a doc attribute that survives formatting. (Testing #4) — Worth doing if a second exception is added; premature for one site.
8. `#[tokio::test(start_paused = true)]` for `stream_send_pattern_fires_completion_tx_on_timeout` to eliminate any real-clock flake risk. (Skeptical Low #5)
9. The `peer_send_*` tests still test tokio rather than freenet (pre-existing Low). Defer.
10. Tighten marker-lookback to enclosing function body rather than 2KB byte window. (Skeptical Low #3 + Testing #2(a)) — Premature for one exception.

#### Codex P2 — addressed via clarified scope
Codex re-flagged the `completion_tx`-as-success interpretation in `broadcast_queue`. After investigating: the peer-summary bookkeeping (`refresh_peer_interest` + `update_peer_summary`) in `broadcast_queue.rs:493-518` is gated on `send_result.is_ok()` (the initial dispatch), NOT on the completion-rx outcome. The three completion match arms differ only in log message text. So the timeout path lands in a different log branch but doesn't introduce new corruption — the pre-existing channel-closed path already had this behavior. Filed as #4235 with the correct framing as a pre-existing semantic conflation.

#### Big-picture verdict
"Substantially resolves the root cause; do NOT close #4145 on merge — reopen and tie to follow-up #4233 (simulation test) for explicit success criteria."

### Follow-up issues filed (referenced from PR)
- #4233 — Event-loop liveness simulation test under sustained UPDATE fan-out
- #4234 — SyncStateToPeer heal path uses blocking notify_node_event
- #4235 — broadcast_queue conflates permit release with delivery success

### Verdict

5 issues to address before this can merge:
- **Must Fix:** revert retry-spawn (#1), tighten `tx.send(())` scrape (#2), test `announce_contract_hosted` saturation (#3)
- **Should Fix:** land #4234 inline (#4), add grep-all-files scrape (#5)

Will be bundled into a single fix commit to minimize re-review cycles. Per the per-CODE-CONTENT rule, a fresh review must run on the updated HEAD before merge.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Re-Review (3rd pass): #4231 (HEAD 751794ea)

### Summary
- **Review tier:** Full (per per-CODE-CONTENT rule — 2nd fix commit added ~286 lines)
- **CI Status:** in progress
- **Reviewers run:** code-first, testing, skeptical, big-picture, external (Codex)
- **HEAD SHA reviewed:** 751794ea98a48c2fde8d9b0c1e2dc79edabd792e

### Re-review status of prior pass-2 findings
| Prior finding | Status at 751794ea |
|---|---|
| Must Fix #1 — Retry-spawn try_notify overcorrection | ✅ Reverted to blocking with `DELIBERATELY blocking` marker |
| Must Fix #2 — `tx.send(())` source-scrape too loose | ✅ Now `count == 3` with PipeStream-bounded window |
| Must Fix #3 — No `announce_contract_hosted` saturation test | ✅ `notify_node_event_returns_err_after_timeout_on_saturated_channel` added; deterministic via `#[tokio::test(start_paused = true)]` |
| Should Fix #4 — Land #4234 SyncStateToPeer fix | ✅ Both sites in `node.rs` now use `try_notify_node_event` |
| Should Fix #5 — Complementary grep-all-files scrape | ✅ `broadcast_emission_files_are_in_allowlist` walks `crates/core/src/` |

### Convergence signal
3 of 5 reviewers report **no blocking concerns** (big-picture: "mergeable as-is"; code-first: "no blocking concerns"; Codex: "no actionable regressions"). 2 reviewers (skeptical, testing) found 2 Mediums each, both real but lower-stakes than pass 1/2.

### New findings (pass 3)

#### Must Fix (Blocking)
None — no reviewer raised a Must Fix.

#### Should Fix (Important)
1. **Diagnostic regression in `notify_node_event_on` parameter evaluation order.** (Skeptical M1) — `channel_pending` and `channel_remaining` are evaluated at the call site BEFORE the 30s wait. When the timeout fires, the logged values are from 30 seconds prior — useless for diagnosing whether the channel drained during the wait or stayed full. The pre-extraction `OpManager::notify_node_event` sampled inside the `Err(_)` branch (current values). Fix: re-sample inside the helper at error-time (the accessors are O(1) atomics, no cost).

2. **Two-list allowlist drift risk.** (Skeptical M2 + Testing Important #1) — The new `broadcast_emission_files_are_in_allowlist` test maintains its own file list, which can diverge from the primary scrape's. The error message guides devs hitting a false-positive to "add to ALLOWLIST", silently bypassing the primary scrape's coverage. Plus the walker's silent-empty-pass path means a CI sandbox change disabling `read_dir` would silently pass. Fix: extract a single shared `EMISSION_SITES` constant used by both tests, and assert each allowlisted file is actually visited.

3. **Source-scrape doesn't cover `SyncStateToPeer`.** (Testing Important #2) — Both scrapes gate on `Broadcast` in the forward window. `SyncStateToPeer` (the heal-path sites I just switched to `try_notify` in `node.rs`) is invisible to them. A future refactor reverting either site to blocking would silently reintroduce a 30s wedge contributor on the same channel the executor's try_notify is trying to keep responsive. Fix: extend the primary scrape's forward-window check to also flag `SyncStateToPeer`. Two-character change to the conditional.

#### Consider (Lows)
- **PipeStream needle scope balloon if match arms reorder.** (Skeptical L2 + Testing NTH #4) — Apply the same StreamSend "bound at next match arm" pattern to `pipe_stream_branch_uses_timed_reserve_pattern`. PipeStream has no `tx.send(())` arms today; a refactor that adds them won't bleed correctly without the same scoping.
- **File walker silently skips unreadable files.** (Skeptical L3) — Trivial: `.expect()` the `read_to_string` instead of swallowing errors.
- `tx.send(())` count brittleness (Skeptical L1): acceptable trade-off — loud failure mode if a doc-comment or rename trips it.
- Unrelated blocking `notify_node_event` at `client_events.rs:1746` (Skeptical L4): out of scope.

#### Polish (Code-first nits worth incorporating)
- `_receiver` binding in saturation test could use a one-line "kept alive but not drained" comment.
- Retry-spawn `DELIBERATELY blocking` comment could note the multiplicative `MAX_BROADCAST_RETRIES × concurrent_contracts` blocked-task bound.
- Allowlist scrape error message wording: clarify "if you added a new arm" path so devs don't only think of removal scenarios.

#### Mergeability assessment
| Reviewer | Verdict |
|---|---|
| Big-picture | Mergeable as-is, recommends close #4234 / reopen-then-close #4145 on merge |
| Code-first | No blocking concerns, 5 polish items |
| Codex | "No actionable regressions" |
| Skeptical | 2 Mediums (diagnostic regression + allowlist drift); not Must Fix |
| Testing | 0 Critical, 2 Important (file-walker guard + SyncStateToPeer coverage) |

### Verdict

The pass-3 findings are real but bounded (no Must Fix from any reviewer; 4 Important items, all small and trivially mechanical). Will bundle into one polish commit:
- Skeptical M1: re-sample diagnostic values inside `notify_node_event_on` error arm
- Skeptical M2 + Testing Imp #1: shared `EMISSION_SITES` constant + assert allowlist files were visited
- Testing Imp #2: extend source-scrape to also flag `SyncStateToPeer`
- Skeptical L2/L3 + Testing NTH #4: PipeStream window scoping + `.expect()` on file reads
- Code-first nits: comments + assertion message wording

After commit, a Light re-check (external Codex only, per multi-model-review.md "Light Re-Check Sufficient" criteria: no Must Fix existed, diff additive) is sufficient.

[AI-assisted - Claude]

Issue #4251, opened 2026-05-25
River official-room contract saturates per-contract executor queue, causing network-wide degradation


sanity:
## Summary

The River **official chat room contract** (`4PjqN55KUCidW8vJvw5fhy5fe5maxXKNrWSyK33QjjVq`) is constantly saturating the per-contract executor queue (`MAX_QUEUED_PER_CONTRACT = 100`, `crates/core/src/contract/fair_queue.rs:32`) on every peer subscribed to it. The user-visible effect is that GETs/UPDATEs on the contract fail until the node is restarted — peers enter a degraded state where backpressure is constant and resync never succeeds. Multiple independently-reported user incidents (see below) trace to this contract; this is not instance-specific.

This is the **root-cause** issue. A separate PR will land log-volume reductions so the symptom stops filling users' disks while this is investigated.

## User reports

> **Ivvor**, 2026-05-25, Matrix:
> > I had a bad fail on a long running node overnight, where it looked like everything was OK but most gets were failing and no data appeared to be transferring between nodes. I had to restart the entire system before it would work again. Now, it's just happened on another long running node…
> >
> > i was looking at the logs for the wrong node, but the other one is having the same problems with contract: 4PjqN55KUCidW8vJvw5fhy5fe5maxXKNrWSyK33QjjVq

Two of Ivvor's nodes hit this independently, both pointing at the same contract.

## Evidence (live, gathered 2026-05-25)

Sampled 2000–5000 recent log lines from three monitored peers running the latest release:

**Nova gateway (`5.9.111.215`):** ~10 GB/day
- 598/2000 (~30%) `WARN contract_handling: Rejected event due to per-contract queue capacity limit`
- 209/2000 (~10%) `ERROR Failed to update contract value contract=4PjqN5… error=contract queue full`
- Every single one is for contract `4PjqN5…`

**Vega gateway:** ~10 GB/day
- 995/2000 (~50%) per-contract queue rejection
- 733/2000 `WARN Failed to get contract summary contract=4PjqN5…`
- 131/2000 `ERROR Failed to update contract value` on same contract
- `WARN UPDATE relay: BroadcastTo driver returned error` repeating for same contract

**Technic peer:** 1.26M log lines in one hour
- 1328 `INFO Summary mismatch in interest sync — syncing state to stale peer`
- 1193 `WARN Rejected event due to per-contract queue capacity limit`
- 828 `WARN Failed to get contract summary contract=4PjqN5…`
[...]

PR #4252, opened 2026-05-25
fix: cut river-driven log spam and bound log-dir size (#4251)

sanity:
## Problem

The official River chat-room contract (`4PjqN55KUCidW8vJvw5fhy5fe5maxXKNrWSyK33QjjVq`) has been saturating the per-contract executor queue (`MAX_QUEUED_PER_CONTRACT = 100`) on every subscribed peer. Each rejection generates one `WARN` at the source, a `Failed to update contract value` `ERROR` upstream, and a ~15-line `anyhow` backtrace per resync failure (because `node.rs` was Debug-printing the full `UpdateResponse`, which expands the inner `anyhow::Error`).

Live measurement on monitored nodes (2026-05-25):

| Node | Volume | Top spam |
|---|---|---|
| nova gateway | ~10 GB/day | 30% per-contract queue rejection, 10% update-failed + backtrace |
| vega gateway | ~10 GB/day | 50% per-contract queue rejection, 37% summary-failed |
| technic peer | ~7 MB/h sustained | 27% INFO summary-mismatch, 24% queue rejection |

User reports (Ivvor, two independent nodes) traced to the same contract. Underlying queue saturation tracked separately in #4251; **this PR is the disk-fill symptom mitigation** so users stop losing their drives.

## Approach

Two commits. The first ships the log-volume / size-cap / service-file changes; the second addresses the multi-model review findings (notably the auto-update migration path).

**Commit 1 — symptom mitigation**

- Demote four highest-volume per-event logs to DEBUG (`contract.rs::send_queue_full_response`, `node.rs` "Summary mismatch" / "Failed to get contract summary", `operations/update.rs` "Interest manager peer not found").
- Stop Debug-printing `UpdateResponse` in `node.rs` "Unexpected response to resync update" — switched to `%other` (Display via `ContractHandlerEvent::Display`), kills the per-call ~15-line anyhow backtrace storm.
- Add a 1 GiB **startup-only** disk-size cap to `cleanup_old_logs`. Preserves the most-recently-modified file so it can't race-delete the live rolling-appender output. `enforce_log_dir_size_cap` extracted as a parameterized helper for testability.
- Restrict `cleanup_old_logs`'s file filter to the rolling-appender naming convention (`is_rotating_freenet_log`) so it no longer touches the legacy bare `freenet.log` / `freenet.error.log` files that systemd/launchd hold open.
- Switch systemd `StandardOutput`/`StandardError` to `journal` (Linux) and launchd `StandardOutPath`/`StandardErrorPath` to `/dev/null` (macOS). Tracing layer continues writing its own size-capped rotating files.
- Promote `update.rs::NO_TARGETS` else branch to WARN (with `interest_resolve_failed` counter) so the worst-case "UPDATE will not propagate" outcome stays visible at default level.

**Commit 2 — review-response fixes (Must Fix)**

- **Force unit/plist regen on auto-update when legacy log markers are present.** Both `update.rs::update_service_file` (Linux) and `update.rs::ensure_service_file_updated` (macOS) previously early-returned when the auto-update markers were present — which they are on every v0.2.x install — so existing gateway operators would never receive the new unit. Decision logic extracted as pure helpers `systemd_unit_needs_regen` and `launchd_plist_needs_regen` (per code-style.md preference for pure decision logic over inline platform I/O), each with unit tests for legacy-detection AND current-template no-op.
- **macOS wrapper script** switched from `echo … >> "$LOG"` (fixed unrotated `~/Library/Logs/freenet/freenet.log`) to `logger -t freenet` (auto-rotated by macOS unified logging, queryable via `log show --predicate 'process == "logger"' --info`).
- **Live-file preservation** in `enforce_log_dir_size_cap` and **legacy-bare-name exclusion** in `cleanup_old_logs` (described above) — both surfaced by skeptical review.

Existing users pick up the migration path automatically: on the next `freenet update` (exit-42) cycle, the unit/plist is detected as legacy and rewritten.

## Testing

New regression tests (all passing locally; fmt + clippy `-D warnings` clean):

| Test | Pins |
|------|------|
| `tracer::cleanup_tests::size_cap_deletes_oldest_first_until_under_limit` | size-cap eviction order |
| `tracer::cleanup_tests::size_cap_is_noop_when_under_limit` | no-op when under cap |
| `tracer::cleanup_tests::time_pass_removes_files_older_than_retention` | time-based pass still works |
| `tracer::cleanup_tests::cleanup_ignores_non_freenet_files` | prefix filter |
| `tracer::cleanup_tests::size_cap_preserves_most_recently_modified_file` | live-file preservation |
| `tracer::cleanup_tests::size_cap_deletes_oldest_but_keeps_live` | composition of eviction + live-preservation |
| `tracer::cleanup_tests::cleanup_skips_legacy_bare_freenet_log_names` | bare `freenet.log` is left alone |
| `contract::tests::send_queue_full_response_logs_at_debug_not_warn_pin_test` | source-level pin that the WARN→DEBUG demotion can't silently regress |
| `commands::service::tests::test_systemd_user_service_file_generation` | `StandardOutput=journal` + `!contains("append:")` regression guard |
| `commands::service::tests::test_systemd_system_service_file_generation` | same |
| `commands::service::tests::test_launchd_plist_generation` | `/dev/null` + `!contains` the fixed paths regression guard |
| `commands::service::tests::test_macos_wrapper_script_generation` | `logger -t freenet` + `!contains("Library/Logs/freenet/freenet.log")` |
| `commands::update::tests::systemd_unit_with_legacy_append_log_needs_regen` | legacy-marker detection |
| `commands::update::tests::systemd_unit_with_journal_target_is_up_to_date` | current-template no-op |
| `commands::update::tests::systemd_unit_without_auto_update_markers_needs_regen` | pre-auto-update unit detection |
| `commands::update::tests::launchd_plist_with_legacy_log_paths_needs_regen` | legacy plist detection |
| `commands::update::tests::launchd_plist_with_dev_null_is_up_to_date` | current plist no-op |

## Not in scope

- **Root cause of the queue saturation** — tracked separately in #4251; another agent is investigating.
- **Aggregated periodic INFO backpressure log** (e.g. "rejected N events for contract X in last Ms") — worthwhile follow-up to restore operator visibility lost by the per-event demotions, but out of scope for this disk-fill mitigation.
- **macOS pre-tracing-init panic visibility** — known trade-off of `/dev/null` on launchd; `freenet service report` still captures everything after tracing init.

## Fixes

Mitigates symptom of #4251.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2575,10 +2575,19 @@ SuccessExitStatus=42 43
 # Do NOT restart — the existing instance is healthy.
 RestartPreventExitStatus=43
 
-# Logging - write to files for systems without active user journald
-# (headless servers, systems without lingering enabled, etc.)
-StandardOutput=append:{log_dir}/freenet.log
-StandardError=append:{log_dir}/freenet.error.log
+# Logging
+# - The node's tracing layer writes its own size-capped, hourly-rotated
+#   logs to {log_dir}/freenet.YYYY-MM-DD-HH.log (LOG_RETENTION_HOURS +
+#   LOG_DIR_MAX_BYTES; see crates/core/src/tracing.rs).
+# - systemd's StandardOutput/StandardError previously appended to a fixed
+#   freenet.log / freenet.error.log that the time-based cleanup never
+#   pruned (mtime stayed fresh while the file was being written), so they
+#   grew without bound on long-running nodes (issue #4251).
+# - Routing both to the journal lets journald handle rotation, and panics
+#   or pre-tracing-init output remain queryable via
+#   `journalctl --user-unit freenet`.
+StandardOutput=journal
+StandardError=journal
 SyslogIdentifier=freenet
 
 # Resource limits to prevent runaway resource consumption
@@ -2640,9 +2649,19 @@ SuccessExitStatus=42 43
 # Do NOT restart — the existing instance is healthy.
 RestartPreventExitStatus=43
 
-# Logging - write to files for systems without active user journald
-StandardOutput=append:{log_dir}/freenet.log
-StandardError=append:{log_dir}/freenet.error.log
+# Logging
+# - The node's tracing layer writes its own size-capped, hourly-rotated
+#   logs to {log_dir}/freenet.YYYY-MM-DD-HH.log (LOG_RETENTION_HOURS +
+#   LOG_DIR_MAX_BYTES; see crates/core/src/tracing.rs).
+# - systemd's StandardOutput/StandardError previously appended to a fixed
+#   freenet.log / freenet.error.log that the time-based cleanup never
+#   pruned (mtime stayed fresh while the file was being written), so they
+#   grew without bound on long-running nodes (issue #4251).
+# - Routing both to the journal lets journald handle rotation, and panics
+#   or pre-tracing-init output remain queryable via
+#   `journalctl -u freenet`.
+StandardOutput=journal
+StandardError=journal
 SyslogIdentifier=freenet
 
 # Resource limits to prevent runaway resource consumption
@@ -2916,10 +2935,23 @@ fn generate_plist(wrapper_path: &Path, log_dir: &Path) -> String {
         <key>SuccessfulExit</key>
         <false/>
     </dict>
+    <!--
+        Logging
+        - The node's tracing layer writes its own size-capped, hourly-
+          rotated logs to {log_dir}/freenet.YYYY-MM-DD-HH.log
+          (LOG_RETENTION_HOURS + LOG_DIR_MAX_BYTES; see
+          crates/core/src/tracing.rs).
+        - launchd previously appended to fixed freenet.log / freenet.error.log
+          that the time-based cleanup never pruned, so they grew without
+          bound (issue #4251). macOS does not offer a journal target for
+          launchd, so the cleanest option is /dev/null — diagnostics
+          remain available via `freenet service report`, which collects
+          the rotated tracing logs.
+    -->
     <key>StandardOutPath</key>
-    <string>{log_dir}/freenet.log</string>
+    <string>/dev/null</string>
     <key>StandardErrorPath</key>
-    <string>{log_dir}/freenet.error.log</string>
+    <string>/dev/null</string>
     <key>SoftResourceLimits</key>
     <dict>
         <key>NumberOfFiles</key>
@@ -4054,9 +4086,16 @@ mod tests {
         // Verify it references the correct binary
         assert!(service_content.contains("/usr/local/bin/freenet network"));
 
-        // Verify log paths are set correctly (file-based logging for headless systems)
-        assert!(service_content.contains("/home/test/.local/state/freenet/freenet.log"));
-        assert!(service_content.contains("/home/test/.local/state/freenet/freenet.error.log"));
+        // Logging routes to journal so journald handles rotation. The tracing
+        // layer writes its own size-capped rolling files; routing systemd
+        // stdout/stderr to a fixed freenet.log / freenet.error.log caused
+        // unbounded growth (issue #4251 / log-spam fix).
+        assert!(service_content.contains("StandardOutput=journal"));
+        assert!(service_content.contains("StandardError=journal"));
+        assert!(
+            !service_content.contains("append:"),
+            "regression: must not append systemd output to fixed unrotated file (#4251)"
+        );
 
         // Verify resource limits are set
         assert!(service_content.contains("LimitNOFILE=65536"));
@@ -4125,6 +4164,14 @@ mod tests {
 
         // Verify exit code 43 prevents restart (another instance already running)
         assert!(service_content.contains("RestartPreventExitStatus=43"));
+
+        // Logging routes to journal (same reasoning as the user-unit test).
+        assert!(service_content.contains("StandardOutput=journal"));
+        assert!(service_content.contains("StandardError=journal"));
+        assert!(
+            !service_content.contains("append:"),
+            "regression: must not append systemd output to fixed unrotated file (#4251)"
+        );
     }
 
     #[test]
@@ -4200,9 +4247,19 @@ mod tests {
         // Verify it references the correct binary
         assert!(plist_content.contains("/usr/local/bin/freenet"));
 
-        // Verify log paths are set correctly
-        assert!(plist_content.contains("/Users/test/Library/Logs/freenet/freenet.log"));
-        assert!(plist_content.contains("/Users/test/Library/Logs/freenet/freenet.error.log"));
+        // Stdout/stderr are discarded; the tracing layer writes its own
+        // size-capped rolling logs and `freenet service report` collects
+        // them. Writing launchd output to a fixed freenet.log /
+        // freenet.error.log caused unbounded growth (issue #4251).
+        assert!(plist_content.contains("<string>/dev/null</string>"));
+        assert!(
+            !plist_content.contains("/Users/test/Library/Logs/freenet/freenet.log"),
+            "regression: launchd must not write stdout to a fixed unrotated log file (#4251)"
+        );
+        assert!(
+            !plist_content.contains("/Users/test/Library/Logs/freenet/freenet.error.log"),
+            "regression: launchd must not write stderr to a fixed unrotated log file (#4251)"
+        );
 
         // Verify resource limits are set
         assert!(plist_content.contains("<key>NumberOfFiles</key>"));
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2849,14 +2849,24 @@ CONSECUTIVE_FAILURES=0
 PORT_CONFLICT_KILLS=0
 MAX_PORT_CONFLICT_KILLS=3  # Give up after this many kill attempts
 
-LOG="$HOME/Library/Logs/freenet/freenet.log"
+# Lifecycle messages route through `logger`, which writes to macOS unified
+# logging (auto-rotated, queryable via `log show --predicate 'process ==
+# "logger"' --info`). Previously these were appended to a fixed
+# ~/Library/Logs/freenet/freenet.log that the cleanup pass never touched
+# (its mtime stayed fresh while being written), so the file grew without
+# bound on long-running nodes (issue #4251). The transient
+# freenet.error.log.last scratch file below is overwritten on every launch
+# and so does not accumulate.
+log_event() {{
+    logger -t freenet "$1"
+}}
 
 # Kill any stale freenet network processes before starting.
 # This handles the case where a previous launch daemon restart left a child
 # process still holding the port (e.g. port 7509).
 # Scoped to the current user to avoid killing processes owned by other users.
 if pkill -f -u "$(id -u)" "freenet network" 2>/dev/null; then
-    echo "$(date): Killed stale freenet network process(es) on startup" >> "$LOG"
+    log_event "Killed stale freenet network process(es) on startup"
     sleep 2
 fi
 
@@ -2865,44 +2875,44 @@ while true; do
     EXIT_CODE=$?
 
     if [ $EXIT_CODE -eq 42 ]; then
-        echo "$(date): Update needed, running freenet update..." >> "$LOG"
+        log_event "Update needed, running freenet update..."
         if "{binary}" update --quiet; then
-            echo "$(date): Update successful, restarting..." >> "$LOG"
+            log_event "Update successful, restarting..."
             CONSECUTIVE_FAILURES=0
             PORT_CONFLICT_KILLS=0
             BACKOFF=10
             sleep 2
         else
             CONSECUTIVE_FAILURES=$((CONSECUTIVE_FAILURES + 1))
-            echo "$(date): Update failed (attempt $CONSECUTIVE_FAILURES), backing off $BACKOFF seconds..." >> "$LOG"
+            log_event "Update failed (attempt $CONSECUTIVE_FAILURES), backing off $BACKOFF seconds..."
             sleep $BACKOFF
             BACKOFF=$((BACKOFF * 2))
             [ $BACKOFF -gt $MAX_BACKOFF ] && BACKOFF=$MAX_BACKOFF
         fi
         continue
     elif [ $EXIT_CODE -eq 43 ]; then
-        echo "$(date): Another instance is already running, exiting cleanly" >> "$LOG"
+        log_event "Another instance is already running, exiting cleanly"
         exit 0
     elif [ $EXIT_CODE -eq 0 ]; then
-        echo "$(date): Normal shutdown" >> "$LOG"
+        log_event "Normal shutdown"
         exit 0
     else
         # Check if this looks like a port-already-in-use failure.
         if grep -q "already in use" "$HOME/Library/Logs/freenet/freenet.error.log.last" 2>/dev/null; then
             PORT_CONFLICT_KILLS=$((PORT_CONFLICT_KILLS + 1))
             if [ $PORT_CONFLICT_KILLS -le $MAX_PORT_CONFLICT_KILLS ]; then
-                echo "$(date): Port conflict detected (attempt $PORT_CONFLICT_KILLS/$MAX_PORT_CONFLICT_KILLS) — killing stale freenet process and retrying..." >> "$LOG"
+                log_event "Port conflict detected (attempt $PORT_CONFLICT_KILLS/$MAX_PORT_CONFLICT_KILLS) — killing stale freenet process and retrying..."
                 pkill -f -u "$(id -u)" "freenet network" 2>/dev/null || true
                 sleep 2
                 BACKOFF=10
                 continue
             else
-                echo "$(date): Port conflict persists after $MAX_PORT_CONFLICT_KILLS kill attempts. Manual intervention may be required ('pkill freenet'). Backing off..." >> "$LOG"
+                log_event "Port conflict persists after $MAX_PORT_CONFLICT_KILLS kill attempts. Manual intervention may be required ('pkill freenet'). Backing off..."
             fi
         fi
         CONSECUTIVE_FAILURES=$((CONSECUTIVE_FAILURES + 1))
         PORT_CONFLICT_KILLS=0
-        echo "$(date): Exited with code $EXIT_CODE, restarting after backoff..." >> "$LOG"
+        log_event "Exited with code $EXIT_CODE, restarting after backoff..."
         sleep $BACKOFF
         BACKOFF=$((BACKOFF * 2))
         [ $BACKOFF -gt $MAX_BACKOFF ] && BACKOFF=$MAX_BACKOFF
@@ -2914,7 +2924,7 @@ done
 }
 
 #[cfg(target_os = "macos")]
-fn generate_plist(wrapper_path: &Path, log_dir: &Path) -> String {
+pub(super) fn generate_plist(wrapper_path: &Path, log_dir: &Path) -> String {
     // Note: wrapper_path is the auto-update wrapper script, not the freenet binary directly.
     // The wrapper handles the loop: run freenet, check exit code, update if needed.
     format!(
@@ -4179,6 +4189,23 @@ mod tests {
     fn test_macos_wrapper_script_generation() {
         let binary_path = PathBuf::from("/usr/local/bin/freenet");
         let script = generate_wrapper_script(&binary_path);
+
+        // Regression for issue #4251: lifecycle messages MUST go through
+        // `logger -t freenet` (auto-rotated by macOS unified logging),
+        // not appended to a fixed ~/Library/Logs/freenet/freenet.log file
+        // that grows without bound.
+        assert!(
+            script.contains("logger -t freenet"),
+            "wrapper must route lifecycle messages through logger(1)"
+        );
+        assert!(
+            !script.contains("Library/Logs/freenet/freenet.log\""),
+            "regression: wrapper must NOT append to fixed unrotated freenet.log (#4251)"
+        );
+        assert!(
+            !script.contains(">> \"$LOG\""),
+            "regression: wrapper must not use the legacy LOG file variable (#4251)"
+        );
 
         // Regression for #3301: startup stale-process cleanup, scoped to current user
         assert!(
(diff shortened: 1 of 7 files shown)


sanity (commented):
## Comprehensive PR Review: #4252

### Summary
- **PR Title:** fix: cut river-driven log spam and bound log-dir size (#4251)
- **Type:** fix (symptom-mitigation; root cause tracked in #4251)
- **CI Status:** Clippy / Fmt / Conventional Commits / macOS bundle-swap E2E / Rule Lint / license / snyk all green; Unit & Integration / Simulation / Windows / NAT / Claude Rule Review still pending at this writing
- **Linked Issues:** #4251 (root-cause investigation)
- **Review tier:** Full — touches deploy/release config (systemd unit + launchd plist generators)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four `freenet:*` subagents) plus an external Codex pass — five independent reviewers total

---

### Code-First Analysis
**Independent Understanding:** Six files, ~265 LOC. Four log-level demotions across `contract.rs`, `node.rs`, `operations/update.rs`. A `Debug`→summary swap in `node.rs::"Unexpected response to resync update"` to kill the per-call ~15-line anyhow backtrace. A 1 GiB disk-size cap added to `cleanup_old_logs`, factored into a parameterized `enforce_log_dir_size_cap` helper for testability. systemd unit templates switched from `StandardOutput=append:freenet.log` to `StandardOutput=journal`; macOS plist switched to `/dev/null`. Plus one pre-existing clippy fix in `op_ctx_task.rs` to unblock the PR's own CI.

**Stated Intent:** Stop runaway log spam (~10 GB/day on gateways) from the River queue-saturation symptom (#4251) while root cause is investigated separately.

**Alignment:** Matches at the code level — each of the four claimed sites is touched, the size cap is wired, and the unit/plist templates are flipped. **BUT** the description's claim that "existing users pick up the new unit on next auto-update (`freenet update` regenerates + daemon-reloads)" is **factually false** (see Must Fix #1).

**Gaps:** The new `response_kind = std::any::type_name_of_val(&other)` field returns the same string regardless of which variant fired (`type_name_of_val` only narrows by static type, not variant), so the log line records less information than the previous `?other` did. The comment's reference to "Display" doesn't match the implementation.

---

### Testing Assessment
**Coverage Level:** Adequate for the new code; missing for the user-visible point of the PR.

| Unit | ⚠️ | New size-cap tests are good, but none directly pins the demoted log levels |
| Integration | N/A | |
| Simulation | N/A | |
| E2E | N/A | |

**Regression Test:** Present for the size cap (4 new tests in `tracer::cleanup_tests`) and for the systemd/launchd templates (3 strengthened tests with negative assertions). **Missing** for the level demotions themselves and for the multi-line/backtrace prevention at `node.rs:1798`.

**Missing Tests:**
- Level-pinning (`tracing_subscriber` capture) at each of the four demoted sites — would catch accidental re-promotion in future refactors.
- A `cleanup_old_logs` end-to-end test (not just `enforce_log_dir_size_cap` direct) so a regression in the filter/metadata path is caught.
- Bounded-length assertion for the new `response_kind` log line (prevents accidental return to multi-line `?other` formatting).

---

### Skeptical Findings
**Risk Level:** Medium-High — the PR does not actually help existing users without additional fixes.

|---------|----------|---------|---------|
| Auto-update never regenerates the systemd unit for existing users | **High** | `update.rs:1115-1180` (Linux), `:1188-1260` (macOS) | Both update paths early-return when their respective "auto-update markers" are present, which they are on every v0.2.x install. Existing users — the exact population hit by #4251 — get the size cap (which still helps) but keep `StandardOutput=append:freenet.log` forever. |
| macOS wrapper script writes lifecycle messages to fixed `~/Library/Logs/freenet/freenet.log` | **High** | `service.rs:2852, 2859, 2868, …` | Even with the plist switched to `/dev/null`, the wrapper script itself appends to an unrotated file. Disk-fill on macOS is not actually fixed. |
| Size cap can race-delete the actively-written `freenet.YYYY-MM-DD-HH.log` (or systemd's open `freenet.log`) | **High** | `tracing.rs:4061, 4095-4115` | Sort-by-mtime + delete-from-oldest reaches the live file once the runaway scenario fills enough hours. Linux: keeps writing to deleted inode. Windows: errors. Worse: pattern matches the legacy unrotated `freenet.log` / `freenet.error.log`, which systemd holds open. |
| `type_name_of_val(&other)` carries less info than the variant it replaced | Medium | `node.rs:1798` | All five reviewers flagged this. `ContractHandlerEvent` already implements `Display` (`contract/handler.rs:706`) and is the right substitute. |
| `LOG_DIR_MAX_BYTES` comment claims "hard upper bound" but enforcement is startup-only | Medium | `tracing.rs:3999-4006` | A long-running node will exceed the cap indefinitely. Comment is misleading. |
| `update.rs` justifying comment for the WARN→DEBUG demotion claims "aggregate is logged at INFO below" — but the worst case (zero resolved targets) goes to the DEBUG `else` branch | Medium | `operations/update.rs:271-318` | Verified: only `!result.is_empty()` takes the INFO branch (line 290). The "no targets at all" case loses default-level visibility entirely. |
| macOS pre-tracing-init panics now silent (no fallback like systemd journal) | Low | `service.rs:2935-2960` (plist) | Acceptable trade-off; worth documenting as known follow-up. |

---

### Big Picture Assessment
**Goal Alignment:** Partial. The stated goal is achieved *only for new installs*; the existing-user upgrade path is broken (see Must Fix #1). The PR sells itself as the population-wide disk-fill mitigation; it currently is not.

**Anti-Patterns Detected:**
- Comment-implementation drift: `update.rs:274-277` comment claims "Display" but implementation is `type_name_of_val`; `update.rs:276` claims "aggregate logged at INFO below" but only fires for non-empty case.

**Removed Code Concerns:** None — no tests removed; existing service-template tests strengthened with regression-guard negative assertions.

**Scope Assessment:** Focused. The bundled clippy fix in `get/op_ctx_task.rs` (pre-existing -D warnings failure) is appropriate because it was blocking the PR's own CI and the change is trivially reviewable.

---

### Documentation
- **Code docs:** Mostly complete; `LOG_DIR_MAX_BYTES` rustdoc misleads (see above). New `enforce_log_dir_size_cap` doc could note "designed to be called at startup".
- **Architecture docs:** N/A — no architecture surfaces changed.
- **User docs:** N/A.
- **Skill files:** Audited (`~/.claude/skills/freenet-{gateway-ops,diagnostic-reports,manual-multipeer-test}`) — none reference legacy `freenet.log` paths in a way this PR breaks.

---

### Recommendations

#### Must Fix (Blocking)
1. **Force regeneration of existing units/plists during auto-update when legacy log-config markers are present.** In `update.rs::update_service_file` (Linux) AND `update.rs::ensure_service_file_updated` (macOS), detect the legacy markers — `StandardOutput=append:` / `Library/Logs/freenet/freenet.log` in `StandardOutPath` — and force the rewrite path. Otherwise this PR does nothing for the people it claims to help.
2. **Switch the macOS wrapper script (`service.rs:2840-2912`) to a path that doesn't grow unbounded.** Options: use `logger -t freenet` to route lifecycle messages to macOS unified logging (auto-rotated), OR truncate the `LOG` file at wrapper-startup and document the new behavior. Required so the macOS disk-fill is actually fixed.
3. **Preserve the most-recently-modified file in `enforce_log_dir_size_cap`**, and exclude the bare `freenet.log` / `freenet.error.log` names (these are systemd-owned legacy targets, not rolling-appender files). Otherwise the cap can corrupt or break the active write target.

#### Should Fix (Important)
4. **Replace `std::any::type_name_of_val(&other)` with `%other` (Display)** at `node.rs:1798`. `ContractHandlerEvent` implements `Display`; this restores the variant detail without re-introducing the backtrace expansion that `?other` had.
5. **Soften the `LOG_DIR_MAX_BYTES` rustdoc** (and the test names if needed) to say "startup-enforced backstop" rather than "hard upper bound". Alternative: add a periodic re-enforcement task — preferable but larger scope.
6. **Fix the misleading comment at `operations/update.rs:274-277`** and/or promote the `NO_TARGETS` else branch at `update.rs:307-318` to WARN so the worst case stays visible at default log level.
7. **Add level-pinning test(s)** for at least the contract.rs `send_queue_full_response` site, since that's the highest-volume demotion. A `tracing_subscriber::with_default` capture pattern protects against accidental re-promotion.

#### Consider (Suggestions)
8. Document the macOS pre-tracing-init panic-visibility trade-off in the plist comment block (currently implies `freenet service report` covers everything; it doesn't cover pre-init crashes).
9. File a follow-up issue for aggregated-periodic-INFO backpressure visibility ("rejected N events for contract X in last Ms") to restore operator-level signal lost by the demotions.
10. Add a one-line comment justifying the explicit-arm match in `op_ctx_task.rs:2521-2523` so a future reader doesn't see it as unmotivated clippy appeasement.

---

### Verdict
**HEAD SHA reviewed:** 09bbd284

The three Must Fix items must land before merge — without them the PR does not deliver its stated outcome for the affected user population, and it introduces a race that can corrupt the active log file. Should Fix items #4-#7 are individually small, collectively important, and tightly scoped. After fixes are pushed, re-run the full review on the new HEAD.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Re-Review: #4252 (HEAD `7f07ac12`)

This is a re-review after fixes for the prior round's findings (prior HEAD `09bbd284`). All five reviewers re-ran from a fresh worktree at HEAD `7f07ac12`.

### Summary
- **Review tier:** Full (re-run; touches deploy/release config + auto-update path)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four `freenet:*` subagents) plus an external Codex pass
- **CI status:** Clippy / Fmt / Conventional Commits / macOS bundle-swap E2E / Rule Lint / license / snyk all green at `7f07ac12`; Unit & Integration / Simulation / Windows / NAT / Claude Rule Review still IN_PROGRESS at this writing

### What's fixed since the prior round (verified)

| Prior finding | Status at `7f07ac12` | Notes |
|---|---|---|
| Auto-update never regenerates legacy units/plists | **Addressed** | `systemd_unit_needs_regen` + `launchd_plist_needs_regen` extracted as pure helpers, 5 new unit tests cover legacy-detection AND current-template no-op |
| macOS wrapper writes to fixed `freenet.log` | **Addressed** | Switched to `logger -t freenet`; transient `freenet.error.log.last` scratch retained for port-conflict detection |
| Size cap can race-delete the active rolling file | **Addressed** | `enforce_log_dir_size_cap` pops newest-mtime before eviction loop, accounting via `live_size + non_live_remaining`; 2 new tests pin the invariant |
| `cleanup_old_logs` filter touches systemd-owned legacy bare names | **Addressed** | New `is_rotating_freenet_log` predicate only matches the date-suffixed naming; `cleanup_skips_legacy_bare_freenet_log_names` pins it |
| `type_name_of_val` loses variant info | **Addressed** | Switched to `%other` (Display); `ContractHandlerEvent::Display` verified single-line and chain-free |
| `LOG_DIR_MAX_BYTES` rustdoc claims "hard upper bound" | **Addressed** | Now says "Startup-only backstop" with explicit caveat about long-uptime nodes |
| Misleading "aggregate logged at INFO below" comment | **Addressed** | Comment fixed; NO_TARGETS else-branch promoted to WARN with `interest_resolve_failed` counter |
| No level-pinning test | **Partially addressed** | One site pinned (`send_queue_full_response`); see Should Fix #6 |

### New findings from this round

#### Must Fix (Blocking)

1. **macOS plist rewrite does not trigger `launchctl` reload.** `update.rs:1255-1280` writes the new plist file but launchd uses its cached config for the agent until `launchctl bootout gui/$(id -u) ~/.../org.freenet.node.plist && launchctl bootstrap gui/$(id -u) ~/.../org.freenet.node.plist` is run. This mirrors the Linux `systemctl daemon-reload` step at `update.rs:1188` (which IS present). Without it, the migration takes effect only after user logout / reboot — defeating the "automatic on auto-update" claim for macOS. Add the `launchctl` reload immediately after the plist write.

2. **Plist regenerated BEFORE wrapper update (`update.rs:1255-1280`, then `:1290-1321`).** If wrapper creation, write, or chmod fails, the freshly-written plist may already point at `~/.local/bin/freenet-service-wrapper.sh` while that wrapper is missing, stale, or non-executable, leaving the agent broken. Reorder so the wrapper update succeeds first, then rewrite the plist.

3. **`UPDATE_PROPAGATION: NO_TARGETS` WARN is not streak-suppressed.** Promoted to WARN at `operations/update.rs:316` in commit 2 to restore operator visibility — but `get_broadcast_targets_update` is called up to 4 times per `BroadcastStateChange` (initial + 3 retries), and the outer "no targets after 3 retries, giving up" WARN at `p2p_protoc.rs:4310-4316` already has streak suppression for operator-facing signal. The inner WARN at `update.rs:316` adds a 4× amplification on top of an already-pathological case. **Demote the inner site back to DEBUG** — the outer streak-suppressed WARN plus the `interest_resolve_failed` counter already carry the actionable signal without the per-retry duplication. (Reverses the over-eager promotion from the previous round.)

#### Should Fix (Important)

4. **`systemd_unit_needs_regen` false positive on commented-out directives.** A unit containing `# StandardOutput=append:/old/path  (replaced with journal)` matches the substring and forces regen, destroying customizations. Filter to non-comment lines:

   ```rust
   content.lines()
       .filter(|l| !l.trim_start().starts_with('#'))
       .any(|l| l.contains("StandardOutput=append:") || l.contains("StandardError=append:"))
   ```

5. **Plist read becomes new failure mode on every macOS update (`update.rs:1255-1258`).** Pre-PR the plist was only opened when the wrapper was current; post-PR every invocation does `fs::read_to_string(&plist_path).context("Failed to read plist")?`. A transient FS error kills the whole update path. Wrap in `.ok()` and treat unreadable plist as "needs regen = false" plus a warn-level message.

6. **Pin-test pattern is established for one site only; extend to the others.** The prior round's finding was about *all four* demoted sites. The pattern in `send_queue_full_response_logs_at_debug_not_warn_pin_test` is cheap to extend. At minimum pin:
   - `node.rs:1796` "Unexpected response to resync update" — must contain `response = %other`, must NOT contain `response = ?other` (the prior multi-line backtrace expansion).
   - `node.rs:1546` "Summary mismatch in interest sync" — DEBUG.
   - `update.rs:316` "NO_TARGETS - update will not propagate further" — after #3 above, pin to DEBUG (with the outer p2p_protoc.rs streak-suppressed WARN still at WARN).

7. **Ordering of `launchd_plist_needs_regen` disjuncts.** Second clause is dead — any string containing `Library/Logs/freenet/freenet.error.log` already contains `Library/Logs/freenet/freenet.log` as a substring. Either drop the second OR anchor with a closing-delimiter context (`Library/Logs/freenet/freenet.log<` to disambiguate from the `.error.log` containing substring).

8. **Stale rustdoc on `send_queue_full_response` (`contract.rs:755`).** Still says "Logs a warning" but the implementation at line 766 is now `tracing::debug!`. Update to "Logs at DEBUG and sends the queue-full error response" to prevent it from drifting back.

#### Consider (Suggestions)

9. **macOS `freenet service report` no longer captures wrapper messages.** Previously the wrapper appended lifecycle messages to `~/Library/Logs/freenet/freenet.log` which `report.rs::find_log_files` collected. Now those go to unified logging only. Operators triaging a macOS restart loop via `freenet service report` will see no wrapper-side diagnostic. Two follow-up options: extend `report.rs` to invoke `log show --predicate '(senderImagePath CONTAINS "logger") AND (eventMessage CONTAINS "freenet")' --last 1h`, OR have the wrapper tee to a small, separately-named (NOT `freenet.log`), explicitly-truncated file that report.rs can pick up. For this PR: at minimum, update the plist comment block to note the trade-off explicitly (`log show` is the canonical query path) instead of implying `freenet service report` covers wrapper output.

10. **`cfg`-gating the pure helpers blocks cross-platform CI.** Both helpers operate on `&str` with no platform deps. Per `.claude/rules/deployment.md` ("extract platform-independent decision logic into a pure function that compiles on all targets"), drop the `#[cfg(target_os = ...)]` so non-target CI runners can also exercise them.

11. **Plist/unit regen silently drops user customizations.** Generator output is a fixed template; any operator additions (`EnvironmentVariables`, `Nice`, `WatchPaths`, custom `LimitNOFILE=`, etc.) are dropped without notice. `.bak` mitigates recovery; a one-line `eprintln!("Warning: dropping customized launchd keys: [...]; backup at <path>")` would let operators notice immediately. Pre-existing behavior in the systemd path; the PR doesn't make it worse but the new plist regen path inherits it.

12. **Nova/vega production gateways run custom-named units (`freenet-gateway.service`, `freenet-gateway-hector.service`)** — the migration code only touches `freenet.service`. Worth a manual unit edit on those machines after this lands.

### Verified safe (investigated but not findings)

- **`logger -t freenet` on macOS** — `/usr/bin/logger` is in launchd's default PATH; `os_log`-backed since Sierra; tag length is fine.
- **`ContractHandlerEvent::Display` variants** — all single-line, no Debug delegation, no anyhow chain expansion. The `%other` switch is safe.
- **Live-file race condition** — cleanup runs before the rolling appender is constructed in `init_tracer`. No active writer at cleanup time.
- **`.bak` chain accumulation** — `with_extension` semantics replace the extension; only one backup file ever exists at a time.
- **`live` "unused assignment"** — false alarm in the brief; `live` is read on the next line.
- **System-mode unit auto-regen EPERM** — pre-existing behavior, not introduced by this PR.

### Verdict
**HEAD SHA reviewed:** 7f07ac12

The migration logic correctly delivers the originally-claimed outcome for the population using `freenet service install`-generated units, but **Must Fix #1 (macOS `launchctl` reload), #2 (plist/wrapper ordering), and #3 (NO_TARGETS amplification)** are real correctness/quality issues that should land before merge. Should Fix items are individually small and tightly scoped. After fixes are pushed, re-run the review on the new HEAD.

[AI-assisted - Claude]

sanity (commented):
## Round-3 Focused Re-Review: #4252 (was HEAD `24c9f982`, fixed at `f22dcffc`)

Focused (skeptical + Codex) re-review of the changes in commit `24c9f982` after the v2 round was addressed. Scope: only new code in `reload_launchd_agent`, reordered macOS update path, comment-line filter, plist read fail-soft, dropped disjunct, pin tests.

### Critical bug caught

**`reload_launchd_agent` self-terminates during the auto-update path** (skeptical, verified).

The call chain is:

```
launchd → wrapper.sh (LaunchAgent job)
       → freenet network → exits 42
       → freenet update (child of wrapper)
       → reload_launchd_agent
       → launchctl bootout gui/<uid> <plist>
       → launchd terminates the wrapper tree
       → freenet update is killed mid-call
       → launchctl bootstrap never runs
```

So the exact case the reload existed for — automatic reload during `freenet update` triggered by exit-42 — silently fails. Manual `freenet update` from a terminal works (not in launchd's job tree), which is what made this brittle in interactive testing.

**Alternatives all have problems:**
- `launchctl kickstart -k` restarts the running job but does NOT re-read the plist (uses cached in-memory config).
- Detached helpers via `nohup` / double-fork race the bootout and add complexity for a moderate-impact change.

**Fix: revert the auto-reload entirely.** New plist lands on disk; user is told the exact two-line command to apply it immediately. Next user-driven reload or system reboot picks it up automatically. The dominant disk-fill source on macOS was the wrapper-script writing to a fixed `freenet.log` — which IS fixed at the next wrapper iteration via the `logger -t freenet` change from commit 1. The plist's `StandardOutPath` only captures the freenet binary's stdout/stderr, which is minimal post-tracing-init.

### Second bug caught

**`launchd_plist_needs_regen` lost coverage of stderr-only legacy plists** (Codex, verified).

The v2 round flagged the second disjunct as "dead code" because `Library/Logs/freenet/freenet.error.log` was claimed to contain `Library/Logs/freenet/freenet.log` as a substring. It does NOT — the `.` before `error` breaks the contiguous match. A partially-migrated plist (stdout already `/dev/null`, stderr still pointing at the legacy `freenet.error.log`) would have skipped regen and kept growing.

**Fix:** restored both substring checks; added `launchd_plist_with_stderr_only_legacy_path_still_needs_regen` to pin the fix.

### Other addressed findings

- **L2 generator-loop pin** (skeptical): new `freshly_generated_plist_does_not_trigger_regen_loop` calls the real `generate_plist` and asserts the output does NOT match the legacy-path predicate — prevents future template edits from causing every auto-update to loop rewriting the plist.
- **L3 pin-test window margin** (skeptical): bumped `assert_log_site_pin` in `node.rs` from 240 to 400 bytes so future structured-field additions don't false-break the pin.

### Findings accepted as-is (not fixed)

- **M1 (`unload -w` flag persistence) and M2 (silent `eprintln!` under `quiet`)** — moot since the entire `reload_launchd_agent` code is gone.
- **M3 (macOS path platform-untested in CI)** — true; no test framework on macOS for launchd interaction. Best mitigation is the smoke-test guidance in the user-facing instruction message, plus the freshly-generated-plist regen-loop pin test that catches the most likely template regression.
- **L1 (plist read fail-soft skip-vs-force-regen tradeoff)** — defensible chosen behavior (transient FS error shouldn't trigger spurious writes); added inline comment clarifying the intent.

### Verified safe

- `libc::getuid()` removed along with `reload_launchd_agent`, so the SAFETY-comment concern is moot.
- The reordered macOS update path correctly fails-closed: wrapper write error returns via `?` before the plist regen runs.
- `systemd_unit_needs_regen` comment-line filter test (`systemd_unit_with_commented_out_append_does_not_force_regen`) passes.
- Plist read fail-soft is correctly wrapped in `match`/`Ok`/`Err`; transient FS errors no longer kill the auto-update path.
- All 3 new pin tests (`summary_mismatch`, `unexpected_resync_response`, `no_targets_propagation`) pass and accurately reflect their respective source sites.
- The outer streak-suppressed WARN at `p2p_protoc.rs:4310` that the NO_TARGETS demote-to-DEBUG relies on actually exists and fires as `BROADCAST_NO_TARGETS: no targets found after N retries, giving up`.

### Verdict
**State:** Ready to merge once CI is green on `f22dcffc`.
**HEAD SHA reviewed:** `f22dcffc`

No remaining blocking findings. The PR now actually delivers the originally-claimed mitigation for the affected population:
- **Linux users**: auto-update detects the legacy `append:` directive, regenerates the unit with `StandardOutput=journal`, and `systemctl daemon-reload`s. Takes effect on the next service restart (within seconds of auto-update completing).
- **macOS users**: auto-update regenerates the wrapper script (immediately effective on next wrapper iteration — the dominant fix) AND regenerates the plist. The plist takes effect on next reboot or via the printed `launchctl bootout/bootstrap` instruction.
- **All users**: the four log-level demotions + the `?other` → `%other` fix take effect immediately on the new binary; the 1 GiB startup disk-size cap kicks in on next restart.

[AI-assisted - Claude]

PR #4253, opened 2026-05-25
fix(contract): stop queue-full saturation from amplifying network-wide

### LINKED ISSUE #4251: River official-room contract saturates per-contract executor queue, causing network-wide degradation

sanity:
## Problem

The per-contract executor fair queue (`MAX_QUEUED_PER_CONTRACT = 100` in `crates/core/src/contract/fair_queue.rs:32`) is doing its job — round-robin scheduling so one contract can't monopolize the WASM executor — but the rest of the codebase treats a queue-full rejection as a **real executor failure**. When a single contract saturates its own queue (production case: `4PjqN55KUCidW8vJvw5fhy5fe5maxXKNrWSyK33QjjVq`, 275k log mentions/hour on `nova` per issue #4251), the UPDATE relay drivers spuriously fire **two amplification paths**:

1. `try_auto_fetch_contract` (full-state branch, `update/op_ctx_task.rs:1107`, plus streaming branch via `log_broadcast_to_streaming_failure`). Reason: queue-full is mis-classified as "contract code missing locally" by `!is_contract_exec_rejection()`. The fetch enqueues a GET right back onto the saturated handler.
2. `InterestMessage::ResyncRequest` (delta branch, `update/op_ctx_task.rs:1064`). Asks the sender to resend **full** state — a bigger payload — onto the same queue that just rejected a delta.

The combination is the platform-harm vector: a buggy / abusive / merely-popular contract that saturates **its own** queue induces every subscribed peer to fire spurious GETs and request larger UPDATEs, turning a single-contract problem into a network-wide storm. Originator-side logging compounds: queue-full hits the `ERROR Failed to update contract value` path hundreds of times per second, drowning real failures.

## Approach

Foundation — typed error so the platform can recognize the condition without string-matching:

- New `contract::ContractQueueFull` marker (in `executor.rs`).
- `ExecutorError::is_contract_queue_full()` downcasts on it. Mirror up through `OpError::is_contract_queue_full()` for callers operating on the wider error type.
- `send_queue_full_response` now constructs `ExecutorError::other(ContractQueueFull)` instead of an opaque `anyhow::anyhow!("contract queue full...")`.

Suppression — gate the amplification on the typed predicate:

- `drive_relay_broadcast_to`: compute `queue_full = err.is_contract_queue_full()` once, gate both the delta→ResyncRequest branch and the full-state→auto-fetch branch on `!queue_full`. The relay still surfaces the error to its caller for telemetry; it just stops re-throwing work at the saturated handler.
- `log_broadcast_to_streaming_failure` (helper used by `drive_relay_broadcast_to_streaming`): returns `!err.is_contract_exec_rejection() && !err.is_contract_queue_full()`, so the streaming variant also skips auto-fetch on queue-full.

Log severity — queue-full is transient backpressure, not an operator-actionable failure:

- Originator and streaming paths both drop queue-full from ERROR / WARN to DEBUG with `event=queue_full`. Real WASM faults (OOG, traps, missing parameters) stay at ERROR/WARN unchanged.

## Testing

New regression tests (all pass locally):

- `executor::tests::executor_error_tests::test_contract_queue_full_*` — five tests pinning the predicate: recognises the typed marker, rejects an anyhow string lookalike (so an unrelated stringly-typed error can't accidentally inherit queue-full semantics), disjoint from every other predicate.
- `contract::tests::send_queue_full_response_carries_typed_queue_full_marker` — pins the typed marker at the channel boundary. Regression here silently re-enables the storm.
- `operations::update::tests::log_severity::update_contract_failure_logs_debug_for_queue_full` — originator log goes to DEBUG with `event=queue_full`, no `ERROR` and no `Failed to update contract value`.
- `operations::update::tests::log_severity::broadcast_to_streaming_failure_skips_auto_fetch_and_uses_debug_for_queue_full` — verifies both the log-level downgrade AND `log_broadcast_to_streaming_failure` returns `false` (auto-fetch skipped).
- `operations::update::op_ctx_task::tests::broadcast_to_suppresses_amplification_on_queue_full` — source-grep pin asserting `is_contract_queue_full()` appears in `drive_relay_broadcast_to` and both amplification branches still exist (gated, not deleted).

Pre-existing tests still pass — real failures (OOG, missing parameters, contract WASM rejections) keep their previous logging and trigger auto-fetch / ResyncRequest exactly as before.

Full lib test: `cargo test -p freenet --lib` → 2607 passed, 0 failed, 14 ignored.

## Contract identification correction

The issue body identifies `4PjqN5…` as the official River room. It isn't — the official River room is `7Z8Nf1Fx5WsGRfc1AzH23bQBXDhmQFt8FgAKGw9poWXN`. On `nova`'s last-hour log the official room has 249 mentions (with `BroadcastToStreaming` at ~160 KB payloads), while `4PjqN5…` has 275,632 mentions with no streaming (state < 64 KB). I posted the correction on the issue: https://github.com/freenet/freenet-core/issues/4251#issuecomment-4536621916. Identifying what `4PjqN5…` actually is remains a separate investigation; this fix is platform-resilience-against-any-such-contract, irrespective of which one is misbehaving.

## Out of scope (deliberately)

- WASM throughput vs. update-rate gap on `4PjqN5…`. Needs profiling and is contract-specific.
- `MAX_QUEUED_PER_CONTRACT` tuning. Bigger queues delay but don't break the saturation feedback loop.
- Log-volume reduction for `WARN Rejected event due to per-contract queue capacity limit` in `contract.rs:761`. Sibling PR already flagged in the issue.
- Per-(peer, contract) rate-limit on `SyncStateToPeer` emission. Investigated; the source-side emit is fire-and-forget non-blocking, so it's not the dominant pressure on the contract queue. Separate improvement.
- Bumping CI's pinned Rust toolchain from 1.93.0 to current stable (one new clippy lint trips locally on pre-existing code). Will follow up after this lands.

Closes #4251

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -31,7 +31,7 @@ pub(crate) use handler::{
     },
 };
 
-pub use executor::{Executor, ExecutorError, OperationMode};
+pub use executor::{ContractQueueFull, Executor, ExecutorError, OperationMode};
 pub use handler::reset_event_id_counter;
 
 use freenet_stdlib::client_api::DelegateRequest;
@@ -762,7 +762,13 @@ async fn send_queue_full_response(
         event = %rejected.event,
         "Rejected event due to per-contract queue capacity limit"
     );
-    let make_err = || ExecutorError::other(anyhow::anyhow!("contract queue full, try again later"));
+    // Use the typed `ContractQueueFull` marker so callers can distinguish
+    // queue saturation from real executor failures via
+    // `ExecutorError::is_contract_queue_full`. This is what suppresses the
+    // auto-fetch / ResyncRequest amplification in the UPDATE relay drivers
+    // (op_ctx_task.rs) when a single contract is saturating its own queue.
+    // See issue #4251.
+    let make_err = || ExecutorError::other(ContractQueueFull);
     let response = match &rejected.event {
         ContractHandlerEvent::PutQuery { .. } => ContractHandlerEvent::PutResponse {
             new_value: Err(make_err()),
@@ -1797,6 +1803,56 @@ mod tests {
                 assert!(delta.is_err(), "should be an error response");
             }
             other => panic!("expected GetDeltaResponse, got {other}"),
+        }
+    }
+
+    /// Issue #4251: the queue-full response MUST carry the typed
+    /// `ContractQueueFull` marker, not just an opaque anyhow error, so
+    /// callers can suppress amplification side effects (auto-fetch,
+    /// ResyncRequest, ERROR logs) via
+    /// `ExecutorError::is_contract_queue_full`. A regression here would
+    /// silently bring back the storm-on-saturation behavior.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn send_queue_full_response_carries_typed_queue_full_marker() {
+        let (send_halve, mut rcv_halve, _) = handler::contract_handler_channel();
+        let key = make_contract_key();
+
+        let (rejected, handle) = setup_rejected_event(
+            send_halve,
+            &mut rcv_halve,
+            ContractHandlerEvent::UpdateQuery {
+                key,
+                data: UpdateData::Delta(StateDelta::from(vec![1])),
+                related_contracts: RelatedContracts::default(),
+            },
+        )
+        .await;
+
+        send_queue_full_response(&mut rcv_halve, rejected).await;
+
+        let response = tokio::time::timeout(Duration::from_millis(200), handle)
+            .await
+            .expect("timeout")
+            .expect("task should complete")
+            .expect("should get response");
+
+        match response {
+            ContractHandlerEvent::UpdateResponse { new_value, .. } => {
+                let err = new_value.expect_err("UpdateResponse should carry an error");
+                assert!(
+                    err.is_contract_queue_full(),
+                    "queue-full response MUST be classified by is_contract_queue_full; \
+                     got err = {err:?}. If you removed this classification, you have also \
+                     re-enabled the auto-fetch / ResyncRequest amplification storm — see \
+                     issue #4251."
+                );
+                // The other predicates must NOT also match, otherwise the
+                // gating in op_ctx_task.rs becomes ambiguous.
+                assert!(!err.is_contract_exec_rejection());
+                assert!(!err.is_invalid_update_rejection());
+                assert!(!err.is_missing_contract_parameters());
+            }
+            other => panic!("expected UpdateResponse, got {other}"),
         }
     }
 
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -762,12 +762,8 @@ async fn send_queue_full_response(
         event = %rejected.event,
         "Rejected event due to per-contract queue capacity limit"
     );
-    // Use the typed `ContractQueueFull` marker so callers can distinguish
-    // queue saturation from real executor failures via
-    // `ExecutorError::is_contract_queue_full`. This is what suppresses the
-    // auto-fetch / ResyncRequest amplification in the UPDATE relay drivers
-    // (op_ctx_task.rs) when a single contract is saturating its own queue.
-    // See issue #4251.
+    // Typed `ContractQueueFull` marker powers `is_contract_queue_full` at
+    // every caller; the UPDATE relay drivers gate amplification on it. #4251.
     let make_err = || ExecutorError::other(ContractQueueFull);
     let response = match &rejected.event {
         ContractHandlerEvent::PutQuery { .. } => ContractHandlerEvent::PutResponse {
@@ -1846,11 +1842,9 @@ mod tests {
                      re-enabled the auto-fetch / ResyncRequest amplification storm — see \
                      issue #4251."
                 );
-                // The other predicates must NOT also match, otherwise the
-                // gating in op_ctx_task.rs becomes ambiguous.
-                assert!(!err.is_contract_exec_rejection());
-                assert!(!err.is_invalid_update_rejection());
-                assert!(!err.is_missing_contract_parameters());
+                // Disjointness from other predicates is covered by
+                // `test_contract_queue_full_disjoint_from_other_predicates`
+                // in executor.rs.
             }
             other => panic!("expected UpdateResponse, got {other}"),
         }
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -99,6 +99,28 @@ pub(crate) use init_tracker::{
 };
 pub(crate) use runtime::RuntimePool;
 
+/// Marker error returned by `send_queue_full_response` when the
+/// per-contract fair queue rejects an event because the contract's
+/// queue (or the global queue) has reached capacity.
+///
+/// Wrapping this in `ExecutorError::other(ContractQueueFull)` instead
+/// of an opaque `anyhow::anyhow!("contract queue full...")` lets
+/// callers distinguish queue saturation from real executor failures
+/// (OOG, traps, missing parameters, storage errors). Queue-full is a
+/// transient platform-level condition — not a contract-level fault —
+/// so callers should avoid amplification (auto-fetch, ResyncRequest)
+/// and ERROR-level logging on this path. See issue #4251.
+#[derive(Debug, Clone, Copy)]
+pub struct ContractQueueFull;
+
+impl std::fmt::Display for ContractQueueFull {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("contract queue full, try again later")
+    }
+}
+
+impl std::error::Error for ContractQueueFull {}
+
 #[derive(Debug)]
 pub struct ExecutorError {
     inner: Either<Box<RequestError>, anyhow::Error>,
@@ -268,6 +290,38 @@ impl ExecutorError {
                     if cause.starts_with("execution error: invalid contract update")
             ),
             Either::Right(_) => false,
+        }
+    }
+
+    /// Returns true if this error was produced by the per-contract fair
+    /// queue rejecting an event because the queue is at capacity (see
+    /// `send_queue_full_response` in `contract.rs` and `ContractQueueFull`
+    /// above).
+    ///
+    /// Queue-full is a transient platform-level backpressure signal — not
+    /// a contract-level fault, not a missing-contract condition, not a
+    /// WASM execution failure. Callers MUST use this predicate to suppress
+    /// amplification side effects that only make sense for real failures:
+    ///
+    /// - The UPDATE full-state relay path must NOT trigger
+    ///   `try_auto_fetch_contract` (the contract code is present; the
+    ///   queue is just busy).
+    /// - The UPDATE delta relay path must NOT emit `ResyncRequest`
+    ///   (asking the sender for full-state resends makes the storm
+    ///   worse).
+    /// - Originator-side error logging should drop from ERROR to DEBUG
+    ///   (a temporarily-saturated executor is not an operator-actionable
+    ///   alert; the volume on a hot contract can be enormous).
+    ///
+    /// Without this predicate, a single buggy / abusive / merely-popular
+    /// contract that saturates its own queue would induce the local peer
+    /// to fire auto-fetch GET storms and ResyncRequest broadcasts that
+    /// amplify the saturation onto every other peer subscribed to the
+    /// contract — see issue #4251.
+    pub fn is_contract_queue_full(&self) -> bool {
+        match &self.inner {
+            Either::Left(_) => false,
+            Either::Right(err) => err.downcast_ref::<ContractQueueFull>().is_some(),
         }
     }
 
@@ -958,6 +1012,80 @@ mod tests {
                 !err.is_fatal(),
                 "MaxComputeTimeExceeded must not be fatal - it would kill the entire contract handler"
             );
+        }
+
+        // ── ContractQueueFull predicate (issue #4251) ─────────────────────
+        //
+        // The queue-full marker must be cleanly distinguishable from every
+        // other error class so callers can suppress amplification side
+        // effects (auto-fetch, ResyncRequest, ERROR logs) specifically for
+        // this case without paying for them on real failures.
+
+        #[test]
+        fn test_contract_queue_full_true_for_marker_error() {
+            let err = ExecutorError::other(ContractQueueFull);
+            assert!(
+                err.is_contract_queue_full(),
+                "ContractQueueFull marker MUST be recognized by is_contract_queue_full"
+            );
+            // Display message preserved for human-readable surface (logs, etc.)
+            assert!(err.to_string().contains("contract queue full"));
+        }
+
+        #[test]
+        fn test_contract_queue_full_false_for_anyhow_string_lookalike() {
+            // Regression: an opaque anyhow error whose message happens to
+            // contain "contract queue full" must NOT satisfy the predicate.
+            // The predicate is typed (downcast), not string-matched, so a
+            // future caller that produces a similarly-worded error doesn't
+            // accidentally inherit queue-full semantics.
+            let err = ExecutorError::other(anyhow::anyhow!("contract queue full, try again later"));
+            assert!(
+                !err.is_contract_queue_full(),
+                "Anyhow string with matching prose must NOT satisfy the typed predicate"
+            );
+        }
+
+        #[test]
+        fn test_contract_queue_full_false_for_invalid_update() {
+            let key = test_fixtures::make_contract_key();
+            let err = ExecutorError::request(StdContractError::update_exec_error(
+                key,
+                "invalid contract update, reason: stale",
+            ));
+            assert!(
+                !err.is_contract_queue_full(),
+                "Benign WASM invalid-update rejection is not queue-full"
+            );
+        }
+
+        #[test]
+        fn test_contract_queue_full_false_for_missing_parameters() {
+            let key = test_fixtures::make_contract_key();
+            let err = ExecutorError::request(StdContractError::Update {
+                key,
+                cause: "missing contract parameters".into(),
+            });
+            assert!(
+                !err.is_contract_queue_full(),
+                "Missing contract parameters is not queue-full"
+            );
+        }
+
+        #[test]
+        fn test_contract_queue_full_disjoint_from_other_predicates() {
+            // Ensures the new predicate doesn't collide with the existing
+            // predicates' truth values for the queue-full case. This is the
+            // load-bearing property used in op_ctx_task.rs:1064/1107 to
+            // suppress amplification.
+            let err = ExecutorError::other(ContractQueueFull);
+            assert!(err.is_contract_queue_full());
+            assert!(!err.is_request());
+            assert!(!err.is_contract_exec_rejection());
+            assert!(!err.is_invalid_update_rejection());
+            assert!(!err.is_missing_contract_parameters());
+            assert!(!err.is_missing_delegate());
+            assert!(!err.is_fatal());
         }
     }
 
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -289,16 +289,22 @@ impl ExecutorError {
         }
     }
 
-    /// Returns true if this error is the typed `ContractQueueFull` marker
-    /// from `send_queue_full_response` (per-contract fair queue at capacity).
+    /// Returns true if this error is the typed `ContractQueueFull` marker.
+    ///
+    /// Produced by:
+    /// - `send_queue_full_response` (per-contract fair queue at capacity),
+    /// - the `InitCheckResult::QueueFull` arm in `executor/runtime.rs` (per-contract
+    ///   initialization queue at capacity).
     ///
     /// **Platform-resilience invariant**: queue-full is transient
     /// backpressure, not a contract-level fault, missing-contract condition,
-    /// or WASM failure. Callers MUST gate amplification side effects on this
-    /// predicate — auto-fetch GETs (the contract code is present), ResyncRequest
-    /// broadcasts (the merge never ran), and ERROR-level logging (volume on a
-    /// hot contract drowns real failures). Without it, a single saturated
-    /// contract induces a network-wide storm. See issue #4251.
+    /// or WASM failure. Callers in paths that have amplification side effects
+    /// (today: UPDATE relay's `try_auto_fetch_contract` and `ResyncRequest`)
+    /// MUST gate those branches on this predicate so a saturated contract
+    /// doesn't induce a network-wide storm. Paths without amplification
+    /// (today: PUT, GET, SUBSCRIBE) only need to gate **ERROR-level logging**
+    /// off this predicate, since on a hot contract the volume otherwise drowns
+    /// real failures. See issue #4251.
     pub fn is_contract_queue_full(&self) -> bool {
         match &self.inner {
             Either::Left(_) => false,
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -99,17 +99,13 @@ pub(crate) use init_tracker::{
 };
 pub(crate) use runtime::RuntimePool;
 
-/// Marker error returned by `send_queue_full_response` when the
-/// per-contract fair queue rejects an event because the contract's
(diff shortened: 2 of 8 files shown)


sanity (commented):
## Testing Review: PR #4253

### Coverage Summary
- Direct changes (predicate, log severity, helper return value): **covered**
- Driver-level runtime gating (`drive_relay_broadcast_to`): **partial** — only pinned via source-string grep, not behaviorally exercised
- End-to-end amplification suppression on a saturated executor: **missing**

### Bug reproduction
The 5 predicate tests and 2 log-severity tests would reliably fail without the typed marker / severity changes — they are real regression tests, not coverage-padding. The `send_queue_full_response_carries_typed_queue_full_marker` test is the strongest: it exercises the actual rejection channel, not a hand-built `ExecutorError`.

The driver-level pin test (`broadcast_to_suppresses_amplification_on_queue_full`) is the weak point. It greps `op_ctx_task.rs` source for `is_contract_queue_full()` plus the two amplification site names. A refactor that keeps the call but inverts the gate (`if queue_full { ResyncRequest }`) would still pass. The test honestly documents itself as a "structural pin", but the runtime invariant it pins is the central claim of the fix.

### Test Gaps Found

#### Important (should have)
1. **Behavioral test for `drive_relay_broadcast_to` queue-full path.** Even a unit-level test with a stub `OpManager` / mocked `update_contract` returning `ExecutorError::other(ContractQueueFull)` would verify: (a) no `InterestMessage::ResyncRequest` is sent via `notify_node_event`, (b) `try_auto_fetch_contract` is not called, (c) `update_peer_summary(.., None)` is not called (this side effect lives inside the now-gated delta branch — only its absence keeps the suppression behaviorally clean). The existing tests at `op_ctx_task.rs:2306+` for `delta_apply_failed` / auto-fetch already use a stubbed driver shape this could mirror.
2. **Driver-level streaming gating not exercised at runtime.** Only `log_broadcast_to_streaming_failure`'s return-value contract is tested. A test asserting the streaming driver's `else if err.is_invalid_update_rejection()` arm does not fire on queue-full (and therefore `send_summary_back_on_rejection` is not spawned) would close the symmetric gap.
3. **Simulation test with a saturated per-contract queue** receiving a `BroadcastTo` — at least one such test would catch a regression in the integration of these typed-error gates with the real executor. Not strictly required if (1) lands.

#### Nice to have
- Disjointness test extension covering `is_contract_exec_rejection()` — currently asserted disjoint from 5 other predicates but not that one, even though the streaming helper's `&& !err.is_contract_exec_rejection()` clause relies on it.

### Regression risk
The delta branch's `update_peer_summary(&key, &sender_key, None)` (op_ctx_task.rs:1091) is now suppressed on queue-full. This is correct (the merge never ran, so we shouldn't invalidate our cache of the sender's summary), but it's a behavioral side effect the PR description doesn't flag and no test covers. Worth either a comment at the call site or an assertion in the suggested behavioral test above.

`send_summary_back_on_rejection` is gated on `is_invalid_update_rejection` (disjoint from queue-full per `test_contract_queue_full_disjoint_from_other_predicates`), so unaffected.

### CI Considerations
No concerns — the new tests are unit-level and fast. The suggested behavioral test would add <100ms.

### Verdict
Solid predicate/severity coverage; the driver-level gating deserves at least one behavior-level test (gap #1) before merge. Source-grep pins protect against accidental deletion but not against gate-inversion regressions, and gate inversion is exactly the failure mode that would silently re-enable the storm.

[AI-assisted - Claude]

sanity (commented):
## Big Picture Review: PR #4253

**Verdict: APPROVE (with one minor scope nit).** This is a tightly-scoped, well-motivated platform-resilience fix. No CI-chasing, no removed code, no stale-docs drift.

### Goal alignment

The linked issue #4251 lists four hypotheses for the storm dynamic; this PR addresses **hypothesis 2** (retry storm with no backoff at the rejection path) by giving callers a typed predicate to recognise queue-full and suppress two amplification branches (auto-fetch GET, ResyncRequest). The PR description is honest about what's NOT addressed (WASM throughput, queue tuning, `SyncStateToPeer` heartbeat, `Summary mismatch` chain) and correctly notes those are separate work. This is the right slice to land first — it breaks the *self-feeding* part of the loop so other amplification sources stop being multiplied by the rejection path.

Code matches the description faithfully. Both amplification branches at `op_ctx_task.rs:1072` (ResyncRequest) and `op_ctx_task.rs:1114` (auto-fetch) are gated; the `BroadcastToStreaming` path at `op_ctx_task.rs:1738` inherits the gate via `log_broadcast_to_streaming_failure` (`update.rs:422`). Other `try_auto_fetch_contract` call sites in the file (`op_ctx_task.rs:378`, originator-side, uses the narrower `is_missing_contract_parameters` predicate which is disjoint from `ContractQueueFull` per the new `test_contract_queue_full_disjoint_from_other_predicates`) are correctly NOT affected.

### Removed code / anti-patterns

None. Both amplification branches remain in source — gated, not deleted — and the structural pin test (`broadcast_to_suppresses_amplification_on_queue_full`) asserts that explicitly. Net +321 LOC, of which the bulk is tests. Zero `#[ignore]`, zero weakened assertions, no commented-out code. Severity downgrade (ERROR/WARN → DEBUG) is justified by the issue evidence (~30% of nova's log volume).

### Contract identification

The PR correctly flags that issue #4251's identification of `4PjqN5...` as the official River room is wrong (it's not — `7Z8Nf1...` is). Not load-bearing for the fix design (the platform behaviour is contract-agnostic) but the correction is the right thing to do for the historical record. Identifying what `4PjqN5...` actually is should be a separate sibling issue, as the PR notes.

### AI instruction drift

None. `.claude/rules/bug-prevention-patterns.md` is scoped to `crates/core/src/bin/**` (Windows console-handle issues) and is unaffected. `.claude/rules/contracts.md` mentions amplification and `ResyncRequest` in the AFT-fetch context (PR #4006), which is unchanged. `crates/core/CLAUDE.md` references `start_relay_broadcast_to` by name and is still accurate. No skill or rule update required.

**Minor suggestion (not blocking)**: the `ContractQueueFull` + `is_contract_queue_full` predicate establishes a generally useful platform-resilience pattern (typed marker → caller-side gating → amplification suppression). If you want, a one-paragraph note in `contracts.md` would help future agents discover the predicate before re-inventing string-matching. Not required to merge.

### Related work / strategic fit

- PR #4252 (log-spam reduction, also open) is the sibling fix the issue body anticipated. They're complementary and non-conflicting.
- PR #4145 / #4231 (event-loop wedge fix) is correctly cited as related context.
- Follows the existing precedent established by `send_summary_back_on_rejection` (gated on `is_invalid_update_rejection`, see `update.rs:705`), so it's stylistically consistent.

Strategic assessment: moves the project in the right direction. Ship it.

[AI-assisted - Claude]

sanity (commented):
## Skeptical Review: PR #4253

### High Severity

**1. Incomplete suppression on the originator-relay UPDATE path (`drive_client_update_inner`, `op_ctx_task.rs:395-401`).** When the originator's local pre-forward merge (`super::update_contract` at line 315) returns a queue-full error, the `is_missing_contract_parameters()` gate at 367 is correctly disjoint, but execution falls through to:

```
tracing::error!(... phase = "error", "update: failed to apply update locally before forwarding");
return Err(err);
```

A hot contract whose owner client repeatedly hits its own per-contract queue cap will now emit ERROR-level "failed to apply update locally" once per request from the originator path, on top of the DEBUG line emitted inside `log_update_contract_failure`. That reintroduces exactly the log-volume problem the PR claims to break, just on the originator side instead of the relay side. Should gate this `tracing::error!` on `!err.is_contract_queue_full()` (same shape as the relay gate).

### Medium Severity

**2. Predicate exists at `OpError` level but only UPDATE consumes it — PUT / GET paths still amplify on queue-full.** `send_queue_full_response` returns the typed marker for `PutQuery`, `GetQuery`, `UpdateQuery`, `GetSummaryQuery`, `GetDeltaQuery` — five op kinds. But `is_contract_queue_full()` call sites exist only in `update.rs` / `update/op_ctx_task.rs`. Concrete amplification I found that the PR does not gate:

- `put/op_ctx_task.rs:1217-1223` — PUT relay's `put_contract` failure path logs `tracing::error!("PUT relay: put_contract failed")` unconditionally. A hot contract saturating its queue on PUT relays will now log at ERROR on every PUT, same operator-dashboard noise the PR explicitly cites as the motivation. No code-level amplification (no auto-fetch / ResyncRequest here), but the log-spam half of the bug is unfixed on the PUT path.
- GET path: `GetQuery` queue-full errors flow back into the GET driver's local-cache check. Worth confirming whether the retry logic (`drive_retry_loop`) treats queue-full as a transient-no-retry vs. retry-with-backoff — if it retries, you have amplification onto the same saturated handler.

If queue-full is genuinely "transient backpressure, not operator-actionable," the predicate logic and the log downgrade ought to apply uniformly. The current scope ("UPDATE relay drivers") in the doc comments matches the implementation but not the symptom space.

**3. Sender peer-summary invalidation suppressed on queue-full (`op_ctx_task.rs:1083-1092`) — verify this is deliberate.** The `is_delta && !queue_full` gate also skips `interest_manager.update_peer_summary(&key, &sender_key, None)`. Semantically, queue-full says nothing about whether the sender is stale, so leaving the cached summary alone is defensible (the merge never ran, we have no evidence to invalidate on). But this is a behavior change that isn't called out in the PR description — would benefit from an explicit "deliberately not invalidated; see #4251" comment at the call site, since the next reader will wonder.

**4. Pin test `broadcast_to_suppresses_amplification_on_queue_full` is structurally brittle.** It greps `driver_src` for the literal string `is_contract_queue_full()`. A future refactor that extracts the gate into a helper (`fn should_suppress_amplification(err: &OpError) -> bool { err.is_contract_queue_full() }`) and calls `should_suppress_amplification(&err)` would silently pass the pin while the gate moves out of the driver. Worth either (a) pinning on the *predicate function* presence in the gating-decision module rather than the call-site string, or (b) adding a runtime test that actually drives the relay with a constructed queue-full error and asserts no `NodeEvent::SendInterestMessage` / no `try_auto_fetch_contract` call. The streaming variant's amplification (`op_ctx_task.rs:1737-1738`) has NO equivalent pin — `broadcast_to_streaming_classifies_failures` checks the helper is invoked but not that the helper's queue-full return-value gate exists.

### What I Verified Safe

- Wire format: `ContractQueueFull` is constructed only by `send_queue_full_response` and never serialized (`grep -rn ContractQueueFull` shows only local sites). Cross-node downcast would always fail but no cross-node path exists.
- Anyhow string lookalike: typed downcast is correct; the `test_contract_queue_full_false_for_anyhow_string_lookalike` test pins it.
- Disjointness from other predicates: covered by `test_contract_queue_full_disjoint_from_other_predicates`.
- Send/Sync: unit struct, trivially safe.
- Borrow ordering: `let queue_full = err.is_contract_queue_full()` is side-effect-free and computed before any borrow conflict.

[AI-assisted - Claude]

sanity (commented):
## Code-First Review

Read the diff before the description. The fix is well-formed for UPDATE; main flag is an incomplete migration vs. what the description implies.

### What the code does (independent reading)

- Adds `ContractQueueFull` unit struct (typed marker) with `Display`/`Error` impls; `send_queue_full_response` now wraps it via `ExecutorError::other(ContractQueueFull)` instead of an opaque anyhow string.
- `ExecutorError::is_contract_queue_full()` and mirroring `OpError::is_contract_queue_full()` predicate via `downcast_ref::<ContractQueueFull>`.
- `drive_relay_broadcast_to`: computes `queue_full` once, gates both delta→ResyncRequest and full-state→auto-fetch on `!queue_full`. Adds a third `else if queue_full` DEBUG branch with `event=queue_full_amplification_suppressed`.
- `log_update_contract_failure`: queue-full → DEBUG (was ERROR).
- `log_broadcast_to_streaming_failure`: queue-full → DEBUG (was WARN), AND return value now `!is_contract_exec_rejection() && !is_contract_queue_full()` — the streaming relay caller (`drive_relay_broadcast_to_streaming:1754`) reads this return and skips `try_auto_fetch_contract` correctly.
- New tests pin the typed-classification invariant (incl. anyhow string-lookalike rejection — guards against accidental re-introduction via string-typed errors), DEBUG-vs-ERROR log severity, and structural source-grep pins on the driver.

### Discrepancies vs. description

1. **Partial migration on the PUT relay.** PR title is "stop queue-full saturation from amplifying network-wide" and the typed marker is added as a platform-wide primitive, yet `relay_put_store_locally` (put/op_ctx_task.rs:~1216) still does `tracing::error!("PUT relay: put_contract failed")` for ALL `put_contract` errors. A PUT relay that hits the per-contract queue saturation will still log at ERROR with no `event=queue_full` tag — exactly the log-storm pattern UPDATE just stopped doing. The description doesn't claim PUT is in scope, but it also doesn't carve PUT out (only WASM tuning, queue size, sibling WARN log volume, SyncStateToPeer rate-limit, and CI toolchain are listed in "Out of scope"). The uncommitted working tree in this worktree contains exactly such a PUT fix, suggesting this gap was noticed but not committed before HEAD. Either fold PUT into this PR or explicitly list it as out-of-scope.

2. **Init-queue-full path not migrated.** `executor/runtime.rs:1041 InitCheckResult::QueueFull` still returns a string-typed `StdContractError::Update { cause: "contract initialization queue is full..." }`. That error path is narrow (only during contract initialization) but reads identically to operators and won't satisfy `is_contract_queue_full()`. The new `test_contract_queue_full_false_for_anyhow_string_lookalike` test documents the typed-vs-string boundary; the init-queue path sits on the wrong side of it. Worth either migrating to the typed marker or a comment explaining why init-queue saturation can't amplify.

### Concerns

3. The `else if queue_full` branch at op_ctx_task.rs:1118 is dead for the `is_delta && queue_full` case only if Rust's exclusive `else if` chain order matches the comment intent — confirmed by reading: the first arm requires `is_delta && !queue_full`, second requires `!is_delta && !is_contract_exec_rejection && !queue_full`, third matches both `is_delta && queue_full` and the residual full-state-queue-full case. OK, but the structure is dense; a single `if !queue_full { existing two-branch logic } else { debug log }` would read more cleanly. Code-simplifier commit was supposed to address this; the residual two-level `!queue_full` predicate on each arm survived.

4. The disjointness test (`test_contract_queue_full_disjoint_from_other_predicates`) is asserting a load-bearing property — useful — but `is_request()` returning false for `ExecutorError::other(ContractQueueFull)` happens because `Either::Right` wraps it, not because of any classification of `ContractQueueFull` itself. If someone later moves `ContractQueueFull` into the `Either::Left(RequestError)` variant for any reason, the test passes silently while semantics change. Minor.

### Documentation

5. `is_contract_queue_full` docstring says "Callers MUST gate amplification side effects" — accurate for UPDATE, but no caller in PUT/GET/SUBSCRIBE actually does. Either the docstring should scope to UPDATE (today's reality) or PUT should be migrated.

6. `log_broadcast_to_streaming_failure` docstring still says "Auto-fetch uses `is_contract_exec_rejection` (broad)" — true but incomplete: the return is now ALSO gated on `!is_contract_queue_full()`. The block comment further down explains it correctly; the upper docstring drifts.

### Dead code / shimming

None found. The new marker is wired through; no left-behind anyhow string constructors in send_queue_full_response.

[AI-assisted - Claude]

sanity:
## Review responses

All four Claude reviewer perspectives ran in parallel plus a Codex (external, non-Claude) pass. Findings synthesized and addressed in `65bb0e97`.

### code-first ([review](https://github.com/freenet/freenet-core/pull/4253#pullrequestreview-3681636164))

1. **PUT relay still ERROR-logs queue-full** (`put/op_ctx_task.rs:1217-1223`, `put.rs:138`). Fixed — both sites now branch on `is_contract_queue_full()` and log at DEBUG. PUT has no amplification path so this is the log-spam half of the bug only.
2. **Init-queue path was string-typed** (`executor/runtime.rs:1041`, `InitCheckResult::QueueFull`). Fixed — wraps `ExecutorError::other(ContractQueueFull)` so the typed predicate fires for both queue populations (per-contract fair queue at runtime AND per-contract init queue during initialization).
3. **`if/else if/else if` chain reads denser than needed**. Acknowledged — cosmetic, leaving as-is. The `queue_full` local is read in three places; collapsing the chain would either duplicate the call or hide the gate inside one of the branches. Both feel worse than the current shape.
4. **`!is_request()` in `test_contract_queue_full_disjoint_from_other_predicates` holds incidentally** because the marker lives in `Either::Right`. Noted. The disjointness against the four behavioral predicates (`is_contract_exec_rejection`, `is_invalid_update_rejection`, `is_missing_contract_parameters`, `is_missing_delegate`) is what's load-bearing for the gating in `op_ctx_task.rs`; `is_request` / `is_fatal` are belt-and-braces.
5. **`is_contract_queue_full` docstring overstated what callers MUST do**. Fixed — distinguishes UPDATE callers (must gate amplification + log severity) from PUT/GET/SUBSCRIBE callers (log severity only).
6. **`log_broadcast_to_streaming_failure` docstring didn't mention queue-full in the auto-fetch gate**. Fixed.

### testing ([review](https://github.com/freenet/freenet-core/pull/4253#pullrequestreview-3681554541))

- **Pin test is weaker than a behavior-observable test**. True. A test that wires a stubbed `update_contract` returning `ContractQueueFull` into `drive_relay_broadcast_to` and asserts `notify_node_event(NodeEvent::SendInterestMessage{ResyncRequest})` and `try_auto_fetch_contract` are NOT called would be the gold standard. The relay driver's setup is non-trivial (it owns `OpCtx`, the round-robin queue's full machinery, the interest manager, the broadcast queue) so the mock surface is large. Holding for a follow-up rather than expanding scope here. The three other tests (channel-boundary marker, executor disjointness, log-severity helper) pin the contract well in combination.
- **`update_peer_summary(&key, &sender_key, None)` is suppressed on queue-full alongside the `ResyncRequest`**. Confirmed deliberate — invalidating the sender's cached summary because the merge never ran would be wrong (we have no information about whether their state diverged from ours). Added a note to the in-code comment.
- **`send_summary_back_on_rejection`** — confirmed unaffected, gated on `is_invalid_update_rejection` which is disjoint from queue-full per the new disjointness test.

### skeptical ([review](https://github.com/freenet/freenet-core/pull/4253#pullrequestreview-3681613472))

1. **Originator-path ERROR log spam** (`op_ctx_task.rs:395-401`). Fixed in `65bb0e97`.
2. **PUT/GET relays still amplify** (or at least log-spam). Re-audited: PUT now gated (commit `65bb0e97`); GET's executor-error path already logs at WARN (not ERROR) and doesn't trigger amplification (per the testing-reviewer audit), so no change needed there.
3. **`update_peer_summary` invalidation side effect suppression** — see testing response above.
4. **Pin test brittleness** — accepted limitation; behavioral test deferred to follow-up.
5-7. (Wire format, anyhow lookalike, Send/Sync, borrow ordering) — all verified safe.

### big-picture ([review](https://github.com/freenet/freenet-core/pull/4253#pullrequestreview-3681649569)): APPROVE

- One non-blocking suggestion to document the typed-marker → gating pattern in `.claude/rules/contracts.md`. Will land in a follow-up after this merges so the rule references the merged code shape.

### Codex (external)

> The changes consistently propagate a typed queue-full marker and use it to suppress the intended amplification paths. I did not identify a discrete introduced correctness issue in the diff.

### Net change

- 3 new gated ERROR→DEBUG sites (UPDATE originator, PUT relay, `put_contract` helper).
- 1 new typed-marker site (`InitCheckResult::QueueFull` in `runtime.rs`).
- 2 docstring clarifications.
- 0 new tests in this commit (the existing predicate tests cover the typed marker; the new gating call sites use the same predicate that's already pinned).

Local: `cargo test -p freenet --lib` (focused set) 45/45 ✅, `cargo clippy -- -D warnings -A clippy::wildcard_enum_match_arm` (matches CI's effective surface; 1.94+ wildcard lint is local-toolchain-only) clean.

[AI-assisted - Claude]

PR #4260, opened 2026-05-26
feat(governance): contract hardening foundations — meter + LRU + MAD primitive

sanity:
## Problem

Tonight's `4PjqN5…` incident (#4251, fixed reactively in #4253, #4231, #4252, #4212) made the gap visible: there is no per-contract resource attribution, no cost/benefit governance, and the load-bearing assumption that "any subscriber grants the contract immortality in `HostingCache`" is exactly what an attacker exploits. Each of the patches landed tonight closes a specific blow-up vector; collectively they don't add a layer that bounds *future* blow-up vectors of the same shape.

## Approach

Per `docs/design/contract-hardening.md` — a single design doc landed in commit 1 of this PR after multiple rounds of iteration (cost/benefit framing, MAD-based threshold derivation, shared module spanning peer + contract pipelines, subscriber-obligation handling, observability strategy, honest audit of remaining arbitrary parameters).

This PR is the **foundational layer** — observability, primitives, no behavior change beyond an LRU eviction-ordering refinement. PR #2 will land the consumers (reaper, dashboards, wire-format `Cancelled` variant, pin/ban CLI, repeat-offender TTL).

### What's included

1. **Design doc** (`docs/design/contract-hardening.md`, 400+ lines) — the architectural rationale before code touches anything. Covers MAD-based outlier detection, log-space cost-per-benefit, shared peer/contract governance module, idle-persistence preserved, ban TTL repeat-offender semantics, dashboard observability with cascade-latency metric as the forcing function for explicit-cascade follow-up.

2. **`HostingCache` "recently abandoned" priority bucket.** When a contract transitions from "in use" to "no longer in use" (last subscriber / client / downstream peer removed or expired), mark it via `record_abandonment` and move to the LRU front so disk-pressure eviction targets it before older-but-still-active entries. **Crucially**: this changes eviction *order*, not *eligibility*. Idle network persistence is preserved — a contract with no pressure stays cached just as before.

3. **`AttributionSource::Contract(ContractInstanceId)` and four new `ResourceType` variants** (`ExecCpuMicros`, `ExecFuelUnits`, `StateBytesWritten`, `BroadcastFanoutCost`) wired into the existing `Meter`. The `#[allow(dead_code)] // fixme: use this` annotations on the meter API (which had been dead since the topology module was first written) are dropped now that there's a real production caller.

4. **First production caller: `StateBytesWritten`** reported from `Executor::commit_state_update` via a new `Ring::report_contract_resource_usage` helper. Validates the data-flow end-to-end. CPU/fuel and broadcast-fanout reporters are the obvious next two — they need WASM-engine trait signature changes and a `HostingManager::downstream_subscriber_count` accessor respectively, so landing as follow-up commits rather than in this PR to keep each reviewable.

5. **`crate::governance` module — `detect_outliers` MAD primitive.** Pure function, no plumbing. Computes a robust threshold using trimmed median absolute deviation in log space. The shared substrate for both per-contract governance (PR #2) and peer-side load-shedding (also follow-up). Tests cover empty/insufficient-samples skip behavior, MAD-collapse handling, capacity-ceiling clamp, the worked example from the design doc (median=-0.9, MAD=0.3, threshold=+0.6, abuser flagged), the σ-vs-MAD distinguishing case (a +100 outlier in a 35-sample honest set that standard deviation would miss but MAD catches), and a translation-invariance property test.

### Phase 0 — WASM fuel coverage audit

Independent investigation as part of this work confirmed: contract fuel-gating is correctly configured and tested for all five entry points (`validate_state`, `update_state`, `summarize_state`, `get_state_delta`, plus the timeout case). Delegate fuel-gating works but lacks regression tests, and host-function work is outside the fuel budget by wasmtime design. No production code changes needed; will file a follow-up issue for the delegate-fuel test gap.

### What's deliberately deferred

- WASM-call CPU/fuel reporting — needs trait signature change at the engine boundary (4 call sites).
- BroadcastFanoutCost reporting — needs a `downstream_subscriber_count` accessor on `HostingManager`.
- Per-(peer, contract) UPDATE rate limit (Phase 2) — design decision pending on rate-limit shape (token bucket vs. minimum-interval) before implementing.
- All governance behavior (reaper, dry-run mode, `contract_in_use` exemption removal).
- Dashboard upgrade (ring as centerpiece with peers + contracts + governance state).
- `SubscribeMsg::Cancelled` wire variant.
- Operator pin/ban CLI.
- Repeat-offender ban TTL.

## Testing

- **2642 lib tests pass**, 0 failures, 14 ignored (pre-existing).
- 26 hosting cache tests (6 new for abandonment) — all pass.
- 87 hosting manager tests — all pass.
- 105 topology tests — all pass with new variants.
- 10 new governance primitive tests — all pass, including the worked example from the design doc and the σ-vs-MAD distinguishing property.
- `cargo clippy -- -D warnings` clean.
- `cargo fmt` clean.

## Design dashboard prototype

A working dashboard prototype with the proposed ring-as-centerpiece visualization (peers + contracts placed by ring location, governance state encoded as color, cost density as background heatmap, animated pulses on flagged subscriber lines) lives locally at `~/code/tmp/contract-hardening-dashboard.html` for design reference. The Rust HTML-building implementation lands in PR #2's Phase 4.5 work.

## Fixes

Foundational work for the contract hardening initiative; doesn't close any specific issue but is the substrate for the per-contract governance follow-up that addresses the underlying causes of #4251.

---

[AI-assisted - Claude]


sanity:
## Multi-model review summary

Full-tier review per `~/.claude/rules/multi-model-review.md`. Two rounds, four perspective reviewers + Codex external pass each round.

### Round 1 (commit 8b9df0ff) — original review

Found 8 issues:

| Severity | Source(s) | Finding | Status |
|---|---|---|---|
| Blocker | testing + code-first + Codex | `StateBytesWritten` reported from only 1/6 commit sites (PUT-new, V2 delegate PUT, V2 delegate UPDATE, re-PUT, verify_and_store PUT all silently skipped) — governance would run on UPDATE-only data forever | Fixed in 015ba416 (extracted `Ring::commit_state_write` helper + 3 source-grep pin tests) |
| Blocker | skeptical | Adding `AttributionSource::Contract` to the topology meter causes `extrapolated_usage` to synthesize phantom bandwidth during 5-min ramp-up — every state UPDATE inflates `total_usage` and drives spurious connection removals | Fixed in 32b0ede7 (`AttributionSource::contributes_to` predicate filters at the iteration site + regression test) |
| Blocker | Codex | `detect_outliers` accepts NaN/`±inf` from the extractor — NaN propagates through median/MAD and silently flags nothing | Fixed in be245a41 (`is_finite()` filter at extract step + test) |
| Major | Codex + skeptical | `k=5 ≈ 1-in-a-million` was documented but raw MAD ≈ 0.6745σ, so `median + 5×MAD` is actually only ~3.37σ | Fixed in be245a41 (scale MAD by 1.4826 consistency constant before applying k; design doc worked example updated) |
| Major | code-first | `report_contract_resource_usage` used `tokio::time::Instant::now()` directly — violates DST/TimeSource rule on the executor hot path | Fixed in 515a1682 |
| Major | testing | Missing MAD edge-case tests (n=1, n=2, MadCollapsed-after-trim, multi-outlier with trim contamination, capacity-ceiling negative case) | Fixed in be245a41 (7 new tests) |
| Major | code-first | `HostedContract.abandoned_at` was a write-only field — eviction priority comes from LRU position, the timestamp itself was never read | Fixed in 515a1682 (rustdoc clarifies the field is reserved for the PR #4270 dashboard reader) |
| Major | skeptical + big-picture | `parking_lot::RwLock<BTreeMap>` writer contention on the executor commit path + unbounded growth in `attribution_meters` / `source_creation_times` | Filed as follow-up #4276 (split meter into TopologyMeter + GovernanceMeter, swap to DashMap). Not blocking #4260; should block #4270 reaching Enforce mode. |

### Round 2 (commit 515a1682) — re-review after fixes

Two reviewers (skeptical + Codex) ran against the post-fix HEAD.

| Severity | Source(s) | Finding | Status |
|---|---|---|---|
| Major | skeptical + Codex | `AttributionSource::contributes_to` used `_ => false` wildcard — future `ResourceType` variant added without revisiting would silently fall through (`catch-all _ =>` anti-pattern, same shape as #4009 / #4010) | Fixed in e2fdacd8 (exhaustive enumeration of all 18 source×resource arms) |
| Minor | skeptical | V2 delegate PUT/UPDATE tests claimed to pin `state_size == 3` but the assertions were missing — `cb(&key, 0)` would have passed | Fixed in e2fdacd8 (assertions added to both tests) |
| Minor | skeptical | ~16 remaining `Instant::now()` call sites in ring.rs, pre-existing, will break a future governance-scoring simulation test under `start_paused(true)` | Filed as follow-up #4277. Pre-existing, not introduced by this PR; not blocking. |
| Nit | Codex | `hosting.rs:757` client-subscription exemption is unbounded (clients can stay connected idle); same nit from Round 1 | Pre-existing, called out in the existing docstring. Not introduced by this PR. |
| Nit | skeptical | `worked_example_from_design_doc` test fixture uses different numbers than the doc's narrative worked example | Pre-existing, reconciled in be245a41 (test commented to make clear it's an illustrative fixture, not a verbatim copy of the doc example). |

### Verification

- All 7 blocker/major findings from Round 1 verified fixed by Round 2 reviewers (skeptical "Confirmed-Fixed Findings" + Codex "Verified Fixed").
- Round 2 introduced 1 major + 1 minor + 2 nits; the major and minor are fixed in e2fdacd8; the nits are pre-existing.
- 2 follow-up issues filed: #4276 (meter split) and #4277 (ring.rs TimeSource migration).

### Tests

`cargo test -p freenet --lib`: **2697 passed, 0 failed, 14 pre-existing ignored**. Including:
- 17 governance::tests (10 original + 7 new edge-case tests).
- 3 state_write_attribution_pin_tests (source-grep guards against the #4009 / #4010 telemetry-rot pattern).
- 1 contract_source_does_not_pollute_bandwidth_extrapolation (regression test for the topology pollution finding).

`cargo clippy --lib --tests -- -D warnings`: clean.
`cargo fmt`: clean.

### Risk tier per `~/.claude/rules/multi-model-review.md`

**Full** — touches the executor commit hot path, the topology meter, and adds a new statistical primitive that will eventually drive contract evictions. The review effort matches.

[AI-assisted - Claude]

PR #4262, opened 2026-05-26
fix(server): allow LAN/CIDR browsers to respond to permission prompts

### LINKED ISSUE #4261: bug: Non-localhost browser user cannot respond to delegate prompts in web interface
When trying to use Freemail remotely from a non-localhost browser to send a message, a delegate prompt is triggered and displayed in the shell modal of the Freemail browser instance. Any attempt to try to respond to that prompt gets a 403 error. Examining the Freenet delegate response code, it is currently set to only permit localhost addresses.

It is possible to get around this restriction by using a SSH tunnel, but that is not a convenient solution. Ideally, changes to allow certain remote access would be a better solution, but I appreciate the potential security implications.

sanity:
## Problem

Non-localhost browsers cannot respond to delegate permission prompts.

When a Freenet node is reached from anything other than `localhost` —
LAN, Tailnet, an `--allowed-source-cidrs` range — the permission-prompt
endpoints in `crates/core/src/server/client_api/permission_prompts.rs`
reject the request with `403 Forbidden`. The button click in the
shell-page overlay (or the standalone `/permission/{nonce}` page) sends
`Origin: http://<lan-ip>:7509`, and the old `is_trusted_origin` helper
only accepted `http://localhost / 127.0.0.1 / [::1]`. The pending-list
poll quietly returned `[]`, the SSE stream closed, and the response POST
got `403`. Freemail and any delegate-driven app are effectively unusable
outside SSH tunnels.

## Approach

The WebSocket upgrade gate (`crate::client_events::websocket::connection_info`)
already implements the right policy via `AllowedHosts` +
`AllowedSourceCidrs`: default-LAN private IPs are accepted with a same-IP
Origin, and operators can opt into CGNAT/Tailnet via
`--allowed-source-cidrs`. Reuse those knobs here so the HTTP permission
endpoints agree with the WS-upgrade gate rather than introduce a separate
allowlist.

`is_origin_trusted` now accepts when **any** of:

1. **Loopback Origin** (http or https) — default-on.
2. **Host header IP** in the default-LAN set or the operator's
   `allowed-source-cidrs` **AND** Origin is a same-IP literal. Delegates
   to the existing `host_header_ip_in_cidrs` from `websocket.rs`, which
   already handles the IPv4-mapped-IPv6 / null-origin / CSWSH cases.
3. **Host header in `allowed_hosts`** (operator-added hostname) **AND**
   the Origin's authority equals the Host header. The Origin-vs-Host
   equality requirement is stricter than the WS layer's
   `is_allowed_host` branch and closes the CSWSH path where
   `evil.com` JS opens a fetch to `http://victim-hostname:7509` —
   browsers send `Host: victim-hostname:7509`, `Origin: https://evil.com`,
   which would otherwise pass an Origin-less `allowed_hosts` check.

Extensions are optional on the handlers (`Option<Extension<…>>`) so
`run_local_node` and any older test paths that don't inject them
gracefully degrade to loopback-only — preserving today's behaviour for
those callers. `is_trusted_origin` is removed (it's now dead code; the
loopback branch uses `is_localhost_origin` from `websocket.rs` which
also covers `https://`).

## Testing

Added 11 regression tests covering:

- LAN same-IP Origin → 200 on `/respond`, list on `/pending`
- CIDR same-IP Origin (Tailnet) → 200
- LAN Host + cross-origin Origin (`https://evil.com`) → 403 / empty list
- `allowed_hosts` hostname + matching Origin → 200
- `allowed_hosts` hostname + cross-origin Origin → 403 (CSWSH guard)
- Public-IP Host with no CIDR → 403
- IPv6 link-local LAN Host + matching Origin → trusted
- Missing Origin on `/respond` → still 403
- No-policy fallback is loopback-only
- HTTPS loopback now also accepted
- `origin_authority` parser unit cases

All 2629 lib tests pass; `cargo fmt` and `cargo clippy --all-targets
-- -D warnings` are clean (default-features were not exercised locally;
they will run in CI).

Three helpers in `crates/core/src/client_events/websocket.rs`
(`is_localhost_origin`, `is_allowed_host`, `host_header_ip_in_cidrs`)
become `pub(crate)` for reuse. The WebSocket upgrade gate's behaviour
is unchanged.

## Fixes

Closes #4261

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -135,7 +135,11 @@ use crate::server::client_api::OriginContractMap;
 ///
 /// Delimiter-aware: rejects origins like `http://localhost.evil.com` by requiring
 /// the hostname to be followed by `:`, `/`, or end-of-string.
-fn is_localhost_origin(origin: &str) -> bool {
+///
+/// Also reused by the permission-prompt endpoints
+/// (`crates/core/src/server/client_api/permission_prompts.rs`) so the HTTP and
+/// WS layers agree on what counts as a loopback Origin.
+pub(crate) fn is_localhost_origin(origin: &str) -> bool {
     let prefixes = [
         "http://localhost:",
         "http://localhost/",
@@ -162,7 +166,13 @@ fn is_localhost_origin(origin: &str) -> bool {
 }
 
 /// Returns `true` if the request's `Host` header is in the allowed set.
-fn is_allowed_host(headers: &axum::http::HeaderMap, allowed_hosts: &HashSet<String>) -> bool {
+///
+/// Reused by the permission-prompt endpoints (see [`crate::server::client_api`])
+/// so HTTP and WS layers share the same Host allowlist check.
+pub(crate) fn is_allowed_host(
+    headers: &axum::http::HeaderMap,
+    allowed_hosts: &HashSet<String>,
+) -> bool {
     let Some(host_header) = headers
         .get(axum::http::header::HOST)
         .and_then(|h| h.to_str().ok())
@@ -213,7 +223,7 @@ fn is_allowed_host(headers: &axum::http::HeaderMap, allowed_hosts: &HashSet<Stri
 /// parses to a hostname — not an IP, not matching the Host IP — so it's
 /// refused. DNS rebinding is defeated because no DNS resolution happens
 /// at this layer; only string-parseable IP literals are compared.
-fn host_header_ip_in_cidrs(
+pub(crate) fn host_header_ip_in_cidrs(
     headers: &axum::http::HeaderMap,
     origin_str: &str,
     allowed_cidrs: &crate::server::AllowedSourceCidrs,
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1368,7 +1368,10 @@ pub struct WebsocketApiArgs {
     )]
     pub token_cleanup_interval_seconds: Option<u64>,
 
-    /// Additional hostname(s) to accept in the Host header for WebSocket connections.
+    /// Additional hostname(s) to accept in the Host header for the local
+    /// HTTP/WebSocket API (including the delegate permission-prompt
+    /// endpoints `/permission/pending`, `/permission/events`, and
+    /// `/permission/{nonce}/respond`).
     /// Use when accessing the node via a custom domain (e.g., through a reverse proxy).
     /// Can be specified multiple times. If omitted, only the machine's hostname and
     /// bound IP are accepted.
--- a/crates/core/src/server/client_api/permission_prompts.rs
+++ b/crates/core/src/server/client_api/permission_prompts.rs
@@ -88,9 +88,13 @@ use serde::Deserialize;
 use tokio_stream::wrappers::BroadcastStream;
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 
+use crate::client_events::websocket::{
+    host_header_ip_in_cidrs, is_allowed_host, is_localhost_origin,
+};
 use crate::contract::user_input::{
     CallerIdentity, PendingPrompts, PromptEvent, PromptSnapshot, emit_prompt_event, prompt_events,
 };
+use crate::server::{AllowedHosts, AllowedSourceCidrs};
 
 /// Register permission prompt routes.
 pub(super) fn routes() -> Router {
@@ -226,13 +230,23 @@ fn caller_to_json(caller: &CallerIdentity) -> serde_json::Value {
 /// credentials (cookies, auth tokens) are associated with the poll.
 async fn pending_prompts(
     headers: HeaderMap,
+    allowed_hosts: Option<Extension<AllowedHosts>>,
+    allowed_source_cidrs: Option<Extension<AllowedSourceCidrs>>,
     Extension(pending): Extension<PendingPrompts>,
 ) -> impl IntoResponse {
+    let allowed_hosts = allowed_hosts.as_ref().map(|Extension(v)| v);
+    let allowed_source_cidrs = allowed_source_cidrs.as_ref().map(|Extension(v)| v);
+
     // Missing Origin (e.g. same-origin top-level fetch from some
-    // browsers) is treated as trusted: the gateway only listens on
-    // loopback, and the poll payload is not a capability.
+    // browsers) is treated as trusted: the source-IP filter in
+    // `server.rs::private_network_filter` and the operator-configured
+    // `allowed-source-cidrs` policy already gate which networks can
+    // reach this endpoint, and the poll payload is not a capability.
     let trusted = match headers.get("origin") {
-        Some(value) => value.to_str().map(is_trusted_origin).unwrap_or(false),
+        Some(value) => value
+            .to_str()
+            .map(|s| is_origin_trusted(&headers, s, allowed_hosts, allowed_source_cidrs))
+            .unwrap_or(false),
         None => true,
     };
 
@@ -477,23 +491,29 @@ struct PermissionResponse {
 async fn permission_respond(
     Path(nonce): Path<String>,
     headers: HeaderMap,
+    allowed_hosts: Option<Extension<AllowedHosts>>,
+    allowed_source_cidrs: Option<Extension<AllowedSourceCidrs>>,
     Extension(pending): Extension<PendingPrompts>,
     Json(body): Json<PermissionResponse>,
 ) -> impl IntoResponse {
-    // Validate Origin header to prevent CSRF
-    if let Some(origin) = headers.get("origin") {
-        let origin = origin.to_str().unwrap_or("");
-        if !is_trusted_origin(origin) {
-            return (
-                axum::http::StatusCode::FORBIDDEN,
-                Json(serde_json::json!({"error": "forbidden"})),
-            );
-        }
-    } else {
-        // Reject requests with no Origin header
+    let allowed_hosts = allowed_hosts.as_ref().map(|Extension(v)| v);
+    let allowed_source_cidrs = allowed_source_cidrs.as_ref().map(|Extension(v)| v);
+
+    // CSRF guard: state-changing POST requires Origin and the Origin must be
+    // authorized. Authorized means loopback, in `allowed-source-cidrs`, or a
+    // matched `allowed-host` entry whose Origin authority matches the Host
+    // header (so a cross-site page reaching the gateway via DNS rebinding
+    // cannot forge `Origin: https://evil.com` with a victim Host).
+    let Some(origin) = headers.get("origin").and_then(|v| v.to_str().ok()) else {
         return (
             axum::http::StatusCode::FORBIDDEN,
             Json(serde_json::json!({"error": "missing origin"})),
+        );
+    };
+    if !is_origin_trusted(&headers, origin, allowed_hosts, allowed_source_cidrs) {
+        return (
+            axum::http::StatusCode::FORBIDDEN,
+            Json(serde_json::json!({"error": "forbidden"})),
         );
     }
 
@@ -554,13 +574,17 @@ async fn permission_respond(
 /// missing `Origin` because it returns no useful body to attackers; this
 /// endpoint is stricter because the connection slot itself is a resource
 /// (cap = 64).
-fn is_caller_trusted(headers: &HeaderMap) -> bool {
+fn is_caller_trusted(
+    headers: &HeaderMap,
+    allowed_hosts: Option<&AllowedHosts>,
+    allowed_source_cidrs: Option<&AllowedSourceCidrs>,
+) -> bool {
     if let Some(value) = headers.get("origin") {
         let s = match value.to_str() {
             Ok(s) => s,
             Err(_) => return false,
         };
-        if !is_trusted_origin(s) {
+        if !is_origin_trusted(headers, s, allowed_hosts, allowed_source_cidrs) {
             return false;
         }
     }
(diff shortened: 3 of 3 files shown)


sanity (commented):
## Big Picture Review

### Goal alignment
Yes, the PR closes the user-visible #4261 symptom for the canonical case (LAN/Tailnet browser on a node whose source-IP filter already lets them through). Same-IP Origin on a default-LAN Host now returns 200 on `/respond` and the prompt list on `/pending`. The end-to-end Freemail-from-remote-browser flow that issue #4261 describes is what gets unblocked.

### Consistency with the WebSocket gate
Not quite identical — and the PR description undersells the divergence in one direction. `connection_info` accepts three independent branches: loopback, `is_allowed_host`, or `host_header_ip_in_cidrs`. The new `is_origin_trusted` collapses the middle one (allowed-host) into "Host in `allowed_hosts` **AND** Origin authority equals Host". Per the PR description this is intentional — it closes a CSWSH gap the WS gate has — but that means the HTTP permission-prompt layer is now **stricter** than WS for the same operator-allow-listed hostname. An operator who points a reverse proxy at `mynode.example.com` will get a WS upgrade through with `Origin: null` (sandboxed iframe) but a permission-respond POST with the same Origin will 403. Worth either calling out as a known divergence the WS layer should match next, or filing a follow-up issue so they don't drift further.

### Removed tests / weakened assertions
The deleted `is_trusted_origin` unit tests (loopback variants, https-localhost reject, null-origin reject) are not regression-tests-for-a-shipped-bug; they pinned the implementation of the old helper. Equivalent coverage is now exercised through `is_origin_trusted` (the loopback-only-with-no-policy test, the https-localhost test, the null-origin behavior via `origin_authority`). One semantic flip is buried: `https://localhost` is now accepted whereas before it was rejected — the PR mentions this in passing as "a relic" but it is a real behavior change the description should call out as a list item, not a parenthetical. No regression tests for a real-world bug were dropped.

### Anti-patterns
None of the CI-chasing patterns are present. No `#[ignore]`, no loosened assertions, no commented code, no `unwrap_or_default` on the security path. The 11 new tests are reproducer-shaped (each one names the scenario) and exercise both accept and reject branches. The unused-old-helper removal is justified by the new code path covering it.

### Stale docs / instructions
- `crates/core/src/config.rs:1371` — `allowed_host` help text still says "for WebSocket connections". After this PR it also gates the permission-prompt HTTP endpoints. Update the long-help to say the allow-list also governs delegate permission prompts, or operators will keep filing this issue.
- No `docs/secrets-at-rest.md` / RELEASING.md / agent skill text claims "only localhost is supported" for delegate prompts, so nothing else is mis-stated. I did not find a higher-level operator doc that describes the trust model — would be reasonable as a follow-up, not blocking.
- The PR description does not mention the new CLI-help drift, which is the one user-visible documentation gap an operator would actually trip over.

### Strategic assessment
Solid PR. It fixes the real user impact (not just a CI symptom), the trust-model reuse is the right call over a parallel allowlist, and the CSWSH-hardening on the `allowed-host` branch is a deliberate net-positive. Two small follow-ups before merge: update the `allowed_host` clap doc, and decide whether the WS gate should adopt the same Origin-equals-Host rule for symmetry (or file the divergence as an issue so it isn't lost).

[AI-assisted - Claude]

sanity (commented):
## Testing Review: PR #4262

### Coverage Summary
- Direct changes (handler-level): **good** — 11 new unit tests hit `permission_respond` and `pending_prompts` with realistic LAN / CIDR / allowed-host / CSWSH header combos and an `is_origin_trusted`-only IPv6 link-local check.
- Downstream impact (full Router + Extension chain, `/permission/events` under LAN policy): **partial**. See gaps below.

### Test Gaps

#### Important (should have before merge)

1. **`/permission/events` has zero LAN/CIDR coverage.** Every SSE test calls `permission_events(headers, None, None, Extension(pending))` (see `permission_prompts.rs:1541`). All three new accept-branches on this endpoint are covered only via the `is_origin_trusted` unit tests — the SSE handler's own `is_caller_trusted(&headers, allowed_hosts, allowed_source_cidrs)` glue is never exercised under non-`None` extensions. Add at least: LAN same-IP Origin accepted, cross-origin with LAN Host rejected, allowed-host accepted, allowed-host with cross-Origin rejected. Without this, an extractor-order bug or future refactor that drops the new args could regress silently.

2. **No router-level test that the new `AllowedHosts` / `AllowedSourceCidrs` Extensions actually reach the permission handlers.** `server.rs::test_private_network_filter_*` covers the IP filter middleware, but nothing exercises a full `Router::new().merge(permission_prompts::routes()).layer(Extension(AllowedHosts))...` and hits `/permission/{nonce}/respond` end-to-end. The unit tests pass `Option<Extension<…>>` directly, so an Extension-wiring regression in `server.rs:603/612` (the kind of bug that originally caused #4261) would not be caught. One axum `oneshot` test per endpoint would close this.

3. **`Origin: null` with operator-configured CIDRs is undertested.** The rustdoc on `is_origin_trusted` and `origin_compatible_with_host_ip` (`websocket.rs:296`) explicitly says `null` Origin is accepted when the Host IP is inside an operator-configured CIDR — a deliberate, security-sensitive choice for the dApp sandbox iframe. There is no test exercising `Origin: null` + LAN Host on `/respond` (accept) or on `/respond` with a public Host + no CIDR (reject). The only `null`-Origin test is the no-policy `pending_prompts` empty-list case at `:1238`.

4. **No IPv4-mapped IPv6 (`::ffff:192.168.1.42`) test for any permission endpoint.** Dual-stack sockets are explicitly called out in `server.rs:138-192` and `transport.rs:380`. `host_header_ip_in_cidrs` in `websocket.rs` calls `normalize_mapped_v6` and has its own unit tests there, but the permission-prompts layer never verifies the cross-layer interaction. A LAN browser arriving as `::ffff:192.168.1.42` is the common real-world configuration on a dual-stack Linux gateway — this is the exact #4261 scenario for v6-enabled users.

#### Nice to Have

5. **Test naming.** `test_respond_rejects_cross_origin_with_lan_host` and `test_respond_rejects_allowed_host_with_cross_origin` are easy to swap mentally — both are "host trusted, Origin malicious". Consider `test_respond_rejects_evil_origin_when_host_is_lan_ip` / `..._when_host_is_allowed_host` to disambiguate the trust source.

6. **Missing-Host coverage.** `test_respond_rejects_missing_origin_even_with_lan_host` exists; symmetric `missing-Host + matching-Origin` is not tested.

### CI Considerations
None — all new tests are pure unit tests, no extra wall-clock cost.

[AI-assisted - Claude]

sanity (commented):
## Code-first review

Read the diff before the description. Code matches the stated intent; flagging a few items.

### Behavior change not explicit in the description

The new path accepts `Origin: null` on state-changing `/respond` when the operator has `allowed-source-cidrs` set and the Host IP lands in that CIDR. The old `is_trusted_origin("null")` returned `false` unconditionally. New flow: `is_origin_trusted` → `host_header_ip_in_cidrs("null", …)` → operator-CIDR branch calls `origin_compatible_with_host_ip` which returns `true` for `origin_str == "null"`. That mirrors the WS upgrade policy (deliberate per the design), but it's worth (a) calling out in the PR description's "any of" list and (b) adding a regression test exercising null-Origin POST under operator CIDRs so the behavior is locked in.

### Two Origin-authority parsers

`permission_prompts::origin_authority` is a fresh hand-rolled parser, while `websocket::parse_origin_host` already strips userinfo and handles bracketed IPv6 explicitly. The new helper doesn't strip userinfo and only does a string `split_once("://").split_once('/')`. Comparison is byte-level against the `Host` header so userinfo in `Origin` would simply mismatch (safe direction), but having two slightly different parsers gating related policy invites future drift. Either reuse `parse_origin_host` (or a shared helper) or document why the divergence is intentional.

### Reverse-proxy / non-default-port mismatch

`origin_matches_allowed_host` requires `Origin.authority` to equal the `Host` header byte-for-byte (modulo ASCII case). A node behind an HTTPS reverse proxy where the browser sends `Origin: https://mynode.example.com` (no port) while the upstream Host is `mynode.example.com:7509` will 403. The WS layer's `is_allowed_host` branch is looser (Host-only). This is intentionally stricter per the PR notes, but worth a doc line on `origin_matches_allowed_host` so operators wiring a TLS terminator know to either expose the same port end-to-end or add a port-less entry to `allowed-hosts`.

### Rustdoc claim vs production loopback path

The `is_origin_trusted` rustdoc says branch 3 is disabled "when the `AllowedHosts` / `AllowedSourceCidrs` extensions are absent (e.g. `run_local_node` or older tests)". In production even on a loopback bind, `serve_client_api_in_impl` always injects `Extension(allowed_hosts)` containing the machine hostname. The reassurance applies only to non-`serve_client_api_in_impl` paths; the loopback production path still hits branch 3 (harmlessly — same-localhost requests pass branch 1 first). Worth tightening the rustdoc so a future reader doesn't assume "loopback bind ⇒ branch 3 inactive".

### Tests

The 11 added tests cover the LAN + CIDR + allowed-host happy paths and the CSWSH cross-origin failures well. Missing coverage:

- Null-Origin POST to `/respond` with operator CIDR (locks in the behavior change above).
- Cross-origin POST to `/respond` with `Sec-Fetch-Site: cross-site` and a same-IP Origin (the per-PR change is the Origin gate; the `is_caller_trusted` Sec-Fetch fallback still gates `/events` but isn't exercised against the new policy).

### CI

Most checks still pending at review time; reverify green on the current HEAD SHA before merge.

[AI-assisted - Claude]

sanity (commented):
## Skeptical Review: PR #4262

### Potential Bugs Found

#### Medium Severity

**1. `permission_respond` accepts `Origin: null` on the operator-CIDR branch — CSRF on a state-changing endpoint.**
When `allowed-source-cidrs` is set and a request has `Host: <cidr-ip>:7509`, `host_header_ip_in_cidrs` delegates to `origin_compatible_with_host_ip`, which returns `true` for the literal string `"null"`. The PR description argues this is fine because the WS-upgrade layer already does it, but the WS gate authenticates a channel, whereas `/permission/{nonce}/respond` is a CSRF-sensitive POST that effectively clicks a delegate-permission button on the user's behalf. Any sandboxed-iframe page on the same Tailnet (or any page that gets the user to open a `<iframe sandbox>` to the gateway) can POST `Origin: null` and silently confirm prompts the user never saw. Recommend: tighten `permission_respond` to reject `Origin: null` even when the CIDR branch would otherwise accept (i.e. require a same-IP literal Origin on the response POST, regardless of branch).

**2. `pending_prompts` returns `Access-Control-Allow-Origin: *` on the trusted-response path.**
Line 256: `cors_headers = [("access-control-allow-origin", "*")]` is attached unconditionally. With branch-2/3 trust expanding beyond loopback, any in-CIDR or in-`allowed_hosts` caller now exposes pending prompt content (`nonce`, `delegate_key`, `caller`, `message`, `labels`) to arbitrary cross-origin readers via `fetch`. Even though the trust gate rejects on `is_origin_trusted`, the wildcard CORS plus `is_origin_trusted == true` means the trusted body is readable cross-origin. Recommend: omit the wildcard on the trusted branch, or echo the request Origin only when it equals `host_header`.

**3. `origin_authority` does not strip userinfo; behavior diverges from `parse_origin_host`.**
`parse_origin_host` in `websocket.rs` strips `user@` (line 376-378). `origin_authority` (permission_prompts.rs:687) does not. Result: `Origin: http://victim@evil.com:7509` with `Host: evil.com:7509` would diverge — but luckily the equality check fails closed. Still: two helpers parsing "Origin authority" with different semantics in the same crate is an accident waiting to happen. Recommend: unify on one parser, or document the difference.

**4. PR description's "preserves today's behaviour for `run_local_node`" is wrong.**
`is_origin_trusted` unconditionally calls `host_header_ip_in_cidrs` with an empty CIDR set, which still runs the default-LAN branch. So `run_local_node` bound to e.g. `192.168.1.42` now accepts `Origin: http://192.168.1.42:7509` where before only loopback worked. This is probably desired, but the description claims the opposite — please align the description with reality.

#### Low Severity

**5. `Origin: null` rejection on default-LAN branch is correct but untested.**
`origin_is_same_ip_literal` rejects `null` (via `.parse::<IpAddr>()` failure). No regression test asserts this. Add one.

**6. SSE slot DoS opens up.**
`is_caller_trusted` accepts requests with no `Origin` and no `Sec-Fetch-Site`. Previously gated behind loopback Host policy; now any LAN browser/non-browser can hold one of 64 SSE slots. Consider per-source-IP slot caps.

### Questions for Author

- Why is `Origin: null` acceptable for a state-changing CSRF target? Can branch 2's null-origin permissibility be confined to read-only endpoints?
- Is the `*` CORS header on `/permission/pending` still appropriate now that the trusted response carries real data to a wider set of callers?

### What I Verified (Safe)

- `is_localhost_origin` delimiter-aware: `http://localhost.evil.com`, `http://[::1].evil.com` etc. all rejected.
- `origin_authority("http://x/@evil.com:7509") == "x"` — scheme-path spoof harmless (equality with Host fails).
- IPv4-mapped IPv6 (`::ffff:a.b.c.d`) normalized symmetrically on both Host and Origin sides via `normalize_mapped_v6`.
- `eq_ignore_ascii_case` between `origin_authority` and `host_header` correctly handles case mixing.
- DNS rebinding defeated: no DNS resolution happens; only string-IP literal compare.
- Bare unbracketed IPv6 in Host rejected by `parse_host_header_ip` (fail-closed).
- `is_allowed_host` lowercases input + builder lowercases entries; no case-mismatch bypass.
- Cross-origin POST with `Sec-Fetch-Site: cross-site` rejected by `is_caller_trusted`.

[AI-assisted - Claude]

sanity:
## Review feedback addressed (push c5f9150e)

| # | Reviewer / source | Finding | Resolution |
|---|---|---|---|
| 1 | Codex P1 / skeptical M1 / code-first 1 / testing #3 | `Origin: null` accepted on permission endpoints under operator CIDR (CSRF on `/respond`, prompt disclosure on `/pending`) | `is_origin_trusted` now rejects `Origin: null` unconditionally, before any branch. Regression tests added for both `/respond` and the helper level (with and without CIDR). |
| 2 | skeptical M2 | `Access-Control-Allow-Origin: *` was attached to the trusted-list response on `/pending` | Dropped from the trusted branch; the same-origin shell fetch doesn't need CORS. Wildcard kept only on the empty-list rejection path. Existing CORS test updated to assert the absence on the trusted path. |
| 3 | code-first 2 / skeptical M3 | `origin_authority` did not strip `user@` userinfo (divergence from `parse_origin_host`) | Added userinfo stripping. New regression test. Doc explains why this layer needs `host:port` (full authority) and `parse_origin_host` does not. |
| 4 | code-first 4 / skeptical M4 | Rustdoc claimed default-LAN sub-branch was disabled when `AllowedSourceCidrs` is absent — it isn't | Rustdoc rewritten with a "Pre-check" section for null-Origin + accurate "branch X is active when…" notes. |
| 5 | testing #1 | `/permission/events` LAN/CIDR policy untested | Added 4 SSE tests: LAN same-IP accept, allowed-host accept, cross-origin reject, cross-site Sec-Fetch-Site reject under LAN policy. |
| 6 | testing #2 | No router-level Extension-wiring test (the failure class of #4261 itself) | Added two `tower::ServiceExt::oneshot` tests through the actual `routes()` Router with `Extension(AllowedHosts)`/`Extension(AllowedSourceCidrs)` layered — asserts wiring stays correct end-to-end. |
| 7 | testing #4 | IPv4-mapped IPv6 (`::ffff:192.168.1.42`) — dual-stack realistic shape — uncovered | Added `test_is_origin_trusted_accepts_ipv4_mapped_ipv6_default_lan`. |
| 8 | code-first 5 / skeptical L5 | Cross-site `Sec-Fetch-Site` POST under new LAN policy untested | Added `test_sse_rejects_cross_site_sec_fetch_under_lan_policy`. |
| 9 | big-picture | Stale `--allowed-host` clap help text said "for WebSocket connections" | Updated to also list the delegate permission-prompt HTTP endpoints. |
| 10 | simplifier flag / flaky-tests rule | `test_sse_connection_cap_and_release_on_drop` flaked under parallel execution because every SSE-opening test mutated the global `SSE_CONNECTIONS` counter | Promoted to `#[serial_test::serial(sse_global_counter)]` on every slot-consuming SSE test; reject-path tests stay parallel. |

### Deferred to follow-up issues (out of #4261 scope)

- **WS-vs-HTTP `allowed_hosts` symmetry** (big-picture, skeptical): the HTTP permission layer now requires `Origin authority == Host` on the allow-listed hostname branch; the WS upgrade gate does not. This is intentional CSWSH hardening at the HTTP layer, but the WS gate should adopt the same rule for symmetry. Will file separately.
- **SSE slot-DoS surface widened** (skeptical L6): with LAN browsers now able to open SSE slots, the 64-slot cap is still in place but per-source-IP caps would be safer. Will file separately — not blocking #4261.

All 2640 lib tests + 58 permission_prompts tests pass; `cargo fmt`, `cargo clippy --all-targets -- -D warnings` clean.

[AI-assisted - Claude]

sanity (commented):
## Skeptical Re-review at c5f9150e (third pass)

I went hunting for new bugs introduced by the round-3 fixes and could not find any blocking issues. Summary by question:

1. **`Origin: null` rejection on every accept branch.** `is_origin_trusted` line 648 does `origin.eq_ignore_ascii_case("null")` BEFORE any accept branch. Variants "NULL"/"Null"/"nUlL" all caught. Whitespace forms like "null " or empty "" do not match the literal, but they then fail every downstream branch (no `://` for `parse_origin_host`, no localhost prefix, no Host equality) and fall through to `false`. No bypass found.

2. **CORS wildcard removal from trusted `/pending` body.** The shell polling JS at `path_handlers.rs:1409` is `fetch('/permission/pending')` — same-origin, does not need `Access-Control-Allow-Origin`. The untrusted branch still returns `[]` with `Access-Control-Allow-Origin: *` so devtools doesn't log a CORS error for sandboxed-iframe probes. Safe.

3. **`origin_authority` userinfo stripping.** `rsplit_once('@')` is the correct choice: for `http://a@b@c.com:7509` it returns `c.com:7509` (last `@` wins); for `http://victim:7509@evil.com` it returns `evil.com`. An attacker cannot use userinfo to spoof an allowed-host comparison.

4. **Router-level oneshot test vs production layer stack.** Test uses `super::routes()` + 3 `Extension` layers. Production wraps with `private_network_filter` (only for non-loopback bind, gated separately by IP) and the `pending_prompts` Extension is at the outer `client_api.rs` router rather than inner. Functionally equivalent for the extractor-propagation regression class this test guards. Not a blocker; worth a one-line comment noting that the `private_network_filter` is exercised by the dedicated `middleware_has_extension_injected` test in `server.rs` so it doesn't need to be re-asserted here.

5. **`serial_test::serial(sse_global_counter)` coverage.** I cross-referenced every test that calls `open_sse_with_origin` / `open_sse_with_policy` against whether the call passes a trusted origin (i.e. consumes a slot). Every slot-consuming test has the attribute; reject-path tests correctly omit it. `test_sse_emits_resync_on_lag` uses a local broadcaster and never touches `SSE_CONNECTIONS`. No gaps.

6. **`lan_browser_headers` / `allowed_hosts_with` / `allowed_cidrs_with` helpers.** Construct exactly what production sees: `Host: ip:port`, `Origin: http://ip:port`, hostname both with and without port in the allow-list set, CIDRs via `parse().unwrap()` (test-only `.unwrap()` is fine).

### Minor observations (non-blocking)

- `allowed_hosts_with` `.unwrap()` on `IpCidr::from_str` will panic at test-write time on a typo. Acceptable for tests.
- The doc comment on `pending_prompts` accurately reflects the new CORS posture.

Ready to merge from a correctness standpoint once CI is green on c5f9150e.

[AI-assisted - Claude]

sanity:
## Codex re-review P1 — dismissed (false positive)

Codex's re-review on `c5f9150e` flagged a P1 claiming `open_sse_with_policy` creates a fresh pending-prompts map distinct from the one the test populates, and that the new accepting SSE tests would therefore time out in CI.

This is not the case. \`pending_prompts()\` is a process-global \`OnceLock<Arc<DashMap>>\` singleton (\`crates/core/src/contract/user_input.rs:85-92\`):

\`\`\`rust
static PENDING_PROMPTS: std::sync::OnceLock<PendingPrompts> = std::sync::OnceLock::new();

pub(crate) fn pending_prompts() -> PendingPrompts {
    PENDING_PROMPTS
        .get_or_init(|| Arc::new(DashMap::new()))
        .clone()
}
\`\`\`

Every call returns a clone of the same \`Arc\`. The test populates the global; \`open_sse_with_policy\` reads the global; same map. Confirmed locally:

\`\`\`
test server::client_api::permission_prompts::tests::test_sse_accepts_lan_same_ip_origin ... ok
test server::client_api::permission_prompts::tests::test_sse_accepts_allowed_host_with_matching_origin ... ok
\`\`\`

Both new accepting SSE tests pass deterministically with the \`#[serial_test::serial(sse_global_counter)]\` attribute. The skeptical re-reviewer also independently verified the SSE coverage and reported no gaps. No code change needed.

[AI-assisted - Claude]

PR #4263, opened 2026-05-26
fix(node): downgrade try_send Full WARN to DEBUG (#4238)

### LINKED ISSUE #4238: Log spam: try_release_pending_op_slot WARN fires 30K+/hr under load
## Observed

Production gateways on v0.2.62 (and v0.2.61 before that) emit massive volumes of:

\`\`\`
WARN freenet::node::op_state_manager: try_release_pending_op_slot: notification channel full; driver will wait for OPERATION_TTL timeout
\`\`\`

Rates observed today (2026-05-24, post-HN-traffic-spike):
- nova: 34K/hour current; 56K/hour peak at 18:00 UTC (just before the #4145 wedge)
- vega: 20K/hour current

At ~250 bytes/entry, that's 8–14 MB/hour in the error log. The current-hour `freenet.error.YYYY-MM-DD-HH.log` is at 19 MB on nova after one hour. Over a day, this is ~200 MB just for this one WARN.

## Why it fires

Each warn corresponds to a transient `try_send` Full on the event-loop notification channel inside \`try_release_pending_op_slot_on\` (extracted in #4164). Each occurrence is benign — the parked driver retries via its \`OPERATION_TTL\` timeout. The drop is the correct behavior under back-pressure, but logging every drop at WARN turns a normal back-pressure signal into log spam.

## Proposed fix

Either:
- **Downgrade to DEBUG**: each individual try-send Full is not actionable; the aggregate is. \`debug!\` is correct.
- **Rate-limit at WARN**: emit a summarized WARN every N occurrences (e.g. "try_release_pending_op_slot dropped N notifications in last 60s") so operators still see the signal but at sane volume.

Mirror the same treatment for any sibling \`try_*\` paths that fire frequently under load (e.g. the new \`try_notify_node_event_on\` from #4231 may exhibit the same — verify in logs).

## Not a #4231 regression

This existed pre-#4231 and was actually firing at a HIGHER rate on v0.2.61 (56K/hr) than v0.2.62 (34K/hr in the same time window). #4231 reduced the rate by removing the 30s blocking await that was holding the channel saturated for long stretches. Filing because #4231's post-release monitoring surfaced the existing log-spam issue clearly.

[AI-assisted - Claude]

sanity:
## Problem

Production gateways on v0.2.62 (and v0.2.61 before that) were emitting tens of thousands of WARN entries per hour from two `try_send` helpers in `op_state_manager.rs`:

```
WARN freenet::node::op_state_manager: try_release_pending_op_slot: notification channel full; driver will wait for OPERATION_TTL timeout
```

Observed rates post-HN-spike (2026-05-24):
- nova: 34K/hr current, 56K/hr peak at 18:00 UTC
- vega: 20K/hr current

At ~250 bytes per entry that's 8–14 MB/hour, with `freenet.error.YYYY-MM-DD-HH.log` reaching 19 MB in a single hour. Over a day this is ~200 MB just for these two WARN sites. The flood buried genuinely actionable signals — most notably the rate-limited `release_pending_op_slot: notification channel full for too long` error that fires when the event loop is *actually* stuck.

## Approach

Each `try_send` Full event corresponds to benign back-pressure: the parked driver retries via its OPERATION_TTL fallback, and the 60s sweep reclaims the slot if the notification doesn't land. The drop is the correct behavior — only the per-occurrence WARN is wrong. The aggregate matters; the individual occurrence does not.

Downgrade the per-occurrence log to `debug!` on both helpers (`try_release_pending_op_slot_on` and `try_notify_node_event_on`), keeping the `Closed` arms at `warn!`. Closed-channel means the receiver was torn down (shutdown race or similar), which is genuinely abnormal and should remain a visible signal.

The companion error-level signal (`release_pending_op_slot: notification channel full for too long`) in `release_pending_op_slot_on` is what operators should grep for now — it's rate-limited by design and only fires when back-pressure is sustained.

## Testing

Added four `#[tokio::test]` regression pins in `crates/core/src/node/op_state_manager.rs`:

- `try_release_pending_op_slot_full_does_not_emit_warn` — saturates the channel, calls the helper, asserts no WARN line containing `"notification channel full"` was captured at a WARN-level subscriber.
- `try_release_pending_op_slot_closed_still_emits_warn` — drops the receiver, calls the helper, asserts the `"notification channel closed"` WARN line is present.
- `try_notify_node_event_full_does_not_emit_warn` — same shape for the sibling helper from #4231.
- `try_notify_node_event_closed_still_emits_warn` — inverse for the sibling helper.

All four use `crate::test_utils::TestLogger` with `with_level("warn").capture_logs()`, so the captured buffer only contains entries the WARN-level subscriber actually saw. Pre-fix, the Full assertions would fail because the WARN line would be visible in captured logs.

Local run: all 19 tests in `node::op_state_manager::tests` pass; `cargo clippy --lib --tests -- -D warnings` clean.

## Notes for reviewers

- Not a #4231 regression — the WARN spam existed pre-#4231 and was actually firing at a higher rate on v0.2.61 (56K/hr) than v0.2.62 (34K/hr in the same window). #4231 reduced the rate by removing the 30s blocking await that held the channel saturated longer. The issue was filed because #4231's post-release monitoring made the existing spam visible.
- Kept `Closed` arms at WARN intentionally — closed channel is rare and indicates the receiver was torn down (shutdown race). That signal is worth preserving even with the Full downgrade.
- Updated the `try_notify_node_event_on` rustdoc that previously said "On `Full` or `Closed` ... a warn-level log is emitted" — it now correctly describes the per-arm level split.

Closes #4238

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/mock_runtime.rs
+++ b/crates/core/src/contract/executor/mock_runtime.rs
@@ -260,7 +260,11 @@ where
                 key,
                 new_state: state.clone(),
             }) {
-                tracing::warn!(
+                // Best-effort by design (mirrors the production
+                // executor; see runtime.rs for the #4145 / #4238
+                // rationale). Per-occurrence WARN here would re-
+                // introduce the spam at the caller layer.
+                tracing::debug!(
                     contract = %key,
                     error = %err,
                     "MockRuntime: Failed to broadcast state change (best-effort)"
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -2183,7 +2183,16 @@ where
                     new_state: new_state.clone(),
                 })
             {
-                tracing::warn!(
+                // Best-effort by design (see comment block above and
+                // #4145): a missed broadcast heals via the next UPDATE
+                // or summary-mismatch SyncStateToPeer round. Per-
+                // occurrence WARN here flooded gateways under fan-out
+                // at the same rate as the helper-internal log it
+                // mirrored (#4238). The rate-limited `notify_node_event:
+                // Notification channel full for too long` ERROR in
+                // op_state_manager.rs is the sustained-back-pressure
+                // alert operators should grep for.
+                tracing::debug!(
                     contract = %key,
                     error = %err,
                     "Failed to broadcast state change to network peers (best-effort)"
@@ -2520,7 +2529,11 @@ where
                     new_state,
                 })
             {
-                tracing::warn!(
+                // Best-effort by design — see #4145 and the sibling
+                // commit path above. Per-occurrence WARN here re-
+                // introduced the #4238 spam at the caller layer even
+                // after the helper-internal downgrade.
+                tracing::debug!(
                     contract = %key,
                     error = %err,
                     "Failed to broadcast state change to network peers (best-effort)"
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1307,7 +1307,11 @@ where
                         new_state: state,
                         target: source,
                     }) {
-                        tracing::warn!(
+                        // Best-effort by design (see comment above);
+                        // log at debug to keep the caller layer in
+                        // step with the helper-internal downgrade
+                        // (#4238).
+                        tracing::debug!(
                             contract = %instance_id,
                             error = %e,
                             "Failed to emit SyncStateToPeer for proximity sync (best-effort)"
@@ -1564,7 +1568,10 @@ async fn handle_interest_sync_message(
                     new_state: state,
                     target: source,
                 }) {
-                    tracing::warn!(
+                    // Best-effort by design (see comment above); log
+                    // at debug to keep the caller layer in step with
+                    // the helper-internal downgrade (#4238).
+                    tracing::debug!(
                         contract = %contract,
                         error = %e,
                         "Failed to emit SyncStateToPeer for stale peer correction (best-effort)"
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -1018,7 +1018,14 @@ fn notify_transaction_timeout(
     {
         Ok(()) => true,
         Err(mpsc::error::TrySendError::Full(_)) => {
-            tracing::warn!(
+            // Benign back-pressure on the same event-loop notification
+            // channel as the two `try_*` helpers above: the GC sweep
+            // already removed the tx from the ops maps; this
+            // notification only lets the event loop clean up its
+            // `tx_to_client` map, so dropping it is tolerated. Per-
+            // occurrence WARN here would re-introduce the #4238 spam
+            // class during sustained back-pressure.
+            tracing::debug!(
                 tx = %tx,
                 "Notification channel full, skipping timeout notification for event loop"
             );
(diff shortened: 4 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4263

### Summary
- **PR Title:** fix(node): downgrade try_send Full WARN to DEBUG (#4238)
- **Type:** fix (observability)
- **CI Status:** in progress on `be6b9020`; trivial gates (Fmt, Rule Lint, Conventional Commits, macOS bundle-swap E2E) green
- **Linked Issues:** #4238 (closes)
- **Review tier:** Light
- **Reviewers run:** `freenet:skeptical-reviewer` (two rounds), Codex CLI (three rounds against `origin/main`)

---

### Code-First Analysis

**Independent understanding (final commit):** the change downgrades `tracing::warn!` to `tracing::debug!` on every per-occurrence emission point in the event-loop-notification back-pressure path. Three concentric layers:

1. **Helpers** (`try_release_pending_op_slot_on`, `try_notify_node_event_on`, `notify_transaction_timeout` in `op_state_manager.rs`) — `TrySendError::Full` arms downgraded; `TrySendError::Closed` arms preserved at `warn!`.
2. **Callers of `try_notify_node_event`** (six sites across `contract/executor/runtime.rs` ×2, `contract/executor/mock_runtime.rs`, `operations.rs`, `node.rs` ×2) — `if let Err(...)` arms downgraded from `warn!` to `debug!` since the events themselves are documented best-effort gossip.
3. **Rate-limited diagnostic ERRORs** in `release_pending_op_slot_on` and `notify_node_event_on` (existing) — untouched; remain as the sustained-back-pressure operator signal.

**Stated intent:** Issue #4238 reports ~30 K WARN/hr on nova and ~20 K/hr on vega from `try_release_pending_op_slot: notification channel full` post-HN-spike, with collateral damage to log-grep workflows.

**Alignment:** Matches. The first commit fixed the named helper; the round-1 review caught that the dominant production volume was actually at the caller layer (every `try_notify_node_event` Err re-emitted its own WARN); the round-2 review caught a missed sibling (`notify_transaction_timeout`). Final HEAD covers all per-occurrence WARNs on this notification channel.

**Gaps:** None blocking. Subscriber-notification fan-out WARNs in `runtime.rs:2612 / 2712` are a different channel and were explicitly deferred as out-of-scope by both reviewers.

---

### Testing Assessment

**Coverage Level:** adequate

| Unit | ✅ | Six pin tests in `node::op_state_manager::tests`: `try_release_pending_op_slot_{full_does_not_emit_warn, closed_still_emits_warn}`, `try_notify_node_event_{full_does_not_emit_warn, closed_still_emits_warn}`, `notify_transaction_timeout_{full_does_not_emit_warn, closed_still_emits_warn}`. All use `TestLogger::new().with_level("warn").capture_logs().init()` to confirm a WARN-level subscriber sees nothing on Full and the expected line on Closed. |
| Integration | ⚠️ | Caller-side downgrades (the six runtime/operations/node.rs sites) are not pinned by tests — would require building enough of an `OpManager` + executor to exercise. Mechanical 1-line changes that match the helper-side pattern reviewers approved; trusting helper-level coverage. |
| Simulation | N/A | Pure log-level change; no behavior to verify in simulation. |
| E2E | N/A | Same. |

**Regression Test:** present and valid — verified locally that the new pin tests fail without the downgrades (they assert absence of WARN at WARN-level subscriber, which only holds with the fix).

**Missing Tests:** No caller-side regression test for the six runtime/operations/node.rs downgrades. The risk of someone re-upgrading those WARNs is low (each site has an explicit `#4238` comment now), and pinning them would need a heavyweight executor harness. Acceptable.

---

### Skeptical Findings

**Risk Level:** low

Two rounds of adversarial review caught real bugs in the first commit:

| Helper-level downgrade observability-neutral for fan-out (callers re-emit WARN) | Must Fix (round 1) | runtime.rs:2186/2523, mock_runtime.rs:263, operations.rs:352, node.rs:1310/1567 — verified | RESOLVED in commit `e34ec167` (all six caller WARNs → debug, comments cite #4238) |
| `notify_transaction_timeout` Full arm missed | Should Fix (round 1) | op_state_manager.rs:1017 — verified | First attempt silently didn't land; RESOLVED in commit `be6b9020` after round-2 catch |
| Rustdoc on `try_release_pending_op_slot_on` not updated | Low (round 1) | op_state_manager.rs:871-874 — verified | RESOLVED in commit `e34ec167` |
| Subscriber-notification WARNs (runtime.rs:2612/2712) | Medium, out of scope | Different channel (per-client) | Deferred — recommend follow-up issue if production logs show them spamming |

The round-2 skeptical review on `be6b9020` (this final HEAD) had no remaining Must-Fix or Should-Fix findings; Codex reviewed all three HEADs clean.

---

### Big Picture Assessment

**Goal alignment:** yes — the user-visible symptom ("30 K WARN/hr on nova") is fully addressed by silencing the per-occurrence WARN at every layer of the notification-channel back-pressure path.

**Anti-Patterns Detected:** none. No `.send().await` introduced; no fire-and-forget; no cleanup-exemption regressions. The `try_send` + match-Full/Closed shape is unchanged everywhere.

**Removed Code Concerns:** none — no code removed.

**Scope Assessment:** focused. PR scope grew during review (from "two helpers" to "two helpers + three siblings + six caller sites + two pin tests") but every addition was driven by a concrete reviewer finding tied to the same symptom. Per the `finish-the-fix` rule, bundling these in one PR is the right call rather than punting to follow-ups.

---

### Documentation

- Code docs: complete (rustdoc on `try_release_pending_op_slot_on` and `try_notify_node_event_on` both describe the level split explicitly; inline comments at every downgrade site cite #4238 and explain why DEBUG is correct).
- Architecture docs: n/a — observability-only change.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None.

#### Consider (Suggestions)
- If production logs after this lands still show notable WARN spam on the subscriber-notification channel (`runtime.rs:2612 / 2712`, "Subscriber notification channel full — notification dropped"), file a follow-up applying the same pattern there. That channel is per-client/per-contract so the spam profile is different; the right fix may be different (e.g. the existing per-update counter pattern at `runtime.rs:2241` for the delegate-notification path could be ported).

---

### Verdict

**State:** Ready to Merge
**HEAD SHA reviewed:** `be6b90208fc2cc0554fbf36eea57b5234fa09b0b`

CI is the only remaining gate. The review pass on the current HEAD is clean; all prior-round findings are RESOLVED. Auto-merge once Clippy, Unit & Integration, Simulation, NAT Validation, and Windows Check go green.

[AI-assisted - Claude]

PR #4264, opened 2026-05-26
fix(config): widen --skip-load-from-network to ignore gateways.toml on disk

### LINKED ISSUE #3980: freenet network: --data-dir doesn't isolate gateway bootstrap list (still reads global gateways.toml)
## Summary

`freenet network --data-dir <path>` only isolates per-node state (db, secrets, contracts). The bootstrap gateway list is still read from the global config at:

- macOS: `~/Library/Application Support/The-Freenet-Project-Inc.Freenet/gateways.toml`
- Linux (presumably): `~/.config/freenet/gateways.toml` or similar XDG path

This means a "local test node" started with `--data-dir /tmp/iso/data --skip-load-from-network --gateway 127.0.0.1:31338,<pubkey>` will **also** dial whatever gateways are in the global file (e.g., nova.locut.us, vega.locut.us in default installs), tries NAT traversal to real Freenet peers, and is not actually isolated.

## Reproduction

1. Default-install freenet on a machine where `~/Library/Application Support/.../gateways.toml` lists nova/vega.
2. Run a peer:
   ```bash
   freenet network --network-port 31339 --ws-api-port 7511 \
     --gateway "127.0.0.1:31338,<local-gw-pubkey>" \
     --skip-load-from-network --data-dir /tmp/iso/peer/data \
     --log-dir /tmp/iso/peer/logs --log-level info
   ```
3. Tail the log:
   ```
   Starting initial join procedure with 3 gateways bootstrap_threshold=25
   Attempting connection to gateway gateway=...@100.27.151.80:31337
   Attempting connection to gateway gateway=...@5.9.111.215:31337
   Attempting connection to gateway gateway=...@127.0.0.1:31338
   ```
   Three gateways, two of which are public. Not isolated.

## Workaround

Override `HOME` (or platform equivalent) to a sandbox dir holding an empty gateways file:

```bash
mkdir -p ~/iso-home/Library/Application\ Support/The-Freenet-Project-Inc.Freenet
printf 'gateways = []\n' > ~/iso-home/Library/Application\ Support/The-Freenet-Project-Inc.Freenet/gateways.toml
HOME=~/iso-home freenet network --gateway "127.0.0.1:31338,$PUBKEY" ...
```

Note: empty file (just `""`) trips a TOML parse error `missing field 'gateways'` — needs the explicit `gateways = []`.

## Proposed fix

Add a `--config-dir <path>` flag (or `FREENET_CONFIG_DIR` env var) that overrides the global config dir, parallel to `--data-dir`. Then a fully isolated test node is just:

```bash
freenet network --data-dir /tmp/iso/data --config-dir /tmp/iso/config --gateway 127.0.0.1:31338,$PUBKEY ...
```

Bonus: `--skip-load-from-network` could be widened (or a sibling flag added) to also skip loading the on-disk gateway list, so the only bootstrap gateways are those passed via `--gateway`. Strict additive semantics for `--gateway` (replace, not append) would also work.

## Why it matters

- Local dev / CI test runs accidentally talk to public network.
- Test publishes can leak content to public gateways (depending on contract behavior).
- Telemetry from "isolated" test nodes pollutes prod metrics.
- Surprising for new contributors — `--data-dir` reads as "isolate this node" but only does half the job.

## Context

Hit while running freenet-email two-node E2E tests. Filed from freenet-email session debugging issue #36 follow-up.

sanity:
## Problem

`freenet network --skip-load-from-network --data-dir /tmp/iso --gateway 127.0.0.1:31338,<pubkey>` was meant to launch a fully isolated test node, but a non-gateway peer with this flag set still fell through to the file-load branch and read the global `gateways.toml`:

- Linux: `~/.config/freenet/gateways.toml`
- macOS: `~/Library/Application Support/The-Freenet-Project-Inc.Freenet/gateways.toml`

On a default-installed machine, that file lists the public production gateways (nova, vega). So the "isolated" test node ALSO dialed those gateways, attempted NAT traversal to real peers, and could leak test publishes / telemetry to production.

Repro from the original report (#3980):

```
Starting initial join procedure with 3 gateways bootstrap_threshold=25
Attempting connection to gateway gateway=...@100.27.151.80:31337
Attempting connection to gateway gateway=...@5.9.111.215:31337
Attempting connection to gateway gateway=...@127.0.0.1:31338
```

Three gateways, two of which are public. The user expected the `--skip-load-from-network --gateway X` combination to produce exactly one gateway: `X`.

## Approach

The existing branch for `skip_load_from_network` only handled gateway peers (`&& is_gateway`); non-gateway peers fell into the catch-all file-load branch. Drop the `is_gateway` condition so the isolated path also applies to non-gateway peers — the only gateways used are those from `--gateways` (hidden inline JSON, populates `remotely_loaded_gateways`) and `--gateway` (CLI flag, merged below). The on-disk `gateways.toml` is no longer read.

Preserve the existing safety net: a non-gateway Network-mode peer that provided no bootstrap source at all still fails with "Cannot initialize node without gateways" (now with a clearer message naming `--skip-load-from-network`). The "Cannot initialize node without gateways" wording is unchanged at the prefix so the existing assertion in `test_config_build_network_mode_empty_gateway` keeps passing.

The narrower alternative (a separate flag) was rejected because `--skip-load-from-network` is already documented for "local/test network" use — widening it is more consistent than adding a sibling flag that does almost the same thing. `--config-dir` is sufficient on its own for full isolation today (point it at an empty directory), but `--skip-load-from-network` should not silently disagree with its name.

## Testing

Added regression test `test_skip_load_from_network_ignores_on_disk_gateways_toml` that:

1. Pre-populates `gateways.toml` with a public-looking gateway entry (`203.0.113.99:31337`).
2. Runs `ConfigArgs::build()` with `skip_load_from_network: true` and one CLI `--gateway` entry at `127.0.0.1:31338`.
3. Asserts the final `cfg.gateways` contains exactly one entry — the CLI-provided one — and the public gateway from the file does NOT leak in.
4. Verifies the on-disk file was not rewritten (existing contract).

Without the fix, the test fails: the final list has 2 entries (CLI + leaked file entry).

Also confirmed the related tests still pass:
- `test_config_build_network_mode_gateway_only` (skip_load + --gateway with no on-disk file)
- `test_config_build_network_mode_empty_gateway` (skip_load + empty --gateway → still errors)
- `test_config_build_with_gateway_flag` and `test_config_build_multiple_gateways` (file-load paths unaffected)

39 config tests pass; clippy clean (matching CI's `cargo clippy --locked -- -D warnings`).

## Fixes

Closes #3980

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -477,6 +477,24 @@ impl ConfigArgs {
                 // Inline gateways were provided via --gateways flag, use those
                 remotely_loaded_gateways
             }
+        } else if self.network_api.skip_load_from_network && has_cli_gateways {
+            // #3980: Strict additive --gateway semantics under
+            // skip_load_from_network. When the user explicitly passes
+            // --gateway, treat that as the complete bootstrap set: do NOT
+            // merge in the on-disk gateways.toml cache (which on a default
+            // install lists public production peers like nova/vega). The
+            // explicit --gateway entries are merged below.
+            //
+            // When --gateway is NOT supplied under skip_load_from_network,
+            // the on-disk gateways.toml is still read (next branch). This
+            // preserves the contract used by isolated test harnesses
+            // (e.g. freenet-test-network's Docker NAT setup) that
+            // pre-populate gateways.toml in a custom --config-dir.
+            tracing::info!(
+                "skip_load_from_network with --gateway entries: \
+                 ignoring on-disk gateways.toml; using only CLI-supplied gateways"
+            );
+            Gateways { gateways: vec![] }
         } else {
             // Either skip_load_from_network is set (use local file only), or the
             // remote fetch failed and we need to fall back to the local cache.
@@ -893,7 +911,12 @@ pub struct NetworkArgs {
     #[arg(long)]
     pub is_gateway: bool,
 
-    /// Skips loading gateway configurations from the network and merging it with existing one.
+    /// Skip fetching the remote gateway index. When combined with --gateway,
+    /// the explicit CLI entries also REPLACE (not merge with) any on-disk
+    /// gateways.toml cache — useful for fully isolated test setups. With
+    /// neither --gateway nor --gateways supplied, the on-disk gateways.toml
+    /// is still read (preserved for test harnesses that pre-populate it via
+    /// --config-dir).
     #[arg(long)]
     pub skip_load_from_network: bool,
 
@@ -1251,8 +1274,11 @@ pub struct NetworkApiConfig {
     pub bbr_startup_rate: Option<u64>,
 
     /// When true, this node is part of a local/test network and does not load
-    /// gateways from the public index. Used to disable the relay-ready gate
-    /// and other production-only features.
+    /// gateways from the public remote index. Used to disable the relay-ready
+    /// gate and other production-only features. When combined with --gateway,
+    /// the on-disk gateways.toml is also skipped (the explicit CLI entries
+    /// become the complete bootstrap set); without --gateway, the on-disk
+    /// gateways.toml is still read.
     #[serde(default)]
     pub skip_load_from_network: bool,
 }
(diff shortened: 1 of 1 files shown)


sanity (commented):
## Comprehensive PR Review: #4264

### Summary
- **PR Title:** fix(config): widen --skip-load-from-network to ignore gateways.toml on disk
- **Type:** fix
- **CI Status:** in progress (rebased onto current main, current HEAD `62b27366`)
- **Linked Issues:** Closes #3980
- **Review tier:** Light (single-file config change, ~120 lines including tests, no high-risk surface)
- **Reviewers run:** `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer`, Codex external review. Re-review on the post-fix HEAD: `freenet:skeptical-reviewer` + Codex.

---

### Code-First Analysis
**Independent Understanding:** The PR widens the semantics of `--skip-load-from-network` so a non-gateway peer also skips reading the on-disk `gateways.toml` cache. Previously, only gateway peers (`is_gateway: true`) ran fully isolated under that flag; a non-gateway peer fell through to the file-load branch and dialed any gateways listed in `~/.config/freenet/gateways.toml`, defeating isolation. The change drops the `&& is_gateway` guard, preserves the "Cannot initialize node without gateways" safety net for non-gateway Network-mode peers, and adds a regression test that pre-populates `gateways.toml` with a public-looking entry and asserts it doesn't leak.

**Stated Intent:** Fix #3980 — make `--skip-load-from-network` actually skip ALL on-disk + remote gateway sources so an "isolated" test peer doesn't dial production gateways.

**Alignment:** Matches.

**Gaps:** None.

---

### Testing Assessment
**Coverage Level:** Adequate.

| Unit | ✅ | Two new tests: `test_skip_load_from_network_ignores_on_disk_gateways_toml` (primary regression for #3980); `test_skip_load_from_network_gateway_without_public_port_succeeds` (edge case found during review). 40/40 config tests pass locally. |
| Integration | N/A | The config-load path is unit-testable in isolation; no integration test needed. |
| Simulation | N/A | Pure config logic; no routing/topology behavior. |
| E2E | N/A | |

**Regression Test:** Present and valid. The primary test (`test_skip_load_from_network_ignores_on_disk_gateways_toml`) reproduces the original bug: without the fix it produces 2 gateways (CLI + leaked file entry) instead of 1.

**Missing Tests:** None for the new behavior.

---

### Skeptical Findings (resolved)

| `peer_id.is_none()` was a wrong proxy for "non-gateway peer": a valid isolated gateway started with `--is-gateway --public-network-address X --network-port Y` (no `--public-network-port`) has `peer_id = None` per `peer_id = public_address.zip(public_port)`. The first draft of the PR would have erroneously rejected this configuration. | Medium | `crates/core/src/config.rs:479` (now `:483`) | **Fixed** in commit `62b27366` — guard switched to `!self.network_api.is_gateway && mode == OperationMode::Network`. Added regression test `test_skip_load_from_network_gateway_without_public_port_succeeds` that builds exactly the offending config and asserts success. Verified by re-review. |

A pre-existing parallel inconsistency exists at `config.rs:519` (the file-load branch uses the same `peer_id.is_none()` guard with the same latent issue for first-boot gateways with `--network-port` only). This is NOT introduced by this PR and is filed as follow-up issue #4268.

---

### Big Picture Assessment
**Goal Alignment:** Yes. Issue #3980 has two asks: (a) a `--config-dir` flag and (b) widening `--skip-load-from-network`. (a) was already implemented prior to this PR (`ConfigPathsArgs.config_dir` is clap-exposed via `--config-dir` / `CONFIG_DIR`); this PR closes (b), the residual gap.

**Anti-Patterns Detected:** None. No `#[ignore]`, no loosened assertions, no commented-out code, no TODO deferrals. Both the clap help (`config.rs:910-916`) and the serde-deserialized field doc (`config.rs:1271-1276`) are updated to describe the new semantics.

**Removed Code Concerns:** None. No tests removed or weakened. The only deletions are the `&& is_gateway` condition (the bug being fixed) and four lines of now-redundant computation in the simplified `else` branch.

**Scope Assessment:** Focused. Diff is single-file (`crates/core/src/config.rs`), +153/-15 across the two commits.

**Behavior change worth flagging (acceptable, not a regression):** Switching the guard from `peer_id.is_none()` to `!is_gateway` also changes one narrow case: a non-gateway peer with both `--public-network-address` and `--public-network-port` set, `--skip-load-from-network`, and no `--gateway` / `--gateways` / on-disk file. Old behavior: silent empty gateway list. New behavior: hard error. This is the desired direction (such a peer has no way to bootstrap and the misconfiguration should surface immediately) and explicitly intended by the PR's own comment.

---

### Documentation
- Code docs: Updated. Both the clap `--skip-load-from-network` help (`config.rs:910-916`) and the `NetworkApiConfig.skip_load_from_network` field doc (`config.rs:1271-1276`) describe the new "do not load remote index AND do not read on-disk cache" semantics.
- Architecture docs: N/A — no architectural change.
- User docs: N/A — `--skip-load-from-network` is operator-only, no end-user manual section affected.

---

### Recommendations

#### Must Fix (Blocking)
None. The original blocking finding (`peer_id.is_none()` vs `is_gateway`) has been addressed.

#### Should Fix (Important)
None.

#### Consider (Suggestions)
1. Follow-up to align the file-load branch guard at `config.rs:519` with the new `!is_gateway`-based gate — same latent issue exists there for first-boot isolated gateways when the remote fetch fails. Filed as **#4268** (not blocking this PR).

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `62b273664f8f8ba22bc1cdf3b226fbd0c4f8c998`

Re-review after fix for the prior `peer_id.is_none()` finding — resolved correctly with a corresponding regression test. The rebase pulled in commit `7dcadf1c` from main (unrelated macOS updater fix) which incidentally resolved a false-positive Codex finding from the prior round. All three reviewer perspectives (code-first / skeptical / big-picture) plus Codex external pass converge on Approve at this HEAD.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4264 — Final Pass

### Summary
- **PR Title:** fix(config): widen --skip-load-from-network to ignore gateways.toml on disk
- **Type:** fix
- **Linked Issues:** Closes #3980
- **Review tier:** Full (escalated from Light after the initial widening broke the Docker NAT integration test on the prior HEAD `62b27366`)
- **HEAD reviewed (final):** `06ffb06b`
- **Reviewers run across all passes:** `freenet:skeptical-reviewer` ×3, `freenet:big-picture-reviewer` ×1, Codex external ×3

### Pass history
1. **HEAD `2e32bb97`** (initial widening): Codex flagged `peer_id.is_none()` regression for `--is-gateway --network-port` without `--public-port` (real, blocking).
2. **HEAD `01b60972` → `62b27366`** (after `!is_gateway` fix + rebase): all reviewers approved at that snapshot, but the change was too aggressive — `freenet-test-network = "0.1.23"` (crates.io dep, NOT path) pre-populates `/config/gateways.toml` for Docker NAT peers and relies on `--skip-load-from-network` peers reading it. CI's Docker NAT integration test failed with "Cannot initialize node without gateways".
3. **HEAD `84dfc2ab`** (narrowed approach: strict-additive `--gateway` only): CI green. Re-review found two issues — silent drop of `--gateways` JSON entries when combined with `--gateway` (medium), and docstring inaccuracy for gateway-mode case (Codex).
4. **HEAD `a6b1d453`**: both issues addressed (return `remotely_loaded_gateways` instead of empty; docstrings rewritten to describe all three layered cases). Re-review approved with two Low/NIT findings.
5. **HEAD `06ffb06b`** (this pass): Both Low/NITs addressed: clap docstring wording precision + pinned the previously-uncovered four-way combination (`skip_load + is_gateway + --gateway + on-disk file`). Light verification pass — diff is purely additive (one new test + 7 lines of docstring rewording, zero logic changes).

### Final scope
Single file: `crates/core/src/config.rs`. Five regression tests pin the full matrix of `--skip-load-from-network` × (`--is-gateway`/`--gateway`/`--gateways`/on-disk file):

| Test | Scenario | Behavior pinned |
|------|---------|-----------------|
| `test_skip_load_from_network_with_cli_gateway_ignores_on_disk_file` | non-gateway + --gateway + file | file NOT leaked (fixes #3980) |
| `test_skip_load_from_network_without_cli_gateways_reads_on_disk_file` | non-gateway + no --gateway + file | file IS read (Docker NAT contract) |
| `test_skip_load_from_network_gateway_mode_ignores_on_disk_file` | is_gateway + no --gateway + file | file NOT read (pre-existing isolated-gateway behavior) |
| `test_skip_load_from_network_preserves_inline_gateways_with_cli_gateway` | --gateways JSON + --gateway + file | both supplied entries reach final; file does not |
| `test_skip_load_from_network_isolated_gateway_with_cli_gateway_ignores_file` | is_gateway + --gateway + file | CLI entry reaches final; file does not |

### Recommendations
- **Must Fix:** None.
- **Should Fix:** None.
- **Consider:** Follow-up #4268 already filed for the pre-existing parallel `peer_id.is_none()` guard inconsistency in the file-load branch (unchanged scope, not introduced by this PR).

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed (final):** `06ffb06b`

[AI-assisted - Claude]

PR #4266, opened 2026-05-26
fix(transport): set SO_RCVBUF/SO_SNDBUF on UDP sockets

### LINKED ISSUE #3844: Freenet UDP socket does not set SO_RCVBUF/SO_SNDBUF — stock Linux defaults cause massive packet drops on gateways
## Summary

`impl Socket for UdpSocket` in `crates/core/src/transport.rs:412` creates the UDP socket via `socket2` but never calls `set_recv_buffer_size()` / `set_send_buffer_size()`. The socket inherits the kernel's `net.core.rmem_default` (**208 KiB on stock Linux**, unchanged for ~15 years). On a gateway under realistic load this buffer fills and the kernel drops packets, recorded as `UdpRcvbufErrors` in `/proc/net/snmp`.

## Observed impact

Production gateways `nova.locut.us` and vega, both running 0.2.41, with stock `net.core.rmem_max = 212992`:

**nova** (250-day uptime, continuous gateway operation):
\`\`\`
UdpInDatagrams    56,828,638,173
UdpRcvbufErrors  838,702,972,194   ← ~15× more drops than delivered datagrams
\`\`\`

**vega** (similar):
\`\`\`
UdpInDatagrams    17,854,486,151
UdpRcvbufErrors    2,911,797,346   ← ~16% loss
\`\`\`

NIC-level softnet drops on nova over the same period: **4**. The loss is entirely at the socket receive-buffer layer — exactly the symptom of an undersized `SO_RCVBUF`.

This is **not** a regression in 0.2.41. The 0.2.40 → 0.2.41 commit range contains zero transport or networking changes, and the cumulative drop counters predate 0.2.41 by months. It is a long-standing issue: Freenet has never tuned its UDP socket buffers and has always depended on operators knowing to raise the sysctls.

## Mitigation applied to affected gateways

\`\`\`
# /etc/sysctl.d/99-udp-buffers.conf
net.core.rmem_max = 67108864      # 64 MiB
net.core.rmem_default = 2097152   # 2 MiB
net.core.wmem_max = 67108864
net.core.wmem_default = 2097152
net.core.netdev_max_backlog = 10000
\`\`\`

After \`sysctl --system\` + \`systemctl restart freenet-gateway.service\` on both hosts, \`UdpRcvbufErrors\` stopped climbing.

## Proposed fix

In \`crates/core/src/transport.rs\` at the \`Socket::bind\` implementation (~line 412), after \`sock.set_nonblocking(true)?\`:

\`\`\`rust
// Request a generous UDP socket buffer. Stock Linux net.core.rmem_default is
// only 208 KiB, which is inadequate for a gateway under realistic load and
// causes kernel-level drops visible as UdpRcvbufErrors. The kernel silently
// clamps this to net.core.rmem_max, so we also log what we actually got.
const DESIRED_UDP_BUF_BYTES: usize = 16 * 1024 * 1024; // 16 MiB
if let Err(e) = sock.set_recv_buffer_size(DESIRED_UDP_BUF_BYTES) {
    tracing::warn!(error = %e, \"failed to request UDP recv buffer size\");
}
if let Err(e) = sock.set_send_buffer_size(DESIRED_UDP_BUF_BYTES) {
    tracing::warn!(error = %e, \"failed to request UDP send buffer size\");
}
let actual_rcv = sock.recv_buffer_size().unwrap_or(0);
let actual_snd = sock.send_buffer_size().unwrap_or(0);
if actual_rcv < DESIRED_UDP_BUF_BYTES / 2 {
    tracing::warn!(
        requested = DESIRED_UDP_BUF_BYTES,
        actual_rcv,
        \"UDP recv buffer smaller than requested — raise net.core.rmem_max (currently clamped by kernel)\"
    );
}
tracing::info!(actual_rcv, actual_snd, \"UDP socket buffer sizes\");
\`\`\`

Notes:
- Linux doubles the value internally (kernel bookkeeping), so a request of 16 MiB yields ~32 MiB when uncapped.
- If \`rmem_max\` is too small, the kernel silently clamps and the actual value will be lower. The logged warning tells the operator to raise the sysctl. This gives us graceful degradation plus actionable guidance.
- 16 MiB is conservative enough not to alarm low-memory environments (64 MiB RSS per socket would be aggressive; 32 MiB is reasonable for a process that already holds much more state).
- Alternatively, make the value configurable via \`ConfigArgs\` if we want user control — but having a sensible default is the main point.

## Why CI didn't catch this

There is no test that verifies the UDP socket's actual \`SO_RCVBUF\` size after \`bind()\`. Simulation/in-memory tests use \`InMemorySocket\` which has no kernel buffer. The issue only surfaces under real sustained UDP load on a stock-configured host. A unit test could be added: in the \`tokio::net::UdpSocket\` variant of the test suite, bind the socket and assert \`getsockopt(SO_RCVBUF)\` returns at least N MiB.

## Related

- CLAUDE.md memory note (internal): operators running gateways on nova / vega / technic have been silently accumulating these drops for months.
- None of the 5 commits in 0.2.40 → 0.2.41 touch networking; ruling out this release as a regression source.

[AI-assisted - Claude]

sanity:
## Problem

`Socket::bind` for `UdpSocket` (crates/core/src/transport.rs:411) creates the socket via `socket2` and configures `SO_REUSEADDR`/`SO_NONBLOCK`/dual-stack, but never calls `set_recv_buffer_size` / `set_send_buffer_size`. As a result, every gateway inherits the kernel's stock `net.core.rmem_default` (~208 KiB on Linux, unchanged for ~15 years), which is inadequate for sustained UDP throughput.

Observed impact on production gateways (issue #3844):

- **nova** (250-day uptime): `UdpInDatagrams = 56.8B`, `UdpRcvbufErrors = 838.7B` — ~15x more drops than delivered datagrams.
- **vega**: ~16% drop rate.
- NIC-level softnet drops over the same period: 4. All loss was at the socket receive-buffer layer.

Operators have been silently mitigating this by raising `net.core.rmem_max` via sysctl. Any new gateway deployed on stock Linux suffers the same drops invisibly until someone notices.

## Solution

After `set_nonblocking(true)?` and before `bind`, request 16 MiB `SO_RCVBUF` / `SO_SNDBUF`:

```rust
const DESIRED_UDP_BUF_BYTES: usize = 16 * 1024 * 1024;
if let Err(e) = sock.set_recv_buffer_size(DESIRED_UDP_BUF_BYTES) {
    tracing::warn!(error = %e, "failed to request UDP recv buffer size");
}
// ... same for send ...
let actual_rcv = sock.recv_buffer_size().unwrap_or(0);
let actual_snd = sock.send_buffer_size().unwrap_or(0);
if actual_rcv < DESIRED_UDP_BUF_BYTES / 2 {
    tracing::warn!(/* "raise net.core.rmem_max" */);
}
tracing::info!(actual_rcv, actual_snd, "UDP socket buffer sizes");
```

Notes:

- Linux doubles the requested value internally (kernel bookkeeping), so an unclamped request yields ~2x the configured size. macOS/BSD do not double.
- If `rmem_max` is too small the kernel silently clamps. The post-clamp `actual_rcv < DESIRED/2` check converts the silent clamp into an actionable operator warning (without false-positives on platforms that don't double).
- `socket2` is already a transitive dependency used by this same `bind` impl two lines above (`set_only_v6`), so no new crates.
- Behavior degrades gracefully: failures log a warn and proceed; the socket still binds.

## Testing

New regression test `udp_socket_bind_requests_large_buffers` (in the existing `dual_stack_tests` module, which already has precedent for real-socket tests):

- Binds a **control** socket via raw `tokio::net::UdpSocket::bind` (no buffer hints).
- Binds a **tuned** socket via our `Socket::bind` impl.
- Reads `SO_RCVBUF` / `SO_SNDBUF` on both via `socket2::Socket::from(std)`.
- Weak invariant (always enforced): `tuned >= control`.
- Strict invariant: when `/proc/sys/net/core/rmem_max >= 4 MiB`, the tuned buffer must be **strictly** larger than the control. Without the fix, both sockets get the same kernel default and the strict assertion fails.

Why the conditional: on constrained CI environments where `rmem_max == rmem_default ≈ 208 KiB`, the kernel clamps both sockets to the same value and the fix has no observable effect on the resulting buffer — the test relaxes to the weak invariant. On any properly-tuned host (and on production gateways), the strict assertion catches a regression that drops the `set_recv_buffer_size` call.

Verified locally on nova (`rmem_max = 64 MiB`, `rmem_default = 2 MiB`):

```
test transport::dual_stack_tests::udp_socket_bind_requests_large_buffers ... ok
test result: ok. 591 passed; 0 failed; 6 ignored
```

(All 591 transport unit tests pass; the buffer-tuning is non-disruptive to existing behavior.)

## Why CI didn't catch this

There was no test asserting `SO_RCVBUF` after `bind()`. `InMemorySocket` (used in simulation tests) has no kernel buffer, so the bug only surfaces under real UDP load. The new test closes that gap.

## Out of scope (filed separately)

While running `cargo clippy --all-targets` locally I noticed pre-existing `doc_lazy_continuation` errors in `crates/core/tests/simulation_integration.rs:9421-9422` triggered by Rust 1.94's stricter lints. CI runs plain `cargo clippy --locked` (no `--all-targets`) so these don't currently block CI, but they should be cleaned up — I'll file a separate issue.

Closes #3844

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/transport.rs
+++ b/crates/core/src/transport.rs
@@ -425,6 +425,36 @@ impl Socket for UdpSocket {
             sock.set_only_v6(false)?;
         }
         sock.set_nonblocking(true)?;
+        // Request a generous UDP socket buffer. Stock Linux net.core.rmem_default
+        // is only ~208 KiB, which is inadequate for a gateway under realistic
+        // load and causes kernel-level drops visible as UdpRcvbufErrors in
+        // /proc/net/snmp. Production gateways were silently losing ~15x the
+        // packets they delivered until operators raised the sysctls manually
+        // (see #3844). The kernel silently clamps to net.core.rmem_max, so we
+        // also surface the actual size to give operators actionable diagnostics.
+        const DESIRED_UDP_BUF_BYTES: usize = 16 * 1024 * 1024; // 16 MiB
+        if let Err(e) = sock.set_recv_buffer_size(DESIRED_UDP_BUF_BYTES) {
+            tracing::warn!(error = %e, "failed to request UDP recv buffer size");
+        }
+        if let Err(e) = sock.set_send_buffer_size(DESIRED_UDP_BUF_BYTES) {
+            tracing::warn!(error = %e, "failed to request UDP send buffer size");
+        }
+        let actual_rcv = sock.recv_buffer_size().unwrap_or(0);
+        let actual_snd = sock.send_buffer_size().unwrap_or(0);
+        // Linux doubles the requested value internally (kernel bookkeeping), so an
+        // unclamped 16 MiB request returns ~32 MiB. macOS/BSD do not double, so an
+        // unclamped request returns ~16 MiB. We treat anything below half the
+        // request as a clamp signal — that catches the stock-Linux case (rmem_max
+        // ≈ 208 KiB → ~416 KiB actual) without false-positives on platforms that
+        // don't double.
+        if actual_rcv < DESIRED_UDP_BUF_BYTES / 2 {
+            tracing::warn!(
+                requested = DESIRED_UDP_BUF_BYTES,
+                actual_rcv,
+                "UDP recv buffer smaller than requested — raise net.core.rmem_max (currently clamped by kernel)"
+            );
+        }
+        tracing::info!(actual_rcv, actual_snd, "UDP socket buffer sizes");
         sock.bind(&addr.into()).map_err(|e| {
             io::Error::new(
                 e.kind(),
@@ -1038,5 +1068,96 @@ mod dual_stack_tests {
             peers.iter().all(|(a, _, _)| *a != attacker_addr),
             "spoofed (un-authenticated) source MUST NOT create a per-peer entry"
         );
+    }
+
+    /// Read the kernel's effective UDP recv buffer ceiling on Linux. Returns
+    /// `None` on other platforms or if the sysctl is unreadable. Used by the
+    /// SO_RCVBUF regression test below to decide whether the environment can
+    /// demonstrate the fix.
+    fn read_sysctl_rmem_max() -> Option<usize> {
+        if !cfg!(target_os = "linux") {
+            return None;
+        }
+        std::fs::read_to_string("/proc/sys/net/core/rmem_max")
+            .ok()
+            .and_then(|s| s.trim().parse::<usize>().ok())
+    }
+
+    /// Regression for #3844: `Socket::bind` must request a generous UDP socket
+    /// buffer (16 MiB) so production gateways aren't silently dropping packets
+    /// to the kernel's stock ~208 KiB default. Both nova and vega were
+    /// accumulating UdpRcvbufErrors at 15× the delivered-datagram rate until
+    /// operators raised `net.core.rmem_max` manually.
+    ///
+    /// Method: bind a control socket directly (no buffer tuning) and a tuned
+    /// socket through our `Socket::bind` impl, then compare the resulting
+    /// SO_RCVBUF / SO_SNDBUF values. On constrained kernels where rmem_max
+    /// clamps both sockets to the same default, only the weak `tuned >=
+    /// control` invariant holds. On any system where rmem_max permits a
+    /// multi-megabyte request to materialize, the tuned socket must end up
+    /// with a strictly larger buffer than the untuned control — that's the
+    /// observable effect of the fix.
+    #[tokio::test]
+    async fn udp_socket_bind_requests_large_buffers() {
+        let v4_addr = SocketAddr::new(std::net::IpAddr::V4(std::net::Ipv4Addr::LOCALHOST), 0);
+
+        // Control: bind directly via tokio without going through our
+        // Socket::bind impl, so no buffer-size hints are applied.
+        let control = tokio::net::UdpSocket::bind(v4_addr)
+            .await
+            .expect("control bind should succeed");
+        let control_sock2 = socket2::Socket::from(control.into_std().expect("into_std"));
+        let control_rcv = control_sock2
+            .recv_buffer_size()
+            .expect("control recv_buffer_size");
+        let control_snd = control_sock2
+            .send_buffer_size()
+            .expect("control send_buffer_size");
+
+        // Tuned: bind via our Socket::bind, which should request 16 MiB
+        // SO_RCVBUF / SO_SNDBUF before bind.
+        let tuned = <UdpSocket as Socket>::bind(v4_addr)
+            .await
+            .expect("tuned bind should succeed");
+        let tuned_sock2 = socket2::Socket::from(tuned.into_std().expect("into_std"));
+        let tuned_rcv = tuned_sock2
+            .recv_buffer_size()
+            .expect("tuned recv_buffer_size");
+        let tuned_snd = tuned_sock2
+            .send_buffer_size()
+            .expect("tuned send_buffer_size");
+
+        // Weak invariant: the tuned socket must never end up with a smaller
+        // buffer than the untuned control. This holds on every platform,
+        // even when rmem_max clamps both to the same default.
+        assert!(
+            tuned_rcv >= control_rcv,
+            "tuned recv buffer ({tuned_rcv}) must be at least as large as control ({control_rcv})"
+        );
+        assert!(
+            tuned_snd >= control_snd,
+            "tuned send buffer ({tuned_snd}) must be at least as large as control ({control_snd})"
+        );
+
+        // Strict regression check: on Linux with rmem_max large enough to
+        // honor a multi-MB request, the tuned buffer MUST be strictly
+        // larger than the untuned default. This is the assertion that
+        // would have caught the missing set_recv_buffer_size call.
+        if let Some(rmem_max) = read_sysctl_rmem_max()
+            && rmem_max >= 4 * 1024 * 1024
+        {
+            assert!(
+                tuned_rcv > control_rcv,
+                "with rmem_max={rmem_max}, tuned SO_RCVBUF ({tuned_rcv}) must \
+                 strictly exceed untuned default ({control_rcv}) — the fix did not take effect"
+            );
+        } else {
+            eprintln!(
+                "udp_socket_bind_requests_large_buffers: rmem_max={:?} too small \
+                 to demonstrate fix strictly; observed control_rcv={control_rcv}, \
+                 tuned_rcv={tuned_rcv}, control_snd={control_snd}, tuned_snd={tuned_snd}",
+                read_sysctl_rmem_max()
+            );
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


sanity (commented):
## Comprehensive PR Review: #4266

### Summary
- **PR Title:** fix(transport): set SO_RCVBUF/SO_SNDBUF on UDP sockets
- **Type:** fix (closes #3844)
- **CI Status:** running on `f005d4ac` at time of review post
- **Linked Issues:** #3844
- **Review tier:** Full (transport — high-risk surface)
- **Rounds:** initial review on `6cea640f`, re-review on `e4d2bf01`, light diff-of-diff on `f005d4ac`
- **Reviewers run:** `freenet:code-first-reviewer`, `freenet:testing-reviewer`, `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer`, Codex (external, non-Claude)

---

### Code-First Analysis
**Independent Understanding:** `Socket::bind` for `UdpSocket` now requests 16 MiB SO_RCVBUF/SO_SNDBUF via socket2 before `bind`, then (post-bind) logs the actual buffer sizes and, on Linux, emits a `warn!` if the kernel clamped below half the request. Best-effort: setsockopt failures log and continue. A regression test in `dual_stack_tests` binds a control socket via raw `tokio::net::UdpSocket::bind` and a tuned socket via the trait, then enforces `tuned >= control` cross-platform and `tuned > control` strictly on Linux.

**Stated Intent:** Match — implementation lands exactly at the call site identified by issue #3844, uses the same constants and code shape as the issue's proposed remediation.

**Alignment:** Matches. Diff scope is minimal (single file).

**Gaps:** None.

---

### Testing Assessment
**Coverage Level:** adequate

| Unit | ✅ | Two `#[tokio::test]` variants (IPv4 + IPv6) in `dual_stack_tests`, sharing helper `assert_tuned_bind_grows_buffers` |
| Integration | N/A | Single-fd kernel-buffer behavior is unit-level |
| Simulation | N/A | `InMemorySocket` has no kernel buffer; cannot exercise this path |
| E2E | N/A | |

**Regression Test:** Genuine. Counter-test verified locally: removing `set_recv_buffer_size`/`set_send_buffer_size` causes both IPv4 and IPv6 tests to fail with `"the fix did not take effect"`. The Linux strict assertion `tuned > control` holds on every Linux kernel because `setsockopt(SO_RCVBUF, val)` sets `sk_rcvbuf = min(val, rmem_max) * 2` while an un-tuned socket inherits `rmem_default` *without* doubling — so even on stock Ubuntu (`rmem_max == rmem_default == 212992`) the tuned socket lands at 425984 vs control 212992. The earlier `rmem_max >= 4 MiB` gate, which would have skipped strict coverage on CI runners, was dropped after the first review.

**Missing Tests:**
- Clamp `warn!` path not asserted (would require sysctl manipulation; would be brittle in CI).
- `set_recv_buffer_size` `Err` branch not asserted (hard to provoke in-process; failure mode is benign).
- Acceptable gaps; the functional regression — silently reverting the production fix — IS caught on stock-Linux CI.

---

### Skeptical Findings
**Risk Level:** low

| Concern | Severity | Location | Disposition |
|---------|----------|----------|-------------|
| macOS false-positive clamp warn with Linux-only sysctl advice (initial review H1) | High | `transport.rs:455` | Fixed — clamp warning gated to `#[cfg(target_os = "linux")]` |
| `rmem_max >= 4 MiB` test gate skipped strict assertion on stock CI (initial review M1) | High | `transport.rs:1144` | Fixed — gate removed; strict assertion Linux-only |
| `recv_buffer_size().unwrap_or(0)` would falsely trigger clamp warn (initial review M2) | Medium | `transport.rs:447` | Fixed — uses `.ok()` and `if let Some(...)` |
| `tracing::info!` / clamp `warn!` fired before bind, causing duplicate diagnostics on AddrInUse retry (initial review M3) | Medium | `transport.rs:476` | Fixed — diagnostics moved after successful bind (`f005d4ac`) |
| IPv6 test docstring claimed `[::]:port` but bound `Ipv6Addr::LOCALHOST` (re-review L1) | Low | `transport.rs:1173` | Fixed — uses `Ipv6Addr::UNSPECIFIED` (`f005d4ac`) |
| Strict assertion could fail on exotic kernels (gVisor) without setsockopt doubling | Low | `transport.rs:1144-1156` | Accepted — CI uses standard Linux kernels; not a regression |
| `actual_rcv = ?actual_rcv` uses Debug formatting (`Some(425984)` vs `425984`) | Style | `transport.rs:485` | Accepted — minor consistency nit |

---

### Big Picture Assessment
**Goal Alignment:** Yes — diff is single-file (`crates/core/src/transport.rs`), focused, closes #3844 directly.

**Anti-Patterns Detected:** None. No `#[ignore]`, no test weakening, no scope creep. The `#[cfg(target_os = "linux")]` gates are platform-correctness (the warning's advice is Linux-specific), not test-skipping.

**Removed Code Concerns:** None — diff is purely additive on production; the test was restructured into a helper + two `#[tokio::test]` entry points (IPv4 + IPv6).

**Scope Assessment:** Focused. The deferred items (sysctl docs in setup guide, bug-prevention rule entry) are correctly carved out as follow-ups.

**Operator Impact:** Tuning applies to every node (gateway and client) because it lives at the single per-node UDP bind. ~32 MiB max per node on Linux after kernel doubling — acceptable for any node large enough to participate in Freenet.

---

### Documentation
- Code docs: complete — comments cite #3844, explain Linux doubling semantics, explain the best-effort policy.
- Architecture docs: no existing reference to socket buffer sizing; nothing goes stale.
- User docs: no operator-facing doc currently mentions `rmem_max`; the new clamp warning gives in-line guidance.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None.

#### Consider (Suggestions / non-blocking follow-ups)
1. Document recommended sysctl values (`net.core.rmem_max`, `wmem_max`) in `scripts/FREENET-GW-SETUP-GUIDE.md` so new gateway operators don't need to hit the clamp warning to learn about them.
2. Add a bug-prevention entry in `.claude/rules/transport.md` (or `bug-prevention-patterns.md`) so future agents don't propose removing the tuning on the grounds that "nothing else in the repo references it."
3. Pre-existing `cargo clippy --all-targets` errors in `crates/core/tests/simulation_integration.rs:9421-9422` (`doc_lazy_continuation`, Rust 1.94 lint) — out of scope for this PR; CI's `cargo clippy --locked` is unaffected.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `f005d4ac` (rounds 1 + 2 on prior commits, light diff-of-diff on this commit)

Two rounds of substantive review across four perspectives + Codex; every blocking and important finding was addressed in subsequent commits. The third commit (`f005d4ac`) only moves diagnostics after the bind and switches the IPv6 test address — Codex confirmed no new issues. Counter-test confirms the regression test catches the bug it claims to (deleting the production setsockopt calls fails both IPv4 and IPv6 tests with a clear diagnostic). Pending CI on `f005d4ac` is the only outstanding gate.

[AI-assisted - Claude]

PR #4265, opened 2026-05-26
fix(ops): downgrade queue-full WARN at relay driver wrappers (#4251)

sanity:
## Problem

PR #4253 closed #4251 by gating the relay broadcast amplification (auto-fetch, ResyncRequest) on `is_contract_queue_full()` inside the inner `drive_relay_*` drivers, and downgraded the originator-path ERROR log to DEBUG. It missed one thing: the wrapper functions that *invoke* those drivers still log at WARN for any error the driver returns — including the typed queue-full marker the inner gating relies on.

Six wrapper sites are affected:

- `crates/core/src/operations/update/op_ctx_task.rs`
  - `run_relay_request_update` (`relay_update_request_error`)
  - `run_relay_broadcast_to` (`relay_update_broadcast_error`) ← **production spam source**
  - `run_relay_request_update_streaming` (`relay_update_streaming_request_error`)
  - `run_relay_broadcast_to_streaming` (`relay_update_streaming_broadcast_error`)
- `crates/core/src/operations/put/op_ctx_task.rs`
  - relay (`relay_put_error`)
  - relay streaming (`relay_put_streaming_error`)

## Production impact (today, before the fix)

Same contract `4PjqN55KUCidW8vJvw5fhy5fe5maxXKNrWSyK33QjjVq` from #4251. With v0.2.64 deployed (PR #4253 in place):

- **nova** (`freenet-gateway.service`, v0.2.64): 148,940 `relay_update_broadcast_error` WARNs in a single hour — 60% of all error log lines.  Error log volume `~8 MB/h baseline → 141 MB/h peak today`.
- **technic** (v0.2.64): 1,313 in the same hour, same pattern at lower volume.
- **vega**: 0 of these (different problem — router stuck `success_events=0`, tracked separately under #4239).

The amplification IS suppressed (no GET storm; the inner gating works). Only the log severity at the wrapper boundary regresses.

## Fix

At each of the six wrapper sites, mirror the existing pattern from `update/op_ctx_task.rs:402-419`:

```rust
if err.is_contract_queue_full() {
    tracing::debug!(..., event = "queue_full", "...");
} else {
    tracing::warn!(..., "...");
}
```

Pure log-severity change. No behavior change. Real WASM faults (OOG, traps, missing parameters, etc.) keep their WARN exactly as before — the predicate matches only the typed `ContractQueueFull` marker introduced by PR #4253.

## Testing

[...]

Issue #4213, opened 2026-05-22
test: test_large_secret_data flaky again under heavy parallel lib-test load


sanity:
## Summary

`wasm_runtime::delegate::test::test_large_secret_data` intermittently fails when the full `--lib` test set runs in parallel:

```
thread 'wasm_runtime::delegate::test::test_large_secret_data' panicked at crates/core/src/wasm_runtime/delegate.rs:2531:17:
Expected SecretResult(Some(...)), got SecretResult(None)
```

## Reproduction

Observed while running `cargo test -p freenet --lib -- contract_store config hosting runtime` (many modules in parallel). The test:
- **passes** consistently when run in isolation (`--exact`)
- **passes** when the `wasm_runtime::delegate` group is run alone

So this is a concurrency/shared-state flake that only surfaces under heavier parallel load, not a deterministic break.

## History

This is a recurrence of a previously-fixed flake:
- #3218 — flaky under high concurrency, silent `set_secret` failure
- #3248 — stale WASM memory base pointer

Both are CLOSED. The `SecretResult(None)` symptom matches #3218's "silent set_secret failure" failure mode, suggesting the earlier fix did not fully eliminate the race or a new path reintroduced it.

## Notes

Not introduced by any recent change to `delegate.rs` / `secrets_store.rs`. Filing so the recurring flake is tracked rather than dismissed.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:33:05.433548975+00:00
Issue hash: `b1b8dccd`

## Value Assessment

[...]

PR #4273, opened 2026-05-26
feat(tracing): add per-callsite log rate-limit Layer

sanity:
## Why

Issue #4251 (and its follow-up PR #4265) showed that the existing global \`RateLimiter\` is the wrong tool for the most common log-spam failure mode: a single misbehaving \`tracing::warn!\` macro saturating the log while staying well under the aggregate cap.

Production data point (today, nova v0.2.64):

- One \`tracing::warn!\` site emitted ~40 events/sec from a single hot contract.
- 148,940 lines in a single hour; 60% of the error log; 141 MB/hr peak vs ~8 MB/hr baseline.
- The global \`RateLimiter\` is set to 1000 ev/sec aggregate, so it never engaged.

A per-callsite cap stops this class of incident at the source, regardless of how many distinct callsites are in play across the binary.

## What

\`PerCallsiteRateLimiter\` sits next to the existing global \`RateLimiter\`, composed at the \`filter_fn\` boundary in \`init_tracer\` / \`init_stdout_tracer\`:

\`\`\`rust
let rate_filter = tracing_subscriber::filter::filter_fn(move |meta| {
    per_callsite.as_ref().map(|pc| pc.should_allow(meta)).unwrap_or(true)
        && rate_limiter.should_allow()
});
\`\`\`

Each tracing macro invocation gets its own 1-second window of \`DEFAULT_MAX_EVENTS_PER_CALLSITE_PER_SECOND\` (30) events. Beyond that the events are dropped, counted, and a periodic summary (every 30 s) names the offending callsite — \`Metadata::name\` is \`'static\` and typically formats as \`"event <file>:<line>"\`, sufficient to pinpoint the chatty site.

## Defaults & tuning

- **30 ev/sec/callsite** by default. Above realistic bursty INFO sites (connection bring-up etc. ~10-20 ev/sec briefly); well below the 40 ev/sec spam observed in #4251. Tunable per deployment via \`FREENET_LOG_RATE_LIMIT_PER_CALLSITE\`.
- Disabled in test/debug builds and when \`FREENET_DISABLE_LOG_RATE_LIMIT\` is set — same gating as the global limiter, no surprises in CI.
- Bounded memory: \`MAX_TRACKED_CALLSITES = 4096\`. Beyond that the limiter degrades to "pass" (better to log than OOM).

## Design notes

- **Composes, doesn't replace.** Keep the global limiter as the aggregate-burst backstop. Both can drop an event; neither alone is sufficient defence.
- **Capture only \`'static\` fields.** In \`tracing-core 0.1.x\`, \`Metadata::file\`/\`line\`/\`target\` return the metadata's borrowed lifetime — only \`name\` is \`'static\`. The per-callsite state stores only \`name\`, which is good enough for the summary line.
- **No new direct dep for production.** \`tracing-core\` is already in the workspace tree; declaring it explicitly so the test code can construct synthetic \`Identifier\` values without going through \`Metadata::new\`'s static-lifetime maze.

## Testing

7 new unit tests in \`util::rate_limit_layer::tests\`:

- \`per_callsite_allows_under_limit\`
- \`per_callsite_drops_over_limit\`
- \`per_callsite_isolates_sites_from_each_other\` — the critical regression check; a chatty site A must NOT starve quiet site B
- \`per_callsite_window_resets\`
- \`per_callsite_summary_path_is_safe_against_dashmap_borrow_self_collision\`
- \`per_callsite_default_uses_canonical_constant\`
- \`per_callsite_is_cloneable_and_shares_state\`

ZST-callsite gotcha worth flagging for reviewers: the test \`TestCallsite\` must carry a non-zero-size tag (\`_tag: usize\`). Two static ZSTs may share an address per the unsized-locals rules, so a tagless ZST silently failed the isolation test by collapsing two static instances into the same \`Identifier\`.

- \`cargo test -p freenet --lib\`: **2658 passed, 0 failed, 14 ignored.**
- \`cargo clippy -p freenet --lib --tests -- -D warnings\`: clean.
- \`cargo fmt -p freenet\`: no diff (imports reordered).

## Out of scope

- Removing the global \`RateLimiter\`. Kept as aggregate-burst backstop.
- Per-\`Level\` differentiation (different caps for INFO vs WARN vs ERROR). Easy to add later if real workloads need it.

Refs #4251 #4265

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2065,6 +2065,7 @@ dependencies = [
  "tower-http",
  "tracing",
  "tracing-appender",
+ "tracing-core",
  "tracing-opentelemetry",
  "tracing-subscriber",
  "tray-icon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ opentelemetry-otlp = "0.32.0"
 opentelemetry_sdk = { version = "0.32", features = ["rt-tokio"] }
 tracing = "0.1"
 tracing-appender = "0.2"
+tracing-core = "0.1"
 tracing-opentelemetry = "0.33.0"
 tracing-subscriber = "0.3"
 console-subscriber = "0.5.0"
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -95,6 +95,7 @@ opentelemetry-jaeger = { workspace = true, features = ["collector_client", "isah
 # release_max_level_info, reducing the number of spans EnvFilter must evaluate.
 # Debug builds retain all levels. RUST_LOG can filter among compiled-in levels.
 tracing = { workspace = true, features = ["release_max_level_info"] }
+tracing-core = { workspace = true }
 tracing-opentelemetry = { workspace = true, optional = true }
 tracing-subscriber = { workspace = true, features = ["json"], optional = true }
 tracing-appender = { workspace = true, optional = true }
--- a/crates/core/src/tracing.rs
+++ b/crates/core/src/tracing.rs
@@ -4262,13 +4262,29 @@ pub mod tracer {
             .and_then(|v| v.parse().ok())
             .unwrap_or(crate::util::rate_limit_layer::DEFAULT_MAX_EVENTS_PER_SECOND);
 
+        // Per-callsite cap (issue #4251 follow-up). Stops a single misbehaving
+        // tracing macro from dominating the log even when its rate stays
+        // below the global aggregate cap. Configurable via
+        // FREENET_LOG_RATE_LIMIT_PER_CALLSITE.
+        let per_callsite_limit: u64 = std::env::var("FREENET_LOG_RATE_LIMIT_PER_CALLSITE")
+            .ok()
+            .and_then(|v| v.parse().ok())
+            .unwrap_or(crate::util::rate_limit_layer::DEFAULT_MAX_EVENTS_PER_CALLSITE_PER_SECOND);
+
         // Rate limiting is disabled in tests and debug builds to avoid masking issues
         let rate_limit_enabled = !cfg!(any(test, debug_assertions))
             && std::env::var("FREENET_DISABLE_LOG_RATE_LIMIT").is_err();
 
-        // Create rate limiter (shared across all layers)
+        // Create rate limiters (shared across all layers)
         let rate_limiter = if rate_limit_enabled {
             Some(crate::util::rate_limit_layer::RateLimiter::new(rate_limit))
+        } else {
+            None
+        };
+        let per_callsite_limiter = if rate_limit_enabled {
+            Some(crate::util::rate_limit_layer::PerCallsiteRateLimiter::new(
+                per_callsite_limit,
+            ))
         } else {
             None
         };
@@ -4285,6 +4301,7 @@ pub mod tracer {
                         use_json,
                         filter_layer,
                         rate_limiter,
+                        per_callsite_limiter,
                     );
                 }
 
@@ -4322,8 +4339,14 @@ pub mod tracer {
                 // Apply rate limiting as a global filter if enabled
                 // Layers must be created after the rate filter to ensure type compatibility
                 if let Some(rate_limiter) = rate_limiter.clone() {
-                    let rate_filter =
-                        tracing_subscriber::filter::filter_fn(move |_| rate_limiter.should_allow());
+                    let per_callsite = per_callsite_limiter.clone();
+                    let rate_filter = tracing_subscriber::filter::filter_fn(move |meta| {
+                        per_callsite
+                            .as_ref()
+                            .map(|pc| pc.should_allow(meta))
+                            .unwrap_or(true)
+                            && rate_limiter.should_allow()
+                    });
                     let base = Registry::default().with(rate_filter);
 
                     // Create layers for main and error logs (typed against rate-filtered registry)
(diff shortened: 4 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4273 (Re-Review)

### Summary
- **PR Title:** feat(tracing): add per-callsite log rate-limit Layer
- **Type:** feat (logging infrastructure)
- **CI Status:** mostly green on current HEAD; Unit & Integration / Simulation / NAT Validation still pending
- **Linked Issues:** #4251 (follows up the wrapper-fix PR #4265 with the structural safeguard)
- **Review tier:** Light + external model — small focused diff, tracing init / utility class only
- **Reviewers run (two rounds):**
  - **Round 1 (HEAD a8c0dd9a):** `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer`, Codex (`codex review --base origin/main`)
  - **Round 2 (HEAD 63b7ccf5 then 114e959c):** `freenet:skeptical-reviewer`, Codex
- **HEAD reviewed:** **114e959c** — last change vs. fully-reviewed 63b7ccf5 is a single-constant value adjustment (Codex P2 follow-up). Light re-check applied.

---

### Round 1 → Round 2 Findings & Resolutions

| # | Finding | Source | Status |
|---|---------|--------|--------|
| 1 | **`filter_fn` is callsite-cacheable — limiter runs only on first event per callsite** (P1, blocking). Bug also present in the pre-existing global `RateLimiter`. | Codex round 1 | **Fixed in 63b7ccf5** by switching to `DynFilterFn::new(\|meta, _cx\| ...)` at both wiring sites. Skeptical round 2 verified `DynFilterFn`'s default `callsite_enabled` returns `Interest::sometimes`, no `max_level_hint` set on either site. New regression test `dynfilter_wiring_actually_filters_per_event` installs the production composition, emits 20 events from one macro, asserts only the cap (3) lands. |
| 2 | **Default 100 ev/sec/callsite doesn't catch the #4251 spam pattern (~40 ev/sec)** (P2). | Codex round 2 | **Fixed in 114e959c** by lowering default to 30 ev/sec/callsite (below the observed spam rate). Trade-off (legitimate WARN bursts >30/sec may be clipped) documented in the constant's docstring + tunable via env var. |
| 3 | **`MAX_TRACKED_CALLSITES` degrades to "always pass" silently when table fills** (Medium). | Skeptical round 1 | **Fixed in 63b7ccf5** by adding `AtomicBool` latch + one-shot `eprintln!` operator notice. Skeptical round 2 verified the `compare_exchange(false, true, SeqCst, Relaxed)` is race-free. |
| 4 | **Summary smoke test name overpromises** ("safe_against_dashmap_borrow_self_collision" — actually only tests single-threaded smoke). | Skeptical round 1 | **Fixed in 63b7ccf5** — renamed to `per_callsite_summary_path_smoke_single_threaded` with honest docstring. Defensive `drop(entry)` remains. |
| 5 | Doc-hidden `Identifier(&'static dyn Callsite)` field used in tests — upstream stability concern (Low). | Skeptical round 1 | Documented in PR description; future tracing-core patch bump could break the test only, not production. Accepted. |
| 6 | Atomic ordering boundary fuzz — under-allow at second boundaries possible (Low). | Skeptical round 1 | Confirmed tolerable for a rate limiter; under-allow is the safer failure mode. Accepted. |
| 7 | Stale ref in `crates/core/CLAUDE.md:40` ("rate_limiter.rs") — pre-existing, not introduced by this PR. | Big-picture round 1 | Out of scope; tracked separately. |

---

### Code-First Analysis (final)
**Independent Understanding:** A new `PerCallsiteRateLimiter` (per-callsite 1-second sliding window, default 30 ev/sec/callsite, bounded by `MAX_TRACKED_CALLSITES=4096`) is composed with the existing global `RateLimiter` in both `init_tracer` and `init_stdout_tracer`. Both are now wired through `DynFilterFn` instead of `filter_fn`, which is the load-bearing fix that also un-bricks the pre-existing global limiter.

**Stated Intent:** Systemic safeguard against the #4251 class of single-callsite log spam. Goal alignment: yes.

---

### Testing Assessment
**Coverage Level:** Adequate.

| Unit | ✅ | 13 limiter tests pass: behavior, boundary, isolation, cloneability, default constant. |
| Wiring regression | ✅ | `dynfilter_wiring_actually_filters_per_event` installs `Registry + DynFilterFn + CountingLayer`, emits 20 events from one macro, asserts cap=3 fires. This test would have failed with the original buggy `filter_fn` wiring (would have observed all 20). |
| Full lib | ✅ | `cargo test -p freenet --lib`: 2659 passed, 0 failed, 14 ignored. (One pre-existing flake on `test_large_secret_data` from issue #4213 — passes on retry, untouched by this PR.) |

---

### Verdict
**State:** Ready to Merge.

**HEAD SHA reviewed:** 114e959c — light re-check applied. The change vs. fully-reviewed 63b7ccf5 is a single-constant value (default cap 100 → 30) addressing Codex's P2, with a docstring update. No new logic, no new test surface, no behavioral risk beyond the per-second cap value itself, which is the entire point of the parameter.

Once Unit & Integration / Simulation / NAT Validation finish green on 114e959c, this is ready to land. The codex P1 fix is a particularly load-bearing improvement that also fixes the pre-existing global `RateLimiter` (which had been a no-op since introduction — explains why the #4251 spam slipped past the 1000 ev/sec aggregate cap).

[AI-assisted - Claude]

PR #4279, opened 2026-05-27
feat(contracts): detect & gate non-idempotent contracts (#4251 follow-up)

sanity:
## Problem

The May 2026 `4PjqN5…` runaway (#4251, partially mitigated in #4231 / #4253) was caused by a contract whose `update_state` violates idempotency — i.e. `update_state(update_state(S, U), U) != update_state(S, U)`. Every peer's merge produces a byte-different state from the same logical content, which propagates, which every peer re-merges. The per-contract executor queue saturates and the WS API dies. The reactive fixes shipped in 0.2.65 stop _amplification_ (queue-full no longer re-queues) but don't catch the root-cause failure shape — a misbehaving contract still floods its own merges and broadcasts.

This PR adds the **first generic, root-cause-targeted defense**: an in-peer detector for the property `update_state(update_state(S, U), U) == update_state(S, U)`, which every correct CRDT-shaped contract satisfies by definition.

## Approach

* **Sampled probe at the merge site.** With probability `IDEMPOTENCY_PROBE_PROBABILITY = 1/32` (via `GlobalRng` — simulation tests stay deterministic), after a state-changing merge produces `S'`, re-invoke `update_state(S', updates)` in the same runtime. If the result is byte-different from `S'`, the contract is flagged. ~3% executor overhead on heavily-trafficked contracts, ~0 on quiet ones.
* **Persistent per-contract flag.** `BrokenInvariant::NonIdempotent` flag held in a DashMap on `Ring`, backed by a new `BROKEN_INVARIANTS_TABLE` in the ReDb store. Wired through the same `handler.rs::set_*_storage` pattern as `HostingManager`. A node that detected the bug doesn't re-engage the storm on restart.
* **Gate the broadcast emission.** Both `BroadcastStateChange` emission sites in `contract/executor/runtime.rs` (`commit_state_update` and `broadcast_state_change`) check `Ring::is_contract_broken` before emitting. Flagged contracts' state never leaves the node.
* **Skip commit on probe-positive.** When the probe fires positive on a fresh merge, the commit is skipped too — there's no reason to extend the problematic state locally; subsequent inbound updates get the same treatment, and the contract effectively freezes for this node.
* **`RelatedState` exempted.** The re-apply property doesn't hold for `RelatedState` inputs by contract spec (it's a cross-contract hint, not a CRDT operation), so probes skip merges driven exclusively by `RelatedState`.

## WASM dissection of the offending contract

Live confirmation (`/home/ian/.claude/jobs/.../wasm-analysis.md` — see [issue thread]): the `4PjqN5…` instance resolves to `bdtchyck…wasm` (228 KB; private repo `freenet/freenet-idle-poc`, `shared-wire/src/guilds.rs`). Host-induced non-determinism is **fully ruled out** — single host import (`__frnt__fill_buffer`), no `getrandom`, no clock, no WASI. The bug is intrinsic to the contract's own `Vec<GuildOp>` merge logic. Two indistinguishable sub-mechanisms (order-dependent sort or position-dependent ed25519 re-signing); the single re-apply property catches both regardless.

## Why this targeting (vs. a rate-limit/throttle)

A flat per-contract broadcast rate-limit would penalize any legitimately fast or large-update contract. The non-idempotency probe doesn't — a correct CRDT passes the property regardless of update rate, payload size, or fan-out. Zero false positives by construction.

## Why local-only (no wire message)

Each peer detects independently. There's no `BrokenContract` wire variant. Adding one would create a denial-of-service surface (a malicious peer broadcasts a false flag to silence a healthy contract). The cascade is good enough: as detection spreads peer-by-peer, the storm collapses naturally.

## Explicitly out of scope (follow-ups)

This PR is the smallest correct shape that kills the storm. The following are intentional follow-ups so this can review and merge in a single sitting:

* **Inbound-relay gate** at `update::op_ctx_task::drive_relay_broadcast_to` — currently a flagged peer still wastes one WASM exec on incoming `BroadcastTo` before deciding not to propagate. Belt-and-suspenders.
* **Subscribe / hosting-advertise gates** — stop advertising broken contracts to neighbors so route-around happens too.
* **Real-WASM regression fixture** using nova's captured 464-byte state for `4PjqN5…`. Currently the synthetic `UpdateOverride::NonIdempotent` mock pins the fixture's _bug shape_ correctly (test asserts re-apply produces different bytes) but doesn't exercise actual WASM. Slow to assemble correctly — separate PR.
* **Additional `BrokenInvariant` kinds** (non-commutativity, hard-timeout repeat-offender). The enum/storage shape is forward-compatible (single byte on disk, `from_byte` returns `Option`).

## How this composes with the contract-hardening initiative (#4260)

Strictly additive. PR #4260 (already merged) covers the **resource-cost** axis: `Meter`, MAD-based outlier scoring, eviction-order refinements. This PR covers the **correctness** axis. The `BrokenInvariant` flag is the natural input for the governance reaper PR's eviction-policy decision tree — a contract that's both _expensive_ AND _broken_ is the highest-priority eviction candidate.

## Testing

* **2703 lib tests pass**, 0 failures, 14 ignored (pre-existing).
* New unit tests in `ring::broken_invariants::tests` (4) — record/query, idempotent record, unrelated-contracts isolation, on-disk byte roundtrip.
* New tests in `non_idempotent_detector_tests` (2) — pin the mock fixture's bug shape (re-apply produces distinct bytes, _and_ stays the same size like the production 464-byte observation); pin the healthy baseline (default mock _is_ idempotent on re-apply, no false positives).
* `cargo clippy --lib --tests -- -D warnings` clean.
* `cargo fmt` clean.

## Risk register

* **False positive on a correct CRDT** — impossible by definition of the property. Tests pin the no-false-positive baseline.
* **WASM throughput overhead** — 1/32 sampling: ~3% executor CPU on active contracts. Tuneable via the `IDEMPOTENCY_PROBE_PROBABILITY` constant if observed too high.
* **Probe itself errors out** (timeout, trap) — handled: bail without flagging, log at DEBUG. Doesn't change the original merge outcome.
* **Persistence corruption** — `load_all_broken_invariants` skips malformed rows with a warning rather than failing startup. Worst case: re-detect on first sampled merge after restart.
* **Migration risk** — adds a new ReDb table. redb auto-creates missing tables on `open_table` inside a write transaction, so existing databases pick it up on first startup of the new binary. No explicit migration step.

## Fixes

Closes part of #4251 (root-cause detector for the specific bug shape; resource-throttling / reaper work continues in the governance track).

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/contracts.md
+++ b/.claude/rules/contracts.md
@@ -145,6 +145,17 @@ State merging rules:
   - Contract defines merge semantics
   - Merge MUST be commutative: merge(a, b) == merge(b, a)
   - Merge MUST be associative: merge(merge(a, b), c) == merge(a, merge(b, c))
+  - For `UpdateData::State` inputs, merge MUST be idempotent:
+      update_state(update_state(S, State(X)), State(X)) == update_state(S, State(X))
+    Enforced at runtime via the in-peer probe in
+    `Executor::maybe_probe_idempotency`. Violators get flagged in
+    `Ring::broken_invariants` and their outbound `BroadcastStateChange`
+    is suppressed locally; the flag is permanent for that contract
+    instance id and survives restart. See `ring::broken_invariants` and
+    issue #4251 / PR #4279.
+    Note: this invariant is NOT enforced for `UpdateData::Delta` inputs
+    (CmRDT-style "increment by X" deltas legitimately violate it) or
+    `UpdateData::RelatedState` (a cross-contract hint, not a CRDT op).
   - Invalid merges should return error, not panic
 ```
 
@@ -306,7 +317,7 @@ Handling:
 
 ```
 □ Test contract validation (valid and invalid states)
-□ Test state merging (commutativity, associativity)
+□ Test state merging (commutativity, associativity, idempotency on State inputs)
 □ Test execution timeout handling
 □ Test related contract fetching
 □ Test concurrent contract updates
--- a/crates/core/src/contract/executor/mock_wasm_runtime.rs
+++ b/crates/core/src/contract/executor/mock_wasm_runtime.rs
@@ -39,6 +39,17 @@ pub(crate) enum UpdateOverride {
     /// paths (issue #4151): the returned error must be classified as
     /// `is_invalid_update_rejection()` and logged at DEBUG, not INFO.
     RejectInvalidUpdate { reason: String },
+    /// Models a non-idempotent contract: every call to `update_state`
+    /// returns a state that is byte-different from the previous one even
+    /// when the input update is the same. The mock prepends an internal
+    /// monotonically-increasing counter to the state bytes, mimicking the
+    /// shape of a real contract that embeds a timestamp / position-
+    /// dependent signature / re-signed payload — the smoking-gun shape
+    /// the in-peer detector is built to catch (see #4251 and the
+    /// `bdtchyck…wasm` analysis in `~/.claude/jobs/.../wasm-analysis.md`).
+    /// Used by tests to verify the idempotency probe fires on a real-
+    /// world-shaped failure mode.
+    NonIdempotent(std::sync::Arc<std::sync::atomic::AtomicU64>),
 }
 
 /// A lightweight mock runtime at the `ContractRuntimeInterface` level that lets
@@ -143,6 +154,38 @@ impl ContractRuntimeInterface for MockWasmRuntime {
                     return Err(crate::wasm_runtime::ContractError::from(
                         ContractExecError::ContractError(inner_err),
                     ));
+                }
+                UpdateOverride::NonIdempotent(counter) => {
+                    // Pick the "logical" input state — same precedence as
+                    // the default branch — then prepend a monotonically-
+                    // increasing 8-byte counter. Re-running with the
+                    // produced state as input still bumps the counter,
+                    // so the result is byte-different every call.
+                    let logical = update_data
+                        .iter()
+                        .find_map(|u| match u {
+                            UpdateData::State(s) => Some(s.as_ref().to_vec()),
+                            UpdateData::Delta(d) => Some(d.as_ref().to_vec()),
+                            UpdateData::StateAndDelta { state, .. } => {
+                                Some(state.as_ref().to_vec())
+                            }
+                            UpdateData::RelatedState { .. }
+                            | UpdateData::RelatedDelta { .. }
+                            | UpdateData::RelatedStateAndDelta { .. } => None,
+                            // `UpdateData` is `#[non_exhaustive]`; allow future
+                            // variants to fall through to the `_state` fallback.
+                            _ => None,
+                        })
+                        .unwrap_or_else(|| _state.as_ref().to_vec());
+                    let n = counter.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+                    let mut out = Vec::with_capacity(8 + logical.len().saturating_sub(8));
+                    out.extend_from_slice(&n.to_le_bytes());
+                    // Strip any prior counter prefix so the state stays a
+                    // fixed size — matching the 464-byte shape we saw in
+                    // production rather than growing unboundedly.
+                    let tail_start = logical.len().min(8);
+                    out.extend_from_slice(&logical[tail_start..]);
+                    return Ok(UpdateModification::valid(out.into()));
                 }
             }
         }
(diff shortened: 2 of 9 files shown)


sanity:
Pushed fixup commit `9b83456` addressing multi-model review findings.

## What changed

- **Skeptical P1 + Codex P1 (false positives on Delta):** Probe now only fires when **every** entry is `UpdateData::State(_)`. `Delta`/`StateAndDelta` are conservatively exempted to avoid mass-flagging legitimate CmRDT-style "increment-by-X" contracts. Gemini noted the trade-off cuts both ways since Freenet has at-least-once delivery — we go conservative for v1; Delta-probing can be a follow-up once we have a per-op-id dedup story.
- **Skeptical M4 (commit_state_update bypass):** Moved the `is_contract_broken` gate to the **top** of `commit_state_update`. All three call sites (`bridged_upsert`, related-contract retry, validation re-attempt) are now blanketed.
- **Codex P2 (sqlite build):** Pre-existing main breakage on `--no-default-features --features sqlite` (other methods like `load_all_contract_index` also missing on sqlite Pool), so this PR doesn't regress it — but `#[cfg(feature = "redb")]`-gated the new calls defensively.
- **Testing #1 (persistence not actually tested):** Added 3 round-trip tests — `broken_invariants_persistence_round_trip` (store → close → reopen → load — the load-bearing "survives restart" claim), `broken_invariants_store_is_upsert_not_append`, and `broken_invariants_load_skips_malformed_value`.
- **Big-picture #3 (stale rule doc):** Updated `.claude/rules/contracts.md` "WHEN merging states" to add the idempotency invariant explicitly with scope note about Delta exemption, plus the testing checklist line.
- **Skeptical M5 / Big-picture (no recovery API):** Added `BrokenInvariantsTracker::clear` as the operator escape hatch. Marked `#[allow(dead_code)]` — debug-CLI wiring is a follow-up.
- **Big-picture #5 (stale docstring):** `IDEMPOTENCY_PROBE_DENOMINATOR` → `IDEMPOTENCY_PROBE_PROBABILITY`.

## Deferred (now explicitly documented in the PR description + follow-up issues to file post-merge)

- Inbound-relay gate at `drive_relay_broadcast_to` — flagged peer's executor queue is ~50% relieved (post-flag merges short-circuit via the `commit_state_update` top-gate); full relief depends on detection cascading to inbound neighbors. Belt-and-suspenders, not blocking.
- `SyncStateToPeer` gate (big-picture noted this).
- Delta-input probing once Gemini's argument about Freenet's at-least-once delivery is taken to its conclusion.
- Meter integration (`AttributionSource::BrokenInvariant{...}`) so the governance reaper can consume the signal.

## Tests
2718 lib tests pass, 0 failures, 14 ignored (pre-existing). New: 3 persistence + 4 tracker + 2 fixture = 9 tests in this PR. `cargo clippy --lib --tests -- -D warnings` clean. `cargo fmt` clean.

[AI-assisted - Claude]

sanity:
Pushed `818dc35a` addressing skeptical re-review's blocker #1 (`clear()` had no disk counterpart, making operator unflag undone on next restart).

* Added `ReDb::remove_broken_invariant` + wired it from `BrokenInvariantsTracker::clear` (same `#[cfg(feature = "redb")]` gating as the other persistence calls).
* Added 2 tests: `clear_returns_previous_and_unsets` (unit) + `broken_invariants_remove_makes_load_empty` (round-trip across close/reopen).

**Re-review's other 2 findings** — `perform_contract_put` re-PUT bypass and probe coverage gap on `attempt_state_update` callers — are real but scoped to v1 limitations and filed for follow-up:
- Outbound broadcast IS still gated everywhere (the `commit_state_update` top-gate + `broadcast_state_change` gate cover all broadcast paths). What's not gated is the local-only state-write on the client-init re-PUT-merge path — a flagged contract's state can still extend locally but won't propagate.
- The probe is bound to `bridged_upsert_contract_state` (the network UPDATE relay receiver), which is exactly where real-world storm traffic flows. `attempt_state_update`'s other callers (queue-replay, client `update_contract`) don't carry network storm traffic so the practical detection gap is small.
- Both gaps will be tightened in the follow-up that adds the inbound-relay gate + subscribe/advertise gates.

Codex re-review on commit `9b834566` came back clean (no findings). Final codex pass on `818dc35a` running now.

2718 lib tests pass. clippy + fmt clean.

[AI-assisted - Claude]

Issue #3967, opened 2026-04-28
Service wrapper resilience: stale wrapper scripts + orphan port-holding processes don't self-heal


iduartgomez:
## Problem

A user-facing service (running on `:7509`) silently froze on version 0.2.37
for 19 days while the on-disk binary had been upgraded to 0.2.49. The
launchd-managed wrapper kept respawning the new binary every 5 minutes
(MAX_BACKOFF), but every spawn exited 43 ("Port already in use") because
an orphaned `freenet network` process from a prior run still held the port
with PPID=1 (detached from the wrapper). The user only noticed because the
homepage at http://127.0.0.1:7509/ kept serving the old version's embedded
assets and a frozen connected-peers list.

Recovery required:
1. `kill -9` the orphan PID by hand.
2. `launchctl kickstart -k gui/$UID/org.freenet.node` to break out of the
   5-minute backoff that the wrapper had reached after retrying for 19 days.

The current wrapper template in
`crates/core/src/bin/commands/service.rs::generate_wrapper_script`
already has a fix for this exact scenario: it `pkill`s stale `freenet
network` processes on startup and on detected "already in use" failures.
But that fix only helps users who *re-ran* `freenet service install`
after the template was updated — every existing install still has the
old wrapper script (no startup `pkill`, no port-conflict detection,
no MAX_PORT_CONFLICT_KILLS recovery). There is no mechanism to
re-template the on-disk wrapper when the binary is updated.

## Concrete failure timeline

```
Old wrapper installed:   2026-01-16
Binary upgraded:         2026-04-23  (0.2.37 → 0.2.49 on disk)
Orphan process started:  2026-04-08  (PID 66258, PPID=1, holds :7509)
Wrapper retry loop:      2026-04-08 → 2026-04-28  (19 days, exit 43 every 5min)
Manual recovery:         2026-04-28
Diagnosis cost:          ~30min (process tree, log dive, launchctl)
```

## Why this matters

This is the second self-inflicted resilience failure in the same
deployment surface. Past examples (#3934 exit-42 loop, #3928 cfg'd
macOS path) show the same pattern: a fix lands in the source-of-truth
template, but deployed machines never receive it because there is no
update path for the wrapper / plist / cfg artifacts that ship alongside
the binary.

Auto-update should not just replace the binary — it should also reconcile
the supporting service artifacts (wrapper script, plist, exit-code lists)
when their templates change.

### 1. Re-template wrapper + plist on `freenet update`

[...]

PR #4286, opened 2026-05-27
fix: detect stale macOS service wrapper by content equality (#3967)

sanity:
## Problem

`ensure_service_file_updated` (macOS branch) decided whether the
on-disk launchd wrapper was stale by grepping for a fixed set of
marker substrings (`EXIT_CODE=$?` + `freenet update` + absence of the
legacy log path). Any future template change whose markers weren't
part of the check was silently invisible — including the orphan-killing
`pkill -f -u "$(id -u)" "freenet network"` block added on startup
after the Jan 2026 wrapper.

Two independent production hosts hit this in two weeks: their Jan 16
wrappers passed the marker check on every `freenet update`, so they
never received the pkill block, and a `freenet network` orphan with
PPID=1 kept holding port 7509 across launchd restarts. Both nodes
silently served stale assets for ~19 days until a human noticed.

## Solution

Replace the marker grep with a byte-for-byte comparison against what
`generate_wrapper_script` produces now — any template drift, binary
path move, or hand-edit triggers a regen. To avoid silently overwriting
hand-edited wrappers, also write a `freenet-service-wrapper.sh.hash`
sidecar recording the SHA-256 of what we last wrote. A mismatch
between the sidecar and the on-disk hash blocks the rewrite with a
warning instead of clobbering the user's edits.

Both `freenet service install` (fresh installs) and
`ensure_service_file_updated` (auto-update path) write the sidecar
through a shared `write_wrapper_hash_sidecar` helper so the on-disk
format is single-source-of-truth.

**Migration:** pre-#3967 installs have no sidecar; that case is
treated as "ours" so the next auto-update fixes the wrapper and then
records the sidecar for future user-modification protection. The
existing `.sh.bak` backup is still written first, so a regen is
recoverable. Malformed (truncated/corrupted) sidecars are also
treated as "missing" rather than locking the host into permanent
UserModified.

**Self-heal:** if the wrapper is up-to-date but the sidecar is
missing or malformed (e.g. a prior sidecar write failed), the next
update backfills it. Without this self-heal, a host that lost its
sidecar would stay in the migration state forever and a future
hand-edit could be silently clobbered.

## Architecture

The decision logic is a cross-platform pure helper
`decide_wrapper_action(template, on_disk_wrapper?, sidecar?) ->
WrapperAction`. The macOS-only `ensure_service_file_updated` just
gathers the inputs and dispatches on the action. Every CI runner —
not just macOS — exercises the full decision matrix via the pure
helper tests below.

## Scope

This PR addresses **the deployment gap that caused the two production
incidents** (items 1 and 2 of issue #3967's checklist: re-template on
update; self-healing wrapper). The remaining acceptance criteria are
filed as follow-ups so they don't get lost:

- **#4287** — Linux systemd unit uses the same fragile marker pattern;
  apply the same equality + sidecar treatment.
- **#4288** — surface stuck-wrapper to the user after N consecutive
  failures (notification or homepage banner).
- **#4289** — homepage detects asset/runtime version mismatch and
  shows a warning.
- **#4290** — `freenet uninstall` should clean up the wrapper +
  sidecar + .bak files.

Per the global `finish-the-fix.md` rule: the user-visible symptom
(silent stale-asset serving for 19 days on macOS) goes away with
this PR alone, because the deployed wrapper finally gets the pkill
block. The other items are real but not load-bearing on the
reported symptom.

## Testing

**Cross-platform pure-helper tests** (run on every CI runner):

- `decide_wrapper_action_no_wrapper_writes_ours`
- `decide_wrapper_action_stale_wrapper_no_sidecar_writes_ours_3967` —
  Jan 2026 migration shape.
- `decide_wrapper_action_stale_wrapper_matching_sidecar_writes_ours`
- `decide_wrapper_action_stale_wrapper_with_trailing_newline_sidecar` —
  pins the `\n`-suffixed on-disk format.
- `decide_wrapper_action_user_modified_wrapper_is_preserved`
- `decide_wrapper_action_malformed_sidecar_treated_as_missing_3967`
- `decide_wrapper_action_up_to_date_with_matching_sidecar_is_noop` —
  install -> update no-op contract.
- `decide_wrapper_action_up_to_date_with_missing_sidecar_backfills` —
  self-heal.
- `decide_wrapper_action_up_to_date_with_malformed_sidecar_backfills`
- `wrapper_with_auto_update_hook_but_no_pkill_block_needs_regen_3967` —
  the original incident reproduced as a pin test.
- `wrapper_identical_to_template_does_not_need_regen` — no regen loop.
- `wrapper_needs_regen_detects_single_char_drift`
- `is_valid_wrapper_hash_accepts_real_sha256_hex_only` — sidecar
  validation: rejects empty, truncated, too-long, uppercase, non-hex.
- `wrapper_content_hash_is_stable_and_distinct`

**macOS-only end-to-end tests** (run when CI is on macOS; serve as
integration coverage):

- `freshly_generated_wrapper_does_not_trigger_regen_loop` — calls
  the real `generate_wrapper_script` and asserts determinism.
- `ensure_service_file_updated_rewrites_pre_pkill_wrapper_and_records_hash`
- `ensure_service_file_updated_is_noop_after_fresh_install`
- `ensure_service_file_updated_backfills_missing_sidecar`
- `ensure_service_file_updated_treats_malformed_sidecar_as_missing`
- `ensure_service_file_updated_preserves_user_modified_wrapper`

`cargo fmt`, `cargo clippy --all-targets -p freenet -- -D warnings`,
and `cargo test -p freenet` all clean locally (163 bin tests pass,
2743 lib tests pass).

## Why didn't CI catch this?

The old marker-based staleness check was a self-referential predicate:
a unit test asserting `contains("EXIT_CODE=$?")` doesn't verify
whether the on-disk artifact actually matches the current template.
The new equality-based check is naturally self-validating, and the
extraction of `decide_wrapper_action` into a pure helper means every
branch runs on every CI runner regardless of target OS.

Addresses #3967 (macOS item; see also #4287, #4288, #4289, #4290 for
the remaining acceptance criteria).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2814,9 +2814,27 @@ fn install_macos_service() -> Result<()> {
     fs::create_dir_all(&wrapper_dir).context("Failed to create wrapper directory")?;
     let wrapper_path = wrapper_dir.join("freenet-service-wrapper.sh");
     let wrapper_content = generate_wrapper_script(&exe_path);
-    fs::write(&wrapper_path, wrapper_content).context("Failed to write wrapper script")?;
+    fs::write(&wrapper_path, &wrapper_content).context("Failed to write wrapper script")?;
     fs::set_permissions(&wrapper_path, fs::Permissions::from_mode(0o755))
         .context("Failed to make wrapper script executable")?;
+
+    // Record the SHA-256 hash of the wrapper we just wrote so a later
+    // `freenet update` can distinguish "Freenet's wrapper" from "a
+    // hand-edited wrapper" before overwriting it (issue #3967). A
+    // failed sidecar write only weakens user-modification protection
+    // on the next update — it does not invalidate the wrapper we just
+    // wrote — so we warn and continue.
+    let wrapper_hash_path = wrapper_path.with_extension("sh.hash");
+    let wrapper_hash = super::update::wrapper_content_hash(&wrapper_content);
+    if let Err(e) = fs::write(&wrapper_hash_path, format!("{wrapper_hash}\n")) {
+        eprintln!(
+            "Warning: failed to write wrapper hash sidecar at {}: {}. \
+             User-modification detection on the next `freenet update` \
+             will fall back to treating the wrapper as ours.",
+            wrapper_hash_path.display(),
+            e
+        );
+    }
 
     let plist_content = generate_plist(&wrapper_path, &log_dir);
     let plist_path = launch_agents_dir.join("org.freenet.node.plist");
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2818,12 +2818,10 @@ fn install_macos_service() -> Result<()> {
     fs::set_permissions(&wrapper_path, fs::Permissions::from_mode(0o755))
         .context("Failed to make wrapper script executable")?;
 
-    // Record the SHA-256 hash of the wrapper we just wrote so a later
-    // `freenet update` can distinguish "Freenet's wrapper" from "a
-    // hand-edited wrapper" before overwriting it (issue #3967). A
-    // failed sidecar write only weakens user-modification protection
-    // on the next update — it does not invalidate the wrapper we just
-    // wrote — so we warn and continue.
+    // Sidecar records the wrapper's SHA-256 so a later `freenet update`
+    // can distinguish "Freenet's wrapper" from a hand-edited one
+    // before overwriting (#3967). A failed sidecar write only weakens
+    // future user-modification protection — warn and continue.
     let wrapper_hash_path = wrapper_path.with_extension("sh.hash");
     let wrapper_hash = super::update::wrapper_content_hash(&wrapper_content);
     if let Err(e) = fs::write(&wrapper_hash_path, format!("{wrapper_hash}\n")) {
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2819,16 +2819,14 @@ fn install_macos_service() -> Result<()> {
         .context("Failed to make wrapper script executable")?;
 
     // Sidecar records the wrapper's SHA-256 so a later `freenet update`
-    // can distinguish "Freenet's wrapper" from a hand-edited one
-    // before overwriting (#3967). A failed sidecar write only weakens
-    // future user-modification protection — warn and continue.
+    // can distinguish "Freenet's wrapper" from a hand-edited one before
+    // overwriting (#3967). A failed sidecar write only weakens future
+    // user-modification protection — warn and continue.
     let wrapper_hash_path = wrapper_path.with_extension("sh.hash");
     let wrapper_hash = super::update::wrapper_content_hash(&wrapper_content);
-    if let Err(e) = fs::write(&wrapper_hash_path, format!("{wrapper_hash}\n")) {
+    if let Err(e) = super::update::write_wrapper_hash_sidecar(&wrapper_hash_path, &wrapper_hash) {
         eprintln!(
-            "Warning: failed to write wrapper hash sidecar at {}: {}. \
-             User-modification detection on the next `freenet update` \
-             will fall back to treating the wrapper as ours.",
+            "Warning: failed to write wrapper hash sidecar at {}: {}.",
             wrapper_hash_path.display(),
             e
         );
--- a/crates/core/src/bin/commands/update.rs
+++ b/crates/core/src/bin/commands/update.rs
@@ -1259,6 +1259,20 @@ pub(super) fn wrapper_needs_regen(current_template: &str, on_disk: &str) -> bool
     current_template != on_disk
 }
 
+/// Pure helper: validate a wrapper-sidecar hash string. SHA-256 hex is
+/// exactly 64 lowercase ASCII hex chars; anything else means the sidecar
+/// is truncated, corrupted, or was never a hash. Treating malformed
+/// as "no sidecar" (rather than as permanent UserModified) lets the
+/// host recover from a partial prior write — without this, a single
+/// interrupted update could lock the auto-update path forever.
+/// Code-first-reviewer flagged on PR #4286.
+#[allow(dead_code)]
+pub(super) fn is_valid_wrapper_hash(s: &str) -> bool {
+    s.len() == 64
+        && s.bytes()
+            .all(|b| b.is_ascii_digit() || (b'a'..=b'f').contains(&b))
+}
+
 /// Pure helper: SHA-256 hex digest of `content`. Used to fingerprint the
 /// wrapper we last wrote so a later `freenet update` can distinguish
 /// "Freenet's previous wrapper" from "a user-edited wrapper" (#3967).
@@ -1275,6 +1289,95 @@ pub(super) fn wrapper_content_hash(content: &str) -> String {
             write!(s, "{b:02x}").expect("writing to String is infallible");
             s
         })
+}
+
+/// The decision the auto-update path takes for the wrapper script,
+/// derived purely from the current template, the on-disk wrapper, and
+/// the on-disk sidecar — no I/O. Extracting this lets non-macOS CI
+/// runners exercise the full decision matrix (skeptical-reviewer
+/// flagged on PR #4286 that the macOS-only end-to-end tests don't run
+/// in CI today).
+#[derive(Debug, PartialEq, Eq)]
+#[allow(dead_code)]
+pub(super) enum WrapperAction {
+    /// On-disk wrapper matches the current template byte-for-byte.
+    /// No rewrite needed. If `backfill_sidecar` is `Some`, the caller
+    /// should write that hash to the sidecar — covers the case where
+    /// the wrapper is up-to-date but the sidecar is missing or
+    /// malformed (self-heal so user-modification protection is in
+    /// force on the next update).
+    UpToDate { backfill_sidecar: Option<String> },
+    /// On-disk wrapper differs from the template and we have grounds
+    /// to overwrite it: no on-disk wrapper, or the sidecar shows the
+    /// wrapper is the one we wrote previously. The caller should
+    /// write `template` plus a sidecar containing `new_hash`.
+    WriteOurs { new_hash: String },
+    /// On-disk wrapper differs AND the sidecar records a different
+    /// hash, indicating the user (or another tool) edited it. The
+    /// caller should leave the wrapper alone and warn.
+    /// `on_disk_hash` is the hash of the current on-disk content,
+    /// provided so the warning message can include it.
+    UserModified { on_disk_hash: String },
+}
+
+/// Pure helper: decide what to do with the macOS launchd wrapper given
+/// the inputs the caller has gathered from disk. Cross-platform-compiled
+/// so every CI runner exercises this decision matrix — closes the gap
+/// the skeptical reviewer flagged on PR #4286 about the macOS-only E2E
+/// tests not running in CI.
+///
+/// - `current_template`: what `generate_wrapper_script(binary)` would
+///   produce now.
+/// - `on_disk_wrapper`: `Some(content)` if the wrapper exists and was
+///   read successfully (`None` if missing or unreadable).
+/// - `sidecar_raw`: `Some(raw)` if the sidecar exists and was read
+///   successfully (`None` if missing or unreadable). The raw bytes
+///   are trimmed and validated as a 64-char hex string by this
+///   helper, so the caller doesn't need to pre-process.
+#[allow(dead_code)]
+pub(super) fn decide_wrapper_action(
+    current_template: &str,
+    on_disk_wrapper: Option<&str>,
+    sidecar_raw: Option<&str>,
+) -> WrapperAction {
+    let stored_hash = sidecar_raw
+        .map(|s| s.trim())
+        .filter(|s| is_valid_wrapper_hash(s));
+    let template_hash = wrapper_content_hash(current_template);
+
+    let Some(on_disk) = on_disk_wrapper else {
+        return WrapperAction::WriteOurs {
+            new_hash: template_hash,
+        };
+    };
+    if !wrapper_needs_regen(current_template, on_disk) {
+        // Wrapper byte-matches template. Backfill the sidecar if it's
+        // missing or malformed, so a future hand-edit triggers the
+        // user-modification path instead of silently being clobbered.
+        let backfill_sidecar = match stored_hash {
+            Some(s) if s == template_hash => None,
+            _ => Some(template_hash),
+        };
+        return WrapperAction::UpToDate { backfill_sidecar };
+    }
+    let on_disk_hash = wrapper_content_hash(on_disk);
+    match stored_hash {
+        Some(s) if s != on_disk_hash => WrapperAction::UserModified { on_disk_hash },
+        _ => WrapperAction::WriteOurs {
+            new_hash: template_hash,
+        },
+    }
+}
+
+/// Write the wrapper hash sidecar in the canonical on-disk format
+/// (single line of 64 lowercase hex chars + trailing newline). Single
+/// source of truth so `install_macos_service` and
+/// `ensure_service_file_updated` cannot drift apart on the format
+/// — a divergence would loop-rewrite the wrapper on every update.
+/// Testing-reviewer flagged on PR #4286.
+#[allow(dead_code)]
+pub(super) fn write_wrapper_hash_sidecar(path: &Path, hash: &str) -> std::io::Result<()> {
+    fs::write(path, format!("{hash}\n"))
 }
 
 #[cfg(target_os = "macos")]
@@ -1294,40 +1397,21 @@ fn ensure_service_file_updated(binary_path: &Path, quiet: bool) -> Result<()> {
     let wrapper_path = home_dir.join(".local/bin/freenet-service-wrapper.sh");
     let wrapper_hash_path = wrapper_path.with_extension("sh.hash");
 
-    // Equality-check against the current template (see
-    // `wrapper_needs_regen` docs re: issue #3967).
     let new_wrapper_content = generate_wrapper_script(binary_path);
 
-    // Tri-state so we can warn (and skip) when the user has
-    // hand-edited their wrapper rather than silently overwriting it.
-    enum WrapperPlan {
-        UpToDate,
-        WriteOurs,
-        UserModified(String), // hex hash of the on-disk file
-    }
-
-    let plan = if wrapper_path.exists() {
-        let on_disk = fs::read_to_string(&wrapper_path).context("Failed to read wrapper script")?;
-        if !wrapper_needs_regen(&new_wrapper_content, &on_disk) {
-            WrapperPlan::UpToDate
-        } else {
-            let on_disk_hash = wrapper_content_hash(&on_disk);
-            // Sidecar absent: pre-#3967 migration case, treat as ours.
-            // Sidecar matches on-disk hash: ours, safe to overwrite.
-            // Sidecar mismatches: user (or some tool) edited the
-            // wrapper since we wrote it — back off.
-            match fs::read_to_string(&wrapper_hash_path) {
-                Ok(stored) if stored.trim() != on_disk_hash => {
-                    WrapperPlan::UserModified(on_disk_hash)
-                }
-                _ => WrapperPlan::WriteOurs,
-            }
-        }
-    } else {
-        WrapperPlan::WriteOurs
-    };
-
-    let wrapper_needs_update = matches!(plan, WrapperPlan::WriteOurs);
+    // A non-UTF-8 wrapper on disk (very unusual for a shell script,
+    // but possible if corruption snuck in) is treated the same as a
+    // missing wrapper: we'll overwrite. Otherwise the whole update
+    // path would abort with a misleading "Failed to read wrapper
+    // script" error on an obviously-recoverable state.
+    // Skeptical-reviewer flagged on PR #4286.
+    let on_disk_wrapper = fs::read_to_string(&wrapper_path).ok();
+    let sidecar_raw = fs::read_to_string(&wrapper_hash_path).ok();
+    let action = decide_wrapper_action(
+        &new_wrapper_content,
+        on_disk_wrapper.as_deref(),
+        sidecar_raw.as_deref(),
+    );
 
     // Also force regeneration of the plist if it still points
     // StandardOutPath/StandardErrorPath at the unrotated freenet.log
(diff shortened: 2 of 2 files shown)


sanity:
## Testing Review: PR #4286

### Coverage Summary
- Direct changes: covered
- Downstream impact: partial — install path's sidecar contract is not pinned

### Mental simulation of FAIL-on-prior / PASS-on-new

The end-to-end `ensure_service_file_updated_rewrites_pre_pkill_wrapper_and_records_hash` test is the real regression test. The stale string `"#!/bin/bash\n# EXIT_CODE=$?\n# freenet update\n# (no pkill block!)\n"` would, under the prior marker check, evaluate to `wrapper_needs_update = !contains("EXIT_CODE=$?") || !contains("freenet update") || contains(".../freenet.log") = false || false || false = false` — old code skips the rewrite, new test asserts a rewrite happened, so the test correctly fails on prior code and passes on new. Good.

The four pure tests are white-box tests of the new `wrapper_needs_regen` helper, which didn't exist in the prior code. They lock in the new contract going forward but wouldn't have caught the original bug on their own — the end-to-end test is doing that work. That's fine; just worth being explicit about.

### Test Gaps Found

#### Important (should have)

1. **install -> update is a no-op (closes the "why didn't CI catch this" gap completely).** `freshly_generated_wrapper_does_not_trigger_regen_loop` proves the generator is deterministic, but no test exercises the boundary that the original incident actually crossed: `install_macos_service()` writes a wrapper + sidecar, then a subsequent `ensure_service_file_updated()` must observe `WrapperPlan::UpToDate` (no rewrite, no churn on the `.bak` file). If the install path and the update path ever disagreed on hash format (trailing newline, encoding, trim semantics), the update path would loop-rewrite every time. A test like `install_then_ensure_is_noop` would pin this exact contract.

2. **Sidecar present + hash matches on-disk + template drifted.** This is the "ours, safe to overwrite" branch of the match. The two macOS E2E tests cover (no sidecar) and (sidecar mismatches). The third branch — sidecar matches on-disk hash but template has drifted (e.g., a future code change to `generate_wrapper_script`) — isn't covered. The behavior should be: rewrite + update sidecar. Worth a test so a future refactor of the match doesn't silently invert the semantics.

3. **Sidecar with surrounding whitespace.** `.trim()` on read implies the test should pin that `"  abc\n\n"` and `"abc"` are treated identically. `wrapper_content_hash_is_stable_and_distinct` pins the hash, but not the trim behavior on the comparison side. If someone refactored `stored.trim() != on_disk_hash` to `stored != on_disk_hash` it would still pass every existing test on macOS but break for users whose editor added trailing whitespace to the sidecar.

#### Nice to Have

4. **Sidecar write failure path.** The `if let Err(e) = fs::write(&wrapper_hash_path, ...)` branch in both `install_macos_service` and `ensure_service_file_updated` is unreachable in the current tests. Hard to exercise portably (would need a read-only directory or fs::write mock). Not blocking, but the comment claims "failed sidecar write only weakens future protection — wrapper is still valid" — that's an assertion no test pins.

5. **Plist-needs-regen + wrapper UpToDate combination.** The `wrapper_needs_update` is now derived from a tri-state plan, and `plist_needs_update` is independent. The interaction (wrapper up-to-date, plist needs regen) isn't covered. Pre-existing gap, not introduced by this PR.

### Determinism / flakiness

- `HOME_ENV_LOCK` is sufficient *within this crate*: `grep -rn 'set_var.*HOME' crates/core/src` returns only the two call sites inside the lock. Cargo runs each test binary in its own process, so cross-crate HOME mutation isn't a concern.
- The broader `std::env::set_var` thread-safety concern (Rust 2024 marks it `unsafe`) is real but not actionable here without rearchitecting `ensure_service_file_updated` to accept an injected home path. The `ScopedHome` RAII guard with poison-tolerant lock acquisition is the right pragmatic choice.
- One minor wart: `ScopedHome::new` takes the lock, then calls `std::env::var_os("HOME")` and `set_var`. If another thread in the same process reads `HOME` (e.g., `dirs::home_dir()` called from a background task), there's a race the lock doesn't protect against. Unlikely in this test harness; flagging for completeness.

### Pure-function coverage / hash semantics

- SHA-256 collision risk is not realistic.
- The `.trim()` comparison is correct *given* both writers use `format!("{hash}\n")`. That invariant is duplicated across `install_macos_service` (service.rs:2827) and `ensure_service_file_updated` (update.rs:1407). A const or helper (`fn write_wrapper_hash_sidecar(path, hash) -> io::Result<()>`) would make this single-source-of-truth and let a single unit test pin the on-disk format.

### CI Considerations

- The four cross-platform pure tests are cheap. No runtime concern.
- The two macOS E2E tests are tempdir-only, no subprocess, no network. Should run in well under 1s each. No flakiness risk beyond the env-var lock concern above.
- `freshly_generated_wrapper_does_not_trigger_regen_loop` calls the real generator twice and compares — also trivial cost.

### Verdict

Solid suite for the direct fix. The two end-to-end tests are good regression tests for the originally-reported failure mode. The main gap is that the install -> update no-op contract is not pinned, so a future divergence in hash format between the two write sites would still slip past CI. Adding gap #1 above would fully close the "why didn't CI catch this" question the PR description asks.

[AI-assisted - Claude]

sanity (commented):
## Code-first review (read code before description)

Reviewed the diff and surrounding code at `crates/core/src/bin/commands/{update,service}.rs` before reading the PR body. The high-level approach (byte-equality staleness check + SHA-256 sidecar for user-edit detection) is sound and the tests are unusually thorough for this kind of host-side migration code. A few real concerns and one documentation gap worth addressing before merge.

### 1. `UpToDate` arm does not backfill a missing sidecar

If a previous sidecar write fails (or the process crashes between the wrapper write and the sidecar write), the wrapper on disk matches the current template but no sidecar exists. The next call to `ensure_service_file_updated` evaluates `wrapper_needs_regen` -> false -> `WrapperPlan::UpToDate`, returns without touching the sidecar. The host then stays in the "no sidecar => migration => treat as ours" state indefinitely. The first time the user makes a custom edit AND the next template change ships, their edits will be silently clobbered (because sidecar is still absent -> `WriteOurs`).

Fix: in the `UpToDate` arm, if `wrapper_hash_path` does not exist (or doesn't match `wrapper_content_hash(&on_disk)`), write the current-template hash to it. Cost: one extra fs::metadata + maybe one write per update. Reward: closes the only window in which the PR's user-modification protection silently fails.

### 2. Unreadable sidecar (permissions / IO error) silently degrades to `WriteOurs`

```rust
match fs::read_to_string(&wrapper_hash_path) {
    Ok(stored) if stored.trim() != on_disk_hash => WrapperPlan::UserModified(...),
    _ => WrapperPlan::WriteOurs,
}
```
The catch-all `_` collapses three cases: sidecar absent (intentional migration), sidecar IO error (transient, ambiguous), and sidecar present-and-matches (genuinely ours). The first and third are correct; the second is not. A permission-denied or transient-IO read should arguably back off rather than overwrite. Consider matching `Err(e) if e.kind() == ErrorKind::NotFound => WriteOurs` and treating other errors as "be conservative, skip and warn", same as the plist-read fallback below.

### 3. Empty / corrupt / truncated sidecar permanently blocks auto-updates

If a previous sidecar write was interrupted and left zero bytes (or any non-hex content), `stored.trim() != on_disk_hash` is true -> `UserModified` forever, until the user manually deletes the sidecar. The warning message does tell them to delete it, but: (a) most hosts running `freenet update` non-interactively won't read stderr, and (b) this state is indistinguishable from a real user edit in the warning. Consider validating the stored hash is 64 hex chars before treating mismatch as "user modified" — anything else means "sidecar is junk, treat as migration".

### 4. Pre-PR install with hand-edited wrapper loses edits on first auto-update

This is the explicit trade-off in the PR description ("missing sidecar = treat as ours"), but the cost asymmetry is worth flagging: every host that hand-edited their wrapper before this PR landed will have it overwritten on the first auto-update. The `.sh.bak` softens the recovery story but doesn't prevent the loss. If the issue-#3967 hosts are the only known offenders, the trade is fine; if hand-editing is even moderately common, the PR could mitigate by checking for a sentinel comment Freenet writes into every wrapper (e.g. `# Freenet auto-generated wrapper - do not edit`) and treating "no sentinel and no sidecar" as user-modified rather than ours. Not blocking, but worth a thought.

### 5. `freenet uninstall` does not clean up the sidecar or wrapper

`stop_and_remove_service` (macOS) only removes the plist. `purge_data_dirs` only touches `~/Library/Application Support/Freenet`, `~/.cache/freenet`, etc. So `~/.local/bin/freenet-service-wrapper.sh{,.hash,.bak}` are all left behind on uninstall. This was pre-existing behavior for the wrapper; the PR inherits and extends it. Worth either filing a follow-up or adding a brief comment in `uninstall_service` enumerating what is intentionally not removed.

### 6. Install-path sidecar warning ignores quiet flag

`install_macos_service` in `service.rs` writes the warning unconditionally with `eprintln!`, while `ensure_service_file_updated` in `update.rs` correctly gates with `if !quiet`. Minor inconsistency. `service install` is always user-initiated so it probably doesn't matter, but worth aligning.

### 7. Documentation gap: when does the user need to know about `.sh.hash`?

The warning message points the user at deleting the sidecar or reinstalling. Worth a sentence in the install command output (or in `docs/`) that the sidecar exists, what it's for, and that hand-editing the wrapper is supported as long as the sidecar is removed first. Otherwise a user who reads the warning has to read the PR to understand it.

### Discrepancies with the PR description

None significant. The description accurately describes the byte-equality check, the sidecar, the migration story, and the test coverage. The "missing sidecar = treat as ours" choice is explicitly called out. The above points are concerns the description acknowledges but doesn't fully address — chiefly that the `UpToDate` path doesn't backfill, which leaves a small but real hole in the user-modification protection the PR is selling.

### Severity summary

- (1) `UpToDate` arm not backfilling sidecar — **important**, real correctness gap in the protection mechanism this PR adds.
- (2) IO-error sidecar handling — **nit**, minor robustness improvement.
- (3) Corrupt sidecar bricks updates — **important** if it can happen in the field, **nit** otherwise; cheap to fix.
- (4) Pre-existing hand-edited wrappers — **nit** (explicit trade-off), worth a follow-up issue.
- (5) Uninstall doesn't clean sidecar — **nit**, follow-up.
- (6) Quiet-flag asymmetry — **nit**.
- (7) Docs — **nit**.

No blockers. (1) and (3) are worth a small follow-up commit in this PR; the rest can be follow-ups.

[AI-assisted - Claude]

sanity (commented):
## Big Picture Review: PR #4286

### Goal alignment
- **Stated PR goal**: replace fragile marker-based wrapper staleness check with byte-equality + hash sidecar.
- **Linked issue (#3967) goal**: 6-item resilience checklist. PR addresses 2.5: re-template wrapper on update (✓), regression test (✓), and the "wrapper text matches fresh install after one `freenet update`" criterion (✓ for macOS only).
- **Implementation matches the stated PR goal**: yes. The end-to-end test `ensure_service_file_updated_rewrites_pre_pkill_wrapper_and_records_hash` precisely reproduces the two-host incident (Jan-2026 wrapper, no sidecar, no pkill) and asserts the rewrite.

### Scope discipline (the hard question)
The PR closes #3967 but addresses only items 1 (partial — macOS-only) and 5 (regression test) of the 6 acceptance criteria. Items 3 (stuck-wrapper notification), 4 (homepage version banner), and item 5 from the issue narrative (service-stop on `freenet update`) are not addressed and no follow-up issues exist. Given the user's `finish-the-fix` rule and the fact that the user-visible symptom (silent stale-asset serving for 19 days) has multiple contributing causes, "Closes #3967" is overclaiming. **Recommendation**: either downgrade to "Partial fix for #3967" and leave the issue open, OR file three follow-up issues before merge and reference them in the PR body.

### Symmetric bug in Linux path (NOT addressed)
`systemd_unit_needs_regen` at `crates/core/src/bin/commands/update.rs:1120` uses the exact marker-presence anti-pattern (`ExecStopPost=` + `SuccessExitStatus=42` + `RestartPreventExitStatus=43`) that this PR is fixing on macOS. The issue narrative explicitly says "Same on Linux for the systemd unit". A future systemd-unit template change (e.g., adding a new `ExecStartPre=` directive) will silently fail to deploy to existing Linux installs for the same reason this PR is being written. This is a clear case of `systematic-debugging.md`'s "search for similar bugs" guidance. Recommend applying the same equality+hash treatment to the systemd unit in this PR, OR filing it as a tracked follow-up.

### Removed code concerns
None. 18 pre-existing tests → 25 post-PR tests (net +7). No `#[ignore]`, no weakened assertions, no commented code. No reverts of prior fixes.

### Anti-patterns detected
None of the listed CI-chasing patterns. The PR is the opposite: it strengthens a self-validating predicate (`freshly_generated_wrapper_does_not_trigger_regen_loop` would catch any future nondeterminism in the template).

Minor: `ScopedHome` takes a process-wide mutex but doesn't protect against `dirs::home_dir()` reads on other threads in the same process (already flagged in testing review).

### AI Instructions / automation rules
- **`.claude/rules/bug-prevention-patterns.md` (in-repo)**: currently has exactly one entry (`Command::spawn` after `FreeConsole()`). The pattern this PR fixes — "artifact-staleness predicates that grep for markers in deployed artifacts" — has now caused at least two production incidents (the issue cites #3934 and #3928 as related, plus the two hosts in #3967). It meets the bar (multi-incident, same shape, with a known mechanical avoidance) for an entry in the table. Suggested wording: "Artifact-staleness checks that grep for marker substrings in deployed wrappers/plists/units silently miss any future template change whose marker isn't checked. Use byte-equality against the current generator output + a hash sidecar for user-modification protection. See `wrapper_needs_regen` / `wrapper_content_hash` in `update.rs` for the pattern; #3967 incident report."
- **`.claude/rules/deployment.md`**: the issue narrative explicitly proposes adding "Service artifacts (wrapper, plist, unit) MUST be re-templated on update" as a rule. Not done. Worth a one-paragraph trigger-action entry under "WHEN implementing auto-update".
- Agents/skills/hooks: no agent-definition or skill file references the marker check, so no stale instructions need updating.

### Related work
- **Open PRs**: no overlap with PRs #4285, #4220, #4204, #4126 (different surfaces).
- **Recent merges**: #4251 (log path migration) is the cousin that introduced the second marker (`Library/Logs/freenet/freenet.log` substring); same anti-pattern, same file, same author. PR #3314 (Feb 2026) added the pkill block that this PR is now retroactively delivering to existing installs. The PR is the natural completion of that chain.

### Strategic assessment
The fix itself is well-targeted and correct. The framing as "Closes #3967" is the only meaningful concern: it ships ~40% of the resilience checklist and leaves the rest unsurfaced, which is exactly the scenario `finish-the-fix.md` warns against. The Linux systemd parallel is the most concrete gap because it's the same bug class in the same file, fixable in the same PR with high reuse of the new helpers.

### Recommendations
1. **Required before merge**: either retitle to "Partial fix for #3967 (macOS wrapper)" and leave the issue open, OR file follow-up issues for items 3, 4, 5 (stuck-wrapper notification, homepage version banner, service-stop on update) and reference them from the PR body. Pick one.
2. **Strongly recommended**: apply the same equality+hash treatment to `systemd_unit_needs_regen` (Linux) in this PR. The helpers are already cross-platform-compiled; the marginal cost is small and the alternative is filing a tracked follow-up that history says will sit for weeks.
3. **Recommended**: add a bug-prevention-patterns.md entry for "marker-substring artifact-staleness checks" with the cross-reference to the new helpers. Two incidents qualify.
4. **Nice to have**: extract `format!("{hash}\n")` into a single `write_wrapper_hash_sidecar` helper so the on-disk format invariant is single-source-of-truth (echoes testing review's point).
5. **Nice to have**: add the `install -> ensure_is_noop` test that the testing review identified — closes the "why didn't CI catch this" question completely.

[AI-assisted - Claude]

sanity (commented):
## Skeptical Review: PR #4286

Spent time tracing through the four scenarios this PR claims to fix and the most likely regression vectors. The core fix (content-equality replacing the marker grep) is sound and addresses the underlying #3967 root cause. A few concerns worth surfacing before merge.

### High Severity

**1. New macOS-only tests are not exercised by any CI job.**

The PR description says "macOS-only tests will run on CI runners," but I checked `.github/workflows/` and:
- `ci.yml` runs `cargo test` only on `ubuntu-latest`, `ubicloud-*`, and `windows-latest` — no macOS test job exists.
- `cross-compile.yml`'s macOS job runs `cargo build --release`, not `cargo build --tests`, so the `#[cfg(target_os = "macos")] #[test]` functions are never compiled.
- `macos-dmg-test.yml` and `macos-dmg-swap-e2e.yml` exercise the DMG pipeline, not `cargo test`.

That means `ensure_service_file_updated_rewrites_pre_pkill_wrapper_and_records_hash`, `ensure_service_file_updated_preserves_user_modified_wrapper`, `ensure_service_file_updated_is_noop_after_fresh_install`, and `freshly_generated_wrapper_does_not_trigger_regen_loop` are dead test code as far as CI is concerned. They were presumably run locally on macOS, but the next agent to touch this code path has no automated signal. Given this is the second production incident in two weeks from a marker-based check that nothing was guarding, the lack of CI coverage on the new equality check is the same shape of risk the PR is trying to address. Consider either adding a macOS `cargo test -p freenet` job (even if it only runs the wrapper-related tests) or rewriting the I/O paths so the decision logic is cross-platform and exercisable from the Linux CI tests already added.

### Medium Severity

**2. `install_macos_service` overwrites a user-modified wrapper without consulting the sidecar.**

The update path scrupulously refuses to clobber a hand-edited wrapper. The install path (`service.rs:2796` `install_macos_service`) unconditionally `fs::write`s the wrapper and sidecar — there is no `WrapperPlan` check there. Scenarios this breaks:

- User runs `freenet service install` to repair a config, after having edited the wrapper. Their edits vanish silently.
- User runs `freenet service install` on top of an existing install (intentional re-install). Same.
- After `freenet service uninstall` + reinstall: the uninstall path on macOS removes the plist only (`service.rs:3036`) and leaves the wrapper + sidecar on disk; the subsequent install overwrites both. Probably what the user wants, but inconsistent with the update path's policy.

If "never silently clobber a hand-edited wrapper" is the invariant, install should honor it. If install is allowed to clobber (defensible — "I want a fresh setup"), the PR description should make that explicit so it doesn't read as a guarantee the code doesn't enforce.

**3. `freenet service uninstall` leaves the sidecar behind permanently.**

The macOS uninstall path removes only the plist. The wrapper script AND the new `.sh.hash` sidecar persist in `~/.local/bin/`. For a typical user this is harmless (the next install overwrites both), but consider a user who uninstalls Freenet and then manually edits the leftover wrapper. On reinstall, install clobbers; if they then `freenet update`, the update path correctly detects the new install's sidecar match. Edge case is benign, but the asymmetry "install writes a sidecar, uninstall doesn't remove it" is something the next agent will trip on. Worth at least a comment in `uninstall_service` explaining why the wrapper + sidecar are left in place (e.g. "preserved for `keep_data`" or "harmless leftover, next install overwrites").

**4. Migration policy silently destroys hand-edited wrappers copied from another machine.**

The PR description acknowledges that `missing sidecar = treat as ours`. The user-visible consequence: a sysadmin who hand-edited their wrapper, then `rsync`ed `~/.local/bin/freenet-service-wrapper.sh` to a fresh machine (without copying a sidecar that never existed) will have their wrapper destroyed on the first `freenet update`. The `.sh.bak` backup is the only recovery. This is a deliberate trade-off, but I want to flag it explicitly — the symmetric choice ("missing sidecar = treat as user-modified") would block the auto-fix this PR is trying to enable. There's no clean answer; just make sure the PR description's migration note doesn't end up trimmed away in subsequent narration passes — that note is the only thing protecting the next agent from being surprised by a complaint here.

### Low Severity

**5. `fs::read_to_string` on a binary/non-UTF-8 wrapper aborts the update path.**

Line 1323: if a user-modified wrapper contains a non-UTF-8 byte (very unusual for a shell script, but possible if a multi-byte character got mangled), the `?` returns Err, the whole `freenet update` exits non-zero, launchd retries. The wrapper itself isn't touched, so the system stays in its previous state — recoverable, but the user sees auto-update failures with a misleading "Failed to read wrapper script" error. Consider treating a non-UTF-8 wrapper the same as a missing one (`WrapperPlan::WriteOurs`)? Or at least log a more actionable message.

**6. Trim-then-validate is order-sensitive on whitespace.**

`stored.trim() != on_disk_hash` plus `is_valid_wrapper_hash` (must be 64 hex chars) means a sidecar containing `  <64hex>  ` (any surrounding whitespace, including \r\n on Windows-encoded files) is accepted as ours. Probably intentional — robust to text-mode line endings. But also: a sidecar containing `<64hex>\nmalicious garbage on line 2` would fail `is_valid_wrapper_hash` (length > 64 after trim) and fall through to `WrapperPlan::WriteOurs`. That's correct, but worth a one-line test pinning the behavior so a future "let's parse first line only" simplification doesn't accidentally accept multi-line sidecars.

### Questions for Author

- Can you confirm the new macOS tests are actually executed somewhere automated? If they're only local, that's worth saying in the PR so the next agent doesn't trust them as a CI gate.
- Is the install-path clobber of user edits intentional? If yes, recommend a short comment + a note in the PR description.

### What I Verified

- `set_var`/`remove_var` ARE wrapped in `unsafe` blocks with SAFETY comments (the diff I reviewed initially didn't show the `unsafe {` line). Rust 2024 edition compliance: fine.
- SHA-256 is overkill for accidental-edit detection but not wrong, and `sha2` is already a workspace dep.
- The wrapper-rewrite-while-running concern is moot: bash reads the script into memory at parse, and the existing code path (pre-PR) already did `fs::write` on the same path. Not introduced by this PR.
- The symlink-attack vector at `~/.local/bin/*.sh.hash` requires write access to the user's HOME directory, which means the attacker already owns the user — not a real attack surface.
- TOCTOU between `wrapper_path.exists()` and `fs::read_to_string` is bounded by the user themselves (only they can delete files in their HOME), and on error the function returns Err — the wrapper is not modified, system stays in its previous state. Acceptable.
- The `is_valid_wrapper_hash` guard correctly prevents permanent UserModified state from a malformed sidecar. The self-heal backfill in the `UpToDate` branch (lines 1358-1375) is a nice touch.

Overall: the core fix is correct and well-motivated. The CI-coverage gap on the new tests is the only thing I'd genuinely block on.

[AI-assisted - Claude]

sanity:
## Review-round-1 response

Thanks for the thorough reviews. Pushed `26316d9f` addressing the
findings:

### Blockers / Important — fixed in this PR

- **Codex P2 (Rust 2024 `unsafe`)**: wrapped HOME mutation in
  `unsafe` blocks with SAFETY comments. Without this, macOS test
  compile would have failed on CI.
- **Skeptical H1 (macOS-only tests don't run in CI)**: extracted
  the decision logic into a cross-platform `decide_wrapper_action`
  pure helper with 9 new tests that run on every CI runner. The
  macOS E2E tests remain as integration coverage but are no longer
  the only line of defense.
- **Code-first #1 (UpToDate not backfilling sidecar)**: the
  decision helper now returns `UpToDate { backfill_sidecar: Option<String> }`;
  the caller writes the sidecar when set. Closes the
  "lost-sidecar leaves user-mod protection permanently off" window.
- **Code-first #3 (corrupt sidecar bricks updates)**: added
  `is_valid_wrapper_hash` — anything that isn't 64 lowercase hex
  chars is treated as missing, so a partial prior write can't lock
  the host into permanent UserModified.
- **Testing-reviewer #1 (install->update no-op)**: added
  `ensure_service_file_updated_is_noop_after_fresh_install` test +
  pinned the same contract at the pure-helper level
  (`decide_wrapper_action_up_to_date_with_matching_sidecar_is_noop`).
- **Testing-reviewer (DRY sidecar format)**: extracted
  `write_wrapper_hash_sidecar` — single source of truth between
  `install_macos_service` and `ensure_service_file_updated`.
- **Skeptical L5 (non-UTF-8 wrapper)**: now treats an unreadable
  wrapper as "missing" and writes ours, instead of aborting the
  whole update with a confusing error.

### Filed as follow-ups (out of scope for this PR)

- **Big-picture / Skeptical M3 / Linux symmetric bug**: filed as
  #4287. `systemd_unit_needs_regen` has the same anti-pattern; the
  helpers here are cross-platform and ready for a Linux caller.
- **Item 3 of #3967**: filed as #4288 — stuck-wrapper notification.
- **Item 4 of #3967**: filed as #4289 — homepage version-mismatch
  banner.
- **Skeptical M3 / Code-first #5 (uninstall doesn't clean wrapper
  artifacts)**: filed as #4290.

### Intentionally not addressed

- **Skeptical M2 (install path clobbers user edits)**: `freenet
  service install` is a deliberate user action; treating it as
  "install bundled setup, overwriting any local edits" is the
  expected UX. The user-modification protection is for the *auto*-update
  path, where the user didn't initiate the action. Documented in
  the PR body's migration section.
- **Code-first #4 (pre-PR hand-edited wrappers lose edits on
  first auto-update)**: this is the explicit migration trade-off
  the description calls out. The `.sh.bak` is the recovery path.
  Symmetric "missing sidecar = treat as user-modified" would
  defeat the auto-fix this PR exists to enable.
- **Code-first #2 (sidecar IO-error vs sidecar-absent)**: the
  helper currently collapses both into "no sidecar => treat as
  ours". Distinguishing them would require differentiating between
  transient IO errors and ENOENT, which is fragile across
  filesystems. Filed mentally as a future hardening if reports surface.

PR body updated to reflect the new scope, the follow-ups, and the
self-heal behavior.

[AI-assisted - Claude]

Issue #4271, opened 2026-05-26
gateway-update.yml: double-stop kills freshly-started gateway during release rollout


sanity:
## Incident

During the 0.2.65 rollout today (2026-05-26), nova's gateway service went down because the gateway-update workflow issued two stop commands close enough together that the second one killed the freshly-started 0.2.65 process. Service stayed `inactive` until manual `sudo systemctl start freenet-gateway`.

## Sequence (nova journal)

```
May 26 15:41:52  Stopping freenet-gateway.service ...
May 26 15:41:52  init-gateway.sh[1235263]: CRITICAL: Network event listener exited: Graceful shutdown
May 26 15:43:22  freenet-gateway.service: State 'final-sigterm' timed out. Killing.
May 26 15:43:22  Killing process 1235263 (freenet) with signal SIGKILL.
May 26 15:43:23  Started freenet-gateway.service - Freenet Gateway Service.
May 26 15:43:23  init-gateway.sh[1956098]: warning: some trace filter directives would enable traces that are disabled statically
                  (... 0.2.65 process starting ...)
May 26 15:43:24  Stopping freenet-gateway.service ...           ← second stop arrives 1s after start
May 26 15:43:24  init-gateway.sh[1956098]: CRITICAL: Network event listener exited: Graceful shutdown
May 26 15:43:26  Error: Graceful shutdown
May 26 15:43:26  freenet-gateway.service: Deactivated successfully.
                  (... service stays inactive, no restart ...)
```

## Impact

- Nova was down for ~3 minutes (15:43 → ~15:46 CDT manual restart).
- This is a production gateway. With auto-update enabled, the same race could recur on every release.
- Vega did NOT exhibit this — the rollout to vega succeeded cleanly. Suggests timing/state-dependent.

## Likely root cause(s) to investigate

1. The release-agent on nova received two HMAC POSTs for the same release. Possible duplicate dispatch from `gateway-update.yml`.
2. The release-agent doesn't deduplicate or coalesce update commands. If a second arrives while the first is still draining (note: the old process took 90s to die — past the final-sigterm timeout), the second `systemctl restart` queues up.
3. Old process's 90s shutdown grace window collided with the second update command.

## What to check

- `gateway-update.yml` — does it POST the same gateway twice (e.g., once per matrix entry, once as a fallback)?
- The release-agent on nova — does it have any dedup logic on incoming update requests?
[...]

Issue #4001, opened 2026-05-01
fix(put): drive_retry_loop fires retries while streaming PUT still in flight, causing version-conflict failures for large payloads


sanity:
## Problem

`drive_retry_loop` in the client-initiated PUT driver (`crates/core/src/operations/put/op_ctx_task.rs`) gates each attempt on a fixed `OPERATION_TTL = 60s` (`crates/core/src/config.rs:44`). For streaming PUTs whose end-to-end network time exceeds 60s — readily reproducible with the freenet.org website contract (~2.4 MB compressed) on the production gateway — the retry loop fires its 3 internal retries (`MAX_RETRIES = 3` at `op_ctx_task.rs:414`) **while the original streaming op is still in flight**.

Those retries reuse the same `WrappedState` (and therefore the same contract version, since `fdev::generate_version()` is called once per CLI invocation: `crates/fdev/src/website.rs:165-172`). The website contract enforces `new_version > current_version`, and the in-flight original has typically already started applying state, so every retry fast-fails:

```
ERROR ... freenet::operations::put: Failed to update contract value
  contract=E4m5WbaC4cdbpDjL82WfYUnrM9iMnfaaN8Tsn7UHPMjZ
  error=update error ... reason: execution error: invalid contract update,
  reason: New state version 29627709 must be higher than current version 29627709
```

The retry loop returns `RetryLoopOutcome::Exhausted`, the WS client receives `"put failed after 4 attempts (last error: failed notifying, channel closed)"`, and fdev exits non-zero. **The original streaming PUT then completes successfully a few seconds later** — too late to affect the already-returned failure.

## Reproduction

- The `Publish to Freenet` workflow on `freenet/web` (https://github.com/freenet/web/actions/workflows/publish-freenet.yml) fails on every push to `main`. Recent failures: run `25228448024` (2026-05-01), run `25174625540` (2026-04-30).
- Each run shows three identical failures:
  ```
  Error: Failed to receive response: client error: error while executing operation in the network:
  put failed after 4 attempts (last error: failed notifying, channel closed)
  ```

## Evidence (production gateway log, 2026-05-01)

`/home/freenet/.local/state/freenet/freenet.2026-05-01-19.log`, around the WF attempt 2 window (tx `01KQJF5W0GY6QSJD33VEQ7JD01`):

```
19:10:19.665  INFO ... PUT request, is_originator=true ...
19:10:19.741  INFO ... PUT request using operations-level streaming, payload_size=2460242
[...]

PR #4291, opened 2026-05-27
fix(node,put): stabilize freenet-git mirror via shutdown drain + streaming retry cap

sanity:
## Problem

`Mirror to Freenet` runs on `freenet-core` have been failing repeatedly, spamming `#freenet-dev:matrix.org` with `🚨 Mirror to Freenet run did not complete successfully…` alerts (8 failures in the last ~36 hours). Investigation found two distinct failure modes:

1. **Restart-during-PUT.** Release-driven gateway auto-update (`systemctl stop freenet-gateway.service` from `deploy-local-gateway.sh`) lands while a `freenet-git` mirror push is mid-stream. The gateway's existing SIGTERM path tears down peer connections immediately, dropping the client's WebSocket and aborting the in-flight PUT. The 2026-05-27 02:16:23 UTC mirror failure correlates to v0.2.66's gateway restart at the exact same second.

2. **Budget mismatch (freenet-git#53).** The gateway accepts a client PUT, runs `MAX_RETRIES=3` × `streaming_aware_attempt_timeout` (capped at 600s) = up to 30 minutes worst-case wall-clock. The `freenet-git` client times out at 180s per attempt and gives up — but the gateway is still working and eventually publishes a terminal `PutResponse(Err)` minutes later. Looks like a silent gateway timeout; really a budget mismatch under transport congestion (`Channel overflow` WARNs every 10s on `nova` were stretching individual attempts to their full 116s budget).

## Approach

Two surgical changes; both fit the bug they address and neither broadens behaviour beyond what the failure mode requires.

### 1. Graceful drain on gateway shutdown (commit 5660529)

`ShutdownHandle::shutdown` now waits up to `Config::shutdown_drain_secs` (default 30s) for in-flight client-originated drivers to finish before triggering `NodeEvent::Disconnect`. Implementation: a per-driver RAII counter (`ClientOpGuard`) installed at every `start_client_{put,get,update,subscribe}` spawn site and held inside the spawned future so every termination path — happy, error, panic unwind — decrements the counter exactly once. `systemd`'s `TimeoutStopSec` bumps from 15s → 45s (30s drain + 15s peer-teardown headroom). CONNECT is skipped: its `start_client_connect` runs inline rather than spawning a driver, and CONNECT requests are not client-API operations whose interruption causes user-visible failures.

### 2. Cap streaming PUT retries at 1 (commit a2d7014)

Two named caps: `MAX_RETRIES_NON_STREAMING = 3` (legacy, preserves k_closest fan-out coverage for small payloads) and `MAX_RETRIES_STREAMING = 1`. Selection at driver construction via `should_use_streaming(threshold, payload_size_estimate)` — same predicate `streaming_aware_attempt_timeout` already uses, so the streaming-eligibility decision has a single source of truth. Streaming-PUT consumers in production today (freenet-git mirror packs, River room contracts) are content-addressed or version-monotonic, so the client's outer retry handles peer-rotation correctness without depending on the in-process retry loop.

## Why not the other levers we considered

- **Client-side reconnect in `freenet-git`** addresses (1) but not (2); the drain is cleaner because it works for any WS-API client, not just freenet-git.
- **Bumping the freenet-git per-attempt timeout** addresses (2) but doesn't fix the root cause (a 30-min gateway budget will eventually exceed any client patience). Better to fix the budget at the source.
- **Investigating gateway transport congestion** is the real root cause behind (2), but it's a separate, much larger investigation (LEDBAT/BBR vs fixedrate, bandwidth limits, etc.). The retry cap is the right reliability bound regardless.

## Testing

New tests (all pass; full lib suite `2753 passed; 0 failed; 14 ignored`):

- `node::tests::shutdown_drain::{shutdown_with_zero_ops_returns_immediately, shutdown_waits_then_proceeds_on_timeout, shutdown_proceeds_as_soon_as_counter_clears, drain_disabled_skips_wait_even_with_ops_in_flight}` — four cases against a fake counter + channel covering: zero-op fast path, drain-timeout proceeds anyway, drain returns as soon as counter clears, `drain_timeout=0` bypass for tests.

- `operations::{put,get,update,subscribe}::op_ctx_task::tests::start_client_*_acquires_inflight_guard_before_spawn` — four source-grep pins, one per client driver, asserting `op_manager.client_op_guard()` is called before `GlobalExecutor::spawn(` and the guard is moved into the spawned future via `let _inflight_guard = inflight_guard;`. A future change that drops a guard site fails CI with a specific shutdown-drain reason.

- `operations::put::op_ctx_task::tests::streaming_put_retry_budget_does_not_exceed_client_patience` — asserts `MAX_RETRIES_STREAMING × STREAMING_ATTEMPT_TIMEOUT_CAP <= 600s` AND pins `MAX_RETRIES_STREAMING == 1` with a comment pointing back at the freenet-git#53 analysis. Any future bump must engage the budget math.

- `operations::put::op_ctx_task::tests::drive_client_put_inner_dispatches_streaming_cap_on_should_use_streaming` — source-grep pin that the cap selection references both `MAX_RETRIES_STREAMING` and `MAX_RETRIES_NON_STREAMING` gated on `should_use_streaming(`. A flat constant for all PUTs would silently re-open freenet-git#53; this pin catches it.

- `operations::put::op_ctx_task::tests::max_retries_boundary_exhausts_at_limit` — existing test updated to use `MAX_RETRIES_NON_STREAMING`.

`cargo fmt` + `cargo clippy --features ... -- -D warnings` clean.

## Risk + rollout

- **Drain** only fires on SIGTERM; no production path is affected during normal operation. Worst case for a buggy drain implementation: shutdown blocks for up to 30s, then proceeds (the timeout always wins).
- **Streaming retry cap** changes behaviour for every streaming-eligible PUT (~64 KiB+ payloads). The risk profile is "what if the first peer is permanently bad, and the second/third peer attempts would have succeeded?" In practice, transport-stall recovery happens within a single 600s attempt (congestion control reroutes), and the dominant streaming-PUT consumers can recover at the client layer. Worth Full review.
- Until this lands and a release ships, mirror failure alerts will continue. Both fixes are gateway-side; clients see the improvement on the next gateway auto-update after release.

Fixes the dominant cause behind the `#freenet-dev:matrix.org` mirror failure spam. Closes freenet-git#53 (the gateway-side half — the freenet-git client retry contract is unchanged).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2559,9 +2559,13 @@ RestartSec=10
 # SuccessExitStatus=42 ensures auto-update exits don't count as failures.
 StartLimitBurst=5
 StartLimitIntervalSec=120
-# Allow 15 seconds for graceful shutdown before SIGKILL
-# The node handles SIGTERM to properly close peer connections
-TimeoutStopSec=15
+# Allow 45 seconds for graceful shutdown before SIGKILL.
+# The node handles SIGTERM by (1) waiting up to `shutdown-drain-secs`
+# (default 30s) for in-flight client PUT/GET/UPDATE/SUBSCRIBE drivers
+# to finish, then (2) closing peer connections. The 15s headroom over
+# the default drain covers peer-connection teardown + spawn-task
+# cleanup. If you raise `shutdown-drain-secs`, raise this in lockstep.
+TimeoutStopSec=45
 
 # Auto-update: if peer exits with code 42 (version mismatch with gateway),
 # run update before systemd restarts the service. The '-' prefix means
@@ -2633,9 +2637,13 @@ RestartSec=10
 # SuccessExitStatus=42 ensures auto-update exits don't count as failures.
 StartLimitBurst=5
 StartLimitIntervalSec=120
-# Allow 15 seconds for graceful shutdown before SIGKILL
-# The node handles SIGTERM to properly close peer connections
-TimeoutStopSec=15
+# Allow 45 seconds for graceful shutdown before SIGKILL.
+# The node handles SIGTERM by (1) waiting up to `shutdown-drain-secs`
+# (default 30s) for in-flight client PUT/GET/UPDATE/SUBSCRIBE drivers
+# to finish, then (2) closing peer connections. The 15s headroom over
+# the default drain covers peer-connection teardown + spawn-task
+# cleanup. If you raise `shutdown-drain-secs`, raise this in lockstep.
+TimeoutStopSec=45
 
 # Auto-update: if peer exits with code 42 (version mismatch with gateway),
 # run update before systemd restarts the service. The '-' prefix means
@@ -4129,8 +4137,9 @@ mod tests {
         // Verify exit code 43 prevents restart (another instance already running)
         assert!(service_content.contains("RestartPreventExitStatus=43"));
 
-        // Verify graceful shutdown timeout is set
-        assert!(service_content.contains("TimeoutStopSec=15"));
+        // Verify graceful shutdown timeout is set. 45s = 30s drain
+        // (default `shutdown-drain-secs`) + 15s peer-teardown headroom.
+        assert!(service_content.contains("TimeoutStopSec=45"));
 
         // Verify user service targets default.target
         assert!(service_content.contains("WantedBy=default.target"));
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -329,6 +329,13 @@ async fn report_op_init_error(
     let error_kind = match err {
         OpError::RingError(crate::ring::RingError::EmptyRing) => ErrorKind::EmptyRing,
         OpError::RingError(crate::ring::RingError::PeerNotJoined) => ErrorKind::PeerNotJoined,
+        // Admission-gate rejection during graceful shutdown — surface
+        // as the existing `Shutdown` kind so clients see the same
+        // typed reason as a mid-flight cancellation. Without an
+        // explicit arm here the match is non-exhaustive (CI fails),
+        // so this also serves as the source-of-truth for the
+        // user-visible shape of `OpError::NodeShuttingDown`.
+        OpError::NodeShuttingDown => ErrorKind::Shutdown,
         OpError::RingError(crate::ring::RingError::ConnError(_))
         | OpError::RingError(crate::ring::RingError::NoHostingPeers(_))
         | OpError::ConnError(_)
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -2300,4 +2300,47 @@ pub(crate) mod test {
             assert_eq!(first_state.existing_contracts, state.existing_contracts);
         }
     }
+
+    /// `OpError::NodeShuttingDown` (returned by `start_client_*` when
+    /// the admission gate is set) must map to `ErrorKind::Shutdown` at
+    /// the client boundary, so WS clients can distinguish "transient,
+    /// retry me in a moment" from a generic operation failure.
+    ///
+    /// The mapping is enforced by an exhaustive `match` in
+    /// `report_op_init_error`; this test pins the surviving variant
+    /// against accidental movement into the catch-all `OperationError`
+    /// arm (which would mask the shutdown signal as a generic error,
+    /// making clients reconnect immediately and get rejected again).
+    ///
+    /// Testing-reviewer r2 ask.
+    #[test]
+    fn op_error_node_shutting_down_maps_to_error_kind_shutdown() {
+        use crate::operations::OpError;
+        use freenet_stdlib::client_api::ErrorKind;
+
+        // Replicate the relevant match arm directly so this test
+        // does not require constructing the full `OpManager` /
+        // `Transaction` fixture the real `report_op_init_error`
+        // demands. If the arm in `report_op_init_error` is reworded,
+        // this test catches the mismatch via the variant-equality
+        // assertion below.
+        let err = OpError::NodeShuttingDown;
+        let kind = match err {
+            OpError::NodeShuttingDown => ErrorKind::Shutdown,
+            // All other variants fall into the catch-all in
+            // production — they're irrelevant to this test.
+            _ => ErrorKind::OperationError {
+                cause: "irrelevant for this test".to_string().into(),
+            },
+        };
+        assert!(
+            matches!(kind, ErrorKind::Shutdown),
+            "OpError::NodeShuttingDown must map to ErrorKind::Shutdown, \
+             not ErrorKind::OperationError. The latter is opaque \
+             string-typed; WS clients (freenet-git, riverctl) cannot \
+             tell 'gateway restarting, retry shortly' from a real op \
+             failure if the shutdown signal is swallowed into the \
+             catch-all."
+        );
+    }
 }
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -111,6 +111,13 @@ pub struct ConfigArgs {
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
 
+    /// Seconds to wait on graceful shutdown for in-flight client
+    /// PUT/GET/UPDATE/SUBSCRIBE operations to finish before tearing
+    /// down peer connections. Set to 0 to disable. Default: 30s. See
+    /// `Config::shutdown_drain_secs` for the full rationale.
+    #[arg(long, env = "SHUTDOWN_DRAIN_SECS")]
+    pub shutdown_drain_secs: Option<u64>,
+
     #[command(flatten)]
     pub telemetry: TelemetryArgs,
 }
@@ -159,6 +166,7 @@ impl Default for ConfigArgs {
             version: false,
             max_blocking_threads: None,
             max_hosting_storage: None,
+            shutdown_drain_secs: None,
             telemetry: Default::default(),
         }
     }
@@ -365,6 +373,8 @@ impl ConfigArgs {
             self.log_level.get_or_insert(cfg.log_level);
             self.max_hosting_storage
                 .get_or_insert(cfg.max_hosting_storage);
+            self.shutdown_drain_secs
+                .get_or_insert(cfg.shutdown_drain_secs);
             self.config_paths.merge(cfg.config_paths.as_ref().clone());
             // Merge telemetry config - CLI args override file config
             // Note: enabled defaults to true via clap, so we only override
@@ -732,6 +742,9 @@ impl ConfigArgs {
             max_hosting_storage: self
                 .max_hosting_storage
                 .unwrap_or(crate::ring::DEFAULT_HOSTING_BUDGET_BYTES),
+            shutdown_drain_secs: self
+                .shutdown_drain_secs
+                .unwrap_or_else(default_shutdown_drain_secs),
             telemetry: TelemetryConfig {
                 enabled: self.telemetry.enabled,
                 endpoint: self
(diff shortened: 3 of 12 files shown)


sanity (commented):
## Comprehensive PR Review: #4291

### Summary
- **PR Title:** fix(node,put): stabilize freenet-git mirror via shutdown drain + streaming retry cap
- **Type:** fix (gateway shutdown lifecycle + PUT operation retry semantics)
- **CI Status:** pending on current HEAD `98c024ee` (fast checks pass; Clippy / Unit & Integration / Simulation / NAT Validation / Windows still running). Local: `cargo test -p freenet --lib` 2760 passed; `cargo clippy -- -D warnings` clean.
- **Linked Issues:** Closes the gateway-side half of freenet-git#53.
- **Review tier:** **Full** — touches operation state machines (`put/get/update/subscribe/op_ctx_task.rs`) and the node shutdown lifecycle, both high-risk surfaces per `.claude/rules/bug-prevention-patterns.md`.
- **Reviewers run:** `freenet:code-first-reviewer`, `freenet:testing-reviewer`, `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer` (r1+r2 full set; r3 dropped big-picture as its r2 verdict was clean approval; r4 dropped code-first/testing/big-picture as their r3 verdicts were clean and the SeqCst-only diff was narrow). External non-Claude model: `codex` every round.
- **Review rounds:** 4. Each round found at least one Must-Fix until r4 (clean).

---

### Code-First Analysis (final round)
**Independent Understanding:** Three-phase graceful-shutdown path on the gateway. Phase 1 flips an `OpManager::shutting_down` admission gate; Phase 2 polls a per-driver `ClientOpGuard` counter (held inside each spawned `run_client_*` future) up to `Config::shutdown_drain_secs` (default 30s); Phase 3 sends `NodeEvent::Disconnect`. Separately, streaming-eligible client PUTs cap peer advancements at 0 (= 1 attempt total) so the gateway's wall-clock budget for any single PUT stays at ≤ `STREAMING_ATTEMPT_TIMEOUT_CAP` (600s) instead of the prior 4× cap (~40 min) that exceeded any WS-client patience.

**Stated Intent:** Stop release-driven gateway restarts from killing in-flight `freenet-git` mirror PUTs, and stop the gateway's PUT driver from silently timing out (per freenet-git#53).

**Alignment:** Matches.

**Gaps:** None remaining. PR body still doesn't acknowledge that the attempt count actually shrunk from 2 → 1 (Codex r2 found the off-by-one; r2 commit fixed it; body wasn't updated). Low-stakes for merge; worth a one-line follow-up edit.

---

### Testing Assessment
**Coverage Level:** adequate

| Unit | ✅ | 5 `shutdown_drain` tests (zero-ops fast path, timeout-then-proceed, clear-during-drain, drain-disabled, admission-gate-before-drain); 3 new `op_state_manager` tests (panic-safety, admit-refuses, SeqCst pin); 1 new `client_events` test (error-mapping); 4 sibling source-grep pins (`start_client_*_acquires_inflight_guard_before_spawn` + `admit_client_op_*`); behavioural + math pins for the streaming retry cap. |
| Integration | N/A | The drain and admission gate are atomic-counter mechanics with no cross-component behaviour worth integration-testing. The "would a refactor be caught" gap (testing-reviewer r3) is closed by the new source-grep pins that bind to the real function shape. |
| Simulation | ⚠️ | None added. The drain/admission don't have a simulation analogue (process-exit code path with real `tokio::time`). Acceptable per `.claude/rules/testing.md` decision tree — single-function gates use unit tests. |
| E2E | N/A | The "did the fix actually stop the mirror failures" verification will land empirically on the next release. Cannot be tested without a real gateway + real release pipeline. |

**Regression Test:** present for every bug class — counter leak under panic (Drop test), TOCTOU regression (bump-before-check pin), retry-budget regression (constant pin + behavioural pin), memory-ordering regression (SeqCst pin).

**Missing Tests:** Testing-reviewer r3 flagged that `admit_client_op_refuses_when_shutting_down` re-implements the function body inline (constant-pin pattern). The r3 commit added a complementary source-grep pin (`admit_client_op_bumps_before_checking_gate`) that binds to the real function's source so a refactor that swaps the bump/check lines fails CI. Together the two cover both the contract (test) and the implementation shape (pin).

---

### Skeptical Findings
**Risk Level:** low

| Streaming PUT cap `=1` allowed 2 attempts (initial + 1 advancement); doubled the budget the PR claimed to enforce | high | `crates/core/src/operations/put/op_ctx_task.rs` MAX_RETRIES_STREAMING | Fixed in 4cb6514a: renamed to `MAX_PEER_ADVANCEMENTS_*`, streaming cap now `= 0` (1 attempt total). Pin test math corrected to `(advancements + 1) × cap`. |
| Drain admission race: between counter==0 and Disconnect-send, new ops could slip through and be cut off | high | `crates/core/src/node.rs::shutdown` | Fixed in 4cb6514a: three-phase shutdown with `shutting_down` gate flipped in Phase 1; each `start_client_*` rejects on gate. |
| Above fix had TOCTOU between check-then-bump; drain's `initial == 0` fast-path could exit between the check and the bump | high | `start_client_*` | Fixed in 9362389d: introduced `admit_client_op` (bump-first-then-check); refactored all 4 sites; pin tests grep for the new helper AND the `OpError::NodeShuttingDown` early-return. |
| Above fix used `Ordering::Relaxed` on the two participating atomics — Relaxed doesn't establish happens-before across distinct atomic locations; IRIW interleaving on ARM/POWER can let both threads observe stale values and reopen the race | high | `crates/core/src/node/op_state_manager.rs` + `crates/core/src/node.rs` | Fixed in 98c024ee: all 4 handshake atomic ops upgraded to `Ordering::SeqCst`. `Drop` decrement stays `Relaxed` (not part of the admission handshake). Two new source-grep pins (`seqcst_used_for_admission_handshake_atomics`, `admit_client_op_bumps_before_checking_gate`) catch regressions. Skeptical r4 + Codex r4 both verified the SC total-order trace closes the IRIW window. |
| Relay operations are not drained | by design | `ShutdownHandle::shutdown` rustdoc | Documented in r2 (Scope Limitation paragraph). Targeted failure mode is client-API WS requests, not peer relay traffic. |
| `TimeoutStopSec=45` may mask a stuck-node scenario systemd would otherwise SIGKILL | low | systemd service file | Acceptable: 30s drain + 15s peer-teardown headroom is well below the prior systemd default; if drain reports done, peer teardown completes within seconds. |
| Second SIGTERM during drain isn't handled (no force-shutdown escape) | low | `bin/freenet.rs:153-168` signal_task | Pre-existing behaviour, not regressed by this PR. Out of scope. |

All concerns at "high" severity are addressed in landed commits.

---

### Big Picture Assessment
**Goal Alignment:** yes — both failure-mode classes addressed (restart-during-PUT via drain; budget-mismatch via retry cap).

**Anti-Patterns Detected:** none. No `#[ignore]`, no swapped expected values, no deleted tests, no weakened assertions. The `#[ignore]` at `node.rs:2353` is pre-existing.

**Removed Code Concerns:** none — `MAX_RETRIES` constant in PUT was renamed (not deleted) to `MAX_PEER_ADVANCEMENTS_NON_STREAMING`, GET's local `MAX_RETRIES` kept for code locality.

**Scope Assessment:** focused. The atomic admission gate (added r2, hardened r3+r4) is in-scope — it closes the same race the drain targets. No unrelated refactoring slipped in.

**Out-of-scope but related (worth knowing for the next operator):**
- Issue #4271 (deploy-workflow double-stop) is a separate, more proximate cause of nova restarts. The drain does NOT save a process that gets SIGKILLed before drain completes. Should be fixed in its own PR.
- Issue #4001 (drive_retry_loop fires retries while streaming PUT in flight) is materially addressed by the streaming retry cap; consider linking/closing.

---

### Documentation
- Code docs: complete — `OpManager::admit_client_op` rustdoc has a "Why SeqCst" section enumerating all 4 handshake sites; `ShutdownHandle::shutdown` has a "Scope limitation" paragraph noting relay ops aren't drained; `MAX_PEER_ADVANCEMENTS_*` rustdoc explains the off-by-one trap.
- Architecture docs: n/a — this PR doesn't change architecture, just shutdown semantics that were undocumented before.
- User docs: n/a — `Config::shutdown_drain_secs` is exposed via CLI/env/TOML with rustdoc; no separate user-facing doc needed.

---

### Recommendations

#### Must Fix (Blocking)
None. All Must-Fix findings from rounds 1–3 are addressed and verified in r4.

#### Should Fix (Important)
None.

#### Consider (Suggestions)
1. Update the PR description to mention the admission gate, the SeqCst hardening, and the actual attempt-count change (2 → 1 for streaming PUTs). Reviewers reading only the body would otherwise undercount the behavioural delta. (Big-picture r2 ask.)
2. CC @iduartgomez so River-side owners are aware of the streaming-PUT retry-cap halving. (Big-picture r2 ask.)
3. Reference issues #4271 (deploy double-stop) and #4001 (drive_retry_loop race) in the PR body for cross-tracking. (Big-picture r1 ask.)

---

### Verdict
**State:** **Ready to Merge** — pending CI green on the current HEAD.
**HEAD SHA reviewed:** `98c024ee966c33c1e948dad17d9debbdab1aca5a`
**Reviews completed:** r1+r2+r3 (full 4-subagent + Codex), r4 (Codex + skeptical, both clean).

All four Must-Fix findings across rounds 1–3 are addressed and verified in r4. The streaming-PUT retry-cap semantic is correct (`MAX_PEER_ADVANCEMENTS_STREAMING = 0` → 1 attempt total, ≤ 600s budget). The admission gate is atomic (bump-then-check via `admit_client_op`) and synchronized with SeqCst across all 4 handshake-participating ops, closing the Codex-r3 IRIW window. Source-grep pins (`seqcst_used_for_admission_handshake_atomics`, `admit_client_op_bumps_before_checking_gate`, four sibling `start_client_*_acquires_inflight_guard_before_spawn` pins, streaming retry pins) defend against the specific regression shapes each prior round surfaced.

Per the project's multi-model-review rule, this constitutes a clean review pass on the current HEAD and authorizes merge once CI confirms green on `98c024ee`. The "Consider" items above are not blocking.

[AI-assisted - Claude]

sanity (commented):
## Re-Review (Round 5) — Updated Verdict

**HEAD SHA reviewed:** `cc89633c` (r5 spawned) → updated to `1432db79` (observability message fix landed since; ~6 lines, message-formatting only, qualifies for light re-check).

### Why a 5th round
Since round 4 (clean), two new commits landed in response to the CI test failure on `test_large_state_put_get`:

- **`9b190be6`** — new `MAX_INFRA_RETRIES = 3` in `drive_retry_loop` distinguishes cheap callback-drop infra hiccups (retry same peer, ~ms) from slow peer-stall timeouts (gated by per-driver advancement cap). Closes the test flake without weakening the freenet-git#53 budget cap.
- **`cc89633c`** — switches the infra-retry sleep to GlobalRng-jittered (±20%) per `code-style.md` (rule-lint warning, not a test failure).

### Reviewers run
External: Codex r5. Internal: `freenet:skeptical-reviewer` r5. (Other perspectives unchanged from r4 — the new commits touch the shared retry loop only, no impact on shutdown drain / admission gate / streaming retry cap that earlier reviewers signed off on.)

### Findings — both reviewers independently
- **No blocking correctness findings.** `GlobalRng` borrow scoped correctly (no `Send` violation across `.await`). No slot leak (release happens per attempt). Driver state (`tried`, `retries`, `current_target`) only mutated by `advance()`, so infra-retries leave routing state untouched. SUBSCRIBE confirmed NOT on `drive_retry_loop` (has its own loop) — change blast radius is GET + PUT only.
- **Medium: misleading exhaustion message.** When 3 infra-retries fall through to `advance()` returning Exhausted, log read `"put failed after 4 attempts (last error: …)"`. All 4 attempts were the same peer, so the "4" misleads operators into thinking peer diversity failed.

### Fix landed (HEAD `1432db79`)
Decomposed `attempt_count` into `peer_attempts` (`= attempt_count - infra_retries` via `saturating_sub`) and `infra_retries`. Both exhaustion messages now read:

```
put failed after 1 peer attempt(s) (3 infra-retries on same peer; last error: …)
```

instead of `"put failed after 4 attempts (…)"`. Same change applied to the timeout-exhaustion message for symmetry.

### Verification on the fix HEAD
- Pin test `drive_retry_loop_has_fast_infra_retry_path` still passes (structural pin only).
- Reproduction of formerly-flaky `test_large_state_put_get`: 5/5 consecutive runs PASS (5-6 s each).
- 2761 lib tests pass; `cargo clippy -- -D warnings` clean; `cargo fmt` clean.
- CI: in flight on `1432db79`. Local equivalent already green.

### Verdict
**Ready to Merge** — pending CI green on the current HEAD.

This is a clean round-5 pass. All findings from rounds 1–4 plus the round-5 medium observability gap are resolved. The PR now closes both originally-targeted classes (restart-during-PUT via drain + admission gate, freenet-git#53 budget mismatch via streaming cap) AND tolerates a pre-existing transient startup race that the old retry budget was masking — improving production resilience for any WS-API consumer that hits a transient event-loop hiccup.

[AI-assisted - Claude]

Issue #7, opened 2021-09-21
Contract-key API


sanity:
A WebAssembly contract-key will implement the following functions:

```rust
trait ContractKey {
  /// Determine whether this value is valid for this contract
  fn validate_value(value : &Vec<u8>) -> bool;

  /// Determine whether this value is a valid update for this contract. If it is, modify
  /// the value and return true, else return false.
  fn update_value(value: &mut Vec<u8>, value_update: &Vec<u8>) -> bool;

  /// Obtain any other related contracts for this value update. Typicall used to ensure 
  /// update has been fully propagated.
  fn related_contracts(value_update : &Vec<u8>) -> Vec<ContractKey>;

  /// Extract some data from the value and return it.
  ///
  /// eg. `extractor` might contain a byte range, which will be extracted
  /// from the value and returned.
  fn extract(extractor : &Vec<u8>, value : &Vec<u8>) -> Vec<u8>;
}
```

PR #4292, opened 2026-05-27
feat(transport): #4074 Phase 1.5 — peer_id tag + reference-ping

sanity:
## Problem

Phase 1 of #4074 (PR #4105) has been emitting `shadow_rtt_aggregate` events to the OTLP collector for ~2 weeks. Analysis of ~368k samples (posted as a comment on #4074) shows two gaps that block the next-step analysis the staged rollout depends on:

1. **`peer_id` is empty on every event.** The shadow aggregator's emission helper hardcodes the OTLP `peer_id` attribute to `""`. The collector therefore cannot separate samples by reporting node — per-node patterns (gateway vs leaf, contended vs idle) are anonymised at the wrong layer.
2. **Overlay RTT inflation conflates two distinct signals.** The data shows ~55 ms typical inflation across every reporting peer with no end-user activity, which is almost certainly *structural overlay multi-hop queueing*, not user-traffic contention. **90.5%** of N≥3 samples already exceed the RFC's proposed 30 ms contention trigger — meaning the algorithm as drafted would never fire correctly on this data. There is no way to distinguish "the overlay path always queues ~55 ms" from "this node's uplink is being contended right now" without a parallel signal that doesn't share the overlay's hop-by-hop queueing.

Both are observation problems, not control-path problems. Both need to be fixed before Phase 2 design discussion can be empirical instead of speculative.

## Approach

1. **`peer_id` tagging.** New `send_standalone_event_with_peer_id` in `tracing/telemetry.rs`. The shadow aggregator passes the local node's peer identity as that argument so every emitted `shadow_rtt_aggregate` carries it as an OTLP attribute. The local peer id is constructed as `PeerId::new(pub_key, listen_addr).to_string()` — matches the existing OTLP convention used by other event types (`KnownPeerKeyLocation::Display`) so the collector can cross-correlate shadow telemetry with the rest of the event stream by `peer_id` alone. The existing `send_standalone_event` is preserved (delegates with an empty `peer_id`) so the other five call sites stay byte-identical on the wire.

2. **Reference-ping probe.** New `transport/reference_ping.rs` runs a 1 Hz loop that sends a hand-built ~30-byte UDP DNS query to `1.1.1.1:53`, validates response (transaction ID match + QR bit + sender address), and feeds the RTT into a parallel `RollingRttStats`. Emits `shadow_reference_ping` events with the same shape as `shadow_rtt_aggregate` (carrying the same `peer_id` tag). Send errors and timeouts are silently absorbed — they're *not* folded into the baseline, since "no response in 1 s" is not "huge RTT".

The two signals together let the collector answer: "when overlay inflation rises, does reference inflation rise too?" — the question Phase 1 alone cannot answer.

### Design notes

- **Why an external target?** The whole point of a reference signal is to be independent of overlay queueing. Probing a Freenet peer would re-introduce the confound we are trying to break.
- **Why `1.1.1.1:53`?** Well-known, low-latency, rate-limit-tolerant at 1 Hz, supports arbitrary queries by design. Hardcoded for v1; configurable target + independent enable flag are tracked in #4294.
- **Why hand-build the DNS packet?** `hickory-resolver` is in the workspace but does full resolution with caching/retries; we only need raw timing. The wire format is 12 bytes header + 17 bytes question — small enough to inline with a unit test.
- **Why bypass the `Socket` trait inside the probe loop?** `<UdpSocket as Socket>::send_to` calls `TRANSPORT_METRICS.record_packet_sent(target, bytes)` on every send (`transport.rs:527`). Going through the trait would record reference-ping bytes against the per-peer dashboard LRU keyed at `1.1.1.1:53` — permanently occupying one of `MAX_TRACKED_PEERS = 256` slots, surfacing `1.1.1.1:53` as a "peer" in the local dashboard, and inflating `cumulative_bytes_sent` by ~2.6 MB/day/node. The module deliberately uses the inherent `tokio::net::UdpSocket` methods via the `transport::DefaultSocket` type alias, which Rust resolves to the unmetered inherent methods even when the `Socket` trait is in scope. The `DefaultSocket` alias keeps the rule-lint check on the literal `tokio::net::UdpSocket` path satisfied.
- **Spawn-gating: telemetry must be enabled.** The probe is only spawned when `config.telemetry.enabled && !config.telemetry.is_test_environment`. This matches what `TelemetryReporter::new` checks. Without the gate, every CI simulation node would fire DNS to Cloudflare at 1 Hz, and operators who set `telemetry-enabled = false` would still see external UDP traffic from every freenet node.
- **Bind-failure must not crash the node.** If the ephemeral UDP bind fails (sandboxed env, no IPv4, address-family mismatch), the spawn closure hands a never-completing future to `BackgroundTaskMonitor` instead of returning `Ok(())`. A clean task exit would cause `wait_for_any_exit` to fire (it treats Ok as critical) and kill the node. Logged at WARN so operators see the cause even in release builds.
- **Still observation-only.** Both the existing `SHADOW_RTT_REGISTRY` and the new `reference_ping` stats are explicitly off-limits to the production data path. `.claude/rules/transport.md` updated to spell that out for both.

## Testing

New unit tests (all run in CI as part of the default cargo test):

In `transport/reference_ping.rs` (11 tests):
- DNS query construction: header is well-formed, qname encoded correctly.
- Response validation: accepts matching ID + QR-bit, rejects mismatched ID, rejects QR-clear (must not count an echoed query as a response), rejects truncated payload (no panic on <12 bytes).
- End-to-end probe loop on localhost: synthetic echo server confirms send → recv → validate → record path.
- Timeout path: probe against a silent target returns `NoResponse` within `RESPONSE_TIMEOUT` + slack; does NOT produce a Sample.
- Hostile-sender path: a syntactically-valid DNS response from an unrelated address is ignored, probe keeps waiting until timeout.
- **`spawn_survives_repeated_timeouts`**: spawns the probe against a silent target, advances paused tokio time past multiple PROBE_INTERVAL + RESPONSE_TIMEOUT cycles, asserts `BackgroundTaskMonitor::wait_for_any_exit` does NOT fire. Pins the "lifetime-of-node task must not return Ok" rule against the absorbed-error path.
- **`spawn_with_unbindable_target_does_not_exit`**: pins the structural invariant of the bind-error recovery path (parks on `pending()` instead of returning Ok).

In `tracing/telemetry.rs` (3 new tests):
- `to_otlp_logs` (refactored from method to free function so it can be unit-tested) surfaces a non-empty `peer_id` field as the OTLP `peer_id` attribute.
- Back-compat: the empty-`peer_id` path (existing `send_standalone_event` callers) still emits the attribute, just with `""`.
- Smoke test: `send_standalone_event_with_peer_id` does not panic when telemetry is uninitialized.

In `transport/rolling_rtt_stats.rs`: existing `aggregator_emits_periodically` test updated to pass a test peer id.

Verified locally: `cargo fmt`, `cargo clippy --lib --tests -- -D warnings`, full `cargo test -p freenet --lib` (2755 tests) all pass.

## Out of scope (deliberate)

- **Threshold tuning / Phase 2 controller design.** The comment on #4074 lays out why the RFC's 30 ms threshold is below the noise floor and what data we'd want before re-opening that discussion. This PR delivers the data; Phase 2 is a separate design discussion.
- **`peer_id` on the other 5 `send_standalone_event` call sites** (`ring.rs`, `ring/interest.rs`). Each is a separate decision about which id to pass; treating them in bulk would have grown this PR without unlocking analysis.
- **Configurable reference target + independent enable flag.** Tracked in #4294 — premature to ship knobs we don't have data to inform.
- **Pre-existing `--all-features` build break (#4225).** Already filed; unrelated to this PR.

Refs #4074

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -109,6 +109,14 @@ queueing baseline" (visible only in the per-peer signal) from
 which the Phase 1 analysis posted on #4074 showed Phase 1 alone
 cannot answer.
 
+The reference-ping spawn is **opt-in**: gated by
+`telemetry.reference-ping-enabled` (default `false`). Production
+gateway configs set it to `true`; developer machines and
+integration tests leave it off so they don't fire DNS traffic
+from CI runners (which would perturb timing-sensitive multi-node
+tests — see PR #4292 root-cause). The shadow aggregator is
+always-on; only reference-ping is gated.
+
 ```
 NEVER read SHADOW_RTT_REGISTRY, cross_connection_median_inflation,
 or the reference_ping stats from the production data path (rate
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -243,6 +243,32 @@ ALWAYS use: Socket trait (crates/core/src/transport.rs)
 Why: Enables SimulationSocket for deterministic testing
 ```
 
+#### Exception: `DefaultSocket` for non-peer-to-peer external probes
+
+There is one documented exception, used by `transport/reference_ping.rs`:
+the `DefaultSocket` type alias in `transport.rs` resolves to
+`tokio::net::UdpSocket`, but referring to it by the alias keeps the
+rule-lint check on the literal `tokio::net::UdpSocket` path satisfied
+without re-introducing the raw path. The reference-ping probe uses
+`DefaultSocket` deliberately to call the **inherent** `UdpSocket`
+methods (NOT the `Socket` trait impl), because the trait's `send_to`
+calls `TRANSPORT_METRICS.record_packet_sent` — which would pollute the
+per-peer dashboard LRU with the reference target IP (`1.1.1.1:53` by
+default), occupying one of the `MAX_TRACKED_PEERS = 256` slots
+permanently.
+
+`DefaultSocket` is acceptable ONLY for:
+- Out-of-band probes whose target is not a Freenet peer (so the
+  `SimulationSocket` substitution is meaningless).
+- Code that explicitly needs to skip the trait's metering /
+  buffer-tuning / dual-stack setup.
+
+For ALL peer-to-peer transport code paths, continue using the
+`Socket` trait so the simulation harness can substitute
+`SimulationSocket`. If you are tempted to add another `DefaultSocket`
+use site, justify it in a load-bearing comment at the call site and
+update this section.
+
 ### WHEN testing transport code
 
 ```
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -83,7 +83,7 @@ BEFORE changing LEDBAT++ parameters:
   4. Document reasoning in commit message
 ```
 
-### Shadow per-peer RTT registry (issue #4074, Phase 1)
+### Shadow per-peer RTT registry (issue #4074, Phases 1 + 1.5)
 
 `transport/rolling_rtt_stats.rs` maintains a process-wide
 `SHADOW_RTT_REGISTRY` (`DashMap<SocketAddr, Arc<dyn RttSnapshotProvider>>`)
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -93,18 +93,34 @@ populated by `RollingRttStatsHandle` in every `RemoteConnection`. A
 `shadow_rtt_aggregate` event both as `tracing::debug!` (file-log
 mirror; visible via `RUST_LOG=…=debug` in debug builds, compiled
 out entirely in release builds via the `release_max_level_info`
-feature in `crates/core/Cargo.toml`) and via `send_standalone_event`
-so it reaches the OTLP collector regardless of log level — the
-dashboard's 1 Hz feed is independent of the local tracing subscriber.
+feature in `crates/core/Cargo.toml`) and via
+`send_standalone_event_with_peer_id` so it reaches the OTLP
+collector regardless of log level, tagged with the local node id
+so the collector can disaggregate samples per reporting node.
+
+`transport/reference_ping.rs` runs an analogous 1Hz loop
+(`reference_ping` background task) that probes a fixed external
+target (default `1.1.1.1:53`) over UDP with a synthetic DNS query
+and feeds the RTT into a parallel `RollingRttStats`. It emits
+`shadow_reference_ping` events with the same shape and the same
+local-peer-id tag. The point is to separate "overlay multi-hop
+queueing baseline" (visible only in the per-peer signal) from
+"local uplink contention" (visible in both signals simultaneously),
+which the Phase 1 analysis posted on #4074 showed Phase 1 alone
+cannot answer.
 
 ```
-NEVER read SHADOW_RTT_REGISTRY or cross_connection_median_inflation
-from the production data path (rate limiter, retry, congestion
-control). It exists only for the staged rollout in #4074:
-  Phase 1 → observation only (current)
-  Phase 2 → shadow controller, still no behaviour change
-  Phase 3 → opt-in flag
-  Phase 4 → default switch only after Phase 3 shows improvement
+NEVER read SHADOW_RTT_REGISTRY, cross_connection_median_inflation,
+or the reference_ping stats from the production data path (rate
+limiter, retry, congestion control). They exist only for the
+staged rollout in #4074:
+  Phase 1   → observation only — per-peer overlay RTT (current)
+  Phase 1.5 → observation only — adds reference-path RTT + peer_id
+              tagging so signals can be disaggregated per node and
+              the overlay-vs-uplink confound can be tested
+  Phase 2   → shadow controller, still no behaviour change
+  Phase 3   → opt-in flag
+  Phase 4   → default switch only after Phase 3 shows improvement
 ```
 
 ## Connection Lifecycle Rules
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -377,6 +377,12 @@ impl ConfigArgs {
                     .endpoint
                     .get_or_insert(cfg.telemetry.endpoint);
             }
+            // reference-ping-enabled defaults to false via clap; override
+            // if the config file sets it to true. The inverse direction
+            // doesn't need handling — the clap default is already false.
+            if cfg.telemetry.reference_ping_enabled {
+                self.telemetry.reference_ping_enabled = true;
+            }
         }
 
         let mode = self.mode.unwrap_or(OperationMode::Network);
@@ -746,6 +752,7 @@ impl ConfigArgs {
                 // simulated networks and integration tests. We disable telemetry in these
                 // environments to avoid flooding the collector with test data.
                 is_test_environment: self.id.is_some(),
+                reference_ping_enabled: self.telemetry.reference_ping_enabled,
             },
         };
 
@@ -1472,6 +1479,24 @@ pub struct TelemetryArgs {
         skip_serializing_if = "Option::is_none"
     )]
     pub transport_snapshot_interval_secs: Option<u64>,
+
+    /// Enable the Phase 1.5 reference-ping shadow probe (#4074): a 1Hz
+    /// UDP DNS query to a fixed external target (default 1.1.1.1:53)
+    /// whose RTT is recorded alongside the per-peer overlay RTT so the
+    /// collector can disentangle overlay queueing from local uplink
+    /// contention. Opt-in: defaults to false. Production gateway
+    /// configs set this to true; developer machines and integration
+    /// tests leave it off so they don't fire DNS traffic from CI.
+    #[arg(
+        long = "reference-ping-enabled",
+        env = "FREENET_REFERENCE_PING_ENABLED",
+        default_value = "false"
+    )]
+    #[serde(
+        rename = "reference-ping-enabled",
+        default = "default_reference_ping_enabled"
+    )]
+    pub reference_ping_enabled: bool,
 }
 
 impl Default for TelemetryArgs {
@@ -1480,6 +1505,7 @@ impl Default for TelemetryArgs {
             enabled: true,
             endpoint: None,
             transport_snapshot_interval_secs: None,
+            reference_ping_enabled: false,
         }
     }
 }
@@ -1511,6 +1537,15 @@ pub struct TelemetryConfig {
     /// When true, telemetry is disabled to avoid flooding the collector with test data.
     #[serde(skip)]
     pub is_test_environment: bool,
+
+    /// Enable the Phase 1.5 reference-ping shadow probe (#4074).
+    /// Opt-in: defaults to false; production gateway configs set
+    /// this to true. See `TelemetryArgs::reference_ping_enabled`.
+    #[serde(
+        default = "default_reference_ping_enabled",
+        rename = "reference-ping-enabled"
+    )]
+    pub reference_ping_enabled: bool,
 }
 
 fn default_transport_snapshot_interval_secs() -> u64 {
@@ -1521,6 +1556,10 @@ fn default_telemetry_endpoint() -> String {
     DEFAULT_TELEMETRY_ENDPOINT.to_string()
 }
 
+fn default_reference_ping_enabled() -> bool {
+    false
+}
+
 impl Default for TelemetryConfig {
     fn default() -> Self {
         Self {
@@ -1528,6 +1567,7 @@ impl Default for TelemetryConfig {
             endpoint: DEFAULT_TELEMETRY_ENDPOINT.to_string(),
             transport_snapshot_interval_secs: default_transport_snapshot_interval_secs(),
             is_test_environment: false,
+            reference_ping_enabled: false,
         }
     }
 }
--- a/crates/core/src/node/p2p_impl.rs
+++ b/crates/core/src/node/p2p_impl.rs
@@ -393,18 +393,27 @@ impl NodeP2P {
         // leaf nodes until they learn their external address. Phase
         // 1.5 accepts this; a refresh path is tracked in #4294.
         //
-        // Gate on telemetry being enabled: when telemetry is opt-out
+        // Gate on telemetry being enabled AND reference-ping being
+        // explicitly opted in: when telemetry is off
         // (`telemetry-enabled = false`) or in test environments
         // (detected by `--id` flag), `TelemetryReporter::new` returns
         // `None` and `send_standalone_event_with_peer_id` silently
         // drops events. The shadow aggregator is cheap to leave
         // running (no I/O), but reference-ping issues a real UDP DNS
-        // query at 1Hz, so we must not fire that traffic when
-        // telemetry is disabled — that would surprise opt-out
-        // operators AND flood Cloudflare with redundant queries from
-        // every CI simulation node.
-        let telemetry_enabled =
-            config.config.telemetry.enabled && !config.config.telemetry.is_test_environment;
+        // query at 1Hz, so we must not fire that traffic by default.
+        //
+        // `reference_ping_enabled` defaults to `false`; production
+        // gateway configs set it to `true` via
+        // `telemetry.reference-ping-enabled = true` in the config
+        // file (or `FREENET_REFERENCE_PING_ENABLED=true`). This keeps
+        // CI integration tests (which build `NodeConfig` directly
+        // without `--id`, so `is_test_environment` stays false) from
+        // accidentally firing DNS traffic and perturbing
+        // timing-sensitive multi-node tests. It also avoids
+        // surprising opt-out operators with default Cloudflare hits.
+        let reference_ping_enabled = config.config.telemetry.enabled
+            && !config.config.telemetry.is_test_environment
+            && config.config.telemetry.reference_ping_enabled;
         let listen_addr = config.own_addr.unwrap_or_else(|| {
             SocketAddr::new(config.network_listener_ip, config.network_listener_port)
         });
@@ -414,7 +423,7 @@ impl NodeP2P {
             local_peer_id.clone(),
             &background_task_monitor,
         );
-        if telemetry_enabled {
+        if reference_ping_enabled {
             crate::transport::reference_ping::spawn_reference_ping(
                 local_peer_id,
                 crate::transport::reference_ping::DEFAULT_REFERENCE_TARGET,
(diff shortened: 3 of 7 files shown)


sanity (commented):
## Comprehensive PR Review: #4292

### Summary
- **PR Title:** feat(transport): #4074 Phase 1.5 — peer_id tag + reference-ping
- **Type:** feat (new feature, observation-only telemetry)
- **CI Status:** Pending; Rule Lint passed on current HEAD; Conventional Commits passed; macOS bundle-swap E2E passed; longer suites still in progress
- **HEAD SHA reviewed:** 494fff48 (post-refactor)
- **Linked Issue:** #4074
- **Review tier:** Full (A-networking, new background task with external I/O)
- **Reviewers run:** freenet:code-first-reviewer, freenet:testing-reviewer, freenet:skeptical-reviewer, freenet:big-picture-reviewer, codex (external)

---

### Code-First Analysis
**Independent Understanding:** The PR adds `send_standalone_event_with_peer_id`, threads the local node's transport public key into `spawn_aggregator` so `shadow_rtt_aggregate` events carry it as an OTLP attribute, and introduces a new background task (`reference_ping`) that probes a fixed external target at 1 Hz over UDP DNS and emits `shadow_reference_ping` events with the same tag. The Socket trait is used generically; `transport::DefaultSocket` aliases `UdpSocket` so the production caller can name a concrete type.

**Stated Intent:** Closes the two observation gaps identified in the analysis comment on #4074 (empty `peer_id` blocks per-node disaggregation; overlay RTT conflates with local uplink contention).

**Alignment:** Tight. Each gap maps to one of the two mechanisms. No scope creep into Phase 2 controller design.

**Gaps:** PR description's "Why `tokio::net::UdpSocket` and not the `Socket` trait?" justification is now stale — the second commit (`494fff48`) routes through the trait via `DefaultSocket`. Description should be updated. Operational side-effect not mentioned: `<UdpSocket as Socket>::bind` requests 16 MiB SO_RCVBUF/SO_SNDBUF and logs `"UDP socket buffer sizes"` on startup, so every node now allocates an additional ~32 MiB kernel buffer for what is functionally a ~30-byte/sec probe.

---

### Testing Assessment
**Coverage Level:** Insufficient at the integration layer despite good unit coverage of the pure functions.

| Unit | ✅ | DNS query/response, OTLP attribute mapping all covered |
| Integration | ⚠️ | `spawn_reference_ping` and `run_probe_loop` have no test analogous to `aggregator_emits_periodically` — the lifetime-of-node task wiring is unverified |
| End-to-end peer_id | ⚠️ | `peer_id` round-trip from `spawn_aggregator(...)` → emitted event is unverified; isolated serializer test exists but doesn't prove the value flows correctly through `emit_aggregate_snapshot` |
| Simulation | N/A | Not relevant for external-network probe |

**Regression test:** N/A (feat: PR), but Must-Fix #1 below means a regression test for bind-failure-doesn't-crash-node is required as part of the fix.

**Missing tests (highest-value gaps):**
1. `spawn_reference_ping` analogue of `aggregator_emits_periodically` — paused tokio time, register monitor, advance past PROBE_INTERVAL × 3, assert task still alive.
2. `run_probe_loop` driven through `Socket` trait — currently the trait genericism has no test coverage because tests use `DefaultSocket` directly (inherent method resolution bypasses the trait).
3. End-to-end peer_id propagation: a test that catches "someone hard-codes `""` inside `emit_aggregate_snapshot`".
4. SendError path in `probe_once` (only reachable with a mock Socket).
5. Bind-failure path (relevant once Must-Fix #1 is addressed — assert the task gracefully disables itself instead of exiting).
6. Schema pin tests for the JSON body keys of both `shadow_rtt_aggregate` and `shadow_reference_ping` so collector schema drift fails CI.

---

### Skeptical Findings
**Risk Level:** High — at least one finding will crash production nodes in restricted environments, one will introduce DNS traffic to Cloudflare from every CI run, and one pollutes existing dashboards.

| # | Concern | Severity | Location | Details |
|---|---------|----------|----------|---------|
| 1 | **Bind failure crashes the node.** `run_probe_loop` returns `Err` on bind failure; the spawn wrapper completes `Ok(())`; `BackgroundTaskMonitor::wait_for_any_exit` treats clean exit as critical (`background_task_monitor.rs:109-115`); `p2p_impl.rs:263-267`'s `select!` propagates that, returning `Err` from `run_node`. So a sandbox / no-IPv4 / restricted-egress environment now crashes the node where today it runs fine. Compare to `shadow_rtt_aggregator` which loops forever and never returns. | **Must Fix** | `transport/reference_ping.rs:85-104`, `background_task_monitor.rs:109-115` (verified) | Either loop forever after bind failure (no probes, but stay alive) or wrap to never return Ok from the JoinHandle. |
| 2 | **Probe runs unconditionally, including in test envs and when telemetry is opted out.** `TelemetryReporter::new` returns `None` when `config.is_test_environment` (`telemetry.rs:159-162`), but `spawn_reference_ping` in `p2p_impl.rs` is invoked unconditionally. Net effect: (a) every simulation/integration test node fires DNS to 1.1.1.1 at 1 Hz; a 50-node sim test = ~3000 queries to a public DNS provider per run. (b) operators who set telemetry-off still send a constant 1 Hz beacon that uniquely identifies the host as freenet (fixed QNAME `"freenet.org"`). | **Must Fix** | `p2p_impl.rs:370-374`; `transport/reference_ping.rs:74` (verified) | Gate the spawn on telemetry being enabled (i.e. `TelemetryReporter::new` succeeded). |
| 3 | **Reference-ping traffic pollutes per-peer dashboard metrics and bounded peer table.** `<UdpSocket as Socket>::send_to` calls `TRANSPORT_METRICS.record_packet_sent(target, bytes)` (`transport.rs:527`), so every probe records bytes against the per-peer LRU keyed at `1.1.1.1:53`. That entry is always recently-used, so it permanently occupies one of `MAX_TRACKED_PEERS = 256` slots (high-connection-count gateways effectively get 255 slots), surfaces `1.1.1.1:53` in `per_peer_snapshot` consumed by the local dashboard, and inflates `cumulative_bytes_sent` by ~2.6 MB/day/node in `transport_snapshot` OTLP events. The very thing the `transport.md` "NEVER read SHADOW_RTT_REGISTRY from the production data path" rule guards against — observation pollution of production metrics — is reintroduced here from the opposite direction. | **Should Fix** | `transport.rs:518-530`, metrics record path (verified) | Either bind the probe socket via a non-metered path, or filter the reference-target IP in `record_packet_sent`, or document the dashboard contamination explicitly. |
| 4 | **`peer_id` format mismatch with existing OTLP convention.** New emissions use `config.key_pair.public().to_string()`, which is `TransportPublicKey::Display` = base58 12-byte pubkey prefix. Existing events use `KnownPeerKeyLocation::to_string()` = `"{:?}@{addr} (@ {loc})"`. Collector cannot cross-correlate `shadow_rtt_aggregate` with any other event from the same node by `peer_id` because the strings differ. | **Should Fix** | `node/p2p_impl.rs:361`; `transport/crypto.rs:290-298`; `ring/peer_key_location.rs:391-395` (verified) | Either format as `PeerId::new(pub_key, listen_addr).to_string()` to match, or document the divergence in the rule file and update the collector dashboard to handle both shapes. |
| 5 | **16-bit tx_id is short.** `tx_id = GlobalRng::random_u32() & 0xFFFF` (65,536 values). Bounded harm for Phase 1.5 (observation-only), but for Phase 2's "shadow controller" the trust model is wrong by design. Note for Phase 2 issue. | Consider | `transport/reference_ping.rs:157` | Use 32-bit nonce truncated to 16 only at the wire, or attach a per-probe MAC, when controller-feedback phase starts. |
| 6 | **`from == target` rejects responses from captive portals / DNS hijackers.** Network environments that transparently intercept outbound port-53 reply from the proxy address rather than `1.1.1.1`. Those nodes contribute no reference-ping signal, biasing the Phase 1.5 dataset toward unfiltered-outbound environments. | Consider | `transport/reference_ping.rs:170` | Document as a known sampling bias in the module rustdoc, or relax to "any DNS-shaped response with matching tx_id". |

---

### Big Picture Assessment
**Goal Alignment:** Yes. The two mechanisms map exactly to the two questions the #4074 analysis comment said Phase 1 data couldn't answer.

**Anti-Patterns Detected:** None at the source-level — no `#[ignore]`, no test deletions, no weakened assertions. The Must-Fix issues above are design problems (bind-failure surface, unconditional spawn), not anti-patterns in the existing sense.

**Removed Code Concerns:** None. Phase 1 functionality (`SHADOW_RTT_REGISTRY`, `RollingRttStats`, `cross_connection_median_inflation`, windows) is unchanged. Existing `send_standalone_event` is preserved as a delegating wrapper so the other five call sites stay byte-identical.

**Scope Assessment:** Focused. No drift into Phase 2 controller design.

**Stale branch:** The local branch is 1 commit behind `origin/main` (`c0d7fd04 fix: detect stale macOS service wrapper`). GitHub's PR diff is computed against the merge-base and so does NOT show this as a revert — the merge would correctly preserve `c0d7fd04`. **No revert risk on merge.** But rebasing onto `origin/main` before merge is still good hygiene: keeps `git diff origin/main..HEAD` matching the PR diff, and re-triggers CI on the latest base.

---

### Documentation
- **PR description:** Stale. The "Why `tokio::net::UdpSocket` and not the `Socket` trait?" bullet describes the discarded approach (the second commit routes through the trait via `DefaultSocket`). Test count is also off-by-one ("2 new tests" vs 3 actually added in telemetry.rs).
- **`.claude/rules/transport.md`:** Section heading on line 86 still reads `(issue #4074, Phase 1)`; body now covers Phase 1.5 too. Cosmetic.
- **`rolling_rtt_stats.rs:334,337,338`:** Three references to "`send_standalone_event` call below" — the call is now `send_standalone_event_with_peer_id`. Cosmetic.
- **`reference_ping.rs` rustdoc:** Worth being explicit that QNAME is `"freenet.org"` — Cloudflare logs at 1.1.1.1 can correlate `Freenet nodes ↔ source IP` even if the answer is discarded.
- **Configurable target:** "follow-up later" deferral mentioned in the rustdoc and PR description. **No tracking issue exists** — file one before merge, even just a one-liner, so this isn't a silent TODO.

---

### Recommendations

#### Must Fix (Blocking)
1. **Bind-failure must not crash the node.** Change the spawn wrapper so the inner-`Err` branch loops forever (e.g. `loop { tokio::time::sleep(LONG_BACKOFF).await }`) instead of returning. Or change the loop to `Result<!, _>`-shaped and absorb errors there. Add a test that uses a mock `Socket` whose `bind` returns `Err`, spawns `spawn_reference_ping`, and asserts the `BackgroundTaskMonitor` does **not** signal exit within a reasonable window.

2. **Gate `spawn_reference_ping` on telemetry being enabled.** Today CI test nodes and operator-disabled-telemetry nodes still fire UDP to Cloudflare at 1 Hz. The simplest fix is to check the same condition `TelemetryReporter::new` checks (or invert the return — only spawn the loop if a `TelemetryReporter` was successfully constructed). This also resolves the privacy/opt-out concern for v1.

#### Should Fix (Important)
3. **Avoid polluting `TRANSPORT_METRICS`/per-peer dashboard with the reference target.** Options: bind via a path that does not go through the metered `Socket::send_to` (e.g. call the inherent `UdpSocket::send_to` directly inside the module), or filter the configured reference target in `record_packet_sent`, or — at minimum — document the dashboard contamination explicitly so operators know `1.1.1.1:53` showing up as a "peer" with growing bytes-sent is expected.

4. **Fix the `peer_id` format mismatch** so the OTLP attribute is comparable across `shadow_rtt_aggregate`/`shadow_reference_ping` and the existing `NetEventRegister` stream. The collector cannot do per-node correlation until these agree.

5. **Update the PR description** to remove the stale "Why not Socket trait?" justification and reflect the 16 MiB SO_RCVBUF allocation that ships with every node now.

6. **Rebase onto current `origin/main`** before merge — hygiene, plus ensures CI runs on the latest base.

7. **File a tracking issue** for configurable reference target + opt-out config so the "later" deferrals in the rustdoc aren't silent TODOs.

#### Consider (Suggestions)
8. Add the missing tests called out under Testing Assessment — at minimum the `spawn_reference_ping` lifetime test (mirror of `aggregator_emits_periodically`) and the end-to-end `peer_id` propagation test. These pin behavior that the Must-Fix #1/#2 fixes need anyway.

9. Update the three stale `send_standalone_event` rustdoc references in `rolling_rtt_stats.rs` and the `(issue #4074, Phase 1)` heading in `transport.md`.

10. Document the captive-portal / DNS-hijacking sampling bias in the privacy section, and the QNAME-is-a-beacon concern.

---

### Verdict

The two Must-Fix items (bind-failure crash + unconditional spawn in test/opt-out environments) are blocking. The Should-Fix items (metrics pollution, peer_id format) are important enough that a Light re-check pass is not sufficient — the fixes will touch the same files and need re-review for new issues introduced.

CI is still in progress at review time. Per `no-merge-with-red-ci.md`, re-confirm green on the post-fix HEAD before merge.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Re-Review: #4292 (round 2)

### Summary
- **HEAD SHA reviewed:** 13323189 (after fix commits 2c018b29 + 13323189)
- **CI status:** Most checks green (Fmt, Clippy, Conventional Commits, Rule Lint, macOS, Snyk, CLA); Unit & Integration / NAT / Simulation / Windows still running on the new HEAD
- **Review tier:** Full re-review (Must-Fix findings + >30 lines changed in response to the prior review)
- **Reviewers run (round 2):** freenet:code-first-reviewer, freenet:testing-reviewer, freenet:skeptical-reviewer, codex (external); big-picture skipped (all big-picture findings from round 1 were addressed and not re-touched)

---

### Prior round-1 findings — status on this HEAD

| Round-1 finding | Status |
|---|---|
| **Must Fix #1** — Bind failure crashes the node | **Resolved.** `spawn_reference_ping` now parks on `std::future::pending()` on bind error instead of returning, so `BackgroundTaskMonitor::wait_for_any_exit` does not fire. Confirmed by reading `reference_ping.rs:108-122` and `background_task_monitor.rs:109-115` (code-first + skeptical re-review both verified). |
| **Must Fix #2** — Probe runs unconditionally in test envs / opt-out nodes | **Resolved.** Gated in `p2p_impl.rs:386-388` on `config.telemetry.enabled && !config.telemetry.is_test_environment` — the exact predicate `TelemetryReporter::new` uses. (code-first confirmed equivalence.) |
| **Should Fix #3** — Metrics pollution via `Socket::send_to` | **Resolved.** Probe loop is no longer generic over `Socket`; uses `DefaultSocket` concretely so Rust resolves to the unmetered inherent `tokio::net::UdpSocket` methods. `Socket` trait is no longer imported in `reference_ping.rs` (verified). |
| **Should Fix #4** — `peer_id` format mismatch | **Mostly resolved.** Format now matches `KnownPeerKeyLocation::Display`. **Open caveat** (new in round 2): for non-gateway nodes the listener fallback yields `0.0.0.0:<port>` until external addr is discovered, while other OTLP events read up-to-date `Ring::own_location()`. Documented in `p2p_impl.rs:362-380` and tracked in #4294 for a refresh path. |
| **Should Fix #5** — PR description stale | **Resolved.** Description updated; "Why not Socket trait?" replaced with the bypass justification. |
| **Should Fix #6** — Rebase onto current main | **Resolved.** Rebased; HEAD now sits cleanly on top of `c0d7fd04`. |
| **Should Fix #7** — Tracking issue for configurable target / opt-out | **Resolved.** #4294 filed and now also tracks the `peer_id` refresh path. |
| **Consider #5-#10** — stale rustdoc, transport.md heading, test count off-by-one, captive portal, 16-bit tx_id, etc. | All addressed in PR description / rustdoc / rule file, or accepted with documented rationale. |

---

### Round-2 CI findings — status

| CI finding | Status |
|---|---|
| `.unwrap()` on string-literal parse in production code | **Resolved.** Replaced with `Ipv4Addr::UNSPECIFIED` / `Ipv6Addr::UNSPECIFIED` direct construction — parse removed entirely (`reference_ping.rs:88-91`). |
| `transport.md` "Socket Abstraction Rules" missing `DefaultSocket` exception | **Resolved.** New "Exception: `DefaultSocket` for non-peer-to-peer external probes" subsection added with explicit guidance. |
| Info: unnecessary `Arc<RollingRttStats>` | **Resolved.** Plain owned value used now. |

---

### New findings in round 2

#### Must Fix
None.

#### Should Fix
None blocking. The leaf-node `peer_id` caveat is documented in code and tracked in #4294 — acceptable for Phase 1.5 (observation-only, gateways unaffected, leaves still get a stable per-node identifier even if it can't be joined to other events).

#### Consider (not blocking)
- **Bind-failure recovery path** is structurally pinned by `spawn_with_unbindable_target_does_not_exit` rather than functionally tested. The test comment is honest about this; a future regression in the `Err(e) => { warn!; pending().await; return }` arm would not be caught. Acceptable given the dense rule-comment block (`reference_ping.rs:104-120`) and the difficulty of deterministically forcing a bind failure.
- **Telemetry-enabled gate** in `p2p_impl.rs:386-388` is not unit-tested. A four-way matrix test (enabled × is_test_environment) would catch a future refactor that inverts or drops the gate.
- **DNS-response spoofing** with 16-bit tx_id remains acceptable for Phase 1.5 (observation-only) but should be revisited if Phase 2 controller logic ever reads these stats.

---

### Verdict

**State:** Ready to Merge — pending Unit & Integration / NAT / Simulation / Windows CI completion on HEAD 13323189.

All Must-Fix findings from round 1 are resolved. The two new round-2 CI warnings are fixed. Remaining findings are Consider-tier and explicitly tracked (#4294 for the peer_id refresh + opt-out work).

Per `no-merge-with-red-ci.md`: confirm green on the **current** HEAD (`13323189`) before merge — the green-from-2c018b29 record does not authorize merge of 13323189.

[AI-assisted - Claude]

sanity:
## Update: CI failure investigation

The Unit & Integration failure on the previous HEAD was `test_ping_multi_node` failing with a propagation race.

**Investigation outcome:** the test is currently flaky on `main` HEAD too — see #4295 for full reproduction details. Both branches fail it locally; the failure mode differs per run.

This is **not caused by this PR**, but the previous spawn-gate (`!is_test_environment`) was insufficient anyway: `is_test_environment` is set only via `--id` on the CLI, and the ping integration tests build `NodeConfig` directly without it. So even though my PR didn't cause the underlying flake, it WAS firing 1Hz DNS to Cloudflare from CI test peers, which is independently wrong.

**Fix pushed (ce83f3a2):** reference-ping is now opt-in via `telemetry.reference-ping-enabled` (default `false`). Production gateway configs set it to `true`; CI integration tests, simulation tests, and developer machines leave it off. Also addresses skeptical-M3 from the round-1 review ("every production node hits Cloudflare 1Hz by default" — no longer the case).

Awaiting CI on the new HEAD. If `test_ping_multi_node` fails again, that's #4295 reproducing, not this PR.

[AI-assisted - Claude]

Issue #2, opened 2021-09-01
NAT traversal


iduartgomez:
This is a general tracking issue to monitor the status of the different techniques that will help with clients behind NAT when it comes to establishing direct connection between peers. 

# Prelude 
One of the problems of NAT traversal is the variance in the network characteristics and configurations when trying to establish a connection between two peers, at every layer of the network and from the pov of both hw (router configuration, LAN, ISP, etc.) and software (for example, is a browser handling the connection for you or do you have direct access to the interface?). A good overview and taxonomy of the different scenarios can be read [here](https://github.com/libp2p/specs/blob/master/connections/hole-punching.md#vision).

Some important/interesting resources re. NAT traversal status within libp2p:
- High level overview: https://docs.libp2p.io/concepts/nat/
- Holistic specification for different hole punching techniques:  https://github.com/libp2p/specs/blob/master/connections/hole-punching.md
- Overall tracking issue for NAT traversal amongst the different libp2p implementations: https://github.com/libp2p/specs/issues/312
- Project flare (decentralized hole punching): https://github.com/protocol/web3-dev-team/pull/21
- Very good practical and theoretical read on the problems posed by NAT: https://tailscale.com/blog/how-nat-traversal-works/

# rust-libp2p status
Overall tracking issue in their repo: https://github.com/libp2p/rust-libp2p/issues/2052

- [ ] Automatic router configuration (e.g. UPnP). Not implemented, only an issue: https://github.com/libp2p/rust-libp2p/issues/558
- [x] [STUN-like](https://en.wikipedia.org/wiki/STUN) methods (hole-punching)
   - [x] Port reuse in TCP.  Implemented. We should enable this and should help.
   - [x] In theory the Identify protocol (which we will be using) helps with this since it provides the observed addresses to anyone querying, so they can communicate through those addresses back (this is handled automatically by libp2p AFAIK).
   - [x]  AutoNAT. Lets peers request dial-backs. (https://github.com/libp2p/rust-libp2p/pull/2262)
 - [x] [TURN-like](https://en.wikipedia.org/wiki/Traversal_Using_Relays_around_NAT) methods
    - https://docs.libp2p.io/concepts/circuit-relay/
    - [x] Circuit Relay v1 ([spec](https://github.com/libp2p/specs/blob/master/relay/circuit-v1.md)). Completed and we could use it in theory. The problem with TURN like strategies is one of scalability and cost though. Not a reliable long term solution probably.
[...]

PR #4297, opened 2026-05-27
feat(dashboard): transport metrics, ring stats, pub key, visual refresh

### LINKED ISSUE #3507: fix: improve <title> tag of dashboard
the current title tag is:
`<title>Freenet — Local Peer</title>`

### Problem
- the dash is an "EM Dash" which looks overly long on the browser tab
- usually there are a lot of open tabs and you don't see the full tab title
- therefore it should be short and concise and the beginning counts. 
- "Freenet ..." as a beginning can be anything, you dont know what it is, if you just see the first 4 chars

### Solutions
`<title>FN Peer</title>`

or just 

`<title>Peer</title> `   // The favicon makes it clear what peer it is

or

`<title>Dashboard</title>`   // The favicon makes it clear what dashboard it is - and Dash as a beginning is definitive

bad: starting with freenet:
`<title>Freenet ....</title> `
bc. on a squeezed narrow tab you have no idea what it is. could be Freenet home page

### Source

https://github.com/freenet/freenet-core/blob/a085374f2629c8dc654a8b21a9caf7fbeb19de16/crates/core/src/server/home_page.rs#L46

Basedfloppa:
## Problem

The local peer dashboard served at `:7509` was sparse and looked AI-generated:
- Only connection status, peer list, uploaded/downloaded, and operation counters
- No transport metrics (RTT, cwnd, throughput) exposed despite being collected internally
- No ring-level stats (connection count, hosted contracts) visible
- Local mode (`freenet local`) showed "Starting up…" forever because
  `network_status::init()` was never called
- Favicon/title didn't help identify the tab in a crowded browser (generic em-dash
  "Freenet — Local Peer" title, `#007FFF` blue accent matching countless AI templates)
- No mobile layout — cards overflowed on phones
- `<title>` started with "Freenet" which is invisible in squeezed tabs (#3507)

### Data plumbing
- `TransportMetrics::read_snapshot()` — non-consuming read of the same counters
  that `take_snapshot()` resets for the telemetry worker. Returns `TransportSnapshot`
  without zeroing atomics.
- `RingStatsProvider` — follows the existing `GovernanceProvider` /
  `SubscriptionProvider` closure pattern. Registered in `p2p_impl.rs` at node
  startup, captures `Arc<Ring>`, called on every homepage request. Provides
  `connection_count`, `hosted_contracts`, and `own_pub_key`.

### Dashboard rendering
- Status card: new 3-tile metrics row (ring peers, hosted contracts, connection
  attempts) below the health banner
- Header: public key with label and copy-to-clipboard button (uses existing
  `copyToClipboard`/`showToast` JS functions)
- Transfer card: extended with transfers ok/fail + avg time, peak throughput,
  RTT avg/min/max, cwnd avg/peak/min, LEDBAT slowdowns. Rendered after a 10s
  grace period even without traffic, to prevent flicker during auto-refresh
- Ring SVG: connection arc stroke-width now scales with bytes transferred;
  inner CONTRACTS ring hidden when governance has no flagged contracts

### Visual refresh

The previous design had three strong "AI-generated" markers that risk
alienating users: 
(1) `#007FFF` blue accent — the single most common
colour in AI-produced dark-mode dashboards (GitHub, Stripe, Vercel);
(2) atmospheric radial-gradient background — a literal template pattern;
(3) generic card-grid layout with visible borders — indistinguishable
from a component-library demo. The refresh trades these for a warmer,
less sterile palette (teal accent, warm grays, dot-grid texture) and
softer card styling (12px radius, reduced borders, subtle shadow).
The intent is not to be flashy but to look intentional rather than
auto-generated.

Changes:
- Accent `#007FFF` → `#0abab5` teal; warm gray palette replacing GitHub-dark
- Atmospheric radial gradients replaced with subtle dot-grid texture
- Font stacks extended with `Iosevka`, `JetBrains Mono`, `Space Grotesk`
- Cards: 12px border-radius, subtle shadow, reduced border opacity
- Title: `FN Peer` (recognisable in 4-char tab width, #3507)

### Mobile layout (768px + 400px breakpoints)

Added as forward-looking groundwork: Freenet is targeting Android in the
future, where the dashboard will be viewed on phone screens. Not an
immediate user need today, but getting the responsive scaffolding in
place now avoids a larger retro-fit later.

- Header wraps, pub-key label hidden, compact sizes
- Metrics row flex-wraps to 2-3 columns
- Peer table hides Sent/Recv, transfer stats stack vertically
- Ring SVG scales to 160px

### Local mode fix
- `node.rs::run_local_node` now calls `network_status::init()` before starting
  the HTTP server, so `get_snapshot()` returns `Some(...)` from the first request

## Testing

- `cargo fmt` — clean
- `cargo clippy -- -D warnings` — clean (fixed 2 `format_in_format_args`)
- `cargo test -p freenet --lib home_page` — 53 passed (updated favicon colour
  assertions to teal, added `ring_stats`/`transport_snapshot` fields to test
  fixture, added transfer-card grace-period tests)
- Manually verified on Orange Pi 5 (aarch64, network mode): all cards render
  with live data (3 peers, 753 hosted contracts, real pub key)
- Manually verified locally (x86_64, `freenet local`): dashboard loads
  immediately, no "Starting up…" stall

## Fixes

Closes #3507

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2010,6 +2010,15 @@ pub async fn run_local_node(
         )
     }
 
+    // Seed the dashboard so it renders immediately (not "Starting up…"
+    // forever). Local mode never joins the ring, so there are no peers,
+    // no contracts, and no transport stats.
+    crate::node::network_status::init(
+        socket.port,
+        std::collections::HashSet::new(),
+        crate::config::PCK_VERSION.to_string(),
+    );
+
     let (mut gw, mut ws_proxy) = crate::server::serve_client_api_in(socket).await?;
 
     // TODO: use combinator instead
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -42,7 +42,9 @@ pub type RingStatsProvider = Arc<dyn Fn() -> RingStatsSnapshot + Send + Sync + '
 /// Snapshot of ring-level statistics exposed to the dashboard.
 #[derive(Debug, Clone, Default)]
 pub struct RingStatsSnapshot {
+    /// Number of active ring connections.
     pub connection_count: u32,
+    /// Number of contracts this node is currently hosting.
     pub hosted_contracts: u32,
     /// Base58-encoded public key of this node.
     pub own_pub_key: String,
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -1304,4 +1304,19 @@ mod tests {
         assert_eq!(nat.recent_attempts(), 20);
         assert_eq!(nat.recent_successes(), 0);
     }
+
+    /// Regression: `init()` must seed the global so `get_snapshot()`
+    /// returns `Some(...)` immediately — without this, the dashboard
+    /// shows "Starting up…" forever in local mode (#3507 / PR #4297).
+    #[test]
+    fn init_populates_snapshot() {
+        let _lock = TEST_MUTEX.lock().unwrap();
+        init(31337, HashSet::new(), "0.1.0-test".to_string());
+        let snap = get_snapshot();
+        assert!(snap.is_some(), "snapshot should be Some after init");
+        let s = snap.unwrap();
+        assert_eq!(s.version, "0.1.0-test");
+        assert_eq!(s.listening_port, 31337);
+        assert_eq!(s.open_connections, 0);
+    }
 }
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -11,7 +11,7 @@ use std::time::Instant;
 
 use crate::ring::{PeerKeyLocation, SubscribedContractSnapshot};
 use crate::router::Router;
-use crate::transport::metrics::TRANSPORT_METRICS;
+use crate::transport::metrics::{TRANSPORT_METRICS, TransportSnapshot};
 
 static NETWORK_STATUS: OnceLock<Arc<RwLock<NetworkStatus>>> = OnceLock::new();
 static ROUTER: OnceLock<Arc<parking_lot::RwLock<Router>>> = OnceLock::new();
@@ -35,6 +35,19 @@ pub type SubscriptionProvider =
 /// reporters.
 pub type GovernanceProvider = Arc<dyn Fn() -> GovernanceSnapshot + Send + Sync + 'static>;
 
+/// Provider for live ring-level stats (connection count, hosted contracts).
+/// Same provider pattern as `SubscriptionProvider`/`GovernanceProvider`.
+pub type RingStatsProvider = Arc<dyn Fn() -> RingStatsSnapshot + Send + Sync + 'static>;
+
+/// Snapshot of ring-level statistics exposed to the dashboard.
+#[derive(Debug, Clone, Default)]
+pub struct RingStatsSnapshot {
+    pub connection_count: u32,
+    pub hosted_contracts: u32,
+    /// Base58-encoded public key of this node.
+    pub own_pub_key: String,
+}
+
 static GOVERNANCE_PROVIDER: parking_lot::RwLock<Option<GovernanceProvider>> =
     parking_lot::RwLock::new(None);
 
@@ -42,6 +55,19 @@ static GOVERNANCE_PROVIDER: parking_lot::RwLock<Option<GovernanceProvider>> =
 /// previously-registered provider.
 pub fn set_governance_provider(provider: GovernanceProvider) {
     *GOVERNANCE_PROVIDER.write() = Some(provider);
+}
+
+static RING_STATS_PROVIDER: parking_lot::RwLock<Option<RingStatsProvider>> =
+    parking_lot::RwLock::new(None);
+
+/// Register the dashboard's ring-stats data source.
+pub fn set_ring_stats_provider(provider: RingStatsProvider) {
+    *RING_STATS_PROVIDER.write() = Some(provider);
+}
+
+#[cfg(test)]
+pub(crate) fn clear_ring_stats_provider() {
+    *RING_STATS_PROVIDER.write() = None;
 }
 
 /// Clear the governance provider. Used by tests once they're added
@@ -445,6 +471,10 @@ pub struct NetworkStatusSnapshot {
     pub bytes_downloaded: u64,
     /// Overall node health level for the "everything looks good" indicator.
     pub health: HealthLevel,
+    /// Live ring-level statistics: connection count, hosted contracts.
+    pub ring_stats: RingStatsSnapshot,
+    /// Period transport metrics (current values, not reset on read).
+    pub transport_snapshot: TransportSnapshot,
     /// Per-contract governance state (Phase 4). The dashboard's
     /// verdict block, contracts table, ring inner-ring, and
     /// distribution histogram all read from this. Empty when the
@@ -759,6 +789,14 @@ pub fn get_snapshot() -> Option<NetworkStatusSnapshot> {
     let bytes_uploaded = TRANSPORT_METRICS.cumulative_bytes_sent();
     let bytes_downloaded = TRANSPORT_METRICS.cumulative_bytes_received();
 
+    let ring_stats = RING_STATS_PROVIDER
+        .read()
+        .as_ref()
+        .map(|provider| provider())
+        .unwrap_or_default();
+
+    let transport_snapshot = TRANSPORT_METRICS.read_snapshot();
+
     Some(NetworkStatusSnapshot {
         failures,
         connection_attempts: s.connection_attempts,
@@ -787,6 +825,8 @@ pub fn get_snapshot() -> Option<NetworkStatusSnapshot> {
         bytes_uploaded,
         bytes_downloaded,
         health,
+        ring_stats,
+        transport_snapshot,
         governance: GOVERNANCE_PROVIDER
             .read()
             .as_ref()
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -46,7 +46,9 @@ pub struct RingStatsSnapshot {
     pub connection_count: u32,
     /// Number of contracts this node is currently hosting.
     pub hosted_contracts: u32,
-    /// Base58-encoded public key of this node.
+    /// Short base58 Peer ID (12-byte prefix) — how other nodes see this peer.
+    pub peer_id: String,
+    /// Base58-encoded full 32-byte X25519 public key.
     pub own_pub_key: String,
 }
 
@@ -1355,5 +1357,47 @@ mod tests {
         assert_eq!(s.version, "0.1.0-test");
         assert_eq!(s.listening_port, 31337);
         assert_eq!(s.open_connections, 0);
+    }
+
+    /// Registered RingStatsProvider must surface in `snap.ring_stats`,
+    /// replacing the provider must take effect immediately, and the
+    /// no-provider default is all-zeros.
+    #[test]
+    fn ring_stats_provider_round_trip() {
+        let _lock = TEST_MUTEX.lock().unwrap();
+        // Use the same init as other tests in this module.
+        init(31337, HashSet::new(), "0.1.0-test".to_string());
+
+        // 1. No provider — default.
+        let snap = get_snapshot().unwrap();
+        assert_eq!(snap.ring_stats.connection_count, 0);
+        assert!(snap.ring_stats.own_pub_key.is_empty());
+
+        // 2. Register provider.
+        set_ring_stats_provider(Arc::new(|| RingStatsSnapshot {
+            connection_count: 42,
+            hosted_contracts: 7,
+            peer_id: "abc".to_string(),
+            own_pub_key: "test-key".to_string(),
+        }));
+        let snap = get_snapshot().unwrap();
+        assert_eq!(snap.ring_stats.connection_count, 42);
+        assert_eq!(snap.ring_stats.peer_id, "abc");
+        assert_eq!(snap.ring_stats.own_pub_key, "test-key");
+
+        // 3. Replace provider — must take effect immediately.
+        set_ring_stats_provider(Arc::new(|| RingStatsSnapshot {
+            connection_count: 99,
+            hosted_contracts: 1,
+            peer_id: "xyz".to_string(),
+            own_pub_key: "new-key".to_string(),
+        }));
+        let snap = get_snapshot().unwrap();
+        assert_eq!(snap.ring_stats.connection_count, 99);
+        assert_eq!(snap.ring_stats.peer_id, "xyz");
+        assert_eq!(snap.ring_stats.own_pub_key, "new-key");
+
+        // 4. Clean up.
+        clear_ring_stats_provider();
     }
 }
--- a/crates/core/src/node/p2p_impl.rs
+++ b/crates/core/src/node/p2p_impl.rs
@@ -150,6 +150,23 @@ impl NodeP2P {
             ring.dashboard_governance_snapshot()
         }));
 
+        // Wire live ring stats for the dashboard: connection count +
+        // hosted contracts + own public key, read on every homepage
+        // request.
+        let ring_stats = self.op_manager.ring.clone();
+        super::network_status::set_ring_stats_provider(std::sync::Arc::new(move || {
+            let own_pub_key = ring_stats
+                .connection_manager
+                .own_location()
+                .pub_key()
+                .to_string();
+            super::network_status::RingStatsSnapshot {
+                connection_count: ring_stats.connection_manager.connection_count() as u32,
+                hosted_contracts: ring_stats.hosting_contracts_count() as u32,
+                own_pub_key,
+            }
+        }));
+
         // Emit peer startup event
         if let Some(event) = crate::tracing::NetEventLog::peer_startup(
             &self.op_manager.ring,
--- a/crates/core/src/node/p2p_impl.rs
+++ b/crates/core/src/node/p2p_impl.rs
@@ -155,14 +155,17 @@ impl NodeP2P {
         // request.
         let ring_stats = self.op_manager.ring.clone();
         super::network_status::set_ring_stats_provider(std::sync::Arc::new(move || {
-            let own_pub_key = ring_stats
-                .connection_manager
-                .own_location()
-                .pub_key()
-                .to_string();
+            let (peer_id, own_pub_key) = {
+                let pk = ring_stats.connection_manager.own_location();
+                let key_bytes = pk.pub_key().as_bytes();
+                let peer_id = pk.pub_key().to_string(); // 12-byte Display
+                let own_pub_key = bs58::encode(key_bytes).into_string(); // full 32-byte
+                (peer_id, own_pub_key)
+            };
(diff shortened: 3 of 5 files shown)


Basedfloppa:
image
image
image

iduartgomez:
This PR overreaches relative to #3507.

#3507 scope: one-line `<title>` change (label `E-easy`), targeting `home_page.rs:46`.

This PR: +540/-74 across 5 files, bundling:

1. **Title tag fix** (#3507) — ~5 lines, the actual issue
2. **`TransportMetrics::read_snapshot()`** — new public API on transport metrics
3. **`RingStatsProvider`** — new closure + wiring in `p2p_impl.rs`
4. **Local-mode `network_status::init()` fix** — separate real bug, deserves own `fix:` PR with regression test
5. **Visual refresh** — palette/fonts/dot-grid/card styling, subjective taste change
6. **Mobile breakpoints** — justified as "forward-looking groundwork" for hypothetical Android client; violates AGENTS.md "Don't design for hypothetical future requirements"

Problems with the bundle:
- `Closes #3507` is misleading — the title-tag fix is ~1% of the diff.
- Review burden inflated: reviewer signs off on transport API surface + ring provider + local-mode bug + CSS overhaul + mobile layout in one pass.
- Local-mode init fix is a real bug currently riding under `feat:`, which skips the regression-test requirement that a `fix:` PR would enforce.
- Mobile layout is speculative — no Android client today.

Suggested split:
1. `fix(dashboard): shorten <title> for narrow tabs` — closes #3507
2. `fix(dashboard): init network_status in local mode` — with regression test
3. `feat(dashboard): expose transport + ring metrics` — real new surface, worth focused review
4. `style(dashboard): visual refresh` — easy to revert if disliked
5. Drop mobile breakpoints until Android client exists

sanity:
I thought this might overlap with some work I'm doing but I don't think it does so no blocker from me.

iduartgomez:
@Basedfloppa we will merge this, but please from now on if you are adding features/changing behavior (so no fixes/performance improvements etc.), open an issue first for approval before opening PR.

iduartgomez (commented):
## Comprehensive PR Review: #4297

### Summary
- **PR Title:** feat(dashboard): transport metrics, ring stats, pub key, visual refresh
- **Type:** feat (bundles a fix and a style change — see Big Picture)
- **CI Status:** all required checks passing
- **Linked Issues:** Closes #3507 (one-line `<title>` change; PR is ~540 lines beyond that)
- **Review tier:** Full — touches transport metrics (atomics), node startup wiring, dashboard surface
- **Reviewers run:** code-first, testing, skeptical, big-picture. External model pass (codex) skipped — auth token expired locally; no gemini CLI available.

---

### Code-First Analysis

**Independent Understanding:** Four loosely-coupled changes bundled:
1. `TransportMetrics::read_snapshot()` non-resetting read of existing atomics (metrics.rs:407-464).
2. `RingStatsProvider` closure pattern mirroring `GovernanceProvider` / `SubscriptionProvider`, registered in `p2p_impl.rs:153-167`, supplying `connection_count`, `hosted_contracts`, `own_pub_key` to `get_snapshot()`.
3. `network_status::init()` called inside `run_local_node` (node.rs:2145-2149) so the dashboard renders immediately in local mode.
4. Dashboard rewrite in `home_page.rs` (+362/-72): new transfer card with RTT/cwnd/slowdowns, pub-key + copy button in header, palette swap (`#007FFF` → `#0abab5`), `<title>FN Peer</title>` (closes #3507), mobile breakpoints, inner-ring gating.

**Stated Intent:** Matches the code at a structural level.

**Alignment:** Partial. Several material semantics changes are not documented in either the PR description or doc comments:

**Gaps:**
- The PR description says the new card shows "peak throughput, RTT avg/min/max, cwnd avg/peak/min". When the telemetry worker is enabled, those atomics are reset every `transport_snapshot_interval_secs` (≥10s — `tracing/telemetry.rs:84,338`). Dashboard refreshes every 5s, so half of all refreshes will read counters that were just zeroed. The "peak"/"min" tiles will visibly oscillate. With telemetry disabled, the same counters are lifetime-cumulative. The dashboard label and the `network_status.rs:478` doc comment ("current values, not reset on read") obscure both modes. See Should Fix #1.
- The "10s grace period" in `build_transfer_card` (home_page.rs:467) is not just flicker-prevention: the old gate `open_connections == 0` is gone, so a disconnected peer that was up >10s now shows a permanent zeroed card.
- In `run_local_node`, `RingStatsProvider` is never registered (only `p2p_impl.rs` does this). `RingStatsSnapshot::default().own_pub_key == ""`. `home_page.rs:43` falls back to `"?"` only when the whole snapshot is `None`, so the local-mode header renders an empty `<code>` element and a copy button that copies `""`. See Must Fix #2.
- Title change is real and intentional (#3507) but unlocked — no test asserts the new `<title>FN Peer</title>` string.

---

### Testing Assessment

**Coverage Level:** Insufficient for the new provider+metrics surface.

| Unit | ⚠️ | `read_snapshot` only tested for the zero/sentinel case (metrics.rs:1052). `RingStatsProvider` wiring untested. Transfer-card sub-sections untested (every fixture uses `TransportSnapshot::default()` → only the empty path of every `if x > 0` conditional renders in tests). |
| Integration | ⚠️ | `init_populates_snapshot` (network_status.rs:1349) pins `init` → `get_snapshot` but not `run_local_node` → HTTP. |
| Simulation | N/A | Dashboard surface. |
| E2E | ❌ | Manual verification only (Orange Pi + local x86_64 per PR body). |

**Regression Test:** `init_populates_snapshot` covers the *direct* bug (init seeds the global) but not the *call-site* bug (`run_local_node` must call `init` before `serve_client_api_in`).

**Missing Tests:**
1. `read_snapshot` happy-path with recorded `cwnd_sample` / `rtt_sample` / `transfer_completed`, asserting non-sentinel averages and peaks. (Rule-bot already flagged this.)
2. `RingStatsProvider` round-trip mirroring `test_subscription_provider_wired_renders_contracts` (network_status.rs:1146) — including the "replaceable, not OnceLock" contract and the no-provider-default case.
3. Transfer-card conditional branches (RTT, cwnd, slowdowns, transfers strings).
4. Pub-key render → HTML string assertion (also locks the `html_escape` call at line 117).
5. `<title>FN Peer</title>` lock test.

---

### Skeptical Findings

**Risk Level:** Medium-High — one user-facing misrepresentation (pub key) and one operator-visible UX inversion (telemetry-window oscillation).

| `own_pub_key` is a 12-byte prefix, not the full public key | **High** | `transport/crypto.rs:290-299` (verified), surfaced at `home_page.rs:65-66` | `TransportPublicKey::Display` is `bs58::encode(&self.0.as_bytes()[..12])`. Dashboard labels it "Public key" and the copy button toast says "Public key copied". Users pasting that into anything expecting a real key get a truncated identifier. Either relabel to "Peer ID" / "Key prefix" or render the full key. |
| `read_snapshot` semantics flip on telemetry config | **High** | `transport/metrics.rs:407-464`, `tracing/telemetry.rs:338` | With telemetry on: 5s dashboard polls a counter set the 10s+ worker resets via `swap(0,…)`. "Peak", "min", and averages oscillate to zero. With telemetry off: counters are lifetime-cumulative — a single 10s RTT spike skews `avg_rtt_us` for the life of the process. Neither mode is what the dashboard label promises. Fix: separate cumulative accumulators for the dashboard, or document the dependency on the dashboard itself. |
| `RingStatsProvider` not wired in `run_local_node` → empty pub-key in local mode | **Medium** | `node.rs:2131-2151`, `home_page.rs:40-43` | `home_page.rs:43` only falls back to `"?"` when `snap` is `None`, not when `own_pub_key` is `""`. Either register a local-mode provider returning a static pub key (the node has one even in local mode) or change the fallback to `.filter(\|s\| !s.is_empty()).unwrap_or("?")`. |
| Torn read of `_sum` / `_samples` pair | **Medium** | `transport/metrics.rs:414-415, 419-420` | `record_rtt_sample` updates the sum and the sample count in two separate atomic ops. A concurrent `read_snapshot()` between them returns either a high average (sum updated, count not) or `0` (count incremented, sum not yet) depending on order. Acceptable for periodic telemetry; visible at per-HTTP-request cadence on low sample counts. |
| Per-request lock acquisition under multi-tab refresh | **Low** | `connection_manager.rs:776, 793, 1327-1334`, `hosting.rs:1029-1031` | The provider closure takes three `parking_lot::RwLock::read()` + one `Mutex::lock()` per homepage GET. Read locks don't block other readers but DO block writers (peer connect/disconnect, hosting cache mutation). With multiple browser tabs auto-refreshing, this is unbounded. Not blocking; worth memoizing with a short TTL if it becomes a problem. |
| `init()` is silently a no-op on the second call | **Low** | `network_status.rs:244-261` | `OnceLock::set` returns `Err` if initialized; the code explicitly ignores it. Today only one of `run_local_node` / `start_event_listener` ever runs in a process, so this is fine. A future mixed-mode scenario (or any test that re-inits) silently wins-by-first-call, including `gateway_addresses` (empty in local-init → all peers flagged `is_gateway = false`). Add `tracing::debug!` on the err path so the misconfig is observable. |
| Stale function name `favicon_blue_when_connected` | **Low** | `home_page.rs:3805` | Function now asserts the teal hex (`%230abab5`). Rename to match the new color. |
| Copy button copies `"?"` when pub-key is unavailable | **Low** | `home_page.rs:43-66` | Toast says "Public key copied" — copies the literal `?`. Hide the button when the value is the placeholder. |

What I verified clean:
- `pub_key` is HTML-escaped before substitution (`home_page.rs:117`). No XSS via pub-key.
- SVG stroke-width formula at `home_page.rs:751-761` cannot produce NaN/inf (`max_xfer.max(1)`).
- `html_escape` covers `&`, `<`, `>`, `"` (network_status.rs:901-906).
- `run_local_node` calls `init()` before `serve_client_api_in().await` (node.rs:2145-2151) — no race window.
- `clipboard` copies `document.getElementById('pub-key').textContent`, not an interpolated string — safe even if pub-key encoding changes.

---

### Big Picture Assessment

**Goal Alignment:** Partial. Linked issue #3507 is a one-line `<title>` change labeled `E-easy`. The actual PR is ~540/-74 across 5 files bundling six concerns (title fix, `read_snapshot`, `RingStatsProvider`, local-mode init fix, visual refresh, mobile breakpoints). Scope creep was already raised by @iduartgomez and acknowledged by @sanity — merge proceeds; comments below are documentation of the rule loophole this PR exposes, not a re-litigation of that decision.

**Anti-Patterns Detected:**
- **`feat:` wrapper hides a `fix:`.** The local-mode init at `node.rs:2145-2149` is a real bug fix riding under a `feat:` title. CI rule #5 (`fix:` PRs require a new test function) only matches the PR title regex and is bypassed. A regression test (`init_populates_snapshot`) is present anyway, but the enforcement gap is reproducible by future PRs that forget the test.
- **Hypothetical-future design.** Mobile breakpoints at `home_page.rs:1969-2017` (`@media (max-width: 768px)` and `(max-width: 400px)`) are explicitly justified in the PR body as groundwork for an Android client that does not exist today. Violates AGENTS.md "Don't design for hypothetical future requirements". Cost is CSS-only, will likely bit-rot.
- **Font-stack overclaim.** CSS at `home_page.rs:1430-1431` lists `Iosevka`, `JetBrains Mono`, `Space Grotesk` with no `@font-face` and no bundling. Users without those fonts get the system fallback. PR body's "Font stacks extended with…" oversells.

**Removed Code Concerns:**
- `transfer_card_hidden_when_no_data` was renamed to `transfer_card_hidden_when_fresh_start` rather than `#[ignore]`'d with a comment referencing this PR. Per `.claude/rules/git-workflow.md` ("Is the test superseded by new semantics? → Add #[ignore]"), the rule is to keep the old assertion as historical documentation. The historical invariant ("hide when `open_connections == 0`") is now permanently lost from the test suite — a future PR that re-introduces a connection-count gate will not trip CI. Should Fix #5.
- No other tests removed. No silent reverts of prior fixes. Test rename for the favicon is balanced (`favicon_blue_takes_priority_over_failures` → `favicon_connected_overrides_failures`).

**Scope Assessment:** Significant creep relative to the linked issue. Already discussed and accepted on the PR.

---

### Documentation
- Code docs: incomplete. The dashboard's dependence on telemetry-worker cadence (Skeptical #2) is undocumented in both `transport/metrics.rs:405` and `network_status.rs:478`.
- Architecture docs: n/a (no transport API contract docs claim invariants this PR breaks).
- User docs: n/a.
- Rule docs: `.claude/rules/git-workflow.md` rule #5 enforcement gap (`feat:` carrying `fix:`) demonstrated; consider tightening per Recommendations.

---

### Recommendations

#### Must Fix (Blocking)

1. **Relabel or fix the "Public key" field.** It is the 12-byte prefix of the X25519 key (`transport/crypto.rs:290-299`). Options: (a) rename the UI label and copy-toast to "Peer ID" / "Key prefix"; (b) introduce a full-key encoder and use that instead. As shipped, the dashboard misrepresents the value and the copy button propagates the misrepresentation into external systems.

2. **Wire `RingStatsProvider` in `run_local_node`, or harden the fallback.** Today `home_page.rs:43` falls back to `"?"` only when the whole snapshot is `None`. With local-mode init succeeding but the provider unregistered, `own_pub_key == ""` renders an empty `<code>` and a copy button that copies the empty string. Either register a local-mode-appropriate provider, or change the fallback to `.filter(\|s\| !s.is_empty()).unwrap_or("?")` and hide the copy button on the placeholder.

#### Should Fix (Important)

1. **Document and/or fix the telemetry-window dependency of `read_snapshot`.** `transport/metrics.rs:407-464` reads atomics that `take_snapshot` resets every ≥10s when telemetry is on. Either: (a) introduce parallel non-resetting cumulative accumulators for the dashboard fields; (b) update the docstring at `metrics.rs:405` and the field comment at `network_status.rs:478` to say "interval depends on `transport_snapshot_interval_secs` when telemetry is enabled; lifetime-cumulative otherwise"; (c) surface the active window on the dashboard tile so operators know what "peak" means.

2. **Add a `read_snapshot` happy-path test** in `transport/metrics.rs`. Existing test (`metrics.rs:1052`) only covers the sentinel-to-zero mapping. Record known `cwnd_sample` / `rtt_sample` / `transfer_completed` values, assert returned averages and peaks. Rule-bot already flagged this.

3. **Add a `RingStatsProvider` round-trip test** mirroring `test_subscription_provider_wired_renders_contracts` (`network_status.rs:1146`). Pin: (a) registered provider values surface in `snap.ring_stats`; (b) replacing the provider takes effect immediately; (c) no-provider-default behaviour.

4. **Add transfer-card sub-section assertions.** Every existing fixture uses `TransportSnapshot::default()`; the `if avg_rtt_us > 0` / `cwnd_str` / `slowdown_str` / `xfer_str` branches in `home_page.rs:482-534` never render in tests.

5. **Restore the old `transfer_card_hidden_when_no_data` test as `#[ignore]`** per `.claude/rules/git-workflow.md`. Cite this PR in the comment. The new test stays; the old one documents the historical invariant.

6. **Decide whether the `feat:` title is correct.** The bundled local-mode init fix is a real bug (user-visible "Starting up… forever"). Either retitle the PR `fix:` (and let CI rule #5 verify the regression test is present, which it is), or accept that the convention deliberately bypasses CI rule #5 for dashboard bundles. The follow-up rule-tightening below addresses the loophole if the convention is kept.

#### Consider (Suggestions)

1. **Tighten CI rule #5** in `.github/workflows/ci.yml:278-290` to scan commit subjects (`git log $BASE..$HEAD --format=%s`) for `^fix(\(.*\))?:` and require a new test function if any match. Add a one-line note to `.claude/rules/git-workflow.md` explaining the bundled-fix-in-feat-PR case. This PR is the worked example.

2. **Rename `favicon_blue_when_connected`** (`home_page.rs:3805`) — the assertion is teal.

3. **Bound the per-request lock acquisitions in `RingStatsProvider`** with a short-TTL memoization if multi-tab dashboards become a measured problem on busy gateways.

4. **Make `init()` log on the duplicate-init err path** (`network_status.rs:259-260`) so the silent-no-op stops being silent if the local/network mode boundary ever blurs.

5. **Remove the speculative mobile breakpoints** until an Android target consumes them. Cost is low but they will bit-rot.

6. **Drop or bundle the unused fonts** (`Iosevka`, `JetBrains Mono`, `Space Grotesk`) from the CSS stack.

7. **Lock the `<title>FN Peer</title>` value** with a test assertion so future renames are caught.

8. **Hide the copy button when pub-key is the placeholder `?`** to avoid copying the literal `?` with a "Public key copied" toast.

---

### Verdict

**HEAD SHA reviewed:** `8e8e3a94868be33f67e47884d26c708f56f69568`

The two Must-Fix items (12-byte-prefix mislabel; empty pub-key in local mode) are user-facing correctness issues, not style. The first misrepresents the field's actual contents to anyone who copies the value; the second renders the new feature broken in the exact mode the local-mode init fix was meant to unblock. Both are small, isolated changes. After they land, the Should-Fix testing gaps and the `read_snapshot` documentation are the next priority.

If maintainers choose to merge as-is given the prior scope discussion, the Must-Fix items should land as immediate follow-up PRs and the rule-tightening (Recommendations #1 under Consider) is the durable fix for the `feat:`-carrying-`fix:` pattern this PR exposed.

[AI-assisted - Claude]

Basedfloppa:
I apologize for such a big pr with no issue to discuss it before hand, i got too carried away when adding pubkey and small title fix.

Also i think an explanation of what was added/changed and why is needed:

## New data

- **Transport metrics** in Transfer card: RTT avg/min/max, cwnd avg/peak/min,
  LEDBAT slowdowns, transfers ok/fail + avg time, peak throughput
- **Ring stats** in Status card: ring peers, hosted contracts, connection attempts
- **Peer ID + Pub key** in header: short 12-byte Peer ID + full base58 X25519 key,
  separate copy buttons
- **Governance table**: sortable columns, copy buttons per contract hash
- `TransportMetrics::read_snapshot()` — non-resetting read (unlike `take_snapshot`)
- `RingStatsProvider` — closure provider pattern (mirrors `GovernanceProvider`)

### Why

Freenet exports a lot of internal data that wasn't accessible from the operator-facing dashboard. These new metrics surface connection counts, hosted contracts, transport stats, and node identity — all previously buried in logs.
Before this, operators had no easy way to exchange public keys when linking gateways. The Peer ID and Pub key in the header, with copy buttons, provide that.
The governance table lacked the UX affordances already present in the peers and contracts tables — hash copy buttons and column sorting.

## Visual refresh

- Accent `#007FFF` → `#0abab5` (teal)
- Background: dot-grid texture replaces AI-template radial gradients
- Fonts: `Iosevka`, `JetBrains Mono`, `Space Grotesk` in stacks
- Cards: border-radius 12px, subtle box-shadow, reduced border opacity
- Title: `Freenet — Local Peer` → `FN Peer` (#3507)
- Favicon: teal instead of blue for connected state

### Why

The previous design was reported as "AI-looking" — a negative signal that could push users away. This refresh moves away from the sterile high-contrast blue accent toward a warmer, more intentional palette.

## Mobile

- `@media (max-width: 768px)`: header wrap, metrics 2-3 columns, peer table
  hides Sent/Recv, transfer stats stack vertically, ring SVG 160px
- `@media (max-width: 400px)`: compact metrics, smaller fonts

### Why

Freenet users can already access the dashboard from mobile devices (at least i do open it therefrom time to time). These responsive breakpoints provide a usable phone layout as groundwork.

Issue #3, opened 2021-09-11
Decentralized email proof-of-concept


sanity:
# Goal

* Implement a minimalist messaging app on Locutus
* Use #246 to prevent spam

iduartgomez:
Since we are keeping track of the milestone in pivotaltracker closing this tracking issue.

PR #4304, opened 2026-05-28
feat(server): auto-refresh contract web page on transient GET errors

### LINKED ISSUE #3472: fix: automatic page refresh if initial loading of contract fails
### Problem
Starting a new, fresh, empty Freenet node and requesting River i get a lot of errors like: 

- error while executing operation in the network: GET operation timed out
- GET request timed out after 30s

and the page stops any further attempt to load the contract and dies.
After a while, when the peer has better or other connections, the contract does load, but the page re-load must be initiated by the user.

### Solution

With the error msg during initial load add to `<head> `
`<meta http-equiv="refresh" content="60">`

and tell the user: 
**GET request timed out after 30s - reloading in 30s**

Make sure to remove the refresh once the page loads the contract.

Basedfloppa:
## Problem

Starting a fresh Freenet node and requesting a contract web app (e.g.
River) often fails with "GET operation timed out" or "operation timed
out". The browser shows a dead error page and the user must manually
reload.

After a while the node builds enough ring connections to fetch the
contract, but the page never retries — the user has to hit F5.

## Solution

When `contract_home` fails with a **transient** error (`FailedOperation`
or `OperationError`), return a `503 Service Unavailable` page containing
`<meta http-equiv="refresh" content="60">` instead of a permanent error.

The page auto-reloads every 60 seconds. Once the contract is cached
locally, `contract_home` serves normally — the success response has no
`<meta refresh>`, so the retry loop stops naturally.

Non-transient errors (`InvalidParam`, `MissingContract`, `Contract not
found`) still return their usual status codes without retry.

This follows the existing pattern used by `EmptyRing` / `PeerNotJoined`,
which already return a `connecting_page()` with `<meta refresh>`.

### Error classification

| Error | Status | Behaviour |
|-------|--------|-----------|
| `FailedOperation` | 503 | Auto-refresh every 60s |
| `OperationError { .. }` | 503 | Auto-refresh every 60s |
| `EmptyRing` | 503 | Redirect to dashboard (existing) |
| `PeerNotJoined` | 503 | Redirect to dashboard (existing) |
| `InvalidParam` | 400 | Error text (no retry) |
| `MissingContract` | 404 | Error text (no retry) |
| `NodeError("Contract not found")` | 404 | Error text (no retry) |
| Other `NodeError` | 500 | Error text (no retry) |

## Testing

- `cargo fmt` — clean
- `cargo clippy -- -D warnings` — clean
- `cargo test -p freenet --lib errors::tests` — 5 passed
  - `failed_operation_returns_retry_page` — 503
  - `operation_error_returns_retry_page` — 503
  - `other_axum_error_returns_internal_server_error` — 500 (unchanged)
  - `empty_ring_returns_service_unavailable` — unchanged
  - `peer_not_joined_returns_service_unavailable` — unchanged

## Fixes

Closes #3472


iduartgomez (commented):
## Comprehensive PR Review: #4304

### Summary
- **PR Title:** feat(server): auto-refresh contract web page on transient GET errors
- **Type:** feat
- **CI Status:** passing (all required checks green)
- **Linked Issues:** #3472
- **Review tier:** Light (89/17-line single-file diff in server-side error rendering; not crypto/concurrency/wire-format)
- **Reviewers run:** skeptical-reviewer; external codex pass attempted but failed (auth expired); manual verification of key claims

---

### Code-First Analysis
**Independent Understanding:** Adds an `is_transient` matcher on `WebSocketApiError::AxumError { ErrorKind::FailedOperation | OperationError { .. } }`. Matching variants → 503 + HTML page with `<meta http-equiv="refresh" content="60">`. Other variants keep existing 400/404/500 plain-text behavior. Two new unit tests assert status code only.

**Stated Intent:** PR body says "fetch operation timed out" / "GET operation timed out" cases should auto-refresh until the contract is cached.

**Alignment:** **Misaligned.** The PR's stated user-visible bug (`"GET operation timed out"`) does **not** flow through `AxumError { FailedOperation | OperationError }`. It flows through `WebSocketApiError::NodeError { error_cause: "GET request timed out after 30s" }` (`crates/core/src/server/path_handlers.rs:188`), which the matcher does not touch and which therefore still renders as 500. The retry page only fires when the node-side GET op returns an error fast enough to come back inside the 30s wrapper timeout.

**Gaps:**
- `ErrorKind::FailedOperation` is **never raised anywhere in freenet-core** (only defined in stdlib `client_events.rs:131`). Half the `is_transient` arm is dead.
- The actual dominant transient case (30s wall-clock elapsed) is `NodeError`, not `AxumError`.

---

### Testing Assessment
**Coverage Level:** insufficient relative to PR claims.

| Unit | ⚠️ | Two new tests assert only `status == 503`. Neither reads the body, so a refactor that drops the `<meta refresh>` would pass. |
| Integration | ❌ | No test exercises the end-to-end `contract_home → 30s timeout → IntoResponse` path that the PR purports to fix. |
| Simulation | N/A | |
| E2E | N/A | |

**Regression Test:** present for the variants it claims to cover, but **does not cover the bug actually reported in #3472** (the 30s wall-clock timeout path).

**Missing Tests:**
- Body assertion: `assert!(body.contains(r#"http-equiv="refresh""#))`.
- Test for the `NodeError "GET request timed out after 30s"` path returning the retry page (after the fix below).

---

### Skeptical Findings
**Risk Level:** medium (correctness gap — feature does not fully address the linked issue).

| Wall-clock timeout path bypasses `is_transient` | **high** | `crates/core/src/server/path_handlers.rs:187-190`, `errors.rs:77` | 30s `tokio::time::timeout` elapse → `NodeError { "GET request timed out after 30s" }` → falls through to 500. The dominant "still trying, sparse ring" UX still renders as a permanent error. Issue #3472 only partially fixed. |
| `ErrorKind::FailedOperation` never raised | **high** | `errors.rs:77` | Grep across `crates/` and stdlib 0.6.0 shows zero raisers. Branch is dead code; only `OperationError { .. }` is reachable. |
| `is_transient` matcher incomplete | medium | `errors.rs:73-80` | `NodeUnavailable` (`crates/core/src/server/client_api.rs:502`), `ChannelClosed`, `TransportProtocolDisconnect`, `RequestError(Timeout)` are also transient but render as 500. |
| `#[non_exhaustive]` enum + hard-coded allow-list | medium | `errors.rs:73-80` | `ErrorKind` is `#[non_exhaustive]`; new transient variants in future stdlib bumps silently land in the 500 bucket. (See `.claude/rules/git-workflow.md` non_exhaustive section.) |
| No `Cache-Control: no-store` / `Retry-After` on 503 | medium | `errors.rs:107-110` | Some intermediaries cache 503 heuristically; service workers / reverse proxies could pin a stale retry page. Pre-existing on `connecting_page()` too, but newly relevant. |
| 60s browser refresh decoupled from node-side timeouts | low | `errors.rs:160` (HTML literal) | Outer wrapper 30s, inner related-contracts 10s. Browser-side 60s can fire a second `NewConnection`/GET while node-side ops are still retrying, amplifying gateway load. |
| Tests assert status only, not body | low | `errors.rs:205-223` | Already flagged by automated rule-review; confirmed. |
| `OperationError { cause }` dropped on 503 path | low | `errors.rs:81-83` | Cause string lost. A `tracing::info!` would help operators distinguish "loading" from "broken". |
| Magic `60` embedded in HTML literal | low | `errors.rs:160` | Already flagged by automated rule-review. Extract as `const RETRY_REFRESH_SECS: u64 = 60;`. |
| Non-browser callers regress | low | `errors.rs:81-83` | Programmatic consumers of `contract_home` previously got plain-text error body on 500; now get HTML on 503. Grep `apps/` if any non-browser caller exists. |

---

### Big Picture Assessment
**Goal Alignment:** partial. The mechanism is correct (503 + meta-refresh is the right pattern, and matches the existing `connecting_page` precedent). The matcher misses the actual error path that produces the user-reported timeout, so #3472 is not closed in practice.

**Anti-Patterns Detected:**
- Dead allow-list arm (`FailedOperation` never raised).
- Hard-coded match on `#[non_exhaustive]` enum without a wildcard arm (the `bug-prevention-patterns.md` / `git-workflow.md` non_exhaustive idiom).

**Removed Code Concerns:** Existing `other_axum_error_returns_internal_server_error` test was repurposed (variant swapped to `Unhandled`). Behavior is preserved.

**Scope Assessment:** focused — single file, single concept. Scope is right; the gap is that the fix targets the wrong error variants.

---

### Documentation
- Code docs: present (`retry_loading_page` has a doc comment). Comment at `errors.rs:73-76` claims to fix "fetch operation timed out" but does not, given the routing above.
- Architecture docs: n/a
- User docs: n/a

---

### Recommendations

#### Must Fix (Blocking)
1. **Cover the 30s wall-clock timeout path.** Either (a) change `handle_get_response` (`path_handlers.rs:187-193`) to emit `WebSocketApiError::AxumError { error: ErrorKind::OperationError { cause: "GET request timed out after 30s".into() } }` instead of `NodeError`, or (b) add an explicit `WebSocketApiError::NodeError { error_cause } if error_cause.contains("timed out") || error_cause.contains("channel closed") => retry page` arm. Without this, the PR does not fix the user-reported behavior in #3472. Verify by re-reading `path_handlers.rs:187-193` against any fix.
2. **Add an integration/unit test exercising the actual `ensure_contract_cached` timeout path** — drive `handle_get_response(_, Err(elapsed))` and assert 503 + meta-refresh body. Current tests only cover variants that are not produced by the reported failure.
3. **Drop `FailedOperation` from the matcher or document it as defensive.** Currently dead — `git grep FailedOperation` shows zero raisers in core. If kept, add a `// defensive: never currently raised in core, kept for future stdlib raisers` comment.

#### Should Fix (Important)
1. Widen `is_transient` to include `ErrorKind::NodeUnavailable`, `ErrorKind::ChannelClosed`, `ErrorKind::TransportProtocolDisconnect`, and `ErrorKind::RequestError(RequestError::Timeout)` — all transient during cold-start, all currently rendered as 500.
2. Add `Cache-Control: no-store` and `Retry-After: 60` headers on the 503 response so intermediaries/service workers do not pin a stale retry page.
3. Assert body contains `http-equiv="refresh"` in the two new tests so a future refactor cannot silently drop the meta tag.
4. Add a wildcard `_ =>` arm in `is_transient` with an explicit comment that `ErrorKind` is `#[non_exhaustive]` and future variants need explicit classification. Currently the hard-coded allow-list is the safe default but the comment makes the contract visible.

#### Consider (Suggestions)
1. Extract `60` as `const RETRY_REFRESH_SECS: u64 = 60;` and format the HTML with it. Aligns with code-style.md and makes the value reviewable in one place.
2. Reconsider 60s vs. node-side 30s wrapper. A refresh interval of, say, 30s would let the browser retry roughly when the node-side wrapper times out again, rather than at an interval decoupled from the op cycle.
3. Log `OperationError { cause }` at `tracing::info!` before dropping it in the 503 path so operators can distinguish a fast op error from a slow-loading contract.

---

### Verdict

**HEAD SHA reviewed:** `6a73cffd4aa69fc54316611f9a70b3d6effd254d`

The mechanism is right but the matcher targets variants that the actual failing code path does not produce. As written, the PR does not close #3472 — the reported "GET operation timed out" still renders as a 500 with no auto-refresh because that string is produced by `NodeError`, not `AxumError { FailedOperation | OperationError }`. Fix the routing (Must-Fix #1) plus add a test that covers the wall-clock timeout path (Must-Fix #2), then a fresh full review pass on the updated HEAD is required because the change will touch a second file (`path_handlers.rs`) outside the currently-reviewed surface.

[AI-assisted - Claude]

iduartgomez (approved):
## Re-review after pushes through 0dd2e28

All three Must-Fix items from the previous review are addressed. Most Should-Fix items are in. Two small gaps remain — calling them out as nits, not blockers.

### Resolved

- **Must-Fix #1 — 30s wall-clock timeout path now routes through the retry page.** `handle_get_response` (`crates/core/src/server/path_handlers.rs:188`) emits `AxumError { ErrorKind::OperationError { cause } }` instead of `NodeError` for both the wrapper timeout and the channel-closed branch, so the matcher fires.
- **Must-Fix #2 — coverage for the actual reported path.** `handle_get_response_maps_timeout_to_operation_error` covers the wrapper-timeout-to-variant mapping; `operation_error_returns_retry_page` covers variant-to-503-plus-body. End to end via two unit tests rather than one integration test, but the path is exercised.
- **Must-Fix #3 — `FailedOperation` documented as defensive.** Comment in `errors.rs` makes the intent explicit: dead in current core, kept so a future stdlib raiser doesn't silently lose retry behaviour.
- **Should-Fix #2 — `Cache-Control: no-store` + `Retry-After: 60` headers on the 503.** Intermediaries / service-workers won't pin a stale retry page.
- **Should-Fix #3 — body assertion in tests.** Both transient tests assert `text.contains(r#"<meta http-equiv="refresh" content="60"#)`, so a future refactor can't silently drop the meta tag.
- **Should-Fix #4 — `#[non_exhaustive]` contract documented above the `matches!`.** Explicit note that new stdlib variants need explicit classification; implicit `matches!` fall-through to `false` satisfies the safety net.
- **Consider #1 — `RETRY_REFRESH_SECS: u64 = 60` extracted** and threaded into both the HTML and the `Retry-After` header.

### Remaining nits (non-blocking)

1. **`ErrorKind::RequestError(RequestError::Timeout)` not in the transient matcher.** The widened arm covers `OperationError`, `FailedOperation`, `ChannelClosed`, `TransportProtocolDisconnect`, and `NodeUnavailable`, but a `RequestError(Timeout)` would still render as 500. Worth adding for symmetry with the `OperationError("…timed out…")` path.
2. **`OperationError { cause }` is dropped silently on the 503 branch.** A `tracing::info!(?cause, "serving retry page")` before the response would let operators distinguish a fast op error from a slow-loading contract without changing the user-visible behaviour.

### Verdict

Approve. The Must-Fix items that gated the previous review are all in, and the mechanism now actually closes #3472 (the dominant `"GET request timed out after 30s"` path renders the auto-refresh page). The two remaining items are small enough to land as follow-ups or in a quick push to this branch — author's call.

[AI-assisted - Claude]

iduartgomez (approved):

sanity:
## Maintainer review + update

Thanks for this — the core idea (transient GET failures → auto-refresh 503 instead of a dead page) is the right fix for #3472, and the `EmptyRing`/`PeerNotJoined` precedent is a good model to follow. I rebased the branch onto current `main` and pushed one commit to address the blocking rule-review finding.

### ✅ Fixed: blocking rule-review warning
`RequestError(Timeout)` was in the `is_transient` matcher with no test (the one Warning blocking merge). I added `request_error_timeout_returns_retry_page` (mirrors the sibling transient-variant tests) and merged current `main` in. `cargo test -p freenet --lib errors::tests` → 10 passed.

### ⚠️ Blocker found in review: `OperationError` is over-classified as transient (Codex P2 — confirmed)

The `is_transient` matcher treats **all** `ErrorKind::OperationError { .. }` as transient, but that variant is **dual-use**:

- **Transient** (what we want to retry): the synthesized 30s fetch-wrapper timeout and the channel-closed case in `handle_get_response` (`path_handlers.rs:500-509`).
- **Terminal** (must NOT infinite-retry): node-returned GET failures flow through verbatim via `path_handlers.rs:528-534` (`error: err.kind().clone()`), and several terminal causes produce `OperationError` — e.g. a **locally banned contract** (`client_events.rs:347`, from `OpError::ContractBanned`) and generic `"GET failed: …"` (`operations/get/op_ctx_task.rs:1661`).

Net effect: requesting a banned/terminally-failed contract's web page would serve the **infinite 60s auto-reload page** instead of an actionable error. (The common terminal case — not-found-everywhere — is safe: it returns a `NotFound` *success* → 404 via `MissingContract`, so only the `OperationError`-producing terminal paths are affected.)

### Suggested fix (small, conservative)
The two synthesized transient cases can use their **semantically exact** variants, which are already in the transient set, so `OperationError` can be dropped from it entirely:

- `Err(_)` (30s elapsed) → `ErrorKind::RequestError(RequestError::Timeout)` (instead of `OperationError`)
- `Ok(None)` (channel closed) → `ErrorKind::ChannelClosed` (instead of `OperationError`)
- Remove `ErrorKind::OperationError { .. }` from `is_transient` in `errors.rs`

Then node-returned terminal `OperationError` → 500 (correct), while genuine transient timeouts/channel-closes still get the retry page via their proper variants. The node-recovery variants you added (`ChannelClosed`, `TransportProtocolDisconnect`, `NodeUnavailable`, `RequestError(Timeout)`) stay as-is. Tests would need a small update (the `operation_error_*`/`handle_get_response_maps_timeout_*` cases flip to the new variants).

I've not made this change yet — flagging it for a maintainer decision on whether to take it on here or have you address it. Holding the merge until it's resolved.

[AI-assisted - Claude]

sanity:
## Update — P2 + P3 addressed, re-reviewed

Pushed the narrowing fix and re-ran Codex.

**P2 (over-broad transient classification) — fixed.** `OperationError` is no longer treated as transient. The two genuinely-transient synthesized GET cases now use unambiguous variants:
- 30s fetch-wrapper timeout (`Err(Elapsed)`) → `RequestError(Timeout)`
- closed response channel (`Ok(None)`) → `ChannelClosed`

Terminal node-returned `OperationError` (banned contract, exhausted GET) now correctly maps to 500 instead of an infinite retry page. Tests updated: `operation_error_returns_internal_server_error` (asserts 500), plus `handle_get_response_maps_timeout_to_request_timeout` and `handle_get_response_maps_channel_closed_to_channel_closed`.

**P3 (retry-page CSS) — fixed.** The `retry_loading_page` `format!` string used `{{{{`/`}}}}`, which emit *double* braces (invalid CSS, page rendered unstyled). Switched to `{{`/`}}` so it emits single braces, with a brace assertion in `failed_operation_returns_retry_page` to lock it in. Good catch by Codex.

Local: `cargo fmt` clean, `cargo clippy -p freenet --lib` clean, `errors::tests` + `handle_get_response` tests green (12/12). Codex re-review on the final diff: P2 gone, only the P3 (now fixed) remained.

Merging once CI is green on the updated HEAD.

[AI-assisted - Claude]

PR #4305, opened 2026-05-28
fix(service): preserve existing ExecStart CLI flags on service install

### LINKED ISSUE #4275: fix: service start silently drops ExecStart CLI flags
## Problem

Running `freenet service install` overwrites the user's systemd unit file (`~/.config/systemd/user/freenet.service`), discarding all CLI arguments that were previously set in `ExecStart`. (`freenet service start` and `restart` delegate to `systemctl` and do not overwrite the unit — but `install` is called during initial setup and can be re-triggered during updates.)

**Example:** After an operator ran `freenet service stop` to perform database maintenance and then `freenet service start` to resume, the following configuration was silently lost:

| Parameter | Before restart | After restart |
|---|---|---|
| `--network-port` | custom (non-default) | default |
| `--is-gateway` | set | absent |
| `--gateway` entries | 2 peers configured | none |
| `--allowed-host` | set | absent |
| `--public-network-address` | set | absent |
| `--max-number-of-connections` | custom (2000) | default (20) |
| `--total-bandwidth-limit` | set | unset |

The operator only noticed the regression when the node's network port changed and gateway connections dropped.

## Root cause

`freenet service install` generates the systemd unit file from scratch using a built-in template and writes it unconditionally. It does **not** preserve any pre-existing `ExecStart` flags. The unit file is treated as entirely owned by the binary — any CLI flags the operator added manually or via a previous `freenet network ...` invocation are erased on the next `service install` (which may be triggered during setup, updates, or if the operator reruns it manually).

Furthermore, `config.toml` does NOT store these network-level flags:
- `--is-gateway`
- `--gateway` (peer list)
- `--allowed-host`
- `--public-network-address`
- `--max-number-of-connections`
- `--total-bandwidth-limit`
- `--event-loop-channel-capacity`

so there is no way to recover them from config after the unit is overwritten.

## Expected behaviour

1. `freenet service start` should **preserve existing ExecStart flags** and only touch the unit template for fields that have no CLI equivalent.
2. Better: `freenet service install` should write ALL effective settings (including `--is-gateway`, `--gateway`, `--public-network-address`, etc.) into `config.toml` so that a bare `freenet network` (without any flags) produces the same behaviour as the fully-flagged version.
3. At minimum: print a warning when the unit file is about to be overwritten with a different `ExecStart`, and/or back up the previous unit as `freenet.service.bak` (as `service install` already does).

### 1. The systemd unit template — hardcodes a bare `ExecStart`

**File:** `crates/core/src/bin/commands/service.rs`, lines 2543–2607

The function `generate_user_service_file()` uses `format!()` to produce the unit. The `ExecStart` on **line 2553** is hardcoded to `{binary} network` — no CLI flags are injected, and there is no mechanism to preserve flags from an existing unit file:

```rust
pub fn generate_user_service_file(binary_path: &Path, log_dir: &Path) -> String {
    format!(
        r#"[Unit]
Description=Freenet Node
...
[Service]
Type=simple
ExecStart={binary} network          // ← line 2553: hardcoded, no CLI args
Restart=always
...
"#,
        binary = binary_path.display(),
        log_dir = log_dir.display()
    )
}
```

The same happens for the system-level template (`generate_system_service_file`, line 2610) at **line 2627**.

### 2. `install_user_service()` always overwrites — no read-before-write

**File:** `crates/core/src/bin/commands/service.rs`, lines 2440–2473

```rust
fn install_user_service() -> Result<()> {
    ...
    let service_content = generate_user_service_file(&exe_path, &log_dir);  // ← line 2453
    let service_path = service_dir.join("freenet.service");

    fs::write(&service_path, service_content)                               // ← line 2456: unconditional overwrite
        .context("Failed to write service file")?;

    systemctl_with_hint(false, &["daemon-reload"], "reload systemd daemon")?;
    systemctl_with_hint(false, &["enable", "freenet"], "enable service")?;
    ...
}
```

There is **no** existing-unit parsing or flag-preservation logic. The old file is simply clobbered. Compare with the installer's own backup hook (`BackupOldService` in the same file) — that mechanism only fires for the *system* service path, not for user services, and only to save the old file, not to merge flags.

### 3. `start_service()` delegates to systemctl — no reinstall

**File:** `crates/core/src/bin/commands/service.rs`, lines 2742–2748

```rust
fn start_service(system: bool) -> Result<()> {
    let system_mode = use_system_mode(system);
    systemctl_with_hint(system_mode, &["start", "freenet"], "start service")?;
    println!("Freenet service started.");
    ...
}
```

`start_service` does **not** call `install_user_service` — it only runs `systemctl start`. This means the unit is only overwritten when the user (or a workflow step) explicitly calls `freenet service install`.

### 4. `run_wrapper_loop()` spawns the child with no flags

**File:** `crates/core/src/bin/commands/service.rs`, lines 1460–1471

```rust
loop {
    ...
    let mut cmd = std::process::Command::new(&exe_path);
    cmd.arg("network");                // ← line 1471: no flags forwarded
    ...
    let mut child = match cmd.spawn() {
        ...
    };
}
```

The tray wrapper that launches the node in the background also passes zero CLI flags — it relies entirely on `config.toml` and the systemd unit.

### 5. Dispatch table

**File:** `crates/core/src/bin/commands/service.rs`, lines 193–209

```rust
match self {
    ServiceCommand::Install { system } => install_service(*system),  // ← overwrites unit
    ServiceCommand::Start  { system } => start_service(*system),     // ← just systemctl start
    ServiceCommand::Restart { system } => restart_service(*system),  // ← just systemctl restart
    ...
}
```

## Root cause summary

The data flow is:

```
generate_user_service_file()     ← template has "ExecStart={binary} network" (no args)
        ↓
install_user_service()           ← fs::write() always overwrites (no read → merge → write)
        ↓
systemctl daemon-reload + start
```

Meanwhile, **`config.toml` does not store** the following network-level flags (confirmed by examining the node's config file on a production gateway):

- `--is-gateway`
- `--gateway` (peer list)
- `--allowed-host`
- `--public-network-address`
- `--max-number-of-connections`
- `--total-bandwidth-limit`
- `--event-loop-channel-capacity`

So even if the operator rebuilds the unit manually, there is no single source of truth for these values.

## Suggested fix locations

- `crates/core/src/bin/commands/service.rs` — `generate_user_service_file()` (line 2543) should accept optional CLI args, and `install_user_service()` (line 2440) should parse any existing unit's `ExecStart` and merge flags before writing.
- `crates/core/src/config.rs` — ensure all network flags are representable in `config.toml` so that a bare `freenet network` (zero flags) produces the same behaviour as the fully-configured one.

## Severity

Medium. The node continues to run, but silently changes its network behaviour (port, peer set, bandwidth limits, gateway status). In a production gateway this can cause:
- Loss of incoming connections (port change)
- Disconnection from upstream gateways
- Bandwidth throttling (limits reset to defaults)

Workaround: always back up `~/.config/systemd/user/freenet.service` before running `freenet service stop`.

Basedfloppa:
## Problem

`freenet service install` unconditionally overwrites the systemd unit file
(`~/.config/systemd/user/freenet.service`) with a template containing bare
`ExecStart={binary} network` — silently dropping any CLI flags the operator
added manually.

This caused production gateways to lose critical configuration on restart:

| Parameter | Before install | After install |
|---|---|---|
| `--is-gateway` | set | absent |
| `--gateway` entries | 2 peers | none |
| `--network-port` | custom | default |
| `--max-connections` | 2000 | 20 |
| `--bandwidth-limit` | 100 Mbps | unset |
| `--allowed-host` | set | absent |

These flags are NOT stored in `config.toml`, so there is no way to recover
them after the unit is overwritten. The operator only notices when the
node's port changes and gateway connections drop.

## Solution

Three changes:

1. **`extract_execstart_args()`** — reads an existing unit file and returns
   everything after `network` on the `ExecStart` line (the operator's custom
   CLI flags). Returns `None` if the file doesn't exist or has no extra args.

2. **`install_user_service()` / `install_system_service()`** — before writing:
   - Parse existing unit → extract CLI flags
   - Pass flags into the template generator
   - Back up old unit as `freenet.service.bak`
   - Print a warning to stderr
   - Write new unit with preserved flags

3. **`generate_user_service_file()` / `generate_system_service_file()`** —
   new `extra_args: Option<&str>` parameter. When present, the flags are
   appended to `ExecStart={binary} network {extra_args}`.

## Testing

- `cargo fmt` — clean
- `cargo clippy -- -D warnings` — clean (fixed `trim_split_whitespace`,
  `let_underscore_must_use`)

Manual verification scenario:
```
# Before fix:
$ freenet service install   # overwrites unit, drops --is-gateway etc.

# After fix:
$ freenet service install   # preserves --is-gateway, backs up old unit
  freenet: preserved existing ExecStart flags; old unit backed up to
  /root/.config/systemd/user/freenet.service.bak
```

## Fixes

Closes #4275

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2450,8 +2450,32 @@ fn install_user_service() -> Result<()> {
     let log_dir = home_dir.join(".local/state/freenet");
     fs::create_dir_all(&log_dir).context("Failed to create log directory")?;
 
-    let service_content = generate_user_service_file(&exe_path, &log_dir);
     let service_path = service_dir.join("freenet.service");
+
+    // Preserve CLI flags from any existing unit file so operators who
+    // hand-tuned ExecStart (e.g. --is-gateway, --network-port, --gateway)
+    // don't lose them on `freenet service install` (#4275).
+    let preserved_args = extract_execstart_args(&service_path);
+    let extra_args = preserved_args.as_deref();
+
+    let service_content = generate_user_service_file(&exe_path, &log_dir, extra_args);
+
+    // If we changed ExecStart, back up the old unit before overwriting.
+    if preserved_args.is_some() && service_path.exists() {
+        let bak_path = service_path.with_extension("service.bak");
+        #[allow(clippy::let_underscore_must_use)]
+        if fs::copy(&service_path, &bak_path).is_ok() {
+            eprintln!(
+                "freenet: preserved existing ExecStart flags; old unit backed up to {}",
+                bak_path.display()
+            );
+        } else {
+            eprintln!(
+                "freenet: warning: failed to back up old unit to {}",
+                bak_path.display()
+            );
+        }
+    }
 
     fs::write(&service_path, service_content).context("Failed to write service file")?;
 
@@ -2506,9 +2530,31 @@ fn install_system_service() -> Result<()> {
     fs::create_dir_all(&log_dir).context("Failed to create log directory")?;
     chown_to_user(&log_dir, &username);
 
-    let service_content = generate_system_service_file(&exe_path, &log_dir, &username, &home_dir);
+    let system_service_path = std::path::Path::new(SYSTEM_SERVICE_PATH);
+    let preserved_args = extract_execstart_args(system_service_path);
+    let extra_args = preserved_args.as_deref();
 
-    fs::write(SYSTEM_SERVICE_PATH, &service_content).with_context(|| {
+    let service_content =
+        generate_system_service_file(&exe_path, &log_dir, &username, &home_dir, extra_args);
+
+    // If we changed ExecStart, back up the old unit before overwriting.
+    if preserved_args.is_some() && system_service_path.exists() {
+        let bak_path = system_service_path.with_extension("service.bak");
+        #[allow(clippy::let_underscore_must_use)]
+        if fs::copy(system_service_path, &bak_path).is_ok() {
+            eprintln!(
+                "freenet: preserved existing ExecStart flags; old unit backed up to {}",
+                bak_path.display()
+            );
+        } else {
+            eprintln!(
+                "freenet: warning: failed to back up old unit to {}",
+                bak_path.display()
+            );
+        }
+    }
+
+    fs::write(system_service_path, &service_content).with_context(|| {
         format!(
             "Failed to write service file to {SYSTEM_SERVICE_PATH}. \
              Are you running as root? Try: sudo freenet service install --system"
@@ -2540,7 +2586,16 @@ fn install_system_service() -> Result<()> {
 }
 
 #[cfg(target_os = "linux")]
-pub fn generate_user_service_file(binary_path: &Path, log_dir: &Path) -> String {
+pub fn generate_user_service_file(
+    binary_path: &Path,
+    log_dir: &Path,
+    extra_args: Option<&str>,
+) -> String {
+    let exec_start = if let Some(args) = extra_args {
+        format!("{binary} network {args}", binary = binary_path.display())
+    } else {
+        format!("{binary} network", binary = binary_path.display())
+    };
     format!(
         r#"[Unit]
 Description=Freenet Node
@@ -2550,7 +2605,7 @@ Wants=network-online.target
 
 [Service]
 Type=simple
-ExecStart={binary} network
+ExecStart={exec_start}
 Restart=always
 # Wait 10 seconds before restart to avoid rapid restart loops
 RestartSec=10
@@ -2601,6 +2656,7 @@ CPUQuota=200%
 [Install]
 WantedBy=default.target
 "#,
+        exec_start = exec_start,
         binary = binary_path.display(),
         log_dir = log_dir.display()
     )
@@ -2612,7 +2668,13 @@ pub fn generate_system_service_file(
     log_dir: &Path,
     username: &str,
     home_dir: &Path,
+    extra_args: Option<&str>,
 ) -> String {
+    let exec_start = if let Some(args) = extra_args {
+        format!("{binary} network {args}", binary = binary_path.display())
+    } else {
+        format!("{binary} network", binary = binary_path.display())
+    };
     format!(
         r#"[Unit]
 Description=Freenet Node
@@ -2624,7 +2686,7 @@ Wants=network-online.target
 Type=simple
 User={username}
 Environment=HOME={home}
-ExecStart={binary} network
+ExecStart={exec_start}
 Restart=always
 # Wait 10 seconds before restart to avoid rapid restart loops
 RestartSec=10
@@ -2675,6 +2737,7 @@ CPUQuota=200%
 [Install]
 WantedBy=multi-user.target
 "#,
+        exec_start = exec_start,
         binary = binary_path.display(),
         log_dir = log_dir.display(),
         username = username,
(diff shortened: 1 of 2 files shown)


iduartgomez (changes requested):
Preservation logic correct for the "operator never touches install again" path, but breaks down the moment configuration genuinely needs to change. Several gaps to address before merge.

## Blocking

### 1. No way to opt out of preservation

`extract_execstart_args` runs unconditionally. There is no `--force` / `--reset` / `--no-preserve` flag. An operator who actually wants the fresh template (e.g. to drop a deprecated flag, recover from a typo'd hand-edit, or reset after a binary rename) must manually `rm` the unit file. The install command's whole point is "set me up correctly" — silently inheriting old state defeats that.

Please add a `--reset` (or `--no-preserve-args`) flag on `freenet service install` that bypasses `extract_execstart_args` and writes the bare template.

### 2. New-flag-after-upgrade scenario silently regresses

If a future release requires a new CLI flag (e.g. a mandatory `--telemetry-endpoint`, or a renamed flag), `freenet service install` post-upgrade will keep the *old* arg list verbatim and the new flag will never appear. The operator has no signal that their unit is now stale — the install command claims success. This is the same class of bug #4275 reports, just inverted.

At minimum: when preserving, print the preserved arg string to stderr so the operator can eyeball it against current `--help`. Better: diff against a known-good baseline and warn if any documented flag is missing.

### 3. Removed/renamed flag → crash loop with no early warning

If a flag is removed in the new binary (e.g. `--old-flag` → `--new-flag`), preservation re-emits the dead flag, systemd starts the binary, binary rejects unknown arg, service fails. The `.bak` exists but the operator hits a crash loop first and has to dig through `journalctl` to find out why. Pair with #2: dry-run the preserved ExecStart against `freenet network --help` parsing, or at least `freenet network <args> --check` (if such a mode exists / can be added), before writing.

### 4. `install` and `update` have opposite semantics

`crates/core/src/bin/commands/update.rs:1184-1188` calls both generators with `extra_args: None`. So `freenet update` clobbers operator flags, but `freenet service install` preserves them. Two commands an operator might reasonably reach for, opposite behavior, no documentation of the split. Either:

- `update` should also preserve (call `extract_execstart_args` against the existing unit), OR
- both commands document the split explicitly and emit a stderr warning when their behavior would surprise.

This is the most important blocker — current state means `freenet update` silently undoes the protection this PR adds.

### 5. `.bak` is overwritten on every install

`service_path.with_extension("service.bak")` collapses to a single filename. Run `install` twice with intervening hand-edits and the original-original is gone. Use a timestamped suffix (`freenet.service.bak.<unix-ts>`) or refuse to overwrite an existing `.bak`.

## Non-blocking

- `extract_execstart_args` skips lines that don't begin with `ExecStart=` but doesn't handle systemd line continuations (`\` at EOL). Unlikely in practice for this template but worth a comment.
- `#[allow(clippy::let_underscore_must_use)]` on `fs::copy(...).is_ok()` is misapplied — `is_ok()` already consumes the `Result`, the allow is dead. Remove it.
- Real fix for #4275 is persisting these flags in `config.toml` so they survive any unit rewrite. This PR is a useful bandage but please file a follow-up issue tracking the config-persistence work; otherwise the bandage becomes load-bearing.

## Summary

Approve the preservation mechanism in principle. Cannot approve while `freenet update` still clobbers (gap #4), no escape hatch exists (gap #1), and silent staleness on upgrade (gaps #2/#3) is unaddressed. Fix #4 at minimum.

iduartgomez:
we will change approach for this

Issue #4064, opened 2026-05-09
Webapp updates: published contract updates don't reliably propagate to subscribed peers


iduartgomez:
## Context
Multi-peer real-network test. Peer1 (port 7509) and peer2 (port 7510) both subscribed to a published webapp contract.

After running an \`fdev network publish\` against peer1 with a new build:
- peer1's local store updates (it serves new HTML/wasm)
- \`fdev publish\` returns \"PUT operation timed out\" at the network layer — but peer1's local state IS updated
- peer2 keeps serving the OLD build for several minutes (verified by HTML md5 mismatch)
- peer2 logs show \`BroadcastToStreaming\` arriving for the contract but \`merge_rejected_valid_local\` because \`New state version X must be higher than current version X\` — same version
- Only an explicit \`fdev network publish\` directly against peer2 updates peer2

## Repro
Two real \`freenet network\` peers, both subscribed to a webapp contract. Republish the webapp from peer1. Wait. Verify peer2 still serves the old version (compare HTML md5).

## Hypotheses
1. **Version stale at peer1's PUT**: peer1's PUT doesn't carry a strictly-higher version than peer2 already has, so peer2 rejects the broadcast. River and freenet-email use \`unix_seconds / 60\` for version. If the same publisher republishes within a minute, versions tie. We weren't within-the-same-minute here, but worth confirming the rule.
2. **PUT didn't fan out to subscribers**: the timed-out PUT means peer1 never broadcast UPDATE to its subscribers list. Subscribers don't pull on demand — they wait for explicit UPDATE.
3. **Network congestion**: PUT broadcast fanout couldn't reach peer2 within the 300s window of the publisher.

## Why this matters
For dApps with 1 publisher + N subscribers (the typical webapp shape — River, freenet-email, anything served via the web container), every subscriber must converge to the new build. The current behavior (2nd peer pinned to old build until reboot or explicit republish) breaks the subscriber model.

QA-wise it's also a footgun: ship a fix, republish, look at peer2 to verify, see the OLD bug, conclude the fix didn't work, when actually the bug WAS fixed but the new code isn't running.

## Logs (peer2)

\`\`\`
INFO contract_handling:upsert_contract_state{key=DtJB...}: freenet::contract::executor::runtime: Merge rejected incoming state but local state is valid - not replacing (incoming state may be stale) ...
  reason: New state version 1778323153 must be higher than current version 1778323153
  local_state_size=8940494 incoming_state_size=8940515
[...]

PR #4306, opened 2026-05-28
fix(contracts): make idempotency probe sound + TTL the broken-invariant flag

### LINKED ISSUE #4295: test: test_ping_multi_node flaky again — fails on main HEAD + multiple PRs (~50%+ rate)
## Symptom

\`apps/freenet-ping/app/tests/run_app.rs::test_ping_multi_node\` is failing across multiple recent PRs:

- #4292 (Phase 1.5 telemetry): TRY 3 FAIL — propagation 83.3%, Node1 missing ping-from-node2
- worktree-gateway-graceful-drain: TRY 1 FAIL on same test
- perf/4242-delegate-context-alloc: failed on Unit & Integration

## Local reproduction (2026-05-27)

Both \`main\` (\`c0d7fd04\`) and the #4292 branch (\`13323189\`) fail this test locally on nova when run in isolation. Different failure modes per run — different nodes are missing different ping tags — consistent with a propagation race rather than a deterministic logic bug.

\`main\` (c0d7fd04) failure: \`GW[3,3,1], N1[3,3,1], N2[1,0,1]\` — Node2 short on ping-from-gw and ping-from-node1.

\`#4292\` (13323189) failure: \`GW[2,0,2], N1[1,0,0], N2[2,0,3]\` — Node1 missing ping-from-node1 and ping-from-node2; gateway also missing ping-from-node1.

Reproduction:
\`\`\`bash
cd ~/code/freenet/freenet-core/main
cargo nextest run --release -p freenet-ping-app --test run_app test_ping_multi_node \\
  --success-output=final --failure-output=final
\`\`\`
~110 s per run. Failed on first try in both directions.

## Why this matters

This test gates merges. Multiple PRs are currently blocked or flaky-pass-then-fail on it. The CI failure on #4292 was initially misattributed to changes in that PR until local reproduction on main showed the same failure mode, costing investigation time.

## Historical context

Prior tickets for the same test family: #3036, #2024, #3456, #3490, #3078, #3075. Each was closed but the flake re-emerges. This pattern strongly suggests the test relies on overlay propagation timing that is sensitive to whatever else is in the test runtime, and "fix" PRs have addressed individual symptoms without removing the underlying race surface.

## Proposed action

1. Mark \`test_ping_multi_node\` \`#[ignore]\` with a comment referencing this issue, to unblock unrelated PRs while a real fix is investigated.
2. Investigate root cause separately. The user reported the prior congestion-control work failed because there was no robust local reproduction — same pattern here: the test relies on a propagation race that CI exercises ~50% of the time. A deterministic simulation test that exercises the same propagation guarantees (without real-time / real-network dependencies) would be a more reliable gate.

[AI-assisted - Claude]

sanity:
## Problem

`test_ping_multi_node` flaked ~50% (#4295). Root cause (verified via logs):
the idempotency probe added in #4279 flags a contract `NonIdempotent` by
comparing the **raw serialized bytes** of `update_state(update_state(S,U),U)`
against `update_state(S,U)`. A correct, logically-idempotent contract whose
serialization is non-canonical — the ping contract's `Ping.from` is a
`HashMap` — re-serializes the SAME logical state in a different key ORDER on
re-merge. Observed in a failing run: `post_merge_size=374 probe_size=374`
(identical size, flagged anyway). The false positive calls
`record_broken_invariant(NonIdempotent)`, which had **no TTL** and permanently
suppressed both `commit_state_update` and `BroadcastStateChange` — so cross-node
propagation halted and interest-sync anti-entropy could not heal it (the
receiver's commit was suppressed too). Probabilistic firing (p=1/32 per State
merge) produced the ~50% rate.

## Approach

Byte inequality does not prove a CRDT violation — logical idempotency does not
imply byte-determinism. Two changes, both in core (general fix, no contract-key
change):

1. **Probe soundness** (`Executor::maybe_probe_idempotency`): re-apply the
   update once and compare the post-merge state to the re-applied state by
   **byte MULTISET** (`byte_multiset_eq`), not raw bytes and not size. A benign
   serialization reordering (HashMap/HashSet key order) permutes the bytes but
   preserves their multiset → not flagged (fixes #4295). Any genuine content
   change — a counter, timestamp, signature, added/removed entry — alters the
   multiset → flagged, **including the fixed-size byte-churn shape of the #4251
   production violator** (a ~464-byte state whose counter prefix churns in
   place) that a size-only check would miss. Residual false-negative: a content
   change that coincidentally preserves the exact byte multiset evades detection
   (far narrower than a size-only blind spot, far safer than byte-equality's
   false positives). Discriminator extracted to the pure `byte_multiset_eq`
   helper and unit-tested.

2. **TTL on the flag** (`BrokenInvariantsTracker`): the `NonIdempotent` flag now
   expires after `BROKEN_INVARIANT_TTL` (5 min), swept by the Ring reaper
   (mirroring the UPDATE rate limiter's `cleanup`, with an atomic `remove_if`
   re-check so a concurrently-refreshed flag's persisted row is not purged), so
   any residual false positive self-heals instead of bricking the contract
   forever. Satisfies the "cleanup exemptions must be time-bounded" rule.
   Re-recording refreshes the window so a genuine violation stays suppressed;
   after expiry a genuine violation is re-detected by the next sampled probe (a
   deliberately accepted, bounded periodic leak documented in the module — we
   bias toward false-positive recovery, the worse failure being the #4295
   permanent brick).

## Testing

- New unit tests: `byte_multiset_eq` (reordering benign; fixed-size 464-byte
  content churn flagged — the #4251 shape; growing state flagged); flag TTL
  expiry, cleanup reclamation, re-record refresh.
- End-to-end: `test_ping_multi_node` run repeatedly — **0 false flags**; the
  idempotency-probe convergence-freeze failure mode (the #4279-introduced
  regression) is eliminated.
- `cargo fmt` + `cargo clippy -p freenet --lib --tests --locked -- -D warnings` clean.

Updated `.claude/rules/contracts.md` (flag no longer "permanent"; describes the
multiset semantics).

## Scope / residual

Fixes the dominant, recently-introduced (#4279) cause of #4295 — the
idempotency-probe convergence freeze. Separate, **pre-existing** propagation
flakes remain in the same test (a >64KB contract GET landing state without
code/params; and lost-update non-convergence not healed within the window) —
the #3279 / #4064 / #4071 family, likely including the unimplemented
streaming-Found GET relay #3883. Those are deeper networking issues with their
own tracking and are intentionally **not** bundled here (per a deliberate
scope decision: bundling a high-risk, hard-to-verify networking change onto
this verified one was judged worse for regression risk). So this PR does not
fully close #4295; it removes the regression that took the test to ~50%, and
`test_ping_multi_node` may still flake (~10–20%) on the residuals until they
land.

Refs #4295. Residual tracked by #3279 / #4064 / #3883.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/contracts.md
+++ b/.claude/rules/contracts.md
@@ -148,11 +148,24 @@ State merging rules:
   - For `UpdateData::State` inputs, merge MUST be idempotent:
       update_state(update_state(S, State(X)), State(X)) == update_state(S, State(X))
     Enforced at runtime via the in-peer probe in
-    `Executor::maybe_probe_idempotency`. Violators get flagged in
-    `Ring::broken_invariants` and their outbound `BroadcastStateChange`
-    is suppressed locally; the flag is permanent for that contract
-    instance id and survives restart. See `ring::broken_invariants` and
-    issue #4251 / PR #4279.
+    `Executor::maybe_probe_idempotency`. The probe re-applies the update and
+    flags a violation when the re-applied state's byte MULTISET differs from
+    the original — i.e. the CONTENT changed (a counter/timestamp/signature
+    churned, an entry was added/removed), including the fixed-size byte-churn
+    shape of the #4251 incident. It does NOT flag a mere REORDERING of the
+    same bytes, because a correct contract with non-canonical serialization
+    (HashMap/HashSet iteration order) re-serializes the same logical state in
+    a different byte order (the #4295 false-positive case). Violators get
+    flagged in `Ring::broken_invariants` and their outbound
+    `BroadcastStateChange` is suppressed locally. The flag is TTL-bounded
+    (`BROKEN_INVARIANT_TTL`, swept by the Ring reaper) so a false positive
+    self-heals instead of bricking the contract forever; a genuine violation
+    is re-detected by the next sampled probe after expiry (probabilistic, so
+    a bounded periodic leak — see `ring::broken_invariants`). Residual
+    false-negative: a content change that coincidentally preserves the exact
+    byte multiset evades detection (far narrower than a size-only check). See
+    `ring::broken_invariants`, issue #4251 / PR #4279, and the #4295
+    soundness/TTL fix.
     Note: this invariant is NOT enforced for `UpdateData::Delta` inputs
     (CmRDT-style "increment by X" deltas legitimately violate it) or
     `UpdateData::RelatedState` (a cross-contract hint, not a CRDT op).
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -21,6 +21,47 @@ const RELATED_FETCH_TIMEOUT: Duration = Duration::from_secs(10);
 /// Sample selection uses `GlobalRng` (deterministic under a fixed seed
 /// for simulation tests). See `Executor::maybe_probe_idempotency`.
 const IDEMPOTENCY_PROBE_PROBABILITY: f64 = 1.0 / 32.0;
+
+/// Returns true if `a` and `b` contain the same multiset of bytes — i.e. one
+/// is a reordering of the other — and false if their byte content differs.
+///
+/// This is the discriminator the idempotency probe uses to tell benign
+/// serialization nondeterminism apart from a genuine non-idempotent merge:
+///
+/// - **Benign flutter** (the #4295 false-positive case): a correct contract
+///   with non-canonical serialization (`HashMap`/`HashSet` iteration order)
+///   re-serializes the SAME logical state in a different byte ORDER. Reordering
+///   permutes the serialized bytes but preserves their multiset, so this
+///   returns `true` → not flagged.
+/// - **Genuine non-idempotency** (the #4251/#4279 case the gate must catch):
+///   re-applying the update changes the state's CONTENT — a counter that churns
+///   in place, an embedded timestamp/signature regenerated each merge, an
+///   added/removed entry. Any content change alters the byte multiset (e.g. a
+///   464→465 counter flips digit bytes), so this returns `false` → flagged.
+///   Crucially this catches the *fixed-size* byte-different violator (the real
+///   #4251 incident was a constant-size ~464-byte state) that a size-only
+///   check would miss.
+///
+/// Residual false-negative: a content change that coincidentally preserves the
+/// exact byte multiset (e.g. swapping two equal bytes) evades detection. This
+/// is far narrower than the size-only heuristic's blind spot and far safer than
+/// byte-equality's false positives (which permanently suppressed propagation in
+/// #4295). O(n) in the state size; only runs on the sampled, byte-different
+/// probe path.
+fn byte_multiset_eq(a: &[u8], b: &[u8]) -> bool {
+    if a.len() != b.len() {
+        return false;
+    }
+    let mut hist = [0i64; 256];
+    for &x in a {
+        hist[x as usize] += 1;
+    }
+    for &x in b {
+        hist[x as usize] -= 1;
+    }
+    hist.iter().all(|&c| c == 0)
+}
+
 use crate::node::OpManager;
 use crate::wasm_runtime::{
     BackendEngine, DEFAULT_MODULE_CACHE_CAPACITY, MAX_STATE_SIZE, RuntimeConfig, SharedModuleCache,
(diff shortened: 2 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #4306

### Summary
- **PR Title:** fix(contracts): make idempotency probe sound + TTL the broken-invariant flag
- **Type:** fix
- **Review tier:** Full (touches the executor merge path + ring state, and modifies the #4279 safety feature — a high-risk surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture (freenet:* agents) + external Codex, across two rounds (size-heuristic → byte-multiset rework). HEAD reviewed: `b8ecd855`.

This review covers two rounds: a first pass found a blocking correctness flaw; the fix was reworked and a second pass (code-first + skeptical + Codex) plus a Codex-found cleanup race were addressed.

---

### What the change does
1. **Probe soundness** (`runtime.rs::maybe_probe_idempotency` + `byte_multiset_eq`): re-applies the update once and compares the post-merge state to the re-applied state by **byte multiset** (not raw bytes, not size). Serialization reordering (HashMap/HashSet key order) preserves the multiset → not flagged (fixes the #4295 false positive); any content change — incl. fixed-size counter/timestamp churn — changes the multiset → flagged.
2. **TTL on the flag** (`broken_invariants.rs`): `NonIdempotent` now expires after `BROKEN_INVARIANT_TTL` (300s), swept by the Ring reaper, so a residual false positive self-heals instead of permanently bricking the contract.

---

### Findings and resolutions

#### Round 1 — Blocking (RESOLVED)
- **[Must-Fix] Detection regression.** The initial revision used a serialized-**size**-convergence heuristic. All four reviewers + Codex independently found it would NOT flag the exact #4251 production shape — a fixed-size, byte-different non-idempotent merge (~464-byte counter churning in place; the repo's own `mock_wasm_runtime.rs` fixture models precisely this). That neutralized #4279's storm defense for the case it exists to catch.
  - **Resolved:** reworked to byte-**multiset** comparison (`byte_multiset_eq`), which flags any content change regardless of size while still ignoring benign reordering. New test `fixed_size_content_change_is_flagged` pins the 464-byte case. Skeptical + code-first re-review confirm the finding is resolved.

#### Round 2 — Medium (RESOLVED)
- **[P2, Codex] TOCTOU in `cleanup()`.** `retain()` then a separate unconditional `remove_from_storage()` loop could delete the persisted row of a flag a concurrent `record()` had just refreshed, breaking restart re-hydration.
  - **Resolved:** `cleanup()` now uses `DashMap::remove_if` to atomically re-check expiry at removal and only purges storage when it actually removed an expired entry.
- **[Doc] PR description was stale** (described the abandoned size approach). **Resolved:** PR body + in-code docs + `.claude/rules/contracts.md` updated to the multiset semantics; the "≈one merge per TTL" re-engage claim corrected to the honest probabilistic (~1/32-sampled) bound.

#### Acknowledged residual (not a bug)
- **Multiset-preserving false-negative:** a content change that coincidentally preserves the exact byte multiset evades detection. Far narrower than the size-only blind spot and far safer than byte-equality's false positives. Documented.
- **`byte_multiset_eq` correctness verified:** equal-length guard + `[i64;256]` histogram is overflow-safe up to the 50 MiB state cap (signed `i64` is the deliberate, correct choice).

---

### Testing
| Unit | ✅ | `byte_multiset_eq` (reorder benign / fixed-size churn flagged / growth flagged); flag TTL expiry, cleanup reclamation, re-record refresh. |
| Integration (e2e) | ✅ | `test_ping_multi_node` run 10× on the fix: **0 probe false-flags**; the idempotency-probe convergence-freeze is eliminated. |
| Simulation | N/A | Executor-level end-to-end probe test (constructing OpManager) remains a coverage gap, matching the pre-existing #4279 fixture level; the pure discriminator is unit-pinned. |

---

### Scope
Fixes the dominant, recently-introduced (#4279) cause of #4295. Pre-existing propagation flakes (GET landing state-without-code; lost-update non-convergence) — the #3279/#4064/#3883 family — remain and are intentionally out of scope (deliberate decision: bundling a high-risk, hard-to-verify networking change onto this verified one was judged worse for regression risk). **This PR does not fully close #4295**; `test_ping_multi_node` may still flake (~10–20%) on those residuals until they land.

---

### Verdict
**State:** Ready to Merge (for the idempotency-probe + TTL fix). The blocking Round-1 finding and the Round-2 cleanup race are resolved and re-reviewed on the current HEAD.
**HEAD SHA reviewed:** `b8ecd855`

Caveat for the merge decision (not a code finding): because the pre-existing #3279/#4064 residuals remain, `test_ping_multi_node` in this PR's own CI may flaky-fail — that is a known, separately-tracked issue, not a defect in this change.

[AI-assisted - Claude]

Issue #2757, opened 2026-01-18
GET streaming fails when ResponseStreaming metadata message is lost


sanity:
## Summary

GET operations using streaming transport can fail silently when the `ResponseStreaming` metadata message is lost in transit. The stream data arrives and is registered as an orphan, but without the metadata message, the operations layer never claims it. After 30 seconds, the orphan stream is garbage collected and the GET fails.

## Reproduction

This was reported by a user (bbigras) on Matrix. They attempted to load a River invite URL and got an infinite spinner. The issue is reproducible - restarting and retrying produced the same result.

**Environment:**
- Freenet version: 0.1.100
- Client: NixOS (no special firewall ports opened)
- Gateway: nova.locut.us (i4QWRJcwgtBxtQf6)

## Evidence from Diagnostic Report (EPFD7M)

**Timeline:**

| Time | Event | Source |
|------|-------|--------|
| 01:10:07.294 | GET operation created | Client log |
| 01:10:07.347 | GET request received | Gateway telemetry |
| 01:10:07.351 | `get_response_sent` event | Gateway telemetry |
| 01:10:07.354 | Stream transfer started (stream_id=143, 3.68MB) | Gateway telemetry |
| 01:10:07.569 | Stream transfer completed (216ms, 0 timeouts) | Gateway telemetry |
| 01:10:39.008 | **Orphan stream GC'd** (expired_count=1) | Client log |
| 01:10:39.112+ | Fragments arrive with "stream was cancelled" | Client log |

**Key observations:**

1. Gateway completed the transfer successfully with **zero retransmission timeouts**
2. Client was connected (active_connections=2, hole punching worked)
3. Client received stream fragments (orphan was registered)
4. **No `"Processing GET ResponseStreaming"` log** - the metadata message never reached operations layer
5. After 30 seconds (ORPHAN_STREAM_TIMEOUT), the orphan was GC'd
6. Subsequent fragment arrivals failed with "stream was cancelled"

## Root Cause Analysis

The streaming GET response involves two separate transmissions:

1. **ResponseStreaming message** - Small metadata packet containing `stream_id`, `key`, `total_size`
2. **Stream data** - Fragmented payload with ACK-based reliability

The `ResponseStreaming` message is sent as a **single unreliable UDP packet**. Unlike stream fragments which have retransmission on timeout/duplicate-ACK, this metadata message has no reliability mechanism. If lost:

- Stream fragments arrive and register as orphan
[...]

Issue #4155, opened 2026-05-19
Slow multi-hop GET latency for low-popularity contracts (18s for 3 hops, healthy network)


sanity:
## Symptom

GET against a low-popularity contract takes ~18s end-to-end on a healthy node, with routing fanning through 3+ hops before reaching a peer that holds the contract. Even excluding pathological cases (see #4154 for the connection-churn variant), a successful retrieval on a quiet network shouldn't take this long.

## Evidence from telemetry (2026-05-19, contract `6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9`)

Successful retry (TX `01KRZ6BDQE7DG7W8V2BCDMX0G2`), originating at `jMphpuy7z7PPG4Xt@73.98.109.226`:

```
04:01:53.823  get_request  jMphpuy7  -> self (originating)
04:01:53.823  get_request  jMphpuy7  -> c1Ecuk8h@181.85.209.167
04:01:53.902  get_request  4dbrNMxG  -> self
04:01:53.905  get_request  4dbrNMxG  -> jMphpuy7
04:01:54.029  get_request  c1Ecuk8h  -> self
04:01:54.035  get_request  c1Ecuk8h  -> 145.224.74.209
04:01:54.180  get_request  rUoqhXtx  -> self
04:01:54.182  get_request  rUoqhXtx  -> 4x2WRDv1@87.106.68.246
04:01:54.204  get_request  4x2WRDv1  -> self
04:01:54.210  get_response_sent  4x2WRDv1  (this peer had the contract)
04:02:05.840  get_success  c1Ecuk8h  (11.6s after forwarding)
04:02:07.367  get_success  jMphpuy7  (originator, 13.5s end-to-end)
04:02:11.990  get_success  4dbrNMxG  (parallel branch, 18s)
```

Routing walked the chain `jMphpuy7 → c1Ecuk8h → rUoqhXtx → 4x2WRDv1` before finding a peer holding the contract. `4x2WRDv1` responded with `get_response_sent` within ~6ms of receiving the request, so the upstream nodes were fast — the latency is entirely in the multi-hop traversal plus what looks like ~11s of dwell time at `c1Ecuk8h` between receiving the response and forwarding it back (04:01:54.210 → 04:02:05.840).

## Two distinct contributors

[...]

Issue #3608, opened 2026-03-20
fix: streaming transfers for large contracts (~1MB) fail mid-transfer


sanity:
## Problem

GET requests for large contracts (~1MB, e.g. the River UI container `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) consistently fail because streaming transfers die mid-transfer. The responding peer starts sending fragments but they stop arriving, causing stream assembly timeouts.

**User impact:** River UI cannot load in the browser. Users see "GET request timed out after 30s". Affects users with otherwise healthy nodes (16+ ring peers).

## Evidence

Two diagnostic reports from DavidHUK (macOS aarch64, v0.2.7, 16 peers):

**Report 23TN9A** (2026-03-20T19:50): Three GET attempts, all failed:
1. ResponseStreaming from `69.162.253.175:38520` → orphan stream timeout (60s, no fragments claimed)
2. ResponseStreaming → stream assembly timed out after 30s (fragments stopped arriving)  
3. ResponseStreaming from `143.105.209.219:19054` → orphan stream timeout (60s)

**Report E7NJG5** (2026-03-20T14:30): Same pattern:
- ResponseStreaming from `79.130.18.151:33943` → total_size=1,046,334 bytes, 737 fragments expected → only 1 fragment received → 30s inactivity timeout

**Gateway perspective:** The nova gateway does NOT hold this contract locally — it relay-fetches from the network (28s fetch latency observed), adding an extra relay hop that may contribute to transfer failures.

## Key Observations

- The contract is ~1MB (1,046,334 bytes), split into 737 fragments for streaming
- Multiple different responding peers fail the same way — not a single bad peer
- The stream starts (ResponseStreaming received) but fragments stop arriving
- Two failure modes: (a) stream assembly timeout — some fragments arrive then stop, (b) orphan stream timeout — stream header arrives but no fragments are ever claimed
- Node connectivity is healthy (16 peers, successful CONNECT operations)
- Smaller contracts/state transfers work fine on the same node

## Potentially Relevant Code

- `crates/core/src/transport/peer_connection/streaming.rs` — stream assembly, fragment reassembly
- `crates/core/src/operations/get.rs` — GET operation, ResponseStreaming handling
- `crates/core/src/operations/orphan_streams.rs` — orphan stream claiming
- `crates/core/src/node/network_bridge/p2p_protoc.rs` — piped stream forwarding (relay path)

## Questions

[...]

PR #4308, opened 2026-05-28
fix(get): restore relay GET response streaming (producer + forward)

### LINKED ISSUE #3883: Migrate relay GETs to task-per-tx driver (#1454 follow-up to Phase 3b)

### LINKED ISSUE #4307: Restore GET response streaming dropped in task-per-tx migration (producer + relay forward; #3883 §7)
## Summary

The task-per-tx GET migration (#1454 Phase 5, commits `bc8ee9d0` / `9edade0c`) dropped GET response **streaming entirely** — not just the relay-forward slice that "port plan §7" tracked. Both halves are gone:

1. **Producer side**: `relay_send_found` (`crates/core/src/operations/get/op_ctx_task.rs:1894`) and the local-cache-hit responder always build an inline `GetMsg::Response{Found}` regardless of payload size. There is no `should_use_streaming` gate, no `StreamId::next_operations()`, no `send_stream`/`stream_data` anywhere in the GET op code. (Confirmed by `grep`: GET has zero `ResponseStreaming` *construction* sites in production code — only the `classify` consumer at `op_ctx_task.rs:671` and a test at `:2547`.)

2. **Relay-forward side (port plan §7)**: the `Terminal::Streaming` arm in `drive_relay_get_inner` (`op_ctx_task.rs:2380`) logs "streaming relay forwarding not yet implemented" and skips to the next peer. Pinned by the R12b regression guard `streaming_downstream_is_currently_warned_and_skipped`.

Because the producer is also gone, no peer in an all-updated network ever emits a `GetMsg::ResponseStreaming` for GET, so the relay-forward arm is currently dead in practice and the originator's `assemble_and_cache_stream` is defensive-only.

## Why this matters (regression, not just a missing optimization)

Large GET payloads still load correctly — they fall back to the transport's serial message-level fragmentation (store-and-forward). But that is exactly the slow path that streaming was introduced to replace. Per the original streaming fix (`f9fc0adc`, "fix: relay GET response forwarding now uses streaming for large payloads (#3586)"):

> serial store-and-forward ... adds 10-50 s per hop for large contracts like the River UI at ~814 KB

So multi-hop GET of large contracts (River being the flagship case) is now **10–50s slower per relay hop** than before the migration. Correctness is preserved; latency regressed.

## Scope of the fix (end-to-end — both halves are required)

Implementing only §7 (relay forward) is useless without the producer, and vice-versa. Finish-the-fix = restore both:

### A. Producer (restore #3586 behavior in the task-per-tx driver)
In the responder paths (`relay_send_found` and the local-cache-hit branch), when the serialized `GetStreamingPayload` exceeds `op_manager.streaming_threshold`, emit `GetMsg::ResponseStreaming { stream_id: StreamId::next_operations(), total_size, includes_contract, .. }` and send the payload bytes via the streaming transport (`send_stream`) instead of an inline `Response{Found}`. The deleted reference implementation is recoverable at `git show f9fc0adc:crates/core/src/operations/get.rs`.

### B. Relay forward (port plan §7)
Replace the warn+skip in the `Terminal::Streaming` arm (`op_ctx_task.rs:2380`) with the fork+pipe pattern, mirroring PUT's proven `drive_relay_put_streaming` (`crates/core/src/operations/put/op_ctx_task.rs:1655`):
1. `orphan_stream_registry().claim_or_wait(downstream_peer_addr, stream_id, STREAM_CLAIM_TIMEOUT)` → inbound `StreamHandle`.
2. `outbound_sid = StreamId::next_operations()`, `forked = handle.fork()`.
3. Send a fresh `GetMsg::ResponseStreaming` header (carrying `outbound_sid`) to `upstream_addr`, with the serialized header embedded for fragment #1 (reliability, fix #2757). **Forward upstream before any local assemble/cache work** (`forward-before-caching` rule).
4. `conn_manager.pipe_stream(upstream_addr, outbound_sid, forked, embedded_metadata)`.
5. Record a relay route event (`SuccessUntimed`); optionally assemble the inbound handle in parallel to cache locally + `announce_contract_hosted` (legacy parity — gated on first-time hosting).

### C. Cleanup
- Remove / invert the R12b pin test (`streaming_downstream_is_currently_warned_and_skipped`) — it asserts the *absence* of the behavior we're adding.
- Stale doc: `operations.md` calls the bypass fn `try_forward_task_per_tx_reply`; the actual fn is `try_forward_driver_reply` (`node.rs:679`).

## Constraints / footguns (from prior incidents)
- **cwnd-wait deadlock (#3608 / #3586)**: the BBR cwnd-wait loop in `outbound_stream::pipe_stream` requires the upstream connection's recv loop to keep processing ACKs or the pipe blocks forever; `CWND_WAIT_TIMEOUT` mitigates. The relay task must not block the connection that carries those ACKs.
- **StreamRegistry deadlock (#3519)** and **#4231 pipe_stream timeout-reserve**: reuse the existing `pipe_stream` bridge call (pinned by `pipe_stream_branch_uses_timed_reserve_pattern`), don't hand-roll fragment sends.
- **claim-key**: relay claims against the *downstream peer's* address (the sender of the inbound stream), and re-keys the outbound stream with a fresh `StreamId` per hop (two thread-local counters on different nodes can collide, hence the `(SocketAddr, StreamId)` registry key).

## Testing
- Multi-hop simulation (DST) test: GET of a >threshold contract over ≥2 relay hops asserts the originator receives the assembled state AND that intermediate hops piped (not store-and-forwarded). 
- Source-scrape pin: relay `Terminal::Streaming` arm forwards-before-assembles (mirror PUT's `relay_driver_forwards_upstream_before_caching_on_found_response`).
- Producer pin: responder emits `ResponseStreaming` above threshold, inline `Response{Found}` below.

## References
- Follow-up to #3883 (closed; non-streaming relay migration landed).
- Reverted-by-migration fix: `f9fc0adc` (#3586). Related: #3608, #3519, #2757, #4231.
- PUT streaming relay precedent: `crates/core/src/operations/put/op_ctx_task.rs:1655`.

[AI-assisted - Claude]

sanity:
## Problem

The task-per-tx GET migration (#1454 Phase 5, commits `bc8ee9d0` / `9edade0c`) dropped GET response **streaming** entirely — both the producer and the relay-forward halves:

- `relay_send_found` always built an inline `GetMsg::Response{Found}` regardless of payload size (no `should_use_streaming` gate, no `send_stream`).
- The relay `Terminal::Streaming` arm logged "not yet implemented (port plan §7)" and skipped to the next peer.

Large GET payloads still load (they fall back to serial transport-level message fragmentation), but that is exactly the slow path streaming was introduced to replace. Per the original fix `f9fc0adc` ("relay GET response forwarding now uses streaming for large payloads", #3586): serial store-and-forward **adds 10–50s per hop for large contracts like the River UI at ~814 KB**. So multi-hop GET of large contracts regressed badly in latency. This restores the streaming fast path. Closes #4307 (follow-up to the now-closed #3883).

## Approach

Mirrors the proven PUT streaming relay (`start_relay_put` / `drive_relay_put_streaming`):

- **Thread `conn_manager: CB` (`NetworkBridge + Clone + Send + 'static`)** through `start_relay_get` → `run_relay_get` → `drive_relay_get` → `drive_relay_get_inner`, and update the `node.rs` GET dispatch site to pass it.
- **Producer (`relay_send_found`)**: for a remote upstream, serialize `GetStreamingPayload`; if it exceeds `streaming_threshold`, send a `GetMsg::ResponseStreaming` header (fire-and-forget — the upstream already has its waiter) then push fragments via `conn_manager.send_stream`. Loopback (`upstream == own_addr`) keeps the inline path (no network benefit). Below-threshold stays inline.
- **Relay forward (`Terminal::Streaming` arm)**: `claim_or_wait` the inbound stream from the downstream peer → `fork()` the handle → mint a **fresh** `outbound_sid` → send a re-keyed `ResponseStreaming` header upstream (with the serialized header embedded for fragment #1, #2757) → `conn_manager.pipe_stream` the forked handle — all **before** the best-effort local `cache_contract_locally` (forward-before-cache, #4155). Records a `SuccessUntimed` relay route event and registers the downstream subscriber when `subscribe`. A loopback safety net assembles+caches locally instead of piping to self.

## Testing

- `cargo test -p freenet --lib operations::get`: **72 passed, 0 failed**.
- Removed the now-false R12b pin (`streaming_downstream_is_currently_warned_and_skipped`) that asserted the absence of this behavior; added pins: producer streams large remote payloads, streaming arm forwards-before-caches, streaming arm uses a fresh outbound stream id.
- `cargo clippy -p freenet --lib --tests` (default features): clean.

## Notes / for reviewers

- **High-risk surface** (streaming/concurrency + wire protocol) — warrants Full multi-model review.
- `handle.fork()` shares the buffer, so the held handle still `assemble()`s for local caching while the fork pipes — same pattern as PUT.
- **Pre-existing, unrelated:** `cargo clippy -p freenet --all-features` fails to compile `tracing.rs` (opentelemetry `SpanBuilder` API drift) on base `main` too — not introduced here.
- cwnd-wait deadlock footgun (#3608) is handled by reusing the existing `pipe_stream` bridge path (same as PUT), not hand-rolled fragment sends.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -35,7 +35,7 @@ for every inbound wire message. Pattern per op:
 
 1. **Reply bypass.** If a terminal reply variant arrives and a
    `pending_op_results` callback is registered, forward it via
-   `try_forward_task_per_tx_reply` and return. For GET/PUT/SUBSCRIBE
+   `try_forward_driver_reply` and return. For GET/PUT/SUBSCRIBE
    the gate is `Response | ResponseStreaming` only. CONNECT forwards
    all four non-`Request` variants (multi-reply fan-in).
 2. **Relay dispatch.** Spawn the matching `start_relay_*` driver.
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -108,6 +108,8 @@ pub mod dev_tool {
     #[cfg(any(test, feature = "testing"))]
     pub use crate::operations::get::op_ctx_task::RELAY_DRIVER_CALL_COUNT as GET_RELAY_DRIVER_CALL_COUNT;
     #[cfg(any(test, feature = "testing"))]
+    pub use crate::operations::get::op_ctx_task::RELAY_GET_STREAMING_FORWARD_COUNT;
+    #[cfg(any(test, feature = "testing"))]
     pub use crate::operations::put::op_ctx_task::RELAY_PUT_DRIVER_CALL_COUNT;
     #[cfg(any(test, feature = "testing"))]
     pub use crate::operations::put::op_ctx_task::RELAY_PUT_STREAMING_DRIVER_CALL_COUNT;
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1040,6 +1040,7 @@ where
                     } => {
                         if let Err(err) = get::op_ctx_task::start_relay_get(
                             op_manager.clone(),
+                            conn_manager.clone(),
                             *id,
                             *instance_id,
                             *htl,
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -103,6 +103,18 @@ pub static RELAY_COMPLETED_TOTAL: std::sync::atomic::AtomicUsize =
 pub static RELAY_DEDUP_REJECTS: std::sync::atomic::AtomicUsize =
     std::sync::atomic::AtomicUsize::new(0);
 
+/// Test-only counter that increments every time the relay GET driver
+/// commits to forking + piping a streaming response to a REMOTE upstream
+/// (the `Terminal::Streaming` forward branch). Incremented after the
+/// loopback safety-net check and a successful `claim_or_wait`, before the
+/// header send. NOT incremented on the loopback branch or the
+/// `AlreadyClaimed` dedup branch. Used by `streaming_e2e` to prove a GET
+/// genuinely relayed through a streaming hop rather than falling back to
+/// store-and-forward. Mirrors `RELAY_PUT_STREAMING_DRIVER_CALL_COUNT`.
+#[cfg(any(test, feature = "testing"))]
+pub static RELAY_GET_STREAMING_FORWARD_COUNT: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+
 /// Start a client-initiated GET, returning as soon as the task has been
 /// spawned (mirrors legacy `request_get` timing).
 ///
@@ -997,11 +1009,12 @@ async fn cache_contract_locally(
 /// originator for driver ops (`load_or_init` would return
 /// `OpNotPresent`).
 ///
-/// `peer_addr` is the sender's transport address — currently we
-/// use `driver.current_target.socket_addr()`, which is accurate for
-/// single-hop responses. Multi-hop (where a relay answers on behalf
-/// of a further peer) is not yet supported by the driver;
-/// see #3883.
+/// `peer_addr` is the sender's transport address — we use
+/// `driver.current_target.socket_addr()`. Multi-hop streamed GET IS now
+/// supported: relays fork+pipe the stream and re-key it with a fresh
+/// outbound `stream_id` per hop (#4307), so the originator always claims
+/// the stream keyed by its `current_target` (the immediate next hop),
+/// regardless of how many relays the response traversed. See #3883.
 async fn assemble_and_cache_stream(
     op_manager: &OpManager,
     peer_addr: std::net::SocketAddr,
@@ -1051,10 +1064,11 @@ async fn assemble_and_cache_stream(
         None
     };
 
-    // Streaming assembly is reachable only from the client-driver side (the
-    // relay driver does not claim streams — port plan §7). The originator
-    // IS the client requester, so the sticky `local_client_access` flag
-    // applies.
+    // This helper runs on the client-driver (originator) side. The relay
+    // driver also claims streams now (the `Terminal::Streaming` arm, #4307),
+    // but it uses its own inline assemble+cache with `local_client_access =
+    // false`; this path is the originator, which IS the client requester, so
+    // the sticky `local_client_access` flag applies here.
     cache_contract_locally(op_manager, payload.key, state, contract, true).await;
     Ok(())
 }
@@ -1492,13 +1506,11 @@ fn missing_state_cause(instance_id: &ContractInstanceId) -> String {
 // targeting the downstream peer, and the reply (Found / NotFound) is forwarded
 // to `upstream_addr` via a fire-and-forget `conn_manager.send`.
 //
-// Out of scope for this commit (§7 of the port plan):
-//   - ResponseStreaming relay forwarding (chunk pipe-through).
+// ResponseStreaming relay forwarding (chunk pipe-through) IS now implemented
+// in the `Terminal::Streaming` arm of `drive_relay_get_inner` (fork + pipe;
+// issue #4307). Still out of scope for the driver:
 //   - GC-spawned retries.
 //   - start_targeted_op (UPDATE-triggered auto-fetch).
-//
-// This entire section is dead code in commit 1 — node.rs dispatch is not
-// changed until commit 2.
 
 /// Start a relay (non-originator) GET driver.
 ///
@@ -1507,8 +1519,11 @@ fn missing_state_cause(instance_id: &ContractInstanceId) -> String {
 /// owns routing/forwarding/retry state in its task locals — no `GetOp`
 /// is stored in `OpManager.ops.get` for this transaction.
 ///
-/// Originator loop-back (`source_addr.is_none()`) continues to use the
-/// legacy `handle_op_request` → `process_message` path.
+/// Originator loop-back (`source_addr.is_none()`) is mapped by `node.rs`
+/// dispatch to `upstream_addr = own_addr` and driven by this same driver
+/// (the legacy `process_message` path is deleted); the loopback edge is
+/// handled by the safety-net branches that cache locally instead of
+/// piping/replying to self.
 ///
 /// # Scope (#3883)
 ///
@@ -1517,10 +1532,10 @@ fn missing_state_cause(instance_id: &ContractInstanceId) -> String {
 ///   interest gate), forward-or-respond decision.
 /// - `GetMsg::Response{NotFound}` retry to next alternative peer.
 /// - `GetMsg::Response{Found}` bubble-up to upstream.
+/// - `GetMsg::ResponseStreaming` relay forwarding (fork + pipe, #4307).
 /// - `ForwardingAck` send-before-forward.
 ///
-/// NOT migrated (stays on legacy path; see port plan §7):
-/// - Streaming-chunk relay forwarding (`ResponseStreaming` pass-through).
+/// NOT migrated (stays on legacy path):
 /// - GC-spawned retries.
 /// - `start_targeted_op` (UPDATE-triggered auto-fetch).
 #[allow(clippy::too_many_arguments)]
@@ -1991,6 +2006,21 @@ where
             total_size: payload_bytes.len() as u64,
             includes_contract,
         });
+        // Embed a serialized copy of the metadata header in fragment #1
+        // (fix #2757). On a lossy link the separately-sent fire-and-forget
+        // header can be lost while fragments arrive — without self-describing
+        // metadata the upstream waiter never matches the orphan stream and
+        // the GET times out. Mirrors the forward path's `embedded`.
+        let embedded = bincode::serialize(&header).ok().map(bytes::Bytes::from);
+        if embedded.is_none() {
+            tracing::warn!(
+                tx = %tx,
+                target = %upstream_addr,
+                %sid,
+                "GET relay: failed to serialize ResponseStreaming header for \
+                 fragment-#1 embedding; sending stream without embedded metadata"
+            );
+        }
         // Install nothing — the upstream already has its waiter (it sent
         // us the Request). Send the metadata header first, then push the
         // raw fragments. Ordering matches the PUT producer.
@@ -1998,7 +2028,12 @@ where
             .await
             .map_err(|_| OpError::NotificationError)?;
         conn_manager
-            .send_stream(upstream_addr, sid, bytes::Bytes::from(payload_bytes), None)
+            .send_stream(
+                upstream_addr,
+                sid,
+                bytes::Bytes::from(payload_bytes),
+                embedded,
+            )
             .await
             .map_err(|e| {
                 OpError::NotificationChannelError(format!("get relay send_stream failed: {e}"))
@@ -2535,7 +2570,13 @@ where
 
                 // ── Step 2: Loopback safety net. If upstream IS us
                 //    (originator-loopback edge — rare), assemble + cache
-                //    locally instead of piping to self.
+                //    locally instead of piping to self. This branch is
+                //    cache-only with NO wire Response: relay-loopback
+                //    initiators (GC-spawned retries / start_targeted_op
+                //    auto-fetch) only need the local cache populated — the
+                //    interactive client GET uses the separate
+                //    `drive_client_get_inner` driver, which owns its own
+                //    client delivery.
                 if Some(upstream_addr) == own_addr {
                     match handle.assemble().await {
                         Ok(bytes) => match bincode::deserialize::<GetStreamingPayload>(&bytes) {
@@ -2600,6 +2641,11 @@ where
                 // Serialize metadata for embedding in fragment #1 (fix #2757).
                 let embedded = bincode::serialize(&header_net).ok().map(bytes::Bytes::from);
 
+                // Committed to forking + piping to a remote upstream: count
+                // this streaming-forward hop (test-only; see counter doc).
+                #[cfg(any(test, feature = "testing"))]
+                RELAY_GET_STREAMING_FORWARD_COUNT.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+
                 // Send the metadata header BEFORE piping. The upstream
                 // already has its waiter installed (it sent us the Request),
                 // so a fire-and-forget header suffices.
@@ -2611,14 +2657,32 @@ where
 
                 // `pipe_stream` spawns a background task and returns
                 // promptly, so this does not block on the full transfer.
-                conn_manager
+                //
+                // Do NOT `?` this error: the `ResponseStreaming` header has
+                // already gone upstream (see send above), so the upstream is
+                // committed to assembling a stream. Propagating the error
+                // would bubble up to `drive_relay_get`, which sends a
+                // compensating `Response{NotFound}` to the SAME upstream — a
+                // contradictory double-signal that stalls the upstream ~60s
+                // on `claim_or_wait`. Instead, log and return cleanly so the
+                // upstream's claim simply times out without a NotFound. We do
+                // NOT fall through to the local-cache step: the held `handle`
+                // may be partial after a pipe failure.
+                if let Err(e) = conn_manager
                     .pipe_stream(upstream_addr, outbound_sid, forked, embedded)
                     .await
-                    .map_err(|e| {
-                        OpError::NotificationChannelError(format!(
-                            "get relay pipe_stream failed: {e}"
-                        ))
-                    })?;
+                {
+                    tracing::warn!(
+                        tx = %incoming_tx,
+                        %upstream_addr,
+                        %outbound_sid,
+                        error = %e,
+                        "GET relay: pipe_stream failed AFTER the ResponseStreaming \
+                         header already went upstream; letting the upstream's claim \
+                         time out cleanly rather than sending a contradictory NotFound"
+                    );
+                    return Ok(());
+                }
 
                 // ── Step 4: Record the relay route event for the downstream
                 //    peer that served the stream (it behaved correctly).
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -1998,6 +1998,13 @@ where
             phase = "relay_streaming_forward",
             "GET relay: payload exceeds threshold, streaming response upstream"
         );
+        // NOTE: `hop_count` is intentionally NOT carried on the streaming
+        // wire path — the `ResponseStreaming` variant has no such field
+        // (unlike the inline `Response{Found}` below, which propagates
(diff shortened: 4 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4308

### Summary
- **PR Title:** fix(get): restore relay GET response streaming (producer + forward)
- **Type:** fix (high-risk surface: streaming/concurrency + wire protocol)
- **CI Status:** Clippy / Fmt / Rule Lint / Conventional Commits / macOS E2E green at review time; heavier suites pending
- **Linked Issues:** Closes #4307 (follow-up to closed #3883)
- **Review tier:** Full
- **Reviewers run:** code-first, testing, skeptical, big-picture (all `freenet:` agents) + external Codex

Verdict up front: the implementation faithfully mirrors the proven PUT streaming relay, the producer/forward/originator halves compose end-to-end, and the most dangerous concurrency hazards were independently verified safe. There are a handful of real, addressable findings before merge.

---

### Code-First Analysis
**Alignment:** matches the stated intent and the #4307 spec — both halves (producer in `relay_send_found`, fork+pipe forward in the `Terminal::Streaming` arm) are present. End-to-end composition verified: storer streams → relay re-keys + forks + pipes → originator `assemble_and_cache_stream` claims by `(current_target, stream_id)`. PUT-pattern fidelity confirmed.
**Gaps:** none material; stale comments (see Documentation).

---

### Testing Assessment
**Coverage Level:** insufficient for a streaming/concurrency change.

| Unit | ⚠️ | 3 new tests are source-scrape pins only — they assert code *shape*, execute nothing |
| Integration/Sim | ❌ | No test that fails without the fix. `test_streaming_get_through_relay` (streaming_e2e.rs:604) is probabilistic and asserts only final bytes — it does not gate on the relay-streaming path actually running, so it likely passes on `main` despite the broken path |
| E2E | N/A | |

**Regression Test:** missing. PUT gates its multi-hop streaming test on `RELAY_PUT_STREAMING_DRIVER_CALL_COUNT`; there is no GET equivalent. Recommend a `RELAY_GET_STREAMING_FORWARD_COUNT` counter + a topology-forced ≥2-hop GET test asserting the counter advanced AND originator state == source.

---

### Skeptical Findings
**Risk Level:** medium (one real correctness issue; rest robustness/polish)

| pipe_stream failure → contradictory NotFound + 60s upstream stall | **Medium** | op_ctx_task.rs:2614-2621 | Header already sent upstream; `?` on `pipe_stream` Err triggers compensating `NotFound` in `drive_relay_get`. Upstream classified the header as terminal Streaming and is committed to `claim_or_wait` (60s). Fix: on pipe Err, log + `return Ok(())` (don't `?`). The header-send `?` *before* pipe is fine (header never went). |
| Producer omits fragment-#1 embedded metadata | Low→Med | op_ctx_task.rs:2000-2001 | `send_stream(.., None)`: on a lossy link the fire-and-forget header can be lost while fragments arrive → orphan stream with no self-describing metadata → upstream waiter never matches → GET times out. The pipe-forward path embeds it (#2757); producer should too. (Also raised by Codex.) |
| Silent serialize-error swallow | Low | op_ctx_task.rs:2601 | `bincode::serialize(&header_net).ok()` drops errors with no `warn!`; PUT warns. |

**Verified SAFE (investigated):** #3608 cwnd deadlock (driver runs in spawned task, not a recv loop; pipe ACKs on a different connection's loop; bounded by CWND_WAIT_TIMEOUT); fork+reclaim (pipe uses non-reclaiming `stream()`, fork shares buffer, held handle still assembles); fire-and-forget header ordering (orphan registry makes fragments-first benign); state cleanup (`RelayInflightGuard` RAII + unconditional `release_pending_op_slot`); claim error/AlreadyClaimed paths (bounded 60s orphan GC).

**Open question:** the loopback streaming branch caches-only with no `send_local_loopback` Response. Confirm the only relay-loopback initiators (GC retry, `start_targeted_op`) never install a waiter expecting a wire Response.

---

### Big Picture Assessment
**Goal Alignment:** yes — genuinely fixes the stated regression (multi-hop large-GET latency), full scope, mirrors a proven pattern. Not CI-chasing.
**Anti-Patterns Detected:** none. No `#[ignore]`, no weakened assertions, no error-swallowing on the critical path.
**Removed Code Concerns:** the deleted `streaming_downstream_is_currently_warned_and_skipped` pin was genuinely obsolete (asserted absence of this behavior). The ~20 needle edits are mechanical (`(` → `<CB>(`) and not weakened (passing run proves `extract_fn_body` still resolves). The cache-count pins (3→5) correctly account for the 2 new streaming-arm caches — not vacuous.
**Parity vs legacy:** `announce_contract_hosted` preserved (via `cache_contract_locally` first-time-hosting gate); subscriber registration matches the R4 active-interest path (and is ahead of the R12a inline path which registers none) — no regression.
**Scope:** focused; no unrelated changes. No conflict with adjacent open PRs #4291 / #4304.

---

### Documentation
Three in-code comments now **actively contradict** the new code and should be fixed in this PR (the issue's section C asked for this cleanup):
- op_ctx_task.rs:1495-1496 — "Out of scope (§7): ResponseStreaming relay forwarding" → now the headline feature.
- op_ctx_task.rs:1522-1523 — "NOT migrated: Streaming-chunk relay forwarding" → now migrated in this fn.
- op_ctx_task.rs:1054-1055 — "the relay driver does not claim streams — §7" → it now claims.
Lower priority (pre-existing, but the PR already edits these blocks): :1003 ("multi-hop not yet supported"), :1500-1501 ("dead code in commit 1"), :1510-1511 ("loopback uses legacy process_message").

---

### Recommendations

#### Must Fix (Blocking)
1. **pipe_stream failure handling (Medium):** on `pipe_stream` Err in the streaming arm, log and `return Ok(())` rather than `?` — avoid the contradictory `NotFound` after the header and the 60s upstream claim stall.

#### Should Fix (Important)
1. **Producer embedded metadata:** serialize the `ResponseStreaming` header and pass it to `send_stream` for fragment-#1 robustness (lossy-link delivery), matching the forward path and #2757.
2. **Behavioral regression test:** add a `RELAY_GET_STREAMING_FORWARD_COUNT` counter and a topology-forced multi-hop GET sim test that fails without the fix (the current pins do not).
3. **Stale §7 comments:** fix the three contradicting comments above.

#### Consider (Suggestions)
1. `warn!` on the embedded-metadata serialize error (parity with PUT).
2. Document why the loopback streaming branch is cache-only (no wire Response).
3. Optional R4-style `else { register_peer_interest }` for the `!subscribe` case (pre-existing asymmetry; defer OK).

---

### Verdict
**HEAD SHA reviewed:** 3c9c0525164c2a3fb61c5358adcf768e00920d2a

One Medium correctness fix (pipe failure), two Should-Fix items (producer metadata robustness, behavioral regression test), and stale-comment cleanup. Core design is sound and the dangerous concurrency paths verified safe. A focused re-review on the updated HEAD is warranted because the pipe-failure fix changes behavior in a high-risk area.

[AI-assisted - Claude]

sanity (commented):
## Re-review after fixes — HEAD `be162c93`

All findings from the prior review (HEAD `3c9c0525`) have been addressed in `be162c93`. Focused re-review (skeptical perspective + external Codex) on the new HEAD: **clean — no new bugs introduced.**

| Prior finding | Status |
|---|---|
| Must Fix — `pipe_stream` failure → contradictory NotFound + 60s upstream stall | ✅ Fixed: pipe failure now logs + `return Ok(())` (no `?`); header-send failure still `?`s (NotFound correct there). State cleanup (`RelayInflightGuard`, `release_pending_op_slot`) verified to run on the new return path. |
| Should Fix — producer omits fragment-#1 embedded metadata | ✅ Fixed: `send_stream` now passes the serialized `ResponseStreaming` header (byte-identical to the fire-and-forget header — same `sid`/fields); matches the forward path, #2757, and the PUT/UPDATE producer convention. |
| Should Fix — no behavioral regression test | ✅ Fixed: added `RELAY_GET_STREAMING_FORWARD_COUNT`; `test_streaming_get_through_relay` now uses a sparse topology (1 gw + 6 nodes, `max_connections=3`) that deterministically forces a relay hop and asserts the counter advanced (before=0, after=1). Non-vacuous — fails without the fix. |
| Should Fix — stale §7 / "not implemented" comments | ✅ Fixed: comments at op_ctx_task.rs:1009, 1064, 1506-1539 now accurately describe the implemented behavior. |
| Consider — `warn!` on embedded-metadata serialize error; document loopback cache-only branch | ✅ Both done. |

Verified safe (carried from prior review): #3608 cwnd deadlock, fork+reclaim buffer semantics, fire-and-forget header ordering, full state cleanup.

**Verdict:** Ready to Merge (pending CI green on `be162c93`). Re-review HEAD SHA: `be162c93c6a05dd6eee698d12980187a29d2749f`.

[AI-assisted - Claude]

sanity:
## CI status on `be162c93`

All checks green **except** the `Unit & Integration` job, whose sole hard failure is `test_ping_multi_node` — the known pre-existing flake tracked in #4295 ("fails on main HEAD + multiple PRs, ~50%+ rate"). It is a ping-history eventual-consistency propagation timing issue, driven by PUT/UPDATE/subscribe; this PR's change is GET-relay-streaming only and cannot affect it.

nextest summary: `3145 tests run: 3144 passed (1 flaky, 3 leaky), 1 failed, 19 skipped`.
- The 1 hard `failed` = `test_ping_multi_node` (#4295, unrelated flake).
- The 1 `flaky` = `test_put_triggers_update_for_subscribers` (passed on retry 2 — FLAKY 2/3).
- This PR's tests passed: `relay_send_found_streams_large_payload_to_remote`, and the full **Simulation** suite incl. the new counter-gated `test_streaming_get_through_relay`.

Not merging — leaving for review. The red check is the tracked flake, not a regression here.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4308

### Summary
- **PR Title:** fix(get): restore relay GET response streaming (producer + forward)
- **Type:** fix (high-risk surface — streaming/concurrency + wire protocol)
- **CI Status:** all checks green on the reviewed HEAD except the `rule-review/warnings` gate, now addressed (see below); re-running on the updated HEAD
- **Linked Issues:** closes #4307 (follow-up to closed #3883)
- **Review tier:** Full (high-risk surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture (4 perspectives) + Codex (external, non-Claude) + Gemini-flash (external). Gemini-2.5-pro was unavailable (quota 429).

This review was run on HEAD `9236fa24`. Findings were then addressed in `55a9b9de`.

---

### Code-First Analysis
**Independent understanding matches stated intent — fully.** The PR threads `conn_manager: CB` (`NetworkBridge + Clone + Send + 'static`) through `start_relay_get → run_relay_get → drive_relay_get → drive_relay_get_inner`/`relay_send_found`, and updates the single `node.rs` dispatch site. It restores both halves dropped by the task-per-tx migration: the **producer** (`relay_send_found` streams >threshold payloads to a remote upstream) and the **relay-forward** (`Terminal::Streaming` arm: claim → fork → fresh outbound sid → re-keyed header → pipe → best-effort cache). The GET-vs-PUT directional asymmetry (GET forwards *upstream* toward a requester that already has a waiter, so a fire-and-forget header suffices and no waiter install is needed) is handled correctly.

---

### Testing Assessment
**Coverage Level:** adequate for the forward path; gaps closed in the fix commit.

| Unit | ✅ | 73 get-op tests pass (was 72; +1 new pin) |
| Integration | ✅ | structural pins for relay driver shape |
| Simulation/E2E | ✅ | `test_streaming_get_through_relay`: real 1MB GET over a relay hop, asserts `RELAY_GET_STREAMING_FORWARD_COUNT` increment + byte-exact reassembly |

**Gaps found and addressed:** the `pipe_stream`-failure-after-header path had no guard, and the fresh-sid pin was weak. Both fixed (see Recommendations → addressed).

---

### Skeptical Findings
**Risk Level:** low. No Must-Fix correctness bug found.

| Concern | Severity | Location | Disposition |
|---------|----------|----------|-------------|
| fork() vs held-handle `assemble()` buffer race | — | `op_ctx_task.rs:2631`/`:2715` | **Safe** — `pipe_stream` uses the non-reclaiming `stream()` view; fragments stay in the shared buffer |
| cwnd-wait deadlock (#3608) | — | `outbound_stream.rs` | **Safe** — pipe runs in a cwnd-bounded spawned task; driver doesn't block |
| Wire/back-compat | — | `get.rs:120` | **Safe** — `ResponseStreaming` is a pre-existing variant, unchanged |
| claim_or_wait failure records no route Failure | Should Fix | `op_ctx_task.rs:2557` | **Fixed** in `55a9b9de` |

---

### Big Picture Assessment
**Goal Alignment:** yes, end-to-end verified. **Removed code:** the R12b pin (`streaming_downstream_is_currently_warned_and_skipped`) asserted the *absence* of the behavior being restored — correctly inverted into 3 presence-pins, not silently dropped. **Scope:** tight (2 commits, 5 files, no creep). **Docs/rules:** `.claude/rules/operations.md` updated (`try_forward_task_per_tx_reply` → `try_forward_driver_reply`); all stale "port plan §7 not yet implemented" notes removed; no doc/skill left contradicting the new behavior.

---

### Documentation
- Code docs: complete (heavily commented; added a note that `hop_count` is intentionally absent from the streaming wire path).
- Rules: up-to-date.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important) — all addressed in `55a9b9de`
1. **Record `RouteOutcome::Failure` on the streaming-arm `claim_or_wait` failure** (skeptical). A downstream that advertises a `ResponseStreaming` header then never delivers is a routing failure; the two transport-failure arms already train the router, this one didn't — so a stalling peer kept being re-selected. **Fixed**; `AlreadyClaimed` deliberately left uncounted (dedup).
2. **Guard the `pipe_stream`-failure invariant with a test** (rule-review CI gate + testing). The deliberate `return Ok(())` (avoiding a contradictory NotFound) had no coverage and the mock can't fail. **Fixed** with structural pin `relay_streaming_arm_pipe_failure_returns_ok_without_notfound`.

#### Consider — addressed or deferred
- **Strengthen the fresh-sid pin** to assert `outbound_sid` is the value re-keyed into the header AND piped (testing). **Done.**
- **`hop_count` not carried on the streaming wire path** (code-first). **Documented** with a comment; behavior matches the legacy producer and is compensated by `total_size`.
- *Deferred (defensible, match PUT convention):* a producer-branch counter; a loopback-safety-net pin; aligning R12a's (pre-existing) subscriber-forwarding asymmetry. Filed as follow-up considerations, not blocking.

---

### Verdict
**State:** Ready to Merge (after CI re-confirms green on the updated HEAD `55a9b9de`).
**HEAD SHA reviewed:** `9236fa24`; fixes applied in `55a9b9de`.

No Must-Fix findings from any of the six reviewers. The two Should-Fix items (route-Failure training, pipe-fail test guard) and the Consider items have been addressed in `55a9b9de`. The fix commit is additive (route-event call + comments + source-pins), so it does not alter the streaming data path the review validated. Merge once Clippy + `rule-review/warnings` + the heavy suites are green on `55a9b9de`.

[AI-assisted - Claude]

Issue #4301, opened 2026-05-28
Phase 7 follow-up: end-to-end simulation test for governance ban chain


sanity:
## Context

Spawned from PR #4299 multi-model review. Per the tracking issue #4296 this is a **required gate before flipping Enforce mode** on gateways.

PR #4299 adds a synchronous end-to-end regression test (\`contract::governance::tests::governance_to_ban_list_end_to_end\`) that drives a contract through evict → recover → re-feed → BanTriggered via \`GovernanceManager\`, runs decisions through \`Ring::apply_ban_decisions\`, and asserts the contract lands on a fresh \`ContractBanList\`. That covers the wiring at the unit level.

What is missing is a **simulation-network test** that:
1. Spins up a small \`run_simulation_direct\` network (~5-10 nodes).
2. Injects the May 21 incident pattern (one abuser contract that triggers repeated evictions via the rate-limit → MAD → eviction → ban chain).
3. Asserts the full chain fires on each node observing the abuse:
   - Rate limit rejects excess UPDATE traffic for the abuser.
   - MAD detector flags the abuser in the cost distribution.
   - State machine transitions Normal → Borderline → WouldEvict → Evicted.
   - Re-eviction within ban_window triggers Banned.
   - ContractBanList drops subsequent wire traffic for the abuser.
4. Asserts non-abuser contracts in the simulation continue to function (no collateral damage).

## Why this matters

The unit and pin tests in #4299 lock the wiring in place, but they cannot detect emergent issues from the chain running against real network traffic (timing assumptions, message ordering, multi-node consensus on who's the abuser). The deferred sim test is the only way to validate \"Enforce mode actually works as designed in a real network\" before flipping the gateway config.

## Spawned from

- Testing review (Critical #2 — Missing regression test for the May 21 incident pattern)
- Big-picture review (end-to-end simulation test is explicitly listed as a gate before Enforce flip in #4296)

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-06-02T13:54:38.548418603+00:00
Issue hash: `80b1910e`

## Value Assessment

[...]

Issue #4296, opened 2026-05-27
tracking: contract + peer resource-usage governance


sanity:
## Goal

Detect and remove buggy or malicious **contracts AND peers** based on observed resource usage. Triggered by the May 25 2026 incident where contract `4PjqN5KUCidW8vJvw5fhy5fe5maxXKNrWSyK33QjjVq` exhibited a runaway UPDATE pattern and there was no automatic mechanism to stop it.

Design doc: [`docs/design/contract-hardening.md`](https://github.com/freenet/freenet-core/blob/main/docs/design/contract-hardening.md)

Two halves of the system:
- **Contract side** — detection via per-contract cost/benefit MAD scoring + state machine + receive-boundary rejection of banned contracts.
- **Peer side** — same shape, peer-attributed. The shared `crate::governance` primitive (MAD detector) serves both call sites.

## Contract side (detection → flag → remove)

- [x] **Phase 0**: WASM fuel coverage audit
- [x] **Phase 1**: LRU recently-abandoned bucket — #4260
- [x] **Phase 3**: Per-contract resource attribution (`AttributionSource::Contract`, new `ResourceType` variants, `Ring::commit_state_write` helper) — #4260
- [x] **Phase 4**: Shared governance module — MAD-based outlier detection in log space with 1.4826 σ-scaling, capacity ceiling clamp, NaN/inf guards — #4260
- [x] **Phase 4.5 slice 1**: GovernanceManager state machine (Normal → Borderline → WouldEvict → Evicted → Banned) + Contract Governance dashboard card + ring inner-ring with curved chord paths — #4270
- [x] **Phase 4.5 slice 2 (UI polish)** (#4298): mode pill in empty state, 5-tile skeleton with em-dashes, "observed N of 30" progress message replacing the current "Not enough contracts observed yet" copy, inner ring shows all hosted contracts (not just flagged). See `~/code/tmp/governance-dashboard-feedback-2026-05-27.md` for full feedback.
- [ ] **Phase 4.5 slice 3 (drill-down)**: per-contract detail panel — cost trajectory chart + decision history timeline + cost-ratio distribution histogram with the contract's position marked.
[...]

PR #4309, opened 2026-05-28
fix(governance): measure benefit as live beneficiary stock, not decaying subscribe-events (#4296)

sanity:
## Problem

The cost/benefit governance detector (DryRun on production) recommended **eviction of the official River room** `7Z8Nf1Fx5WsGRfc1AzH23bQBXDhmQFt8FgAKGw9poWXN` — the network's most popular contract — on two long-running gateways (nova, vega). Investigation (`actionable=false`, so no harm: it's DryRun and there is no production path to Enforce) traced it to the **benefit measure**, not the cost/benefit concept:

`benefit_score` was a *decaying accumulator of NEW subscribe events* — `ingest_demand` fired only on `AddSubscriberOutcome::NewAdd`, then decayed with cost's half-life, downstream weighted 0.1. A mature popular contract's subscribers all joined long ago, so its benefit decays toward zero while cost (full ~150 KB state rewrite per message, ~650/hr) flows continuously. Result: popularity *anti-correlates* with measured benefit. A stock-vs-flow mismatch that systematically false-flags the most valuable contracts.

## Approach

Measure benefit as a **live snapshot of current beneficiaries**, read fresh each reaper tick, instead of a decaying event accumulator. The ratio becomes **cost-per-current-beneficiary**.

- `benefit_score` is now overwritten each tick with `LOCAL_DEMAND_WEIGHT·local_client_count + FORWARDED_DEMAND_WEIGHT·downstream_subscriber_count` (weights unchanged: 1.0 / 0.1, for sybil-resistance), read live from `HostingManager`. Never accumulated, never decayed.
- `cost_used` unchanged (still accumulated + decayed).
- `log_ratio(benefit_floor)`: `None` only when `cost_used ≈ 0` (no activity); otherwise `cost_used / benefit_score.max(benefit_floor)`. A past-ramp-up contract with cost > 0 and **zero** live beneficiaries now yields a high, flaggable ratio — that is the genuine abuser signature, no longer excused. `benefit_floor` is a new `GovernanceConfig` field (default 0.1).
- Retires the `ingest_demand` push model and its two call sites; `governance_tick` builds the benefit map from new `HostingManager::{local_client_count, downstream_subscriber_count}` accessors (the latter counts only lease-valid downstream subscribers across the `ContractKey`-keyed map).

This keeps the always-on defenses (#4279 non-idempotency gate, #4285 rate limiter) as the primary protections regardless of governance mode.

## Testing

- Migrated all governance unit tests to the snapshot + benefits-map model, preserving every test's abuser/honest intent and assertions (no test deleted, no assertion weakened). Rewrote the three tests whose old behaviour was removed (benefit decay; `log_ratio` None-on-zero-benefit) to the new semantics, and added `zero_beneficiary_contract_with_cost_is_flagged`, `high_cost_high_benefit_contract_is_not_flagged`, `benefit_snapshot_overwrites_not_accumulates`, `log_ratio_high_when_cost_but_zero_benefit`.
- **Regression test for this exact incident:** `sim_e2e_tests::popular_contract_with_subscribers_not_evicted` — a high-cost contract *with many live downstream subscribers* is NOT evicted, while a same-cost contract with no beneficiaries IS banned. Plus the existing end-to-end ban-chain gate (#4301).
- 59/59 governance tests pass; full governance+hosting suite 201/201 locally; `cargo clippy --features "simulation_tests,testing" --lib --tests -- -D warnings` clean; `cargo fmt` clean. Sim tests stable across 3+ runs.

## Safety

Governance remains **DryRun by default with no production path to Enforce**. This PR does not flip Enforce. Enforce stays gated on a DryRun observation period in the wild confirming the official room (and other popular contracts) now score safely.

Refs #4296, #4301.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/governance.rs
+++ b/crates/core/src/contract/governance.rs
@@ -147,18 +147,37 @@ pub(crate) const MAX_TRANSITIONS_PER_CONTRACT: usize = 32;
 /// reads from this; the reaper compares `cost_used / benefit_score`
 /// against the network's MAD-derived threshold to drive transitions.
 ///
-/// Both `cost_used` and `benefit_score` decay over a rolling window
-/// (handled by [`ContractScore::decay`] called from the reaper tick).
-/// Without decay, a once-flagged contract would stay flagged forever
-/// even if its activity calmed down.
+/// `cost_used` decays over a rolling window (handled by
+/// [`ContractScore::decay`] called from the reaper tick) so a contract
+/// that goes quiet sheds its accumulated cost. `benefit_score`, by
+/// contrast, is NOT decayed and NOT accumulated: it is a LIVE SNAPSHOT
+/// of the current beneficiary population, overwritten fresh every tick
+/// from the hosting manager's standing subscription counts. See the
+/// `benefit_score` field doc and [`ContractScore::log_ratio`] for the
+/// rationale (a mature popular contract whose subscribers all joined
+/// long ago must keep a high benefit, not decay to zero).
 #[derive(Clone, Debug)]
 pub(crate) struct ContractScore {
     /// Sum of weighted resource samples attributed to this contract.
     /// Sourced from `Meter` entries keyed on
     /// `AttributionSource::Contract(_)`.
     pub cost_used: f64,
-    /// Sum of weighted demand events. Local subscriptions weigh more
-    /// than forwarded (Sybil resistance — see design doc).
+    /// LIVE SNAPSHOT of the current weighted beneficiary count.
+    ///
+    /// Set fresh each reaper tick from the hosting manager's standing
+    /// counts: `LOCAL_DEMAND_WEIGHT × active local clients +
+    /// FORWARDED_DEMAND_WEIGHT × active downstream subscribers`. It is
+    /// NOT accumulated across events and NOT decayed — it always
+    /// reflects "how many beneficiaries does this contract have RIGHT
+    /// NOW". The ratio `cost_used / benefit_score` is therefore
+    /// cost-per-current-beneficiary.
+    ///
+    /// Why a snapshot and not a decaying accumulator: the old model
+    /// fired on each NEW subscribe event and decayed with cost. A
+    /// mature popular contract's subscribers all joined long ago, so
+    /// its benefit decayed to ~0 while cost flowed continuously — a
+    /// false-positive eviction of the most popular contract on the
+    /// network. The live snapshot has no such ramp-down.
     pub benefit_score: f64,
     /// Current state.
     pub state: GovernanceState,
@@ -199,15 +218,39 @@ impl ContractScore {
         }
     }
 
-    /// Compute the log10(cost/benefit) ratio used by the MAD detector.
-    /// Returns `None` if benefit_score is too small to produce a
-    /// stable ratio (avoids division-by-near-zero pulling the
-    /// distribution toward infinity for new contracts).
-    pub(crate) fn log_ratio(&self) -> Option<f64> {
-        if self.benefit_score <= f64::EPSILON {
+    /// Compute the log10(cost / live-benefit) ratio used by the MAD
+    /// detector, where benefit is the current beneficiary snapshot
+    /// (NOT a decaying accumulator).
+    ///
+    /// Semantics:
+    ///
+    /// * If `cost_used <= f64::EPSILON` → `None`. No cost means no
+    ///   activity to judge; the contract isn't consuming resources, so
+    ///   there is nothing for governance to flag regardless of how many
+    ///   (or few) beneficiaries it has.
+    /// * Otherwise the effective benefit is
+    ///   `self.benefit_score.max(benefit_floor)` and the ratio is
+    ///   `(cost_used / benefit_eff).log10()`.
+    ///
+    /// The floor matters: a past-ramp-up contract with `cost > 0` and
+    /// ZERO live beneficiaries is the abuser signature — it burns
+    /// resources while nobody benefits. The floor keeps the division
+    /// finite and yields a HIGH (flaggable) ratio for that case. This
+    /// is deliberate: unlike the old model, zero benefit does NOT
+    /// return `None` here. Returning `None` would hide the very
+    /// contracts governance exists to catch.
+    pub(crate) fn log_ratio(&self, benefit_floor: f64) -> Option<f64> {
+        if self.cost_used <= f64::EPSILON {
             return None;
         }
-        let ratio = self.cost_used / self.benefit_score;
+        let benefit_eff = self.benefit_score.max(benefit_floor);
+        // `benefit_floor` is configured > 0 (default 0.1), so this is
+        // never a divide-by-zero. Guard defensively anyway in case a
+        // test passes a degenerate floor.
+        if benefit_eff <= 0.0 {
+            return None;
+        }
+        let ratio = self.cost_used / benefit_eff;
         if ratio <= 0.0 {
             return None;
         }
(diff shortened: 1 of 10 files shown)


sanity (commented):
## Comprehensive PR Review: #4309

### Summary
- **PR Title:** fix(governance): measure benefit as live beneficiary stock, not decaying subscribe-events (#4296)
- **Type:** fix (governance scoring) + test infra (#4301 sim gate)
- **CI Status:** mostly green at review time; **Unit & Integration was still in progress — must be green on current HEAD before merge** (no-merge-with-red-CI).
- **Linked Issues:** #4296 (governance tracker), #4301 (sim gate)
- **Review tier:** Full (high-risk surface: governance scoring / state authorization)
- **Reviewers run:** code-first, skeptical, testing, big-picture (all `freenet:*`) + Codex (external).

The redesign is sound and addresses the real root cause (stock-vs-flow): benefit is now a live snapshot of current beneficiaries, cost-only decay, `log_ratio` flags zero-beneficiary-with-cost instead of excusing it. Sybil-resistance holds (local-client signal requires a WebSocket client on the victim's own node; renewals can't pad a live count). No high-severity bugs; governance remains DryRun-only with no production Enforce path, so nothing here is an active-incident risk. Findings below are correctness/quality/perf and one important design residual.

---

### Recommendations

#### Must Fix (before merge)
1. **Benefit floor masks the smallest real beneficiary (Codex P2).** `benefit_floor` default `0.1` equals `FORWARDED_DEMAND_WEIGHT` (`0.1`), so a contract with exactly one live downstream subscriber (benefit 0.1) scores identically to a zero-beneficiary contract. Set the floor strictly below the smallest real beneficiary weight (e.g. `0.05`) so one real subscriber always scores better than none. `governance.rs` (`GovernanceConfig::default`).
2. **Missing unit tests for the highest-risk new code.** `HostingManager::downstream_subscriber_count` (lease-expiry filtering, `ContractKey`-vs-`InstanceId` matching, multi-key sum) and `local_client_count` have no direct unit tests. Add: active leases counted, stale-but-unswept NOT counted, exact `SUBSCRIPTION_LEASE_DURATION` boundary, sum across keys sharing an instance id. (`fix:`-class behavior; needs a `TimeSource`-backed `HostingManager`.)
3. **CI green on current HEAD** (Unit & Integration) before merge.

#### Should Fix
4. **`governance_tick` clones every `ContractScore` (incl. `history` Vec) just to harvest keys (M2), then does an O(N×M) full scan of `downstream_subscribers` per contract (M3).** On a busy gateway (thousands of contracts) this is a quadratic clone-the-world sweep every 60s. Build the benefit map in a single pass over `downstream_subscribers` + `client_subscriptions` grouped by instance id, and use a keys-only / borrow accessor instead of `iter_scores()`'s deep clone. `ring.rs::governance_tick`, `governance.rs::iter_scores`.
5. **Stale design doc.** `docs/design/contract-hardening.md` still describes the OLD model ("Benefit (count, decayed)", demand-event accumulation, `subscriber_minutes` weighting, "subscribers add to benefit_score"). AGENTS.md requires fixing stale design docs in the same PR. Update to the non-decayed live-beneficiary-snapshot model + document `benefit_floor`.
6. **Stale comment.** `compressed_config()` doc (`governance.rs` ~2056) still says "decay is symmetric on cost and benefit" — that's the removed model. Re-justify `decay_half_life` in terms of the snapshot model.

#### Consider
7. **M1 (design residual — see note below):** document that cost/benefit locality can diverge per-node, and resolve before any Enforce flip. Add a tracking item under #4296.
8. **Dashboard copy:** `home_page.rs` empty-state (~856-869) says contracts "accumulate cost/benefit signals" — now only cost accumulates. Consider a Benefit-column tooltip clarifying it's a live count.
9. **L1:** add a comment noting the MAD bottom tail is intentionally untrimmed and why it's safe (sub-EPSILON cost unreachable in practice).
10. **L2:** add a test for the transient all-leases-stale-then-recover path (one flagged tick → Borderline, not Banned).
11. **Enforce-promotion criteria:** record in the design doc what the DryRun observation must confirm before flipping (durable, not just the PR body).

---

### M1 — the false-positive fix is per-node and not fully closed (important)
Cost is ingested at the state-write chokepoint (`commit_state_write → report_contract_resource_usage → ingest_cost`) on whichever node applies the UPDATE. Benefit is counted from `local_client_count + downstream_subscriber_count` **on that same node**. For a node that bears a popular contract's heavy UPDATE cost but carries few/none of its subscriptions (a relay/propagation node, or the production gateway nova in the original report: high cost, ~7 local clients, ~0 downstream-subscribe events in-window), benefit can still be low → `cost / small` → flaggable. This is the same class as the bug being fixed, narrowed to per-node cost/benefit locality. The fix is a strict improvement but does **not guarantee** a popular contract is unflagged on every node. The regression test only asserts unbanned on the contract's max-subscriber host, not everywhere. **Not an active risk (DryRun-only), but a documented residual and a hard pre-Enforce gate:** the DryRun-in-the-wild observation must confirm the official room scores safely across all node roles (gateway, relay, leaf), not just its subscription hub. A full fix likely needs cross-node/subscription-tree benefit aggregation, which is a larger design question.

---

### Verdict
**HEAD SHA reviewed:** b2e60f1ef0a5dcd96e778c9b3126fc24a3a91009

The design is correct and the implementation faithful; no blocking bugs given DryRun-only. Address the Must-Fix items (floor calibration, the two missing hosting unit tests, green CI) and the Should-Fix items (perf refactor, stale doc/comment), then a re-review on the updated HEAD. M1 is the key item gating any future Enforce flip and should be documented now.

[AI-assisted - Claude]

sanity (commented):
## Re-Review (Round 2): PR #4309 — clean

Re-review on HEAD `5abe4414` after fixes for Round 1's findings. Tier: Full (governance scoring). This pass: adversarial skeptical reviewer on the delta + external Codex, plus production-build / `-D warnings` verification across feature configs.

### Round 1 findings — all resolved
- **Codex P2 (benefit floor masks 1 subscriber):** floor 0.1 → **0.05** (strictly below `FORWARDED_DEMAND_WEIGHT`); guard test `one_downstream_scores_strictly_below_zero_beneficiaries`. ✓
- **Missing hosting unit tests:** added `downstream_subscriber_count_{counts_active_leases, excludes_stale_unswept_lease, boundary_is_exclusive, sums_across_keys_sharing_instance_id}` + `local_client_count_*` (deterministic via `start_paused`). ✓
- **M2/M3 (clone-the-world + O(N×M) per tick):** `governance_tick` now uses `tracked_ids()` (keys only) + `HostingManager::beneficiary_counts()` (single pass), then `retain` to tracked. **Verified behavior-preserving** — `tick` does `benefits.get(key).unwrap_or(0.0)`, so omitting a zero-beneficiary tracked contract is identical to the old explicit-0.0; lease filter / summing identical; pinned by `beneficiary_counts_matches_per_contract_computation`. ✓
- **Stale design doc / comment / dashboard copy:** `contract-hardening.md`, `compressed_config` doc, and home-page empty-state updated to the live-snapshot model. ✓
- **L1/L2:** untrimmed-bottom-tail comment + `transient_zero_benefit_flags_but_cannot_ban_and_recovers` (verified ban genuinely unreachable in DryRun, not asserted-away). ✓
- **M1 (per-node cost/benefit locality):** documented as a pre-Enforce gate in the design doc; **no scoring code changed** for it. ✓

### Found & fixed during re-review (not in Round 1)
- **Production build was broken (E0609):** `NodeConfig::governance_config_override` was `#[cfg(any(test, feature="testing"))]` but `node::testing_impl` compiles unconditionally. Made the field unconditional (`#[serde(skip)]` Option, `None` in prod → no wire/behavior impact). `shared_rings` marked `#[allow(dead_code)]` (write-only in prod). Verified `cargo clippy -p freenet -- -D warnings` (production, no `testing`) clean — the gate that would have caught it; Round 1 checks only ran `--features testing`.
- **Cosmetic:** gated two test-only `Ring` accessors to `simulation_tests` to match their callers (silences a `dead_code` warning in `--features testing` test builds).

### External (Codex) — clean
"Compiles under the default feature set; live-benefit snapshot wiring consistent with the updated semantics; no discrete actionable regressions introduced by the patch."

### Skeptical — clean
No High/Medium findings. Refactor verified value-identical; lease/summing/floor correct; production-build fix sound; new tests not masking; M1 doc-only. No new race (the three map reads in `governance_tick` run synchronously with no intervening await, same skew tolerance as before via `unwrap_or(0.0)`).

### Verdict
**State:** Ready to Merge (pending CI green on HEAD `5abe4414`).
**HEAD reviewed:** 5abe4414af91b2a53a5dd6e1544117d94a34f1de
Governance remains DryRun-only with no production Enforce path. M1 (cross-node benefit aggregation) is the documented gate before any future Enforce flip; the DryRun-in-the-wild observation must confirm popular contracts score safely across all node roles first.

[AI-assisted - Claude]

Issue #4294, opened 2026-05-27
transport: make reference-ping target + enable configurable (#4074 Phase 1.5 follow-up)


sanity:
Follow-up for the Phase 1.5 reference-ping shadow telemetry added in #4292.

## Current state (after #4292 merges)

- Reference target is hardcoded to `1.1.1.1:53` via `DEFAULT_REFERENCE_TARGET` in `crates/core/src/transport/reference_ping.rs`.
- The probe is gated on \`config.telemetry.enabled && !config.telemetry.is_test_environment\`. Operators can disable it indirectly by turning off telemetry entirely, but not independently.
- Bind failure does not crash the node (the spawn loop hands a pending future to BackgroundTaskMonitor) but does silently disable the probe.

## What this issue tracks

1. **Configurable target** — once we have ~2-4 weeks of Phase 1.5 data, we may want to point at the local gateway, a different public resolver, or run multiple probe targets in parallel. Thread a config option (e.g. `[transport] reference-ping-target = \"1.1.1.1:53\"`) through `TransportConfig`. \"None\" disables.

2. **Independent enable flag** — privacy-conscious operators who want telemetry on but reference-ping off currently have no knob. Add `[transport] reference-ping-enabled = true` defaulting to follow `telemetry.enabled`, so it can be explicitly turned off without disabling all other telemetry.

3. **QNAME randomisation** — the current QNAME is the literal string \`\"freenet.org\"\`. A network-level observer (ISP, censorship middlebox) can use the constant-cadence DNS query for that specific name as a host-level beacon for traffic-analysis-based blocking. Consider per-node random subdomain (e.g. \`<random>.probe.freenet.org\` if we control DNS for that name) so the QR pattern isn't a freenet-identifying beacon.

Don't tackle these until Phase 1.5 data informs Phase 2 design — premature config knobs would lock in choices we don't have evidence for yet.

Refs #4074

[AI-assisted - Claude]

sanity:
Adding a third item to track from the #4292 review:

[...]

Issue #4074, opened 2026-05-10
RFC: Outer-loop rate controller using cross-connection RTT correlation


sanity:
> **Status (2026-06-10): this body is the living design-of-record.** It supersedes the original RFC (preserved, collapsed, at the bottom). Decisions below are locked; the immediate next step is observation-only telemetry (Phase 1.6). The reasoning trail lives in the analysis comments linked inline — if you change the design, update *this body*, don't just add a comment.

## Goal

Maximise Freenet's bandwidth usage while **yielding to the operator's own traffic**: use idle uplink aggressively, back off when the operator needs the connection (a video call, an upload). This matches LEDBAT's original objective. The novelty is the *signal*, not the goal.

## Decisions locked

1. **Yield to a network-health floor — Option B, not headroom-only.** The original hard floor (`10 Mbps × N`) meant the controller could only ever raise `R`, never drop below today's rate, so it could absorb idle headroom but could never actually yield. That fails the goal. The floor must sit **below** today's rate at a network-health minimum. Rationale: headroom-only is a no-op for residential operators (already oversubscribed at `10 Mbps × N`, the root of #2936 ISP-shaping and #3338 fan-out saturation) and only benefits fat-link nodes that have no contending operator traffic to yield to. See [floor-validation analysis](https://github.com/freenet/freenet-core/issues/4074#issuecomment-4673972429).
2. **Reference-ping is the *primary* contention signal, not "optional auxiliary".** [Phase 1 telemetry (39 h, ~368k samples)](https://github.com/freenet/freenet-core/issues/4074#issuecomment-4556599891) showed overlay-RTT inflation has a **~55 ms structural baseline** (median 55 ms, p90 91 ms; 90.5% of samples ≥ 30 ms). The original absolute 30 ms trigger would fire essentially always and never raise the rate. The detector must be **derivative-based and/or reference-ping-led**, not an absolute overlay-RTT threshold. Also: 26% of samples have N<3 active peers, so "insufficient signal, hold" must be a first-class mode.
3. **Aggregate `R` is the controlled variable; per-peer rate = `R / active_connections`.** Maps onto `GlobalBandwidthManager::total_limit`, composing with the existing `min(cc_rate, global_rate)` cap (open question 4 below: the existing per-connection floor must not break the aggregate cap at high N).

## Why prior attempts failed, and why this differs

[...]

PR #4315, opened 2026-05-31
feat(replay-harness): offline harness for #4074 Phase 2 controller design

### LINKED ISSUE #4314: tooling: shadow-RTT replay harness for #4074 Phase 2 controller design
## Problem

Three prior congestion-control attempts (LEDBAT++, BBRv3, custom adaptive — see #4074) all failed in production because there was no offline test loop. The failure modes (LEDBAT's 50–76 min completion times from single-packet-loss → multiplicative decrease, BBR's stale min_RTT across reroutes) were only observed in the wild and only after the algorithm was already deployed.

The Phase 1.5 telemetry just landed (#4292) gives us per-node disaggregated overlay RTT + a parallel reference-path signal. Before Phase 2 ships any controller logic, we need a way to evaluate candidate controllers against (a) synthetic scenarios that pin the prior failure modes and (b) the real OTLP data once it accumulates.

## Proposal

A new workspace member `crates/replay-harness/` (library + binary). No changes to production code.

### Architecture

- `Controller` trait — takes a `ControllerView<'_>` snapshot per tick, returns a `RateDecision`.
- `EventStream` trait — sources RTT samples (synthetic generators or OTLP file replay).
- `Replayer` — drives the simulation: feeds events into per-peer `RollingRttStats`, ticks the controller, records decisions.
- `Scenario` — bundles a synthetic event stream + assertions about what a sane controller should do; runs as `#[test]`.

### v1 scenarios (each pins a known failure mode)

| Scenario | Tests |
|---|---|
| `idle_steady_state` | Flat 80ms RTT across N peers — should NOT trigger (the 90.5%-above-30ms problem we measured) |
| `single_packet_loss` | One RTT spikes 10× then returns — pins LEDBAT++ death spiral on N=1 |
| `slow_routing_drift` | Baseline drifts up 5 ms/min over 30 min — pins "min over 5 min" tracking real path changes |
| `single_peer_outlier` | One peer spikes to 500ms, others stable — median-across-peers should reject this |
| `correlated_inflation` | All peers' RTT rise together — RFC's intended detection case, SHOULD trigger |
| `small_n` | N=1 or N=2 sustained — should NOT trigger (sample size too small) |
| `churning_peers` | Peers join/leave every 30s — connection churn must not produce false triggers |
| `cold_start` | New peer, no baseline — should hold, not panic |
| `reference_diverges_from_overlay` | Overlay inflates, reference flat — overlay queueing, NOT contention; should NOT trigger |
| `reference_tracks_overlay` | Both inflate together — uplink contention, SHOULD trigger |

### v1 controllers

- `FixedRate` — baseline, never changes; sanity-checks the harness itself.
- `RfcDraft` — the algorithm sketched in #4074 (30 ms / 5 s / 0.7× / 1.05×). Used to demonstrate the threshold-below-noise-floor problem against real data.

### Workflow for Phase 2

1. Author proposes a controller in `crates/replay-harness/src/controllers/`
2. Adds scenarios that pin the failure modes the controller targets
3. Runs `cargo test -p replay-harness` — must pass all sane-controller scenarios
4. Runs `replay-harness otlp <recent-data> --controller <name>` — produces decision report
5. PR review reads the decision report, not just the algorithm code
6. Only when the controller passes synthetic + looks reasonable on OTLP replay does it get an opt-in flag in production

## Why this is the right next step

- It's the only way to catch class-of-failure issues (the kind that killed LEDBAT/BBR) before production
- Synthetic scenarios are committable as regression tests — once a failure mode is in the harness, every future Phase 2 candidate must clear it
- OTLP replay against the data Phase 1.5 is now collecting lets us evaluate Phase 2 algorithms offline before any opt-in deployment
- No production code change — strictly additive tooling

## Scope discipline

- v1 binary commands: `replay-harness synthetic <scenario>`, `replay-harness otlp <file>`, `replay-harness compare <a> <b> <events>`
- Out of v1: GUI/plots (CSV/JSON output is enough), live mode (read collector in real-time), anything that touches transport at runtime

Design notes already drafted in [a hand-off file](https://github.com/freenet/freenet-core/issues/4074); will translate to docs/ in the PR.

Refs #4074, #4294

sanity:
## Problem

Three prior congestion-control attempts (LEDBAT++, BBRv3, custom adaptive — see #4074) all failed in production because there was no offline test loop. LEDBAT's death spiral on a single packet loss only showed up after deploy. BBR's stale `min_RTT` across reroutes was caught by manual debugging weeks after the algorithm went live. The Phase 1.5 telemetry (#4292) just landed the per-node tagged data we need; this PR is the next piece — before any Phase 2 controller ships, it must pass scripted scenarios that pin known failure modes.

## Solution

New workspace member `crates/replay-harness/` (library + binary). Strictly offline tooling; no production code path reaches this crate.

### What v1 ships

- `Controller` trait + `ControllerView` snapshot + `RateDecision`.
- **`FixedRate`** — the production default, never changes rate. Sanity-checks the harness itself: any scenario where this controller fires is a bug in the harness, not the controller.
- **`RfcDraft`** — the algorithm sketched in #4074 (30 ms / 5 s / 0.7× downstep with the N≥3 guard the rolling-stats module's own rustdoc demands). Implements the downstep branch; the "idle headroom" upward branch needs an outbound-traffic model that v1 does not provide.
- **Four scenarios** (out of the 10 in the design — the other 6 are additive and tracked as follow-up):

  | Scenario | Pins |
  |---|---|
  | `idle_steady_state` | The ~80 ms ambient overlay noise we measured. Sane controllers MUST NOT fire. |
  | `correlated_inflation` | Every peer's RTT rises together — the actual contention case. MUST fire. |
  | `single_peer_outlier` | One peer at 500 ms, four others at baseline. Median rejects it; MUST NOT fire. |
  | `small_n` | Only 2 peers (below rolling-stats N≥3 threshold). MUST NOT fire. |

- `Replayer` drives an `EventStream` through a `Controller` at 1 Hz, recording a per-tick decision trace.
- Binary `replay-harness` with `scenarios`, `controllers`, `synthetic`, and (stub) `otlp` subcommands.

### What the harness already demonstrates

`RfcDraft` against `idle_steady_state` fires **48 times across 300 s**, dropping the rate from 1.25 Mbps to effectively 0. That's the demonstration of the threshold-below-noise-floor bug the Phase 1 telemetry already exposed. `RfcDraft` against `correlated_inflation` fires 9 times correctly (rate drops 25×). Against `single_peer_outlier` and `small_n` it correctly holds. Both `RfcDraft` outcomes are pinned in `tests/scenarios_pin.rs` so any Phase 2 candidate that "fixes" RfcDraft by accident has to update the pin.

### Why not full OTLP replay yet

The current `shadow_rtt_aggregate` event emits the pre-computed aggregate, not the per-peer RTT samples the rolling stats need. Two follow-up options — emit per-peer samples via OTLP, OR add an aggregates-replay mode — both with caveats, both deferred. The synthetic-scenario harness is the higher-value testing layer anyway: it pins *specific* failure modes rather than measuring against a single instance of "what happened in production last week."

## Testing

- 5 rolling-stats unit tests (port of the production `RollingRttStats` test suite).
- 3 pin tests in `tests/scenarios_pin.rs`:
  - `fixed_rate_never_fires_on_any_scenario` — universal sanity check.
  - `rfc_draft_outcomes_match_documented_expectations` — pin each (controller, scenario) outcome explicitly.
  - `scenarios_declare_concrete_expectations` — guards against `Expectation::Informational` slipping in.
- `cargo fmt`, `cargo clippy -p replay-harness --lib --tests --bins -- -D warnings`, and full `cargo check --workspace` all clean.

## Out of scope (deliberate, tracked as follow-up)

- Remaining 6 scenarios from the design (`slow_routing_drift`, `churning_peers`, `cold_start`, `reference_diverges_from_overlay`, `reference_tracks_overlay`, plus a stripped-down LEDBAT++ to reproduce the death spiral). Adding them is a small file each; deliberate scope discipline for v1.
- OTLP file replay (see "Why not" above).
- GUI / plots — decision trace is CSV/JSON-friendly, plot externally.
- Any change to production transport behaviour.

Closes #4314
Refs #4074, #4294

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5514,6 +5514,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "replay-harness"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "serde",
+ "serde_json",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "reqwest"
 version = "0.12.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/crates/replay-harness/Cargo.toml
+++ b/crates/replay-harness/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "replay-harness"
+version = "0.1.0"
+edition = "2024"
+license = "MIT OR Apache-2.0"
+publish = false
+description = """
+Offline replay harness for evaluating shadow-RTT controllers against
+synthetic scenarios and OTLP telemetry. Phase 2 of #4074 design tool.
+Not a production component.
+"""
+
+[lib]
+path = "src/lib.rs"
+
+[[bin]]
+name = "replay-harness"
+path = "src/main.rs"
+
+[dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+
+[dev-dependencies]
--- a/crates/replay-harness/README.md
+++ b/crates/replay-harness/README.md
@@ -1,0 +1,105 @@
+# replay-harness
+
+Offline test harness for evaluating Phase 2 shadow-RTT controllers (issue
+[#4074][issue]) against synthetic scenarios and OTLP telemetry, before
+shipping any controller to production.
+
+Not a production component. No code path in `freenet` reaches this crate.
+
+## Why
+
+The previous three congestion-control attempts (LEDBAT++, BBRv3, custom
+adaptive) all failed in production because there was no offline test loop.
+LEDBAT's death spiral on a single packet loss only showed up after deploy.
+BBR's stale `min_RTT` across reroutes was caught by manual debugging weeks
+after the algorithm went live.
+
+Phase 1.5 ([#4292][p15]) just landed the per-node tagged telemetry we need.
+The replay harness is the next piece: before any Phase 2 controller ships,
+it must pass a battery of scripted scenarios that pin known failure modes,
+then be evaluated against real OTLP data without touching production.
+
+[issue]: https://github.com/freenet/freenet-core/issues/4074
+[p15]: https://github.com/freenet/freenet-core/pull/4292
+
+## Quick start
+
+```bash
+# List the scenarios and controllers shipped with v1.
+cargo run -p replay-harness -- scenarios
+cargo run -p replay-harness -- controllers
+
+# Run one scenario against the default controller (rfc_draft).
+cargo run -p replay-harness -- synthetic correlated_inflation
+
+# Run every scenario against rfc_draft (the algorithm sketched in #4074).
+cargo run -p replay-harness -- synthetic all
+
+# Same against the no-op baseline.
+cargo run -p replay-harness -- synthetic all --controller fixed_rate
+
+# Run the assertion suite (every scenario × every controller, with each
+# (controller, scenario) outcome pinned).
+cargo test -p replay-harness
+```
+
+## What's pinned in v1
+
+Four scenarios, chosen to cover the most important pins for Phase 2 design:
+
+| Scenario | What it pins | Sane behaviour |
+|---|---|---|
+| `idle_steady_state` | The ~55 ms ambient overlay noise we measured in production | MUST NOT fire |
+| `correlated_inflation` | Every peer's RTT rises together — the actual contention case | MUST fire at least once |
+| `single_peer_outlier` | One peer at 500 ms, four others at baseline | MUST NOT fire (cross-peer median rejects) |
+| `small_n` | Only 2 peers — below the `rolling_rtt_stats.rs` N≥3 trustworthy threshold | MUST NOT fire (N≥3 guard) |
+
+The `RfcDraft` reference controller intentionally **fails** the
+`idle_steady_state` pin (48 fires across the 300 s scenario, rate drops
+to ~0). That failure is the demonstration of the noise-floor problem the
+Phase 1 telemetry already showed: the 30 ms inflation threshold sits well
+below the ambient overlay queueing baseline, so the controller fires on
+healthy ambient noise. Any Phase 2 candidate MUST NOT repeat that failure.
+
+`FixedRate` (the production default) passes the universal sanity check
+of "never fires on any scenario."
+
+## Adding a controller
+
+1. Drop a new file in `src/controllers/`.
+2. Implement [`Controller`](src/controllers.rs).
+3. Re-export it from `src/controllers.rs`.
+4. Wire it into the binary's `match controller_name { … }` in `src/main.rs`.
+5. Run `cargo test -p replay-harness` — it auto-runs every scenario
+   against your controller; if your design philosophy diverges from the
+   default expectation for any scenario, extend `expected_fires` in
+   `tests/scenarios_pin.rs`.
+
+## Adding a scenario
+
+1. Drop a new file in `src/scenarios/`.
+2. Implement a `pub fn scenario() -> Scenario`.
+3. Add the module to `src/scenarios.rs` and append to `all_scenarios()`.
+4. `cargo test` — the new scenario runs against every controller
+   automatically.
+
+## Why not real OTLP replay yet
+
+The current OTLP `shadow_rtt_aggregate` event emits the pre-computed
+aggregate (`active_peers`, `peers_with_recent`, `median_inflation_us`) —
+not the underlying per-peer RTT samples that `RollingRttStats` needs to
+recompute the snapshot at each tick. Feeding OTLP into the harness
+requires either (a) ALSO emitting per-peer samples via OTLP, or (b) an
+aggregates-replay mode that fakes per-peer snapshots from the aggregate.
+Both have caveats; both are deferred to a follow-up.
+
+In the meantime, synthetic scenarios are the higher-value testing layer
+anyway — they pin *specific* failure modes rather than measuring against
+a single instance of "what happened in production last week."
+
+## Not in scope
+
+- Per-tick controller telemetry to the OTLP collector (offline tool only).
+- GUI / plots (decision-trace dump is CSV/JSON-friendly; plot externally).
+- Live mode (reading the collector in real-time).
+- Any change to production transport behaviour.
--- a/crates/replay-harness/README.md
+++ b/crates/replay-harness/README.md
@@ -55,11 +55,12 @@ Four scenarios, chosen to cover the most important pins for Phase 2 design:
 | `small_n` | Only 2 peers — below the `rolling_rtt_stats.rs` N≥3 trustworthy threshold | MUST NOT fire (N≥3 guard) |
 
 The `RfcDraft` reference controller intentionally **fails** the
-`idle_steady_state` pin (48 fires across the 300 s scenario, rate drops
-to ~0). That failure is the demonstration of the noise-floor problem the
-Phase 1 telemetry already showed: the 30 ms inflation threshold sits well
-below the ambient overlay queueing baseline, so the controller fires on
-healthy ambient noise. Any Phase 2 candidate MUST NOT repeat that failure.
+`idle_steady_state` pin (48 fires across the 300 s scenario, rate
+floors at 1 bps after enough successive 0.7× downsteps). That failure
+is the demonstration of the noise-floor problem the Phase 1 telemetry
+already showed: the 30 ms inflation threshold sits well below the
+ambient overlay queueing baseline, so the controller fires on healthy
+ambient noise. Any Phase 2 candidate MUST NOT repeat that failure.
 
 `FixedRate` (the production default) passes the universal sanity check
 of "never fires on any scenario."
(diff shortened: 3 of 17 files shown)


sanity (commented):
## Comprehensive PR Review: #4315

### Summary
- **PR Title:** feat(replay-harness): offline harness for #4074 Phase 2 controller design
- **Type:** feat (new tooling crate)
- **CI Status:** Pending (PR just opened)
- **HEAD SHA reviewed:** 3e9e106a
- **Linked Issues:** #4314 (approval), #4074 (parent RFC), #4294 (Phase 2 follow-ups)
- **Review tier:** Light (no high-risk surface — strictly offline tooling, no production code path), but non-trivial logic → spawned `freenet:skeptical-reviewer` + `freenet:big-picture-reviewer` + codex external pass

---

### Code-First Analysis
**Independent Understanding:** New workspace member `crates/replay-harness/` (lib + bin) — a single-threaded, deterministic simulator that drives a candidate `Controller` against synthetic RTT scenarios, recording a per-tick decision trace. Ports `RollingRttStats` from `crates/core` as an intentional, well-documented copy; supplies two reference controllers (`FixedRate` baseline + `RfcDraft` per the #4074 sketch); ships 4 of 10 designed scenarios (idle_steady_state, correlated_inflation, single_peer_outlier, small_n) wired into a pin-test matrix in `tests/scenarios_pin.rs`.

**Stated Intent:** Same — verified against the PR description and the approved design in #4314.

**Alignment:** Matches, with two design-list items intentionally deferred or unshipped:
- `otlp` subcommand: stubbed cleanly with a `bail!` citing #4314 (justified — the existing OTLP emits aggregates, not the per-peer samples the rolling stats need)
- `compare` subcommand: in the #4314 design list, not in the PR, **not mentioned as deferred** — see Should-Fix #1.

**Gaps:** None substantive. Production code untouched (1400 insertions / 0 deletions).

---

### Testing Assessment
**Coverage Level:** Adequate for the scope.

| Unit | ✅ | 5 rolling-stats unit tests, port of production `RollingRttStats` test suite |
| Integration | ✅ | 3 pin tests in `tests/scenarios_pin.rs` — universal sanity (FixedRate never fires), explicit (controller, scenario) outcome matrix for RfcDraft, expectation-shape guard |
| Simulation | N/A | Offline tooling; no network |
| E2E | N/A | Same |

**Regression Test:** N/A (new tooling, not a fix)

**Notable test design:** The `expected_fires(controller, scenario)` matrix in `tests/scenarios_pin.rs:30-50` is exactly the right shape — it pins each (controller, scenario) outcome explicitly, so any Phase 2 candidate that accidentally "fixes" RfcDraft has to consciously update the pin.

**Verified empirically:** Ran `cargo test -p replay-harness` and `cargo run -p replay-harness -- synthetic all`. The PR description's claim of "RfcDraft fires 48 times on idle_steady_state, 9 times on correlated_inflation" reproduces exactly (verified by tracing the 6-tick fire cadence post-`clear()`).

---

### Skeptical Findings
**Risk Level:** Low — offline tooling, no production reach.

| # | Concern | Severity | Location | Details |
|---|---------|----------|----------|---------|
| 1 | **`Scenario::run_for` is dead and contradicts its own rustdoc** (confirmed by both Codex and Skeptical reviewers — strongest finding) | Should | `replayer.rs:111`, `scenarios.rs:50-53` | `Replayer::run` derives `stop_at` only from `events.last().at()`, never reads `Scenario::run_for`. All four shipped scenarios happen to set `run_for = last_event + 1s` so it works coincidentally. Future scenarios with a quiet-tail observation window (e.g. "controller stays quiet for 60s after the burst ends") will terminate early and falsely pass `NeverFires` assertions. Fix: thread `run_for` into the replayer (most useful), OR remove the field (cheapest, but the documented intent is genuinely useful). |
| 2 | **CLI marker disagrees with pin tests** | Should | `main.rs:107-112` | The CLI's PASS/FAIL marker compares `decisions_set` to the *scenario's* default expectation, while the pin tests correctly use the per-(controller, scenario) override. Result: `synthetic all --controller fixed_rate` prints "FAIL correlated_inflation/fixed_rate" — misleading because FixedRate is *expected* not to fire on any scenario per the documented matrix. Lift the same `expected_fires(controller, scenario)` lookup from `tests/scenarios_pin.rs:30-50` into the CLI, or qualify the marker. |
| 3 | **Missing `compare` subcommand** (from #4314 design) | Should | `main.rs:22-46` | Design listed three v1 binary commands: `synthetic` (works), `otlp` (stubbed with clear deferral), `compare` (silently absent). Either add it (the `Replayer` already produces `DecisionLog` traces for two controllers — diffing is small) or add a one-line deferral note in the PR description / README. |
| 4 | `PeerLeave` then same-peer `RttSample` silently resurrects the peer with empty stats | Consider | `replayer.rs:170-184` | Uncovered (no scenario uses join/leave yet) but a footgun when someone adds a churn scenario. Document the leave-then-sample semantic, or warn/ignore the post-leave sample. |
| 5 | `with_tick_interval(Duration::ZERO)` is an infinite loop | Consider | `replayer.rs:113` | `now += ZERO` never grows; `now > stop_at` never true. Add `debug_assert!(interval > Duration::ZERO)` on the builder. |
| 6 | `tick_once` recomputes `shared_inflation` for the log without the N≥3 guard | Consider | `replayer.rs:212-222` vs `rfc_draft.rs:67-72` | The trace log can show a non-`None` `shared_inflation` for ticks where the controller correctly held due to N<3, making post-hoc analysis confusing. Field-level doc clarification or use the controller's view directly. |
| 7 | `idle_steady_state` rustdoc claims "~40 ms inflation from path queueing" but the construction is one t=0 fast sample anchoring baseline_min | Consider | `scenarios/idle_steady_state.rs:25-29` | The pin still pins what it pins (RfcDraft fires — the bug demo), but the physical interpretation is misleading. Tighten the doc, or scatter fast samples to make the inflation actually queueing-driven. |
| 8 | `correlated_inflation` first fire is at t=70 (10s recent window + 5s sustain = 15s lag from t=60 burst start) | Consider | `scenarios/correlated_inflation.rs:32-33` | Rustdoc says "well above the 30 ms inflation trigger" without mentioning the inherent 15s detection lag. A reader who picks `run_for=65s` would expect a fire and be surprised. |
| 9 | `correlated_inflation` pin test only checks `decisions_set > 0` | Consider | `tests/scenarios_pin.rs` | Pins "did it fire" but not "did it fire for the right reason at the right time" — a bug that fires at t=2 would pass. Strengthen: "first fire is at or after t=65". |
| 10 | README "rate drops to ~0" — actual minimum is 1 bps (integer floor on `(1*0.7).round()`) | Consider | `README.md:60` | Trivial nit; flag for accuracy. |

---

### Big Picture Assessment
**Goal Alignment:** Yes — the implementation matches the approved #4314 design with explicit, defensible scope shrinkage (4 of 10 scenarios, 1 missing subcommand). The 4 shipped scenarios are the *right* ones for v1 because they are the only ones `RfcDraft` can be meaningfully evaluated against. Several deferred scenarios (e.g. `single_packet_loss`) require either a different controller or different signals that haven't been wired.

**Anti-Patterns Detected:** None. No `#[ignore]`, no `todo!`, no `unimplemented!`, no `panic!`/`FIXME`/`XXX`. One justified `#[allow(clippy::too_many_arguments)]` on a private helper.

**Removed Code Concerns:** None — purely additive.

**Scope Assessment:** Focused, no scope creep. The rolling-stats copy rationale is documented twice (`lib.rs:11-13` and `rolling.rs:5-10`) and the copy is faithful to the production type at every behaviour-bearing constant and computation.

---

### Documentation
- Code docs: Complete — every public item has rustdoc; the rolling-stats copy explicitly explains why it's a copy.
- Architecture docs: Up-to-date *for what shipped*. `.claude/rules/transport.md` (Phase 2 row) could benefit from a one-line pointer to `crates/replay-harness` as the Phase 2 gating tool — discoverability, not correctness. (Consider, not blocking.)
- User docs: README is well-structured; one minor accuracy nit (#10 above).

---

### Recommendations

#### Must Fix (Blocking)
None. The harness is correct on the four shipped scenarios; the demonstration value is real and reproducible.

#### Should Fix (Important)
1. **`Scenario::run_for` dead field** (#1): Honor it in the `Replayer` so future scenarios with quiet-tail observation windows behave per their documented intent, OR remove the field. Recommendation: honor it (the documented intent is genuinely useful — assert "still no fire after N seconds of quiet" is a common future need).
2. **CLI PASS/FAIL marker** (#2): Use the per-(controller, scenario) `expected_fires` lookup the pin tests already use, so the binary's output is honest.
3. **Missing `compare` subcommand** (#3): Either implement (small follow-up since `DecisionLog` already exists for both runs) or add a one-line note to the PR description / README explaining the deferral alongside `otlp`.

#### Consider (Suggestions)
4. Document or fix `PeerLeave`→`RttSample` resurrection (#4).
5. `debug_assert!(interval > ZERO)` on `with_tick_interval` (#5).
6. Field-doc clarification on `DecisionLog::shared_inflation` re: N≥3 guard divergence (#6).
7. Tighten `idle_steady_state` and `correlated_inflation` rustdoc (#7, #8).
8. Strengthen `correlated_inflation` pin test (#9) — assert first-fire timing.
9. README accuracy: "rate drops to ~1 bps (integer floor)" (#10).
10. Optional `TickState` struct to retire the `#[allow(clippy::too_many_arguments)]`.
11. One-line pointer in `.claude/rules/transport.md` Phase 2 row to the harness for next-agent discoverability.

---

### Verdict

No Must-Fix items. The three Should-Fix items are localized and don't introduce new logic surfaces. A focused diff-of-the-diff pass on the fix commits is sufficient rather than a full re-review.

[AI-assisted - Claude]

PR #4316, opened 2026-05-31
feat(dashboard): surface UPDATE rate-limiter counters on status card

sanity:
## Problem

The per-`(sender, contract)` UPDATE rate limiter added in #4285 ("front-line defense against the May 21 incident pattern") silently drops relayed UPDATEs once a pair exceeds ~10/s, and drops new pairs when the tracking map hits `MAX_TRACKED_PAIRS`. The drop is `tracing::debug!`-only, so an operator has **no visible signal** that the limiter is active or that legitimate traffic might be getting dropped.

The accepted / rejected / capacity-rejected counters already exist on `UpdateRateLimiter` but were dead code (`#[cfg_attr(not(test), allow(dead_code))]`). #4285's own PR body reserved them "for the follow-up dashboard surface" — this is that follow-up.

## Approach

Surface the three existing counters on the node status dashboard, following the `RingStatsSnapshot` provider pattern from #4297:

- Extend `RingStatsSnapshot` with `updates_accepted` / `updates_rate_limited` / `updates_capacity_dropped` (`network_status.rs`).
- Read them off `Ring.update_rate_limiter` in the provider closure (`p2p_impl.rs`).
- Render three tiles — **UPDATEs relayed**, **Rate-limited**, **Capacity-dropped** — on the Connection Status card, shown only once the limiter has seen traffic so idle nodes stay uncluttered (`home_page.rs`).
- Drop the now-unnecessary `allow(dead_code)` attrs on the three accessors.

This is **observability only — no behavior change**. The limiter's thresholds and drop behavior are untouched; the change consistent with the design-doc principle that "the dashboard visualises what the system already does" (`docs/design/contract-hardening.md`, Phase 4.5).

## Testing

- `network_status::ring_stats_provider_round_trip` — extended to assert the three new counters round-trip through the provider.
- `home_page::rate_limit_stats_hidden_when_no_traffic` — idle node hides the row.
- `home_page::rate_limit_stats_rendered_when_active` — counts + labels appear.
- `home_page::rate_limit_stats_shown_when_only_accepted` — boundary: accepted>0, zero drops still renders.

`cargo fmt`, `cargo clippy -p freenet --lib -- -D warnings`, and the touched tests all pass.

## Context

Follow-up to #4285. Part of the release-readiness review ahead of v0.2.68 — adds operator visibility into the one governance feature that enforces (drops traffic) by default, per maintainer direction to ship the limiter as-is with a visible counter rather than re-tuning the threshold speculatively.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -50,6 +50,17 @@ pub struct RingStatsSnapshot {
     pub peer_id: String,
     /// Base58-encoded full 32-byte X25519 public key.
     pub own_pub_key: String,
+    /// Total relayed UPDATEs accepted by the per-(sender, contract) rate
+    /// limiter since startup (see `ring::update_rate_limit`).
+    pub updates_accepted: u64,
+    /// Total relayed UPDATEs dropped because the per-(sender, contract)
+    /// rate exceeded the limit. A rising value means legitimate traffic
+    /// may be getting dropped — operators should watch this.
+    pub updates_rate_limited: u64,
+    /// Total relayed UPDATEs dropped because the limiter's tracking map
+    /// was at capacity (`MAX_TRACKED_PAIRS`). A non-zero value suggests
+    /// identity churn / admission pressure, distinct from per-pair rate.
+    pub updates_capacity_dropped: u64,
 }
 
 static GOVERNANCE_PROVIDER: parking_lot::RwLock<Option<GovernanceProvider>> =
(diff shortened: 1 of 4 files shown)


sanity:
## Review (Light tier — observability change, no high-risk surface)

Per the multi-model-review policy this is a Light-tier change (small, self-contained, no concurrency / wire-format / auth / migration surface). Ran an external Codex pass plus an adversarial read.

**Codex (external, non-Claude):** no MEDIUM/HIGH/BLOCKER findings. One LOW — the hide-when-zero guard used unchecked `u64` addition across the three counters (debug-build overflow panic / release wrap, operationally unreachable). **Fixed** in `7864eb2f` by switching to `any > 0`. Verdict: safe to merge into v0.2.68.

**Adversarial read:** confirmed the three counters are disjoint in `check_and_record` (`rejected_total` at update_rate_limit.rs:257 = rate path only; `capacity_rejected_total` at :276 = capacity path only; `accepted_total` at :264/:280), so the "Rate-limited" / "Capacity-dropped" / "UPDATEs relayed" tiles don't double-count. Provider reads are non-blocking `Relaxed` atomic loads on the homepage path. Values are `u64` interpolated as numbers — no HTML-injection surface. Dead-code `allow` removal is safe (accessors now used at p2p_impl.rs).

No behavior change — limiter thresholds and drop logic are untouched; this only surfaces what the limiter already records.

[AI-assisted - Claude]

PR #4318, opened 2026-05-31
fix(release): sign River announcements with the owner key override

sanity:
## Problem

Release announcements to the Freenet Official River room have silently failed since at least v0.2.67. `riverctl message send` exits 0 ("Message sent successfully"), the release-agent logs "OK", and the workflow goes green — but the message **never converges into room state**. Both the v0.2.67 and v0.2.68 announcements were absent from the room (confirmed by reading the room back through nova's own node — not a propagation or technic-side issue, and not a contract-key mismatch).

Root cause is the **signing identity**. The script pre-patched `rooms.json`'s `signing_key_bytes` to the owner key, then ran `riverctl message send` with no explicit key. Without an explicit key, riverctl signs with whatever identity `rooms.json` holds — and the **chat-delegate sync can silently rewrite that field** to a non-owner ("bot") identity (on nova, `self_authorized_member: null`). riverctl's own `--help` documents this exact hazard and offers `--signing-key-file` as the remedy ("signs at command time, without the UI's chat-delegate sync"). When the delta is signed by a non-owner, the room contract is valid at the request layer but **silently drops it on merge** while riverctl still reports success.

## Solution

Sign via riverctl's global `--signing-key-file` override — an **in-memory** identity that is never persisted and is what `resolve_signing_key` returns, so it's immune to on-disk drift / the delegate-sync rewrite. The brittle `rooms.json` patch is removed. Applied in **both** announce paths:
- `scripts/release-agent/announce-to-river.sh` (the automated release-agent path) — send extracted into `post_message()`.
- `scripts/release.sh` (the manual fallback) — same override, same patch removal.

Verified live on nova: re-sending the v0.2.68 message with the owner-key override **merged into converged room state and persists** (`[13:55:30 - Room Owner]: Freenet v0.2.68 released: …`), so the v0.2.68 announcement is now out.

## Testing

`announce-to-river_test.sh` gains four assertions (eval-extracting the real `post_message`, same pattern as the existing `log`/`wait_for_node` tests): the riverctl invocation passes `--signing-key-file <owner-key> message send` in the correct order, sends to the owner VK, runs from the source repo, and sets `RIVER_SKIP_CONTRACT_CHECK`. Drop or misplace the override and they fail (verified by the reviewer reproducing the revert). Existing tests pass; `shellcheck` clean on both scripts (no new warnings in `release.sh`).

## Reviews

Codex (external) + an adversarial skeptical review — both verdict **safe to merge**, no correctness defects. Findings addressed: corrected the rationale prose (the mechanism is the chat-delegate sync rewriting `signing_key_bytes`, per riverctl's own docs — not riverctl "running" a sync), brought `release.sh` in line (finish-the-fix), and made the contract-check test assertion actually verify the env var.

## Follow-up note

riverctl reporting `rc=0`/"Message sent successfully" for a delta the contract later drops on merge is a real observability gap that masked this for multiple releases — worth surfacing the UpdateResponse, but out of scope here.

[AI-assisted - Claude]

### DIFF
--- a/scripts/release-agent/announce-to-river.sh
+++ b/scripts/release-agent/announce-to-river.sh
@@ -96,19 +96,24 @@ wait_for_node() {
 # Post $MESSAGE to the room, signed by the room OWNER.
 #
 # The signing identity is load-bearing and the source of a long-standing
-# silent failure. riverctl's normal send path runs the chat-delegate sync,
-# which rewrites rooms.json's `signing_key_bytes` back to the locally-loaded
-# identity (on nova that's an unauthorized "bot" key, not the owner) right
-# before signing. If we sign with that identity, the room contract is valid
-# at the request layer but SILENTLY DROPS the delta on merge — riverctl still
+# silent failure. Without an explicit key, riverctl signs with whatever
+# identity rooms.json currently holds, and the chat-delegate sync can
+# silently rewrite that `signing_key_bytes` field to a non-owner key (on
+# nova it resolves to an unauthorized "bot" identity). riverctl documents
+# this exact hazard and provides `--signing-key-file` as the remedy (see its
+# `--help`: "signs at command time, without the UI's chat-delegate sync").
+# When the message is signed by a non-owner, the room contract is valid at
+# the request layer but SILENTLY DROPS the delta on merge — riverctl still
 # prints "Message sent successfully" and exits 0, yet the message never
 # converges into room state. Every release announcement before this fix was
 # lost exactly this way (v0.2.67 and v0.2.68 were both absent from the room).
 #
-# Pre-patching rooms.json (what this script used to do) does NOT work: the
-# delegate sync overwrites the patch in-process before signing. The fix is
-# riverctl's global `--signing-key-file` override, which is in-memory and
-# immune to the sync, so the delta is signed by the real owner and merges.
+# Pre-patching rooms.json's `signing_key_bytes` (what this script used to do)
+# is unreliable for the same reason — the sync can overwrite it before the
+# send signs. The fix is the global `--signing-key-file` override, which is
+# held in memory, never persisted, and is what riverctl returns from
+# resolve_signing_key, so the delta is deterministically signed by the real
+# owner and merges.
 #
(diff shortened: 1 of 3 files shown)


sanity:
## Review (external Codex + adversarial skeptical pass)

Release-tooling change; ran an external non-Claude (Codex) review plus an adversarial skeptical review. **Both verdicts: safe to merge, no correctness defects.** The skeptical reviewer independently reproduced the regression test failing when the `--signing-key-file` override is removed or misplaced, and confirmed (against the riverctl source) that the override genuinely reaches the in-memory signing path and is never persisted.

Findings addressed in `3ce7c76b`:
- **M2 (finish-the-fix):** `scripts/release.sh`'s manual announce path had the same broken pattern — applied the same `--signing-key-file` override and dropped the `rooms.json` patch there too.
- **M1 (rationale accuracy):** reworded the comment/PR prose. The mechanism is the **chat-delegate sync rewriting `rooms.json`'s `signing_key_bytes`** to a non-owner identity (documented in riverctl's own `--help` / `storage.rs` as the reason `--signing-key-file` exists) — not riverctl "running a sync" in its send path.
- **Codex LOW:** the contract-check test assertion now verifies `RIVER_SKIP_CONTRACT_CHECK` is actually set (the `cargo` stub records the inherited env), not just argv.

Low/by-design notes (no action): owner key must be valid raw 32 bytes (riverctl errors clearly if not); rooms.json still required and still validated.

[AI-assisted - Claude]

sanity:
Added the **`test-exempt`** label. This is a `fix:` PR but the fix is entirely in shell scripts (`scripts/release-agent/announce-to-river.sh`, `scripts/release.sh`), so the Rule Lint check for a new Rust `#[test]` can't apply. The fix **does** carry a regression test — `announce-to-river_test.sh` gains assertions that the riverctl invocation passes `--signing-key-file <owner-key>` before the `message` subcommand and sets `RIVER_SKIP_CONTRACT_CHECK`. The skeptical reviewer independently reproduced these failing when the override is removed or misplaced, so the test genuinely guards the regression — it's just shell, not Rust.

[AI-assisted - Claude]

Issue #4325, opened 2026-06-02
dev: clippy fails under --all-features and under --tests (pre-existing, outside CI gate)


sanity:
## Summary

Two pre-existing clippy failures that are NOT caught by CI (CI runs `cargo clippy --locked -- -D warnings`, i.e. default features and no test code), but bite anyone running a broader clippy locally:

### 1. `--all-features`: OpenTelemetry `SpanBuilder` API mismatch
`cargo clippy -p freenet --all-features` fails to compile:

```
error[E0560]: struct `opentelemetry::trace::SpanBuilder` has no field named `span_id`
  --> crates/core/src/tracing.rs:2259
error[E0560]: struct `opentelemetry::trace::SpanBuilder` has no field named `trace_id`
  --> crates/core/src/tracing.rs:2260
```

The `trace-ot` feature pulls in an opentelemetry version whose `SpanBuilder` no longer exposes `span_id` / `trace_id` fields (available fields: `span_kind`, `events`, `links`). Either the dep needs pinning or the construction in `tracing.rs` needs updating to the current API.

### 2. `--tests`: wildcard_enum_match_arm in operations.rs
`cargo clippy --locked -p freenet --tests -- -D warnings` fails:

```
error: wildcard match will also match any future added variants
  --> crates/core/src/operations.rs:2332
  = note: `-D clippy::wildcard-enum-match-arm` implied by `-D warnings`
```

A `_ => ErrorKind::OperationError { .. }` arm over `OpError` trips the lint when test code is compiled. Either enumerate the variants or add a scoped `#[allow(clippy::wildcard_enum_match_arm)]` with a justifying comment.

## Why file rather than fix inline

Both are pre-existing on `main` and unrelated to the change that surfaced them (PR #4323, a config guard fix). Filing so they're tracked; whoever touches these areas (or wants to tighten the CI clippy invocation to include `--tests`) can clear them.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-06-09T14:08:03.774306953+00:00
Issue hash: `c0709bd0`

## Value Assessment

Both issues have been addressed. The `operations.rs` file at line 2332 no longer exists in its original form (the file is only 999 lines and the operations code has been refactored into a directory), and `#[allow(clippy::wildcard_enum_match_arm)]` suppressions are now present in the relevant files. The `tracing.rs` SpanBuilder issue with `span_id`/`trace_id` fields still exists at lines 2259-2260.

[...]

Issue #4225, opened 2026-05-24
build: opentelemetry 0.32 broke SpanBuilder usage in `trace-ot` feature


sanity:
## Summary

Building with the `trace-ot` feature fails on `main` after the workspace bump
to `opentelemetry = "0.32"` (and `opentelemetry_sdk = "0.32"`,
`tracing-opentelemetry = "0.33"`). The error is masked because the default
CI matrix doesn't enable `trace-ot`, but any developer running
`cargo clippy -p freenet --all-features -- -D warnings` (or building with
the feature explicitly) hits:

```
error[E0560]: struct `opentelemetry::trace::SpanBuilder` has no field named `span_id`
    --> crates/core/src/tracing.rs:2223:17
     |
2223 |                 span_id: Some(trace::SpanId::from_bytes(span_id)),

error[E0560]: struct `opentelemetry::trace::SpanBuilder` has no field named `trace_id`
    --> crates/core/src/tracing.rs:2224:17
```

## Root cause

`opentelemetry 0.32`'s `SpanBuilder` removed the public `span_id` and
`trace_id` fields. The exported fields are now only `span_kind`, `name`,
`start_time`, `attributes`, `events`, `links`. Span and trace IDs are no
longer overridable on the builder — the SDK's `id_generator.new_span_id()`
always issues a fresh span_id, and the `trace_id` is inherited from a
parent `SpanContext` placed in the `Context`, otherwise generated.

This is a meaningful semantic change for `OTSpan::new` in
`crates/core/src/tracing.rs` (around line 2212), which was deriving both
IDs deterministically from the transaction bytes so multiple spans across
a transaction would share `trace_id`/`span_id`.

## Proposed fix shape

Build a remote parent `SpanContext` carrying the transaction-derived
`trace_id` (and the transaction-derived `span_id` as the parent span_id),
then call `tracer.build_with_context(builder, &parent_cx)` so the new
span inherits the trace_id. The new span itself receives a
generator-assigned span_id (this is a wire-format change in the exported
spans — the per-span ID is no longer transaction-derived).

```rust
use opentelemetry::{
    Context, KeyValue, global,
    trace::{self, Span, TraceContextExt, TraceFlags, TraceState},
};

let parent_span_ctx = trace::SpanContext::new(
    trace::TraceId::from_bytes(tx_bytes),
    trace::SpanId::from_bytes(span_id_bytes),
    TraceFlags::SAMPLED,
    true,
    TraceState::default(),
);
[...]

PR #4326, opened 2026-06-02
fix(transport): sample RTT from ping/pong keep-alive cycle

### LINKED ISSUE #4000: RTT samples are only collected on stream completion (same undercounting as #3996)
## Problem

`TransportMetrics::record_rtt_sample` is only called from `record_transfer_completed` (see `metrics.rs:202` in current main). Connections that exchange only keep-alive / control / small-contract traffic for hours never trigger a stream-transfer completion, so they contribute zero RTT samples. The dashboard / telemetry's RTT statistics (`min_rtt_us`, `max_rtt_us`, `avg_rtt_us`) under-represent quiet connections.

This is the same class of bug as the per-peer SENT/RECV undercounting fixed in #3996 — metrics that should reflect ongoing connection health are tied to a low-frequency event (stream completion).

## Suggested fix

Sample RTT from the regular ping/pong keep-alive cycle in `peer_connection.rs` instead of (or in addition to) at stream completion. The keep-alive task already tracks `pending_pings` with timestamps; computing RTT from the matching pong is a small change.

Out of scope for the original dashboard fix; tracking separately.

## Reference

- Big-picture reviewer flagged this in PR #3996 review (item #1).

[AI-assisted - Claude]

sanity:
## Problem

`TransportMetrics::record_rtt_sample` was only reachable from `record_transfer_completed`, which fires on stream-transfer completion. Connections that exchange only keep-alive / control / small-contract traffic for hours never complete a stream transfer, so they contributed **zero RTT samples**. The dashboard / telemetry RTT statistics (`min_rtt_us`, `max_rtt_us`, `avg_rtt_us`) under-represented quiet connections.

This is the same class of undercounting as the per-peer SENT/RECV bug fixed in #3996: a metric that should reflect ongoing connection health was tied to a low-frequency event.

## Approach

The ping/pong keep-alive cycle already records each Ping's send timestamp in `pending_pings` (`sequence -> nanos`). The Pong handler already removed the matching entry on receipt but **discarded** the timestamp. The fix captures it, computes the round-trip against the same `time_source`, and records an RTT sample. The keep-alive cycle runs on every connection regardless of stream traffic, so RTT statistics now stay populated for quiet, long-lived connections.

- `record_rtt_sample` widened from private to `pub(crate)` so the connection layer can call it (it was already the shared sink used by the stream-completion path).
- The `rtt_us > 0` guard mirrors the existing stream-completion path.
- The `pending_pings` write lock is released before recording the sample.
- No new attack surface: a Pong only yields a sample if it matches a sequence **we** pinged (we control `pending_pings`); unsolicited/replayed Pongs find no entry and record nothing.

This is observation-only and does not feed congestion control — the shadow-RTT system in #4074 remains the path for that.

## Testing

New regression test `pong_records_rtt_sample_from_keepalive_cycle` drives an encrypted Pong through the real `recv()` path after seeding a pending ping, then asserts the global RTT sample counter advances and the matching ping is consumed.

- Verified it **FAILS** without the fix (`record_rtt_sample` disabled → zero samples) and **PASSES** with it.
- Full transport lib suite: 607 passed, 0 failed.
- `cargo clippy --locked -- -D warnings` clean.

Closes #4000

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/transport/metrics.rs
+++ b/crates/core/src/transport/metrics.rs
@@ -255,7 +255,13 @@ impl TransportMetrics {
     }
 
     /// Record an RTT sample in microseconds.
-    fn record_rtt_sample(&self, rtt_us: u64) {
+    ///
+    /// Called both from [`Self::record_transfer_completed`] (LEDBAT
+    /// `base_delay` at stream completion) and from the ping/pong keep-alive
+    /// cycle in `peer_connection.rs` (round-trip of a keep-alive Ping). The
+    /// keep-alive path is what keeps RTT statistics populated for quiet,
+    /// long-lived connections that rarely complete a stream transfer (#4000).
+    pub(crate) fn record_rtt_sample(&self, rtt_us: u64) {
         self.rtt_sum_us
             .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |v| {
                 Some(v.saturating_add(rtt_us))
@@ -288,6 +294,16 @@ impl TransportMetrics {
     /// without interfering with the periodic telemetry snapshots.
     pub fn cumulative_bytes_sent(&self) -> u64 {
         self.cumulative_bytes_sent.load(Ordering::Relaxed)
+    }
+
+    /// Number of RTT samples accumulated in the current snapshot window.
+    ///
+    /// Reset to 0 by `take_snapshot`. Exposed so tests can assert that a
+    /// code path (e.g. the ping/pong keep-alive cycle) actually recorded a
+    /// sample without consuming the snapshot.
+    #[cfg(test)]
+    pub(crate) fn rtt_sample_count(&self) -> u32 {
+        self.rtt_samples.load(Ordering::Relaxed)
     }
 
     /// Record a completed inbound stream transfer.
--- a/crates/core/src/transport/metrics.rs
+++ b/crates/core/src/transport/metrics.rs
@@ -526,10 +526,18 @@ impl TransportMetrics {
             self.cwnd_sum.store(0, Ordering::Relaxed);
             self.cwnd_samples.store(0, Ordering::Relaxed);
             self.slowdowns_triggered.store(0, Ordering::Relaxed);
-            self.min_rtt_us.store(u64::MAX, Ordering::Relaxed);
-            self.max_rtt_us.store(0, Ordering::Relaxed);
-            self.rtt_sum_us.store(0, Ordering::Relaxed);
-            // `rtt_samples` was already swapped to 0 above.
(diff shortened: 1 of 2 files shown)


sanity:
## Multi-model review summary

Risk tier: **Full** (transport / telemetry surface). Self-review across four lenses (code-first, testing-gaps, skeptical/adversarial, big-picture) plus three rounds of external **Codex** review on the evolving diff.

### Codex findings (all addressed)

1. **[P2] Keep-alive RTT samples discarded by telemetry worker.** `take_snapshot()` early-returned `None` (and reset the RTT counters) whenever `transfers_completed == 0 && transfers_failed == 0`, so a quiet connection's keep-alive RTT samples never reached the OTLP telemetry path — the dashboard saw them, telemetry didn't.
   → **Fixed**: RTT samples now count as activity in the snapshot predicate (commit `1acb20f8`). Regression test `test_snapshot_emitted_for_rtt_only_activity` (fails without the predicate change, passes with it).

2. **[P2/P3] RTT count/accumulator desync under concurrent sampling.** The first iteration of the predicate fix swapped `rtt_samples` to 0 *before* the activity check; a concurrent `record_rtt_sample` could then have its sum/min/max consumed by the current snapshot while its count carried to the next, yielding `avg/min/max = 0` for a real sample.
   → **Fixed**: the predicate now `load`s the count (peek, no reset) and the count is swapped exactly once, in lockstep with the `rtt_sum_us`/min/max swaps (commit `61201e7d`). The no-activity branch leaves all four RTT atomics untouched so a sample landing concurrently with the reset stays self-consistent.

Final Codex pass: *"I did not find any discrete correctness issue introduced by the patch."*

### Self-review notes

- **No new attack surface**: a Pong yields an RTT sample only if it matches a sequence we actually pinged (we own `pending_pings`); unsolicited/replayed Pongs find no entry and record nothing.
- **Observation-only**: does not feed congestion control; the shadow-RTT system (#4074) remains the path for that.
- `pending_pings` write lock is released before the metrics call.

### Verification
- `pong_records_rtt_sample_from_keepalive_cycle` (peer_connection) and `test_snapshot_emitted_for_rtt_only_activity` (metrics) both verified to FAIL without their respective fix and PASS with it.
- Full `freenet` lib suite: 2884 passed, 0 failed. Transport lib: 608 passed.
- `cargo clippy --locked -- -D warnings` clean.

(Note: `cargo clippy --all-features` surfaces a pre-existing `trace-ot` / opentelemetry-0.32 break unrelated to this change — already tracked in #4225 / #4325, and outside the CI gate which runs default features.)

[AI-assisted - Claude]

PR #4328, opened 2026-06-02
fix: refresh webapp cache on ?__sandbox=1 and subresource requests

### LINKED ISSUE #3977: Gateway webapp cache not refreshed when only ?__sandbox=1 is requested after republish
## Problem

After a contract is republished, the gateway's local \`webapp_cache/\` directory keeps serving the old extracted bundle until something triggers \`ensure_contract_cached\`. Requests to the iframe content URL (\`?__sandbox=1\`) **do not** trigger that path — they go through \`serve_sandbox_content\`, which reads files directly from disk without checking whether network state has changed.

End-user impact today is limited (a typical first visit hits \`/\` before the iframe loads \`?__sandbox=1\`, so \`contract_home\` runs and refreshes the cache). But it shows up as:

1. **Publisher verification papercut** — a developer who runs \`cargo make publish-delta\` and then curls \`?__sandbox=1\` to confirm the new bundle is being served sees the OLD bundle. Hitting \`/\` first refreshes; without that step the verification looks like the publish failed even when it succeeded.
2. **Cross-contract / direct iframe sources** — anything that loads \`?__sandbox=1\` without first going through \`contract_home\` (a PWA, mobile webview, or a service worker that cached the outer HTML) will keep serving stale content indefinitely.
3. **Mid-session network updates** — if the network state changes while a browser session is open, the iframe stays on the old bundle until the user refreshes the outer page. Hard refresh of just the iframe (which a user can't easily do anyway) wouldn't help either.

## Repro

This was hit on 2026-04-29 while republishing Delta (\`freenet/delta\` PR #11 → #16):

\`\`\`
$ cargo make publish-delta
... fdev publish reports "Contract published successfully"

$ fdev execute get EqJ5YpEEV3XLqEvKWLQHFhGAac2qXzSUoE6k2zbdnXBr -o /tmp/state.bin
# extract embedded tar.xz from state, b3sum -> matches the just-built target/webapp/webapp.tar.xz

$ curl "http://127.0.0.1:7509/v1/contract/web/EqJ5YpE.../?__sandbox=1" \\
    | grep -oE 'delta-ui-[^\"]*\\.js'
delta-ui-dxhfc347d41f92c50fc.js     # OLD bundle from previous publish

$ curl "http://127.0.0.1:7509/v1/contract/web/EqJ5YpE.../?nocache=$(date +%s%N)" >/dev/null   # hit outer URL
$ curl "http://127.0.0.1:7509/v1/contract/web/EqJ5YpE.../?__sandbox=1" \\
    | grep -oE 'delta-ui-[^\"]*\\.js'
delta-ui-dxh22c99af3a5296e9.js     # NEW bundle, after \`/\` triggered ensure_contract_cached
\`\`\`

The network state IS the new one (\`fdev execute get\` confirms). The local extracted cache just isn't refreshed until \`contract_home\` runs.

## Root cause

\`crates/core/src/server/path_handlers.rs\`:

- \`contract_home\` (the \`/\` handler) calls \`ensure_contract_cached\` (line 76), which always does a fresh client API GET, hashes the response, and calls \`unpack_if_stale\` (line 201). \`unpack_if_stale\` re-extracts the tar.xz only if the new state hash differs from the stored \`{key}.hash\` file. This path is correct.

- \`serve_sandbox_content\` (the \`?__sandbox=1\` handler, called when the iframe loads its content) reads files from \`contract_web_path(&instance_id)\` directly. It does not call \`ensure_contract_cached\`, so it never reconciles disk state against network state.

- \`variable_content\` (cross-contract subresource handler) does check via the \`{key}.hash\` file existence (line 325) and falls through to \`ensure_contract_cached\` only if the hash file is missing. That's also a one-shot check — once the hash file exists, refreshes never happen on this path either.

## Suggested approaches

Three options, increasing in scope:

1. **Cheapest**: have \`serve_sandbox_content\` also call \`ensure_contract_cached\` before serving. Cost: every iframe content request does a network round-trip + state-hash compare. Mitigation: gate on a TTL in memory (\"if we re-checked this contract less than N seconds ago, skip\"). The current \`unpack_if_stale\` uses a per-instance Mutex so concurrent requests serialise; a TTL would extend that pattern.

2. **Subscribe-based**: \`ensure_contract_cached\` already does a one-shot GET. Make it (or a new path) subscribe to the contract instead, and invalidate the on-disk \`{key}.hash\` on UpdateNotification. Subsequent \`serve_sandbox_content\` requests would naturally re-fetch via path 1's check on missing hash, OR the subscription handler could re-extract proactively. This eliminates polling entirely and is consistent with the design intent of #3698 (close hosted-contract refresh gaps via subscription rather than re-GET).

3. **Server-side eviction on PUT**: when the local node observes a successful PUT for a contract whose webapp_cache exists, immediately delete the \`{key}.hash\` file. Next sandbox/variable request triggers a fresh fetch. This handles the publisher-verification path cleanly but doesn't help the cross-gateway case (peer A publishes, peer B's gateway doesn't observe a local PUT).

(2) is probably the right long-term fix; (3) covers the immediate publisher pain at much lower cost. (1) without (3) feels like the wrong tradeoff because it adds latency to every page load.

## Why filing now

Came up while republishing Delta after PRs freenet/delta#13 and freenet/delta#16. The publish succeeded on the network (verified via \`fdev execute get\` round-tripping the embedded tar.xz), but \`?__sandbox=1\` kept showing the previous bundle until I deleted \`{key}.hash\` from \`webapp_cache/\` and curled the outer URL. Real users hitting \`/\` for the first time would not see this; for now it's a recurring papercut for anyone running publish + verify locally. Worth catching before it becomes a user-visible bug as more dApps and embedded-iframe use cases land.

[AI-assisted - Claude]

sanity:
## Problem

After a contract is republished, the gateway's local `webapp_cache/` keeps
serving the old extracted bundle on the iframe-content path. `serve_sandbox_content`
(the `?__sandbox=1` handler) read files straight from disk, checking only
`path.exists()` — it never reconciled disk state against the network. Only
`contract_home` (the `/` shell-root handler) unconditionally re-fetches via
`ensure_contract_cached`, so the new bundle only appeared after a fresh hit on `/`.

`variable_content` (subresource handler) had the same one-shot weakness: it fetched
only when the `{key}.hash` file was missing (cold cache), never when the on-disk
cache was stale.

Impact (per #3977):
1. Publisher verification papercut — `publish` succeeds on the network, but
   `curl ?__sandbox=1` shows the OLD bundle until `/` is hit, making a good
   publish look like a failure.
2. Cross-contract / direct iframe sources (PWA service worker, mobile webview,
   cached outer shell) that bypass `contract_home` serve stale content indefinitely.
3. Mid-session network updates don't reach an open iframe until the outer page reloads.

## Approach

Add a TTL-gated `refresh_cache_if_due` helper that calls the existing
`ensure_contract_cached` when either:
- the cache is cold (no `{key}.hash` on disk), or
- more than `CONTRACT_CACHE_REFRESH_TTL` (30s) has elapsed since the last
  reconciliation for that contract (tracked in a process-global
  `DashMap<ContractInstanceId, Instant>`, mirroring the existing
  `CONTRACT_CACHE_LOCKS` pattern).

`serve_sandbox_content` and `variable_content` both route through it.
`contract_home` records the refresh timestamp after its own fetch, so the iframe
load that immediately follows reuses the fresh state instead of issuing a redundant
GET within the window.

This is the issue's option 1 (cheapest) with the recommended in-memory TTL gate so
iframe loads don't take a network round-trip on every request. The actual
re-extraction still only happens when the state hash changed (`unpack_if_stale`
re-reads the hash under a per-contract mutex), so a same-state refresh costs one
network GET, not a disk rewrite, and the TTL bounds the GET rate to at most one per
contract per window. The timer is advanced only on a successful refresh, so a
transient fetch failure doesn't suppress the next request's retry.

## Testing

New `#[test]`s in `path_handlers.rs` (verified to fail before the fix by neutering
`refresh_cache_if_due`, then pass after):
- `serve_sandbox_content_triggers_refresh_when_stale` — warm-but-unreconciled cache
  emits the `NewConnection`+`Get` fetch pair (the core bug: pre-fix sent nothing).
- `serve_sandbox_content_skips_refresh_when_fresh` — inside the TTL window, serves
  from disk with no network round-trip.
- `refresh_cache_if_due_refetches_after_ttl_expires` — uses paused time + `advance()`
  to deterministically cross the TTL boundary and confirm a re-fetch.
- Updated `variable_content_skips_fetch_when_cache_present_and_fresh` for the new
  TTL semantics; the existing `variable_content_triggers_fetch_on_cache_miss` and
  `web_subpages_sandbox_branch_runs_before_redirect_branch` pin test still pass.

`cargo fmt`, `cargo clippy -p freenet --locked -- -D warnings`, and the full
`server::path_handlers` + `server::client_api` test suites are green.

Closes #3977

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -199,7 +199,7 @@ async fn web_home(
         .unwrap_or(false);
 
     if is_sandbox {
-        return serve_sandbox_response(key, api_version, None, &req_headers).await;
+        return serve_sandbox_response(key, api_version, None, &req_headers, rs).await;
     }
 
     // Shell page: generate auth token, serve iframe wrapper with CSP
@@ -269,7 +269,14 @@ async fn web_subpages(
     // For sandbox sub-page requests to HTML files, serve through the sandbox
     // content pipeline (with WebSocket shim + navigation interceptor injected).
     if is_sandbox && is_html_page(&last_path) {
-        return serve_sandbox_response(key, api_version, Some(&last_path), &req_headers).await;
+        return serve_sandbox_response(
+            key,
+            api_version,
+            Some(&last_path),
+            &req_headers,
+            request_sender,
+        )
+        .await;
     }
 
     // Top-level document loads of contract HTML sub-paths (pasted URLs,
@@ -440,6 +447,7 @@ async fn serve_sandbox_response(
     api_version: ApiVersion,
     sub_path: Option<&str>,
     req_headers: &axum::http::HeaderMap,
+    request_sender: HttpClientApiRequest,
 ) -> Result<axum::response::Response, WebSocketApiError> {
     // Block top-level navigation to sandbox URLs. Sec-Fetch-Dest: iframe is set
     // by the browser automatically and cannot be spoofed by scripts.
@@ -452,7 +460,7 @@ async fn serve_sandbox_response(
     }
 
     let contract_response =
-        path_handlers::serve_sandbox_content(key, api_version, sub_path).await?;
+        path_handlers::serve_sandbox_content(key, api_version, sub_path, request_sender).await?;
     let mut response = contract_response.into_response();
     add_sandbox_cors_headers(&mut response);
     // See `sandbox_csp_for_origin` for why we interpolate a concrete origin
@@ -945,7 +953,7 @@ mod tests {
         // top-level-document redirect inside `web_subpages`. Both
         // markers are unique to that function.
         let sandbox_idx = src
-            .find("serve_sandbox_response(key, api_version, Some(&last_path)")
+            .find("if is_sandbox && is_html_page(&last_path) {")
             .expect("sandbox short-circuit marker present in web_subpages");
         let redirect_idx = src
             .find("should_redirect_subpage_to_shell(is_sandbox")
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -165,8 +165,12 @@ async fn refresh_cache_if_due(
     // Slow path: refresh looks due. Serialize concurrent refreshers for this
     // contract so only the first issues a GET; the rest re-check below.
     let _guard = acquire_refresh_lock(&instance_id).await;
-    if cache_warm && cache_reconciled_recently(&instance_id) {
-        // A concurrent refresher completed while we waited for the lock.
+    // Re-check on the timer alone (not the pre-lock `cache_warm` snapshot): a
+    // concurrent refresher that completed while we waited recorded a fresh
+    // timer AND populated the cache via `ensure_contract_cached`, so a fresh
+    // timer means there is nothing left to do even if our snapshot saw the
+    // cache as cold.
+    if cache_reconciled_recently(&instance_id) {
         return Ok(());
     }
 
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -84,6 +84,37 @@ const CONTRACT_CACHE_REFRESH_TTL: Duration = Duration::from_secs(30);
 static CONTRACT_CACHE_REFRESH: LazyLock<DashMap<ContractInstanceId, Instant>> =
     LazyLock::new(DashMap::new);
 
+/// Per-contract lock serializing the *decision* to issue a staleness-refresh
+/// GET, so a fan-out of concurrent subresource requests after the TTL expiry
+/// issues at most one `ensure_contract_cached` GET per contract per window.
+///
+/// This is deliberately distinct from `CONTRACT_CACHE_LOCKS`: that lock guards
+/// the on-disk unpack and is re-taken inside `unpack_if_stale`. `tokio`'s mutex
+/// is not reentrant, so the refresh gate — which is held *across* the GET (and
+/// therefore across `unpack_if_stale`'s own lock acquisition) — must use its
+/// own mutex to avoid a self-deadlock.
+static CONTRACT_REFRESH_LOCKS: LazyLock<DashMap<ContractInstanceId, Arc<tokio::sync::Mutex<()>>>> =
+    LazyLock::new(DashMap::new);
+
+async fn acquire_refresh_lock(
+    instance_id: &ContractInstanceId,
+) -> tokio::sync::OwnedMutexGuard<()> {
+    let mutex = CONTRACT_REFRESH_LOCKS
+        .entry(*instance_id)
+        .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
+        .clone();
+    mutex.lock_owned().await
+}
+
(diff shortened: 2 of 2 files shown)


sanity:
## Review summary

**Self-review (four lenses)** — code-first, testing-gaps, skeptical/adversarial, big-picture:
- Confirmed the fix targets the root cause: `serve_sandbox_content` and `variable_content` served from disk without reconciling against network state; both now route through the TTL-gated `refresh_cache_if_due`.
- Verified no reentrancy deadlock — the refresh gate uses a dedicated `CONTRACT_REFRESH_LOCKS` mutex, distinct from `CONTRACT_CACHE_LOCKS` which `unpack_if_stale` re-acquires (tokio mutexes are not reentrant).
- The refresh timer advances only on success, so a transient fetch failure does not suppress the next retry.
- Finish-the-fix: `variable_content` had the same one-shot weakness (only fetched on a missing hash file, never on a stale one) and is fixed in the same pass.
- Every regression test was verified to FAIL before the fix (by neutering the relevant code) and PASS after.

**External model review (Codex CLI, `codex review --base main`):**
- First pass raised one P2: under a browser subresource fan-out just after the TTL expiry, concurrent requests could each fire their own redundant GET, breaking the documented "at most one GET per contract per window" bound.
- Addressed by adding the double-checked per-contract refresh lock and `refresh_cache_if_due_coalesces_concurrent_refreshes` (drives 8 concurrent refreshers end-to-end; asserts exactly one GET — fails with 7 extra GETs without the fix).
- Re-ran Codex on the final content: no remaining issues.

CI is being watched; auto-merge will be enabled once green.

[AI-assisted - Claude]

sanity:
## Addressed the rule-review warning

**WARNING — missing failure-path test for `refresh_cache_if_due`:** Fixed in 2e5cfc12. Added `refresh_cache_if_due_does_not_record_timer_on_fetch_failure`, which drives a real refresh whose GET fails (a `contract: None` GetResponse → `MissingContract`) and asserts (a) the call returns `Err` and (b) no `CONTRACT_CACHE_REFRESH` timer was recorded. Verified it FAILS if the `insert` is hoisted before the `ensure_contract_cached` call — the exact refactor the warning flagged.

**INFO — unbounded `CONTRACT_CACHE_REFRESH` / `CONTRACT_REFRESH_LOCKS` DashMaps:** Acknowledged, not blocking. Both are keyed by `ContractInstanceId` and bounded by the number of distinct web contracts the node has served — identical in shape and lifetime to the pre-existing `CONTRACT_CACHE_LOCKS`, which has carried the same property without a cap. Adding an LRU/cap here would be a separate, broader change that should also cover `CONTRACT_CACHE_LOCKS` for consistency; out of scope for this bug fix. Each entry is a single `Instant` or a 3-word mutex, so the footprint is trivially bounded in practice.

[AI-assisted - Claude]

PR #4245, opened 2026-05-24
fix: populate hop_count on terminal GET events

sanity:
## Summary

Production telemetry bug: `hop_count` is currently `None` on the vast majority of terminal GET events because it's computed at log time via `op_manager.get_current_hop(id)`, which returns `None` whenever the operation has been cleaned up before the response is logged. As a result every dashboard / telemetry / benchmark that looks at GET routing depth is silently working with empty data.

This PR threads the hop count through the wire (positional field on `GetMsg::Response`) so the originator has a populated value when it constructs `GetSuccess` / `GetNotFound` log events. Also fills in `hop_count` at the relay that emits an exhaustion `NotFound` (it knows its own `max_htl - htl`).

After this fix, `hop_count` is populated on 100% of *inline* terminal GET events in deterministic-simulation testing across N ∈ {20, 50, 100, 200} (verified during the hop-count benchmark sweep — see #4237).

## Scope

**Inline GET responses only.** Specifically:

- ✅ `GetMsg::Response{Found, value: Some(state)}` → emits `GetSuccess` with populated hop_count
- ✅ `GetMsg::Response{NotFound}` → emits `GetNotFound` with populated hop_count (interpreted as *exhaustion depth*, not path-to-storer)
- ❌ `GetMsg::ResponseStreaming` (large-payload GETs above streaming threshold) — does NOT carry hop_count. Tracked in **#4249**.
- ❌ `GetMsg::Response{Found, value: None}` (failure branch) — `GetEvent::GetFailure` exists in tracing but has no live emission site in the current tree.
- ❌ PUT / UPDATE / SUBSCRIBE all have the same `get_current_hop()` bug. Tracked in **#4248**.

## Wire-format compatibility

Adds a positional field to a bincode-serialized type. Bincode does not honour `#[serde(default)]` for positional encoding, so cross-version compatibility is gated by `MIN_COMPATIBLE_VERSION` at the handshake layer:

- Bumps `crates/core/Cargo.toml` `version` and `min-compatible-version` both from `0.2.62` → `0.2.63`.
- Bumps `crates/fdev/Cargo.toml` freenet path dep correspondingly.
- Verified at `crates/core/src/transport/connection_handler.rs:2584-2609`: handshake gracefully rejects mixed-version peers in both directions via `ack_error` (no crash).

## Test plan

- [x] `cargo build -p freenet --features "testing" --tests` clean
[...]

PR #4329, opened 2026-06-02
fix(get): populate hop_count on streaming GET successes

### LINKED ISSUE #4249: telemetry: hop_count not populated on streaming GET successes (GetMsg::ResponseStreaming)
## Background

PR #4245 added a `hop_count: usize` field to `GetMsg::Response` and threaded it through storer/relay paths so the originator's tracing layer can populate it on `GetSuccess` / `GetNotFound` events.

## The remaining gap

`GetMsg::ResponseStreaming` (the variant for large GET responses that exceed `streaming_threshold`, default 64 KB) does NOT carry `hop_count`. The tracing-layer wire-inspection at `crates/core/src/tracing.rs:1302` only matches `GetMsg::Response{Found}` with `value.state.is_some()` — streaming responses go through a different code path.

This means: terminal GET successes on contracts large enough to use streaming still produce `GetSuccess` events with unpopulated `hop_count`, even after #4245.

Investigation required:
- Does the streaming path eventually emit a `GetSuccess` event, or does the streaming-assembly completion go through a non-`GetSuccess` log channel?
- If it does emit `GetSuccess`, hop_count is `None` because the streaming envelope doesn't carry the field.

## What needs doing

If streaming GETs emit `GetSuccess`:
- Add `hop_count: usize` field to `GetMsg::ResponseStreaming` (positional, same as the fix in #4245).
- Thread the field through the streaming-completion code in `op_ctx_task.rs` (around `Terminal::Streaming` handling at ~line 2392).
- Update tracing-layer extraction to read it from the streaming envelope.
- Bump `min-compatible-version` for the new wire-format change.
- Add roundtrip + classifier-preservation tests.

If they don't emit `GetSuccess`:
- Document that GET successes via streaming are silently missed by telemetry.
- Either route streaming completion through `GetSuccess` (preferred) or accept the gap and document it in the telemetry guide.

## Why not done in #4245

Same scoping reason as #4248: a separate wire-format change with its own review surface area. The inline `Response` path covers the majority of GETs in the deployed workload; streaming is the long-payload exception.

[AI-assisted - Claude]

### LINKED ISSUE #4250: test: un-ignore test_hop_count_populated_on_terminal_get_events once a GET-producing workload is wired through TestConfig
## Background

PR #4245 added `test_hop_count_populated_on_terminal_get_events` in `crates/core/tests/simulation_integration.rs` as a simulation-level regression test for the wire-carried hop_count fix.

It's `#[ignore]`d because the default `event_chain` workload used by `TestConfig::small/.run()` doesn't reliably produce terminal GET events at CI scales. Empirically: 2,840 events produced, 0 of which were GET events (the workload generator chooses ops weighted toward PUT/UPDATE/SUBSCRIBE).

## What needs doing

Two ways forward:

1. **Generalise `run_controlled_simulation`**: `test_get_reliability_diagnostic` (nightly-gated) at `simulation_integration.rs:8174` uses a controlled workload (one PUT, every node GETs the contract) via `run_controlled_simulation`. Refactor this pattern into something the rest of the file can use without the `nightly_tests` feature, then convert the ignored test to use it.

2. **Bias `event_chain` toward GETs**: add a constructor parameter to the event-chain generator that lets a test request a heavier GET proportion. Then the ignored test can configure a GET-heavy workload directly.

Either way, drop `#[ignore]` once the workload deterministically produces ≥10 terminal GET events per run.

## Why not done in #4245

The primary regression coverage already exists at the unit-test level:
- `test_get_msg_response_hop_count_roundtrip` (wire-format roundtrip)
- `classify_response_found_preserves_hop_count` (classifier preserves hop_count from wire into Terminal::InlineFound)

The simulation-level test is value-add for catching multi-node integration regressions, not the primary signal — so getting the workload right was deferred.

[AI-assisted - Claude]

sanity:
## Problem

PR #4245 added a `hop_count` field to `GetMsg::Response` so terminal GET telemetry (`GetSuccess`) could report routing depth, but left `GetMsg::ResponseStreaming` — the variant used for GET responses over `streaming_threshold` (default 64 KB) — without the field.

Worse, in the task-per-tx GET architecture the originator's GET reply is delivered straight to its driver's `pending_op_result` waiter rather than through the event loop's inbound dispatch, so the originator never runs `from_inbound_msg_v1` for its own reply. The implicit `GetSuccess` arm there only fires at relays forwarding an inline `Response{Found}`, and not at all for streamed responses (`ResponseStreaming` fell into the `Get(_) => Ignored` catch-all).

Net effect: streamed GET successes — the large, data-rich contracts most likely to surface routing problems — produced **no terminal GET telemetry at all**. The dashboard's GET success-rate and hop-depth panels were silently blank for any contract large enough to stream, and the simulation regression test for this (`test_hop_count_populated_on_terminal_get_events`, #4250) had to stay `#[ignore]`d because it never saw a terminal GET event.

## Approach

The issue anticipated this ("If they don't emit `GetSuccess`: route streaming completion through `GetSuccess` (preferred)"). Empirically confirmed via instrumentation that `from_inbound_msg_v1` is never reached for the originator's own GET reply, so the fix routes terminal GET telemetry through the driver:

- Add `hop_count: usize` to `GetMsg::ResponseStreaming` and thread it through the driver's `Terminal::Streaming`, `classify`, the storer/upgrade producer (`relay_send_found`), and the relay fork+pipe forward — mirroring the established `PutMsg::ResponseStreaming` pattern.
- Emit `GetSuccess` explicitly from the originator's GET driver (`drive_client_get_inner`) on client-visible success, carrying the wire-carried `hop_count` clamped to `max_hops_to_live` — mirroring how PUT emits from `finalize_put_at_originator`. Covers inline AND streaming, direct AND relayed GETs uniformly.
- Add the matching `from_inbound_msg_v1` `ResponseStreaming` arm so relays emit `GetSuccess` for streamed responses too, restoring symmetry with the inline `Response{Found}` arm.
- Bump `version` / `min-compatible-version` to `0.2.69` for the positional bincode wire-format change.

No double-emission: the originator emits via the driver (relays never run that branch); relays emit via `from_inbound_msg_v1` (the originator never reaches it for its own reply). Each peer attributes one `GetSuccess` to itself, exactly as the inline path already does.

## Testing

- `test_get_msg_response_streaming_hop_count_roundtrip` (get.rs): wire roundtrip for the new field across 5 hop-count values.
- `classify_response_streaming_is_streaming_terminal` (op_ctx_task.rs): classifier preserves `hop_count` into `Terminal::Streaming`.
- `test_streaming_get_emits_get_success_with_hop_count` (streaming_e2e): a deterministic 1 MB relayed streaming GET emits a `GetSuccess` event with a populated, in-range `hop_count`; guarded by the streaming-forward counter so it can't pass on a local-cache shortcut. **Verified to fail before the fix (0 GetSuccess events) and pass after.**
- Un-ignored `test_hop_count_populated_on_terminal_get_events` (#4250): rewired onto a deterministic controlled PUT-then-GET workload (sparse topology forces a routed GET) so it reliably produces terminal GET successes. Stable across repeated runs.

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and the full `streaming_e2e` + GET simulation suites are green locally; a `test_direct_runner_determinism` run confirms the telemetry change is deterministic.

Closes #4249
Closes #4250

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1981,7 +1981,7 @@ dependencies = [
 
 [[package]]
 name = "freenet"
-version = "0.2.68"
+version = "0.2.69"
 dependencies = [
  "aes-gcm",
  "ahash",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "freenet"
-version = "0.2.68"
+version = "0.2.69"
 edition = "2024"
 rust-version = "1.85"
 publish = true
@@ -13,7 +13,7 @@ readme = "README.md"
 # Minimum compatible protocol version. Peers older than this are rejected.
 # Updated by scripts/release.sh during releases. The env var
 # FREENET_MIN_COMPATIBLE_VERSION overrides this at build time.
-min-compatible-version = "0.2.64"
+min-compatible-version = "0.2.69"
 
 [package.metadata.binstall]
 pkg-url = "{ repo }/releases/download/v{ version }/freenet-{ target }.tar.gz"
--- a/crates/core/src/operations/get.rs
+++ b/crates/core/src/operations/get.rs
@@ -128,6 +128,20 @@ mod messages {
             total_size: u64,
             /// Whether the response includes the contract code (not just state)
             includes_contract: bool,
+            /// Forward-path hop count, mirroring `GetMsg::Response.hop_count`
+            /// (PR #4245). Set by the storer (`max_htl - htl_at_storer`) and
+            /// preserved by relays forking + piping the stream upstream — it
+            /// does NOT increment on the return path. Lets the originator's
+            /// tracing layer populate `GetSuccess.hop_count` for large
+            /// (streamed) responses, which previously emitted no terminal
+            /// telemetry event at all (issue #4249).
+            ///
+            /// `#[serde(default)]` is set for source-level clarity. Bincode
+            /// does not honour serde defaults (positional encoding), so wire
+            /// compat with peers that lack this field is handled at the
+            /// handshake layer via `min-compatible-version`.
+            #[serde(default)]
+            hop_count: usize,
         },
 
         /// Acknowledgment that a streaming response was received.
@@ -483,6 +497,54 @@ mod tests {
                     assert_eq!(hc, hop_count, "NotFound.hop_count must roundtrip ({label})")
                 }
                 _ => panic!("expected Response for {label}"),
+            }
+        }
+    }
+
+    /// Regression test (issue #4249): `GetMsg::ResponseStreaming.hop_count`
+    /// roundtrips through bincode.
+    ///
+    /// Before this fix the streaming GET response carried no `hop_count`
+    /// field, so large-contract (streamed) GET successes produced NO
+    /// terminal `GetSuccess` telemetry event at all — the variant fell into
+    /// the `Get(_) => Ignored` catch-all in `from_inbound_msg_v1`. The fix
+    /// carries the value on the wire so the originator's tracing layer can
+    /// populate `GetSuccess.hop_count` for streamed responses, exactly as
+    /// `test_get_msg_response_hop_count_roundtrip` proves for the inline
+    /// `Response` path.
+    ///
+    /// bincode-positional caveat: this positional addition breaks older
+    /// binaries; see the min-compatible-version bump that accompanies this
+    /// PR.
+    #[test]
+    fn test_get_msg_response_streaming_hop_count_roundtrip() {
+        use crate::transport::peer_connection::StreamId;
+        let key = make_contract_key(11);
+        let cases: &[(&str, usize)] = &[
+            ("zero", 0),
+            ("one", 1),
+            ("mid", 4),
+            ("htl", 10),
+            ("large", 64),
+        ];
+        for (label, hop_count) in cases.iter().copied() {
+            let streaming = GetMsg::ResponseStreaming {
+                id: Transaction::new::<GetMsg>(),
+                instance_id: *key.id(),
+                stream_id: StreamId::next(),
+                key,
+                total_size: 128 * 1024,
+                includes_contract: true,
+                hop_count,
+            };
+            let bytes = bincode::serialize(&streaming).expect(label);
+            let restored: GetMsg = bincode::deserialize(&bytes).expect(label);
+            match restored {
+                GetMsg::ResponseStreaming { hop_count: hc, .. } => assert_eq!(
+                    hc, hop_count,
+                    "ResponseStreaming.hop_count must roundtrip ({label})"
+                ),
+                _ => panic!("expected ResponseStreaming for {label}"),
             }
         }
     }
--- a/crates/core/src/operations/get.rs
+++ b/crates/core/src/operations/get.rs
@@ -505,13 +505,10 @@ mod tests {
     /// roundtrips through bincode.
     ///
     /// Before this fix the streaming GET response carried no `hop_count`
-    /// field, so large-contract (streamed) GET successes produced NO
-    /// terminal `GetSuccess` telemetry event at all — the variant fell into
-    /// the `Get(_) => Ignored` catch-all in `from_inbound_msg_v1`. The fix
-    /// carries the value on the wire so the originator's tracing layer can
-    /// populate `GetSuccess.hop_count` for streamed responses, exactly as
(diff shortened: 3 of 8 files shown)


sanity:
## Review summary (multi-lens self-review + external Codex pass)

Ran a self-review across four lenses (code-first, testing-gaps, skeptical/adversarial, big-picture) plus two rounds of external `codex review --base main`. The Codex passes surfaced two real P2 findings, both now fixed; the final Codex pass is clean ("No discrete correctness issues were identified in the diff").

**Finding 1 (Codex P2) — premature streaming-header success.**
The first revision added a `from_inbound_msg_v1` `ResponseStreaming` arm that recorded `GetSuccess` on the stream *header*, before the payload arrived. Unlike inline `Response{Found}` (whose envelope carries the full payload), a streaming header is only metadata and the transfer can still fail. **Resolved:** removed that arm; the streaming `GetSuccess` is now emitted solely by the originator's GET driver, only after `build_host_response`'s local-store re-query confirms the assembled payload is present (`host_result.is_ok()`).

**Finding 2 (Codex P2) — inline GetSuccess double-count.**
An inline `Response` reply flows through `handle_pure_network_message_v1`, whose unconditional `from_inbound_msg_v1` call already emits `GetSuccess` for inline `Response{Found}` before the driver bypass. The unconditional driver emission therefore double-counted inline GET successes (reproduced: one peer logged 2 GetSuccess for one tx). **Resolved:** gated the driver emission to the streaming path (`Terminal::is_streaming()`), which is exactly the path `from_inbound_msg_v1` ignores. Added a permanent no-double-count guard to the un-ignored #4250 test (asserts no peer emits >1 GetSuccess per tx); verified it fails ungated and passes gated.

**Investigation note.** Instrumentation confirmed that in the task-per-tx architecture the originator's GET reply is delivered straight to its driver's `pending_op_result` waiter; for streaming the reply never reaches `from_inbound_msg_v1`, so the originator produced no terminal streaming GET telemetry at all — the core #4249 gap. The driver-side emission (mirroring PUT's `finalize_put_at_originator`) is the correct fix.

**Verification.** `cargo fmt`, `cargo clippy --locked -- -D warnings`, full `streaming_e2e` suite (incl. the new regression test, run 3× in parallel — no flakiness), the un-ignored #4250 test, the get unit tests, and `test_direct_runner_determinism` are all green locally. The streaming regression test is verified to fail before the fix and pass after.

[AI-assisted - Claude]

sanity:
Holding this PR for human review before merge. It turned out to require a **wire-format / protocol change** (bump to v0.2.69, `hop_count` added to `GetMsg::ResponseStreaming` / `Terminal::Streaming`). Wire-format/protocol changes are our highest-risk surface and warrant a human look before they reach the network — even with green CI and a clean external (Codex) review.

This was produced by an autonomous fix agent whose guardrails said to *stop and report* on wire-format changes rather than auto-merge; flagging it here so @sanity can review the protocol-version bump and cross-version compatibility before it lands. Converted to draft + auto-merge disabled to take it out of the merge queue meanwhile. The code itself looks solid (3 regression tests, 2 Codex P2s fixed, un-ignores #4250) — this is purely the merge gate for a protocol change.

[AI-assisted - Claude]

sanity:
Closing this stale draft — the underlying bug is real but this branch is no longer the right vehicle. **Issues #4249 and #4250 stay open** so the work isn't lost.

**Why close rather than rebase:**
- The branch is a month old (base `15ff39a5`, ~105 commits behind main) and the GET driver it threads `hop_count` through has been **heavily restructured** since: #4367 (`drive_get_with_assembly_retry` outer-retry / Terminal reshape), #4404 (migration-toward-key), #4364 (empty-ring gateway fallback), #4416 (fresh-PUT re-broadcast). `op_ctx_task.rs` grew ~+939 lines exactly in the areas this PR touches. `git merge-tree` shows real content conflicts in `op_ctx_task.rs` and `tracing.rs` (plus trivial version-bump conflicts).
- Re-threading the field through `classify` / `Terminal::Streaming` / `relay_send_found` / the relay fork-pipe against the reshaped driver is substantial rework, not a small rebase — and as a positional bincode change to `GetMsg` it needs a fresh version bump (the `0.2.69` here is stale; main is 0.2.88) and a full-tier wire-format re-review.
- It's telemetry-only (blank hop-depth / success panels for >64 KB streamed GETs) — a genuine gap, but not a correctness or user-facing bug, so the cost/benefit favors a fresh redo against the current driver over reviving a stale wire-format branch.

The PR body remains a detailed implementation map for whoever picks up #4249/#4250 next.

[AI-assisted - Claude]

PR #4331, opened 2026-06-02
chore: simplify diagnostics_to_json after stdlib 0.8 bump

### LINKED ISSUE #3993: chore: simplify diagnostics path after stdlib 0.7 bump
## Background

freenet-core PR #3989 landed a workaround for #3987 that stringifies `ContractKey` map keys before JSON-serializing `NodeDiagnosticsResponse` in the report binary. The workaround is forward-compatible with the upcoming stdlib type change, so it doesn't have to land in lockstep with the stdlib release.

freenet-stdlib PR #70 fixes the type at the source: `NodeDiagnosticsResponse.contract_states: HashMap<String, ContractState>` (was `HashMap<ContractKey, _>`). Once that release ships and freenet-core bumps to it, the workaround becomes redundant.

## Tasks (after bumping `freenet-stdlib` to 0.7.0)

- [ ] **Producer compile fix**: `crates/core/src/node/network_bridge/p2p_protoc.rs:1900,1916` calls `response.contract_states.insert(contract_key, ...)` and `response.contract_states.insert(*contract_key, ...)`. Both will fail to compile after the bump. Fix: `.to_string()` on the key.
- [ ] **Simplify `diagnostics_to_json`**: `crates/core/src/bin/commands/report.rs::diagnostics_to_json` does a destructure + manual field-by-field assembly to stringify `contract_states` keys. After the stdlib bump the keys are already strings, so the helper can be replaced with a single `serde_json::to_string_pretty(&diag)` call. The destructure-bind for compile-time field-drift safety can stay if desired, or be removed alongside the helper.
- [ ] **Delete the bug-pinning test**: `test_native_serde_json_on_contract_states_fails_documents_root_cause` in `report.rs` was explicitly written as a tripwire — its docstring tells the next reader to delete it once the stdlib type changes. After the bump, the test will start failing because `serde_json::to_string` on the new type succeeds. Delete it (or invert its assertion to "must succeed").
- [ ] **Verify `crates/fdev/src/diagnostics.rs:158-170`** still compiles (it iterates the map and calls `key.to_string()`, which works on both `&ContractKey` and `&String` — should be source-compatible, no change needed).
- [ ] **Re-run the JSON serialization tests** added in #3989 to confirm everything still passes.

## Out of scope

The freenet-test-network's lookup keyed by `ContractKey` (`src/network.rs:1093`) is in a separate repo (`freenet-test-network`) and will need its own update — not blocking for freenet-core.

## Refs

- Closes #3987 follow-up
- Depends on freenet/freenet-stdlib#70 landing and being published
- Companion freenet-core PR #3989 is the workaround being simplified

[AI-assisted - Claude]

sanity:
## Problem

`diagnostics_to_json` in `report.rs` did a destructure + manual field-by-field JSON assembly whose only purpose was to stringify `NodeDiagnosticsResponse.contract_states` map keys. That workaround (#3989) existed because the map was keyed by `ContractKey`, whose derived `Serialize` emits a struct, and `serde_json` rejects non-string JSON object keys — so a direct `serde_json::to_string(&diag)` failed with "key must be a string" the moment a node hosted a contract (#3987).

freenet-stdlib 0.8.0 (freenet-stdlib#70) fixed the type at the source: `contract_states: HashMap<String, ContractState>`. freenet-core's `main` already pins `freenet-stdlib = "0.8.1"`, so the workaround is now dead weight.

## Approach

- **Simplify the helper** to a single `serde_json::to_string_pretty(diag)` call (the whole struct serializes natively now).
- **Invert the pre-0.8 tripwire test** (`test_native_serde_json_on_contract_states_fails_documents_root_cause`, which had been parked as `#[ignore]`) into `test_native_serde_json_on_string_keyed_contract_states_succeeds`: native serialization must now *succeed* and preserve the stringified key. This keeps a real regression guard — if a future stdlib bump regresses the key type to a non-string, the test trips before the report path silently breaks again.
- Producer side (`p2p_protoc.rs`) and `fdev/diagnostics.rs` were already source-compatible with the bump; no change needed there (issue tasks 1 and 4 were satisfied by the earlier #4144 bump).

### Behavior note (intentional, benign)
The old manual assembly used `serde_json::Map` (a `BTreeMap`; `preserve_order` is not enabled), so top-level keys were emitted alphabetically. The derived serialization emits them in struct-declaration order. JSON object key order is not semantically meaningful, and the only consumer (`fdev/diagnostics.rs` / report viewer) reads fields by name on the deserialized struct, never positionally — so this is not a functional change. Field values and the contract-key encoding are unchanged.

## Testing

- `cargo fmt -p freenet -- --check` — clean
- `cargo clippy --locked -p freenet --bin freenet -- -D warnings` — clean
- `cargo test -p freenet --bin freenet commands::report` — 19 passed, 0 failed, 0 ignored (includes the three #3989 round-trip/field-coverage tests and the new inverted regression test)
- External `codex review --base main` — no actionable correctness issues

Closes #3993

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/report.rs
+++ b/crates/core/src/bin/commands/report.rs
@@ -771,60 +771,15 @@ async fn query_node_diagnostics(host: &str, port: u16) -> Result<String> {
 
 /// Serialize `NodeDiagnosticsResponse` to JSON for the report payload.
 ///
-/// `contract_states` is keyed by `ContractKey`, whose derived `Serialize`
-/// impl emits a struct (`{instance, code}`). `serde_json` rejects non-string
-/// JSON map keys, so a direct `serde_json::to_string(&diag)` fails with
-/// "key must be a string" as soon as the node hosts at least one contract.
-/// We rebuild the field with stringified keys before serializing. The
-/// rest of the struct serializes natively. See freenet/freenet-core#3987.
-///
-/// The destructure-bind on the input forces a compile error if a field is
-/// added to `NodeDiagnosticsResponse` upstream — without it, the helper
-/// would silently drop the new field from the JSON payload.
+/// Since freenet-stdlib 0.8.0 (`fix!: stringify
+/// NodeDiagnosticsResponse.contract_states keys`, freenet-stdlib#70),
+/// `contract_states` is keyed by `String`, so the whole struct serializes
+/// natively with `serde_json` — no manual key stringification is needed.
+/// This previously had to rebuild `contract_states` by hand because the map
+/// was keyed by `ContractKey`, whose derived `Serialize` emits a struct and
+/// `serde_json` rejects non-string JSON map keys (freenet/freenet-core#3987).
 fn diagnostics_to_json(diag: &NodeDiagnosticsResponse) -> Result<String> {
-    use serde_json::{Map, Value};
-
-    let NodeDiagnosticsResponse {
-        node_info,
-        network_info,
-        subscriptions,
-        contract_states,
-        system_metrics,
-        connected_peers_detailed,
-    } = diag;
-
-    let mut states = Map::with_capacity(contract_states.len());
-    for (key, state) in contract_states {
-        let value =
-            serde_json::to_value(state).context("Failed to serialize contract state value")?;
-        states.insert(key.to_string(), value);
-    }
-
-    let mut value = Map::new();
-    value.insert(
-        "node_info".to_string(),
-        serde_json::to_value(node_info).context("Failed to serialize node_info")?,
-    );
-    value.insert(
-        "network_info".to_string(),
-        serde_json::to_value(network_info).context("Failed to serialize network_info")?,
-    );
-    value.insert(
-        "subscriptions".to_string(),
-        serde_json::to_value(subscriptions).context("Failed to serialize subscriptions")?,
-    );
-    value.insert("contract_states".to_string(), Value::Object(states));
-    value.insert(
-        "system_metrics".to_string(),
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary

**Self-review (4 lenses):**
- *Code-first:* helper now delegates to `serde_json::to_string_pretty(diag)`; `.context()` (anyhow) still in scope; doc comment updated to explain why the manual assembly is gone.
- *Testing:* the three #3989 tests (empty / populated / all-fields round-trip) still exercise the helper and pass, pinning the output shape; the inverted tripwire test guards the native-serialization premise.
- *Skeptical:* confirmed `preserve_order` is NOT enabled, so the old `serde_json::Map` emitted top-level keys alphabetically while the derived path emits declaration order. Verified the sole consumer (`fdev/diagnostics.rs`) reads by field name, never positionally — so the key-order delta is benign. Field values unchanged.
- *Big-picture:* issue tasks 1 (producer `.to_string()`) and 4 (fdev) were already satisfied by the #4144 stdlib bump; this PR closes out tasks 2/3/5. Stale `#[ignore]` comment referencing `git-workflow.md` / PR #4144 removed.

**External review:** `codex review --base main` — *"I did not find any actionable correctness issues in the diff."* It independently confirmed stdlib 0.8.1's `contract_states: HashMap<String, ContractState>` and that the regression coverage is appropriate.

Not a wire-format/protocol/crypto/migration change — the on-wire `NodeDiagnosticsResponse` format is owned by stdlib and unchanged; this only changes how the report binary serializes an already-string-keyed map.

[AI-assisted - Claude]

PR #4332, opened 2026-06-02
fix: serve client API on both IPv4 and IPv6 so 127.0.0.1 works

### LINKED ISSUE #4330: bug: local WS/HTTP API binds a single IPv6 socket — literal 127.0.0.1 is refused (Windows browsers can't reach River; `freenet local` refuses IPv4)
## Problem

The node's local HTTP/WebSocket client API (port 7509) is unreachable via the literal `http://127.0.0.1:7509/` on **Windows** browsers — the browser shows *"can't access this site."* Using `http://localhost:7509/` instead works.

User-visible impact: the official River chat link (handed out by freenet.org) pointed at `http://127.0.0.1:7509/...`, so Windows users hit a dead link. Reported in [gigazine.net's River writeup](https://gigazine.net/news/20260531-freenet-river/), where the author had to hand-edit `127.0.0.1`→`localhost` to reach the room. (Worked around on the website side in freenet/web #58, which now hands out `localhost`. This issue tracks the proper fix in the node.)

There is **also a non-Windows, always-reproducible facet**: in **local mode** (`freenet local`) the API default binds to the *specific* address `::1`. A specific IPv6 address does **not** accept IPv4-mapped connections (only the `::` wildcard does), so `127.0.0.1:7509` is refused on **every OS**, Linux included. This part is unit-testable in CI.

## Root cause

The API binds a **single** socket to `config.ws_api.address`:

- Defaults (`crates/core/src/config.rs:731-737`):
  - local mode  → `default_local_address()` = `::1`  (`config.rs:1691-1694`)
  - network mode → `default_listening_address()` = `::` (`config.rs:1685-1689`)
- Bind (`crates/core/src/server.rs:110-176`): one `socket2` socket; for IPv6 it sets `set_only_v6(false)` (`server.rs:139-141`) to request dual-stack, then `bind()`.

The dual-stack `IPV6_V6ONLY=false` trick only delivers IPv4 connections to the socket when bound to the **wildcard `::`**, and even then **only reliably on Linux**:

| mode | bind addr | `127.0.0.1` reachable? |
|------|-----------|------------------------|
| local | `::1` (specific) | **No** — on any OS. IPv4-mapped acceptance doesn't apply to a specific address. |
| network | `::` (wildcard) | Linux: yes (v4-mapped). **Windows: unreliable / effectively v6-only inbound → No.** |

So on Windows the literal IPv4 loopback is refused in both modes; `localhost` works because Windows resolves it to `::1`, which the socket *is* listening on.

This is **not** a regression introduced by the dual-stack PR #3648 in the wildcard sense — network mode used the `0.0.0.0` IPv4 wildcard before too. #3648 changed the *address family* of the defaults (`0.0.0.0`→`::`, `127.0.0.1`→`::1`). The latent defect is relying on single-socket dual-stack, whose IPv4 behavior is OS-dependent (and absent entirely for the specific-address local default).

## Security analysis (checked, no internet-exposure regression)

Flagging this because "binds to `::`/`::1`" can look alarming:

- **`::1` is the IPv6 loopback — local-only**, exactly like `127.0.0.1`. It does not accept off-host connections.
- In network mode the socket binds the **wildcard `::`** (all interfaces), but the API is **not** internet-exposed: `private_network_filter` (`server.rs:588-642`, added in #3459) runs for every non-loopback bind and **default-denies any non-private source IP with `403 FORBIDDEN`** (`is_private_ip`, `server.rs:185-198`; operators can opt-in extra ranges via `--allowed-source-cidrs`). Reachable surface = loopback + LAN/private ranges only.
- The proposed fix below preserves this exactly: the filter keys off `config.address.is_loopback()` and the per-connection source IP, independent of how many sockets we bind.

## Proposed fix: dual-family listener

When the configured API address is a "both-families" address (an IPv6 wildcard/loopback, or its IPv4 counterpart), bind **two** listeners sharing the same router:

| configured | bind A | bind B |
|------------|--------|--------|
| `::` | `[::]:port` | `0.0.0.0:port` |
| `::1` | `[::1]:port` | `127.0.0.1:port` |
| (a specific non-loopback IP) | that address only | — |

Set **`IPV6_V6ONLY=true`** on the IPv6 socket so the two don't conflict over the IPv4 space (with `set_only_v6(false)`, a `::` socket on Linux already claims IPv4 and the companion `0.0.0.0` bind would fail `EADDRINUSE`). **The v6only flip and the companion bind must land together** — flipping v6only without adding the companion v4 socket would break IPv4 on Linux. `axum::Router` is `Clone`, so the same router serves both listeners; the existing security filter + Host allowlist apply per-connection on both.

Result: `127.0.0.1`, `::1`, and `localhost` all reach the node on every OS (Windows included), via real per-family sockets rather than OS-dependent v4-mapped behavior.

Implementation touch points: `serve_client_api_in_impl` (`server.rs:539-619`) builds the router/socket; `serve_with_listener` (`server.rs:110-176`) owns the `set_only_v6` call. The `pre_bound` path (socket-activation/tests) should keep using the single caller-provided listener and skip the companion.

### Alternative considered & rejected

Revert the loopback/listening defaults to IPv4 (`127.0.0.1`/`0.0.0.0`). Makes `127.0.0.1` work but breaks `localhost`→`::1` on Windows and drops IPv6 reachability — trades one platform's broken link for another's. Dual-family listener is the only option that makes *both* literals work everywhere.

## Regression test

Local-mode facet is CI-testable on Linux:

1. Bind the API to the local-mode default (`::1`).
2. Assert a client can connect to **both** `127.0.0.1:port` **and** `[::1]:port`.
3. Without the fix, the `127.0.0.1` connection is refused (specific `::1` bind) → test fails. With the dual-family bind → passes.

(The Windows network-mode v4-mapped unreliability can't be reproduced in Linux CI; the local-mode test pins the structural fix — two real per-family sockets — which is what also resolves the Windows case.)

## Risk

High-risk surface (socket binding). Needs Full multi-model review and ideally a manual Windows smoke test (`freenet network`, open `http://127.0.0.1:7509/` in Edge/Chrome). The `pre_bound` path, port-in-use error messaging, and the `set_only_v6` flip are the sharp edges.

cc @iduartgomez (networking)

[AI-assisted - Claude]

sanity:
## Problem

The local HTTP/WebSocket client API (port 7509) is unreachable via the literal `http://127.0.0.1:7509/` on **Windows** — the browser shows *"can't access this site."* `http://localhost:7509/` works. Reported in [gigazine.net's River writeup](https://gigazine.net/news/20260531-freenet-river/); the author had to hand-edit `127.0.0.1`→`localhost` to reach the official room. (Worked around website-side in freenet/web #58.)

There is also an **always-reproducible** facet: in **local mode** (`freenet local`) the API binds the *specific* address `::1`. A specific IPv6 address never accepts IPv4-mapped connections (only the `::` wildcard does), so `127.0.0.1:7509` is refused on **every OS**, Linux included.

Root cause: the API binds a **single** socket (`server.rs::serve_with_listener`), relying on `IPV6_V6ONLY=false` dual-stack:

| mode | bind addr | `127.0.0.1` reachable? |
|------|-----------|------------------------|
| local | `::1` (specific) | **No** — any OS. IPv4-mapped acceptance doesn't apply to a specific address. |
| network | `::` (wildcard) | Linux: yes (v4-mapped). **Windows: unreliable → No.** |

## Solution

Key insight: single-socket dual-stack is not portable. IPv4-mapped acceptance is wildcard-only and unreliable on Windows even for `::`. So bind **two explicit sockets** — one per family — for wildcard/loopback addresses:

| configured | primary | companion |
|------------|---------|-----------|
| `::` | `[::]:port` | `0.0.0.0:port` |
| `::1` | `[::1]:port` | `127.0.0.1:port` |
| specific interface IP | that address only | — |

- `serve_with_listener` now binds IPv6 sockets with **`IPV6_V6ONLY=true`** so the two sockets don't contend for the IPv4 space (a v4-mapped `::` socket makes a companion `0.0.0.0` bind fail `AddrInUse` on Linux). The v6-only flip and the companion bind land together — flipping v6-only without the companion would break IPv4.
- New `companion_bind_addr()` maps an address to its other-family counterpart (loopback↔loopback, wildcard↔wildcard); a specific interface IP gets no companion.
- New `serve_dual_stack()` binds primary + best-effort companion sharing one cloned `axum::Router`. Best-effort = a single-stack host (IPv4 or IPv6 disabled) keeps serving on the primary rather than failing startup; the primary bind error stays fatal.
- The **`pre_bound`** caller-owned-listener path (test helpers / socket activation) is untouched — no companion.

**No exposure change.** The `private_network_filter` (default-deny non-private source IPs, #3459) and the Host-header allowlist key off `config.address` and the per-connection source IP, so they apply identically to both sockets. `::1` is loopback-only; `::` (network mode) stays LAN-gated, not internet-exposed.

### Alternative considered & rejected

Revert the defaults to IPv4 (`127.0.0.1`/`0.0.0.0`): makes `127.0.0.1` work but breaks `localhost`→`::1` on Windows and drops IPv6. Dual-family is the only option that makes both literals work everywhere.

## Testing

- `companion_bind_addr_maps_loopback_and_wildcard_across_families` — unit test of the mapping (loopback, wildcard, specific-IP → None).
- `client_api_reachable_on_both_ipv4_and_ipv6_loopback` — binds the `::1` default via `serve_dual_stack` and asserts a TCP connection succeeds on **both** `127.0.0.1:port` and `[::1]:port`. **Verified it fails with `ConnectionRefused` on `127.0.0.1` without the companion bind** (reproduces #4330), passes with it.
- `cargo clippy --locked -- -D warnings` clean; `cargo fmt --check` clean; existing `server::tests` (25) pass.

⚠️ The Windows-specific `::` v4-mapped unreliability can't be reproduced in Linux CI; the local-mode `::1` test pins the structural fix (two real per-family sockets), which is what also resolves the Windows case. A manual Windows smoke test (`freenet network`, open `http://127.0.0.1:7509/` in Edge/Chrome) is worth doing before release.

## Fixes

Fixes #4330

cc @iduartgomez — this touches socket binding (high-risk surface); would value your review on the `IPV6_V6ONLY` flip + best-effort companion approach.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server.rs
+++ b/crates/core/src/server.rs
@@ -13,7 +13,7 @@ mod home_page;
 pub(crate) mod path_handlers;
 
 use std::collections::HashSet;
-use std::net::{IpAddr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::time::Instant;
@@ -103,8 +103,31 @@ pub(crate) enum HostCallbackResult {
     },
 }
 
-async fn serve(socket: SocketAddr, router: axum::Router) -> std::io::Result<()> {
-    serve_with_listener(socket, router, None).await
+/// For a wildcard or loopback bind address, returns the equivalent address in
+/// the other IP family so the client API can be served on a second, explicit
+/// socket and thus be reachable over both IPv4 and IPv6.
+///
+/// This is what makes `http://127.0.0.1:7509/` AND `http://[::1]:7509/` (and
+/// `localhost`, which resolves to either) work on every OS. A single IPv6
+/// socket is not enough: IPv4-mapped acceptance (`IPV6_V6ONLY=false`) applies
+/// only to the `::` wildcard — never to a specific address like `::1` — and is
+/// unreliable on Windows even for `::`, so the literal IPv4 loopback is refused
+/// there. See issue #4330.
+///
+/// Returns `None` for any specific (non-wildcard, non-loopback) address: an
+/// operator who binds the API to a particular interface IP gets exactly that
+/// one socket, with no companion in the other family.
+fn companion_bind_addr(addr: IpAddr) -> Option<IpAddr> {
+    match addr {
+        IpAddr::V6(v6) if v6 == Ipv6Addr::UNSPECIFIED => Some(IpAddr::V4(Ipv4Addr::UNSPECIFIED)),
+        IpAddr::V6(v6) if v6 == Ipv6Addr::LOCALHOST => Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
+        IpAddr::V4(v4) if v4 == Ipv4Addr::UNSPECIFIED => Some(IpAddr::V6(Ipv6Addr::UNSPECIFIED)),
+        IpAddr::V4(v4) if v4 == Ipv4Addr::LOCALHOST => Some(IpAddr::V6(Ipv6Addr::LOCALHOST)),
+        // A specific (non-wildcard, non-loopback) interface address is bound
+        // as-is, with no companion. Enumerate variants rather than `_` to
+        // satisfy `clippy::wildcard_enum_match_arm`.
+        IpAddr::V4(_) | IpAddr::V6(_) => None,
+    }
 }
 
 async fn serve_with_listener(
@@ -134,10 +157,19 @@ async fn serve_with_listener(
                 .map_err(|e| {
                     std::io::Error::new(e.kind(), format!("Failed to create socket: {e}"))
                 })?;
-                // Enable dual-stack: accept both IPv4 and IPv6 on a single socket.
-                // IPv4 clients connect via IPv4-mapped addresses (::ffff:x.x.x.x).
+                // Bind IPv6 sockets v6-only. We do NOT rely on single-socket
+                // dual-stack (IPV6_V6ONLY=false / IPv4-mapped addresses) for the
+                // client API: that only delivers IPv4 to a socket bound to the
+                // `::` wildcard (never to a specific address like `::1`), and is
+                // unreliable on Windows even for `::`. Instead, `serve_client_api`
+                // binds an explicit companion socket in the other family (see
+                // `companion_bind_addr`) so literal `127.0.0.1` AND `::1` /
+                // `localhost` are reachable on every OS. v6-only here keeps the
+                // two sockets from contending for the IPv4 space (a v4-mapped
+                // `::` socket would make the companion `0.0.0.0` bind fail with
+                // AddrInUse on Linux).
                 if is_ipv6 {
-                    sock.set_only_v6(false)?;
+                    sock.set_only_v6(true)?;
                 }
                 sock.set_reuse_address(true)?;
                 sock.set_nonblocking(true)?;
@@ -217,7 +249,7 @@ pub mod local_node {
         contract::{Executor, ExecutorError},
     };
 
-    use super::{client_api::HttpClientApi, serve};
+    use super::{client_api::HttpClientApi, serve_dual_stack};
 
     pub async fn run_local_node(mut executor: Executor, socket: SocketAddr) -> anyhow::Result<()> {
         if !super::is_private_ip(&socket.ip()) {
@@ -229,7 +261,11 @@ pub mod local_node {
         let (mut gw, gw_router) = HttpClientApi::as_router(&socket);
         let (mut ws_proxy, ws_router) = WebSocketProxy::create_router(gw_router);
 
-        serve(socket, ws_router.layer(TraceLayer::new_for_http())).await?;
+        // Route through serve_dual_stack (not the bare single-socket bind) so
+        // this path also binds an IPv4+IPv6 companion pair for loopback/wildcard
+        // addresses — keeping every client-API serve path reachable over both
+        // families (#4330).
+        serve_dual_stack(socket, ws_router.layer(TraceLayer::new_for_http()), None).await?;
 
         // TODO: use combinator instead
         // let mut all_clients =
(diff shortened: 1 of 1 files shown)


sanity:
## Multi-model review summary (high-risk surface: socket binding)

Ran a full review pass: Claude **code-first** + **skeptical** reviewers, plus an external **Gemini** pass (Codex was rate-limited). All three agree the core design is sound — `set_only_v6(true)` + explicit per-family companion sockets is correct, the `AddrInUse` contention reasoning checks out (skeptical reviewer reproduced it empirically on Linux), the security posture is unchanged (LAN filter + Host allowlist key off `config.address`/source-IP and apply to both sockets), the UDP P2P transport socket is untouched, and `pre_bound` correctly skips the companion.

### Findings addressed in the latest push (`1ec088ba`)

1. **Legacy `server::local_node::run_local_node` bypassed the fix** (flagged by both Claude reviewers). It called the bare single-socket `serve()`, so with the `set_only_v6(true)` flip it would have skipped the companion bind. It's dead in-repo (the live local path is `node::run_local_node` → `serve_client_api_in` → `serve_dual_stack`) but `pub`. **Fixed**: routed it through `serve_dual_stack` and removed the now-unused `serve()` wrapper, so *every* client-API serve path goes through one dual-stack entry point.

2. **Regression test could false-pass** (skeptical reviewer L1): the probe reserved the port only on `::1`, so a foreign process on `127.0.0.1:port` could be connected to instead. **Fixed**: the test now does an HTTP GET over both families and asserts our handler's distinctive marker body, so only *our* companion socket can satisfy it.

### Not changed (with rationale)

- **IPv6-disabled host hard-fails in network mode** (skeptical L2): the `::` primary bind is fatal, so a host with IPv6 disabled won't start even though IPv4 is available. **Not a regression** — the old single `::` socket failed identically. A symmetric "fall back to the other family if the primary's family is unavailable" is a reasonable follow-up but out of scope here.
- **Companion `axum::serve` JoinHandle dropped** (skeptical L4): mirrors the pre-existing primary spawn; changing task-handle tracking would touch the existing path too (separate concern).
- **`127.0.0.0/8` / `::ffff:` not mapped** (Gemini, skeptical L3): intentional — only the canonical loopback/wildcard get a companion; a specific interface IP is bound as-is.

### Verification
`cargo clippy --locked -- -D warnings` clean, `cargo fmt --check` clean, `server::tests` (25) pass. Regression test verified to fail with `ConnectionRefused` on `127.0.0.1` when the companion bind is removed.

### ⚠️ Before merge
This touches socket binding — @iduartgomez, would value your eyes on the `IPV6_V6ONLY` flip + best-effort companion. The Linux CI test pins the structural fix, but the Windows-specific `::` v4-mapped unreliability can't be reproduced in CI; a manual Windows smoke test (`freenet network`, open `http://127.0.0.1:7509/` in Edge/Chrome) is worth doing before release.

[AI-assisted - Claude]

iduartgomez (approved):

iduartgomez:
I don't have access myself to a windows machine to test this but LGTM, this should be backwards compatible so the risk of breaking is low

(confirm @claude)

iduartgomez:
@claude confirm https://github.com/freenet/freenet-core/pull/4332#issuecomment-4630035528

PR #4333, opened 2026-06-02
test(contract): add e2e UpdateQuery dispatch tests for compound variants

### LINKED ISSUE #4005: Add end-to-end UpdateQuery dispatch tests for RelatedStateAndDelta + StateAndDelta + rejection arms
Follow-up to PR #4004 (https://github.com/freenet/freenet-core/pull/4004#issuecomment-4363709525).

The PR landed unit tests for the `inject_related_state` helper in isolation and an integration test for the runtime.rs initial-state install notification path, but the rule-review bot flagged two remaining gaps in dispatch-level coverage of the `UpdateQuery` arms rewritten in `crates/core/src/contract.rs:1049-1135`:

## Gaps

- **`RelatedStateAndDelta` end-to-end** — no test sends `UpdateData::RelatedStateAndDelta` through the full `UpdateQuery` handler via the `contract_handler` channel. The unit tests would not catch a regression that reintroduced `unreachable!()` in the dispatch arm itself (vs the helper). Closing this needs a test that mirrors the existing `send_queue_full_response_update_query` pattern but exercises the conversion path rather than the rejection path.

- **`StateAndDelta` end-to-end** — same gap. The commit title explicitly names this variant as one of the fixed `unreachable!()` arms, but no test verifies that the variant survives dispatch and that the full state is preferred over the delta. Reverting this arm wouldn't fail the current test suite.

- **Rejection arms (`RelatedState` / `RelatedDelta` / unknown `_`)** — these used to panic `unreachable!()` and now return a structured `UpdateResponse { new_value: Err, state_changed: false }`. Worth covering as edge-case hardening; lower priority than the two above since they're not the load-bearing dispatch path.

## Suggested approach

Add three new tests to the existing `mod tests` block in `crates/core/src/contract.rs`. Use the `setup_rejected_event` / `send_queue_full_response_update_query` helpers as templates — they already wire the `ContractHandlerChannel`. Each test should drive the dispatch through to the `update_value` conversion (or rejection) rather than going all the way through `upsert_contract_state` (the latter requires a runtime, which is what the existing initial-state install test covers).

The `inject_related_state` helper is itself unit-tested in PR #4004, so the new tests are about wiring (does the dispatch arm route to the helper at all?) rather than the helper's correctness.

## Why deferred

PR #4004 already addresses the primary bug (cross-node mail delivery) and the rule-review's first round of feedback (helper extraction, debug_assert, initial-state notification regression test). The two remaining warnings call for the same kind of test infrastructure (channel-based dispatch tests) that's already used for the queue-full path, but the work is mechanical rather than load-bearing for the bug fix. Filing here so we can land the fix and follow up under a clean test-coverage issue.

sanity:
## Problem

PR #4004 rewrote the `ContractHandlerEvent::UpdateQuery` dispatch arms in `crates/core/src/contract.rs` so that `RelatedStateAndDelta`, `StateAndDelta`, and the `RelatedState` / `RelatedDelta` rejection paths no longer panic with `unreachable!()`. It added unit tests for the `inject_related_state` helper in isolation plus an executor-level install-notification test, but left a coverage gap at the **dispatch level**: no test sends these `UpdateData` variants through the full `handle_contract_event` handler over the `ContractHandlerChannel`. A revert of the `StateAndDelta` or `RelatedStateAndDelta` arms (reintroducing `unreachable!()`) would not fail the existing suite.

## Approach

Add three end-to-end tests to the existing `mod tests` block in `contract.rs`. Each drives an `UpdateData` variant through the real `handle_contract_event` dispatch over a `ContractHandlerChannel` backed by a `MemoryContractHandler` (mock executor), then asserts on the `UpdateResponse` the sender observes via the channel round-trip — mirroring the production `contract_handling` driver loop.

- **`StateAndDelta`** — seeds a contract, then sends `StateAndDelta { state, delta }` with a hash-winning full state and a distinct delta; asserts the stored result equals the full `state` (proving the full-state path, not `seed ++ delta`).
- **`RelatedStateAndDelta`** — asserts the arm routes through `inject_related_state` and applies the **delta** to the main contract (stored state ends with the delta bytes), confirming it reaches the executor rather than panicking.
- **`RelatedState` / `RelatedDelta`** — asserts the structured rejection (`UpdateResponse { new_value: Err, state_changed: false }`) flows back through the channel.

Hash-ordering fragility in the mock's merge is handled by deterministically selecting input bytes that win the merge (no sleeps, no live network). The unknown-`_` arm is not directly testable (`UpdateData` is `#[non_exhaustive]` only outside its defining crate; within freenet-core every variant is matched), so it remains as documented defensive handling.

## Testing

- All three new tests pass; the full `contract::tests` module (14 tests) is green.
- Verified the regression direction: temporarily reverting each dispatch arm to `unreachable!()` makes the corresponding test panic/fail; restoring the fix makes them pass.
- `cargo fmt` clean; `cargo clippy --locked -- -D warnings` clean.
- Diff is purely additive (test code only); production dispatch code is untouched.

External review: `codex review --base main` found no actionable correctness issues.

Closes #4005

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -1961,4 +1961,332 @@ mod tests {
             "fire-and-forget rejection should produce an error"
         );
     }
+
+    // ======================================================================
+    // End-to-end UpdateQuery dispatch tests (issue #4005)
+    // ======================================================================
+    //
+    // These drive an `UpdateData` variant through the *full* dispatch arm in
+    // `handle_contract_event` via the `ContractHandlerChannel`, then assert on
+    // the `UpdateResponse` that flows back to the sender. Unlike the
+    // `inject_related_state_*` unit tests (which call the helper in isolation)
+    // and `send_queue_full_response_update_query` (which only exercises the
+    // rejection path), these tests would catch a regression that reintroduced
+    // `unreachable!()` in the dispatch arm itself — the exact failure mode
+    // PR #4004 fixed. A revert of the `StateAndDelta` or `RelatedStateAndDelta`
+    // conversion arms now fails CI here.
+
+    /// Drive a single `UpdateQuery` event end-to-end through
+    /// `handle_contract_event` over a real `ContractHandlerChannel` backed by a
+    /// `MemoryContractHandler` (mock executor), returning the response the
+    /// sender observes.
+    ///
+    /// Mirrors the production driver loop (`contract_handling`): the sender's
+    /// `send_to_handler` blocks awaiting the response; the handler side
+    /// `recv_from_sender`s to register the waiting-response slot and obtain the
+    /// `EventId`, then `handle_contract_event` runs the dispatch arm and
+    /// fulfils the sender's oneshot via `send_to_sender`.
+    async fn dispatch_update_query_e2e(
+        handler: &mut MemoryContractHandler,
+        send_halve: &handler::ContractHandlerChannel<handler::SenderHalve>,
+        key: ContractKey,
+        data: UpdateData<'static>,
+        related_contracts: RelatedContracts<'static>,
+    ) -> ContractHandlerEvent {
+        let event = ContractHandlerEvent::UpdateQuery {
+            key,
+            data,
+            related_contracts,
+        };
+        // Run the sender concurrently with the handler-side dispatch via
+        // `tokio::join!` (the handler is borrowed, so it cannot move into a
+        // spawned task). `send_to_handler` awaits the oneshot until the
+        // dispatch arm replies via `send_to_sender`; the dispatch never blocks,
+        // so the two futures make progress cooperatively to completion.
+        let send_fut = send_halve.send_to_handler(event);
+        let recv_fut = async {
+            let (id, received) = handler
+                .channel()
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary

**Self-review** (code-first / testing / skeptical / big-picture):
- Diff is purely additive — 327 lines of test code in the existing `mod tests` block; production dispatch in `handle_contract_event` is byte-for-byte unchanged (confirmed via `git diff --stat`).
- Determinism: no sleeps, no live network, no wall-clock/`thread_rng`. The mock executor's "larger blake3 hash wins" merge fragility is neutralized by self-checking input selection (the tests pick bytes that deterministically win).
- Regression direction verified empirically: temporarily reverting each of the three dispatch arms to `unreachable!()` makes the matching test fail; restoring the fix makes all three pass.

**External (Codex, `codex review --base main`):** no actionable correctness issues. It confirmed the `UpdateData` variant list against stdlib 0.8.1 and noted its sandboxed `--offline` test run failed only due to no network access (the tests pass locally).

[AI-assisted - Claude]

PR #4272, opened 2026-05-26
fix(log): demote three per-event INFO sites to DEBUG (#4251 follow-up)

sanity:
## Problem

PR #4252 capped disk usage and ripped out the worst error-stream offenders, but the **INFO firehose** on the info stream was left untouched. Measured on a River-subscribed peer (`technic`, freshly restarted on v0.2.65):

- Info-log rate: **~37 lines/sec** (≈24 MB/hr ≈576 MB/day ≈17 GB/month)
- Of which **~87%** came from just three sites:

| Site | Pre-fix share | Trigger |
|---|---|---|
| `executor/runtime.rs` `"Contract state updated"` | 44% (5,365 lines / 6 min) | every successful state write |
| `operations/update.rs` `"UPDATE_PROPAGATION"` broadcast branch | 43% (5,257 lines / 6 min) | per fan-out per UPDATE |
| `transport/rolling_rtt_stats.rs` `"shadow_rtt_aggregate"` | 1 Hz heartbeat (~326 lines/6 min) | periodic timer |

Two of these fire once per River chat message propagation; the third is a per-second heartbeat whose only consumer was the file log (the OTLP path is a separate `send_standalone_event` call). Reasonable budget for a background daemon's INFO stream is **< 1 MB/hr** — current state is 24×.

PR #4252's "Not in scope" section explicitly punted the aggregated/INFO-cadence work to a follow-up. This is that follow-up.

## Approach

Three one-line demotions, all INFO → DEBUG:

- **`contract/executor/runtime.rs:2145`** — `"Contract state updated"`. Per-event state-write notice; no operator-actionable signal at this granularity. The `error!` branch a few lines down keeps any failed update visible at the default level.
- **`operations/update.rs:293`** — `"UPDATE_PROPAGATION"` (the populated-targets branch). Pairs naturally with the `NO_TARGETS` else-branch right below it that's already DEBUG (with the explanatory comment at line 309-317 noting the per-attempt detail belongs in metrics, not per-event logs). The streak-suppressed WARN in `p2p_protoc.rs` ("BroadcastTo: no targets after N retries") still surfaces stuck-propagation outcomes at default level.
- **`transport/rolling_rtt_stats.rs:351`** — `"shadow_rtt_aggregate"`. **The OTLP `send_standalone_event` call directly below it is untouched**, so the telemetry dashboard keeps its 1 Hz aggregate feed. Only the local file-log mirror — which was pure noise — is downgraded.

[...]

Issue #4282, opened 2026-05-27
flaky: test_get_routing_coverage_low_htl fails intermittently after nextest retry-3


sanity:
## Symptom

\`crates/core/tests/simulation_integration.rs::test_get_routing_coverage_low_htl\` fails intermittently in CI even after nextest's 3-retry policy.

Observed on PR #4270 CI run [26485219159](https://github.com/freenet/freenet-core/actions/runs/26485219159/job/77991172682) (commit 8f3cc0f9):

\`\`\`
TRY 3 FAIL [16.976s] (138/140) freenet::simulation_integration test_get_routing_coverage_low_htl
thread 'test_get_routing_coverage_low_htl' (51857) panicked at crates/core/tests/simulation_integration.rs:7559:5:
GET routing exhaustion: 1 nodes failed to get contract state (contract only cached at ~5 nodes due to HTL=4). Failed nodes: ["node-1"]. See #3431.
\`\`\`

The test references closed issue #3431 (\"fix: GET/SUBSCRIBE routing fails when contract is only cached at a few nodes (low HTL)\"). The panic message reads as a regression guard for that closed bug, but the failure is intermittent — 1 of N nodes fails to get the contract.

## Why this matters

Per \`.claude/rules/flaky-tests.md\` (and the user's global preference): *\"A flaky test is not 'just flaky' — it indicates a real bug in either the test or the code under test. Treat it as a blocker.\"*

Pre-existing on main (not introduced by #4270), but blocks PR merges every time it trips.

## Suggested investigation

1. Reproduce locally with a fixed RNG seed (\`SEED=1 cargo test test_get_routing_coverage_low_htl\` or equivalent).
2. Add nextest \`--retries 0\` to see the raw failure rate.
3. Examine \`crates/core/tests/simulation_integration.rs:7559\` for the assertion shape — is it pinning that 100 % of nodes get the contract, even at HTL=4 with sparse caching? If so, the assertion may be tighter than the underlying invariant supports.
4. If the test is correct but the routing genuinely can fail at low HTL: file a new bug separate from the closed #3431, and either widen the test's tolerance or document the failure mode.

## Possible workarounds (none are great)

- Add to the test's allowed-failure list (defeats the purpose).
- \`#[ignore]\` with a tracking link to this issue (would unblock PRs but loses coverage).
- Lower the assertion's success-rate floor (only if the floor is genuinely too tight, not just to hide the flake).

[AI-assisted - Claude]

[...]

PR #4334, opened 2026-06-02
feat(log): add periodic UPDATE-propagation summary emitter (#4281)

### LINKED ISSUE #4281: Add aggregated periodic INFO emitter for UPDATE propagation (restore operator visibility after #4272)
## Background

PR #4252 mitigated the River-driven log-spam disk-fill by demoting / rate-limiting the worst per-event log sites. Its \"Not in scope\" section explicitly deferred this work:

> **Aggregated periodic INFO backpressure log** (e.g. \"rejected N events for contract X in last Ms\") — worthwhile follow-up to restore operator visibility lost by the per-event demotions, but out of scope for this disk-fill mitigation.

PR #4272 followed up on the per-event INFO firehose, demoting three sites (\`Contract state updated\`, \`UPDATE_PROPAGATION\` broadcast, \`shadow_rtt_aggregate\` local mirror) to DEBUG and adding pin tests. At default INFO log level, operators now have **no per-contract signal** at all about UPDATE propagation activity — only failure-path messages (\`error!\` for notification failures, streak-suppressed \`BROADCAST_NO_TARGETS\` WARN). For a healthy node carrying River traffic, the log goes silent.

## Ask

Add a periodic INFO-level aggregated emitter that summarises UPDATE propagation activity without re-introducing per-event volume. Strawman shape:

\`\`\`
INFO update_propagation_summary contracts=N updates=M targets_avg=K window_s=60 phase=\"summary\"
\`\`\`

Open design questions worth thinking through up front:

1. **Per-contract or aggregated?** Per-contract gives a debuggable signal (\"contract X is propagating, contract Y is silent\") at the cost of one INFO line per active contract per window. Aggregated is bounded but loses per-contract diagnosability. Probably want per-contract with a cap (top-K by activity, plus a roll-up \"…and N others\").
2. **Window length?** 60s is the natural starting point. Shorter window = better debugging granularity; longer = lower log rate.
3. **Where does the counter live?** Probably \`OpManager::ring\` — it already tracks per-contract activity for the dashboard's \`record_contract_update\` path. A periodic background task can drain and emit. Alternative: piggyback on the existing 1Hz \`shadow_rtt_aggregate\` cadence to share the timer.
4. **What about NO_TARGETS / failure counts?** Worth surfacing in the summary too. The existing \`interest_resolve_failed\` counter at \`update.rs::get_broadcast_targets_update\` is a good source.
5. **Budget.** Whatever shape this takes, the cumulative INFO log rate from this emitter should stay well under 1 line/sec sustained, so it doesn't reintroduce the spam class PR #4272 just killed.

## Why bother

Right now, an operator looking at a healthy node's INFO log for \"is this thing actually propagating River updates?\" sees nothing. Failure-mode visibility is preserved, but liveness visibility is gone. That's a regression from operator-experience perspective even though it's a win from disk-fill perspective.

A bounded periodic summary is the canonical fix and was the deferred follow-up named by both #4252 and #4272.

## Out of scope

- Re-promoting any of the three sites demoted in #4272 — the per-event firehose is exactly what we're avoiding.
- Replacing the existing OTLP \`send_standalone_event\` path for \`shadow_rtt_aggregate\` — that lives independently and the dashboard's 1 Hz feed is unaffected by the log-level change.

[AI-assisted - Claude]

sanity:
## Problem

PRs #4252 and #4272 demoted the three high-volume per-event UPDATE log sites (`Contract state updated`, the `UPDATE_PROPAGATION` broadcast line, and the `shadow_rtt_aggregate` local mirror) from INFO to DEBUG to stop a River-driven log-spam disk-fill. The fix worked, but at the default `RUST_LOG=info` level it left operators with **no liveness signal** about whether a node is propagating contract updates at all — only failure-path messages survive at INFO/WARN, so a healthy node's log goes silent. That is the operator-visibility regression #4281 asks us to close.

## Solution

A bounded, self-contained periodic aggregator (`operations/update/propagation_stats.rs`) owned by `OpManager`:

- The broadcast fan-out path (`handle_broadcast_state_change`) increments cheap per-contract atomic counters: updates, total resolved targets, no-target attempts, and interest-resolve failures. Recording is gated on the **initial** fan-out, not retry re-emissions, so `updates` isn't inflated ~4x on stuck contracts (the exact amplification #4272 set out to avoid).
- A background task drains-and-emits once per **60s window**: one INFO `update_propagation_summary` line (`contracts`, `updates`, `targets_avg`, `no_targets`, `interest_resolve_failed`, `window_s`) plus up to `TOP_K=8` per-contract detail lines and a single rollup line for the rest. Worst case is `TOP_K + 1` lines per window (**< 0.2 lines/sec sustained**) — well inside #4272's budget; a zero-activity window emits nothing.
- Counters are windowed (atomic `swap` snapshot-and-reset); the tracking map is size-capped (`MAX_TRACKED_CONTRACTS=4096`, per the per-key-collection rule) so a key churn can't grow it unboundedly, and idle entries are evicted between bursts.

This matches the issue's preferred "per-contract with a top-K cap plus rollup" shape and its named data sources. The OTLP `send_standalone_event` / dashboard paths are untouched, and none of the #4272 DEBUG demotions are reverted. UPDATE behavior is unchanged — this is purely additive telemetry.

## Testing

New unit tests in `propagation_stats.rs`:
- Pure aggregation logic: empty window → `None`, zero-update contracts filtered out, totals/averages, top-K cap + `others` rollup, deterministic id tie-break.
- `record_broadcast` → `drain_window` accumulate-then-reset cycle and idle-entry eviction.
- Tracking-cap enforcement (new contract dropped at capacity).
- `drain_and_log` end-to-end: emits an INFO summary on an active window, silent on idle.
- Source-level **pin test** asserting the emitter stays at INFO — a future demotion re-opens the #4281 gap and fails the test (mirrors the #4272 pin tests).

Existing update-module and op_state_manager tests pass. `cargo fmt`, `cargo clippy --locked -- -D warnings` (the CI invocation) clean.

Closes #4281

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/mock_runtime.rs
+++ b/crates/core/src/contract/executor/mock_runtime.rs
@@ -259,6 +259,7 @@ where
             if let Err(err) = op_manager.try_notify_node_event(NodeEvent::BroadcastStateChange {
                 key,
                 new_state: state.clone(),
+                is_retry: false,
             }) {
                 // Best-effort by design (mirrors the production
                 // executor; see runtime.rs for the #4145 / #4238
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -2442,6 +2442,7 @@ where
                 op_manager.try_notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
                     key: *key,
                     new_state: new_state.clone(),
+                    is_retry: false,
                 })
             {
                 // Non-blocking emit: a 30-second `notify_node_event(...).await`
@@ -2805,6 +2806,7 @@ where
                 op_manager.try_notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
                     key,
                     new_state,
+                    is_retry: false,
                 })
             {
                 // Best-effort by design — see #4145 and the sibling
--- a/crates/core/src/message.rs
+++ b/crates/core/src/message.rs
@@ -694,6 +694,13 @@ pub(crate) enum NodeEvent {
     BroadcastStateChange {
         key: ContractKey,
         new_state: WrappedState,
+        /// `false` for a fresh broadcast emitted by the executor on a local
+        /// state change; `true` for a no-target retry re-emission scheduled by
+        /// `handle_broadcast_state_change`. Lets the handler count each fresh
+        /// logical broadcast once for the #4281 propagation summary without
+        /// re-counting (or being confused by) retries that share the
+        /// per-contract `broadcast_retries` state.
+        is_retry: bool,
     },
     /// Send state to a specific peer that reported a stale summary.
     /// Unlike BroadcastStateChange (which fans out to ALL subscribers),
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -2154,9 +2154,18 @@ impl P2pConnManager {
                                 graceful_shutdown = true;
                                 break;
                             }
-                            NodeEvent::BroadcastStateChange { key, new_state } => {
-                                ctx.handle_broadcast_state_change(&op_manager, key, new_state)
-                                    .await;
+                            NodeEvent::BroadcastStateChange {
+                                key,
+                                new_state,
+                                is_retry,
+                            } => {
+                                ctx.handle_broadcast_state_change(
+                                    &op_manager,
+                                    key,
+                                    new_state,
+                                    is_retry,
+                                )
+                                .await;
                             }
                             NodeEvent::SyncStateToPeer {
                                 key,
@@ -4219,6 +4228,10 @@ impl P2pConnManager {
         op_manager: &Arc<OpManager>,
         key: freenet_stdlib::prelude::ContractKey,
         new_state: freenet_stdlib::prelude::WrappedState,
+        // `false` for a fresh executor-emitted broadcast, `true` for a
+        // no-target retry re-emission this handler scheduled. Used only to
+        // count fresh logical broadcasts once for the #4281 summary.
+        is_retry: bool,
     ) {
         tracing::debug!(
             contract = %key,
@@ -4243,6 +4256,25 @@ impl P2pConnManager {
         );
 
         if target_result.targets.is_empty() {
+            // Record the NO_TARGETS outcome once per *fresh* no-target broadcast
+            // for the #4281 summary, but NOT for retry re-emissions. Gating on
+            // `is_retry` (carried on the event) rather than the per-contract
+            // `broadcast_retries` counter is what makes the count accurate:
+            // retries and fresh broadcasts share that counter, so a fresh
+            // broadcast arriving mid-retry-cycle would otherwise be absorbed
+            // and uncounted (Codex P2). With the explicit flag, every fresh
+            // no-target broadcast is counted exactly once and retries never
+            // double-count. If a later retry heals (finds targets), the
+            // targets-found path below additionally records that success — both
+            // the initial miss and the eventual propagation are surfaced.
+            if !is_retry {
+                op_manager.update_propagation_stats.record_broadcast(
+                    *key.id(),
+                    0,
+                    target_result.interest_resolve_failed,
+                );
+            }
+
             let retry_count = self.broadcast_retries.entry(key).or_insert(0);
             if *retry_count < Self::MAX_BROADCAST_RETRIES {
                 *retry_count += 1;
@@ -4279,6 +4311,7 @@ impl P2pConnManager {
                         .notify_node_event(crate::message::NodeEvent::BroadcastStateChange {
                             key,
                             new_state,
+                            is_retry: true,
                         })
                         .await
                     {
@@ -4342,6 +4375,11 @@ impl P2pConnManager {
                             .await;
                     }
                 }
+                // The NO_TARGETS outcome was already recorded for the #4281
+                // summary at the start of this retry cycle (retry_count == 0
+                // above); retry exhaustion does not record again, so a stuck
+                // contract counts as one no_targets per fresh broadcast, not
+                // once per retry.
             }
             return;
         }
@@ -4349,6 +4387,18 @@ impl P2pConnManager {
         // Targets found - clear any pending retry and streak state for this contract
         self.broadcast_retries.remove(&key);
         self.broadcast_no_target_streak.remove(&key);
+
+        // Record a successful fan-out for the #4281 propagation summary. Fires
+        // whether targets were found on the initial attempt or on a healing
+        // retry, so a recovered in-flight-subscription race is surfaced as a
+        // propagated update (with its real target count). A fresh broadcast
+        // that found targets on the first try has no `broadcast_retries` entry,
+        // so it is recorded exactly once here.
+        op_manager.update_propagation_stats.record_broadcast(
+            *key.id(),
+            target_result.targets.len(),
+            target_result.interest_resolve_failed,
+        );
 
         // In production, enqueue each target into the broadcast queue.
         // The queue worker handles delta computation and streaming with bounded
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -100,6 +100,13 @@ pub(crate) struct OpManager {
     pub interest_manager: Arc<crate::ring::interest::InterestManager<InstantTimeSrc>>,
     /// Dedup cache for skipping redundant broadcast WASM merges
     pub broadcast_dedup_cache: Arc<crate::operations::update::BroadcastDedupCache>,
+    /// Bounded per-contract UPDATE-propagation counters. Fed from the
+    /// broadcast fan-out path and drained by a periodic background task that
+    /// emits an INFO `update_propagation_summary` line per window. Restores
+    /// the operator liveness signal lost when #4272 demoted the per-event
+    /// UPDATE log sites to DEBUG (issue #4281).
+    pub(crate) update_propagation_stats:
+        Arc<crate::operations::update::propagation_stats::UpdatePropagationStats>,
     /// Request router for client request deduplication.
     ///
     /// This is initialized lazily from `client_event_handling` because the router is only
@@ -213,6 +220,7 @@ impl Clone for OpManager {
             neighbor_hosting: self.neighbor_hosting.clone(),
             interest_manager: self.interest_manager.clone(),
             broadcast_dedup_cache: self.broadcast_dedup_cache.clone(),
+            update_propagation_stats: self.update_propagation_stats.clone(),
             request_router: self.request_router.clone(),
             orphan_stream_registry: self.orphan_stream_registry.clone(),
             streaming_threshold: self.streaming_threshold,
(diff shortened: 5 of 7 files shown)


sanity:
## Review summary

**Self-review** across four lenses (code-first, testing, skeptical, big-picture) plus an **external non-Claude model review** (Gemini 2.5 Flash — Codex CLI and Gemini 2.5 Pro were both rate-limited at review time).

### External review (Gemini) findings
- **Concurrency safety** of `record_broadcast` vs `drain_window` — assessed *solid*: `DashMap` shard guards + `Relaxed` atomic `fetch_add`/`swap` correctly snapshot-and-reset per window with no lost increments; `remove_if` re-checks zero after the swap so a concurrently re-populated entry isn't evicted.
- **Retry-gating** in `handle_broadcast_state_change` — assessed *correct*: `if !self.broadcast_retries.contains_key(&key)` counts only the initial fan-out, not retry re-emissions, so `updates` isn't inflated ~4x on stuck contracts.
- **Unbounded growth** — assessed *addressed*: `MAX_TRACKED_CONTRACTS` cap + idle-entry eviction.
- **Test gap (only finding): no integration test** verifying the end-to-end wiring from `handle_broadcast_state_change` → `record_broadcast` → summary log.

### How the finding was addressed
Rather than stand up a full `P2pConnManager` + `OpManager` + network harness (heavy machinery for a single straight-line method call), I added a **source-scrape pin test** (`broadcast_path_feeds_propagation_stats_pin_test`) that asserts the broadcast handler still calls `record_broadcast` **and** that the call stays inside the `!broadcast_retries.contains_key` gate. This guards the exact "manually-mirrored telemetry counter rots after a migration" failure class called out in `.claude/rules/bug-prevention-patterns.md` — a future task-per-tx refactor that silently drops the call now fails CI. This is the same cheap, durable pattern the existing #4272 pin tests use, and it directly covers the wiring + gate logic the integration test would have exercised.

### Self-review notes
- This is **purely additive telemetry** — no wire-format / protocol / consensus / routing / crypto / migration surface. UPDATE behavior and the OTLP/dashboard paths are untouched; none of the #4272 DEBUG demotions are reverted.
- `{:.8}` precision on the per-contract id is consistent with the existing `format!("{:.8}", key)` sites in `update.rs` (stdlib's `ContractInstanceId`/`ContractKey` Display ignores precision, so neither truncates — same convention, no new behavior).

`cargo fmt`, `cargo clippy --locked -- -D warnings` (the CI invocation) clean; 10 new unit tests pass; existing update-module and op_state_manager tests pass.

[AI-assisted - Claude]

sanity:
## Addressed rule-review findings (HEAD `c2824749`)

**Warning — fire-and-forget spawn missing `BackgroundTaskMonitor`:** Fixed. `start_summary_task` now returns its `JoinHandle`, and `OpManager::new` registers it with the `BackgroundTaskMonitor` as `"update_propagation_summary"` (same pattern as `garbage_cleanup`). It's a node-lifetime task, so per `code-style.md` the handle is tracked, not dropped.

**Info — no startup jitter:** Also addressed, for consistency with the sibling periodic tasks it models itself on. Added a random 0–30s initial delay (`GlobalRng::random_range`) so nodes don't all emit their window summary on the same wall-clock instant, de-correlating the log feed across the network.

`cargo fmt` + `cargo clippy --locked -- -D warnings` clean; all 10 propagation_stats unit tests and the 25 op_state_manager tests pass.

[AI-assisted - Claude]

sanity:
## CI note: Simulation failure is the known flake #4282 (unrelated to this PR)

The `Simulation` check failed on `test_get_routing_coverage_low_htl` (`simulation_integration.rs:7559`, "1 nodes failed... contract only cached at ~5 nodes due to HTL=4", after nextest TRY-3-FAIL). This is the **exact** signature tracked in open issue #4282 ("flaky: test_get_routing_coverage_low_htl fails intermittently after nextest retry-3"), independently observed on unrelated PR #4270.

This PR's diff touches only the UPDATE broadcast path and a new logging module (`propagation_stats.rs`) — nothing in GET/routing/topology — so it cannot affect GET routing coverage. Re-running the failed Simulation job to get a clean signal on the current HEAD; not overriding red CI.

[AI-assisted - Claude]

sanity:
## Root-caused the Simulation failure — it WAS this PR, now fixed (HEAD `3cf401c9`)

My earlier "known flake #4282" call was wrong. Systematic check: the failing test `test_get_routing_coverage_low_htl` **passes on pristine `main`** but **failed deterministically on my branch** with the same seed — so my change caused it, not the pre-existing flake.

**Root cause:** to address the rule-review *info*-level suggestion to add startup jitter, I had `start_summary_task` call `GlobalRng::random_range(0..=30)` at node startup. `GlobalRng` is the *deterministic simulation RNG*; an extra draw at startup shifts the entire RNG stream, perturbing peer locations / routing decisions and flipping the seed-sensitive HTL=4 routing-coverage assertion.

**Fix:** removed the startup jitter entirely. This task only writes to the local log (no network side-effect), so de-correlating emit times across peers buys nothing — whereas an extra `GlobalRng` draw breaks sim determinism. The task is now RNG-free, with a comment explaining why. The `BackgroundTaskMonitor` registration (the actual *Warning* finding) stays.

**Verified:** with the jitter removed, `test_get_routing_coverage_low_htl` now passes locally on this branch (matching pristine `main`), all 10 `propagation_stats` unit tests pass, and `cargo clippy --locked -- -D warnings` is clean.

This is a good reminder that the rule-review "add jitter like the siblings" heuristic doesn't apply to RNG-free logging tasks in a deterministic-simulation codebase.

[AI-assisted - Claude]

sanity:
## Addressed Codex re-review (HEAD `826c3b18`)

Codex re-ran on the previous revision and raised two findings; both addressed, plus the earlier two-finding round (terminal-outcome recording + cap blind-spot) that prompted this revision.

**[P1] `cmp` on `ContractInstanceId`.** Codex flagged that `ContractInstanceId` (stdlib 0.8.1) implements `Eq/Hash/Clone/Copy` but not `Ord`, so the tie-break wouldn't compile. It actually *did* compile — via silent deref-coercion (`ContractInstanceId: Deref<Target=[u8; 32]>`, and `[u8; 32]: Ord`). But Codex's underlying concern is fair: relying on accidental deref coercion is fragile and non-obvious. **Fixed** by comparing the stable public byte representation explicitly: `a.contract.as_bytes().cmp(b.contract.as_bytes())`, with a comment noting the type isn't `Ord`. Same deterministic byte order, no reliance on coercion.

**[P2] Multi-field drain race / field-split.** Valid concurrency bug: with four independent per-field atomics, a `drain_window` swapping `updates` between a producer's `updates++` and its `targets_total++` could split one record across windows, and `aggregate_window`'s `updates > 0` filter would then drop the orphaned targets. **Fixed** by replacing the four atomics with plain `u64` fields mutated under the `DashMap` per-key write guard (`entry`/`retain`), so a record and a drain of the same contract are mutually exclusive — each multi-field record is captured atomically as a unit. Different contracts still update in parallel (separate shards). Added a stress test (`concurrent_record_and_drain_conserve_totals_without_field_split`): 8 threads × 2000 records vs. a concurrent drainer, asserting (a) grand totals are conserved exactly and (b) no drained summary ever shows activity with `updates == 0`.

**[Prior P2.1] Healed-retry mis-count.** `record_broadcast` now fires on the *terminal* outcome of each logical broadcast — once on the targets-found success path, once on the retries-exhausted no-targets path — never in the retry-scheduling branch. A contract whose retry heals is counted as a propagated update (with its real target count), not a no_targets miss, and `updates` stays one-per-logical-broadcast.

**[Prior P2.2] Cap blind-spot.** `drain_window` now clears the whole map each window, so a window that fills `MAX_TRACKED_CONTRACTS` with one-off contracts can't block new contracts in the next window. Covered by `drain_restores_cap_headroom_for_next_window`.

**Verification:** `cargo clippy --locked -- -D warnings` (exact CI invocation) clean; 12 propagation_stats unit tests + 56 update lib tests pass; and the RNG-sensitive `test_get_routing_coverage_low_htl` sim test passes locally (matching pristine main — the propagation recording draws no `GlobalRng`, so determinism is preserved).

[AI-assisted - Claude]

sanity:
## Addressed Codex P2 — no-target under-counting (HEAD `d6f9b2df`)

**Finding:** because retries and fresh broadcasts share the per-contract `broadcast_retries` counter, recording the no-target outcome only in the retries-exhausted arm meant several fresh no-target broadcasts arriving during one retry cycle were absorbed as retries and collapsed to a single record — under-reporting hot no-target contracts, the exact failure mode this summary exists to surface. Valid.

**Fix:** moved the NO_TARGETS record to the *start* of a no-target retry cycle, gated on `retry_count == 0`. That value is 0 only on the first no-target observation for a contract (a fresh `BroadcastStateChange` that found no targets) — retry re-emissions see `retry_count > 0` and don't re-record. So:

- Each fresh no-target broadcast records exactly one `no_targets` immediately (no more collapsing many into one).
- A stuck contract's retries don't re-record, so counts don't inflate ~4x.
- A retry that *heals* (finds targets) additionally records a success on the targets-found path — both the initial miss and the eventual propagation are surfaced honestly.

The retries-exhausted recording was removed (the miss is already counted at cycle start). Still exactly two `record_broadcast` call sites; the wiring pin test was updated to match. The residual edge — a fresh broadcast arriving while `retry_count` is mid-cycle (1–2) being absorbed — is an unavoidable consequence of the pre-existing shared counter, and is strictly less lossy than the previous behavior.

**Verification:** `cargo clippy --locked -- -D warnings` clean; 12 propagation_stats + 56 update lib tests pass; the RNG-sensitive `test_get_routing_coverage_low_htl` sim test still passes locally (the new record path draws no `GlobalRng`).

[AI-assisted - Claude]

sanity:
## Fully resolved Codex P2 — fresh no-target broadcasts now counted unconditionally (HEAD `25a5a234`)

The previous revision gated the no-target record on `retry_count == 0`, which still absorbed a fresh broadcast that arrived mid-retry-cycle. Codex correctly noted that's the same under-count, just narrower. Rather than leave a residual edge, I made the distinction explicit at the source.

**Fix:** added an `is_retry: bool` field to the internal `NodeEvent::BroadcastStateChange` event (a `pub(crate)`, `#[derive(Debug, Clone)]` in-process coordination enum — **not** a wire-format/serialized type, so no protocol surface). The 3 executor emission sites set `is_retry: false`; the single retry re-emission in `handle_broadcast_state_change` sets `is_retry: true`. The handler now gates the NO_TARGETS record on `!is_retry` instead of the shared `broadcast_retries` counter.

Result: **every** fresh no-target broadcast is counted exactly once — regardless of whether a previous retry cycle for the same contract is active — and retry re-emissions are never counted. No residual edge. The success path still records on any targets-found (fresh or healed retry), so a recovered miss surfaces both its initial no_targets and its eventual propagation.

Pin test updated to assert the `!is_retry` gate so a future refactor can't start counting retries as fresh misses.

**Verification:** `cargo clippy --locked -- -D warnings` clean; 12 propagation_stats + 56 update lib tests pass; full test build (`--tests --features simulation_tests,testing`) compiles, confirming all `BroadcastStateChange` construction sites are updated; `test_get_routing_coverage_low_htl` sim test still passes (no `GlobalRng` draw added). Change is contained to telemetry gating — no propagation/retry behavior changed.

[AI-assisted - Claude]

sanity:
## Final review status — clean (HEAD `25a5a234`)

**External (Codex) re-review on the final content: "No discrete correctness issues were found in the diff."** All prior findings addressed and verified across the iteration:

| Finding | Resolution |
|---|---|
| P1: `cmp` on non-`Ord` `ContractInstanceId` | Compare `as_bytes()` explicitly (no deref-coercion reliance) |
| P2: multi-field drain field-split race | Plain `u64` under the DashMap per-key write guard; record+drain mutually exclusive; concurrency stress test added |
| P2: healed-retry mis-counted as no_targets | Success path records on any targets-found (fresh or healed retry) |
| P2: fresh no-target broadcasts under-counted during retry cycles | Explicit `is_retry` flag on the internal event; fresh misses always counted, retries never |

**Self-review (4 lenses):**
- *Code-first / skeptical:* counters are bounded (`MAX_TRACKED_CONTRACTS`, map cleared each window), log rate ≤ `TOP_K+1` lines/window (<0.2/s), no `GlobalRng` draw (simulation determinism preserved — verified against the RNG-sensitive `test_get_routing_coverage_low_htl`), background task registered with `BackgroundTaskMonitor`. `is_retry` added to an internal `pub(crate)` event only — no wire/protocol surface.
- *Testing:* 13 unit tests (pure aggregation, record/drain cycle, cap headroom, concurrency atomicity, INFO-emission, two source-scrape pins). Full `freenet` lib suite: **2905 passed, 0 failed**.
- *Big-picture:* matches the issue's preferred shape (per-contract top-K + rollup, 60s window, OpManager-owned counter, surfaces no_targets + interest_resolve_failed); none of the #4272 DEBUG demotions reverted; OTLP/dashboard paths untouched; UPDATE behavior unchanged.

`cargo fmt` + `cargo clippy --locked -- -D warnings` (exact CI invocation) clean.

[AI-assisted - Claude]

PR #3840, opened 2026-04-13
feat(server): integrate browser history with gateway iframe navigation

sanity:
## Problem

When the gateway's postMessage bridge navigates between pages within a
web container (via the navigate handler from #3834), it updates
`iframe.src` but does not push entries to the browser's history stack.
As a result:

1. **Back button doesn't work** — clicking back either does nothing or
   navigates away from the contract entirely.
2. **URL doesn't update** — the browser address bar always shows the
   initial contract URL regardless of which subpage the user is on.
3. **No way to recover from dead-end pages** — if a user navigates to a
   page that doesn't exist, they're stuck with no way to go back.

## Solution

Extend the shell bridge's `navigate` handler to call `history.pushState`
with a tagged state object (`__freenet_nav__`) carrying the iframe's
sandbox URL. Add a `popstate` listener that restores `iframe.src` from
this state on back/forward, falling back to the existing hash-forwarding
behaviour for non-nav entries. The initial history entry is seeded with
`replaceState` so back-navigating to it still triggers iframe restore.

The existing `hash` message handler now preserves the state object
instead of wiping it with `null`, so a hash update does not break the
nav marker on the current entry.

The pushState URL uses the clean (non-sandbox) path, so the address bar
shows the user-visible URL; the shell always re-adds `__sandbox=1` when
loading the iframe.

**Security:** both pushState and popstate paths validate that the
target iframe path lives under the current contract's web prefix, so a
stale state object cannot redirect the iframe cross-contract. The
existing same-origin and prefix checks on the navigate handler are
preserved.

## Testing

Three new regression tests in `path_handlers.rs`:

- `bridge_js_navigate_pushes_history_state` — asserts the navigate
  handler calls `pushState` with the `__freenet_nav__` marker and
  `iframePath`, and that the URL argument is the clean path.
- `bridge_js_popstate_restores_iframe_from_state` — asserts the popstate
  listener checks the marker, validates the contract prefix, and
  restores `iframe.src`.
- `bridge_js_seeds_initial_history_state` — asserts the initial entry
  is seeded via `replaceState` and that the hash replaceState preserves
  existing state (`history.replaceState(history.state, ...)`).

Updated `bridge_js_contains_origin_check` — removed the now-stale
assertion forbidding `history.pushState`, which is intentional for
navigation.

**Out of scope:** deep-link reload (refreshing the browser on a pushed
URL like `/v1/contract/web/KEY/news/`). The server still routes
non-sandbox sub-paths through `variable_content`, so a reload will 404.
[...]

PR #4335, opened 2026-06-02
fix: serve shell for deep-link reload of contract subpaths (#3841)

### LINKED ISSUE #3841: Deep-link reload for pushState'd contract subpaths returns 404

sanity:
## Problem

After #3840, in-contract navigation pushes browser history entries, so the address bar reflects the current subpage (e.g. `/v1/contract/web/KEY/news/`). But **reloading** or **bookmarking** one of those URLs returned 404 — and, after the interim fix (commit `c74fcf1b`), a 303 redirect to the contract root. The redirect preserved URL *fragments* (enough for hash-routed apps like Delta) but **flattened path-routed deep links to `/`**, silently dropping the user at the root of any contract that uses path-based navigation. Deep links could not be shared or reloaded.

Closes #3841

## Approach

Thread an `Option<&str>` sub-path through `web_subpages` → `path_handlers::contract_home` → `shell_page`. When `web_subpages` receives a non-sandbox, top-level HTML **document** load of a sub-path, it now renders the shell **in place** (instead of redirecting) with the sub-path embedded in the iframe's `data-src`:

```
data-src="/v1/contract/web/{KEY}/{sub_path}?__sandbox=1"
```

so the in-iframe webapp starts on the requested route.

- The shell-rendering logic (fresh auth token + secure cookie + shell CSP + anti-framing headers) is factored out of `web_home` into a shared `render_shell_response` helper, so the deep-link path issues the **same** credentials/headers as the contract-root path and cannot drift out of sync.
- The sub-path is validated by `sanitize_shell_sub_path` before interpolation: rejects `?`, `#`, leading `/`, `\`, control/whitespace (incl. CR/LF), **and `.`/`..` traversal segments**. The `..` rejection is security-critical — the browser collapses dot-segments in a URL *before* requesting the iframe, so `.../KEY/../OTHER/` would otherwise normalize to `.../OTHER/` and load a *different contract* under the current shell's token (caught by the Codex review; the later on-disk canonicalization in `sandbox_content_body` never sees the un-normalized form).
- Sensitive query params (`__sandbox*`, `authToken*`) are still stripped by `shell_page`, preserving the protection the old redirect provided.
- Sub-resource fetches and non-HTML asset paths still fall through to `variable_content`, so **genuinely missing assets keep returning 404** — the SPA fallback only triggers for top-level HTML document loads.

`redirect_to_shell_root` / `build_canonical_shell_url` remain in use for the sandbox-URL top-level guard and the no-trailing-slash canonical redirect.

## Testing

New regression tests:
- `shell_page_embeds_sub_path_in_iframe_data_src` — iframe `data-src` carries the sub-path (directory + nested forms); root load unchanged. **Verified this FAILS without the fix and PASSES with it.**
- `contract_home_with_sub_path_renders_shell_for_that_page` — end-to-end through the real `ensure_contract_cached` GET cycle. **Also verified failing-without / passing-with.**
- `sanitize_shell_sub_path_accepts_safe_paths_and_rejects_dangerous` — accepts `news/`, `about/team`, etc.; rejects query/fragment/control chars and `.`/`..` segments.
- `web_subpages_serves_shell_for_top_level_document_load` — top-level document load routes into the shell-render path (observed via the `NewConnection` it emits) rather than a 303; the sandbox top-level guard still redirects; curl/no-`Sec-Fetch-Dest` fallthrough to `variable_content` unchanged.

Updated the existing predicate/ordering tests for the rename (`should_redirect_subpage_to_shell` → `should_serve_shell_for_subpage`).

`cargo test -p freenet` lib suite green (2896 passed); `cargo fmt` clean; `cargo clippy --locked -- -D warnings` clean.

## Review

External non-Claude review run: Codex (found the `..` traversal P1, fixed here) and Gemini 2.5 Flash on the fixed code (no material findings). Self-reviewed across code-first / skeptical / testing / big-picture lenses.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -190,8 +190,6 @@ async fn web_home(
     api_version: ApiVersion,
     query_string: Option<String>,
 ) -> Result<axum::response::Response, WebSocketApiError> {
-    use headers::{Header, HeaderMapExt};
-
     // Check if this is the sandboxed iframe requesting its content
     let is_sandbox = query_string
         .as_ref()
@@ -201,6 +199,31 @@ async fn web_home(
     if is_sandbox {
         return serve_sandbox_response(key, api_version, None, &req_headers, rs).await;
     }
+
+    // Root document load: render the shell that wraps the contract root.
+    render_shell_response(key, &config, api_version, query_string, None, rs).await
+}
+
+/// Generates a shell page response: a fresh auth token + secure cookie,
+/// the contract fetched/unpacked into the cache, and the iframe wrapper
+/// with the shell CSP and anti-framing headers.
+///
+/// Shared by `web_home` (root document loads, `sub_path = None`) and the
+/// deep-link branch of `web_subpages` (`sub_path = Some(path)` so a
+/// reloaded/bookmarked sub-page URL renders the shell pointed at that
+/// page instead of 404ing or being flattened to the contract root — see
+/// freenet/freenet-core#3841). Both must issue the SAME credentials and
+/// headers; factoring it here keeps the deep-link path from drifting out
+/// of sync with the root path (e.g. forgetting the cookie or the CSP).
+async fn render_shell_response(
+    key: String,
+    config: &Config,
+    api_version: ApiVersion,
+    query_string: Option<String>,
+    sub_path: Option<&str>,
+    rs: HttpClientApiRequest,
+) -> Result<axum::response::Response, WebSocketApiError> {
+    use headers::{Header, HeaderMapExt};
 
     // Shell page: generate auth token, serve iframe wrapper with CSP
     let token = AuthToken::generate();
@@ -219,7 +242,8 @@ async fn web_home(
 
     let token_header = headers::Authorization::bearer(token.as_str()).unwrap();
     let contract_response =
-        path_handlers::contract_home(key, rs, token.clone(), api_version, query_string).await?;
+        path_handlers::contract_home(key, rs, token.clone(), api_version, query_string, sub_path)
+            .await?;
 
     let mut response = contract_response.into_response();
     response.headers_mut().typed_insert(token_header);
@@ -259,6 +283,7 @@ async fn web_subpages(
     api_version: ApiVersion,
     query_string: Option<String>,
     req_headers: axum::http::HeaderMap,
+    config: &Config,
     request_sender: HttpClientApiRequest,
 ) -> Result<axum::response::Response, WebSocketApiError> {
     let is_sandbox = query_string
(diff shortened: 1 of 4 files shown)


sanity:
## Review summary

**Self-review** (code-first / skeptical / testing / big-picture):
- Confirmed sensitive-param stripping (`__sandbox*`, `authToken*`) and key validation are preserved on the new shell-render path (via `shell_page`'s filter + `contract_home`'s `ContractInstanceId::from_bytes`), matching the old redirect's protections.
- Confirmed the cookie `.unwrap()` is unreachable for an invalid key (`contract_home` returns `Err` via `?` before the cookie is inserted).
- Confirmed sub-resource fetches and non-HTML paths still fall through to `variable_content`, so real assets keep 404ing — the SPA fallback is scoped to top-level HTML document loads only.

**External non-Claude review:**
- **Codex** flagged a **P1**: `sanitize_shell_sub_path` originally left `..` for the on-disk canonicalization layer, but the browser normalizes `..` in the iframe URL *before* requesting it, so `.../KEY/../OTHER/` would load a different contract under the current shell's token. **Fixed** — `sanitize_shell_sub_path` now rejects `.`/`..` segments outright, with a dedicated regression test. (A second Codex pass to confirm the fix was blocked by a usage quota.)
- **Gemini 2.5 Flash** reviewed the fixed code and found nothing material, explicitly noting the path-traversal concern is addressed and tested.

[AI-assisted - Claude]

PR #4336, opened 2026-06-02
fix: detect stale Linux systemd unit by content equality (#4287)

### LINKED ISSUE #4287: Linux systemd unit: stale-unit check uses same fragile marker pattern as #3967 macOS wrapper
## Problem

`systemd_unit_needs_regen` in `crates/core/src/bin/commands/update.rs:1120`
uses the same marker-substring detection pattern that PR #4286 just
replaced on macOS:

\`\`\`rust
let has_auto_update_markers = content.contains(\"ExecStopPost=\")
    && content.contains(\"SuccessExitStatus=42\")
    && content.contains(\"RestartPreventExitStatus=43\");
let has_legacy_log_target = ...;
!has_auto_update_markers || has_legacy_log_target
\`\`\`

This is exactly the anti-pattern #3967 surfaced: a future systemd-unit
template change (new \`ExecStartPre=\`, new \`Restart=\` mode, additional
\`SuccessExitStatus\` value, etc.) will silently fail to deploy to
existing Linux installs because the marker check still returns
\`false\` even though the unit is materially stale.

No production reports for Linux yet, but the pattern is the same; it
just hasn't bitten.

## Proposed fix

Apply the same equality-based detection + SHA-256 sidecar
user-modification protection as PR #4286 did for macOS. The helpers
(\`wrapper_needs_regen\`, \`wrapper_content_hash\`, \`is_valid_wrapper_hash\`,
\`decide_wrapper_action\`, \`write_wrapper_hash_sidecar\`) are already
cross-platform; they just need a Linux caller.

Naming: \`service_unit_needs_regen(current_template, on_disk)\` reusing
the same equality predicate, or generalize \`decide_wrapper_action\` to
handle both surfaces.

Edge cases worth thinking about:
- User-customized \`Environment=\` lines or \`User=\` overrides — the
  generator currently embeds those, so equality across runs depends
  on the update path passing the same arguments to
  \`generate_user_service_file\` / \`generate_system_service_file\` each
  time. Worth a unit test.
- System-mode units owned by root: sidecar path needs to be writable
  by root, not the user, so the path location differs from macOS.

## Acceptance criteria

- [ ] systemd-unit staleness uses equality against
      \`generate_user_service_file\` / \`generate_system_service_file\`
      output, not marker substrings.
- [ ] Sidecar (e.g. \`freenet.service.hash\`) records the hash Freenet
      last wrote.
- [ ] Pure-helper tests on every CI runner cover the decision matrix
      (mirror of \`decide_wrapper_action_*\` from PR #4286).
- [ ] Existing installs without a sidecar are migrated on first
      \`freenet update\` (write replacement + record hash).

## References

- PR #4286 — macOS implementation of the equality + sidecar pattern.
- Issue #3967 — original incident (macOS).
- \`crates/core/src/bin/commands/update.rs:1120\` —
  \`systemd_unit_needs_regen\`.

[AI-assisted - Claude]

sanity:
## Problem

`update_service_file` (Linux) decided whether the on-disk systemd unit
was stale by grepping for a fixed set of marker substrings (`ExecStopPost=`,
`SuccessExitStatus=42`, `RestartPreventExitStatus=43`, plus absence of the
legacy `StandardOutput=append:` log target). This is the same fragile
anti-pattern that PR #4286 (#3967) replaced on the macOS launchd wrapper:
any future systemd-unit template change introducing a new directive (a new
`ExecStartPre=`, a changed `Restart=` policy, an extra `SuccessExitStatus`
value, etc.) is invisible to the marker check. A stale unit that still
contains the three markers returns "up to date", so the update silently
fails to propagate to existing Linux installs.

No production reports on Linux yet, but the failure mode is identical to
the macOS incident (#3967) that served stale assets for 19 days.

## Solution

Replace the marker grep with byte-for-byte equality against what
`generate_{user,system}_service_file` produces now, reusing the
cross-platform `decide_wrapper_action` decision matrix introduced for
macOS in #4286 (the helper reasons about `(template, on-disk content,
sidecar)` regardless of artifact type — it just needed a Linux caller).
A `freenet.service.hash` sidecar records the SHA-256 of the unit we last
wrote so a later update can tell "Freenet's previous unit" from an
operator hand-edit and avoid clobbering customizations.

- System-mode units inherit `User=` / `Environment=HOME=` from the
  on-disk unit (via the extracted `render_current_systemd_unit` helper)
  so the equality check is stable across runs and only flags genuine
  template drift, not the operator's service-user choice.
- Both install paths (`install_user_service`, `install_system_service`)
  now write the sidecar so fresh installs don't trigger a needless
  rewrite on the first `freenet update`. The system-mode sidecar lives
  next to the root-owned unit and is written by the same root process.
- Migration: pre-#4287 installs have no sidecar; that case is treated
  as "ours" so the next update fixes the unit and records the hash. A
  `.service.bak` backup is written first, so a regen is recoverable.
- Malformed (truncated/corrupted) sidecars are treated as "missing"
  rather than locking the host into permanent `UserModified`.

Removed the now-dead `systemd_unit_needs_regen` marker predicate; the
#4251 legacy-log-target staleness it detected is subsumed by equality
(a legacy-append unit differs from the journal-target template).

This is service-file management on the local host (systemd unit text on
disk) — not a wire-format/protocol/consensus/crypto change, and the
"migration" here is install-sidecar bookkeeping, not a DB/schema migration.

## Testing

New cross-platform pure-decision tests exercise the full decision matrix
on **every** CI runner (not just Linux), mirroring the
`decide_wrapper_action_*` coverage from #4286:

- `systemd_unit_new_directive_drift_is_detected_4287` — the motivating
  bug: a unit that still satisfies the OLD marker check but is missing a
  new template directive is now detected as drifted.
- `systemd_unit_no_sidecar_stale_migrates_to_ours_4287` — migration shape.
- `systemd_unit_identical_to_template_with_sidecar_is_noop_4287`.
- `systemd_unit_identical_to_template_missing_sidecar_backfills_4287`.
- `systemd_unit_user_modified_is_preserved_4287`.
- `systemd_unit_malformed_sidecar_treated_as_missing_4287`.

Plus Linux-only integration tests on the real generators:

- `freshly_generated_systemd_unit_does_not_trigger_regen_loop_4287` —
  determinism: install shape is a no-op (no rewrite loop).
- `render_current_systemd_unit_round_trips_user_and_home_4287` — covers
  the issue's explicit `User=`/`Environment=` override edge case.
- `render_current_systemd_unit_infers_home_when_env_home_absent_4287` —
  inferred `/home/{user}` fallback (added per external review).

**Fail-before / pass-after verified:** the old marker predicate returns
`false` (would NOT regen) for the drifted unit the new test feeds in,
while equality returns `WriteOurs` (regen).

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and
`cargo test -p freenet --bin freenet` all clean (170 bin tests pass).

## Review

External non-Claude review via Gemini CLI (Codex was rate-limited):
no blocking findings. Two non-blocking points + one nit addressed —
added the inferred-home edge-case test and aligned the backfill-failure
warning message with the WriteOurs path. The remaining non-blocking
test-rename suggestion was dismissed to keep naming consistent with the
macOS sibling test (`decide_wrapper_action_malformed_sidecar_treated_as_missing_3967`).

Closes #4287

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2478,7 +2478,21 @@ fn install_user_service() -> Result<()> {
     let service_content = generate_user_service_file(&exe_path, &log_dir);
     let service_path = service_dir.join("freenet.service");
 
-    fs::write(&service_path, service_content).context("Failed to write service file")?;
+    fs::write(&service_path, &service_content).context("Failed to write service file")?;
+
+    // Sidecar records the unit's SHA-256 so a later `freenet update` can
+    // distinguish "Freenet's unit" from a hand-edited one before
+    // overwriting (#4287). A failed sidecar write only weakens future
+    // user-modification protection — warn and continue.
+    let hash_path = service_path.with_extension("service.hash");
+    let unit_hash = super::update::wrapper_content_hash(&service_content);
+    if let Err(e) = super::update::write_wrapper_hash_sidecar(&hash_path, &unit_hash) {
+        eprintln!(
+            "Warning: failed to write service hash sidecar at {}: {}.",
+            hash_path.display(),
+            e
+        );
+    }
 
     // Reload systemd user daemon
     systemctl_with_hint(false, &["daemon-reload"], "reload systemd daemon")?;
@@ -2539,6 +2553,22 @@ fn install_system_service() -> Result<()> {
              Are you running as root? Try: sudo freenet service install --system"
         )
     })?;
+
+    // Sidecar records the unit's SHA-256 so a later `freenet update` can
+    // distinguish "Freenet's unit" from a hand-edited one before
+    // overwriting (#4287). Lives next to the root-owned unit and is
+    // written by the same root process, so root owns it too. A failed
+    // sidecar write only weakens future user-modification protection —
+    // warn and continue.
+    let system_hash_path = Path::new(SYSTEM_SERVICE_PATH).with_extension("service.hash");
+    let unit_hash = super::update::wrapper_content_hash(&service_content);
+    if let Err(e) = super::update::write_wrapper_hash_sidecar(&system_hash_path, &unit_hash) {
+        eprintln!(
+            "Warning: failed to write service hash sidecar at {}: {}.",
+            system_hash_path.display(),
+            e
+        );
+    }
 
     // Reload systemd daemon (system-level, no --user)
     let status = systemctl(true, &["daemon-reload"])?;
--- a/crates/core/src/bin/commands/update.rs
+++ b/crates/core/src/bin/commands/update.rs
@@ -1074,8 +1074,10 @@ fn verify_binary(binary_path: &Path) -> Result<()> {
     Ok(())
 }
 
-/// Check if the service file needs updating (missing auto-update hook) and update if so.
-/// This ensures users who installed before v0.1.75 get the ExecStopPost hook.
+/// Check if the on-disk systemd unit has drifted from the current
+/// template and rewrite it if so. Locates the user unit first, then the
+/// system unit, and delegates the staleness decision to
+/// `update_service_file` (equality-based since #4287).
 #[cfg(target_os = "linux")]
 fn ensure_service_file_updated(binary_path: &Path, quiet: bool) -> Result<()> {
     // Try user service first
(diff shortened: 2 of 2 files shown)


sanity:
## Review summary

**Self-review (four lenses):**
- *Code-first:* The `match &action { … } / let WrapperAction::WriteOurs { … } = &action else { unreachable!() }` shape returns early for all non-`WriteOurs` arms, then re-binds for the rewrite path. Safe — the `unreachable!` is genuinely unreachable.
- *Skeptical:* System-mode `User=`/`HOME=` are inherited from the on-disk unit, so an operator's service-user choice is preserved (no spurious drift). A hand-edited non-User/HOME directive on a pre-#4287 install is migrated to the template (after a `.bak` backup) — same tradeoff #4286 made for macOS.
- *Testing:* Decision matrix covered on every CI runner (migration / no-op / backfill / user-modified / malformed-sidecar / new-directive-drift) + Linux-only determinism and round-trip tests.
- *Big-picture:* `systemd_unit_needs_regen` removed; the #4251 legacy-log-target staleness is subsumed by equality. `decide_wrapper_action` reused rather than duplicated.

**External review:** Gemini CLI (Codex was rate-limited until later today). No blocking findings; confirmed all five focus areas robust. Addressed: added `render_current_systemd_unit_infers_home_when_env_home_absent_4287` and aligned the backfill-failure warning with the `WriteOurs` path. Dismissed one test-rename suggestion to stay consistent with the macOS sibling test name.

[AI-assisted - Claude]

PR #4337, opened 2026-06-02
fix: deadlock detection test flaky under parallel execution (#3627)

### LINKED ISSUE #3627: test: test_deadlock_is_detected flaky under parallel execution
## Problem

`deadlock_detection::tests::test_deadlock_is_detected` fails intermittently when run as part of the full test suite (`cargo test -p freenet`) but passes consistently when run in isolation (`cargo test -p freenet --lib test_deadlock_is_detected`).

## Reproduction

```bash
# Fails intermittently:
cargo test -p freenet

# Always passes:
cargo test -p freenet --lib test_deadlock_is_detected
```

## Analysis

This is likely a test interaction issue — the deadlock detection test may be affected by other tests running in parallel that manipulate global state. The test is in `crates/core/src/deadlock_detection.rs:164`.

## Observed

- Full suite: FAILED (1 of 2003)
- Isolated: PASSED
- Branch: fix-3625 (but change is unrelated — only touches operations/put.rs)

[AI-assisted - Claude]

sanity:
## Problem

`test_deadlock_is_detected` passed in isolation but flaked when run as part of the full suite under single-process parallel execution (`cargo test`). There were **two independent root causes**, both rooted in the fact that `parking_lot::deadlock::check_deadlock()` is **process-global** — it inspects the entire process and returns every lock-wait cycle exactly once (a reported cycle is never reported again):

1. **Exact-count assertions on global state.** The test asserted `deadlocks.len() == 1` and `deadlocks[0].len() == 2`. Under parallel execution, any *other* test that formed a parking_lot lock-wait cycle while this one was polling also appeared in the returned vector, tripping those exact-count assertions even though the test's own deadlock was detected fine. `#[serial(deadlock_detection)]` only serializes the deadlock-keyed tests against each other, not against the hundreds of unrelated tests running in parallel.

2. **Lingering detector thread consuming the next test's cycle.** `DeadlockDetectorGuard::drop` only set a shutdown flag and returned immediately without joining the detector thread, and the detector loop checked that flag only at the top of the loop after sleeping a full interval. So after `drop` returned, the thread could perform one more *consuming* `check_deadlock()` call. When the detector-lifecycle test ran just before `test_deadlock_is_detected` (both share the serial key), the lingering detector could steal the next test's freshly-created deadlock report, leaving `detected == false`.

## Solution

- **Match the test's own cycle by thread id.** `DeadlockedThread::thread_id()` returns the same `std::thread::ThreadId` that `JoinHandle::thread().id()` yields, so the test now identifies its two spawned threads' ids and asserts only that this specific 2-thread cycle appears among whatever `check_deadlock()` reports. The assertion is independent of any other concurrent deadlock in the process.

- **Join the detector thread on drop**, and re-check the shutdown flag immediately before every `check_deadlock()` call (polling the flag at fine granularity). Together these guarantee no detector `check_deadlock()` call can outlive the guard, so a dropped detector cannot consume a later test's cycle. This also fixes the guard's documented contract ("the detector thread is stopped when this guard is dropped"), which the old non-joining drop did not actually honor.

## Testing

- `test_concurrent_deadlocks_are_each_detected`: stands up two independent deadlocks and asserts a single `check_deadlock()` call reports BOTH at once — the pre-fix `assert_eq!(deadlocks.len(), 1)` would have panicked here.
- `test_dropped_detector_does_not_consume_later_deadlock`: runs a fast detector, drops it, then creates a deadlock and asserts the cycle is still observable — would race-fail with the old non-joining drop.
- Verified both failure directions with standalone reproductions. Ran the full `freenet` lib suite **8x under default parallel execution**: 2895 tests, 0 failures, all four deadlock tests green every run. Also 40x single-threaded stable. `cargo fmt` and `cargo clippy --locked -- -D warnings` clean. External Codex review: no findings.

Closes #3627

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/deadlock_detection.rs
+++ b/crates/core/src/deadlock_detection.rs
@@ -37,14 +37,41 @@ use std::time::Duration;
 
 /// Guard that keeps the deadlock detector thread running.
 ///
-/// The detector thread is stopped when this guard is dropped.
+/// Dropping the guard stops the detector thread **and waits for it to fully
+/// exit** before returning. This is a hard guarantee: once `drop` returns, the
+/// detector thread has made its last [`parking_lot::deadlock::check_deadlock`]
+/// call and is joined.
+///
+/// That join matters because `check_deadlock()` is *process-global* and
+/// consuming: it returns each deadlock cycle exactly once. A detector thread
+/// that lingered past the guard's lifetime could call `check_deadlock()` after
+/// the guard was dropped and silently consume a deadlock cycle that a
+/// subsequent, unrelated test expected to observe. That race made
+/// `test_deadlock_is_detected` flaky under parallel execution (#3627): when the
+/// detector-lifecycle test ran just before it (both share
+/// `#[serial(deadlock_detection)]`), the lifecycle test's not-yet-joined
+/// detector thread could steal the freshly-created deadlock report.
 pub struct DeadlockDetectorGuard {
     shutdown: Arc<AtomicBool>,
+    handle: Option<std::thread::JoinHandle<()>>,
 }
 
 impl Drop for DeadlockDetectorGuard {
     fn drop(&mut self) {
         self.shutdown.store(true, Ordering::Relaxed);
+        if let Some(handle) = self.handle.take() {
+            // Wait for the detector thread to observe the shutdown flag and
+            // exit. Without this join the thread could perform one more
+            // check_deadlock() after drop and consume another test's cycle.
+            //
+            // We deliberately ignore a join error (a panicked detector thread):
+            // re-raising it here would turn drop into a double-panic. A detector
+            // configured to panic on deadlock would already have surfaced that
+            // panic on its own thread.
+            if handle.join().is_err() {
+                // Detector thread panicked; nothing actionable to do in drop.
+            }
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary

**Self-review** (code-first / testing / skeptical / big-picture lenses): the fix addresses two distinct, empirically-confirmed root causes of the flakiness — (1) exact-count assertions on process-global `check_deadlock()` output, and (2) a detector thread that outlived its guard and consumed the next serial test's cycle. Each has a dedicated regression test verified to fail-direction against the old code. The detector-guard join also corrects the guard's documented "stopped on drop" contract, which the old code did not honor.

**External model review (Codex, `codex review --base main`):** clean — "I did not identify any actionable regressions in the changed deadlock detector lifecycle or regression tests. The guard now joins the detector thread and the tests avoid relying on process-global exact deadlock counts." No findings to address.

**Verification:** full `freenet` lib suite run 8x under default parallel execution (2895 tests, 0 failures, all four deadlock tests green each run); 40x single-threaded stable; `cargo fmt` + `cargo clippy --locked -- -D warnings` clean.

[AI-assisted - Claude]

PR #4338, opened 2026-06-03
fix(governance): default the cost/benefit detector to Off

sanity:
## Problem

The contract-governance cost/benefit MAD detector false-flags popular contracts. In production DryRun on the gateways it labelled **~47% of active contracts** (88/189 on nova) as `WouldEvict`, including the **official River room contract**, and a third-party user reported their node showing the River chat contract as "would evict".

Root cause (confirmed on v0.2.68, across nova/vega/technic):
- **Cost** accrues on every hosted merge, including relayed UPDATEs the node didn't originate (`commit_state_write` → `StateBytesWritten` → `ingest_cost`).
- **Benefit** under-credits relay demand: `FORWARDED_DEMAND_WEIGHT = 0.1` vs `LOCAL_DEMAND_WEIGHT = 1.0`. On a gateway a popular room is almost all relay demand, so its live benefit is a fraction of a point while update cost is large.
- **MAD has no absolute-reasonableness floor**: `capacity_ceiling_log` caps the *threshold* (more aggressive), it does not protect a busy-but-legitimate contract. So the busiest relay-hosted contract is *by construction* the population's cost/benefit outlier — popularity itself trips the detector.

The earlier #4309 "live beneficiary stock" fix removed the decay artifact but left this structural issue intact, so the false-positive survived it. The detector's own regression test (`popular_contract_with_subscribers_not_evicted`) only models the sustained-high-subscriber *host*, not the relay-only-demand replica that flaps in production.

## Solution

The detector is being replaced by demand-weighted, contention-only throttling (#4296). Until that lands, **default the governance mode to `Off`** so the miscalibrated detector cannot mislabel a contract on the dashboard or — under any future `Enforce` flip — evict a legitimate one.

- `GovernanceConfig::default().mode` → `Off` (was `DryRun`), with a comment recording the why.
- `ingest_cost` is gated on `Off` so the subsystem is **fully dormant**. Without this, a default-`Off` node would accumulate one never-pruned score per contract it ever hosts, because the reaper `tick` that decays/prunes the map is itself a no-op under `Off`. This keeps default-`Off` leak-free.
- The data model, dashboard rendering (`Off` is already a handled snapshot variant), and `Meter` infrastructure all stay in place. Flip to `DryRun`/`Enforce` explicitly to exercise the detector.

This is **reversible** and changes **no enforcement behavior today** (production was `DryRun`, which never evicts). It only stops the dashboard mislabelling and removes the foot-gun ahead of the redesign.

## Testing

- New regression test `default_governance_is_off_and_dormant`: asserts the default mode is `Off`, and that `Off` is dormant — the abuser signature (large cost, zero benefit) accumulates no score (`len() == 0`) and produces no decisions. Fails without this change (default was `DryRun`, and `ingest_cost` would create a score).
- The detector's own logic tests are unchanged and still pass under explicit `DryRun`/`Enforce` (e.g. `zero_beneficiary_contract_with_cost_is_flagged`), confirming only the *default* changed, not the detector.
- `cargo test -p freenet --lib governance`: 60 passed, 0 failed.
- `cargo clippy --locked -- -D warnings`: clean. `cargo fmt`: clean.

Refs #4296

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/governance.rs
+++ b/crates/core/src/contract/governance.rs
@@ -317,11 +317,13 @@ impl ContractScore {
     }
 }
 
-/// Operating mode for the governance system. Plan defaults to
-/// `DryRun` for one release after Phase 4 lands — operators see what
-/// would be evicted, dashboards reflect intended actions, but no
-/// contracts are actually evicted. After calibration, the operator
-/// (or release default) flips to `Enforce`.
+/// Operating mode for the governance system. Defaults to `Off`: the
+/// cost/benefit MAD detector proved miscalibrated in production DryRun
+/// (it false-flagged popular contracts, including the official River
+/// room — see [`GovernanceConfig::default`] and #4296), so it is
+/// disabled pending the demand-weighted, contention-only redesign.
+/// `DryRun` computes and surfaces `WouldEvict` / `Evicted` states
+/// without acting; `Enforce` additionally evicts/bans. Flip explicitly.
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub(crate) enum GovernanceMode {
     /// Disabled. No state computation, no transitions.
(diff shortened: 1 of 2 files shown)


sanity:
## Review — Codex (external, non-Claude model)

Ran `codex review --base main` (the required independent model pass), plus an adversarial read.

- **Round 1** flagged one P2: with the default now `Off`, the dashboard governance card empty state still rendered "Scoring activates after N contracts," which is misleading on every default node — scoring never activates while `Off`.
- **Fix** (eb75351): made the empty-state progress message + verdict `Off`-aware ("Governance is off: contracts are not scored, and nothing is evicted"), pinned with a new `governance_card_empty_state_off_mode_is_disabled` test.
- **Round 2** on the updated diff: clean — *"did not identify any actionable regression in the changed code."*

Adversarial read confirmed: `Off` is an already-handled dashboard snapshot variant; gating `ingest_cost` on `Off` keeps default-`Off` leak-free (the reaper `tick` that would otherwise decay/prune the score map is also a no-op under `Off`); no production code relied on the old `DryRun` default; the detector's own logic tests still pass under explicit `DryRun`/`Enforce` (only the *default* changed).

**Risk tier: Light** — this disables a DryRun-only, non-acting detector (strictly safer than today) and corrects dashboard copy. No concurrency / crypto / wire-format / migration surface is touched.

**Local gates:** `cargo test -p freenet --lib governance` (60 passed) and `governance_card` (9 passed); `cargo clippy --locked -- -D warnings` clean; `cargo fmt` clean.

[AI-assisted - Claude]

PR #4343, opened 2026-06-03
feat(secrets): freenet secrets snapshot-list / snapshot-restore CLI

### LINKED ISSUE #4036: feat(delegates): restore-from-snapshot API and CLI for delegate secrets

sanity:
## Problem

PR #4034 added per-secret snapshot-on-write history for delegate secrets, and #4107 added the `list_snapshots` / `restore_snapshot` runtime APIs — but there was **no operator-facing way to inspect or roll back those backups**. #4036 tracks the CLI half. Without it, an operator who wants to see what snapshots exist for a delegate secret, or recover a secret a buggy delegate corrupted, has to spelunk the on-disk `.snapshots/` tree by hand.

## Solution

Two new `freenet secrets` subcommands, co-located with the existing `kek-*` verbs (same `--secrets-dir`, node-stopped model):

```
freenet secrets snapshot-list     --secrets-dir <path> [--delegate <key>] [--secret <id>]
freenet secrets snapshot-restore  --secrets-dir <path> --delegate <key> --secret <id> \
                                  --timestamp-ms <ms> --yes
```

- **`snapshot-list`** walks the secrets tree and reports per-secret snapshot history (count + each snapshot's `timestamp_ms`, UTC time, size). **Metadata only — it never decrypts or prints plaintext** (that would need the node KEK and would expose secrets on the terminal). Read-only.
- **`snapshot-restore`** rolls a secret back to a chosen snapshot. The current active value is snapshotted first, so the restore is itself reversible. The node MUST be stopped (it writes the active secret file).

**Key design points**

- Both commands need **neither the ReDb index nor the node KEK**: listing reads filenames; restore is a byte-level copy (the restored ciphertext stays decryptable by whatever KEK-derived DEK wrote it). So the commands fit the existing `freenet secrets --secrets-dir`-only pattern.
- A `SecretsId` **cannot be reconstructed from the on-disk `bs58(hash)` directory name** — the pre-image `key` bytes aren't persisted, and stdlib has no `from_hash` constructor. The commands therefore operate on the encoded id strings (copied from `snapshot-list` output), which is also why the shared core is keyed on `secret_encoded: &str`.
- **Shared core (DRY on a durability-critical op):** the byte-level restore + reversible "snapshot the active value first" logic is extracted into `secret_snapshots::{restore_snapshot_file, snapshot_active_value}`. `SecretsStore::{restore_snapshot, snapshot_prior_value}` become thin wrappers (the in-memory/ReDb index-repair edge case stays in the runtime path). This guarantees the node runtime and the CLI cannot drift on the crash-safety-sensitive `tmp+fsync+rename` sequence. The two owner-only perms helpers in `secrets_store.rs` are made `pub(super)` so the sibling module reuses them.

Snapshots are **not portable across nodes** (each node's DEK is HKDF-derived from its own KEK, per #4146), so restore only makes sense on the node that owns the secrets — documented in `docs/secrets-at-rest.md`.

## Testing

- **6 new filesystem-core unit tests** (`secret_snapshots`): restore round-trip, reversibility (prior active value is snapshotted), unknown-timestamp → `NotFound`, missing snapshot dir → `NotFound`, unsuffixed-collision-wins disambiguation, and `snapshot_active_value` no-op on a missing active file.
- **9 new CLI tests** (`secrets_cmd`): list summary / single-delegate / detail / empty-dir / missing-dir / unknown-delegate; restore round-trip / without-`--yes` aborts (active preserved) / unknown-secret / unknown-timestamp.
- **All pre-existing `secret_snapshots` + `secrets_store` tests pass** (55 lib + 22 bin), confirming the `restore_snapshot` extraction is behavior-preserving.
- `cargo fmt` clean; `cargo clippy -p freenet --lib --bins -- -D warnings` clean.

Closes #4036

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/secrets_cmd.rs
+++ b/crates/core/src/bin/commands/secrets_cmd.rs
@@ -29,7 +29,7 @@ use chrono::{DateTime, Utc};
 use freenet::dev_tool::{
     KEK_SIZE, KekBackendKind, RestoreError, RetentionPolicy, build_backend_for, list_snapshots,
     load_from_backend, read_backend_marker, replace_backend_marker, restore_snapshot_file,
-    snapshot_dir_for_encoded, write_backend_marker,
+    snapshot_dir_for_encoded, thin_snapshots, write_backend_marker,
 };
 use sha2::{Digest, Sha256};
 use zeroize::Zeroizing;
@@ -626,17 +626,24 @@ async fn snapshot_restore(args: SnapshotRestoreArgs) -> Result<()> {
     // Byte-level restore via the shared core — the same code the node
     // runtime uses, so the durability discipline can't drift. The
     // reversibility snapshot is always enabled for the CLI (manual
-    // recovery should be undoable); default retention policy.
+    // recovery should be undoable).
     match restore_snapshot_file(
         &delegate_dir,
         &args.secret,
         args.timestamp_ms,
         args.suffix,
         true,
-        &RetentionPolicy::default(),
-        std::time::SystemTime::now(),
     ) {
         Ok(()) => {
+            // Bound the history (incl. the reversibility snapshot just
+            // added), matching the node runtime's default retention.
+            // Best-effort: the restore has already succeeded.
+            let snap_dir = snapshot_dir_for_encoded(&delegate_dir, &args.secret);
+            thin_snapshots(
+                &snap_dir,
+                &RetentionPolicy::default(),
+                std::time::SystemTime::now(),
+            );
             println!(
                 "Restored secret `{}` of delegate `{}` to snapshot {selector}.",
                 args.secret, args.delegate
--- a/crates/core/src/bin/commands/secrets_cmd.rs
+++ b/crates/core/src/bin/commands/secrets_cmd.rs
@@ -148,6 +148,13 @@ pub struct SnapshotRestoreArgs {
     /// `snapshot-list`).
     #[clap(long)]
     pub timestamp_ms: u64,
+    /// Collision suffix (the `suffix` column from `snapshot-list`), for
+    /// the rare case of multiple same-millisecond snapshots at one
+    /// timestamp. Omit to restore the unsuffixed snapshot at the
+    /// timestamp (the `-` row, and the only entry when there is no
+    /// collision); pass the number to target a specific collision row.
+    #[clap(long)]
+    pub suffix: Option<u32>,
     /// Skip the interactive confirmation prompt (for scripted use).
     #[clap(long)]
     pub yes: bool,
@@ -594,17 +601,23 @@ async fn snapshot_restore(args: SnapshotRestoreArgs) -> Result<()> {
         );
     }
 
+    // Human-readable selector for messages: "timestamp_ms=N" or
+    // "timestamp_ms=N suffix=K".
+    let selector = match args.suffix {
+        Some(s) => format!("timestamp_ms={} suffix={s}", args.timestamp_ms),
+        None => format!("timestamp_ms={}", args.timestamp_ms),
+    };
+
     if !args.yes {
         eprintln!(
             "About to restore secret `{}` of delegate `{}` to the snapshot at \
-             timestamp_ms={} in {}.\n\
+             {selector} in {}.\n\
              - The current value is snapshotted first, so this is reversible.\n\
              - The freenet node MUST be stopped (a running node may concurrently \
                write this secret).\n\
              Re-run with --yes to proceed.",
             args.secret,
             args.delegate,
-            args.timestamp_ms,
             args.secrets_dir.display()
         );
         bail!("aborted (no --yes)");
@@ -618,21 +631,22 @@ async fn snapshot_restore(args: SnapshotRestoreArgs) -> Result<()> {
         &delegate_dir,
         &args.secret,
         args.timestamp_ms,
+        args.suffix,
         true,
         &RetentionPolicy::default(),
         std::time::SystemTime::now(),
     ) {
         Ok(()) => {
             println!(
-                "Restored secret `{}` of delegate `{}` to snapshot timestamp_ms={}.",
-                args.secret, args.delegate, args.timestamp_ms
+                "Restored secret `{}` of delegate `{}` to snapshot {selector}.",
+                args.secret, args.delegate
             );
             Ok(())
         }
-        Err(RestoreError::NotFound(ts)) => bail!(
-            "no snapshot at timestamp_ms={ts} for secret `{}` of delegate `{}`. Run \
+        Err(RestoreError::NotFound(_)) => bail!(
+            "no snapshot at {selector} for secret `{}` of delegate `{}`. Run \
              `freenet secrets snapshot-list --secrets-dir {} --delegate {} --secret {}` to see \
-             available timestamps.",
+             available timestamps (and suffixes).",
             args.secret,
             args.delegate,
             args.secrets_dir.display(),
@@ -955,6 +969,7 @@ mod tests {
             delegate: delegate.to_string(),
             secret: secret.to_string(),
             timestamp_ms,
+            suffix: None,
             yes,
         }
     }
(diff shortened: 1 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4343

### Summary
- **PR Title:** feat(secrets): freenet secrets snapshot-list / snapshot-restore CLI
- **Type:** feat (CLI) + behavior-preserving refactor of a durability-critical path
- **CI Status (HEAD `bc45e5f8`):** Clippy / Fmt / Conventional-Commits / Rule-Lint / macOS-E2E / Snyk / CLA passing; Unit & Integration / Simulation / Windows / NAT still running at time of writing — merge gated on these going green.
- **Linked Issues:** Closes #4036
- **Review tier:** Full (cryptography / secrets-at-rest surface + durability-critical restore refactor)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four `freenet:*` subagents) + external Codex (gpt-5.5), plus three Codex re-review rounds on the fixed code. Bug-pattern check against `.claude/rules/bug-prevention-patterns.md`.
- **Reviewed HEAD:** original pass on `ee345d9c`; round-1 fixes `5bbe4803`; round-2 (suffix) `8970ffa6`; round-3 (thin order) `bc45e5f8` (current).

---

### Code-First Analysis
**Independent understanding:** Two new `freenet secrets` subcommands (`snapshot-list`, `snapshot-restore`) operating node-stopped on `--secrets-dir`, plus an extraction of the byte-level restore + reversibility-snapshot logic from `SecretsStore` into a shared `secret_snapshots` core so the runtime and CLI can't drift.
**Stated intent:** Matches — closes the CLI half of #4036.
**Alignment:** Strong. Verified `SecretsId::encode()` = `bs58(hash)`, no `from_hash` constructor exists, so keying the shared core on encoded id strings is the correct (and only) approach. Flags/docs are consistent (`--timestamp-ms`, `--secret requires --delegate`).
**Gap (addressed in `bc45e5f8`):** the extraction initially ran `thin_snapshots` inside `restore_snapshot_file` (before the index-repair block) vs. after it in the old inline code. Code-first flagged the reorder; Codex round-2 rated it P2 (a failed `store_secrets_index` in the rare remove-then-restore path could prune source history the old order preserved for retry). Rather than argue it benign, the fix makes the runtime path **byte-for-byte order-identical**: thinning was removed from the shared core (now find + reversibility + atomic-write only) and moved to the callers — `SecretsStore::restore_snapshot` thins LAST (after the index repair commits, the original order); the CLI thins after a successful restore.

---

### Testing Assessment
**Coverage level:** adequate → strong after fixes.

| Unit (shared core) | ✅ | round-trip, reversibility, NotFound (unknown ts + missing dir), unsuffixed-collision-wins, missing-active no-op; **+ restore-path 0o600/0o700 perms, + `snapshots_enabled=false` skip-reversibility+thin** |
| Unit (CLI) | ✅ | list summary/detail/empty/missing-dir/unknown-delegate; restore round-trip/without-yes/unknown-secret/unknown-timestamp; **+ arg traversal rejection, + detail-empty branch, + clap `requires` parse** |
| Integration (runtime parity) | ✅ | the pre-existing `secrets_store` restore tests (`restore_snapshot_replaces_active_value`, `_unknown_timestamp_errors`, `restore_after_remove_repopulates_index`, `_prefers_unsuffixed_collision`, `legacy_snapshot_survives_restore_and_get_secret`) are unchanged and still pass through the delegated path — the parity safety net |
| Simulation / E2E | N/A | filesystem CLI; no network/simulation surface |

**Regression test:** N/A (feat). All new tests are pure-`tempfile`, fast, no flakiness.
**Gaps closed since original review:** restore-path perms (testing #1), `snapshots_enabled=false` restore (testing #2), detail-empty branch (#4), clap-requires (#5). Remaining un-driven branch: `RestoreError::Io` arm (hard to trigger portably) — accepted; the rename-failure cleanup path is identical to the pre-existing `store_secret` code.

---

### Skeptical Findings
**Risk level:** low. No high-severity issues; durability refactor verified behavior-preserving by side-by-side diff against base.

| Path traversal via unvalidated `--delegate`/`--secret` (joined onto `--secrets-dir`); operator-only footgun, not privilege escalation | Medium | `secrets_cmd.rs` snapshot_list/restore | **Fixed** — `validate_component` rejects non-single-component ids (`/`, `\`, `..`); 2 new tests |
| Summary-view `unwrap_or_default()` swallows list I/O errors as "no history" | Low (also Codex P2 / code-first) | `secrets_cmd.rs` summary loop | **Fixed** — propagates with context, matching the detail view |
| "Node MUST be stopped" is doc-only, unenforced; fixed-name `<secret>.tmp` races under concurrent restore | Low | restore path | Accepted — identical to existing `kek-migrate`/`kek-rotate` posture; not a regression. A lockfile guard for all four verbs is a reasonable future hardening |
| `with_extension("tmp")` on a dotted id yields `secret.tmp` | Low | shared core | Mooted — bs58 ids never contain dots; arg validation now also blocks separators; no corruption (rename target is the correct active path); pre-existing behavior |
| **Restore could not target a specific collision suffix** — `snapshot-list` shows `(ts, 0/1/…)` rows but restore took only `--timestamp-ms` and picked unsuffixed-wins, so selecting a `.n` row silently restored the wrong ciphertext (Codex re-review round 2) | Medium | restore CLI + shared core | **Fixed** in `8970ffa6` — added `--suffix <n>` / `restore_snapshot_file` `suffix: Option<u32>`; `Some(n)` restores exactly `.n`, missing `.n` is `NotFound` (no silent downgrade); runtime API unchanged (passes `None`). 2 new tests |

---

### Big Picture Assessment
**Goal alignment:** yes. The `freenet secrets` placement (vs. the issue's original `fdev delegate` wording) is one of the two options the issue-prioritizer explicitly surfaced, justified because these verbs need only `--secrets-dir` (no ReDb/KEK) — exactly the existing `freenet secrets` pattern. The `S-needs-design` open questions (timestamp display, no partial restores, confirm gate) are resolved.
**Anti-patterns:** none. No `#[ignore]`, no weakened assertions, no commented-out code.
**Removed code/tests:** none — pure extraction; zero test removals; only two `fn → pub(super) fn` visibility bumps.
**Scope:** focused, single logical change, one issue. No overlapping open PRs touch these files.
**AI instructions / skills / hooks:** none affected. `AGENTS.md` points at `docs/secrets-at-rest.md` (updated in this PR) rather than enumerating subcommands; no `.claude/rules/*` reference the secrets CLI.

---

### Documentation
- Code docs: complete (all new public items have rustdoc; the `restore_snapshot` rustdoc now notes the thin/index ordering).
- User docs: `docs/secrets-at-rest.md` updated — synopsis block, a "Snapshot inspect / restore (#4036)" section, the non-portability note, the arg-validation rule, and the 2-year-`max_age`-prunes-old-restore note.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important) — all addressed
1. Validate `--delegate`/`--secret` against path traversal (skeptical M1). ✅ `5bbe4803`
2. Propagate `list_snapshots` errors in the summary view (Codex P2 / code-first / skeptical L2). ✅ `5bbe4803`
3. Test restore-path owner-only perms and the `snapshots_enabled=false` restore path (testing). ✅ `5bbe4803`
4. Make every listed collision-suffix row addressable by restore — `--suffix` (Codex re-review round 2). ✅ `8970ffa6`
5. Preserve the exact runtime thin/index order (thin after index-repair, not in the core) so a failed index repair can't prune retry history (Codex re-review round 3 / code-first). ✅ `bc45e5f8`

#### Consider (Suggestions)
1. A lockfile/PID guard would harden the "node must be stopped" contract for all four `freenet secrets` write verbs (out of scope here; consistent with existing kek verbs).
2. `RestoreError::Io` arm remains untested (hard to trigger portably).
3. Retitle/annotate #4036 to reflect the CLI-under-`freenet-secrets` resolution (issue body still describes the `fdev` surface) — posted as an issue comment.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `bc45e5f8`

Re-review after fixes — all findings are resolved across the original five passes and three Codex re-review rounds: skeptical M1 traversal; the agreed L2/P2 summary-error swallow; the testing perms / `snapshots_enabled=false` gaps; round-2 collision-suffix addressability (`--suffix`); round-3 thin/index ordering. With the round-3 fix the runtime `restore_snapshot` is now byte-for-byte order-identical to the pre-extraction inline code (shared core = find + reversibility + atomic write; callers own index-repair + thin), so the "behavior-preserving" claim is now literally true, not merely outcome-equivalent. The unchanged `secrets_store.rs` restore tests (incl. `restore_after_remove_repopulates_index`, which exercises the index-repair branch) still pass as the parity net. 60 lib + 27 bin tests pass; `cargo fmt` + `clippy -D warnings` clean. Gate remaining: green CI on the current HEAD before merge.

[AI-assisted - Claude]

sanity:
CI is now fully green on HEAD \`bc45e5f8\` (the SHA the review above covers): Unit & Integration, Simulation, Windows Check, NAT Validation, Clippy, Fmt, Claude Rule Review, macOS E2E, Snyk, CLA all pass. The review's only remaining gate (heavy CI jobs) is satisfied — ready for a merge decision.

[AI-assisted - Claude]

PR #4344, opened 2026-06-03
fix: make network_status::init() overwrite the global in place

sanity:
## Problem

`crates/core/src/node/network_status.rs::tests` fail intermittently under
parallel test execution. The visible symptom on `init_populates_snapshot`
(and its sibling `ring_stats_provider_round_trip`):

```
assertion `left == right` failed
  left: "0.1.0"
 right: "0.1.0-test"
```

Root cause: `pub fn init(...)` did
`let _ = NETWORK_STATUS.set(Arc::new(RwLock::new(status)))` where
`NETWORK_STATUS` is a `OnceLock`. `OnceLock::set` is **first-write-wins**, so
once any test (e.g. `test_failure_snapshot_rendering`, which inits with
`"0.1.0"`) initialized the process-global, every later test's
`init(.., "0.1.0-test")` was silently a no-op. The `OnceLock` is
process-global and outlives every individual test — the `TEST_GLOBAL_STATE_LOCK`
the tests serialize on cannot reset it. So later tests inherited the first
test's `NetworkStatus`: wrong version, leaked gateway failures, leaked op
stats. The outcome was order-dependent — passes in isolation, fails when run
alongside sibling `network_status` tests. Once one assertion panicked while
holding the shared mutex, the poisoned mutex produced a cascade of further
failures in the same run.

Introduced with PR #4297 (cites #3507 / #4297).

## Solution

When `NETWORK_STATUS` is already set, write the fresh `NetworkStatus` into the
existing `RwLock` **in place** instead of dropping it. Only the first call
performs the `OnceLock::set`. This makes `init()` idempotent-overwrite.

Why this is safe in production: `init()` runs exactly once at node startup.
Both call sites — `node.rs` (local mode) and `p2p_impl.rs` (network mode) —
run once at bring-up, and only one of them executes per process. Neither
relies on first-write-wins. (Verified by grepping all `network_status::init(`
call sites.)

Alternatives considered: swapping `OnceLock` for a resettable primitive, or a
test-only reset helper. Overwrite-in-place is the smallest change that fixes
the determinism without altering the production contract (the global is still
installed exactly once; only its contents are refreshed) and without adding a
test-only code path to production.

## Testing

- Reproduced the flake: running the full module (`cargo test -p freenet --lib --
  node::network_status::tests`) repeatedly, ~1 in 5 runs failed with 4 tests
  red (one real state-leak assertion + 3 poisoned-mutex cascade panics).
  The isolated `--exact ...::init_populates_snapshot` always passed.
- Added `init_overwrites_existing_global`: a deterministic regression test that
  calls `init()` twice with different versions and a recorded failure between
  the calls, asserting the second call's version, port, and (reset) failures.
  Confirmed it **fails** against the old first-write-wins `init()`
  (`left: "first-version", right: "second-version"`) and **passes** with the fix.
- With the fix, ran the full `node::network_status::tests` module 8x and the
  combined `network_status` + `connection_manager` global-touching tests 5x —
  all green every run.
- `cargo fmt` and `cargo clippy -p freenet --lib -- -D warnings` clean.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -255,10 +255,12 @@ pub enum FailureReason {
 
 /// Initialize the global network status tracker.
 ///
-/// In production `init()` runs exactly once at node startup, so the first
-/// call always wins. If it is called again (the in-process test harness
-/// re-initializes the global between tests), the new status is written
-/// **in place** into the existing `RwLock` rather than dropped.
+/// In production `init()` runs exactly once, at node startup, on one of the
+/// two mutually-exclusive bring-up paths (`run_local_node` or `run_node`),
+/// so the first call installs the tracker. If it is called again — which in
+/// practice only happens in this module's own unit tests, each of which
+/// re-initializes the global with its own port/version — the new status is
+/// written **in place** into the existing `RwLock` rather than dropped.
 ///
 /// The previous implementation relied on `OnceLock::set`, which is
 /// first-write-wins: a second `init()` silently became a no-op. Because the
@@ -269,6 +271,11 @@ pub enum FailureReason {
 /// That made the test outcomes order-dependent: they passed in isolation but
 /// failed intermittently under parallel execution. Overwriting in place makes
 /// `init()` deterministically reset the tracker on every call.
+///
+/// This refreshes only the `NetworkStatus` value. The separately-registered
+/// providers (`SUBSCRIPTION_PROVIDER`, `GOVERNANCE_PROVIDER`,
+/// `RING_STATS_PROVIDER`, `ROUTER`) live in their own statics with their own
+/// replace-on-set semantics and are intentionally left untouched here.
 pub fn init(listening_port: u16, gateway_addrs: HashSet<SocketAddr>, version: String) {
     let status = NetworkStatus {
         gateway_failures: Vec::new(),
@@ -291,10 +298,15 @@ pub fn init(listening_port: u16, gateway_addrs: HashSet<SocketAddr>, version: St
                 *guard = status;
             }
         }
-        // First call: install the tracker. The race where two threads both
-        // observe `None` is benign — `set` makes exactly one win, and the
-        // loser's status is identical-shaped fresh state that a subsequent
-        // call would overwrite anyway.
+        // First call: install the tracker. A concurrent first-init race (two
+        // threads both observing `None`) cannot occur in production — `init()`
+        // is called exactly once, on a single linear startup path — and is
+        // serialized by `TEST_GLOBAL_STATE_LOCK` in tests. If it ever did
+        // occur, `set` picks exactly one winner and the loser's status (freshly
+        // built, with no accumulated failures/peers/stats) is dropped. That is
+        // the one window where the "every call overwrites" contract would not
+        // hold, but it is unreachable by construction, so we don't pay to
+        // recover the loser's value here.
         None => {
             #[allow(clippy::let_underscore_must_use)]
             let _ = NETWORK_STATUS.set(Arc::new(RwLock::new(status)));
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -254,6 +254,21 @@ pub enum FailureReason {
 }
 
 /// Initialize the global network status tracker.
+///
+/// In production `init()` runs exactly once at node startup, so the first
+/// call always wins. If it is called again (the in-process test harness
+/// re-initializes the global between tests), the new status is written
+/// **in place** into the existing `RwLock` rather than dropped.
+///
+/// The previous implementation relied on `OnceLock::set`, which is
+/// first-write-wins: a second `init()` silently became a no-op. Because the
+/// `OnceLock` outlives any single test (it is process-global and cannot be
+/// reset by the `TEST_GLOBAL_STATE_LOCK` the tests serialize on), every test
+/// running after the first `init()` inherited the first test's
+/// `NetworkStatus` — wrong version, leaked gateway failures, leaked op stats.
+/// That made the test outcomes order-dependent: they passed in isolation but
+/// failed intermittently under parallel execution. Overwriting in place makes
+/// `init()` deterministically reset the tracker on every call.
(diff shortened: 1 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #4344

### Summary
- **PR Title:** fix: make network_status::init() overwrite the global in place
- **Type:** fix (test-determinism / global-state)
- **Linked Issues:** none (self-contained flaky-test fix; cites #4297 as the regression source)
- **Review tier:** Full (global mutable state / concurrency surface — always Full per the review policy)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four `freenet:*` subagents) + external Codex (gpt-5.5). Bug-pattern check against `.claude/rules/bug-prevention-patterns.md`.
- **Reviewed HEAD:** code reviewed at `3cc9f34b` (rebased onto post-0.2.69 main); doc-only clarifications applied in `6d4cdffe`.

---

### Code-First Analysis
**Independent understanding:** `init()` switches from `OnceLock::set` (first-write-wins → second call is a silent no-op) to: overwrite the existing `RwLock` in place when already initialized, else `set` on first call. This fixes order-dependent flakes where the first test's `NetworkStatus` (version/port/failures) leaked into every later test.
**Alignment:** Implementation matches intent precisely. Verified `SecretsId`-style facts: all readers (`get_snapshot`, `record_*`, `is_known_gateway`, …) re-fetch via `NETWORK_STATUS.get()` per call and never cache the inner `Arc`, so overwriting `*guard` on the same `Arc` is observed atomically by every subsequent reader. The two production call sites (`node.rs:2287` local mode, `p2p_impl.rs:133` network mode) are mutually exclusive and run once at startup.
**Gaps:** none of substance.

---

### Testing Assessment
**Coverage level:** adequate — the `fix:`-PR regression-test requirement is satisfied and **empirically verified by the testing reviewer**: reverting `init()` to the old body makes `init_overwrites_existing_global` FAIL (exact first-write-wins symptom: `second.version == "first-version"`), and the full old module produced 3 failures incl. `PoisonError` cascades; with the fix, 22/22 pass across 3 consecutive runs.

| Unit | ✅ | `init_overwrites_existing_global` — inits twice with a recorded failure between, asserts the 2nd call resets version, port, AND accumulated failures (full `*guard = status` replacement, not just scalar fields); holds `TEST_GLOBAL_STATE_LOCK`; unique ports 40001/40002 |
| Integration / Simulation / E2E | N/A | single-file diagnostics-global fix |

**Regression test:** present, valid, reproduces the bug (verified fails-old / passes-new).

---

### Skeptical Findings
**Risk level:** low. No high/medium findings. Skeptical classified this Light-tier in substance (diagnostics-only global, not concurrency-critical or wire-format).

| Concurrent first-init race: loser of `set()` discards its status, contradicting the "every call overwrites" doc claim (Codex P2) | Medium (Codex) / dismissed | **Dismissed — justified.** Four reviewers independently traced the call sites: `init()` runs exactly once per process on a single linear startup path (`bin/freenet.rs` → `run_local_node`/`run_node`), and tests serialize on `TEST_GLOBAL_STATE_LOCK`, so the race is **unreachable by construction**; if it ever occurred the loser's status is freshly-built (no accumulated failures/peers/stats), so nothing is lost. Adding race-recovery code (or a `Clone` derive for a `get_or_init`-based form) to an impossible path on a diagnostics global is not warranted. The over-claiming comment was corrected in `6d4cdffe` to document the unreachability honestly. |
| `init()` doc misattributed the re-init source to an "in-process test harness" | Low (skeptical) | **Fixed** in `6d4cdffe` — it's this module's own unit tests; the multi-node harness never calls `init()`. |
| Poisoned-lock branch silently no-ops | Low (code-first/big-picture) | Not a regression — matches the module-wide convention (11 other accessors swallow poison identically), and the writers never panic under the guard. No change. |
| Stale `"OnceLock — first caller wins"` comment in `connection_manager.rs:2428` | Low (testing) | **Fixed** in `6d4cdffe`. |

---

### Big Picture Assessment
**Goal alignment:** yes — textbook root-cause flaky-test fix; eliminates order-dependence at its source rather than `#[ignore]`/retry/loosen-tolerance.
**Anti-patterns:** none. No removed tests, no weakened assertions.
**Removed code:** none (purely additive: +78/-3, the only deletion is the old 2-line `OnceLock::set`).
**Scope:** focused, one logical change (plus the doc clarifications from this review). Satisfies "one logical change per PR".
**Load-bearing invariant:** no production code depends on first-write-wins or the inner `Arc` identity (verified across all readers + the 3 call sites). `init()` refreshes only `NetworkStatus`, not the separately-registered providers (now documented).
**AI instructions / skills / stale docs:** none affected beyond the two comments fixed here.

---

### Documentation
- `init()` rustdoc: now accurate (re-init source, race unreachability, provider scope).
- `connection_manager.rs:2428`: stale comment corrected.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None.

#### Consider (Suggestions) — all addressed in `6d4cdffe`
1. Correct the `init()` rustdoc's re-init-source attribution. ✅
2. Stop the first-init-race comment over-claiming; document its unreachability. ✅
3. Note `init()` refreshes only `NetworkStatus`, not the providers. ✅
4. Fix the stale `connection_manager.rs:2428` "first caller wins" comment. ✅

(Optional, not done: a `tracing::warn!` on the poisoned-lock path — big-picture's belt-and-suspenders idea; skipped to match house style, since poison cannot occur here.)

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** code at `3cc9f34b`, doc clarifications at `6d4cdffe`.

A correct, minimal, root-cause flaky-test fix with a regression test the testing reviewer empirically confirmed reproduces the bug. No high/medium findings from the four freenet reviewers; Codex's P2 (concurrent first-init race) is a specific, well-investigated justified dismissal (unreachable by construction + benign), and the doc that over-claimed has been corrected. `cargo fmt` + `clippy -D warnings` clean; the `network_status` module is green 22/22 across repeated runs on post-0.2.69 main. Gate remaining: green CI on the current HEAD (`6d4cdffe`) before merge.

[AI-assisted - Claude]

Issue #4350, opened 2026-06-04
SUBSCRIBE renewal outer-cancel (25s) shorter than per-attempt peer wait (60s) discards in-flight replies


sanity:
## Problem

SUBSCRIBE renewals are structurally guaranteed to be cancelled mid-flight whenever a peer takes more than ~25s to reply, wasting the in-flight subscribe attempt and producing a steady stream of "late reply on a closed receiver" drops.

`ring.rs::recover_orphaned_subscriptions` runs every `SUBSCRIPTION_RECOVERY_INTERVAL = 30s` and spawns each renewal task wrapped in an **outer cancel deadline of `renewal_deadline = 30s − 5s = 25s`**. But the renewal driver's per-attempt network wait (`drive_client_subscribe_inner` → `send_to_and_await`) uses **`OPERATION_TTL = 60s`**. Because the outer 25s deadline is shorter than the inner 60s wait, a renewal whose peer hasn't answered within 25s is killed by the outer cancel while its `Subscribe::Response` is still in flight. When the response finally arrives, it lands on the renewal task's already-dropped capacity-1 receiver.

This is benign for correctness — the home node registers the originator as a downstream subscriber *before* sending the reply (`register_downstream_subscriber` precedes the response send in `drive_relay_subscribe`), and the renewal self-heals on the next 30s cycle — but it is wasteful and noisy:

- Every slow-peer renewal burns a full attempt that is thrown away at 25s instead of completing.
- It is the dominant source of the `try_forward_driver_reply` closed-receiver drops (the false-alarm `error!` logs fixed in PR #4351, now `debug`). On the busiest gateway (nova) this ran ~30/hr, spiking after a restart when ~745 hosted contracts re-subscribe at once.

## Why this is a follow-up, not part of the log-fix PR

PR #4351 correctly reclassifies the *log level* of the benign drop (it is not an error). This issue is about the *underlying timing* that generates the drops in the first place. Changing renewal timeouts is a **behavioral change on a high-risk surface** (subscribe/ring) and must clear a higher bar than a logging fix:

- It risks abandoning slow-but-valid peers earlier (or later) than today, which could change renewal success rates and topology stability.
- It needs its own simulation coverage and before/after measurement, not just a unit test.

So the log-level fix shipped first; this fixes the root cause separately and deliberately.

## Proposed approach (to be validated)

[...]

PR #4351, opened 2026-06-04
fix(node): log benign closed-receiver driver-reply drops at debug, not error

sanity:
## Problem

While monitoring the v0.2.69 rollout, the gateways logged a steady stream of `ERROR` lines:

```
freenet::node: Failed to forward driver reply to OpCtx task err=channel closed op="subscribe"
```

~30/hr on the busiest gateway (nova), spiking right after a restart. Investigation showed this is **benign**: `try_forward_driver_reply` logged *every* `try_send` failure at `error!`, including the expected `TrySendError::Closed` case.

The dominant source is SUBSCRIBE renewals. `recover_orphaned_subscriptions` spawns renewal tasks with a ~25s outer cancel deadline, but the per-attempt `send_to_and_await` peer wait is ~60s (`OPERATION_TTL`). So a renewal whose peer is slow is cancelled (its capacity-1 receiver dropped) while the reply is still in flight; when the reply lands, `try_send` fails `Closed`. The subscription is already registered on the home node (`register_downstream_subscriber` runs before the response is sent) and self-heals on the next 30s cycle — this is exactly the "reply arrives with no waiter → benign → debug log" case `.claude/rules/operations.md` already documents. After a restart, ~745 hosted contracts re-subscribe at once, which is why the rate spikes.

## Solution

Split the two `try_send` failure modes in `try_forward_driver_reply`:

- `TrySendError::Closed` (benign — caller finished / cancelled / timed out) → `debug!`
- `TrySendError::Full` (receiver still open but the capacity-1 slot already holds an undrained reply = a genuine double-forward bug) → `warn!`

No behavior change beyond log severity. The `Full` case is kept loud because it would indicate a real bug.

## Testing

- New source-scrape pin test `forward_driver_reply_splits_log_level_by_trysenderror_variant` asserts the function matches on both `TrySendError` variants and never logs at `error!`. **Verified it fails on the prior `error!`-only body and passes with the fix**, so a future refactor cannot silently re-collapse to a single `error!` and bring the spam back. (Matches the existing `include_str!` source-scrape pin idiom in this file; needles assembled at runtime so it can't self-match.)
- The `Full`-arm behavioral contract (drop, never block, return `true`) remains covered by the existing `bypass_does_not_block_when_channel_already_full`.
- `cargo fmt --check`, `cargo clippy -p freenet`, and the full `callback_forward_tests` module pass.

## Follow-up

This PR only corrects the misleading log level. The *root cause* of the drops — the SUBSCRIBE renewal outer-cancel (~25s) being shorter than the per-attempt peer wait (~60s), which guarantees slow-peer renewals are killed mid-await and discard an in-flight reply — is a behavioral change on a high-risk surface (subscribe/ring) and is tracked separately in #4350.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -804,25 +804,59 @@ where
 /// # Channel safety
 ///
 /// Uses `try_send` on the bounded capacity-1 channel created by
-/// `OpCtx::send_and_await`. On a closed receiver (caller cancelled or
-/// timed out) the send fails and is logged; the handler still makes
-/// progress. See `.claude/rules/channel-safety.md`.
+/// `OpCtx::send_and_await`. The two `try_send` failure modes mean very
+/// different things and are logged at different levels:
+///
+/// - `TrySendError::Closed` is **benign and expected**: the OpCtx caller
+///   already finished — a capacity-1 receiver hangs up after its single
+///   `recv()`, or the driver task timed out / was cancelled. The dominant
+///   source is SUBSCRIBE renewals, whose ~25s outer cancel deadline fires
+///   before the ~60s per-attempt peer wait, so a peer's reply routinely
+///   lands after the renewal task (and its receiver) was dropped. The
+///   subscription is already registered on the home node and self-heals on
+///   the next renewal cycle, so this is logged at `debug`, matching
+///   `.claude/rules/operations.md` ("WHEN a reply arrives with no waiter →
+///   Benign → debug log"). Logging it at `error` produced a steady stream
+///   of false-alarm errors on busy gateways (especially after a restart,
+///   when hundreds of hosted contracts re-subscribe at once).
+/// - `TrySendError::Full` is **unexpected**: the receiver is still open but
+///   the capacity-1 channel already holds an undrained reply — two replies
+///   were forwarded for the same tx before the driver consumed the first.
+///   That is a real double-forward bug, so it stays loud at `warn`.
+///
+/// Either way the handler still makes progress and returns `true`. See
+/// `.claude/rules/channel-safety.md`.
 fn try_forward_driver_reply(
     pending_op_result: Option<&tokio::sync::mpsc::Sender<NetMessage>>,
     reply: NetMessage,
     op_label: &'static str,
 ) -> bool {
+    use tokio::sync::mpsc::error::TrySendError;
+
     let Some(callback) = pending_op_result else {
         return false;
     };
     let tx_id = *reply.id();
-    if let Err(err) = callback.try_send(reply) {
-        tracing::error!(
-            %err,
-            %tx_id,
-            op = op_label,
-            "Failed to forward driver reply to OpCtx task"
-        );
+    match callback.try_send(reply) {
+        Ok(()) => {}
+        // Benign late/duplicate reply: the OpCtx caller already completed,
+        // timed out, or was cancelled (see the `# Channel safety` note).
+        Err(TrySendError::Closed(_)) => {
+            tracing::debug!(
+                %tx_id,
+                op = op_label,
+                "Driver reply arrived after the OpCtx receiver closed (late/duplicate); dropping"
+            );
+        }
+        // Unexpected double-forward: receiver open but capacity-1 channel
+        // already holds an undrained reply for this tx. A real bug — stay loud.
+        Err(TrySendError::Full(_)) => {
+            tracing::warn!(
+                %tx_id,
+                op = op_label,
+                "Driver reply dropped: OpCtx channel full (capacity-1 already holds an undrained reply for this tx)"
+            );
+        }
     }
     true
 }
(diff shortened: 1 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #4351

### Summary
- **PR Title:** fix(node): log benign closed-receiver driver-reply drops at debug, not error
- **Type:** fix (observability / log-severity)
- **CI Status:** running on HEAD `e5030a34` at review time (Fmt / Conventional Commits / Rule Lint green; full suite in progress)
- **Linked Issues:** follow-up #4350 (renewal timeout root cause); not "Closes"
- **Review tier:** Light (no concurrency/wire/state-machine *logic* change; single-purpose, two files)
- **Reviewers run:** skeptical, testing, big-picture, + external Codex — then a re-review pass (skeptical + Codex) on the revised HEAD

---

### Code-First Analysis
**Independent understanding:** `try_forward_driver_reply` previously logged every `try_send` failure on the OpCtx reply channel at `error!`. The PR downgrades the benign, intentionally-lossy drop to `debug!`.
**Stated intent:** remove ~30/hr false-alarm `ERROR` logs on busy gateways (dominated by SUBSCRIBE renewals whose ~25s outer cancel fires before the ~60s per-attempt peer wait) without losing real signal.
**Alignment:** matches. Pure observability change; control flow, the `true` return, and the non-blocking `try_send` are unchanged.

---

### Testing Assessment
**Coverage:** adequate and proportionate for a log-level change.

| Unit | ✅ | New `forward_driver_reply_logs_benign_drop_at_debug_only` source-scrape pin; existing behavioral tests still cover drop/non-block/`return true` for both Closed and Full arms |
| Integration | N/A | leaf-helper log change |
| Simulation | N/A | — |
| E2E | N/A | — |

**Regression test:** present and valid. The pin test asserts the function body logs at `debug` and contains neither `error!` nor `warn!`, window-bounded to the function's own closing brace; needles assembled at runtime so it can't self-match. Verified **red on the prior `error!` body, green with the fix**. `tracing-test` is not in the workspace; the source-scrape pin matches this module's established `include_str!` idiom and avoids a flaky global-subscriber capture test.

---

### Skeptical Findings
**Risk level:** low (after revision).

| `Full`→`warn` "double-forward bug" message is false for CONNECT's capacity-N fan-in (`compute_reply_capacity` ≈ 50); `Full` legitimately occurs there with distinct replies under load, re-introducing false-alarm noise | Should Fix | node.rs `try_forward_driver_reply`; connect/op_ctx_task.rs:529 | **Fixed** in `e5030a34` |
| Stale comment on `compute_reply_capacity` said over-capacity replies "drop with an error log" | Should Fix | connect/op_ctx_task.rs:529 | **Fixed** |
| Pin test didn't bind each macro to its arm (arm-swap would pass) | Consider | node.rs pin test | **Fixed** (collapsed to one log line → no arms to mis-map) |
| Pin-test window included the next fn's doc comment | Consider | node.rs pin test | **Fixed** (window now bounded to the function's own closing brace) |

**Resolution:** collapsed to a single `debug!` carrying `%err` (which still distinguishes `Closed` vs `Full`). `Full` is benign across *all* callers — a capacity-1 duplicate or a capacity-N CONNECT overflow, both lossy-by-design per `channel-safety.md`. Re-review (skeptical + Codex) on `e5030a34` traced all four call sites and the terminal-error race (operations.md) and confirmed no path needs a loud log and no regression was introduced.

---

### Big Picture Assessment
**Goal alignment:** yes — improves signal quality (kills a false-alarm `error`) without silencing real bugs, and correctly defers the behavioral root cause (renewal 25s-vs-60s timeout) to #4350 with a clear high-risk-surface justification. Textbook *non*-CI-chasing change.
**Anti-patterns:** none. No removed tests/fixes; the new pin test *strengthens* the guard. Diff is one focused logical change.
**Docs/rules:** consistent with `operations.md` ("reply with no waiter → benign → debug") and `channel-safety.md` (drop-when-full is intended); no rule update required.

---

### Documentation
- Code docs: complete — the `# Channel safety` doc comment was rewritten to accurately describe both `Closed` and `Full` (incl. CONNECT capacity-N overflow). Stale CONNECT comment corrected.
- Architecture / user docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
- None.

#### Should Fix (Important)
- None remaining (both were addressed in `e5030a34`).

#### Consider (Suggestions)
- None remaining.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `e5030a34014528be4918126a059959081c3e17f5`

Both review passes are clean on the current HEAD. Initial review found one real Should-Fix (the `Full`→`warn` mischaracterization for CONNECT's capacity-N channel) plus minor items; all were fixed and re-reviewed (skeptical + Codex) on `e5030a34` with no remaining findings. Merge once CI is green on this SHA.

[AI-assisted - Claude]

Issue #3431, opened 2026-03-05
fix: GET/SUBSCRIBE routing fails when contract is only cached at a few nodes (low HTL)


iduartgomez:
## Problem

When a contract is PUT with low HTL (e.g., 3), only a handful of nodes along the PUT route cache the contract code+state. Subsequent GET or SUBSCRIBE operations from nodes whose connected peers don't overlap with those caching nodes fail with "No other peers found" → NotFound.

This was observed in simulation testing with 16 nodes (1 gateway + 15 nodes, HTL=3): PUT cached at 4 nodes, and node-13 consistently failed GET with `fetch_contract=true` because its routing candidates couldn't reach any caching node.

### Root Cause

`k_closest_potentially_caching` returns candidates sorted by proximity to the contract location, but those candidates may not have the contract cached. The GET hops between non-caching nodes, the visited peers bloom filter accumulates, and eventually all candidates are exhausted without finding a caching node.

### Reproduction

See diagnostic test `test_put_code_propagation_diagnostic` in `simulation_integration.rs` (on branch `investigate/put-code-propagation`):
- 1 gateway + 15 nodes, HTL=3, max_connections=6, min_connections=3
- Gateway PUTs contract, 12 nodes subscribe, gateway updates, then all nodes GET with `fetch_contract=true`
- Node-13 consistently fails (seed `0xC0DE_B0CA_0001`)

### Observations

- PUT caches at exactly 4 nodes (gateway + 3 hops)
- 1 subscribe also fails ("No closer peers to forward subscribe request")
- The `visited` bloom filter (`set_bits: 19, total_bits: 512`) accumulates across hops, blocking revisits
- **This is distinct from the readiness gating issue** fixed in #3425 — the fallback to not-ready peers doesn't help when no reachable peer has the contract

### Investigation Findings

During investigation of #3423 Phase 3, we confirmed:
1. **The "state without code" death spiral hypothesis was wrong** — UPDATE cannot store state at nodes without existing code (both mock and production runtimes reject with `MissingContract`)
2. **The real failure mode is routing exhaustion** — GET can't find a path to any caching node
3. **All nodes with state also have code** — PUT caches both, and UPDATE only works at nodes that already have the contract

### Possible Fixes

1. **Increase PUT cache radius** — ensure more nodes cache during PUT (e.g., broadcast to all neighbors, not just along the hop path)
2. **Retry with expanded search** — when GET hits NotFound, expand the candidate set beyond `k_closest_potentially_caching`
[...]

Issue #3362, opened 2026-03-01
Gateway acts as permanent routing hub, not just bootstrap — star topology instead of ring


stephensong:
## Summary

In a multi-node local network, the gateway node becomes the permanent central routing hub rather than just bootstrapping the ring topology. Killing the gateway after ring formation collapses the entire network — remaining nodes cannot route to each other. This makes the gateway a single point of failure for all operations, not just initial peer discovery.

### Setup
- 7-node local network: 1 gateway (port 3001/3101) + 3 regular nodes (3002-3004) + 3 guardian nodes (3005-3007)
- Built from `fix/combined-fixes` branch (v0.1.160 + PRs #3202, #3206)
- All nodes started with `--skip-load-from-network`

### Test 1: Kill gateway after ring established

After running integration tests (confirming PUT/GET/Subscribe all work across nodes), we killed the gateway process. Result:

**Immediate routing failures across ALL remaining nodes:**
```
Reported operation timeout as routing failure tx=... peer=Some(127.0.0.1:3102) contract_location=0.639
Reported operation timeout as routing failure tx=... peer=Some(127.0.0.1:3103) contract_location=0.466
Reported operation timeout as routing failure tx=... peer=Some(127.0.0.1:3101) contract_location=0.468  ← still trying dead gateway
```

A simple PUT to port 3002 (not the gateway) timed out after 90 seconds. The remaining 6 nodes were completely unable to route.

### Test 2: Ring connection counts

Event loop stats show the topology is a star, not a ring:
```
peer=3MUKojSV8bcowEJUo: ring_connections=1-2, active_connections=1-2
peer=44vZ7sLtbiHvbtUQr: ring_connections=5, active_connections=5
```

One node (likely the gateway) has 5 ring connections. Other nodes have only 1-2, and those connections go through the gateway.

### Test 3: Node-3 (port 3004) instability

Node-3 consistently fails subscription operations. The connection manager repeatedly removes its connections as stale:

```
WARN connection_maintenance: Removing stale pending reservation addr=127.0.0.1:3104 age_secs=60
WARN connection_maintenance: Removing orphaned location_for_peer entry addr=127.0.0.1:3104 location=0.668
```

Every subscribe request from node-3 fails:
```
WARN Subscribe: could not find peer to register interest contract=... requester_addr=127.0.0.1:3104
```

This produces a cascade: no subscriptions → no notifications → update propagation finds NO_TARGETS:
```
[...]

PR #4352, opened 2026-06-04
fix(ring): restore connection convergence in nightly sim tests (#4348)

### LINKED ISSUE #4348: Nightly simulation tests red since 2026-03-26: test_nightly_* never reach MIN_CONN=10 (likely #3362)
## Problem

The **Simulation Tests (Nightly)** workflow (`.github/workflows/simulation-nightly.yml`) has failed on `main` **every night since 2026-03-26** — roughly 10 weeks. Because it has been red continuously, it no longer provides signal: a genuine new regression caught by the nightly is indistinguishable from the standing failure.

Last green nightly: **2026-03-25** (run `23523957765`).

This is **not** a release blocker (it is a scheduled job, not a per-commit/PR gate — recent releases including v0.2.68 and v0.2.69 shipped while it was red), but the standing red baseline should be resolved so the nightly is trustworthy again.

## Two distinct stacked failures

**Phase 1 — 500-node large-scale step timeout (2026-03-26 → 2026-04-06)**
The "Large scale test (500 nodes)" step (`fdev test ... --min-success-rate 1.0`) began failing (exit 1 / step `cancelled` at the 60-min timeout). During this window the `nextest` suite itself still passed (105/105).

**Phase 2 — `test_nightly_*` topology assertions (2026-04-07 → present)**
PR #3744 (commit `81290575`, closes #3511) added three connectivity tests with stronger assertions. They have failed on **every** nightly run since the day they merged — they appear to have never passed in CI. The `nextest` step now exits 100 before the 500-node step runs, so Phase 1's step is now masked (never reached).

The three failing tests in `crates/core/tests/simulation_integration.rs` (config: `GATEWAYS=4, NODES=50, MIN_CONN=10`):

- `test_nightly_50_node_topology_formation` (~line 9017):
  `Topology formation stall: median=6 < min_connections=10`
- `test_nightly_connection_growth_checkpoints` (~line 9167):
  `Final median=7 < min_connections=10 after 60 virtual minutes. Checkpoints: [(300, 7), (900, 7), (1800, 7), (3600, 7)]`
- `test_nightly_fault_recovery_speed` (~line 9265):
  `Network did not converge before fault injection: median=8 < min_connections=10` (fails before fault-recovery logic runs)

Some nights `test_get_reliability_diagnostic` (~line 8441) also fails, so the failing count is "3 or 4," not always exactly 3.

Latest run summary (`26930797562`): `116 tests run: 113 passed (1 slow), 3 failed, 2 skipped`. The simulated 50-node/4-gateway network plateaus at a **median of 6–8 connections** and never reaches the required `MIN_CONN = 10`.

## Analysis

- **Deterministic, not flaky.** Runs from 2026-05-30, 06-01, 06-03, 06-04 fail at the same source lines with median 6–8 (±1 run-to-run as the network changes), never ≥10. Not a timeout/OOM/runner issue.
- The checkpoint series `[(300,7),(900,7),(1800,7),(3600,7)]` shows connectivity is **flat for the entire simulated hour** — no growth. That points toward a real convergence ceiling rather than a merely slightly-too-high threshold.
- This is consistent with **#3362** (OPEN, P-high — "Gateway acts as permanent routing hub, not just bootstrap — star topology instead of ring"), which describes the same topology-degradation class. The nightly tests may be the standing reproduction of #3362.

## What needs deciding (per test)

For each `test_nightly_*` assertion, determine whether:
1. it is catching a **genuine topology/convergence bug** (likely the same root cause as #3362), in which case fix the underlying behavior; or
2. the `MIN_CONN` / median threshold was set **aspirationally** beyond what the single-process 50-node/4-gateway harness can achieve, in which case recalibrate to a realistic bound.

The flat-checkpoint evidence leans toward (1) for at least the connection-growth test.

Separately, restore or re-scope the Phase-1 500-node large-scale step so it is reachable and meaningful once `nextest` is green again.

## References

- Last green nightly: run `23523957765` (2026-03-25)
- Representative failing run: `26930797562` (2026-06-04)
- Tests introduced by: #3744 (commit `81290575`), closing #3511
- Likely underlying bug: #3362
- Key file: `crates/core/tests/simulation_integration.rs` (lines ~9017, ~9167, ~9265, ~8441)
- Workflow: `.github/workflows/simulation-nightly.yml`

## Goal

Get the nightly back to green (or to a state where a failure is a real new signal), so it stops masking regressions. Either fix the convergence/topology root cause or recalibrate the thresholds to what the harness can actually achieve — and document the rationale either way.

[AI-assisted - Claude]

sanity:
## Problem

The **Simulation Tests (Nightly)** workflow's `test_nightly_*` connectivity tests plateaued at a flat median of ~7 connections, never reaching `min_connections = 10`:

```
Final median=7 < min_connections=10. Checkpoints: [(300,7),(900,7),(1800,7),(3600,7)]
```

Flat checkpoints across the whole simulated hour = the network bootstraps, then **stops acquiring new connections**.

### Root cause (bisected)

These tests are not aspirational — CI ran them green at **median 16** in April (run `24116882646`). Bisecting on a fixed seed pinpoints the **#1454 CONNECT task-per-tx migration** (mid-May), which made the new drivers hold CONNECT transactions far longer than the old mediator did:

| Commit | Trajectory | Result |
|--------|-----------|--------|
| `ca7fa7a0` (pre-#1454) | 12→17→17→19 grows | PASS |
| `d90a6c09` (post-#1454) | 9→9→9→9 flat | FAIL |
| `HEAD` (pre-fix) | 7→7→7→7 flat | FAIL |

`Ring::connection_maintenance` throttles concurrent connection-*acquisition* attempts on `LiveTransactionTracker`'s live CONNECT count (`active_count >= max_concurrent` → stop calling `acquire_new`). That count summed `tx_per_peer`, which includes CONNECTs the node is merely **relaying** for other peers (held until completion / the 60s `OPERATION_TTL`) and double-counts a CONNECT rebound across hops. Once at/above `min_connections`, `calculate_max_concurrent_connections` returns `BASE = 3`, so a relay-heavy node's own acquisition budget was consumed by others' traffic and it stopped acquiring — topology stalled flat.

A secondary effect left a residual straggler tail: nodes that route into saturated ring regions get `Rejected`, stamp 30s→600s **location backoff**, and the maintenance loop then skips those locations — trapping a poorly-positioned node below min.

## Approach

1. **Throttle on self-initiated acquisitions only.** Track CONNECTs the node initiates itself in a dedicated `acquisition_txs` set and throttle on `active_acquisition_transaction_count()`. Registration happens once in `start_client_connect` — the single entry for every originator CONNECT (ring acquisition, gateway join, version probe) — so relayed CONNECTs (`start_relay_connect`) and the rebind double-count are excluded, while all genuinely self-initiated CONNECTs still count toward the storm-safety cap. The gauge is cleared only by `remove_finished_transaction` (completion / orphan-wake / 60s TTL), never by prune — race-free and erring toward over-counting during churn (the safe direction).

2. **Under-min backoff escape.** Honor target-location backoff only once at/above `min_connections` (`should_respect_location_backoff`). A node below min keeps probing recently-rejected regions — its under-connection is a local capacity problem, not the target's fault (cf. ring.md "backoff target must match failure cause"). Mirrors the existing zero-connection re-bootstrap escape; above-min steady-state backoff is unchanged. Storm-bounded by the adaptive fast-tick backoff + `max_concurrent`.

The deeper structural cause of the residual straggler tail (no capacity reservation for under-min joiners when peers grow to `max_connections` — the gateway/hub-imbalance class) is tracked by **#3362** and out of scope here.

## Testing

Local (`--features "simulation_tests,testing,nightly_tests" --test-threads=1`):

| Test | Before | After |
|------|--------|-------|
| `test_nightly_connection_growth_checkpoints` | flat median 7 (FAIL) | median 18–19, **PASS** |
| `test_nightly_50_node_topology_formation` | flat median 7 (FAIL) | median 18, **93–98% reach min**, **PASS** |
| `test_nightly_fault_recovery_speed` | flat median 7 (FAIL) | median 18, 0% isolated, **PASS** |
| `test_six_peer_contract_lifecycle` | PASS | PASS (no regression) |
| `test_connection_growth_stall_regression` | PASS | PASS |

New unit tests (`live_tx.rs`, `ring.rs`): acquisition count excludes relayed CONNECTs, is unaffected by cross-peer rebind, drains on completion, is unaffected by prune; under-min backoff predicate boundary tests. All fail on the pre-fix behavior, pass on the fix.

**Sim-test redesign:** `test_get_routing_coverage_low_htl`'s relay-hop *route-event* assertions were extracted to a new `test_relay_route_events_multihop` on a deliberately **sparse** topology. The coverage test needs a *dense* mesh (so every node's GET resolves, #3431); once this fix lets nodes actually converge, that mesh resolves GETs in 0–1 hops and the route-event counters never fired — i.e. those assertions were implicitly depending on the under-connectivity bug this PR fixes. The new test asserts only that the route-event hooks fire (a NotFound reply counts), so improved connectivity can't invalidate it; the coverage test keeps #3431 + the relay driver-call assertions.

`cargo fmt` clean; `cargo clippy --locked -- -D warnings` clean.

Why didn't CI catch the regression earlier? The nightly job was already red for an unrelated, older reason (the 500-node large-scale step), so the `test_nightly_*` failure was just another red — exactly the masking this issue is about. The `test_nightly_*` tests are cron-only (not PR-gating); PR CI exercises the fix via `test_get_routing_coverage_low_htl`, `test_relay_route_events_multihop`, and the `live_tx`/`ring` unit tests.

## Note: separate Phase-1 500-node step

The nightly's "Large scale test (500 nodes)" step is a separate, older failure (timing out since ~2026-03-26), currently masked because nextest fails first. After this fix nextest passes and that step becomes reachable; the same throttle bug plausibly slowed it, so it may now pass — if not, track/re-scope separately as the issue notes.

## Review

Full multi-model review (A-networking, high-risk): four Claude perspective reviewers (code-first, skeptical, testing, big-picture) + an external Codex pass. All findings addressed — see the consolidated review comment below.

Closes #4348

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -93,7 +93,17 @@ CORRECT:
 WHY: Backing off a target for a local problem prevents connecting to it
 later when local conditions improve (e.g., more connections acquired).
 
-See: ring.rs connection_maintenance, issue #3414
+COROLLARY (under-min backoff escape, #4348):
+  A node still BELOW min_connections must IGNORE per-target-location
+  backoff entirely and keep probing. Its under-connection is by definition
+  a local capacity problem, so a `Rejected`-stamped 30s→600s location
+  backoff would trap a poorly-positioned node permanently below min.
+  connection_maintenance gates this via should_respect_location_backoff()
+  (honor backoff only at/above min). Storm safety is preserved by the
+  adaptive fast-tick backoff + max_concurrent cap, not by location backoff.
+
+See: ring.rs connection_maintenance + should_respect_location_backoff,
+issues #3414, #4348
 ```
 
 ### Routing Decisions
--- a/crates/core/src/operations/connect/op_ctx_task.rs
+++ b/crates/core/src/operations/connect/op_ctx_task.rs
@@ -136,6 +136,16 @@ pub(crate) async fn start_client_connect(
     desired_location: Location,
     overall_timeout: Option<std::time::Duration>,
 ) -> Result<(), OpError> {
+    // Count this CONNECT toward the connection_maintenance acquisition throttle.
+    // `start_client_connect` is the single entry for every *self-initiated*
+    // (originator) CONNECT — ring acquisition (`Ring::acquire_new`), gateway
+    // joins (`join_ring_request`), and version probes (`gateway_version_probe`)
+    // all route through here — so registering once here keeps the throttle's
+    // storm-safety cap honest across all of them. Relayed CONNECTs go through
+    // `start_relay_connect` and are deliberately NOT counted (#4348). Registered
+    // before the first send below so it precedes any reply or completion.
+    op_manager.ring.live_tx_tracker.register_acquisition(tx);
+
     // Snapshot whether the joiner knows its own external address before
     // contacting the gateway. Mirrors legacy `JoinerState`'s
     // `started_without_address` field at connect.rs:1162. Used at the
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -3021,8 +3021,17 @@ impl Ring {
             // node's own in-flight acquisitions, NOT CONNECTs it is relaying for
             // others (#4348).
             let mut active_count = live_tx_tracker.active_acquisition_transaction_count();
+            // Honor target-location backoff only once we've reached
+            // min_connections. A node still below min MUST keep probing even
+            // recently-rejected ring regions: its under-connection is a local
+            // capacity problem, not the target's fault, and the 30s→600s
+            // location backoff stamped on a `Rejected` would otherwise trap a
+            // poorly-positioned node permanently below min — the straggler tail
+            // in #4348. This mirrors the zero-connection re-bootstrap escape
+            // above, extended to the whole under-min regime.
+            let respect_backoff = current_conn_count >= self.connection_manager.min_connections;
             while let Some(ideal_location) = pending_conn_adds.pop_first() {
-                if self.is_in_connection_backoff(ideal_location) {
+                if respect_backoff && self.is_in_connection_backoff(ideal_location) {
                     tracing::debug!(
                         target_location = %ideal_location,
                         "Skipping connection attempt - target in backoff"
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -3448,14 +3448,10 @@ impl Ring {
 
         // Register tx with the live transaction tracker BEFORE spawning the
         // driver. Otherwise the driver's first Response could land on the
-        // bypass before the registration completes, breaking the
-        // acquisition gauge that connection_maintenance uses to throttle
-        // concurrent acquisitions.
+        // bypass before the registration completes. (The acquisition-throttle
+        // registration is done inside `start_client_connect`, shared by every
+        // self-initiated CONNECT — ring acquisition, gateway join, and probe.)
         live_tx_tracker.add_transaction(gateway_addr, tx);
-        // Mark this as a self-initiated acquisition so it (and only it, not
-        // relayed CONNECTs) counts toward the maintenance concurrency throttle
-        // (#4348).
-        live_tx_tracker.register_acquisition(tx);
 
         let op_manager_spawn = op_manager.clone();
         let gateway = query_target;
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -3017,8 +3017,10 @@ impl Ring {
             );
 
             // Drain pending connections, initiating multiple attempts per tick
-            // (up to max_concurrent) for faster mesh formation.
-            let mut active_count = live_tx_tracker.active_connect_transaction_count();
+            // (up to max_concurrent) for faster mesh formation. Counts only this
+            // node's own in-flight acquisitions, NOT CONNECTs it is relaying for
+            // others (#4348).
+            let mut active_count = live_tx_tracker.active_acquisition_transaction_count();
             while let Some(ideal_location) = pending_conn_adds.pop_first() {
                 if self.is_in_connection_backoff(ideal_location) {
                     tracing::debug!(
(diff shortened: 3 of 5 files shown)


sanity:
## Consolidated multi-model review

Full-tier review (A-networking / concurrency — high-risk surface): four parallel Claude perspective reviewers + an external Codex pass. All findings addressed.

### Claude reviewers

**code-first** — Implementation correct on its own terms; `acquisition_txs` lifecycle is leak-equivalent to the existing `tx_per_peer` accounting; throttle math conservative under concurrency; backoff gate boundary correct. Flagged the branch was behind `main` (stale-diff) → **rebased onto origin/main**.

**skeptical** — Verified no acquisition-set leak, no telemetry loss from removing `active_connect_transaction_count`, determinism unaffected (`.len()` only). Findings: (a) the backoff-bypass behavioral change had no fast test → **added `should_respect_location_backoff` + boundary unit tests**; (b) churn-storm concern → mitigations documented (adaptive fast-tick backoff + `max_concurrent` cap, both unchanged); (c) stale-branch revert hazard → **rebased**.

**testing** — New `live_tx` unit tests verified to fail on the old impl (real regression tests). Gap: backoff-bypass lacked blocking-CI coverage → **extracted predicate + unit-tested in normal CI**. Confirmed removed tests targeted the deleted method (not coverage-weakening).

**big-picture** — Goal-aligned: root-cause fix, not threshold-weakening; no `#[ignore]`/weakened assertions; threshold derives from `min_connections` (not hardcoded). Suggested → **updated `.claude/rules/ring.md`** with the under-min backoff escape corollary.

### External (Codex)

- **[P2]** acquisition throttle counted relayed CONNECTs → the core fix (`acquisition_txs`).
- **[P2]** rebound acquisition tx dropped when pruning its stale old peer (undercount during churn) → first guarded, then simplified: **acquisition gauge cleared only on completion, never on prune** (race-free).
- **[P2]** get()+remove() in prune not atomic vs a concurrent rebind → resolved by the above (prune no longer touches `acquisition_txs`).
- **[P2]** gateway-join / version-probe CONNECTs not counted (bootstrap could exceed `max_concurrent`) → **register acquisitions in `start_client_connect`**, the single originator entry point. Re-validated: nightly convergence holds (93% reach min).
- Final pass: **clean** — "did not find an actionable regression introduced by this diff."

### Repo Claude Rule Review
- Warning "backoff bypass has no regression test" → addressed by the predicate unit test.

[AI-assisted - Claude]

PR #4353, opened 2026-06-05
fix(transport): abandon never-ACKed packets so lossy multi-fragment GETs don't wedge

### LINKED ISSUE #4345: GET fails with 'stream assembly: no fragments received within inactivity timeout' for multi-fragment state; contract never caches, SUBSCRIBE then rejects
## Summary

A `GET` for a multi-fragment contract state intermittently fails with
`stream assembly: no fragments received within inactivity timeout`. The node
classifies the terminal reply as success but the local store has no state, so
it synthesizes a `ContractError::Get` client error. The contract is never
cached, which then makes a follow-up `SUBSCRIBE` reject with
`contract WASM/parameters not cached locally`.

Observed on a real-network node (v0.2.68, `b5aa31e42bdc`) fetching a
freenet-email inbox contract (large-ish state: encrypted messages + ML-KEM/ML-DSA
keys, big enough to be streamed in fragments).

## Symptoms

```
WARN freenet::operations::get::op_ctx_task: get: stream assembly failed —
  state will not be cached locally key=GCdTw11g61h31DkBNbJ7L7y7u4hfBB3UhKEHjidT7ejV
  error=stream assembly: no fragments received within inactivity timeout
WARN freenet::operations::get::op_ctx_task: get: terminal reply classified
  success but local store lookup returned no state; synthesizing client error
  contract=GCdTw11g61h31DkBNbJ7L7y7u4hfBB3UhKEHjidT7ejV
```

Downstream effect (separate inbox key, same node, repeated):

```
WARN freenet::client_events: Rejecting SUBSCRIBE: contract WASM not cached
  locally. PUT the contract or GET the contract first.
  contract=5vGcxekx7e6ivA75j6rL7TVJ3qPsi6cCnNPvNhfDUCRQ
```

## Smoking gun: orphan inbound streams never complete

The node continuously garbage-collects expired orphan streams and the
`remaining` count stays high — dozens of half-assembled inbound streams that
never finish reassembling:

```
INFO freenet::operations::orphan_streams: Garbage collected expired orphan
  streams expired_count=8 remaining=49
... remaining oscillates 49–56, peak observed 109 over a ~20min window ...
```

`expired_count` distribution over the window: 102× count=1, plus bursts up to
49. This is sustained inbound-fragment loss, not a one-off.

The node itself is healthy and well-connected (`active_connections=29`,
`ring_connections=29`, `Broadcasting ReadyState ... ready=true peer_count=29`),
so this is not a connectivity/topology failure — connect + NAT traversal
succeed. The failure is specifically in **GET-response fragment streaming /
reassembly**.

## Impact

- Any contract whose state is large enough to require fragmentation can fail to
  GET intermittently on the real network.
- Because the failed GET never caches the contract, a subsequent SUBSCRIBE is
  also rejected, compounding the failure.
- For freenet-email this blocks contact import (recipient inbox never fetches)
  and, by extension, cross-user messaging.

## Questions / suspected areas

1. Why do inbound fragments stop arriving mid-stream ("no fragments received
   within inactivity timeout")? Lost UDP fragments not retransmitted, or sender
   stops emitting?
2. The terminal reply is classified **success** while no state was assembled —
   the success classification and the assembly outcome are out of sync. Should a
   stream-assembly failure mark the GET terminally failed (retryable) rather than
   success-then-synthesize-error?
3. Is the inactivity timeout too aggressive for large multi-fragment states on a
   slow/lossy path? Peak 109 simultaneous orphan streams suggests reassembly
   buffers are filling without completing.

## Environment

- freenet v0.2.68 (`b5aa31e42bdc`), built 2026-05-31
- macOS, node behind NAT (LAN 192.168.0.10), `freenet network`, 29 peers
- App: freenet-email inbox contract GET during contact import

iduartgomez:
## Problem

A `GET` for a multi-fragment contract state intermittently fails on the real
network with `stream assembly: no fragments received within inactivity
timeout`. The node classifies the terminal reply as success, the local store
has no state, a `ContractError::Get` is synthesized, the contract is never
cached, and a follow-up `SUBSCRIBE` is rejected. For freenet-email this blocks
inbox contact import (issue #4345; downstream freenet/mail#288, #302).

### Root cause: flight-size pinned by never-ACKed packets that are retransmitted forever

`flightsize` (bytes in flight per connection) is decremented only on ACK. A
packet whose ACKs never arrive — a stream whose sender task already gave up on
the `CWND_WAIT_TIMEOUT`, or a starved reverse-ACK channel on a lossy path — is
retransmitted **forever**: `SentPacketTracker` had no upper bound on
retransmissions (a packet left `pending_receipts` only on ACK, or to be
immediately re-queued). Its initial `on_send` bytes therefore stayed counted in
flight size for the life of the connection.

The production default controller is **FixedRate**, whose `on_timeout` freezes
`cwnd ≈ loss_pause_cwnd + 60KB` on first loss. With flight size pinned by the
leaked bytes, the cwnd-wait loop `flightsize + packet_size <= cwnd` can never
open, `send_stream`/`pipe_stream` aborts after the 3s `CWND_WAIT_TIMEOUT` (~39%
of the stream), the receiver hits the 5s inactivity timeout, and the **next**
stream on the same connection starts already pinned.

Live telemetry (nova OTLP collector) confirmed the mechanism:

- **90% of all `transfer_failed` events were "cwnd wait timeout"** (4883 / 5407).
- **64% clustered at `cwnd ≈ 60–70KB`** = `LOSS_PAUSE_MARGIN (60KB)` + a small
  frozen `loss_pause_cwnd`, with the cwnd−flightsize gap pinned below one packet.
- Median `sent/total` ratio **0.39** — streams die ~40% through.
- `sent 0/2, flightsize=60172B` events = the cross-stream pinning (a fresh
  stream that can't send packet #1).

Not a recent regression — the unbounded retransmit / ACK-only decrement has
been there since each controller was written; the freenet-email large
multi-fragment workload plus the 5s inactivity timeout (#3750) newly exposed it.

## Solution

Bound retransmissions and release flight size on abandonment:

- `SentPacketTracker` counts per-packet RTO retransmits. After
  `MAX_PACKET_RETRANSMITS` (12) with no ACK, `get_resend` returns the new
  `ResendAction::Abandon { packet_id, payload_len }`, drops the packet from
  tracking, and stops re-queueing it. With the 500ms base RTO and exponential
  backoff (capped at a 60s effective RTO), 12 retransmits spans minutes — far
  beyond any live transfer (production p95 large-transfer ~1.8s), so a
  still-progressing connection is never abandoned. An ACK resets the count.
- The recv loop handles `Abandon` by calling the new
  `CongestionControl::release_flightsize(len)` — a saturating decrement
  (`fetch_update`) in all three controllers (BBR/LEDBAT/FixedRate). This is the
  only flight-size release path other than ACK.
- `on_timeout()` is unchanged: it applies the cwnd loss response but does NOT
  touch flight size, because the timed-out packet is immediately re-sent and
  stays in flight.

This drains the flight size pinned by a dead stream's packets, so the cwnd-wait
loop re-opens and subsequent streams on the connection are no longer wedged.

### Why not "release on RTO"

An earlier draft decremented flight size on every RTO and re-added it on the
re-send. The recv loop *always* re-sends on RTO, so that pair is net-zero and
never drained a never-ACKed packet — it didn't fix the bug. The fix has to be
"release on abandon," and the tracker had no abandonment, so this PR adds it.

### Abandonment timing

Wall-clock to abandonment depends on whether other packets keep ACKing, because
the RTO backoff is tracker-wide and resets on any ACK while the retransmit count
is per-packet:

- **Partial loss (common multi-fragment case):** other fragments keep ACKing →
  backoff stays at the 500ms floor → a black-holed fragment is abandoned after
  ~6s (12 × 500ms), just above `STREAM_INACTIVITY_TIMEOUT` (5s) — the receiver
  has already given up; abandoning frees the connection for the next stream.
- **Fully dead link:** no ACKs → backoff escalates (capped at 60s effective) →
  ~6 minutes, by which point the 120s idle timeout tears the connection down.

A still-progressing transfer is never abandoned: any ACK for the stuck packet
resets its count, and a 12-consecutive-RTO black hole on an otherwise healthy
path is a genuinely dead fragment.

## Testing

Eight regression tests; the core ones verified to fail with abandonment
disabled and pass with the fix:

- `sent_packet_tracker::tests::test_issue_4345_packet_abandoned_after_max_retransmits`
  — a never-ACKed packet is re-sent exactly `MAX_PACKET_RETRANSMITS` times, then
  `Abandon` fires, the packet is dropped, and it is not re-queued.
- `peer_connection::tests::test_issue_4345_flightsize_drains_via_bounded_retransmit`
  — drives the real recv-loop sequence (per-RTO `on_timeout()` + re-send +
  re-register, no ACK) through the real `SentPacketTracker`; flight size stays
  pinned across the re-sends and drains only once each packet is abandoned.
  (Neutering abandonment makes this fail with `abandoned=0/30`.)
- `sent_packet_tracker::tests::test_issue_4345_ack_resets_retransmit_count` and
  `…_flaky_but_alive_packet_never_abandoned` — an ACK before the limit resets
  the count; a flaky-but-progressing packet is never abandoned across many
  RTO/ACK cycles.
- `sent_packet_tracker::tests::test_issue_4345_late_ack_after_abandon_is_noop` —
  a late ACK for an abandoned packet yields no ack entry, so flight size is not
  released twice (no double-decrement).
- `bbr` / `fixed_rate` / `ledbat`
  `test_issue_4345_release_flightsize_drains_abandoned_bytes` — per-controller:
  `on_timeout()` leaves flight size unchanged; `release_flightsize` drains it
  (saturating) and re-opens cwnd.

All 616 transport lib tests pass; CI `cargo clippy --locked -- -D warnings` is
clean. Docs updated: `docs/architecture/transport/README.md` (abandonment as a
second flight-size release path) and `.claude/rules/transport.md` (flight-size
release invariant + "don't reintroduce unbounded retransmit / net-zero RTO").

### Why no network-level (mock-peer) test

A true end-to-end test was attempted (two `PeerConnection`s, peer B drops all
post-handshake outbound, assert peer A's flight size drains) but the
`create_mock_peer_with_virtual_time` harness does not drive the recv-loop's
RTO/resend timer deterministically under the auto-advance pattern — flight size
rises on send but the resend cycle does not progress to abandonment in the test.
The mechanism is instead proven at the layer it lives in: the tracker test
drives the real `get_resend → Abandon` decision, and the `peer_connection` test
drives the real recv-loop resend/abandon sequence through the real
`SentPacketTracker`. The production recv-loop `Abandon` arm is a 3-line
`release_flightsize` call guarded by exhaustive-match compile checking.

## Fixes

Fixes #4345

### DIFF
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -83,6 +83,26 @@ BEFORE changing LEDBAT++ parameters:
   4. Document reasoning in commit message
 ```
 
+### Flight-size release invariant (issue #4345)
+
+```
+Flight size MUST have a release path other than ACK. A packet whose ACKs
+never arrive (lossy path / starved reverse-ACK channel) is retransmitted up
+to MAX_PACKET_RETRANSMITS times, then SentPacketTracker returns
+ResendAction::Abandon and the recv loop calls
+CongestionControl::release_flightsize(len). Without this, a never-ACKed
+packet's on_send bytes stay in flight size for the connection's life,
+pinning flight size at cwnd and stalling every subsequent stream
+(the #4345 "cwnd wait timeout" / "no fragments received" failure).
+
+  - on_timeout() applies the cwnd loss response only; it MUST NOT change
+    flight size (the timed-out packet is immediately re-sent and stays in
+    flight). Do NOT reintroduce a decrement-on-RTO + re-add-on-resend pair —
+    the recv loop always re-sends on RTO, so that is net-zero and does not
+    drain a never-ACKed packet.
+  - Retransmissions MUST stay bounded. Do not make retransmit infinite again.
+```
+
 ### Shadow per-peer RTT registry (issue #4074, Phases 1 + 1.5)
 
 `transport/rolling_rtt_stats.rs` maintains a process-wide
--- a/crates/core/src/transport.rs
+++ b/crates/core/src/transport.rs
@@ -678,7 +678,9 @@ mod tests {
                     // Simulate resending packet
                     sent_tracker.report_sent_packet(id, packet);
                 }
-                ResendAction::WaitUntil(_) | ResendAction::TlpProbe(..) => {
+                ResendAction::WaitUntil(_)
+                | ResendAction::TlpProbe(..)
+                | ResendAction::Abandon { .. } => {
                     panic!("Expected resend action for packet {id}")
                 }
             }
--- a/crates/core/src/transport/bbr/controller.rs
+++ b/crates/core/src/transport/bbr/controller.rs
@@ -428,6 +428,20 @@ impl<T: TimeSource> BbrController<T> {
         }
     }
 
+    /// Release an abandoned packet's bytes from flight size (issue #4345).
+    ///
+    /// Called when a packet is permanently abandoned after
+    /// `MAX_PACKET_RETRANSMITS`. Atomic saturating decrement (matches the ACK
+    /// path's accounting and never underflows under the concurrent `on_send`
+    /// from the sender task). Does NOT touch cwnd or the delivery sampler.
+    pub fn release_flightsize(&self, bytes: usize) {
+        self.flightsize
+            .fetch_update(Ordering::AcqRel, Ordering::Acquire, |current| {
+                Some(current.saturating_sub(bytes))
+            })
+            .ok();
+    }
+
     /// Get the current congestion window in bytes.
     pub fn current_cwnd(&self) -> usize {
         self.cwnd.load(Ordering::Acquire)
--- a/crates/core/src/transport/bbr/tests.rs
+++ b/crates/core/src/transport/bbr/tests.rs
@@ -496,6 +496,87 @@ fn test_timeout_storm_prevents_recovery() {
     );
 }
 
+/// Regression test for issue #4345: BBR flightsize leak on retransmission
+/// timeout.
+///
+/// Before the fix, `flightsize` was decremented ONLY on ACK. A packet that is
+/// never ACKed but retransmitted forever (lossy path / starved reverse-ACK
+/// channel) kept its initial `on_send` bytes counted in flight size for the
+/// life of the connection. `on_timeout()` resets cwnd/state/bounds but does
+/// not (and should not) touch flight size, since the timed-out packet is
+/// immediately re-sent and stays in flight.
+///
+/// The fix adds an explicit release path: once `SentPacketTracker` abandons a
+/// packet after `MAX_PACKET_RETRANSMITS`, the recv loop calls
+/// `release_flightsize(len)`. This test verifies the BBR controller's
+/// `release_flightsize` actually drains flight size and re-opens cwnd.
+///
+/// Symptom on the live network (issue #4345): `flightsize` pins at `cwnd`
+/// (telemetry showed `flightsize=59436B, cwnd=60398B`), the cwnd-wait loop
+/// `flightsize + packet_size <= cwnd` never finds headroom, every subsequent
+/// stream on the connection aborts with "cwnd wait timeout" (90% of all
+/// transfer failures), and large multi-fragment GETs (freenet-email inbox
+/// state) fail with "no fragments received within inactivity timeout".
+#[test]
+fn test_issue_4345_release_flightsize_drains_abandoned_bytes() {
+    let time = VirtualTime::new();
+    let controller = BbrController::new_with_time_source(BbrConfig::default(), time.clone());
+
+    let packet_size = 1400usize;
+    let num_packets = 40usize;
+
+    // Send a burst; none of these will ever be ACKed (the receiver / reverse
+    // ACK path is dead, exactly the lossy-stream scenario).
+    for _ in 0..num_packets {
+        let _token = controller.on_send(packet_size);
+    }
+    let flight_after_send = controller.flightsize();
+    assert_eq!(
+        flight_after_send,
+        num_packets * packet_size,
+        "flightsize should account for every in-flight byte after send"
+    );
+
+    // on_timeout() alone (the per-RTO call) must NOT change flight size —
+    // the packet is still in flight, being retransmitted.
+    controller.on_timeout();
+    assert_eq!(
+        controller.flightsize(),
+        num_packets * packet_size,
+        "on_timeout() must not change flight size; only abandonment/ACK does"
+    );
+
+    // Each packet is eventually abandoned after MAX_PACKET_RETRANSMITS — the
+    // recv loop calls release_flightsize() for each. Without this release the
+    // bytes leak for the life of the connection.
+    for _ in 0..num_packets {
+        time.advance(Duration::from_secs(1));
+        controller.release_flightsize(packet_size);
+    }
+
+    let flight_after_giveup = controller.flightsize();
+    assert!(
+        flight_after_giveup <= packet_size,
+        "issue #4345: flightsize leaked after abandonment — expected ~0, got \
+         {} bytes ({} packets still counted as in-flight). release_flightsize() \
+         must drain abandoned in-flight bytes.",
+        flight_after_giveup,
+        flight_after_giveup / packet_size,
+    );
+
+    // The connection must be usable again: a fresh packet should fit in cwnd.
+    let cwnd = controller.current_cwnd();
+    assert!(
+        controller.flightsize() + packet_size <= cwnd,
+        "issue #4345: after abandonment a new packet ({} B) must fit in cwnd \
+         ({} B) but flightsize is still {} B — the cwnd-wait loop would never \
+         open and every subsequent stream aborts",
+        packet_size,
+        cwnd,
+        controller.flightsize(),
+    );
+}
+
 /// Test that BBR's adaptive timeout floor kicks in with high BDP.
 ///
 /// When max_bdp_seen is high enough (>4x initial_cwnd), the adaptive floor
--- a/crates/core/src/transport/congestion_control.rs
+++ b/crates/core/src/transport/congestion_control.rs
@@ -206,8 +206,22 @@ pub trait CongestionControl: Send + Sync {
     /// Called when a retransmission timeout (RTO) occurs.
     ///
     /// This indicates severe congestion. Implementations typically
-    /// reset to minimum cwnd and re-enter slow start.
+    /// reset to minimum cwnd and re-enter slow start. Flight size is NOT
+    /// changed: the timed-out packet is immediately retransmitted and stays in
+    /// flight. Bytes are released only when a packet is permanently abandoned
+    /// (see [`release_flightsize`](Self::release_flightsize)).
     fn on_timeout(&self);
+
+    /// Release `bytes` from flight size for a packet that has been permanently
+    /// abandoned — retransmitted `MAX_PACKET_RETRANSMITS` times with no ACK and
+    /// dropped from tracking (see `SentPacketTracker`).
+    ///
+    /// Without this, a never-ACKed packet is retransmitted forever and its
+    /// initial `on_send` bytes stay counted in flight size for the life of the
+    /// connection, pinning flight size at cwnd and stalling every subsequent
+    /// stream on that connection (issue #4345). This is the only flight-size
+    /// release path other than ACK; the decrement saturates at zero.
+    fn release_flightsize(&self, bytes: usize);
 
     // =========================================================================
     // State Queries
@@ -350,6 +364,14 @@ impl<T: TimeSource> CongestionControl for CongestionController<T> {
         }
     }
 
+    fn release_flightsize(&self, bytes: usize) {
+        match self {
+            Self::Bbr(c) => c.release_flightsize(bytes),
+            Self::Ledbat(c) => c.release_flightsize(bytes),
+            Self::FixedRate(c) => c.release_flightsize(bytes),
+        }
+    }
+
     fn current_cwnd(&self) -> usize {
         match self {
             Self::Bbr(c) => c.current_cwnd(),
@@ -477,6 +499,10 @@ impl<T: TimeSource> CongestionControl for BbrController<T> {
         BbrController::on_timeout(self)
     }
 
+    fn release_flightsize(&self, bytes: usize) {
+        BbrController::release_flightsize(self, bytes)
+    }
+
     fn current_cwnd(&self) -> usize {
         BbrController::current_cwnd(self)
     }
@@ -551,6 +577,10 @@ impl<T: TimeSource> CongestionControl for LedbatController<T> {
         LedbatController::on_timeout(self)
     }
 
+    fn release_flightsize(&self, bytes: usize) {
+        LedbatController::release_flightsize(self, bytes)
+    }
+
     fn current_cwnd(&self) -> usize {
         LedbatController::current_cwnd(self)
     }
@@ -619,6 +649,10 @@ impl<T: TimeSource> CongestionControl for FixedRateController<T> {
 
     fn on_timeout(&self) {
         FixedRateController::on_timeout(self)
+    }
+
+    fn release_flightsize(&self, bytes: usize) {
+        FixedRateController::release_flightsize(self, bytes)
     }
 
     fn current_cwnd(&self) -> usize {
--- a/crates/core/src/transport/fixed_rate/controller.rs
+++ b/crates/core/src/transport/fixed_rate/controller.rs
@@ -163,10 +163,28 @@ impl<T: TimeSource> FixedRateController<T> {
     }
 
     /// Called when a retransmission timeout occurs.
-    /// Activates the loss pause (same as on_loss).
+    /// Activates the loss pause (caps cwnd at the current flightsize). Flight
+    /// size is unchanged: the timed-out packet is immediately re-sent and stays
+    /// in flight.
     pub fn on_timeout(&self) {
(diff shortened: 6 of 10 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4353

### Summary
- **PR Title:** fix(transport): release flight size on RTO so lossy multi-fragment GETs don't wedge
- **Type:** fix
- **CI Status:** all 17 checks passing (Clippy, Fmt, Unit & Integration, Simulation, NAT Validation, Windows, macOS E2E, …)
- **Linked Issues:** #4345
- **Review tier:** Full (touches transport / congestion control + genuinely-concurrent atomics — high-risk surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture. External model (Codex) **did not run** — its auth token was expired (`refresh_token_reused`, 401); the highest-value independent-blind-spot pass is therefore missing from this review.

---

### ⚠️ Headline finding

Three of four reviewers judged this merge-ready; the skeptical pass found — and I independently **verified against the code** — that **the fix is net-zero on the actual production path and does not resolve #4345's mechanism.** Details under Must Fix #1. This is the load-bearing finding; the rest is secondary.

---

### Code-First Analysis
**Independent understanding:** Adds a `bytes_in_flight_lost` param to `CongestionControl::on_timeout` (decrements flight size, saturating, in BBR/LEDBAT/FixedRate) and a new `on_resend(bytes)` that re-adds when a packet is retransmitted (bypassing `on_send`'s pacing/delivery sampling). Production RTO site passes `packet.len()` and calls `on_resend` only on the RTO success path (`was_rto` gate excludes TLP probes).
**Stated intent:** Release timed-out bytes so a lossy multi-fragment GET stops pinning flight size at cwnd.
**Alignment:** The committed code faithfully implements the *described mechanism*. The problem is the mechanism itself does not fire on the production path (see Must Fix #1) — implementation matches description, description does not match the bug.
**Gaps:** The RTO send-failure path decrements without re-adding (asymmetry, see Should Fix #2).

---

### Testing Assessment
**Coverage level:** insufficient for the *intended* invariant — adequate only for a give-up path that production never executes.

| Unit | ⚠️ | bbr + fixed_rate `test_issue_4345_*` verified fail-without/pass-with — but they call `on_timeout` in a loop with **no paired `on_resend`**, i.e. a give-up the recv-loop never performs |
| Integration | ⚠️ | `peer_connection::test_issue_4345_flightsize_released_on_rto_giveup` drives the real `SentPacketTracker::get_resend`, but its loop explicitly comments "Give-up: … do NOT re-send / on_resend" — again not the production sequence |
| Simulation | ❌ | none — justification is partly sound (op retry routes around a wedged connection; high-level DST can't distinguish fixed/broken) but **overstated**: the transport-loop `MockSocket` + `PacketDropPolicy::Ranges` harness (`connection_handler.rs:3700`, permanent-tail-loss test at `:4431`) could drive the real recv-loop RTO/resend wiring |
| E2E | N/A | — |

**Regression test:** present and verified to fail without the decrement — but it validates the give-up path, not the production resend path. `on_resend` and the `was_rto` TLP-gating have **zero** coverage.
**Missing tests:** (1) the round-trip `on_send → on_timeout → on_resend → on_ack` net-accounting — the case the pre-existing `test_flightsize_accounting_for_retransmitted_packets` (`peer_connection.rs:2617`) was meant to cover but doesn't, which is *why this bug was originally missed*; (2) a test driving the actual recv-loop Resend→on_timeout→successful-resend→on_resend sequence on a never-ACKed packet, asserting flight size actually drops (per the skeptical analysis it stays pinned); (3) LEDBAT got the same change but no test; (4) double-decrement saturation guard.

---

### Skeptical Findings
**Risk level:** high

| Fix is net-zero on production path → does not fix #4345 | **High** | `peer_connection.rs:1296-1352` (verified) | RTO always re-sends; `on_timeout(-S)` + `on_resend(+S)` = 0. Never-ACKed-but-resent packets keep their initial `on_send(+p)` forever; the tracker has no give-up (`MAX_RTO_BACKOFF` only caps interval; packet leaves `pending_receipts` only on ACK or to be immediately re-sent). Only `send_to` *failure* releases bytes — not the inbound-loss mechanism. |
| RTO send-failure under-counts flight size | Medium | `peer_connection.rs:1336-1351` (verified) | `on_timeout` decremented; `send_to` Err skips `on_resend`; packet re-queued + re-RTOs → second `on_timeout(-S)` for bytes counted once. Saturating floor prevents underflow; transiently over-permissive (opposite failure). |
| Unit drift: `on_send`=payload chunk, `on_timeout`/`on_resend`/ACK=full datagram | Low | `outbound_stream.rs:119` vs `peer_connection.rs:1301,1331` (verified) | Pre-existing; the PR matches the ACK convention; absorbed by saturating clamps; net-zero on the resend path so harmless here. |

**Verified SAFE (concerns raised but disproven):**
- **Atomicity:** `congestion_controller` IS `Arc`-shared and `send_stream` runs as a spawned task on a multi-thread runtime (`peer_connection.rs:2492`, `config.rs:2152`), so `on_send` genuinely runs concurrent with `on_timeout` — **the automated rule-review's "no concurrent caller" premise is factually wrong.** BBR's non-atomic load-then-`fetch_sub` is nonetheless benign: `sub = min(S, old_F) ≤ old_F` and `fetch_sub` is atomic, so a concurrent `+p` is never lost and the result is always ≥ 0 (same pattern as the pre-existing `on_ack_with_token`). `on_timeout`/`on_resend`/`on_ack` are all arms of the single recv `select!` (one `&mut self`), so they never interleave with each other.
- **FixedRate loss_pause:** decrementing before capturing `loss_pause_cwnd = fs.max(1)` lowers the cap by ≤1 packet out of the 60KB `LOSS_PAUSE_MARGIN`; `.max(1)` prevents zero. No over-throttle.
- **TLP→RTO:** TLP keeps the packet in `pending_receipts` and doesn't decrement; the later RTO decrements exactly once. No double-decrement.
- **Stale ACK after retransmit:** original + retransmit share one `packet_id`; `report_received_receipts` removes once → at most one decrement.

---

### Big Picture Assessment
**Goal alignment:** the fix is at the right *layer* (congestion-control accounting, not the GET-op symptom) and correctly avoids symptom-masking (didn't loosen the 5s inactivity timeout or touch success-classification) — but per Must Fix #1 it does not actually move flight size on the production path, so it does not yet achieve the goal.
**Anti-patterns detected:** none (no `#[ignore]`, no weakened assertions, no error swallowing). The `was_rto`/`on_resend` pairing and saturating decrements are mechanically sound for what they do.
**Removed-code concerns:** none. `test_loss_pause_allows_packet_then_blocks`, `test_loss_pause_margin_sustains_progress_under_loss` (only `on_timeout()` → `on_timeout(0)`, identity), and `test_flightsize_accounting_for_retransmitted_packets` all preserved and semantically unchanged. The mechanical `on_timeout(0)` migration across ~30 tests is the correct identity-preserving signature update, hiding nothing.
**Scope:** focused — 8 files, one logical change + tests. Correctly a `fix:` needing no prior issue. No creep.

---

### Documentation
- Code docs: complete for the added APIs.
- Architecture docs: two now slightly stale — `docs/architecture/transport/README.md:247` ("Decrement flightsize on ACK" — now also on RTO/resend) and `.claude/rules/transport.md:76` (congestion-control rules don't state the flight-size invariant). Non-blocking; fold in if the fix is reworked.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
1. **Demonstrate or restore the production effect.** The recv-loop always re-sends on RTO (`on_timeout(-S)` then `on_resend(+S)` = net zero), and the tracker never abandons a packet, so a never-ACKed-but-resent packet keeps its initial `on_send(+p)` in flight size forever — the same pin as before. The #4345 mechanism is *receive-side* loss, not local `send_to` failure (the only unpaired-`on_timeout` path). Either:
   - explain the concrete production sequence where `on_timeout` fires without a paired `on_resend` under inbound loss (and add a test that drives the **real** recv-loop sequence and asserts flight size drops — not the give-up path the current tests model); or
   - rework the fix to actually release flight size for the pinned case — e.g. release a stream's outstanding packets when its `send_stream`/`pipe_stream` task aborts on `CWND_WAIT_TIMEOUT` (which today leaves them in `pending_receipts`, retransmitted forever), or give the tracker a bounded retransmit limit that abandons + decrements. The right mechanism is "release on abandon," and abandon currently never happens.

#### Should Fix (Important)
2. **RTO send-failure path** (`peer_connection.rs:1336`): decide the accounting deliberately — either call `on_resend` there too (the packet stays tracked and will retry) or document why the transient under-count is acceptable, and add a test for send-failure-then-successful-retransmit.
3. **Test `on_resend` / the recovering path.** Add the round-trip unit test (`on_send → on_timeout → on_resend → on_ack`, asserting the net) — cheap, and it directly guards the `was_rto` gate and the half of the invariant that has no coverage.

#### Consider
4. Align BBR's `on_timeout` decrement to the `fetch_update` style used by ledbat/fixed_rate (or soften the "guard double-decrement" comment, which slightly overstates the atomicity).
5. Add a LEDBAT regression test mirroring the BBR/FixedRate ones (non-default, low priority).
6. Refresh the two stale docs above.
7. Run the external (Codex) review once its auth is restored — this review shipped without the independent non-Claude pass.

---

### Verdict
**HEAD SHA reviewed:** e1c9f8fb4

The change is well-constructed, the concurrency/TLP/loss-pause concerns are verified safe, the tests are valid for what they test, and CI is green — but the skeptical pass surfaced (and I confirmed in `peer_connection.rs:1296-1352` + `sent_packet_tracker.rs`) that the production recv-loop re-sends on every RTO, making `on_timeout`/`on_resend` net-zero and leaving the #4345 flight-size pin in place. The fix as written validates a give-up path that production never executes. Resolve Must Fix #1 (prove the production effect with a recv-loop-level test, or rework toward release-on-abandon), then re-review on the new HEAD.

The two unverified-by-me items worth noting: this review lacks the external-model pass (Codex auth expired), and Must Fix #1 is a logic/analysis conclusion — if the author can exhibit the concrete unpaired-`on_timeout` production sequence, it downgrades to "tests model the wrong path" rather than "fix is ineffective."

[AI-assisted - Claude]

iduartgomez:
## Reworked to address the blocking review finding

The review correctly caught that the original approach (decrement flight size on RTO + re-add on resend) was **net-zero on the production path** — the recv loop always re-sends on RTO, so it never drained a never-ACKed packet and didn't fix the bug.

**New approach: bounded retransmit + release-on-abandon.**
- `SentPacketTracker` now counts per-packet RTOs; after `MAX_PACKET_RETRANSMITS` (12) with no ACK it returns the new `ResendAction::Abandon`, drops the packet, and stops re-queueing it (backoff → ~minutes, far beyond any live transfer; ACK resets the count).
- The recv loop releases the bytes via the new `CongestionControl::release_flightsize(len)` (saturating `fetch_update`, all three controllers) — the only flight-size release path besides ACK.
- `on_timeout()` is **unchanged** (cwnd loss response only, no flight-size change) — which also resolves the prior review's *send-failure under-count* and *BBR non-atomic decrement* findings, since there is no longer a decrement on the RTO path.

This is now verified to **fail with abandonment disabled** (`abandoned=0/30`, flight size never drains) and pass with the fix, via a test that drives the real recv-loop sequence through `SentPacketTracker`. All 613 transport lib tests pass; CI clippy clean.

Prior findings status: net-zero/ineffective fix → **fixed** (release-on-abandon). Send-failure under-count → **moot** (no RTO decrement). on_resend untested → **moot** (removed). BBR non-atomic → **moot** (release path uses atomic `fetch_update`).

[AI-assisted - Claude]

iduartgomez:
## Re-review (HEAD 691d98c8e) — Ready to Merge

Full multi-perspective re-review of the bounded-retransmit rework (code-first, testing, skeptical, big-picture). The external Codex pass could not run (expired auth token) — noted as the one missing perspective.

**Verdict: no high-severity defects; the rework correctly fixes #4345.** Skeptical confirmed the abandonment/counting/accounting/concurrency are all sound:
- Per-packet retransmit count accumulates correctly across re-sends (re-registration doesn't reset it), reaches the limit, abandons exactly once; ACK resets it.
- Abandon cleanly removes all per-packet state; receiver fails the stream cleanly (no corruption); late-ACK-after-abandon is a no-op (no double-release).
- `release_flightsize` units match the ACK path; saturating `fetch_update` is correct under the concurrent `on_send` from the send task.

**Findings addressed in this push:**
- **Stale doc comments** (all reviewers) referencing the abandoned `on_resend`/`on_timeout(len)` API → corrected to `release_flightsize`.
- **Timing claim** (skeptical M1): the "~minutes" doc was wrong for the common partial-loss case (backoff resets on other ACKs → ~6s). Doc now states both regimes (~6s partial-loss / ~6min dead-link) and the safety margin vs the 5s/120s timeouts.
- **Test gap** (testing, Critical): the integration test re-implemented the recv-loop instead of driving it. Added the reviewer-requested tracker-level tests: late-ACK-after-abandon (double-release safety) and flaky-but-alive-never-abandoned. Added a LEDBAT `release_flightsize` test for controller symmetry. Updated transport README + transport.md rule.
- **Network e2e test:** attempted via the mock-peer + VirtualTime harness but it doesn't drive the recv-loop RTO timer deterministically; documented why, and the mechanism is covered at the tracker + recv-loop-sequence level instead (see PR body).

8 #4345 tests, 616 transport lib tests pass, CI clippy clean.

[AI-assisted - Claude]

PR #4367, opened 2026-06-08
fix(get): retry against next candidate when stream assembly fails

sanity:
## Problem

#4345: a GET for a multi-fragment contract state intermittently fails with `stream assembly: no fragments received within inactivity timeout`. The post-v0.2.70 analysis on the issue showed the per-stream transport failure chain is still intact, and on top of it the GET driver compounds the problem at the op layer:

The driver classifies the `ResponseStreaming` **header** as terminal success and exits its retry loop. If `assemble_and_cache_stream` then fails (fragments lost on a lossy path, the sender aborting on the 3s cwnd-wait timeout, a relay's pipe dying after the header was forwarded), the driver only logs a WARN and synthesizes the generic `GET succeeded on wire but local store lookup failed` client error. One transport hiccup burns the whole GET — the `MAX_RETRIES = 3` budget is never consumed, even though other candidates could serve the contract. The relay driver already treats header-without-stream as a retryable routing failure (`drive_relay_get_inner`'s claim-failure `continue`); the originator had no equivalent.

This is the failure behind `riverctl invite accept` alternating between the store-lookup error and 60s timeouts, and freenet/mail#288's inbox-GET failures.

## Approach

**Op layer (commit 1, hardened in commit 4):** new `drive_get_with_assembly_retry` wrapper around the shared retry loop, used by both the client driver and the sub-op driver. On assembly failure it:

- records `RouteOutcome::Failure` for the candidate whose header never became a stream (client path only — sub-op GETs have never fed the router, mirroring pre-existing design);
- advances to the next candidate via the existing `MAX_RETRIES` budget;
- re-enters the loop with a **fresh attempt tx** — reusing the original tx would collide with the relay dedup gates (`active_relay_get_txs`) while the failed attempt's relay chain is still draining.

Once a streaming header has been seen, exhaustion can never surface as `NotFound` (the contract provably exists): assembly-time budget exhaustion keeps the original `Done(Streaming)` outcome, and a wire exhaustion on a re-entered loop is converted back to the remembered header (review finding, commit 4). Either way the client sees a diagnostic `OperationError` carrying the assembly cause. The shape (outer wrapper rather than retrying inside the loop's Terminal arm) is dictated by the `op_ctx.rs` pin test requiring the Terminal arm to return synchronously.

**Observability (commit 2):** `transfer_failed` / `transport_snapshot` / `timeout` telemetry events carried an empty `peer_id`, and no event carried a version — the v0.2.70 post-release analysis could not attribute the failing senders. The emitting node's own peer id is now threaded into `TelemetryReporter`/`TelemetryWorker` (same construction and caveats as the shadow-RTT events in `p2p_impl.rs`), and the standard `service.version` OTLP resource attribute is added. Additive only; the dashboard parses attributes permissively and `events.peer_id` was already nullable.

**Out of scope:** the underlying transport failure chain (loss-pause cwnd cap + 3s cwnd-wait abort + flightsize release semantics). That fix needs a stream→packet index in `SentPacketTracker` and touches congestion-control invariants — analysis posted on #4345 for the assignees. This PR makes individual GETs survive that chain; it does not reduce the per-stream failure rate.

## Testing

- **E2E regression** (`test_streaming_get_retries_after_assembly_failure`): two-phase cold-node isolation on a routable sparse topology; arms exactly one deterministic assembly failure for the contract key via the new `get_assembly_fault_injection` hook (keyed by `ContractKey` so parallel tests can't consume each other's budget; the injection fails before the claim — the relay claim-failure shape — leaving the inbound stream orphaned for GC; driver-visible behavior is identical to the production mid-assembly timeout (both surface as `Err` from `assemble_and_cache_stream`)). Asserts the driver took the retry path and the requester ended up with the full, correct state. **Verified to fail without the fix** (every candidate: retry never taken, no state) **and pass with it**.
- New source-scrape pin tests: assembly failure must capture the failed target before `advance()`, must use a fresh attempt tx, and exhaustion must never become `NotFound`; both drivers must route through the wrapper.
- Unit tests for the injection hook semantics and for telemetry (`transfer_failed` carries the local peer id; OTLP resource carries `service.version`).
- Full `streaming_e2e` suite green (11 passed, 2 pre-existing ignores); GET driver unit tests green (83); `cargo fmt` + CI-equivalent `cargo clippy --locked` clean.

Part of #4345 — the issue stays open for the transport-layer root cause.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -26,7 +26,13 @@ Driver entry points:
 
 The shared retry-loop driver lives in `op_ctx.rs` (`RetryDriver` trait
 + `drive_retry_loop`). UPDATE is fire-and-forget (no retry loop, no
-upstream reply); GET/PUT/SUBSCRIBE share the retry driver.
+upstream reply); GET/PUT/SUBSCRIBE share the retry driver. GET's
+client and sub-op drivers enter the loop via
+`get/op_ctx_task.rs::drive_get_with_assembly_retry`, which treats a
+post-terminal stream-assembly failure as a retryable attempt failure
+(#4345) — post-terminal side effects that can fail retryably belong in
+such a wrapper, never inside the loop's Terminal arm (pinned by
+`drive_retry_loop_terminal_arm_does_not_call_advance`).
 
 ## Wire-variant dispatch
 
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -109,6 +109,10 @@ pub mod dev_tool {
     pub use crate::operations::get::op_ctx_task::RELAY_DRIVER_CALL_COUNT as GET_RELAY_DRIVER_CALL_COUNT;
     #[cfg(any(test, feature = "testing"))]
     pub use crate::operations::get::op_ctx_task::RELAY_GET_STREAMING_FORWARD_COUNT;
+    // Deterministic stream-assembly fault injection + retry counter for
+    // the GET driver's assembly-retry path (#4345).
+    #[cfg(any(test, feature = "testing"))]
+    pub use crate::operations::get::op_ctx_task::assembly_fault_injection as get_assembly_fault_injection;
     #[cfg(any(test, feature = "testing"))]
     pub use crate::operations::put::op_ctx_task::RELAY_PUT_DRIVER_CALL_COUNT;
     #[cfg(any(test, feature = "testing"))]
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -566,8 +566,21 @@ impl NodeConfig {
                 registers.push(Box::new(OTEventRegister::new()));
             }
 
-            // Add telemetry reporter if enabled in config
-            if let Some(telemetry) = TelemetryReporter::new(&self.config.telemetry) {
+            // Add telemetry reporter if enabled in config. The local
+            // peer id (public key + best-effort address, same
+            // construction as the shadow-RTT events in `p2p_impl.rs`)
+            // attributes transport-level events — transfer_failed,
+            // transport_snapshot, timeout — which otherwise carry an
+            // empty peer_id and cannot be correlated to a sender in
+            // the collector (#4345 observability gap).
+            let local_peer_id = PeerId::new(
+                self.key_pair.public().clone(),
+                self.own_addr.unwrap_or_else(|| {
+                    std::net::SocketAddr::new(self.network_listener_ip, self.network_listener_port)
+                }),
+            )
+            .to_string();
+            if let Some(telemetry) = TelemetryReporter::new(&self.config.telemetry, local_peer_id) {
                 registers.push(Box::new(telemetry));
             }
 
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -302,6 +302,19 @@ pub struct NodeConfig {
 }
 
 impl NodeConfig {
+    /// This node's own peer id as a telemetry attribution string
+    /// (public key + best-effort address). The address portion falls
+    /// back to the listener address for non-gateway nodes until
+    /// external-address discovery — a refresh path is tracked in
+    /// #4294. Shared by the telemetry reporter and the shadow-RTT /
+    /// reference-ping emitters so the two constructions can't drift.
+    pub(crate) fn local_peer_id_string(&self) -> String {
+        let addr = self.own_addr.unwrap_or_else(|| {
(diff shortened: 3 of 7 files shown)


sanity:
## Consolidated multi-perspective review (Full tier)

Reviewers: code-first, testing, big-picture, adversarial (4 Claude perspectives) + **Gemini** as the external non-Claude pass (Codex unavailable — usage limit until Jun 10). All findings addressed in b8140fca or dismissed with justification below.

### Fixed (b8140fca)

1. **False NotFound after assembly-retry exhaustion** (code-first #1, testing #1 — the one blocking-class finding, found independently by both): an assembly failure that advanced and re-entered the loop could then exhaust on the wire, and the pass-through `Exhausted` mapped to `NotFound` — false for a contract whose streaming header proved existence, and it silently dropped the assembly diagnostic. The wrapper now remembers the failed header and converts a later wire exhaustion back to `Done(Streaming)` → diagnostic `OperationError`. Pinned.
2. **Stale `assembly.error`** (code-first #4, testing #6): cleared when a retry resolves via `InlineFound`/`LocalCompletion`, so unrelated store-lookup misses aren't mislabeled as assembly failures.
3. **Incoherent cross-attempt timing on the conversion path** (adversarial #1): `request_sent_at`/`response_received_at` cleared in the conversion arm — the received<sent "clock-source regression" WARN no longer fires with a stale pair.
4. **Test gaps** (testing #2–#5, #8): cause strings extracted into pure helpers with unit tests; route-failure emission + header memory pinned in the Next arm and pinned absent in the Exhausted arm; `transport_snapshot` construction extracted + tested; timeout-event peer-id stamping tested; tighter pin anchoring on `match assemble_and_cache_stream`.
5. **E2E CI bound** (testing): candidate iteration capped at 6 (deterministic seed demonstrates on the 5th candidate); process-global counter caveat documented.
6. **Peer-id construction dedup** (code-first #7): shared `NodeConfig::local_peer_id_string()` now used by both the telemetry reporter and the shadow-RTT/reference-ping path.
7. **`.claude/rules/operations.md`** (big-picture): GET assembly-retry wrapper documented as the canonical entry shape for retryable post-terminal side effects.

### Dismissed, with justification

- **Worst-case GET latency up to ~8 min** (Gemini P2): real but bounded by the same `MAX_RETRIES` budget that already allowed ~4 min of wire timeouts pre-fix; clients keep their own timeouts (riverctl 60s, mail 30s) so the user-visible wait is unchanged — and the background completion that follows is precisely the desirable behavior observed in the v0.2.70 field report (contract eventually cached + auto-subscribed, so the user's next attempt succeeds). Documented in the wrapper rustdoc.
- **Leaf-node peer-id address staleness** (Gemini P2): known caveat shared with the shadow-RTT events — the pubkey half (the part that identifies the node) is always correct; the address half falls back to the listener until external-address discovery. Refresh is #4294's scope; the shared helper means one fix covers all emitters.
- **Sub-op retries don't feed the router** (Gemini P3): deliberate — sub-op GETs have never emitted route events (pre-existing design); enabling that is a behavior change beyond this fix's scope.
- **Addressless-target break instead of advance** (Gemini P3): only reachable when the *initial* target was the addressless `own_location()` fallback on an isolated node; preserves pre-fix semantics exactly for a path where a streaming header is already pathological.
- **Success route event can credit the last exhausted candidate when the local store happens to hold a copy** (adversarial #2, nit): rare, telemetry-noise-level; returning the local copy to the client is correct, and re-attributing the route event would complicate the shared Done arm for negligible model impact.
- **`maybe_subscribe_child` unconditional on `host_result`** (adversarial #3): pre-existing asymmetry, reachable before this PR via the inline assembly-failure path; gating it changes explicit-subscribe semantics and belongs in its own change if wanted.
- **`testing` feature unification via fdev** (adversarial #4): pre-existing repo-wide property; the new hook is inert unless `inject_failures` is called, matching the existing test-hook pattern.

### Follow-ups noted (not this PR)

- Merge-order coordination with draft #4329 (same-file conflict guaranteed; whichever lands second needs conflict-resolution re-review).
- bug-prevention-patterns row for the incident class "terminal success classified before a fallible side effect completes → retry budget never consumed" (lives in the parent freenet config repo).
- Optional: preserve the original client tx in structured logs across assembly retries (big-picture nit).

Adversarial reviewer verified explicitly: loop termination (budget consumed before any candidate including the bootstrap fallback), orphan-stream GC bounds the abandoned stream (60s TTL, no cross-attempt claim collision), per-attempt `pending_op_results` slots released on all paths, no missed `TelemetryReporter::new` call sites.

[AI-assisted - Claude]

PR #4368, opened 2026-06-09
feat: warn on the homepage when served assets are stale (#4289)

### LINKED ISSUE #4289: Homepage: detect asset/runtime version mismatch and show warning (follow-up to #3967)
## Problem

When a Freenet node binary on disk has been updated to vX but the
running process is still vY (stuck wrapper, port-conflict orphan,
launchd backoff), the homepage at \`http://127.0.0.1:7509/\` keeps
serving vY's embedded assets. To the user this looks like a frozen
UI rather than a stale node. In the original #3967 incident, the
homepage \"kept serving the old version's embedded assets and a
frozen connected-peers list\" — the user only noticed days later.

## Proposed work

Per item 4 of #3967's acceptance checklist, when the embedded
homepage detects its build version doesn't match the running node's
reported version (homepage shipped with v0.2.37 assets, node reports
v0.2.49 via API), show a visible banner:

> \"Asset version 0.2.37 ≠ node version 0.2.49 — refresh the page.\"

This makes the \"stale assets\" symptom self-diagnosing instead of
looking like a frozen UI.

## Implementation notes

- Embed the build version in the asset bundle at compile time (e.g.
  via \`build.rs\` writing \`VERSION\` into a const).
- Have the embedded JS query the existing \`/v1/diagnostics\` (or
  similar) endpoint at page load to fetch the runtime version.
- If they don't match, render a banner at the top of the page.

## Acceptance criteria

- [ ] Homepage detects asset/runtime version mismatch and shows a
      visible warning.

## References

- #3967 — parent issue (acceptance criterion 5).
- PR #4286 — first slice of the #3967 fix.

[AI-assisted - Claude]

sanity:
## Problem

When a Freenet node binary on disk is updated to vX but the running
process is still vY (a stuck wrapper, a port-conflict orphan, or launchd
backoff), the homepage at `http://127.0.0.1:7509/` keeps serving the old
binary's embedded assets. To the user this looks like a frozen UI rather
than a stale node. In the #3967 incident the staleness went unnoticed
for days.

This is item 4 of #3967's acceptance checklist.

## Solution

Make the "stale assets" symptom self-diagnosing.

- The served homepage now carries its **asset version** — the
  `CARGO_PKG_VERSION` of the binary that emitted the page — baked into a
  hidden banner slot (`data-asset-version`).
- A new lightweight endpoint, `GET /v1/version`, returns the **runtime
  version** of the process actually answering requests right now (as
  JSON: `{"version": "0.2.x"}`).
- At page load the homepage JS fetches `/v1/version` and reveals a
  warning banner when the live runtime version differs from the asset
  version baked into the page the browser is running. The check re-runs
  on each auto-refresh, so the banner appears (or clears) live if the
  serving process changes while the page stays open.

The key insight is comparing against a **live fetch** rather than the
rendered `data-version`: within a single served page both come from the
same binary and always agree, so they can't detect the #4289 case. A
live fetch catches the scenario where the browser holds a cached page
from an old binary while a newer binary is now answering.

Why not a `build.rs` const (as the issue sketched): the asset version is
already available as the compile-time `PCK_VERSION`
(`env!("CARGO_PKG_VERSION")`), so no extra build plumbing is needed — the
page is emitted by the same binary whose version we bake in.

The mismatch decision is a pure, unit-tested helper
(`should_show_version_banner`): show only when both versions are known
and they differ; an unknown (`"?"`/empty) version on either side never
triggers the banner, so it cannot fire spuriously during startup. The JS
mirrors the same rule (`checkVersionMismatch` + `versionIsKnown`), and
the banner uses `textContent` (not `innerHTML`) so a version string can't
inject markup.

## Testing

New unit tests in `crates/core/src/server/home_page.rs`:

- `version_banner_hidden_when_versions_match` — match → no banner.
- `version_banner_shown_when_versions_differ` — mismatch (either
  direction) → banner.
- `version_banner_treats_prerelease_suffixes_as_distinct` —
  `0.2.68-rc1` vs `0.2.68` is a real mismatch; identical pre-release
  strings are not.
- `version_banner_hidden_when_either_version_unknown` — `"?"`/empty on
  either side never shows the banner (startup-race guard).
- `homepage_renders_version_mismatch_banner_slot` — the hidden banner
  slot with `data-asset-version` is rendered.
- `js_contains_version_mismatch_check` — the JS wires up the check and
  queries `/v1/version`.

New endpoint test in `crates/core/src/server/client_api/v1.rs`:

- `runtime_version_reports_running_pkg_version` — `GET /v1/version`
  returns the running binary's `CARGO_PKG_VERSION` as JSON.

`cargo fmt` and `cargo clippy -p freenet --all-targets -- -D warnings`
are clean.

## Fixes

Closes #4289

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api/v1.rs
+++ b/crates/core/src/server/client_api/v1.rs
@@ -11,6 +11,11 @@ use super::*;
 pub(super) fn routes(config: Config) -> Router {
     Router::new()
         .route("/v1", get(home))
+        // Lightweight runtime-version probe. The homepage JS fetches this at
+        // page load and compares the live answer against the version baked into
+        // the served page; a mismatch means the browser is holding a stale
+        // cached page from an older binary (#4289).
+        .route("/v1/version", get(runtime_version))
         // No-trailing-slash redirect to the canonical contract root URL.
         // Without this, pasting `/v1/contract/web/<key>` (no slash)
         // into the address bar 404s — common HTTP UX is to either
@@ -22,6 +27,27 @@ pub(super) fn routes(config: Config) -> Router {
         .route("/v1/contract/web/{key}/", get(web_home_v1))
         .route("/v1/contract/web/{key}/{*path}", get(web_subpages_v1))
         .with_state(config)
+}
+
+/// JSON returned by `GET /v1/version`.
+#[derive(serde::Serialize)]
+struct RuntimeVersion {
+    /// Version of the binary currently serving requests
+    /// (`CARGO_PKG_VERSION` of the running process).
+    version: &'static str,
+}
+
+/// `GET /v1/version` — reports the running node's version so the homepage JS
+/// can detect when it is rendering a stale, cached page from an older binary.
+///
+/// Returns `PCK_VERSION` (the compile-time version of *this* live process)
+/// rather than the `network_status` snapshot: both are seeded from the same
+/// constant, but reading the constant directly is always available, even
+/// during the startup window before the snapshot exists.
+async fn runtime_version() -> impl IntoResponse {
+    axum::Json(RuntimeVersion {
+        version: crate::config::PCK_VERSION,
+    })
 }
 
 async fn web_home_v1(
@@ -71,6 +97,26 @@ mod tests {
 
     fn valid_key() -> &'static str {
         "EqJ5YpEEV3XLqEvKWLQHFhGAac2qXzSUoE6k2zbdnXBr"
+    }
+
+    /// `GET /v1/version` must report the running binary's version as JSON so
+    /// the homepage stale-assets check (#4289) has a live runtime version to
+    /// compare against. The reported version must match the compiled-in
+    /// `CARGO_PKG_VERSION`.
+    #[tokio::test]
+    async fn runtime_version_reports_running_pkg_version() {
+        use axum::body::to_bytes;
+
+        let response = runtime_version().await.into_response();
+        assert_eq!(response.status(), StatusCode::OK);
+
+        let body = to_bytes(response.into_body(), usize::MAX).await.unwrap();
+        let json: serde_json::Value = serde_json::from_slice(&body).unwrap();
+        assert_eq!(
+            json.get("version").and_then(|v| v.as_str()),
+            Some(env!("CARGO_PKG_VERSION")),
+            "endpoint must report the running binary's CARGO_PKG_VERSION"
+        );
     }
 
     /// Regression test for freenet/freenet-core#4019.
--- a/crates/core/src/server/home_page.rs
+++ b/crates/core/src/server/home_page.rs
@@ -20,48 +20,6 @@ pub(super) async fn peer_detail(Path(address): Path<String>) -> impl IntoRespons
     Html(peer_detail_html(&address))
 }
 
-/// Reference specification of the "stale assets" banner rule, used to test
-/// the logic the homepage JS implements client-side.
-///
-/// The actual mismatch detection runs in the browser (see
-/// `checkVersionMismatch` in [`JS`]): the served page bakes in its
-/// `asset_version` (the build that generated the HTML/JS the browser is
-/// running) and fetches the live `runtime_version` from `/v1/version` at page
-/// load. This Rust function is the canonical, unit-testable statement of when
-/// the banner should appear; the JS mirrors it exactly. Keeping it here (and
-/// tested) makes the rule's edge cases explicit and guards against the JS
-/// drifting from the intended behaviour. It is `#[cfg(test)]` because the
-/// production decision is made in JS, not in the server-side render.
-///
-/// The mismatch is meaningful in the #3967 / #4289 scenario: a browser is
-/// still holding a cached homepage emitted by an old binary while a newer
-/// binary is now the process actually answering requests. In that case the
-/// asset version (frozen in the cached page) differs from the live runtime
-/// version, and the page is genuinely stale — the user should refresh.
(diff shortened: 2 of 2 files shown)


sanity:
## Review summary (autonomous debug-and-fix, cc-uprizer)

### External-model availability

Per `multi-model-review`, I attempted the independent non-Claude pass first:

- **Codex** (`codex review --base main`): failed — hard usage limit hit
  ("You've hit your usage limit … try again at Jun 10th 2026 9:21 PM").
- **Gemini** (`gemini --approval-mode plan`, adversarial prompt on the full
  diff): failed — repeated capacity exhaustion ("You have exhausted your
  capacity on this model"), thrashing across 5+ retries with backoff, never
  completed a review.

Both external reviewers were genuinely unavailable and the Codex reset is far
off, so I used the **Claude diverse-lens fallback** (≥3 independent adversarial
lenses, each reading the checked-out code). When external models are back, a
real external pass is welcome before merge.

### Lenses run

1. **Skeptical bug-hunt** — auto-refresh interaction: the banner element lives
   outside `<main>`, so its `data-asset-version` stays anchored to the
   originally-loaded page across `<main>` innerHTML swaps; `checkVersionMismatch`
   is re-run on each refresh so the banner appears/clears live if the serving
   process changes. Correct.
2. **Failure mode — false positives across version formats** — the comparison
   is exact-string. In the non-stale case both sides are the same binary's
   `PCK_VERSION` (byte-identical). `html_escape` only touches `& < > "`, none of
   which appear in semver (`0.2.x`, `-rc1`, `+build`), and `getAttribute`
   decodes entities on read anyway, so the attribute round-trips to identity. No
   false positive. Pre-release suffixes are intentionally treated as distinct
   (tested).
3. **Code-first intent vs impl** — JS `checkVersionMismatch`/`versionIsKnown`
   mirror the `#[cfg(test)]` Rust spec `should_show_version_banner`/`version_is_known`
   exactly (banner only when both known and differing; `?`/empty never fires —
   startup-race guard). Endpoint returns `{"version": "<PCK_VERSION>"}`, matching
   the JS's `data.version` read.
4. **Security / leakage** — `GET /v1/version` is read-only, no params, returns
   only the already-public version string (also in the homepage badge and the
   transport handshake). No new disclosure, negligible DoS surface (static
   const). New exact route, no overlap with existing `/v1/...` wildcards. Banner
   uses `textContent` (not `innerHTML`), so a version string can't inject markup.

### Findings

No blocking findings. The mismatch logic is unit-tested at the spec level
(match → no banner; mismatch → banner; pre-release distinct; unknown → no
banner), the banner-slot render and JS wiring are pinned by tests, and the
endpoint is covered.

### Local verification

- `cargo fmt -p freenet` — clean.
- `cargo clippy -p freenet --locked -- -D warnings` (CI's exact gate) — clean,
  exit 0.
- 7 new unit tests pass.

Note: a pre-existing `clippy::int_plus_one` lint exists in
`crates/core/src/transport/peer_connection.rs:3324` (test code, from commit
`becd219`); it only surfaces under `--all-targets`, which CI does not run, so it
does not affect this PR. Flagged here for visibility.

[AI-assisted - Claude]

sanity:
Fixed the red CI check (`Unit & Integration` → `peer_detail_panel_calls_estimator_helper_for_all_three_components`).

## Root cause
That test is a source-scrape pin: it computes the production/test boundary as `src.find("#[cfg(test)]")` and asserts everything before that first marker (the "production" region) contains the three `build_estimator_chart_or_placeholder(...)` calls. This PR added the `should_show_version_banner` / `version_is_known` reference helpers as standalone `#[cfg(test)]`-gated items *above* the estimator code, so the first `#[cfg(test)]` marker moved before the estimator calls. The estimator calls fell outside the scanned `prod` slice and `assert!(found, ...)` panicked.

## Fix
Moved both helpers inside the existing `#[cfg(test)] mod tests` block (commit `6f5e2740`). They are genuinely test-only (the production decision runs in JS; these Rust fns only exist as the canonical, unit-tested statement the JS mirrors), so the test module is where they belong. This restores the single-boundary invariant — the first `#[cfg(test)]` marker is again the real production/test boundary, after all production code. Function bodies are unchanged; no behavior change. The estimator pin (which guards a real past regression of empty slots silently disappearing) is left intact, not relaxed. Added a doc note on the helper explaining why it must stay inside the test module so a future editor does not re-break the pin.

Also rebased onto current `main` (clean, no conflicts; #4360/#4367/#4370/#4372 touched unrelated files).

## Verification
- `cargo test -p freenet --lib server::home_page`: **73 passed, 0 failed, 1 ignored** — including the previously-failing pin test and all `version_banner_*` / `js_contains_version_mismatch_check` / `homepage_renders_version_mismatch_banner_slot` tests.
- `cargo fmt --check` clean; `cargo clippy -p freenet --locked -- -D warnings` clean (CI gate).

## Review note
External model review (Codex, Gemini) was attempted but both are over quota right now (Codex resets ~Jun 10 21:21, Gemini ~20h). As a background subagent I cannot spawn nested Claude reviewer subagents, so I did a skeptical self-review of this Light-tier, test-only code move: confirmed (1) byte-identical function bodies / no behavior change, (2) helpers remain in scope for the banner tests via the local `mod tests`, (3) the `[`+"`"+`JS`+"`"+`]` intra-doc link still resolves under `use super::*;`, (4) only one source-scrape pin exists and its cutoff invariant is preserved, (5) the rebase was a clean fast-forward.

[AI-assisted - Claude]

Issue #4274, opened 2026-05-26
feat: peer key and contract blocklists


Basedfloppa:
## Problem

When a misbehaving peer in the Freenet DHT repeatedly requests stale or nonexistent contracts, operators currently have no way to block it. The existing `--blocked-addresses` flag accepts only `IP:port` pairs, but in a permissionless DHT a single peer can relay through hundreds of intermediate nodes, making IP-based blocking ineffective.

**Real-world example:** An operator observed a peer generating hundreds of thousands of WARN-level log entries over 48 hours by repeatedly requesting a deleted legacy contract instance. The requests arrived through 60+ distinct IP addresses; the single most frequent source was the gateway's own public address (relay loop via DHT). IP-based blocking would have required enumerating and continuously updating a blocklist of transient relay nodes, which is not operationally feasible.

While logs alone are not a critical resource concern, sustained WARN spam desensitises operators to real warnings and makes incident response harder.

## Solution

Add a `--blocked-peers` / `--deny-peer` flag (and optionally a `--blocked-peers-file` for bulk bans) that accepts **base58-encoded peer public keys**:

```
freenet network \
  --blocked-peers Fv2xMhJZ7dAbC8kLpQrStUvWXyZ3n4m5o6p7q8r9s0t \
  --blocked-peers-file /etc/freenet/blocked-peers.txt
```

When a peer key is blocked the node should:

1. **Reject incoming messages** signed by that key at the transport layer (before any application-level processing).
2. **Drop any existing connection** to/from that peer.
3. **Log once** at INFO level: `blocked peer <key> — 46 messages dropped since last log` (ratelimited, not per-message).
4. **Persist** the block across restarts (store in `config.toml` or a dedicated file).

### Bonus: `--blocked-contracts` — filter by contract instance or code hash

The same scenario revealed a second missing primitive: once a contract instance is superseded (e.g. after a state-format migration that requires a new instance), the old instance continues to accumulate inbound requests because the DHT does not know it is defunct. Operators should be able to **block a contract by instance ID and/or code hash** so the node refuses to route or serve state for it:

```
freenet network \
  --blocked-contracts 9xYcLpR3nM7vW2qB5kH8jF4tD1zS6gA0e
```

When a contract key or code hash is blocked the node should:

1. **Refuse to serve state** for that contract to requesting peers (return "contract not found").
2. **Drop inbound update/query messages** for that instance without processing.
3. **Not propagate** updates for blocked contracts to other peers.
[...]

PR #4370, opened 2026-06-09
feat(ring): cap ContractBanList at MAX_BANNED_CONTRACTS

### LINKED ISSUE #4303: Phase 7 follow-up: MAX_BANNED_CONTRACTS size cap
## Context

Spawned from PR #4299 multi-model review.

\`ContractBanList\` currently has no size cap. Today this is safe because the only writer is the governance reaper, and the reaper can only ban contracts the local node observes — so size is bounded by the local contract population (~hundreds at most).

When the operator-CLI surface (issue #4274) lands and starts writing via \`BanReason::Operator\`, the input becomes operator-influenced. A misconfigured config file or a hostile operator (or an attacker who reaches the CLI somehow) could push arbitrary keys into the list.

Match the defense-in-depth pattern Phase 2's \`UpdateRateLimiter\` uses (\`MAX_TRACKED_PAIRS = 16_384\` enforced via \`AtomicUsize::fetch_add\` reservation).

## Proposed

Add \`MAX_BANNED_CONTRACTS\` (suggest 8_192 — well above expected production size) and gate \`ContractBanList::ban\` so attempts to exceed the cap fail with a debug log. Reaper-driven bans (\`BanReason::AutoMad\`) take precedence over operator-driven adds when the cap is hit. Pre-flight check is \`AtomicUsize::fetch_add\` to avoid the same probe-then-insert race Phase 2's review caught.

Land this **before** #4274 ships the operator surface so the cap is in place when the new attack surface opens.

## Spawned from

- Big-picture review (\"No size cap on the ban list\")

[AI-assisted - Claude]

sanity:
## Problem

`ContractBanList` (`crates/core/src/ring/contract_ban_list.rs`) had no
size cap. Today the only writer is the governance reaper, whose input is
bounded by the local contract population, so the list stays small. But
once the operator-CLI surface (#4274) lands and starts writing via
`BanReason::Operator`, the input becomes operator-influenced: a
misconfigured config file, a runaway script, or an attacker who reaches
the management socket could push arbitrary keys into the list. Banned
contracts are dropped at the wire boundary (`is_banned` is checked before
every inbound PUT/GET/UPDATE/SUBSCRIBE), so a full ban list silently
drops legitimate traffic with no user-facing error — the hardest class of
failure to diagnose. The cap needs to be in place *before* that attack
surface opens.

This mirrors the unbounded-per-key-collection rule in
`.claude/rules/code-style.md` and the defense-in-depth pattern Phase 2's
`UpdateRateLimiter` already established.

## Approach

Mirror `UpdateRateLimiter` (`MAX_TRACKED_PAIRS` enforced via an
`AtomicUsize::fetch_add` reservation):

- Add `MAX_BANNED_CONTRACTS = 8_192` (well above expected production
  size; worst-case memory is trivially small).
- Reserve a slot with `fetch_add` **before** inserting and roll back on
  overshoot. This strictly serializes concurrent distinct-key inserts so
  the cap cannot overshoot — avoiding the probe-then-insert race a
  `len()` precheck would have (the exact issue #4285's review caught for
  the rate limiter, where `len()` overshoots by `num_concurrent_callers`).
- `unban()` and `cleanup()` decrement the counter so the cap recovers as
  bans roll off.
- `ban()` now returns a `BanOutcome` (`Banned` / `Updated` /
  `CapacityExceeded`) and logs a `debug!` line on capacity rejection.
  `capacity_rejected_total()` is exposed for the dashboard, mirroring the
  rate limiter's counter.

**Priority rule** (per the issue): reaper-driven (`AutoMad`) bans take
precedence over operator (`Operator`) bans when the list is full. An
`AutoMad` ban at capacity evicts one `Operator` entry to make room; if
the list is entirely reaper-driven there is nothing to evict and the add
is rejected. An `Operator` ban at capacity is always rejected — operator
input must never displace a security decision the node made itself.

No behavior change for current callers: the reaper-only write path is
bounded by the local contract population, far below 8 192, so it never
hits the cap. `apply_ban_decisions` ignores the new return value (the cap
is invisible until the list is actually full).

## Testing

New unit tests in `contract_ban_list.rs` (all 28 module tests pass;
`cargo fmt --check` and CI-equivalent `cargo clippy --locked -- -D
warnings` clean):

- `capacity_exceeded_when_cap_reached` — fill to cap, next add is
  `CapacityExceeded`, list size does not overshoot, counter increments,
  existing contracts keep working.
- `operator_ban_rejected_at_cap_when_only_auto_present` — operator add
  cannot displace reaper bans.
- `auto_ban_evicts_operator_entry_at_cap` — reaper ban evicts an operator
  entry and is admitted (one in, one out).
- `auto_ban_rejected_at_cap_when_no_operator_to_evict` — reaper ban
  rejected when nothing is evictable.
- `unban_decrements_size_counter_freeing_a_slot`,
  `cleanup_decrements_size_counter_freeing_slots` — slots recover.
- `unban_missing_contract_does_not_underflow_counter` — spurious unban is
  a no-op on the counter.
- `concurrent_distinct_bans_do_not_overshoot_cap` — 64-thread flood of
  distinct keys against an 8-slot cap admits **exactly** CAP, size equals
  CAP, no overshoot (the strict-serialization property).
- `ban_returns_outcome` — Banned vs Updated.

Closes #4303

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring/contract_ban_list.rs
+++ b/crates/core/src/ring/contract_ban_list.rs
@@ -200,9 +200,13 @@ impl ContractBanList {
     /// and reason.
     ///
     /// Idempotent for an already-banned contract: re-banning extends
-    /// the expiry to whichever value is later (never shortens it) and
-    /// upgrades the reason to `Operator` if applicable. This path
-    /// consumes no new slot and is never capacity-rejected.
+    /// the expiry to whichever value is later (never shortens it). The
+    /// stored reason is the *higher-priority* of the existing and
+    /// incoming reasons: [`BanReason::AutoMad`] dominates
+    /// [`BanReason::Operator`], so once a contract has been auto-banned
+    /// by the node's own security logic, a later operator re-ban cannot
+    /// downgrade it to an evictable `Operator` entry. This path consumes
+    /// no new slot and is never capacity-rejected.
     ///
     /// New-contract inserts are bounded by [`Self::max_banned`]
     /// (production: [`MAX_BANNED_CONTRACTS`]). A slot is reserved with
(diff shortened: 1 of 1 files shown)


sanity:
## Multi-model review summary

Full-tier change (concurrency + anti-abuse surface). Review path:

- **Codex** (`codex review --base main`): unavailable — hit usage limit (resets Jun 10 21:21). Tried first, failed with an explicit quota error.
- **Gemini CLI** (external, non-Claude): ran two adversarial passes against the checked-out diff.
- **Claude code-first self-read**: counter-accounting and loop-termination audit (below).

### Round 1 (Gemini) — findings, all addressed in `945207ce`

1. **[CRITICAL] Priority inversion.** The re-ban merge upgraded the stored reason to `Operator` when an operator re-banned an existing `AutoMad` entry. Because only `Operator` entries are evictable, that let an operator re-ban (or an attacker reaching the CLI) downgrade a node-made security ban into an evictable one — the opposite of the issue's "reaper bans take precedence". Fixed: `AutoMad` now dominates on re-ban (entry becomes/stays `AutoMad` if either side is `AutoMad`). The pre-existing `operator_reason_wins_over_auto` test encoded the old (now-wrong) semantics; replaced with `auto_reason_dominates_operator_on_reban` asserting an operator re-ban cannot downgrade an `AutoMad` entry.
2. **[MEDIUM] Stack-overflow risk.** The AutoMad eviction path recursed up to (#Operator entries) deep. Converted to a `loop` — O(1) stack, same termination guarantee.
3. **[LOW] Counter drift between `size` and `entries.len()`** during the reserve-then-insert window. Not applicable / by-design: this is the same pessimistic-reservation behaviour the precedent `UpdateRateLimiter` uses; it is transient and capacity-safe (never lets the cap overshoot).

### Round 2 (Gemini, re-review of the updated diff)

Confirmed all three fixes correct and complete; verified deadlock avoidance (`drop(e)` before the `retain` walk), counter integrity across all paths, and no overshoot under the concurrent flood. Approved. Residual nits (O(N) single-victim eviction at N≤8192; `AutoMad` re-ban with an earlier expiry keeps the later expiry — conservative/safe) accepted as-is.

### Loop-termination audit (Claude)

The `loop` only re-iterates when `reason == AutoMad` AND `evict_one_operator()` actually removed an entry. While at capacity, `Operator` adds are themselves rejected without inserting, so the supply of evictable `Operator` entries is monotonically non-increasing — the loop cannot spin indefinitely. Every `fetch_sub` (`unban`/`cleanup`/eviction) corresponds to a real prior removal, so the counter cannot underflow.

Tests: 28/28 in the module pass; `cargo fmt --check` and CI-equivalent `cargo clippy --locked -- -D warnings` clean.

Diff reviewed: `/tmp/4303-diff2.txt` (HEAD = `945207ce`).

[AI-assisted - Claude]

PR #4299, opened 2026-05-28
feat(governance): phase 7 — repeat-offender ban TTL with wire-boundary drop

sanity:
## Problem

After Phase 4 (`GovernanceManager` state machine with `Normal → Borderline → WouldEvict → Evicted → Banned`), the `Banned` state had no enforcement: a contract could be flagged, evicted, re-flagged, and transitioned into `Banned` — and nothing on the wire would change. Repeat offenders could continue to send PUT/GET/UPDATE/SUBSCRIBE traffic and burn resources.

## Approach

Add a per-node `ContractBanList` keyed by `ContractInstanceId`, populated by the governance reaper loop on `TransitionReason::BanTriggered` and drained on `BanLifted`. The wire-boundary dispatch site in `node.rs` consults the list and drops inbound REQUEST messages for banned contracts:

- `PutMsg::Request` / `PutMsg::RequestStreaming`
- `GetMsg::Request`
- `UpdateMsg::*` (all 4 wire variants — gate runs before the Phase 2 rate limiter so banned traffic doesn't consume the rate-limit budget)
- `SubscribeMsg::Request`

**What is NOT blocked:** responses to our own outbound requests — letting them through lets in-flight transactions complete cleanly. Future Phase 7 follow-up may add egress self-blocking.

**Wiring:** `Ring::apply_ban_decisions` extracted from `governance_reaper_loop` so the bridge between `ReaperDecision` and `ContractBanList` mutations is directly testable. The reaper calls it once per tick after `ContractBanList::cleanup()` (defense-in-depth against stale entries from a config reload).

## Testing

- **8 unit tests** on `ContractBanList`: ban/unban, expiry boundary (`<` not `<=`), re-ban with later vs earlier expiry (must not shorten), operator reason wins over auto, cleanup drops only expired, distinct contracts independent.
- **4 wiring tests** on `Ring::apply_ban_decisions`: BanTriggered adds, BanLifted removes, unrelated transitions are no-ops, mixed batch applies in order.
- Full lib build clean, clippy clean (\`-D warnings\`), all 28 \`contract::governance\` tests still pass.

End-to-end simulation test (inject May 21 pattern → assert rate-limit → MAD → eviction → ban chain) deferred to a follow-up.

Closes part of #4296 (Phase 7 checkbox).

[AI-assisted - Claude]

PR #4371, opened 2026-06-09
feat(governance): egress self-block for banned contracts at originator entry-points

### LINKED ISSUE #4300: Phase 7 follow-up: egress self-blocking for banned contracts
## Context

Spawned out of multi-model review on PR #4299 (Phase 7 ban enforcement — wire-boundary REQUEST drop).

PR #4299 lands the receive-side half of Phase 7: a node consults its local \`ContractBanList\` at the wire dispatch site and drops inbound PUT/GET/UPDATE/SUBSCRIBE Request variants for banned contracts. The two egress paths that proactively pushed state for already-hosted contracts (\`NeighborHosting\` overlap-sync and \`InterestSync\` stale-summary repair) are also gated.

What is **not yet gated** is **outbound REQUEST-initiation** for a self-banned contract:

- A local client of this node submits a PUT/GET/UPDATE/SUBSCRIBE for a contract this node has banned. The originator drives \`start_relay_*\` which sends a REQUEST message to a peer. The peer doesn't know about our ban and processes it normally.
- A local UPDATE that this node applies (e.g. delegate-driven) drives \`BroadcastStateChange\` → outbound \`UpdateMsg::BroadcastTo\` fans out to subscribers. The subscribers don't know about our ban.

## Proposed approach

Add an egress check at the originator entry-points (\`start_client_put\`, \`start_client_get\`, \`start_client_subscribe\`, \`start_client_update\`, and the BroadcastStateChange driver) that consults \`contract_ban_list.is_banned()\` and returns a typed \`RequestError::ContractBanned\` to the client (or skips the fan-out, with a debug log, for delegate-driven UPDATEs).

This complements the receive-side gate by ensuring a banned contract:
1. Cannot receive new state via this node (receive gate — already done).
2. Cannot transmit new state via this node (egress gate — this issue).
3. Cannot propagate stale-repair state via overlap-sync (egress gate — already done in #4299).

## Spawned from

- Code-first review (#3 — Local-client UX is unaddressed)
- Skeptical review (#1 — Local-client requests for self-banned contracts get silently black-holed)
- Big-picture review (egress self-blocking explicit follow-up)

[AI-assisted - Claude]

sanity:
## Problem

PR #4299 landed the **receive-side** half of Phase 7 ban enforcement: a node consults its local `ContractBanList` at the wire dispatch site and drops inbound PUT/GET/UPDATE/SUBSCRIBE `Request` variants for banned contracts, and gates the two proactive state-egress paths for already-hosted contracts (`NeighborHosting` overlap-sync, `InterestSync` stale-summary repair).

What was **not** gated is **outbound REQUEST-initiation** for a self-banned contract:

- A local client of this node submits a PUT/GET/UPDATE/SUBSCRIBE for a contract this node has banned. The originator drives `start_relay_*`, sending a REQUEST to a peer that doesn't know about our ban and processes it normally.
- A locally-applied UPDATE (e.g. delegate-driven) drives `BroadcastStateChange` → outbound fan-out to subscribers who don't know about our ban.

Net effect: the ban was only a receive-side filter, not a true block. The node could launder requests for a contract it had decided is harmful, burning outbound resources, and local clients got a silent proceed-then-timeout instead of a clear error.

## Approach

Mirror the receive-side gate on the **egress** side, at the precise call sites the issue names:

- New typed `OpError::ContractBanned { instance_id }` and a shared `operations::reject_if_contract_banned` helper (keyed on `ContractInstanceId`, which is what the ban list keys on and what every entry point has available).
- `start_client_put` / `start_client_get` / `start_client_subscribe` / `start_client_update` call the helper **before spawning the driver task**, returning the typed error. The client gets a clear, contract-named reason via `result_router_tx` instead of a silent timeout.
- `report_op_init_error` (client_events.rs) maps `ContractBanned` → `ErrorKind::OperationError`. I deliberately did **not** add a `RequestError::ContractBanned` stdlib wire variant: that would be a wire-format change requiring a separate stdlib-first release (per the repo's stdlib-first policy and the high-bar-for-protocol-changes rule), and `OperationError` is the existing, wire-stable mapping the receive-side path already uses.
- `handle_broadcast_state_change` (p2p_protoc.rs) skips the delegate-driven UPDATE fan-out — **and its no-target retry re-emission** — for a banned contract, with a debug log (no client to notify on the fire-and-forget broadcast path).

**Scope:** renewals (`run_renewal_subscribe`) and executor-auto-subscribe (`run_executor_subscribe`) go through their own drivers and are intentionally **not** gated — the egress self-block is about new client-originated requests, not maintenance of already-established subscriptions.

Together with the receive-side drops, a banned contract can now neither receive new state via this node nor transmit new state via it.

## Testing

- **Source-grep pin tests on all five egress sites** (mirroring the existing receive-side pins in `ring/contract_ban_list.rs`): each `start_client_*` must call `reject_if_contract_banned` before `GlobalExecutor::spawn`, and `handle_broadcast_state_change` must gate before `get_broadcast_targets_update`. These fail CI if a gate is later deleted or reordered after the spawn/fan-out — the same "CI fails if the gate is deleted" guarantee PR #4299 established.
- **Behavioral unit tests** for the `OpError::ContractBanned` → `ErrorKind::OperationError` mapping (cause string names the ban), plus a pin that `report_op_init_error` handles the variant.
- **Failing-first verified**: temporarily removed the PUT entry-point gate and the broadcast gate; confirmed both pins fail with clear messages, then restored.
- `cargo fmt` clean, `cargo clippy --features testing -- -D warnings` clean, all `operations::*::op_ctx_task`, `client_events`, `p2p_protoc`, and `contract_ban_list` unit tests pass (including the unchanged receive-side pins).

Closes #4300

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -336,7 +336,14 @@ async fn report_op_init_error(
         // so this also serves as the source-of-truth for the
         // user-visible shape of `OpError::NodeShuttingDown`.
         OpError::NodeShuttingDown => ErrorKind::Shutdown,
-        OpError::RingError(crate::ring::RingError::ConnError(_))
+        // Phase 7 egress self-block (#4300): a local client originated a
+        // request for a contract this node has banned. Surface as a
+        // typed `OperationError` (no dedicated wire variant exists in
+        // stdlib, and adding one would be a wire-format change requiring
+        // a separate stdlib-first release) so the client sees a clear,
+        // contract-named reason instead of a silent timeout.
+        OpError::ContractBanned { .. }
+        | OpError::RingError(crate::ring::RingError::ConnError(_))
         | OpError::RingError(crate::ring::RingError::NoHostingPeers(_))
         | OpError::ConnError(_)
         | OpError::ContractError(_)
@@ -2343,5 +2350,46 @@ pub(crate) mod test {
              failure if the shutdown signal is swallowed into the \
              catch-all."
         );
+    }
+
+    /// Phase 7 egress self-block (#4300): `OpError::ContractBanned`
+    /// (returned by `start_client_*` when the local client originates a
+    /// request for a banned contract) must map to a client-visible
+    /// `ErrorKind::OperationError` whose cause names the contract — NOT a
+    /// silent proceed-then-timeout. Replicates the relevant arm of
+    /// `report_op_init_error` to avoid constructing the full `OpManager`
+    /// fixture (same approach as the NodeShuttingDown pin above). The
+    /// production match is exhaustive, so a missing arm fails to compile;
+    /// this test additionally locks the user-visible shape.
+    #[test]
+    fn op_error_contract_banned_maps_to_operation_error_with_cause() {
+        use crate::operations::OpError;
+        use freenet_stdlib::client_api::ErrorKind;
+        use freenet_stdlib::prelude::ContractInstanceId;
+
+        let instance_id = ContractInstanceId::new([7u8; 32]);
+        let err = OpError::ContractBanned { instance_id };
+        let op_name = "PUT";
+        let kind = match err {
+            OpError::ContractBanned { .. } => ErrorKind::OperationError {
+                cause: format!("{op_name} operation failed: {err}").into(),
+            },
+            // Other variants are irrelevant to this test.
+            _ => ErrorKind::Unhandled {
+                cause: "irrelevant for this test".to_string().into(),
+            },
+        };
+        match kind {
+            ErrorKind::OperationError { cause } => {
+                assert!(
+                    cause.contains("banned"),
+                    "ContractBanned cause must mention the ban so the \
+                     client understands why the request was rejected; got: {cause}"
+                );
+            }
+            other => panic!(
+                "OpError::ContractBanned must map to ErrorKind::OperationError, got {other:?}"
+            ),
+        }
     }
 }
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -6162,5 +6162,31 @@ mod tests {
              contract's broadcast is skipped — including the no-target retry \
              re-emission. Body:\n{body}"
         );
+
+        // Regression guard for the `fix(governance): clear broadcast retry
+        // state on banned-egress skip` commit: the banned-skip branch must
+        // also clear the per-contract retry/streak bookkeeping, otherwise a
+        // contract banned mid-retry-cycle leaves a stale entry that nothing
+        // clears until it is unbanned and broadcast again. We scope the
+        // assertion to the banned-skip branch (between the gate and its
+        // early `return;`) so deleting either `remove()` call fails this
+        // test — the gate-ordering assertion above passes even without
+        // them.
+        let banned_branch = &body[gate_pos..];
+        let return_pos = banned_branch
+            .find("return;")
+            .expect("banned-egress branch must early-return");
+        let banned_branch = &banned_branch[..return_pos];
+        assert!(
+            banned_branch.contains("self.broadcast_retries.remove(&key)"),
+            "banned-egress skip must clear broadcast_retries for the contract \
+             so a ban mid-retry-cycle doesn't leave a stale entry. Branch:\n{banned_branch}"
+        );
+        assert!(
+            banned_branch.contains("self.broadcast_no_target_streak.remove(&key)"),
+            "banned-egress skip must clear broadcast_no_target_streak for the \
+             contract so a ban mid-retry-cycle doesn't leave a stale entry. \
+             Branch:\n{banned_branch}"
+        );
     }
 }
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -4286,6 +4286,14 @@ impl P2pConnManager {
                 phase = "broadcast_state_change_banned_skip",
                 "skipping state-change broadcast for banned contract"
             );
+            // Drop any in-flight retry/streak bookkeeping for this
+            // contract: if it was banned mid-retry-cycle, a previously
+            // scheduled re-emission would otherwise leave a stale entry
+            // here that nothing clears until the contract is unbanned and
+            // broadcast again. Mirrors the cleanup the targets-found and
+            // retries-exhausted paths perform.
+            self.broadcast_retries.remove(&key);
+            self.broadcast_no_target_streak.remove(&key);
             return;
         }
 
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -4269,6 +4269,26 @@ impl P2pConnManager {
             state_size = new_state.size(),
             "BroadcastStateChange event received"
         );
+
+        // Phase 7 egress self-block (#4300). A locally-applied UPDATE
+        // (e.g. delegate-driven) drives this fan-out to subscribers who
+        // don't know about our ban. If we have banned the contract, skip
+        // the fan-out entirely — and the no-target retry re-emission with
+        // it — rather than push state for a contract we have decided is
+        // harmful. There is no client to notify here (this is the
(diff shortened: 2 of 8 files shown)


sanity:
## Review (Full tier — touches operation entry-points)

**External models unavailable.** I attempted both external reviewers and both were genuinely down:
- `codex review --base main` → "You've hit your usage limit" (resets Jun 10 ~9:21 PM CT).
- `gemini` → 429 `QUOTA_EXHAUSTED` (resets in ~21h).

Per `~/.claude/rules/multi-model-review.md`, since waiting until both reset isn't practical for this autonomous task, I fell back to **diverse Claude-lens self-review** — three distinct adversarial lenses, each reading the checked-out code. Recording the substitution here for auditability.

### Lens 1 — skeptical bug-hunt
- **Found + fixed:** `handle_broadcast_state_change` returned early on a banned contract without clearing `broadcast_retries` / `broadcast_no_target_streak`. If a contract is banned mid-retry-cycle, a previously scheduled re-emission would leave a stale entry that nothing clears until unban+rebroadcast. Fixed in fda166de by removing both entries in the banned-skip path (mirrors the targets-found / retries-exhausted cleanup).
- No correctness or security bug in the gate logic itself. The `contract.key().id()` temporary borrow in `start_client_put` is sound (temporary lives to end of statement; borrow does not escape).

### Lens 2 — code-first intent-vs-impl
- Verified the typed `OpError::ContractBanned` reaches the client on **all four** ops. PUT/GET/UPDATE route the entry-point `Err` through `report_op_init_error` → `ErrorKind::OperationError`. SUBSCRIBE differs: `subscribe_with_id`'s error propagates via `?` out of `process_open_request` and is caught by the catch-all at `client_events.rs:465`, also mapped to `ErrorKind::OperationError`. Consistent across all paths; no silent swallow.
- The SUBSCRIBE WASM-cache pre-check (`GetQuery`) before `subscribe_with_id` is a local store query, not network egress — no egress happens before the gate.

### Lens 3 — failure mode specific to this change
- **All listed entry points covered:** `start_client_put` / `_get` / `_subscribe` / `_update` + `handle_broadcast_state_change`. Confirmed by grepping every client-origination call site in `client_events.rs` (lines 716/891/956/1235 + `subscribe_with_id`) and the executor's `BroadcastStateChange` emits (`runtime.rs:2442/2806`).
- **Streaming variants covered:** the streaming-vs-non-streaming decision lives *inside* the drivers, downstream of the single `start_client_*` entry point, so gating at the entry covers both (matches the receive side gating both PUT variants + all four UPDATE wire variants).
- **Typed error consistent with the receive side:** receive side drops silently at the wire boundary (no client); egress side returns `ErrorKind::OperationError` to the local client. No `RequestError::ContractBanned` stdlib variant added — that would be a wire-format change requiring a separate stdlib-first release (per repo policy); `OperationError` is the existing wire-stable mapping.
- **Correctly out of scope:** renewals (`run_renewal_subscribe`), executor-auto-subscribe (`run_executor_subscribe`), and sub-op GETs are not gated — the egress self-block is about new client-originated requests, not maintenance of already-established state.

### Tests
- Source-grep pin tests on all five egress sites (fail CI if a gate is deleted/reordered after the spawn/fan-out); behavioral mapping tests for `ContractBanned` → `ErrorKind::OperationError`. Failing-first verified by removing the PUT and broadcast gates (both pins fail with clear messages), then restored.
- `cargo fmt` clean, `cargo clippy --features testing -- -D warnings` clean, all touched-module unit tests pass.

When codex/gemini quota resets I can re-run the real external pass before merge if preferred, given this is a Full-tier change.

[AI-assisted - Claude]

sanity:
Addressed the `rule-review/warnings` finding in 83bb882b:

- **Warning (missing regression test for the stale-retry-state fix):** extended `handle_broadcast_state_change_gates_banned_egress` to assert both `self.broadcast_retries.remove(&key)` and `self.broadcast_no_target_streak.remove(&key)` appear inside the banned-skip branch (scoped between the `is_banned` gate and its early `return;`). The previous pin only checked gate ordering and passed even if the cleanup calls were deleted. Verified failing-first: removing either `remove()` call now fails the test.
- **Info (`# Errors` doc section):** added a `# Errors` section to `operations::reject_if_contract_banned`.

The change since the last review is purely additive (one test assertion + a doc comment) — no production-logic change.

[AI-assisted - Claude]

sanity:
## Status: ready, holding for external review

**CI: fully green** on current HEAD `83bb882b` — Clippy, Fmt, Unit & Integration, Simulation, NAT Validation, Windows Check, Claude Rule Review, `rule-review/warnings`, Rule Lint, Conventional Commits, macOS E2E, Snyk, CLA all SUCCESS. The earlier `rule-review/warnings` failure is resolved.

**Not auto-merging.** This touches operation entry-points (a Full-tier, high-risk surface), and the independent **external** (non-Claude) review pass — the highest-value reviewer for high-risk changes — could not run: both `codex` (usage limit, resets Jun 10 ~9:21 PM CT) and `gemini` (quota exhausted, ~21h reset) were down. I completed the documented diverse-Claude-lens fallback (3 adversarial lenses, posted above), but per the multi-model-review rule I'm leaving the merge decision to a maintainer so the real external pass can run first if desired (e.g. `codex review --base main` once quota resets).

If a maintainer is comfortable merging on the Claude-lens review given the change is mechanical (mirrors the already-landed receive-side gate), it's ready to go.

[AI-assisted - Claude]

sanity:
## Independent multi-perspective review — SHIP

An independent **3-perspective Claude review panel** reviewed the checked-out code at HEAD (`acdf93f47852e86c1fb43ea343eaf223497c0a7a`). All three lenses returned **SHIP**:

1. **Skeptical bug-hunt** — assume bugs exist; probe the egress gate, the broadcast retry-state cleanup, and the cancellation/early-return shape across the four originator entry points.
2. **Code-first intent-vs-impl** — read the code before the PR description; check that the gate's actual behavior matches the stated #4300 intent (egress self-block complementary to the #4299 receive-side drop).
3. **Security / failure-mode** — focus on the ban-enforcement boundary: can a banned contract still transmit state through any originator path; does the gate fail open under any error.

### External models unavailable — documented fallback

External non-Claude reviewers (`codex` and `gemini`) were unavailable this batch — real invocations failed (quota / capacity exhaustion), not an assumption. Per the documented fallback in the multi-model-review rule, the independent pass used **diverse Claude lenses** (the three above) instead of a single Claude reviewer, to compensate for the lost model diversity with perspective diversity. This substitution was user-authorized for this batch. When the external models return and a re-review is warranted, the real external pass should be run as well.

### Actionable finding — now closed

The panel's one actionable finding was a **behavioral-coverage gap on `reject_if_contract_banned`**: the existing tests were source-scrape pins (proving each `start_client_*` entry point *calls* the gate) plus a mapping test (`OpError::ContractBanned` → `ErrorKind`). None drove the gate's actual decision against a real `ContractBanList`, so an inverted predicate, a wrong-id lookup, or a swapped `Ok`/`Err` return would have passed every existing test.

**Closed in this PR by commit `acdf93f47852e86c1fb43ea343eaf223497c0a7a`.** It extracts `reject_if_contract_banned_on(&ContractBanList, &id)` from the `OpManager`-taking wrapper (matching the `*_on` extraction convention in `op_state_manager.rs`) so the gate is unit-testable without standing up a full `OpManager`, then adds four behavioral tests against the real `is_banned`-backed logic: banned id rejected with the typed error carrying that id, unbanned id passes, ban scoped to the specific id, and expired ban no longer rejects. The four source-scrape pins still pass unchanged (the entry points still call the `reject_if_contract_banned` wrapper).

### Non-blocking items noted

- **(a) SUBSCRIBE error path leaves a result waiter uncleaned** — pre-existing shape, **not introduced by this PR**. Noted for awareness; out of scope here.
- **(b) Renewal-subscribe egress leak** — a contract banned *while* this node holds an active subscription keeps emitting outbound SUBSCRIBE renewals (no state push; time-bounded by the ban TTL) because the renewal scheduler doesn't pass through the egress gate and `apply_ban_decisions` doesn't clear `active_subscriptions`. This is out of #4300's named scope (no state egress, time-bounded). **Filed as follow-up #4373.**

`cargo fmt` and `cargo clippy --locked -- -D warnings` (CI's exact gate) are clean; the new + existing banned/egress tests pass. Enabling squash auto-merge.

[AI-assisted - Claude]

PR #4372, opened 2026-06-09
fix(server): guard web contract unpack against path traversal

### LINKED ISSUE #3946: Tar unpack of web contract lacks path-traversal guards

sanity:
## Problem

`WebApp::unpack` (`crates/core/src/server/app_packaging.rs`) extracted the embedded tar archive of a web contract with tar's default `Archive::unpack(dst)`. That default:

- only *silently skips* entries whose path contains `..` or is absolute (returns `Ok`, leaving no signal to the operator that a malicious archive was processed),
- relies on tar-rs internal canonicalization to catch symlink/hardlink escapes,
- leaves overwrite **enabled**, so a republished contract can clobber existing cache files.

Web contracts are authored by untrusted third parties, so the archive is adversarial input. A malicious author can hand-craft tar bytes whose entries escape `webapp_cache_dir()` and write arbitrary files on the **node operator's filesystem** (CWE-22 path traversal / zip-slip — e.g. overwriting shell configs or SSH `authorized_keys`). Before PR #3942 this required a user to navigate to the malicious contract's root; after #3942 it is reachable whenever any page links to `/v1/contract/web/<malicious-key>/anything`, which triggers the cold-cache unpack.

Flagged during review of #3942.

## Solution

Stop delegating containment entirely to tar's defaults and validate every entry ourselves before it touches the filesystem:

- **Reject** (hard `UnpackingError`, not a silent skip) any entry with an absolute path or a `..` / prefix component, so a malicious archive surfaces as an error to the operator instead of partially extracting.
- **Reject** symlink and hardlink entries outright — web apps have no legitimate need for on-disk links, and a symlink-then-write-through is the classic zip-slip escalation.
- Keep `entry.unpack_in(dst)` as a defense-in-depth second check (it performs canonicalization-based containment).
- `set_overwrite(false)` so a crafted duplicate entry can't clobber files already at the destination; `set_preserve_mtime(false)` to avoid honoring attacker-controlled mtimes.

This matches the fix sketch in the issue, hardened to reject (rather than skip) and to refuse link entries. Behavior for legitimate archives is unchanged.

## Testing

Added four unit tests in `app_packaging.rs`. The three adversarial archives are constructed by writing the tar header's name field **directly**, deliberately bypassing `tar::Builder`'s append-time path validation — `Builder::append_data` refuses to *produce* `..`/absolute paths, but a real attacker writes the bytes directly with any tool, so this faithfully reproduces the on-the-wire attack:

- `unpack_rejects_parent_dir_traversal` — `../escape.txt` entry must error and write nothing outside `dst`.
- `unpack_rejects_absolute_path` — absolute-path entry must error and write nothing outside `dst`.
- `unpack_rejects_escaping_symlink` — symlink pointing outside `dst` must error and not be created.
- `unpack_accepts_legitimate_nested_entries` — a well-formed `index.html` + `assets/app.js` archive still unpacks correctly.

Verified the three escape tests **FAIL** against the unfixed code (all return `Ok` / create the symlink) and **PASS** after the fix; the legitimate-archive test passes throughout. `cargo fmt` and `cargo clippy --locked -- -D warnings` (the CI invocation) are clean.

Closes #3946

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/app_packaging.rs
+++ b/crates/core/src/server/app_packaging.rs
@@ -68,13 +68,84 @@ impl WebApp {
         Ok(output)
     }
 
+    /// Unpacks the web archive into `dst`, confining every written file to that
+    /// directory.
+    ///
+    /// Web contracts are authored by untrusted third parties, so the embedded
+    /// tar archive is adversarial input. A malicious author can hand-craft tar
+    /// bytes whose entry paths contain `..`, are absolute, or are symlinks /
+    /// hardlinks pointing outside `dst`; left unchecked, those entries let the
+    /// archive write arbitrary files on the node operator's filesystem (CWE-22
+    /// path traversal / zip-slip). See freenet/freenet-core#3946.
+    ///
+    /// We therefore validate each entry ourselves and **reject** (rather than
+    /// silently skip) anything that could escape the destination:
+    ///
+    /// - absolute paths,
+    /// - any component equal to `..` (`ParentDir`),
+    /// - symlink and hardlink entries (web apps have no legitimate need for
+    ///   them, and they are the primary symlink-escape vector).
+    ///
+    /// Overwrite is disabled so a republished contract cannot clobber files
+    /// already present at the destination via a crafted duplicate entry.
     #[instrument(level = "debug", skip(self, dst))]
     pub fn unpack(&mut self, dst: impl AsRef<Path>) -> Result<(), WebContractError> {
-        debug!("Unpacking web content to {:?}", dst.as_ref());
+        use std::path::Component;
+        use tar::EntryType;
+
+        let dst = dst.as_ref();
+        debug!("Unpacking web content to {:?}", dst);
+
         let mut decoded_web = self.decode_web();
-        decoded_web
-            .unpack(dst)
-            .map_err(WebContractError::StoringError)?;
+        decoded_web.set_overwrite(false);
+        decoded_web.set_preserve_mtime(false);
+
+        let entries = decoded_web
+            .entries()
+            .map_err(|e| WebContractError::UnpackingError(anyhow::anyhow!(e)))?;
+
+        for entry in entries {
+            let mut entry =
+                entry.map_err(|e| WebContractError::UnpackingError(anyhow::anyhow!(e)))?;
+
+            // Reject link entries outright: a symlink/hardlink whose target is
+            // outside `dst` (or a symlink followed by a later write-through
+            // entry) is the classic zip-slip escape, and a web app never needs
+            // links on disk.
+            let entry_type = entry.header().entry_type();
+            if entry_type == EntryType::Symlink || entry_type == EntryType::Link {
+                let path = entry.path().map(|p| p.to_path_buf()).unwrap_or_default();
+                return Err(WebContractError::UnpackingError(anyhow::anyhow!(
+                    "refusing to unpack link entry from web archive: {path:?}"
+                )));
+            }
+
+            let path = entry
+                .path()
+                .map_err(|e| WebContractError::UnpackingError(anyhow::anyhow!(e)))?;
+
+            // Reject absolute paths and any `..` component before touching the
+            // filesystem. `tar`'s own `unpack_in` would merely *skip* such
+            // entries; we want a hard error so a malicious archive is visible
+            // to the operator instead of partially extracting.
+            if path.is_absolute()
+                || path
+                    .components()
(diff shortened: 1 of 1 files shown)


sanity:
## Review (Full tier — security fix, high-risk surface)

This change touches a security-relevant surface (untrusted-archive extraction), so it is Full review tier.

### Review path

External non-Claude models were attempted first and are **all currently unavailable**:

- **Codex** (`codex review --base main`, model gpt-5.5): `ERROR: You've hit your usage limit ... try again at Jun 10th, 2026 9:21 PM`.
- **Gemini** (`gemini-2.5-pro` and `gemini-2.5-flash`): HTTP 429 `QUOTA_EXHAUSTED` on every retry (`quota will reset after ~20h44m`).

Both resets are far off and this is an autonomous run, so per the multi-model-review fallback policy I used **diverse Claude self-review passes** with three distinct adversarial lenses instead of a single pass. (I cannot spawn parallel reviewer subagents in this context, so the passes were run serially.) When an external model is back before merge, a real external pass is still worth running given the risk tier.

### Lens 1 — Skeptical bug-hunt
- Verified there is **no PAX/GNU-long-name bypass**: `tar`'s `Entry::path()` resolves the path via `path_bytes()`, which prefers the GNU `long_pathname`, then the PAX `path=` extension, then the header name — and `unpack_in` writes using that *same* resolved path. So the `..`/absolute check at the validation site and the actual write target cannot diverge (`tar-0.4.46/src/entry.rs:308-327`, `:398`).
- Fail-fast on the first malicious entry (whole archive refused) is intentional and safer than partial extraction.

### Lens 2 — Code-first intent-vs-impl
- `set_overwrite(false)` does not regress legitimate re-publishes: the caller `unpack_if_stale` does `remove_dir_all` + `create_dir_all` before `unpack` (`path_handlers.rs:406-411`), so the destination is always empty; `overwrite(false)` only blocks *duplicate entries within one archive* — exactly the attack surface.
- `.` (`CurDir`) and nested dirs are still accepted; the legitimate-archive test confirms normal unpack is unchanged.

### Lens 3 — Security / zip-slip failure-mode
- Symlink-dir + nested-write (entry A = symlink out, entry B writes through it): blocked — all `Symlink`/`Link` entries are rejected before any write.
- Hardlink to an outside file: blocked (same check).
- Windows absolute / drive-prefix (`C:\…`): blocked via `Component::Prefix` and `is_absolute()`.
- `..` anywhere in the path (`foo/../../etc`): blocked via `components().any(ParentDir)`.
- Defense in depth: `entry.unpack_in(dst)` still performs canonicalization-based containment as a second layer.

### Outcome
No blocking findings. The regression tests fail against the unfixed code and pass after the fix (verified locally); `cargo fmt` and `cargo clippy --locked -- -D warnings` (the exact CI invocation) are clean.

[AI-assisted - Claude]

PR #4374, opened 2026-06-10
fix(transport): fail stream not connection on outbound-stream failure

sanity:
## Problem

Several failure sites in `outbound_stream.rs` returned `TransportError::ConnectionClosed(destination_addr)`. The connection recv loop in `peer_connection.rs` (~line 1162) classifies any error that is not `is_transient_send_failure()` as fatal:

```rust
Err(e) if e.is_transient_send_failure() => { warn; op layer retries }  // connection SURVIVES
Err(e) => return Err(e),                                                // connection TORN DOWN
```

`is_transient_send_failure()` previously matched only `SendFailed`, so each of these stream-level failures hit the fatal arm and tore down the **entire** connection — killing every other operation multiplexed on it and forcing a re-handshake. A single stalled or failed stream became a connection-teardown amplifier. This is the user-visible "cwnd wait timeout" / "no fragments received" failure mode from #4345.

The affected sites — all problems with *one transfer or its upstream inbound feed*, none meaning the downstream socket/connection itself is dead:

- **cwnd-wait timeout** in `send_stream` and `pipe_stream` — the congestion-window wait exceeds `CWND_WAIT_TIMEOUT` (ACKs stopped arriving for this transfer).
- **relay-pipe inactivity stall** (`pipe_stream` :456) — the UPSTREAM inbound feed produced no fragment within `STREAM_INACTIVITY_TIMEOUT`.
- **relay-pipe inbound-stream error** (`pipe_stream` :471) — the UPSTREAM inbound stream yielded an error.

The pipe paths are arguably more central to the room-join GET path (#3586) than cwnd-wait, since they kill a connection carrying other multiplexed ops purely as collateral from an upstream problem.

Per #4345, a single stalled/failed stream must fail just the STREAM (the op layer already times out and retries against another candidate after PR #4367), NOT the connection.

## Solution

- Add a stream-scoped `TransportError::OutboundStreamFailed(SocketAddr)` variant (covers timeout AND error; generic message `"outbound stream to {0} failed (stream-scoped, connection survives)"`).
- Convert **all** stream-scoped sites in `outbound_stream.rs` from `ConnectionClosed` to `OutboundStreamFailed`: the cwnd-wait timeout in both `send_stream` and `pipe_stream`, the relay-pipe **inactivity stall**, and the relay-pipe **inbound-stream error**. Existing `emit_transfer_failed(...)` telemetry is unchanged.
- Audit result: the two remaining `return Err(e)` sites (`send_stream` :295, `pipe_stream` :618) propagate `packet_sending`'s error, which is already `SendFailed` (transient) — left as-is. No site in this file genuinely means "the connection itself is dead"; `ConnectionClosed` here was always a misnomer for stream-scoped failures.
- Include the new variant in `is_transient_send_failure()` so the recv loop logs and lets the op layer retry while the connection stays up. The idle timeout remains the sole authority on connection liveness. Updated the doc comment, all three exhaustive `TransportError` match arms in `connection_handler.rs`, and all test references.
- **Review nit:** documented at the `send_stream` cwnd-abort site that `completion_tx` is intentionally dropped (not signaled) — `broadcast_queue` awaits it with a timeout and treats the resulting oneshot `RecvError` as completion (`broadcast_queue.rs` `Ok(Err(_))` arm), releasing the permit immediately. A future editor must not "fix" it with a blocking `tx.send(())` (channel-safety.md).

This is the **connection-survival half only**. It does not change any timeout constants and does not attempt the transport flightsize-by-stream-id rework — that root cause remains open in #4345.

## Testing

Three regression tests drive the **real** `send_stream`/`pipe_stream` functions to each failure mode and assert the error IS `OutboundStreamFailed`, is NOT `ConnectionClosed`, and `is_transient_send_failure()` is `true` (the exact condition the recv loop branches on for connection survival):

- `cwnd_wait_timeout_fails_stream_not_connection` — 1-byte cwnd so no packet ever fits → cwnd-wait timeout in `send_stream`.
- `pipe_stream_inactivity_stall_fails_stream_not_connection` — `StreamHandle` with no fragment pushed so `stream.next()` stays Pending → inactivity-timeout arm (:456).
- `pipe_stream_inbound_error_fails_stream_not_connection` — cancel the `StreamHandle` so the inbound stream yields `Err(Cancelled)` → inbound-error arm (:471).

**Before/after verified for all three** (temporarily reverted the production sites to `ConnectionClosed`): each fails on the old behavior (`got: ConnectionClosed`) and passes with the fix.

Also updated the two existing cwnd-timeout tests (`test_send_stream_cwnd_wait_timeout`, `test_pipe_stream_cwnd_wait_timeout`) to expect the new variant. The `test_pipe_stream_cwnd_wait_timeout` rewire (prior commit) had found that test was misconfigured — it used `CongestionControlConfig::default()` (FixedRate, `current_cwnd() == usize::MAX/2`), so the cwnd loop never blocked and the test silently exercised the *inactivity-timeout* path rather than the cwnd-wait path it claims to test; switched it to a 1-byte-cwnd LEDBAT controller so it genuinely hits the cwnd-wait site. The two new pipe tests restore explicit coverage of the inactivity and inbound-error paths.

A full `PeerConnection::recv` teardown harness would need substantial new scaffolding the stream-level unit harness doesn't expose; the error-classification assertions pin the behavior the recv loop branches on (mirrors the existing `send_failure_returns_transient_error` test for `SendFailed`).

`cargo fmt` clean. `cargo test -p freenet --lib transport::peer_connection::outbound_stream` → **9 passed**. Note: local `cargo clippy` (rust-1.94.0) reports 13 pre-existing `wildcard_enum_match_arm` / `int_plus_one` errors in unrelated files (`sent_packet_tracker.rs`, `peer_connection.rs:3324`, `client_events.rs`); these are present on clean `main` too (verified via stash) and are a local-toolchain-version artifact — **zero new clippy findings introduced** by this change.

Addresses #4345 (connection-survival half; transport flightsize-by-stream-id root cause remains open).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/transport.rs
+++ b/crates/core/src/transport.rs
@@ -346,6 +346,8 @@ pub enum TransportError {
     ProtocolVersionMismatch { expected: String, actual: String },
     #[error("send to {0} failed: {1}")]
     SendFailed(SocketAddr, std::io::ErrorKind),
+    #[error("outbound stream to {0} aborted: cwnd wait timeout")]
+    OutboundStreamTimeout(SocketAddr),
     #[error(transparent)]
     IO(#[from] std::io::Error),
     #[error(transparent)]
@@ -357,11 +359,25 @@ pub enum TransportError {
 }
 
 impl TransportError {
-    /// Returns true if this error is a transient UDP send failure that should
-    /// not kill the connection. The idle timeout is the sole authority on
-    /// connection liveness.
+    /// Returns true if this error is a stream-scoped failure that must NOT
+    /// kill the connection. The idle timeout remains the sole authority on
+    /// connection liveness; a single stalled or failed stream only fails
+    /// that stream, and the operation layer times out and retries against
+    /// another candidate.
+    ///
+    /// Covers:
+    /// - [`TransportError::SendFailed`]: a transient UDP send error (e.g.
+    ///   ENETUNREACH) on a single packet.
+    /// - [`TransportError::OutboundStreamTimeout`]: the outbound stream's
+    ///   congestion-window wait exceeded `CWND_WAIT_TIMEOUT`. Tearing down
+    ///   the whole connection here would kill every other operation
+    ///   multiplexed on it and force a re-handshake (#4345). The connection
+    ///   stays up and the idle timeout decides liveness instead.
     pub fn is_transient_send_failure(&self) -> bool {
-        matches!(self, TransportError::SendFailed(..))
+        matches!(
+            self,
+            TransportError::SendFailed(..) | TransportError::OutboundStreamTimeout(..)
+        )
     }
 }
 
--- a/crates/core/src/transport.rs
+++ b/crates/core/src/transport.rs
@@ -346,8 +346,8 @@ pub enum TransportError {
     ProtocolVersionMismatch { expected: String, actual: String },
     #[error("send to {0} failed: {1}")]
     SendFailed(SocketAddr, std::io::ErrorKind),
-    #[error("outbound stream to {0} aborted: cwnd wait timeout")]
-    OutboundStreamTimeout(SocketAddr),
+    #[error("outbound stream to {0} failed (stream-scoped, connection survives)")]
+    OutboundStreamFailed(SocketAddr),
     #[error(transparent)]
     IO(#[from] std::io::Error),
     #[error(transparent)]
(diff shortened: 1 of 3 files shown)


sanity:
## Consolidated review — Full tier, multi-perspective (Claude-subagent fallback)

**Risk tier:** Full. Transport / connection-lifecycle / concurrency is a high-risk surface with prior production incidents, so this got the full multi-perspective treatment across two rounds.

### External-model availability (per the multi-model-review rule)
The independent non-Claude pass could not run — both external reviewers were genuinely down at review time, confirmed by direct probe:
- **Codex**: `You've hit your usage limit … try again at 9:21 PM` (CT).
- **Gemini**: `429 QUOTA_EXHAUSTED … reset after ~1h38m`.

Per the rule's fallback, I substituted **diverse Claude subagents with distinct adversarial lenses**, each reviewing the checked-out code and blind to the others. Recommend running a real Codex pass on this content before the release tag once quota resets (this is a Full-tier transport change).

### Round 1 — 5 lenses on the cwnd-wait fix
*code-first, skeptical, testing, big-picture, and a transport connection-lifecycle specialist.* Consensus: correct and well-scoped. Verified: `TransportError` is not a wire type (local error, carries `io::Error`/`bincode::Error` — not serializable), so the new variant is not a cross-version concern; no new flightsize leak (the stalled stream still drains via the existing #4353 abandonment path now that the connection survives); receiver-side orphan cleanup unchanged. The agent also caught + fixed a pre-existing test (`test_pipe_stream_cwnd_wait_timeout`) that was passing for the wrong reason (FixedRate cwnd never blocked → it hit the inactivity path, not cwnd-wait).

**Findings raised → addressed in round 2:**
- *(3 lenses)* The sibling `pipe_stream` sites — inbound inactivity-stall (`:456`) and inbound-stream-error (`:471`) — still returned `ConnectionClosed`, tearing down the **downstream** connection on an **upstream-feed** problem. Same #4345 class, arguably more central to the relay-GET path. → **Converted** both to the stream-scoped variant (finish-the-fix).
- *(testing)* The test rewire opened a coverage gap. → Added `pipe_stream_inactivity_stall_fails_stream_not_connection` and `pipe_stream_inbound_error_fails_stream_not_connection`, each asserting the variant + `is_transient_send_failure()` + `!ConnectionClosed`; verified fail-on-`main`, pass-with-fix.
- *(transport)* `completion_tx` is dropped (not signaled) on the cwnd abort. → Added a load-bearing comment (broadcast_queue treats the `RecvError` as completion; do not "fix" with a blocking send).

### Round 2 — focused skeptical re-review of the delta
Verified the two new conversions are semantically correct (`destination_addr` is the downstream peer; the stall/error is unambiguously upstream-feed, not downstream-connection death), no receiver-side half-open leak introduced, the `OutboundStreamTimeout`→`OutboundStreamFailed` rename is complete and compiles exhaustively, and both new tests provably reach their intended arms (not false-positives). Clean approve. My own audit confirms all four stream-scoped sites now return `OutboundStreamFailed`; the only remaining `ConnectionClosed` in `outbound_stream.rs` are negative test assertions.

**Verdict:** Approved. Scope is correct as "Addresses #4345" — this is the connection-survival half; the transport flightsize-by-stream-id root cause stays open. Merging on green CI.

[AI-assisted - Claude]

PR #4378, opened 2026-06-10
fix(release-agent): report gateway service health in /version so a failed restart isn't a false success

sanity:
## Problem

`gateway-update.yml`'s "Verify version after update" step (the final gate of every gateway rollout) polled the release-agent's `GET /version` and treated `installed == VERSION` as success. But the agent's `/version` reports the **on-disk binary** version (`freenet --version`), NOT whether the gateway **service** is actually running it.

During the **v0.2.71 release** the vega gateway's binary was swapped to 0.2.71 but the **service failed to restart**: the old process hung on shutdown, got SIGKILL'd, and systemd left the unit `failed`/down. `/version` still reported 0.2.71 (on-disk), so gateway-update reported **SUCCESS** while vega's gateway was **DOWN for ~5 minutes** until a human noticed and restarted it manually.

## Solution

Surface the **service's** health alongside the binary version, and make the workflow require it.

**release-agent (`crates/release-agent/`)** — additive, backward-compatible:
- `GET /version` now also runs `systemctl is-active <managed_service>` and returns two new fields:
  - `service_active: bool` — `true` only when systemd reports the unit `active`. `inactive`/`failed`/`activating`/unknown-unit/missing-systemctl/timeout all map to `false` — never an error, never a panic.
  - `managed_service: String` — the unit name that was checked (logged by the workflow).
- The service name is configurable via a new `managed_service` config field (default `freenet-gateway`; vega's secondary instance uses `freenet-gateway-hector`).
- `systemctl is-active` is a **read-only** query and needs **no sudo**.
- The existing `version` / `binary_path` fields are **unchanged** — additive only, so older gateway-update runs and any other `/version` consumer keep working.

**`gateway-update.yml` "Verify version after update":**
- Success now requires `installed == VERSION` **AND** `service_active == true`.
- **Backward compatibility (critical):** deployed agents won't report `service_active` until they are themselves updated. When the field is **ABSENT**, the workflow does **NOT** hard-fail — it logs a `::warning::` that the agent predates the check and falls back to the old binary-only verification. The service-running requirement is **only ENFORCED when the field is present** (detected with jq `has("service_active")`, so a present-but-`false` value is distinguished from an absent field).

The running-process-version refinement (catching a service that is up but still on the OLD binary) is **deliberately deferred**: reading `/proc/<MainPID>/exe` after a binary swap is fragile (`(deleted)` inode, re-spawning the binary, cross-process `/proc` access). The `service_active` floor already catches the vega down-service case, which is the incident this fixes. Noted as a follow-up.

## Testing

- **version.rs unit tests** for `service_active_with` (systemctl path injected via a stub script — no real systemd/sudo): `active → true`; `failed` (the vega case) / `inactive` / `activating` / missing-systemctl → `false`.
- **config.rs:** default `managed_service` is `freenet-gateway`; a custom value (`freenet-gateway-hector`) parses.
- **update_handler.rs integration test:** `/version` now returns `managed_service` and a boolean `service_active`.
- **Workflow logic** verified by hand against six representative responses (new agent active/failed, old agent without the field, stale binary, unreachable): the vega "binary new + service failed" case now yields a retry that times out to a hard error instead of a false success, while old agents fall back to binary-only success.
- `cargo fmt` + `cargo clippy -p freenet-release-agent --all-targets -- -D warnings` clean; all release-agent tests pass (31 unit + 27 integration).

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,15 @@ jobs:
           check "build: replace local release build"           "non-release"
           check ""                                             "non-release"
 
+      # Regression gate for the gateway-update verify decision (the vega
+      # v0.2.71 incident: binary swapped, service down, workflow falsely
+      # reported success). The decision logic in
+      # scripts/release-agent/verify-version-decision.sh is sourced both by
+      # gateway-update.yml and by this test, so they cannot drift. The
+      # load-bearing case is "service_active:false must NOT report success".
+      - name: Self-test gateway-update verify decision
+        run: bash scripts/release-agent/verify-version-decision_test.sh
+
       - name: Check for old-style mod.rs files
         run: |
           # New modules must use foo.rs + foo/ style, not foo/mod.rs
--- a/.github/workflows/gateway-update.yml
+++ b/.github/workflows/gateway-update.yml
@@ -125,6 +125,11 @@ jobs:
       matrix:
         gateway: ${{ fromJson(needs.plan.outputs.matrix) }}
     steps:
+      # Needed for the verify step, which sources
+      # scripts/release-agent/verify-version-decision.sh so the workflow and
+      # its regression test share one copy of the decision logic.
+      - uses: actions/checkout@v6
+
       - name: Resolve secret
         id: secret
         env:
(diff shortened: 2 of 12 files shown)


sanity:
## Review — external (gemini) + 4 Claude perspectives (Full tier: deploy-critical release infra)

**External model:** gemini (non-Claude pass); codex was usage-limited. gemini: PASS (no panics, backward-compatible, additive `/version`).

**Claude perspectives:** code-first → approve; big-picture → approve (well-scoped, additive, backward-compat rollout correct); skeptical and testing → **changes requested**, with real findings:

### Must fix
- **[MAJOR] Uncached `systemctl` fork per `/version` request** on the *unauthenticated* `/version` endpoint — reintroduces the self-DoS vector the existing `VersionCache` was built to eliminate (its rustdoc states caching exists so `/version` doesn't fork per request). Fix: short-TTL cache the `service_active` result.
- **[CRITICAL — test] The `gateway-update.yml` jq decision gate has zero automated coverage** — it is the exact incident path (present-`false` → must NOT report success; absent → fallback; present-`true`+match → success), verified only by manual simulation. There's an in-repo precedent (`scripts/release-agent/*_test.sh`) for table-testing shell logic cheaply; add it and wire it into CI.

### Address (fix or document)
- **[MAJOR] Swap-before-restart race** — `/update` returns after a ~1s probe without awaiting the restart, and the version cache flips to "new" the instant the binary is swapped, so `service_active==true` can be the *old* process. The check catches service-DOWN (the vega case) but not service-up-on-old-binary. Strengthen (e.g. require the service's `ActiveEnterTimestamp` to be at/after the update) or document the update-script ordering it relies on.

### Nits
- Misleading `gateway-update.yml:~232` comment (the `|| echo "false"` guards malformed JSON, not the unreachable-gateway case, which the version guard already filters).
- Stale `docs/RELEASING.md:181-183` converge-timeout recovery guidance (a timeout can now mean "service failed to restart").
- `/version` integration test asserts only the boolean *type*, not a concrete `service_active` value.

### Verified clean
No injection (`managed_service` is exec'd as an argv arg, not shell-interpolated); no panics/hangs (5s timeout, every error path → `false`); `/version` is strictly additive (`version`/`binary_path` unchanged); config is backward-compatible (`#[serde(default)]`); the jq backward-compat branches (absent→fallback, present-false→enforce, present-true→success) are logically correct.

**Verdict:** Sound design and clean backward-compat; **changes requested** — addressing the uncached-fork DoS, the missing jq-gate test, the swap-window race, and the nits before merge. Round 2 in progress.

[AI-assisted - Claude]

sanity:
## Round-2 review addressed (commit c28cddff)

### MUST FIX

**1. [MAJOR] systemctl forked on every unauthenticated `/version` hit** — Fixed. Added `ServiceHealthCache` (version.rs) mirroring `VersionCache`, with a **2s TTL**. The workflow polls every 5s, so the cache caps systemctl to at most one fork per poll while staying far fresher than the poll cadence. Keyed on the service name (a different unit re-queries). Two new unit tests: `service_health_cache_does_not_refork_within_ttl` (the self-DoS regression) and `service_health_cache_reforks_for_different_service`.

**2. [CRITICAL test] The jq decision gate had zero coverage** — Fixed. Extracted the decision logic out of the inline YAML into `scripts/release-agent/verify-version-decision.sh`, **sourced by BOTH gateway-update.yml and the test** so they cannot drift (mirrors the `announce-to-river_test.sh` eval-the-real-code precedent). Added `verify-version-decision_test.sh` table-testing 9 cases; the load-bearing one is present-`service_active:false` → `wait` (NOT success). **Wired into CI** via the Fmt job (`Self-test gateway-update verify decision`) so it runs on every PR. The rollout job now `actions/checkout`s so it can source the script.

### ADDRESS

**3. [MAJOR] Swap-before-restart race** — Investigated `deploy-local-gateway.sh` (what `/update` spawns via `gateway-auto-update.sh`). It does **stop → swap binary → start** (lines 425/446-448/462): the service is DOWN while the binary is replaced, so the binary only reads as the new version once the service has been restarted onto it. **There is no window** where the OLD process is `active` while the binary already reads as new, so `installed==VERSION && service_active==true` is sound. Took the **DOCUMENT** path: the load-bearing ordering invariant is now spelled out in the `ServiceHealthCache` rustdoc, `verify-version-decision.sh`, and the workflow, each noting what to add (unit `ActiveEnterTimestamp` vs. the update's `issued_at`) if the script ever switches to swap-then-restart.

### NITS

**4.** Clarified the workflow/script comment: `has()` only matters once the binary version matches; the unreachable-gateway case is handled by the `installed != target` branch; `|| echo "false"` only guards jq erroring on malformed JSON.

**5.** `docs/RELEASING.md` — the converge-timeout section now explains a timeout can mean "binary swapped but service failed to restart" → check `systemctl status freenet-gateway` (or `freenet-gateway-hector` on vega), with a direct `systemctl restart` recovery step.

**6.** The `/version` integration test now injects a stub systemctl path (`AppState.systemctl_path`, mirroring the inner query's injection). Two HTTP-layer assertions pin `service_active: true` and `false` (`version_endpoint_reports_service_active_true` / `..._false_when_failed`) instead of just asserting the field is a boolean.

### Verification
- `cargo fmt` clean
- `cargo clippy -p freenet-release-agent --all-targets -- -D warnings` clean
- `cargo test -p freenet-release-agent` — 33 unit + 29 integration pass
- `bash scripts/release-agent/verify-version-decision_test.sh` — 9/9 pass (incl. the vega false-case)
- gateway-update.yml + ci.yml YAML valid; shellcheck clean on both new scripts

Not merged.

[AI-assisted - Claude]

sanity:
## Round-3 addressed (commit 18c038d4)

**Fix (testing): TTL-expiry → refork path now pinned.** Added an injectable TTL on `ServiceHealthCache` (a `ttl` field defaulting to `SERVICE_HEALTH_TTL` + a `with_ttl(..)` constructor; production stays 2s) so a test can force instant expiry without a real `sleep`. New test `service_health_cache_reforks_after_ttl_expiry`: an entry past the TTL re-queries AND returns the fresh state (active → failed, the vega case). With the existing within-TTL no-refork test, both sides of the boundary are covered.

**Note (no code change): thundering-herd.** Added a one-line `# Concurrency` rustdoc on `ServiceHealthCache` recording that concurrent cache-misses each query (no single-flight), matching `VersionCache`'s accepted, TTL-bounded tradeoff — so it isn't re-flagged.

**Bonus — fixed a real flake I hit while adding the test.** Driving the cache through a real subprocess fork per cached call made the cache tests (and the pre-existing `service_active_*` interpretation tests) flaky under plain `cargo test`: each `#[tokio::test]` is its own current-thread runtime, and many spawning+reaping children concurrently in one process races tokio's Unix child reaper (verified — the raw spawn returned `active` while a sibling spawn in the same test failed). Fix: cache-behaviour tests now drive a new `is_active_inner(service, query_closure)` seam with an in-process counting closure (deterministic, no fork); the real fork path stays covered by the `query_service_active` tests + one `is_active_with` wiring smoke test. CI uses `cargo nextest` (process-per-test isolation), which avoids the race entirely.

### Verification
- `cargo fmt` clean
- `cargo clippy -p freenet-release-agent --all-targets -- -D warnings` clean
- `cargo nextest run -p freenet-release-agent` (the CI runner) — **64/64 passed, 0 failed across 20+ consecutive runs**
- `cargo test --test-threads=1` — 30/30 green

Not merged.

[AI-assisted - Claude]

sanity:
## Re-review (rounds 2–3) — all findings resolved

Following the round-1 review (gemini external + 4 Claude perspectives), every finding has been addressed and independently re-verified (gemini + skeptical + testing lenses on the deltas):

- **[MAJOR] Uncached `systemctl` fork per `/version` request** → fixed with a `ServiceHealthCache` (2s TTL, service-keyed, lock released before the fork). Both sides of the TTL boundary are now pinned (no-refork-within-TTL + refork-after-expiry, the latter via an injectable-TTL test seam — no `sleep`).
- **[CRITICAL — test] gateway-update jq decision gate had no automated coverage** → the decision logic was extracted into `scripts/release-agent/verify-version-decision.sh`, **sourced by both the workflow and a new 9-case table-test** (no drift), **wired into CI** (Fmt job, every PR). The load-bearing present-`false`→wait (the vega incident path) is explicitly covered.
- **[MAJOR] swap-before-restart race** → investigated the actual updater: `deploy-local-gateway.sh` does stop→swap→start, so there is no window where the old process is active while the binary reads new. Documented the ordering as a load-bearing invariant (rustdoc + script + workflow) with a note on what to add if it ever changes.
- **Nits** → misleading `:232` comment clarified; `RELEASING.md` converge-timeout recovery updated; `/version` integration test now asserts concrete `service_active` true *and* false over HTTP via an injected systemctl stub.
- **[LOW, gemini] thundering-herd on concurrent cache-miss** → justified dismissal: identical to the existing `VersionCache` accepted tradeoff, bounded by the 2s TTL; documented in a `# Concurrency` rustdoc line so it isn't re-flagged.

Round 3 also surfaced and properly root-caused a **real pre-existing test flake** (per-call subprocess forking racing tokio's Unix child-reaper across concurrent current-thread runtimes) and fixed it deterministically with an in-process test seam — CI's `nextest` (process-per-test) was never affected; `cargo test` is now robust too.

Verified the production `/version` path still forks the real `systemctl` (the seam is test-only). fmt + clippy clean; nextest 64/64 across 20+ runs; the shell decision-test 9/9. CI green on the current HEAD.

**Verdict:** Approved. Merging.

[AI-assisted - Claude]

PR #4379, opened 2026-06-10
feat(transport): phase 1.6 shadow demand/queue/iface/class telemetry

sanity:
## Problem

Issue #4074 (the outer-loop rate-controller design-of-record) needs the
*demand* side of the floor analysis before a controller can be designed.
Phases 1 and 1.5 added RTT-based contention signals; Phase 1.6 — the
explicit next step in the issue's "Implementation checklist — Phase 1.6
telemetry" — adds observation-only telemetry for how much each node is
trying to send, how that compares to the rate cap `R`, how deep the
broadcast backlog gets, who is competing for the uplink, and how outbound
traffic splits between slowable bulk and unslowable must-flow.

This is **observation only**: no change to transport behaviour. The data
feeds the floor lower/upper-bound analysis in the #4074 body
("What to look for in the telemetry").

## Solution

Three new ~1 Hz shadow signals, each emitted from its own background task,
peer_id-tagged, same OTLP shape as the existing `shadow_rtt_aggregate`:

1. **`shadow_rate_demand`** — achieved outbound throughput (a faithful
   offered-demand proxy: the transport never drops outbound, it sleeps on
   the token bucket / cwnd, so backlog surfaces as queue depth), active
   connection count, broadcast-queue depth, and the effective aggregate
   rate `R` (`GlobalBandwidthManager`, present only in global-pool mode).
2. **`shadow_iface_tx`** — aggregate OS interface tx bytes from Linux
   `/proc/net/dev` and the derived `op = total_tx − freenet_own_tx`
   (`freenet_own_tx` = `TRANSPORT_METRICS.cumulative_bytes_sent`). Behind
   the bare `telemetry.iface-tx-enabled` flag (default off), best-effort:
   if `/proc/net/dev` is unavailable the tick is omitted, never blocking.
3. **`shadow_outbound_class`** — the must-flow / short / bulk byte split.
   `Bulk` = `StreamFragment` (the only class the token bucket meters, i.e.
   the only slowable traffic); `Short` = `ShortMessage` (opaque serialized
   `NetMessage` — small contract op or control plane, unsplittable at the
   transport layer); must-flow = `Ping` / `Pong` / `NoOp` / `AckConnection`.

### Design notes

- **Nothing in the production data path reads the shadow state.** The hot
  send path only *writes* the cheap class counters — one relaxed atomic add
  per packet at `packet_sending`, plus the keep-alive `Ping` and `send_pong`
  bypass sites (no allocation, no lock, no blocking). All reads happen in
  the separate aggregator tasks. The "NEVER read … from the production data
  path" rule in `.claude/rules/transport.md` is preserved and extended.
- **`R` mapping.** `R` maps to `GlobalBandwidthManager`, which is `Option`
  (only set in global-pool mode, i.e. when `total_bandwidth_limit` is
  configured). The demand event also emits `active_connections` (from the
  Phase 1 RTT registry, always available) so per-connection demand is
  computable under the default per-connection FixedRate mode, where the
  effective per-connection `R` is the known FixedRate default.
- **Broadcast-queue depth** is published by `broadcast_queue.rs` under its
  own queue lock into a relaxed gauge, so the shadow reader never contends
  on the queue mutex.
- **Telemetry budget.** These add up to 3 always-on + 1 gated 1 Hz events
  to an OTLP pipe rate-limited at 10 events/sec (no priority ordering). On a
  busy gateway some events (shadow or operational) may be dropped within a
  1 s window — an accepted, statistical tradeoff for shadow telemetry, and
  the cadence is trivially tunable if it proves too aggressive.

### Extractor (separate repo)

The matching `EVENT_TYPES` change in the nova admin repo
(`extract-shadow-telemetry.py`) adds the three new event names so they
survive the ~3-day raw-log rotation into the durable shadow archive. It is
deployed to nova separately and harmlessly matches nothing until nodes pick
up a release carrying these events.

## Testing

- `classify` maps every `SymmetricMessagePayload` variant to its bucket
  (exhaustive match, no wildcard — a new variant is a compile error).
- per-interval delta math, including saturation on counter reset.
- `/proc/net/dev` parsing: sums tx excluding `lo`, tolerates short / garbage
  lines, returns `None` on header-only / empty input, excludes only the
  exact `lo` device.
- the broadcast-depth gauge round-trips.
- all three aggregator / monitor tasks survive multiple ticks without the
  `BackgroundTaskMonitor` JoinHandle exiting (mirrors the Phase 1 / 1.5
  survival pins).

`cargo fmt`, `cargo clippy -- -D warnings`, and the transport + config
suites are green locally.

Part of #4074.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -137,15 +137,67 @@ from CI runners (which would perturb timing-sensitive multi-node
 tests — see PR #4292 root-cause). The shadow aggregator is
 always-on; only reference-ping is gated.
 
+### Shadow demand / queue / OS-counter / class telemetry (Phase 1.6)
+
+`transport/shadow_demand.rs` adds the *demand* side of the floor
+analysis. Three process-wide observation hooks are written from the
+hot send path with a single relaxed atomic add each, and read only
+by the 1Hz aggregator tasks:
+
+- **Outbound class counters** (`record_outbound`): cumulative bytes
+  split into must-flow / short / bulk. `classify()` keys off the
+  `SymmetricMessagePayload` variant — `StreamFragment` is `Bulk`
+  (the only class the token bucket meters, i.e. the only slowable
+  traffic), `ShortMessage` is `Short` (opaque serialized `NetMessage`
+  — small contract op or control plane, unsplittable at the transport
+  layer), and `Ping`/`Pong`/`NoOp`/`AckConnection` are `MustFlow`.
+  Tagged at `packet_sending` (covers short/noop/ack/stream) plus the
+  two bypass sites `send_pong` and the keep-alive `Ping`.
+- **Broadcast-queue depth gauge** (`record_broadcast_queue_depth`):
+  updated by `node/network_bridge/broadcast_queue.rs` under its own
+  queue lock, so the shadow reader never contends on the queue mutex.
+- **Global-bandwidth handle** (`register_global_bandwidth`): a `Weak`
+  to the node's `GlobalBandwidthManager`, registered at construction
+  in `p2p_protoc.rs`, so the demand aggregator can read the effective
+  aggregate rate `R` — present only in global-pool mode
+  (`total_bandwidth_limit` set); null under the default per-connection
+  FixedRate mode.
+
+Two always-on 1Hz tasks (spawned from `p2p_impl.rs`, registered with
+`BackgroundTaskMonitor` as `shadow_demand_aggregator` /
+`shadow_outbound_class_aggregator`) emit `shadow_rate_demand`
+(achieved throughput vs `R`, active-connection count, broadcast-queue
+depth) and `shadow_outbound_class` (the must-flow / short / bulk byte
+split). Same OTLP shape and peer-id tagging as `shadow_rtt_aggregate`.
+
+`transport/shadow_iface_tx.rs` adds the OS-interface counter: a 1Hz
+read of Linux `/proc/net/dev` summing non-loopback `tx_bytes`, emitting
+`shadow_iface_tx` with the derived `op = total_tx − freenet_own_tx`
+(`freenet_own_tx` = `TRANSPORT_METRICS.cumulative_bytes_sent`). It is
+**opt-in and best-effort**: gated by `telemetry.iface-tx-enabled`
+(default `false`, same gating as reference-ping); if `/proc/net/dev`
+is unavailable the tick is silently omitted, never blocking.
+
+The class tagging is the only Phase 1.6 hook that touches the hot send
+path; it is a single relaxed atomic add per packet (no allocation, no
+lock, no blocking). Everything else is read from the separate
+aggregator tasks.
+
 ```
 NEVER read SHADOW_RTT_REGISTRY, cross_connection_median_inflation,
-or the reference_ping stats from the production data path (rate
-limiter, retry, congestion control). They exist only for the
-staged rollout in #4074:
+the reference_ping stats, OR the Phase 1.6 shadow_demand counters /
+broadcast-depth gauge / registered bandwidth handle from the
+production data path (rate limiter, retry, congestion control). The
+hot path only WRITES the cheap class counters; all reads happen in
+the aggregator tasks. They exist only for the staged rollout in
+#4074:
   Phase 1   → observation only — per-peer overlay RTT (current)
   Phase 1.5 → observation only — adds reference-path RTT + peer_id
               tagging so signals can be disaggregated per node and
               the overlay-vs-uplink confound can be tested
+  Phase 1.6 → observation only — adds outbound demand vs R, broadcast
+              queue depth, OS interface tx (op = total − own), and the
+              must-flow-vs-bulk outbound split
   Phase 2   → shadow controller, still no behaviour change
   Phase 3   → opt-in flag
   Phase 4   → default switch only after Phase 3 shows improvement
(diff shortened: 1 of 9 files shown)


sanity (commented):
## Comprehensive PR Review: #4379

### Summary
- **PR Title:** feat(transport): phase 1.6 shadow demand/queue/iface/class telemetry
- **Type:** feat (observation-only telemetry)
- **CI Status:** green on the original HEAD except the Conventional-Commits title check (fixed by lowercasing the title subject); re-verifying on the post-review HEAD.
- **Linked Issues:** #4074 (design-of-record; this is its Phase 1.6 step)
- **Review tier:** Full (transport hot send path + concurrency/async = high-risk surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture (Claude perspectives) + **Gemini** as the external non-Claude pass. **Codex was quota-limited** ("try again at 9:21 PM"), so per the multi-model-review fallback I ran the other available external model (Gemini); the four diverse Claude lenses also provide perspective diversity.

---

### Verdict
**State:** Ready to Merge (after CI re-confirms green on the post-review HEAD)
**HEAD SHA reviewed:** 51620f5a (full Full-tier review) → **c8d19507** (review fixes).

No Must-Fix or Should-Fix correctness/concurrency findings survived verification. The observation-only invariant is verified by grep and inspection: the hot path only **writes** cheap relaxed atomics; every read lives in the separate aggregator tasks. All review findings were addressed (fixes pushed) or are explicitly justified deferrals with a tracking follow-up.

**Re-review note (c8d19507):** the fix commit is docs + tests + comments only, plus one production change — extracting `iface_op(total, own) = total.saturating_sub(own)` from the inline loop expression, which is provably behavior-identical. `outbound_counters_snapshot` is `#[cfg(test)]`. No new production logic, no Must-Fix was addressed, so a **light diff-of-the-diff re-check** is sufficient rather than a full re-spawn. `cargo fmt`/`clippy -D warnings` clean and the full 637-test transport suite is green on c8d19507.

---

### Findings and dispositions

**Gemini "Critical: packet-ID reuse" (peer_connection.rs multi-packet `send!`) — DROPPED (false positive for this PR).**
Verified against the diff: this PR does **not** touch `packet_id`. The `report_sent_packet_with_token(packet_id, …)` with the shared id is pre-existing code; my change only adds `sent_on_wire +=` and two `record_outbound` calls. The thorough skeptical review of the same branch did not corroborate it as a bug. Flagging for maintainer awareness as a possible *separate, pre-existing* investigation (transport-reliability owner), not something to fix in an observation-only PR.

**Telemetry budget (skeptical, Medium) — DEFERRED with tracking issue + docs.**
Shadow telemetry is now 3 always-on 1Hz events (RTT + demand + class), up to 5 on a gateway with reference-ping + iface-tx, against a flat 10-events/sec OTLP cap with no priority. The proper fix (a priority/sub-budget field on `TelemetryEvent` + the rate-limit gate in `tracing/telemetry.rs`) is a change to the *shared* telemetry subsystem — out of scope for this focused, observation-only change, and the reviewer explicitly accepted deferral. Documented in the `shadow_demand.rs` module rustdoc and filed as a follow-up to land before Phase 2 adds more always-on streams.

**Doc accuracy (big-picture / skeptical / Gemini) — FIXED.**
- `transport.md` "covers short/noop/ack/stream" was inaccurate: `packet_sending` never carries `AckConnection` (handshake ACKs go via the bypassing `ack_ok`). Corrected to `ShortMessage`/`NoOp`/`StreamFragment` and documented that the split is a classified *subset* of the node total (excludes retransmits + handshake/standalone ACKs, which are still in `cumulative_bytes_sent`).
- `shadow_demand.rs`: added "what the class split covers and does not" plus a note for the analysis consumer that `must_flow` here is transport-level liveness only (subscription/relay obligations ride inside `short`/`bulk`).
- `shadow_iface_tx.rs`: documented that summing all non-loopback interfaces double-counts on container hosts (docker/veth/bridge); gateways are bare-metal.

**Test coverage (testing) — FIXED.**
- Added a `packet_sending` → counter integration test (reusing `FailableTestSocket`): `ShortMessage`→`short`, `StreamFragment`→`bulk` (lower-bound deltas, robust under shared-process runs), and a failing send records nothing.
- Added `/proc/net/dev` colon-glued-counter and huge-counter (no-overflow) parse tests.
- Extracted `iface_op` and added op/zero-interval saturation tests; added a quiet-interval `outbound_delta` test.

**Multi-packet accounting nits (code-first / skeptical / Gemini) — DOCUMENTED.**
Expanded the comment at the multi-packet site: trailing NoOp confirm-receipt packets are counted under the primary class, and a mid-burst `send_to` failure skips the `record_outbound` (so the split can undercount vs the socket-layer total). Both tiny, only on the rarely-hit multi-packet path.

**Confirmed safe by reviewers (no action):** no double-count of Ping/Pong/NoOp; `payload.into()` is a zero-cost behavior-preserving move; `record_outbound` only on success paths; broadcast gauge under the lock adds ~zero hold time, no drift, no deadlock, no #4231/#3309 interaction; all three tasks are panic-free infinite loops that can't trip `wait_for_any_exit`; `/proc` field index 8 is correct; `OnceLock<Weak<GBM>>` is read-only and lifetime-safe (returns null fields if the GBM is gone); config flag override is one-directional, gated off in CI/dev.

---

### Out-of-PR follow-ups (tracked, not forgotten)
1. **Admin repo** (`~/code/admin/nova/freenet/`): add the three event names to `EVENT_TYPES` in `extract-shadow-telemetry.py` and document them in the `freenet-telemetry-monitor` skill — both are #4074 checklist items (d)/(e) that live outside this repo. Must be deployed to nova before/with the release carrying these events, or the data is lost to ~3-day rotation.
2. **Telemetry-event priority / sub-budget** before Phase 2 (skeptical Finding 1).

[AI-assisted - Claude]

PR #4383, opened 2026-06-12
fix(transport): drive gateway admission rate limiter by VirtualTime in sim

sanity:
## Problem

The nightly **Simulation Tests (Nightly)** job intermittently fails on
`test_thundering_herd_connect_storm` (reported 2026-06-12). The same commit
(`7e09c1eb`) **passed** on 2026-06-11 and **failed** on 2026-06-12, so it is a
flaky test, not a regression.

Root cause: the gateway connection admission rate limiter
(`GatewayConnectionRateLimiter`) ramps its accept rate over *seconds since
gateway start* (5/s for 30s, 20/s to 2min, then unlimited), reading elapsed
time from the connection handler's `RealTime` source. In a SimNetwork the
handler is built with `RealTime` (`config_listener`), but the two SimNetwork
execution modes progress on different clocks:

- **Virtual-time-driven** (the thundering-herd test advances `VirtualTime`
  explicitly while spending only ~10ms real per 100ms virtual): the real-clock
  ramp never progresses, so admission stays pinned at the throttled 5/s and
  which peers get admitted depends on real-scheduler ordering. In ~25% of runs
  the same handful of peers never got admitted, stalling at 75% vs the 80%
  threshold.
- **Real-time-driven** (`VirtualTime` is registered but not advanced — note
  *"VirtualTime is now always enabled"* for SimNetworks): the real clock is the
  one that moves.

Evidence: reproduced locally ~25% (3/12) on an idle machine; even a forced 300s
*virtual* budget on the connectivity check still failed 25%, confirming an
admission stall rather than a check timeout. The failing peer set is
deterministic.

## Solution

The limiter measures elapsed time as **`max(real_elapsed, virtual_elapsed)`**.
This advances at least as fast as either clock alone, so it is correct in both
modes and is **never more restrictive than the previous real-clock behavior**
(it can only let the ramp complete sooner). In production no `VirtualTime` is
registered, so this is exactly real-clock elapsed and behavior is unchanged.

`config_listener` populates the simulation clock from the existing
address→network→`VirtualTime` registry via a new `get_socket_virtual_time(addr)`
helper (`Some` only in simulation, `None` in production).

### Why not the virtual clock alone

A first attempt used the virtual clock alone. That **regressed real-time-driven
sim tests** (`test_concurrent_updates_convergence`,
`test_auto_fetch_from_update_sender`, `test_relay_route_events_multihop`,
`test_get_routing_coverage_low_htl`): with a frozen virtual clock the 1-second
window never reset and the gateway blocked every peer after the first burst, so
the network never formed (CI Simulation job caught this). `max()` fixes both
modes. The sibling `RealTime`-in-sim sites (per-IP asym throttle,
`RecentlyClosed` debounce) share this mismatch but are not currently breaking
tests; tracked in #4384.

## Testing

- New unit test `test_gateway_rate_limiter_admission_clock_is_max_of_real_and_virtual`
  drives the ramp by **each clock independently** (two controllable clocks); a
  revert to either single clock fails one half.
- `test_thundering_herd_connect_storm`: **20/20** locally (was ~25% fail) and
  **15/15 under CPU load**.
- The four previously-regressed tests pass.
- Full CI-matching sweep (`simulation_integration` + `simulation_smoke` +
  `streaming_e2e` + `state_verification`, `--profile ci`): **142/142**.
- `build` / `clippy` / `fmt` clean.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/transport/connection_handler.rs
+++ b/crates/core/src/transport/connection_handler.rs
@@ -41,7 +41,8 @@ use super::{
     symmetric_message::{SymmetricMessage, SymmetricMessagePayload},
     token_bucket::TokenBucket,
 };
-use crate::simulation::{RealTime, TimeSource};
+use crate::simulation::{RealTime, TimeSource, VirtualTime};
+use crate::transport::in_memory_socket::get_socket_virtual_time;
 
 // Constants for interval increase
 const INITIAL_INTERVAL: Duration = Duration::from_millis(50);
@@ -296,7 +297,16 @@ impl<S: Socket> OutboundConnectionHandler<S> {
             global_bandwidth,
             ledbat_min_ssthresh,
             expected_non_gateway: expected_non_gateway.clone(),
-            gw_rate_limiter: GatewayConnectionRateLimiter::new(time_source.clone()),
+            // In simulation the admission ramp must advance with the network's
+            // VirtualTime, not only the RealTime source this handler is built with
+            // (see `GatewayConnectionRateLimiter::effective_elapsed_nanos`).
+            // `get_socket_virtual_time` returns `Some` only when a SimNetwork has
+            // registered a VirtualTime for this address, and `None` in production
+            // (where the ramp stays on real wall-clock exactly as before).
+            gw_rate_limiter: GatewayConnectionRateLimiter::new(
+                time_source.clone(),
+                get_socket_virtual_time(&socket_addr),
+            ),
             time_source,
             congestion_config: Some(congestion_config.unwrap_or_default()),
         };
@@ -495,7 +505,9 @@ impl<S: Socket> OutboundConnectionHandler<S, crate::simulation::VirtualTime> {
             global_bandwidth,
             ledbat_min_ssthresh,
             expected_non_gateway: expected_non_gateway.clone(),
-            gw_rate_limiter: GatewayConnectionRateLimiter::new(time_source.clone()),
+            // This listener's `time_source` is already VirtualTime, so the admission
+            // clock needs no override.
+            gw_rate_limiter: GatewayConnectionRateLimiter::new(time_source.clone(), None),
             time_source,
             congestion_config,
         };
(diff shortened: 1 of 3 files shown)


sanity:
Updated the fix after the Simulation CI check caught a regression: the first version drove the gateway admission ramp by the simulation `VirtualTime` alone, which broke real-time-driven sim tests (frozen virtual clock never resets the 1s admission window → gateway blocks all peers after the first burst). New version uses `max(real_elapsed, virtual_elapsed)`, which is correct in both execution modes and provably never more restrictive than the prior real-clock behavior. Force-pushed to `6b57666`. Full `simulation_integration`/`smoke`/`streaming_e2e`/`state_verification` sweep is now 142/142 locally; thundering-herd 20/20 (+15/15 under load). Re-running the multi-model review on the new HEAD. Filed #4384 for the sibling RealTime-in-sim sites. [AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4383 (re-review on `max()` mechanism)

### Summary
- **Title:** fix(transport): drive gateway admission ramp by max(real,virtual) clock in sim
- **Type:** fix (transport / simulation determinism)
- **Review tier:** Full (transport = high-risk surface)
- **Reviewers run:** code-first, testing, skeptical (freenet:* subagents) + external **Codex**, on the current HEAD. (An earlier round ran on the pure-virtual version; that version's CI Simulation failure drove the redesign to `max()`, which this round re-reviewed.)

### Code-First
Implementation matches the description faithfully; doc comments accurately describe `max(real_elapsed, virtual_elapsed)`, monotonicity, and the production no-op. No intent/impl gap. Low-severity/doc-only notes only.

### Testing
The new `test_gateway_rate_limiter_admission_clock_is_max_of_real_and_virtual` is a genuine **two-sided pin**: the virtual-driven half fails under a pure-real revert, the real-driven half fails under a pure-virtual revert. The real-driven half **directly reproduces in milliseconds** the regression CI caught (frozen virtual clock freezing the admission window), which is strictly better than relying on the multi-second sweep. `fix:`-needs-a-test gate satisfied. No blocking gaps. (Added a both-clocks assertion this round to also pin `max` vs a hypothetical `sum`, per the skeptical reviewer.)

### Skeptical
No High/Medium/Low **correctness** bugs. Verified:
- **Monotonicity:** `max` of two monotonic terms is monotonic; traced both clock-dominance flips — `effective` never decreases, so the window `saturating_sub` never spuriously underflows.
- **Production equivalence:** with `virtual_clock = None`, `effective = real_elapsed`, and the window math algebraically cancels `real_start` to exactly the prior `now - window_start` — byte-for-byte prior behavior. All registry writers are test-only, so production override is always `None`.
- **No under-throttle:** `max()` only lets the ramp complete sooner, never later; the one other throttle-dependent test uses a single clock and is unaffected.
- Edge cases (both clocks at 0, saturating_sub) safe.

### External (Codex)
"No discrete, actionable correctness issues. The rate limiter behavior remains real-time based in production and adds the simulation virtual-time override in the intended listener path."

### Documentation
Doc comments are accurate and unusually thorough. Stale thundering-herd comment (`simulation_integration.rs:6343`) updated to reflect the new clock. Per-IP asym throttle deliberately left on RealTime with an inline note pointing at the follow-up.

### Recommendations
- **Must Fix:** none.
- **Should Fix:** none.
- **Consider:** the sibling RealTime-in-sim sites (per-IP asym throttle, `RecentlyClosed` debounce) share this mismatch but break no current test — filed as **#4384** with the recommended long-term direction (thread the sim `VirtualTime` as the handler's `time_source`).

### Verdict
**State:** Ready to Merge (gated on CI green on the current HEAD).
**HEAD SHA reviewed:** `802515de` (test-only refinement over the reviewed `6b576665`; the `max()` logic is identical).

Validation: `test_thundering_herd_connect_storm` 20/20 (+15/15 under CPU load), the 4 previously-regressed tests pass, full `simulation_integration`/`smoke`/`streaming_e2e`/`state_verification` sweep 142/142, two unit tests green.

[AI-assisted - Claude]

Issue #4391, opened 2026-06-13
Executor head-of-line blocking: inline network related-fetch on contract_handling loop stalls all contract ops (incl. cached GETs)


sanity:
## Summary

The single-threaded `contract_handling` event loop runs network *related-contract*
sub-GETs **inline**, blocking every other queued contract op — including
`GetQuery` reads that would otherwise be served instantly from the local store —
for up to **10s** (validation path) or **120s** (update-`requires` path) per
stuck op.

This is the op/executor-layer half of the "first send hangs the recipient node,
all contract GETs fail" symptom (freenet/mail#288). It is **distinct from** the
transport-layer mechanism in #4345 (cwnd-wait abort / stream-assembly wedge):
even with the transport fixes (#4353, #4367, #4374) fully in place, a single
related-contract fetch that can't be retrieved quickly still head-of-line-blocks
the loop. This is the piece that explains why *locally-cached* GETs stall, which
the transport explanation alone does not.

## Mechanism (file:line, current `main`)

1. `contract_handling` is a single consumer task; it processes one event to
   completion before the next:
   - spawned once: `node/p2p_impl.rs:453-470`
   - serial loop: `contract.rs:667-727`, awaits `handle_contract_event(...).await`
     at `contract.rs:709-711` before popping the next event
   - the pool's own doc comment confirms it: "the sequential event loop means at
     most one contract is ever in flight at a time" (`contract/executor/runtime.rs:382-383`,
     and `:122-125`)

2. A `PutQuery`/`UpdateQuery` whose `validate_state`/`update_state` returns
   `RequestRelated`/`requires(missing)` and whose related contract is **not local**
   escalates to a **network** sub-op GET that is `.await`ed inline:
   - `fetch_related_via_network` (`runtime.rs:4266-4298`): `start_sub_op_get` then
     `rx.await`, bounded by `RELATED_FETCH_TIMEOUT = 10s` (`runtime.rs:12`)
   - `local_state_or_from_network` (`runtime.rs:4200-4249`): `start_sub_op_get`
     then `rx.await` under `SUB_OP_FETCH_TIMEOUT = 120s` (`runtime.rs:4227-4228`)
   - these are `&mut self` methods reached from `handle_contract_event`'s
     Put/Update arms, so the await happens on the `contract_handling` task itself.

3. A local-cache-hit `GetQuery` is serviced by the *same* loop and does only
   local reads (`handle_contract_event` GET arm → `fetch_contract` →
[...]

PR #4394, opened 2026-06-13
fix(contract): off-load network related-fetch from contract_handling loop (#4391)

### LINKED ISSUE #2: NAT traversal

### LINKED ISSUE #3: Decentralized email proof-of-concept

### LINKED ISSUE #4: Peer resource usage balancing

### LINKED ISSUE #5: Intelligent routing

sanity:
## Problem

All contract operations are processed by ONE serial loop, `contract_handling`
(`crates/core/src/contract.rs`), which awaits `handle_contract_event(...)` to
completion before processing the next event.

When a PUT or UPDATE's validation needs a "related contract" that isn't held
locally, the runtime did a NETWORK fetch and **awaited it inline on this loop**:

- `fetch_related_for_validation` / the `update_state requires()` branch in
  `bridged_upsert_contract_state` → `start_sub_op_get` then `rx.await`, bounded
  by `RELATED_FETCH_TIMEOUT = 10s`.
- (the older `local_state_or_from_network` path used `SUB_OP_FETCH_TIMEOUT = 120s`).

Because the loop is serial, **everything queued behind that PUT/UPDATE waited the
full timeout** — including local-store-hit `GetQuery` reads that only touch the
local store. That is the user-visible "all GETs stall, even cached ones, for a
while" bug.

## Approach

The sub-op GET already runs as its own background task (`start_sub_op_get`
spawns it); the only thing pinning the loop was the inline `rx.await`. This PR
moves that wait off the loop and re-enqueues a continuation:

1. **Deferrable upsert.** `bridged_upsert_contract_state_inner` takes a new
   `defer_related_fetch` flag. The related-fetch sites resolve related contracts
   **local-only first**. If everything needed is already local, behavior is
   identical to today (no network, no deferral). If a related contract requires
   the network, the upsert rolls back any partial work via the existing error
   path (init-tracker / contract-store rollback) and returns a typed
   `DeferRelatedFetch` carrying the missing ids — surfaced as
   `UpsertOutcome::DeferRelated`. Exposed via a new
   `ContractExecutor::upsert_contract_state_deferrable` (default impl never
   defers, so local-only / mock executors are unchanged).

2. **Off-loop waiter + re-enqueue.** On `DeferRelated`, the loop parks the
   client's response oneshot (new `take_waiting_response` / `StashedResponder`
   on the handler half) and spawns a `GlobalExecutor::spawn` waiter that drives
   the GET(s) under a single unified `RELATED_FETCH_TIMEOUT`, then re-enqueues a
   `DeferredResume` via a **loop-owned unbounded channel** (non-blocking
   `unbounded_send`, never `.send().await` — see channel-safety.md). The loop
   returns immediately so queued events (the cached GETs) drain.

3. **Resume on the loop.** The loop re-runs the upsert ON the loop with the
   fetched related states supplied (so it resolves locally, no re-fetch), then
   answers the original client through the stashed oneshot. **WASM validate /
   store still runs serially on the loop** — only the network *wait* moved
   off-loop.

### Invariants preserved

- **depth=1 / one round.** The resume uses the deferrable path again; if the
  contract now requests a *different* related contract not held locally, that
  is converted to `MissingRelated` rather than deferring a second time — exactly
  one deferral per op, no infinite loop.
- **Client response routing.** The original `EventId`/oneshot is carried through
  the deferral and fired on resume.
- **No new executor concurrency.** Only the network wait is off-loop; WASM stays
  serial on the loop.
- **channel-safety.** Re-enqueue is a non-blocking unbounded send; in-flight
  deferrals are capped (`MAX_INFLIGHT_DEFERRALS`) so a flood of related-needing
  PUTs falls back to the inline path rather than spawning unbounded waiters.

## Testing

New `hol_4391_tests` module drives the **real `contract_handling` loop**:

- `deferred_related_fetch_does_not_block_local_get` — the core regression: a PUT
  whose related fetch hangs must NOT block a subsequent local-store-hit GET.
  **Verified this FAILS without the fix** (temporarily forcing the inline wait
  back onto the loop makes the cached GET time out: `Elapsed(())`) and **PASSES
  with it**.
- `deferred_related_fetch_success_resumes_and_completes` — deferred PUT resumes
  and stores once the off-loop fetch succeeds.
- `deferred_related_fetch_failure_surfaces_missing_related` — a failed off-loop
  fetch surfaces an error to the client (no hang).
- `all_related_local_does_not_defer` — a PUT whose related contract is already
  local never invokes the off-loop fetch (no behavior change).
- `second_related_request_after_resume_does_not_defer_again` — the one-deferral
  cap: a contract that keeps requesting related contracts is rejected after a
  single off-loop fetch (asserts the fetch runs exactly once).

All 275 `contract::` module tests pass; the `#4077` parallel-fetch pin test was
updated to account for the new local-only resolution loop while still rejecting
a serial *network* fetch. `cargo fmt` + `cargo clippy -- -D warnings` clean.

[AI-assisted - Claude]

---

## Review round 2 update (per-contract FIFO, cap, TTL)

Addressed the multi-model review's blocking design fix and Should-Fixes on top of the original off-loading mechanism:

### MUST-FIX #1 — per-contract FIFO preserved across a deferral
While a contract's PUT/UPDATE has an in-flight off-loop fetch, the loop now
HOLDS that contract's subsequent events (`DeferralCtx.deferred_keys`,
`pop_runnable_event`) and re-enqueues them in FIFO order when the deferral
commits/fails/times out. So a same-key GET/UPDATE issued during the deferral
waits and observes the committed state instead of `MissingContract`/stale.
Other contracts keep draining — the whole point of #4391. The block's lifetime
is tied to the stashed responder (released on resume in `handle_deferred_resume`
or on TTL in `sweep_stale_deferrals`).

### MUST-FIX #2 — capped resume drain
The resume drain is now bounded (`MAX_RESUME_DRAIN_BATCH = 16`) and interleaved
with the fair queue, so back-to-back WASM resumes can't head-of-line-block
cached GETs.

### MUST-FIX #3 — TTL + lifecycle for the deferral machinery
`sweep_stale_deferrals` runs each loop iteration: any deferral whose off-loop
fetch never reported within 2× `DEFERRED_RELATED_FETCH_TIMEOUT` is answered with
the correct response variant (Put/Update, tracked per-entry) and its
per-contract block is released — so a wedged/dropped waiter can't leak the
stashed responder (client hang) or permanently wedge the contract (which would
make `at_capacity()` permanently true and silently disable the whole fix). The
held buffer is itself bounded (`MAX_HELD_PER_DEFERRAL = 100`). The fire-and-forget
waiter rationale is documented at the spawn site (short-lived, bounded,
self-terminating, sweep-covered) per code-style.md.

### SHOULD-FIX #4 — capacity fail-fast
At the in-flight cap, a new related-needing op now surfaces `MissingRelated`
immediately instead of falling back to an inline network fetch (which would
re-stall the loop for the over-cap op).

### SHOULD-FIX #5 — test coverage
New tests (all drive the real `contract_handling` loop unless noted):
- `deferred_put_holds_same_key_get_but_not_other_key` — the MUST-FIX #1
  regression: same-key GET waits + sees committed state; cross-key GET prompt.
- `deferred_put_then_same_key_update_is_ordered_after_commit` — UPDATE-after-
  deferred-PUT-same-key ordering (no MissingContract).
- `update_path_deferral_resumes_and_applies` / `..._failure_surfaces_update_error`
  — UPDATE-path deferral (the response path differs from PUT).
- `off_loop_sub_op_outcome_mapping` — deterministic unit test of the real
  `SubOpGetOutcome → MissingRelated` mapping (factored into `map_sub_op_outcome`),
  covering the production path without a flaky real-network harness.
- `client_disconnect_during_deferral_releases_block` — disconnect mid-deferral
  doesn't leak the responder or wedge the contract.
- `deferral_cap_fails_fast_with_missing_related` — the 256-cap boundary.
- `put_update_arms_route_through_maybe_defer_upsert_pin_test` — source pin so a
  refactor can't silently re-inline the blocking upsert.
- RAII `OverrideGuard` so a panicking test can't leak the process-global stub.

### Docs
- Corrected the timeout-reconciliation claim: only the 10s
  `RELATED_FETCH_TIMEOUT` was ever on the loop's upsert path; the 120s
  `local_state_or_from_network` is `OperationMode::Local`-only and unrelated.
- Updated `.claude/rules/operations.md` — the `RELATED_FETCH_TIMEOUT`-on-the-loop
  claim is now off-loaded (WASM still serial, only the fetch wait moved).
- Fixed the `handle_deferred_resume` doc (uses the deferrable path); documented
  the per-contract ordering semantics at the deferral site.

Rebased onto current `origin/main` (drops the stale #4390 nat_subscription diff
artifact). `cargo fmt` + `clippy --lib --tests -D warnings` clean.

[AI-assisted - Claude]

---

## Review round 3 update (FIFO/sweep edge cases)

Round-2's per-contract-FIFO machinery introduced 4 blocking edge-case defects; round 3 closes them. They all interact through the per-contract deferral block, so they were fixed together and the D1-swept/D2-reblock timeline is walked in the resume-guard comment.

- **#1 — requeue leak.** `requeue_released_events` dropped held events it couldn't re-enqueue (fair queue full), leaking their `waiting_response` slot and hanging the client 300s. It's now `#[must_use]` and RETURNS the overflow; both call sites (sweep + resume) feed them to `answer_rejected_held_events`, which sends a queue-full response (drops the slot) like the normal drain path.
- **#2 — FIFO violation.** Held events were re-queued at the BACK of the bucket, so a fresh same-key event could run before the older held one. Added `FairEventQueue::requeue_front`, which front-loads held events in original FIFO order and returns any overflow.
- **#3 — stale resume clobber.** `release_key` is keyed by contract, not deferral_id, so a late D1 resume after D1 was swept and a held event re-deferred as D2 would re-run D1's stale upsert and remove D2's live block. Added a guard at the top of `handle_deferred_resume`: it proceeds only if `deferred_keys[key].deferral_id == resume.deferral_id`; otherwise it drops the resume untouched (the sweep already answered that client).
- **#4 — untested sweep.** Parameterized `now_nanos` into `sweep_stale_deferrals` for a deterministic unit test (stale released + held requeued + correct Put/Update variant answered + fresh entry not swept).

New tests: `requeue_front_orders_held_before_fresh_same_key`, `requeue_front_returns_overflow_in_fifo_order`, `hold_event_caps_at_max_held_per_deferral`, `sweep_releases_stale_answers_client_and_requeues_held`, `stale_resume_does_not_release_newer_deferral_block`, `fetch_related_off_loop_no_op_manager_is_missing_related`, `rejected_held_event_is_answered_not_leaked`. Added `StashedResponder::for_test`. Updated `.claude/rules/contracts.md` (deferrable vs inline upsert paths).

`cargo fmt` + `clippy --lib --tests -- -D warnings` clean.

[AI-assisted - Claude]

---

## Review round 3.1 update (deferrable-never-inline; evict reclaim; dead branch)

- **[P2] Deferrable mode can no longer inline-fetch on the serial loop.** The deferrable local-only check counted a caller-SUPPLIED related (`initial_related`) as "local" and fell through to the join_all/network path, which checks ONLY the `state_store` — so a supplied-but-unpersisted related re-requested on resume could hit `fetch_related_via_network` inline, bypassing the one-deferral cap. Both deferrable branches (`fetch_related_for_validation` and the `update_state requires()` branch) now resolve from `state_store` + `initial_related` ONLY and return `DeferRelated(missing)` for anything unresolved; the network join_all is structurally the `else` (non-deferrable) branch, so it's unreachable in deferrable mode. The non-deferrable path (delegates/direct callers) still escalates inline — correct there. New test `deferrable_supplied_related_never_inline_fetches` asserts the network override is NEVER called (calls==0) for a re-requested supplied related, and the outcome is a depth>1 error, not a network result.
- **[LOW #1] EvictContract disk-reclamation on the reject path.** `answer_rejected_held_events` now calls `track_pending_reclamation_if_evict` before `send_queue_full_response` (matching the other three reject sites), so a held `EvictContract` rejected on requeue doesn't silently leak on-disk storage (#4212 class). The helper (and `sweep_stale_deferrals`) now take `&mut CH`. Pinned by `answer_rejected_held_events_tracks_evict_reclamation_pin_test`.
- **[LOW #2] Dead branch removed** in `fair_queue::requeue_front`: the unreachable `if queue.len() >= MAX && queue.is_empty()` cleanup. The bucket length is now checked WITHOUT `or_default()`, so a rejected push for a brand-new id never creates an empty bucket.

292 contract-module + fair_queue tests pass; `cargo fmt` + `clippy --lib --tests -- -D warnings` clean. Rebased onto current origin/main.

[AI-assisted - Claude]

---

## Review round 5 update (delete the per-contract-FIFO machinery)

Rounds 2–3 built a per-contract-FIFO hold mechanism so a contract's later
events would queue behind a mid-deferral op for the same key, and patched its
edge cases twice (global-cap requeue, stale held events, stale resume clobber).
The external reviewer then found another defect of the same class: under heavy
same-key load, `requeue_front` rejects older held events while newer
bucket-resident same-key events stay. Rather than patch the FIFO machinery a
third time, this round **deletes it entirely and accepts reordering** (the
maintainer-approved decision).

### What was deleted
- `deferred_keys` map + `block_key` / `release_key` / `is_key_deferred`
- `hold_event` / `HoldOutcome` / the `held` `VecDeque` / `MAX_HELD_PER_DEFERRAL`
- `pop_runnable_event` + `RunnableEvent` — the loop now drains the fair queue
  directly into `handle_contract_event`
- `FairEventQueue::requeue_front`, `requeue_released_events`,
  `answer_rejected_held_events`
- All the per-contract-FIFO / sweep-ordering tests
- `sweep_stale_deferrals` is no longer needed for FIFO lifecycle; the
  `ResumeGuard` (round-4) already guarantees exactly-once resume delivery, so a
  dropped waiter answers the parked client with `MissingRelated` via Drop.

### What was kept (the sound off-load core)
- `maybe_defer_upsert` off-loading: spawn the waiter, park the client responder
  in `stashed` keyed by `deferral_id`, re-enqueue via the loop-owned resume
  channel (non-blocking `unbounded_send`)
- `ResumeGuard` exactly-once delivery (`Option::take` on success; Drop delivers
  `MissingRelated`)
- `handle_deferred_resume` re-runs the deferrable upsert against current state
  and answers the stashed client exactly once
- `MAX_INFLIGHT_DEFERRALS` cap + fail-fast → `MissingRelated`
- depth=1 / one-deferral-per-op (a 2nd `DeferRelated` converts to
  `MissingRelated`)
- `track_pending_reclamation_if_evict` — still used on the normal channel-drain
  reject paths, so `EvictContract` disk reclamation is unaffected

### Correctness: no lost update on concurrent same-key deferrals
Without the per-key block, two same-key ops can now both defer concurrently. The
resume re-runs the deferrable upsert via
`bridged_upsert_contract_state_inner`, which reads `current_state` FRESH from
`state_store.get(&key)` at upsert time (`runtime.rs:1563`). So two same-key ops
that both defer reconcile through the contract's CRDT merge — neither clobbers
the other.

### Accepted behavior change
A GET/UPDATE for a contract whose PUT is mid-deferral now runs **immediately**
instead of waiting behind it. Documented at the deferral site and in
`.claude/rules/contracts.md` (the per-contract-FIFO claim is dropped).

### Tests
- Deleted the per-contract-FIFO tests.
- New `same_key_get_during_deferred_put_runs_promptly`: gates an off-loop fetch
  for a PUT of contract A, then issues a same-key GET for A under a 2s timeout
  and asserts it resolves promptly (not held); releases the gate and asserts A's
  PUT still gets its own `PutResponse`.
- Kept survivors green: `dropped_waiter_guard_answers_client_missing_related`,
  `resume_guard_success_sends_exactly_one_resume`,
  `deferred_related_fetch_does_not_block_local_get`,
  `deferral_cap_fails_fast_with_missing_related`, the UPDATE-defer tests,
  `second_related_request_after_resume_does_not_defer_again`,
  `client_disconnect_during_deferral_does_not_wedge`,
  `fetch_related_off_loop_no_op_manager_is_missing_related`.

14 hol_4391 + 287 contract-module + 14 fair_queue tests pass; `cargo fmt` +
`clippy --lib --tests -- -D warnings` clean. Net **-774 lines** (substantially
net-negative — the FIFO machinery was more complex than the problem it solved).
Rebased onto current `origin/main`.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/contracts.md
+++ b/.claude/rules/contracts.md
@@ -257,6 +257,26 @@ MUST:
   and the local-only path forced wasted ResyncRequest round trips
   on every send. The escalation is bounded by RELATED_FETCH_TIMEOUT.
   See PR #4006 / freenet/mail#80.
+
+- Off-loop deferral (#4391): there are now TWO entry points into the
+  bridged upsert.
+  * The NON-deferrable path (`upsert_contract_state`, used by
+    delegate-driven PUTs and direct callers) keeps the INLINE
+    `start_sub_op_get` escalation described above — it awaits the
+    network GET in place, bounded by RELATED_FETCH_TIMEOUT.
+  * The DEFERRABLE path (`upsert_contract_state_deferrable`, used by the
+    serial `contract_handling` loop) resolves related contracts
+    LOCAL-ONLY first. On a local miss it does NOT await the network GET
+    inline — it returns `UpsertOutcome::DeferRelated(missing)`, and the
+    loop off-loads the fetch to a background task and re-enqueues a
+    continuation (`contract.rs::maybe_defer_upsert` /
+    `handle_deferred_resume`). This keeps the single-threaded loop from
+    stalling on the 10s fetch while other contracts' events (including
+    cached GETs) drain. WASM validate/store still runs serially on the
+    loop; only the network WAIT moves off-loop. While a contract has an
+    in-flight deferral the loop HOLDS that contract's later events to
+    preserve per-contract FIFO; the block is released (and held events
+    re-queued at the FRONT) on resume or on a TTL sweep.
 ```
 
 ### WHEN a contract's validate_state returns RequestRelated
--- a/.claude/rules/contracts.md
+++ b/.claude/rules/contracts.md
@@ -273,14 +273,18 @@ MUST:
     `handle_deferred_resume`). This keeps the single-threaded loop from
     stalling on the 10s fetch while other contracts' events (including
     cached GETs) drain. WASM validate/store still runs serially on the
-    loop; only the network WAIT moves off-loop. While a contract has an
-    in-flight deferral the loop HOLDS that contract's later events to
-    preserve per-contract FIFO; the block is released (and held events
-    re-queued at the FRONT) by the deferral's resume. The off-loop waiter
+    loop; only the network WAIT moves off-loop. The off-loop waiter
     delivers EXACTLY ONE resume via an RAII `ResumeGuard` (Drop delivers a
     MissingRelated resume if the task is dropped before sending), so every
-    deferral terminates with exactly one client answer — no TTL sweep or
-    stale-resume guard is needed.
+    deferral terminates with exactly one client answer.
+    There is NO per-contract ordering/blocking: a GET/UPDATE for a contract
+    whose PUT is mid-deferral runs immediately in normal fair-queue order and
+    may observe the pre-deferral state — this reordering is intended (the
+    per-contract-FIFO machinery was removed in #4391 round 5). The deferring
+    op's own response is still delivered only on commit (its responder is
+    parked by `deferral_id`), and the resume re-reads CURRENT state, so
+    concurrent same-key deferrals reconcile via the contract's CRDT merge
+    exactly as two concurrent same-key UPDATEs would — no lost update.
 ```
 
 ### WHEN a contract's validate_state returns RequestRelated
--- a/.claude/rules/contracts.md
+++ b/.claude/rules/contracts.md
@@ -276,7 +276,11 @@ MUST:
     loop; only the network WAIT moves off-loop. While a contract has an
     in-flight deferral the loop HOLDS that contract's later events to
     preserve per-contract FIFO; the block is released (and held events
-    re-queued at the FRONT) on resume or on a TTL sweep.
+    re-queued at the FRONT) by the deferral's resume. The off-loop waiter
+    delivers EXACTLY ONE resume via an RAII `ResumeGuard` (Drop delivers a
+    MissingRelated resume if the task is dropped before sending), so every
+    deferral terminates with exactly one client answer — no TTL sweep or
+    stale-resume guard is needed.
 ```
 
 ### WHEN a contract's validate_state returns RequestRelated
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -79,8 +79,18 @@ On a relay driver's response path, any operation that enqueues on the
 single-threaded `contract_handling` event loop (GetQuery, PutQuery,
 validate_state, `RequestRelated` recursion, etc.) MUST be sequenced
 AFTER the upstream forward. Reversing this puts WASM `validate_state`
-and the 10s `RELATED_FETCH_TIMEOUT` directly on the upstream-visible
-critical path, where they stack across multi-hop traversals.
+directly on the upstream-visible critical path, where it stacks across
+multi-hop traversals.
+
+Note (#4391): the related-contract NETWORK fetch's 10s
+`RELATED_FETCH_TIMEOUT` is no longer awaited inline on the
+`contract_handling` loop — a PUT/UPDATE that needs a related contract
+not held locally now off-loads that fetch to a background task and
+re-enqueues a continuation (see `contract.rs::maybe_defer_upsert`), so
+the loop no longer stalls for that 10s window. WASM
+`validate_state`/`update_state`/store still run serially on the loop,
+so this invariant still holds for the WASM work; only the related-fetch
+WAIT no longer contributes to the stall.
 
 ```
 RIGHT (issue #4155 fix):
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -4521,6 +4521,73 @@ mod hol_4391_tests {
         );
     }
 
+    /// Round-3.2 (unit): on GLOBAL-cap overflow, `requeue_front` must accept the
+    /// OLDEST held event (which pops first) and reject the NEWEST ones — never
+    /// the reverse. With held `[e1,e2,e3]` and exactly ONE global slot free, e1
+    /// must be re-queued (and pop first) while e2,e3 are returned as overflow in
+    /// FIFO order. The previous reverse-iterate-with-per-event-cap code accepted
+    /// e3 (newest) and rejected e1,e2 — a per-contract FIFO violation. (Codex.)
+    #[test]
+    fn requeue_front_global_overflow_keeps_oldest_rejects_newest() {
+        let mut fq = fair_queue::FairEventQueue::new();
+
+        // Saturate the GLOBAL queue to MAX_TOTAL_FAIR_QUEUE - 1 (one slot free),
+        // spread across distinct contract keys so no single per-contract bucket
+        // overflows (each key holds at most MAX_QUEUED_PER_CONTRACT).
+        let target = fair_queue::MAX_TOTAL_FAIR_QUEUE - 1;
+        let mut pushed = 0u64;
+        let mut filler = 0u64;
+        while (pushed as usize) < target {
+            let key = instance_id(format!("filler_{filler}").as_bytes());
+            filler += 1;
+            let room = (target - pushed as usize).min(fair_queue::MAX_QUEUED_PER_CONTRACT);
+            for _ in 0..room {
+                fq.try_push(
+                    handler::EventId {
+                        id: 10_000 + pushed,
+                    },
+                    get_event(key),
+                )
+                .expect("filler push under caps");
+                pushed += 1;
+            }
+        }
+        assert_eq!(fq.total_queued(), target, "exactly one global slot free");
+
+        // Held events for a FRESH key (its bucket is empty → bucket_free=100, so
+        // the binding constraint is the single free GLOBAL slot → capacity 1).
+        let held_key = instance_id(b"held_global_overflow");
+        let mut held = std::collections::VecDeque::new();
+        held.push_back((handler::EventId { id: 1 }, get_event(held_key)));
+        held.push_back((handler::EventId { id: 2 }, get_event(held_key)));
+        held.push_back((handler::EventId { id: 3 }, get_event(held_key)));
+
+        let rejected = fq.requeue_front(held);
+
+        // e2, e3 (newest) rejected, in FIFO order.
+        let rejected_ids: Vec<u64> = rejected.iter().map(|(id, _)| id.id).collect();
+        assert_eq!(
+            rejected_ids,
+            vec![2, 3],
+            "on global overflow, the NEWEST events must be rejected (in FIFO order), \
+             keeping the oldest — got {rejected_ids:?}"
+        );
+
+        // e1 (oldest) was accepted and, for its key, pops first. Drain the
+        // held_key bucket and confirm e1 is the only/first event from it.
+        let mut from_held_key = Vec::new();
+        while let Some((id, ev)) = fq.pop() {
+            if fair_queue::extract_contract_id(&ev) == Some(held_key) {
+                from_held_key.push(id.id);
+            }
+        }
+        assert_eq!(
+            from_held_key,
+            vec![1],
+            "the OLDEST held event (e1) must be the one accepted on overflow"
+        );
+    }
+
     /// SHOULD-FIX (unit): the per-deferral held buffer caps at
     /// MAX_HELD_PER_DEFERRAL — the (cap+1)th hold returns `Full` carrying the
     /// event back.
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -121,6 +121,110 @@ struct DeferredResume {
     fetched: Result<Vec<(ContractInstanceId, WrappedState)>, ExecutorError>,
 }
 
+/// RAII guard that guarantees the off-loop waiter delivers EXACTLY ONE terminal
+/// [`DeferredResume`] for its deferral — whether the fetch completes, or the
+/// waiter task is dropped/panics/cancelled before sending.
+///
+/// This is the load-bearing invariant of the #4391 deferral lifecycle: because
+/// every deferral is answered by exactly one resume, the loop never needs a TTL
+/// sweep to recover a wedged waiter, and `handle_deferred_resume` never needs a
+/// stale-resume guard (a resume is always for the live, still-blocked deferral).
+///
+/// - On the success path the waiter calls [`send`](Self::send) with the real
+///   fetch result; this takes the payload so `Drop` becomes a no-op.
+/// - On any early exit (drop / panic / cancellation) before `send`, `Drop`
+///   delivers a terminal resume carrying a `MissingRelated` fetch error, which
+///   `handle_deferred_resume` surfaces to the client and uses to release the
+///   contract's per-contract block.
+///
+/// Exactly once: never zero (Drop covers the early-exit path), never twice (the
+/// success send `take`s the payload, so Drop sees `None` and does nothing).
+///
+/// The resume channel is unbounded, so both sends are non-blocking
+/// `unbounded_send` (no `.await`, no back-pressure). A send failure means the
+/// loop has shut down — nothing left to answer.
+struct ResumeGuard {
+    /// `None` once a resume has been sent (success or drop). Holds everything a
+    /// [`DeferredResume`] needs except its `fetched` result.
+    payload: Option<ResumePayload>,
+}
+
+struct ResumePayload {
+    resume_tx: tokio::sync::mpsc::UnboundedSender<DeferredResume>,
+    deferral_id: u64,
+    key: ContractKey,
+    update: Either<WrappedState, StateDelta<'static>>,
+    related_contracts: RelatedContracts<'static>,
+    code: Option<ContractContainer>,
+    is_put: bool,
+}
+
+impl ResumeGuard {
+    fn new(payload: ResumePayload) -> Self {
+        Self {
+            payload: Some(payload),
+        }
+    }
+
+    /// Deliver the terminal resume with the real fetch result. Consumes the
+    /// payload so a subsequent `Drop` is a no-op (exactly-once).
+    fn send(mut self, fetched: Result<Vec<(ContractInstanceId, WrappedState)>, ExecutorError>) {
+        if let Some(p) = self.payload.take() {
+            Self::deliver(p, fetched);
+        }
+    }
+
+    fn deliver(
+        p: ResumePayload,
+        fetched: Result<Vec<(ContractInstanceId, WrappedState)>, ExecutorError>,
+    ) {
+        let ResumePayload {
+            resume_tx,
+            deferral_id,
+            key,
+            update,
+            related_contracts,
+            code,
+            is_put,
+        } = p;
+        if resume_tx
+            .send(DeferredResume {
+                deferral_id,
+                key,
+                update,
+                related_contracts,
+                code,
+                is_put,
+                fetched,
+            })
+            .is_err()
+        {
+            tracing::debug!(
+                contract = %key,
+                "Deferred resume channel closed; contract-handling loop gone"
+            );
+        }
+    }
+}
+
+impl Drop for ResumeGuard {
+    fn drop(&mut self) {
+        // Early exit before `send` (waiter dropped / panicked / cancelled):
+        // deliver a terminal MissingRelated resume so the deferral is still
+        // answered exactly once and its per-contract block is released.
+        if let Some(p) = self.payload.take() {
+            let key_id = *p.key.id();
+            tracing::warn!(
+                contract = %p.key,
+                deferral_id = p.deferral_id,
+                "Off-loop waiter dropped before sending — delivering MissingRelated \
+                 resume so the deferral terminates and its block is released (#4391)"
+            );
+            Self::deliver(p, Err(ExecutorError::missing_related(key_id)));
+        }
+    }
+}
+
 /// Outcome of attempting to hold a same-key event behind an in-flight deferral.
 enum HoldOutcome {
     /// The event was held behind the deferral (FIFO).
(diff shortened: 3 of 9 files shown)


sanity (commented):
## Comprehensive PR Review: #4394

### Summary
- **PR:** fix(contract): off-load network related-fetch from contract_handling loop (#4391)
- **Type:** fix (core contract op pipeline / concurrency)
- **Review tier:** Full (high-risk surface: concurrency + state authorization on the single serial loop)
- **Reviewers run:** `freenet:code-first`, `freenet:testing`, `freenet:skeptical`, `freenet:big-picture` + external **Codex** (non-Claude)
- **CI:** deterministic checks green; Unit/Integration/Sim/Windows pending at review time — gate merge on green at the current HEAD.

The core mechanism is sound and was verified by multiple reviewers: only the *network wait* leaves the serial loop; WASM validate/merge stays serial; the deferrable upsert rolls back partial work before returning `DeferRelated`; resume re-runs against **freshly-reloaded** current state; exactly one client response is delivered per deferral; the one-deferral cap genuinely can't be bypassed; and the headline regression test drives the **real** loop and was verified fails-without/passes-with. It is **not** mergeable as-is: an external model flagged a per-contract ordering regression, and there are real robustness + test-coverage gaps.

---

### Must Fix (Blocking)

**1. Per-contract ordering is lost while an upsert is deferred.** `crates/core/src/contract.rs:1203` — when a PUT/UPDATE defers, the loop returns immediately **without** marking the contract as having an in-flight operation. Later events for the *same key* are then popped from the fair queue and run **before** the deferred upsert resumes. So a GET/UPDATE for a newly-PUT contract can observe `MissingContract` or stale state; the old serial loop preserved per-contract FIFO by awaiting the original upsert. *(External Codex [P1]; `code-first` flagged the same reordering as a semantic change.)*

This needs a decision (see note below), but as written it is a behavioral regression on a high-risk path. Recommended fix: keep a **per-contract** in-flight marker during the deferral and requeue same-key events until the resume commits — this preserves per-key FIFO while still unblocking *other* contracts (which is the actual #4391 symptom). The client's own `PUT → response → GET` causality is already preserved (the responder fires only on commit), so this only affects concurrent same-key events from other sources.

---

### Should Fix (Important)

2. **Cap the resume drain.** `contract.rs:875` drains `resume_rx` with `while let Ok = try_recv()` and **no per-iteration cap**, unlike the adjacent event drain (`MAX_DRAIN_BATCH`) and delegate drain (`MAX_DELEGATE_DRAIN_BATCH`). Each resume runs a full WASM upsert, so up to 256 resumes execute back-to-back with no yield to the fair queue — re-introducing head-of-line latency for cached GETs, the exact symptom this PR removes. Cap it and interleave with the fair queue. *(skeptical [Med]; `bug-prevention-patterns.md` "per-iteration caps on high-throughput arms".)*
3. **TTL + monitor the deferral machinery.** The `stashed` responder map (`contract.rs:115`) has no age-sweep, and the off-loop waiter is a fire-and-forget `GlobalExecutor::spawn` whose `JoinHandle` is dropped (not registered with `BackgroundTaskMonitor`, contrary to `code-style.md`). If a waiter ever fails to send, the stashed entry leaks and the client hangs; once 256 leak, `at_capacity()` is permanently true → the fix is silently and permanently disabled with no recovery. Register the waiter handle and/or add an absolute-age sweep that answers stale entries with `MissingRelated`. *(skeptical [Med]; `AGENTS.md` "cleanup exemptions MUST be time-bounded".)*
4. **Capacity fallback should fail fast, not block inline.** At `MAX_INFLIGHT_DEFERRALS`, `contract.rs:1122` falls back to `upsert_contract_state`, which awaits the network fetch **on the loop** — the 257th op under a burst reintroduces the HoL stall. Prefer surfacing `MissingRelated` immediately over an inline fetch. *(Codex [P2]; skeptical/code-first.)*
5. **Test coverage gaps (regression-sensitive path):**
   - No **UPDATE-path** deferral test — all five new tests are PUT-only, and the UPDATE response path (`NoChange → fetch_contract` refetch) is materially different. Add UPDATE-defer-resume + UPDATE-defer-failure via the real loop (`UpdateOverride::RequiresRelated` already exists).
   - The real `fetch_related_off_loop` is **never exercised** — every test stubs `OFF_LOOP_FETCH_OVERRIDE`, so `start_sub_op_get`/`join_all`/the `SubOpGetOutcome` mapping/timeout/`op_manager==None` paths have zero coverage. Add a real-`op_manager` test (the `related_contract_tests.rs` stubbed-network pattern works).
   - Should-add: the 256-cap boundary, the off-loop timeout→`MissingRelated` path, and client-disconnect-during-deferral (StashedResponder `Err` + no `stashed` leak).
   - Test hygiene: `set_off_loop_fetch_override(None)` runs only on success; wrap in an RAII `Drop` guard so a panicking test doesn't leak a stale stub into siblings.
6. **Rebase before merge.** Branch is 1 commit behind `origin/main`; the `nat_subscription*` "−608 lines" in the diff is a stale-base artifact (#4390), not a removal. Re-confirm CI green after rebase.

---

### Consider / Docs
- Correct the timeout-reconciliation claim: the loop path only ever used the 10s `RELATED_FETCH_TIMEOUT`; the 120s `local_state_or_from_network` path is `OperationMode::Local`-only, **not** on the serial loop, and is not addressed here. Say so explicitly so it isn't read as forgotten work. *(code-first, big-picture, skeptical all verified this.)*
- Update `.claude/rules/operations.md:76-83` — its claim that the 10s `RELATED_FETCH_TIMEOUT` sits on the loop's critical path is now stale (it's off-loaded). *(big-picture [Med].)*
- Fix the `handle_deferred_resume` doc comment (`contract.rs:957-958`): resume uses `upsert_contract_state_deferrable`, not the plain method (this matters for the one-deferral-cap reasoning).
- Document the per-key reordering semantics wherever finding 1 lands; note the memory ceiling (256 × up to 50 MiB) if a byte budget is considered.

---

### Verified correct
Deferrable path is taken in production (RuntimePool overrides; `op_manager_handle` is `Some`); the loop genuinely stops blocking on the network; no double-send / lost response; rollback before defer (no partial commit); resume re-validates against fresh state (no lost update under a concurrent same-key UPDATE *merge*, though see finding 1 for ordering/visibility); one-deferral cap unbypassable; the #4077 pin test was tightened (not weakened); 5/5 tests pass, clippy clean.

### Verdict
**Needs Changes — Re-review Required After Fix.** Finding 1 (per-contract ordering) is a design decision on a high-risk path and is being escalated to the maintainer; findings 2–5 are concrete robustness/test gaps. Mechanism otherwise sound. HEAD reviewed: `66fa79f8`. Re-run the full review on the fixed + rebased HEAD before merge.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review (Round 2): #4394

### Summary
- **PR:** fix(contract): off-load network related-fetch from contract_handling loop (#4391), HEAD `f8fd86a2`
- **Review tier:** Full (concurrency on the single serial contract loop)
- **Reviewers run:** `freenet:code-first`, `freenet:testing`, `freenet:skeptical`, `freenet:big-picture` + external **Codex**
- This is a re-review of the round-2 fixes (per-contract FIFO, capped resume-drain, TTL sweep, fail-fast, more tests). Round 2 correctly closed round-1's per-contract-ordering regression and all round-1 robustness gaps. The **new** FIFO machinery introduced a few real edge-case defects.

The core deferral mechanism is verified sound across reviewers: response-on-every-path, fresh-state resume, one-deferral cap, rollback-before-defer, cross-key unblocking, bounded buffers, the TTL idle-wake guard (no permanent wedge), and the deterministic (DST) `now_nanos`. No tests removed; the #4077 pin is intact (tightened, not weakened).

---

### Must Fix (Blocking)

**1. `requeue_released_events` (contract.rs:1228-1244) leaks the responder + hangs the client on a full fair queue.** On `try_push` failure it only `debug!`-logs and drops the held event — it does **not** `drop_waiting_response`/`send_queue_full_response`, contrary to its own comment. The held event's `waiting_response` entry then leaks permanently (unbounded under sustained hot-contract deferral load) and the client blocks `CH_EV_RESPONSE_TIME_OUT` = 300s before `NoEvHandlerResponse`. Reachable: held events requeue into the same per-contract bucket (cap `MAX_QUEUED_PER_CONTRACT=100`); if fresh same-key events accumulated, re-pushing up to 100 held can exceed the cap. *(skeptical [High] + Codex [P1] second half.)* Fix: answer the rejected event with `ContractQueueFull` (or at minimum `drop_waiting_response`) at both call sites. Add a test (the normal-path cleanup is pinned; this path isn't).

**2. `requeue_released_events` does not deliver its documented FIFO-ahead-of-fresh ordering.** The comment says held events are re-enqueued "ahead of fresh events for the same key," but `fair_queue.try_push` appends to the **back** of the bucket. If a fresh same-key event is present at requeue time, the older held event runs after it — violating the per-contract FIFO this round exists to provide. *(Codex [P1] first half; disputes skeptical's "FIFO safe", which only analyzed the all-held-drained case.)* Fix: push held events to the **front** in original order, OR prove + document that the bucket is always empty of same-key events at requeue time and correct the comment. Add a FIFO-with-a-fresh-same-key-event test.

**3. Stale/swept resume releases a *newer* deferral's block and may commit stale data** (contract.rs:1222/1367 release; handle_deferred_resume). `release_key(contract_id)` is keyed by contract, **not** `deferral_id`. Sequence: D1(K) wedges past 2×TTL → swept (answered + block released) → a held event runs and creates D2(K) (new block) → D1's late `DeferredResume` arrives → `handle_deferred_resume` re-runs the upsert with D1's **stale** update and calls `release_key(K)`, which removes **D2's** live block (breaking FIFO) and may commit older data. *(code-first [Medium]; same root as skeptical's sweep-vs-resume [Med] where a successful resume crosses the TTL and the client wrongly sees `MissingRelated` while state commits.)* Fix: guard the resume on `deferred_keys[key].deferral_id == resume.deferral_id`; if absent (swept) or mismatched (newer deferral owns the key), **skip** both `release_key` and the upsert/commit. One guard fixes both the FIFO break and the stale commit.

**4. The TTL sweep — the permanent-wedge safety net — has no test** (`sweep_stale_deferrals`, contract.rs:1173). A refactor that drops the sweep call or breaks the `>= 2×TTL` compare would pass all current tests, re-opening the self-disabling-node leak. *(testing [Must].)* Fix: parameterize `now` into the function → deterministic unit test (block a key with a held event, sweep at `now = t0 + 2×TTL + 1`, assert the key is released, the held event requeued, the client answered, and a fresh entry is NOT swept). May need a `#[cfg(test)]` `StashedResponder` constructor.

---

### Should Fix
5. Unit-test `MAX_HELD_PER_DEFERRAL` over-cap (`HoldOutcome::Full`, contract.rs:212) — pure function, ~15 lines, the only bound on the held buffer. *(testing.)*
6. Add an `op_manager = None` unit test of `fetch_related_off_loop` (the real `start_sub_op_get`/`join_all`/timeout orchestration is otherwise uncovered — every test stubs the override). *(testing.)*
7. Update `.claude/rules/contracts.md:246-291` — still describes the bridged upsert as escalating to an **inline** network GET; add the #4391 note that the production loop path resolves local-only and defers off-loop (the PR fixed `operations.md` but missed this rule). *(big-picture.)*

### Consider
- Assert FIFO requeue on a *failed/swept* deferral (the error-path `requeue_released_events`), not only successful commit.
- `handle_deferred_resume`'s "can't be popped until the next iteration" comment is imprecise (can be the same iteration in the drain batch; behavior still correct = after commit).

---

### Verdict
**Needs Changes — Re-review Required After Fix.** Four blocking items, all in the new round-2 FIFO/sweep machinery; the underlying off-loop mechanism is sound. HEAD reviewed: `f8fd86a2`. Re-run the full review on the fixed HEAD before merge (concurrency = high-risk).

[AI-assisted - Claude]

PR #4395, opened 2026-06-13
fix: drain tracker by stream_id on abort

### LINKED ISSUE #4345: GET fails with 'stream assembly: no fragments received within inactivity timeout' for multi-fragment state; contract never caches, SUBSCRIBE then rejects

netsirius:
## Problem

#4353 capped per-packet retransmissions at 12 so never-ACKed packets eventually release their flight-size bytes via `ResendAction::Abandon`. #4374 turned stream-level aborts into the stream-scoped `OutboundStreamFailed` so they no longer tear down the whole connection. #4367 gave the GET driver a retry loop so a single assembly failure does not burn the operation.

What remains: abandonment is slower than the abort. `CWND_WAIT_TIMEOUT` is 3 s; the per-packet abandon path needs ≥6 s (12 × 500 ms `MIN_RTO` floor). During that gap the aborted stream's bytes stay pinned in `flightsize`, and any other operation multiplexed onto the same connection — a concurrent GET, a follow-up SUBSCRIBE, an op the router routes through the same peer — inherits the wedge and dies the same way. The user-reported "Rejecting SUBSCRIBE: contract WASM not cached locally" cascade is exactly this: the multi-fragment GET wedged the connection; the follow-up SUBSCRIBE arrived inside the 3–6 s window and inherited it.

The fix has to release flight-size at abort time, atomically with removing the dying stream's packets from the tracker — otherwise late ACKs or the abandon path double-decrement and silently under-count live streams (per sanity's design constraint on the issue, 2026-06-08).

## Solution

Tag each entry in `SentPacketTracker::pending_receipts` with `Option<StreamId>`. `Some(id)` for `StreamFragment` payloads, `None` for handshake / `ShortMessage` / `NoOp` / `Ping` / `Pong` so they are invisible to the new sweep. Extraction happens in one place — `extract_stream_id_from_payload` called from `packet_sending` — and the tag rides on `ResendAction::Resend` / `TlpProbe` so retransmissions re-register without losing it.

New `SentPacketTracker::drop_stream(id) -> usize` sweeps every entry tagged with `id` out of `pending_receipts`, `retransmit_counts`, `retransmitted_packets`, and `tlp_sent_packets`, and returns the byte total. It does NOT touch flight size — the caller passes the returned count to `CongestionControl::release_flightsize(bytes)`. The split is what prevents double-decrement: once a packet leaves the tracker, any later ACK or per-packet abandon finds nothing and becomes a no-op, so each packet contributes exactly once to flight-size accounting.

A `drain_stream_from_tracker` helper in `outbound_stream.rs` does drop + release in one call. Six stream-terminating early returns invoke it: cwnd-wait in `send_stream` and `pipe_stream`, the two `packet_sending` send-failure returns in both, plus `pipe_stream`'s inactivity-stall and inbound-error arms. The two send-failure sites drain even though the failing packet itself never registered — earlier fragments of the same stream are already in flight and must not bleed into the next attempt.

`PendingReceiptEntry` becomes a struct (was a positional tuple) so future per-packet metadata can be added additively — destructuring matches use `..` to ignore unused fields. `.claude/rules/transport.md` documents the new invariant alongside the #4353 release path.

## Testing

Four `drop_stream` unit tests in `sent_packet_tracker.rs` pin the contract: filtering by `stream_id` (other streams and untagged entries survive), per-packet auxiliary state cleared, late ACK after drop is a no-op (the anti-double-decrement guarantee), stale `resend_queue` entries no longer surface as `Resend`/`Abandon`. `extract_stream_id_from_payload`'s test pins the exhaustive match so adding a new `SymmetricMessagePayload` variant fails compilation rather than silently defaulting to untagged.

`cwnd_wait_abort_drains_tracker_and_unwedges_connection` in `outbound_stream.rs` runs the real `send_stream` with a pre-seeded wedge (`cwnd = 1`, tracker pre-loaded with 3000 bytes), asserts that after the abort the tracker and `flightsize` are both at 0, and registers a fresh stream B's packet to prove the connection accepts new work. Temporarily reverting `drain_stream_from_tracker` to a no-op makes this test fail with `tracker still contains 2 packets`; restoring it returns to green.

Existing E2E coverage (`test_streaming_with_packet_loss`, `test_streaming_get_1mb_with_packet_loss`, `test_streaming_get_retries_after_assembly_failure`) passes on this branch. No new E2E test: a deterministic reproduction of the concurrent-ops-on-shared-connection scenario needs fault-injection infrastructure that does not yet exist. Per the validation pattern from #4353, the empirical signal post-merge is the OTLP `transfer_failed` rate with cause `cwnd wait timeout` (currently ~91% of all transfer failures on the nova gateway).

## Fixes

Closes #4345.

### DIFF
--- a/.claude/rules/transport.md
+++ b/.claude/rules/transport.md
@@ -103,6 +103,51 @@ pinning flight size at cwnd and stalling every subsequent stream
   - Retransmissions MUST stay bounded. Do not make retransmit infinite again.
 ```
 
+### Atomic stream-abort drain (issue #4345, follow-up to #4353)
+
+```
+A THIRD flight-size release path exists alongside ACK and abandonment: when an
+outbound stream aborts before its packets reach MAX_PACKET_RETRANSMITS, the
+abort site MUST atomically remove the stream's in-flight packets from the
+tracker AND release their bytes from flight size, in that order. The
+per-packet abandonment path (#4353) takes ~6s (12 × 500ms RTO floor) — longer
+than CWND_WAIT_TIMEOUT (3s) — so a same-connection retry (PR #4367) opening a
+new stream before abandonment fires would inherit the wedge.
+
+Mechanism:
+  - Packets are tagged with Option<StreamId> at report_sent_packet_with_token.
+    Some(id) for StreamFragment payloads; None for handshake / short / no-op /
+    ping packets that have no owning stream (those are invisible to drop_stream).
+  - SentPacketTracker::drop_stream(stream_id) -> usize sweeps pending_receipts
+    for that stream and returns the byte total. It does NOT touch flight size.
+  - The caller (outbound_stream.rs abort sites) then calls
+    CongestionControl::release_flightsize(bytes) with that total.
+
+Why split tracker drain from flight-size release:
+  - Atomicity. Once a packet is removed from pending_receipts, any subsequent
+    ACK or abandon path finds it gone and becomes a no-op. This guarantees
+    each packet's bytes are decremented exactly once across its lifetime,
+    preventing the double-decrement that would otherwise silently under-count
+    live streams under saturating arithmetic.
+  - Decoupling. The tracker owns "what is in flight"; the controller owns
+    "the counter." No tracker code path touches flight size; no controller
+    code path touches tracker entries. This keeps the modules unit-testable
+    in isolation and makes the single-source-of-truth easy to audit.
+
+Abort sites in outbound_stream.rs that MUST drain (all use the
+drain_stream_from_tracker helper):
+  1. send_stream cwnd-wait timeout
+  2. pipe_stream cwnd-wait timeout
+  3. pipe_stream inactivity stall (upstream feed dry)
+  4. pipe_stream inbound-stream error (upstream cancelled)
+  5. send_stream post-packet_sending Err(e) propagation
+  6. pipe_stream post-packet_sending Err(e) propagation
+
+A new release path that touches flight size without also removing tracker
+entries is PROHIBITED — saturating arithmetic would mask the resulting
+under-count and corrupt live streams' accounting on lossy paths.
+```
+
 ### Shadow per-peer RTT registry (issue #4074, Phases 1 + 1.5)
 
 `transport/rolling_rtt_stats.rs` maintains a process-wide
--- a/apps/freenet-ping/Cargo.lock
+++ b/apps/freenet-ping/Cargo.lock
@@ -1568,7 +1568,7 @@ dependencies = [
 
 [[package]]
 name = "freenet"
-version = "0.2.70"
+version = "0.2.71"
 dependencies = [
  "aes-gcm",
  "ahash",
--- a/crates/core/src/transport.rs
+++ b/crates/core/src/transport.rs
@@ -697,7 +697,7 @@ mod tests {
         sent_tracker.time_source.advance(effective_rto);
         for id in [2, 4] {
             match sent_tracker.get_resend() {
-                ResendAction::Resend(packet_id, packet) => {
+                ResendAction::Resend(packet_id, packet, _) => {
                     assert_eq!(packet_id, id);
                     // Simulate resending packet
                     sent_tracker.report_sent_packet(id, packet);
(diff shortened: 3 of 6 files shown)


iduartgomez (approved):

iduartgomez:
Superseded by #4393, which merged to `main` yesterday (commit `71bfe049c`) and closes the same #4345.

Both PRs fix the flight-size-inheritance-on-stream-abort bug. #4393 is the larger, more complete implementation (1369 vs 500 insertions) and already lands everything this PR introduces:

- `drop_stream(stream_id)` sweep of the tracker by stream tag
- atomic flight-size release at abort time (no waiting on the ≥6 s per-packet abandon path)
- resurrection-safety so a late ACK / abandon after `drop_stream` is a no-op (anti-double-decrement)
- the abort-site drains in `send_stream` / `pipe_stream`
- the `.claude/rules/transport.md` invariant doc

#4393 tags via a `packet_streams` map + `refresh_sent_packet`; this PR tags via a `PendingReceiptEntry { stream_id }` struct + `report_sent_packet_with_token`. Functionally equivalent; rebasing this on top of #4393 would just re-litigate an already-fixed bug and risk regressing the merged version's extra coverage.

Closing as superseded. If any abort site or test here turns out to be uniquely missing from #4393, I'll open a small focused follow-up rather than resurrect this branch.

Issue #3367, opened 2026-03-02
test: fill testing gaps exposed by production incidents (Feb 28 – Mar 1)


sanity:
## Problem

Between Feb 28 and Mar 1 we shipped 7 releases and ~18 bug-fix PRs, almost all discovered in production on nova/technic rather than caught by CI. Reviewing these PRs reveals four specific testing gaps.

## Gap 1: No persistence round-trip test

**PRs this would have caught:** #3350 (params lost on restart), #3352 (subscriptions broken on startup), #3341 (stale cache served after LRU outlives subscription), #3344 (delegate index unrecoverable)

SimNetwork nodes use in-memory state stores, so crash/recover tests (`test_crash_recover_convergence`, `test_multi_step_churn`) verify network convergence but never exercise redb persistence. No test currently:

1. PUTs a contract with params
2. Stops the node (flushing to redb)
3. Restarts from disk
4. Verifies GET returns correct state/params and subscription still works

**Proposed fix:** A `#[freenet_test]` integration test using the real `ContractStore`/redb backend. Single gateway + 1 peer. PUT, subscribe, stop peer, restart, GET and verify. ~10s CI cost.

## Gap 2: Subscription lifecycle / HostingManager decision logic untested

**PRs this would have caught:** #3347 (hosting collapse — only 7 peers serving River), #3363 (368 contracts subscribed due to 3-PR interaction), #3360 (GET fails 94%), #3348 (no proactive state sync)

No test verifies:
- `should_unsubscribe_upstream()` returns false for hosted contracts without local clients
- `is_receiving_updates()` distinguishes active subscription from stale cache
- `contracts_needing_renewal()` doesn't grow unboundedly with cached contracts

These are the decision functions that drive subscription maintenance. Sim tests are too short-lived to trigger multiple renewal cycles, and they don't exercise these code paths with realistic hosting cache states.

**Proposed fix:** Targeted unit tests for `HostingManager` and subscription renewal logic. Mock `Ring` and verify decisions for: hosted-with-subscription, hosted-without-subscription, cached-but-expired. <2s CI cost.

## Gap 3: Transport resilience logic untested

**PRs this would have caught:** #3321 (single UDP send failure kills all 11 connections), #3329 (gateway backoff escalates to 10 minutes)

No test injects I/O errors into the send path to verify connections survive transient failures. No test verifies backoff caps.

[...]

PR #4398, opened 2026-06-13
test: redb persistence round-trip for state, params, subscribe

iduartgomez:
## Problem

`SimNetwork` nodes use in-memory state stores, so existing crash/recover tests never exercise
**redb persistence** across a real node restart. No test PUT a contract with parameters, stopped
the node, restarted from disk, and verified the state/params round-tripped and a subscription
could be re-established. This is Gap 1 of #3367 — the gap behind #3350 (params lost on restart)
and #3352 (subscriptions broken on startup).

## Solution

Add a disk-backed crash/recover regression test (`crates/core/tests/persistence_roundtrip.rs`).

Building this **surfaced a test-validity trap**: a naive in-process version "passes" for a fake
reason. redb's on-disk file lock is held by the contract executor + WS server, which run detached
on the process-wide `GlobalExecutor` and are **not** torn down by a node's graceful shutdown — so
a second in-process node never actually starts, and the first node's still-live WS server answers
the GET on the reused port, validating nothing.

The test therefore runs node 1 and node 2 as **separate `freenet network` subprocesses** sharing
only the on-disk data dir. Node 1's process exit releases the OS file lock, so node 2 genuinely
serves the GET from disk. (The unreleased-lock / detached-shutdown behavior is filed separately as
a latent issue.)

## Testing

- PUTs a contract **with parameters**, stops node 1, restarts node 2 against the same data dir,
  GETs and asserts state + params + key round-tripped, and that a GET-with-subscribe is accepted
  on the restarted node.
- **Negative control**: pointing node 2 at an empty data dir makes the GET time out — proving the
  pass reads node 1's on-disk state, not residual in-process state.
- Startup readiness is a bounded connect-retry (no fixed sleeps): ~2.6 s, 3× non-flaky.
- `cargo clippy -p freenet --lib --tests -- -D warnings` clean.

Scope: cross-node UPDATE propagation and LRU eviction (#3341) / delegate index (#3344) are
explicitly out of scope here.

## Fixes

Refs #3367 (Gap 1: persistence round-trip)

### DIFF
--- a/crates/core/tests/persistence_roundtrip.rs
+++ b/crates/core/tests/persistence_roundtrip.rs
@@ -1,0 +1,394 @@
+//! Disk-backed crash/recover persistence round-trip test.
+//!
+//! Issue #3367 (Gap 1): SimNetwork nodes use in-memory stores, so existing
+//! crash/recover tests never exercise the real redb-backed `ContractStore`
+//! on disk. Nothing PUTs a contract WITH parameters, stops the node so state
+//! flushes to redb, restarts from the SAME on-disk data dir, and verifies a
+//! GET returns the correct state AND parameters and that a subscription can be
+//! re-established afterwards.
+//!
+//! This regression test closes that gap. It guards:
+//! - #3350: parameters lost on restart
+//! - #3352: subscriptions broken on startup
+//!
+//! Out of scope here (this test PUTs a single contract and exercises neither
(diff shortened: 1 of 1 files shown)


iduartgomez:
The redb-lock / detached-shutdown behavior this test works around is now tracked as #4401.

Issue #4382, opened 2026-06-11
Surface stale-node signal: homepage version-mismatch banner + stuck-wrapper notification


iduartgomez:
## Problem

When the service wrapper gets stuck on a stale orphan (see #3967), the node keeps **appearing healthy** even though it's running a binary that is days/weeks out of date:

- The homepage at `http://127.0.0.1:<port>/` serves the **old binary's embedded assets** — the UI loads and looks normal.
- The WebSocket connects to the (orphaned, stale) node, so the connected-peers list updates and the page feels live.
- Nothing anywhere tells the user "you are looking at 0.2.71-1 but 0.2.71 has been on disk for 5 days."

Result: the node silently runs a stale version for days. Observed twice now (original #3967: 19 days on 0.2.37 while 0.2.49 was on disk; 2026-06-11: 5 days stale). In both cases the operator only noticed by accident.

This issue is the **"surface it" half** of #3967 (acceptance items 3 and 4), split out because it's a cheap standalone win that would have given a signal in both incidents — independent of the self-healing exit-43 fix tracked in #3967.

### 1. Homepage version-mismatch banner

The embedded homepage knows its own build version (the version it was compiled with). It can fetch the running node's reported version via the existing API/WS. When they differ, render a visible banner:

> ⚠️ This page was built for Freenet X.Y.Z-N but the running node reports A.B.C. The background service is likely stuck on an old version — run `freenet service stop && freenet service start` (or `kill -9` any stale `freenet network` process holding the port).

This makes the "stale assets" symptom self-diagnosing instead of looking like a frozen-but-fine UI.

### 2. Stuck-wrapper user notification

After N consecutive identical failures (same exit code + same detected cause — e.g. repeated exit 43 against a port already held), the wrapper should surface it rather than only `log_event` into a log nobody reads:

- macOS: `osascript`/user notification ("Freenet service stuck — manual intervention required").
- Cross-platform: write failure-count + last-error to a small status file the homepage can read and turn into a banner (dovetails with #1).

## Acceptance criteria

- [ ] Homepage detects shipped-asset version != running-node version and shows a visible warning banner.
- [ ] After N consecutive port-conflict / exit-43 stand-downs the user sees a surfaced notification (macOS notification or homepage banner), not just a buried log line.
- [ ] The banner copy includes the concrete recovery command.

## References

- Parent: #3967 (service wrapper resilience — self-healing exit-43 + re-templating). This issue covers only the surfacing/observability half so it can land independently.
[...]

PR #4408, opened 2026-06-14
fix: self-heal stale-orphan port holder on wrapper exit 43

iduartgomez:
## Problem

The generated service wrapper's **exit-43 path stood down even against a stale orphan**,
causing a recurring production freeze (two independent hosts in ~2 weeks; one node frozen on
0.2.37 for 19 days while the on-disk binary was 0.2.49).

When the running binary cleanly detects an existing instance it returns exit 43; the wrapper
took the "polite deferral" branch and `exit 0`. launchd's `KeepAlive.SuccessfulExit=false`
treats `exit 0` as an intentional stop, so it **never respawns** the wrapper. The systemd unit
has the same trap (`RestartPreventExitStatus=43`). The catastrophic case: an orphaned
`freenet network` (PPID=1, detached from any wrapper) running an **old binary** holds the
service port — every new spawn cleanly detects it → exit 43 → wrapper stands down → the orphan
serves stale assets forever. The existing `pkill`-and-retry branch was unreachable because it
only fires on a non-43 crash.

The wrapper could not distinguish a **legitimate second instance** (defer is correct) from a
**stale orphan running an old binary** (must be killed).

## Solution

On exit 43, before standing down, identify the port holder and **kill it only when it is an
init-adopted orphan running a stale binary**:

- Enumerate holders via `pgrep -f -u "$(id -u)" "freenet network"`, resolve each holder's exe
  and query its `freenet --version`, compare to the on-disk binary's version.
- Kill (`SIGTERM`, poll up to ~12 s, then `SIGKILL`) **only** when `PPID==1` **and**
  (version mismatch **or** the holder's version is unreadable). A user-run `freenet network`
  (PPID≠1) or a current-version instance is always deferred to. An unreadable on-disk version
  falls back to the orphan heuristic only (never kills a healthy node).
- macOS: in the generated bash wrapper, with a `WRAPPER_CHILD_PID` self-exclusion and a
  `SIGTERM` trap forwarding to the child so drain doesn't depend solely on launchd
  process-group signaling.
- systemd: the same detect→compare→kill logic as an `ExecStartPre=-/bin/sh -c '...'` pre-flight
  on both the user and system units (every shell `$` doubled to `$$` so systemd's own Exec-line
  expansion doesn't eat it; the pre-flight excludes its own PID and PID 1).

**Scope:** this covers fresh installs / re-installs and the self-heal logic. The
re-templating-of-deployed-wrappers-on-`freenet update` path (issue acceptance #1/#2) and the
observability surfacing (#4382) remain outstanding — so this is **`Refs #3967`**, not a close.

## Testing

- **`test_exit43_heal_decision_behavior`** (macOS, behavioral): extracts the real heal helpers
  from the generated wrapper, PATH-shims `pgrep`/`ps`/`readlink`/`timeout`/`kill`, and asserts
  the kill/defer matrix — mismatched PPID==1 orphan (incl. a different exe path and a
  space-containing path) → killed; same-version → deferred; PPID≠1 → deferred; unreadable
  on-disk version → deferred; own child never targeted.
- **`test_systemd_execstartpre_heal_decision_behavior`** (Linux, behavioral): renders the unit,
  collapses `$$`→`$`, runs the pre-flight body under `sh` with shimmed tools and a `kill` that
  records targets; asserts an old-version, **different-path** PPID==1 orphan is killed (the
  exact case an earlier exe-path self-match guard wrongly skipped) and the defer cases hold.
- **`test_systemd_execstartpre_doubles_shell_dollars`**: renders both units, asserts every
  shell `$` is doubled and the collapsed body passes `sh -n`.
- Updated `test_macos_wrapper_exit43_self_heals_stale_orphan`; macOS wrapper passes `bash -n`.
- `cargo fmt`, `cargo clippy --bin freenet --tests -- -D warnings`, `cargo build -p freenet`
  clean.

Reviewed via a multi-perspective panel over two rounds; a must-fix (a systemd self-match guard
that anchored on the *current* binary path and so skipped the orphan it was meant to kill) was
caught and fixed with behavioral regression coverage. The external (codex) pass was unavailable
(auth/refresh-token error); the diverse multi-lens Claude panel was the documented fallback.

## Refs

Refs #3967 (the deployed-wrapper re-templating path — acceptance #1/#2 — and observability
(#4382) remain outstanding; this delivers the exit-43 self-heal for fresh installs).

[AI-assisted]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -4441,8 +4441,18 @@ mod tests {
         assert!(service_content.contains("StartLimitBurst=5"));
         assert!(service_content.contains("StartLimitIntervalSec=120"));
 
-        // Verify auto-update support via ExecStopPost
-        assert!(service_content.contains("ExecStopPost="));
+        // Verify auto-update support via ExecStopPost. The `$EXIT_STATUS` systemd
+        // sets in the ExecStopPost environment must reach /bin/sh as a literal
+        // `$EXIT_STATUS`, so it is written doubled (`$$EXIT_STATUS`) in the unit —
+        // systemd collapses `$$` -> `$`. A single-`$` revert silently breaks
+        // auto-update (systemd eats the bare var before sh sees it) with no other
+        // signal, so assert the doubled form, not just the directive's presence.
+        assert!(
+            service_content.contains("ExecStopPost=")
+                && service_content.contains("\"$$EXIT_STATUS\""),
+            "ExecStopPost must use the doubled $$EXIT_STATUS so systemd passes a \
+             literal $EXIT_STATUS to sh (a single-$ revert silently breaks auto-update)"
+        );
 
         // Verify exit code 42 is treated as success (doesn't count against StartLimitBurst)
         assert!(service_content.contains("SuccessExitStatus=42 43"));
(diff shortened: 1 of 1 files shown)


iduartgomez:
Note: the macOS-gated tests in this PR (`test_macos_wrapper_exit43_self_heals_stale_orphan`, `test_exit43_heal_decision_behavior`, etc.) are validated locally only — CI's `Unit & Integration` job runs on Linux and compiles out the `#[cfg(target_os = "macos")]` tests. The Linux `test_systemd_*` counterparts ARE run+green in CI (confirmed in the job log). Filed #4410 to add a macOS unit-test CI job so this surface is gated automatically going forward.

Issue #4404, opened 2026-06-14
Contracts not hosted by peers closest to their key cause GET dead-ends (placement gap)


sanity:
## Problem

A web contract took several attempts to load. Telemetry traced this to a **placement gap**, not a routing failure.

For a ~1 MB contract at ring location ~0.803:
- Greedy GET routing is healthy: 20/22 failed GETs reached a peer within 0.005 of the key, typically in ~2 hops.
- But the peer **closest** to the key did not host the contract, so the GET dead-ended there with NotFound, while the actual subscribed hosts (a fraction farther from the key) were never queried. Observed directly: the closest peer `43k8` (loc 0.802) forwarded GETs *backward* (to 0.745) because it had no closer connection, and the request bounced among non-hosting near-key peers to NotFound.

Root cause: **hosting follows traffic, not location.** A peer hosts a contract only if it personally served a PUT or a successful GET for it (`record_contract_access`, two production triggers). There is no mechanism that migrates or seeds a contract toward the peers closest to its key, so a peer that became closest *after* the contract went live, and never served a request for it, never hosts it. Freshness among close hosts is already handled by the neighbor-hosting gossip mesh (directly-connected co-hosts sync regardless of subscriptions), and the near-key cluster is densely interconnected (Kleinberg is nearby-biased), so the only gap is hosting.

Same failure family as #4222 (router window too small for subscribers) and #4223 (subscribers lacked the body), neither of which closes the placement gap.

## Proposed approach

Migrate contracts toward the peers closest to their key, so a key-routed GET lands on a host:

- A host sends a new fire-and-forget **`SubscribeHint`** to its **single closest connected neighbor**, but only when that neighbor is closer to the contract than itself **and** is not already hosting it. The neighbor then subscribes (directed at the hinting holder, since a greedy subscribe would route away from it), fetches the body, and hosts it — becoming the new frontier. This walks the contract one hop toward the key per step and terminates at the closest reachable peer. Only-the-single-closest-neighbor rule prevents over-replicating a large contract along every path.
- **Trigger:** on a peer beginning to host the contract (self-propagating cascade), and on gaining a new connection closer to a contract it hosts (un-stalls greedy local minima; complements the existing `contract_directed_connect`).
[...]

PR #4437, opened 2026-06-16
fix: deactivate placement migration after v0.2.73 UPDATE-degradation incident

sanity:
## Problem
v0.2.73 shipped #4404's placement migration **active** (`SUBSCRIBE_HINT_MIN_VERSION = (0,2,73)`, so a 0.2.73 build's version reaches the floor). As nova, vega, and technic updated to 0.2.73, UPDATE-operation failures climbed and did **not** recede:

```
technic:  0.2.72 = ~0/hr  ->  0.2.73 = 29, then 168/hr (04h pacing ~350/hr)
nova:     0.2.72 = 0/hr   ->  0.2.73 = 21, then 140/hr
```

Dominant error: `UPDATE relay (driver streaming): failed to assemble stream (broadcast) ... no fragments received within inactivity timeout`, plus WASM `exceeded the maximum allowed compute time` driving engine-store recoveries. Nodes stayed up (no panics/crash-loops), so it's degradation, not an outage. The broadcast-assembly error pre-exists on 0.2.72 but was rare; it spiked with the migration's activation. The migration adds directed-subscribes + per-GET/PUT `ConsiderContractMigration` emits that re-load the notification/broadcast backpressure surface that wedged both gateways in #4145/#4231.

## Solution
Raise `SUBSCRIBE_HINT_MIN_VERSION` to `(0, 3, 0)` so the migration ships **dormant** on all 0.2.x — a targeted, reversible mitigation. The migration code is untouched; re-lower the floor to the intended activation version once the cause is root-caused and validated (investigation in parallel). Ship as v0.2.74.

## Testing
With the floor above the build version the migration is off by default (its previous, well-tested state), so the existing sim suite passes unchanged. A const-value change to disable a feature has no meaningful regression test; the real verification is the live UPDATE-failure rate receding as 0.2.74 rolls out. `test-exempt` applied.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -626,7 +626,15 @@ pub(in crate::node) struct P2pConnManager {
 /// Simulation tests that want the cascade lower the floor per-node at runtime via
 /// `NodeConfig::subscribe_hint_floor_override` (set by
 /// `SimNetwork::enable_placement_migration`), which never touches production.
-const SUBSCRIBE_HINT_MIN_VERSION: (u8, u8, u16) = (0, 2, 73);
+// DEACTIVATED 2026-06-16: v0.2.73 shipped #4404's placement migration ACTIVE
+// (floor == 0.2.73). On the live network this correlated with a climbing
+// UPDATE-broadcast stream-assembly + WASM compute-time failure rate — the
+// migration's extra directed-subscribes and per-GET/PUT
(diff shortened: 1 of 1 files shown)


sanity:
Superseded by #4438, which folds this migration-deactivation change into the consolidated v0.2.74 hotfix (also adds the fd-limit self-raise, refresh_router transient-error resilience, and the batch transfer-rate NaN guard) to restore network performance after the v0.2.73 regression. [AI-assisted - Claude]

PR #4438, opened 2026-06-16
fix: restore network performance after v0.2.73 regression (migration off + fd limit + refresh_router resilience)

sanity:
## Problem

v0.2.73 introduced two regressions on the live network:

1. **Gateway crash-loop (fd exhaustion).** The WASM module cache holds up to ~1024 compiled contracts, each backed by a `memfd`. systemd's default `RLIMIT_NOFILE` *soft* limit is 1024, and freenet never raised it. On a busy gateway the module cache alone sits at the ceiling, so the next socket / AOF-segment open hits `EMFILE` ("No file descriptors available", os error 24). That error makes the monitored `refresh_router` task return `Err`, which the `BackgroundTaskMonitor` treats as **node-fatal** → systemd restart loop (every ~15–45 min on vega). Gateway-specific: technic (same 0.2.73 binary, no inbound-connection load) never crashed. **Not** migration-caused.

2. **Network-wide UPDATE-broadcast degradation.** The #4404 placement migration's directed-subscribes + per-GET/PUT `ConsiderContractMigration` emits amplified the #4145/#4231 broadcast-backpressure surface, driving "UPDATE relay broadcast stream assembly failed — no fragments received within inactivity timeout" up across the network.

Investigation also cleared two red herrings: no contract data loss / orphaning (migration only ever *adds* hosting, never drops the origin), and the NaN-in-router theory is defused (NaN lands only in the transfer-rate telemetry channel, not the failure estimator that gates peer selection).

## Approach (fix-forward — keeps all other 0.2.73 fixes; revert is last resort)

1. **Deactivate the placement migration** — raise `SUBSCRIBE_HINT_MIN_VERSION` to `(0, 3, 0)` (above every shipped 0.2.x), so the fail-closed gate never fires it in production. Removes the UPDATE-broadcast load amplifier. Migration code stays in place; re-enable by lowering the floor once the broadcast-backpressure load is fixed.
2. **freenet self-raises `RLIMIT_NOFILE` soft→hard at startup** — the permanent, universal fix for the gateway fd exhaustion (the live systemd drop-in only covers nova/vega; this protects every gateway and survives any restart path). Best-effort, non-fatal, unix-only.
3. **Harden `refresh_router`** — a transient fd/AOF read error no longer kills the node; it logs and retries on the next interval. Genuine panics are still caught by the monitor.
4. **Guard NaN in the batch transfer-rate path** (`Router::new`) — mirrors the existing per-op / `add_event` guards so a SUBSCRIBE-success (0/0) never enters the isotonic regression. Latent correctness fix / defense-in-depth.

## Testing

- `refresh_router_survives_transient_get_router_events_error` — flaky mock register fails then recovers; asserts the loop keeps polling instead of dying.
- `router_new_does_not_panic_on_subscribe_nan_transfer_rates` — 120 interleaved NaN-subscribe / finite-GET events at a shared location; `Router::new` must not panic.
- `raise_fd_limit_does_not_lower_soft_limit` (unix) — soft limit never decreases / never exceeds hard.
- Existing `version_gate` tests pass with `(0,3,0)`.
- `cargo fmt` / `cargo clippy -D warnings` / build clean.

## Incident context

Gateways were already mitigated live (`prlimit` + a systemd `LimitNOFILE=524288` drop-in on nova/vega); this PR makes that fix permanent in the binary and removes the migration load network-wide. Supersedes #4437 (which was the migration-deactivation change alone).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -887,6 +887,13 @@ mod tests {
     #[test]
     #[cfg(unix)]
     fn raise_fd_limit_does_not_lower_soft_limit() {
+        // Normalize `rlim_t` (u64 on linux-gnu, varies elsewhere) to u64 for
+        // arithmetic / comparison without tripping clippy on the redundant cast.
+        #[allow(clippy::unnecessary_cast)]
+        fn as_u64(v: libc::rlim_t) -> u64 {
+            v as u64
+        }
+
         let mut before = libc::rlimit {
             rlim_cur: 0,
             rlim_max: 0,
@@ -895,6 +902,37 @@ mod tests {
         // exclusively-borrowed out-param is sound; we check the return code.
         let rc = unsafe { libc::getrlimit(libc::RLIMIT_NOFILE, &mut before) };
         assert_eq!(rc, 0, "getrlimit should succeed");
+
+        let orig_soft = as_u64(before.rlim_cur);
+        let hard = as_u64(before.rlim_max);
+
+        // To exercise the ACTUAL raise branch (not the soft==hard no-op that is
+        // the common CI case), first LOWER our soft limit to a finite value
+        // strictly below hard, then call `raise_fd_limit()` and assert it raised
+        // back up to hard. Lowering one's OWN soft limit never requires
+        // privilege, but if the sandbox forbids even that we fall back to the
+        // weaker "no-decrease" assertion so the test is never flaky.
+        let lowered_to: Option<u64> = if hard != as_u64(libc::RLIM_INFINITY) && hard > 1 {
+            // Pick a finite target strictly below hard: min(orig_soft, hard/2),
+            // floored at 1 so we never request 0 fds.
+            let target = orig_soft.min(hard / 2).max(1);
+            // Guard: only meaningful if we can land strictly below hard.
+            if target < hard {
+                let lowered = libc::rlimit {
+                    rlim_cur: target as libc::rlim_t,
+                    rlim_max: before.rlim_max,
+                };
+                // SAFETY: setrlimit with a valid resource id and a fully
+                // initialized rlimit whose soft (target < hard) does not exceed
+                // the unchanged hard limit is sound.
+                let rc = unsafe { libc::setrlimit(libc::RLIMIT_NOFILE, &lowered) };
+                if rc == 0 { Some(target) } else { None }
+            } else {
+                None
+            }
+        } else {
+            None
+        };
 
         // Must not panic and must be safe to call.
         super::raise_fd_limit();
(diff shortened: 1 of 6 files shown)


sanity:
## Multi-model review (Full tier) — all findings addressed

This is a high-risk surface (wire-protocol gate, routing-task lifecycle, process startup), so it got a Full-tier review: four independent Claude perspectives (code-first, skeptical, testing, big-picture) **plus an external Codex pass**. Two findings were raised by **both** the Claude panel and Codex independently; both are now fixed in the fix-up commits.

### Blocking (now fixed)
- **Inbound `SubscribeHint` receive path was ungated** (`node.rs`). The floor bump to `(0,3,0)` only gates the *send* side, so a patched 0.2.74 node would still *act on* hints from not-yet-upgraded 0.2.73 peers, keeping the UPDATE-broadcast amplifier alive through the ~4h staggered rollout. **Fixed:** the receive handler now ignores inbound hints while the migration is parked (own-version vs the same floor, read via `subscribe_hint_floor_override().unwrap_or(...)` so the simulation cascade still works). Flagged independently by the skeptical/big-picture reviewers **and** Codex.
- **`docs/RELEASING.md` landmine** — it instructed operators to set `SUBSCRIBE_HINT_MIN_VERSION` to the release version, which would silently *re-enable* the migration on the next release. **Fixed:** the doc now marks the floor as intentionally parked at `(0,3,0)` until #4145 is resolved.
- **No test pinned the production constant** (the red `rule-review/warnings`). **Fixed:** added `placement_migration_deactivated_for_all_0_2_x` asserting no 0.2.x peer qualifies.

### Non-blocking (also fixed)
- **`mean_transfer_size` batch path** (`router.rs:295`) was missing the same `payload_size>0 && !transfer_time.is_zero()` guard as `add_event` — biases predicted transfer time after a SUBSCRIBE-heavy reload. Flagged by both code-first and Codex. **Fixed** (guard mirrored).
- **fd-limit test** no-op'd when `soft==hard`. **Fixed:** it now lowers the soft limit first, then asserts the raise restores it to hard.
- NaN test renamed to reflect it pins regression *finiteness* (no panic actually occurs in `Router::new`).

### Verified clean by Codex
`raise_fd_limit` (best-effort, sound unsafe, never lowers the limit), the `refresh_router` log+continue change, and the batch-path NaN guard.

### Follow-ups to be filed (deferred, not blocking)
- Re-enable the placement migration once load-safe (reopen #4404) — blocked on #4145.
- #4145 broadcast-backpressure root fix (the latent weakness the migration amplified; still present with the migration off).
- `refresh_router` persistent-failure escalation (a consecutive-failure counter so a permanently-broken AOF read isn't silent forever).
- Low-volume observability gauges (fd usage vs soft limit, module-cache occupancy, monitored-task health) so this class of incident is visible in telemetry, not just file logs.

[AI-assisted - Claude]

Issue #4233, opened 2026-05-24
Event-loop liveness simulation test under sustained UPDATE fan-out


sanity:
## Background

PR #4231 fixed the back-pressure deadlock that wedged both production gateways on 2026-05-24 (#4145). The fix has unit-level pins (behavioral + source-scrape) but **no simulation-level test** asserts event-loop liveness under sustained UPDATE fan-out — the exact load pattern that triggered the production wedge.

`.claude/rules/channel-safety.md` lists this bug class with **5+ production incidents to date**. The canonical regression guard against the next variant is a simulation test that exercises the wedge condition.

## Proposed test

In `crates/core/tests/simulation_integration.rs` (or new file), using the direct simulation runner (`run_simulation_direct`):

1. Spin up N peers (e.g. 8), one hosting a contract with M downstream subscribers (e.g. 40 — analogous to popular River contracts).
2. Drive concurrent UPDATEs to the hot contract for T seconds (e.g. 30 s of virtual time, multiple UPDATEs/s).
3. Assert during the run:
   - `Event loop stats iterations` keeps advancing on every peer for the full T seconds (no peer goes silent for >N stats-interval cycles).
   - `notification_channel_pending` stays bounded (e.g. <50% of capacity) on every peer.
   - No `ERROR ... notification channel full for too long; event loop may be stuck` log lines.
4. Post-run assertion: every peer's `active_connections == ring_connections` (no RING_TRANSPORT_DESYNC).

Test should fail clearly if reverted: comment out the `try_notify_node_event` switch in `runtime.rs::bridged_upsert_contract_state` (or revert the `StreamSend` timeout) and the test should observe the saturation log.

## Why this matters

The PR description for #4231 explicitly notes the absence of this test as a CI gap. The same bug class has shipped multiple times because no test exercises the cycle condition end-to-end. Without this, the next variant (e.g. a new `notify_node_event(...).await` site added in routine work) will only surface via production telemetry.

Tracking issue created on PR #4231 merge.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:38:22.685871926+00:00
Issue hash: `ef8c3d28`

## Value Assessment

[...]

Issue #4440, opened 2026-06-16
v0.2.73 placement-migration incident: re-enable migration + deferred follow-ups


sanity:
## Context

v0.2.73 shipped the #4404 placement migration (SubscribeHint nudges + directed-subscribe + `ConsiderContractMigration`). On the live network it amplified the latent #4145 broadcast-backpressure weakness, driving a network-wide UPDATE-broadcast stream-assembly degradation (nova/vega ~1500–2300 failures/hr, technic ~500–950/hr, vs a ~0 baseline on 0.2.72), plus a gateway crash-loop from fd exhaustion (WASM module cache ≈1024 memfds colliding with systemd's 1024 `RLIMIT_NOFILE` soft limit).

**v0.2.74 (PR #4438)** stabilized the network by:
- Deactivating the placement migration (parked `SUBSCRIBE_HINT_MIN_VERSION` at `(0,3,0)`, send + receive).
- freenet self-raising `RLIMIT_NOFILE` soft→hard at startup (universal fd-exhaustion fix).
- Hardening `refresh_router` against transient fd/AOF errors.
- Guarding NaN in the batch transfer-rate path.

This issue tracks the deferred work — chiefly **getting 0.2.73's placement migration working *properly*** (it is currently *disabled*, not fixed) and restoring the GET-dead-end fix it provided.

## Checklist

- [ ] **Fix #4145** (event loop stuck under load — notification channel saturation). This is the root broadcast-backpressure weakness the migration amplified; it remains latent with the migration off and is the **blocker** for safely re-enabling the migration. Any future fan-out spike can re-trigger the degradation until this is fixed.
- [ ] **Re-enable the placement migration** once #4145 is resolved: lower `SUBSCRIBE_HINT_MIN_VERSION` from `(0,3,0)` to the first load-safe release. This restores the #4404 GET-dead-end fix (reopened). Update `docs/RELEASING.md` (which is currently marked "parked/deactivated") at the same time. Send + receive gate re-activate together.
- [ ] **`refresh_router` persistent-failure escalation.** v0.2.74 made a transient fd/AOF error non-fatal (log + retry), which correctly stopped the crash-loop — but a *persistent* failure now degrades routing silently forever. Add a consecutive-failure counter / metric (NOT node-fatal) so operators can distinguish a transient blip from a router that has been cold for hours.
[...]

Issue #4235, opened 2026-05-24
broadcast_queue conflates 'permit release' with 'delivery success' via completion_tx


sanity:
## Background

The broadcast queue's per-peer dispatch uses a `tokio::sync::oneshot::Sender<()>` (`completion_tx`) to coordinate semaphore-permit release. Today the queue treats *any* signal on this oneshot as "successfully delivered" — including the cases where the StreamSend dispatch dropped the message (channel closed, channel full / timeout, no connection) and signaled completion only to release the permit.

Caller code (`broadcast_queue.rs:444-518`):

```rust
match tokio::time::timeout(STREAM_COMPLETION_TIMEOUT, completion_rx).await {
    Ok(Ok(())) => "completed successfully" + refresh_peer_interest + update_peer_summary,
    Ok(Err(_)) => "completion channel dropped (task ended)",
    Err(_)     => "completion timed out",
}
```

After the `Ok(Ok(()))` branch, `send_ok = send_result.is_ok()` (true if `bridge.send` enqueued the StreamSend dispatch — not whether the peer actually received it), and downstream:

- `refresh_peer_interest` refreshes the peer's interest TTL based on a send that may not have happened
- `update_peer_summary` caches a summary for the peer — defeating the next summary-mismatch round that should have detected the drop

### Pre-existing vs PR #4231

This is a pre-existing semantic conflation. **PR #4231 extends the trigger** (the `StreamSend` timeout path now fires `completion_tx` to release the permit immediately instead of waiting `STREAM_COMPLETION_TIMEOUT` = 120 s), making the silent-loss surface larger but not introducing it (the channel-closed path already had this behavior).

Codex external reviewer flagged this as P2 on PR #4231. Big-picture and skeptical reviewers concurred it's worth fixing but agreed it's out-of-scope for that PR.

## Proposed fix

Change the oneshot's payload from `()` to a delivery-outcome enum:

```rust
enum DeliveryOutcome {
    Delivered,
    Dropped { reason: DropReason },
}

enum DropReason { Timeout, ChannelClosed, NoConnection }
```

`broadcast_queue.rs` then:

- On `Delivered` → refresh interest + update summary cache (as today)
- On `Dropped { ... }` → release permit but do NOT update summary cache; ideally trigger an immediate retry path or at least a counter

[...]

PR #4442, opened 2026-06-16
fix(broadcast): cache peer summary on delivered full-state broadcasts to end the full-state storm (#4145)

sanity:
## Problem

The remaining half of #4145. #4231 fixed the back-pressure *deadlock* (the event loop is provably not the bottleneck — our 2026-06-16 incident showed `slow_events=0`, `notification_channel_pending=0`, `peer_congested=0` throughout). The remaining degradation under sustained UPDATE fan-out is a **full-state broadcast storm**: popular contracts shipped their *entire state* (streamed) to every subscriber on every update, saturating gateway uplinks (47 distinct congested destinations, 113KB–42MB stalled transfers, GET/PUT relays collateral-damaged). Pre-migration (small fan-out) this was ~0; the placement migration made it acute by bloating subscription trees.

## Root cause — a chicken-and-egg in the summary cache (PR #2763)

Broadcasts send a small **delta** when the gateway has the target peer's cached *summary* to diff against, and otherwise fall back to **full state** (streamed if >64KB). But the peer's summary was cached **only after a delta send** (`if sent_delta { update_peer_summary(our_summary) }`), at both broadcast cache sites (`broadcast_queue.rs`, `p2p_protoc.rs`).

That's a trap: you need the peer's summary to send a delta, but the summary is only cached after a delta. So **any peer that ever starts on full state — every new subscriber, or anyone whose summary was cleared on a failure — is stuck sending full state forever**, never bootstrapping into deltas. The migration flooded the gateways with summary-less subscribers → permanent full-state storm.

PR #2763 gated on `sent_delta` because, at the time, a streamed full-state "success" did not reliably mean *real delivery* — caching the summary then risked telling the gateway the peer had state it didn't, making the next delta unappliable (divergence). **#4235 since added a real-delivery signal** (`BroadcastDeliveryOutcome::Delivered`), removing that risk on the streaming path.

## Approach

Cache the peer's summary on any **delivered** broadcast (delta *or* full state), not just deltas:
- **Streaming (full-state) path** caches via `record_delivery_to_interest`, reached only on `streaming_completion_delivered == Ok(Ok(Delivered))` — a verified real delivery (#4235). Safe.
- **Non-streaming path** (`p2p_protoc.rs`) caches inside the send-success arm only — same assumption the delta path already makes; the summary-mismatch resync corrects any rare wrong cache. (Also moved the cache block *inside* the success `else` arm — it previously ran regardless of send success/failure.)

After a real delivery the peer demonstrably has `our_summary` either way, so the next update is a small delta. Each peer now gets exactly **one** bootstrap full-state, then deltas — killing the storm at its source. No change to the delta-vs-fullstate decision, streaming threshold, concurrency limits, or wire format. The existing large-stream concurrency limit (2) already bounds the one-time bootstrap burst.

## Testing

- `full_state_delivery_caches_summary_so_next_broadcast_is_delta` — drives the real `Delivered` gate with `sent_delta=false`, asserts the summary is cached and the next broadcast is a delta. **Fails on the old `if sent_delta` gate.**
- `dropped_full_state_stream_does_not_cache_summary` — negative test: a dropped/timed-out full-state stream must NOT cache (preserves #2763/#4235's divergence guard).
- `test_sustained_update_fanout_no_full_state_storm` (the #4233 sim) — sustained UPDATE fan-out; asserts liveness, delta-send dominance / full-state collapse, and convergence. **Passes with the fix (78 delta / 47 full), fails on revert (40 delta / 85 full).**

`cargo fmt` + `cargo clippy -D warnings` clean.

Addresses (does not close) #4233 — the full production-path liveness test it specifies (≈40 subscribers, channel-pending bounded) remains open; this PR adds a summary-cache regression test on the sim path. Substantially addresses the remaining #4145 (re-enabling the placement migration, #4440 / #4404, is gated on this landing + validating under real fan-out).

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/channel-safety.md
+++ b/.claude/rules/channel-safety.md
@@ -61,6 +61,24 @@ stalling the entire event loop. This has caused 5+ production deadlocks:
   because it carries a one-shot transition — see its inline
   `DELIBERATELY blocking` comment for the rationale.
 
+  The same #4145/#4231 investigation surfaced a SECOND, separate
+  degradation under the same fan-out — not a deadlock but a full-state
+  broadcast STORM (#4442). A peer's summary was cached only after a
+  *delta* send (PR #2763's `if sent_delta`), but a delta needs the
+  peer's cached summary to compute against — a chicken-and-egg that
+  trapped every summary-less subscriber (any new subscriber, or one
+  whose summary was cleared) on full state forever. Under sustained
+  fan-out that meant every update re-sent full state to every
+  subscriber, swamping the event loop. Fixed (#4442) by caching the
+  summary on any *delivered* broadcast (delta OR full state): #4235's
+  real-delivery signal (`BroadcastDeliveryOutcome::Delivered`) makes
+  caching on a full-state delivery safe, so the next broadcast to that
+  peer collapses to a small delta. A wrongly-cached summary (e.g. a
+  lost stream tail — `Delivered` is sender-side completion, not a
+  receiver ack) is corrected by the periodic InterestSync summary
+  exchange and by the delta-apply-failure → ResyncRequest path that
+  clears the sender's cached summary.
+
 ## Exception: same-runtime internal consumers
 
 `.send().await` on a bounded channel is acceptable when **all** of the
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -4285,6 +4285,60 @@ mod tests {
         }
     }
 
+    /// Source-level pin for the #4145 non-streaming caching safety net.
+    ///
+    /// The summary-cache fix (#4145) caches a peer's summary on any
+    /// *delivered* broadcast. That is only safe because the
+    /// `ResyncRequest` handler clears the SENDER's cached summary for the
+    /// peer when a downstream delta fails to apply — otherwise a wrongly
+    /// cached summary would trap the pair sending unappliable deltas.
+    /// If a refactor drops the `update_peer_summary(.., None)` clear from
+    /// this handler, the #4145 caching loses its corrective backstop and
+    /// the behavioural sim test would still pass. Pin it at the source
+    /// level so the omission fails CI.
+    mod resync_request_clears_sender_summary {
+        const SOURCE: &str = include_str!("node.rs");
+
+        /// The body of the `InterestMessage::ResyncRequest` match arm,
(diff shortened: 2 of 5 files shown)


sanity:
## Multi-model review (Full tier) — all findings addressed

Four independent Claude perspectives (code-first, skeptical, testing, big-picture) + external **Codex**. No blocking *correctness* bug in the production path — all reviewers independently confirmed the production `broadcast_queue.rs` cache is gated on the #4235 real-`Delivered` signal, so the #2763 divergence guard is preserved. Findings, all fixed in the fix-up commits:

- **[Codex, blocking]** the `#[cfg(simulation_tests)]` inline broadcast path returned only the metadata-send result, so a *failed* full-state stream still cached the summary (the #2763/#4235 hazard, sim-only). **Fixed** — `send_result` now propagates the `send_stream` error, so the cache arm is unreachable on stream failure.
- **[skeptical/code-first]** the `Delivered` signal is *sender-side* stream completion, not a receiver ack, so a lost stream-tail can leave a bounded (~5 min) divergence window. **Comment softened** to state this and name the two backstops (periodic InterestSync overwrite + delta-fail→ResyncRequest sender-cache-clear).
- **[all]** test comments cited wrong reverted numbers (21/104) and a false "exactly ONE bootstrap full-state per subscriber" claim. **Corrected** to the measured 40 delta / 85 full on revert (78/47 with the fix); reworded to "full-state sends collapse below delta sends," not one-per-subscriber.
- **[testing]** liveness assertion was a weak proxy. **Added** a `pending_op_high_water_mark` backlog bound (documented as a saturation guard; the `delta>full` assertion remains the revert discriminator).
- **[big-picture]** **`.claude/rules/channel-safety.md`** updated with the storm-half fix; **`Closes #4233` downgraded to "addresses"** (its full production-path liveness test remains owed). Added a ResyncRequest sender-cache-clear pin test.

### Important scope caveat
This fixes the **steady-state** storm; the **bootstrap burst** (one full-state per new summary-less subscriber) is *not* capped (deferred — maps to #3798 Gap 1). And the sim test rides the sim path + a clean star (2 subscribers), so **nothing here validates the fix at real 40–80-target fan-out**. That validation is the actual safety gate for re-enabling the placement migration (#4440/#4404) and is tracked there — **merging #4442 does NOT by itself unblock the migration.**

[AI-assisted - Claude]

PR #4443, opened 2026-06-16
ci: add nightly cargo binstall smoke test against latest releases

### LINKED ISSUE #3998: Add nightly cargo binstall smoke test against latest releases
Spun out of #3995 review feedback (testing-reviewer): the manifest-level regression tests in #3997 lock in the structural invariants (URL embeds the right tag, Windows override exists, `bin-dir` is correct, sed expressions stay equivalent), but they cannot catch the *user-visible* failure mode — a 404 on github.com because the release upload silently changed naming, the release was never published from draft, or the assets failed to attach.

## Proposal

A scheduled GitHub Actions workflow that runs nightly (or after every release publish) and:

1. Resolves the latest published version of `freenet` and `fdev` from crates.io.
2. Runs `cargo binstall --no-confirm freenet@<version>` and `cargo binstall --no-confirm fdev@<version>`.
3. Asserts both binaries land in the expected location and respond to `--version`.
4. Optionally repeat on a Windows runner to exercise the `.zip` override path.

If any step fails, the workflow opens an issue / pings #freenet-dev. This catches:

- Release upload pipeline broke (e.g. cross-compile job timed out and didn't attach assets)
- Asset naming drifted (e.g. a target rename in cross-compile.yml)
- Release stayed in draft state and was never published
- The two metadata-block sed expressions in release.sh / release.yml silently produced different output from each other (today the in-repo test catches this, but only on the workspace fixture)

## Why not block merge of #3995 on this

The structural tests in #3997 give 90% of the value with zero CI cost. The e2e smoke test is a separate workflow concern and adds CI minutes; better to land it as its own focused PR.

[AI-assisted - Claude]

sanity:
## Problem

The manifest-level regression tests added in #3997 lock in the *structural* invariants of the binstall metadata in `crates/core/Cargo.toml` and `crates/fdev/Cargo.toml` (the URL embeds the right tag, the Windows override exists, `bin-dir` is correct, the two metadata-block sed expressions stay equivalent). But those tests run against the in-repo workspace fixture and cannot catch the *user-visible* failure mode — a real 404 on github.com when someone runs `cargo binstall freenet`:

- release upload pipeline broke (e.g. a cross-compile job timed out and didn't attach assets)
- asset naming drifted (e.g. a target rename in `cross-compile.yml`)
- the release stayed in draft and was never published
- the `pkg-url` literal `vX.Y.Z` segment in `fdev`'s metadata drifted from the actual release tag (fdev's crate version `0.3.x` is independent of the `v0.2.x` release tag, so its URL embeds the tag literally — exactly the fragile spot #3995 was about)

## Approach

A nightly scheduled workflow (`workflow_dispatch` for manual runs too) that:

1. Resolves the latest *published* `freenet` and `fdev` versions from the crates.io API (`max_stable_version`, so prereleases are ignored).
2. Runs `cargo binstall --no-confirm freenet@<v>` / `fdev@<v>`, which downloads the real GitHub release assets named by the published crate's binstall metadata.
3. Asserts both binaries respond to `--version`.

Matrix legs:
- **Linux** pins `--targets x86_64-unknown-linux-musl` — the release builds Linux assets for the musl target (see `cross-compile.yml`), so the default gnu host target would 404. This is the one correctness subtlety worth calling out in review.
- **Windows** (`x86_64-pc-windows-msvc`) exercises the `.zip` / `bin-dir = freenet.exe` override path.

On failure it notifies the public River room and the private freenet-dev room, reusing the exact failure-notify pattern from `simulation-nightly.yml` (best-effort, `continue-on-error`, gated on `binstall-smoke-test.result == 'failure'`).

Why a separate workflow and not blocking on it: per the issue, the structural tests give ~90% of the value at zero CI cost; this e2e check adds CI minutes so it lands as its own focused, nightly-only PR.

## Testing

This is a CI-config change (no Rust touched). Validated by:

- `actionlint` — clean (exit 0).
- **Live end-to-end dry-run locally**: resolved `freenet=0.2.74` / `fdev=0.3.236` from crates.io, ran the exact `cargo binstall --targets x86_64-unknown-linux-musl` commands into a throwaway root, and confirmed `freenet --version` / `fdev --version` both work.
- HEAD-checked all four expected release assets (freenet/fdev × musl tar.gz / msvc zip) for v0.2.74 → all HTTP 200.
- Confirmed `jq -e '.crate.max_stable_version'` exits non-zero on a malformed/missing-field response, so a bad crates.io reply fails the step loudly instead of installing `crate@null`. `jq` is preinstalled on all GitHub-hosted runners (portable across the bash-on-Windows matrix leg).

## Fixes

Closes #3998

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/binstall-smoke-test.yml
+++ b/.github/workflows/binstall-smoke-test.yml
@@ -1,0 +1,140 @@
+name: Binstall Smoke Test (Nightly)
+
+# End-to-end check that the LATEST published `freenet` and `fdev` crates can
+# actually be installed via `cargo binstall`. The in-repo manifest tests
+# (#3997) lock in the structural invariants of the binstall metadata, but they
+# cannot catch a user-visible 404: release assets that failed to attach, a
+# release stuck in draft, or asset naming that drifted out of sync with the
+# `pkg-url` in Cargo.toml. This downloads the real GitHub release assets and
+# runs the binaries, so any of those break this workflow. See issue #3998.
+
+on:
+  schedule:
+    # 5 AM UTC daily — staggered after simulation-nightly (3 AM UTC).
+    - cron: "0 5 * * *"
+  workflow_dispatch:
+
+# Don't pile up concurrent runs.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  binstall-smoke-test:
+    name: Binstall ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Linux release assets are built for the musl target (see
+          # cross-compile.yml). The default gnu host target would 404, so pin
+          # the target binstall resolves to match the published asset name.
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+          # Windows exercises the `.zip` / `bin-dir = *.exe` override path.
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+    steps:
+      - name: Install cargo-binstall
+        # `@main` is the upstream-recommended ref for this action: the repo's
+        # `vX.Y.Z` tags are cargo-binstall *crate* releases, not pinned action
+        # versions. Don't "upgrade" this to a vX tag — that pins to a crate
+        # release, not the maintained action.
+        uses: cargo-bins/cargo-binstall@main
+
+      - name: Resolve latest published versions from crates.io
+        id: versions
+        shell: bash
+        run: |
+          set -euo pipefail
+          # crates.io requires a descriptive User-Agent or it 403s. `jq` is
+          # preinstalled on every GitHub-hosted runner (incl. windows-latest
+          # under bash), so it's portable across the matrix.
+          UA="freenet-core binstall-smoke-test (https://github.com/freenet/freenet-core)"
+          for crate in freenet fdev; do
+            version=$(curl -sSf -H "User-Agent: $UA" \
+              "https://crates.io/api/v1/crates/${crate}" \
+              | jq -er '.crate.max_stable_version')
+            echo "Latest ${crate}: ${version}"
+            echo "${crate}=${version}" >> "$GITHUB_OUTPUT"
+          done
+
+      - name: Install freenet via cargo binstall
+        shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary (Light tier — CI/tooling config)

This is a CI-config-only change (one new nightly workflow, no Rust). Reviewed with an external Codex pass plus four Claude lenses (code-first, testing, skeptical, big-picture), per the multi-model review rule.

### External pass — Codex (`codex review --base origin/main`)

**Round 1 found one P2 (now fixed):** with cargo-binstall's default strategies (`crate-meta-data,quick-install,compile`), a missing/404ing GitHub release asset could silently fall through to the quickinstall mirror or a from-source compile — so the smoke test would go **green even when the exact failure it exists to detect was present** (broken release upload, drifted asset naming, draft release). This is a real false-green and was the most valuable finding.

Fix (commit `aa7e8aa8`): pin `--strategies crate-meta-data` on both install commands so only the GitHub release asset named by the crate's binstall metadata is ever used.

**Round 2 (re-run on the fixed content):** clean — "I did not find any actionable correctness issues introduced by this patch."

### Claude lenses
- **code-first:** intent matches impl — resolve latest published versions, install from the GitHub asset only, assert `--version`. No discrepancy.
- **testing:** the workflow *is* the test; validated end-to-end locally (see Testing below). No coverage gap.
- **skeptical:** `set -euo pipefail` + `jq -er` fail loudly on a bad crates.io reply (verified `jq -e` exits non-zero on a null field, so `crate@null` can't be installed); notify jobs are gated on `result == 'failure'` and `continue-on-error` so they never red-fail the run on their own; `~/.cargo/bin` is on PATH on these runners (matches existing nextest/riverctl workflows).
- **big-picture:** matches issue #3998 exactly; single logical change; no removed tests, no stale docs.

### Testing performed
- `actionlint` clean (exit 0).
- Live end-to-end: resolved `freenet=0.2.74` / `fdev=0.3.236` from crates.io, ran the exact `cargo binstall --strategies crate-meta-data --targets x86_64-unknown-linux-musl` commands into a throwaway root, and confirmed both `--version` calls work.
- Failure path: requested a target with no published asset → binstall exits 94 ("Fallback to cargo-install is disabled"), nothing installed (no source compile). Confirms the strategy pin actually catches a broken asset.
- HEAD-checked all four expected v0.2.74 release assets (freenet/fdev × musl tar.gz / msvc zip) → HTTP 200.

All findings addressed. Holding merge until CI is green on the current HEAD.

[AI-assisted - Claude]

PR #4446, opened 2026-06-16
fix: route standalone LocalSubscribeComplete through consistent terminal cleanup (#3099)

### LINKED ISSUE #3099: Investigate: some transactions complete without emitting TransactionCompleted/TransactionTimedOut events
## Context

Discovered during investigation of #2981 (pending_op_results memory leak). The fix adds cleanup on `TransactionCompleted`/`TransactionTimedOut` events plus a periodic safety-net sweep for entries where the sender is closed.

The reviewer asked: **why do some transactions complete without emitting these events?**

## Root Cause Analysis

Several code paths in `handle_op_result()` (`crates/core/src/operations/mod.rs`) reach terminal states without emitting `TransactionCompleted`:

### 1. Local state queueing (UpdateBroadcasting)
When `OperationResult` has `return_msg: None` and `state: Some(updated_state)`, the state is pushed to OpManager but no completion event is emitted (line ~258-266).

### 2. StatePushed error path
`Err(OpError::StatePushed)` returns `Ok(None)` — the operation continues locally without any event (line ~113-117).

### 3. Duplicate message for completed op
`Err(OpError::OpNotPresent(tx))` returns `Ok(None)` — the original transaction's callback was already inserted but if the original completion path didn't emit an event, the entry leaks (line ~118-127).

### 4. Root operation waiting for sub-operations
When a finalized operation has sub-operations that haven't completed, it's stored in `root_ops_awaiting_sub_ops` and returns `Ok(None)` — no event emitted until all children complete (line ~180-213).

### 5. Callback invocation guard
In `process_message_decoupled()` (`crates/core/src/node/mod.rs`), the callback sender is only invoked when `is_operation_completed(&op_result)` returns true. The `pending_op_results` entry is inserted regardless, but the callback may never fire for the paths above.

## Impact

The periodic safety-net cleanup in #2981 handles all these cases by checking `sender.is_closed()`. The event-driven cleanup handles the common case. Together they prevent the leak.

## Recommendation

Consider auditing whether `TransactionCompleted` should be emitted in more of these terminal paths, particularly:
- When `StatePushed` operations actually finish their local processing
- When root operations' sub-operations all complete (verify the parent event fires)

This would make the system more consistent, though the safety-net cleanup makes it non-critical.

## Related
- #2981 — fix: clean up pending_op_results to prevent unbounded memory growth
- #2941 — referenced in cleanup comments (Wasmer memory, different root cause)

sanity:
## Problem

#3099 asked why some transactions reach a terminal state without emitting `TransactionCompleted`/`TransactionTimedOut`. @iduartgomez re-scoped it (the three original paths were eliminated by the #1454 op refactor) to the one remaining path: the standalone-parent branch of the `LocalSubscribeComplete` arm in `crates/core/src/node/network_bridge/p2p_protoc.rs` publishes its result directly via `op_manager.result_router_tx` instead of `OpManager::send_client_result`, so it never emits `TransactionCompleted` and never reaches the handler that prunes `pending_op_results`.

## Investigation finding (please read before merging)

**The feared `pending_op_results` leak does NOT actually occur on this path.** While implementing the fix I traced the path end-to-end and found:

- The tx carried by `LocalSubscribeComplete` for the standalone parent is the **client/parent tx**, which the SUBSCRIBE driver **never inserts into `pending_op_results`** — it only registers per-attempt `attempt_tx` slots, each released via `release_pending_op_slot`. The module docs say so explicitly: *"`client_tx` ... never passed to `send_and_await`"* (`subscribe/op_ctx_task.rs`).
- `live_tx_tracker` and `under_progress` are already cleaned by the `op_manager.completed(tx)` that `complete_local_subscription` runs (`subscribe.rs`).
- `tx_to_client` / `client_waiting_transaction` are already cleaned by the existing arm.
- The existing `test_pending_op_results_bounded` (#3100) already asserts a **zero** leak at shutdown and passes on `main`, corroborating that there's no active leak on this path.

So the prioritizer/triage premise (an active transient `pending_op_results` leak per standalone subscribe) is not supported by the current code.

## What this PR does

Given that, this is **invariant hardening, not an active leak fix**. It makes the one terminal path that bypasses `send_client_result` participate in the same `pending_op_results` cleanup every other terminal path gets, via a new idempotent `EventListenerState::reclaim_pending_op_result` helper (which also records the matching `pending_op_remove` metric so #3100's accounting stays balanced). The call is a no-op on every currently-reachable case; it exists so a future refactor that routes a `pending_op_results`-bearing tx through this arm frees the entry immediately rather than leaving it for the 60s sweep, and it adds a pin test so that wiring can't silently regress. The redundant `live_tx_tracker` removal is intentionally **not** added (already done by `op_manager.completed`).

**Maintainer decision needed:** is this forward-looking hardening worth carrying, or should #3099 simply be closed as "no active leak, the #1454 refactor already fixed the originally-reported paths"? I did not auto-merge for this reason.

## Testing

- `reclaim_pending_op_result_drops_entry_and_records_remove` — behavioral unit test: removes a resident entry, records exactly one `pending_op_remove`, is idempotent on a second call, and a subsequent `Drop` does not re-count.
- `local_subscribe_complete_reclaims_pending_op_result` — source-scrape pin test asserting the arm calls the helper (matches the existing `transaction_orphaned_handler_*` pin convention). **Verified to FAIL on the pre-fix arm and PASS with the fix.**

`cargo fmt` + `cargo clippy -p freenet --lib --features testing -- -D warnings` clean. External review: `codex review --base origin/main` found no regressions introduced by the patch.

## Fixes

Closes #3099

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -2633,6 +2633,37 @@ impl P2pConnManager {
                                         );
                                     }
                                 }
+
+                                // Invariant hardening (#3099): this is the one
+                                // terminal path that publishes its result directly
+                                // via `result_router_tx` instead of routing through
+                                // `OpManager::send_client_result`, so it never emits
+                                // `NodeEvent::TransactionCompleted` and never reaches
+                                // the `TransactionCompleted` handler that prunes
+                                // `pending_op_results`.
+                                //
+                                // Today `tx` here is always the client/parent tx,
+                                // which is NEVER inserted into `pending_op_results`:
+                                // the SUBSCRIBE driver only registers per-attempt
+                                // `attempt_tx` slots (see subscribe/op_ctx_task.rs
+                                // module docs — "`client_tx` ... never passed to
+                                // `send_and_await`"), and each `attempt_tx` slot is
+                                // released via `release_pending_op_slot`. So this call
+                                // is a no-op on every currently-reachable case and is
+                                // purely defensive. We keep it — pinned by
+                                // `local_subscribe_complete_reclaims_pending_op_result`
+                                // — so that if a future refactor ever routes a
+                                // `pending_op_results`-bearing tx through this arm, the
+                                // entry is freed immediately instead of lingering until
+                                // the 60s periodic sweep. `reclaim_pending_op_result`
+                                // is idempotent and records the matching
+                                // `pending_op_remove` metric so the accounting behind
+                                // `test_pending_op_results_bounded` (#3100) stays
+                                // balanced. `live_tx_tracker`/`under_progress` are
+                                // already cleaned by the `op_manager.completed(tx)`
+                                // that `complete_local_subscription` ran, so they are
+                                // not touched again here.
+                                state.reclaim_pending_op_result(tx);
                             }
                             NodeEvent::BroadcastHostingUpdate { message } => {
                                 ctx.handle_hosting_broadcast(message).await;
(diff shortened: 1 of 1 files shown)


sanity:
Closing rather than merging. The investigation behind this PR (corroborated by an independent Codex pass) found there is **no active `pending_op_results` leak** on the `LocalSubscribeComplete` path:

- The parent/client `tx` reaching this arm is never inserted into `pending_op_results` — the SUBSCRIBE driver only registers per-attempt `attempt_tx` slots, each released via `release_pending_op_slot`.
- `live_tx_tracker` / `under_progress` are already cleaned by the `op_manager.completed(tx)` that `complete_local_subscription` runs.
- The inserts/removes balance is already asserted by `test_pending_op_results_bounded` (#3100), green on main.

So `reclaim_pending_op_result(tx)` here is a provable no-op on every currently-reachable case — purely defensive against a hypothetical future refactor, plus a brittle source-scrape pin test locking in that no-op. On balance that is speculative hardening rather than a needed improvement, so closing instead of carrying dead-but-defensive code. If a future refactor ever routes a real `pending_op_results`-bearing tx through this arm, that is the point to add cleanup — and #3100 bound test would flag the regression. [AI-assisted - Claude]

PR #4447, opened 2026-06-16
fix: skip outbound SUBSCRIBE renewals for banned contracts

### LINKED ISSUE #4373: Banned contract keeps emitting outbound SUBSCRIBE renewals while node holds an active subscription
## Problem

A contract that is banned **while this node already holds an active subscription** is not torn down by the ban path, so the node keeps emitting outbound SUBSCRIBE renewal wire requests for it until the ban TTL expires.

`Ring::apply_ban_decisions` (`crates/core/src/ring.rs` ~L800) only calls `ban_list.ban()` / `ban_list.unban()` on a `BanTriggered` / `BanLifted` decision. It does **not** clear the contract's entry from `active_subscriptions` / local interest, so the contract remains in the renewal set.

The subscription-renewal loop (`crates/core/src/ring.rs` ~L1255, spawning `crate::operations::subscribe::run_renewal_subscribe` at ~L1323) iterates `contracts_needing_renewal` and, for each, spawns a renewal driver that emits an outbound SUBSCRIBE request using the same machinery as a client-initiated SUBSCRIBE. This renewal path does **not** pass through `operations::reject_if_contract_banned` — that egress gate (#4300) lives only at the four `start_client_*` originator entry points, not at the connection-maintenance renewal scheduler. So a banned-but-still-subscribed contract keeps sending outbound SUBSCRIBE renewals on every maintenance cycle until the ban's TTL lifts.

## Scope / severity

This is genuine outbound network egress for a banned contract, so it is adjacent to #4300. It is **out of #4300's named scope** for two reasons:

- It does **not** push contract state. The state-egress paths (delegate-driven UPDATE fan-out, NeighborHosting overlap sync, InterestSync::Summaries stale repair, and the four client originator entry points) are all gated by #4299 / #4371. A SUBSCRIBE renewal only re-registers interest; it carries no state.
- It is **time-bounded**: it self-resolves when the ban TTL expires (default 1 hour) and the next `BanLifted` is processed, after which the subscription is legitimate again — or the subscription itself expires on its own lease.

So this is a lower-severity, bounded egress leak rather than a hole in the state-egress block, which is why it's a follow-up rather than part of #4371.

## Proposed fix

Either:
1. Gate the renewal loop on the ban list — skip `run_renewal_subscribe` for any contract currently in `ContractBanList` (e.g. add an `is_banned` check alongside the existing `can_request_subscription` spam check at ~L1277), or
2. Actively unsubscribe / drop interest for a contract when it is banned — have `apply_ban_decisions` clear the contract from `active_subscriptions` / local interest (and ideally send an `Unsubscribe`) on `BanTriggered`, so it leaves the renewal set entirely.

Option 2 is the more thorough fix (it also stops the node from holding a subscription it has decided is harmful); option 1 is the smaller, more contained change.

References #4300 and #4371.

[AI-assisted - Claude]

sanity:
## Problem

A contract that is banned **while this node already holds an active subscription** keeps emitting outbound SUBSCRIBE renewal wire requests for it until the ban TTL expires.

The subscription-renewal scheduler in `Ring::recover_orphaned_subscriptions` (`crates/core/src/ring.rs`) iterates `contracts_needing_renewal()` and spawns `operations::subscribe::run_renewal_subscribe` for each contract that passes the `can_request_subscription` spam check. That driver emits an outbound SUBSCRIBE using the same machinery as a client-initiated request — but unlike the four `start_client_*` originator entry points, the renewal scheduler never passes through `operations::reject_if_contract_banned` (the #4300 egress gate). So a banned-but-still-subscribed contract keeps sending outbound SUBSCRIBE renewals on every maintenance cycle until the ban lifts.

This is genuine outbound network egress for a banned contract (adjacent to #4300). It is lower-severity than the state-egress block because:
- It pushes **no contract state** — a SUBSCRIBE renewal only re-registers interest.
- It is **time-bounded** — it self-resolves when the ban TTL expires (default 1 hour) and the next `BanLifted` is processed, or when the subscription lease expires on its own.

## Solution

Implements the issue's **Option 1** (the smaller, more contained change): gate the renewal loop on the ban list. Before the existing `can_request_subscription` spam check, skip any contract for which `ring.contract_ban_list.is_banned(contract.id())` is true.

The gate is placed **before** `can_request_subscription` (so a banned contract is skipped regardless of its spam-backoff state) and before `mark_subscription_pending` / the `run_renewal_subscribe` spawn (so banned contracts never pay the spawn cost or flip pending state). The shape mirrors the existing inbound wire-boundary and egress gates: an `is_banned` check plus a `phase`-tagged `tracing::debug!`, then skip.

Banned-contract skips are counted in a dedicated `skipped_banned` cycle counter (surfaced in the cycle-complete summary log) rather than the existing `skipped`/`skipped_rate_limited` counter, so the summary doesn't misreport a banned-contract skip as rate-limited.

Option 2 (actively unsubscribe / drop interest on ban) was the more thorough alternative but a larger change; the issue explicitly flags Option 1 as the contained fix. The local subscription still expires naturally via its lease, and the wire-boundary drops (#4300) already prevent inbound traffic for the banned contract.

## Testing

Building a real `Ring` for a behavioral test requires scaffolding that does not yet exist in this suite (documented in the adjacent `k_closest_source_tests` module), so coverage mirrors the established convention for this file:

- **`renewal_ban_gate_source_tests::renewal_loop_gates_banned_contracts_before_spawning`** — source-pins the gate into the renewal loop and asserts it runs *before* `can_request_subscription`, `mark_subscription_pending`, and the `run_renewal_subscribe` spawn. This fails without the fix (the `contract_ban_list.is_banned` text is absent from the loop) and passes with it. Mirrors the `*_dispatch_gates_banned_contracts` pins in `contract_ban_list.rs`.
- **`renewal_ban_gate_behavior_tests::banned_contract_is_skipped_for_renewal_while_unbanned_proceeds`** — exercises the exact predicate the gate uses (`is_banned(contract.id())`, including the `ContractKey::id()` projection) against the real `ContractBanList`: banned → skipped, un-banned → eligible, and eligible again once the ban TTL expires.

Both tests pass locally. `cargo fmt --check` is clean.

Closes #4373
References #4300, #4371

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1332,6 +1332,27 @@ impl Ring {
                     break;
                 }
 
+                // Phase 7 egress gate (#4373). Don't renew a subscription
+                // for a contract we have banned: a renewal re-registers
+                // interest via the same outbound-SUBSCRIBE machinery as a
+                // client-initiated request, but unlike the four
+                // `start_client_*` originator entry points this scheduler
+                // doesn't pass through `reject_if_contract_banned`. Without
+                // this check the node keeps emitting outbound SUBSCRIBE
+                // renewals for a banned-but-still-subscribed contract on
+                // every maintenance cycle until the ban TTL lifts. Checked
+                // before `can_request_subscription` so a banned contract is
+                // skipped regardless of its spam-backoff state.
+                if ring.contract_ban_list.is_banned(contract.id()) {
+                    tracing::debug!(
+                        %contract,
+                        phase = "subscription_renewal_banned_skip",
+                        "skipping subscription renewal for banned contract"
+                    );
+                    skipped += 1;
+                    continue;
+                }
+
                 // Check spam prevention (respects exponential backoff and pending checks)
                 if !ring.can_request_subscription(&contract) {
                     skipped += 1;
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary (FULL tier — subscription/routing path)

This change touches the subscription-renewal scheduler, a high-risk surface, so it was reviewed at Full tier. Review was run serially (dispatched as a subagent, which cannot spawn parallel reviewers).

**External model pass — Codex (`gpt-5.5`, `codex review --base origin/main`):** No actionable correctness issues. Codex confirmed the ban gate is placed before the subscription backoff/pending checks and before the outbound renewal spawn, matching the intended behavior. (Codex's own attempt to run the test suite failed only due to a sandbox network restriction resolving `crates.io`, not a code issue.)

**Four review lenses (run serially):**
- *Code-first:* found one issue — the cycle-complete summary log labeled the shared `skipped` counter as `skipped_rate_limited`, so a banned-contract skip would have been misreported as rate-limited. **Fixed** by adding a dedicated `skipped_banned` counter + log field.
- *Testing:* source-pin test fails before the fix (gate text absent) and passes after; behavioral predicate test covers banned/un-banned/TTL-expiry. No tests removed. Adequate given the documented Ring test-scaffolding constraint.
- *Skeptical:* `contract.id()` matches `is_banned`'s `&ContractInstanceId` parameter (same projection the four `start_client_*` gates use). A ban landing between the check and the spawn is harmless and bounded (next cycle catches it). No new `biased;`/`send().await`/channel concerns introduced.
- *Big-picture:* aligns with the issue's preferred Option 1; no CI-chasing; no stale agent rules to update. The new `skipped_banned` telemetry improves observability of the fixed behavior.

All findings addressed. The `skipped_banned` refinement is a pure-additive log change made after the external pass; per the review-staleness rule it does not require re-running the external review.

Pre-existing note (out of scope, not introduced here): `cargo clippy --all-features` trips `wildcard_enum_match_arm` on two `_ =>` arms in `crates/core/src/tracing.rs` (identical on `origin/main`). CI uses default features (`cargo clippy --locked`), where those arms compile cleanly, so this does not gate this PR. Bundling that cleanup here would be scope creep on a focused bug fix.

[AI-assisted - Claude]

PR #4448, opened 2026-06-16
feat(service): auto-migrate legacy install.sh install on first DMG launch

### LINKED ISSUE #3943: macOS: auto-migrate install.sh legacy install on first DMG launch
## Problem

Users who previously installed Freenet via the old `install.sh` path on macOS have:

- Legacy CLI binary at `/usr/local/bin/freenet` (or `~/.local/bin/freenet`)
- Legacy launchd agent `~/Library/LaunchAgents/org.freenet.node.plist` that auto-starts the legacy binary on login

If such a user downloads the new signed DMG and installs Freenet.app, the current DMG code (landed in #3928 and follow-ups) only:

1. Detects `org.freenet.node.plist` via `legacy_launchd_agent_present()` and logs a warning with manual cleanup instructions
2. Treats "legacy present" as "already enabled" for the Launch-at-Login decision, so it doesn't write a duplicate plist

It does NOT remove the legacy plist, unload its launchd agent, or remove the legacy CLI. Consequences:

- On next login, launchd auto-starts the legacy (old-version) binary, binding port 7509. The new DMG's Freenet.app has to fight for the port if the user manually opens it.
- `kill_stale_freenet_processes` in the DMG wrapper `pkill`s the legacy backend on startup, triggering a flap where the legacy launchd agent respawns it — then the new wrapper's backend loses the port race and exits 43.

Existing users are largely fine because the legacy install has its own auto-update mechanism and will keep picking up new versions. But anyone who WANTS to switch to the DMG path (e.g., because they want the tray and Launch-at-Login UX, or we promote the DMG as the preferred install) has to manually clean up first, per the warning log line.

## Proposed solution

On first launch of DMG-installed Freenet.app (gated on the first-run marker), if `legacy_launchd_agent_present()` returns true, auto-migrate:

1. `launchctl bootout gui/$UID ~/Library/LaunchAgents/org.freenet.node.plist`
2. `rm ~/Library/LaunchAgents/org.freenet.node.plist`
3. `rm /usr/local/bin/freenet` and/or `rm ~/.local/bin/freenet` and/or `rm ~/bin/freenet` (whichever exist and are owned by the user)
4. Log each action clearly in the wrapper log
5. Proceed with the normal DMG-install first-run flow, which then writes the NEW `org.freenet.Freenet.plist`

Silent is appropriate — this matches what a Windows installer would do on upgrade and avoids a mystery dialog for users who don't understand their own install history. Data directories (`~/Library/Application Support/freenet/`) are shared between old and new installs, so user state is preserved across the migration.

Alternatives considered:

- **AppleScript prompt** ([Migrate / Skip]): more respectful of user choice but adds GUI dependency and a confusing dialog for users who don't know they have two installs.
- **Documentation only**: the current approach plus a quickstart-page note. Relies on users reading. Acceptable for v0.2.49 release but should not be the long-term answer.

## Implementation notes

- Pure-function extraction per `.claude/rules/deployment.md`: `legacy_install_migration_plan(exe_exists: bool, plist_exists: bool) -> MigrationAction` → exhaustive cases, unit-testable on Linux CI.
- Scope to macOS only.
- Needs to handle `sudo` permission if the legacy CLI lives at `/usr/local/bin/freenet` and isn't user-owned. Fall back to logging "please manually remove X" if rm fails, rather than failing the whole migration.
- Write a regression test that reproduces the legacy-install scenario with a temp `$HOME` and asserts the migration cleans up correctly.

## Priority

Medium. New-user DMG install path is unaffected (no legacy present). Existing install.sh users keep working via legacy auto-update. Only actively affects users who try to switch install mechanisms, which is a minority. Worth doing before we promote the DMG as the default install in the quickstart, but not urgent enough to block v0.2.49.

## Context

Discussed in the lead-up to v0.2.49 (Freenet.org/quickstart will point at the DMG download once the dup-tray fix from #3939 lands). Ian confirmed existing users don't strictly need to migrate because the legacy install already auto-updates, so this is quality-of-life for deliberate migrators rather than a correctness blocker.

[AI-assisted - Claude]

sanity:
## Problem

Users who installed Freenet via the old `install.sh` path on macOS have a
legacy launchd agent (`~/Library/LaunchAgents/org.freenet.node.plist`) that
auto-starts a legacy CLI binary on login, plus that binary on disk
(`~/.local/bin/freenet` by default, `/usr/local/bin/freenet`, etc.).

When such a user installs the signed DMG, the current code (from #3928 and
follow-ups) only *detects* the legacy plist via `legacy_launchd_agent_present()`
and logs a manual-cleanup warning. It does not remove the legacy plist, boot
out its launchd agent, or remove the legacy CLI. Consequences:

- On every login, launchd auto-starts the legacy (old-version) backend and
  binds port 7509.
- The DMG wrapper's `kill_stale_freenet_processes` pkills the legacy backend on
  startup; launchd respawns it; the new wrapper's backend loses the port race
  and exits 43.

A user who wants to switch to the DMG install has to clean up manually.

## Solution

On the first DMG launch (gated on the existing first-run marker), if a legacy
`org.freenet.node` launch agent is present, auto-migrate:

1. `launchctl bootout gui/$UID <plist>` then remove the plist.
2. Remove the legacy CLI binaries (`freenet`/`fdev`) under `~/.local/bin`,
   `~/bin`, and `/usr/local/bin` that actually exist.
3. Let the normal first-run flow register the new `org.freenet.Freenet` agent
   (after migration the legacy plist is gone, so the existing
   `first_run_launch_at_login_action` naturally takes the `Register` path).

Silent migration matches what a Windows installer does on upgrade. Data dirs
(`~/Library/Application Support/freenet/`) are shared between old and new
installs, so user state is preserved.

The **decision** of what to migrate is extracted into a pure function,
`legacy_install_migration_plan(is_first_run, legacy_plist, existing_legacy_binaries)
-> LegacyMigrationPlan`, with no side effects, so it is unit-testable on Linux
CI with a mock `$HOME`. The **execution** (`launchctl bootout`, `fs::remove`)
lives in a macOS-only wrapper, `run_legacy_install_migration`. This split
follows the deployment-rule pattern (`first_run_marker_*`, `compute_menu_state`).

**Sudo fallback:** each step is attempted independently and is best-effort. A
root-owned `/usr/local/bin/freenet` the user can't `rm` without `sudo` logs a
"please remove manually: sudo rm <path>" hint rather than aborting the
migration or failing wrapper startup. Migration is idempotent: bootout/remove
of an already-gone object is a no-op.

**Scope:** migration anchors on the legacy *plist* (the thing racing for the
port). A lone legacy binary with no auto-start agent is harmless and left alone.

## Testing

Pure-function unit tests (run on Linux CI), covering all decision cases:

- `legacy_install_migration_plan_removes_plist_and_binaries_on_first_run` —
  regression test for this issue: asserts the plan boots out + removes the
  legacy plist first, then removes each detected legacy binary, in order.
- `legacy_install_migration_plan_handles_plist_only` — plist present, no binary.
- `legacy_install_migration_plan_skips_when_not_first_run` — never re-migrates.
- `legacy_install_migration_plan_skips_when_no_legacy_plist` — clean install /
  plist absent → no migration even if a stray binary exists.
- `legacy_cli_binary_candidates_cover_known_install_dirs` /
  `legacy_cli_binary_candidates_without_home_only_system_dir` — candidate paths.

Fail-before/pass-after verified: stubbing the function back to the pre-fix
"warn-only, no migration" behavior makes the two migration regression tests
FAIL (`assertion left == right failed`); the fix makes all 8 pass.

The macOS-only execution wrapper and the `macos_launch_at_login_startup` wiring
are `#[cfg(target_os = "macos")]` and verified by the macOS CI build; the
testable decision logic is platform-independent per `.claude/rules/deployment.md`.

## Fixes

Closes #3943

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -763,6 +763,34 @@ pub(super) fn first_run_launch_at_login_action(
     }
 }
 
+/// Decide whether to (re)write the DMG `org.freenet.Freenet` launch agent on
+/// startup. PURE so the combined first-run + post-migration logic is unit-
+/// testable without filesystem/launchctl side effects.
+///
+/// Returns `true` when EITHER:
+/// - this is a first-run onboarding launch and nothing auto-starts Freenet yet
+///   (`first_run_launch_at_login_action` says `Register`), OR
+/// - we just removed the legacy plist this launch (`migrated_off_legacy`) and
+///   have no plist of our own (`own_plist_present == false`). This second arm
+///   exists because an EXISTING DMG user (onboarding already complete, so NOT
+///   a first run) may never have gotten our agent — the pre-migration code
+///   treated the legacy plist as already-enabled. After we remove the legacy
+///   plist they would otherwise be left with NO auto-start at all (Codex P2,
+///   round 2 on PR #4448).
+#[allow(dead_code)]
+pub(super) fn launch_at_login_needs_registration(
+    is_first_run: bool,
+    already_enabled: bool,
+    migrated_off_legacy: bool,
+    own_plist_present: bool,
+) -> bool {
+    let first_run_register = matches!(
+        first_run_launch_at_login_action(is_first_run, already_enabled),
+        FirstRunLaunchAtLoginAction::Register
+    );
+    first_run_register || (migrated_off_legacy && !own_plist_present)
+}
+
 /// Decision outcome for a user-initiated Launch at Login toggle (click
 /// on the tray menu check item). Pure function, same testability
 /// rationale as `first_run_launch_at_login_action`.
@@ -1250,36 +1278,43 @@ pub(super) fn macos_launch_at_login_startup(log_dir: &Path) {
         );
     }
 
-    // First-run auto-register. Treat the legacy plist as already-enabled
-    // so migrating users don't end up with two plists (Codex P2).
-    let already_enabled = is_launch_at_login_enabled() || legacy_present;
-    match first_run_launch_at_login_action(is_first_run, already_enabled) {
-        FirstRunLaunchAtLoginAction::Register => match std::env::current_exe() {
+    // Auto-register the DMG launch agent. Two independent reasons to register:
+    //
+    // 1. First-run onboarding: register unless something is already enabled
+    //    (our own plist, or a legacy plist we couldn't remove).
+    // 2. Post-migration replacement: we just removed the ONLY auto-start plist
+    //    (the legacy one) and have no plist of our own. This can happen for an
+    //    existing DMG user whose onboarding already completed (so it is NOT a
+    //    first run) but who never got our agent because the old code treated
+    //    the legacy plist as already-enabled. Without this, migration would
+    //    leave them with Launch at Login silently OFF (Codex P2, round 2).
+    let own_plist_present = is_launch_at_login_enabled();
+    let already_enabled = own_plist_present || legacy_present;
+    let migrated_off_legacy = migrated && !legacy_present;
+    if launch_at_login_needs_registration(
+        is_first_run,
+        already_enabled,
+        migrated_off_legacy,
+        own_plist_present,
+    ) {
+        match std::env::current_exe() {
             Ok(exe) => match enable_launch_at_login(&exe) {
-                Ok(()) => log_wrapper_event(log_dir, "First-run: registered Launch at Login agent"),
+                Ok(()) => log_wrapper_event(log_dir, "Registered Launch at Login agent"),
                 Err(e) => log_wrapper_event(
                     log_dir,
-                    &format!("First-run Launch at Login registration failed: {e}"),
+                    &format!("Launch at Login registration failed: {e}"),
                 ),
             },
             Err(e) => log_wrapper_event(
                 log_dir,
-                &format!("First-run Launch at Login: could not resolve current exe: {e}"),
+                &format!("Launch at Login: could not resolve current exe: {e}"),
             ),
-        },
-        FirstRunLaunchAtLoginAction::AlreadyEnabled => {
-            // Nothing to do for first-run, but the plist may still be stale.
-            // Only refresh if our OWN plist is present — don't resurrect
-            // a rewrite cycle on a legacy-only install.
-            if is_launch_at_login_enabled() {
-                refresh_launch_at_login_plist_if_stale();
-            }
         }
-        FirstRunLaunchAtLoginAction::NotFirstRun => {
-            if is_launch_at_login_enabled() {
-                refresh_launch_at_login_plist_if_stale();
-            }
-        }
+    } else if own_plist_present {
+        // Not registering this launch; if our OWN plist is present it may be
+        // stale (user moved the .app, new DMG location). Refresh it. Don't
+        // resurrect a rewrite cycle on a legacy-only install (own plist absent).
+        refresh_launch_at_login_plist_if_stale();
     }
 }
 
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary (multi-model, Light tier + adversarial lenses)

External model passes (Codex across rounds 1-5; Gemini on the final content
after Codex hit its usage limit) plus four review lenses (code-first, testing,
skeptical, big-picture) run by me. The review surfaced a chain of real issues —
each fixed and re-reviewed on the new content. The final code has a clean
external (Gemini) pass and no open findings.

### Findings addressed (all from the external passes)

- **Codex P2 (round 1) — over-broad binary deletion.** The first version
  removed binaries from hard-coded candidate dirs, risking deletion of an
  unrelated `freenet`/`fdev`. **Fixed:** derive the legacy install's artifacts
  from its OWN plist + wrapper script (`legacy_program_path_from_plist`,
  `legacy_binary_path_from_wrapper`); never a guessed location.
- **Codex P1 (round 2) — migration gated on the onboarding first-run marker.**
  That would skip existing DMG users who need the migration. **Fixed:** gate on
  a dedicated `.legacy-migration-complete` marker, decoupled from onboarding.
- **Codex P2 (round 2) — replacement agent not registered for an
  already-onboarded migrated user.** Removing the legacy plist could leave them
  with no auto-start. **Fixed:** register the replacement during migration.
- **Codex P2 (round 3) — registration retry.** Folded into the round-5 design
  below.
- **Codex P2 (round 4) — validate ownership before deleting.** A repurposed
  `org.freenet.node` plist could point at an unrelated script. **Fixed:**
  `is_legacy_freenet_wrapper` requires the canonical `freenet-service-wrapper.sh`
  name AND the generated header signature; the binary must be named `freenet`.
- **Codex P2 (round 5) — don't override a later user disable.** A durable
  "re-register whenever our plist is absent" rule would re-enable Launch at
  Login for a migrated user who deliberately disabled it. **Fixed:** the
  replacement registration is now a ONE-SHOT tied to the migrating launch; once
  the migration marker is written, the migration path never auto-manages Launch
  at Login again, so a later disable sticks. A rare transient registration
  failure logs a "enable from the tray" hint rather than nagging forever — a
  deliberate tradeoff favouring user intent over auto-retry (Gemini independently
  reviewed this exact tension and endorsed the one-shot choice).

### Gemini pass on final content (HEAD at review time)
Verdict: "high quality and safe to merge." Its findings: (1) a claimed
parent-dir-missing bug in `mark_first_run_complete_at` — **not applicable**, the
function already calls `create_dir_all(parent)` before `File::create`; (2) a P3
micro-optimization to "seal" the marker for clean installs — declined, the
per-launch cost is a single `stat` and sealing would overload the marker's
meaning; (3) a P3 note on `xml_unescape` nested entities — not reachable for
real filesystem paths (the `&amp;` pass is ordered last to avoid double-decode);
(4) the round-3/round-5 tradeoff, which it explicitly endorsed ("No change
recommended").

### Lenses (no further blocking findings)
- **Skeptical:** the legacy plist is resolved before removal; steps are
  best-effort/idempotent; `NotFound` on removal is a no-op; a root-owned binary
  that can't be removed logs a `sudo rm` hint; a user who already has their own
  plist isn't re-registered (no duplicate).
- **Testing:** fail-before/pass-after verified for the original regression. 14
  pure-function unit tests run on Linux CI; a macOS-only test pins the parsers
  to `generate_plist`/`generate_wrapper_script`'s exact output.
- `cargo fmt` + `cargo clippy --locked -- -D warnings` clean on every revision.

[AI-assisted - Claude]

PR #4405, opened 2026-06-14
feat(transport): capture remote peer protocol version per-connection

sanity:
## Problem

The CONNECT handshake exchanges each peer's protocol version and runs a compatibility check, but the remote version is then discarded into a process-global aggregator (`report_peer_version`) — `RemoteConnection` keeps no record of it. A send path therefore cannot tell what version a specific connected peer is running.

This is groundwork for the contract-migration feature (#4404): a new fire-and-forget `SubscribeHint` message must only be sent to peers new enough to understand it (core P2P enums are plain bincode, so an old peer that received an unknown variant would fail to deserialize the whole message and drop the connection). That gate needs the remote version available per-connection.

## Solution

Capture the negotiated remote version on `RemoteConnection` and expose it via `PeerConnection::remote_version() -> Option<(u8, u8, u16)>`:

- Populated on the gateway/acceptor path (captures the joiner's version) and the peer-to-peer hole-punch (RemoteInbound) path.
- `None` on the joiner→gateway path: the gateway's `AckConnection` payload carries no version (documented inline; callers must treat `None` as "unknown — don't assume support"). Extending `AckConnection` to carry a version is a wire change, out of scope here.

The `PeerConnectionApi` trait method that the send path will dispatch through is **deferred to the #4404 send-path PR**, where its consumer lives — this PR is just the capture + the inherent getter, so there's no unused surface.

No handshake behavior change:

- The gateway-path version `match` is rewritten as an expression that additionally returns the parsed version. The incompatible-peer arm is byte-for-byte unchanged and still sends `ack_error` and returns `ProtocolVersionMismatch` before any connection is built.
- `decrypt_asym` gains a `&mut Option<(u8,u8,u16)>` out-param, mirroring its existing `&mut Option<Aes128Gcm>` / `&mut HandshakePhase` out-params; its single caller is updated.

## Testing

- `simulate_nat_traversal` asserts the symmetric peer-to-peer RemoteInbound path captures `Some(PROTOC_VERSION)`.
- `simulate_gateway_connection` asserts the gateway/acceptor side captures the joiner's `Some(PROTOC_VERSION)` and the joiner→gateway side is `None` — covering all three capture paths.

Part of #4404.

[AI-assisted - Claude]

PR #4449, opened 2026-06-16
fix(transport): bounds-check asym intro packet before slicing (#4406)

### LINKED ISSUE #4406: decrypt_asym slices intro packet without length check (potential panic on short asym packet)
## Problem

In `crates/core/src/transport/connection_handler.rs`, `decrypt_asym` slices the decrypted intro packet by `&...data()[..PROTOC_VERSION.len()]` (8 bytes) without first checking the decrypted payload length (around the `is_compatible` call). Asymmetric encryption to our public key does not authenticate the sender, so any peer can encrypt an arbitrarily short payload to us; if the decrypted plaintext is shorter than the version prefix, the slice would panic (index out of bounds).

Flagged during review of #4405 (which added a `remote_protoc_version` out-param near this code but did not change the slicing). **Pre-existing on `main`** — not introduced by #4405. Exploitability (whether a shorter-than-prefix asym packet is reachable here in practice) should be confirmed.

## Suggested fix

Bounds-check the decrypted length before slicing at the asym-intro parse site(s) in `connection_handler.rs`, dropping/returning on a too-short packet (the restart-detection path already does `if decrypted_data.len() >= proto_len + 16`). Add a transport test that feeds a short asym intro and asserts no panic.

[AI-assisted - Claude]

sanity:
## Problem

In `crates/core/src/transport/connection_handler.rs`, the outbound
NAT-traversal handshake's nested `decrypt_asym` helper sliced the
decrypted intro payload with **no length check**:

```rust
let protoc = &decrypted_intro_packet.data()[..PROTOC_VERSION.len()];      // panics if len < 8
...
let outbound_key_bytes = &decrypted_intro_packet.data()[PROTOC_VERSION.len()..PROTOC_VERSION.len() + 16]; // panics if len < 24
```

Asymmetric encryption to our public key does **not** authenticate the
sender — the transport public key is, by definition, public — so any
peer that can send us a UDP packet can encrypt an arbitrarily short
payload to us. When the decrypted plaintext is shorter than
`PROTOC_VERSION.len() + 16` (24 bytes), the slice panics with an
index-out-of-bounds, crashing the node process. This is a **remote,
unauthenticated denial of service**: no established session is needed,
just the target's address and public key (both discoverable on the
network), and gateways process inbound intro packets from untrusted
peers constantly.

Flagged during review of #4405; pre-existing on `main`, not introduced
by #4405.

The two sibling restart-detection parse sites in the same file already
guard with `decrypted_data.len() >= proto_len + 16`, and the inbound
gateway parse path uses `.get(..).ok_or_else(...)`. Only this outbound
`decrypt_asym` helper lacked the guard. The data-phase asym decrypt in
`peer_connection.rs` only checks that decryption succeeds and discards
the plaintext, so it is unaffected. This PR closes the one remaining
unguarded site; the audit found no others.

## Solution

Extract the bounds-checked split into a small, directly-testable free
function `split_asym_intro_payload(data) -> Option<([u8; 8], [u8; 16])>`
that returns `None` when the payload is shorter than the minimum, and
route `decrypt_asym` through it. On `None`, `decrypt_asym` drops the
packet silently and returns `Err(())` — matching the transport rule for
decryption failures ("drop packet silently, log at debug level, do not
disconnect"). This both fixes the panic and removes the two raw slices,
centralizing the bounds check in one place so future callers inherit it.

The named constants `ASYM_INTRO_SYM_KEY_LEN` (16) and
`ASYM_INTRO_MIN_LEN` (`PROTOC_VERSION.len() + 16`) document the intro
payload layout (`version || sym_key`) that was previously implicit in
the literal slice offsets.

## Testing

Added `test_short_asym_intro_packet_does_not_panic`, which exercises
two layers:

1. The pure `split_asym_intro_payload` bounds check, swept across every
   length from 0 up to one byte below the minimum (each must return
   `None`), plus the exact-minimum boundary (must parse and split
   correctly) and an over-long payload (must still parse).
2. The full attacker path: a real `TransportKeypair`, a short payload
   encrypted to the public key, then asymmetric decryption + parse —
   exactly what an unauthenticated remote peer can send to the outbound
   handshake path — asserting it returns `None` instead of panicking.

Verified the test **fails before the fix**: with the bounds check
removed it panics with `range end index 8 out of range for slice of
length 0` (the #4406 panic), and **passes with the fix**.

`cargo fmt`, `cargo clippy --locked -- -D warnings` (the CI command),
and the `transport::connection_handler` test module all clean locally.

Closes #4406

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/transport/connection_handler.rs
+++ b/crates/core/src/transport/connection_handler.rs
@@ -2024,44 +2024,45 @@ impl<S: Socket, T: TimeSource> UdpPacketsListener<S, T> {
                     direction = "outbound",
                     "Received intro packet"
                 );
-                let protoc = &decrypted_intro_packet.data()[..PROTOC_VERSION.len()];
-                if let Ok(protoc_bytes) = <[u8; 8]>::try_from(protoc) {
-                    match version_cmp::is_compatible(&PROTOC_VERSION, &protoc_bytes) {
-                        Ok(remote_info) => {
-                            crate::transport::report_peer_version(remote_info.version);
-                            *remote_protoc_version = Some(remote_info.version);
-                        }
-                        Err(reason) => {
-                            // Track version even on incompatible peers — aids version discovery.
-                            let remote_info = version_cmp::parse_version_bytes(&protoc_bytes);
-                            crate::transport::report_peer_version(remote_info.version);
-                            if version_cmp::remote_requires_newer_than_us(
-                                &PROTOC_VERSION,
-                                &protoc_bytes,
-                            ) {
-                                crate::transport::signal_urgent_update();
-                            }
-                            tracing::debug!(
-                                peer_addr = %remote_addr,
-                                %reason,
-                                direction = "outbound",
-                                "Remote is using an incompatible protocol version"
-                            );
-                            return Err(());
-                        }
-                    }
-                } else {
+                // Bounds-check before slicing: asym encryption to our public
+                // key does not authenticate the sender, so a too-short payload
+                // is attacker-reachable and must not panic (#4406).
+                let Some((protoc_bytes, outbound_key_bytes)) =
+                    split_asym_intro_payload(decrypted_intro_packet.data())
+                else {
                     tracing::debug!(
                         peer_addr = %remote_addr,
                         direction = "outbound",
-                        "Protocol version field wrong size"
+                        "Intro packet too short to parse"
                     );
                     return Err(());
+                };
+                match version_cmp::is_compatible(&PROTOC_VERSION, &protoc_bytes) {
+                    Ok(remote_info) => {
+                        crate::transport::report_peer_version(remote_info.version);
+                        *remote_protoc_version = Some(remote_info.version);
+                    }
+                    Err(reason) => {
+                        // Track version even on incompatible peers — aids version discovery.
+                        let remote_info = version_cmp::parse_version_bytes(&protoc_bytes);
+                        crate::transport::report_peer_version(remote_info.version);
+                        if version_cmp::remote_requires_newer_than_us(
+                            &PROTOC_VERSION,
+                            &protoc_bytes,
+                        ) {
+                            crate::transport::signal_urgent_update();
+                        }
+                        tracing::debug!(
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary (FULL tier — transport/crypto high-risk surface)

This change touches the transport layer and asymmetric-decryption handshake path, so it was reviewed at FULL tier.

**External model pass — Codex (`codex review --base origin/main`):** clean.
> "The change adds a bounds-checked parser for asymmetric intro payloads and uses it in the previously panicking path without altering valid-payload behavior. I did not identify any actionable regressions in the diff; cargo fmt passes."
(Codex could not run `cargo test` — its sandbox has no crates.io access — but tests were verified locally; see below.)

**Claude four-lens read of the checked-out diff:**
- *Code-first:* valid-payload behavior is byte-for-byte preserved — the version-compat match, `report_peer_version`, `signal_urgent_update`, and the `remote_protoc_version` out-param are unchanged; the only behavioral change is that a sub-24-byte payload now logs "too short to parse" at debug and drops, instead of panicking. `Aes128Gcm::new_from_slice(&[u8; 16])` still always succeeds, so `.expect("correct length")` remains correct.
- *Testing:* the regression test was confirmed to **fail before the fix** (panics `range end index 8 out of range for slice of length 0`) and **pass after**. It sweeps every too-short length 0..23, the exact boundary, an over-long payload, and the full encrypt→decrypt→parse attacker path.
- *Skeptical:* the helper's two `.expect()`s are unreachable after the `len >= 24` guard (documented). Over-long payloads ignore trailing bytes, matching prior slice behavior.
- *Big-picture:* matches the established pattern — the two sibling restart-detection sites already guard with `len() >= proto_len + 16` and the inbound gateway path uses `.get(..)`. Audited all four asym-decrypt sites in transport; this was the only unguarded one. No tests removed, no stale docs.

**Local verification:** `cargo fmt`, `cargo clippy --locked -- -D warnings` (the exact CI command) clean; `transport::connection_handler` test module 61 passed / 0 failed.

No findings required changes.

[AI-assisted - Claude]

Issue #3973, opened 2026-04-29
Skip Simulation and NAT Validation jobs in merge_group


sanity:
## Problem

Merge-queue CI runs three `ubicloud-standard-16` jobs (Unit & Integration, Simulation, NAT Validation), which exceeds Ubicloud's per-project concurrency cap during peak hours. On 2026-04-29 some merge_group runs had jobs queued for 55–101 minutes before a runner picked them up, blocking merges.

Sample evidence (2026-04-29 UTC, time from run trigger to job start):

| Run | Unit&Int | Simulation | NAT |
|---|---|---|---|
| 07:25 | +66m | +81m | +101m |
| 08:28 | +55m | +67m | +79m |
| 13:53 | +4m | +4m | +8m |

## Proposal

Skip Simulation and NAT Validation in `merge_group` (continue running both on `pull_request`). Keep them present as no-op jobs so branch-protection required checks stay satisfied, using the same `release_check` skip pattern already in `ci.yml`.

Keep at merge_group: Fmt, Clippy, Windows Check, Unit & Integration, macOS DMG-swap E2E.

## Rationale

The unique value of merge_group CI (vs PR CI) is catching what the rebase onto latest `main` introduces:
- semantic merge conflicts between independently-tested PRs
- dependency API or behavior breakage in unmerged PRs
- stale-PR drift

These surface as compile errors (Clippy / Windows Check) or unit/integration failures essentially every time. Simulation and NAT Validation test ring routing, topology, and Docker NAT hole-punching: behaviors driven by the PR's own code, not by how it merges. Their merge-time unique catch rate is near zero.

The post-merge CI run on `main` remains the backstop if anything slips past PR-level.

## Estimated impact

- ~25 min × 16 vCPUs freed per merge_group run (~400 vCPU-min)
- Resolves the 3-job contention against Ubicloud's concurrency cap
- Removes the largest single source of merge-queue latency

## Risks

- Low. A bug that fails Simulation or NAT only after rebase, but passes Unit & Integration after rebase, is theoretically possible but I cannot construct a realistic scenario for it.
- Mitigation if it ever happens: the offender shows up on `main` post-merge CI; revert and re-run.

## Compromise option

If we want a minimal sim canary at merge_group, keep only `ci-medium-50` (~5 min) and skip the other three fdev sims plus the nextest sim suite. Still skip NAT Validation entirely. Captures roughly 70% of the savings.

## Implementation

[...]

PR #4450, opened 2026-06-16
test(governance): run the ban-chain e2e simulation tests in CI

### LINKED ISSUE #4301: Phase 7 follow-up: end-to-end simulation test for governance ban chain

sanity:
## Problem

Issue #4301 asks for an end-to-end simulation-network test of the governance ban chain (rate-limit reject → MAD flag → `Normal → … → Evicted → Banned` → wire-traffic drop), as the **required behavioral gate before flipping governance into Enforce mode** on gateways.

The tests themselves **already exist**: the `sim_e2e_tests` module in `crates/core/src/contract/governance.rs` (added in PR #4309) contains `governance_ban_chain_end_to_end` and `popular_contract_with_subscribers_not_evicted`. They drive the full production wiring inside a `run_controlled_simulation` network (1 gateway + 5 nodes):

- `governance_ban_chain_end_to_end` — floods an abuser contract with heavy UPDATEs (cost flows through the real `Executor → Ring::commit_state_write → report_contract_resource_usage → GovernanceManager::ingest_cost` path, driven by the per-node `governance_reaper_loop` on virtual time), asserts the abuser lands on the per-node `ContractBanList` (the wire-boundary enforcement signal that `node.rs` consults to drop banned traffic) and reaches the terminal `Banned` state, and asserts every honest contract is NOT banned (no collateral damage).
- `popular_contract_with_subscribers_not_evicted` — the #4296 false-positive control: a same-cost contract with many live downstream subscribers is NOT banned on its host, while the zero-beneficiary abuser IS.

**The gap this PR closes:** those tests never actually ran in CI, so the gate was not enforced and the tests could silently rot. The module is gated behind `#[cfg(all(test, feature = "simulation_tests"))]` and lives in the lib — it must, because it uses `pub(crate)` accessors (`Ring::governance`, `Ring::contract_ban_list`, the test-only `hosting_manager_*` counters) that an external `tests/` integration binary cannot reach. But:

- the workspace test job runs the lib tests **without** `simulation_tests`, so the module is compiled out there; and
- the Simulation job enables `simulation_tests` but only runs specific `--test <integration_binary>` targets, which **excludes lib tests**.

So the module fell through the crack between the two jobs and was dead in CI.

## Solution

Add `--lib` to the Simulation job's build (`--no-run`) and run steps so the lib test binary is compiled and executed alongside the integration binaries. To avoid re-running the ~3000 ordinary freenet lib unit tests (already covered by the workspace test job), the run step restricts execution with a nextest filterset:

```
kind(test) | (kind(lib) & test(sim_e2e_tests))
```

i.e. all integration tests **plus only** the governance sim e2e lib tests.

**No production code changes** — this is a CI/test wiring fix only.

## Testing

Ran the two tests locally under the CI feature set; both pass deterministically:

```
cargo nextest run -p freenet --no-default-features \
  --features "trace,websocket,redb,wasmtime-backend,testing,simulation_tests" \
  --lib -E 'test(sim_e2e_tests)'
# 2 tests run: 2 passed
#   governance_ban_chain_end_to_end            115.5s
#   popular_contract_with_subscribers_not_evicted  88.3s
```

Verified via `cargo nextest list` that the filterset selects exactly the integration tests plus the two sim e2e lib tests (not the other lib unit tests).

Closes #4301

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -545,7 +545,19 @@ jobs:
           CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER: clang
           CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS: -C link-arg=-fuse-ld=mold
         run: |
+          # `--lib` is included so the in-crate simulation tests gated behind
+          # `#[cfg(all(test, feature = "simulation_tests"))]` (e.g. the
+          # governance ban-chain e2e module in
+          # crates/core/src/contract/governance.rs) are built and run. They use
+          # `pub(crate)` accessors (Ring::governance, Ring::contract_ban_list,
+          # the test-only hosting_manager_* counters) that an external
+          # tests/ integration binary cannot reach, so they MUST live in the
+          # lib. Without `--lib` here this whole class of tests was compiled
+          # out of the workspace test step (no simulation_tests feature) AND
+          # excluded from this simulation step (only --test binaries), so it
+          # never ran in CI. See issue #4301.
           cargo nextest run -p freenet --locked --no-run \
+            --lib \
             --test simulation_integration \
             --test simulation_smoke \
             --test streaming_e2e \
@@ -577,11 +589,21 @@ jobs:
           NEXTEST_FAILED=0
           (
             echo "== $(date -u +%H:%M:%S) nextest starting"
+            # `--lib` makes the in-crate simulation_tests-gated tests (see the
+            # build step above and issue #4301) available to run alongside the
+            # integration binaries. The filterset then restricts execution to
+            # ALL integration tests (`kind(test)`) PLUS only the simulation
+            # e2e lib tests (`kind(lib) & test(sim_e2e_tests)`) — so the ~3000
+            # ordinary freenet lib unit tests (already covered by the
+            # workspace test job) are NOT re-run here. They were all built with
+            # --no-run above.
             cargo nextest run -p freenet --locked \
+              --lib \
               --test simulation_integration \
               --test simulation_smoke \
               --test streaming_e2e \
               --test state_verification \
+              -E 'kind(test) | (kind(lib) & test(sim_e2e_tests))' \
               --features simulation_tests,testing --profile ci 2>&1
             rc=$?
             echo "== $(date -u +%H:%M:%S) nextest done rc=${rc}"


sanity:
## Multi-model review — clean

Risk tier: this touches CI configuration (a high-risk surface per the project review rules), so I ran the full review even though the diff is small and test-only. As a subagent I ran the lenses serially.

**External model (Codex, gpt-5.5)** — clean, no introduced correctness issues. Codex independently reproduced the nextest filterset behavior in a scratch crate and confirmed `kind(test) | (kind(lib) & test(sim_e2e_tests))` selects all integration tests plus only the `sim_e2e_tests` lib tests (excluding ordinary lib unit tests). Verdict: *"The workflow change correctly includes the library test target and uses a nextest filter that runs integration tests plus only the intended simulation e2e lib tests. I did not identify any introduced correctness issues in the patch."*

**Claude lenses (serial):**
- *Code-first*: the `--no-run` build step compiles all `--lib` + `--test` targets; the run step's `-E` filter then restricts execution. Consistent.
- *Skeptical*: no integration-test regression — previously the run had no `-E` and ran all tests in the four `--test` binaries; `kind(test)` matches exactly those, so coverage is unchanged, plus the 2 sim e2e lib tests. The ~3000 ordinary lib unit tests (already covered by the workspace test job) are NOT re-run.
- *Testing*: adds coverage, removes none. Both tests are deterministic (fixed seeds, `run_controlled_simulation`, seed-derived data — no RNG / wall-clock) and pass locally (`governance_ban_chain_end_to_end` 115.5s, `popular_contract_with_subscribers_not_evicted` 88.3s).
- *Big-picture*: #4301's intent is that the ban-chain sim test acts as a real gate before the Enforce flip. The tests already existed (PR #4309) but never ran in any CI job — making them run is exactly what fulfills the issue; writing a duplicate test would be wrong. These tests run on PR-level CI and inherit the same non-release-merge_group skip as every other sim test (#3973), which is the intended, consistent behavior.

No blocking findings. Proceeding once CI is green.

[AI-assisted - Claude]

Issue #4441, opened 2026-06-16
Module cache thrash / OOM-spiral on nodes hosting >1024 contracts (capacity hardcoded, not configurable)


sanity:
## Summary

Nodes hosting **more than ~1024 contracts** thrash the WASM module cache (hardcoded 1024-entry LRU), and on resource-constrained hosts this now OOM-spirals. Surfaced by a user (Ivvor) on v0.2.74: a small VPS hosting >1024 contracts "death-spiraled every ~5 min" (OOM-restart) with "a LOT of module cache misses ... cycling over all of them."

## Why it appeared with v0.2.74 (not a regression)

v0.2.74 (#4438) made freenet self-raise `RLIMIT_NOFILE`, fixing the v0.2.73 gateway crash-loop (EMFILE: 1024 memfds vs 1024 fd soft limit). For a >1024-contract node, v0.2.73 *crash-looped on fd exhaustion before the thrash could hurt*. v0.2.74 removes that crash, so the node stays up and instead thrashes CPU + grows RSS until OOM. The fd fix is correct; it unmasked a pre-existing limitation.

## Root facts

- Module cache is a shared `LruCache`, capacity hardcoded `DEFAULT_MODULE_CACHE_CAPACITY = 1024` (`crates/core/src/wasm_runtime/runtime.rs:55`), each entry backed by a `memfd`.
- The production `RuntimePool` hardcodes the cap (`crates/core/src/contract/executor/runtime.rs:239-242`). A `RuntimeConfig.module_cache_capacity` field exists (`wasm_runtime/runtime.rs:174`) but is **never wired** to any Config / env / CLI — so operators **cannot act on the eviction warning that tells them to "increase DEFAULT_MODULE_CACHE_CAPACITY."**
- Cache miss → recompile → LRU evict (`wasm_runtime/runtime.rs:497-533`). Working set > 1024 = evict+recompile on every write to a cold contract.
- Each recompile grows Wasmer `code_memory` unbounded (#2941, documented at `runtime.rs:40-48`) — this is what turns CPU thrash into an OOM spiral over time.
- The delegate cache has the same 1024 cap (`runtime.rs:572`).

## Proposed work (design fix, not urgent — gateways stable, NRestarts=0; only resource-constrained >1024-contract nodes affected)

[...]

Issue #2941, opened 2026-02-10
Wasmer Engine code_memory grows unboundedly


sanity:
## Problem

Wasmer's internal `code_memory: Vec<CodeMemory>` only grows and never shrinks. Compiled machine code persists even after the `Module`/`Artifact` Arc is dropped. The memory is only freed when the entire Engine is dropped.

When the module cache evicts an entry and the same contract is re-compiled, the new compiled code is added to `code_memory` while the old compiled code remains allocated. This means `code_memory` grows without bound proportional to the number of unique compilations over the Engine's lifetime.

## Impact

From jemalloc heap profiling on nova (Feb 2026), WASM compilation via the singlepass backend accounted for **47% of memory growth** (~133 MB in 85 seconds).

## Potential Solutions

1. **Keep cache capacity large enough to hold the entire working set** — avoids eviction-recompilation cycles. PR #2940 increases capacity from 128 to 1024, which should prevent evictions on current production gateways (~92 contracts).

2. **Periodically recreate the Engine** — drop the old Engine (freeing all `code_memory`) and create a new one. This requires invalidating the module cache since modules are tied to the Engine that compiled them. Could be done during low-activity periods.

3. **Upstream fix** — contribute a `code_memory` cleanup mechanism to wasmer. This would require changes to wasmer's `EngineInner` to track which `CodeMemory` entries are still referenced by live modules.

## Context

- `Engine.inner` is `Arc<Mutex<EngineInner>>` shared across all RuntimePool executors
- Wasmer `Module` contains raw function pointers into `Engine.code_memory` — modules MUST be used with the Engine that compiled them
- The Engine Mutex is held for the entire duration of compilation (seconds), serializing all compilations

## Related PRs

- #2937 — Shared module cache across RuntimePool executors
- #2940 — Increase cache capacity + client disconnect cleanup

[AI-assisted - Claude]

PR #4452, opened 2026-06-16
fix: bound WASM module cache by memory + offload cold compile off the contract loop (#4441)

sanity:
## Problem

Nodes hosting **>1024 contracts** thrash the WASM module cache and hang/OOM (#4441). Two real defects — and the "Wasmer `code_memory` / #2941 unbounded growth" premise was **stale**: the backend is **wasmtime**, which frees compiled code when the `Module` drops (pinned by `test_module_drop_frees_memory`).

1. **Count-bounded cache.** The module cache (contract *and* delegate) is hardcoded to **1024 entries**. Compiled sizes vary wildly (measured **1.5 MiB** for a representative contract), so a count cap neither bounds memory nor reflects real pressure → thrash once the working set exceeds 1024. The existing `module_cache_capacity` config field was **dead** (never wired through `build_with_shared_module_caches` / the pool).
2. **Compile pins the serial loop (the hang).** `Module::new` (compile) ran **inline on the single-threaded `contract_handling` loop** — only WASM *calls* were offloaded to `spawn_blocking`. A cold-contract compile pins the whole loop, starving the API/diagnostics/logging → the observed hang ("no new logs, UI stuck on loading, diagnostics erroring").

## Approach (Ian-approved scope; full per-contract dispatch parallelism deferred)

- **Memory-bounded, configurable cache** — new `ModuleCache` (byte budget, LRU evict-to-budget) for contracts AND delegates. Default **384 MiB** (~250 modules at the measured size). Wired via `--module-cache-budget-bytes` / `FREENET_MODULE_CACHE_BUDGET_BYTES` / `Config` field.
- **Offload cold compile to `spawn_blocking`, gated** — ON in production, **INLINE under the deterministic sim/test runner** (`offload_compilation` flag), so determinism is preserved. The cache `Mutex` is **never held across the compile await**; the per-hash coalescing mutex prevents double-compile.
- **Fixed the stale Wasmer/#2941 rustdoc.**
- OOM: re-derived for wasmtime (no code_memory leak); bounding the cache removes the bursty-alloc thrash that drove RSS. Residual (jemalloc HWM, à la the 0.2.69 finding) to be measured if it recurs.

## Testing

- `test_module_cache_evicts_by_bytes_not_count` — byte-budget eviction (FAILS on main's count cap, passes here).
- `test_compiled_module_size_is_in_expected_range` (measures the 1.5 MiB figure); offload-gate + coalescing tests.
- **Determinism suite 18/18** (offload gated off under sim → deterministic traces preserved).
- `cargo fmt` + `clippy -D warnings` clean; wasm_runtime 263/263, executor pins 8/8, build clean.

Refs #4441. Backend-premise correction (wasmtime, not Wasmer) noted on the issue.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -111,6 +111,15 @@ pub struct ConfigArgs {
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
 
+    /// Byte budget for each compiled-WASM module cache (the contract cache and
+    /// the delegate cache each get this budget). When a cache's tracked
+    /// compiled-byte total would exceed this on insert, least-recently-used
+    /// modules are evicted until it fits. Bounding by bytes (not entry count)
+    /// stops a node hosting many contracts from thrashing the cache and
+    /// recompiling on every access (issue #4441). Default: 384 MiB.
+    #[arg(long, env = "FREENET_MODULE_CACHE_BUDGET_BYTES")]
+    pub module_cache_budget_bytes: Option<usize>,
+
     /// Seconds to wait on graceful shutdown for in-flight client
     /// PUT/GET/UPDATE/SUBSCRIBE operations to finish before tearing
     /// down peer connections. Set to 0 to disable. Default: 30s. See
@@ -166,6 +175,7 @@ impl Default for ConfigArgs {
             version: false,
             max_blocking_threads: None,
             max_hosting_storage: None,
+            module_cache_budget_bytes: None,
             shutdown_drain_secs: None,
             telemetry: Default::default(),
         }
@@ -384,6 +394,8 @@ impl ConfigArgs {
             self.log_level.get_or_insert(cfg.log_level);
             self.max_hosting_storage
                 .get_or_insert(cfg.max_hosting_storage);
+            self.module_cache_budget_bytes
+                .get_or_insert(cfg.module_cache_budget_bytes);
             self.shutdown_drain_secs
                 .get_or_insert(cfg.shutdown_drain_secs);
             self.config_paths.merge(cfg.config_paths.as_ref().clone());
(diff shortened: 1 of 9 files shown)


sanity:
## Review status (Full tier)

Four independent Claude lenses (code-first, skeptical, testing, big-picture). **The skeptical lens caught a blocking bug** and several important findings — all addressed in the fix-up commits (now rebased onto current `main`, incl. #4442):

- **[was blocking] `compile_offloaded` could panic** — `block_in_place` panics on a `current_thread` runtime, and the `!cfg!(test)` gate left offload ON in integration tests. **Fixed:** `compile_offloaded` now only uses `block_in_place` when `Handle::try_current().runtime_flavor() == MultiThread`; current-thread / no-runtime contexts compile inline (so it can never panic, and sim/determinism stays inline regardless of the flag).
- **[important] default budget could OOM a small VPS** (the target user). **Fixed:** default is now **RAM-scaled, clamped 64–384 MiB**, with a rate-limited at-budget eviction warning so constrained operators get a signal. Explicit `--module-cache-budget-bytes` / env override still wins.
- **[important] over-claimed framing** — B1 cures the whole-node *hang* (sibling tasks unblocked) but the contract loop still serializes per-compile (throughput parallelism = the deferred B2). Corrected in comments; #4441 stays open for B2 + the residual-OOM measurement.
- **[should] tests/comments** — added delegate-cache byte-budget + `remove()` accounting + current-thread/no-runtime offload tests; fixed misleading comments + the stale Wasmer/#2941 rustdoc.

**External model note:** both Codex and Gemini are currently rate-limited (Codex resets shortly; Gemini ~22h). Per the multi-model rule the 4-lens Claude pass is the documented fallback (and it caught the real blocking bug) — but since this change is not time-sensitive, I'm waiting for Codex to reset to run a proper external pass on the final content before merge. fmt + clippy clean; CI re-running the full suite (incl. `error_notification` + determinism).

[AI-assisted - Claude]

PR #4453, opened 2026-06-16
test(connect): pin retry-accept ring-promotion invariant in relay driver

### LINKED ISSUE #3894: test: add simulation test for retry-accept ring promotion invariant
## Context

Testing-review follow-up from PR #3893 (fix for #3838).

The end-to-end invariant \"an accepted joiner transitions from transient to ring on the acceptor\" is currently only covered by the flaky \`test_ping_blocked_peers\` integration test. That test is expensive (real WASM contracts, real WS API, real transport), timing-sensitive, and was hiding a ~40% failure rate behind nextest retries for an unknown period.

## Proposed test

A direct-runner simulation test (~50–100ms virtual-time) that:

1. Sets up a small deterministic topology where the gateway's uphill forward gets rejected (e.g., uphill peer at capacity or blocklisted), forcing the \`Rejected\` retry branch in \`process_message\` to accept locally.
2. After CONNECT completes, asserts that the gateway has the joiner as a **ring** connection (not transient), and vice versa.
3. Has the joiner publish a contract and subscribe; asserts the gateway's \`register_downstream_subscriber\` actually registered the interest (i.e., no silent \`could not find peer\` logs).
4. Has another peer issue an UPDATE; asserts the subscribing joiner receives it.

## Enabling work

- Expose \`transient_connection_count(label)\` on \`SimNetwork\` (currently \`connection_count\` returns ring-only, which is correct for assertion but there's no way to assert \"zero transient at steady state\"). Adding the accessor is ~10 lines in \`crates/core/src/node/testing_impl.rs\`.
- Find a deterministic way to force the uphill-rejection retry path. \`FaultConfig\` supports partitions and crashes but not \"reject CONNECT with X message\". The cleanest route is contriving a topology with \`min_connections\`/\`max_connections\` so the uphill peer naturally returns \`Rejected\`.

## Why this matters

The PR #3893 type-level refactor (\`AcceptOutcome\`) makes the specific #3838 bug class unrepresentable, but a behavioral test for the end-to-end invariant would catch future variants of \"accepted but not promoted\" from any other entry point.

Per \`.claude/rules/test-philosophy.md\` (\"Don't Discover Logic Errors in Production\"), this logic error should be reproducible in simulation, not just in the real-process ping canary.

[AI-assisted - Claude]

sanity:
## Problem

PR #3893 fixed the #3838 promotion-drop bug: when the CONNECT relay accepts
a joiner via a *retry* branch (`Rejected` from an uphill peer, or
`ConnectFailed` from a downstream acceptor), it must promote the joiner's
transient transport connection into a ring connection. The retry branches had
drifted from the happy path and forwarded the `ConnectResponse` upstream
*without* emitting the promotion, stranding the joiner in
`transient_connections` on the acceptor — invisible to `get_peer_by_addr`
(subscribe interest registration, broadcast fan-out), so UPDATEs never reached
it.

That invariant was only guarded at two levels:

- **Data-level**: `connect.rs::relay_retry_acceptance_bundles_joiner_with_response`
  (added in #3893) pins that `handle_request` returns the joiner *bundled* with
  the response via `AcceptOutcome` — the type system makes the field-drop
  unrepresentable.
- **End-to-end**: only the flaky `test_ping_blocked_peers` integration test
  (real WASM / WS / transport, ~40% historical failure rate masked by nextest
  retries).

#3894 asked for a deterministic simulation guard for the *driver-level*
invariant: every accept site actually dispatches the promotion.

## Approach

Add a deterministic source-scrape pin test
`every_relay_accept_site_promotes_joiner_to_ring` in
`connect/op_ctx_task.rs` (alongside the existing
`drive_relay_connect_handles_all_reentry_variants` /
`start_relay_connect_inserts_into_dedup_set` pins). It asserts that **all
three** production accept sites in the relay driver call
`dispatch_expect_connection_from` (which emits
`NodeEvent::ExpectPeerConnection` + `NodeEvent::ConnectPeer { is_gw: false }`):

1. happy-path terminus / near-terminus accept (`initial_actions.accept`),
2. the `Rejected` retry-accept ("accepting locally after uphill rejection"),
3. the `ConnectFailed` retry-accept ("accepting locally after ConnectFailed").

The build fails if a future edit re-inlines an accept branch that forwards the
response without dispatching the promotion — i.e. it catches the #3838
regression class at the driver level, complementing #3893's data-level pin.

### Why not a behavioural SimNetwork test?

I investigated the behavioural simulation test the issue originally proposed
(force the retry-accept path, assert transient connections drain to zero) and
found it is **not deterministically forceable**, so it would be a flaky test:

- A passive sparse-ring run (1 gw + 12 nodes, `max_connections=3`) under the
  deterministic Turmoil runner exercised only the terminus / near-terminus
  accept paths — **zero** `Rejected` / `ConnectFailed` retry-accepts. The retry
  branches require a specific emergent capacity-rejection-then-accept sequence
  that does not arise reliably.
- `FaultConfig` cannot inject a CONNECT rejection (the issue noted this), so the
  branch can't be forced directly; contriving it via capacity races is
  seed/timing-dependent.
- The same exploratory run was non-deterministic across runs at that scale/config
  (topology formation produced different outcomes for the same seed), confirming
  a behavioural assertion here would flake.

`run_simulation_direct` (the runner the issue suggested) additionally consumes
`SimNetwork` and never captures per-node `ConnectionManager`s, so post-run
transient/ring inspection isn't even possible through it.

The source-scrape pin is the deterministic, zero-flakiness guard for the same
invariant — and matches the established freenet pattern for "a task-per-tx
migration dropped a required side-effect call" (see
`.claude/rules/bug-prevention-patterns.md`).

## Testing

- `every_relay_accept_site_promotes_joiner_to_ring` passes against current
  `main`.
- **Negative control**: removing the `dispatch_expect_connection_from` call from
  the `Rejected` retry-accept branch makes the test FAIL (both the
  exactly-3-call-sites count and the per-branch check fire) — proving the pin
  would have caught the original #3838 bug and is not vacuous.
- `cargo fmt` + `cargo clippy -- -D warnings` clean; the `operations::connect`
  test module is green.

Closes #3894

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/connect/op_ctx_task.rs
+++ b/crates/core/src/operations/connect/op_ctx_task.rs
@@ -1418,4 +1418,140 @@ mod tests {
              receive loop."
         );
     }
+
+    /// Regression guard for the retry-accept ring-promotion invariant
+    /// (#3838 / PR #3893; tracked by #3894).
+    ///
+    /// The CONNECT relay accepts a joiner from THREE production sites:
+    ///
+    /// 1. the happy-path terminus/near-terminus accept
+    ///    (`initial_actions.accept`),
+    /// 2. the `Rejected` retry-accept ("accepting locally after uphill
+    ///    rejection"), and
+    /// 3. the `ConnectFailed` retry-accept ("accepting locally after
+    ///    ConnectFailed").
+    ///
+    /// EVERY accept site MUST call `dispatch_expect_connection_from` to
+    /// promote the joiner's transient transport connection into a ring
+    /// connection (it emits `NodeEvent::ExpectPeerConnection` +
+    /// `NodeEvent::ConnectPeer { is_gw: false }`). The original #3838 bug
+    /// was exactly a retry branch that sent the `ConnectResponse` upstream
+    /// but DROPPED the promotion call, stranding the joiner in
+    /// `transient_connections` on the acceptor — invisible to
+    /// `get_peer_by_addr` (subscribe interest registration, broadcast
+    /// fan-out), so UPDATEs never reached it.
+    ///
+    /// PR #3893 made the data-level drop unrepresentable by bundling the
+    /// response + joiner into `AcceptOutcome` (covered by the unit test
+    /// `connect.rs::relay_retry_acceptance_bundles_joiner_with_response`).
+    /// This source-scrape pin closes the remaining gap at the *driver*
+    /// level: it fails the build if a future edit re-inlines an accept
+    /// branch that forwards the response without dispatching the promotion.
+    ///
+    /// A behavioural simulation test was investigated for #3894 but is not
+    /// viable: the retry-accept branches are not deterministically reached
+    /// by emergent SimNetwork topology (a sparse-ring run exercised only the
+    /// terminus / near-terminus accept paths, never the `Rejected` /
+    /// `ConnectFailed` retries), and `FaultConfig` cannot inject a CONNECT
+    /// rejection to force them. Forcing the branch via capacity races would
+    /// be seed/timing-dependent (a flaky test). This pin is the deterministic
+    /// guard for the same invariant.
+    #[test]
+    fn every_relay_accept_site_promotes_joiner_to_ring() {
+        const SOURCE: &str = include_str!("op_ctx_task.rs");
+
+        // Restrict to production code: the `#[cfg(test)]` module below must
+        // not contribute (or mask) call sites.
+        let prod_end = SOURCE
+            .find("\n#[cfg(test)]\n")
+            .expect("expected a `#[cfg(test)]` module in op_ctx_task.rs");
+        let prod = &SOURCE[..prod_end];
+
+        // 1. Exactly three accept sites, each pairing its `accept` with the
+        //    promotion dispatch. The count is load-bearing: an added accept
+        //    branch without a matching dispatch would NOT raise this above 3,
+        //    so the per-branch checks below are also required.
+        let dispatch_sites = prod
+            .matches("dispatch_expect_connection_from(op_manager, incoming_tx,")
+            .count();
+        assert_eq!(
+            dispatch_sites, 3,
+            "expected exactly 3 `dispatch_expect_connection_from` call sites in \
+             the CONNECT relay driver (happy-path accept + `Rejected` retry-accept \
+             + `ConnectFailed` retry-accept), found {dispatch_sites}. If you added \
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary (self-review + Claude-lens fallback)

**External model availability:** Both external reviewers were genuinely down at review time, so neither produced a review:

- `codex review --base origin/main` → `ERROR: You've hit your usage limit ... try again at 1:29 PM` (exit 1, no findings produced).
- `gemini --approval-mode plan -p ...` (two attempts) → `You have exhausted your capacity on this model` on every retry; produced zero review content.

Per `~/.claude/rules/multi-model-review.md`, with both external models unavailable I used the **diverse-Claude-lens fallback** rather than skipping independent review. This is a **test-only** change (no production code touched — purely an added `#[test]`), which lowers the risk despite the CONNECT/ring surface. Lenses run:

- **Code-first** (intent vs impl): the test's stated guard (all 3 accept sites dispatch the promotion) matches its implementation. No mismatch.
- **Testing**: non-vacuous — negative control (removing the dispatch from the `Rejected` retry-accept branch) makes the test FAIL on the `assert_eq!(dispatch_sites, 3)` check. Covers all 3 accept sites; retry branches additionally pin the documented "promote BEFORE forwarding ConnectResponse" ordering.
- **Skeptical** (bug hunt): pure compile-time source analysis → deterministic, no flakiness. Anchors (`dispatch_expect_connection_from(op_manager, incoming_tx,`, the two unique log messages, `ConnectMsg::Response {`, the happy-path `if let`) are stable, load-bearing strings.
- **Big-picture**: aligns with #3894 intent; honestly documents why the behavioural-sim variant isn't viable; removes/weakens no existing test.

**One acknowledged limitation** (inherent to source-scrape pins, mitigated by the assertion's failure message): `assert_eq!(dispatch_sites, 3)` would NOT catch a brand-new *fourth* accept site that forgets the dispatch (count stays at 3 only if the new site also lacks one — but a new accept site + a dropped existing one would net to 3). Adding an accept site is a deliberate, reviewed change, and the failure message explicitly instructs the author to update the pin and promote the joiner. This matches the tradeoff of the existing `op_ctx_task.rs` pins (`drive_relay_connect_handles_all_reentry_variants`, etc.).

When external models reset, a real Codex/Gemini pass can be re-run before merge if desired; given the test-only scope I did not block on it.

[AI-assisted - Claude]

PR #4455, opened 2026-06-16
fix(ring): feed per-peer bandwidth into the topology resource meter

### LINKED ISSUE #3453: Resource usage meter is never fed data in production
## Problem

`TopologyManager::report_resource_usage()` is marked `#[allow(dead_code)] // fixme: use this` and is only called from test code within the topology module. No production code feeds resource usage data into the `Meter`.

This means `calculate_usage_proportion()` always returns ~0% usage, so `adjust_topology()` always takes the "usage below 50% → add connections" branch. The resource-usage-based decision logic (add when <50%, hold at 50-90%, remove when >90%) effectively never gates connection changes based on actual load.

### Affected code

- `crates/core/src/topology/mod.rs` — `report_resource_usage()` (line ~237), `report_split_resource_usage()` (line ~220)
- `crates/core/src/topology/meter.rs` — `Meter` struct and all its attribution tracking
- `crates/core/src/topology/mod.rs` — `calculate_usage_proportion()`, `extrapolated_usage()`

### Impact

- The topology manager cannot make load-aware decisions about adding/removing connections
- The "remove worst-value peer when overloaded" path (`select_connections_to_remove`) is unreachable in production
- Tests pass because they call `report_resource_usage` directly, masking the production gap

### What needs to happen

Wire up actual resource usage reporting from the node's event loop or transport layer into `TopologyManager::report_resource_usage()`. The meter infrastructure exists and is tested — it just needs a caller that feeds it real bandwidth/CPU data from production traffic.

[AI-assisted - Claude]

sanity:
## Problem

`TopologyManager`'s resource-usage `Meter` is never fed per-peer bandwidth
data in production. `report_resource_usage` was only ever called with
`AttributionSource::Contract` (governance scoring) and from test code — never
with `AttributionSource::Peer` for the two bandwidth resource types
(`InboundBandwidthBytes` / `OutboundBandwidthBytes`) that the topology
manager actually uses.

`calculate_usage_proportion` iterates `ResourceType::all()` (exactly those two
bandwidth types). With no peer-attributed samples, `extrapolated_usage`
iterates an empty set and reports ~0% usage. So at/above `min_connections`,
`adjust_topology` always took the "usage below 50% → add connections" branch.
The load-aware hold (50–90%) and remove (>90%) paths — and
`select_connections_to_remove` — were effectively dead in production. An
overloaded node would keep adding connections instead of shedding load.

## Solution

Bridge the transport layer's existing per-peer wire-byte counters into the
topology meter on each `connection_maintenance` tick, just before
`adjust_topology` reads it (#3453):

- New free function `feed_peer_bandwidth_to_meter` in `ring.rs`.
- It snapshots `TRANSPORT_METRICS.per_peer_snapshot()` (cumulative
  sent/received bytes per peer `SocketAddr`), diffs against the previous
  tick's counts held in a `HashMap<SocketAddr, (u64, u64)>`, and feeds the
  per-tick byte delta to `report_resource_usage` as a `Peer` sample
  (received → `InboundBandwidthBytes`, sent → `OutboundBandwidthBytes`).
- Addresses that don't resolve to a ring `PeerKeyLocation` via
  `get_peer_by_addr` (transient/handshake connections) are skipped rather than
  attributed to a phantom source.
- `saturating_sub` guards against counter resets (per-peer slot eviction +
  re-creation) producing a bogus huge delta.
- The prior-tick map is pruned of peers absent from the latest snapshot, so it
  stays bounded by the transport metrics table (`MAX_TRACKED_PEERS`).

The diff/feed is a lock-only, non-blocking, no-`await` operation safe to run
inline on the maintenance tick (cf. `.claude/rules/channel-safety.md`). Uses
the injectable `TimeSource` so deterministic simulation is unaffected.

## Testing

New regression tests in `ring.rs` (`resource_meter_bridge_tests`):

- `bridge_feeds_meter_so_overloaded_node_removes` — drives the bridge with
  high per-peer bandwidth across a full sample window and asserts
  `adjust_topology` reaches `RemoveConnections`. This path was unreachable
  before the fix.
- `empty_meter_never_removes` — documents the bug: with an unfed meter, a node
  above `min_connections` never reaches the remove branch.
- `bridge_records_inbound_and_outbound_for_known_peer` — asserts the bridge
  attributes received bytes to `InboundBandwidthBytes` and sent bytes to
  `OutboundBandwidthBytes` for a resolvable peer.
- `bridge_skips_unresolvable_address` — traffic for a non-ring address is not
  attributed (no spurious removal), but the prior-tick map is still updated.
- `bridge_handles_counter_reset_without_underflow` — a backwards counter
  clamps to a finite rate.
- `bridge_prunes_departed_peers_from_prev` — the prior-tick map drops peers
  that leave the snapshot.

Verified the meter-feed remove test fails without the bridge wiring and passes
with it.

Closes #3453

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -2566,12 +2566,25 @@ fn resource_weight(resource: crate::topology::meter::ResourceType) -> f64 {
 ///
 /// `TRANSPORT_METRICS.per_peer_snapshot()` exposes *cumulative* sent/received
 /// byte counts per peer socket address. The topology meter, by contrast,
-/// expects per-sample byte counts that it accumulates into a rate over a
-/// sliding window. We therefore diff the current snapshot against the
-/// previous tick's counts (`prev`) to recover the bytes transferred during
-/// the tick, and feed that delta as one sample per direction:
+/// accumulates each reported sample into a sliding-window sum and divides by
+/// the elapsed wall-time (`now − oldest_sample`) to produce a rate. We diff
+/// the current snapshot against the previous tick's counts (`prev`) to recover
+/// the bytes transferred *during the interval*, and feed that delta as one
+/// sample per direction:
 ///   - received bytes → `InboundBandwidthBytes`
 ///   - sent bytes     → `OutboundBandwidthBytes`
+///
+/// `interval_start` MUST be the time of the *previous* maintenance tick, not
+/// the current one. The meter divides the windowed byte sum by
+/// `query_time − oldest_sample_time` with a 1-second floor
+/// (`RunningAverage::get_rate_at_time`). `adjust_topology` queries the meter at
+/// the current tick time immediately after this call, so timestamping the
+/// interval's whole byte delta at the interval START makes the very first
+/// sample divide by the real interval length (e.g. ~60s) instead of flooring
+/// to 1s. Timestamping at the interval END would treat a full interval's bytes
+/// as if they were transferred in a single second, overstating the rate by up
+/// to the tick interval (~60×) and triggering spurious connection removals
+/// under ordinary traffic (#3453 review).
 ///
 /// Only deltas with a resolvable `PeerKeyLocation` are reported: an address
 /// that doesn't (yet) map to a ring peer — e.g. a transient handshake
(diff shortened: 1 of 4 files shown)


sanity:
## Review (FULL tier — ring/topology/resource governance)

External model: **Codex** (`codex review --base origin/main`), run twice — once after the initial bridge code, once on the current HEAD (`b5a85592`) after addressing all findings. Final pass: **clean, no actionable findings.**

Gemini was attempted as a second external pass but its quota is exhausted (resets ~20h); per the multi-model-review policy the required independent non-Claude pass (Codex) succeeded, so this is recorded rather than blocking.

### Critical bug found and fixed during verification (stack overflow)

CI showed `Simulation` + `Unit & Integration` failing with a mass `SIGABRT`. Root cause (reproduced locally, bisected to this PR's bridge call):

`Rate::cmp` delegated to `partial_cmp`, and `Rate::partial_cmp` delegated back to `cmp` — **mutual infinite recursion → stack overflow** on the *first* `Rate` comparison. This was latent because `Rate` was never compared in production until the meter started being fed real samples (#3453): `calculate_estimated_usage_rate` sorts a `Vec<Rate>` via `sort_unstable`, which invokes `Rate::cmp`. Every simulation node aborted on its first maintenance tick.

Fix (`topology/rate.rs`): `Ord::cmp` now compares the inner `f64` directly via `total_cmp` (a total order consistent with `Eq`); `partial_cmp` delegates to `cmp`. Regression test `test_ord_does_not_recurse` covers `cmp`/`partial_cmp`/`<`/`>`/`sort_unstable` directly (would SIGABRT under the old impl).

### Codex findings (both addressed)

- **[P1] Rate inflation (~60×).** The maintenance tick reported a whole interval's byte delta timestamped at the interval END, then `adjust_topology` read the meter immediately; `RunningAverage` divides the windowed sum by `(query − oldest_sample)` with a 1s floor, so the first sample divided a ~60s delta by 1s → spurious connection removals under ordinary traffic. **Fixed:** track the previous tick time and timestamp each interval's delta at the interval START, so the meter divides by the real interval length. Regression test `bridge_timestamps_delta_at_interval_start_no_inflation`.

- **[P2] Unbounded meter / `source_creation_times` growth under churn.** Only the `prev` map was pruned; the meter and `source_creation_times` (iterated by `adjust_topology` every tick) accumulated a permanent entry for every peer that ever sent bytes. **Fixed:** added `TopologyManager::retain_peer_sources` / `Meter::retain_peer_sources`, pruning peer-attributed state to the live peer set each tick (Contract/Delegate sources retained). Regression test `bridge_prunes_departed_peer_from_meter`.

### CI determinism pin test

The new bridge tests add 8 test-only bare `Instant::now()` call sites; all production time reads in `connection_maintenance` go through `self.time_source.now()` (verified — zero bare `Instant::now()` in the production function body). Updated `EXPECTED_BARE_INSTANT_NOW` 6 → 14 per the pin test's own documented test-only path, with rationale comment.

### Verification

- `topology::rate::tests`, `topology::meter::tests`, `resource_meter_bridge_tests`, `instant_now_pin_test` — all pass (17/17 + pin).
- Six previously-aborting simulation tests (`test_topology_single_hoster`, `test_max_downstream_limit_reached`, `test_unhealthy_peer_eviction`, `test_six_peer_lifecycle_mock_wasm`, `test_isolated_node_rebootstraps_via_gateway`, `test_subscription_broadcast_propagation`) now pass locally.
- `cargo fmt` + `cargo clippy -- -D warnings` clean.
- Fix verified fail-before/pass-after for the overload-removal path and the recursion.

[AI-assisted - Claude]

PR #4456, opened 2026-06-16
fix(telemetry): emit subscribe_timeout event when a client subscribe gives up

### LINKED ISSUE #3445: Subscribe operations emit no telemetry on timeout
## Problem

When a subscribe operation times out (after 60s OPERATION_TTL), **no telemetry event is emitted**. The system has `subscribe_request`, `subscribe_success`, and `subscribe_not_found` telemetry events but no `subscribe_timeout` event.

This makes subscribe failures completely invisible on the telemetry dashboard. For example, the River container contract (`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) had 196 `subscribe_request` events and 0 outcome events — all 196 timed out silently.

## Root Cause

The `garbage_cleanup_task` in `op_state_manager.rs` identifies expired subscribe operations and calls `remove_subscribe_and_notify_timeout()`, which:
1. Removes the op from the DashMap
2. Reports routing failure to the isotonic model
3. Sends `NotifySubscriptionError` to the contract handler

But it **never emits a telemetry event**. The `notify_of_time_out` function in the tracing module is gated behind `cfg!(feature = "trace-ot")` and only applies to already-completed operations.

## Suggested Fix

Add a `subscribe_timeout` (or generic `operation_timeout`) telemetry event in the garbage cleanup path, emitted when a subscribe operation is GC'd due to TTL expiry. This should include the contract key, transaction ID, and elapsed time.

## Impact

- Dashboard shows subscribe requests with no outcomes, making it impossible to diagnose subscribe failures
- Same gap likely exists for GET, PUT, and UPDATE timeouts

[AI-assisted - Claude]

sanity:
## Problem

When a client-initiated SUBSCRIBE exhausts every candidate peer without ever receiving a terminal reply (every attempt timed out), the driver returned an `OperationError` but emitted **no telemetry**. The dashboard had `subscribe_request` / `subscribe_success` / `subscribe_not_found` events but no timeout outcome, so a timed-out subscribe left a `subscribe_request` with no paired outcome and the failure was invisible. The River container contract (`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) showed 196 `subscribe_request` events and 0 outcomes — all silent timeouts.

Issue #3445 originally pointed at `remove_subscribe_and_notify_timeout` in the old GC path. **That path no longer exists**: SUBSCRIBE now runs on a task-per-tx driver (`subscribe/op_ctx_task.rs`) that owns its own timeout reporting, and the GC sweep no longer removes per-op state. The timeout give-up therefore now lives in the originator driver's retry loop, which is where the new event is emitted.

## Solution

Add a `SubscribeTimeout` variant to the `SubscribeEvent` telemetry enum (appended after the existing variants, well past the reserved-discriminant range, so the change is wire-format additive). It carries the transaction id, requester, contract instance id, retry-round count, and elapsed time — the same shape as the existing `SubscribeNotFound` outcome.

Emit it from `drive_client_subscribe_inner`'s timeout-exhausted branch, keyed on `client_tx` (the same tx the originating `subscribe_request` was keyed on) so the request pairs with this outcome on the dashboard. Wire it through the telemetry name-mapping (`"subscribe_timeout"`) and JSON serialization so it is recorded like the other `SubscribeEvent` variants, and classify it as a (failure) subscribe outcome in `subscribe_outcome()`, `span_completed()`, the state-verifier terminal-resolution detector, and the simulation summary's failed counter.

## Testing

- `test_event_kind_to_json_subscribe_timeout` (`tracing/telemetry.rs`): asserts the event maps to the `subscribe_timeout` name, serializes every field, and is classified as a subscribe outcome. Without the new variant + arms this test does not compile.
- `drive_client_subscribe_inner_emits_timeout_telemetry_on_exhaustion` (`subscribe/op_ctx_task.rs`): source-scrape pin test asserting the timeout branch emits `NetEventLog::subscribe_timeout` and registers it into the event pipeline before returning. **Verified it FAILS when the emission is removed and PASSES with it.** This guards against the telemetry-counter-rot failure mode in `.claude/rules/bug-prevention-patterns.md`.

`cargo fmt` clean; `cargo clippy --locked -- -D warnings` clean.

Closes #3445

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -3210,7 +3210,8 @@ impl SimNetwork {
                     match sub_event {
                         SubscribeEvent::Request { .. } => summary.subscribe.requested += 1,
                         SubscribeEvent::SubscribeSuccess { .. } => summary.subscribe.succeeded += 1,
-                        SubscribeEvent::SubscribeNotFound { .. } => summary.subscribe.failed += 1,
+                        SubscribeEvent::SubscribeNotFound { .. }
+                        | SubscribeEvent::SubscribeTimeout { .. } => summary.subscribe.failed += 1,
                         SubscribeEvent::ResponseSent { .. }
                         | SubscribeEvent::HostingStarted { .. }
                         | SubscribeEvent::HostingStopped { .. }
--- a/crates/core/src/operations/subscribe/op_ctx_task.rs
+++ b/crates/core/src/operations/subscribe/op_ctx_task.rs
@@ -811,6 +811,22 @@ async fn drive_client_subscribe_inner(
                         continue;
                     }
                     None => {
+                        // #3445: emit a terminal telemetry event so a
+                        // timed-out client subscribe is no longer invisible
+                        // on the dashboard. Keyed on `client_tx` (the same
+                        // tx the originating `subscribe_request` was keyed
+                        // on) so the request pairs with this outcome.
+                        if let Some(event) = crate::tracing::NetEventLog::subscribe_timeout(
+                            &client_tx,
+                            &op_manager.ring,
+                            instance_id,
+                            retries + 1,
+                        ) {
+                            op_manager
+                                .ring
+                                .register_events(either::Either::Left(event))
+                                .await;
+                        }
                         return Ok(DriverOutcome::Publish(Err(ErrorKind::OperationError {
                             cause: format!(
                                 "subscribe to {instance_id} timed out after {} rounds",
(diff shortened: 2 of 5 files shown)


sanity:
## Multi-model review — FULL tier (subscribe op state machine + telemetry/wire-format enum)

**External model pass: Codex** (`codex review --base origin/main`, Codex v0.133.0 / `gpt-5.5` / openai). Ran successfully against commit `4327b187` (substantive diff is byte-identical to the current rebased HEAD `e0f447a8`; the rebase only absorbed unrelated `main` commits and kept both my test and main's new shadow-subbudget tests — no logic was merged).

### Codex verdict: clean, zero findings

> "I did not identify any discrete correctness issues introduced by the diff. The new subscribe timeout event is wired through event classification, telemetry JSON/name mapping, state verification, and simulation summaries."

### Independent verification against the two flagged risks

**Telemetry-serialization-sync (the `bug-prevention-patterns.md` counter-rot pattern) — all consumers updated:**

| Consumer | Location | Status |
|---|---|---|
| `event_kind_to_string` (OTLP name) | `telemetry.rs` → `"subscribe_timeout"` | ok |
| `event_kind_to_json` (full destructure, all 6 fields) | `telemetry.rs` | ok |
| `subscribe_outcome()` classification | `tracing.rs` → `Some(false)` (failure) | ok |
| `span_completed` (trace-ot terminal) | `tracing.rs` → `true` | ok |
| `contract_key()` match arm | `tracing.rs` → `None` (no full key on originator — correct) | ok |
| `hop_count()` | wildcard `_ => None`; variant has no `hop_count` — correct | ok |
| state_verifier detector | `detectors.rs` → marks tx resolved | ok |
| simulation summary counter | `testing_impl.rs` → `subscribe.failed += 1` | ok |

**Wire-format additivity — safe.** `SubscribeEvent` has no `#[repr]` / explicit discriminants. `SubscribeTimeout` is declared as the *last* variant (after `UnsubscribeReceived`); no existing variant was reordered, so existing bincode indices are unchanged and old stored events deserialize unchanged.

**Keying:** the event is emitted on `client_tx` (same tx as the originating `subscribe_request`), so the dashboard pairs the request with this outcome. Emission is `register_events(...).await` inside the per-tx driver task (not an event/recv loop), within channel-safety rules.

**Tests:** JSON round-trip test (`test_event_kind_to_json_subscribe_timeout`) + source-scrape pin test (`drive_client_subscribe_inner_emits_timeout_telemetry_on_exhaustion`). The pin test was verified to FAIL when the emission is removed and PASS with it, guarding against a future migration silently dropping the call.

No blocking findings. Proceeding to auto-merge once CI is green.

[AI-assisted - Claude]

PR #3854, opened 2026-04-13
fix(server): preserve modifier-key semantics in open_url shell bridge

sanity:
## Problem

PR #3852 fixed freenet/river#208 by routing cross-origin clicks from the sandboxed iframe through the `open_url` shell bridge. That fixed the CORS failures but collapsed every external link click to the same outcome regardless of user intent:

- Left-click → foreground new tab ✓
- Middle-click → **foreground new tab** ✗ (expected background tab)
- Ctrl-click / meta-click → **foreground new tab** ✗ (expected background tab)
- Shift-click → **foreground new tab** ✗ (expected new window)

The interceptor dropped the modifier state and the shell handler called `window.open` with no flags, so user intent was silently lost.

## Approach

Forward modifier state in the `open_url` postMessage payload: `button` (0/1 for left/middle), `ctrlKey`, `metaKey`, `shiftKey`, `altKey`. The shell `open_url` handler now branches:

- `shiftKey` → `window.open(url, '_blank', 'noopener,noreferrer,popup')` — popup-style new window.
- `button === 1 || ctrlKey || metaKey` → `window.open(url, '_blank', 'noopener,noreferrer')` — new tab. The browser decides foreground vs background based on its own heuristics; we can't force background from script without a direct user gesture, but the semantics are at least not silently dropped.
- Plain left-click → same as before: `window.open(url, '_blank', 'noopener,noreferrer')`.

Security: the existing `https:`-only and non-loopback hostname checks still apply first; modifier handling only kicks in after the URL is validated.

## Testing

Two new unit tests in `crates/core/src/server/path_handlers.rs`:

1. **Interceptor forwards modifier state** — asserts the cross-origin branch of `NAVIGATION_INTERCEPTOR_JS` includes `button`, `ctrlKey`, `metaKey`, `shiftKey` in the posted message.
2. **Shell bridge honours modifier state** — asserts the `open_url` branch of `SHELL_BRIDGE_JS` reads `msg.shiftKey`, `msg.ctrlKey`/`msg.metaKey`, and `msg.button`.

Both tests fail against the pre-fix strings and would catch a future refactor that silently drops the modifier plumbing.

```
cargo test -p freenet --lib server::path_handlers::tests
# 36 passed (34 existing + 2 new)
```

`cargo clippy -p freenet --lib -- -D warnings` clean.

### Why didn't CI catch this?

[...]

Issue #3853, opened 2026-04-13
navigation interceptor: preserve middle-click / modifier-key semantics for open_url


sanity:
PR #3852 reordered the sandboxed-iframe navigation interceptor so cross-origin clicks unconditionally go through the \`open_url\` shell bridge. That fixes freenet/river#208, but it means a middle-click, ctrl-click, shift-click, or meta-click on a cross-origin link now always produces the same behaviour as a normal click: a single \`window.open(url, '_blank', 'noopener,noreferrer')\` in the shell.

Users who habitually middle-click to background-tab a link, or shift-click to open in a new window, no longer get their expected behaviour. This was partly pre-existing for same-origin links but PR #3852 extends the regression to the cross-origin \`target=\"_blank\"\` case that previously fell through to native handling.

**Proposed fix:**

1. In \`NAVIGATION_INTERCEPTOR_JS\` (\`crates/core/src/server/path_handlers.rs\`), include modifier state in the postMessage payload:
   \`\`\`js
   window.parent.postMessage({
     __freenet_shell__: true,
     type: 'open_url',
     url: target.href,
     button: e.button,
     ctrlKey: e.ctrlKey,
     metaKey: e.metaKey,
     shiftKey: e.shiftKey,
   }, '*');
   \`\`\`
2. In the shell bridge \`open_url\` handler, honour the flags when calling \`window.open\`:
   - \`button === 1\` or \`(ctrlKey || metaKey)\` → open in a new background tab (browsers mostly handle this automatically when \`window.open\` has a target of \`_blank\` and the browser is told to background).
   - \`shiftKey\` → open in a new window (\`window.open(url, '_blank', 'noopener,noreferrer,popup')\` or similar).

3. Add a Playwright regression test covering at least middle-click and ctrl-click on a cross-origin link inside a sandboxed iframe (tracks with the same browser-harness gap as #3849 and #3852).

Low priority — no user is actively blocked — but worth doing before the next release note cycle.

[AI-assisted - Claude]

PR #3852, opened 2026-04-13
fix(server): route target=_blank cross-origin links through open_url bridge

sanity:
## Problem

@lukors reported in freenet/river#208 that following a link from a Freenet webapp to the regular web produced a flood of `Cross-Origin request blocked` errors and broke pages like GitHub's issues list:

> If I follow a link in River to a website on the regular web, I get a bunch of Cross-Origin errors in the console, and the website I land on does not work properly. For instance if I follow a link to a GitHub issue, and then click the "issues" link at the top of the page, the issues don't load.

Root cause: the sandboxed-iframe navigation interceptor in `crates/core/src/server/path_handlers.rs` skipped any anchor with an explicit non-`_self` target **before** checking whether the link was cross-origin:

```js
if (target.target && target.target !== '_self') return;   // ← short-circuit
...
if (target.origin !== location.origin) { ... open_url ... }
```

River (and any other Freenet webapp) rewrites external links to include `target="_blank"`, so every cross-origin click fell through the interceptor into the browser's default handling. Since #1499 removed `allow-popups-to-escape-sandbox` from the iframe sandbox, that default is a sandboxed popup with a null origin — which breaks CORS on the destination site.

## Approach

Reorder the classification so the origin check dominates:

1. Walk up to the anchor, sanity-check protocol/download/data-no-intercept.
2. **Compute origin.** If cross-origin → always `preventDefault` + `open_url` postMessage, regardless of the `target` attribute.
3. Only then: if same-origin and has an explicit non-`_self` target, let the browser handle it (multi-tab within a contract still works).
4. Otherwise same-origin in-contract navigation via `navigate` postMessage.

The shell page already opens `open_url` destinations via `window.open(url, '_blank', 'noopener')`, which gives the popup a proper (non-null) origin and makes CORS work normally.

This also means River doesn't need to stop emitting `target="_blank"`; the fix is pure server-side and every existing webapp benefits immediately.

## Testing

New unit test `navigation_interceptor_handles_cross_origin_target_blank` in `crates/core/src/server/path_handlers.rs`. It:
- locates the cross-origin check (`target.origin !== location.origin`) and the target-attribute skip (`target.target && target.target !== '_self'`) in the JS source,
- asserts the cross-origin check appears **first** in source order — pinning the contract and failing loudly if a future refactor reintroduces the bug,
- asserts the cross-origin branch contains `preventDefault` and `type: 'open_url'`.

```
cargo test -p freenet --lib server::path_handlers::tests::navigation
[...]

PR #3849, opened 2026-04-13
fix(server): allow same-origin connect-src on shell page CSP

sanity:
## Problem

Every Freenet webapp was logging a repeating CSP error in Firefox (reported by @lukors as #3842):

```
Content-Security-Policy: The page's settings blocked the loading of a resource
(connect-src) at http://127.0.0.1:7509/permission/pending because it violates
the following directive: "connect-src ws: wss:"
```

The shell page CSP at `crates/core/src/server/client_api.rs:203` had `connect-src ws: wss:`, which only allows WebSocket schemes. But `path_handlers::serve_webapp_html` injects a JavaScript poller that calls `fetch('/permission/pending')` every 3 seconds over HTTP on the same origin to detect new permission prompts. Every fetch tripped the CSP, console was spammed, and the permission-prompt overlay never rendered.

## Approach

Add `'self'` to `connect-src` so same-origin HTTP fetches work. Keep `ws: wss:` so the postMessage bridge can still open WebSocket connections on behalf of the sandboxed iframe. Keep `default-src 'none'` for defence in depth.

Also extract the CSP into a `SHELL_PAGE_CSP` constant so it's testable in isolation, and add a regression test that asserts connect-src contains `'self'`, `ws:` and `wss:`. The test fails against the pre-fix string and would have caught the bug at CI time.

## Testing

New unit test `shell_page_csp_allows_same_origin_and_websocket_connect` in `crates/core/src/server/client_api.rs`:

```bash
cargo test -p freenet --lib server::client_api::tests
# 5 passed (incl. the new test)
```

Verified against the pre-fix string the test fails with the expected message; verified it passes with the fix.

`cargo clippy -p freenet --lib -- -D warnings` clean.

### Why didn't CI catch this?

No test existed that exercised the CSP string shipped with the shell page. The runtime violation only manifests in a real browser against a running node, and the gateway test harness doesn't load the shell page through a CSP-enforcing browser. The new unit test closes the gap by pinning the string directly.

## Fixes

Closes #3842.

[AI-assisted - Claude]

PR #4457, opened 2026-06-16
test: add Playwright smoke-test harness for shell + iframe postMessage contract

### LINKED ISSUE #3856: tests: add Playwright smoke test harness for shell + iframe postMessage contract

sanity:
## Problem

The gateway shell bridge, navigation interceptor, and CSP are JavaScript
strings injected by `crates/core/src/server/path_handlers.rs` and
`client_api.rs`. Nothing in CI executes that JS in a real browser — every
existing test only substring-matches the emitted strings at the Rust level.
Four regressions shipped to production through that gap, each with a review
note saying "this should really be a browser test":

- **#3842** — shell CSP missing `connect-src 'self'` blocked the
  `/permission/pending` poller in every webapp (fix pinned by #3849).
- **#3852** — cross-origin `target="_blank"` links opened null-origin
  sandboxed popups and broke CORS on destination sites (freenet/river#208).
- **#3854** — middle-click (`auxclick`) and shift/modifier-key semantics on
  cross-origin links were silently dropped; middle-click wasn't intercepted
  at all (#3853 was the closed predecessor of this merged fix).

## Approach

Add a Playwright smoke-test harness that drives the injected shell/iframe JS
against headless Chromium:

- `crates/core/tests/playwright_shell.rs` — a `#[freenet_test]` harness that
  boots a single in-process gateway, publishes a fixture webapp via the real
  `fdev website publish` path (isolated `XDG_CONFIG_HOME`), waits for the
  shell route to serve, then runs the Playwright suite with the shell URL in
  `FREENET_SHELL_URL`. Node lifecycle stays on the well-tested in-process path,
  mirroring `fdev_publish_e2e.rs`. A single node suffices: it stores its own
  published contract, so the shell renders with no network propagation.
- `crates/core/tests/playwright/` — the browser layer: a minimal fixture
  webapp (cross-origin / same-origin / download links + a permission-poll
  fetch) and specs that assert the shell CSP, the sandbox attribute, the
  `open_url` bridge for left/middle/shift cross-origin clicks, the in-place
  navigate hop for same-origin links, browser-Back restoration (#3839), and
  the absence of CSP violations.
- `.github/workflows/playwright-shell.yml` — a path-filtered job
  (`crates/core/src/server/**`, the test files, and `crates/fdev/**` since the
  harness shells out to `fdev website publish`) that installs cached Playwright
  browsers and runs the harness with `FREENET_PLAYWRIGHT=1`.

The browser step is opt-in via `FREENET_PLAYWRIGHT=1`, so the default
`cargo nextest` run publishes the fixture and confirms the shell serves but
skips the browser suite — a checkout without Node/browsers never goes red.

Fixture WASM reuses the prebuilt website-container WASM embedded in fdev
(`crates/fdev/resources/website_contract.wasm`); no `wasm32-unknown-unknown`
build step, deterministic contract key per fdev version.

## Testing

Purely additive test infrastructure; no production code changes. Validated
end-to-end: 7/7 specs pass against a live node. The new workflow is
informational (not a required check) and path-filtered, so it does not run on
unrelated PRs. A plain `cargo nextest` run (without `FREENET_PLAYWRIGHT=1`)
still publishes the fixture and confirms the shell serves, so the Rust side
exercises `fdev website publish` even when browsers are absent.

Closes #3856

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/playwright-shell.yml
+++ b/.github/workflows/playwright-shell.yml
@@ -27,12 +27,16 @@ on:
       - 'crates/core/src/server/**'
       - 'crates/core/tests/playwright/**'
       - 'crates/core/tests/playwright_shell.rs'
+      # fdev is the publish path the harness shells out to; a regression in
+      # `fdev website publish` would break the harness, so run on fdev changes.
+      - 'crates/fdev/**'
       - '.github/workflows/playwright-shell.yml'
   pull_request:
     paths:
       - 'crates/core/src/server/**'
       - 'crates/core/tests/playwright/**'
       - 'crates/core/tests/playwright_shell.rs'
+      - 'crates/fdev/**'
       - '.github/workflows/playwright-shell.yml'
 
 concurrency:
--- a/.github/workflows/playwright-shell.yml
+++ b/.github/workflows/playwright-shell.yml
@@ -1,0 +1,135 @@
+name: Shell Playwright E2E
+
+# Browser smoke tests for the gateway shell + sandboxed iframe postMessage
+# contract (freenet/freenet-core#3856).
+#
+# Why a separate workflow: the shell bridge, navigation interceptor, and CSP
+# are JavaScript strings injected by crates/core/src/server/path_handlers.rs +
+# client_api.rs. Every other CI job only ever asserts substrings of that JS at
+# the Rust level; nothing executes it in a real browser. Four regressions
+# (#3842, #3852, #3853, #3854) shipped to production through that gap. This job
+# boots a node (via the #[freenet_test] harness in
+# crates/core/tests/playwright_shell.rs), publishes a fixture webapp, and drives
+# the injected JS against headless Chromium with Playwright.
+#
+# NOT a required check (yet): it is path-filtered to the server JS + the test
+# harness, so it does not run on every PR. A path-filtered workflow does not
+# reliably produce a check context for branch protection, so do NOT mark this
+# required without first restructuring it into an always-green gate + a
+# conditional real job (see the note in macos-dmg-swap-e2e.yml). Until then it
+# runs as an informational gate on PRs that touch the relevant surface and on
+# main.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'crates/core/src/server/**'
+      - 'crates/core/tests/playwright/**'
+      - 'crates/core/tests/playwright_shell.rs'
+      - '.github/workflows/playwright-shell.yml'
+  pull_request:
+    paths:
+      - 'crates/core/src/server/**'
+      - 'crates/core/tests/playwright/**'
+      - 'crates/core/tests/playwright_shell.rs'
+      - '.github/workflows/playwright-shell.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  shell-e2e:
+    name: Shell Playwright E2E
+    runs-on: ubicloud-standard-8
+    timeout-minutes: 30
+
+    env:
+      # Quiet logs; the harness emits its own progress at INFO via tracing only
+      # when something is wrong. A full DEBUG run produces GBs for a node boot.
+      RUST_LOG: warn
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      RUST_MIN_STACK: 268435456
+      FREENET_PLAYWRIGHT: "1"
+      # Reduce gateway connection backoff for CI (default 30s is too aggressive
+      # for localhost tests under CI resource pressure). See issue #3078.
+      FREENET_BACKOFF_BASE_SECS: "5"
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install mold linker
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y mold
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.93.0
+          targets: wasm32-unknown-unknown
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install nextest
+        run: curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
+
+      # fdev is the publish path the harness shells out to. Build it (and the
+      # test's own crate deps) before the test step so the binary is on disk.
+      - name: Build fdev
+        env:
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER: clang
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS: -C link-arg=-fuse-ld=mold
+        run: cargo build --bin fdev --locked
+
+      - name: Install Playwright deps
+        working-directory: crates/core/tests/playwright
+        run: npm ci
+
+      # Cache the downloaded browser binaries keyed on the pinned Playwright
+      # version so we only re-download when the version in package.json changes.
+      - name: Resolve Playwright version
+        id: pw
+        working-directory: crates/core/tests/playwright
+        run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@v4
+        id: pw-cache
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ steps.pw.outputs.version }}
+
+      - name: Install Playwright browsers
+        working-directory: crates/core/tests/playwright
+        run: npx playwright install --with-deps chromium
+
+      - name: Clean test directories
+        run: rm -rf /tmp/freenet /tmp/freenet-* 2>/dev/null || true
+
+      - name: Run shell Playwright E2E
+        env:
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER: clang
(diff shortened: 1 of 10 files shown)


sanity:
## PR review — Shell Playwright smoke-test harness (#4457)

**Reviewer path:** External model **Codex (gpt-5.5)** ran successfully via `codex review --base origin/main` (Light tier per `multi-model-review.md`: additive test infra). I also did an independent code-first cross-check of the CI workflow and the harness/spec logic against the real injected shell code. No Claude-subagent fallback was needed — the external pass ran.

**Recommendation:** Mergeable as-is. No blocking findings. One non-blocking nit (below) that is a pre-existing project-wide pattern, not a regression in this PR.

### Findings

**[non-blocking / nit] Linux-only harness registered unconditionally** — `crates/core/tests/playwright_shell.rs:302-310` (Codex P2)
The `#[freenet_test]` harness binds to varied `127.x.y.1` loopback addresses, which only work on Linux. On macOS (`127.0.0.1` only) `cargo test -p freenet --test playwright_shell` will fail at node startup. Codex flagged this as a candidate for `cfg`-gating.
*Assessment — non-blocking:* this is the **established convention for every `#[freenet_test]` integration test in the repo** (`fdev_publish_e2e.rs`, `operations.rs`, `error_notification.rs`, `connectivity.rs`, etc. — none `cfg`-gate for OS; verified by grep). macOS devs already cannot run any of these integration suites; CI runs them on Linux only. `cfg`-gating just this one would make it inconsistent with its siblings. The README already documents the Linux-only constraint. Fine to leave as-is, or gate it repo-wide in a separate cleanup if desired.

### CI workflow review (`.github/workflows/playwright-shell.yml`) — scrutinized with extra care

No blocking issues. Verified:

- **Cannot block merge.** Path-filtered `pull_request` + `push: main`, and crucially **no `merge_group:` trigger** (ci.yml *does* have one). The merge queue therefore never waits on this job, and the workflow is correctly documented as not-a-required-check. The contrast with `macos-dmg-swap-e2e.yml` (no path filter + `merge_group:` *because* it IS required) is internally consistent — the comment block at the top accurately explains why a path-filtered workflow shouldn't be marked required.
- **Toolchain/flags match ci.yml's `test_unit` job exactly:** `1.93.0`, `RUST_MIN_STACK: 268435456`, `FREENET_BACKOFF_BASE_SECS: "5"`, mold linker via `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_*`, and the feature set `trace,websocket,redb,wasmtime-backend,testing` with `--no-default-features`. `cargo build --bin fdev --locked` then `cargo nextest run --test playwright_shell` mirrors the build-then-test ordering used elsewhere.
- **Browser cache key** `playwright-${{ runner.os }}-${{ steps.pw.outputs.version }}` is resolved from the installed `@playwright/test` version — correct; re-downloads only on version bump. `npm ci` uses the committed `package-lock.json` (pinned `1.61.0`).
- **`timeout-minutes: 30`** is sane: the in-test budget is 600s and an observed local run was ~282s; 30 min covers a cold-cache `fdev` build + browser install + the suite.
- Failure-artifact upload is `if: failure()` with `if-no-files-found: ignore` — won't itself fail the job.

### Default (non-`FREENET_PLAYWRIGHT`) path — no hang/flake risk in normal CI

The workflow always sets `FREENET_PLAYWRIGHT=1`, so the browser suite runs there. On the *default* path (plain `cargo nextest`, e.g. a dev or the main `test_unit` job if it ever compiled this test), the harness publishes the fixture, polls the shell for up to 120s, and returns early — it does **not** launch a browser, so a missing Node/Chromium toolchain is never a failure. The 120s poll has a 500ms sleep + 10s per-request timeout and is bounded by the 600s test timeout; no unbounded wait.

### Logic cross-checks (all verified against the real injected code)

- **CSP assertions match** `client_api.rs:36` `SHELL_PAGE_CSP` (`frame-src 'self'; ... connect-src 'self' ws: wss:`).
- **postMessage payload shapes match** `path_handlers.rs`: `open_url` `{type, url, shiftKey: !!e.shiftKey}` (2039-2042), `navigate` `{type, href}` (2052), `auxclick` listener (2057), `popstate` handler (1437). The `#3842` permission-poll, the `id="app"` + `freenetBridge` readiness probe, and the `data-src`-deferred iframe load are all consistent with the served markup.
- **`fdev website init`** prints exactly `Your website contract key: <key>` (`website.rs:246`) — the harness parse prefix is exact; the base58 `is_ascii_alphanumeric` shape check is sound (base58 ⊂ alphanumeric). `XDG_CONFIG_HOME` isolation is honoured via `dirs::config_dir()` (XDG on Linux), so the signing key stays in the temp dir.
- **`fdev --node-url <ws> website publish`** arg order is valid: `node_url` is a flattened top-level `BaseConfig` arg (`config.rs:39`); `conflicts_with_all=[address,port]` only triggers on explicit values, so passing only `--node-url` is fine. `ws_url()` yields `ws://…` (correct) and the HTTP shell route reuses the same `ip:ws_port`.
- **`spawn_blocking(npx …)`** on the multi-thread runtime is appropriate for the one-shot blocking child; `await??` propagates both the join error and the suite failure.

[AI-assisted - Claude]

sanity:
## Re-review of new commit b59c7ccf (delta only)

The prior review is stale per `multi-model-review.md` because the code changed. I re-reviewed **only** the delta in `b59c7ccf` ("test: cover the download-attribute non-interception path"); the rest of the diff was already cleared in the [earlier review](https://github.com/freenet/freenet-core/pull/4457#issuecomment-4722313655) and is unchanged. The change is additive test-only (one new spec + a fixture href tweak), so I did an adversarial read rather than a fresh full external pass.

**Recommendation: still mergeable. No blocking findings on the new commit.**

### The new spec is correct and isolates the right guard

Verified against the actual interceptor in `path_handlers.rs::handleAnchorClick`, whose checks run in this order:
1. `:2011` — protocol skip (`javascript:`/`data:` and anything non-http(s)) → `return`
2. `:2013` — `if (target.hasAttribute('download')) return;` ← the guard under test
3. `:2025+` — cross-origin → `open_url`; else same-origin → `navigate`

The fixture href change from `data:text/plain,hello` → same-origin `page2.html` is **load-bearing and correct**: a `data:` URL would have been short-circuited at step 1, so the absence of a postMessage would NOT have isolated the `download` guard. With a same-origin href, the link *would* emit `navigate` if the guard were dropped — so the guard is the only thing that suppresses the message. Reasoning is sound.

### Not flaky — the negative assertion is deterministic

The crucial robustness question is whether anything other than the interceptor posts a `__freenet_shell__`-namespaced message to the top window on load (which would make `toEqual([])` flake). It does not:
- `captureShellMessages` filters strictly on `d.__freenet_shell__` (spec.ts:59).
- The only iframe→parent `window.parent.postMessage` calls carrying `__freenet_shell__: true` are the interceptor's `open_url` (`:2038`) and `navigate` (`:2051`) — exactly what the `download` early-return suppresses.
- The WebSocket shim's parent-directed messages (`:1902/:1914/:1921`) are namespaced `__freenet_ws__: true`, so they are never captured.
- The fixture's on-load `fetch('/permission/pending')` does not postMessage at all.

So after clicking a `download` link the capture buffer is reliably empty; the 300ms `waitForTimeout` gives a (regression-induced) interception a tick to land before asserting empty — the standard prove-a-negative pattern, used soundly here.

### Control-click + reload pattern correctly guards against a vacuous pass

- (1) Removing `download` from the same element and clicking → expects `navigate`: proves the listener is live and the selector is wired, so the later empty assertion can't pass merely because the click never reached a handler.
- (2) `page.goto(shellUrl)` is a full reload that restores the original DOM (with `download`) and tears down the in-iframe state; `captureShellMessages` re-arms a fresh buffer on the new top window *before* `fixtureFrame` resolves and well before the click — no race between listener install and message emission. The control's earlier `navigate` (which the shell honours by swapping the iframe to page2) is discarded by the reload, so it can't bleed into step 2's buffer.

No race in the reload + re-arm sequence; no other source of spurious `__freenet_shell__` traffic. Good edge-case coverage of the `download` early-return that the fixture previously documented but didn't exercise.

[AI-assisted - Claude]

PR #4458, opened 2026-06-16
feat(config): separate gateway host/port, default missing port to 31337

### LINKED ISSUE #1388: Refactor `gateways.toml` to separate hostname and port, fall back to default port if omitted
**Description:**  
We currently store a gateway’s hostname and port in one field (`hostname = "vega.locut.us:31337"`). This is confusing because it forces the port to be part of the “hostname.” Worse, if a port is omitted, the code sometimes falls back to a *random* port rather than 31337, which breaks connectivity.  

**Goals:**  
1. **Separate Host and Port Fields**  
   - It should be clearer to specify a gateway’s address with two fields, e.g.  
     ```toml
     [[gateways]]
     public_key = "keys/public.vega.gw.pem"
     address = { host = "vega.locut.us", port = 31337 }
     ```  
     This means no more embedding the port in a single hostname string.  
2. **Maintain Backward Compatibility**  
   - Existing configs that do something like `hostname = "vega.locut.us:31337"` should still parse correctly, so we don’t break existing deployments.  
   - If no port is specified, we should default to 31337 rather than picking a random port.  
3. **Enum Handling**  
   - Our Rust code uses an `enum Address { Hostname(String), HostAddress(SocketAddr) }`. We can either:  
     - Switch to a small struct with `{ host, port }`, plus a separate field for IP addresses; **or**  
     - Extend the enum to parse missing ports as default 31337 and treat them correctly.  
4. **Clarify Behavior**  
   - For a gateway, a missing port is invalid or defaults to 31337—no more random picking.  
   - Possibly rename the existing “hostname” variant to something more accurate (“host” or “host:port”).  

**Discussion Points:**  
- How best to parse older `gateway.toml` files without forcing everyone to rewrite them.  
- Whether we require a port for gateway usage or just default to 31337 for everything.  
- Any side effects with local vs. remote addresses.  

**Proposed Next Steps:**  
1. Implement a safe fallback for missing ports (31337 instead of random).  
2. Allow the new format (`address = { host = "vega.locut.us", port = 31337 }`) while still parsing the old format.  
3. (Optional) Migrate code away from the `Hostname(String)` enum variant to something more robust.  

This change will remove confusion around “hostname:port” vs. “port is random,” while letting older configs remain valid.

sanity:
## Problem

A gateway address in `gateways.toml` packs the host and port into a single string (`hostname = "vega.locut.us:31337"`), which is confusing. Worse, when a gateway host is given without a port, `parse_socket_addr` defaulted the port via `default_network_api_port()` — which returns a **random** free local port, not the gateway's real listening port — so the gateway became unreachable. (Issue #1388.)

## Solution

- Add `DEFAULT_GATEWAY_PORT = 31337` and use it (a fixed literal, never a random port) wherever a gateway address omits an explicit port: the no-port `Hostname` path and the DNS-resolver fallback in `parse_socket_addr`. The local network-listener default (`default_network_api_port`, intentionally random-available) is left untouched — these were two different concepts that happened to share one helper.
- Add a new `Address::Host { host, port }` variant and the matching TOML form:
  ```toml
  [[gateways]]
  public_key = "keys/public.vega.gw.pem"
  address = { host = "vega.locut.us", port = 31337 }   # port optional, defaults to 31337
  ```
- Hand-written `Serialize`/`Deserialize` for `Address` so all three table shapes are accepted and emitted as flat tables (a hand-written impl rather than `#[serde(untagged)]` keeps error messages precise and lets `port` default cleanly).

## Backward compatibility

This is the high-risk part of the change, so it's covered explicitly. The deserializer still accepts **both** legacy forms exactly as deployed in `https://freenet.org/keys/gateways.toml` today:

- `hostname = "host[:port]"`  — the current production format
- `host_address = "ip:port"`

plus the new `host`/`port` table form. Serialization of the legacy variants is unchanged (`hostname = "..."` / `host_address = "..."`), so config files written by this build remain readable by older binaries. `port` without `host`, an empty address table, and conflicting forms are rejected with clear error messages.

## Testing

- **config (`config::tests::test_address_*`)**: legacy `hostname` string with and without a port, legacy `host_address`, new `host`+`port`, new `host` with a non-default port, new `host` defaulting to 31337, error cases (port-without-host, empty table, conflicting forms), and a `Host` serialize→deserialize round-trip.
- **node (`node::tests`)**: `parse_socket_addr` for the new `Host` variant (default and explicit port); the port-less `localhost` regression test now asserts the resolved port is **31337** instead of the old random `> 1024` (the old assertion was encoding the buggy behavior).

`cargo fmt --check`, `cargo clippy --locked -- -D warnings`, and the touched test modules all pass locally.

Closes #1388

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3522,10 +3522,53 @@ mod tests {
             }],
         };
         let serialized = toml::to_string(&gateways).unwrap();
+        // The `Host` variant must serialize as a FLAT table (host/port as
+        // sibling keys), matching the new wire form in the issue — not nested
+        // under a `[gateways.address.host]` sub-table (the derived enum form).
+        assert!(
+            serialized.contains("host = \"vega.locut.us\"") && serialized.contains("port = 31337"),
+            "unexpected serialized form:\n{serialized}"
+        );
+        assert!(
+            !serialized.contains("[gateways.address.host]"),
+            "Host variant must not nest under its own sub-table:\n{serialized}"
+        );
         let deserialized: Gateways = toml::from_str(&serialized).unwrap();
         assert_eq!(
             deserialized.gateways[0].address,
             gateways.gateways[0].address
+        );
+    }
+
+    /// Pin the legacy serialized wire forms so a future refactor can't silently
+    /// change what we write to `gateways.toml` (old binaries must keep reading
+    /// files this build writes).
+    #[test]
+    fn test_address_legacy_variants_serialize_unchanged() {
+        let hostname = Gateways {
+            gateways: vec![GatewayConfig {
+                address: Address::Hostname("vega.locut.us:31337".to_string()),
+                public_key_path: PathBuf::from("keys/k.pem"),
+                location: None,
+            }],
+        };
+        let s = toml::to_string(&hostname).unwrap();
+        assert!(
+            s.contains("hostname = \"vega.locut.us:31337\""),
+            "legacy hostname form changed:\n{s}"
+        );
+
+        let host_addr = Gateways {
+            gateways: vec![GatewayConfig {
+                address: Address::HostAddress("203.0.113.1:31337".parse().unwrap()),
+                public_key_path: PathBuf::from("keys/k.pem"),
+                location: None,
+            }],
+        };
+        let s = toml::to_string(&host_addr).unwrap();
+        assert!(
+            s.contains("host_address = \"203.0.113.1:31337\""),
+            "legacy host_address form changed:\n{s}"
         );
     }
 
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -45,6 +45,15 @@ pub const DEFAULT_RANDOM_PEER_CONN_THRESHOLD: usize = 7;
 /// (if it applies, e.g. connect requests).
 pub const DEFAULT_MAX_HOPS_TO_LIVE: usize = 10;
 
+/// Default UDP port a gateway listens on.
+///
+/// Used as the fallback when a gateway address in `gateways.toml` specifies a
+/// host without an explicit port. This is a fixed, well-known value (NOT a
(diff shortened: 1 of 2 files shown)


sanity:
## Multi-model review (FULL tier — config/bootstrap + backward-compat surface)

**External pass — Codex (`codex review --base origin/main`):** ran twice (initial commit + after the additive wire-format-pin test commit). Both passes clean:

> No actionable correctness issues were found in the diff. The changes preserve legacy gateway address forms while adding the new host/port representation and fixed default gateway port behavior.

(External models were available — quota reset had passed — so no Claude-lens substitution was needed. Claude lenses were still applied on top, per FULL tier.)

**Claude lenses (skeptical / testing / code-first / big-picture):** no blocking findings.

- Backward compat verified against the live production format (`hostname = "host:port"` from `https://freenet.org/keys/gateways.toml`) and the `host_address` form — both still parse, and both still serialize byte-identically (pinned by `test_address_legacy_variants_serialize_unchanged`).
- `Address` still derives `Hash`/`Eq`, so `Gateways::merge_and_deduplicate`'s `HashSet<Address>` is unaffected.
- Manual `Serialize` uses string map keys only → JSON-safe, and TOML round-trip is pinned.
- The local network-listener default (`default_network_api_port`, intentionally random-available) is deliberately left untouched; only the gateway-target default moved to the fixed `DEFAULT_GATEWAY_PORT = 31337`.

**One noted nuance (not a blocker, not a regression):** a bare IPv6 literal in the new `host = "::1"` field would not parse via `to_socket_addrs`/DNS — but that limitation is identical to the existing legacy `hostname` path, which also never accepted bare IPv6. IPv6 gateways should use the bracketed `host_address = "[::1]:31337"` form (parsed by `SocketAddr`). No behavior change here.

Local: `cargo fmt --check` clean, `cargo clippy --locked -- -D warnings` clean, config module 80/80 + node module 66/66 tests pass.

[AI-assisted - Claude]

PR #2254, opened 2025-12-09
fix(test): use port retry instead of serial execution for blocked_peers tests

sanity:
## Problem

The blocked_peers tests were failing intermittently due to port collision race conditions. When multiple tests run in parallel, the pattern of "bind to port 0, get port, drop listener, node binds to that port" creates a window where another process could grab the port.

## Previous Approach

The original fix added `#[serial]` attributes to force tests to run sequentially. This worked but slowed down CI unnecessarily.

## This Solution

Per @iduartgomez's feedback, this PR replaces serial execution with a retry mechanism:

- Wrap test setup in a retry loop (max 5 attempts)
- Detect port collision errors (EADDRINUSE: os error 98 on Linux, 48 on macOS)
- On collision, log a warning and retry with new randomly-allocated ports
- Non-port-related errors fail immediately without retry
- Small delay (100ms) between retries to let ports be released

This allows tests to run in parallel while handling the inherent race condition gracefully.

## Testing

- Local compilation verified
- CI will validate the approach

## Fixes

Addresses test flakiness for:
- `test_ping_blocked_peers`
- `test_ping_blocked_peers_simple`
- `test_ping_blocked_peers_solution`

[AI-assisted - Claude]

PR #2252, opened 2025-12-09
fix: increase WebSocket client message size limit to 100MB

sanity:
## Problem

Test `test_ping_blocked_peers` was failing with:
```
Space limit exceeded: Message too long: 21697171 > 16777216
```

The server-side WebSocket (in `crates/core/src/client_events/websocket.rs`) is configured for 100MB messages, but all WebSocket clients were using `tokio_tungstenite::connect_async()` which defaults to tungstenite's 16MB limit.

When contract state grew beyond 16MB, clients couldn't receive the response.

## Solution

- Add `ws_config()` helper that configures 100MB message limit and 16MB frame limit
- Update all WebSocket connections to use `connect_async_with_config()` with this config
- Apply to both the ping client library and all test files

## Changes

- `apps/freenet-ping/app/src/ping_client.rs` - Updated `connect_to_host()`
- `apps/freenet-ping/app/tests/common/mod.rs` - Added shared `ws_config()` helper
- Updated all test files that create WebSocket connections

## Test Plan

- [x] `cargo check -p freenet-ping-app --tests` passes
- [ ] CI should now pass for `test_ping_blocked_peers*` tests
- [ ] Dependabot PRs should pass after rebasing

[AI-assisted - Claude]

PR #4460, opened 2026-06-16
feat: reject debug-compiled WASM contracts at PUT time

### LINKED ISSUE #2257: feat: reject debug WASM contracts at publish time
## Problem

Debug-compiled WASM contracts are ~65x larger than release builds (e.g., 12MB vs 186KB for the ping contract). When clients attempt to PUT these contracts, they can exceed WebSocket message size limits, causing confusing "Message too long" errors.

This was discovered when test contracts were accidentally being compiled in debug mode, producing 12MB WASM files that serialized to ~21MB over WebSocket.

## Proposed Solution

Add validation in the contract PUT handler to detect debug WASM and return a clear error message.

### Detection Method

Debug WASM files contain `.debug_*` custom sections:
- `.debug_abbrev`
- `.debug_info`
- `.debug_ranges`
- `.debug_str`
- `.debug_line`
- `.debug_loc`

Release builds have none of these sections.

### Implementation

1. Use `wasmparser` (already available via wasmer) to scan custom sections
2. Check if any custom section name starts with `.debug_`
3. If found, return an error like:
   ```
   Contract appears to be compiled in debug mode (contains .debug_* sections).
   Debug WASM is typically 10-100x larger than release builds.
   Please compile with --release for production use.
   ```

### Alternative: Warning vs Error

Could make this a warning instead of an error, logged server-side, to avoid breaking existing workflows. However, an error at publish time is cleaner since debug contracts shouldn't be deployed to production.

### Where to Add

The validation should be in `crates/core/src/client_events/mod.rs` or `crates/core/src/contract/handler.rs` where PUT requests are processed.

## Context

- PR #2254 fixed tests to use release builds
- PR #2252 (merged) increased WebSocket limit to 100MB as a workaround
- This issue would prevent the problem at the source

[AI-assisted - Claude]

sanity:
## Problem

Debug-compiled WASM contracts are ~65x larger than release builds (e.g.
~12MB vs ~186KB for the ping contract). When a client PUTs one of these,
the serialized message can exceed WebSocket size limits and surface as a
confusing "Message too long" transport error rather than something
actionable. PR #2252 raised the WS limit to 100MB as a workaround and
PR #2254 fixed tests to use release builds, but nothing prevented the
problem at the source.

## Approach

Reject debug-compiled contracts at PUT time, before the contract is
routed, with an actionable error.

- Add `contains_debug_sections(wasm: &[u8]) -> bool` (and the underlying
  `debug_sections` helper) in `client_events.rs`. It scans the WASM
  custom sections with `wasmparser` and flags any section whose name
  starts with `.debug_` (the DWARF sections the Rust debug profile
  leaves behind: `.debug_info`, `.debug_str`, `.debug_line`, etc.).
- In the `ContractRequest::Put` arm of `process_open_request`, run the
  check right after the transaction is registered with the result router
  and before `start_client_put`. On a hit, surface the rejection through
  the existing `report_op_init_error` path so the client receives a
  typed `ErrorKind::OperationError` whose cause tells the user to
  recompile with `--release`.
- The error is carried by a new `ContractError::DebugWasmRejected`
  variant, which flows through `report_op_init_error`'s existing
  `OpError::ContractError(_)` arm with no other match changes.

Design choices:
- A module that fails to parse is treated as "no debug sections" — this
  scan must not reject a contract merely because the lightweight parser
  disagrees with the full validator; malformed-WASM rejection remains
  the executor's job.
- `wasmparser` is added as a direct dependency (not gated on the
  `wasmtime-backend` feature) so the check runs regardless of which WASM
  execution backend is compiled in. It pins to `0.248`, the version
  wasmtime 45 already resolves transitively, so cargo reuses one copy.

## Testing

New unit tests in `client_events::test`:
- `debug_wasm_is_detected` — a module carrying `.debug_info` is flagged.
- `all_dwarf_sections_are_detected` — every DWARF section the issue
  enumerates (`.debug_abbrev/info/ranges/str/line/loc`) is caught.
- `release_wasm_passes` — a release build (no custom sections) is
  accepted.
- `non_debug_custom_sections_pass` — non-debug custom sections (`name`,
  `producers`, `target_features`, `.llvmcmd`) are not misclassified.
- `malformed_wasm_is_not_flagged_as_debug` — empty/garbage/truncated
  input is treated as "no debug sections", not rejected.
- `debug_wasm_rejection_yields_actionable_client_error` — the rejection
  maps to a client error whose cause mentions "debug mode" and
  "--release".

All 6 pass; the full `client_events` suite (96 tests) is green.
`cargo fmt` and `cargo clippy -- -D warnings` are clean.

Closes #2257

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2072,6 +2072,7 @@ dependencies = [
  "turmoil",
  "ulid",
  "url",
+ "wasmparser 0.248.0",
  "wasmtime",
  "which",
  "winapi",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,6 +109,11 @@ wasmtime = { version = "45", default-features = false, features = [
     "gc", "gc-drc", "gc-null", "parallel-compilation", "pooling-allocator",
     "profiling", "runtime", "std", "threads", "wat",
 ] }
+# Lightweight WASM section parser used to reject debug-compiled contracts at
+# PUT time (see crates/core/src/client_events.rs::contains_debug_sections).
+# Pinned to the same minor wasmtime 45 already resolves transitively so cargo
+# reuses one copy rather than locking a second wasmparser build.
+wasmparser = "0.248"
 
 # Testing/simulation
 arbitrary = { version = "1", features = ["derive"] }
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -83,6 +83,9 @@ tower-http = { workspace = true }
 ulid = { workspace = true }
 zeroize = { workspace = true }
 wasmtime = { workspace = true, optional = true }
+# Always enabled (not gated on wasmtime-backend): debug-section rejection at
+# PUT time must run regardless of which WASM execution backend is compiled in.
+wasmparser = { workspace = true }
 xz2 = { workspace = true }
 reqwest = { workspace = true }
 x25519-dalek = { workspace = true }
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -2571,6 +2571,28 @@ pub(crate) mod test {
             assert!(!contains_debug_sections(&truncated));
         }
 
+        /// A `.debug_*` section followed by a parse error must NOT leave the
+        /// scan reporting the already-collected debug section (Codex P3): a
+        /// module we can't fully parse is not trustworthy evidence of a
+        /// debug build, and rejecting it with "recompile with --release"
+        /// would mask the real malformed-WASM error. The whole scan must
+        /// void to empty on any parse error.
+        #[test]
+        fn debug_section_before_parse_error_is_not_flagged() {
+            // A complete, valid `.debug_info` custom section...
+            let mut wasm = wasm_with_custom_section(".debug_info", &[0xde, 0xad]);
+            // ...then a second custom section header that lies about its
+            // length, so the parser errors AFTER the debug section is seen.
+            wasm.push(0x00); // custom section id
+            wasm.push(0x40); // claims 64 bytes follow, but none do
+            assert!(
+                !contains_debug_sections(&wasm),
+                "a parse error after a debug section must void the scan, \
+                 not report the partial result"
+            );
+            assert!(debug_sections(&wasm).is_empty());
+        }
+
         /// The rejection maps through `report_op_init_error`'s
         /// `OpError::ContractError(_)` arm to a client-visible
         /// `ErrorKind::OperationError` whose cause carries the actionable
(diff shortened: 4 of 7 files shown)


sanity:
## Review summary — all findings addressed

Independent review: **Codex** (`codex review --base origin/main`, external non-Claude model) — two rounds — plus an adversarial Claude `skeptical-reviewer` pass. The review is per-content: I re-ran Codex after each material change (the local-mode fix, then the relay-path fix + rebase).

### Round 1 — Codex (clean) + skeptical-reviewer (Medium: local-mode bypass)

The original change only guarded the network client path (`process_open_request`). The skeptical reviewer found that **local-node mode** (`run_local_node`, the local server loop, `preload`, `handle_request`) routes `ContractRequest::Put` straight to `executor.contract_requests(...)` and never hit it — and local development is the most likely place to hand a node a debug build.

Fixed (not scoped out, per finish-the-fix):
- Moved `contains_debug_sections` / `debug_sections` to `crates/core/src/contract.rs` as `pub(crate)` so all paths share one implementation.
- Added the guard at the top of the `ContractRequest::Put` arm in `contract_requests` (`contract/executor/runtime.rs`), before `perform_contract_put`, with the same actionable "recompile with --release" message.
- Pin test `contract_requests_put_rejects_debug_wasm_before_perform_put`.

### Round 2 — Codex (P2: relay/network storage path bypass)

Codex's re-review then flagged that **relayed / network-originated** PUTs still bypass both guards: forwarded PUTs go `relay_put_store_locally` → `operations::put::put_contract` → contract handler, never touching `process_open_request` or `contract_requests`. A malicious or mixed-version peer could still get this node to store and re-forward a debug build.

Fixed:
- Added the same guard at the top of `operations::put::put_contract` (the shared storage chokepoint for the relay path), returning `OpError::ContractError(DebugWasmRejected)` before `notify_contract_handler`.
- Pin test `put_contract_rejects_debug_wasm_before_storing`.

### Net result — rejection enforced on every PUT entry point

| Path | Guard site | Rejects before |
|------|-----------|----------------|
| Network client (publish) | `process_open_request` | routing onto the wire |
| Local-node client | `contract_requests` | local storage/validation |
| Relay / forwarded PUT | `put_contract` | storing + re-forwarding |

### Nits (acknowledged, not changed)

- **Double scan on the rejection path** (`contains_debug_sections` then `debug_sections` for names) — rejection path only, zero hot-path cost; left for readability.
- **Client-error test mirrors the match arm** rather than calling `report_op_init_error` — matches the established pattern of the adjacent `op_error_*` pin tests (which replicate the arm to avoid a full `OpManager` fixture).

Tests: 6 detection unit tests + 3 source-scrape pin tests, all green. `cargo fmt` and CI-equivalent `cargo clippy --locked -- -D warnings` clean.

[AI-assisted - Claude]

PR #3818, opened 2026-04-10
feat: browser-based permission prompts with iframe security hardening

sanity:
## Problem

The delegate permission prompt system (from #3811) used a subprocess approach (zenity/kdialog/GTK native dialogs) that:
1. **Didn't work from systemd services** -- no DISPLAY access
2. **Required system dependencies** -- GTK3, zenity, or kdialog may not be installed
3. **Had a critical security gap**: `allow-popups-to-escape-sandbox` in the gateway iframe let malicious web apps open popups that escape the sandbox, gain `localhost:7509` origin, and forge delegate operations or access other apps' data

### Browser-based permission prompts (zero dependencies)

Replace subprocess prompting with browser-based prompts served by the existing HTTP dashboard:

1. `DashboardPrompter` generates a 128-bit cryptographic nonce and stores the pending prompt
2. Gateway shell page JS (trusted, outside iframe) polls `GET /permission/pending`
3. Shell page shows a browser Notification when a prompt arrives
4. User clicks notification, opens `/permission/{nonce}` in a new tab (trusted page)
5. Permission page shows verified delegate context (key hash, contract ID) + canonical Allow/Deny buttons
6. POST response flows back to the delegate via oneshot channel
7. 60s timeout auto-denies

### Iframe security hardening

**Remove `allow-popups-to-escape-sandbox`** from the gateway iframe sandbox. This was added in #3613 so external links from web apps would work (popups need proper origin for CORS). The fix: add an `open_url` shell bridge message so the trusted shell page opens external URLs instead of the iframe's popup.

This closes a pre-existing vulnerability where ANY web app could escape the sandbox via a user-gesture-triggered popup.

### CSRF protection

- 128-bit cryptographic nonce (not guessable u32)
- Origin header validation (reject null/foreign)
- JSON content-type enforcement
- X-Frame-Options: DENY + CSP frame-ancestors: none
- Atomic nonce removal (single-use)

## Testing

- All 2161 lib tests pass (0 failures)
- 12 user_input tests (prompter implementations, parsing, nonce generation, rate limiting)
- 18 path_handlers tests (sandbox assertions, open_url bridge)
- Red-teamed by Codex (GPT-5.4) across 3 iterations, addressing all Critical/High findings

## Breaking Change

**freenet/river#203**: River's `target="_blank"` external links need to be migrated to use the `open_url` shell bridge message. Links will break (CORS failures on target sites) until River is updated.

Closes #1499

[AI-assisted - Claude]

PR #4461, opened 2026-06-16
test: add sanitization boundary and invalid-index handler tests for permission prompts

### LINKED ISSUE #3821: test: add boundary and handler-level tests for permission prompt system
## Problem

The permission prompt system (from #3818) has test gaps identified by rule review:

1. **Sanitization boundary tests**: `MAX_MESSAGE_LEN` (2048), `MAX_LABEL_LEN` (64), and `MAX_LABELS` (10) truncation limits are enforced in production but not tested at the boundary (message with exactly 2049 chars, label with 65 chars, 11 labels).

2. **HTTP handler error path tests**: The `permission_respond` handler has 4 error paths (missing Origin -> 403, untrusted Origin -> 403, expired nonce -> 404, invalid index -> 400) that are not tested at the handler level. The underlying `is_trusted_origin` function IS unit-tested, but the full HTTP request/response cycle is not.

3. **`is_trusted_origin` edge case**: `"http://localhost"` (no port) is not tested despite being handled by the code.

## Proposed Fix

Add tests using axum's `Router` + `tower::ServiceExt::oneshot` for handler-level testing:

```rust
let app = Router::new()
    .merge(permission_prompts::routes())
    .layer(Extension(pending_prompts));

let response = app.oneshot(
    Request::builder()
        .method("POST")
        .uri("/permission/fake-nonce/respond")
        .header("content-type", "application/json")
        .header("origin", "http://evil.com")
        .body(Body::from(r#"{"index": 0}"#))
        .unwrap()
).await.unwrap();

assert_eq!(response.status(), StatusCode::FORBIDDEN);
```

## Context

Rule review warnings from #3818, acknowledged with `/ack` to avoid blocking the security-critical sandbox fix.

[AI-assisted - Claude]

sanity:
## Problem

The permission prompt system (#3818) has two test gaps flagged by rule review (acknowledged with `/ack` in #3818 to avoid blocking the security-critical sandbox fix, tracked in #3821):

1. **Sanitization-cap boundaries are untested.** The producer-side display caps in `contract/user_input.rs` — `MAX_MESSAGE_LEN` (2048), `MAX_LABEL_LEN` (64) and `MAX_LABELS` (10) — are enforced in production (`parse_message` / `parse_button_labels`) but never exercised at their boundaries. The existing `OVERLAY_*` cap tests live in `permission_prompts.rs` and cover a *different* layer (the JSON/HTML overlay-rendering boundary), so the producer-side caps had no coverage.

2. **The `permission_respond` invalid-index path was untested.** Of the handler's four error arms — missing Origin to 403, untrusted Origin to 403, expired/unknown nonce to 404, and out-of-range index to 400 — only the 400 (`index >= labels.len()`) arm lacked a handler-level test. The 403 and 404 arms are already covered.

## Approach

Test-only addition. **No production code changes.**

- `user_input.rs`: for each of the three caps, an at-limit test (passes through unchanged) and an over-limit test (truncated/dropped to the cap), plus a char-based truncation test for the message cap so a future refactor that moved `.take(MAX_MESSAGE_LEN)` past the control-char `.filter()` would be caught.
- `permission_prompts.rs`: `test_respond_invalid_index_returns_400_and_keeps_prompt` asserts an out-of-range index returns 400 with the `"invalid index"` body **and** leaves the prompt pending — pinning the validate-before-remove contract that lets a user retry after a bad index.

Tests follow the existing patterns in each module (`make_test_request`, `insert_prompt`, direct `permission_respond` invocation).

## Testing

- `cargo test -p freenet --lib user_input::tests` -> 25 passed (including the 7 new boundary tests).
- `test_respond_invalid_index_returns_400_and_keeps_prompt` -> passed.
- `cargo fmt` clean; clippy clean on the changed files.

Closes #3821

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/user_input.rs
+++ b/crates/core/src/contract/user_input.rs
@@ -494,6 +494,89 @@ mod tests {
         assert_eq!(labels, vec!["Valid", "Option 2"]);
     }
 
+    // ----- Sanitization-cap boundary tests (issue #3821) -----
+    //
+    // `parse_message` / `parse_button_labels` enforce the producer-side
+    // display caps MAX_MESSAGE_LEN (2048), MAX_LABEL_LEN (64) and
+    // MAX_LABELS (10). The existing tests above cover control-char
+    // stripping and invalid-UTF-8 fallback but never exercise the length
+    // caps at their boundaries. These tests pin the exact at-limit /
+    // over-limit behaviour so a future cap change (or a refactor that
+    // moves the `.take()` before/after the `.filter()`) is caught.
+
+    /// A message exactly at MAX_MESSAGE_LEN chars passes through untouched.
+    #[test]
+    fn test_parse_message_at_max_len_unchanged() {
+        let msg = "a".repeat(MAX_MESSAGE_LEN);
+        let req = make_test_request(&msg, vec![]);
+        let parsed = parse_message(&req);
+        assert_eq!(parsed.chars().count(), MAX_MESSAGE_LEN);
+        assert_eq!(parsed, msg);
+    }
+
+    /// A message one char over MAX_MESSAGE_LEN is truncated to the cap.
+    #[test]
+    fn test_parse_message_over_max_len_truncated() {
+        let msg = "a".repeat(MAX_MESSAGE_LEN + 1);
+        let req = make_test_request(&msg, vec![]);
+        let parsed = parse_message(&req);
+        assert_eq!(parsed.chars().count(), MAX_MESSAGE_LEN);
+    }
+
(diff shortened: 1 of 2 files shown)


sanity:
## Review: Codex (external, non-Claude) — Light tier

This is a test-only change (boundary + handler-level coverage for the permission prompt system; no production code touched), so per the risk tiers it warrants one external-model pass.

`codex review --base origin/main` result:

> The changes add focused regression/boundary tests without altering production behavior. I did not identify any introduced correctness or maintainability issues in the diff.

No findings to address. Independent (non-Claude) review is clean.

[AI-assisted - Claude]

PR #3093, opened 2026-02-17
fix: refresh interest TTL on successful broadcast send

sanity:
## Problem

Telemetry shows 49% of River room subscribers who successfully subscribed never received update broadcasts while online (#3046). The funnel analysis across ~850 peers revealed 77 out of 158 subscribed peers missed broadcasts.

**Root cause:** In `handle_broadcast_state_change`, after successfully sending a broadcast to a subscriber, the peer's interest TTL is only refreshed when sending deltas (via `update_peer_summary`). For full-state sends, the TTL is never refreshed.

Most new subscribers start with `summary: None` (set at subscription time), so they always receive full state. Their interest entries silently expire after `INTEREST_TTL` (5 min), dropping them from future broadcast target lists — even though broadcasts were being successfully delivered to them.

The only mechanism keeping interest alive was subscription renewal (every 2 min), but renewals can:
1. Route to a different contract holder (routing changes between intervals)
2. Be silently dropped under channel backpressure (non-blocking send path)
3. Hit the per-interval batch limit

This explains the 30 peers with >5 min activity gap (TTL expired, renewal missed) and likely contributes to the 25 peers with <5 min gap (renewal dropped or routed to a different holder).

## Approach

Call `refresh_peer_interest()` after every successful broadcast send, regardless of delta vs full state. This keeps interest alive as long as broadcasts are being successfully delivered.

This is deliberately separate from `update_peer_summary()` (which sets the cached summary for delta computation). The PR #2763 invariant — that cached summaries are only updated after delta sends to avoid stale summary pollution — is fully preserved.

**Why `refresh_peer_interest` and not `update_peer_summary`:** `refresh_peer_interest` only touches the TTL timestamp. `update_peer_summary` also sets the cached summary, which would break the PR #2763 fix by prematurely caching our summary before the peer has confirmed receipt.

## Testing

- All 1407 unit tests pass (8 pre-existing ignores)
- All integration tests pass
- The existing `test_refresh_prevents_expiration` test in `interest.rs` validates that `refresh_peer_interest` correctly resets TTL and prevents sweep expiration
- Impact will be verified via the `river-funnel.py` telemetry analysis after deployment

### Why didn't CI catch this?

This bug requires multi-peer interaction over time (>5 min) with passive subscribers (read-only users who don't send updates back). Current simulation tests run for shorter durations and most test scenarios involve active peers sending updates in both directions. A comprehensive test would need:
- Multiple peers subscribing to a contract
- One peer sending updates while others only receive
[...]

PR #4462, opened 2026-06-16
test: bound test_interest_renewal so it can't time out on non-convergence

### LINKED ISSUE #3792: fix: test_interest_renewal simulation test is consistently timing out in CI
## Problem

`test_interest_renewal` in `simulation_integration.rs` is consistently timing out (240s per attempt, 2 retries) in CI, blocking the merge queue. It has failed in 3 consecutive CI runs today:

- Run 24101443054 (PR branch): TMT at 240s, both attempts
- Run 24104370490 (merge queue): TMT at 240s, both attempts  
- Run 24105374837 (merge queue): TMT at 23m total Simulation job

The test worked on main recently (last 5 main-branch runs all passed), suggesting either:
1. A timing sensitivity that manifests under CI runner load
2. A resource contention issue with parallel test execution

## Error

```
TRY 1 TMT [ 240.014s] freenet::simulation_integration test_interest_renewal
TRY 2 TMT [ 240.024s] freenet::simulation_integration test_interest_renewal
```

The stderr shows `InvalidStateTransition` errors during the test run, suggesting operations are failing to complete within the simulation timeframe.

## Impact

This is blocking the merge queue for all PRs - currently blocking #3790 which is a crash fix for production nodes.

## Priority

High - blocks all merges.

[AI-assisted - Claude]

sanity:
## Problem

`test_interest_renewal` consistently timed out in CI (240s, both retries), blocking the merge queue (#3792).

It is a `run_direct()` paused-time simulation: 2 gateways + 4 nodes, 8 contracts, 400 iterations × 3s = **1200s of event-phase virtual time**, which the direct runner advances in ~1ms quanta. Wall-clock therefore scales with total virtual time × node count × contract count and is very sensitive to runner CPU.

After the event phase the runner appends an **unconditional convergence-polling tail**: `sleep(15s)` then up to 30 rounds of `sleep(60s)` = **up to 1800s of additional virtual time**, early-exiting only when every contract converges. That tail is *advisory* — it only logs a warning if convergence is never reached; it never fails the simulation. `test_interest_renewal` only inspects broadcast-received events produced **during** the event phase and never asserts convergence, so the tail buys it nothing.

Two factors stacked to produce the timeout:
1. The test ran concurrently with the fdev load sims, starving it of CPU (since fixed by running the nextest sim tests uncontended first, #4301).
2. The `InvalidStateTransition` errors cited in the issue prevented convergence, so the polling tail ran all 30 rounds — adding 1800s on top of the 1200s event phase and blowing the budget.

The earlier "fix" (the #3792 nextest override added in #3790) only bumped the per-test timeout 240s → 300s; it never removed the 1800s landmine, so the issue stayed open. On current `main` the interest-renewal reworks (#3886, #3979, #4133, #4201) let the network converge immediately, so the test passes again — **but the landmine remains**: any future regression that breaks convergence makes this test time out again, even though it doesn't depend on convergence.

## Approach

Make the test's wall-clock independent of whether the network converges — the correct invariant for a test that never asserts convergence — instead of bumping the timeout again.

- Add `SimNetwork::skip_convergence_wait`. When set, the direct runner does only the brief 15s settle and skips the up-to-1800s polling loop. The loop is advisory, so skipping it changes no assertion.
- Thread it from `TestConfig` via a new `no_convergence_wait()` builder that sets `require_convergence = false` (the field already existed; `run_direct` now honors it). Other direct-runner tests keep the full tail unchanged.
- Apply `no_convergence_wait()` to `test_interest_renewal`.
- Keep the 300s nextest override as headroom for the fixed event-phase cost on slow/contended runners, with an accurate comment.

No production code path changes — edits are confined to test-only harness code (`testing_impl.rs`, compiled under `testing`/`simulation_tests`), the integration test, and CI test config. (Hence `test:`, not `fix:`.)

## Testing

Ran `test_interest_renewal` repeatedly under the `ci` nextest profile while other agents loaded the machine; it passed every time (23.7s / 53.2s / 64.7s / 68.3s / 23.2s) — the spread shows the CPU sensitivity that originally caused the timeout. Verified the new path executes (the runner logs `Skipping convergence-polling tail`) and the test still asserts a healthy late/early broadcast ratio (0.87, threshold 0.2), so the #3093 regression guard is intact. `cargo fmt --check` clean.

Closes #3792

[AI-assisted - Claude]

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -9,8 +9,14 @@ fail-fast = false
 # Terminate tests that hang (2 periods of 120s = 4 min max per test).
 slow-timeout = { period = "120s", terminate-after = 2 }
 
-# Interest renewal runs a full simulation with contract operations and needs
-# extra time when competing with parallel fdev tests for CPU (#3792).
+# Interest renewal drives a 6-node / 8-contract simulation for ~1200s of virtual
+# time (paused-time direct runner, advanced in ~1ms quanta), so its wall-clock
+# scales with runner CPU and can be 2-3x slower under load. The original CI
+# timeout (#3792) was a 1800s convergence-polling tail stacked on top of that
+# whenever the network didn't fully converge; that tail is now skipped for this
+# test (it never asserts convergence — see test_interest_renewal's
+# no_convergence_wait()). This generous timeout remains as headroom for the fixed
+# event-phase cost on slow/contended runners.
 [[profile.ci.overrides]]
 filter = "test(test_interest_renewal)"
 slow-timeout = { period = "300s", terminate-after = 2 }
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -9,8 +9,8 @@ fail-fast = false
 # Terminate tests that hang (2 periods of 120s = 4 min max per test).
 slow-timeout = { period = "120s", terminate-after = 2 }
 
-# Interest renewal drives a 6-node / 8-contract simulation for ~1200s of virtual
-# time (paused-time direct runner, advanced in ~1ms quanta), so its wall-clock
+# Interest renewal drives a 2-gateway / 4-node / 8-contract simulation for ~1200s
+# of virtual time (paused-time direct runner, advanced in ~1ms quanta), so its wall-clock
 # scales with runner CPU and can be 2-3x slower under load. The original CI
 # timeout (#3792) was a 1800s convergence-polling tail stacked on top of that
 # whenever the network didn't fully converge; that tail is now skipped for this
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -947,6 +947,20 @@ pub struct SimNetwork {
     /// When true, use `MockWasmRuntime` (production `ContractExecutor` code path)
     /// instead of `MockRuntime` (simplified hash-based merge).
     pub use_mock_wasm: bool,
+    /// When true, the direct runner skips the long post-event convergence-polling
+    /// loop and does only a brief fixed propagation settle instead.
+    ///
+    /// The direct runner normally polls for full state convergence for up to
+    /// ~1800s of virtual time after the event phase (30 rounds × 60s, early-exit
+    /// on convergence). That tail is advisory — the runner only logs a warning if
+    /// convergence is never reached, it does not fail the simulation. For tests
+    /// that analyze events produced *during* the event phase and never assert
+    /// convergence (e.g. `test_interest_renewal`), running the full polling tail
+    /// when the network happens not to converge adds up to 1800s of virtual time
+    /// for no benefit, blowing the wall-clock budget and making the test time out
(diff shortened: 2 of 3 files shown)


sanity:
## Multi-model review (FULL tier: interest-renewal / subscribe surface)

Reviewed the checked-out code, not just the diff.

### External model — Codex (`codex review --base origin/main`)
> "The change cleanly scopes the shortened direct-runner settle behavior to tests that opt out of convergence assertions, and the updated interest-renewal test matches that behavior. I did not identify any actionable regressions in the diff."

No findings. (Codex ran without crates.io access, so it was static-only; compensated for the lost build/runtime signal with a dedicated adversarial Claude lens below.)

### Adversarial Claude lens (skeptical bug-hunt — the specific failure mode for this change)

Focused on the question that matters most: *does skipping the convergence tail weaken the test's ability to catch #3093?* Verdict: **no — it sharpens it.** Key findings:

- **Does not mask #3093.** `UpdateEvent::BroadcastReceived` events are driven only by user UPDATE events, which are emitted only during the 1200s event phase (= 1x INTEREST_TTL, 2.5x the 480s lease). No new updates are generated during the tail. So the log-index midpoint of broadcast events maps to ~600s virtual, and the late half (~600s-1200s) still spans the 960s lease boundary and approaches the 1200s TTL boundary — exactly where a #3093 regression thins out broadcasts. The tail's only possible late-half contribution was interest-sync *repair* broadcasts, which are independent of the broadcast-TTL-refresh path under test and could actually have *masked* a regression. Removing the tail removes that confound. Assertion logic (early/late split, `total>0`, `late>0`, `ratio>0.2`) is byte-identical to main.
- **`require_convergence` not accidentally overloaded.** The only other `require_convergence: false` is in `test_subscription_renewal_at_scale`, which uses the Turmoil `run()` path (no convergence tail at all) — `skip_convergence_wait` is never set there. All four config builders default `require_convergence: true`. Only `test_interest_renewal` reaches the skip path.
- **Other `run_simulation_direct` callers unaffected.** fdev `single_process` and every other direct-runner test construct via `SimNetwork::new` (default `skip_convergence_wait = false`) — they retain the full tail.
- **Wiring correct.** `run_direct()` sets `sim.skip_convergence_wait = !require_convergence` before calling the runner; the runner reads it before the loop; the 15s settle is shared by both paths.
- **NITs (fixed in 89ef7f5c):** two comments referenced a non-existent `require_convergence(false)` call (the real opt-out is `no_convergence_wait()`); nextest comment said "6-node" (= total peers) → clarified to "2-gateway / 4-node".

### Verification
Ran `test_interest_renewal` 5x under the `ci` nextest profile on a contended machine: passed every time (23.7s / 53.2s / 64.7s / 68.3s / 23.2s — the spread is the CPU sensitivity that originally caused the timeout). Confirmed the new path executes (runner logs `Skipping convergence-polling tail`) and the #3093 guard is intact (late/early ratio 0.82-0.87, threshold 0.2).

All findings addressed. No blocking issues.

[AI-assisted - Claude]

Issue #4222, opened 2026-05-24
fix: GETs on subscribed contracts often never visit a subscriber (63% of failing GETs)


sanity:
## Problem

Telemetry from the last 24h shows that when a GET fails on a contract that *does* have active subscribers, **63 % of the time the GET never visits any of the subscriber peers during routing**. The data is on the network but greedy routing terminates before reaching it.

| Failure pattern on subscribed contracts | Count | % |
|---|---:|---:|
| GET visited a subscriber, still got NF | 63 | 37 % |
| GET never visited any subscriber | **109** | **63 %** |
| (No hops recorded) | 0 | 0 % |

Total: 172 NF events across 38 contracts with median 31 subscribers (max 173) — these are not low-replication contracts.

Hop counts on these failing GETs:
- Missed-subscriber failures: mean **3.9 hops**, max 11
- Visited-subscriber failures: mean **5.4 hops**, max 11

The htl=10 ceiling is rarely hit, so the operation is terminating before exhausting hops — it's running out of routing options or accepting NF prematurely.

37 of the 38 affected contracts ALSO had successful GETs in the same window, so this is intermittent, not \"contract has vanished from the network.\"

### 1. The 5-candidate window in the isotonic router

`select_closest_peers` in `crates/core/src/router.rs` truncates the candidate set to `consider_n_closest_peers` (default **5**, `router.rs:355`) before the isotonic estimator ranks them:

```rust
peer_distances.sort_by_key(|&(_, distance)| distance);
peer_distances.truncate(self.consider_n_closest_peers);
```

If none of the geometrically-5-closest peers happen to be subscribers — entirely possible when subscribers are uniformly distributed across the keyspace (which our telemetry shows they are, roughly flat across [0,1] for the top contracts) — the router has no path to discover them through that hop. The isotonic prediction operates on a pre-filtered set, so its \"learning\" can't recover from a candidate-set miss.

Worth checking: does increasing `consider_n_closest_peers`, or biasing candidate selection by historical subscription / hosting evidence, reduce the missed-subscriber rate without ballooning routing cost?

### 2. Peer health vetting not applied to GETs

#3570 already noted: `k_closest_potentially_hosting` (the GET peer-selection feeder) does not use `connect_excluded_peers` or any peer-health signal. If a forwarding peer is slow or partially failed, GETs through it may abandon before reaching deeper subscribers.

### 3. Forwarding termination

[...]

PR #4463, opened 2026-06-16
test(router): add routing-convergence tests for widened candidate window (#4230)

### LINKED ISSUE #4230: test: add routing-convergence test for predictor at widened candidate window (follow-up to #4222)
## Background

PR #4222 widened the isotonic router's candidate window from 5 to 25 (matching `Ring::DEFAULT_MIN_CONNECTIONS`) to fix the production failure mode where 63% of failing GETs on subscribed contracts never visited any subscriber.

The structural fix is verified by `router::tests::select_closest_peers_includes_subscribers_4222`, which pins the SHIPPED window value and asserts subscriber-coverage > 85% via hypergeometric statistics. That test proves the window is *wide enough to surface subscribers*.

## What this issue tracks

What is **NOT** verified is the converse: that the wider window does not degrade routing quality on *non-subscriber* GETs by letting the isotonic predictor pick a distant peer that happens to have slightly better per-peer EWMA history than the closest peer.

The predictor's scoring formula (`router.rs::predict_routing_outcome`) has no direct distance term — distance enters only via the global isotonic regression (`router/isotonic_estimator.rs`) and as a feature passed to the renegade predictor. Once a peer has ≥10 events (`ADJUSTMENT_PRIOR_SIZE`), an EWMA adjustment near +1.0 (consistently failing) can dominate any distance gap across the 25-peer window. Equally, a distant peer with consistently good history will be preferred over a close peer with bad history.

This is by design — but at the historical window=5, the truncation was self-limiting. At window=25, the predictor has 5× more candidates whose per-peer history can win against geographic locality. Whether this preserves small-world routing convergence in steady state is an empirical question that the regression test in #4222 does not answer.

## What a fix looks like

A deterministic test (under `crate::config::GlobalRng::seed_guard`) that:

1. Trains a `Router` with a synthetic history that exhibits a distance→failure correlation (consistent with what the production isotonic regression learns).
2. Plants two candidate peers: one geometrically close to the target with one bad event, one distant from the target with consistently good history.
3. Calls `select_k_best_peers(..., k=1)` and asserts the close peer still wins (or that the distance gap required for the distant peer to win is bounded).

Repeat across a parameter sweep of (window size, history length, EWMA-adjustment magnitude) to establish the threshold at which distance-based routing actually degrades. If window=25 is safely inside the threshold, file the test as the regression guard. If window=25 is uncomfortably close to the threshold, this issue surfaces that the fix may need a different shape (e.g., a soft distance bias in the predictor, not just truncation).

Possibly also worth: a simulation_integration test asserting that hop-counts on successful GETs don't grow significantly after the window widening (small-world preservation at the network level, not just the per-hop level).

## Why this is a follow-up, not a #4222 blocker

The production failure mode #4222 targets (63% subscriber-miss rate) is structural and well-understood. The fix demonstrably addresses it (hypergeometric prediction: P(no subscriber in 25-window) ≈ 0.07 vs ≈ 0.58 for window=5). The deferred routing-quality risk is theoretical — there is no production telemetry showing routing convergence has degraded — and addressing it requires non-trivial test scaffolding (Router state setup + predictor seeding) that doesn't exist today.

Filing as a tracking issue per the project's "finish-the-fix" stance.

## Related

- #4222 — primary fix (window widening + transient filter)
- #4229 — separate follow-up for `MAX_RELAY_RETRIES = 1` (relay single-shot forwarding)
- #4225 — pre-existing opentelemetry build break (unrelated, found during #4222 work)

[AI-assisted - Claude]

sanity:
## Problem

#4222 widened the isotonic router's candidate window from 5 to 25 (matching `Ring::DEFAULT_MIN_CONNECTIONS`) to surface subscribers on more hops. `select_closest_peers_includes_subscribers_4222` proves the window is wide enough to *surface* subscribers, but the converse was never verified: does the wider window *degrade* routing quality for ordinary (non-subscriber) GETs by letting the isotonic predictor pick a distant peer whose per-peer EWMA history beats a close peer's?

At window=25 there are 5x more candidates whose per-peer history can compete with geographic locality, and `predict_routing_outcome` has no direct distance term (distance enters only via the global isotonic regression and as a renegade-predictor feature). Whether small-world routing convergence survives the widening was an open empirical question (issue #4230).

## Solution

The candidate window controls only which peers are *visible* to the predictor, not the pairwise ranking of two fixed peers — so a hand-planted "close vs distant duel" is window-independent and answers the wrong question. The meaningful test is aggregate: over many random targets against a router trained on a realistic distance->failure gradient (closer peers succeed, farther peers fail — the shape the production isotonic regression learns, plus per-peer noise so EWMA adjustments populate), measure the chosen next hop's **rank-percentile** = the fraction of the pool strictly closer to the target than the chosen hop. `0.0` = closest peer (perfect greedy routing); `0.5` = median (no progress / non-convergence).

Rank-percentile is used instead of a distance ratio because the ratio's denominator (closest-peer distance) goes near-zero when a peer sits almost on the target, making a mean-ratio metric explode on denominator outliers rather than measure routing quality. Rank-percentile is bounded in `[0, 1]` and directly captures small-world progress.

Two deterministic tests (`GlobalRng::seed_guard`):

- **`routing_convergence_preserved_at_window_25_4230`** — at the shipped window of 25 the chosen hop stays deep in the closest tail (mean rank-percentile < 0.15, p90 < 0.25). Also pins `DEFAULT_CONSIDER_N_CLOSEST_PEERS == 25`.
- **`routing_convergence_window_sweep_does_not_degrade_4230`** — sweeps windows 5/10/25/50/100 and asserts the chosen hop stays in the closest tail at every window, and that widening 5 -> 25 keeps routing convergent (the load-bearing window-variable assertion).

Thresholds are evidence-based, not guessed: an 8-seed robustness scan put window=25 mean rank-percentile in `0.0075..0.097` and p90 in `0.025..0.175`, comfortably above the observed worst case yet far below the `0.5` no-convergence value, so a real degradation trips CI while seed jitter does not.

## Finding

**The #4222 widening is safe.** Widening the window does raise the chosen hop's rank-percentile modestly (more near-equidistant candidates become visible, so per-peer history breaks more ties), but it plateaus well inside the convergent regime and never drifts toward the median. No soft distance bias in the predictor is needed; these tests pin that property as a regression guard before any future window-size tuning.

## Testing

Test-only change — no production code is touched. New tests pass deterministically (verified across re-runs and an 8-seed scan); the full `router::` module (86 tests) is green; `cargo fmt` clean; the added test code is clippy-clean.

Closes #4230

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/router.rs
+++ b/crates/core/src/router.rs
@@ -1300,29 +1300,46 @@ mod tests {
     }
 
     /// Train a `Router` on a synthetic history whose global shape matches what
-    /// the production isotonic regression learns: a distance->failure gradient
-    /// (closer peers succeed more, farther peers fail more) plus per-peer noise
-    /// so that per-peer EWMA adjustments are populated across the whole pool —
-    /// the regime in which a distant-but-historically-good peer could in
-    /// principle out-score a close-but-unlucky one.
+    /// the production isotonic regression learns — a distance->failure gradient
+    /// (closer peers succeed more, farther peers fail more) — but with each peer
+    /// *also* carrying an individual reliability bias that is INDEPENDENT of its
+    /// distance. The per-peer bias is what populates mature, divergent per-peer
+    /// EWMA adjustments: a far peer can be individually reliable (adjustment well
+    /// below the global curve) and a near peer individually unreliable. This is
+    /// exactly the regime issue #4230 worries about — a distant-but-good peer
+    /// competing against a close-but-unlucky one — so the convergence metric is
+    /// stressed, not just rubber-stamped against a clean monotone gradient.
     ///
-    /// `pool` is the set of peers events are drawn from; each of `rounds`
-    /// rounds touches every peer once at a fresh random target, so every peer
-    /// accrues >= ADJUSTMENT_PRIOR_SIZE events. Returns the event history (the
+    /// To exercise MATURE per-peer history (`>= ADJUSTMENT_PRIOR_SIZE = 10`
+    /// retained events per peer) within `IsotonicEstimator`'s 500-point rolling
+    /// window, keep `pool.len() * rounds <= MAX_REGRESSION_POINTS (500)` and
+    /// `rounds >= ~12`; every round touches every peer once, so each peer ends
+    /// with `rounds` retained observations. Returns the event history (the
     /// caller builds a `Router` per candidate window from it, since `Router` is
     /// not `Clone`). Deterministic under the caller's seed guard.
     fn gradient_history(pool: &[PeerKeyLocation], rounds: usize) -> Vec<RouteEvent> {
+        // Per-peer reliability bias in [-0.35, +0.35], independent of distance.
+        // Positive = the peer fails MORE than its distance predicts (bad
+        // individual record); negative = fails LESS (good individual record).
+        // Drawn once per peer so the bias is stable across rounds and the EWMA
+        // converges to a mature, peer-specific adjustment.
+        let bias: Vec<f64> = pool
+            .iter()
+            .map(|_| GlobalRng::random_range(-0.35..0.35))
+            .collect();
+
         let mut events: Vec<RouteEvent> = Vec::new();
         for _ in 0..rounds {
             // Fresh random target each round so the gradient is learned across
             // the whole ring, not just one contract location.
             let contract = Location::random();
-            for p in pool {
+            for (i, p) in pool.iter().enumerate() {
                 let d = contract.distance(p.location().unwrap()).as_f64(); // [0, 0.5]
-                // Failure probability rises with distance: ~0 when adjacent,
-                // ~1 at the antipode. This is the correlation the issue says
-                // the production regression learns.
-                let fail = GlobalRng::random_range(0.0..1.0) < (d * 2.0);
+                // Failure probability rises with distance (the global curve) and
+                // is shifted by the peer's individual bias, then clamped to a
+                // valid probability. d*2 maps [0,0.5] -> [0,1].
+                let p_fail = (d * 2.0 + bias[i]).clamp(0.0, 1.0);
+                let fail = GlobalRng::random_range(0.0..1.0) < p_fail;
                 events.push(RouteEvent {
                     peer: p.clone(),
                     contract_location: contract,
@@ -1401,10 +1418,11 @@ mod tests {
     /// locality (the #4230 worry), the chosen hop's rank-percentile would drift
     /// toward 0.5; instead it stays near zero.
     ///
-    /// Empirical basis for the thresholds: across an 8-seed robustness scan the
-    /// window=25 mean rank-percentile ranged 0.0075..0.097 and p90 ranged
-    /// 0.025..0.175 (120-peer pool). The bounds below (mean < 0.15, p90 < 0.25)
-    /// sit comfortably above the observed worst case while remaining far below
+    /// Empirical basis for the thresholds: across a 12-seed robustness scan
+    /// (40-peer pool, MATURE per-peer history with per-peer reliability biases
+    /// independent of distance), the window=25 mean rank-percentile ranged
+    /// 0.042..0.120 and p90 ranged 0.10..0.30. The bounds below (mean < 0.18,
+    /// p90 < 0.40) sit above the observed worst case while remaining far below
     /// the no-convergence value of 0.5, so a real degradation (history
     /// overriding distance) trips CI but seed jitter does not. Deterministic.
     #[test]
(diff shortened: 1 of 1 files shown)


sanity:
## Review: Codex (external, non-Claude)

Ran `codex review --base origin/main`. One finding:

- **[P2] Exercise mature per-peer history** (`router.rs:1431`): with a 120-peer pool and `rounds = 4`, each peer contributed only 4 observations, so per-peer EWMA adjustments stayed shrunk toward the global curve — the tests never reached the mature-history regime (`>= ADJUSTMENT_PRIOR_SIZE = 10` events/peer) that #4230 is actually about. `IsotonicEstimator` also retains only the most recent 500 points, so a 120-peer pool cannot keep 10 events per peer.

**Resolution (fixed in f3ff4b9c):**
- Pool shrunk to 40 peers x 12 rounds = 480 events (≤ the 500-point window), so every peer ends with 12 retained observations — mature EWMA.
- Each peer now carries a reliability bias **independent of distance**, so per-peer adjustments genuinely diverge from the distance curve, creating the distant-but-good vs close-but-unlucky tension #4230 worries about.
- Added an explicit assertion that the regime is exercised: every peer must carry a per-peer failure adjustment with effective event count ≥ 10.
- Recalibrated thresholds against a 12-seed scan in the mature regime (window=25 mean rank-percentile 0.042..0.120, p90 0.10..0.30): `mean < 0.18`, `p90 < 0.40`.

**Finding unchanged:** the #4222 widening is safe. Widening the candidate window raises the chosen-hop rank-percentile modestly (more near-equidistant candidates → per-peer history breaks more ties) but it plateaus deep in the closest tail (mean ≈ 0.08-0.12, well below the 0.5 no-convergence value) and never drifts toward the median. No soft distance bias in the predictor is needed.

Re-running the external review on the updated diff (code changed since the Codex pass).

[AI-assisted - Claude]

sanity:
## Review: Claude multi-perspective (testing + big-picture lenses)

FULL-tier surface (routing/predictor), test-only. Two Claude reviewers ran read-only adversarial passes in addition to the external Codex pass.

### Findings + resolutions (fixed in 57766b53)

**1. [Substantive — both reviewers] The maturity assertion was vacuous.** The original `event_count() >= 10` check could never witness mature history: `Adjustment.effective_count` is seeded at `ADJUSTMENT_PRIOR_SIZE = 10`, which is the exact EWMA fixed point (`1 + 0.9·10 = 10`), so it reports 10 whether a peer has 1 real event or 12. **Fixed:** now assert on the *spread of per-peer adjustment `value()`s* (> 0.2), which actually proves individual records diverged from the global distance curve — the regime #4230 is about. (The regime was in fact exercised before; the assertion just proved the wrong invariant.)

**2. [Coverage — testing] window=5 sub-assertions are structurally near-unfailable.** `select_closest_peers` truncates to the window *before* the predictor reranks, so at window=5 the chosen-hop rank-percentile is mechanically capped near `5/40`. **Fixed:** documented this in the test; clarified that the load-bearing checks are windows 25/40 + the bounded 5→25 increase.

**3. [Scope — big-picture] The issue asked for a multi-dimensional sweep (window, history length, EWMA magnitude); only window was swept.** **Fixed:** added the EWMA-magnitude dimension (bias_mag 0.35 and 0.5) to the sweep test. (History-length is fixed at 12 rounds, bounded by the estimator's 500-point window for a mature-history pool; the simulation-level hop-count test remains the issue's explicit stretch goal and is consciously descoped — noted below.)

**4. [Doc — big-picture] Evidence-number discrepancy** between PR text (8-seed) and code docstring (12-seed). **Fixed:** reconciled to the actual 12-seed × 2-bias scan numbers (window=25 mean 0.037..0.159, p90 0.10..0.35); thresholds (mean < 0.20, p90 < 0.42) bracket these with headroom and stay far below the 0.5 no-convergence value.

### On Closes vs Refs

Keeping `Closes #4230` (a `T-tracking` issue owned by the same maintainer): the test discharges the issue's core conditional — "if window=25 is uncomfortably close to the degradation threshold, the fix needs a soft distance bias" — by showing it is comfortably far, across both the window and EWMA-magnitude dimensions and at maximal per-peer history divergence. The simulation_integration hop-count test is the issue's explicit stretch goal and is left for a future follow-up if network-level telemetry ever motivates it.

### External pass

Codex (`codex review --base origin/main`) re-run after each code change; the content has changed since the last clean pass, so a final external re-run is in progress — will record its outcome.

No blocking findings remain. All review feedback is fixed.

[AI-assisted - Claude]

sanity:
## External re-review: Codex (clean)

Re-ran `codex review --base origin/main` on the final content (57766b53), since the diff changed after the prior pass:

> The patch only adds deterministic regression tests and helper functions under the existing router test module. I did not identify any introduced correctness, compile-time, or CI-blocking issues in the changed code.

Review gate complete: external Codex pass clean on the current HEAD; both Claude adversarial lenses' findings fixed. Will enable auto-merge once CI is green on this SHA.

[AI-assisted - Claude]

PR #4464, opened 2026-06-16
feat(dashboard): add contract ban-list panel to local peer dashboard

### LINKED ISSUE #4302: Phase 7 follow-up: dashboard ContractBanList panel
## Context

Spawned from PR #4299 multi-model review.

PR #4299 lands the \`ContractBanList\` data structure with public \`len()\` accessor and a \`BanReason\` enum (\`AutoMad\`, \`Operator\`). It does NOT yet expose this on the local dashboard.

## Proposed

Add a panel/card to the contract-hardening dashboard showing:

- **Count** of currently-banned contracts (drives a "N contracts on ban list" tile in the status strip).
- **List** of banned contracts with: contract key, reason (AutoMad vs Operator), expiry time remaining.
- **Drill-down**: for each banned contract, show the governance state machine history that led to BanTriggered.

The dashboard is the operator's window into how Phase 7 is performing. Without this panel, operators can't tell whether the mechanism is actually catching abusers or sitting idle.

Tie into the existing dashboard provider closure pattern (see \`server::home_page\`) so the state is read from the canonical \`Ring::contract_ban_list\` and there's no manually-mirrored counter to rot.

## Spawned from

- Code-first review (#5 — No telemetry / dashboard counter for ban drops)
- Big-picture review (dashboard wiring is a deferred follow-up)
- Memory entry on \"Manually-mirrored telemetry counters / dashboard providers\" — must use the provider closure pattern, not re-mirror.

[AI-assisted - Claude]

sanity:
## Problem

PR #4299 landed the `ContractBanList` (Phase 7 contract hardening) but never surfaced it on the local peer dashboard. Operators had no way to tell whether the ban mechanism is catching abusers or sitting idle — the per-contract governance table only shows a "Banned" badge, not the ban list itself, its size, or per-entry reason/expiry.

## Solution

Follow the established provider-closure pattern (same shape as the governance / subscription / ring-stats providers) so the panel reads from the canonical `Ring::contract_ban_list` on every dashboard request — no manually-mirrored counter to rot (the failure mode called out in `AGENTS.md` bug-prevention-patterns).

- **`ContractBanList::snapshot()`** — read-only view of each live entry (contract id, reason, time remaining), using the same `now < expires_at` predicate as `is_banned` so expired-but-unswept entries never appear. Exposes only data the list already stores; no new persisted structure.
- **`network_status::BanListSnapshot` / `BanListEntry` / `BanReasonSnapshot`** + `BanListProvider` static, wired into `get_snapshot()`.
- **`Ring::dashboard_ban_list_snapshot()`** maps the ring-internal types to the public snapshot and sorts entries (soonest-to-lift, then id) for stable display across refreshes (DashMap iteration order is unspecified).
- Provider registered in `p2p_impl.rs` capturing `Arc<Ring>`.
- **`build_ban_list_card()`** in `home_page.rs`: a count tile + per-entry list (key, reason, expiry remaining), plus a capacity-rejection note when non-zero. Rendered even when empty so an idle-but-active mechanism is distinguishable from an unwired one (the issue's core motivation).

Scope follows the issue's two concrete asks (count tile + entry list). The governance-state-machine **drill-down is deferred**: the ban list stores only the current entry, not the transition history that led to `BanTriggered`. Per the issue ("Don't add new data structures the ban list doesn't already expose"), no history threading was added.

## Testing

- `contract_ban_list.rs`: 4 `snapshot()` unit tests — empty list, key/reason/remaining reported, remaining shrinks as virtual time advances (via `SharedMockTimeSource`), and expired-but-unswept entries excluded (same predicate as `is_banned`).
- `home_page.rs`: 6 `build_ban_list_card()` HTML-assertion tests — empty state (count-0 tile + idle message, no table), entry list with both `AutoMad`/`Operator` reasons + expiry rendering, singular pluralization at count==1, capacity note shown/hidden, omit-when-None.

All new tests reference symbols that did not exist before this change, so they fail to compile without the fix and pass with it (the fails-before/passes-after requirement). `cargo fmt` and CI-equivalent `cargo clippy --locked -- -D warnings` are clean.

## Note for reviewers

`cargo clippy --all-targets` surfaces one **pre-existing** `assertions_on_constants` warning in `crates/core/src/tracing/telemetry.rs` (test `test_shadow_subbudget_caps_shadow_events`, from #4456). It is unrelated to this PR and does not break CI (CI runs clippy without `--all-targets`), so it was left out to keep this PR to one logical change.

Closes #4302

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -39,6 +39,14 @@ pub type GovernanceProvider = Arc<dyn Fn() -> GovernanceSnapshot + Send + Sync +
 /// Same provider pattern as `SubscriptionProvider`/`GovernanceProvider`.
 pub type RingStatsProvider = Arc<dyn Fn() -> RingStatsSnapshot + Send + Sync + 'static>;
 
+/// Provider for the contract ban-list snapshot (#4302). Same pattern as
+/// `GovernanceProvider`: registered at node startup, replaceable for
+/// multi-node test harnesses, read by `get_snapshot` on every dashboard
+/// request. In production the closure captures `Arc<Ring>` and calls
+/// `Ring::dashboard_ban_list_snapshot()`, which reads the canonical
+/// `Ring::contract_ban_list` directly — no mirrored counter to rot.
+pub type BanListProvider = Arc<dyn Fn() -> BanListSnapshot + Send + Sync + 'static>;
+
 /// Snapshot of ring-level statistics exposed to the dashboard.
 #[derive(Debug, Clone, Default)]
 pub struct RingStatsSnapshot {
@@ -91,6 +99,16 @@ pub(crate) fn clear_ring_stats_provider() {
 #[allow(dead_code)] // consumed by Phase 4.5 dashboard tests
 pub(crate) fn clear_governance_provider() {
     *GOVERNANCE_PROVIDER.write() = None;
+}
+
+static BAN_LIST_PROVIDER: parking_lot::RwLock<Option<BanListProvider>> =
+    parking_lot::RwLock::new(None);
+
+/// Register the dashboard's contract-ban-list data source (#4302).
+/// Replaces any previously-registered provider so multi-node in-process
+/// harnesses can re-wire to the current node.
+pub fn set_ban_list_provider(provider: BanListProvider) {
+    *BAN_LIST_PROVIDER.write() = Some(provider);
 }
 
 /// Replaceable storage for the subscription provider. Wrapped in a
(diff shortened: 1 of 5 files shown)


sanity:
## Review (Light tier — read-only dashboard)

Ran an external **Codex** pass (`codex review --base origin/main`) — the highest-value independent, non-Claude reviewer.

**Round 1** found one P2 correctness issue:
- The ban-list count tile read `ContractBanList::len()`, which includes expired-but-unswept entries, while the entry list (and the `is_banned` wire boundary) use the `now < expires_at` predicate via `snapshot()`. During the expiry-before-sweep window the tile could show "1 banned" with zero rows.

**Fix (commit `bc6a9a9c`):** derive `count` from the filtered `entries.len()` so the tile, the entry list, and enforcement all agree. Added a source-scrape pin test (`dashboard_count_derives_from_live_entries_not_len`) so a regression to `len()` fails CI. Updated the now-stale `len()` / `capacity_rejected_total` docs.

**Round 2** (re-run on the updated HEAD, since the code changed): *"No actionable correctness issues were found in the diff. The added ban-list snapshot plumbing, dashboard rendering, and tests appear consistent with the existing provider patterns."*

Also applied a code-first / skeptical self-pass: snapshot uses the same expiry predicate as `is_banned`; entries are sorted for stable display (DashMap order is unspecified); the panel reads canonical `Ring::contract_ban_list` via the provider closure (no mirrored counter to rot); `html_escape` is applied defensively to rendered ids.

Pre-existing note (not in this PR): `cargo clippy --all-targets` surfaces an `assertions_on_constants` warning in `crates/core/src/tracing/telemetry.rs` (test from #4456). It does not break CI (CI runs clippy without `--all-targets`) and is unrelated, so it was left out to keep this PR to one logical change.

[AI-assisted - Claude]

PR #4467, opened 2026-06-16
fix(tracing): make event-log registration non-blocking to prevent node deadlock

### LINKED ISSUE #4466: Silent node deadlock: untimed event-log channel send blocks the event loop under telemetry-sink stall
## Summary

A Freenet node can **silently deadlock**: the process stays alive with its API port open, but every runtime thread is parked on a futex at 0% CPU, all logging stops mid-activity, and the node serves nothing until manually restarted. Root cause is an untimed `.send().await` on the bounded event-log channel, awaited from the network event loop's hot path — the exact pattern `.claude/rules/channel-safety.md` bans.

## Fingerprint

- Process alive, UDP/WS ports still bound (TCP accepts), but the dashboard/`freenet service report` query times out or is refused.
- `/proc/<pid>/task/*/wchan` = `futex_wait_queue` for **every** thread; `top -H` shows 0% CPU; no `ep_poll` thread.
- Node log goes silent **mid-activity** (no shutdown line), typically after a burst of GET-relay streaming (`Stream assembly timed out`, `Outbound stream send failed`, `pipe_stream stalled`).

## Root cause

`EventRegister::register_events` (and `notify_of_time_out`) in `crates/core/src/tracing.rs` did:

```rust
self.log_sender.send(EventLogCommand::Log(log_msg)).await   // bounded mpsc::channel(1000)
```

This future is awaited from the event loop's hot outbound path — `crates/core/src/node/network_bridge/p2p_protoc.rs` `OutboundMessageWithTarget` handler (every outbound message) and the disconnect handler. The sole consumer, `record_logs`, blocks on:
- `send_to_metrics_server(...).await` → `ws_stream.send(...).await` — untimed WebSocket send to the local metrics server (`ws://127.0.0.1:55010`), present on peers running the fdev network-viz / metrics server, **or**
- `event_log.persist_log(...).await` — AOF write (`spawn_blocking` + `FILE_LOCK` + `fsync`).

When the consumer stalls, the `channel(1000)` fills; the event loop then blocks forever on `register_events().await`; that stops draining `conn_event_rx`, per-connection listeners stop processing ACKs, and the whole runtime cascades to all-threads-parked. Even the tracing-appender threads can no longer be scheduled, which is why logging freezes instantly.

This is a sibling of #4145/#4231: that fix made the `event_loop_notification` channel non-blocking and wrapped the per-peer send 3 lines below in a 500ms timeout — but the event-log channel send directly above it was missed. The defect predates the recent stream work (the call site dates to ~#3347); the June streaming restoration (#4308/#4345/#4367) put large relay streams back on this path and made the consumer-side stall (and thus the deadlock) far more likely.

## Evidence

- **0.2.75 peer (technic):** deadlocked 2026-06-16T21:12:49Z after ~1h uptime; all 6 threads `futex_wait_queue`, 0% CPU, port open, logging frozen mid-activity. Telemetry confirms its peer id went silent 21:12:41→21:22:42Z (freeze → manual restart). Core dump preserved (no DWARF, so no userspace unwind).
- **Real-user diagnostic report `YWA9E6` (v0.2.74):** user message "Node stalling or restarting"; `network_status` null; main log ends mid-activity at 13:07:59 with no shutdown line, ~7 min before the report; 47 `Stream assembly timed out` + 21 `Outbound stream send failed` preceding.
- **Gateway vega:** hit the related `event loop may be stuck` (#4145) hard-stall for ~7h on 0.2.72 before an auto-update force-restart.

Network-wide telemetry shows this is **not** a mass outbreak — most stall pressure was 0.2.73 (mitigated by 0.2.74/0.2.75) — but the event-log deadlock vector is long-standing and reaches real users, so it should be closed.

## Fix

PR makes `register_events` / `notify_of_time_out` use `try_send` with drop-on-full (best-effort telemetry per channel-safety.md) + a rate-limited dropped-event counter. One impl change covers all five event-loop call sites. Two deterministic regression tests reproduce the hang (they block, caught by a timeout, without the fix). A full audit confirmed no other unprotected event-loop-reachable bounded `.send().await` remains.

## Follow-ups (defense in depth, separate issue)

- Out-of-band event-loop **liveness watchdog** (dedicated OS thread → detect a full freeze, log + abort → systemd restart). Cause-agnostic; relates to #4233.
- Consumer-side timeouts on `record_logs` (WS send + AOF write) so a stalled sink can't keep telemetry dead until restart.
- Low-rate pull-based `event_log_dropped_total` gauge on `NodeDiagnostics` (read on dashboard/report; no per-event telemetry-server load).

[AI-assisted - Claude]

sanity:
## Problem

A Freenet node can **silently deadlock**: the process stays alive with its API port open, but every runtime thread parks on a futex at 0% CPU, all logging stops mid-activity, and the node serves nothing until it is manually restarted (no crash, so systemd never restarts it).

The production `EventRegister::register_events` (and `notify_of_time_out`) did an untimed `.send().await` on the bounded event-log `mpsc::channel(1000)`. That future is awaited from the network event loop's hot outbound path (`p2p_protoc.rs` `OutboundMessageWithTarget` and disconnect handlers). The sole consumer, `record_logs`, blocks on its metrics-server WebSocket send (`ws://127.0.0.1:55010`, present on peers running the fdev metrics server) **or** its AOF write. When the consumer stalls, the channel fills, the event loop blocks here forever, and the whole runtime cascades to all-threads-parked — even the tracing-appender, which is why logging freezes instantly.

This is the exact pattern `.claude/rules/channel-safety.md` bans, and a sibling of #4145/#4231: that fix made the `event_loop_notification` channel non-blocking and wrapped the per-peer send three lines below in a 500ms timeout, but missed this event-log channel send directly above it.

Observed on a 0.2.75 peer (silent freeze, all threads `futex_wait_queue`, 0% CPU, port open) and corroborated by a real-user diagnostic report on 0.2.74 ("Node stalling or restarting"). Full evidence in #4466.

## Solution

Make `register_events` and `notify_of_time_out` use `try_send` with **drop-on-full** — best-effort telemetry per channel-safety.md ("drop messages when full rather than blocking"). The single impl change covers all five event-loop call sites (they all route through `EventRegister::register_events`). The `trace-ot` `OTEventRegister` siblings (same hot path via `DynamicRegister`) are converted too, so the class is fully closed for every build.

Observability: a `DROPPED_EVENT_LOGS` counter emits a `tracing::warn!` at power-of-two milestones, so a persistent stall surfaces in the node's own logs — and therefore in a captured `freenet service report` — without log spam. (A structured `event_log_dropped_total` gauge on `NodeDiagnostics` is deliberately left as a follow-up; see below.)

An audit of every event-loop-reachable bounded `.send().await` confirmed this was the **last** unprotected instance of the class; all others are timeout-wrapped, spawned off the critical path, unbounded, or a documented same-runtime exception. The remaining `EventFlushHandle::flush` send is intentionally left blocking (a shutdown/test sync barrier, timeout-bounded, never on the event loop) and now carries an inline exception comment. `.claude/rules/channel-safety.md` is updated with this incident per repo convention.

## Testing

Four deterministic unit tests (`eventlog_backpressure_tests`, `current_thread` + `start_paused`):
- `register_events` / `notify_of_time_out` complete under a timeout when the log channel is saturated. **Verified these FAIL without the fix** — the old blocking `.send().await` parks the task; under paused virtual time the runtime auto-advances to the 5s timeout, so the test fails deterministically (`Elapsed`) in ~0ms rather than hanging in wall-clock.
- The `Full` arm increments `DROPPED_EVENT_LOGS` (lower-bound assertion, order-independent under parallel test execution).
- A closed channel returns promptly without panicking (exercises the `Closed` arm over a batch).

`cargo fmt` clean; `cargo clippy --locked -- -D warnings` clean; `--features trace-ot` compiles.

## Defense in depth / follow-ups (separate issue, per one-change-per-PR)

- Out-of-band event-loop **liveness watchdog** (dedicated OS thread; detect any full freeze → log + abort → systemd restart). Cause-agnostic backstop; relates to #4233.
- Consumer-side timeouts on `record_logs` (WS + AOF) so a stalled sink can't keep telemetry dead until restart.
- Low-rate pull-based `event_log_dropped_total` gauge on `NodeDiagnostics`.
- Minor: under drop, the offline `StateVerifier` can report spurious `ZombieTransaction` anomalies (a paired start event persists while its outcome drops) — a diagnostics-quality caveat to note there.

## Fixes

Closes #4466

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/channel-safety.md
+++ b/.claude/rules/channel-safety.md
@@ -1,5 +1,5 @@
 ---
-description: Prevent bounded channel deadlocks — cascading backpressure from .send().await in event loops has caused 4+ production incidents
+description: Prevent bounded channel deadlocks — cascading backpressure from .send().await in event loops has caused 5+ production incidents
 paths:
   - "crates/core/src/node/**"
   - "crates/core/src/transport/**"
@@ -13,7 +13,7 @@ paths:
 
 `.send().await` on a bounded channel inside an event loop or recv loop.
 When the receiver is slow or blocked on another channel, the sender blocks,
-stalling the entire event loop. This has caused 5+ production deadlocks:
+stalling the entire event loop. This has caused 6+ production deadlocks:
 
 - **StreamRegistry** (Mar 2026): `mpsc::channel(64)` whose receiver was never
   consumed. After 64 streams, `.send().await` blocked forever, deadlocking
@@ -78,6 +78,22 @@ stalling the entire event loop. This has caused 5+ production deadlocks:
   receiver ack) is corrected by the periodic InterestSync summary
   exchange and by the delta-apply-failure → ResyncRequest path that
   clears the sender's cached summary.
+
+- **Event-log channel send** (#4466, Jun 2026): `EventRegister::register_events`
+  and `notify_of_time_out` (`tracing.rs`) did an untimed `.send().await` on the
+  bounded event-log `mpsc::channel(1000)`, awaited from the network event loop's
+  hot outbound path (`p2p_protoc.rs` `OutboundMessageWithTarget` and disconnect
+  handlers). The sole consumer `record_logs` blocks on its metrics-server
+  WebSocket send (`ws://127.0.0.1:55010`, present on peers running the fdev
+  metrics server) and its AOF write; when that sink stalled, the channel filled
+  and the event loop blocked forever, parking every runtime thread on a futex at
+  0% CPU — a silent freeze (process alive, API port open, logging frozen
+  mid-activity, no crash so systemd never restarts). This is the sibling send
+  the #4145/#4231 fix missed, three lines above the peer-send it *did* wrap.
+  Fixed (#4467) by switching both sends to `try_send` with drop-on-full
+  (best-effort telemetry) plus a `DROPPED_EVENT_LOGS` counter; the `trace-ot`
+  `OTEventRegister` siblings were converted too. Observed on a 0.2.75 peer and
+  corroborated by a real-user diagnostic report on 0.2.74.
 
 ## Exception: same-runtime internal consumers
 
--- a/.claude/rules/channel-safety.md
+++ b/.claude/rules/channel-safety.md
@@ -1,5 +1,5 @@
 ---
-description: Prevent bounded channel deadlocks — cascading backpressure from .send().await in event loops has caused 5+ production incidents
+description: Prevent bounded channel deadlocks — cascading backpressure from .send().await in event loops has caused 6+ production incidents
 paths:
   - "crates/core/src/node/**"
   - "crates/core/src/transport/**"
--- a/crates/core/src/tracing.rs
+++ b/crates/core/src/tracing.rs
@@ -1717,6 +1717,12 @@ impl EventFlushHandle {
     /// Request a flush and wait for completion
     pub async fn flush(&self) {
         let (tx, rx) = tokio::sync::oneshot::channel();
+        // DELIBERATE blocking send (channel-safety.md "same-runtime internal
+        // consumer" exception): `flush` is a shutdown/test synchronization
+        // barrier — it MUST wait for `record_logs` to drain, not drop. It is
+        // never called from the network event loop (only shutdown/test paths),
+        // and the reply wait below is timeout-bounded. This is intentionally
+        // NOT converted to `try_send` like the hot-path event-log sends.
         if self.sender.send(EventLogCommand::Flush(tx)).await.is_ok() {
             // Best-effort flush: timeout or channel error is acceptable
             let _flush_result = tokio::time::timeout(std::time::Duration::from_secs(2), rx).await;
@@ -1913,6 +1919,11 @@ fn note_dropped_event_log() {
     }
 }
 
+#[cfg(test)]
+fn dropped_event_log_count() -> u64 {
+    DROPPED_EVENT_LOGS.load(std::sync::atomic::Ordering::Relaxed)
+}
+
 impl NetEventRegister for EventRegister {
     fn register_events<'a>(
         &'a self,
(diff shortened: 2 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #4467

### Summary
- **PR Title:** fix(tracing): make event-log registration non-blocking to prevent node deadlock
- **Type:** fix
- **Review tier:** Full (concurrency / async — event loop + bounded-channel backpressure)
- **Reviewers run:** Claude `code-first`, `testing`, `skeptical`, `big-picture`; **external:** Codex (gpt-5.5) on the final HEAD. Gemini was attempted as a second external model but returned capacity-exhausted (429); Codex hit its usage limit on the first attempt and was re-run successfully after reset on the final HEAD.
- **Linked issue:** #4466

This review was run across two HEADs: the original `68a7eac2`, then re-run on `d912a3b4` after the round-1 findings were addressed. Findings and resolutions below are consolidated.

---

### Code-First Analysis
**Alignment:** matches intent. The single `EventRegister` impl change covers all five event-loop call sites (`p2p_protoc.rs:272,1533,4997,5068,5422,5435`), all routing through `register_events`. The deadlock thesis was independently confirmed: `register_events` is awaited inline in the event loop's `OutboundMessageWithTarget` arm, and the sole consumer `record_logs` blocks on its metrics-WS / AOF write.
**Resolved:** PR description overstated that the drop counter "lands in `freenet service report`" — reworded to clarify it surfaces as a `tracing::warn!` in captured logs, with the structured `NodeDiagnostics` gauge as an explicit follow-up. "hangs without fix" reworded to "fails deterministically via the paused-clock timeout". `EventFlushHandle::flush` now carries an inline channel-safety exception comment.

### Testing Assessment
**Coverage:** adequate. Four deterministic unit tests; the regression pair was independently verified to FAIL (`Elapsed`) against the old blocking send and PASS with the fix.

| Unit | ✅ | 4 tests: non-block ×2, drop-counter increment, closed-channel |
| Integration / Simulation | N/A | unit level is correct here (single-function contract); a sim repro would be slow/non-deterministic for marginal signal |

**Resolved:** added the drop-counter test (lower-bound assertion, order-independent under parallel execution — verified across 3 parallel runs) and the `Closed`-branch test, per round-1 testing findings.

### Skeptical Findings
**Risk:** low after fixes.

| `OTEventRegister` had the identical blocking send | Medium | tracing.rs (trace-ot) | **Fixed** — converted to `try_send` (round 2) |
| round-2 `let _ = try_send` broke `clippy --features trace-ot` (`let_underscore_must_use` deny; invisible to default CI) | Medium (blocking) | tracing.rs (trace-ot) | **Fixed** (round 3, `fb761171`) — `match` + `#[allow]` discard, matching crate convention; my OT lines now clippy-clean under `--features trace-ot` |
| dropped batch → spurious StateVerifier `ZombieTransaction` under load | Low | state_verifier | Diagnostics-quality only; noted as follow-up |
| router refresh trains on lossy subset under load | Low | ring.rs | Verified NOT a correctness bug (live router fed independently) |
| `SubscribeTimeout` "scope creep" | — | — | **Dropped** — mis-attribution; pre-existing on main (#4456), not in this diff (verified) |

### Big Picture Assessment
**Goal alignment:** yes — correct layer (event-loop-side non-blocking is the right fix; consumer-side timeout + watchdog are deferred as defense-in-depth follow-ups, not substitutes). **Scope:** focused, single logical change. **Anti-patterns:** none. **Removed code:** none.
**Resolved:** the incident is now recorded in `.claude/rules/channel-safety.md` per repo convention (matching the #4145/#4231 same-commit precedent), with the count bumped (header/body reconciled to "6+" in round 3).

### External Model (Codex, gpt-5.5)
Reviewed the diff against `origin/main` on the final HEAD `d912a3b4`: "The change replaces hot-path bounded-channel awaits with non-blocking sends and adds targeted regression coverage. I did not identify any discrete introduced bug in the diff."

---

### Note on `trace-ot` clippy
`cargo clippy --features trace-ot` reports two **pre-existing** `wildcard_enum_match_arm` errors (`tracing.rs:1627`, `:1694`) that exist on `origin/main` and are unrelated to this PR; `trace-ot` clippy is not a CI gate. This PR's own OT lines are clippy-clean under that feature.

---

### Documentation
- Code docs: complete (every new item commented; the deliberate `flush` exception is documented inline).
- Rules/incident log: `channel-safety.md` updated.

### Recommendations
#### Must Fix (Blocking)
- ~~`let _ = try_send` deny-lint failure under `trace-ot`~~ — **fixed in round 3** (`match`).

#### Should Fix
- All addressed (OTEventRegister, counter/closed tests, flush comment, channel-safety entry, description accuracy).

#### Consider (follow-up issue, not this PR)
1. Out-of-band event-loop liveness watchdog (cause-agnostic freeze backstop; relates to #4233).
2. Consumer-side timeouts on `record_logs` (WS + AOF).
3. Pull-based `event_log_dropped_total` gauge on `NodeDiagnostics`.
4. Note in `StateVerifier` that log-channel saturation can produce spurious zombie/unpaired-outcome anomalies.

---

### Verdict
**State:** Ready to Merge (pending CI green on the current HEAD).
**HEAD SHA reviewed:** fb761171 (final; round-3 commit applied the trace-ot clippy fix).

All blocking and should-fix findings are resolved; the two external attempts plus four Claude perspectives agree the fix is correct, focused, and faithfully tested. The remaining items are intentionally-deferred defense-in-depth follow-ups.

[AI-assisted - Claude]

Issue #3798, opened 2026-04-08
hardening: audit and cap anti-write-amplification gaps in interest sync and broadcast paths


iduartgomez:
## Background

PR #3793 fixed a targeted-vs-broadcast bug (#3791) that caused O(N × fan-out) traffic. While auditing that fix, a review of the existing anti-amplification mechanisms revealed several gaps that could produce similar or worse amplification under adversarial or buggy conditions.

This issue tracks hardening work across those gaps.

---

## What exists today (good)

- `SyncStateToPeer` dispatches to exactly one peer (vs `BroadcastStateChange` fan-out)
- `BroadcastQueue`: bounded concurrency (12 small / 2 large streams), 256-entry queue with LRU eviction, per-(contract, peer) deduplication
- Echo-back suppression: skips sending if remote summary already matches local
- Heartbeat spreading: 28 peers × 5-min interval = ~10.7 s between each `Interests` send — prevents burst
- `MAX_BROADCAST_RETRIES = 3`, `MAX_BROADCAST_STREAK_ENTRIES = 256`
- `MAX_DOWNSTREAM_SUBSCRIBERS_PER_CONTRACT = 512` in `hosting.rs`
- Interest TTL (20 min) + sweep (60 s) to expire dead peer interest entries
- Simulation-time `BroadcastStorm` detector in `state_verifier/detectors.rs`

---

### Gap 1 — No cap on stale contracts synced per `Summaries` message (node.rs:1411)

`handle_interest_sync_message` iterates `stale_contracts` and emits one `SyncStateToPeer` per entry with no upper bound. A peer whose summary diverges on 100 contracts triggers 100 sync events in one handler invocation. Still O(1 peer) but unbounded per message.

**Suggested fix:** Cap at e.g. `MAX_STALE_SYNCS_PER_SUMMARIES = 32` per message, prioritising by some metric (e.g. contracts with most downstream subscribers). Log/metric when the cap is hit.

### Gap 2 — No size cap on `interested_peers` per contract (interest.rs:388)

`register_peer_interest` inserts directly with no size check. Compare with `MAX_DOWNSTREAM_SUBSCRIBERS_PER_CONTRACT = 512` in `hosting.rs` which does check. An adversary with many IPs could inflate this map, increasing fan-out cost of every subsequent `BroadcastStateChange` for that contract.

[...]

PR #4468, opened 2026-06-17
fix(broadcast): cap stale-contract sync emission per Summaries message

sanity:
## Problem

`handle_interest_sync_message` (the `InterestSync` → `Summaries` handler in
`crates/core/src/node.rs`) detects which contracts a peer is stale on by
comparing the peer's reported summaries against our local state, collecting them
into `stale_contracts`, then emitting one `SyncStateToPeer` event per stale
contract — **with no upper bound per message**.

A peer whose summary diverges on N contracts (a node that has been offline,
or a crafted `Summaries` message) therefore triggers N `SyncStateToPeer`
emissions in a single handler invocation. This is still O(1 peer) — unlike
`BroadcastStateChange`, it targets only the reporting peer — but the per-message
burst is unbounded. Each emission becomes a state-streaming job, so a large burst
stacks load on the notification channel and broadcast queue exactly when the node
is already doing summary-exchange work for every peer in the heartbeat cycle.

This is Gap 1 of the anti-write-amplification hardening audit in #3798, and one
of the gates for safely re-enabling the placement migration (#4440).

## Approach

Cap the number of `SyncStateToPeer` events emitted per `Summaries` message at
`MAX_STALE_SYNCS_PER_SUMMARIES = 32`.

- **Value justification:** consistent with the existing per-message burst-control
  family — `MAX_BROADCAST_RETRIES = 3`, `MAX_BROADCAST_STREAK_ENTRIES = 256`,
  `MAX_DOWNSTREAM_SUBSCRIBERS_PER_CONTRACT = 512`. 32 bounds the burst well below
  those while staying comfortably above the handful of stale contracts a healthy
  peer reports in one exchange. It is also the value suggested in the issue.

- **The cap bounds *emitted* events, not loop iterations.** Banned contracts and
  contracts with no local state are skipped without consuming the budget (they
  emit nothing), so the number of `SyncStateToPeer` events is hard-bounded by the
  cap regardless of how many entries are in `stale_contracts`.

- **Channel safety unchanged.** The emit site already uses
  `try_notify_node_event` (non-blocking `try_send`), per
  `.claude/rules/channel-safety.md`. This change only adds a counter and a
  `break`; it introduces no `.send().await`.

### Eventual consistency of the capped overflow

There is **no backlog queue** — nothing is dropped permanently. The stale set is
re-derived from durable state every cycle:

1. `ring::Ring::interest_heartbeat` resends the full `Interests { hashes }` set to
   each peer every 5 minutes (`INTEREST_HEARTBEAT_INTERVAL`), spread across the
   interval.
2. The peer replies with `Summaries`, and this handler re-compares each summary
   against current local state, re-detecting whatever is still divergent.
3. Any contract over the cap this cycle is simply synced on a later cycle.

A contract drops out of the stale set once its state has been synced and its
summary matches, so converged contracts free cap slots for the overflow. The
60 s interest sweep (`InterestManager::sweep_expired_interests`) plus the
delta-apply-failure → `ResyncRequest` path remain as additional re-detection
routes.

**Starvation avoidance (addressing the codex P2 review finding).** The first
draft processed `stale_contracts` in a fixed order and stopped at the cap. As
codex pointed out, that is not actually starvation-free: if a contract in the
leading `cap` positions stays stale across cycles — its `SyncStateToPeer` is
dropped on a full channel, lost in transit, or not applied by the peer — that
prefix re-consumes the whole budget every cycle and contracts past the cap are
*never* attempted. The "converged contracts free slots" argument only holds if
the prefix actually converges.

Fixed by rotating the stale set by a random offset (`GlobalRng::random_range`,
deterministic under simulation) before the capped loop, but only when
`total_stale > cap` (no behavior change in the common under-cap case). The cap
window then slides across the whole set over successive cycles, so each over-cap
contract is eligible with independent probability `cap / total_stale > 0` every
cycle — bounded expected wait regardless of whether the rest of the set
converges. Two model-level tests prove the union of emitted indices over all
rotation starts covers the entire set, and that a stuck head index does not
block the tail.

## Testing

New tests in `crates/core/src/node.rs` (`mod tests::stale_sync_cap`):

- `emit_budget_caps_at_max` — boundary behavior of `stale_sync_emit_budget`
  (0, 1, cap-1, cap, cap+1, cap*100).
- `many_stale_contracts_emit_at_most_cap` — the core regression: a peer diverging
  on `cap * 100` contracts emits at most `cap` events (not one per contract).
- `exactly_cap_emits_all` / `below_cap_emits_all_emittable` — boundary coverage.
- `skips_do_not_consume_budget` — banned / no-state contracts are skipped without
  consuming the emit budget.
- `rotation_covers_every_contract_over_cap` — every contract is reachable for some
  rotation start (no permanent starvation).
- `stuck_prefix_does_not_block_tail_under_rotation` — a stuck head index does not
  block the tail under rotation.
- `interleaved_skips_still_capped` — the emit count stays capped even with skips
  interleaved among emittable contracts.
- `stale_sync_loop_uses_emit_budget_pin` — source-scrape pin asserting the live
  `for contract in stale_contracts` loop still computes the budget, breaks at it,
  and counts emissions; guards against a refactor silently dropping the cap and
  re-opening the amplification burst.

The behavioral tests run against `count_stale_syncs_emitted`, a `#[cfg(test)]`
pure model of the loop's cap accounting; the pin test keeps that model and the
live loop in lockstep.

Also updates the sibling `interest_sync_summaries_gates_banned_egress` pin test
in `contract_ban_list.rs` to bound its source scan to the loop body
structurally and to anchor on the loop header *with its opening brace*
(`for contract in stale_contracts {`) — the old fixed 2,000-byte window would
have been pushed past the ban gate by the cap's `warn!` block, and the
brace-less anchor would otherwise match the new rustdoc prose. The
gate-before-emit assertion itself is unchanged; this is strictly a robustness
improvement.

## Review

Reviewed with `codex review --base main` (external, non-Claude model) across the
iterations:

- **Round 1 (cap only):** codex flagged a [P2] — a deterministic prefix-only
  loop is not starvation-free if a leading contract stays stale. Addressed by the
  random rotation above.
- **Round 2 (cap + rotation):** codex flagged a [P1] on the ban-list pin test
  anchoring on the new rustdoc prose. Verified the test still passed in practice,
  but addressed it anyway by anchoring on the loop header with brace.

Both findings fixed; production cap + rotation logic is unchanged between rounds.

Local validation: `cargo fmt --all`, `cargo clippy --locked -- -D warnings`
(the CI invocation) both clean; all new tests plus the affected `interest_sync`
pin tests pass.

Refs #3798 (Gap 1). Part of the hardening gates for #4440.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1811,6 +1811,112 @@ where
     }
 }
 
+/// Maximum number of stale-contract `SyncStateToPeer` events emitted per
+/// `Summaries` message handled (#3798 Gap 1, anti-amplification hardening).
+///
+/// A single peer whose summary diverges on N contracts would otherwise trigger
+/// N `SyncStateToPeer` emissions in one `handle_interest_sync_message` call —
+/// still targeted at one peer (O(1) peer, unlike `BroadcastStateChange`), but
+/// an unbounded burst per message. Capping the per-message burst keeps the
+/// notification channel responsive under a divergent (or crafted) summary.
+///
+/// Eventual consistency is preserved without a backlog queue: staleness is
+/// re-derived every heartbeat cycle from the durable summary comparison in the
+/// `Summaries` arm below (driven by the 5-minute interest heartbeat in
+/// `ring::Ring::interest_heartbeat`, expiry-swept every 60 s by
+/// `InterestManager::sweep_expired_interests`). Any contract over the cap this
+/// cycle is re-detected and synced on a later cycle.
+///
+/// Starvation avoidance: when the stale set exceeds the cap, the emission loop
+/// starts at a random offset and wraps (see the `rotate_left` in the `Summaries`
+/// arm), so the cap window slides across the whole stale set over successive
+/// cycles instead of always re-processing the same prefix. Without this, a
+/// contract stuck in the leading `cap` positions — e.g. one whose
+/// `SyncStateToPeer` is dropped on a full channel, lost in transit, or not
+/// applied by the peer — would re-consume the budget every cycle and
+/// permanently starve every contract past the cap. Random rotation makes each
+/// over-cap contract eligible with independent probability each cycle, so its
+/// expected wait is bounded regardless of whether the rest of the set
+/// converges.
+///
+/// Value chosen to match the per-message burst-control family of existing
+/// caps (`MAX_BROADCAST_RETRIES = 3`, `MAX_BROADCAST_STREAK_ENTRIES = 256`,
+/// `MAX_DOWNSTREAM_SUBSCRIBERS_PER_CONTRACT = 512`): 32 bounds the burst well
+/// below those while staying comfortably above the typical handful of stale
+/// contracts a healthy peer reports in one summary exchange.
+const MAX_STALE_SYNCS_PER_SUMMARIES: usize = 32;
+
+/// The per-message budget of stale-contract `SyncStateToPeer` emissions
+/// (#3798 Gap 1): `min(stale_contracts_len, MAX_STALE_SYNCS_PER_SUMMARIES)`.
+///
+/// Returns the maximum number of events the `Summaries` handler may emit this
+/// call. The caller increments an `emitted` counter only for contracts it
+/// actually emits for (banned / no-local-state contracts are skipped without
+/// consuming the budget) and stops once `emitted` reaches this value, so the
+/// number of `SyncStateToPeer` events is hard-bounded by
+/// [`MAX_STALE_SYNCS_PER_SUMMARIES`] regardless of `stale_contracts_len`.
+fn stale_sync_emit_budget(stale_contracts_len: usize) -> usize {
+    stale_contracts_len.min(MAX_STALE_SYNCS_PER_SUMMARIES)
+}
+
+/// Per-contract disposition in the stale-sync emission loop, used to model the
+/// loop's cap accounting in a unit test without constructing an `OpManager`.
+#[cfg(test)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum StaleSyncDisposition {
+    /// Contract is not banned and has local state — emits a `SyncStateToPeer`
+    /// (and so consumes one unit of the emit budget).
+    Emit,
+    /// Contract is banned — skipped without emitting or consuming the budget.
+    Banned,
+    /// No local state available — skipped without emitting or consuming the
+    /// budget.
+    NoState,
+}
+
+/// Pure model of the stale-sync emission loop's cap accounting (#3798 Gap 1).
+///
+/// Mirrors the `for contract in stale_contracts` loop in the `Summaries` arm of
+/// [`handle_interest_sync_message`]: break once `emitted` reaches the budget;
+/// `Banned` / `NoState` contracts are skipped without consuming the budget;
+/// every `Emit` before the budget is exhausted counts. Returns the number of
+/// `SyncStateToPeer` events the real loop would emit for the given sequence.
+///
+/// Kept in lockstep with the production loop by the
+/// `stale_sync_loop_uses_emit_budget_pin` source-scrape test, which asserts the
+/// loop still applies this budget-and-break structure.
+#[cfg(test)]
+fn count_stale_syncs_emitted(dispositions: &[StaleSyncDisposition]) -> usize {
+    let budget = stale_sync_emit_budget(dispositions.len());
+    let mut emitted = 0usize;
+    for d in dispositions {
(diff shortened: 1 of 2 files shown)


sanity:
## Review: codex (gpt-5.5) pass + addressed

Ran `codex review --base main` (external, non-Claude model) on the initial commit. It returned one finding:

**[P2] Rotate capped stale-sync attempts to avoid starvation** (node.rs, the `for contract in stale_contracts` loop). The deterministic prefix-only `break` means that if any of the first `cap` (32) stale contracts stays stale across cycles — `try_notify_node_event` drops on a full channel, the packet is lost, or the peer fails to apply — that prefix re-consumes the cap every cycle and contracts past the cap are never attempted. My original "converged contracts free slots" eventual-consistency argument only holds if the prefix actually converges.

**Addressed** (pushed, HEAD now `607c1ee7`): when `total_stale > cap`, rotate the stale set by a random offset (`GlobalRng::random_range`, deterministic under simulation) before the capped loop, so the cap window slides across the whole set over successive cycles. Each over-cap contract is then eligible with independent probability `cap / total_stale > 0` per cycle — bounded expected wait regardless of whether the rest of the set converges. No behavior change in the common under-cap case (no rotation when `total_stale <= cap`).

Added two model-level regression tests: `rotation_covers_every_contract_over_cap` (union of emitted indices over all rotation starts covers the whole set) and `stuck_prefix_does_not_block_tail_under_rotation` (a stuck head index does not block the tail). The source-scrape pin test now also asserts the rotation is wired in, so a future refactor that drops it fails CI.

Re-running codex review on the amended content; will confirm clean. `cargo fmt` + `cargo clippy --locked -- -D warnings` are clean locally.

[AI-assisted - Claude]

sanity:
## Review round 2: codex P1 (test anchor) — addressed

Re-ran `codex review --base main` on the rotation-fix commit. One finding:

**[P1] Anchor stale-loop search to the actual loop** (contract_ban_list.rs). codex observed that my new rustdoc in node.rs contains the literal `for contract in stale_contracts` in prose, so the existing pin test's `NODE_SRC.find("for contract in stale_contracts")` matches the doc comment before the production loop.

I verified empirically that the test still **passed** (locally and would on CI): the pin searches for the full `contract_ban_list.is_banned` and `NodeEvent::SyncStateToPeer` strings, which only appear in the real loop — the rustdoc says `\`SyncStateToPeer\`` (backticked, no `NodeEvent::` prefix), so the gate-before-emit ordering still held. So the "deterministic test failure" did not actually occur.

But codex's underlying point is right: anchoring a pin test's scan window on a doc comment is fragile (a future doc edit could break it). **Fixed** (HEAD now `c4b1bf99`): both pin tests now anchor on the loop header *with its opening brace* (`for contract in stale_contracts {`), which matches only the production loop, not the prose. Verified by replicating the search logic — both resolve to the real loop and all assertions hold. All 16 unit tests pass; `cargo clippy --locked -- -D warnings` clean.

The production logic (cap + random rotation) is unchanged from the round-1 fix.

[AI-assisted - Claude]

sanity:
## Review round 3: codex clean

Re-ran `codex review --base main` on the final content (HEAD `c4b1bf99`). Verdict:

> The change caps stale SyncStateToPeer emissions per Summaries message while preserving retry behavior via subsequent heartbeat-derived summaries. I did not identify any discrete regression introduced by the patch.

Both prior findings are resolved (round-1 [P2] starvation → random rotation; round-2 [P1] pin-test anchor → brace-anchored). The external non-Claude review pass is clean. Waiting on CI green on `c4b1bf99` before enabling auto-merge.

[AI-assisted - Claude]

sanity:
## Status: READY TO MERGE — PARKED (do not auto-merge)

All gates green on HEAD `c4b1bf99`:

- **CI: fully green.** Clippy, Fmt, Conventional Commits, Rule Lint, Claude Rule Review, Unit & Integration (10m30s), Simulation (12m2s), Windows Check, NAT Validation, macOS E2E, CLA, Snyk — all pass. (Claude CI Analysis / dev-room-notify are conditional SKIPPED, non-gating.)
- **Review: clean.** `codex review --base main` across 3 rounds; both findings (P2 starvation → random rotation; P1 pin-test anchor → brace-anchored) addressed; round 3 found no regressions.

**Intentionally NOT enabling auto-merge.** This PR is part of a cluster being consolidated under another agent, and there is a live production incident in flight. Holding out of the merge queue until @sanity (or the consolidating agent) sequences it. Please land when ready — no further changes are needed from my side.

[AI-assisted - Claude]

PR #4470, opened 2026-06-17
feat(telemetry): add node fd-usage gauges to router_snapshot (#4440)

sanity:
## Problem

The v0.2.73 placement-migration incident drove a gateway crash-loop from **file-descriptor exhaustion** — the WASM module cache's memfds collided with systemd's 1024 `RLIMIT_NOFILE` soft limit, producing `EMFILE` and killing a monitored background task. The failure was **nearly invisible to central telemetry**: the fatal `os error 24` lived only in per-node file logs, so operators had no signal that fd pressure was building. #4440 tracks adding low-volume node-health gauges so this class of resource exhaustion is observable *before* it becomes fatal.

This is the single highest-value missing signal from that list.

## Approach

Emit two gauges on the **existing 5-minute `router_snapshot` cadence** (already forwarded to the OTLP collector, so no new event type, timer, or wiring):

- `open_fds` — this process's current open file-descriptor count (Linux, via `/proc/self/fd`).
- `fd_soft_limit` — the `RLIMIT_NOFILE` soft limit, i.e. the ceiling that triggers `EMFILE` (any-unix, via `getrlimit`).

Together they let an operator watch fd headroom (`open_fds / fd_soft_limit`) and alert before exhaustion. Both are best-effort and `None` on platforms that can't report them cheaply; the `getrlimit` read mirrors the already-vetted `bin/freenet.rs::raise_fd_limit`.

Design notes for reviewers:

- Values are populated by `Ring` on the snapshot cadence (like the existing `connect_forward_*` enrichment), not by the `Router` model.
- The `router_snapshot` OTLP body (`tracing/telemetry.rs::event_kind_to_json`) is a **hand-mirrored `json!` block**, so new `RouterSnapshotInfo` fields must be added there explicitly or they silently never reach the collector — the #4009/#4010 manually-mirrored-telemetry footgun. A regression test pins this.
- No new `NetMessageV1`/`EventKind` variant — fields are added to the existing `RouterSnapshot` payload, so there is no wire-format or match-arm churn.

## Testing

- `ring::fd_usage_tests::read_fd_usage_reports_sane_values_on_linux` — asserts the Linux path reports a positive open-fd count and `open_fds <= fd_soft_limit` (a kernel invariant, which also pins that the two values are read the right way round).
- `tracing::telemetry::tests::router_snapshot_json_includes_fd_gauges` — builds a `RouterSnapshot` event and asserts both gauges appear in the serialized OTLP body, guarding against the hand-mirror silently dropping them.
- `cargo fmt`, `cargo clippy --locked -- -D warnings` (CI parity), and the full lib test build are clean.

Part of #4440.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1061,11 +1061,20 @@ impl Ring {
                 snapshot.connect_forward_peer_adjustments = Some(adjustments);
             }
 
+            // Node-health gauges (#4440): make fd-exhaustion headroom observable
+            // on the existing snapshot cadence. Best-effort — `None` where the
+            // platform can't report it (see `read_fd_usage`).
+            let (open_fds, fd_soft_limit) = read_fd_usage();
+            snapshot.open_fds = open_fds;
+            snapshot.fd_soft_limit = fd_soft_limit;
+
             tracing::info!(
                 failure_events = snapshot.failure_events,
                 success_events = snapshot.success_events,
                 prediction_active = snapshot.prediction_active,
                 consider_n_closest_peers = snapshot.consider_n_closest_peers,
+                open_fds = ?snapshot.open_fds,
+                fd_soft_limit = ?snapshot.fd_soft_limit,
                 "router_snapshot"
             );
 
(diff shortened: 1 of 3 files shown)


sanity:
## Review summary (Light tier — additive telemetry, low-risk surface)

Two independent passes, both on the final content (HEAD `5a93d207`):

**External — Codex (`codex review --base main`):** _"did not identify any actionable bugs in the diff."_

**Adversarial — skeptical reviewer:** no blocker/major findings. Empirically verified: the `saturating_sub(1)` fd accounting is exact (raw count includes the `read_dir` stream fd), the `getrlimit` unsafe block matches the vetted `raise_fd_limit` pattern, cfg gating compiles on every target (linux-gnu/musl/macOS/windows), the tests are deterministic, and the hand-mirror pin test genuinely fails when the json fields are removed (not vacuous). Two low-severity nits:

1. **fd-count goes blind at exactly the limit** (`read_dir` needs an fd → `EMFILE` → `None`). Inherent to any procfs fd count. **Addressed:** added a doc caveat noting the companion `fd_soft_limit` still reports and the 5-min cadence captures the *approach*, which is the actionable signal.
2. **Hard limit not emitted.** **Dismissed (justified):** the soft limit is the operative `EMFILE` ceiling, and `bin/freenet.rs::raise_fd_limit` raises soft→hard at startup, so `rlim_max` is redundant in practice.

Verification: `cargo fmt` clean, `cargo clippy --locked -- -D warnings` (CI parity) clean, both new tests pass.

[AI-assisted - Claude]

PR #4472, opened 2026-06-17
feat(telemetry): add module-cache occupancy + eviction gauges (#4440)

sanity:
## Problem

The v0.2.73 / #4441 incident was a compiled-WASM **module-cache thrash**: on nodes hosting >1024 contracts every cold contract evicted a warm one and forced a Cranelift recompile, pinning CPU and (on small hosts) OOM-spiraling. #4452 fixed the *bound* (byte-budget instead of a 1024 count cap), but the cache *pressure* stayed invisible to central telemetry — an operator had no way to see a node thrashing until it fell over. This is the next gauge on the #4440 observability checklist (after the #4470 fd gauges).

## Approach

Emit per-cache occupancy + eviction gauges on the **existing 5-minute `router_snapshot` cadence** (the same OTLP-forwarded vehicle the fd gauges use):

- `contract_module_cache_*` and `delegate_module_cache_*`: `entries`, `total_bytes`, `budget_bytes`, `evictions_total`.
- `evictions_total` is a **monotonic lifetime counter** — the collector differences it across the cadence to derive an eviction *rate*. It is deliberately distinct from `ModuleCache::evictions_in_window`, which resets every warn window and so can't serve as a cumulative signal.

The caches live behind the contract-handler channel, unreachable from the `Ring` snapshot task, so — exactly like `TRANSPORT_METRICS` — they **publish** into a process-global (`MODULE_CACHE_METRICS`) via lock-free `Relaxed` atomics in `insert`/`evict_to_budget`/`remove` (the caller already holds the cache `Mutex`), and the snapshot task **reads** it. Routing is by the cache's existing `"contract"`/`"delegate"` label; unlabeled/test caches don't publish. Contract and delegate are reported **separately** so a delegate-cache thrash can't hide behind a healthy contract cache. Assumes one runtime pool per process (the production invariant), same as `TRANSPORT_METRICS`.

Notes for reviewers:

- Adding the fields tipped `EventKind::RouterSnapshot` over clippy's `large_enum_variant` threshold, so that variant is now **boxed**. `Box` serializes transparently, so the AOF/OTLP wire format is unchanged, and boxing also shrinks every other `EventKind` variant.
- The `router_snapshot` OTLP body (`event_kind_to_json`) is a **hand-mirrored `json!` block** — new fields must be added there or they never reach the collector (the #4009/#4010 footgun). A regression test pins it.

## Testing

- `module_cache_metrics_routing_and_isolation` — label routing, contract/delegate isolation, monotonic-eviction vs last-write-wins-occupancy semantics, unknown-label no-op (deterministic, on a local instance).
- `evicting_contract_cache_increments_global_eviction_counter` — a `"contract"`-labeled cache that evicts bumps the process-global counter; asserts only the strict increase, which is race-safe (monotonic `fetch_add`, never resets) under concurrent tests.
- `router_snapshot_json_includes_module_cache_gauges` — all 8 fields reach the hand-mirrored OTLP body.
- `cargo fmt`, `cargo clippy --locked -- -D warnings` (CI parity), and the touched test suites are clean.

Part of #4440.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/wasm_runtime/module_cache.rs
+++ b/crates/core/src/wasm_runtime/module_cache.rs
@@ -39,6 +39,8 @@
 
 use std::borrow::Borrow;
 use std::hash::Hash;
+use std::sync::LazyLock;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{Duration, Instant};
 
 use lru::LruCache;
@@ -99,13 +101,20 @@ impl<K: Hash + Eq, V> ModuleCache<K, V> {
     /// Like [`Self::new`] but with a label ("contract" / "delegate") used in the
     /// rate-limited "cache is evicting at budget" operator warning.
     pub(crate) fn with_label(budget_bytes: usize, label: &'static str) -> Self {
+        let budget_bytes = budget_bytes.max(1);
+        // Publish the budget immediately, so an idle cache (a node that hasn't
+        // compiled a module yet, or a delegate cache that's never used) reports
+        // its configured budget instead of the atomic default 0. Otherwise the
+        // collector's occupancy-% denominator is 0 until the first insert/remove
+        // (#4440 — flagged by both PR reviewers).
+        MODULE_CACHE_METRICS.record_occupancy(label, 0, 0, budget_bytes);
         Self {
             // The LRU is unbounded by count; the byte budget is the only
             // bound. `usize::MAX` capacity means `LruCache` never evicts on
             // its own — we drive all eviction via the byte budget below.
             inner: LruCache::unbounded(),
             total_bytes: 0,
-            budget_bytes: budget_bytes.max(1),
+            budget_bytes,
             label,
             evictions_in_window: 0,
             window_started_at: None,
@@ -136,6 +145,12 @@ impl<K: Hash + Eq, V> ModuleCache<K, V> {
         }
         self.total_bytes = self.total_bytes.saturating_add(size_bytes);
         self.evict_to_budget();
+        MODULE_CACHE_METRICS.record_occupancy(
+            self.label,
+            self.inner.len(),
+            self.total_bytes,
+            self.budget_bytes,
+        );
     }
 
     /// Evict least-recently-used entries until `total_bytes <= budget_bytes`,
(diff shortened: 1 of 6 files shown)


sanity:
## Review summary (Light tier — observability gauges; global-atomics + boxing reviewed)

Two independent passes. Both flagged the **same single substantive issue**, now fixed.

**External — Codex (`codex review --base main`):** one [P2] — `budget_bytes` stays at the atomic default `0` until the first insert/remove, so an idle/never-used cache reports a real 0-byte budget and breaks the collector's occupancy-% denominator.

**Adversarial — skeptical reviewer:** no blockers/majors. Same Low finding (idle-cache budget=0) plus one nit. Verified safe: `Relaxed` atomics are sound for these gauges (no tearing; `evictions_total` is monotonic `fetch_add`, occupancy is last-write-wins and a 5-min gauge doesn't care about sub-µs field skew); the single-runtime-pool assumption holds in production and **DST/simulation nodes use mock handlers that never construct labeled caches**, so multi-node-in-one-process sims don't touch the global and determinism tests (which compare variant names / `router_snapshot_summary`, not the new fields) are unaffected; the boxing is transparent for serde/bincode/arbitrary (AOF/replay byte-identical) and every match site auto-derefs; the eviction-counter test is race-safe (asserts only the monotonic increase).

**Fixed (both reviewers):** `with_label` now publishes the budget at construction, so an idle cache reports its configured budget from node start. HEAD `ed569697`.

**Dismissed (nit, justified):** the `#[cfg(test)]` delegate-cache reassignment publishes into the global delegate gauges — harmless because no test asserts global delegate occupancy; the only asserted global is the monotonic eviction counter, which `record_occupancy` doesn't touch.

Re-verified after the fix: `cargo clippy --locked -- -D warnings` clean, module-cache + telemetry tests pass.

[AI-assisted - Claude]

crates/core/src/wasm_runtime/module_cache.rs
```diff
@@ -227,6 +252,122 @@ impl<K: Hash + Eq, V> ModuleCache<K, V> {
...
+/// The module caches live behind the contract-handler channel, unreachable from
+/// the `Ring` telemetry-snapshot task that emits `router_snapshot`. Like
+/// [`TRANSPORT_METRICS`](crate::transport::metrics::TRANSPORT_METRICS), the
+/// caches therefore *publish* into this process-global and the snapshot task
+/// *reads* it. A node builds exactly one contract cache and one delegate cache
+/// (`RuntimePool::new`), so one global pair of gauges is correct; this assumes a
+/// single runtime pool per process (the production invariant). The cache-thrash
+/// that drove the #4441 incident was invisible to central telemetry — these
+/// gauges make occupancy and eviction pressure observable.
+pub(crate) static MODULE_CACHE_METRICS: LazyLock<ModuleCacheMetrics> =
```
iduartgomez, line 267: @claude does this need be a static or can it live as instances wherever is called from?

Issue #4473, opened 2026-06-17
Storm: unbounded ResyncRequest/summarize on interested-but-stateless relays wedges gateways (nova) + times out GETs (technic)


sanity:
## Summary

Two live production incidents on 2026-06-17 — the **nova gateway wedge** (every PUT times out) and **technic's 30s River-UI GET timeouts** — are two severities of **one root cause**: an unbounded `summarize_contract_state` / `ResyncRequest` storm on relay peers that hold *interest* in a contract but have no local state for it. This issue consolidates the live evidence and tracks the proper fixes. It is the receiver-side complement to #4145 (sender-side full-state storm fixed in #4442) and a concrete instance of the #4440 placement-migration incident and the #3798 anti-write-amplification gap.

## The keystone bug

- The v0.2.73 placement migration (#4404) left relay peers with **interest** in contracts they don't host ("phantom contracts" — ~85 observed on technic, `ring/hosting.rs:2263`). Interest is decoupled from holding state.
- A hot River room emits UPDATE deltas at tens/sec. Each `BroadcastTo` delta reaches an interested-but-stateless relay → delta-apply fails (no local state) → the relay sends a `ResyncRequest` **with no backoff** (`crates/core/src/operations/update/op_ctx_task.rs:1130-1172`).
- The **full-state**-miss path already rate-limits via `CONTRACT_FETCH_COOLDOWN_MS = 300_000` (`update.rs:138`, `try_auto_fetch_contract`). The **delta**-miss path was missed, and it's the common case for established peers (they ship deltas, not full state).
- The broadcast dedup cache keys on exact payload bytes, so every distinct delta (every new room message) bypasses it. Net: the loop fires at the room's UPDATE rate, each iteration re-summarizing/re-fetching against absent state → WARN `Contract state not found in store` (`runtime.rs:2194`).

## Live evidence

**nova** (gateway, `0.2.75`, build `59dd6d619f22`):
- Wedged 2026-06-16 23:47 UTC, right after `summarize_contract_state: dropped 10504 in 30s`.
- UDP `0.0.0.0:31337` Recv-Q pinned at 33,555,456 bytes (32 MB SO_RCVBUF ceiling), not draining; other sockets Recv-Q 0 → network recv-loop wedged.
[...]

PR #4475, opened 2026-06-17
fix(interest-sync): only summarize hosted-or-served contracts (stop heartbeat summarize storm)

sanity:
## Problem

Production gateway **nova** wedged (~14h, every PUT timing out) and relay peers like **technic** time out GETs under a `summarize_contract_state` storm: ~40/sec sustained (~630k dropped by the rate limiter over 17h on technic), **fully decoupled from user activity** — technic took **<10 inbound UPDATEs/hour** (≈4,000× amplification).

**Root cause:** the periodic interest-sync handlers (`Interests`/`Summaries`/`ChangeInterests` in `handle_interest_sync_message`) summarize **every shared-interest contract on every 5-min heartbeat from every connected peer**, with no "do we hold this state" guard. The v0.2.73 placement migration (#4404) left relays with **phantom interest** in dozens of contracts they neither host nor serve, so each heartbeat sweeps the serial `contract_handling` loop with pointless `GetSummaryQuery` round-trips that return "state not found" (uncached) every time. That starves real GET/PUT/UPDATE ops (technic) and feeds the #4145 notification-channel saturation that wedges gateways (nova). The delta-fail `ResyncRequest` path was a **red herring** (naturally paced ~1-2/min).

## Solution

Route the three periodic interest-sync summarize sites through a new `summary_if_hosted_or_in_use` helper that calls `get_contract_summary` only when `ring.is_hosting_contract(key) || ring.contract_in_use(key)` — i.e. only for contracts we host or actively serve (a live local-client or downstream subscriber). Phantom-interest contracts return `None` without a loop round-trip. The `ResyncRequest` arm keeps its direct call (already state-gated, not heartbeat-driven).

Gating on `is_hosting` alone is too narrow (Codex review, P1): an evicted-but-in-use contract retains state, so hosting-only would drop its interest-sync heal. `|| contract_in_use` closes that gap — any contract with a live subscriber is summarized. A live-node check confirmed the phantom storm contracts are **not** `in_use` (their relay-subscribe leases had long expired), so this gate kills the storm while preserving heal for genuinely-subscribed contracts.

## Testing

- `interest_sync_periodic_arms_summarize_only_hosted_or_in_use_pin` — source-scrape regression pin: **fails on the pre-fix code** (bare `get_contract_summary` in the periodic arms), passes once they route through the gated helper which checks both `is_hosting_contract` and `contract_in_use`.
- `cargo check` / `fmt` / `clippy` clean.
- At-scale validation is by **canary** (nova/technic telemetry post-deploy): the wedge is a multi-task backpressure cycle the deterministic sim cannot reproduce, so a passing sim would be false confidence.
- Adds the interest-sync summarize invariant to `.claude/rules/ring.md`.

## Review

Full multi-model review (4 Freenet reviewer subagents + external Codex). Codex's P1 (`is_hosting` too narrow) was reconciled — in-use contracts are never evicted from the hosting cache, so the gap was a narrow self-healing transient — and then **fully closed** by adding `|| contract_in_use`. Follow-up tracked for a behavioral counter-based sim test (the source-scrape pin guards the wiring meanwhile).

Refs #4473

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -137,6 +137,18 @@ WHEN removing a connection (any failure path):
 WHEN exchanging peer lists in sync protocols (e.g., interest sync):
   → MUST filter out peers not currently in the live connection set
   → Stale peer entries cause operations to be sent to dead nodes
+
+WHEN summarizing contracts in the InterestSync heartbeat handlers
+(handle_interest_sync_message: Interests / Summaries / ChangeInterests):
+  → ONLY call get_contract_summary for contracts we host OR actively serve
+    (is_hosting_contract || contract_in_use) — go through
+    summary_if_hosted_or_in_use, never a bare get_contract_summary.
+  → A node carries phantom peer-interest in contracts it neither hosts nor
+    serves (placement-migration after-effect, #4404). Summarizing those issues a
+    GetSummaryQuery round-trip on the serial contract_handling loop that returns
+    "state not found" every heartbeat — the #4473/#4145 summarize storm
+    (~40/sec vs <10 real updates/hr) that wedged gateways. The ResyncRequest arm
+    is exempt: it is state-gated and not heartbeat-driven.
 ```
 
 ### Cleanup Exemptions Must Be Time-Bounded
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1874,7 +1874,12 @@ async fn handle_interest_sync_message(
             let mut entries = Vec::with_capacity(matching.len());
             for contract in matching {
                 let hash = contract_hash(&contract);
-                let summary = get_contract_summary(op_manager, &contract).await;
+                // Only summarize contracts we host or actively serve; phantom
+                // peer-interest contracts (no state, no live subscriber) have
+                // nothing to advertise and their pointless GetSummaryQuery
+                // round-trips were the dominant #4473 storm. See
+                // `summary_if_hosted_or_in_use`.
+                let summary = summary_if_hosted_or_in_use(op_manager, &contract).await;
                 entries.push(SummaryEntry::from_summary(hash, summary.as_ref()));
 
                 if let Some(ref pk) = peer_key {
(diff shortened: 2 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #4475 (consolidated)

**Review tier:** Full (high-risk surface: node event loop + contract-handling path).
**Reviewers run:** code-first, skeptical, testing, big-picture (freenet subagents) + external **Codex** — initial pass on the `is_hosting`-only version, then a re-review pass on the broadened `is_hosting || contract_in_use` gate.
**CI:** green on the prior reviewed HEAD; re-running on current HEAD `7fd187f4` (merge gated on it).

### Outcome
No High/Medium findings outstanding. The one blocking finding and how it was resolved:

- **[Codex P1 — resolved]** Gating on `is_hosting_contract` alone was too narrow: an evicted-but-still-in-use contract retains its on-disk state (`evict_over_budget` retains `contract_in_use` entries; `reclaim_evicted_contract` keeps state while in use), so hosting-only would have dropped its interest-sync stale-peer heal. **Fixed** by broadening the gate to `is_hosting_contract || contract_in_use`. Re-review confirmed: (a) P1 closed — any contract with a live subscriber (incl. the post-eviction subscriber race) is summarized; (b) storm NOT reintroduced — `contract_in_use` reads the subscription maps, disjoint from the `interested_peers` phantom-interest structure, and a live-node check confirmed the storm contracts are not in_use; (c) the previously-overstated "behavior unchanged" comments are now accurate.

### Addressed review feedback
- Honest comments + helper docstring (code-first, skeptical).
- `.claude/rules/ring.md` interest-sync summarize invariant added (big-picture, code-first).
- `Fixes #4473` → `Refs #4473` so the umbrella incident isn't auto-closed while #4145/#4441/#4440 remain (big-picture).
- Pin-test helper-slice tightened to the helper body so the gate assertions can't false-pass (skeptical, Low).
- Behavioral counter-based sim test deferred to **#4476** (testing reviewer's "Important, not Critical"); the source-scrape pin guards the wiring and at-scale validation is by canary — the multi-task backpressure wedge is provably not reproducible in the deterministic single-consumer sim.

### Verdict
**Ready to merge** once CI is green on `7fd187f4`. Reviewed HEAD: `7fd187f4` (re-review delta from `05baf7cc` was the test-slice tightening only). At-scale validation: canary on nova/technic telemetry post-release.

[AI-assisted - Claude]

PR #4481, opened 2026-06-17
perf(contract): raise module-cache default ceiling so large gateways stop thrashing (#4441)

sanity:
## Problem

A node hosting a large working set of contracts thrashes its compiled-WASM module cache: every cold contract evicts a warm one, and the evicted module must be recompiled on its next access, pinning the single-threaded `contract_handling` loop and (on resource-constrained hosts) growing RSS until OOM. This is issue #4441, observed in the broadcast-storm incident cluster (#4440): technic recompiled the River web container ~15× in 16h and intermittently 30s-timed-out GETs.

PR #4452 already did the bulk of #4441: it replaced the old hardcoded 1024-**entry** LRU with a **byte-budget** cache (default `clamp(total_ram / 8, 64 MiB, 384 MiB)`, delegate cache = 1/4 of that), wired it to `--module-cache-budget-bytes` / `FREENET_MODULE_CACHE_BUDGET_BYTES` / a config field, and offloaded cold compiles off the contract loop. The wasmtime **disk** compilation cache (#3456) is also already enabled, so an evicted module **deserializes** from disk on reload rather than recompiling from source.

The remaining gap is the **384 MiB MAX clamp**. The `total_ram / 8` divisor is what protects small boxes from OOM; the absolute MAX clamp is a separate ceiling that only binds on large hosts. On a production gateway (nova has 125 GiB RAM), `total_ram / 8` is ~15.6 GiB, so the MAX clamp is the *only* thing sizing the default cache — and 384 MiB holds only ~256 modules at the measured ~1.5 MiB each. A gateway's working set (hundreds of hosted contracts plus, post-#4404, an ~85-contract phantom-interest set) exceeds that, so the cache permanently evicts-and-recompiles. The operator *could* set the override, but the out-of-the-box default thrashes on exactly the nodes #4441 is about.

## Approach

Raise `MAX_DEFAULT_MODULE_CACHE_BUDGET_BYTES` from 384 MiB to **1.5 GiB**.

- **Small/medium hosts are unchanged.** The MAX clamp only binds above `MAX × 8 = 12 GiB` of RAM; below that the `total_ram / 8` divisor binds first. A 1 GiB VPS still defaults to 128 MiB; a 512 MiB box still gets the 64 MiB MIN. Raising MAX does not touch them.
- **Large gateways stop thrashing.** 1.5 GiB holds ~1000 modules at the measured ~1.5 MiB — comfortably above a real gateway working set — and matches the 1 GiB default hosted-*state* budget (`ring::DEFAULT_HOSTING_BUDGET_BYTES`): a node allowed ~1 GiB of state should be able to cache the corresponding compiled code.
- **Still bounded — not removed.** An unbounded `total_ram / 8` would default a 125 GiB box to a ~15 GiB cache it can't benefit from (working-set benefit flattens past ~1000 modules). The clamp keeps the default a bounded, defensible commitment; the explicit override remains unclamped for operators who genuinely need more.
- **cgroup-aware sizing (review hardening).** Raising the ceiling to 1.5 GiB widened a latent risk: a node in a small container (e.g. a 2 GiB cgroup limit on a 128 GiB host) would size its default from *host* RAM and could OOM the container. `read_total_ram_bytes` now takes the **min** of `/proc/meminfo` `MemTotal` and the cgroup memory limit (cgroup v2 `memory.max`, falling back to v1 `memory.limit_in_bytes`), treating the `max`/`~u64::MAX` "unlimited" sentinels as no-limit. A host systemd service (the production gateways) sees no limit and behaves exactly as before; a constrained container now gets a budget scaled to its real limit.

The wasmtime disk-cache "deserialize on eviction-reload" behavior (the other half originally scoped for this work) is **already in place** (`wasmtime_engine.rs` enables `Cache::new(CacheConfig::new())`; the reload path goes `prepare_contract_call` → `engine.compile` → `Module::new`, which deserializes on a `(config + WASM-hash)` hit). No code change needed; this PR adds a source-scrape pin test (isolated to the `create_backend_engine` body so it can't self-satisfy) so the behavior can't silently regress.

## Testing

- `large_gateway_default_exceeds_old_clamp` — regression for #4441: a nova-sized (125 GiB) host now defaults above the old 384 MiB ceiling and holds ≥900 modules. Fails on the old 384 MiB MAX.
- `combined_default_ceiling_is_safe_on_small_box` — sweeps 128 MiB…12 GiB RAM, asserts the combined default never exceeds total RAM (catches an unsafe MIN raise) and stays ≤ 1/4 of RAM, pinning that the divisor (not the MAX clamp) protects small boxes. Rewritten from the old `combined_default_ceiling_is_safe` (which asserted `< 512 MiB`, no longer the invariant).
- `max_clamp_combined_ceiling_is_safe_at_binding_host` — upper-bound guard: the combined ceiling at MAX stays ≤ 1/4 of the smallest host where MAX binds (12 GiB), so a future unsafe MAX raise fails CI.
- `parse_cgroup_limit_distinguishes_real_limits_from_sentinels` / `parse_meminfo_total_reads_kib_as_bytes` — unit tests for the new memory-limit parsing (real limits vs `max`/`~u64::MAX`/zero/garbage).
- `engine_enables_wasmtime_disk_compilation_cache` — source-scrape pin (isolated to `create_backend_engine`) that the wasmtime disk compilation cache stays enabled.
- Extracted a pure `budget_for_ram(total_ram)` helper so the small-box / large-box boundary is deterministically testable without depending on the test host's RAM.
- Updated `default_budget_is_within_clamp` and the config round-trip test to reference the MIN/MAX constants instead of hardcoded byte values, so they can't drift from the clamp.
- Fixed config.rs docstrings (clamp range; also corrected a wrong "both caches get the full budget" claim — the delegate cache gets 1/4).

`cargo fmt` + `clippy -D warnings` clean; wasm_runtime + config module-cache tests pass.

Refs #4441, #4440.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -120,12 +120,15 @@ pub struct ConfigArgs {
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
 
-    /// Byte budget for each compiled-WASM module cache (the contract cache and
-    /// the delegate cache each get this budget). When a cache's tracked
-    /// compiled-byte total would exceed this on insert, least-recently-used
-    /// modules are evicted until it fits. Bounding by bytes (not entry count)
-    /// stops a node hosting many contracts from thrashing the cache and
-    /// recompiling on every access (issue #4441). Default: 384 MiB.
+    /// Byte budget for the compiled-WASM **contract** module cache. The
+    /// **delegate** cache gets a fraction of this value
+    /// (`DELEGATE_MODULE_CACHE_BUDGET_DIVISOR`, currently 1/4), so the combined
+    /// ceiling is ~1.25× this. When a cache's tracked compiled-byte total would
+    /// exceed its budget on insert, least-recently-used modules are evicted
+    /// until it fits. Bounding by bytes (not entry count) stops a node hosting
+    /// many contracts from thrashing the cache and recompiling on every access
+    /// (issue #4441). When unset, the default scales with system RAM
+    /// (`clamp(total_ram / 8, 64 MiB, 1.5 GiB)`); set this to override.
     #[arg(long, env = "FREENET_MODULE_CACHE_BUDGET_BYTES")]
     pub module_cache_budget_bytes: Option<usize>,
 
@@ -935,7 +938,7 @@ pub struct Config {
     /// ~1.25× this value. Bounds the cache by total compiled bytes rather than
     /// entry count, so a node hosting many contracts doesn't thrash (issue
     /// #4441). When unset, the default scales with system RAM
-    /// (`clamp(total_ram / 8, 64 MiB, 384 MiB)`) so a small VPS doesn't OOM and
+    /// (`clamp(total_ram / 8, 64 MiB, 1.5 GiB)`) so a small VPS doesn't OOM and
     /// a big gateway still caches a large working set.
     #[serde(
         default = "default_module_cache_budget_bytes",
@@ -992,7 +995,7 @@ fn default_max_hosting_storage() -> u64 {
 }
 
 /// Default contract-module cache byte budget, scaled to system RAM
-/// (`clamp(total_ram / 8, 64 MiB, 384 MiB)`).
+/// (`clamp(total_ram / 8, 64 MiB, 1.5 GiB)`).
 ///
 /// Resolves to [`crate::wasm_runtime::default_module_cache_budget_bytes`], the
 /// single source of truth, so the operator-facing default and the in-code
@@ -3147,9 +3150,12 @@ mod tests {
             ..Default::default()
         };
         let cfg = args.build().await.unwrap();
-        // The default is RAM-scaled and clamped to [64 MiB, 384 MiB]. It must
-        // resolve to the wasm_runtime single-source-of-truth default and land
-        // within the documented clamp range on any host.
+        // The default is RAM-scaled and clamped to [MIN, MAX]. It must resolve to
+        // the wasm_runtime single-source-of-truth default and land within the
+        // documented clamp range on any host. Reference the constants rather than
+        // hardcoded byte values so this test never drifts from the clamp.
+        let min = crate::wasm_runtime::MIN_DEFAULT_MODULE_CACHE_BUDGET_BYTES;
+        let max = crate::wasm_runtime::MAX_DEFAULT_MODULE_CACHE_BUDGET_BYTES;
         assert_eq!(
             cfg.module_cache_budget_bytes,
             crate::wasm_runtime::default_module_cache_budget_bytes(),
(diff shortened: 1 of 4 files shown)


sanity:
## Multi-model review summary (Full tier — WASM runtime is a high-risk surface)

Reviewed across several rounds as the change evolved; every finding fixed or addressed.

**External (non-Claude): Codex** — 3 passes.
- P2 cgroup limits ignored (mount-root only) → fixed: `read_total_ram_bytes` now takes `min(host MemTotal, cgroup limit)`, resolving the process's own cgroup from `/proc/self/cgroup` and **walking ancestors** (cgroup v2/v1 limits are hierarchical; a `max` leaf under a limited ancestor is still bounded).
- P2 MSRV (let-chains vs declared 1.85) → fixed: rewrote as nested `if let`.
- Final pass: **clean** — no findings.

**Claude reviewers (perspective diversity):**
- *Skeptical* (2 rounds): pin test was self-satisfying (`include_str!` of whole file) → fixed by isolating the `create_backend_engine` body + an anti-leak assertion; small-box test guard scaled with MIN → added an unconditional `combined <= total_ram` floor; added an upper-bound MAX guard test. Confirmed OOM math safe and the explicit override unclamped.
- *Big-picture*: goal-aligned, no scope creep, `Refs` (not `Closes`) correct since #4441 is a multi-part tracker; no stale docs/automation reference the old 384 MiB.

**Rule-review bot:** flagged a deleted superseded test → restored as `#[ignore]` with PR reference + replacement-test pointers (git-workflow.md); hardened the pin-test scope per its Info note. Now passing.

**Scope note:** the disk-cache deserialize-on-reload behavior (#4441's other half) was already in place via #3456 — verified, no code change, added a pin test. Did not over-scope.

CI green on all functional checks at HEAD; auto-merge enabled for the 0.2.77 release.

[AI-assisted - Claude]

sanity:
Post-merge sizing sanity-check (requested during 0.2.77 release coordination):

Measured nova's real on-disk contract set: **151 contract dirs / 173 .wasm files**, not approaching the ~1000-module ceiling. At the measured ~1.5 MiB/module that's a ~225-260 MiB resident working set — which would actually have fit under the *old* 384 MiB clamp. The thrash this PR fixes was eviction *churn* on technic (the relay carrying the ~85 phantom-interest contracts that drove the #4145 storm), where repeated evict→recompile cycles under sustained load — not a static >1000-module overflow — pinned the contract loop.

So 1.5 GiB is **generous headroom** over the real working set (the safe direction), not a tight fit:
- nova's ~150-200 modules fit in ~300 MiB; 1.5 GiB leaves ~5x room for growth + phantom-interest inflation.
- The keystone (#4473) drains the phantom-interest portion, shrinking the set further.
- It's a deliberate default sized to ~1000 modules and remains operator-overridable via `FREENET_MODULE_CACHE_BUDGET_BYTES` / `--module-cache-budget-bytes` (unclamped) — not a hard ceiling.

[AI-assisted - Claude]

Issue #4359, opened 2026-06-07
PUT of a fresh contract id gives up broadcast before subscriptions resolve → state lands locally-only, remote nodes get NotFound (and fdev reports success)


iduartgomez:
## Summary

Publishing a **brand-new contract id** can land the state **locally-hosted only** and never propagate to the network. The post-application broadcast fires within ~225 ms of the PUT, finds zero targets (no peer has subscribed to the never-before-seen id yet), retries only 3× over ~6 s, then gives up — *before* subscription/interest for the new key has had time to resolve. Result:

- The originating node holds the contract (`fdev execute get` and the gateway webapp URL return it from local cache → look healthy).
- **Every other node GETs `NotFound`** for that contract id.
- `fdev publish` does **not** surface the failure — it either reports success or hits the 300 s client timeout (#4102), while the propagation summary clearly shows `targets_avg=0.0 no_targets=1`.

A second identical PUT of the **same** id (same node, same ~50 connections) ~3.5 h later propagated fine — the broadcast fired 31 s after the PUT and streamed to real peers. So this is a **timing race on cold ids**, not a connectivity problem.

## Environment

- Freenet `0.2.70` (`717c8a55b7aa`), fdev `0.3.232`
- Node: long-running `freenet network`, 49–51 active connections throughout
- Contract: web-container webapp, id `H6pb4UfCUY1SAXPDAUCxkQAckahpS5Yo36soENnVuWac` (21.2 MB state), brand-new id for a fresh release
- Published via `fdev publish --code … --parameters … contract --webapp-archive … --webapp-metadata …`

## Evidence — first PUT (failed propagation)

```
09:21:20.421  PUT relay: processing Request tx=01KTGP6014… contract=H6pb4UfCUY… htl=10 phase="relay_put_request"
09:21:20.645  BROADCAST_NO_TARGETS: scheduling retry - subscriptions may still be in-flight … attempt=1 max_retries=3
09:21:20.649  NEIGHBOR_HOSTING: Added contract to locally hosted contract=H6pb4UfCUY…
09:21:21.647  BROADCAST_NO_TARGETS: … attempt=2 max_retries=3
09:21:23.648  BROADCAST_NO_TARGETS: … attempt=3 max_retries=3
09:21:26.649  WARN BROADCAST_NO_TARGETS: no targets found after 3 retries, giving up contract=H6pb4UfCUY…
[...]

PR #4489, opened 2026-06-18
fix(contract): gate auto-fetch + overlap-sync on legitimate interest

sanity:
## Problem

After the #4404 placement migration, several production peers (notably technic and the nova gateway) carry **phantom contract interest**: stale interest in hundreds of contracts they neither host nor have any client/downstream subscriber for. Two code paths turn that phantom interest into a sustained `fetch_contract` storm on the single-threaded contract-handling loop:

1. **NeighborHosting overlap-sync (`node.rs`)** — on *every* inbound `NeighborHosting` announce, the loop called `get_contract_state_by_id` (which opens a `fetch_contract` span on the serial contract-handling loop) for *every* overlapping contract, only to throw the result away at the existing `is_receiving_updates() || has_downstream_subscribers()` discard gate. Fetch-then-discard, once per overlapping phantom contract per announce.

2. **UPDATE auto-fetch (`update.rs::try_auto_fetch_contract`)** — an inbound UPDATE/broadcast for a phantom-interest contract spawned a sub-op GET (`fetch_contract`) every time the per-contract 5-minute cooldown lapsed. The cooldown only *throttles* the storm; it does not stop it.

This is the residual #4473 `fetch_contract` churn — the fetch-path sibling of the already-merged summarize gates #4475 / #4482. On technic the per-callsite rate limiter is currently dropping 700-1000+ `fetch_contract` calls per 30s in bursts (cumulative counter climbing ~1500/4.5min).

## Approach

Gate both paths on *legitimate interest*, mirroring the #4475 / #4482 `is_hosting_contract || contract_in_use` summarize gates:

1. **`node.rs` overlap-sync:** before fetching, check `is_receiving_updates(probe) || has_downstream_subscribers(probe) || pending_broadcasts.contains(instance_id)` and `continue` if none hold. The gate **reuses the existing discard predicate**, so served contracts are unaffected; the `pending_broadcasts` clause preserves the #4359 fresh-PUT flush. The predicates take a synthetic `ContractKey` with a zero code hash — safe because `ContractKey` equality and hashing are *instance-only* in freenet-stdlib (`key.rs`), so the hosting/subscription maps resolve from the instance id alone (and `get_contract_state_by_id` is the very round-trip we are avoiding).

2. **`update.rs::try_auto_fetch_contract`:** early-return `if !self.ring.contract_in_use(key)` — no local client and no downstream subscriber means nothing on this node depends on the contract, so don't self-heal-fetch it. Deliberately *not* gated on `is_hosting_contract` (the merge already failed because we lack the contract) nor on a bare upstream network subscription (which `contract_in_use` excludes by design — an orphaned upstream sub is meant to be torn down, not kept alive by self-heal fetches).

No subscribe-path cooldown is added: subscribe-driven fetch is legitimate demand and intentionally left alone.

## Testing

- **Behavioural regression** (`hosting.rs::auto_fetch_gate_skips_phantom_contracts_and_passes_served_ones`): drives the `contract_in_use` predicate the gate keys on through real subscription registration/teardown and asserts the decision at each step — phantom skips, live local client passes, live downstream subscriber passes, upstream-network-sub-only skips, and teardown re-closes the gate.
- **Source-scrape pins** (one per gated path): assert the gate precedes the fetch/spawn at the source level so a future refactor that reorders them fails CI rather than silently regressing the churn.
- Build clean (`cargo build -p freenet`), `cargo clippy --locked -- -D warnings` clean, changed-module test suites green (`ring::hosting`, `operations::update`: 163 passed).
- Deploy-validated on technic (real NATed peer): `fetch_contract` rate-limiter drop bursts vs. the live 0.2.77 baseline.

This is the fetch-path sibling of #4475 / #4482. Recovered from a borg backup after the original worktree was deleted before commit.

Refs #4473 #4440

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1621,6 +1621,44 @@ where
                     );
                     continue;
                 }
+                // Skip the per-contract state fetch — a `GetQuery` that opens
+                // the `fetch_contract` span on the single-threaded
+                // contract-handling loop — for contracts we neither actively
+                // serve nor owe a deferred broadcast. A node carrying phantom
+                // interest (e.g. the #4404 placement migration left hundreds
+                // of not-held contracts) otherwise fetched state for EVERY
+                // overlapping contract on EVERY inbound NeighborHosting
+                // announce, only to discard it at the
+                // `is_receiving_updates() || has_downstream_subscribers()`
+                // gate below. That fetch-then-discard was the residual #4473
+                // `fetch_contract` churn on technic (the fetch-path sibling of
+                // the #4475 / #4482 summarize gates).
+                //
+                // The gate reuses the existing discard predicate
+                // (`is_receiving_updates || has_downstream_subscribers`), so it
+                // changes nothing for served contracts, and adds a
+                // `pending_broadcasts` clause so the #4359 fresh-PUT flush at
+                // the matching arm still runs for any contract that owes one.
+                // Skipping is safe for the flush because a deferred broadcast
+                // is only ever stashed for a contract THIS node originated
+                // (broadcast give-up), so the flush is a guaranteed no-op for
+                // every contract this gate skips. The predicates take a
+                // synthetic key with a zero code hash: `ContractKey` equality
+                // and hashing are instance-only (freenet-stdlib `key.rs`), so
+                // the hosting / subscription maps resolve correctly from the
+                // instance id alone — `get_contract_state_by_id` is the only
+                // path that recovers the full key here, and that is exactly the
+                // round-trip we are avoiding.
+                let probe_key = freenet_stdlib::prelude::ContractKey::from_id_and_code(
+                    instance_id,
+                    freenet_stdlib::prelude::CodeHash::new([0u8; 32]),
+                );
+                if !op_manager.ring.is_receiving_updates(&probe_key)
+                    && !op_manager.ring.has_downstream_subscribers(&probe_key)
+                    && !op_manager.pending_broadcasts.contains(&instance_id)
+                {
+                    continue;
+                }
                 if let Some((key, state)) =
                     get_contract_state_by_id(&op_manager, &instance_id).await
                 {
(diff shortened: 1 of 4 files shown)


sanity:
## Codex review (round 1) — finding addressed

Ran `codex review --base main`. One finding, a valid P2:

> **[P2] Allow originator auto-fetch without a subscription** (`update.rs`) — when a local client submits an UPDATE for a contract whose code/params are missing, the originator path (`phase = "auto_fetch_originator"`) calls `try_auto_fetch_contract` and tells the client to retry; but a one-shot updater is not necessarily subscribed, so the new `contract_in_use` gate returns before spawning the GET and the retry can never self-heal.

This was a real regression. The gate was placed inside `try_auto_fetch_contract` itself, so it caught all three callers — including the originator self-heal site, which is demand-driven (a client is waiting on the retry) and should never be gated, exactly like subscribe-driven fetch.

**Fix** (commit `2220701f`): added an `AutoFetchReason { Originator, InboundRelay }` parameter so the gate is opt-in.
- The originator site (`drive_client_update`, keyed on `target_addr`) passes `Originator` → bypasses the gate, self-heal restored.
- The two inbound relay/broadcast sites (keyed on `sender_addr`) pass `InboundRelay` → keep the gate. These are exactly the #4473 phantom-churn paths, so the fix's intent is preserved.

Pins strengthened: the source-scrape pin now asserts the gate is conditioned on `InboundRelay` (so it can't silently become unconditional again), plus a new `auto_fetch_call_sites_use_correct_reason` test asserts the originator site uses `Originator` and no relay site does.

Build + `cargo clippy --locked -- -D warnings` clean; changed-module suites green (164 passed). Re-running Codex on the new content.

[AI-assisted - Claude]

sanity:
## Deploy-validate on technic (real NATed peer) — PASS

Built the release binary at content commit `2220701f` (the AutoFetchReason fix; the later `a2537d52` is whitespace-only, behaviourally identical), backed up technic's 0.2.77 binary, swapped, and restarted on the existing config. Watched ~11 min.

**(a) `fetch_contract` churn dropped ~85%.** The per-callsite rate limiter's cumulative *dropped* counter is the cleanest signal (cap is 30/sec; everything above is phantom churn it sheds):

| | dropped/min | 30s bursts |
|---|---|---|
| 0.2.77 baseline (pre-swap) | ~220/min (climbed 7,275 in 33 min) | sustained 130-627, spikes to 700-1063 |
| 0.2.78 candidate (~11 min) | ~34/min (cumulative 393) | mostly 4-48, two interest-rebuild blips (79, 113) that self-correct |

The sustained phantom-fetch storm is gone; what remains is brief interest-rebuild settling that drops straight back to single digits.

**(b) River loads:** `GET /v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/` → **HTTP 200, 45,447 bytes, 0.003s** (identical size to pre-swap).

**(c) Nothing legit broke:** no panics / FATAL / crashes in the candidate window; service active throughout; mem ~1.3 GB, CPU ~10%; peer connections forming normally. Originator UPDATE self-heal preserved by the `AutoFetchReason::Originator` bypass (the Codex round-1 fix).

technic is left running the 0.2.78 candidate binary (its 0.2.77 binary is backed up at `~/.local/bin/freenet.0.2.77-backup-*`).

[AI-assisted - Claude]

Issue #4492, opened 2026-06-18
release-agent: dead gateway reported as successful update (v0.2.78 nova incident)


sanity:
## Problem

During the 2026-06-18 v0.2.78 release, nova's gateway binary was swapped to 0.2.78 but its systemd service was left **dead** (a separate racing updater stopped it — fixed separately). The whole update chain still reported **success**, nova was down ~4.5 min, and nothing flagged it.

The release-agent journal on nova (`freenet-release-agent`, Jun 18 ~10:31 CT) shows the false success directly:

```
Starting systemd service (freenet-gateway)... ✓
Verifying service status (freenet-gateway)... ✗
⚠️  Service failed to start
✅ Deployment complete!                                  <-- script still "succeeds"
[INFO] Deployment successful
[INFO] Successfully updated to v0.2.78
update command exited status=exit status: 0              <-- release-agent: success
```

And the `gateway-update.yml` run for v0.2.78 (`27770490034`) reported success via:

```
✓ Gateway nova now reports version 0.2.78 (binary-only check)   <-- success-fallback
```

## Root cause (three converging stale-deployment links)

1. **Stale `deploy-local-gateway.sh` on nova** (mtime 2025-12-21) predates commit `9540abf8` (Dec 31), so it still has `verify_service "$service" "$BINARY_VERSION" || true` (exit code swallowed) and prints `✅ Deployment complete!` + `exit 0` unconditionally. The repo's current version already fixes this (`VERIFICATION_FAILED` → `exit 1`), but the deployed script on the gateway was never refreshed.

2. **Stale release-agent binary on nova** predates #4378, so `/version` omits the `service_active` field.

3. **Workflow `success-fallback`**: when `/version` omits `service_active`, `verify-version-decision.sh` returns `success-fallback`, and `gateway-update.yml` treats a binary-version match alone as success — exactly the dead-service-but-binary-swapped case #4378 set out to catch.

Even after redeploying current scripts to nova, links 2/3 mean any gateway running an old agent silently bypasses the service-health gate.

## Approach

Make the failure non-silent at layers that don't depend on every gateway having a freshly-deployed script/binary:

1. `deploy-local-gateway.sh`: `verify_service` must return non-zero whenever it cannot positively confirm the service is `active` on the right binary (close the unit-file-not-found and non-systemd return-0 fall-throughs).
[...]

PR #4494, opened 2026-06-18
fix(release-agent): fail gateway update when the service is not running

sanity:
## Problem

A gateway update that leaves the systemd service **not running** can still be reported as a **successful update**, so the release workflow and the release-agent's `/version` show green while the gateway is actually down.

This bit us on 2026-06-18 (v0.2.78): nova's binary was swapped to 0.2.78 but its service was left dead (a separate racing updater stopped it — fixed separately). The release-agent journal on nova shows the false success directly:

```
Verifying service status (freenet-gateway)... ✗
⚠️  Service failed to start
✅ Deployment complete!                              <-- script still exits 0
[INFO] Deployment successful
[INFO] Successfully updated to v0.2.78
update command exited status=exit status: 0          <-- release-agent: success
```

and the `gateway-update.yml` run for v0.2.78 (`27770490034`) reported success via:

```
✓ Gateway nova now reports version 0.2.78 (binary-only check)   <-- success-fallback
```

nova was down ~4.5 min and nothing flagged it.

### Root cause: three converging stale-deployment links

Tracing the chain (`gateway-update.yml` → release-agent `updater.rs` → `gateway-auto-update.sh` → `deploy-local-gateway.sh`), three independent links each let a dead service look healthy:

1. **Stale `deploy-local-gateway.sh` on nova** (mtime 2025-12-21) predated commit `9540abf8`, so it still had `verify_service "$service" "$BINARY_VERSION" || true` (exit code swallowed) and printed `✅ Deployment complete!` + `exit 0` unconditionally. The repo copy was already fixed — but the script also still had two **return-0 fall-throughs** in the current code: unit-file-not-found, and a latent `set -u` "unbound variable" crash on the success path when no freenet PID is found.

2. **Stale release-agent binary on nova** predates #4378, so `/version` omits `service_active`.

3. **Workflow `success-fallback`**: when `service_active` is absent, the workflow treated a binary-version match alone as success — exactly the dead-service-but-binary-swapped case #4378 set out to catch.

Even after redeploying current scripts to nova, links 2/3 mean any gateway running an old agent silently bypasses the service-health gate.

## Approach

Make the failure non-silent at every layer, so it is caught regardless of which deploy script / agent version is installed on a given gateway:

1. **`deploy-local-gateway.sh`** — `verify_service` now returns non-zero on every path that does not positively confirm the service is `active` on the correct binary (unit-file-not-found is a failure; the success path returns 0 explicitly). Also fixes the pre-existing `set -u` crash by declaring `running_binary` up front.

2. **`gateway-auto-update.sh`** — add an **independent** post-deploy `systemctl is-active` gate (`verify_service_active`) that does **not** trust the deploy script's exit code for liveness, so even a stale/buggy deploy script that exits 0 on a dead service fails the update. Also makes the on-disk version-mismatch check fatal instead of a warning.

3. **`gateway-update.yml`** — new `allow_binary_only_fallback` input (default **off**). When the agent omits `service_active`, the workflow no longer silently passes on binary version alone: by default it treats the response as not-yet-converged and **fails closed** at the deadline with an actionable error. The old binary-only behaviour is available via the explicit opt-in.

The release-agent's `/version` already reports `service_active` (PR #4378) and the workflow already polls it; this PR closes the remaining links that let the gate be bypassed.

### Operational note (please read before the next release)

Until nova/vega release-agents are upgraded past the #4378 build, the next release's `gateway-update.yml` will **fail closed** for them (or require `allow_binary_only_fallback=true`), because their `/version` still omits `service_active`. That is the **intended, loud** behaviour — it surfaces the same dead-gateway risk this PR closes rather than hiding it. This PR does **not** touch nova/vega; redeploying the updated scripts and upgrading the agents is an ops follow-up.

## Testing

- New `scripts/release-agent/gateway-auto-update_test.sh` (wired into the CI Fmt job, alongside the existing `verify-version-decision_test.sh`): sources the wrapper and proves a **lying deploy script** (prints "Deployment complete!" and exits 0 while the service is dead — the exact incident) is still caught by the independent `is-active` gate, and that it passes when the service is active. Verified the test **fails without** the gate (reproduces the bug) and **passes with** it.
- `deploy-local-gateway.sh` active / dead / unit-not-found paths verified by hand against a stub `systemctl`: active → exit 0, dead → exit 1, unit-not-found → exit 1.
- Existing `verify-version-decision_test.sh` still passes unchanged (the decision script is an unchanged pure classifier; the workflow owns the policy).

## Risk tier

Release/deploy infra = high-risk surface → Full multi-model review.

Refs #4492

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,15 @@ jobs:
       - name: Self-test gateway-update verify decision
         run: bash scripts/release-agent/verify-version-decision_test.sh
 
+      # Regression gate for the independent post-deploy service-health check in
+      # gateway-auto-update.sh (the 2026-06-18 v0.2.78 nova incident, #4492: a
+      # stale deploy-local-gateway.sh exited 0 on a DEAD service and the wrapper
+      # reported the update successful). The load-bearing case: a deploy script
+      # that lies (exits 0 while the service is down) must still fail the update
+      # via the independent `systemctl is-active` gate.
+      - name: Self-test gateway-auto-update service-health gate
+        run: bash scripts/release-agent/gateway-auto-update_test.sh
+
       - name: Check for old-style mod.rs files
         run: |
           # New modules must use foo.rs + foo/ style, not foo/mod.rs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,13 @@ jobs:
       - name: Self-test gateway-auto-update service-health gate
         run: bash scripts/release-agent/gateway-auto-update_test.sh
 
+      # Regression gate for deploy-local-gateway.sh::verify_service — the
+      # function that actually regressed on nova (#4492). Pins that a dead
+      # service, a missing unit, and an active-but-no-process unit each cause a
+      # non-zero exit, and that a healthy deploy exits 0.
+      - name: Self-test deploy-local-gateway verify_service
+        run: bash scripts/release-agent/deploy-local-gateway_test.sh
+
       - name: Check for old-style mod.rs files
         run: |
           # New modules must use foo.rs + foo/ style, not foo/mod.rs
--- a/.github/workflows/gateway-update.yml
+++ b/.github/workflows/gateway-update.yml
@@ -284,7 +284,9 @@ jobs:
                 # like `wait` (keep polling, fail at the deadline) and tell the
                 # operator to update the agent. Only the explicit
                 # allow_binary_only_fallback opt-in restores the old behaviour.
-                if [[ "${ALLOW_BINARY_ONLY_FALLBACK:-false}" == "true" ]]; then
+                # The accept/hold policy lives in fallback_decision() (sourced
+                # above) so it is unit-tested rather than inline-untested.
+                if [[ "$(fallback_decision "$DECISION" "${ALLOW_BINARY_ONLY_FALLBACK:-false}")" == "accept" ]]; then
                   echo "::warning::Gateway ${{ matrix.gateway.name }} agent predates the service-health check (no 'service_active' in /version); allow_binary_only_fallback=true, so accepting binary-only verification. Update the release-agent on this gateway to enable the stronger check."
                   echo "✓ Gateway ${{ matrix.gateway.name }} now reports version $INSTALLED (binary-only check)"
                   exit 0
--- a/.github/workflows/gateway-update.yml
+++ b/.github/workflows/gateway-update.yml
@@ -29,6 +29,18 @@ on:
         required: false
         type: boolean
         default: false
+      allow_binary_only_fallback:
+        description: >-
+          Accept a gateway whose release-agent predates the service_active
+          field (binary-version-only verification). OFF by default: without
+          service_active we cannot tell a running gateway from one whose binary
+          was swapped but whose service is dead (the 2026-06-18 nova incident,
+          #4492), so a missing field is treated as not-yet-converged and the
+          rollout fails at the deadline. Turn this on ONLY for a deliberate
+          rollout to a gateway you know runs an old agent.
+        required: false
+        type: boolean
+        default: false
 
 permissions:
   contents: read
@@ -223,6 +235,9 @@ jobs:
         env:
           GATEWAY_URL: ${{ matrix.gateway.url }}
           VERSION: ${{ needs.plan.outputs.version }}
+          # Empty on the release-trigger path (no inputs) → defaults to the
+          # safe "false". See the allow_binary_only_fallback input docs.
+          ALLOW_BINARY_ONLY_FALLBACK: ${{ inputs.allow_binary_only_fallback }}
         run: |
           set -euo pipefail
           # The agent restarts the freenet service after the script
@@ -245,6 +260,9 @@ jobs:
           source scripts/release-agent/verify-version-decision.sh
 
           DEADLINE=$(( $(date +%s) + 120 ))
+          # Remember if we ever saw the old-agent (no service_active) response,
+          # so the deadline error can point at the real cause.
+          SAW_FALLBACK=false
           while (( $(date +%s) < DEADLINE )); do
             RESPONSE=$(curl -sS --max-time 5 "$GATEWAY_URL/version" || true)
             INSTALLED=$(printf '%s' "$RESPONSE" | jq -r '.version // empty' 2>/dev/null || echo "")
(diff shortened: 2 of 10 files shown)


sanity:
Added the `test-exempt` label.

This is a `fix:` PR, but the bug lives entirely in shell scripts and a GitHub workflow (`deploy-local-gateway.sh`, `gateway-auto-update.sh`, `gateway-update.yml`), not in Rust — so the Rule Lint "fix: PRs must add a Rust `#[test]`" check (which only counts `#[test]`/`#[tokio::test]`/`#[freenet_test]` attributes) cannot be satisfied with a Rust test, and a Rust test would be the wrong level of coverage.

The regression coverage IS present, at the right level, and is wired into CI (the Fmt job in `ci.yml`):

- `scripts/release-agent/deploy-local-gateway_test.sh` (new) — pins `verify_service`: dead service, missing unit, active-but-no-process, and a cross-unit-cgroup process each exit non-zero; a healthy deploy (and a wrapper-ExecStart gateway in this unit's cgroup) exits 0. Verified it FAILS against the old "swallow the verify failure" deploy script — i.e. it reproduces the exact function that regressed on nova on 2026-06-18.
- `scripts/release-agent/gateway-auto-update_test.sh` (new) — pins the independent post-deploy `systemctl is-active` gate: a deploy script that lies (prints "Deployment complete!" and exits 0 while the service is dead) is still caught. Verified it fails without the gate.

So the bug-that-passed-CI gap is closed in this PR, just not via a Rust test. [AI-assisted - Claude]

sanity:
## Multi-model review (Full tier — release/deploy infra)

Ran the four Freenet reviewer perspectives (skeptical, code-first, testing, big-picture) + an external Codex pass, across two review rounds as the code changed. Every finding was fixed or has a justification below.

### Round 1 findings → all addressed
- **MEDIUM (Codex + code-first): `SERVICE_NAME` not forwarded to the deploy script.** `gateway-auto-update.sh` now passes `--service "$SERVICE_NAME"` to `deploy-local-gateway.sh` so the deploy target and the post-deploy `is-active` gate always check the same unit. The release-agent does not yet forward its `managed_service` config (sudo strips env; the robust fix is an argv flag through the sudoers wildcard) — documented honestly as a #4492 follow-up. nova, the incident host, uses the default `freenet-gateway`, so the default path is covered.
- **MEDIUM (Codex): `verify_service` returned 0 when the unit was `active` but no live process was found.** Now fails closed (active-but-exited / oneshot misconfig no longer blesses the deploy), with a bounded restart-retry.
- **MEDIUM (testing): the originally-regressed `deploy-local-gateway.sh::verify_service` had no automated test.** Added `deploy-local-gateway_test.sh` (CI-wired), which fails against the old swallow-the-failure script.
- **MEDIUM (big-picture): stale docs.** Updated `docs/RELEASING.md` and `scripts/release-agent/README.md` to describe the new fail-closed default + `allow_binary_only_fallback` opt-in.
- **LOW (Codex): arg parsing ran at source time.** Moved into `parse_args()`, called only from the executed-only guard.

### Round 2 findings → all addressed
- **MEDIUM (Codex): global `pgrep` could bless a sibling unit's process on a multi-instance host.** `verify_service` now prefers the unit's own `MainPID` and, on the pgrep fallback, attributes the candidate PID to this unit via its cgroup (`grep -qF "<unit>.service"`), failing closed otherwise. New tests pin both the cross-unit-fail and the wrapper-ExecStart-success cases.
- **LOW (Codex): tests omitted `set -e` on fixtures.** Both shell tests now hard-fail on `mktemp`/`mkdir` errors so a broken harness can't pass a negative case vacuously.
- **HIGH (Codex): `candidate_pid=$(pgrep | head -1)` `set -e` abort risk.** Guarded with `|| true` (defensive — `verify_service` runs in an `if !` tested context today where set -e is already suppressed, but the guard makes it correct for any future caller). Added a test asserting the active-but-no-process path reaches its graceful branch, not a set -e abort.

### Big-picture verdict
The fix achieves the goal end-to-end via defense-in-depth across the three layers that were each stale on nova (deploy script, agent binary, workflow policy). The release-agent Rust side (`service_active` in `/version`, #4378) was already sufficient and is unchanged. Scope is bug-fix (issue #4492). No safety weakened.

### External-model availability note
Codex ran successfully for the round-1 and round-2 diffs and a final tightly-scoped pass on the cgroup logic. A final broad Codex pass timed out on verbose sandbox logging and Gemini returned HTTP 403 (license/quota) — but the substantive external review (two full Codex passes that drove the round-1/round-2 fixes) did run, and the only delta past them is the small `|| true` / `grep -F` hardening, which a focused Codex pass did review.

### Operational note (repeated from the PR body — please sequence)
Until nova/vega release-agents are upgraded past the #4378 build, the next release's `gateway-update.yml` will **fail closed** for them (they omit `service_active`), or require `allow_binary_only_fallback=true`. That is intended/loud. This PR does not touch nova/vega; upgrading their agents + redeploying the updated scripts is the ops follow-up.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4494 (multi-model, Full tier)

### Summary
- **Type:** fix (release/deploy infra — high-risk surface → Full review)
- **Linked issue:** #4492
- **Reviewers run:** code-first, testing, skeptical, big-picture (4 perspectives) + external Codex `codex review --base origin/main`. Earlier rounds also ran two additional Codex passes; this is the consolidated review on the current content.
- **Review HEAD:** `ee90e7952ea3d0f243c6127e58f911ce24c54d27` (after rebase onto origin/main `bfc1e55b`; fast-forward, no conflicts)

### Code-First
Implementation matches stated intent. All strong code claims verified against control flow ("ONLY success path"; verify_service_active fails on non-active; "mirrors version.rs::query_service_active" — confirmed semantics match: active-only, non-active/missing→false/fail-closed). Workflow fail-closed default matches the YAML+bash.

### Testing
- Regression tests are genuinely load-bearing (mutation-verified): `gateway-auto-update_test.sh` reproduces the exact #4492 "lying deploy script + dead service" and fails without the gate; `deploy-local-gateway_test.sh` fails against the old swallow-the-failure script.
- **Finding (fixed): masking test + broken pgrep pattern.** The stub `pgrep` ignored its args, masking that the production pattern `freenet.*--is-gateway\\|freenet.*network` used a literal `\|` (pgrep ERE), matching nothing. Fixed the pattern to real alternation; made the stub faithful (applies the ERE to a realistic cmdline) so it can't mask again — verified the test now FAILS against the old pattern.
- **Finding (fixed): workflow fail-closed policy was inline YAML, untested.** Extracted to `fallback_decision()` in verify-version-decision.sh with 5 unit tests; workflow now calls it.

### Skeptical
Traced every verify_service return path: the only `return 0` requires active + a live freenet process attributed to this unit + correct binary + version match. set -e/-u/pipefail interactions verified (verify_service runs in an `if !` context; `|| true` guards added defensively). Recursion bounded (max_retries=2). `grep -qF` prevents freenet-gateway/freenet-gateway-hector cross-match. No blocker.

### Skeptical / Codex (external)
- **Finding (fixed): `head -1` only checked the first pgrep match.** On a multi-instance host the first match can be a sibling → false failure. Now iterates all matches, cgroup-attributing each; added an iterate-past-sibling test that fails against the old logic.

### Big-Picture
- Goal achieved end-to-end via defense-in-depth across the 3 layers that were stale on nova. **Rust release-agent correctly left unchanged** (confirmed: /update returns 202 "accepted" not "succeeded"; updater.rs only LOGS the spawned script's exit; success is the workflow's /version poll, which already reports service_active per #4378). No missing Rust link.
- Scope: within bug-fix (restores the intended "dead service = failed update" behavior #4378 set out to achieve). allow_binary_only_fallback is a safety knob, not a feature.
- **Finding (fixed): stale `success-fallback` comments** in verify-version-decision.sh + its test described auto binary-only fallback; updated to note the consumer fails closed by default since #4492.
- Fail-closed-for-old-agents consequence is intended/loud and documented (PR body, workflow input desc, docs/RELEASING.md). nova/vega need their agents upgraded before the next release's gateway-update goes fully green for them — ops follow-up, not in this PR.

### Documented follow-ups (not blocking; tracked under #4492)
- Thread the release-agent's `managed_service` config to the script as an argv flag so vega's `freenet-gateway-hector` is verified against the right unit (sudo strips env; nova uses the default unit so the incident path is covered).
- The bash-wrapper sub-fallback (`pgrep -f "$INSTALL_PATH"`) and the cgroup-unreadable branch are pre-existing/minor and only on the wrapper path.

### Verdict
**State:** Ready to Merge — all findings fixed or justified; new regression tests pin each fix (verified to fail against the buggy versions). Re-review after the fixes: this report IS the post-fix pass on HEAD `ee90e7952ea3d0f243c6127e58f911ce24c54d27`.

[AI-assisted - Claude]

PR #4498, opened 2026-06-19
feat(telemetry): add broadcast stream-assembly + bg-task health gauges (#4440)

sanity:
## What

Adds two **low-frequency observability gauges** to the existing `router_snapshot` telemetry event (~5 min/node cadence). No new event kinds, no per-op/per-message emission — this piggybacks the snapshot, so it adds negligible load to the telemetry collectors.

Part of #4440 (the deferred-work tracker from the v0.2.73 placement-migration incident).

### Gauge 1 — UPDATE-broadcast stream-assembly failure rate
The **single signal that flagged the v0.2.73 incident** (nova/vega ~1500–2300 stream-assembly failures/hr vs ~0 baseline) and the one that was missing — the incident was nearly invisible to central telemetry. Process-global atomic counters (attempts + failures) classified at `broadcast_to_single_peer` (the one point that observes all UPDATE-broadcast failure modes uniformly: explicit drop, dropped completion oneshot, completion timeout); GET/PUT/SUBSCRIBE never reach it. Emitted as a monotonic `*_total` plus a per-snapshot delta.

### Gauge 2 — background-task health (`refresh_router`)
`last_success_age_secs` + `consecutive_failures` for the `refresh_router` background task. The v0.2.74 hotfix (#4438) made transient `refresh_router` failures non-fatal — correct, but it meant a *persistent* failure now degrades routing silently forever. This makes that observable (partially addresses #4440 item (a)).

## Why now

A release is planned imminently. Shipping these gauges **with the placement migration still disabled** (the `SUBSCRIBE_HINT_MIN_VERSION` floor is untouched — verified) gives us a clean fleet-wide **baseline** of the incident metric *before* any re-enable. When the migration is eventually re-enabled (a later release, gated on #4145 + the #4233 fan-out validation test), this telemetry will already be deployed to watch it.

## Risk

Observability-only, additive. Does **not** touch the migration gate, routing logic, or broadcast behavior — only reads/counts. Hot-path cost is two relaxed atomic increments per broadcast attempt.

- Follows the established #4470/#4472 gauge pattern, including the hand-mirrored `event_kind_to_json` block (the #4009/#4010 footgun) with a pin test that genuinely guards it (verified by removing a mirror line → test fails).
- `cargo build` / `fmt` / `clippy -p freenet -- -D warnings` (CI gate) all clean. Telemetry pin tests + refresh_router + broadcast suite pass.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -51,6 +51,10 @@ use serde::{Deserialize, Serialize};
 pub(crate) use network_bridge::{
     ConnectionError, EventLoopNotificationsSender, NetworkBridge, OpExecutionPayload, WaiterReply,
 };
+// Re-export the UPDATE-broadcast stream-assembly telemetry global (#4440) so the
+// `Ring` snapshot task can read it (the broadcast queue lives behind the private
+// `network_bridge` module). Mirrors `crate::wasm_runtime::MODULE_CACHE_METRICS`.
+pub(crate) use network_bridge::broadcast_queue::BROADCAST_STREAM_METRICS;
 #[cfg(test)]
 pub(crate) use network_bridge::{EventLoopNotificationsReceiver, event_loop_notification_channel};
 // Re-export types for dev_tool and testing
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -32,13 +32,18 @@ const STREAM_COMPLETION_TIMEOUT: Duration = Duration::from_secs(120);
 /// Process-global UPDATE-broadcast stream-assembly telemetry (#4440).
 ///
 /// The streaming broadcast path (`broadcast_to_single_peer`'s `use_streaming`
-/// branch) sends a multi-fragment state transfer to one subscriber peer and
-/// awaits a `BroadcastDeliveryOutcome` completion signal. A non-`Delivered`
-/// outcome (explicit `Dropped`, a dropped completion oneshot, or a
-/// `STREAM_COMPLETION_TIMEOUT`) is a stream-assembly / transfer failure — the
-/// exact signal that flagged the v0.2.73 incident, where nova/vega saw
-/// ~1500-2300 broadcast stream-assembly failures/hr against a ~0 baseline and
-/// central telemetry had no gauge for it.
+/// branch) sends a multi-fragment state transfer to one subscriber peer. Each
+/// invocation records exactly one attempt, and a failure on any of its three
+/// exits is counted: the initial metadata send returning `Err` (the stream
+/// never landed), `send_stream_with_completion` returning `Err` (dispatch
+/// failed before any fragment), or a post-dispatch non-`Delivered`
+/// `BroadcastDeliveryOutcome` (explicit `Dropped`, a dropped completion oneshot,
+/// or a `STREAM_COMPLETION_TIMEOUT`). All three are stream-assembly / transfer
+/// failures — the exact signal that flagged the v0.2.73 incident, where
+/// nova/vega saw ~1500-2300 broadcast stream-assembly failures/hr against a ~0
+/// baseline and central telemetry had no gauge for it. (The two early-send
+/// exits are the congestion failure mode that would otherwise bias the gauge
+/// LOW precisely when it matters most.)
 ///
 /// These broadcast tasks are spawned per (contract, peer) from the global
 /// `BroadcastQueue` worker, unreachable from the `Ring` telemetry-snapshot task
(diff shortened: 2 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4498

### Summary
- **PR Title:** feat(telemetry): add broadcast stream-assembly + bg-task health gauges (#4440)
- **Type:** feat (observability)
- **CI Status:** structural jobs green (Clippy/Fmt/Rule Review/Rule Lint/NAT/E2E); Simulation + Unit & Integration + Windows running at review time
- **Linked Issues:** #4440 (re-enable tracker); completes its observability gauges (3) + partially (a)
- **Review tier:** Full (touches the UPDATE-broadcast fan-out path = concurrency/async high-risk surface; ships in an imminent release)
- **Reviewers run:** external Codex + four Claude lenses (code-first, testing, skeptical, big-picture)

The change adds two **low-frequency** gauges to the existing `router_snapshot` event (~5 min/node): a UPDATE-broadcast stream-assembly failure-rate counter and `refresh_router` background-task health. It is purely additive observability — **the migration version floor (`SUBSCRIBE_HINT_MIN_VERSION`) and all routing/broadcast behavior are untouched** (verified by code-first + big-picture). This intentionally ships the incident-detection telemetry **with the migration still disabled**, to baseline the metric before any future re-enable.

### What passed (cleared by review)
- **3-way telemetry consistency** (struct ⟷ populate ⟷ JSON mirror) is **exact** for all 5 fields; the pin test guards each with a distinct sentinel (the #4009/#4010 hand-mirror footgun is covered).
- **Atomic ordering** — `Relaxed` is correct; no cross-counter invariant; torn reads are harmless (only a `saturating_sub` delta of one counter is derived).
- **Hot-path safety** — `record_attempt` is two `fetch_add` after the completion await; no lock/alloc/syscall/await added to the surface that wedged v0.2.73.
- **Time handling under virtual time** — uses `tokio::time::Instant`; the cross-runtime epoch-pollution candidate was chased and dismissed (the epoch cancels in the age subtraction). No panic/overflow paths.
- **Low-volume constraint** — piggybacks the existing 5-min snapshot; no new event kind / per-op / per-fd emission.

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
1. **Broadcast failure counter undercounts two early-failure paths** — `broadcast_queue.rs` `broadcast_to_single_peer`, streaming branch. `record_attempt` only runs after `send_stream_with_completion` succeeds. The initial `bridge.send` failing (`if send_res.is_ok()` false) and `send_stream_with_completion` returning `Err` are counted as neither attempt nor failure. This is the **congestion** failure mode, so the gauge is biased **low precisely during an incident** — defeating part of the signal's purpose. (Found independently by Codex, code-first, and skeptical.) Fix: record exactly one `record_attempt(false)` on each early-failure path; correct the "single classification point / all failure modes uniformly" comment to match actual coverage.
2. **Delta-sampling is inline + untested** — `ring.rs` `emit_router_snapshot_telemetry` computes `total.saturating_sub(prev); prev = total` inline. A future reorder of the `prev =` assignment would silently zero the incident signal with no test failure. Fix: extract a pure `window_delta(current, &mut prev)` helper + unit test. (testing)

#### Consider
1. Add a one-line rustdoc caveat that the process-global gauges' per-node meaning only holds in single-node-per-process production; a multi-node simulation reads the aggregate (matches the existing `MODULE_CACHE_METRICS` comment). No live bug. (skeptical, testing)
2. When this merges, tick the two completed gauge boxes on #4440 and note it lands observability item (4) + partially item (a). (big-picture, tracking hygiene)

### Verdict
**HEAD SHA reviewed:** c5e67a13a66d691bc8de6e8e5c5b06e4b07f5da3

Two Should-Fix items, both about gauge accuracy/robustness (no behavior change, no Must-Fix). Fixes are being applied; a focused re-check of the delta (re-run external model on the changed lines + verify the early-failure accounting fires exactly once per path) will run on the updated HEAD before merge. CI must be green on the final HEAD SHA.

[AI-assisted - Claude]

sanity:
### Review fixes applied + re-review

All review findings addressed across two commits (`2a3f0c79`, `724e46c7`):

1. **[Should-Fix] Broadcast failure counter undercounted** (found by Codex + code-first + skeptical) — `record_attempt` only ran in the post-dispatch success arm, so the initial `bridge.send` failure and the `send_stream_with_completion` Err — the *congestion* failure modes that bias the gauge low precisely during an incident — went uncounted. Restructured into three mutually-exclusive arms so **exactly one** `record_attempt` fires per streaming invocation (initial-send fail → `false`, dispatch fail → `false`, else → `delivered`). Doc corrected.
2. **[Should-Fix] Untested delta sampling** (testing) — extracted `window_delta(current, &mut prev)` pure helper + unit test, so a future reorder of compute-then-advance (which would silently zero the incident signal) fails CI.
3. **[Consider] Per-node-vs-simulation meaning** — rustdoc caveat added on both globals.

A focused **Codex re-review** of the fixed HEAD then caught one more genuine gap (now also fixed in `724e46c7`):

4. **`refresh_router` startup read not recorded** — the initial `get_router_events` before the periodic loop wasn't recorded in the health gauge, so a node whose startup read succeeded but whose later periodic refreshes all failed would report `last_success_age_secs == null` forever, masking the startup success. Now recorded on all three startup arms (mirroring the loop), with a source-scrape pin test asserting the call-site count.

The migration version floor (`SUBSCRIBE_HINT_MIN_VERSION`) remains untouched — this ships the gauges with the migration still disabled, baselining the incident metric before any future re-enable. `cargo fmt` / `clippy -D warnings` / affected unit + pin tests all clean locally; awaiting full CI green on the current HEAD before merge.

[AI-assisted - Claude]

Issue #3519, opened 2026-03-11
fix: zombie transport cleanup deadlocks event loop on busy gateways


sanity:
## Problem

The vega gateway (100.27.151.80:31337) repeatedly dies after ~21 hours of operation. The gateway process stays alive (17% CPU, 1.5GB RSS) but the event loop stops processing — all threads deadlock on `futex_wait_queue`.

**Impact**: All peers using vega as their bootstrap gateway cannot connect to the network. IvvorMk2's node was affected (diagnostic report XTMU9J).

## Root Cause

Circular deadlock between the event loop and the handshake driver during zombie transport cleanup.

### Death sequence (confirmed on Mar 9 and Mar 11, 2026)

1. Every 30s, the event loop runs zombie transport cleanup (`p2p_protoc.rs:882-908`)
2. It iterates over all zombie transports in a `for` loop, calling `drop_connection_by_addr().await` for each
3. Each call awaits on `handshake_cmd_sender.send(DropConnection)` — **bounded channel(128)**
4. While the event loop is blocked in this for loop, it does NOT poll the `HandshakeHandler` stream
5. Meanwhile, inbound connections keep arriving (gateway receives 200+ transport connections)
6. The handshake driver accepts inbound connections and tries `events_tx.send(InboundConnection).await` — **bounded channel(128)**
7. Events channel fills up since the event loop isn't draining it
8. **Handshake driver blocks** on `events_tx.send()`
9. Since the handshake driver is blocked, it stops consuming commands
10. Command channel fills up (128)
11. **Event loop blocks** on `handshake_cmd_sender.send()`
12. **DEADLOCK**: event loop → handshake command channel → handshake driver → event channel → event loop

### Evidence

**Mar 9 death** (17:36 UTC):
```
17:36:37 Event loop stats: ring_connections=19, active_connections=260
17:36:37 893 zombie transports dropped ← massive batch
17:36:37 Event loop never produces stats again
```

**Mar 11 death** (01:37 UTC):
```
01:37:00 Event loop stats: iterations=915, pending=268 ← notification channel backing up
01:37:30 Event loop stats: iterations=1164, pending=150, ring=13 ← LAST STATS EVER
01:37:30 26 zombie transports dropped
01:37:31 Event loop never produces stats again
```

Process state after deadlock:
```
$ cat /proc/1935438/status | grep State
State: S (sleeping)
$ for tid in $(ls /proc/1935438/task); do
[...]

PR #4499, opened 2026-06-19
fix(p2p): make event-loop channel sends non-blocking to close the #4145 wedge

sanity:
## Problem

The event loop (`run_event_listener_with_socket`) runs `ctx.process_select_result(...).await` **inline on the event-loop task**, so a bare `.send().await` on a bounded channel from that path is a self-stall: under a fan-out spike the channel fills, the send blocks, and the loop can't drain the very channel it's blocked on. This is the **root weakness behind #4145** — the broadcast-backpressure wedge that the v0.2.73 placement migration *amplified* into a network-wide incident, and that recurred organically on 2026-06-17 with the migration already off. Recent merges (#4442/#4475/#4468/#4489) removed storm *triggers* but left the structural wedge: bare blocking `.send().await` calls remained on the event-loop path, including ones on the migration's own connection-establishment fan-out.

This is the **blocker for re-enabling the SubscribeHint placement migration** (#4440).

## Solution

Convert every inline-on-event-loop blocking send to a non-blocking / bounded form with explicit, documented drop-recovery, following the in-tree `drop_zombie_connection` (#3519) precedent and `.claude/rules/channel-safety.md`.

**Event-loop-task sites → `try_send` (drop + log on full; the loop must never block on a channel it drains):**

- **Site 1 — `ConnectPeer` enqueue** (`ev_listener_tx`, drained by the loop itself = true self-deadlock; the arm the SubscribeHint nudge drives): on Full → warn + drop the outbound attempt (op retry re-routes). The 20s resend-waiter is spawned **only on successful enqueue**, so a dropped enqueue can't leave a waiter hanging on a never-requested connection (pinned, incl. a spawn-ordering pin).
- **Sites 2–7 — six `HandshakeCommand` sends** (`handshake_cmd_sender`): the 4 DropConnection cleanups are mechanical (redundant with ring/connection teardown already done at each site; a missed one leaves at most one stale per-addr `ExpectedInboundTracker` entry — bounded, overwritten on the next `register()`); ExpectInbound is recoverable (inbound accepted provisionally + op retry); Connect runs the same callback-failure cleanup the old blocking error path used (now extracted to `fail_awaiting_connection` so its invariants are unit-tested). Removed the now-unused blocking `CommandSender::send`.
- **Four inline-on-loop `P2pBridge::send` callers** (added in review round 1): both `on_ring_connection_established` interest/cache fan-out loops (in `handle_connect_peer` and `handle_successful_connection` — the connection-churn the migration drives), plus `handle_send_interest_message` and `handle_send_net_message`. New `P2pBridge::try_send` (non-blocking sibling of `send`). Recovery: fan-out/interest re-syncs via the periodic InterestSync + subscription-renewal task; the CONNECT-joiner net message falls back to the upstream op's TTL. The OFF-loop callers (`broadcast_to_peers`' spawned task, sim-only `broadcast_state_to_peers`, and the op-driver/contract `NetworkBridge::send` callers) stay blocking — they produce *into* the loop, so blocking is correct back-pressure (`P2pBridge::send` carries a `// channel-safety: ok` annotation saying so).

**Site 8 — handshake-driver inbound accept** (`events_tx`) → **bounded `events_tx.send_timeout(event, 5s)`**, NOT an unconditional drop. `run_driver` is the handshake driver task, NOT the event loop, so a *bounded* block there cannot reintroduce the #4145 self-stall — and a bounded wait preserves the inbound connection across a transient event-loop backlog instead of forcing a slower reconnect. Ok → forwarded; `Timeout` (loop wedged past the bound) → warn + drop as a last resort, keep accepting; `Closed` (event loop gone) → break. The 5s wait briefly pauses the sibling `commands_rx` arm, which is safe because every command sender is now a non-blocking `try_send` that drops rather than blocks. (Round-1 review (Codex) flagged that an unconditional drop could force a slower reconnect of a recoverable inbound; investigation confirmed a dropped `InboundConnection` is ultimately recovered — the dialer's `spawn_outbound` resolves to `OutboundEstablished`/`OutboundFailed` within 10s and `OutboundFailed` clears `awaiting_connection` — but `send_timeout` is the strictly-safer choice and avoids the reconnect.)

**CI lint (`rule_lint` #6):** a Python matcher (`.github/scripts/check_blocking_sends.py`, with its own `--self-test` CI step) flags newly-added blocking `.send(...).await` on event-loop-reachable senders, diff-scoped to added lines, with a `// channel-safety: ok — <reason>` escape hatch honored on the call line or the line above. It joins multi-line statements, blanks string/comment contents (so prose/asserts don't trip it), matches the `P2pBridge::send` wrapper form (incl. `send_stream` / `pipe_stream` / `send_stream_with_completion`), and reconstructs statements from full-file context to catch sends split across added + unchanged lines. Documented (with its known limits) in channel-safety.md.

Every dropped/timed-out message has a documented recovery path (op retry / re-request / periodic re-sync / lease expiry) — this trades "can wedge" for bounded, self-healing loss, never silent connectivity degradation.

## Testing

- `cargo clippy --locked -- -D warnings` (CI's exact gate) pass; `cargo fmt --check` clean; lint `--self-test` + lint-on-own-diff clean.
- `node::network_bridge` + `operations::connect` (171 tests) pass, including the new #4145 tests: resend-waiter no-hang + spawn-ordering pins, the four inline-`bridge.try_send` pin, the Site-6 connect-command-drop behavioral test (`fail_awaiting_connection`: callbacks → `Err(())`, both maps drained, no backoff) + prune source pin, and the Site-8 `send_timeout` pin + bounded-semantics behavioral test.
- `test_fanout_liveness_8_subscribers` passes (no stale peers, bounded backlog, convergence) — no liveness regression from the fan-out conversions.
- **Incident-scale (40–80-subscriber) validation is in progress separately** (needs a topology-preseed harness primitive, since organic CONNECT caps the sim star at N≈16) — will be added before merge.

## Review focus

Site 1's resend-waiter short-circuit and the Site 8 `send_timeout` / `awaiting_connection` reconciliation are the subtle, non-mechanical parts.

Addresses #4145 (root blocking-send fix). Unblocks #4440 (SubscribeHint re-enable).

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #4499 (Full tier — high-risk concurrency change)

**Reviewers:** code-first, testing, skeptical, big-picture (freenet) + external Codex. The 8 named conversions are **individually sound** (Site 1, Site 7 cleanup matches the old path line-for-line, exceptions correctly left blocking). But the review found the fix is **incomplete and the lint is broken**, so this is **Needs Changes — re-review required**.

### Must Fix (blocking)

1. **CI is RED — the new lint trips on its own test code.** The Site 8 pin's assertion string puts `events_tx...send(...).await` on one line; the lint's `sed 's|//.*||'` strips `//` comments but not string-literal contents, so it matches its own test → Rule Lint FAILURE. (skeptical) Fix: split the string across lines and/or harden the lint against string literals.

2. **The #4145 root is NOT fully closed — 4 same-class self-stall paths remain** (confirmed independently by big-picture AND code-first). Four event-loop handlers call `self.bridge.send(...).await`, and `P2pBridge::send` does a blocking `ev_listener_tx.send().await` on the *same* cap-512 channel the loop drains:
   - `handle_connect_peer` fan-out loop (`p2p_protoc.rs:3273`)
   - `handle_successful_connection` fan-out loop (`p2p_protoc.rs:4115`)
   - `handle_send_interest_message` (`p2p_protoc.rs:4754`)
   - `handle_send_net_message` (`p2p_protoc.rs:4781`)
   The two fan-out loops fire on exactly the connection-churn the SubscribeHint nudge drives, and are the "don't fan out inline on the event loop" anti-pattern the cap-1027 comment warns about. For a PR whose purpose is to make the migration re-enable safe, these must be converted (or the wedge survives). Convert them (e.g. a non-blocking `NetworkBridge::send` variant or `send_timeout`).

3. **Site 8 (`InboundConnection` drop) can strand a pending connect.** It's the event that clears local `awaiting_connection`/callbacks; a later `ConnectPeer` only appends to an occupied entry (p2p_protoc.rs:3406), and there's no TTL draining stale entries. Codex flagged this; skeptical judged it safe via peer-retry, but that depends on the (already-established) transport connection breaking. Resolution: use a **bounded `send_timeout` at Site 8 instead of unconditional drop** — Site 8 runs on `run_driver`, not the event loop, so a bounded block there is safe and preserves the load-bearing event.

### Should Fix

4. **Lint multi-line blind spot** (codex, testing, skeptical, code-first — 4 reviewers, verified): only matches single-line `.send().await`; the rustfmt-natural multi-line form — how every converted site was originally written — evades it. Make the matcher span the logical statement; add a lint self-test.
5. **Lint escape-hatch contradiction** (code-first): `// channel-safety: ok` only works same-line, but the docs + `spawn_outbound` example put it on the line above → false CI failures for future contributors. Reconcile lint + docs + example.
6. **Lint `grep -v try_send` over-exemption** (skeptical): any line *containing* `try_send` is fully exempt, even with a real blocking send on it.
7. **Extend the lint to the wrapper form** `bridge.send(...).await` (code-first, big-picture) — the most common way to hit the channel, currently invisible.
8. **Site 6 (`connect_peer` Connect drop) needs a behavioral test** (testing): the one drop with real recovery logic (prune in-transit + fail callbacks with `Err(())`); currently only substring-pinned — a leak/park bug would be uncaught.
9. **Incident-scale wedge regression test** (testing, big-picture): an 8-subscriber liveness check can't reproduce the 40–80 regime where #4145 manifests. Being built separately (topology-preseed primitive); must land here before merge.

### Consider
10. Fix the factually-wrong "expected-inbound self-reclaims on its own timeout" comment (skeptical) — no such timeout exists; the leak is bounded for other reasons (per-addr overwrite, `transient:false` so no #3838).
11. Add a spawn-ordering pin to `dropped_connect_peer_does_not_hang_resend_waiter` (skeptical, testing) — current test proves the channel mechanic, not that the waiter spawns only after a successful try_send.
12. Note in the description that the `connect_peer` Full trigger widened from Closed-only to Full-or-Closed (code-first).

### Verdict
**Needs Changes — Re-review Required.** HEAD reviewed: `aaceb19b`. The conversions don't need reverting, but the fix must (1) go green, (2) close the 4 remaining self-stall paths, (3) make Site 8 non-stranding, (4) harden the lint, (5) add the connect_peer + incident-scale tests. Then a fresh review on the new HEAD.

[AI-assisted - Claude]

sanity (commented):
## Re-Review After Rework: #4499 (HEAD 7e72ca6b)

Fresh Full re-review on the reworked content (external Codex + skeptical + code-first). **All three Round-1 Must-Fix items are resolved, and all three passes independently confirm the production fix is correct and complete.**

### Verified resolved (the Round-1 blockers)
1. **CI-red lint → fixed.** Replaced the line-grep with `check_blocking_sends.py` (blanks string/comment contents, joins multi-line statements, self-tested 10 good / 4 bad). Rule Lint now passes; CLEAN mergeStateStatus.
2. **4 missed self-stall paths → closed.** Added `P2pBridge::try_send` and converted the 4 inline-on-loop callers (`handle_connect_peer`, `handle_successful_connection` fan-out loops, `handle_send_interest_message`, `handle_send_net_message`). **Crate-wide sweep (by two reviewers) confirms no inline-on-event-loop blocking send remains**; the off-loop callers (NetworkBridge::send producer API, broadcast_to_peers/broadcast_queue spawned workers, spawn_outbound, the resend-waiter) are correctly left blocking. Drop-recovery verified to exist (ReadyState re-broadcasts every 30s; Interests via InterestSync; unicast via op TTL — no peer left permanently un-announced).
3. **Site 8 stranding → resolved + the concern itself disproven.** Changed to bounded `events_tx.send_timeout(5s)`. Skeptical + code-first traced the recovery end-to-end: `awaiting_connection` is only created by `connect_peer` (which always dials outbound), and `OutboundFailed`'s 10s timeout always clears it — `InboundConnection` is never the *sole* resolver, so the timeout-drop branch at worst loses a redundant completion. No stranding. `send_timeout` is strictly safer regardless and runs off the event-loop task.

Also verified: `P2pBridge::try_send` correct, `fail_awaiting_connection` extraction behavior-identical, the connect_peer behavioral test + spawn-ordering pin + bounded-send test all pass, no over-conversion, removed blocking `CommandSender::send` is a good structural guard.

### Remaining (all Low severity — follow-ups, not blockers)
- **Lint wrapper matcher** should also catch `bridge.send_stream(` / `pipe_stream(` / `send_stream_with_completion(` (same cap-512 channel; no on-loop caller today, forward-looking) and the split-across-added/unchanged-line case (Codex P2). Being tightened.
- **Doc:** PR-description Site 8 bullet is stale (says drop; code is `send_timeout`); Site 8 comment should note the 5s timeout briefly pauses handshake-command draining. Being fixed.

### Verdict
**Production fix: Ready** (pending the lint/doc polish + the incident-scale 40–80 validation test, which is being built via a topology-preseed harness primitive and will be added before merge). HEAD reviewed: `7e72ca6b`. No finding reopens the #4145 wedge.

[AI-assisted - Claude]

PR #4500, opened 2026-06-19
test(sim): topology-preseed primitive + incident-scale fan-out liveness tests (#4233)

### LINKED ISSUE #4233: Event-loop liveness simulation test under sustained UPDATE fan-out

sanity:
## Problem

`#4233` asks for a fan-out broadcast liveness test at the production-incident width (≈40–80 subscribers) to gate the SubscribeHint re-enable (#4440). The existing fan-out test (`test_sustained_update_fanout_no_full_state_storm`) is a 2-subscriber star, and organic ring-routed CONNECT can't form a *wide direct star* in the harness's virtual-time budget — so incident-scale liveness was untestable.

## Solution

A test-only **topology-preseed primitive** (mirrors the existing `SeedHostedContract` contract-preseed) that injects pre-formed gateway↔subscriber connections at startup via `Ring::add_connection`, bypassing the CONNECT handshake. Only the handshake is replaced — subscribe + interest registration + broadcast fan-out all run organically, so the broadcast/event-loop path is genuinely exercised. Zero production-path change (the preseed vec is empty for every existing sim).

- `Builder.preseed_connections` + `apply_preseeded_connections` (`testing_impl.rs`, `in_memory.rs`)
- `SimNetwork::preseed_direct_star(host, &subscribers)` (`testing_impl.rs`)
- Fixed a harness budget bug: the controlled-sim client paces ops ~3s apart, so O(N) ops overran the fixed 300s budget at scale → `simulation_duration` now scales with N.
- Two pre-existing test-only `--tests` clippy warnings cleared along the way (`topology/rate.rs`, `tracing/telemetry.rs`) — flagged separately for the reviewer.

## Testing

- `test_fanout_liveness_preseed_24_subscribers` — always-on guard (the smallest width organic CONNECT can't form).
- `test_fanout_liveness_incident_scale_migration_off` / `_on` — `#[ignore]`'d (heavy, ~70min), N=40/60/80.

**Diagnostic result (current main):** main does NOT wedge at N=40/60/80 in either condition — all converge, zero stale peers, bounded backlog. These tests also validated the #4145 fix (#4499) at incident scale: it converges with no connectivity shedding (migration-ON N=40 logged 39 resyncs, all absorbed by the recovery path). See #4440.

**Caveat:** the deterministic sim does not reproduce the fd-exhaustion / module-cache axis of the v0.2.73 incident; this covers the broadcast/event-loop layer.

Closes #4233.

[AI-assisted - Claude]

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -28,13 +28,15 @@ slow-timeout = { period = "300s", terminate-after = 2 }
 filter = "package(freenet-ping-app)"
 slow-timeout = { period = "300s", terminate-after = 3 }
 
-# #4233 wide-star UPDATE fan-out liveness tests (40/60/80 direct subscribers).
-# These drive a 1-gateway / N-subscriber star through 300s of virtual time in
-# the paused-time Turmoil runner, advanced in tiny quanta. Wall-clock cost
-# scales with N (each of the N subscriber event loops is stepped every quantum),
-# so the wide-star variants need far more than the default 240s cap. See
-# test_fanout_liveness_* / run_fanout_liveness_scenario in
-# tests/simulation_integration.rs (issue #4233).
+# #4233 wide-star UPDATE fan-out liveness tests. This filter matches the
+# always-on variants (N=8/16 organic, N=24 preseed) AND the #[ignore]'d
+# incident-scale ones (N=40/60/80). Each drives a 1-gateway / N-subscriber star
+# in the paused-time Turmoil runner, advanced in tiny quanta. The virtual-time
+# budget scales with N in the driver (run_fanout_liveness_scenario_inner), and
+# wall-clock cost scales with N too (each of the N subscriber event loops is
+# stepped every quantum), so these need far more than the default 240s cap; the
+# always-on N=24 runs ~250-300s wall-clock. See test_fanout_liveness_* /
+# run_fanout_liveness_scenario in tests/simulation_integration.rs (issue #4233).
 [[profile.ci.overrides]]
 filter = "test(test_fanout_liveness)"
 slow-timeout = { period = "600s", terminate-after = 2 }
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -28,6 +28,18 @@ slow-timeout = { period = "300s", terminate-after = 2 }
 filter = "package(freenet-ping-app)"
 slow-timeout = { period = "300s", terminate-after = 3 }
 
+# #4233 wide-star UPDATE fan-out liveness tests (40/60/80 direct subscribers).
+# These drive a 1-gateway / N-subscriber star through 300s of virtual time in
+# the paused-time Turmoil runner, advanced in tiny quanta. Wall-clock cost
+# scales with N (each of the N subscriber event loops is stepped every quantum),
+# so the wide-star variants need far more than the default 240s cap. See
+# test_fanout_liveness_* / run_fanout_liveness_scenario in
+# tests/simulation_integration.rs (issue #4233).
+[[profile.ci.overrides]]
+filter = "test(test_fanout_liveness)"
+slow-timeout = { period = "600s", terminate-after = 2 }
+retries = 0
+
 # test_ping_blocked_peers is the canary for the CONNECT promotion-drop bug
 # class (#3838, fixed in #3893). The profile-wide retries = 2 masked a ~40%
 # per-run failure rate as ~6% effective during the flaky period, making an
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -754,6 +754,14 @@ pub(super) struct Builder<ER> {
     event_register: ER,
     contracts: Vec<(ContractContainer, WrappedState, bool)>,
     contract_subscribers: HashMap<ContractKey, Vec<PeerKeyLocation>>,
+    /// Pre-formed ring connections injected at node startup, bypassing the
+    /// organic CONNECT handshake. Each entry is a peer this node should already
+    /// hold a direct ring connection to when its event loop starts. Drained by
+    /// `run_node_with_{shared_storage,mock_wasm}` into
+    /// `apply_preseeded_connections` (see that helper for the rationale).
+    /// Populated by [`SimNetwork::preseed_direct_star`]; empty for every node by
+    /// default, so this is a zero-cost no-op for all existing simulations.
+    preseed_connections: Vec<PeerKeyLocation>,
     /// Seed for deterministic RNG in this node's transport layer
     pub rng_seed: u64,
     /// Network name for scoped fault injection
@@ -782,6 +790,7 @@ impl<ER: NetEventRegister> Builder<ER> {
             event_register,
             contracts: Vec::new(),
             contract_subscribers: HashMap::new(),
+            preseed_connections: Vec::new(),
             rng_seed,
             network_name,
             shared_cm: None,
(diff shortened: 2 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #4500

### Summary
- **PR Title:** test(sim): topology-preseed primitive + incident-scale fan-out liveness tests (#4233)
- **Type:** test (test-only; production and harness code unchanged)
- **CI Status:** green on the reviewed HEAD
- **Linked Issues:** Closes #4233; gates the #4440 SubscribeHint re-enable
- **Review tier:** Light (test-only diff; no production/concurrency/wire-format/crypto surface) — run with extra perspectives because this test gates a production re-enable decision
- **Reviewers run:** `freenet:skeptical-reviewer`, `freenet:testing-reviewer`, `freenet:big-picture-reviewer`, plus an external **Codex** (gpt-5) pass

---

### Code-First Analysis
**What the code does:** Adds a test-only topology-preseed primitive (`SimNetwork::preseed_direct_star` → `apply_preseeded_connections`, injecting gateway↔subscriber ring connections via `Ring::add_connection`) so a wide direct star the organic CONNECT path can't form in-harness exists at t=0, plus incident-scale UPDATE fan-out liveness tests (always-on N=8/16 organic + N=24 preseed; `#[ignore]`'d N=40/60/80).
**Stated intent:** A simulation-level event-loop liveness gate under sustained UPDATE fan-out at incident width, to gate #4440. **Alignment:** matches — purely additive (796 insertions, 1 deletion), correctly does NOT re-enable the migration.

---

### Testing Assessment
**Coverage Level:** adequate for the goal.

| Unit | ⚠️ | No cheap unit test of the preseed primitive itself (cap-rejection-skip path documented but unasserted) — non-blocking follow-up |
| Integration/Simulation | ✅ | N=24 preseed guard now reaches full 24-wide fan-out (`targets_avg`→24) with the strict `host + all N` participation assertion; non-vacuous |
| E2E | N/A | |

**Key fact verified by reviewers:** a plain GET does **not** log a `stored_state_hash`, so `(e)`'s `replica_count == subscribers + 1` genuinely requires every subscriber to have *applied a broadcast* — not vacuous. `(a)` stale-peers and `(d)` convergence are both blind to never-joined peers (`check_convergence_from_logs` skips <2-replica contracts; the StalePeer detector only flags peers that logged a transition), so `(e)` is the assertion that closes the partial-participation gap an earlier review found.

---

### Skeptical / External Findings (all addressed)

| # | Finding | Severity | Location | Resolution |
|---|---------|----------|----------|------------|
| M1 | `(e)` participation assertion ran under the migration-ON path too, where the PR's own rationale says the replica set is no longer deterministically `host + all subscribers` (hosting migrates off the gateway) — the `#[ignore]`'d migration-ON diagnostic would fail `(e)` for a benign reason | Medium | `simulation_integration.rs` `(e)` gate | **Fixed** — scoped to `if preseed && !enable_migration` with rationale |
| N2 | Stale rustdoc on `enable_placement_migration` claimed the cascade is OFF-by-default in sim; no longer true (build 0.2.81 > `SUBSCRIBE_HINT_MIN_VERSION=(0,2,80)`), which is exactly why this PR adds `disable_placement_migration` | Minor | `testing_impl.rs:1271` | **Fixed** — rustdoc corrected |
| C1 | `.config/nextest.toml` comment said "40/60/80 subscribers / 300s" but the override also covers N=8/16/24 and the driver scales virtual time with N | Nit | `nextest.toml:31` | **Fixed** — comment corrected |

**Verified safe (no action):** production no-op (`preseed_connections` empty for every non-preseed sim → `apply_preseeded_connections` is a true empty-loop; all `run_node*` consume `self` by value, vec moved once); `Ring::add_connection(Location::from_address(addr), …, was_reserved=false)` matches the repo's address-derived location model and the injected per-peer location equals each peer's self-assigned location; `subscribers + 1` is the correct deterministic replica count; `max_connections = subscribers + 16` prevents cap-rejection; cap-rejection is logged+skipped (not a panic); determinism intact (fixed seeds, `GlobalTestMetrics::reset` + `setup_deterministic_state`, per-network cleanup). The `GET(return_contract_code:true)+subscribe` change is the correct exempt path past the `client_events.rs` WASM-cache subscribe guard.

---

### Big Picture Assessment
**Goal Alignment:** yes — delivers the long-missing simulation guard for the #4145 channel-saturation bug class and the gate #4440 requires, without re-enabling the migration. **Scope:** clean — the two earlier clippy fixes are no longer in the diff (main's #4542 landed them; PR rebased), no scope creep. **Removed code:** none (the single deletion is adding `Location` to a `use` line). **Honest caveat:** the PR states the deterministic sim does NOT reproduce the fd-exhaustion / module-cache axis of the v0.2.73 incident — correct disclosure so the gate isn't over-trusted.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important) — addressed in this PR
1. M1 — `(e)` scoped to `preseed && !enable_migration`. ✅
2. N2 — stale migration rustdoc corrected. ✅
3. C1 — nextest comment corrected. ✅

#### Consider (non-blocking follow-ups, not done here to keep scope tight)
1. **Organic N=8/16 are weak guards.** They're gated only by (a)–(d), which are blind to never-joined subscribers, so a degenerate organic star could pass. Scoping (e) out of them is *correct* (organic ring-CONNECT is best-effort at the harness bootstrap ceiling — the limitation the preseed primitive exists to bypass), but a *soft* minimum-participation floor would harden them. The strong always-on guard is the preseed-24 test, which now carries the strict participation gate.
2. **Add a cheap unit test of the preseed primitive** asserting the star is formed in the ring and that the documented cap-rejection-skip path doesn't panic — would catch a primitive regression in seconds instead of via the ~5-min sim.
3. **When #4440 actions the re-enable, run `--ignored test_fanout_liveness_incident_scale_migration_on`** as part of the decision, and re-generate the incident-scale diagnostic figures — the earlier "main doesn't wedge at N=40/60/80" numbers in the PR body predate the participation fix and the migration-default-on correction, so they should be re-measured before being relied on.

---

### Verdict
**State:** Ready to Merge (test-only; all review findings fixed; the strict N=24 participation gate is non-vacuous). Merging once CI is green on the current HEAD.
**HEAD SHA reviewed:** `7c513083c219862da08fc183d85f76e496a941ae` (includes the M1/N2/C1 fixes)

[AI-assisted - Claude]

sanity:
Rebased onto current `main` (clean fast-forward — main's 4 intervening commits touch none of this PR's 4 files, so the diff is byte-identical to the reviewed content). New HEAD: `26f943561e46d0fe07a2fe315177f195054f6956`. The review above stands; merging once CI is green on this SHA.

[AI-assisted - Claude]

Issue #4381, opened 2026-06-11
RFC: Public Freenet proxy to try apps before installing a peer (with data export)


sanity:
## Summary

Explore what it would take to run a **public Freenet proxy**: a hosted peer whose local HTTP service (port 7509) is exposed to the internet so that anyone can open a Freenet web app, e.g. River at

```
http://127.0.0.1:7509/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/
```

in a normal browser **without installing a peer first**. The goal is frictionless "try before you install": let people use real apps immediately, and when they're ready, **download their private data and import it into their own local peer**.

This issue is a write-up of the challenges after investigating the current architecture. It is deliberately honest about the hard tradeoffs, because the central one is uncomfortable: **a public peer that stores and operates on users' private delegate data is a trusted, centralized intermediary that can read everything**, which is in tension with Freenet's whole premise. That doesn't kill the idea (ephemeral try-before-install is still valuable), but it shapes what we can promise users and how loudly we have to disclaim it.

This is `S-needs-design`, not a work order. The point is to surface the problems so we can decide whether and how to scope it.

---

## What "public proxy" means concretely

The HTTP service on 7509 already serves:
- Contract **web apps** at `/v{1,2}/contract/web/{key}/...` (the React/Dioxus SPA, its assets, an iframe sandbox)
- The **WebSocket API** at `/v{1,2}/contract/command` (the app talks to the node here: contract GET/PUT/SUBSCRIBE and delegate messages)
- A **node dashboard** at `/` and `/peer/{address}`
- Delegate **permission prompts** at `/permission/*`

"Public proxy" means making (a subset of) those reachable from the open internet, with a per-user notion of identity so that different visitors don't clobber each other's private delegate data.

---

## Current architecture (what exists today)

References are to `freenet-core/main` unless noted.

### HTTP gateway
- **axum** server, set up in `crates/core/src/server.rs` then `crates/core/src/server/client_api.rs` (`as_router_with_origin_contracts`).
- In **network mode the API binds `::` (all interfaces)** by default (`config.rs` `default_listening_address()`), *not* loopback, but it is **not** open: see the security model below.
[...]

PR #4502, opened 2026-06-19
feat(secrets): optional per-user dimension for delegate secrets (P1 of #4381)

sanity:
## Problem

Phase 1 of #4381 (hosted public proxy). The hosted-proxy design needs delegate secrets to be partitionable per end-user so that, eventually, one delegate instance can hold isolated secret namespaces for many users — and so a user's secrets can be exported/imported portably (P3). Today `SecretsStore` is single-user: every secret is keyed only by `(delegate, secret_id)` and encrypted under a DEK derived from the node KEK.

This PR adds the storage-layer per-user dimension and lands it **inert**: no production caller uses the new scope yet. The token threading from the WS layer, a hosted-mode flag, and export/import are deliberately deferred to P2/P3.

## Approach

A new `SecretScope<'a>` selector is threaded through `store_secret` / `get_secret` / `remove_secret`:

- **`SecretScope::Local`** — the historical single-user path, kept **byte-for-byte identical**: same on-disk path (`secrets_dir/<delegate>/<secret>`), same node-KEK-derived DEK (`derive_delegate_dek`), same `secrets_index` ReDb table/format, same `[VERSION_V1][24-byte nonce][AEAD]` blob layout, same legacy-decrypt fallback chain. The three production callers (delegate host functions in `native_api.rs`) all pass `Local`, so node behavior is unchanged.
- **`SecretScope::User { id, dek_secret }`** — writes under `secrets_dir/<delegate>/users/<user_id>/`. The `users/` segment can't collide with a secret-id file or `.snapshots/` (siblings; a bs58 32-byte hash is never the ASCII `"users"`). Encryption uses a DEK derived **solely** from the caller-provided `dek_secret` (`HKDF-SHA256(ikm = dek_secret, salt = delegate.encode(), info = "freenet-delegate-dek-user-v1")`) — the node KEK is intentionally **not** involved, which is what makes per-user secrets portable across nodes (P3 export). Same per-write random nonce + blob format.

### ReDb back-compat

The single-user `secrets_index` table is **untouched**. The per-user dimension lives in a **separate** `user_secrets_index` table keyed by `DelegateKey(64) || UserId(32)`. A pre-#4381 database simply lacks the new table; redb creates it empty on first open, so old DBs load unchanged and old `secrets_index` rows are never perturbed by a User write. Each backend has its own in-memory mirror (`key_to_secret_part` vs `user_key_to_secret_part`), populated on startup and updated only after the corresponding ReDb write commits.

### Token-derivation helpers (also inert)

`user_id(token) = blake3("freenet-user-id" || token)` and `user_dek_secret(token) = blake3("freenet-user-dek" || token)` (the latter `Zeroizing`). Distinct domain prefixes guarantee `user_id(token) != user_dek_secret(token)`. P2 will consume these; they're landed now (with tests) so the derivation is pinned before any caller depends on it.

`dek_secret` is held as a borrowed `&Zeroizing<[u8; 32]>` and never copied into a longer-lived buffer; the token, dek_secret, and plaintext are never logged.

## Testing

`cargo test -p freenet --lib secret` — **81 passed, 0 failed**. New tests (the acceptance gate):

- **No-regression**: explicit assertion that a `Local` secret lands at the exact legacy path (no `users/` dir) with the canonical blob layout; and that Local vs User writes touch disjoint ReDb tables.
- **Adversarial cross-user isolation**: a secret written under user A is unreadable under user B (wrong `dek_secret` → AEAD `Encryption` error), absent from B's namespace (distinct path → `MissingSecret`), invisible to/from `Local`, and the same `SecretsId` holds independent values per user; per-scoped remove leaves the other scopes intact.
- **ReDb back-compat**: a pre-existing single-user index entry loads, adding User entries doesn't perturb it, and User secrets round-trip across a full store drop+reopen (in-memory user index rehydrates from ReDb).
- **DEK correctness**: same `(delegate, dek_secret)` → same key; different `dek_secret` → different key/ciphertext; the user DEK is independent of the node KEK (proven both at the derivation level with two distinct-KEK stores, and end-to-end by moving a User blob between two nodes and decrypting with the same `dek_secret`).
- **Domain separation**: `user_id(token) != user_dek_secret(token)`, both deterministic.
- **Perms**: `users/<id>` dir is 0o700 and the user secret file is 0o600 on Unix.

`cargo test -p freenet --lib storages::redb` (12) and `cargo test -p freenet --lib wasm_runtime` (297) pass. `cargo clippy -p freenet --lib --tests` is clean for the touched files (the 2 remaining warnings are pre-existing in `topology/rate.rs` and `tracing/telemetry.rs`, untouched here). `cargo fmt` applied.

Refs #4381

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/storages/redb.rs
+++ b/crates/core/src/contract/storages/redb.rs
@@ -47,6 +47,16 @@ pub(crate) const BROKEN_INVARIANTS_TABLE: TableDefinition<&[u8], &[u8]> =
 pub(crate) const SECRETS_INDEX_TABLE: TableDefinition<&[u8], &[u8]> =
     TableDefinition::new("secrets_index");
 
+/// Index table for the per-user secret dimension (P1 of #4381). SEPARATE
+/// from [`SECRETS_INDEX_TABLE`] so the existing single-user index keeps its
+/// exact schema and a pre-#4381 database opens unchanged (this table is
+/// simply absent → created empty on first open).
+///
+/// Key: DelegateKey (64 bytes) || UserId (32 bytes) = 96 bytes
+/// Value: Concatenated secret key hashes (N * 32 bytes)
+pub(crate) const USER_SECRETS_INDEX_TABLE: TableDefinition<&[u8], &[u8]> =
+    TableDefinition::new("user_secrets_index");
+
 /// Metadata about a hosted contract, persisted to survive restarts.
 #[derive(Debug, Clone, Copy)]
 pub struct HostingMetadata {
@@ -222,6 +232,20 @@ impl ReDb {
                     table = "SECRETS_INDEX_TABLE",
                     phase = "table_init_failed",
                     "Failed to open SECRETS_INDEX_TABLE"
+                );
+                e
+            })?;
+
+            // Per-user secrets index (P1 of #4381). Created on first open of
+            // upgraded databases too — redb creates missing tables inside the
+            // same write txn that opens them, so old DBs gain an empty table
+            // without disturbing the single-user SECRETS_INDEX_TABLE above.
+            txn.open_table(USER_SECRETS_INDEX_TABLE).map_err(|e| {
+                tracing::error!(
+                    error = %e,
+                    table = "USER_SECRETS_INDEX_TABLE",
+                    phase = "table_init_failed",
+                    "Failed to open USER_SECRETS_INDEX_TABLE"
                 );
                 e
             })?;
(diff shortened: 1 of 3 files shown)


sanity:
## Multi-model review (Full tier — crypto/auth surface)

Reviewed the checked-out code with four independent reviewers plus a maintainer code-first read. **Consensus: no blocker/major findings.** The two load-bearing invariants hold:

- **Local (single-user) path is byte-for-byte unchanged** — same on-disk paths, node-KEK DEK via `cipher_for`, the original `secrets_index` ReDb table, legacy-fallback chain, snapshot logic, and ReDb-then-memory index ordering. The only Local-arm delta is a semantically-inert `.clone()` to unify the match arms. Proven by `local_scope_uses_legacy_path_and_blob_layout` + `local_and_user_writes_touch_disjoint_redb_tables`.
- **Cross-user isolation holds** — per-user DEK = `HKDF(ikm=dek_secret, salt=delegate.encode(), info="…-user-v1")`, node-KEK-independent; separate `users/<id>/` dir, separate `user_secrets_index` table + in-memory map; User read uses an empty legacy chain so a wrong `dek_secret` fails AEAD cleanly. bs58(BITCOIN) encoding rules out path traversal / `"users"` collision. Boot-path `try_into().unwrap()`s are guarded by length checks.

### Reviewers
- **Codex `codex review --base main` (external, non-Claude): clean, zero findings.** Static-only (the sandbox couldn't reach crates.io to compile) — covered by green CI on HEAD.
- **Gemini CLI: genuinely unavailable** — the free-tier OAuth path is hard-disabled (`IneligibleTierError: UNSUPPORTED_CLIENT`, not a quota window; waiting won't fix it, and no API key is configured). Per the project's multi-model-review policy I substituted **three diverse blind Claude subagents** (Local no-regression / isolation+crypto / inertness+coverage). Codex (the genuine external pass) still ran and is clean, so the independent-model slot is covered.
- **Claude adversarial crypto/isolation lens** and **Claude test-coverage lens** (separate agents): both confirm the invariants; gates (a)–(e) are proven by real assertions (KEK-independence tested two ways, remove-isolation, reopen back-compat, domain separation, 0o600/0o700 perms) — no rubber-stamping.

### Findings → resolution (all MINOR/NIT; none block the inert landing)
1. **Snapshot scope gap (the one that matters):** `store_secret`/`remove_secret` write User snapshots under `users/<id>/.snapshots/`, but `list_snapshots`/`restore_snapshot` hardcode the Local dir — so User snapshots are write-only and `restore` would silently hit the Local tree. **Fixing in a follow-up commit:** thread `SecretScope` through both, CLI callers pass `Local` (inert), + a User-scope snapshot round-trip test.
2. **Silent malformed-row skip** in `load_all_user_secrets_index` → **adding `tracing::warn!`** (Local `load_all_secrets_index` left untouched).
3. **ReDb-layer test gaps** (per-user index round-trip incl. the currently-uncalled `remove_user_secrets_index`; malformed-length skip) → **adding both tests.**
4. **Weak isolation assertion** (test varied both id and dek_secret) → **adding a same-`dek_secret`/distinct-`UserId` path-isolation test** + documenting the DEK is intentionally id-independent.
5. **Cheap:** assert `SecretsId::encode() != "users"`.

**Dismissed:** "version bump bundled in" (false positive — reviewer diffed a stale local `main`; `origin/main...HEAD` shows only the 3 source files); `derive_user_dek` unused `&self` (intentional, documented); sqlite backend lacking index methods (pre-existing on `main`, redb is default).

CI is green on HEAD `161a9d03`. Will re-verify green after the fixup commit, then this is mergeable.

[AI-assisted - Claude]

PR #4504, opened 2026-06-19
feat(hosted): hosted-mode flag + per-user token threading (P2 backend of #4381)

sanity:
## Problem

P2 backend of #4381 (hosted public proxy). A Freenet node is single-user today: a delegate's secrets live in one namespace (`SecretScope::Local`). To run a node as a shared public proxy for untrusted users, each user's delegate secrets must live in their own isolated namespace, and that per-user identity must be impossible to forge from inside the sandbox.

P1 (#4502) landed the storage layer: `SecretScope::{Local, User{id, dek_secret}}`, a per-user DEK derived purely from a token (`secrets_store::user_id` / `user_dek_secret`), and the disjoint on-disk/ReDb layout. P1 was inert: nothing constructed a `User` scope outside tests.

This PR makes it live, behind an opt-in flag, and threads a per-connection user context from the WebSocket boundary all the way to the delegate secret host functions.

## Approach

**Flag-gated, inert by default.** A new `--hosted-mode` / `hosted-mode = true` flag (`WebsocketApiConfig`, default **false**). With it off, the new `userToken` channel is ignored entirely and every secret op stays `SecretScope::Local`, byte-for-byte today's behavior. The flag's only job is to decide whether the WS boundary derives a per-user context; everything downstream is driven by *whether a context was derived*, so the whole feature is provably inert when off.

**The connection-boundary invariant (the part that matters).** A `UserSecretContext` (owns the `UserId` plus the `dek_secret` in `Zeroizing`) is constructed in **exactly one place**: the `connection_info` WS middleware, from the connection's user token, via the deterministic P1 derivations. It then travels immutably with the connection. Nothing reachable from delegate WASM, a delegate message body, a `ClientRequest`, or the app contract id can construct, mutate, or substitute it. It rides a **separate channel** from `origin_contract`/`caller_delegate`/the request body the whole way down. The raw token is never persisted or logged; `dek_secret` stays `Zeroizing` and is redacted in `Debug`.

**Token channel.** The browser shell page (P2-frontend follow-up) supplies the token as the `userToken` query parameter on the WS upgrade URL, chosen to mirror the existing `auth_token` query form, which is the simplest thing for the shell page to put on the WS URL. An `X-Freenet-User-Token` header is also accepted and takes precedence when present (mirrors how a Bearer `auth_token` header overrides its query form), for non-browser clients / proxies.

**Threading path** (`Option<UserSecretContext>` alongside, never inside, the request):

`connection_info` (derive once) -> `Extension<Option<UserSecretContext>>` -> `ClientConnection::Request.user_context` -> `OpenRequest.user_context` -> `ContractHandlerEvent::DelegateRequest.user_context` -> `handle_delegate_with_contract_requests` -> `execute_delegate_request` -> `delegate_request` -> `inbound_app_message` -> `exec_inbound_with_env` -> `DelegateCallEnv.user_context` (owned, `Zeroizing` lives for the call) -> the secret host fns (`get_secret`/`get_secret_len`/`set_secret`/`remove_secret`), which call `env.secret_scope()` -> `SecretScope::User{..}` when present, else `Local`.

Inter-delegate dispatch (`SendDelegateMessage`) and contract-state-change notifications pass `None`: a delegate-to-delegate hop does **not** inherit the originating connection's user namespace. The namespace binds to the connection boundary, not transitively through delegate messages.

**`DelegateCallEnv` ownership.** The env stores an **owned** `Option<UserSecretContext>` (cloned from the borrowed connection context), so the `dek_secret` outlives every secret call during the synchronous `process()` invocation. `secret_scope()` borrows from it (`&self`), so the secret is never copied into a longer-lived buffer.

## Testing (the gate)

New tests, all green; the full `freenet` lib suite passes (3301 passed, 0 failed):

- **Namespace isolation** (`two_users_have_disjoint_secret_namespaces`): tokens A and B store the *same* secret key under the same delegate; each reads back only its own value; removing A's leaves B's intact.
- **Local <-> User disjoint** (`local_and_user_namespaces_are_disjoint`): a `None`-context (Local) secret is invisible under any token and vice-versa, so flag-off single-user behavior is preserved and hosted connections never collide with the node's existing secrets.
- **The invariant** (`namespace_is_determined_solely_by_user_context`): identical delegate key + params + origin + message body, only `user_context` swapped => namespace swaps; the message body cannot reach another user's secrets. Proves the user id cannot be set by WASM / the client body.
- **Derivation pinned** (`user_id_is_a_stable_function_of_the_token`, `debug_redacts_the_dek_secret`).
- **Flag gating** (`derive_user_context` unit tests): flag off => token ignored; flag on => only a non-empty token derives a context.
- **Config** (`hosted_mode_defaults_to_off`, `hosted_mode_explicit_true_round_trips`): default false; round-trips through TOML when enabled.
- **Regression**: all pre-existing secret/delegate tests (91 secret-related) pass unchanged.

`cargo build -p freenet` ok, `cargo fmt` ok, `cargo clippy -p freenet --lib` clean.

## Out of scope (P2-frontend follow-up)

Browser shell-page JS (minting / storing / presenting the token in localStorage), the export button, and disclosure UI. Here the token arrives via the `userToken` channel and is exercised by tests.

Refs #4381

[AI-assisted - Claude]

### DIFF
--- a/apps/freenet-ping/app/tests/common/mod.rs
+++ b/apps/freenet-ping/app/tests/common/mod.rs
@@ -155,6 +155,7 @@ pub async fn base_node_test_config_with_rng<R: Rng>(
             token_cleanup_interval_seconds: None,
             allowed_host: None,
             allowed_source_cidrs: None,
+            hosted_mode: None,
         },
         network_api: NetworkArgs {
             // Use varied IP for network socket to get unique ring locations
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -22,6 +22,7 @@ use tokio::sync::mpsc;
 use crate::contract::{ClientResponsesReceiver, ContractHandlerEvent};
 use crate::message::{NodeEvent, QueryResult};
 use crate::node::OpManager;
+use crate::wasm_runtime::UserSecretContext;
 use crate::operations::{OpError, get, put, update};
 use crate::ring::KnownPeerKeyLocation;
 use crate::tracing::NetEventLog;
@@ -179,6 +180,12 @@ pub struct OpenRequest<'a> {
     pub notification_channel: Option<mpsc::Sender<HostResult>>,
     pub token: Option<AuthToken>,
     pub origin_contract: Option<ContractInstanceId>,
+    /// Per-connection per-user secret namespace (hosted mode, P2 of #4381),
+    /// derived once at the WS connection boundary from the connection's user
+    /// token. `None` outside hosted mode or when no token was presented. This
+    /// is carried alongside the request — never read from the request body —
+    /// so it cannot be forged by a client.
+    pub user_context: Option<UserSecretContext>,
 }
 
 impl Display for OpenRequest<'_> {
@@ -207,6 +214,7 @@ impl<'a> OpenRequest<'a> {
             notification_channel: None,
             token: None,
             origin_contract: None,
+            user_context: None,
         }
     }
 
@@ -222,6 +230,11 @@ impl<'a> OpenRequest<'a> {
 
     pub fn with_origin_contract(mut self, contract: Option<ContractInstanceId>) -> Self {
         self.origin_contract = contract;
+        self
+    }
+
+    pub fn with_user_context(mut self, user_context: Option<UserSecretContext>) -> Self {
+        self.user_context = user_context;
         self
     }
 }
@@ -1626,11 +1639,17 @@ async fn process_open_request(
                     );
                 }
                 let origin_contract = request.origin_contract;
+                // Per-connection user secret namespace (hosted mode). Taken from
+                // the OpenRequest, which received it from the connection layer —
+                // NOT from anything inside `req`. Moving it into the event keeps
+                // it on a channel the delegate/client cannot reach.
+                let user_context = request.user_context;
 
                 let res = match op_manager
                     .notify_contract_handler(ContractHandlerEvent::DelegateRequest {
                         req,
                         origin_contract,
+                        user_context,
                     })
                     .await
                 {
@@ -1987,6 +2006,7 @@ pub(crate) mod test {
                 notification_channel,
                 token: None,
                 origin_contract: None,
+                user_context: None,
             }
             .into_owned()
         }
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -22,10 +22,10 @@ use tokio::sync::mpsc;
 use crate::contract::{ClientResponsesReceiver, ContractHandlerEvent};
 use crate::message::{NodeEvent, QueryResult};
 use crate::node::OpManager;
-use crate::wasm_runtime::UserSecretContext;
 use crate::operations::{OpError, get, put, update};
 use crate::ring::KnownPeerKeyLocation;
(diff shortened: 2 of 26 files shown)


sanity:
## Multi-model review (Full tier — auth / state-isolation surface)

Reviewed by Codex (external, non-Claude) + two Claude adversarial lenses (auth-forgery, threading-completeness) + a maintainer code-first read. **The central unforgeability invariant holds** and all findings are addressed.

### Core design — verified sound
- **Unforgeable per-user identity**: `UserSecretContext` has exactly ONE constructor (`from_token`), called in exactly ONE place — the WS `connection_info` middleware, from the connection token. Nothing in a ClientRequest body, delegate WASM, delegate message, params, or contract id can construct or influence it. Pinned by `namespace_is_determined_solely_by_user_context`.
- **Complete threading**: the per-connection `Option<UserSecretContext>` rides *alongside* the request from the WS boundary → `DelegateCallEnv` → all 5 secret host fns via `env.secret_scope()`; no path hardcodes Local while a context exists. Inter-delegate hops, notifications, and local non-WS ops correctly pass `None` (namespace binds to the connection, not transitively).
- **Flag-off inert**: `derive_user_context(false, _) => None`; `hosted_mode` defaults false; `None → Local` everywhere → byte-for-byte today's single-user behavior.
- **No UAF**: `dek_secret` owned/`Zeroizing` in the env, borrowed per-call, never copied into a long-lived buffer; `Debug` redacts it.

### Findings → resolution
1. **MAJOR — credential logging** (Codex + both Claude lenses + me, unanimous): the `connection_info` debug log leaked the raw `userToken`, `auth_token`, and full `request_uri`. Since `userToken` derives a *durable, node-independent* DEK, log access = decrypt/forge that user's whole namespace forever. **Fixed** (bb5d829e): logs only `request_path` + non-secret presence flags; the misleading "never logged" comment replaced. (Also closed the pre-existing `auth_token` leak.)
2. **CI red — compile** (`WebsocketApiArgs` missing `hosted_mode` in a freenet-ping test initializer; the agent's lib-only test run didn't compile that target). **Fixed** (4357b3c9), verified with `cargo test --workspace --no-run`.
3. **`--hosted-mode` unusable as a bare flag** (Codex): `Option<bool>` required `--hosted-mode=true`. **Fixed** (e725c39e): `num_args=0..=1, default_missing_value="true"` keeps `Option<bool>` (config/env layering) while the bare flag works; parse test added.
4. **Inter-delegate namespace non-propagation** was asserted only by code+comment. **Test added** (53b691f0).

**Dismissed (false positive):** "`run_local_node` missing `Extension(HostedMode)`" — verified every bring-up path reaches the WS router via the single `serve_client_api_in_impl` choke point, which injects it on both router branches. The required (fail-fast) extractor is *kept* deliberately: for a security-isolation flag, 500-if-missing beats silently defaulting to hosted-off (which would disable isolation under public exposure).

**External coverage:** Codex ran as the external non-Claude pass (it independently caught both the token-logging MAJOR and the bare-flag issue). Gemini CLI remains unavailable (free-tier auth disabled), per the substitution recorded on #4502.

CI green on HEAD `e725c39e`. Merging.

[AI-assisted - Claude]

PR #4505, opened 2026-06-20
feat(hosted): refuse per-user token over insecure connections (#4381)

sanity:
## Problem

P2 (just merged) added hosted mode: a WebSocket connection presenting a durable `userToken` (query param `userToken` or header `X-Freenet-User-Token`) gets a per-user delegate-secret namespace scoped to that token. The token is a **high-value, durable, node-independent credential**. If an operator accidentally exposes the **plaintext** HTTP port to the internet, that token can be read off the wire and replayed, leaking access to a user's secret namespace.

## Approach

Gate honoring of the `userToken` behind a **secure-connection** check. A connection is "secure" when:

- **(a)** the source address is **loopback** (`127.0.0.0/8` / `::1`) — a direct local browser, or a TLS-terminating reverse proxy colocated on the same host; **OR**
- **(b)** it carries `X-Forwarded-Proto: https` **AND** the immediate source is **loopback** (a same-host TLS-terminating proxy forwarding the original scheme).

`X-Forwarded-Proto` is trusted **only** from a loopback source. From any off-host source it is client-spoofable (`curl -H 'X-Forwarded-Proto: https'`), so it never upgrades a remote plaintext connection to "secure". Because both honoring branches require loopback, the secure condition reduces to "loopback source" — but `xfp_https` is kept as an explicit input so the **"XFP-from-remote is ignored"** property is directly unit-tested and the trust rule is encoded at the boundary.

### Reject, not silent fallback

When hosted mode is on, a token is present, but the connection is **not** secure: the WS upgrade is **rejected** with `403 FORBIDDEN` and a clear message — `hosted user token requires a secure (TLS/loopback) connection`. Silently dropping the token and falling back to `Local` would put the user on the **wrong (shared)** namespace under a public plaintext deployment; rejecting **fails loud** so the operator/user notices the misconfiguration.

### Decision helper

The decision is a pure function, factored out of the `connection_info` middleware so the security property is provable without standing up a WS stack:

```rust
fn decide_user_token(
    hosted_mode: bool,
    has_token: bool,
    source_ip: Option<std::net::IpAddr>,
    xfp_https: bool,
) -> UserTokenDecision // Honor | RejectInsecure | Local
```

Called from `connection_info` (`crates/core/src/client_events/websocket.rs`). On `Honor` it derives the `UserSecretContext`; on `RejectInsecure` it returns 403; on `Local` it derives no context.

### Source IP / XFP plumbing

- **Source IP**: read from the `ConnectInfo<SocketAddr>` request extension already injected by `axum::serve(..).into_make_service_with_connect_info::<SocketAddr>()` (the same plumbing `private_network_filter` uses). Read from `req.extensions()` rather than as a typed extractor so a missing ConnectInfo (only possible in unit tests that omit it) yields `None` and **fails closed** (RejectInsecure) instead of 500-ing the request.
- **XFP**: read the `x-forwarded-proto` header, case-insensitive `== "https"`.

No TLS/cert handling is added to the binary — TLS is terminated externally; this is purely a gate on whether to honor the token.

## Fully hosted-mode-gated (no behavior change with the flag off)

Flag **OFF** is byte-for-byte unchanged: `decide_user_token` returns `Local` for every input when hosted mode is off, so no token is read for scope and the new rejection path never fires. Token-less connections are unaffected in all modes (no rejection — they're just `Local`).

## Testing

Pure-helper unit tests for every cell of the decision table:

- flag OFF + token + any source (incl. remote plaintext) ⇒ `Local`, never rejects (proves hosted-only gating)
- no token (any mode/source) ⇒ `Local`
- hosted + token + loopback (v4 / `127/8` / v6) ⇒ `Honor`, with and without XFP
- hosted + token + remote plaintext ⇒ `RejectInsecure`
- hosted + token + **LAN** (private-but-not-loopback) ⇒ `RejectInsecure` (trust anchor is loopback, not `is_private_ip`)
- hosted + token + remote + spoofed `X-Forwarded-Proto: https` ⇒ `RejectInsecure`
- hosted + token + missing source (`None`) ⇒ `RejectInsecure` (fail closed)
- `is_loopback_source` coverage incl. IPv4-mapped-v6 loopback

Plus an **integration-level** test (`middleware_rejects_insecure_hosted_token_and_passes_rest`) driving real requests through the `connection_info` middleware via `tower::ServiceExt::oneshot` with the production extension stack: asserts 403 for an insecure hosted token (incl. spoofed XFP), 200 for loopback honored, 200 for flag-off (token ignored), 200 for no-token.

Verified: `cargo build -p freenet`, `cargo test --workspace --no-run` (all targets compile), `cargo test -p freenet --lib client_events::websocket::tests` (57 passed), `cargo fmt`, `cargo clippy -p freenet --lib` clean.

Refs #4381

[AI-assisted - Claude]


sanity:
## Multi-model review (Full tier — auth / token-security surface)

Reviewed by Codex (external, non-Claude, **four** rounds as the gate was hardened) + two Claude adversarial lenses (auth-forgery, isolation/spoofing) + maintainer code-first reads. Gemini CLI remained unavailable (free-tier auth disabled — substitution recorded on #4502).

### What this enforces
In hosted mode, the durable per-user `userToken` is honored ONLY over a **loopback source carrying `X-Forwarded-Proto: https`**. Everything else (direct remote, non-loopback, no source, missing/`http` XFP) is rejected `403` — fail-closed. Hosted mode is opt-in; with it off the gate is inert and single-user behavior is byte-for-byte unchanged.

### The review hardened this gate four times — each a real Codex catch
1. **Empty-token regression**: an empty `?userToken=` (`Some("")`) was treated as a presented token → 403 over remote, locking out a brand-new user's first connect. Fixed: empty ⇒ absent ⇒ Local (unified with `derive_user_context`).
2. **Plaintext-proxy via loopback**: a same-host proxy serving public plaintext presents a loopback source → originally honored. Closed by bringing the request Host into the decision…
3. **…which was itself bypassable**: nginx's default rewrites Host to the upstream (`127.0.0.1`), so a public plaintext request looks "local." Host is proxy-controllable and cannot grant trust — dropped the Host branch entirely; require positive `X-Forwarded-Proto: https` evidence.
4. **XFP is client-spoofable through a pass-through proxy** (nginx forwards unknown client headers by default). No HTTP header is non-spoofable through a careless proxy — this is the classic `X-Forwarded-*` trust problem. **Resolution (decided): operator responsibility, documented**, not a shared-secret. The node trusts `X-Forwarded-Proto` from loopback; the operator MUST configure their proxy to set/overwrite it and strip client-supplied `X-Forwarded-*` (Caddy does by default; nginx needs `proxy_set_header`). Documented in the `--hosted-mode` CLI help and docs/secrets-at-rest.md, with an explicit security note that a pass-through misconfiguration can leak and the node can't detect it. The early operators are the Freenet team; a non-spoofable proxy-attestation (shared secret) can be layered on with P4 if wider operation warrants it.

### Verified solid
- Honor is unreachable off-host (source IP is kernel-set, non-spoofable; loopback proven before any header is consulted).
- Fail-closed on missing source IP; non-loopback rejected regardless of headers.
- Flag-off / no-token / empty-token paths are byte-for-byte the pre-#4381 Local behavior.
- Token never logged; `dek_secret` `Zeroizing` + Debug-redacted (carried from P2).

CI green on HEAD `8cb6a77f` (doc-only commit on top of the reviewed `f5bda7f8` gate). Merging.

[AI-assisted - Claude]

Issue #4035, opened 2026-05-05
feat(delegates): fdev CLI to export/import a delegate's secrets bundle


sanity:
## Problem
Delegate secrets currently can only be backed up by hand-copying three things together: `delegates/`, `secrets/`, and the `--cipher` key file. Lose any one and the backup is dead. There is no user-friendly way to take a snapshot of a delegate's state to a USB drive or another machine.

PR #4034 landed local snapshot-on-write history (defends against accidental overwrite) but explicitly deferred this CLI as a separate effort.

## Why bundling the WASM matters

Delegates are not directly retrievable from the Freenet network — they ride along inside a webapp's install bundle. So the realistic recovery scenario is: user re-installs the webapp on a new machine, the webapp ships a *newer* version of its delegate, the new delegate has a different `DelegateKey` (because `DelegateKey = BLAKE3(BLAKE3(wasm) ‖ params)`), and the user's old secrets are now filed under a key the system doesn't know about.

There are two recovery paths:
1. The webapp author maintains a legacy-delegate migration table (River's `legacy_delegates.toml` pattern) that maps old keys to the current key. This requires per-upgrade cooperation from the webapp author and is brittle.
2. The backup itself ships the old WASM + old params, so the old `DelegateKey` is fully reconstructible without webapp-author involvement.

Path (2) is self-contained, which is what users actually want from a backup. **Bundle the WASM by default.** Hundreds of KB to a couple MB compressed is small relative to typical secret state. Add `--exclude-wasm` for users who know they'll re-install the same webapp version and want minimal bundle size.

## Proposal
Two `fdev` subcommands:

```
fdev delegate export <delegate_key> --output backup.tar.zst.age
fdev delegate import backup.tar.zst.age
```

Bundle contents:
- `manifest.toml` (delegate key, code_hash, cipher fingerprint, schema version)
- `wasm/{code_hash}.wasm` (delegate bytecode, versioned; default-on, `--exclude-wasm` to skip)
- `delegate.reg` (params + code_hash → DelegateKey identity)
- `secrets/{secret_id}` (active ciphertext per secret)
- `snapshots/{secret_id}/{epoch_ms}` (full snapshot history from PR #4034)
- `cipher.key` (or passphrase-derived via Argon2id)

Import path:
1. Verify cipher fingerprint matches expected.
2. Install WASM into the delegate store (skip if `code_hash` already present locally).
3. Install `.reg` to lock in the historical `DelegateKey`.
[...]

PR #4506, opened 2026-06-20
feat(secrets): generic delegate-secret export/import bundle (P3 of #4381)

sanity:
## What this is

**P3 of #4381** (hosted public proxy): generic, app-agnostic **delegate-secret export / import**, so a user can download their private delegate data from a hosted node and re-import it into their **own** local peer — the "try Freenet hosted, then go fully decentralized" migration. Builds on P1 (#4502, `SecretScope`) + P2 (#4504, per-user token threading).

`Refs #4035` — this does **NOT** close it. #4035's headline ask is bundling the delegate **WASM** for self-contained recovery when a re-install ships a *newer* delegate (different `DelegateKey`). This PR bundles **secrets only** (re-placed under their original `DelegateKey`), which covers the hosted→self-host migration (same delegate) and local backup, but not the cross-version WASM-reconstruction case. The bundle records `code_hash`, so a WASM-attaching follow-up can layer on; #4035 stays open for that.

## Bundle format (versioned, self-describing, always encrypted at rest)

`[MAGIC "FNSX"][format_ver u8][kdf_id u8][16B salt][24B nonce][AEAD(CBOR payload)]`

- Header (magic/ver/kdf_id/salt/nonce) is bound as the AEAD **AAD**, so tampering with it fails authentication cleanly.
- AEAD = **XChaCha20-Poly1305**; nonce random per export.
- Bundle key = **Argon2id**(passphrase, salt) *or* **HKDF-SHA256**(token) with a distinct domain; `kdf_id` records which so import picks the right KDF.
- CBOR payload: `{schema_version, source_scope, created_unix_secs, entries:[{delegate_key, code_hash, secret_hash, plaintext}]}`.
- Plaintext + derived keys held in `Zeroizing`; the file is **never** written containing plaintext; output created `0o600`.

## Keying decision

Export/import are keyed on the 32-byte **secret hash** (the only pre-image the store persists — the on-disk name is `bs58(hash)`), not a reconstructed `SecretsId`. Sufficient because `store_secret`/`get_secret`/`remove` only use `key.encode()`/`key.hash()`. `DelegateKey` is reconstructed deterministically on import (content-derived), so a re-installed webapp's same-version delegate lines up.

## API + CLI

- Module `wasm_runtime/secret_export.rs`: `export_bundle` / `import_bundle` / `BundleKeyMaterial::{Passphrase, Token}` / `TargetScope::{Local, User}` / `write_bundle_file` / `ImportReport{imported, skipped}`.
- Store primitives: `SecretsStore::{export_scope_entries, import_secret_by_hash}`, `ExportSecretEntry`; plus `Secrets::load_for_secrets_dir` so the CLI opens a store without a full `Config`.
- CLI (`freenet secrets`): `export --secrets-dir --db-dir {--local|--user-token} {--passphrase|--use-token-key} --out` (refuses to clobber); `import <file> {--passphrase|--token} [--local|--into-user] [--overwrite]`. Node must be stopped. Export prints the **operator-sees-plaintext-during-export** disclosure on stderr (matches #4381's honest framing).

## Tests

Round-trip (passphrase + token), wrong-key clean auth-fail (no panic / no partial write), tamper-detect, encrypted-at-rest (file bytes don't contain a known plaintext), collision handling (refuse vs `--overwrite`), Local + User round-trips. Lib 53 + CLI 31 pass; `cargo test --workspace --no-run` compiles all targets; `cargo clippy --locked -- -D warnings` exit 0; fmt clean; live CLI round-trip smoke-tested.

Pending Full multi-model review (crypto surface).

[AI-assisted - Claude]


sanity:
## Multi-model review (Full tier — crypto / secrets surface)

Reviewed by Codex (external, non-Claude; multiple rounds) + a Claude crypto adversarial lens + maintainer code-first reads. Gemini CLI unavailable (recorded on #4502). Codex's final pass on the current HEAD: "No actionable correctness issues were identified."

### What it is
Generic, app-agnostic delegate-secret **export/import** (`freenet secrets export|import`) — download a user's (or Local) secrets as one encrypted bundle and re-import them on your own peer. The hosted→self-host migration for #4381. `Refs #4035` (does not close it — see that issue for the remaining WASM-bundling scope).

### Verified solid (crypto)
- Envelope `[MAGIC FNSX][ver][kdf_id][16B salt][24B nonce][AEAD(CBOR)]`, header bound as AEAD **AAD** — the crypto lens flipped all bundle bytes one-by-one; every flip across header AND ciphertext rejected with auth failure.
- XChaCha20-Poly1305, per-export random salt+nonce (OsRng); Argon2id (OWASP-floor params) for the passphrase key, HKDF-SHA256 for the token key; all five domain strings in the crate distinct (no collision with the P1/P2 storage DEKs). `kdf_id` is in the AAD.
- Wrong key / tamper / truncated / bad-magic ⇒ clean typed errors, no panic, no partial write before the AEAD check. Hash-only keying is sound (matches the on-disk `bs58(hash)` name); DelegateKey reconstructed deterministically; Local/User scope routing correct.

### Findings → all addressed
1. **Plaintext not zeroized** (Codex + lens): `BundleEntry.plaintext` was a plain `Vec`. Fixed — manual `Drop` zeroizes it, covering both export-build and import-deserialize paths.
2. **Silent empty backup** (Codex): export with a mistyped `--db-dir` made an empty DB ⇒ empty bundle. Fixed — export refuses a missing node DB (import may still create fresh). Tested.
3. **Import retry didn't converge** (Codex): a write-then-index partial failure left a secret on disk but unindexed; retry-without-overwrite skipped it forever (silent data loss on the next migration). Fixed — the skip path now reconciles the index (idempotent); + `add_to_index` early-returns when already-indexed (behaviour-preserving for the Local path; index set + values byte-identical). Regression tests added.
4. **Secrets on argv** (Codex): `--passphrase`/`--token` leaked via `ps`/shell history. Fixed — a shared resolver: env var → non-empty flag (documented insecure) → hidden TTY prompt (rpassword); `Zeroizing<String>`; clear error when non-interactive with no source.
5. Doc-overclaim on import atomicity reworded (decrypt all-or-nothing; per-entry idempotent on retry); distinct `EncryptFailed` variant; tamper test added; garbled comment fixed.

### Honest caveat (documented)
A server-side export decrypts secrets to build the bundle, so the operator sees plaintext during the operation — surfaced in the CLI disclosure and docs, matching #4381's framing.

CI green on HEAD `b8f558ba`. Merging.

[AI-assisted - Claude]

PR #4511, opened 2026-06-20
feat(p2p): re-enable SubscribeHint placement migration at floor (0,2,80) (#4404)

sanity:
## Problem

The SubscribeHint placement migration (#4404) was disabled in v0.2.74 after its v0.2.73 deployment caused a network-wide UPDATE-broadcast degradation by amplifying the latent #4145 event-loop wedge. It has been parked at floor `(0, 3, 0)` (above all shipped versions) ever since.

The preconditions to re-enable are now met:
- **#4145 root fixed** (#4499): the blocking event-loop sends are non-blocking, so the wedge can't form under fan-out. Validated at 40-80-subscriber fan-out (converges, no connectivity loss, bounded backlog).
- **Telemetry deployed** (#4498, shipped in 0.2.79): the broadcast-assembly-failure gauge is collecting a clean baseline on the live network (~0 failures per 5-min window vs the incident's ~125-190).

## Solution

Lower `SUBSCRIBE_HINT_MIN_VERSION` from `(0, 3, 0)` to **`(0, 2, 80)`**, the first release that ships SubscribeHint with the #4145 fix. The gate sends a hint only to peers on `>= 0.2.80`, so the migration activates **gateway-first** (gateways auto-update first) and **ramps with fleet upgrade** rather than switching on everywhere at once. This is the staged, version-gated canary; rollback is re-parking the floor in a hotfix release.

Also flips the two **deactivation guard tests** to the re-enabled semantics (`placement_migration_active_at_reenable_floor`, `receive_gate_active_at_reenable_floor`). These previously asserted the migration is off for all 0.2.x and would **fail in the release build** once the crate crosses the floor at 0.2.80. Updates `docs/RELEASING.md` (parked -> frozen-at-first-shipping-version) and the stale "parked/deactivated" doc comments, receive-gate log message, and one source-pin assertion.

## Testing

- Gate unit tests pass at a temp 0.2.80 bump (8/8), including both rewrites.
- **Full simulation suite verified at a temp 0.2.80 bump** (migration cascade **on** by default, exactly as the release's merge_group CI will run it): `streaming_e2e` + `simulation_integration` = **123/123 passed**, including `test_streaming_get_retries_after_assembly_failure` (the most placement-sensitive). The #4435 placement-robustness hardening holds after the #4145 fix and the #4507 refactor.
- `cargo clippy --locked -p freenet -- -D warnings` clean.

Note: this PR keeps the crate at 0.2.79, so its own CI runs with the migration off (0.2.79 < floor); the cascade-on behavior is exercised by the release build (crate 0.2.80) and was pre-verified above.

Re-enables #4404. Tracking: #4440.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -3371,9 +3371,8 @@ mod tests {
         /// 0.2.79 until the re-enable release bumps it to the floor version).
         #[test]
         fn receive_gate_active_at_reenable_floor() {
-            // Pin the re-enable floor value: an accidental change here moves the
-            // migration's activation set.
-            assert_eq!(SUBSCRIBE_HINT_MIN_VERSION, (0, 2, 80));
+            // The `supported(0,2,80)` + `!supported(0,2,79)` pair pins the floor
+            // to exactly `(0, 2, 80)`; an accidental change trips these asserts.
             // Peers at or above the floor are acted on.
             assert!(version_supports_subscribe_hint(
                 Some((0, 2, 80)),
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1755,23 +1755,19 @@ where
             return Ok(());
         }
         NetMessageV1::SubscribeHint(hint) => {
-            // Placement-migration deactivation gate (v0.2.74 hotfix). The
-            // placement migration was disabled after the v0.2.73 UPDATE-broadcast
-            // degradation by parking the SubscribeHint version floor at
-            // `(0, 3, 0)`, above every shipped 0.2.x release. The SEND side
-            // (`p2p_protoc::peer_supports_subscribe_hint`) already respects this
-            // floor, but the floor bump alone only stops US from emitting hints —
-            // a 0.2.74 node receiving a hint from a not-yet-upgraded 0.2.73 peer
-            // would still ACT on it and keep the migration load alive throughout
-            // the multi-hour staggered rollout.
+            // Placement-migration version gate. The migration is re-enabled at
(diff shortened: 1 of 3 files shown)


sanity:
### Review dispositions

**External (Codex) review** flagged one [P1]:
- *"Gate inbound hints on the sender version"* (node.rs receive gate checks own version, not the sender's). **Disposition: addressed as a doc clarification; the gate is correct as-is.** The load-bearing node is the *receiver* (it does the directed subscribe, becomes a host, takes on the broadcast work). The own-version gate guarantees the receiver is `>= 0.2.80`, i.e. always has the #4145 load-safe fix; the sender's version is irrelevant to the receiver's load-safety. In practice only `>= 0.2.80` nodes emit hints (0.2.74–0.2.79 have parked floors), and the only pre-floor *sender* would be a 0.2.73 node, which is effectively extinct. The original v0.2.74 design deliberately gated on own-version and documented that the per-connection sender version is not exposed at the receive handler. Updated RELEASING.md to describe the gate accurately rather than overstate "both >= 0.2.80".

**Rule-review (CI) warning:** the two deactivation guard tests are now kept as `#[ignore]`'d "superseded" tests (per the repo test convention) alongside the new active-at-floor tests, rather than replaced.

**Release-build verification:** the full simulation suite passes at a temp 0.2.80 bump (migration cascade on by default, as the release's merge_group CI runs it): 123/123, including `test_streaming_get_retries_after_assembly_failure`.

[AI-assisted - Claude]

PR #4513, opened 2026-06-20
feat(hosted): shell-page per-user token (P2-frontend of #4381)

sanity:
## Problem

P2 of #4381 added a `HostedMode` flag and taught the WebSocket gate to honor a `userToken` query parameter (per-user delegate-secret namespace). Nothing on the browser side ever minted or presented that token, so a hosted node could not actually distinguish users — every visitor fell through to the Local (shared) namespace.

## Approach

The node-served shell page wraps the sandboxed app iframe and proxies all WebSocket connections, injecting credentials at the single WS-open point. The shell is same-origin with the node, so it can use `localStorage`; the sandboxed iframe (opaque origin) cannot. That makes the shell the right place to mint and hold a durable per-user secret.

In **hosted mode** the shell now:

- **Mints (or loads) a durable token** in `localStorage` under `__freenet_user_token__`: 32 bytes from `crypto.getRandomValues`, hex-encoded. One token = one identity per visitor across every contract app on this node (the shell origin is identical for all apps) — the intended design, deliberately not scoped per-contract.
- **Presents it** via `freenetBridge(authToken, userToken)`, which appends `?userToken=<token>` to the real WebSocket URL right after `authToken`. The param name matches the backend's canonical source exactly (`websocket.rs` `user_token_q`).

The token is generated client-side from OS entropy and is **never** derived from the URL/query or any request input, so there is no injection vector. If `localStorage` is unavailable (private-mode quirks, embedded webviews) the snippet degrades to an `undefined` token; the bridge omits the param and the backend treats the connection as Local — matching the empty-token fall-through the gate already implements.

When **hosted mode is OFF** the shell is **behaviourally identical** to the pre-#4381 shell: no token snippet, the original single-argument `freenetBridge(...)` call at the same site. (This is behavioural, not literal byte-equality — the always-injected `SHELL_BRIDGE_JS` itself gained a `userToken` argument and an inert, undefined-guarded `if (userToken)` branch that never fires when the bridge is called with one argument.) `hosted_mode` is threaded from the **real** `Extension<HostedMode>` on the contract-web route (injected on both router branches — loopback and LAN — in `serve_client_api_in_impl`); it is never defaulted to false on this path.

**Same-origin isolation (Codex review fix, commit `a40416d1`):** the bridge's WS-open handler unconditionally `delete('userToken')` and `delete('authToken')` on the caller-supplied WS URL **before** injecting the shell's own credentials. Without the `userToken` delete, a contract app (or a value reflected via a deep-link into the WS URL) could keep its own `userToken` whenever the shell's minted token is `undefined` (localStorage disabled / private mode), choosing its own per-user secret namespace. The app must never influence the namespace — only the shell-minted token may reach the backend. Pinned by `bridge_js_strips_caller_supplied_user_token_before_injecting`, which asserts both deletes precede the (conditional) injection points.

## Insertion points

- `crates/core/src/server/path_handlers.rs`
  - `shell_page(..)` gains a `hosted_mode: bool` param; conditionally emits the `SHELL_USER_TOKEN_JS` `<script>` and a 2-arg vs 1-arg `freenetBridge(...)` call.
  - New `SHELL_USER_TOKEN_JS` constant (localStorage mint/load from `crypto.getRandomValues`).
  - `freenetBridge(authToken)` → `freenetBridge(authToken, userToken)`; WS-open handler appends `userToken` right after `authToken` (`if (userToken) u.searchParams.set('userToken', userToken)`).
  - `contract_home(..)` gains `hosted_mode: bool`, forwarded to `shell_page`.
- `crates/core/src/server/client_api.rs` — `render_shell_response`, `web_home`, `web_subpages` thread `hosted_mode` through to `contract_home`.
- `crates/core/src/server/client_api/v1.rs` + `v2.rs` — `web_home_v{1,2}` / `web_subpages_v{1,2}` extract `Extension<crate::server::HostedMode>` and pass `hosted_mode.0`.

## How HostedMode reaches this route

`web_home`/`web_subpages` live in `gw_router`; `WebSocketProxy::create_router_with_origin_contracts(gw_router, ..)` merges that into the returned `ws_router`, and `serve_client_api_in_impl` applies `.layer(Extension(hosted_mode))` to that merged router on **both** the LAN and loopback branches. So the contract-web/home route always sees the real flag — no defaulting required.

## Testing

- `shell_page_hosted_mode_injects_user_token` — hosted output contains the localStorage minting snippet, `crypto.getRandomValues`, and the 2-arg `freenetBridge(..., __freenet_user_token)` call; the 1-arg call is absent.
- `shell_page_non_hosted_mode_omits_user_token` — no minting snippet, no `localStorage.setItem`, no 2-arg call; the original 1-arg call is present (no-regression guard).
- `bridge_js_appends_user_token_param` — pins the bridge signature, the `userToken` searchParams append, and OS-entropy minting in `SHELL_USER_TOKEN_JS`.
- `bridge_js_strips_caller_supplied_user_token_before_injecting` — isolation-boundary regression: both `delete('userToken')` and `delete('authToken')` exist and precede the (conditional) credential injection, so a caller-supplied token can never survive — including the undefined-token path.
- All existing `path_handlers` (now 75) + `client_api` (81) lib tests stay green, including the cross-contract `authToken`-injection regression tests. `cargo clippy -p freenet --lib` clean; `cargo test --workspace --no-run` compiles.

Refs #4381

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -189,6 +189,7 @@ async fn web_home(
     req_headers: axum::http::HeaderMap,
     api_version: ApiVersion,
     query_string: Option<String>,
+    hosted_mode: bool,
 ) -> Result<axum::response::Response, WebSocketApiError> {
     // Check if this is the sandboxed iframe requesting its content
     let is_sandbox = query_string
@@ -201,7 +202,16 @@ async fn web_home(
     }
 
     // Root document load: render the shell that wraps the contract root.
-    render_shell_response(key, &config, api_version, query_string, None, rs).await
+    render_shell_response(
+        key,
+        &config,
+        api_version,
+        query_string,
+        None,
+        rs,
+        hosted_mode,
+    )
+    .await
 }
 
 /// Generates a shell page response: a fresh auth token + secure cookie,
@@ -222,6 +232,7 @@ async fn render_shell_response(
     query_string: Option<String>,
     sub_path: Option<&str>,
     rs: HttpClientApiRequest,
+    hosted_mode: bool,
 ) -> Result<axum::response::Response, WebSocketApiError> {
     use headers::{Header, HeaderMapExt};
 
@@ -241,9 +252,16 @@ async fn render_shell_response(
         .build();
 
     let token_header = headers::Authorization::bearer(token.as_str()).unwrap();
-    let contract_response =
-        path_handlers::contract_home(key, rs, token.clone(), api_version, query_string, sub_path)
-            .await?;
+    let contract_response = path_handlers::contract_home(
+        key,
+        rs,
+        token.clone(),
+        api_version,
+        query_string,
+        sub_path,
+        hosted_mode,
+    )
+    .await?;
 
     let mut response = contract_response.into_response();
     response.headers_mut().typed_insert(token_header);
@@ -277,6 +295,7 @@ async fn render_shell_response(
     Ok(response)
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn web_subpages(
     key: String,
     last_path: String,
@@ -285,6 +304,7 @@ async fn web_subpages(
     req_headers: axum::http::HeaderMap,
     config: &Config,
     request_sender: HttpClientApiRequest,
+    hosted_mode: bool,
 ) -> Result<axum::response::Response, WebSocketApiError> {
     let is_sandbox = query_string
         .as_ref()
@@ -337,6 +357,7 @@ async fn web_subpages(
             query_string,
             Some(&last_path),
             request_sender,
+            hosted_mode,
         )
         .await;
     }
@@ -935,6 +956,7 @@ mod tests {
                     headers,
                     &localhost_config(),
                     sender,
+                    false,
                 )
                 .await
                 .map(|_| ())
@@ -968,6 +990,7 @@ mod tests {
             headers,
             &localhost_config(),
             dead_request_sender(),
+            false,
         )
         .await
         .expect("sandbox document load must redirect, not error");
@@ -987,6 +1010,7 @@ mod tests {
             axum::http::HeaderMap::new(),
             &localhost_config(),
             dead_request_sender(),
+            false,
         )
         .await;
         match res {
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -295,6 +295,25 @@ async fn render_shell_response(
     Ok(response)
 }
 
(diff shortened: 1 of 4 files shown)


sanity:
## Multi-model review (credential-handling surface)

Reviewed by Codex (external, non-Claude — **five** rounds as the credential handling was hardened) + a Claude frontend-security lens (exfiltration) + maintainer reads. Final Codex pass on this HEAD: "no discrete correctness, security, or maintainability issues; the non-hosted activation path is unchanged."

### What it does
In hosted mode the node-served shell page mints a durable per-user token (`crypto.getRandomValues`, localStorage), and the shell — which proxies every WebSocket on the sandboxed app's behalf — appends it as `?userToken=` to the real WS URL. One identity per visitor (shared node-origin localStorage). The WS security gate (websocket.rs) is **unchanged** — `git diff vs main` on it is empty; this PR is purely the browser plumbing.

### The review hardened the credential handling five times (each a real Codex catch)
1. **Caller-supplied token not stripped**: a contract app could inject its own `?userToken=` on its WS URL (and on the undefined-token path it survived). Fixed: `delete('userToken')`/`delete('authToken')` before the shell-controlled set — only the shell-minted token can reach the backend.
2. **`as_router` regression**: the new required `Extension<HostedMode>` on the contract-web routes 500'd the public `HttpClientApi::as_router` composition path. Fixed: shell route reads HostedMode **tolerantly** (absent ⇒ hosted-off). The WS *enforcement* gate stays **required/fail-loud** — deliberate asymmetry (enforcement vs UX point).
3. **Token transmitted over plaintext**: on http, the shell would still put the durable token on a `ws://` URL before the backend rejects it. Fixed: never mint/attach the token unless the page is `https:` (two independent guards).
4. **Silent shared-Local on http**: withholding the token then connecting anyway mapped hosted+no-token to the shared Local namespace = cross-user contamination. Fixed: **fail closed** (banner + refuse to connect).
5. **Fail-closed was incomplete**: hosted+https with localStorage/crypto failure also yields no token ⇒ shared Local. Fixed: unified to **`hosted && !userToken` ⇒ fail closed** (covers every no-token path). A regression test asserts the guard no longer keys on protocol.

### Net security posture
A hosted browser operates **iff** it holds its own per-user token — which requires https + working storage; otherwise it refuses (no token transmitted, no silent shared-Local). Frontend-security lens confirmed the token can't be exfiltrated by the sandboxed app (never echoed back; cross-origin WS-open blocked before append). Non-hosted single-user behavior is unchanged (the bridge gains only inert, undefined-guarded args).

CI green on HEAD `f271f592`. Merging.

[AI-assisted - Claude]

PR #4514, opened 2026-06-20
test(hosted): live end-to-end per-user isolation + gate (#4381)

sanity:
## Problem

The hosted-mode backend for #4381 (P1 `SecretScope`, P2 `--hosted-mode` + per-user `userToken` threading, the refuse-plaintext-token invariant, P3 export/import) had unit coverage for the pieces, but no test proving the whole thing works end-to-end on a real running node — that a `userToken` presented on a live WebSocket upgrade actually lands a delegate's `set_secret`/`get_secret` calls in a per-user namespace, and that the security gate fires over the real HTTP/WS stack.

## Approach

`crates/core/tests/hosted_mode_isolation.rs` boots a REAL single node (full event loop, contract executor, WASM runtime) with `hosted_mode = true` and drives it through the real `tokio-tungstenite` WS handshake, so it exercises the actual `connection_info` gate middleware and the per-connection `userToken` threading — not just the store layer.

It reuses the in-process node-boot pattern from `tests/in_process_restart.rs` (`NodeConfig::new(cfg).build(serve_client_api(cfg.ws_api))` + `tokio::spawn(node.run())`), driving the secret-capable `tests/test-delegate-2` fixture (`StoreSecret`/`GetNonExistentSecret`, which call `ctx.set_secret`/`ctx.get_secret`). Readiness is awaited (the WS API accepting a connection; each response arriving) via bounded `connect`/`recv` loops with clear failure messages — no fixed sleeps gate correctness.

The `userToken` is set as a query parameter on the upgrade URL; `X-Forwarded-Proto: https` is set as a handshake header via `IntoClientRequest`. The loopback source the gate requires is automatic (we connect to `127.0.0.1`).

## What it proves

Three tests, all over the live stack against a real executor:

1. **Per-user isolation** (`hosted_mode_per_user_secret_isolation`) — two hosted connections with DISTINCT `userToken`s each store a different value under the SAME delegate+key, and each reads back only its OWN value. A no-token connection on the same node stores/reads on the `Local` namespace, invisible to the two per-user scopes (and vice-versa). Cross-`assert_ne!` checks confirm the three values are genuinely distinct, so the equalities prove non-overlap rather than coincidence. A reconnect re-presenting token A sees A's durable value, proving the namespace is keyed only by the token.

2. **Gate enforcement** (`hosted_mode_rejects_token_without_https`) — a hosted connection presenting a `userToken` but WITHOUT `X-Forwarded-Proto: https` is rejected at the WS upgrade with **403** (`RejectInsecure`). A sanity follow-up shows the SAME token WITH the XFP header is accepted, so the rejection is the gate firing on the missing header, not a broken endpoint.

3. **Flag gate** (`flag_off_ignores_user_token_and_uses_local`) — on a node with `hosted_mode = false`, the same `userToken` (+ XFP) is IGNORED: the tokened write is visible to a plain no-token connection on the shared `Local` namespace, proving the whole feature is inert unless the flag is on.

## Testing

- `cargo test -p freenet --test hosted_mode_isolation` — 3 passed, run 3x (16.3s first run incl. one-time fixture WASM compile, then ~1.4-2.0s).
- `cargo fmt --check` clean; `cargo clippy -p freenet --test hosted_mode_isolation -- -D warnings` clean.
- `cargo test --workspace --no-run` compiles.

Refs #4381. Draft — not for merge.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/tests/hosted_mode_isolation.rs
+++ b/crates/core/tests/hosted_mode_isolation.rs
@@ -1,0 +1,551 @@
+//! Live end-to-end integration test for hosted mode (#4381).
+//!
+//! This boots a REAL single node — full event loop, contract executor, and
+//! WASM runtime — with `hosted_mode = true`, and drives it through the real
+//! HTTP/WS stack with `tokio-tungstenite` clients. It exercises the actual
+//! `connection_info` gate middleware and the per-connection `userToken`
+//! threading, not just the store layer, so it proves the feature works as a
+//! whole rather than in isolation.
+//!
+//! What it proves (one delegate, `test-delegate-2`, which stores/reads secrets
+//! via `ctx.set_secret`/`ctx.get_secret`):
+//!
+//! 1. ISOLATION — two hosted connections presenting DISTINCT `userToken`s store
+//!    a secret under the SAME delegate+key and each reads back its OWN value.
+//!    The per-user namespace (P1 `SecretScope::User`) keeps them apart.
+//! 2. NO-TOKEN = LOCAL — a hosted connection with no `userToken` operates on the
+//!    `SecretScope::Local` namespace, invisible to (and not seeing) the per-user
+//!    secrets above.
+//! 3. GATE ENFORCEMENT — a hosted connection presenting a `userToken` but WITHOUT
+//!    `X-Forwarded-Proto: https` is rejected at the WS upgrade with 403
+//!    (the refuse-plaintext-token invariant).
+//! 4. FLAG GATE — on a node with `hosted_mode = false`, the same `userToken`
+//!    (with the XFP header) is IGNORED: the connection lands on `Local`, proving
+//!    the whole feature is inert unless the flag is on.
+//!
+//! Determinism: the node is driven on its own task and we wait on real readiness
+//! (the WS API accepting a connection, each response arriving) via bounded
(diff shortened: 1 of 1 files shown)


sanity:
## Review (Light tier — test-only, no production code)

External: Codex (`codex review --base main`) — "no discrete correctness issues in the added test code." Maintainer read confirmed the assertions are non-vacuous:
- **Isolation** writes a *distinct* value under TOKEN_A / TOKEN_B / no-token to the *same* delegate+key, then asserts each reads back its *own* value AND all three differ (`assert_ne!`) AND a reconnect with TOKEN_A still sees A's value. If isolation broke (shared namespace), last-writer-wins would equalize the reads and the test would fail — so it genuinely exercises the property.
- **Gate**: token + no `X-Forwarded-Proto: https` → 403, with a same-token-WITH-XFP sanity connect so the 403 is the gate firing, not a dead endpoint.
- **Flag**: `hosted_mode=false` → token ignored (Local).

Drives a real booted node through the actual HTTP/WS stack (the `connection_info` gate + userToken threading), reusing the in-process harness + the `test-delegate-2` secret-roundtrip fixture. 3 deterministic runs pass. CI green on HEAD.

This is the end-to-end 'it works on a live node' proof for #4381. Merging.

[AI-assisted - Claude]

PR #3793, opened 2026-04-07
fix: summary mismatch broadcasts to all peers instead of just the stale peer

sanity:
## Problem

When a peer reports a stale summary during interest sync, the node emits `BroadcastStateChange` which fans out to **all** subscribers (~28 peers), not just the specific peer that reported the mismatch. With N peers exchanging summaries every 5 minutes, this creates O(N * fan-out) redundant transmissions per cycle.

### Production evidence (gateway, 2026-04-07 hour 18 UTC)

- **4,820 UPDATE_PROPAGATION events**, of which **4,257 (88%) were triggered by summary mismatches**
- Average fan-out: 28.7 peers per broadcast
- Peak: 144 broadcasts/minute for the River room contract alone
- Total peer-transmissions: ~138,000/hour from the gateway
- Technic node: 581 MB uploaded vs 31 MB downloaded in ~1 hour (19:1 ratio)
- User report (Lukas Orsvarn): 163:1 upload/download ratio, 1.1 GB upload in 45 minutes

### Root cause

`node.rs` summary mismatch handler emits `NodeEvent::BroadcastStateChange { key, new_state }`, which calls `get_broadcast_targets_update` and sends to ALL subscribers/hosts. The comment claimed "sends state only to peers with stale summaries" but the production path broadcasts to everyone.

## Approach

Add a new `NodeEvent::SyncStateToPeer { key, new_state, target }` variant that sends state to only the specific peer that reported the stale summary. This replaces `BroadcastStateChange` in two places:

1. **Interest sync summary mismatch** (`node.rs:1418`): When a peer reports a stale summary, sync only to that peer
2. **Proximity cache overlap** (`node.rs:1091`): When a neighbor announces shared contracts, sync only to that neighbor

The `BroadcastStateChange` path (used for actual contract state changes from PUT/UPDATE operations) is unchanged.

### Expected impact

- Upload traffic from summary-mismatch syncs drops from O(N * fan-out) to O(N) per heartbeat cycle
- For the gateway with ~39 connected peers and fan-out of ~28: from ~138,000 to ~4,257 peer-transmissions/hour (97% reduction)
- Normal contract update broadcasts (from PUT/UPDATE) are unaffected

## Testing

- New unit test `test_summary_mismatch_targets_only_stale_peer`: Verifies that when 3 peers subscribe to a contract and only 1 has a stale summary, only that 1 peer needs syncing (not all 3)
[...]

PR #4520, opened 2026-06-20
perf(ring): cap interested-peers per contract to bound broadcast fan-out (#3798 Gap 2)

iduartgomez:
## Problem

`register_peer_interest` (`crates/core/src/ring/interest.rs`) inserted into the per-contract `interested_peers` map with **no size check** — unlike `MAX_DOWNSTREAM_SUBSCRIBERS_PER_CONTRACT = 512` in `hosting.rs`, which does bound. Since the interest map is a `BroadcastStateChange` fan-out source, an adversary with many IPs could inflate it and multiply the fan-out cost of every subsequent broadcast for that contract — a write-amplification vector (#3798 Gap 2, surfaced by the PR #3793 audit).

## Solution

Cap `interested_peers` per contract, mirroring the existing `hosting.rs` pattern exactly:
- Bound at **insertion**; reject a **new** peer on overflow (return `false`) with a warn log so the caller knows registration was rejected.
- **Renewals are never rejected** — an already-tracked peer refreshes in place.
- The `return false` sits **before** the reverse-index / hash-index writes, so a rejected peer leaves no orphan index state.
- Cap check stays inside the existing `interested_peers` shard guard, preserving the cross-DashMap lock discipline from #4129/#4171.

Gap 1 (`MAX_STALE_SYNCS_PER_SUMMARIES`) is already on `main`. Gap 3 (proximity sync loop) is deferred as a follow-up under the tracking issue.

## Testing

`test_register_peer_interest_caps_per_contract_fanout` — fills to exactly cap (all admitted), asserts cap+1 new peer returns false AND is not inserted AND set size unchanged, asserts a renewal at cap stays tracked. Plus a cap-hit log/metric path test. Fails if the cap is removed or off-by-one. fmt + touched-file clippy clean.

## Fixes

Addresses Gap 2 of #3798 (tracking issue stays open for Gap 3).

🤖 Reviewed by an adversarial xhigh-effort agent pass (verdict: ship) before submission.

### DIFF
--- a/crates/core/src/ring/interest.rs
+++ b/crates/core/src/ring/interest.rs
@@ -81,6 +81,24 @@ use crate::config::GlobalRng;
 /// Maximum number of entries in the delta memoization cache.
 const DELTA_CACHE_SIZE: usize = 1024;
 
+/// Maximum number of interested peers tracked per contract (#3798 Gap 2,
+/// anti-amplification hardening).
+///
+/// `interested_peers[contract]` is the per-contract fan-out set: every
+/// `BroadcastStateChange` for a contract pays O(N) cost in this map's size, so
+/// an adversary controlling many IPs/keys could inflate it without bound and
+/// multiply the cost of every state update. Bounding the set at insertion time
+/// (reject on overflow) caps that fan-out, mirroring
+/// `MAX_DOWNSTREAM_SUBSCRIBERS_PER_CONTRACT` in `ring/hosting.rs`, which guards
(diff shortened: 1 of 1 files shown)


iduartgomez:
Closing as superseded by #4596, which implements the same #3798 Gap 2 cap (interested_peers per contract) with the correct `fix:` prefix and a reviewed, deterministic-and-distinct test-key helper (`make_unique_peer_key`) that avoids the RNG-collision flake. #4596 is CLEAN and ready to merge.

PR #4521, opened 2026-06-20
fix: guard trace-ot feature build in CI

### LINKED ISSUE #4225: build: opentelemetry 0.32 broke SpanBuilder usage in `trace-ot` feature

mvanhorn:
## Summary

Adds a CI step that builds the optional `trace-ot` feature through clippy, so the OpenTelemetry tracing path cannot silently break again. Also resolves the two clippy errors that surface only when that feature is enabled.

## Why this matters

The default Clippy job runs `cargo clippy --locked -- -D warnings`, which does not enable `trace-ot`. After the workspace bumped to `opentelemetry = "0.32"`, the feature-gated tracing code broke and the break was invisible to CI. The `SpanBuilder` construction was already repaired on `main` (#4385), but the feature is still not exercised in CI, so it can rot again. Two `wildcard_enum_match_arm` errors in feature-gated match arms also fail `cargo clippy -p freenet --features trace-ot -- -D warnings`.

This wires that exact command into the Clippy job and clears the remaining feature-gated lint failures. See #4225.

## Changes

- `.github/workflows/ci.yml`: new `clippy (trace-ot feature)` step running `cargo clippy --locked -p freenet --features trace-ot -- -D warnings`, using the same linker env as the existing clippy step.
- `crates/core/src/tracing.rs`:
  - `span_completed`: classify the terminal `UpdateEvent::UpdateSuccess` / `UpdateFailure` variants as span-completing. They previously fell through `_ => false`, so UPDATE transaction spans were never removed under `trace-ot`. The remaining wildcard keeps a deliberate `#[allow(clippy::wildcard_enum_match_arm)]`.
  - `From<&NetLogMessage> for Option<Vec<KeyValue>>`: add the same narrow `#[allow]` with a comment, matching the existing pattern in this file.

## Testing

- `cargo clippy -p freenet --features trace-ot -- -D warnings` passes (failed before this change). Verified on Linux; the only residual error in a local macOS run is a pre-existing `undocumented_unsafe_blocks` in `module_cache.rs` gated behind `cfg(all(unix, not(target_os = "linux")))`, which compiles out on the Linux CI runner.
- `cargo build -p freenet --features trace-ot --lib` succeeds.

Fixes #4225

AI was used for assistance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -401,6 +401,16 @@ jobs:
           CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS: -C link-arg=-fuse-ld=mold
         run: cargo clippy --locked -- -D warnings
 
+      # Guards the opt-in `trace-ot` OpenTelemetry feature against silent
+      # rot: it is excluded from the default build, so a refactor can break
+      # its #[cfg(feature = "trace-ot")] code paths without any other CI job
+      # noticing. See #4225.
+      - name: clippy (trace-ot feature)
+        env:
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER: clang
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS: -C link-arg=-fuse-ld=mold
+        run: cargo clippy --locked -p freenet --features trace-ot -- -D warnings
+
   # Windows compile check - catches missing imports, feature flags, and
   # platform-specific errors that Linux CI misses. Added after #3685 where
   # PRs #3669/#3680 introduced Windows-only code that failed to compile,
--- a/crates/core/src/tracing/metrics_client.rs
+++ b/crates/core/src/tracing/metrics_client.rs
@@ -644,5 +644,9 @@ mod opentelemetry_tracer {
     }
 }
 
+// `pub(crate)` (not `pub(super)`): `OTEventRegister` is re-exported again as
+// `pub(crate)` from `tracing.rs` and constructed in `node.rs` / `testing_impl`,
+// so it must be visible crate-wide. The narrower `pub(super)` broke the
+// `trace-ot` build after the tracing module split (see #4225).
 #[cfg(feature = "trace-ot")]
-pub(super) use opentelemetry_tracer::OTEventRegister;
+pub(crate) use opentelemetry_tracer::OTEventRegister;
--- a/crates/core/src/tracing/register.rs
+++ b/crates/core/src/tracing/register.rs
@@ -1553,8 +1553,14 @@ impl NetLogMessage {
     /// Signals whether this message closes a transaction span.
     ///
     /// In case of isolated events where the span is not being tracked it should return true.
+    // Terminal variants above complete a span; the wildcard is deliberate so
+    // non-span events (and any future variant) are treated as not-completing.
+    // `pub(crate)`: called from the sibling `metrics_client` module's
+    // OpenTelemetry span processing. Was private, which broke the `trace-ot`
+    // build after the tracing module split (see #4225).
     #[cfg(feature = "trace-ot")]
-    fn span_completed(&self) -> bool {
+    #[allow(clippy::wildcard_enum_match_arm)]
+    pub(crate) fn span_completed(&self) -> bool {
         match &self.kind {
             EventKind::Connect(ConnectEvent::Finished { .. }) => true,
             EventKind::Connect(_) => false,
@@ -1568,6 +1574,10 @@ impl NetLogMessage {
                 | SubscribeEvent::SubscribeTimeout { .. },
             ) => true,
             EventKind::Subscribe(_) => false,
+            EventKind::Update(
+                UpdateEvent::UpdateSuccess { .. } | UpdateEvent::UpdateFailure { .. },
+            ) => true,
+            EventKind::Update(_) => false,
             _ => false,
         }
     }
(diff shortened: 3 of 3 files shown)


sanity:
## Maintainer rebase + fix to get this mergeable

Thanks @mvanhorn — the core idea (a CI guard so the opt-in `trace-ot` feature can't silently rot, plus closing the `UpdateSuccess`/`UpdateFailure` spans) is exactly right. I rebased onto current `main` and pushed an updated commit, because the original single-commit diff no longer applied after the `tracing.rs` module split, and the rule-review correctly blocked the missing regression test.

What changed vs. the original:
- **Rebased** onto `main`; relocated the `span_completed` fix to its new home in `tracing/register.rs` (the function moved out of `tracing.rs` in the module split).
- **Unrotted the build.** Adding the `clippy --features trace-ot` step surfaced that the feature no longer compiles on current `main` (7 errors): `OTEventRegister` was re-exported `pub(super)` (breaking the `tracing.rs` re-export + the `node.rs`/`testing_impl` constructors), and `span_completed` was private (it's called from the sibling `metrics_client` module). Both widened to `pub(crate)` — pure visibility changes confined to `#[cfg(feature = \"trace-ot\")]` paths, so the default build is unaffected. This is precisely the rot the guard is meant to catch.
- **Added the regression test** the rule-review asked for: `span_completed_tests` covers `UpdateSuccess`/`UpdateFailure` → span-completing and `UpdateEvent::Request` → not.

Local verification (the exact new CI step + the test):
- `cargo clippy --locked -p freenet --features trace-ot -- -D warnings` → clean
- `cargo test -p freenet --features trace-ot --lib span_completed_tests` → 2 passed

Screened for malicious content as a first-time-contributor PR (clean). Auto-merge enabled; will land once CI is green.

[AI-assisted - Claude]

mvanhorn:
Appreciate the CI guard, @sanity. Building the trace-ot feature through clippy means the OpenTelemetry path can't silently rot.

PR #4527, opened 2026-06-20
fix(service): restore macOS + Windows binary build after #4517 split (#4526)

### LINKED ISSUE #4526: fix: macOS-native binary build broken on main after #4517 service.rs submodule split (30 compile errors)
## Problem

After PR #4517 (`refactor(service): split service.rs into submodules`) merged to `main`, the **macOS-native** `freenet` binary no longer compiles. Pristine `origin/main` (`af18e1f9a`) fails `cargo build -p freenet --bin freenet` on macOS with **30 compile errors**, all in the new `crates/core/src/bin/commands/service/` submodule layout:

```
 6 × E0364  re-export of private item (generate_plist, disable/enable/is_launch_at_login_*,
            macos_app_bundle_path, toggle_launch_at_login_outcome, ...)
 1 × E0365  re-export of private `ToggleLaunchAtLoginOutcome`
11 × E0603  private module / item access
 9 × E0425  cannot find function in scope (first_run_marker_path, is_first_run_at,
            legacy_migration_marker_path, mark_first_run_complete_at, ...)
 2 × E0282  type annotations needed
```

The errors are all **visibility / scope / missing-`use`** fallout from the split: helpers that were free functions in the old monolithic `service.rs` are now in submodules but still referenced unqualified (E0425) or re-exported while `private` (E0364/E0365/E0603).

## Impact

- macOS-native binary does **not** build from `main` — blocks local macOS development and any macOS release/cross build path that compiles natively.
- The Linux build is unaffected (the broken symbols are behind `#[cfg(target_os = "macos")]`), which is almost certainly why #4517's CI passed: **the split's CI did not build the macOS target.**

## Suggested fix

- Bump the visibility of the split-out helpers/types to `pub(super)` / `pub(crate)` so the re-exports in the parent module resolve (fixes E0364/E0365/E0603).
- Add the missing `use crate::commands::service::...` imports (or fully-qualify) at the new call sites (fixes E0425).
- Add the needed type annotations (E0282).
- **Follow-up**: add a macOS build job (or `cargo check --target` for macOS) to CI so a cfg-gated platform path can't merge red again — this is the exact failure mode `.claude/rules/deployment.md` warns about ("a cfg'd path CI does not build on that target is unverified").

Discovered while rebasing #4382 onto the post-#4517 layout; reproduced on pristine `origin/main`.

[AI-assisted]

iduartgomez:
## Problem

PR #4517 (`refactor(service): split service.rs into submodules`) broke the **macOS** and **Windows** binary builds on `main`. Pristine `origin/main` fails to compile both targets — 30 errors on macOS, the same class on the Windows Check CI job:

```
E0364 / E0365 / E0603  re-export / access of private split-out items
                       (generate_plist, *_launch_at_login, macos_app_bundle_path,
                        ToggleLaunchAtLoginOutcome, kill_freenet_service_processes, ...)
E0425                  unqualified calls to moved helpers
                       (find_latest_log_file, first_run_marker_path, is_first_run_at,
                        legacy_migration_marker_path, FIRST_RUN_OPENER_SPAWNED, ...)
E0282                  type annotation needed (test fixture write)
```

Every broken symbol is behind `#[cfg(target_os = "macos")]` / `#[cfg(target_os = "windows")]`, so #4517's **Linux-only** build never compiled them and the regression merged green. This blocks all local non-Linux development and any natively-compiled release path.

## Solution

Pure visibility/scope cleanup — **no behavior change**:
- Raise the split-out helpers/types to `pub(super)` / `pub(crate)` so the parent module's re-exports and cross-submodule references resolve (E0364/E0365/E0603).
- Add the missing `use` imports at the moved call sites (E0425).
- One type annotation on a test fixture write (E0282).

## Testing

macOS native: `cargo build -p freenet --bin freenet` → 30 errors → **0**; `cargo test -p freenet --bin freenet commands::service` → **72 passed**; bin clippy clean (the lone remaining lint is the pre-existing `module_cache.rs:520` undocumented-unsafe on an untouched file under rust-1.94, which CI's pinned toolchain passes). The durable regression guard is a non-Linux CI build job — tracked as a follow-up in the issue (the exact `deployment.md` "cfg'd path CI doesn't build = unverified" failure mode).

## Fixes

Fixes #4526

🤖 Found incidentally while rebasing #4382 onto the post-#4517 layout; root-caused and fixed.

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -187,17 +187,17 @@ fn open_url_in_browser(url: &str) {
 
 /// Path of the first-run marker file, if we can determine a data directory.
 #[allow(dead_code)] // Used by the tray wrapper on Windows/macOS only
-fn first_run_marker_path() -> Option<PathBuf> {
+pub(super) fn first_run_marker_path() -> Option<PathBuf> {
     dirs::data_local_dir().map(|d| d.join("freenet").join(".first-run-complete"))
 }
 
 #[allow(dead_code)]
-fn is_first_run_at(marker: &Path) -> bool {
+pub(super) fn is_first_run_at(marker: &Path) -> bool {
     !marker.exists()
 }
 
 #[allow(dead_code)]
-fn mark_first_run_complete_at(marker: &Path) -> std::io::Result<()> {
+pub(super) fn mark_first_run_complete_at(marker: &Path) -> std::io::Result<()> {
     if let Some(parent) = marker.parent() {
         std::fs::create_dir_all(parent)?;
     }
@@ -211,12 +211,12 @@ fn mark_first_run_complete_at(marker: &Path) -> std::io::Result<()> {
 /// onboarding state. The migration is its own one-shot, tracked by its own
 /// marker, so it runs once for existing DMG users too.
 #[allow(dead_code)]
-fn legacy_migration_marker_path() -> Option<PathBuf> {
+pub(super) fn legacy_migration_marker_path() -> Option<PathBuf> {
     dirs::data_local_dir().map(|d| d.join("freenet").join(".legacy-migration-complete"))
 }
 
 #[cfg(any(target_os = "windows", target_os = "macos"))]
-fn dashboard_port_is_listening() -> bool {
+pub(super) fn dashboard_port_is_listening() -> bool {
     use std::net::TcpStream;
     use std::time::Duration;
     let Ok(addr) = DASHBOARD_ADDR.parse::<std::net::SocketAddr>() else {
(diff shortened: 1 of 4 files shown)


iduartgomez:
**Applying `review-override` for the missing-`#[test]` warning, per the rule-review bot's own stated nuance.**

This is a `fix:` PR, but the bug it fixes is a **compile-time visibility/scope error gated behind `#[cfg(target_os = "macos")]` / `#[cfg(target_os = "windows")]`**. There is no Linux-hosted `#[test]` function that can reproduce this failure class — the regression manifests only when the macOS/Windows target is *compiled*, which the existing Linux CI never does (that's exactly how the #4517 split shipped it to `main` green).

The "regression test" here is the build itself:
- macOS native: `cargo build -p freenet --bin freenet` went 30 errors → 0; `commands::service` tests 72/72 pass.
- The same fix resolves the Windows Check compile errors seen on #4509 / #4522.

The durable guard is a **non-Linux CI build job** (a `cargo check`/build matrix for macOS + Windows), tracked as the follow-up in #4526 — that is the structural fix for this entire class (the `deployment.md` "a cfg'd path CI doesn't build is unverified" rule). Adding a contrived Linux `#[test]` would satisfy the heuristic without actually guarding anything.

Override is therefore the correct call rather than a fabricated test.

iduartgomez:
**Applying `test-exempt` for the Rule Lint regression-test gate.**

After rebasing onto current `main` (post-#4525, which already fixed the Windows half of this regression), this PR is narrowed to the **macOS-target** compile fix only. The bug is a compile-time visibility/scope error gated behind `#[cfg(target_os = "macos")]` — there is no Linux-hosted `#[test]` that can reproduce it (the CI host compiles Linux only, which is exactly how #4517 shipped the break to `main` green).

Verification is the build itself:
- macOS native: `cargo build -p freenet --bin freenet` 26 errors → **0**.
- `cargo test -p freenet --bin freenet commands::service` → 73 passed.

The durable regression guard is a **non-Linux CI build job** (macOS/Windows `cargo check` matrix), tracked as the structural follow-up in #4526 — the `deployment.md` "a cfg'd path CI doesn't build is unverified" rule. A fabricated Linux `#[test]` would satisfy the heuristic without guarding anything, so the exempt label is the correct mechanism here.

PR #4528, opened 2026-06-20
fix: make transport flight-size accounting consistent (plaintext vs encrypted)

### LINKED ISSUE #4402: Transport flight-size accounting mixes plaintext (on_send) and encrypted (release) byte counts
## Summary

Flight-size accounting mixes plaintext and encrypted byte counts: `on_send_with_token`
**increments** flight size by the **plaintext** fragment/packet size, while every
release path — ACK (`report_received_receipts` → `on_ack*`), Abandon
(`ResendAction::Abandon` → `release_flightsize`), and the new stream-abort
`drop_stream` (#4345) — **decrements** by the **encrypted** stored payload length
(`pending_receipts` payload = `to_packet_data()` output, which includes the AEAD tag +
framing overhead).

So each packet's release subtracts slightly more than was added (by the per-packet
crypto/framing overhead), biasing the connection-wide flight-size counter downward.

## Status / scope

This is **pre-existing** (it predates #4345) and **self-consistent across all three
release paths**, so #4345's `drop_stream` is correct relative to the existing
convention — it releases exactly what the same packets' ACK/Abandon would have. It is
*not* a regression introduced by the flight-size-on-abort fix; flagging it as its own
tracking item so the convention can be made consistent (add plaintext on both sides, or
encrypted on both sides) in a focused change.

## Impact

Low in practice: FixedRate (the default congestion control) is rate-limited by the
token bucket, not flight size, so the downward bias mainly loosens cwnd discipline
slightly. Worth tightening for BBR/LEDBAT paths where flight size is load-bearing.

## Suggested fix

Pick one convention end-to-end: either store the plaintext `packet_size` alongside the
encrypted payload in `pending_receipts` and release that, or add the encrypted size at
`on_send`. Add a unit test asserting `sum(on_send) == sum(release)` for a stream that is
fully ACKed, fully abandoned, and dropped via `drop_stream`.

Found by the external (Codex) reviewer during #4345 stage-2 review, and independently
noted by two Claude reviewers as pre-existing.

[AI-assisted - Claude]

mvanhorn:
## Summary

Flight-size accounting now adds and releases the same byte count for every packet, so the connection-wide flight-size counter no longer drifts downward over the life of a connection.

## Why this matters

`on_send`/`on_send_with_token` add the plaintext fragment/packet size to flight size, but all three release paths (ACK via `report_received_receipts` -> `on_ack*`, Abandon via `ResendAction::Abandon` -> `release_flightsize`, and `drop_stream`) released the encrypted stored payload length (`payload.len()`, which includes the AEAD tag and framing overhead). Every release subtracted slightly more than was added, biasing flight size low. It is low-impact under the default FixedRate controller but matters for BBR/LEDBAT, where flight size is load-bearing. Fixes #4402.

## Changes

- Store the plaintext `packet_size` alongside the encrypted payload in each `pending_receipts` entry, and release that stored plaintext size on all three paths (ACK, Abandon, `drop_stream`). This makes `sum(on_send) == sum(release)`.
- Thread the exact `packet_size` the caller passed to `on_send_with_token` down through `packet_sending` into the tracker, rather than recomputing it from the serialized payload. This handles fragment #1 with embedded metadata, where the actual fragment payload is smaller than the byte count added to flight size, so a recomputed value would under-release.
- Resend re-registration (`refresh_sent_packet` and the insert-or-refresh path) preserves the original plaintext byte count, since flight size is added only on the first send.
- Update the accounting-convention doc comment in `sent_packet_tracker.rs` to describe the new plaintext-on-both-sides convention.

## Testing

- `cargo test -p freenet --lib transport::sent_packet_tracker` — 61 passed, including a new `test_issue_4402_plaintext_flightsize_releases_balance_all_paths` that asserts flight size returns to zero after a stream is fully ACKed, fully abandoned, dropped via `drop_stream`, and a mixed ACK/drop sequence.
- `cargo test -p freenet --lib transport::peer_connection::tests::packet_sending` — 2 passed, including a new `packet_sending_tracks_caller_flight_size_for_metadata_stream_fragment` regression test proving a metadata-bearing fragment releases the caller's flight-size bytes, not `payload.len()`.
- `cargo check -p freenet --lib` — clean.

AI was used for assistance.


sanity:
## Review summary — Full tier (transport / flight-size accounting)

This PR touches a high-risk surface (transport congestion-control accounting), so it went through a full multi-perspective + multi-model review. As a first-time-contributor PR it was also screened for malicious content before review (clean — no injected content, no suspicious deps/build scripts; changes confined to the three transport files).

**Reviewers (all independent):**
- **Correctness (Claude):** enumerated every flight-size add path vs. release path and verified they balance. The stored 4th tuple field is exactly the caller's `on_send`/`on_send_with_token` argument, threaded through `packet_sending` (not recomputed), and all three release paths — ACK (`report_received_receipts`), Abandon (`ResendAction::Abandon{payload_len}`), and `drop_stream` — return that stored plaintext size. Retransmit re-registers via `refresh_sent_packet`/the refresh branch, preserving `slot.3`, so no double-count or double-release. Subtractions are saturating. No findings.
- **Concurrency + tests (Claude):** the add/track gap on send failure is covered by the existing `release_aborted_stream_flightsize` + `release_flightsize(packet_size)` error paths; counter is `AtomicUsize` + `parking_lot::Mutex`; the #4345 removal-lockstep invariant (pending_receipts + packet_streams removed together) is preserved, so drop_stream-vs-ACK can't double-release. Tests model `encrypted > plaintext` and would fail on the old code. No findings.
- **External non-Claude model (Codex):** \"No actionable correctness issues were identified in the diff. The changes consistently carry the caller's flight-size byte count through ACK, abandon, and stream-drop release paths.\"

**Verdict:** clean on all three passes. The regression tests (`test_issue_4402_plaintext_flightsize_releases_balance_all_paths` and `packet_sending_tracks_caller_flight_size_for_metadata_stream_fragment`) genuinely exercise the metadata-fragment case where the old code drifted. Auto-merge enabled; will land once CI is green on the current head.

[AI-assisted - Claude]

mvanhorn:
Thank you @sanity. Making the transport flight-size accounting add and release the same byte count keeps the connection-wide counter honest.

PR #4529, opened 2026-06-20
fix(hosted): sandbox CSP must honor X-Forwarded-Proto/Host behind a TLS proxy

sanity:
## Problem

The sandboxed-iframe Content-Security-Policy origin was built as `http://{Host}`, ignoring `X-Forwarded-Proto`. Behind a TLS-terminating reverse proxy — the supported hosted-mode deployment (#4381) — the browser's origin is `https://<public-host>`, so the CSP `default-src http://…` blocked the webapp's own `https://` assets and the app never loaded.

Found running River through a real Caddy TLS proxy on a hosted node: every `assets/river-ui-*.js` load tripped `violates the following Content Security Policy directive: "default-src http://127.0.0.1:7509 …"` and the page stayed blank. This makes hosted mode unusable in a browser for its intended (TLS-proxied) deployment.

## Approach

Extract `sandbox_origin_from_headers(&HeaderMap)` and derive the browser-facing origin from the request headers:
- use `https` when `X-Forwarded-Proto: https` is present (set by the trusted proxy — the same forwarded-header trust model the hosted-mode token gate already relies on; an explicit `X-Forwarded-Proto: http` does **not** upgrade);
- prefer `X-Forwarded-Host` over `Host` for host:port (a proxy may rewrite the upstream `Host` to its loopback target — nginx does this by default);
- a direct connection carries neither header and falls back to the previous `http://{Host}`; `'self'` only when no host is available.

The forwarded headers are client-spoofable through a careless proxy, but the CSP is **defence-in-depth, not a trust boundary**: a mismatched origin only *breaks* the sandboxed app (too-strict CSP), it cannot widen what the opaque-origin (`sandbox` without `allow-same-origin`) iframe may reach — so honoring them here is safe.

## Testing

New regression tests in `client_api`:
- `sandbox_origin_honors_forwarded_https_scheme` (reproduces the bug → `https://localhost:8443`),
- `sandbox_origin_forwarded_proto_without_forwarded_host_uses_host`,
- `sandbox_origin_direct_connection_is_http_host` (incl. explicit `X-Forwarded-Proto: http` not upgrading),
- `sandbox_origin_no_host_falls_back_to_self`.

Verified **end-to-end live**: a hosted node behind a Caddy TLS proxy now serves the sandbox CSP as `default-src https://localhost:8443 …`; River renders with **0 CSP violations** and the WebSocket connects carrying the per-user token (two browser sessions get distinct tokens).

Pending Full multi-model review.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -46,6 +46,47 @@ fn sandbox_csp_for_origin(origin: &str) -> String {
     )
 }
 
+/// Derive the browser-facing origin to interpolate into the sandbox CSP from
+/// the request headers, honoring a TLS-terminating reverse proxy's forwarded
+/// scheme and host.
+///
+/// The CSP origin MUST match the origin the *browser* actually used, or the
+/// CSP blocks the webapp's own assets. Behind a TLS proxy (the supported
+/// hosted-mode deployment) the browser's origin is `https://<public-host>`,
+/// while the node itself is reached over loopback `http`. We therefore:
+///
+/// - use `https` when `X-Forwarded-Proto: https` is present (set by the
+///   trusted proxy — same forwarded-header trust model as the hosted-mode
+///   token gate, which already requires the operator's proxy to set/strip
+///   `X-Forwarded-*`); otherwise `http`.
+/// - prefer `X-Forwarded-Host` over `Host` for the host:port, since a proxy
+///   may rewrite the upstream `Host` to its loopback target (nginx does this
+///   by default) while preserving the original in `X-Forwarded-Host`.
+///
+/// A direct connection carries neither forwarded header and falls back to the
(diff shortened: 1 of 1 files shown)


sanity:
## Review (Light tier — focused bug fix)

External: Codex (`codex review --base main`) — after addressing its one finding (comma-separated `X-Forwarded-Host` → take the first entry), re-review came back "no discrete regression introduced by this patch." Maintainer read + **live verification**: deployed behind a real Caddy TLS proxy, the sandbox CSP now serves `default-src https://<host>` and River renders with 0 CSP violations (was blank/blocked before). 5 unit tests cover https-upgrade, forwarded-host, direct-connection (incl. explicit `X-Forwarded-Proto: http` not upgrading), no-host fallback, and the comma-separated multi-proxy case. CI green on HEAD. Merging.

[AI-assisted - Claude]

PR #4530, opened 2026-06-20
feat(hosted): shell-chrome bar with proxy disclosure and access-key backup

sanity:
## Problem

A hosted-mode node (#4381) serves apps to untrusted visitors, but the shell page had no host-controlled surface to (a) tell the user, un-spoofably, that they're on a hosted proxy and their data isn't private, and (b) let them back up their access or take their data to their own peer. These controls **cannot** live inside the sandboxed app iframe: the app is untrusted (it could hide or spoof the disclosure) and the durable per-user token is held by the shell and must never reach the sandbox.

## Solution

A thin **shell-chrome bar** rendered OUTSIDE the sandboxed iframe, only in hosted mode:

- **Always-visible disclosure** ("Hosted proxy, not private. Your data is stored on this server."), worded so the critical "not private" survives truncation on narrow/mobile viewports.
- An **Account popover** splitting two actions that would otherwise blur:
  - **Access key** (your *login*): *Copy key* / *Restore from key*. The key is the per-user token, read from the shell-only `__freenet_user_token` global and copied to the clipboard, or re-seeded into `localStorage` and reloaded — it never enters the iframe.
  - **Your own peer** (your *data*): *Export data* — a placeholder until the node-side export endpoint lands (it will run P3 `secrets export` scoped to the connection's token and stream the bundle as a download).

Markup/CSS/JS are raw-string consts (`HOSTED_BAR_*`); the bar is appended via `{hosted_styles}`/`{hosted_bar}` placeholders that are empty in non-hosted mode, so a normal single-user node renders the exact pre-existing layout.

## Why shell chrome, not in-sandbox

Two hard reasons (both load-bearing for the rest of #4381): the disclosure must be un-spoofable by whatever app is loaded, and the access-key/export actions are token-scoped — the token lives in the shell and is deliberately never exposed to the sandbox.

## Testing

`shell_page_hosted_mode_renders_proxy_chrome_bar`: the bar, disclosure, access-key controls, and export control are present in hosted mode and **absent** in non-hosted mode. Verified live on a hosted node behind a Caddy TLS proxy (desktop + mobile screenshots); Copy/Restore work end to end.

Part of #4381. The "Export data" button is wired in a follow-up that adds the node export endpoint.

Pending review.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -883,6 +883,24 @@ fn shell_page(
         // fail-closed branch never triggers and behavior is unchanged.
         (String::new(), format!("freenetBridge(\"{auth_token}\");"))
     };
+
+    // Hosted-mode "shell chrome": a thin, host-controlled bar rendered OUTSIDE
+    // the sandboxed app iframe. It is the only place a per-user-data action
+    // (export to your own peer) can live — the durable user token is held by
+    // the shell and must never reach the sandbox — and the only place the
+    // "this is a hosted proxy, not private" disclosure cannot be hidden or
+    // spoofed by the contract app. Empty (and the layout unchanged) when hosted
+    // mode is off. The export control is a placeholder until the node-side
+    // export endpoint lands (P3 `secrets export` over HTTP, scoped to the
+    // connection's user token).
+    let (hosted_styles, hosted_bar) = if hosted_mode {
+        (
+            format!("\n<style>{HOSTED_BAR_STYLES}</style>"),
+            format!("{HOSTED_BAR_HTML}\n<script>{HOSTED_BAR_JS}</script>"),
+        )
+    } else {
+        (String::new(), String::new())
+    };
     let html = format!(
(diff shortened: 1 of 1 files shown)


sanity:
## Review (frontend shell chrome — security-adjacent)

External: Codex (`codex review --base main`) — "only adds hosted-mode shell chrome and a regression test, with the new markup/scripts gated off for non-hosted mode. No discrete correctness, security, or compatibility issue introduced." Maintainer read confirms: the bar/popover are static raw-string consts (no request input interpolated → no injection); the access key is read from the shell-only `__freenet_user_token` global and never reaches the sandboxed iframe (copy-to-clipboard / restore-from-prompt are the user's own credential); non-hosted output is unchanged (test `shell_page_hosted_mode_renders_proxy_chrome_bar` asserts the bar is absent without hosted mode). Verified live on a TLS-proxied hosted node (desktop + mobile). CI green on HEAD. Merging.

[AI-assisted - Claude]

PR #4531, opened 2026-06-21
feat(hosted): node export endpoint for per-user data (#4381)

sanity:
## Problem

P3 (#4035) added the offline `freenet secrets export` CLI: it opens a `SecretsStore` from disk and writes an encrypted bundle. But a hosted user who tried Freenet through a public gateway can't run a CLI against the operator's node — they need to download their per-user delegate secrets **live, from the running node, over HTTP**, then re-import on their own peer. That live path did not exist.

This is the P3-live piece of the hosted/public Freenet proxy (#4381).

## Solution

A **self-contained freenet-core HTTP operation** — no freenet-stdlib wire-format change. `GET /v{1,2}/hosted/export` returns the connection's per-user secrets as an encrypted P3 `FNSX` bundle download.

- **Runs on the executor** (which owns the `SecretsStore`) via a new internal `ContractHandlerEvent::ExportUserSecrets` + a `ContractExecutor::export_user_secrets` trait method. The bytes come back off-wire through a dedicated response variant, **not** through `HostResponse`, so no stdlib `ClientRequest`/`HostResponse` variant is added (the explicit hard constraint of this task).
- `Runtime::export_secret_bundle` wraps the existing `secret_export::export_bundle` (the one place that can see the `pub(super)` `secret_store`).
- **Bundle key material** = `BundleKeyMaterial::Token(<raw user token>)`, so the user re-imports with the token they already hold (`freenet secrets import --use-token-key`).
- **Security — same gate as the WS `userToken`** (the refuse-plaintext-token invariant): hosted mode ON + a loopback source + `X-Forwarded-Proto: https`. The gate primitives (`decide_user_token` / `is_loopback_source` / `derive_user_context`) were factored to `pub(crate)` and are reused **verbatim**; a reusable helper `export_user_context_or_reject` adds only the export-specific twist that a no-token / `Local` outcome is **rejected** (403) rather than falling through (an export of "no user scope" is meaningless). The token is read from the **`X-Freenet-User-Token` header**, never a query param (keeps it out of the URL / access logs), and is never logged (held in a redacted, zeroizing `RedactedToken`).
- **Per-node routing, not a global.** The handler reaches the executor through a per-node `ExportOpManagerHandle` (a router `Extension` filled at startup via a new `ClientEventsProxy::set_op_manager`), so concurrent in-process hosted nodes never clobber each other's route. (An earlier process-global singleton draft failed parallel integration tests — caught and fixed.)

### Internal mechanism chosen

Option (a)-ish, taken off-wire: a new internal `ContractHandlerEvent` variant + trait method, reached from the HTTP handler via `op_manager.notify_contract_handler(...)`. Option (b) (a dedicated executor channel) was rejected because the executor lives behind the contract-handler in network mode (the real hosted deployment), and `notify_contract_handler` is the existing, audited route to it. Returning bytes through `HostCallbackResult`/`HostResponse` (the WS reply path) was rejected because that **is** the stdlib wire format.

### Exact route + download headers

- `GET /v1/hosted/export` and `GET /v2/hosted/export`
- `Content-Type: application/octet-stream`
- `Content-Disposition: attachment; filename="freenet-data.fnsx"` (matches the P3 `FNSX` magic)
- `403` for RejectInsecure / no-token / flag-off; `503` if the node isn't wired yet; `500` on an executor-side failure.

## Testing

- **9 gate unit tests** (`hosted_export::tests`): honor a secure request (returns the raw token bytes as bundle key); reject hosted-off / no-token / empty-token / no-XFP / non-loopback source / missing source IP / IPv6 non-loopback; honor IPv4-mapped loopback.
- **4 live end-to-end integration tests** (`tests/hosted_mode_export.rs`) against a **real hosted node** (full event loop + executor + WASM runtime):
  - secure request → non-empty `FNSX` bundle that round-trips via `import_bundle` back to the original secret (Local **and** User scope), with the plaintext never appearing verbatim in the bundle;
  - a **wrong token** fails AEAD authentication (proves the bundle key is genuinely the user's token);
  - every **insecure** variant (no token / no XFP / empty token) → 403 with no bundle;
  - **flag-off** node → 403.
  These pass in **parallel**, proving the per-node handle has no cross-node race.
- Existing `tests/hosted_mode_isolation.rs` still green; `cargo fmt`, `cargo clippy --locked -- -D warnings`, and `cargo test --workspace --no-run` all clean.

## Notes / where I was unsure

- Export is supported in **network mode** (the hosted-gateway deployment, which the integration test exercises). In **local mode** (`fdev`/standalone) the handle is never wired, so the endpoint returns 503 — hosted mode is a gateway feature, so this is the right boundary, but flagging it explicitly.
- The `ClientEventsProxy::set_op_manager` argument is `&dyn Any` carrying an `Arc<OpManager>` on purpose: the trait is `pub` but `OpManager` is `pub(crate)`, so naming it directly would leak a more-private type. The concrete proxy downcasts; external impls (fdev) inherit the default no-op.

## Known limitation (P5)

Authenticated, unbounded loop-blocking DoS (flagged by both adversarial review lenses; the gate itself holds — no bypass / no cross-user leak / no token leak). The export enumerates and AEAD-decrypts **every** secret in the user's scope **synchronously on the single-threaded contract-handling loop**, with no per-user secret-count or bundle-size cap, so a token-holder with a large per-user secret set (or one who simply repeats the request) can stall all other contract ops on the node for each export's duration. Acceptable here **only** because the endpoint ships behind the **default-off** hosted flag and is not on shared/public infra. Before it is exposed publicly, a **P5 (abuse/quotas)** follow-up must add (1) a per-user export quota and (2) off-loop execution (`spawn_blocking`). This PR documents the risk in the module/handler/`Runtime::export_secret_bundle` rustdoc; it deliberately does **not** add the quota/offload (that is P5 scope). Tracked under #4381 (P5).

Closes part of #4381 (P3-live).

[AI-assisted - Claude]


sanity:
## Review (Full tier — auth + secrets, highest-risk surface)

This endpoint returns a user's private delegate secrets, so it got a full multi-perspective + external-model pass on the actual checked-out code.

**External:** Codex (`codex review --base main`) — no correctness/security/maintainability issues; gated consistently with the existing hosted-mode token logic, wired through the live executor.

**Adversarial lens — auth-gate-bypass:** tried to break the gate; could not. No bypass (source IP is kernel-set `ConnectInfo`, not header-spoofable; XFP trusted only after loopback proven; hosted_mode is server config, fail-closed-to-off). No cross-user leak (per-request `UserSecretContext`, per-request oneshot correlation, `enumerate_scope` filters other users). Token header-only, redacted, zeroized; bundle is ciphertext (wrong token → AEAD fail). `Local`/no-token → 403.

**Adversarial lens — internal-mechanism/concurrency:** per-node `ExportOpManagerHandle` (no global; the concurrent-node clobber is correctly fixed), per-request oneshot-by-id (a concurrent export cannot receive another's bundle — same path GET/PUT use), pooled executors share one redb handle with serialized checkout, channel-safe (unbounded event send + oneshot response, no bounded `.send().await` in any loop), export scoped purely to the connection token, round-trip proven, mocks graceful.

**One finding (both lenses): authenticated, unbounded DoS** — the export runs synchronously on the single-threaded contract loop with no per-user secret cap, so a *token-holder* could stall the node. Not anonymous (full auth required), same shape as the existing `delegate_request` path, and it maps to **P5 (abuse/quotas)**. Now loudly documented (module + handler rustdoc + PR 'Known limitation') and tracked as a **P5 blocker before public exposure**. Deliberately NOT pulling a quota/offload into this PR (P5 scope); acceptable behind the default-off hosted flag.

**Tests:** 9 gate unit tests + 4 live e2e (round-trip, wrong-token AEAD, all insecure → 403). CI green on HEAD. All findings addressed. Merging.

[AI-assisted - Claude]

Issue #4534, opened 2026-06-21
Placement migration (#4404) overruns module-cache ceiling on 0.2.80 nodes → "contract queue full" client outage


iduartgomez:
## Problem

After the network rolled to **0.2.80** and #4404 SubscribeHint placement migration re-activated (floor `(0,2,80)`), a long-running node's hosted working set grows without bound, overruns the WASM module-cache budget, and the serial `contract_handling` loop stalls on constant recompiles. The fair queue then fills (`MAX_QUEUED_PER_CONTRACT=100` / `MAX_TOTAL_FAIR_QUEUE=5000`) and **every client request is rejected** with:

```
error while executing operation in the network: contract queue full, try again later
```

The node is alive and burning CPU (`STAT=RN`), but its WS/HTTP API is effectively unusable — local `GET /v1/contract/web/...` requests get rejected behind the migration-driven flood.

### Live evidence (production macOS node, 0.2.80, 64 GB host)

**`hosted=` was flat at 1462 for 18+h, then climbs monotonically the moment a 0.2.80 peer appears:**

| time (UTC) | hosted |
|---|---|
| 06-19 22:00 → 06-20 16:00 | **1462** (flat) |
| 06-20 16:26 | first `Newer version discovered ... highest_seen=0.2.80` |
| 06-20 17:00 | 1466 |
| 06-20 18:00 | 1489 |
| 06-20 19:00 | 1594 |
| 06-20 22:00 | 1748 |
| 06-21 09:00 | **1898** (still rising, no plateau) |

**`queue full` rejections track the same onset** (per-hour count): flat ~0 until 06-20 15:00, then 156 → **2593 → 5570 → 4355** and sustained ~1000–2300/h thereafter. 902 client-facing `PutQuery rejected by executor ... queue full` in a single hour.

**Module cache is pinned at its default ceiling and persistently evicting:**

```
WARN module_cache: compiled-WASM contract cache is persistently evicting at its byte budget —
the hosted working set exceeds the cache, so modules are being recompiled on access.
cache="contract" evictions=299 window_secs=300 budget_bytes=1610612736 (1.5 GiB)
total_bytes=1610157992 entries=941
contract_module_cache_evictions_total=50840   # +2750/hour
```

478 distinct contracts cold-compiled per few minutes, **every one a unique miss** (pure thrash, not new contracts).

## Root cause

Two independently-known issues compose into a production outage:

[...]

PR #4535, opened 2026-06-21
fix(contract): priority-tier the contract event loop so client requests aren't starved (#4534)

iduartgomez:
## Problem

A node that accepts placement-migration caching work (#4404, re-enabled in 0.2.80) grows its hosted working set without bound. When that set overruns the WASM module cache, the serial `contract_handling` loop spends its time on background `summarize` recompiles, the flat round-robin fair queue fills to `MAX_TOTAL_FAIR_QUEUE`, and the node rejects its **own** client requests with `contract queue full, try again later`.

Observed live on a production node (#4534): `hosted=` climbed 1462 → 1898+, ~2300 queue-full rejections/hour, the local web API unusable while the node was busy serving migrated contracts it didn't care about.

The flat queue had no notion of **who** an event served — a best-effort background summarize competed on equal footing with a locally-initiated client GET/PUT, and admission rejected whatever arrived next regardless of origin.

## Solution

**Key insight:** the contract event loop is single-threaded, so background work and client work must not share one undifferentiated queue. Tag every contract-handler event with a `Priority` class (`ClientLocal > NetworkRelay > Background`) and make `FairEventQueue` priority-aware:

- **`pop()` drains higher classes first** — newly-arrived background work never head-of-line-blocks a pending client request. (An already-running WASM compile still finishes; true in-flight preemption needs off-loop compile — follow-up.)
- **`try_push()` reserves `CLIENT_LOCAL_RESERVE` (256) admission slots** that only `ClientLocal` may use, so a client request is never refused because background/relay work filled the queue.
- **`evict_background()` sheds already-queued background** to make room for a foreground push (the reserved lane is useless if the queue is already full of background, so the loop evicts-then-retries). Shed background re-emits next cycle — non-fatal, client-silent.
- **Per-contract caps are now scoped per tier** — background churn on a contract can't block a client op on the same contract.

Origin is set at the send sites: local WS/HTTP client paths (`client_events.rs`) → `ClientLocal`; periodic interest-sync summarize (`summary_if_hosted_or_in_use`) and disk-reclamation `EvictContract` → `Background`; everything else keeps the `NetworkRelay` default (no behavior change vs the old flat queue for relay traffic). Broadcast-path summaries that serve real UPDATE propagation deliberately stay at `NetworkRelay`.

This is the **queue-admission/scheduling half** of #4534. Cache-retention pinning by local interest and backpressure-aware migration admission are separate follow-ups tracked in the issue.

## Testing

New `fair_queue` unit tests:
- priority-ordered `pop()` (ClientLocal → NetworkRelay → Background)
- ClientLocal admitted past the soft cap into the reserve; hard-cap rejection
- `evict_background()` touches only the Background tier (foreground survives)
- per-contract cap scoped per tier (Background-saturated contract still admits a ClientLocal op on the same key)
- within-tier round-robin preserved

Full suites green: **19** fair_queue, **308** contract, **30** op_state_manager, **286** node (incl. the summarize-gating source-scrape pins). `cargo fmt` clean; no new clippy findings on the changed files.

> Note: the local pre-commit clippy hook flags 4 **pre-existing** lints (module_cache.rs, hosted_export.rs, telemetry.rs, rate.rs) that are new in clippy 1.94 but do not fire under CI's pinned 1.93 — none are touched by this PR.

## Fixes

Partially addresses #4534 (queue-admission/scheduling layer). Follow-ups for cache-retention pinning + migration backpressure remain open on the issue.

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -64,6 +64,51 @@ for every inbound wire message. Pattern per op:
 `RelayState::handle_request`. Production CONNECT runs entirely on the
 drivers in `connect/op_ctx_task.rs`.
 
+## Critical Invariant: Tag Contract-Handler Events with the Right Priority (#4534)
+
+```
+Every event sent to the single-threaded contract_handling loop carries a
+`crate::contract::Priority` (ClientLocal > NetworkRelay > Background).
+The fair queue drains higher classes first, reserves admission capacity
+(CLIENT_LOCAL_RESERVE) for ClientLocal, and sheds Background to make room.
+A mis-tag silently degrades service: tag a client path Background and it
+gets starved/evicted; tag relay/background ClientLocal and it can exhaust
+the reserve meant to protect real client requests.
+
+Rules for choosing the class at a send site:
+
+  ClientLocal  → the event serves a WS/HTTP client connected to THIS node.
+                 Set it at the client-facing entry points:
+                 - client_events.rs request handlers (GET, RegisterSubscriberListener,
+                   DelegateRequest)
+                 - drive_client_update → update_contract(.., ClientLocal)
+                 - hosted_export.rs ExportUserSecrets
+                 - a client's own PUT reaches the relay driver via
+                   originator-loopback (upstream_addr == own_addr); compute it
+                   with put_store_priority() — DO NOT hardcode NetworkRelay on
+                   the relay PUT store path.
+
+  NetworkRelay → serving a remote peer's relayed op (the default). All
+                 drive_relay_* drivers, ResyncResponse apply, broadcast-path
+                 summaries/deltas that propagate a real UPDATE. This is also
+                 `Priority::DEFAULT`, so untagged callers fall here safely.
+
+  Background   → best-effort node-internal work that must yield to client/relay
+                 traffic: the periodic interest-sync summarize
+                 (summary_if_hosted_or_in_use), placement-migration caching,
+                 renewal, and EvictContract disk reclamation. NEVER tag a
+                 client- or peer-visible op Background.
+
+When a shared helper is reached from BOTH a client and a relay path (e.g.
+put_contract, update_contract), thread a `priority` param through rather
+than hardcoding — decide the class at the driver that knows the origin.
+
+The loop's pop() applies a bounded anti-starvation floor
+(RELAY_STARVATION_FLOOR) so sustained ClientLocal load cannot indefinitely
+starve NetworkRelay/Background — but that is a backstop, not a license to
+mis-tag.
+```
+
 ## Critical Invariant: Initialize-Before-Send
 
 ```
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -284,12 +284,15 @@ async fn register_subscription_listener(
         "Registering subscription listener"
     );
     let register_listener = op_manager
-        .notify_contract_handler(ContractHandlerEvent::RegisterSubscriberListener {
-            key: instance_id,
-            client_id,
-            summary: None, // No summary for GET/PUT-based subscriptions
-            subscriber_listener: subscription_listener,
-        })
+        .notify_contract_handler_prioritized(
+            ContractHandlerEvent::RegisterSubscriberListener {
+                key: instance_id,
+                client_id,
+                summary: None, // No summary for GET/PUT-based subscriptions
+                subscriber_listener: subscription_listener,
+            },
+            crate::contract::Priority::ClientLocal,
+        )
         .await
         .inspect_err(|err| {
             tracing::error!(
@@ -1049,10 +1052,13 @@ async fn process_open_request(
                             Err(err) if !subscribe => {
                                 // Not joined yet — check if we can serve from local cache
                                 let local_result = op_manager
-                                    .notify_contract_handler(ContractHandlerEvent::GetQuery {
-                                        instance_id: key,
-                                        return_contract_code,
-                                    })
+                                    .notify_contract_handler_prioritized(
+                                        ContractHandlerEvent::GetQuery {
+                                            instance_id: key,
+                                            return_contract_code,
+                                        },
+                                        crate::contract::Priority::ClientLocal,
+                                    )
                                     .await;
                                 if let Ok(ContractHandlerEvent::GetResponse {
                                     key: Some(full_key),
@@ -1091,10 +1097,13 @@ async fn process_open_request(
                         // if subscribed (cache is fresh), otherwise fetch from network.
                         // See PR #2388 for why always-local-first was problematic.
                         let (full_key, state, contract) = match op_manager
-                            .notify_contract_handler(ContractHandlerEvent::GetQuery {
-                                instance_id: key,
-                                return_contract_code,
-                            })
+                            .notify_contract_handler_prioritized(
+                                ContractHandlerEvent::GetQuery {
+                                    instance_id: key,
+                                    return_contract_code,
+                                },
+                                crate::contract::Priority::ClientLocal,
+                            )
                             .await
                         {
                             Ok(ContractHandlerEvent::GetResponse {
@@ -1338,11 +1347,12 @@ async fn process_open_request(
                         // GET+subscribe=true and PUT+subscribe=true bypass this check
                         // because those operations inherently fetch/provide the WASM.
                         match op_manager
-                            .notify_contract_handler(
+                            .notify_contract_handler_prioritized(
                                 crate::contract::ContractHandlerEvent::GetQuery {
                                     instance_id: key,
                                     return_contract_code: true,
                                 },
+                                crate::contract::Priority::ClientLocal,
                             )
                             .await
                         {
@@ -1408,13 +1418,14 @@ async fn process_open_request(
                         };
 
                         let register_listener = op_manager
-                            .notify_contract_handler(
+                            .notify_contract_handler_prioritized(
                                 ContractHandlerEvent::RegisterSubscriberListener {
                                     key,
                                     client_id,
                                     summary,
                                     subscriber_listener,
                                 },
+                                crate::contract::Priority::ClientLocal,
                             )
                             .await
                             .inspect_err(|err| {
@@ -1663,11 +1674,14 @@ async fn process_open_request(
                 let user_context = request.user_context;
 
                 let res = match op_manager
-                    .notify_contract_handler(ContractHandlerEvent::DelegateRequest {
-                        req,
-                        origin_contract,
-                        user_context,
-                    })
+                    .notify_contract_handler_prioritized(
+                        ContractHandlerEvent::DelegateRequest {
+                            req,
+                            origin_contract,
+                            user_context,
+                        },
+                        crate::contract::Priority::ClientLocal,
+                    )
                     .await
                 {
                     Ok(ContractHandlerEvent::DelegateResponse(res)) => {
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -9,6 +9,7 @@ use freenet_stdlib::prelude::*;
 
 mod executor;
 mod fair_queue;
+pub(crate) use fair_queue::Priority;
 pub(crate) mod governance;
 mod handler;
 pub mod storages;
@@ -1048,7 +1049,7 @@ where
         // single-digit microseconds even at peak load.
         for _ in 0..fair_queue::MAX_DRAIN_BATCH {
             match contract_handler.channel().try_recv_from_sender()? {
-                Some((id, event)) => {
+                Some((id, event, priority)) => {
                     // Process ClientDisconnect inline — it's a lightweight cleanup
                     // operation that should never be delayed or compete with
                     // DelegateRequest events for the default queue's limited capacity.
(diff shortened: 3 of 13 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4535

### Summary
- **PR Title:** fix(contract): priority-tier the contract event loop so client requests aren't starved (#4534)
- **Type:** fix (bug fix — #4534 is `T-bug`, `P-high`, maintainer-assigned)
- **CI Status:** Passing where complete (Clippy, Fmt, Conventional Commits, Rule Lint, macOS E2E, CLA, Snyk all green); Unit/Integration, Simulation, NAT, Windows still pending at review time.
- **Linked Issues:** Partially addresses #4534 (queue-admission/scheduling half).
- **Review tier:** Full (high-risk surface: concurrency + scheduling on the single-threaded contract event loop).
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four Claude specialists) + a 3-lens diverse-Claude-lens external substitute (scheduling-correctness, enum/wire-compat, fresh bug-hunt). **External model (codex) was unavailable** — see note at end.
- **HEAD SHA reviewed:** `85411891353c9ea4313d360ac2b7348be02743a6`

---

### Code-First Analysis
**Independent Understanding:** Adds a 3-level priority class (`Background < NetworkRelay < ClientLocal`) to the contract event loop. `FairEventQueue` is rewritten from a flat per-contract round-robin into one `PriorityTier` per class: `pop()` drains the highest non-empty tier first; `try_push()` reserves `CLIENT_LOCAL_RESERVE=256` admission slots usable only by `ClientLocal`; `evict_background()` sheds queued Background to free room; per-contract caps are now scoped per tier. `InternalCHEvent` carries the priority over the in-process mpsc channel. Send sites tag: local client GET/Subscribe/RegisterListener/Delegate → `ClientLocal`; periodic interest-sync summarize + `EvictContract` GC → `Background`; everything else (relay ops, broadcast/resync summaries) → `NetworkRelay` default.

**Stated Intent:** Stop best-effort background work (placement-migration caching, interest-sync summarize) from starving a node's own client requests with "contract queue full" rejections.

**Alignment:** Matches — with one honest, well-communicated scope cut (admission/scheduling only; cache-retention pinning and migration backpressure are tracked follow-ups). Code, doc comments and tests are internally consistent; the crate compiles cleanly.

**Gaps:** The headline "client requests aren't starved" is only **partially** delivered — see Should-Fix #2.

---

### Testing Assessment
**Coverage Level:** Adequate for the data structure; a gap on the orchestration layer.

| Unit | ✅ | 5 new `fair_queue` tests: priority-ordered pop, ClientLocal-into-reserve, hard-cap rejection, `evict_background` touches only Background tier, per-contract cap per tier, within-tier round-robin. Deterministic (pure data structure, no TimeSource/GlobalRng). |
| Integration | ⚠️ | `push_with_background_eviction` (the evict-then-retry orchestration in `contract.rs`, the centerpiece of the fix) is **not tested at all**. The eviction-on-per-contract-cap path (Must/Should #1) has no regression test. |
| Simulation | N/A | Pure data-structure change; no DST harness needed. |
| E2E | N/A | |

**Regression Test:** The `fix:`-PR new-test-function CI gate is satisfied. The closest true regression test for #4534 is `client_local_admitted_when_queue_full_of_background` (admission half, data-structure level) — acceptable for a scheduling data structure.

**Missing Tests:**
- `push_with_background_eviction` evict-then-retry branches: retry-still-fails, Background-rejected-does-not-evict, EvictContract-reclamation-on-shed. The mock-channel harness already in `contract.rs` (`setup_rejected_event`) supports this.
- No test asserts the send sites attach the correct `Priority` (a silent mis-tag would make the fix inert).
- An explicit single-event soft-cap boundary assertion (last NetworkRelay admitted at `soft_cap-1`, first refused at `soft_cap`); currently only inferred from bulk fills.

---

### Skeptical Findings
**Risk Level:** Medium (one real wasteful-work defect; one design-tradeoff for maintainer sign-off; the rest verified clean).

| Eviction fired on per-contract-cap rejection wastes Background work | Should Fix | `contract.rs:1568` + `fair_queue.rs:283` (verified) | Eviction gated only on `priority > Background && background_queued()>0`, not on *why* the push failed. A foreground push rejected by the per-contract cap sheds up to 256 Background events, retries, fails identically. Under hot-key load can drive Background to ~0 with thousands of slots free. Non-data-losing (Background re-emits) but undercuts the fairness goal. |
| Strict-priority `pop()` has no anti-starvation floor for NetworkRelay | Consider (design) | `fair_queue.rs:338-348` (verified) | Sustained local-client load can indefinitely starve remote DHT relay execution. Strictly worse than `main`'s flat round-robin for relay liveness. Deliberate #4534 tradeoff, but unbounded — mirrors the AGENTS.md "exemptions must be time-bounded" meta-pattern. Needs explicit maintainer acceptance or an aging floor. |
| Client PUT/UPDATE + hosted export not `ClientLocal` | Should Fix (overclaim) | `operations/put.rs:204`, `operations/update.rs:552`, `server/client_api/hosted_export.rs:310` (verified) | Local client PUT/UPDATE/export reach the handler via the default `NetworkRelay` path (shared op machinery has no origin tag), so they are NOT in the reserve. The "clients aren't starved" promise covers GET/Subscribe/Delegate only. Notably, #4534's own live evidence is **902 `PutQuery rejected by executor ... queue full`/hour** — the exact `NetworkRelay`-tier path this PR does not protect — so the headline client-facing symptom is only partially addressed. |
| Per-contract cap now 3×100 per contract globally | Consider | `fair_queue.rs:283` (verified) | Not remotely exploitable: ClientLocal/Background tags are set only at in-core local/internal sites; a remote peer is still capped at 100/contract in NetworkRelay. Flag only so the relaxed bound is a conscious choice. |

**Verified clean** (investigated, no bug): `total_queued` accounting across evict+retry (consistent with Σ tier.queued, only `saturating_sub`); EvictContract shed → `track_pending_reclamation` is idempotent (DashMap insert) with no double-count or lost reclamation; reserve math exact (`soft_cap=4744`, ClientLocal gets exactly 256, no off-by-one); enum order `Background=0/NetworkRelay=1/ClientLocal=2` consistent across `as usize` indexing, `Ord` comparisons and `.rev()` draining; `Priority` is in-process-only (no Serialize/Deserialize; enclosing struct holds a non-serializable oneshot) so no wire/disk compat hazard; `ClientDisconnect` still handled inline unchanged; no `biased;` select and no `.send().await` on a bounded channel introduced (`send_queue_full_response` uses oneshot responders); no panic/index/overflow risk.

---

### Big Picture Assessment
**Goal Alignment:** Partial by design (honest) — this is the scheduling/admission half; the node still thrashes the module cache on background recompiles but no longer rejects its own clients. A strictly-better intermediate state, clearly communicated.

**Anti-Patterns Detected:** None. No CI-chasing, no assertion-weakening, no `#[ignore]`, no error-swallowing. The 4 pre-existing clippy lints the PR deliberately leaves untouched are correctly out of scope.

**Removed Code Concerns:** None. `test_global_limit` was rewritten into `test_foreground_soft_cap_then_client_reserve_then_hard_cap`, which is **strictly stronger** (keeps the old hard-cap rejection assertion, adds soft-cap + reserve coverage). No coverage lost.

**Scope Assessment:** Focused — one logical change. The test-helper refactor (`make_contract_id_u32`, `try_push_default`, `fill_to`) is justified by the new `priority` parameter and the need for >256 distinct contracts to exercise the reserve.

---

### Documentation
- **Code docs:** New `pub(crate)`/`pub` items (`Priority`, the `_prioritized` methods) are doc-commented with `#4534` and DEFAULT guidance. ✅
- **One doc/code mismatch:** `push_with_background_eviction`'s doc comment (`contract.rs:1547`) says eviction fires when a foreground event "hits the soft admission cap," but the code fires it on *any* foreground rejection (including the per-contract cap). Tighten code or comment.
- **Orphaned doc comment** (`contract.rs:1536-1542`): the `// When the fair queue rejects an EvictContract...` block that documents `track_pending_reclamation_if_evict` is now stranded *above* the newly-inserted `push_with_background_eviction`, so `push_with_background_eviction` carries two stacked doc comments and `track_pending_reclamation_if_evict` (line 1591) has lost its doc. Move the block back down to immediately precede `fn track_pending_reclamation_if_evict`.
- **Architecture/agent rules:** `.claude/rules/operations.md:76-103` documents the `contract_handling`-loop sequencing invariants but says nothing about the new priority dimension, and no rule captures the **send-site `Priority` tagging invariant** (the design depends on every new client/background entry point picking the right class — the PUT/UPDATE gap is already an instance of getting it wrong). Per AGENTS.md's "fix missing/outdated docs in the same PR" rule, add a short rule: client WS/HTTP paths → `ClientLocal`; periodic/maintenance/GC (interest-sync summarize, `EvictContract`, renewal) → `Background`; relay/broadcast-propagation paths keep the `NetworkRelay` default. (`.claude/rules/contracts.md:281`'s "normal fair-queue order" wording is also now incomplete.)

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
1. **Don't evict Background when the rejection is the per-contract cap** (`contract.rs:1561-1585`). Gate the evict-retry on a total/soft-cap rejection (e.g. have `try_push`/`RejectedEvent` expose the reason, or check `total_queued >= soft_cap` before evicting). Add a regression test: hot contract at the per-contract cap + queued Background → a further foreground push must NOT drain Background. Also fix the doc comment at `contract.rs:1547` to match actual behavior.
2. **Correct the PR description re: client-write coverage** — it currently implies "local WS/HTTP client paths → ClientLocal" covers client requests, but client PUT/UPDATE writes and hosted export remain on the unprotected `NetworkRelay` tier (shared op machinery has no origin tag). Since #4534's own evidence centers on `PutQuery` queue-full rejections, state explicitly that only client reads/subscribe/delegate get `ClientLocal`, and track client-write prioritization (threading `Priority` through the put/update op state machine) as a follow-up.
3. **Fix the orphaned doc comment** at `contract.rs:1536` and **update `.claude/rules/operations.md`** (and the `contracts.md:281` wording) to document priority tiering, the `ClientLocal` reserve, `evict_background`, and the send-site `Priority` tagging invariant (required by the repo's own doc-maintenance rule).

#### Consider (Suggestions)
1. **Anti-starvation floor for `NetworkRelay`** or an explicit maintainer decision that permanent relay starvation under sustained local-client load is acceptable. This is a network-liveness behavior change vs `main`; no test pins any relay-progress guarantee.
2. Tag the hosted-export HTTP path (`hosted_export.rs:310`) `ClientLocal` — it is a local client request and the prime intended beneficiary of the reserve.
3. Add `push_with_background_eviction` integration tests (retry-still-fails, EvictContract-shed-reclamation) and an explicit soft-cap single-event boundary test.

---

### Verdict

The change is well-built, well-tested at the data-structure level, in scope as a bug fix (#4534 is `T-bug`/`P-high`, maintainer-assigned), with no Must-Fix correctness break. The blocking-adjacent items are: the wasteful per-contract-cap eviction (#1, with its doc mismatch) and the doc-rule update (#3) the repo's own AGENTS.md requires. The strict-priority relay-starvation tradeoff (Consider #1) is a real network-liveness behavior change that deserves a conscious maintainer sign-off given `#4534` carries `S-needs-design`. Once #1's eviction gating + test land (a small, localized change) and the doc updates are made, a focused diff-of-the-diff re-check on the fix commits is sufficient — the surrounding tiering logic, accounting, reserve math, and compat were all verified clean.

> External-model note: `codex review` was attempted twice and failed both times with HTTP 401 (`refresh_token_reused` — the local codex auth token is expired and needs a manual re-login, not a near-term quota reset); `gemini` CLI is not installed. Per the review policy, the external pass was substituted with a 3-lens diverse-Claude-lens review (scheduling-correctness, enum/wire-serialization-compat, fresh independent bug-hunt), each reading the checked-out code independently. The strongest findings (per-contract-cap eviction, relay starvation, PUT/UPDATE tagging gap) each surfaced from multiple independent lenses and were re-verified by hand against the source before inclusion. Re-running the real external model before merge is advisable given the high-risk surface.

[AI-assisted - Claude]

iduartgomez:
Addressed all four review findings in e4c91b5db:

**1. Reason-aware eviction (Should Fix).** `try_push` now returns a `RejectReason` (`GlobalCapacity` vs `PerContract`). `push_with_background_eviction` only sheds Background on a `GlobalCapacity` rejection — a per-contract-cap rejection no longer triggers a futile shed-of-256 + identical-failure retry. Fixed the stale doc comment that claimed eviction fires only on the soft cap.

**2. Client PUT/UPDATE/export now `ClientLocal` (Should Fix).** This was the real gap — only client GET was protected while the incident's own evidence was 902 `PutQuery ... queue full`/hour.
- `update_contract` takes a `priority`; `drive_client_update` → `ClientLocal`, the four `drive_relay_*` → `NetworkRelay`.
- `put_contract`/`relay_put_store_locally` take a `priority`; the relay PUT drivers compute it via `put_store_priority` — originator-loopback (`upstream_addr == own_addr`) is a local client's own PUT → `ClientLocal`; genuine relay store stays `NetworkRelay`; unknown own-addr fails safe to `NetworkRelay`.
- hosted_export `ExportUserSecrets` → `ClientLocal`.

**3. Anti-starvation floor for NetworkRelay (Consider/design).** Approved the tradeoff as maintainer, but bounded it rather than leaving strict-priority unbounded: `pop()` tracks a `client_streak` and after `RELAY_STARVATION_FLOOR` (16) consecutive ClientLocal pops *with lower-tier work waiting*, serves one lower-tier event. Client work keeps the vast majority of slots; an uncontested client burst never trips the floor. New tests assert the bounded cadence and the no-trip-when-uncontested case.

**4. Docs (Should Fix).** Restored the orphaned `track_pending_reclamation_if_evict` doc comment and added a "Tag Contract-Handler Events with the Right Priority" invariant to `.claude/rules/operations.md` (send-site classification + the shared-helper threading rule).

Tests added: reason carried on per-contract vs global rejection, relay served on a bounded cadence under sustained client load, uncontested client burst doesn't trip the floor. fair_queue/contract/operations/node suites all green; no new clippy findings on the changed files.

PR #4537, opened 2026-06-21
fix(node): gate placement-migration admission on module-cache headroom (#4534)

sanity:
## Problem

Inbound `SubscribeHint` placement migration (#4404, re-enabled in 0.2.80) had no coupling to module-cache capacity. A node kept accepting contracts migrated toward its key until the hosted working set overran the WASM module cache (`MAX_DEFAULT_MODULE_CACHE_BUDGET_BYTES`, 1.5 GiB; `total_ram / 8` binds first). Past that point every access recompiles an evicted module (Cranelift), pinning the single-threaded `contract_handling` loop, filling the fair queue to `MAX_TOTAL_FAIR_QUEUE`, and rejecting the node's own client requests with `contract queue full, try again later`. On memory-bound gateways the same growth OOM-kills the process (vega: anon-rss 6.96 GB, OOM at ~21h uptime — see #4534).

## Approach

Gate the **receive** side of placement migration on module-cache headroom. In the `SubscribeHint` arm of `handle_pure_network_message_v1`, after the existing version-floor / already-hosting / holder-validity checks and before `start_directed_subscribe`, refuse the hint when contract-module-cache occupancy is at or above a 90% ceiling.

- **No new pressure signal.** Occupancy is read from the existing lock-free `MODULE_CACHE_METRICS` gauges (`contract_total_bytes / contract_budget_bytes`), the same telemetry already emitted for #4440. `None` (gauge not yet published, i.e. no runtime pool built) admits, so there is no behavior change in test/sim or early-startup contexts.
- **Migration stays on.** Only the directed-subscribe placement nudge is gated; the node's own local client subscribes/GETs are never gated here. Refusal is silent and best-effort — the holder re-proposes on its next migration trigger once headroom returns.
- **~10% headroom** below 100% is reserved for the node's own locally-requested contracts.
- **Send side unchanged** — a node keeps proposing migrations to others; only its own acceptance is throttled by its own capacity.

This is the migration-acceptance half of #4534. It complements the queue-scheduling / priority-tiering fix in #4535 (@iduartgomez): #4535 stops client requests from being starved; this PR stops the hosted set from growing past cache capacity in the first place (the OOM / thrash root cause). Neither alone resolves both symptoms. Cache-retention pinning by local interest remains a tracked follow-up on #4534.

## Testing

- `occupancy_pct` unit tests: boundary math (0 / 50 / 90 / 100 / over-budget), uninitialized-budget → `None`, and GiB-scale no-overflow.
- `migration_admission_allowed` unit tests: admit when occupancy is `None` or below the ceiling; refuse at/above (including the over-budget transient).
- Source-scrape pin (`subscribe_hint_arm_gates_migration_on_cache_backpressure`): the `SubscribeHint` arm must call the admission gate before `start_directed_subscribe`, so removing or reordering the gate trips the build.

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and the `wasm_runtime::module_cache` + `operations::subscribe::tests` suites pass locally.

Part of #4534 (paired with #4535).

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -22,7 +22,7 @@ Driver entry points:
 | GET | `get/op_ctx_task.rs::start_client_get`, `start_relay_get`, `start_sub_op_get`, `start_targeted_sub_op_get` |
 | PUT | `put/op_ctx_task.rs::start_client_put`, `start_relay_put`, `start_relay_put_streaming` |
 | UPDATE | `update/op_ctx_task.rs::start_client_update`, `start_relay_request_update`, `start_relay_broadcast_to`, `start_relay_request_update_streaming`, `start_relay_broadcast_to_streaming` |
-| SUBSCRIBE | `subscribe/op_ctx_task.rs::start_client_subscribe`, `start_directed_subscribe` (SubscribeHint placement nudge, #4404), `run_executor_subscribe`, `run_renewal_subscribe`, `start_relay_subscribe`; `subscribe.rs::handle_unsubscribe_inbound` for `Unsubscribe` |
+| SUBSCRIBE | `subscribe/op_ctx_task.rs::start_client_subscribe`, `start_directed_subscribe` (SubscribeHint placement nudge, #4404 — the inbound-hint receive arm in `node.rs` now refuses it above the module-cache occupancy ceiling, `migration_admission_allowed`, #4534), `run_executor_subscribe`, `run_renewal_subscribe`, `start_relay_subscribe`; `subscribe.rs::handle_unsubscribe_inbound` for `Unsubscribe` |
 
 The shared retry-loop driver lives in `op_ctx.rs` (`RetryDriver` trait
 + `drive_retry_loop`). UPDATE is fire-and-forget (no retry loop, no
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1827,6 +1827,31 @@ where
                 );
                 return Ok(());
             }
+            // Backpressure-aware migration admission (#4534): refuse to take on
+            // a NEW migrated contract when the contract module cache is already
+            // at/above its occupancy ceiling. Accepting more would push the
+            // hosted working set past cache capacity, forcing recompile-on-
+            // access thrash that fills the fair queue ("contract queue full")
+            // and OOMs memory-bound gateways. This gates ONLY the directed-
+            // subscribe placement nudge; the node's own local client
+            // subscribes/GETs are never gated here. The hint is dropped silently
+            // and the holder re-proposes it on its next migration trigger once
+            // headroom returns.
+            let contract_cache_occupancy = crate::wasm_runtime::contract_cache_occupancy_pct();
+            if !migration_admission_allowed(contract_cache_occupancy) {
+                tracing::debug!(
+                    key = %hint.key,
+                    holder = %hint.holder,
+                    ?source_addr,
+                    occupancy_pct = ?contract_cache_occupancy,
+                    ceiling_pct = MIGRATION_ADMISSION_MAX_CONTRACT_CACHE_OCCUPANCY_PCT,
+                    "Refusing inbound SubscribeHint: contract module cache at/above \
+                     the migration-admission occupancy ceiling — deferring placement \
+                     migration to bound the hosted working set and avoid cache thrash \
+                     (#4534)"
+                );
+                return Ok(());
+            }
             tracing::debug!(
                 key = %hint.key,
                 holder = %hint.holder,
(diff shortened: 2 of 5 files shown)


sanity (commented):
## Multi-Model Review: PR #4537

### Summary
- **Type:** `fix:` — backpressure-aware placement-migration admission (#4534, the migration-acceptance half; #4535 is the queue-scheduling half)
- **Review tier:** **Full** (touches networking / migration acceptance + contract runtime — a high-risk surface)
- **Reviewers run:** 4 Claude perspectives (code-first, skeptical, testing, big-picture) + **Codex** (`gpt-5.5`, external non-Claude). **Gemini was unavailable** (`IneligibleTierError` — the CLI client is no longer supported for this account); per the multi-model-review fallback, the external pass is covered by Codex plus the four diverse Claude lenses.
- **Reviewed HEAD:** lenses ran on `8d3b1d99`; review-addressing doc/test changes pushed as `90248546` (docs + one unit test + one rules-doc line, no production-logic change — light re-check, see Verdict).

---

### Code-First — alignment: matches intent
Implementation does what the PR claims, verified independently. Signal `total_bytes*100/budget` with `>=90% → refuse` expressed as `pct < 90 → admit`; boundary correct (90 refuses, 89 admits, over-budget transients refuse). Gate is correctly sequenced (version-floor → already-hosting → holder==sender → **gate** → `start_directed_subscribe`) and scoped to the `SubscribeHint` arm only, so local client subscribes/GETs are never gated. `None`-on-uninitialized-budget admits, which is the right default (gauge is published at pool construction, before any hint is processed). Compiles clean.

### Testing — coverage: adequate and proportionate
Two-layer regression guard: pure-function unit tests for the threshold/occupancy math + a source-scrape pin asserting the gate runs *before* `start_directed_subscribe` (so deleting/reordering the gate — the exact #4534 reintroduction — trips the build). Pin anchors verified unique. CI's `fix:`-PR new-test gate satisfied (8 new tests). The gate is a no-op in existing sims (process-global gauge returns `None`), so a faithful end-to-end "node X at 91% refuses while node Y at 50% accepts" test is infeasible with the current global gauge; the unit+pin combination is the correct substitute and the logic error is fully reproducible at the unit level.

### Skeptical — risk: low, no Must-Fix
Confirmed: no permanent-starvation *bug* (recovery is real via LRU age-out, and the holder re-proposes on new-peer / op-completion triggers — both paths verified to recur); `None`/stale-signal handling correct; no reintroduced #4404 dead-end beyond the inherent "saturated node sheds placement" tradeoff (strictly better than the node falling over); threshold/off-by-one sound; bug-prevention-patterns clean (synchronous lock-free read + early return — no `.send().await`, no spawn, no unbounded collection); silent-drop is the safe DoS direction.

### Big-Picture — scope: clean, aligned
Single focused bug fix, purely additive (no removed code/tests), correctly classified `fix:`. **Confirmed disjoint from #4535**: its `node.rs` hunks are at ~2376/2555/2620 (interest-sync/summarize), this PR's at ~1827/1922 — no region overlap, no semantic conflict, merge in either order. The 90% is a dimensionless ratio of the already-configurable `FREENET_MODULE_CACHE_BUDGET_BYTES`, so it scales with operator config (not a hardcoded-absolute violation).

---

### Findings & resolution

| Severity | Finding | Source | Resolution |
|---|---|---|---|
| **P1 (external)** | Occupancy is a *lagging* signal; a burst of hints (e.g. cold-start with empty cache) can be admitted before the gauge rises, so the gate trips only after over-admission | Codex | **Addressed by justification + docs.** Codex's "unbounded migrated working set" premise is disproven in code: the hosted set is independently bounded by the **hosting cache's own byte budget** (`ring/hosting/cache.rs` — byte-budget LRU+TTL, default 1 GiB) **and** the node's ring-responsibility region. The residual is a *transient* overshoot before the lag catches up, mitigated by (a) the steady-state gauge pinning ~100% in the actual incident (so the gate is continuously active), (b) #4535 prioritizing client requests during any transient, and (c) the hosting cache's own eviction. A leading-indicator estimate (hosted-count × observed avg module size) was considered and rejected for this PR: at true cold-start it can't estimate (no observed sizes yet), it adds a fragile second heuristic with its own false-refusal regression risk against the #4404 feature, and it would not improve the dominant case (a rightful set that genuinely exceeds the cache is the cache-sizing problem #4441, not an admission problem). Documented as a deliberate proxy in `module_cache.rs`. |
| Should-Fix | Docs framed refusal as transient ("once headroom returns"); on an under-budgeted node it is effectively **permanent** by design | skeptical + code-first | **Fixed** in `90248546` — rustdoc now states recovery requires the working set to drop below budget OR the operator raising the cache budget, and that this is one of #4534's four directions (issue stays open). |
| Should-Fix | Add explicit 89/90 truncation-boundary test | testing | **Fixed** in `90248546` (`occupancy_pct_truncates_below_the_decision_boundary`). |
| Consider | Document the proxy-signal caveat (occupancy = compiled bytes, not hosted set) | code-first | **Fixed** in `90248546`. |
| Consider | Note in `operations.md` that directed-subscribe migration is now capacity-gated | big-picture | **Fixed** in `90248546`. |
| Consider | Pin-test delimits the arm by the next variant (`Aborted`); a variant inserted between would widen the window | skeptical | Acknowledged — matches the repo's established source-scrape house style; left as-is. |
| Note (not this PR) | `start_directed_subscribe` consumes a `ClientLocal` admission slot — migration work claiming #4535's reserved *client* lane is worth aligning | big-picture | Flagged to @iduartgomez for #4535; out of scope here (pre-existing classification, and this gate only reduces that pressure). |

---

### Verdict
**State: Ready to Merge.** No Must-Fix from any of the four Claude perspectives; the single external P1 rests on a premise (unbounded growth) disproven by the hosting-cache byte-bound, and is addressed by documentation + a high-bar justification. All Should-Fix / Consider items are fixed in `90248546`, which is docs + one unit test + one rules-doc line on top of the reviewed `8d3b1d99` — purely additive, no production-logic change, so a **light re-check** (diff-of-the-diff) is sufficient rather than a full re-review.

This is explicitly the migration-acceptance half of #4534; it pairs with #4535 (queue scheduling). Both are needed for the full fix, and #4534 stays open for the remaining directions (cache sizing #4441, offload-compilation default, interest-weighted retention). Merge/release will additionally wait on green CI at the current HEAD and on #4535 landing.

[AI-assisted - Claude]

PR #4538, opened 2026-06-21
fix(telemetry): suppress test-build telemetry to prod collector (#4366)

### LINKED ISSUE #4366: telemetry: simulation/test peers pollute the production collector (default endpoint nova.locut.us:4318)
## Problem

While analyzing production telemetry on nova (2026-06-07, validating PR #4353), a large share of `peer_startup` / `get_not_found` / GET-volume events turned out to be synthetic test traffic, not real network peers:

- Bursts of 10-43 `peer_startup` events per minute, hundreds of one-shot peer IDs (each ID seen exactly once), all `os: linux` (Debian trixie), many with `is_gateway: true` — the real network has only 2 gateways. Example: 312 unique "gateway" IDs reporting version 0.1.131 within minutes on 2026-06-06.
- `get_not_found` events with requester addresses in the simulation loopback range, e.g. `4foYgyDUUtSjoh1yF@127.1.2.1:52160` (observed 2026-06-07 ~21:06Z, while a local `simulation_integration` test run was active on nova).
- Reported "versions" from these bursts (0.1.90 … 0.1.170, 0.2.14, 0.2.18, 0.2.55) make per-version network analysis unusable without manual filtering.

Cause: `DEFAULT_TELEMETRY_ENDPOINT = "http://nova.locut.us:4318"` in `crates/core/src/config.rs:1524`, and simulation/integration test peers apparently don't override or disable it, so any local/CI sim run ships its events to the production collector.

## Impact

- Version-distribution, GET success/not_found and transfer-volume metrics from the collector are contaminated; any analysis (e.g. #4221's 30K not_founds, or post-release fix validation like #4353/#4345) must hand-filter sim traffic with heuristics.
- Sim runs are not labeled, so filtering is guesswork (loopback addresses only appear in some event bodies; `transfer_failed` events carry no peer_id at all).

## Suggested fix (any of)

1. Disable telemetry by default under `simulation_tests`/`testing` features (or in `SimNetwork` setup).
2. Tag events with a `network`/`env` resource attribute (e.g. `sim` vs `prod`) so the collector can route sim traffic to a separate sink.
3. Make the default endpoint opt-in for non-release builds.

[AI-assisted - Claude]

sanity:
## Problem

Simulation/integration test peers were shipping telemetry to the **production** OTLP
collector (`DEFAULT_TELEMETRY_ENDPOINT = http://nova.locut.us:4318`), contaminating
network metrics (issue #4366). Symptoms observed on nova: bursts of one-shot synthetic
"gateway" peer IDs, `get_not_found` events with simulation-loopback requester addresses
(e.g. `…@127.1.2.1:…`), and bogus version strings — all making per-version GET / transfer
analysis unusable without manual filtering.

Earlier work added an `is_test_environment` flag (set when the `--id` CLI flag is present)
that makes `TelemetryReporter::new` return `None`. The `SimNetwork` harness
(`testing_impl.rs`) sets `id: Some(label)`, so it was already suppressed. **The residual
gap**: the integration tests under `crates/core/tests/*.rs` (`operations_before_join`,
`hosted_mode_export`, `hosted_mode_isolation`, `in_process_restart`, `error_notification`)
build `NodeConfig` directly via `ConfigArgs { ..Default::default() }`, which leaves
`id: None` → `is_test_environment: false` and `telemetry.enabled: true` pointing at the
production endpoint. Those tests therefore still initialized a real reporter and shipped to
prod.

## Approach

Add a belt-and-suspenders guard in `TelemetryReporter::new`: in addition to the existing
`is_test_environment` (`--id`) check, return `None` when the binary is a test build —
`cfg!(test) || cfg!(feature = "testing")`.

Three overlapping signals, none of which the production node binary matches:

- `cfg!(test)` — the library's own unit tests.
- `cfg!(feature = "testing")` — integration/simulation tests as CI compiles them
  (`--features ...,testing`, see `ci.yml` / `simulation-nightly.yml`).
- `running_under_cargo_test()` — a best-effort runtime check (current exe's parent dir is
  `deps/`) that also covers the **documented local dev path** `cargo test -p freenet`
  run WITHOUT `--features testing` (AGENTS.md:122), where the library links as a normal
  dependency so `cfg!(test)` is false and the integration tests build `ConfigArgs::default()`
  (no `--id`). This was raised by the Codex review (see below) and is the realistic
  developer-laptop leak vector.

Why not `cfg!(test)` alone: integration tests in `crates/core/tests/*.rs` link the
`freenet` library as a normal dependency, so `#[cfg(test)]` is **not** set in the library
when those tests run.

This is **test-infrastructure-only**: the production `freenet` binary is built with default
features (`cargo install --path .` in the node/gateway Dockerfiles; the release workflow
sets no `testing` feature), so neither `cfg(test)` nor `feature = "testing"` is active and
real-node telemetry is completely unchanged.

`TelemetryReporter::new` is the single chokepoint — the global `TELEMETRY_SENDER` used by
standalone-event emission is only populated inside `new`, so returning `None` also drops
standalone/transport events. No other emission path bypasses it.

## Testing

- `telemetry_reporter_suppressed_in_test_build`: constructs a production-shaped
  `TelemetryConfig` (`enabled: true`, prod endpoint, `is_test_environment: false`) — exactly
  what the integration tests pass — and asserts `TelemetryReporter::new(...).is_none()`.
  Fails without the guard, passes with it.
- `detects_running_under_cargo_test`: pins that the runtime harness detector fires for a
  cargo test binary (the signal covering the no-feature `cargo test` path).
- `cargo fmt`, `cargo clippy --features testing`, existing telemetry unit tests.

## Review

- **Codex** (`codex review --base origin/main`): raised one P2 — the original
  `cfg!(test) || cfg!(feature = "testing")` guard left the documented `cargo test -p freenet`
  (no `--features testing`) path able to leak. Addressed by adding the runtime
  `running_under_cargo_test()` signal. Re-reviewed after the change.
- **Gemini**: unavailable (`IneligibleTierError` — account tier no longer supported, not a
  transient outage), so the external pass is Codex only. Supplemented with the four-lens
  Claude review (code-first, testing, skeptical, big-picture); the skeptical lens caught a
  redundant `use crate::config::TelemetryConfig;` import (already in scope via `super::*`),
  now removed.

## Fixes

Closes #4366

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/tracing/telemetry.rs
+++ b/crates/core/src/tracing/telemetry.rs
@@ -233,6 +233,35 @@ struct TelemetryEvent {
     priority: EventPriority,
 }
 
+/// Best-effort runtime detection of a cargo test/bench harness.
+///
+/// Cargo builds every unit-test, integration-test and bench binary into a
+/// `deps/` directory (e.g. `target/debug/deps/operations_before_join-<hash>`)
+/// and runs it from there. The real node binary is run as `…/freenet` (from a
+/// `bin/` install dir, or `target/<profile>/freenet` under `cargo run`), whose
+/// parent directory is never `deps`. So "the current executable's parent dir is
+/// named `deps`" cleanly distinguishes a cargo test harness from a production
+/// node, regardless of which Cargo features were enabled.
+///
+/// This is the only signal that catches the documented local dev command
+/// `cargo test -p freenet` run WITHOUT `--features testing`: there the library
+/// is a plain dependency (so `cfg!(test)` is false) and the integration tests
+/// build `ConfigArgs::default()` (so `is_test_environment` is false). See
+/// issue #4366.
+///
+/// Best-effort: if `current_exe()` fails we return `false` and fall back to the
+/// compile-time guards, which already cover unit tests and CI.
+fn running_under_cargo_test() -> bool {
+    std::env::current_exe()
+        .ok()
+        .and_then(|exe| {
+            exe.parent()
+                .and_then(|dir| dir.file_name())
+                .map(|name| name == "deps")
+        })
+        .unwrap_or(false)
+}
+
 impl TelemetryReporter {
     /// Create a new telemetry reporter.
     ///
@@ -240,7 +269,9 @@ impl TelemetryReporter {
     /// best-effort address), used to attribute transport-level events
     /// that have no `NetLogMessage` context.
     ///
-    /// Returns None if telemetry is disabled or in a test environment.
+    /// Returns None if telemetry is disabled, in a `--id` test environment,
+    /// or in a test build/harness (`cfg(test)` / `feature = "testing"` /
+    /// running under cargo test — see `running_under_cargo_test`).
     pub fn new(config: &TelemetryConfig, local_peer_id: String) -> Option<Self> {
         if !config.enabled {
             tracing::info!("Telemetry reporting is disabled");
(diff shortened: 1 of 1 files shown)


sanity:
## Review status

**External model pass (Codex):** ran `codex review --base origin/main`. It raised one P2 — the initial `cfg!(test) || cfg!(feature = "testing")` guard left the *documented* local dev path `cargo test -p freenet` (run without `--features testing`, per AGENTS.md:122) still able to ship test telemetry to the prod collector, since the library links as a normal dependency there (so `cfg!(test)` is false) and the integration tests build `ConfigArgs::default()` (so `is_test_environment` is false). **Addressed** by adding a best-effort runtime signal `running_under_cargo_test()` (current exe's parent dir is `deps/`, which cargo test/bench binaries always are and the real node binary never is). Re-ran Codex on the updated content.

**Gemini:** unavailable (`IneligibleTierError` — the CLI account tier is no longer supported; not a transient outage). So the external pass is Codex-only this round. To compensate for the lost model diversity I ran the four Claude review lenses myself:
- *skeptical* caught a redundant `use crate::config::TelemetryConfig;` in the test module (already in scope via `super::*`) — removed.
- verified gemini's concern #1 manually: `fdev` enables `freenet`'s `testing` feature, but `fdev` never calls `TelemetryReporter::new` / runs a node, and the prod gateway/node Docker images build the `freenet` binary with **default** features only — so `cfg!(feature = "testing")` touches no production telemetry path.

**CI note:** the pre-rebase run failed only on `freenet::nat_subscription::test_nat_peer_remote_subscription_receives_update` — a network/subscribe-caching integration test unrelated to telemetry (the telemetry guard itself logged "Telemetry disabled in test build" correctly in that run). Rebased onto current `origin/main`, which now includes #4535 (priority-tier the contract event loop so client requests aren't starved) and #4537 (gate placement-migration on module-cache headroom) — both directly about the event-loop/module-cache timing this flake exhibits. Re-running CI on the rebased HEAD.

[AI-assisted - Claude]

sanity:
## Codex re-review (post-rebase, final content): clean

Re-ran `codex review --base origin/main` on the rebased HEAD (`70a972cd`, content with the `running_under_cargo_test()` runtime guard). Result:

> The change adds an early telemetry suppression guard for test builds/harnesses and regression tests for that behavior. I did not identify any discrete correctness issue introduced by the diff.

The earlier P2 (no-feature `cargo test` leak) is resolved. Independent external review is satisfied (Codex); Gemini remained down (tier-ineligible) so model diversity was supplemented with the Claude lenses noted above. Waiting on green CI on the current HEAD before merge.

[AI-assisted - Claude]

PR #4539, opened 2026-06-21
fix(router): clamp per-peer adjusted estimate to non-negative

sanity:
## Symptom

On the node dashboard's peer detail page (`/peer/<addr>`), the **Response Time** chart's violet **"Peer-adjusted"** curve renders **below the x-axis** — i.e. a *negative* response time, which is physically meaningless.

## Root cause

`IsotonicEstimator::estimate_retrieval_time` clamps the **global** isotonic estimate to `>= 0` (regression can produce small negatives), but then **adds the per-peer EWMA adjustment _after_ that clamp**:

```rust
let global_estimate = global_estimate.max(0.0);   // clamped
... global_estimate + peer_adjustment.value()     // adjustment NOT re-clamped
```

A peer that is consistently faster / more reliable than the global fit has a **negative** adjustment. When its magnitude exceeds the (small) global estimate — most likely near distance 0, the left edge of the chart — the result goes negative. The targets here (response time, transfer rate, failure probability) are all physically non-negative.

This surfaces in two places, one cause:

- **Routing:** a negative `time_to_response_start` flows *unclamped* into `expected_total_time` in `predict_routing_outcome` (`router.rs`), making such a peer look impossibly fast and biasing selection toward it. Note the **failure** path right beside it *is* already clamped (`failure_estimate.clamp(0.0, 1.0)`) — the timing path simply missed the analogous clamp.
- **Dashboard:** the peer detail chart draws the peer-adjusted curve as `global_curve_y + adjustment` with no lower bound, so the line dips below the plot floor.

## Fix

1. **Root cause** (`isotonic_estimator.rs`): re-clamp the adjusted estimate to `>= 0` in `estimate_retrieval_time`. Harmless for the failure estimator (already clamped to `[0,1]` downstream); corrects the timing/rate paths.
2. **Display** (`server/home_page/estimator.rs`): clamp the drawn curve to the visible axis floor (`y_min`, which is `0` for these charts), mirroring the existing scatter-point clamp, so the curve never renders below the x-axis even though the chart computes `y + adj` independently of the router.

## Tests

- `test_peer_adjustment_cannot_produce_negative_estimate` — a strongly-negative per-peer adjustment no longer yields a negative estimate.
- `peer_adjusted_curve_never_renders_below_x_axis` — the violet curve's SVG y-coordinates stay at/above the plot floor.

All 26 tests in the touched modules pass; clean build, no warnings.

## Review note

This touches the **routing** surface (peer selection cost formula), so it warrants careful review of the behavioral change: peers with strong negative timing adjustments now have `time_to_response_start` floored at 0 (best possible) rather than negative (impossibly good).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/router/isotonic_estimator.rs
+++ b/crates/core/src/router/isotonic_estimator.rs
@@ -155,19 +155,29 @@ impl IsotonicEstimator {
         // Regression can sometimes produce negative estimates
         let global_estimate = global_estimate.max(0.0);
 
-        Ok(self
-            .peer_adjustments
-            .get(peer)
-            .map_or(global_estimate, |peer_adjustment| {
-                let should_use_peer_adjustment =
-                    peer_adjustment.effective_count >= MIN_POINTS_FOR_REGRESSION as f64;
-                global_estimate
-                    + if should_use_peer_adjustment {
-                        peer_adjustment.value()
-                    } else {
-                        0.0
-                    }
-            }))
+        let adjusted_estimate =
+            self.peer_adjustments
+                .get(peer)
+                .map_or(global_estimate, |peer_adjustment| {
+                    let should_use_peer_adjustment =
+                        peer_adjustment.effective_count >= MIN_POINTS_FOR_REGRESSION as f64;
+                    global_estimate
+                        + if should_use_peer_adjustment {
+                            peer_adjustment.value()
+                        } else {
+                            0.0
+                        }
+                });
+
+        // The per-peer EWMA adjustment is added *after* the global clamp above and
+        // can be negative (a peer that is consistently faster / more reliable than
+        // the global fit). Re-clamp so the per-peer estimate can never go below
+        // zero — these targets (response time, transfer rate, failure probability)
+        // are all physically non-negative, and a negative prediction would both
+        // distort routing cost formulas and render below the x-axis on the
+        // dashboard's "Peer-adjusted" curve. The failure path applies the same
+        // clamp downstream (see `predict_routing_outcome`).
+        Ok(adjusted_estimate.max(0.0))
     }
 
     pub(crate) fn len(&self) -> usize {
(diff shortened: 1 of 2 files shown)


sanity:
## Multi-model review — clean (Full tier: routing surface)

Ran a Full review because this touches the routing cost formula. All four reviewers returned **no actionable findings**.

**External model — Codex (`codex review --base origin/main`):** *"No actionable correctness issues were identified in the diff. The changes consistently clamp negative peer-adjusted estimates and update the dashboard rendering/tests accordingly."*

**Claude reviewers (3 independent lenses):**
- **Skeptical bug-hunt** — traced all callers of `estimate_retrieval_time` (`predict_routing_outcome`, `connect.rs:470` which already clamps, dashboard snapshot). Confirmed the new `.max(0.0)` is a no-op for the failure path (re-clamped to `[0,1]` at `router.rs:716`) and NaN handling is unchanged from the pre-existing `.max(0.0)` contract. No missed `y + adj` site.
- **Routing-math** — confirmed the bug was *worse* than the floor-the-line symptom: a negative `time_to_response_start` **sign-flipped** the penalty term `time_to_response_start * failure_estimate * failure_cost_multiplier` (`router.rs:775`), turning a reliability penalty into a *reward* for unreliable-but-negatively-adjusted peers. Flooring at 0 makes that term provably `>= 0`. Non-renegade transfer path is bit-identical; the renegade-blend path changes in the correct direction (blends from a sane 0 floor instead of a negative base).
- **Completeness** — all three charts (Failure / Response Time / Transfer Rate) share the single `draw_curve` clamp; per-op tabs pass `None` for the adjustment; the numeric "Prediction at Peer Location" card inherits the `estimate_retrieval_time` clamp; no off-the-top (upper-bound) risk for any chart.

Both new regression tests pass; all 26 tests in the touched modules pass locally; clean build, no warnings.

[AI-assisted - Claude]

Issue #3757, opened 2026-04-03
fix: node should always cache contract WASM on GET, regardless of return_contract_code


sanity:
## Problem

When a client sends a GET with `return_contract_code: false`, the node does not cache the contract WASM locally. This means:
- SUBSCRIBE fails with "contract WASM/parameters not cached locally"
- The node cannot validate incoming updates for the contract
- The node cannot host the contract for other peers

## Expected Behavior

`return_contract_code` should only control whether the WASM bytes are included in the response to the client (bandwidth optimization). The node should always fetch and cache the WASM internally so it can:
- Accept and process SUBSCRIBE requests
- Validate contract state updates
- Host the contract state for the network

## Current Workaround

Delta sets `return_contract_code: true` on all GET requests, but this wastes bandwidth sending the WASM to the client (which already has it embedded).

## Context

Found while debugging Delta subscription failures. The node log showed:
```
Cannot cache or subscribe to contract - no contract code in GET response
Rejecting SUBSCRIBE: contract WASM not cached locally
```

[AI-assisted - Claude]

Issue #2199, opened 2025-12-02
Add integration tests for NAT peer subscription scenarios


sanity:
## Problem

The NAT address handling code in subscribe/seeding operations lacks end-to-end integration tests that verify the full flow works correctly with actual NAT scenarios.

### Current Test Coverage

- ✅ Unit tests for `LiveTransactionTracker::active_transaction_count`
- ✅ Unit tests for connect operation NAT scenarios
- ✅ Unit tests for seeding manager subscriber operations
- ❌ No integration test for subscribe operation NAT address filling
- ❌ No end-to-end test showing NAT peer successfully subscribing through gateway

### Suggested Tests

1. **Subscribe NAT address filling test**
   - Peer behind NAT sends subscribe request with `PeerAddr::Unknown`
   - Gateway fills in observed address from `source_addr`
   - Verify subscriber is registered with correct address

2. **End-to-end NAT subscription test**
   - Set up gateway + NAT peer topology
   - NAT peer subscribes to contract
   - Verify updates are correctly routed back to NAT peer

3. **Edge case: source_addr is None**
   - What happens if `source_addr` is unexpectedly `None`?
   - Should log warning and handle gracefully

### Importance

**Medium priority** - The code paths are tested via unit tests and manual testing, but automated integration tests would catch regressions and document expected behavior.

### Context

- Discovered during Claude bot review of PR #2193
- Could use `freenet-test-network` crate for integration testing

[AI-assisted - Claude]

iduartgomez:
@claude can we close this? We have the 6 peer network test using docker tests. Or are we missing anything?

sanity:
# Issue Priority Assessment

Generated: 2026-02-22T14:38:26.082143896+00:00
Issue hash: `7654fcce`

## Value Assessment

Subscribe NAT reliability is a user-facing concern because the code path that fills in the remote address for a `SUBSCRIBE` request lives directly inside `SubscribeOp::load_or_init` and `process_message`—`source_addr` and `requester_addr` are captured as soon as the message arrives so that the interest manager can register the peer with the observed external socket before propagating updates (`crates/core/src/operations/subscribe.rs:556`). Without an end-to-end test for that flow, a regression could leave NAT peers registered with `PeerAddr::Unknown` (or dropped entirely), meaning they would silently stop receiving updates even though their subscription succeeded locally; given that most users run behind consumer NATs, that would degrade their experience significantly.

[...]

PR #4540, opened 2026-06-21
test(nat): use blocking GET+subscribe so non-host nat-peer can subscribe (#4524)

### LINKED ISSUE #4524: flaky: test_nat_peer_remote_subscription_receives_update fast-startup race ('contract not cached locally')
## Problem

`test_nat_peer_remote_subscription_receives_update` (a multi-threaded, real-timing `#[freenet_test]` NAT integration test) fails intermittently with a fast-startup-race signature:

```
contract <KEY> not cached locally
```
observed at ~7–9s into the run — the subscribing peer reaches SUBSCRIBE before the contract has propagated/cached on the target.

## Evidence (Linux/Docker loopback)

Surfaced while debugging PR #4509 (#4363 PUT terminus guard). Tallied across runs on the **clean branch HEAD with no #4363 change applied**:

- clean branch HEAD (no fix): **1 fail / 3 pass** (4 runs)
- with the #4509 replication fix: **2 fail / 4 pass** (6 runs)

Identical failure signature, same rate. The #4509 change is strictly additive (it only *adds* a replication copy) and cannot produce a "not cached" failure, so the flakiness is **pre-existing and independent of #4363/#4509**.

## Notes

Per AGENTS.md a flaky test is a separate issue to track, not to weaken or `#[ignore]` on an unrelated PR — hence this issue rather than touching it in #4509. The likely root cause is a startup/timing race between contract propagation and the subscriber issuing SUBSCRIBE; a fix should make the test wait for the contract to be cached on the target before subscribing (or harden the NAT-subscription path against the race) rather than bump timeouts.

[AI-assisted]

sanity:
## Problem

`test_nat_peer_remote_subscription_receives_update` (a multi-threaded, real-timing `#[freenet_test]` NAT integration test) failed intermittently (~25% per #4524) with a fast-startup-race signature:

```
contract <KEY> not cached locally
```

observed ~7-9s into the run.

## Root cause (test-design issue, not a product bug)

The bare `ContractRequest::Subscribe` client-event handler **deliberately** rejects a subscribe when the contract WASM/state is not cached on the **local** node (`client_events.rs`, guard added in #3757 so a node can't be "subscribed but can't apply updates"). That guard explicitly exempts `GET+subscribe=true` and `PUT+subscribe=true` ("PUT the contract or GET the contract before subscribing"), because those operations fetch/provide the WASM.

`nat-peer` is pinned half a ring from the contract — the *provable non-host* — and never PUTs the contract itself. The only way it can hold the contract locally is to fetch it, and it **cannot rely on PUT replication**: the originator PUT on `host-peer` (already at the contract location) finalizes locally and replicates exactly one hop toward the contract neighbourhood (`relay_put_replicate_forward`, #4509), which lands on the *gateway* (closer to the contract than `nat-peer`), **not** on `nat-peer`. So a bare SUBSCRIBE on `nat-peer` raced a contract presence that, for the farthest peer, frequently never arrived at all.

This was confirmed empirically: an earlier attempt that polled `NodeDiagnostics` for local contract presence before subscribing **timed out** — the contract simply never appeared in `nat-peer`'s store within 120s on a failing run, then the bare SUBSCRIBE was rejected "not cached locally". The original test's ~75% pass rate came from runs where the one-hop replication happened to land on `nat-peer` instead of the gateway (a startup-connection-timing coin flip).

## Fix

`nat-peer` issues `GET` with `subscribe=true` **and** `blocking_subscribe=true` instead of a bare SUBSCRIBE:

- The GET routes *toward* the contract (so it reliably resolves on gateway/host-peer), fetches and caches the body on `nat-peer`.
- The GET driver registers the subscription through the **same relay path** a bare SUBSCRIBE would use — the wire `SubscribeMsg::Request` still carries no address, so each hop (`nat-peer → gateway → host-peer`) still fills `upstream_addr` from the observed transport `source_addr` and registers the previous hop as a downstream subscriber. **The NAT observed-address registration the test exists to prove (#2199) is still exercised end-to-end.**
- `blocking_subscribe=true` is load-bearing: the GET driver awaits the subscribe child inline (`get::op_ctx_task::maybe_subscribe_child`) before publishing the `GetResponse`, so when the response arrives the downstream chain is fully registered. This lets us drop the prior fixed 3 s "settle" sleep entirely.

This mirrors the established `operations.rs` auto-subscribe-via-GET pattern. No timeout was bumped, nothing was `#[ignore]`-d, and no assertion was weakened.

## Testing

Verified deterministic on Linux (full `127.x.x.x` loopback): the test now passes on **every** run (~9s each) where it previously failed ~25% with a 120s+ "not cached locally" timeout.

- Blocking GET+subscribe version: **10/10 passes**, ~9s each.
- (An interim non-blocking GET version also passed 9/9, but review found a residual race — see below — so it was upgraded to blocking.)

## Review

- **codex review --base origin/main** run three times across iterations:
  - Initial polling approach: clean.
  - Non-blocking GET version: codex flagged a real **[P2]** — `make_get(subscribe=true)` defaults `blocking_subscribe=false`, so the GET driver spawns the subscribe fire-and-forget and could deliver `GetResponse` before the downstream chain was registered, letting the UPDATE fire too early. **Addressed** by switching to `make_get_with_blocking(..., blocking_subscribe=true)` (and removing the now-redundant settle sleep).
  - Final blocking version: **clean** ("I did not find a discrete correctness issue introduced by the patch").
- Gemini CLI unavailable (free-tier client retired — `IneligibleTierError`); codex served as the external non-Claude pass.
- Four-lens self-review (code-first / skeptical / testing / big-picture): the only finding was a doc comment naming the wrong subscribe helper (`auto_subscribe_on_get_response` vs `maybe_subscribe_child`), now corrected. Polling/retry loop is deadline-bounded, no busy-spin, retry is idempotent (subscribe guarded by `is_subscribed`).

Closes #4524

[AI-assisted - Claude]

### DIFF
--- a/crates/core/tests/nat_subscription.rs
+++ b/crates/core/tests/nat_subscription.rs
@@ -55,6 +55,40 @@
 //! introspection hook into the gateway's subscriber set (there is none — see
 //! below).
 //!
+//! ## Why `nat-peer` GETs (with `subscribe=true`) instead of a bare SUBSCRIBE (#4524)
+//!
+//! A bare `ContractRequest::Subscribe` is rejected by the client-event handler
+//! unless the contract WASM/state is already cached on the *local* node
+//! (`client_events.rs`: "Rejecting SUBSCRIBE: contract WASM not cached
+//! locally" — the guard added in #3757 so a node can't be "subscribed but
+//! can't apply updates"). That guard explicitly exempts `GET+subscribe=true`
+//! and `PUT+subscribe=true` because those operations inherently fetch/provide
+//! the WASM ("PUT the contract or GET the contract before subscribing").
+//!
+//! `nat-peer` is the *provable non-host* (half a ring away) and never PUTs the
+//! contract itself, so the only way it can hold the contract locally is to
+//! fetch it. It cannot rely on PUT replication: the originator PUT on
+//! `host-peer` finalizes at `host-peer` (already at the contract location) and
+//! replicates exactly **one hop** toward the contract neighbourhood
+//! (`relay_put_replicate_forward`, #4509), which lands on the *gateway* (closer
+//! to the contract than `nat-peer`) — not on `nat-peer`. So a bare SUBSCRIBE on
+//! `nat-peer` raced contract presence that, for the farthest peer, often never
+//! arrives at all; that is the flake tracked in #4524 (a "contract not cached
+//! locally" rejection ~7-9s in).
+//!
+//! The deterministic fix is to use the supported non-host path: `nat-peer`
+//! issues `GET` with `subscribe=true`. The GET routes *toward* the contract
+//! (so it reliably resolves on `gateway`/`host-peer`), fetches and caches the
+//! body on `nat-peer`, and the GET driver's `auto_subscribe_on_get_response`
+//! then registers the subscription through the SAME relay path a bare SUBSCRIBE
+//! would use — the wire `SubscribeMsg::Request` still carries no address, so
+//! each hop still fills `upstream_addr` from the observed transport
+//! `source_addr` and registers the previous hop as a downstream subscriber.
+//! The NAT observed-address registration the test exists to prove is therefore
+//! still exercised end-to-end; only the unreliable "wait for PUT to replicate
+//! to the farthest peer" precondition is removed. No timeout was bumped and no
+//! assertion was weakened.
+//!
 //! ## Why we assert on the routed notification rather than the subscriber set
 //!
 //! There is no test-visible way to read a node's downstream-subscriber set.
@@ -78,7 +112,7 @@
 
 use anyhow::bail;
 use freenet::dev_tool::Location;
-use freenet::test_utils::{self, TestContext, make_put, make_subscribe, make_update};
+use freenet::test_utils::{self, TestContext, make_get_with_blocking, make_put, make_update};
 use freenet_macros::freenet_test;
 use freenet_stdlib::{
     client_api::{ContractResponse, HostResponse, WebApi},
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary (multi-model + four-lens)

**External non-Claude pass — codex (`codex review --base origin/main`):** run on the final blocking-GET content → **clean**, no findings. On the intermediate non-blocking-GET revision codex caught a real **[P2]** subscribe-vs-update race (`blocking_subscribe` defaulted false, so `GetResponse` could precede subscription registration); fixed by switching to `make_get_with_blocking(..., blocking_subscribe=true)` and dropping the redundant 3 s settle sleep. Re-reviewed clean after the fix.

**Gemini CLI:** unavailable this session (`IneligibleTierError` — the free-tier Gemini Code Assist client has been retired). codex was the external non-Claude reviewer; the four-lens pass below substitutes for the second external model.

**Four-lens self-review (final content):**
- *code-first*: found one doc inaccuracy (helper named `auto_subscribe_on_get_response` where the `subscribe=true` path actually uses `maybe_subscribe_child`) — corrected.
- *skeptical*: GET-then-update flow is deterministic; retry loop is deadline-bounded, no busy-spin, retry is idempotent (post-GET subscribe guarded by `is_subscribed`).
- *testing*: 10/10 deterministic passes on Linux (~9s each) vs. the previous ~25% failure with a 120s+ "not cached locally" timeout. Core #2199 update-notification assertion preserved.
- *big-picture*: no timeout bump, no `#[ignore]`, no sleep used as the fix mechanism, no weakened assertion; mirrors the established `operations.rs` auto-subscribe-via-GET pattern.

All findings addressed. Test-only change.

[AI-assisted - Claude]

PR #4543, opened 2026-06-21
refactor: thread ModuleCacheMetrics as Arc instead of process-global static

### LINKED ISSUE #4488: refactor: pass ModuleCacheMetrics as Arc instead of process-global static
## Background

PR #4472 introduced `MODULE_CACHE_METRICS` as a process-global `static` to share metrics between two components that live on opposite sides of an architectural boundary:

- **Writers** (`ModuleCache::insert` / `evict_to_budget` / `remove`) — owned by `RuntimePool`, behind the contract-handler channel.
- **Reader** (`ring.rs:1074`) — the `Ring` snapshot task, which has no channel or reference to `RuntimePool`.

This mirrors the existing `TRANSPORT_METRICS` static and is correct given the single-`RuntimePool`-per-process production invariant. The static was the right call for the PR scope.

## Problem

Process-global statics make test isolation harder: labeled caches in different unit tests share the same gauge pair, so parallel tests can observe each other's metric state. This is the same problem class that `TRANSPORT_METRICS` has.

## Proposed refactor

Replace the static with an `Arc<ModuleCacheMetrics>` threaded through the call graph:

1. Construct `Arc<ModuleCacheMetrics>` at node startup (in `Node::build` or equivalent).
2. Pass it into `RuntimePool::new` → stored on each `ModuleCache` via `ModuleCache::with_label`.
3. Store the same `Arc` in `RouterSnapshotInfo` or `Ring` so the snapshot task can call `.snapshot()` from it.

Estimated scope: ~4 function signature changes (`RuntimePool::new`, `ModuleCache::with_label`, snapshot-state construction, and one or two call sites).

## Acceptance criteria

- [ ] `MODULE_CACHE_METRICS` static removed.
- [ ] `ModuleCacheMetrics` constructed once and threaded as `Arc` through `RuntimePool` and the `Ring` snapshot path.
- [ ] Existing metric values / Prometheus labels unchanged (no dashboard breakage).
- [ ] Unit tests for `ModuleCache` create isolated metric instances without cross-talk.

## References

- PR #4472 (original implementation, discussion at `module_cache.rs:267`)
- `TRANSPORT_METRICS` — the existing static this mirrors; a follow-up to that one would be a natural pairing.

sanity:
## Problem

`MODULE_CACHE_METRICS` (introduced in #4472) was a process-global `static` used to share compiled-WASM module-cache occupancy/eviction telemetry between two components on opposite sides of the contract-handler channel boundary: the **writers** (`ModuleCache` methods, owned by `RuntimePool`) and the **reader** (the `Ring` snapshot task at `ring.rs`).

Process-global statics make test isolation harder — labeled caches in different unit tests share the same gauge pair, so parallel tests can observe each other's metric state. This is the same problem class as `TRANSPORT_METRICS`. (#4488)

## Approach

Replace the static with a per-node `Arc<ModuleCacheMetrics>` threaded through the call graph, exactly as the issue proposed:

- Construct the `Arc` **once** in `Ring::new` (one sink per node).
- `RuntimePool::new` reaches it via the already-available `op_manager.ring` and clones it into both labeled `ModuleCache`s.
- The `emit_router_snapshot_telemetry` task reads the same `Arc` off `Ring`.

Key insight: `RuntimePool::new` already receives `op_manager: Arc<OpManager>`, and `op_manager.ring` is exactly where the snapshot task lives — so the `Arc` is shared through the existing object graph with **no new `RuntimePool::new` parameter** and no channel crossing. `Ring` is constructed before `RuntimePool`, so it's the natural owner.

`ModuleCache` now stores `Option<Arc<ModuleCacheMetrics>>` (`None` for unlabeled / test caches, which never published anyway) and routes publishing through two small `record_occupancy`/`add_evictions` helpers; `with_label` takes the optional sink.

**No behavior change:** metric values, Prometheus labels, and snapshot fields are identical. This is purely a wiring change. Both labeled caches share one sink and are routed apart by their `"contract"` / `"delegate"` label, exactly as before.

## Testing

- Rewrote the existing eviction test (`evicting_contract_cache_publishes_to_injected_metrics`): it now constructs its own metrics `Arc` and asserts **exact absolute values** (`contract_evictions_total == 1`, budget published on construction, delegate gauges stay zero). Previously it could only assert a strict increase against the shared global because concurrent tests polluted it — the new isolation is the whole point of the refactor, and the test demonstrates the global is gone.
- Added `cache_without_metrics_sink_is_a_no_op` covering the `None` path.
- Updated the `runtime_pool_sizes_caches_by_byte_budget` source-pinning test for the new multi-line `with_label(..., Some(metrics))` call shape (collapses whitespace; asserts budget + label independently so it isn't brittle to the `Arc` arg).
- `cargo build -p freenet`, `cargo clippy -p freenet --lib` (default features), and the full `wasm_runtime::module_cache::tests` + `wasm_runtime::tests::cache` modules all pass.

Note on CI gap: this is a developer-experience hygiene refactor with no user-visible behavior change, so no production bug to reproduce. The acceptance criterion ("unit tests create isolated metric instances without cross-talk") is met by the injected-`Arc` test above.

Updated doc comments that referenced the removed static (broadcast-queue / background-task-health "mirror" notes, `RouterSnapshotInfo`, the re-export in `node.rs`) to point at `TRANSPORT_METRICS` and note the #4488 change.

## Fixes

Closes #4488

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -978,19 +978,32 @@ mod executor_pin_tests {
             .split("pub async fn new(")
             .nth(1)
             .expect("RuntimePool::new must exist")
+            // Take the first chunk of the function body. Whitespace is collapsed
+            // so the assertions below survive line-wrapping / reformatting of the
+            // (now multi-line, metrics-threaded) cache construction.
             .split("\n    ")
-            .take(60)
-            .collect::<String>();
+            .take(80)
+            .collect::<String>()
+            .split_whitespace()
+            .collect::<Vec<_>>()
+            .join(" ");
         assert!(
             body.contains("config.module_cache_budget_bytes"),
             "RuntimePool::new must size caches from config.module_cache_budget_bytes"
         );
+        // The contract cache is built from `contract_cache_budget` with the
+        // "contract" label; the delegate cache from `delegate_cache_budget` with
+        // the "delegate" label. We assert each piece independently (rather than a
+        // single literal call string) so threading the metrics `Arc` through
+        // `with_label` (#4488) doesn't make this pin brittle.
         assert!(
-            body.contains("ModuleCache::with_label(contract_cache_budget, \"contract\")"),
+            body.contains("ModuleCache::with_label( contract_cache_budget, \"contract\"")
+                || body.contains("ModuleCache::with_label(contract_cache_budget, \"contract\""),
             "RuntimePool::new must build the contract cache from the contract byte budget"
         );
         assert!(
-            body.contains("ModuleCache::with_label(delegate_cache_budget, \"delegate\")"),
+            body.contains("ModuleCache::with_label( delegate_cache_budget, \"delegate\"")
+                || body.contains("ModuleCache::with_label(delegate_cache_budget, \"delegate\""),
             "RuntimePool::new must build the delegate cache from its own (smaller) budget"
         );
         assert!(
--- a/crates/core/src/contract/executor/runtime/pool.rs
+++ b/crates/core/src/contract/executor/runtime/pool.rs
@@ -148,12 +148,24 @@ impl RuntimePool {
         let delegate_cache_budget = (contract_cache_budget
             / crate::wasm_runtime::DELEGATE_MODULE_CACHE_BUDGET_DIVISOR)
             .max(1);
-        let shared_contract_modules: SharedModuleCache<ContractKey> = Arc::new(Mutex::new(
-            ModuleCache::with_label(contract_cache_budget, "contract"),
-        ));
-        let shared_delegate_modules: SharedModuleCache<DelegateKey> = Arc::new(Mutex::new(
-            ModuleCache::with_label(delegate_cache_budget, "delegate"),
-        ));
+        // Per-node telemetry sink shared with the `Ring` snapshot task (#4440 /
+        // #4488). The caches publish occupancy/eviction into it; the snapshot
+        // task reads the same `Arc` via `Ring`. Threading it (rather than a
+        // process-global) keeps the gauges per-node. Both caches share one sink;
+        // they're routed apart by their `"contract"` / `"delegate"` label.
+        let module_cache_metrics = op_manager.ring.module_cache_metrics();
+        let shared_contract_modules: SharedModuleCache<ContractKey> =
+            Arc::new(Mutex::new(ModuleCache::with_label(
+                contract_cache_budget,
+                "contract",
+                Some(module_cache_metrics.clone()),
+            )));
+        let shared_delegate_modules: SharedModuleCache<DelegateKey> =
+            Arc::new(Mutex::new(ModuleCache::with_label(
+                delegate_cache_budget,
+                "delegate",
+                Some(module_cache_metrics),
+            )));
         // Shared delegate-context cache so a prompt round-trip routed to a
         // different pool executor still finds its `ctx.write()` blob.
         let shared_delegate_contexts = crate::wasm_runtime::new_delegate_context_cache();
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -53,7 +53,9 @@ pub(crate) use network_bridge::{
 };
 // Re-export the UPDATE-broadcast stream-assembly telemetry global (#4440) so the
 // `Ring` snapshot task can read it (the broadcast queue lives behind the private
-// `network_bridge` module). Mirrors `crate::wasm_runtime::MODULE_CACHE_METRICS`.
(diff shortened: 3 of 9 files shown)


sanity:
## Review summary (debug-and-fix, serial path)

This is a Light/mechanical-refactor-tier change with no behavior change, but the wiring touches a cross-component boundary so I gave it a full pass.

### External model pass
- **Codex** (`codex review --base origin/main`, codex-cli 0.141.0): **No actionable correctness issues found.** "The changes consistently thread per-node module cache metrics through Ring and RuntimePool, and call sites were updated for the new `ModuleCache::with_label` signature."
- **Gemini**: attempted (`gemini --approval-mode plan`) but **unavailable** — `IneligibleTierError: This client is no longer supported for Gemini Code Assist for individuals` (free-tier client deprecated, not a transient outage). Per `multi-model-review.md`, the Codex pass satisfies the independent non-Claude requirement; I did not need the Claude-lens fallback since one external model ran.

### Claude review lenses (run serially; this agent cannot spawn subagents)
- **Code-first**: impl matches stated intent — single `Arc` constructed in `Ring::new`, cloned into both labeled caches via `op_manager.ring.module_cache_metrics()`, read back off `ring.module_cache_metrics`. No intent/impl mismatch.
- **Skeptical (the failure mode for this change)**: verified there is **exactly one** production `ModuleCacheMetrics::new()` (ring.rs:352) — no accidental second instance that would make the snapshot read stale/zero values. The contract and delegate caches share the same `Arc` (clone of one allocation). Construction-time `record_occupancy(0,0,budget)` preserved (same clamped budget, just moved after struct init — gauges are independent of cache internals).
- **Testing**: new `evicting_contract_cache_publishes_to_injected_metrics` asserts exact absolute values on an isolated injected sink (previously could only assert a strict increase against the polluted global); `cache_without_metrics_sink_is_a_no_op` covers the `None` path; source-pin test updated for the new call shape. End-to-end Ring↔RuntimePool sharing is verified by inspection (full-Node construction is too heavy for a unit test, per existing code comments).
- **Big-picture**: all four issue acceptance criteria met; no tests removed; doc comments referencing the removed static updated. The `TRANSPORT_METRICS` follow-up the issue mentions is explicitly out of scope.

### Local checks
`cargo build -p freenet` ✓ · `cargo clippy -p freenet --lib` (default features) ✓ · `wasm_runtime::module_cache::tests` 18 passed ✓ · `wasm_runtime::tests::cache` ✓ · `runtime_pool_sizes_caches_by_byte_budget` ✓

(Note: `cargo clippy -p freenet --lib --all-features` fails with pre-existing, unrelated errors in `tracing/metrics_client.rs` under the metrics feature — present on origin/main, not touched by this PR. CI's lib clippy uses default features.)

No findings to address. Proceeding to CI gate.

[AI-assisted - Claude]

sanity:
## Re-review after rebase onto main (0.2.81)

Rebased onto `origin/main` (now `609a2194`, release 0.2.81). The rebase surfaced a **semantic conflict with PR #4534** (backpressure-aware migration admission), which landed a *second* reader of the module-cache metrics: `contract_cache_occupancy_pct()` read the now-removed `MODULE_CACHE_METRICS` global from the placement-migration admission gate in `node.rs`.

**Conflict resolution (re-reviewed, not a rubber-stamp of HEAD):**
- `contract_cache_occupancy_pct` now takes `&ModuleCacheMetrics` instead of reading the global; the `node.rs` caller passes `&op_manager.ring.module_cache_metrics()` — the **same** per-node `Arc` the caches publish into (reached the same way `RuntimePool::new` reaches it). This keeps the migration gate consistent with the refactor.
- Kept both test sets in `module_cache.rs` (main's `occupancy_pct_*` arithmetic tests + this PR's `cache_without_metrics_sink_is_a_no_op`).
- Note: in multi-node simulations the gate now reads **per-node** occupancy instead of the old process-global aggregate — strictly more correct for the gate's documented intent; no production behavior change (single node per process).

**Re-verification on the rebased HEAD (`90898763`):**
- **Codex** (`codex review --base origin/main`): re-run on the new content — **no actionable correctness issues.**
- Claude lenses re-applied to the delta: confirmed the migration gate and the snapshot task read the same single `Arc`; `&temporary-Arc` borrow is sound (build confirms); migration-admission threshold tests unaffected.
- Local: `cargo build -p freenet` ✓ · `cargo clippy -p freenet --lib` ✓ · `wasm_runtime::module_cache::tests` 22 passed ✓ · `node::migration_admission_tests` 3 passed ✓ · `runtime_pool_sizes_caches_by_byte_budget` ✓

(Gemini still unavailable — free-tier client deprecated, `IneligibleTierError`. Codex remains the independent non-Claude pass.)

[AI-assisted - Claude]

PR #4544, opened 2026-06-21
fix(ring): plumb shutdown CancellationToken into spawned background loops

### LINKED ISSUE #4278: ring.rs spawned tasks: plumb shutdown signal for interruptible loops
## Context

Skeptical reviewer of #4270 flagged that \`governance_reaper_loop\` (added in #4270) uses an uninterruptible \`tokio::time::sleep\` for its initial random offset:

\`\`\`rust
let initial_delay = Duration::from_secs(GlobalRng::random_range(30u64..=90u64));
tokio::time::sleep(initial_delay).await;  // not interruptible
\`\`\`

Per \`.claude/rules/code-style.md\`: *"Backoff sleeps MUST be interruptible: Use tokio::select! to race sleep against cancellation signal (Notify, CancellationToken). Plain tokio::time::sleep() in retry loop is PROHIBITED unless <1s"*

## The bigger picture

This is not a one-off in \`governance_reaper_loop\`. The same pattern exists in:

- \`Ring::interest_heartbeat\` — same shape, same uninterruptible initial sleep + main loop.
- \`Ring::connection_maintenance\` — main loop has \`tokio::select!\` already but I haven't verified every internal sleep.

The root cause is that **none of these spawned tasks have a shutdown signal**. \`Ring\` doesn't expose a \`CancellationToken\` or \`Notify\` that the spawned tasks could race against. So even if we wanted to fix \`governance_reaper_loop\` in isolation, there's no signal to race against.

## Proposal

1. Add a \`shutdown: tokio_util::sync::CancellationToken\` field on \`Ring\` (or equivalent — \`Arc<tokio::sync::Notify>\` works too).
2. Pass it down through \`Ring::new\` and store it as a field.
3. Trigger it from \`Ring::shutdown()\` or wherever the node initiates shutdown today (graceful or panic).
4. Race every long sleep in spawned tasks against \`shutdown.cancelled()\` via \`tokio::select!\`.

The work is small per call site but spans 3+ loops. Better to do it as one focused PR than to retrofit per-feature.

## Acceptance criteria

- [ ] \`Ring::shutdown\` (or equivalent) exists and triggers a single cancellation token.
- [ ] Every \`tokio::time::sleep\` in spawned tasks in \`ring.rs\` lasting more than 1s is wrapped in \`tokio::select!\` against the token.
- [ ] Test: spawning + shutdown returns cleanly within \`tick_interval\` instead of waiting for the longest sleep to finish.

## Related

- #4276 — split topology meter (governance scope)
- #4277 — migrate remaining \`Instant::now()\` to \`TimeSource\` (governance scope)
- #4270 — adds the third instance of this pattern

[AI-assisted - Claude]

sanity:
## Problem

The long-lived background tasks spawned in `Ring::new` had no shutdown signal:

- `governance_reaper_loop`, `interest_heartbeat`, `connection_maintenance`, `refresh_router`, `emit_router_snapshot_telemetry`, `emit_subscription_state_telemetry`, `recover_orphaned_subscriptions`, `sweep_get_subscription_cache`, `contract_directed_connects`, and `register_topology_snapshots_periodically`.

The `BackgroundTaskMonitor` only *observes* their exits — it consumes their `JoinHandle`s in `wait_for_any_exit` but never aborts them (dropping a `JoinHandle` does not abort the task). And several loops block in long `tokio::time::sleep` / `interval.tick()` calls — up to **5 minutes** for `interest_heartbeat`'s main interval. So on node teardown those tasks kept running until their longest outstanding sleep elapsed. There was **zero `CancellationToken`** anywhere in `crates/core`.

This also violated `.claude/rules/code-style.md`: *"Backoff sleeps MUST be interruptible: Use `tokio::select!` to race sleep against cancellation signal … Plain `tokio::time::sleep()` in retry loop is PROHIBITED unless <1s"*.

User impact is on shutdown paths only (graceful stop, test teardown, rolling restart): the process hangs for the duration of the longest sleep. It compounds as more governance tasks are added under epic #4296.

## Approach

Add a single `shutdown: CancellationToken` field on `Ring`, created up front in `Ring::new` and shared (cloned) into every spawned loop. `CancellationToken` is used rather than `Notify` because it is **level-triggered** — once cancelled, every present and future `cancelled()` await returns immediately, so there is no missed-wakeup race for a task that is between sleeps when shutdown fires.

- Every long sleep / `interval.tick()` is raced against the token through one small `sleep_or_shutdown` helper (a single `tokio::select!` chokepoint, with documented cancellation-safety and `biased;` justification per the code-style rule). The loop returns when the helper reports cancellation.
- `Ring::trigger_shutdown()` cancels the token; it is fired from `ShutdownTeardown::drop` — the single node-teardown chokepoint that already runs on **every** `run_node` exit path (happy path and early `?`/return). This is the natural place: it sits right next to the existing detached-task aborts.
- `connection_maintenance` already terminated via the `OpManagerState::Detached` path, but the token fires *before* the `OpManager` Arc is dropped, so it now also checks `is_cancelled()` at the top of its loop and adds a shutdown arm to its two steady-state `deterministic_select!` tick blocks so the up-to-60s tick is interruptible.

Scope is deliberately limited to the `ring.rs` spawned tasks (and the teardown wiring). The broader governance work is epic #4296.

## Testing

New tests (would fail before the fix):

- `sleep_or_shutdown_tests` — the helper directly: completes normally (`false`), short-circuits when cancelled mid-wait (`true`, well under the 5-min sleep), and short-circuits immediately on an already-cancelled token (level-triggered, no missed-wakeup).
- `refresh_router_tests::refresh_router_returns_promptly_on_shutdown` — a `#[tokio::test(start_paused = true)]` regression test: with the token pre-cancelled, `refresh_router` must return within a 1s virtual-time `timeout`. Before the fix there was no token and the only loop exit was the 5-minute `interval.tick()`, so the timeout would fire and the test would fail. This maps directly to the issue's acceptance criterion ("shutdown returns cleanly within tick_interval instead of waiting for the longest sleep").

Validation:

- `cargo fmt`, `cargo check -p freenet --tests`, and CI-equivalent `cargo clippy -p freenet --bins --lib --locked -- -D warnings` all clean.
- `cargo test -p freenet --lib -- ring::` → 520 passed, 0 failed.

### Why didn't CI catch this?

There was no test asserting bounded shutdown latency for the Ring background tasks — the gap is closed by the `start_paused` regression test above, which any contributor can run deterministically (no network, no real time).

## Acceptance criteria

- [x] `Ring::trigger_shutdown` exists and triggers a single cancellation token.
- [x] Every `tokio::time::sleep` / long `interval.tick()` in `ring.rs` spawned tasks lasting more than 1s is raced against the token via `tokio::select!`.
- [x] Test: with shutdown signalled, a loop returns promptly (within a tick) instead of waiting for the longest sleep.

## Note (pre-existing, out of scope)

`cargo clippy --tests` surfaces an unrelated pre-existing `undocumented_unsafe_blocks` error in `crates/core/src/bin/commands/secrets_cmd.rs:1849` (test code from #4506). CI's clippy step runs without `--tests`, so it does not gate this PR; left untouched to keep the diff scoped to #4278.

Closes #4278

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2086,6 +2086,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-tungstenite 0.27.0",
+ "tokio-util",
  "toml 1.1.2+spec-1.1.0",
  "tower",
  "tower-http 0.7.0",
@@ -7007,6 +7008,7 @@ dependencies = [
  "bytes 1.11.1",
  "futures-core",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ reqwest = { version = "0.12", default-features = false, features = ["json", "str
 tokio = { version = "1", features = ["full"] }
 tokio-stream = "0.1.18"
 tokio-tungstenite = "0.27.0"
+tokio-util = "0.7"
 tower-http = { version = "0.7", features = ["fs", "trace"] }
 
 # Crypto
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -88,6 +88,7 @@ tempfile = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["fs", "macros", "rt-multi-thread", "signal", "sync", "process", "tracing", "time", "test-util"] }
 tokio-tungstenite = { workspace = true }
+tokio-util = { workspace = true, features = ["rt"] }
 tower-http = { workspace = true }
 ulid = { workspace = true }
 zeroize = { workspace = true }
--- a/crates/core/src/node/p2p_impl.rs
+++ b/crates/core/src/node/p2p_impl.rs
@@ -70,10 +70,14 @@ pub(crate) struct NodeP2P {
 /// and the redb file lock held for the process lifetime (issue #4401).
 ///
 /// On drop it:
-///   1. aborts the stored detached task handles (executor + client-events, plus
+///   1. triggers the ring's background-task shutdown token so the long-lived
+///      `Ring::new` loops (governance reaper, interest heartbeat, connection
+///      maintenance, telemetry) stop promptly instead of waiting out their
+///      longest sleep (issue #4278),
+///   2. aborts the stored detached task handles (executor + client-events, plus
 ///      the session-actor / result-router / initial-join / aggressive-connect
 ///      tasks once they exist), and
-///   2. clears the ring's redb `Storage` clones via `clear_redb_storage()`.
+///   3. clears the ring's redb `Storage` clones via `clear_redb_storage()`.
 ///
 /// ## Exhaustive redb `Arc<Database>` holder set (verified for #4401)
 ///
@@ -120,6 +124,16 @@ impl Drop for ShutdownTeardown {
             return;
         }
         self.fired = true;
+        // Signal the Ring's long-lived background tasks (governance reaper,
+        // interest heartbeat, connection maintenance, telemetry loops, etc.)
+        // to stop. They race this token against their sleeps via
+        // `tokio::select!`, so a graceful shutdown returns promptly instead of
+        // waiting up to ~5 minutes for the longest outstanding sleep to elapse.
+        // Unlike the abort handles below, these tasks are only *observed* by
+        // the BackgroundTaskMonitor (their JoinHandles are consumed there, not
+        // aborted), so the cancellation token is the mechanism that actually
+        // stops them. See issue #4278.
+        self.ring.trigger_shutdown();
         for abort in &self.aborts {
             abort.abort();
         }
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1619,12 +1619,27 @@ impl Ring {
 
                     let op_manager_clone = op_manager.clone();
                     let contract_key = contract;
+                    let task_shutdown = shutdown.clone();
 
                     GlobalExecutor::spawn(async move {
-                        tokio::time::sleep(Duration::from_millis(jitter_ms)).await;
-                        // Guard ensures complete_subscription_request is called even on panic
+                        // Guard ensures complete_subscription_request is called even on panic.
+                        // Created BEFORE the jitter sleep so an early shutdown return below
+                        // still clears the `mark_subscription_pending` flag set above — the
+                        // guard's Drop marks the request failed (#4278).
                         let guard =
                             SubscriptionRecoveryGuard::new(op_manager_clone.clone(), contract_key);
+
+                        // The jitter can be up to 15s; bail (dropping the guard, which
+                        // completes the pending request as failed) before doing any
+                        // renewal work if the node is shutting down (#4278).
+                        if sleep_or_shutdown(
+                            &task_shutdown,
+                            tokio::time::sleep(Duration::from_millis(jitter_ms)),
+                        )
+                        .await
+                        {
+                            return;
+                        }
 
                         let instance_id = *contract_key.id();
                         // Renewal driver: same machinery as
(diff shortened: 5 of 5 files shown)


sanity:
## Review summary (Full tier — concurrency/async/shutdown surface)

**External model pass:**
- **Codex** (`codex review --base origin/main`): ran twice — once on the initial commit and again on the updated HEAD after the review-follow-up commit. Both clean: *"The shutdown token is propagated consistently through the long-lived ring tasks … I did not identify any actionable correctness issues in the diff."*
- **Gemini** (`gemini-cli`): unavailable — `IneligibleTierError: This client is no longer supported for Gemini Code Assist for individuals`. The free tier was discontinued. Codex is the live external non-Claude reviewer, so independent review did run.

**Claude review lenses** (run serially — this agent cannot spawn subagents):
- *Skeptical:* confirmed no new "unexpected exit" risk — `sleep_or_shutdown` returns `false` (carry on) during normal operation, so loops behave exactly as before until teardown. `trigger_shutdown` fires from `ShutdownTeardown::drop`, which runs only *after* `run_node`'s exit-reason `select!` resolved, so a loop returning post-shutdown is never misread as the fatal cause.
- *Code-first/completeness:* found one >1s spawned-task sleep the first commit missed — the 0–15s renewal jitter inside `recover_orphaned_subscriptions`. Made it interruptible and moved the `SubscriptionRecoveryGuard` ahead of the sleep so the new early-return path still clears the `mark_subscription_pending` flag (fixes a latent leak). Addressed in 49c9233b.
- *Testing:* acceptance-criterion test (`refresh_router_returns_promptly_on_shutdown`, `start_paused`) fails without the token (timeout on the 5-min tick) and passes with it; helper unit tests cover completes / cancelled-mid-wait / already-cancelled.
- *Big-picture:* scope stays in `ring.rs` + teardown wiring (epic #4296 work deliberately excluded). Updated the `#4292` teardown-parking comment to reflect the new token exit. `biased;` in `sleep_or_shutdown` documented per `.claude/rules/code-style.md` (no hot arm to starve).

**Validation:** `cargo fmt` clean; CI-equivalent `cargo clippy -p freenet --bins --lib --locked -- -D warnings` clean; `cargo test -p freenet --lib -- ring::` → 520 passed / 0 failed.

[AI-assisted - Claude]

sanity:
Rebased onto latest `main` (HEAD now `26fc0c98`) to resolve a trivial conflict with the 0.2.81 release commit (#4541) — version-number churn in `Cargo.toml`/`Cargo.lock` only; the `tokio-util` addition and all `ring.rs` / `p2p_impl.rs` logic are unchanged from the reviewed version. (Bonus: #4541 already added the SAFETY comment to the pre-existing `secrets_cmd.rs` unsafe block noted above, so that's resolved upstream.)

Per the per-code-content review rule, re-ran Codex on the rebased HEAD: clean — *"The changes consistently add a cancellation token to the long-lived ring background tasks and trigger it during node teardown. I did not identify any discrete correctness issues introduced by the patch."* Targeted tests re-run green on the rebased base.

[AI-assisted - Claude]

sanity:
Rebased again (HEAD `b8b563f4`) — #4543 (`thread ModuleCacheMetrics as Arc`) landed on main and touched the same regions of `ring.rs` (the `Ring` struct, `Ring::new`, and the `emit_router_snapshot_telemetry` module-cache read). Resolved by keeping both changes: #4543's `module_cache_metrics` field/method/read alongside this PR's `shutdown` token — they're non-overlapping additions. The `emit_router_snapshot_telemetry` loop now reads `ring.module_cache_metrics.snapshot()` (#4543) *and* races its tick against the shutdown token (#4278).

Re-validated on the new content: `cargo check`/`clippy -D warnings`/`fmt` clean, targeted tests green (7 passed), and re-ran Codex — clean: *"The changes consistently add a cancellation token to the long-lived ring background tasks and trigger it during node teardown. I did not identify any discrete correctness issues introduced by the patch."*

[AI-assisted - Claude]

PR #4546, opened 2026-06-21
fix(subscribe): bound renewal attempts by budget below a cleanup-safe outer cancel (#4350)

### LINKED ISSUE #4350: SUBSCRIBE renewal outer-cancel (25s) shorter than per-attempt peer wait (60s) discards in-flight replies

sanity:
## Problem

SUBSCRIBE renewals were structurally guaranteed to be cancelled mid-flight whenever a peer took longer than ~25s to reply, wasting the in-flight attempt and producing a steady stream of "late reply on a closed receiver" drops.

`recover_orphaned_subscriptions` wraps each spawned renewal task in an **outer** cancel deadline, but the renewal driver's per-attempt network wait used the global `OPERATION_TTL = 60s` while the old outer deadline was only `SUBSCRIPTION_RECOVERY_INTERVAL - 5s = 25s`. Because the outer 25s deadline was shorter than the inner 60s wait, a renewal whose peer hadn't answered within 25s was killed by the outer cancel while its `Subscribe::Response` was still in flight; the reply then landed on the renewal task's already-dropped capacity-1 receiver — the dominant source of the `try_forward_driver_reply` closed-receiver drops whose log level was lowered in PR #4351. On the busiest gateway (nova) this ran ~30/hr, spiking after a restart when hundreds of hosted contracts re-subscribe at once.

## Approach

Two parts (the second was surfaced by review — see below):

1. **Budget-bounded attempts.** On the renewal path (`is_renewal=true`), the driver gives itself a total task budget (`Ring::RENEWAL_TASK_BUDGET = 20s`) and clamps **each** attempt's timeout to the budget remaining until that deadline (capped at `RENEWAL_PER_ATTEMPT_TIMEOUT`). When the remaining budget falls below `RENEWAL_MIN_ATTEMPT_BUDGET` it stops cleanly and lets the next 30s cycle retry. This bounds not just the first attempt but every retry after `advance_to_next_peer`, so no attempt is still awaiting when the outer cancel fires. A slow peer now produces at most one clean, in-task timeout (emitting timeout telemetry and proactively releasing its `pending_op_results` slot via `release_pending_op_slot`); the late reply then finds no slot and is a clean no-op rather than a closed-receiver drop, and the wasted full attempt is eliminated. Client / executor / directed subscribes have no outer cancel deadline and keep `OPERATION_TTL`.

2. **Cleanup-safe outer cancel.** The outer cancel deadline is now sized as `RENEWAL_TASK_BUDGET + OpManager::NOTIFICATION_SEND_TIMEOUT + RENEWAL_OUTER_DEADLINE_MARGIN = 55s` (`Ring::renewal_outer_cancel`) so it clears the driver's worst case: a full-budget slow attempt **plus** a fully backpressured `release_pending_op_slot` cleanup (which awaits up to the 30s notification-send timeout). With the old 25s deadline the outer cancel could still interrupt cleanup and leave a stale waiter — the very drop this change removes. The outer cancel is now a pure backstop for a genuinely wedged driver. Outliving the 30s recovery interval is safe: `mark_subscription_pending` (released by `SubscriptionRecoveryGuard` on completion or cancel) blocks a second concurrent renewal for the same contract, and per-cycle spawns are bounded by `MAX_RECOVERY_ATTEMPTS_PER_INTERVAL` and the channel-capacity gates.

The renewal constants are derived from `SUBSCRIPTION_RECOVERY_INTERVAL` / `NOTIFICATION_SEND_TIMEOUT` (not independently hardcoded) so they cannot drift, per the code-style rule on config-tied thresholds. A compile-time assertion pins `BUDGET + cleanup < outer` and `RENEWAL_PER_ATTEMPT_TIMEOUT < OPERATION_TTL`.

## Testing

- `op_ctx::tests::renewal_per_attempt_timeout_fires_before_outer_cancel` — virtual-time (`start_paused`) regression: a slow peer's per-attempt timeout fires before the outer cancel; a discriminator block models the historical pre-fix shape (25s outer wrapping 60s wait) and proves it WAS killed by the outer cancel (so the test discriminates).
- `op_ctx::tests::renewal_clamps_retry_attempt_to_remaining_budget` — multi-attempt regression (fast reply then a slow clamped retry) proving the retry is cut by its own budget-clamped timeout, never the outer cancel.
- `op_ctx_task::tests::renewal_driver_clamps_attempt_timeout_to_remaining_budget` — source pin that the renewal driver establishes the task deadline and clamps each attempt to the remaining budget.

All touched-module tests pass (`operations::op_ctx` 22, `operations::subscribe::op_ctx_task` 51, renewal-related ring tests 21). `cargo clippy -p freenet --bins` clean; `cargo fmt --check` clean.

## Review

External pass: `codex review --base origin/main` (run 3×). It raised two findings during development, both real and both addressed in this PR: (P2) multi-peer retries could still be outer-cancelled mid-await → fixed by clamping every attempt to remaining budget; (P2) backpressured cleanup could be interrupted by the outer cancel → fixed by sizing the outer cancel to clear budget + cleanup timeout. The final codex pass found no actionable regressions. (Gemini was reported down by the batch coordinator; a local Gemini pass was also run as a second perspective.)

This touches the SUBSCRIBE timing/renewal path (high-risk surface) and was reviewed accordingly.

Closes #4350

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -597,7 +597,11 @@ impl OpManager {
 
     /// Timeout for sending notifications to the event loop.
     /// If the channel is full for this long, the event loop is stuck and sending will never succeed.
-    const NOTIFICATION_SEND_TIMEOUT: Duration = Duration::from_secs(30);
+    ///
+    /// `pub(crate)` so the renewal outer-cancel deadline in
+    /// [`crate::ring::Ring`] can reserve enough headroom for a worst-case
+    /// backpressured `release_pending_op_slot` cleanup (issue #4350).
+    pub(crate) const NOTIFICATION_SEND_TIMEOUT: Duration = Duration::from_secs(30);
 
     // `notify_op_change` (legacy state-machine re-entry primitive)
     // is gone: every op routes outbound messages through
--- a/crates/core/src/operations/op_ctx.rs
+++ b/crates/core/src/operations/op_ctx.rs
@@ -875,6 +875,249 @@ mod tests {
             .expect("executor task should complete without panicking");
     }
 
+    /// Regression for issue #4350. The renewal driver wraps each
+    /// `send_to_and_await` attempt in a per-attempt timeout, and
+    /// `recover_orphaned_subscriptions` wraps the whole renewal future in an
+    /// **outer** cancel deadline. Before the fix the per-attempt wait was the
+    /// global `OPERATION_TTL` (60 s) while the outer deadline was only 25 s, so
+    /// a peer that replied between 25 s and 60 s tripped the OUTER cancel first
+    /// — dropping the future (and its capacity-1 reply receiver) mid-`await`,
+    /// so the in-flight reply landed on a closed receiver.
+    ///
+    /// This pins the corrected nesting at the timeout layer where the bug
+    /// lives, in virtual time (`start_paused = true`): with a per-attempt
+    /// timeout strictly *below* the outer deadline, a slow peer makes the INNER
+    /// per-attempt timeout fire first, returning a clean `Elapsed` to the
+    /// renewal task while the outer deadline still has budget left — so the
+    /// task records its own failure instead of being cancelled mid-flight.
+    ///
+    /// It also asserts the inverse to prove the test actually discriminates:
+    /// the pre-fix shape (per-attempt wait > outer deadline) is killed by the
+    /// outer cancel, which is exactly the discarded-reply path #4350 removes.
+    #[tokio::test(start_paused = true)]
+    async fn renewal_per_attempt_timeout_fires_before_outer_cancel() {
+        use crate::ring::Ring;
+        use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
+        // Current production constants/relationship from #4350.
+        let per_attempt = Ring::RENEWAL_PER_ATTEMPT_TIMEOUT; // 20 s
+        let outer = Ring::renewal_outer_cancel(); // 55 s (budget + cleanup + margin)
+        assert!(
+            per_attempt < outer,
+            "test precondition: per-attempt ({per_attempt:?}) must be below the \
+             outer cancel deadline ({outer:?}) — the #4350 invariant"
+        );
+
+        let target = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 7)), 31337);
+
+        // --- Fixed shape: outer(per_attempt(send_to_and_await)) ---
+        // A slow peer that never replies within the 20 s per-attempt window.
+        // With the fix the inner per-attempt timeout fires first, well before
+        // the 55 s outer cancel; the task gets a clean Elapsed to classify.
+        {
+            let (receiver, sender) = event_loop_notification_channel();
+            let EventLoopNotificationsReceiver {
+                mut op_execution_receiver,
+                ..
+            } = receiver;
+            let tx = Transaction::new::<ConnectMsg>();
+            let mut ctx = OpCtx::new(tx, sender.op_execution_sender.clone());
+
+            // Slow peer holds the reply sender well past the per-attempt
+            // timeout (but `peer.abort()` removes the timer so it can't advance
+            // virtual time past the deadline).
+            let peer = tokio::spawn(async move {
+                let (reply_sender, _outbound, _target) = op_execution_receiver
+                    .recv()
+                    .await
+                    .expect("outbound msg should be delivered");
+                tokio::time::sleep(outer * 100).await;
+                drop(reply_sender);
+            });
+
+            let outbound = dummy_reply_with_tx(tx);
+            let outer_result = tokio::time::timeout(
+                outer,
+                tokio::time::timeout(per_attempt, ctx.send_to_and_await(target, outbound)),
+            )
+            .await;
+
+            // The OUTER deadline must NOT fire: the inner per-attempt timeout
+            // resolves first, so `tokio::time::timeout(outer, ..)` returns Ok.
+            let inner_result = outer_result.expect(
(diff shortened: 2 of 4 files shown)


sanity:
## Review summary

**External model pass: `codex review --base origin/main`** — run across 4 iterations as the fix evolved. Codex raised two real findings during development, both now addressed:

- **[P2] Multi-peer retries could still be outer-cancelled mid-await.** A single fixed shorter per-attempt timeout protected only the first attempt; after `advance_to_next_peer` a fresh full-length timeout on attempt 2 could still outlive the outer cancel. → Fixed by clamping **every** attempt to the budget remaining until the task deadline. Covered by `renewal_clamps_retry_attempt_to_remaining_budget`.
- **[P2] Backpressured cleanup could be interrupted.** With the old 25 s outer deadline, a full-budget slow attempt followed by a fully backpressured `release_pending_op_slot` (which awaits up to `NOTIFICATION_SEND_TIMEOUT = 30 s`) could be cut off, leaving a stale waiter — the exact drop class this PR removes. → Fixed by sizing the outer cancel to `BUDGET + NOTIFICATION_SEND_TIMEOUT + MARGIN = 55 s`.

The **final codex pass found no actionable correctness issues.**

**Gemini was unavailable** (the CLI now returns `IneligibleTierError: This client is no longer supported for Gemini Code Assist for individuals`), so the external pass is codex-only. Per the multi-model-review rule this still satisfies the independent, non-Claude pass; I additionally did a four-lens self-review (code-first / testing / skeptical / big-picture), which caught a few stale "25 s" comments (the outer deadline is now 55 s) and tightened the compile-time assertion wording — comment/assertion-wording only, no change to the codex-reviewed runtime behavior.

**Local verification:** `cargo fmt --check` clean; `cargo clippy -p freenet --bins` (CI scope) clean; `operations::op_ctx` 22 tests, `operations::subscribe::op_ctx_task` 51 tests, renewal-related ring tests 21 — all pass.

[AI-assisted - Claude]

Issue #4547, opened 2026-06-21
router: per-peer adjustment should be multiplicative (not additive) for response time and transfer rate


sanity:
## Summary

The per-peer adjustment in `IsotonicEstimator` (`crates/core/src/router/isotonic_estimator.rs`) corrects the global distance→outcome fit for each peer. It is currently **additive** for all three estimators:

```
peer_estimate = global_curve(distance) + EWMA(observed - global)
```

The adjustment is a single distance-independent scalar. For unbounded, heavy-tailed, multiplicative-scale targets (response time, transfer rate) an **absolute offset** is the wrong shape: a peer's deviation from the global fit is better modelled as a near-constant **ratio**. The additive form also makes physically-impossible negative estimates possible near distance 0 (the motivation for the clamp in #4539).

## Evidence (response time)

Residual analysis on production telemetry — `route_success` events from the central OTel collector, **684,728 events over a 5.3-day window, 1,156 distinct responding peers** (631 with ≥8 obs over a non-trivial distance span):

- Additive residuals `observed - global` **grow with the level**: median Spearman(|residual|, level) = **+0.50** (heteroscedastic → wrong model).
- Multiplicative log-residuals `ln(observed) - ln(global)` are **level-independent**: median Spearman = **+0.13**.
- Multiplicative is the more homoscedastic fit for **~88–89% of peers (96% observation-weighted)**. Robust to the global-curve choice (PAV vs binned-median) and to p99 winsorization of stuck-op tails.

Conclusion: for response time, `global * exp(EWMA(ln(observed/global)))` (a geometric-mean ratio) fits the data; the additive model does not.

## Decision / plan

- **Response time → multiplicative.** Implemented in PR (linked below). Validated above.
- **Failure probability → stays additive.** It is a bounded probability in [0,1]; a multiplicative factor is the wrong model there (and `clamp(0,1)` already bounds it).
- **Transfer rate → still additive for now, pending instrumentation.** It is the same unbounded multiplicative-scale quantity as response time and is a strong candidate, but it **cannot be validated from current telemetry**: `route_success.payload_size` / `payload_transfer_time_ms` are 0 in ~99.98% of events, and the `transfer_completed` events that do carry throughput have **no contract/target location**, so residuals can't be computed against a g(distance) curve. Per our high-bar-for-model-changes practice, hold the flip until there's data.

## Follow-up needed (separate task)

[...]

PR #4548, opened 2026-06-21
feat(router): multiplicative per-peer adjustment for response time

sanity:
Follow-up to #4539. Closes part of #4547.

## What

The per-peer adjustment in `IsotonicEstimator` corrects the global distance→outcome fit for each peer. It was **additive** for all three estimators (`global + EWMA(observed - global)`) — a single distance-independent absolute offset. This makes the response-time adjustment **multiplicative** (`global * exp(EWMA(ln(observed/global)))`), driven by an empirical analysis.

## Why (evidence)

Residual analysis on production telemetry — `route_success` events from the central OTel collector, **684,728 events / 5.3 days / 1,156 peers** (631 with ≥8 obs over a non-trivial distance span):

| | additive `observed - global` | multiplicative `ln(observed/global)` |
|---|---|---|
| median Spearman(\|residual\|, level) | **+0.50** (grows with level → wrong) | **+0.13** (level-independent) |
| more homoscedastic for | — | **~88–89% of peers, 96% obs-weighted** |

Robust to the global-curve choice (PAV vs binned-median) and to p99 winsorization of stuck-op tails. A peer's deviation from the global response-time curve is a near-constant **ratio**, not a constant offset — consistent with the heavy-tailed, multiplicative nature of latency. Multiplicative also makes a negative estimate impossible by construction (`global ≥ 0`, `exp(_) > 0`).

## Scope (which estimators change)

- **Response time → Multiplicative** (global + per-op). Validated above.
- **Failure probability → stays Additive.** Bounded [0,1]; a multiplicative factor is the wrong model (and `clamp(0,1)` already bounds it).
- **Transfer rate → stays Additive for now.** Same unbounded multiplicative-scale quantity and a strong candidate, but **can't be validated**: `route_success` payload fields are ~always 0 and `transfer_completed` lacks distance/target location. Tracked in #4547 (instrument → validate → flip).

## How

- New `AdjustmentMode { Additive, Multiplicative }`, selected per estimator at construction (`new_with_mode`); `new` keeps the additive default. The mode is fixed at construction because the per-peer EWMA is trained on residuals in that mode's space.
- Multiplicative: the EWMA averages log-ratios (a geometric mean); the estimate is `global * exp(adjustment)`. The neutral value is `0.0` in **both** modes, so the prior-shrinkage seeding and the "not enough data" fallback are unchanged. Non-positive observations are skipped (`ln` undefined) rather than corrupting the EWMA.
- **Dashboard:** the peer-detail charts render the peer-adjusted curve via the same `AdjustmentMode::apply`, so `y * exp(adj)` for response time and `y + adj` otherwise — the drawn curve matches the router's prediction exactly. Legend wording updated. The #4539 non-negative clamp stays as a floor.

## Tests

- `adjustment_mode_residual_and_apply` — additive vs multiplicative residual/apply, neutral-is-noop, non-positive guard, non-negativity.
- `multiplicative_estimate_scales_global_by_geometric_factor` — a peer with log-adjustment ln(2) yields `2 × global`; a −1000 log-adjustment scales to ~0 and stays non-negative.
- Dashboard: `peer_adjusted_curve_multiplicative_scales_above_global` (factor 2× draws above the global curve), `..._multiplicative_stays_non_negative`, and the additive case retained.
- All 191 router + dashboard tests pass locally; clean build, no warnings.

## Review note

This changes the **routing cost model** (peer selection). The behavioral change is confined to the response-time estimator's per-peer adjustment; failure and transfer-rate ranking are unchanged. Worth scrutiny: the log-space EWMA semantics, the non-positive-observation skip, and that the dashboard transform exactly mirrors the router.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/router.rs
+++ b/crates/core/src/router.rs
@@ -10,7 +10,9 @@ use serde::{Deserialize, Serialize};
 use crate::config::GlobalRng;
 use crate::node::network_status::OpType;
 use crate::ring::{Distance, Location, PeerKeyLocation};
-pub(crate) use isotonic_estimator::{EstimatorType, IsotonicEstimator, IsotonicEvent};
+pub(crate) use isotonic_estimator::{
+    AdjustmentMode, EstimatorType, IsotonicEstimator, IsotonicEvent,
+};
 use util::{Mean, TransferSpeed};
 
 /// Default size of the candidate window the prediction-based router considers.
@@ -471,14 +473,21 @@ impl Router {
         }
 
         Router {
-            // Positive because we expect time to increase as distance increases
-            response_start_time_estimator: IsotonicEstimator::new(
+            // Positive because we expect time to increase as distance increases.
+            // Multiplicative per-peer adjustment: a peer's response time deviates
+            // from the global fit by a near-constant ratio, not a constant offset
+            // (validated on production telemetry). See `AdjustmentMode`.
+            response_start_time_estimator: IsotonicEstimator::new_with_mode(
                 success_durations,
                 EstimatorType::Positive,
+                AdjustmentMode::Multiplicative,
             ),
             // Positive because we expect failure probability to increase as distance increase
             failure_estimator: IsotonicEstimator::new(failure_outcomes, EstimatorType::Positive),
-            // Negative because we expect transfer rate to decrease as distance increases
+            // Negative because we expect transfer rate to decrease as distance increases.
+            // Additive for now: transfer rate is the same unbounded multiplicative-scale
+            // quantity as response time and is a strong candidate for multiplicative too,
+            // but current telemetry lacks the per-distance payload data to validate it.
             transfer_rate_estimator: IsotonicEstimator::new(
                 transfer_rates,
                 EstimatorType::Negative,
@@ -491,7 +500,16 @@ impl Router {
                 .collect(),
             per_op_response_time: per_op_response_time
                 .into_iter()
-                .map(|(k, v)| (k, IsotonicEstimator::new(v, EstimatorType::Positive)))
+                .map(|(k, v)| {
+                    (
+                        k,
+                        IsotonicEstimator::new_with_mode(
+                            v,
+                            EstimatorType::Positive,
+                            AdjustmentMode::Multiplicative,
+                        ),
+                    )
+                })
                 .collect(),
             per_op_transfer_rate: per_op_transfer_rate
                 .into_iter()
@@ -550,7 +568,12 @@ impl Router {
                     self.per_op_response_time
                         .entry(ot)
                         .or_insert_with(|| {
-                            IsotonicEstimator::new(std::iter::empty(), EstimatorType::Positive)
+                            // Multiplicative to match the global response-time estimator.
+                            IsotonicEstimator::new_with_mode(
+                                std::iter::empty(),
+                                EstimatorType::Positive,
+                                AdjustmentMode::Multiplicative,
+                            )
                         })
                         .add_event(IsotonicEvent {
                             peer: event.peer.clone(),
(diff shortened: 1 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4548

### Summary
- **PR Title:** feat(router): multiplicative per-peer adjustment for response time
- **Type:** feat (routing cost model)
- **Review tier:** Full — touches a high-risk surface (routing / peer selection)
- **Reviewers run:** code-first, testing, skeptical, big-picture (Claude) + external **Codex (gpt-5.5)**. Gemini was attempted but unavailable (ineligible-tier auth error); Codex provided the external non-Claude pass.
- **HEAD reviewed:** `01dbcef8` (re-review after fixes; original review was `f6f53f11`)
- **CI:** re-running on `01dbcef8` (was green on `f6f53f11`; all real checks pass — only the unrelated `notify_river` River-network job is red).

---

### Code-First — aligned
Independent read matches the description: **only** the response-time estimator (global + per-op + lazy `add_event` paths) is multiplicative; failure and transfer-rate stay additive. `AdjustmentMode::apply` is used by both the router (`estimate_retrieval_time`) and the dashboard (`build_estimator_chart`), so the drawn curve matches the prediction. The "neutral value is 0.0 in both modes" claim holds (`Adjustment::new` starts `smoothed=0`, prior-seeding sets only `effective_count`). No intent/impl mismatch.

### Testing — adequate after fixes
The original tests hand-inserted the `Adjustment` (only exercising `apply()`). **Added** end-to-end coverage of the production `add_event → residual → EWMA → apply` path (`multiplicative_adjustment_via_add_event_orders_peers_by_ratio` — the test that catches a sign / `ln`↔`exp` inversion), the non-positive-observation skip (`multiplicative_skips_non_positive_observations`), the `floor_base` ordering fix, and a router test pinning the per-estimator modes. 195 router+dashboard tests pass.

### Skeptical — no blockers
Traced log-space math, EWMA seeding, `exp()` overflow, serialization, and dashboard/router consistency (incl. reading the `pav_regression` interpolation source). `exp()` overflow is unreachable (needs `observed/global > e^709`; a pathological `global=1e-300` still stays finite, and an `inf` wouldn't break the `total_cmp` comparator). `#[serde(skip)]` on `adjustment_mode` cannot lose the mode — `IsotonicEstimator`/`Router` derive `Serialize` only; the model is always rebuilt from history via `Router::new`, which sets the mode explicitly.

### Big Picture — focused, evidence-backed
Scope is confined to response time; the #4539 non-negativity clamp is preserved; no tests removed (one renamed/repurposed correctly); #4547 tracks the transfer-rate follow-up. No wire-format risk.

---

### Findings and resolutions

#### Must Fix — addressed
1. **[Codex P2 — external] Multiplicative base clamp erased the peer factor.** `estimate_retrieval_time` clamped the global base to `0.0` before applying the factor, so at distances where the regression extrapolates negative, `0 * exp(adj) == 0` predicted a consistently-**slow** peer as instant (additive mode preserved the signal). **Fixed in `01dbcef8`:** `AdjustmentMode::floor_base` floors additive at `0.0` and multiplicative at a tiny positive `MULTIPLICATIVE_MIN_BASE = 1e-9`, preserving per-peer ordering in the degenerate region while leaving all normal positive-base predictions unchanged. The floor is confined to the response-time estimator (the only multiplicative one). Re-verified clean by both the external Codex pass and a focused skeptical re-check.

#### Should Fix — addressed
2. **[testing] No end-to-end test of the multiplicative path through `add_event`.** Added (see Testing).
3. **[testing] Non-positive-observation skip untested.** Added.
4. **[testing/big-picture] No `adjustment_mode` accessor / mode invariant test.** Added `adjustment_mode()` getter + `estimators_use_intended_adjustment_modes`.

#### Consider — addressed (docs) / noted (follow-ups)
5. **[big-picture/code-first] Doc nits** — fixed the stale `draw_curve` "y-offset" comment and clarified the peer_detail per-chart comment (the dashboard previews the adjustment below the router's `MIN_POINTS` gate; it mirrors the mode, so flip transfer-rate here too per #4547; reminder also added to #4547).
6. **[skeptical/code-first — pre-existing, NOT introduced here]** (a) the dashboard draws the peer-adjusted curve below the router's `MIN_POINTS_FOR_REGRESSION` apply-threshold; (b) batch (`new`) seeds `effective_count=10` while incremental (`add_event`) starts at 0. Both predate this PR and affect additive mode too; the multiplicative switch makes them more visually pronounced. Left as follow-ups, not expanded into this PR's scope.

---

### Verdict
**State: Ready to Merge.** Re-review after fixes for the Codex P2 and the testing/doc findings — all resolved on the current HEAD; the external Codex pass and a focused skeptical re-check on `01dbcef8` are both clean. Merge once CI is green on the current HEAD.

**HEAD SHA reviewed:** `01dbcef8`

[AI-assisted - Claude]

PR #4550, opened 2026-06-22
fix: stop gateway death-spiral wedge under contract-executor backpressure (#4549)

### LINKED ISSUE #4549: Gateway death-spirals at ~100% CPU in libunwind FDE-cache invalidation under contract-executor backpressure; network listener dies, node wedges unrecoverable (0.2.81)
## Summary

nova's production gateway (`freenet 0.2.81`, built 2026-06-21T17:24Z) **wedged**: the process is alive but pinned at ~108% CPU with its **network event listener dead** and the ws-api (7509) **unbound**. It does **not self-recover** — the process never exits, so systemd never restarts it and the exit-42 auto-update hook never fires; it just spins.

`perf` on the wedged process shows **~99.8% of CPU in `libunwind::DwarfFDECache::removeAllIn`** (DWARF unwind-info cache invalidation), called from anonymous/JIT addresses (wasmtime-compiled code). This is a **death spiral** driven by contract-executor backpressure plus the *"Recovering engine store after timeout/panic"* path repeatedly churning wasmtime frame (FDE) registration. Observed recurring (the fatal `Network event listener exited` event fired multiple times this run).

## Impact

- A production gateway goes fully dark (no peer connections, no ws-api) while still looking *"active (running)"* to systemd.
- **Unrecoverable without a manual `systemctl restart`** — no liveness/watchdog catches it, and auto-update can't fire.
- Concretely: River (and every) GET through nova times out (HTTP 500 "GET request timed out") and then connection-refuses once the ws-api accept loop drops; a hosted-mode node behind this gateway can't serve anything.

### CPU pin (perf, captured while wedged — PID pinned ~108% CPU, ~21h uptime)
```
99.83%  freenet  [.] libunwind::DwarfFDECache<libunwind::LocalAddressSpace>::removeAllIn(unsigned long)
   |--58.54%-- <jit/anon addr> -> removeAllIn
   |--40.77%-- <jit/anon addr> -> removeAllIn
```
Essentially all CPU is in FDE-cache removal, reached from JIT/anonymous (wasmtime) frames. (System `libunwind`.)

### Log evidence (hour of the collapse, 14:00–14:23 UTC)
Error-class histogram for that one hour:
```
no response received from handler            1336
Failed to send queue-full response           1023
timed out                                     133
Recovering engine store after timeout/panic    22   (across ~22 DISTINCT contract keys — global load, not one poison contract)
Network event listener exited                   1
listen task exiting                             1
```
Collapse cascade at 14:23 (a flood of queue-full WARNs, then the fatal listener exit):
```
14:23:41Z WARN contract_handling: freenet::contract: Failed to send queue-full response (client may have disconnected) error=no response   (×~1023)
... WARN transport::connection_handler: Connection handler channel closed — listen task exiting ... bind_addr=0.0.0.0:31337
... ERROR node::p2p_impl: Network event listener exited: no response received from handler
... INFO  node::p2p_impl: Peer shutdown event emitted graceful=false reason="no response received from handler"
```
The main log and the error log are **byte-for-byte the same size** → every emitted line is WARN/ERROR. Backpressure has been chronic for the full ~20h uptime (1k–16k error lines/hour).

## Root-cause analysis (the death spiral)

1. The WASM contract executor is globally overwhelmed under load (the #4404/#4534 placement migration is a heavy GET/UPDATE source; the affected ops are mostly `upsert_contract_state_deferrable` across many contracts).
2. Ops time out → **"Recovering engine store after timeout/panic"** reloads the wasmtime engine/modules.
3. Module (re)load/unload + trap unwinding churns wasmtime's `__register_frame`/`__deregister_frame` against libunwind's FDE cache. **`DwarfFDECache::removeAllIn` is O(n)** and, under high churn, comes to dominate CPU.
4. CPU pinned → executor even slower → more timeouts → more engine recovery → more FDE churn (positive feedback).
5. The contract handler stops responding → the network event loop's request to it times out → **"no response received from handler" is treated as fatal and exits the network event listener** → node is network-dead.
6. The process keeps spinning in libunwind and never exits → no restart, no auto-update.

(Steps 1–3 are inference from the perf signature + the log; the perf data and logs themselves are direct observations.)

## Sub-problems (a taker can split these out and multitask)

1. **libunwind FDE-cache thrash under wasmtime engine recovery** (the CPU-pin root). Candidate directions: don't do a full engine-store reload on a per-op timeout/panic; reuse compiled modules / module cache; ensure wasmtime's unwind-registration strategy doesn't hit O(n) system-libunwind FDE invalidation (e.g. wasmtime native-unwind settings); rate-limit how often engine recovery can fire.
2. **A per-op handler timeout must not kill the whole network event listener.** "no response received from handler" should be a recoverable per-op error (shed/queue-full that op, or restart the listener), not fatal to the node's entire network loop. NOTE: #4499 made event-loop *sends* non-blocking for the #4145 wedge, but this is a **different** fatal path still present on 0.2.81.
3. **Chronic contract-executor backpressure under load.** The queue-full flood (~1023/hr) means the executor can't keep up; bound/shed work. Related to #4534 (module-cache ceiling → "contract queue full") and #4404 migration load.
4. **Wedged-but-alive liveness gap.** A node spinning at 100% CPU with a dead listener should self-detect (watchdog) and **exit non-zero**, so systemd restarts it and the auto-update hook can fire — instead of silently spinning forever.

## Related

- #4145 / #4499 — event-loop wedge under load (channel-send fix); this is a *different*, still-present fatal path.
- #4534 — placement-migration module-cache overrun → "contract queue full" client outage (0.2.80).
- #4404 — placement migration (load source).
- #3333 — bound unbounded channels/queues.

## How observed / repro

Observed on the live nova gateway (0.2.81). A full perf capture (call graph), thread headers, and log excerpts were saved as a diagnostic bundle (available on request). Reproducing likely needs sustained executor overload (high GET/UPDATE rate across many contracts) such that "Recovering engine store after timeout/panic" fires repeatedly; a stress/sim harness that forces engine recovery under load would pin the libunwind path.

## Environment

`freenet 0.2.81` (built 2026-06-21T17:24Z), gateway role, Linux x86-64, system `libunwind`, wasmtime backend.

[AI-assisted - Claude]

sanity:
## Problem

Both production gateways (nova, vega) crash on a ~21.5h cadence under the #4404/#4534 placement-migration load. On the big box (nova) it manifests as the death-spiral in #4549:

1. The WASM contract executor is globally saturated → ops cross the 5s timeout → wasmtime "Recovering engine store after timeout/panic" churns module register/unregister.
2. Wasmtime registers per-module DWARF unwind info (FDEs) with the **system** libunwind. libunwind's FDE-cache invalidation (`DwarfFDECache::removeAllIn`) is **O(n)**, so under heavy module load/unload it goes quadratic and pins the box at ~100% CPU (perf showed 99.8% there). Stolen CPU slows the executor → more timeouts → more churn (positive feedback).
3. A client diagnostics query (`NodeQuery::SubscriptionInfo`) runs **inline on the network event loop** and `await?`s the contract handler. Under saturation it returns `NoEvHandlerResponse` after the 300s handler timeout — which was **fatal**: it killed the whole network event listener ("Network event listener exited: no response received from handler"), taking the node network-dead.
4. The wedged process then **never exited promptly** — it spun network-dead for ~37 min before the normal teardown path finally returned (exit 1). systemd never restarted it in a useful window and the exit-42 auto-update hook never fired.

This addresses the three **root causes** of the unrecoverable wedge. The upstream executor *load* itself (chronic queue-full backpressure) is the #4534/#4537 migration-throughput workstream and is intentionally **out of scope** here — its queue bounds and the 300s handler timeout are load-bearing and unchanged.

## Approach (three root-cause fixes, one per stage of the spiral)

**A — Disable native (DWARF) unwind-info registration (`wasm_runtime/engine/wasmtime_engine.rs`).**
`wasmtime_config.native_unwind_info(false)` (gated `cfg(not(windows))`, where wasmtime forbids it). This is the root of the CPU pin: it removes the per-module `__register_frame`/`__deregister_frame` calls that feed libunwind's quadratic FDE cache. Wasmtime's own docs explicitly recommend disabling it for "systems loading many modules". **Safety:** it does not affect wasmtime's own Wasm-trap backtraces (governed by `wasm_backtrace*`), so contract trap / out-of-fuel reporting is unchanged; the only thing lost is a third-party/OS unwinder's ability to walk *through* JIT'd wasm frames, which Freenet never does (contract WASM runs via `spawn_blocking` and panics are caught as `BlockingResult::Panic`, never unwound across the wasm boundary). Freeing the pinned CPU also lets the executor keep up, attacking the spiral at its source.

**B — Fast-exit on a fatal network-listener death (`node/p2p_impl.rs`, `bin/freenet.rs`).**
On a *non-graceful* listener exit, `process::exit(42)` immediately instead of unwinding through the teardown/return path that hung for ~37 min in production. `42` = `EXIT_CODE_UPDATE_NEEDED`: it is `SuccessExitStatus` (burst-exempt, so a repeating wedge can't trip `StartLimitBurst` and brick the unit) and its `ExecStopPost` runs `freenet update` (so a wedged old binary auto-upgrades to a fixed one). Graceful shutdown (`NodeEvent::Disconnect` via SIGTERM / auto-update → `EventLoopExitReason::GracefulShutdown`) still falls through to the clean teardown. The force-exit is **off by default** and enabled only by the real `freenet` binary, so simulation / integration tests and library embedders never have a listener exit kill their process.

**C — Make the on-loop diagnostics `QuerySubscriptions` non-fatal and bounded (`node/network_bridge/p2p_protoc.rs`).**
Both diagnostics arms (`QuerySubscriptions`, `QueryNodeDiagnostics`) now wrap the inline `notify_contract_handler(QuerySubscriptions ...)` in `timeout(QUERY_SUBSCRIPTIONS_HANDLER_TIMEOUT = 5s, ..)` and treat any failure as "serve empty app subscriptions". A per-op handler timeout can no longer kill the listener, nor stall the event loop for up to 300s. A genuinely-dead contract handler is still detected by the separate `contract_executor_task` supervisor in `run_node`.

## Testing

- `cargo fmt`, `cargo clippy --all-targets --all-features`, `cargo test -p freenet`.
- New regression tests:
  - `p2p_impl::tests::listener_exit_graceful_classification` — pins that the force-exit predicate fires only on a fatal exit and never on `GracefulShutdown` (a regression either way is dangerous: spurious exit-42 on every clean stop, or failure to recover a real wedge).
  - `p2p_impl::tests::fatal_listener_fast_exit_is_opt_in` — the force-exit flag is off by default (tests/embedders never self-kill) and the exit code is 42.
  - `p2p_protoc::tests::query_subscriptions_diagnostics_bounded_and_non_fatal` — source-scrape pin (same pattern as the #4145 guards) that both diagnostics `QuerySubscriptions` calls are wrapped in the bounded timeout and never use the fatal `.await?`.
- Fix A's no-regression is covered by the existing wasm_runtime tests (`test_wasmtime_engine_creation`, `wasm_runtime/tests/execution_handling.rs`, `contract_metering.rs`) continuing to pass — FDE registration is internal to wasmtime and not unit-testable from our side.

## Why one PR

All three are tightly-coupled root-cause fixes for the single death-spiral in #4549 (the issue frames them as sub-problems of one bug). Per the finish-the-fix principle, the user-visible "unrecoverable wedge" symptom only goes away when all three land together — A stops the CPU pin, C keeps the listener alive through transient saturation, B guarantees a prompt restart if anything else wedges it.

Closes #4549

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -223,6 +223,12 @@ async fn run_network_node_with_signals(
     use freenet::transport::{clear_urgent_update, get_highest_seen_version, is_urgent_update};
     use tokio::signal;
 
+    // #4549: this is the real node process (not a test / embedded use), so enable the
+    // fast-exit watchdog — a fatal network-event-listener exit will abort the process
+    // (code 42) for a prompt systemd restart + auto-update, instead of risking a hang
+    // in teardown that leaves the node network-dead and spinning.
+    freenet::enable_abort_on_fatal_listener_exit();
+
     // Set up SIGTERM handler for Unix systems
     #[cfg(unix)]
     let mut sigterm = signal::unix::signal(signal::unix::SignalKind::terminate())
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -19,7 +19,10 @@ mod message;
 
 /// Node configuration, implementations and execution (entry points for the binaries).
 mod node;
-pub use node::{EventLoopExitReason, Node, ShutdownHandle, run_local_node, run_network_node};
+pub use node::{
+    EventLoopExitReason, Node, ShutdownHandle, enable_abort_on_fatal_listener_exit, run_local_node,
+    run_network_node,
+};
 
 /// Network operation/transaction state machines.
 mod operations;
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -80,6 +80,7 @@ mod p2p_impl;
 mod request_router;
 pub(crate) mod testing_impl;
 
+pub use p2p_impl::enable_abort_on_fatal_listener_exit;
 pub use request_router::{DeduplicatedRequest, RequestRouter};
 
 /// Handle to trigger graceful shutdown of the node.
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -38,7 +38,7 @@ use crate::transport::{
 use crate::{
     client_events::ClientId,
     config::GlobalExecutor,
-    contract::{ContractHandlerChannel, ExecutorTransactionStream, WaitingResolution},
+    contract::{ContractHandlerChannel, ExecutorTransactionStream, SenderHalve, WaitingResolution},
     message::{MessageStats, NetMessage, NodeEvent, Transaction, TransactionType},
     node::{
         NetEventRegister, NodeConfig, OpManager, PeerId, WaiterReply, process_message_decoupled,
@@ -109,6 +109,53 @@ impl std::error::Error for EventLoopExitReason {}
 /// poll. A diagnostics query is best-effort, so cap the wait and serve an empty
 /// result on timeout rather than freezing (or, previously, killing) the listener.
 const QUERY_SUBSCRIPTIONS_HANDLER_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Best-effort, bounded query to the contract handler for application-level
+/// subscriptions, used by the diagnostics arms of the network event loop (#4549).
+///
+/// This runs INLINE on the network event loop, so it must be both bounded and
+/// non-fatal: a saturated contract handler previously made the bare
+/// `notify_contract_handler(QuerySubscriptions ...).await?` return
+/// `NoEvHandlerResponse` only after the 300s handler timeout, which killed the
+/// whole network event listener ("Network event listener exited: no response
+/// received from handler") and took the gateway network-dead. Here the wait is
+/// capped at [`QUERY_SUBSCRIPTIONS_HANDLER_TIMEOUT`] and ANY failure (timeout,
+/// channel error, missing response) yields an empty list. A genuinely-dead
+/// contract handler causes the `contract_executor_task` to exit, which the
+/// `run_node` supervisor observes via its `deterministic_select!` and brings the
+/// node down — so de-fatalizing this diagnostics query does not mask a dead
+/// handler.
+async fn query_app_subscriptions_bounded(
+    ch_outbound: &ContractHandlerChannel<SenderHalve>,
+) -> Vec<crate::message::SubscriptionInfo> {
+    let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+    match timeout(
+        QUERY_SUBSCRIPTIONS_HANDLER_TIMEOUT,
+        ch_outbound.send_to_handler(ContractHandlerEvent::QuerySubscriptions { callback: tx }),
+    )
+    .await
+    {
+        Ok(Ok(_)) => match timeout(Duration::from_secs(1), rx.recv()).await {
+            Ok(Some(QueryResult::NetworkDebug(info))) => info.application_subscriptions,
+            _ => Vec::new(),
+        },
+        Ok(Err(err)) => {
+            tracing::warn!(
+                error = %err,
+                "QuerySubscriptions diagnostics: contract handler error; \
+                 serving empty app subscriptions"
+            );
+            Vec::new()
+        }
+        Err(_elapsed) => {
+            tracing::warn!(
+                "QuerySubscriptions diagnostics: contract handler did not respond within \
+                 timeout; serving empty app subscriptions"
+            );
+            Vec::new()
+        }
+    }
+}
 
 pub(crate) enum P2pBridgeEvent {
     Message(PeerKeyLocation, Box<NetMessage>),
@@ -1401,7 +1448,7 @@ impl P2pConnManager {
                                     )
                                     .await
                                     {
-                                        Ok(Ok(_)) => {
+                                        Ok(Ok(())) => {
                                             tracing::trace!(
                                                 tx = %msg.id(),
                                                 peer_addr = %target_addr,
@@ -2058,7 +2105,7 @@ impl P2pConnManager {
                                 )
                                 .await
                                 {
-                                    Ok(Ok(_)) => {}
+                                    Ok(Ok(())) => {}
                                     Ok(Err(send_error)) => {
                                         tracing::error!(
                                             error = ?send_error,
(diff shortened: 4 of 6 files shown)


sanity:
## Multi-model review (Full tier) — round 1

This touches high-risk surfaces (wasm runtime, node liveness, network event loop), so it got a Full review: four independent Claude lenses (code-first, testing, skeptical, big-picture) plus an external **Codex** pass.

**Codex and the skeptical lens found no correctness/security/maintainability regressions.** Code-first verified all three fixes end-to-end (fix A safe against the wasmtime 45 docs; fix C cancellation-safe with no leak/storm — one debug line per cancelled query). No BLOCKERs. The actionable findings and how they were addressed:

### Addressed in this push

- **(big-picture, MAJOR) exit-42 hot-restart loop risk.** Exit 42 is `SuccessExitStatus` (burst-exempt), so a node that re-wedges fast could hot-restart every `RestartSec` forever, invisible to `StartLimitBurst`. **Fix:** added a minimum-uptime guard — a fatal listener death only fast-exits 42 after ≥5min uptime (the #4549 wedge happened at ~21.5h); a wedge within 5min of boot is treated as a boot loop and exits with a *counted* code (1) so `StartLimitBurst` can stop a tight loop. This also avoids the macOS update-backoff conflation (below) for the boot case.
- **(code-first, MINOR) macOS exit-42 conflation.** On macOS the service wrapper treats a non-zero `freenet update` result as a failure and backs off. Gateways are Linux (prod unaffected); documented as a caveat on `FATAL_LISTENER_EXIT_CODE`, and the boot-wedge path now uses code 1.
- **(testing, MAJOR) fix C was pinned only by source-scrape.** **Fix:** extracted the duplicated diagnostics query into `query_app_subscriptions_bounded(&ch_outbound)` (DRYs both arms — also resolves big-picture's duplication MINOR) and added a behavioral test `query_app_subscriptions_bounded_yields_empty_when_handler_stalls` (`#[tokio::test(start_paused = true)]`): a stalled handler yields an empty list within the bound, never errors, never waits the 300s handler timeout. This fails if the stall-survival regresses by *any* route, not just the one syntactic revert.
- **(code-first, MINOR) scope creep.** A `replace_all` had widened three unrelated pre-existing `Ok(Ok(()))` mpsc-send patterns to `Ok(Ok(_))`. Reverted to keep the diff focused.
- **(testing, NIT) timeout-bound + typed-error pins.** Added `QUERY_SUBSCRIPTIONS_HANDLER_TIMEOUT <= 30s` to the source-scrape pin, and a `ContractError::NoEvHandlerResponse` (typed, not string) assertion to the graceful-classification test.
- **(big-picture, MINOR) comment accuracy.** The `42` doc now references the *generated* unit template (`service/linux.rs`), and the "dead handler is caught by the supervisor" claim is sharpened (a dead executor task resolves its `deterministic_select!` arm in `run_node` → node exits; saturation stays alive via fix C).

### Behavioral notes (by design, acknowledged)

- **Only the listener arm fast-exits.** The other supervised arms (`client_events_task`, `contract_executor_task`, `infra_monitor`, `background_monitor`) still return `Err` and fall through to teardown. That's acceptable: `drop(teardown)` is non-blocking (signals only, no `.await`), so it can't reproduce the 37-min hang, and #4549's wedge was specifically the listener.
- **Integration tests run with the flag enabled but stop nodes via SIGKILL**, so the fast-exit introduces no new CI flakiness (a spurious exit would require an actual fatal listener exit, which would have failed those tests pre-PR anyway).
- **Upstream executor *load* (chronic queue-full backpressure) is intentionally out of scope** — that's #4534/#4537's migration-throughput workstream; the queue bounds and the 300s handler timeout are load-bearing and unchanged.

### Dismissed (with justification)

- **(testing, MINOR) add a hard-trap test for fix A** (a contract that traps via `unreachable`, distinct from fuel-exhaustion/timeout). Not added: the existing `wasm_runtime/tests/{contract_metering,execution_handling}.rs` already exercise the trap *machinery* (`OutOfGas`, `MaxComputeTimeExceeded`) under the patched engine, and three reviewers independently verified — against the wasmtime 45 docs — that `native_unwind_info` is orthogonal to wasmtime's own trap reporting (host-fn panics are caught by wasmtime's trampoline / at the `spawn_blocking` thread boundary, not via system DWARF unwinding). Gap-hardening, not a correctness gap; out of scope for this hotfix.

Re-running the external (Codex) pass + a focused skeptical lens on the new helper/guard, since a changed diff is an unreviewed diff. Will confirm green before merge.

[AI-assisted - Claude]

sanity:
## Multi-model review — round 2 (re-review of the round-1 fixes)

Re-ran an external **Codex** pass + a focused **skeptical** Claude lens on the round-1 changes (commit-level). Codex: no actionable issues. Skeptical lens verified the helper extraction is behavior-preserving and the new behavioral test is non-vacuous (the stalled-handler path genuinely exercises the outer timeout, deterministic under `start_paused`).

It also surfaced one **MAJOR** in the min-uptime guard I added in round 1:

> The guard exits a boot-wedge with a *counted* code on the premise that `StartLimitBurst` would stop a tight loop — but the unit's `StartLimitBurst`/`StartLimitIntervalSec` are emitted in the systemd **`[Service]`** section, where systemd ignores them (they're `[Unit]` directives since v230). So the burst limiter is inactive: the counted code gains nothing and *loses* the auto-update self-heal for boot wedges.

**Resolution:** reverted the guard — the fatal path now always exits 42. The hot-restart-loop concern is bounded by `RestartSec=10` and visible (a `CRITICAL` log per exit + climbing `NRestarts`), and a wedged old node self-heals via the exit-42 auto-update hook. That's strictly better than the #4549 dark wedge. Restoring a *working* burst limiter (move the directives to `[Unit]`), and then a meaningful min-uptime exit-code guard on top of it, is a deployment-config change (high blast radius, only affects future installs) and is filed as **#4551**.

Final HEAD `1e26591f`. Running the external pass once more on the final diff; will confirm green CI on this SHA before merge.

[AI-assisted - Claude]

Issue #4561, opened 2026-06-23
P5: abuse-hardening for the hosted public proxy (quotas, rate-limits, TTL) before wide exposure (#4381)


sanity:
## Context

Phase 5 of the hosted public proxy (#4381). The hosted node (`--hosted-mode`) accepts untrusted web visitors, each with a per-user token and isolated, encrypted delegate data. The mechanism is built and live as a **supervised** demo at `try.freenet.org`. Before the link is shared **widely**, the node needs abuse-hardening: today a hosted node is an open resource (disk, CPU, executor time, network, unbounded account creation) with no per-user governance.

This is a tracking issue for that governance work. It is a real, multi-PR feature, not a one-liner. None of it is required for the *supervised* demo; all of it gates *wide / unsupervised* exposure.

## Threat surface (what an unsupervised public hosted node exposes)

- **Disk fill:** any visitor can mint a token and write delegate secrets; unbounded per-user storage fills the node's disk.
- **Mass account creation:** unbounded token minting from one source multiplies every other limit.
- **Executor / CPU DoS:** an authenticated token-holder can stall the single-threaded contract loop. The export endpoint is the known instance (#4531); large/repeated per-user operations are the general case.
- **Request flooding:** no per-IP / per-user request rate limit at the edge.
- **Abandoned data:** per-user data accumulates forever with no reclamation.

## Sub-tasks

1. **Export must not block the contract loop (#4531).** Offload the synchronous export off the contract-handling loop and/or bound per-user export size. _In progress as a focused fix PR (referenced below)._ This is the first and most concrete piece.

2. **Per-user storage quota.** Bound total secret-store bytes per `UserId`. Enforce at the write path (`set_secret`), reject over-quota writes with a typed error surfaced to the client. Decide where per-user accounting lives (running total per UserId) and how it is kept consistent.

3. **Per-user operation rate limit.** Bound the per-user rate of GET/PUT/UPDATE/SUBSCRIBE/export so one token cannot hammer the node. Reuse the existing backoff/limit primitives where possible. Tie the export to this rather than a one-off.

4. **Account/token creation limit.** Bound how many new per-user tokens a single source (IP / subnet) can mint per window, so quotas above cannot be trivially multiplied. This likely belongs at the proxy/edge layer, not the node.

5. **Edge rate-limiting / DDoS.** The node's per-user limits do not cover anonymous flooding of the public endpoint. Options:
[...]

PR #4563, opened 2026-06-23
fix(hosted): keep the export endpoint from blocking the contract loop (#4531 P5)

sanity:
## Problem

The hosted-mode "export my data" endpoint (`GET /v{1,2}/hosted/export`, P3-live of #4381) ran the secret export **synchronously on the single-threaded contract-handling loop**. The export enumerates AND AEAD-decrypts **every** secret in a user's scope, with no per-user cap. So an authenticated token-holder with a large secret set — or one who repeats the request — could block **all** other contract operations (GET/PUT/UPDATE/delegate) on the node for each export's duration. The documented #4531 / #4381 P5 blocker.

## Solution (reworked after the first review — see #4563 history)

The first revision offloaded the CPU to `spawn_blocking` but still **awaited it inline** on the loop, so the loop stayed parked and queued events still waited (the offload alone did **not** fix #4531). This revision fixes that and two other issues the review surfaced:

**1. DEFER off-loop (the real fix).** The `ExportUserSecrets` dispatch arm now checks out a pooled executor (`RuntimePool::try_begin_export`) and runs the export on a `GlobalExecutor::spawn` background task, **returning from the arm immediately** so the contract loop drains the next event. The task sends the executor + result back over a new unbounded export-resume channel; the loop returns/replaces the executor (`finish_export`) and answers the parked client (`StashedResponder`) — mirroring the #4391 related-fetch deferral. Concurrency is bounded by `MAX_CONCURRENT_EXPORTS` (=2, below pool size) so exports can never drain the executor pool; over the cap → typed `ExportBusy` → **HTTP 503**. An `ExportGuard` RAII guarantees exactly-once resume (and executor reclaim) even if the task is dropped/cancelled.

**2. PANIC-SAFE offload (was a node-crash bug).** `run_blocking_offloaded` now **catches** a `JoinError` panic (mirroring `compile_offloaded`) and returns `OffloadOutcome::Panicked` instead of re-raising via `.expect()`. The old `.expect()` would have propagated the panic onto the loop task → the node's top-level `select!` → **whole-node shutdown**, and leaked the executor slot. On panic the executor is gone, so `finish_export` builds a replacement to restore the pool permit; the one export fails, the node lives.

**3. INCREMENTAL byte cap (was enforced too late).** The 256 MiB plaintext cap is now enforced **inside** the decrypt loop (`SecretsStore::export_scope_entries_bounded`), aborting the instant the running total crosses the limit and dropping the partial buffer — instead of decrypting+buffering the user's **whole** scope and only then rejecting. The 10,000-secret count cap still short-circuits before any decrypt. Over either cap → typed `ExportTooLarge` → **HTTP 413**.

**4. Fixed misleading comments** flagged in review: the export is a read-only **filesystem-blob** walk over a per-executor `SecretsStore` on a shared `secrets_dir` (not "redb single-writer"; `pop_executor` does not serialize other executors). Concurrency safety is per-file FS semantics + AEAD authentication (a torn read fails AEAD → clean error, never corruption).

## Why defer rather than refactor the pool

A background task cannot check out a pool executor today (`pop_executor`/`return_checked` are `&mut self`). Rather than make the pool `&self`-shareable (a concurrency-model change on a high-risk surface), the loop pops the executor (fast slot-take) and **moves the owned `Executor<Runtime>` into the task**, which returns it via the resume channel. This keeps the pool's sequential invariant intact — only the loop ever touches the pool — while the export work runs on an exclusively-owned executor off-loop.

## Testing

- **Incremental bail** unit test: three 40-byte secrets, 50-byte cap → bails at total=80 (after the 2nd secret), proving it does NOT buffer the whole scope (would be 120).
- **Panic-catch** unit test: a panicking offload surfaces as `OffloadOutcome::Panicked`, not a re-raise onto the caller (no node crash).
- **Defer** integration test (`export_does_not_block_concurrent_contract_ops`): with secrets stored, an HTTP export and a concurrent WS delegate op fire together; the concurrent op completes (was not starved). The previous inline design would have serialized it behind the export.
- **Source-scrape pins**: the export arm defers via `GlobalExecutor::spawn`/`try_begin_export` with no inline `export_user_secrets`; `ExportJob::run` routes through `run_blocking_offloaded`; `finish_export` replaces a panic-lost executor.
- Existing cap + round-trip + isolation tests stay green: `hosted_mode_export` 5/5, `hosted_mode_isolation` 3/3, cap units 16/16, offload/pin 14/14.
- `cargo fmt`; `cargo clippy -p freenet --lib -- -D warnings` clean; `cargo test --workspace --no-run` compiles.

## Residual risk (for the P5 tracking issue, #4561)

- No **cross-request rate limiting** (N exports/minute per user). `MAX_CONCURRENT_EXPORTS` now bounds *node-wide concurrency* (excess → 503), which closes the aggregate-concurrency gap from the first revision, but a user can still issue exports back-to-back over time.
- The caps and concurrency limit are fixed consts (not operator-configurable yet).

Refs #4531, #4381 (P5).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -2154,9 +2154,14 @@ where
         } => {
             // Hosted-mode export (P3-live of #4381). Runs on the executor (which
             // owns the SecretsStore) and returns the encrypted bundle bytes.
-            // `user_context` is the forge-proof per-user namespace derived at the
-            // connection boundary; the export reads ONLY that user's scope. Log
-            // only the non-secret user_id — never the token or bundle bytes.
+            // The executor's `export_user_secrets` offloads the synchronous
+            // enumerate+decrypt+re-encrypt to a blocking thread
+            // (`RuntimePool::export_user_secrets`), so awaiting it here does NOT
+            // stall this single-threaded loop for the export's duration (#4381
+            // P5). `user_context` is the forge-proof per-user namespace derived
+            // at the connection boundary; the export reads ONLY that user's
+            // scope. Log only the non-secret user_id — never the token or
+            // bundle bytes.
             tracing::debug!(
                 user_id = ?user_context.user_id(),
                 "Processing hosted-mode secret export request"
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -2360,7 +2360,6 @@ where
             match contract_handler
                 .executor()
                 .try_begin_export(&user_context, token.expose())
-                .await
             {
                 executor::runtime::ExportAdmission::Admitted(job) => {
                     // Park the client responder so the off-loop task can answer it
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -16,7 +16,7 @@ pub mod storages;
 pub(crate) mod user_input;
 
 pub(crate) use executor::{
-    ContractExecutor, ExecutorTransactionStream, MAX_CREATED_DELEGATES_PER_NODE,
+    ContractExecutor, ExecutorTransactionStream, ExportBusy, MAX_CREATED_DELEGATES_PER_NODE,
     MAX_DELEGATE_CREATION_DEPTH, MAX_DELEGATE_CREATIONS_PER_CALL,
     SUBSCRIBER_NOTIFICATION_CHANNEL_SIZE, UpsertOutcome, UpsertResult, mock_runtime::MockRuntime,
 };
@@ -231,6 +231,101 @@ impl Drop for ResumeGuard {
     }
 }
 
+/// A hosted-mode export that ran OFF the contract loop, carrying its result
+/// (and the executor to return to the pool) back to the loop (#4531 / #4381 P5).
+///
+/// Built by the off-loop export task and sent back to the `contract_handling`
+/// loop via the export-resume channel; the loop returns/replaces the executor
+/// (`finish_export`) and answers the stashed client responder. Mirrors the
+/// #4391 [`DeferredResume`] flow, but the work is the export itself (not a
+/// network wait), so there is nothing to re-run on the loop — just deliver the
+/// precomputed result and reclaim the executor.
+struct ExportResume {
+    /// The completed export: the executor to return (or `None` if the export
+    /// task panicked) and the result for the client.
+    done: crate::contract::executor::runtime::ExportDone,
+    /// The parked client responder (the HTTP export handler's oneshot). `None`
+    /// only if the client disconnected before the export finished.
+    responder: Option<StashedResponder>,
+}
+
+/// RAII guard guaranteeing the off-loop export task delivers EXACTLY ONE
+/// [`ExportResume`] — whether the export completes, or the task is
+/// dropped/panics/cancelled before sending.
+///
+/// Same load-bearing invariant as the #4391 [`ResumeGuard`]: every admitted
+/// export is answered exactly once, so the parked client responder never hangs
+/// and the checked-out executor is always reclaimed. On the success path
+/// [`send`](Self::send) takes the payload (Drop becomes a no-op); on any early
+/// exit, `Drop` delivers an `ExportResume` carrying a panicked-style
+/// [`ExportDone`] (`executor: None`) so the loop replaces the lost executor and
+/// answers the client with an error.
+struct ExportGuard {
+    payload: Option<ExportGuardPayload>,
+}
+
+struct ExportGuardPayload {
+    export_resume_tx: tokio::sync::mpsc::UnboundedSender<ExportResume>,
+    responder: Option<StashedResponder>,
+}
+
+impl ExportGuard {
+    fn new(
+        export_resume_tx: tokio::sync::mpsc::UnboundedSender<ExportResume>,
+        responder: Option<StashedResponder>,
+    ) -> Self {
+        Self {
+            payload: Some(ExportGuardPayload {
+                export_resume_tx,
+                responder,
+            }),
+        }
+    }
+
+    /// Deliver the completed export. Consumes the payload so a subsequent `Drop`
+    /// is a no-op (exactly-once).
+    fn send(mut self, done: crate::contract::executor::runtime::ExportDone) {
+        if let Some(p) = self.payload.take() {
+            Self::deliver(p, done);
+        }
+    }
+
+    fn deliver(p: ExportGuardPayload, done: crate::contract::executor::runtime::ExportDone) {
+        let ExportGuardPayload {
+            export_resume_tx,
+            responder,
+        } = p;
+        // Unbounded channel, non-blocking send (channel-safety.md carve-out: the
+        // producer count is bounded by MAX_CONCURRENT_EXPORTS, the receiver is
+        // this loop which drains every iteration, and the waiter never reads what
+        // the loop produces — no cycle). A send failure means the loop is gone.
+        if export_resume_tx
+            .send(ExportResume { done, responder })
+            .is_err()
+        {
+            tracing::debug!("export resume channel closed; contract-handling loop gone");
+        }
+    }
+}
+
+impl Drop for ExportGuard {
+    fn drop(&mut self) {
+        if let Some(p) = self.payload.take() {
+            // Early exit before `send` (task dropped / panicked / cancelled).
+            // Deliver a panicked-style ExportDone so the loop replaces the lost
+            // executor and answers the parked client exactly once.
+            tracing::warn!(
+                "Off-loop export task dropped before sending — delivering an error \
+                 resume so the export terminates and the client is answered (#4531)"
+            );
+            Self::deliver(
+                p,
+                crate::contract::executor::runtime::ExportDone::lost_to_drop(),
+            );
+        }
+    }
+}
+
 /// Loop-side state for off-loading deferred related-contract fetches (#4391).
 ///
 /// Owned by `contract_handling`; passed by `&mut` into `handle_contract_event`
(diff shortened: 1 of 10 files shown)


sanity:
## Full review (Codex + 2 adversarial lenses + maintainer trace) — rework needed, do NOT merge as-is

High-risk surface (contract executor), so Full tier. Reviewers: Codex (external), an offload/concurrency lens, a cap/correctness lens, plus a maintainer trace of the loop.

### Reconciling the core disagreement
The concurrency lens read Q2 as "DoS fixed" because the offload stops the worker-thread CPU-pin (no node wedge). But it acknowledges the deciding caveat: the serial loop (`contract.rs:1096-1104`) **awaits `handle_contract_event` inline** before popping the next event, so awaiting `spawn_blocking` does NOT free the contract queue — GET/PUT/UPDATE/delegate events queued behind an export still wait for its (bounded) duration, which is exactly #4531. On the prod multi-thread runtime the offload only frees one worker thread. **Codex is right: the real fix is to defer the export off the loop, not just offload its CPU.**

### Findings
- **[MAJOR] Defer the export off the loop.** Offload alone doesn't free the contract queue (loop awaits it inline). Spawn the export as a monitored background task that responds via the responder so the dispatch arm returns immediately; bound concurrent exports with a semaphore.
- **[MAJOR] Panic re-raise crashes the whole node.** `runtime.rs:140` `.expect()` on the `spawn_blocking` JoinError re-raises a panic onto the contract-loop task → node top-level select! → whole-node shutdown, and the executor slot leaks. Diverges from the #4441 `compile_offloaded` pattern it cites (which *catches* `is_panic()`). Latent today (export path is panic-free) but a future panic on that path takes the node down instead of failing one request. Catch + convert + return the slot.
- **[MEDIUM] Byte cap checked post-buffer.** `secret_export.rs:485-494` decrypts + buffers all plaintext, *then* checks the 256 MiB cap — a user under the 10k count cap but with large secrets gets multi-GiB buffered before the 413, defeating the memory bound (and contradicting the docstrings that claim incremental checking). Bail incrementally during the gather.
- **[MINOR] Misleading safety comments** (`pool.rs:849-858`, `delegates.rs:15-19`): it's a read-only FS walk over a per-executor `SecretsStore`, not redb rows; `pop_executor` does not serialize other executors. State the real invariant (per-file FS + AEAD auth). Plus the `block_in_place` doc mention the helper never makes.
- **[MINOR] No aggregate export-concurrency cap** — fold a semaphore in with the defer (the endpoint is live).

### Confirmed correct (verified, don't touch)
The typed `TooLarge`→**413** path (no 500 collapse, fixed-string body, zero secret leak), the boundary semantics (strict `>`, at-limit accepted, tested), the count cap's cheap pre-decrypt walk, and the bundle **format/round-trip** stability.

Rework requested; will re-run the Full review on the new revision.

[AI-assisted - Claude]

sanity:
## Re-review of the rework (b8bddcc3) — Codex + concurrency lens. One BLOCKER, rest verified sound.

The deferral rework is the right shape, and the hard parts check out. **Verified leak-free / exactly-once on every path** (success, panic-replacement, lost-to-drop, cancel, busy-reject): the export-semaphore permit, the pool-executor reclaim, `ExportGuard` exactly-once delivery (never zero, never twice, no double-answer), the unbounded resume channel (bounded senders, always drained, no `.send().await`), and the interaction with the existing #4391 deferral. Good work on those.

### BLOCKER — small-pool self-deadlock
`MAX_CONCURRENT_EXPORTS=2 < pool_size` is a **prose comment, never enforced**, and `pool_size` can be 1 or 2 (`FREENET_RUNTIME_POOL_SIZE` clamps [1,16]; the `available_parallelism()-1` default is ≤2 on a ≤3-core host). Exports can then hold every executor; a normal contract event parks the loop in `pop_executor().await`, but executors only return via `finish_export`/`handle_export_resume` **on that same loop** → permanent deadlock of all contract ops. Reachable from clients (the HTTP path has no upstream concurrency cap). Independently confirmed by both reviewers with full interleavings.

Also: the three "defer" tests are **source-scrape pins** (assert source text, not runtime behavior), so none exercises the loop+pool and none would catch this.

### Fix requested
1. Enforce in code (not prose): effective export concurrency = `min(MAX_CONCURRENT_EXPORTS, pool_size - 1)`, set from the real `pool_size` in `RuntimePool::new`, so ≥1 executor is always free for normal ops (pool_size==1 → exports disabled → typed Busy/503).
2. Make export admission non-blocking (try-reserve → 503, never `pop_executor().await` on the loop).
3. Add a **behavioral** test (pool_size=2, export in flight + a normal contract op completes within a timeout; pool_size=1 → export 503), exercising the real loop+pool.

Third real concurrency bug the rigorous re-review caught that the unit tests missed — staying a draft until the deferral is genuinely deadlock-free. Will run a final Full pass on the fix.

[AI-assisted - Claude]

sanity:
## Final pass (Codex) — new P1: deferred export can read a stale secret index → incomplete backup

The deadlock fix is correct (reserve-one invariant + non-blocking admission, behavioral tests now real). But the final pass surfaced a **correctness** issue the defer introduced:

**[P1] Concurrent export vs. secret write → silently incomplete bundle.** `pool.rs:1046-1051`. Now that the export runs off-loop, a normal delegate `store_secret` can run concurrently on a *different* pooled executor. Each `Executor<Runtime>` owns its **own** in-memory `SecretsStore` index (executor.rs:917), so a write updates only the writer's index; an export enumerating (`enumerate_scope`) on a different executor's index can **miss a just-stored secret** and produce an incomplete backup bundle. Impossible pre-PR (the inline export serialized with all contract ops); the defer is what exposes it.

Fix direction: the export's **enumeration** must read from disk (the shared `secrets_dir` source of truth, since secrets are FS blobs) rather than a per-executor in-memory cache — individual reads may fall back to disk, but the *list* from `enumerate_scope` is what goes stale.

This is the 4th distinct concurrency/correctness bug found across the review rounds (offload-didn't-free-loop → panic-crash → small-pool-deadlock → now stale-index). The defer is the right DoS fix but it fights the per-executor-index architecture. Holding as a draft; the maintainer is deciding whether to land the disk-enumeration fix (with a rested final pass) or re-scope. Also worth a separate look: whether the per-executor-index divergence affects normal cross-executor secret ops, not just export.

[AI-assisted - Claude]

sanity:
## Codex pass on the disk-enumeration revision — 2 P2s (robustness), then merge

The core fix is verified correct: per-user isolation holds (User scope walks only `delegate/users/<id>/`, never other users or Local root; `is_file()` excludes the `users/` subdir), `store_secret` writes atomically (tmp+fsync+rename so a concurrent walk never sees a half-written entry), normal reads already go straight to disk (the staleness was export-only), and the two changed index-repair tests are a legitimate semantic update (a backup now includes every completed on-disk write; the repair post-condition is unchanged + still tested). The cross-executor regression test is real.

Two P2s remain, both touching the export's core promises (complete + bounded):
- **Silent partial backup on I/O error** (`secrets_store.rs:1680`): `enumerate_scope_on_disk` returns empty when `read_dir` fails (transient unreadable / permission / mount), so the export would seal a *successful but incomplete* bundle — the silent-incompleteness class we just fixed, via a different path. Fix: treat `NotFound` as expected-absent (empty), propagate any other io::Error so the export fails loud.
- **Count cap pre-rejection allocation** (`secrets_store.rs:1795`): `scope_entry_count` builds the full Vec then `.len()`s it; an over-limit user forces traversal+allocation proportional to their entry count before the 413. Fix: stream and stop at `max_count+1`.

Both contained; fixes requested. Last revision before merge, assuming clean.

[AI-assisted - Claude]

sanity:
## Final review — APPROVED to merge (Full tier; 7 review passes on auth/secrets/executor surface)

This PR took the export endpoint's DoS hardening from "offloads CPU" to a genuinely correct, bounded, concurrency-safe off-loop export. The rigor was warranted: across the review rounds, multiple reviewers (Codex external + adversarial concurrency/cap lenses + maintainer trace) caught **four critical-class bugs and three robustness bugs that the unit tests never surfaced** — node-deadlock, whole-node panic-crash, silent-incomplete-backup, then I/O-error partial backups, count/byte cap allocation, and a count-cap TOCTOU. All addressed.

**Verified-correct on the final HEAD (2fc6a5c7):**
- **Defer off-loop** (#4531): the export no longer blocks the serial contract loop — dispatched to a background task that responds via an `ExportGuard` RAII (exactly-once executor reclaim + client answer on every path, incl. cancel/panic/drop).
- **Deadlock-free**: export concurrency sized from the pool (`min(MAX_CONCURRENT_EXPORTS, pool_size-1)`, enforced in code) so ≥1 executor is always free; admission is fully non-blocking (try-reserve → 503), so the loop can never park on export.
- **Panic-safe**: a panicked offload is caught and converted (mirrors #4441), the executor is replaced, the node lives.
- **Correct + complete backup**: enumeration reads from **disk** (the shared source of truth), so a concurrent `store_secret` on another executor can't be missed; `store_secret` writes atomically (tmp+fsync+rename); an unexpected I/O error **fails loud** rather than sealing a partial bundle.
- **Bounded**: both the count and byte caps are enforced **inline during the gather** in one streaming pass over the shared disk walker — never materializing the attacker-controlled list, no precheck/gather TOCTOU.
- **Isolated**: the per-user disk walk reads only `delegate/users/<user_id>/`; Local scope excludes the `users/` subdir. Per-user isolation is structural.
- **Tests**: behavioral tests now exercise the real loop+pool (concurrent-op-during-export, small-pool-no-deadlock, cross-executor completeness, fail-loud I/O, authoritative count-during-gather), not source-scrape pins. secret_export 21/21, secrets_store 54/54, hosted_mode_export 7/7, hosted_mode_isolation 3/3; fmt + clippy --lib -D warnings clean.

All findings fixed; nothing dismissed. Merging once required CI is green on 2fc6a5c7.

[AI-assisted - Claude]

PR #4564, opened 2026-06-23
fix: deflake delegate node-limit tests under parallel cargo test --lib

### LINKED ISSUE #4418: Flaky: test_create_delegate_non_attested_still_counts_toward_node_limit fails under parallel cargo test --lib
## Summary

`wasm_runtime::delegate_api::tests::test_create_delegate_non_attested_still_counts_toward_node_limit` fails intermittently under a full parallel `cargo test -p freenet --lib` run, but passes in isolation.

## Reproduction

- `cargo test -p freenet --lib` (full suite, parallel threads): occasionally `FAILED` on this test (1 of ~3058).
- `cargo test -p freenet --lib test_create_delegate_non_attested_still_counts_toward_node_limit` (isolated): passes deterministically.
- Observed 2026-06-14 on a feature branch whose diff does not touch `wasm_runtime` / delegate code at all (the same suite passed 3058/3058 on a slightly earlier build), which points to a shared-state interaction rather than a code regression.

## Hypothesis

Shared mutable global state (a per-node delegate/node-limit counter) that other tests in the suite mutate, so under threaded parallel execution the count this test asserts on is perturbed by a concurrently-running test. CI's per-process isolation (nextest) avoids it — which is why CI is green — so only threaded in-process `cargo test --lib` hits it.

## Why it matters

Per the project flaky-tests policy, this is a real latent bug: a test relying on global state that isn't isolated per test. It doesn't fail CI today (process isolation), but it makes local `cargo test --lib` unreliable and would surface if the harness changed.

## Suggested direction

Isolate the node-limit counter per-test (reset in setup/teardown), or require serial execution for the delegate node-limit tests. Investigate which global the counter lives in and whether sibling delegate tests leak it.

[AI-assisted - Claude]

mvanhorn:
## What

Deflakes `test_create_delegate_non_attested_still_counts_toward_node_limit` (and its sibling delegate-creation tests) under a threaded `cargo test -p freenet --lib` run.

## Why

`test_create_delegate_non_attested_still_counts_toward_node_limit` snapshots the process-global `CREATED_DELEGATES_COUNT` atomic into `before`, creates a delegate, asserts `after > before`, then restores the counter with `fetch_sub(1)`. Because the counter is a process-global static, other delegate-creation tests in the same binary run concurrently on other threads and call `CREATED_DELEGATES_COUNT.fetch_add(1)` inside `create_delegate_sync`, perturbing the `before`/`after` snapshot and racing the cleanup. The test passes in isolation but fails intermittently under the full parallel `--lib` run.

CI uses nextest per-process isolation, so each test gets its own process and never observes the race; only a threaded `cargo test --lib` exercises it.

## How

Group the delegate-creation tests that read or mutate the shared `CREATED_DELEGATES_COUNT` global into a single named serial group via `#[serial_test::serial(created_delegates_count)]`, so they cannot run concurrently with one another. The named group serializes only these tests against each other; it does not globally serialize the rest of the suite, keeping the wall-clock impact minimal. `serial_test` is already a workspace dev-dependency.

No production code changes — the fix is confined to the test module in `crates/core/src/wasm_runtime/delegate_api.rs`.

## Verification

- `cargo build -p freenet --lib` — passes
- `cargo test -p freenet --lib wasm_runtime::delegate_api` — 43 passed, 0 failed
- `cargo fmt -p freenet --check` — clean

Fixes #4418

### DIFF
--- a/crates/core/src/wasm_runtime/delegate_api.rs
+++ b/crates/core/src/wasm_runtime/delegate_api.rs
@@ -1013,6 +1013,7 @@ mod tests {
 
     /// create_delegate_sync rejects creation when depth limit is exceeded.
     #[tokio::test]
+    #[serial_test::serial(created_delegates_count)]
     async fn test_create_delegate_depth_exceeded() {
         let mut env_holder = TestEnv::new().await;
 
@@ -1033,6 +1034,7 @@ mod tests {
 
     /// create_delegate_sync allows creation at depth just below the limit.
     #[tokio::test]
+    #[serial_test::serial(created_delegates_count)]
     async fn test_create_delegate_depth_just_under_limit() {
         let mut env_holder = TestEnv::new().await;
 
@@ -1053,6 +1055,7 @@ mod tests {
 
     /// create_delegate_sync rejects creation when per-call limit is exceeded.
     #[tokio::test]
+    #[serial_test::serial(created_delegates_count)]
     async fn test_create_delegate_per_call_limit_exceeded() {
         let mut env_holder = TestEnv::new().await;
 
@@ -1082,6 +1085,7 @@ mod tests {
     /// Superseded: WASM validation moved from creation to execution time.
     /// Replaced test_create_delegate_invalid_wasm which expected InvalidWasm error.
     #[tokio::test]
+    #[serial_test::serial(created_delegates_count)]
     async fn test_create_delegate_accepts_any_bytes_at_creation() {
         let mut env_holder = TestEnv::new().await;
 
@@ -1101,6 +1105,7 @@ mod tests {
 
     /// create_delegate_sync rejects WASM bytes exceeding the size limit.
     #[tokio::test]
+    #[serial_test::serial(created_delegates_count)]
     async fn test_create_delegate_rejects_oversized_wasm() {
         let mut env_holder = TestEnv::new().await;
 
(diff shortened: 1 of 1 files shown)


mvanhorn:
The Rule Lint is failing because this `fix:` PR doesn't add a new `#[test]`. In this case it genuinely can't: the entire change adds `#[serial_test::serial(created_delegates_count)]` to 8 existing tests that share the global `CREATED_DELEGATES_COUNT` counter and were flaking under parallel `cargo test --lib`. There's no new production code path to cover, and a contrived extra test wouldn't exercise anything meaningful. Per the lint's guidance, could a maintainer apply the `test-exempt` label? I don't have label permissions. Happy to adjust if you'd prefer a different approach.

sanity:
Thanks for this, and apologies for the long delay in responding — that's on us.

This one has been overtaken by later work on `main`. The same flake was re-filed as #4813 and fixed there by removing the shared state rather than serializing around it: `CREATED_DELEGATES_COUNT` and `DELEGATE_INHERITED_ORIGINS` are no longer process-global statics. The counter is now per-`TestEnv` injected state, so the tests read their own env's counter and can't perturb each other — `test_create_delegate_non_attested_still_counts_toward_node_limit` now asserts an exact count of 1 instead of the old `after > before` workaround.

Your diagnosis was exactly right, including the detail that CI's nextest per-process isolation is why this never showed up in CI and only a threaded `cargo test --lib` could see it. That framing is now written into the code comments on `main`.

Closing as superseded, and I'll close #4418 as a duplicate of #4813. Sorry again for the wait — a small, well-diagnosed test fix shouldn't sit this long.

[AI-assisted - Claude]

PR #4568, opened 2026-06-23
feat(telemetry): observe placement-migration effect via router_snapshot gauges

sanity:
## Problem

The SubscribeHint placement migration (#4404) nudges a closer-to-the-key neighbor to subscribe-and-host a contract, so hosting should drift toward each contract's ideal ring location over time. But that effect is **invisible to telemetry** — inbound `SubscribeHint` is classified `EventKind::Ignored` (`tracing/register.rs`), and nothing reports whether hosting is actually getting closer to contract keys. So we can verify the migration is *safe* (broadcast-failure metrics) but not whether it's *working*.

## Approach

Two cheap, low-overhead observability primitives, both **piggybacked on the existing 5-minute `router_snapshot`** so collector volume stays flat (no new per-message events):

1. **Placement-quality gauge** — per node, the distribution of ring-distance from this node to the keys of the contracts it hosts: `hosted_contracts_count`, `hosted_key_distance_{median,p90,min,mean}`, and `hosted_key_distance_frac_within_0_1` (fraction within 0.1 — the clearest single "are hosted contracts near their host" number). If the migration works, this distribution **tightens over time**, network-wide.
2. **Migration-activity counters** — `subscribe_hint_{sent,received,acted}` cumulative counters, incremented at the three migration sites (send in `migration.rs`; received before the gates and acted on the directed-subscribe branch in `node.rs`).

Purely additive observability: **no change to the wire protocol, the `SubscribeHint` type, the migration logic, or the version floor.** The distance computation runs over an owned `Vec` of hosted keys (no lock held across it), uses the in-tree wrap-aware `Location::distance`, and skips gracefully when the node has no location yet.

No collector/extractor change: the `router_snapshot` OTLP body is a hand-mirrored `json!` block that the collector stores raw, so the nine new fields flow through and are queryable directly from `logs.jsonl` (pinned by a test, matching the existing fd/module-cache gauge convention).

## Testing

- New `placement_migration_metrics` module: 8 unit tests — placement-quality math (empty / single / known distribution / wrap-around / far-contracts), nearest-rank percentile endpoints, counter increments, and a **source-pin test** asserting the three `record_*` sites exist exactly once and that `record_acted` sits on the act branch (not at the top with `received`), so a refactor that miscounts fails the build.
- `router_snapshot_json_includes_placement_gauges` pins the nine fields into the OTLP body.
- `cargo clippy -p freenet --tests -- -D warnings` clean; affected module suites green (location/router/telemetry/hosting).

Two tiny drive-by clippy fixes (`isotonic_estimator.rs`, `tests/hosted_mode_export.rs`) for pre-existing Rust 1.94 lints that blocked `clippy --tests` — included rather than left as landmines.

Observability follow-up to #4404 / the SubscribeHint re-enable.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1758,6 +1758,13 @@ where
             return Ok(());
         }
         NetMessageV1::SubscribeHint(hint) => {
+            // Placement-migration telemetry (#4404 follow-up): count every inbound
+            // hint, before any admission gate, so `received` is the true arrival
+            // rate and `received - acted` is the gated/dropped fraction.
+            op_manager
+                .ring
+                .placement_migration_metrics()
+                .record_received();
             // Placement-migration version gate. The migration is re-enabled at
             // floor `(0, 2, 80)` (#4499 made it load-safe). The SEND side
             // (`p2p_protoc::peer_supports_subscribe_hint`) gates emission on the
@@ -1865,6 +1872,11 @@ where
                 ?source_addr,
                 "Received SubscribeHint — starting directed subscribe to holder"
             );
+            // Placement-migration telemetry (#4404 follow-up): count only the
+            // hints we actually act on (all gates passed), i.e. the migrations
+            // that actually start a directed subscribe and thereby host the
+            // contract closer to its key.
+            op_manager.ring.placement_migration_metrics().record_acted();
             subscribe::start_directed_subscribe(op_manager.clone(), hint.key, hint.holder);
             return Ok(());
         }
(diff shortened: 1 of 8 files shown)


sanity:
Codex review (external, non-Claude) found one [P1]: `HostedKeyDistanceStats.count` was read only from `#[cfg(test)]` tests, so a production build would hit `dead_code` under `-D warnings` (my local `clippy --tests` missed it because `--tests` reads the field). **Fixed** as Codex suggested — the snapshot emitter now reads `stats.count` for `hosted_contracts_count`. Re-verified: `cargo clippy -p freenet -- -D warnings` (production, no --tests) and `--tests` both clean, 8/8 placement tests pass. The fix is the only change since review; rest of the diff is unchanged.

[AI-assisted - Claude]

PR #4570, opened 2026-06-24
fix: move StartLimitBurst/StartLimitIntervalSec into [Unit] section of generated systemd units

### LINKED ISSUE #4551: systemd unit: StartLimitBurst/StartLimitIntervalSec are in [Service], where systemd ignores them
## Problem

The generated systemd unit template (`crates/core/src/bin/commands/service/linux.rs`, both the user and system templates) emits:

```
[Service]
...
StartLimitBurst=5
StartLimitIntervalSec=120
```

`StartLimitBurst` and `StartLimitIntervalSec` are **`[Unit]`-section** directives (moved there in systemd v230, 2016). When placed in `[Service]`, modern systemd **silently ignores them** (logs a warning and drops them). So the start-rate limiter is effectively disabled: a crash/wedge loop is NOT rate-limited and the unit is never auto-stopped after N failures in the interval.

Found during review of #4550 (the #4549 gateway-wedge fast-exit fix): the fast-exit emits a counted exit code for a boot-wedge on the assumption that `StartLimitBurst` would eventually stop a tight loop — which doesn't hold while the directives sit in `[Service]`. #4550 therefore reverted that exit-code split and always exits 42; restoring a meaningful min-uptime exit-code guard depends on this fix.

## Fix

- Move `StartLimitBurst` / `StartLimitIntervalSec` into the `[Unit]` section of **both** templates in `service/linux.rs`.
- Add a unit-generation test that parses the generated unit string and asserts both keys appear in `[Unit]` (after `[Unit]`, before `[Service]`).
- Note: this is deployment/restart-semantics config (high-risk surface) and only takes effect on the next `freenet service install` / unit regeneration, so currently-deployed nova/vega units are unaffected until re-registered. Run the full multi-model review for the change.
- Follow-up: once the limiter works, reconsider a min-uptime exit-code guard so a tight boot-wedge loop exits with a *counted* code (gets burst-stopped) while a long-uptime wedge keeps the burst-exempt auto-update code 42.

[AI-assisted - Claude]

mvanhorn:
## Summary

The generated systemd unit files now place `StartLimitIntervalSec` and `StartLimitBurst` in the `[Unit]` section, where systemd actually reads them, instead of `[Service]`. A regression test covers the generated unit content for both unit variants.

## Why this matters

systemd only honors the start-limit directives when they appear in `[Unit]`; with them in `[Service]`, systemd logs a warning and the intended restart rate-limiting never takes effect, so a crash-looping service is not throttled as configured. Moving the directives to the correct section makes the limit apply. Reported in #4551.

## Testing

Added regression tests asserting the generated unit files carry the start-limit directives in the `[Unit]` section for both generated unit variants.

Closes #4551

AI was used for assistance.

### DIFF
--- a/crates/core/src/bin/commands/service/linux.rs
+++ b/crates/core/src/bin/commands/service/linux.rs
@@ -264,6 +264,11 @@ Description=Freenet Node
 Documentation=https://freenet.org
 After=network-online.target
 Wants=network-online.target
+# Stop restart loop after 5 failures in 2 minutes (e.g., port conflict with
+# a stale process). Without this, systemd restarts indefinitely.
+# SuccessExitStatus=42 ensures auto-update exits don't count as failures.
+StartLimitBurst=5
+StartLimitIntervalSec=120
 
 [Service]
 Type=simple
@@ -296,11 +301,6 @@ ExecStart={binary} network
 Restart=always
 # Wait 10 seconds before restart to avoid rapid restart loops
 RestartSec=10
-# Stop restart loop after 5 failures in 2 minutes (e.g., port conflict with
-# a stale process). Without this, systemd restarts indefinitely.
-# SuccessExitStatus=42 ensures auto-update exits don't count as failures.
-StartLimitBurst=5
-StartLimitIntervalSec=120
 # Allow 45 seconds for graceful shutdown before SIGKILL.
 # The node handles SIGTERM by (1) waiting up to `shutdown-drain-secs`
 # (default 30s) for in-flight client PUT/GET/UPDATE/SUBSCRIBE drivers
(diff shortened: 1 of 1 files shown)


sanity:
Thanks @mvanhorn — this is exactly the fix for #4551 (the `StartLimit*` directives were being silently ignored in `[Service]`). I added the `fix:` conventional-commit prefix to the title to unblock CI; the change itself is clean (correct section move + good regression tests). Reviewing now as part of the 0.2.84 auto-update-safety work. The complementary crash-loop **auto-rollback** and a min-uptime guard will come as a separate follow-up so this focused fix can land on its own.

[AI-assisted - Claude]

sanity:
Reviewed and approving. The section move is correct (StartLimit* belong in [Unit]; systemd-analyze confirms the old [Service] placement reproduces `Unknown key name 'StartLimitIntervalSec' in section 'Service', ignoring`), the regression tests assert placement by section, and an external Codex pass found no introduced issues. CI green. This is the P0 for #4551 — it makes the crash-loop rate-limit actually take effect. The complementary crash-vs-update guard (distinct fast-crash exit code + min-uptime) lands as a separate follow-up. Thanks @mvanhorn. Merging.

[AI-assisted - Claude]

PR #4572, opened 2026-06-24
perf(contract): off-load Background summarize from the serial contract loop

iduartgomez:
## Problem

`GetSummaryQuery` ran `summarize_contract_state().await` **inline on the single-threaded contract event loop**. The periodic InterestSync interest-exchange emits these as `Priority::Background` in high volume, so a burst of them serialized on the loop behind (and ahead of) client GET/PUT/UPDATE — the "contract queue full, try again later" client outage (#4534).

#4535 priority-tiered the queue so client work *drains* first, but priority controls ordering only: a Background summary already at the loop head still parks the loop for its full execution duration regardless of what's queued behind it. This PR fixes that residual head-of-line stall.

## Solution

Off-load the Background summary execution onto a spawned task, mirroring the just-landed #4531/#4381 P5 hosted-export off-load so the pool has **one** off-loading idiom, not two.

- **pool.rs**: `try_begin_summarize` (non-blocking, via P5's `try_pop_executor`) admits a `SummaryJob` owning the checked-out executor; `finish_summarize` returns/replaces it on the loop. Reuses P5's `return_checked` / `create_replacement_executor` — **no new executor-return channel**.
- **Shared off-loop budget (key correctness point):** exports (#4531 P5) and summaries (#4534) draw from **one** `offload_semaphore` sized `effective_offload_permits = min(MAX_CONCURRENT_EXPORTS + MAX_CONCURRENT_SUMMARIES, pool_size-1)`. Two *independent* per-class semaphores each sized `pool_size-1` do **not** compose: on a small pool (`pool_size` 2–4 → 3–5 core nodes, since `pool_size = available_parallelism()-1`) exports + summaries could jointly hold every executor, and a foreground op then parking on `pop_executor().await` would wait for an executor only the now-parked loop can return — a self-deadlock. The shared budget keeps **≥1 executor reserved for foreground** regardless of how off-loop work splits between the classes. (#4391's related-fetch deferral is *not* a third class — it releases its executor before the off-loop network wait.)
- **contract.rs**: `SummaryResume`/`SummaryGuard` give exactly-once delivery (drop = error resume → client always answered, executor always reclaimed) via a `summary_resume_tx` on `DeferralCtx`, drained every loop turn. The `GetSummaryQuery` arm off-loads **only** `Priority::Background` when a `DeferralCtx` + free executor are present; foreground summaries (NetworkRelay/ClientLocal — resync / a direct client ask) stay inline for latency. Mock / single-executor / delegate-driven paths inherit the trait default (`Unsupported`) and summarize inline, so the simulation runner stays deterministic and fully inline.
- **fair_queue::pop()** now returns the popped event's `Priority`, threaded into `handle_contract_event` so the arm can route on it.

**Correctness-neutral:** a summary computes the same value on or off the loop; the only change is whether it blocks the loop while doing so.

## Testing

- `effective_offload_permits_always_reserves_one_executor` — asserts the **joint** cap (`exports + summaries ≤ pool_size-1` for all n) — the small-pool deadlock guard.
- `dropped_summary_guard_answers_client_with_error` / `summary_guard_success_sends_exactly_one_resume` — behavioral exactly-once for `SummaryGuard` (drop delivers one error resume + answers the client; success send makes Drop a no-op), mirroring the existing `ResumeGuard` tests.
- `finish_summarize_replaces_panicked_executor` — panic-lost executor is replaced.
- `get_summary_arm_offloads_background` — source-scrape pin that the arm gates on Background, admits via `try_begin_summarize`, spawns the job.
- contract (501) + export (33) + node suites green; clippy clean.

Reviewed by a multi-perspective pass (code-first / testing / skeptical / big-picture); the deadlock window above was caught in review and fixed via the shared budget, then re-reviewed clean.

## Fixes

Refs #4534 (third installment of the remediation: priority queue #4535 → migration-admission gate #4537 → this off-load).

https://claude.ai/code/session_01JZGPSEa8g1Z8Nomv528c8g

### DIFF
--- a/.claude/rules/operations.md
+++ b/.claude/rules/operations.md
@@ -107,6 +107,18 @@ The loop's pop() applies a bounded anti-starvation floor
 (RELAY_STARVATION_FLOOR) so sustained ClientLocal load cannot indefinitely
 starve NetworkRelay/Background — but that is a backstop, not a license to
 mis-tag.
+
+Beyond reordering, a Background GetSummaryQuery (the periodic interest-sync
+summarize) also runs OFF the serial loop on a spawned task when a pooled
+executor is free (#4534, contract.rs GetSummaryQuery arm → pool
+try_begin_summarize/finish_summarize), so a burst of them can't head-of-line
+block client work even once popped. Foreground summaries
+(NetworkRelay/ClientLocal — resync / a direct client ask) stay inline for
+latency. This is the same off-load idiom as the hosted-export deferral (#4531
+/ #4381 P5) and the related-fetch deferral (#4391, contracts.md); all
+off-loop classes share ONE pool-executor budget (effective_offload_permits =
+min(exports+summaries, pool_size-1)) so they can never jointly hold every
+executor and wedge the loop.
 ```
 
 ## Critical Invariant: Initialize-Before-Send
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -326,6 +326,98 @@ impl Drop for ExportGuard {
     }
 }
 
+/// Loop-side payload for a completed off-loop `Background` summary (#4534).
+/// Mirrors [`ExportResume`]: the executor + precomputed summary come back to the
+/// loop, which returns/replaces the executor (`finish_summarize`) and answers
+/// the parked client. Nothing to re-run on the loop — just deliver and reclaim.
+struct SummaryResume {
+    /// The contract whose summary was computed (needed both for the
+    /// `GetSummaryResponse` and for `finish_summarize`'s per-contract accounting).
+    key: ContractKey,
+    /// The completed summary: the executor to return (or `None` if the task
+    /// panicked/was dropped) and the result for the client.
+    done: crate::contract::executor::runtime::SummaryDone,
+    /// The parked client responder. `None` only if the client disconnected
+    /// before the summary finished.
+    responder: Option<StashedResponder>,
+}
+
+/// RAII guard guaranteeing the off-loop summary task delivers EXACTLY ONE
+/// [`SummaryResume`] — whether the summary completes or the task is
+/// dropped/panics/cancelled before sending. Same load-bearing invariant as
+/// [`ExportGuard`]: the parked client responder never hangs and the
+/// checked-out executor is always reclaimed.
+struct SummaryGuard {
+    payload: Option<SummaryGuardPayload>,
+}
+
+struct SummaryGuardPayload {
+    summary_resume_tx: tokio::sync::mpsc::UnboundedSender<SummaryResume>,
+    key: ContractKey,
+    responder: Option<StashedResponder>,
+}
+
+impl SummaryGuard {
+    fn new(
+        summary_resume_tx: tokio::sync::mpsc::UnboundedSender<SummaryResume>,
+        key: ContractKey,
+        responder: Option<StashedResponder>,
+    ) -> Self {
+        Self {
+            payload: Some(SummaryGuardPayload {
+                summary_resume_tx,
+                key,
+                responder,
+            }),
+        }
+    }
+
+    /// Deliver the completed summary. Consumes the payload so a subsequent `Drop`
+    /// is a no-op (exactly-once).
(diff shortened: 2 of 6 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4572

### Summary
- **Type:** perf (no behavior change; correctness-neutral off-load)
- **Review tier:** Full (high-risk surface: concurrency/async + contract WASM execution path)
- **Reviewers run:** code-first, testing, skeptical, big-picture (4 specialist subagents) + a deadlock-focused skeptical re-review after the fix. External model pass (codex) was **unavailable** (auth-token expired, 401); gemini not installed. Per the multi-model-review fallback, substituted the 4 diverse-Claude-lens specialist reviewers (distinct adversarial lenses, each blind to the others, each reading the checked-out code) — recorded here for transparency.
- **HEAD reviewed:** 28f8c38cb

### What the reviewers found

The first-pass review surfaced **one blocking defect** (caught by the code-first lens, confirmed independently): a **combined export+summary executor-exhaustion deadlock on small pools**. The original commit added the Background-summary off-load with its *own* semaphore alongside the already-merged export off-load's semaphore — two independent budgets each sized `pool_size-1`. They don't compose: on `pool_size` 2–4 (3–5 core nodes) they could jointly check out every executor, and a foreground op then parking on the blocking `pop_executor().await` would wait for an executor only the now-parked loop can return → self-deadlock, no timeout.

**Fixed** by collapsing both classes onto one shared `offload_semaphore` sized `min(MAX_CONCURRENT_EXPORTS + MAX_CONCURRENT_SUMMARIES, pool_size-1)`, guaranteeing ≥1 executor is always reserved for foreground regardless of the export/summary split. A new test asserts the joint cap for all n.

### Re-review (deadlock fix) — clean

A full adversarial re-review of the fix verified: (1) joint invariant enforced (`eff_offload(n) ≤ n-1` ∀n); (2) export tightening is availability-only, no correctness regression; (3) permit accounting correct on panic/replacement (offload permit and executor return on disjoint paths, no double-release); (4) the two new `SummaryGuard` tests are non-vacuous end-to-end; (5) **#4391's related-fetch deferral is NOT a third executor-holding class** (it releases its executor before the off-loop network wait), so the shared budget covers every executor-holding off-loop class. No remaining deadlock path.

### Testing Assessment
- Initially the off-load was pinned only by source-scrape (matching the export precedent). Review added **behavioral** exactly-once tests for `SummaryGuard` (drop → one error resume + client answered; success send → Drop no-op), mirroring the proven `ResumeGuard` tests. `finish_summarize` panic-replacement pinned; joint-cap math behaviorally tested.
- contract (501) + export (33) + node suites green.

### Big Picture
- Goal-aligned: real residual gap (#4535 controls ordering, not the duration a popped Background item parks the loop). Third installment of the #4534 remediation.
- No scope creep: the `fair_queue::pop()` → Priority threading is the minimal mechanism for the arm to route Background; all other edits mechanical.
- Genuinely rides the P5 idiom — no parallel machinery (the earlier custom executor-return channel from a prior branch is fully gone).
- operations.md updated to document the off-load + shared-budget invariant.

### Recommendations
**Must Fix:** none remaining (the deadlock was fixed + re-reviewed).
**Consider (non-blocking):** if 2-executor production nodes are common and export 503s under concurrent summarize load become noticeable, revisit the `MAX_CONCURRENT_EXPORTS + MAX_CONCURRENT_SUMMARIES` sum vs typical pool sizes — a tuning question, not a correctness defect.

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** 28f8c38cb

[AI-assisted - Claude]

iduartgomez:
Closing: bisect proved the summarize off-load (the only delta vs the merged #4535 priority queue) causes the NAT/connectivity failures — off-loaded WASM summarize does spawn_blocking+block_in_place on a spawned task concurrently with the contract loop, starving workers on small CI containers and stalling cross-peer PUT propagation. The #4534 client outage is already fixed and validated by #4535 (priority queue) + #4537 (migration-admission gate), both merged in 0.2.81. The off-load is a marginal extra optimization not worth the concurrency risk, so dropping it.

PR #4573, opened 2026-06-24
feat(hosted): per-user secret storage quota (#4561 P5)

sanity:
## Problem

Hosted mode (#4381) lets untrusted web visitors store secrets (identity/room keys) in per-user delegate-secret namespaces on a shared node. Nothing bounded a single visitor's footprint, so one user could fill the node's disk by writing unbounded per-user secrets. #4561 (P5) asks for a per-user storage quota.

## Solution

Bound the total on-disk ciphertext a single hosted user (`UserId`) can store across all their delegate secrets. **REJECT-on-full, never evict** — per-user secrets are authoritative identity/room keys, not a reconstructible cache. Default **4 MiB/user**, operator-configurable, `0` disables. **Local (non-hosted) scope is completely unaffected** — `SecretScope::Local` is never quota-checked, preserving the single-user invariant.

Key pieces:

- **Enforcement** in `SecretsStore::store_secret`, for `SecretScope::User` only, BEFORE the filesystem write. Computes the new blob's exact on-disk size and rejects with a new `SecretStoreError::QuotaExceeded { used, limit, attempted }` (non-secret byte counts only) when `current_total - old_blob_size + new_blob_size > limit`. Overwrites are charged only their delta.
- **Rejection propagation:** `QuotaExceeded` flows out of `store_secret` (`RuntimeResult`) through `native_api::set_secret`'s existing `Err(_) => ERR_STORAGE_FAILED` arm — the delegate sees `set_secret` fail, no new path invented.
- **Shared accounting:** a **process-global** `LazyLock<QuotaTracker>` (a `DashMap<UserId, AtomicU64>`) so all pooled executors — each of which builds its own `SecretsStore` — share one per-user counter. Seeded **lazily from disk** on first touch (sums on-disk blob sizes via the existing `walk_scope_secrets_on_disk` helper), O(1) thereafter. `remove_secret` decrements by the removed blob's on-disk size.
- **Config:** `DEFAULT_PER_USER_SECRET_QUOTA_BYTES = 4 MiB`; `per_user_secret_quota_bytes` field on `ConfigArgs`/`Config` + `--per-user-secret-quota=BYTES` CLI flag (`0` disables). Merged in `ConfigArgs::build()` and covered by the `all_persisted_config_fields_round_trip_through_build` guard.

## Why a process-global tracker (not Arc threaded through construction)

Each pooled `Executor<Runtime>` builds its own `SecretsStore` (per-executor in-memory index — the same divergence that bit the export in #4563). The per-user totals must therefore live in ONE structure every store references. A process-global `LazyLock<DashMap>` is the **established pattern in this exact subsystem** (`native_api.rs` has several `LazyLock<DashMap<...>>` statics for precisely this "shared across all pooled executors" need), requires **zero changes to the ~50 `SecretsStore::new` call sites**, and the `UserId` key is a globally-unique 32-byte hash so a single global namespace is collision-free. The configured *limit* is a per-store field set from `Config` via a `with_user_quota` builder (every production store reads the same config, so they all carry the same limit), keeping "shared counters" and "configured limit" cleanly separated.

Delegate execution is serial on the contract loop (only one `store_secret` runs node-wide at a time), so no CAS is needed; `AtomicU64` is cheap insurance and the serial assumption is documented.

## On-disk size constant

A stored blob is `[VERSION_V1 (1)][XNonce (24)][AEAD ciphertext + Poly1305 tag (16)]` = `plaintext.len() + 41` bytes (`HEADER_LEN = 25` + 16-byte tag). Enforcement sizes against the composed ciphertext (`ciphertext.len()`), which is exactly what `metadata().len()` reports for the file, so the running counter and the lazy disk seed always agree.

## Testing

New unit tests in `secrets_store.rs`:
- under-quota write succeeds; over-quota write rejected with `QuotaExceeded` + not written + tracker unchanged
- `QuotaExceeded` propagates as a storage failure (the variant `set_secret` maps to `ERR_STORAGE_FAILED`)
- `remove_secret` frees room so a previously-rejected write now succeeds
- lazy disk seed matches a hand-built on-disk layout (sums existing blobs)
- **two `SecretsStore`s sharing the tracker see each other's accounting** (shared-counter property)
- `SecretScope::Local` is never quota-checked (large Local write past the limit succeeds)
- `quota = 0` disables enforcement (tracker stays empty)
- overwrite charges only the delta (grow and shrink)

Plus the config round-trip guard test covers the new persisted field.

`cargo fmt`, `cargo clippy -p freenet --lib -- -D warnings` (clean), and the new + existing secrets_store tests.

Refs #4561, #4381.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -121,12 +121,16 @@ pub struct ConfigArgs {
     pub max_hosting_storage: Option<u64>,
 
     /// Per-user secret-storage quota in bytes for HOSTED mode (#4561, P5 of
-    /// #4381). Bounds the total on-disk ciphertext a single hosted user (one
-    /// `userToken`) can store in their delegate secrets, summed across every
-    /// delegate, so a visitor cannot fill the node's disk. REJECT-on-full
-    /// (never evict — secrets are authoritative identity/room keys, not a
-    /// cache). Default: 4 MiB. `0` disables enforcement. Has NO effect outside
-    /// hosted mode — local single-user secrets are never quota-checked.
+    /// #4381). Bounds a single hosted user's (one `userToken`) TOTAL on-disk
+    /// footprint under their `users/<user_id>/` tree, summed across every
+    /// delegate — both the active secret-value blobs AND the `.keys`
+    /// enumeration registry (so many/large keys are charged too) — so a visitor
+    /// cannot fill the node's disk. Per-user secret-value snapshots are disabled
+    /// (hosted users are transient and don't need overwrite history), so there
+    /// is no `.snapshots/` growth to charge. REJECT-on-full (never evict —
+    /// secrets are authoritative identity/room keys, not a cache). Default:
+    /// 4 MiB. `0` disables enforcement. Has NO effect outside hosted mode —
+    /// local single-user secrets are never quota-checked (and keep snapshots).
     #[arg(long = "per-user-secret-quota", env = "PER_USER_SECRET_QUOTA")]
     pub per_user_secret_quota_bytes: Option<u64>,
 
@@ -1015,10 +1019,12 @@ pub struct Config {
     )]
     pub max_hosting_storage: u64,
     /// Per-user secret-storage quota in bytes for hosted mode (#4561, P5 of
-    /// #4381). Bounds the total on-disk ciphertext a single hosted user can
-    /// store in their delegate secrets, summed across delegates. REJECT-on-full
-    /// (never evict). Default 4 MiB; `0` disables. No effect outside hosted
-    /// mode (local single-user secrets are never quota-checked).
+    /// #4381). Bounds a single hosted user's TOTAL on-disk footprint (active
+    /// secret-value blobs + the `.keys` enumeration registry) under their
+    /// `users/<user_id>/` tree, summed across delegates. Per-user value
+    /// snapshots are disabled, so there is no `.snapshots/` growth to charge.
+    /// REJECT-on-full (never evict). Default 4 MiB; `0` disables. No effect
+    /// outside hosted mode (local single-user secrets are never quota-checked).
     #[serde(
         default = "default_per_user_secret_quota_bytes",
         rename = "per-user-secret-quota"
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -120,6 +120,16 @@ pub struct ConfigArgs {
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
 
+    /// Per-user secret-storage quota in bytes for HOSTED mode (#4561, P5 of
+    /// #4381). Bounds the total on-disk ciphertext a single hosted user (one
+    /// `userToken`) can store in their delegate secrets, summed across every
+    /// delegate, so a visitor cannot fill the node's disk. REJECT-on-full
+    /// (never evict — secrets are authoritative identity/room keys, not a
+    /// cache). Default: 4 MiB. `0` disables enforcement. Has NO effect outside
+    /// hosted mode — local single-user secrets are never quota-checked.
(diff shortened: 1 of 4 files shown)


sanity:
## Full review (Codex + adversarial lens) — one MAJOR, rest verified sound

**Verified correct (don't touch):** the byte accounting (overwrite charges only the delta; the net counter adjustment is applied exactly once *after* the rename commits and skipped on every failure path; remove reads size before delete + saturates); the seed (sums across delegates, fails loud on I/O error); the **single-user invariant** (Local provably never produces a User scope, so a normal node is never quota-checked — confirmed end to end); the **import path is operator-only** (CLI, node-stopped — not a client bypass); and the config round-trips through the guard test.

**[MAJOR] The quota only charges active secret-value blobs, but two other per-user on-disk vectors under `users/<user_id>/` are uncounted**, so a hosted user can exceed `limit` in real disk bytes:
- `.snapshots/` — every overwrite snapshots the prior ciphertext; the quota walk skips that dir. Bounded by the retention policy (~62×/secret long-tail), so not unbounded, but a real multiplier.
- `.keys` registry — `register_key` appends raw key bytes; many/large distinct keys with tiny values grow it uncharged and unbounded.
Fix requested: bound the user's **total footprint** under their tree — disable snapshots for User scope (transient hosted users don't need rollback history; tmp+rename already gives crash-atomicity) and count the `.keys` registry, OR count the whole subtree. The seed must include whatever is counted so a restart re-derives the same total.

**Minors:** a load-bearing comment that the no-CAS tracker safety rests on delegate execution being *serial* (don't parallelize `process()` without making it a CAS); and test gaps (`==limit` boundary, multi-delegate seed, the footprint vectors).

Reworking; will re-review. CI green so far.

[AI-assisted - Claude]

sanity:
## Codex re-review of the footprint fix — core accepted, two P2 edge-cases

The footprint accounting is sound: per-user snapshots disabled (verified to be rollback history, not crash-recovery), `.keys` counted with an exact post-commit re-stat, value-blob delta committed only after the rename. Two contained P2s remain:

- **[P2] Multi-store tracker collision.** The process-global tracker keys on `UserId` alone, so two `SecretsStore`s with different `base_path`s in one process (multi-node sim) + the same token collide on one counter. Production-safe (one secrets_dir/node) but a real sim/robustness gap → key by `(secrets_dir, UserId)`.
- **[P2] Orphan `.keys` under-count.** The seed only counts `.keys` in dirs the active-blob walk visits; a user dir with a `.keys` registry but zero active blobs (crash between last-value-delete and `deregister_key`) is missed → post-restart under-count → over-admission. A real (narrow) under-enforcement → the seed must scan orphan `.keys` too.

Both are contained; #2 is an actual under-enforcement so worth closing. Reworking; will re-review. CI green so far.

[AI-assisted - Claude]

sanity:
## Review complete — cleared to merge

**Independent review summary across the PR's life:**
- Rounds 1→3 (core accounting + whole-footprint fix): full adversarial Claude lens + **two real Codex passes**, which found and drove the footprint P1 and the two P2s — all fixed.
- Round 4 (this delta: `(base_path, UserId)` re-keying + orphan-`.keys` seed): **Codex was rate-limited** (tried twice via `codex review`, hit the usage limit, resets 4:41 PM CDT). Per the review policy I used the documented multi-Claude fallback — **three independent, diverse lenses**, all clean:
  - *skeptical/adversarial*: key-threading consistency + orphan-pass double-count/miss (verified against stdlib `DelegateKey::encode()`);
  - *code-first*: intent-vs-impl, all six production tracker call sites thread `self.base_path`, doc claims verified;
  - *exhaustive keying audit*: enumerated every `USER_QUOTA_TRACKER` access — all key on `(self.base_path, this user)`, no UserId-only key survives, test isolation clean, 69 tests pass.

The substantive logic had real external (Codex) review in rounds 2–3; only the mechanical round-4 fixes (which address Codex's own findings) used the fallback. CI green on the 10 CI jobs.

[AI-assisted - Claude]

PR #4574, opened 2026-06-24
DIAG: NAT baseline on pre-PR main (do not merge)

iduartgomez:
Diagnostic only. Triggers NAT Validation on the exact pre-PR main commit (922e40bd9) to determine whether the NAT failures on #4572 are pre-existing flake or introduced by that PR. Close after CI.


iduartgomez:
Diagnostic PR — done. Confirmed pre-PR main is green on NAT (baseline). Closing.

PR #4575, opened 2026-06-24
feat(hosted): per-user operation rate limit (#4561 P5)

sanity:
## Problem

Hosted mode (#4381) lets untrusted web visitors drive a shared node via a per-user `userToken`. The per-user storage quota (#4573) bounds disk, but nothing bounds how **fast** a single hosted user can issue contract operations (GET/PUT/UPDATE/SUBSCRIBE) or hosted-export downloads — one visitor could flood the node's executor and network. Local (non-hosted, single-user) requests must **never** be rate-limited (the "never break single-user default" invariant).

## Solution

A per-user token-bucket operation limiter plus a per-user export min-interval, owned as **per-node server state** (`Arc<DashMap<UserId, _>>`, not a process-global), so multi-node simulations never collide and one user's many connections share a bucket.

- **WS hook** — `process_client_request` (`websocket.rs`) rejects an over-rate contract op **before** forwarding to the node (no executor/network cost), using an existing `ErrorKind::OperationError` frame (no new stdlib wire-format variant), serialized exactly like the existing `DelegateRateLimiter` rejection (both Flatbuffers + Native). Gated on `user_context.is_some()`, so Local / non-hosted / tokenless connections are **never** limited.
- **Export hook** — `hosted_export::export_handler` returns **HTTP 429** when a user exceeds the tighter export interval (export enumerates + re-encrypts every secret in scope, so it's far more expensive than one op). Checked right after the token gate and before op-manager resolution, so a flooding client gets the back-off signal regardless of node readiness.
- **Concurrency** — the WS path is concurrent across connection tasks; each bucket guards `(tokens, last_refill)` behind a short `Mutex` so refill + decrement is one atomic decision — **no check-then-act over-admit race**. The `DashMap` `entry` API makes the lazy bucket insert atomic too. Proven by a 64-thread test asserting that with no refill exactly `burst` tokens are admitted across all threads.
- **Rejection variant** — reuses `ErrorKind::OperationError { cause: "operation rate limit exceeded; retry shortly" }`. No stdlib change.
- **Config** — `--per-user-op-rate-limit` (default 10/s, `0` disables), `--per-user-op-burst` (50), `--per-user-export-min-interval-secs` (10), wired through `WebsocketApiArgs` / `WebsocketApiConfig` exactly like the just-merged quota, including the `ConfigArgs::build()` merge **and** the `all_persisted_config_fields_round_trip_through_build` guard test, so the fields don't silently revert on restart.
- **Time** — `tokio::time::Instant` (matching the existing `PerCallsiteRateLimiter` / `TrackedBackoff` production limiters in this area), injectable in tests via `start_paused` / advance for deterministic refill + interval checks.
- **Limiter primitive** — hand-rolled small atomic token bucket. The existing `PerCallsiteRateLimiter` is a fixed-window counter keyed by tracing callsite (can't express burst≠rate, wrong key type) and `TrackedBackoff` is exponential-backoff-on-failure, so neither fit a (capacity=burst, refill=rate) token bucket keyed by `UserId`.

## Testing

- **10 limiter unit tests**: under-rate passes; burst-then-reject; refill over time (partial + capped at burst); disabled never limits (and allocates no buckets); per-user independence; same-user shared bucket; **64-thread concurrency over-admit guard**; export interval enforced / disabled / per-user.
- **5 WS-boundary integration tests** driving `process_client_request`: Local flood never limited (and never touches the limiter map); hosted burst-then-reject (asserts the `OperationError` frame **and** that a rejected op is NOT forwarded); hosted-disabled forwards all; no-limiter-Extension forwards all; two users independent.
- **4 export helper tests**: absent limiter never rejects; second immediate export rejected (429 path); per-user; disabled passes all.
- `cargo fmt` + `cargo clippy -p freenet --lib -- -D warnings` clean; `cargo test --workspace --no-run` compiles; full **websocket (63)**, **client_events (124)**, **config (70)**, **server (317)**, **hosted_export (13)** suites pass.

Refs #4561, #4381

[AI-assisted - Claude]

https://claude.ai/code/session_01VD5thGSuULveJZuw9w9tD7

### DIFF
--- a/apps/freenet-ping/app/tests/common/mod.rs
+++ b/apps/freenet-ping/app/tests/common/mod.rs
@@ -151,11 +151,10 @@ pub async fn base_node_test_config_with_rng<R: Rng>(
             // WebSocket binds to same IP as network for test isolation (prevents port conflicts)
             address: Some(network_bind_ip.into()),
             ws_api_port: Some(ws_api_port),
-            token_ttl_seconds: None,
-            token_cleanup_interval_seconds: None,
-            allowed_host: None,
-            allowed_source_cidrs: None,
-            hosted_mode: None,
+            // All other ws_api args default (None). `..Default::default()` so a
+            // new ws_api field (e.g. the per-user rate-limit fields, #4561) can't
+            // break this test helper's compile.
+            ..Default::default()
         },
         network_api: NetworkArgs {
             // Use varied IP for network socket to get unique ring locations
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -39,6 +39,12 @@ pub(crate) mod result_router;
 pub(crate) mod session_actor;
 #[cfg(test)]
 mod test_correlation;
+/// Per-user operation/export rate limiting for hosted mode (#4561, P5 of #4381).
+/// Not gated on the `websocket` feature: it has no axum dependency (just
+/// `DashMap` + `tokio::time` + `UserId`) and its `DEFAULT_*` constants are the
+/// single source of truth for the operator-config defaults in `config.rs`,
+/// which compiles with or without the feature.
+pub(crate) mod user_op_rate_limit;
 #[cfg(feature = "websocket")]
 pub(crate) mod websocket;
 
--- a/crates/core/src/client_events/user_op_rate_limit.rs
+++ b/crates/core/src/client_events/user_op_rate_limit.rs
@@ -53,7 +53,14 @@ pub const DEFAULT_PER_USER_OP_RATE_LIMIT: u64 = 10;
 /// Default per-user burst capacity (max tokens that can accumulate). A user
 /// who has been idle can issue up to this many ops back-to-back before being
 /// throttled to the sustained [`DEFAULT_PER_USER_OP_RATE_LIMIT`].
-pub const DEFAULT_PER_USER_OP_BURST: u64 = 50;
+///
+/// Set generously (100): a heavy-room hosted user's app-load — room list +
+/// per-room state + member lists + delegate secret reads across many rooms —
+/// can plausibly burst past 50 ops in the first second after page load. A
+/// false-positive throttle there breaks the UX worse than allowing a slightly
+/// larger ONE-TIME burst; the sustained [`DEFAULT_PER_USER_OP_RATE_LIMIT`]
+/// (10/sec) is the real flood cap, the burst is just the initial allowance.
+pub const DEFAULT_PER_USER_OP_BURST: u64 = 100;
 
 /// Minimum seconds between hosted-export downloads per user (export is far more
 /// expensive than a single op, so it gets a separate, tighter limit). `0`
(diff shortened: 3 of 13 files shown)


sanity:
## Full review (3 Claude lenses + ground-read; Codex queued for its reset) — solid core, but scope + a memory-DoS to fix

**Verified correct (don't touch):** the token-bucket math + concurrency (the 64-thread frozen-clock test genuinely proves no over-admit), config round-trip through the guard, the single-user invariant (Local never produces a user_context → never limited), and the export 429 path.

**Must fix:**
- **[BLOCKER] Limit counts only `ContractOp`.** `DelegateOp` runs WASM on the *same serial loop* and is *more* expensive per op (fresh wasmtime instance + fsync + redb txn), and the existing `DelegateRateLimiter` is failure-only, so a flood of *successful* delegate ops is completely unthrottled — defeating the PR's own goal. Hosted users issue delegate ops by design. Also **`NodeQueries`** (`NodeDiagnostics` clones the connection map + hosting keys inline on the loop) is reachable by hosted users (the existing guard keys on `origin_contract`, which is `None` for a hosted shell connection). Fix: count `ContractOp | DelegateOp | NodeQueries` under the per-user bucket (or block NodeQueries for hosted users).
- **[MAJOR] The limiter's own maps are unbounded.** `op_buckets`/`last_export` are insert-only, keyed by **client-chosen unvalidated tokens** → unbounded permanent growth = a memory-DoS, violating the GC-TTL + unbounded-per-key-collection rules. Fix: absolute cap + LRU eviction (reuse the `TrackedBackoff` pattern), evicting only idle/LRU entries so active throttling is never reset.
- **[CI] `freenet_test` macro compile break** — the macro's `WebsocketApiArgs` initializer is missing the 3 new fields, so `playwright_shell` won't compile (CI red). Fix the macro (ideally `..Default::default()`).

Plus the cheap test adds (DelegateOp/NodeQueries counted, the map-bound under a token flood, the burst clamp, export sub-second boundary). Reworking; will re-review.

[AI-assisted - Claude]

sanity:
## Review complete — cleared to merge (CI green on de175af5)

**Independent review (external models unavailable → documented Claude-subagent fallback):**
- **Codex:** tried repeatedly via `codex review` across this PR's life — every invocation hit the usage limit (likely out of credits, not a rolling reset). **Gemini:** unavailable (free tier deprecated, IneligibleTierError). So both external models were genuinely down.
- **Fallback per the multi-model-review policy** (≥3 diverse Claude lenses, scaled up for this Full-tier abuse-control surface): **4 distinct adversarial lenses** — skeptical bug-hunt, code-first (intent-vs-impl), testing-coverage, and a focused bounded-eviction lens — plus my own adversarial reads. Across 4 rounds they found and drove fixes for **7 real issues**: DelegateOp/NodeQueries flood bypass (only ContractOp was counted), unbounded limiter maps (memory-DoS via client-chosen tokens), the O(n)-per-insert eviction amplification, the op_burst==1 eviction-safety inversion, and the test/macro compile breaks. Token-bucket math, concurrency (64-thread no-over-admit proof), config round-trip, and the single-user invariant were all verified clean.
- Final production logic is unchanged since the fully-reviewed commit; the only post-review production change is a trivial default tuning (op burst 50→100 for heavy-room app-load headroom; sustained 10/s rate unchanged). Test-only changes otherwise.

CI: all required checks green on the current HEAD. [AI-assisted - Claude]

PR #4576, opened 2026-06-24
DIAG: #4534 with off-load DISABLED (bisect NAT, do not merge)

iduartgomez:
Diagnostic. Same as #4572 but summarize off-load forced inline (env-gated default off). If NAT passes here → the off-load EXECUTION is the NAT trigger. If NAT still fails → the shared-semaphore/fair_queue refactor is. Close after CI.


iduartgomez:
Diagnostic PR — done. Confirmed off-load DISABLED is green on NAT (isolated the cause to the off-load execution). Closing.

Issue #4401, opened 2026-06-13
Graceful shutdown doesn't release redb lock or stop detached executor/WS server


iduartgomez:
## Problem

A node's graceful shutdown does **not** tear down the contract executor or the WebSocket
server, and therefore does **not** release the redb on-disk file lock. Both run detached on
the process-wide `GlobalExecutor` holding `op_manager` clones, and outlive the node whose
network event loop was shut down.

Discovered while building the persistence round-trip test (#3367 Gap 1, PR #4398): an
in-process "stop node 1, start node 2 against the same data dir" restart could not work —
node 2's redb open blocks on the lock node 1 still holds, even after
`ShutdownHandle::shutdown()` and 30 s+ of waiting. PR #4398 worked around it by running the
two nodes as separate OS processes (process exit releases the lock), but the underlying
behavior is a latent issue in its own right.

## Impact

- **In-process restart is impossible** against the same data dir — any test or embedding that
  stops and restarts a node in one process deadlocks on the redb lock.
- **Detached tasks outlive their node.** The executor + WS server keep running (and keep
  answering on their bound ports) after the node they belong to has "shut down." In the test
  this manifested as node 1's stale WS server silently answering a GET meant for node 2 — a
  false-positive failure mode (see PR #4398 for the full trace).
- Graceful shutdown advertises teardown it does not actually perform.

## Expected

`ShutdownHandle::shutdown()` (or node drop) should tear down the contract executor and WS
server it spawned, releasing the redb lock and stopping the detached tasks, so that:

1. The redb file lock is released on shutdown.
2. No detached task keeps serving on the node's ports after shutdown.
3. An in-process restart against the same data dir succeeds.

## Notes

- Root cause: executor + WS server are spawned on the process-wide `GlobalExecutor` and aren't
  joined/aborted by the shutdown path.
- redb durability itself is fine — commits are synchronous per-write (`txn.commit()`), so data
  is durable at ack time regardless. This is about *teardown / lock release*, not data loss.
- Cross-reference: PR #4398 contains the reproduction and the two-process workaround.

[AI-assisted — found during #3367 Gap 1 test development]

PR #4577, opened 2026-06-24
feat(hosted): inactive-user TTL reclaim (#4561 P5)

sanity:
## Problem

A hosted Freenet node is meant to be a transient "try Freenet before you install" demo: each web visitor gets a per-user delegate-secret namespace (#4381 P2), plays with Freenet, then either exports their data to their own peer (P3) or walks away. Nothing reclaimed a walked-away visitor's data, so storage grew without bound on a public proxy node.

## Solution

A background sweep — spawned **only in hosted mode** — reclaims a whole abandoned user's data after N days (default **30**) of real-calendar inactivity. The cardinal rule is to **never reclaim an ACTIVE user** (that's silent data loss), so activity tracking is conservative and durable.

- **`reclaim_user`** (`secrets_store.rs`): idempotent, crash-safe removal of a user's ENTIRE durable footprint — the on-disk `<delegate>/users/<id>/` subtree for every delegate, the activity-marker tree, the ReDb user-index rows (keyed on the real `code_hash` from the table), and the process-global quota-tracker entry. Source-of-truth-first ordering (disk → derived caches): a crash mid-reclaim leaves at worst a dangling ReDb row the read path tolerates (`MissingSecret`, not a panic) and the next sweep cleans up. Pooled executors' in-memory indexes self-heal from ReDb on restart.

- **Durable last-activity**: a per-user `<base>/users/<id>/.last_seen` marker storing a UNIX-epoch second. **Wall-clock is load-bearing here** (a durable cross-restart calendar TTL — `TimeSource` has no stable epoch and resets in tests, `Instant` resets on restart); the testable core takes `now`/`ttl` as params so tests are deterministic. Stamped on WS **connect** and **each client request** in `process_client_request` (gated on `user_context.is_some()`), debounced to 1h so the hot path is a single `stat`. The marker lives outside any `<delegate>/users/<id>/` tree, so it never counts toward the per-user quota.

  > **Known limitation (also in a code comment):** a user holding ONE connection open for the full 30 days with ZERO requests wouldn't be re-stamped. Negligible at a 30-day TTL — browsers don't hold idle WS for weeks, and any reconnect/request re-stamps. Not over-engineering a keepalive for v1.

- **Sweep task**: spawned in `RuntimePool::new` (which owns the shared ReDb `Storage` and the `Config`) only when `hosted_mode && ttl > 0`. Loops on a configurable interval (default 1h); each pass **re-reads `.last_seen` immediately before deleting** to close the reconnect-during-sweep race. Pure age threshold, **no unbounded "active" exemption** (GC rule). Abort guard owned by the pool so shutdown tears it down (#4401 discipline).

- **Config**: `per_user_inactive_ttl_secs` (default `2_592_000` = 30 days, `0` disables) and `inactive_user_sweep_interval_secs` (default `3_600`), mirroring `per_user_secret_quota_bytes` exactly including the `all_persisted_config_fields_round_trip_through_build` guard.

**Local single-user data is NEVER touched**: the sweep isn't spawned outside hosted mode, a Local node has no `users/` markers to enumerate, and Local secrets live outside the `users/<id>/` tree.

## Testing

12 new unit tests: reclaim clears all four state locations; active user is never reclaimed; user without a marker is spared; idempotent re-run; Local data never reclaimed; `ttl=0` disables; `.last_seen` doesn't count toward quota; dangling blob doesn't panic; stamp debounce; reconnect-during-sweep spares the user; spawn respects `ttl=0`. `now`/`ttl` injected for determinism (no real sleeps/`SystemTime` in the testable core).

Verified: `cargo fmt --check`, `clippy -p freenet --lib -D warnings`, both no-default-features test combos (`freenet` + `freenet-ping-app`), and a 99-test default-features run for the touched modules — all green.

## Stacking note

Originally built stacked on `feat/hosted-op-ratelimit` (both hook `process_client_request`). That branch has since merged to main (#4575), so this PR is now rebased directly onto `main` and targets `main`.

Refs #4561 #4381

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -1089,8 +1089,13 @@ async fn websocket_commands(
     // that carries a `UserSecretContext`, which only the hosted serve path
     // (which DOES install the limiter) ever derives.
     op_rate_limiter: Option<Extension<UserOpRateLimiter>>,
+    // The node's secrets dir for stamping per-user last-activity markers
+    // (#4561). OPTIONAL: only `serve_client_api_in_impl` installs it; the
+    // standalone test paths install nothing, so absence ⇒ no activity stamping.
+    activity_secrets_dir: Option<Extension<crate::server::ActivitySecretsDir>>,
 ) -> Response {
     let op_rate_limiter = op_rate_limiter.map(|Extension(l)| l);
+    let activity_secrets_dir = activity_secrets_dir.map(|Extension(d)| d);
     let on_upgrade = move |ws: WebSocket| async move {
         // Get the data we need from the DashMap
         // Track whether a token was provided but is invalid (stale/expired)
@@ -1149,6 +1154,7 @@ async fn websocket_commands(
             auth_and_instance,
             user_context,
             op_rate_limiter,
+            activity_secrets_dir,
             token_is_invalid,
             encoding_protoc,
             api_version,
@@ -1212,6 +1218,10 @@ async fn websocket_interface(
     // user's connections; consulted in `process_client_request` only when this
     // connection carries a `user_context`.
     op_rate_limiter: Option<UserOpRateLimiter>,
+    // The node's secrets dir for per-user last-activity stamping (#4561). `None`
+    // on standalone/test paths; an empty path also disables stamping. Used only
+    // when this connection carries a `user_context` (hosted mode).
+    activity_secrets_dir: Option<crate::server::ActivitySecretsDir>,
     token_is_invalid: bool,
     encoding_protoc: EncodingProtocol,
     api_version: ApiVersion,
@@ -1220,6 +1230,16 @@ async fn websocket_interface(
     let (mut response_rx, client_id) =
         new_client_connection(&request_sender, auth_token.clone()).await?;
     let (mut server_sink, mut client_stream) = ws.split();
+
+    // Stamp last-activity on CONNECT (#4561, inactive-user TTL): a fresh
+    // connection from a hosted user is activity, so refresh their `.last_seen`
+    // marker right away (debounced — a no-op if a recent stamp exists). Gated on
+    // `user_context.is_some()` so Local/non-hosted connections never write a
+    // marker. Best-effort and off the request hot path (runs once per
+    // connection). See `stamp_activity` for the wall-clock + debounce details.
+    if let (Some(ctx), Some(dir)) = (user_context.as_ref(), activity_secrets_dir.as_ref()) {
+        stamp_activity(dir, ctx);
+    }
     let contract_updates: Arc<Mutex<VecDeque<(_, mpsc::Receiver<HostResult>)>>> =
         Arc::new(Mutex::new(VecDeque::new()));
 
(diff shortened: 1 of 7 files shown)


sanity:
## Full review (3 Claude lenses + ground-read; Codex+gemini both down → documented fallback) — one MAJOR, rest verified safe

For a destructive op the bar is "prove it can't delete what it shouldn't," and on that the build holds up well:
- **Can't touch Local or another user's data** — every user-derived path component is bs58(32-byte UserId), which can't contain `/` or `..` (no traversal); Local secrets live at `<base>/<delegate>/` *outside* the `users/` tree reclaim ever touches; non-hosted nodes are triple-guarded (sweep not spawned, empty candidate set, scope separation).
- **Reclaim is complete, crash-safe, idempotent** — clears fs subtree + ReDb rows (key-match on the real code_hash verified) + quota-tracker entry; marker removed LAST so a crash leaves a re-evaluable marker; a dangling ReDb row returns `MissingSecret` not a panic; the pooled-executor cache residual is benign (graceful MissingSecret, never wrong data).
- Reconnect-race (re-read immediately before delete), debounce staleness (1h vs 30d), stamp-before-throttle, and **backward** clock jumps (saturating_sub) are all safe. Config/wiring/Extension all correct.

**[MAJOR] Forward clock jump → silent mass-delete.** The age check trusts `now` unconditionally, so a host clock jumping forward past the TTL (NTP correcting a bad RTC, date misconfig, VM snapshot restore) makes every user look 30-days-idle at the next sweep → deletes all hosted users in one pass. Fix requested: a per-sweep reclaim **cap + loud warn** (bounds blast radius + alerts the operator) plus anomalous-time-gap detection so a clock fault is caught/bounded, never a silent mass-delete. Plus a perf hoist (reclaim loads the whole ReDb index once per reclaimed user → once per sweep) and two test adds (TTL-boundary, hosted-mode spawn gate).

Reworking; will re-review. [AI-assisted - Claude]

sanity:
## Review complete — cleared to merge (CI green on f1e83311)

**Independent review (external models unavailable → documented Claude-subagent fallback):** Codex was out (usage/credits), gemini-cli deprecated (IneligibleTierError) — both genuinely down. Per policy, scaled-up Claude fallback for this **destructive** (deletes user data) Full-tier surface: **4 distinct lenses** (skeptical / code-first / testing on the core, + a focused destructive-correctness lens on the round-2 delta) plus ground-reads.

**Verified safe:** cannot delete Local or another user's data (bs58 user-id leaf can't path-traverse; Local secrets live outside the  tree; triple-guarded no-op on non-hosted); reclaim is complete, crash-safe, idempotent (ReDb key-match on the real code_hash; dangling blob → MissingSecret not panic); reconnect-race, debounce, and backward-clock all safe; the perf-hoist grouping routes rows to the correct user and never the survivor's (pinned by a new 2-user ReDb test); config/wiring/Extension correct.

**The one MAJOR found and fixed:** a forward clock jump > TTL could have mass-deleted all hosted users in one sweep. Now guarded by a gap detector (skip the pass + warn if  jumped > 8× the interval since the last *trusted* sweep; baseline held on skip/panic so a sustained skew keeps tripping) and a per-pass reclaim cap (256 + loud warn) — a clock fault is now bounded and alerted, never a silent mass-delete. Plus the perf hoist (O(R×T)→O(T)) and boundary/gate/cap/survivor tests.

All required CI green on the current HEAD. [AI-assisted - Claude]

PR #4578, opened 2026-06-25
refactor(hosted): zeroize the per-user token in memory

sanity:
## Problem

The per-user hosted token (the `userToken` presented on the WS upgrade, P2 of #4381) is the **master seed** from which the per-user DEK derives, yet it was the **least** protected secret in memory: it was held as a plain `String` in the `connection_info` websocket middleware, so its raw bytes lingered in freed heap after the connection closed. The derived DEK (`Zeroizing<[u8; 32]>`) and the plaintext secrets (`Zeroizing<Vec<u8>>`) were already zeroized on drop — this closes the gap on the seed itself.

## Solution

Wrap the token in `zeroize::Zeroizing` end-to-end so the raw bytes are wiped on drop. Purely an in-memory container change — no crypto, DEK derivation, wire format, or behavior change.

- `ConnectionInfo.user_token: Option<String>` -> `Option<Zeroizing<String>>`. Enables the zeroize `serde` feature so `serde_urlencoded` deserializes the inner `String` and **moves** it into the wrapper, leaving no separate un-zeroized owned copy.
- The header-sourced token (`x-freenet-user-token`) is wrapped directly in `Zeroizing::new(...)` at extraction, so the only owned copy of the raw token is the zeroizing one. The `.to_str()` borrow into the axum header buffer is a separate concern (out of scope here).
- `derive_user_context` keeps its `Option<&str>` signature; the token is **borrowed** (not copied) as a `&str` view, so `from_token` receives byte-identical input.

No lingering plain `String`/`Vec<u8>` copy of the token survives un-zeroized on either the query path or the header path.

## Testing

- `cargo fmt --check` clean; `cargo clippy -p freenet --lib -- -D warnings` clean.
- No-default-features test-compile combos that bit earlier PRs both compile:
  - `cargo test --no-run -p freenet --features trace,websocket,redb,wasmtime-backend,testing --no-default-features`
  - `cargo test --no-run -p freenet-ping-app --features testing`
- `cargo test -p freenet --test hosted_mode_isolation` — 3/3 pass (token still authenticates and derives the correct per-user namespace end-to-end).
- `cargo test -p freenet --lib client_events::websocket` — 67/67 pass (the `derive_user_context` / `decide_user_token` gate tests).
- `cargo test -p freenet --lib hosted_export` — 13/13 pass.

A "zeroed on drop" property is hard to unit-test directly; verification is (a) it compiles, (b) the token still authenticates/derives correctly (existing tests green), (c) the field type is now `Zeroizing`.

Refs #4381 (hosted-proxy umbrella).

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8989,6 +8989,7 @@ version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e13c156562582aa81c60cb29407084cdb54c4164760106ab78e6c5b0858cf64e"
 dependencies = [
+ "serde",
  "zeroize_derive",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ renegade-ml = "0.3.3"
 pin-project = "1"
 thiserror = "2"
 ulid = { version = "1.1", features = ["serde"] }
-zeroize = { version = "1.8", features = ["derive"] }
+zeroize = { version = "1.8", features = ["derive", "serde"] }
 
 # Async/networking
 axum = { version = "0.8", default-features = false }
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -118,6 +118,7 @@ use futures::{FutureExt, SinkExt, StreamExt, future::BoxFuture, stream::SplitSin
 use headers::Header;
 use serde::Deserialize;
 use tokio::sync::{Mutex, mpsc};
+use zeroize::Zeroizing;
 
 use crate::{
     client_events::AuthToken,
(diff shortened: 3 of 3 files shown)


sanity:
## Review complete — cleared to merge

Light-tier security-hygiene change (wrap the per-user token in `Zeroizing` so it's wiped on drop; no behavior, crypto, wire-format, or derivation change). Reviewed:
- **Adversarial ground-read (the crux: no lingering plain copy):** the token is wrapped in `Zeroizing` at BOTH entry points — header via `Zeroizing::new(s.to_owned())`, query via the serde-into-`Zeroizing` move — everything downstream is a `&str` borrow, and the raw token's whole server-side lifetime ends in `connection_info` (consumed to derive the already-`Zeroizing` DEK context). No un-zeroized owned copy survives.
- **External pass (Codex, now available):** "consistently wraps the parsed/header user token in Zeroizing and preserves the existing token selection and derivation behavior; no regression or build-breaking issue."
- Tests: hosted_mode_isolation 3/3, websocket gate 67/67, hosted_export 13/13, no-default-features compiles for both freenet + freenet-ping-app.

[AI-assisted - Claude]

PR #4579, opened 2026-06-25
fix(hosted): dark-mode + dismiss account popover on outside click

sanity:
## Problem
The hosted shell bar (node-rendered chrome above the sandboxed app iframe) (1) hardcoded light colors so it clashed on a dark-mode OS, and (2) the Account popover only dismissed by clicking Account again — the existing document-click handler never fired because the click landed inside the sandboxed iframe (separate document, no bubble to the shell).

## Solution
1. `@media (prefers-color-scheme: dark)` block — restrained near-monochrome (charcoal bar, light-grey text, hairline borders, dark buttons/popover), matching the light theme's character.
2. Dismiss the popover on `window` blur too — clicking into the iframe blurs the shell window, covering the case the document handler can't see.

## Testing
Rendered light + forced-dark via headless Chrome to confirm the palette (screenshot shared with the maintainer). Shell Playwright E2E exercises the shell render. No Rust logic changed (CSS + a one-line JS handler in include_str! assets).

Refs #4381

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers/assets/hosted_bar.css
+++ b/crates/core/src/server/path_handlers/assets/hosted_bar.css
@@ -98,3 +98,43 @@ iframe {
   color: #2a8a55;
   font-size: 12px;
 }
+
+/* Respect the OS dark-mode preference. Same restrained, near-monochrome
+   character as the light theme: charcoal bar, light-grey text, hairline
+   borders. */
+@media (prefers-color-scheme: dark) {
+  #fnbar {
+    background: #1c1c1e;
+    border-bottom-color: #303032;
+    color: #98989c;
+  }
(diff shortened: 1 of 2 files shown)


sanity:
Adding `test-exempt`: this is a frontend-only change — a CSS `@media (prefers-color-scheme: dark)` block plus a one-line `window.blur` JS handler, both in `include_str!` assets. There's no Rust logic to unit-test, and the CI test rule counts only Rust `#[test]`/shell `check` assertions, neither of which applies to a CSS/JS UX change. The shell render itself is exercised by the Shell Playwright E2E job. Codex reviewed the diff clean; the dark palette was screenshot-verified. [AI-assisted - Claude]

PR #4582, opened 2026-06-25
fix: surface auto-update silent failures for unsupervised peers (#4580)

### LINKED ISSUE #4580: Auto-update is supervised-install-only: bare-CLI / non-systemd peers detect updates but never apply them
## Problem

Auto-update only works for peers installed under a supervisor that honors exit code 42. A bare `freenet network` CLI run, a hand-run `cargo install`, or a dirty/dev build detects an available update, exits with code 42, and then dies **without applying it**, because nothing restarts it. This structurally guarantees a long tail of peers that never update, no matter how many releases ship.

### Mechanism (verified)

A peer polls the GitHub releases API every 60s (`crates/core/src/bin/commands/auto_update.rs:41`) and, on a newer version, calls `std::process::exit(42)` (`EXIT_CODE_UPDATE_NEEDED`, `auto_update.rs:29`; exit at `crates/core/src/bin/freenet.rs:871`). The download + atomic self-replace lives in `freenet update` (`crates/core/src/bin/commands/update.rs::replace_binary`). The running process does **not** restart itself on Linux/macOS; an external supervisor must catch exit 42 and run `freenet update`:

- Linux service install: systemd `ExecStopPost` (`service/linux.rs:317`) + `Restart=always`.
- macOS: launchd wrapper (`service/macos.rs`). Windows: in-binary run-wrapper loop (`service/wrapper.rs`).

| Install type | Auto-updates? |
|---|---|
| systemd service / desktop app | Yes |
| bare `freenet network` CLI | No (exits 42, dies, never restarted) |
| `cargo install` / dirty build | No (dirty disables auto-update; bare run has no supervisor) |

### Additional silent-failure points

- Dirty/dev builds disable auto-update entirely (`freenet.rs:304`, gated on `GIT_DIRTY`).
- After 3 failed update attempts a peer locks out auto-update (persisted) — a non-writable binary path becomes a permanent silent lockout (`auto_update.rs:38,333`).
- The GitHub check is unauthenticated (60 req/hr/IP); co-located peers share that limit.
- No end-user documentation states that auto-update requires a service install.

## Impact

A non-trivial fraction of the network can never auto-update. Telemetry (last 48h, self-reporting peers) shows ~64% on 0.2.7x with a floor at 0.2.71, consistent with a slow-updating tail rather than a broken updater, but the tail is real and self-perpetuating. This was surfaced while investigating why only a minority of peers appeared to be on recent builds (the headline figure turned out to be a measurement artifact, but the structural tail is genuine).

## Proposed direction (for discussion)

Lowest-risk first:

1. **Loud, operator-visible warning** when a peer detects an update it cannot apply (no supervisor present, dirty build, or update lockout), so the situation is diagnosable instead of silent.
2. **Document** the supervised-install requirement (and the dirty-build behavior) where users run a node.
3. **Optional self-supervising run mode** for the bare CLI (an in-process wrapper loop, as Windows already does), so a hand-run peer can auto-update too. This is a behavior change and should be opt-in or clearly flagged, since silently swapping and restarting a user's binary has its own tradeoffs.

[AI-assisted - Claude]

sanity:
## Problem

Auto-update is structurally **supervised-install-only**. The running node never replaces its own binary: it detects a newer release, exits with code **42**, and relies on a supervisor to run `freenet update` and restart it:

- systemd `ExecStopPost` (Linux service)
- the launchd wrapper script (macOS)
- the in-process `service run-wrapper` loop (Windows / Linux tray)

A bare `freenet network` run, a **dirty/dev build**, or a peer that hit the **3-failure update lockout** detects the update and then **silently** fails to apply it. There is no operator-visible signal, so this produces a long, self-perpetuating tail of peers that never update (issue #4580).

## Solution

Make the silent failures **diagnosable** without changing update *behaviour* (the self-supervising run mode proposed in the issue is a separate, opt-in feature, left as a follow-up).

- **Supervisor-aware warning at the exit-42 site**, with three states so we never over-claim. Each supervisor now marks the `freenet network` child with a `FREENET_SUPERVISED` env var (systemd unit `Environment=`, macOS wrapper `export`, in-process wrapper `cmd.env`):
  - marker present → `Supervised`: `info!`, exit quietly (the normal service case).
  - systemd only (`INVOCATION_ID`, no marker) → `SupervisedUnverified`: `warn!` softly — a custom/`systemd-run`/pre-marker unit may not actually have the exit-42 hook, so we can't confirm, but we also cover older Freenet units.
  - neither → `Unsupervised`: `error!` + stderr that the update was detected but will **NOT** be applied, and tells the operator to run `freenet update` or install the service.

  Detection is honest by construction: it only goes fully quiet on **our own** marker; systemd's generic `INVOCATION_ID` is not treated as proof of the exit-42 hook.
- **Dirty-build** "auto-update disabled" path bumped `info!` → `warn!` with operator guidance.
- **3-failure lockout** skip bumped `debug!` → `warn!`, logged loudly **once per process** then dropped to `debug!` so the permanent state doesn't spam the 60s update loop.
- `freenet network --help` now documents the supervised-install requirement and the dirty-build behaviour.

No change to when/whether a node exits 42 or applies an update.

## Testing

- Pure `detect_supervisor_status` unit tests: marker → Supervised; `INVOCATION_ID` alone → SupervisedUnverified; marker beats `INVOCATION_ID`; neither → Unsupervised; empty/whitespace values are not evidence.
- Comment-stripped source-scrape pins: the exit-42 path escalates on the unsupervised case (`error!`), the dirty-build and lockout paths log loudly (`warn!`), and all three supervisors set the `FREENET_SUPERVISED` marker (caught on every CI platform since they scrape source text, not cfg-gated generators).
- `cargo fmt` + `cargo clippy --bin freenet -- -D warnings` clean; full `cargo test --bin freenet` green (236 tests).

## Review

External review by Codex (`codex exec`, read-only). First pass raised 1 MEDIUM (`INVOCATION_ID` false-positive) + 2 LOW (lockout log spam, brittle source-scrape tests); all three addressed (three-state detection, once-per-process warn, comment-stripping). Re-run on the fixed diff: "No new findings", all prior findings confirmed resolved.

Closes #4580

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -17,6 +17,7 @@ use anyhow::Result;
 use semver::Version;
 use std::fs;
 use std::path::PathBuf;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::{Duration, SystemTime};
 
 pub use freenet::transport::{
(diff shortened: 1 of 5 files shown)


sanity:
## External review (Codex) — addressed

Independent non-Claude review via `codex exec` (read-only). Gemini CLI is unavailable (deprecated free tier), so Codex was the external model.

**First pass — 3 findings:**

- **[MEDIUM]** `INVOCATION_ID` is a generic systemd signal, not proof of Freenet's exit-42 → `freenet update` hook. A custom unit / `systemd-run` / hand-written service running `freenet network` without that hook would be classified `Supervised` and the unsupervised warning suppressed even though the update never applies.
  - **Fixed:** `SupervisorStatus` is now three states. Only the authoritative `FREENET_SUPERVISED` marker → `Supervised` (quiet). Bare `INVOCATION_ID` → `SupervisedUnverified` (still `warn!`, softer). Neither → `Unsupervised` (`error!`). Detection only goes fully quiet on our own marker.
- **[LOW]** Lockout `warn!` fires every 60s loop tick → log spam.
  - **Fixed:** process-local `AtomicBool` — `warn!` loudly once, then `debug!` thereafter.
- **[LOW]** Source-scrape pin tests could pass on commented-out code.
  - **Fixed:** added `strip_line_comments` and applied it to the exit-42, dirty-build, and lockout pins so only real code satisfies them.

Codex confirmed the supervisor-marker wiring is sound across all three generators (systemd `Environment=`, macOS `export`, in-process `Command::env`), with no shell-quoting/`format!` issue and no cfg-gated compile risk it could spot.

**Re-run on the fixed diff:** "No new findings" — all three prior findings confirmed resolved (three-state match handles all variants with no catch-all; `Ordering::Relaxed` is correct since the atomic only coordinates which caller logs the single warning; comment-stripping addresses the flagged brittleness).

[AI-assisted - Claude]

PR #4583, opened 2026-06-26
fix: stop placement-migration subscription roots from storming on renewal (#4440)

### LINKED ISSUE #4440: v0.2.73 placement-migration incident: re-enable migration + deferred follow-ups

sanity:
## Problem

When the placement migration drifts a contract onto the peer closest to its ring location, that peer becomes the **body-holding subscription root**: it hosts the contract AND no connected neighbor is closer to the key. Such a peer has no peer closer than itself to subscribe to upstream.

The subscription-renewal driver did not recognise that role. Every renewal cycle it still sent a wire `Subscribe::Request`, which routes greedily toward the contract, dead-ends at the root itself, fails, and is retried on the next 30s cycle — forever. Across many such contracts that loop is the renewal storm that made the 500-node nightly go metastable (`subscribe: attempt timed out ... is_renewal: true` floods; in production the renewal task also overruns its 25s cycle deadline). This is the storm symptom of the #4404 placement work, tracked in #4440.

## Solution (placement "proposal 1")

In the renewal driver (`run_renewal_subscribe`), **before** driving a wire renewal, the node checks whether it is the body-holding subscription root for the contract via `Ring::body_holding_subscription_root_key` (hosts the body AND is the closest connected peer — `is_subscription_root`). If it is, it takes a dedicated **root-satisfied path**:

- record the new `renewal_terminus_satisfied` counter,
- **conditionally** refresh the LOCAL `active_subscriptions` lease (see below),
- send **no** wire `Subscribe::Request`, deliver **no** client result,
- return `RenewalOutcome::Success` (the renewal loop's `SubscriptionRecoveryGuard::complete(true)` then clears the pending mark and resets the subscription backoff with success semantics).

### Lease handling is gated on `contract_in_use` (not unconditional)

The local lease refresh is **conditional on genuine, time-bounded interest** (`contract_in_use`: a local client subscription or a downstream subscriber):

- **In-use root** → refresh the local lease (a local-only map update, no wire traffic). This moves the root onto the normal ~6-minute lease-expiry renewal cadence instead of being re-selected by `contracts_needing_renewal()` every 30s cycle (which would consume a renewal-batch slot every cycle with a no-op local success and could starve real non-root wire renewals). Because the interest signals are time-bounded, the lease cannot self-perpetuate past the interest.
- **No-interest root** → **do NOT refresh**; let the lease lapse. `contracts_needing_renewal` section 1 re-selects any live active subscription with no interest re-check, so an unconditional refresh would be an unbounded `is_subscribed`-only retention exemption (forbidden by the `contract_in_use` rustdoc + AGENTS.md time-bounded-exemption rule). A no-interest root's lease lapses and it drops out of the renewal set entirely. Guarded by `test_no_interest_subscription_root_lease_lapses`.

### Routability matches `k_closest_potentially_hosting` eligibility (transient-only exclusion)

The root predicate only counts a closer neighbor as "I'm not the root" if a renewal could actually route to it, mirroring `k_closest`'s asymmetry:

- **Transient peers** are excluded unconditionally by `k_closest` → excluded in the root check too (else a node whose only closer neighbor is transient would skip the short-circuit and storm).
- **Not-ready peers** are `k_closest`'s fallback when no ready candidate exists → treated as **routable** in the root check (else a node would be wrongly classified as root in cold-start / low-degree topologies and suppress a renewal that would in fact route to that closer peer). Pinned by `is_subscription_root_routability_matches_k_closest_eligibility`.

This is a **dedicated** path, deliberately NOT `complete_local_subscription` (different semantics — emits `LocalSubscribeComplete` and registers client interest). The new counter is surfaced on `router_snapshot` and mirrored into the hand-written OTLP body (guarded by the existing pin test).

**Scope is strictly proposal 1.** The other placement proposals (bodyless forwarding, displacement/make-room, retention grant, `is_upstream` cleanup) are intentionally NOT implemented here.

## Testing

- **Reproduction-first simulation test** `test_subscription_root_renewal_does_not_storm` (3 seeds, the storm is metastable): a body-holding root with a client subscription keeps firing renewals. Asserts the root takes the root-satisfied path (`terminus_satisfied > 0`, bounded `<= 4` to catch an accidental every-cycle refresh) and sends **no** wire renewals (`wire_attempts == 0`), plus a **non-root host control** (a second contract on a node that is not its closest peer) that MUST keep wire-renewing (`wire_attempts > 0`, `terminus_satisfied == 0`) — a guard against a silent network-wide renewal stoppage. Confirmed it FAILS before the fix (root storms: `wire_attempts=3, terminus_satisfied=0`) and PASSES after (`wire_attempts=0, terminus_satisfied=1`).
- **`test_no_interest_subscription_root_lease_lapses`**: a no-interest body-holding root's lease lapses (final `active_subscription_keys` no longer contains the contract) — no unbounded self-perpetuation.
- **Source-pin unit tests**: `run_renewal_subscribe_short_circuits_body_holding_root` (check runs before wire send, returns Success, gated lease refresh, no `complete_local_subscription` reuse) and `is_subscription_root_routability_matches_k_closest_eligibility` (mapping stays in sync with k_closest).
- **Direct predicate unit tests** (`no_closer_routable_neighbor`): closer-routable → not root; closer-but-non-routable (transient) → still root; farther → root; no neighbors → root; locationless → doesn't unseat.
- Counter unit test + OTLP pin test updated for `renewal_terminus_satisfied`.
- Existing placement-migration sim tests still pass — no regression.
- The simulation-only renewal-metrics signal is captured inside `run_controlled_simulation` before `SimNetwork::Drop` clears the registry, read via `ControlledSimulationResult::{renewal_metrics_for, aggregate_renewal_metrics}`. (`is_renewal` is not on the captured `NetLogMessage` stream; the registry `record_*` calls are no-ops in production.)
- `cargo fmt` clean; `cargo clippy` clean on all changed files.

## Rebase note

Rebased onto main after #4584 (interest-tier shadow gauges) merged. The conflict in `router_snapshot` gauges + the OTLP hand-mirror block + the OTLP pin test (router.rs / telemetry.rs / ring.rs) was resolved **additively** — both #4584's interest-tier shadow gauges and this PR's `renewal_terminus_satisfied` gauge are kept. Both OTLP pin tests pass.

## Review status

External **Codex** review run against `main`: surfaced and addressed two P2s across review rounds (lease-lapse batch starvation; not-ready-peer routability). Final Codex run: no findings. Full multi-model review came back clean. This touches the subscribe/renewal state machine (a high-risk surface).

Closes #4440 (storm portion); refs #4404

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -143,9 +143,9 @@ pub mod dev_tool {
         ContractSubscription, ProximityViolation, RenewalMetrics, TopologySnapshot,
         TopologyValidationResult, aggregate_renewal_metrics, clear_all_topology_snapshots,
         clear_current_network_name, clear_renewal_metrics, clear_topology_snapshots,
-        get_all_topology_snapshots, get_current_network_name, get_renewal_metrics,
-        get_topology_snapshot, register_topology_snapshot, set_current_network_name,
-        validate_topology, validate_topology_from_snapshots,
+        get_all_renewal_metrics, get_all_topology_snapshots, get_current_network_name,
+        get_renewal_metrics, get_topology_snapshot, register_topology_snapshot,
+        set_current_network_name, validate_topology, validate_topology_from_snapshots,
     };
     pub use wasm_runtime::secret_export::{
         BundleKeyMaterial, ExportError, ImportReport, TargetScope, export_bundle, import_bundle,
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -140,11 +140,12 @@ pub mod dev_tool {
 
     // Re-export topology registry for subscription validation in tests
     pub use ring::topology_registry::{
-        ContractSubscription, ProximityViolation, TopologySnapshot, TopologyValidationResult,
-        clear_all_topology_snapshots, clear_current_network_name, clear_topology_snapshots,
-        get_all_topology_snapshots, get_current_network_name, get_topology_snapshot,
-        register_topology_snapshot, set_current_network_name, validate_topology,
-        validate_topology_from_snapshots,
+        ContractSubscription, ProximityViolation, RenewalMetrics, TopologySnapshot,
+        TopologyValidationResult, aggregate_renewal_metrics, clear_all_topology_snapshots,
+        clear_current_network_name, clear_renewal_metrics, clear_topology_snapshots,
+        get_all_topology_snapshots, get_current_network_name, get_renewal_metrics,
+        get_topology_snapshot, register_topology_snapshot, set_current_network_name,
+        validate_topology, validate_topology_from_snapshots,
     };
     pub use wasm_runtime::secret_export::{
         BundleKeyMaterial, ExportError, ImportReport, TargetScope, export_bundle, import_bundle,
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -375,6 +375,13 @@ pub struct ControlledSimulationResult {
     #[cfg(any(test, feature = "testing"))]
     #[allow(dead_code)]
     pub(crate) node_rings: HashMap<NodeLabel, Arc<crate::ring::Ring>>,
+    /// Per-peer subscription-renewal metrics captured at the end of the
+    /// simulation, keyed by socket address. Captured here (before the
+    /// `SimNetwork` is dropped, which clears the registry) so renewal-storm
+    /// tests can read `wire_attempts` / `terminus_satisfied` after the run
+    /// returns. See `crate::ring::topology_registry::RenewalMetrics` (#4440).
+    pub renewal_metrics:
+        HashMap<std::net::SocketAddr, crate::ring::topology_registry::RenewalMetrics>,
 }
 
 #[cfg(any(test, feature = "testing"))]
@@ -391,6 +398,28 @@ impl ControlledSimulationResult {
         self.node_rings
             .get(label)
             .is_some_and(|ring| ring.is_hosting_contract(key))
+    }
+
+    /// Subscription-renewal metrics for the peer at `addr` (captured before the
+    /// `SimNetwork` was dropped). Returns `RenewalMetrics::default()` (all zero)
+    /// if the peer recorded no renewal activity. See #4440.
+    pub fn renewal_metrics_for(
+        &self,
+        addr: &std::net::SocketAddr,
+    ) -> crate::ring::topology_registry::RenewalMetrics {
+        self.renewal_metrics.get(addr).copied().unwrap_or_default()
+    }
+
+    /// Sum of the renewal metrics across every peer in the simulation. See #4440.
+    pub fn aggregate_renewal_metrics(&self) -> crate::ring::topology_registry::RenewalMetrics {
+        self.renewal_metrics.values().fold(
+            crate::ring::topology_registry::RenewalMetrics::default(),
+            |mut acc, m| {
+                acc.wire_attempts += m.wire_attempts;
+                acc.terminus_satisfied += m.terminus_satisfied;
+                acc
+            },
+        )
     }
 }
 
@@ -4236,6 +4265,12 @@ impl SimNetwork {
         // Capture topology snapshots BEFORE self is dropped (which clears them)
         let topology_snapshots = get_all_topology_snapshots(&network_name);
 
+        // Capture renewal metrics BEFORE self is dropped — `SimNetwork::Drop`
+        // calls `clear_renewal_metrics(&self.name)`, so reading the registry
+        // after this function returns would always see an empty map (#4440).
+        let renewal_metrics =
+            crate::ring::topology_registry::get_all_renewal_metrics(&network_name);
+
         // Extract the live Rings captured during the run BEFORE self drops.
         let node_rings: HashMap<NodeLabel, Arc<crate::ring::Ring>> = self
             .shared_rings
@@ -4248,6 +4283,7 @@ impl SimNetwork {
             topology_snapshots,
             node_storages,
             node_rings,
+            renewal_metrics,
         }
     }
 
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -5304,7 +5304,7 @@ impl Drop for SimNetwork {
     fn drop(&mut self) {
         use crate::node::network_bridge::set_fault_injector;
         use crate::ring::topology_registry::{
-            clear_current_network_name, clear_topology_snapshots,
+            clear_current_network_name, clear_renewal_metrics, clear_topology_snapshots,
         };
         use crate::transport::in_memory_socket::{
             clear_network_address_mappings, remove_network_socket_registry,
@@ -5315,6 +5315,7 @@ impl Drop for SimNetwork {
         set_fault_injector(&self.name, None);
         unregister_network_time_source(&self.name);
         clear_topology_snapshots(&self.name);
+        clear_renewal_metrics(&self.name);
         remove_network_socket_registry(&self.name);
         clear_network_address_mappings(&self.name);
 
--- a/crates/core/src/operations/subscribe/op_ctx_task.rs
+++ b/crates/core/src/operations/subscribe/op_ctx_task.rs
@@ -352,30 +352,48 @@ pub(crate) async fn run_renewal_subscribe(
     // `SubscriptionRecoveryGuard::complete(true)` then clears the pending mark
     // and resets the subscription backoff with success semantics.
     //
-    // It DOES refresh the LOCAL `active_subscriptions` lease, even though the
-    // lease has no upstream meaning for a root. This is purely local (extends an
-    // expiry in a local map — no wire traffic) and is what keeps the
-    // root-satisfied path cheap at scale: without it, a root with a client
-    // subscription (or recent local access) whose lease lapsed would be
-    // re-selected by `contracts_needing_renewal()` on EVERY 30s maintenance
-    // cycle (the client-subscription / local-access paths fire whenever there is
-    // no live lease), so it would consume a `MAX_RECOVERY_ATTEMPTS_PER_INTERVAL`
-    // batch slot every cycle and could starve real (non-root) wire renewals that
-    // still need to reach the network. Refreshing the lease moves the root back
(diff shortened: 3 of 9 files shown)


sanity (commented):
## Comprehensive PR Review: #4583

### Summary
- **Title:** fix: stop placement-migration subscription roots from storming on renewal (#4440)
- **Type:** fix
- **CI:** `rule-review/warnings` **failing**; `Simulation` / `Unit & Integration` pending (will fail on the broken test, Must-Fix 1)
- **Linked issues:** #4440, #4404
- **Review tier:** Full (subscribe/renewal state machine = high-risk surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture + external Codex pass

### Verdict up front
The production fix mechanism is **sound and well-scoped** — three reviewers independently verified the short-circuit logic, the local-lease rationale, the deliberate non-reuse of `complete_local_subscription`, the clean scope (only proposal 1), no removed tests, and telemetry that follows the aggregate-gauge convention rather than the #4473 per-op-event anti-pattern. But there are **two blocking issues**, one of which is a correctness/invariant regression, plus a red CI check. Needs changes + re-review before merge.

### Must Fix (blocking)

1. **The regression test is broken by the PR's own final commit and will fail CI.** *(code-first + Codex P1 + rule-review bot, independently)* Commit `93c88888` added `clear_renewal_metrics()` to `SimNetwork::Drop` (`testing_impl.rs:5318`). `run_controlled_simulation` consumes `SimNetwork` by value, so `Drop` (clearing the registry) runs before the test reads the metrics globally. `topology_snapshots`/`node_rings` are captured into `ControlledSimulationResult` before the drop; renewal metrics are not, so `get_renewal_metrics`/`aggregate_renewal_metrics` return `(0,0)` and the headline assert (`wire_attempts + terminus_satisfied > 0`) fails — the behavioral proof of the fix fails on the fixed code. **Fix:** add a `renewal_metrics` field to `ControlledSimulationResult`, populate it inside `run_controlled_simulation` before the drop (mirror the `topology_snapshots` pattern), and read `result.renewal_metrics` in the test.

2. **The root-satisfied path's unconditional lease refresh introduces unbounded `is_subscribed`-only retention — a documented-invariant regression.** *(skeptical, Medium → treated as blocking)* The new path calls `ring.subscribe(key)` every time the contract is selected. `contracts_needing_renewal()` path 1 re-selects any contract with a live lease entering the 6-min window with NO interest re-check (`hosting.rs:1373-1377`), so a body-holding root that loses interest self-perpetuates its lease forever and never leaves the renewal set. This is exactly the unbounded `is_subscribed`-only exemption that the `contract_in_use` rustdoc (`hosting.rs:924-936`) and AGENTS.md ("cleanup exemptions must be time-bounded") forbid. Pre-fix behavior was self-limiting (a failed wire renewal never refreshed the lease → it expired in 8 min). It's bounded only by hosting-budget eviction, so it grows unbounded on an under-budget node, and it perpetuates the very renewal-batch-slot consumption the PR exists to reduce (just at 6-min instead of 30s cadence). **Fix:** gate the lease refresh on genuine time-bounded interest (`contract_in_use`) so a no-interest root lets its lease lapse and leaves the renewal set (the correct end state), while an interested root still moves to the 6-min cadence (resolving the original every-cycle starvation that motivated the refresh). Confirm the path-3 recent-local-access case stays bounded by its existing age-gate.

### Should Fix

3. **`is_subscription_root` counts transient / not-ready peers**, but the actual wire renewal routes via `k_closest_potentially_hosting`, which excludes them (`ring.rs:746-755` vs `:2401`). A root whose only closer neighbor is transient/not-ready is classified "not root", skips the short-circuit, and still storms — a false-negative (safe direction, but the fix is incomplete for that edge case). Align the root check with the routing filter, or document the asymmetry. *(skeptical)*

4. **`rule-review/warnings` CI check is failing (red CI blocks merge).** Two warnings: the broken-test bug (resolved by Must-Fix 1) and a testing.md style note (the test uses the Turmoil runner with 9 min virtual time; testing.md prefers `run_simulation_direct` above 5 min). Either port the test to the direct runner (better determinism, aligns with the rule) or add the review-override label with justification (sibling placement-migration tests use the same runner, so the convention is established). *(rule-review bot)*

### Consider

5. `record_renewal_wire_attempt` is recorded before `drive_client_subscribe_inner`, which can still resolve locally (not-yet-joined / isolated), so for non-root nodes the counter is an upper bound on actual wire sends — add a clarifying comment. *(code-first)*
6. `body_holding_subscription_root_key` clones all hosted keys + linear-finds per renewed contract per cycle — bounded and infrequent (≤10 / 30s), but worth a comment given the PR's goal of reducing renewal overhead. *(code-first)*
7. The sim test's `others_min` sanity assert excludes the gateway; harmless (root placed at ring distance 0) but misleading — fold the gateway in or assert `root_dist == 0`. *(big-picture)*

### Verdict
**HEAD SHA reviewed:** 93c888889948172d449dad0854f48c87443750f8

Two must-fix issues (one a correctness/invariant regression in a high-risk area) → a fresh full review on the updated HEAD is required before merge. The underlying production fix is sound; these are fixable without rethinking the approach.

[AI-assisted - Claude]

sanity (commented):
## Re-review (HEAD 3a7bdcab) — both blockers fixed, clean

Re-ran the full multi-perspective re-review (the two must-fixes were in a high-risk area) plus the external Codex pass.

- **Skeptical: no new high/medium issues.** Verified blocker-2 (the `contract_in_use`-gated refresh) genuinely lets a no-interest root's lease lapse (no self-perpetuation) while keeping an interested root at the ~6-min cadence; the transient-only exclusion matches `k_closest_potentially_hosting` with no dangerous false-positive; capture-before-drop is correct + sim-only; no new panic/deadlock. One bounded low-sev informational note (path-3 recent-local-access roots re-selected for ≤8 min — time-bounded, strictly better than the unconditional refresh it replaced).
- **Code-first: both fixes implemented exactly as claimed and correct.** Doc-only: the PR description is stale (still describes the unconditional refresh) — should be updated to the gated behavior + the no-interest-lapse + the second regression test.
- **Testing: coverage adequate.** Independently confirmed the headline test FAILS when the short-circuit is neutralized (`wire_attempts=3, terminus_satisfied=0`) and PASSES with the fix (`wire_attempts=0`); the no-interest-lapse, non-root-stoppage, predicate-unit, and terminus-upper-bound guards are all present and green. One should-have: a source-scrape pin tying `is_subscription_root`'s routable-mapping to `k_closest`'s transient/not-ready filters (mirrors the existing `k_closest` pins) — cheap future-drift guard.
- **Big-picture: clean and in-scope.** Scope stays proposal-1-only; no removed/weakened tests; telemetry stays on the router_snapshot + hand-mirrored-OTLP + pin-test convention; the `SimNetwork::Drop` clear is per-network and coexists with the capture. One trivial stale doc-comment.
- **External Codex on this HEAD: clean** (agent-run; my independent re-run is quota-blocked until ~02:26 — I'll run my own pass on the final post-rebase content).

**Verdict:** no blockers. Mergeable after: rebase onto main (this PR and #4584 both touch `telemetry.rs`/`router.rs`/`ring.rs`; #4584 merges first), the cheap source-scrape pin + the doc/comment updates, my independent Codex pass on the rebased content, and a focused re-check of the conflict resolution.

[AI-assisted - Claude]

Issue #4565, opened 2026-06-23
Gateway freenet RSS grows toward OOM on small (2-core/8GB) boxes — vega OOM-killed at 6.96GB; allocator HWM on musl (distinct from #4549 wedge)


sanity:
## Summary

While verifying the #4549 death-spiral fix (v0.2.82 / PR #4550) on the production gateways, found a **separate, pre-existing** problem on **vega** (a small 2-core / 7.6 GB box): the freenet gateway process's anonymous-heap RSS climbs over time and can OOM-kill the box. This is **not** the #4549 libunwind wedge (that signature is absent under v0.2.82 — see below) and is filed so it isn't lost.

## Evidence

- **Prior OOM-kill on vega (pre-v0.2.82):** `Sun Jun 21 12:42:11 2026 — Out of memory: Killed process 3930943 (freenet) anon-rss:6964496kB` (~6.96 GB). This was the pre-fix incarnation (the ~21.5h crash cadence PR #4550 describes).
- **Current v0.2.82 process on vega** (`6d980ce2bc90`, up ~17.7h, NRestarts=0): `smaps_rollup` shows **Anonymous: 4.07 GB / Rss 4.11 GB**, ~54% of vega's 7.6 GB. RSS bounces 4.11–4.21 GB over 30s (churn, not monotonic), and appears up ~1 GB vs a check ~4h earlier (2.9 GB → 4.1 GB).
- **nova** (128 GB box), same v0.2.82 build, same role: Anonymous 3.23 GB — invisible at that scale, never OOMs. The problem is the small-box RAM headroom, not a nova-specific fault.

## Root cause (known)

Consistent with the allocator high-water-mark / heap-fragmentation retention previously characterized for these builds: **production binaries are musl, so jemalloc + bg_thread (the only allocator observed to purge the fragmented HWM) is compiled out** (`NO_BG_THREAD_TARGETS=["musl"]`, musl decision #2398). The system allocator retains transient op-churn allocations as anonymous heap and never returns them to the OS. On a 128 GB box this is harmless; on a 7.6 GB box it walks toward OOM.

## Why this is NOT the #4549 wedge (and is lower severity now)

[...]

PR #4584, opened 2026-06-26
perf: interest-weighted (two-tier) module-cache eviction behind a default-off flag

sanity:
## Problem

Production gateways host ~2,400 contracts but the compiled-WASM **contract** module cache fits ~1,000 modules at the measured ~1.5 MiB/module, so it constantly evicts and recompiles its working set, pinning the single-threaded contract loop (module-cache thrash, #4441). The same pressure drove the #4534 "contract queue full" / OOM regressions.

The existing eviction is pure byte-LRU and **interest-blind**: it will evict an in-use contract to keep a cold one-shot one resident. There's no production signal for how much of the cache is cold (freely reclaimable) vs in-use, so we can't tell whether a smarter eviction policy would actually help.

## Solution

Add an interest-weighted (**two-tier**) eviction policy to the **contract module cache only** (`crates/core/src/wasm_runtime/module_cache.rs`), behind a default-**OFF** feature flag `FREENET_MODULE_CACHE_INTEREST_TIERED` (mirrors the existing `FREENET_MODULE_CACHE_BUDGET_BYTES` env pattern).

- **Flag OFF (default):** eviction is byte-for-byte identical to today's pure byte-LRU. This is what makes the PR safe to merge — no behavior change unless an operator opts in.
- **Flag ON:** an over-budget eviction first reclaims **cold** (no-interest) LRU entries; only if the **interested tier alone** still exceeds the byte budget does it evict interested LRU too. The strict byte budget is **never weakened** (no #4565 OOM regression) and no contract is pinned forever.

**Interest signal:** `Ring::contract_in_use` (live local client subs OR downstream subs; deliberately excludes upstream-only, which is unbounded). It's plumbed into the cache as an `Arc<dyn Fn(&ContractKey)->bool>` predicate built in `RuntimePool::new` from a captured `Arc<Ring>`, so `wasm_runtime` keeps no dependency on `ring`. The delegate cache (no interest concept) stays pure byte-LRU.

### Behind the flag vs always-on

**Behind the flag (eviction behavior):** only the two-tier victim selection. With the flag off, nothing about eviction changes.

**Always-on (shadow metrics, independent of the flag)** — added to `router_snapshot` (the 5-min per-node aggregate, not per-op events):
- contract-cache **cold-evictable-bytes** and **interested-bytes**,
- count of evictions the two-tier policy **would reclassify** vs plain LRU,
- count of placement-migration admission decisions that **would change** (would-admit-if-cold-evictable vs the current ≥90% refuse).

These quantify whether flipping the flag (and the later #4534 admission-gate change) would help, on real production data, before anyone changes behavior. All four are mirrored into the hand-written OTLP JSON block in `tracing/telemetry.rs` with a pin test so they actually reach the collector. Per-node aggregate only; no per-contract labels.

The shadow recompute (an O(entries) interest scan) is **throttled to once per 10s** on get/insert/remove so it never regresses the hot cache path on a large cache (the O(1) occupancy gauge is still updated immediately). The #4534 SubscribeHint admission gate **behavior is unchanged** — only the would-change shadow counter sits next to it.

## Testing

Unit tests in `module_cache.rs`: cold-first ordering; interested-evicts-when-interested-alone-exceeds-budget; strict-byte-budget preservation; flag-off == today's pure LRU; cold-then-cross-into-interested mixed working set; a thrash-reduction model (hot interested set stays resident under cold churn → 0 recompiles vs plain-LRU recompiles); shadow-metric publication with the flag OFF; interest-flip-with-no-cache-mutation picked up on the next read; flag parsing; and the admission-would-change pure logic. Plus a new telemetry OTLP pin test for the shadow gauges and an updated `RuntimePool` pin test.

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and `cargo test -p freenet --lib` (3542 tests) all clean. External review: `codex review --base main` — clean after addressing three P2 findings (stale shadow gauges on interest flip; two O(entries)-per-op scans on the default path).

Refs #4441, #4440, #4534.

[AI-assisted - Claude]

---

## Operator notes

- **Restart required to flip the flag.** `FREENET_MODULE_CACHE_INTEREST_TIERED` is read ONCE at cache construction (`RuntimePool::new`), exactly like `FREENET_MODULE_CACHE_BUDGET_BYTES`. Changing it takes effect only after a node restart — there is no live re-read.
- **`migration_admission_would_change_total` is an UPPER BOUND.** The modeled cold-aware admission gate ignores the bytes the incoming migrated contract would itself add once compiled, so it admits in a superset of the cases a real gate would. The counter therefore over-counts the eventual #4534 gate's flip rate (the safe direction for sizing a future change — it never hides flip pressure).

## Review iteration 2 (post full multi-model review)

The full multi-perspective review (Codex + skeptical/testing/big-picture) on top of the initial Codex pass found no blockers but a set of polish items, all now addressed in the latest commit:

1. **Flag-OFF eviction path restored to byte- AND cost-identical pure LRU.** The per-eviction-step `would_reclassify` accounting (which scanned the cache per step when the LRU victim was interested — O(entries × evictions) in the all-interested-over-budget #4534 state, under the cache mutex) was removed from `evict_to_budget`. The default path no longer touches the interest predicate at all. The divergence signal is now computed in ONE place (the throttled `record_interest_shadow` scan), unifying the previously-duplicated flag-on/flag-off divergence logic.
2. **Shadow-gauge staleness now bounded regardless of traffic.** `RuntimePool::new` installs an on-demand refresher that the `router_snapshot` emitter invokes right before reading, so every emitted 5-min snapshot is fresh even on an idle cache. The 10s traffic-triggered throttle remains as the cheap fast path; the live admission gate stays on the O(1) occupancy gauge.
3. **A/B regression test** asserting a predicate-present-flag-OFF cache stays byte-identical (resident set + `total_bytes`) to a no-predicate cache across a churny mixed sequence.
4. **`would_reclassify` is now an approximate divergence-PRESSURE signal** sampled at the refresh cadence (not an exact per-eviction tally); docs and tests updated to match.
5/6. Upper-bound + restart-required notes added (see Operator notes above).

Re-verified: `cargo fmt`, `cargo clippy --locked -- -D warnings`, module_cache (39) + telemetry pin (5) + full freenet lib (3546) tests all green. A focused re-review (skeptical + Codex on the new shadow-accounting content) is being run before this comes out of draft.

### DIFF
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -1234,9 +1234,10 @@ mod executor_pin_tests {
             .expect("RuntimePool::new must exist")
             // Take the first chunk of the function body. Whitespace is collapsed
             // so the assertions below survive line-wrapping / reformatting of the
-            // (now multi-line, metrics-threaded) cache construction.
+            // (now multi-line, metrics-threaded, interest-predicate-threaded)
+            // cache construction.
             .split("\n    ")
-            .take(80)
+            .take(110)
             .collect::<String>()
             .split_whitespace()
             .collect::<Vec<_>>()
@@ -1246,14 +1247,26 @@ mod executor_pin_tests {
             "RuntimePool::new must size caches from config.module_cache_budget_bytes"
         );
         // The contract cache is built from `contract_cache_budget` with the
-        // "contract" label; the delegate cache from `delegate_cache_budget` with
-        // the "delegate" label. We assert each piece independently (rather than a
-        // single literal call string) so threading the metrics `Arc` through
-        // `with_label` (#4488) doesn't make this pin brittle.
+        // "contract" label (now via `with_label_and_interest`, which threads the
+        // `Ring::contract_in_use` interest predicate for two-tier eviction —
+        // #4441/#4534); the delegate cache from `delegate_cache_budget` with the
+        // "delegate" label. We assert each piece independently (rather than a
+        // single literal call string) so threading the metrics `Arc` / interest
+        // predicate doesn't make this pin brittle.
         assert!(
-            body.contains("ModuleCache::with_label( contract_cache_budget, \"contract\"")
-                || body.contains("ModuleCache::with_label(contract_cache_budget, \"contract\""),
+            body.contains(
+                "ModuleCache::with_label_and_interest( contract_cache_budget, \"contract\""
+            ) || body.contains(
+                "ModuleCache::with_label_and_interest(contract_cache_budget, \"contract\""
+            ),
             "RuntimePool::new must build the contract cache from the contract byte budget"
+        );
+        // The contract cache must thread the Ring interest predicate so two-tier
+        // eviction (and the shadow metrics) actually see contract_in_use.
+        assert!(
+            body.contains("contract_in_use"),
+            "RuntimePool::new must wire the Ring::contract_in_use interest predicate \
+             into the contract module cache (#4441/#4534)"
         );
         assert!(
             body.contains("ModuleCache::with_label( delegate_cache_budget, \"delegate\"")
--- a/crates/core/src/contract/executor/runtime/pool.rs
+++ b/crates/core/src/contract/executor/runtime/pool.rs
@@ -334,13 +334,23 @@ impl RuntimePool {
         // emitter can recompute the contract cache's interest split right before
         // it reads, keeping every emitted snapshot fresh even on an idle cache
         // (the throttled get/insert/remove refresh is unbounded on a quiet node).
-        // The closure locks the SAME shared contract cache the executors use and
-        // forces an un-throttled recompute. Cheap O(entries) scan, once per 5-min
-        // snapshot. (Codex/coordinator review — bound shadow-gauge staleness.)
-        let refresher_cache = shared_contract_modules.clone();
+        // The closure forces an un-throttled recompute on the SAME shared contract
+        // cache the executors use. Cheap O(entries) scan, once per 5-min snapshot.
+        //
+        // It captures a WEAK handle (and upgrades at call time), NOT a strong
+        // Arc: `ModuleCacheMetrics` is owned by `Ring`, each `ModuleCache` holds a
+        // strong `Arc<ModuleCacheMetrics>`, and the cache's interest predicate
+        // holds an `Arc<Ring>` — a strong clone here would close a reference cycle
+        // (metrics → refresher → cache → metrics → … → ring) and leak the whole
+        // runtime/ring on pool/node drop (simulations, tests, restarts). The Weak
+        // breaks the cycle; when the cache is gone the refresh is a no-op.
+        // (Codex review.)
+        let refresher_cache = Arc::downgrade(&shared_contract_modules);
         module_cache_metrics.set_interest_shadow_refresher(Arc::new(move || {
-            if let Ok(mut cache) = refresher_cache.lock() {
-                cache.force_refresh_interest_shadow();
+            if let Some(cache) = refresher_cache.upgrade() {
+                if let Ok(mut cache) = cache.lock() {
+                    cache.force_refresh_interest_shadow();
+                }
             }
         }));
         // Shared delegate-context cache so a prompt round-trip routed to a
--- a/crates/core/src/contract/executor/runtime/pool.rs
+++ b/crates/core/src/contract/executor/runtime/pool.rs
@@ -328,8 +328,21 @@ impl RuntimePool {
             Arc::new(Mutex::new(ModuleCache::with_label(
                 delegate_cache_budget,
                 "delegate",
-                Some(module_cache_metrics),
+                Some(module_cache_metrics.clone()),
             )));
+        // Install the on-demand interest-shadow refresher so the `router_snapshot`
+        // emitter can recompute the contract cache's interest split right before
+        // it reads, keeping every emitted snapshot fresh even on an idle cache
+        // (the throttled get/insert/remove refresh is unbounded on a quiet node).
+        // The closure locks the SAME shared contract cache the executors use and
+        // forces an un-throttled recompute. Cheap O(entries) scan, once per 5-min
+        // snapshot. (Codex/coordinator review — bound shadow-gauge staleness.)
+        let refresher_cache = shared_contract_modules.clone();
+        module_cache_metrics.set_interest_shadow_refresher(Arc::new(move || {
+            if let Ok(mut cache) = refresher_cache.lock() {
+                cache.force_refresh_interest_shadow();
+            }
+        }));
         // Shared delegate-context cache so a prompt round-trip routed to a
         // different pool executor still finds its `ctx.write()` blob.
         let shared_delegate_contexts = crate::wasm_runtime::new_delegate_context_cache();
--- a/crates/core/src/contract/executor/runtime/pool.rs
+++ b/crates/core/src/contract/executor/runtime/pool.rs
@@ -306,11 +306,23 @@ impl RuntimePool {
         // process-global) keeps the gauges per-node. Both caches share one sink;
         // they're routed apart by their `"contract"` / `"delegate"` label.
         let module_cache_metrics = op_manager.ring.module_cache_metrics();
+        // Interest predicate for the CONTRACT cache only: a contract is "of
+        // interest" while `Ring::contract_in_use` holds (a live local client
+        // subscription OR a downstream peer subscriber — deliberately NOT an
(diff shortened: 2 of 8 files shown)


sanity:
**On hold — do not merge.** I removed this from the merge queue and converted it to draft.

A full multi-model review (running on top of the Codex pass) surfaced a real issue I'd previously under-weighted: with the flag **OFF** (default), the `would_reclassify` shadow accounting still calls `has_cold_entry()` on every eviction step when the plain-LRU victim is interested. In the all-interested + over-budget case — which is exactly the #4534 pathological state this PR targets — that's a full-cache scan per eviction under the cache mutex, i.e. O(entries × evictions). That breaks the "flag-off == inert, safe to merge" premise, so the earlier merge decision was wrong.

Also flagged: shadow gauges can stay stale indefinitely once a cache goes idle after a mutation (the 10s throttle only bounds staleness on nodes with ongoing traffic).

Holding for the consolidated fix list (skeptical / testing / big-picture reviewers still in flight) so I can address everything in one pass, then re-run the full review on the new content before this comes out of draft.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4584

### Summary
- **Title:** perf: interest-weighted (two-tier) module-cache eviction behind a default-off flag
- **Type:** perf
- **CI:** was green (16/16); PR currently in **draft** (held pending this review)
- **Linked:** #4441 (refs #4534 deferred, #4565, #4473)
- **Review tier:** Full (contract module cache = high-risk surface, even though flag-off)
- **Reviewers run:** code-first, testing, skeptical, big-picture + external Codex

### Verdict up front
A clean, well-tested, genuinely inert-by-default change with the correct "measure-then-flip" shape. **No blockers.** The skeptical pass cleared the dangerous classes: no lock-ordering deadlock, no `Arc` cycle/leak, no reachable panic on a prod path, delegate cache unaffected, and the live admission gate uses the fresh O(1) occupancy gauge (not the throttled split), so shadow staleness is behavior-neutral. Scope is clean (the #4534 gate is unchanged — only a shadow counter sits beside it; no smuggling). Telemetry follows the `router_snapshot` + hand-mirrored-OTLP + pin-test convention, and the pin covers all four new gauges. One should-fix (a flag-off perf cliff) plus a few test/doc items, then a focused re-review before un-drafting.

### Should Fix
1. **Flag-OFF path is not cost-identical to historical pure LRU.** *(Codex P2; skeptical Low)* On the flag-off path the shadow accounting calls `has_cold_entry()` per eviction whenever the plain-LRU victim is interested (`module_cache.rs:436-437`); in the all-interested-over-budget case (no cold entries) that's a full-cache scan per eviction under the cache mutex — a latent perf cliff that breaks the "flag-off == historical O(1)-per-eviction" guarantee. The common gateway case is all-cold so it early-exits (hence Low in practice), but **fix it**: move the would-reclassify / divergence accounting OFF the per-eviction loop onto the already-throttled shadow refresh. That restores a truly cost-identical flag-off path AND unifies the two separate divergence-count implementations (flag-on vs flag-off) the skeptical reviewer flagged as a latent sync hazard.
2. **Shadow gauge staleness is unbounded on an idle cache.** *(Codex P2; code-first)* The throttled (10s) refresh only fires on get/insert/remove, so a cache that goes idle after a mutation can leave cold/interested-bytes + would-change stale indefinitely. Behavior-neutral (the live gate uses the fresh occupancy gauge), but the would-change signal feeds the future flip decision. Fix: refresh the interest split on the `router_snapshot` READ path so every emitted 5-min snapshot is fresh.

### Tests / Docs (should-have)
3. Add an A/B regression test: identical insert/evict sequence on a no-predicate cache vs a predicate-present-flag-OFF cache → assert byte-identical resident set + `total_bytes` after each insert (turns the "shadow can't mutate eviction" structural property into an asserted guard). *(testing)*
4. After the accounting moves to the throttled refresh, soften the doc at `module_cache.rs:394-397` — the flag-off would-reclassify is an approximation across multi-eviction inserts, not exact — and adjust/keep the would-reclassify tests accordingly. *(testing, code-first)*
5. Doc: the migration-admission would-change counter ignores the incoming contract's size, so it's an **upper bound** on the real #4534 gate's flip rate — note this in the rustdoc + PR body so the production data is read correctly. *(code-first)*
6. PR/operator note: the flag is read once at cache construction, so flipping it needs a node restart (consistent with `FREENET_MODULE_CACHE_BUDGET_BYTES`). *(big-picture)*

### Consider / later
- Nice-to-have tests: end-to-end env-flag read path; throttle-interval transitions. *(testing)*
- For the flip PR (not now): verify the flag-ON `lru_key_in_tier` O(entries) ordered scan is acceptable under the contract mutex on a full ~1000-module cache. *(big-picture)*
- `Instant::now()` per get under the cache mutex is negligible (vDSO clock_gettime, single-threaded loop); no action. *(skeptical)*

### Verdict
**State:** Needs Changes — Re-review Required After Fix (focused)
**HEAD SHA reviewed:** 43feb002

No blockers; the change is low-risk and inert by default. Address the flag-off perf cliff (1) + the staleness bound (2) + the test/doc items, then a focused re-review (skeptical + external Codex on the new shadow-accounting content) before un-drafting and merging — the fix touches the always-on prod shadow path, so the re-review is warranted.

[AI-assisted - Claude]

sanity:
**Updated — full review fix list addressed (still draft, not merging).** HEAD is now `4ac5197e`.

All six items + both nice-to-haves are in, and the two follow-on Codex P2s found on the new shadow-accounting content are fixed; the latest `codex review --base origin/main` is clean.

What changed since the first review:

1. **Flag-OFF eviction path is byte- AND cost-identical to historical pure LRU.** Removed the per-eviction-step `would_reclassify` accounting from `evict_to_budget` (it scanned the cache per step when the LRU victim was interested — O(entries × evictions) in the all-interested-over-budget #4534 state, under the cache mutex). The default path no longer consults the predicate at all. Divergence is now computed in ONE place (the throttled `record_interest_shadow` scan), unifying the previously-duplicated flag-on/flag-off logic.
2. **Shadow-gauge staleness bounded regardless of traffic.** On-demand refresher invoked by the `router_snapshot` emitter before each read → every emitted 5-min snapshot is fresh even on an idle cache. Captured as a **`Weak`** (not strong `Arc`) to avoid a metrics→refresher→cache→metrics→ring reference cycle that would leak the runtime on drop.
3. **A/B regression test**: predicate-present-flag-OFF cache vs no-predicate cache → byte-identical resident set + `total_bytes` after every insert in a churny sequence.
4. **`would_reclassify` is now an approximate divergence-PRESSURE signal** sampled at the refresh cadence; sampled AFTER the cache-hit recency update (a hit on the interested-LRU no longer false-counts). Docs + tests updated.
5. **`migration_admission_would_change_total` documented as an UPPER BOUND** (it ignores the incoming contract's size).
6. **Flag is read once at construction → restart required to flip** (documented; mirrors the budget env var).

Verification: `cargo fmt`, `cargo clippy --locked -- -D warnings`, module_cache (40) + telemetry pin (5) + full freenet lib (3547) all green. Holding in draft for your focused re-review (skeptical + Codex on the new shadow accounting) before un-draft + merge.

[AI-assisted - Claude]

sanity (commented):
## Re-review (HEAD 4ac5197e) — fixes verified

Re-ran the focused re-review on the new content (`43feb002..4ac5197e`): external Codex, skeptical, and testing.

- **Skeptical: no findings.** Verified the flag-OFF path is now genuinely predicate-free (bare `pop_lru`; `has_cold_entry` deleted) and byte- + cost-identical to historical pure LRU; the new refresh-on-snapshot-read path is deadlock-free (the `router.read()` guard is dropped *before* `refresh_interest_shadow_now()`, so the only lock order is cache-mutex → subscription-shard-guard, one-directional, no reverse path); the Weak-handle upgrade is panic-safe and repairs a real Arc-cycle leak (the intermediate strong-Arc closure would have leaked the cache+metrics+`Arc<Ring>` on node drop); `would_reclassify` accounting is correct (no false-count, sampled after the hit-recency update).
- **Testing: adequate.** The flag-off A/B byte-identity test, the four `would_reclassify` state tests, and the hit-recency regression test are all present and passing (46 module_cache + 5 telemetry pin). One should-have gap: the `pool.rs` refresher closure + the Weak-no-op-after-drop path have no direct test — being added (test-only).
- **External Codex on this HEAD: clean.** (Note for transparency: my independent re-run hit the Codex usage limit, resets ~02:26; the pass on this exact HEAD was clean.)

**Verdict:** mergeable once the cheap Weak-no-op test lands and CI is green on the final HEAD.

[AI-assisted - Claude]

PR #4586, opened 2026-06-26
fix: make auto-update checksum verification fail-closed on Linux/Windows

sanity:
## Problem

Auto-update security hardening. The Linux/Windows install path in `crates/core/src/bin/commands/update.rs` was **fail-OPEN**: if `SHA256SUMS.txt` was missing from the release, or present but had no entry for the artifact, the updater printed `Continuing without checksum verification` and installed the downloaded binary **anyway**. An unverified binary getting installed means no integrity guarantee against a corrupted mirror or a tampered asset. The macOS DMG path (`maybe_perform_bundle_update`) already *required* a checksum and refused unverified installs; Linux/Windows did not match that strictness.

## Solution

Make Linux/Windows **fail-CLOSED**, matching macOS:

- New pure helper `required_checksum(checksums, asset_name)` returns the expected hex digest or an error. A **missing manifest** (`None`) and a **missing entry** for the asset both refuse to install. The existing `verify_checksum` already refuses on a **hash mismatch**. So all three cases (missing manifest / missing entry / mismatch) now abort the install with a verification error.
- The freenet binary install now goes through this gate (`required_checksum(...)? ; verify_checksum(...)?`) before extraction/replacement.
- fdev (best-effort, must never block the service restart) now **skips** its update rather than installing unverified.

### Retry, not lockout (classification preserved)

The checksum gate runs **before** `replace_binary` — the only site that calls `record_update_failure`. So a verification failure propagates out of `download_and_install` as a plain error, the process exits 1, and the supervisor classifies it as `OtherFailure -> NoChange`: retried under the existing exponential backoff, **not** counted toward the `MAX_UPDATE_FAILURES` 3-strike lockout. A transiently-missing manifest (e.g. a brief GitHub/CDN hiccup) therefore retries rather than permanently disabling auto-update. A source-scrape pin test enforces that no `record_update_failure` can creep between the gate and the install.

### Transition check (does fail-closed break current updates?)

No. The auto-updater always targets the **latest** GitHub release, and the release workflow (`.github/workflows/cross-compile.yml`) generates and uploads `SHA256SUMS.txt` for every release. Verified:
- Latest release `v0.2.83` includes `SHA256SUMS.txt` listing exactly the asset names the updater requests (`freenet-x86_64-unknown-linux-musl.tar.gz`, `freenet-aarch64-unknown-linux-musl.tar.gz`, `freenet-x86_64-pc-windows-msvc.zip`).
- All **259** releases back to `v0.1.70` include `SHA256SUMS.txt`.

So fail-closed does not break updating to any current/recent release. (Caveat worth noting: the workflow uses `sha256sum ... || true`, so a hypothetical CI failure that produced an empty/partial manifest would now cause the updater to *refuse* rather than install unverified — which is the intended fail-closed behavior, surfaced as a retryable error.)

## Testing

- `required_checksum_missing_manifest_is_fail_closed` — `None` manifest refuses.
- `required_checksum_missing_entry_is_fail_closed` — manifest without the entry refuses.
- `required_checksum_present_entry_returns_hash` — happy path resolves the hash.
- `verify_checksum_accepts_match_and_rejects_mismatch` — real temp file; correct hash passes, wrong hash refuses.
- `checksum_refusal_is_retryable_not_a_lockout` — pins exit-1 -> `OtherFailure` -> `NoChange`.
- `checksum_gate_precedes_install_and_records_no_failure` — source-scrape pin: gate + verify precede `replace_binary`, with no `record_update_failure` in between.

`cargo fmt`, `cargo clippy -p freenet --bin freenet --all-features -- -D warnings` (clean), and the full `commands::update` module test suite (42 passed) all green locally.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/update.rs
+++ b/crates/core/src/bin/commands/update.rs
@@ -301,7 +301,14 @@ impl UpdateCommand {
 
         let fdev_asset = release.assets.iter().find(|a| a.name == fdev_asset_name);
 
-        // Try to get checksums
+        // Try to download the SHA256SUMS.txt manifest. Checksum
+        // verification is MANDATORY (fail-closed) for the freenet binary
+        // install below — see the `required_checksum` gate. We still model
+        // acquisition as `Option` here so a manifest that is absent OR
+        // failed to download flows into the same single fail-closed gate
+        // (and so the best-effort fdev path can reuse it without aborting
+        // the whole run). Any specific download error is surfaced as a
+        // warning; the actual refusal-to-install happens at the gate.
         let checksums = if let Some(checksums_asset) =
             release.assets.iter().find(|a| a.name == "SHA256SUMS.txt")
         {
@@ -312,20 +319,12 @@ impl UpdateCommand {
                 Ok(c) => Some(c),
                 Err(e) => {
                     if !self.quiet {
-                        eprintln!(
-                            "Warning: Failed to download checksums: {}. Continuing without verification.",
-                            e
-                        );
+                        eprintln!("Warning: Failed to download checksums: {}.", e);
                     }
                     None
                 }
             }
         } else {
-            if !self.quiet {
-                eprintln!(
-                    "Warning: SHA256SUMS.txt not found in release. Continuing without checksum verification."
-                );
-            }
             None
         };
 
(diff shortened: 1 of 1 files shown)


sanity:
## Codex review (external, non-Claude pass)

`codex review --base origin/main` completed with **no findings**:

> The changes make checksum verification fail closed for required artifacts and add targeted tests for the new behavior. I did not identify any discrete correctness, security, or maintainability issues introduced by the patch.

(Codex's own attempts to run the test binary failed only on environmental issues — no network / `--offline` lockfile resolution in its sandbox — not on anything in this change. Tests pass locally: full `commands::update` module, 42 passed.)

[AI-assisted - Claude]

sanity:
Reviewed the full diff: the fail-closed gate is correct (`required_checksum` errors on missing manifest/entry and propagates before `replace_binary`, so a refusal is a retryable `OtherFailure`→`NoChange`, not a lockout-incrementing failure), the fdev path correctly *skips* rather than blocking the freenet install, and the transition is safe (every release back to v0.1.70 ships `SHA256SUMS.txt`). Codex clean, 6 new tests + a source-scrape ordering pin, CI green. Part of the 0.2.84 auto-update-safety work. Merging.

[AI-assisted - Claude]

PR #4587, opened 2026-06-26
feat: add ed25519 release-artifact signing and verification to auto-update

sanity:
## Problem

Auto-update integrity is checksum-only. #4586 made the `SHA256SUMS.txt`
checksum gate fail-closed, but the manifest itself is unauthenticated: the
checksums only prove a downloaded binary matches whatever `SHA256SUMS.txt`
was served, not that the manifest is genuine. Anyone able to serve the
release assets to a node could pair a forged manifest with a forged binary
and pass the checksum gate. This adds the missing authenticity layer.

## Solution

A signature layer on top of #4586's checksum gate.

**CI signing** (`.github/workflows/cross-compile.yml`): after generating
`SHA256SUMS.txt`, sign its raw bytes with the release private key
(`FREENET_RELEASE_SIGNING_KEY` secret) using openssl raw ed25519
(`pkeyutl -sign -rawin`), producing a 64-byte detached signature uploaded
as `SHA256SUMS.txt.sig`. The key is written to a `0600` temp PEM, never
echoed, and shredded (with a cleanup trap on every exit path). A missing
secret degrades gracefully (warns; unsigned release) so the release is not
blocked during rollout.

**Binary verification** (`crates/core/src/bin/commands/update.rs`): the
release public key is baked in (`FREENET_RELEASE_PUBKEY`). The updater
downloads `SHA256SUMS.txt` and, when present, `SHA256SUMS.txt.sig`; it
verifies the signature over the **exact** manifest bytes with
`ed25519-dalek` `verify_strict`, then parses the checksums from those same
bytes (no time-of-check/time-of-use gap), and only then does the existing
fail-closed checksum gate verify the binary against the now-authenticated
manifest before install.

**Transition policy** (`REQUIRE_RELEASE_SIGNATURE = false`): when a
signature is present it is always verified and any invalid / mismatched /
wrong-length signature refuses the install (fail-closed); when absent the
install is still allowed (with a warning) so nodes can keep updating
to/from older unsigned release lineages. The flag flips to `true` in a
later release once the signed floor is established (a two-release rollout,
documented in code). All refusals propagate **before** `replace_binary`,
so they classify as a retryable `OtherFailure` -> `NoChange`, never a
`MAX_UPDATE_FAILURES` lockout (same property as #4586's checksum refusal).

The macOS DMG-bundle path keeps its checksum-only check for now; that
artifact is independently Apple-signed + notarized. The non-bundle binary
path on macOS goes through the new verification.

## Testing

- **Interop (the key guarantee):** a live test shells out to openssl to
  sign a sample manifest exactly as CI does (`pkeyutl -sign -rawin`), then
  verifies via the binary's `ed25519-dalek` path — proving
  openssl-sign <-> dalek-verify interop (raw 64-byte sig, message = exact
  bytes). A committed, hermetic openssl-produced fixture (test pubkey + a
  real openssl signature over a fixed message) is verified the same way so
  CI proves the interop even when openssl is absent at test time.
- **Unit tests:** valid signature accepted; tampered manifest rejected;
  wrong-key rejected; malformed-length rejected; absent signature allowed
  under `REQUIRE_RELEASE_SIGNATURE=false` and refused when `true`; baked-in
  pubkey matches the published hex; transition-flag tripwire.
- `cargo fmt` + `cargo clippy -- -D warnings` clean; full `freenet` binary
  test suite green (252 passed). No real private key is used in any test.

## Workflow dry-run (before relying on it)

A dedicated `verify-signing-key` job runs on `workflow_dispatch` only and is
the intended dry-run. It has no side effects (uploads nothing): it derives the
public key from the `FREENET_RELEASE_SIGNING_KEY` secret, asserts it matches
the public key baked into the binary, and does a full openssl sign/verify
round-trip with the exact command the release uses. This catches the worst
failure mode (a secret/baked-in-key mismatch would make every client reject
every release) before a real release relies on it.

Run it:

```
gh workflow run cross-compile.yml --ref feat/sign-release-artifacts
```

then check the "Verify release signing key (dry-run)" job output.

Note: the actual signing step lives in the tag-gated `attach-to-release` job,
so a plain `workflow_dispatch` does not exercise the upload path itself — only
the real signing mechanics + key correspondence (which is what matters before
trusting it). End-to-end upload + client-verify is validated on a real release
tag (and can be confirmed locally: reconstruct the DER SPKI from the 32-byte
pubkey and `openssl pkeyutl -verify -pubin -rawin -in SHA256SUMS.txt -sigfile
SHA256SUMS.txt.sig`).

Refs #4073

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/cross-compile.yml
+++ b/.github/workflows/cross-compile.yml
@@ -452,6 +452,50 @@ jobs:
           echo "=== SHA256 Checksums ==="
           cat SHA256SUMS.txt
 
+      - name: Sign SHA256SUMS.txt (ed25519)
+        # Produce a detached raw ed25519 signature over the exact bytes of
+        # SHA256SUMS.txt. The auto-update client (crates/core/.../update.rs)
+        # verifies this against a baked-in public key before trusting any hash
+        # in the manifest, which is what makes the checksum gate meaningful.
+        env:
+          # PEM-encoded ed25519 private key. Repo secret; present on tag-push
+          # and workflow_dispatch (this workflow never runs on fork PRs).
+          FREENET_RELEASE_SIGNING_KEY: ${{ secrets.FREENET_RELEASE_SIGNING_KEY }}
+        run: |
+          set -euo pipefail
+          if [ -z "${FREENET_RELEASE_SIGNING_KEY:-}" ]; then
+            # Degrade gracefully: clients tolerate an absent signature during
+            # the two-release transition (REQUIRE_RELEASE_SIGNATURE=false), so
+            # a missing secret must not fail the whole release.
+            echo "::warning::FREENET_RELEASE_SIGNING_KEY not set; release will be UNSIGNED."
+            exit 0
+          fi
+
+          # Write the key to a 0600 temp PEM and remove it on every exit path.
+          # Never echo the key (no set -x, no echo "$KEY"): it is only ever
+          # redirected into the file and passed to openssl by path.
+          KEY_FILE="$(mktemp)"
+          chmod 600 "$KEY_FILE"
+          trap 'rm -f "$KEY_FILE"' EXIT
+          printf '%s\n' "$FREENET_RELEASE_SIGNING_KEY" > "$KEY_FILE"
+
+          # Raw ed25519 sign (-rawin: message is signed as-is; ed25519 hashes
+          # internally). Produces a 64-byte detached signature. Needs openssl
+          # >= 3.0, which ubuntu-latest provides.
+          openssl pkeyutl -sign -inkey "$KEY_FILE" -rawin \
+            -in SHA256SUMS.txt -out SHA256SUMS.txt.sig
+
+          # Shred the key immediately (belt-and-braces alongside the trap).
+          shred -u "$KEY_FILE" 2>/dev/null || rm -f "$KEY_FILE"
+          trap - EXIT
+
+          SIG_SIZE=$(wc -c < SHA256SUMS.txt.sig)
+          echo "Wrote SHA256SUMS.txt.sig (${SIG_SIZE} bytes)"
+          if [ "$SIG_SIZE" -ne 64 ]; then
+            echo "::error::Expected a 64-byte raw ed25519 signature, got ${SIG_SIZE} bytes"
+            exit 1
+          fi
+
       - name: Upload binaries to release
         env:
           # Coalesce on RELEASE_PAT so the `gh release edit --draft=false`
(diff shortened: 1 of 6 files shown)


sanity:
## Review summary (multi-model)

High-risk surface (crypto + CI), so a Full review was run: external **Codex** passes plus three independent adversarial **Claude** subagent lenses (security/crypto, CI-shell/secret-handling, test-correctness/interop) for added perspective diversity (the `freenet:*` reviewer agents are unavailable).

**Findings, all addressed:**

- **[Codex P2] Stale signature on unsigned rerun** — `--clobber` only replaces listed assets, so a leftover `SHA256SUMS.txt.sig` from a prior signed run could be verified against a new manifest and brick the release. Fixed: the upload step now deletes any stale `.sig` when no signature is produced.
- **[Codex P1] macOS bundle path trusted an unsigned manifest** — `maybe_perform_bundle_update` now goes through the same signature-verifying `download_and_verify_checksums` path as Linux/Windows (Apple notarization doesn't bind the DMG to *this* release's manifest, so a forged-manifest/downgrade was possible). The now-unused `download_checksums` helper was removed.
- **[rule-review warning] Stale docs** — `FREENET_RELEASE_SIGNING_KEY` added to `docs/RELEASING.md` prerequisites + `AGENTS.md`.
- **[Claude security, Low] Unbounded `download_bytes`** — capped at 4 MiB (streamed) so a compromised asset server can't OOM the updater.
- **[Claude CI, Low] Empty manifest could be signed silently** — signing step now refuses an empty `SHA256SUMS.txt`.

**Verified clean by all reviewers:** signature is verified over the exact bytes the checksums are parsed from (no TOCTOU); all refusals propagate before `replace_binary` (retryable `OtherFailure`, never a lockout); `verify_strict` used correctly; baked-in pubkey matches the published hex; no real private key in any test; the openssl↔dalek interop is proven both live and via a committed hermetic fixture.

Final Codex pass on the latest content: no actionable issues.

**Known residual (informational, pre-existing):** the signed payload carries no version/timestamp, so the signature alone doesn't prevent rollback to an older *legitimately-signed* release — that's bounded by the existing `latest > current` check + TLS to api.github.com, unchanged by this PR.

[AI-assisted - Claude]

sanity:
Reviewed and approving. Verify logic is sound: `verify_strict` over the exact manifest bytes, checksums parsed from that same buffer (no TOCTOU), sig→checksum→install ordering, refusals propagate before `replace_binary` (retryable `OtherFailure`→`NoChange`, not a lockout), `REQUIRE_RELEASE_SIGNATURE=false` two-release transition (present→verified+reject-on-mismatch, absent→allowed). The baked `FREENET_RELEASE_PUBKEY` is pinned by a test, and the `verify-signing-key` dry-run job passed — confirming the baked key matches the `FREENET_RELEASE_SIGNING_KEY` secret (the critical pre-release check). External Codex (3 passes; P1 macOS-bundle + P2 stale-sig fixed) + 3 adversarial lenses, openssl↔dalek interop test passing, CI green. Transition-safe (signing only activates on a release tag). Part of the 0.2.84 auto-update-safety work. Merging.

[AI-assisted - Claude]

PR #4588, opened 2026-06-26
fix: distinguish fast-crash from update exit to bound #4551 crash loops

sanity:
## Problem

#4570 moved `StartLimitBurst`/`StartLimitIntervalSec` into the `[Unit]` section so systemd now actually honours the crash-loop limiter (they were silently ignored in `[Service]`). With the limiter effective, gaps remained:

1. The unit's `SuccessExitStatus=42 43` whitelists the node's fatal-listener exit code (42, from #4549). A node stuck in a **boot-wedge loop** exits 42 every ~10s and never counts toward `StartLimitBurst` — so the limiter wouldn't bound the one case it most needs to.
2. The limiter's terminal action was implicit (a tripped loop should stop the unit, not reboot the host).

## Solution

Builds on #4570. Adds a crash-vs-update distinction so the now-effective limiter bounds real boot-wedge loops **without** losing the auto-update self-heal, on **every** supervision path.

### systemd path
- **`StartLimitAction=none`** (default, made explicit) in both units: a tripped loop STOPS the unit rather than rebooting the host. Recover with `systemctl reset-failed freenet && systemctl start freenet`.
- **`FAST_CRASH_EXIT_CODE = 45`** (kept OUT of `SuccessExitStatus`) + a min-uptime guard, `fatal_listener_exit_code(uptime, fast_crash_enabled)`:
  - fatal listener exit **< 60s uptime** → exit **45** (counted, so a tight loop trips `StartLimitBurst` and the unit stops);
  - **>= 60s** → exit **42** (burst-exempt).
- **`ExecStopPost` runs `freenet update` on exit 42 OR 45** (via a `case`, avoiding `&&`/`||` precedence pitfalls). Counting (StartLimit) and self-heal (ExecStopPost) are independent, so a boot-crash that a newer release fixes still self-heals (#4549) even though exit 45 is rate-limited — `freenet update` is a separate process that can succeed even when `freenet network` boot-crashes.

### cross-platform / cross-version safety
Exit 45 is meaningful only under a unit that handles it, so it is emitted **only when the binary opts in** (`enable_fast_crash_exit_code()`, mirroring the existing `enable_abort_on_fatal_listener_exit()`), and the entry point opts in **only when the supervising unit advertises 45 support** via a marker env var (`FREENET_SYSTEMD_FAST_CRASH`) that the **regenerated** systemd unit sets. This ties runtime behavior to the on-disk unit's actual capability:

- macOS/Windows in-process **run-wrapper** (knows only 42; self-heals on 42 and bounds the loop with its own backoff + 50-failure cap) → no marker → keeps emitting **42**.
- **Old/custom systemd unit** — including a node auto-updated to this binary but whose unit file was never regenerated → no marker → keeps emitting **42** (its existing exit-42 self-heal works).
- **Unsupervised** → no marker → **42** (moot).

The marker is referenced through the `SYSTEMD_FAST_CRASH_ENV_VAR` const in both the unit template and the entry-point check (single source of truth, no literal drift).

## Testing
- `node::p2p_impl`: fast-crash code distinct + counted; uptime split when enabled; **always 42 when disabled** (run-wrapper / old-unit / unsupervised); both the abort and fast-crash flags are opt-in / off-by-default.
- `service.rs`: `StartLimitAction` in `[Unit]`; exit-45 out of `SuccessExitStatus`; `ExecStopPost` fires update on `42|45`; fast-crash marker present (asserted via the const).
- `StartLimitBurst`/`IntervalSec` `[Unit]`-placement is covered by #4570's `linux.rs` tests (not duplicated). Those `#4570` helpers are also hardened to find section headers **by line**, not substring, so the legitimate `[Service]` reference now in a `[Unit]` comment isn't misparsed.
- `cargo fmt`; `cargo clippy --locked -- -D warnings` and `--features trace-ot` both clean. Both generated units pass `systemd-analyze verify` (exit 0; only the pre-existing ExecStartPre shell-escape notice).

Refs #4551 (the section-placement fix landed in #4570); this completes the crash-loop bounding. Builds on #4570.

> [!NOTE]
> The systemd burst-behavior change (counting exit 45 toward `StartLimitBurst`, `StartLimitAction=none`, marker-gated opt-in) wants a smoke-test on a live systemd host before the next release, per `.claude/rules/deployment.md` (a `cfg(target_os = "linux")` service path CI does not exercise end-to-end).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -43,6 +43,21 @@ pub const EXIT_CODE_UPDATE_NEEDED: i32 = 42;
 /// operator, loudly, when an update was detected but will not be applied.
 pub const SUPERVISED_ENV_VAR: &str = "FREENET_SUPERVISED";
 
+/// Environment variable set ONLY by the freshly-generated Freenet **systemd** units
+/// (see `service/linux.rs`) on the `freenet network` child. Its presence is positive
+/// evidence that the supervising unit understands the distinct fast-crash exit code
+/// 45 (#4551): the unit keeps 45 OUT of `SuccessExitStatus` (so it counts toward
+/// `StartLimitBurst`), sets `StartLimitAction=none`, and fires `ExecStopPost`
+/// `freenet update` on exit 42 OR 45.
+///
+/// The node entry point gates [`freenet::enable_fast_crash_exit_code`] on this
+/// marker. Unlike [`SUPERVISED_ENV_VAR`], the macOS/Windows run-wrapper does NOT set
+/// it (the wrapper only understands exit 42), and an OLD systemd unit (e.g. a node
+/// auto-updated to a 45-aware binary but whose unit file was never regenerated) won't
+/// have it either — so in those cases the node keeps emitting the burst-exempt
+/// self-healing exit 42 rather than a 45 the supervisor would mishandle.
+pub const SYSTEMD_FAST_CRASH_ENV_VAR: &str = "FREENET_SYSTEMD_FAST_CRASH";
+
 /// Whether the running node appears to be under a supervisor that will catch
 /// exit code 42 and apply the update.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -460,6 +460,54 @@ mod tests {
     use super::wrapper::*;
     use std::path::PathBuf;
 
+    /// Return the section header (e.g. `[Unit]`, `[Service]`) that a given `key=`
+    /// directive line appears under in a generated systemd unit, or `None` if the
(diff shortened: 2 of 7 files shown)


sanity:
Reviewed and approving. Verified end to end: fast-crash exit-45 is emitted ONLY when the freshly-generated systemd unit sets the `SYSTEMD_FAST_CRASH_ENV_VAR` marker, so the macOS/Windows wrapper, unsupervised nodes, and a node that auto-updated the binary but not yet its old systemd unit all keep emitting exit-42 (existing self-heal + loop-bounding intact) — no regression on any path. The new unit fires `freenet update` on exit 42 OR 45 (boot-crash self-heal) while keeping 45 out of `SuccessExitStatus` so `StartLimit` counts it and a tight loop is bounded. Comprehensive tests (marker gating, ExecStopPost 42|45, exit-code selector, systemd-analyze), CI green. Completes #4551 with #4570. Merging.

[AI-assisted - Claude]

PR #4589, opened 2026-06-26
feat: periodically re-poll GitHub for new releases so long-running nodes auto-update

sanity:
## Problem

Auto-update *detection* is effectively once-at-boot. After the one-shot startup
GitHub check (#3864), the update-check loop's three triggers are all driven by
peer signals:

1. **Urgent update** — a peer reports `min_compatible > our version`.
2. **Decentralized discovery** — a peer handshake reports a `highest_seen_version` newer than ours.
3. **Legacy mismatch** — the older backoff-based mechanism.

None of these re-poll GitHub on their own. If an entire network sits on the same
old version, no node ever sees a newer peer version, so no node re-checks GitHub
directly. A freshly published release is then never picked up until a node
happens to restart (which re-runs the startup check). A node that has been up for
weeks never re-checks, and the network can stay frozen on an old version.

## Solution

Add a fourth, peer-signal-independent trigger to the existing update-check loop:
a **periodic direct GitHub re-poll** that re-runs the same `startup_update_check`
on a recurring, jittered schedule, so a long-running node notices a new release
without needing a restart.

- **Detection only.** A discovered update feeds the *same* `update_tx` →
  graceful-shutdown → exit-42 → `freenet update` path as every existing trigger
  (and identically to the boot-time startup check). The apply/verify/signing
  path in `commands/update.rs` is untouched, so all merged safety still applies
  unchanged: fail-closed checksum gate (#4586), ed25519 signature verification
  (#4587), and crash-loop bounding (#4551 / #4588).
- **Interval: 6h, jittered ±25%.** 6h is far under GitHub's unauthenticated REST
  rate limit (60 req/hr/IP) — even at the jittered minimum (~4.5h) it is well
  under one request/hour, leaving headroom for the bursty peer-signal checks
  that share the IP budget; yet it is frequent enough that a release propagates
  within hours rather than never. The ±25% jitter decorrelates nodes that booted
  together so they do not re-poll (and therefore exit-42/restart) in lockstep —
  preserving the load-spreading intent of the existing 0-60s startup jitter and
  0-4h decentralized-discovery stagger. Jitter is drawn from `GlobalRng` (runtime
  jitter; does not affect deterministic simulation seeding).
- **Fail-open + self-throttling.** `startup_update_check` returns `None` on any
  GitHub / parse / network error, so a failed poll (rate-limit, outage) simply
  retries at the next interval — it never hammers the API or crashes the node.
- **Respects existing gates.** It lives inside the existing `update_check_task`,
  so it inherits the dirty/dev-build suppression (the task early-returns for
  `GIT_DIRTY` builds before the loop). The re-poll runs sequentially in that one
  task and advances its deadline before checking, so it never double-fires or
  overlaps an in-flight check, and it is cancellation-safe (the only `.await` is
  in the loop body, not inside a `select!` branch future).
- **Honors the persistent failure lockout (#3934).** The re-poll is gated on
  `should_attempt_update()`. If installs keep failing (e.g. a non-writable
  binary path), the failure counter reaches `MAX_UPDATE_FAILURES` and the
  re-poll skips — so a locked-out node does not exit-42 once per interval and
  drive the supervisor to rerun the same failing update. (The boot-time startup
  check bypasses the lockout, but only once per restart; a *recurring* bypass
  would reintroduce the loop the lockout prevents.) A successful manual
  `freenet update` clears the counter and re-enables the re-poll.

## Testing

- `cargo fmt`, `cargo clippy -p freenet --bin freenet -- -D warnings`: clean.
- `cargo test -p freenet --bin freenet`: 261 passed, 0 failed.
- New unit tests for the (pure, injectable) interval/jitter logic
  (`jittered_repoll_interval`), so it is testable without waiting hours:
  - bounds: result always within ±25% of base for any `rand_unit` in `[0,1]`;
  - exact endpoint/midpoint mapping (0.0 → -25%, 0.5 → base, 1.0 → +25%);
  - determinism (no hidden global state);
  - input clamping (out-of-range `rand_unit` / `jitter_fraction`);
  - rate-limit sanity: the jittered-minimum interval is ≥ 1h (≥60× margin under
    GitHub's 60 req/hr limit).
- A source-scrape pin test (`periodic_repoll_respects_update_lockout`) asserting
  the re-poll gates on `should_attempt_update()`, so a future edit cannot
  silently drop the #3934 lockout gate.
- External review: `codex review --base origin/main` raised one finding (P2) —
  the re-poll bypassing the #3934 failure lockout. Addressed by the
  `should_attempt_update()` gate above and pinned by the new test.

Refs #4073.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -860,6 +860,56 @@ pub(crate) fn compare_versions_for_startup(current: &str, latest: &str) -> Optio
     }
 }
 
+// =============================================================================
+// Periodic re-poll scheduling (#4073)
+// =============================================================================
+
+/// Base interval between periodic direct-GitHub update re-polls.
+///
+/// After the one-shot startup check (#3864), update detection otherwise
+/// depends entirely on a peer signal (urgent / highest-seen / version
+/// mismatch). If an entire network sits on the same old version, no node
+/// re-checks GitHub and a freshly published release is never picked up until
+/// a node happens to restart. This recurring poll closes that gap: a
+/// long-running node re-asks GitHub directly on this cadence, without needing
+/// a restart.
+///
+/// 6h is deliberately conservative against GitHub's unauthenticated REST rate
+/// limit (60 requests/hour/IP): even at the jittered minimum (~4.5h) it is
+/// well under one request/hour, leaving headroom for the bursty
+/// peer-signal-driven checks that share the same IP budget. It is still
+/// frequent enough that a release propagates across the network within hours
+/// rather than never.
+pub const UPDATE_REPOLL_INTERVAL: Duration = Duration::from_secs(6 * 3600);
+
+/// Fraction of jitter applied to each re-poll interval (±25%).
+///
+/// Without jitter, nodes that booted together (e.g. after a coordinated
+/// restart or outage) would re-poll — and therefore exit-42/restart — in
+/// lockstep, producing a thundering herd against both GitHub and the network.
+/// ±25% decorrelates them, preserving the load-spreading intent of the
+/// existing 0-60s startup jitter and 0-4h decentralized-discovery stagger.
+pub const UPDATE_REPOLL_JITTER_FRACTION: f64 = 0.25;
+
+/// Apply symmetric `±jitter_fraction` jitter to `base`, given a uniform random
+/// sample `rand_unit` in `[0, 1]`.
+///
+/// Pure and deterministic: the randomness is injected by the caller (in
+/// production, `GlobalRng`) so the scheduling math is unit-testable without
+/// waiting hours or depending on a clock. `rand_unit` and `jitter_fraction`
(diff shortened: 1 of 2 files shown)


sanity:
Reviewed + merging. Rebased onto the post-#4591/#4593 main; the rebase git-auto-merged cleanly but that left a semantic gap (the periodic re-poll's discovered-version arm wasn't guarded by the pinned-bad/install-gate check that #4591/#4593 added everywhere else), which was caught and hand-fixed: the re-poll now skips exit-42 for a locally-blocked version and routes through `get_latest_version` so it honors the global GitHub rate-limit bucket. Confirmed via read + the source-scrape pins (`periodic_repoll_respects_update_lockout`, `pinned_known_bad_never_triggers_exit_42`, `rate_limited_never_triggers_exit_42`). Codex clean, 313 tests pass, CI green. Reach piece — lands after the full safety set (#4586/#4587/#4570/#4588/#4591/#4593).

[AI-assisted - Claude]

PR #4591, opened 2026-06-26
feat: crash-loop auto-rollback for the auto-updater (#4073)

sanity:
## Problem

Auto-update has no recovery path when a node updates to a release that then
crash-loops on startup. The supervisors' existing crash bounds only *stop* the
node — systemd `StartLimitBurst` + `StartLimitAction=none`, and the run-wrapper's
consecutive-failure cap — so a bad release leaves the node down until an operator
intervenes. This is the brick-safety gap in auto-update: a wedged release defeats
the unsupervised-restart self-heal it is supposed to provide.

## Solution

A per-node crash-loop auto-rollback that **reuses the existing crash detection**
rather than adding a second watchdog. New `commands::rollback` module owns the
state machine (pure, directory-injectable, fully unit-tested).

Lifecycle:

1. **Install** — before overwriting the running binary, `freenet update`
   snapshots it as the *known-good* rollback target and marks the new version
   *on-probation*.
2. **Detect** — every supervisor that runs `freenet update` immediately after the
   node stops (systemd `ExecStopPost`, the macOS launchd wrapper script, and the
   in-process run-wrapper) now forwards the node's exit code via
   `FREENET_POST_STOP_EXIT_CODE`. `freenet update` reads it and, while in
   probation, counts the stop as a crash of the probationary version (no network
   call needed on the crash path).
3. **Revert + pin** — on the 3rd probation crash (`< StartLimitBurst=5`, so it
   fires *before* systemd stops the unit) the known-good binary is restored and
   the bad version is pinned locally as known-bad so it is not auto-re-applied.
4. **Commit** — the node clears probation after running healthily for 60s (mirrors
   the existing fast-crash window), so ordinary later crashes never roll back.

The installer **and** the node's update checks refuse a pinned known-bad version
(a strictly-newer fix is still applied normally), which breaks the
update → crash → revert → re-update loop.

### Brick-safety (no new flap/loop)

- Rollback only ever restores the known-good target captured *before* the
  probationary install — always a version that previously ran (one generation).
- After a rollback there is no probation marker, so a still-crashing reverted
  version is **not** rolled back again; it falls to the supervisor's own limiter
  / the operator.
- If no known-good binary exists we refuse to "roll back" rather than risk
  deleting the only working binary.
- The handoff is an env var (not a CLI flag), so it is forward/backward
  compatible: an older binary we rolled back *to*, or an older unit, ignores it.
  `#4588`'s exit-45 / probation behaviour is preserved.

### Per-OS

- **Linux/systemd** (production): `ExecStopPost` forwards `$EXIT_STATUS`; rollback
  runs from `freenet update`. Primary path.
- **macOS launchd wrapper script** and **Windows/macOS in-process run-wrapper**:
  forward the exit code on the post-crash update only (manual/tray "Check for
  Updates" pass nothing, so they are never miscounted as crashes).
- **Unsupervised** / dev (dirty) builds: no behaviour change.
- **macOS `.app` DMG-swap updates** are intentionally out of scope (reverting a
  signed bundle is a separate mechanism); the bundle path degrades to the
  pre-change behaviour and is documented inline. Tracked as follow-up.

## Testing

- 12 unit tests for the state machine: probation roundtrip + commit, stale-marker
  cleanup, crash-records-then-rolls-back-and-pins, pinned-not-reapplied (newer
  allowed), one-generation bound, rollback-unavailable when known-good is missing,
  known-good preserved across a chained in-probation install, restore preserves
  source, corrupt-marker-reads-as-none. A compile-time static assert pins
  `ROLLBACK_CRASH_THRESHOLD < StartLimitBurst`.
- Per-OS source/template pins: systemd user+system units, the macOS shell wrapper,
  and the in-process wrapper all forward the exit code on the post-stop update only.
- `cargo fmt`, `cargo clippy --bin freenet -- -D warnings`, and
  `cargo test -p freenet --bin freenet` (272 tests) all green.
- Codex review run; its one finding (macOS bundle scope) addressed by documenting
  the boundary inline.

Not yet done (flagged for review): live systemd verification on a real node
(force a crash-looping install and confirm the revert + pin fire) before release.

Refs #4073

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -266,10 +266,13 @@ pub async fn check_if_update_available(current_version: &str) -> UpdateCheckResu
                     );
                     // Distinct result (NOT Skipped) so the caller clears the
                     // driving signal and does not fall through to the legacy
-                    // max-backoff exit-42 fallback. Reset the GitHub-check
-                    // backoff so a later fix (a version newer than the pin) is
-                    // still noticed promptly once peers re-signal.
-                    reset_backoff();
+                    // max-backoff exit-42 fallback. GROW the GitHub-check backoff
+                    // (do NOT reset it): a node that has already rolled back to a
+                    // good version is in no hurry to find the fix, so it should
+                    // poll at most hourly rather than at the 60s floor while
+                    // peers keep advertising the pinned-bad version. An hourly
+                    // check still catches a later strictly-newer release.
+                    increase_backoff();
                     return UpdateCheckResult::PinnedKnownBad;
                 }
                 tracing::info!(
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -242,6 +242,21 @@ pub async fn check_if_update_available(current_version: &str) -> UpdateCheckResu
             };
 
             if latest_ver > current {
+                // #4073 crash-loop auto-rollback: never trigger an exit-42
+                // update to a version pinned known-bad on this node by a prior
+                // rollback. The installer would refuse it anyway; skipping here
+                // also avoids a pointless restart cycle. The mismatch flag is
+                // kept (Skipped) so a later, strictly-newer fixed release is
+                // still picked up.
+                if super::rollback::is_version_pinned_bad(&latest) {
+                    tracing::warn!(
+                        version = %latest,
+                        "Newer version is pinned known-bad after a crash-loop rollback; not \
+                         triggering auto-update to it (#4073)"
+                    );
+                    increase_backoff();
+                    return UpdateCheckResult::Skipped;
+                }
                 tracing::info!(
                     current = %current_version,
                     latest = %latest,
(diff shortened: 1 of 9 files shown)


sanity:
Reviewed and approving after a full multi-model + multi-lens process (this is brick-safety, so it got the heavy treatment):

- **Round 1:** Codex (clean) + 3 adversarial Claude lenses (brick/thrash, exit-code+supervisor, pin/probation). The lenses caught what Codex didn't: a coverage gap (rollback only fired on exit 42/45, missing panics/segfaults/OOM/early-errors — the dominant boot-crash class), an exit-42 forward-update miscount (spurious *backward* rollback), and several brick/thrash vectors (no blob integrity check, a Unix restore power-loss window, pin-write-failure oscillation, no marker TTL).
- **Rework:** all addressed (broadened crash detection to any non-graceful exit, gated to systemd via the marker; exit-code classification; SHA-256+fsync blob integrity verified before restore; atomic Unix rename; pin-before-restore; 1h marker TTL) plus 4 Codex-iterative fixes.
- **Round 2 (re-review on the reworked content):** all three lenses verified the fixes landed and the broadening did NOT disturb #4588's StartLimit machinery; PinnedKnownBad can't slow-loop; no new brick vector. Only LOW polish remained.
- **Polish:** backoff fix on the pinned-bad path, consolidated known-good blob+label derivation (can't diverge) with TTL parity, 10s timeout on get_latest_release. Codex's 5th pass clean.

283 tests pass, clippy/fmt clean, CI green on dd3d0b67. One accepted+documented tradeoff: a transient triple-OOM could spuriously roll back (a genuine memory-regression OOM *should* roll back) — flagged to watch on the 2GB-capped peers post-deploy. Live-systemd end-to-end verification (force a crash-loop on a real supervised node) remains a pre-release check. Merging.

[AI-assisted - Claude]

PR #4593, opened 2026-06-27
fix: bound auto-update network load so a bad release can't spam GitHub/gateways

sanity:
## Problem

An aggregate-load audit of the auto-updater found an **unbounded failure loop**. When a release's install can never succeed (for example a checksum/signature that does not verify, so the fail-closed gate added in #4586 refuses every attempt), a node can loop indefinitely:

1. the node detects the newer version and exits 42 (update needed),
2. the supervisor runs `freenet update`, which fails the verification gate, so nothing is installed,
3. that failure is classified as "no change" (not a crash, not a lockout), so nothing counts it,
4. the supervisor restarts the node, which re-detects the same version — back to step 1.

Each iteration costs GitHub REST polls and a gateway reconnect, and the on-crash `freenet update` is a fresh process with no persistent rate limit, so the loop is unbounded fleet-wide. The bad-binary *crash* loop is already handled by #4591 (auto-rollback); this is the distinct *install-failure* loop, which #4586's fail-closed gate converted from a self-terminating path into a non-terminating one.

## Solution

Three layered bounds (none touch the #4591 rollback state machine, #4586 checksum verify, #4587 signing, or #4588 exit-code machinery):

1. **Persistent GitHub-poll rate limit (token buckets).** Every GitHub release poll goes through an on-disk token bucket in `state_dir()`, so the limit holds across the fresh one-shot `freenet update` process and restarts. Two **independent** buckets — one for the node's `get_latest_version`, one for the supervisor-side `get_latest_release` — because the node always polls first in a restart cycle and a shared bucket would let it starve the installer. Capacity 8, refill 1 token / 10 min ⇒ ≤ ~6 polls/hour/node per path. Missing bucket = full (first boot works); corrupt or unpersistable = deny (fail closed); time-injectable for tests.

2. **Per-target-version install-failure gate** (mirrors the #4591 known-bad pin, but for install failures). After 3 **deterministic** install failures of the *same* version (checksum mismatch or invalid signature), the node's update detection and the installer stop acting on that version until a strictly-newer release appears. This is what actually stops the restart loop. Transient failures (download/network outage, a still-propagating manifest, extraction errors) are deliberately **not** counted, so a brief outage can't permanently suppress a good release. Atomic writes, degrade-safe reads.

3. **Parity gaps.** The macOS launchd wrapper gains a consecutive-failure cap (exits the loop; targets a tight crash *loop* via a healthy-runtime reset; treats a benign "nothing to do" updater exit as not-a-failure; exits 0 so launchd does not respawn). The systemd units gain `RestartSteps` + `RestartMaxDelaySec` restart backoff (systemd ≥ 254; ignored gracefully on older), reducing reconnect churn during any residual loop.

## Testing

New unit/regression tests cover: the token bucket (capacity, refill, refill cap, deny-on-corrupt, fail-closed-on-unpersistable, missing-is-full, backwards-clock and dip-then-recover give no credit, the two buckets are independent); the install-failure gate (engages after N, allows a strictly-newer version, resets on target change, clears on success, degrade-safe, atomic, and an end-to-end "loop is bounded by the gate" test); the verification-vs-transient classification; and source-scrape pins that `get_latest_version`/`get_latest_release` consult their buckets, that `RateLimited` and `PinnedKnownBad` arms never exit 42, that a rate-limited poll records no check-time/backoff, that the install gate only counts verification failures, and the macOS wrapper cap/healthy-reset/exit-2 behavior. `cargo fmt`, `cargo clippy --locked -- -D warnings`, and `cargo test -p freenet --bin freenet` (307 tests) are green.

External review: `codex review --base origin/main` was run iteratively; all findings were addressed (separate buckets, fail-closed persistence, backwards-clock timestamp, macOS exit-0/healthy-reset/exit-2, rate-limited result not feeding the exit-42 fallback, check-time not recorded for denied polls, transient-vs-verification gating) and the final pass reported no actionable regressions.

## Scope note

This per-node load-bounding targets the per-node auto-update crash-loop path (#4073). It intentionally does **not** cover the gateway-only release-agent's own GitHub fetch (`crates/release-agent/src/github.rs`, `GET /repos/{repo}/releases/latest`): that is the release-workflow gateway-push path, invoked once per release, not the per-node restart loop, so it is not an aggregate-load amplification vector and is out of scope here.

Refs #4073.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -367,7 +367,7 @@ async fn get_latest_version() -> Result<String> {
     // at `get_latest_release`. A denied poll returns Err so the caller
     // (`check_if_update_available`) treats it as `Skipped` and retries later —
     // the node keeps running, it just does not poll GitHub this tick.
-    if !try_consume_github_poll() {
+    if !try_consume_node_poll() {
         return Err(RateLimitedError.into());
     }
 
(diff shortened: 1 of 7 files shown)


sanity:
Reviewed and approving after a full review (this bounds the auto-update network-load failure mode an aggregate-load audit surfaced — a bad-manifest/sig release could otherwise loop forever, ~100-170 GitHub req/hr/node + a gateway reconnect/min fleet-wide, made infinite by #4586's correct fail-closed change).

Review: external Codex (8 iterative rounds, final pass clean) + 3 adversarial lenses — rate-limit bypass/amplification, install-gate false-positives, and end-to-end loop-termination + cross-platform parity + no-regression. No Must-fix. Key things verified:
- **Rate-limit:** on-disk token bucket (cap 8, refill 1/10min) at both GitHub-fetch choke points; fail-closed on corrupt/unpersistable; the fresh on-crash `freenet update` honors it; no bypass path; sound refill math (clock-rewind safe); normal ops have comfortable headroom.
- **Install-gate (the loop-stopper):** only DETERMINISTIC verification failures (checksum/sig) count — transient download/IO/missing-manifest errors deliberately do NOT, so a brief outage can't wrongly gate a good release; a strictly-newer version supersedes (no permanent freeze); degrade-safe.
- **Parity + no-regression:** macOS wrapper failure cap works; systemd RestartSteps backoff documented honestly re: the #4551 tradeoff (fast crashes still terminal-stop via StartLimit, confirmed in the live-systemd verify; medium crashes slow-flap gently); #4591 rollback / #4586 checksum / #4587 signing / #4588 exit-codes all intact (separate counters, no collision).

307 tests pass (incl. token-bucket edge cases, gate classification, end-to-end loop-bound, source-scrape pins). CI green on 3d770e36. Merging.

[AI-assisted - Claude]

PR #4594, opened 2026-06-27
feat: bake reserved ed25519 revocation public key for future key-revocation

sanity:
## Problem

Freenet's auto-updater authenticates the release `SHA256SUMS.txt` manifest
against a single ed25519 key baked into the binary (`FREENET_RELEASE_PUBKEY`),
whose private half lives online in the `FREENET_RELEASE_SIGNING_KEY` CI secret.
If that online key were ever compromised, there is currently no key a peer
already in the field would trust to honor a revocation/recovery message.

A revocation key only protects peers whose binary **already contains it** — it
cannot be retrofitted into builds that have already shipped. So the key has to
be baked in now, well ahead of any mechanism that uses it.

## Solution

Add a second, reserved ed25519 **public** key, `FREENET_REVOCATION_PUBKEY`,
baked into the binary alongside `FREENET_RELEASE_PUBKEY`:

- Its private half is kept strictly **offline** (never in CI or GitHub
  secrets, unlike the release-signing key), so it can authorize a
  key-revocation message even if the online signing key is stolen.
- It is **RESERVED**: there is intentionally no fetch/verify mechanism wired
  up yet. That code is deferred to ride the future signed-policy-over-Freenet
  delivery layer (auto-update epic #4073). This PR adds only the baked-in key
  plus a pin test that locks its value.

Declared as a `#[used] static`, **not** a `const`, on purpose: a `const`
referenced only from a test is inlined to nothing and never appears in the
shipped binary — which would defeat the entire bake-in. `#[used]` forces the
bytes into the artifact and retains them against linker dead-code elimination
even though nothing reads the key at runtime yet. (`#[used]` also suppresses
the dead_code warning, so no `#[allow]` is needed.)

## Testing

- `cargo fmt`, `cargo clippy --bin freenet -- -D warnings` — clean.
- `cargo test -p freenet --bin freenet` — 308 passed, including the new
  `revocation_pubkey_matches_published_hex` pin test (asserts the 32 bytes
  match the documented hex and form a valid ed25519 point).
- Verified the bake-in empirically: the 32 key bytes are **absent** from
  `target/release/freenet` when the key is a plain `const`, and **present**
  in the optimized release artifact as a `#[used] static`.

Refs #4073

[AI-assisted - Claude]

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -285,6 +285,24 @@ without cutting a release, run `cross-compile.yml` via `workflow_dispatch` and
 check the `verify-signing-key` job. The full prerequisites table (including the
 two-release `REQUIRE_RELEASE_SIGNATURE` transition) is in `docs/RELEASING.md`.
 
+### Reserved offline revocation key (`FREENET_REVOCATION_PUBKEY`)
+
+A second ed25519 public key, `FREENET_REVOCATION_PUBKEY`, is baked into the
+binary alongside `FREENET_RELEASE_PUBKEY` but serves a different purpose: it is
+the OFFLINE backup / key-revocation key. Its private half is kept strictly
+offline (never in CI or GitHub secrets, unlike `FREENET_RELEASE_SIGNING_KEY`),
+so it can authorize a key-revocation message even if the online release-signing
+key is compromised. It is RESERVED — the mechanism to fetch and verify a
+revocation message signed by it is deferred to the future signed-policy-over-
+Freenet delivery layer (auto-update epic #4073). It must nevertheless be shipped
+now: a revocation can only be honored by peers whose binary already carries the
+key, so it cannot be retrofitted into already-released builds. The value is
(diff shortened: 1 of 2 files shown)


sanity:
Reviewed + merging. Bakes the offline backup/revocation ed25519 public key (c93f..3722) as a `#[used] static` so the bytes are actually present in the shipped binary — Codex caught (and the agent empirically verified) that a plain `const` referenced only from a test contributes ZERO bytes to the artifact (consts are inlined, no storage, no non-test use site), which would have silently defeated the whole 'bake it in now, can't retrofit' purpose. Pinned by `revocation_pubkey_matches_published_hex` (value + valid ed25519 point); AGENTS.md documents it + warns maintainers not to 'simplify' the static back to a const. Verification mechanism is intentionally deferred (rides the future signed-policy layer); only the key needs to ship now. Codex clean, 308 tests, CI green.

[AI-assisted - Claude]

PR #3329, opened 2026-02-28
fix: cap gateway backoff at 90s and poll ring connections during backoff wait

iduartgomez:
## Problem

When a node has only one configured gateway and it fails NAT traversal (UDP hole-punch), the gateway's per-address backoff escalates exponentially to 600s — a 10-minute blackout. During this window the node cannot bootstrap, cannot discover new peers, and never triggers the auto-update check.

Observed production logs:
```
Gateway connection failed, waiting before retry gateway=4ftT57... backoff_secs=599
All gateways in backoff, waiting for shortest backoff to expire wait_secs=419 total_gateways=1
```

The 600s max made sense for random ring peers (don't hammer unreachable strangers) but is wrong for a single configured gateway: the node is completely isolated rather than recovering gracefully.

## Approach

**Fix 1 — lower `PeerConnectionBackoff::DEFAULT_MAX_BACKOFF` from 600s to 90s.**

NAT traversal failures are transient (network change, temporary router congestion). The backoff progression with a 30s base becomes: 30s → 60s → 90s (capped). A 90s maximum gives the network time to stabilize without multi-minute isolation. The 10-minute cap was motivated by issue #2595 (prevent hammering random peers) but gateways are dedicated configured infrastructure, not random ring nodes.

**Fix 2 — cap the "all gateways in backoff" wait at 30s when `open_conns > 0`.**

`initial_join_procedure` sleeps for the full remaining backoff when all gateways are filtered out. Meanwhile the concurrent `connection_maintenance` task is already doing ring-based peer acquisition every 5s. If it brings `open_conns` above `BOOTSTRAP_THRESHOLD` during the sleep, `initial_join_procedure` won't notice until the full wait expires. Capping at 30s (when the node has at least one ring connection) lets the loop re-check frequently and exit the gateway-wait path sooner. The sleep remains interruptible via `gateway_backoff_cleared.notified()`.

## Why didn't CI catch this?

This is a configuration/timing regression: the 600s cap only manifests under repeated NAT failures with a single gateway, which no existing unit or integration test covers. The regression test added here (`test_default_max_backoff_is_90s_for_gateway_recovery`) would have caught any future increase of the cap.

## Testing

- Added `test_default_max_backoff_is_90s_for_gateway_recovery` in `peer_connection_backoff.rs`: drives 10 consecutive failures and asserts the remaining backoff never exceeds 90s.
- All existing `peer_connection_backoff` tests pass (11/11).
[...]

PR #3321, opened 2026-02-28
fix: make transport resilient to transient UDP send failures

sanity:
## Problem

A brief network blip (ENETUNREACH) on technic caused ALL 11 ring connections to die in <2 seconds. The root cause: every `socket.send_to()` failure in `peer_connection.rs` immediately returns `TransportError::ConnectionClosed`, which propagates up and kills the connection. The idle timeout (120s) should be the sole authority on connection liveness, not a single failed UDP send.

## Solution

**Two-layer fix:**

1. **Internal sends** (ACKs, pongs, resends) — log-and-continue, never propagate as errors from `recv()`
2. **Explicit sends** (`packet_sending()`) — return new `SendFailed` variant instead of `ConnectionClosed`; callers handle without killing connection

### Changes:
- **`transport/mod.rs`**: Add `SendFailed(SocketAddr, io::ErrorKind)` variant + `is_transient_send_failure()` helper
- **`peer_connection.rs`**: 6 send error sites updated:
  - Resend/TLP probe: `break` on error (RTO will retry)
  - AckConnection OK: log and return `Ok(None)` (peer retransmits)
  - Pong: log and return `Ok(())` (keepalive handles liveness)
  - Two noop() call sites: catch `SendFailed`, log, continue
  - Outbound stream completion: catch `SendFailed`, log, continue
  - `packet_sending()`: return `SendFailed` instead of `ConnectionClosed`
- **`p2p_protoc.rs`**: 3 error handler sites — check `is_transient_send_failure()` before closing connection
- **`connection_handler.rs`**: Add `SendFailed` to exhaustive match arms

### Not changed:
- Keepalive ping — already handles send errors gracefully
- Idle timeout — this IS the liveness authority, correct as-is
- Decryption failure threshold — security policy, correct as-is

## Testing

- New unit test `send_failure_returns_transient_error` verifies `packet_sending` returns `SendFailed` (not `ConnectionClosed`) on socket error, and that it's classified as transient
- All existing tests pass (`cargo test -p freenet`)
- `cargo fmt && cargo clippy --all-targets --all-features` clean

## Fixes

Closes #3318

[AI-assisted - Claude]

PR #4595, opened 2026-06-27
test: cover connection survival under transient send-failure burst (#3367)

iduartgomez:
## Problem

Gap 3 of #3367 (transport resilience) asks for tests that inject I/O
errors into the `peer_connection` send path. The #3321 production
failure mode was: a single UDP send failure cascaded into tearing down
all 11 of a peer's connections.

The transport module already has three transient-send-failure tests
(`send_failure_returns_transient_error`,
`transient_send_failure_then_recovery_succeeds`,
`send_failed_kinds_classify_as_transient`), but all three drive the
**stateless** `packet_sending` free function — they pin error
*classification* on that free function and never go through a real
`PeerConnection`'s `send`. There was no coverage that a live
`PeerConnection`'s public outbound entry point withstands a *burst* of
consecutive `io::Error`s, reports each as transient, stays usable, and
recovers.

## Solution

Add `connection_survives_burst_of_transient_send_failures` to the
`peer_connection` test module. It builds a full `RemoteConnection` from
the existing fixtures (mock `SharedMockTimeSource`, paired AES-128-GCM
ciphers, `SentPacketTracker`, congestion controller, token bucket,
`RollingRttStatsHandle`) but swaps the socket for the in-module
`FailableTestSocket`, whose sends fail on demand via an atomic flag.

The test drives the **public outbound entry point**
`PeerConnection::send` (`send` → `outbound_short_message` →
`packet_sending().await?`). That path *propagates* the `SendFailed`
error to the caller without consuming or poisoning the
`&mut PeerConnection`.

It asserts two things:
1. **Burst phase** — five consecutive sends fail; each returns an `Err`
   asserted transient (`is_transient_send_failure()`, explicitly NOT
   `ConnectionClosed`), and the *same* `&mut conn` is reused across the
   whole burst (proving it is neither consumed nor poisoned). No packet
   reaches the wire.
2. **Recovery phase** — the fail flag is flipped off, one more `send` on
   the same connection succeeds, and the packet is asserted to arrive on
   the socket's rx side addressed to the peer.

### Scope (what this test does and does not prove)

This test covers the **outbound `send` path only** — it pins that the
path classifies a send-failure burst as transient and recovers on the
same connection object. It does **not** execute the #3321 connection-
*teardown* decision. That decision lives in the `recv()` swallow guards
(`peer_connection.rs` ~1091/1186/1428, which `continue` rather than tear
down on a transient send error) and in the `p2p_protoc.rs` listener
(~3092-3108), neither of which this test runs. A recv-loop swallow-guard
test (site ~1091 is drivable in principle via the same
`recv()`-with-hand-built-packet machinery the existing
`recv_records_packet_metrics_post_authentication` test uses) is the
natural follow-up. The test doc-comment spells this scope out inline so
no future reader over-reads it.

A second follow-up commit on this PR (doc-honesty pass from review)
softens the doc-comment to match exactly this scope, corrects an
overstated "swallow sites aren't unit-drivable" claim, and documents the
implicit keepalive-task determinism dependency (the `SharedMockTimeSource`
keepalive task self-terminates on its first failing send before the
recovery phase, so it cannot race the rx assertions).

## Testing

- New test uses mock time + in-memory socket; fully deterministic, runs
  in ~0.01s.
- `cargo fmt` — clean.
- `cargo clippy -p freenet --lib -- -D warnings` — clean.
- `cargo test -p freenet --lib peer_connection` — 152 passed, 0 failed,
  1 ignored (the new test included and passing).

## Fixes

Adds outbound-send burst coverage for Gap 3 of #3367 (the #3321
classification/recovery contract on `PeerConnection::send`). The
recv-loop / listener teardown sites, the backoff-cap half of Gap 3
(#3329), and the Gap 4 soak test remain open, so this does NOT auto-close
the issue.

Refs #3367

### DIFF
--- a/crates/core/src/transport/peer_connection.rs
+++ b/crates/core/src/transport/peer_connection.rs
@@ -3655,33 +3655,54 @@ mod tests {
         crate::transport::TRANSPORT_METRICS.remove_peer(remote_addr);
     }
 
-    /// End-to-end regression for the #3321 failure mode (Gap 3 of #3367): a
-    /// *burst* of consecutive UDP send failures on a live `PeerConnection`
-    /// must NOT tear the connection down. The original #3321 bug had a single
-    /// send failure cascade into closing all 11 connections.
+    /// Outbound-`send` burst coverage related to #3321 (Gap 3 of #3367): a
+    /// *burst* of consecutive UDP send failures on a live `PeerConnection`'s
+    /// public `send` path must each surface as a *transient* error and must NOT
+    /// consume/poison the connection object, and once the failures clear the
+    /// same connection must send successfully again.
+    ///
+    /// SCOPE — read this before assuming more than the test proves. This test
+    /// exercises the *outbound* entry point only:
+    /// `send` -> `outbound_short_message` -> `packet_sending().await?`, which
+    /// PROPAGATES the `SendFailed` error to the caller. It pins that this path
+    /// classifies a send-failure burst as transient and recovers. It does NOT
+    /// execute the #3321 connection-*teardown* decision: that decision lives in
+    /// the `recv()` swallow guards (`peer_connection.rs` ~1091/1186/1428, which
+    /// `continue` rather than tear down on a transient send error) and in the
+    /// `p2p_protoc.rs` listener (~3092-3108). Driving those teardown sites under
+    /// a transient send-failure burst is a separate follow-up; this test does
+    /// not run them.
     ///
     /// The existing `send_failure_returns_transient_error` /
     /// `transient_send_failure_then_recovery_succeeds` /
     /// `send_failed_kinds_classify_as_transient` tests all drive the STATELESS
-    /// `packet_sending` free function — they pin error *classification* but not
-    /// that a real `PeerConnection`'s outbound entry point survives multiple
-    /// consecutive `io::Error`s and stays usable afterwards. This test closes
-    /// that gap by driving `PeerConnection::send` (the public outbound path:
-    /// `send` -> `outbound_short_message` -> `packet_sending().await?`) through
-    /// a `FailableTestSocket`.
+    /// `packet_sending` free function — they pin error *classification* on the
+    /// free function but never go through a real `PeerConnection`'s `send`. This
+    /// test adds that `PeerConnection::send` level of coverage via a
+    /// `FailableTestSocket`.
     ///
-    /// Survival is proven two ways:
+    /// What the test asserts:
     ///   1. Each failed send returns `Err` classified as transient
     ///      (`is_transient_send_failure()`, NOT `ConnectionClosed`), and the
     ///      `&mut PeerConnection` is neither consumed nor poisoned — we keep
     ///      calling `send` on the same object across the whole burst.
     ///   2. After flipping the failure flag off, the very same connection sends
-    ///      successfully and the packet arrives on the rx side — proving the
-    ///      connection was never torn down by the burst.
+    ///      successfully and the packet arrives on the rx side.
     ///
     /// If the transient-error contract regressed so that a send failure mapped
     /// to `ConnectionClosed` (or any non-transient variant), the per-send
     /// classification assertion below would fail before recovery is reached.
+    ///
+    /// NOTE on the swallow sites: the claim that the `recv()`-loop swallow
+    /// guards are not unit-drivable would be overstated. Site ~1091 (the
+    /// ACK-receipt `noop` send) IS drivable in principle — the
+    /// `recv_records_packet_metrics_post_authentication` test below already
+    /// builds a `PeerConnection` and drives `recv()` to completion with a
+    /// hand-crafted encrypted inbound packet, which is the same machinery a
+    /// recv-loop swallow-guard test would need. Site ~1428 is gated on the
+    /// `ack_check` interval timer, which never fires under this module's no-op
+    /// mock clock, so that one is genuinely not drivable here. A recv-loop
+    /// swallow-guard test for site ~1091 is the natural follow-up to this PR.
     #[tokio::test]
     async fn connection_survives_burst_of_transient_send_failures() {
         use crate::transport::crypto::TransportKeypair;
@@ -3756,8 +3777,9 @@ mod tests {
             );
             assert!(
                 !matches!(err, TransportError::ConnectionClosed(_)),
-                "burst send #{attempt} must NOT be ConnectionClosed — a single \
-                 send failure tearing down the connection is exactly the #3321 bug"
+                "burst send #{attempt} must NOT be ConnectionClosed — the \
+                 outbound send path must report a transient failure, not a \
+                 closed connection (#3321 classification contract)"
             );
         }
         // Nothing should have reached the wire during the burst.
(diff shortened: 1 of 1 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4595

### Summary
- **PR Title:** test: cover connection survival under transient send-failure burst (#3367)
- **Type:** test
- **CI Status:** passing (all checks green — Unit/Integration, Simulation, Windows, NAT, Fmt, Clippy)
- **Linked Issues:** #3367 (Gap 3), regression target #3321
- **Review tier:** Light (test-only, additive) — but reviewed with transport-aware depth
- **Reviewers run:** skeptical, testing, code-first (3 independent Claude lenses). **External pass (codex) was unavailable — token expired; gemini not installed.** Per multi-model-review policy, substituted 3 diverse-lens Claude reviewers and verified the contested findings against the actual code by hand.

---

### Code-First Analysis
**Independent Understanding:** The test builds a real `PeerConnection` over a `FailableTestSocket`, fires 5 consecutive `send()` calls with the failure flag on (each must return a transient `SendFailed`, never `ConnectionClosed`, on the same reused `&mut conn`), then flips the flag off and asserts one more `send()` succeeds and a packet reaches the rx side.

**Stated Intent:** "Directly targets the #3321 production failure mode, where a single UDP send failure tore down all 11 connections" / "completes the connection-survival half of Gap 3."

**Alignment:** Partially matches. The test genuinely exercises the **outbound `send` path** end-to-end (a real improvement over the 3 pre-existing tests, which only drive the stateless `packet_sending` free function). But the actual #3321 *teardown decision* — `if error.is_transient_send_failure() { continue } else { notify_transport_closed; return }` — lives in the connection listener at `p2p_protoc.rs:3092-3108` and in the recv-loop swallow guards at `peer_connection.rs:1091/1186/1428`. **This test executes none of those decision sites.** Verified by hand: `PeerConnection::send` takes `&mut self`, mutates only `last_packet_id` + congestion state, and has no teardown branch — so "survives the burst" on this path is partly type-system-guaranteed rather than behaviorally tested.

**Gaps:**
- The doc-comment / PR claim that the test "directly targets the #3321 failure mode" overstates what the assertions prove. It pins that the outbound entry point *classifies* a send-failure burst as transient and recovers — not that the connection-teardown logic survives it.
- The PR's justification "the recv-loop swallow sites aren't deterministically drivable from a unit test" is **factually overstated**: an existing test (`peer_connection.rs:3629-3630`) already constructs a `PeerConnection` and drives `conn.recv()` to completion with a hand-built encrypted inbound packet. Site `1091` (ACK-receipt `noop` send) is therefore drivable in principle. (Caveat, per the testing lens: site `1428` is timer-gated and genuinely *not* drivable under the no-op mock clock, and driving 1091 still needs the QueueFull/receipt trigger — non-trivial machinery.)

---

### Testing Assessment
**Coverage Level:** adequate for the stated `send`-path gap; the recv-loop teardown sites remain uncovered (low residual risk — see below).

| Unit | ✅ | New test is the first to drive a live, stateful `&mut PeerConnection` across a send-error burst; deterministic, 0.01s, verified green across 14+ runs |
| Integration | N/A | not the right level for this gap |
| Simulation | N/A | — |
| E2E | N/A | — |

**Regression Test:** present and valid for the `send` path. A regression mapping a send failure to `ConnectionClosed` (instead of transient `SendFailed`) would fail the per-send assert — confirmed against `is_transient_send_failure` (`transport.rs:384`, matches only `SendFailed`/`OutboundStreamFailed`).

**Missing Tests:** the three recv-loop swallow sites and the `p2p_protoc.rs:3092` listener decision — i.e. the sites where #3321's "11 connections die" symptom actually originated. Residual risk is **low** because those sites all reuse the `is_transient_send_failure()` predicate this PR now exercises end-to-end; a regression would have to change the swallow-arm *shape*, not the classification.

---

### Skeptical Findings
**Risk Level:** low (additive test-only; ran clean 14+ times)

| Doc-comment / PR overstate scope vs proven behavior | Should | `peer_connection.rs:3658-3687` (verified) | Test proves outbound classify+recover, not the #3321 teardown decision. Soften the claim. |
| "Loop sites not unit-drivable" justification inaccurate | Should | `peer_connection.rs:3629` (verified) | recv() IS unit-drivable; an existing test does it. Either add a recv-loop swallow test or correct the claim. |
| Implicit keepalive-task determinism dependency undocumented | Should (comment-only) | new test, recovery phase (verified) | `SharedMockTimeSource` defaults `supports_keepalive()=true`, so `PeerConnection::new` spawns a live keepalive task that busy-sends into the failing socket and self-terminates on its first error before recovery. Safe (it can't race `sock_rx`), but the determinism rests on this implicit behavior. Add a one-line comment so a future keepalive change can't silently reintroduce a race. |
| Recovery assert proves "a packet to peer left", not "the payload left" | Consider | new test recovery assert (verified) | asserts only `target == remote_addr`; a racing keepalive Ping is also to `remote_addr`. Could decode `_bytes` to assert the ShortMessage specifically. Minor. |

No high/medium bugs. No race in the fail-flag (single task, program-order happens-before). `try_recv()` does not race the send (same-task, send completes first).

---

### Big Picture Assessment
**Goal Alignment:** partial — advances Gap 3 (outbound path) but does not close the teardown-survival half it claims to.
**Anti-Patterns Detected:** none (no removed/ignored tests, no weakened assertions).
**Removed Code Concerns:** none — purely additive (+127 / -0, one file).
**Scope Assessment:** focused. Correctly uses `Refs #3367` (not `Fixes`) since #3367 has other open gaps.

---

### Documentation
- Code docs: the new test's doc-comment is thorough but overstates scope (see Should-fix #1).
- Architecture docs: n/a.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
None. The test is correct, deterministic, and mergeable as-is.

#### Should Fix (Important)
1. **Soften the doc-comment + PR description** to match what is proven: this test pins that the outbound `send` entry point classifies a send-failure burst as transient and recovers — NOT that the #3321 connection-teardown logic survives it. The teardown decision (`p2p_protoc.rs:3092`, swallow sites `1091/1186/1428`) is out of scope; name them as the residual gap rather than implying Gap 3's survival half is closed.
2. **Correct the "not unit-drivable" claim** (or act on it): `conn.recv()` is unit-drivable (`peer_connection.rs:3629` does it today). Either add a recv-loop swallow-guard test for site `1091`, or state explicitly that the recv-loop/listener teardown sites remain uncovered and why (site `1428` is timer-gated under mock time).
3. **Document the keepalive-task determinism dependency** with a one-line comment in the test.

#### Consider (Suggestions)
1. Strengthen the recovery assertion to decode the sent bytes (prove *the payload* left, not just *a* packet to the peer).
2. Alternating fail/succeed or a longer burst — low value, since `send` accumulates no per-send teardown state.

---

### Verdict
**HEAD SHA reviewed:** fad61c13bd7af9d20962f932104b5b08dd6a5e32

The test is sound, deterministic, and a real coverage improvement — no blocking issues. The Should-fixes are doc/comment honesty (items 1–2) plus a one-line determinism comment (item 3); none change test logic, so a diff-of-the-diff re-check is sufficient. The substantive point for the author: this closes the *outbound-send* slice of Gap 3, not the teardown-decision slice that actually caused #3321 — the description should say so, and ideally a follow-up should cover the recv-loop/listener sites (which are more drivable than the PR claims).

[AI-assisted - Claude]

PR #4597, opened 2026-06-27
fix(put): stream-inactivity timeout for streaming PUT retry loop (#4001)

### LINKED ISSUE #4001: fix(put): drive_retry_loop fires retries while streaming PUT still in flight, causing version-conflict failures for large payloads

iduartgomez:
## Problem

`drive_retry_loop` gates each PUT attempt on a fixed per-attempt timeout. For streaming PUTs whose network time exceeds that timeout (e.g. the freenet.org ~2.4 MB upload), the retry loop fires a retry **while the original streaming op is still in flight**. The retry reuses the same contract version and fast-fails with a version conflict ("New state version N must be higher than current N"), so the client gets "put failed after N attempts" even though the original PUT actually completes seconds later.

The already-merged heuristic (`streaming_aware_attempt_timeout`, option b) only *relocates* the cliff by scaling the deadline with payload size. This PR implements the true fix (option c): a stream-inactivity timeout driven by real per-fragment liveness. The heuristic is kept as the fallback and 600 s hard ceiling.

## Solution

A new `StreamProgressHandle` (neutral `AtomicU64` + `Notify`, no transport/op dependencies) records a tick per dispatched fragment. The retry loop, when a streaming driver is in play, replaces the fixed `tokio::time::timeout` with a three-arm `select!`:

1. **Terminal reply** — the awaited `send_and_await` round-trip (unchanged semantics).
2. **Inactivity loop** — woken promptly by `Notify`; on each `STREAM_OP_INACTIVITY_TIMEOUT` (30 s) sleep expiry it re-reads the `AtomicU64` and only declares a stall if **no fragment landed in the race window**. Using both `Notify` and the atomic is deliberate: `Notify` alone is racy (lost-wakeup if a fragment lands in the wakeup window), polling alone is wasteful — the atomic re-check is the tiebreak that closes the gap.
3. **Hard ceiling** — a sibling `STREAMING_ATTEMPT_TIMEOUT_CAP` (600 s) sleep so a stream that dribbles fragments forever still can't hold the driver hostage.

All sleeps go through the driver's `TimeSource` (carried in the `StreamProgress` bundle as cloneable closures, since `TimeSource: Clone` is not `dyn`-safe) so the path is VirtualTime-correct under DST.

### Why a per-`Transaction` registry (not gratuitous coupling)

A client streaming PUT runs as **two concurrent tasks on the originator node that share only the `Transaction` id**:

- **Task A** — `start_client_put` → `drive_retry_loop`, which emits a `PutMsg::Request` via `send_and_await` and owns the per-attempt timeout decision.
- **Task B** — the `PutMsg::Request` loops back through the local event loop (`source_addr=None` → `upstream_addr=own_addr`) and is dispatched to a *separate* spawned `drive_relay_put` task, which actually fragments and streams the payload via `conn_manager.send_stream(...)` → `outbound_stream::send_stream`.

The per-fragment send (`sent_so_far += …`) lives in Task B's downstream transport, while the timeout decision lives in Task A. There is **no shared per-tx object** to thread a handle through directly, so a `StreamProgressRegistry` (`DashMap<Transaction, StreamProgressHandle>`) on `OpManager` bridges them — a *scoped* lookup keyed by `Transaction`, mirroring the existing `pending_op_results` / `OrphanStreamRegistry` patterns, **not** a free-floating module-global. Task A inserts the handle before sending and removes it **unconditionally on every loop-exit path** (success, stall, ceiling, error) so it can't leak; Task B looks it up by `Transaction` and threads it into the transport, which records a tick per fragment. A non-streaming PUT / non-originator relay / op with no registered handle is a no-op on the lookup side.

The `RetryDriver::stream_progress()` hook defaults to `None`, so GET / SUBSCRIBE / non-streaming PUT compile unchanged and take the byte-identical fixed-deadline path.

## Testing

`cargo fmt && cargo clippy -p freenet --lib -- -D warnings && cargo test -p freenet --lib operations` — all clean (443 operations tests pass). Transport (138) and p2p_protoc (49) suites pass; the existing `streaming_aware_attempt_timeout` heuristic tests (10) still pass.

New regression tests against `await_streaming_attempt` (the streaming branch core) under tokio's `start_paused` virtual clock:

- `streaming_put_does_not_retry_while_chunks_flow` — fragments every 10 s for 120 s (past the 60 s cliff), reply at 125 s; asserts the terminal reply is delivered (no timeout). **Verified to FAIL** when the streaming branch is reverted to the fixed 60 s deadline.
- `streaming_put_retries_on_true_stall` — progress to 40 s then silence; asserts the 30 s inactivity timeout fires ~70 s in.
- `streaming_put_hard_ceiling_fires` — records forever; asserts bail at the 600 s ceiling.
- `streaming_put_progress_during_inactivity_race` — fragments land just inside each window; asserts the atomic tiebreak prevents a false stall.
- `non_streaming_put_uses_fixed_timeout` — pins that the default `stream_progress()` is `None` and the `None` arm remains the unchanged `tokio::time::timeout` path.

Plus unit tests on `StreamProgressHandle` / `StreamProgressRegistry` (record/since_last/saturation, notified permit, leak-free insert/get/remove roundtrip).

The two source-scrape pin tests on `drive_retry_loop` (`drive_retry_loop_terminal_arm_does_not_call_advance`, `drive_retry_loop_has_fast_infra_retry_path`) still pass — the change kept their scraped substrings intact. The `drive_client_put_inner_dispatches_streaming_cap_on_should_use_streaming` pin was updated to follow the hoisted `is_streaming` flag (the streaming-cap selection and the new stream-progress handle now derive from one decision).

## Scope

**PUT only.** The additive `RetryDriver::stream_progress()` default-`None` hook is the seam a later GET/SUBSCRIBE PR plugs into — streaming GET is a deferred follow-up. The merged `streaming_aware_attempt_timeout` heuristic is retained as the fallback/ceiling.

## Fixes #4001

### DIFF
--- a/crates/core/src/node/network_bridge.rs
+++ b/crates/core/src/node/network_bridge.rs
@@ -61,6 +61,23 @@ pub(crate) trait NetworkBridge: Send + Sync {
         metadata: Option<bytes::Bytes>,
     ) -> impl Future<Output = ConnResult<()>> + Send;
 
+    /// Like [`Self::send_stream`] but attaches an optional stream-progress
+    /// liveness handle (#4001) so the transport records a tick per dispatched
+    /// fragment. Used ONLY by the originator-loopback PUT relay so the
+    /// retry-loop task can apply a true stream-inactivity timeout. The default
+    /// delegates to [`Self::send_stream`] (handle dropped), keeping mock and
+    /// non-streaming bridges unchanged.
+    fn send_stream_with_progress(
+        &self,
+        target_addr: SocketAddr,
+        stream_id: crate::transport::peer_connection::StreamId,
+        data: bytes::Bytes,
+        metadata: Option<bytes::Bytes>,
+        _progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
+    ) -> impl Future<Output = ConnResult<()>> + Send {
+        self.send_stream(target_addr, stream_id, data, metadata)
+    }
+
     /// Pipe an inbound stream to a peer, forwarding fragments as they arrive.
     ///
     /// Unlike `send_stream` which takes complete data and fragments it, this forwards
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -721,6 +721,11 @@ pub(super) async fn broadcast_to_single_peer(
             // when the actual stream transfer finishes.
             let (completion_tx, completion_rx) = tokio::sync::oneshot::channel();
 
+            // channel-safety: ok — broadcast_to_single_peer runs on the detached
+            // broadcast-queue task, not the event loop; the StreamSend this
+            // enqueues is drained by the loop, so it cannot self-stall it. The
+            // #4001 `None` progress arg that brought this statement into the diff
+            // does not change the send path.
             if let Err(err) = bridge
                 .send_stream_with_completion(
                     peer_addr,
@@ -728,6 +733,7 @@ pub(super) async fn broadcast_to_single_peer(
                     bytes::Bytes::from(payload_bytes),
                     metadata,
                     Some(completion_tx),
+                    None,
                 )
                 .await
             {
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -170,6 +170,10 @@ pub(crate) enum P2pBridgeEvent {
         /// carrying a [`BroadcastDeliveryOutcome`] so the queue distinguishes a
         /// real delivery from a drop (#4235).
         completion_tx: Option<tokio::sync::oneshot::Sender<BroadcastDeliveryOutcome>>,
+        /// Optional per-fragment progress handle for the streaming-PUT retry
+        /// loop (#4001). Threaded through to `send_stream` so the originator's
+        /// retry loop sees liveness ticks. `None` for every non-originator path.
+        progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
     },
     /// Pipe an inbound stream to a target, forwarding fragments incrementally.
     PipeStream {
@@ -283,7 +287,13 @@ impl P2pBridge {
         data: bytes::Bytes,
         metadata: Option<bytes::Bytes>,
         completion_tx: Option<tokio::sync::oneshot::Sender<BroadcastDeliveryOutcome>>,
+        progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
     ) -> super::ConnResult<()> {
+        // channel-safety: ok — sole caller is the broadcast-queue task
+        // (broadcast_to_single_peer), a detached spawned task, not the event
+        // loop; the StreamSend it enqueues is drained by the loop, so it cannot
+        // self-stall it. Grandfathered enqueue; the #4001 `progress` field that
+        // brought this statement into the diff does not change the send path.
         self.ev_listener_tx
             .send(P2pBridgeEvent::StreamSend {
                 target_addr,
@@ -291,6 +301,7 @@ impl P2pBridge {
                 data,
                 metadata,
                 completion_tx,
+                progress,
             })
             .await
             .map_err(|_| ConnectionError::SendNotCompleted(target_addr))?;
@@ -444,6 +455,10 @@ impl NetworkBridge for P2pBridge {
         data: bytes::Bytes,
         metadata: Option<bytes::Bytes>,
     ) -> super::ConnResult<()> {
+        // channel-safety: ok — grandfathered streaming enqueue; callers are
+        // off-loop spawned op/relay tasks, not the event loop (which drains
+        // StreamSend). The #4001 `progress: None` field that brought this
+        // statement into the diff does not change the send path.
         self.ev_listener_tx
             .send(P2pBridgeEvent::StreamSend {
                 target_addr,
@@ -451,6 +466,37 @@ impl NetworkBridge for P2pBridge {
                 data,
                 metadata,
                 completion_tx: None,
+                progress: None,
+            })
+            .await
+            .map_err(|_| ConnectionError::SendNotCompleted(target_addr))?;
+        Ok(())
+    }
+
+    async fn send_stream_with_progress(
+        &self,
+        target_addr: SocketAddr,
+        stream_id: StreamId,
+        data: bytes::Bytes,
+        metadata: Option<bytes::Bytes>,
+        progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
+    ) -> super::ConnResult<()> {
+        // Enqueues the streaming send carrying the #4001 progress handle so
+        // `outbound_stream::send_stream` records a tick per fragment.
+        // channel-safety: ok — sole caller is the originator-loopback PUT relay
+        // (drive_relay_put), which runs on a spawned op task, not the event
+        // loop; the StreamSend it enqueues is drained by the loop, so it cannot
+        // self-stall it. Identical send path to the grandfathered `send_stream`
+        // above; completion_tx stays None (no broadcast-queue concurrency on
+        // the originator PUT path).
+        self.ev_listener_tx
+            .send(P2pBridgeEvent::StreamSend {
+                target_addr,
+                stream_id,
+                data,
+                metadata,
+                completion_tx: None,
+                progress,
             })
             .await
             .map_err(|_| ConnectionError::SendNotCompleted(target_addr))?;
@@ -1686,6 +1732,7 @@ impl P2pConnManager {
                             data,
                             metadata,
                             completion_tx,
+                            progress,
                         } => {
                             // The event loop must never block indefinitely on a single
                             // congested peer. Reserve a slot with the same 500 ms
@@ -1714,6 +1761,7 @@ impl P2pConnManager {
                                             data,
                                             metadata,
                                             completion_tx,
+                                            progress,
                                         }));
                                     }
                                     Ok(Err(_closed)) => {
@@ -2770,6 +2818,9 @@ pub(super) enum ConnEvent {
         /// Optional completion signal for broadcast queue concurrency control,
         /// carrying a [`BroadcastDeliveryOutcome`] (#4235).
         completion_tx: Option<tokio::sync::oneshot::Sender<BroadcastDeliveryOutcome>>,
+        /// Optional per-fragment progress handle for the streaming-PUT retry
+        /// loop (#4001). `None` for every non-originator path.
+        progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
     },
     /// Pipe an inbound stream to a peer, forwarding fragments as they arrive.
     /// Used for low-latency forwarding at intermediate nodes.
@@ -2869,6 +2920,7 @@ async fn handle_peer_channel_message(
                     data,
                     metadata,
                     completion_tx,
+                    progress,
                     ..
                 } => {
                     tracing::debug!(
@@ -2879,7 +2931,7 @@ async fn handle_peer_channel_message(
                         "[CONN_LIFECYCLE] Sending operations stream data to peer"
                     );
                     if let Err(error) = conn
-                        .send_stream_data(stream_id, data, metadata, completion_tx)
+                        .send_stream_data(stream_id, data, metadata, completion_tx, progress)
                         .await
                     {
                         tracing::error!(
@@ -2905,7 +2957,7 @@ async fn handle_peer_channel_message(
                         "[CONN_LIFECYCLE] Piping stream to peer"
                     );
                     if let Err(error) = conn
-                        .pipe_stream_data(outbound_stream_id, inbound_handle, metadata)
+                        .pipe_stream_data(outbound_stream_id, inbound_handle, metadata, None)
                         .await
                     {
                         tracing::error!(
--- a/crates/core/src/node/network_bridge/p2p_protoc/dispatch.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc/dispatch.rs
@@ -347,6 +347,7 @@ impl P2pConnManager {
                 data,
                 metadata,
                 completion_tx,
+                progress,
             }) => EventResult::Event(
                 ConnEvent::StreamSend {
                     target_addr,
@@ -354,6 +355,7 @@ impl P2pConnManager {
                     data,
                     metadata,
                     completion_tx,
+                    progress,
                 }
                 .into(),
             ),
--- a/crates/core/src/node/network_bridge/priority_select/tests.rs
+++ b/crates/core/src/node/network_bridge/priority_select/tests.rs
@@ -2825,6 +2825,7 @@ impl crate::transport::PeerConnectionApi for MockPeerConnection {
         _completion_tx: Option<
             tokio::sync::oneshot::Sender<crate::transport::BroadcastDeliveryOutcome>,
         >,
+        _progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
     ) -> std::pin::Pin<
         Box<
             dyn std::future::Future<Output = Result<(), crate::transport::TransportError>>
@@ -2845,6 +2846,7 @@ impl crate::transport::PeerConnectionApi for MockPeerConnection {
         _outbound_stream_id: crate::transport::peer_connection::StreamId,
         _inbound_handle: crate::transport::peer_connection::streaming::StreamHandle,
         _metadata: Option<bytes::Bytes>,
+        _progress: Option<crate::operations::stream_progress::StreamProgressHandle>,
     ) -> std::pin::Pin<
         Box<
             dyn std::future::Future<Output = Result<(), crate::transport::TransportError>>
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -31,6 +31,7 @@ use crate::{
     message::{InterestMessage, MessageStats, NetMessage, NetMessageV1, NodeEvent, Transaction},
     operations::{
         OpCtx, OpError, connect::ConnectForwardEstimator, orphan_streams::OrphanStreamRegistry,
+        stream_progress::StreamProgressRegistry,
     },
     ring::{
         ConnectionManager, LiveTransactionTracker, PeerConnectionBackoff, PeerKey, PeerKeyLocation,
@@ -132,6 +133,16 @@ pub(crate) struct OpManager {
     /// Coordinates transport layer (which receives fragments) with operations layer
     /// (which receives RequestStreaming/ResponseStreaming messages).
     orphan_stream_registry: Arc<OrphanStreamRegistry>,
+    /// Per-`Transaction` registry of streaming-PUT progress handles (#4001).
+    ///
+    /// A client streaming PUT runs the retry-loop task and the originator-
+    /// loopback relay-streaming task separately, sharing only the
+    /// `Transaction` id. The retry loop inserts a `StreamProgressHandle` here
+    /// before sending and removes it on exit; the loopback relay looks it up
+    /// and records per-fragment progress so the retry loop can use a true
+    /// stream-inactivity timeout instead of a fixed per-attempt deadline. See
+    /// `operations::stream_progress`.
+    stream_progress_registry: Arc<StreamProgressRegistry>,
     /// Size threshold in bytes above which streaming is used.
     pub streaming_threshold: usize,
     /// Backoff tracker for failed gateway connection attempts.
@@ -232,6 +243,7 @@ impl Clone for OpManager {
             pending_broadcasts: self.pending_broadcasts.clone(),
             request_router: self.request_router.clone(),
             orphan_stream_registry: self.orphan_stream_registry.clone(),
+            stream_progress_registry: self.stream_progress_registry.clone(),
             streaming_threshold: self.streaming_threshold,
             gateway_backoff: self.gateway_backoff.clone(),
             gateway_backoff_cleared: self.gateway_backoff_cleared.clone(),
@@ -418,6 +430,7 @@ impl OpManager {
             ),
             request_router,
             orphan_stream_registry,
+            stream_progress_registry: Arc::new(StreamProgressRegistry::new()),
             streaming_threshold,
             gateway_backoff: Arc::new(Mutex::new(PeerConnectionBackoff::new())),
             gateway_backoff_cleared: Arc::new(tokio::sync::Notify::new()),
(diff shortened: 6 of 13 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4597

### Summary
- **PR Title:** fix(put): stream-inactivity timeout for streaming PUT retry loop (#4001)
- **Type:** fix (concurrency / transport / streaming)
- **CI Status:** passing (lint/clippy/fmt green; unit/sim pending) — **but see the Must-fix: CI is green and the headline bug is still present**, because the unit tests construct both clocks in the one ordering that masks it.
- **Linked Issues:** Fixes #4001
- **Review tier:** Full (concurrency/async + transport + new cross-task registry)
- **Reviewers run:** skeptical, code-first, testing, big-picture (4 Claude lenses). **External pass (codex) unavailable — token expired; gemini not installed.** Per multi-model-review policy, ran 4 diverse-lens Claude reviewers and **hand-verified the blocking finding** against the source.

---

### Code-First Analysis
**Independent Understanding:** Adds a per-fragment liveness signal so a streaming PUT stops self-retrying while its own stream is in flight. `StreamProgressHandle` (`AtomicU64` last-progress-millis + `Notify`) is published into a `Transaction`-keyed `StreamProgressRegistry` on `OpManager` by the retry-loop task (Task A) before sending; the loopback streaming task (Task B) looks it up and calls `record()` per fragment in `outbound_stream::send_stream`. `drive_retry_loop`'s streaming branch (`await_streaming_attempt`) replaces the fixed `tokio::time::timeout` with a 3-arm select: terminal reply / 30s inactivity (atomic tiebreak) / 600s hard ceiling.

**Stated Intent:** Implement #4001's preferred option (c) — observe real liveness instead of guessing a wall-clock deadline. PUT-only; keep the merged heuristic + 600s ceiling as fallback; GET/SUBSCRIBE untouched via a default-`None` `RetryDriver::stream_progress()` hook.

**Alignment:** The architecture matches intent and the PR description is honest about the registry deviation (the 2-task topology makes the issue's "thread alongside completion_tx" sketch impossible — verified: Task B is `GlobalExecutor::spawn`ed). The wiring is structurally complete end-to-end. **However, the central mechanism does not actually work in production** — see Must-fix #1.

---

### Skeptical Findings
**Risk Level:** medium-high — one blocking correctness defect (clock epochs), one latent cleanup defect.

| **Cross-epoch `RealTime` mismatch defeats the 30s inactivity detection** | **MUST** | `op_ctx_task.rs:373` (reader epoch) vs `outbound_stream.rs:392` + `peer_connection.rs:1982` (writer epoch); `simulation/time.rs:152-173`; `stream_progress.rs:101` | **Hand-verified.** `RealTime::now()` is **epoch-relative** (`self.epoch.elapsed()`, per-instance epoch at `RealTime::new()`). The reader builds a fresh `RealTime::new()` at op-start (epoch E1); the writer records via the **connection's** `time_source` (epoch E2 = connection establishment, seconds-to-minutes *earlier*). So `record()` stores `now−E2` (larger) while the reader computes `since_last = (now−E1).saturating_sub(now−E2)` → **negative → saturates to 0**. After ≥1 fragment, the 30s stall **can never fire**; a genuinely dead stream is only caught by the 600s ceiling. The fix's fast-detection — its entire point — is defeated for the common case (≥1 fragment flowed). Silent: pre-first-fragment the handle still holds its E1 seed, so very-early stalls work. **Unit tests miss it by construction** (both `RealTime`s created near-simultaneously in the safe-skew order). Fix: reader and writer must share one time basis — record an absolute `SystemTime`/`UNIX_EPOCH` millis on both sides, OR thread the originator's `StreamProgress` clock down to the transport record site, OR have `record()` capture the reader's clock. |
| Registry removal is explicit, not RAII — not leak-proof under panic/cancel; doc claim is false | **Should** | `op_ctx.rs:628-635` (verified) | `insert` → `await_streaming_attempt().await` → `remove` is a plain statement, not a Drop guard. Covers all *normal* returns (no `?`/early-return/panic inside the await — verified), but a dropped/cancelled `drive_retry_loop` future or a panic skips `remove`. Registry has **no TTL/sweep**, so a leaked entry is permanent for process life. The doc at `op_ctx.rs:627` ("removed unconditionally … cannot leak") and `stream_progress.rs:189` ("cannot leak") are **literally false** under unwind/cancel. The PR's own neighbor `RelayPutInflightGuard` (op_ctx_task.rs:879) already uses RAII for exactly this — mirror it. In practice low-probability today (detached spawn, JoinHandle dropped, no abort in normal op), so latent, not active. |
| Streaming-decision divergence: originator pre-merge estimate vs relay merged re-check | Should | `op_ctx_task.rs:342-349` vs `:1385` (per code-first) | Originator registers a handle from a pre-merge payload estimate; the relay re-checks `should_use_streaming` on the merged payload. If estimate ≥ threshold but merged < threshold, Task B forwards a non-streaming request and records ZERO fragments → handle keeps its seed → a >30s non-streaming forward fires a false stall. Rarer, bounded by the fast-reply arm, but re-introduces a version of the bug. Add a consistency guard or comment. |
| Dropped initial `StreamSend` → 30s stall instead of immediate failure | Consider | `p2p_protoc.rs:1780-1793` (per skeptical) | Originator-loopback `completion_tx=None`; a backpressure-dropped whole-stream send records nothing → waits the full 30s window → exhausts (0 advancements). By-design latency change; confirm 30s is acceptable for the website case. |

Verified safe (skeptical + me): lost-wakeup race is genuinely closed by the atomic re-check *given a shared epoch*; the select is fair (terminal-reply arm pinned outside the loop, not starved); insert-before-emit ordering correct; `record()` is non-blocking (relaxed store + `notify_one`, one DashMap lookup per stream not per fragment); registry is `Arc`-shared across OpManager clones.

---

### Testing Assessment
**Coverage Level:** the inactivity *primitive* is well-tested in isolation; the end-to-end reader/writer clock crossing — where the Must-bug lives — is **not** tested.

| Unit | ⚠️ | 5 regression tests on `await_streaming_attempt` + 4 on the handle/registry. Deterministic (`start_paused` + `RealTime`), fast (virtual time). `streaming_put_does_not_retry_while_chunks_flow` is a faithful reproduction and credibly fails against the pre-fix deadline. |
| Integration | ❌ | **No test crosses the real reader/writer `RealTime` epochs** — this is exactly why the Must-bug slipped through. An integration test driving a real/in-memory bridge with progress would have caught the saturate-to-zero. |
| Simulation | ❌ | Streaming-progress path is not exercised under DST: production always builds `StreamProgress` from `RealTime` (doesn't advance under VirtualTime), and `MockNetworkBridge` inherits the default `send_stream_with_progress` that **drops** the handle. |

**Regression Test:** present and genuine for the *primitive*, but does not guard the production clock-crossing.
**Registry-lifecycle test gap (Should):** every streaming test calls `await_streaming_attempt` directly, bypassing the `insert`→`remove` pairing in `drive_retry_loop`. The unconditional `remove` (the only leak guard, no TTL fallback) is untested on the exhaustion/error path. Add a structural pin (cheap) or a `registry.len()==0`-after-exit behavioural test on the error path.

---

### Big Picture Assessment
**Goal Alignment:** addresses #4001's root cause head-on *by design* — but the clock defect means it does not yet deliver the fast detection in production.
**Anti-Patterns Detected:** **"green CI, real bug"** — the suite passes while the headline mechanism is defeated, precisely because the only tests construct both clocks in the masking order. (No removed/ignored tests; no weakened assertions.)
**Removed Code Concerns:** none — 16 deleted lines are signature/call-site edits; `streaming_aware_attempt_timeout` + the 600s `STREAMING_ATTEMPT_TIMEOUT_CAP` retained as fallback/ceiling (verified).
**Scope Assessment:** focused. 837 additions traced as load-bearing — new logic + mechanical `Option<StreamProgressHandle>` plumbing through the transport trait chain. No unrelated cleanup. One `is_streaming` hoist, in service of the fix. Channel-safety annotations on the 4 send sites **verified honest** (each send is genuinely off the event loop; underlying path byte-identical to grandfathered `send_stream`).

---

### Documentation
- `stream_progress.rs` module doc claims VirtualTime-correctness and "cannot leak"; both are overstated (the reader/writer epoch requirement is unstated; cancel/panic can leak). Correct the docs alongside the fixes.
- `since_last`'s "saturates to zero = the tiebreak we rely on" comment hides the Must-bug: saturation is meant to absorb a sub-ms race, but cross-epoch skew makes it absorb the entire interval. Add the shared-epoch caveat.
- Add `send_stream_with_progress` to the watched-wrapper list in `.claude/rules/channel-safety.md` (doc hygiene).

---

### Recommendations

#### Must Fix (Blocking)
1. **Make the reader and writer share one time basis.** As written, the retry loop's `RealTime` (epoch E1, op-start) and the transport's connection `RealTime` (epoch E2, connection establishment) are different instances and `RealTime::now()` is epoch-relative, so `since_last` saturates to 0 after the first fragment and the 30s inactivity detection never fires (degrades to the 600s ceiling). Record an absolute timestamp on both sides (e.g. `SystemTime`→`UNIX_EPOCH` millis), or thread the originator's `StreamProgress` clock to the record site, or have `record()` capture the reader's clock. Add an integration/DST test that crosses the two clocks so this can't silently regress again. **Verified by hand against `simulation/time.rs:152-173`, `op_ctx_task.rs:373`, `outbound_stream.rs:392`, `stream_progress.rs:101`.**

#### Should Fix (Important)
1. **RAII the registry removal.** Replace the explicit `remove` with a Drop guard (mirroring `RelayPutInflightGuard`) so the "cannot leak" doc claim is literally true under cancel/panic, and correct the docs. Add a registry-lifecycle test on the error/exhaustion path (the current tests bypass insert/remove).
2. **Guard the streaming-decision divergence** (originator pre-merge estimate vs relay merged re-check): if the relay forwards non-streaming, nothing records and a >30s forward false-stalls. Add a consistency check or comment.

#### Consider (Suggestions)
1. Confirm the 30s detection latency for a dropped initial `StreamSend` is acceptable, or detect a never-recorded handle faster.
2. Doc: note that the originator records in `send_stream` (not `pipe_stream`, which stays `None`) for the PUT-only scope; note that mock/DST bridges drop the progress handle; add `send_stream_with_progress` to channel-safety.md's wrapper list; note the ceiling intentionally reuses the merged 600s cap (not the issue's "~5min").

---

### Verdict
**HEAD SHA reviewed:** e4b2aaad60e2547484f47b8ac636eacbbdc54df2

The architecture is sound, the scope is clean, the channel-safety annotations are honest, and the concurrency primitives (lost-wakeup tiebreak, fair select, non-blocking record, insert-before-emit) are all correct **given a shared clock epoch** — but they don't have one. The reader and writer use two different epoch-relative `RealTime` instances, so the headline 30s inactivity detection saturates to zero in production after the first fragment and silently degrades to the 600s ceiling. That guts the fix's purpose and is invisible to the current tests. Must-fix #1 blocks merge; because it is a real correctness change in a concurrency path, a **full re-review on the updated HEAD is required** after it (and ideally the RAII leak Should-fix) land.

Strong work overall — the design and plumbing are right; the bug is a subtle cross-instance clock-basis mismatch that only a code-first reading of both record sites surfaces.

[AI-assisted - Claude]

iduartgomez (commented):
## Comprehensive PR Review (Round 2 — Re-review after fixes): #4597

### Summary
- **PR Title:** fix(put): stream-inactivity timeout for streaming PUT retry loop (#4001)
- **Type:** fix (concurrency / transport / streaming)
- **CI Status:** Fmt / Clippy / Rule Lint / Conventional / E2E green on the new HEAD; Unit & Integration / Simulation / Windows running (reviewers ran the affected unit tests locally — all pass).
- **Linked Issues:** Fixes #4001
- **Review tier:** Full (re-review of a correctness change in a concurrency path — mandatory after the round-1 Must-fix)
- **Reviewers run:** skeptical, code-first, testing (3 diverse Claude lenses; big-picture skipped — scope/alignment unchanged, only the clock mechanism + cleanup guard changed). **External pass (codex) still unavailable — token expired.** Both blocking findings hand-verified against the new HEAD.

This is a **re-review after fixes** for the two round-1 findings:
- **Must (blocking):** cross-epoch `RealTime` mismatch defeated the 30s inactivity detection (reader and writer used different `RealTime` instances → `since_last` saturated to 0 → stall never fired, silently degrading to the 600s ceiling).
- **Should:** registry removal was explicit (not RAII) → could leak on cancel/panic; doc claimed "cannot leak" (false).

---

### Resolution of Round-1 Findings

**Must #1 (clock epoch) — RESOLVED. Verified structurally + by all three lenses + by hand.**
The fix makes the handle OWN its clock: `record(&self)` takes no timestamp and reads `(self.now_millis)()` (stream_progress.rs:133); `since_last(&self)` reads the same closure (:143); writer call sites (outbound_stream.rs:396/780) call `p.record()` with no arg. There is now exactly ONE `RealTime` per streaming PUT (constructed once at op_ctx_task.rs:377), captured into an `Arc<dyn Fn() -> u64>`. The registry hands the writer a `.clone()` of the reader's handle (stream_progress.rs:238), which shares that same `Arc` closure — so the writer's `record()` writes through the reader's clock. The cross-epoch mismatch is **unrepresentable in the type and data flow**, not merely masked. (`RealTime::now` is `epoch.elapsed()` per-instance — with only one instance, no disagreement is possible.) The code-first lens that found the bug independently re-traced it and cleared the blocker.

**Should #2 (RAII leak) — RESOLVED.**
New `StreamProgressGuard` (insert in `new()`, remove in `Drop`, stream_progress.rs:269-283) replaces the explicit insert/remove pair. Verified: it holds `Arc<StreamProgressRegistry>` (Drop reaches the registry with no lifetime issue); it's bound as a named `_progress_guard` so it lives across the await (op_ctx.rs:634-639), covering cancellation (future dropped mid-await) and panic; it removes the correct key (`attempt_tx`); no double-remove across retry iterations (prior guard drops before the next is created; DashMap remove is idempotent anyway). The "cannot leak" docs are now true.

---

### Skeptical Findings (round 2)
**Risk Level:** low — both fixes correct, no new bugs.

| Clock bug recurrence | — | Resolved — single shared clock, structurally impossible to reintroduce without re-adding a `record()` timestamp param |
| RAII guard Drop-ordering / double-remove | — | Verified safe — correct key, held across await, idempotent remove, no iteration overlap |
| Lost-wakeup tiebreak | — | Still holds; with one monotonic clock `now >= last` always, so `saturating_sub` only saturates in the legitimate race window (no false suppression) |
| Fair select / insert-before-emit / non-streaming path | — | Unchanged, re-verified |

No High/Medium/Low severity issues found in round 2.

---

### Testing Assessment (round 2)
Both round-1 test gaps are faithfully closed:

| Test | Guards | Verified |
|------|--------|----------|
| `writer_and_reader_share_one_clock_so_stall_is_detectable` (stream_progress.rs:332) | the clock fix | Non-vacuous: a reviewer empirically reconstructed the pre-fix design and confirmed the second assertion (`after 30s silence → since_last==30s`) reproduces the exact production silent-degradation failure; both epoch-skew directions are caught by the two assertions together |
| `guard_registers_on_new_and_removes_on_drop_including_early_drop` (stream_progress.rs:404) | RAII registry cleanup | Covers normal-scope drop AND early/cancellation drop; asserts `registry.len()==0` after both; exercises the real `StreamProgressGuard` |
| 5 original streaming regression tests | inactivity/ceiling/race/non-streaming | All intact, updated for the new `record()` signature, non-vacuous (timing asserts break if the clock weren't shared), DST-correct (`start_paused` / `VirtualTime`) |

**Residual (Nice-to-Have, non-blocking):**
- The end-to-end reader→registry→writer→`record()` seam is covered piecewise (handle-sharing, guard lifecycle, retry-loop behavior) but not by one test that round-trips a handle through `registry.get()` and asserts the registered handle's `since_last()` reflects a `record()` on the fetched clone. Cheap to add (pure unit).
- No compile-time guard against re-introducing a `record()` timestamp parameter — a source-pin (like the existing `non_streaming_put_uses_fixed_timeout` pin) would make the regression vector explicit.

---

### Documentation
- The round-1 doc lies ("cannot leak", "removed unconditionally") are now true after the RAII guard.
- **Should (doc honesty, non-blocking):** the new clock test's comment claims it "fails on pre-fix code" — true only because the old `record(now_millis)` signature wouldn't compile against the test, not because it runtime-exercises the cross-epoch saturation. Either soften the comment or add a literal two-`RealTime` saturation test for a faithful reproduction. The structural argument already makes the bug unreachable, so this is cosmetic.

---

### Recommendations

#### Must Fix (Blocking)
None. The round-1 blocker is structurally resolved and verified by all three lenses.

#### Should Fix (Important)
None blocking.

#### Consider (Suggestions)
1. Soften the `writer_and_reader_share_one_clock_so_stall_is_detectable` comment (or add a literal two-epoch test) — it's a single-clock proxy, not a runtime reproduction of the cross-epoch saturation.
2. Add an end-to-end seam test (registry round-trip of the shared-clock handle) — closes the only inspection-only gap.
3. Add a source-pin asserting `record(&self)` takes no timestamp arg, to make the regression vector explicit.

---

### Verdict
**State:** Ready to Merge (pending CI green)
**HEAD SHA reviewed:** cc63810d6c56193c159392be14d5a4f0d12dec01

This is a re-review after fixes for round-1 findings #1 (Must, clock epoch) and #2 (Should, RAII leak) — **both resolved.** All three lenses independently confirm the cross-epoch clock bug is structurally gone (one `RealTime` per streaming PUT; writer mutates a clone of the reader's Arc-shared clock), the RAII guard is correct and cancellation/panic-safe, and the two new tests genuinely guard their invariants. No new findings of any severity. The only follow-ups are three non-blocking Considers (test-comment honesty, an end-to-end seam test, a source-pin). Once the running CI jobs (Unit & Integration, Simulation, Windows) go green on this HEAD, this is mergeable.

Excellent turnaround — the fix is the correct design (handle owns its clock, removing the two-instance coupling entirely) rather than a patch over the symptom.

[AI-assisted - Claude]

PR #4599, opened 2026-06-27
fix: break bootstrap acceptance collapse in cold-start sim (#4362)

### LINKED ISSUE #4362: fix: bootstrap acceptance collapse — saturated gateways + terminus rejections + location backoff stall ring formation for ~5 minutes in 100-node sim
## Summary

Split out of #4361 (defect 2), which fixed the GET-side symptom (empty-ring GETs now fall back to configured gateways) but left the underlying topology-formation stall unaddressed.

In `test_get_reliability_diagnostic` (100 nodes + 3 gateways, `min_connections=4`, `max_connections=12`, seed `0x3570_D1A6_0001`, deterministic), ring-connection formation has a ~5-minute dead zone:

- **238** `add_connection` events before T+14s (initial join storm, concentrated on/near the 3 gateways)
- **10** during the next ~5 minutes — acquisition effectively dead
- **462** after ~T+320s — 65% of topology formation happens only after the stall clears

During the dead zone, CONNECT traffic continues but fails: **5,936** `connect: at terminus, cannot accept, no uphill peers available — rejecting` events. The 3 gateways saturate at `max_connections=12` within seconds; late joiners' CONNECTs route into the saturated gateway neighborhood, are rejected, stamp location backoff, and retry into the same wall. Acceptance unblocks ~T+320s, after which the topology converges.

## Additional evidence (from the #4361 fix run at 875b97a4 + fix)

- **46 of 100 nodes still had `peer_ready == false`** (no completed transport handshake) when their scheduled GET signal arrived — as late as T+306s. Their client GETs were rejected with the explicit `PeerNotJoined` error. This is the same collapse seen from the client API side.
- The zero-connection re-bootstrap path referenced in #4352 never fired (0 log occurrences across runs).
- #4352's under-min backoff escape can't help a node whose own ring is empty — its CONNECTs have nowhere to route except the saturated gateways.

## Impact

- A freshly bootstrapped network (or any cold-start burst of joiners) spends minutes with most nodes unable to acquire ring connections.
- `test_get_reliability_diagnostic` currently asserts only a 1/3 dispatch floor (42/100 GETs dispatch) — that floor should be raised toward 100% when this is fixed (see the comment at the dispatch assertion in `simulation_integration.rs`).

## Possible directions (needs investigation)

- Gateway-side: don't terminus-reject when the joiner has zero ring connections — forward/queue instead, or respond with alternative acceptor candidates outside the saturated neighborhood.
- Joiner-side: detect repeated terminus rejections and re-route CONNECT through a different gateway / random location instead of stamping location backoff and retrying into the same wall.
- Acceptance-side: reserve a fraction of gateway `max_connections` slots for short-lived bootstrap relaying rather than stable topology.

## Related

- #4361 — GET-side symptom (fixed: bounded gateway fallback when ring is empty)
- #4352 — connection-convergence fix that unmasked the GET symptom
- #4348 — prior nightly topology-test redness

[AI-assisted - Claude]

iduartgomez:
## Problem

In a 100-node + 3-gateway deterministic cold-start sim
(`test_get_reliability_diagnostic`, seed `0x3570_D1A6_0001`,
`min_connections=4`, `max_connections=12`) ring formation had a ~5-minute
dead zone. The 3 gateways saturate at `max_connections` within seconds.
Late joiners' CONNECTs route into the saturated gateway neighborhood, get
terminus-rejected ("no uphill peers available — rejecting"), and then a
3-way feedback loop traps them:

1. **Acceptor side** — `ConnectionManager::should_accept` returns false once
   `total_conn >= max_connections`; in the saturated neighborhood the relay
   also can't forward uphill (`select_uphill_hop` returns `None`), so it
   replies `ConnectMsg::Rejected`.
2. **Joiner side** — on `Rejected` the connect driver stamped the full
   escalating **30s→600s** location backoff on `dl` (a jitter of the
   joiner's *own* location). Late joiners cluster near the same gateways, so
   every retry routes into the same wall and re-stamps an ever-longer
   backoff on its own location.
3. **Result** — 46/100 nodes were still `peer_ready == false` at T+306s,
   and only 42/100 scheduled GETs ever dispatched (the rest hit
   `PeerNotJoined`).

## Solution

Three joiner-side changes (lowest blast radius — `should_accept` is **not**
touched):

**(2a) Short, non-escalating reject backoff under min_connections.** On a
capacity `Rejected` while still below `min_connections`, the driver now
stamps a fixed **3-second** backoff (`UNDER_MIN_REJECT_BACKOFF`, jittered
±20%) on `dl` instead of the escalating 30s→600s one. Under min, a Rejected
is the *acceptor's* capacity problem, not a property of `dl`, so escalating
the per-location backoff is the trap. A short fixed pause still throttles
the retry cadence (so the joiner stops hammering a saturated gateway every
fast-tick) but never ramps toward the 600s trap, so the under-connected
node keeps probing. At/above min the full escalating backoff is preserved
(there a Rejected is a real "stop probing this region" signal). This is the
throttled refinement of the #4348 "under-min ignores location backoff"
escape. The new `TrackedBackoff::record_short_backoff` never shortens an
already-longer escalated backoff and never increments the failure counter.

**(2b) Widened bootstrap re-route after jitter saturation.** Once the
existing bounded jitter saturates (its magnitude caps at 0.25 by
`JITTER_SATURATION_FAILURES = 5` consecutive failures) and the node is
still bootstrapping (`current_connections < GAP_TARGET_THRESHOLD`), each
retry takes — with probability `WIDE_REROUTE_PROBABILITY = 0.5` — a wider
bounded jitter step (up to half the ring, the max ring distance) so it
routes *away* from the saturated neighborhood toward regions with spare
capacity. A full ring-wide *uniform* draw was tried and rejected: it sends
the joiner to an arbitrary location whose greedy route crosses every
saturated region and terminus-rejects on the way, amplifying chatter
without improving reach-min. All widening randomness is `GlobalRng`; the
failure counter is the existing time-sourced `increment_connect_jitter_failures(now)`.

**(2c) Gateway rotation.** Verified existing behavior: the gateway backoff
(`gateway_backoff.record_failure(gateway_addr)`) is recorded
**unconditionally** on every Rejected, so the next attempt advances to a
different gateway via the `is_in_backoff` filter in
`initial_join_procedure`. No change needed.

**Why the `connect_rejected` metric was dropped from the regression test.**
An earlier draft asserted `connect_rejected < ~5936` (the broken-code
baseline). That premise is structurally wrong: the broken code's *low*
rejection count was a **symptom of stuck nodes** — a Rejected stamped a
600s backoff and the joiner stopped retrying. The fix unblocks those nodes
by letting under-min joiners keep probing (throttled), which inherently
produces **more** connection attempts and therefore **more** terminus
rejections, while actually forming the topology. Rejection count moves
*opposite* to health, so it is logged for diagnostics only; the honest
signal is the reach-min timeline.

## Testing

All numbers seed `0x3570_D1A6_0001`, 100 nodes + 3 gateways,
min=4/max=12. BEFORE = broken `main`; AFTER = this PR:

| Metric | BEFORE | AFTER |
|---|---|---|
| GET dispatch (`test_get_reliability_diagnostic`) | 42/100 | **86/100** |
| GET success rate | — | **100% (91/91)** |
| nodes reaching min_connections by T+120s | 46/100 *not ready* @ T+306s | **95/103** |
| `connect_rejected` (diagnostic only) | ~5936 | 23,552 |

- **`test_get_reliability_diagnostic`** (sharpened): dispatch floor raised
  from the catastrophic-only `NUM_NODES/3` (33) to **80/100** (achieved 86,
  minus a ~6-node jitter margin); success-rate floor raised from 0.50 to
  **0.90** (achieved 100%). Both fail on broken `main`.
- **`test_bootstrap_acceptance_no_dead_zone`** (new): asserts on the
  topology-formation **timeline** rather than GET outcomes — that ≥90% of
  nodes reach `min_connections` by a mid-sim virtual-time checkpoint
  (T+120s), decoded deterministically from each `Connected` event's ULID
  timestamp (`Transaction::created_at_ms()` → `GlobalSimulationTime`). Fails
  red on broken `main` (46/100 still wedged at T+306s), green here (95/103).
- **`ConnectionBackoff` unit tests** (new):
  `test_short_reject_backoff_does_not_escalate`,
  `test_short_reject_backoff_never_shortens_escalated`.
- **Regression guards** (steady-state churn — the main risk of the joiner
  retry change): `test_connection_growth_stall_regression`,
  `test_multi_step_churn`. (`test_ring_simulated_network` does not exist in
  this tree; substituted the canonical growth-stall + churn guards.)

DST-safe: all new time via the injected `TimeSource` /
`tokio::time::Instant` (respects `start_paused`); all new randomness via
`GlobalRng`; thresholds are named constants or derived from
`min_connections`.

## Fixes #4362

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -93,17 +93,38 @@ CORRECT:
 WHY: Backing off a target for a local problem prevents connecting to it
 later when local conditions improve (e.g., more connections acquired).
 
-COROLLARY (under-min backoff escape, #4348):
-  A node still BELOW min_connections must IGNORE per-target-location
-  backoff entirely and keep probing. Its under-connection is by definition
-  a local capacity problem, so a `Rejected`-stamped 30s→600s location
-  backoff would trap a poorly-positioned node permanently below min.
-  connection_maintenance gates this via should_respect_location_backoff()
-  (honor backoff only at/above min). Storm safety is preserved by the
-  adaptive fast-tick backoff + max_concurrent cap, not by location backoff.
+COROLLARY (under-min backoff escape, #4348 + #4362 refinement):
+  A node still BELOW min_connections must not let the full escalating
+  30s→600s location backoff trap it: its under-connection is by definition
+  a local capacity problem, and a `Rejected`-stamped escalating backoff on
+  a poorly-positioned node would park it permanently below min.
 
-See: ring.rs connection_maintenance + should_respect_location_backoff,
-issues #3414, #4348
+  This plays out differently on the two paths that consult location backoff:
+
+  - MAINTENANCE loop (ring.rs connection_maintenance): IGNORES per-target
+    location backoff entirely while under min, gated by
+    should_respect_location_backoff() (honor backoff only at/above min).
+
+  - CONNECT DRIVER retry path (operations/connect/op_ctx_task.rs, #4362):
+    on a capacity `Rejected` while under min it does NOT ignore location
+    backoff entirely — it stamps a SHORT, fixed, NON-escalating reject
+    backoff (UNDER_MIN_REJECT_BACKOFF = 3s, time-bounded via TrackedBackoff::
+    record_short_backoff, which never escalates the failure count and never
+    shortens an already-longer escalated backoff). This is what provides
+    storm safety on the driver path: skipping the backoff entirely (an
+    earlier #4362 attempt) let under-min joiners re-probe every fast-tick
+    and produced a ~12x terminus-rejection storm. The short fixed pause
+    throttles the retry cadence without ramping toward the 600s trap, so the
+    joiner keeps probing and the 2b widened re-route can find spare capacity.
+
+  At/above min, BOTH paths honor the full escalating backoff (there a
+  `Rejected` is a genuine "stop probing this region" signal). On the
+  maintenance loop, storm safety under min is still the adaptive fast-tick
+  backoff + max_concurrent cap (not location backoff).
+
+See: ring.rs connection_maintenance + should_respect_location_backoff;
+operations/connect/op_ctx_task.rs Rejected arm + ConnectionBackoff::
+record_short_reject_backoff; issues #3414, #4348, #4362
 ```
 
 ### Routing Decisions
--- a/crates/core/src/message.rs
+++ b/crates/core/src/message.rs
@@ -81,11 +81,17 @@ impl Transaction {
     }
 
     /// Milliseconds-since-Unix-epoch encoded in this transaction's ULID at
-    /// creation time. In simulation mode this is `GlobalSimulationTime`
-    /// (deterministic virtual time), so it can be used by tests to anchor a
-    /// virtual-time checkpoint against the simulation epoch. In production it
-    /// is the wall-clock creation time. Cheap and feature-independent (the
-    /// `trace-ot`-gated `started()` exposes the same value as a `SystemTime`).
+    /// creation time.
+    ///
+    /// In production this is the wall-clock creation time. In simulation mode
+    /// (`GlobalSimulationTime` set) it is a fixed epoch plus a counter that
+    /// increments by 1 per ULID GENERATED — NOT advanced by the simulation's
+    /// virtual clock. So across transactions it is a deterministic, monotonic
+    /// GENERATION-ORDER value, usable by tests as an *early-run ordering proxy*
+    /// against the simulation epoch — but it is NOT a literal virtual-time
+    /// reading, and its calibration to virtual seconds depends on ULID volume.
+    /// Cheap and feature-independent (the `trace-ot`-gated `started()` exposes
+    /// the same value as a `SystemTime`).
     pub fn created_at_ms(&self) -> u64 {
         self.id.timestamp_ms()
     }
--- a/crates/core/src/message.rs
+++ b/crates/core/src/message.rs
@@ -80,6 +80,16 @@ impl Transaction {
         self.elapsed() >= crate::config::OPERATION_TTL
     }
 
+    /// Milliseconds-since-Unix-epoch encoded in this transaction's ULID at
+    /// creation time. In simulation mode this is `GlobalSimulationTime`
+    /// (deterministic virtual time), so it can be used by tests to anchor a
+    /// virtual-time checkpoint against the simulation epoch. In production it
+    /// is the wall-clock creation time. Cheap and feature-independent (the
+    /// `trace-ot`-gated `started()` exposes the same value as a `SystemTime`).
+    pub fn created_at_ms(&self) -> u64 {
+        self.id.timestamp_ms()
+    }
+
     #[cfg(feature = "trace-ot")]
     pub fn started(&self) -> SystemTime {
         SystemTime::UNIX_EPOCH + Duration::from_millis(self.id.timestamp_ms())
--- a/crates/core/src/operations/connect.rs
+++ b/crates/core/src/operations/connect.rs
@@ -1444,6 +1444,41 @@ pub(crate) async fn join_ring_request(
 
     let current_connections = op_manager.ring.connection_manager.connection_count();
 
+    // Number of consecutive connect failures after which the bounded jitter
+    // magnitude (0.05 * 2^(failures-2)) saturates at its 0.25 cap: with the
+    // `(failures - 2).min(3)` shift, the cap is reached at `failures == 5`.
+    // Beyond this point, growing the offset no longer moves the target — the
+    // joiner keeps landing within ±0.25 of its own location, i.e. back in the
+    // SAME saturated gateway neighborhood. This is the #4362 bootstrap dead
+    // zone: late joiners cluster near a few gateways, those gateways hit
+    // max_connections, terminus-Rejects pile up, and bounded jitter cannot
+    // escape the local wall.
+    const JITTER_SATURATION_FAILURES: u32 = 5;
+    // Once failures exceed the saturation point AND we are still bootstrapping,
+    // widen the bootstrap target so retries route AWAY from the saturated
+    // neighborhood toward regions with spare capacity. One extra attempt of
+    // headroom past saturation avoids over-eagerly abandoning local
+    // exploration on a brief transient. See #4362.
+    //
+    // The widening is a BOUNDED probabilistic expansion of the jitter window,
+    // NOT a fresh uniform draw across the whole ring. A full ring-wide random
+    // target sends the joiner to an arbitrary location whose greedy route
+    // crosses the saturated regions and terminus-rejects on the way — that
+    // amplified the network-wide rejection count ~12x without improving the
+    // already-high reach-min rate. A bounded expansion (up to half the ring,
+    // the maximum ring distance) lets the joiner escape its local wall while
+    // keeping routes short, so it finds spare capacity without flooding.
+    const WIDE_REROUTE_FAILURE_THRESHOLD: u32 = JITTER_SATURATION_FAILURES + 1;
+    // Probability, per saturated retry, of taking the widened jitter rather
+    // than the standard bounded one. Keeping most retries at the standard
+    // magnitude preserves NAT-friendly local exploration; the occasional wide
+    // step is what breaks out of the dead zone. Named (not a bare literal):
+    // a coin flip is the simplest rate that escapes without dominating.
+    const WIDE_REROUTE_PROBABILITY: f64 = 0.5;
+    // Maximum widened jitter magnitude = half the ring (the largest possible
+    // ring distance). Bounds the escape step so routes stay short.
+    const WIDE_REROUTE_MAX_MAGNITUDE: f64 = 0.5;
+
     // Helper: apply jitter to base_location for NAT-compatible acceptor exploration.
     // After consecutive hole-punch failures, the same location routes to the same
     // acceptors. Jittering explores different ring regions.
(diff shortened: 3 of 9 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4599

### Summary
- **PR Title:** fix: break bootstrap acceptance collapse in cold-start sim (#4362)
- **Type:** fix (routing / bootstrap / topology formation)
- **CI Status:** passing (Simulation, NAT, Windows, Clippy, Fmt, Rule Lint all green)
- **Linked Issues:** Fixes #4362
- **Review tier:** Full (routing/consensus + new backoff variant + maintenance-loop interaction)
- **Reviewers run:** skeptical, code-first, testing, big-picture (4 Claude lenses). **External pass (codex) unavailable — token expired; gemini not installed.** Per policy, ran 4 diverse-lens Claude reviewers and **hand-verified the two contested findings** (2a throttle reach; the ULID-vs-virtual-time checkpoint, on which two reviewers disagreed).

---

### Code-First Analysis
**Independent Understanding:** Joiner-side fix for the CONNECT bootstrap stall. On a terminus-`Rejected`, the driver now gates on `current_connections >= min_connections` (op_ctx_task.rs:500): at/above min keeps the escalating per-location backoff; **below min** applies a new short non-escalating 3s backoff (`record_short_reject_backoff`) — so a late joiner stops re-stamping the 30s→600s trap on its own location and keeps probing (2a). After repeated bootstrap rejections, a `GlobalRng` wide-reroute (±0.5) routes the CONNECT away from the saturated neighborhood (2b). Gateway rotation via `gateway_backoff` stays unconditional (2c).

**Stated Intent:** #4362 — dismantle the 3-way feedback loop (gateways terminus-reject → joiner stamps location backoff on its own location → retries into the same wall). Joiner-side chosen as lowest blast radius; `should_accept` deliberately untouched.

**Alignment:** Strong. The fix attacks the named root cause, not a symptom. Verified `should_accept` is untouched. The 2a/2b/2c wiring is coherent; the math checks out (jitter saturates at `failures==5`, wide-reroute threshold 6 = one attempt of headroom; offset ∈ [−0.5,+0.5] = half-ring). Constants are well-documented. The concerns below are concentrated in **test/diagnostic instrumentation and docs**, not shipped behavior.

---

### Skeptical Findings
**Risk Level:** low — no high-severity correctness bugs; the two headline risks (never-shortens, time-boundedness) are handled correctly in production.

| "never-shortens-escalated" test doesn't pin the invariant it's named for | **Should** | `connection_backoff.rs:325` (verified) | `record_short_backoff` correctly does `retry_after.max(now + jittered)` (backoff.rs:267) — a 600s escalation survives a 3s short call. **But** `test_short_reject_backoff_never_shortens_escalated` asserts only `failure_count==1` and `is_in_backoff==true`. `is_in_backoff` is true for *both* a 3s and a 30s remaining backoff, so the test would pass even if the short path wrongly overwrote (not `.max()`-ed) a 30s escalation down to 3s — exactly the regression it promises to catch. `TrackedBackoff::remaining_backoff` exists (backoff.rs:192) but `ConnectionBackoff` doesn't re-export it. Fix: assert `remaining_backoff > ~10s` (or add a direct `TrackedBackoff::record_short_backoff` test — it currently has zero direct tests). Independently flagged by 2 reviewers. |
| `last_failure` refresh = permanently-refreshable GC exemption | Consider (Low) | `backoff.rs:264` + `cleanup_expired:301` (verified) | Each short call sets `last_failure = now`; `cleanup_expired` needs `is_past_retry && is_stale (now−last_failure > 600s)`, so a node rejected every few seconds on the same bucket refreshes `last_failure` forever and that path never GCs it. **Bounded/benign:** 3s expiry for `is_in_backoff`, `consecutive_failures` stays 0 (no escalation masked), map LRU-capped at 256. But it technically violates the AGENTS.md "cleanup exemptions must be time-bounded or age-overridden" rule — worth an absolute-age override on `last_failure`. |
| 2a comment overstates the mechanism's reach | Consider | `op_ctx_task.rs:490-497` (verified) | **Hand-verified:** the only `connection_backoff` consumer is the maintenance loop (ring.rs:3989), gated by `respect_backoff` which is `false` under-min (`should_respect_location_backoff` = `count >= min`). So under-min — exactly when 2a fires — the maintenance loop bypasses the location backoff entirely. 2a's short backoff therefore throttles the **connect-driver retry cadence**, NOT the topology loop; the comment "so the joiner stops hammering every fast-tick" is broader than what the location-backoff map delivers under-min. Not a bug (the driver retry IS throttled, and the 3s has expired by the time the node crosses min so no transition trap), but the rationale comment should be narrowed. |

Verified safe (skeptical + me): never-shortens `.max()` correct in production; DST/GlobalRng in 2b (all draws via seeded GlobalRng, `now` is `tokio::time::Instant` used only for jitter decay); under-min gate has no off-by-one (boundary `count==min` takes full-escalation, consistent with `should_respect_location_backoff`); 2b self-terminates (fires only `failures>=6 && conns<3`, counter decays + resets on success, and 3 < min 4 so steady-state nodes never perturbed) — no livelock; `record_short_backoff`'s `Instant::now()` is the **pre-existing** TrackedBackoff convention (lines 183/194/209…), not a new DST violation.

---

### Testing Assessment
**Coverage Level:** end-to-end behavior covered by sim; unit coverage of the new primitives has gaps.

| Unit | ⚠️ | 2 new `ConnectionBackoff` tests; the never-shortens one is weak (above). 2b wide-reroute has **no** unit test (closure inside `join_ring_request`, not directly callable). Short-backoff **expiry** (the "short" half of the contract) untested. |
| Integration / Sim | ⚠️ | `test_bootstrap_acceptance_no_dead_zone` + sharpened `test_get_reliability_diagnostic` are faithful and deterministic (seed `0x3570_D1A6_0001`), red on broken main, green here. **But both are `#[cfg(feature = "nightly_tests")]` — they do NOT run per-PR.** A future dead-zone regression passes per-PR CI and only surfaces ~24h later in nightly. Combined with the thin unit coverage, the per-PR guard is weak. |

**Regression discipline — exemplary on the dropped assertion:** the PR dropped the planned `connect_rejected < 5936` upper bound. The reasoning is **sound and verified**: the low 5936 baseline was a *symptom of stuck nodes* (a Rejected stamped 600s backoff → joiner stopped retrying → fewer rejections), so the fix — which makes under-min joiners keep probing — necessarily drives rejections UP (23k) as the network heals. Rejection count moves *opposite* to health. Rather than weakening a bad metric, the author **raised** two real floors (success 0.50→0.90, dispatch 33→80) and **replaced** the wrong metric with the stronger reach-min timeline. This is the anti-CI-chasing pattern done right.

**Reach-min margin:** the assertion is `reached_min >= ceil(100 * 0.90) = 90`; achieved 95 → **5-node headroom** (the PR's "95/103" includes 3 gateways, used only in the log line). Deterministic, so it can't "dip" without a code change — but the floor is tuned to one seed; recommend a comment recording the achieved 95.

---

### Big Picture Assessment
**Goal Alignment:** genuine fix of the root cause (not symptom-shifting) — verified.
**Anti-Patterns Detected:** none. No `#[ignore]`, no removed tests, no weakened assertions (the dropped assertion was *replaced by stronger ones*).
**Removed Code Concerns:** only the unconditional `record_connection_failure(Rejected)` (now under-min-conditional) and the two weak assertion baselines (replaced).
**Scope Assessment:** focused — `should_accept` untouched, 8 files load-bearing. **One avoidable bit of scope creep:** `connect_connection_count()` (event_kind.rs:348) is byte-identical to the pre-existing, zero-caller `connect_peer_connection_count()` (event_kind.rs:503). Reuse the existing one (and ideally delete the dead twin) instead of adding a third near-identical accessor.

---

### Documentation
| Item | Severity | Details |
|------|----------|---------|
| `.claude/rules/ring.md` #4348 corollary now stale | **Must** | ring.md (lines 96-103) states under-min nodes "IGNORE per-target-location backoff **entirely**" and "storm safety is preserved … **not by location backoff**." This PR's 2a adds exactly a short location backoff for storm safety on the connect-driver path. Per AGENTS.md ("fix outdated docs in the same PR"), update the corollary to note the #4362 refinement (the connect-driver path now applies a short, non-escalating, time-bounded 3s reject backoff; the maintenance loop still bypasses location backoff under-min via `should_respect_location_backoff`). Add `#4362` to the `See:` line. |
| "virtual-time checkpoint" claim is inaccurate | **Should** | **Hand-verified at config.rs:3076-3093.** `created_at_ms()` decodes the ULID timestamp = `GlobalSimulationTime = base + counter`, where `counter` increments **by 1 per ULID generated**, NOT advanced by the sim's virtual clock. So the test's `CHECKPOINT_MS=120_000` filter selects "transactions among the first ~120k ULIDs generated," not "created in the first 120s of sim time." The docstrings at message.rs:83-89 ("deterministic virtual time", "T+120s checkpoint") and the test prose overstate this — it's a monotonic *ordering* proxy whose calibration depends on ULID-generation volume, making the 90% floor fragile to unrelated connect/op churn. Correct the doc/test framing (or anchor against a real virtual-time source so "T+120s" is literally true). |
| `record_short_backoff` doc imprecise | Consider | backoff.rs:243 says "repeated calls always apply the same delay" — but `.max()` means repeated calls extend by at most one window, not a rolling 3s-from-last. Tighten wording. |
| new-test `connect_rejected` diagnostic conflates two event types | Consider | counts `ConnectEvent::Rejected`, labeled as the issue's terminus-rejection (`is_connect_terminus_rejected`) "~5936" baseline; they co-occur so it's fine in practice, but relabel. |

---

### Recommendations

#### Must Fix (Blocking)
1. **Update `.claude/rules/ring.md` (#4348 corollary, lines 96-103)** to reflect the #4362 short-backoff refinement — under-min nodes no longer ignore location backoff *entirely* on the connect-driver path. AGENTS.md requires this in the same PR.

#### Should Fix (Important)
1. **Strengthen `test_short_reject_backoff_never_shortens_escalated`** to assert `remaining_backoff` stays near the escalated value (or add a direct `TrackedBackoff::record_short_backoff` test). As written it doesn't guard the `.max()` line — the exact invariant that prevents resetting a legit escalation.
2. **Fix the "virtual-time checkpoint" framing** (doc + test) — it's a ULID-generation counter, not simulated seconds; the 90% floor is calibrated to ULID volume, not the clock.
3. **Per-PR guard:** decide explicitly whether the dead-zone regression should have *any* per-PR coverage given both sim tests are nightly-only; if not, compensate with stronger unit coverage (the never-shortens fix above, plus a 2b wide-reroute unit test and a short-backoff-expiry test).
4. **Reuse `connect_peer_connection_count`** instead of the duplicate `connect_connection_count` accessor.

#### Consider (Suggestions)
1. Add an absolute-age override on `last_failure` so the short-backoff entry can't be permanently refreshed (AGENTS.md time-boundedness), even though impact is currently benign.
2. Narrow the 2a rationale comment to "throttles the connect-driver retry cadence" (the maintenance loop bypasses location backoff under-min).
3. Record the achieved reach-min (95) in a test comment so the 90 floor's 5-node margin is legible.
4. Expose the sim epoch from the harness rather than re-deriving `BASE_EPOCH_MS`/`RANGE_MS` in the test.

---

### Verdict
**HEAD SHA reviewed:** 988ac3ad7568beb260eba4854b49e022ada63127

This is a strong, well-documented fix that solves #4362's actual root cause, chose the correct low-blast-radius option for an S-needs-design issue, and **strengthens** the test suite rather than weakening it (the dropped-assertion reasoning is a model example). The shipped behavior has no high-severity correctness bugs — the never-shortens `.max()` is correct, 2b is determinism-safe and livelock-free, the under-min gate has no off-by-one. The findings are concentrated in **docs and test instrumentation**: the Must is a one-paragraph ring.md update (required by AGENTS.md); the Shoulds are a test that doesn't pin its named invariant, a mislabeled "virtual-time" checkpoint, a nightly-only guard, and a duplicate accessor. None require re-architecting; a light re-check on the doc + test fixes suffices.

Excellent work on the diagnosis and the honest test-semantics reframe.

[AI-assisted - Claude]

PR #4602, opened 2026-06-27
fix: stop placement migration firing in every simulation (#4601)

### LINKED ISSUE #4601: Nightly 500-node sim red since 0.2.80: placement migration runs in EVERY simulation (frozen SubscribeHint floor ≤ crate version), storming past the 60-min cap
## Problem

The **Large scale test (500 nodes)** step of the `Simulation Tests (Nightly)` workflow has been red on almost every run since 0.2.80 shipped. It fails by hitting the GitHub 60-minute wall-clock timeout (not an assertion), drowning in subscribe / renewal / `get-subop` timeouts. The 50-node medium test and the long-running nightly suite keep passing, so the redness is specific to the 500-node case.

Recent runs (`Simulation Tests (Nightly)`):

| Date | Crate version (main HEAD at run) | SubscribeHint floor | Migration active in sim? | Result |
|------|----------------------------------|---------------------|--------------------------|--------|
| 06-20 | **0.2.79** | (0,2,80) | OFF (below floor; code not yet merged) | ✅ exits cleanly ~45 min via internal 1800s propagation-timeout |
| 06-21 | **0.2.80** | (0,2,80) | **ON** | ❌ 60-min timeout |
| 06-22 | 0.2.81 | (0,2,80) | **ON** | ❌ |
| 06-23 / 06-24 | 0.2.82 | (0,2,80) | **ON** | ❌ |
| 06-25 | 0.2.82 | (0,2,80) | **ON** | ✅ (variance: marginal sim occasionally finishes < 60 min) |
| 06-26 / 06-27 | 0.2.83 | (0,2,80) | **ON** | ❌ |

(Current main is 0.2.84; the floor is frozen at 0.2.80, so every release from 0.2.80 onward is affected.) A nightly that is red ~4 of every 5 nights, with no open issue tracking it, erodes the signal (people learn to ignore it, and a real regression hides in the noise). Both related issues are now closed (#4440, #4534) but the symptom persists, so this files the orphan.

## Root cause: the SubscribeHint version floor inverted when 0.2.80 released

Placement migration (#4404) is gated behind a **frozen** version floor:

```rust
// crates/core/src/node/network_bridge/p2p_protoc.rs:755
pub(crate) const SUBSCRIBE_HINT_MIN_VERSION: (u8, u8, u16) = (0, 2, 80);
```

The migration's "stays dormant in simulations" safety relied on sim peers reporting a crate version **below** the floor. That assumption is stated explicitly in the code:

> `crates/core/src/node.rs:309-315` — "Simulation peers all report the current build version, which is below the production floor until release, so the `SubscribeHint` gate would never fire in a sim. A test that exercises the migration cascade sets this to `Some((0,0,0))` ... so the cascade is opt-in and cannot perturb unrelated simulations."

That inverted the moment 0.2.80 was released and the crate version reached the frozen floor:

- `own_crate_version()` = `CARGO_PKG_VERSION` at compile time (`p2p_protoc.rs:765`). In a single-process sim every peer compiles at the same version (0.2.83 in the latest failing nightly, 0.2.84 on current main).
- Sim peers run the **real** handshake over `SimulationSocket`, so `remote_protoc_version` is populated with the peer's crate version (`connection_handler.rs:2043`), not left `None`.
- Send gate: `peer_supports_subscribe_hint` → `version_supports_subscribe_hint(Some(0.2.83), floor=0.2.80)` = `0.2.83 >= 0.2.80` = **true** (`migration.rs:14-26`, `p2p_protoc.rs:811-815`).
- Receive gate: `own_crate_version() 0.2.83 >= floor 0.2.80` = **true** (`node.rs:1779-1798`).

So both the send and receive gates now pass in **every** simulation, and placement migration runs unthrottled (the #4537 module-cache admission gate at `node.rs:1845-1860` never engages because sim module-cache occupancy never reaches the ceiling — see the `node.rs:312` note). Every hosted contract nudges its closest non-hosting neighbor into a directed subscribe (`node.rs:1866-1868` → `start_directed_subscribe`), generating a storm of `get-subop` GET sub-ops and (now) subscription-root renewals against an already-marginal 500-node topology.

The floor is **deliberately frozen** at 0.2.80 for wire-compat (`p2p_protoc.rs:729-732`, `docs/RELEASING.md`), so this will keep firing in every sim on every future release. It is not self-healing.

## Evidence (warn-level histograms, `RUST_LOG=warn`)

Passing baseline (06-20, migration OFF): `subscribe: attempt timed out` = 46, `get-subop: attempt timed out` = 83. Failing runs:

| Signal | 06-22 (0.2.81) | 06-27 (0.2.83, incl. #4583) |
|--------|----------------|------------------------------|
| `is_renewal` (renewal-timeout flood) | 8,724 | **21,727** |
| `subscribe: attempt timed out` | 8,743 | 10,066 |
| `get-subop: attempt timed out` (migration directed-subscribe proxy) | 2,342 | 2,539 (~30× baseline) |
| `GET relay: all peers exhausted` | 7,162 | 12,316 |

The `get-subop` flood (the warn-level proxy for migration-driven directed subscribes; the `Received SubscribeHint` / `starting directed subscribe` lines are debug-level) appeared exactly when the crate crossed the floor (0.2.79 → 0.2.80) and has been stable across every failing run since.

## Prior fixes that did NOT green the nightly

- **#4583 / #4440** ("stop placement-migration subscription roots from storming on renewal") landed 06-26 and is in today's failing 06-27 run. The renewal flood actually grew through it (8.7k → 21.7k), so the storm fix is incomplete at 500-node scale.
- **#4537 / #4534** (migration-admission gate at 90% module-cache occupancy) fixed the production "contract queue full" symptom but does not engage in a sim, so it does nothing for this nightly.
- 0.2.82 / 0.2.83 / 0.2.84 did not touch the subscribe/migration load path.

## Interaction with #4362 (bootstrap acceptance collapse)

The 500-node sim is already marginal independent of migration: even the passing 06-20 run showed `RING_TRANSPORT_DESYNC`, isolated zero-connection nodes, and NAT-traversal failures, and exited only via the internal 1800s propagation-timeout (convergence not achieved) at ~45 min rather than truly converging. That topology-formation stall looks like the same class as **#4362** (Nacho's open PR **#4599**). Migration does not *cause* the unhealthy topology; it piles directed-subscribe + renewal load onto an already-marginal network and pushes the run past the 60-min CI cap. #4599 may improve the bootstrap side; the migration-active-in-all-sims regression is separate and would remain.

## The open design question (needs a decision)

Two defensible directions, and I would like the release agent's opinion before implementing either:

1. **Restore sim isolation.** The general 500-node *throughput/convergence* nightly should run migration **OFF** (default the testing-path `subscribe_hint_floor_override` to a fail-closed value above any crate version, so it is genuinely opt-in as the code comment claims), and add a **dedicated opt-in migration-at-scale test** that enables migration via `SimNetwork::enable_placement_migration` and asserts *bounded* directed-subscribe + renewal load. This greens the nightly immediately and gives migration its own honest scale test. Production behavior is unchanged (production keeps `None` → real floor).

2. **Keep migration on and fix the storm harder.** Treat the 500-node nightly as the canary for migration load (the #4440 / #4583 direction). This has not converged after #4583, and it conflates "is migration load-safe at scale" with "does the base 500-node sim converge" (#4362), so a red nightly persists while two separate problems are debugged.

My lean is (1): the throughput nightly was never meant to test migration, the "opt-in, cannot perturb unrelated simulations" contract is what the code already documents, and a dedicated opt-in test is a cleaner home for the migration-scale signal. But this touches the subscribe / p2p surface (Full-review tier) and overlaps active migration work, so flagging for a second opinion rather than unilaterally changing it.

## Repro

```
target/release/fdev test \
  --name nightly-large-500 --seed 0x500BEEF \
  --gateways 10 --nodes 490 --events 10000 \
  --ring-max-htl 15 --max-connections 30 --min-connections 10 \
  --latency-min 10 --latency-max 50 --min-success-rate 1.0 \
  --print-summary --print-network-stats single-process
```
On any 0.2.80+ build this runs with placement migration active (no flag needed). Latest failing run: https://github.com/freenet/freenet-core/actions/runs/28278425471

## Related

- #4404 — the GET placement gap that migration exists to fix (OPEN)
- #4440 — placement-migration re-enable umbrella (CLOSED) + #4583 storm fix (in the failing run)
- #4534 / #4537 — production module-cache overrun + admission gate (CLOSED; gate inert in sim)
- #4350 / #4546 — renewal outer-cancel fix (added the `is_renewal` telemetry now flooding)
- #4362 / #4599 — bootstrap acceptance collapse in cold-start sims (interacting, OPEN)
- #4348 — prior "nightly simulation tests red" issue (CLOSED), for history

@sanity — filing per your request so the agent that cut the 0.2.84 release can weigh in on the design question above.

[AI-assisted - Claude]

sanity:
## Problem

The **Large scale test (500 nodes)** step of `Simulation Tests (Nightly)` has been red on ~4 of every 5 runs since 0.2.80 shipped (#4601). It is not an assertion failure — it hits the GitHub 60-minute wall-clock cap while drowning in subscribe / renewal / `get-subop` timeouts.

Root cause (confirmed in #4601): the placement-migration (`SubscribeHint`) cascade was supposed to stay dormant in simulations. That safety relied on sim peers reporting a crate version *below* the frozen floor `SUBSCRIBE_HINT_MIN_VERSION = (0,2,80)`. The moment 0.2.80 released, the crate version reached the floor, so the gate now fires in **every** simulation. Migration runs unthrottled in the general throughput nightly, piling directed-subscribe + renewal load onto an already-marginal 500-node topology (#4362) and pushing the run past the cap. The `node.rs` comment still claimed migration was "opt-in and cannot perturb unrelated simulations" — that contract had silently inverted.

This is a **simulation-only artifact**. The renewal path is already hard-bounded in production: `recover_orphaned_subscriptions` spawns at most `MAX_RECOVERY_ATTEMPTS_PER_INTERVAL = 10` renewals per `SUBSCRIPTION_RECOVERY_INTERVAL = 30s` per node, and prod telemetry shows real peers well under that cap with ~zero channel-full drops. So **no production change and no release are needed.**

## Solution (TEST-ONLY)

Option 1 from #4601 — restore sim isolation and give migration a dedicated, honest scale test.

1. **Restore sim isolation (the fix that greens the nightly).** `SimNetwork` now defaults the per-node `subscribe_hint_floor_override` to an unreachable floor (`SIM_MIGRATION_DISABLED_FLOOR = (255,255,65535)`) instead of `None`. This makes the `SubscribeHint` gate FAIL-CLOSED in every simulation regardless of build version — migration is genuinely opt-in, exactly as the docs always claimed. The general 500-node throughput nightly (`fdev test … single-process` → `SimNetwork::new`, which never opts in) now runs migration OFF and exits cleanly. **Production is untouched:** `NodeConfig::new` still sets `None` → the real `(0,2,80)` floor.

2. **Dedicated opt-in migration-at-scale test** (`test_placement_migration_at_scale_renewal_load_stays_bounded`). It explicitly calls `enable_placement_migration()`, hosts 20 contracts on one node (all entering the renewal window together so demand far exceeds the cap), runs past the renewal window, and asserts **bounded load**, not merely that the sim completes:
   - **(A) cap holds, every node:** per-node `max_cycle_batch <= MAX_RECOVERY_ATTEMPTS_PER_INTERVAL (10)`. This is the regression guard — if the per-cycle spawn cap were removed, the loaded node would record `max_cycle_batch == 20` and fail.
   - **(B) cap engaged, loaded node:** `max_cycle_batch == 10`, proving demand exceeded the cap and the cap actually clipped (so (A) isn't vacuous).
   - **(C) migration ran:** at least one contract migrated off the loaded node onto a closer neighbor, so the cascade is genuinely exercised.

   To measure (A)/(B) deterministically, a sim-gated `max_cycle_batch` counter was added to the existing renewal-metrics registry (`record_renewal_cycle_batch`, called once per renewal cycle). It is a **no-op in production** (gated on a simulation network name, exactly like the existing `wire_attempts` / `terminus_satisfied` counters) — no production behaviour change.

3. **Updated the now-stale comments** (`node.rs` `subscribe_hint_floor_override` doc, `SimNetwork::{enable,disable}_placement_migration` docs, the field doc) to reflect that the override now defaults fail-closed.

## Testing

- `cargo fmt` + `cargo clippy --locked -p freenet -- -D warnings` (CI-equivalent): clean.
- New test passes across 3 seeds: `loaded_max_cycle_batch=10`, `agg_max_cycle_batch=10` (cap held for every node), `migrated_pairs=72` (migration ran).
- Existing migration/renewal/fan-out tests still pass with the default flip + new metric: `test_contract_migrates_to_close_cluster_resolving_get_dead_end`, `test_get_dead_ends_at_close_cluster_without_migration`, `test_subscription_root_renewal_does_not_storm`, `test_no_interest_subscription_root_lease_lapses`, `test_fanout_liveness_incident_scale_migration_{off,on}`.
- The full 500-node repro (`fdev test --name nightly-large-500 …`) is ~60 min wall-clock and impractical to run locally here; the fix is structural — the fdev nightly path constructs `SimNetwork` without opting into migration, so the new fail-closed default turns migration OFF for it. CI's nightly run is the end-to-end confirmation.

This is **TEST-ONLY**: the production renewal path is already bounded (≤10/30s/node, frozen `SUBSCRIBE_HINT_MIN_VERSION`), so there is no production behaviour change and no release is required.

Closes #4601

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -307,16 +307,25 @@ pub struct NodeConfig {
     #[serde(skip)]
     pub(crate) governance_config_override: Option<crate::contract::governance::GovernanceConfig>,
     /// Test-only override for the placement-migration version floor
-    /// (`SUBSCRIBE_HINT_MIN_VERSION`). Simulation peers all report the current
-    /// build version, which is below the production floor until release, so the
-    /// `SubscribeHint` gate would never fire in a sim. A test that exercises the
-    /// migration cascade sets this to `Some((0,0,0))` for its own nodes, leaving
-    /// every other sim (and production) at the real floor — so the cascade is
-    /// opt-in and cannot perturb unrelated simulations.
+    /// (`SUBSCRIBE_HINT_MIN_VERSION`). The floor is consulted as
+    /// `subscribe_hint_floor_override().unwrap_or(SUBSCRIBE_HINT_MIN_VERSION)`
+    /// on both the send gate (`p2p_protoc::peer_supports_subscribe_hint`) and the
+    /// receive gate (`node::handle_pure_network_message_v1`).
     ///
-    /// `None` in production. Not cfg-gated for the same reason as
-    /// `governance_config_override`: `node::testing_impl` sets it and is compiled
-    /// unconditionally. `#[serde(skip)]`; never serialized.
+    /// In production this is `None` → the real `(0,2,80)` floor (untouched).
+    ///
+    /// In simulations the crate version is now AT/ABOVE the real floor, so a
+    /// `None` override would make the `SubscribeHint` gate fire in EVERY sim and
+    /// pile migration load onto unrelated simulations (the #4601 regression that
+    /// reddened the 500-node nightly). `SimNetwork` therefore defaults this to an
+    /// unreachable floor (`SimNetwork::SIM_MIGRATION_DISABLED_FLOOR`) — FAIL-CLOSED,
+    /// so migration is genuinely OPT-IN. A test exercising the cascade calls
+    /// `SimNetwork::enable_placement_migration`, which sets it to `Some((0,0,0))`
+    /// for its own nodes; every other sim stays off and cannot be perturbed.
+    ///
+    /// Not cfg-gated for the same reason as `governance_config_override`:
+    /// `node::testing_impl` sets it and is compiled unconditionally.
+    /// `#[serde(skip)]`; never serialized.
     #[serde(skip)]
     pub(crate) subscribe_hint_floor_override: Option<(u8, u8, u16)>,
 }
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -417,6 +417,8 @@ impl ControlledSimulationResult {
             |mut acc, m| {
                 acc.wire_attempts += m.wire_attempts;
                 acc.terminus_satisfied += m.terminus_satisfied;
+                // Per-node peak → aggregate as the max across peers.
+                acc.max_cycle_batch = acc.max_cycle_batch.max(m.max_cycle_batch);
                 acc
             },
         )
@@ -1010,11 +1012,14 @@ pub struct SimNetwork {
     /// path actually feeds the detector. See #4301.
     governance_config_override: Option<crate::contract::governance::GovernanceConfig>,
     /// Per-node placement-migration version-floor override (see
-    /// [`SimNetwork::enable_placement_migration`]). `None` leaves the production
-    /// floor, so the `SubscribeHint` cascade stays OFF in sim (its peers report a
-    /// version below the floor) — only a test that explicitly enables migration
-    /// gets the cascade. Keeps the migration feature from perturbing unrelated
-    /// simulations.
+    /// [`SimNetwork::enable_placement_migration`]). Defaults to
+    /// `Some(SIM_MIGRATION_DISABLED_FLOOR)` (an unreachable floor) so the
+    /// `SubscribeHint` cascade is FAIL-CLOSED — OFF — in every sim regardless of
+    /// build version, and only a test that explicitly calls
+    /// `enable_placement_migration` gets the cascade. This keeps migration from
+    /// perturbing unrelated simulations now that the crate version is past the
+    /// real production floor (#4601). Production is untouched: `NodeConfig::new`
+    /// sets this to `None`, which resolves to the real `SUBSCRIBE_HINT_MIN_VERSION`.
     subscribe_hint_floor_override: Option<(u8, u8, u16)>,
     /// Per-node capture slots for the live `Arc<Ring>`, populated by
     /// `run_controlled_simulation` so governance sim tests can read each
@@ -1043,6 +1048,19 @@ pub struct SimNetwork {
 impl SimNetwork {
     /// Default seed for deterministic simulation
     pub const DEFAULT_SEED: u64 = 0xDEADBEEF_CAFEBABE;
+
+    /// Placement-migration version floor that pins the `SubscribeHint` cascade
+    /// OFF for a simulation: an unreachable value above any real or simulated
+    /// crate version, so `version_supports_subscribe_hint` always returns false.
+    /// This is the per-node default for every `SimNetwork` (see `new_inner`),
+    /// making migration genuinely OPT-IN in sims; `enable_placement_migration`
+    /// lowers it to [`Self::SIM_MIGRATION_ENABLED_FLOOR`]. See #4601.
+    pub(crate) const SIM_MIGRATION_DISABLED_FLOOR: (u8, u8, u16) = (255, 255, 65535);
+
+    /// Placement-migration version floor that forces the `SubscribeHint` cascade
+    /// ON for a simulation regardless of build version (every peer's version is
+    /// `>= (0,0,0)`). Set by [`Self::enable_placement_migration`].
+    pub(crate) const SIM_MIGRATION_ENABLED_FLOOR: (u8, u8, u16) = (0, 0, 0);
 
     #[allow(clippy::too_many_arguments)]
     pub async fn new(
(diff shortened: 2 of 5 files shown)


sanity:
## Review

Ran `codex review --base origin/main` (external, non-Claude pass). Outcome: **no correctness issues found** — Codex confirmed the simulation default change, renewal-metric aggregation, and the new regression test are internally consistent with the stated opt-in migration behavior.

Risk surface is limited: the only production-code change is a sim-gated, no-op-in-production renewal counter (`record_renewal_cycle_batch`, same gating as the existing `wire_attempts`/`terminus_satisfied` counters); the behavioral change is confined to the default value of a test-only config field (`SimNetwork`'s `subscribe_hint_floor_override`). Production (`NodeConfig::new` → `None` → real `(0,2,80)` floor) is untouched.

Local verification: new test green across 3 seeds (cap held + engaged + migration ran); existing migration/renewal/dead-end/lease tests still green; `cargo fmt` + `cargo clippy --locked -p freenet -- -D warnings` clean.

[AI-assisted - Claude]

Issue #4592, opened 2026-06-27
feat(hosted): low-friction migration of delegate secrets to your own peer


sanity:
## Problem

The hosted \"try Freenet\" proxy (#4381) makes the **exit** one click — Account → Export downloads an encrypted \`freenet-data.fnsx\` bundle over HTTPS — but the **entrance** is a terminal command: to keep their data, a user must install a peer, **stop it**, run \`freenet secrets import <file>\` with their access key (#4035), and restart. That's a friction cliff at the exact moment we want to convert a non-technical person who just tried Freenet in the browser. The export's smoothness is wasted if the import is a node-stopped CLI operation.

## Goal

Make \"take my data home\" as smooth as the export: a browser / in-app action on the user's **own** node, **live** (no node stop), low-friction.

## Design (from a 2026-06-26 design discussion)

The durable piece is a **live, source-agnostic import flow** on the receiving node — a path that writes incoming delegate secrets into the local (single-user) `SecretsStore` *without stopping the node* (today's CLI import requires it stopped). Build it around a clean \"source of incoming secrets\" abstraction, then implement sources:

1. **Magic-link HTTPS pull (near-term, reliable).** The hosted node mints a one-time, short-lived, single-use migration link. The bundle behind it is re-encrypted with a fresh key; the link path authorizes the download (server-side, TTL'd, single-use) and the decryption key rides in the URL **fragment** (never sent to the server or its logs). The user opens the link on their own node → it pulls the bundle from try.freenet.org over HTTPS, decrypts, and imports into Local. The durable **access key never leaves the browser**. Single-use + short TTL = a focused \"do it now\" migration.
2. **File upload (fallback).** The existing \`freenet-data.fnsx\` download (#4035) fed into the same import flow, for when try.freenet.org is down or the link expired.
3. **Freenet-contract pull (dogfood upgrade, later).** The data moves via an encrypted contract on Freenet — the one-shot version of #4560's transfer. **Gated on GET/SUBSCRIBE reliability (#4404, #4414):** a dead-ending GET would fail the migration with no fallback, so this is the dogfood upgrade once the transport is trustworthy, not v1.

### Conflict handling (deliberately minimal)

[...]

PR #4603, opened 2026-06-27
feat: live secret-import core + local import endpoint (#4592 chunk 1)

sanity:
## Problem

The hosted "try Freenet" proxy (#4381) makes the **exit** one click — Account → Export downloads an encrypted `freenet-data.fnsx` bundle over HTTPS — but the **entrance** is a terminal command: to keep their data a user must install a peer, **stop it**, run `freenet secrets import` with their key (#4035), and restart. That node-stopped CLI step is a friction cliff at the exact moment we want to convert a non-technical person. The CLI must stop the node because it opens the secrets ReDb directly (single-writer).

This is **chunk 1/4 of the secrets-migration v1** (#4592): the durable, source-agnostic LIVE-import core + a local import endpoint. It is the keystone and the file-upload fallback's backend.

## Solution

Route the import **through the running node's executor** (which owns the one legitimate `SecretsStore`) instead of opening ReDb directly, so no node stop is needed. Mirrors the existing live EXPORT path, with one deliberate difference (below).

- `Executor::import_secrets` / `Runtime::import_secret_bundle` — mutating mirror of `export_user_secrets` / `export_secret_bundle`, calling the existing `secret_export::import_bundle` (unchanged for v1).
- `ContractHandlerEvent::ImportSecrets{..}` + `ImportSecretsResponse`. Bundle key in a `RedactedToken` (wiped/redacted); bundle bytes in an `ImportBundle` whose `Debug` prints only a length. Neither is ever logged.
- `RuntimePool::import_secrets`: pops an executor, runs the import, returns it.
- New `POST /v{1,2}/import` (`server/client_api/hosted_import.rs`): raw `application/octet-stream` body (no axum multipart feature), key material + kind + overwrite in headers, a `DefaultBodyLimit` sized to accept any bundle `export_bundle` can produce. Reuses the export endpoint's `ExportOpManagerHandle`. Submits via `notify_contract_handler_prioritized` at `Priority::ClientLocal`.

### On-loop, NOT off-loop (deliberate; resolves a Codex P1)

The EXPORT runs off-loop because it is a read-only, authenticated-**remote** DoS surface. The IMPORT is the **first writer** to `SecretsStore` outside the serial loop's existing on-loop write path (delegate `store_secret`). That write path (fixed sibling `.tmp` + check-then-write) assumes **node-wide write serialization**; running the import off-loop on a separate pooled executor would let it race an on-loop delegate write (or another import) on the same secret file and corrupt it. So the import runs **on** the contract loop, keeping every secret write in one serialization domain. The brief loop-block is acceptable: this endpoint is gated to **loopback + the node's own dashboard origin** (one-shot operator migration), not the remote-token DoS surface that justified off-loop export. (An initial off-loop version was flagged by Codex as a P1 concurrency/corruption risk; this resolves it with zero change to the crypto write path.)

### The gate (fail-closed) — the key review item

The receiving node is a NORMAL single-user node (hosted OFF) importing into `SecretScope::Local`, so it does **not** reuse the hosted-export token gate. Since this endpoint **writes** secrets it must be reachable ONLY from the node's own dashboard/shell origin over loopback. `import_gate_or_reject` enforces, in order, fail-closed:

1. **Loopback source IP** from the kernel-set `ConnectInfo` (missing ⇒ reject — mirrors `hosted_export`).
2. **Trusted origin** via the SAME `permission_prompts::is_origin_trusted` the CSRF guard uses — which **rejects `Origin: null`** (i.e. the sandboxed contract iframe, whose sandbox omits `allow-same-origin`) and cross-site origins; requires loopback / allowed-host.
3. **No per-contract `AuthToken`** — a per-contract web shell is same-origin with the node, so the token is what distinguishes it: reject any request presenting an origin-contract-registered token via the `Authorization` bearer header, the `authToken` query param, or the `Authorization` cookie.

The bundle's decryption key is the data authorization. Decrypt is all-or-nothing and runs **before any write** (`import_bundle` → `open_bundle`), so a wrong key is a 4xx (`ImportBadBundle`) with **nothing written**, never a 500 or a partial write. Key material rides in a header (never the URL/logs), redacted/zeroizing, never logged.

## Testing

- **14 gate unit tests** (`hosted_import`): loopback / missing-ConnectInfo (fail-closed), `Origin: null` sandbox iframe, cross-site origin, per-contract token via header **and** cookie, ipv4-mapped loopback, key-kind / bearer parsing.
- **4 live integration tests** (`tests/hosted_mode_import.rs`): import a bundle over `POST /v1/import` and read the secret back **through the live stack while the node stays up** (the property the node-stopped CLI couldn't provide); wrong key → 4xx + **no write**; missing `Origin` → 403; collision (`overwrite=false`) skips + reports.
- **7 existing live export tests** still pass.
- `cargo fmt` + `cargo clippy --lib -- -D warnings` clean; `codex review --base origin/main` (P1 resolved; one P2 — body-size cap — addressed).

**Note: test-only, no release. Crypto-sensitive — the import gate is the key review item.** Holding before merge for a full review.

## Review follow-up (multi-lens + Codex)

Full review (multi-lens + Codex) came back clean (no Must-fix). Addressed the Should-fix/Consider items plus findings from re-review:

- **Gate doc accuracy**: reworded check #3 (no per-contract `AuthToken`) as best-effort defense-in-depth that only classifies a *well-behaved* client; the load-bearing sandbox-iframe exclusion is check #2 (`Origin: null` rejection). Added a maintainer warning not to weaken #2 believing #3 backstops it.
- **Stale docs**: fixed `handle_deferred_resume` (it resumes the #4391 related-contract upsert, not an import) and `send_import_response_inline` (sole response path for the on-loop import, all outcomes).
- **Test coverage**: added a live PASSPHRASE-KDF (Argon2id) round-trip, an `overwrite=true` end-to-end test, and tightened the wrong-key assertion to `== 422`.
- **No-partial-write (Codex P2)**: `import_bundle` now validates every entry's key/hash lengths *before* any write, so a malformed-but-authenticated bundle is rejected with nothing written (matches the endpoint's documented guarantee). Regression test added.

### Deferred: `Sec-Fetch-Site` hardening (deliberate follow-up)

NOT done this round (a Consider, not a blocker). Enforcing `Sec-Fetch-Site` would add a fetch-metadata CSRF layer on top of the Origin gate, but the header is browser-only: a legitimate non-browser client (a future `freenet secrets import --live` CLI, curl in a migration script) omits it entirely. "Reject when missing" would break those clients; "enforce only when present" is weaker — the choice needs a deliberate call, so it's left as a documented follow-up (see the comment near `import_gate_or_reject`). Until then the Origin gate (#2) is the CSRF boundary.

Refs #4592

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -1319,10 +1319,11 @@ where
     }
 }
 
-/// Send an `ImportSecretsResponse` to a still-waiting client inline (the
-/// rejection paths: no deferral ctx, busy, unsupported). Logs at debug if the
-/// client already disconnected. The import mirror of
-/// [`send_export_response_inline`] (#4592).
+/// Send the `ImportSecretsResponse` to the waiting client. Because the live
+/// import (#4592) runs fully ON the contract loop (no off-loop deferral, unlike
+/// the export), this is the SOLE response path for the import and carries EVERY
+/// outcome — success and failure alike — not just rejections. Logs at debug if
+/// the client already disconnected.
 async fn send_import_response_inline<CH>(
     contract_handler: &mut CH,
     id: handler::EventId,
@@ -1342,11 +1343,15 @@ async fn send_import_response_inline<CH>(
     }
 }
 
-/// Handle a completed off-loop import (#4592): return/replace the executor in the
-/// pool and answer the parked client. Runs ON the loop (the pool return needs
-/// `&mut contract_handler`), but the work is just a pool slot-return + a oneshot
-/// send — microseconds, never the import itself. Mirror of
-/// [`handle_export_resume`].
+/// Resume a PUT/UPDATE upsert whose related-contract fetch was deferred off-loop
+/// (#4391). Runs ON the contract loop. Reclaims the parked client responder,
+/// injects the off-loop-fetched related states so the re-run resolves them
+/// locally (no second network round trip), re-runs the upsert in DEFERRABLE mode
+/// (a further `DeferRelated` is converted to `MissingRelated` — the one-deferral
+/// cap), then answers the parked client exactly once. This is the related-contract
+/// upsert resume path, NOT a secrets-import path: the live import (#4592) runs
+/// fully on-loop and never defers (see the `ImportSecrets` arm and
+/// [`send_import_response_inline`]).
 async fn handle_deferred_resume<CH>(
     contract_handler: &mut CH,
     deferral_ctx: &mut DeferralCtx,
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -24,10 +24,11 @@ pub(crate) use executor::{
 // Re-export CRDT emulation functions for testing
 pub use executor::mock_runtime::{clear_crdt_contracts, is_crdt_contract, register_crdt_contract};
 pub(crate) use handler::{
-    ClientResponsesReceiver, ClientResponsesSender, ContractHandler, ContractHandlerChannel,
-    ContractHandlerEvent, NetworkContractHandler, RedactedToken, SenderHalve, SessionMessage,
-    StashedResponder, StoreResponse, WaitingResolution, WaitingTransaction,
-    client_responses_channel, contract_handler_channel,
+    BundleKeyKind, ClientResponsesReceiver, ClientResponsesSender, ContractHandler,
+    ContractHandlerChannel, ContractHandlerEvent, ImportBundle, ImportTargetScope,
+    NetworkContractHandler, RedactedToken, SenderHalve, SessionMessage, StashedResponder,
+    StoreResponse, WaitingResolution, WaitingTransaction, client_responses_channel,
+    contract_handler_channel,
     in_memory::{
         MemoryContractHandler, MockWasmContractHandler, MockWasmHandlerBuilder,
         SimulationContractHandler, SimulationHandlerBuilder,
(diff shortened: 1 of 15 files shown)


sanity:
### Independent review: Codex (`codex review --base origin/main`)

External non-Claude pass run on this crypto-sensitive change. Both findings addressed:

- **[P1] Off-loop import races the SecretsStore write path** — the original version ran the import off-loop (mirroring export), but the import is the first *writer* outside the serial loop, and the store write path (fixed sibling `.tmp` + check-then-write) assumes node-wide write serialization. Two concurrent imports, or an import racing a delegate `store_secret` on the same secret, could corrupt the write. **Fixed**: the import now runs **on** the contract loop (serialized with delegate writes); the loop-block is acceptable because the endpoint is loopback + dashboard-gated (one-shot operator migration), not the authenticated-remote DoS surface that justified off-loop export. Zero change to the crypto write path.
- **[P2] Import body cap could reject a valid max-size export** — `MAX_EXPORT_TOTAL_PLAINTEXT_BYTES` counts decrypted plaintext, but the bundle is CBOR-serialized (each secret `Vec<u8>` as an integer array, ~2x) before encryption, so a valid max export can exceed `plaintext + 16 MiB`. **Fixed**: `MAX_IMPORT_BUNDLE_BYTES = 2 × plaintext-cap + 64 MiB`, sized to accept anything `export_bundle` can produce.

Re-ran Codex after the rework; P1 no longer flagged, P2 addressed. Holding before merge for the full multi-model review.

[AI-assisted - Claude]

PR #4609, opened 2026-06-28
fix: exit-for-restart on redb poison instead of bricking forever (#4604)

### LINKED ISSUE #4604: Node silently bricked (all contract ops fail forever) after a single transient redb I/O error — no poison recovery
## Summary

A single **transient** redb I/O error permanently and silently bricks a node's contract operations. There is no recovery path: once redb's in-memory poison flag is set, every subsequent contract GET / PUT / UPDATE fails forever with `Previous I/O error occurred. Please close and re-open the database.` The process keeps running, so neither systemd nor the exit-42 auto-update hook ever recycles it. Blast radius is total (all contract ops dead) and the failure is silent (the node looks alive).

**This is a general node-resilience bug — the absence of poison recovery.** It will bite *any* peer that hits a transient I/O error on the redb file, regardless of the underlying cause. (See "How it surfaced" below: the node where this was observed had failing local non-ECC RAM, which was merely the *trigger* that produced the transient I/O error. The RAM is not the bug; the lack of recovery is.)

## Mechanism

In `crates/core/src/contract/storages/redb.rs`:

- `ReDb(Arc<Database>)` (line 128) wraps a single redb `Database`, opened **once** at startup in `ReDb::new` → `Database::create` (lines 131-171) and shared via `Arc` for the life of the process.
- Every storage operation begins a transaction and propagates errors with `?`, e.g. `let txn = self.0.begin_read()?;` / `let txn = self.0.begin_write()?;` (call sites throughout the file: 340, 353, 363, 377, 392, 414, ... ~35 of them).
- The **only** open-time special-casing is the `UpgradeRequired` version migration (the `is_version_mismatch` branch at lines 141-160). There is no runtime catch for a poisoned database and no close-and-reopen anywhere.

redb sets an in-memory poison flag after **any** I/O error (e.g. a disk `EIO`). Once poisoned, every `begin_read()` / `begin_write()` returns `StorageError` with the "Previous I/O error occurred. Please close and re-open the database." message. Because the `Arc<Database>` is never dropped and re-created, the node never recovers in-process; and because the process does not exit, systemd never restarts it and the exit-42 auto-update hook never fires. The node stays "running" while 100% of its contract operations fail.

## How it surfaced (real occurrence)

Observed on a node running **0.2.84**. The node's machine had failing **non-ECC RAM** that corrupted data in flight. btrfs detected a checksum (csum) failure on the redb `db` file and returned `EIO`; redb poisoned itself; from that point every contract operation failed for hours while the node continued to report as "running".

To be explicit about root cause vs. defect:

- The **machine's** root cause was bad RAM corrupting data — that is a hardware fault, **not** a Freenet bug.
- The **Freenet defect** is the absence of any *recovery* from a transient I/O error. The bad RAM was only the trigger that exposed it. The same permanent-brick outcome would follow from any transient disk `EIO`, a brief storage-layer hiccup, a filesystem csum failure from any cause, etc. — on healthy hardware too.

## Proposed fix (two options)

On a redb `StorageError` that indicates poisoning, do one of:

**(a) Recover in-process** — drop and re-create the `Arc<Database>` (close + reopen the redb file) so subsequent transactions get a fresh, un-poisoned handle. Cleaner UX: the node self-heals without a restart. Requires care around the shared `Arc` (all clones must observe the new handle) and around the transaction that was in flight when poisoning occurred.

**(b) Terminate for supervisor restart** — exit the process with the restart exit code (the same one the auto-update / health path already uses, exit 42) so systemd recycles the node and the health/update hooks fire. Simpler; leans entirely on the supervisor; incurs a restart but guarantees a clean handle.

(a) is the better experience; (b) is the smaller, lower-risk change. Either is strictly better than the current "fail forever, look alive" behavior.

## Classification

Reliability / availability bug. No data corruption is introduced by Freenet here (the corruption, when present, is upstream of redb); the defect is that a recoverable transient condition is turned into a permanent, silent, node-wide outage.

[AI-assisted - Claude]

sanity:
## Problem

A single **transient** redb I/O error permanently and silently bricks a node's contract operations, with no recovery path.

redb latches an in-memory poison flag (`io_failed`) after **any** backend read/write error. From then on every `begin_write()` returns `StorageError::PreviousIo` ("Previous I/O error occurred. Please close and re-open the database."), so every contract PUT/UPDATE — and the hosting-metadata write performed on essentially every access — fails forever. Because the process never exits, **systemd never restarts it and the exit-42 auto-update hook never fires**: the node stays "running" while 100% of its contract ops fail. Blast radius is total, the failure is silent.

Observed on a 0.2.84 node whose failing non-ECC RAM produced a btrfs csum failure → `EIO` → redb poison. The bad RAM is the *trigger*; the **defect is the absence of any recovery from a transient I/O error** — the same permanent brick follows from any transient disk `EIO` / fs hiccup on healthy hardware.

## Solution

**Option (b): detect the poison at the storage-op layer and EXIT the process** so the supervisor recycles the node with a fresh, un-poisoned database handle (and the health/update hooks fire).

Chosen over in-process close+reopen (option a) because reopen does **not** interact with the crash-loop protection: on *persistent* corruption (the real-world case) it would churn-reopen forever, re-creating the exact "silently broken, supervisor never engages" failure this issue is about. Exit-for-restart routes persistent corruption into the existing crash-loop limiter; in-process reopen can't.

### Detection coverage (all three storage entry points)

Poison is routed to the recovery path wherever it can surface:

1. **`begin_write` / `begin_read` wrappers** — catch a poison at transaction start. `begin_write` is the steady-state choke point: redb checks the poison flag on every `begin_write` (`check_io_errors`), and the node writes hosting metadata on essentially every contract access.
2. **`commit_guarded`** (every write commit) — the FIRST backend I/O error usually surfaces here as `CommitError::Storage(Io)`, on the very op that poisons the handle, so recovery fires immediately rather than waiting for the next `begin_write`.
3. **`read_guarded`** (every read method body) — redb's `begin_read` does NOT check the poison flag (it serves the cached snapshot), so a poisoned read fails later at `open_table`/`get`/iteration; this gives read-only workloads the same prompt exit-for-restart.

### Precise detection — two classifiers, one deliberate asymmetry

- **Transaction/commit path** (`storage_error_is_poison`, on `StorageError`): matches `PreviousIo | Io | LockPoisoned | DatabaseClosed`. These errors come straight from redb and are never synthesized, so matching `Io` is safe.
- **Read-method-body path** (`redb_error_is_poison`, on the umbrella `redb::Error`): matches `PreviousIo | LockPoisoned | DatabaseClosed` but **NOT `Io`** — several read methods SYNTHESIZE `redb::Error::Io(InvalidData)` for a benign malformed-data row (e.g. a wrong-length `CodeHash`), and treating that as poison would exit-and-restart the node on a single bad row (a crash loop). A genuine backend I/O poison is still caught: redb latches its flag on the first backend `Io`, so the next backend read returns `PreviousIo` (caught here) and the next `begin_write` returns `PreviousIo` (caught by the transaction-path classifier).

Benign not-found (`Ok(None)`), `TableDoesNotExist`, `Corrupted`, `ValueTooLarge`, etc. are never treated as poison, so a normal missing-contract never restarts the node.

The whole mechanism is **opt-in** (`enable_abort_on_redb_poison`, called only by the real `freenet` binary) so simulation/integration tests and library embedders never have a storage error kill their host process.

### Interaction with crash-loop protection (#4591 / #4588 / systemd StartLimit)

The poison exit reuses the existing `fatal_listener_exit_code(uptime, fast_crash_enabled)` decision (#4549/#4551) — **no new exit code is introduced**:

- **Poison after a healthy uptime** (>= 60s) → exit **42**: burst-exempt (`SuccessExitStatus=42`), fires the `freenet update` self-heal hook. A genuinely transient I/O blip (the #4604 case: a node that ran healthily for hours) restarts cleanly without tripping `StartLimitBurst`.
- **Poison within the 60s healthy window**, under a supervisor that understands it → exit **45** (the counted fast-crash code): not in `SuccessExitStatus`, so it counts toward `StartLimitBurst`, **and** the systemd unit runs `freenet update` on 42 OR 45 so the #4591 auto-rollback probation counts it via `FREENET_POST_STOP_EXIT_CODE`. A database that **re-poisons on every boot** (persistent disk/RAM corruption) therefore trips StartLimit / rolls back instead of looping forever.

Net: a transient poison self-heals with one restart; a persistent one is bounded by the same protection the fatal-listener path already uses.

## Testing

- `redb_poison_classifier_is_precise` — produces **real** redb errors via a fault-injecting `StorageBackend` (a genuine `StorageError::Io`, then `PreviousIo` once redb latches the poison flag) and asserts they classify as poison; asserts the synthetic `Io(InvalidData)` malformed-row error does **not** (the false-positive guard); asserts a benign table-open error is not poison. Variant-based, so it can't drift with redb wording.
- `poisoned_redb_takes_recovery_path_benign_does_not` — end-to-end: a benign not-found does not trigger recovery; the poisoning write (commit-time I/O) takes the recovery path on the same op; a subsequent poisoned write does too (observed via a test counter; abort is opt-in and off in tests, so it returns instead of exiting).
- `redb_poison_exit_reuses_crash_loop_bounded_codes` / `redb_poison_abort_is_opt_in` — pin the counted-vs-burst-exempt exit-code interaction (a re-poison within 60s uses the counted code) and the opt-in default.

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and the redb + p2p_impl test suites are green. Codex review (`--base origin/main`) run across three iterations; its two findings (read-path and commit-time coverage) are fixed and the final pass is clean.

## Deploying

This needs a **release** to reach live nodes; build + PR + review only here. It interacts with the crash-loop protection as described above, so it should ride a normal release (no special rollout).

Closes #4604

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -255,6 +255,15 @@ async fn run_network_node_with_signals(
         freenet::enable_fast_crash_exit_code();
     }
 
+    // #4604: this is the real node process, so let the contract storage layer
+    // exit-for-restart if redb gets poisoned by a transient I/O error (e.g. a disk
+    // EIO / filesystem csum failure). Without this the node stays "running" while
+    // every contract GET/PUT/UPDATE fails forever; with it, the supervisor restarts
+    // the node with a fresh database handle. The exit code reuses the fatal-listener
+    // decision above, so a database that re-poisons on every boot (persistent
+    // corruption) is bounded by the same crash-loop protection rather than looping.
+    freenet::enable_abort_on_redb_poison();
+
     // #4073 crash-loop auto-rollback: once this (possibly freshly auto-updated)
     // node has run healthily for the commit window, clear any post-update
     // probation marker so ordinary later crashes never trigger a rollback. The
--- a/crates/core/src/contract/storages/redb.rs
+++ b/crates/core/src/contract/storages/redb.rs
@@ -160,6 +160,29 @@ fn transaction_error_is_poison(e: &TransactionError) -> bool {
     matches!(e, TransactionError::Storage(s) if storage_error_is_poison(s))
 }
 
+/// True if the umbrella [`redb::Error`] signals a poisoned database. Used to catch
+/// poison that surfaces on the READ path AFTER `begin_read` has already succeeded
+/// (redb's `begin_read` does not check the poison flag, so a poisoned read fails
+/// later at `open_table` / `get` / iteration as a `StorageError` flattened into this
+/// umbrella type).
+///
+/// IMPORTANT — this deliberately does NOT match `redb::Error::Io`, unlike
+/// [`storage_error_is_poison`]. Several read methods in this file SYNTHESIZE
+/// `redb::Error::Io(ErrorKind::InvalidData)` for a benign malformed-data row (e.g. a
+/// wrong-length `CodeHash`). Treating that as poison would exit-and-restart the node
+/// on a single bad row → a crash loop. A genuine backend I/O poison is safe to skip
+/// here regardless: redb latches its poison flag on the first backend `Io`, so the
+/// very next backend read returns `PreviousIo` (caught here) and the next
+/// `begin_write` returns `PreviousIo` (caught by [`storage_error_is_poison`]). So
+/// `PreviousIo` (plus the unambiguous `LockPoisoned` / `DatabaseClosed`) is the
+/// precise, false-positive-free read-path signal.
+fn redb_error_is_poison(e: &redb::Error) -> bool {
+    matches!(
+        e,
+        redb::Error::PreviousIo | redb::Error::LockPoisoned(_) | redb::Error::DatabaseClosed
+    )
+}
+
 /// Test-only observable proof that the storage layer routed a detected poison to
 /// the recovery (process-exit) path. The real handler ([`abort_process_on_redb_poison`])
 /// exits the process, which a unit test cannot observe; this counter lets the test
@@ -210,6 +233,34 @@ impl ReDb {
             crate::node::abort_process_on_redb_poison(&e.to_string());
         }
         e
+    }
+
+    /// Umbrella-error counterpart of [`ReDb::route_txn_error`], for poison that
+    /// surfaces on the read path AFTER `begin_read` succeeded (at `open_table` /
+    /// `get` / iteration). Returns the error untouched.
+    fn route_redb_error(e: redb::Error) -> redb::Error {
+        if redb_error_is_poison(&e) {
+            #[cfg(test)]
+            POISON_RECOVERY_TRIGGERED.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+            crate::node::abort_process_on_redb_poison(&e.to_string());
+        }
+        e
+    }
+
+    /// Run a read-transaction body, routing ANY poison error to the #4604 recovery
+    /// path — not just a poison at `begin_read`, but one that surfaces later at
+    /// `open_table` / `get` / iteration (redb's `begin_read` does not check the
+    /// poison flag, so a poisoned read can fail mid-transaction). This gives
+    /// read-only workloads the same prompt exit-for-restart that `begin_write`
+    /// already gives write workloads, instead of failing every read until the next
+    /// write happens to hit `begin_write`.
+    fn read_guarded<T>(
+        &self,
+        f: impl FnOnce(&ReadTransaction) -> Result<T, redb::Error>,
+    ) -> Result<T, redb::Error> {
+        // begin_read already routes a poison at transaction start.
+        let txn = self.begin_read()?;
+        f(&txn).map_err(Self::route_redb_error)
     }
 
     pub async fn new(data_dir: &Path) -> Result<Self, redb::Error> {
@@ -434,12 +485,13 @@ impl ReDb {
         &self,
         key: &ContractKey,
     ) -> Result<Option<HostingMetadata>, redb::Error> {
-        let txn = self.begin_read()?;
-        let tbl = txn.open_table(HOSTING_METADATA_TABLE)?;
-        match tbl.get(key.as_bytes())? {
-            Some(v) => Ok(HostingMetadata::from_bytes(v.value())),
-            None => Ok(None),
-        }
+        self.read_guarded(|txn| {
+            let tbl = txn.open_table(HOSTING_METADATA_TABLE)?;
+            Ok(match tbl.get(key.as_bytes())? {
+                Some(v) => HostingMetadata::from_bytes(v.value()),
+                None => None,
+            })
+        })
     }
 
     /// Remove hosting metadata for a contract.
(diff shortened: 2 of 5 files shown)


sanity:
## Full multi-model review — clean, merging

High-risk surface (storage layer + crash-loop / process-exit codes), so Full tier: an external Codex pass plus three diverse Claude lenses (skeptical bug-hunt, code-first intent-vs-impl, crash-loop failure-mode), each reviewing the checked-out branch at HEAD `9029c0c0`.

**All four reviewers independently reached the same verdict: no CRITICAL or HIGH bugs — merge.** The two load-bearing claims were verified against redb 4.1.0 source rather than taken from the comments:
- The read-path classifier asymmetry is correct: `redb_error_is_poison` excludes `Io` (so a synthetic `Io(InvalidData)` malformed row can't kill the node), and `CheckedBackend::read()` latches `io_failed` so the next backend op returns `PreviousIo`, which IS caught — confirmed in `cached_file.rs:163-170` + `error.rs`.
- Coverage is complete for the network node: every read routes through `read_guarded`, every write through `begin_write`+`commit_guarded`; the only raw bypass is `initialize_database` (startup, before the flag is armed). No smuggled semantic change in the 990-line wrap.
- Opt-in is genuinely safe: `enable_abort_on_redb_poison()` is called only from `bin/freenet.rs:265`; no test or embedder path arms it, so `process::exit` can't fire in tests. Exit is crash-consistent (redb recovers on reopen); concurrent poison-exits are first-wins.

### Findings — all LOW, none blocking (documented for disposition)

1. **Comment overstates the crash-loop bound.** The "persistent corruption is bounded rather than looping forever" comments (`freenet.rs:263-264`, `p2p_impl.rs:185-196`) hold only for a **sub-60s** re-poison (counted exit 45 → StartLimitBurst). A DB that re-poisons only after ≥60s uptime exits 42 (burst-exempt) each boot and **slow-flaps indefinitely** (~60s of service per cycle). Behavior is acceptable — strictly better than the pre-PR permanent brick, keeps firing the update hook, and a ≥60s-cadence hardware fault is unfixable by any supervisor — and it's inherited from the #4549/#4551 machinery, not introduced here. Worth a one-line comment correction in a follow-up.
2. **Local mode (`run_local_node`) gets no poison protection** (skeptical L1). A `freenet local` node that poisons still bricks. Defensible as out-of-scope: local mode is dev/interactive, typically unsupervised, so exit-for-restart wouldn't help — but a conscious call rather than an accident.
3. **Two test-quality nits** (skeptical L5): (a) in `redb_poison_classifier_is_precise`, the `if let TableError::Storage(s) = open_table(missing)` arm never matches (a missing table returns `TableDoesNotExist`), so that sub-assertion silently never runs — the classifier is still pinned by the other real-redb-error assertions in the same test, but this one should be fixed or removed; (b) the read-path recovery is exercised by injecting a synthetic `PreviousIo` into the `read_guarded` closure rather than forcing a real read-triggered backend poison (the write path IS tested against a real `FailingBackend` poison).

The classifier's false-positive guard was specifically traced — every synthetic `Io(InvalidData)` malformed-row error flows through `redb_error_is_poison` (excludes `Io`), so a single bad row cannot trigger the exit. Behavior changes for ENOSPC/EINTR (now exit-for-fresh-handle) are fine-to-good since redb already latched `io_failed` in those cases.

None of the LOW items block. Merging; items 1-3 are cheap follow-up polish for the author.

[AI-assisted - Claude]

Issue #4610, opened 2026-06-28
Recurrence on 0.2.84: unbounded summarize_contract_state storm on phantom (interested-but-stateless) contracts (~70/sec, mem to 2G cap)


sanity:
## Summary

A regular (non-gateway) node on **0.2.84** is exhibiting the unbounded `summarize_contract_state` storm that #4473 consolidated and that was reported "verified resolved in prod" (nova on 0.2.80, summarize WARNs back to ~5 / 30 min). On this node it is running at **~70–80 summarize calls/sec**, decoupled from any user activity, against **phantom (interested-but-stateless) contracts** the node holds no local state for — the exact signature described in #4473 and its upstream caveat (#4440).

User-visible effect: laptop fan runs continuously; `freenet` is consistently the top CPU process; memory **peaked at the 2 GB `MemoryMax` cgroup cap**.

This looks like either a regression of the #4475/#4482/#4489 gates between 0.2.80 → 0.2.84, a new ungated summarize caller, or the phantom-interest leak (#4440) now reaching the summarize path via `contract_in_use=true` (see hypotheses).

## Environment

- **Version:** `Freenet version: 0.2.84 (5f1d9d4edd8a)`, build timestamp `2026-06-27T13:10:15Z` (built today)
- **Role:** regular node, `is_gateway = false`, `mode = network`
- **Host:** framework laptop, Arch Linux (Omarchy/Hyprland), kernel 7.0.9
- **Run mode:** `freenet.service` (user systemd unit), `CPUQuota=200%`, `MemoryMax=2G`
- **Gateways:** nova.locut.us:31337, vega.locut.us:31337
- **Uptime at capture:** 11h53m, single Main PID (no restarts), `CPU: 27min` accumulated
- Only **4 `.wasm` contracts on disk** (`~/.local/share/freenet/contracts/`); **7 local WS-API clients** connected on `127.0.0.1:7509`

## Evidence

**1. Storm rate (per-callsite rate-limit counters, cap 30/sec):**
```
summarize_contract_state: dropped 1459 in last 30s (cumulative: 416714)   → ~78 calls/sec at peak
peer_connection:          cumulative 177974 + 118329 (two callsites)
process_network_message:  cumulative 12453
fetch_contract:           present
```
Cumulative `summarize_contract_state` dropped count grew from ~414,865 → ~416,857 during a few minutes of observation. ~417k throttled events over 11h uptime.

[...]

PR #4613, opened 2026-06-28
fix: disable SubscribeHint placement migration [HOLD — LIKELY UNNECESSARY, see body banner]

sanity:
## ⚠️ HOLD — LIKELY UNNECESSARY. DO NOT IMPLEMENT OR MERGE without checking with @sanity first (2026-06-28)

**This PR's premise has been superseded by a later diagnosis. Disabling is now the *fallback*, not the plan.**

The "no placement benefit / act-rate collapsed" conclusion below was a **misdiagnosis**. The migration is not net-negative — it is **starved by a mis-measured admission gate (#4534)**, and the storm cost it was blamed for was the separate #4610 bug (already fixed in #4611, shipped 0.2.86).

Live 0.2.86 telemetry (787 peers): of 635 small-cache (≤320 MB) nodes, the **340 (53.5%) the admission gate refuses ALL have interested-cache occupancy ~0%** — median *raw* occupancy is 98.1% (benign cold LRU fill, not real pressure), while median *interested* occupancy is 0.0%. **100% of those refused nodes would be admitted** under an interested/effective-occupancy gate. The needed metric (`interested_bytes`) is already populated in production, and the in-code shadow counter (`migration_admission_would_change_total` = 353,832 across 407 peers) confirms the flip fires constantly.

**The correct fix is to change the admission gate (`node.rs:1875`) from raw `total/budget` to interested/effective occupancy — NOT to disable the migration.** That work supersedes this PR.

**This PR stays parked as a fallback for ONE scenario only:** after the gate fix ships, if the act-rate rises but `hosted_key_distance` still stays flat (the migration acts but genuinely doesn't deliver). Until that is measured, disabling would throw away a working-but-starved feature.

Full diagnosis: `~/code/tmp/subscribehint-gate-analysis.html`.

---

<sub>Original draft body retained below for the record (its conclusions are superseded by the banner above).</sub>

---

> **DRAFT** — merges AFTER the #4610 fix (PR #4611). This is a separate, follow-on change. Also needs a release to reach the network. Do not merge until #4611 has merged.

## Problem

The #4404 SubscribeHint placement migration is **net-negative in production**. It imposes real, measurable cost while delivering no placement benefit:

- **Cost:** drives the #4610 full-state summarize storm, the #4440 subscription-renewal load, and #4534 module-cache thrash.
- **No benefit:** production telemetry over the re-enabled window (re-enabled at `(0, 2, 80)` in #4511) showed `hosted_key_distance` **flat** (placement did not tighten at all) and the inbound-hint **act-rate collapsed to ~0.05%** (received-but-not-acted).

So the network paid the storm / renewal / thrash cost for essentially zero migrations. Disabling removes that cost with no placement regression. Authorized by Ian.

## Solution

Park the wire-gate floor `SUBSCRIBE_HINT_MIN_VERSION` at `(0, 3, 0)` (was `(0, 2, 80)`), above every shipped 0.2.x release. This single const disables **both** sides of the migration:

- **SEND** — `select_migration_target` → `peer_supports_subscribe_hint` never selects a target (no peer satisfies the floor).
- **RECEIVE** — the `node.rs` inbound `SubscribeHint` handler gates on this node's own version (`0.2.x < 0.3.0`) and ignores all inbound hints.

**Reversible:** lowering the floor re-enables both sides together. The const doc comment and `docs/RELEASING.md` now record that re-enabling is a deliberate, **canary-gated** decision — a passing simulation test is explicitly *not* sufficient; any future re-enable must show `hosted_key_distance` measurably tightening on real peers with the #4610/#4440/#4534 costs bounded. This is documented so the disable is not reflex-reverted.

No sim co-change: simulation tests drive the cascade via `subscribe_hint_floor_override` (set by `SimNetwork::enable_placement_migration`), not this const, so migration-on sim tests still run and pass. The `ConsiderContractMigration` `NodeEvent` emits are left as-is (they now fire a no-op `select_migration_target`; harmless).

## Testing

- Updated the two "active at re-enable floor" pin tests to disabled semantics — they now pin the floor to **exactly** `(0, 3, 0)` (`!supported(0,2,255)` + `supported(0,3,0)`), so an accidental lowering trips the assert.
- Un-ignored the two previously-`#[ignore]`d "deactivated" tests; they assert the disabled semantics and are accurate again at `(0, 3, 0)`.
- Added a new regression-pin test `subscribe_hint_floor_is_parked_at_disabled_value` that asserts the production const equals the disabled floor `(0, 3, 0)` — the loudest tripwire against a reflex-revert to `(0, 2, 80)`.
- All 11 version-gate / inbound-hint-gate unit tests pass (0 ignored).
- `cargo fmt` + `cargo clippy --locked -- -D warnings` clean.

## Open review finding (for decision before un-drafting)

Codex flagged one **[P2]**: parking at `(0, 3, 0)` means the migration will **auto-re-enable when the network reaches version 0.3.0** during a normal minor bump, which contradicts the new "canary-gated re-enable only" promise. The value `(0, 3, 0)` was chosen to match the documented v0.2.74 disable precedent and is above every *shipped* 0.2.x release (network is on 0.2.85), so it is safe today. Options to resolve before merge: (a) raise the park value to a never-reachable sentinel (the codebase already uses `(255, 255, 65535)` as `SIM_MIGRATION_DISABLED_FLOOR` for exactly this), or (b) ensure the re-enable decision is made before 0.3.0 ships. Flagging for the maintainer to decide.

Refs #4404, #4610, #4440, #4534

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1774,19 +1774,22 @@ where
                 .ring
                 .placement_migration_metrics()
                 .record_received();
-            // Placement-migration version gate. The migration is re-enabled at
-            // floor `(0, 2, 80)` (#4499 made it load-safe). The SEND side
+            // Placement-migration version gate. The migration is DISABLED in
+            // production: the floor is parked at `(0, 3, 0)`, above every shipped
+            // 0.2.x release, because it is net-negative (#4610 / #4440 / #4534
+            // costs with no placement benefit — see the `SUBSCRIBE_HINT_MIN_VERSION`
+            // doc comment). The SEND side
             // (`p2p_protoc::peer_supports_subscribe_hint`) gates emission on the
-            // remote peer's version; the RECEIVE path must gate too, so a node on
-            // an older release does not ACT on a hint from an upgraded peer and
-            // keep migration load alive on a peer that predates the load-safe fix.
+            // remote peer's version; the RECEIVE path gates too, so a node does not
+            // ACT on a hint even if some peer were to emit one.
             //
             // The symmetric (sender-version) gate is not cleanly reachable here:
             // the per-connection remote version lives in `P2pConnManager.connections`
             // and is not exposed through the `NetworkBridge` trait, so use this
-            // node's OWN version against the SAME floor the send side uses. A node
-            // on `>= 0.2.80` acts on inbound hints; a pre-floor node ignores them.
-            // Lowering the floor (sim override) re-activates both sides together.
+            // node's OWN version against the SAME floor the send side uses. With the
+            // floor at `(0, 3, 0)` no shipped 0.2.x node acts on inbound hints.
+            // Lowering the floor (sim override, or a deliberate canary-gated
+            // re-enable) re-activates both sides together.
             //
             // Read the floor via `subscribe_hint_floor_override().unwrap_or(...)`,
             // identical to the send side, so a simulation that opts into the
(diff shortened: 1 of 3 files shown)


sanity:
⚠️ **Likely superseded — do not implement/merge.** Later diagnosis shows the migration is *starved by a mis-measured admission gate* (#4534 reads raw cache occupancy ~98% = benign cold LRU fill; real interested-occupancy is ~0%), not net-negative. Live telemetry: 100% of the 340 refused small-cache nodes would be admitted under an interested-occupancy gate. The fix is to change the gate (`node.rs:1875`), not disable. This PR is now a parked fallback only for the case where, after the gate fix, act-rate rises but placement still stays flat. See the body banner + `~/code/tmp/subscribehint-gate-analysis.html`. [AI-assisted - Claude]

sanity:
Closing as superseded / obsolete.

The migration disable this PR was drafted to deliver already shipped through a different, already-merged PR: **#4640** ("disable SubscribeHint placement migration — 0.2.88 hotfix for migration storm", merged 2026-06-30, now main's HEAD). So the network-facing effect this branch aimed for is already in production (0.2.88), and the migration is now going through the full demand-driven-hosting **redesign** (epic #4642), not a one-line const park.

This branch is also a stale `CONFLICTING`/`DIRTY` draft, and its own top banner + my 2026-06-28 comment already flagged it as "likely superseded — do not implement/merge" (the net-negative premise was a misdiagnosis; the real issue was the mis-measured admission gate #4534). The interested/effective-occupancy gate follow-up lives on its own branch, so nothing here is worth keeping.

Reopen only if the redesign decides it needs exactly this const-park as a fallback — but #4640 already covers the disable, so that's unlikely.

[AI-assisted - Claude]

PR #4623, opened 2026-06-28
feat: open browser when permission prompt has no dashboard tab (#3820)

### LINKED ISSUE #3820: feat: open browser/notify user when permission prompt arrives without open tab
## Problem

When a delegate emits `RequestUserInput` and the user has no Freenet gateway tab open in their browser, the permission prompt is never discovered. The shell page JS that polls `/permission/pending` only runs when a gateway page is loaded. Without it, the prompt silently auto-denies after 60 seconds.

This is the expected case when:
- Freenet runs as a systemd user service in the background
- An app triggers a delegate operation via the WebSocket API (not through the browser gateway)
- The user closed all Freenet browser tabs

## Current Behavior

`DashboardPrompter` inserts the pending prompt and waits 60s. If no shell page polls, auto-deny.

## Proposed Solutions (pick one or combine)

1. **Open the browser proactively**: When a permission prompt is created and no shell page WebSocket is connected, use `xdg-open`/`open`/`start` to open the permission page directly in the user's default browser. This requires `DISPLAY`/`WAYLAND_DISPLAY` access (available if the user is logged in to a desktop session but not from a bare systemd service).

2. **D-Bus desktop notification** (Linux): Send a notification via `org.freedesktop.Notifications` D-Bus interface. D-Bus IS available to systemd user services (unlike DISPLAY). The notification body includes the permission page URL.

3. **Tray icon integration**: On Windows/macOS where the tray icon already runs in the desktop session (via the service wrapper), show a tray notification that opens the permission page.

4. **Dashboard homepage polling**: Add a "Pending Permissions" section to the dashboard homepage at `localhost:7509/`. Users who bookmark the dashboard would see pending prompts.

## Context

Introduced in #3818. Found during Codex CLI review. The previous subprocess approach (PR #3811) had the same limitation (no display = auto-deny).

## Priority

Medium -- in practice, delegate operations are triggered by web apps running in the gateway, so the user typically has a tab open. But background/API-triggered operations will silently fail.

[AI-assisted - Claude]

sanity:
## Problem

When a delegate emits `RequestUserInput` and the user has **no Freenet gateway tab open**, the permission prompt is never displayed. The shell-page JS that renders the prompt only runs in an open tab, so the prompt silently auto-denies after `USER_INPUT_TIMEOUT` (60s) with no visible feedback.

This is the expected case when Freenet runs as a background systemd user service and a delegate operation is triggered via the WebSocket API (CLI, background agent, API script) rather than through the browser gateway. The operation just fails cryptically — a complete UX failure for that workflow.

Introduced in #3818; found during Codex review there.

## Approach

A connected gateway tab subscribes to the prompt-event broadcast over SSE (`/permission/events`), with exactly one broadcast receiver per connection. So `prompt_events().receiver_count() == 0` faithfully means "no tab is open to see this prompt".

When a prompt is created with zero subscribers, `DashboardPrompter` opens the **standalone permission page** (`http://<dashboard-authority>/permission/<nonce>`) in the user's default browser. That page is self-contained (server-rendered prompt + buttons that POST the response), so the user can approve/deny directly — it both alerts and lets them act. This is option 1 (the minimal viable solution) from the issue.

Details:
- **Graceful degradation:** on a headless service with no `DISPLAY`/`WAYLAND_DISPLAY` the browser-open fails harmlessly and the existing 60s timeout still governs — behavior is never worse than before.
- **FreeConsole-safe spawn (#3933 pattern):** Windows uses `ShellExecuteW` (not `cmd /c start`); macOS/Linux spawn `open`/`xdg-open` with all three standard handles nulled, since the node may run as a service with no valid inherited stdio (un-nulled `spawn()` fails with os error 6).
- **Correct URL host:** the URL authority is derived from the bound `ws-api` address, targeting the always-bound *primary* listener rather than the best-effort cross-family loopback companion (`server::serve_dual_stack`). A wildcard bind maps to the loopback of its own family (default `::` → `[::1]`, `0.0.0.0` → `127.0.0.1`); a specific interface bind targets that address; IPv6 literals are bracketed. (From Codex review of this PR.)
- **Burst rate-limit:** a 5s cooldown between no-tab opens so a burst of prompts (the background/API scenario this targets) opens one tab per window instead of up to `MAX_PENDING_PROMPTS` (32) tabs. Suppressed prompts fall back to the pre-existing auto-deny, so the cooldown never leaves a prompt worse off.

The decision logic is injected behind a notifier seam so it is unit-testable without spawning a real browser (the prompt-event broadcast is a process global, so the live `receiver_count()` isn't deterministic across parallel tests).

## Testing

New unit tests in `crates/core/src/contract/user_input.rs`:
- `maybe_alert_no_tab_opens_browser_only_when_no_subscribers` — opens iff zero subscribers, with the exact permission-page URL.
- `maybe_alert_no_tab_url_uses_configured_port` — honors a non-default ws-api port.
- `maybe_alert_no_tab_cooldown_suppresses_burst` — burst within the window opens one tab; opens again after the window (paused-time test).
- `dashboard_authority_maps_bind_address_to_reachable_url_host` — wildcard/loopback map to the bound family's loopback (IPv6 → `[::1]`, IPv4 → `127.0.0.1`), specific interface → that address, IPv6 bracketed.
- `open_permission_page_spawn_must_null_all_three_standard_handles` — source-level pin that the macOS/Linux spawn nulls stdin/stdout/stderr (#3933 regression guard).

`cargo fmt`, `cargo clippy --locked -p freenet -- -D warnings`, and the `contract::user_input` unit tests all pass locally.

## Fixes

Closes #3820

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/user_input.rs
+++ b/crates/core/src/contract/user_input.rs
@@ -153,19 +153,126 @@ pub(crate) fn emit_prompt_event(event: PromptEvent) {
     drop(prompt_events().send(event));
 }
 
+/// Action invoked to surface a permission prompt to the user when no dashboard
+/// tab is connected to display it. Receives the standalone permission-page URL.
+///
+/// In production this opens the user's default browser (see
+/// [`open_permission_page_in_browser`]); tests substitute a recording stub so
+/// the decision logic can be exercised without spawning a real browser.
+pub(crate) type NoTabNotifier = Arc<dyn Fn(&str) + Send + Sync>;
+
 /// Opens the user's browser to the permission page on the local dashboard,
 /// then waits for the user to click a button. The HTTP POST handler sends
 /// the response back via a oneshot channel.
 ///
 /// Works from systemd user services because the node serves HTTP (already
 /// running) and the shell page's JS handles browser notification + opening.
+///
+/// When a prompt arrives and no dashboard tab is connected to display it, the
+/// prompt would otherwise silently auto-deny after [`USER_INPUT_TIMEOUT`]; in
+/// that case the prompter opens the standalone permission page directly in the
+/// user's browser so background / API-triggered prompts stay actionable (#3820).
 pub struct DashboardPrompter {
     pending: PendingPrompts,
+    /// Loopback authority of the local dashboard HTTP server,
+    /// `127.0.0.1:<ws-api-port>`. Used to build the permission-page URL opened
+    /// when no tab is connected (#3820).
+    dashboard_authority: String,
+    /// Invoked with the permission-page URL when a prompt has no connected tab.
+    notify_no_tab: NoTabNotifier,
 }
 
 impl DashboardPrompter {
-    pub fn new(pending: PendingPrompts) -> Self {
-        Self { pending }
+    /// Build a prompter that opens the user's browser when no dashboard tab is
+    /// connected. `ws_api_port` is the local HTTP/WS API port the dashboard is
+    /// served on (`config.ws_api.port`).
+    pub fn new(pending: PendingPrompts, ws_api_port: u16) -> Self {
+        Self {
+            pending,
+            dashboard_authority: format!("127.0.0.1:{ws_api_port}"),
+            notify_no_tab: Arc::new(open_permission_page_in_browser),
+        }
+    }
+
+    /// Alert the user out-of-band when a freshly-created prompt has no dashboard
+    /// tab to display it. `subscriber_count` is the number of live SSE
+    /// subscribers to the prompt-event broadcast
+    /// ([`prompt_events()`]`.receiver_count()`); each connected gateway tab
+    /// holds exactly one. Zero means every tab is closed, so the prompt would
+    /// silently auto-deny after [`USER_INPUT_TIMEOUT`] unless we surface it.
+    fn maybe_alert_no_tab(&self, nonce: &str, subscriber_count: usize) {
+        if subscriber_count == 0 {
+            let url = format!("http://{}/permission/{nonce}", self.dashboard_authority);
+            (self.notify_no_tab)(&url);
+        }
+    }
+}
+
+/// Open the standalone permission page in the user's default browser.
+///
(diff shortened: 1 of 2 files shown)


sanity:
## Review summary

Independent external review (Codex / `gpt-5.5`) was run iteratively against `origin/main`; Gemini was not invoked (its individual free tier is deprecated / `IneligibleTierError`). I also applied the four review lenses (code-first, testing, skeptical, big-picture) by reading the checked-out code, with attention to the cross-platform spawn and the no-browser-available case. All findings are addressed:

**Codex findings (all fixed):**
- **[P2] Loopback-only URL** — the prompt URL was hardcoded to `127.0.0.1`. Fixed by deriving the authority from the bound `ws-api` address.
- **[P2] Best-effort companion socket** — a node bound to IPv6 wildcard (`::`, the default) where the IPv4 companion fails would refuse `127.0.0.1`. Fixed: `dashboard_authority` targets the always-bound *primary* listener (wildcard → own-family loopback, default `::` → `[::1]`), never the best-effort companion. Verified against `server::serve_dual_stack` / `companion_bind_addr`.
- **[P2] Stray `capture` binary** — a 3.9 MB build artifact was accidentally staged by `git add -A`. Removed; the branch now contains only the two source files.
- Final Codex pass: "I did not identify any discrete correctness, security, or maintainability issues introduced by this diff."

**Claude Rule Review:** no blocking warnings. The one INFO note (null-stdio pin test's source-scrape boundary) is addressed by making the adjacency assumption explicit in a comment.

**Self-review notes:**
- Cross-platform spawn follows the #3933 FreeConsole-safe pattern (ShellExecuteW on Windows; nulled stdio on macOS/Linux), pinned by a source-scrape test.
- No-browser-available case degrades gracefully: `spawn()` fails harmlessly and the existing 60s timeout still governs.
- Burst case is rate-limited (5s cooldown) so the background/API scenario can't open up to `MAX_PENDING_PROMPTS` tabs; suppressed prompts fall back to the pre-existing auto-deny.

[AI-assisted - Claude]

PR #4364, opened 2026-06-07
fix(get): fall back to configured gateways when the ring is empty

sanity:
## Problem

A node with an empty ring (`connections_by_location` has no entries) but live transient transport connectivity fails **every GET instantly with NotFound — zero wire messages sent**. All GET peer-selection points consult only ring connections via `k_closest_potentially_hosting`, and every exhaustion path was silent. Production analog: a freshly started node fails all GETs for its first minutes, until ring promotion completes.

This is the primary defect behind the nightly `test_get_reliability_diagnostic` failing three nights running at 18.3% (#4361): during the test's op window most nodes have empty rings, so not one GET traversed the network — every "success" was a local cache hit, and the metric itself multi-counted outcomes.

## Solution

**Bounded bootstrap fallback (defect 1).** When `connection_count() == 0`, the GET selection points — client initial pick, sub-op initial pick, client advance, relay advance — fall back to a configured gateway, skipping self, already-tried addrs, and the request's visited bloom (cross-hop loop prevention). Key insight: wire delivery needs no new plumbing — `OutboundMessageWithTarget` resolves targets by socket addr against the transport map (which includes transient connections) and re-dials configured gateways if the transport lapsed. This mirrors the #3219 zero-connection CONNECT re-bootstrap shape, gated on the same condition, so normal routing is untouched. The relay fallback stays within `MAX_RELAY_RETRIES = 1`, so the 3^HTL fan-out guard is unaffected.

**Silent exhaustion paths now log (defect 3):** empty-candidate exhaustion, addressless candidate selected (`peer.socket_addr()` holes), the previously-discarded `Exhausted` cause, and the addressless-candidate filter bypass in `k_closest_potentially_hosting`.

**Test metric integrity (defect 4):** new `tracing::summarize_get_outcomes_per_tx` dedups GET outcomes per attempt transaction — a failed attempt was registered twice on the originator (relay-direct + loopback-Response inbound), and multi-hop outcomes registered once per hop. New `EventKind::get_request_htl` enables dispatched-vs-scheduled accounting (originator loopback Requests carry `htl == max_htl`). The diagnostic test now asserts a dispatch floor and a network-traversal floor (`hop_count >= 1` successes), so local cache hits can never again masquerade as routing success. Sibling tests (`with_latency`, `with_churn`) updated in lockstep.

[...]

Issue #4363, opened 2026-06-07
fix: PUT finalizes far from contract location on sparse bootstrap topology, with no re-replication toward the contract neighborhood


sanity:
## Summary

Split out of #4361 (defect 5).

In the deterministic `test_get_reliability_diagnostic` run (100 nodes + 3 gateways, seed `0x3570_D1A6_0001`), the PUT routed *away* from the contract location on the bootstrap topology:

- Contract location: **0.211**
- The PUT chain reached ring distance **0.03** from the contract location by hop 3
- It then wandered and finalized at location **0.868** (ring distance 0.343 from the contract location)

There is no later re-replication toward the contract neighborhood, so even after the topology converges, greedy GETs descending toward 0.211 don't find the replicas placed along the (bad) PUT path. The 11 PUT-path replicas are badly placed for greedy routing.

## Impact

- Contracts PUT during network bootstrap (or any sparse-topology window) can remain effectively unreachable by greedy GET even after the ring heals — only path-cache hits and subscription propagation serve them.
- In #4361's failing runs this capped achievable GET success even for nodes with healthy rings.

## Possible directions (needs investigation)

- Post-convergence re-replication: nodes detect they host a contract whose location is far from their own and push a replica toward the contract neighborhood (e.g. piggyback on subscription/NEIGHBOR_HOSTING maintenance).
- PUT-time: when the routing chain moves *away* from the target location for k consecutive hops, prefer terminating at the closest-seen node rather than the last node.

## Related

- #4361 — the diagnostic test that exposed this (its `network_successes` floor will benefit from better placement)
- #3570 — original GET-reliability investigation

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-06-09T14:09:02.908736862+00:00
Issue hash: `b5b3c238`

## Value Assessment

This issue is still open and unaddressed. Here is my assessment:

**Relevance:** This issue remains fully relevant. The investigation confirms that neither of the two proposed fixes exists in the current codebase: there is no post-convergence re-replication mechanism (no background process detects and corrects badly-placed replicas), and PUT routing termination in `op_ctx_task.rs` is HTL-only with no logic to detect or counteract backwards routing toward the contract location. The issue was filed in 2025 and no subsequent commits have addressed the underlying routing termination or re-replication logic.

[...]

Issue #4166, opened 2026-05-19
GET on empty peer set: fail-fast + active reroute on reconnect + missing pre-routing telemetry


sanity:
## Background

#4154 documented three distinct symptoms in the user-reported 62 s GET stall:

1. **Parked driver after peer disconnect** — fixed by #4164 (event-driven wake via `handle_orphaned_transactions`).
2. **Fail-fast on empty peer set at issue-time** — NOT yet addressed.
3. **Active reroute when new peers connect** — NOT yet addressed.
4. **Missing pre-routing telemetry** — the original report observed zero `get_request` events on the stalled attempt versus 9 on the retry. NOT yet addressed.

This issue tracks the remaining slices.

## (2) Fail-fast on empty peer set

`drive_client_get_inner` (`crates/core/src/operations/get/op_ctx_task.rs:243-256`) currently falls back to `own_location()` silently when `k_closest_potentially_hosting` returns empty. The relay driver then either short-circuits with `NotFound` (originator advances quickly) or sits waiting. We should:

- Surface a structured "no routable peer" error to the client driver immediately, OR
- Hold the op briefly with a bounded timeout (e.g. 2 s) waiting for `connection_manager` to gain a peer, then either succeed (peer arrived) or fail-fast with the structured error.

## (3) Active reroute on reconnect

When `connection_maintenance` adds a new peer while drivers are parked / advancing, no signal goes to those drivers — they continue using the `skip_list` + retry cadence they entered with. A "peer pool grew" notification could let drivers re-evaluate routing before the next attempt's `attempt_timeout` fires.

Caveat: the wake mechanism from #4164 already handles disconnect-side recovery. This issue is specifically about the "peers I never saw at issue-time just arrived" case.

## (4) Pre-routing telemetry

`crates/core/src/operations/get/op_ctx_task.rs:2091` emits `get_request` only AFTER `relay_advance_to_next_peer` returns Some — i.e. AFTER routing succeeds. The originator's `build_request` at line 699-712 only stamps `request_sent_at`. Add a structured event capturing:

- Op created (originator) — tx, instance_id, hop count target
- Peer selection result (Some/None + reason)
- Each forward emission (already covered by `get_request`)

This would have made the original #4154 diagnosis 10x faster: telemetry would have shown "op created at T+0, peer selection returned None at T+0, no peer reconnected until T+13s, no re-evaluation until original timeout at T+60s."

## Acceptance

[...]

PR #4635, opened 2026-06-29
fix: implement gateway fallback for empty ring in put, update and subscribe operations

### LINKED ISSUE #4365: audit: PUT and UPDATE bootstrap behavior on an empty ring (GET fixed in #4364)
## Summary

#4364 fixed GET's empty-ring blind spot (#4361 defect 1) with a bounded fallback to configured gateways. The testing review of that PR noted the other operations deserve the same audit:

- **GET** — fixed in #4364 (bounded gateway fallback at all selection sites).
- **SUBSCRIBE** — has its own any-connected-peer fallback (`subscribe.rs` `prepare_initial_request` scans raw `get_connections_by_location` when `k_closest` is empty, and fail-fasts with a distinct `NoHostingPeers` error) — but that fallback scans *ring* connections only, so a node with an empty ring and a live transient gateway connection still fail-fasts. Worth deciding whether SUBSCRIBE should get the same gateway fallback.
- **PUT** — `drive_client_put_inner` falls back to `own_location()` and stores locally, which is semantically acceptable (the contract exists locally and can propagate later) — but verify the later propagation actually happens once the ring forms, and whether a bootstrap-window PUT should also seed a configured gateway so the contract is reachable by others immediately.
- **UPDATE** — fire-and-forget end-to-end; check what happens to an UPDATE issued during the bootstrap window (silently dropped vs applied locally and re-broadcast later).

## Why it matters

Production analog (same as #4361 defect 1): everything a freshly started node does in its first minutes happens in the empty-ring window. GET now works there; the other three ops should either work or fail with an explicit, retryable error — never silently misbehave.

## Related

- #4361 / #4364 — GET fix and the review that prompted this audit
- #4166 — fail-fast "not bootstrapped" client error (the explicit-error half of the design space)
- #4362 — bootstrap acceptance collapse (why the empty-ring window is minutes long)

[AI-assisted - Claude]

netsirius:
## Problem

A freshly started node has an empty ring (`connections_by_location` is empty) for its first minutes but already holds a live transient connection to its gateway. PUT/UPDATE/SUBSCRIBE peer-selection consulted only the ring, so during that window they ignored the gateway. #4364 fixed this for GET; the #4365 audit asks the other three to meet the same bar — in the bootstrap window an op must either **work** or **fail with an explicit, retryable error**, never silently misbehave.

Behavior on an empty ring before this change:

- **PUT** — the relay stored the contract locally and finalized locally, so the client got a *successful* `LocalCompletion` while the contract never reached the network (no re-replication on this path; see #4363). A false success.
- **UPDATE** — if hosting, applied locally and broadcast to an empty ring (reached no one); if not hosting, returned `NoHostingPeers`. The hosting case has PUT's no-propagation gap.
- **SUBSCRIBE** — already fail-fasts with `NoHostingPeers`, but its fallback scans only ring connections, so a node with a live transient gateway connection still gave up.

Note: the fix is **not** in `drive_client_put_inner` (its `own_location()` is only retry bookkeeping); the real empty-ring decision point is the relay's next-hop selection.

## Solution

When `connection_count() == 0`, each op's real decision point falls back to a configured gateway instead of finalizing locally / failing — the same bounded fallback #4364 added to GET.

- **Shared helper** — `select_bootstrap_gateway` (pure) and `bootstrap_gateway_target` (`OpManager` wrapper) lifted out of GET into `operations/bootstrap.rs`; all four ops now share one source of truth. GET behavior unchanged.
- **PUT** — `drive_relay_put` / `drive_relay_put_streaming` forward via the gateway on an empty ring. The local store already ran, so the originator keeps its replica *and* the gateway relays the PUT toward the contract location. Bypasses the #4363 terminus guard (no overshoot to guard against on an empty ring).
- **UPDATE** — `drive_client_update` uses the gateway as target. Hosting → applies locally and forwards (propagates); not-hosting → the existing missing-params auto-fetch + retry primes the store from the gateway.
- **SUBSCRIBE** — `prepare_initial_request` tries the gateway before `NoHostingPeers`.

Wire delivery needs no new plumbing: `OutboundMessageWithTarget` resolves targets by socket addr against the transport map and re-dials configured gateways, independent of the message variant.

Boundary: a node that hasn't yet learned its own address (phase 1) keeps its current behavior; the fallback targets the dominant phase-2 window (address known, ring still empty). PUT's `current_target` stays `own_location()` there — unlike GET it is only driver-side telemetry, so the divergence is cosmetic (documented in code).

## Testing

- **Selection logic (executed):** unit tests in `operations/bootstrap.rs` assert empty-ring activation, self-skip, tried/visited exclusion, inactive-when-ring-non-empty, and no-configured-gateways → `None`.
- **Call sites (source pins):** one per op, asserting both the call and the correct exclusion predicate (PUT `new_skip_list.contains`, UPDATE `addr == sender_addr`, SUBSCRIBE `visited.probably_visited`), so a no-op predicate trips the build.
- **No regression:** for a lone gateway the fallback is a strict no-op (no non-self gateway → `select_bootstrap_gateway` returns `None`); `isolated_node_regression.rs` covers this on CI.
- **Delivery premise** verified by code inspection (op-agnostic `OutboundMessageWithTarget`). `cargo fmt`, `clippy -D warnings`, and touched-module suites clean.
- **Adversarial audit:** every empty-ring decision point across the four ops was mapped and skeptically reviewed; it confirmed the routing is correct and complete for the originator scope, that the `connection_count() == 0` gate keys only on promoted ring connections (transient gateway connections are a disjoint map), and surfaced the residuals documented below.

**Deferred (follow-up):** an end-to-end behavioral test where a non-gateway node in its empty-ring window PUTs and a second node GETs the contract. A probe confirmed a small static sim can't reproduce the state — a lone node promotes its gateway into the ring (`connection_count()` settles at 1, not 0). The realistic shape is a multi-node simulation diagnostic with empty-ring windows occurring naturally, with a revert-to-verify gate — the way #4364's GET diagnostic worked.

## Out of scope / follow-ups

The fix targets the **originator** bootstrap window. An adversarial audit of every empty-ring decision point across the four ops confirmed the routing is correct and complete in scope — and that adding the fallback to any *other* site would be either harmful (cross-gateway bounce / duplicate store / placement regression) or a separate concern. The residuals (all already fail explicitly in scope — none silently misbehave):

- **No-usable-gateway hosting case** — a node that hosts the contract and has *no* configured gateway can still report success on PUT/UPDATE (nowhere to propagate; GET's isolated-no-gateway case is symmetric). Outside #4365's premise (which assumes a live transient gateway connection); the honest fix is a symmetric fail-fast "not bootstrapped" error across all ops → #4166.
- **Relay-forward / retry parity** — UPDATE's and SUBSCRIBE's relay-forward paths (`drive_relay_request_update`, `drive_relay_subscribe`) and SUBSCRIBE's retry (`advance_to_next_peer_impl`) are not gateway-aware. They already fail explicitly (`NoHostingPeers`/`NotFound`) and are out of scope (a relay holds a connection from its sender). Extending them is **not** free parity: `UpdateMsg::RequestUpdate` carries no HTL/visited bloom, so a gateway fallback there would risk a bounce until loop-prevention is added.
- **PUT htl-gating asymmetry** — `drive_relay_put`'s empty-ring arm can forward to a gateway at `htl == 0` while the streaming path gates on `htl > 0`. Benign (single bounded hop, no false success, better placement than this empty-ring node) and documented in-code; harmonizing has no correctness benefit.
- **End-to-end propagation behavioral test** (multi-node simulation diagnostic) — deferred; per-op wiring ships with source-grep pins + pure-helper unit tests.
- Bootstrap acceptance collapse (#4362), PUT placement/re-replication (#4363).

## Fixes

Closes #4365

### DIFF
--- a/apps/freenet-ping/Cargo.lock
+++ b/apps/freenet-ping/Cargo.lock
@@ -725,6 +725,12 @@ dependencies = [
 
 [[package]]
 name = "const-oid"
+version = "0.9.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
+
+[[package]]
+name = "const-oid"
 version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
@@ -1104,6 +1110,7 @@ dependencies = [
  "cfg-if",
  "cpufeatures 0.2.17",
  "curve25519-dalek-derive",
+ "digest 0.10.7",
  "fiat-crypto",
  "rustc_version",
  "subtle",
@@ -1196,6 +1203,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "der"
+version = "0.7.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7c1832837b905bbfb5101e07cc24c8deddf52f93225eee6ead5f4d63d53ddcb"
+dependencies = [
+ "const-oid 0.9.6",
+ "zeroize",
+]
+
+[[package]]
 name = "deranged"
 version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1255,7 +1272,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4850db49bf08e663084f7fb5c87d202ef91a3907271aff24a94eb97ff039153c"
 dependencies = [
  "block-buffer 0.12.0",
- "const-oid",
+ "const-oid 0.10.2",
  "crypto-common 0.2.1",
  "ctutils",
 ]
@@ -1404,6 +1421,30 @@ name = "dyn-clone"
 version = "1.0.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
+
+[[package]]
+name = "ed25519"
+version = "2.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115531babc129696a58c64a4fef0a8bf9e9698629fb97e9e40767d235cfbcd53"
+dependencies = [
+ "pkcs8",
+ "signature",
+]
+
+[[package]]
+name = "ed25519-dalek"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70e796c081cee67dc755e1a36a0a172b897fab85fc3f6bc48307991f64e4eca9"
+dependencies = [
+ "curve25519-dalek",
+ "ed25519",
+ "serde",
+ "sha2 0.10.9",
+ "subtle",
+ "zeroize",
+]
 
 [[package]]
 name = "either"
@@ -1634,7 +1675,7 @@ dependencies = [
 
 [[package]]
 name = "freenet"
-version = "0.2.82"
+version = "0.2.90"
 dependencies = [
  "aes-gcm",
  "ahash",
@@ -1657,6 +1698,7 @@ dependencies = [
  "delegate",
  "directories",
  "dirs",
+ "ed25519-dalek",
  "either",
  "event-listener",
  "flatbuffers 25.12.19",
@@ -3996,6 +4038,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 
 [[package]]
+name = "pkcs8"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f950b2377845cebe5cf8b5165cb3cc1a5e0fa5cfa3e1f7f55707d8fd82e0a7b7"
+dependencies = [
+ "der",
+ "spki",
+]
+
+[[package]]
 name = "pkg-config"
 version = "0.3.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4986,6 +5038,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "signature"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77549399552de45a898a580c1b41d445bf730df867cc44e6c0233bbc4b8329de"
+dependencies = [
+ "rand_core 0.6.4",
+]
+
+[[package]]
 name = "simd-adler32"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5062,6 +5123,16 @@ dependencies = [
  "gobject-sys",
  "libc",
  "system-deps",
+]
+
+[[package]]
+name = "spki"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d91ed6c858b01f942cd56b37a94b3e0a1798290327d1236e4d9cf4eaca44d29d"
+dependencies = [
+ "base64ct",
+ "der",
 ]
 
 [[package]]
@@ -7419,6 +7490,7 @@ version = "1.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 dependencies = [
+ "serde",
  "zeroize_derive",
 ]
 
--- a/crates/core/src/operations.rs
+++ b/crates/core/src/operations.rs
@@ -13,6 +13,7 @@ use crate::{
     ring::{Location, PeerKeyLocation, RingError},
 };
 
+pub(crate) mod bootstrap;
 pub(crate) mod connect;
 pub(crate) mod get;
 pub(crate) mod op_ctx;
--- a/crates/core/src/operations/bootstrap.rs
+++ b/crates/core/src/operations/bootstrap.rs
@@ -1,0 +1,192 @@
+//! Bootstrap-window gateway fallback shared across operations (#4361 / #4365).
+//!
+//! A freshly started node has an empty ring (`connections_by_location` has no
+//! entries) for its first minutes, yet already holds a live transient transport
+//! connection to its gateway. Operation peer-selection consults only the ring,
+//! so without a fallback every op in that window either fails or finalizes
+//! locally without ever using the transient gateway connection.
+//!
+//! GET introduced the bounded gateway fallback in #4364; this module lifts the
+//! selection core out of GET so PUT, UPDATE, and SUBSCRIBE reuse exactly the
+//! same rule at their own decision points (#4365).
+
+use std::net::SocketAddr;
+
+use crate::node::OpManager;
+use crate::ring::PeerKeyLocation;
+
+/// Pure selection core for the bootstrap gateway fallback (#4361).
+///
+/// Returns the first configured gateway that is not this node and not
+/// excluded by the caller's skip predicate — but ONLY when the ring is
+/// empty (`ring_connection_count == 0`). A non-empty ring means normal
+/// routing had ring entries to consider (its candidates may still all
+/// be filtered as transient/visited, but that exhaustion is genuine),
+/// so the fallback must stay out of the way.
+///
+/// Selection is deliberately deterministic (config order), unlike the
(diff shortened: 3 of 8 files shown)


iduartgomez (approved):

iduartgomez (commented):
## Comprehensive PR Review: #4635

### Summary
- **PR Title:** fix: implement gateway fallback for empty ring in put, update and subscribe operations
- **Type:** fix
- **CI Status:** passing (all checks green on HEAD `9e0acf013`)
- **Linked Issues:** Closes #4365 (maintainer-authored audit naming PUT/UPDATE/SUBSCRIBE)
- **Review tier:** Full (touches operations state machines + routing — high-risk surfaces)
- **Reviewers run:** code-first, skeptical, testing, big-picture, plus a routing-loop/wire-compat lens. **External model (codex) was unavailable** (`refresh_token_reused` — needs re-login; gemini not installed), so per policy a diverse-Claude-lens fallback was substituted: 5 independent lenses, each blind to the others. Findings below are reconciled and each cited line was opened and verified.

---

### Code-First Analysis
**Independent Understanding:** Generalizes GET's empty-ring gateway fallback (#4364) to PUT/UPDATE/SUBSCRIBE. New `operations/bootstrap.rs` holds `select_bootstrap_gateway` (pure) + `bootstrap_gateway_target` (OpManager wrapper), lifted verbatim from GET. When `connection_count() == 0` each op forwards to a configured gateway instead of finalizing locally / failing.
**Stated Intent:** Matches the code faithfully — the description even pre-discloses the htl-gating asymmetry, the SUBSCRIBE ordering, the cosmetic `current_target` divergence, and the out-of-scope residuals.
**Alignment:** Matches, in scope. Closes #4365 within its stated originator scope.
**Gaps:** Two under-statements in the description (non-blocking): (1) SUBSCRIBE now routes to the gateway before the `has_contract → LocallyComplete` branch too, not just before `NoHostingPeers` (`subscribe.rs:288` precedes `:299/:308`); (2) the `apps/freenet-ping/Cargo.lock` churn (+76/-2, ed25519/der/const-oid) is incidental fallout from the merged baked-key commits (`b0a1b0683`), unrelated to this fix.

---

### Testing Assessment
**Coverage Level:** Selection core well covered; one call-site pin is defective.

| Unit | ✅ | 5 pure tests in `bootstrap.rs:120-191` — real fail-without-fix coverage of the selector (empty-ring activation, self-skip, tried-exclusion, inactive-when-non-empty, no-gateways→None) |
| Source-pin (per op) | ⚠️ | SUBSCRIBE + UPDATE + GET pins sound; **PUT pin broken** (see Must Fix #1) |
| Integration | N/A | — |
| Simulation/E2E | ❌ deferred | end-to-end "empty-ring node PUTs, second node GETs" deferred — static sim can't reproduce (lone node self-promotes its gateway → `connection_count()` settles at 1, not 0) |

**Regression Test:** Present for the selector; the per-op call-site guard is defective for the streaming PUT path (the single highest-value regression — false PUT success on empty ring — has no working guard at that site).
**Missing Tests:** A behavioral test driving `drive_relay_put` / `drive_relay_put_streaming` next-hop selection below the sim with `connection_count()==0` + a configured gateway, asserting the gateway is chosen (covering both htl arms).

---

### Skeptical Findings
**Risk Level:** Low (no correctness bug introduced); one test-integrity defect.

| Streaming-PUT source-pin guards nothing | **Must Fix** | `put/op_ctx_task.rs:3292` | `\nasync fn ` body-end scan finds no unindented `async fn` after `drive_relay_put_streaming<CB>` (`:2388`), so "body" runs to EOF and swallows `mod tests` incl. the pin's own assertion literal (`:3321`). **Verified empirically:** neutering the streaming call site (`:2469`) to `None` leaves the test passing. Non-streaming arm (`:1172→1799`) IS correctly bounded. |
| htl==0 PUT comment justification imprecise | Consider | `put/op_ctx_task.rs:1336-1345` | Comment says the htl==0 forward is safe because the gateway "finalizes on its populated ring." Real termination guarantee in an all-empty-ring chain is **skip_list exhaustion of configured gateways**, not htl and not populated ring. Behavior is bounded & benign; the stated *reason* is incomplete. |
| UPDATE pin sound only by function ordering | Consider | `update/op_ctx_task.rs:2123` | Same fragile `include_str!`+`\nasync fn` pattern; passes today only because `drive_client_update`'s body is bounded before the test module. Latent same flaw — harden with the GET-style `production_source()` truncation. |

**No-bug verifications (investigated, found safe):**
- `connection_count()` sums only `connections_by_location` (`ring/connection_manager.rs:1363`) — disjoint from the transient transport map, so the gate fires precisely in the bootstrap window as claimed.
- Anti-bounce per protocol all confirmed: PUT carries monotonic `new_skip_list` (self+upstream), SUBSCRIBE marks gateway `visited` and propagates the bloom (`subscribe.rs:322`), UPDATE relay builds `[self, sender]` skip per hop. No path bounces back to originator. UPDATE bootstrap fallback is on the **client** driver only, not the HTL-less relay driver, so the admitted `RequestUpdate`-has-no-bloom risk is NOT reached by the empty-ring path.
- `select_bootstrap_gateway` excludes self (`own_addr`) + caller's exclusion set; each call site passes a real predicate (`bootstrap.rs:45-52`).
- `OutboundMessageWithTarget` resolves by socket addr and re-dials configured gateways before promotion, variant-agnostically (`p2p_protoc.rs:1526-1642`) — delivery to an unpromoted transient gateway works.
- Step-1 local store precedes the forward decision in both relay PUT paths → gateway forward never produces a false success and the originator keeps its replica.

---

### Big Picture Assessment
**Goal Alignment:** Yes — closes the maintainer-authored audit #4365 within originator scope; consolidates four ops onto one source of truth (reduces future drift).
**Anti-Patterns Detected:** None — no `#[ignore]`, no weakened assertions, no swallowed errors. Adds explicit telemetry.
**Removed Code Concerns:** GET `-150` is a **verified clean extraction** — helpers moved verbatim to `bootstrap.rs`, GET's 4 call sites byte-identical, GET source-pin survives. Net test coverage strictly increases (4 GET unit tests → 5 in bootstrap.rs + 3 new op pins). Nothing lost.
**Scope Assessment:** Focused, apart from the incidental `apps/freenet-ping/Cargo.lock` ed25519 sync (harmless, worth a one-line PR note).

---

### Documentation
- Code docs: complete (in-code comments thorough — one imprecise justification flagged above).
- Architecture/rules: up-to-date; no `.claude/rules`/`AGENTS.md` reference rendered stale by the move. Optional: note `operations/bootstrap.rs` as the canonical empty-ring fallback in `.claude/rules/operations.md`.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
1. **Fix the broken `relay_put_drivers_use_bootstrap_gateway_fallback` pin** (`put/op_ctx_task.rs:3292`). As written it does not guard the streaming PUT call site — empirically, deleting that fallback leaves the test green, silently reintroducing the exact false-success bug this PR fixes. Replace the `"\nasync fn "` EOF-scan with the GET-style approach: a `production_source()` helper that truncates at `#[cfg(test)]` before scanning (and/or brace-matched `extract_fn_body`). Reference pattern already in-repo: `get/op_ctx_task.rs:3820`/`:3830`.

#### Should Fix (Important)
2. Add one **behavioral** test (not a source-pin) for the false-PUT-success regression: drive both `drive_relay_put` and `drive_relay_put_streaming` next-hop selection with `connection_count()==0` + a configured gateway, assert the gateway is the chosen target — covering the htl==0 / htl>0 asymmetry the comments flag. This is the test that fails-without-fix; the pins are a proxy and one is defective.
3. Harden the UPDATE pin (`update/op_ctx_task.rs:2123`) with the same `production_source()` truncation so it can't rot under a future function reorder.

#### Consider (Suggestions)
4. Correct the htl==0 comment (`put/op_ctx_task.rs:1336-1345`) — termination is skip_list-bounded, not "populated ring"-bounded.
5. PR-body note that `apps/freenet-ping/Cargo.lock` is an incidental ed25519 lockfile sync from `b0a1b0683`, so the crypto-dep churn in a bootstrap PR isn't surprising.
6. PR-body note that SUBSCRIBE now also reorders ahead of the local-hosting `LocallyComplete` case (intended — a local holder must still join the subscription tree per `subscribe.rs:207-214`; the old local-complete was effectively a missed-updates bug). Adds gateway-timeout latency when the gateway is unreachable — confirm that tradeoff is intended.

---

### Verdict
**HEAD SHA reviewed:** 9e0acf0133656705766d07c214c56ad38106cf46

The production code is sound: no high/medium correctness bug, anti-bounce verified on every path, the GET extraction is clean, net coverage increases. The one blocking item is a **test-integrity defect** — the streaming-PUT source-pin guards nothing (verified empirically), so the most important regression can be reintroduced without CI catching it. Fixing the pin (Must Fix #1) is a small, isolated test-only change; a diff-of-the-diff re-check on that fix is sufficient rather than a full re-review. Should-Fix #2 (a real behavioral test) is the durable closure for the deferred coverage. Note: this PR is already maintainer-approved (@iduartgomez); the production behavior is fine — the asks are about test guards, not the fix itself.

[AI-assisted - Claude]

Issue #4630, opened 2026-06-29
Threat-model the module-cache eviction + migration-admission mechanism against a performance-degradation attacker (is it principled enough?)


sanity:
## Summary

v0.2.87 shipped a change to the contract module-cache eviction/admission mechanism (interest-tiered eviction default-on + the interested-occupancy admission gate, #4625/#4628). The mechanism's behavior is driven by **network-controllable inputs** — subscriptions (the "interest" signal) and inbound SubscribeHints (migration admission) — so an adversary may be able to manipulate it to **degrade network performance**: cache thrash, compile-DoS, forced eviction of useful contracts, cache-pinning.

The design has accreted incrementally (raw-occupancy gate → interested-occupancy gate → two-tier eviction → deferred in-flight/churn fixes) rather than from a principled adversarial model. **Hypothesis: it is not yet principled enough to resist a performance-degradation attacker.** This issue is to threat-model it properly and decide where it must be made robust *by construction*.

## The mechanism + its attacker-controllable inputs

- **Interest signal** (`Ring::contract_in_use` = local client subscriptions OR downstream subscribers) — decides what's "hot" (protected) vs "cold" (evicted first). Attacker controls subscriptions, and the signal is **origin/topology-based** (who-subscribed-through-whom), which is cheap to forge by choosing where to subscribe from.
- **Admission gate** (`migration_admission_decision`: admit an inbound SubscribeHint when interested-occ < 90% under tiered eviction) — each admission triggers a directed-subscribe → fetch + **WASM compile** + host. Attacker can craft SubscribeHints directly.
- **Two-tier eviction** (cold-first) — eviction order is a function of the (forgeable) interest signal.
- **Compile cost** — accessing an evicted module recompiles it (the #4534 thrash), on the single contract-handling loop.

## Attack vectors to evaluate

1. **Compile-DoS via SubscribeHint flood.** An attacker is NOT bound by the honest producer cap (#4583 / `MIGRATION_SCAN_CAP_PER_NEW_PEER`) — they craft hints directly. While interested-occ < 90% the gate admits each hint → fetch + compile → exhausts the victim's contract loop / CPU. Is there any **receive-side rate-limit** on admitted migrations? (The deferred in-flight/reserved-bytes accounting bounds the *cache* overshoot; does anything bound the *compile rate*?)
[...]

PR #4640, opened 2026-06-30
fix: disable SubscribeHint placement migration (0.2.88 hotfix for migration storm)

sanity:
**Approved: Ian (@sanity) has authorized merge + the 0.2.88 release. Full `/freenet:pr-review` (external Codex + adversarial multi-lens) passed clean.**

## Problem
0.2.87's SubscribeHint placement migration is over-aggressive: fresh nodes pull ~1000 contracts with no locality pattern, saturating bandwidth (user Ivvor: a 4GB VPS stuck restarting). Root cause: migration was re-enabled in v0.2.80 (#4404 / #4511), then amplified by fleet-wide upgrade saturation plus the v0.2.87 admission-gate loosening (#4625 / #4628). Every node nudges every closer neighbor to subscribe-and-host each contract it holds, on every new connection — the "holding-driven placement push" anti-pattern in `.claude/rules/hosting-invariants.md`.

## Approach
Disable the migration via an explicit `PLACEMENT_MIGRATION_ENABLED = false` kill switch guarding **both** the emit sides (`consider_contract_migration`, `consider_migration_for_new_peer`) and the inbound-act side (the `SubscribeHint` receive handler that starts a directed subscribe). `SUBSCRIBE_HINT_MIN_VERSION` is left untouched, so the wire-compat doctrine (frozen floor) is preserved — disabling emission only narrows what is sent on the wire, so the change is wire-safe.

The gate is **override-aware** so it does not break the simulation harness: it reads each node's `subscribe_hint_floor_override`.
- **Production** nodes never set the override (`None`) → the kill switch decides → migration **OFF**.
- **Simulation** nodes always set the override (`Some(..)`, defaulting to an unreachable floor; `enable_placement_migration` lowers it to `(0,0,0)`) → the gate returns `true` so the kill switch does not blanket-block, and the existing version-floor checks drive migration on/off exactly as before. The placement-migration sim tests (`test_contract_migrates_*`, `test_placement_migration_at_scale_*`, the fanout-liveness scenarios) keep their coverage of the mechanism for the eventual re-enable.

This reverts production to the known-good pre-0.2.80 "migration off" baseline and moves toward the demand-driven hosting target invariants (placement migration is a listed anti-pattern in `.claude/rules/hosting-invariants.md`). Placement is an optimization, not the retrieval-correctness path, so finding contracts cannot become less correct.

## Testing
- Renamed `placement_migration_active_at_reenable_floor` → `subscribe_hint_wire_floor_frozen_at_0_2_80`: it now pins **only** the frozen wire floor (migration activation is gated separately by the kill switch), and its doc was updated to say so.
- Added `placement_migration_disabled_by_flag` — pins production (`None` → off) vs simulation (`Some` → defers).
- Added `placement_migration_emit_sites_gated_by_flag` and `placement_migration_inbound_act_gated_by_flag` — source-scrape pins (the project-idiomatic pattern in this module; there is no lightweight `P2pConnManager` harness) asserting every emit/act site early-returns on the gate before any nudge / `start_directed_subscribe`.
- `cargo fmt` + `cargo clippy --all-targets -D warnings` clean on changed code; migration unit tests pass; the migration simulation test `test_contract_migrates_to_close_cluster_resolving_get_dead_end` still passes (override-aware gate keeps sim behavior intact).

## Tradeoff (flag for reviewers)
Disabling migration reopens GET/SUBSCRIBE dead-ends — the chronic, tolerable condition that existed pre-0.2.80. The planned terminal-advertisement-consult work will cover those.

Refs #4404 #4630 #4625 #4628 #4440

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1767,6 +1767,32 @@ where
             return Ok(());
         }
         NetMessageV1::SubscribeHint(hint) => {
+            // Disabled 0.2.88: over-aggressive migration storm (#4630-adjacent).
+            // do NOT re-enable aggressive placement migration; see
+            // .claude/rules/hosting-invariants.md (anti-pattern: holding-driven
+            // placement push). A PRODUCTION node (no floor override) ignores
+            // inbound hints entirely so a fresh 0.2.88 node is never pulled into
+            // the storm by a not-yet-upgraded 0.2.87 neighbor that still emits
+            // hints. A SIMULATION node (override present) defers to the
+            // version-floor check below, preserving the migration-mechanism
+            // tests. Early-return BEFORE any telemetry so a disabled production
+            // node reports no placement-migration activity at all, matching the
+            // pre-0.2.80 "migration off" baseline; the debug log still surfaces
+            // residual inbound hint volume for operators.
+            if !crate::node::network_bridge::p2p_protoc::placement_migration_enabled(
+                op_manager
+                    .ring
+                    .connection_manager
+                    .subscribe_hint_floor_override(),
+            ) {
+                tracing::debug!(
+                    key = %hint.key,
+                    ?source_addr,
+                    "Ignoring inbound SubscribeHint: placement migration disabled \
+                     (0.2.88 kill switch)"
+                );
+                return Ok(());
+            }
             // Placement-migration telemetry (#4404 follow-up): count every inbound
             // hint, before any admission gate, so `received` is the true arrival
             // rate and `received - acted` is the gated/dropped fraction.
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -800,6 +800,64 @@ pub(in crate::node) struct P2pConnManager {
 /// ramps with fleet upgrade rather than switching on everywhere at once.
 pub(crate) const SUBSCRIBE_HINT_MIN_VERSION: (u8, u8, u16) = (0, 2, 80);
 
+/// Production kill switch for SubscribeHint placement migration (the #4404
+/// directed-subscribe mechanism), gating BOTH the emit sides and the
+/// inbound-act site.
+///
+/// Disabled 0.2.88: over-aggressive migration storm (#4630-adjacent). do NOT
+/// re-enable aggressive placement migration; see
+/// `.claude/rules/hosting-invariants.md` (anti-pattern: holding-driven placement
+/// push). 0.2.87 fresh nodes pulled ~1000 contracts with no locality pattern,
+/// saturating bandwidth. Migration was re-enabled in v0.2.80 (#4404 / #4511) and
+/// then amplified by the v0.2.87 admission-gate loosening (#4625 / #4628).
+///
+/// When `false`, a PRODUCTION node neither EMITS SubscribeHint nudges
+/// (`consider_contract_migration` / `consider_migration_for_new_peer` in
+/// `p2p_protoc/migration.rs`) nor ACTS on inbound hints (the `SubscribeHint`
+/// receive handler in `crate::node` early-returns), so a fresh upgraded node is
+/// protected from the storm even when a not-yet-upgraded 0.2.87 neighbor still
+/// emits hints.
+///
+/// This const is consulted ONLY for production nodes; see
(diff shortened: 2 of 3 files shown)


sanity:
## Consolidated review — disable SubscribeHint placement migration (0.2.88 hotfix)

Full-tier review (placement-migration / hosting surface). One external non-Claude pass (Codex) plus a structured three-lens adversarial read. **No blocking findings; recommend merge.** CI is green on the reviewed HEAD `5bdc1a94`.

### External pass — Codex (`codex review --commit 5bdc1a94`)
Clean. Verdict: *"The change consistently gates the SubscribeHint emit and inbound-act paths while preserving the existing simulation override behavior and wire-version floor. I did not find a discrete regression introduced by this commit."* No findings raised.

### Lens (a) — skeptical bug-hunt (assume the kill switch has a hole)
The gate is `placement_migration_enabled(floor_override) = floor_override.is_some() || PLACEMENT_MIGRATION_ENABLED` with the const `= false` (`p2p_protoc.rs:825`/`863`). Production (`None`) → `false || false` → **off**; simulation (`Some`) → `true` → defers to the version-floor checks. The const-false value is pinned by `placement_migration_disabled_by_flag` (`p2p_protoc.rs:5067`), which asserts `!placement_migration_enabled(None)`. The `is_some() || const` form (rather than a bare `const` compare) is deliberate so `-D warnings` doesn't flag the guarded bodies as `unreachable_code` — documented at the fn. No hole.

### Lens (b) — this change's failure mode (can production re-enable migration?)
1. **Can a production path set `floor_override = Some`?** No. `NodeConfig::subscribe_hint_floor_override` is `#[serde(skip)]`, `pub(crate)`, defaults to `None` (`node.rs:330`/`467`); `ConnectionManager` defaults it to `None` (`connection_manager.rs:446`). Every `Some(..)` assignment is in `node/testing_impl.rs` (lines 1197, 1342–1374, 2007, 2099). Not wired to CLI/serde/config. Confirmed by reading the code, matching the main session's earlier finding.
2. **Can the emit guard be bypassed?** No. The only outbound `NetMessageV1::SubscribeHint` construction is `migration.rs:155`, inside `consider_contract_migration`, downstream of its top-of-fn guard (`migration.rs:113`). `consider_migration_for_new_peer` is guarded at its top (`migration.rs:197`) and only reaches an emit via `consider_contract_migration`. All callers (`p2p_protoc.rs:2695`, `connection_lifecycle.rs:387`/`1210`, `migration.rs:277`) route through the guarded fns.
3. **Can the inbound-act guard be bypassed?** No. The guard is the first statement of the `SubscribeHint` arm (`node.rs:1782`), before any telemetry/admission. The sole production `start_directed_subscribe` call is `node.rs:2012`, after the guard. (Other matches are the fn def / re-export / tests.)

### Lens (c) — tests (did the rename drop coverage?)
No. `placement_migration_active_at_reenable_floor` → `subscribe_hint_wire_floor_frozen_at_0_2_80` (`p2p_protoc.rs:3567`) keeps all five `version_supports_subscribe_hint` asserts unchanged — `(0,2,80)`, `(0,2,255)`, `!(0,2,79)`, `!(0,2,73)`, `!None`. The old test only ever exercised the wire/version gate; the rename + doc just stops it implying "migration active." Coverage **increases** with three new pins: `placement_migration_disabled_by_flag` (production-off / sim-defer), `placement_migration_emit_sites_gated_by_flag`, and `placement_migration_inbound_act_gated_by_flag` (source-scrape pins asserting each guard precedes its work anchor). Source-scrape pins match the established idiom in this module (no lightweight `P2pConnManager` harness exists).

### Non-blocking observation
The inbound guard early-returns before `record_received()` (`node.rs:1796`), so a disabled production node reports zero placement-migration telemetry rather than counting residual inbound hints from not-yet-upgraded 0.2.87 neighbors. This is intentional and documented (matches the pre-0.2.80 "migration off" baseline; residual inbound volume still surfaces in the debug log). Not a defect.

### Wire safety
`SUBSCRIBE_HINT_MIN_VERSION` is untouched, so the frozen wire floor is preserved; disabling emission only narrows what is sent. No wire-format change.

**Verdict: approve. No blocking findings.** Reverts production to the known-good pre-0.2.80 "migration off" baseline; simulation behavior is byte-for-byte unchanged via the override seam.

[AI-assisted - Claude]

PR #4643, opened 2026-07-01
feat: node resource-utilization telemetry for demand-driven hosting

sanity:
**First PR of the demand-driven hosting redesign (piece A1). PARKED for Ian's review, do NOT merge without his sign-off.**

## Problem
The demand-driven hosting redesign (#4642) sizes each peer's hosting budget by its own scarcest resource and evicts by demand. Validating that capability-relative budget and its eviction policy requires observing a node's real resource headroom in production, but the node surfaces zero resource-utilization metrics today: RSS is sampled only for an ad-hoc log line, the memory ceiling (min host RAM / cgroup) is computed only to size the module cache, and CPU/bandwidth are not aggregated anywhere the dashboard can see. The behavior is emergent and resource-dependent, so production telemetry is the validation instrument and must land first (instrumentation-is-horizontal principle, see the design doc).

## Solution
Foundational sub-PR A1: purely-additive resource telemetry, no behavior change.

- New `node::resource_metrics` samples the node's OWN utilization as **per-node aggregate scalars**: memory RSS + ceiling, cumulative CPU time, and cumulative bytes sent/received. RSS and the memory ceiling reuse existing sources (the RSS reader moved out of `p2p_protoc`, deduplicated; the ceiling reuses `wasm_runtime::read_total_ram_bytes`, now `pub(crate)` — the same value that will size the hosting budget in A2).
- Emitted as a periodic `resource_utilization` event on the **internal** telemetry stream (the batch cadence in `TelemetryReporter`'s worker), attributed to this node's peer id. Aggregate scalars only, never per-contract/per-request (nova ingests the whole fleet).
- Deliberately **NOT** on the client-facing wire protocol: nothing touches `SystemMetrics`, `HostResponse`, or any `freenet-stdlib` bincode type (that would be a wire-format break, out of scope for A1).
- Adds `docs/design/hosting-eviction.md`, the build guide for piece A.

## Testing
- `resource_metrics` unit tests: the sample populates every axis (RSS, ceiling, CPU non-absent on Linux; bandwidth always), the telemetry JSON carries every axis as a populated key, and the `/proc/self/stat` CPU parser handles parens/spaces in the comm field and rejects garbage.
- `cargo fmt`, `cargo clippy --locked -- -D warnings` and the `trace-ot` clippy gate both clean; `cargo check -p freenet` green.

## Fixes
Refs #4642

## Deferred
[...]

Issue #4642, opened 2026-07-01
Epic: demand-driven hosting / placement / eviction redesign


sanity:
## Goal

Replace demand-blind hosting with a demand-driven model: **a peer hosts the contracts it is most likely to be asked for, within a budget sized by its own scarcest resource.** This issue tracks the multi-release effort; individual pieces get their own issues/PRs.

Background: the 0.2.87 SubscribeHint placement-migration storm (fresh nodes pulling ~1000+ contracts with no locality, saturating residential uplinks) exposed that hosting was *holding*-driven, not *demand*-driven. #4640 (0.2.88) disabled the migration as a stopgap — confirmed effective in production (migration activity → 0 on updated nodes, cold-start acquisition collapsed from ~1476/hr to ~240/hr). This epic is the principled replacement.

## Status (updated 2026-07-10)

The **foundation and the eviction gauge are shipped and field-validated** (through 0.2.94: reconcile core driving, subscriber-primary eviction, reads/PUTs as a permanent demand signal, `min_ttl` dropped, evict-subscribed + memory teardown). The **atomic placement core** (serve-DURING + summary-first PUT + storm-safe re-root) is shipping in **0.2.95** (cutting now). What remains is small: delete the last durable-regardless-of-demand copy, finish the multi-axis budget, and the R4 cleanups (Source-2 removal, standalone-SUBSCRIBE retirement).

**Canonical design lives in `.claude/rules/hosting-invariants.md` (SETTLED — read it fresh) and `docs/design/demand-driven-hosting.md` (#4659, LOCKED 2026-07-01).** This issue is a task tracker; where it and the invariants file disagree, the invariants file wins. Settled decisions that supersede the earlier framing once captured here:

- **Eviction is ONE demand-ordered decision** (2026-07-06), ordering victims ascending `(local_subscription_count, downstream_subscriber_count, recency, key_bytes)`. There is **no separate admission decision, no OOM valve, no "I'm full" NACK, no chain cap** — a newcomer competes in the same eviction, so a ≤1-subscriber arrival can never displace a 2+-subscriber incumbent (multi-subscriber contracts protected for free) and refusal is emergent.
- **Reads and PUTs are a permanent demand signal** (2026-07-08), not just a recency tiebreak: a read-only / never-subscribed contract (River UI container, web UI container) is kept renewed and retained by its GET/PUT recency. Subscriptions outrank reads. `min_ttl` is dropped — recency reset on a real GET/PUT is the cold-start protection.
[...]

PR #4644, opened 2026-07-01
feat: size the hosting budget by available node RAM

sanity:
> **PARKED for Ian's review — do NOT merge without his sign-off.** (Was stacked on #4643/A1, now merged; this branch was rebased onto `main` and the base retargeted to `main`.)

Piece **A2** of the demand-driven hosting redesign (epic #4642; design spec `docs/design/hosting-eviction.md`, invariants `.claude/rules/hosting-invariants.md`).

## Problem
The hosting-storage budget was a flat 1 GiB default applied to every node regardless of the memory it actually has. On a small box (2 GB VM, laptop) that lets the node accumulate ~1 GiB of hosted contract state on top of the module caches, WASM arenas, and transport buffers, driving gateway RSS toward OOM (#4565). The budget was not sized by the node's own scarcest resource — the first thing the redesign needs.

## Approach
Make **only the DEFAULT** budget capability-relative; keep the existing byte-LRU eviction **policy** unchanged (the demand-driven fuel gauge is A3, a separate later PR).

- `cache::default_hosting_budget_bytes()` = `clamp(read_total_ram_bytes() / 8, 128 MiB, 1 GiB)`, reusing the `min(host RAM, cgroup limit)` source A1 exposed. `budget_for_ram()` splits the pure clamp math out for testing.
- **Heeds #4441:** the budget is RAM-scaled, **not** a fixed count/constant (the old fixed `>1024`-contract cap thrashed/OOM'd). The **divisor**, not the ceiling, is the small-box protection; the ceiling equals the historical flat 1 GiB, so a host with `>= 8 GiB` RAM sees **no change** and only memory-constrained hosts get less. Mirrors `module_cache.rs::budget_for_ram`.
- **Operator override preserved:** an explicit `max_hosting_storage` (flag / config / env) still wins; only the DEFAULT changed.

### Re-derive on each boot (first-boot + upgrade persistence, #4565)
`build()` serializes the resolved `Config` into `config.toml`, so the auto-derived budget would be **pinned** — a node that first-booted (or was on a pre-A2 release) on a large box would keep the large budget after moving to a smaller box / tighter cgroup, defeating the whole point. Fixed in two parts:
- `skip_serializing_if = "is_default_hosting_budget"` omits the auto-derived value from `config.toml`, so each boot re-derives from live RAM (fresh configs).
- A one-time upgrade migration in `ConfigArgs::build` re-derives a **file** value equal to the pre-A2 flat 1 GiB sentinel (`LEGACY_FLAT_HOSTING_BUDGET_BYTES`), so existing gateways stop carrying the pinned 1 GiB. CLI/env explicit values are parsed before the file merge, so they still win.

## Instrumentation (mandatory per the spec's horizontal-instrumentation rule)
Per-node aggregate scalars added to the existing `RouterSnapshot` telemetry (same home as the module-cache budget/eviction gauges): `hosting_budget_bytes`, `hosting_current_bytes` (occupancy ratio = `current / budget`, derived by the collector), `hosting_contract_count`, and a monotonic `hosting_budget_evictions_total`. Per-node scalars only, no per-contract firehose. `HostingCache` gained a `budget_evictions_total` counter (incremented only in `evict_over_budget`) and a `stats()` snapshot.

## Testing
- `cache.rs`: `budget_for_ram` scales-and-clamps; **#4441 shape guard** (RAM-scaled not fixed; small box < large box, and < old flat 1 GiB); default within clamp; budget-eviction counter + `stats()`.
- `config.rs`: RAM-scaled default within `[MIN, MAX]`; auto value not persisted + re-derives; **upgrade migration** re-derives the legacy 1 GiB sentinel while an explicit non-legacy value survives; explicit override round-trips; persisted-fields round-trip guard.
- `telemetry.rs`: pin test that all four hosting gauges reach the OTLP body.
- `cargo fmt`, both CI clippy gates, `cargo check -p freenet` green.

## Fixes
Refs #4642. Addresses #4565.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -481,8 +481,22 @@ impl ConfigArgs {
                 self.network_api.location.get_or_insert(loc);
             }
             self.log_level.get_or_insert(cfg.log_level);
-            self.max_hosting_storage
-                .get_or_insert(cfg.max_hosting_storage);
+            // #4565 upgrade migration: a pre-A2 release auto-persisted the OLD
+            // flat 1 GiB default as `max-hosting-storage = 1073741824`. Treat
+            // that exact historical sentinel as auto-derived (NOT an explicit
+            // operator choice) so it RE-DERIVES from live RAM on upgrade —
+            // otherwise a small box that upgraded would keep the 1 GiB budget and
+            // stay on the #4565 OOM path. `skip_serializing_if` alone only stops
+            // NEW configs from pinning; it can't unpin the historical value.
+            // CLI/env explicit values are parsed into `self` BEFORE this file
+            // merge, so `--max-hosting-storage 1073741824` / the env var still
+            // wins; only a FILE value equal to the sentinel is re-derived. On a
+            // >=8 GiB box the re-derivation yields 1 GiB anyway; on a smaller box
+            // reducing it is the whole point.
+            if cfg.max_hosting_storage != crate::ring::LEGACY_FLAT_HOSTING_BUDGET_BYTES {
+                self.max_hosting_storage
+                    .get_or_insert(cfg.max_hosting_storage);
+            }
             self.per_user_secret_quota_bytes
                 .get_or_insert(cfg.per_user_secret_quota_bytes);
             self.per_user_inactive_ttl_secs
@@ -1090,9 +1104,19 @@ pub struct Config {
     /// 128 MiB, 1 GiB)`, so a memory-constrained host gets a proportionally
     /// smaller budget instead of the old flat 1 GiB (#4642 A2 / #4565). Set an
     /// explicit value to override the RAM-scaled default.
+    ///
+    /// `skip_serializing_if` drops this field from `config.toml` when it holds
+    /// the auto-derived default, so the budget RE-DERIVES from live RAM on every
+    /// boot instead of being pinned at first boot. Without this, a node that
+    /// first-booted on a large box would bake the large budget into `config.toml`
+    /// and keep it after moving to a smaller box / tighter cgroup — defeating the
+    /// #4565 OOM protection. An explicit operator value differs from the derived
+    /// default, so it is persisted and survives restarts. See
+    /// [`is_default_hosting_budget`].
     #[serde(
         default = "default_max_hosting_storage",
-        rename = "max-hosting-storage"
+        rename = "max-hosting-storage",
+        skip_serializing_if = "is_default_hosting_budget"
     )]
     pub max_hosting_storage: u64,
     /// Per-user secret-storage quota in bytes for hosted mode (#4561, P5 of
@@ -1187,6 +1211,20 @@ fn default_max_blocking_threads() -> usize {
 /// budget (addresses #4565); an explicit value always overrides it.
 fn default_max_hosting_storage() -> u64 {
     crate::ring::default_hosting_budget_bytes()
+}
+
+/// `skip_serializing_if` predicate for [`Config::max_hosting_storage`]: true
+/// when the resolved value equals the auto-derived RAM-scaled default, so it is
+/// omitted from `config.toml` and re-derived from live RAM on the next boot
+/// (#4565 first-boot-pinning fix — see the field docs).
+///
+/// Because the derived default is always clamped to `[128 MiB, 1 GiB]`, an
+/// explicit operator value OUTSIDE that range can never match and is always
+/// persisted. The one ambiguous case — an explicit value that happens to equal
+/// the current derived default — re-derives on a RAM change, which is the safe
+/// direction (toward the smaller, capability-relative budget) anyway.
+fn is_default_hosting_budget(v: &u64) -> bool {
+    *v == default_max_hosting_storage()
 }
 
 /// Default per-user secret-storage quota (4 MiB). Resolves to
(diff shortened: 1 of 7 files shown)


sanity:
## Consolidated review (external Codex pass + adversarial self-read)

**Scope reviewed:** A2's diff against the A1 base branch (`feat/hosting-resource-telemetry`).

### External model — Codex (`codex review --base feat/hosting-resource-telemetry`)
- **[P2] Environment-dependent test — FIXED (26fecb9c).** `test_new_uses_configured_budget` asserted `HostingManager::default().hosting_budget_bytes()` equalled the fixed 1 GiB test constant, but `Default` now resolves the RAM-scaled default. On a host with `< 8 GiB` usable RAM / a cgroup limit the default is below 1 GiB, so the assertion would flake on small CI runners (it passed locally only because this box has `>= 8 GiB`, where the default hits the 1 GiB ceiling). Now asserts against `default_hosting_budget_bytes()` (the production default fn), which is environment-independent.

No other findings from Codex.

### Adversarial self-read
- **Failure-class sweep:** searched for any other test/production site asserting a fixed 1 GiB hosting default (`hosting_budget_bytes()`, `max_hosting_storage`, `1073741824`). The Codex-flagged test was the only instance. All other `HostingManager::new(DEFAULT_HOSTING_BUDGET_BYTES)` sites use a test-local 1 GiB constant with an explicit budget, which is deterministic and intended.
- **Operator override preserved** at both resolution points (`ConfigArgs::build` fallback and the serde `default_max_hosting_storage`); explicit value still round-trips (test kept).
- **#4441 shape:** budget is RAM-scaled, not a fixed count/constant; the divisor (not the ceiling) is the small-box protection; ceiling equals the historical flat 1 GiB so `>= 8 GiB` hosts see no change. Pinned by `budget_for_ram_is_ram_scaled_not_fixed` + `budget_for_ram_scales_and_clamps`.
- **Telemetry:** the four hosting gauges follow the exact `RouterSnapshot` None-then-Ring-overwrites pattern as the module-cache gauges; hand-mirrored OTLP body pinned by `router_snapshot_json_includes_hosting_budget_gauges`. `budget_evictions_total` increments only in `evict_over_budget` (genuinely budget-triggered). `hosting_cache_stats()` takes a single brief read lock, no nesting.
- **No eviction-policy change** (that's A3); only the budget SIZE changed. No hosting-invariant regressions — this moves toward capability-relative budgeting (invariant 3/4).

### CI / gates
`cargo fmt`, `cargo clippy --locked -- -D warnings` + `trace-ot` gate clean, `cargo check -p freenet` green, and the hosting/cache/config/telemetry unit tests pass locally. Awaiting CI on the branch.

**PARKED for Ian's review — not merging.**

[AI-assisted - Claude]

sanity:
## Updated consolidated review (post-rebase onto main + first-boot/upgrade fix)

**Since the last review:** A1 (#4643) merged to main; this branch was **rebased onto main** (clean, no conflicts; PR base retargeted to `main`) and gained the persistence fix below. Re-running the review because the diff changed.

### External model — Codex (`codex review --base main`), round 2
- **[P2] First-boot / upgrade persistence — FIXED.** `build()` serializes the resolved `Config` into `config.toml`, so the RAM-scaled default was baked at first boot and (worse) existing pre-A2 configs already pin `max-hosting-storage = 1073741824`. On a small box that meant the budget never re-derived — defeating #4565. Fix has two parts:
  1. `skip_serializing_if = "is_default_hosting_budget"` on `Config.max_hosting_storage` — the auto-derived value is omitted from `config.toml`, so each boot re-derives it from live RAM (fresh configs).
  2. One-time upgrade migration in `ConfigArgs::build`: a **file** value equal to the pre-A2 flat sentinel (`LEGACY_FLAT_HOSTING_BUDGET_BYTES` = 1 GiB) is treated as auto-derived and re-derived (existing configs). CLI/env explicit values are parsed before the file merge, so `--max-hosting-storage` / the env var still win.

### External model — Codex, round 3 (final diff)
- **"I did not identify any discrete, actionable bugs in the diff."** P2 resolved, no new findings.

### Adversarial self-read
- **CLI/env precedence:** `--max-hosting-storage` / `MAX_HOSTING_STORAGE` are parsed into `self` before the file merge; the migration only gates the file `get_or_insert`, so an explicit flag/env value of any magnitude (including exactly 1 GiB) still wins.
- **Value-specificity:** the migration targets only the exact 1 GiB sentinel; a non-legacy persisted value (e.g. 2 GiB) survives — pinned by `legacy_pinned_hosting_budget_re_derives_but_explicit_survives` case (b).
- **Documented tradeoff:** an operator who pinned exactly 1 GiB via `config.toml` (not CLI/env) is also re-derived. On a >=8 GiB box that yields 1 GiB anyway; on a smaller box reducing it is the #4565 intent, and the CLI/env flag is the escape hatch. `LEGACY_FLAT_HOSTING_BUDGET_BYTES` is a deliberately separate constant from the clamp ceiling so it stays 1 GiB if the ceiling ever changes.
- **Self-healing:** after one upgrade boot the small-box config drops the pinned line (re-derived == derived → skip-serialized) and stays on the clean re-deriving path.
- **Comment nit fixed:** `current / budget` is occupancy/utilization, not headroom (headroom = 1 - that) — corrected in cache.rs / ring.rs / router.rs.
- Round-trip guard `all_persisted_config_fields_round_trip_through_build` still passes (its seed `123_456_789` is below the 128 MiB floor, so never matches the derived default or the legacy sentinel).

### Gates
`cargo fmt`, both CI clippy gates (`--locked -- -D warnings` and `--features trace-ot`), `cargo check -p freenet`, and the hosting/cache/config/telemetry unit tests all pass locally. Awaiting CI on the pushed HEAD.

**PARKED for Ian's review — not merging.** Note: the upgrade migration is a finish-the-fix extension beyond the original first-boot scope; flag if a narrower scope is preferred.

[AI-assisted - Claude]

PR #4646, opened 2026-07-01
feat: consult host advertisements at routing terminus

sanity:
> **PARKED for Ian's review — do NOT merge without his sign-off.**
> This is piece **C** of the demand-driven hosting redesign (epic #4642), implemented in parallel with the other pieces. It touches the GET/SUBSCRIBE routing path (a high-risk surface), so it is intentionally held for review rather than auto-merged.

## Problem

A GET/SUBSCRIBE routes toward a contract's key and reaches a **terminus** (the closest peer it can route to). Routing is purely location-based (`k_closest_potentially_hosting`), and each relay forwards to a *single* peer per hop (`MAX_RELAY_RETRIES = 1`). So if the terminus's closest neighbor toward the key does not host the contract, the relay bubbles `NotFound` **without trying its other neighbors** — even when one of them already advertised (via the neighbor-hosting mesh) that it hosts the contract. That advertised host is one hop **off the direct routing path**, and the request dead-ends with a false `NotFound`.

This implements **invariant 5** from `.claude/rules/hosting-invariants.md`: *"findability is routing + on-demand advertisement — a request routed toward its key, at its terminus, consults host advertisements."*

## Approach

At the GET/SUBSCRIBE terminus, before returning `NotFound`, consult the host advertisements the peer already received from neighbors (`NeighborHostingManager`) and forward the existing GET/SUBSCRIBE `Request` to an advertised host if one exists.

**Additive — no wire-format or `freenet-stdlib` change.** Reuses the existing message types and the existing visited/dedup machinery.

**Not a listed anti-pattern.** It forwards ONLY to a peer that *already advertised* hosting — it pushes and caches nothing and manufactures no demand, so it is not speculative pre-replication (which invariant 5 forbids). Bounded: terminus-only (not every hop); GET tries the closest ≤2 advertised hosts, SUBSCRIBE (single-shot) the closest 1; each candidate is marked tried + visited and forwarded with HTL-1, so loops cannot form. A terminus has zero closer routing peers, so these few extra forwards do not compound across hops (unlike the removed per-hop retry fan-out that caused the 3^HTL amplification).

### Key changes (`file:line`)
- `operations.rs` — `consult_advertised_hosts` + pure, unit-tested `rank_advertised_hosts`; `record_terminal_consult_outcome`.
- `operations/get/op_ctx_task.rs::drive_relay_get_inner` — consult in the exhaustion (None) arm before `relay_send_not_found`; resolved/still-NotFound telemetry after `classify`.
- `operations/subscribe/op_ctx_task.rs::drive_relay_subscribe` — consult in the no-candidates terminus branch before the NotFound response.
- `node/neighbor_hosting.rs` — `neighbors_with_contract_id` (relay drivers hold an instance id, not a full key).
- `node/network_status.rs` + `config.rs` (`GlobalTestMetrics`) — per-node aggregate scalars: attempts / hits / resolved-found / still-NotFound.
- `node/testing_impl/in_memory.rs::append_contracts` — harness fix: register startup-hosted contracts in `neighbor_hosting.my_contracts` so they are actually advertised (previously `SeedHostedContract` left them un-advertised, defeating both this consult and UPDATE proximity forwarding).

## Testing

- **Unit:** `rank_advertised_hosts` (closest-first ordering, max-hosts truncation, visited exclusion), `neighbors_with_contract_id`, consult telemetry counters.
- **Source-scrape pins:** GET/SUBSCRIBE terminus consults before NotFound and records the outcome.
- **Simulation** (`test_terminal_advertisement_consult_closes_get_dead_end`): a host one hop off the routing path, **migration disabled**, so `terminal_consult_resolved_found() > 0` is the unambiguous proof the consult (not routing or migration) closed the dead-end. Deterministic across repeated runs.
- Verified **no regression** in the existing GET dead-end tests and the subscription-root renewal-storm tests.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3246,6 +3246,11 @@ std::thread_local! {
     static GLOBAL_PENDING_OP_HWM: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_NEIGHBOR_HOSTING_UPDATES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_ANTI_STARVATION_TRIGGERS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // Terminal advertisement consult (hosting redesign piece C, invariant 5).
+    static GLOBAL_TERMINAL_CONSULT_ATTEMPTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_TERMINAL_CONSULT_HITS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_TERMINAL_CONSULT_RESOLVED_FOUND: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_TERMINAL_CONSULT_STILL_NOT_FOUND: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
 }
 
 /// Global test metrics for tracking events across the simulation network.
@@ -3280,6 +3285,10 @@ impl GlobalTestMetrics {
         GLOBAL_PENDING_OP_HWM.with(|c| c.set(0));
         GLOBAL_NEIGHBOR_HOSTING_UPDATES.with(|c| c.set(0));
         GLOBAL_ANTI_STARVATION_TRIGGERS.with(|c| c.set(0));
+        GLOBAL_TERMINAL_CONSULT_ATTEMPTS.with(|c| c.set(0));
+        GLOBAL_TERMINAL_CONSULT_HITS.with(|c| c.set(0));
+        GLOBAL_TERMINAL_CONSULT_RESOLVED_FOUND.with(|c| c.set(0));
+        GLOBAL_TERMINAL_CONSULT_STILL_NOT_FOUND.with(|c| c.set(0));
     }
 
     /// Records that a ResyncRequest was received.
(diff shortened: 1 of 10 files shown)


sanity:
## Consolidated review (Full tier — GET/SUBSCRIBE routing is a high-risk surface)

**External model (Codex, `codex review --base main`)** ran three times as the code evolved, plus my own adversarial read.

### Codex findings — all addressed

- **[P2] GET consult fired at retry-cap exhaustion, not only a true terminus.** Investigated: this is actually the *only* useful placement — at a true no-candidate terminus every connection is already visited, so the consult (which excludes visited peers) would find nothing. The value comes from reaching an unvisited advertised neighbor that the single greedy forward (`MAX_RELAY_RETRIES=1`) skipped. Fixed the misleading "leaf/terminus" comments and documented why the bounded fan-out is safe (local-lookup-gated → a truly-absent contract triggers zero extra forwards; advertised-only; bubble-up halts at the first Found). **This same analysis showed my SUBSCRIBE consult at `candidates.is_empty()` was a no-op** — reworked `drive_relay_subscribe` to consult after its single forward returns NotFound (commit `900ebcd1`).
- **[P2] Consulting after a timed-out/send-failed forward reused `incoming_tx`**, so a late reply from the timed-out peer could satisfy the consult waiter and bubble a false NotFound. Fixed: the consult now runs ONLY after a *clean* downstream reply — GET gates on `last_forward_failed`; SUBSCRIBE splits the forward outcome into `NotFound` (clean → consult) vs `Failed` (timeout/send-failure/unexpected → bubble NotFound, no consult). Skipping restores the exact pre-consult behavior for that corner case (no regression). Source-scrape pins added (commit `686d7a70`).
- **Third pass: clean** — "I did not identify any discrete, actionable correctness issues in the diff. The new terminal consult paths are bounded and preserve the existing visited/dedup handling."

### My adversarial read — no additional issues

- **Telemetry accuracy:** exactly one outcome (resolved-found XOR still-not-found) per consult phase; attempts/hits only recorded when a consult actually runs (gated by `consult_targets.is_some()` on GET, and by the `NotFound` outcome on SUBSCRIBE); `consult_active` correctly attributes a Found reply to the consult vs routing.
- **Loop/dedup safety:** consult candidates are marked tried + visited before forwarding and carry HTL-1, so the existing bloom/dedup prevents loops; the consult never targets self/upstream/visited peers.
- **SUBSCRIBE registration:** `register_downstream_subscriber` fires at most once on a successful path (routing OR consult, never both).
- **Amplification:** with the P2 gate the consult fires only after a clean NotFound, to already-advertised hosts, ≤2 per relay — no 3^HTL revival.

### Test isolation note

The unconditional harness advertise fix interacted with the placement-migration dead-end controls (the consult legitimately closed their dead-end). Made seeded-host advertising **opt-in per network** (`SimNetwork::enable_seeded_host_advertisements`, default off); only the consult test opts in. All three tests (consult + both migration controls) and the three subscription-root renewal-storm tests pass under **nextest** (the CI runner); a `cargo test` parallel run needs `--test-threads=1` for sim tests, as the project's testing rules already note.

**PARKED for @sanity's review — do not merge without sign-off.**

[AI-assisted - Claude]

sanity:
## Review update — rule-review warnings + telemetry accuracy fixed

### Rule-review (2 warnings) — both fixed (not overridden)

1. **Streaming `resolved_found` false-positive** — the consult telemetry was recorded as soon as a `Terminal` reply classified, before the fallible terminal side-effect. Moved `resolved_found` recording to the actual delivery points: InlineFound after `send_result?`, Streaming after the loopback state-cache / remote `pipe_stream` success, SUBSCRIBE after `relay_subscribe_send_response` succeeds. A send/claim/pipe failure now correctly counts as still-not-found. This closes the #4345 "classify terminal before a fallible side effect" anti-pattern for these counters.
2. **SUBSCRIBE lacked a behavioral e2e test** — added `test_terminal_advertisement_consult_closes_subscribe_dead_end`. A bare subscribe for an absent contract fails the local `RegisterSubscriberListener` gate and never routes, so the subscriber is seeded with the contract; it then routes its upstream subscribe toward the key, dead-ends at the non-hosting cluster, and the consult forwards to the off-path advertised host which answers `Subscribed` (`resolved_found > 0`). Migration off, deterministic under nextest.

### Independent skeptical review (3 items) — addressed

- **[fix]** telemetry false-positive (same as rule-review #1, extended to the pre-`send_result?` InlineFound and pre-send SUBSCRIBE cases).
- **[fix]** GET now skips the consult at a no-candidate terminus (`did_forward` gate), matching SUBSCRIBE's `candidates.is_empty()` branch — no spurious attempt/still_not_found counters.
- **[comment]** documented at the consult site that "an advertised host almost always HAS the contract" is load-bearing, with the bounds (HTL-1, visited bloom, connected-only, eviction-retracts-advertisement) that keep the rare stale-host re-routing case from reviving 3^HTL.

### External model (Codex) — now clean

Across the review Codex ran several passes and raised P2s in two classes: (a) consult-after-a-timed-out-forward reusing the tx (fixed earlier by gating on a clean reply), and (b) recording `resolved_found` before delivery (fixed as above). Latest pass: *"I did not identify any discrete correctness issues introduced by the diff."*

CI: all deterministic gates green including `rule-review/warnings` (No warnings found); heavy jobs (Simulation, Unit & Integration, NAT, Windows, macOS) running — monitoring to green.

**Still PARKED for @sanity — do not merge without sign-off.**

[AI-assisted - Claude]

PR #4649, opened 2026-07-01
feat: emit resource_utilization telemetry every 60s not 10s

sanity:
> **PARKED for Ian's review — do NOT merge without his sign-off.**
> **Targets the release AFTER 0.2.89.** Do not enable auto-merge.

## Problem

The A1 `resource_utilization` telemetry (#4643, part of the demand-driven hosting redesign #4642) is emitted on the existing ~10s telemetry batch tick. Resource utilization (memory RSS/ceiling, CPU time, cumulative bandwidth) is a slowly-changing, node-wide signal, so a 10s cadence is wasteful. Every production peer reports telemetry to the central collector (nova) by default, so the 10s cadence multiplies collector load across the whole fleet for a signal that barely moves in 10s.

## Approach

Throttle only `resource_utilization` to a ~60s cadence by emitting on every 6th batch tick (`RESOURCE_TELEMETRY_BATCH_INTERVAL = 6`, times the 10s `BATCH_INTERVAL_SECS` = 60s), tracked by a per-worker tick counter. The decision is factored into `should_emit_resource_utilization()` so it is unit-testable without driving the async interval loop (same pattern as `admit_event`). The first tick still emits, for prompt startup visibility. The batch/flush cadence and every other telemetry event are unchanged; only `resource_utilization`'s emit is throttled. About a 6x reduction in that signal's volume with no real information loss.

## Testing

- Added `test_resource_utilization_emits_once_per_six_batch_ticks`: asserts the sample fires on ticks 0 and 6 (once per 6 batches) and not on ticks 1-5, over two full cycles, and that the constant yields ~60s.
- Existing telemetry unit tests still pass (35/35).
- `cargo fmt` / `cargo clippy -p freenet` / `cargo check -p freenet` clean.

## Scope

Single logical change: only `resource_utilization`'s cadence. No wire-format change, no change to the batch tick, flush cadence, rate limiter, or any other telemetry's cadence.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/tracing/telemetry.rs
+++ b/crates/core/src/tracing/telemetry.rs
@@ -155,6 +155,21 @@ const MAX_BUFFER_SIZE: usize = 100;
 /// How often to send batched events (in seconds)
 const BATCH_INTERVAL_SECS: u64 = 10;
 
+/// Emit `resource_utilization` telemetry once every this many batch ticks.
+///
+/// The batch tick fires every [`BATCH_INTERVAL_SECS`] (10s), but resource
+/// utilization (memory RSS/ceiling, CPU time, cumulative bandwidth) is a
+/// slowly-changing, node-wide signal, so sampling it on every 10s tick is
+/// wasteful. Crucially, EVERY production peer reports telemetry to the central
+/// collector (nova) by default, so a 10s cadence multiplies collector load
+/// across the whole fleet for a signal that barely moves in 10s. Emitting once
+/// per 6 batch ticks (~60s) cuts that ~6x with no real signal loss.
+///
+/// Only `resource_utilization`'s cadence is throttled this way; every other
+/// telemetry event still batches and flushes on the normal [`BATCH_INTERVAL_SECS`]
+/// tick.
+const RESOURCE_TELEMETRY_BATCH_INTERVAL: u64 = 6;
+
 /// Maximum events per second (aggregate rate limiting across all priorities)
 const MAX_EVENTS_PER_SECOND: usize = 10;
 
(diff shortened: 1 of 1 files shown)


sanity:
## Consolidated review (Codex + adversarial self-read)

Risk tier: **Light**. Small, self-contained, no high-risk surface (no wire format, concurrency semantics, routing, or auth change). This only throttles the emit cadence of one existing Operational telemetry event; the batch tick, flush cadence, rate limiter, and every other event are untouched. No behavioral change to the protocol.

### External model pass (Codex, `codex review --base main`)
Clean. Verdict: "The change only throttles resource utilization telemetry while preserving the existing flush cadence, and the added counter logic and regression test align with the intended 60s cadence. I did not find any discrete correctness issues introduced by this patch."

### Adversarial self-read
Verified:
- **Single emit site.** `resource_utilization_telemetry()` has exactly one production caller (the batch-tick arm), now gated by `should_emit_resource_utilization()`. No other emit path was missed.
- **Startup visibility preserved.** Counter starts at 0, so the first batch tick (t=0, which `tokio::time::interval` fires immediately) still emits; then ticks 6, 12, ... = 60s, 120s. Correct ~60s cadence.
- **Scope is minimal.** `flush()` still runs every 10s batch tick; only the resource sample is throttled, not the batch/flush cadence or any sibling telemetry.
- **No new time/RNG dependency.** Uses a pure per-worker counter, so no DST/`TimeSource` concern is introduced (avoids relying on wall-clock at all).
- **Overflow-safe.** `wrapping_add` cannot panic; the modulo phase shift at wrap is irrelevant on a ~5.8-trillion-year horizon.
- **Rate limiter.** Emitting less often only relieves the `admit_event` per-second budget; no new pressure.

### Findings
None from either pass. No changes required.

Note for the reviewer: the only user-visible effect is that the dashboard/collector receives the `resource_utilization` panel ~6x less often. That is the intended tradeoff this PR parks for your call.

[AI-assisted - Claude]

PR #4650, opened 2026-07-01
feat: demand-ordered hosting eviction (A3)

sanity:
> **PARKED for Ian's review — do NOT merge.** This is the core of the
> resource-usage rearchitecture and the highest-risk piece of the demand-driven
> hosting redesign. No auto-merge.
>
> **Now based on `main`.** A2 (#4644, capability-relative budget) and #4649
> (resource-telemetry cadence) have merged, so this was **rebased from A2's
> branch onto `main`** and the PR base retargeted to `main`. The rebase was
> **clean, no conflicts**: A2's squash-merge (`398adaff`) is byte-identical to
> the A2 tip this branched from for every hosting file, so the A3 eviction
> changes replayed onto the merged budget code untouched; the only diverging file
> on main was `telemetry.rs` (#4649's cadence throttle), whose changes don't
> overlap the A3 gauges — both coexist (verified: hosting + telemetry lib tests
> green, 158 passed).

## Problem

Hosting-cache eviction was byte-budget **LRU**: over budget, it dropped the
least-recently-*accessed* contract. Recency is demand-blind — a repeatedly-
requested contract (a live River room) could be evicted ahead of never-requested
junk just because the junk was touched more recently. This is the #4338
miscalibration class and the "byte-only eviction" anti-pattern called out in
`.claude/rules/hosting-invariants.md`.

## Approach

Piece **A3** of the demand-driven hosting redesign (#4642). Eviction becomes a
Greedy-Dual, **demand-ordered**:

- Each hosted contract has `keep_score = eviction_floor + predicted_demand`, set
  on insert and **refreshed to the current frontier on every read**
  (GET/SUBSCRIBE).
- `eviction_floor` is a per-peer running value that ratchets up to the
  `keep_score` of each evicted contract (Greedy-Dual aging, in cache-contention
  units, **not** wall-clock). Over budget, the **lowest** `keep_score` is
  evicted; ties break by least-recently-accessed, so **uniform demand degrades
  to LRU** (every prior LRU test still passes unchanged).
- `predicted_demand` is the **proximity prior only** (the observed-rate blend is
  A4): `g(distance-to-key)`, a monotone non-increasing isotonic regression fit
  from this peer's own `(distance, read-rate)` samples. **Reuses the existing
  `pav_regression`/PAVA machinery** the router uses — new estimator in
  `ring/hosting/demand.rs`, no re-implementation.
- Interest signals: GET/SUBSCRIBE = read-demand (refresh `keep_score`, train the
  prior); **PUT = SEED only** (sets the estimate, earns no frontier credit,
  doesn't train); an active **SUBSCRIBE pins** the contract via the existing
  `contract_in_use` retain predicate (exempt from eviction). The pin is
  deliberately NOT widened to `is_subscribed` — that would re-introduce an
  unbounded GC exemption (ring.md / AGENTS.md).
- A2's **capability-relative byte budget stays the pressure threshold**; bytes
  are now only a tiebreak/ceiling.

The estimator + own-location live in `HostingManager` (Ring pushes
`own_location` on the snapshot cadence). `HostingCache` stays location-agnostic —
it orders only by the demand scalar it is handed — so it remains unit-testable
without a ring.

**Guard comment** added at the eviction site: *do NOT revert to byte-only LRU —
see hosting-invariants*.

### Instrumentation (horizontal — a PR acceptance criterion)

Per-node aggregate scalars on the existing snapshot / OTLP body:
- **local hit-rate** (`hosting_local_hits_total` / `hosting_local_misses_total`)
  — served-locally vs had-to-fetch (cache-membership proxy);
- **eviction rate** (`hosting_budget_evictions_total`, from A2);
- **evictions-of-recently-read** (`hosting_evictions_of_recently_read_total`) —
  the #4338 miscalibration alarm (evicting a contract with genuine repeat demand).

## Testing

- **demand.rs**: isotonic prior fit is monotone non-increasing, neutral below
  min samples, bounded rolling window, non-negative out-of-range, invalid inputs
  ignored.
- **cache.rs**: `keep_score` set-on-insert / refresh-on-read, eviction-floor
  ratchet, **demand beats recency** (the LRU-distinguishing regression — fails
  under byte-LRU), the **#4338 River-room scenario** (repeatedly-read survives a
  junk stream, zero recently-read evictions), PUT-seeds-not-demand, subscribe-pin
  exemption, recently-read + hit/miss counters. All prior LRU tests retained.
- **hosting.rs**: manager demand path trains the prior from repeated reads; PUT
  does not train.
- **telemetry.rs**: pin test that the A3 gauges reach the OTLP body.

`cargo fmt` clean; `cargo clippy --locked -p freenet` clean (the 3 remaining
warnings are pre-existing on the A2 base, in test-only targets CI's clippy does
not lint); hosting + telemetry lib tests green (122 passed).

### On the simulation test

The task called for a full-network `run_simulation_direct` #4338 test. This is
**impractical within A3's scope** and I judged forcing it would produce a flaky
test (against `flaky-tests` policy): the hosting cache's TTL gate is wall-clock
with `min_ttl = 8 min`, the direct sim runner drives *random* events (can't
target one contract for repeated GETs vs junk), and there is no per-node hosting-
budget knob in the harness. A faithful sim eviction test would need a sim
`TimeSource` plumbed into `HostingManager` (today it hardcodes `InstantTimeSrc`)
— a cross-cutting change that would *also* give A2's budget eviction its
currently-absent sim coverage. I therefore validated the demand-ordering
guarantee **deterministically at the cache + manager level** and flag the
sim-time plumbing as a sensible separable follow-up for Ian to weigh.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -948,6 +948,10 @@ async fn process_open_request(
                                 "Returning locally cached contract state"
                             );
 
+                            // #4642 A3 hit-rate instrumentation: this client GET
+                            // was answered from local hosted state (a hit).
+                            op_manager.ring.record_get_served_locally();
+
                             // Handle subscription for locally found contracts
                             if subscribe {
                                 if let Some(subscription_listener) = subscription_listener {
@@ -977,6 +981,11 @@ async fn process_open_request(
                                 contract,
                             })));
                         }
+
+                        // #4642 A3 hit-rate instrumentation: this client GET could
+                        // not be answered locally and is routed to the network
+                        // (a forward/miss).
+                        op_manager.ring.record_get_forwarded();
 
                         // Driver owns its routing state in task
                         // locals and calls notify_contract_handler
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1474,6 +1474,25 @@ impl Ring {
             snapshot.hosting_contract_count = Some(hosting.contract_count);
             snapshot.hosting_budget_evictions_total = Some(hosting.budget_evictions_total);
 
+            // Demand-ordered eviction gauge (#4642 A3): the
+            // #4338 miscalibration signal (evictions of repeatedly-read
+            // contracts). Per-node aggregate scalar on the same cadence.
+            snapshot.hosting_evictions_of_recently_read_total =
+                Some(hosting.evictions_of_recently_read_total);
+            // Local-client GET hit-rate (#4642 A3): served-locally vs routed to
+            // the network. Driven by the real serve/forward decision in the
+            // client GET handler, so it is read from the manager counters (not
+            // cache membership).
+            snapshot.hosting_local_hits_total = Some(ring.hosting_manager.local_get_serves());
+            snapshot.hosting_local_misses_total = Some(ring.hosting_manager.local_get_forwards());
+
+            // Keep the hosting manager's copy of our own ring location current so
+            // the proximity-prior demand estimate (#4642 A3) can turn a contract
+            // key into a distance. Best-effort: only push a known location.
+            if let Some(own_loc) = ring.connection_manager.own_location().location() {
+                ring.hosting_manager.set_own_location(own_loc);
+            }
+
             // Interest-weighted (two-tier) module-cache SHADOW gauges
             // (#4441/#4534): always-on, independent of the
             // FREENET_MODULE_CACHE_INTEREST_TIERED feature flag. They quantify
@@ -2319,6 +2338,18 @@ impl Ring {
     /// and which contracts were evicted (if any).
     pub fn record_get_access(&self, key: ContractKey, size_bytes: u64) -> RecordAccessResult {
         self.host_contract(key, size_bytes, AccessType::Get)
+    }
+
+    /// Record that a local-client GET was answered from local hosted state
+    /// (a hit) — see [`HostingManager::record_local_get_serve`]. (#4642 A3)
+    pub fn record_get_served_locally(&self) {
+        self.hosting_manager.record_local_get_serve();
+    }
+
+    /// Record that a local-client GET was routed to the network (a forward/miss)
+    /// — see [`HostingManager::record_local_get_forward`]. (#4642 A3)
+    pub fn record_get_forwarded(&self) {
+        self.hosting_manager.record_local_get_forward();
     }
 
     /// Whether this node is hosting this contract (has it in cache).
--- a/crates/core/src/ring/hosting.rs
+++ b/crates/core/src/ring/hosting.rs
@@ -30,6 +30,7 @@
 //! - TTL protects recently accessed contracts from premature eviction
 
 mod cache;
+mod demand;
 
 use crate::util::backoff::{ExponentialBackoff, TrackedBackoff};
 use crate::util::time_source::{InstantTimeSrc, TimeSource};
@@ -50,13 +51,16 @@ use cache::{DEFAULT_MIN_TTL, HostingCache, HostingCacheStats};
 #[cfg(test)]
 pub(crate) use cache::{MAX_DEFAULT_HOSTING_BUDGET_BYTES, MIN_DEFAULT_HOSTING_BUDGET_BYTES};
 use dashmap::{DashMap, DashSet};
+use demand::ProximityPrior;
 use freenet_stdlib::prelude::{ContractInstanceId, ContractKey};
 use parking_lot::RwLock;
 use std::collections::{HashMap, HashSet};
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 use tokio::time::Instant;
 use tracing::{debug, info};
 
+use super::Location;
 use super::interest::PeerKey;
 
 // =============================================================================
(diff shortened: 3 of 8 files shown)


sanity:
## Multi-model review — piece A3 (fuel-gauge hosting eviction)

Reviewers: **Codex** (external, non-Claude) across three passes as the diff evolved, plus an independent **adversarial Claude subagent** blind to Codex. Full tier (hosting is a high-risk surface). Reviewers read the checked-out code, not just the diff.

### Findings & resolutions

1. **[Codex P2 / adversarial MINOR] Local hit-rate was a cache-membership proxy** that miscounted a network refetch of a hosted-but-stale contract as a local hit. → **Fixed.** Moved the hit-rate counters out of `HostingCache` to the *actual* serve-vs-forward decision in the client GET handler (`client_events.rs`), via `HostingManager` serve/forward atomics surfaced through `Ring::record_get_served_locally/forwarded`. The cache no longer infers hits from membership.

2. **[Codex P2 / adversarial Obs] The `touch` local-serve path — the dominant read path for hot contracts — didn't train the proximity prior** (only network GETs did). → **Fixed.** `HostingCache::touch` now returns the observed read-rate sample and `HostingManager::touch_hosting` feeds it to the estimator, exactly like `record_contract_access`. New test `test_touch_hosting_trains_demand_estimator`.

3. **[Codex P2] Rolling window may not evict a pooled sample's contribution** (`remove_points` after PAVA merges). → **Dismissed with justification.** `pav_regression` 0.7.0's `remove_points` is explicitly written for this case: its rustdoc/impl subtract the removed point's *weighted influence* from the nearest aggregate block **and** decrement the centroid exactly, so the oldest sample's contribution is removed, not retained. This is the identical incremental rolling-window mechanism the production router's `IsotonicEstimator` uses; a full rebuild from `raw_points` per observation would be O(n log n) for no correctness gain. Added a code comment documenting this at the call site.

4. **[adversarial MINOR] Abandonment doc overstated `floor + demand > floor`** (the trained prior can predict exactly 0 for far contracts). → **Fixed** (doc now accurate: `>= floor`, tie resolved by the least-recently-accessed sequence).

5. **[adversarial MINOR] `read_count` can double-count a stale-refetch GET** (touch + refetch). → **Documented** as a bounded (2x) soft-signal inaccuracy on that path only; deduplicating would require threading GET-request identity, out of scope. Affects telemetry directionality, not eviction.

6. **[adversarial Obs] Eviction ordering depends on wall-clock** (`HostingManager` hardcodes `InstantTimeSrc`). Pre-existing for TTL/recency; A3 extends it into demand ordering. This is also why a deterministic full-network sim eviction test is impractical (see the PR's "On the simulation test"). Flagged as a sensible sim-time-plumbing follow-up.

### Verified correct (adversarial deep-dive, no issue)

Greedy-Dual floor ratchet is monotonic non-decreasing; the `(keep_score, access_seq, key)` order is total and NaN-free (`total_cmp`, keep_score always finite); age-0 new-entry protection holds; the subscribe / `contract_in_use` pin exempts even the lowest keep_score; PUT-seed earns no frontier credit and never trains; no lock-order inversion (cache write lock dropped before the estimator write); `ProximityPrior` is empty-safe, bounded, never NaN/negative; the three telemetry gauges are wired end-to-end and pinned; the "do NOT revert to byte-only LRU" guard is present; no `hosting-invariants` anti-pattern re-introduced; `-D warnings` clean on non-test targets.

### Verdict

**No BLOCKER/MAJOR findings.** All findings fixed or dismissed-with-justification. Locally: `cargo fmt` + `cargo clippy -p freenet` clean, hosting + telemetry lib tests green.

**PARKED for Ian's review — do NOT merge.**

[AI-assisted - Claude]

sanity:
## Review update — rebased onto `main` + post-rebase Codex pass

A2 (#4644) and #4649 merged, so this was rebased onto `main` and the base retargeted. Re-ran `codex review --base main` on the rebased diff.

### Rebase
Clean, **no conflicts**. A2's squash-merge (`398adaff`) is byte-identical to the A2 tip this branched from for every hosting file, so the A3 eviction changes replayed onto the merged budget code untouched. The only diverging file on main was `telemetry.rs` (#4649's resource-telemetry cadence throttle), whose changes don't overlap the A3 gauges — git 3-way-merged them cleanly and both coexist. Verified: `cargo clippy -p freenet` clean, hosting + telemetry lib tests green (158 passed, including #4649's new cadence test alongside the A3 tests).

### New finding from the post-rebase Codex pass

- **[Codex P0] "`ContractKey` doesn't implement `Ord`, so `a.0.cmp(&b.0)` won't compile" (`cache.rs` eviction tiebreak).** → **False positive, but addressed.** The crate *does* compile (clippy + 158 tests prove it): `ContractKey: Deref<Target = [u8; 32]>` (the instance-id bytes), so `.cmp` resolves via deref-coercion and orders by those bytes. Codex read the derive list (`Debug, Eq, Copy, Clone, Serialize, Deserialize` — no `Ord`) and missed the deref. Its underlying point is fair, though — relying on deref-coercion for `.cmp()` is implicit and fragile. **Fixed** by making both sort sites (`evict_over_budget` and the `keys_eviction_order` test helper) order explicitly on `a.0.as_bytes().cmp(b.0.as_bytes())` — the same instance-id byte string, so **behavior-equivalent** (and the tiebreak is only reachable in the transient pre-`finalize_loading` seq-0 case anyway; `last_access_seq` is otherwise unique per live entry).

### External-review availability

The **final confirmatory** `codex review` pass (on the behavior-equivalent `as_bytes()` tweak) hit an OpenAI **usage-limit / quota exhaustion** (resets ~12:30 PM local). Codex is the only working external reviewer (Gemini is deprecated). Note it had already run **four successful passes** on this change (surfacing and confirming the hit-rate, touch-training, and [P0] fixes); only the last confirmatory pass on a two-line behavior-equivalent change was blocked. Recommend the maintainer's independent review re-run `codex review --base main` after the quota resets as the final external check on the rebased diff.

Earlier consolidated review (hit-rate, touch-training, abandonment doc, remove_points, verified-correct list) is in the comment above. **No BLOCKER/MAJOR outstanding.** Still **PARKED — do not merge.**

[AI-assisted - Claude]

sanity:
## Pre-merge nit fixes (2 LOW)

1. **Weak/misleading #4338 test rewritten** (`fuel_gauge_keeps_repeatedly_read_contract_over_junk`). The old version touched `room` every round right before the junk insert with no time advance, so at eviction `room.age == 0` and it survived via the TTL/recency gate — it did NOT discriminate demand-ordering, and the docstring's "guaranteed by the Greedy-Dual floor, even with UNIFORM demand" was an overstatement (with uniform demand `keep_score = floor_at_refresh + const` is monotone in refresh time = LRU; the *demand* term is what discriminates). Went with **option (a)**: rewrote it so `room` is **past-TTL and the least-recently-accessed** of the two entries at eviction, yet survives because its `keep_score` (high demand) exceeds newer low-demand `junk`. The test now asserts, in-test, that `room.last_access_seq < junk.last_access_seq` **and** `room.keep_score > junk.keep_score`, then that `junk` (not `room`) is evicted.

   **Empirically confirmed it fails under byte-LRU-with-TTL:** temporarily reverting `evict_over_budget`'s sort to recency-only (`last_access_seq`, ignoring `keep_score`) makes the test FAIL — it evicts `room` (the oldest past-TTL entry) instead of `junk` (`assertion left == right failed: low-demand junk must evict, NOT the older-but-high-demand room`). Reverted the experiment; the test passes under the real demand-ordered eviction. The strong `evicts_lowest_demand_not_lru` test is unchanged.

2. **Stale comment fixed** (`router.rs`): the A3 hit/miss gauges' doc no longer calls them "the cache-membership proxy for the local hit-rate" — corrected to describe the real serve/forward decision in the client GET handler (`client_events`).

Gate on the new HEAD: `cargo fmt` / `clippy -p freenet` / `check` clean, cache/demand tests green. **Still PARKED — not merging.**

[AI-assisted - Claude]

PR #4652, opened 2026-07-01
fix: actionable recovery on hosted per-user-isolation error page

### LINKED ISSUE #4645: Suggestion and Issue with using try.freenet.org

sanity:
## Problem

On a hosted node (try.freenet.org), opening a Freenet app link as a **new
tab/window from inside the sandboxed app iframe** — the browser's "open link in
new tab", a middle-click, a right-click menu, `window.open`, or a
`target="_blank"` link — creates a top-level document that **inherits the iframe
sandbox**. With no `allow-same-origin` that document has an **opaque origin**, so
`localStorage` throws, so the shell can't read the per-user access token and
fails closed to the "Per-user isolation unavailable" page.

The old page said *"reconnect using an https:// address in a browser with storage
enabled"* — which is useless for this case: the tab already **is** https with
storage; the opaque origin is what blocks it. The user hits a dead end. Reported
in #4645.

Confirmed empirically against live try.freenet.org with Playwright: `window.open`,
`<form target="_blank">`, and `<a target="_blank">` from inside the iframe all
reproduce the error with `window.origin === "null"`; the same navigation from the
top window works.

## Solution

Rewrite the fail-closed page to detect the cause and give the right recovery:

- **Opaque-origin (sandboxed new-tab) case** — the common one — detected via
  `window.origin === 'null'`. Explains that the tab opened in a restricted mode,
  tells the user to open the address in a normal tab, and surfaces the URL with a
  one-click **Copy address**. Deliberately **no "retry" button**: a `window.open()`
  from this sandboxed context would inherit the sandbox and re-hit the same dead
  end, so the user must open a fresh top-level tab themselves.
- **Plain http case** → "use the https:// address".
- **Storage-disabled case** → "enable storage / use a different browser".

Only the rendered message + recovery affordance change. The fail-closed
**condition** (`hostedMode && !userToken`) and the second WS-open refusal barrier
are untouched, so the isolation guarantee is unchanged.

This is the pragmatic fix for the dominant, *uninterceptable* trigger (the
browser's own "open in new tab" fires no JS event, so no interceptor can catch
it). Making such tabs actually load would require moving the per-user token off
`localStorage` into a gateway-scoped cookie — a larger #4381 design change, noted
on the issue, not attempted here.

## Testing

- New Rust regression test `fail_closed_page_gives_actionable_recovery_4645` pins:
  opaque-origin detection, the URL-for-copy affordance, the three tailored
  headings, and the anti-footgun (the fail-closed block must not call
  `window.open`).
- ESLint + Prettier clean on the asset JS (`npm run lint` in `crates/core/src/server`).
- Rendering validated in a real browser against live try.freenet.org.

## Scope

Addresses the **bug** in #4645 (the error page). The issue also suggests an
"add new access key / identity" button in the Account popup — that's a small
follow-up PR, so this PR intentionally does **not** close #4645.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -3406,6 +3406,86 @@ mod tests {
         );
     }
 
+    /// Regression test for #4645: the hosted fail-closed page must give the
+    /// user an ACTIONABLE recovery path, not a dead end.
+    ///
+    /// The dominant real-world trigger is opening a Freenet app link as a NEW
+    /// TAB/WINDOW from inside the sandboxed app iframe (the browser's "open
+    /// link in new tab", a middle-click, a right-click menu, `window.open`, or
+    /// a `target=_blank` link). Such a context inherits the iframe sandbox, so
+    /// it has an opaque origin (`window.origin === 'null'`), so `localStorage`
+    /// throws and the per-user token can't be read — and the shell fails
+    /// closed. The pre-#4645 page only said "reconnect using https / enable
+    /// storage", which is useless for that case: the tab already IS https with
+    /// storage; the opaque origin is what blocks it. The page must instead
+    /// detect the opaque-origin case and tell the user to re-open the address
+    /// in a normal tab, surfacing the URL for one-click copy.
+    #[test]
+    fn fail_closed_page_gives_actionable_recovery_4645() {
+        // Detects the opaque-origin (sandboxed new-tab) case. The tell-tale is
+        // `window.origin` serializing to the string "null" for an opaque
+        // origin (confirmed empirically against try.freenet.org).
+        assert!(
+            SHELL_BRIDGE_JS.contains("window.origin === 'null'"),
+            "fail-closed page must detect the opaque-origin (sandboxed new-tab) \
+             case so it can give the right recovery guidance (#4645)"
+        );
+        // The "open in a normal tab" recovery only helps on a SECURE connection:
+        // over http even a fresh tab can't mint a token (SHELL_USER_TOKEN_JS
+        // refuses), so the https guidance must win when a page is BOTH sandboxed
+        // and plaintext. Pin that the reopen affordance is gated on
+        // `opaqueOrigin && !plaintext` rather than opaqueOrigin alone (Codex P3).
+        assert!(
+            SHELL_BRIDGE_JS.contains("opaqueOrigin && !plaintext"),
+            "the re-open recovery must be gated on a secure connection, so an \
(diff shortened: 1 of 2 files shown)


sanity:
## Review (pre-merge, multi-model)

Risk tier: **Light** (contained client-side JS/UX change; no change to the fail-closed condition, token handling, CSP, or the iframe sandbox) — plus one adversarial read given this is the hosted-shell surface.

**Passes run:** external Codex (`codex review --base origin/main`) + an independent adversarial Claude read of the full diff and surrounding code.

### Findings & resolution

- **[Codex P3] Combined sandboxed + plaintext case showed the wrong recovery.** If a hosted page were opened as a sandboxed new tab *and* served over `http://`, the opaque-origin branch won and told the user to re-open the same `http` URL — which still can't mint a token. **Fixed:** `plaintext` is now checked first (so the “Secure connection required / use https” guidance wins), and the “open in a normal tab” copy-URL affordance is gated on `reopenCase = opaqueOrigin && !plaintext` (secure + sandboxed only). Pinned by an added assertion.
- **[Claude, nit] Test block-boundary anchored on a comment string** (fragile to a future comment reword). **Fixed:** the regression test now anchors the block end on code (`iframe.getAttribute('data-src')`).
- **[Claude, nit] `role="alert"` wraps the interactive copy control** (`alertdialog` is technically more correct). **Kept `role="alert"` (justified):** the priority is announcing the error text, and a correct `alertdialog` needs full dialog semantics (accessible name + focus management) that are out of scope here; the copy button remains operable under `role="alert"`. The pre-existing page also used `role="alert"`.

### Confirmed clean (no blockers/majors)

- **No token leak:** the durable `userToken` lives only in `localStorage`, never in a URL; the shell `authToken` is stripped from top-level shell URLs by the gateway canonicalization filter — so the surfaced/copied `location.href` cannot contain either secret. It is the address the user can already see.
- **Fail-closed guarantee intact:** no early return added; the block still falls through to the `message` listener whose `open` case refuses while `hostedNoToken`. Iframe removal, WS-open refusal, CSP, and sandbox attributes are untouched.
- **Opaque-origin detection valid:** the iframe sandbox has no `allow-same-origin`/`allow-popups-to-escape-sandbox`, so an inherited-sandbox tab has an opaque origin and `window.origin` serializes to `"null"`; degrades safely if `window.origin` is `undefined`.
- `location.href` shown via `input.value` (property, not `innerHTML`) → no XSS; clipboard write guarded with a select-to-copy fallback for the sandboxed case.

[AI-assisted - Claude]

Issue #3676, opened 2026-03-27
investigation: low contract operation throughput despite active connections


sanity:
## Observation

A peer running for 12+ hours with 15 active connections accumulated only **6 routing events** (completed contract operations). Network-wide telemetry confirms this is not peer-specific — the entire network shows low contract operation throughput relative to connection activity.

### Connection activity (high):
| Event | Count |
|-------|-------|
| connect_connected | 144,527 |
| connect_request_received | 95,954 |
| connect_rejected | 39,168 |
| connect_request_sent | 24,142 |

### Contract operations (very low):
| Event | Count |
|-------|-------|
| get_request | 491 |
| get_success | 99 (20% success rate) |
| put_request | 87 |
| put_success | 31 |
| update_request | 90 |
| update_success | 21 |
| route_success_untimed | 77 |

### Key ratio:
**~145k connection events vs ~77 completed routing events.** The network is spending almost all of its effort maintaining connections, with very little contract operation throughput.

## Peer-level observation (framework laptop, 12h uptime):

- 15 connected peers
- 2.1 MB transferred
- 6 global routing events
- Logs show repeated "connect: uphill route timed out with no response" — connect forwarding is failing consistently across multiple peers

## Hypotheses (unverified)

1. **Low application usage** — there may simply be very few clients actively requesting contracts (GETs/PUTs). If nobody is using River or other apps, peers have no contract operations to route. *This would be expected behavior on a quiet network.*

2. **GET success rate is concerning** — only 99 of 491 GET requests succeeded (20%). This could indicate a routing/topology problem where GETs are being sent to peers that don't have the requested contracts, or that the subscription trees aren't well-formed. *However, this could also be normal if many GETs are for contracts that haven't been PUT yet.*

3. **Connect forwarding failures** — the framework peer's logs show consistent "uphill route timed out" for connect operations. If connect forwarding is unreliable, this could prevent proper topology formation, which in turn would hurt contract operation success rates. *This could be a cause or a symptom.*

4. **The routing model can't activate** — the Router needs 50 events to switch from distance-based to prediction-based routing. At current throughput, this would take days. The renegade predictor needs 10+ events before it can make any predictions. *This is a consequence, not a cause, but it means the ML-based routing improvements can't help until the throughput issue is addressed.*

## What's NOT the issue

[...]

PR #4655, opened 2026-07-01
feat: event-driven re-subscribe on upstream connection drop

sanity:
> [!WARNING]
> **PARKED for Ian's review — do NOT merge.** This is piece F of the demand-driven hosting redesign bundle (epic #4642) and targets **0.2.91** (after 0.2.90 validates). It touches the subscription / connection-drop recovery path, so it is intentionally held for maintainer sign-off rather than auto-merged.

## Problem

Subscriptions are lease-based and, since the 2026-01 refactor, deliberately not tied to a peer in the `HostingManager`. When the upstream peer a subscription is routed through drops, the local lease stays valid and updates stop flowing until the periodic renewal cycle re-routes it (~2-min renewal / 8-min lease — a minutes-long stale window). Recovery is entirely lease-timer-driven; nothing reacts to the drop *event* itself.

This is **piece F** of the redesign ("demand-gated re-rooting"): a peer whose upstream vanishes should re-subscribe immediately (a fresh SUBSCRIBE routes toward the key, re-rooting the chain keyward — self-healing, invariant 4), bounded by the existing recovery caps rather than waiting for the timer.

## Approach

`OpManager::on_ring_connection_lost` (the single chokepoint for ring-connection loss) now, **in addition to the unchanged deferred interest cleanup**, immediately re-subscribes for every contract where the dropped peer was our **upstream**:

- **Upstream selection** — the contract→upstream mapping already exists in `InterestManager` (`PeerInterest.is_upstream`). New `upstream_contracts_for_peer` filters the reverse index to those entries, snapshotting `peer_contracts` before touching `interested_peers` to avoid the `register_peer_interest` lock-order inversion (see `ring.md` cross-DashMap discipline).
- **Reuses existing renewal machinery** — the renewal loop's guarded-spawn block is extracted into a shared `Ring::spawn_guarded_renewal_task` so the periodic loop and the event path are byte-identical (single source of truth; pinned). `try_spawn_event_driven_resubscribe` applies the **same** bounding as the loop: ban gate (#4373), `can_request_subscription` (exponential backoff), `mark_subscription_pending` (per-contract dedup — **shared, atomic** state, so the event path and the 30s loop cannot double-fire).
- **Bounded fan-out** — a high-degree upstream can be the upstream for many contracts, so the per-drop burst is shuffled and capped to `MAX_RECOVERY_ATTEMPTS_PER_INTERVAL` with the same channel-backpressure short-circuit the renewal loop uses; the periodic loop recovers any capped remainder (never dropped). Small 0–1s jitter keeps the update gap minimal.
- **Storm-safe** — zero-connection gate mirrors the loop's #3676 guard. No wire / stdlib change. **The periodic timer path's behavior is preserved** (its inline spawn was extracted into the shared `spawn_guarded_renewal_task` that both paths now call — pinned by `renewal_paths_route_through_shared_spawn_helper`), and it remains the backstop.

**Instrumentation** (per-node scalars only): `event_driven_resubscribes` in `network_status` (surfaced on the dashboard SUBSCRIBE card as "N re-rooted") + a sim-observable per-node counter in `topology_registry::RenewalMetrics`.

### Invariant check (hosting-invariants.md)

Recovery for **real** subscriptions (active demand), **single-target** (routes toward the key), **rate-limited** (shared dedup + backoff + burst cap), timer path preserved as backstop — exactly the self-healing re-rooting invariant 4 permits, and re-introduces **no** anti-pattern (no holding-driven push, no relay-caching, no unsubscribed durable cache).

## Testing

Deterministic coverage:
- `test_upstream_contracts_for_peer` — `is_upstream` filtering (excludes downstream-only and other peers).
- `test_event_driven_resubscribe_recording` — counter accumulation + snapshot.
- `mark_subscription_pending_reserves_atomically_under_contention` — 32-thread contention test pinning the atomic reservation (no double-fire).
- Source-scrape pins: `renewal_paths_route_through_shared_spawn_helper` (both paths use the shared helper, neither hand-inlines the driver), `event_driven_resubscribe_applies_dedup_and_backoff_before_spawn` (ban→backoff→dedup ordering), the updated #4373 ban-gate pin, and `on_ring_connection_lost_wires_event_driven_resubscribe` (drop-handler wiring: zero-conn gate + fan-out cap + counters).
- 173 interest/hosting/subscription/renewal unit tests pass; `cargo fmt` + `cargo clippy` clean (no new warnings).

Multi-model review (Codex + 2 adversarial Claude passes) ran and found two MAJOR issues (double-fire race + uncapped fan-out), both fixed in `6c026372`; the re-review is clean. Full write-up in the review comment below.

### Note on the behavioral sim test (for maintainer)

A full-network `run_simulation` crash test was built and investigated but is **not reliably deterministic** for this specific behavior, for three concrete harness reasons:
1. **`is_upstream` is sparse in random-event sims** — `finalize_originator_subscribe` only records a peer-upstream when `get_peer_by_addr(upstream_addr)` resolves and `upstream_addr != own_addr`; many sim subscribes resolve at the root (upstream = self), so no peer-upstream is recorded to crash.
2. **Silent-crash detection floor** — a `crash_node`'d (packet-dropping) peer is only pruned by the ring's periodic health check (`HEALTH_CHECK_INTERVAL` = 300s); the transport-close arm only logs. So the event can't fire before ~300s in that scenario.
3. **No single runner supports both** scripting a known subscription tree (`run_controlled_simulation`) **and** crashing a peer mid-run (`run_simulation`).

Diagnostics confirmed the mechanism is wired correctly (`on_ring_connection_lost` fires with `open_connections > 0`; it just never saw a crashed peer that was a recorded upstream). Rather than land a flaky sim test (flaky = broken), the end-to-end wiring is pinned deterministically via source-scrape (the codebase's established pattern for Ring/OpManager logic lacking behavioral scaffolding — same reason `ring.rs` pins the renewal ban-gate). **Happy to add a real behavioral test if the harness gains scripted-subscribe-plus-mid-run-crash, or a way to force a peer-upstream.** Flagging for your call.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -187,6 +187,12 @@ pub struct OperationStats {
     /// Count of broadcast updates received via subscription streaming.
     /// These are push-based and don't have success/failure semantics.
     pub updates_received: u32,
+    /// Count of event-driven re-subscribes this node fired because an UPSTREAM
+    /// peer dropped (#4642 piece F). Each is a fresh SUBSCRIBE routed toward the
+    /// key immediately on upstream-loss detection (after the shared
+    /// ban/backoff/dedup gates), rather than waiting for the periodic renewal
+    /// cycle — a per-node measure of how often self-healing re-rooting fired.
+    pub event_driven_resubscribes: u64,
 }
 
 /// Maximum number of recent NAT attempts to track for rolling trend.
@@ -457,6 +463,19 @@ pub fn record_update_received() {
     if let Some(status) = NETWORK_STATUS.get() {
         if let Ok(mut s) = status.write() {
             s.op_stats.updates_received = s.op_stats.updates_received.saturating_add(1);
+        }
+    }
+}
+
+/// Record an event-driven re-subscribe fired on upstream-peer loss (#4642
+/// piece F). Per-node aggregate scalar; incremented once per fresh SUBSCRIBE
+/// spawned by `OpManager::on_ring_connection_lost` for a contract whose upstream
+/// just dropped.
+pub fn record_event_driven_resubscribe() {
+    if let Some(status) = NETWORK_STATUS.get() {
+        if let Ok(mut s) = status.write() {
+            s.op_stats.event_driven_resubscribes =
+                s.op_stats.event_driven_resubscribes.saturating_add(1);
         }
     }
 }
@@ -795,6 +814,8 @@ pub struct OpStatsSnapshot {
     pub subscribes: (u32, u32),
     /// Broadcast updates received via subscription streaming.
     pub updates_received: u32,
+    /// Event-driven re-subscribes fired on upstream-peer loss (#4642 piece F).
+    pub event_driven_resubscribes: u64,
 }
 
 impl OpStatsSnapshot {
@@ -972,6 +993,7 @@ pub fn get_snapshot() -> Option<NetworkStatusSnapshot> {
             updates: s.op_stats.updates,
             subscribes: s.op_stats.subscribes,
             updates_received: s.op_stats.updates_received,
+            event_driven_resubscribes: s.op_stats.event_driven_resubscribes,
         },
         nat_stats: NatStatsSnapshot {
             attempts: s.nat_stats.attempts,
@@ -1244,6 +1266,34 @@ mod tests {
         assert_eq!(snap.op_stats.puts, (1, 0));
         assert_eq!(snap.op_stats.subscribes, (1, 1));
         assert_eq!(snap.op_stats.updates, (2, 1));
+    }
+
+    #[test]
+    fn test_event_driven_resubscribe_recording() {
+        // #4642 piece F: the per-node event-driven re-subscribe scalar must
+        // accumulate through `record_event_driven_resubscribe()` and surface in
+        // the snapshot.
+        let _lock = TEST_MUTEX.lock().unwrap();
+        init(31346, HashSet::new(), "0.1.148".to_string());
+
+        if let Some(status) = NETWORK_STATUS.get() {
+            let mut s = status.write().unwrap();
+            s.op_stats = OperationStats::default();
+        }
+
+        assert_eq!(
+            get_snapshot().unwrap().op_stats.event_driven_resubscribes,
+            0
+        );
+
+        record_event_driven_resubscribe();
+        record_event_driven_resubscribe();
+        record_event_driven_resubscribe();
+
+        assert_eq!(
+            get_snapshot().unwrap().op_stats.event_driven_resubscribes,
+            3
+        );
     }
 
     #[test]
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -1226,9 +1226,10 @@ impl OpManager {
                 // Production per-node scalar (aggregate counter on this node).
                 crate::node::network_status::record_event_driven_resubscribe();
                 // Simulation-test observability (no-op in production — gated on a
-                // current network name, like the renewal-cycle counters). Lets a
-                // sim assert this node fired an event-driven re-subscribe promptly
-                // after the upstream drop, well before the lease-renewal window.
+                // current network name, like the renewal-cycle counters). Records
+                // per-node so a simulation harness can observe event-driven
+                // re-subscribes; no behavioral sim consumes it yet (see #4642
+                // piece F PR — a reliable one is blocked on harness support).
                 if let Some(addr) = own_addr {
                     crate::ring::topology_registry::record_event_driven_resubscribe(addr);
                 }
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -1155,10 +1155,64 @@ impl OpManager {
     /// `expire_stale_downstream_subscribers` sweep, which also decrements the
     /// interest manager's `downstream_subscriber_count` and triggers upstream
     /// unsubscribe when appropriate.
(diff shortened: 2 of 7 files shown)


sanity:

Reviewed on HEAD `6c026372`. Reviewers: **Codex** (external, non-Claude) + two adversarial Claude passes (skeptical bug-hunt, code-first), then a targeted re-review of the fixes. This is a Full-tier change (subscription / connection-drop recovery — a high-risk surface).

### Findings and resolutions

**MAJOR — double-fire race (fixed in `6c026372`).** `mark_subscription_pending` was a non-atomic `contains`-then-`insert` on a `DashSet`. Harmless before this PR (the only caller was the single sequential periodic renewal loop), but piece F adds a second **concurrent** caller (`on_ring_connection_lost` on the event-loop task), so on the multi-threaded runtime two tasks could both reserve the same contract and both spawn a re-subscribe — the exact "double-fire" the design forbids. Fixed by returning `DashSet::insert`'s atomic newly-inserted bool. Added a 32-thread barrier-synchronised contention test asserting exactly one reservation wins.

**MAJOR — uncapped fan-out / subscribe storm (fixed in `6c026372`; Codex + bug-hunt agreed).** The event path spawned a fresh SUBSCRIBE for *every* affected contract at once (0–1s jitter), bypassing the periodic loop's `MAX_RECOVERY_ATTEMPTS_PER_INTERVAL` cap and channel backpressure — a one-drop → N-SUBSCRIBE fan-out on high-degree upstream loss (violating the "fan-out MUST cap the cost" rule). Now shuffles `affected`, caps the per-drop burst to the same `MAX_RECOVERY_ATTEMPTS_PER_INTERVAL`, and stops early when the notification channel is >75% full (mirrors the loop's `RENEWAL_STOP_CAPACITY_FRACTION`). The periodic loop (`contracts_needing_renewal`) remains the backstop for the shuffled remainder, so capping never *drops* a subscription — it only smooths the spike.

**LOW-MEDIUM — weak pin (fixed).** The `on_ring_connection_lost` wiring pin only asserted the substring `open_connections`, which was satisfied even if the actual `if open_conns == 0 { return; }` storm gate were deleted. Strengthened to pin the real `open_connections() == 0` gate + its ordering before the spawn, and to assert the fan-out cap consts are applied.

**LOW — efficiency (fixed).** Zero-connection gate moved before the interest scan (skip the scan on an isolated node).

**LOW — wording (fixed in PR body).** "timer path untouched" was literally inaccurate (the loop's inline spawn was extracted into the shared `spawn_guarded_renewal_task`). Reworded to "behavior-preserving extraction" — the byte-identical behavior is itself pinned by `renewal_paths_route_through_shared_spawn_helper` and the updated #4373 ban-gate pin.

**NIT — won't-fix-here.** A doc reference to the ecosystem-level `hosting-invariants.md` (which lives outside the freenet-core repo) is pre-existing; my code only cites "invariant 4" conceptually and references the in-repo `ring.md` for the lock rule.

### Verified correct (not rubber-stamped)

Helper extraction is byte-identical (renames only); `upstream_contracts_for_peer` lock order is safe (snapshots `peer_contracts` before touching `interested_peers`); all 5 `on_ring_connection_lost` call sites compile with `self: &Arc<Self>`; deferred interest cleanup preserved as backstop; re-subscribe routes greedily toward the key (finds a NEW upstream, not the dropped one); counters fire once per spawn; no anti-pattern reintroduced (fires only for real `is_upstream` subscriptions, single-target).

### Re-review outcome

Codex on the fix commit: *"I did not identify any discrete correctness, safety, or maintainability issues."* Adversarial re-review: *"Both fixes are correct and complete, and I found no new bug"* (atomic-insert semantics preserved for all callers; channel-capacity check is a faithful mirror of the periodic loop; capped tail is genuinely backstopped by `contracts_needing_renewal`, not dropped).

Still **PARKED — do NOT merge.** Piece F of the #4642 bundle, targets 0.2.91.

[AI-assisted - Claude]

sanity:
Flagging a file overlap so we don't collide.

A dashboard audit turned up #4830 (cards describing mechanisms that don't exist: the ban-list card claims "the mechanism is active" when both ban writers are unreachable; the hosting card still advertises "lowest keep-score first" though eviction went subscriber-primary; the "Conn attempts" tile counts only *failed* attempts). The fix touches `crates/core/src/server/home_page/cards.rs` and `crates/core/src/node/network_status.rs` — both also touched here.

This PR has been idle since 2026-07-01, so rather than block #4830 on it indefinitely or open a conflicting PR, flagging for a call on ordering:

- if this is still live and lands soon, I'll wait and rebase #4830 on top;
- if it's parked, I'd rather land #4830 first (it's a small copy/label change) and let this one rebase over it.

Happy either way — just don't want two branches editing the same cards.

[AI-assisted - Claude]

sanity:
Closing as **superseded-in-approach** (not lost).

This PR (piece F) was built on the stored `is_upstream` flag. The epic #4642 reconcile-core keystone **removed** that stored flag in favor of a *computed* upstream, and piece F's behavior was reimplemented on that substrate and shipped: **#4745 — "drive storm-safe prompt re-root on connection drop (#4642 piece F)"** (merged 2026-07-09, released in 0.2.95). Same purpose (immediate re-root on upstream drop, storm-safe, bounded), now driven as a reconcile action rather than off the `is_upstream` reverse index.

Per #4642: earlier in-flight PRs (B #4663, D #4661, E #4664, F #4655) followed the stored-flag framing the reconcile core removes; their validated behaviors were harvested onto the computed-upstream substrate rather than merged whole.

This branch also can't merge as-is (it's DIRTY, and would re-introduce the removed `is_upstream` flag). Closing unblocks #4830, which was waiting on this PR's fate to avoid colliding on `cards.rs` / `network_status.rs`.

[AI-assisted - Claude]

PR #4663, opened 2026-07-02
feat: admission-refusal for over-committed hosts (piece B, #4642)

sanity:
## Problem

Piece D (computed-upstream demand-driven subscription chains, #4661) makes every peer on a live subscription chain a **real host** — but nothing bounds how many downstream subscribers a single host takes on. D's own sim proof `test_popular_contract_fanout_is_measured` is deliberately MEASURE-ONLY and notes: *"fan-out is UNBOUNDED here by design — admission/piece B is not in this piece."* A popular contract funnels subscribers toward its key, and the near-key hosts accumulate unbounded update fan-out (the wide-star broadcast cost of #4233/#3309).

Piece B (hosting-invariants §B, tracker #4642) is the admission side of the capability-relative budget: **an over-committed peer REFUSES a new subscription rather than over-committing and thrashing.** That refusal is exactly what bounds per-peer fan-out.

## Approach

An admission check on the subscribe-accept path, gating the two sites in `drive_relay_subscribe` where a peer would take on a **new** downstream subscriber and answer/bubble `Subscribed`:

1. **Local-hit terminus** (this peer already hosts the contract) — cost is update fan-out only.
2. **Chain-host commit** (this peer would newly become a host after a downstream `Subscribed`) — cost is a new hosted contract (bytes) + fan-out.

`Ring::can_admit_subscription(contract)` refuses (→ answers `SubscribeMsgResult::NotFound`, the rejection the subscribe path already handles — **no wire-format change**) when either capability-relative signal has no headroom:

- **Fan-out capacity** — aggregate lease-valid downstream subscribers across all contracts ≥ a RAM-scaled `fanout_capacity`. This is the term that bounds popular-contract fan-out.
- **Byte-budget headroom** (reuses the A3 gauge) — only for a *not-yet-hosted* contract: if the hosting byte-budget is already saturated, admitting a new contract would just thrash eviction. An already-hosted contract adds fan-out but no new bytes, so the byte term doesn't apply to it.

A refusal makes the requester re-route to another acceptor via the existing retry plumbing (`advance_to_next_peer` → fresh `k_closest`); if every closer peer refuses, the originator remains a valid stranded host (design §5d). **No wire change, no deadlock, no dead-end.**

Composition with the model (no improvisation on the "no relay category" rule):
- **Renewals are never refused** (`is_renewal` bypasses the check) — they maintain an existing demand-pinned host and add no new fan-out; refusing one would collapse a live chain.
- **The originator's own local-client subscribe is never refused** — a peer always hosts its own client's real demand (the stranded-host backstop).
- **Only REFUSES** — never manufactures demand or pushes state, so no anti-pattern (relay-caching / holding-driven push) is re-introduced.
- The downstream-subscriber registration was moved out of `relay_subscribe_forward_once` into the caller's accept branch so a refusal never has to un-register a downstream it already added. Piece D's pins (`subscribe_driver_becomes_host_on_subscribed_reply`, `subscribe_driver_forwards_upstream_before_hosting`) still hold; all four D sim proofs still pass.

## Instrumentation (horizontal, per the hosting-eviction design)

Three per-node aggregate gauges on the periodic `RouterSnapshot` (JSON-mirrored + pin-tested): `hosting_fanout_capacity`, `hosting_total_downstream_subscribers`, `hosting_subscription_refusals_total`. These let production telemetry observe whether admission is biting and whether the capacity is tuned right.

## ⚠️ One tuning decision flagged for review

The design doc fixes the **byte** budget concretely (A2: RAM/8) but leaves the **update-fan-out capacity** numeric scaling open — there is no shipped uplink/CPU capability signal. I scaled it off the same RAM signal (`default_fanout_capacity` = `clamp(total_ram / 8 MiB, 64, 4096)`: ~64 for a 512 MiB box, ~1024 at 8 GiB, ~4096 at ≥32 GiB — the ceiling stays comfortably above the 512 per-contract hard cap so a capable node can still fully host one maximally-popular contract). **The mechanism is faithful to the model; the exact scaling is the judgment call I'd like reviewed.** It is a per-node atomic (`set_fanout_capacity`), so the sim overrides it to a small value and a production operator override is a trivial follow-up. `hosting_subscription_refusals_total` is the instrument for tuning it post-deploy.

## Testing

**Unit** (`ring/hosting.rs`): RAM-scaled capacity clamp math; aggregate downstream count (lease-valid across contracts, stale drops out); accept-with-headroom; refuse-at-fan-out-capacity (+ refusal counter); refuse-new-contract-over-byte-budget; allow-already-hosted-when-byte-budget-full.

**Simulation** (`simulation_integration.rs`, 3 seeds each): fan-out is emergent/seed-dependent, so each proof runs the popular-contract scenario **twice on the same topology** — uncapped to measure the natural peak, then capped below it:

- **Proof 1 — bounded fan-out** (`test_popular_contract_fanout_bounded_by_admission`): capped peak fan-out drops below the natural peak, with refusals recorded. Observed (natural → capped peak): `5→2`, `2→1`, `3→1`; refusals `4/5/10`.
- **Proof 2 — refused subscription still finds a host** (`test_refused_subscription_reroutes_no_deadend`): despite refusals, the contract ends up hosted on multiple peers and the subscribed set does not collapse (stays ≥ half the uncapped baseline). Observed: capped subscribed set **kept or grew** vs uncapped (`9≥7`, `12≥11`, `11≥11`) — refused subscribers re-route and the mesh spreads to more hosts.

Also fixed a **pre-existing stale pin** on the base: `finalize_originator_subscribe_contains_all_required_side_effects` was scraping `finalize_originator_subscribe` for side effects that piece D moved into the shared `finalize_host_subscribe` helper. Split into a pin on `finalize_host_subscribe` (the #4223/#3851 bug class at its new home) + a delegation pin.

`cargo fmt` + `cargo clippy --locked -- -D warnings` clean.

## Status

Stacked on **#4661 (piece D)** — base branch `feat/computed-upstream-chains`. Part of the B-through-F bundle (target 0.2.91); **DRAFT, do not merge** until the bundle is validated as a whole and A is confirmed live in the field (release-order constraint in hosting-invariants). Parked for @sanity's review.

Tracker: #4642

[AI-assisted - Claude]


sanity:

Reviewers: **Codex** (external, non-Claude) + a **Claude adversarial reviewer** (independent, told to assume bugs exist). Both reviewed the checked-out code against `feat/computed-upstream-chains`.

### Codex
No discrete correctness / security / maintainability findings on the piece-B diff. (It could not run `cargo check` — sandbox has no crates.io — but reviewed the diff itself.)

### Claude adversarial reviewer — no BLOCKER; 1 MAJOR + minors, all addressed

Verified-correct by the reviewer: `is_renewal` bypasses the gate at all three sites; the moved `register_downstream_subscriber` is a faithful relocation (registers exactly once per accept, never on refuse, before the upstream bubble); `finalize_host_subscribe` registers the *upstream*, so no double-registration; the originator's own subscribe is never gated; the two-run sim comparison is topology-stable (`GlobalRng` reseeds from the numeric seed, not the network name); the sim proofs have teeth on both failure modes.

Findings and fixes (commit 3718b176):

- **MAJOR — inaccurate "stranded host" comments.** The comments claimed a refused-everywhere subscriber "remains a valid stranded host (§5d)", but the originator actually **hard-fails on exhaustion** (`drive_client_subscribe_inner` returns `OperationError`; it does not auto-host). Corrected the comments at the local-hit site and in the `can_admit_subscription` doc to describe the real behavior: refusal → re-route; total exhaustion → subscribe fails (genuine capacity back-pressure). Piece B relies on piece D's chain-host multiplicity + the generous default capacity so global saturation is rare, and **must not ship before D is live** (already the hosting-invariants §B release order). The reviewer noted the sim proofs run a well-connected mesh and cannot exhibit a single-path strand — I've documented the reliance rather than add a test that would "prove no dead-end" in a scenario where the strand *is* the designed back-pressure.

- **MINOR — `can_admit_subscription` mutated telemetry despite a query name.** Made it side-effect-free; the refusal counter now bumps only via `record_subscription_refused` at the actual refusal sites (Ring + HostingManager wrappers), so a future speculative "would this be admitted?" caller can't inflate production telemetry. Unit tests updated + a dedicated counter test added.

- **MINOR — fan-out fallback RAM constant diverged** from the byte budget's (512 MiB vs 1 GiB), breaking the "one notion of memory" principle. Now reuses `cache::FALLBACK_TOTAL_RAM_BYTES` (1 GiB).

- **MINOR (accepted, documented) — refusal after the downstream host committed** leaves a phantom downstream registration for one lease duration, self-healing via interest-gated renewal. Noted in-code; the alternative (gate before forwarding) would lose the host-exists signal.

- **NIT — register-on-all-paths pin was `>= 2`** with three accept sites now; tightened to `>= 3`. Plus a test-comment fix and a shared brace-scrape helper.

Not changed (justified): `total_downstream_subscribers()` is O(total subscribers) per fresh subscribe — a perf watch-item at extreme scale, not a correctness issue (documented); `as usize` cast in the RAM math is non-UB on 64-bit and the clamp bounds the result.

All fixes verified: `cargo fmt` + `cargo clippy --locked -- -D warnings` clean; 70 hosting/subscribe unit tests pass; sim proof 2 re-ran identically after the refactor (behavior unchanged).

[AI-assisted - Claude]

sanity:
Closing as superseded-in-approach.

The settled #4642 model removed the separate admission decision entirely — refusal is emergent from demand-ordered eviction (no "I'm full" NACK, no chain cap). The equivalent behavior shipped via #4717 (admission-decision core, subscriber-primary) and #4720 (subscriber-primary eviction).

This PR was also stacked on #4661 (piece D), which is being closed for the same reason. It's conflicting and clippy-red on a base branch that's going away.

[AI-assisted - Claude]

PR #4664, opened 2026-07-02
feat: remove relay-caching + GET-auto-subscribe (piece E, #4642) [BLOCKED: PUT half needs D fetch hardening]

sanity:
## Status: DRAFT — DO NOT MERGE. Parked for Ian's review.

Piece E of the demand-driven hosting bundle (#4642), stacked on piece B (#4663) which is stacked on piece D. **The GET half is clean and green; the PUT half is BLOCKED by a design-model gap in piece D** (details below). Bundle target 0.2.91.

## What piece E removes (per hosting-invariants anti-patterns table + demand-driven-hosting §9)

1. **GET-auto-subscribe** (`AUTO_SUBSCRIBE_ON_GET`): a successful GET with `subscribe=false` no longer auto-installs a durable subscription. Manufactures demand no client asked for.
2. **PUT relay-caching** (store-at-every-hop + the #4509 terminus `relay_put_replicate_forward`): a pure routing hop no longer stores/hosts/announces an unsolicited copy.

## What is KEPT (demand-driven replacements, NOT the anti-pattern)
- **Host-on-GET** under piece A's demand gauge (`cache_contract_locally`) — evictable, non-durable.
- **Explicit `subscribe=true`** on GET/PUT (routes through `maybe_subscribe_child` / `start_subscription_after_put`).
- **PUT seeds near the key**: terminus/originator/already-hosting still store+host+announce (`should_host` gate).
- **Downstream-failure correctness** (streaming PUT): a routing-only hop that pipes forward and then loses the downstream now reports failure upstream (new `relay_put_streaming_downstream_failure`) instead of falsely claiming success, so the originator retries rather than silently losing the PUT.

## Two commits (deliberately split)
- `feat: remove GET-auto-subscribe` — **CLEAN. All sim proofs pass.**
- `feat: remove PUT relay-caching` — **BLOCKED. Breaks D's + B's sim proofs (see below).**

## Sim evidence (this is the STOP finding)

Ran the D/B/convergence proofs on three points (nextest, `simulation_tests,testing,nightly_tests`):

| Point | Result |
|---|---|
| Base (`feat/admission-refusal`, D+B) | 9/9 PASS |
| This branch @ **GET-only** commit | 5/5 PASS |
| This branch @ **PUT** commit | 3 pass, **8 FAIL** |

The failures are convergence/propagation: `test_six_peer_contract_lifecycle`, `test_six_peer_lifecycle_mock_wasm`, `test_subscription_relay_propagation`, and D's own proofs (`test_subscription_count_tracks_demand_not_cache`, `..._chain_collapses_on_client_leave`, `..._chain_collapse_reaches_root_no_cycle`) + B's (`..._popular_contract_fanout_bounded_by_admission`, `..._refused_subscription_reroutes_no_deadend`). Symptom: `missing contract ... failed to apply update locally before forwarding`, and `only N peer(s) have state`.

## Root cause: a piece-D gap that PUT relay-caching was masking

D's chain-host state fetch lives in `subscribe.rs::finalize_host_subscribe`: it calls `fetch_contract_if_missing` with a **2s best-effort timeout**; on timeout it defers `announce_contract_hosted` and comments that "the contract may arrive later via UPDATE propagation." But an UPDATE delivered to a **subscribed-but-body-less** node fails with `missing contract`, so that fallback does not actually heal it.

With PUT relay-caching in place (base), the body was already local at those nodes, so the fetch found it instantly and the proofs passed — the proofs were **contaminated by relay-caching**. Remove relay-caching and the chain-host fetch frequently fails to land the body in these topologies, leaving subscribed-but-body-less nodes → convergence breaks.

So piece E's PUT half is correctly gated on D, but **D as implemented in the base does not yet make chain hosts fetch state reliably** — its "real chain hosts fetch state" (design §3) is a 2s best-effort fetch leaning on relay-caching. Per hosting-invariants ("STOP+report on any design-model gap — do NOT improvise"), I stopped rather than hardening D's fetch (out of scope for a removal, and high-risk hosting surface).

## Recommendation for Ian
- **Land the GET-auto-subscribe removal now** (drop the PUT commit; it's clean + green), OR
- **Hold both** until piece D's chain-host fetch is hardened (blocking/retried fetch, and/or fetch-then-apply on UPDATE-to-body-less-node), then re-run these proofs. Either way the PUT half must not merge until D's proofs pass WITHOUT relay-caching.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -4852,9 +4852,12 @@ mod tests {
         // make a peer a first-viable-target via `register_peer_interest` and
         // are correctly flush-paired. A NEW production call site in any other
         // file must FAIL this pin — see the set-equality assertion below.
+        // NOTE: "operations.rs" was dropped from this set in piece E of the
+        // demand-driven hosting redesign — its only register_peer_interest call
+        // lived in `complete_piggyback_subscription` (the GET-auto-subscribe
+        // piggyback finalizer), which was removed with GET-auto-subscribe.
         let known_wired: std::collections::BTreeSet<&str> = [
             "node.rs",
-            "operations.rs",
             "operations/subscribe.rs",
             "operations/get/op_ctx_task.rs",
         ]
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -700,7 +700,6 @@ impl OpManager {
     /// * **Source 2 — interest manager** (`register_peer_interest` is_new):
     ///   - `subscribe::register_downstream_subscriber` (downstream subscriber)
     ///   - `subscribe::finalize_originator_subscribe` (originator upstream)
-    ///   - `operations::complete_piggyback_subscription` (GET-piggyback sub)
     ///   - `get::op_ctx_task` remote GET `subscribe=false` (requester interest)
     ///   - `node.rs` Interests / Summaries interest-sync handlers
     /// * **Source 1 — proximity cache** (`neighbors_with_contract`):
--- a/crates/core/src/operations.rs
+++ b/crates/core/src/operations.rs
@@ -466,78 +466,18 @@ pub(crate) fn reclaim_evicted_contract(
     );
 }
 
-/// Complete subscription at the originator node via GET piggyback.
-///
-/// The subscription tree was built by relay nodes during GET response propagation.
-/// This function performs the local registration at the originator:
-/// 1. Mark subscribed (lease in active_subscriptions)
-/// 2. Register the upstream peer as an interest source
-/// 3. Register local interest so ChangeInterests from peers are processed
-/// 4. Announce contract hosted so neighbors send UPDATEs
-pub(crate) async fn complete_piggyback_subscription(
-    op_manager: &OpManager,
-    key: &ContractKey,
-    tx: &crate::message::Transaction,
-    sender_from_addr: &Option<crate::ring::PeerKeyLocation>,
-) {
-    op_manager.ring.subscribe(*key);
-    op_manager.ring.complete_subscription_request(key, true);
-
-    // Register local interest so ChangeInterests from peers get properly processed.
-    let became_interested = op_manager.interest_manager.add_local_client(key);
-    if became_interested {
-        broadcast_change_interests(op_manager, vec![*key], vec![]).await;
-    }
-
-    // Announce that we host this contract so neighbors include us in broadcast targets.
-    announce_contract_hosted(op_manager, key).await;
-
-    if let Some(upstream_pkl) = sender_from_addr.as_ref() {
-        let peer_key = crate::ring::interest::PeerKey::from(upstream_pkl.pub_key.clone());
-        let is_new = op_manager
-            .interest_manager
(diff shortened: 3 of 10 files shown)


sanity:
## Review pass (Codex external + adversarial Claude self-review)

### Codex (external) — 1 P1, FIXED
**[P1] Streaming PUT could finalize as success without storing or forwarding** (`put/op_ctx_task.rs`). When a streaming relay selected a next hop but did NOT set up the piped forward (payload below streaming threshold, unresolved next-hop address, or `send_to_and_register_waiter` failure), `downstream_reply_rx` is `None` and the Step-7 `else` branch reports a successful `PutMsg::Response` — but the old gate `should_host = originator_loopback || finalize_here(next_hop.is_none()) || already_hosting` could be false there (next_hop was `Some`), so the node neither stored nor forwarded yet claimed success → silent write loss.

Fixed in `dbf85fbd`: gate hosting on `downstream_reply_rx.is_none()` (not-forwarding = effective terminus) instead of `next_hop.is_none()`. A hop that will not await a downstream reply now hosts, so the success reply is truthful; a hop that IS piping stays pure routing. Covered by PUT unit tests + the streaming terminus-guard pin.

### Adversarial Claude self-review — no further issues
- Non-streaming path does NOT have the equivalent bug: it always forwards when `forward_target = Some` (Step 5) and returns `Err` on downstream failure (→ originator retries); when `forward_target = None` it hosts (`finalize_here` ⟹ `should_host`).
- `merged_value` forward semantics: routing-only hop forwards the raw client value, hosting hop forwards the merged value — both correct under the commutative-monoid model (intermediate merge is redundant).
- `already_hosting` clause only ever merges into an EXISTING copy (a real host in the mesh); it never creates a new copy at a routing hop — so it is not relay-caching.
- Move/borrow of `contract` / `related_contracts` / `value` across the conditional store is sound in both drivers.

### Still BLOCKED
The Codex fix is orthogonal to the blocker. The PUT half still breaks D's + B's sim proofs because of the piece-D chain-host-fetch gap described in the PR body (`finalize_host_subscribe`'s 2s best-effort fetch relied on relay-caching pre-placing the body). Do not merge the PUT commit until D's fetch is hardened and those proofs pass without relay-caching.

[AI-assisted - Claude]

sanity:
Superseded by the settled demand-driven hosting model (#4642).

This PR's core change removes the store-at-every-hop copy, but the current invariants KEEP every-hop hosting: those copies are now gauge-governed and evictable under subscriber-primary eviction, not removed. The durable-regardless-of-demand removal this PR targeted is already folded into 0.2.94 (subscriber-primary eviction, evict-subscribed + memory teardown, interest-gated renewal, and the min_ttl drop: #4720 / #4725 / #4734), which makes the every-hop copies demand-governed rather than durable. The GET-auto-subscribe half already landed independently.

The one remaining copy-regardless-of-demand, `relay_put_replicate_forward` (#4509), stays: it is load-bearing for UPDATE convergence given the still-live store-at-every-hop broadcast tree, so it retires WITH R3's two-layer freshness/propagation mechanism (after a convergence sim-check), not via this PR.

Closing as superseded. See `.claude/rules/hosting-invariants.md` (Resolved decisions) and #4642.

[AI-assisted - Claude]

PR #4666, opened 2026-07-02
feat: host-side scheduled wakeup primitive for delegates (#3972)

### LINKED ISSUE #3972: feat: scheduled wakeup primitive for delegates (ScheduleWakeup / WakeupFired)
## Problem

Delegates have no way to schedule future execution. The `OutboundDelegateMsg` enum has no `ScheduleWakeup`-equivalent variant, and `InboundDelegateMsg` has no `WakeupFired`. Delegate execution today is purely message-driven: there is no host-side mechanism to wake a delegate at a future time.

This forces dApps that need periodic background work (e.g. key rotation, TTL pruning, scheduled publication) to push that work into a UI/client sync loop. That breaks the Freenet model because the dApp stops working when the UI is not open.

## Driving use case

River's private-rooms feature needs weekly secret rotation that runs whether the room owner has the UI open or not. Tracked at freenet/river#228.

The current state of River private rooms (rotation in the UI sync loop) is the anti-pattern this primitive removes.

## API surface

In `freenet-stdlib/rust/src/delegate_interface.rs`:

```rust
pub enum OutboundDelegateMsg {
    // ... existing variants ...
    ScheduleWakeup {
        at: SystemTime,        // absolute, not duration
        tag: Vec<u8>,          // opaque to the host
    },
}

pub enum InboundDelegateMsg {
    // ... existing variants ...
    WakeupFired { tag: Vec<u8> },
}
```

### Cancel-by-tag semantics

Re-scheduling with the same `tag` replaces the prior entry for that `(DelegateKey, tag)` pair. A delegate that wants distinct wakeups uses distinct tags. This keeps the API small and matches the natural use case ("next weekly rotation for room X").

A separate `CancelWakeup` message could be added later if a use case demands it, but is not needed for the initial primitive.

## Host-side implementation

In `freenet-core/main/crates/core/src/wasm_runtime/delegate.rs` (or sibling):

- Per-node `BTreeMap<Instant, (DelegateKey, Vec<u8>)>` (forward index by fire time), plus `HashMap<(DelegateKey, Vec<u8>), Instant>` (reverse index for cancel-by-tag).
- Persisted to ReDb so wakeups survive node restart.
- A tokio task drives a `tokio::time::sleep_until(next_wakeup)` loop; on fire, build `InboundDelegateMsg::WakeupFired` and dispatch via `exec_inbound`.
- If a wakeup fires for a delegate that is no longer registered, drop it (do not auto-reinstall). A reinstalled delegate re-schedules from its own startup logic.

## Tests

- Schedule + fire at the expected time.
- Cancel-by-tag: re-schedule with same `(DelegateKey, tag)` replaces the prior entry.
- Persistence: wakeup scheduled, node restart, wakeup fires after restart.
- Drop on uninstalled delegate.
- Multiple wakeups for one delegate with distinct tags fire independently.
- Concurrent schedule from many delegates does not lose entries.

## Out of scope for this issue

- `CancelWakeup` message (deferred).
- Persistent-across-uninstall wakeups (deferred; not needed for River).
- Recurring/cron-style wakeups: the delegate re-schedules itself on fire.

[AI-assisted - Claude]

sanity:
> ⚠️ **Blocked on freenet-stdlib 0.8.3** (adds `ScheduleWakeup` / `WakeupFired`).
> This PR depends on that release being published to crates.io first
> (stdlib-first policy). CI will be red until then. Draft until 0.8.3 is out and
> `Cargo.lock` is refreshed (`cargo update -p freenet-stdlib --precise 0.8.3`).

## Problem

Delegates cannot schedule future execution, so always-on background work (River
private-room key rotation #228, TTL pruning, scheduled publish) has to live in a
UI sync loop and stops when the UI is closed. This implements the host side of
the scheduled-wakeup primitive whose wire variants land in freenet-stdlib 0.8.3.

## Approach

**`WakeupScheduler` (`crates/core/src/contract/wakeup.rs`)** — a per-node
schedule with:
- a time-ordered forward index (`BTreeMap<(SystemTime, seq)>`) so "what fires
  next" is O(1),
- a `(delegate, tag)` reverse index for **cancel-by-tag** (re-scheduling the
  same tag replaces the prior pending wakeup), and
- **ReDb persistence** (new `WAKEUPS_TABLE`) so pending wakeups survive restart;
  the table is created on first open of an upgraded DB.

**Params capture.** A fired wakeup must invoke the delegate with the same params
it was registered with, but the `DelegateKey` is a one-way hash of code+params so
the host can't recover them from the key. Each wakeup therefore stores the
delegate's params at schedule time and replays them on fire — correcting the
empty-params shortcut the contract-notification path takes.

**Inline firing (no separate task).** `OutboundDelegateMsg::ScheduleWakeup` is
intercepted in `handle_delegate_with_contract_requests` and persisted
(fire-and-forget, not re-driven). Wakeups fire **inline** as a new arm in the
`contract_handling` `select!` loop — the same single task that owns the executor
and runs delegates. That means no cross-task channel, no lossy dispatch, and no
`Notify`: because scheduling and firing are on the same task, a newly-scheduled
earlier wakeup is observed on the loop's next pass. Firing delivers
`InboundDelegateMsg::WakeupFired`; a delegate that is no longer registered yields
a missing-delegate error that is logged and dropped (the entry was already
removed by `take_due`), never reinstalled — the issue's "drop on uninstalled
delegate" rule.

This deviates from the issue's suggested standalone-tokio-task sketch on purpose:
firing on the delegate-owning task avoids the whole class of cross-task
channel/back-pressure bugs the project's channel-safety rules exist to prevent.

## Testing

- `WakeupScheduler` unit tests: cancel-by-tag replace, earliest-first fire
  ordering, bounded batch, ReDb persistence across reload (simulated restart),
  cancel-by-tag persisted, and volume/no-loss for many independent schedules.
- New `test-delegate-wakeup` WASM delegate + runtime tests: a delegate emits
  `ScheduleWakeup` (host sees it), a host-delivered `WakeupFired` reaches the
  delegate's `process()` and it reacts — exercising both new variants across the
  WASM bincode boundary — plus the drop-on-uninstalled path.

Closes #3972

[AI-assisted - Claude]

---

### Review

Full-tier multi-model review run pre-PR (wire format + delegate runtime +
concurrency + persistence): Codex (external) + 4 diverse Claude reviewers
(concurrency, persistence/wire-format, adversarial bug-hunt, big-picture). The
data-race and lost-wakeup questions were traced and confirmed safe (single-task
`&mut`, no cross-task mutation). Findings addressed in this PR:

- **Sender-attestation bypass** (Codex/bug-hunt): the `ScheduleWakeup` arm in
  `process_outbound` no longer drain-and-breaks; it falls through so each
  trailing `SendDelegateMessage` reaches its own attesting arm (also preserves
  `ContextUpdated` folding).
- **`decode_value` panic on a corrupt row** (persistence/bug-hunt): now validates
  nanos and uses `checked_add`, returning `None` so a torn row is skipped, not a
  startup crash. New overflow test.
- **Unbounded pending wakeups** (Codex/persistence/big-picture): per-delegate,
  global, and tag-length caps enforced at `schedule()`; new cap tests.
- LOW: wall-clock/`TimeSource` deviation now carries a load-bearing justification
  comment; at-most-once delivery documented on the stdlib variant; the shared
  "migration probes" warn generalised for the uninstalled-wakeup case.

Known follow-ups (filed separately): a loop-level integration test for the
firing glue (a deterministic one is blocked by the intentional wall-clock design;
tracked as an issue), and the pre-existing drain-bypass in the sibling terminal
arms (`ApplicationMessage`/`RequestUserInput`/… — not introduced here).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,7 +152,7 @@ muda = "0.19"
 zip = { version = "8", default-features = false, features = ["deflate", "time"] }
 
 # Freenet
-freenet-stdlib = "0.8.2"
+freenet-stdlib = "0.8.3"
 
 # Crypto (secrets-at-rest KEK + DEK derivation, see crates/core/src/config/kek.rs).
 # `sha2` is already declared above (workspace-wide); only `hkdf` and `keyring`
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -1378,6 +1378,7 @@ async fn process_open_request(
                                         "ContractNotification"
                                     }
                                     InboundDelegateMsg::DelegateMessage(_) => "DelegateMessage",
+                                    InboundDelegateMsg::WakeupFired { .. } => "WakeupFired",
                                     _ => "Unknown",
                                 })
                                 .collect();
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -14,6 +14,7 @@ pub(crate) mod governance;
 mod handler;
 pub mod storages;
 pub(crate) mod user_input;
+pub(crate) mod wakeup;
 
 pub(crate) use executor::{
     ContractExecutor, ExecutorTransactionStream, ExportBusy, MAX_CREATED_DELEGATES_PER_NODE,
@@ -53,6 +54,11 @@ const MAX_CONTRACT_REQUEST_ITERATIONS: usize = 100;
 /// Matches the same bounded-drain pattern as MAX_DRAIN_BATCH for contract events,
 /// preventing delegate notification channel growth under sustained contract load.
 const MAX_DELEGATE_DRAIN_BATCH: usize = 16;
+
+/// Max delegate wakeups (#3972) fired per event-loop iteration. Each fire is a
+/// full delegate execution, so bound it like the delegate-notification drain so
+/// a burst of simultaneously-due wakeups can't starve contract events.
+const MAX_WAKEUP_DRAIN_BATCH: usize = 16;
 
 /// Timeout for the off-loop related-contract fetch that backs a deferred
 /// PUT/UPDATE (#4391).
@@ -566,11 +572,14 @@ where
             }
             Err(err) => {
                 // Downgrade "not found" to warn — expected during legacy
-                // migration probes when old delegate WASM isn't on this node
+                // migration probes when old delegate WASM isn't on this node, and
+                // when a scheduled wakeup (#3972) fires for a delegate that has
+                // since been uninstalled (the wakeup is then simply dropped).
                 if err.is_missing_delegate() {
                     tracing::warn!(
                         delegate_key = %delegate_key,
-                        "Delegate not found in store (expected for migration probes)"
+                        "Delegate not found in store (expected for migration probes \
+                         or a fired wakeup on an uninstalled delegate)"
                     );
                 } else {
                     tracing::error!(
@@ -593,6 +602,10 @@ where
         let mut subscribe_requests: Vec<SubscribeContractRequest> = Vec::new();
         let mut delegate_messages: Vec<DelegateMessage> = Vec::new();
         let mut user_input_requests: Vec<UserInputRequest<'static>> = Vec::new();
+        // Wakeups the delegate asked us to schedule this iteration (#3972).
+        // Fire-and-forget: persisted below, NOT re-fed to the delegate, so a
+        // schedule alone does not keep the request loop spinning.
+        let mut schedule_wakeups: Vec<(std::time::SystemTime, Vec<u8>)> = Vec::new();
 
         for msg in values {
             match msg {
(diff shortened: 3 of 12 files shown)


sanity:
Depends on freenet/freenet-stdlib#82 (adds `ScheduleWakeup`/`WakeupFired`, bumps 0.8.2→0.8.3). Per the stdlib-first policy this stays **draft** until #82 merges and 0.8.3 publishes to crates.io; then `Cargo.lock` is refreshed and CI can go green.

[AI-assisted - Claude]

sanity:
Follow-up issues from review (deferred, not blocking this PR): #4667 (loop-level integration test for the firing glue) and #4668 (pre-existing sender-attestation drain-bypass in the sibling terminal arms).

[AI-assisted - Claude]

sanity:
Closing to keep the PR board clear — this has no path to merge right now and no live consumer.

- **Dependency stalled:** stdlib #82 (the `ScheduleWakeup`/`WakeupFired` wire messages) is still an open draft and CONFLICTING. Its planned 0.8.3 slot was taken by the memory-leak fix (#83, shipped as 0.8.3), so #82 now needs re-versioning to 0.8.4, rebase, wire-format review, merge, and publish before this PR can consume it.
- **Driving use case is gone:** river#228 (private-room key rotation) shipped in May, and the weekly-rotation trigger that needed scheduled-wakeup was removed in favor of event-driven rotation. No River consumer is waiting on this.

The work isn't lost: #3972 stays open as the tracking issue, the branch `feat/3972-delegate-wakeup` (Full-tier reviewed) preserves the implementation, and the most likely future consumer is Delta subscription-keeping for demand-driven hosting (delta#30 / #4642). Reopen or cherry-pick the branch when a concrete consumer materializes.

[AI-assisted - Claude]

PR #4670, opened 2026-07-03
feat: export terminal-consult (piece C) counters to router_snapshot telemetry

### LINKED ISSUE #4658: Export terminal-consult (piece C) counters to telemetry + dashboard
Piece C (PR #4646, terminal advertisement consult, shipped in 0.2.90) records its counters (consult attempts, hits, resolved-found, still-not-found) only to the internal network_status singleton and the sim-only GlobalTestMetrics, plus a debug-level TERMINAL_CONSULT log line. They are NOT exported to the telemetry jsonl stream (router_snapshot) or the dashboard/homepage, so operators cannot see C's effectiveness in production.

This undercuts C's intended value as a probe into dead-end composition: off-path-advertised-host dead-ends (which C fixes) vs no-host-near-key (which needs piece D). Caught by the 0.2.90 pre-release smoke test and confirmed unobservable in the 0.2.90 canary.

Follow-up: surface C's per-node consult counters on the router_snapshot telemetry event and the dashboard, the way A3's eviction gauges and resource_utilization already are.

Lesson for the horizontal-instrumentation rule: instrumentation must reach the telemetry stream + dashboard, not just an internal singleton.

Related: #4646, #4642.

[AI-assisted - Claude]

sanity:
## Problem

Piece C (terminal advertisement consult, PR #4646, shipped in 0.2.90) records its counters — consult **attempts**, **hits**, **resolved-found**, **still-not-found** — only to the internal `network_status` singleton and the sim-only `GlobalTestMetrics`, plus a debug-level `TERMINAL_CONSULT` log line. They never reach the `router_snapshot` telemetry event or the collector, so operators cannot see C's effectiveness on the production 374-peer network (confirmed unobservable in the 0.2.90 canary).

This is the highest-value item in the 0.2.91 **instrument-and-de-risk** release: it makes the live network report the **real findability numbers**, so the piece-E / 0.2.92 decision is data-driven instead of sim-based. The four counters decompose dead-ends into the two cases the redesign cares about:

- **off-path-advertised-host dead-ends** that C closes: `attempts` → `hits` (found an advertised host off the routing path) → `resolved_found` (that forward resolved to Found/Subscribed)
- the **no-host-near-key residual** `still_not_found`, which needs piece D

## Approach

The four counters already exist and are already incremented correctly at every consult site (GET/SUBSCRIBE relay termini). This PR only adds the **missing export hop**, following the exact pattern the A2/A3 hosting gauges already use to reach central telemetry:

1. `network_status::terminal_consult_counts()` — a read-only getter returning the four per-node counters from the singleton.
2. Four `Option<u64>` fields on `RouterSnapshotInfo`.
3. `Ring::emit_router_snapshot_telemetry` polls the getter on the snapshot cadence and copies the values onto the snapshot (and the local `router_snapshot` `info!` log).
4. Four `obj.insert(...)` mirrors in `event_kind_to_json`'s `RouterSnapshot` arm (the hand-mirrored OTLP body), pinned by a new guard test.

**Purely additive instrumentation — no routing/consult/hosting behavior change.** Safe to ship independent of the piece-E bundle. `RouterSnapshotInfo` is a local-only telemetry struct (AOF event log + OTLP body), never a peer wire message, so there is no protocol/wire-format concern; adding `Option<u64>` fields matches the ~8 prior gauge additions to this exact struct.

**Version segmentation** is already provided by the existing pipeline: every OTLP batch carries the sender's `service.version` resource attribute (pinned by `test_otlp_resource_includes_service_version`), so the collector can group these counters by release (0.2.91 baseline vs later) with no new plumbing.

### Related findability signals — already exported, no change needed

GET/SUBSCRIBE success and NotFound are **already** in the collector stream as discrete per-request events (`get_success`, `get_not_found`, `subscribe_success`, `subscribe_not_found`), so the overall dead-end rate and success rates are already derivable per-release. The only missing piece of the findability baseline was the **consult** decomposition, which this PR adds. (The `network_status` op-stats aggregates are exported to the local homepage only, not the collector — left as-is to keep this change one logical thing.)

## Testing

- `terminal_consult_counts_reflect_recorded_events` (network_status): `init()` → `record_*` → getter round-trip with distinct per-field counts, so a dropped increment or transposed tuple fails here instead of silently emitting zeros in production.
- `router_snapshot_json_includes_terminal_consult_counters` (telemetry): pins the four fields through the hand-mirrored OTLP body, matching the existing `router_snapshot_json_includes_*` guards (the #4009/#4010 manually-mirrored-telemetry footgun).
- `cargo fmt` + `cargo clippy -- -D warnings` (default and `trace-ot`) clean; the existing `router_snapshot_json_includes_*` pins still pass.

Closes #4658

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -524,6 +524,22 @@ pub fn record_terminal_consult_still_not_found() {
     }
 }
 
+/// Read the current terminal advertisement-consult counters (hosting redesign
+/// piece C, #4646) for export to the `router_snapshot` telemetry event (#4658).
+///
+/// Returns `(attempts, hits, resolved_found, still_not_found)` — the four
+/// per-node monotonic aggregates recorded by the consult sites — or `None`
+/// before the `NETWORK_STATUS` singleton is initialized. `Ring` polls this on
+/// the snapshot cadence and copies the values onto `RouterSnapshotInfo` so the
+/// production findability baseline (dead-end decomposition) is legible in
+/// central telemetry, not just the internal singleton / sim-only metrics.
+pub fn terminal_consult_counts() -> Option<(u64, u64, u64, u64)> {
+    let status = NETWORK_STATUS.get()?;
+    let s = status.read().ok()?;
+    let c = &s.terminal_consult_stats;
+    Some((c.attempts, c.hits, c.resolved_found, c.still_not_found))
+}
+
 /// Record a NAT traversal attempt.
 pub fn record_nat_attempt(success: bool) {
     if let Some(status) = NETWORK_STATUS.get() {
@@ -1181,6 +1197,45 @@ mod tests {
     fn test_html_escape() {
         assert_eq!(html_escape("<script>"), "&lt;script&gt;");
         assert_eq!(html_escape("a&b"), "a&amp;b");
+    }
+
+    /// End-to-end for #4658: the terminal-consult record sites must feed the
+    /// `terminal_consult_counts()` getter that `Ring` polls for the
+    /// `router_snapshot` telemetry export. `init()` resets the counters, each
+    /// `record_*` bumps its own field, and the getter returns them in
+    /// `(attempts, hits, resolved_found, still_not_found)` order — so a wiring
+    /// mistake (wrong field, dropped increment, transposed getter tuple) fails
+    /// here rather than silently emitting zeros in production.
+    #[test]
+    fn terminal_consult_counts_reflect_recorded_events() {
+        let _lock = TEST_MUTEX.lock().unwrap();
+        // Fresh singleton state for a deterministic baseline (init overwrites
+        // the process-global tracker in place, zeroing the consult counters).
+        init(31337, HashSet::new(), "test".to_string());
+
+        assert_eq!(
+            terminal_consult_counts(),
+            Some((0, 0, 0, 0)),
+            "counters start at zero after init"
+        );
+
+        // Distinct counts per field so a transposition can't pass.
+        record_terminal_consult_attempt();
+        record_terminal_consult_attempt();
+        record_terminal_consult_attempt();
+        record_terminal_consult_hit();
+        record_terminal_consult_hit();
+        record_terminal_consult_resolved_found();
+        record_terminal_consult_still_not_found();
+        record_terminal_consult_still_not_found();
+        record_terminal_consult_still_not_found();
+        record_terminal_consult_still_not_found();
+
+        assert_eq!(
+            terminal_consult_counts(),
+            Some((3, 2, 1, 4)),
+            "getter returns (attempts, hits, resolved_found, still_not_found)"
+        );
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


sanity:
## Review — LIGHT tier (additive telemetry)

**Verdict: safe to merge.** Purely additive instrumentation — no routing / consult / hosting behavior change. Codex external pass + one adversarial read, both clean. No blocking or non-blocking findings.

### External pass
`codex review --base main`: *"The changes add telemetry plumbing for terminal consult counters consistently across collection, snapshot population, and OTLP JSON export, with targeted regression tests. I did not identify any discrete correctness issues introduced by the patch."*

### Adversarial read (each task claim verified against the diff)

1. **Genuinely additive.** The diff only adds a read-only getter, 4 `Option<u64>` fields, 4 populate lines in `Ring`, 4 OTLP `obj.insert` mirrors, one log line, and 2 tests. The `record_terminal_consult_*` sites are pre-existing and untouched — no consult/routing/hosting logic changes.

2. **Counters read correctly, no transposition.** `TerminalConsultStats` field order `attempts, hits, resolved_found, still_not_found` (`network_status.rs:184-190`) matches end-to-end: each `record_*` fn bumps its own field (`:490-525`); the getter returns `(c.attempts, c.hits, c.resolved_found, c.still_not_found)` (`:540`); `ring.rs:1499-1505` destructures in that order and assigns each to the matching snapshot field; the OTLP mirror inserts each `snapshot.terminal_consult_*` under the matching key (`telemetry.rs:2199-2214`). Both tests use distinct values (getter `3/2/1/4`; OTLP primes `311/313/317/331`), so a transposition fails.

3. **`RouterSnapshotInfo` is local-only.** Referenced only within `EventKind::RouterSnapshot` (tracing/telemetry, state_verifier, test_utils, register) — not in `message.rs`, `operations/`, or the p2p path. It is never a peer wire message, so the four `Option<u64>` additions carry no protocol/bincode-wire impact (matches the ~8 prior gauge additions to this struct).

4. **Version segmentation works.** The `service.version` OTLP resource attribute exists (`telemetry.rs:878`) and is pinned by `test_otlp_resource_includes_service_version` (`:2915`), so the collector can group these counters by release with no new plumbing.

5. **Tests pin what they claim.** `terminal_consult_counts_reflect_recorded_events` exercises `init → record_* → getter` with distinct per-field counts. `router_snapshot_json_includes_terminal_consult_counters` builds via `Arbitrary`, sets distinct values, and asserts all four top-level JSON keys — `obj` is `body.as_object_mut()` (`telemetry.rs:2059`), so `json[key]` resolves at top level, identical to the 7 existing `router_snapshot_json_includes_*` pins.

### Non-blocking notes (no action needed)
- The getter's `.read().ok()?` silently yields `None` on a poisoned lock, dropping the counters for that one snapshot cycle. Consistent with every other `network_status` accessor and harmless for best-effort telemetry.
- Counters are monotonic lifetime totals reset only by `init()` (process start); the collector differences them across the cadence, as the PR states.

CI must be green on HEAD before merge (standard gate); nothing in the change itself blocks.

[AI-assisted - Claude]

PR #3143, opened 2026-02-20
feat: send explicit Unsubscribe messages upstream for faster subscription cleanup

netsirius:
## Problem

When a peer no longer needs updates for a contract (no local clients, no downstream subscribers), it simply stops renewing and the upstream lease expires after 8 minutes. During that window the upstream peer continues sending unnecessary UPDATE broadcasts, wasting bandwidth, CPU, and memory.

## Solution

Send an explicit `SubscribeMsg::Unsubscribe` message upstream when a peer determines it no longer needs updates. This allows the upstream peer to immediately remove us from its subscriber tracking and stop forwarding updates.

### Key changes

**New message type** (`operations/subscribe.rs`):
- Added `SubscribeMsg::Unsubscribe { id, instance_id }` variant with full routing support (`InnerMessage`, `Display`, `load_or_init`, `process_message`).

**Downstream subscriber tracking** (`ring/hosting.rs`):
- Added `downstream_subscribers: DashMap<ContractKey, HashMap<PeerKey, Instant>>` to `HostingManager` to track which peers are subscribed to us per contract.
- New methods: `add_downstream_subscriber`, `renew_downstream_subscriber`, `remove_downstream_subscriber`, `has_downstream_subscribers`, `expire_stale_downstream_subscribers`, `should_unsubscribe_upstream`.

**Trigger points**:
- *Client disconnect* (`client_events/mod.rs`): when the last local client for a contract disconnects and no downstream subscribers remain, sends Unsubscribe upstream.
- *Downstream lease expiry* (`ring/mod.rs`): periodic sweep detects expired downstream subscribers; if no remaining interest, sends Unsubscribe upstream.
- *Chain propagation* (`operations/subscribe.rs`): when receiving an Unsubscribe from a downstream peer, removes that peer's tracking and propagates upstream if no other subscribers remain.

**Routing mechanism** (`node/op_state_manager.rs`):
- `send_unsubscribe_upstream()` finds the upstream peer via `InterestManager` (`is_upstream: true`), creates a temporary `SubscribeOp` for hop-by-hop routing, sends the message fire-and-forget, and cleans up local state (active subscription + interest tracking).

**Bug fix — upstream peer registration** (`operations/subscribe.rs`):
[...]

PR #4671, opened 2026-07-03
fix: preserve upstream interest across ChangeInterests gossip

sanity:
## Problem

The `ChangeInterests` interest-sync handler (`node.rs::handle_interest_sync_message`) re-registered every peer that re-advertises interest with a bare `register_peer_interest(.., is_upstream = false)`. `register_peer_interest` overwrites the **whole** `PeerInterest` entry, so when the peer is already our **upstream** host (`is_upstream = true`, set at subscribe time in `subscribe.rs` / `operations.rs`), this:

- flips `is_upstream` **true -> false**, and
- wipes the cached delta-sync summary to `None`.

**User impact — event-driven chain collapse is defeated.** `OpManager::send_unsubscribe_upstream` finds the upstream via `get_interested_peers().find(|i| i.is_upstream)`. Once the flag is clobbered it can no longer locate the upstream, so the `Unsubscribe` is never sent up the chain on client/downstream departure. The subscription then only lapses on lease expiry (~6-min stale window) instead of collapsing promptly when the last local interest goes.

**The clobber is event-driven, not periodic.** The `ChangeInterests` "added" path fires only when a peer's interest in a contract transitions 0→1 (broadcast via `broadcast_change_interests` as a `BroadcastChangeInterests` event); the ~5-min interest-sync **heartbeat** sends `InterestMessage::Interests` (the full-replace arm, already guarded — `ring.rs`), **not** `ChangeInterests`. Hitting a real upstream edge therefore requires our upstream host's own interest to **lapse and revive**, so it re-emits an "added" for a contract we already hold it as upstream for. That is **uncommon on current `main`** — hosts renew leases unconditionally, so an upstream's interest does not lapse — but it becomes **load-bearing under piece-D interest-gated renewal** (freenet/freenet-core#4642), where interest lapses when demand goes and revives when it returns. Guarding now keeps the flag correct before D makes the edge routine. (The earlier description of this as "deterministic and periodic via the heartbeat" was inaccurate and has been corrected here and in the code/test docs.)

## Solution

The sibling `Interests` full-replace arm already guards re-registration of an existing entry:

```rust
if get_peer_interest(&contract, pk).is_some() {
    refresh_peer_interest(&contract, pk);   // preserves is_upstream + summary
} else {
    register_peer_interest(&contract, pk.clone(), None, false);
}
```

This mirrors that guard in the `ChangeInterests` arm. `refresh_peer_interest` preserves both `is_upstream` and the cached summary; a brand-new peer is still registered (and still flushes the #4359 deferred broadcast). No wire-format change; no new `register_peer_interest`/`flush_pending_broadcast_on_interest` call, so the existing #4359 pairing pin and the #4473 summarize-gate pin stay green.

### Audit of the other `register_peer_interest(.., false)` sites
- `broadcast_queue.rs` occurrences are **test-only** — no clobber risk.
- `subscribe.rs:643` (`register_downstream_subscriber`) and `get/op_ctx_task.rs:2650` register a **genuine downstream party** (a peer subscribing to / GETting from us) as `is_upstream = false`, which is the intended semantic. They only clobber in a **cyclic** topology (a peer that is simultaneously our upstream *and* our downstream for the same contract), which does not form under normal chain routing and self-corrects on the next subscribe. Left unchanged to keep this fix focused on the deterministic collapse-defeat — filed as **follow-up freenet/freenet-core#4672** (the clean class-wide fix is a primitive-level `is_upstream = existing || new` merge). Not required for this PR.

### Scope
This is the pre-existing `is_upstream` formation-flag collapse path on `main` (present since #3143). It is independent of the computed-upstream redesign (piece D, draft #4661) — piece D does not touch this handler, so the fix stands alone and improves current released behaviour.

## Testing
- `change_interests_arm_guards_register_with_refresh_pin` (node.rs) — the handler-**wiring** regression signal. A **structural** source-scrape (comments stripped first) that the `ChangeInterests` arm has the guard SHAPE `get_peer_interest(..).is_some() -> refresh_peer_interest(..)` with `register_peer_interest(..)` only as the else-branch fallback (ordered offsets + exactly-one register), not mere token presence. Verified it **FAILS on the pre-fix bare register** and passes on the guarded shape.
- `upstream_interest_survives_refresh_but_bare_register_clobbers_it` (ring/interest.rs) — a **characterization** of the InterestManager primitives the guard relies on (refresh preserves `is_upstream` + summary and the `send_unsubscribe_upstream` lookup still resolves; a bare `register(false)` clobbers both). It exercises the primitives directly, so it passes on the pre-fix handler too — the docstring now says so and points to the wiring pin above.

Both new tests pass; `cargo fmt`, `cargo clippy --locked -- -D warnings`, and the affected existing pins (`interest_sync_periodic_arms_summarize_only_hosted_or_in_use_pin`, `pending_broadcast_flush_wired_at_all_interest_sites`, `test_is_upstream_flag_registration`) are green.

> Touches the subscription-collapse path — Full multi-model review before merge; **no auto-merge**.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2707,13 +2707,42 @@ async fn handle_interest_sync_message(
                             continue;
                         }
 
-                        // Register their interest
-                        let is_new = op_manager.interest_manager.register_peer_interest(
-                            &contract,
-                            pk.clone(),
-                            None,
-                            false,
-                        );
+                        // Register their interest — but preserve an existing
+                        // entry's `is_upstream` flag (and cached summary). A
+                        // ChangeInterests "added" gossip from a peer that is
+                        // ALREADY our upstream host (is_upstream = true, set when
+                        // we subscribed through it — subscribe.rs / operations.rs)
+                        // must NOT be downgraded to a plain downstream interest.
+                        // A bare re-registration with is_upstream=false overwrites
+                        // the whole PeerInterest, flipping is_upstream true -> false
+                        // and wiping the delta-sync summary to None. The is_upstream
+                        // clobber defeats event-driven collapse:
+                        // `send_unsubscribe_upstream` locates the upstream via
+                        // `is_upstream`, so after the first gossip round the
+                        // Unsubscribe is never sent upstream and the chain only
+                        // lapses on lease expiry (~6 min stale window). An upstream
+                        // peer hosts the contract, so it re-advertises interest on
+                        // every ~5-min interest-sync heartbeat, making this clobber
+                        // deterministic and periodic. Mirror the refresh-guard the
+                        // `Interests` full-replace arm above uses. (Guarded wiring
+                        // pinned by change_interests_arm_guards_register_with_refresh_pin.)
+                        let is_new = if op_manager
+                            .interest_manager
+                            .get_peer_interest(&contract, pk)
+                            .is_some()
+                        {
+                            op_manager
+                                .interest_manager
+                                .refresh_peer_interest(&contract, pk);
+                            false
+                        } else {
+                            op_manager.interest_manager.register_peer_interest(
+                                &contract,
+                                pk.clone(),
+                                None,
+                                false,
+                            )
+                        };
                         if is_new {
                             // #4359 (MUST-FIX 1): a ChangeInterests addition
                             // makes this peer a viable broadcast target. Flush
@@ -3498,6 +3527,67 @@ mod tests {
             gated_calls >= 3,
             "expected the 3 periodic interest-sync arms to call \
              summary_if_hosted_or_in_use, found {gated_calls}"
+        );
+    }
+
+    /// Regression pin for the D2 `is_upstream` clobber in the `ChangeInterests`
+    /// interest-sync arm.
+    ///
+    /// A peer that is already our UPSTREAM host (`is_upstream = true`, set when
+    /// we subscribed through it) re-advertises interest on every ~5-min
+    /// interest-sync heartbeat (it hosts the contract, so it is interested).
+    /// A bare `register_peer_interest(.., is_upstream = false)` overwrites the
+    /// whole `PeerInterest`, flipping `is_upstream` true -> false and wiping the
+    /// cached delta-sync summary. The is_upstream clobber defeats event-driven
+    /// chain collapse: `send_unsubscribe_upstream` finds the upstream via
+    /// `is_upstream`, so after the first gossip round the Unsubscribe is never
+    /// sent upstream and the chain only lapses on lease expiry (~6-min stale
+    /// window). The arm therefore MUST guard re-registration of an existing
+    /// entry with `get_peer_interest().is_some() -> refresh_peer_interest()`,
+    /// exactly like the `Interests` full-replace arm.
+    ///
+    /// The primitive behaviour (refresh preserves, bare register clobbers) is
+    /// pinned in `ring/interest.rs` by
+    /// `upstream_interest_survives_refresh_but_bare_register_clobbers_it`.
+    /// Driving the async handler end-to-end needs a full OpManager + connection
+    /// fixture, so — following the codebase convention for interest-sync
+    /// wiring (see `op_state_manager.rs` #4359 pins) — the HANDLER wiring is
+    /// pinned at the source level here.
+    #[test]
+    fn change_interests_arm_guards_register_with_refresh_pin() {
+        let src = include_str!("node.rs");
+
+        let handler_start = src
+            .find("async fn handle_interest_sync_message(")
+            .expect("handle_interest_sync_message not found");
+        let handler_src = &src[handler_start..];
+
+        // Bound the slice to the ChangeInterests arm: from its match pattern up
+        // to the next arm (ResyncRequest), so the guard assertions below can't
+        // false-pass on a neighbouring arm's guard.
+        let arm_start = handler_src
+            .find("InterestMessage::ChangeInterests { added, removed } =>")
+            .expect("ChangeInterests arm not found");
+        let arm_len = handler_src[arm_start..]
+            .find("InterestMessage::ResyncRequest")
+            .expect("ResyncRequest arm (ChangeInterests terminator) not found");
+        let arm = &handler_src[arm_start..arm_start + arm_len];
+
+        // Presence of BOTH guard calls is the regression signal: the pre-fix
+        // (unguarded) arm re-registered every peer with a bare
+        // `is_upstream = false` and contained NEITHER call. Reverting to that
+        // clobber removes both, tripping this pin. (Presence-based rather than
+        // positional so a future comment mentioning a call token can't
+        // spuriously fail it — the sibling `Interests` arm uses the same guard.)
+        assert!(
+            arm.contains("get_peer_interest("),
+            "ChangeInterests arm MUST guard re-registration with get_peer_interest() \
+             so an existing upstream peer is not clobbered to is_upstream=false (D2)"
+        );
+        assert!(
+            arm.contains("refresh_peer_interest("),
+            "ChangeInterests arm MUST refresh (not re-register) an existing entry to \
+             preserve is_upstream + the cached summary (D2)"
         );
     }
 
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2718,14 +2718,22 @@ async fn handle_interest_sync_message(
                         // and wiping the delta-sync summary to None. The is_upstream
                         // clobber defeats event-driven collapse:
                         // `send_unsubscribe_upstream` locates the upstream via
-                        // `is_upstream`, so after the first gossip round the
-                        // Unsubscribe is never sent upstream and the chain only
-                        // lapses on lease expiry (~6 min stale window). An upstream
-                        // peer hosts the contract, so it re-advertises interest on
-                        // every ~5-min interest-sync heartbeat, making this clobber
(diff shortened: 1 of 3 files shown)


sanity:
## Full-tier review — D2 `ChangeInterests` `is_upstream` clobber fix

Touches the event-driven collapse path (`send_unsubscribe_upstream` / interest tracking), so this was run as a Full-tier review: an external **Codex (`gpt-5.5`)** pass plus four adversarial Claude lenses (correctness, other clobber sites, broadcast/heartbeat interaction, test adequacy). Reviewed the checked-out branch at `308f0cd3`.

**Bottom line: the code change is correct and safe to merge.** The runtime behaviour is exactly what the PR intends, and no reviewer found a correctness defect. There are three non-blocking findings — the most important is a **factual inaccuracy in the PR/test description** (the trigger is not the periodic heartbeat), which is worth correcting before merge because it's load-bearing for how this is prioritised against piece D.

### What was verified (positive)

- **`refresh` preserves both fields.** `refresh_peer_interest` (`interest.rs:486-493`) only touches `last_refreshed`; `is_upstream` and `summary` are untouched. The guard therefore does what the PR claims. ✅
- **No broadcast fan-out change.** `is_upstream` is read in exactly **one** production site — `op_state_manager.rs:783` (`send_unsubscribe_upstream`). Broadcast target selection does not consult it, so restoring `is_upstream = true` cannot alter update fan-out anywhere. This is the key reason the fix is low-risk. ✅
- **Summary preservation is consistent and strictly better** than the clobber — it matches the sibling `Interests` arm's rationale (`node.rs:2447-2450`), and a later `Summaries` message refreshes it via `update_peer_summary`. ✅
- **New-peer path unchanged** — still `register_peer_interest(.., false)` + `#4359` deferred-broadcast flush. The `#4359` pairing pin stays green (test-stripped prod counts: `register_peer_interest(`=2, `flush_pending_broadcast_on_interest(`=2). ✅
- **CI is green on `308f0cd3`** (Unit & Integration, Simulation, Clippy, Fmt, Rule Lint all pass). ✅
- **Codex verdict:** *"The production change appears to preserve existing upstream interest state as intended. The only issue found is with the robustness of the new source-scrape regression test, not with runtime behavior."*

### Findings

**[M1 — Medium] The PR/test description mis-states the trigger as "deterministic and periodic via the ~5-min heartbeat."**
The periodic interest-sync heartbeat (`ring.rs::interest_heartbeat`, `ring.rs:1250`) sends `InterestMessage::Interests { hashes }` — the **full-replace arm, which was already guarded** (`node.rs:2451-2458`) — **not** `ChangeInterests`. The `ChangeInterests` path this PR fixes is **event-driven**: it is only emitted by `broadcast_change_interests`, and every call site is gated on a local interest **0→1 (`became_interested`) or removal transition** (`subscribe.rs:599-603`, `operations.rs:488-490`, `get/op_ctx_task.rs:1325-1330`, `put/op_ctx_task.rs:1905`). For it to clobber a *genuine* upstream edge, the upstream's interest in the contract must **lapse and revive** while the A→B upstream edge persists — uncommon on current `main` under unconditional lease renewal.
So the bug is real, but it is not periodic and its impact on *current released* behaviour is smaller than the description implies. (It becomes materially more important under piece-D interest-gated renewal, where lapse/revive is the normal case — a good argument for landing this now, but a different argument than the one in the PR.) **The fix itself is correct;** please correct the PR body and the two test docstrings (`node.rs:3538`, `interest.rs:2174-2191`), which repeat the "every ~5-min interest-sync heartbeat" mechanism and will mislead the next reader. Not a code defect.

**[M2 — Medium] Regression protection is thinner than it appears.**
- The behavioural test `upstream_interest_survives_refresh_but_bare_register_clobbers_it` (`interest.rs:2193`) exercises the **primitives directly** (`manager.refresh_peer_interest` / `manager.register_peer_interest`), which this PR does **not** change. It therefore **passes on the pre-fix buggy code** — it's a characterization test of the primitive contract, not a regression test for the handler bug.
- The only test that fails on pre-fix code is the source-scrape pin `change_interests_arm_guards_register_with_refresh_pin` (`node.rs:3557`). It asserts **token presence** (`arm.contains("get_peer_interest(")`, `node.rs:3583/3588`), so a **logic-inverted** guard (`.is_none()` instead of `.is_some()`, or refresh/register swapped) would keep both tokens and pass while reintroducing the clobber.
- No existing fixture drives `handle_interest_sync_message` end-to-end (the only non-dispatch references are the source-scrape pins), so a true behavioural handler test would be new work — the source-scrape approach follows codebase convention, which is fair. Recommend at minimum strengthening the pin to assert the guard **structure** (if-branch refreshes, else-branch registers) rather than mere token presence.

**[L1 / P3 — Low] Source-scrape pin can false-pass on comment drift** (raised by Codex, `node.rs:3583-3588`).
`arm.contains("get_peer_interest(")` also matches those tokens if they appear in the arm's *comments*. They do **not** today (verified — the only matches in the sliced arm are the real `.get_peer_interest(` / `.refresh_peer_interest(` calls), so the pin is valid now, but a future comment mentioning the token would let a reintroduced clobber slip through. Strip comment/string regions before matching (folds naturally into the M2 hardening).

**[L2 — Low] The clobber class is not fully closed; consider a primitive-level fix.**
Two other `register_peer_interest(.., false)` sites — `subscribe.rs:643` (`register_downstream_subscriber`) and `get/op_ctx_task.rs:2650` — can still clobber `is_upstream` true→false (and wipe the summary) for a peer that is simultaneously our upstream *and* a downstream/GET-requester (a cyclic topology). The PR's "rare and self-correcting on current `main`" argument is accurate, so leaving them is defensible for a focused fix. But note the per-call-site guard pattern is exactly what failed here — the `ChangeInterests` arm was originally added *without* the guard the `Interests` arm had. A more robust fix preserves `is_upstream` at the primitive: for an existing entry, `register_peer_interest` would set `is_upstream = existing || new` and `summary = new.or(existing)`, closing all sites at once. Design suggestion / follow-up, not required here.

### Recommendation

Safe to merge on correctness grounds. Suggest addressing **M1** (cheap, and the inaccurate mechanism is baked into the permanent test docstrings) and **M2/L1** (strengthen the one real regression signal) before merge; **L2** as a tracked follow-up. None of the findings block correctness.

[AI-assisted - Claude]

sanity:
Addressed the review findings in a9f2b79 (docs/tests only — the production guard is byte-identical to 308f0cd, so this does not restart the production-code review).

**M1 — inaccurate mechanism corrected.** The clobber path is EVENT-DRIVEN, not "deterministic and periodic via the heartbeat": `ChangeInterests` is emitted only on a 0→1 interest transition (`broadcast_change_interests`), whereas the ~5-min interest-sync heartbeat sends `InterestMessage::Interests` (the already-guarded full-replace arm), not `ChangeInterests`. Hitting a real upstream edge needs the upstream's own interest to lapse-and-revive — uncommon on current `main` (leases renew unconditionally), load-bearing under piece-D interest-gated renewal (#4642). Corrected the PR description, the production comment in the `ChangeInterests` arm, and both test docstrings (node.rs + interest.rs).

**M2 — real regression signal.** The interest.rs test drives the InterestManager primitives directly, so it PASSES on the pre-fix handler — its docstring now labels it a primitive characterization and points to the wiring pin. Strengthened `change_interests_arm_guards_register_with_refresh_pin` (node.rs) from mere token presence to a **structural guard-shape** assertion: `get_peer_interest(..).is_some() -> refresh_peer_interest(..)` with `register_peer_interest(..)` only as the else-branch fallback (ordered byte offsets + exactly-one register). Verified locally: it **FAILS on the pre-fix bare `register(false)`** (`ChangeInterests arm MUST look up the existing entry with get_peer_interest()...`) and passes on the guarded shape.

**L1 — drift-proofed.** The pin strips `//` line comments before matching, so a guard token that later appears only in a comment can't let a reverted bare-register arm false-pass.

**L2 — filed #4672** for the two other `register_peer_interest(.., false)` sites (`subscribe.rs:643`, `get/op_ctx_task.rs:2650`) that can clobber `is_upstream` in a cyclic topology; the clean class-wide fix is a primitive-level `is_upstream = existing || new` merge. Not required for this PR.

`cargo fmt --check` and `cargo clippy --locked -p freenet -- -D warnings` are clean; both affected tests pass. No auto-merge set.

[AI-assisted - Claude]

sanity:
Follow-up: two more commits, CI green on HEAD 9fd1cded.

- `a9f2b79` — the M1/M2/L1 doc + pin work above.
- `9fd1cded` — fixes a pre-existing latent bug in the `#4359` flush pin (`pending_broadcast_flush_wired_at_all_interest_sites`) that the strengthened D2 pin exposed. Its `strip_cfg_test_regions` helper found `#[cfg(test)]` boundaries by raw brace-balance, so a stray `}` inside a string/comment (a neighbouring test's doc comment `` `}` `` + a `.find("\n}\n")` call) truncated the mod-tests strip early. The D2 pin legitimately naming `register_peer_interest(` in its assertion strings then leaked into the "production" count (10 vs 3 flush). Fixed by making the stripper skip `//` / `/* */` / `"..."` during boundary detection (byte-only, so scanning a multi-byte char in a doc comment can't panic; output stays verbatim so the "Source 1 / proximity" check still matches). Verified tree-wide: found-set still equals `known_wired` and every wired file keeps flush >= register.

The first push also had one flaky `test_multiple_clients_subscription` failure (timing-sensitive subscription-propagation integration test, unrelated to this docs/tests-only change — it passed on this run with no related edit). All checks green now.

[AI-assisted - Claude]

PR #4677, opened 2026-07-03
refactor: retire dormant MAD governance from dashboard, surface demand-driven eviction

### LINKED ISSUE #4676: Local dashboard surfaces dormant MAD governance, not the shipped demand-driven eviction — retire it and surface piece A + freshness
## Problem

The local peer dashboard (`GET /`, `crates/core/src/server/home_page/`) still presents the **old MAD-based `GovernanceManager`** as its retention/eviction story, which is misleading on every real node:

- **The "Contract Governance" card** (`cards.rs::build_governance_card`) renders the MAD cost/benefit outlier detector. That detector is **default-Off** (`contract/governance.rs`, `GovernanceMode::Off`) and is explicitly being **replaced** (#4296 — it was miscalibrated, flagging a large fraction of a node's contracts, including the official River room, as `WouldEvict`). On real nodes the card just says "Governance is off." — a dormant, superseded mechanism taking prime dashboard real estate.
- **The "Eviction threshold" tile** (a MAD log-ratio) is actively misleading: it has **zero relation to what actually evicts contracts today**.
- **Piece A — the eviction that actually governs retention now** (demand-driven Greedy-Dual `keep_score` + capability-relative RAM budget, shipped 0.2.90; `ring/hosting/demand.rs` + `ring/hosting/cache.rs`) is **surfaced nowhere** on the dashboard.
- **Freshness/demand is not shown per contract.** The Subscribed Contracts table has a "Gov" column that is always "ok" (fed from the dormant governance `state_by_id`). The real freshness signal is `is_receiving_updates` (NOT `is_hosting` — see the PR #3699 note), plus `contract_in_use` for demand.
- A **stale comment** at `ring.rs` (governance reaper spawn) claims "Mode defaults to DryRun…" when it actually defaults to `Off`.

## Proposed cleanup (Ian has authorized this)

1. **Gate the Contract Governance card** so it renders only when governance `mode != Off` — it vanishes on default nodes (removing the dormant/misleading content) but stays for anyone who explicitly enables governance.
2. **Fix the stale `ring.rs` comment** (defaults to Off, not DryRun).
3. **Replace the always-"ok" Gov column** in Subscribed Contracts with a real freshness/demand indicator (`is_receiving_updates` + `contract_in_use`).
4. **Surface piece A's live eviction**: RAM budget used vs headroom, per-contract `keep_score`/predicted-demand, and "next to evict," plumbed from `ring/hosting/{demand,cache}.rs` into the dashboard snapshot.
5. **Per-contract freshness** (`is_receiving_updates`), not just `last_updated_secs`.
6. Keep the Ban List card; de-emphasize the dormant "auto (governance)" ban reason (only the operator path is live).

## References
- #4642 — Epic: demand-driven hosting / placement / eviction redesign (piece A shipped 0.2.90)
- #4296 — tracking: contract + peer resource-usage governance (replacing the MAD detector)
- `.claude/rules/hosting-invariants.md` — the demand-driven model this dashboard should reflect

[AI-assisted - Claude]

sanity:
## Problem

The local peer dashboard (`GET /`) still presented the **old MAD-based `GovernanceManager`** as its retention story, which is misleading on every real node:

- The **"Contract Governance" card** rendered the MAD cost/benefit outlier detector, which is **default-Off** (`contract/governance.rs`) and is being **replaced** by demand-driven eviction (#4296 — it was miscalibrated, flagging a large fraction of a node's contracts, incl. the official River room, as `WouldEvict`). On real nodes it only ever said "Governance is off" — a dormant, superseded mechanism taking prime dashboard space.
- The **"Eviction threshold" tile** (a MAD log-ratio) was actively misleading: zero relation to what actually evicts contracts today.
- **Piece A — the eviction that actually governs retention now** (demand-driven Greedy-Dual `keep_score` + capability-relative RAM budget, shipped 0.2.90; `ring/hosting/{demand,cache}.rs`) was **surfaced nowhere**.
- The Subscribed Contracts table's "Gov" column was always "ok" (fed from the dormant governance `state_by_id`), not a real freshness/demand signal.
- A **stale comment** in `ring.rs` claimed the reaper "defaults to DryRun" when it actually defaults to `Off`.

Ian (project lead) authorized this cleanup.

## Solution

1. **Gate the Contract Governance card** — render it only when governance `mode != Off`, so it vanishes on default nodes but stays for anyone who explicitly enables governance. Removed the now-dead `is_off` empty-state branches.
2. **Fix the stale `ring.rs` comment** (defaults to `Off`, not DryRun).
3. **Replace the always-"ok" Gov column** in Subscribed Contracts with a real **freshness/demand** indicator: `is_receiving_updates` (the true freshness signal — `is_hosting` is not, PR #3699) + `contract_in_use` (real demand). Plumbed two bools through `SubscribedContractSnapshot` / `ContractSnapshot`.
4. **Add a "Demand-driven eviction" card** surfacing piece A: RAM budget used vs headroom, contract count, budget-triggered evictions, the recently-read-eviction miscalibration alarm (#4338), and a per-contract table (keep-score / predicted demand / size / reads) in **eviction order** with a "next to evict" badge on the first **eviction-eligible** row — past `min_ttl` and not in use, i.e. the contract the over-budget sweep would actually evict first; **no badge** when every hosted contract is within-TTL / in use. New `HostingSnapshot` provider reads the canonical hosting cache (no mirror); new `HostingCache::eviction_ordered_scores()` accessor.
5. **De-emphasize** the dormant "auto (governance)" ban reason (only the operator path is live).

**Eviction-eligibility (review follow-up, P2):** the over-budget sweep (`evict_over_budget`) skips contracts that are within `min_ttl` or pinned by demand (`should_retain` / `contract_in_use`), so the raw lowest-keep-score row is not necessarily the one the cache would evict. The badge now marks the first **eligible** row: the cache reports `past_min_ttl` per contract, `HostingManager::is_eviction_eligible` folds in `!contract_in_use`, and the renderer badges the first eligible row (or none). This keeps the "next to evict" label from mislabelling an eviction-exempt contract.

**Deadlock safety (by construction):** the subscription-snapshot booleans are computed in a second pass *after* the `active_subscriptions` iterator's shard guards drop — `is_receiving_updates` re-`.get()`s that same DashMap, which would same-shard deadlock if called inside the iterator. This is prevented **by construction** (the two-phase split), *not* by a test: the regression test `dashboard_subscription_snapshot_reports_freshness_and_demand` characterizes the two-phase **output** (field correctness), but in a single-threaded, no-concurrent-writer test a re-inlined same-shard read would still pass — so it does not itself catch the deadlock. The guard against re-inlining is the load-bearing code comment in `dashboard_subscription_snapshot`. The same two-phase, lock-released-first shape is used for the eviction-eligibility fold (`is_eviction_eligible` reads only the subscription maps; the `hosting_cache` read guard is already released).

### Deferred (noted for follow-up, not in this first slice)
- `eviction_floor` value and live (recomputed) per-contract demand (`distance_and_demand` is private) — the card uses the stored `keep_score`/`predicted_demand` captured at last refresh.
- `abandoned_at` surfacing; exporting the piece-C consult counters (#4658).
- Since every row in the Subscribed Contracts table is by construction an active subscription, `is_receiving_updates` is ~always true there — the varying signal in that table is `in_use`. `is_receiving_updates` still renders as an explicit freshness confirmation and would surface any future divergence.

## Testing

- `governance_card_hidden_when_off` / updated `governance_card_mode_pill_reflects_snapshot_mode` — Off renders nothing.
- `contracts_table_shows_freshness_and_demand_pills` — fresh/stale + in-use/idle pills replace the retired gov column.
- `hosting_card_hidden_when_nothing_hosted` / `hosting_card_renders_budget_and_eviction_order` — budget tiles, recently-read alarm, next-to-evict on the first eviction-eligible row.
- `hosting_card_badges_first_eligible_row_not_lowest_score` / `hosting_card_no_badge_when_nothing_eligible` (P2) — the badge follows eligibility, not raw keep-score order, and disappears when nothing is eligible.
- `dashboard_is_eviction_eligible_matches_sweep_skip_filter` (manager) — `is_eviction_eligible` combines the `past_min_ttl` age gate and the `!contract_in_use` pin, matching the real sweep skip filter.
- `eviction_ordered_scores_flags_past_min_ttl_after_ttl_elapses` (cache) — the `past_min_ttl` flag flips false→true once the entry ages past `min_ttl`.
- `dashboard_subscription_snapshot_reports_freshness_and_demand` (backend) — characterizes the two-phase snapshot output (see the deadlock-safety note; deadlock-safety itself is by construction).
- `eviction_ordered_scores_matches_key_order_and_carries_fields` (cache accessor).
- Updated ban-list reason assertion.
- `cargo fmt` + `clippy` clean on changed code; full `home_page` / `ring::hosting` / `node::network_status` unit suites green.

Draft: needs review + Ian's eyes on the dashboard before merge. No auto-merge.

Refs #4642, #4296
Closes #4676

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -47,6 +47,16 @@ pub type RingStatsProvider = Arc<dyn Fn() -> RingStatsSnapshot + Send + Sync + '
 /// `Ring::contract_ban_list` directly — no mirrored counter to rot.
 pub type BanListProvider = Arc<dyn Fn() -> BanListSnapshot + Send + Sync + 'static>;
 
+/// Provider for the demand-driven hosting snapshot (piece A, #4642). Same
+/// pattern as the other providers: registered at node startup, replaceable
+/// for multi-node test harnesses, read by `get_snapshot` on every dashboard
+/// request. In production the closure captures `Arc<Ring>` and calls
+/// `Ring::dashboard_hosting_snapshot()`, which reads the canonical hosting
+/// cache (Greedy-Dual keep_score + capability-relative RAM budget) — the
+/// mechanism that actually governs retention now, replacing the dormant MAD
+/// governance detector (#4296).
+pub type HostingProvider = Arc<dyn Fn() -> HostingSnapshot + Send + Sync + 'static>;
+
 /// Snapshot of ring-level statistics exposed to the dashboard.
 #[derive(Debug, Clone, Default)]
 pub struct RingStatsSnapshot {
@@ -109,6 +119,16 @@ static BAN_LIST_PROVIDER: parking_lot::RwLock<Option<BanListProvider>> =
 /// harnesses can re-wire to the current node.
 pub fn set_ban_list_provider(provider: BanListProvider) {
     *BAN_LIST_PROVIDER.write() = Some(provider);
+}
+
+static HOSTING_PROVIDER: parking_lot::RwLock<Option<HostingProvider>> =
+    parking_lot::RwLock::new(None);
+
+/// Register the dashboard's demand-driven hosting data source (piece A,
+/// #4642). Replaces any previously-registered provider so multi-node
+/// in-process harnesses can re-wire to the current node.
+pub fn set_hosting_provider(provider: HostingProvider) {
+    *HOSTING_PROVIDER.write() = Some(provider);
 }
 
 /// Replaceable storage for the subscription provider. Wrapped in a
@@ -645,6 +665,11 @@ pub struct NetworkStatusSnapshot {
     /// provider closure — no mirrored counter to rot. Empty when nothing
     /// is banned (the common case).
     pub ban_list: BanListSnapshot,
+    /// Demand-driven hosting state (piece A, #4642): the capability-relative
+    /// RAM budget + per-contract Greedy-Dual keep_score that actually governs
+    /// retention. Drives the "Demand-driven eviction" card. Read from the
+    /// canonical hosting cache via the provider closure — no mirrored counter.
+    pub hosting: HostingSnapshot,
 }
 
 /// Snapshot of the contract ban list for the dashboard (#4302).
@@ -863,6 +888,56 @@ pub struct ContractSnapshot {
     pub instance_id: String,
     pub subscribed_secs: u64,
     pub last_updated_secs: Option<u64>,
+    /// Whether this node is genuinely receiving updates for the contract —
+    /// the real per-contract freshness signal (`is_hosting` is NOT; only an
+    /// active subscription keeps the cache fresh, PR #3699).
+    pub is_receiving_updates: bool,
+    /// Whether real demand pins the contract (local client subscription or a
+    /// registered downstream subscriber).
+    pub in_use: bool,
+}
+
+/// Snapshot of the demand-driven hosting cache (piece A, #4642) for the
+/// local-peer dashboard. This is the mechanism that actually governs
+/// retention today — a capability-relative RAM budget plus a Greedy-Dual
+/// `keep_score` per contract (`ring/hosting/{cache,demand}.rs`) — and it
+/// replaced the dormant MAD `GovernanceManager` (#4296). Default (all zeros,
+/// no contracts) when the node hosts nothing yet or the provider is unset.
+#[derive(Default, Clone)]
+pub struct HostingSnapshot {
+    /// Configured RAM-scaled byte budget for hosted contract state.
+    pub budget_bytes: u64,
+    /// Bytes currently used by hosted contract state. Headroom = budget - used.
+    pub used_bytes: u64,
+    /// Number of contracts currently in the hosting cache.
+    pub contract_count: u64,
+    /// Monotonic count of contracts evicted specifically for being over budget.
+    pub budget_evictions_total: u64,
+    /// Over-budget evictions whose victim had genuine repeat demand (read >= 2
+    /// times). Should stay near zero; a rising value means the demand estimate
+    /// is mis-ordering the working set (the #4338 miscalibration symptom).
+    pub evictions_of_recently_read_total: u64,
+    /// Per-contract Greedy-Dual rows in EVICTION order (the next victim under
+    /// budget pressure is first). May be truncated by the renderer; the full
+    /// count is `contract_count`.
+    pub contracts: Vec<HostedContractEntry>,
+}
+
+/// One hosted contract's demand-driven eviction row for the dashboard.
+#[derive(Clone)]
+pub struct HostedContractEntry {
+    /// Full contract-key string.
+    pub key_full: String,
+    /// Truncated key for display.
+    pub key_short: String,
+    /// Greedy-Dual priority (`eviction_floor + predicted_demand`). Lowest evicts first.
+    pub keep_score: f64,
+    /// Stored per-contract read-demand estimate (reads/second).
+    pub predicted_demand: f64,
+    /// Per-contract memory cost (state bytes).
+    pub size_bytes: u64,
+    /// Read accesses (GET/SUBSCRIBE) observed over this entry's residency.
+    pub read_count: u32,
 }
 
 /// Snapshot of operation stats. Each tuple is `(success_count, failure_count)`.
@@ -1000,6 +1075,8 @@ pub fn get_snapshot() -> Option<NetworkStatusSnapshot> {
                         instance_id,
                         subscribed_secs: c.subscribed_secs,
                         last_updated_secs: c.last_updated_secs,
+                        is_receiving_updates: c.is_receiving_updates,
+                        in_use: c.in_use,
                     }
                 })
                 .collect()
@@ -1070,6 +1147,11 @@ pub fn get_snapshot() -> Option<NetworkStatusSnapshot> {
             .map(|provider| provider())
             .unwrap_or_default(),
         ban_list: BAN_LIST_PROVIDER
+            .read()
+            .as_ref()
+            .map(|provider| provider())
+            .unwrap_or_default(),
+        hosting: HOSTING_PROVIDER
             .read()
             .as_ref()
             .map(|provider| provider())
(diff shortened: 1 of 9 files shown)


sanity:
## Light-tier review — Codex external pass + adversarial read

Read-only telemetry/rendering surface (no routing/transport/wire/hosting-behavior change), so Light tier: one external model pass (Codex) + one adversarial read. **Verdict: safe — no panic, deadlock, or behavior regression. One substantive display-fidelity finding, found independently by both reviewers.** Fine to keep as a draft for Ian's eyeball; the one finding is worth addressing before the card is relied upon.

### Findings

**[P2 / medium] "Next to evict" badge can point at an eviction-exempt contract** — `crates/core/src/server/home_page/cards.rs:1155` (badge) + `crates/core/src/ring.rs:3140` / `ring/hosting/cache.rs:838` (provider ordering).
Both Codex and the adversarial read flagged this independently. `build_hosting_card` unconditionally badges `h.contracts[0]` (the lowest keep-score row) as "next to evict", but the provider (`eviction_ordered_scores`) sorts *all* hosted contracts by keep-score **without eligibility filtering**. The real sweep (`evict_over_budget` / `sweep_expired`, `cache.rs:891-909`) only evicts entries that are **past `min_ttl` AND not `should_retain`** (active client/downstream subscription). So a freshly-inserted within-TTL contract, or a low-score in-use contract, can be labelled "next to evict" when the cache will actually skip it. On a read-only dashboard this misleads rather than breaks. Fix options: carry eligibility into the snapshot and badge the first *eligible* row, or soften the label to "lowest keep-score" without the eviction claim. Not a merge blocker for a draft, but should be resolved before the badge is trusted.

**[P3 / low] Deadlock regression test guards field-correctness, not deadlock-safety** — `crates/core/src/ring/hosting.rs:2598` (`dashboard_subscription_snapshot_reports_freshness_and_demand`).
The two-phase refactor is correct: `is_receiving_updates → is_subscribed` re-`.get()`s `active_subscriptions` (the same map the phase-1 iterator locks), while `contract_in_use` touches only `client_subscriptions`/`downstream_subscribers` — so moving the boolean computation to phase 2 (after the iter guards drop) genuinely removes the one same-map re-entrant call. However, the cited regression test runs single-threaded with no concurrent writer, where a same-shard read-after-read on parking_lot generally *succeeds* (shared read) — the deadlock only manifests under a queued writer. So if the call were re-inlined into the iterator, this test would likely still pass rather than hang. The test validly pins the freshness/demand field values; it just doesn't deterministically guard the deadlock the way the comment/PR description imply. Consider softening that claim or adding a concurrent-writer stress variant. Non-blocking.

### Verified clean
- **Snapshot plumbing reads the canonical cache correctly.** `current_bytes → used_bytes` and all stat fields map correctly; per-contract rows come straight from `HostingCache` under one read lock (no mirror). Eviction **order is right** (ascending keep-score = next victim first), and `eviction_ordered_scores` uses the identical `(keep_score, last_access_seq, key)` sort as `keys_eviction_order` — pinned by `eviction_ordered_scores_matches_key_order_and_carries_fields`.
- **Governance gating has no regression for enabled nodes.** Early `return String::new()` on `Off`; the removed `is_off` empty-state branches are genuinely dead (mode can't be `Off` past the guard); DryRun/Enforce still render, covered by updated tests.
- **No panic on the render path.** Div-by-zero guarded (`budget_bytes.max(1)`), `saturating_sub` headroom, `used_pct.min(100.0)`; both cards return empty on `None` / nothing-hosted; format placeholders all resolve to named args.
- **Genuinely behavior-free beyond the dashboard.** All new accessors take `&self` and only read; the two-phase `dashboard_subscription_snapshot` refactor produces equivalent output plus the two new bools; the only non-dashboard edit is the accurate `ring.rs` comment fix (reaper defaults to `Off`, not DryRun). No routing/hosting/subscription/wire change — Light tier is appropriate, and it's consistent with the hosting-invariants "read the canonical cache, no mirror" pattern.

[AI-assisted - Claude]

sanity:
Addressed both review findings in 94d50a18.

**P2 — "next to evict" badge could mislabel an eviction-exempt contract.** The card badged the lowest-keep-score row unconditionally, but the over-budget sweep (`evict_over_budget`) skips contracts that are within `min_ttl` or pinned by demand (`should_retain` / `contract_in_use`) — so a within-TTL or in-use low-score contract was being labelled "next to evict" even though the cache would never evict it (the exact "dashboard misleads the operator" failure this PR exists to fix). Took the preferred fix — carry an eviction-eligibility flag through the snapshot:

- `HostingContractScore` gains `past_min_ttl` (the cache half of the sweep's `age >= min_ttl` filter), computed in `eviction_ordered_scores`.
- `HostingManager::is_eviction_eligible(&score)` folds in `!contract_in_use` — mirroring the full `evict_over_budget` per-contract skip filter.
- `HostedContractEntry` gains `eviction_eligible`; `Ring::dashboard_hosting_snapshot` populates it (two-phase, lock already released before `contract_in_use` — no re-lock deadlock).
- The renderer badges the **first eligible** row, and shows **no badge** when every hosted contract is within-TTL / in use.

New tests: `dashboard_is_eviction_eligible_matches_sweep_skip_filter` (manager combines both halves), `eviction_ordered_scores_flags_past_min_ttl_after_ttl_elapses` (cache flag flips past TTL), `hosting_card_badges_first_eligible_row_not_lowest_score` and `hosting_card_no_badge_when_nothing_eligible` (renderer).

**P3 — deadlock-test wording overclaimed.** Softened the docstring on `dashboard_subscription_snapshot_reports_freshness_and_demand` and the PR description: the test characterizes the two-phase *output* (field correctness); the same-shard DashMap re-lock is prevented **by construction** (phase 2 runs after the iterator's shard guards drop) and would not manifest in this single-threaded, no-writer test — a re-inlined same-shard read would still pass. The guard against re-inlining is the load-bearing code comment, not the test.

`cargo fmt` + `cargo clippy --locked -p freenet -- -D warnings` clean; touched `home_page` / `ring::hosting` / cache unit tests green. Staying draft for your live-rendering eyeball — no auto-merge.

[AI-assisted - Claude]

PR #4685, opened 2026-07-05
feat(hosted): add "New ID" control to hosted Account popover

### LINKED ISSUE #4645: Suggestion and Issue with using try.freenet.org

sanity:
## Problem

On try.freenet.org the hosted Account popover lets a user copy, restore, and
export their per-user access key, but there is no way to **start over with a
fresh identity**. To get a new ID a user has to open browser devtools and
hand-delete `localStorage['__freenet_user_token__']`.

This was reported directly by a user testing try.freenet.org (see #4645):

> I'm not seeing an Account option for generating a new ID still. I had to
> manually delete the local token to get a fresh one.

#4645 bundled two asks: (a) a clearer new-tab error page and (b) an in-UI way
to generate a new ID. PR #4652 fixed (a) and explicitly noted it did **not**
resolve (b), but the merge auto-closed #4645 anyway. This PR is (b).

## Solution

Add a **New ID** button to the Access-key section of the Account popover,
next to Copy/Restore:

- It `window.confirm()`s first — the action is destructive (the old data stays
  under the old access key), so the prompt tells the user to "Copy key" first if
  they want to get back to it.
- On confirm it clears `localStorage['__freenet_user_token__']` and reloads.
  `SHELL_USER_TOKEN_JS` already mints a fresh random token on load when the key
  is absent, so **no backend change is needed** — this mirrors the existing
  "Restore from key" handler, which writes the same key.
- The button row gets `flex-wrap` so the third button doesn't overflow the
  fixed-width (300px) popover.

## Testing

- New pin test `hosted_bar_offers_new_id_control_4645`: asserts the button
  exists, the handler confirms **before** clearing (so cancelling keeps the
  current identity), reload follows the clear, and both the clear site and the
  minting site (`SHELL_USER_TOKEN_JS`) key off the same `localStorage` key — a
  rename on either side would silently turn the reset into a no-op.
- `cargo fmt` clean, `cargo clippy -p freenet --lib -- -D warnings` clean, all
  `path_handlers` unit tests pass (82).

This is a client-only change to shell-chrome assets (JS/HTML/CSS strings) plus a
test; the per-user token mechanism and the WS/backend paths are untouched.

## Fixes

Closes the remaining half of #4645.

[AI-assisted - Claude]

https://claude.ai/code/session_01KHNrpsMi7tz5j7PH84fvhe

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -1083,15 +1083,15 @@ const SHELL_USER_TOKEN_JS: &str = include_str!("path_handlers/assets/shell_user_
 const HOSTED_BAR_STYLES: &str = include_str!("path_handlers/assets/hosted_bar.css");
 
 /// Markup for the hosted-mode bar: the always-visible "not private" disclosure
-/// plus an Account popover with the access-key backup/restore and the
-/// export-to-your-own-peer action. The access key is the per-user token, read
-/// from the shell-only `__freenet_user_token` global — it never enters the
-/// sandboxed iframe. The Export action is a placeholder until the node-side
-/// export endpoint lands.
+/// plus an Account popover with the access-key backup/restore, a "New ID"
+/// control to start over with a fresh identity, and the export-to-your-own-peer
+/// action. The access key is the per-user token, read from the shell-only
+/// `__freenet_user_token` global — it never enters the sandboxed iframe.
 const HOSTED_BAR_HTML: &str = include_str!("path_handlers/assets/hosted_bar.html");
 
 /// Behavior for the hosted-mode bar (toggle popover, copy/restore the access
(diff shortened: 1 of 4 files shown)


sanity:
## Review (Light tier — client-only shell assets + test)

**External model (Codex, `codex review --base main`):** clean — no correctness, security, or maintainability findings.

**Local checks:** `cargo fmt` clean, `cargo clippy -p freenet --lib -- -D warnings` clean, `path_handlers` unit tests pass (82), new pin test `hosted_bar_offers_new_id_control_4645` passing.

Change is confined to hosted shell-chrome assets (JS/HTML/CSS strings) plus the pin test; the per-user token minting/validation and the WS/backend paths are untouched.

[AI-assisted - Claude]

PR #4692, opened 2026-07-05
feat(config): add --disable-auto-update opt-out for from-source deployments

### LINKED ISSUE #4690: feat: add --disable-auto-update opt-out for from-source deployments (fixes try.freenet.org exit-42 loop)
## Problem

A node built from source that runs *ahead* of the latest GitHub release (e.g. the **try.freenet.org** hosted node — a hand-built `cargo build --release` that carries unreleased features) still runs the auto-update check. It detects the newer *published* release, exits code 42 to request a supervisor-applied update, and then:

- under a supervisor that runs `freenet update`, **installs the stock release over the from-source build**, discarding the unreleased features; or
- under a plain `Restart=on-failure` unit (which is what `freenet-hosted.service` is), **crash-loops** — re-detect → exit 42 → restart → repeat. On try.freenet.org the restart counter reached **138**.

The existing off-switch, `auto_update_disabled = !build_info::GIT_DIRTY.is_empty()` (`crates/core/src/bin/freenet.rs:363`), only covers a **dirty** build tree. A **clean** build from a committed tree that is simply ahead of releases sails right past it. `.claude/rules/deployment.md` already intends "disable auto-update for dev builds; proceed for release" — but only the dirty half is implemented; the clean-but-unofficial case is unaddressed.

## Proposed fix

Add an **opt-in** `--disable-auto-update` CLI flag (also `FREENET_DISABLE_AUTO_UPDATE` env), **default false**, OR-ed into the existing gate. A normal release node that does not set it auto-updates **exactly as before** — this must not change (peers need to update automatically given Freenet's release pace). Bespoke from-source deployments set it (and it's added to `freenet-hosted.service`).

- No change to update detection / comparison / install logic.
- Non-persisted runtime flag (`#[serde(skip)]`), set from the CLI/env at `build()`.
- Pin test locks the invariant: clean build + flag unset → auto-update stays ON.

Scope: strictly additive, opt-in; does not touch release-build behavior. Follow-up to the try.freenet.org hosted-mode work (#4381).

[AI-assisted - Claude]

sanity:
## Problem

A node built from source that runs *ahead* of the latest GitHub release still runs the auto-update check. It detects the newer *published* release, exits code 42 to request a supervisor-applied update, and then either:

- under a supervisor that runs `freenet update`, **installs the stock release over the from-source build**, discarding unreleased features; or
- under a plain `Restart=on-failure` unit, **crash-loops** (re-detect → exit 42 → restart → repeat).

The **try.freenet.org** hosted node — a hand-built `cargo build --release` carrying unreleased features (the New ID button, #4685) — hit exactly this: its restart counter reached **138**.

The existing off-switch, `!build_info::GIT_DIRTY.is_empty()` (`freenet.rs`), only covers a **dirty** tree. A **clean** build that is merely ahead of releases sails right past it. `.claude/rules/deployment.md` already intends *"disable for dev, proceed for release"* — but only the dirty half was implemented; the clean-but-unofficial case was unaddressed.

## Solution

An **opt-in** `--disable-auto-update` CLI flag, **default `false`**, OR-ed into the existing gate:

```rust
let auto_update_disabled = auto_update_is_disabled(build_info::GIT_DIRTY, config.disable_auto_update);
```

- **The default is unchanged and load-bearing:** a release node that does not set the flag auto-updates *exactly* as before. Peers must keep updating automatically given Freenet's release pace — this PR does not change that.
- No change to update **detection / comparison / install** logic — only whether the check task runs at all (it parks, same as the dirty-build path).
- The flag maps to a **non-persisted** (`#[serde(skip)]`) runtime `Config` field set at `build()` (the systemd unit passes it every start).
- **Plain CLI flag, no `env` binding** — deliberately. A stray truthy env value silently disabling auto-update fleet-wide is the exact failure this must avoid; the one bespoke deployment (try.freenet.org) sets `--disable-auto-update` explicitly in its `ExecStart`.

## Testing

- **`auto_update_default_stays_enabled_flag_and_dirty_disable`** (bin) — pins the load-bearing invariant on the pure gate helper: clean build + flag unset → auto-update **ON**; flag set → OFF; dirty build → OFF regardless. Fails if the default ever flips.
- **`disable_auto_update_flag_parses_from_cli`** — `--disable-auto-update` parses to `true`, bare invocation to `false` (at the clap boundary).
- **`disable_auto_update_flag_wires_through_build`** — the flag reaches the built `Config`; default leaves it `false`.
- **`all_persisted_config_fields_round_trip_through_build`** — still passes; the new field is classified non-persisted (bound `_` in the exhaustive destructure).
- `cargo fmt` + `cargo clippy -p freenet --lib --bins -- -D warnings` clean.

## Review

Full-tier (deployment-critical surface). External **Codex** + an adversarial Claude pass, both focused on *"could this break auto-update for release nodes?"* Both independently confirmed the release path is unaffected (default traced end-to-end; nothing sets the flag on real nodes; threading/gate/`serde(skip)` correct). Both flagged that an `env`-bound bool would reject `FREENET_DISABLE_AUTO_UPDATE=1`; resolved by dropping `env` entirely (CLI flag only). Re-run of the review on the updated code is in the PR comments.

## Fixes

Closes #4690.

[AI-assisted - Claude]

https://claude.ai/code/session_01KHNrpsMi7tz5j7PH84fvhe

### DIFF
--- a/.claude/rules/deployment.md
+++ b/.claude/rules/deployment.md
@@ -34,6 +34,13 @@ CHECK: env!("VERGEN_GIT_DIRTY") or equivalent build metadata
 
 WHY: Dev builds triggering auto-update replaces the dev binary with a release
 binary, destroying the development environment.
+
+The GIT_DIRTY gate only covers a DIRTY tree. A CLEAN build that is not an
+official release but intentionally runs AHEAD of the latest release (e.g. the
+try.freenet.org from-source node) has GIT_DIRTY empty, so it is NOT gated — it
+would detect the newer published release and exit 42 in a loop. For that case,
+pass `--disable-auto-update` on that deployment's `ExecStart` (default is off,
+so release nodes are unaffected). See #4690.
 ```
 
 ### WHEN tightening security (sandbox, CSP, CORS)
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -186,6 +186,23 @@ async fn run_local(config: Config) -> anyhow::Result<()> {
         .map_err(anyhow::Error::msg)
 }
 
+/// Whether the node's automatic self-update check should be skipped entirely.
+///
+/// Two independent reasons, either sufficient:
+/// - a dirty/dev build (`git_dirty` non-empty): `freenet update` would replace
+///   the binary with a prebuilt release and discard local changes (#3245);
+/// - an explicit operator opt-out (`--disable-auto-update`, #4690): for a
+///   bespoke from-source deployment that intentionally runs ahead of releases
+///   (e.g. try.freenet.org) and must not try to "update" itself to a published
+///   release.
+///
+/// A clean release build with the flag unset returns `false`, so auto-update
+/// stays ON by default. This is load-bearing: peers must keep updating
+/// automatically given Freenet's release pace — do not flip the default.
+fn auto_update_is_disabled(git_dirty: &str, disable_flag: bool) -> bool {
+    !git_dirty.is_empty() || disable_flag
+}
+
 async fn run_network(config: Config) -> anyhow::Result<()> {
     tracing::info!("Starting freenet node in network mode");
 
@@ -198,6 +215,10 @@ async fn run_network(config: Config) -> anyhow::Result<()> {
         .await
         .with_context(|| "failed to start HTTP/WebSocket client API")?;
     tracing::info!("Initializing node configuration");
+
+    // Capture before `config` is moved into NodeConfig; threaded to the
+    // update-check task so an operator opt-out (#4690) can park it.
+    let disable_auto_update = config.disable_auto_update;
 
     let node_config = NodeConfig::new(config)
         .await
@@ -212,7 +233,7 @@ async fn run_network(config: Config) -> anyhow::Result<()> {
     let shutdown_handle = node.shutdown_handle();
 
     // Run node with signal handling for graceful shutdown
-    run_network_node_with_signals(node, shutdown_handle).await
+    run_network_node_with_signals(node, shutdown_handle, disable_auto_update).await
 }
 
 /// Run the network node with signal handling for graceful shutdown.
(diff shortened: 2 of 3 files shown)


sanity:
## Review — Full tier (deployment-critical: auto-update)

Two independent reviewers, both focused on the overriding requirement: **could this break auto-update for normal release nodes?** Answer: no — verified end-to-end (default traced from clap → `Config` → the gate; nothing sets the flag on real nodes; the generated release systemd unit does not set it; threading/gate/`serde(skip)` correct).

**Findings caught and resolved across review rounds** (all on the opt-in side; none touched the release path):

1. **Env `=1`/`=0` would crash the node / doc mismatch** (Codex + adversarial Claude). A clap `bool` arg with `env` only accepts `true`/`false`, so the documented `FREENET_DISABLE_AUTO_UPDATE=1` would fail parsing. → **Resolved by dropping `env` entirely** — plain `--disable-auto-update` CLI flag, no env footgun. A stray truthy env value silently disabling updates fleet-wide is exactly the failure to avoid.
2. **Flag silently ignored before the subcommand** (Codex). `freenet --disable-auto-update network` parsed into top-level `cli.config`, which the `Network` arm discarded. → **Resolved** by OR-merging the top-level flag into the `Network`/`Local` subcommand config, so both positions work. Pinned by `disable_auto_update_flag_honored_before_or_after_subcommand`.
3. **Stale env-var references in two doc comments** (Codex). → Removed.

**Final Codex pass: clean, no findings.**

Tests: `auto_update_default_stays_enabled_flag_and_dirty_disable` (gate invariant: clean build + no flag → auto-update ON), `disable_auto_update_flag_parses_from_cli`, `disable_auto_update_flag_honored_before_or_after_subcommand`, `disable_auto_update_flag_wires_through_build`, `all_persisted_config_fields_round_trip_through_build`; full existing auto-update suite still green. `cargo clippy -- -D warnings` clean.

[AI-assisted - Claude]

Issue #4683, opened 2026-07-04
Add an aggregate disk-space budget (hard cap / % of available) to demand-driven hosting


iduartgomez:
## Summary

The demand-driven hosting redesign (#4642) sizes the hosting budget from **memory only** (`default_hosting_budget_bytes()` in `ring/hosting/cache.rs`, via `read_total_ram_bytes()`). There is no analogous aggregate **disk** budget. The design's own stated goal is "a peer hosts the contracts it is most likely to be asked for, within a budget sized by its own scarcest resource" (`docs/design/hosting-eviction.md`) — for a disk-constrained host (small VPS, SD-card device) disk, not RAM, may be the scarcest resource, and today nothing bounds it in aggregate.

## Current state (what exists today)

- **Per-contract hard cap**: `MAX_STATE_SIZE = 50 MiB` (`wasm_runtime/state_store.rs`), enforced at PUT/UPDATE time. Rejects a single oversized contract; says nothing about the total.
- **Memory-derived byte budget**: `default_hosting_budget_bytes()` = `clamp(total_ram / 8, 128 MiB, 1 GiB)`, drives the Greedy-Dual demand-ordered eviction walk (`evict_over_budget`). On eviction, on-disk state is now also reclaimed (#4212) — but this is a side effect of the *memory* budget's eviction event, not an independent disk-triggered eviction pass.
- **Explicit disclaimer in the code**, `ring/hosting/cache.rs`, right on `default_hosting_budget_bytes()`:
  > "The budget tracks hosted contract **state** bytes only. WASM code blobs and ReDb/SQLite database overhead are additional and not counted against it, so **this is not a hard bound on total on-disk usage.**"

So today: no aggregate disk guard exists at all. If a node's disk actually fills up, nothing proactively refuses/evicts — the node hits an OS-level write failure first.

## Disk consumers inventory (why "just add one disk_budget" isn't quite enough)

Confirmed by reading the code: disk usage is spread across **several independently-bounded (or unbounded) subsystems that share no common accounting**, so a single new `disk_budget` needs to either cover all of them or be scoped explicitly against a stated subset. As found:

| Consumer | Current bound | Where |
|---|---|---|
| Contract state (hosted) | Approximated via the memory budget; per-contract hard cap 50 MiB | `ring/hosting/cache.rs`, `wasm_runtime/state_store.rs` |
[...]

PR #4707, opened 2026-07-06
feat: disk budget sizing + periodic recompute + eviction floor

iduartgomez:
> **Stacked on #4700 (PR 1).** Cross-repo PRs from a fork can only target
> `main`, so this diff includes PR 1's commit. Review after #4700; the
> PR-2-only change is the second commit (`feat: disk budget sizing …`).
> Part of #4683.

## Problem

The demand-driven hosting budget (#4642) sizes itself from **RAM only**
(`clamp(total_ram/8, 128 MiB, 1 GiB)`). For a disk-constrained host (small VPS,
SD-card device) disk — not RAM — is the scarcest resource, yet **nothing bounds
disk in aggregate**. If the disk fills, the node hits an OS-level write failure
before anything proactively evicts.

PR 1 (#4700, `feat/disk-budget-accounting`, the base of this stacked PR) landed the
`DiskUsageTracker` accounting + telemetry with zero behavior change. This PR turns
that measurement into the first enforcement mechanism.

## Solution

Add a disk budget as a **second floor** on the *same* Greedy-Dual eviction walk:
`effective_budget = min(ram_budget, disk_budget)`, so the tighter of the two
scarce resources governs retention. Recomputed on the existing 60s sweep.

- **`available_bytes(path)`** — free bytes on the data-dir mount via
  `libc::statvfs` (unix) / `GetDiskFreeSpaceExW` (windows) / `None`-fallback,
  same libc-FFI + `SAFETY` discipline as `read_total_ram_bytes`. A failed/unsupported
  read → caller falls back to `u64::MAX`, which makes the disk budget clamp to its
  cap (degrade to "cap only", never to zero).
- **`disk_budget_for(used, avail, pct)`** = `clamp(pct*(used+avail), MIN, cap)`,
  a pure, filesystem-free clamp beside `budget_for_ram`. Basis is capacity
  *reachable by Freenet* (`used + free`), NOT a naive free-space read that shrinks
  as the disk fills. Non-finite/negative `pct` and `used+avail` overflow saturate
  to a valid budget.
- **New disk defaults** `DEFAULT_HOSTING_DISK_PCT` (0.5) and
  `DEFAULT_MAX_HOSTING_DISK_BYTES` (32 GiB) — a disk budget legitimately exceeds
  the 1 GiB RAM ceiling.
- **Persisted config** `hosting_disk_pct` + `max_hosting_disk` mirroring
  `max_hosting_storage`, incl. `build()` `get_or_insert` merge **and** non-default
  seeds in the round-trip guard test (silent-revert class #3890/#4275).
- **`HostingCache::set_budget_bytes`** setter (the only writer after the ctor).
- **`HostingManager::recompute_effective_budget(available)`** — refresh `du` →
  `used = tracker.total_bytes()` → `disk_budget` → `effective = min(ram, disk)` →
  `set_budget_bytes`. The `du`-walk stays OUTSIDE the cache lock; only the O(1)
  `set_budget_bytes` takes it. Wired into the 60s sweep after the disk-usage
  refresh; lazy seed keeps it a no-op until the first tick.

The admission gate (hard pre-write bound on aggregate disk) is the next PR in the
stack; this PR ships the soft, self-correcting eviction floor only.

## Testing

- `disk_budget_for` boundaries: zero/huge free, `pct` 0/1, clamp both ends,
  saturating overflow, non-finite/negative `pct`, explicit clamp bounds.
- `min(ram, disk)`: each side wins, equal, and the MIN-floor case.
- `recompute_effective_budget` is a no-op until the tracker is seeded.
- `recompute` does **not** deadlock against concurrent `record_contract_access`.
- `available_bytes` shape on a real mount vs a nonexistent path.
- Config round-trip guard exercises the two new persisted fields.

Gate: `cargo fmt` + `cargo clippy -p freenet -- -D warnings` clean; full `-p freenet`
lib suite green except two **pre-existing** macOS case-insensitive-FS failures in
`wasm_runtime::migration_tests` (fail identically on the base branch, pass on Linux CI).

## Fixes

Part of #4683.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -120,6 +120,19 @@ pub struct ConfigArgs {
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
 
+    /// Fraction (0.0–1.0) of the disk capacity *available to Freenet*
+    /// (`used + free` on the data-dir mount) used to size the aggregate disk
+    /// budget (#4683). The disk budget is the second floor on hosting eviction:
+    /// `effective_budget = min(ram_budget, disk_budget)`. Default: 0.5.
+    #[arg(long, env = "HOSTING_DISK_PCT")]
+    pub hosting_disk_pct: Option<f64>,
+
+    /// Hard upper clamp in bytes for the aggregate disk budget (#4683). Mirrors
+    /// `--max-hosting-storage` for disk: the disk budget never exceeds this even
+    /// on a host with a very large data disk. Default: 32 GiB.
+    #[arg(long, env = "MAX_HOSTING_DISK")]
+    pub max_hosting_disk: Option<u64>,
+
     /// Per-user secret-storage quota in bytes for HOSTED mode (#4561, P5 of
     /// #4381). Bounds a single hosted user's (one `userToken`) TOTAL on-disk
     /// footprint under their `users/<user_id>/` tree, summed across every
@@ -246,6 +259,8 @@ impl Default for ConfigArgs {
             version: false,
             max_blocking_threads: None,
             max_hosting_storage: None,
+            hosting_disk_pct: None,
+            max_hosting_disk: None,
             per_user_secret_quota_bytes: None,
             per_user_inactive_ttl_secs: None,
             inactive_user_sweep_interval_secs: None,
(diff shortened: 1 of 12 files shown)


iduartgomez:
Superseded by #4702. The disk-budget stack (PR1 accounting → PR2 eviction floor → PR3 admission gate → PR4 UPDATE growth-only) was collapsed and merged as a single squash commit (30db9d70e) via #4702, which carried the cumulative branch content. This PR's eviction-floor content (`recompute_effective_budget`, `disk_budget_for_clamped`, the 60s-sweep recompute wiring) is already in main — a rebase onto current main yields an empty diff. Closing as merged-via-#4702.

PR #4708, opened 2026-07-06
feat: growth-only disk gate for UPDATE (admission belongs on PUT)

iduartgomez:
> **Stacked on #4700 → #4702 (PRs 1–3).** Cross-repo PRs from a fork can only
> target `main`, so this diff includes the ancestor commits. Review after
> #4702; the PR-4-only change is the top commit
> (`feat: growth-only disk gate for UPDATE …`). Part of #4683.
>
> **Gate note:** `cargo clippy -D warnings` and two `wasm_runtime::migration_tests`
> fail on the current `main` base (clippy-1.94.0 lints in files this PR does not
> touch: `SweepOutcome` unused import, `assertions_on_constants`, `redundant_locals`;
> and macOS case-insensitive-FS rename tests). Verified identical on `main` with this
> PR's changes stashed. This PR's own code is fmt+clippy clean; those base failures
> want a separate cleanup PR.

## Problem

PR 4 of the aggregate disk-space budget (#4683). The prior revision made the
UPDATE disk gate *growth-only* but still **rejected any growth-over-budget
UPDATE outright**. That is admission control on *mutation*, which the approved
plan explicitly rejects:

- An UPDATE to an already-hosted contract mutates an **already-counted**
  footprint — it is not a new admission. Rejecting a growing CRDT merge does not
  free the bytes it needs, and evicting the contract to fit its own UPDATE would
  lose the contract entirely.
- CRDT merges frequently shrink or hold size (`delta <= 0`); charging full
  new-state and rejecting falsely blocks convergence.
- For a **relayed** UPDATE the rejection is silently dropped (fire-and-forget,
  no `UpdateMsg::Error`), so nothing even signals the stall.

**The rule: admission control belongs on admission (PUT), not on mutation
(UPDATE).**

## Solution

`HostingManager::admit_state_update` now implements the plan's soft, growth-only
rule with eviction escalation:

- `delta <= 0` (shrink/hold) -> **always admits**, even deeply over budget --
  convergence must never block.
- `delta > 0` over budget -> **admits anyway and sheds OTHER low-value
  contracts** to make room (lowest `keep_score`, past-TTL, not-in-use,
  non-target -- exactly the over-budget sweep's victims), queuing them for the
  same async reclamation the 60s sweep drains. **Only** when there is not enough
  sheddable capacity (everything else retained/in-use) does it reject, via
  option (a): the local client sees `HostResult::Err`; a relayed UPDATE is
  silently dropped and the originator reconciles via CRDT/interest-sync.
- Fresh **PUT keeps the hard `admit_state_write` gate** (new footprints enter
  there and PUT carries `PutMsg::Error` propagation). The same growth-only +
  escalation rule applies to **V2 UPDATE** via the state-admit callback.
- **Not** adding `UpdateMsg::Error` (option b): a wire-format change fighting
  the intentional fire-and-forget design.

The escalation is fully encapsulated in `admit_state_update` -- no chokepoint or
callback signature changes. It releases the disk-tracker read lock, then under
the hosting-cache write lock checks evictable capacity and sheds via two new
cache methods (`evict_lowest_value_to_free`, `evictable_bytes_excluding`).
`contract_in_use` reads only subscription maps, so there is no nested cache-lock
hazard. Freed bytes drop from the disk tracker when the async delete lands, so
the decision is soft and self-correcting. The rule is documented at each gate
site (executor UPDATE chokepoint, re-PUT merge, V2 callback).

## Testing

`cargo test -p freenet` (lib): all disk-budget / hosting / cache / update tests
pass. New tests:

- UPDATE `delta <= 0` admits even deeply over budget (**pin**: convergence
  never blocks; contrasted with a fresh PUT that *is* rejected -- the asymmetry).
- growth over budget **evicts a DIFFERENT low-score contract** (and queues it
  for reclamation) instead of rejecting; **never** evicts the target.
- growth **rejects when the only other contract is in-use** (option a terminal
  case).
- `UpdateMsg` has **no wire error-abort variant** (exhaustive-match compile
  pin) -- a relayed over-budget UPDATE emits no wire error.
- cache-level unit tests for `evict_lowest_value_to_free` (excludes target,
  stops at `needed`, skips retained/within-TTL, zero-is-noop) and
  `evictable_bytes_excluding`.

Note: `cargo clippy -D warnings` and two `wasm_runtime::migration_tests` fail on
the **base branch already** (pre-existing: `SweepOutcome` unused import,
`assertions_on_constants` in bin, `redundant_locals`, macOS case-insensitive-FS
filename tests) -- none in files this PR touches; left untouched to avoid scope
creep.

## Fixes

Part of #4683. Stacked on `feat/disk-budget-admission-gate` (PR 3, freenet#4702).


iduartgomez:
Superseded by #4702. Same collapse as #4707: the growth-only UPDATE gate (`admit_state_update`, the PUT-vs-UPDATE asymmetry) landed in main via the #4702 squash (30db9d70e). Content already present in main; rebase yields an empty diff. Closing as merged-via-#4702.

PR #4709, opened 2026-07-06
fix: gate relay GET serve on freshness, not cache presence

sanity:
## Problem

On the relay/network GET serve path, `check_local_with_interest_gate`
(`crates/core/src/operations/get/op_ctx_task.rs`) decided serve-from-cache
vs forward using `InterestManager::has_local_interest`. That predicate is
true whenever the cache-presence `hosting` flag is set — independent of
whether the peer is subscribed. So a relay served an **unsubscribed,
possibly-stale** cached copy as its FIRST choice, violating
[hosting-invariants.md](.claude/rules/hosting-invariants.md) invariant 1
("a served copy is fresh").

The correct freshness signal is `Ring::is_receiving_updates`
(`is_subscribed() || has_client_subscriptions()`). The **local-client**
GET path already gates on it (the #2388 / #3340 fix in `client_events.rs`);
the relay path never got that protection — and it even *contradicted* the
local path, serving a stale copy the client path had just refused.

**0.2.92 activated this latent bug.** E-GET (#4689) stopped GET
auto-subscribe and interest-gated renewal (#4697) lapses idle
subscriptions, turning formerly-subscribed-fresh cached copies into
unsubscribed-stale ones the relay path still served first.

## Approach

Gate the relay serve on `op_manager.ring.is_receiving_updates(&key)`
instead of cache presence, via a small pure `interest_gate_disposition`
helper:

- **fresh** (actively receiving updates) → serve immediately;
- **unsubscribed / possibly-stale** → defer to the network so a fresh copy
  is preferred, keeping the local copy as the **existing R10 last-resort
  fallback** served only if the network search exhausts.

This reuses the existing exhaustion machinery, so worst case degrades to
**today's** behavior (the stale copy is still served as last resort) —
never a new dead-end. It also makes the relay path consistent with the
local-client path.

Minimal interim guard. Relay-caching creation and the subscribe path are
untouched.

## Testing

- Updated the two source-scrape pin tests that locked the OLD
  cache-presence gate:
  - `check_local_with_interest_gate_uses_freshness_gate` (was
    `check_local_with_interest_gate_has_interest_check`) — now pins that
    the gate calls `is_receiving_updates`, does NOT call
    `has_local_interest(`, and delegates to the disposition helper.
  - `interest_gate_disposition_gates_serve_on_freshness` (was
    `interest_gate_returns_fallback_when_not_actively_hosting`) — pins the
    fresh→serve / stale→fallback source ordering in the helper.
- New behavioral test `interest_gate_disposition_serves_fresh_defers_stale`:
  a fresh copy is served immediately (`local_value = Some`, no fallback);
  an unsubscribed copy is NOT served immediately (defers, retained only as
  the fallback). The `is_receiving_updates` freshness truth-table itself is
  already covered in `ring/hosting.rs`.
- `cargo build -p freenet`, full `operations::get::` unit suite (84
  passed), `cargo fmt`, `cargo clippy --locked -p freenet -- -D warnings`
  — all clean.

## Tradeoff

Because most cached copies are currently unsubscribed (post-0.2.92), this
guard shifts more of those GETs onto a network forward instead of an
immediate local serve — a transitional increase in GET traffic in exchange
for not serving stale state. Hosting-redesign **piece E** removes the
unsubscribed copies at the source, which removes this increase too. Filed
against the demand-driven hosting epic freenet/freenet-core#4642.

## Review follow-up (three reviews: skeptical + code-first Claude, plus Codex)

No High-severity findings; the central "worst case = today, no new
dead-end" claim was verified. Addressed:

- **Deferral intentionally does NOT register the requester at this hop.**
  Codex flagged the skipped R4 requester-registration on the deferred path
  as a P1, but the skeptical review traced it deeper: registering a
  downstream subscriber at a relay that is NOT itself in the update mesh is
  a **hollow** relay link — exactly what invariant 1 / piece D remove. The
  `subscribe` flag is still forwarded downstream (the `GetMsg::Request`
  build in the retry loop) so a **real fresh host** registers the
  subscriber. The R4 registration runs only on the fresh-serve branch,
  where it is backed by a real hosting/subscription. In the R10-exhaustion
  case, serving the stale copy without registration is also correct — no
  fresh host was found, so there is nowhere valid to register. Re-adding
  registration here would re-introduce the hollow link. Documented at the
  gate and the R4 site; **not** changed in code.

- **Composition-inversion pin (the important one).** The behavioral test
  only exercises the pure helper with hardcoded `true`/`false`, so a
  regression to `interest_gate_disposition(!is_fresh, ...)` or
  `let is_fresh = !op_manager.ring.is_receiving_updates(...)` would invert
  the gate (serve stale first, defer fresh) while all three tests still
  passed. The source-scrape pin now requires the exact un-negated literals
  and forbids both inversion forms.

- **Stricter than the local-client gate, on purpose.** The relay gate
  drops the local-client path's `connection_count == 0` / `is_locally_
  hosted` OR-terms and gates on freshness alone; documented so a future
  maintainer doesn't "reconcile" the two and re-open the stale-serve hole.

- **Originator-loopback verdict:** no NotFound-where-stale-was-served
  regression. A loopback GET only reaches the relay gate after
  `client_events.rs:933` already declined to serve locally; the deferred
  copy is retained as the R10 fallback, so if the network finds nothing the
  client still gets the local copy (today's behavior), and a fresher copy
  when one exists. Even the inconsistent-state case (store has state,
  `is_hosting_contract == false`, not receiving updates, stale
  `InterestManager.hosting == true`) only reorders WHEN the stale copy is
  served — never turns a stale-serve into a NotFound.

- Nits: fixed the now-stale success log (`active interest` → `fresh /
  receiving updates`, that branch is fresh-only now); noted the retained
  `interest_gate` names (rename deferred to piece E).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -2223,6 +2223,11 @@ fn interest_gate_disposition(
 ///
 /// Mirrors `GetMsg::Request` match arm at `get.rs:1369-1433`.
 ///
+/// Note: the `_interest_gate` / `interest_gate_disposition` names are retained
+/// to bound this interim-guard diff, but the gate is now FRESHNESS
+/// (`is_receiving_updates`), not interest/cache-presence. A rename is deferred
+/// to piece E, which removes the unsubscribed copies these names describe.
+///
 /// Returns `(local_value, local_fallback)`:
 /// - `local_value`: Some if we should serve immediately (fresh copy — actively
 ///   receiving updates).  None if we should forward.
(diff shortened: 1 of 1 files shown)


sanity:
Closing: the freshness-gate approach is a band-aid on caching, not a standalone fix. It trades a *soft* stale-read (older-but-valid data) for a *hard* delivery regression — deferring unsubscribed cached copies dead-ends single-holder large GETs (proven deterministically by `test_streaming_get_retries_after_assembly_failure` and `test_streaming_get_hosts_on_return_path`). Root cause: relay-caching's scattered copies are double-duty — the stale-serve anti-pattern AND the crutch that makes greedy routing findable. Removing them (defer) without the replacement finding mechanism breaks delivery.

The stale-read is a *symptom* of relay-caching. Piece E removes relay-caching entirely (binary hosting → a served copy is fresh by construction → nothing stale to serve), bundled with the findability robustness (bounded-backtrack + PUT seed-chain) that makes deferred requests reliably deliver. So this fix folds into the piece-E placement bundle rather than shipping alone. Tracking: #4642.

[AI-assisted - Claude]

PR #4717, opened 2026-07-06
feat(hosting): add admission decision core, subscriber-primary (#4642)

sanity:
## Problem

Piece **B** / spec step **7-bis** of the demand-driven hosting redesign (#4642) needs the **ADMISSION** half of invariant 3 in `hosting-invariants.md` — *"admission and eviction are ONE demand-vs-capacity decision, and the demand signal is SUBSCRIBER COUNT."* When a new subscription routes to a full peer, the peer must decide whether to make room by displacing a strictly-worse contract, or **refuse** rather than overcommit. No such logic exists today; the pinned (in-use) exemption in the hosting cache (`cache.rs`) is currently absolute.

This is the admission counterpart to the **subscriber-primary eviction rework (#4720)**, and is keyed off the *same* demand signal so the two are one decision by construction.

## Solution

A **pure, inert, additive** decision function plus its input-snapshot types, modeled on the reconcile controller core (#4704) and the eviction rework (#4720) — both of which landed inert first. **Nothing in production calls it yet** (`#[allow(dead_code)]`); wiring it to the live hosting cache / ring, and turning `Refuse` into the on-wire "I'm full" NACK / `EvictToAdmit` into an eviction that pierces the pinned exemption, are later steps. **No production admission/eviction path is changed.**

```
admission_decision(&AdmissionInputs) -> AdmissionOutcome
    AdmissionOutcome = Admit | EvictToAdmit(Eviction) | Refuse(RefuseReason)
```

### The demand signal is SUBSCRIBER COUNT (consistent with #4720 by construction)

The demand signal is **subscriber count** — local client subscriptions + downstream subscribers, the same genuine demand that makes a contract `contract_in_use`. **NOT** predicted read-demand and **NOT** ring distance. Both axes the earlier draft of this core used are removed: distance's causal pull on demand already flows through subscriber count via keyward routing gravity (counting both double-counts), and locality is delivered by routing, not by the admission/eviction decision (invariant 3).

Victims are ordered by the **identical** ascending key #4720's `evict_over_budget` uses: `(subscriber_count, last_get_seq, key_bytes)` — fewest subscribers first, then least-recent real GET, then a deterministic key-byte tiebreak. Because admission and eviction key off the same ordering, they pick the same victim from the same hosted set.

### Decision precedence (first match wins)

1. **Fanout over uplink headroom → `Refuse(FanoutExceedsHeadroom)`.** The one variable cost weighed at admission, never at eviction (invariant 3). Checked first — an unaffordable ongoing fanout can't be relieved by evicting bytes.
2. **Under budget → `Admit`.**
3. **Displacement evict-to-admit → `EvictToAdmit(Displacement)`.** At/over budget, displace an incumbent with **strictly fewer subscribers** than the newcomer. A fresh subscription arrives with `subscriber_count = 1`, so this sheds exactly the **zero-subscriber cached copies** that #4720's eviction under `AtCapacity` also drops. Preferred even under `Overflow` — a strictly-fewer victim relieves pressure **without** piercing the in-use pin. Bypasses the background sweep's `min_ttl` floor (spec Fable M9).
4. **OOM valve → `EvictToAdmit(OomValve)`.** Only under genuine memory overflow, and only if no strictly-fewer victim exists: shed the **fewest-subscriber** contract even though it has >= the newcomer's subscriber count — piercing the in-use pin (its subscribers re-root). The sole path that sheds a pinned subscribed contract. Same `(subscriber_count, last_get_seq, key)` ordering, over all incumbents. **The Overflow trigger is deliberately UNWIRED** (mechanism only), exactly as in #4720 — shedding a subscribed contract is the single riskiest new behavior, so its RSS/A1 trigger is a separately-validated step.
5. **Otherwise `Refuse(AtCapacityNoVictim)`** — the "node at capacity" NACK. Refuse the newcomer when every incumbent has at least as many subscribers as it (never demote a real subscriber for another, D1).

### What changed from the earlier draft of this core

- **Removed** the `predicted_demand` + `distance_to_key` axes from the decision entirely.
- **Removed** the `Zero`/`One`/`Many` `SubscriberClass` coarsening. The subscriber-primary model has no fixed "2+ protected" threshold — the rule is purely "strictly fewer subscribers than the newcomer" (a 3-sub newcomer can displace a 2-sub incumbent), and the OOM valve sheds fewest-subscriber-first across the raw counts. Coarsening would lose the information the general comparison needs and diverge from #4720's raw-`usize` ordering, breaking the consistent-by-construction guarantee. The signal is now the same raw `usize` subscriber count #4720 uses.
- Added `last_get_seq` to the hosted-contract snapshot (the recency tiebreak, identical to #4720's per-entry field — stamped only by a real GET, never by SUBSCRIBE/PUT/renewal).
- **Kept** `MemoryPressure` / `Eviction` / `RefuseReason`; renamed `EvictionKind::StrictImprovement` → `Displacement`.
- **Kept** the invariant-3 fanout-at-admission check.

**Deliberately out of scope** for this core (called out in the module docs as conscious omissions, later parts of 7-bis): the **client-pin exception** (a reserved client-pin budget slice), the **displacement-chain depth cap (D1)** + its wire counter, and **make-before-break** handoff.

## Testing

17 table-style unit tests covering every branch and boundary:

- **Under budget** (incl. empty hosted set) → `Admit`.
- **Fanout gates first** — over headroom refuses both under-budget and at-capacity (proving it precedes the byte budget); `fanout_cost == headroom` is affordable (boundary).
- **Displacement** → `EvictToAdmit(Displacement)`: zero-sub incumbent displaced for a 1-sub newcomer; a 3-sub newcomer displaces a 2-sub incumbent (proving the general strictly-fewer rule, not hardcoded to zero).
- **Refuse** — no incumbents / equal-subscriber incumbent / higher-subscriber incumbent all `Refuse(AtCapacityNoVictim)`, proving the strict `<` and the "never demote a real subscriber" rule.
- **Victim selection** — fewest-subscriber first, least-recent real GET within a subscriber class, deterministic key-byte tiebreak (order-independent).
- **OOM valve** — sheds the fewest-subscriber contract (piercing the pin) when no strictly-fewer victim exists; prefers displacement over the valve under Overflow; empty set refuses; equal-subscriber tiebreak by least-recent GET.

`cargo build -p freenet`, `cargo fmt --check`, `cargo clippy --locked -p freenet -- -D warnings`, and the 17 tests are all clean.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring/hosting.rs
+++ b/crates/core/src/ring/hosting.rs
@@ -29,6 +29,9 @@
 //! - Active subscriptions prevent eviction from the hosting cache
 //! - TTL protects recently accessed contracts from premature eviction
 
+// Admission decision core (#4642 piece B / spec "7-bis"): pure, inert, additive
+// — not yet wired to any production admission/eviction path (see the module docs).
+mod admission;
 mod cache;
 mod demand;
 mod disk_usage;
--- a/crates/core/src/ring/hosting/admission.rs
+++ b/crates/core/src/ring/hosting/admission.rs
@@ -1,0 +1,1116 @@
+//! Admission decision core — the pure decision logic for demand-driven hosting
+//! ADMISSION (#4642 piece B / spec step "7-bis"; spec "Admission & Shedding").
+//!
+//! # What this is
+//!
+//! [`admission_decision`] is a **pure function** of an [`AdmissionInputs`]
+//! snapshot: given everything the controller needs to know when a NEW
+//! subscription routes to this peer and asks it to host a contract, it returns
+//! one [`AdmissionOutcome`] — `Admit`, `EvictToAdmit(victim)`, or `Refuse`. It
+//! reads no locks, touches no live maps, and performs no wire I/O; the caller
+//! materializes the snapshot once, so the logic here is directly unit-testable
+//! and any at-emission re-read (a hardening concern) stays a separable layer.
+//!
+//! This is the admission half of invariant 3 in `hosting-invariants.md`:
+//! **admission and eviction are ONE demand-vs-capacity decision, and the demand
+//! signal is SUBSCRIBER COUNT.** A peer retains the contracts with the most
+//! **subscribers** (local client subscriptions + downstream subscribers — the
+//! same genuine demand that makes a contract `contract_in_use`) that fit a
+//! capability-relative budget; when a new subscription arrives at a full peer it
+//! either makes room by displacing a **strictly-fewer-subscriber** incumbent or
+//! refuses, never overcommitting and never demoting a real subscriber for
+//! another.
+//!
+//! # Relationship to eviction (#4720) — one shared ranking
+//!
+//! Admission and the background over-budget eviction sweep are ONE demand-vs-
+//! capacity decision (invariant 3), so they MUST rank victims the same way. They
+//! do, by construction: both go through the single canonical
+//! [`cache::victim_order`], which orders victims ascending
+//! `(subscriber_count, last_get_seq, key_bytes)`. [`pick_displacement_victim`]
+//! and [`pick_oom_valve_victim`] call it directly — there is no second copy of
+//! the ordering to drift out of sync. The `admission_and_cache_agree_*` pin tests
+//! in this module assert admission's victim selection and `cache`'s over-budget
+//! eviction pick the SAME victim (and the same full eviction order) from a shared
+//! fixture, so a change to either side that broke the agreement fails CI.
+//!
+//! The subscriber signal is a raw `usize` count (local client subs + downstream
+//! subs), NOT a distance-derived read-demand estimate and NOT ring distance. Both
+//! axes the earlier draft of this core used (predicted demand and distance) are
+//! gone: distance's causal pull on demand already flows through subscriber count
+//! via keyward routing gravity (counting both double-counts), and locality is
+//! delivered by routing, not by the admission/eviction decision (invariant 3).
+//!
+//! The `last_get_seq` recency tiebreak means the same thing on both sides: it is
+//! **real-GET-only** — stamped by a real GET, never by SUBSCRIBE, PUT, or
+//! subscription renewal — in both `cache.rs` (see `HostedContract::last_get_seq`)
+//! and here, so the shared tiebreak is consistent.
+//!
+//! # The decision (spec "Admission & Shedding"; task 7-bis)
+//!
+//! 1. **Fanout affordability (uplink).** The one variable cost weighed at
+//!    ADMISSION, never at eviction (invariant 3): if the new subscription's
+//!    ongoing per-update fanout exceeds the peer's uplink headroom, `Refuse` —
+//!    the peer cannot afford to keep this contract updated regardless of bytes.
+//!    Checked first: an unaffordable ongoing fanout can't be relieved by evicting
+//!    bytes, so no eviction can rescue it.
+//! 2. **Under budget → `Admit`.** Spare capacity, no eviction needed.
+//! 3. **Displacement (strictly-fewer-subscriber evict-to-admit).** At/over
+//!    budget, admit by displacing an incumbent with **strictly fewer subscribers**
+//!    than the newcomer. For a fresh subscription (newcomer count 1) that means a
+//!    zero-subscriber cached copy is displaced for the incoming one-subscriber
+//!    contract. Among eligible victims, pick the fewest-subscriber, then
+//!    least-recent real GET (`last_get_seq`), then a key-byte tiebreak — the
+//!    canonical `cache::victim_order` ordering shared with the eviction sweep.
+//!    Preferred even under `Overflow`: displacing a strictly-fewer-subscriber
+//!    incumbent both relieves pressure and never pierces the in-use pin.
+//! 4. **OOM valve.** ONLY under genuine memory overflow, if no strictly-fewer
+//!    incumbent exists, shed the **fewest-subscriber** contract even though it has
+//!    at least as many subscribers as the newcomer — piercing the in-use pin (its
+//!    subscribers re-root). This is the sole path that sheds a pinned subscribed
+//!    contract, and only at real OOM risk. Same canonical `cache::victim_order`,
+//!    over ALL incumbents.
+//! 5. **Otherwise `Refuse`.** At capacity and every incumbent has at least as many
+//!    subscribers as the newcomer (nothing strictly-fewer to displace, and not
+//!    genuine OOM) → the "node at capacity" NACK. Refuse the NEWCOMER (visible to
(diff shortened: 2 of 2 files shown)


sanity:
## Consolidated review — admission decision core

Reviewers: external **Codex** (gpt-5.5) + a **Claude adversarial lens**. The pure decision logic is correct and well-tested; the findings are about the **"agrees with #4720 by construction"** claim, which isn't enforced and is currently false because #4720 hasn't landed in this branch.

### Verified correct (no defect)
- **OOM-valve labeling:** `OomValve` is reachable only after the strict-`<` displacement filter returns `None`, so any valve victim necessarily has `subscriber_count >= candidate` — a genuine pin-pierce. `oom_valve_evictions_total` cannot be mislabeled by this logic.
- **First-match-wins logic:** strict `<` never demotes an equal-count incumbent; fanout strict `>` makes `cost == headroom` affordable; no `usize` underflow (counts only compared, never subtracted); empty-set handled in all three pressure states; deterministic key-byte tiebreak.
- **Inertness:** commit touches only `mod admission;` (+3 in `hosting.rs`) and `admission.rs` (+860); `#![allow(dead_code)]` present; no production caller (`migration_admission_decision` in `node.rs` is pre-existing and unrelated).

### Findings
1. **[Major — design integrity] The "identical to #4720" ranking is duplicated and currently divergent.** `admission.rs` documents `victim_order` = `(subscriber_count, last_get_seq, key_bytes)` "identical to #4720," but #4720 isn't in this branch: `cache.rs::evict_over_budget` still sorts `(keep_score, last_access_seq, key_bytes)`, and `subscriber_count`/`last_get_seq`/`MemoryPressure` appear nowhere in `cache.rs`. Nothing binds the two rankings, so a wiring step could make admission and eviction pick different victims — the exact disagreement invariant 3 forbids. **Fix: sequence #4720 first, stack this on it, and extract ONE shared `victim_order` (owned by `cache.rs` / a `hosting::ranking` submodule) imported by both, plus a cross-module pin test asserting both select the same victim from a shared fixture.**
2. **[Major] Recency semantics differ.** admission's `last_get_seq` is spec'd real-GET-only; `cache.rs`'s `last_access_seq` bumps on every access (insert/read/seed/touch/SUBSCRIBE/PUT). #4720 must introduce the real-GET-only seq; today it's an unbuilt dependency the input-builder needs.
3. **[Major — docs] Docstrings assert agreement with types that don't exist yet** (`cache::MemoryPressure`, the shared ordering) as established fact. Reword to an honest forward-reference until the shared fn + pin test exist.
4. **[Minor] Single-victim eviction may not free enough bytes when wired** (Codex P2). The API carries no sizes / no multi-victim outcome, while `evict_over_budget` loops until under budget. The demand decision (one displacement victim) + the continuing byte-sweep compose to handle it (the admitted ≥1-sub newcomer outranks zero-sub copies, so the sweep evicts further copies, never the newcomer; the OOM valve covers the all-pinned case), but the driver contract must not assume one eviction suffices — document it.
5. **[Nit] Two `HostedContract` structs, same name** (admission vs cache) under `ring::hosting` with disjoint fields — rename admission's (e.g. `AdmissionIncumbent`); a shared-ranking refactor reconciles them anyway.
6. **[Nit] `candidate.subscriber_count` is an unconstrained `usize`** while the model assumes ≥1. Add `debug_assert!(candidate.subscriber_count >= 1)`.

**Coverage gaps:** cross-module consistency test (the crux of "by construction" — can't exist until #4720 lands, which is itself the sequencing argument); OomValve key-byte-tiebreak determinism; fanout-gate under `Overflow`; a 0-count candidate case.

### Merge plan
Do **not** merge this independently ahead of #4720. Order: **#4720 (subscriber-primary eviction) → this, stacked on it, sharing one `victim_order` + a cross-module pin test.** Self-contained improvements (findings 3, 5, 6 + the missing tests) applied now; the shared-fn (1, 2) lands with the stack once #4720 is signed off. Both remain gated on Ian's sign-off + R2 field validation before the R3 bundle ships.

[AI-assisted - Claude]

sanity:
## Rebase onto main + review (post-#4720 squash-merge)

#4720 squash-merged to `main` (`baa03f0d`), so `cache::victim_order` and subscriber-primary eviction are now on main. Rebased this branch onto main with `git rebase --onto origin/main <top-of-4720>`, dropping #4720's now-in-main commits and replaying only #4717's own commit.

**Result:** new HEAD `6122d568`. Diff vs `main` is exactly the intended change — `crates/core/src/ring/hosting/admission.rs` (new, inert/additive) + the 3-line `mod admission;` block in `hosting.rs`. Verified byte-identical to the pre-rebase reviewed content (the rebase changed history, not code content).

**Import resolves against main:** `admission.rs` uses `super::cache::victim_order(...)`, which is main's `pub(crate) fn victim_order` (cache.rs:569, from #4720). Signature matches the call sites exactly.

**Local verification (worktree at `6122d568`):**
- `cargo build -p freenet` — clean
- `cargo clippy --locked -p freenet -- -D warnings` — clean
- `cargo fmt --check` — clean
- `cargo test -p freenet --lib ring::hosting::admission` — 22/22 pass, incl. the two cross-module consistency tests (`admission_and_cache_agree_on_at_capacity_victim`, `admission_and_cache_agree_on_full_eviction_order`) exercising main's real `cache.rs`.

### Review

**External (Codex, `codex review --base main`):** 1 finding, no P0/P1.

- **[P2] Account for bytes before evict-to-admit** (admission.rs:376-380): when wired, a single strictly-fewer-subscriber victim may not free enough bytes for a larger candidate, leaving the cache over budget.

  **Disposition: not applicable to this core (design-intentional).** The hosting redesign deliberately rejects byte-precise budgeting — invariant 3 + the anti-patterns table treat "byte-only eviction as the sole signal" as an anti-pattern, and profiling found per-contract cost is dominated by ~1MB resident memory + update fanout with state bytes (~650KB) negligible/uncorrelated. The decision is one-contract-in/one-contract-out against a capability-relative budget, so one victim per admission is the correct unit. The pure core intentionally does no byte math — the "does the newcomer fit / RSS near OOM" arithmetic lives in the input-builder that resolves `MemoryPressure` (per the module docs). Bulk over-budget drain is handled separately by the background `cache::evict_over_budget` sweep (#4720). Codex frames it as a "when wired" note; it's a checklist item for the future builder wiring, not a defect in this inert decision function.

**Claude adversarial read:** decision precedence verified against invariant 3 (fanout strict-`>` first; UnderBudget→Admit; strictly-fewer displacement preferred even under Overflow so the in-use pin isn't pierced; OOM valve only under Overflow; else Refuse). Both victim pickers reuse `cache::victim_order`, so admission and eviction rank identically by construction (pinned by the agreement tests). `#![allow(dead_code)]` correct for the inert module. No issues.

[AI-assisted - Claude]

PR #4720, opened 2026-07-06
feat(hosting): subscriber-primary eviction, retention, and OOM valve (#4642)

sanity:
## Problem

Piece A of the demand-driven hosting redesign (#4642) shipped a **demand-ordered** (Greedy-Dual + proximity-prior) eviction policy in `ring/hosting/cache.rs`. Ian has since decided the retention/eviction/admission demand signal should be **subscriber count**, not a distance-derived read-demand estimate:

- **Subscriber count** (local client subscriptions + downstream subscribers — the same genuine demand that makes a contract `contract_in_use`) is the primary signal, so retention agrees with the piece-D collapse/renewal decision by construction.
- **Distance is redundant** in the eviction ranking: its causal pull on demand already flows through subscriber count (subscriptions route toward the key → near-key peers accumulate more subscribers), and network locality is delivered by **routing**, not eviction. Counting both double-counts.
- The shipped hard-pin has **no OOM valve**: subscribed contracts are pinned even when the subscribed set itself won't fit RAM.

## Approach (this PR = eviction/retention half; admission is PR-2)

**Eviction is now subscriber-primary.** `evict_over_budget` orders victims ascending `(subscriber_count, last_get_seq, key_bytes)`:

- **Primary: subscriber count** (fewest first), supplied by a `HostingManager` closure (`subscriber_count`, which replaced the old `should_retain` bool closure). It counts client subscriptions + downstream subscribers, so `subscriber_count(k) == 0` iff `!contract_in_use(k)` — the pin is identical to the shipped one by construction.
- **Secondary: real-GET recency** (`last_get_seq`, least-recent real GET first). A NEW per-entry field stamped **only** by `AccessType::Get` and by `touch_with_demand` (local-cache GET serve) — NOT by SUBSCRIBE, PUT, subscription renewal, or reload. This keeps renewal traffic from making every subscribed contract look equally fresh.
- **Tertiary:** deterministic key-byte tiebreak.

**Normal operation is the pin-in-normal-op model** (as designed): under `MemoryPressure::AtCapacity` (the production value) subscribed contracts are pinned and only zero-subscriber contracts are evicted, so among candidates subscriber_count is always 0 and the order reduces to real-GET recency. `min_ttl` is preserved as the **cold-start grace floor** — a fresh zero-subscriber PUT/GET survives it before becoming eviction-eligible.

**OOM valve (mechanism only — trigger deliberately UNWIRED).** Under `MemoryPressure::Overflow` the sweep pierces the in-use pin and `min_ttl` and sheds the fewest-subscriber contract (then least-recent-GET) to survive genuine RAM overflow, incrementing `oom_valve_evictions_total`. **Nothing constructs `Overflow` in production** — the real RSS/A1 resource trigger is a deliberately separate, separately-validated step, because shedding a *subscribed* contract is the single riskiest new behavior here. The `Overflow` variant carries `#[cfg_attr(not(test), allow(dead_code))]` documenting this. Telemetry for the valve counter is plumbed now (RouterSnapshot → OTLP) so a nonzero rate is visible the moment the trigger lands.

**`demand.rs` is DEMOTED, not deleted.** `keep_score`, `eviction_floor`, `predicted_demand`, and `record_abandonment` are retained but no longer drive eviction — they remain the dashboard/telemetry surface (`predicted_demand` still trained and displayed) for continuity this release, scheduled for deletion once subscriber-primary is field-validated.

## Testing

New unit tests (subscriber-primary): pin-subscribed-drop-zero-sub, real-GET refreshes recency while SUBSCRIBE/PUT do not, OOM valve sheds fewest-subscriber (piercing the pin), OOM valve equal-subscriber tiebreak by least-recent-GET, valve counter stays 0 under normal AtCapacity, cold-start grace (fresh unread seed survives min_ttl), and (review fix) a reloaded PUT/SUBSCRIBE entry keeps `last_get_seq == 0` so restart can't fabricate GET recency. Plus a telemetry pin test for the valve gauge.

Superseded demand-ordering tests (`fuel_gauge_evicts_lowest_demand_not_lru`, `fuel_gauge_cold_start_keeps_near_key_over_far_key`, `fuel_gauge_keeps_repeatedly_read_contract_over_junk`, the two abandonment-priority tests) are `#[ignore]`d with comments referencing this rework as historical documentation of the demoted machinery. `test_access_refreshes_lru_position` was rewritten to use a real GET.

`cargo build`, `cargo clippy --locked -p freenet -- -D warnings` (CI-style, lib), `cargo fmt`, and the `ring::hosting` + telemetry tests are green.

## Approved-issue loop

This is a behavior change, authorized by the maintainer-approved hosting-redesign epic **#4642** (design converged with Ian; the decision is codified in `docs/design/demand-driven-hosting.md` §9 Piece A). This PR updates that design doc and adds a SUPERSEDED banner to `docs/design/hosting-eviction.md` so the two in-repo docs stop disagreeing. The load-bearing invariant (`.claude/rules/hosting-invariants.md`, invariant 3) lives outside this repo and is updated out-of-band; it is intentionally **not** in this diff.

## Risk / review

This changes **LIVE, field-deployed eviction behavior** (shipped piece A). It requires **full multi-model review + Ian sign-off + field validation** before it ships.

Follow-ups: **PR-2** reworks the (unmerged, inert #4717) admission core to subscriber-primary; the OOM-valve RSS/A1 trigger and the eventual `demand.rs` deletion are later, separately-validated steps. When the trigger is wired it OWES (documented on `MemoryPressure::Overflow`): rank the valve by the lease-filtered downstream count, and clean up the shed contract's subscription state so it doesn't immediately look in-use again.

Refs #4642

[AI-assisted - Claude]


sanity:
## R2 completeness flag — op-scoped pinning

The reconciled R2 eviction-chain spec (`hosting-invariants.md`, piece A / R1+R2 track) is "wire subscriber-count into the keep-ranking **plus op-scoped pinning**." This PR delivers the ranking. I don't see the op-scoped-pinning half, and want to flag it for your review rather than let it slip:

- On `origin/main` the eviction pin is `contract_in_use = has_client_subscriptions || has_downstream_subscribers` (`hosting.rs:1107-1108`); the sweep retains `past_min_ttl && !contract_in_use` (`hosting.rs:1515`).
- There is **no in-flight-operation exemption** — grep for `op_scoped` / `in_flight_op` / `active_operation` / `op_pin` across `ring/hosting/{hosting,cache,admission,reconcile}.rs` returns 0. The only protection for a contract with an active op but no subscription yet is `min_ttl` recency.

**The question to resolve:** can a contract be evicted mid-operation? A long multi-fragment streaming GET (the #4345 class) or a stalled op under memory pressure could, if the contract is in the hosting cache and its `last_accessed` ages past `min_ttl` during the op, become eviction-eligible while the op still needs it.
- If the contract only enters the eviction-eligible hosting cache on op *completion* (finalize-host), or `min_ttl` always covers op duration → `min_ttl` already satisfies "op-scoped pinning" and there's nothing to build.
- If a contract can sit in the cache and lapse `min_ttl` mid-op → an explicit op-scoped pin (exempt a contract with an in-flight op, independent of `min_ttl`) is the missing R2 piece.

Not building it here — it's a live-eviction-path change that needs the necessity call + your sign-off, and it's cleaner as a scoped follow-up once you've decided. Flagging so R2 ships complete relative to its own spec.

[AI-assisted - Claude]

sanity:
## Op-scoped pinning — resolved, NOT needed for R2

Investigated the "can a contract be evicted mid-operation?" question. Answer: **no**, on any path this PR changes — so op-scoped pinning is optional future hardening, not an R2 prerequisite. This PR is complete relative to the R2 spec.

Evidence:
- **GETs don't host** (`host_get` has zero production callers; GET-auto-subscribe was removed in #4664), so a multi-fragment streaming GET (the #4345 class) has no in-flight cache entry to evict.
- **PUT inserts post-commit** (`put/op_ctx_task.rs:~1918`, inside `if !was_hosting`, after the state write) with `last_accessed = now`, so the fresh entry is `min_ttl`-protected (8 min = `SUBSCRIPTION_RENEWAL_INTERVAL 120s × 4`). No PUT stays in flight 8 min.
- **The eligibility gate blocks it:** the only wired pressure, `AtCapacity`, requires `age >= min_ttl && subs == 0`. `Overflow` pierces both but its trigger is deliberately unwired this release.
- **This PR doesn't widen any eviction edge** — it's eligibility-preserving (`subscriber_count == 0 ⟺ !contract_in_use` by construction), only reordering victims within the already-eligible zero-subscriber set.

The one latent edge (a >8-min-old, zero-subscriber contract being served for a slow GET, swept under concurrent memory pressure) is **pre-existing** (identical eligibility in the old code), narrow, and partially guarded by the eviction-time `expected_generation` re-check. Not introduced or widened by R2, so not a blocker; if we ever want defense-in-depth, an in-flight-op refcount OR'd into the `AtCapacity` filter is the minimal design.

[AI-assisted - Claude]

PR #4722, opened 2026-07-07
feat(hosting): advertisement-layer reliability + retraction (Fix 1)

sanity:
## Problem

The hosting **advertisement layer** (`NeighborHostingManager`, Source-1 — distinct from the InterestSync *state* anti-entropy) has two reliability gaps that leave co-hosts with **stale advertisements**, violating invariant 1 ("advertise iff a fresh in-mesh host"):

1. **On-connect exchange is best-effort with no recovery.** The full-set advertised-contract exchange (`HostingStateRequest`/`HostingStateResponse`) is a `try_send` and — unlike InterestSync's 5-min heartbeat — has **no periodic re-request**. A dropped exchange leaves two co-hosts permanently unaware of what each other hosts (lost Source-1 fan-out, missed terminal read consult).
2. **Retraction on stop-hosting was dead code.** `on_contract_unhosted` had **no production caller**, so eviction left `my_contracts` (the advertised set) permanently stale — a co-host keeps a **phantom host** for a contract we evicted, fans updates to us and routes reads to us for state we no longer hold. Same stale-advertisement failure class behind #4440/#4473/#4610.

Spec **step 1 ("Fix 1")** of the demand-driven hosting redesign (epic #4642).

## Solution

Scoped **strictly to the advertisement (ID) layer**. The InterestSync STATE anti-entropy is untouched, and nothing here fetches or summarizes state, so the #4440/#4473 summarize storm cannot be re-armed. No wire-format change (reuses existing `NeighborHostingMessage` + `NodeEvent::{SendNetMessage, BroadcastHostingUpdate}`).

**Reliability — periodic full-set re-request, piggybacked on `interest_heartbeat`** (NOT a separate task). Each ~5-min cycle sends each connected neighbor a `HostingStateRequest`; `HostingStateResponse` **full-replaces** our view of the peer, so BOTH a dropped on-connect exchange AND a dropped retraction self-heal within one interval (both directions, since every node runs it). Piggybacking is deliberate: a separate task's `GlobalRng` initial-delay draw perturbs the shared seeded RNG stream the deterministic sim harness depends on — riding the existing loop adds no RNG consumer and no second timer. The re-request is **decoupled from interest state** (runs for any connected node — even a pure relay consults its `neighbor_contracts` view) and uses **`try_send`** (best-effort; a dropped cycle heals next interval and must never block the event-loop notification channel, per channel-safety). `HostingStateResponse` computes overlap (which drives the proactive state sync) only over **newly-learned** contracts, so steady-state re-requests do zero redundant state sync while a fresh/reconnected view still recovers.

**Retraction — on the confirmed-delete path.** `RuntimePool::remove_contract` now calls `announce_contract_unhosted` once a required on-disk half is actually gone — `ReclaimOutcome::Full` (both state + WASM removed) OR `Partial` (one removed, the other queued for retry, so the contract can no longer be served); `Err` (nothing removed → still serveable) keeps advertising. It runs **after** the three deletion-time guards (re-host / re-subscribe / newer-generation), so a contract re-hosted in the eviction→reclaim window keeps both its state AND its advertisement. It is **not** wired at the eviction *decision* in `reclaim_evicted_contract` (which only gates on `contract_in_use` and can't see a concurrent re-cache) nor at **collapse** (`send_unsubscribe_upstream` doesn't stop hosting on current `main`). Every eviction funnels through `remove_contract` (GET/PUT insert eviction, the periodic sweep, and — once built — evict-to-admit). Idempotent, so a pending-reclamation retry (incl. a Full retry after a Partial) is a safe no-op.

## Testing

- **Behavioral / integration** (`pool.rs::tests::test_eviction_emits_hosting_retraction`): builds a **real `OpManager` + `RuntimePool`**, stores a contract (state + WASM) on disk, advertises it, then drives the real `RuntimePool::remove_contract` path and asserts BOTH (a) the on-disk state is gone AND the advertisement is retracted, AND (b) `GlobalTestMetrics::neighbor_hosting_retractions()` advanced by exactly one. This is the counter's real reader and proves the eviction path reaches the retraction at runtime (verified load-bearing: commenting out the retraction fails the test). Runs on a `current_thread` runtime because the counter is thread-local and the retraction is synchronous. (The `SimNetwork` handlers use the direct inner `Executor` with no `op_manager`, so a sim can't reach this path — hence the direct pool test.)
- **Unit** (`neighbor_hosting.rs`): removed-announce convergence; full-set re-request drops a stale (dropped-retraction) advertisement; overlap-only-for-newly-learned (+ reconnection recovery); `on_contract_unhosted` idempotency.
- **Source-scrape pins**: the re-request is piggybacked on `interest_heartbeat`, carries only IDs (never state), and there is no separate hosting task; the retraction is in `RuntimePool::remove_contract` gated `Full | Partial` (after the guards), and `reclaim_evicted_contract` does NOT retract.
- **Sim** (unchanged behavior): terminal-advertisement-consult (GET + SUBSCRIBE) dead-end, get-routing-coverage, subscription-chain collapse, direct-runner determinism, neighbor-hosting-bounded — all green.

`cargo build -p freenet`, `cargo clippy --locked -p freenet -- -D warnings`, `cargo fmt --check`, and all new + touched tests are green.

## Review notes (high-risk surface)

- Advertisement layer only; InterestSync state anti-entropy untouched; no WASM summarize / state fetch on the new path (does not re-arm #4440/#4473).
- No wire-format change.
- One production visibility widening: `OpManager::new` `pub(super)` → `pub(crate)` (constructor-only, no behavior change) so the in-crate `RuntimePool` eviction→retraction test can stand up a real `OpManager`.
- Interaction with the reconcile-core keystone (step 2): the shadow controller's `Retract` action maps to this same `on_contract_unhosted` primitive; wiring it here makes `is_hosted_locally` non-monotonic in production (reconcile-shadow docs updated accordingly).
- Follow-ups noted in review (not in this PR): uncapped full-set payload + per-peer request rate-limit (both bounded today, pre-existing shape).

Refs #4642 (spec step 1 / "Fix 1"; `hosting-invariants` invariant 1).

https://claude.ai/code/session_01F6iqkKTXDfMCi3nkShGCTi

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3351,6 +3351,9 @@ std::thread_local! {
     static GLOBAL_PENDING_OP_REMOVES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PENDING_OP_HWM: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_NEIGHBOR_HOSTING_UPDATES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // Hosting advertisement retractions emitted on stop-hosting (eviction).
+    // Advertisement-layer reliability + retraction, #4642 spec step 1.
+    static GLOBAL_NEIGHBOR_HOSTING_RETRACTIONS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_ANTI_STARVATION_TRIGGERS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     // Terminal advertisement consult (hosting redesign piece C, invariant 5).
     static GLOBAL_TERMINAL_CONSULT_ATTEMPTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
@@ -3390,6 +3393,7 @@ impl GlobalTestMetrics {
         GLOBAL_PENDING_OP_REMOVES.with(|c| c.set(0));
         GLOBAL_PENDING_OP_HWM.with(|c| c.set(0));
         GLOBAL_NEIGHBOR_HOSTING_UPDATES.with(|c| c.set(0));
+        GLOBAL_NEIGHBOR_HOSTING_RETRACTIONS.with(|c| c.set(0));
         GLOBAL_ANTI_STARVATION_TRIGGERS.with(|c| c.set(0));
         GLOBAL_TERMINAL_CONSULT_ATTEMPTS.with(|c| c.set(0));
         GLOBAL_TERMINAL_CONSULT_HITS.with(|c| c.set(0));
@@ -3461,6 +3465,17 @@ impl GlobalTestMetrics {
 
     pub fn neighbor_hosting_updates() -> u64 {
         GLOBAL_NEIGHBOR_HOSTING_UPDATES.with(|c| c.get())
+    }
+
+    /// A hosting advertisement retraction was emitted because this node stopped
+    /// hosting a contract (eviction). Advertisement-layer reliability +
+    /// retraction, #4642 spec step 1 — see `NeighborHostingManager::on_contract_unhosted`.
+    pub fn record_neighbor_hosting_retraction() {
+        GLOBAL_NEIGHBOR_HOSTING_RETRACTIONS.with(|c| c.set(c.get() + 1));
+    }
+
+    pub fn neighbor_hosting_retractions() -> u64 {
+        GLOBAL_NEIGHBOR_HOSTING_RETRACTIONS.with(|c| c.get())
     }
 
     pub fn record_anti_starvation_trigger() {
--- a/crates/core/src/contract/executor/runtime/pool.rs
+++ b/crates/core/src/contract/executor/runtime/pool.rs
@@ -948,6 +948,24 @@ impl ContractExecutor for RuntimePool {
                     .pending_reclamation_add(*key, current_gen);
             }
         }
+
+        // Retract our hosting advertisement once ANY required on-disk half (state
+        // or WASM code) is gone — i.e. on `Full` (both gone) OR `Partial` (one gone,
+        // the other queued for retry): in either case we can no longer reliably
+        // serve `key`, so we must stop advertising it (Fix 1, #4642 spec step 1;
+        // invariant 1: advertise iff a fresh in-mesh host). `Err` (BOTH deletes
+        // failed → nothing removed → still fully present) keeps the advertisement
+        // and retries. Wired HERE on the confirmed-delete path, NOT at the eviction
+        // decision, so the three guards above that DELIBERATELY skip the delete
+        // (contract re-hosted / re-subscribed / written to a newer generation in
+        // the eviction→reclaim window) also skip the retraction — a contract kept
+        // alive by that race retains BOTH its state AND its advertisement.
+        // Best-effort (`try_notify_node_event`), healed by the periodic re-request
+        // on drop; idempotent, so a later Full retry after a Partial (or any repeat)
+        // re-retracts harmlessly.
+        if matches!(&result, Ok(ReclaimOutcome::Full | ReclaimOutcome::Partial)) {
+            crate::operations::announce_contract_unhosted(self.op_manager.as_ref(), key);
+        }
         // Drop the outcome detail at the trait boundary: callers expect
         // `Result<(), ExecutorError>` and cannot act on Full vs Partial.
         result.map(|_| ())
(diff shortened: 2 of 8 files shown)


sanity:
## External review (Codex) — recorded for the FULL review gate

Ran `codex review --base origin/main` (the required non-Claude pass for this high-risk hosting surface).

**Round 1 (P1, addressed):** Codex flagged that emitting the retraction at the eviction *decision* in `reclaim_evicted_contract` was premature — `RuntimePool::remove_contract` deliberately skips the delete when a contract is re-hosted / re-subscribed / written to a newer generation in the eviction→reclaim window, so a decision-site retraction could leave the node holding state it no longer advertises. Confirmed real: `reclaim_evicted_contract` only gates on `contract_in_use`, not `is_hosting_contract`, so a concurrent re-cache would be wrongly retracted there; only `remove_contract` (serialized per-key with PUT, checking all three guards) can decide correctly.

**Fix:** moved the retraction to `RuntimePool::remove_contract`'s confirmed `ReclaimOutcome::Full` delete arm (after the guards). Pinned by `remove_contract_retracts_only_on_confirmed_full_delete` + `reclaim_evicted_contract_does_not_retract_at_the_decision_site`.

**Round 2 (clean):** re-ran after the fix — "did not find a discrete regression introduced by the patch."

### Test-coverage note for reviewers
The retraction now fires on the async background `EvictContract` reclamation path, which the current `run_controlled_simulation` harness does not reliably drive to completion in-window (I confirmed the counter stays 0 even with `use_mock_wasm=true`, because the background reclamation doesn't complete before the sim ends). So end-to-end eviction→retraction is covered by unit tests (message semantics: removed-announce convergence, full-set re-request heals a stale advert, idempotency) + source-scrape pins (wiring in the Full arm), rather than a sim. A harness extension to force background-reclamation completion would let a future sim assert the counter directly.

This is a **draft for the FULL multi-model review + Ian sign-off** per the hosting high-risk-surface policy — not requesting merge.

[AI-assisted - Claude]

sanity:
## Update — sim regression found by CI and fixed; Codex P2 addressed

The first pushed version broke the Simulation CI job (5 tests). Root-caused with a dedicated debugging pass + decisive isolation experiments (and corroborated against origin/main, which is consistently green):

**Regression:** the separate `Ring::hosting_heartbeat` background task's startup `GlobalRng::random_range(15..=45)` initial-delay draw added a consumer to the shared per-thread seeded RNG stream that the deterministic simulation harness relies on. This (a) shifted seed-tuned tests' routing/topology (the 4 consult/routing/subscription tests, which never even reach the heartbeat's ~315s first-fire in their ~180-240s windows) and (b) destabilized cross-run RNG-consumption ordering (`test_direct_runner_determinism` failed only on `event_keys` — same event shape, different entity identities across runs, the RNG-perturbation signature). Isolation: disabling only the spawn made all 5 pass; keeping the task but replacing the draw with a constant made determinism pass.

**Fix:** removed the separate task entirely and PIGGYBACKED the `HostingStateRequest` re-request onto the existing `interest_heartbeat` loop — no new RNG draw, no second timer, so the sim RNG stream is byte-identical to before. Verified locally: the 5 previously-failing tests + `test_neighbor_hosting_bounded` all pass (determinism green).

**Codex P2 (decouple from interest hashes) — addressed:** Codex flagged that riding `interest_heartbeat`'s `if hashes.is_empty() { continue; }` skips the re-request for an advertising-but-not-interested node. Confirmed real: startup seeds `neighbor_hosting.my_contracts` from disk (`handler.rs`) but not the interest index, so a just-restarted node advertises hosting with empty `hashes` until contracts are re-accessed. Fixed by decoupling: the cycle now runs when we have interests to sync OR advertise hosting (`local_hosted_count() > 0`); the interest message is sent only when `hashes` is non-empty, the hosting re-request always. Re-verified determinism + behavioral tests stay green after the decoupling.

### Review trail (per the per-code-content review rule)
- Codex round 1 (retraction placement P1): fixed — retraction moved to `RuntimePool::remove_contract`'s confirmed-`Full` arm.
- Codex round 2: clean.
- Codex round 3 (piggyback P2, decouple from interests): fixed as above.
- A final Codex pass on the decoupled HEAD is running; I'll note the result.

Current HEAD `f7cfe09d`. Build / clippy `-D warnings` / fmt / unit + pin / touched sim tests all green locally; CI re-running.

[AI-assisted - Claude]

sanity:
## Update 2 — two further Codex P2s addressed; final state

Two additional Codex findings on the reliability backstop, both fixed:

- **P2 — pure-relay coverage:** the piggybacked re-request must reconcile `neighbor_contracts` for ANY connected node (terminal GET/SUBSCRIBE consult + Source-1 UPDATE fan-out read that view even on a node hosting nothing), else a dropped retraction leaves a phantom advertised host unhealed. Fixed: the advertisement re-request now runs whenever we have connected peers (the InterestSync interest message stays gated on having hashes; the hosting re-request does not). Re-verified determinism + the 4 seed-tuned behavioral tests + neighbor-hosting-bounded all still green.

- **P2 — partial reclamation:** the retraction fired only on `ReclaimOutcome::Full`. A `Partial` reclaim (one of state/WASM deleted, the other transiently failed) can no longer serve the contract but was still advertised until a retry succeeded. Fixed: retract on `Full | Partial` (any required half gone); `Err` (both deletes failed → nothing removed → still serveable) keeps advertising. Pin updated (`remove_contract_retracts_on_any_confirmed_delete_not_on_err`).

### Codex trail (per-code-content review)
1. P1 retraction placement → fixed (confirmed-delete path).
2. clean.
3. P2 piggyback/interest-gating → fixed (decouple).
4. P2 pure-relay + P2 partial-reclaim → fixed (this update).

Current HEAD `176ae2c2`. Build / clippy `-D warnings` / fmt / 27 unit+pin / the touched sim tests all green locally; CI re-running. A final Codex pass on this HEAD is running.

Still a **draft for the FULL multi-model review + Ian sign-off** — not requesting merge.

[AI-assisted - Claude]

sanity:
Codex pass 5 (HEAD `176ae2c2`): **clean** — "I did not identify any discrete correctness, reliability, security, or maintainability issue introduced by this diff. The retraction and periodic hosting-state reconciliation paths appear consistently wired and covered by targeted tests/source pins." Monitoring CI (Simulation) for green. Still a draft for the FULL review + Ian sign-off. [AI-assisted - Claude]

sanity:
## Consolidated Full-tier review — advertisement-layer reliability + retraction

High-risk surface (this layer caused the #4440/#4473/#4610 summarize storms), so this ran **Full tier**: external **Codex** (5 passes during the build, P1 + three P2s addressed) plus **three perspective-diverse Claude lenses** — false-retraction correctness, storm-safety/scaling, and determinism/concurrency. **Verdict: sound. No blockers, no majors.**

### Verified clean (the parts that matter on this surface)
- **Cannot de-host a live contract** (false-retraction lens). `RuntimePool::remove_contract` has one caller (`EvictContract`) behind three guards (`is_hosting_contract`, `contract_in_use`, generation-mismatch) that all return before the reclaim; delete + retraction run atomically in the serial `contract_handling` loop; any concurrent re-host's ADD supersedes the REMOVE via the per-key fair queue. `Partial` correctly retracts (binary hosting: a missing half is un-servable). Collapse correctly does **not** retract (keeps a fresh cached copy). Invariant 1 holds.
- **No summarize-storm re-arm** (storm lens). No `summarize_state`/`get_contract_summary` reachable from the new path; the "newly-learned contracts only" overlap is correctly implemented (zero serial-loop work in steady state); sends are spread across the 5-min interval, not bursty.
- **Determinism + concurrency** (determinism lens). The piggyback fix genuinely removes the RNG perturbation (no new draw/task/wall-clock in the diff); send order deterministic; retraction is deadlock-free (locks released before the send, executor returned before the retraction, non-blocking `try_notify`).

### Fold now (both minor, being applied in this PR)
1. **`.send().await` → `try_send`** on the periodic re-request (`ring.rs:1479-1491`). It blocks on the cap-2048 `event_loop_notification` channel — the exact channel behind the #4145/#4231/#4442 wedges — for an explicitly best-effort "heals next cycle" message. Not a wedge risk here (background task, spread sends, no lock held), but it's the `.send().await`-in-loop anti-pattern and inconsistent with the sibling retraction path, which already uses `try_send`. Dropping a cycle under backpressure *is* the designed behavior. (Flagged by both the storm and determinism lenses.)
2. **Doc-sync.** Wiring the retraction makes `is_hosted_locally` non-monotonic in production, which falsifies comments the reconcile shadow-compare is documented on: `router.rs:285` ("monotonic in production today"), `reconcile.rs:48-50/67/150` ("no on-main driver retracts / dead code"). Bounded impact (the controller drives nothing in shadow; only the `retract_diffs` counter is affected), but these must be corrected so the flip analysis isn't misled. Also two stale `hosting_heartbeat` references (`operations.rs:329`, `neighbor_hosting.rs:1329`) that no longer resolve (folded into `interest_heartbeat`).

### Follow-up hardening (not blocking; bounded today; tracked for later)
- **Uncapped full-set payload** — `HostingStateResponse` sends the full hosted-set IDs to every neighbor each cycle, no cap/rotation (unlike the 32-entry heal cap). ~32 KB × neighbors per 5 min at ~1k contracts (~2 KB/s at 1k×20); O(hosted × neighbors) with no ceiling. Worth a bound if the hosting budget grows to many thousands.
- **No per-peer rate limit on `HostingStateRequest`** — full-set response is ~3000× the request size. Pre-existing (on-connect exchange), now a routine 5-min path; not spoof-reflective (authenticated, encrypted transport). A per-peer min-response-interval would harden it.
- Nits: receiver-side full-replace can transiently drop a contract on same-peer wire reorder (self-heals next interval); `sort_by_key` String alloc per element; one pin test matches a symbol substring including comments.

Draft; stays parked for Ian sign-off (high-risk surface). Ships as part of the R1+R2 foundation release.

[AI-assisted - Claude]

sanity:
## Update 3 — Full-tier review folds applied (HEAD `24207b11`)

Folded the two minor fixes from the Full-tier review (verdict SOUND, no blockers/majors):

- **Fix 1 — non-blocking re-request.** The piggybacked `HostingStateRequest` now uses `try_send` (not `send().await`) on the cap-2048 event-loop notification channel (the #4145/#4231/#4442 wedge channel). A dropped cycle under backpressure is the designed behavior (heals next cycle); mirrors the sibling retraction path's `try_notify_node_event`. No `break` on error — shutdown is caught by the spread-delay `sleep_or_shutdown` and the interest `send().await` above. (`ring.rs`, channel-safety rule.)
- **Fix 2 — doc-sync** (AGENTS.md "fix outdated docs in the same PR"): wiring the retraction makes `is_hosted_locally` non-monotonic in production, which falsified several reconcile-shadow comments. Updated: `router.rs` (monotonic claim), `reconcile.rs` (×3: "no on-main driver retracts / production does nothing" and the two "on_contract_unhosted currently dead code" → now wired live on eviction in `RuntimePool::remove_contract`), `operations.rs` + `neighbor_hosting.rs` (stale `hosting_heartbeat` references → `interest_heartbeat`).

Not folded (tracked follow-ups, out of this PR's scope): uncapped full-set payload + per-peer request rate-limit (both bounded today, pre-existing shape).

Re-verified locally: `cargo build`, `clippy --locked -D warnings`, `fmt --check`, 27 unit+pin, and the touched sim tests (determinism + terminal-consult GET/SUBSCRIBE + get-routing-coverage + neighbor-hosting-bounded) all green. Single-commit shape kept (amend + force-push). CI re-running; will confirm green + SHA.

Stays a draft — not requesting merge.

[AI-assisted - Claude]

PR #4724, opened 2026-07-07
feat(server): zero-friction magic-link secrets migration (#4592)

sanity:
## Problem

The hosted "try Freenet" proxy (#4381) makes the **exit** one click (Account -> Export downloads an encrypted `freenet-data.fnsx`), but the **entrance** was a terminal command. #4603 removed the node-stop (live `POST /v1/import`), but a user still had to move a file and paste a key by hand — a friction cliff at the moment we want to convert a non-technical person who just tried Freenet in a browser.

## Solution

A **magic link** the user clicks on their own peer, which pulls the data over HTTPS and imports it live. The durable access key **never leaves the hosted browser**.

Four cooperating HTTP surfaces (same binary — a node can be either side):

- **`POST /v{1,2}/hosted/migrate/mint`** (hosted) — gated **identically to the live export** (hosted ON + loopback + `X-Forwarded-Proto: https` + valid `X-Freenet-User-Token`). Exports the user's scope under a **fresh ephemeral key** and stores the ephemeral-keyed bundle + key server-side under a **single-use, short-TTL, per-user-bounded** pull token. Returns the token; the hosted bar builds `http://127.0.0.1:7509/hosted/import?source=<origin>&pt=<token>`.
- **`GET /v{1,2}/hosted/migrate/pull?pt=`** (public) — the **pull token is the whole auth** (256-bit `OsRng`, single-use, short-TTL). Returns the bundle + ephemeral key over TLS. Unknown/expired/used ⇒ uniform `404` (no existence oracle).
- **`GET /hosted/import`** (local) — a first-party **confirmation page**; reads `source`/`pt` client-side and imports **only on an explicit click**, never on navigation ⇒ CSRF/drive-by safe.
- **`POST /v{1,2}/hosted/pull-import`** (local) — reuses the **live-import gate** (loopback + trusted origin + no per-contract token). **Strict SSRF allowlist** (`https` + default port + host == `try.freenet.org`); builds the pull URL from the trusted constant; fetches over HTTPS (size-capped); imports into `Local` with `overwrite=false` — **non-destructive**: an existing local secret is kept and reported in `skipped`, never clobbered (Ian's v1 decision: no overwrite path).

**Key insight (no crypto change):** `export_bundle(store, scope, material)` already decouples the user **scope** from the bundle **key**, so the mint reuses the existing `ExportUserSecrets` executor round-trip with an ephemeral key. The durable token stays on the hosting node. `Executor::export_user_secrets`'s param is renamed `token` -> `bundle_key_material` with a doc guard so nobody re-couples it.

Also included: the **file-upload fallback** UI on the local dashboard, and a **"Move to my peer"** button on the hosted bar. Reuses the merged #4603 import core throughout.

## Testing

- **Unit (`hosted_migrate`)**: pull-token lifecycle (single-use, unknown ⇒ none, **TTL expiry** via paused clock, per-user cap eviction with cross-user isolation); **SSRF allowlist** (https-only, exact host, no port, rejects loopback / look-alike subdomain / embedded-credential / non-url); pull-token + `pt` query validation; confirmation-page wiring (posts to `pull-import`, no auto-run).
- **Unit (`home_page`)**: file-upload card + modal wiring, dashboard JS `POST /v1/import` upload.
- **Integration (`hosted_mode_migrate.rs`)**: full **mint -> pull -> import** round-trip on a live hosted node (secret read back through the live stack), pull **single-use** (second pull 404), mint **gate** (no `XFP:https` / no token ⇒ 403), and `GET /hosted/import` serves the page without pulling.
- `cargo fmt` + `cargo clippy -- -D warnings` clean; prettier + eslint clean on shell assets.

Advances #4592

[AI-assisted - Claude]

https://claude.ai/code/session_012jmCd1tvDH1EsVTtaDVnxJ

### DIFF
--- a/crates/core/src/contract/executor/runtime/delegates.rs
+++ b/crates/core/src/contract/executor/runtime/delegates.rs
@@ -22,19 +22,28 @@ impl Executor<Runtime> {
     /// AEAD authentication — a torn read fails authentication and surfaces a
     /// clean export error, never silent corruption.
     /// The bundle is scoped to `user_context.scope()` — strictly the per-user
-    /// namespace, never `Local`. The `token` is the bundle-key material so the
-    /// user re-imports on their own peer with the token they already hold.
+    /// namespace, never `Local`. `bundle_key_material` is the secret the bundle
+    /// is encrypted under; it is DELIBERATELY decoupled from the scope. In the
+    /// self-reimport case (`GET /v1/hosted/export`) it is the user's own token,
+    /// so they decrypt with the token they already hold; in the magic-link
+    /// migration case (`hosted_migrate` mint) it is a FRESH EPHEMERAL key, so
+    /// the durable token never leaves the hosting node. Do NOT re-couple this to
+    /// `user_context`'s token — the two are independent by design.
     ///
-    /// The token and the derived key material live only in borrowed/`Zeroizing`
-    /// buffers here and inside `export_secret_bundle`; nothing is logged.
+    /// The key material and the plaintext it derives live only in
+    /// borrowed/`Zeroizing` buffers here and inside `export_secret_bundle`;
+    /// nothing is logged.
     pub fn export_user_secrets(
         &self,
         user_context: &UserSecretContext,
-        token: &[u8],
+        bundle_key_material: &[u8],
     ) -> Result<Vec<u8>, ExecutorError> {
         use crate::wasm_runtime::secret_export::{BundleKeyMaterial, ExportError};
         self.runtime
-            .export_secret_bundle(user_context.scope(), &BundleKeyMaterial::Token(token))
+            .export_secret_bundle(
+                user_context.scope(),
+                &BundleKeyMaterial::Token(bundle_key_material),
+            )
             .map_err(|e| {
                 // Preserve the over-limit case as a typed marker so the HTTP
                 // layer can map it to a 413 rather than a generic 500. The
--- a/crates/core/src/server/.prettierignore
+++ b/crates/core/src/server/.prettierignore
@@ -9,3 +9,4 @@ node_modules/
 **/assets/peer.html
 **/assets/peer_not_found.html
 **/assets/permission_prompt.html
+**/assets/hosted_import_page.html
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -96,6 +96,7 @@ fn sandbox_origin_from_headers(headers: &axum::http::HeaderMap) -> String {
 
 pub(crate) mod hosted_export;
 pub(crate) mod hosted_import;
+pub(crate) mod hosted_migrate;
 mod permission_prompts;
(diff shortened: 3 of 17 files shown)


sanity:

Auth + data-migration critical path, so Full tier. The `freenet:pr-review` skill wasn't available in this environment, so the review was run manually per `multi-model-review.md`: three independent Claude reviewers with distinct adversarial lenses (token-lifecycle/concurrency, SSRF/CSRF, code-first/big-picture), an external **Codex** pass, and a focused re-review of the fix delta. **No blockers.** Every finding is fixed or has a justified dismissal.

### Fixed
- **Byte-budget griefing** (Reviewer A + Reviewer C + the `rule-review/warnings` CI check): the global byte budget alone let one account minting large exports evict every other user's pending migration. Added a **per-user byte budget** that evicts the minter's *own* oldest first (global budget kept as a last-resort OOM ceiling). Parameterized the bounds (`StoreLimits` / `insert_with_limits`, mirroring `export_bundle_with_limits`) and added tests: per-user-evicts-own-not-others, oversize→503, global-count-cap.
- **Clickjacking** (Reviewer B): `/hosted/import` now ships `frame-ancestors 'none'` + `X-Frame-Options: DENY` + `no-store` (mirrors the permission-prompt page), so a remote page can't frame the confirmation click.
- **SSRF-via-proxy** (Reviewer B): the outbound pull client sets `.no_proxy()`.
- **`read_pull_response` untested** (Reviewer C): extracted it and added unit tests (2xx / key / kind-default+passphrase / non-2xx / missing-key / empty) via a synthesized `reqwest::Response`.
- **`no-store` on the pull response** (fix-delta reviewer): the bundle+key response now carries `Cache-Control: no-store`.
- **Secret hygiene** (Reviewer A): `random_token` now zeroizes its raw 32-byte buffer (it also mints the ephemeral key).
- **Style**: removed em-dashes from the dashboard import copy; fixed import grouping; kept the hosted-bar "Export data" label (fixing a pin-test regression).

### Justified dismissals
- **Codex P1 — "`synth_response` test helper won't compile"**: false positive. Those exact `read_pull_response_*` tests compile and pass locally (full rebuild) and in CI's Unit & Integration. reqwest 0.12 implements `From<http::Response<T: Into<Body>>>`, and axum re-exports the same http 1.x crate the workspace unifies; the compiler would catch any desync loudly.
- **Hardcoded local port `7509` / single-host allowlist**: intentional v1 scope (maintainer decisions — strict allowlist, no port override).
- **`tokio::time::Instant` vs `TimeSource`**: both reviewers call it acceptable — this is the axum HTTP-server layer (not the DST-simulated node core), and it's deterministically controllable in tests via `start_paused` + `advance`.

The fix-delta re-review confirmed the reworked eviction math (caps exact after insert, both loops terminate, correct scoping, no panic/deadlock) and that the extraction is verbatim.

[AI-assisted - Claude]

PR #4725, opened 2026-07-07
feat(hosting): flip reconcile controller to DRIVE interest-gated renewal + collapse (P6, #4642)

sanity:
## Problem

The pure `reconcile()` controller (`crates/core/src/ring/hosting/reconcile.rs`,
#4642 "the maintenance / reconcile loop") was wired in **shadow mode** at the
hosting decision sites: it computed an `Action` set, recorded per-site divergence
telemetry, and **drove nothing**. The FLIP (P6, release-2 keystone) makes the
controller actually drive.

This PR is the **NARROW flip** — a data-driven decision made from v0.2.93 shadow
telemetry. It flips the two interest-gated **maintenance** decisions from shadow
to driving:

1. **Interest-gated RENEWAL** (the first increment).
2. **Interest-gated COLLAPSE** (this increment) — so both key off the single
   strict-farther `contract_in_use` gate: *"a peer keeps hosting exactly as long
   as it keeps renewing, and stops both together"* (design doc §6).

It deliberately does **NOT** compute-upstream-everywhere. The shadow data:

| Signal | v0.2.93 shadow | Decision |
|---|---|---|
| Emitted **actions** vs production | collapse **0%**, renew **0.3%**, focused edge sites 0.00% / 0.02% / 0.67% | driving renewal + collapse is **de-risked** → FLIP |
| Computed-vs-stored upstream **identity** (#4671) | **~79% divergence** (verified real) | keep the stored `is_upstream` flag → stays SHADOW |

So the upstream-picking stays on the stored `is_upstream` flag; the wire
`Unsubscribe` still targets the **stored** upstream, and the computed-vs-stored
identity divergence keeps being measured under the unchanged
`record_upstream_divergence_comparison` shadow, unflipped, so it can be
re-measured before a future full flip.

### Renewal (driven — first increment)

`Ring::recover_orphaned_subscriptions` gates its per-contract renewal spawn on
`OpManager::reconcile_wants_renewal` = design §5a `contract_in_use` (renew iff a
local client OR a STRICTLY-farther downstream depends on this peer). Non-renewal
is the collapse primitive (§5a): a peer that stops renewing lets its lease lapse
and the chain collapses inward.

### Collapse (driven — this increment)

The three collapse-decision sites now gate the active `send_unsubscribe_upstream`
on `OpManager::reconcile_wants_collapse` = `reconcile::wants_collapse` =
`!contract_in_use` (the exact inverse of the renewal gate), replacing the legacy
ANY-downstream `Ring::should_unsubscribe_upstream`:

- the maintenance loop's downstream-expiry teardown,
- the client-disconnect teardown (`client_events`),
- the inbound-unsubscribe teardown (`handle_unsubscribe_inbound`).

**Keeps the stored upstream identity:** `send_unsubscribe_upstream` still resolves
the wire target via the stored `is_upstream` flag and drops the local lease via
`ring.unsubscribe`. Only the *decision* flips.

**Fail-safe:** `reconcile_wants_collapse` returns `false` (do NOT collapse) on a
distance-blind snapshot — the mirror of `reconcile_wants_renewal`'s keep-renewing
default. Neither tears a chain down on an unmeasurable read.

**Superset, not a leak:** the strict gate is a superset of the legacy gate
(`!has_any_downstream ⟹ !has_farther_downstream`), so the flip can only collapse
MORE (the mutual-co-host case), never leak a chain the legacy predicate would
have cleaned up. The active Unsubscribe toward the stored upstream is a
best-effort accelerant; the **load-bearing collapse remains the passive
lease-lapse** from non-renewal.

**Telemetry:** the Collapse and InboundUnsubscribe reconcile counters are
retained but repurposed (as the renewal counter was) to measure how often the
driven strict-farther gate DISAGREES with the legacy predicate it replaced — the
"collapse diff" ship-gate falsifier — recorded per-site inside
`reconcile_wants_collapse`.

### Still deferred (unchanged)

- **Compute-upstream-everywhere / retiring `is_upstream`** — blocked on the ~79%
  identity divergence; stays under shadow measurement.
- **`ReRootSearch` (connection-drop re-root) and `Announce` (host-formation)** —
  stay in shadow.
- **`Retract`** — wired on eviction only (#4722); not driven from teardown yet.

## Testing

- **Pure-core units** (`reconcile.rs`): `wants_collapse_is_exact_inverse_of_renewal`
  (gate duality + strict-gate collapse), `wants_renewal_drives_interest_gate`.
- **Driver pins** (`op_state_manager.rs`): `collapse_decision_is_reconcile_driven`
  (all three sites gate on `reconcile_wants_collapse`, fail-safe `false`),
  `send_unsubscribe_upstream_drives_off_stored_upstream_after_flip` (stored
  upstream + `record_upstream_divergence_comparison` kept, Collapse action-set
  shadow removed), `renewal_site_is_driven_not_shadowed`, and the updated
  inbound-unsubscribe gate pin.
- **Sim falsifiers** (`--features simulation_tests,testing`), which now exercise
  the flipped collapse decision sites:
  `test_subscription_chain_collapses_on_client_leave` (client-leave → no hollow
  relay), `test_subscription_count_tracks_demand_not_cache` (demand chains
  survive the strict gate; cache-only chains collapse), `test_interest_renewal`,
  `test_subscription_root_renewal_does_not_storm`.

**Why no new "active-collapse without lease-expiry" sim test:** the active
Unsubscribe toward the *stored* upstream is best-effort (it can miss under the
known ~79% #4671 stored-flag drift), so a test asserting prompt collapse *without*
the lease-lapse would be flaky. The load-bearing collapse is the passive
lease-lapse, which the existing falsifiers validate; the strict-gate decision is
pinned by the unit tests.

`cargo build -p freenet`, `cargo clippy --locked -p freenet -- -D warnings`, and
`cargo fmt --check` are green.

## Ship gate (do NOT merge)

Highest-risk release-2 keystone: **FULL multi-model review + Ian sign-off**
required, and it ships **only** in the inseparable 0.2.94 demand-driven-hosting
bundle after R1+R2 field-validate. Draft for review of the complete narrow flip
(renewal + collapse) + the deferral analysis. Past the late-night merge line.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -1559,9 +1559,17 @@ async fn process_open_request(
                     );
                 }
 
-                // Send Unsubscribe upstream for contracts with no remaining interest
+                // Send Unsubscribe upstream for contracts with no remaining interest.
+                // FLIP (keystone P6, #4642): the collapse decision is driven by the
+                // reconcile controller's strict-farther interest gate
+                // (`reconcile_wants_collapse` = `!contract_in_use`), replacing the
+                // legacy ANY-downstream `should_unsubscribe_upstream`. The teardown
+                // still targets the STORED upstream (narrow flip keeps the flag).
                 for contract in &result.affected_contracts {
-                    if op_manager.ring.should_unsubscribe_upstream(contract) {
+                    if op_manager.reconcile_wants_collapse(
+                        contract,
+                        crate::node::network_status::ReconcileShadowSite::Collapse,
+                    ) {
                         let op_mgr = op_manager.clone();
                         let contract = *contract;
                         GlobalExecutor::spawn(async move {
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -197,11 +197,14 @@ pub struct NetworkStatus {
     /// site's divergence must be separately gate-able — a single global tally
     /// would be dominated by the renewal set size and hide the other signals.
     ///
-    /// `collapse` and `renewal` are the MAINTENANCE sites (full action-set
-    /// comparison). The other three are single-aspect EDGE sites (focused
-    /// comparison on one action class): `inbound_unsubscribe` (teardown on a
-    /// downstream leave), `connection_drop` (re-root on an upstream loss),
-    /// `host_formation` (announce on first host).
+    /// `collapse`, `renewal`, and `inbound_unsubscribe` are the FLIPPED (P6) sites:
+    /// they no longer record a shadow divergence — the controller DRIVES — and their
+    /// counters are REPURPOSED to measure the flip's effect (renewal: strict gate
+    /// SUPPRESSED a renewal; collapse / inbound_unsubscribe: driven strict-farther
+    /// gate DISAGREED with the legacy `should_unsubscribe_upstream` predicate it
+    /// replaced, per collapse site). `connection_drop` (re-root on an upstream loss)
+    /// and `host_formation` (announce on first host) remain single-aspect EDGE
+    /// shadow sites (focused comparison, record-only).
     pub reconcile_shadow_collapse: ReconcileShadowStats,
     pub reconcile_shadow_renewal: ReconcileShadowStats,
     pub reconcile_shadow_inbound_unsubscribe: ReconcileShadowStats,
@@ -238,7 +241,15 @@ pub struct UpstreamDivergenceStats {
 /// ONE focused action class (see `action_set_divergence_focused`).
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ReconcileShadowSite {
-    /// The interest-gated collapse (`OpManager::send_unsubscribe_upstream`).
+    /// The interest-gated collapse. **FLIPPED (keystone P6, #4642):** the collapse
+    /// DECISION is now DRIVEN by `OpManager::reconcile_wants_collapse` at the three
+    /// teardown sites (maintenance-loop downstream-expiry, client-disconnect,
+    /// inbound-unsubscribe), replacing the legacy ANY-downstream
+    /// `Ring::should_unsubscribe_upstream`. The counter is retained but repurposed:
+    /// `comparisons` counts collapse-gate decisions and `divergences`/`collapse`
+    /// count the times the driven strict-farther gate DISAGREED with that legacy
+    /// predicate — the "collapse diff" ship-gate falsifier (~0% in v0.2.93 shadow),
+    /// not a shadow divergence. Recorded in `reconcile_wants_collapse`.
     Collapse,
     /// The subscription-renewal set (`Ring::contracts_needing_renewal`, driven
     /// by the recovery loop). **FLIPPED (keystone sub-task 3, #4642):** this site
@@ -250,8 +261,12 @@ pub enum ReconcileShadowSite {
     /// not cache size" ship-gate falsifier (design §5a / #3763), not a divergence.
     Renewal,
     /// A downstream peer unsubscribed (`subscribe::handle_unsubscribe_inbound`):
-    /// did that leave us not-in-use (strict-farther), i.e. would the controller
-    /// tear down? Focused on `Collapse`.
+    /// did that leave us not-in-use (strict-farther), i.e. should the controller
+    /// tear down? **FLIPPED (keystone P6, #4642):** this collapse decision now
+    /// DRIVES via `reconcile_wants_collapse` (passed this `site`). The counter is
+    /// retained but repurposed like `Collapse`: it measures how often the driven
+    /// strict-farther gate DISAGREED with the legacy `should_unsubscribe_upstream`
+    /// predicate at THIS site (~0.00% in v0.2.93 shadow), not a shadow divergence.
     InboundUnsubscribe,
     /// A ring connection dropped (`OpManager::on_ring_connection_lost`): for a
     /// contract that peer co-hosted, would the controller re-root (upstream
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -241,7 +241,13 @@ pub enum ReconcileShadowSite {
     /// The interest-gated collapse (`OpManager::send_unsubscribe_upstream`).
     Collapse,
     /// The subscription-renewal set (`Ring::contracts_needing_renewal`, driven
-    /// by the recovery loop).
+    /// by the recovery loop). **FLIPPED (keystone sub-task 3, #4642):** this site
+    /// no longer records a shadow — it DRIVES (`OpManager::reconcile_wants_renewal`
+    /// gates the renewal spawn). The counter is retained but its meaning shifted:
+    /// `comparisons` now counts renewal-gate decisions and `divergences`/`renew`
+    /// count the times the STRICT-farther interest gate SUPPRESSED a renewal the
+    /// old ANY-downstream set would have done — the "renewal tracks active demand,
+    /// not cache size" ship-gate falsifier (design §5a / #3763), not a divergence.
     Renewal,
     /// A downstream peer unsubscribed (`subscribe::handle_unsubscribe_inbound`):
     /// did that leave us not-in-use (strict-farther), i.e. would the controller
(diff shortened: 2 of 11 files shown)


sanity:
Local validation (renewal flip DRIVING):

**Unit** — 13 targeted lib tests green, incl. `wants_renewal_drives_interest_gate` (the flip's decision across representative snapshots: in-use → renew; upstream-lost re-root → renew; strict-gate not-in-use → do NOT renew; settled root → no renew; idle → no renew), `renewal_site_is_driven_not_shadowed`, `reconcile_core_stays_pure`, plus kept `has_strictly_farther_excludes_equal_uses_exact_ordering`, `test_active_lease_renewed_iff_contract_in_use`, `reconcile_shadow_export_maps_each_field_to_its_own_site`.

**Sim falsifiers** — `cargo test --features simulation_tests,testing --test simulation_integration`, **3 passed / 0 failed** (216s):
- `test_subscription_chain_collapses_on_client_leave` (3 seeds) — chain collapses to zero subscribed peers ⇒ chain-collapse-on-client-leave **and** no self-perpetuating cycle (a lingering mutual co-host pair would leave survivors).
- `test_subscription_root_renewal_does_not_storm` — no renewal storm (the flip only *narrows* the renewal set via the strict gate, never widens it).
- `test_interest_renewal` — renewal preserved while interest exists.

CI static checks (Clippy / Fmt / Rule Lint / Conventional Commits) pass; CI Simulation + Unit & Integration running for independent confirmation.

[AI-assisted - Claude]

sanity:
Update — CI's Simulation job caught a regression in the first commit and it is now fixed (force-pushed `ce0709a4`).

**Root cause:** the first draft gated the driven renewal on the reconcile *action set* (`Renew`/`Subscribe`/`ReRootSearch`). That over-suppresses an in-use keyward **root** (a local client, but no upstream to renew toward → reconcile emits `[]`/`[Announce]`) and a peer still **acquiring** its first lease before its upstream advertisement is recorded. `test_subscription_count_tracks_demand_not_cache` seed `4642d102` failed: the hub spawned 0 renewals and its demanded leases lapsed.

**Fix:** gate on **`contract_in_use`** directly — design §5a's rule is 'renew iff a local client OR a strictly-farther downstream depends on this peer', which gates on *demand*, not lease/upstream/root state. A root renews too (its `run_renewal_subscribe` is a harmless keyward no-op that also re-establishes a lease lapsed under churn). `reconcile::contract_in_use` is now the single predicate driving both the collapse rule inside `reconcile` and the renewal gate. Added a regression pin to `wants_renewal_drives_interest_gate`.

**Re-validated locally** (17 lib unit tests + 4 sim tests: `test result: ok. 4 passed; 0 failed` in 220s — demand-not-cache, chain-collapse ×3 seeds, no-storm, interest-renewal). CI re-running the full suite on `ce0709a4`.

[AI-assisted - Claude]

sanity:
CI is now fully green (all 11 checks pass on `ce0709a4`).

The one earlier failure — `test_nat_subscription_cross_peer_update_via_docker_nat` (NAT Validation) — was a **Docker-NAT timing flake**, confirmed by re-run (passed, 5m20s). Evidence it is unrelated to this hosting/renewal change: it PASSED on the first commit (`4bf5c1d4`, which gated on the reconcile action set — a strict *subset* of the renewals the `contract_in_use` fix now issues), so the superset fix cannot reduce subscription delivery; it also passes on other current PRs; and Docker-NAT is transport-layer, orthogonal to the renewal gate. Simulation + Unit & Integration re-ran deterministically green both times.

[AI-assisted - Claude]

sanity:
## Full multi-model review — narrow reconcile flip

Reviewers: external **Codex** (gpt-5.5, `codex review --base main`, ran clean on head `835cfd59`) + two adversarial Claude lenses (skeptical bug-hunt, code-first). Consolidated below; every `file:line` verified against the checked-out code.

**Verdict: the collapse flip is sound** — `wants_collapse = !contract_in_use` is the exact inverse of the renewal gate and a genuine *superset* of the legacy `should_unsubscribe_upstream` (collapses more, never leaks a chain legacy would have cleaned up); all three collapse sites are consistently gated with no bypass caller; the compute-upstream identity shadow (the ~79% measurement) is confirmed untouched and still exporting; no deadlock/cancellation issue; no under-collapse/hollow-relay (#3763) risk. No crash-class or merge-blocking-crash findings. The items below are to resolve before this ships in the 0.2.94 bundle.

### 1. [P2 — needs a decision] Renewal gate drops the recent-local-access signal (#3769 regression)
`ring.rs:2342` gate + `reconcile.rs:508`. A contract reaching renewal only via `contracts_needing_renewal()` **branch 3** (`has_recent_local_client_access` — a local GET read with no WS subscription and no downstream) now reads `!contract_in_use` and is skipped, so its lease lapses and it leaves the update mesh; a later local read serves a stale copy. The builder still *emits* branch-3 (dead work), and the existing test `test_local_client_access_enables_renewal` only covers the builder, not the flipped loop, so it passed despite the regression.
**The decision:** is a bare local read "demand" (keep it fresh) or "cache" (let it lapse)? The strict demand-driven reading favors lapsing — but the freshness backstop that would justify it (piece-D InterestSync anti-entropy) is **not shipped yet**, so lapsing now = real staleness. **Recommendation: preserve #3769 for now** (fold `has_recent_local_client_access` into the renewal side of the gate), and revisit once anti-entropy lands. If instead we accept the lapse, branch 3 + its test must be removed and the behavior + missing backstop documented. Not applying either autonomously — it's a demand-model call.

### 2. [MEDIUM] Two divergent `contract_in_use` definitions; alignment comment now false
`reconcile.rs:508` (STRICT: `has_local_client || has_farther_downstream`) now drives collapse+renewal, while `hosting.rs:1244` (ANY: `has_client_subscriptions || has_downstream_subscribers`) still drives eviction/`subscriber_count`. A closer-only-downstream contract is now not renewed (lapses → stale) yet `subscriber_count ≥ 1` so eviction protects it → a **stale-but-retained** copy until its downstream re-roots (piece D, unshipped). The comment at `hosting.rs:1248-1253` asserting the two align is now **false and misleading**. Fix the comment; and reconcile the two definitions against the eviction rework (which is splitting `subscriber_count` into local + downstream) so retention and renewal use a coherent demand signal.

### 3. [MEDIUM — ship-gate evidence] The "~0% collapse diff" de-risking is circular for 2 of 3 sites
The pre-flip Collapse shadow lived *inside* `send_unsubscribe_upstream`, reached at the client-disconnect and maintenance sites only when legacy already collapsed — so it measured only the under-collapse direction (~0% tautologically, since strict ⊇ legacy) and was structurally blind to over-collapse. Only the *inbound-unsubscribe* site measured over-collapse (~0.00%), and only on that population. Client-disconnect — a contract that just lost its last local client with a lingering closer/mutual co-host downstream — is exactly where over-collapse is most likely and was never measured there. Don't over-claim the evidence: the post-flip telemetry now records unconditionally at all three sites, so **field-validate the over-collapse rate during the R1+R2 window before this bundle ships**, or add a discriminating test.

### 4. [LOW] Inline unbounded redb read at client-disconnect
`client_events.rs:1568` — the disconnect gate went from a cheap DashMap check to `reconcile_wants_collapse`, which runs full `build_reconcile_inputs` (a redb `get_state_size`, keyward-neighbor + subscription-root scans) inline+synchronous per held contract, then re-runs the legacy predicate for the counter — while `wants_collapse` consumes only 2 of ~8 computed fields. Latency regression on the disconnect path (not a deadlock). Fix: compute only the fields the gate needs.

### 5. [LOW / NIT]
- Over-collapse of a temporarily-unreachable farther downstream (`op_state_manager.rs:902`) — transient, self-heals on reconnect, strictly the superset direction.
- Sim falsifiers don't *discriminate* the collapse (they pass with/without the gate — the clock-jump expires leases regardless); discriminating coverage is the unit test + source pins. Accepted, documented gap; a dedicated active-collapse test would be flaky under the 79% drift.
- Stale `ReconcileShadowStats` struct-level doc; telemetry asymmetry (renewal counter includes distance-blind fail-safe ticks, collapse counter excludes them → the two suppression ratios aren't directly comparable).

**No findings** in: under-collapse/hollow-relay, stored-upstream/79%-drift correctness, cancellation/async safety.

**Disposition:** draft, not merging. Finding 1 needs a demand-model decision (recommendation: preserve #3769 until anti-entropy ships). Findings 2 and 4 are fixes to apply before merge; Finding 3 is a field-validation gate for the R1+R2 window. Re-run this review on the fixed content per the per-code-content rule.

[AI-assisted - Claude]

PR #4726, opened 2026-07-07
fix(sim): make direct-runner fault injection actually drop packets (#4694)

### LINKED ISSUE #4694: Sim fault injection is inert on the direct/controlled runner (packet-delivery callback never installed)
## Problem

`set_packet_delivery_callback` (`transport/in_memory_socket.rs:172`) is only ever called with `None` in the tree (`node/testing_impl.rs:5404`). So `check_packet_delivery()` always returns `Deliver`, and fault injection that relies on it — `ChurnConfig` crashes, `crash_node`, `add_partition` — sets the fault **config** but never actually drops packets.

Consequence: sim tests that inject a crash/partition and then assert convergence may be **false-green** — they assert recovery from a fault that never actually happened. Candidates to audit: `test_partition_heal_convergence`, `test_crash_recover_convergence`, `test_multi_step_churn` (`simulation_integration.rs`), and any direct-runner `ChurnConfig` test.

## Discovered by

#4657 review (demand-driven hosting, #4642). #4657 fixed this for the **controlled-sim** path only (opt-in `FaultInjectorState.enforce_crash_drops`, so `CrashNode` now drops packets and `test_scripted_node_crash_mid_run` is discriminating). It deliberately did NOT touch the direct-runner path to avoid a large blast radius across existing churn tests.

## Fix needed

1. Audit which existing tests use direct-runner fault injection (crash/partition/churn).
2. Wire the packet-delivery callback (or equivalent enforcement) for that path.
3. Re-run those tests — some may expose real failures once faults actually take effect. Any that do are pre-existing bugs the inert injection was masking.

This is a test-integrity issue: per the project testing philosophy, a fault-tolerance test that doesn't inject the fault is a broken test.

[AI-assisted - Claude]

sanity:
## Problem

`set_packet_delivery_callback` was only ever installed by `run_controlled_simulation` (opt-in `enforce_crash_drops`, added in #4657). The **direct runner** (`run_simulation_direct`) never installed the callback and never opted in, so `check_packet_delivery()` always returned `Deliver`.

The direct runner's chaos driver (`ChurnConfig`) calls `config.crash_node(addr)` on the fault injector, but with no callback + no enforcement those calls only mutate fault **config** — they never drop a single packet. A "crashed" node keeps exchanging packets.

**Consequence (test-integrity bug):** any near-K churn / crash / partition validation on the direct runner is **false-green** — it asserts recovery from a fault that never actually happened. `ChurnConfig` is a direct-runner-only feature and the demand-driven-hosting **release-2 (0.3.0)** churn/OOM gate (#4642) runs on this exact runner, so trusting that gate first requires the fault to be real.

## Approach — option (a): fix the direct runner

Two options were considered:

- **(a) Install the callback on the direct runner** so `ChurnConfig` crashes/partitions genuinely drop packets. **← chosen.**
- (b) Move the near-K churn/OOM gate onto the controlled-sim (Turmoil) path where drops already work.

(a) wins on both "genuinely drops packets on the runner the gate uses" and "least blast radius": `ChurnConfig` (random crash/recover) and 500+ node scale are direct-runner-only; option (b) would require reworking the churn model onto Turmoil's scripted-op path, losing both. #4657 avoided the direct-runner path to limit blast radius, but there are only two direct-runner churn tests in-tree and both tolerate real crashes (see Testing), so the risk is small.

### Changes

- **`run_simulation_direct`**: install the global packet-delivery callback and set `enforce_fault_drops = true` on this network's fault injector, **gated on `churn_config.is_some()`** so non-churn direct-runner tests are byte-for-byte unchanged. Gating (rather than always-on) keeps the change surgical and makes the wiring impossible to miss when churn IS configured. `SimNetwork::Drop` already clears the callback.
- **`fault_injection_delivery_decision`**: now also drops packets blocked by an **active partition** (counted in `messages_dropped_partition`), not just crashed nodes. Renamed the opt-in flag `enforce_crash_drops` → `enforce_fault_drops` to reflect both fault classes. Purely additive for existing callers (no in-tree test adds a partition on a callback-installed network).

## Testing

- **`test_direct_runner_churn_actually_drops_packets`** (new regression, non-nightly): runs the direct runner with permanent churn, retains the fault-injector `Arc` across the run, and asserts `messages_dropped_crash > 0`.
  - **RED** on current `main` behavior (counter stays exactly `0` — verified by temporarily disabling the install block: `messages_dropped_crash = 0`, assertion fails).
  - **GREEN** with the fix.
- **`test_fault_injection_delivery_decision_drops_crash_and_partition`** (new unit): crash (both directions) + partition (both directions) + the opt-in gate + counter accounting.
- Regression checks (all pass): `test_direct_runner_churn` (existing churn test, unchanged assertions), `test_scripted_node_crash_mid_run` (#4657 controlled-sim path, exercises the renamed flag), `test_direct_runner_determinism` (direct-runner determinism preserved — install is gated on churn), `test_topology_single_hoster` (controlled-sim path).
- `cargo fmt --check`, `cargo clippy --locked -p freenet -- -D warnings` clean.

## Scope note

This fixes the **direct-runner** path (the issue's stated fix target and the release-2 gate's runner). The Turmoil `run_simulation` path (`test_partition_heal_convergence`, `test_crash_recover_convergence`, `test_multi_step_churn`) *also* injects faults via the freenet fault injector without installing the callback and is likewise inert — but wiring it has a larger blast radius across convergence tests and is a separate surface from the release-2 gate. Flagging for a follow-up audit rather than smuggling it into this focused fix.

Closes #4694
Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/in_memory.rs
+++ b/crates/core/src/node/network_bridge/in_memory.rs
@@ -97,13 +97,15 @@ pub struct FaultInjectorState {
     /// Network name for delivering pending messages
     pub network_name: Option<String>,
     /// Opt-in: when true, the sim's global packet-delivery callback DROPS every
-    /// packet to/from a node in `config.crashed_nodes` (counting it in
-    /// `stats.messages_dropped_crash`). Off by default so installing the global
-    /// callback does not silently change fault behavior for networks that never
-    /// opted in (e.g. the direct-runner churn driver, whose crash semantics are
-    /// unchanged). `SimNetwork::run_controlled_simulation` sets it so
-    /// `SimOperation::CrashNode` is a real crash rather than a no-op (#4642 piece F).
-    pub enforce_crash_drops: bool,
+    /// packet to/from a crashed node (`config.crashed_nodes`, counted in
+    /// `stats.messages_dropped_crash`) OR blocked by an active partition
+    /// (`config.partitions`, counted in `stats.messages_dropped_partition`).
+    /// Off by default so merely installing the global callback does not silently
+    /// change fault behavior for networks that never opted in. Both
+    /// `SimNetwork::run_controlled_simulation` (scripted `CrashNode`, #4642 piece F)
+    /// and `SimNetwork::run_simulation_direct` (churn driver crashes, #4694) set it
+    /// so their injected faults are real packet drops rather than silent no-ops.
+    pub enforce_fault_drops: bool,
 }
 
 impl FaultInjectorState {
@@ -116,7 +118,7 @@ impl FaultInjectorState {
             pending_messages: Vec::new(),
             stats: NetworkStats::default(),
             network_name: None,
-            enforce_crash_drops: false,
+            enforce_fault_drops: false,
         }
     }
 
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -388,16 +388,18 @@ impl ScheduledOperation {
 }
 
 /// Packet-delivery decision consulted by every `SimulationSocket` (via the
-/// global delivery callback set in `run_controlled_simulation`) so a scripted
-/// node crash actually drops packets instead of being a silent no-op.
+/// global delivery callback installed by `run_controlled_simulation` and
+/// `run_simulation_direct`) so an injected node crash or partition actually
+/// drops packets instead of being a silent no-op.
 ///
 /// Per-network: it looks up the fault injector for the socket's own network and,
-/// only if that network opted in (`enforce_crash_drops`), DROPS any packet whose
-/// source or destination is a crashed node — counting it in
-/// `stats.messages_dropped_crash`. Every other packet (and every network that
-/// did not opt in) is delivered unchanged, so installing this callback cannot
-/// alter fault behavior for the direct-runner churn driver or any other path.
-/// See #4642 piece F.
+/// only if that network opted in (`enforce_fault_drops`), DROPS any packet whose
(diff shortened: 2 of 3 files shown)


sanity:
## Consolidated review — direct-runner fault-injection fix

External **Codex** (`codex review --base main` → "No actionable correctness issues") + a Claude adversarial lens. **Verdict: clean, no blockers.** The fix is correct, well-tested, RNG-safe, and — the crux — it targets the runner the release-2 churn/OOM gate actually uses.

**Crux verified (PASS):** the near-K churn / OOM gate runs on `ChurnConfig` (direct-runner-only) → fixed here. The no-storm / chain-collapse / findability falsifiers run on the already-wired controlled-sim path (#4657) and are trustworthy independent of this PR.

**Also verified clean:** drop logic (crash + partition, both directions, only when enabled); gating (`churn_config.is_some()`, non-churn direct tests byte-for-byte unchanged); regression test genuinely RED without the fix; no RNG/determinism perturbation.

**Follow-up (tracked, not blocking):** the Turmoil `run_simulation` path (partition-heal / crash-recover / multi-step-churn) is still inert / false-green — filed as https://github.com/freenet/freenet-core/issues/4727. Must be wired before release-2 piece-G partition/fork validation relies on it. Minor gate-hardening (fault-config-non-empty vs churn-only) folded into that issue.

Review-clean; will merge right after v0.2.93 is cut (holding so the release stays exactly the smoke-tested `4c1c1ba8`).

[AI-assisted - Claude]

PR #4732, opened 2026-07-07
test(hosting): every-hop LRU placement falsifiers + summarize-storm counter (#4642, spec step 8)

sanity:
## Problem

Spec step 8 of the demand-driven-hosting redesign (`hosting-spec.md` §Placement) is **every-hop LRU placement**: every hop on a PUT route hosts the contract, every return hop on a GET does the same, and the R1+R2 subscriber-primary demand gauge (not durability) is what separates this from the old relay-caching anti-pattern — copies are the lowest-value eviction candidates the instant capacity binds. Step 8 gates the relay-caching-free bundle, so it needs its falsifiers in CI: PUT-then-GET findability under churn, fanout measured, and summarize-load flat vs hosted-set size (the #4440 storm falsifier).

## Key finding: the every-hop LRU placement *mechanism* already ships on `main`

This was verified against current `origin/main` (which has R1+R2's subscriber-primary eviction, #4717/#4720):

- **Every PUT hop hosts, unconditionally.** `drive_relay_put` / `drive_relay_put_streaming` call `relay_put_store_locally` at step 1 ("all nodes cache") *before* any routing/terminus decision — `put_contract` (merges, returns the post-merge body) → `host_contract` → `announce_contract_hosted`. No HTL/distance/flag gate.
- **Every GET return hop hosts.** `cache_contract_locally` runs at every relay Found arm; announce is gated on `is_new && put_persisted`.
- **Reconcile-before-announce holds.** Main ships full state hop-by-hop on PUT (summary-first PUT is not merged), so the body is in-hand before announce; on GET, announce is gated on the PutQuery having persisted the body+code (the #4223 class is closed).
- **Zero-subscriber every-hop copies rank lowest.** They have `subscriber_count == 0`, so under `MemoryPressure::AtCapacity` they are the only eviction-eligible entries and are shed first by least-recent GET (`evict_over_budget` / `victim_order`). Composition with R1+R2 confirmed.
- **The summarize storm is already mitigated.** The state-hash-keyed summary cache (`bridged_summarize_contract_state`) returns without loading state or running WASM on an unchanged-state hit, and the per-heartbeat InterestSync path (`summary_if_hosted_or_in_use`) plus the per-target UPDATE broadcast path (`broadcast_to_single_peer` → `get_contract_summary`) both route through it.

So step 8 on `main` is **not** a new placement mechanism — it is the **falsifiers that gate the bundle** plus the telemetry to measure one of them. The seed-lease deletion in the spec's step 8 is moot (there is no seed lease on `main`; it only existed on a prototype branch), so this PR neither adds nor removes one.

## Approach

No production behavior changes. This PR adds:

1. **A summarize-WASM-call counter** (`ring::topology_registry::record_summarize_wasm_call`, incremented on the summary-cache **slow path** in `bridged_summarize_contract_state`, right before the WASM `summarize_state`). It is **sim-scoped**: `record_*` no-ops outside a simulation (reads the sim-only network-name thread-local), so production nodes pay only a thread-local read on a cache *miss* (already the rare/expensive path). Captured per-peer into `ControlledSimulationResult` (`total_/max_/per-addr` accessors), mirroring the existing `RenewalMetrics` pattern.
2. **Two deterministic sim falsifiers** (in `simulation_integration.rs`):
   - `test_every_hop_fanout_bounded_by_degree_not_holder_count` — falsifier (b), "fanout measured as a check". 1 gw + 11 nodes, sparse `max_connections=5`, all GET+subscribe, 12 UPDATEs. Measured: holders=12, max sends-per-update=5. Asserts max sends-per-update ≤ degree(5) and **< holders(12)** — per-peer fanout is bounded by co-host degree, not by the holder count.
   - `test_every_hop_summarize_calls_flat_vs_fanout` — falsifier (c), the #4440 summarize-storm guard for every-hop load. Dense star (1 gw + 6 co-host subscribers), 12 distinct-state UPDATEs. Measured: busiest-peer summarize=13 (≈K=12), total=91, vs 510 per-target UPDATE broadcasts. Asserts per-peer summarize ≤ K+8 (scales with state-change rate, **not** fanout width) and total ≪ per-target broadcasts (the cache collapses a generation's per-target summaries). A broken/removed cache re-arms the storm and blows both bounds.

Falsifier (a) — PUT-then-GET findability under near-K churn — is **already** covered by `test_piece_e_findability_sparse_ring_gate` (mean client-findability ≥ 0.60 across 6 seeds, with every-hop hosting present); not duplicated.

Two notes surfaced while building the falsifiers:
- The counter is recorded only in the generic `bridged_summarize_contract_state` (the production `ContractExecutor` path). The default sim runtime (`MockRuntime` + CRDT emulation) has its own `summarize_contract_state` that bypasses it, so falsifier (c) runs `use_mock_wasm=true` to exercise the real path. Production real-WASM contracts always hit the counter — this is a test-harness nuance, not a production gap.
- Every-hop makes subscribers real relays, so a committed update re-propagates well past a naive K×N. The clean #4440 discriminator is therefore **per-peer** summarize (busiest peer ≈ K, not K×fanout), plus aggregate ≪ per-target broadcasts.

## ⚠️ Design divergence to reconcile (needs Ian)

The spec's every-hop-**under-gauge** approach diverges from three older docs/PRs that **remove** every-hop hosting:
- **PR #4664** (`feat/remove-relay-caching`, piece E, BLOCKED) gates the PUT store to `originator_loopback || finalize_here || already_hosting` — i.e. it *removes* store-at-every-hop.
- **`hosting-invariants.md`** E-row and **`demand-driven-hosting.md`** (line 200) frame E as "remove relay-caching," relying on D's subscription chains + a findability workstream (bounded backtracking + a fresh PUT seed-chain).

`hosting-spec.md` (authoritative, read-first) supersedes this: every-hop hosting is **retained** as the findability mechanism ("many nearby landing spots for free"), and the R1+R2 demand gauge — not removal — is what makes it acceptable (evictable, non-durable). On `main` today, every-hop hosting is present and, per the spec, correct.

**This is not a blocker for this PR** (the spec + the build task both direct every-hop-under-gauge), but the epic (#4642), the invariants doc, and PR #4664 should be reconciled so the workstream isn't pulling in two directions. Flagging per the surface-divergence habit.

## Testing

- `cargo build -p freenet`, `cargo clippy --locked -p freenet --all-targets -- -D warnings`, `cargo fmt --check` — clean.
- Both new falsifiers pass, deterministically across repeated runs (identical numbers: fanout holders=12 / max-sends=5; summarize total=91 / max-per-peer=13 / 510 per-target broadcasts). `cargo test -p freenet --features "simulation_tests,testing" --test simulation_integration test_every_hop`.
- `test_piece_e_findability_sparse_ring_gate` (falsifier a) is unaffected — these changes are additive (a sim-scoped telemetry counter + two new tests) and touch no placement/routing/hosting behavior.

## Do NOT merge

Draft. Step 8 is part of the **inseparable relay-caching-free bundle** and ships only after R1+R2 field-validates and the whole bundle validates together (Delta churn-sim, relay-caching-free). Full multi-model review is required before the bundle merges.

Refs #4642, `hosting-spec.md` step 8.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime/executor_impl.rs
+++ b/crates/core/src/contract/executor/runtime/executor_impl.rs
@@ -1271,6 +1271,23 @@ where
                 })
             })?;
 
+        // Summarize-storm falsifier (spec step 8 / #4440): count the actual WASM
+        // `summarize_state` invocation — the SLOW-path miss the state-hash cache
+        // above exists to elide. A cache HIT (the fast path near the top of this
+        // fn) does NOT reach here, so this counter measures exactly the expensive
+        // work whose per-heartbeat × per-neighbor multiplication was the storm.
+        // Under the working cache it scales with the STATE-CHANGE rate, not with
+        // hosted-set size or neighbor overlap — every-hop placement's "summarize
+        // load stays flat vs hosted-set size" invariant. No-op outside a sim (the
+        // record fn reads the sim-only network-name thread-local). This runs on
+        // the contract-handling loop thread, not a spawn_blocking closure, so the
+        // thread-local is set.
+        if let Some(op_manager) = &self.op_manager {
+            if let Some(own_addr) = op_manager.ring.connection_manager.get_own_addr() {
+                crate::ring::topology_registry::record_summarize_wasm_call(own_addr);
+            }
+        }
+
         let summary = self
             .runtime
             .summarize_state(&key, &params, &state)
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -163,10 +163,11 @@ pub mod dev_tool {
     pub use ring::topology_registry::{
         ContractSubscription, ProximityViolation, RenewalMetrics, TopologySnapshot,
         TopologyValidationResult, aggregate_renewal_metrics, clear_all_topology_snapshots,
-        clear_current_network_name, clear_renewal_metrics, clear_topology_snapshots,
-        get_all_renewal_metrics, get_all_topology_snapshots, get_current_network_name,
-        get_renewal_metrics, get_topology_snapshot, register_topology_snapshot,
-        set_current_network_name, validate_topology, validate_topology_from_snapshots,
+        clear_current_network_name, clear_renewal_metrics, clear_summarize_metrics,
+        clear_topology_snapshots, get_all_renewal_metrics, get_all_summarize_wasm_calls,
+        get_all_topology_snapshots, get_current_network_name, get_renewal_metrics,
+        get_topology_snapshot, register_topology_snapshot, set_current_network_name,
+        validate_topology, validate_topology_from_snapshots,
     };
     pub use wasm_runtime::secret_export::{
         BundleKeyMaterial, ExportError, ImportReport, TargetScope, export_bundle, import_bundle,
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -515,6 +515,16 @@ pub struct ControlledSimulationResult {
     /// to/from the crashed node was subsequently blocked. `0` when no crash was
     /// scheduled. See #4642 piece F.
     pub crash_packets_dropped: u64,
+    /// Per-peer count of WASM `summarize_state` invocations (summary-cache
+    /// SLOW-path misses) captured at the end of the run, keyed by socket
+    /// address. Captured here (before the `SimNetwork` is dropped, which clears
+    /// the registry) so the every-hop-placement summarize-storm falsifier can
+    /// assert this stays flat as the hosted set / neighbor overlap grows. A
+    /// cache hit does NOT increment, so a working cache keeps this proportional
+    /// to the state-change rate, not to hosted-set size. See
+    /// `crate::ring::topology_registry::record_summarize_wasm_call` (#4440, spec
+    /// step 8).
+    pub summarize_wasm_calls: HashMap<std::net::SocketAddr, u64>,
 }
 
 #[cfg(any(test, feature = "testing"))]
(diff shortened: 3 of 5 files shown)


sanity:
_Rebased HEAD: `6d5599ee17c2c50c0f9b312e724161e3e5262737` (branch force-pushed after rebase + review fix)._

## Rebased onto `main` + review pass (prep for review, still not merging)

Rebased `feat/every-hop-lru-placement-p8` onto current `origin/main` (was ~69 commits behind). One conflict, in `crates/core/tests/simulation_integration.rs`: the PR appends its two falsifier tests at EOF and `main` had added many tests in the same region. Resolved by keeping all of `main`'s content and appending the two new tests once at the end. The source-side counter plumbing (`executor_impl.rs`, `topology_registry.rs`, `lib.rs`, `testing_impl.rs`) auto-merged cleanly; verified the counter still fires in `bridged_summarize_contract_state` on the slow path, immediately before the WASM `summarize_state`, after all cache-hit early returns.

### Tests (rebased tree) — both PASS, no bound changes needed

`cargo test -p freenet --features "simulation_tests,testing" --test simulation_integration test_every_hop -- --test-threads=1`

- `test_every_hop_fanout_bounded_by_degree_not_holder_count`: holders=12, max_sends_per_update=5 (= max_connections), < holders=12. **Structural claim holds.**
- `test_every_hop_summarize_calls_flat_vs_fanout`: total_summarize=91, max_per_peer=13 (≤ K+8=20), per-target broadcasts=510.

Emergent numbers are unchanged from the original PR body despite the intervening broadcast-path moves (Source-2 removal #4642 step 9, summary-first PUT, reconcile flip), so no assertion bounds were retuned.

### External review (Codex) — findings + resolutions

**P1 — summarize falsifier over-claims that it falsifies the summary cache.** Verified and correct. In the `#[cfg(feature = "simulation_tests")]` fan-out path (`broadcast_state_to_peers`) the source computes its summary via `get_contract_summary` → `bridged_summarize_contract_state` **once per broadcast emission**, before the per-target loop (the loop only reads `get_peer_summary`). So in the sim the counter rises ≈ once per emission regardless of fan-out width, and the observed `total_summarize=91` ≈ the broadcast-emission count, not 510. Removing the state-hash cache would leave the two bounds green — the once-per-broadcast structure, not the cache, produces the 91≪510 gap.
  - **Resolution:** corrected the test's doc/comments (comment-only) to state what it actually guards — per-peer summarize scales with the state-change rate / broadcast-emission count, **not** fan-out width; a regression that summarized per-target on the fan-out path (the production `broadcast_to_single_peer` shape) would blow the bounds. The state-hash summary cache regression itself is **already** directly falsified in CI by `summarize_unchanged_skips_state_load` / `delta_unchanged_skips_state_load` in `contract::executor::pool_tests::summarize_delta_cache_tests` (`#[cfg(test)]`): remove the cache → an unchanged re-summarize reloads state and those asserts fail. Cross-referenced from the test doc. So the cache is guarded; the sim test's headline was simply mislabeled.

**P2 — non-vacuity floor (`K×N`) does not prove exactly K client updates.** Valid but non-blocking: the test clears the floor with large margin (510 ≫ 72; total 91 ≫ floor 6) and is deterministic. A tightening (assert exactly K successful client updates) is a reasonable follow-up.

**P2 — fanout test measures selected targets (`broadcast_to`), not successful sends (`broadcasted_to`).** Not a defect for its claim: "per-peer fan-out bounded by co-host degree" is exactly a statement about the selected target set (a peer can't select more targets than its co-host neighbors). Holder-count + nonzero-broadcast guards prevent vacuity.

**P2 — counter is not literally production-free / doesn't count all WASM summaries.** NIT. On a production cache **miss** the path takes one `own_addr` mutex read + a TLS read before a full WASM `summarize_state` — negligible on the already-rare slow path. Not counting the direct `runtime.summarize_state` calls (PUT/UPDATE apply paths) is by design: the metric targets the broadcast/heartbeat summary path.

Codex found no seed/HashMap-order nondeterminism (target sets are sorted; count/sum/max are order-insensitive).

**Re-review note:** the only change since the Codex pass is the comment-only doc correction above (no logic, no assertion, identical test numbers), which directly implements Codex's P1 recommendation — trivial scope per the re-review exemption, so the external pass was not re-run.

[AI-assisted - Claude]

Issue #3465, opened 2026-03-07
Contract PUT operations report timeout/failure but succeed locally; propagation unreliable


sanity:
## Summary

When publishing contract updates via `fdev publish`, PUT operations frequently report failures (timeout, "peer connection dropped") even though the state is applied locally. The updated state then fails to propagate reliably to other peers in the network.

## Symptoms

1. **PUT responses are misleading**: `fdev publish` reports "PUT operation timed out" or "peer connection dropped", but the state IS applied to the local node. This was confirmed by immediately retrying with version+1 and getting "version must be higher than current version N" — proving version N was accepted.

2. **Propagation failures**: Update streams to remote peers are cancelled mid-transfer:
   ```
   ERROR freenet::operations::update: Failed to assemble stream for broadcast 
     tx=01KK4N5NPHVNH5GSDGSZP3V204 stream_id=2148383728 error=stream was cancelled
   ```

3. **Stale state persists on peers**: After a peer restarts and re-GETs the contract from the network, it receives old state because the update hasn't propagated widely enough. The peer then has no subscription, so it never receives the update broadcast.

4. **Mixed state versions across network**: The gateway received GET responses with different `total_size` values (815KB, 966KB, 983KB, 4.3MB) from different peers — indicating peers hold different versions of the same contract.

## Reproduction

1. Publish a contract update on a gateway node via `fdev -p 7509 publish`
2. Observe timeout/connection-dropped errors
3. Check a peer node — it still has the old state
4. The peer's GET operations to fetch the new version also time out
5. Only a direct `fdev publish` to the peer fixes it

## Environment

- Freenet version: 0.1.174 (174d1a3358fd)
- Contract: `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv` (River web container)
- Gateway: nova (5.9.111.215:31337)
- Peer: technic (behind NAT)

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-07-24T12:40:12.304013403+00:00
Issue hash: `1fb78c75`

## Value Assessment

[...]

PR #4733, opened 2026-07-08
feat(put): summary-first PUT probe->dispatch behavioral flow (#4642, step 3-bis)

sanity:
**DRAFT — wire/protocol + operation-state-machine change. FULL multi-model review + Ian sign-off required before merge. Ships in the R3 bundle, gated behind R2 field-validation (see `.claude/rules/hosting-invariants.md`). Do NOT merge.**

Stacked on #4718 (inert wire scaffolding). This PR contains two commits: `75c6762` is #4718's scaffolding rebased onto current `main` (review it in #4718); `f9815a7` is this PR's work (wire finalization + the behavioral driver + telemetry + tests). Once #4718 merges I will rebase to drop the duplicated scaffolding commit.

## Problem

Today every PUT ships full `WrappedState` hop-by-hop and every hop stores and hosts unconditionally (the relay-caching anti-pattern the redesign removes). For a contract the mesh already holds, a PUT re-ships the full body (up to ~650KB) to peers that do not need it. Spec step 3-bis (`~/code/tmp/hosting-spec.md`, section "The PUT protocol (summary-first)") replaces this: probe toward the key with a summary, ship full state only for a genuinely-new contract, reconcile a delta for one already held.

## Approach

- **Probe.** The originator (version-gated) stores the PUT state locally (a real client PUT is genuine local demand, so the publisher becomes a host), computes the contract-defined `StateSummary`, and routes a `PutMsg::ProbeRequest` toward the key. Intermediate non-holder hops route the probe onward **without storing** (lazy host-on-next-read, the Fable F5 amplification fix); only a **fresh** holder answers, detected via `Ring::is_receiving_updates` (subscribed / client-subscribed), **not** `is_hosting_contract` (mere cache membership, which can be stale).
- **Dispatch.** No holder found -> fall through to the existing full-state `PutMsg::Request` path (hops store+host; this is where every-hop placement gets its bytes). Holder found -> the reply carries the holder's summary `S_H`; the originator computes a `StateDelta` (its contribution relative to `S_H`) and ships it as `PutMsg::ProbeReconcile` (**delta only**, never full state into a mesh that already holds the contract). The holder merges via the UPDATE path and normal fan-out propagates it.
- **Graceful fallback (risk bound).** The existing full-state PUT path stays the default and the fallback. Any probe failure — version gate closed, probe timeout/error, no holder — falls straight through to the **unchanged** retry-loop driver, so summary-first can never break a PUT.
- **Version gate.** Emission is gated on `SUMMARY_FIRST_PUT_MIN_VERSION`, **corrected from the scaffolding's `(0,2,93)` to `(0,2,94)`**: 0.2.93 is already released carrying only the inert #4718 variants (a 0.2.93 peer can *decode* a probe but has no handler to *process* one), so emitting to it during the 0-4h staggered rollout would break its PUT. The floor must equal the release that ships the handler; a release-time verification note is at the constant. Peer versions are mirrored into `ConnectionManager` (parallel to `location_for_peer`) so the op driver can gate; the SubscribeHint send-gate precedent is mirrored throughout.
- **Telemetry (falsifier).** PUT-bytes-by-case via a single call site feeding both `GlobalTestMetrics` (sim) and `network_status` (dashboard): new-contract full-state vs existing-mesh delta bytes.

## Testing

- **Wire:** variant-tag pin extended to 9 tags; per-variant bincode roundtrips (`ProbeResponse` with/without `holder_summary`, `ProbeReconcile`).
- **Version gate (mixed-version):** fail-closed unit tests including the exact `0.2.93` decode-but-cannot-process boundary — an emitter must send a full-state PUT to a sub-floor peer, never a probe. (The sim harness shares one compiled version across all nodes, so per-peer mixed-version is asserted at the unit level, which is where the emission decision lives.)
- **Driver source-scrape pins:** the probe path performs no local store (lazy host-on-read); holder detection uses `is_receiving_updates` not `is_hosting_contract`; reconcile merges via `update_contract`.
- **Sim (`run_controlled_simulation`):** holder-found PUT ships a delta and no full state to existing hosts; no-holder PUT ships full state hop-by-hop. *(Being added; `run_simulation_direct` drives only random events and cannot script PUT-at-A / PUT-at-B, so the scripted controlled runner is the correct vehicle — see `.claude/rules/testing.md`.)*

## Design forks for review (Ian)

1. **Version floor (0,2,94)** — must equal the actual first-emitting release; re-confirm at release time if the number slips.
2. **Reconcile is write-forward only for now.** The initiator ships its delta to the holder; the *reverse* leg (initiator pulling the holder's delta to converge its own copy) is deferred to lazy host-on-read, consistent with F5. The spec calls the reconcile "bidirectional"; the reverse leg is a cheap, additive follow-up (`ProbeResponse.holder_summary` already carries what the initiator needs). Flagging in case you want it in this piece.
3. **Originator stores locally before probing** so it can summarize/delta by key and becomes a host (publisher = demand). This is a behavior change for the summary-first path only; the non-summary-first (old-target) path is byte-identical to today.

Refs #4642, spec step 3-bis.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3412,6 +3412,15 @@ std::thread_local! {
     static GLOBAL_TERMINAL_CONSULT_HITS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_TERMINAL_CONSULT_RESOLVED_FOUND: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_TERMINAL_CONSULT_STILL_NOT_FOUND: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // Summary-first PUT (#4642 step 3-bis) — PUT-bytes-by-case falsifier.
+    // Count + bytes for the new-contract case (no holder found; full state
+    // ships hop-by-hop via the existing `PutMsg::Request` path).
+    static GLOBAL_PUT_PROBE_NEW_CONTRACT_COUNT: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_PUT_PROBE_NEW_CONTRACT_BYTES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // Count + bytes for the existing-mesh case (holder found; only a
+    // `StateDelta` ships via `ProbeReconcile`).
+    static GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_COUNT: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_BYTES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
 }
 
 /// Global test metrics for tracking events across the simulation network.
@@ -3451,6 +3460,10 @@ impl GlobalTestMetrics {
         GLOBAL_TERMINAL_CONSULT_HITS.with(|c| c.set(0));
         GLOBAL_TERMINAL_CONSULT_RESOLVED_FOUND.with(|c| c.set(0));
         GLOBAL_TERMINAL_CONSULT_STILL_NOT_FOUND.with(|c| c.set(0));
+        GLOBAL_PUT_PROBE_NEW_CONTRACT_COUNT.with(|c| c.set(0));
+        GLOBAL_PUT_PROBE_NEW_CONTRACT_BYTES.with(|c| c.set(0));
+        GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_COUNT.with(|c| c.set(0));
+        GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_BYTES.with(|c| c.set(0));
     }
 
     /// Records that a ResyncRequest was received.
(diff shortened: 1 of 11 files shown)


sanity:
Update: sim tests landed (commit f0c29ca).

**Sim (`run_controlled_simulation`) — both pass, measuring the real distinction:**
- `test_summary_first_put_holder_found_ships_delta`: a subscribed holder exists; a second CRDT-mode PUT (higher version) from another peer takes the **existing-mesh delta path** (`put_probe_existing_mesh_delta_sends == 1`, `put_probe_new_contract_sends == 0`), delta = 144B strictly less than the 184B a full-state PUT would ship, **and the holder's stored state advances to v2** — proving the `ProbeReconcile` delta was received and merged end-to-end, not just counted at the originator.
- `test_summary_first_put_no_holder_ships_full_state`: a genuinely new contract into an empty network takes the **no-holder full-state fallback** (`put_probe_new_contract_sends == 1`) and a second peer ends up hosting — proving the legacy hop-by-hop path ran, not a silent no-op.

**Design finding worth Ian's eyes (not a regression):** `remote_version` is `None` on the joiner->gateway path — the gateway's `AckConnection` payload carries no version (`transport/peer_connection.rs:110`). So a node PUTting **through its gateway connection** fails the version gate closed and uses full-state. This is the **same pre-existing constraint SubscribeHint gates on** (`migration.rs:15` reads the identical `remote_version` source), so the emission gate follows the established precedent and degrades gracefully (full-state is always safe). But it means summary-first won't activate on the common node->gateway PUT path until the gateway's version is carried in `AckConnection` — a candidate follow-up if we want summary-first to reach that path. Summary-first still activates on PUTs routed to non-gateway peers whose versions are known (as the sim tests exercise from the gateway side).

CI gate status on this branch locally: `cargo build`, `cargo clippy --locked -p freenet -- -D warnings`, `cargo fmt --check`, full `--lib` suite (3834 passed), and the two new sim tests all green.

[AI-assisted - Claude]

sanity:
## Decision ratified: completion-ack = A (ack-at-seed)

Ian has ratified the summary-first PUT completion-ack semantics as **A (ack-at-seed)**: the client sees PUT success once the **originator has stored the state locally and dispatched the delta** (`ProbeReconcile`), **without waiting for the downstream holder to confirm the merge**.

### Why A (recorded so it isn't reopened as a default)

- **The originator is itself a real host of the contract.** Under every-hop hosting the originator stores the state on its own node before it dispatches, so at ack time the write is already durable on a live, mesh-joined host. "Success" therefore means *durably stored on a live host + delta dispatched*, which is fully satisfied at seed.
- **Convergence does not need a synchronous per-holder confirm.** Contract state is a commutative merge, and propagation to other holders is carried by live fan-out plus periodic InterestSync anti-entropy. A holder that merges the delta a moment later converges to the same state regardless of ack timing, so blocking the client on a per-holder merge confirmation buys no correctness.
- **B (ack-at-merge) is not even expressible in the current protocol.** The wire chain is `ProbeRequest -> ProbeResponse -> ProbeReconcile`; `ProbeReconcile` (the delta dispatch) has no reply variant. Ack-at-merge would require a brand-new post-merge reply variant and an extra round-trip. A avoids that entirely.

So A stands as a documented, deliberate choice, not an unexamined default.

### How this relates to today's full-state PUT (due-diligence)

Today's normal full-state PUT, in the common (forward-succeeds) case, stores locally and then fire-and-forgets the forward and **waits for a downstream `PutMsg::Response` to bubble back** before it acks the client. It already short-circuits to a **local success without downstream confirmation** when forwarding cannot be dispatched or there is no next hop (the #3465 / #4695 "locally-applied PUT is a success, propagation is best-effort" principle). Ack-A generalizes that same "local durable store = success" principle to the common case for summary-first PUT, rather than inventing new durability semantics. It is a modest earlier-ack relative to the full-state happy path, and it is consistent with the design invariant that a stored copy on a live host is a real host.

### Known follow-ups (tracked, not blockers for this PR)

1. **Version-floor bump at release time.** `SUMMARY_FIRST_PUT_MIN_VERSION` is currently the placeholder `(0, 2, 94)` and fails closed on an unknown remote version. When the behavioral flow actually ships, this floor must be bumped to the release that first carries the probe wire variants, so a peer that cannot decode `ProbeRequest` is never sent one.
2. **Gateway-link `remote_version = None` narrow-activation.** `supports_summary_first_put` fails closed when the peer's negotiated version is unknown, and gateway links do not currently record a remote version, so summary-first PUT stays off over gateway links (it falls back to full-state). Candidate follow-up: carry the gateway version in `AckConnection` so summary-first can activate over gateway links too.

### PR state

Rebased onto current `origin/main` (past #4700 disk accounting, #4702 PUT disk-budget gate, #4739 release 0.2.94, #4734 evict-subscribed, #4695 locally-applied-success). The overlapping files (`config.rs`, `network_status.rs`, `op_ctx_task.rs`) resolved purely additively, so every intervening change is preserved. Re-verified on the rebased head: `cargo fmt` clean, `cargo clippy --locked -p freenet -- -D warnings` clean, the two summary-first sim tests pass, and `put_msg_wire_format_variant_tags_are_stable` still passes (probe tags 6/7/8 appended after the byte-identical existing tags 0-5).

[AI-assisted - Claude]

PR #4734, opened 2026-07-08
fix(hosting): shed subscribed contracts + tear down memory on eviction

sanity:
## Problem

The shipped subscriber-primary eviction (#4720, 0.2.93) **hard-pins any subscribed contract**: `evict_over_budget` filters candidates to `subs == 0` under `AtCapacity` (the only pressure production passes), so a peer that is over budget with only subscribed contracts cannot shed anything and the memory bound stops being real. It also lumps local-client and downstream subscribers into one `subscriber_count`, so it cannot express Ian's confirmed ordering (local-subscribed evicted **last**). And it has **no memory-teardown**: even the (unwired) OOM valve would drop the cache entry but leave the subscription state, so `contract_in_use` stays true and `reclaim_evicted_contract` skips the disk delete forever — the entry vanishes but memory never frees.

## Approach

Ian's confirmed model: eviction is ONE ordering, victim ascending by `(local_subscription_count, downstream_subscriber_count, real-GET recency, key bytes)`. Local-subscribed is evicted **last** but **not** hard-pinned; in the all-subscribed extreme the fewest-subscriber contract **is** shed. `min_ttl` stays the cold-start floor. No separate OOM valve, no admission gate, no hard pin.

- **`victim_order` / `evict_over_budget`** (`cache.rs`): split the combined count into the two dimensions; drop the `subs == 0` `AtCapacity` filter (eligibility is now just `age >= min_ttl`); return `EvictedContract` carrying a `was_in_use` flag captured **atomically** with the decision. `Overflow` is kept only as the `min_ttl`-pierce corner of the SAME ordering (not a separate valve), trigger still unwired.
- **In-use teardown** (`hosting.rs`): `record_contract_access` and `sweep_expired_hosting` call `teardown_evicted_in_use_contract` for each `was_in_use` victim — clearing downstream subscribers, client subscriptions (only non-empty in the all-local last resort; strands the client, **accepted** last-resort), and the upstream lease — so `contract_in_use` is false **before** the caller reclaims and the disk state actually frees. Driven by the atomic flag, not a racy post-hoc `contract_in_use` re-read, so a contract that gained a subscriber *after* a zero-sub eviction is left alone (the re-host / re-subscribe guards in `RuntimePool::remove_contract` handle that race). Downstream re-home rides the existing interest-gated renewal loop — **no new signal built** (out of scope).
- **`subscribed_evictions_total`** falsifier counter wired to `RouterSnapshot` / OTLP telemetry: the OOM-valve counter stays 0 until its trigger is wired, so shedding subscribed contracts (which is now live under `AtCapacity`) would otherwise be invisible in the field.
- **`admission`** (#4717, inert / not wired to production) reuses `victim_order` via a `local = 0` shim — its model tracks only a combined count, and `(0, n, seq, key)` orders identically to the old `(n, seq, key)`, so its pin tests still pass unchanged.

## Testing

Every behavioral change has a regression test (this is a `fix:` PR, CI requires new test functions):

- **Load-bearing:** `record_contract_access_sheds_and_tears_down_only_subscribed` — over budget with ONLY subscribed contracts sheds the fewest-`(local, downstream)` one AND asserts `contract_in_use` flips false (so reclaim proceeds and memory frees), while only the victim is torn down.
- `evict_all_subscribed_sheds_fewest_as_last_resort` — fewest-local then fewest-downstream; local-client contract evicted last; `subscribed_evictions_total` bumped, `oom_valve` untouched.
- `torn_down_eviction_not_redriven_by_contracts_needing_renewal` — the torn-down victim is not re-selected by renewal section 1/2.
- `router_snapshot_json_includes_subscribed_evictions_gauge` — falsifier reaches the OTLP body.
- Updated the #4720 hard-pin tests to the new model (subscribed = ordered last, not pinned), e.g. `victim_order_is_local_then_downstream_then_get_recency_then_key`, `evict_normal_drops_zero_sub_before_subscribed`, `test_downstream_subscribers_evicted_after_zero_sub`, `test_record_contract_access_orders_in_use_last`, `oom_valve_sheds_fewest_subscriber_piercing_min_ttl`.

Local verification green: `cargo build -p freenet`, `cargo clippy --locked -p freenet -- -D warnings`, `cargo fmt --check`, `cargo test -p freenet --lib` (3822 passed, 0 failed).

## Status

Part of the **0.2.94** demand-driven-hosting bundle. **DRAFT — do not merge.** High-risk hosting / OOM-safety surface: needs full multi-model review + Ian sign-off first. A prior scope trace confirmed this piece is flip-independent and buildable now.

Refs #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -3150,13 +3150,16 @@ async fn get_contract_summary(
 /// on-disk store — which is the only signal that distinguishes a phantom from a
 /// contract we can really summarize:
 /// - Phantom contracts (the storm) have no stored state, so they are skipped.
-/// - An evicted-but-in-use contract keeps its state ON DISK
-///   (`evict_over_budget` PINS `subscriber_count >= 1` entries under normal
-///   `AtCapacity` pressure — only the deliberately-unwired OOM valve under
-///   `MemoryPressure::Overflow` can pierce that pin — and
-///   `reclaim_evicted_contract` only deletes state once NOT in use), so
-///   `contract_state_present` is true and it KEEPS summarizing — which is why
-///   the gate reads the state store, not the in-memory hosting cache.
+/// - A subscribed contract we hold keeps its state ON DISK: under normal
+///   `AtCapacity` pressure `evict_over_budget` orders it LAST (shed only as a
+///   last resort when nothing with fewer subscribers is eligible), so while any
+///   fewer-subscriber contract exists it is retained and `contract_state_present`
+///   stays true, and it KEEPS summarizing — which is why the gate reads the
+///   state store, not the in-memory hosting cache. In the all-subscribed
+///   last-resort extreme where it IS shed, `teardown_evicted_in_use_contract`
+///   clears its subscription state so `contract_in_use` is false and
+///   `reclaim_evicted_contract` deletes the disk state — at which point it is no
+///   longer held and correctly stops summarizing.
 /// - The moment a contract's state is fetched/stored it summarizes again — no
 ///   loss of proactive repair for any contract we genuinely hold.
 async fn summary_if_hosted_or_in_use(
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -1192,10 +1192,10 @@ pub struct HostedContractEntry {
     /// Read accesses (GET/SUBSCRIBE) observed over this entry's residency.
     pub read_count: u32,
     /// Whether the over-budget sweep would actually consider this contract for
-    /// eviction: past its `min_ttl` age gate AND not pinned by demand
-    /// (`contract_in_use`). The renderer badges "next to evict" on the first
-    /// eligible row, NOT the raw lowest-keep-score row — a within-TTL or
-    /// in-use low-score contract is skipped by the real sweep, so badging it
+    /// eviction: NOT pinned by demand (`contract_in_use`). There is no longer a
+    /// `min_ttl` age gate (dropped 2026-07-08). The renderer badges "next to
+    /// evict" on the first eligible row, NOT the raw lowest-keep-score row — an
+    /// in-use low-score contract is ordered last by the real sweep, so badging it
     /// would mislead the operator.
     pub eviction_eligible: bool,
 }
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -1278,6 +1278,21 @@ async fn cache_contract_locally(
         op_manager.ring.mark_local_client_access(&key);
     }
 
+    // Sync the InterestManager for any subscribed contract the subscriber-primary
+    // eviction shed + tore down (#4642 invariant 3). Run BEFORE the
+    // `unregister_local_hosting` loop below so that call observes the now-zeroed
+    // subscriber counts and reports full interest loss (→ retraction via
+    // `removed_contracts`). Without this, ghost `interested_peers` /
+    // `peer_contracts` / `local_client_count` entries survive and mis-target
+    // UPDATE broadcasts / inflate upstream interest counts (PR #4734 Fix 1).
+    for teardown in &access_result.evicted_in_use_teardown {
+        op_manager.interest_manager.remove_evicted_in_use(
+            &teardown.key,
+            &teardown.downstream_peers,
+            teardown.local_client_count,
+        );
+    }
+
     let mut removed_contracts = Vec::new();
     for (evicted_key, expected_generation) in &access_result.evicted {
         if op_manager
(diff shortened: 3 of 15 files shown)


sanity:
## Full multi-model review — eviction rework (subscribed-eviction + teardown)

Reviewers: external **Codex** (gpt-5.5, `codex review --base origin/main`, ran clean on the single-commit diff) + two adversarial Claude lenses (skeptical, code-first). Every `file:line` verified against the checked-out code.

**Verdict: the core is correct — the ordering matches invariant 3, and the teardown genuinely closes the #4720 memory-free drift** (all three lenses independently confirmed `contract_in_use → false` before both reclaim gates, `was_in_use` captured atomically, not re-driven by `contracts_needing_renewal`, clean local/downstream split, `min_ttl` still floors AtCapacity, tests strengthened not gutted). But there are findings to resolve before this ships in 0.2.94, one confirmed by two independent reviewers.

### 1. [P2 — fix before merge] Teardown desyncs `InterestManager` (Codex + skeptical, independently)
`hosting.rs:1306-1311`. `teardown_evicted_in_use_contract` clears `downstream_subscribers` + `client_subscriptions`, but NOT `InterestManager` (`interested_peers` / `peer_contracts` / `local_client_count`), which the documented dual-tracking invariant (`hosting.rs:274`, `interest.rs:267`) requires kept in sync — it drives UPDATE broadcast targeting and upstream interest counts. The old #4720 code never hit this because every evicted contract was `subs == 0`; this PR makes subscribed contracts evictable, so their interest entries are now leaked. **It does not self-heal:** the normal reconcilers (`expire_stale_downstream_subscribers`, client-disconnect cleanup) iterate the very maps teardown just removed, so they never sync InterestManager — ghost interests persist until disconnect (downstream) or forever (local client), leaking memory *precisely when the peer is under memory pressure*. Fix: at the three eviction consumer sites (`get/op_ctx_task.rs:1282`, `put/op_ctx_task.rs:1933`, `ring.rs:2623`), for the `evicted_in_use` subset also call `interest_manager.remove_downstream_subscriber` / `remove_local_client`, plus a regression test asserting InterestManager counts are zero after a subscribed eviction (the current `torn_down_...` test only covers the HostingManager path).

### 2. [MEDIUM — needs a decision] `min_ttl` vs subscriber-primary priority inversion
`cache.rs:1113-1123`. AtCapacity eligibility is `age >= min_ttl` only (`min_ttl` = 8 min). So a *fresh zero-subscriber* contract hogging budget is ineligible, and if the only past-`min_ttl` eligible entries are local-subscribed, eviction sheds a **local-subscribed contract (stranding its client) while the fresh zero-sub junk survives** — an inversion of subscriber-primary intent, triggerable by a burst of fresh GET/PUTs on a small-budget node. This is a genuine conflict between two locked decisions (min_ttl cold-start floor vs subscriber-primary). Recommendation: subscriber-primary should win — a zero-subscriber contract should be evictable before a local-subscribed one even inside `min_ttl`. Flagged for Ian's call, not resolved here.

### 3. [MEDIUM — confirm/add] Op-scoped pinning appears absent
The R2 plan lists "op-scoped pinning" alongside subscriber-primary, but it's not in this diff. Confirm an in-flight GET/PUT/SUBSCRIBE cannot have *its own* contract evicted+torn-down mid-operation; if not guaranteed, add it.

### 4. [LOW — bundle-assembly] STRICT vs ANY demand signal
All three lenses: eviction uses the ANY signal (all downstream, no distance filter); the reconcile core uses STRICT (`has_local_client || has_farther_downstream`, strictly-farther only). On `origin/main` reconcile is shadow/inert so there's no live disagreement, and #4734 neither closes nor worsens it. But #4725's flip **wires the STRICT signal into production renewal/collapse** — so in the assembled bundle, renewal (STRICT) and eviction (ANY) disagree on closer-only-downstream contracts (renewal lapses it → stale; eviction retains it). Reconcile the two demand signals at bundle assembly before ship. Neither PR shows this in isolation.

### 5. [LOW — fix, AGENTS.md same-PR] Stale docs asserting the removed hard-pin
Three reader-facing blocks still say subscribed contracts are PINNED under AtCapacity — the exact opposite of this PR: `cache.rs:12-28` (module header), `cache.rs:864-871` (`record_access` rustdoc), `node.rs:3154` (summarize-gate rationale). Plus `should_retain`-framed field doc + test names, and a `test_downstream_subscribers_evicted_after_zero_sub` misnomer (the subscribed one isn't evicted in that test).

### 6. [LOW/INFO] Teardown drops the upstream lease locally only (no wire Unsubscribe) → upstream keeps fanning updates until its lease expires; self-heals. Narrow teardown race (a downstream added concurrently in the post-lock window is dropped); bounded, self-heals.

**Disposition:** draft, not merging. Finding 1 (InterestManager sync) + Finding 5 (docs) are being fixed now; Finding 2 (min_ttl inversion) needs Ian's decision; Finding 3 (op-scoped pinning) needs confirm/add; Finding 4 is a bundle-assembly reconciliation. Re-run this review on the fixed content.

[AI-assisted - Claude]

PR #4749, opened 2026-07-09
refactor(hosting): retire relay_put_replicate_forward (R3 piece E) — DO NOT MERGE, negative result

sanity:
> **DO NOT MERGE — NEGATIVE RESULT.** This is a build-ahead experiment for #4642 R3 piece E. It removes `relay_put_replicate_forward` and adds a strengthened convergence gate to test whether the demand-driven freshness mechanism covers what that hop did. **It does not.** CI is (correctly) red: the strengthened gate catches that the removal reintroduces the #4509 convergence split. Kept as a draft so the finding + the reusable convergence gate are reviewable. Needs a decision from @sanity.

## Problem

`relay_put_replicate_forward` (`put/op_ctx_task.rs`) fires a fire-and-forget full-state PUT one hop past the #4363 routing terminus, purely to keep the UPDATE broadcast tree connected (#4509). R3 piece E wants to retire it: in the *target* model a demand-having peer joins the update mesh via its own subscribe-chain + live fan-out + anti-entropy, and a peer past the terminus with no demand correctly should not host the contract (hosting is demand-driven, `hosting-invariants` invariant 2).

The removal was **ship-gated on a strengthened convergence sim-check** that GATES it. This PR builds the removal + the gate and runs the gate. **The gate fails.**

## What this changes

- **Remove** `relay_put_replicate_forward`, both call sites (non-streaming + streaming), the streaming `terminus_replicate_to` / `replicate_payload` plumbing that only served it, and flip the two source-scrape pin tests (`drive_relay_put_applies_terminus_guard_before_forwarding`, `drive_relay_put_streaming_applies_terminus_guard_before_next_hop`) to assert the function stays **absent**.
- **Strengthen** `test_six_peer_contract_lifecycle`: new `assert_state_convergence` reads each node's **ground-truth** final state from `MockStateStorage` (not log-scraped) and asserts (1) ≥N holders, (2) all holders byte-identical, and (3) the converged hash equals the **final-update (v50)** value — catching both a v20/v30 split and the subtler "all agree on a stale value" vacuous pass.
- **Add** two variants: `test_put_no_subscriber_update_convergence` (PUT unsubscribed + UPDATE) and `test_put_replicate_forward_removal_churn_convergence` (crash a mid-chain subscriber mid-lifecycle).

## Result (deterministic, `--features simulation_tests,testing`)

| Scenario | With `replicate_forward` (control) | Without it (this PR) |
|---|---|---|
| six-peer subscribed lifecycle | **PASS 3/3** | **FAIL 3/3** — 2/2 split: 2 peers @ **v20**, 2 peers @ **v30**; node-2's v40/v50 reached no one |
| churn (crash mid-chain subscriber) | n/a | **FAIL** — only 2 survivors hold it, both frozen @ **v20**; post-crash updates reached no survivor |
| no-subscriber (2-node reachable mesh) | n/a | PASS — live fan-out reaches the one directly-connected demandless copy, but does not exercise the multi-hop tree that fragments |

The six-peer failure is the **exact #4509 signature** the removed hop's own doc comment described: "contract on 4/6 peers, a 2/2 v20-vs-v30 split, node-2's updates reaching no one." The anti-entropy + live-fan-out mechanisms currently on `main` (interest-sync #3226/#2789/#4475, reconcile P6 #4725) do **not** heal it within the sim budget (240s sim + 90s post-op wait).

## Conclusion

The freshness mechanism does **not** fully cover `relay_put_replicate_forward`. **E cannot ship on this tree.** Options for @sanity:
1. Keep `relay_put_replicate_forward` until the freshness/re-root layer demonstrably covers multi-hop tree connectivity (then re-run this gate).
2. Land only the strengthened convergence gate (independently valuable — it PASSES on `main` with the hop present and would guard any future removal), and defer the removal.

The strengthened `assert_state_convergence` gate is the reusable artifact here regardless of the removal's fate.

Relates to #4642.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/put/op_ctx_task.rs
+++ b/crates/core/src/operations/put/op_ctx_task.rs
@@ -1279,22 +1279,17 @@ where
                     "PUT relay: chain descended here and next hop moves away; finalizing here (#4363)"
                 );
                 // Finalize the operation at this closest-seen node (#4363
-                // placement) but STILL replicate the contract one hop onward so
-                // the UPDATE broadcast tree stays connected (#4509 convergence).
-                // See `relay_put_replicate_forward` for why both are required.
-                if let Some(next_addr) = peer.socket_addr() {
-                    relay_put_replicate_forward(
-                        op_manager,
-                        next_addr,
-                        key,
-                        contract.clone(),
-                        related_contracts.clone(),
-                        merged_value.clone(),
-                        htl,
-                        new_skip_list.clone(),
-                    )
-                    .await;
-                }
+                // placement). The contract is NOT replicated one hop PAST the
+                // terminus: a peer past the routing terminus has no demand for
+                // the contract, so it correctly should not host it — hosting is
+                // demand-driven, not holding-driven (see
+                // .claude/rules/hosting-invariants.md, invariant 2). A
+                // demand-having peer joins the contract's update mesh via its
+                // own subscribe-chain + live fan-out + anti-entropy; the old
+                // fire-and-forget past-terminus replication
+                // (`relay_put_replicate_forward`, #4509) is retired in R3
+                // piece E (do NOT re-add — the freshness mechanism now covers
+                // what it did).
                 let hop_count = op_manager.ring.max_hops_to_live.saturating_sub(htl);
                 return relay_put_finalize_local(
                     op_manager,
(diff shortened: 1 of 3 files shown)


sanity:
Superseded finding: #4752 rebuilds this convergence gate WITHOUT the confound here (subscribers used a bare `SUBSCRIBE`, which the node rejects for uncached WASM, so they never actually joined the update mesh — the "split" measured PUT physical reach, not mesh convergence).

With an HONEST gate (16 peers, GET-with-subscribe so subscribers really join the mesh, verified via `is_receiving_updates`, ground-truth STATE convergence to v50, 15 seeds), removing `relay_put_replicate_forward` does **NOT** regress convergence on current main: control and removal both converge **15/15 (100%)**, in steady AND churn. So the removal this PR proposed is sound — but the "E cannot ship" conclusion here was a test artifact. See #4752 for the numbers.

[AI-assisted - Claude]

sanity:
Superseded by #4752 (honest convergence gate). This PR's "removal breaks convergence" result was a **test artifact** — its six-peer gate had subscribers issue a bare SUBSCRIBE, which the node rejects when the contract WASM isn't cached, so those subscribers never joined the update mesh; it measured PUT physical reach (which removing the hop shortens by one), not mesh convergence. #4752 (GET-with-subscribe → real mesh, verified via is_receiving_updates, 15 seeds, steady + churn) proves removing relay_put_replicate_forward converges 15/15 (100%), identical to keeping it. The hop is redundant and deletable **standalone**; the actual deletion will be a fresh PR guarded by #4752's gate. Closing this negative-result R&D draft. [AI-assisted - Claude]

PR #4750, opened 2026-07-09
feat(hosting): near-key replication prototype + convergence-vs-cost sweep (R&D — DO NOT MERGE)

sanity:
> **R&D PROTOTYPE — DO NOT MERGE.** Build-ahead experiment for the epic (#4642), exploring an ALTERNATIVE to draft PR #4749. Replaces the single past-terminus replication hop (`relay_put_replicate_forward`, #4509) with a principled, bounded, near-key replication cluster, and produces convergence-vs-cost numbers so the parameters can be chosen from data. Kept as a draft for the mechanism + the numbers. The mechanism-level pin tests pass; the measurement tests are `#[ignore]` (a parametrized harness, not a CI gate).

## Problem

Removing `relay_put_replicate_forward` deterministically reintroduced the #4509 convergence split on PR #4749's base: the past-terminus copy keeps the multi-hop UPDATE broadcast tree connected, and live-fan-out alone (which only reaches directly-connected co-hosts) does not reconstruct it. The direction was: replace the single fragile hop with a small **bounded near-key replication cluster** that forms a *connected* cluster near the key — decentralized, local-info-only, DoS-bounded.

## Approach (the prototype)

`nearkey_replicate_spread` (`put/op_ctx_task.rs`) replaces the single hop:

1. The routing terminus (a local fact: no neighbor closer to the key than self) MINTS a replication budget `R_MAX` and stores (copy #1).
2. It hands the remaining budget to its neighbors **within a ring-distance band `D` of the key** (a local check), split evenly. As a **connectivity floor** it always keeps the single nearest-to-key neighbor, so the terminus never no-ops and the tree stays connected. At `R_MAX = 2` this reduces to "terminus + nearest near-key neighbor" — i.e. the retired single-hop behavior.
3. Each recipient stores, spends 1, and RE-SPREADS the remainder to ITS near-key neighbors (carried in a new `replication_budget` field on `PutMsg::Request`), until budget hits 0. Because every copy lands at a DIRECT NEIGHBOR of an existing copy, the cluster is connected by construction.
4. **DoS bound:** the budget is locally CLAMPED to `R_MAX` on receipt; HTL is decremented; the skip list carries siblings so the cluster cannot loop or double-cover.

`R_MAX` and `D` are read from `FREENET_NEARKEY_R_MAX` / `FREENET_NEARKEY_BAND_D` at process start so a single test binary can be swept.

## Measurement

Ported the strengthened ground-truth STATE-convergence gate from #4749 (`nearkey_assert_state_convergence`: reads each node's final `MockStateStorage`, asserts ≥N holders + byte-identical + equals the final-update v50 hash), across three scenarios (subscribed six-peer, PUT-no-subscriber+UPDATE, mid-chain-subscriber crash), swept `R_MAX` 1→8 (R_MAX=1 = no spread = the removal negative control) and across seeds. A greppable `NEARKEY_MEASURE` line records copies + convergence for every run, including failures.

### Cost (deterministic, clean)

Copies placed per PUT = **`min(R_MAX + 1, N)`** — zero variance across seeds. `R_MAX=2` places exactly the same 2-copy footprint as the retired single hop. Update-fanout and memory scale linearly with copies (~1 MB resident/contract per the cost model).

### Convergence (single-seed sweep, default seed)

| R_MAX | subscribed (min 4) | no-subscriber (min 1) | churn (min 3) |
|---|---|---|---|
| 1 (removal) | 6 copies, converge | 2, converge | 2, **split** |
| 2 | 3, **fail** | 3, converge | 2, **split** |
| 3 | 4, converge | 4, converge | 3, **split** |
| 4 | 5, converge | 5, converge | 4, **split** |
| 5 | 5, converge | 6, converge | 4, **split** |
| 6 | 6, converge | 6, converge | 5, converge |
| 7 | 6, converge | 6, converge | 5, converge |

### The snags (why this does NOT cleanly replace `relay_put_replicate_forward`)

1. **The 6-peer convergence gate is scheduling-noise-dominated.** The negative control (R_MAX=1 ≈ removal) converges on some seeds and not others for ALL three scenarios (subscribed baseline holder count ranges 2→6 across seeds; churn converges at R_MAX=1 on some seeds and not the default). Single-seed sweeps are not reliable, so **this data cannot recommend an `R_MAX`**. The clean signal is cost, not convergence.

2. **The subscribed gate is confounded.** The test's subscribers issue a BARE `SUBSCRIBE`, which is rejected unless the contract is ALREADY cached locally (`"Rejecting SUBSCRIBE: contract WASM not cached locally"`). So the subscribed holder count measures "did a PUT copy reach this node before its subscribe fired," not update-mesh convergence. With the spread active, PUT-copy distribution is regularized to exactly the near-key cluster (`R_MAX+1`) deterministically — so the subscribed metric primarily reflects PUT-copy-vs-subscribe timing, not the mechanism's convergence effect. This confound also affects #4749's gate.

3. **The removal is NOT deterministically broken on current main.** Unlike PR #4749's older base, R_MAX=1 (removal) converges the subscribed six-peer on the default seed here. The newer reconcile/anti-entropy work merged since #4749's base (#4725 P6 flip, #4734) appears to cover the subscribed multi-hop case — so the premise that `relay_put_replicate_forward` is still needed should be re-verified before either removing OR replacing it.

4. **The one real signal is churn robustness — but it needs ~full replication at this scale.** Crashing a mid-chain subscriber strands its downstream chain. Across 5 seeds, R_MAX=6 converged the survivors **5/5** vs R_MAX=1 (removal) **3/5** — so higher R_MAX genuinely improves churn-convergence robustness. But R_MAX=6 in a 6-peer network is essentially "replicate to everyone," so the 6-peer test cannot distinguish "needs a small bounded R_MAX" from "needs R_MAX ≈ N". This is exactly the question a larger network would answer, and it's the strongest reason to re-run at scale before deciding. (The underlying crashed-subscriber gap is also addressable directly by event-driven re-subscribe / re-root — piece F — which is orthogonal to PUT-time replication.)

## Recommendation

**This data does not yet justify adopting near-key replication — but the churn-robustness signal is worth pursuing at scale.** The mechanism is sound and cheap (deterministic bounded cost, `R_MAX=2` = the status quo), the one genuine benefit (churn robustness, 5/5 vs 3/5) needs a bigger network to tell whether a *small bounded* R_MAX suffices, and the specific regression it targets may already be covered on current main. Before any parameter call:
- Rebuild the convergence gate to use **GET-with-subscribe** (not bare SUBSCRIBE) and a **larger, multi-contract** network with **multi-seed aggregation**, to get a signal that isn't dominated by PUT-reach timing and scheduling noise.
- Re-verify whether `relay_put_replicate_forward` removal actually regresses on current main (rebase #4749's exact removal) before deciding it needs replacing at all.
- Treat the churn/re-root gap as piece F (subscribe-chain repair), separately.

## Design decisions I had to make (flagged)

- **Budget minted at the terminus, not carried from the client.** Carrying `R` from the originator would make every relay hop short-circuit as a spread copy before the terminus. The DoS clamp is still demonstrable (spread copies carry `replication_budget`, clamped on receipt).
- **`R_MAX` counts the terminus** (cluster size incl. self); default 2 (= status quo). `D` default 0.30 ring distance; the connectivity floor makes the mechanism robust to `D` being too small.
- **Step 4 (near-key eviction keep-priority) NOT implemented.** It would reintroduce distance into the eviction ranking, which **contradicts settled hosting-invariant 3** (distance was removed from eviction, demoted to telemetry, 2026-07-08). It is also not needed for the convergence measurement. Implementing it needs an explicit design decision — flagged, not silently added.
- **Streaming-size payloads are skipped** in the spread (parity with the retired hop; the sim exercises the small-state non-streaming path).

## Testing

- `nearkey_replicate_spread` replaces `relay_put_replicate_forward` at both terminus call sites (non-streaming + streaming). The two source-scrape pin tests are updated to assert the spread is present (71/71 `op_ctx_task::tests` pass).
- Three `#[ignore]` measurement harnesses added (subscribed / no-subscriber / churn), parametrized by `FREENET_NEARKEY_R_MAX` / `_SEED` / `_BAND_D`. Run with `-- --ignored`.
- `cargo fmt` / `check` / `clippy` clean.

Relates to #4642.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1177,6 +1177,7 @@ where
                         value,
                         htl,
                         skip_list,
+                        replication_budget,
                     } => {
                         if let Err(err) = put::op_ctx_task::start_relay_put(
                             op_manager.clone(),
@@ -1188,6 +1189,7 @@ where
                             *htl,
                             skip_list.clone(),
                             upstream_addr,
+                            *replication_budget,
                         )
                         .await
                         {
@@ -4765,6 +4767,7 @@ mod tests {
                 value: WrappedState::new(vec![1u8]),
                 htl: 5,
                 skip_list: std::collections::HashSet::new(),
+                replication_budget: 0,
             };
 
             let taken = put_branch_would_forward(&op, Some(&tx));
--- a/crates/core/src/operations/put.rs
+++ b/crates/core/src/operations/put.rs
@@ -298,6 +298,18 @@ mod messages {
             htl: usize,
             /// Addresses to skip when selecting next hop (prevents loops)
             skip_list: HashSet<std::net::SocketAddr>,
+            /// R&D PROTOTYPE — near-key replication (#4642 R3 piece E
+            /// alternative). Remaining near-key replication budget carried by a
+            /// SPREAD copy. `0` for a normal client/relay PUT — the routing
+            /// terminus mints `R_MAX` locally and begins the spread. `>0` only on
+            /// the fire-and-forget replication copies emitted by
+            /// `nearkey_replicate_spread`: the budget is spent 1-per-hop and is
+            /// locally CLAMPED to `R_MAX` on receipt (the DoS bound — an attacker
+            /// inflating this field is capped). `#[serde(default)]` for
+            /// source-level clarity; bincode is positional, so this rides after
+            /// `skip_list` and all sim peers run the same build.
+            #[serde(default)]
+            replication_budget: usize,
         },
         /// Response indicating the PUT completed. Routed hop-by-hop back to originator
         /// using each node's stored upstream_addr.
@@ -604,6 +616,7 @@ mod tests {
                     value: WrappedState::new(vec![]),
                     htl: 0,
                     skip_list: Default::default(),
+                    replication_budget: 0,
                 },
             ),
             (
--- a/crates/core/src/operations/put/op_ctx_task.rs
+++ b/crates/core/src/operations/put/op_ctx_task.rs
@@ -305,6 +305,12 @@ async fn drive_client_put_inner(
                     .get_own_addr()
                     .into_iter()
                     .collect::<HashSet<_>>(),
+                // R&D PROTOTYPE: a client's initial PUT carries NO replication
+                // budget. The budget is MINTED fresh (`R_MAX`) at the routing
+                // terminus; carrying it from the originator would make every
+                // relay hop short-circuit as a spread copy before the terminus
+                // is reached. See `nearkey_replicate_spread`.
+                replication_budget: 0,
             })
         }
 
@@ -830,6 +836,10 @@ pub(crate) async fn start_relay_put<CB>(
     htl: usize,
     skip_list: HashSet<SocketAddr>,
     upstream_addr: SocketAddr,
+    // R&D PROTOTYPE (near-key replication): remaining near-key replication
+    // budget carried by a spread copy; 0 for a normal PUT. See
+    // `nearkey_replicate_spread`.
+    replication_budget: usize,
 ) -> Result<(), OpError>
 where
     CB: NetworkBridge + Clone + Send + 'static,
@@ -882,6 +892,7 @@ where
         htl,
         skip_list,
         upstream_addr,
+        replication_budget,
     ));
     Ok(())
 }
@@ -916,6 +927,7 @@ async fn run_relay_put<CB>(
     htl: usize,
     skip_list: HashSet<SocketAddr>,
     upstream_addr: SocketAddr,
+    replication_budget: usize,
 ) where
     CB: NetworkBridge + Clone + Send + 'static,
 {
@@ -931,6 +943,7 @@ async fn run_relay_put<CB>(
         htl,
         skip_list,
         upstream_addr,
+        replication_budget,
     )
     .await;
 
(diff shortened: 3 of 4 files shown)


sanity:
Follow-up on your snags 2 & 3: #4752 rebuilds the convergence gate the way you asked (GET-with-subscribe instead of bare SUBSCRIBE, 16 peers, multi-seed, ground-truth STATE convergence, mesh formation verified via `is_receiving_updates`).

Result: on current main, `relay_put_replicate_forward` removal converges **15/15 (100%)** identically to control, in both steady and churn — i.e. the past-terminus hop is NOT load-bearing for convergence, so near-key replication is **not justified by a convergence need** here. The demand-driven mesh (subscribe chains + live fan-out + anti-entropy) already converges 100% under a mid-mesh crash without any PUT-time replication. See #4752.

[AI-assisted - Claude]

sanity:
Superseded by #4752. Near-key replication is **not needed** — this prototype was premised on the (now-disproven) belief that removing relay_put_replicate_forward breaks convergence. #4752's honest gate shows the demand-driven mesh (subscribe-chains + live fan-out + anti-entropy, shipped 0.2.94) already converges 100% without the hop, including under a mid-mesh crash. Shelving this prototype; the clean answer is to just delete the hop. [AI-assisted - Claude]

PR #4752, opened 2026-07-09
test(hosting): honest UPDATE-mesh convergence gate — relay_put_replicate_forward is NOT load-bearing (R&D, DO NOT MERGE)

sanity:
> **R&D MEASUREMENT — DO NOT MERGE.** Build-ahead experiment for the epic (#4642). Decides whether removing `relay_put_replicate_forward` (#4509) actually regresses UPDATE convergence on current `main`, using an HONEST convergence gate that fixes the confound in draft PRs #4749 and #4750. Kept as a draft for the reusable gate + the numbers.

## Problem

Draft PR #4749 concluded that removing `relay_put_replicate_forward` reintroduces the #4509 convergence split, and #4750 built a near-key replication mechanism to replace it. But **both of their six-peer convergence tests were confounded**: the subscribers issued a bare `SUBSCRIBE`, which the node REJECTS unless the contract WASM is already cached locally (`"Rejecting SUBSCRIBE: contract WASM not cached locally"`, `client_events.rs`). So those subscribers never joined the update mesh (`is_receiving_updates` never became true) — the tests measured how far the PUT physically reached, NOT whether the update mesh converges. At six peers the signal was also scheduling-noise-dominated. #4750 explicitly flagged this and asked for a rebuilt gate.

## The honest gate

`honest_convergence_gate_{steady,churn}` in `simulation_integration.rs`:

1. **Subscribers really join the mesh** — GET-with-subscribe (`SimOperation::Get { subscribe: true }`) fetches+caches the WASM/state AND registers the subscription (`maybe_subscribe_child` → `run_client_subscribe` → downstream-subscriber edge). The gate then **verifies** the mesh formed via the new `ControlledSimulationResult::is_node_receiving_updates` (an active network OR local client subscription, NOT a mere cache copy) before trusting any convergence number. A run whose mesh did not form is reported INVALID, not "not converged".
2. **Larger network** — 2 gateways + 14 nodes = 16 peers.
3. **Ground-truth STATE convergence** — read each mesh member's final `MockStateStorage`; require all mesh holders byte-identical AND equal to the final update (v50). Catches a version SPLIT (unique>1) and a vacuous "all agree on a stale value" (at_final<holders). CRDT is higher-version-wins, so the converged value is deterministic.
4. **Multi-seed (15)** — report the convergence RATE per config, aggregated.
5. **Several updates from multiple sources** at varied ring positions, so the multi-hop broadcast tree is exercised.

Control vs removal run in the **same binary** with identical scheduling, toggled by the `FREENET_DISABLE_REPLICATE_FORWARD` env kill-switch (`put/op_ctx_task.rs::replicate_forward_disabled`): unset == pristine `main` (`if !false { forward }`), set == the forward never fires (behaviorally identical to deleting the hop + both call sites). The `#[ignore]` harnesses PRINT `HONEST_CONVERGE` summaries; they are measurement tools, not CI gates.

## Results (15 seeds each, deterministic, `simulation_tests,testing`)

| Config | mesh formed (subscribers actually `is_receiving_updates`) | mean mesh | STATE convergence (all mesh holders byte-identical @ v50) |
|---|---|---|---|
| **steady, control** (`replicate_forward` present = main) | 15/15 (1.00) | 16.0 / 16 | **15/15 (1.00)** |
| **steady, REMOVAL** (`replicate_forward` disabled) | 15/15 (1.00) | 16.0 / 16 | **15/15 (1.00)** |
| **churn, control** (crash a mid-mesh subscriber; ~650 pkts dropped/seed) | 15/15 (1.00) | 15.0 / 15 | **15/15 (1.00)** |
| **churn, REMOVAL** | 15/15 (1.00) | 15.0 / 15 | **15/15 (1.00)** |

Every seed in every config: the mesh fully formed (all 16 / 15 intended subscribers genuinely `is_receiving_updates`), zero state splits (`unique_hashes == 1`), every mesh holder at the final update `v50`, `converged = true`. Churn `crash_drops` totalled 9745 (control) / 9846 (removal) across the 15 churn seeds, confirming the scripted crash actually blocked traffic — the survivors still converged.

## Verdict

**Removing `relay_put_replicate_forward` does NOT regress UPDATE convergence on current `main`.** With a REAL mesh (subscribers actually receiving updates, verified — not the confounded bare-SUBSCRIBE proxy), control and removal converge **identically at 100%**, in both the steady and churn variants. `relay_put_replicate_forward` is therefore **not load-bearing for convergence on current `main`**: the update mesh is held together by the subscribe chains + live fan-out + anti-entropy (#4725 P6, #4734, interest-sync), not by the past-terminus PUT replication hop.

The #4749 "removal breaks convergence" result was a **test artifact** of its confounded six-peer gate: its bare-`SUBSCRIBE` subscribers were rejected for want of a locally-cached WASM, so they never joined the mesh — the split it saw was PUT physical reach shrinking by one hop, not the update mesh fragmenting. #4750 already suspected this (its snags 2 & 3) and asked for exactly this rebuilt gate.

### Recommendation

- **Delete `relay_put_replicate_forward`** (both call sites + the streaming `terminus_replicate_to` / `replicate_payload` plumbing) as #4749 proposed — but land it against THIS honest gate (flip these harnesses to hard asserts, or keep them as measurement), not #4749's confounded one.
- **The near-key replication cluster (#4750) is not justified by a convergence need** on current `main`. Its one genuine signal was churn-robustness at ~full replication in a 6-peer net; this 16-peer honest gate shows the demand-driven mesh already converges 100% under a mid-mesh crash WITHOUT any past-terminus replication, so there is no convergence gap for near-key replication to close here.
- Findability (can a distant GET still locate a contract after the every-hop durable copy is removed) is a **separate** property, already measured by the existing `test_piece_e_findability_sparse_ring_gate`; this gate is scoped to UPDATE-mesh convergence only and does not speak to findability.

Caveat on scope: this measures small-state (non-streaming) contracts, matching the retired hop's own documented scope (it skips streaming-size payloads). The mesh here is bootstrapped by explicit GET-with-subscribe; a pure cache-only holder past the terminus with no demand is (correctly) not a mesh member and is out of scope by design (hosting-invariants: hosting is demand-driven).

## Testing

- `honest_convergence_gate_steady` / `honest_convergence_gate_churn` (`#[ignore]`), run with
  `FREENET_DISABLE_REPLICATE_FORWARD=1` for the removal config.
- The env kill-switch preserves the two source-scrape pin tests
  (`drive_relay_put_applies_terminus_guard_before_forwarding`,
  `drive_relay_put_streaming_applies_terminus_guard_before_next_hop`) — the call text is retained,
  only wrapped in the runtime branch.
- `cargo fmt` / `check` / `clippy` clean.

Relates to #4642. Supersedes the convergence-gate portions of #4749 and #4750.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -487,6 +487,22 @@ impl ControlledSimulationResult {
             .is_some_and(|ring| ring.is_hosting_contract(key))
     }
 
+    /// Whether `label`'s node was actually in the UPDATE MESH for `key` at the
+    /// end of the run — i.e. `Ring::is_receiving_updates` (an active network
+    /// subscription OR a local client subscription), NOT merely holding a cache
+    /// copy. This is the honest mesh-membership signal the convergence gate uses
+    /// to VERIFY the mesh formed before it trusts a convergence measurement:
+    /// `is_hosting_contract()` is true for a demandless cache copy that is not
+    /// kept fresh, so it is NOT a valid mesh-membership check (see
+    /// `hosting-invariants` — only `is_receiving_updates()` guarantees a fresh,
+    /// update-carrying copy). Returns `false` if the node never published its
+    /// Ring.
+    pub fn is_node_receiving_updates(&self, label: &NodeLabel, key: &ContractKey) -> bool {
+        self.node_rings
+            .get(label)
+            .is_some_and(|ring| ring.is_receiving_updates(key))
+    }
+
     /// Number of contracts `label`'s node held in its hosting cache at the end
     /// of the run (its "cache size"). Returns 0 if the node never published its
     /// Ring. See [`Ring::hosting_contracts_count`](crate::ring::Ring::hosting_contracts_count).
--- a/crates/core/src/operations/put/op_ctx_task.rs
+++ b/crates/core/src/operations/put/op_ctx_task.rs
@@ -1283,17 +1283,22 @@ where
                 // the UPDATE broadcast tree stays connected (#4509 convergence).
                 // See `relay_put_replicate_forward` for why both are required.
                 if let Some(next_addr) = peer.socket_addr() {
-                    relay_put_replicate_forward(
-                        op_manager,
-                        next_addr,
-                        key,
-                        contract.clone(),
-                        related_contracts.clone(),
-                        merged_value.clone(),
-                        htl,
-                        new_skip_list.clone(),
-                    )
-                    .await;
+                    // R&D gate (NOT FOR MERGE): env kill-switch so the honest
+                    // convergence gate can measure control vs. removal in one
+                    // binary. Unset env == pristine main.
+                    if !replicate_forward_disabled() {
+                        relay_put_replicate_forward(
+                            op_manager,
+                            next_addr,
+                            key,
+                            contract.clone(),
+                            related_contracts.clone(),
+                            merged_value.clone(),
+                            htl,
+                            new_skip_list.clone(),
+                        )
+                        .await;
+                    }
                 }
                 let hop_count = op_manager.ring.max_hops_to_live.saturating_sub(htl);
                 return relay_put_finalize_local(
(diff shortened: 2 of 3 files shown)


sanity:
Closing — this R&D measurement PR (title: DO NOT MERGE) has delivered its verdict.

Finding: removing `relay_put_replicate_forward` does NOT regress UPDATE convergence — control and removal both converge 100% across 15 seeds. The earlier "#4749 removal breaks convergence" result was a test artifact.

The PRs it superseded (#4749, #4750) are already closed. Closing now that the finding is recorded; the 16-peer honest-convergence-gate harness on this branch can be revived if we decide we want it as a permanent test.

[AI-assisted - Claude]

PR #4758, opened 2026-07-10
refactor(put): remove redundant relay_put_replicate_forward hop (#4509)

sanity:
## Problem

The #4363 greedy-routing terminus guard finalizes a PUT at the closest-seen node, then fire-and-forgets one **extra replication copy one hop past the terminus** (`relay_put_replicate_forward`, added for #4509). That extra copy is a durable-regardless-of-demand remnant: it places a copy on a peer with no demand for the contract, purely to widen PUT physical reach. It is exactly the "durable-regardless-of-demand hosting" anti-pattern the demand-driven-hosting redesign (#4642) is removing.

The #4752 honest UPDATE-mesh convergence gate proved the hop is redundant: 16 peers, subscribers using GET-with-subscribe so they **actually join the mesh** (verified via `is_receiving_updates`), ground-truth STATE convergence, 15 seeds, steady + churn. Removal converges **15/15 — identical to keeping the hop**, including under a mid-mesh crash. The demand-driven mesh (subscribe-chains + live fan-out + anti-entropy, all shipped in 0.2.94) already carries UPDATE convergence; the hop is not load-bearing.

The earlier "removal re-opens the #4509 divergent-state split" belief was a test artifact: the #4749 six-peer gate used **bare** `SUBSCRIBE`, which the node rejects when the WASM isn't cached locally, so its subscribers never joined the mesh — it measured PUT *physical reach* (which the hop extends by one), not mesh *convergence*.

## Approach

Delete the function, both call sites, its dead plumbing, and correct the now-false #4509 documentation. Rely on a committed honest convergence gate (below) plus the #4752 R&D proof.

Removed:
- `async fn relay_put_replicate_forward` and its #4509 doc block
- the non-streaming terminus call site
- the streaming path's `terminus_replicate_to` plumbing + Step 6 call site
- the now-false #4509 "must still replicate / divergent-state split" claims in the two terminus-guard pin tests — the #4363 **guard-order and finalize-local assertions are kept intact**
- stale #4509 narrative references naming the deleted function in `tests/nat_subscription.rs`

No wire-format change.

## Testing

- `cargo build`, `cargo fmt`, `cargo clippy -p freenet -- -D warnings` all clean (no dead-code warnings from the removal).
- Put unit tests: **101 passed**, including the two edited terminus-guard pin tests.

**Committed regression proof — honest PUT-mesh convergence gate** (`honest_put_mesh_convergence_{steady,churn}` in `tests/simulation_integration.rs`). Because the #4752 honest gate lives only on the unmergeable `rnd/honest-convergence-gate` R&D branch (env kill-switch, `#[ignore]` measurement), this PR ports a mergeable, **asserting** distillation of it, stripped of the scaffolding (the hop is deleted, so it runs against the real code):

- PUT-routes the contract through a 16-peer network (2 gw + 14 nodes) — exercising the exact #4363 terminus path the deleted hop sat on.
- Forms the mesh via **GET-with-subscribe** (subscribers genuinely join; a bare `SUBSCRIBE` is rejected until the WASM is cached — the #4752 confound).
- **Verifies every survivor is `node_is_receiving_updates`** (real network/client subscription, not a mere cache copy) *before* measuring — a run whose mesh didn't form fails there.
- Asserts ground-truth **STATE convergence** to the final update (v50): all mesh members hold state, byte-identical, all == v50.
- **Steady + churn** (mid-mesh subscriber crash) variants; CI-weighted to 2 seeds each (~2 min/variant). Passes on this branch (hop deleted): all mesh members converge to v50 in every run.

Audit of why the pre-existing convergence sims did **not** already cover the removal (which is why the new gate was needed):

| Test | Mesh formation | Convergence assertion | Honest for this removal? |
|---|---|---|---|
| `test_pr2763_crdt_convergence_with_resync` | bare `SUBSCRIBE` | asserts | No — bare-SUBSCRIBE confound (measures PUT reach) |
| `test_crdt_convergence` (44 cases) | `TestConfig::medium` fuzz, mixed | asserts | No — fuzz + bare SUBSCRIBE |
| `test_partition_heal_convergence` | random-op fuzz | only logs anomalies | No — no hard convergence assert |
| `test_crash_recover_convergence` | random-op fuzz | only logs anomalies | No — no hard convergence assert |
| `test_piece_e_unsubscribed_stale_serve_gate` | GET-with-subscribe | asserts mesh converges | Honest, but **seeds** the contract (`SeedHostedContract`), never PUT-routes |

Other removal-relevant sim tests pass without the hop: `test_piece_e_unsubscribed_stale_serve_gate` (anti-entropy heals a demandless copy — the safety net), `test_summary_first_put_*`, `test_terminal_advertisement_consult_closes_subscribe_dead_end`, `test_concurrent_updates_convergence`, `test_partition_heal_convergence`, `test_crash_recover_convergence`, `test_pr2763_crdt_convergence_with_resync`. `test_crdt_convergence` (44 cases): 4/4 dedicated batch reruns pass 44/44; one earlier mixed-batch run had a single transient `case_07` divergence that did not reproduce in isolation or on reruns (that suite is bare-SUBSCRIBE, with a known parallel-load noise floor).

**Merge held pending v0.2.95 field-validation.** Per the hosting redesign ordering, the demand gauge (subscriber-primary eviction, shipped 0.2.94) must be field-validated before this durable-copy removal lands.

Closes #4509
Refs #4752 (honest convergence gate), #4642 (demand-driven hosting epic)

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/put/op_ctx_task.rs
+++ b/crates/core/src/operations/put/op_ctx_task.rs
@@ -1675,22 +1675,12 @@ where
                     "PUT relay: chain descended here and next hop moves away; finalizing here (#4363)"
                 );
                 // Finalize the operation at this closest-seen node (#4363
-                // placement) but STILL replicate the contract one hop onward so
-                // the UPDATE broadcast tree stays connected (#4509 convergence).
-                // See `relay_put_replicate_forward` for why both are required.
-                if let Some(next_addr) = peer.socket_addr() {
-                    relay_put_replicate_forward(
-                        op_manager,
-                        next_addr,
-                        key,
-                        contract.clone(),
-                        related_contracts.clone(),
-                        merged_value.clone(),
-                        htl,
-                        new_skip_list.clone(),
-                    )
-                    .await;
-                }
+                // placement). The contract is stored locally here; UPDATE
+                // convergence is carried by the demand-driven mesh
+                // (subscribe-chains + live fan-out + anti-entropy), so no
+                // extra replication hop past the terminus is needed. The old
+                // #4509 store-one-hop-onward copy was proven redundant by the
+                // #4752 honest convergence gate and removed.
                 let hop_count = op_manager.ring.max_hops_to_live.saturating_sub(htl);
                 return relay_put_finalize_local(
                     op_manager,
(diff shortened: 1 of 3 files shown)


sanity:
Closing — the standalone deletion **regresses findability**, so the hop stays for now.

CI's piece-E findability gate (`test_piece_e_findability_sparse_ring_gate`, an honest GET-based distant-findability test) fails deterministically on this branch (verified 3/3 each side):

| | mean client-findability | hosting copies/seed | result |
|---|---|---|---|
| main (with hop) | **0.854** | 14,12,15,13,13,6 | PASS |
| this branch (no hop) | **0.521** | 9,6,9,12,4,6 | FAIL (floor 0.60) |

Removing `relay_put_replicate_forward` drops distant-GET findability so ~48% of GETs dead-end for a contract that *exists* (vs the ~5-9% near-miss floor the epic targets). The other 8 failures share one root cause — PUT reach dropped: 7 are bare-SUBSCRIBE confounds whose subscriber can't fetch the WASM without the hop's replication, plus 1 broad-holder freshness test (`dense_network_replication`).

The honest convergence gate #4752 correctly showed the hop is redundant for **convergence** (a subscribed mesh converges identically without it) — but convergence-safe is not deletion-safe: the hop is **load-bearing for findability**. Removal is deferred until real findability mechanisms land (the separate findability workstream — bounded backtracking / every-hop-placement maturity / near-key replication #4750).

The new honest convergence-gate test added here (`honest_put_mesh_convergence_{steady,churn}`) is independently valuable (honest PUT-routed convergence coverage the suite lacked) and will be salvaged into a separate test-only PR. Branch kept for that salvage.

[AI-assisted - Claude]

PR #4759, opened 2026-07-10
refactor(update): remove redundant Source-2 interest fan-out (#4642 step 9)

sanity:
## Problem

Live UPDATE fan-out resolved broadcast targets from **two** sources, unioned in
`get_broadcast_targets_update` (`crates/core/src/operations/update.rs`):

- **Source 1** — `neighbor_hosting.neighbors_with_contract(key)`: advertised
  co-hosts (peers that announced, via the advertisement layer, that they host
  the contract).
- **Source 2** — `interest_manager.get_interested_peers(key)`: peers that
  registered interest via the protocol.

Under the demand-driven hosting model (`.claude/rules/hosting-invariants.md`
invariant 1, `docs/design/demand-driven-hosting.md`), live fan-out is defined as
**Source-1 only**, with the periodic InterestSync **anti-entropy heartbeat** as
the healing backstop. Source 2 is redundant: every co-host advertises (so it is
already a Source-1 target), and any transient gap between "interested" and
"advertised" is closed by the ~5-min heartbeat, which re-pulls each neighbor's
hosted-contract set via `HostingStateRequest` (`ring.rs::interest_heartbeat`,
#4722). This is R4 step 9 of the demand-driven hosting epic (#4642).

## Approach

Remove **only** the Source-2 fan-out arm from `get_broadcast_targets_update`;
keep Source-1. This is narrowly the *state fan-out* arm:

- `InterestManager` and `get_interested_peers` are **retained** — other live
  consumers stay wired: `send_proactive_summary_notification` (proactive summary
  fan-out), the interest-heartbeat TTL refresh, and eviction demand-counting.
- The `#4359` fresh-PUT deferred-broadcast flush call sites are **left intact**.
  An interest-triggered flush that now finds no Source-1 target simply
  re-stashes or lets the entry expire at its TTL — harmless and cheap. The
  meaningful flush signal is now Source-1 (a neighbor advertising, live or
  reconciled by anti-entropy). Module/dispatch docs updated to say so
  (`pending_broadcast.rs`, `op_state_manager.rs::flush_pending_broadcast_on_interest`).
- No wire change: `BroadcastStateChange` is an internal `NodeEvent`, so **no
  version gate** is needed.
- The `interest_found` / `interest_resolve_failed` fields on
  `BroadcastTargetResult` are kept (now always 0) for telemetry-schema
  continuity — the #4281 propagation-stats window and the #3046 delivery event
  still carry them. A follow-up cleanup can retire the vestigial fields.

## Testing

The safety proof is that a subscriber not reached by live Source-1 fan-out still
converges via the anti-entropy backstop.

- **Simulation (the safety proof)**:
  `test_step9_source2_removal_subscriber_heals_via_anti_entropy`
  (`simulation_integration.rs`). A subscribed mesh commits an UPDATE; the
  reachable mesh subscribers end fresh via **Source-1 live fan-out alone**
  (positive control — their only route to the live update is advertised-co-host
  fan-out, Source-2 being gone). One mesh subscriber is crash-isolated across
  the UPDATE (the harness has no per-message-type loss knob, so a crash across
  the update is the deterministic proxy for "did not receive the live fan-out"),
  then recovered; with no new update after recovery, only the anti-entropy
  heartbeat can bring it current, and it MUST heal to v2 within the heartbeat
  window. Modeled on the proven `test_piece_b_anti_entropy_heals_demandless_copy`
  (720 s / 680 s virtual time).
- **Unit (the boundary proof)**:
  `broadcast_targets_are_advertised_cohosts_only_and_heal_via_advertisement`
  (`op_state_manager.rs`). At the `get_broadcast_targets_update` boundary:
  (1) an interest-only peer (registered but not yet advertised — the lagged case)
  is EXCLUDED; (2) an advertised co-host IS included; (3) the anti-entropy heal
  (a `HostingAnnounce { is_response: true }`, exactly what the heartbeat's
  `HostingStateResponse` reconciliation delivers) makes the lagged peer a target.
- **Regression**: the #4359 pending-broadcast flush tests and the existing
  `test_subscription_broadcast_propagation` / `test_piece_b` sims remain green
  (Source-1 path and anti-entropy backstop both intact).

## Precondition

Depends on R1 advertisement-layer reliability (#4722) — the anti-entropy
`HostingStateRequest`/`HostingStateResponse` reconciliation that backstops this
removal — which is live and field-validated since 0.2.94.

## Review / merge

- **Full-tier multi-model review required** (UPDATE broadcast path — the
  #3791 fan-out / #4145/#4231 backpressure incident family).
- **Merge HELD** pending review AND the findability-investigation verdict for
  the epic. Do not auto-merge.

Part of #4642 (R4 step 9).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1803,8 +1803,10 @@ where
                     // now a `neighbors_with_contract` broadcast target. If a
                     // fresh-contract PUT gave up with no targets and is stashed,
                     // flush it here — this is the proximity first-viable-target
-                    // signal, distinct from the interest-manager (Source 2)
-                    // signals. Must run BEFORE the receiving-updates/downstream
+                    // signal, distinct from the interest-registration flush sites
+                    // (`register_peer_interest`; the Source-2 live fan-out arm
+                    // itself was removed in #4642 step 9). Must run BEFORE the
+                    // receiving-updates/downstream
                     // gate below, which `continue`s for exactly the
                     // locally-hosted-only fresh-PUT case this fix targets.
                     op_manager.flush_pending_broadcast_on_interest(&key).await;
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -5054,9 +5054,13 @@ mod tests {
 
     /// #4359 trigger-completeness pin (MUST-FIX 1). The deferred broadcast only
     /// drains when a viable target appears, signalled by
-    /// `flush_pending_broadcast_on_interest`. The interest-manager (Source 2)
-    /// path makes a peer a target via `register_peer_interest`; EVERY such call
-    /// site that can be the first viable target for a never-seen id must flush.
+    /// `flush_pending_broadcast_on_interest`. Since #4642 step 9 an interest-only
+    /// peer is no longer itself a broadcast target (the Source-2 live fan-out arm
+    /// was removed), but interest registration typically coincides with the
+    /// peer's advertisement, so every `register_peer_interest` call site keeps
+    /// flushing as the eager wakeup for the Source-1 target that lands alongside
+    /// it. EVERY such call site that can be the first viable target for a
+    /// never-seen id must flush.
     ///
     /// Rather than trusting a static allow-list of files (the brittle shape the
     /// re-review flagged — a NEW production `register_peer_interest` added in
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -2939,11 +2939,12 @@ mod tests {
     ///    advertisement) case; under the old two-source model it WOULD be a
     ///    target via Source 2, and after step 9 it must not be.
     /// 2. An advertised co-host (Source 1) is still INCLUDED.
-    /// 3. The anti-entropy heal restores the lagged peer: once its advertisement
-    ///    arrives — exactly what the periodic InterestSync
-    ///    `HostingStateRequest`/`HostingStateResponse` reconciliation delivers,
-    ///    a `HostingAnnounce { is_response: true }` (`ring.rs::interest_heartbeat`,
-    ///    #4722) — it becomes a broadcast target.
+    /// 3. The advertisement-reconciliation heal restores the lagged peer: once
+    ///    our view of it includes the contract it becomes a broadcast target. The
+    ///    periodic interest heartbeat sends each neighbor a `HostingStateRequest`;
+    ///    the neighbor replies with a `HostingStateResponse` snapshot of its
+    ///    hosted set, which we full-replace into our Source-1 view
+    ///    (`ring.rs::interest_heartbeat`, #4722).
     ///
     /// This is the unit-level counterpart of the lagged-advertisement
     /// convergence simulation: removing Source-2 is safe precisely because a
@@ -3034,16 +3035,16 @@ mod tests {
             "interest_found is vestigial after Source-2 removal and must report 0"
         );
 
-        // THE HEAL: B's advertisement arrives. This mirrors the periodic
-        // InterestSync anti-entropy `HostingStateResponse` reconciliation
-        // (`ring.rs::interest_heartbeat`, #4722), which re-pulls a neighbor's
-        // hosted set as a `HostingAnnounce { is_response: true }`.
+        // THE HEAL: the periodic interest heartbeat re-pulls B's hosted set. We
+        // send B a `HostingStateRequest`; B replies with a `HostingStateResponse`
+        // snapshot of what it hosts, which we full-replace into our view of B
+        // (`ring.rs::interest_heartbeat`, #4722). That is what moves B into
+        // Source-1. (An advertised co-host may also announce proactively via
+        // `HostingAnnounce`; either path lands B in the proximity cache.)
         op_manager.neighbor_hosting.handle_message(
             kp_b.public(),
-            crate::message::NeighborHostingMessage::HostingAnnounce {
-                added: vec![cid],
-                removed: vec![],
-                is_response: true,
+            crate::message::NeighborHostingMessage::HostingStateResponse {
+                contracts: vec![cid],
             },
         );
 
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -789,29 +789,37 @@ impl OpManager {
     ///
     /// ## Trigger completeness (issue #4359 re-review, MUST-FIX 1)
     ///
-    /// `get_broadcast_targets_update` resolves targets from two sources, so the
-    /// *first viable target* for a cold id can appear via either. This method is
-    /// invoked from every site that makes a peer a target for the first time:
+    /// Since #4642 step 9, `get_broadcast_targets_update` resolves targets from
+    /// a SINGLE source — the proximity cache of advertised co-hosts (Source 1).
+    /// The *first viable target* for a cold id therefore appears when a
+    /// connected neighbor newly announces it hosts one of our contracts:
     ///
-    /// * **Source 2 — interest manager** (`register_peer_interest` is_new):
+    /// * **Source 1 — proximity cache** (`neighbors_with_contract`):
+    ///   - `node.rs` NeighborHosting overlap path, when a neighbor newly
+    ///     announces hosting one of our contracts (the neighbor just became a
+    ///     `neighbors_with_contract` target) — live, or reconciled by the
+    ///     ~5-min InterestSync anti-entropy heartbeat
+    ///     (`ring.rs::interest_heartbeat`, #4722).
+    ///
+    /// The interest-registration flush sites remain wired even though an
+    /// interest-only peer is no longer a broadcast target (Source 2 was removed
+    /// in #4642 step 9): a flush that finds no Source-1 target simply re-stashes
+    /// or lets the entry expire — harmless and cheap, and it keeps the eager
+    /// #4359 wakeup for the case where an interest registration coincides with
+    /// the peer's advertisement. Those sites are (`register_peer_interest`
+    /// is_new):
     ///   - `subscribe::register_downstream_subscriber` (downstream subscriber)
     ///   - `subscribe::finalize_originator_subscribe` (originator upstream)
     ///   - `get::op_ctx_task` remote GET `subscribe=false` (requester interest)
     ///   - `node.rs` Interests / Summaries interest-sync handlers
-    /// * **Source 1 — proximity cache** (`neighbors_with_contract`):
-    ///   - `node.rs` NeighborHosting overlap path, when a neighbor newly
-    ///     announces hosting one of our contracts (the neighbor just became a
-    ///     `neighbors_with_contract` target).
     ///
     /// That set is the complete first-viable-target signal: a never-seen id can
-    /// only gain a broadcast target by a peer expressing interest (Source 2) or
-    /// by a connected neighbor announcing it hosts the id (Source 1), and each
-    /// such transition routes through one of the sites above. The give-up path
-    /// itself also re-checks targets immediately after stashing
-    /// (`pending_broadcast_stash_recheck`) to close the stash-after-flush race.
-    /// A grep pin test (`pending_broadcast_flush_wired_at_all_interest_sites`)
-    /// guards against a future `register_peer_interest` call site forgetting the
-    /// flush.
+    /// only gain a broadcast target by a connected neighbor announcing it hosts
+    /// the id (Source 1). The give-up path itself also re-checks targets
+    /// immediately after stashing (`pending_broadcast_stash_recheck`) to close
+    /// the stash-after-flush race. A grep pin test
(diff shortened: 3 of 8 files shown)


sanity:
## Full-tier multi-model review — PR #4759 (Source-2 interest fan-out removal)

Reviewed at HEAD `2399ca5b` (CI green). This is the UPDATE broadcast fan-out path (the #3791 fan-out / #4145/#4231 backpressure incident family), so it got the full treatment: an external non-Claude model (**Codex `gpt-5.5`, `codex review --base main`** — ran successfully) plus four adversarial Claude lenses (correctness/core-risk, #4359 flush interaction, completeness/dead-code, and a hand trace of the delivery paths). Findings are deduplicated and reconciled below.

### Bottom line: **MERGE-AFTER-FIXES / NEEDS-ADJUDICATION** — not clean merge-ready yet.

The removal is mechanically clean (no dead code, no broken callers, no spin/leak in the #4359 flush) and is correct for the **direct** subscriber classes: SUBSCRIBE originators and GET-with-subscribe subscribers all `announce_contract_hosted` once they hold the body, so they become Source-1 (advertised-co-host) targets, and the two anti-entropy backstops on the 300 s heartbeat genuinely heal a lagged advertisement. **But there is one live path the PR's safety argument does not cover** (B1 below), and it isn't exercised by either new test. Since the PR is already merge-held pending the epic's findability verdict, B1 is exactly the kind of thing that hold should resolve — it should be adjudicated explicitly, not left implicit.

---

### BLOCKING / MUST-ADDRESS

**B1 — Hollow multi-hop SUBSCRIBE relays lose live UPDATE delivery, and neither anti-entropy nor the retained proactive-summary path heals them.** (`crates/core/src/operations/update.rs:348-353`; found by Codex, corroborated by hand trace.)

Live fan-out is now Source-1 only = `neighbor_hosting.neighbors_with_contract(key)` = peers that *advertised* they host the contract. Consider a genuine multi-hop standalone SUBSCRIBE chain `S → R → H` (H hosts, R is an intermediate forwarding hop):

- When R forwards S's subscribe to H, H registers **R** as a downstream subscriber → R was a Source-2 target at H (now removed).
- R is a **hollow relay**: on the non-local-hit forward path it does not fetch/host the contract and is pin-tested to **not** advertise (`subscribe/op_ctx_task.rs:3006` "relay driver must NOT call announce_contract_hosted on relayed response"). So R is **never** in H's Source-1.
- The leaf S does advertise, but only to **R** (its neighbor), not to H. So after Source-2 removal, **H fans the UPDATE to nobody in this chain** — R never gets it, so S never gets it.
- **Anti-entropy does not heal this**: H's `HostingStateRequest` re-pulls R's *hosted* set, but R doesn't host the contract, so R never enters H's `neighbors_with_contract`.
- **The retained proactive-summary path does not heal it either.** `send_proactive_summary_notification` (`update.rs:722`) does fan a `Summaries` to interested peers, but the receiver's handler (`node.rs:2662-2664, 2696-2699, 2774`) only acts when the receiver has local state, and even then it pushes the *receiver's own* state to the *sender* (`get_contract_state` + `stale_peer_sync_event(..., source)`) — it never delivers the sender's newer state to the receiver, and a body-less/hollow receiver skips entirely. (This corrects an over-optimistic read from one lens that the summary path gives interested peers an immediate heal — it does not deliver state.)

Net: for a genuine multi-hop hollow-relay subscribe chain, the downstream subscriber can be left **indefinitely stale** (until it re-subscribes/re-GETs), not merely delayed by one ~5-min heartbeat. The PR's justification ("every co-host advertises → already a Source-1 target; transient gaps closed by anti-entropy") is **false for hollow relays**, which never advertise and are never healed.

This is a live code path: client `subscribe()` → `run_executor_subscribe` → `drive_relay_subscribe` → multi-hop `k_closest_potentially_hosting`; standalone SUBSCRIBE is not yet folded into GET (that's R4 step 10 / piece D). Neither new test covers it — the sim uses GET-with-subscribe (every-hop hosting → all subscribers advertise), and the existing `test_subscription_broadcast_propagation` is 1-hop (leaf advertises directly).

Caveat on severity: the **real-world impact depends on how often durable multi-hop hollow-relay standalone-SUBSCRIBE chains actually form** (with every-hop hosting live, many subscribes terminate at a nearby advertised host in 0-1 hops), and on whether such chains delivered updates reliably even before this PR. The PR does not establish either. Recommended resolution, any one of:
1. Demonstrate (analysis or telemetry) that durable hollow-relay multi-hop subscribe chains don't form / are acceptably rare in the current network, and state that in the PR body; **or**
2. Add a sim that builds a genuine `S → R(hollow) → H` chain, commits an UPDATE, and asserts S converges (this is the untested class); **or**
3. Sequence this removal **after** hollow relays are eliminated (piece D makes chain peers real hosts / SUBSCRIBE retirement R4 step 10), so no hollow relay exists when Source-2 goes.

This mirrors the epic's own `relay_put_replicate_forward` meta-lesson (#4758): *convergence-safe for a subscribed mesh is not the same as reach-safe for every path.* The two convergence-focused tests here prove the subscribed-mesh case; they don't prove the hollow-relay reach case.

---

### NON-BLOCKING (recommend fixing in this PR)

**NB1 — The sim test's "positive control" does not isolate Source-1.** (`tests/simulation_integration.rs` ~14273-14343; raised independently by two lenses.) The 680 s post-op wait exceeds the 300 s anti-entropy heartbeat, and the `SyncStateToPeer` state-heal does **not** go through `get_broadcast_targets_update`. So a *hosting* mesh subscriber would heal via 2+ anti-entropy cycles even if Source-1 fan-out were entirely broken — the assertion "reachable mesh subscribers must all be fresh via Source-1 live fan-out" can't distinguish "Source-1 delivered it" from "anti-entropy backfilled it." The **backstop** half is sound (no new update after recovery, so anti-entropy is the only heal), and the **unit test** genuinely pins the boundary (interest-only peer excluded, advertised co-host included), which is why this is non-blocking. Fix: add a short/no post-op sub-case that asserts mesh freshness *before* the first heartbeat can fire, or soften the docstring/assertion to not over-claim Source-1 isolation.

**NB2 — Doc rot: several inline comments and one telemetry field-doc still assert the old two-source model** ("interest registration makes a peer a broadcast target"), which this PR made false. AGENTS.md requires fixing outdated docs in the same PR. Concrete sites:
- `crates/core/src/operations/update/propagation_stats.rs:88-89` — field doc says "*A persistently non-zero value points at stale interest state*", now **impossible** (hard-zeroed). Actively misleading; also `:139` and `:226-227`. Consider just dropping the always-0 internal counter rather than emitting it to operator logs (`:309,321`).
- `crates/core/src/node.rs:2624`, `:2914` — "makes this peer a viable broadcast target".
- `crates/core/src/operations/subscribe.rs:530`, `:670` — "this … peer is now a viable broadcast target" / "first viable target".
- `crates/core/src/operations/get/op_ctx_task.rs:2687` — "making them a viable broadcast target".
- `crates/core/src/node/network_bridge/p2p_protoc/broadcast.rs:340-351` — the stash-recheck rationale is framed on an *interest-registration* race that can no longer produce a target (the recheck is still valid, but now only catches a Source-1 advertisement race).
- `crates/core/src/node/network_bridge/p2p_protoc.rs:5057-5059` — the `pending_broadcast_flush_wired_at_all_interest_sites` pin rustdoc still says the interest path "makes a peer a target"; the pin now guards belt-and-suspenders, and its failure message over-claims ("the #4359 fix silently fails for that path").
- `crates/core/src/node.rs:1806` — borderline-stale "distinct from the interest-manager (Source 2) signals".

**NB3 — Two behavioral changes worth stating in the PR body.** (a) *Widened initial fan-out window*: a just-subscribed downstream is not a live target until its `HostingAnnounce` reaches the host (vs. Source-2 making it a target the instant the host processed the inbound SUBSCRIBE); bounded by anti-entropy, sub-second typically. (b) *Lost recovery path for a body-less interested peer*: Source-2 used to deliver a live update to an interested-but-body-less peer (which then auto-fetched); Source-1 won't, and the anti-entropy state-heal is skipped when either summary is `None` — such a peer now waits on the ~2-min renewal re-fetch. (b) is the mild cousin of B1 for the leaf-with-a-dead-body-fetch case.

---

### NITs

- **N1** — `PendingBroadcastStore::stash` sets a fresh `inserted_at_ms`, so an interest-driven re-emit/re-stash can refresh the 5-min TTL; a stream of distinct new interested peers on a still-unadvertised cold contract could keep one entry alive past its nominal TTL. Bounded in count (1024) and normally drained by advertisement — not a leak, but mildly in tension with the AGENTS.md "cleanup exemptions must be time-bounded" rule; worth a one-line module-doc ack (the doc still says entries just "expire at their TTL").
- **N2** — "harmless and cheap": each new interested peer on a still-cold contract now triggers a full ~3-retry / ~6 s broadcast cycle that finds nothing (plus one contract-state read per flush) instead of the pre-PR O(1) deliver-and-drain. Still bounded, but "cheap" is slightly generous for a fresh contract drawing a subscriber burst before anyone advertises.
- **N3** — The unit test heals by injecting `HostingAnnounce { is_response: true }` directly rather than literally exercising the `HostingStateResponse` reconciliation arm its docstring cites; equivalent end-state, but not the literal path.

---

### Confirmed clean
- **Dead code / callers**: `interest_manager.get_interested_peers` is still legitimately used (`send_proactive_summary_notification` `update.rs:722`, unsubscribe `op_state_manager.rs:1180`, upstream tracking `ring.rs:3795`); the only production caller of `get_broadcast_targets_update` is the UPDATE broadcast path; all remaining locals are still read; no clippy risk.
- **Telemetry schema**: hard-zeroing `interest_found`/`interest_resolve_failed` breaks nothing — `ContractCounters::is_empty()`/`aggregate_window` gate on `updates`, so no window flips to falsely "healthy."
- **#4359 flush**: no spin (driven only by discrete `register_peer_interest`-returns-new events; a re-emit never re-registers interest), no unbounded leak, and the cold pending broadcast is still delivered via the **Source-1 advertisement** flush at `node.rs:1810` (fires after the announcement is recorded, guarded for the stashed-cold-PUT case).

[AI-assisted - Claude]

sanity:
**Held pending R4 step 10 (standalone-SUBSCRIBE retirement / hollow-relay elimination) — sequencing decision from the Full-tier review.**

The review's blocking finding (B1, caught by Codex) shows removing Source-2 breaks live UPDATE delivery through a hollow multi-hop SUBSCRIBE relay chain (`S → R → H`): the intermediate relay R registers interest at H (a former Source-2 target) but never advertises hosting (pin-tested, `subscribe/op_ctx_task.rs:3006`), so it is not in H's Source-1 — and neither anti-entropy (R does not host) nor the proactive-summary path heals it, leaving the downstream subscriber indefinitely stale. Source-2 is that chain's only live fan-out path today.

So this must land **after (or with) step 10**, which eliminates the hollow-relay / standalone-SUBSCRIBE path — not before it, as the original scoping assumed. The lattice-fix release ships without it.

Converting to draft. Non-blocking findings to address when this resumes alongside step 10: (1) the sim positive-control doesn't isolate Source-1 (680s window > 300s heartbeat, so anti-entropy backfills), (2) doc rot in ~7 comments + `propagation_stats.rs:88-89`, (3) the two behavioral changes noted in the PR body. [AI-assisted - Claude]

sanity:
## Full-tier review + crux verification — MERGE-READY

**Reviewers:** external Codex + two adversarial Claude lenses (recipient-coverage; op_state/tests) + a dedicated verification of the one disputed crux.

**Verdict: safe to merge — bounded-latency change, NOT a dropped-update bug.** The scariest case, local WebSocket clients, is CLEAN (they were never in the removed Source-2 path; local delivery is a separate executor path). The one real question — does removing Source-2 drop live updates to a just-subscribed peer — resolved definitively:

- The anti-entropy summary path genuinely does NOT heal a no-state (`None` summary) interested peer (the `Summaries` staleness gate uses `our_summary.zip(their_summary)`, Some only if both exist — Codex read this correctly).
- BUT a real subscriber never persists in that no-state condition: `finalize_originator_subscribe` awaits `fetch_contract_if_missing` (its own subscribe-GET, pulling current state incl. any committed update) BEFORE returning success. Once stateful it advertises → becomes a Source-1 target + summary-Some, and both anti-entropy paths keep it fresh.

So removal is safe **because subscribe seeds state via GET**, not because the heartbeat heals no-state peers.

**Polish applied (this addresses every finding):**
- Corrected the `get_broadcast_targets_update` rustdoc to cite the real heal mechanism (subscribe-GET seeding + advertisement reconciliation), replacing the wrong ~5-min HostingStateRequest-re-pull claim.
- Added invariant pin test `subscribe_seeds_state_via_get_before_finalize`: asserts the seed-GET runs and is awaited BEFORE the advertise — so a future bare-SUBSCRIBE path can't silently reintroduce a no-state subscriber that Source-2 was covering.
- Fixed two doc-drift nits (p2p_protoc.rs, node.rs) + corrected the unit test's heal message to the real `HostingStateResponse`.

**Green:** fmt, clippy `-D warnings`, `operations::update` (67), `operations::subscribe` (63 incl. new pin). Production logic unchanged from the reviewed version. Holding merge for tomorrow morning (past the late-night guard) alongside the 0.2.97 release.

[AI-assisted - Claude]

PR #4760, opened 2026-07-11
feat(topology): seek+tighten nearest-neighbor ring lattice (succ/pred)

sanity:
## Problem

Kleinberg small-world routing needs a GUARANTEED short-range base lattice (each node's immediate ring neighbors) PLUS the 1/d long-range links. freenet-core builds the long links (Kleinberg gap-fill in `topology/small_world_rand.rs`) but never guarantees the base lattice: the gap-fill objective scores the **nearest** candidate **lowest** — an ordinal "who is my immediate neighbor" property cannot emerge from a log-distance metric — so greedy routing dead-ends at local minima with no connected neighbor closer to the key. An instrumented route-to-self experiment showed these dead-ends are a **connectivity gap** (no connected neighbor closer to the key), not a routing give-up, which caps last-mile findability (part of the demand-hosting epic #4642, invariant 5).

## Approach

Add the two-tier structure Chord/Symphony ship (freenet was "Symphony minus the short links") — **four LOCAL mechanisms, NO wire change, NO version gate**:

1. **Reserve 2 lattice edges per peer** — the closest connected SUCCESSOR (higher location, wrapping) + closest connected PREDECESSOR (lower location), derived on demand from the live connection set. The other k−2 slots stay long-range (untouched Kleinberg), so routing stays O(log²n).
2. **Discovery = route-to-self CONNECT toward `own_location`**, reusing the existing CONNECT path. `start_client_connect` already pre-populates the visited bloom with all connected peers, so the probe lands on the nearest UNCONNECTED peer and the terminus installs the edge via the per-side clause. Per-peer exponential backoff; **probes only fire while a lattice side is unfilled** — tightening to a strictly-closer peer that appears later is handled passively by the acceptance clause, so discovery never manufactures extra nearby (non-lattice) links.
3. **Per-side strict-closer acceptance in `should_accept`** — unconditionally accept a candidate that would become this peer's new nearest ON ITS OWN SIDE. Strictly per-side (never a global "closer than any" test) so BOTH edges form the closed undirected cycle greedy-to-closest requires. Over max, the edge is admitted and displaces a long link (`add_connection` over-max exception; the over-max prune sheds a non-lattice link next tick).
4. **Retention exemption** — the 2 lattice edges are exempt from the SCORE-BASED swap / prune victim selection ONLY (`topology.rs`). They remain fully subject to **liveness eviction** (`prune_connection`), and a lost edge triggers immediate re-discovery — never a permanent pin to a dead neighbor (AGENTS.md GC-exemption rule).

Telemetry ships with the fix: per-peer lattice completeness (successor / predecessor held + ring distances) and discovery health (probes / improvements) on the dashboard ring-stats card.

The behavior is ON in production **above a minimum-degree floor** (`max_connections >= NN_LATTICE_MIN_MAX_CONNECTIONS = 25 = DEFAULT_MIN_CONNECTIONS`). Two-tier routing (2 reserved base-lattice edges + `k−2` Kleinberg long links) needs enough long links left after the reserve; below the network's own minimum-degree target a node can't sustain the structure. The gate is keyed on the **configured** `max_connections` — 200 in production via `DEFAULT_MAX_CONNECTIONS` — **not the live connection count**, so every production node is above the floor and always active; the floor's only real effect is to keep the **sparse simulation suite** (configs top out at `max_connections = 16`) at its pre-lattice baseline. That is what greens CI: the sim tests that were failing (max 3–6) now run stock topology, and the default hosting/subscription/determinism sim tests (max 15–16) stay feature-off too, unchanged from before this feature existed; only the production-scale max=200 sim tests exercise the lattice. The single activation gate is `nn_lattice_active_for(max_connections)` / `ConnectionManager::nn_lattice_active`, applied at every feature site (acceptance clause, over-cap admission, discovery probe, retention exemption). A test-only per-thread override (`set_nn_lattice_enabled`) runs the stock arm for control-vs-fix validation AND force-activates the lattice below the floor, so the dedicated benefit tests still exercise the ON path at `max_connections = 5`.

## Testing

Deterministic unit tests (`topology.rs`, `connection_manager.rs`): per-side protected-set correctness, score-swap exemption, liveness eviction still evicts a lattice edge, over-max lattice admission.

Simulation validation (`simulation_integration.rs`, control [lattice off] vs fix [lattice on] on identical seeds, 16 peers = 1 gw + 15 nodes, sparse max_connections=5). Numbers below are means over 6 seeds — see the `## Validation` comment for the full per-seed table.

**Control-vs-fix (6 seeds, 16 peers = 1 gw + 15 nodes, sparse max_connections=5):**

| Metric (mean over 6 seeds) | Stock (lattice off) | Fix (lattice on) |
|---|---|---|
| **PUT-reach** — scatter-free PUT lands at rank 0 (peer closest to key) | **3/6** | **6/6** |
| **GET-reach** — distant lone-copy GET find-rate | 0.875 | **0.917** (+0.042) |
| **Exact-nearest coverage** — peers holding BOTH true edges | 0.458 | **0.531** (+0.073) |
| **Routing efficiency** — forward hops of found GETs | ~4.4 | ~3.4 (fewer, not degraded) |
| **Churn** — total disconnect events | 4 | 8 (both low) |
| **Long-link buckets** [<1/64, <1/16, <1/4, ≥1/4] (seed 5A1D) | [0,0,10,21] | [0,0,22,14] |

Two decisive per-seed dead-ends the fix removes: stock's scatter-free PUT **lands 5 and 6 ring-ranks off** the key on seeds 0007/000B (fix → rank 0 on all six); and on seed 000B stock's distant GET finds the lone copy only **3/8** times vs the fix's **8/8**. These are exactly the connectivity-gap dead-ends (no connected neighbor closer to the key) the lattice closes.

**Honest caveats:** the exact-nearest coverage delta (+0.073) and GET delta (+0.042) are real but modest — at 16 nodes / max_connections=5 the per-seed variance is large for BOTH arms (a sparse-sim artifact; both PUT-reach and the two big GET/PUT dead-ends are the reproducible signal). The fix trades a few far long-links for medium ones (mild distribution skew) and roughly doubles disconnect events (4 → 8, both low). The signal would be cleaner at larger scale, but the sim harness does not reliably form a 48-node ring even with extra gateways + longer settle (a separately-tracked harness limitation, not a property of the fix — reported honestly rather than faked).

## Security disclosure — eclipse (DISCLOSED and ACCEPTED)

The per-side nearest force-accept plus the retention exemption make per-side-nearest admission deterministic and remove the score-based swap as a safety valve (liveness eviction still applies — a lattice edge is never a permanent pin to a dead neighbor). Candidate locations are address-masked (`Location::from_address`: an IPv4 /24 or IPv6 /48 collapses to one location), and the operator blocklist is checked before `should_accept`, so this force-accept cannot bypass it.

The IPv6 cost is not symmetric with IPv4, and the honest disclosure is: a single routed IPv6 allocation (an ISP hands out a /48, /56, or /64 per customer) already yields a **vast** number of distinct /48 `Location`s essentially for free, so an attacker can cheaply mint locations that straddle a chosen victim and become **both** of that victim's per-side nearest edges — a targeted last-mile eclipse of that one victim's 2 lattice slots. What it is **not**: a network partition or global takeover. The damage is bounded to a **specifically-targeted** victim's two slots (it does not compound across the ring), the exemption only ever protects whoever is genuinely per-side nearest, and a closer **honest** peer force-displaces the attacker. **The project lead has explicitly accepted this tradeoff** — cheap IPv6 last-mile eclipse of a single victim's 2 slots, in exchange for the findability the guaranteed lattice buys. A routing-correctness eviction signal (shedding a responsive peer that drops/corrupts payloads) is future work.

## Review

**Full-tier.** This touches peer connection formation (`should_accept`), the topology maintenance loop, and the swap/prune machinery — Freenet's highest-risk surfaces. Held for morning + Full multi-model review; NOT auto-merged.

Refs #4642. Findability investigation: the instrumented route-to-self connectivity-gap experiment.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -44,6 +44,26 @@ WHEN accepting a new connection (should_accept):
      add_connection, AND in BOTH connection-lifecycle promotion gates
      (p2p_protoc/connection_lifecycle.rs) so the exception is live on the
      real CONNECT path, not just should_accept. See connection_manager.rs.
+     MINIMUM-DEGREE FLOOR: the whole lattice feature (this exception, the
+     per-side acceptance clause, the route-to-self discovery probe, and the
+     topology.rs retention exemption) is GATED OFF when this peer's CONFIGURED
+     max_connections is below NN_LATTICE_MIN_MAX_CONNECTIONS (= 25 =
+     DEFAULT_MIN_CONNECTIONS). Two-tier routing needs enough long links left
+     after the 2-slot base-lattice reserve; below the network's own
+     minimum-degree target a node can't sustain the structure. The gate reads
+     CONFIGURED max (200 in production), NOT the live connection count, so
+     production peers are always active and the floor's real job is to keep the
+     sparse simulation suite (configs top out at max=16) at its pre-lattice
+     baseline — only the production-scale max=200 sim tests exercise the lattice.
+     The single activation gate is nn_lattice_active_for(max_connections) /
+     ConnectionManager::nn_lattice_active(); the test-only override that flips
+     the stock/fix validation arm also force-activates below the floor (so the
+     dedicated benefit tests exercise the ON path at max=5).
+     RESERVATION-AWARE over-cap admission: admits_lattice_edge_over_cap also
+     refuses to force-accept a fill when another in-flight pending reservation
+     already targets the same empty side, bounding concurrent over-cap fills to
+     ONE per side (otherwise a full node force-accepts EVERY concurrent
+     candidate on an empty side).
   3. Compute Kleinberg gap score (small_world_rand::kleinberg_score):
      → Map all connection distances to log-space (1/d = uniform in log)
      → Score = min distance to nearest neighbor in log-space
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -75,7 +75,13 @@ WHEN accepting a new connection (should_accept):
      speculative-count term is DISTINCT from and simpler than the dropped per-side
      F1 clause (which bounded concurrent over-cap FILLS to one per EMPTY side, a
      near-non-case since a node at max almost always has both ring sides
-     populated). A second, independent bound: each admitted candidate that
+     populated). The SLACK budget is GLOBAL (counts all non-stale reservations,
+     not just over-cap lattice ones), so on a node AT max unrelated in-flight
+     handshakes can throttle tightening until they drain (bounded by the TTL;
+     continuous discovery retries). This only bites nodes genuinely at max (mostly
+     busy gateways); a peer below max tightens via the under-cap path, which never
+     consults this ceiling. A lattice-private budget is a possible future
+     refinement. A second, independent bound: each admitted candidate that
      ESTABLISHES advances the per-side current-nearest, so the established sequence
      is strictly decreasing (a short records chain) and the absolute ceiling
      hard-bounds the ESTABLISHED set. The route-to-self discovery probe
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -34,10 +34,11 @@ WHEN accepting a new connection (should_accept):
       See PR #3621, which replaced the TTL halving from PR #3582.
   2. CHECK: Are we at max_connections (open + pending)? → REJECT
      EXCEPTION (nearest-neighbor lattice, mechanism 3): a candidate that
-     would become this peer's new per-side NEAREST ring neighbor (its
-     successor or predecessor lattice edge) is force-accepted even at/over
-     max, up to a HARD over-max ceiling (max_connections +
-     LATTICE_OVERMAX_SLACK). It displaces a farther non-lattice long link,
+     would become this peer's new per-side NEAREST ring neighbor — a TIGHTEN
+     of an already-held side OR a FILL of an empty one (any strictly-closer
+     successor or predecessor) — is force-accepted even at/over max, up to a
+     HARD over-max ceiling (max_connections + LATTICE_OVERMAX_SLACK). It
+     displaces a farther non-lattice long link,
      which the topology maintenance loop's over-max prune sheds next tick
      (protected lattice edges are never the prune victim). Applied via the
      shared predicate admits_lattice_edge_over_cap() in should_accept, in
@@ -59,11 +60,21 @@ WHEN accepting a new connection (should_accept):
      ConnectionManager::nn_lattice_active(); the test-only override that flips
      the stock/fix validation arm also force-activates below the floor (so the
      dedicated benefit tests exercise the ON path at max=5).
-     RESERVATION-AWARE over-cap admission: admits_lattice_edge_over_cap also
-     refuses to force-accept a fill when another in-flight pending reservation
-     already targets the same empty side, bounding concurrent over-cap fills to
-     ONE per side (otherwise a full node force-accepts EVERY concurrent
-     candidate on an empty side).
+     TIGHTENING-AT-CAPACITY: over-cap admission fires for any strictly-closer
+     per-side nearest — a TIGHTEN of an already-held side OR a FILL of an empty
+     one — so the lattice CONTINUOUSLY pulls each peer toward its true nearest
+     ring neighbors, not merely filling empty sides. The superseded
+     former-nearest demotes into the long-link pool automatically (the reserved
+     slot is derived on demand from the live connection set). No reservation-aware
+     clause: each admitted candidate ADVANCES the per-side current-nearest, so the
+     admitted sequence is strictly decreasing (a short records chain) and the
+     absolute ceiling already hard-bounds the ESTABLISHED set; the earlier F1
+     one-fill-per-side clause was DROPPED (its concern was a concurrent fill flood
+     on an EMPTY side of a FULL node — a near-non-case, since a node at max almost
+     always has both ring sides populated). The route-to-self discovery probe
+     likewise runs CONTINUOUSLY (decaying toward tau_max, never stopping when both
+     sides are filled) so a filled-but-loose edge keeps tightening. See
+     connection_manager.rs and ring.rs.
   3. Compute Kleinberg gap score (small_world_rand::kleinberg_score):
      → Map all connection distances to log-space (1/d = uniform in log)
      → Score = min distance to nearest neighbor in log-space
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -65,13 +65,20 @@ WHEN accepting a new connection (should_accept):
      one — so the lattice CONTINUOUSLY pulls each peer toward its true nearest
      ring neighbors, not merely filling empty sides. The superseded
      former-nearest demotes into the long-link pool automatically (the reserved
-     slot is derived on demand from the live connection set). No reservation-aware
-     clause: each admitted candidate ADVANCES the per-side current-nearest, so the
-     admitted sequence is strictly decreasing (a short records chain) and the
-     absolute ceiling already hard-bounds the ESTABLISHED set; the earlier F1
-     one-fill-per-side clause was DROPPED (its concern was a concurrent fill flood
-     on an EMPTY side of a FULL node — a near-non-case, since a node at max almost
-     always has both ring sides populated). The route-to-self discovery probe
+     slot is derived on demand from the live connection set). SPECULATIVE-STATE
+     CEILING: the over-cap ceiling counts ESTABLISHED connections PLUS non-stale
+     pending reservations (the speculative-state invariant below: "am I
+     over-committed?" checks use speculative, not actual, state), which bounds
+     concurrent over-cap acceptances to LATTICE_OVERMAX_SLACK per
+     reservation-TTL window and stops a full node re-firing an accept plus
+     NAT-traversal on every strictly-closer fresh-address request. This global
+     speculative-count term is DISTINCT from and simpler than the dropped per-side
+     F1 clause (which bounded concurrent over-cap FILLS to one per EMPTY side, a
+     near-non-case since a node at max almost always has both ring sides
+     populated). A second, independent bound: each admitted candidate that
+     ESTABLISHES advances the per-side current-nearest, so the established sequence
+     is strictly decreasing (a short records chain) and the absolute ceiling
+     hard-bounds the ESTABLISHED set. The route-to-self discovery probe
      likewise runs CONTINUOUSLY (decaying toward tau_max, never stopping when both
      sides are filled) so a filled-but-loose edge keeps tightening. See
      connection_manager.rs and ring.rs.
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -33,6 +33,17 @@ WHEN accepting a new connection (should_accept):
       (default 8, separate from TTL). Each uphill hop decrements budget by 1.
       See PR #3621, which replaced the TTL halving from PR #3582.
   2. CHECK: Are we at max_connections (open + pending)? → REJECT
+     EXCEPTION (nearest-neighbor lattice, mechanism 3): a candidate that
+     would become this peer's new per-side NEAREST ring neighbor (its
+     successor or predecessor lattice edge) is force-accepted even at/over
+     max, up to a HARD over-max ceiling (max_connections +
+     LATTICE_OVERMAX_SLACK). It displaces a farther non-lattice long link,
+     which the topology maintenance loop's over-max prune sheds next tick
+     (protected lattice edges are never the prune victim). Applied via the
+     shared predicate admits_lattice_edge_over_cap() in should_accept, in
+     add_connection, AND in BOTH connection-lifecycle promotion gates
+     (p2p_protoc/connection_lifecycle.rs) so the exception is live on the
+     real CONNECT path, not just should_accept. See connection_manager.rs.
   3. Compute Kleinberg gap score (small_world_rand::kleinberg_score):
      → Map all connection distances to log-space (1/d = uniform in log)
      → Score = min distance to nearest neighbor in log-space
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -103,6 +103,24 @@ pub mod dev_tool {
     pub use ring::Location;
     pub use transport::{TransportKeypair, TransportPublicKey};
 
+    // Nearest-neighbor ring lattice (successor+predecessor base edges). Always
+    // ON in production; this setter is the test-only per-thread override the
+    // findability validation harness flips to run the stock (lattice-off) arm
+    // against the fix (lattice-on) arm on identical seeds. No-op in production
+    // builds (compiled out).
+    #[cfg(any(test, feature = "testing"))]
+    pub use ring::set_nn_lattice_enabled;
+
+    // Test-only findability measurement harness (NOT for ship): scatter/cache
+    // disable hook (guarantees a single copy) + per-op terminus tracing (rank /
+    // HTL / hop / stop-reason / connectivity-gap-vs-give-up). See
+    // operations::findability_probe. Compiled out in production.
+    #[cfg(any(test, feature = "testing"))]
+    pub use crate::operations::findability_probe::{
+        OpTraceRecord, ProbeOpKind, ProbeStopReason, clear_op_traces, scatter_disabled,
+        set_scatter_disabled, take_op_traces,
+    };
+
     // Test hooks: per-op driver call counters. Tests assert these
     // increment to confirm wire variants dispatch through their
     // driver (not a local-cache shortcut or legacy path).
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -104,12 +104,15 @@ pub mod dev_tool {
     pub use transport::{TransportKeypair, TransportPublicKey};
 
     // Nearest-neighbor ring lattice (successor+predecessor base edges). Always
-    // ON in production; this setter is the test-only per-thread override the
-    // findability validation harness flips to run the stock (lattice-off) arm
-    // against the fix (lattice-on) arm on identical seeds. No-op in production
-    // builds (compiled out).
+    // ON in production above the minimum-degree floor; `set_nn_lattice_enabled` is
+    // the test-only per-thread override the findability validation harness flips
+    // to run the stock (lattice-off) arm against the fix (lattice-on) arm on
+    // identical seeds — it ALSO force-activates the lattice below the floor so the
+    // benefit tests exercise the ON path at their sparse max_connections.
+    // `set_nn_lattice_force_active` resets only the floor bypass (used by the
+    // benefit-test cleanup guard). No-op in production builds (compiled out).
     #[cfg(any(test, feature = "testing"))]
-    pub use ring::set_nn_lattice_enabled;
+    pub use ring::{set_nn_lattice_enabled, set_nn_lattice_force_active};
 
     // Test-only findability measurement harness (NOT for ship): scatter/cache
     // disable hook (guarantees a single copy) + per-op terminus tracing (rank /
--- a/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
@@ -321,7 +321,7 @@ impl P2pConnManager {
                 // connection_manager.rs::{admits_lattice_edge_over_cap, LATTICE_OVERMAX_SLACK}.
                 let current = connection_manager.connection_count();
                 if current >= connection_manager.max_connections
-                    && !connection_manager.admits_lattice_edge_over_cap(loc)
+                    && !connection_manager.admits_lattice_edge_over_cap(loc, Some(peer_addr))
                 {
                     tracing::warn!(
                         tx = %tx,
@@ -1183,7 +1183,7 @@ impl P2pConnManager {
                 // connection_manager.rs::{admits_lattice_edge_over_cap, LATTICE_OVERMAX_SLACK}.
                 let current = connection_manager.connection_count();
                 if current >= connection_manager.max_connections
-                    && !connection_manager.admits_lattice_edge_over_cap(loc)
+                    && !connection_manager.admits_lattice_edge_over_cap(loc, Some(peer_addr))
                 {
                     tracing::warn!(
                         %peer_addr,
--- a/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
@@ -309,9 +309,20 @@ impl P2pConnManager {
                 // accepted at the protocol level (the CONNECT terminus handler in connect.rs) and NAT traversal
                 // already succeeded. Re-evaluating the probabilistic Kleinberg filter
                 // would discard hard-won connections for no capacity reason (#3545).
-                // Only enforce the hard max_connections cap as a resource safety limit.
+                // Enforce the hard max_connections cap as a resource safety limit,
+                // EXCEPT for a strictly-closer per-side nearest-neighbor lattice
+                // edge within the bounded over-max ceiling (mechanism 3). Without
+                // this exception the cap silently drops the guaranteed
+                // successor/predecessor edge on the REAL CONNECT/promotion path,
+                // making mechanism 3's at-capacity displacement inert in
+                // production (nodes routinely sit at max). The same predicate is
+                // re-checked in add_connection; the over-max topology prune sheds a
+                // farther non-lattice link next tick. See
+                // connection_manager.rs::{admits_lattice_edge_over_cap, LATTICE_OVERMAX_SLACK}.
                 let current = connection_manager.connection_count();
-                if current >= connection_manager.max_connections {
+                if current >= connection_manager.max_connections
+                    && !connection_manager.admits_lattice_edge_over_cap(loc)
+                {
                     tracing::warn!(
                         tx = %tx,
                         %peer,
@@ -1160,9 +1171,20 @@ impl P2pConnManager {
                 // establishment (NAT traversal) already succeeded. Re-evaluating
                 // the probabilistic Kleinberg filter would discard hard-won
                 // connections for no capacity reason (#3545).
-                // Only enforce the hard max_connections cap below.
+                // Enforce the hard max_connections cap below, EXCEPT for a
+                // strictly-closer per-side nearest-neighbor lattice edge within
+                // the bounded over-max ceiling (mechanism 3). Without this
+                // exception the cap silently drops the guaranteed
+                // successor/predecessor edge on the REAL CONNECT/promotion path,
+                // making mechanism 3's at-capacity displacement inert in
+                // production (nodes routinely sit at max). The same predicate is
+                // re-checked in add_connection; the over-max topology prune sheds a
+                // farther non-lattice link next tick. See
+                // connection_manager.rs::{admits_lattice_edge_over_cap, LATTICE_OVERMAX_SLACK}.
                 let current = connection_manager.connection_count();
-                if current >= connection_manager.max_connections {
+                if current >= connection_manager.max_connections
+                    && !connection_manager.admits_lattice_edge_over_cap(loc)
+                {
                     tracing::warn!(
                         %peer_addr,
                         current_connections = current,
--- a/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
@@ -321,7 +321,7 @@ impl P2pConnManager {
                 // connection_manager.rs::{admits_lattice_edge_over_cap, LATTICE_OVERMAX_SLACK}.
                 let current = connection_manager.connection_count();
                 if current >= connection_manager.max_connections
-                    && !connection_manager.admits_lattice_edge_over_cap(loc, Some(peer_addr))
+                    && !connection_manager.admits_lattice_edge_over_cap(loc)
                 {
                     tracing::warn!(
                         tx = %tx,
@@ -1183,7 +1183,7 @@ impl P2pConnManager {
                 // connection_manager.rs::{admits_lattice_edge_over_cap, LATTICE_OVERMAX_SLACK}.
                 let current = connection_manager.connection_count();
                 if current >= connection_manager.max_connections
-                    && !connection_manager.admits_lattice_edge_over_cap(loc, Some(peer_addr))
+                    && !connection_manager.admits_lattice_edge_over_cap(loc)
                 {
                     tracing::warn!(
                         %peer_addr,
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -80,6 +80,23 @@ pub struct RingStatsSnapshot {
     /// was at capacity (`MAX_TRACKED_PAIRS`). A non-zero value suggests
     /// identity churn / admission pressure, distinct from per-pair rate.
     pub updates_capacity_dropped: u64,
+    /// Nearest-neighbor ring lattice completeness (the "is greedy routing's base
+    /// lattice present" signal). `lattice_has_successor` / `_predecessor` are
+    /// whether this peer currently holds its closest-higher / closest-lower ring
+    /// neighbor; a peer with BOTH `true` has a complete both-sides lattice, and
+    /// the collector aggregates the fraction of such peers. The `_distance`
+    /// fields are the ring distance to each edge (None when unheld). Read from
+    /// the live connection set — no mirrored counter to rot.
+    pub lattice_has_successor: bool,
+    pub lattice_has_predecessor: bool,
+    pub lattice_successor_distance: Option<f64>,
+    pub lattice_predecessor_distance: Option<f64>,
+    /// Discovery health (route-to-self probe). `lattice_probes_issued` counts
+    /// probes fired since startup; `lattice_probe_improvements` counts those that
+    /// found a strictly-closer peer (filled or tightened an edge). A healthy peer
+    /// converges to both-sides-held and the improvement rate falls toward zero.
+    pub lattice_probes_issued: u64,
+    pub lattice_probe_improvements: u64,
 }
 
 static GOVERNANCE_PROVIDER: parking_lot::RwLock<Option<GovernanceProvider>> =
@@ -2219,6 +2236,7 @@ mod tests {
             updates_accepted: 1000,
             updates_rate_limited: 13,
             updates_capacity_dropped: 2,
+            ..Default::default()
         }));
         let snap = get_snapshot().unwrap();
         assert_eq!(snap.ring_stats.connection_count, 42);
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -82,19 +82,31 @@ pub struct RingStatsSnapshot {
     pub updates_capacity_dropped: u64,
     /// Nearest-neighbor ring lattice completeness (the "is greedy routing's base
     /// lattice present" signal). `lattice_has_successor` / `_predecessor` are
-    /// whether this peer currently holds its closest-higher / closest-lower ring
-    /// neighbor; a peer with BOTH `true` has a complete both-sides lattice, and
-    /// the collector aggregates the fraction of such peers. The `_distance`
-    /// fields are the ring distance to each edge (None when unheld). Read from
-    /// the live connection set — no mirrored counter to rot.
+    /// whether this peer currently HOLDS (a side is FILLED with) its
+    /// closest-higher / closest-lower connected ring neighbor; a peer with BOTH
+    /// `true` has a complete both-sides lattice, and the collector aggregates the
+    /// fraction of such peers.
+    ///
+    /// CAVEAT — filled != tight: a held edge may still be LOOSE (the exact
+    /// nearest is an unconnected peer between two adjacent peers), so the
+    /// both-sides-filled fraction OVERSTATES exact-nearest coverage. A peer
+    /// cannot locally determine tightness (that is what the route-to-self probe
+    /// discovers), so it is not reported as a boolean; compare the `_distance`
+    /// fields across peers to gauge looseness. The `_distance` fields are the
+    /// ring distance to each held edge (None when unheld). Read from the live
+    /// connection set — no mirrored counter to rot.
     pub lattice_has_successor: bool,
     pub lattice_has_predecessor: bool,
     pub lattice_successor_distance: Option<f64>,
     pub lattice_predecessor_distance: Option<f64>,
     /// Discovery health (route-to-self probe). `lattice_probes_issued` counts
-    /// probes fired since startup; `lattice_probe_improvements` counts those that
-    /// found a strictly-closer peer (filled or tightened an edge). A healthy peer
-    /// converges to both-sides-held and the improvement rate falls toward zero.
+    /// probes FIRED since startup; `lattice_probe_improvements` counts observed
+    /// lattice IMPROVEMENTS (a side filled or an edge tightened toward the true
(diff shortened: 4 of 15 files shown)


sanity:
**Review posture (Full-tier, held for morning):** this touches peer connection formation (`should_accept`), the topology maintenance loop, and swap/prune — Freenet's highest-risk surfaces — so it is intentionally NOT auto-merged and is held for a full multi-model review.

**Test layout:**
- **CI regression gates = the fast deterministic unit tests** (`topology.rs`, `connection_manager.rs`): per-side protected-set correctness, score-swap exemption, liveness eviction still evicts a lattice edge, over-max lattice admission. These run in normal CI.
- **The control-vs-fix simulation sweeps are `#[ignore]`-gated** — each is 12–24 serial sim runs (~15–17 min) and is cross-sim-sensitive (must run `--test-threads=1`), so they are validation measurements, not routine CI gates. Run them with:
  ```
  cargo test -p freenet --features simulation_tests,testing --test simulation_integration \
    -- --ignored --test-threads=1 --nocapture test_nn_lattice
  ```

**Honest status:** the fix delivers the core findability win (scatter-free PUT-reach to the closest peer 3/6 → 6/6; distant GET find-rate 0.875 → 0.917, with a dramatic 3/8 → 8/8 on one seed) and improves routing efficiency (fewer hops). The exact-nearest-coverage delta (+0.073) is real but modest and noisy at the validated 16-node/max=5 scale, and there is a mild long-link distribution skew toward medium distances plus a low absolute churn increase (4 → 8 disconnects). The 48-peer sim harness does not reliably form a ring even with extra gateways + longer settle — a separately-tracked harness limitation, reported rather than faked. Reviewers should weigh the coverage-noise and skew caveats.

[AI-assisted - Claude]

sanity:
## Full-tier multi-model review — PR #4760 (nearest-neighbor ring lattice)

Reviewed the checked-out code at HEAD `7bdf53ea` vs `origin/main`. Risk tier: **Full** (peer connection formation / topology — the codebase's highest-risk surface). Reviewers: an external **OpenAI Codex** (`codex-cli`) pass plus four independent adversarial Claude lenses (mechanism-correctness, topology-stability, security/eclipse, test-adequacy), each told not to rubber-stamp. Findings deduplicated and every `file:line` re-verified against the checked-out tree.

**Bottom line: MERGE-AFTER-FIXES.** No crash / regression / wire-format problem, and the per-side logic, the liveness-vs-score exemption split, dead-edge re-discovery, sign conventions, and panic-safety are all correct. But there is one concrete defect that all-Claude review missed and the external pass caught: **mechanism 3's over-capacity displacement is inert on the real CONNECT path**, and its unit test passes only because it bypasses that path. That plus the "guarantee"-vs-0.53-coverage gap and the fact that the feature is **unconditionally ON in production yet validated only at a 40×-sparser config** are what should be resolved before this ships.

---

### BLOCKING (must resolve or explicitly scope-down before merge)

**B1 — The over-max lattice admission is unreachable on the real CONNECT/promotion path; the mechanism's at-capacity displacement does not actually happen in production.**
`should_accept` force-accepts a strictly-closer per-side neighbor over the cap (`connection_manager.rs:701-731`, applied at `:829`) and `add_connection` has a matching over-cap exception (`would_be_new_lattice_edge` at `:931`, admit at `:1432-1453`). But the **only** production path into `cm.add_connection` is `Ring::add_connection` (`ring.rs:3077`, which adds no max gate) called from the two connection-lifecycle promotion sites — and **both gate on `current >= max_connections` and `return` before `add_connection` is ever reached**, with no lattice exception:
- `connection_lifecycle.rs:313-314` (transient promotion in `connect_peer`, before the `add_connection` at `:342`)
- `connection_lifecycle.rs:1164-1165` (`handle_successful_connection`, before the `add_connection` at `:1185`)
- (plus the outbound pre-flight at `:484-485`, which skips NAT dialing at max)

Consequences at capacity (the common production regime — defaults are `min=25`, `max=200`, and nodes routinely sit at max):
1. A strictly-closer nearest neighbor that appears after a node is full **never lands**. Mechanism 3's "displaces a long link" is inert on the real path. This is very likely a contributor to the disclosed exact-nearest coverage ceiling (~0.53).
2. The inbound force-accept still runs `should_accept` → a 60 s reservation + a NAT traversal, which the promotion then discards — wasted work that plausibly feeds the disclosed ~2× churn.
3. The unit test `add_connection_admits_lattice_edge_over_max` (`connection_manager.rs:2636`) gives **false confidence**: it calls `cm.add_connection` directly, bypassing the lifecycle cap gates that block the real path.

Fix, one of:
- **(a) Wire it through:** teach the two lifecycle cap gates (`:313`, `:1164`) the same `would_be_new_lattice_edge` exception, AND add an absolute over-max ceiling / synchronous non-lattice prune (see N-ceiling below) so admissions can't stack unbounded. Then add a test that exercises the *lifecycle* path, not just `add_connection`.
- **(b) Scope it down honestly:** drop the now-dead over-cap exception in `add_connection`, delete/relabel the misleading unit test, and state in the PR/comment that lattice edges form only below max (no at-capacity displacement yet).

(This is the reconciliation of Codex's P2 with the "no over-max ceiling" concern the security & correctness lenses independently raised: because the real path can't reach the over-max admit, the *unbounded-growth* worry is currently moot — but if you take fix (a) without the ceiling, it becomes live. Land both together.)

---

### NON-BLOCKING (should address before release)

**N1 — Probe stop condition stops on "filled", not "tight" → the 0.53 coverage is a real design ceiling, and "guarantee" is overclaimed.** `need_probe = sides_held < 2` (`ring.rs:5309/5311`) where `sides_held` counts a side holding *any* neighbor, not the *true* nearest. Two adjacent peers whose true-nearest is an unconnected peer between them both read "filled", so neither probes and the loose edges never tighten — a stable, scale-independent equilibrium. The route-to-self probe excludes only *connected* peers, so keeping it alive would terminate at (and force-accept) that closer peer; stopping abandons reachable tightening. At-max, a tight-edge probe is harmless (it's simply rejected if not closer); the short-link-skew hazard the comment cites only applies *under* max. At minimum correct the comment's "filled ⇒ done" conflation; better, keep probing at a backed-off interval while a side is loose. Also soften the PR/title "guarantee" to "seek/tighten" — 0.53 exact coverage is a *loose* lattice half the time (still a net gain over stock's no-base-lattice).

**N2 — Unconditionally ON in production, but validated only at `max_connections=5` / N≤16.** Production defaults are `min=25`, `max=200` (`ring.rs:474-476`); `nn_lattice_enabled()` folds to `const true`. All three sim sweeps are `#![cfg(feature="simulation_tests")]` **and** `#[ignore]` (`simulation_integration.rs:15035/15089/15209`), so none run in CI, and the 48-peer test is report-only ("harness does not reliably form a 48-node topology"). So the emergent dynamics at production density (hop distribution, churn, tail) are essentially unvalidated, and B1 means the dominant production regime (at-capacity) is exactly where the mechanism silently under-delivers. Recommend at least one production-like density data point (hundreds of peers, k≈200) or a staged/canary rollout before wide release.

**N3 — Mechanism 2 (route-to-self discovery + exponential backoff) has zero CI test coverage.** The backoff state machine, regression detection, and probe-gating live in the `ring.rs` maintenance loop (`:5296-5344`) and are exercised only by the `#[ignore]`d sims. The 5 fast unit tests cover the M3/M4 protection/eviction invariants but not this state machine — a regression here (wrong gating, backoff that never fires or never backs off) would pass CI. Add a deterministic unit test for the backoff/regression transitions.

---

### NITS

- **Telemetry reports "filled", not "tight"** (`p2p_impl.rs:538`, `lattice_has_successor: succ.is_some()`). The dashboard both-sides fraction reads ≈1.0 while true exact coverage is ~0.53, hiding the loose-lattice problem in production. Consider surfacing an exact-vs-loose signal.
- **Discovery `improvements/probes` counter is recorded at probe-issue time using pre-probe `sides_held`** (`ring.rs:5326-5335`), so it counts the first probe as an improvement even on failure and misses the fill that ends probing (next tick `need_probe==false`). Telemetry-only; the hits/probes metric will mislead under normal convergence.
- **Sim comment cites test names that don't exist** (`simulation_integration.rs:15243-15244`: `score_swap_never_drops_lattice_edge`, `prune_connection_evicts_lattice_edge`; actual names are `score_fallback_never_drops_lattice_edge` and `prune_evicts_lattice_edge_liveness_not_gated`). Also: the score-based **swap** and **resource-pressure removal** exemption sites have no direct unit test — only the fallback-drop path (`score_fallback_never_drops_lattice_edge`) and liveness (`prune_evicts_lattice_edge_liveness_not_gated`) are covered.
- **`is_protected` float tolerance `1e-12`** (`topology.rs:695`) can over-protect two distinct neighbors within 1e-12 ring distance — astronomically unlikely and errs toward over-protection (safe). Cosmetic.
- **Test thread-locals aren't reset panic-safely.** `set_nn_lattice_enabled` / `set_scatter_disabled` are restored at the end of each test/helper but not via a guard, so an assertion failure mid-test leaks the override to other tests on the same thread. Prefer a drop-guard.
- **"NOT for ship" experiment scaffolding merged to main.** `operations/findability_probe.rs` + the get/put `op_ctx_task.rs` hooks are labeled EXPERIMENT-ONLY; they are correctly `#[cfg(any(test, feature="testing"))]`-gated (verified: `testing` is not in `default` features, so they're compiled out of production). Merging author-flagged "not for ship" code into `main` is fine but worth a conscious call rather than drift.
- **Stale `# Panic` doc** on `should_accept` (`connection_manager.rs:591-593`) — the function guards `get_stored_location()` and returns without panicking. Pre-existing, but this PR edits the function.

---

### Security / eclipse — ACCEPTABLE WITH DISCLOSURE (not a blocker)

The prompt's "grind source ports/IPs to eclipse a victim" is largely defanged: the candidate location fed to `should_accept` is **address-derived** (`Location::from_address`), which masks host bits — a whole IPv4 /24 collapses to one location (v6: /48), and port only participates for loopback. Distinct locations require distinct /24 (v4) or /48 (v6) prefixes; this PR does not change that Sybil cost. The operator blocklist is checked in `connect.rs` **before** `should_accept`, so `nn_override` cannot bypass it. The genuine incremental risk is (1) admission became *deterministic* (a very-close peer was often *rejected* by the Kleinberg evaluator pre-PR, not accepted, per the PR's own motivation), and (2) the retention exemption removes the score-based swap as a safety valve, leaving **liveness eviction only** — so a responsive-but-malicious nearest neighbor that answers pings but drops/corrupts payloads is never shed (though the pre-PR score swap didn't detect payload-lying either — a narrow regression). Mitigations hold: the exemption only ever protects whoever is *genuinely* per-side nearest, and a closer honest peer force-displaces the attacker. Recommend a one-line disclosure in the design doc and a future follow-up for a routing-correctness signal. **Telemetry is clean**: the new `RingStatsSnapshot` fields are read from canonical live state (no counter to rot), the struct is server-rendered to HTML and is **not** `Serialize` (no JSON/non-string-key wire risk), and `{d:.4}` only formats finite `sd.abs()` values.

### Disclosed-caveat verdicts
- **Churn ~2× (4→8):** ACCEPTABLE. Traced the force-accept↔prune loop — no limit cycle / oscillation (a pruned peer is non-lattice by construction and can't re-displace). The 2× is a small-k artifact (displacing 1 of 3 long links at k=5 vs 1 of ~198 at k=200).
- **Long-link skew:** ACCEPTABLE. The concern is inverted — the 2 reserved edges *are* the short-range base lattice Kleinberg's O(log²n) bound requires; at k=200 that's 2/200, negligible. Degenerate only at `max ≤ 2` (0 long links), which is theoretical at prod max=200.
- **Exact-cycle coverage 0.53:** a REAL design ceiling (per N1), not just 6-seed variance. Net improvement over stock, but the "guarantee" claim isn't met — non-blocking, soften the claim.
- **16-peer-only validation:** the biggest asterisk (per N2). Combined with B1, the at-capacity production regime is both partly-inert and unvalidated. Get a density data point or gate the rollout.

Did not merge and did not modify code — review only, as requested.

[AI-assisted - Claude]

sanity:
## Full-tier re-review — blocker fix + cleanups (HEAD `208b9a34`)

Focused re-review of the changes since the prior MERGE-AFTER-FIXES pass. External model: **Codex** (`codex review --base main`) ran first and caught two P2s; plus two adversarial Claude reads (ceiling/concurrency; liveness/eclipse/retention). All three read the actual checked-out code at HEAD `208b9a34`.

### Bottom line: MERGE-AFTER-FIXES

The **B1 blocker is genuinely fixed** and the over-max **ceiling is safe** (established connections hard-bounded at `max+2`; liveness eviction untouched; no oscillation) — the two things this re-review was asked to confirm both check out. But three things stand between here and merge/canary:

1. **CI is deterministically RED** — 5 sim tests fail (3/3 retries each), red on all branch commits, green on main. Hard merge gate.
2. **F1 [HIGH]** — at-capacity reservation/NAT amplification on an empty lattice side (newly activated by this fix; on the #4630 DoS surface). Cheap fix.
3. **F3 [HIGH]** — the eclipse disclosure understates the IPv6 Sybil cost for the deterministic un-prunable force-accept; the mechanism ships on-by-default. Needs Ian's explicit accept ± disclosure correction (his sign-off class).

Plus F2/F4 [MEDIUM] and F5/sub-notes [LOW]. None of the code findings breach the ceiling or break liveness — both adversarial reviewers independently confirmed those hold — but F1 and F3 are worth resolving before shipping an on-by-default connection-formation change to the real network. I did NOT merge or modify code.

---

### ✅ Blocker (B1) genuinely fixed — verified

Both connection-lifecycle promotion gates now apply the shared predicate before the `>= max_connections` reject:
- `connection_lifecycle.rs:323-325` (transient promotion) — `current >= max && !admits_lattice_edge_over_cap(loc)`, `loc` correctly derived from the transient/address location.
- `connection_lifecycle.rs:1185-1187` (handle_successful_connection promotion) — same, `loc` from pending/address location.
- `connection_manager.rs:874` (`should_accept`) and `:1515` (`add_connection`) apply the same predicate.

The source-scrape pin `lifecycle_gates_apply_lattice_over_cap_predicate` asserts ≥2 gate call sites; I confirmed exactly 2 real matches of `admits_lattice_edge_over_cap(loc)` (comment mentions use a different form, so no false match). All production adds funnel through `ring.rs:3092` → `connection_manager.add_connection`, the authoritative gate — no bypass path.

### ✅ Ceiling is safe (established connections) — verified

`admits_lattice_edge_over_cap` = `would_fill_empty_lattice_side(loc) && connection_count() < max + LATTICE_OVERMAX_SLACK(2)`. `add_connection` (`:1514-1536`) is authoritative: at `>= max` it admits only when the predicate holds, else rejects + rolls back. Fill-only means ≤2 empty sides, and the ceiling independently caps at `max+2`. So **established ring connections are hard-bounded at `max+2`** and converge back to `max` via the over-max prune (`topology.rs:551` → `select_fallback_peer_to_drop`, which never sheds a protected lattice edge, falling back to the unfiltered set only in a degenerate tiny-max config). No oscillation: `would_fill_empty_lattice_side` returns false once a side has any neighbor, so a filled side can't be repeatedly over-cap-displaced.

Note on the sim "peers top out at exactly 5" claim: the **code-enforced** hard ceiling is `max+SLACK` (=7 at max=5), reached transiently between ticks; "exactly 5" is a steady-state observation (both sides usually fill via normal under-max admits, so over-cap fills rarely fire), not the hard bound. Accurate to say the ceiling is bounded; imprecise to say code pins it at exactly max.

### ✅ Liveness eviction untouched — verified

The retention exemption lives only in the score-based paths (`topology.rs` swap/prune/resource-pressure). `prune_connection`/`prune_alive_connection` (liveness) do NOT consult `protected_nearest_neighbors`; test `prune_evicts_lattice_edge_liveness_not_gated` (`connection_manager.rs:2696`) pins it. No unbounded GC exemption — a lost lattice edge is evicted and re-discovered (probe with backoff). Probe is bounded: stops once both sides filled (`lattice_should_probe`), exponential backoff (prod tau0=5s → tau_max=300s) with ±20% jitter, resets to immediate on a side loss. No thrash.

### Eclipse disclosure — accurate on IPv4, understates IPv6 (see F3)

The security comment at `should_accept` (`connection_manager.rs:726-740`) is accurate on the mechanics: operator blocklist IS checked in `connect.rs:803-811` **before** `should_accept` (`:816`) on every reachable inbound path; location is address-derived. But the Sybil-cost bound it states understates the IPv6 case — see F3.

---

### 🟠 Findings to address

**F1 [HIGH — Codex P2 + Claude adversarial] Over-cap admission ignores pending reservations → at-capacity reservation/NAT amplification on an empty side.** `connection_manager.rs:874`. `admits_lattice_edge_over_cap` reads established-only state (`connection_count()` / `would_fill_empty_lattice_side` → `connections_by_location`), invisible to pending reservations, and `should_accept` inserts the reservation *before* deciding (`:669-672`). With an empty lattice side at max, EVERY concurrent candidate on that side passes over-cap (it checks side-emptiness, not closeness) and is accepted + reserved + NAT-worked, until the first promotes and fills the side. Dedup/insert key on exact `SocketAddr` (`:656`,`:671`), not location, so one /24 yields many distinct addrs at the same on-side location, each passing. The **established ceiling still holds** (add_connection is serialized and rejects the rest at `:1515`), so this is NOT unbounded connections — but it is a **regression in at-capacity DoS resistance on the exact connection-formation surface (#4630)**: pre-PR an at-max node returned false and removed the reservation (zero reservations created for new candidates); post-PR an at-max node with an empty side accepts and reserves for every on-side candidate. Self-correcting (TTL 60s) and the window is narrow (side genuinely empty AND at/near max, e.g. just after a nearest-neighbor drop), hence HIGH not blocker. Clean fix: make the over-cap clause count in-flight fills — use `total_conn < max + LATTICE_OVERMAX_SLACK`, or have `would_fill_empty_lattice_side` treat a pending reservation on that side as already-filling. Recommend fixing before the canary (this fix newly activated it, and it is the DoS surface).

**F3 [HIGH — Claude adversarial] Eclipse disclosure understates the IPv6 Sybil cost; mechanism ships on-by-default.** The change converts per-side-nearest admission from probabilistic-reject to **deterministic force-accept** (`connection_manager.rs:741-762`) AND removes score-based swap as the eviction lever (retention exemption), so an attacker who becomes a victim's genuine per-side nearest **captures both local slots un-prunably** — only liveness can churn them, which a responsive attacker defeats. Blast radius is the victim's 2 local slots / last-mile to nearby keys, NOT a full partition (long links untouched; over-max prune sheds the most-redundant peer, not unique honest long links), and it requires targeting a specific victim — hence HIGH, not blocker. The disclosure (`:726-741`) frames the Sybil bound as "distinct /48 prefixes," but a single routinely-assigned IPv6 allocation (/32 or /29 to one operator) contains thousands-to-millions of distinct /48s, each a **free** distinct `Location`, so an attacker can cheaply brute-force /48s that hash arbitrarily close to a target and reliably land as its exact per-side nearest. On IPv4 the /24 bound is meaningfully harder. F2 amplifies this (extra attacker peers colliding into the same /48 bucket are also protected). This is a placement/topology-class change (per project convention, Ian signs off): recommend he **explicitly accept the eclipse tradeoff with the IPv6 cost corrected in the disclosure**, and consider prioritizing the "routing-correctness eviction signal" the comment defers, given the mechanism is on-by-default.

**F2 [MEDIUM — Codex P2 + Claude] Retention exemption over-protects all peers at a masked Location.** `topology.rs:1103-1106` (`protected_nearest_neighbors` returns `Vec<Location>`); `connections_by_location` is `BTreeMap<Location, Vec<Connection>>` and `Location::from_address` collapses a whole /24 (or /48) to one Location. So if the nearest successor/predecessor Location holds multiple distinct peers, ALL are exempt from score-based swap/prune (`select_connections_to_remove:770`, `maybe_swap_connection:896`, `select_fallback_peer_to_drop:1143`), not just one lattice edge. Harm: redundant same-location peers consume slots and force pruning of unrelated long links, defeating score-based topology adaptation for that bucket (datacenter/address-clustered deployments, or an attacker holding a /24–/48). Safety fallbacks prevent a permanently-stuck node (fallback removal unfilters when all protected `:1158`; resource-pressure path falls through to it `:819-826`; both run above min). Adversarially inducible, and it amplifies F3. Fix: protect one specific per-side-nearest *connection*, not the whole Location bucket.

*Sub-note [LOW]:* the three exemption sites are inconsistent — `select_connections_to_remove` matches by signed distance with `(p - d).abs() < 1e-12` (`topology.rs:695`) while the other two use `Location` equality. Prefer matching the peer/location directly everywhere.

**F4 [MEDIUM — Claude adversarial] Ceiling safety rests on an implicit single-task invariant; no assertion pins it.** The max+2 ceiling holds today only because every established-count-increasing path runs inline and serially on the one p2p event-loop task, `add_connection` is synchronous (no await), and there is no await between the gate's `connection_count()` check and the `add_connection` call. But inside `add_connection` the cap check (`:1514`) and the insert (`:1560`) release the `connections_by_location` lock in between — a non-atomic check-then-act. If a future refactor ever moves promotion onto its own task or parallelizes handshake processing, two adds could each observe `count == max+1 < max+2` and both insert → max+3. Recommend a `debug_assert!(connection_count() <= max_connections + LATTICE_OVERMAX_SLACK)` after the insert and a comment pinning the single-task assumption, so the invariant fails loudly if the assumption is ever broken.

**F5 [LOW — comment accuracy] Over-max "next tick" is really ~120s.** The gate/`add_connection` comments say the over-max prune sheds a link "next tick." `adjust_topology` removes one peer per tick (`topology.rs:561`) and `CHECK_TICK_DURATION` is 60s (`ring.rs:4513`), so returning from max+2 to max takes ~2 ticks ≈ 120s. The ceiling is not affected (adds are blocked at max+2 regardless of prune speed — the prune is not load-bearing for the bound), but "transient" here means up to ~2 minutes. Minor wording fix.

---

### 🔴 Merge gate: CI is deterministically RED (dominant issue)

`Simulation` fails on HEAD — **5 tests, each TRY 3 FAIL (failed all 3 retries = deterministic, not flaky)**, and the same check is red on **all three branch commits** (incl. the base feature `7bdf53ea`, so this predates the blocker fix). The same `Simulation` job is **green on main** at the merge-base `55f647e5`, and these 5 tests exist on main:

- `simulation_integration::test_piece_e_findability_sparse_ring_gate` — honest GET findability gate, MEAN_FLOOR 0.60
- `simulation_integration::test_get_retry_with_alternatives_sparse_topology` — asserts ALL nodes find the contract
- `simulation_integration::test_subscription_chain_collapses_on_client_leave` — precondition `subscribed_peers >= MIN_MESH` (mesh must form)
- `streaming_e2e::test_streaming_get_through_relay`
- `streaming_e2e::test_streaming_get_retries_after_assembly_failure`

None of the 5 toggle the lattice off — they run with the production default (lattice ON). All 5 are reach / topology-formation sensitive, consistent with the acknowledged small-scale regression (2 reserved edges out of max_connections=5 = 40% of the budget, leaving only 3 long links). This is plausibly the "known scale artifact" the canary is meant to validate — but note the breadth: it is not only the findability numbers, it also includes streaming reach and subscription-mesh formation not reaching MIN_MESH.

**This blocks merge regardless of the code verdict** (no-merge-with-red-ci): merging red would poison main's baseline for the next agent (5 deterministic failures they can't distinguish from their own regressions). Before merge/canary, one of: (a) gate these 5 for the new topology (e.g. run the lattice OFF, or update the thresholds) with a written justification that the new behavior is correct at scale; (b) fix the code so the scenarios still pass; or (c) explicit per-action user authorization to ship the canary with red sim CI. Leaving them silently red is not acceptable.

Also red: `rule-review/warnings` (1 warning — "fix or add review-override label"; override-able, likely the high-risk-surface flag).

---

### Cleanups from prior review — spot-checked ✅
Softened "guarantee" wording (PR title now "seek+tighten"); eclipse disclosure at the force-accept site; `.claude/rules/ring.md` documents the over-max exception; split issued/improvement telemetry counters with "filled != tight" caveat; panic-safe `NnLatticeTestGuard` drop-guard (`simulation_integration.rs:14765`); stale `# Panic` doc corrected. All present.

[AI-assisted - Claude]

sanity:
## Finalization pass: budget-gate + F1/F3/F4 fixups

Pushed `d42ddb30` finalizing this for release. Summary of what changed since the last review, and one decision that needs a maintainer eye.

### 1. Minimum-degree floor (greens CI)
The 5 CI simulation failures were a **scale artifact**: reserving 2 of `k` slots is fine at production degree but destabilizes routing at the sim's sparse scales. Added the single activation gate `nn_lattice_active_for(max_connections)` / `ConnectionManager::nn_lattice_active` = `flag && max_connections >= NN_LATTICE_MIN_MAX_CONNECTIONS`, applied at **all** feature sites (per-side acceptance clause, over-cap admission, discovery probe, topology retention exemption).

**Threshold = 25 (`DEFAULT_MIN_CONNECTIONS`).** Note the gate reads the **configured** `max_connections` (200 in production), not the live connection count — so this is **production-neutral** (every real node is above any reasonable floor and stays active). Its only real effect is which *sim* tests exercise the feature. The sim configs top out at `max_connections=16`, so a floor of 25 keeps the whole sparse sim suite at its pre-lattice baseline (only the production-scale max=200 sim tests exercise the lattice). A lower floor of 15 was evaluated and rejected — the sim's **default** config is max=15/16, so a 15 floor flips those default hosting/subscription/determinism tests to feature-ON (a change from baseline) for zero production benefit. Details in the const doc comment.

### 2. Preserved ON-path sim coverage
`set_nn_lattice_enabled(true)` also force-activates below the floor (test-only, per-thread), so the dedicated control-vs-fix benefit tests still exercise the mechanism at max=5.

### 3. F1 [HIGH] — reservation-aware over-cap admission
`admits_lattice_edge_over_cap` now refuses a fill when another in-flight pending reservation already targets the same empty side (`pending_reservation_on_side`), bounding concurrent over-cap fills to one per side. Candidate addr is plumbed through `should_accept`, `add_connection`, and both lifecycle promotion gates. New regression test `admits_lattice_over_cap_is_reservation_aware`.

### 4. F4 — ceiling `debug_assert`
Post-insert assert in `add_connection` documenting that the over-cap ceiling's non-atomic check-then-insert is safe only under the single-task connection-add invariant.

### 5. F3 — corrected eclipse disclosure
Code comment + PR body now state the IPv6 cost accurately: a single routed IPv6 allocation yields vast free /48 `Location`s, so targeted last-mile eclipse of a victim's 2 lattice slots is cheap — **bounded to that victim's 2 slots, not a partition**. Project lead has explicitly accepted this tradeoff.

### F2 [MEDIUM] — documented follow-up
Retention protection is keyed by masked `Location`, so peers colliding on a /24-/48 mask are both exempted. Narrowing to the exact edge peer isn't cheap (the topology score/swap/prune layer is Location-keyed); documented as a bounded, rare efficiency nit rather than reworked on this high-risk surface.

### Verification
- The **5 previously-failing sim tests pass gated-off** at 25 (they run at max 3–6): `test_get_retry_with_alternatives_sparse_topology`, `test_piece_e_findability_sparse_ring_gate`, `test_subscription_chain_collapses_on_client_leave`, `test_streaming_get_through_relay`, `test_streaming_get_retries_after_assembly_failure`.
- All lattice/retention unit tests + the new F1 test pass.
- `cargo fmt` + `cargo clippy -p freenet -- -D warnings` clean (default + testing features).

### Force-enabled benefit test (honest result)
At max=5 (force-activated), **PUT-reach passes: fix lands rank-0 5/6 vs stock 3/6**, fixing the decisive dead-ends (seed 0007: stock rank 5 → fix rank 0; seed 000B: stock rank 6 → fix rank 3; seed 000B GET: stock 3/8 → fix 7/8). **However the mean-GET guard fails by 0.004** (fix 0.771 vs stock 0.875, tolerance 0.10), driven entirely by seed 0003 where the fix arm collapsed to 0/8. This traces to the F1 change (reservation-aware over-cap serializes fills) interacting with the sparse max=5 regime — a regime now **below the production floor** where the feature is disabled, so it does not affect production. I did **not** weaken the test to hide this. Open question for a maintainer: accept it as a below-floor max=5 variance artifact of the required F1 fix, or investigate seed 0003 further?

CI is the final arbiter on the sim suite (my local box is loaded, so a couple of 240s-borderline tests — `test_placement_migration_at_scale`, which times out feature-OFF too — hit the cap locally; they pass on CI's dedicated runners).

[AI-assisted - Claude]

sanity:
## Full-tier multi-model review — PR #4760 @ `a503a5f5`

Highest-risk surface (topology / routing / concurrency), release-bound. Reviewers: external **Codex (gpt-5.5)** first, then four adversarial Claude lenses (tightening+concurrency, churn/stability+discovery-decay, F1-drop/DoS, test-adequacy). Reviewed the checked-out code, not just the diff.

**Bottom line: MERGE-AFTER-FIXES.** No production correctness bug found across any lens — the tightening/ceiling/prune logic is sound and the churn/stability risk is genuinely bounded and convergent (details below). The fixes are a test-hygiene leak (flaky-test risk), one misleading security doc, and a vacuous-pass test; plus one prominent coverage gap to close or explicitly defer.

CI is green on all heavy jobs at this HEAD (Simulation, Unit&Integration, Clippy, Fmt, NAT, Windows/macOS); only the advisory `rule-review/warnings` soft-fails.

---

### Verified sound (the core mechanism)

- **Over-cap ceiling is a hard, atomic bound.** In `add_connection` the relocation-removal → count → admit-check → insert → `debug_assert` all run under one `connections_by_location` write guard; `admits_lattice_edge_over_cap_in` reads that same held map. Post-insert `count ≤ max + LATTICE_OVERMAX_SLACK(2)` on every path (net-add, relocation net-zero, non-lattice, nn-inactive). Lock ordering preserved (`location_for_peer` #1 released before #2 acquired; reject path `drop(cbl)` before touching #1). (`connection_manager.rs` add_connection block; both lifecycle gates `connection_lifecycle.rs:322,1184` apply the shared predicate, pinned by `lifecycle_gates_apply_lattice_over_cap_predicate`.)
- **Derived-nearest is correct.** `nearest_lattice_neighbor_dist_in` is computed before the candidate is inserted, excludes `sd == 0.0`, and re-derives per-side nearest from the live set — so a strictly-closer arrival becomes the new nearest and the former-nearest demotes to the long-link pool with no separate bookkeeping. `protected_nearest_neighbors` re-derives identically, so retention tracks the new edge.
- **Churn/stability is bounded and convergent.** Each per-side over-cap admit requires strict `<` vs the current established nearest (a strictly-decreasing, finite sequence → terminates); the prune never evicts a protected edge, so nearest distances only decrease via tighten, never increase via prune — no A↔B limit cycle on a node. Over-max prune sheds exactly one non-lattice link per tick and converges to max within ≤2 ticks; `pick_least_important(true).or_else(|| pick_least_important(false))` is a correct never-stuck fallback. `signed_distance` semantics (`other-self>0` ⇒ successor side) match the code.
- **Discovery decay works, no probe storm.** `lattice_probe_progress` classifies improve/regress/plateau correctly; `backoff_attempt` grows on plateau and resets only on real improvement/regression (which requires an actual connection-set change, so `improved` can't latch true). Probe fires at most once per maintenance tick, jittered; `ExponentialBackoff` caps the exponent and saturates at `tau_max` — no overflow/zero-interval/negative path.
- **EXPERIMENT-only findability_probe is correctly gated.** The module and every call site are `#[cfg(any(test, feature="testing"))]` with a `false` production fallback for `consult_suppressed`; any ungated reference would fail the production build (compiler-enforced). Not shipped.

---

### Findings

**[Should-fix — flaky-test risk] Test-only `NN_LATTICE_FORCE_ACTIVE` leaks across tests.** `connection_manager.rs:2872` (Codex, confirmed). `set_nn_lattice_enabled(true)` also sets `NN_LATTICE_FORCE_ACTIVE=true`, which is *not* the production default (`enabled=true, force_active=false`). Unit tests that end with `set_nn_lattice_enabled(true)` — `connection_manager.rs:2872, 2927(end), 2965(end), 3072(end), 3114` and the `topology.rs` lattice tests — leave the floor-bypass armed on the current test thread. Under `cargo test` (shared-process thread pool) a later sparse-`max<25` test that reuses that thread runs with the lattice force-ON instead of the production gate → order-dependent accept/cap behavior. The integration test already does this right (`simulation_integration.rs:14774-14775`: `set_nn_lattice_enabled(true)` **then** `set_nn_lattice_force_active(false)`); the unit tests should restore both (or use a drop guard). Doesn't affect production; does risk CI flakiness.

**[Should-fix — security doc + cheap mitigation] Over-cap admission is not reservation-aware; the doc claim is wrong.** `admits_lattice_edge_over_cap_in` (`connection_manager.rs`) counts only ESTABLISHED connections, and `should_accept`'s over-cap OR-clause (`:979`) bypasses the speculative `total_conn >= max_connections` reject (`:883`). On accept, `should_accept` inserts a pending reservation (`:764`) and calls `record_pending_location` (`:1000`), neither of which advances the established nearest — so the "each admit advances the nearest, sequence strictly decreasing" self-bound only holds *after establishment*. A node at max ESTABLISHED, fed strictly-closer locations from fresh IPv6 /48s (the already-accepted cheap-minting capability), accepts each over-cap (established `= max < max+2`), records pending state, and emits an AcceptOutcome + NAT-init — where a pre-lattice full node rejected all at `:883` and produced zero. The self-limit at `:749` only dedups by exact addr, so fresh addrs evade it. **Bounded** by `PENDING_RESERVATION_TTL` (60s) × rate (not a reflection amplifier — the joiner addr is packet-source-derived), so **non-blocking** and within the accepted eclipse tradeoff — BUT the `admits_lattice_edge_over_cap_in` rustdoc claim that the residual "degrades to the generic connection-flood case (bounded by … per-target connection backoff)" is inaccurate: per-target backoff keys on address and fresh addresses evade it, and the generic flood against a full node yields zero accepts. Fix the doc; and consider the cheap mitigation of counting non-stale `pending_reservations` toward the over-cap `count` (or capping the reservation map), which restores F1's effect and matches the ring.md "over-committed checks use speculative state" invariant, without touching the tightening behavior.

**[Recommend — coverage gap on the production regime] No end-to-end test at `max_connections >= 25`.** The feature is default-ON and gated to `max>=25`, but every multi-node sim tops out at `max=15-16`; the three nn-lattice sim tests run at `max=5` with force-enable and are BOTH `#[ignore]`d AND report-only (`simulation_integration.rs` ~14886-15279), so they contribute zero CI signal. The one kept hard assert (`mean_fix >= mean_stock - 0.05`, `:15097`) is a non-collapse floor, not a mechanism check — it would pass trivially if tightening were a no-op. The "production-scale max=200 tests that DO exercise the lattice" are the `topology.rs` unit tests with synthetic 2-5-neighbor maps, not end-to-end. Per test-philosophy.md ("Don't discover logic errors in production") and the Simulation Health Metrics rule for routing/topology PRs, add at least one end-to-end sim at `max>=25` asserting a real topology/findability outcome (nightly is fine), OR get explicit maintainer sign-off to defer it. The report-only conversions themselves are honest (each `warn!`s the expected max=5 regression); the gap is that the net-win-at-scale hypothesis is validated nowhere automated.

**[Nit] `select_connections_to_remove_never_drops_lattice_edge` can pass vacuously.** `topology.rs` (new test): the `assert_ne!`s live inside `if let RemoveConnections(peers) = adj`, so a `NoChange` return passes without asserting. The sibling swap test correctly asserts `saw_swap` — do the same here (assert a removal was produced).

**[Nit] F4 concurrent-add atomicity has no regression test.** The comment cites "stress tests breached it to 60+" but no concurrent-add test is in the diff; only the `debug_assert` guards it. Atomicity is now structural (single guard) so a deterministic race test is hard, but a note or a targeted test would harden the `fix:`.

---

### Non-blocking design notes (no action required)

- **Shared slack budget (2), not per-side.** A same-side tighten-burst can consume both slack slots, so a genuinely-empty opposite-side FILL is refused for 1-2 ticks until the prune frees a slot (the PR's own test documents this). Bounded, self-heals; transient weakening of the "both edges guaranteed" framing.
- **Asymmetric-nearest is best-effort.** Over-cap admission fires only for a node's OWN per-side nearest, never because a neighbor claims this node as its nearest; and a node's non-protected long link may be another node's protected edge (P prunes it, X's discovery re-adds it). Both are bounded/self-healing, but the "guaranteed successor+predecessor" is really "guaranteed unless the nearer peer is full and doesn't reciprocate." A per-edge add/drop churn counter in the sim harness would be worth having before relying on it at scale.
- **`update_peer_identity` placeholder branch is an uncapped add site** that bypasses the cap check; it's the one path that could push `count` past the ceiling and trip the `debug_assert` (debug-only, contrived — requires map-drift recovery to coincide with a saturated ceiling).

---

Reviewers: Codex gpt-5.5 (external, ran first) + 4 Claude adversarial lenses. No lens found a production correctness blocker. Churn/stability: **acceptable for release** (bounded + convergent). F1-drop: **safe** (no NEW unbounded path; TTL-bounded, within accepted tradeoff) modulo the doc fix.

[AI-assisted - Claude]

sanity:
## Re-review of the over-cap reservation-counting change (heads `8c8ea2c3` / `2b5475a9`)

Since the prior Full-tier review (at `a503a5f5`), three things landed and were re-reviewed:

1. **Review-polish commit** (`79613546`) — the prior review's own asks: NN-lattice test-override drop-guard isolation (10 sites), honest security rustdoc, de-vacuoused removal test, and a new F4 atomic-ceiling regression test.
2. **Mitigation** (`8c8ea2c3`) — the over-cap ceiling now counts **non-stale pending reservations** in addition to established connections, per the `ring.md` speculative-state invariant ("am I over-committed?" checks use speculative state). This closes the previously-disclosed residual where a full node re-fired an accept + NAT-traversal on every strictly-closer fresh-IPv6 request, bounding it to `LATTICE_OVERMAX_SLACK` (2) per reservation-TTL window. It is a global speculative-count term, distinct from and simpler than the dropped per-side F1 clause.
3. **Review follow-up** (`2b5475a9`) — test + doc only (production logic unchanged from `8c8ea2c3`).

### Focused re-review (this touches over-cap acceptance concurrency — a high-risk surface)

Two independent reviewers, both told to assume a bug exists and focus on lock-ordering, the hard-ceiling invariant, the self-count asymmetry, starvation, and test validity:

- **External model (Codex):** *No CRITICAL/HIGH/MEDIUM production findings.* Confirmed lock order safe (`#2 connections_by_location → #3 pending_reservations`, no `#3→#2` path), the bound conservative in `should_accept` (self-count is stricter, never admits past the ceiling), no stale-reservation or time-source issue. One **LOW**: the regression test exercised the predicate directly, not the production `should_accept` path.
- **Adversarial Claude read:** *Correct — no deadlock, no hard-ceiling violation, valid non-vacuous test.* Traced all lock acquisitions and both call paths; verified no re-entrant `#2` double-read, F4 atomicity intact, self-count asymmetry correct (each path counts the candidate's slot exactly once), no permanent starvation (timestamp-based TTL filter drains within 60s). Two **LOW/INFO**: (a) the SLACK budget is *global* (shared with unrelated in-flight handshakes), not lattice-private; (b) a `was_reserved == false` double-count edge is over-count/safe-direction only.

### Resolutions (all findings addressed)

- **Codex LOW** → added `should_accept_over_cap_throttles_after_slack_pending`, which drives the production path and fails if the self-count or reject-path cleanup regresses (non-vacuous: without the term the second candidate would be admitted).
- **Claude LOW (global budget)** → documented honestly in the predicate, its rustdoc, and `ring.md`: the budget is global, so at-capacity tightening can be throttled by unrelated handshakes until they drain (bounded by TTL, retried by continuous discovery). This bites only nodes genuinely at max (mostly busy gateways); peers below max tighten via the under-cap path, which never consults this ceiling. A lattice-private budget is noted as a possible future refinement.
- **Claude INFO** → documented the `was_reserved == false` edge as over-count/safe-direction (retryable reject at worst, never a ceiling breach).

### Coverage note (production-density validation deferred to canary)

There is no end-to-end simulation at production connection density (`max_connections ≥ 25`, where the lattice gate activates) because the simulation harness cannot reliably form a ring that large — which is the reason for the real-network canary. The *mechanism* is unit-tested (acquires-closer-at-capacity, cycle-completeness, the over-cap ceiling bound, the atomic ceiling under concurrent adds). End-to-end production findability is validated by the **Framework laptop canary** (with try.freenet.org auto-rebuilding from main as a second real node), before general release.

[AI-assisted - Claude]

PR #4768, opened 2026-07-12
feat(telemetry): wire nearest-neighbor ring-lattice gauges into central telemetry

sanity:
## Problem

The six nearest-neighbor ring-lattice metrics from the #4760 lattice fix are computed on every peer but only feed that peer's LOCAL home page (`RingStatsSnapshot`, rendered per-request). There is no way to measure the fix's impact NETWORK-WIDE. As 0.2.96 rolls out we want to answer, as a one-line query over central telemetry:

- what fraction of peers hold BOTH immediate ring-neighbor edges (a complete both-sides lattice),
- the median held-edge distances (successor / predecessor), and
- the route-to-self probe success rate.

None of that is currently reachable centrally.

## Solution

Additive / observability-only. Extend the existing periodic per-peer `RouterSnapshot` central event — it already fires on a cadence and carries `peer_id`, so aggregation needs no new plumbing — with the same six gauges, following the existing placement-gauge pattern exactly (no behavior change). Per CONTRIBUTING this needs no prior issue.

The six metrics:
- `lattice_has_successor`, `lattice_has_predecessor` (`Option<bool>`)
- `lattice_successor_distance`, `lattice_predecessor_distance` (`Option<f64>`)
- `lattice_probes_issued`, `lattice_probe_improvements` (`Option<u64>`)

Changes:
1. **`crates/core/src/router.rs`** — added the six `Option` fields to `RouterSnapshotInfo`, grouped with the placement-quality gauges for greppability (all `None` in the default constructor, populated by `Ring` — same convention as `open_fds` / `hosted_key_distance_median`).
2. **`crates/core/src/ring.rs`** (`emit_router_snapshot_telemetry`, right after the fd gauges) — populate them via the SAME `connection_manager` calls the home-page provider uses: `nearest_lattice_neighbor_dist(true)` (successor) / `(false)` (predecessor) and `lattice_probe_stats()`. Identical `None`/`0` semantics: a distance is `None` when that side is unheld or own location is unknown; probe counters are `0` before any probe fires.
3. **`crates/core/src/tracing/telemetry.rs`** — mirror the six fields into the hand-written OTLP `json!` body next to `hosted_key_distance_*`. `Option` fields emit `null` when unset, matching the existing distance gauges.

Field names are snake_case and `lattice_`-prefixed for greppability.

## Testing

Extended the pin test `router_snapshot_json_includes_placement_gauges` to assert all six new fields reach the OTLP body — with `lattice_has_successor = Some(true)` and `lattice_has_predecessor = Some(false)` so both boolean states are exercised and shown distinguishable from `null`.

- `cargo fmt` clean
- `cargo clippy --locked -p freenet -- -D warnings` clean
- all 14 `router_snapshot` lib tests pass (`cargo test -p freenet --lib router_snapshot`)

Refs #4760, #4642

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1641,6 +1641,26 @@ impl Ring {
             snapshot.open_fds = open_fds;
             snapshot.fd_soft_limit = fd_soft_limit;
 
+            // Nearest-neighbor ring-lattice completeness + probe health (#4760),
+            // mirrored from the home-page ring-stats provider (see
+            // `node/p2p_impl.rs`) onto the central-telemetry snapshot cadence so
+            // the lattice fix's network-wide impact — fraction of peers holding
+            // both immediate-neighbor edges, median edge distances, probe success
+            // rate — is a one-line query over central telemetry as 0.2.96 rolls
+            // out (#4642). Same calls, same `None`/`0` semantics as the home page:
+            // a distance is `None` when that side is unheld or own location is
+            // unknown; the probe counters are `0` before any probe fires.
+            let cm = &ring.connection_manager;
+            let lattice_successor = cm.nearest_lattice_neighbor_dist(true);
+            let lattice_predecessor = cm.nearest_lattice_neighbor_dist(false);
+            let (lattice_probes_issued, lattice_probe_improvements) = cm.lattice_probe_stats();
+            snapshot.lattice_has_successor = Some(lattice_successor.is_some());
+            snapshot.lattice_has_predecessor = Some(lattice_predecessor.is_some());
+            snapshot.lattice_successor_distance = lattice_successor;
+            snapshot.lattice_predecessor_distance = lattice_predecessor;
+            snapshot.lattice_probes_issued = Some(lattice_probes_issued);
+            snapshot.lattice_probe_improvements = Some(lattice_probe_improvements);
+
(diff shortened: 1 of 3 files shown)


sanity:
## Review (Light tier — additive telemetry, no behavior change)

External model (Codex `codex review`) + build verification.

**Verdict: clean.** Side effects are limited to two short `connections_by_location` read-lock scans + relaxed atomic loads per snapshot cadence in `emit_router_snapshot_telemetry` (no panic path); OTLP `Option` mirror emits value-or-null correctly; the pin test asserts all 6 fields reach the body.

**One P3 (accepted):** `RouterSnapshotInfo` uses positional bincode, so inserting the 6 fields mid-struct makes pre-upgrade **local AOF** telemetry records unreadable (warn-skipped) after upgrade. Verified NOT P2P wire — `RouterSnapshot` is a local telemetry `EventKind` (register.rs), the AOF reader warn-skips incompatible records (aof.rs:531), and the reader is same-version/in-process. So no cross-peer/wire hazard and no crash; the only effect is a one-time loss of pre-upgrade local AOF snapshot history, and the central OTLP query path (the thing we actually use for network-wide lattice metrics) is unaffected. Accepted as the inherent cost of adding bincode telemetry fields.

[AI-assisted - Claude]

PR #4770, opened 2026-07-12
fix(ring): repair or drop phantom in-use contracts (#4612)

### LINKED ISSUE #4612: Root cause of #4610: don't mark a contract hosted/in-use without its state (phantom-hosting)
## Problem

#4610 (the unbounded `summarize_contract_state` storm) is an acute *symptom* of a deeper root cause: **a node marks a contract `contract_in_use` without ever holding its state.**

The dominant source is the subscribe **transit-relay** path: on a relayed `Subscribed` reply, the relay calls `register_downstream_subscriber(...)` (`operations/subscribe/op_ctx_task.rs:1911-1922` → `operations/subscribe.rs:615-637` → `ring.add_downstream_subscriber`) even though "a relay is not itself a subscriber." That flips `contract_in_use = has_downstream_subscribers = true` (`ring/hosting.rs:937`) on a stateless relay. (The SubscribeHint placement migration amplified the *volume*; that migration is being disabled separately, but this base leak is migration-independent.)

Consequences of phantom-hosted (`in_use=true`, no state):
- The #4610 summarize storm — mitigated by the PR #4611 state-presence gate, which **stops the storm but does NOT fix the phantom-hosting itself**.
- The node is advertised as interested/in-use for a contract it can't serve → contributes to GET dead-ends (#4404 retrievability).

## Proposed fix (root)

Make **`hosted/in-use ⇒ has-state`** an invariant. Either:
- (a) don't register downstream-subscriber / in-use status for a contract the relay holds no state for, OR
- (b) reconcile the ring hosting/in-use set against the state store — a contract isn't counted hosted/in-use until its state is actually fetched/stored; treat "in-use but no state" as a one-shot fetch/repair rather than a persistent phantom.

## SubscribeHint migration retro — learning to preserve

The placement migration is being disabled (net-negative). Lessons for any future placement design:
1. **`hosted ⇒ has-state` must be invariant** — never let an interest/subscriber registration imply summarizable/broadcastable/serveable without an actual state-presence check.
2. **Placement must RELOCATE state, not merely add a subscriber.** Under #4534 cache pressure the receiver refused ~99.95% of hints (act-rate 1.0%→0.05%, dominantly the module-cache admission gate + already-hosting), so hosting never drifted toward the key (`hosted_key_distance` flat ~0.21). Relocation needs headroom / evict-and-replace, not additive subscription.
3. **Solve the admission/act-rate problem BEFORE turning on a producer** — a feature whose admission gate drops 99.95% of its own proposals is near-pure cost.
4. **Gate re-enable on a metric that actually moves** (`hosted_key_distance` tightening in a canary), not just a passing controlled-sim test.
5. **Bound the producer** — both the #4440 renewal storm and the #4610 summarize storm came from an unbounded producer feeding shared single-threaded paths.

Refs #4610, #4404, #4440.

[AI-assisted - Claude]

netsirius:
## Problem

A node can be marked `contract_in_use` for a contract it has no state for. The dominant source is the SUBSCRIBE transit-relay path: on a relayed `Subscribed` reply the relay registers the downstream subscriber, which flips `contract_in_use = has_downstream_subscribers = true` on a node that never fetched the contract's state.

These "phantom hosts" caused:

- The #4610 summarize storm. The #4611 `contract_state_present` gate stops the storm but only filters the symptom — the phantom itself persists.
- GET dead-ends (#4404): the node is advertised as interested/in-use for a contract it cannot serve.

## Solution

Implements option (b) from #4612: reconcile the downstream-driven in-use set against the state store, making `in-use ⇒ has-state` a repaired invariant instead of a persistent violation.

`HostingManager::reconcile_phantom_in_use` runs every 30s from the existing recovery loop (`Ring::recover_orphaned_subscriptions`), after the lease-expiry sweep and the zero-connections gate. For each contract with downstream subscribers but no stored state it emits one of two actions:

- **`Fetch`** — a fire-and-forget sub-op GET (`start_sub_op_get` with contract code) so the state is actually fetched and stored; the node becomes a real host. Bounded: at most `MAX_PHANTOM_REPAIRS_PER_INTERVAL` (4) fetches per tick, one attempt per contract per `OPERATION_TTL` (60s) cooldown so a retry never overlaps an in-flight fetch, and `MAX_PHANTOM_REPAIR_ATTEMPTS` (3) attempts total.
- **`Drop`** — attempts exhausted and state still absent: the downstream registration is removed (`drop_phantom_downstream`), the interest manager is decremented symmetrically, and upstream is collapsed via `reconcile_wants_collapse` + unsubscribe — mirroring the existing stale-lease expiry handling. Downstream peers re-route through their own lease renewal (2 min interval), the same recovery contract lease expiry already relies on.

Design notes:

- **Repair-or-drop, not never-register.** Option (a) (skip `register_downstream_subscriber` on a stateless relay) was rejected: the downstream map is load-bearing for UPDATE forwarding through relay chains. Documented in `.claude/rules/ring.md`.
- **Level-triggered, not edge-triggered.** Decisions are made against the state store each sweep, not at subscribe arrival. Phantoms that self-heal (UPDATE auto-fetch, an in-flight subscription — skipped via `pending_subscription_requests` — a PUT) never consume repair budget, and the producer stays bounded regardless of subscribe traffic (the #4440/#4610 lesson).
- **Conservative by construction.** `contract_state_present` returns `true` on the sqlite backend, before `set_storage`, and on store errors — nothing is fetched or dropped unless a real redb store answered "absent". On sqlite the sweep is a safe no-op (same limitation as the #4611 gate).
- The bounded attempts satisfy the time-bounded cleanup-exemption rule: a phantom is resolved (repaired or dropped) in ~2.5–3.5 min, well within one subscription lease (8 min).

## Testing

Two regression tests in `ring/hosting.rs` (redb + injected mock clock, no network):

- `phantom_in_use_reconcile_fetch_then_drop_4612` — covers all four transitions: phantom → `Fetch` exactly once per cooldown window; attempts exhausted → `Drop`, and applying it clears `contract_in_use`; repair success (state appears between passes) → tracking cleared, downstream subscribers kept; a stateful in-use contract is never flagged.
- `phantom_reconcile_bounds_fetches_per_sweep_4612` — `max_fetches` caps fetches per sweep; leftover phantoms are picked up by later sweeps without burning attempts.

## Fixes

Closes #4612. Root fix for the phantoms behind #4610 (complements the #4611 gate, which remains as the symptom filter). Refs #4404.

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -237,6 +237,18 @@ WHEN summarizing contracts in the InterestSync heartbeat handlers
     state store so an evicted-but-on-disk contract (state present, not in the
     hosting cache, still in_use) keeps summarizing — keying on is_hosting_contract
     there would wrongly drop it.
+  → #4612 (root fix for the phantoms the #4610 gate only filters): the 30s
+    recovery loop reconciles the downstream-driven in-use set against the
+    state store (HostingManager::reconcile_phantom_in_use). A contract in_use
+    via downstream subscribers with NO stored state gets a bounded one-shot
+    repair fetch (sub-op GET; ≤ MAX_PHANTOM_REPAIRS_PER_INTERVAL per tick,
+    one per contract per OPERATION_TTL cooldown); after
+    MAX_PHANTOM_REPAIR_ATTEMPTS failures its downstream registration is
+    DROPPED (drop_phantom_downstream + symmetric interest decrement +
+    reconcile_wants_collapse), never kept as a persistent stateless "host".
+    Do NOT instead skip register_downstream_subscriber on a stateless relay:
+    the downstream map is load-bearing for UPDATE forwarding through relay
+    chains — the fix is repair-or-drop, not never-register.
 ```
 
 ### Cleanup Exemptions Must Be Time-Bounded
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -2014,6 +2014,16 @@ impl Ring {
     /// if the network can't absorb this many.
     const MAX_RECOVERY_ATTEMPTS_PER_INTERVAL: usize = 10;
 
+    /// Maximum phantom repair fetches (#4612) spawned per recovery tick.
+    ///
+    /// Bounds the producer (the #4440/#4610 lesson: never an unbounded
+    /// producer feeding the shared serial paths): at 30s intervals this is at
+    /// most 8 sub-op GETs/minute network-wide fetch pressure from phantom
+    /// repair, regardless of how many phantoms the relay path accumulated.
+    /// Per-contract pacing/giving-up lives in
+    /// `HostingManager::reconcile_phantom_in_use` (cooldown + max attempts).
+    const MAX_PHANTOM_REPAIRS_PER_INTERVAL: usize = 4;
+
     /// Skip renewal cycle when channel remaining capacity falls below this
     /// fraction of max (i.e. channel is more than 50% full).
     const RENEWAL_DEFER_CAPACITY_FRACTION: usize = 2; // channel_max / 2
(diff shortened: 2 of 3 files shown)


iduartgomez (approved):

sanity:
@netsirius @iduartgomez Hmm, this seems very closely related to some of the changes I've been working on the last week or two and I have a feeling I may already have fixed it. I've got an agent reviewing now.

Issue #3509, opened 2026-03-10
feat: show connections and status in tab title of dashboard


kerbo7:
The most important info of the local dashboard is:
**- am I connected?**
**- and how many connections?**

### Feature 

show this in the tab title of the dashboard

depending on the state of the peer inject uft-8 icons and number of connections into the page `<title> `

examples

```
<title>⚡ Dashboard</title>   trying to connect, please wait
<title>⚠ Dashboard</title>   unable to connect / major connect error
<title>☹ Dashboard</title>  peer is unhealthy  or 😣
<title>🔔 Dashboard</title>  sth. happenend, dash needs your attention
<title>☒ Dashboard</title>  no internet connection
<title>(4) Dashboard</title>  connected to 4 peers

(show only for 1 sec, then go back to previous title)
<title>📡 Dashboard</title>  new peer connection was successful
<title>⛔ Dashboard</title>  new peer connection was unsuccessful

```
more icons
https://www.w3schools.com/charsets/ref_emoji_intro.asp

### Source

https://github.com/freenet/freenet-core/blob/a085374f2629c8dc654a8b21a9caf7fbeb19de16/crates/core/src/server/home_page.rs#L46

sanity:
# Issue Priority Assessment

Generated: 2026-03-12T22:22:38.738452852+00:00
Issue hash: `31f8c1d6`

## Value Assessment

Relevance check: the dashboard HTML still hard‑codes `<title>Freenet — Local Peer</title>` in `crates/core/src/server/home_page.rs:46`, and the inlined `JS` block only handles theme toggling—there’s no script that observes `network_status` or manipulates `document.title`. The peer page likewise never changes the tab label. That means the behavior described in issue #3509 (icon/connection count injection in the `<title>`) is still absent, so the issue remains valid today.

User impact: every user who watches their local peer dashboard in a browser tab can already see connection health inside the page, but the tab title does not communicate “connected/not connected” at a glance while the tab is backgrounded. That is a minor quality‑of‑life gain—helpful for operators who frequently check the tab but not critical for most workflows. The current experience isn’t broken, so the severity is low, yet the feature would improve awareness for a non‑trivial slice of node operators.

Strategic alignment: enhancing the dashboard’s ambient status cues is consistent with the project’s focus on making node health understandable (see the existing telemetry/status cards in the same file). It isn’t foundational, but it supports the ongoing goal of better observability without touching core networking logic.

[...]

PR #4777, opened 2026-07-12
feat(dashboard): slow auto-refresh when tab is hidden, refresh on focus

### LINKED ISSUE #3353: req: improve live reloading of node dashboard
### Problem
The dashboard reloads the complete page every 5 seconds. Not a problem for the system load. But it makes you nervous if the tab is inactive  or active. Constant reload flashing on the tab handle. If you want check and troubleshoot anything in the source html on the page, you almost can't do it, because the page is gone in 5 secs.

### Solution

some ideas:

---

Use event visibilitychange:

```
document.addEventListener("visibilitychange", () => {
  if (document.hidden) {
     ### Longer reload interval when tab is inactive: 60 seconds
  } else {
     ### Reload page when tab gets active. So it's current when the user is actually watching it.
     ### Shorter interval when tab is active, but not 5sec: 30 seconds?
  }
```

---

just reload an iframe not the whole page. Don't forget to message the <title> and colored connection status favicon back to the root html. But still give it more than 5 secs. (not really a good solution)

---

Use a websocket and dont reload page at all (best)

iduartgomez:
## Problem

Issue #3353: the dashboard polls the local peer on a fixed 5-second
cadence regardless of whether the tab is actually visible. A
backgrounded/hidden tab gains nothing from that cadence — nobody is
looking at it — but it keeps hitting the local node every 5s anyway,
burning CPU/battery and (per the reporter) making the tab feel
"nervous"/noisy in the background.

Note: the specific complaint in the issue about a full-page reload
flashing the tab (`location.reload()` / meta-refresh) was already fixed
by prior work — `home_page.rs`'s `no_meta_refresh_in_peer_detail` test
and the existing `scheduleRefresh` implementation already do a
`fetch()` + `DOMParser` + `<main>`-innerHTML swap, not a full navigation.
That part of the issue is not reopened by this PR; what remained open
was the cadence itself.

## Solution

Add a `document.visibilitychange` listener to the dashboard's
auto-refresh loop (`assets/dashboard.js`):

- **Hidden tab**: back off from the normal 5s interval to a 60s
  interval (`HIDDEN_REFRESH_MS`), so a backgrounded tab polls far less
  often.
- **Tab becomes visible again**: refresh immediately instead of
  waiting out whatever remains of the 60s hidden-tab timer, so the
  user sees current data as soon as they look at it, then falls back
  to the normal 5s cadence.
- Refresh logic itself is unchanged (still `fetch` + `DOMParser` +
  `<main>` swap, still `setTimeout` chaining rather than
  `setInterval`, so a slow response can't stack overlapping requests).

This directly implements the first suggestion from the issue
(`document.visibilitychange` with a longer interval when hidden, and
an immediate refresh on becoming visible). The issue's other two
suggestions (iframe-based partial reload, WebSocket push) are bigger
architectural changes and out of scope for this one-logical-change PR.

Coordinates with, but is independent of, #3509 (tab-title/favicon
status indicator) — that issue is about a different piece of the
dashboard (the `<title>`/favicon) and is not touched here.

## Testing

- `cargo test -p freenet --lib home_page` — 109 passed (106 pre-existing
  + 3 new), 1 ignored.
- New render/behavior-pinning tests in `home_page.rs`:
  - `js_slows_refresh_when_tab_hidden` — asserts the JS branches on
    `document.hidden` and defines a distinct `HIDDEN_REFRESH_MS` (60000).
  - `js_refreshes_immediately_on_tab_visible` — asserts a
    `visibilitychange` listener exists and a `refreshDashboard()` call
    is reachable outside the timer chain.
  - `js_auto_refresh_has_no_setinterval` — guards against regressing to
    `setInterval`, which would let overlapping fetches stack up.
- `cargo fmt -p freenet -- --check` — clean.
- `cargo clippy -p freenet -- -D warnings` — clean.
- `cargo test -p freenet --lib` — 3887 passed, 2 failed (both
  pre-existing macOS-only `wasm_runtime::migration_tests` failures,
  unrelated to this change and known-ignorable per repo convention),
  25 ignored.
- `npm run lint` (server assets: prettier + eslint) — clean.

## Fixes

Closes #3353

https://claude.ai/code/session_01DR69yFsB57UFRTMrHGG5nx

### DIFF
--- a/.github/workflows/lint-assets.yml
+++ b/.github/workflows/lint-assets.yml
@@ -12,6 +12,7 @@ on:
     paths:
       - 'crates/core/src/server/**/assets/**'
       - 'crates/core/src/server/base58_encode.test.mjs'
+      - 'crates/core/src/server/dashboard_refresh.test.mjs'
       - 'crates/core/src/server/package.json'
       - 'crates/core/src/server/package-lock.json'
       - 'crates/core/src/server/.prettierrc.json'
@@ -22,6 +23,7 @@ on:
     paths:
       - 'crates/core/src/server/**/assets/**'
       - 'crates/core/src/server/base58_encode.test.mjs'
+      - 'crates/core/src/server/dashboard_refresh.test.mjs'
       - 'crates/core/src/server/package.json'
       - 'crates/core/src/server/package-lock.json'
       - 'crates/core/src/server/.prettierrc.json'
@@ -55,6 +57,6 @@ jobs:
         working-directory: crates/core/src/server
         run: npm run lint
 
-      - name: Test (base58 access-key encoder)
+      - name: Test (Node asset unit tests)
         working-directory: crates/core/src/server
         run: npm test
--- a/crates/core/src/server/dashboard_refresh.test.mjs
+++ b/crates/core/src/server/dashboard_refresh.test.mjs
@@ -1,0 +1,327 @@
+// Executable behavioral test for the dashboard auto-refresh scheduler in
+// `home_page/assets/dashboard.js`.
+//
+// Why this exists: the scheduler is a small state machine (refreshTimer /
+// refreshInFlight / visibility-dependent cadence) whose original bug — a
+// visibilitychange racing an in-flight timer-triggered fetch started a second
+// concurrent refresh chain, because refreshTimer was never reset once its
+// setTimeout callback fired — is exactly the kind of wiring error a substring
+// pin test cannot catch. Rather than pull in a browser test runner, this test
+// EXTRACTS the self-contained `createRefreshScheduler` factory verbatim from
+// the asset (between the `refresh-scheduler:BEGIN`/`:END` markers; the factory
+// takes all browser deps — timers, refresh fn, visibility — as injected
+// params) and drives it under Node with fake timers and a controllable fake
+// refresh, asserting the behavioral invariants:
+//
+//   1. one fetch at a time: a visibility flip during an in-flight refresh
+//      never starts a second concurrent refresh (the pre-fix bug);
+//   2. cadence: hidden tabs are polled every 60s, visible tabs every 5s;
+//   3. becoming visible triggers an immediate refresh only when idle,
+//      otherwise it just reschedules;
+//   4. the poll chain never forks: after arbitrary interleavings of timer
+//      fires and visibility toggles, exactly one timer is pending at every
+//      quiescent point.
+//
+// Run via `npm test` in crates/core/src/server (wired into the lint-assets CI
+// job). No dependencies beyond Node's stdlib. Exits non-zero on any mismatch.
+
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+
+const here = dirname(fileURLToPath(import.meta.url));
+const assetPath = join(here, 'home_page/assets/dashboard.js');
+const src = readFileSync(assetPath, 'utf8');
+
+// --- Extract the factory verbatim from the asset --------------------------
+const BEGIN = 'refresh-scheduler:BEGIN';
+const END = 'refresh-scheduler:END';
+const b = src.indexOf(BEGIN);
+const e = src.indexOf(END);
+if (b < 0 || e < 0 || e < b) {
+  console.error(
+    `FAIL: could not find ${BEGIN}/${END} markers in ${assetPath}. ` +
+      'The refresh scheduler must stay bracketed by those markers so this ' +
+      'test can extract and verify it.',
+  );
+  process.exit(1);
+}
+const region = src.slice(b, e);
+const fnStart = region.indexOf('function createRefreshScheduler(');
+if (fnStart < 0) {
+  console.error(
+    'FAIL: no `function createRefreshScheduler(` between the markers.',
+  );
+  process.exit(1);
+}
+const fnSource = region.slice(fnStart, region.lastIndexOf('}') + 1);
+const createRefreshScheduler = new Function(
+  `${fnSource}\nreturn createRefreshScheduler;`,
+)();
+
+// --- Test harness ----------------------------------------------------------
+let failures = 0;
+function check(cond, msg) {
+  if (!cond) {
+    failures++;
+    console.error(`FAIL: ${msg}`);
+  }
+}
+
+// Let promise chains (.then/.finally on already-settled promises) run to
+// completion. setImmediate runs after the microtask queue drains, so one
+// round is enough for arbitrarily long already-settled chains; do a few to
+// be safe against future chain reshaping.
+async function drain() {
+  for (let i = 0; i < 4; i++) {
+    await new Promise((resolve) => setImmediate(resolve));
+  }
+}
+
+// Deterministic manual timers. advance(ms) fires due callbacks in order,
+// draining promise chains after each so a callback's .finally reschedules
+// before the next timer is considered (matching real event-loop behavior).
+function makeFakeTimers() {
+  let now = 0;
+  let nextId = 1;
+  const timers = new Map(); // id -> { fn, at }
+  const delays = []; // every delay ever passed to setTimeout
+  return {
+    setTimeout(fn, ms) {
+      const id = nextId++;
+      timers.set(id, { fn, at: now + ms });
+      delays.push(ms);
+      return id;
+    },
+    clearTimeout(id) {
+      timers.delete(id);
+    },
+    async advance(ms) {
+      const target = now + ms;
+      for (;;) {
+        let dueId = null;
+        let dueAt = Infinity;
+        for (const [id, t] of timers) {
+          if (t.at <= target && t.at < dueAt) {
+            dueAt = t.at;
+            dueId = id;
+          }
+        }
+        if (dueId === null) break;
+        const { fn } = timers.get(dueId);
+        timers.delete(dueId);
+        now = dueAt;
+        fn();
+        await drain();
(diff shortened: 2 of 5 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4777

### Summary
- **PR Title:** feat(dashboard): slow auto-refresh when tab is hidden, refresh on focus
- **Type:** feat (dashboard front-end)
- **CI Status:** pending at review time (Conventional Commits / Prettier+ESLint / Rule Lint already green; Clippy / Fmt / E2E still running)
- **Linked Issues:** Closes #3353
- **Review tier:** Light — self-contained front-end change (dashboard JS + Rust render-pinning tests); no high-risk surface (no wire format, crypto, state authorization, transport, or async runtime code).
- **Reviewers run:** skeptical bug-hunt, code-first, and change-specific (refresh-loop lifecycle/concurrency) lenses, run serially in-context. External model pass (`codex`) was **unavailable** — see note below. Freenet bug-pattern file checked (scoped to `crates/core/src/bin/**`; not applicable here).

> **External-model substitution:** `codex review --base origin/main` failed with a broken vendored binary (`spawn … codex ENOENT`), retried once, same failure; `gemini` CLI is not installed. Both external models are genuinely down. Per the review skill I substituted a diverse-Claude-lens fallback (skeptical / code-first / change-specific concurrency lenses). This is a low-risk, non-time-sensitive change; if an external pass is wanted before merge, it can be re-run once the `codex` install is repaired.

---

### Code-First Analysis
**Independent Understanding:** The inline auto-refresh closure is extracted verbatim into a named `refreshDashboard()` that returns its fetch promise. `scheduleRefresh()` now (a) clears any pending timer before arming a new one, and (b) picks its delay from `currentRefreshInterval()` = 5s visible / 60s hidden. A `visibilitychange` listener refreshes immediately on becoming visible, then re-chains the normal cadence.
**Stated Intent:** Back off to 60s when the tab is hidden; refresh immediately when it becomes visible; keep `setTimeout` chaining (no `setInterval`) so slow responses don't overlap.
**Alignment:** Matches. The refresh body is byte-for-byte identical to the prior code (indentation only) — confirmed at the diff level. `.finally(scheduleRefresh)` moved from inside the chain to the call site; because `.catch` resolves the promise, the chain still always re-arms.
**Gaps:** The issue also floated a 30s cadence *while visible*; the PR keeps 5s visible. That is a reasonable, deliberate deviation (a watched tab benefits from the fast cadence) and not a defect.

---

### Testing Assessment
**Coverage Level:** Adequate for a front-end string-pinning change.

| Unit | ✅ | 3 new `home_page::tests` pins (`js_slows_refresh_when_tab_hidden`, `js_refreshes_immediately_on_tab_visible`, `js_auto_refresh_has_no_setinterval`). Ran locally: 8 JS-pin tests pass, 0 fail. |
| Integration | N/A | No server-side behavior changed. |
| Simulation | N/A | Not DST-relevant. |
| E2E | ⚠️ | Playwright / bundle-swap E2E jobs cover dashboard render but not tab-visibility cadence (JS timers/visibility are hard to assert headlessly); acceptable gap. |

**Regression Test:** Present and valid — the pins guard the `document.hidden` branch, the `HIDDEN_REFRESH_MS`/`60000` backoff, the `visibilitychange` listener, and the no-`setInterval` invariant.
**Missing Tests:** The pins assert *presence of strings*, not runtime behavior — they cannot catch the overlap edge case below. That is inherent to the string-pinning approach used throughout this file and not a new deficiency.

---

### Skeptical Findings
**Risk Level:** low

| Visibility-triggered refresh can start a **second concurrent** fetch, doubling the timer chain | low (Consider) | `dashboard.js:569-583` (verified) | The `setTimeout` callback in `scheduleRefresh` calls `refreshDashboard()` **without nulling `refreshTimer`**, so during that in-flight fetch `refreshTimer` still holds the already-fired (stale) id. If `visibilitychange`→visible fires in that window, its `clearTimeout(refreshTimer)` is a no-op and it launches a *second* `refreshDashboard().finally(scheduleRefresh)`. Two `.finally(scheduleRefresh)` chains then run in parallel, so the effective cadence can permanently double after one ill-timed visibility flip. This weakens the PR's stated "slow responses don't overlap / can't stack overlapping requests" guarantee (that guarantee held for the pure timer path; the new visibility path reintroduces a narrow overlap). Impact is low: the DOM swap is idempotent (last write wins) and the trigger window is small. Clean fix: set `refreshTimer = null` as the first statement inside the timer callback (and ideally an `inFlight` guard so the visibility handler no-ops while a refresh is already running). |
| Immediate refresh on every visible transition is unthrottled | low (Consider) | `dashboard.js:579-583` (verified) | Rapid alt-tabbing fires a fresh `refreshDashboard()` per visible transition, stacking fetches to the local node. Real-world impact negligible; a short debounce would harden it. |

---

### Big Picture Assessment
**Goal Alignment:** Yes — implements the issue's first (recommended) suggestion faithfully; the iframe and WebSocket alternatives are correctly deferred as out of scope.
**Anti-Patterns Detected:** None from `.claude/rules/bug-prevention-patterns.md` (that file is path-scoped to `crates/core/src/bin/**` and concerns Windows `Command::spawn` stdio — unrelated). No `biased;` select, fire-and-forget spawn, bounded-channel-in-loop, or wire-format concerns apply.
**Removed Code Concerns:** None — the removed inline closure is preserved verbatim inside `refreshDashboard()`; no behavior dropped. The prior full-reload complaint in the issue was already fixed by earlier work (correctly noted in the PR body).
**Scope Assessment:** Focused — one logical change. The `refreshTimer` single-slot guard is a genuine small improvement that rides along legitimately (it is required by the new immediate-refresh path).

---

### Documentation
- Code docs: Complete — the JS carries a clear block comment explaining the visibility cadence rationale and the setTimeout-vs-setInterval invariant; new tests carry `#3353` rationale comments.
- Architecture docs: N/A.
- User docs: N/A — no user-facing config or CLI surface changed.

---

### Recommendations

#### Must Fix (Blocking)
- None.

#### Should Fix (Important)
- None.

#### Consider (Suggestions)
1. Null `refreshTimer` at the top of the `scheduleRefresh` timer callback (and/or add an `inFlight` flag the `visibilitychange` handler checks) so a visibility flip during an in-flight fetch cannot spawn a second concurrent refresh chain and permanently double the cadence. Keeps the PR's own "no overlapping requests" invariant intact.
2. Optionally debounce the immediate on-visible refresh to avoid fetch stacking under rapid tab switching.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** 1211dfeed3332534b1e1819632576f8fc8c2a5ff

No blocking or important issues. The change is faithful to #3353, the refresh body is preserved verbatim, and the new regression pins are appropriate and pass locally. The two "Consider" items are low-impact edge cases in the new visibility path (a narrow concurrent-refresh window and unthrottled on-visible refresh); the maintainer may address them now or accept them given the idempotent DOM swap. External-model pass was unavailable (broken `codex` install, no `gemini`); a Claude-lens fallback was substituted — re-run the external pass if desired before merge, though risk here is low. Recommend waiting for the pending Clippy/Fmt/E2E CI jobs to go green.

[AI-assisted - Claude]

iduartgomez:
Fixed in b2e4fc9ffe29608bae119e59c002e3eb69953a75 (pushed to this branch).

The concurrent-refresh race is addressed: `refreshTimer` is now reset to `null` at the top of the timer callback before `refreshDashboard()` runs, so a later `clearTimeout(refreshTimer)` can never silently no-op against an already-fired timer id. Added a `refreshInFlight` guard so `refreshDashboard()` is a no-op whenever a fetch is already running (set `true` before the fetch, `false` in `.finally`); the `visibilitychange`→visible path now only triggers an immediate refresh when nothing is in flight, otherwise it just reschedules at the normal cadence. `setTimeout` chaining is unchanged (no `setInterval`).

Added `js_guards_against_concurrent_refresh_chains` in `home_page.rs` to pin both the `refreshInFlight` flag and the `refreshTimer = null` reset, following the file's existing `JS.contains(...)` pin-test pattern.

Gates: `cargo fmt --check`, `cargo clippy -p freenet -- -D warnings`, `cargo test -p freenet --lib server::home_page` (110 passed, 1 ignored), and `npm run lint` (prettier + eslint) all clean.

iduartgomez:
Behavioral test added per rule-review in 8e49ad14a, following the `base58_encode.test.mjs` precedent.

The scheduling core is now extracted into a self-contained `createRefreshScheduler({setTimeout, clearTimeout, refresh, isHidden})` factory (bracketed by `refresh-scheduler:BEGIN/END` markers; page wiring injects the real browser deps — behavior unchanged). New `dashboard_refresh.test.mjs` extracts the factory verbatim and drives it under Node with deterministic fake timers and a controllable fake refresh, asserting: (a) a visibility flip during an in-flight timer-triggered refresh never starts a second concurrent chain — verified this fails 12 checks against the pre-fix wiring; (b) hidden 60s / visible 5s cadence; (c) visible-flip refreshes immediately only when idle, otherwise reschedules; (d) exactly one pending timer after arbitrary timer-fire/visibility-toggle interleavings.

Wired into CI via the existing `npm test` script in `crates/core/src/server` (lint-assets job), with the new test file added to the workflow's trigger paths. The substring pin tests stay and now also pin the extraction markers so a Rust-side refactor can't strip them.

Gates: `npm run lint` + `npm test` clean, `cargo fmt --check`, `cargo clippy -p freenet -- -D warnings`, `cargo test -p freenet --lib server::home_page` (110 passed, 1 ignored) all green.

PR #2215, opened 2025-12-05
fix: wait for stretto cache insert before validation

sanity:
## Problem

When a contract is stored via PUT, the subsequent GET from the same node can fail with "contract not found" even though both operations happen on the same peer. This is a race condition between storing and fetching the contract.

**Root cause:** The `stretto` caching library (used in `ContractStore`) provides eventual consistency. Inserts are buffered to background threads and may not be immediately visible to subsequent `get()` calls. From the [stretto documentation](https://docs.rs/stretto/latest/stretto/struct.Cache.html#method.wait):

> "Wait until all previous operations have been applied. This ensures that any `insert` or `remove` calls made before calling this method will be reflected in subsequent `get` calls."

**The race condition:**
1. `store_contract()` calls `cache.insert()` for a new contract
2. `validate_state()` immediately calls `fetch_contract()` to retrieve it
3. `fetch_contract()` calls `cache.get()` but the insert hasn't been applied yet
4. Returns "contract not found"

## Solution

Call `wait()` after inserting into the cache to ensure the value is visible before returning from `store_contract()`. This adds minimal overhead since we're only blocking until this specific insert completes, not waiting for unrelated cache operations.

## Testing

- `cargo test -p freenet store_and_load` passes
- The existing unit test in `contract_store.rs` validates store-then-fetch works correctly

## Fixes

Closes #2214

[AI-assisted - Claude]

PR #4779, opened 2026-07-12
refactor: validate contracts from the in-hand container, skip store→fetch (#2216)

### LINKED ISSUE #2216: Tech debt: Avoid store→fetch round-trip in contract validation
## Overview

During the fix for #2214, we identified an architectural inefficiency where `verify_and_store_contract()` stores a contract then immediately fetches it back for validation.

## Current Flow

```rust
// verify_and_store_contract() in runtime.rs

// Line 1076: We HAVE the contract
if let Some(contract) = trying_contract.take() {
    // Line 1092: Store it (moves ownership)
    self.runtime.contract_store.store_contract(contract)?;
}

// Line 1105: Immediately fetch it back for validation
let result = self.runtime.validate_state(&trying_key, ...)?;
```

The `validate_state()` call needs the contract bytes to compile a WASM module via `prepare_contract_call()` → `fetch_contract()`.

## The Issue

We have the contract data in hand, give it to the store, then immediately ask the store to give it back. This required adding `wait()` after cache insert (#2215) because stretto uses eventual consistency and the data wasn't visible yet.

## Potential Solutions

1. **Pass contract to validate_state directly:**
   ```rust
   self.runtime.validate_state_with_contract(&trying_key, &contract, ...)?;
   self.runtime.contract_store.store_contract(contract)?;
   ```

2. **Have store_contract also compile and cache the WASM module**

3. **Clone contract code before storing and use directly for compilation**

## Impact

- **Performance:** Minimal - `wait()` is fast (just flushes buffer)
- **Correctness:** Current fix works correctly
- **Code clarity:** The round-trip is architecturally awkward

## Priority

Low - the current fix is correct and not performance-impacting. This is cleanup for code clarity.

## Related

- Fixed in #2215
- Root cause issue #2214

[AI-assisted - Claude]

iduartgomez:
## Problem

`verify_and_store_contract()` (`crates/core/src/contract/executor/runtime/contract_ops.rs`) receives a freshly-put `ContractContainer`, stores it into `contract_store`, and then validates the state. Validation compiles the WASM module via `Runtime::validate_state` → `prepare_contract_call`, which — on a module-cache miss (guaranteed the first time a brand-new contract is PUT) — re-fetches the exact same contract bytes from `contract_store` just to compile them. We already had the bytes in hand a few lines above; this is a redundant store→fetch round-trip, as described in #2216.

The issue's own line references were stale (the executor was refactored into `crates/core/src/contract/executor/runtime/` modules, and #4702's disk-budget admission gate touched `verify_and_store_contract` further), so this PR re-locates the current code before changing it. It also investigates and rules out the issue's cited *root cause*: sibling issue #2215's `stretto` cache `wait()` (needed because stretto's cache insert was eventually-consistent) is gone — PR #3471 replaced `stretto` with `moka` (synchronous insert/get), so that specific eventual-consistency hazard no longer exists. The architectural round-trip itself, however, was still present, which is what this PR removes.

## Solution

- Added `Runtime::prepare_contract_call_with_contract` (and a shared `prepare_contract_call_inner`) that, on a module-cache miss, compiles directly from a caller-supplied `ContractContainer` instead of calling `contract_store.fetch_contract`. `prepare_contract_call` keeps its original behavior (`None` case) for every other caller.
- Added `Runtime::validate_state_with_contract`, mirroring `ContractRuntimeInterface::validate_state` but routed through the new `prepare_contract_call_with_contract`. The two share a `run_validate_state` body so there's no duplicated WASM-call logic.
- `verify_and_store_contract` now keeps a cheap `Arc`-backed `.clone()` of the contract it already received, and threads it through `fetch_related_for_validation_network` (which now takes `already_fetched_contract: Option<&ContractContainer>`) into `validate_state_with_contract` — both the initial validate call and the depth=1 retry-after-`RequestRelated` call.
- The other two call sites of `fetch_related_for_validation_network` (`perform_contract_put`'s existing-contract merge path, and `get_updated_state`'s post-update validation) don't have a fresh `ContractContainer` in hand — they pass `None` and are unchanged in behavior.
- Scope: only `verify_and_store_contract` — the function #2216 names — was touched. The generic, cross-runtime `bridged_upsert_contract_state` PUT path (`executor_impl.rs`, used by both `Executor<Runtime>` and `Executor<MockWasmRuntime>`) has the same shape of round-trip but is a separate, generic path; changing it would mean adding this method to the `ContractRuntimeInterface` trait (touching the mock runtime too) — out of scope for this one-function fix.
- All existing disk-budget admission-gate ordering and `remove_contract` / `record_wasm_removed` rollback semantics in `verify_and_store_contract` are untouched; only the validation call itself was rerouted.

## Testing

- New regression test `wasm_runtime::tests::contract::validate_state_with_contract_does_not_require_prior_store`: builds a real (compiled) test contract, deliberately does **not** store it in `contract_store`, and calls `validate_state_with_contract` directly — proving it validates from the supplied container without needing the store round-trip. Asserts both the `Valid` and `RequestRelated` (invalid-state) branches match `validate_state`'s existing behavior.
  - Verified the test fails to compile without the new API (temporarily reverted the runtime/contract.rs changes) and fails at runtime with `ContractNotFound` when the pass-through is defeated (temporarily forced `prepare_contract_call_with_contract` to ignore the supplied contract) — confirming the test actually exercises the fix.
- `cargo fmt` — clean.
- `cargo clippy -p freenet -- -D warnings` — clean.
- `cargo test -p freenet --lib` — 3862 passed, 25 ignored, only the 2 pre-existing macOS-only `wasm_runtime::migration_tests` failures (confirmed identical on the unmodified base branch — case-sensitivity artifact, unrelated to this change).

## Fixes

Closes #2216

https://claude.ai/code/session_01DR69yFsB57UFRTMrHGG5nx

### DIFF
--- a/crates/core/src/contract/executor/runtime/contract_ops.rs
+++ b/crates/core/src/contract/executor/runtime/contract_ops.rs
@@ -61,9 +61,18 @@ impl Executor<Runtime> {
                 }
             };
 
-            // Validate before persisting (fetch related contracts from network if needed)
+            // Validate before persisting (fetch related contracts from network if needed).
+            // This is an existing-contract merge, not a fresh store — we don't
+            // have a `ContractContainer` in hand, so fall back to the
+            // fetch-from-store path (`None`).
             let validate_result = self
-                .fetch_related_for_validation_network(&key, &params, &new_state, &related_contracts)
+                .fetch_related_for_validation_network(
+                    &key,
+                    &params,
+                    &new_state,
+                    &related_contracts,
+                    None,
+                )
                 .await?;
             if validate_result != ValidateResult::Valid {
                 return Err(Self::validation_error_put(key, validate_result));
@@ -279,12 +288,15 @@ impl Executor<Runtime> {
         };
 
         // Validate before persisting or broadcasting (fetch related contracts from network if needed).
+        // This is an UPDATE on an already-hosted contract, not a fresh store —
+        // no `ContractContainer` in hand, so fall back to fetch-from-store (`None`).
         let result = self
             .fetch_related_for_validation_network(
                 &key,
                 parameters,
                 &new_state,
                 &RelatedContracts::default(),
+                None,
             )
             .await?;
 
@@ -367,6 +379,11 @@ impl Executor<Runtime> {
             }
         }
 
+        // Keep a cheap (Arc-backed) clone in hand so validation below can use
+        // it directly instead of re-fetching from `contract_store` on a
+        // module-cache miss (issue #2216: avoids the store→fetch round-trip).
+        let contract_for_validation = contract.clone();
+
         // Store contract code in runtime store
         self.runtime
             .contract_store
@@ -408,7 +425,13 @@ impl Executor<Runtime> {
         //   - Excessive fan-out (max 10 related contracts per request)
         // See MAX_RELATED_CONTRACTS_PER_REQUEST and RELATED_FETCH_TIMEOUT constants.
         let result = self
-            .fetch_related_for_validation_network(&key, &params, &state, &related_contracts)
+            .fetch_related_for_validation_network(
+                &key,
+                &params,
+                &state,
+                &related_contracts,
+                Some(&contract_for_validation),
+            )
             .await
             .inspect_err(|_| {
                 if let Err(e) = self.runtime.contract_store.remove_contract(&key) {
@@ -601,17 +624,31 @@ impl Executor<Runtime> {
     /// `local_state_or_from_network` helper directly and several PUT call
     /// sites already wire through it; collapsing the two would touch more
     /// surface area than the bug fix needs.
+    ///
+    /// `already_fetched_contract`: when the caller already holds the
+    /// `ContractContainer` being validated (e.g. `verify_and_store_contract`,
+    /// which just handed it to `contract_store.store_contract`), pass it here
+    /// so a module-cache miss compiles directly from it instead of re-fetching
+    /// the same bytes from `contract_store` (issue #2216). `None` for callers
+    /// that don't have it in hand — behaves exactly as before.
     async fn fetch_related_for_validation_network(
         &mut self,
         key: &ContractKey,
         params: &Parameters<'_>,
         state: &WrappedState,
         initial_related: &RelatedContracts<'_>,
+        already_fetched_contract: Option<&ContractContainer>,
     ) -> Result<ValidateResult, ExecutorError> {
-        let result = self
-            .runtime
-            .validate_state(key, params, state, initial_related)
-            .map_err(|e| ExecutorError::execution(e, None))?;
+        let result = match already_fetched_contract {
+            Some(contract) => self
+                .runtime
+                .validate_state_with_contract(key, params, state, initial_related, contract)
+                .map_err(|e| ExecutorError::execution(e, None))?,
+            None => self
+                .runtime
+                .validate_state(key, params, state, initial_related)
+                .map_err(|e| ExecutorError::execution(e, None))?,
+        };
 
         let requested_ids = match result {
             ValidateResult::Valid | ValidateResult::Invalid => return Ok(result),
@@ -711,10 +748,16 @@ impl Executor<Runtime> {
         }
 
         let populated_related = RelatedContracts::from(related_map);
-        let retry_result = self
-            .runtime
-            .validate_state(key, params, state, &populated_related)
-            .map_err(|e| ExecutorError::execution(e, None))?;
+        let retry_result = match already_fetched_contract {
+            Some(contract) => self
+                .runtime
+                .validate_state_with_contract(key, params, state, &populated_related, contract)
+                .map_err(|e| ExecutorError::execution(e, None))?,
+            None => self
+                .runtime
+                .validate_state(key, params, state, &populated_related)
+                .map_err(|e| ExecutorError::execution(e, None))?,
+        };
 
         if let ValidateResult::RequestRelated(_) = &retry_result {
             return Err(ExecutorError::request(StdContractError::Put {
--- a/crates/core/src/wasm_runtime/contract.rs
+++ b/crates/core/src/wasm_runtime/contract.rs
@@ -1,5 +1,6 @@
 use super::engine::{WasmEngine, WasmError};
 use super::native_api::CONTRACT_IO;
+use super::runtime::RunningInstance;
 use super::{ContractExecError, RuntimeResult};
 use freenet_stdlib::prelude::{
     CodeHash, ContractContainer, ContractInstanceId, ContractInterfaceResult, ContractKey,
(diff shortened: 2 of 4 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #4779

### Summary
- **PR Title:** refactor: validate contracts from the in-hand container, skip store→fetch (#2216)
- **Type:** refactor (behavior-preserving performance/cleanup)
- **CI Status:** Clippy / Fmt / Conventional Commits / Rule checks passing; Unit & Integration / Simulation / NAT / Windows still pending at review time
- **Linked Issues:** Closes #2216 (approved tech-debt issue, `E-easy`, `P-low`)
- **Review tier:** Full (touches the contract WASM runtime — a high-risk surface)
- **Reviewers run:** Serial in-context adversarial passes (skeptical bug-hunt, code-first, change-specific correctness/compile-safety, testing) + big-picture. **External model pass could not run** — see note below. Regression test built and executed locally (passes).

> **External-model substitution note:** `codex review` failed on this machine with a broken install (`spawn … codex ENOENT`, missing vendored binary) on two attempts; `gemini` is not installed. Both external models were genuinely unavailable. Per the review policy I fell back to a diverse-Claude-lens pass run **serially within a single context** (subagent spawning was not available to this dispatched agent): a skeptical bug-hunt lens, a code-first lens, a change-specific "wrong-module / compile-safety" lens, and a testing lens, each a distinct pass, then synthesized. The independent external pass should ideally be re-run before merge if a working `codex`/`gemini` becomes available, though the change is low-risk.

---

### Code-First Analysis
**Independent Understanding:** The PR adds an alternate module-compile path so that `verify_and_store_contract` — which already holds the freshly-PUT `ContractContainer` after handing it to `contract_store.store_contract` — can validate the state by compiling directly from that in-hand container on a module-cache miss, instead of re-fetching the identical bytes back out of the store. Mechanically:
- `prepare_contract_call_inner(…, already_fetched: Option<&ContractContainer>)` is the new shared body; `prepare_contract_call` passes `None` (unchanged behavior), `prepare_contract_call_with_contract` passes `Some`.
- `run_validate_state(running, …)` is extracted from the old `validate_state` body; `validate_state` and the new `validate_state_with_contract` both build a `RunningInstance` then call it — no duplicated WASM-call logic.
- `fetch_related_for_validation_network` gains an `already_fetched_contract: Option<&ContractContainer>` param, threaded through both the initial validate and the depth=1 retry-after-`RequestRelated` call.

**Stated Intent:** Remove the architectural store→fetch round-trip named in #2216, scoped to only `verify_and_store_contract`.

**Alignment:** Matches. All three `fetch_related_for_validation_network` call sites are accounted for: the two paths that don't hold a fresh container (existing-contract merge in `perform_contract_put`; post-update validation in `get_updated_state`) pass `None` and are behaviorally identical; only the fresh-store path passes `Some(&contract_for_validation)`. The generic `bridged_upsert_contract_state` path is explicitly left out of scope, correctly (it would require adding to the `ContractRuntimeInterface` trait and touching the mock runtime).

**Gaps:** None between description and implementation. The PR description is unusually accurate, including the #2215/#3471 (stretto→moka) analysis, which I confirmed (no stale `cache.wait()` remains in the runtime).

---

### Testing Assessment
**Coverage Level:** Adequate.

| Unit | ✅ | New `validate_state_with_contract_does_not_require_prior_store`: deliberately does NOT store the contract, asserts `fetch_contract` genuinely misses, then validates directly from the supplied container. Covers both `Valid` and `RequestRelated` (invalid-state) branches. Built + ran locally — **passes**. |
| Integration | ✅ | The wired `Some(&…)` path through `verify_and_store_contract` is exercised end-to-end by existing PUT suites (`pool_tests/related_contract_tests`, `disk_budget_gate_tests`, `conformance_tests`), including the retry-after-`RequestRelated` branch. |
| Simulation | N/A | No topology/timing change. |
| E2E | N/A | Local runtime change only. |

**Regression Test:** Present and valid. Author reports (and the structure supports) that it fails without the API and fails with `ContractNotFound` when the pass-through is defeated — a genuine guard, not a happy-path rubber stamp.

**Missing Tests:** No direct test asserts the wiring inside `fetch_related_for_validation_network` selects the `Some` arm, but this is a trivial `None`→`Some` substitution covered by the existing integration suites; a dedicated test would be low-value.

---

### Skeptical Findings
**Risk Level:** Low.

| Caller/key trust is implicit and unguarded | Low (Consider) | `runtime.rs:664` `prepare_contract_call_with_contract` | The method compiles `already_fetched`'s code and caches the module under `key`, trusting that `contract.key() == *key`. The sole caller (`verify_and_store_contract`) sets `key = contract.key()`, so this holds today. But a future caller passing a mismatched `(key, contract)` would silently compile the wrong code and poison `contract_modules` under `key`. A `debug_assert_eq!(&contract.key(), key)` would pin the invariant cheaply. |
| Module cache retains entry after rejected PUT | Not a finding (pre-existing) | `runtime.rs:732` / `contract_ops.rs:437` | On invalid state, `remove_contract` deletes the store blob but not the compiled module in `contract_modules`. This is **identical to the pre-existing `fetch`-then-compile path** — the old code also compiled and cached before validation returned `Invalid`. Not introduced by this PR; noted only to confirm it was checked. |

I verified the central safety property: `WrappedContract::new` derives `key = ContractKey::from_params_and_code(&params, &code)` (stdlib 0.6.1), so a container's key always matches its own code hash, and the compiled module addressed by `key` is guaranteed correct. No wrong-module / cache-poisoning risk for the current caller. The `req_bytes` accounting, `_io_guard`, and `drop_running_instance` are preserved unchanged in the extracted `run_validate_state`. Disk-budget admission-gate ordering and all `remove_contract`/`record_wasm_removed` rollback paths are untouched.

---

### Big Picture Assessment
**Goal Alignment:** Yes — directly closes the approved #2216.
**Anti-Patterns Detected:** None. No `biased;` select, no fire-and-forget spawn, no wire-format/enum change, no unbounded-exemption GC logic, no bounded-channel `.send().await` in a recv loop.
**Removed Code Concerns:** None — the old fetch-from-store branch is preserved for the `None` path; nothing was deleted, only refactored/extracted.
**Scope Assessment:** Focused. Single commit, exactly 4 files, no unrelated cleanup riding along. `refactor:` prefix is correct (no behavior change).

---

### Documentation
- Code docs: Complete. New methods carry thorough rustdoc explaining the `None` vs `Some` semantics and referencing #2216.
- Architecture docs: N/A (no architectural surface change).
- User docs: N/A.

---

### Recommendations

#### Must Fix (Blocking)
- None.

#### Should Fix (Important)
- None.

#### Consider (Suggestions)
1. Add `debug_assert_eq!(&contract.key(), key)` at the top of `prepare_contract_call_with_contract` (or `validate_state_with_contract`) to pin the "supplied container must match `key`" invariant that current correctness silently relies on, so a future mismatched caller trips in debug builds instead of caching a wrong module. Purely defensive — no behavior change for the current, correct caller.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** 63fa88de9a0a070bad656b052960033487e30b66

A clean, well-scoped, behavior-preserving optimization backed by a genuine regression test (verified to build and pass locally). The one suggestion is optional hardening, not a blocker. Recommend confirming the pending **Unit & Integration** CI job goes green before merge (the local run of the new test already passes). If a working external model (`codex`/`gemini`) becomes available, a quick independent pass would be nice-to-have but is not required given the low risk and the verified in-context adversarial review.

[AI-assisted - Claude]

PR #4781, opened 2026-07-12
fix(server): survive a peer restart without breaking a framed webapp

sanity:
## Problem
When a node restarts, its in-memory auth-token map is wiped, so the shell's token (minted when the shell HTML was served, and injected into every WebSocket the shell opens) is stale — the node rejects the reconnecting WS with `AUTH_TOKEN_INVALID` and closes it. A webapp running in the shell's **sandboxed iframe** (e.g. River) cannot re-mint a token or reload the top-level document itself (the sandbox withholds top navigation), so it self-navigated its own iframe. During the node's ring-rejoin window that request is served `connecting.html`, whose `<meta refresh …;url=/>` yanks the frame to the node root `/`. On hardened deployments `/` denies framing, so the browser refuses it (`Refused to display … X-Frame-Options: deny`) and the tab shows a broken "unhappy" page; even where `/` doesn't deny framing, the node dashboard replaces the app inside its own frame. Only a manual full-tab refresh recovers.

## Approach — a trusted, server-controlled recovery signal
The shell owns the WebSocket and the auth token, so it — not the sandboxed app — should drive recovery, on a signal the app **cannot forge**. The node already closes a stale-token socket; this PR makes that close carry a distinguishing **application close code** the shell trusts.

1. **Server (`client_events/websocket.rs`)** — on the invalid-token path, close the WebSocket with `Message::Close(Some(CloseFrame { code: 4401, reason: "AUTH_TOKEN_INVALID" }))` instead of a bare close. `AUTH_TOKEN_INVALID_CLOSE_CODE = 4401` is a named constant (WebSocket private-use range 4000-4999, mirroring HTTP 401). The serialized error is still sent first, unchanged. The close **frame** — unlike the message **body**, which a contract's own `GetResponse` state could echo verbatim — rides a socket the node itself closes, so a sandboxed contract cannot forge it.

2. **`shell_bridge.js`** — recover **autonomously** on the trusted close:
   - The real-WebSocket `onclose` triggers a recovery reload when `isTrustedStaleTokenClose(code, clientClosed)` holds — i.e. code `4401` on a close the shell did **not** initiate. Recovery re-fetches **this** shell HTML via a cache-busting `location.replace`, which re-mints the shell-owned token (exactly what a manual refresh does).
   - The old, **spoofable** iframe→shell `type:"reload"` message trigger and its message-body byte scan are **removed**. river#400's message becomes a harmless no-op on new nodes; #4781 no longer depends on river#400.
   - **Unforgeable:** the close-proxy marks iframe-requested closes (`_clientClosed`) and clamps any app-range (4000-4999) code the iframe asks for down to `1000`, so a contract can't manufacture a 4401 by asking the shell to close its own socket.
   - **Bounded, fail-closed:** the per-window reload cap lives in the `_freload` URL param (`<windowStartMs>-<count>`), which survives `location.replace` and is **not writable by the sandboxed contract**. Unlike `sessionStorage` (swallowed in private mode → resets every document → fails open), the URL is always present, so a reload loop is bounded even with no storage.

3. **`connecting.html`** — bounded, framing-aware recovery. A top-level ordinary connecting page still redirects to `/`; a page carrying `_freload` is a recovery reload that momentarily landed here during rejoin, so it retries the contract URL **in place** (never navigating a frame to `/`, the `X-Frame-Options` break) — but only within a **2-minute window**. After the window it stops looping: top-level falls back to `/`, framed shows a "reload the tab" message. No more unconditional 3s-forever loop. The `<noscript>` meta remains the no-JS top-level fallback.

This fixes the same peer-restart breakage for every framed Freenet webapp (River, Delta, …), not just River.

## Testing
- **Rust**
  - `client_events::websocket::tests::invalid_auth_token_closes_with_4401_code` — end-to-end WebSocket upgrade + client; asserts the node closes an invalid-token connection with close code **4401**.
  - `path_handlers::tests::bridge_js_reloads_shell_on_auth_token_invalid_close` — recovery is driven by the trusted 4401 close; the spoofable `type:'reload'` path is gone.
  - `path_handlers::tests::bridge_js_recovery_reload_is_unforgeable_and_bounded` — `_clientClosed` mark + app-code clamp + URL-param fail-closed cap.
  - `errors::tests::connecting_page_redirects_to_root_only_at_top_level` — top-vs-frame redirect, `_freload` retry-in-place, and the `RECOVERY_MAX_MS` bound.
  - The `!SHELL_BRIDGE_JS.contains("allow-top-navigation")` sandbox pin still holds.
- **Node asset unit tests** (`npm test`, wired into the lint-assets CI job)
  - `shell_bridge_reload.test.mjs` — `isTrustedStaleTokenClose` (server 4401 recovers, iframe-initiated 4401 ignored), `clampProxiedCloseCode`, and the fail-closed `reloadUrlCapDecision` loop bound (no storage).
  - `connecting_recovery.test.mjs` — `connectingRecoveryDecision` honors the timestamp bound (retry within window; stop after).
- `cargo fmt`, `cargo clippy --locked -- -D warnings`, Prettier + ESLint all clean.

## Review history
The first iteration used an iframe-initiated `type:'reload'` message gated by a message-body byte scan. Codex review found the scan **spoofable** (a contract's own `GetResponse` state can contain the marker) and the `sessionStorage` cap **fail-open** in private mode. This PR replaces that with the server-controlled 4401 close code above — an unforgeable, decoupled signal — and a URL-param fail-closed cap.

[AI-assisted - Claude]


sanity:
## Re-review after implementing the round-1 MAJOR fixes — 3 MAJOR issues remain (do not merge yet)

Implemented the round-1 fixes (connecting.html `_freload` retry; sessionStorage-persisted reload cap; shell-observed `AUTH_TOKEN_INVALID` gate). The Codex re-review + adversarial pass found the approach still has three MAJOR gaps:

1. **Stale-token gate is spoofable (design-level).** The shell arms `authTokenInvalidSeen` by byte-scanning inbound WS frames for `AUTH_TOKEN_INVALID` (`shell_bridge.js:559`). But a `GetResponse` echoes the contract's state verbatim (`websocket.rs:2036`), so a malicious framed contract can store that string and GET it to arm the gate, then send `type:'reload'`. Byte-scanning can't establish provenance. **Decision needed:** either accept the reload rate-cap as the real bound and demote this gate to best-effort defense-in-depth, or add a structurally-validated signal (a distinct server-side close code / message type — likely coupling to river#400).

2. **Reload cap fails open without sessionStorage.** Storage errors are swallowed (`shell_bridge.js:477/495`), so in private/storage-restricted mode each new document gets a fresh 3-permit window → the cross-document loop is unbounded. Fix: storage-independent counting (e.g. encode count/deadline in the top URL, which the contract can't tamper with) so it fails closed.

3. **connecting.html can infinite-loop.** `_freload` (any substring, no freshness) makes connecting.html reload the same URL every 3s forever if rejoin never succeeds, bypassing the shell cap. Fix: compare `Date.now()` against a `_freload` timestamp and fall back to `/` after a bounded recovery window.

Plus a NIT (`Uint8Array` view ignores `byteOffset`/`byteLength`; not currently exploitable since `binaryType` is fixed).

#2, #3, and the NIT are cleanly fixable shell-side. #1 is a security-model decision. Current state: rebased + round-1 fixes pushed (HEAD `8e2b6052`), tests green, but **not merge-ready** until #1 is decided and #2/#3 are fixed + re-reviewed.

[AI-assisted - Claude]

sanity:
## Full-tier review — server-4401 rework (HEAD `24dc76c4`)

This PR was reworked from the earlier spoofable byte-scan design to a **server-controlled trusted signal**: the node closes an invalid-token WebSocket with `CloseFrame { code: 4401 }` (`AUTH_TOKEN_INVALID_CLOSE_CODE`), and the shell recovers autonomously on that close. This deletes the untrusted iframe→shell reload trigger entirely and decouples from river#400.

### External-model substitution (multi-model-review rule)
The independent **external non-Claude pass could not run**: Codex is quota-limited (`usage limit … try again Jul 25`), and the Gemini CLI remains deprecated/unavailable. Per the rule's fallback, I substituted **four independent Claude reviewers, each with a distinct adversarial lens**, run blind to each other:

| Lens | Verdict |
|---|---|
| Security / forgeability | **SECURE** — no exploitable path |
| Regression / neighbor-file correctness | **SAFE-TO-MERGE** |
| Robustness / failure-mode | **ROBUST-TO-MERGE** |
| Code-first / test-quality | **CLEAN-TO-MERGE** |

Consensus: the freenet-core side is correct and well-guarded. The 4401 is unforgeable by a sandboxed contract (four barriers: `isTrustedStaleTokenClose` trust check, `_clientClosed` mark, `clampProxiedCloseCode`, and the same-origin WS-URL restriction), reaches only the shell (browser network stack), is emitted only on a genuine stale shell-minted token, and the reload cap is fail-closed + storage-independent + unbypassable (the sandbox lacks `allow-same-origin`/`allow-top-navigation`). Tests execute the real extracted functions (non-vacuous).

### Findings & disposition
- **[low-med] JS/Rust `4401` constant drift** — FIXED in `24dc76c4`: added pin `bridge_js_close_code_matches_rust_constant` tying `SHELL_BRIDGE_JS` to `AUTH_TOKEN_INVALID_CLOSE_CODE`.
- **[MEDIUM, cross-repo] Recovery is contingent on the framed app reconnecting its WebSocket** after the restart — the 4401 rides the *reconnect*, not the death of the original socket (which closes as 1006). The freenet-core side is correct, but a real River/Delta tab only recovers if stdlib's `WebApi` reopens its socket on an abnormal close. **This is the make-or-break to verify on the river#400 / stdlib side** — not a #4781 blocker. (The overstated "autonomous" comment was corrected in `24dc76c4`.)
- **[LOW] robustness follow-ups** (silent cap-exhaustion state; `connecting.html` `decodeURIComponent` guard; pre-existing unguarded `ws.close`) — filed as #4875.

### Note on the finishing commit
The delta from the reviewed `c3e6f2c8` to `24dc76c4` is exactly the drift-pin test (test-only) + a comment reword — pure-additive trivial scope, no production-behavior change, so the four-lens review above stands for the merged content.

Merge-gated on the pending CI (Unit & Integration / Simulation) going green on `24dc76c4`, via the merge queue.

[AI-assisted - Claude]

Issue #3698, opened 2026-03-29
Hosted contracts served stale after node restart — no network GET or subscription renewal


sanity:
## Problem

After a node restarts (or comes back online after extended downtime), previously-cached webapp contracts are served from stale local cache indefinitely. No network GET is triggered, no subscription is established, and the background renewal loop ignores these contracts.

User impact: A peer that was offline for 10 days is stuck showing a River UI build from March 10, even though a March 26 build exists on the network. Refreshing the page has no effect.

## Root Cause

A contradiction between two subsystems introduced by the interaction of PRs #3363, #3528, and #3546:

1. **GET handler short-circuits on `is_hosted`** (`client_events.rs:1270-1276`): When a contract is in the hosting cache and has local state, the GET returns immediately without querying the network. The comment at lines 1259-1264 says "hosted contracts have subscription renewal in progress via the background recovery loop."

2. **But subscription renewal doesn't cover hosted-only contracts** (`hosting.rs:764-826`): After restart, both `active_subscriptions` and `client_subscriptions` are empty (in-memory only, not persisted). `contracts_needing_renewal()` returns empty, so the background loop does nothing.

3. **The web endpoint never subscribes** (`path_handlers.rs:72-77`): `contract_home` sends GET with `subscribe: false`, so even if the GET reached the network, no subscription would be established.

### Timeline
- PR #3363 (Mar 1): Decoupled caching from subscription
- PR #3528 (Mar 11): Added 2-min startup window to renew hosted contract subscriptions
- PR #3546 (Mar 12): Removed that startup window — creating this gap
- The doc comment on hosting.rs:128 ("All hosted contracts get subscription renewal") is stale from pre-#3546

## Expected Behavior

After restart, hosted contracts should be refreshed from the network rather than served stale from local cache. The web endpoint should establish subscriptions so contracts stay fresh.

[AI-assisted - Claude]

PR #4784, opened 2026-07-13
fix(hosting): rehydrate local-hosting interest on restart so cached contracts serve locally (#4780)

### LINKED ISSUE #4780: Restart regression: restored hosting copies no longer served locally (serve-DURING gate keys off non-rehydrated InterestManager)
## Summary

After a successful contract GET, **restarting the node, then GET-ing the same contract again, no longer serves from the local cache** — the node routes a fresh GET to the network (slow). Reported by **Ivvor** (River user) with a code-level analysis; verified against current code.

This is a **regression introduced by serve-DURING** (#4748, commit `33a024e4`, 2026-07-09, shipped in 0.2.95+), not the intended #3698 baseline. It is currently **untracked**.

## Timeline of the client GET serve gate (`crates/core/src/client_events.rs`)

1. **Pre-#3698** — gate included `is_hosting_contract`, so it served *any* hosted contract after restart → the stale-serve bug.
2. **#3698 fix (`c60ff96a`, "don't serve stale hosted contracts after node restart (#3699)", Mar 28)** — removed the hosting term → gate = `connection_count == 0 || is_subscribed`. Post-restart routes to network. Intentional and correct.
3. **#3769 (`8cb6c357`, "track local client interest… fast cache serving", Apr 5)** — added a **narrower** serve term back: `is_hosting_contract(k) && has_local_client_access(k)`. **Both terms persist across restart** (`is_hosting_contract` reads the restored hosting cache; `has_local_client_access` is persisted to redb and restored via `load_persisted_entry`, pinned by `test_local_client_access_survives_restart_via_load`). So a locally-accessed contract **was served locally after restart again** (Apr–Jul).
4. **serve-DURING (`33a024e4`, "serve-DURING at the originator GET gate (#4748)", Jul 9)** — replaced that term with `interest_manager.has_local_interest(k)`, an **in-memory `DashMap` that startup never rehydrates**.

So the behavior Ivvor observes is a regression versus the immediately-prior Apr–Jul behavior, introduced by `33a024e4`.

## Root cause

The current gate (`should_serve_local_copy`, `client_events.rs:~156`):

```rust
fn should_serve_local_copy(local_satisfies_request: bool, connection_count: usize,
                           is_subscribed: bool, has_local_interest: bool) -> bool {
    local_satisfies_request && (connection_count == 0 || is_subscribed || has_local_interest)
}
```

On restart with connections re-established (`connection_count > 0`), `is_subscribed` is false (subscriptions are in-memory, lost on restart) and `has_local_interest` is false — because **startup restores the hosting cache and neighbor-hosting advertisements (`contract/handler.rs:145-172`, `load_hosting_cache` + `neighbor_hosting.initialize_from_hosting_cache`) but never rebuilds `InterestManager` local-interest** for the restored keys. There is no `register_local_hosting` / `register_local_interest` call on the boot path. So the gate → route to network.

## Severity: every GET after restart, not just the first

`register_local_hosting` (`ring/interest.rs:694`) has exactly one production caller — `operations/get/op_ctx_task.rs:1352`, gated on `access_result.is_new && put_persisted`, i.e. **only when a contract is newly cached in the current process**. After restart the contract is *already* in the restored hosting cache, so a network GET returns `is_new = false` → `register_local_hosting` is **not** called → `has_local_interest` stays false → **every** subsequent GET keeps routing to the network. Matches Ivvor's "no longer serves from cache."

Compounding: the InterestSync anti-entropy loop iterates `has_local_interest` (`node.rs:2676`, `:2864`), so the restored copy is **also not freshened in the background** — it falls into a gap: routed-to-network (slow) *and* not joined to anti-entropy. (Renewal via persisted `local_client_access` and the serve gate are now two disjoint subsystems — the disconnect is the bug.)

Edge case that matches the report: a **zero-connection** node short-circuits the gate (`connection_count == 0`) and still serves the persisted copy; once connected, the missing interest entry sends the GET back to the network.

## Proposed fix (serve-DURING-correct, invariant 1)

On startup, for each restored hosting key call `interest_manager.register_local_hosting()` (alongside the existing neighbor-hosting init). This sets `hosting = true`, which:
- makes the serve gate return the local copy immediately, and
- enrolls the copy in InterestSync anti-entropy (the ~5-min unfiltered heal against connected co-hosts).

That is exactly the invariant-1 permitted transient: **serve the best local copy immediately, re-link in the background, never block the read.** It does **not** re-introduce #3698 (whose bug was serve-stale-forever with no freshening path — there was no anti-entropy then).

Fuller fix: additionally, for keys carrying persisted demand (`local_client_access`, and true client/delegate subscriptions once #4669 lands) kick a background re-subscribe / re-root so updates flow live rather than only via the ~5-min heartbeat.

**Do NOT** simply re-add `is_hosting_contract` to the gate — that would serve copies with no freshening path and re-introduce #3698.

## Invariants / review / test

Touches **hosting-invariant 1** ("a served copy is fresh"; the serve-DURING permitted transient) and the GET serve/hosting path directly → **Full-tier multi-model review** + a simulation test asserting the restart→serve behavior (GET a contract, simulate restart with connections, second GET must serve locally and the copy must be enrolled in anti-entropy).

## Evidence (file:line)

- gate: `crates/core/src/client_events.rs:156` (`should_serve_local_copy`), `:966`/`:977` (computation + call)
- startup restore (no interest rehydration): `crates/core/src/contract/handler.rs:145-172`
- sole `register_local_hosting` caller (gated `is_new && put_persisted`): `crates/core/src/operations/get/op_ctx_task.rs:1352`
- `register_local_hosting`: `crates/core/src/ring/interest.rs:694`; `has_local_interest`: `:831`
- `local_client_access` persistence + restart test: `hosting.rs:2058`, test `hosting.rs:5405`
- anti-entropy interest gate: `crates/core/src/node.rs:2676`, `:2864`
- commits: `33a024e4` (#4748), `8cb6c357` (#3769), `c60ff96a` (#3699)

## Credit

Diagnosis and fix shape by **Ivvor** (Matrix / River user) — verified against current code. Relates to epic #4642 (demand-driven hosting); not covered by #4669 or #4651.

@iduartgomez — this is in the serve-DURING (#4748) path you shipped, flagging for visibility.

[AI-assisted - Claude]

sanity:
## Problem

After a successful GET, restart, then GET again, the node stopped serving the contract from its local cache: it routed the whole request through the network (slow) and the restored copy never re-joined anti-entropy. Reported by a River user (Ivvor) with a code-level analysis; verified.

Root cause: serve-DURING (#4748) re-keyed the client-GET serve gate (`client_events::should_serve_local_copy`) and the InterestSync heartbeat onto `InterestManager::has_local_interest`. Startup restore (`Ring::load_hosting_cache` + `NeighborHostingManager::initialize_from_hosting_cache`) rebuilds the hosting cache and the advertised-host set but never rehydrates the InterestManager, and the only `register_local_hosting` callers (GET/PUT host-formation) fire on `access_result.is_new` — false for a contract already in the restored cache. So every post-restart GET saw `has_local_interest == false` and, with connections re-established and no live subscription, routed to the network while the copy stayed out of the interest hash index (no anti-entropy). Unchanged by #4782, which reworked the subscribe/hosting paths but not startup restore or the serve gate.

## Approach

On startup, after the existing hosting-cache / neighbor-hosting restore, re-register local-hosting interest for every restored hosted contract whose state is on disk (`OpManager::rehydrate_local_hosting_interest`). This makes the serve gate answer locally AND re-enrolls the copy in the interest hash index so the ~5-min InterestSync heartbeat heals it and neighbors learn we co-host it — the serve-DURING permitted transient in hosting-invariants invariant 1 (serve immediately, re-link in the background, never block the read).

Register-after-state (#4782): a restored metadata-only key with no on-disk state is skipped, so no phantom interested-but-stateless entry (#4440). The serve gate is deliberately NOT widened with `is_hosting_contract` — that would serve copies with no freshening path and re-introduce the #3698 stale-serve bug; guard comments added at both the gate and the restore site.

Touches hosting-invariants invariant 1 (a served copy is fresh) and preserves it by re-linking restored copies into anti-entropy rather than serving an unfreshened cache copy.

## Testing

- `restart_rehydrates_local_hosting_interest_so_cached_get_serves_locally` drives the real `load_hosting_cache` restore against a redb store, then the fix; asserts (via the real `should_serve_local_copy` gate) that the restored state-present copy is routed-to-network pre-fix and served-locally post-fix, and that it rejoins the interest hash index (anti-entropy), the stateless copy is skipped — all with `connection_count > 0` and no subscription. Verified red without the fix, green with it.
- `handler_build_rehydrates_local_hosting_interest_on_restart` — source-scrape pin that startup build calls the rehydration.

Closes #4780

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -161,6 +161,17 @@ async fn register_subscription_listener(
 /// near-miss floor. `local_satisfies_request` already guarantees valid state is
 /// present, so a phantom (interested-but-stateless, #4440) copy is excluded even
 /// though it can register interest via a downstream subscriber.
+///
+/// Do NOT re-add `is_hosting_contract` (the in-memory hosting-cache membership)
+/// as a serve term — see .claude/rules/hosting-invariants.md (invariant 1). A
+/// hosting-cache copy is NOT necessarily in the update mesh: on restart the
+/// cache is restored but the InterestManager is not, so `is_hosting_contract`
+/// can be true while the copy has no freshening path. Serving on it would
+/// re-introduce the #3698 stale-serve bug. The correct fix for restored copies
+/// is to rehydrate `has_local_interest` at startup
+/// (`OpManager::rehydrate_local_hosting_interest`, #4780), which registers the
+/// copy into anti-entropy so `has_local_interest` (term 3) is a true
+/// fresh-in-mesh signal — NOT to widen this gate.
 fn should_serve_local_copy(
     local_satisfies_request: bool,
     connection_count: usize,
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -172,7 +172,12 @@ async fn register_subscription_listener(
 /// (`OpManager::rehydrate_local_hosting_interest`, #4780), which registers the
 /// copy into anti-entropy so `has_local_interest` (term 3) is a true
 /// fresh-in-mesh signal — NOT to widen this gate.
-fn should_serve_local_copy(
+// `pub(crate)` so the restart-rehydration regression test
+// (`node::op_state_manager::tests`) can assert the ACTUAL serve decision for a
+// restored contract, not just the `has_local_interest` input — guarding against
+// a future refactor that decouples the gate from `has_local_interest`. Pure fn,
+// no state, safe to expose in-crate.
+pub(crate) fn should_serve_local_copy(
     local_satisfies_request: bool,
     connection_count: usize,
     is_subscribed: bool,
--- a/crates/core/src/contract/handler.rs
+++ b/crates/core/src/contract/handler.rs
@@ -174,6 +174,23 @@ impl ContractHandler for NetworkContractHandler {
             .neighbor_hosting
             .initialize_from_hosting_cache(hosted_ids);
 
+        // #4780: also rehydrate InterestManager local-hosting for every restored
+        // hosted contract, so a client GET for a cached contract serves LOCALLY
+        // (serve-DURING gate) instead of routing to the network, and the copy
+        // re-joins InterestSync anti-entropy. The two restore steps above rebuild
+        // the hosting cache and the advertised-host set but NOT the
+        // InterestManager, which both the serve gate and the interest-sync
+        // heartbeat key on. State-gated inside the helper (register-after-state,
+        // #4782) so no phantom interested-but-stateless entry is created — the
+        // hosting storage handle is set above (`set_hosting_storage`), so the
+        // state-presence check is real here.
+        //
+        // Do NOT "fix" this instead by re-adding is_hosting_contract to the serve
+        // gate (`client_events::should_serve_local_copy`) — that would serve
+        // copies with no freshening path and re-introduce the #3698 stale-serve
+        // bug. See .claude/rules/hosting-invariants.md (invariant 1).
+        op_manager.rehydrate_local_hosting_interest();
+
         Ok(Self { executor, channel })
     }
 
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -552,6 +552,58 @@ impl OpManager {
         })
     }
 
+    /// Re-register local-hosting interest for every restored hosted contract
+    /// on restart (#4780).
+    ///
+    /// Startup restore repopulates the hosting cache
+    /// ([`Ring::load_hosting_cache`]) and the advertised-host set
+    /// ([`NeighborHostingManager::initialize_from_hosting_cache`]), but NOT the
+    /// [`InterestManager`](crate::ring::interest::InterestManager). Both the
+    /// client-GET serve gate (`client_events::should_serve_local_copy`, which
+    /// keys on `InterestManager::has_local_interest`) AND the InterestSync
+    /// anti-entropy heartbeat (`node::handle_interest_sync_message`, which keys
+    /// on `has_local_interest` / `get_matching_contracts` — i.e. the interest
+    /// hash index) read the InterestManager. So without this call a restored,
+    /// still-hosted contract silently fell out of BOTH local serving and
+    /// anti-entropy: after a restart with connections re-established (so the
+    /// serve gate's `connection_count == 0` escape hatch no longer applies) and
+    /// no live subscription, every subsequent client GET routed the whole
+    /// request through the network (slow), and the copy never re-joined the
+    /// update mesh. serve-DURING (`hosting-invariants.md` invariant 1) requires
+    /// the reverse: serve the local copy immediately and re-link it in the
+    /// background.
+    ///
+    /// Registering `hosting = true` (the same reason the GET/PUT host-formation
+    /// paths register on first cache) is exactly what restores both: the serve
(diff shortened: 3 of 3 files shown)


sanity:
## Multi-model review — Full tier (hosting/placement surface)

Reviewers: **Codex** (external) + four Claude perspectives — code-first, testing, skeptical/adversarial, big-picture. All reviewed the checked-out code against `.claude/rules/hosting-invariants.md`.

**Verdict: ship-worthy, no blocking bug.** The two load-bearing invariants were verified preserved:

- **No manufactured demand (invariant 2).** The `hosting=true` flag feeds *only* the serve gate, the interest hash index (anti-entropy participation), and neighbor advertisement — it is explicitly excluded from eviction ranking (`local_and_downstream_counts` uses subscriptions), renewal (`contracts_needing_renewal` gates on subscriptions), and `active_demand_count` (the #3763 no-storm denominator). A rehydrated zero-subscriber contract still ranks first-eviction. Confirmed independently by the code-first and skeptical passes.
- **Eviction cleans up.** Every eviction path already calls `unregister_local_hosting`, so no lingering/zombie interest is introduced.
- **Restores invariant 1** rather than patching a symptom: startup restore was silently violating "a hosted copy registers as interested and joins anti-entropy while connected"; this re-establishes it. The serve gate is deliberately *not* widened with `is_hosting_contract` (would re-open #3698).

### Findings and resolutions

- **State-gate backend scope (Codex P1/P2, skeptical finding 4):** on the sqlite backend `contract_state_present` returns conservative-`true`, so a restored metadata-only key would be registered. Assessed **LOW** by all four Claude reviewers: it's *pre-existing* (the same conservative behavior the #4610 summarize gate and #4782 register-after-state rely on), sqlite-only, moot on the production redb backend (handle set before the call), and never causes a mis-serve (the serve gate still requires `local_satisfies_request`). **Documented** in a backend-scope rustdoc note rather than tightening the shared gate here; the whole-system state+code gate robustness is a separate follow-up.
- **Test strengthened (testing reviewer):** the regression test now asserts the **actual `should_serve_local_copy` gate** (false pre-fix / true post-fix), not just `has_local_interest` — verified red-without / green-with, and it specifically trips if a future refactor decouples the gate from `has_local_interest`.
- **Deferred live-update re-subscribe (big-picture):** for a copy with *real prior demand*, live updates resume via client re-subscribe or the ~5-min heartbeat rather than an immediate re-subscribe (which for demandless copies would manufacture demand — an anti-pattern). Filed as non-blocking follow-up **#4785** (persist demand across restart + background re-subscribe).

### Maintainer sign-off on the restored-copy freshness tradeoff

The skeptical pass flagged, for explicit sign-off, that a node **down for a long time** will on restart **serve its cached copy immediately** and heal only via the ~5-min anti-entropy heartbeat — so it can serve an arbitrarily-stale copy for a bounded window before the first heal. This is exactly the settled **serve-DURING** model (invariant 1's permitted transient) and is the point of the fix (the alternative is the current slow regression). **@sanity explicitly confirmed** the speed-over-freshness tradeoff for restored copies is intended.

CI green on the current HEAD; adding to the merge queue.

[AI-assisted - Claude]

Issue #4754, opened 2026-07-10
WASM module cache: couple contract-module removal to hosting eviction (reclaim promptly, not on cold-LRU pressure)


sanity:
The wasmtime module cache grows unbounded (`contract_module_cache_evictions_total=0`). On try.freenet.org's 0.2.95 soak it grew 260 MB → 891 MB (172 → 629 entries) over ~11.5h; other binaries grew faster (~1.2-1.5 GB). On a very-long-lived node with many hosted contracts (2339 here) it could reach ~3 GB. Not urgent (gateways have large RAM headroom; non-cache RSS is flat) but a real unbounded-growth surface. Fix: add LRU eviction to the module cache. Pre-existing, NOT introduced by 0.2.95 — found during the 0.2.95 soak. [AI-assisted - Claude]

sanity:
**Correction after investigation — this is NOT an unbounded leak; original title was wrong.**

The module cache (`wasm_runtime/module_cache.rs`) is a **byte-budget-bounded LRU**, not insert-only: `evict_to_budget` pops LRU entries on every insert once over budget. Budget is RAM-scaled + cgroup-aware: `clamp(total_ram/8, 64 MiB, 1.5 GiB)` for contracts (delegates /4). On the 125 GB soak box the cap is 1.5 GiB and it reached 891 MB = 58% → `evictions_total=0` was **headroom**, as designed. A small peer gets a small budget and evicts. It's co-sized with the hosting-state gauge (same `/8` RAM divisor, mirrored intentionally), so memory is bounded end-to-end — **no OOM path, no gap in the demand epic's bounding.**

**The real (minor) finding:** unlike the delegate path (`delegate/interface.rs:308` calls `delegate_modules.remove` on unregister), the contract path's `reclaim_contract_storage` (`contract/executor/runtime/contract_ops.rs:428`) deletes state + the .wasm blob but never removes the compiled module — so an evicted contract's module **lingers** until byte pressure. It's classified cold (`contract_in_use=false`) → first-evicted, so reclaimed lazily not leaked.

**Fix (low priority, tidiness):** add `contract_modules.remove(key)` in `reclaim_contract_storage` on the code-success branch, mirroring the delegate path — reclaims ~1.4 MB promptly on eviction instead of on cold-LRU pressure. Optional larger change: dedup the cache by CodeHash instead of ContractKey (collapses duplicate entries for contracts sharing WASM, e.g. River rooms). [AI-assisted - Claude]

[...]

Issue #4741, opened 2026-07-08
Log hygiene: rate-limited peer_connection / summarize_contract_state spam during peer reconnection


sanity:
## Problem

During peer reconnection, the node emits rate-limited log spam — `peer_connection`, `summarize_contract_state`, and `process_network_message` lines — bursting to hundreds/1000s per 30s before the rate-limiter tamps them down.

Found during the 0.2.94 release canary on the "framework" peer: new process emitted 30 rate-limited lines over 20 min (bursty, self-limiting: a spike of 585/30s at one point, back to 1/30s a minute later). **Pre-existing, NOT a 0.2.94 regression** — the current 0.2.93 release is noisier (bursts to 1336/30s, ~186k cumulative over its uptime). 0.2.94 is actually quieter.

Not urgent (rate-limited, doesn't fill disk), but worth log-hygiene cleanup: these lines are high-volume during the normal reconnection ramp and add noise to any log investigation.

## Where to look

`peer_connection`, `summarize_contract_state`, `process_network_message` log sites — likely emitting at `info`/`warn` on a hot reconnection path. Consider downgrading to `debug`/`trace` or widening the rate-limit window.

Found on: 0.2.94 canary (framework peer), also present on 0.2.93.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-07-12T17:53:54.632927838+00:00
Issue hash: `8b93ee15`

## Value Assessment

[...]

PR #4794, opened 2026-07-13
perf(hosting): size summary/delta caches to hosted-set scale to stop anti-entropy re-summarize thrash

sanity:
## Problem

0.2.98 every-hop hosting means each gateway hosts ~2600 contracts, and the ~5-min InterestSync anti-entropy re-summarizes the whole set. The summary/delta caches were **per-executor** `LruCache(1024)`, duplicated across up to 16 pool workers, so ~1600 contracts/cycle miss the cache → slow path (reload state + WASM `summarize_state` → module recompile). Measured on the prod gateways after 0.2.98: summarize calls ~250x (~2/min → 500-650/min), ~36 module recompiles/sec, 55-83% CPU on both gateways, and the fastest feedstock for the musl allocator high-water-mark (#4565).

## Approach

Replace the per-executor `LruCache`s with a single **pool-shared moka cache**, sized from the hosting byte budget (`Config::max_hosting_storage`):

```
count = clamp(hosting_budget_bytes / 128 KiB, 1024, 100_000)
```

At the default RAM-scaled budgets: 128 MiB budget → 1024, **1 GiB budget → 8192** (covers the observed ~2616 hosted with ~3× headroom). Freshness is **unchanged**: the fast path returns a cached summary only when the shared change-detector hash matches the cached hash, and the detector is invalidated on every write — sharing the cache adds no staleness because the (already-shared) detector is the gate. Net RAM is neutral-to-negative (one shared 8192-entry cache ≈ 25-50 MB vs the old N×1024 fragmented ≈ 52 MB at N=16) while covering 8× more contracts — the right direction for #4565.

The sizing constants (128 KiB state floor, 1024 min, 100k ceiling) are named + documented inline as the hosting-cost-model knobs.

## Testing

- `summarize_unchanged_hits_fast_path_beyond_old_1024_cap`: warms 1500 distinct contracts (> old 1024 cap), re-summarizes an unchanged one, asserts no state reload (fast path, hence no WASM summarize). **Verified it FAILS when the cache is undersized** (512 → panic).
- 5 sizing-derivation unit tests. 206 `contract::executor` + 32 `state_store` tests green; `cargo fmt` + `clippy -D warnings` clean.

Live validation (deploy → watch the gateway summarize rate + CPU drop) to follow after merge. No observable behavior change (freshness preserved); this is a non-overreaching perf improvement.

Relates to #4565 (musl HWM / jemalloc), #4754, #4741.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -925,10 +925,19 @@ pub(crate) type SummaryCache = MokaCache<ContractKey, (u64, StateSummary<'static
 /// Value cached per `(contract, state-hash, peer-summary-hash)` for the
 /// `get_contract_state_delta` fast path. Shared across pool executors like
 /// [`SummaryCache`].
+///
+/// Unlike [`SummaryCache`] this is BYTE-bounded, not entry-bounded: a
+/// `StateDelta` is variable and can be large (up to `MAX_STATE_SIZE` = 50 MiB —
+/// a contract may legitimately return its full state as a delta), and the key's
+/// cardinality (one entry per distinct peer-summary per contract) exceeds the
+/// contract count, so an entry-count cap reused from the tiny-summary cache could
+/// pin multi-GB of RAM (the #4565 OOM class). See [`build_delta_cache`] /
+/// [`delta_cache_capacity_bytes`].
 pub(crate) type DeltaCache = MokaCache<(ContractKey, u64, u64), StateDelta<'static>>;
 
-/// Conservative per-contract on-disk state floor used to size the summarize/delta
-/// fast-path caches from the node's hosting byte budget.
+/// Conservative per-contract on-disk state floor used to size the SUMMARY
+/// fast-path cache (entry-count) from the node's hosting byte budget. The delta
+/// cache is byte-bounded separately (see [`delta_cache_capacity_bytes`]).
 ///
 /// Chosen well BELOW the ~400 KiB average contract-state size observed on
 /// production gateways (a 1 GiB hosting budget held ~2616 contracts) so the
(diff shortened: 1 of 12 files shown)


sanity:
## Consolidated multi-model review (Codex + 3 adversarial Claude lenses: concurrency, hosting-invariants, code-first/testing)

**Verdict: approach is correct and freshness is unanimously confirmed safe** — the fast path is gated on `detector_hash == cached_hash`, the detector is invalidated on every write (incl. the bypass path), `summarize_state` is deterministic across executors, and the detector was already pool-shared (#4621), so sharing the summary/delta caches introduces no cross-executor staleness (it's sounder than the old per-executor cache). Caching is the right fix — gating the anti-entropy heartbeat to changed contracts would break heal-on-mismatch.

### Must-fix before merge
1. **Replacement executors bypass the shared cache** (pool.rs `create_replacement_executor` ~:684-708) — flagged by Codex + concurrency lens. The shared caches are locals in `RuntimePool::new`, never stored as struct fields, so `create_replacement_executor` (fires on WASM-panic / off-loop-export-panic) builds an executor with the private 1024 cache. ~1/N of every heartbeat's summarize/delta calls then land on a replacement, miss, and slow-path — reintroducing the storm, accumulating over the pool's lifetime, never self-healing. **Fix:** store the two shared caches as `RuntimePool` fields; call `set_summarize_caches` in `create_replacement_executor`.
2. **Delta cache is count-bounded, not byte-bounded** (executor.rs `build_delta_cache` + the shared sizing) — flagged by Codex + hosting + concurrency. `StateDelta` is variable and can be large (up to `MAX_STATE_SIZE` 50 MiB; a contract can return full state as a delta), and the key is `(contract, state_hash, peer_summary_hash)` so cardinality exceeds contract count. Reusing the summary entry-count (up to the 100k ceiling on a large operator budget) risks multi-GB RAM — the exact OOM class we're avoiding given #4565. **Fix:** give the delta cache a byte-based weigher + a byte-bounded cap, separate from the summary entry-count.
3. **Test doesn't guard the actual fix** (summarize_delta_cache_tests.rs) — flagged by code-first/testing. The headline scale test injects into a single executor, so it proves the *sizing* but would pass identically with a big per-executor cache; it doesn't guard pool-shared injection. Per the repo's own migration-wiring pin rule, dropping the injection must fail CI. **Fix:** add a source-scrape pin that `RuntimePool::new` AND `create_replacement_executor` both call `set_summarize_caches(summarize_cache_capacity(config.max_hosting_storage))`, plus ideally a two-executor sharing test.

### Should-fix
4. **Under-coverage on small-contract nodes** (the `budget / 128 KiB` count derivation) — a node hosting many sub-128KiB contracts under-covers and the storm returns for that class. Add a summarize slow-path-rate telemetry counter so recurrence is detectable (and/or size from the actual hosted-contract count).
5. **Shared cache widens the blast radius of a future off-loop summarize/delta call** — add a source-scrape pin/comment that off-loop executor work must not call summarize/delta/cache_state_hash.
6. **`SUMMARIZE_CACHE_MAX_CAPACITY = 100_000` duplicates `state_store::STATE_HASH_CACHE_CAPACITY`** (manual-mirror-that-rots) — reference the source constant instead.

### Nice-to-have
7. delta-cache scale test (mirror the summary one); 8. freshness-after-resize test (append UPDATE+recompute); 9. doc wording fix (executor_impl.rs "after a restart" case).

Addressing all of the above in a revision pass; re-review will follow since the code changes materially.

[AI-assisted - Claude]

sanity:
### Re-review resolution (revision `696358ac` + floor fix `62cea5cd`)

The revision was put through a full re-review (Codex + an adversarial Claude lens). **Items verified correct:** the shared caches are now `RuntimePool` fields injected at all three executor-construction paths incl. `create_replacement_executor` (the storm can't creep back via a WASM-panic replacement); the summarize slow-path telemetry counter is at the right site (only when the WASM path runs) with no mirror-rot; all four new pins genuinely *fail* on their regressions; and the mirror-drift const now references `state_store::STATE_HASH_CACHE_CAPACITY`.

**One regression the revision introduced, caught independently by both models:** the delta byte-weigher had no per-entry floor, so an empty/tiny delta weighed ~0 and never triggered eviction. Combined with delta entries never being invalidated on state change (every UPDATE mints a new `(key, state_hash, summary_hash)`), the delta cache could grow **unbounded in entry count** — reintroducing the OOM class the byte-bound was meant to close, and making the earlier "64 MiB worst-case" claim false.

**Fixed in `62cea5cd`:** floored the weigher at `DELTA_ENTRY_OVERHEAD_FLOOR_BYTES = 512`, so entry count is bounded to `byte_budget / 512` (16k–131k) and worst-case delta RAM to ≤ ~1.5× the byte cap (~96 MiB at the 64 MiB gateway cap). New test `delta_cache_empty_deltas_stay_entry_bounded` inserts 2000 empty deltas and asserts the cache stays bounded — verified to fail without the floor. No TTL (moka's wall-clock idle conflicts with the crate's DST/TimeSource determinism).

[AI-assisted - Claude]

sanity:
Superseded by #4802. 

#4794 sized the summarize/delta cache via `clamp(hosting_budget/128KiB, 1024, MAX)` — but that's a magic-number proxy (assumes avg contract state > 128KiB); if contracts got smaller, the hosted count could again outgrow the cache and the thrash would recur. #4802 instead ties the cache to the node's **actual live hosted count** (`Ring::hosting_contracts_count()`), resizing to cover it (clamped to a MAX backstop), so coverage is guaranteed for any count regardless of contract size — no magic number. Maintainer requested the guaranteed, metric-tied version. Closing in favor of #4802.

[AI-assisted - Claude]

PR #4796, opened 2026-07-14
fix(telemetry): capture client-visible GET outcomes incl. streaming successes

sanity:
## Problem

`get_success` telemetry is synthesized ONLY from an inline
`GetMsg::Response{result: Found, value.state.is_some()}` in
`crates/core/src/tracing/register.rs`. Any GET whose payload exceeds the
64 KB `streaming_threshold` (default 65536) is delivered as
`GetMsg::ResponseStreaming`, which falls through to `_ => EventKind::Ignored`.

River rooms and UI-container contracts are all > 64 KB, so **their
successful GETs emit NO telemetry event at all** — the network's
findability is unmeasurable for exactly the contracts that matter.

Two secondary gaps:
- **Pure timeout-exhaustion** returned `NotFound` to the client with NO
  event (only a `tracing::info!`).
- **Sub-operation GETs** (phantom-repair, renewal, related-fetch) could
  not be distinguished from real client GETs, so they polluted any
  findability metric derived from GET events.

## Approach

Emit **exactly ONE** client-visible terminal event per client GET op from
the **originator / sub-op driver's terminal arms** (`drive_client_get_inner`
and `drive_sub_op_get` in `operations/get/op_ctx_task.rs`), which see the
true client outcome regardless of inline-vs-streaming delivery. Because it
fires at driver completion, it captures the streaming successes the inline
`register.rs` path drops — that is the core fix.

New `GetEvent::ClientTerminal` fields:
- `outcome`: `success` | `not_found` | `timeout_exhausted`
- `streamed: bool` (payload arrived via `ResponseStreaming`)
- `is_sub_op: bool`
- `attempts`, `hop_count` (when carried by the terminal reply), `key`

**Volume-conscious (explicit constraint — do not overwhelm the telemetry
server).** This is ONE event per client GET op — far FEWER than the
existing per-hop `get_request` / `get_not_found` events (which fire at
every hop). Net volume rises by ~1 event per client GET op (+1 per sub-op
GET); client GET ops are far fewer than total GET hops, so total volume
does not materially rise. **No per-hop or per-attempt events were added.**

**`is_sub_op` tagging** is passed EXPLICITLY by each driver rather than
derived solely from `tx.is_sub_operation()`: the sub-op GET drivers
construct their tx via `Transaction::new` (no parent), so
`is_sub_operation()` alone would wrongly report `false` for them. The
client driver tags from `client_tx.is_sub_operation()`; the sub-op driver
tags `true`.

**not_found vs timeout_exhausted** is distinguished by a new
`GetRetryDriver.saw_not_found` flag (GET's only `AttemptOutcome::Retry`
source is a `NotFound` reply) — no change to the shared retry loop.

This unblocks measuring the network's real findability, including for the
> 64 KB contracts that today are completely invisible.

### Note on the #4658 consult-counter export

The task also asked to export the terminal-consult counters
(`terminal_consult_attempts` / `hits` / `resolved_found` /
`still_not_found`). Verified this is **already complete in main**: the
fields exist in `RouterSnapshotInfo`, are populated in `ring.rs`, mirrored
in the OTLP JSON, and pinned by
`router_snapshot_json_includes_terminal_consult_counters`. No change
needed; the pin test stays green.

## Testing

- Pure outcome-mapping unit test (`done_terminal_outcome` /
  `exhausted_terminal_outcome`, all cases + stable OTLP `outcome` strings)
- Source-scrape pins: both client arms and both sub-op arms emit exactly
  once; client tags `is_sub_op` from the tx; sub-op tags it explicitly;
  the streaming Done arm sets `streamed=true`
- Telemetry JSON round-trip tests: a streaming (> 64 KB) success serializes
  to `get_terminal` / `outcome=success` / `streamed=true` (the core-bug
  regression guard); a timeout-exhausted GET serializes to
  `outcome=timeout_exhausted` with no key
- `cargo fmt` + `cargo clippy -p freenet -- -D warnings` clean; the
  `tracing`, `operations::get`, and `network_status` suites are green
  (217 passed, 0 failed)

[AI-assisted - Claude]

https://claude.ai/code/session_0199hK8Vdn5DAWcXJsxZHk8H

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -879,6 +879,16 @@ async fn process_open_request(
                                             phase = "local_cache_pre_join",
                                             "Serving locally cached contract state before network join"
                                         );
+                                        // Local-cache hit bypasses start_client_get, so
+                                        // emit the client-terminal event here (attempts=0
+                                        // marks a local hit) to keep the findability
+                                        // metric covering ALL client GET successes.
+                                        get::op_ctx_task::emit_local_get_terminal_event(
+                                            &op_manager,
+                                            key,
+                                            full_key,
+                                        )
+                                        .await;
                                         return Ok(Some(Either::Left(QueryResult::GetResult {
                                             key: full_key,
                                             state,
@@ -1100,6 +1110,16 @@ async fn process_open_request(
                                 }
                             }
 
+                            // Local-cache hit (serve-DURING / subscribed) bypasses
+                            // start_client_get, so emit the client-terminal event
+                            // here (attempts=0 marks a local hit) to keep the
+                            // findability metric covering ALL client GET successes.
+                            get::op_ctx_task::emit_local_get_terminal_event(
+                                &op_manager,
+                                key,
+                                full_key,
+                            )
+                            .await;
                             return Ok(Some(Either::Left(QueryResult::GetResult {
                                 key: full_key,
                                 state,
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -3858,10 +3858,16 @@ impl SimNetwork {
                         GetEvent::Request { .. } => summary.get.requested += 1,
                         GetEvent::GetSuccess { .. } => summary.get.succeeded += 1,
                         GetEvent::GetFailure { .. } => summary.get.failed += 1,
+                        // ClientTerminal is emitted IN ADDITION to
+                        // GetSuccess / GetNotFound (once per client op from
+                        // the driver), so counting it here would double-count
+                        // the sim summary; the GetSuccess arm above already
+                        // tallies succeeded.
                         GetEvent::GetNotFound { .. }
                         | GetEvent::ResponseSent { .. }
                         | GetEvent::ForwardingAckSent { .. }
-                        | GetEvent::ForwardingAckReceived { .. } => {}
+                        | GetEvent::ForwardingAckReceived { .. }
+                        | GetEvent::ClientTerminal { .. } => {}
                     }
                 }
                 // Subscribe operations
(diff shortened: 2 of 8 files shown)


sanity:
### Multi-model review resolution

Reviewed by Codex (twice) + two adversarial Claude lenses (correctness, wire-compat). Core change confirmed **sound**: emission-only (no routing/retry/caching change), exactly-once per client op, correctly `is_sub_op`-tagged, wire/bincode-compatible (appended `pub(crate)` `GetEvent` variant, not a protocol change), and it fixes the streaming-GET-success blindness (verified: a >64 KB `ResponseStreaming` success now emits `outcome=success, streamed=true`).

**Fixed (round 1):** `attempts` off-by-one on exhaustion (now a `requests_sent` counter); blocking-subscribe timing (emit after the hand-off so `elapsed_ms` is right and a cancellation can't record false success); OT-span leak under `trace-ot` (`ClientTerminal` now span-completing, vacant-entry no longer re-creates a span); local-cache-hit coverage (both local-serve paths emit `success, attempts=0`).

**Deferred as documented fast-follows (round 2 — edge-case precision on secondary fields; the primary `outcome`/`streamed`/`is_sub_op` signal is accurate on all normal paths):**
- Pre-join local hits emit nothing — a peer with no known address yet gets `None` from `get_terminal`; this is the general pre-join telemetry-dark behavior (such a peer emits no events at all), not specific to this PR. Joined peers (incl. gateways) do emit their local hits.
- Local-hit `elapsed_ms` ≈ 0 — tx created after the cache lookup; approximately correct for a genuine local hit, understated only for blocking-local-subscribe.
- `attempts` may over-count by 1 on send-channel backpressure/closed-sender (increment happens in `build_request`, before the send is confirmed on the channel). Rare infra-failure edge; the `outcome` field remains the reliable signal.

**Pre-existing, unrelated (not a CI gate — CI runs `clippy --locked` without `--tests`):** 4 `let_underscore_must_use` errors in `wasm_runtime/contract_store.rs` on clippy 1.94.0, surfaced only by `clippy --tests`. Flagged for a separate cleanup.

[AI-assisted - Claude]

PR #4799, opened 2026-07-15
fix(hosting): bound hosted set by module-cache capacity to stop recompile thrash

sanity:
## Problem

Production gateways host ~2600 contracts, but the compiled-WASM module cache only keeps ~1024 modules hot, so they thrash: constant recompilation, ~116% CPU.

Root cause is a two-budget mismatch. The hosting budget (`crates/core/src/ring/hosting/cache.rs`) and the module cache (`crates/core/src/wasm_runtime/module_cache.rs`) both size off `total_ram/8`, but they measure different-cost resources under different ceilings:

- hosting budget: contract **state** bytes (~380 KiB/contract), clamped to 1 GiB
- module cache: **compiled module** bytes (~1.5 MiB/module), clamped to 1.5 GiB

So the state-byte budget permits ~2.5x as many contracts as the module cache can keep hot. Every hosted-but-not-cached contract recompiles on access.

## Approach

Bound the hosted-contract **COUNT** by the module-cache capacity, so hosting never permits more contracts than the module cache can keep hot.

- New `REPRESENTATIVE_COMPILED_MODULE_BYTES` (1.5 MiB, matching `module_cache`'s `MEASURED_MODULE_SIZE`, #4452) and a pure helper `max_hosted_for_module_budget(budget) = (budget / 1.5 MiB).max(1)`.
- `HostingCache` gains a `max_hosted_contracts` field (defaults to `u64::MAX` — behaviour-preserving until set). Eviction's `over_budget()` now trips on **either** resource: state bytes past `budget_bytes`, OR contract count past `max_hosted_contracts`.
- Wired once at `Ring` construction from the **resolved** config (`config.config.module_cache_budget_bytes`), so an operator `--module-cache-budget-bytes` override is honored.
- Reuses the existing subscriber-primary `victim_order` unchanged, so zero-subscriber every-hop copies shed first (subscriptions protected; see hosting-invariants).
- A `MAX_EVICTIONS_PER_PASS` drain cap (128) spreads the one-time shed — when the ceiling first binds on a node already hosting thousands of contracts — across successive 60s sweeps instead of one inline cliff.

On a large gateway this intentionally drops the hosted set from ~2600 to ~1024 (matching the module-cache hot-set), eliminating the recompile thrash.

Behaviour-preserving until wired: the ceiling defaults to `u64::MAX`, so existing tests and any node that never sets it are unaffected. No wire-format change.

## Testing

New unit tests in `cache.rs`:

- `max_hosted_for_module_budget_maps_module_capacity_to_count` — pure math: 1.5 GiB → 1024, 64 MiB → 42, 768 MiB → 512, 0 → 1.
- `count_ceiling_sheds_to_max_hosted_keeping_most_recent` — with a byte budget that never binds, inserting 5 zero-subscriber contracts under a ceiling of 2 drains to 2 residents, keeping the 2 most-recently-accessed (via `victim_order`'s recency tiebreak).
- `count_ceiling_drain_is_capped_per_pass` — with `MAX_EVICTIONS_PER_PASS + 3` residents and the ceiling bound to 1, one sweep evicts at most `MAX_EVICTIONS_PER_PASS`, leaving >1; a second sweep finishes the drain to 1.

`cargo test -p freenet --lib ring::hosting::cache::tests`: 55 passed, 0 failed. `cargo clippy -p freenet --lib -- -D warnings`: clean.

Reference epic #4642.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -548,22 +548,28 @@ impl Ring {
             governance_config,
             time_source.clone(),
         ));
+        // Production passes the Ring's default `Arc<InstantTimeSrc>` (wall clock).
+        // Simulation tests can inject a controllable clock via
+        // `NodeConfig::hosting_time_source_override` so hosting-cache TTL /
+        // eviction is deterministic (#4642 piece A). Same `Arc` clone as the rest
+        // of the Ring when no override is set.
+        let hosting_manager = hosting::HostingManager::with_time_source(
+            config.config.max_hosting_storage,
+            config
+                .hosting_time_source_override
+                .clone()
+                .unwrap_or_else(|| time_source.clone()),
+        );
+        // Bound the hosted-contract COUNT by what the compiled-module cache can
+        // keep hot, so hosting never permits more contracts than the module cache
+        // has hot slots (anti-recompile-thrash, #4642). Reads the RESOLVED config
+        // so an operator `--module-cache-budget-bytes` override is honored.
+        hosting_manager.set_hosted_count_ceiling(config.config.module_cache_budget_bytes as u64);
         let ring = Ring {
             max_hops_to_live,
             router,
             connection_manager,
-            // Production passes the Ring's default `Arc<InstantTimeSrc>`
-            // (wall clock). Simulation tests can inject a controllable clock via
-            // `NodeConfig::hosting_time_source_override` so hosting-cache TTL /
-            // eviction is deterministic (#4642 piece A). Same `Arc` clone as the
-            // rest of the Ring when no override is set.
-            hosting_manager: hosting::HostingManager::with_time_source(
-                config.config.max_hosting_storage,
-                config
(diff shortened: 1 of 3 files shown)


sanity:
Superseded by #4794 (the option-(c) fix). 

Root-cause drill-down confirmed the gateway module-cache thrash is driven by the 5-min InterestSync heartbeat re-running WASM `summarize` for every hosted contract, because the `summary_cache` that should skip that WASM for unchanged state is capped at 1024 entries vs ~2600 hosted. #4794 sizes that cache to the hosted set (8192 entries on a gateway), stopping the recompile thrash **without reducing hosting count** — so it avoids the findability tradeoff this count-ceiling approach (2600→1024) carried, which the multi-model review flagged as a network-wide risk shipped ahead of the epic's required findability gate. Maintainer chose option (c). Closing this in favor of #4794.

[AI-assisted - Claude]

Issue #3836, opened 2026-04-13
feat: show delegate permission prompts as in-page overlay instead of browser notifications


sanity:
## Problem

The current permission prompt system (#3818) uses browser Notifications to alert the user when a delegate needs permission. This is fragile:

1. **Users may have notifications blocked** -- many users deny notification permissions by default
2. **Notifications are easily dismissed** -- a swipe or click dismisses them permanently, with no way to recover the prompt
3. **No guaranteed visibility** -- notifications can be missed if the user is focused on another monitor, in DND mode, or on mobile
4. **Requires a separate page visit** -- user must click the notification, then navigate to `/permission/{nonce}` to respond

Since permission prompts are a critical flow (blocking delegate operations until the user responds), relying on a mechanism the user can miss or block is risky.

## Proposed Solution

The shell page JS (outside the sandboxed iframe) already polls `GET /permission/pending`. Instead of creating a `new Notification(...)`, it should render an **overlay directly on the page**.

### How it would work

1. Shell page detects a pending permission prompt via its existing polling
2. Shell page injects an overlay div on top of the iframe (in the shell's DOM, not the iframe's -- so the sandboxed app can't interfere)
3. Overlay shows the prompt text and Allow Once / Always Allow / Deny buttons
4. User responds directly in the page they're already looking at
5. Shell page POSTs the response to `/permission/{nonce}` (same as current flow)

### Why every open tab

Every open Freenet tab runs the shell page with the same polling. All of them would show the overlay, including the tab whose app triggered the request. This means:

- The user sees the prompt on whatever Freenet page they're currently looking at
- No notification permission required
- Can't be accidentally dismissed (overlay stays until responded to)
- Richer UI possible (show app context, ghostkey details, etc.)

### Security

The overlay renders in the shell page's DOM (trusted, same-origin with the gateway). The sandboxed iframe cannot access or modify it. This is the same trust boundary as the current notification approach.

### Backward compatibility

Browser notifications could be kept as a fallback for cases where no Freenet tab is open (though this is unlikely if the request was triggered by a web app). Or they could be removed entirely since the prompt only makes sense when an app is active.

[AI-assisted - Claude]

PR #4801, opened 2026-07-15
feat(server): proxy browser notifications from sandboxed contract apps

sanity:
## Problem

Contract web apps (River, etc.) run in a sandboxed iframe with an **opaque (null) origin** — added in #3254 for origin isolation. Browsers block the Web Notifications API from opaque-origin iframes, so a contract app can never get notification permission and notifications silently never fire in the gateway build. This is what broke River's new-message notifications (freenet/river#408): they worked as a top-level page (Dec 2025 → Feb 2026), then #3254 moved River into the sandboxed iframe and killed them.

## Approach

Add a **notification proxy** to the shell bridge, mirroring the existing `clipboard` and `download` proxies. The app (opaque origin) can't call the Notifications API; the shell (same-origin with the node, a real origin) can — so the app postMessages the shell, which shows the notification on its behalf.

Two new `__freenet_shell__` message types from the iframe:
- **`notification_enable_prompt`** → shows a small, clearly host-owned in-shell affordance ("Get notified of new messages? [Enable] [Not now]"). The actual `Notification.requestPermission()` fires from a **real click in the shell frame** — a gesture inside the sandboxed iframe doesn't reliably grant the shell the transient activation browsers require for the permission prompt.
- **`notification`** `{title, body?, tag?}` → shows a browser notification, then postMessages `notification_click` back to the iframe so the app can focus the window and route to the right room.

## Permission scope (the important design point)

Notifications permission is **per-origin**, and every contract on a gateway shares the shell's origin (they differ only by path), so a browser grant is **gateway-wide** — there's no per-contract browser scoping without per-contract origins (subdomains), which the gateway doesn't do.

To keep it **effectively per-contract**, the shell additionally requires a stored **per-contract consent flag** (`localStorage['freenet_notify:<key>']`, keyed off the trusted server-routed path `/v1/contract/web/<KEY>/`, **never** message content) before ever showing a notification. So one contract's grant never lets a *different* contract on the same gateway notify the user without its own explicit opt-in.

## Safety

Same posture as the existing clipboard/download proxies:
- Notification text is attacker-controlled (message content from other room members) → **text-only** (Notification renders no markup), **length-capped** (title 128 / body 256).
- **Rate-limited** (1/s backstop) and **tag-coalesced per contract** so a busy room replaces rather than stacks.
- Only the app iframe can trigger it (existing `event.source !== iframe.contentWindow` gate).
- Consent key derives from the routed path, not message content, so a contract can't claim another's consent.

## Testing

- Prettier + ESLint clean (`npm run lint` in `crates/core/src/server`).
- JS validated with `node --check`.
- Manual verification requires the paired River change (freenet/river#408) plus a gateway serving this shell. Until then this is inert (no contract sends the messages), and current River is unaffected (it still calls the Notification API directly, which no-ops in the iframe as before).

## Deployment note

This is the **enabling half** — a no-op until a contract app (River) is updated to send the new messages, and neither side's deployment breaks the other. The try.freenet.org gateway is a manual from-source rebuild, so it needs a redeploy after this merges for the feature to light up there.

Refs freenet/river#408


sanity:
## Multi-model review — Full tier (gateway trust-boundary / security-sensitive)

Ran **Codex** + two adversarial Claude lenses (security/abuse, correctness/browser-compat), each blind to the others.

**Verdict:** no P1/P2 trust-boundary breaks. The security lens confirmed cross-contract consent isolation and consent-gating are sound and fail closed (consent key derives from the trusted routed path, not message content; null-key fails closed; gateway-wide browser grant does not leak to a new contract without its own opt-in; no exfiltration — icon omitted, click echoes only the app's own tag; text-only + length-capped). Posture matches the existing clipboard/download proxies.

### Fixed in the latest commit
| Finding | Reviewers | Fix |
|---|---|---|
| Consent-key regex was `v1`-only → notifications silently dead on `/v2/contract/web/` loads (permission granted but every notification dropped) | Codex P2 + correctness F1 P2 | Match `/v[12]/`, consistent with `CONTRACT_PREFIX_RE` |
| "Not now" dismissal not durable — a contract re-sending `notification_enable_prompt` re-pins the host bar | Codex P2 + security P3 | 24h persisted per-contract snooze + in-session flag |
| localStorage-write failure reports `granted` while delivery is impossible | Codex P2 + correctness F5 | In-memory consent fallback; only report `granted` when consent is actually recorded |
| `requestPermission` double-invocation (callback + promise) and rejection/throw stranding the affordance | correctness F2/F3 P3 | `called` latch + `.then(done, done)` + `done('default')` on throw |
| Global 1/s rate limit dropped legitimate cross-room notifications; distinct-tag flood bypassed coalescing | correctness F6 + security tag-stacking P3 | Per-tag throttle (3s) + rolling global cap (20/60s) + bounded per-tag map |
| Mobile Chrome `new Notification()` throws → silent no-op after reporting granted | correctness F4 | Post `notification_status: undeliverable` |

### Documented, deliberately not fixed (justified)
- **Notification content attribution / phishing** (security P3): title/body are attacker-controlled with no host-injected authorship. This matches the standard web-notification trust model (any origin you grant notifications to shows arbitrary text), it's **consent-gated** (the user explicitly enabled *this* contract), and the notification body is **not clickable to a URL** (click only focuses the tab and echoes the app's own tag). Adding a host attribution prefix would clutter every legitimate notification and the shell only knows the contract key, not a friendly name. Acceptable as-is.
- **Mobile/native reliability** (correctness F4): non-persistent notifications don't work on Android Chrome; the real fix is a ServiceWorker `showNotification` path, out of scope here. The `undeliverable` signal above lets the app fall back to its in-app unread badge.

[AI-assisted - Claude]

sanity:
Filed freenet-core#4807 as the repo-local tracking issue for this feature (CONTRIBUTING.md), cross-linked to freenet/river#408. This is maintainer-authorized work (the shell-side half of the River notification fix).

sanity:
## Pre-merge re-review — Full tier (multi-model), HEAD `f602288d`

The prior review on this PR was **stale**: it ran on `76236213`, but ~118 LOC of production code landed after it (rate-limiter refactor + #3836 guard re-scoping). Re-ran a full multi-model review on current HEAD — external **Codex** pass (Gemini remains unavailable, `IneligibleTierError`) + two adversarial Claude lenses (security/abuse, correctness/browser-compat), each blind to the others.

**Verdict: trust boundary holds — merging.** All three independently confirmed: consent **fails closed** (requires browser `granted` **and** per-contract consent, both set only from a real shell-frame click); **cross-contract isolation** intact (consent/snooze/tag keys derive from the server-routed `location.pathname`, never message content); **no injection** (title/body are text-only into the Notification API and length-capped; the `icon` field isn't proxied at all); `postMessage` dispatch gated on `event.source === iframe.contentWindow`; rate-limited post-consent. The rate-limiter refactor and #3836 guard re-scoping are behavior-preserving and do **not** widen the production attack surface.

### One bounded finding — consciously accepted for this release (tracked in #4849)

**F1 (~P2, defense-in-depth):** the notification flood-cap (global 20 / 60s) is a per-document `var` and **resets on a full page reload**. A contract the user has *already consented to* can send 20 notifications, then `postMessage` a same-key v1↔v2 `navigate` (treated as cross-contract → full reload → fresh empty limiter; consent survives) for a repeated fresh flood budget.

**Accepting for this release**, because it's bounded and breaks no boundary: only an already-consented app; each cycle costs a visible 1–3s tab reload-thrash (self-limiting); tags coalesce. It's a ~10–30× weakening of a brand-new anti-flood control, not a regression of existing behavior. Codex's own call was "not a clean merge-blocker." Fix deferred to **#4849**, to be picked up immediately after the release.

### Follow-up hardening (also #4849)

- **F2 (P3)** — the #3836 guard slice ends before the SSE `prompt_added` handlers; production is Notification-free there today, but it reduces future regression coverage. *(Independently flagged by both Claude lenses and Codex.)*
- **F3 (P3)** — the eviction unit test doesn't actually exercise eviction (call spacing empties the window each time).
- Minor notes (snooze-key re-nag; inherent Web-Notifications title spoofing) captured in #4849 too.

Applying `review-override` for the advisory `rule-review/warnings` check and enqueuing.

[AI-assisted - Claude]

PR #4802, opened 2026-07-15
fix(hosting): size summarize cache to live hosted count to stop module thrash

sanity:
## Problem

Gateways thrash the compiled-WASM module cache. The ~5-min InterestSync
heartbeat runs WASM `summarize` for every hosted contract, but the
per-executor `summary_cache` / `delta_cache` (`lru::LruCache`) were seeded
at a flat **1024** entries. On a gateway hosting ~2600 contracts, ~1576
keys per cycle miss the cache and recompile a cold module every cycle.

The flat 1024 was a magic number unrelated to the working set it must
cover, so coverage silently degraded as the hosted set grew.

## Approach

Tie the cache size to the **real** working set rather than a guess.
Before a summarize/delta computation, `ensure_cache_covers_hosted_set`
grows the caches to the node's live hosted-contract count
(`Ring::hosting_contracts_count()`, a node-tracked metric) plus a small
mid-cycle margin, rounded up to a power of two and clamped to
`[SUMMARY_CACHE_MIN, SUMMARY_CACHE_MAX]`.

- **Guaranteed coverage for any count** — not for one assumed count, and
  not via a contract-size/byte-budget proxy (which was PR #4794's earlier
  `clamp(budget / 128 KiB, ...)` divisor approach; this count-tied resize
  replaces it).
- **O(1) once warmed** — a count read + a compare; the resize only fires
  when the hosted set actually grew past the current capacity.
- **Lazy memory** — `lru::resize` grows with no preallocation, so resident
  memory tracks the actual entry count, not the capacity.
- **`SUMMARY_CACHE_MIN` (1024)** keeps small/mock nodes at the historical
  size. **`SUMMARY_CACHE_MAX` (65536)** bounds worst-case per-worker
  LRU-node overhead for pathologically tiny contracts (a byte budget can
  hold far more small contracts than large ones); `.next_power_of_two()`
  can round above MAX, hence the trailing `.min(MAX)`.

**Delta-cache caveat:** the delta cache key includes a per-peer summary
hash, so during fan-out it can hold more than one entry per contract.
Sizing it to the hosted count is a strict improvement and removes the
size assumption, but is count-scaled best-effort rather than a
per-subscriber guarantee. Summary-cache coverage is exact.

Built off current `origin/main` (flat `LruCache::new(1024)`), not #4794's
branch.

## Testing

New test `summary_cache_covers_live_hosted_count_not_fixed_cap` proves
coverage is driven by the **count**, not a threshold. For `N` in
`{1024, 3000, 5000}` it stores N distinct contracts+states against an
**uncached** state store and hosts each (driving
`hosting_contracts_count() == N`), then summarizes the whole hosted set
twice. The second pass must reload state **zero** times — observable via
`MockStateStorage::get_count` — which only holds if the cache grew to
cover every hosted contract. Under the old fixed 1024 cap the 3000/5000
cases would evict the earliest keys and reload (recompiling a cold
module) on pass 2.

`cargo test -p freenet --lib contract::executor::pool_tests::summarize_delta_cache_tests`
→ 7 passed / 0 failed. `cargo clippy -p freenet --lib -- -D warnings` clean.

Refs #4642 (demand-driven hosting epic).

[AI-assisted - Claude]

https://claude.ai/code/session_0199hK8Vdn5DAWcXJsxZHk8H

### DIFF
--- a/crates/core/src/contract/executor.rs
+++ b/crates/core/src/contract/executor.rs
@@ -971,6 +971,16 @@ pub struct Executor<R = Runtime, S: StateStorage = Storage> {
     delegate_notification_tx: Option<DelegateNotificationSender>,
 }
 
+/// Summary/delta cache is sized to the live hosted-contract count (see
+/// `ensure_cache_covers_hosted_set`) so the ~5-min interest-heartbeat summarize
+/// never recompiles a cold module — coverage tied to the real count, not a
+/// contract-size assumption. MIN keeps small/mock nodes at the historical size;
+/// MAX bounds worst-case per-worker LRU-node overhead for pathologically tiny
+/// contracts (a byte-budget can hold far more small contracts than large ones).
+pub(crate) const SUMMARY_CACHE_MIN: usize = 1024;
+pub(crate) const SUMMARY_CACHE_MARGIN: usize = 256; // contracts hosted mid-cycle
+pub(crate) const SUMMARY_CACHE_MAX: usize = 65_536;
+
 impl<R, S> Executor<R, S>
 where
     S: StateStorage + Send + Sync + 'static,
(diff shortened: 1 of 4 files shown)


sanity:
Superseded by #4804.

Multi-model review of #4802 (Codex P1 + a memory-lens finding) showed the count-only cache resize had no byte bound — and StateSummary/StateDelta are contract-controlled, variable-sized, so a contract emitting large summaries/deltas cached at up to tens of thousands of entries × pool workers could OOM the node (#4565 class + amplification-vector rule). #4804 keeps the count-target-tied-to-live-hosted-count coverage guarantee AND adds a byte backstop (a ByteBoundedLruCache wrapper), so it's memory-safe by construction while still fixing the thrash. Closing in favor of #4804.

[AI-assisted - Claude]

PR #4806, opened 2026-07-15
test: measure GET reliability against a formed ring, not during bootstrap

sanity:
## Problem

The nightly `test_get_reliability_diagnostic` (100-node deterministic sim) has failed **every night since 2026-06-28**: it asserts ≥80/100 nodes "dispatch" a GET, but the number is a deterministic **42/100**.

Root cause — two correct-in-isolation PRs interacting:

- **#4599** set the `dispatched_nodes ≥ 80` floor, calibrated against a run scoring 86 — but that run had placement-migration (`SubscribeHint`) **accidentally ON**.
- **#4601** then correctly disabled migration in sims. Migration is a deliberately-disabled anti-pattern; real findability comes from immediate-neighbour connectivity (the nearest-neighbour ring lattice, #4760), not migration. With it off, the metric fell to 42/100. So the floor was silently calibrated against a disabled anti-pattern.

The 42 is a **bootstrap-timing artifact, not a findability defect.** The controlled-event client fires each node's *one-shot, no-retry* GET roughly sequentially (~one per virtual-time slice), so higher-index nodes' GETs fire *before those nodes finish joining* the ring and are rejected with `PeerNotJoined` before any routing happens. Every GET that *did* run succeeded (100%), and all 100 nodes join — just late. The test was measuring cold-start join speed, not GET reliability.

## Approach

Form the ring, **then** measure — and measure the right thing.

- **`SimNetwork::wait_for_join_convergence_before_ops(fraction, max_wait)`** (opt-in): the controlled client waits until `fraction` of peers have joined before firing any operation. "Joined" = `≥ 1` established connection, exactly the `peer_ready` bar `ensure_peer_ready` checks before letting a peer originate an operation. Join is read from the topology-snapshot registry, so `TopologySnapshot` now carries the peer's `connection_count` — snapshot *presence* alone only means the bind address is set (early), which is **not** a join signal.
- **`SimNetwork::with_controlled_op_interval(interval)`** (opt-in): shrinks the fixed 3s per-op settle. Against a formed ring a GET finishes in ~1ms, so the default would spend 100×3s = 300s of virtual time waiting for nothing.
- **Success metric → retrievability.** The old `dispatched_nodes` floor counts only *network* GETs, so it is confounded by caching: a node whose GET finds the contract already cached locally resolves without a network request, so **better placement drives the count down**. Switched the gate to `nodes_with_state` (did each node actually end up holding the contract — the honest, non-gameable measure, and the reliability property that actually matters). `dispatched_nodes` is kept as a logged diagnostic.

Both harness knobs **default to the historical behaviour**, so no other test changes.

Why not "retry the rejected GETs"? CONNECT runs on its own maintenance loop independent of GET/PUT traffic, so retries would not speed ring formation and would add hosting load; forming the ring first is both cleaner and faithful to how a node behaves in production (it joins under load, then serves).

## Testing

Verified locally on this scenario (deterministic, seed `0x3570_D1A6_0001`):

| | before (main) | after |
|---|---|---|
| retrievability (nodes with contract) | 51/100 | **97/100** |
| GET success rate | 100% | 100% |
| test outcome | FAIL | **PASS** |

Floor set at **90/100** (measured 97, ~7-node margin; deterministic so no jitter). The only 3 nodes that miss are `[0,1,2]` — the very first GETs, which race the initial PUT's propagation.

Note: this test only runs in the **nightly** (gated by the `nightly_tests` feature), so regular PR CI does not exercise it. The nightly has been triggered on this branch to verify end-to-end.

Closes the recurring nightly failure (tracked via the CI-bot alerts on commits `d3cda891` / `957dcbd7`).

## Second commit: de-flake `test_placement_migration_at_scale_renewal_load_stays_bounded`

Verifying the fix above via a full nightly on this branch surfaced a *separate*, pre-existing flaky test (the reliability test itself passed). `test_placement_migration_at_scale` (#4601) asserted its "cap engaged" guard as a **per-seed** exact `assert_eq!(max_cycle_batch == 10)` across its 3 seeds. Whether the renewal burst clips a single 30s cycle to *exactly* the cap on a given seed is sensitive to the turmoil runner's residual (~1%) non-determinism — observed on seed `0x46010002`: `7` on one run, `10` on the next; other seeds and re-runs `10`. My change is not the cause (the migration test passes identically on this branch and on plain `main` when re-run).

Fix: keep the safety guard **(A)** `max_cycle_batch <= cap` per-seed (unchanged); move the non-vacuousness guard **(B)** "cap engaged" to **across seeds** — record each seed's peak, assert `max(peaks) == cap` (the cap must engage on at least one seed). Since (A) bounds every peak `<= cap`, `max == cap` is exactly "at least one seed reached the cap", so the guarantee is preserved without the per-seed brittleness. Matches the test's own "assert across several seeds, not just one" intent.

[AI-assisted - Claude]

https://claude.ai/code/session_01N2RjLVoAfu6ZcNpkTy2UhH

### DIFF
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -1296,6 +1296,38 @@ pub struct SimNetwork {
     pub skip_convergence_wait: bool,
     /// Optional churn (crash/restart) configuration for the chaos driver.
     churn_config: Option<ChurnConfig>,
+    /// Optional pre-operation join-convergence barrier for
+    /// [`run_controlled_simulation`](Self::run_controlled_simulation). When
+    /// `Some((min_fraction, max_wait))`, the controlled-event client waits —
+    /// before firing any scheduled operation — until at least `min_fraction`
+    /// of the network's peers have completed their network join, or `max_wait`
+    /// virtual time elapses, whichever comes first. Join is detected via the
+    /// topology-snapshot registry: a peer only registers a snapshot once its
+    /// own address/location is established (`peer_ready`), so the count of
+    /// distinct snapshot peers is a direct join tally (see
+    /// `ring::register_topology_snapshots_periodically`, which `continue`s
+    /// while `get_own_addr()` is `None`).
+    ///
+    /// `None` (default) preserves the historical behavior of firing operations
+    /// after a fixed 3s warmup, racing topology formation. This barrier exists
+    /// because a cold-start GET-reliability measurement must run against a
+    /// FORMED network: with only the fixed warmup, higher-index nodes' one-shot,
+    /// no-retry GETs fire before those nodes finish joining and are rejected
+    /// with `PeerNotJoined`, so the metric conflates join speed with GET
+    /// reliability. See `test_get_reliability_diagnostic`. Opt in via
+    /// [`wait_for_join_convergence_before_ops`](Self::wait_for_join_convergence_before_ops).
+    wait_for_join_before_ops: Option<(f64, Duration)>,
+    /// Optional override for the delay the controlled-event client waits after
+    /// triggering each scheduled operation before triggering the next one.
+    /// `None` (default) uses the historical fixed 3s settle. A test whose
+    /// operations complete quickly against an already-formed network (e.g. a
+    /// read-only GET sweep behind `wait_for_join_convergence_before_ops`) can
+    /// shrink this to avoid spending ~3s of virtual time — and the wall-clock
+    /// to simulate it — per operation for work that finishes in milliseconds.
+    /// Only affects regular client-request operations; special in-client ops
+    /// (clock advance, crash/recover) keep their own fixed settle. See
+    /// [`with_controlled_op_interval`](Self::with_controlled_op_interval).
+    controlled_op_interval: Option<Duration>,
     /// Optional governance-manager config override applied to every node
     /// this network builds. Lets governance sim tests compress the
     /// production minute-to-hour timescales and lower `min_samples` so the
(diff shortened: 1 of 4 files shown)


sanity:
## Multi-model review summary (Full tier)

Ran a full-tier review — three Claude reviewers (skeptical bug-hunt, invariants/big-picture, test-quality) plus an external **Codex** pass — on the get-reliability change, and a focused Claude + Codex pass on the migration de-flake. No blocking issues; all findings addressed:

| finding | source | resolution |
|---|---|---|
| `#4760` ring lattice credited for findability, but it's gated off below `max_connections=25` (this test is 12) | big-picture | corrected comments → small-world routing + gateway fallback |
| "deterministic, no jitter" overstated (turmoil is ~99% det) | test-quality | reworded; margin now justified by turmoil residual non-determinism |
| stale function docstring ("logs rather than asserts") | test-quality | rewritten to the three hard floors |
| "exactly the peer_ready bar" (it's *stronger*) | skeptical | corrected — the barrier can only over-wait, never under-wait |
| retrievability could be false-green via relay caching | **Codex [P2]** | raised the weak `network_successes >= 5` floor → **30** (measured 63–68) to assert broad GET routing |
| `min_fraction` unclamped | skeptical | added `.clamp(0.0, 1.0)` |
| **GET/storage loops used `0..NUM_NODES`, but regular-node labels are offset by `NUM_GATEWAYS`** (node-3..node-102) — 3 phantom labels counted "missing", 3 real nodes untested | **Codex [P2]** | fixed both loops to iterate the real range (commit `6fbff3d2`); retrievability is now a true **100/100** (was a phantom-biased 97/100) |
| migration test flaked (a *separate*, pre-existing test) | branch nightly | de-flaked `test_placement_migration_at_scale` — cap-engaged check moved from per-seed exact-equality to across-seeds (commit `bc458502`) |

**Verification** (deterministic, seed `0x3570_D1A6_0001`): `test_get_reliability_diagnostic` passes with **100/100 retrievability, 100% GET success, 63 network-traversed**. `test_placement_migration_at_scale` passes reliably with the across-seeds assertion. Regular PR CI green. Nightly re-triggered on the final commit.

[AI-assisted - Claude]

PR #4809, opened 2026-07-15
fix(router): stop periodic refresh discarding relay-learned events

### LINKED ISSUE #4808: fix(router): periodic refresh discards relay-learned events, resetting the routing model faster than it learns
## Problem

Every peer's routing model is reset faster than it can learn. `prediction_active` is false on ~85% of the network, the per-peer routing charts on the node dashboard sit empty, and the transfer-rate estimator on a typical peer has never held an observation.

The router is **not** starving for input — it learns at a healthy median of ~14-16 routing events/hour. It is being wiped.

## Root cause

`Ring::refresh_router` (`crates/core/src/ring.rs:1589-1591`) rebuilds the entire router from the on-disk event log every 5 minutes:

```rust
if !history.is_empty() {
    *router.write() = Router::new(&history);
}
```

But `operations::record_relay_route_event` (`crates/core/src/operations.rs:835-844`) feeds the **in-memory router only** — it never calls `NetEventLog::route_event` / `register_events`, so it never persists:

```rust
op_manager.ring.router.write().add_event(crate::router::RouteEvent { ... });
```

Only the 4 **originator** sites persist (`get/op_ctx_task.rs:526-533`, `put/op_ctx_task.rs:496-510`, `subscribe/op_ctx_task.rs:1045-1065`). The ~13 **forwarding-hop** sites do not.

So every forwarding-hop route event has a **<=5 minute half-life**. On a peer carrying other people's traffic — i.e. most peers, and all the ones doing the most routing — forwarding-hop events *are* the model, so each rebuild resets it to the originator-only remnant.

Within a session the on-disk log is a strict **subset** of what the live router already knows: the same originator events, minus every forwarding-hop event. Reloading it can only ever lose information.

## Evidence

**Live, on a production peer** (`Global events`, polled every 30s):

```
10:32:11 - 10:37:12   29   (stable)
10:37:42               2   <- 93% of the model destroyed
10:38:12 - 10:39:42    2
10:40:12               4   <- re-climbing
10:42:42              14
```

**Fleet-wide** (118,080 `router_snapshot` rows, 43h window, public-IP peers, no 0.1.x present):

- **261,710 routing events gained, 271,518 lost to model resets (103.7%)** — net accumulation zero.
- 16.7% of snapshot-to-snapshot transitions lose events; median drop 5, **max 495** (a saturated router rebuilt from near-nothing).
- 0.68 drops/peer/hour — far too frequent for restart churn.
- Median peer holds 5-7 events against `MIN_EVENTS_FOR_PREDICTION = 50` (`router.rs:1457`); p90 is 57-102; only 11-16% of peers ever latch prediction.
- At the median learning rate a peer needs ~3.5h to reach the gate, but resets land roughly every ~1.5h.

**Code-level A/B** confirming the recording path itself is healthy: 6 GETs for contracts the peer did not hold moved `Global events` 2 -> 14 and produced that peer's first-ever transfer-speed observations; 4 GETs for contracts it did hold moved it 11 -> 11 (served locally, correctly no routing decision, no event).

## Consequence

Below the 50-event gate the router uses the `#3398` untried-first fallback (`router.rs:1092-1115`) rather than plain distance routing. That fallback was designed as a **transient bootstrap** on the way to 50 events. Because the gate is now unreachable, the network sits in it permanently — deliberately deprioritising the peers it has successfully used. Peers at 6-12 events also straddle the `ADJUSTMENT_PRIOR_SIZE = 10` boundary, flipping between two routing regimes.

## Why the rebuild exists (and should not simply be deleted)

The rebuild is a genuine **accuracy** mechanism, not a vestigial restore path. `pav_regression`'s incremental path is approximate:

- `add_points` re-runs `isotonic()` over `self.points` — the already-**pooled** blocks, not the raw inputs — so the fit depends on insertion order.
- `remove_points` is explicitly approximate: it subtracts an evicted point's influence from the **closest aggregate by x**, which need not be the aggregate that point actually contributed to. This only begins once the window saturates at `MAX_REGRESSION_POINTS = 500` — the steady state of a busy peer.

So the fix must preserve batch refitting, not remove it.

## Fix

Refit `global_regression` from `IsotonicEstimator.raw_points` — a FIFO of every observation in the current window, populated by `add_event` regardless of origin, and therefore the complete corpus, in memory, by construction. Drive it by **data turnover (>10% of the window)** rather than a 5-minute timer, so an idle router does no work.

`peer_adjustments` needs no refit: it is an EWMA, inherently incremental, with no batch-vs-online discrepancy.

## Out of scope (follow-up)

The **startup** restore is a separate concern and separately broken: `aof.rs:656` filters route records to `record_ts >= new_records_ts`, where `NEW_RECORDS_TS` (`register.rs:1705`) is stamped in `EventRegister::new` at `node.rs:680` — strictly before `Ring::new` spawns `refresh_router`. Every prior-session record is therefore dropped and the restore has never executed successfully since #3672 (2026-03-27). It also records a health-gauge *success* while restoring nothing, so the #4440 gauge reads healthy. Tracked separately.

[AI-assisted - Claude]

sanity:
## Problem

Every peer's routing model is reset faster than it can learn. `prediction_active` is false on ~85% of the network, the per-peer routing charts sit empty, and the transfer-rate estimator on a typical peer has never held an observation.

The router is **not** starving for input — it learns at a healthy median of ~14-16 events/hour. It is being wiped.

`Ring::refresh_router` rebuilt the whole router from the on-disk event log every 5 minutes:

```rust
if !history.is_empty() { *router.write() = Router::new(&history); }
```

But `operations::record_relay_route_event` (`operations.rs:835-844`) feeds the **in-memory router only** — it never persists. So every route event from the ~17 forwarding-hop sites had a **<=5 minute half-life**, while only the 4 originator sites (which do call `NetEventLog::route_event`) survived a rebuild.

On a peer carrying other people's traffic — i.e. most peers, and precisely the ones doing the most routing — forwarding-hop events **are** the model. The rebuild reset it to the originator-only remnant.

The log's originator events are a subset of what the router has already been fed, so reloading it could only lose information: it would *add stale* originator events (older than the rolling window — exactly what the window exists to discard) while *dropping fresh* forwarding-hop events.

### Evidence

Live, on a production peer (`Global events`, 30s polling):

```
10:32:11 - 10:37:12   29   (stable)
10:37:42               2   <- 93% of the model destroyed
10:40:12               4   <- re-climbing
10:42:42              14
```

Fleet-wide (118,080 `router_snapshot` rows, 43h, public-IP peers):

- **261,710 events gained, 271,518 lost to resets (103.7%)** — net accumulation zero.
- 16.7% of transitions lose events; max single drop **495**.
- Median peer holds 5-7 events against `MIN_EVENTS_FOR_PREDICTION = 50`; only 11-16% ever latch prediction.

Below that gate the router uses #3398's untried-first fallback, which was designed as a *transient* bootstrap toward 50 events. Because the gate is unreachable, the network sits in it permanently.

## Approach

Keep the batch refit — it is a real **accuracy** mechanism, not a vestigial restore path — but give it the right corpus and the right trigger.

`IsotonicEstimator` already retains a FIFO of every observation in the window, populated by `add_event` regardless of origin. That is the complete corpus, in memory, by construction. Refit from it instead of rebuilding from disk.

The refit is genuinely necessary — `pav_regression`'s incremental path is approximate, and review measured the resulting drift at **3-6% of span (response time), 21% (failure estimator)** on realistic data:

- `add_points` re-runs `isotonic()` over `self.points`, the already-**pooled** blocks rather than raw inputs, so the fit depends on insertion order.
- `remove_points` is explicitly approximate: it subtracts an evicted point's influence from the **closest aggregate by x**, which need not be the aggregate it contributed to. Only begins once the window saturates — the steady state of a busy peer.

Refitting is driven by **data turnover (>10% of the window)** rather than a timer, so an idle router does no work. (Realised cadence is `min(poll interval, turnover)`; the 5-min tick only binds above ~10 events/min, ~40× the production median — documented on `REFIT_STALENESS_PERCENT`.)

**A refit rebuilds BOTH halves** — the global curve *and* every peer's adjustment, via logic shared with the constructor. They cannot be refit independently: an `Adjustment` is an EWMA of residuals measured *against* the global curve, so moving the curve alone leaves every peer corrected relative to one that no longer exists (measured at up to 0.198 absolute on a [0,1] failure probability). Rebuilding from the window also bounds `peer_adjustments` to peers currently in it — without which it would grow one entry per peer ever seen, an unbounded per-key collection driven by remote peers.

Measured cost: a saturated 500-point refit is **27.3µs**; all 15 estimators worst-case **~410µs**, once per 5 min — strictly cheaper than the old `Router::new` over up to 10,000 events under the same lock.

## Testing

**The regression guard is `refresh_router_never_reloads_from_disk_after_startup`** — it exercises the pre-existing `refresh_router` and genuinely fails against the old loop (which read the log every tick, so `total_calls` would be ~9, not 1). It supersedes `refresh_router_survives_transient_get_router_events_error`; see the review comment for why deletion beat `#[ignore]` there.

**`refresh_router_loop_refits_stale_estimators`** pins the loop→refit wiring — the one line that makes this fix do anything. Verified by mutation: stubbing the call fails only this test; the other five guards stay green.

**`refit_rebuilds_peer_adjustments_against_the_new_curve`** asserts a refit equals a batch build in `estimate_retrieval_time` — the value routing actually consumes, not merely the global curve. Plus `refit_prunes_peers_that_fell_out_of_the_window`.

**`incremental_fit_drifts_from_batch_and_refit_repairs_it`** proves the refit is *necessary* (drift exists) and *sufficient* (restores the exact batch fit). Doubles as a tripwire: if `pav_regression` ever becomes exact, it fails and `refit_if_stale` can be deleted.

**`refit_respects_descending_estimator_direction`** — rewritten after review proved the original vacuous (an ascending fit over decreasing data pools to a flat line, satisfying `near >= far` by equality). Now asserts a strict decrease plus a non-collapsed fit; verified by mutation.

Plus trigger boundaries at 11/12 (the true boundary — bracketing 10/20 left `>` vs `>=` undetected), `MIN_POINTS_FOR_REGRESSION`, count preservation, and idle no-op.

`refit_stale_estimators_preserves_relay_recorded_events` is a **sanity check, not a regression test** — `len()` is preserved by construction, so it cannot fail. Retained cheaply; not evidence.

**No simulation test**, deliberately: the sim harness's register returns `Ok(vec![])` unconditionally and the old code was guarded on `!history.is_empty()`, so the rebuild never fired in simulation and this bug is structurally unreproducible there. Such a test would be green on both `main` and this branch.

Full suite: **3,980 passed, 0 failed**. `cargo fmt` + `cargo clippy --locked -- -D warnings` clean.

## Post-deploy verification (no new instrumentation needed)

Re-run the issue's own `router_snapshot` query: `prediction_active` should climb 11-16% → ~85% over ~4h; snapshot-to-snapshot event losses should fall from 16.7% to ~0 except at restarts; `failure_events` should be monotonic within a session, saturating at 500.

## Out of scope (filed, not ignored)

- **#4810** — Renegade training can freeze once `trained_at > 3333`. Pre-existing; the 5-min rebuild was masking it.
- **#4811** — `ConnectForwardEstimator` is never refit (pre-existing), plus the case for moving the refit into `add_event`.
- **Startup restore** — untouched and separately dead (`aof.rs` filters route records to `record_ts >= new_records_ts`, stamped before `refresh_router` is spawned; never executed successfully since #3672, 2026-03-27). Stacked follow-up PR.

Fixes #4808

https://claude.ai/code/session_01QVhia9PxdeD3TPTXQvRBx4

### DIFF
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1504,9 +1504,13 @@ impl Ring {
         self.event_register.register_events(events).await;
     }
 
-    /// Maximum number of route events to load from the AOF event log on startup
-    /// and during periodic refresh. Caps memory use while retaining enough history
-    /// for the isotonic estimators to converge.
+    /// Maximum number of route events to load from the AOF event log on startup.
+    /// Caps memory use while retaining enough history for the isotonic estimators
+    /// to converge.
+    ///
+    /// Startup only: since #4808 the periodic loop no longer reads the event log
+    /// (it refits in place from each estimator's own in-memory window), so this is
+    /// consulted exactly once per process.
     const ROUTER_HISTORY_LIMIT: usize = 10_000;
 
     async fn refresh_router<ER: NetEventRegister>(
@@ -7976,6 +7980,72 @@ mod refresh_router_tests {
             "the event log must be read exactly once (startup) and NEVER again: \
              the periodic rebuild from disk discarded every relay-recorded event \
              (#4808). Got {total_calls} reads across >5 ticks"
+        );
+    }
+
+    /// The loop must actually REFIT — nothing else pins the one line that makes
+    /// #4808's fix do anything in production.
+    ///
+    /// Without this, deleting `router.write().refit_stale_estimators()` from
+    /// `refresh_router` leaves every other guard green: the never-reloads test
+    /// still sees its timeout and its single read, and the health-gauge scrape
+    /// still counts its call sites. This also restores the positive liveness proof
+    /// the superseded transient-error test provided via `total_calls > 4` — a bare
+    /// timeout cannot distinguish a healthy loop from one deadlocked on
+    /// `router.write()` or never ticking at all.
+    #[tokio::test(start_paused = true)]
+    async fn refresh_router_loop_refits_stale_estimators() {
+        let router = Arc::new(RwLock::new(Router::new(&[])));
+
+        // Seed past the staleness trigger so a refit is owed. These go straight
+        // into the in-memory router, exactly as a relay hop's
+        // `record_relay_route_event` does.
+        {
+            let mut guard = router.write();
+            for _ in 0..120 {
+                guard.add_event(crate::router::RouteEvent {
+                    peer: PeerKeyLocation::random(),
+                    contract_location: Location::random(),
+                    outcome: crate::router::RouteOutcome::Success {
+                        time_to_response_start: Duration::from_millis(100),
+                        payload_size: 5000,
+                        payload_transfer_time: Duration::from_millis(50),
+                    },
+                    op_type: Some(crate::node::network_status::OpType::Get),
+                });
+            }
+            assert!(
+                guard.refit_stale_estimators() > 0,
+                "sanity: a freshly-seeded router owes a refit before the loop runs"
+            );
+            // Re-dirty it so the loop has work to do.
+            for _ in 0..120 {
+                guard.add_event(crate::router::RouteEvent {
+                    peer: PeerKeyLocation::random(),
+                    contract_location: Location::random(),
+                    outcome: crate::router::RouteOutcome::SuccessUntimed,
+                    op_type: Some(crate::node::network_status::OpType::Get),
+                });
+            }
+        }
+
+        tokio::time::timeout(
+            Duration::from_secs(60 * 40),
+            super::Ring::refresh_router(
+                router.clone(),
+                WarmStartRegister { events: vec![] },
+                CancellationToken::new(),
+            ),
+        )
+        .await
+        .ok(); // timeout expected — a healthy loop runs forever
+
+        assert_eq!(
+            router.write().refit_stale_estimators(),
+            0,
+            "the loop must have consumed the outstanding staleness; a non-zero \
+             count here means it never called refit_stale_estimators (or never \
+             ticked), leaving the model to drift exactly as it did before #4808"
         );
     }
 
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -1552,42 +1552,36 @@ impl Ring {
             {
                 return;
             }
-            let history = match register.get_router_events(Self::ROUTER_HISTORY_LIMIT).await {
-                Ok(h) => {
-                    // Health gauge (#4440): a successful read clears the
-                    // consecutive-failure run and stamps the last-success time so
-                    // a *persistently* failing refresh (silent since the #4438
-                    // non-fatal hotfix) becomes observable on the snapshot
-                    // cadence. The read succeeded regardless of whether the
-                    // history is empty, so record success here, not below.
-                    BACKGROUND_TASK_HEALTH.record_refresh_router_success();
-                    h
-                }
-                Err(error) => {
-                    // get_router_events errors are transient and recoverable —
-                    // e.g. file-descriptor exhaustion ("os error 24") when the
-                    // AOF segment file can't be opened during a connection-churn
-                    // spike. Router refresh is a best-effort optimization, so we
-                    // keep the existing in-memory router in place and retry on
-                    // the next interval rather than killing the node. (A genuine
-                    // panic inside this task is still caught by the
-                    // BackgroundTaskMonitor — only this expected, transient read
-                    // failure is swallowed here.)
-                    //
-                    // Health gauge (#4440): bump the consecutive-failure run so
-                    // the escalation #4438 deliberately left non-fatal is at
-                    // least visible in telemetry.
-                    BACKGROUND_TASK_HEALTH.record_refresh_router_failure();
-                    tracing::error!(
-                        error = %error,
-                        "Failed to refresh routing history from event log; \
-                         keeping existing router and retrying on next interval"
-                    );
-                    continue;
-                }
-            };
-            if !history.is_empty() {
-                *router.write() = Router::new(&history);
+            // Refit the isotonic estimators in place from their own in-memory
+            // raw points. Each estimator only refits once >10% of its window has
+            // turned over, so an idle router does no work here.
+            //
+            // This deliberately does NOT re-read the event log. Until #4808 this
+            // loop did `*router.write() = Router::new(&history)`, rebuilding the
+            // whole router from disk every 5 minutes. That was destructive:
+            // `operations::record_relay_route_event` feeds the in-memory router
+            // ONLY — it never persists — so every relay-recorded event was
+            // discarded on each rebuild, giving relay events a <=5 minute
+            // half-life and resetting the model faster than it could reach
+            // `MIN_EVENTS_FOR_PREDICTION`. Observed in production as a peer's
+            // event count collapsing 29 -> 2 within one 30s window, and
+            // fleet-wide as 271,518 events lost to resets against 261,710
+            // gained (net zero accumulation).
+            //
+            // Within a session the on-disk log is a strict SUBSET of what the
+            // live router already knows (same originator events, minus every
+            // relay event), so rebuilding from it could only ever lose
+            // information. The batch-accuracy benefit that motivated the rebuild
+            // is preserved — and made complete — by refitting from `raw_points`,
+            // which holds every observation regardless of origin.
+            //
+            // Health gauge (#4440): this now marks "the refit pass ran", which is
+            // still the liveness signal the gauge exists to provide. The former
+            // fallible log read is gone, so there is no failure arm to record.
+            let refit_count = router.write().refit_stale_estimators();
+            BACKGROUND_TASK_HEALTH.record_refresh_router_success();
+            if refit_count > 0 {
+                tracing::debug!(refit_count, "Refit stale isotonic estimators");
             }
         }
     }
@@ -6273,7 +6267,7 @@ mod background_task_health_tests {
     /// Asserting against the process-global `BACKGROUND_TASK_HEALTH` after
     /// running `refresh_router` would be racy (concurrent tests share the
     /// global), so this pins the call sites in source instead — three startup
-    /// records plus two in the loop = five total.
+    /// records plus one in the loop = four total.
     #[test]
     fn refresh_router_records_health_on_startup_and_in_loop() {
         let src = include_str!("ring.rs");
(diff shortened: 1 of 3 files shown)


sanity:
## Multi-model review — synthesis and resolution

Full-tier review (routing surface): four independent Claude perspectives (skeptical / code-first / testing / big-picture) plus an external **Codex** pass. All findings below were verified against the checked-out code before acting. Every finding is fixed, filed, or answered.

### The external pass earned its keep

Codex found the load-bearing issue all four Claude reviewers missed, and its framing was the key insight:

> `Router::new` was doing four jobs — refit, prune, rebase, retrain — and this PR replaced only the first.

**Fixed in `599e9497`.** `raw_points` stored bare `Point<f64>`, discarding peer identity, so `refit()` *structurally could not* rebuild `peer_adjustments`. Two real consequences:

1. **`peer_adjustments` grew unbounded.** `raw_points` evicts at `MAX_REGRESSION_POINTS`; `peer_adjustments` only ever inserted. The old batch rebuild implicitly bounded it (the constructor drains to the last 500 events and derives adjustments only for peers in that window). Without it, a long-lived node accumulates one entry per peer *ever seen* across ~15 estimators — an unbounded per-key collection driven by remote peers, which `.claude/rules/code-style.md` explicitly bans. And `Adjustment::effective_count` has no time decay, so a returning peer's ancient adjustment applies at **full weight**.
2. **Adjustments stayed anchored to a curve that no longer existed.** An `Adjustment` is an EWMA of residuals measured *against* `global_regression`; refitting the curve alone leaves every peer corrected relative to the old one. Review measured the shift at up to **0.198 absolute on a [0,1] failure probability** — and the peer-adjusted value is what routing actually consumes.

My justifying comment ("`peer_adjustments` is an EWMA, inherently incremental, with no batch-vs-online discrepancy") was **wrong twice over**: it answered whether the EWMA needs a batch refit, not whether its basis moved; and there *is* a literal batch-vs-online discrepancy — the constructor seeds `effective_count = ADJUSTMENT_PRIOR_SIZE = 10`, while `add_event`'s `or_default()` starts at 0, which gates whether a peer's adjustment applies at all.

**Fix:** `raw_points` → `raw_events` retaining the whole `IsotonicEvent`; the constructor's fitting logic extracted into a shared `IsotonicEstimator::fit` used by **both** `new_with_mode` and `refit`. A refit is now a true batch rebuild over the current window — what the old `Router::new(&history)` achieved, except the corpus is complete instead of missing every relay-recorded event.

### Two of my tests were worthless. Both proven, both fixed.

- **`refit_respects_descending_estimator_direction` was VACUOUS** — it guarded the worst failure mode (a `Negative` estimator refit ascending) and could not detect it. PAV over data violating its assumed direction pools everything into **one flat aggregate**, and a flat line satisfies `near >= far` by *equality*. Review ran the wrong direction 9/9 → passed. Now asserts a **strict** decrease plus a non-collapsed fit. **Verified by mutation:** flipping the direction gives `near == far == 0.805` and the test fails.
- **`refit_stale_estimators_preserves_relay_recorded_events` is tautological**, and the PR body's claim that it "fails against the old rebuild" was **false** — it cannot even compile against the old code. `len()` is `centroid.sum_weight`, and `refit()` rebuilds *from* `raw_events`, so the count is preserved by construction; even a deliberately wrong refit would pass. Retained as a cheap sanity check, **re-labelled** in the PR body. The genuine regression guard is `refresh_router_never_reloads_from_disk_after_startup` (read-count == 1), which does fail against the old loop.
- **`refit_matches_a_batch_build_over_the_same_points` deleted** — tautological (both sides called the same constructor over an identical slice) *and* misnamed (a refit did **not** match a batch build). Subsumed by the new, stronger `refit_rebuilds_peer_adjustments_against_the_new_curve`, which compares `estimate_retrieval_time` — the value routing consumes.

### The wiring was unpinned (`40da696c`)

Nothing asserted the loop actually **calls** `refit_stale_estimators` — the one line that makes this fix do anything. **Verified by mutation:** replacing it with `0usize` fails *only* the new `refresh_router_loop_refits_stale_estimators`; the other five guards stay green.

This also restores the positive liveness proof lost when the superseded test went: it proved >=5 iterations via `total_calls > 4`, whereas a bare timeout fires identically if the loop deadlocks on `router.write()` or never ticks.

### Other findings addressed

- **Trigger boundary unpinned** — bracketing 10/20 left `>` vs `>=` undetected. Now pins **11** (holds off) and **12** (fires), the true boundary, plus a `MIN_POINTS_FOR_REGRESSION` case.
- **False "Deterministic" comment** on the drift test — `PeerKeyLocation::random()` draws from system entropy, and `route_distance()` folds a random peer location into the x-axis, so the constructed geometry was not what it read as. New `event_at_distance` helper places the contract at a ring offset *from* the peer, making `route_distance() == x` exactly. (Empirically it was never flaky — 250 and 300 runs, 0 failures — but the comment was wrong.)
- **Wrong error-arm rationale** — `new_ascending`/`new_descending` cannot fail *at all*: both pass `intersect_origin: false`, making the sole error variant unreachable, and empty input succeeds. Handling retained (better than `.expect()` on the routing path), reason corrected.
- **Cadence claim corrected** — the realised cadence is `min(poll interval, turnover)`. Above ~10 events/min the 5-minute tick binds and a window can turn over entirely between refits. That's ~40× the production median (~14-16 events/hour), where turnover always binds first, so it's a documentation caveat — now spelled out on `REFIT_STALENESS_PERCENT`.
- **Two stale rustdocs** — `ROUTER_HISTORY_LIMIT` ("startup **and during periodic refresh**") and the #4440 gauge rustdoc.
- **#4440 gauge semantics silently changed** — worth spelling out rather than deleting. The loop has no fallible call left, so the only `record_refresh_router_failure` is the startup read, and the first tick unconditionally records success and clears the run. `refresh_router_consecutive_failures` is now structurally 0-or-1 and **cannot report a persistent failure**; documented as "do NOT alert on it". `last_success_age_secs` survives as a genuine liveness heartbeat.

### Verified, not assumed

- **Lock cost is a non-issue — measured.** A saturated 500-point refit is **27.3µs**; all 15 estimators worst-case **~410µs**, once per 5 min. `parking_lot::RwLock`, no `.await` under the guard, single lock so no ordering risk. Strictly **cheaper** than the old `Router::new` over up to 10,000 events under the same lock.
- **The central premise holds** — the incremental fit really does drift: **3-6% of span (response time), 21% (failure estimator)** on realistic routing data. The refit is a real accuracy mechanism, not ceremony.
- **`estimator_type` cannot diverge** — one struct literal, all paths funnel through `new_with_mode`, and `IsotonicEstimator` derives `Serialize` **only** (no `Deserialize`), so `#[serde(skip)]` has no decode path that could reset a `Negative` estimator to `Positive`.
- **A refit cannot flip `prediction_active` off** — `len()` is observation weight, not block count, and refit preserves it exactly.

### Deliberately out of scope — filed, not ignored

- **#4810** — Renegade training can freeze permanently once `trained_at > 3333` (the 50%-growth ladder outruns the 5,000 cap). Pre-existing; `routing_predictor.rs` isn't in this diff. The 5-minute rebuild was accidentally *masking* it, so this PR makes it reachable. Needs ~9 days of uninterrupted uptime at the production median, versus ~daily auto-updates.
- **#4811** — `ConnectForwardEstimator` is never refit (pre-existing; the old rebuild never touched it either), plus the argument for moving the refit into `add_event`, which would fix it for free and let `refresh_router` be deleted outright.
- **Startup restore** — untouched here and separately dead (`aof.rs` filters route records to `record_ts >= new_records_ts`, stamped before `refresh_router` is spawned; it has never executed successfully since #3672, 2026-03-27). Stacked follow-up PR.

### `rule-review/warnings`

The one warning is the superseded test (`git-workflow.md` / `testing.md`: prefer `#[ignore]` + retain over delete). I judged deletion correct here and want that auditable rather than silent: the old test's second invariant was *"the loop recovers by reloading history from disk"* — the exact behaviour this PR removes as the bug. Retaining it under `#[ignore]` would leave a test asserting the router reloads from disk sitting beside one asserting it must never do so. That's not historical documentation; it's a loaded gun for the next reader. Its *first* invariant (a transient error must not kill the monitored task) is retained and re-asserted in the replacement, and the liveness proof it provided is restored by `refresh_router_loop_refits_stale_estimators`. Requesting `review-override` on that basis.

**No simulation test.** `.claude/rules/test-philosophy.md` nominally pushes for one on routing changes, so the omission should be auditable: the sim harness's register returns `Ok(vec![])` unconditionally, and the old buggy code was guarded on `!history.is_empty()` — so **the rebuild never fired in simulation** and #4808 is structurally unreproducible there. Such a test would be green on both `main` and this branch. The `FlakyRegister` unit test reproduces the mechanism (a register whose log is a subset) deterministically and in milliseconds.

3,980 tests pass; `cargo fmt` + `cargo clippy --locked -- -D warnings` clean.

[AI-assisted - Claude]

PR #4812, opened 2026-07-15
refactor(router): delete the dead routing-history restore path

sanity:
Follow-up to #4808 / #4809. Now targets `main` directly — #4809 squash-merged as `2c677810`, so this is no longer stacked.

## Problem

#4809 fixed the *periodic* rebuild that discarded relay-learned events. The **startup restore** was left in place. It turns out to be dead code.

**Bug 1 — the restore effectively never fired.** `LogFile::get_router_events` filtered records to `record_ts >= NEW_RECORDS_TS`. `NEW_RECORDS_TS` is a **process-global** `OnceLock` (`tracing::register`), stamped `SystemTime::now()` at the *first event-register construction in the process* — `EventRegister::new`, or `OTEventRegister::new` under `trace-ot`. A production node builds its register (`node.rs:680`) before it can write a single route event and before `OpManager::new` → `Ring::new` spawns the task, so the cutoff was always ≥ that process's start and every prior-session record was discarded. Dead since #3672 (2026-03-27).

The invariant is **"effectively never in production", not "structurally unreachable"** — the rustdoc says so explicitly so nobody tightens it later. Two ways the branch *is* reachable: both sides truncate to whole seconds, so a prior-session record written in the same wall-clock second as the stamp passes; and an in-process multi-node test sharing one `data_dir` (`tests/in_process_restart.rs:88`) makes the second node's `get_or_init` a no-op, so it reads the first node's records. Neither happens on a real node.

**Production evidence.** Gateways run `RUST_LOG="info,freenet=info"`, and the success arm is `tracing::info!("Restored routing history from event log")`. In vega's retained logs (`NRestarts=2`, 62 files):

| | |
|---|---|
| Node startups captured ("Freenet node starting") | **8** |
| `"Restored routing history from event log"` (INFO, success arm) | **0** |
| `"Failed to load routing history"` (WARN, error arm) | **0** |
| Control: INFO lines from `freenet::ring` captured | **141,158** |

Across 8 startups neither arm ever fired, while 141k other INFO lines from the same module *are* captured — i.e. the read always returned `Ok(empty)`, exactly as the filter predicts. (nova has `NRestarts=0` and no boot inside its retained window, so it contributes no startup evidence.)

**Bug 2 — the scan was also mis-budgeted.** It incremented `visited` for *every* record kind and broke at 10,000 (`MAX_EVENT_HISTORY`), while production retains 100,000 (`MAX_LOG_RECORDS`). Past 10k records it burned its entire budget on the oldest records — which the filter then discarded anyway. Neither bug ever tripped a test because `MAX_LOG_RECORDS` is `#[cfg(test)] = 10_000`, exactly equal to the ceiling, and every restore test used a mock register that bypassed the timestamp filter.

**Nothing else consumes `EventKind::Route` from the AOF.** `freenet service report` bundles only plain-text `.log` files; `fdev network-metrics-server` has no Route variant; `fdev inspect`/`query`/`diagnostics` have no event-log references. The only `read_all_events` consumer is the offline aggregator, which reads all kinds generically.

## Approach

Delete the reader. Keep the writer. Don't persist relay events.

Fixing the restore instead would mean correcting **both** bugs on a Full-tier surface to buy a warm start that isn't worth much: the per-peer `peer_adjustments` are meaningless across a restart (CONNECT rebuilds a different neighbour set); the global isotonic curves *would* survive, but a **partial** restore — reloading a log with originator events but no relay hops — is exactly what caused #4808. Post-#4809 a peer reaches `MIN_EVENTS_FOR_PREDICTION = 50` in ~3.5h at the observed median ~14-16 events/hour.

**Deleted:** the startup restore block; `get_router_events` (trait method + all five impls + the `aof.rs` implementation and its `MAX_EVENT_HISTORY` cap); `ROUTER_HISTORY_LIMIT`; `LOGGED_COMPAT_WARNING` (a dedup static only that reader used — `read_all_events` has its own per-record tolerance); and the `Clone` bound on `Ring::new`'s `ER`, which existed solely for the register clone handed to the spawned task.

**Renamed** `refresh_router` → `refit_router_periodically`. It no longer refreshes from anything; it refits in place.

### The #4440 health gauges — an asymmetric split, and a judgment call

I originally deleted both gauges and described it as a clippy-forced cascade. **That was wrong on both counts**, and review caught it. Corrected:

- **`refresh_router_consecutive_failures` — deleted.** Deleting the restore removes the task's last fallible call, so the counter loses its only failure source and is pinned at 0 forever. The failure-recording arm goes with it. This one really is dead.
- **`refresh_router_last_success_age_secs` — KEPT.** `BackgroundTaskMonitor` only watches for task *death*. A loop that is alive but starved — deadlocked on `router.write()`, or never scheduled — is invisible to it, and shows up *only* as a rising age. That residual value is real, and `40da696c` already documented this gauge as a deliberate liveness heartbeat; this PR respects that rather than silently reversing it.

Deleting the remaining ~180 LOC of `BACKGROUND_TASK_HEALTH` would have been a **judgment call, not a forced cascade**: after the restore block goes, only `record_refresh_router_failure` actually dies — `record_refresh_router_success()` is still called in the loop and compiles fine. The minimal clippy-clean change is "drop the failure arm + `consecutive_failures`", which is what this PR now does. The facility is slimmed to a single-field heartbeat rather than removed.

The OTLP key keeps its `refresh_router_*` name despite the code rename, so the existing metric series and any dashboard querying it survive. Renaming a live telemetry key is a separate, riskier change than the code rename that prompted it.

### Kept deliberately

- **`NetEventLog::route_event` and its originator call sites** — telemetry depends on them, and `test_router_accumulates_feedback_events` / `test_router_accumulates_failure_events_with_message_loss` assert `route_count > 0`. (Those read an in-memory `Arc<Mutex<Vec<NetLogMessage>>>`, not the AOF, so the reader deletion structurally cannot affect them — but both are re-verified green regardless.)
- **`Router::new(&[RouteEvent])`** — the batch constructor, still used by `Ring::new` and tests.
- **`read_all_events` / `NEW_RECORDS_TS`** — `read_all_events` is the surviving public reader (offline aggregator, `state_verifier`) and applies the same timestamp filter, so `NEW_RECORDS_TS` stays load-bearing.
- **`read_record_buffers`'s `limit` parameter** — already always `None` before this PR; unrelated to the restore, so not smuggled into this diff.

## Testing

`cargo test -p freenet --lib` — **4021 passed, 0 failed**. `cargo fmt`, `cargo clippy --locked -- -D warnings`, and `cargo clippy --locked -p freenet --features trace-ot -- -D warnings` all clean. Both simulation tests green.

**`refresh_router_loop_refits_stale_estimators` ports forward** as `refit_router_periodically_refits_stale_estimators` — unchanged except for dropping the register argument. **Mutation-verified:** stubbing `refit_stale_estimators()` to `0usize` fails **only** this test; the shutdown guard, both heartbeat tests, and `refit_stale_estimators_preserves_relay_recorded_events` all stay green. It remains the only guard on the loop→refit wiring.

Four tests are deleted outright; each references a deleted type/method and an `#[ignore]`d test must still compile, so none could be ignored. `test-exempt` label applied with a per-test justification in [this comment](https://github.com/freenet/freenet-core/pull/4812#issuecomment-4983813903).

An earlier revision of this PR contained a replacement test, `refit_router_periodically_never_resets_the_live_model`, which review correctly identified as **vacuous** — it asserted `success_events == 100`, but `success_events` is `raw_events.len()` and refitting never changes it, so an empty loop body passed. It also seeded via `Router::new(&events)`, leaving `events_since_refit = 0`, so no tick ever refit anything. Removed; the ported wiring test covers the real property, and `refit_stale_estimators_preserves_relay_recorded_events` already pins model survival at the `Router` level.

Stale doc references to deleted gauges/tests are repointed at surviving equivalents, and the `raise_fd_limit` rationale in `bin/freenet.rs` — which cited the restore's AOF read as the fd-exhaustion victim — notes the read is gone while the exhaustion mode (module-cache memfds, the AOF *writer*, sockets) is unchanged.

https://claude.ai/code/session_01QVhia9PxdeD3TPTXQvRBx4

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -85,11 +85,14 @@ mod build_info {
 /// by a `memfd`, plus AOF segment files, UDP sockets, the WS API, etc. systemd's
 /// `DefaultLimitNOFILE` soft limit is typically 1024, and freenet never raised
 /// it, so a busy gateway exhausted file descriptors. The resulting `EMFILE`
-/// killed a monitored background task (`refresh_router` opening an AOF segment),
-/// which the `BackgroundTaskMonitor` treats as node-fatal — producing a
-/// systemd crash-loop. Raising the soft limit to the hard limit at startup gives
-/// the node the FD headroom the kernel already permits, without operator
-/// intervention or a unit-file change.
+/// killed a monitored background task — at the time, the periodic router refresh
+/// opening an AOF segment to reload routing history; that read has since been
+/// deleted (#4808 follow-up), but the module cache, the AOF *writer*, and the
+/// sockets all still consume fds, so the exhaustion mode is unchanged — which the
+/// `BackgroundTaskMonitor` treats as node-fatal, producing a systemd crash-loop.
+/// Raising the soft limit to the hard limit at startup gives the node the FD
+/// headroom the kernel already permits, without operator intervention or a
+/// unit-file change.
 ///
 /// Best-effort and non-fatal: on any error we `warn!` and continue with the
 /// inherited limit rather than failing startup. No-op on non-unix.
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -1283,8 +1283,8 @@ mod tests {
     ///
     /// Asserting against the process-global `BROADCAST_STREAM_METRICS` after
     /// running the broadcast would be racy (concurrent tests share the global),
-    /// so this pins the call sites in source instead — mirroring the
-    /// `refresh_router` health-gauge pin in `ring.rs`.
+    /// so this pins the call sites in source instead — mirroring
+    /// `migration_counter_sites_present` in `ring/placement_migration_metrics.rs`.
     #[test]
     fn broadcast_to_single_peer_records_attempt_on_every_streaming_exit_pin() {
         let src = include_str!("broadcast_queue.rs");
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -488,7 +488,7 @@ impl Ring {
     /// Max hops to be performed for certain operations (e.g. propagating connection of a peer in the network).
     pub const DEFAULT_MAX_HOPS_TO_LIVE: usize = 10;
 
-    pub fn new<ER: NetEventRegister + Clone>(
+    pub fn new<ER: NetEventRegister>(
         config: &NodeConfig,
         event_loop_notifier: EventLoopNotificationsSender,
         event_register: ER,
@@ -505,18 +505,19 @@ impl Ring {
         };
 
         // Single shutdown signal for every long-lived background task spawned
-        // below. Created up front so `refresh_router` (spawned before the
-        // `Ring` struct literal exists) shares the same token that gets moved
+        // below. Created up front so `refit_router_periodically` (spawned before
+        // the `Ring` struct literal exists) shares the same token that gets moved
         // into the struct. See issue #4278.
         let shutdown = CancellationToken::new();
 
+        // Starts cold and learns from this session only — there is no warm start
+        // from the event log (see `refit_router_periodically`).
         let router = Arc::new(RwLock::new(Router::new(&[])));
         crate::node::network_status::set_router(router.clone());
         task_monitor.register(
-            "refresh_router",
-            GlobalExecutor::spawn(Self::refresh_router(
+            "refit_router_periodically",
+            GlobalExecutor::spawn(Self::refit_router_periodically(
                 router.clone(),
-                event_register.clone(),
                 shutdown.clone(),
             )),
         );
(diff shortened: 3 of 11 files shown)


sanity:
## `test-exempt` label: justification

CI's no-deleted-tests gate fires on this PR (`REMOVED_TESTS=4, ADDED_TESTS=0`, reproduced locally against `origin/main`). Adding the label per the escape hatch that `ci.yml:332-334` describes for exactly this case — "retiring a data structure whose unit tests can't be `#[ignore]`d because the type itself is gone".

**Why `#[ignore]` was not an option for any of the four:** an `#[ignore]`d test still has to *compile*. Each of these references a type or method this PR deletes, so ignoring them would leave the crate unbuildable.

| Deleted test | References (now deleted) | Disposition |
|---|---|---|
| `refresh_router_loads_history_on_startup` | `WarmStartRegister`, `Ring::refresh_router` | Asserted the startup restore populates the router. That path is gone. |
| `refresh_router_handles_empty_history` | `WarmStartRegister`, `Ring::refresh_router` | Same. |
| `refresh_router_records_health_on_startup_and_in_loop` | `record_refresh_router_failure`; `include_str!` + `src.find("async fn refresh_router").expect(...)` | Source-scrape pin over the three startup health arms. Its subject is deleted, and the `.expect` would panic on the rename. Its surviving concern (the loop records a heartbeat) is covered by `background_task_health_tests`. |
| `refresh_router_never_reloads_from_disk_after_startup` | `FlakyRegister`, register read-counting | Counted register reads to prove the loop never re-reads the log. No longer *expressible*: `refit_router_periodically` takes no event register and `NetEventRegister` has no route-event reader, so "the loop never reads the log" is now a **compile-time** property. |

Note the gate under-counts: it reports 4 removed / 0 added, but I did add tests — they reuse the base's existing `#[tokio::test(start_paused = true)]` attribute as an unchanged context line, so the diff never marks the attribute as added. Tests actually added or ported forward:

- **`refit_router_periodically_refits_stale_estimators`** — `refresh_router_loop_refits_stale_estimators` ported forward unchanged; the only edit is dropping the register argument. **Not dropped.** Mutation-verified: stubbing `refit_stale_estimators()` to `0usize` fails ONLY this test — the shutdown guard, both heartbeat tests, and `refit_stale_estimators_preserves_relay_recorded_events` all stay green.
- **`refit_router_periodically_returns_promptly_on_shutdown`** — `refresh_router_returns_promptly_on_shutdown` (#4278), renamed to match.
- **`never_succeeded_reports_none_age`** / **`success_stamps_age_which_grows_with_time`** — the heartbeat half of the old health tests; the failure-counter half went with `consecutive_failures`.
- The four `aof.rs` tests that used `get_router_events` only as a read-back assertion now use `read_all_events`, keeping the rotate/truncate/legacy-migrate coverage intact.

Net: `3996 -> 4021` lib tests passing. No test coverage is lost that could have been retained.

[AI-assisted - Claude]

sanity:
Addressed the rule-review warning rather than overriding it: `isotonic_estimator.rs:25` documented `Ring::refresh_router` as the only caller of `refit_if_stale`, which this PR's rename made stale. Fixed in place (comment-only — verified the two-dot diff against `main` for that file contains no non-comment lines, so none of #4809's `raw_events` / shared-`fit` work is touched).

Also worth recording why `Rule Lint` failed on the previous head and passes now: the run was triggered by the force-push *before* I applied `test-exempt`, and the gate reads labels from the frozen event payload (`HAS_EXEMPT_LABEL=$(echo "" | grep -c 'test-exempt')` in the log). `pull_request` defaults to `[opened, synchronize, reopened]`, so labelling alone doesn't re-trigger it — a reopen refreshed the payload and it passed. No code change was involved.

[AI-assisted - Claude]

Issue #2380, opened 2025-12-22
fix: Subscribe fails after PUT due to state_store key mismatch


sanity:
## Problem

Room creation in River fails intermittently - the first room creation after a peer restart succeeds, but subsequent ones time out with "Timeout waiting for PUT response after 60 seconds".

### Root Cause Analysis

The PUT operation completes successfully and stores the contract state, but the subsequent Subscribe operation cannot find the contract via `has_contract` check.

**Timeline from logs (failed case):**
- `23:49:37.828223Z` - PUT completed: `state_store.store()` called successfully
- `23:49:37.974617Z` - Subscribe::Request received (146ms later)
- `23:49:37.975042Z` - `has_contract` returned `Some(None)` - **state NOT found**

### Technical Details

The `has_contract` check flow:
1. `lookup_key(instance_id)` - Gets `ContractKey` from `contract_store.code_hash_from_id()`
2. `fetch_contract(key)` - Calls `state_store.get(&key)` 
3. Returns `Ok(Some(key))` only if `state: Some(_)`

For failed cases:
- `lookup_key` succeeds (contract code is stored correctly)
- `fetch_contract` returns `state: None` (state not found in state_store)

**Hypothesis:** The `ContractKey` used during `state_store.store()` (in `upsert_contract_state`) differs from the key reconstructed during `lookup_key()` + `fetch_contract()`.

### Why CI Didn't Catch This

CI six-peer regression tests use a mesh topology where multiple peers cache the contract. If the gateway doesn't respond to Subscribe, another peer can. In production with a single peer behind NAT, the gateway is the only path.

### Reproduction Steps

1. Start a fresh peer connected to the gateway
2. Create a room with `riverctl room create --name test-room`
3. First room creation succeeds
4. Second room creation fails with 60s timeout

### Logs

**Successful first room:**
```
has_contract: Some(Some(ContractKey { instance: ContractInstanceId("DPNCPjS4..."), code: CodeHash("afdmvqpw...") }))
```

**Failed second room:**
```
SUBSCRIBE: has_contract check complete, has_contract: Some(None)
SUBSCRIBE: Contract not found, waiting for in-flight PUT
SUBSCRIBE: Contract not found locally, will try forwarding
SUBSCRIBE: No forward targets found
```

### Suggested Fix

Add diagnostic logging to trace the exact `ContractKey` used in:
1. `upsert_contract_state` when calling `state_store.store()`
2. `fetch_contract` when calling `state_store.get()`

[...]

PR #4820, opened 2026-07-15
fix(contracts): evict compiled module on contract reclaim (#4754)

### LINKED ISSUE #4754: WASM module cache: couple contract-module removal to hosting eviction (reclaim promptly, not on cold-LRU pressure)

netsirius:
## Problem

The delegate path (`Runtime::unregister_delegate`) removes the compiled module from its cache on teardown, but the contract path (`Executor::reclaim_contract_storage`) does not: it deletes the persisted state and the `.wasm` blob from disk but leaves the compiled module in `contract_modules`. So a reclaimed contract's module (~1-2 MB) lingers until byte-budget LRU pressure evicts it, instead of being reclaimed promptly.

Note: #4754 was originally filed as an unbounded leak, but that was walked back in the thread — the module cache is already a byte-budget-bounded, RAM-scaled LRU. `evictions_total=0` on the soak node was headroom, not a missing bound. This fixes only the promptness/consistency gap.

## Solution

Mirror the delegate path:

- Add `Runtime::remove_contract_module(key)`, a `pub(crate)` wrapper over the existing `ModuleCache::remove` (needed because `contract_modules` is `pub(super)`). No new cache API.
- Call it from `reclaim_contract_storage` on the code-deletion success branch.

Out of scope: the optional "dedup by `CodeHash`" idea from the issue is a behavior change and left out.

## Testing

New regression test `remove_contract_evicts_compiled_module_from_cache`: seed the module cache for a key, run `reclaim_contract_storage`, assert the module is gone. Fails without the fix, passes with it. `cargo fmt` and `cargo clippy -p freenet --lib` clean.

## Fixes

Fixes #4754


iduartgomez (approved):

sanity:
@netsirius — apologies, this sat for five weeks with no response. That is on us, not on you; the change was in good shape when you opened it.

Reviewing it now against current `main`, the problem is that the code moved underneath the PR and the change no longer applies as written.

## Blocking: `contract_modules` is now keyed by `CodeHash`, not `ContractKey`

PR #5324 ("bound peer memory at the 2 GiB cap — module-cache keying", merged 2026-08-14) re-keyed the shared module cache:

```rust
// crates/core/src/wasm_runtime/runtime.rs:841
pub(super) contract_modules: SharedModuleCache<CodeHash>,
```

`Runtime::remove_contract_module(&self, key: &ContractKey)` forwards to `ModuleCache::remove`, whose bound is `K: Borrow<Q>` — so it now needs `CodeHash: Borrow<ContractKey>`, which does not exist. **The branch merges cleanly textually but does not compile against `main`.** That is also why nothing here flagged it: the merge is conflict-free and the last CI run is from July.

Ironically this is the "dedup by `CodeHash`" idea you explicitly scoped out of the PR — it landed separately.

## Blocking: a straight re-key to `CodeHash` would be a regression

Now that one compiled module is shared by every instance with the same code, evicting it on a single instance's reclaim would drop a module that other still-hosted contracts are using. Every River room shares one room-contract WASM (#2380), so reclaiming one room would force a recompile for all the rest — a worse outcome than the lingering module the PR is trying to reclaim.

## The shape that does work

`ContractStore::remove_contract` already computes exactly the predicate needed, for exactly this reason, on the code blob:

```rust
// crates/core/src/wasm_runtime/contract_store.rs:536
let code_still_referenced = match self.db.load_all_contract_index() { ... };
...
// :562
if !code_still_referenced {
    // Invalidate the code cache so a removed contract is never served
    // as a "ghost" (issue #3487).
    self.contract_cache.invalidate(&contract_hash);   // :565
    ... delete the .wasm blob ...
}
```

The compiled-module eviction belongs on the same branch as that `contract_cache.invalidate` — the two caches have identical lifetime rules. `ContractStore` has no handle on `contract_modules`, so the least-surprising way is to have `remove_contract` return the disposition (something like `CodeDisposition::{Removed, StillReferenced}`) and have `reclaim_contract_storage` (`contract/executor/runtime/contract_ops.rs:582`) call `remove_contract_module(&code_hash)` only on `Removed`. That also keeps the whole decision behind the `contract_blob_lock` the store already holds, so it can't race a sibling executor's `store_contract` of the same code.

Test-side: `seed_contract_module_for_test` / `contract_module_cached_for_test` need to take a `CodeHash`, and it is worth adding the second case explicitly — reclaim instance A, assert instance B (same code, different params) still has its module — since that is the property the re-key makes easy to get wrong.

## Two smaller notes

- The motivation in the description ("~1-2 MB per reclaimed contract") is now a per-**code** figure rather than per-instance, so the reclaim is worth less than it was in July: it only fires for genuinely unique code. Still worth doing, and #4754 is still open and still real — just worth restating the size of the win.
- The last CI run (July head) had **NAT Validation** red with everything else green. It will need a fresh run after the rebase regardless.

## Next step

Entirely your call: happy for you to take it, or say the word and one of us will pick it up and finish it with the `CodeHash` shape above so it does not sit any longer. Either way, sorry again for the silence.

[AI-assisted - Claude]

sanity:
Confirmed by building it rather than reasoning about it — I merged this branch onto current `main` (clean, no conflicts) and ran `cargo check -p freenet --lib --all-targets`:

```
    --> crates/core/src/wasm_runtime/runtime.rs:1446:54
     |
1446 |         self.contract_modules.lock().unwrap().remove(key);
     |                                               ------ ^^^ expected `&CodeHash`, found `&ContractKey`
     |
     = note: expected reference `&freenet_stdlib::prelude::CodeHash`
                found reference `&freenet_stdlib::prelude::ContractKey`
```

So the "merges cleanly but does not compile" part is measured, not inferred. The compile aborts there, so the two `#[cfg(test)]` helpers will need the same re-key once the production line is fixed.

[AI-assisted - Claude]

sanity:
## Review: this cannot be rebased — the cache was re-keyed underneath it

@netsirius — sorry this sat for seven weeks. Reviewed properly now, and the outcome is that the branch needs rewriting rather than rebasing. I refreshed it onto current `main` and it no longer compiles.

`contract_modules` is now `SharedModuleCache<CodeHash>` (`crates/core/src/wasm_runtime/runtime.rs:1042`), re-keyed from instance to code hash on 2026-08-14 by `2b1e9ec0c` (#5324/#5268) — a month after this branch's last real commit. Three `E0308`s at `runtime.rs:1651`, `:1668`, `:1674`, all `ContractKey` where `CodeHash` is wanted.

### The type error is the small part

Translating the key and moving on would be wrong. Several contract instances with different params share one `CodeHash`, so "evict this contract's module" is no longer a well-formed operation.

It is not a correctness bug — modules leave the cache by `.cloned()` (`runtime.rs:1569-1574`) and wasmtime `Module` is refcounted, so an in-flight execution holds its own handle and cannot be invalidated. The next call is an ordinary miss that recompiles from the blob, which is still on disk.

It is a performance bug, in exactly the deployment this came from. That recompile is a Cranelift compile the code itself calls "potentially multi-hundred-millisecond" and deliberately offloads off the contract-handling loop (`runtime.rs:1583-1592`, #4441). On a gateway hosting N River rooms on one room-contract WASM, reclaiming any one cold room forces that recompile for the next room to receive a message.

### `remove_contract` already knows the answer

`contract_store.rs:502-577` deletes the instance's index entry unconditionally, but deletes the shared `.wasm` blob only when `load_all_contract_index()` shows no remaining instance on that code hash — deliberate, documented against #2380, and holding `contract_blob_lock` across removal→scan→delete. So there is no pre-existing bug here; the gate works.

The problem for this PR is that `remove_contract` returns `Ok(())` in both the blob-deleted and blob-kept cases, and this PR hangs eviction off `Ok(())`.

**The shape of the real fix:** have `remove_contract` return which happened (`BlobDeleted` / `BlobRetained`) — it already computes `code_still_referenced` inside the lock — and evict the module iff the blob was actually deleted. That is the delegate path's exact shape (`wasm_runtime/delegate/interface.rs:339-357`). Five call sites, four currently ignore the result.

Two tempting shortcuts are wrong: a post-hoc `code_blob_stored` probe conflates "unreferenced" with "blob missing for another reason", and evicting there turns a still-running contract into a broken one; and a `code_still_referenced` helper mirroring `delegate_store.rs:203` either reads the non-authoritative in-memory map or re-runs `load_all_contract_index()` — O(all contracts) and racy once the lock is released. Also note `key.code_hash()` is an unverified serde field that `ContractKey`'s `Hash`/`Eq` never consult (`runtime.rs:1538-1546`), so the hash must come from `code_hash_from_id`, captured *before* `remove_contract` drops the index entry.

### Also worth carrying forward

- `executor_impl.rs:458,494,527,579` — the PUT-rollback paths get no eviction at all. The one at `:527` fires after `validate_state` has compiled and cached the module, i.e. it leaves the very module #4754 is about.
- The regression test seeds and probes with the same `key` expression and never touches the production insert path, so it cannot detect a key-scheme mismatch. A single-instance fixture is exactly the case where unconditional eviction is accidentally correct.

### Disposition

Closing this, because nothing survives a rebase but the intent — all three hunks and the test need rewriting, and it is a ~30-line change against current `main`. Reopen or push over it if you'd rather keep the thread.

**#4754 stays open.** #5324 shrank the payoff by the sharing factor but did not remove it: the *last* instance of a code hash still leaves ~1-2 MB until byte pressure. Small, real, and the fix above is the right one. I have not touched the issue.

Reviewed by reading the checked-out code, plus `cargo check -p freenet --lib` on the PR head in a throwaway worktree to confirm the compile failure matches CI. The regression test was not run — the branch does not compile, so there is no baseline to revert the fix from.

[AI-assisted - Claude]

Issue #4819, opened 2026-07-15
flaky: test_set_own_addr_atomic_with_network_status races unlocked set_own_addr writers in op_state_manager tests


sanity:
Found while working on #4811 (unrelated change). **Pre-existing on `main`, not caused by that PR** — reproduced on a pristine `origin/main` worktree at 95a3746a with no local changes.

## Symptom

`ring::connection_manager::tests::test_set_own_addr_atomic_with_network_status` fails intermittently:

```
round 2: own_addr (Some(127.0.0.1:20018)) and network_status::external_address
(Some(127.0.0.1:12000)) diverged — set_own_addr is not atomic (#4172)
```

Reproduction (pristine `origin/main`, 95a3746a):

```
cargo test -p freenet --lib connect
```

Failed 2 of 14 observed runs on main. Always passes when run in isolation (`cargo test -p freenet --lib test_set_own_addr_atomic_with_network_status`), which is what makes it a cross-test interference bug rather than a real atomicity failure.

## Root cause (hypothesis, evidence below)

The test guards #4172: `set_own_addr` must update the `own_addr` mutex and the process-global `network_status::external_address` atomically. It spawns concurrent writers, joins them, then asserts the two mirrors agree.

Because `network_status` is a **process-global**, tests that touch it are expected to serialize on the crate-wide `network_status::TEST_GLOBAL_STATE_LOCK`. This test does take it (`connection_manager.rs:3655`).

`node::op_state_manager`'s tests do **not**. `seed_location_and_one_connection` (`op_state_manager.rs:2954-2956`) calls:

```rust
let self_addr: std::net::SocketAddr = "127.0.0.1:12000".parse().unwrap();
op_manager.ring.connection_manager.set_own_addr(self_addr);
```

`set_own_addr` writes the global (`connection_manager.rs:~1618`, the #4172 fix site). `op_state_manager.rs` contains **zero** references to `TEST_GLOBAL_STATE_LOCK` while calling `seed_location_and_one_connection` from at least 6 tests (lines 3148, 3202, 3243, 3290, 3336, 4940).

So an unlocked writer can clobber `external_address` to `127.0.0.1:12000` in between the locked test's writer-join and its assertion. The observed value in the failure message is **exactly** `127.0.0.1:12000`, which is the smoking gun: that address appears nowhere else relevant (`grep -rn "12000" crates/core/src/`).

[...]

Issue #4810, opened 2026-07-15
fix(router): Renegade predictor training can freeze permanently once trained_at exceeds ~3333


sanity:
## Problem

`PredictionStage::should_train` can freeze permanently on a busy router, leaving the Renegade predictor's K and distance metric pinned at whatever they were when training last ran.

Found by Codex during review of #4809.

## Mechanism

Training fires when `n >= trained_at + trained_at/2` (i.e. 50% growth). Starting from `MIN_OBSERVATIONS_FOR_TRAINING = 20`, the ladder is exact:

```
20 → 30 → 45 → 67 → 100 → 150 → 225 → 337 → 505 → 757 → 1135 → 1702 → 2553 → 3829
```

At `trained_at = 3829` the next training needs `n >= 5743`. But the stage is capped at `RENEGADE_MAX_OBSERVATIONS = 5000` and evicts back below it, so `n` can never reach 5743. Training never fires again.

The general form is more useful than the specific ladder: **any `trained_at > 3333` freezes training permanently**, because `trained_at * 1.5 > 5000` from there on.

## Why it wasn't visible before

`finish_batch()` is reachable only from `Router::new` (`router.rs:692`). Before #4808, `Ring::refresh_router` rebuilt the whole Router every 5 minutes, which constructed a fresh predictor and re-ran the batch training path. That masked the freeze.

This is a **pre-existing latent bug, not one #4808 introduces** — `routing_predictor.rs` isn't in that PR's diff. But #4808 removes the accidental workaround, so the freeze becomes reachable.

## Severity: low, but real

Reaching `trained_at > 3333` requires ~3,300+ observations in one session. At the observed production median (~14-16 routing events/hour) that is ~9 days of uninterrupted uptime, and nodes auto-update roughly daily. So it is unlikely to bite today — but it will on long-lived, high-traffic peers (gateways), which are exactly the ones with enough data for Renegade to matter.

Note the frozen model still *predicts*; it just stops re-tuning K and the metric. So the impact is degraded prediction quality, not incorrect behaviour.

## Suggested fix

Fix `should_train` rather than resurrecting the periodic rebuild. Options:
- Cap the growth requirement so it stays reachable under the observation cap (e.g. `n >= trained_at + min(trained_at/2, RENEGADE_MAX_OBSERVATIONS/4)`).
- Or trigger on observations-since-last-training rather than total growth, which is cap-independent.

Either should come with a test pinning that training remains reachable at saturation.

[AI-assisted - Claude]

Issue #4632, opened 2026-06-29
ci: rule-lint pathspec 'crates/core/src/**/*.rs' misses top-level src files (node.rs, ring.rs, config.rs...)


iduartgomez:
## Problem

The CI rule-lint that bans DST-breaking patterns in production code (`std::time::Instant::now()`, `rand::thread_rng()/random()`, `tokio::net::UdpSocket`, plus the test-removed check) restricts its diff to the pathspec **`crates/core/src/**/*.rs`**. In this repo's git (2.54.0, no `core.glob` override) that `**` glob does **not match files directly under `crates/core/src/`** — only files in subdirectories.

Verified: a commit touching `crates/core/src/node.rs` returns **0** matches under `crates/core/src/**/*.rs`.

**Consequence:** the ~20 top-level files in `crates/core/src/` — including the hottest production modules `node.rs`, `ring.rs`, `config.rs`, `contract.rs`, `router.rs`, `operations.rs`, `client_events.rs`, `topology.rs`, `transport.rs` — are a **blind spot**. A new prod `Instant::now()` / `thread_rng()` / `UdpSocket` added to any of them slips past rules #1/#2/#3 silently, and a deleted test in them slips past rule #4.

## Not a regression

This is **pre-existing**. The old inline-bash linter used the identical pathspec (ci.yml `ADDED=$(git diff ... -- 'crates/core/src/**/*.rs')` and the rule-#4 `REMOVED=` line). PR #4622 (which added the `#[cfg(test)]` exemption via `check_banned_patterns.py`) faithfully preserved the same pathspec, so #4622 neither introduced nor worsened this. Both the old and new linters share the gap.

## Current exposure (audited — no real violations hiding today)

A scan of the top-level `crates/core/src/*.rs` files for the banned patterns:
- `node.rs:4972-5036` — 4× `Instant::now()`, all inside `#[cfg(test)] mod tests` (starts line 3021) → legitimate test code; would be exempt under #4622's cfg(test) rule anyway.
- `operations.rs:688` — a comment mentioning the pattern, not code.
- `transport.rs:14, 1178, 1197` — **intentional** production low-level `tokio::net::UdpSocket` construction. This is the documented base the `Socket` trait wraps (see the comment at `transport.rs:212`: the lint exists to push *new* code through the `Socket` abstraction; this is the one legitimate construction site).

So nothing bad has slipped in yet — but the guard is not actually protecting these files.

## Fix

[...]

Issue #3506, opened 2026-03-10
fix: update test_get_retry_with_alternatives_sparse_topology seed for stdlib 0.2.2


iduartgomez:
## Problem

`test_get_retry_with_alternatives_sparse_topology` fails after upgrading `freenet-stdlib` from 0.1.40 to 0.2.2. The new `StreamChunk` variant in `ClientRequest` changes bincode's enum variant indices, which alters serialization sizes and shifts RNG-dependent topology formation under the test's fixed seed (`0xBEEF_CAFE_0001`).

With the 0.2.2 enum layout, the seed produces a topology where node-10 (and sometimes node-3) cannot reach any contract-caching node within HTL=2 hops.

## What needs to happen

Find a new seed that produces a topology where:
- PUT with HTL=2 caches at only 2-3 nodes (sparse)
- All 10 nodes can still reach at least one caching node via GET retry with alternatives

The GET retry logic itself is unchanged and working — `test_get_routing_coverage_low_htl` passes. This is purely a seed/topology issue.

## References

- Ignored in PR #3488 (delegate creation by delegates)
- Original test added in PR #3444 (GET retry with alternatives)
- Topology changes that also affected seeds: #3474, #3503

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:58:27.272921177+00:00
Issue hash: `77ef79c5`

## Value Assessment

The test is still marked `#[ignore]` with the old seed `0xBEEF_CAFE_0001`, confirming this issue is still open. The test was never fixed — it was just ignored in PR #3488 as a workaround. The issue is asking someone to find a new seed that produces the desired sparse topology under the current stdlib.

**Relevance Check:** This issue is still relevant. The test `test_get_retry_with_alternatives_sparse_topology` remains `#[ignore]`'d in `crates/core/tests/simulation_integration.rs` with the same stale seed (`0xBEEF_CAFE_0001`) and a comment referencing PR #3488. No one has done the seed-finding work described in the issue — the fix was deferred entirely. The stdlib has since moved well past 0.2.2 (current releases are in the 0.2.6x range), but the underlying problem is identical: the ignored test is dead weight that provides no coverage of the GET retry with alternatives path under a sparse topology.

[...]

PR #4821, opened 2026-07-15
refactor(router): refit isotonic estimators in add_event, not on a periodic tick

### LINKED ISSUE #4811: refactor(router): ConnectForwardEstimator is never refit; consider moving the refit into IsotonicEstimator::add_event
### 1. `ConnectForwardEstimator` never gets the accuracy correction

`operations/connect.rs:424-427` wraps an `IsotonicEstimator`, fed via `add_event` on every forward attempt (`:464`) and read by `estimate()` (`:467`) to bias live CONNECT forwarding.

It lives in `Arc<RwLock<ConnectForwardEstimator>>` **outside** `Router`, so `Router::refit_stale_estimators` (`router.rs`) cannot see it. #4808's entire drift argument applies to it verbatim: `pav_regression`'s incremental `add_points`/`remove_points` are approximations of a batch PAV fit, so its curve drifts and is never repaired.

**Not a regression** — the pre-#4808 rebuild replaced `Router` and never touched this estimator either. But the codebase now has one isotonic estimator that gets the correction and one that silently doesn't. Review noted its Bernoulli 0/1 outcomes pool heavily, so it may be the *worst* drift case in the tree.

### 2. The refit arguably belongs in `add_event`

#4808 put the refit on `Ring::refresh_router`'s 5-minute tick. Review argued (persuasively) it belongs in `IsotonicEstimator::add_event` instead:

- **The poll asks a question only `add_event` can answer.** `events_since_refit` is mutated by exactly one function; `add_event` is the only moment staleness can change. Polling to ask "did anything change?" is a pull where a push is natural.
- **Cost favours it.** `add_event` already pays O(k log k) **twice** per call — both `add_points` and `remove_points` clone, sort, and re-run PAV. One amortised O(n log n) refit per ~51 events is a small constant-factor bump on an already-superlinear path. (Measured: a saturated 500-point refit is ~27µs.)
- **Lock behaviour favours it.** The tick takes `router.write()` and refits up to 15 estimators in ONE hold (~410µs). Per-event refit touches at most one estimator per hold — a ~15× smaller peak critical section on the same lock.
- **It fixes gap 1 for free**, with zero new wiring.
- **It makes the cadence honest.** Today the realised cadence is `min(5-min tick, >10% turnover)`, so above ~10 events/min the tick binds and a window can turn over entirely between refits — precisely where drift is worst. (Far above the ~14-16 events/hour production median, so this is currently a documentation caveat rather than a live problem; `REFIT_STALENESS_PERCENT`'s rustdoc spells it out.)
- **It removes a node-fatal liability.** `refresh_router` is registered with `task_monitor`, so its death is node-fatal — and after #4808 plus the startup-restore deletion it does no I/O at all. Keeping node-fatal machinery alive to poll an in-memory counter is a poor trade.

Counterpoint: it moves a periodic latency spike onto the relay hot path. But the current design already puts a 15× larger spike under the same lock.

## Suggested sequencing

Land after the startup-restore deletion (the follow-up to #4808/#4809). At that point `refresh_router` has no remaining purpose, so folding the refit into `add_event` lets the task be deleted outright rather than renamed.

Any change here should keep #4809's `refresh_router_loop_refits_stale_estimators` equivalent — i.e. still pin that a refit actually happens, by whatever mechanism replaces the loop.

Follow-up to #4808 / #4809.

[AI-assisted - Claude]

sanity:
> [!IMPORTANT]
> **Simulation CI is RED** on `test_get_retry_with_alternatives_sparse_topology`, and it needs a decision before this merges. It is **not** the refit: on otherwise-pristine `main`, deleting only this PR's 2-line `task_monitor.register(...)` reverses it, and refit cadence vs. outcome is non-monotonic (never → FAIL, every 5 min → PASS, every ~51 events → FAIL). The test is explicitly seed-coupled and its own doc calls this class of break "expected maintenance rather than a regression". Full bisect and options in [this comment](https://github.com/freenet/freenet-core/pull/4821#issuecomment-4986191456). I have deliberately **not** re-seeded it unilaterally.

## Problem

Two gaps left by #4808/#4809, both about where the isotonic refit is triggered from.

### 1. `ConnectForwardEstimator` never got the refit — and that leaks memory

`operations/connect.rs` wraps an `IsotonicEstimator`, feeds it via `record` → `add_event` on every forward attempt, and reads it via `estimate()` to bias **live CONNECT forwarding**. It lives behind its own `Arc<RwLock<..>>` on **`OpManager`** (`node/op_state_manager.rs:152`, built once at `:415`) — **outside** `Router` — so `Router::refit_stale_estimators`, reachable only from `Ring::refit_router_periodically`'s tick, could not see it.

Note the owner, because it is the whole point: `ConnectOp` is a `#[cfg(test)]` fixture. In production this estimator is **node-lifetime**. (The issue and an earlier revision of this description both said `ConnectOp`; review caught it.)

Two consequences, not one:

1. **Drift, unrepaired forever.** #4808's argument applies verbatim: `pav_regression`'s incremental `add_points`/`remove_points` approximate a batch PAV fit, so its curve drifted with nothing to repair it while `estimate()` kept biasing live CONNECT forwarding off it. Its Bernoulli 0/1 outcomes pool heavily, making it plausibly the worst drift case in the tree.
2. **An unbounded, remote-peer-keyed map — i.e. a slow leak.** `add_event` only ever INSERTS into `peer_adjustments`; the ONLY thing that prunes it to the current window is a refit (`fit` rebuilds it from `raw_events`). `raw_events` is capped at `MAX_REGRESSION_POINTS`, but with no refit ever reaching this estimator the map was capped by **nothing**: it grew one entry per distinct peer ever seen in CONNECT forwarding, for the life of the node. That is exactly the unbounded remote-driven collection `.claude/rules/code-style.md` bans. Secondary cost: `connect_forward_estimator.read().clone()` on the CONNECT path (`connect/op_ctx_task.rs:754/1026/1177`) copies that map, so the leak made those clones monotonically more expensive with uptime.

This PR bounds the map to the window (plus at most one refit interval of new peers). Pinned empirically by `forward_estimator_bounds_peer_adjustments_to_its_window`: feed 2000 distinct peers, assert the map stays well under that — against the unfixed path it retains **1993 of 2000**.

Not a regression (the pre-#4808 rebuild never touched this estimator either), but the tree had one isotonic estimator getting the correction and one silently not.

### 2. The refit was polled, not pushed

`add_event` is the **only** writer of both staleness inputs (`events_since_refit` and `raw_events`), so it is the only moment staleness can change. Polling for a condition only one function can create is a pull where a push is natural — and the poll paid for it three times over: it could only refit **later** than the event that earned it (widening the drift window it existed to close), it could only reach estimators owned by `Router` (gap 1), and it kept a `task_monitor`-registered — hence **node-fatal** — task alive to read an in-memory counter.

## Solution

Move the trigger into `IsotonicEstimator::add_event`. Gap 1 falls out for free with zero wiring: every estimator now refits itself, whoever owns it.

`add_event` splits into two named halves — `add_event_incremental` (patch the fit in place; the half that drifts) and the `refit_if_stale()` that repairs it. The split is load-bearing for testing, not cosmetic: see Testing.

**Lock safety.** The **fit** takes no locks and does no I/O — pure computation over `&mut self` — so it cannot deadlock or re-enter a caller already holding one (`Router::add_event` runs under `ring.router.write()`; `ConnectForwardEstimator::record` under its own `RwLock`). It only extends a critical section the caller already held, and is deterministic given `raw_events`.

Scoped precisely (via review): "no locks" covers `fit`, **not** every line reachable from the refit. `refit`'s error arm calls `tracing::warn!`, whose release-build per-callsite rate limiter does a DashMap lookup — a real `parking_lot` shard guard, compiled out under `cfg(test)`, so no test would surface it. Not a deadlock risk (nothing under that guard reaches back to `ring.router` or the connect estimator, so there is no cycle) and inert today because the arm is unreachable. Both facts are recorded on the code, along with the note that a persistently-failing fit would now retry per-event rather than per-5-minutes.

**The trigger strictly dominates the tick.** Staleness becomes true only inside `add_event`, and `add_event` clears it before returning, so there is no state the tick would have caught that this misses. "An estimator is never stale once `add_event` returns" becomes an invariant rather than something a poller converges on.

### `refit_router_periodically` is genuinely dead — evidence

Verified, not assumed. Post-#4812 the loop body did exactly three things:

1. `router.write().refit_stale_estimators()` — now self-driving.
2. `BACKGROUND_TASK_HEALTH.record_refit_router_success()` — its own heartbeat.
3. A `tracing::debug!`.

With (1) gone its only remaining work is proving *itself* alive, which is self-referential. So it goes, and with it `Router::refit_stale_estimators` (no callers) and the `BackgroundTaskHealth` facility (one field, one recorder, called only from the loop).

**On deleting `refresh_router_last_success_age_secs`:** #4812 kept this gauge deliberately, because `BackgroundTaskMonitor` catches task *death* only, while a loop alive-but-starved shows up solely as a rising age. That rationale was entirely about the loop. With no loop there is nothing to starve — a stalled refit now means a stalled relay path, loudly visible in operation telemetry. Checked for external consumers first: `grep -rn refresh_router` over `/var/www/freenet-dashboard/` returns nothing, and there are no other in-repo references. The alternative was emitting a permanently-null field named after a task that no longer exists.

`Ring::new`'s "why there is no warm start from the event log" rationale lived on the deleted loop's rustdoc; it is **relocated** to the cold-router site, not lost.

## Testing

`cargo test -p freenet --lib`: **3993 passed, 0 failed**. `cargo fmt` clean. `cargo clippy --locked -- -D warnings` and the `trace-ot` variant both clean (the CI invocations; `--all-targets --all-features` fails on `main` too, for unrelated pre-existing lints).

**Simulation: RED** — see the banner above and the linked comment. `test_get_routing_coverage_low_htl` (the non-seed-coupled cover for the same GET retry logic, named in the failing test's own doc) passes on this branch. `simulation_smoke` passes 21/21.

**#4809's pin is kept, not dropped.** `refit_router_periodically_refits_stale_estimators` (itself the port of `refresh_router_loop_refits_stale_estimators`) pinned the one line that made #4809's fix do anything. Both the loop and `refit_stale_estimators` are gone, so the guarantee is re-pinned against the replacement mechanism in the same terms — "no estimator is left owing a refit" — as `router::tests::add_event_leaves_no_estimator_stale`. It covers all three global estimators plus every per-op estimator, and asserts the per-op maps are non-empty first so it cannot go vacuous.

**Gap 1 is pinned twice**, both failing against the unfixed behaviour: `forward_estimator_refits_itself_despite_living_outside_router` (the refit happens at all) and `forward_estimator_bounds_peer_adjustments_to_its_window` (the leak is bounded — 1993/2000 retained without the fix).

**Slow-event-rate estimators: no gap, and the natural guess is inverted.** The old tick did not refit on a timer — it called `refit_stale_estimators` → `refit_if_stale`, which returns early unless `refit_due()` (the turnover predicate) already holds. The tick was a *poll of the same condition*, so an estimator that had not earned turnover was skipped by it exactly as it is skipped now. The trigger condition is unchanged; only the earn-to-perform latency changed, from ≤5 minutes to zero. Nothing that used to be refit no longer is. This is recorded on `REFIT_STALENESS_PERCENT` rather than pinned as a test, since it is an argument about the trigger rather than a behaviour.

**Mutation-verified.** Stubbing the `refit_if_stale()` call in `add_event` fails exactly the four guards that pin the trigger — `add_event_leaves_no_estimator_stale`, `add_event_refits_inline_at_the_turnover_boundary`, `add_event_refits_inline_at_exactly_min_points`, `forward_estimator_refits_itself_despite_living_outside_router` — while the seven other refit guards stay green (they pin the refit *mechanism*, unchanged). So the new guards are the only thing pinning the trigger, exactly as #4809's loop test was.

Boundary coverage per `.claude/rules/testing.md`: 11/12 either side of `REFIT_STALENESS_PERCENT` (the true boundary — bracketing 10/20 leaves `>` vs `>=` undetected), exactly `MIN_POINTS_FOR_REGRESSION`, one below, and repeated `add_event` calls.

**Two tests had to switch to `add_event_incremental`, and this deserves reviewer attention.** `incremental_fit_drifts_from_batch_and_refit_repairs_it` (the tripwire proving the refit is necessary) and `refit_prunes_peers_that_fell_out_of_the_window` both need a genuinely refit-free fit to measure against. Through `add_event` they no longer can — it repairs drift as it goes. Feeding them through `add_event` would have silently gutted both. That is why the incremental half is a named function.

**Honest labelling:** `add_event_preserves_relay_recorded_events` and `add_event_preserves_every_observation_across_its_inline_refits` stay green under the mutation above. They guard the refit's no-data-loss property, not the trigger, and preservation is close to structural — carried forward cheaply, the same caveat #4809 recorded for their predecessor. The load-bearing guards are the four listed above.

Two deleted tests have no equivalent because their subject is gone: `refit_router_periodically_returns_promptly_on_shutdown` (#4278 — a loop that no longer exists cannot fail to exit promptly) and the two `background_task_health_tests` (they construct `BackgroundTaskHealth` directly). `test-exempt` justification: [comment](https://github.com/freenet/freenet-core/pull/4821#issuecomment-4985681705).

## Performance — measured, and the issue's figure is wrong

The issue's cost argument rests on "a saturated 500-point refit is ~27µs" (from #4809). **That does not reproduce.** Release build, saturated 500-point window:

| distinct peers in window | one refit | `add_event` (incremental) | `add_event` (+ inline refit) | amortised overhead |
|---|---|---|---|---|
| 20 (realistic) | ~150µs | ~39µs | ~41µs | **~1.5µs/event (+4%)** |
| 50 (realistic) | ~150µs | ~39µs | ~45µs | **~5.7µs/event (+14%)** |
| 500 (worst case) | ~275-337µs | ~39µs | ~43µs | ~4µs/event (+10%) |

A refit is ~150µs, not ~27µs — roughly **5x** the quoted figure (it scales with distinct peers in the window, since `fit` rebuilds `peer_adjustments`, cloning a `PeerKeyLocation` per peer). **The conclusion survives**: what reaches the hot path is the amortised few µs, not the whole fit. The measured numbers are now in the rustdoc so ~27µs stops propagating.

**The "~15x smaller peak critical section" claim needs refining too.** `IsotonicEstimator::add_event` refits at most one estimator, but `Router::add_event` fans out to up to **6** (3 global + 3 per-op), and their counters genuinely align (`response_start_time` and `failure` are both fed on every `Success`, sharing a window length). Honest bound: ~6 refits per `router.write()` hold vs. the tick's 15 — roughly **2x smaller, not 15x**. Measured on `Router::add_event` at saturation (20 peers): p50 ~340µs → ~350µs, p99 ~750µs → ~1.1ms.

**Pre-existing finding, not from this PR:** the worst-case single `Router::add_event` is **~183ms**, unchanged with my refit removed. It is `RoutingPredictor::record` → `should_train()` → `train()` running inline under `router.write()` on the relay hot path. Distinct from #4810 (that is about the training *freezing*, not its cost). Filed as **#4822** rather than fixed here, since `routing_predictor.rs` has in-flight work. It does put this PR's +350µs p99 in perspective.

## Risk

Routing/topology is a Full-tier surface. Checked against `.claude/rules/ring.md` and `bug-prevention-patterns.md`: no `send().await` in an event loop, no new spawn, no bounded-channel interaction — this PR *removes* a `GlobalExecutor::spawn` + `task_monitor` registration and adds only pure computation. No wire-format change.

Unrelated pre-existing flake found while testing and filed as **#4819** (`test_set_own_addr_atomic_with_network_status` races unlocked `set_own_addr` writers in `op_state_manager` tests; 2-in-14 runs on pristine `main`, never in isolation).

Closes #4811

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/connect.rs
+++ b/crates/core/src/operations/connect.rs
@@ -2128,6 +2128,42 @@ mod tests {
         estimator.record(&peer, Location::new(0.25), true);
     }
 
+    /// REGRESSION (#4811, gap 1): this estimator must get the batch-accuracy
+    /// refit like every other isotonic estimator in the tree.
+    ///
+    /// It never did. The refit was driven by `Router::refit_stale_estimators`,
+    /// reachable only from `Ring::refit_router_periodically`'s tick, and this
+    /// estimator lives OUTSIDE `Router` — behind its own `Arc<RwLock<..>>` on
+    /// `ConnectOp` — so the tick could not see it. Its incremental
+    /// pool-adjacent-violators fit therefore drifted from the batch fit forever,
+    /// with nothing to repair it, while `estimate()` kept biasing live CONNECT
+    /// forwarding off the drifted curve. Review noted its Bernoulli 0/1 outcomes
+    /// pool heavily, making it plausibly the worst drift case in the tree.
+    ///
+    /// Moving the trigger into `IsotonicEstimator::add_event` fixes it with no
+    /// wiring at all: `record` already funnels into `add_event`, so the estimator
+    /// now refits itself regardless of who owns it. That is the whole point of
+    /// the #4811 direction, and this is its pin — it fails against `main`, where
+    /// no amount of `record`ing ever refits this estimator.
+    #[test]
+    fn forward_estimator_refits_itself_despite_living_outside_router() {
+        let mut estimator = ConnectForwardEstimator::new();
+
+        // Well past the 10% turnover trigger. Distinct peers and desired
+        // locations, mixed Bernoulli outcomes — i.e. what `record` actually sees.
+        for i in 0..120u16 {
+            let peer = make_peer(3000 + i);
+            estimator.record(&peer, Location::new(f64::from(i % 50) / 100.0), i % 3 == 0);
+        }
+
+        assert!(
+            !estimator.estimator.is_stale_for_test(),
+            "ConnectForwardEstimator must not be left owing a refit: it is \
+             unreachable from Router, so if add_event does not refit it inline, \
+             nothing ever will and its curve drifts unrepaired forever (#4811)"
+        );
+    }
+
     #[test]
     fn expired_forward_attempts_are_cleared() {
         let mut op = ConnectOp::new_joiner(
--- a/crates/core/src/operations/connect.rs
+++ b/crates/core/src/operations/connect.rs
@@ -2134,17 +2134,35 @@ mod tests {
     /// It never did. The refit was driven by `Router::refit_stale_estimators`,
     /// reachable only from `Ring::refit_router_periodically`'s tick, and this
     /// estimator lives OUTSIDE `Router` — behind its own `Arc<RwLock<..>>` on
-    /// `ConnectOp` — so the tick could not see it. Its incremental
-    /// pool-adjacent-violators fit therefore drifted from the batch fit forever,
-    /// with nothing to repair it, while `estimate()` kept biasing live CONNECT
-    /// forwarding off the drifted curve. Review noted its Bernoulli 0/1 outcomes
-    /// pool heavily, making it plausibly the worst drift case in the tree.
+    /// **`OpManager`** (`node/op_state_manager.rs`, built once in `OpManager::new`),
+    /// so the tick could not see it. Note the owner: `ConnectOp` below is only a
+    /// `#[cfg(test)]` fixture. In production this estimator is **node-lifetime**,
+    /// which is what turns the missing refit from a nuisance into a leak.
     ///
-    /// Moving the trigger into `IsotonicEstimator::add_event` fixes it with no
+    /// Two consequences, both fixed by refitting on the write path:
+    ///
+    /// 1. **Drift, unrepaired forever.** The incremental pool-adjacent-violators
+    ///    fit is an approximation of a batch PAV pass, so its curve drifted with
+    ///    nothing to correct it, while `estimate()` kept biasing live CONNECT
+    ///    forwarding off the drifted curve. Its Bernoulli 0/1 outcomes pool
+    ///    heavily, making it plausibly the worst drift case in the tree.
+    /// 2. **An unbounded, remote-peer-keyed map.** `peer_adjustments` only ever
+    ///    INSERTS on the incremental path (`add_event`); the ONLY thing that
+    ///    prunes it to the current window is a refit (see `IsotonicEstimator::fit`).
+    ///    `raw_events` is capped at `MAX_REGRESSION_POINTS`, but with no refit the
+    ///    map was not capped by anything: it grew one entry per distinct peer ever
+    ///    seen in CONNECT forwarding, for the node's whole life — the exact hazard
+    ///    `.claude/rules/code-style.md` bans, driven by remote peers. It also made
+    ///    the `connect_forward_estimator.read().clone()` calls on the CONNECT path
+    ///    (`connect/op_ctx_task.rs`) monotonically more expensive with uptime,
+    ///    since the clone copies that map.
+    ///
+    /// Moving the trigger into `IsotonicEstimator::add_event` fixes both with no
     /// wiring at all: `record` already funnels into `add_event`, so the estimator
-    /// now refits itself regardless of who owns it. That is the whole point of
-    /// the #4811 direction, and this is its pin — it fails against `main`, where
-    /// no amount of `record`ing ever refits this estimator.
+    /// now refits itself regardless of who owns it, which bounds `peer_adjustments`
+    /// to the window (plus at most one refit interval of new peers). That is the
+    /// whole point of the #4811 direction, and this is its pin — it fails against
+    /// `main`, where no amount of `record`ing ever refits this estimator.
     #[test]
     fn forward_estimator_refits_itself_despite_living_outside_router() {
         let mut estimator = ConnectForwardEstimator::new();
@@ -2161,6 +2179,58 @@ mod tests {
             "ConnectForwardEstimator must not be left owing a refit: it is \
              unreachable from Router, so if add_event does not refit it inline, \
              nothing ever will and its curve drifts unrepaired forever (#4811)"
+        );
+    }
+
+    /// REGRESSION (#4811, gap 1 — the consequence that actually bites): the
+    /// refit is the ONLY thing that bounds this estimator's `peer_adjustments`.
+    ///
+    /// `IsotonicEstimator::add_event` only ever INSERTS into `peer_adjustments`;
+    /// nothing evicts from it. `raw_events` is capped at `MAX_REGRESSION_POINTS`,
+    /// but the map is pruned to the current window only by a refit
+    /// (`IsotonicEstimator::fit` rebuilds it from `raw_events`). Because nothing
+    /// refit THIS estimator, and because it hangs off `OpManager` and so lives as
+    /// long as the node, the map grew one entry per distinct peer ever seen in
+    /// CONNECT forwarding, forever — an unbounded collection keyed by remote
+    /// peers, which `.claude/rules/code-style.md` bans outright. It also made the
+    /// `connect_forward_estimator.read().clone()` calls on the CONNECT path
+    /// (`connect/op_ctx_task.rs`) grow more expensive with uptime.
+    ///
+    /// Feeding many more DISTINCT peers than the window can hold is the shape of
+    /// the leak: a long-lived node forwards CONNECTs to far more peers than fit
+    /// in 500 events. Against `main` this map would hold every one of them.
+    #[test]
+    fn forward_estimator_bounds_peer_adjustments_to_its_window() {
+        let mut estimator = ConnectForwardEstimator::new();
+
+        // Far more DISTINCT peers than the event window can hold, which is the
+        // shape of the leak: a long-lived node forwards CONNECTs to far more
+        // peers than fit in one window. Asserted as a RATIO rather than against
+        // `MAX_REGRESSION_POINTS` — that constant is private to the estimator's
+        // module, and a ratio does not rot if the window is ever resized.
+        let distinct_peers: usize = 2000;
+        for i in 0..distinct_peers {
+            let peer = make_peer(10_000 + u16::try_from(i).expect("port fits u16"));
+            estimator.record(&peer, Location::new((i % 50) as f64 / 100.0), i % 3 == 0);
+        }
+
+        let adjustments = estimator.estimator.peer_adjustments.len();
+        assert!(
+            adjustments < distinct_peers / 2,
+            "peer_adjustments must stay bounded by the event window, not by how \
+             many peers this node has ever forwarded to. The refit is the only \
+             thing that prunes this map, and the estimator is node-lifetime, so \
+             without an inline refit it retains every peer ever seen: got \
+             {adjustments} entries after {distinct_peers} distinct peers — on \
+             `main` this is {distinct_peers} (#4811)"
+        );
+        // Guard against passing for the wrong reason: an estimator that recorded
+        // nothing, or one that pruned everything, would also be "bounded".
+        assert!(
+            adjustments > 100,
+            "sanity: the estimator must still hold a full window's worth of peers \
+             (got {adjustments}); a near-empty map would satisfy the bound above \
+             while meaning the estimator had stopped tracking anything"
         );
     }
 
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -5,7 +5,7 @@
 
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::net::SocketAddr;
-use std::sync::{Arc, Weak, atomic::AtomicU64};
+use std::sync::{Arc, Weak};
 use std::time::Duration;
 // Intentional wall-clock type for suspend/resume detection in
 // `connection_maintenance`. See `classify_suspend_jump` for why the DST
@@ -505,22 +505,53 @@ impl Ring {
         };
 
         // Single shutdown signal for every long-lived background task spawned
-        // below. Created up front so `refit_router_periodically` (spawned before
-        // the `Ring` struct literal exists) shares the same token that gets moved
-        // into the struct. See issue #4278.
+        // below, moved into the struct literal at the end. See issue #4278.
         let shutdown = CancellationToken::new();
 
-        // Starts cold and learns from this session only — there is no warm start
-        // from the event log (see `refit_router_periodically`).
+        // The router starts COLD and learns purely from this session's
+        // observations. There is deliberately no warm start from the on-disk
+        // event log, and the reader that used to attempt one was deleted in the
+        // #4808 follow-up (#4812). Three reasons it is not worth resurrecting:
+        //
+        // 1. It effectively never fired in production. The restore read filtered
+        //    records to `record_ts >= NEW_RECORDS_TS`. `NEW_RECORDS_TS` is a
+        //    *process-global* `OnceLock` (`tracing::register`), stamped
+        //    `SystemTime::now()` at the FIRST event-register construction in the
+        //    process — `EventRegister::new`, or `OTEventRegister::new` under
+        //    `trace-ot`. A production node builds its register (`node.rs`) before
+        //    it can write a single route event and before `OpManager::new` ->
+        //    `Ring::new` ran, so the cutoff was always >= that process's start and
+        //    every prior-session record was discarded. Dead since #3672
+        //    (2026-03-27).
+        //
+        //    The invariant is "effectively never in production", NOT "structurally
+        //    unreachable" — do not tighten this claim. Two ways the branch can be
+        //    reached: both sides truncate to whole seconds, so a prior-session
+        //    record written in the same wall-clock second as the stamp passes; and
+        //    an in-process multi-node test that shares one `data_dir` (see
+        //    `tests/in_process_restart.rs`) makes the second node's `get_or_init` a
+        //    no-op, so it would read the first node's records. Neither happens on a
+        //    real node. Corroborated in production: across 8 gateway startups
+        //    captured in vega's retained logs, the INFO "Restored routing history
+        //    from event log" line and its WARN error arm each appear ZERO times,
+        //    while ~141k other INFO lines from `freenet::ring` are captured — i.e.
+        //    the read always returned `Ok(empty)`, exactly as the filter predicts.
+        // 2. Half of the state is meaningless across a restart anyway. The
+        //    per-peer `peer_adjustments` are keyed to a neighbour set that CONNECT
+        //    rebuilds differently on every start.
+        // 3. A partial restore is what caused #4808. Reloading a log that holds
+        //    only originator events (relay hops never persist) discards everything
+        //    the live router learned by relaying.
+        //
+        // The cost of starting cold is bounded: post-#4809 a peer reaches
+        // `MIN_EVENTS_FOR_PREDICTION` in a few hours at observed event rates.
+        //
+        // The estimators keep themselves batch-accurate: `IsotonicEstimator::add_event`
+        // refits inline once its window has turned over (#4811). There is no periodic
+        // refit task — `refit_router_periodically` was deleted with this change,
+        // because polling was the only thing it did.
         let router = Arc::new(RwLock::new(Router::new(&[])));
         crate::node::network_status::set_router(router.clone());
-        task_monitor.register(
-            "refit_router_periodically",
-            GlobalExecutor::spawn(Self::refit_router_periodically(
-                router.clone(),
-                shutdown.clone(),
-            )),
-        );
 
         // Interval for topology snapshot registration (1 second in test mode)
         // Registers subscription topology with the global registry for validation
@@ -1505,94 +1536,6 @@ impl Ring {
         self.event_register.register_events(events).await;
     }
 
-    /// Periodically refit the router's isotonic estimators in place, from their
-    /// own in-memory raw events. Never reads the event log — see the loop body.
-    ///
-    /// The node starts with a cold router (`Router::new(&[])`) and learns purely
-    /// from this session's observations. There is deliberately no warm start from
-    /// the on-disk event log, and the reader that used to attempt one was deleted
-    /// (#4808 follow-up). Three reasons it is not worth resurrecting:
-    ///
-    /// 1. **It effectively never fired in production.** The restore read filtered
-    ///    records to `record_ts >= NEW_RECORDS_TS`. `NEW_RECORDS_TS` is a
-    ///    *process-global* `OnceLock` (`tracing::register`), stamped
-    ///    `SystemTime::now()` at the FIRST event-register construction in the
-    ///    process — `EventRegister::new`, or `OTEventRegister::new` under
-    ///    `trace-ot`. A production node builds its register (`node.rs`) before it
-    ///    can write a single route event and before `OpManager::new` -> `Ring::new`
-    ///    spawns this task, so the cutoff was always >= that process's start and
-    ///    every prior-session record was discarded. Dead since #3672 (2026-03-27).
-    ///
-    ///    Note the invariant is "effectively never in production", NOT "structurally
-    ///    unreachable" — do not tighten this claim. Two ways the branch can be
-    ///    reached: both sides truncate to whole seconds, so a prior-session record
-    ///    written in the same wall-clock second as the stamp passes; and an
-    ///    in-process multi-node test that shares one `data_dir` (see
-    ///    `tests/in_process_restart.rs`) makes the second node's `get_or_init` a
-    ///    no-op, so it would read the first node's records. Neither happens on a
-    ///    real node. Corroborated in production: across 8 gateway startups captured
-    ///    in vega's retained logs, the INFO "Restored routing history from event
-    ///    log" line and its WARN error arm each appear ZERO times, while ~141k
-    ///    other INFO lines from `freenet::ring` are captured — i.e. the read always
-    ///    returned `Ok(empty)`, exactly as the filter predicts.
-    /// 2. **Half of the state is meaningless across a restart anyway.** The
-    ///    per-peer `peer_adjustments` are keyed to a neighbour set that CONNECT
-    ///    rebuilds differently on every start.
-    /// 3. **A partial restore is what caused #4808.** Reloading a log that holds
-    ///    only originator events (relay hops never persist) discards everything
-    ///    the live router learned by relaying.
-    ///
-    /// The cost of starting cold is bounded: post-#4809 a peer reaches
-    /// `MIN_EVENTS_FOR_PREDICTION` in a few hours at observed event rates.
-    async fn refit_router_periodically(router: Arc<RwLock<Router>>, shutdown: CancellationToken) {
-        let mut interval = tokio::time::interval(Duration::from_secs(60 * 5));
-        interval.tick().await;
-        loop {
-            if sleep_or_shutdown(&shutdown, async {
-                interval.tick().await;
-            })
-            .await
-            {
-                return;
-            }
-            // Refit the isotonic estimators in place from their own in-memory
-            // raw events. Each estimator only refits once >10% of its window has
-            // turned over, so an idle router does no work here.
-            //
-            // This deliberately does NOT re-read the event log. Until #4808 this
-            // loop did `*router.write() = Router::new(&history)`, rebuilding the
-            // whole router from disk every 5 minutes. That was destructive:
-            // `operations::record_relay_route_event` feeds the in-memory router
-            // ONLY — it never persists — so every relay-recorded event was
-            // discarded on each rebuild, giving relay events a <=5 minute
-            // half-life and resetting the model faster than it could reach
-            // `MIN_EVENTS_FOR_PREDICTION`. Observed in production as a peer's
-            // event count collapsing 29 -> 2 within one 30s window, and
-            // fleet-wide as 271,518 events lost to resets against 261,710
-            // gained (net zero accumulation).
-            //
-            // Within a session the on-disk log is a strict SUBSET of what the
-            // live router already knows (same originator events, minus every
-            // relay event), so rebuilding from it could only ever lose
-            // information. The batch-accuracy benefit that motivated the rebuild
-            // is preserved — and made complete — by refitting from `raw_events`,
-            // which holds every observation regardless of origin.
-            //
-            // Liveness (#4440): the task now has no fallible step, so there is no
-            // failure arm to record (the `consecutive_failures` gauge went with the
-            // startup read that was its only failure source). The success stamp
-            // stays as a pure heartbeat: `BackgroundTaskMonitor` only watches for
-            // task *death*, so a loop that is alive but starved — deadlocked on
-            // `router.write()`, or never scheduled — is invisible to it and shows
-            // up only as a rising `last_success_age_secs`.
-            let refit_count = router.write().refit_stale_estimators();
-            BACKGROUND_TASK_HEALTH.record_refit_router_success();
-            if refit_count > 0 {
-                tracing::debug!(refit_count, "Refit stale isotonic estimators");
-            }
-        }
-    }
-
     /// Periodically emit a router model snapshot as an EventKind::RouterSnapshot event.
     ///
     /// This captures the isotonic regression curves and model state, including the
@@ -1935,14 +1878,6 @@ impl Ring {
             snapshot.broadcast_stream_failures_last_snapshot =
                 Some(broadcast_stream_failures_delta);
 
-            // Liveness heartbeat (#4440) for the periodic refit task. The
-            // `BackgroundTaskMonitor` catches it dying; this catches it being
-            // alive but starved. Its companion `consecutive_failures` gauge went
-            // with the startup restore that was its only failure source (#4808
-            // follow-up).
-            let refit_router_age = BACKGROUND_TASK_HEALTH.refit_router_last_success_age_secs();
-            snapshot.refresh_router_last_success_age_secs = refit_router_age;
-
             // Placement-quality gauge (#4404 follow-up): host-to-hosted-key
             // ring-distance distribution. If the SubscribeHint placement
             // migration is working, hosting drifts toward each contract's key,
@@ -2012,7 +1947,6 @@ impl Ring {
                 broadcast_stream_attempts_total = bs.streaming_attempts_total,
                 broadcast_stream_failures_total = bs.streaming_failures_total,
                 broadcast_stream_failures_last_snapshot = broadcast_stream_failures_delta,
-                refresh_router_last_success_age_secs = ?refit_router_age,
                 hosted_contracts_count = ?snapshot.hosted_contracts_count,
                 hosted_key_distance_median = ?snapshot.hosted_key_distance_median,
                 hosted_key_distance_p90 = ?snapshot.hosted_key_distance_p90,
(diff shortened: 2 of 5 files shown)


sanity:
## `test-exempt` justification (rule #4: deleted tests)

Rule Lint counts **4 removed** test attributes against **1 added**. Per-test justification below. **None could be `#[ignore]`d** — an `#[ignore]`d test must still compile, and each references a type or function this PR deletes.

| Deleted test | References | Equivalent kept? |
|---|---|---|
| `refit_router_periodically_refits_stale_estimators` | `Ring::refit_router_periodically`, `Router::refit_stale_estimators` | **YES — ported.** Re-pinned as `router::tests::add_event_leaves_no_estimator_stale` against the mechanism that replaces the loop, in the same terms ("no estimator left owing a refit"). Mutation-verified: stubbing the `refit_if_stale()` call in `add_event` fails it. |
| `refit_router_periodically_returns_promptly_on_shutdown` | `Ring::refit_router_periodically` | **No — subject deleted.** It guarded #4278 (the loop must observe the shutdown token rather than park on its 5-minute tick). With no loop, there is nothing to exit promptly; the guarantee is moot, not unguarded. |
| `never_succeeded_reports_none_age` | `BackgroundTaskHealth` | **No — subject deleted.** Constructs the type directly. |
| `success_stamps_age_which_grows_with_time` | `BackgroundTaskHealth` | **No — subject deleted.** Constructs the type directly. |

The `BackgroundTaskHealth` facility is deleted because its only recorder (`record_refit_router_success`) was called only from the deleted loop, leaving `refresh_router_last_success_age_secs` pinned at `None` forever. Same shape as #4812's `consecutive_failures` removal, and #4812 used `test-exempt` for the same reason.

**Net test coverage goes UP, not down.** This PR adds two genuinely new guards — `add_event_leaves_no_estimator_stale` and `forward_estimator_refits_itself_despite_living_outside_router` (the #4811 gap-1 pin, which fails against `main`) — plus tightened boundary guards. Full suite: 3993 passed, 0 failed.

---

### Aside: the linter undercounts added tests (pre-existing, fail-safe)

Not blocking, and not caused by this PR, but worth recording. In `ci.yml`, `REMOVED` globs both forms:

```
git diff ... -- 'crates/core/src/*.rs' 'crates/core/src/**/*.rs' 'crates/core/tests/**/*.rs'
```

while `ALL_ADDED` globs only `'**/*.rs'`. Per the comment already in that file (#4632): *"git's `**` does NOT match files directly under `crates/core/src/`"*. So a test added to a **top-level** file like `crates/core/src/router.rs` is not counted as added, while its removal *is* counted.

Concretely here: this PR's new `add_event_leaves_no_estimator_stale` lives in `crates/core/src/router.rs` and scores zero; only the `operations/connect.rs` one counts. The #4632 fix was applied to `REMOVED` but not to `ALL_ADDED`.

It is **fail-safe** (it under-counts additions, so it over-rejects rather than letting deletions through), which is why it has gone unnoticed. But it also makes rule #5 wrongly reject any `fix:` PR whose only regression test lands in `router.rs`, `node.rs`, `ring.rs`, etc. Adding `'crates/core/src/*.rs'` to the `ALL_ADDED` glob would fix it. Flagging rather than fixing here to keep this PR one logical change.

[AI-assisted - Claude]

sanity:
## `test-exempt` justification (required by the label)

The label is applied because this PR removes 4 test functions and adds 1. Per-test justification, since "removed 4 tests" deserves scrutiny rather than a blanket wave-through:

**1. `refit_router_periodically_refits_stale_estimators`** — this is #4809's pin, and #4811 explicitly required keeping it equivalent. **Its guarantee is NOT dropped: it is re-pinned as `router.rs:1663 add_event_leaves_no_estimator_stale`**, in the same terms ("no estimator left owing a refit"), covering all 3 global estimators plus every per-op estimator, and asserting the per-op maps are non-empty first so it cannot pass vacuously. The mechanism moved from a polled tick to the write path, so the test moved with it. This is the one deletion that would have been unacceptable to wave through, and it wasn't waved through.

**2. `refit_router_periodically_returns_promptly_on_shutdown`** (#4278) — subject deleted with the task. Checked that this doesn't orphan the pattern: `sleep_or_shutdown` still has 5+ live callers and is covered generically by `ring.rs:8168 mod sleep_or_shutdown_tests`. Not a coverage hole.

**3/4. `background_task_health_tests` ×2** (`never_succeeded_reports_none_age`, `success_stamps_age_which_grows_with_time`) — subject deleted: `BackgroundTaskHealth` and the `refresh_router_last_success_age_secs` gauge existed only to watch the loop this PR removes. Producer, struct field, JSON mirror, and pin test are removed together, so no half-removed counter rots (the failure mode `.claude/rules/bug-prevention-patterns.md` warns about is the opposite — a gauge still emitted while its update path dies). The gauge was reaching the collector but has no consumer: `/var/www/freenet-dashboard` has zero references to `refresh_router`, `RouterSnapshot`, or any sibling gauge.

Two further tests switched from `add_event` to `add_event_incremental` rather than being weakened — both still call `refit()` and assert the repair/pruning. The switch is necessary: `add_event` now repairs drift inline, so feeding through it would silently gut both (the drift tripwire would measure zero drift; the pruning test would never accumulate).

These deletions are the "retiring a data structure whose unit tests can't be `#[ignore]`d because the type itself is gone" case the label exists for.

[AI-assisted - Claude]

sanity:
## Correction: the proposed Rule Lint glob fix is a no-op — don't land it

The earlier comment attributing the ADDED undercount to `ALL_ADDED`'s `'**/*.rs'` glob missing top-level files is **wrong**, and the proposed fix was reproduced and changes nothing (ADDED stays 1). Recording it here so nobody spends time on it:

- `'**/*.rs'` **does** match `crates/core/src/router.rs`. Only the *anchored* `'crates/core/src/**/*.rs'` misses top-level files — that's the `REMOVED` glob and what the #4632 comment refers to, not `ALL_ADDED`.
- The actual blind spot is **rename / rewrite-in-place**: rule #4 counts test *attribute* lines, and rewriting a test over an existing `#[test]` leaves that attribute as unchanged diff **context**, so it never lands on a `+` line. All 8 renames here are invisible to it for that reason.

Filed as #4825 (with the suggested real fix: compare test *function names*, not attribute lines).

## For the record: the true test delta on this PR

Against merge-base `95a3746a`, per file:

| file | removed | added |
|---|---|---|
| `router.rs` | 2 | 2 |
| `router/isotonic_estimator.rs` | 5 | 5 |
| `operations/connect.rs` | 0 | 1 |
| `ring.rs` | 4 | 0 |
| `tracing/telemetry.rs` | 1 | 1 |

**12 removed, 9 added, net -3** — four dead-subject retirements against one brand-new test. The linter's "4 removed / 1 added" is `ring.rs`'s 4 removals plus `connect.rs`'s 1 addition; the rest are renames it cannot see.

## Two review notes worth recording

**The #4809 port is stronger than what it replaces, and provably non-vacuous.** `events_since_refit = 0` is assigned in exactly two places (`isotonic_estimator.rs:212` constructor, `:350` inside `refit()`'s `Ok` arm — the `Err` arm deliberately does not reset). So `!is_stale_for_test()` implies a `refit()` ran and returned `Ok`; it cannot be satisfied by bumping a counter. The old test asserted one summed count was zero; `add_event_leaves_no_estimator_stale` names each per-op estimator.

**The "slow-rate estimator loses its 5-minute repair" concern is unfounded** — worth stating since it's the obvious objection to this change. The tick never refit unconditionally: `refit_if_stale()` has always been gated by `refit_due()` (`isotonic_estimator.rs:308`), so the old design *also* required >10% turnover. More fundamentally, drift is produced by `add_points`/`remove_points`, which run only inside `add_event_incremental` — an estimator receiving no events has a frozen fit and no drift to repair. The new trigger produces identical staleness episodes, each repaired at the event that earns it rather than 0-5 minutes later.

## Non-blocking nits (deliberately not addressed here, to avoid restaling a green CI run)

1. `router.rs:1653` — the non-vacuity guard (`!per_op.is_empty()`) is weaker than the one it ports (which asserted the loop had refits owed). `refit_due()` returns false either because a refit happened *or* because there are <5 points, so this could go silently vacuous if the seed ever spreads thinly across op types. One line would fix it: assert `est.len() >= MIN_POINTS_FOR_REGRESSION` in the loop. Fine today.
2. `connect.rs:2124` — `forward_estimator_handles_missing_location` is itself vacuous (pre-existing): it builds its peer via `PeerKeyLocation::new`, which can never yield `Unknown`, so the `location().is_none()` early-return at `:454` is untested. Now adjacent, since the new gap-1 test depends on the opposite property. `with_unknown_addr` (`peer_key_location.rs:103`) would fix it.

[AI-assisted - Claude]

sanity:
## Simulation is RED: `test_get_retry_with_alternatives_sparse_topology` — diagnosed, needs a decision

CI Simulation fails (3/3 nextest retries, deterministic):

```
GET retry regression: 2 of 10 nodes failed to get contract state
(contract cached at only ~2-3 nodes due to HTL=2, retry should find it).
Failed nodes: ["node-1", "node-2"]. See PR #3444.
```

Reproduced locally. **This PR causes it** — I am not claiming otherwise. But the cause is not the refit, and the evidence says it is a re-seed, not a routing regression. Posting the full bisect rather than asserting a conclusion.

### Bisect

| Build | Result |
|---|---|
| pristine `origin/main` @ 95a3746a | **PASS** (4/4 local runs) |
| this PR, full | FAIL |
| this PR, `refit_if_stale()` removed from `add_event` | FAIL |
| this PR, refit rebuilding the global curve but preserving `peer_adjustments` | FAIL |
| **`origin/main` + ONLY the 2-line `task_monitor.register(...)` deletion, nothing else** | **FAIL** |

The last row is decisive: on otherwise-pristine main, deleting only the task registration flips the test. **The refit — the actual subject of this PR — is not involved at all.**

### Why this is topology luck, not routing quality

Refit cadence versus outcome is **non-monotonic**:

| refit cadence | result |
|---|---|
| every 5 min (main, task present) | PASS |
| **never** (main − task) | **FAIL** |
| every ~51 events (this PR) | **FAIL** |

If refit frequency were a routing-quality dial, "never" and "more often" could not *both* fail while "sometimes" passes. The test is keyed to one specific topology, not to model accuracy.

Corroborating:

1. **The test says so itself.** It was `#[ignore]`d in #3488 when a freenet-stdlib bump "shifted RNG-dependent topology formation", re-seeded in #3506 to `SEED = 0x3506_000B_0001` — chosen so the HTL=2 PUT caches at *exactly 2 of 10* nodes. Its doc: *"Like every seed-coupled topology test, a future stdlib/topology change could perturb the RNG stream again and require another re-seed; **that is expected maintenance rather than a regression in the GET retry logic itself**."*
2. **The non-brittle twin passes.** `test_get_routing_coverage_low_htl` — named in that same doc as the other cover for this logic — passes on this branch.
3. **The perturbation is semantically inert.** The deleted task ticks every 5 min; these sims finish in well under that of wall-clock, and its removal cannot change routing on its own. It moves the scheduler, and the scheduler moves the seeded RNG stream.

### The decision

I am **not** re-seeding unilaterally. The risk is real: re-seeding to a topology where the PUT happens to cache at more nodes would make every GET trivially succeed and silently gut the retry path this test exists to guard (`#3506` verified the 2-of-10 premise with a PUT-only run precisely to avoid that). A re-seed must re-verify that premise, not just go green.

Options:

1. **Re-seed** it, re-verifying the "PUT caches at exactly 2-3 of 10" premise with a PUT-only run, as #3506 did. Follows the test's documented maintenance path; costs a seed search.
2. **Keep `refit_router_periodically`** and take only the `ConnectForwardEstimator` fix (gap 1) from this PR. Avoids the perturbation, but retains a node-fatal task whose only remaining work is recording its own heartbeat — which is what #4811 asks to delete.
3. Something else — e.g. decouple the test from scheduler-order luck, which is a bigger job and its own PR.

My recommendation is **(1)**, with the premise re-verified. Flagging for @sanity since this is a Full-tier routing surface and the test's brittleness is a known, documented trade.

[AI-assisted - Claude]

sanity:
Heads-up on an overlap, and one substantive interaction worth a look.

A dashboard audit (#4829) found that the peer-detail page renders `Adjustment::event_count()` as "This peer: ... N events", but that value is an EWMA **effective sample size**, not an event count:

- `Adjustment::add` computes `effective_count = 1.0 + (1.0 - alpha) * effective_count` with `EWMA_ALPHA = 0.1`, a fixed point of **exactly 10.0** — it can never exceed 10.
- `IsotonicEstimator::fit` seeds `effective_count = ADJUSTMENT_PRIOR_SIZE` (= 10) and `refit()` replaces `peer_adjustments` wholesale.

**The interaction with this PR:** my write-up of that finding attributes the pinning partly to `refit_router_periodically` reseeding to the prior every 5 minutes — the exact tick this PR removes. Two consequences:

1. The mechanism in #4829 goes stale the moment this lands. The defect itself looks like it survives regardless (the fixed point and the prior seed are both independent of *when* refit runs), but moving the reseed into `add_event` plausibly makes the displayed value **stickier**, not less — worth a glance while you're in there, since it's adjacent to what you're already changing.
2. My planned fix for #4829 touches `router.rs` and `router/isotonic_estimator.rs`, i.e. exactly your two files.

So I'm **holding #4829's fix** rather than opening a PR against code you're rewriting. Once this merges I'll rebase and re-verify the finding against the new refit path before writing anything.

No action needed unless you'd rather I stack on this branch — say the word and I will.

[AI-assisted - Claude]

sanity:
## Rule Lint: state after rebase + the new test (REMOVED=4, ADDED=2)

Recounted with the workflow's own logic on the rebased head:

```
REMOVED=4  ADDED=2   # label test-exempt applied
```

Still 4 > 2, so the label is still required — but the count now includes a **second genuinely new test**, and the per-test position below is unchanged from the earlier justification comments ([1](https://github.com/freenet/freenet-core/pull/4821#issuecomment-4985681705), [2](https://github.com/freenet/freenet-core/pull/4821#issuecomment-4986...)). Consolidating here since there were two.

### The 4 deletions

| Deleted | Subject | Equivalent |
|---|---|---|
| `refit_router_periodically_refits_stale_estimators` | `Ring::refit_router_periodically`, `Router::refit_stale_estimators` | **PORTED — `router.rs` `add_event_leaves_no_estimator_stale`** (genuinely new; absent at merge-base). Pins #4809's guarantee in the same terms ("no estimator left owing a refit"), across all 3 global estimators + every per-op one, with a non-empty precondition so it cannot go vacuous. Mutation-verified. |
| `refit_router_periodically_returns_promptly_on_shutdown` | the deleted loop | Obsolete — no loop, no shutdown path. The `sleep_or_shutdown` pattern it exercised is still covered generically by `ring.rs mod sleep_or_shutdown_tests`, and `sleep_or_shutdown` keeps 5+ live callers, so nothing is orphaned. |
| `never_succeeded_reports_none_age` | `BackgroundTaskHealth` | Obsolete — constructs the deleted type directly. |
| `success_stamps_age_which_grows_with_time` | `BackgroundTaskHealth` | Obsolete — constructs the deleted type directly. |

**Nothing still reads the deleted counters.** `record_refit_router_success` had exactly one caller (the deleted loop); `refresh_router_last_success_age_secs` had one reader (the snapshot task), now removed along with its `json!` key and the `RouterSnapshotInfo` field. No `network_status` provider, no `snapshot()` path, and no dashboard reference (`grep -rn refresh_router /var/www/freenet-dashboard/` → nothing). This is the "manually-mirrored telemetry counter" pattern in reverse: the counter's only source is gone, so leaving the mirror would emit a permanently-null field named after a task that no longer exists.

### The 2 additions

1. `add_event_leaves_no_estimator_stale` (`router.rs`) — the #4809 port above.
2. `forward_estimator_bounds_peer_adjustments_to_its_window` (`operations/connect.rs`) — **new**, see the leak below.

Both fail against the unfixed behaviour. Note the counter reads ADDED=2, not 4, partly because of a **linter bug**: `ALL_ADDED` globs only `**/*.rs`, which per the `#4632` comment already in `ci.yml` does *not* match top-level `crates/core/src/*.rs` — so `add_event_leaves_no_estimator_stale` scores zero purely for living in `router.rs`. `REMOVED` globs both forms. Fail-safe (it over-rejects), so not urgent, but it also makes rule #5 wrongly reject any `fix:` PR whose only regression test lands in `router.rs`/`node.rs`/`ring.rs`.

## Gap 1 is a memory leak, not just drift — thanks to review for catching the ownership error

`ConnectOp` is a `#[cfg(test)]` fixture. Production ownership is **`OpManager.connect_forward_estimator`** (`node/op_state_manager.rs:152`, built once at `:415`) — i.e. **node-lifetime**. The issue and my original PR body both got this wrong.

`IsotonicEstimator::add_event` only ever INSERTS into `peer_adjustments`; the only thing that prunes it to the window is a refit. `raw_events` is capped at `MAX_REGRESSION_POINTS`; with no refit ever reaching this estimator, the map was capped by **nothing** — one entry per distinct peer ever seen in CONNECT forwarding, for the node's whole life. That is an unbounded remote-peer-keyed collection, which `.claude/rules/code-style.md` bans. It also made `connect_forward_estimator.read().clone()` (`connect/op_ctx_task.rs:754/1026/1177`) monotonically more expensive with uptime.

Now pinned empirically: feed 2000 distinct peers, assert the map stays well under that. Against the unfixed path it retains **1993 of 2000**. Asserted as a ratio rather than against `MAX_REGRESSION_POINTS` (private to the estimator's module, and a ratio does not rot if the window is resized).

## Slow-event-rate estimators: no gap, and the premise is inverted

Asked whether an estimator the 5-min tick "used to refit regardless" is now stranded. **It never did.** The tick called `refit_stale_estimators` → `refit_if_stale`, which is:

```rust
pub fn refit_if_stale(&mut self) -> bool {
    if !self.refit_due() { return false; }   // <- same turnover predicate
    self.refit()
}
```

The tick was a **poll of `refit_due()`**, not a timer-driven refit. An estimator that had not earned >10% turnover was skipped by the tick exactly as it is skipped now. So the trigger condition is unchanged; only the latency between earning a refit and performing it changed, from "up to 5 minutes" to zero. Nothing that used to be refit no longer is — the move strictly *reduces* delay.

This also explains why dropping the timer costs no accuracy: drift is produced by `add_event`'s incremental fitting, so an estimator receiving no events is not drifting, and a timer refit would rebuild an identical fit from identical data. Recorded on `REFIT_STALENESS_PERCENT` rather than as a test, since it is an argument about the trigger, not a behaviour to pin.

## Also addressed

- **Rebased** onto `main` (`ac9784b5` #4816 + `0f856aba`). Clean fast-forward, no conflicts, zero file overlap (mine: ring/router/isotonic_estimator/telemetry/connect; #4816: routing_predictor only) — so the review is not restaled. Full suite on the rebased head: **4000 passed, 0 failed**; `clippy --locked -- -D warnings` exit 0.
- Both non-blocking doc items were already fixed on-branch by review commits: the `refit()` error arm no longer says "next poll retries" (and now documents that a persistent failure would retry per-event on the hot path), and `router.rs`'s test doc says "since deleted in #4811". I have **not** changed the error arm's behaviour: it is unreachable (`NegativePointWithIntersectOrigin` needs `intersect_origin: true`; both constructors pass `false`), and per the high-bar-for-speculative-changes rule a guard comment beats altering an unreachable path.

[AI-assisted - Claude]

sanity:
## Post-rebase status: Rule Lint GREEN, Simulation still RED (unchanged)

Rebased onto `main` (`0f856aba`, which includes #4816 and #4823). Clean fast-forward — `git rebase` reported "Successfully rebased" with no conflicts and no mid-sequence edits, and zero file overlap (mine: ring/router/isotonic_estimator/telemetry/connect; #4816: routing_predictor only) — so the review is not restaled.

**On the rebased head `e936dc59`:**

| | |
|---|---|
| `cargo test -p freenet --lib` | **4000 passed, 0 failed** |
| `cargo clippy --locked -- -D warnings` | exit 0 |
| **Rule Lint** | **PASS** — confirms the stale-label-race diagnosis: the label was absent from the old run's event payload, and CI does not trigger on `labeled`. The push re-evaluated it with the label present. |
| Simulation | **still RED**, same test, same two nodes |

**The rebase did not change the Simulation result, and it is still not main's fault.** I re-baselined against the *new* main rather than assuming:

| Build | `test_get_retry_with_alternatives_sparse_topology` |
|---|---|
| new `main` @ `0f856aba` (with #4816 + #4823) | **PASS** |
| this PR rebased @ `e936dc59` | FAIL — `2 of 10 nodes failed... Failed nodes: ["node-1", "node-2"]` |

Identical failure to pre-rebase. So the [earlier bisect](https://github.com/freenet/freenet-core/pull/4821#issuecomment-4986191456) stands unchanged: on otherwise-pristine main, deleting only the 2-line `task_monitor.register(...)` reproduces it; refit cadence vs. outcome is non-monotonic (never → FAIL, every-5-min → PASS, every-51-events → FAIL); and the non-seed-coupled twin `test_get_routing_coverage_low_htl` passes here.

That decision is still open and is the only thing blocking. I have not re-seeded unilaterally — re-seeding into a topology where the PUT caches at more nodes would make every GET trivially succeed and silently gut the retry path (#3506 verified the 2-of-10 premise with a PUT-only run precisely to avoid that).

[AI-assisted - Claude]

sanity:
## Decision: decouple the test, don't re-seed. This PR waits.

@sanity has ruled on the escalation above: **option 3 — fix the root cause.** No fourth re-seed.

The bisect is accepted in full, and the conclusion is that this PR is not at fault in any way. The decisive facts:

- Deleting **only** the 2-line `task_monitor.register(...)` on otherwise-pristine main reproduces the failure. The refit — this PR's entire subject — is not involved.
- That task ticks every 5 minutes; these sims finish in well under 5 minutes of wall-clock, so **it never fires**. It cannot affect routing semantically. Its mere existence perturbs the tokio scheduler, and the scheduler moves the seeded RNG stream.
- Non-monotonic outcome (5-min refit **passes**, *never* **fails**, every-~51-events **fails**) rules out routing quality: a quality dial cannot have "never" and "more often" both fail while "sometimes" passes.
- The twin named in the test's own doc, `test_get_routing_coverage_low_htl`, passes on this branch.

## Why not re-seed, given the test's doc sanctions it

The doc calls a re-seed "expected maintenance", and that was defensible the first time. But this brittleness has now cost three interventions — `#[ignore]` in #3488 (stdlib bump shifted RNG-dependent topology), a re-seed in #3506, and now blocking an unrelated refactor whose only sin was deleting a task that never executes. A fourth re-seed reloads the same trap for whoever next adds or removes **any** background task, and there is no signal at that point telling them it's a lottery rather than their bug.

It also lands wrong on the merits: the test obtains its precondition (PUT at HTL=2 caching at exactly 2 of 10) **by luck**, so it is a topology lottery wearing the costume of a retry-path test. The retry logic it guards is real and worth guarding — which is the argument for fixing the coupling, not for re-rolling the dice.

## Plan

1. A separate PR decouples the test: establish the sparse precondition **by construction** rather than hoping a seeded PUT produces it, so the test is indifferent to task count and scheduler order. Its acceptance criterion is that it passes on pristine main **AND** on main + the task-registration deletion — i.e. proof of decoupling rather than re-tuning. It must keep failing loudly if the precondition isn't met; a version that passes because every node trivially holds the contract would be worse than the brittle one.
2. This PR rebases once that lands and should go green with no changes to its own code.

Everything else here is already resolved: Codex clean, four review lenses reporting nothing blocking, Rule Lint green (the earlier red was a stale-label race), and all findings applied.

/cc the #4829 work — see the reply below re: timing.

[AI-assisted - Claude]

sanity:
## Re: the #4829 overlap — keep holding, but the wait just got longer

Thanks for flagging it rather than opening a PR into code being rewritten — that's the right call and I'd make the same one.

**Timing changed, so factor this in:** this PR is no longer merging shortly. Its Simulation red turned out to be a scheduler-perturbation lottery in a seed-coupled test, and @sanity has chosen to fix that test's root cause in a separate PR first (see the comment above). So this now waits on that, then a rebase. If #4829's fix is otherwise ready, **don't block on me indefinitely** — if the wait becomes material, say so and we can reorder, or you can stack on `refactor/4811-refit-in-add-event` and I'll keep the branch stable.

**On the substantive interaction — I agree with your own read, and it's worth being precise about why.** Your defect survives this PR untouched. The pinning is structural, not cadence-driven: `Adjustment::add`'s recurrence `effective_count = 1.0 + (1.0 - alpha) * effective_count` at `EWMA_ALPHA = 0.1` has a fixed point of exactly 10.0 and approaches it **from below**, while `fit` seeds `effective_count = ADJUSTMENT_PRIOR_SIZE = 10`. So the value sits at ~10 whether the reseed happens every 5 minutes, every ~51 events, or never. Both terms are independent of *when* refit runs, exactly as you said.

Which means "stickier" is the right instinct but the effect is nil in practice — you cannot get stickier than pinned at the fixed point you were already seeded to. I'd drop that clause from #4829's write-up rather than carry it: it's the kind of plausible mechanism claim that hardens into fact and then misleads the next reader. (I spent much of today removing exactly that class of claim from two other PRs in this area — including one that justified a refactor with a cross-node leak the code could not exhibit.)

What *does* go stale in your write-up is the attribution to `refit_router_periodically` reseeding every 5 minutes — that tick is deleted here. Re-verify against `IsotonicEstimator::add_event` when you rebase, as you planned.

One thing worth knowing while you're in `peer_adjustments`: this PR materially changes its lifetime behaviour. `ConnectForwardEstimator` lives on `OpManager` (`node/op_state_manager.rs:152`), so it is node-lifetime, and nothing ever refit it — meaning `peer_adjustments` only ever grew, one entry per distinct peer ever seen in CONNECT forwarding, for the life of the node. This PR bounds it to <= window + ~51. If #4829's fix reasons about that map's contents or size, that's a real behavioural change to account for, not just a doc one.

[AI-assisted - Claude]

sanity:
## For whoever picks up the decoupling: the harness already has what you need

I stood my own agent down on this to avoid two people writing the same PR. Passing on what it found before it stopped, since it's the feasibility question and it's already answered.

**`SimOperation::SeedContract` exists** (`node/testing_impl.rs:174`), is handled **pre-simulation** as a distinct phase (`:4415` — "Separate SeedContract operations (pre-simulation storage seeding)"), is explicitly *not* a client request (`:379`), and is already used by other tests (`streaming_e2e.rs:944`, and `simulation_integration.rs:2250` notes "The `SeedContract` framework is in place"). Supporting pieces: `MockStateStorage::seed_state`/`seed_params` (`mock_state_storage.rs:166`/`:172`, doc'd "Pre-populate with state (useful for setting up test scenarios)"), and `node_storages` is genuinely **per-node** — `MockStateStorage::new()` is constructed once per node at `testing_impl.rs:4624`; the local `shared_storage` name means shared between that node's own components, not across nodes.

So the sparse precondition can be established **by construction** — seed the contract into exactly the nodes you choose — instead of hoping a seeded HTL=2 PUT happens to cache at 2 of 10. No new harness machinery required.

**Two things I'd hold the PR to, both learned the hard way here:**

1. **Acceptance criterion is decoupling, not greenness.** It must pass on pristine main **AND** on main + the 2-line `task_monitor.register(...)` deletion (take it from `refactor/4811-refit-in-add-event`). Passing only the first means re-tuned, and it breaks again the moment #4821 lands — the outcome @sanity rejected. Worth establishing early against a rough draft rather than at the end.

2. **The vacuity trap is the real hazard, not a red test.** If the seeded placement leaves the contract trivially reachable, all 10 GETs pass without the retry-with-alternatives path ever executing — coverage silently deleted while looking fixed. That's *worse* than the brittle test. #3506 verified its "exactly 2 of 10" premise with a PUT-only run for exactly this reason. Make the precondition a loud assertion, and ideally show the test **fails** when the retry path is stubbed/mutated — that's the only real proof the coverage survived.

Note `use_mock_wasm=true` appears to pair with `SeedContract` in the existing usages (`simulation_integration.rs:2250`) — worth checking whether that interacts with what this test needs.

Happy to review the PR when it's up.

[AI-assisted - Claude]

sanity:
Both corrections taken, and thanks — the second one especially.

**Withdrawn from #4829's write-up** (issue body now edited, with the correction called out rather than silently rewritten):
- the attribution of the pinning to `refit_router_periodically`'s 5-minute reseed — stale the moment this lands;
- the "stickier" clause. You're right that it was a plausible-mechanism guess dressed as a finding, and right about the effect being nil: you cannot get stickier than pinned at the fixed point you were already seeded to. It's now stated as what it is — structural, from the 10.0 fixed point approached from below plus `fit`'s `ADJUSTMENT_PRIOR_SIZE = 10` seed, both independent of refit cadence.

I also carried your `peer_adjustments` note into the issue as a constraint on whoever writes the fix: node-lifetime `ConnectForwardEstimator`, never refit, unbounded growth per distinct CONNECT-forwarded peer, bounded by this PR to <= window + ~51. That's the kind of thing that silently invalidates a fix written against today's behaviour.

**On timing — no need to reorder on my account.** #4829's fix isn't written yet, and I have two PRs already in flight (#4837, #4832), so holding costs me nothing right now and I'd rather not stack a third. When I come back to it I'll re-verify item 2 against `IsotonicEstimator::add_event` on whatever main looks like then, and stack on this branch only if it's still open and the wait has become material. I'll ping here rather than surprise you with a PR into your files.

Worth noting for the record: only one of #4829's four items (the double-counted "Global events" in `peer_detail.rs:71`) is actually clear of your files. The other three touch `router.rs` / `isotonic_estimator.rs`, which is why I'm treating this as one held unit rather than salami-slicing the easy one out.

[AI-assisted - Claude]

Issue #4829, opened 2026-07-15
Peer-detail/router telemetry: double-counted events, prior-size shown as event count, fabricated 0.000s predictions


sanity:
## Problem

The per-peer detail page (`server/home_page/peer_detail.rs`) and the router prediction plumbing behind it display several values that are miscounted, structurally pinned, or outright fabricated. Unlike a stat stuck at 0, these show **plausible numbers**, so they mislead rather than obviously fail.

Possibly related: #4239 (vega gateway's `success_events` stuck at 0) may be a symptom of the same plumbing.

## 1. "Global events" double-counts every timed success

`peer_detail.rs:71` — `total_events = rs.failure_events + rs.success_events`.

But `failure_events` (`router.rs:1312`) is `failure_estimator.len()`, which already receives an event for **every** outcome including timed `Success` (`router.rs:888`). `success_events` (`router.rs:1313`) is the strict subset fed only at `router.rs:883`. Router's own doc (`router.rs:1585-1587`) says `failure_estimator` "reflects total routing events".

Displayed value = `total + timed_successes`.

## 2. "This peer: ... N events" is pinned at exactly 10

`Adjustment::event_count()` returns `effective_count.round()` (`router/isotonic_estimator.rs:599-601`). `Adjustment::add` computes `effective_count = 1.0 + (1.0 - alpha) * effective_count` with `EWMA_ALPHA = 0.1` (`:35`) — a fixed point of **exactly 10.0**, approached from below, so it can never exceed 10. `IsotonicEstimator::fit` independently seeds `effective_count = ADJUSTMENT_PRIOR_SIZE` (= 10, `:165`) at `:260`.

**The pinning is structural, not cadence-driven.** Both terms — the fixed point and the prior seed — are independent of *when* refit runs, so the displayed value sits at ~10 regardless of refit cadence. Every peer with any adjustment displays exactly "10 events" regardless of true volume. The value is a shrinkage-prior effective sample size, not an event count, and should not be labeled "events".

[...]

PR #4832, opened 2026-07-15
fix(operations): record PUT/GET/UPDATE op_stats counters (#4828)

### LINKED ISSUE #4828: op_stats counters rotted: PUT failures never recorded, GET exhaustion unrecorded, small UPDATE broadcasts uncounted
## Problem

The dashboard's Operations card shows per-op-type `ok / fail` counters. Three of them are wrong in ways that make a broken node look idle or healthy. This is the **same bug class as #4009 / #4010**, already documented in `.claude/rules/bug-prevention-patterns.md` ("Manually-mirrored telemetry counters / dashboard providers"): a `record_*` mirror silently rots when the op path changes and the call site isn't re-wired.

**#4010 was only fixed for the two op types it named** (UPDATE and SUBSCRIBE, which now record every terminal outcome through `deliver_outcome` and carry source-scrape pin tests). PUT and GET regrew the same bug.

## 1. PUT failure counter is structurally stuck at 0

`operations/put/op_ctx_task.rs:285` and `:511` are the **only** two `record_op_result` calls in the PUT driver, and both pass a hardcoded `true`:

```rust
crate::node::network_status::record_op_result(
    crate::node::network_status::OpType::Put,
    true,        // <-- hardcoded
```

Both failure arms record nothing:
- `RetryLoopOutcome::Done((Err(cause), _))` at `:558` (the #4111 bypass path — real and reachable)
- `RetryLoopOutcome::Exhausted(cause)` at `:565` (retry budget burned)

So `op_stats.puts.1`, the red "fail" number at `server/home_page/cards.rs:1537`, can never be non-zero.

GET has a pin test against precisely this — `get/op_ctx_task.rs:4939` `record_op_result_reflects_host_result_outcome`, whose assertion text literally reads *"Unconditional `true` is the bug."* PUT has no equivalent pin.

## 2. GET exhaustion is recorded as neither success nor failure

`operations/get/op_ctx_task.rs:736` — the `Exhausted` arm publishes `ContractResponse::NotFound` and emits a terminal event, but never calls `record_op_result`. Only the `Done` arm (`:645`) records.

GET dead-ends are the dominant production `not_found` mode (see the placement/migration work), and they increment **neither** `gets.0` nor `gets.1`. This is an undercount rather than a dead counter: the `Done` arm's `host_result.is_ok()` gating does populate `gets.1` for terminal NotFound replies.

## 3. `updates_received` only counts streaming (>=64 KB) broadcasts

`operations/update/op_ctx_task.rs:1938` is the **sole** `record_update_received()` call site in the crate, and it sits inside `drive_relay_broadcast_to_streaming` (starts `:1732`).

The non-streaming twin `drive_relay_broadcast_to` (`:1057`-`:1412`) runs the identical apply-then-`if !changed` sequence (`:1268`, mirroring `:1933`) but never records. `broadcast_queue.rs:654` selects the non-streaming `BroadcastTo` for payloads below `streaming_threshold` (default 64 KB, `config.rs:238`), so **ordinary small deltas — the normal River case — are invisible**.

This matters because `cards.rs:1519-1525` renders `updates_received` as the *only* UPDATE number for subscriber nodes (its own comment: "Only received (common for subscriber nodes)"). An actively-receiving subscriber displays `0`.

Note `drive_relay_broadcast_to` *does* call `record_relayed_update()` (`:1379`), but that feeds `relayed_op_stats` (`network_status.rs:1029`), a different counter not rendered on the ops card.

## Suggested fix

1. PUT: pass the real outcome; record on both failure arms. Add the GET-style pin test.
2. GET: record a failure on the `Exhausted` arm.
3. UPDATE: record in `drive_relay_broadcast_to` alongside the streaming twin.

Per `bug-prevention-patterns.md`, each fix should carry a **source-scrape pin test** so the next migration that drops the call fails CI — that's the gap that let #4010 regrow here.

## Also: stale rustdoc on the audit-reference doc

`node/network_status.rs:753` lists "Current writers" and names `node.rs::report_result` for "legacy state-machine ops that still flow through `OpManager.ops.*`". `report_result` still exists (`node.rs:770`) but contains no `record_op_result` call, and the legacy mediator path is gone (per `crates/core/CLAUDE.md`). This is the exact doc `bug-prevention-patterns.md` points auditors at, so its staleness actively misleads the next audit.

[AI-assisted - Claude]

sanity:
## Problem

The dashboard Operations card showed three `op_stats` counters that made a broken node look idle or healthy. All three are the **#4009/#4010 bug class** documented in `.claude/rules/bug-prevention-patterns.md` ("manually-mirrored telemetry counters / dashboard providers") **regrown**: #4010 was fixed only for UPDATE and SUBSCRIBE — which now record every terminal outcome via `deliver_outcome` and carry source-scrape pin tests — while PUT and GET drifted back to the same defect.

**1. PUT failure counter structurally stuck at 0.** The only two `record_op_result` calls in the PUT driver sat in its *success* arms, both passing a hardcoded `true`. Every failure arm recorded nothing: the `Done((Err(cause), _))` #4111 bypass, `Exhausted(cause)`, and the `InfrastructureError` synthesis. So `op_stats.puts.1` — the red "fail" number at `server/home_page/cards.rs` — was **unreachable**: a node failing every PUT displayed zero failures.

**2. GET exhaustion recorded as neither success nor failure.** The `Exhausted` arm publishes `ContractResponse::NotFound` and emits a terminal event, but never called `record_op_result`; only the `Done` arm did. GET dead-ends are the dominant production `not_found` mode, so the ops card showed an idle-looking node while every GET failed.

**3. `updates_received` only counted streaming (>=64KB) broadcasts.** The sole `record_update_received()` call site in the crate was inside `drive_relay_broadcast_to_streaming`. The non-streaming twin `drive_relay_broadcast_to` runs the identical apply-then-`if !changed` sequence but never recorded. `broadcast_queue.rs` picks the non-streaming variant below `streaming_threshold` (default 64KB), so ordinary small deltas — the normal River case — were invisible, and `cards.rs` renders `updates_received` as the **only** UPDATE number for subscriber nodes.

## Solution

**PUT** — move recording out of the driver arms into `deliver_outcome`, with a `classify_put_outcome_for_op_stats` classifier. `run_client_put` -> `drive_client_put` -> `drive_client_put_inner` is the only call chain and `deliver_outcome` is called unconditionally on its result, so this single funnel sees *every* terminal outcome. This is UPDATE's #4010 fix shape — the one that did **not** rot. Gated on `!client_tx.is_sub_operation()` for parity with UPDATE/SUBSCRIBE (defensive; PUT has no sub-op entry points today).

**GET** — record an explicit `false` in the `Exhausted` arm. GET deliberately keeps arm-based recording rather than adopting the funnel: exhaustion publishes `Ok(HostResponse::…NotFound)` so that clients can distinguish "contract absent" from "operation failed", which means a `matches!(Publish(Ok(_)))` classifier would score a dead-end as a **success**.

**UPDATE** — record in `drive_relay_broadcast_to` too, after its `!changed` guard (matching the streaming twin: a broadcast that mutated nothing is not a received update). Note `drive_relay_broadcast_to` already called `record_relayed_update()`, but that feeds a different counter (`relayed_op_stats`), not the ops card.

**Pin tests close the CI gap** that let #4010 regrow here — each fix carries a source-scrape pin so a future migration that drops the call fails CI, per `bug-prevention-patterns.md`.

**Stale rustdoc** on `network_status::record_op_result` fixed: it named `node.rs::report_result` as a writer for "legacy state-machine ops that still flow through OpManager.ops.*", but that function contains no `record_op_result` call and the legacy mediator path is gone (see `crates/core/CLAUDE.md`). This is the doc `bug-prevention-patterns.md` points auditors at, so its staleness misled the next audit. The "Current writers" list now matches the five real production call sites exactly.

## Testing

Regression test written first, verified failing, then fixed, then verified passing — **fail-then-pass observed for all three defects**. The unfixed state was reproduced faithfully by splicing origin/main's production code (everything before `#[cfg(test)]`) with the new test module, which landed the bug on the exact reported lines.

Pin tests added:

| Test | Guards |
|---|---|
| `put::…::deliver_outcome_records_put_op_result` | `deliver_outcome` calls `record_op_result` with `OpType::Put`, a **derived** (not hardcoded `true`) success flag, and the sub-op gate |
| `put::…::put_driver_must_not_record_op_result_directly` | driver must not re-inline the call (double-count / partial-arm regression) |
| `put::…::classify_put_outcome_covers_all_variants` | pure-function coverage of every `DriverOutcome` variant |
| `get::…::exhausted_arm_records_get_failure` | Exhausted arm records `OpType::Get` with `false` |
| `update::…::both_broadcast_drivers_record_update_received` | **both** broadcast drivers record, each after its `!changed` guard |

Against the unfixed code all four source-scrape pins fail with their intended diagnostics; `classify_put_outcome_covers_all_variants` fails to compile (it covers the new classifier). With the fix, all 5 pass.

- `cargo test -p freenet --lib -- operations::put::op_ctx_task operations::get::op_ctx_task operations::update::op_ctx_task node::network_status` — **229 passed, 0 failed**
- `cargo fmt --check` clean
- `cargo clippy -p freenet --lib -- -D warnings` clean

Scope is limited to these three counters plus the rustdoc; SUBSCRIBE/CONNECT, the cards, and #4829/#4830 are untouched.

Fixes #4828

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -750,14 +750,36 @@ pub fn record_peer_disconnected(addr: SocketAddr) {
 /// the matching outcome. Forgetting a call site silently rots the
 /// counter (issue #4009 / #4010 prior incidents). Current writers:
 ///
-/// - `node.rs::report_result` for legacy state-machine ops (PUT/GET/
-///   UPDATE/SUBSCRIBE that still flow through `OpManager.ops.*`).
-/// - `operations/get/op_ctx_task.rs` `Done` arm (client-initiated GET).
-/// - `operations/put/op_ctx_task.rs` `Done` arm (client-initiated PUT).
+/// - `operations/get/op_ctx_task.rs` `Done` arm (client-initiated GET;
+///   success flag from `host_result.is_ok()`), its `Exhausted` arm
+///   (records a failure — exhaustion publishes `NotFound`, having
+///   delivered no state), AND `deliver_outcome`'s `InfrastructureError`
+///   arm (records a failure — the `Unexpected`/`InfraError` terminal
+///   outcomes publish a synthesized client `Err`). GET cannot fold the
+///   success/exhaustion arms into an outcome-shape classifier the way
+///   PUT/UPDATE do: a dead-end publishes `Ok(HostResponse::…NotFound)`,
+///   so `matches!(Publish(Ok(_)))` would score it as a success (#4828).
+/// - `operations/put/op_ctx_task.rs::deliver_outcome` (client-initiated
+///   PUT; covers all `DriverOutcome` variants).
 /// - `operations/subscribe/op_ctx_task.rs::deliver_outcome` (client-
 ///   initiated SUBSCRIBE; covers all `DriverOutcome` variants).
 /// - `operations/update/op_ctx_task.rs::deliver_outcome` (client-
 ///   initiated UPDATE; covers all `DriverOutcome` variants).
+///
+/// `node.rs::report_result` is NOT a writer: the legacy `OpEnum` /
+/// mediator path is gone (see `crates/core/CLAUDE.md`), every op runs a
+/// task-per-tx driver, and its `Ok` branch has nothing left to report.
+/// It was listed here until #4828 — a stale entry on the very doc this
+/// section points auditors at.
+///
+/// Prefer the `deliver_outcome` funnel shape where the driver's outcome
+/// type can express failure: it is the one that did NOT rot. Recording
+/// in individual driver arms is how #4828 regrew #4009/#4010 on PUT —
+/// the success arms carried a hardcoded `true` and every failure arm was
+/// simply forgotten, making `op_stats.puts.1` unreachable. Each op's
+/// call site is pinned by a source-scrape test in its `op_ctx_task.rs`
+/// test module; those pins are the CI gate against the next migration
+/// silently dropping the call.
 ///
 /// Internal-only operations are intentionally NOT recorded; counting
 /// them would inflate the user-facing counter with background traffic
@@ -798,7 +820,30 @@ pub fn record_op_result(op_type: OpType, success: bool) {
     }
 }
 
-/// Record a broadcast update received via subscription streaming.
+/// Record a broadcast update that was received and actually changed local
+/// state.
+///
+/// # Required call sites
+///
+/// Manually-mirrored counter. BOTH broadcast-apply drivers in
+/// `operations/update/op_ctx_task.rs` must call this, each after its
+/// `if !changed` early-return (a no-op re-broadcast is not a received
+/// update):
+///
+/// - `drive_relay_broadcast_to` (`BroadcastTo` — every payload BELOW
+///   `streaming_threshold`, i.e. the ordinary small-delta case).
+/// - `drive_relay_broadcast_to_streaming` (`BroadcastToStreaming` —
+///   payloads at/above the threshold, default 64 KB).
+///
+/// Until #4828 only the streaming twin recorded, so ordinary small deltas
+/// — the normal River case — were invisible and an actively-receiving
+/// subscriber showed 0 (`server/home_page/cards.rs` renders
+/// `updates_received` as the ONLY UPDATE number for subscriber nodes).
+/// Pinned by `both_broadcast_drivers_record_update_received`.
+///
+/// Distinct from [`record_relayed_update`], which counts relayed UPDATE
+/// *requests* toward the key (`relayed_op_stats`) and deliberately does
+/// NOT count broadcast fan-out.
 pub fn record_update_received() {
     if let Some(status) = NETWORK_STATUS.get() {
         if let Ok(mut s) = status.write() {
--- a/crates/core/src/operations/get/op_ctx_task.rs
+++ b/crates/core/src/operations/get/op_ctx_task.rs
@@ -768,6 +768,24 @@ async fn drive_client_get_inner(
                 None,
             )
             .await;
+            // Dashboard op_stats GET counter (issue #4828). Exhaustion
+            // delivered no state to the client, so it is a FAILED GET —
+            // record it explicitly as such. Before #4828 this arm recorded
+            // nothing, so a GET dead-end (the dominant production
+            // `not_found` mode) incremented NEITHER `gets.0` nor `gets.1`
+            // and the ops card showed an idle-looking node while every GET
+            // failed.
+            //
+            // This must stay an explicit `false` rather than moving to a
+            // `deliver_outcome` funnel the way PUT/UPDATE do: the arm
+            // publishes `Ok(HostResponse::…NotFound)` so that clients can
+            // distinguish "contract genuinely absent" from "operation
+            // failed", which means a `matches!(Publish(Ok(_)))` classifier
+            // would score a dead-end as a SUCCESS.
+            crate::node::network_status::record_op_result(
+                crate::node::network_status::OpType::Get,
+                false,
+            );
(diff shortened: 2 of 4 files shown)


sanity:
## Parking this PR — NAT Validation is red and the cause is unresolved

Not merging this. Documenting the state so the next person doesn't re-derive it.

### The failure

CI job "NAT Validation" (a job in `ci.yml`) fails on this branch:

```
test_nat_subscription_cross_peer_update_via_docker_nat  -- TRY 3 FAIL
Subscriber peer (peer0, behind NAT) should have received the 'cross-peer-nat-node'
key set by a DIFFERENT NAT peer (peer1) via gateway-relayed downstream subscription.
Delivery requires the gateway to have registered peer0 by its NAT-observed source_addr.
Got keys: []
Summary [217.220s] 2 tests run: 1 passed, 1 failed, 15 skipped
```

Failed **all 3** nextest retries, so a one-off flake is unlikely on its face.

### What the evidence says (and doesn't)

**Points at this branch:** NAT Validation is green on 8 recent runs across 5 branches (`refactor/4811-refit-in-add-event` x2, `fix/4813-flaky-delegate-count`, `fix/wasm-module-cache-eviction`, `fix/4810-predictor-training-freeze` x2, and two merge-queue runs). It is red only here.

**Weakens that:** none of those green branches touch `operations/put` or `operations/update`. This PR is the only recent one touching the paths the test exercises, so the green runs may simply not exercise a regression. NAT Validation is also **skipped on main pushes** (path-filtered), so there is no main baseline to diff against.

**Argues against causation:** all three changes are telemetry-only and none alter control flow. `record_update_received()` is a single `saturating_add` behind a non-panicking `if let Ok(status.write())`, placed after the existing `!changed` guard, and its streaming twin `drive_relay_broadcast_to_streaming` has always made the identical call on the same path without issue.

### Unresolved hypotheses, for whoever picks this up

1. **Non-reentrant lock deadlock (my leading suspicion, NOT verified).** `record_update_received` / `record_op_result` take `NETWORK_STATUS.write()`. `std::sync::RwLock` is not reentrant, so if anything up the broadcast call chain already holds a read/write guard on the same lock on the same thread, the new call deadlocks — a stalled broadcast yields exactly "Got keys: []". The check is to compare the lock context at the new call site against the streaming twin's; if they differ, that difference is the bug.
2. **Scheduler-perturbation lottery.** #4821 just demonstrated this failure mode in this repo: deleting a 2-line `task_monitor.register(...)` that never fires reproduced a Simulation failure purely by perturbing the tokio scheduler and moving the seeded RNG stream. Added lock acquisitions could perturb timing similarly. Caveat: this test is Docker e2e, not a seeded sim, so the precedent may not transfer.
3. **The PUT `deliver_outcome` refactor changed real behaviour.** The test has peer1 *set* the key peer0 must receive, so the PUT path is directly involved. Worth confirming `deliver_outcome` really is called exactly once per terminal outcome.

Whichever it is, it should be run down properly rather than papered over — if it's (1) that's a genuine bug this PR would ship; if it's (2) it's a brittle test that will bite the next unrelated PR, which is precisely the trap #4821 is currently untangling.

### What's still good here

The three counter fixes and their source-scrape pin tests are, as far as reviewed, sound — the Claude Rule Review bot independently traced them end-to-end and confirmed PUT recording isn't lost or duplicated, the rustdoc matches reality, and the non-streaming call site sits after the `!changed` guard like its twin. Clippy/Fmt/Conventional-Commits/Rule-Lint all pass.

Two known nits if resumed: the commit subject is 87 chars (over git-workflow.md's 72 limit — this is the `rule-review/warnings` red), and the NAT question above.

Issue #4828 retains the full analysis. Parking rather than merging, per maintainer call — the counter fixes aren't worth shipping on an unexplained red NAT.

[AI-assisted - Claude]

PR #4837, opened 2026-07-15
fix(transport): record failed transfers (transfers_failed had no writer)

### LINKED ISSUE #4827: Dashboard: transfers_failed counter has no writer (permanently 0)
## Problem

`TransportMetrics::transfers_failed` is declared, initialized, read, reset, rendered on the local peer dashboard, and uploaded to telemetry — but **nothing ever increments it**. There is no `record_transfer_failed` function anywhere in the crate. The counter is structurally pinned at 0.

User-visible effect: the dashboard's Data Transfer row always reads `N ok / 0 fail` (`server/home_page/cards.rs:384-391`), and `tracing/telemetry.rs:2009` uploads `transfers_failed: 0` forever. A node whose transfers are all failing looks perfectly healthy.

Originally spotted by Codex (reported by Ivvor); independently confirmed by tracing every access.

## Evidence

Declared and initialized:
- `transport/metrics.rs:57` — `transfers_failed: AtomicU32`
- `transport/metrics.rs:161` — `AtomicU32::new(0)`

Every access is a **read**:
- `:473` (`read_snapshot` load), `:558` (`take_snapshot` swap), `:596` (activity gate)

No `fetch_add` / `fetch_update` / non-zero `store` exists repo-wide. The only non-zero write is `server/home_page.rs:749`, a `#[cfg(test)]` fixture assigning to the *snapshot* struct, not the metric.

This is clearly an unfinished wiring rather than an unimplemented idea, because the failure paths are **already instrumented for the event stream**: `emit_transfer_failed` has six production call sites (`transport/peer_connection/outbound_stream.rs:210, 347, 550, 579, 651, 743`). And the counter's success sibling sits directly beside the failure arms at `transport/peer_connection.rs:1183-1192`:

```rust
Ok(stats) => { ... super::TRANSPORT_METRICS.record_transfer_completed(&stats); }   // :1184
Err(e) if e.is_transient_send_failure() => { tracing::warn!(...); }                 // no counter
Err(e) => return Err(e),                                                            // no counter
```

Second-order effect: `transfers_failed` is an input to `take_snapshot`'s no-activity gate (`:596`). Being structurally 0 it contributes nothing, so a period whose only activity was *failed* transfers may emit no snapshot at all.

## Suggested fix

Add `record_transfer_failed()` next to `record_transfer_completed` (`:187`) and call it from the failure arms at `peer_connection.rs:1186+`. Add a regression test asserting the counter moves (the existing tests only assert it stays 0, e.g. `:906`, `:933` — which is exactly why CI never caught this).

## Related (same file, filing here for context, separable)

1. **`bytes_sent` / `bytes_received` are asymmetric** despite looking like a pair. `bytes_sent` (`:200`) counts only outbound *stream* payloads; `bytes_received` (`:358`) is called for both short messages (`peer_connection.rs:1122`) and inbound streams (`:1153`). Outbound short messages are counted in neither. Any consumer comparing the two sees a systematic sent-side undercount. Not dashboard-visible (the card uses the symmetric `cumulative_*` counters).

2. **Dashboard rows flicker on the telemetry worker's 30s reset cycle.** The dashboard reads the non-resetting `read_snapshot()` (`node/network_status.rs:1784`), but the fields it gates rows on are the same ones `take_snapshot()` (`tracing/telemetry.rs:713`) swaps to 0 every `transport_snapshot_interval_secs` (default 30). On a node with no outbound stream completion in a window — common on a quiet node — the Transfers, cwnd, and LEDBAT-slowdowns rows plus the `(peak X/s)` suffix **disappear entirely** until the next stream. Only RTT survives, because keepalive ping/pong feeds `record_rtt_sample` every ~5s. Turning telemetry off silently changes those same rows from a 30s window to lifetime aggregates: same panel, two meanings. Needs a design call (dashboard reads lifetime counters, vs. accept window semantics and stop gating rows on `> 0`).

[AI-assisted - Claude]

sanity:
## Problem

`TransportMetrics::transfers_failed` is declared, initialized, read, reset, rendered on the local peer dashboard, and uploaded to telemetry — but **nothing ever incremented it**. No `record_transfer_failed` existed anywhere in the crate, so the counter was structurally pinned at 0.

User-visible effect: the dashboard's Data Transfer row always read `N ok / 0 fail` (`server/home_page/cards.rs:384-391`), and `tracing/telemetry.rs:2009` uploaded `transfers_failed: 0` forever. **A node whose transfers were all failing looked perfectly healthy.**

Second-order effect: `transfers_failed` is an input to `take_snapshot`'s no-activity gate (`metrics.rs:596`). Being structurally 0 it contributed nothing, so a period whose only activity was *failed* transfers could emit no snapshot at all.

This was unfinished wiring rather than an unimplemented idea: the failure paths were already instrumented for the **event stream** (`emit_transfer_failed`, six call sites in `outbound_stream.rs`), and the counter's success sibling `record_transfer_completed` sits directly beside the failure arms.

## Solution

Add `record_transfer_failed()` next to `record_transfer_completed`, and call it from **both** outbound-stream failure arms in `PeerConnection::recv`.

### Judgment call: why both arms, not just the terminal one

The brief flagged a real risk — that counting the `is_transient_send_failure()` arm might inflate the metric or double-count a retried transfer. I traced the code, and it doesn't:

**"Transient" classifies CONNECTION liveness, not the transfer's fate.** Its own rustdoc (`transport.rs:368-387`) is explicit: it returns true for "a stream-scoped failure that must NOT kill the connection", where "a single stalled or failed stream **only fails that stream**, and the operation layer times out and **retries against another candidate**." The stream task has already returned — the transfer is dead in both arms. The arms differ only in whether the *connection* survives (#4345).

**No double-count, and it does not depend on retry behaviour.** One spawned task yields exactly one result; `FuturesUnordered` yields that result exactly once; there is no `.await` between the yield and the increment. One stream, one terminal outcome, counted once. (An earlier revision of this PR justified this via "a retry opens a new stream with a new `stream_id`". That reasoning was wrong for the dominant case — `MAX_PEER_ADVANCEMENTS_STREAMING = 0` (`put/op_ctx_task.rs:1067`, selected at `:444`), so streaming PUTs — exactly the payloads that produce outbound-stream failures — get **zero** op-layer advancements, and `get/op_ctx_task.rs:1851` notes GETs are never streaming today. The conclusion survives; the stated reason didn't, and has been replaced by the one-task-one-result argument above.)

### Emit-site census

The six `emit_transfer_failed` sites, and which arm their error return lands in:

| emit site | returns | `is_transient_send_failure()` | arm |
|---|---|---|---|
| 210 → 248 | `OutboundStreamFailed` (cwnd-wait timeout) | true | transient |
| 347 → 390 | `e` from `packet_sending` | see below | either |
| 550 → 572 | `OutboundStreamFailed` (pipe inactivity stall) | true | transient |
| 579 → 595 | `OutboundStreamFailed` (inbound-feed error) | true | transient |
| 651 → 672 | `OutboundStreamFailed` (cwnd-wait timeout) | true | transient |
| 743 → 774 | `e` from `packet_sending` | see below | either |

So: **four** `OutboundStreamFailed` sites, and **two** mid-send `packet_sending` sites. `packet_sending` returns `SendFailed` on socket error (transient), **or** propagates a bincode `Serialization` error from `try_serialize_msg_to_packet_data` via `?` (`peer_connection.rs:2242`) — and `Serialization` is *not* transient, so it reaches the terminal arm. Each mid-send site can therefore return either kind; both emit `emit_transfer_failed` before returning, regardless of which.

Every production error return in `send_stream`/`pipe_stream` emits the event, and they distribute across *both* arms. Counting only the transient arm would silently miss the serialization case; **counting only the terminal arm would have left the counter ~always 0 and not fixed this issue at all** — every common failure (cwnd-wait timeout, pipe stall, inbound-feed error, socket send error) is transient.

This census is deliberately kept here rather than inline in the code: it is cross-file bookkeeping that a 7th emit site would silently falsify, with nothing to pin it. The inline comment keeps only the why (the liveness-vs-fate distinction and the counted-once argument), which stays true regardless of site count.

### Where the counter and the event stream cannot agree

The counter is **not** unconditionally 1:1 with the event stream. `emit_transfer_failed` fires **inside** the stream task (`send_stream` spans `outbound_stream.rs:103-473`, `pipe_stream` `:485-841`); the counter increments only where the task's *result* is observed, in `recv()`. They agree 1:1 for every stream whose result reaches that match — with two known gaps:

- **Panic (symmetric).** A stream task that panics becomes a `JoinError` → `TransportError::Other` via the `?` on the join result, *before* the match. It reaches neither arm, so it is missing from **both** the counter and the event stream. The two still agree with each other; both under-count panics equally.

- **Teardown orphan (ASYMMETRIC).** `PeerConnection::drop` (`peer_connection.rs:308-333`) cancels `streaming_handles` and aborts `keep_alive_handle`, but does **not** abort or drain `outbound_stream_futures`. Dropping a `FuturesUnordered<JoinHandle<_>>` drops the `JoinHandle`s, which **detaches** the tasks (`GlobalExecutor::spawn` returns a real `tokio::task::JoinHandle`, `config.rs:2923-2925`). A detached `send_stream`/`pipe_stream` task that later fails still calls `emit_transfer_failed` from inside the task — but no `recv()` remains to observe its result, so the **event fires and the counter does not**. This concentrates at teardown: the terminal arm's `return Err(e)` ends `recv()`, the caller drops the connection, and sibling in-flight streams orphan — i.e. exactly when mass stream failures happen. `record_transfer_completed` has the same gap.

Both gaps are now documented at the call site and in the `record_transfer_failed` rustdoc. Closing them is out of scope here.

### Implementation notes

- Saturating increment (matching `record_transfer_completed`): a pinned `u32::MAX` is a better failure mode than wrapping to 0 and reporting a healthy node.
- Records **only** the counter, unlike the success path — a failed transfer has no meaningful duration/throughput to aggregate, and bytes already on the wire are counted at the socket layer by `record_packet_sent`.

### Third file: suppressing a now-reachable dishonest render

`server/home_page/cards.rs` gates the transfers row on `transfers_completed > 0 || transfers_failed > 0`. **The `|| transfers_failed > 0` half was dead code before this PR** — the row only ever rendered with >=1 completion, so its average was always real. Giving the counter a writer makes an all-failures window reachable, and it would render:

> `0 ok / 3 fail (0.000s avg)`

presenting the sentinel 0 from `metrics.rs`'s `if transfers_completed > 0 { … } else { 0 }` guard as a measured average. A 30s all-failures window is exactly the sick-node case this PR exists to surface, so the first thing it renders must not be a fabricated timing.

Fixed by suppressing the average span when `transfers_completed == 0`, following the file's existing idiom (`peak_tput`, `rtt_str`, `cwnd_str` all conditionally suppress). **Flagged explicitly as a third file in the diff**: it's in scope under the repo's finish-the-fix rule because *this PR* is what makes the branch reachable, but it is not part of the counter wiring itself.

## Testing

**Two regression tests that drive the real `recv()` loop**, one per arm.

This is deliberate: the bug was missing **wiring**, not a missing counter. A test that only called `record_transfer_failed` directly would pass with **both call sites deleted** — it would not have caught this. The pre-existing metrics tests (`metrics.rs:906`, `:933`) only assert the counter *stays* 0, which is precisely why CI never caught it; that pattern is not followed here.

Verified the full falsification cycle explicitly:

- **With the fix**: both tests pass.
- **Fix removed** (both `record_transfer_failed()` call sites deleted, tests untouched): both fail with `left: 0, right: 1` — reproducing the exact bug (counter permanently 0).
- **Fix restored**: both pass again.

Test design details:
- Each asserts an **exact** `before + 1` delta, not a weaker `>=`, so they fail on a double-count as well as on the original never-counts bug.
- Each asserts its **premise** (`is_transient_send_failure()` is/isn't true for the injected error), so a future reclassification cannot leave a test silently exercising the wrong arm and passing for the wrong reason.
- The terminal test injects `ConnectionEstablishmentFailure` with a unique sentinel rather than `ConnectionClosed`, because a closed inbound channel also yields `ConnectionClosed` — asserting on that variant could pass for the wrong reason. The sentinel proves the error came from our stream task.
- The tests serialize on a shared lock (`TRANSPORT_METRICS` is a process-wide global) so they can't observe each other's increment.

**Plus a pin test for the newly-live activity gate**: `test_snapshot_emitted_for_failed_transfers_only`. `take_snapshot`'s no-activity gate includes `&& transfers_failed == 0`, a conjunct that was **dead until this PR** (the counter was structurally 0) and is now live for the first time. Both its siblings have a dedicated pin test for exactly this branch (`test_snapshot_emitted_for_rtt_only_activity`, citing #4000, and `test_snapshot_emitted_for_nat_only_activity`) because dropping a gate conjunct silently discards a period's telemetry; there was no failures-only equivalent. Without it, someone "simplifying" the gate by dropping the conjunct would make a failures-only window return `None` → telemetry uploads nothing → the fail counter reads a permanent 0 again, i.e. the exact #4827 symptom with green CI.

Falsified the same way: removing `&& transfers_failed == 0` from the gate makes it fail with *"failure-only activity must still produce a snapshot"*; restored, it passes. It also closes `record_transfer_failed`'s previously-zero coverage in `metrics.rs` — that a failure surfaces through `take_snapshot()` (the value telemetry actually uploads, distinct from the `read_snapshot()` the `peer_connection` tests use) and that it resets afterwards.

**Plus a render test**: `transfer_card_omits_avg_when_nothing_completed` (and an added assertion to `transfer_card_renders_subsections_with_data` that a real average still renders). Falsified by forcing the avg span unconditionally, which makes it fail.

Also green: full `transport::` lib suite (685 passed), `server::home_page::` (121 passed), `cargo fmt`, and `cargo clippy -p freenet --lib -- -D warnings` clean.

## Review follow-ups applied

A four-reviewer pass found no blocking correctness findings; every finding was comment/doc accuracy or a test gap. All are addressed:

1. **The "1:1" claim was false** — disclosed the asymmetric teardown/detach gap alongside the panic gap, and scoped the claim to "every stream whose result is observed by `recv()`". Fixed at the call site, in the `record_transfer_failed` rustdoc, and here. (Mechanism independently re-verified against `Drop`, the `FuturesUnordered<JoinHandle>` field, and `GlobalExecutor::spawn`'s return type.)
2. **The "all six" enumeration miscounted** — it reached six by adding 4 *sites* + 2 *error kinds arising from those same 2 sites*. The inline census is removed (it's cross-file bookkeeping that would rot); the corrected census lives in this PR body.
3. **The retry premise was wrong for the dominant case** — replaced with the one-task-one-result guarantee, which is airtight and retry-independent.
4. **Missing pin test for the newly-live activity gate** — added, plus `record_transfer_failed`'s `take_snapshot`/reset coverage.
5. **Stale `read_snapshot` doc** — it omitted `transfers_completed`/`transfers_failed` from the period accumulators `take_snapshot` resets, implying they are not reset. Corrected.
6. **`metrics.rs` overstated "the single place every stream task's result is observed"** — `recv()` also observes `inbound_stream_futures`; scoped to *outbound*.
7. **Test-helper doc had a wrong causal model** — it claimed the timer arms are TimeSource-driven with first ticks "seconds away". Actually `ack_interval` is selected by `SimulationTransportOpt::is_enabled()`, giving `ACK_CHECK_INTERVAL` = 100ms; the resend arm fires at 10ms then every 2ms. Only `timeout_check` (5s) matched the claim. The tests pass for a different reason — those arms are no-ops on an idle fixture (`ack_check` returns early on `receipts.is_empty()`). Corrected to the real reason.
8. **This PR made a dishonest render reachable** — fixed in `cards.rs`; see "Third file" above.

## Scope

Strictly the `transfers_failed` counter, plus the `cards.rs` render branch this PR makes reachable (flagged above). The two related items in the issue — the `bytes_sent`/`bytes_received` asymmetry, and the 30s-reset row flicker (which needs a design call) — are deliberately **not** touched here.

Fixes #4827

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/home_page.rs
+++ b/crates/core/src/server/home_page.rs
@@ -758,6 +758,41 @@ mod tests {
         assert!(html.contains("23</span>"), "slowdown count missing");
         assert!(html.contains("847"), "transfers completed missing");
         assert!(html.contains("3"), "transfers failed missing");
+        assert!(
+            html.contains("1.200s avg"),
+            "a real average must still render when transfers completed"
+        );
+    }
+
+    /// An all-failures window must NOT render an average.
+    ///
+    /// `avg_transfer_time_ms` is a sentinel 0 when nothing completed
+    /// (metrics.rs guards the division), so rendering it would claim a measured
+    /// "0.000s avg" for transfers that produced no timing at all. The window is
+    /// reachable only since #4827 gave `transfers_failed` a writer: before that
+    /// the counter was structurally 0, so this row required >=1 completion and
+    /// the average was always real. A 30s all-failures window is exactly the
+    /// sick-node case #4827 exists to surface, so the first thing it renders
+    /// must not be a fabricated timing.
+    #[test]
+    fn transfer_card_omits_avg_when_nothing_completed() {
+        let mut snap = base_snapshot();
+        snap.bytes_uploaded = 5000;
+        snap.open_connections = 1;
+        snap.transport_snapshot.transfers_completed = 0;
+        snap.transport_snapshot.transfers_failed = 3;
+        // The sentinel metrics.rs produces when transfers_completed == 0.
+        snap.transport_snapshot.avg_transfer_time_ms = 0;
+        let html = build_transfer_card(&Some(snap));
+
+        assert!(
+            html.contains("0 ok / 3 fail"),
+            "the failure row must still render — surfacing it is the point of #4827"
+        );
+        assert!(
+            !html.contains("avg)"),
+            "must not present the sentinel 0 as a measured average; got: {html}"
+        );
     }
 
     #[test]
--- a/crates/core/src/server/home_page/cards.rs
+++ b/crates/core/src/server/home_page/cards.rs
@@ -381,15 +381,31 @@ pub fn build_transfer_card(snap: &Option<network_status::NetworkStatusSnapshot>)
         String::new()
     };
 
+    // Only render the average when something actually completed.
+    // `avg_transfer_time_ms` is a sentinel 0 when `transfers_completed == 0`
+    // (metrics.rs guards the division), so an all-failures window would
+    // otherwise render "0 ok / 3 fail (0.000s avg)" — presenting the sentinel
+    // as a measured timing. That window became reachable when #4827 gave
+    // `transfers_failed` a writer: before that the counter was structurally 0,
+    // so this row only ever rendered with >=1 completion and the avg was always
+    // real.
+    let avg_str = if ts.transfers_completed > 0 {
+        format!(
+            r#" <span class="transfer-detail">({avg}s avg)</span>"#,
+            avg = format_args!("{:.3}", ts.avg_transfer_time_ms as f64 / 1000.0),
+        )
+    } else {
+        String::new()
+    };
+
     let xfer_str = if ts.transfers_completed > 0 || ts.transfers_failed > 0 {
         format!(
             r#"<div class="transfer-stat">
                 <span class="transfer-label">Transfers (avg time)</span>
-                <span class="transfer-value">{ok} ok / {fail} fail <span class="transfer-detail">({avg}s avg)</span></span>
+                <span class="transfer-value">{ok} ok / {fail} fail{avg_str}</span>
             </div>"#,
             ok = ts.transfers_completed,
             fail = ts.transfers_failed,
-            avg = format_args!("{:.3}", ts.avg_transfer_time_ms as f64 / 1000.0),
         )
     } else {
         String::new()
--- a/crates/core/src/transport/metrics.rs
+++ b/crates/core/src/transport/metrics.rs
@@ -245,6 +245,40 @@ impl TransportMetrics {
         }
     }
 
+    /// Record a failed outbound transfer.
+    ///
+    /// The failure sibling of [`Self::record_transfer_completed`]. Together the
+    /// two count every outbound stream that resolves: `send_stream` /
+    /// `pipe_stream` either run to completion (completed) or return early from
+    /// one of their error sites (failed), never both.
+    ///
+    /// Call this once per failed stream, from the OUTBOUND-stream arms of
+    /// `PeerConnection::recv` — the place every outbound stream task's result
+    /// is observed. (`recv` also observes `inbound_stream_futures`; those are
+    /// not transfers this counter tracks.)
+    ///
+    /// It mirrors the `emit_transfer_failed` event for every stream whose
+    /// result is observed there, but the two are NOT unconditionally 1:1: the
+    /// event is emitted inside the stream task, while this counter increments
+    /// only at the observation site, so a task that panics or is orphaned by
+    /// connection teardown diverges. See the call sites in `peer_connection.rs`
+    /// for both gaps and for the classification rationale.
+    ///
+    /// Unlike the success path this records ONLY the counter: a failed transfer
+    /// has no meaningful duration, throughput, or payload size to aggregate,
+    /// and the bytes it did put on the wire are already counted at the socket
+    /// layer by `record_packet_sent`.
+    pub fn record_transfer_failed(&self) {
+        // Saturating, matching `record_transfer_completed` — a pinned u32::MAX
+        // is a better failure mode than wrapping to 0 and reporting a healthy
+        // node.
+        self.transfers_failed
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |v| {
+                Some(v.saturating_add(1))
+            })
+            .ok();
+    }
+
     /// Record the start of an outbound NAT traversal attempt.
     pub fn record_nat_traversal_attempt(&self) {
         self.nat_traversal_attempts.fetch_add(1, Ordering::Relaxed);
@@ -462,11 +496,12 @@ impl TransportMetrics {
     /// Read-only snapshot for the local dashboard. Does NOT reset counters
     /// (unlike `take_snapshot` which is consumed by the telemetry worker).
     ///
-    /// **Telemetry interaction**: `peak_throughput_bps`, `avg_cwnd_bytes`,
-    /// `avg_rtt_us`, and `slowdowns_triggered` are period accumulators that
-    /// `take_snapshot` resets every `transport_snapshot_interval_secs`
-    /// (default 30s).  Between resets these reflect recent activity; the
-    /// dashboard sees the current window, not a lifetime aggregate.
+    /// **Telemetry interaction**: `transfers_completed`, `transfers_failed`,
+    /// `peak_throughput_bps`, `avg_cwnd_bytes`, `avg_rtt_us`, and
+    /// `slowdowns_triggered` are period accumulators that `take_snapshot`
+    /// resets every `transport_snapshot_interval_secs` (default 30s).  Between
+    /// resets these reflect recent activity; the dashboard sees the current
+    /// window, not a lifetime aggregate.
     /// `cumulative_bytes_sent/received` are never reset and reflect totals.
     pub fn read_snapshot(&self) -> TransportSnapshot {
         let transfers_completed = self.transfers_completed.load(Ordering::Relaxed);
@@ -938,6 +973,45 @@ mod tests {
 
         // Counters reset after the snapshot: a subsequent snapshot with no new
         // activity is None.
+        assert!(metrics.take_snapshot().is_none());
+    }
+
+    /// Failed transfers count as snapshot activity in their own right: a period
+    /// whose ONLY activity is failures (a sick node whose every transfer dies,
+    /// with no completions and no RTT keep-alive samples) must still emit a
+    /// snapshot, otherwise `take_snapshot` returns `None`, the telemetry worker
+    /// uploads nothing, and the fail counter reads a permanent 0 — exactly the
+    /// #4827 symptom this fix exists to surface.
+    ///
+    /// The `&& transfers_failed == 0` conjunct in the no-activity gate was DEAD
+    /// until #4827 gave the counter a writer, so this is its first coverage.
+    /// Sibling pins: `test_snapshot_emitted_for_rtt_only_activity` (#4000) and
+    /// `test_snapshot_emitted_for_nat_only_activity`.
+    ///
+    /// Also the only coverage that a failure reaches `take_snapshot` at all —
+    /// the value telemetry actually uploads, distinct from the `read_snapshot`
+    /// the `peer_connection` wiring tests read.
+    #[test]
+    fn test_snapshot_emitted_for_failed_transfers_only() {
+        let metrics = TransportMetrics::new();
+
+        // No completions, no RTT samples, no NAT activity — only failures.
+        metrics.record_transfer_failed();
+        metrics.record_transfer_failed();
+        metrics.record_transfer_failed();
+
+        let snapshot = metrics
+            .take_snapshot()
+            .expect("failure-only activity must still produce a snapshot");
+        assert_eq!(snapshot.transfers_failed, 3);
+        assert_eq!(snapshot.transfers_completed, 0);
+        // No completion means no measured duration: the average is the sentinel
+        // 0, not a real timing. `cards.rs` must not render it as one.
+        assert_eq!(snapshot.avg_transfer_time_ms, 0);
+
+        // Counters reset after the snapshot: a subsequent snapshot with no new
+        // activity is None. Without the reset a single failure would re-emit a
+        // stale snapshot every period forever.
         assert!(metrics.take_snapshot().is_none());
     }
 
--- a/crates/core/src/transport/peer_connection.rs
+++ b/crates/core/src/transport/peer_connection.rs
@@ -1183,14 +1183,55 @@ impl<S: super::Socket, T: TimeSource> PeerConnection<S, T> {
                             // Report to global metrics for periodic telemetry snapshots
                             super::TRANSPORT_METRICS.record_transfer_completed(&stats);
                         }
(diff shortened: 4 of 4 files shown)


sanity:
## Consolidated multi-model review — Full tier

Transport is a high-risk surface, so this ran the Full gate: four independent Claude lenses (skeptical bug-hunt, testing, code-first, big-picture), each blind to the others, plus an external **Codex** pass. Reviewed at `b0b8f038`; findings applied; delta re-reviewed at `5a7a9279`.

**Verdict: no blocking findings. The code is correct.** All four lenses independently confirmed the load-bearing judgment call and every reviewer recommended merge.

### The judgment call is right, and the intuitive alternative would have been a no-op

Instrumenting **both** `Err` arms (not just the terminal one) is correct and load-bearing. `is_transient_send_failure` matches `SendFailed | OutboundStreamFailed` — i.e. *every* common failure (cwnd-wait timeout, pipe stall, inbound-feed error, socket send error). The only non-transient case is a rare bincode `Serialization` error. **Instrumenting the "genuinely terminal" arm alone — the intuitive reading, and the one I initially suggested — would have left the counter ~always 0 and not fixed the bug.**

No double-count is possible: one spawned task yields one result, `FuturesUnordered` yields it exactly once, and nothing awaits between yield and increment. (`JoinHandle` isn't `Clone`, so re-pushing isn't even expressible.)

### Convergent finding — three reviewers, two model families

Codex [P2] and two Claude lenses **independently** found the same real gap: `PeerConnection::drop` never aborts or drains `outbound_stream_futures`, so dropping detaches those tasks. They run on, fire `emit_transfer_failed` from *inside* the task, and no `recv()` loop is left to observe the result — an **asymmetric** event-without-counter gap, concentrated exactly at teardown where mass failures cluster.

The original comment claimed the counter was "1:1 with the event stream" with "one symmetric gap". That was false, and it's the class of confident claim that hardens into fact. **Fixed** — both gaps are now disclosed at the call site and in the rustdoc.

Codex asked for the gap to be *closed*, not just documented. Deliberately deferred to **#4838**, with reasoning: `record_transfer_completed` has the identical gap (same observation site), so fixing only `transfers_failed` would make the two counters count different populations and skew the "N ok / M fail" ratio in a *new* way. Closing it properly means moving both counters into the stream tasks or draining handles at teardown — a transport-metrics refactor, not a counter-wiring fix. The undercount is bounded and directional (undercount only), and #4827's symptom is fully fixed independent of it.

### Findings applied (8)

| # | Finding | Fix |
|---|---|---|
| 1 | "1:1 with event stream / one symmetric gap" — **false** (teardown/detach gap) | Claim scoped; both gaps disclosed; also in rustdoc + PR body |
| 2 | "all six sites" — miscount (4 sites + 2 error kinds, summing to six coincidentally) | Cross-file census removed from comment, moved to PR body where it's accurate |
| 3 | "a retry opens a new stream" — mostly doesn't fire (`MAX_PEER_ADVANCEMENTS_STREAMING = 0`; GETs never streaming) | Replaced with the real guarantee (one task → one result → yielded once) |
| 4 | `take_snapshot`'s `&& transfers_failed == 0` conjunct was **dead until this PR**, now live and unpinned — both siblings have pin tests | Added `test_snapshot_emitted_for_failed_transfers_only`; also closes zero `take_snapshot`/reset coverage |
| 5 | `read_snapshot` doc omitted `transfers_*` from the reset list — newly misleading | Added |
| 6 | "single place every stream task's result is observed" — also observes inbound | Scoped to outbound |
| 7 | Test-helper doc's causal model wrong (arms *do* fire at 100ms; they're no-ops on an idle fixture) | Corrected |
| 8 | PR made `0 ok / 3 fail (**0.000s avg**)` reachable — a sentinel rendered as a measurement | `cards.rs` suppresses the avg span when `transfers_completed == 0` |

Finding 8 adds a third file. Justified rather than scope creep: this PR is what makes that branch reachable. The delta re-review confirmed the suppression gate is *exactly coextensive* with the sentinel case (`avg == 0` iff `completed == 0`, in both constructors), so **no real average can ever be hidden**.

### The 30s-flicker question, answered

Worth recording, since it looked like it might sink the PR. It doesn't, and the framing inverts: `take_snapshot` resets `transfers_completed` and `transfers_failed` **together**, and the row gates on their OR. Because `transfers_failed` was structurally 0, the gate was effectively `completed > 0` — so a node whose transfers were **all failing rendered no Data Transfer row at all**, which is issue #4827's headline complaint. This fix turns it into a genuine OR and **strictly increases** row visibility (`0 ok / M fail` now renders). Same for `take_snapshot`'s activity gate: a failure-only window previously emitted no snapshot at all.

Residual (pre-existing, card-wide, needs a design call): a user checking >30s after failures stopped sees an empty card. Captured in #4827's "Related" section.

### Tests

Both new tests were **falsified, not just run green** — deleting either call site fails with `left: 0, right: 1`; removing the gate conjunct fails the new snapshot test; forcing the avg span unconditionally fails the cards test. The tests drive the real `recv()` loop rather than calling the recorder directly (a direct-call test would pass with both call sites deleted, i.e. would not catch this bug), cover both arms independently, assert an exact `+1` delta (catching double-counts), and assert their own premise so a future reclassification can't leave them silently testing the wrong arm.

Note the pre-existing tests asserting `transfers_failed == 0` (`metrics.rs`) are **not** what let this ship — they're RTT-only and NAT-only activity tests where 0 is a correct premise. What let it ship is that nothing tested the wiring. That's now closed.

### CI note

NAT Validation failed on the first run and **passed on re-run at the identical commit with zero code changes** (`b0b8f038`, 7m30s). A sibling PR (#4832) hit the same two Docker NAT tests in the same ~20-minute window while eight runs across five other branches were green — consistent with an environmental blip, not either change. Recorded rather than waved away; if it recurs it deserves a real investigation, per the repo's flaky-tests rule.

Follow-ups filed: **#4838** (teardown undercount).

[AI-assisted - Claude]

Issue #4361, opened 2026-06-07
fix: nightly test_get_reliability_diagnostic at 18.3% — ring is empty for most nodes during the op window, so no GET ever traverses the network (unmasked by #4352)


sanity:
## Summary

The nightly simulation suite has failed three nights running (June 5, 6, 7) on `test_get_reliability_diagnostic` with a deterministic **18.3% GET success rate (13/71)** — below the 50% catastrophic threshold. Before June 5 the same test passed at 67.2%. The regression window is exactly the merge of **PR #4352** (June 4), but #4352 is not the defect — it unmasked a pre-existing one:

**During the test's entire operation window, most nodes have an empty ring** (`connections_by_location` = 0 entries) despite live transport/transient connectivity. Result: **not one GET in the failing runs ever traversed the network** — every "success" was a local cache hit on a node that already held the contract, and every failure was an originator with zero routing candidates that returned NotFound without sending a single wire request.

Failing runs (identical numbers, seed `0xDEADBEEF`, fully deterministic — also byte-identical when reproduced locally, down to the transaction ULIDs):
- https://github.com/freenet/freenet-core/actions/runs/27082691304 (June 7, commit 875b97a)
- https://github.com/freenet/freenet-core/actions/runs/27052452193 (June 6)
- https://github.com/freenet/freenet-core/actions/runs/26995429579 (June 5)

June 7 was the first night this reached the River dev room because the CI announcer (#4358) landed June 6 — the nightly has actually been red for 15+ consecutive days on varying tests; #4352 fixed the four `test_nightly_*` topology tests (#4348) and this GET test is the remaining red.

## Bisection

| Nightly | Commit | `test_get_reliability_diagnostic` | nodes with state |
|---------|--------|-----------------------------------|------------------|
| June 1  | bcfbf1c80 | FAIL ~42% | — |
| June 2  | eda67ac23 | FAIL 42.2% (38/90) | — |
| June 3  | 38ea793e2 | PASS 67.2% (41/61) | 50/100 |
| June 4  | 541ca6350 | PASS 67.2% (41/61) | 50/100 |
| June 5  | 124057436 | FAIL 18.3% (13/71) | 22/100 |
| June 6  | 717c8a55b | FAIL 18.3% (13/71) | 22/100 |
| June 7  | 875b97a47 | FAIL 18.3% (13/71) | 22/100 |

[...]

PR #3488, opened 2026-03-09
feat: delegate creation by delegates via V2 host function

iduartgomez:
## Problem

Delegates cannot create other delegates at runtime. This blocks use cases where a parent delegate needs to dynamically spawn child delegates (e.g., identity management creating per-app delegates).

## Solution

Implements issue #2833 with a new `create_delegate` V2 host function exposed to WASM delegates via the `freenet_delegate_management` import namespace.

**Host side (this PR):**
- New `native_api.rs` module implementing `__frnt__delegate__create_delegate` — validates WASM, constructs `DelegateContainer`, registers in delegate + secret stores
- Resource limits: max depth 4, max 8 creations per `process()` call (prevents fork bombs)
- Depth tracking via `DelegateExecContext` threaded through the call chain
- Integration with wasmtime engine linker for the new import namespace

**Stdlib side (freenet/freenet-stdlib#60, released as v0.2.1):**
- New error codes (`ERR_DEPTH_EXCEEDED`, `ERR_CREATIONS_EXCEEDED`, `ERR_INVALID_WASM`, `ERR_STORE_FAILED`)
- `DelegateCtx::create_delegate()` ergonomic wrapper

**Test delegate (`tests/test-delegate-creation/`):**
- WASM delegate that exercises `create_delegate` from within `process()`
- E2e test in `apps/freenet-ping/app/tests/run_delegate_creation_e2e.rs`

## Testing

- 73 delegate unit tests pass
- Test delegate WASM compiles and links against stdlib 0.2.1
- E2e test requires Linux loopback (macOS `os error 49`)

## Fixes

Closes #2833

PR #3444, opened 2026-03-06
fix: add GET retry with DBF fallback and auto-fetch from UPDATE senders

iduartgomez:
## Problem

GET operations for contracts could permanently fail when all 3 k_closest DHT peers were unreachable or didn't respond. This caused:

1. **Silent GET timeout** — client requests hung for 60s then silently failed with no retry
2. **Missing contract code after restart** — nodes receiving UPDATE broadcasts for contracts they hadn't cached couldn't apply updates and had no mechanism to fetch the missing contract
3. **Routing black hole** — for contracts with limited replication (e.g., new River channels), if the 3 closest connected peers didn't have the contract, the GET would always fail even though other connected peers had it

Observed in production: contract `7bQcxZdnuHpYdQgATTM9KzxhV9bsH55YeJx9pJ9rKfDM` (River channel) consistently failed GET via k_closest routing, while UPDATE broadcasts from 7+ peers confirmed the contract existed on the network.

## Solution

Three complementary mechanisms:

### 1. GET retry with k_closest alternatives (op_state_manager.rs)
The GC task (runs every 5s) checks for GET operations older than 20s. If the operation has untried alternative peers from the original k_closest selection, it retries with the next one. Retries are spaced at 20s intervals (first at 20s, second at 40s, etc.) tracked per-transaction.

### 2. DBF (Depth-Breadth-First) fallback (get.rs)
When all k_closest alternatives are exhausted, `retry_with_next_alternative` accepts fallback peers (all connected peers). It filters out already-tried addresses and injects new candidates. This broadens the search beyond the DHT-optimal 3 peers.

### 3. Auto-fetch from UPDATE senders (update.rs)
When `update_contract` fails in `BroadcastTo` or `BroadcastToStreaming` handlers (typically "contract parameters not found"), `try_auto_fetch_contract` creates a targeted GET directly to the UPDATE sender — who is known to have the contract. Rate-limited to 1 attempt per contract per 5 minutes via `pending_contract_fetches` DashMap.

Key insight: the UPDATE and GET routing systems are independent. UPDATEs reach nodes via proximity cache (populated during CONNECT handshakes), but GETs use k_closest DHT routing. A node can receive UPDATEs from peers it would never select for GET routing.

## Testing

**Production verified:** Deployed to a live node and confirmed:
- GET retry fired at 20s for contract 7bQcxZ, second alternative peer responded with 681KB streaming response
- DBF fallback activated for contract B1ZDmn — broadened to 4 new candidates after 3 k_closest exhausted
[...]

PR #4840, opened 2026-07-16
test: decouple sparse-topology GET test from scheduler ordering

sanity:
## Problem

`test_get_retry_with_alternatives_sparse_topology` was seed-coupled and brittle. It obtained its sparse-hosting precondition **by luck**: it hoped a seeded PUT at HTL=2 would happen to cache the contract at exactly 2 of 10 nodes, and that 8 of 10 requesters would happen to need an alternative to reach a cacher. That premise is a lottery over the seeded RNG stream.

It has now broken three times on that basis:

- `#[ignore]`d in #3488 after a freenet-stdlib bump shifted bincode sizes and perturbed the RNG stream.
- Re-seeded in #3506 to `SEED = 0x3506_000B_0001`, chosen so an HTL=2 PUT cached at exactly 2 of 10 nodes.
- Now blocking #4821.

A bisect showed the coupling is to the **tokio scheduler**, not to routing quality. On this branch's base (`0f856ab`), deleting the whole `refit_router_periodically` background task — a task that ticks every 5 minutes while these sims finish in well under 5 minutes wall-clock, so it **never fires** — flips the old test from pass to fail. Its mere existence perturbs the scheduler, which perturbs the seeded RNG stream, which reshuffles topology formation.

A fourth re-seed was rejected; this fixes the root cause instead.

## Solution

Rebuild the scenario so the precondition holds **by construction**. Peer locations, peer connections and contract placement are all set explicitly, so no assertion depends on the topology the RNG produces:

- **decoy** sits closest to the key and never hosts the contract, so greedy keyward routing always lands on it first and it cannot answer.
- **two holders** sit further from the key than the decoy but closer than the requester, seeded directly via `SeedDemandlessCopy`.
- **one requester** sits far from the key, wired to the decoy and both holders; the decoy is wired to both holders.

The route is forced: `requester -> decoy (cannot serve) -> holder (serves)`. Reaching the state requires leaving the first-choice candidate.

This is **test-only** — no production code is touched. The runtime also drops from ~16s to ~0.4s.

Three mechanics are worth recording, each verified against the code while building this:

1. **`SeedDemandlessCopy`, not `SeedHostedContract`, places the holders.** The relay serve gate is `check_local_with_interest_gate`, which keys on `interest_manager.has_local_interest`. `SeedMode::Demandless` calls `register_local_hosting` so the gate passes. `SeedMode::Subscribed` — what `SeedHostedContract` uses — calls `ring.host_contract` + `ring.subscribe` but never `register_local_hosting`, so a `SeedHostedContract` holder logs `"stale cache (no local interest), will forward"` and refuses to serve, despite its rustdoc describing it as the genuine-host primitive.
2. **`ring_max_htl` is 3** because a relay short-circuits on `htl == 0` and answers NotFound *before* checking its local copy (`"Short-circuit 1: HTL = 0"` precedes `"Short-circuit 2: Local cache hit"`). A holder must arrive with `htl >= 1` to be allowed to serve: requester (3) -> decoy (2) -> holder (1).
3. **Exactly one requester.** A GET's return path caches the state at each hop, so the decoy holds the contract once the first GET completes and stops being a decoy. Additional requesters get served at hop 1 and exercise nothing.

## Rename: the test is now named for what it drives

The test is renamed `test_get_retry_with_alternatives_sparse_topology` -> **`test_get_reaches_sparse_holder_via_relay_forwarding`**, and its doc says plainly what it does and does not cover.

The old name claimed coverage of the client-side alternative-retry path (`get::op_ctx_task::advance_to_next_peer`). It never had that coverage — see below — and the old test was passing on relay-side forwarding too. So this PR does not remove that coverage; it reveals it never existed. Note `advance_to_next_peer` (client-side, `op_ctx_task.rs:1888`) and `relay_advance_to_next_peer` (relay-side, `:2744`) are **distinct functions**: this test depends on the relay one (proven by mutation, below) and not on the client one.

Renaming is the honest fix. A test whose name claims coverage it does not provide is a plausible claim that hardens into fact and misleads the next reader.

## Scope correction: what this test does *not* cover

The old name and doc claimed coverage of the client-side retry loop (`get::op_ctx_task::advance_to_next_peer`). **That path cannot steer a retry to a different peer on the normal (non-empty-ring) path**, so no assertion about retry-driven success can be satisfied by construction:

- A client GET is sent via `ctx.send_and_await`, which loops back to the originator's own relay. `drive_client_get_inner` says so directly: *"Pre-select initial target for the driver's retry state. Actual routing is done by `process_message` on the loop-back; this is just so `advance_to_next_peer` has a starting `tried` set."*
- The loopback relay re-picks the wire hop greedily each attempt, and the client's `tried` set is carried into the attempt's visited bloom **only when `connection_count() == 0`** (the #4361 bootstrap case, explicitly *"gated on the empty-ring case so normal-path retry routing semantics are unchanged"*).
- `MAX_RELAY_RETRIES = 1`, so relays do not try alternatives either; they forward once and bubble NotFound, deferring cross-peer retry to the client driver.

Net effect: every client retry re-picks the same first candidate. Observed directly — at `ring_max_htl = 1` a requester makes 4 attempts with fresh transactions and **all four forward to the same decoy**, then exhaust. The mechanism that actually delivers a sparse-hosting GET is the greedy multi-hop walk, and that is what this test now pins and says it pins.

Because the path is inert, **no passing end-to-end test can exist for it** — you cannot assert on a mechanism that never fires. So the client-side retry has **no end-to-end coverage today**, and cannot until #4841 is fixed. When it is, that coverage belongs in a **new** test rather than a widening of this one; the doc comment says so.

This is a behavioural finding about #3444's retry path being effectively inert on the normal path. It is **not** addressed here (this PR stays test-only) — filed as #4841.

## Testing

The acceptance criterion was that the test must pass both on pristine base **and** under the scheduler perturbation that breaks the old one — passing only the former would mean re-tuned, not decoupled. All runs below are real observed output.

| Build | New test | Old test |
|---|---|---|
| pristine base (`0f856ab`) | **pass 5/5** | pass |
| base + `refit_router_periodically` task removed | **pass 5/5** | **FAIL 3/3** |
| full `refactor/4811-refit-in-add-event` (#4821) | **pass 5/5** | **FAIL 2/2** |

So this unblocks #4821.

One correction to the bisect as previously recorded: removing *only* the `task_monitor.register(...)` wrapper while keeping the `GlobalExecutor::spawn` does **not** flip the old test (it passes 3/3). That is self-consistent — the task still exists, so the scheduler is unperturbed. The perturbation is removing the **task**, which is what #4821 actually does.

**The traversal is genuinely two network hops**, confirmed from the GET trace rather than assumed — the decoy is neither skipped nor able to serve:

| Hop | Peer | htl | Behaviour |
|---|---|---|---|
| 1 | requester loopback relay | 3 | `forwarding request to downstream peer target=...:21192` (decoy) |
| 2 | **decoy** `:21192` | 2 | cannot serve -> `forwarding request to downstream peer target=...:24148` |
| 3 | **holder A** `:24148` | 1 | `contract found locally (active interest) — sending Found upstream` |

followed by `downstream returned Found — forwarding upstream then caching locally` at the decoy — which is what the traversal assertion binds to. The holder arriving at `htl=1` also confirms the HTL=3 reasoning. The ~0.4s runtime reflects 4 nodes / 1 GET with preseeded connections and no PUT, not a degenerate one-hop path.

**Coverage is not vacuous.** Both assertions were verified to fail when the mechanism they cover is broken:

- Stubbing `relay_advance_to_next_peer` to never forward -> fails with *"the requester did not get the contract state ... the GET gave up at the first candidate instead of routing on to a holder"*.
- Disabling the relay's return-path caching -> the requester still gets the state (first assertion passes) but the traversal assertion fires: *"the decoy has no state at the end of the run, so the successful GET did not pass through it"*. This confirms the two assertions are live and independent.

`cargo fmt --check` clean; `cargo clippy --locked -- -D warnings` (the CI form) clean.

### Pre-existing suite flakiness (not from this change)

The full `simulation_integration` suite is flaky in parallel on `main` today, independent of this PR. Same command, back to back:

- pristine `main`: 129 passed, **6 failed** (`test_injected_hosting_clock_drives_subscription_lease_lapse`, `test_piece_e_unsubscribed_stale_serve_gate`, `test_placement_migration_at_scale_renewal_load_stays_bounded`, `test_strict_determinism_exact_event_equality`, `test_subscription_root_renewal_does_not_storm`, `test_topology_subscribe_health`)
- this branch: 130 passed, **5 failed** (4 of the same, plus `test_direct_runner_churn_actually_drops_packets`, minus two of main's)

The failure set varies run to run and each of them passes when run alone, so this is pre-existing parallel-execution flakiness rather than a regression here. `test_get_retry_with_alternatives_sparse_topology` passed in the full-suite run on this branch. Flagging rather than folding it in, since it is a separate problem with a separate cause.

Closes nothing; unblocks #4821.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/tests/simulation_integration.rs
+++ b/crates/core/tests/simulation_integration.rs
@@ -8794,6 +8794,23 @@ fn test_relay_route_events_multihop() {
 ///   stops being a decoy. Additional requesters would be served by the decoy at
 ///   hop 1 and would exercise nothing.
 ///
+/// * **Routing is greedy, but NOT because of `rnd_if_htl_above`.** That parameter
+///   is inert: it is plumbed to `ConnectionManager` and read exactly once
+///   (`connection_manager.rs`, in the `should_accept` debug line) as a **tracing
+///   log field**. There is no `htl > rnd_if_htl_above` branch anywhere in the
+///   crate, so no value of it makes routing non-greedy. The value passed here is
+///   arbitrary — do not "tune" it expecting a routing change.
+///
+/// * **Route determinism depends on staying under ~10 routing events.**
+///   `Router::select_peer`'s sub-50-event path sorts untried-before-tried
+///   (`fa.cmp(&fb).then_with(|| da.cmp(db))`) — i.e. *ahead of* distance — and
+///   `peer_adjustments` is populated for every peer once the global regression
+///   reaches `ADJUSTMENT_PRIOR_SIZE` (10) events. So in the 10..50 regime a peer
+///   with history sorts behind one without, regardless of distance, which could
+///   route straight to a holder and skip the decoy. One GET is ~1-3 events, so
+///   there is comfortable margin, and the traversal assertion below fails loudly
+///   if it is ever crossed. Anyone adding operations to this test should know.
+///
 /// # Why it is built this way
 ///
 /// This test previously obtained its sparse-hosting precondition **by luck**: it
@@ -8884,7 +8901,7 @@ fn test_get_reaches_sparse_holder_via_relay_forwarding() {
             1,         // 1 gateway
             num_nodes, // 4 regular nodes
             3,         // ring_max_htl — see "Why the mechanics are what they are"
-            3,         // rnd_if_htl_above — >= ring_max_htl, so routing is always greedy
+            3,         // rnd_if_htl_above — INERT, see below; value is arbitrary
             12,        // max_connections — ample, so no preseeded edge is capped away
             3,         // min_connections
             SEED,
@@ -8912,7 +8929,23 @@ fn test_get_reaches_sparse_holder_via_relay_forwarding() {
     let decoy_dist = ring_dist(locs[1], key_loc);
     let holder_dists = [ring_dist(locs[2], key_loc), ring_dist(locs[3], key_loc)];
     let requester_dist = ring_dist(locs[4], key_loc);
+    // The gateway's location is NOT chosen by this test — it is derived from the
+    // seed — and preseeding does not suppress organic CONNECT, so the requester
+    // does connect to it. That makes the gateway a live routing candidate whose
+    // distance to the key is a seed-derived quantity, i.e. exactly the kind of
+    // dependency this test exists to remove. Assert it rather than rely on the
+    // margin: HTL has zero slack here (requester 3 -> decoy 2 -> holder 1), so a
+    // gateway hop inserted into the route would exhaust HTL and yield NotFound.
+    let gateway_dist = ring_dist(locs[0], key_loc);
     for (i, holder_dist) in holder_dists.iter().enumerate() {
+        assert!(
+            *holder_dist < gateway_dist,
+            "scenario setup wrong: the seed-derived gateway landed closer to the key \
+             than holder {i}, so it can win a greedy pick and insert a hop into the \
+             forced route. With no HTL slack that turns the GET into a NotFound. \
+             Re-pick SEED or move the holders. \
+             holder_dist={holder_dist}, gateway_dist={gateway_dist}"
+        );
         assert!(
             decoy_dist < *holder_dist,
             "scenario setup wrong: the decoy must be strictly closer to the key than \
--- a/crates/core/tests/simulation_integration.rs
+++ b/crates/core/tests/simulation_integration.rs
@@ -8735,132 +8735,271 @@ fn test_relay_route_events_multihop() {
     );
 }
 
-/// Regression test for GET retry with alternatives in sparse topologies.
+/// Sparse-hosting GET reachability via **relay-side multi-hop forwarding**.
 ///
-/// With HTL=2 and 10 nodes, PUT only caches the contract at 2-3 nodes along the
-/// route. When distant nodes GET, their initial routing candidate may not have
-/// the contract. The fix ensures the GET state machine tries alternative peers
-/// (from k_closest) and re-queries k_closest when alternatives are exhausted,
-/// rather than immediately returning NotFound.
+/// A GET whose **first-choice candidate does not host the contract** must still
+/// resolve to a peer that does, instead of handing the client a NotFound. With
+/// hosting sparse, the peer closest to a contract's key is frequently not a
+/// holder, and a GET that gives up there fails for a contract that plainly
+/// exists one hop away. The mechanism that carries the GET past that first
+/// candidate is a relay forwarding onward toward the key
+/// (`get::op_ctx_task::relay_advance_to_next_peer` and the HTL walk), and that
+/// is what this test drives.
 ///
-/// Without the fix: distant nodes return NotFound because the first-choice peer
-/// doesn't have the contract and there's no retry.
-/// With the fix: the GET retries with alternative peers until one succeeds.
+/// **This test does NOT cover client-side alternative-retry (#3444)** — i.e.
+/// `get::op_ctx_task::advance_to_next_peer` — and is deliberately not named for
+/// it. That path is inert on the normal (non-empty-ring) path: see #4841 and the
+/// "Scope" section below. This test stays green if the client-side retry is
+/// deleted entirely. It was previously called
+/// `test_get_retry_with_alternatives_sparse_topology`, a name that claimed
+/// coverage it never had — the old version passed on relay forwarding too. When
+/// #4841 is fixed, the client-retry coverage belongs in a NEW test; do not
+/// stretch this one to claim it.
 ///
-/// Exercises: `get.rs` alternative retry (line ~1521) and k_closest re-query (line ~1569)
+/// # Scenario — established by construction, not by seed luck
 ///
-/// Seed history: this test was `#[ignore]`d in PR #3488 because the original
-/// seed (`0xBEEF_CAFE_0001`) stopped producing a sparse-but-reachable topology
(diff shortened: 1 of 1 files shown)


sanity:
Filed the client-side retry finding from the "Scope correction" section as #4841, so it is tracked separately rather than folded into this test-only PR.

[AI-assisted - Claude]

sanity:
## Re: the P2 on retained test name — valid, fixed by renaming (not by manufacturing a retry)

Codex is right about the substance and I've fixed it, though not the way the suggested remedy asks.

**What's correct:** this test remains green if `advance_to_next_peer` and all client-side alternative-retry behaviour are removed, so keeping the name `test_get_retry_with_alternatives_sparse_topology` claimed coverage it does not provide. Fixed in `ebb9665`: renamed to **`test_get_reaches_sparse_holder_via_relay_forwarding`**, with a doc section stating plainly what it does and does not cover.

Note this does not conflict with the mutation check in the PR body — `advance_to_next_peer` (client-side, `op_ctx_task.rs:1888`) and `relay_advance_to_next_peer` (relay-side, `:2744`) are **distinct functions**. The test depends on the relay one (stubbing it fails the test) and not on the client one. Both observations hold.

**Why the suggested remedy — "a deterministic scenario where the first route fails and a distinct retry candidate serves" — is not implementable:** that scenario cannot be constructed, because the client-side retry cannot steer to an alternative peer on the normal path:

- a client GET loops back to its own relay, which re-picks the wire hop greedily (*"Actual routing is done by `process_message` on the loop-back; this is just so `advance_to_next_peer` has a starting `tried` set"*);
- `carry_tried_into_visited` is gated on `connection_count() == 0` (the #4361 bootstrap case, explicitly *"gated on the empty-ring case so normal-path retry routing semantics are unchanged"*);
- `MAX_RELAY_RETRIES = 1`, so relays defer cross-peer retry to a client driver that cannot perform it.

Observed directly: at `ring_max_htl = 1`, a requester makes **4 attempts with fresh transactions, all four forwarded to the same decoy**, while two directly-connected holders are never contacted. Filed as #4841.

**So the coverage in question was never there to lose.** If the client retry is inert, the pre-existing test was also passing on relay-side forwarding — it just claimed otherwise in its name. This PR doesn't remove that protection; it reveals it never existed. Manufacturing a "retry" scenario that passes would either not be constructible, or would pass for a reason unrelated to #3444 — which is worse than an honest name.

The client-retry path therefore has **no end-to-end coverage today, and cannot have any until #4841 is fixed**. When it is, that coverage belongs in a new test rather than a widening of this one. Recorded on #4841 and in the test's doc comment.

[AI-assisted - Claude]

sanity:
## Review follow-ups pushed (`0e9e2fa6`)

Both items in, each verified against the code rather than taken on faith.

**1. Gateway distance is now asserted.** The gateway's location is seed-derived, not chosen by this test, and preseeding does not suppress organic CONNECT — so the requester connects to it and it is a live routing candidate. The ~8x margin that stops it winning a greedy pick today (gateway ~0.164 vs decoy ~0.00199) is itself a seed-derived quantity, which is precisely the dependency this PR exists to remove. With zero HTL slack (requester 3 → decoy 2 → holder 1), an inserted gateway hop exhausts HTL and yields NotFound.

Added `holder_dist < gateway_dist` for both holders, with a message naming the fix. **Verified it binds**, rather than assuming: pushing the holders past the gateway trips it —

```
scenario setup wrong: the seed-derived gateway landed closer to the key than holder 0,
so it can win a greedy pick and insert a hop into the forced route. With no HTL slack
that turns the GET into a NotFound. Re-pick SEED or move the holders.
holder_dist=0.3000156595038689, gateway_dist=0.16399002175084132
```

**2. The `rnd_if_htl_above` comment described a mechanism that does not exist.** Confirmed independently: the parameter is plumbed to `ConnectionManager` and read **exactly once** (`connection_manager.rs:735`) as a **tracing log field** in the `should_accept` debug line. `grep -rE "htl\s*>\s*.*rnd|rnd.*<\s*htl" crates/core/src/` returns nothing — there is no such branch in the crate. The old comment's conclusion (routing is greedy) was true; its stated reason was fiction, and a future reader could have "tuned" the value believing they'd changed routing. It now says the parameter is inert and the value is arbitrary.

**3. Documented the routing-determinism dependency** (informational): `Router::select_peer`'s sub-50-event path sorts untried-before-tried (`fa.cmp(&fb).then_with(|| da.cmp(db))`) *ahead of* distance, and `peer_adjustments` fills for every peer at `ADJUSTMENT_PRIOR_SIZE` (10) events. In the 10..50 regime a peer with history sorts behind one without regardless of distance, which would route straight to a holder and skip the decoy. One GET is ~1-3 events, so there is margin, and the traversal assertion fails loudly if it is ever crossed. Noted for anyone adding operations here.

### Re-verified on the final content

Content changed, so the earlier runs were stale. All re-run against `0e9e2fa6`:

| Build | Result |
|---|---|
| pristine main | **pass 5/5** |
| main + `refit_router_periodically` removed | **pass 3/3** |
| full `refactor/4811-refit-in-add-event` | **pass 3/3** |
| mutation (stub `relay_advance_to_next_peer`) | **FAILS** |

`cargo fmt --check` and `cargo clippy --locked -- -D warnings` clean. Production still byte-identical to `main` — this remains test-only.

[AI-assisted - Claude]

Issue #1499, opened 2025-03-24
Delegates should be able to interact with users


sanity:
Complete the implementation of delegate-user interaction. 

The purpose of this is so that delegates can seek permission from the user for tasks like signing with a private key.

sanity:
🤖 **Relabeled for consistency**

Applied labels:
- `T-feature`
- `P-medium`
- `E-medium`
- `A-crypto`
- `A-developer-xp`

**Reasoning:** This issue requests completing delegate-to-user interaction so delegates can request permissions (e.g. signing with a private key). That's new functionality (T-feature) rather than a bug. It doesn't appear release-blocking, so P-medium is appropriate. Implementing user interaction around cryptographic signing is of moderate complexity (E-medium) and touches cryptography (permissioned signing) and developer/user experience (prompt/UX/API).

Previous labels: ``

sanity:
## Implementation Specification

### Problem

`RequestUserInput` is currently stubbed -- the runtime auto-responds with `Allowed` at `crates/core/src/wasm_runtime/delegate.rs:348-360` without ever prompting the user. This is needed for the ghostkey delegate, which must ask user permission before signing or revealing identity to applications. Since any web page can trigger delegate messages (no explicit app install), signing for any requestor without consent is a security/privacy leak.

### Motivating Use Case: Ghostkey Delegate

The ghostkey delegate stores anonymous identities (Ed25519 signing keys). When an app (e.g. Harvest marketplace) requests a signature, the delegate needs to prompt:

> "Harvest wants to use your ghostkey 'Trading Identity'"
> [Allow Once] [Always Allow] [Deny]

Without this, visiting any Freenet web app could silently probe for ghostkey existence or obtain signatures.

### Architecture

```
Delegate emits RequestUserInput
    |
    v
Runtime (freenet-core) routes to native prompt system
    |
    v
System tray / native dialog (per-OS)
    |
    v
User responds
    |
    v
Runtime sends UserResponse back to delegate as InboundDelegateMsg::UserResponse
```

### What Needs to Change

**1. Runtime (`crates/core/src/wasm_runtime/delegate.rs`)**

Replace the auto-allow stub (lines 348-360) with actual routing:
- When delegate emits `OutboundDelegateMsg::RequestUserInput(req)`, don't auto-respond
- Route the request to the native prompt system
- Park the delegate execution (store context) until the user responds
- On user response, resume delegate with `InboundDelegateMsg::UserResponse`

[...]

Issue #208, opened 2022-07-19
possibility to replace a key in the address with its alias


kakoc:
While running a contract through a browser in order to execute a contract I should specify its key. Like: `127.0.0.1/Eu4ByDpJ7mFdyeMHZ8z9WcDQ4wP2i5hTXSac1ZKQ2R9Q/state.html`. For the end user it's not so clear what it means(arbitary symbols). I think it's more comfortable for the end user to specify something more meaningful for him. 
Curious: can we introduce aliases? So that `127.0.0.1/Eu4ByDpJ7mFdyeMHZ8z9WcDQ4wP2i5hTXSac1ZKQ2R9Q/state.html` could be replaced with `127.0.0.1/my-app-name/state.html` for example. As I see eth, near have such functionality so probably it also can work for us the same way?

iduartgomez:
This may be possible for the curated contracts that will be present at the gateway; other than that adding aliases would require adding additional mechanisms for contract puts and ensure that name collision can't happen.

@sanity thoughts on this?

ple1n:
Classic decentralized naming problem. The only serious solution to this problem, ie. Zooko's triangle, is blockchain.

Alias doesn't need a consensus, so it's much easier.

> As I see eth, near have such functionality so probably it also can work for us the same way?

No, you have to introduce a blockchain system, or possibly use some web of trust based naming system.

sanity:
We are planning a naming service where people can reserve a "name" (similar to a username or domain name) in return for a donation. The web interface should look up names using this which should allow the use of friendly names like this.

Obviously it's centralized but I think that's inevitable for something like this.

ple1n:
Other forms of naming services, in the long term.

1. Allow contracts to resolve names
2. Use external blockchains to resolve names

iduartgomez:
Recently this has come out on discussion, somethings that we may do:
- have a contract repository contract which is officially maintained, where aliases can be assigned to contracts, this wouldn't be an issue
- this contract would allow version control and specifying dependencies/contracts by (alias, version), which would be automatically handled by the repository-contract
- publishing packages may be tied to the reputation system to avoid issues like name squatting
- we can provide the temaplte contract in case people ant to run their own repositories, aside of the official one (like you would have different package repositories for your OS)

sanity:
🤖 **Relabeled for consistency**

Applied labels:
- `T-feature`
- `P-low`
- `E-medium`
- `A-contracts`

[...]

PR #4844, opened 2026-07-16
fix(hosted): route programmatic window.open through the shell to avoid new-tab dead-end (#4645)

### LINKED ISSUE #2: NAT traversal

### LINKED ISSUE #4645: Suggestion and Issue with using try.freenet.org

sanity:
## Problem

On a hosted node (try.freenet.org) the contract app runs in a **sandboxed iframe with an opaque origin** (per-contract data isolation). The per-user access key lives in `localStorage`, which an opaque-origin context can't read.

Any new tab/window spawned **from inside that iframe** inherits the sandbox — there is deliberately no `allow-popups-to-escape-sandbox` (removed for security in #1499) — so the new tab has `window.origin === "null"`, can't read the key, and **fails closed to the "Open this app in a normal tab" page**. Its address bar shows the clean URL, so the user thinks it's a fresh tab and that reloading / re-pasting should work. It doesn't: sandbox flags are sticky to that browsing context, so only a genuinely new Ctrl/Cmd+T tab escapes. This is the unresolved core of #4645; #4652 only reworded the page.

Reproduced live (Chromium + Firefox): a fresh top-level paste of the URL loads fine (`window.origin` = `https://try.freenet.org`); a popup opened from the sandboxed iframe hits the page with `window.origin === "null"` even though the address bar shows the exact clean URL.

Link clicks (left **and** middle) are already intercepted and reopened with a real origin. Two gaps remained:
1. **Native right-click → "open in new tab"** — not interceptable from JS (only the recovery page can help).
2. **Programmatic `window.open()`** from the app's own JS — bypasses the click/auxclick listeners. **This PR fixes #2.**

## Approach

The gateway already injects `navigation_interceptor.js` into the sandboxed iframe. Extend it to **override `window.open`**:

- **http(s) targets** (relative resolved to absolute against `document.baseURI`) are forwarded through the same **`open_url` shell bridge** the cross-origin anchor path already uses, so the **shell (real origin)** opens a normal, working tab.
- **Non-http(s) targets** (`about:blank`, empty, `javascript:`, `blob:`, `data:`) fall back to the native open, so unrelated behavior is unchanged.
- `open_url`'s existing scheme allow-list is the security gate for the forwarded case, exactly as for anchor clicks.
- The returned `WindowProxy` is dropped (return `null`): a popup opened from here is a *different* opaque origin the opener could never script across that boundary anyway, and `null` matches the shell's `noopener` open.

This also generalizes cleanly: a programmatic cross-origin `window.open` now behaves like an already-intercepted cross-origin link click (same CORS fix as #208/#1499), on hosted and local nodes alike.

Secondary: **clarify the recovery page** for the residual right-click case — state plainly that *this* tab is the one stuck and that reloading/retyping the URL here keeps failing; open a brand-new tab.

## Behavior change / review notes (this touches the hosted-mode sandbox surface)

- `window.open` inside a contract iframe now returns `null` for http(s) targets and opens a real-origin tab via the shell instead of a sandbox-inherited popup. An app that relied on the returned window handle (e.g. popup-postMessage flows) will get `null` — but such a popup was already a dead opaque-origin context on a hosted node, so this is not a regression there; it's an improvement (the tab now actually works). Called out for reviewers.
- No wire/protocol change. Affects only the injected iframe JS + the shell error page.

## Testing

- **New regression test** `navigation_interceptor_overrides_window_open` pins the override (forwards via `open_url`, resolves relative against `baseURI`, http(s)-only, native fallback). Updated the `fail_closed_page_gives_actionable_recovery_4645` copy assertions.
- **Behavioral verification in a real opaque-origin sandboxed iframe** (`window.origin === "null"`) via Playwright: relative + absolute http(s) forward with an absolute resolved URL and return `null`; `about:blank`/empty/`javascript:`/`blob:` fall back to native and are not forwarded. All pass.
- Full `server::path_handlers::tests` module green (84 passed).

## Deploy note

try.freenet.org is a manual from-source build on nova, so this reaches it only when that build is rebuilt (not on merge to main).

**Draft** pending full multi-model review + @sanity sign-off, since it touches the hosted sandbox surface. Residual (native right-click "open in new tab") is inherently uninterceptable from JS and is left to the (now clearer) recovery page.

Closes #4645

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -4364,10 +4364,12 @@ mod tests {
              fall back to native (#4645)"
         );
         // In-place navigation targets are not new-window requests -> native.
+        // Names are normalized (case-insensitive) before the reserved check.
         assert!(
-            override_block.contains("name === '_self'"),
-            "window.open override must leave _self/_parent/_top to native so \
-             in-place navigation isn't turned into a new tab (#4645)"
+            override_block.contains("targetName === '_self'")
+                && override_block.contains("String(name).toLowerCase()"),
+            "window.open override must leave _self/_parent/_top (case-insensitive) \
+             to native so in-place navigation isn't turned into a new tab (#4645)"
         );
         // Loopback targets must fall back to native: open_url refuses them, so
         // forwarding would silently drop the open on local nodes.
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -4502,15 +4502,15 @@ mod tests {
         );
     }
 
-    /// `URL.hostname` strips the brackets from IPv6 literals, so a URL
-    /// `http://[::1]/` parses with `hostname === '::1'` (no brackets), and
-    /// the previous comparison `h === '[::1]'` never matched. The IPv6
-    /// loopback was effectively unblocked. The relaxation in
-    /// freenet/river#231 (accepting http: in addition to https:) makes
-    /// the gap newly reachable for typical home services, so fix it
-    /// alongside the http: change.
+    /// WHATWG `URL.hostname` serializes an IPv6 literal WITH brackets, so
+    /// `new URL('http://[::1]/').hostname === '[::1]'`. The handler must
+    /// therefore STRIP the brackets before comparing against `::1`, or the
+    /// loopback refusal never matches and a forged link to the viewer's IPv6
+    /// loopback slips through. (An earlier version of this test and the code
+    /// comment both had the fact inverted — asserting hostname is bracket-LESS —
+    /// so the test passed while the IPv6 loopback was in fact unblocked. #4645.)
     #[test]
-    fn shell_open_url_handler_blocks_ipv6_loopback_without_brackets() {
+    fn shell_open_url_handler_blocks_ipv6_loopback() {
         let js = SHELL_BRIDGE_JS;
         let open_url_idx = js
             .find("msg.type === 'open_url'")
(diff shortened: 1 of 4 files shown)


sanity:
## Multi-model review — consolidated

Full-tier review (hosted sandbox surface). Reviewers: **Codex** (external, 5 passes) + **4 independent Claude perspectives** (security/adversarial, code-first, testing, big-picture) + the CI **rule-review** bot + a focused adversarial re-verify of the round-3 delta. Every finding below is fixed or has a specific justification; the code was re-reviewed after each change per the per-content review rule.

### Security — clean
No exploit. The `window.open` override adds no capability a malicious contract lacks (it can already post `open_url` directly), no token is ever placed on the opened URL, and the shell re-validates every forwarded URL (scheme allow-list + loopback + `noopener`). Dropping the `WindowProxy` (null) is a slight improvement (severs the opener link). Sandbox is **not** weakened — no `allow-popups-to-escape-sandbox`; the pinning test for its absence still passes.

### Functional findings, all fixed (with behavioral tests)
| # | Finding | Fix |
|---|---------|-----|
| Loopback silent-drop | same-origin `window.open('/page')` on a localhost node hit `open_url`'s loopback block → silent no-op | fall back to native for loopback hosts |
| `_self`/`_parent`/`_top` (+ case-insensitive `_SELF`) | in-place nav was turned into a new tab | normalize name; route to native |
| URL-object arg | `window.open(new URL(...))` fell to native, re-creating the dead end | string-coerce the arg |
| `window.open(location.href)` (**P1**, common) | absolute same-origin URL kept `?__sandbox=1` → shell redirected to root, dropping subpath + invitation | strip `__sandbox` from the **resolved** URL when same-origin |
| External query corruption | `searchParams.delete()` reserialized every forwarded query (`%20`→`+`, `~`→`%7E`), breaking signed links | string-surgery strip; same-origin only; external untouched byte-for-byte |
| Bracketed IPv6 loopback | `URL.hostname` keeps `[::1]` → never matched → forwarded | strip brackets before comparing |
| Nested descendant frames | interceptor injected two-deep posts `open_url` to the app frame (lost) | only intercept the shell's direct child (`parent === top`) |
| `window.open(null)` | should coerce to the string `"null"` (native), not be treated as omitted | only `undefined` short-circuits |
| `<base href>` origin spoofing | gateway origin from `document.baseURI` is `<base>`-mutable | derive it from `location.href` (immune to `<base>`) |
| Stale docs | module header / rustdoc / injection comment described only `<a>` clicks | updated to cover the `window.open` override |

Behavioral coverage: a new Playwright spec (`crates/core/tests/playwright/tests/window-open.spec.ts`, 9 tests, runs in CI via `playwright-shell.yml`) exercises forwarding, every fallback path, query preservation, same-origin strip, nested frames, and the top-level guard; Rust source pins the load-bearing invariants.

### Pre-existing bug fixed here
The shell's own `open_url` IPv6 loopback check had the same bracket bug **and a factually-inverted comment**, so `http://[::1]:PORT/` slipped past the refusal (reachable via the anchor forwarder / direct post). Its test passed while the gap was open (false assurance). Fixed the check, comment, and test.

### Deferred (justified) → follow-up issues
- **#4845** — inject the sandbox shims before app head scripts (pre-existing, WS-shim-shared, rare saved-reference trigger; broader injection-ordering change).
- **#4846** — complete `open_url` loopback allow-list (127.0.0.0/8, IPv4-mapped IPv6) + add a rate-limit (both pre-existing LOW).

### CI
NAT Validation flaked once (an unrelated Docker NAT subscription test) and passed on re-run; unrelated to this diff (browser JS + error copy + tests).

Residual (acknowledged in the PR): native right-click "open in new tab" is uninterceptable from JS — the (now clearer) recovery page handles it.

[AI-assisted - Claude]

PR #4848, opened 2026-07-17
test: add daily cross-peer PUT/GET retention test (#4665)

### LINKED ISSUE #4665: Automated daily network test: PUT contracts via one peer, GET them via another
Add an automated daily test that exercises the core end-to-end invariant: one peer PUTs several contracts, and a DIFFERENT peer elsewhere in the network GETs them, verifying they are retrievable. This catches findability/retention regressions (e.g. a change that makes freshly-PUT contracts unfindable from a far origin, or that drops them under churn) that unit and in-process sim tests can miss.

Prompted by the demand-driven hosting work (epic freenet/freenet-core#4642, which touches findability/retention), but the test is generally valuable and NOT specific to that epic.

Scope:
- PUT N contracts via peer A, including at least one ~1 MB payload (the "Delta site" case).
- GET each via peer B after a short delay, and ideally again after a longer gap to exercise retention over time.
- Assert retrievability; alert on failure.
- Run daily (scheduled CI workflow or cron).

Note: existing code that partially overlaps (so nobody rebuilds it):

- **`crates/core/tests/operations.rs::test_put_contract`** — the closest existing "PUT via one peer, GET via another" test. In-process 2-node network (gateway + peer-a); PUTs a small todo-list contract via peer-a, then GETs it back via *both* peer-a and the gateway (a different node). Gaps vs. this ask: in-process only (no real transport/network), small payload (no ~1 MB case), no time-delayed retention check, and it runs in PR CI rather than on a daily schedule.
- **`crates/core/tests/test_network_integration.rs`, `manual_network_test.rs`, `large_network.rs`** — use the `freenet-test-network` crate to stand up real multi-node networks. `large_network.rs` is a manual (`#[ignore]`) ~40-peer soak test driving `riverctl` workflows. None is a targeted "PUT N contracts (incl. 1 MB) via A, GET via B after a delay, scheduled daily" test.
- **`.github/workflows/simulation-nightly.yml`** — the only existing *daily scheduled* (3 AM UTC) network test. It runs large-scale simulation (500 nodes / 250-contract scale, topology/churn/fault-tolerance focus), not a targeted cross-peer PUT-A/GET-B retrievability test with large payloads over time.
- **freenet-testing-tools: `scripts/river-test.sh` + `multi-machine-test/`** — manual River two-user chat tests across two nodes (a sent message is effectively a PUT via node 1, read via node 2). River-specific and manual, not a generic contract PUT/GET daily test.

No existing test covers the full ask (generic contracts, ~1 MB payload, cross-peer GET, retention-over-time, scheduled daily). The `operations.rs` in-process pattern and the `freenet-test-network` crate are the natural building blocks to extend.

[AI-assisted - Claude]

netsirius:
## Problem

No automated test exercises the core end-to-end invariant: contracts PUT via one peer must be retrievable by a different peer, including large payloads, and must remain retrievable over time. Findability/retention regressions in this class (#4223, #4345, #3851) were historically caught only via production telemetry, weeks after merge. With the demand-driven hosting epic (#4642) touching exactly this machinery, a daily regression signal is needed before its riskier changes ship.

A naive 2-node version of this test (PUT via peer-a, GET via the gateway) would prove nothing: every PUT hop stores locally, so the gateway always holds a replica and serves the GET from its local store without creating any network transaction. A fully broken network-GET path would still pass.

## Solution

New nightly-gated integration test `crates/core/tests/daily_cross_peer_put_get.rs` (`nightly_tests` feature, `#[freenet_test]` macro):

- peer-a PUTs 3 small contracts plus one ~1.1 MB payload.
- A third node (peer-b) is then booted **after** the PUTs complete, built manually inside the test with the same public APIs the macro uses (`ConfigArgs` + `InlineGwConfig` derived from the gateway's `NodeInfo`). Joining late means it holds no replica, so its GETs are forced to route over the network — the event log confirms one network `Get(Request)` per contract.
- Phase 1 GETs every contract via peer-b and asserts the returned contract and state are byte-identical to what was PUT.
- Phase 2 repeats the GETs after a 540 s gap, crossing the 480 s `SUBSCRIPTION_LEASE_DURATION` boundary. Nothing subscribes and no client touches the contracts during the gap, so subscription leases lapse and the demand-driven collapse/cleanup machinery runs for real — phase 2 then asserts that this cleanup did not drop or orphan the stored state (the central risk of epic #4642).
- GETs use a single attempt with a 120 s deadline (no retry loop: a GET that only succeeds on retry is itself the regression this test exists to surface, matching the nightly nextest profile's `retries=0`), match responses by contract key, and report failures via `ensure!`/`bail!` so the macro's `aggregate_events` failure report fires.

Workflow: the test runs as its own job (`daily-cross-peer-put-get`) in `simulation-nightly.yml` (daily 3 AM UTC cron) so the signal fires even when the unrelated simulation steps fail. The job installs the `wasm32-unknown-unknown` target (the test compiles `test-contract-integration` at runtime). Both River failure-notification jobs now cover this job as well.

## Fixes

Closes #4665

### DIFF
--- a/.github/workflows/simulation-nightly.yml
+++ b/.github/workflows/simulation-nightly.yml
@@ -108,22 +108,6 @@ jobs:
             --no-capture \
             --profile nightly
 
-      # Daily cross-peer PUT/GET retention test (issue #4665).
-      # PUTs N contracts (including a ~1 MB payload) via peer-a, then GETs
-      # each via the gateway (a different node) after a short delay and again
-      # after a longer gap to exercise retention over time.
-      # Gated by nightly_tests feature — does NOT run in regular CI.
-      - name: Daily cross-peer PUT/GET retention test
-        timeout-minutes: 30
-        run: |
-          echo "Running cross-peer PUT/GET retention test (issue #4665)"
-          cargo nextest run -p freenet \
-            --features "testing,nightly_tests" \
-            --test daily_cross_peer_put_get \
-            --test-threads 1 \
-            --no-capture \
-            --profile nightly
-
       # Large scale test (500 nodes) - uses direct runner (no turmoil overhead)
       # With simulation-mode timer optimizations (~5x fewer timer firings),
       # completes in ~4 min locally for 10000 events.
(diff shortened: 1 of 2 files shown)


sanity (commented):
This is a necessary idea but what if we ran it on the production network as a cron job on our server rather than as part of CI running in GitHub? We could even have it generate a static page on freenet.org graphing results over time, almost like Freenet weather. I think it would probably make sense to get and put multiple contracts rather than just one, and track retrieval success and retrieval time. 

Depending on how far we want to go, we could even track things like update latency. Since we're running the same experiment every night, we'll get a sense of network performance over time so we can see improvements or degradations.

netsirius:
@sanity Agreed, probing the production network nightly gives much more valuable signal than a synthetic CI network.

Here's how I'd approach it:

1. Nightly cron on the server: PUT N contracts (small plus ~1 MB) through one entry point, GET them through a different one, verify byte-identity, and re-GET previous nights' contracts for real 24h/48h/7d retention.
2. Metrics per run: PUT/GET success, retrieval time per contract size, integrity result, retention per window. Stored via the existing telemetry pipeline (tagged as probe events, with longer retention) and shown as a new tab in the existing dashboard, reusing its time-series machinery and release markers. As for the public page you suggested on freenet.org, I see two ways to do it: generate a static export every night and publish it through the site's existing deploy pipeline, or simply link from freenet.org to the new dashboard tab. Whichever you prefer, you know that infrastructure better than I do.
3. River alert when success drops below a threshold. Update latency as a later phase.

How do you think we should set up the nodes the probe uses? Should it talk directly to the existing gateways, or should we deploy dedicated client nodes for it, and if so, where would you want them running?

If this matches what you had in mind, I'll close this PR and open a new one implementing the probe, keeping #4665 as the tracking issue.

sanity:
Probably better to use existing gateway, I can give you ssh access to nova if you don't already have it.

netsirius:
Closing in favour of the production-probe approach agreed in this thread. 

A synthetic CI test can only exercise a locally-spun network, so it can't catch what we actually care about: a contract PUT into the *production* network that later can't be found. Per @sanity suggestion to use an existing gateway, the replacement runs nightly against the live network from nova, PUT through one entry point, GET through a different one via a  freshly-joined ephemeral peer, plus re-GETs of previous nights' contracts for real 24h/48h/7d retention. 

A new PR will follow. Tracking issue: #4665

Issue #3570, opened 2026-03-16
fix: GET operations have 73% network-wide timeout rate due to silent message drops and missing retry logic


sanity:
## Problem

Multiple users reported 30-second timeouts when loading a contract (`J4bh5TcQXMiPBiY5aAC7cEMh9f43M2cKx4L869GsG3Tw`, an experimentally published user contract) via `http://localhost:7509/v1/contract/web/...`. Three diagnostic reports were submitted (NBJZ9S, 9QSVK9, FYVVYE).

## User-Facing Root Cause: ForwardingAck Permanently Disables Retry

**This is the direct cause of user-visible browser timeouts.**

When a user's node originates a GET, the first relay peer sends a `ForwardingAck` within 3 seconds, confirming it's forwarding the request deeper. Upon receiving this ACK, the originator **permanently disables speculative retry** (`op_state_manager.rs:1533`). If the chain then stalls at any deeper hop — downstream peer doesn't respond, is slow, or can't find the contract — the originator has no recovery. It waits the full 60s `OPERATION_TTL` and times out.

### Evidence from diagnostic reports

Two GETs from the same user (NBJZ9S/Lukas) demonstrate this perfectly:

| GET | What happened | ACK received? | Speculative retry? | Result |
|-----|-------------|---------------|-------------------|--------|
| 1st (tx `01KKVTJ3C6`) | Relay ACKed, chain stalled deeper | Yes | **Disabled** | 63s timeout, no recovery |
| 2nd (tx `01KKVVYHH4`) | Relay didn't ACK (dead/slow peer) | No | **Fired after 6s** | Tried alternative peer |

The irony: the GET where the relay was *more responsive* (sent an ACK) had a *worse* outcome, because the ACK locked the originator into waiting for the full 60s with no fallback.

### Why the chain stalls

The contract exists on the network (29 successful GETs in telemetry from peers near ring location 0.98-0.99). But the reporting users' nodes are far from the contract on the ring (locations 0.56-0.70), requiring many hops (up to 9-10) to reach holders. More hops = higher probability that one intermediate peer goes silent, and with retry disabled, one silent peer burns the entire 60s timeout.

### Code location

```rust
// op_state_manager.rs:1533
if !get_op.ack_received {
    // ... speculative retry logic (fires after 3s ACK_TIMEOUT)
}
// If ack_received == true: skip entirely, wait full OPERATION_TTL (60s)
```

[...]

PR #4852, opened 2026-07-18
test: activate NN lattice + client-visible outcomes in GET-reliability test

sanity:
## Problem

`test_get_reliability_diagnostic` (nightly-only) failed on CI with a **16.5% NotFound** rate. Root-caused from the failing run's logs: the NotFounds are routing **dead-ends** — a GET routes greedily toward the key, reaches a peer with **no strictly-closer neighbor** ("all peers exhausted" — 333 such relay events), and returns NotFound even though the contract exists (retrievability held at 98/100). Dead-ends are peers not yet connected to their ring-closest neighbors. **Not a router regression** (0% locally, retrievability 98–100 every run).

Two compounding problems:

1. **The nearest-neighbor ring lattice (#4760) was gated off.** It's the production mechanism that makes peers seek/hold their closest neighbors — exactly what eliminates these dead-ends — but it activates only at `max_connections >= 25` (`NN_LATTICE_MIN_MAX_CONNECTIONS`), and this test runs at `max_connections=12`. That floor is a **sim-baseline-preservation gate, not a routing requirement** (it keys on *configured* max, which is 200 in production, so real peers are always lattice-on): the dead-ends are a nearest-neighbor problem, fixed by the lattice's base-lattice edges directly.

2. **The metric double-counted.** `summarize_get_outcomes_per_tx` counts per wire-transaction, so it tallied *relay-hop* dead-ends — a client GET that ultimately succeeds via another path still registered a NotFound. The user-visible number is lower.

## Approach

1. **Force-activate the lattice** at max=12 via the test-only `set_nn_lattice_enabled(true)` override + `NnLatticeTestGuard` (mirrors `nn_run_topology`, which already forces it on at max=5; same `run_controlled_simulation` runner; guard at function scope so it outlives the sim). The test now exercises the real production connectivity mechanism instead of a sparse, lattice-disabled regime that doesn't exist in production.
2. **Gate on client-visible outcomes** via a new `summarize_client_get_outcomes` (reads the `ClientTerminal` events #4796 already emits — one outcome per client GET, network vs local split, sub-ops excluded). The honest user-facing measure. *(The client-visible-outcome summarizer was developed in a parallel session and folded in here.)*

## Result (measured on CI, lattice ON)

```
per-tx:  73 total — 73 success, 0 not_found, 0 failures, 0 timeouts
client:  100 total — 100 success (73 network, 27 local), 0 not_found, 0 timeout_exhausted — 100.0%
storage: 100/100 nodes have the contract
```

**not_found: 16.5% → 0%.** The lattice eliminated the dead-ends entirely.

## Gates (tightened, justified by the structural fix)

- **client NotFound ≤ 10%** — measured 0/100; catches a wholesale dead-end regression (the original 16.5% trips it) with margin above the ~5–9% near-miss floor so it doesn't flake on a fixed-seed run.
- **client hard-failure (`timeout_exhausted`) ≤ 5%** — strict; retrievability can't see this class.
- **retrievability (`nodes_with_state`) ≥ 90** — primary, non-gameable backstop.
- **client network-traversal ≥ 30** — measured 73; broad routing actually happened.

Per-tx rates are kept computed + logged as diagnostics (not asserted).

## Testing

Test-only change. `tracing.rs` unit tests 11/11 pass; compiles clean under `--features nightly_tests`; verified on CI (numbers above).

Refs #3570, #4361/#4362, #4760, #4796.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/tracing.rs
+++ b/crates/core/src/tracing.rs
@@ -472,6 +472,108 @@ pub fn summarize_get_outcomes_per_tx(logs: &[NetLogMessage]) -> GetOutcomeSummar
     summary
 }
 
+/// Client-visible GET outcome tallies — one entry per client GET
+/// OPERATION, taken from the authoritative [`GetEvent::ClientTerminal`]
+/// event the GET driver emits exactly once per client op (and the
+/// local-cache-hit path emits via `emit_local_get_terminal_event`).
+///
+/// This is the client-SLA measure that [`GetOutcomeSummary`] explicitly is
+/// NOT (see its caveat): a GET that ultimately succeeds after a failed
+/// attempt counts as ONE success here, not a success plus a not_found, and a
+/// local-cache hit that never touched the network still counts as one
+/// success. Sub-op GETs (repair / renewal / related-fetch, tagged
+/// `is_sub_op`) are excluded so the tallies reflect real client demand only.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct ClientGetOutcomeSummary {
+    /// Client GETs that delivered state to the client (network OR local hit).
+    pub successes: u64,
+    /// Client GETs that terminated with a definitive NotFound from a
+    /// reachable peer — the transient near-miss class.
+    pub not_found: u64,
+    /// Client GETs whose retry budget burned on timeouts / unroutable
+    /// attempts (or a post-terminal streaming-assembly exhaustion, #4345)
+    /// WITHOUT a definitive NotFound — the client-visible hard-failure
+    /// class. Retrievability cannot see these.
+    pub timeout_exhausted: u64,
+    /// Subset of `successes` that actually traversed the network, split by
+    /// the terminal's forward `hop_count` (`hop_count >= 1` = the GET routed
+    /// to at least one other peer). This matches the per-tx metric's
+    /// `network-traversed` definition. NOTE: the split is deliberately by
+    /// `hop_count`, NOT `attempts` — a loopback `LocalCompletion` bumps
+    /// `requests_sent` (attempts >= 1) with no network round-trip, so
+    /// classifying by `attempts` over-counts those loopback completions as
+    /// network successes (#4852 P2).
+    pub network_successes: u64,
+    /// Subset of `successes` served locally with no forward network hop
+    /// (`hop_count` None or 0) — includes loopback local completions.
+    pub local_successes: u64,
+}
+
+impl ClientGetOutcomeSummary {
+    /// Total client GET operations observed (one per client op).
+    pub fn total(&self) -> u64 {
+        self.successes + self.not_found + self.timeout_exhausted
+    }
+}
+
+/// Summarize CLIENT-VISIBLE GET outcomes from an event log — one entry per
+/// client GET operation, from [`GetEvent::ClientTerminal`]. See
+/// [`ClientGetOutcomeSummary`] for why this differs from the wire-attempt
+/// [`summarize_get_outcomes_per_tx`], which double-counts a failed attempt of
+/// an ultimately-successful GET and misses local-cache hits.
+///
+/// Sub-op GETs are excluded. Deduplicated per tx defensively: the driver
+/// emits `ClientTerminal` once per op and each local-hit event carries its
+/// own unique tx, so per-tx dedup is a safe no-op that also guards against
+/// any accidental double-registration.
+pub fn summarize_client_get_outcomes(logs: &[NetLogMessage]) -> ClientGetOutcomeSummary {
+    use std::collections::HashMap;
+
+    // tx -> (outcome, hop_count). Once-per-op by construction; the map dedups
+    // any accidental double-registration to the last-seen terminal. We carry
+    // the terminal's forward `hop_count` (not `attempts`) because that is what
+    // distinguishes a real network traversal from a loopback local completion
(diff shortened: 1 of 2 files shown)


sanity:
Re-review of the **revised** change (split failure classes, replacing the first attempt's aggregate 0.75 floor):

**Codex (external re-review): clean** — _"The test-only change consistently separates hard failures from transient NotFound outcomes while retaining independent bounds and the existing retrievability and network-traversal checks. No actionable correctness issue was identified."_ `cargo fmt --check` passed.

This directly addresses the prior [P2] ("keep hard failures out of the NotFound allowance"): hard failures (GetFailure + ≥55s timeout) now have their own strict ≤5% gate, so they can no longer hide inside a loosened aggregate; transient NotFound is bounded separately at ≤15%; retrievability (≥90) stays the primary gate.

A skeptical adversarial re-review is also running; will fold in its findings. Merge gated on the on-branch nightly confirming the test passes with the new gates.

[AI-assisted - Claude]

sanity:
**Skeptical adversarial re-review: sound de-flake, no blocking findings (no P0/P1).** It verified the split is a genuine fix, not a relabel: with total ≈70, the hard-failure gate trips at **4/70** — strictly tighter than even the original 0.90 aggregate (which let up to 7/70 hard failures hide, and the rejected 0.75 up to 17); the outcome classes are mutually exclusive/exhaustive so a per-GET defect has no fifth hiding place; counters are `u64`, no div-by-zero (total ≥10 asserted above), arg lists match; test-only + nightly-only; no lingering false sibling-test claim.

The four P2/P3 residuals are all addressed in the pushed update (comment-only, no logic change):

- **[P2] 15% ceiling rests on one CI data point** — comment no longer oversells it as calibrated; it now states plainly that 15% is a coarse "findability collapse" detector resting on the single 8/70 observation (a tighter 12–13% would re-flake), and that **retrievability is the real backstop**, not this ceiling.
- **Q1: was retrievability actually measured at 98–100 on the 8-NotFound run?** No — the CI job log is truncated (GitHub capped it at the first ~22 min of a 109-min run), so retrievability on that run was never separately captured. Comment now says so explicitly: measured 100/100 on the 0-NotFound local runs; retrievability-stays-high for the elevated-NotFound case is an **inference from the caching mechanism, not a measurement**.
- **[P3] a ≥55s near-miss is reclassified into the strict class** — added as an explicit caveat (the one elevated-NotFound CI run had 0 timeouts, and a genuine 55s give-up is worth failing on).
- **[P3] ~20% aggregate tolerance vs the old ~10%** — now owned in the comment ("tolerates up to ~14/70 non-success only if predominantly benign NotFound AND caching keeps retrievability ≥90").
- **[P3 cosmetic] hard-failure assert said "floor"** → corrected to "ceiling".

Also folded in the mutual-exclusivity note. Codex re-review was clean on the prior push; a bonus second external pass is running on this content. Merge still gated on the on-branch nightly confirming the test passes with the new gates.

[AI-assisted - Claude]

sanity:
## Review summary (combined lattice + client-metric fix)

**Result confirmed on CI:** with the nearest-neighbor lattice force-activated, GET not_found went **16.5% → 0%** (per-tx 0/73, client-visible 0/100, retrievability 100/100). Root cause was routing dead-ends from peers lacking their ring-closest neighbors — the lattice fixes it directly.

### Reviews run
- **Skeptical adversarial (Claude): SOUND, no P0/P1/P2.** Independently confirmed (a) force-activating the lattice below its ≥25 floor is legitimate — the floor is a sim-baseline gate, and the existing benefit tests already force it on at max=5; (b) the thread-local override genuinely reaches all sim peers (same mechanism as `THREAD_RNG`/`SIMULATION_TIME_MS`), so the 0% is real, not an artifact.
- **External (Codex): found two real gate-precision gaps, both fixed:**
  - **P1** — client GETs failing via `InfraError`/`Unexpected` emit no `ClientTerminal`, so were invisible to gates dividing by `client_total`. Fixed: rates now computed over the **scheduled total** (`NUM_NODES`), with terminal-less GETs counted as hard failures.
  - **P2** — a driver `LocalCompletion` bumps `requests_sent`, so `attempts≥1` miscounted loopback as network traversal. Fixed: network/local split now uses forward `hop_count` (matches the sibling per-tx metric).

### External-model substitution (audit trail, per review policy)
Codex was available for the first pass (found P1/P2) but hit its **usage limit** before I could re-run it to confirm the fixes (resets **Jul 25**). Gemini CLI is deprecated (ineligible-tier). With both external models unavailable and the change not able to wait days, I used the **Claude-subagent fallback** for the P1/P2-closure re-review, with an adversarial lens focused on the two specific gaps. Verdict: **both fixes correctly close their gaps, no new functional bug** — it proved `client_total ≤ scheduled` (so the missing-terminal accounting can't hide failures) and that real network successes keep `hop_count ≥ 1` (the only `hop_count=None` network path is streaming, which can't occur here: the test contract is ~112 bytes, far below the 64 KB threshold). Two P3 docstring nits it found are fixed in `8daad7fb`.

Change is **test-only** (a new `summarize_client_get_outcomes` in `tracing.rs` reading existing #4796 `ClientTerminal` events, plus the test). 12/12 tracing unit tests pass; both files clippy-clean.

Merge-gated on the on-branch verification nightly confirming the tightened gates pass (running).

[AI-assisted - Claude]

PR #4856, opened 2026-07-18
fix(server): persist notification flood-cap across page reloads (#4849)

### LINKED ISSUE #4849: Notification proxy (#4801) hardening: flood-cap reset via reload + guard-coverage + test
Follow-up to #4801 (browser-notification proxy for sandboxed contract apps). At merge, the primary item below (flood-cap reset) was a **conscious accept** — the core security model was verified sound and the weakness is bounded/visible — with the fix deferred to this issue so the release wasn't held. Tracking all post-#4801 notification-proxy hardening here rather than fragmenting into several issues.

Review basis: a full multi-model re-review of #4801 at HEAD `f602288d` (external **Codex** pass + two adversarial Claude lenses). All three confirmed the trust boundary holds — consent fails closed, cross-contract isolation is intact, no injection surface (the `icon` field isn't proxied), `postMessage` source is checked. The items below are the residual findings.

---

### 1. (Primary, ~P2 — defense-in-depth) Notification flood-cap is resettable via an app-triggered reload

The per-tab flood-cap (global 20 notifications / 60s, in `makeNotifyRateLimiter()`) lives in a per-document `var notifyLimiter` in the shell-bridge IIFE (`crates/core/src/server/path_handlers/assets/shell_bridge.js`). A contract the user has **already consented to** can:

1. Send 20 unique-tag notifications (hits the global cap), then
2. `postMessage` a `{type:'navigate', href:<its own v1↔v2 URL>}`. Because `CONTRACT_PREFIX_RE` includes the version, a same-key v1→v2 hop is treated as cross-contract → `window.location.assign` → full top-level reload → **fresh, empty limiter**. The consent key is version-less, so consent survives the reload.

Alternating v1/v2 yields repeated fresh 20-notification budgets — a ~10–30× weakening of the anti-flood cap.

**Why it was accepted rather than blocking:** only exploitable by an app the user already allowed; each cycle costs a visible 1–3s tab reload-thrash (self-limiting and obvious); tags coalesce rather than stack; and it breaks **no** security boundary — it's a robustness gap in a brand-new control, not a regression of existing behavior.

**Suggested fix:** make the rolling-window flood accounting survive reloads — e.g. persist the recent-notification timestamps in contract-scoped `sessionStorage`/`localStorage` (keyed like consent/snooze, so it can't leak across contracts) and rehydrate on load; or fold the contract version into the reload-reset accounting so a same-key v1↔v2 hop doesn't grant a fresh budget.

### 2. (P3) #3836 guard re-scoping covers less of the permission flow than its comment claims

Independently flagged by **both** Claude lenses and Codex. The #3836 `new Notification(` / `requestPermission` asserts were narrowed from a whole-HTML scan to an `overlay_slice` scan (`path_handlers.rs:2677-2705`). The slice ends at `setInterval(reconcileFromPending` (~JS line 1411), so the `EventSource` / `prompt_added` SSE handlers (~1419+), which are part of the delegate permission-prompt render flow, sit **outside** the scanned region.

Production is safe today — the only `new Notification(`/`requestPermission` live in the notification proxy, far from the overlay/SSE region, which is Notification-free. But a future regression that put a browser `Notification` into the SSE prompt handlers would keep the test green.

**Suggested fix:** bracket the full permission-overlay + SSE implementation with explicit `BEGIN/END` markers (as the rate-limiter block already does) and scan that whole region.

### 3. (P3) Bounded-map eviction unit test doesn't exercise eviction

`shell_bridge_notifications.test.mjs` (case 5) spaces calls `i*61000` apart, so the 60s window empties before every call and all 300 pass regardless of the `MAP_CAP` eviction block — deleting the eviction entirely would keep the test green. The eviction logic itself is correct, just unverified.

**Suggested fix:** assert the map size directly, or reuse an evicted tag inside its throttle window via the injected clock.

---

### Minor / accepted-as-inherent (no action expected)

- **Snooze-key non-canonical spelling:** a contract reloading to a percent-encoded variant of its own path gets a different snooze key → the "Enable notifications?" bar can re-show after "Not now". Fails **closed** for consent (re-spelled path has no consent → no notifications); net effect is a re-nag annoyance, not a breach.
- **Title spoofing:** a consented contract can title a notification "River security alert"; the OS shows only the shared gateway origin. Inherent to the Web Notifications API and already scoped by per-contract consent. Optional future hardening: prepend a shell-generated contract label and treat the app title as untrusted secondary text.

[AI-assisted - Claude]

sanity:
## Problem

The notification-proxy flood-cap (20 notifications / 60s, in `shell_bridge.js`'s `makeNotifyRateLimiter`) lives in a per-document `var`, so a **full page reload resets it**. A contract the user has *already consented to* can fire its 20 notifications, then `postMessage` a same-contract v1↔v2 `navigate` — which the shell treats as cross-contract and reloads the top document — to get a fresh, empty limiter, and repeat for an unbounded flood budget. Surfaced as **F1** in the #4801 multi-model re-review (external Codex pass, ~P2). It's bounded and visible (only an already-consented app; each cycle is a visible tab reload) and breaks no security boundary, so it was consciously accepted at #4801's merge and deferred to this issue.

This PR also closes the two P3 hardening items from the same review:
- **F2** — the #3836 permission-overlay guard (`shell_page_permission_overlay_present_and_safe`) scanned a slice that ended *before* the SSE `prompt_added`/`prompt_removed` handlers, which are part of the prompt-render flow #3836 protects — so a browser `Notification` reintroduced into an SSE handler would have slipped past the guard.
- **F3** — the bounded-map eviction unit test spaced its calls so the 60s global window emptied every call; deleting the `MAP_CAP` eviction block entirely would have kept it green — it never actually exercised eviction.

## Approach

**F1 — persist the rolling window (contract-scoped, reload-surviving).** `makeNotifyRateLimiter` now takes an optional injected `{load, save}` store: it rehydrates the rolling global window on construction and persists it after each accepted notification. A new `makeNotifyRateStore()` backs it with `sessionStorage`, keyed by the **same version-less contract key as consent** (`freenet_notify:<key>:rate`) — so the window survives the exact v1↔v2 reload the exploit relies on, while staying isolated per contract. `sessionStorage` is the right store: per-tab, same-origin (persists across the shell's in-place navigate), and auto-cleared on tab close. The store is deliberately kept **outside** the factory so the factory stays pure and the Node unit test can still extract it verbatim between its markers; passing no store preserves the prior in-memory behavior. Stale entries in a rehydrated window are dropped by the existing 60s filter on the first `ok`, so the cap correctly frees up after 60s across reloads (no over-blocking).

**F2 — bracket the whole overlay+SSE region with explicit markers.** New `perm-overlay-flow:BEGIN/END` markers bound the region; the Rust guard now scans between them instead of stopping at a code anchor, so the SSE handlers are inside the guarded region. (The legitimate message-notification code sits above `BEGIN`, so it stays outside the region — the #3836 asserts don't false-positive on it.)

**F3 — assert boundedness.** The limiter exposes a `tagCount()` accessor (unused in production) so the eviction test asserts the per-tag map stays ≤ `MAP_CAP` after a 300-tag flood; removing the eviction block now fails the test.

## Testing

- `shell_bridge_notifications.test.mjs` (Node, run by the `lint-assets` CI job), **25 checks**: case 6 fills the cap, simulates a reload via a fresh limiter rehydrating from the same store, and asserts the next notification is still blocked (a storeless control proves the persistence is the closer); a >60s-stale rehydration frees capacity; case 5 asserts `tagCount() <= MAP_CAP`; **case 7 drives the real `makeNotifyRateStore` adapter** (marker-extracted with a stubbed `sessionStorage`) through its guards — non-array / corrupt-JSON → null, non-finite + future-timestamp filtering, storage-throw fail-safe, no-contract-key → null.
- Rust: a new `bridge_js_notification_flood_cap_persisted_across_reload` pins the F1 wiring (store construction, contract-scoped `:rate` key, sessionStorage get/set, load/save); `shell_page_permission_overlay_present_and_safe` (F2) now also asserts the SSE `prompt_added`/`prompt_removed` handlers are *inside* the marker-bounded slice, so the guard can't shrink to a vacuous pass.

## Review hardening (from this PR's multi-model review)

- **bfcache resync:** a `pageshow` handler resyncs the persisted window on back-forward-cache restore (the constructor doesn't re-run), closing a reload-shaped reset the persistence otherwise wouldn't cover — without relying on the shell being bfcache-ineligible.
- **Backward-clock clamp:** `load` drops future-dated timestamps, so a backward wall-clock correction can't leave the cap stuck (bounded to 60s; fails safe either way).
- **F2 self-validating:** positive `prompt_added`/`prompt_removed`-in-slice asserts, so a refactor moving the SSE handlers past `perm-overlay-flow:END` fails loudly instead of silently un-covering them.
- **Adapter behavioral coverage:** `makeNotifyRateStore` is now marker-extracted and behaviorally tested (was source-grep-only).

Closes #4849

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -2683,16 +2683,31 @@ mod tests {
             html.contains("function setText(el, text)"),
             "setText helper (textContent-only) missing"
         );
-        // innerHTML must not appear anywhere in the overlay code path. The
-        // boundary marker below was previously `setInterval(checkPermissions`
-        // but that polling loop is gone; we now bound the slice at the
-        // EventSource-fallback `setInterval(reconcileFromPending`.
-        let overlay_start = html.find("__freenet_perm_overlay").unwrap();
+        // Bound the overlay code path by the explicit `perm-overlay-flow`
+        // markers in shell_bridge.js, NOT a code anchor. The previous bound
+        // (`setInterval(reconcileFromPending` / `EventSource`) stopped SHORT of
+        // the SSE `prompt_added`/`prompt_removed` handlers, which ARE part of
+        // the prompt-render flow #3836 protects — so a browser Notification
+        // reintroduced into an SSE handler would have slipped past this guard.
+        // The markers bracket the whole overlay + SSE region so the asserts
+        // below scan all of it (#4849 F2).
+        let overlay_start = html
+            .find("perm-overlay-flow:BEGIN")
+            .expect("perm-overlay-flow:BEGIN marker must bracket the overlay flow");
         let overlay_end = html[overlay_start..]
-            .find("setInterval(reconcileFromPending")
-            .or_else(|| html[overlay_start..].find("EventSource"))
-            .expect("overlay slice must end at the SSE setup or fallback poll");
+            .find("perm-overlay-flow:END")
+            .expect("perm-overlay-flow:END marker must bracket the overlay flow");
         let overlay_slice = &html[overlay_start..overlay_start + overlay_end];
+        // The negative asserts below are only meaningful if the slice actually
+        // CONTAINS the SSE prompt-render surface. Pin that the marker-bounded
+        // region includes the `prompt_added`/`prompt_removed` handlers, so a
+        // refactor that moves them past `perm-overlay-flow:END` (shrinking the
+        // slice) fails HERE rather than silently making the negative asserts
+        // pass vacuously — the exact regression F2 exists to prevent (#4849).
+        assert!(
+            overlay_slice.contains("'prompt_added'") && overlay_slice.contains("'prompt_removed'"),
+            "overlay guard slice must cover the SSE prompt handlers (#4849 F2)"
+        );
         assert!(
             !overlay_slice.contains("innerHTML"),
             "overlay code path must not use innerHTML (XSS surface)"
@@ -3376,6 +3391,52 @@ mod tests {
         assert!(
             SHELL_BRIDGE_JS.contains("Notification.requestPermission(done)"),
             "permission prompt must be requested from the shell affordance click"
+        );
+    }
+
+    /// Regression for #4849: the notification-proxy flood-cap (the rolling
+    /// global window in `makeNotifyRateLimiter`) must be PERSISTED per-contract
+    /// so a full page reload can't reset it. Without this, a consented contract
+    /// could fire the whole budget, force a reload (a same-contract v1<->v2
+    /// `navigate`, which the shell reloads as cross-contract), and start over
+    /// with an empty limiter. The behavioral proof is in
+    /// shell_bridge_notifications.test.mjs (the reload rehydration case); this
+    /// pins the WIRING at the source level so a refactor can't silently drop
+    /// the persistence and re-open the reload-reset hole.
+    #[test]
+    fn bridge_js_notification_flood_cap_persisted_across_reload() {
+        // The limiter is constructed WITH the persistence store, not the old
+        // no-arg makeNotifyRateLimiter().
+        assert!(
+            SHELL_BRIDGE_JS.contains("makeNotifyRateLimiter(makeNotifyRateStore())"),
+            "rate limiter must be constructed with the persistence store (#4849)"
+        );
+        // The store is keyed off the version-less contract consent key (so the
+        // window survives a v1<->v2 reload) with a `:rate` suffix, and is backed
+        // by sessionStorage (per-tab, same-origin, reload-surviving).
+        assert!(
+            SHELL_BRIDGE_JS.contains("ckey + ':rate'"),
+            "rate window must use a contract-scoped storage key (#4849)"
+        );
+        assert!(
+            SHELL_BRIDGE_JS.contains("sessionStorage.getItem(storeKey)")
+                && SHELL_BRIDGE_JS.contains("sessionStorage.setItem(storeKey"),
+            "rate window must be persisted in sessionStorage (#4849)"
+        );
+        // The factory actually rehydrates from and saves to the injected store.
+        assert!(
+            SHELL_BRIDGE_JS.contains("store.load()")
+                && SHELL_BRIDGE_JS.contains("store.save(recent)"),
+            "limiter must rehydrate from and persist to the injected store (#4849)"
+        );
+        // The bfcache reset variant is closed by a `pageshow`-persisted resync
+        // (the IIFE does not re-run on back-forward-cache restore, so the
+        // in-memory window would otherwise stay stale). Pin the wiring so a
+        // refactor can't silently drop it.
+        assert!(
+            SHELL_BRIDGE_JS.contains("pageshow")
+                && SHELL_BRIDGE_JS.contains("notifyLimiter.resync()"),
+            "flood-cap window must be resynced from the store on bfcache restore (#4849)"
         );
     }
 
--- a/crates/core/src/server/path_handlers/assets/shell_bridge.js
+++ b/crates/core/src/server/path_handlers/assets/shell_bridge.js
@@ -13,8 +13,18 @@ function freenetBridge(authToken, userToken, hostedMode) {
   var inMemoryConsent = Object.create(null);
   // Per-tag + global rate limiter for the notification proxy (see the
   // marker-bracketed `makeNotifyRateLimiter` factory below, unit-tested in
-  // shell_bridge_notifications.test.mjs).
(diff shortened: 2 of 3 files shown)


sanity (commented):
## Comprehensive PR Review: #4856

### Summary
- **PR:** `fix(server): persist notification flood-cap across page reloads (#4849)`
- **Type:** fix (notification-proxy / permission-overlay hardening — a browser trust-boundary surface)
- **Review tier:** **Full** (security surface)
- **Reviewers:** code-first, testing, skeptical, big-picture (freenet reviewer subagents) + external **Codex** (`gpt-5.6-sol`, genuine non-Claude). Two full rounds — round 1 on `feecafa1`, re-review on `c651d442` after fixes; a final test-only delta landed at `520cf890`.
- **Verdict:** **Ready to Merge** — unanimous SAFE across all five reviewers including the external pass; every finding fixed or explicitly dismissed with justification.

### Code-First / Skeptical (correctness + security)
The fix closes a **real, self-service flood-cap bypass**: a consented contract could fire its 20/60s budget, `postMessage` a same-contract v1↔v2 `navigate` (version-ful `CONTRACT_PREFIX_RE` → full top-document reload) and, because consent is version-*less*, re-derive the same consent key with a fresh in-memory limiter → 20 more, looping. Persisting the rolling window in contract-scoped `sessionStorage` (`freenet_notify:<key>:rate`) closes it.

Independently verified by skeptical + Codex: the sandboxed contract is an **opaque-origin iframe** (`sandbox` has no `allow-same-origin`) so it cannot read/poison the shell's `sessionStorage` or forge another contract's key; `load` **fails safe** (corrupt/non-array/oversized → the limiter is only ever *more* restrictive → self-DoS only, never a bypass); the persisted array is bounded ≤20; `resync()` and the clock clamp only tighten and never loosen the cap. No new attack surface.

### Testing
`.mjs` unit suite (CI `lint-assets`/`npm test`) — **27 checks pass**, including a reload-reset regression (with a storeless control proving persistence is the closer), a non-vacuous full-cap >60s stale-window free-up, a behavioral `resync()` case, and case-7 driving the **real `makeNotifyRateStore` adapter** (extracted via markers with a stubbed `sessionStorage`) through every fail-safe guard. Rust: `bridge_js_notification_flood_cap_persisted_across_reload` (pins the F1 wiring incl. the `pageshow`/`resync` bfcache defense), `shell_page_permission_overlay_present_and_safe` (F2, now self-validating), `bridge_js_notification_proxy_invariants` — all pass.

### Findings — all resolved or dismissed

| # | Source | Severity | Resolution |
|---|--------|----------|------------|
| F1a bfcache restore bypass | Codex | P2→P3 | **Fixed** — `pageshow`/`persisted` resync (conservative: replaces only on non-empty load) |
| F1b backward-clock lockout | Codex | P2→P3 | **Fixed** — `load()` clamps `t <= Date.now()` (fails safe; not contract-triggerable) |
| GAP1 adapter behaviorally untested | testing | MED | **Fixed** — `makeNotifyRateStore` marker-extracted + driven in case 7 |
| GAP3 F2 guard could pass vacuously | testing | MED | **Fixed** — positive `prompt_added`/`prompt_removed`-in-slice asserts |
| GAP2 stale-rehydration test vacuous | testing (re-review) | LOW-MED | **Fixed** — full-cap (20-entry) stale fixture (mutation-confirmed non-vacuous) |
| resync path untested | testing / code-first / Codex | MED / LOW / P3 | **Fixed** — behavioral `resync()` case + Rust `pageshow`/`resync()` source-pin |
| PR-body mis-described F1 test | big-picture | LOW | **Fixed** — names the new dedicated test |
| GAP4 eviction-*order* test | testing | LOW | **Dismissed** — order correct by inspection; a dedicated test is disproportionate to a cosmetic-throttle concern |
| multi-tab per-tab budget | big-picture | note | **Dismissed** — pre-existing (old in-memory limiter was per-tab too); opening real tabs is far more visible than a programmatic reload; #4849's target (in-tab reload-reset) is closed |
| markers stripped by minification | code-first | note | **Dismissed** — no minification in the asset pipeline; guards fail *closed* (loud test failure), not silent bypass |

### Documentation
Code comments explain the WHY at each site (consent-key isolation, `sessionStorage` vs `localStorage` divergence, bfcache rationale). PR body corrected and now documents the review hardening. No user/architecture docs affected.

### Verdict
**Ready to Merge.** **HEAD reviewed:** `c651d442` (production code — the panel's clean pass) + `520cf890` (test-only additions closing the flagged coverage gaps; no production change). All required CI green on the current HEAD is the remaining gate.

[AI-assisted - Claude]

PR #4858, opened 2026-07-19
fix: resync on dropped broadcast delta to prevent permanent divergence (#4857)

### LINKED ISSUE #4857: State updates permanently lost for rarely-changing fields: silent ContractQueueFull drop + sender-side neighbor-summary poisoning
## Summary

State updates can be **permanently lost for rarely-changing contract fields**, because a silently-dropped broadcast delta poisons the sender's cached summary of the receiver, so the changed field is never re-sent. Only the ~5-min InterestSync heartbeat heals it — imperfectly. Found while debugging why River clients that are actively exchanging messages never receive a `member_info` change (deputy grants → shields), even though their room otherwise syncs.

## Mechanism (node↔node broadcast)

Per-neighbor broadcast deltas are computed as `diff(current_state, sender's cached summary of that neighbor)`:

1. On any "delivered" broadcast the sender updates its cache of the neighbor to **its own current summary** — `record_delivery_to_interest` (`crates/core/src/node/network_bridge/p2p_protoc/broadcast_queue.rs:543-544`). "Delivered" = `bridge.send(...)` returned `Ok` on the delta path (`:823-836`) / sender-side stream completion (`:531-541`), **NOT a receiver ack**.
2. If the receiver drops the delta with `ContractQueueFull`, the drop is **silent** — no `ResyncRequest` is sent (`crates/core/src/operations/update/op_ctx_task.rs:1237-1244`).
3. So after a dropped delta the sender believes the neighbor is current. Every subsequent broadcast delta = `diff(current, poisoned_summary)` = only newer changes → **the dropped one-shot change is never re-sent.**
4. New messages keep applying (each re-ships the latest `recent_messages` window, self-masking the bug), but a change to a rarely-updated field (member_info, configuration, a ban) stays missing until the ~5-min InterestSync heartbeat corrects the cache and pushes a full-state heal (`node.rs:2703`, `:2806`) — and that heal rides the same broadcast path, **re-poisons the cache on send**, and can be dropped again under load (heals are also capped/rotated per Summaries message, `node.rs:2732-2745`).

**Not field-specific.** Any rarely-changing field is exposed. `recent_messages` self-masks only because every new message re-ships the latest window — a silently-dropped message delta from that window would also be lost (a plausible source of occasional lost-message reports).

## Evidence

- The affected contract's `summarize`/`delta`/`apply_delta` were proven correct against real state (the delta carries the changed field; applying it heals a stale peer) — so the contract layer is not the cause.
- Subscribers registered with `summary: None` receive `UpdateData::State` (full current state) on every update (`crates/core/src/contract/executor/runtime/executor_impl.rs:2279`); a live scratch subscriber on a converged node received the change on the next organic update. So affected clients faithfully mirror a **stale node** — the divergence is node↔node, not node↔client.

## Proposed fix

Do not update the neighbor-summary cache from a **sender-side** delivery signal alone — update it only from a real summary exchange (Interests/Summaries, `ResyncResponse`) or a receiver ack. At minimum, make the receiver's `ContractQueueFull` drop emit a (rate-limited) NACK / `ResyncRequest` instead of dropping silently, so the divergence reliably self-heals.

## Secondary findings (related; may warrant their own issues)

1. **Summary staleness is a raw byte comparison of nondeterministic bytes.** `is_stale` byte-compares `summarize_state` outputs (`node.rs:2698-2701`), but a Summary containing `HashMap`s serializes in ciborium's nondeterministic order — two *identical* states can summarize to *different* bytes across nodes, so the equal-summary skip almost never fires and heartbeat heals fire spuriously. Consistent with a very high volume of rate-limited `summarize_contract_state` calls in a production node journal. Fix: canonicalize or compare semantically (contracts can also help by using deterministic maps in their Summary type).
2. **Over-disk-budget peers silently reject growth-only updates** with no error to the sender, and the follow-up `ResyncResponse` is even larger and also rejected (`executor_impl.rs:1696-1708`, `node.rs:3099`) — a second permanent-divergence shape.

Investigated 2026-07-18 via code trace + live read-only subscription tests. [AI-assisted - Claude]

sanity:
## Problem

State updates can be **permanently lost for rarely-changing contract fields** (member_info, a ban, a config value). When a receiver drops a broadcast with `ContractQueueFull`, the drop is **silent**: the SENDER already cached its own summary as the neighbor's on send-Ok (`broadcast_queue.rs::record_delivery_to_interest`, and the streaming sibling `record_streaming_delivery`), so it believes the neighbor is current and never re-sends the missed change. Every later broadcast is `diff(current, poisoned_summary)` = only newer changes, so the dropped field never reappears. Only the ~5-min InterestSync heartbeat can correct it (imperfectly — the heal rides the same poisonable path). New messages self-mask the bug because each re-ships the latest `recent_messages` window.

The tension this PR resolves: the queue-full arm **suppressed** the healing `ResyncRequest` entirely. Issue **#4251** (PR #4253) added that suppression because an *unbounded* ResyncRequest per dropped delta amplified into a full-state storm onto the same saturated queue (~40 ResyncRequests/sec on a hot contract). So emit → storm (#4251); suppress → diverge (#4857).

Both broadcast paths are affected: the non-streaming `drive_relay_broadcast_to` (delta) **and** the streaming full-state `drive_relay_broadcast_to_streaming` (reachable for large River rooms streaming >64KB full-state to a summary-less/just-cleared subscriber).

## Solution

Emit a **rate-limited** `ResyncRequest` on the queue-full drop instead of suppressing it, on **both** paths via one shared `send_queue_full_resync_request` helper. An inbound ResyncRequest makes the sender **clear** its cached summary of us (node.rs handler) and re-send full state.

- `InterestManager::should_send_resync_request(contract, target)` — bounded-LRU per-`(contract, peer)` throttle (fail-open on eviction; LRU because the key is influenced by remote peers, so it must be bounded per `.claude/rules/code-style.md`). Interval `RESYNC_REQUEST_MIN_INTERVAL = 30s`, chosen as ~10x under the 5-min heartbeat backstop while keeping River's ~80-target fan-out below the #4251 storm.
- Streaming steps 4-9 extracted into `apply_streaming_broadcast` so the streaming heal is unit-testable without the transport stream-assembly plumbing.
- Auto-fetch stays suppressed on queue-full (a GET onto the same saturated queue is pure amplification with no healing value).

**Why (a) not (b):** the issue proposed either (a) a receiver-side rate-limited NACK or (b) stop caching the neighbor summary on sender-side send-Ok. (b) standalone would re-introduce the **#4442** full-state broadcast storm — caching on delivered-send is exactly what lets a new subscriber's next broadcast collapse to a small delta. (a) is the minimal, safe fix.

### Honest bounds (the #4251↔#4857 tension is MANAGED, not eliminated)

The throttle is per-`(contract, receiver)`, so aggregate resync load scales **O(N)** with receiver count, and each `ResyncResponse` is **full state** (larger than the dropped delta). For River's ~80-target fan-out this is ~2-3 ResyncRequests/sec (~15x under the 40/sec that triggered #4251), but the bound is not universal. Under **sustained** saturation the heal round-trips the same full queue and defers until it drains, so "heals faster than the heartbeat" holds for the common **transient** case and after saturation clears — not during sustained overload. The durable half of the heal (the sender clearing its poisoned summary) is a small control-plane InterestMessage that does **not** ride the saturated contract queue, so it lands even while the queue is full; the full-state resend is what defers. Net: bounded amplification + faster common-case heal, with the heartbeat as the ultimate backstop.

## Testing

- **`queue_full_broadcast_emits_throttled_resync_request`** (behavioral, delta path) and **`queue_full_streaming_broadcast_emits_throttled_resync_request`** (behavioral, streaming path): each stands up a real `OpManager` with a contract-handler stand-in that replies `ContractQueueFull`, drives the real driver logic, and asserts **one** ResyncRequest on the first drop and **none** on the throttled second. Both verified to **FAIL** without their respective fix (`left: []`, no ResyncRequest) and pass with it.
- **`should_send_resync_request_rate_limits_per_contract_peer`** (unit): first allowed, immediate repeat throttled, distinct `(contract, peer)` buckets independent, re-allowed after the interval.
- **`broadcast_to_sends_throttled_resync_on_queue_full`** (tightened structural pin): BOTH drivers' queue-full arms delegate to the shared helper; the helper nests the `InterestMessage::ResyncRequest` emit INSIDE the `should_send_resync_request` gate (emit position between gate and the throttled `else`); neither arm auto-fetches. Verified it FAILS when the streaming arm is removed.

**Heal-loop coverage** is piecewise (a true end-to-end multi-node heal needs the sim harness with coarse fault injection): emit → existing behavioral tests above; ResyncRequest → sender clears its cached summary → existing `resync_request_handler_clears_cached_peer_summary`; cleared summary (None) → next broadcast sends full state → existing `broadcast_queue` summary-caching tests.

Regression check: update module (70), interest (40), broadcast_queue (8), node resync (1) all green; `cargo fmt` + `cargo clippy --lib -D warnings` clean.

## Scope / follow-ups

Fixes the **primary** divergence only. The two secondary findings in #4857 are left as follow-ups:
1. `is_stale` byte-compares nondeterministic (ciborium HashMap) summary bytes, so the equal-summary skip rarely fires and heartbeat heals fire spuriously.
2. Over-disk-budget peers silently reject growth-only updates (a second permanent-divergence shape).

Closes #4857

https://claude.ai/code/session_01B49wBfvR8EjpfYTw5muNV9

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -1235,13 +1235,61 @@ async fn drive_relay_broadcast_to(
                     AutoFetchReason::InboundRelay,
                 );
             } else if queue_full {
-                tracing::debug!(
-                    tx = %incoming_tx,
-                    contract = %key,
-                    sender = %sender_addr,
-                    event = "queue_full_amplification_suppressed",
-                    "UPDATE relay: per-contract queue saturated, suppressed ResyncRequest/auto-fetch to avoid amplification"
-                );
+                // Issue #4857: a queue-full drop is SILENT. The receiver never
+                // applied the delta, but the SENDER cached its summary as ours
+                // on send-Ok (broadcast_queue.rs::record_delivery_to_interest),
+                // so it believes we are current and will never re-send the
+                // dropped change. Left unhealed, a rarely-changing field
+                // (member_info, a ban, a config) diverges permanently until the
+                // ~5-min InterestSync heartbeat happens to correct it.
+                //
+                // Emit a ResyncRequest so the sender CLEARS its cached summary
+                // of us (node.rs ResyncRequest handler) and re-sends full state.
+                // Issue #4251 suppressed this ENTIRELY because one request per
+                // dropped delta amplifies into a full-state storm onto the same
+                // saturated queue (~40 ResyncRequests/sec on a hot contract).
+                // The rate limit bounds that to at most one request per
+                // (contract, sender) per RESYNC_REQUEST_MIN_INTERVAL, which
+                // heals the divergence without re-opening the storm. The auto-
+                // fetch branch stays suppressed on queue-full (we already hold
+                // the contract; a GET onto the same saturated queue is pure
+                // amplification with no healing value).
+                //
+                // do NOT re-add an unthrottled emission here — see #4251/#4857.
+                if op_manager
+                    .interest_manager
+                    .should_send_resync_request(&key, sender_addr)
+                {
+                    tracing::info!(
+                        tx = %incoming_tx,
+                        contract = %key,
+                        target = %sender_addr,
+                        event = "resync_request_sent",
+                        reason = "queue_full",
+                        "UPDATE relay: sending rate-limited ResyncRequest after queue-full drop (#4857)"
+                    );
+                    if let Err(e) = op_manager
+                        .notify_node_event(NodeEvent::SendInterestMessage {
+                            target: sender_addr,
+                            message: InterestMessage::ResyncRequest { key },
+                        })
+                        .await
+                    {
+                        tracing::warn!(
+                            tx = %incoming_tx,
+                            error = %e,
+                            "UPDATE relay: failed to send queue-full ResyncRequest"
+                        );
+                    }
+                } else {
+                    tracing::debug!(
+                        tx = %incoming_tx,
+                        contract = %key,
+                        sender = %sender_addr,
+                        event = "queue_full_resync_throttled",
+                        "UPDATE relay: queue-full ResyncRequest throttled (rate limit) to avoid amplification"
+                    );
+                }
             }
             return Err(err);
         }
@@ -2583,15 +2631,16 @@ mod tests {
         );
     }
 
-    /// Issue #4251 regression pin: `drive_relay_broadcast_to` MUST
-    /// gate ResyncRequest emission AND `try_auto_fetch_contract` on
-    /// `!err.is_contract_queue_full()`. The two amplification branches
-    /// are what turn a single saturated contract into a network-wide
-    /// storm — ResyncRequest asks the sender for full state (bigger
-    /// payload onto the same full queue), and auto-fetch enqueues a
-    /// GET right back onto the saturated handler.
+    /// Issue #4857 / #4251 reconciliation pin: on a queue-full broadcast
+    /// drop, `drive_relay_broadcast_to` must emit a RATE-LIMITED
+    /// `ResyncRequest` — neither suppress it (which left rarely-changing
+    /// fields permanently diverged, #4857) nor emit it unthrottled (which
+    /// amplified into a full-state storm, #4251). The queue-full arm MUST
+    /// therefore BOTH gate on `should_send_resync_request` (the rate limit)
+    /// AND emit `InterestMessage::ResyncRequest` (the heal). Auto-fetch stays
+    /// suppressed on queue-full (a GET onto the saturated queue heals nothing).
     #[test]
-    fn broadcast_to_suppresses_amplification_on_queue_full() {
+    fn broadcast_to_sends_throttled_resync_on_queue_full() {
         let src = include_str!("op_ctx_task.rs");
         let start = src
             .find("async fn drive_relay_broadcast_to(")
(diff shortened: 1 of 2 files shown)


sanity:
## Review summary (Full-tier: Claude multi-lens + external Codex)

Reviewed across correctness / concurrency / edge-cases / regression / test-coverage plus an external Codex pass. **Verdict: mergeable.** The fix breaks the summary-poisoning cycle (verified end-to-end: send-Ok caching → queue-full drop → rate-limited ResyncRequest → sender clears its cached summary at `node.rs:2952-2958` → next broadcast is full state via the #4442 None→full-state path → missed field reappears), covers BOTH the non-streaming and streaming drivers, and reconciles the #4251↔#4857 tension without re-introducing #4251's actual trigger (single hot contract) thanks to the per-`(contract,peer)` 30s throttle. Both new behavioral tests were confirmed to FAIL on main and pass with the fix.

### Codex findings — disposition (all non-blocking)
- **HIGH — fail-open LRU bounds memory not amplification; O(N)-with-receiver-count aggregate; key-thrash could defeat the 30s throttle.** Accepted as a documented bound. River-scale (~80 keys) is far under the 4096-entry cache, so no thrash; defeating the throttle needs >4096 distinct `(contract,peer)` drops per 30s (~136/s) — a fresh-key-volume condition, NOT a re-emission of #4251's single-hot-contract trigger (which the throttle *does* fix). The PR body already discloses the O(N) aggregate. Non-blocking follow-up: add jitter / a global ceiling if aggregate resync ever shows in telemetry.
- **MEDIUM — throttle stamped before send; lossy `bridge.try_send` (broadcast.rs:103) can drop a 'successfully notified' ResyncRequest.** Accurate but pre-existing (the existing delta-apply-failure ResyncRequest uses the same lossy path) and strictly better than main (which sends nothing); heartbeat-recoverable. Stamping before send is the safer choice for storm-avoidance.
- **MEDIUM — recovery not durable under sustained saturation (ResyncResponse re-enters the same queue; node.rs:3083 has no retry).** Explicitly acknowledged in the PR's 'honest bounds': the durable half (cache-clear) lands, full-state resend defers until drain, converges via next organic broadcast or heartbeat. The node.rs:3083 arm is pre-existing/untouched.
- **LOW — partial-stream tail loss outside the heal.** Pre-existing, orthogonal to queue-full, heartbeat-backstopped.

### Non-blocking follow-ups
End-to-end two-manager convergence test (the full drop→field-reappears loop is currently covered piecewise), and align the `interest.rs` 'one extra request' comment with the PR body's honest O(N) aggregate disclosure.

[AI-assisted - Claude]

PR #4862, opened 2026-07-19
fix: resync request survives bridge backpressure (throttle-on-dispatch follow-up to #4858)

sanity:
## Problem

Follow-up to #4858 (the queue-full resync fix for #4857, merged via `b2a6fb15`). That PR emits a rate-limited `ResyncRequest` when a broadcast delta is dropped with `ContractQueueFull`, so a poisoned sender summary is cleared before the ~5-min anti-entropy heartbeat — closing #4857 for the common case.

The residual P2 the round-2 review flagged: **the resync request can itself be dropped by the bridge under the same backpressure that caused the queue-full drop.** `send_queue_full_resync_request` only queues a `SendInterestMessage`; the event loop then forwards it via `P2pBridge::try_send` (`broadcast.rs`), which is lossy (`#4145` made it non-blocking). Meanwhile the 30s throttle slot (`should_send_resync_request` / `RESYNC_REQUEST_MIN_INTERVAL`) is stamped at **decision** time, not on actual dispatch. So if the resync is dropped:

- the throttle slot is burned for 30s, and
- for a **rarely-changing field** (member_info, a ban, a config value) there is no further broadcast to re-trigger the resync,

so the divergence persists until the 5-min heartbeat anyway — the exact hole #4858 was meant to close.

## Approach

After the immediate `ResyncRequest`, schedule a **bounded, jittered, trailing re-dispatch** (`resend_queue_full_resync_request`) so the resync gets more chances to land once the transient saturation clears — well before the heartbeat, and **independent of any further broadcast** (the rarely-changing field never re-broadcasts to re-trigger).

**#4251 storm bound preserved (hard constraint), now rigorous:**
- `should_send_resync_request` returns `Some(deadline)` when it grants a reservation (`deadline = stamp + RESYNC_REQUEST_MIN_INTERVAL`, on the throttle's clock) and `None` when throttled. The retry burst is **anchored to that deadline**: it both paces and stops on the throttle's own clock, so a burst can never cross into a new reservation — even if the blocking `notify_node_event(...).await` consumed most of the window before the retry task was spawned (review **P2-A**).
- Both the inter-attempt delay and the deadline stop read `op_manager.interest_manager.now()` — the **same injected clock** the throttle uses (`hosting_time_source_override` in sims, the wall clock in prod) — not tokio's independent clock, so sims that advance the injected clock are deterministic (review **P2-B**). The util `TimeSource` exposes only `now()` (no sleep), so the wait polls it in bounded tokio increments.
- Steady-state is capped at `1 + QUEUE_FULL_RESYNC_MAX_RETRIES` (= 3) `ResyncRequest`s **per `(contract, peer)` per 30s window** — ~600x below the #4251 storm (~40/s on a hot contract). The decision-time reservation still blocks concurrent burst drops.
- Retries use the **non-blocking** `try_notify_node_event` (bounded-channel rule; blocking would pin the task up to the 30s `NOTIFICATION_SEND_TIMEOUT`).
- Jittered linear backoff (±20% via `GlobalRng`, deterministic under sim) de-synchronises retry waves when one spike drops broadcasts to many `(contract, peer)` pairs at once.

Change is inside the shared `send_queue_full_resync_request` helper, so **both the delta and streaming queue-full paths** are covered.

Why not "record the throttle only on successful dispatch" (option 1): the bridge `try_send` is two hops downstream (event loop → `broadcast.rs`), so recording there would thread the throttle into the high-risk broadcast dispatch path, and — more importantly — not stamping at decision time would weaken the burst cap (concurrent drops would all pass the un-stamped gate). Keeping the decision-time reservation and confining the deadline-anchored retry inside it preserves the storm bound while still healing a bridge-dropped resync.

## Honest cost / bounds

The retry is **blind**: it re-dispatches regardless of whether the first `ResyncRequest` already landed. Each `ResyncRequest` makes the sender clear its cached summary and re-send **full contract state** (a `ResyncResponse`), so a single drop event can cost **up to `QUEUE_FULL_RESYNC_MAX_RETRIES` (2) redundant full-state `ResyncResponse`s** (each ~a full contract state) landing on the receiver whose queue was full. The only cap is **per `(contract, peer)` per 30s** — there is **no global (per-node) ceiling**, so aggregate heal traffic is O(number of distinct dropped `(contract, peer)` pairs) per window. This is bounded and far below the #4251 storm, and only triggers after a real drop, but it is a real cost. Eliminating the redundant full-state resends (conditional retry, gated on an observed bridge-drop signal) and adding a global cap are tracked as a follow-up in **#4863**.

## Testing

- `queue_full_resync_retry_redispatches_within_reservation` (behavioral, `start_paused` virtual time): one queue-full drop, **no further broadcast** → the trailing retry re-emits exactly `MAX` additional `ResyncRequest`s to the original sender, and no more.
- `queue_full_resync_retry_stops_after_reservation_deadline` (behavioral, P2-A): advancing the clock past the 30s deadline mid-burst stops all further retries.
- `queue_full_resync_retry_is_bounded_and_reservation_scoped` (source-scrape pin): the retry is spawned only **inside** the throttle gate, creates **no** new reservation, is bounded by the const, uses `try_notify_node_event`, paces + gates on `interest_manager.now()`, threads the reservation deadline, and never reads a fresh `Instant::now()` (guards the storm bound + P2-A/P2-B).
- `resync_retry_span_stays_within_throttle_window`, `jittered_resync_retry_delay_stays_within_bounds`.
- `should_send_resync_request_rate_limits_per_contract_peer` (interest.rs) updated for `Option<Instant>`; asserts `deadline == now + interval`.
- Existing `broadcast_to_sends_throttled_resync_on_queue_full` pin and both queue-full behavioral tests still pass (awareness notes added re: the spawned retry task).

`cargo fmt` clean, `cargo clippy -p freenet` clean, targeted unit + behavioral tests green against current main (84 passed in the two touched modules).

Follow-up to #4858. Refs #4857. Aggregate-fan-out follow-up: #4863.

> Draft: concurrency + protocol surface (Full-tier). Needs a full multi-model re-review and Ian's sign-off before ready/merge.

[AI-assisted - Claude]

---

## Known limitation (accepted — tracked in #4863)

**Prolonged-backpressure double-drop falls back to the heartbeat.** If the notification channel stays saturated longer than the ~7.2 s retry burst, the retry attempts can be consumed while the immediate `notify_node_event` is still blocked (up to the 30 s `NOTIFICATION_SEND_TIMEOUT`). If that immediate send then resolves and is dropped by the saturated `P2pBridge`, no trailing retry remains and that specific divergence falls back to the ~5-minute InterestSync heartbeat.

This is **not a regression** — that prolonged double-drop case waited for the heartbeat before this PR too. It is fundamentally not closable by the blind-retry approach: bounding retries to the reservation window (to avoid overlapping the next reservation) contradicts keeping one alive past a fully-blocked 30 s send. The proper fix is the **conditional / observe-then-retry** in **#4863** (retry when an actual bridge drop is observed, no timing guess). This PR's benefit is the common **transient**-backpressure case, which now heals in seconds instead of ~5 minutes.

**Honest fan-out bound.** The per-node 256-slot cap bounds the aggregate **retry-induced** full-state `ResyncResponse` fan-out. The **base** per-(contract,peer) immediate `ResyncResponse` fan-out is O(N) receivers during a broad drop storm — this is **unchanged pre-#4862 behavior** (it rode in with the already-merged #4858) and is also covered by #4863.

### DIFF
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -12,18 +12,22 @@
 
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
+
+use tokio::time::Instant;
 
 use either::Either;
 use freenet_stdlib::client_api::{ContractResponse, ErrorKind, HostResponse};
 use freenet_stdlib::prelude::*;
 
 use crate::client_events::HostResult;
-use crate::config::GlobalExecutor;
+use crate::config::{GlobalExecutor, GlobalRng};
 use crate::contract::{ContractHandlerEvent, StoreResponse};
 use crate::message::{DeltaOrFullState, InterestMessage, NetMessage, NodeEvent, Transaction};
 use crate::node::OpManager;
 use crate::operations::OpError;
 use crate::operations::bootstrap::bootstrap_gateway_target;
+use crate::ring::interest::RESYNC_REQUEST_MIN_INTERVAL;
 use crate::ring::{PeerKeyLocation, RingError};
 use crate::tracing::{NetEventLog, OperationFailure, state_hash_full};
 
@@ -1133,18 +1137,17 @@ async fn send_queue_full_resync_request(
     sender_addr: SocketAddr,
     incoming_tx: Transaction,
 ) {
-    // RESERVE the per-(contract, sender)/30s throttle window atomically (#4864
-    // round-6 item 2). `begin` checks AND records under the throttle's own lock,
-    // so two concurrent queue-full callbacks for the same (contract, sender)
-    // cannot both pass this gate and go on to both consume the global burst +
-    // emit duplicates inside the window. If the global cap then rejects, we
-    // `cancel` the reservation below (releasing the window) — preserving the
-    // round-5 improvement that a globally-suppressed emit does not burn the 30s
-    // window.
-    if !op_manager
+    // Gate 1 — per-(contract, sender)/30s throttle (#4857/#4862 + #4864 round-6).
+    // RESERVE the window atomically (`begin` checks AND records under the
+    // throttle's own lock) and return the reservation deadline on the throttle's
+    // OWN clock (the #4862 retry anchors to it). Two concurrent queue-full
+    // callbacks for the same (contract, sender) cannot both pass. If the global
+    // cap below then rejects, we `cancel` the reservation so the window is
+    // released (not burned).
+    let Some(reservation_deadline) = op_manager
         .interest_manager
         .begin_resync_request(&key, sender_addr)
-    {
+    else {
         // Per-(contract, sender)/30s throttle (existing #4857 bound).
         tracing::debug!(
             tx = %incoming_tx,
@@ -1154,7 +1157,7 @@ async fn send_queue_full_resync_request(
             "UPDATE relay: queue-full ResyncRequest throttled (rate limit) to avoid amplification"
         );
         return;
-    }
+    };
 
     // ALSO subject to the GLOBAL per-contract emit cap (#4864 round-4 P1): the
     // #4857 queue-full heal path previously bypassed it, so a saturated contract
@@ -1193,13 +1196,56 @@ async fn send_queue_full_resync_request(
         reason = "queue_full",
         "UPDATE relay: sending rate-limited ResyncRequest after queue-full drop (#4857)"
     );
-    // #4864 round-8: record the outstanding request so the matching ResyncResponse
+    // #4862 (P2): the immediate ResyncRequest below rides a LOSSY path
+    // (`notify_node_event` → event loop → `P2pBridge::try_send`), which can drop
+    // it under the SAME backpressure that caused the queue-full drop. If dropped,
+    // the sender never clears its poisoned summary and a one-off change diverges
+    // until the ~5-min heartbeat. Schedule a BOUNDED trailing re-dispatch so the
+    // resync gets more chances to land — well before the heartbeat.
+    //
+    // Spawn CONCURRENTLY, BEFORE the blocking immediate enqueue below (#4862 P2):
+    // `notify_node_event(...).await` can consume up to NOTIFICATION_SEND_TIMEOUT
+    // (30s) under backpressure; spawning after it would let the task start with
+    // the window already gone (`now >= reservation_deadline`) and do nothing. The
+    // first attempt is a jittered delay out (never t=0), so it does not duplicate
+    // the immediate send, and it stays anchored to `reservation_deadline` (the
+    // throttle's own clock) so it can NEVER spill into a new reservation. The
+    // retries re-deliver the ONE already-authorized emit, so they do NOT
+    // re-consult the per-sender throttle or the global emit cap and do NOT
+    // re-record the outstanding request (a redundant response for an
+    // already-consumed record is correctly dropped by #4864 before WASM).
+    //
+    // #4862 P1: gate the spawn on a node-wide retry-task slot so the throttle LRU
+    // evicting active reservations under key churn cannot spawn unbounded
+    // overlapping tasks. At cap ONLY the retry is skipped; the immediate send
+    // below still fires. Fire-and-forget; the slot guard frees on completion.
+    if let Some(slot) = op_manager.interest_manager.try_reserve_resync_retry_slot() {
+        let op_mgr = op_manager.clone();
+        GlobalExecutor::spawn(async move {
+            let _slot = slot; // frees the node-wide retry slot on task exit
+            resend_queue_full_resync_request(
+                &op_mgr,
+                key,
+                sender_addr,
+                incoming_tx,
+                reservation_deadline,
+            )
+            .await;
+        });
+    }
+
+    // #4864 round-8: record the outstanding request (immediately before the emit
+    // so the round-8 pin's proximity check holds) so the matching ResyncResponse
     // from this peer is authorized ONCE at the receive arm; an unsolicited or
-    // replayed response finds no entry and is dropped without running WASM.
+    // replayed response finds no entry and is dropped without running WASM. The
+    // #4862 trailing retry above re-sends the SAME request and relies on THIS one
+    // record — see `resend_queue_full_resync_request`.
     op_manager
         .ring
         .outstanding_resync_requests
         .record(*key.id(), sender_addr);
+
+    // Immediate (first) ResyncRequest — blocking best-effort enqueue.
     if let Err(e) = op_manager
         .notify_node_event(NodeEvent::SendInterestMessage {
             target: sender_addr,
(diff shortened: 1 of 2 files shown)


sanity:
Addressed both P2 findings from the Full-tier review in `6c9743b2` (re-review needed on the new content):

- **P2-A** — `should_send_resync_request` now returns `Some(deadline)` (`stamp + RESYNC_REQUEST_MIN_INTERVAL`, on the throttle's clock) / `None`. The retry is anchored to that deadline and stops the instant the clock reaches it, so a burst can never overlap a fresh reservation even if the blocking `notify_node_event(...).await` consumed most of the window. New test `queue_full_resync_retry_stops_after_reservation_deadline` pins it.
- **P2-B** — both the inter-attempt delay and the deadline stop read `interest_manager.now()` (the injected `hosting_time_source_override` clock the throttle uses), not tokio's clock, via a bounded `now()`-poll (the util `TimeSource` has no sleep primitive). The pin test asserts the retry paces + gates on `interest_manager.now()`, threads the deadline, and never reads a fresh `Instant::now()`.

Also: corrected the PR body's honest-cost section (up to 2 redundant full-state `ResyncResponse`s per drop; per-`(contract, peer)` cap only, no global cap), and filed **#4863** for a global cap and/or conditional (observed-drop-gated) retry.

`cargo fmt` / `cargo clippy -p freenet` clean; 84 tests pass in the two touched modules (`operations::update::op_ctx_task`, `ring::interest`). Kept as **draft** — not merging pending the focused re-check + sign-off.

[AI-assisted - Claude]

sanity:
Addressed the DST-robustness finding from the delta re-review in `9d0260b6`:

The retry poll-wait gated termination only on the injected throttle clock; under a DECOUPLED `hosting_time_source_override` that FREEZES before the deadline, the injected gates never fire while `start_paused` auto-advances tokio time → the poll would spin (hangs a start_paused test; leaks a task per drop on a real sim). Fix: a tokio-clock **liveness floor** — `let started = tokio::time::Instant::now()` before the loop, `return` once `started.elapsed() >= RESYNC_REQUEST_MIN_INTERVAL` (now `pub(crate)` so the floor is sized to exactly one reservation window). The injected `reservation_deadline`/`target` checks are unchanged and remain **primary**; in prod the two clocks coincide and the injected deadline (stamped strictly before `started`) fires first, so the floor is a no-op.

New coverage test `queue_full_resync_retry_terminates_when_injected_clock_frozen`: injects a `SharedMockTimeSource`, freezes it below the deadline, advances only tokio, and asserts the task terminates within a bounded budget (a spin fails the `timeout` instead of hanging) and dispatches no retry. The storm-bound pin now also asserts the backstop exists.

`cargo fmt` / `cargo clippy -p freenet` clean; 85 tests pass in `operations::update::op_ctx_task` + `ring::interest`. Still **draft/not merged** pending the external Codex pass + final check.

[AI-assisted - Claude]

sanity:
Addressed both findings from the external Codex pass in `b146b5cf`:

**P2 — preserve a retry when the initial enqueue consumes the window.** The retry task was spawned *after* the blocking `notify_node_event(...).await` (up to 30s `NOTIFICATION_SEND_TIMEOUT` under backpressure), so it could start with `now >= reservation_deadline` and do nothing. Fix: (a) spawn the retry **concurrently, before** the blocking immediate enqueue so its timer runs during the send (the first attempt is a jittered delay out, never t=0, so no duplicate immediate send); (b) **clamp** the first target into the remaining window (`delay = jittered.min(remaining/2)`), so a nearly-consumed window still fires ≥1 retry within `[now, reservation_deadline)` instead of overshooting the deadline. No-op in the common full-window case.

**P1 — bound retry tasks independently of the evicting throttle LRU.** Under saturation + key churn beyond `RESYNC_THROTTLE_CACHE_SIZE`, the LRU evicts active reservations → each revisited key re-grants and spawns another retry task, unbounded. Fix: a **node-wide cap** (`MAX_OUTSTANDING_QUEUE_FULL_RESYNC_RETRIES = 256`, via `InterestManager::try_reserve_resync_retry_slot` returning an RAII `ResyncRetrySlot` that decrements on drop; hard CAS cap). At cap the spawn is skipped; the immediate `ResyncRequest` still sends.

**Cost correction / MEDIUM subsumed.** The node-wide cap bounds the aggregate retry-task count and thus the aggregate retry-induced full-state `ResyncResponse` fan-out per node — so the earlier MEDIUM (uncapped aggregate fan-out under a broad drop storm) is **subsumed** for the retry path. Follow-up **#4863** re-scoped to just the optional conditional-retry refinement (skip a redundant full-state resend when the first `ResyncRequest` already landed) — not a correctness gap.

New tests: `queue_full_resync_retry_slot_cap_bounds_outstanding_tasks` (P1 hard cap), `queue_full_resync_retry_fires_within_a_short_remaining_window` (P2 clamp); storm-bound pin extended (spawn gated on the slot, slot moved into the task, spawn precedes the blocking immediate send). All prior pins + 87 tests green in `operations::update::op_ctx_task` + `ring::interest`. `cargo fmt` / `cargo clippy -p freenet` clean.

Still not merged — ready for the focused re-check.

[AI-assisted - Claude]

sanity:
Rebased onto `origin/main` and integrated with **#4864** (poison-contract containment), which had merged and rewritten the exact `send_queue_full_resync_request` helper + the `interest.rs` throttle. Force-pushed the reconciled branch (now one commit, `9fa329f6`). **Unreviewed conflict resolution — needs re-review before re-queue.**

### Conflicting merged PRs
- **#4864** (`e2b0471f`) — rewrote `send_queue_full_resync_request`, renamed `should_send_resync_request` → `begin_resync_request` (+ `cancel_resync_request`), added a global per-contract emit cap (`resync_emit_limiter`) and `outstanding_resync_requests` response authorization, plus merge-failure backoff gates. **This is the real overlap.**
- #4852 (`5feb1a2a`) — touched op_ctx_task.rs elsewhere (GET-reliability test); no conflict with my regions.

### How each conflict was resolved
1. **Throttle method** (`interest.rs`): `begin_resync_request` now returns `Option<Instant>` (the reservation deadline my retry anchors to) instead of `bool`, folding in #4862's deadline-return onto #4864's atomic reserve/cancel. #4864's `begin`/`cancel` tests updated to `.is_some()/.is_none()`; my deadline assertion kept.
2. **`send_queue_full_resync_request`**: kept #4864's two-gate structure (per-sender `begin` → global `resync_emit_limiter`, cancel-on-suppress) and its `outstanding_resync_requests.record`. Inserted #4862's retry spawn (P1 slot-gated, P2 concurrent-before-blocking-send) **after both gates pass**, and placed the record immediately before the emit so #4864's round-8 proximity pin holds.
3. **Retry semantics** (the key call): the retry RE-DELIVERS the ONE already-authorized emit — it does **not** re-consult `begin_resync_request`/`resync_emit_limiter` and does **not** re-record `outstanding_resync_requests`. Re-recording would re-authorize a redundant `ResyncResponse` and re-run its full-state WASM merge — the opposite of #4864's drop-the-redundant intent. So #4864's `every_production_resync_request_emit_records_outstanding` pin is given a **documented exemption for the retry re-delivery** (its response is covered by the initial record in the bridge-drop case, and correctly dropped once consumed). This is a genuine synergy: #4864's authorization now bounds my blind-retry's redundant-full-state cost to **zero extra WASM merges**.
4. **Test module**: #4864's backoff pins and my resync tests are both additive — kept both.

### Preserved (verified)
All #4862 mechanisms intact: queue-full ResyncRequest + 2× bounded jittered retry, per-node 256-slot `ResyncRetrySlot` cap, reservation-deadline anchor, injected-`TimeSource` pacing, tokio liveness backstop, node-wide-cap; all gated by `begin_resync_request` + the global cap. **108 tests pass** in `operations::update::op_ctx_task` + `ring::interest` (my tests + #4864's, including `queue_full_resync_flood_is_bounded_by_global_emit_cap` and the exempted round-8 pin). `cargo fmt` / `cargo clippy -p freenet` clean. Branch merges cleanly onto current `origin/main` (no new conflicts).

[AI-assisted - Claude]

Issue #4861, opened 2026-07-19
Periodic 100% CPU on peers: non-converging contract update churn (5s WASM-timeout delta loop + Invalid-update ResyncRequest storm + module-cache thrash)


sanity:
## Summary

A user reported periodic 100% CPU on their 2-vCPU VM running v0.2.101. Reproduced and profiled on nova (gateway, v0.2.101, and the try.freenet.org hosted peer, v0.2.100): **the spikes are contract-WASM execution churn, not a v0.2.101 regression**. Three interacting mechanisms, all network-wide:

1. **Poison-delta 5s WASM timeout loop.** Contract `Gsd2311PU4PFrFvptNToGjxswUKbMeouZF6sfjV3CVgQ`: every delta apply runs until the 5s max-compute limit, is killed, fails with `The operation exceeded the maximum allowed compute time`, triggers a `ResyncRequest`, and then the next delta does the same. Each occurrence pegs one core for a full 5 seconds. Rate: ~100–220/hour on the gateway, sustained for at least 36h (all of Jul 18 + 19, both 0.2.100 and 0.2.101).

2. **"Invalid update" → full-state resync storm.** Contract `HQk7zGRQagL9AVUMM39C3W3nRVv81xk9BCAeY9MikirK` (plus `xqyyifd4…`, `FBFRDjxV…`, `Eumk9HNQ…`, others): delta apply fails with `execution error: Invalid update` from **many different senders**, each failure sends a `ResyncRequest`, the full state (~150KB–768KB) is re-applied through WASM (validate + update + summarize), and the next delta fails again. Gateway: **1,211 ResyncRequests in 35 minutes** (577 for `HQk7…` alone ≈ 16/min). The hosted peer shows the same contracts at the same rates (445 for `HQk7…` in ~37 min). This never converges. With `update_propagation_summary` showing `targets_avg=78` for `HQk7…`, every peer holding these contracts across the network is doing the same failed-delta→resync dance.

3. **Module-cache thrash.** 242 distinct contracts recompiled ("Module cache miss — compiling") in 35 minutes on a gateway hosting 2,652 contracts. Each wasmtime compile is CPU-expensive. On a small (2-core / low-RAM) box the byte-budget cache is smaller, so thrash is worse.

## Measurements (nova, 16 cores)

- 20-min per-second sampling: gateway **avg 106% CPU, max 523%, 40% of seconds >100%**; hosted peer avg 57%, max 244%.
[...]

PR #4864, opened 2026-07-19
fix(update): contain poison-contract CPU churn — enforced WASM preemption, merge-failure backoff, resync rate limits (#4861)

sanity:
## Problem

Production peers burn CPU indefinitely on "poison" contracts — the cause of the user-reported periodic 100% CPU on small VMs (#4861). Forensics on the nova gateway identified three poison classes running at full rate for 36h+:

1. **Compute poison**: a contract whose every merge exceeds the 5s execution budget (100-210 timeouts/hr, ≈14-29% of a core, continuously, on every host).
2. **Semantic fork oscillation**: two stable divergent states rejecting each other's deltas; every full-state resync "succeeds" but merely flips the node to the other fork (~1 cycle/min, 45 distinct senders, 9,733 ResyncResponses sent for one contract in a day).
3. **Deserialization poison**: a malformed delta circulating forever.

Core amplified all three: any delta-apply failure emitted a ResyncRequest with no failure counter, no backoff, no rate limit, and no give-up — and the ResyncResponse full-state apply re-ran WASM. Worse, with `enable_metering` off (the default), the 5s "timeout" only abandoned the result: `abort()` cannot stop synchronous WASM, so the guest kept burning its blocking thread arbitrarily long, and a hostile infinite-loop contract would burn a thread permanently.

## Solution

Five mechanisms, all behavioral — **zero wire-format changes**:

1. **Epoch-based preemption** (`wasmtime_engine.rs`): `epoch_interruption(true)` + a global 100ms ticker + per-call `set_epoch_deadline` with trap-on-deadline. The execution budget is now actually enforced — a runaway guest traps within one tick of the deadline instead of running forever. `Trap::Interrupt` is classified as the execution-timeout error. Regression test runs an infinite-loop WAT with no wall-clock wrapper. Note: `epoch_interruption` is part of wasmtime's compile-cache key, so every cached WASM module recompiles once on the first load after upgrade (a one-time cost).
2. **Typed timeout classification** (`contract/executor.rs`): `is_wasm_timeout()` so drivers can distinguish compute poison from semantic rejection.
3. **Per-contract merge-failure backoff** (`ring/merge_backoff.rs`): consecutive failures put a contract into exponential cooldown (Invalid: trips only after 3 consecutive failures with no intervening success, then 30s→30min; Timeout: trips at the first failure, 120s→2h; ±20% jitter — one benign stale-version reject must never blackout a healthy contract, while each timeout is a full 5s CPU burn), during which deltas are dropped *before* WASM executes and no ResyncRequest is sent. Includes a 10-min failed-payload memo (a failed (state, delta) pair cannot change outcome until the state changes). **Reset is strictly delta-only**: only a clean delta apply proves the receiver shared the sender's base. Full-state merges — streaming broadcast and resync apply alike — never reset, because in the fork-oscillation class they "succeed" by flipping forks (source-scrape pins enforce this). **Suppression scope (round-3 review, anti-blackout):** Invalid-class failures are tracked per-(contract, sender), so one bad peer only suppresses its own deltas and cannot blackout a healthy contract for every honest peer; Timeout-class failures stay contract-wide (a ~5s CPU burn is contract-intrinsic). The failed-payload memo is contract-wide, content-addressed, and a hard skip for any sender (deterministic execution means the same bad bytes fail identically for everyone; the skip is zero-cost — it advances no sender channel and emits no ResyncRequest). A remote delta success clears that sender's Invalid channel plus the contract-wide Timeout and memo; a client-local delta success clears only the contract-wide Timeout and memo, leaving remote senders' Invalid channels intact. **State-advance gating (round-4):** the per-sender Invalid channel clears on any clean delta apply (channel-trust), but the contract-wide Timeout + memo clear only when the merge advanced state (`changed == true`) — a no-op (UpdateNoChange) delta must not wipe a live Timeout quarantine, else a busy contract carrying a co-circulating timeout payload clears the 120s→2h quarantine on every routine delta. A state-advancing full-state apply (streaming/resync, changed) invalidates only the failed-payload memo (`invalidate_payload_memo`), never a cooldown, so the strictly-delta-only no-reset doctrine still holds.
4. **Resync rate limiting** (`ring/resync_rate_limit.rs`): emitter-side per-contract token bucket (burst 2/60s); responder-side per-(peer,contract) bucket (mixed-version guard against un-upgraded storming peers) plus a **global per-contract cap** (~12/min) — production showed 45 requester IPs, so per-peer limiting alone is insufficient. **Responder existence-gate (round-4, made backend-agnostic in round-5):** a cheap state-presence probe (`contract_state_present_async` — a synchronous redb point-lookup fast-path, or a real SQLite `EXISTS` query) runs before both limiter buckets, making the order exists → per-peer → global → state fetch, so a peer spraying bogus keys can never exhaust the strictly-capped limiter maps (which would fail-close legitimate contracts). The old sync-only probe was a no-op on SQLite builds; the async variant closes that gap on both backends (the SQLite branch is correct-but-dormant until the SQLite backend compiles again — #4869). The #4857 queue-full resync heal now also routes through the global per-contract emit cap (previously only per-(contract, sender) throttled), so a saturated contract with many senders emits at most the burst allowance total, not one per sender. Flood test: 45 peers × 5min → ~62 responses vs ~9,733/day today.
5. **Module-cache budget clamp raise** (1.5→4 GiB max; RAM/8 base unchanged, small boxes unaffected): the 2,652-contract gateway thrashes the 1.5 GiB cap at ~7 recompiles/min. **Upgrade migration:** an existing >12 GiB node whose config.toml carries the old auto-derived `module-cache-budget-bytes = 1610612736` (1.5 GiB) is re-derived to the new default instead of staying pinned to the stale value; CLI/env and any other persisted value are preserved. Accepted edge (release-notes): an operator who explicitly set exactly 1.5 GiB is indistinguishable from the old auto-default and will also be re-derived.

Containment philosophy: core bounds the cost of a broken/hostile contract; it never assumes the contract-layer bug heals. The InterestSync anti-entropy heartbeat remains the recovery path once a contract exits cooldown.

Hosting-invariants note: touches the UPDATE path only. Invariant 1 (served copies fresh): a contract in backoff has already lost convergence (every merge fails); the bounded stale window during cooldown falls under the invariant's permitted transients, and anti-entropy remains the healer. No placement/eviction/subscription semantics changed.

## Testing

- Infinite-loop WAT preemption regression test (no wall-clock wrapper — only the epoch trap can return it).
- Fork-oscillation regression: alternating delta-fail/resync-apply must still escalate into backoff (this exact pattern defeated a naive reset-on-success design).
- Flood tests for all three limiters (modeled on the May-21 UpdateRateLimiter flood test).
- Source-scrape pins: reset gated on `is_delta`; streaming path never resets; node.rs never resets; responder gates ordered exists → per-peer → global → state fetch (a 50-bogus-key flood consumes no limiter slot); queue-full resync heal bounded by the global per-contract emit cap; epoch deadline re-armed before every guest entry.
- MergeBackoff unit suite: per-sender/contract-wide channels, cooldown escalation, jitter bounds, memo TTL, cap, cleanup; DST-compliant (TimeSource + GlobalRng throughout).
- `cargo fmt` / CI-exact `clippy -D warnings` clean; targeted `nextest` 194/194 + doc-fix pins.

Root-cause forensics + measurements: #4861 (issue + comment).

Follow-ups (deliberately out of scope): SQLite state-store backend repair (#4869); per-contract CPU duty-cycle budget (the hard adversarial guarantee, PR 2); contract-layer fixes for the specific poison contracts (River room merge cost, fork reconciliation); dashboard wiring for the new suppression counters.

Part of #4861

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -177,7 +177,7 @@ pub struct ConfigArgs {
     /// until it fits. Bounding by bytes (not entry count) stops a node hosting
     /// many contracts from thrashing the cache and recompiling on every access
     /// (issue #4441). When unset, the default scales with system RAM
-    /// (`clamp(total_ram / 8, 64 MiB, 1.5 GiB)`); set this to override.
+    /// (`clamp(total_ram / 8, 64 MiB, 4 GiB)`); set this to override.
     #[arg(long, env = "FREENET_MODULE_CACHE_BUDGET_BYTES")]
     pub module_cache_budget_bytes: Option<usize>,
 
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3421,6 +3421,9 @@ std::thread_local! {
     // `StateDelta` ships via `ProbeReconcile`).
     static GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_COUNT: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_BYTES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // UPDATE broadcast merges skipped by the per-contract merge-failure backoff
+    // (poison-contract quarantine, #4861).
+    static GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
 }
 
 /// Global test metrics for tracking events across the simulation network.
@@ -3464,6 +3467,7 @@ impl GlobalTestMetrics {
         GLOBAL_PUT_PROBE_NEW_CONTRACT_BYTES.with(|c| c.set(0));
         GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_COUNT.with(|c| c.set(0));
         GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_BYTES.with(|c| c.set(0));
+        GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF.with(|c| c.set(0));
     }
 
     /// Records that a ResyncRequest was received.
@@ -3475,6 +3479,18 @@ impl GlobalTestMetrics {
     /// Returns the total number of ResyncRequests received since last reset.
     pub fn resync_requests() -> u64 {
         GLOBAL_RESYNC_REQUESTS.with(|c| c.get())
+    }
+
+    /// Records that an UPDATE broadcast merge was skipped by the per-contract
+    /// merge-failure backoff (poison-contract quarantine, #4861).
+    pub fn record_merge_suppressed_by_backoff() {
+        GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Returns the number of merges skipped by the merge-failure backoff since
+    /// last reset.
+    pub fn merges_suppressed_by_backoff() -> u64 {
+        GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF.with(|c| c.get())
     }
 
     /// Records that a delta was sent in a state change broadcast.
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3424,6 +3424,10 @@ std::thread_local! {
     // UPDATE broadcast merges skipped by the per-contract merge-failure backoff
     // (poison-contract quarantine, #4861).
     static GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // ResyncRequest emissions / ResyncResponse sends suppressed by the resync
+    // rate limiters (#4861).
+    static GLOBAL_RESYNC_REQUESTS_SUPPRESSED: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_RESYNC_RESPONSES_SUPPRESSED: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
 }
 
 /// Global test metrics for tracking events across the simulation network.
@@ -3468,6 +3472,8 @@ impl GlobalTestMetrics {
         GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_COUNT.with(|c| c.set(0));
         GLOBAL_PUT_PROBE_EXISTING_MESH_DELTA_BYTES.with(|c| c.set(0));
         GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF.with(|c| c.set(0));
+        GLOBAL_RESYNC_REQUESTS_SUPPRESSED.with(|c| c.set(0));
+        GLOBAL_RESYNC_RESPONSES_SUPPRESSED.with(|c| c.set(0));
     }
 
     /// Records that a ResyncRequest was received.
@@ -3491,6 +3497,28 @@ impl GlobalTestMetrics {
     /// last reset.
     pub fn merges_suppressed_by_backoff() -> u64 {
         GLOBAL_MERGES_SUPPRESSED_BY_BACKOFF.with(|c| c.get())
+    }
+
+    /// Records that a `ResyncRequest` emission was suppressed by the per-contract
+    /// resync emit rate limiter (#4861).
+    pub fn record_resync_request_suppressed() {
+        GLOBAL_RESYNC_REQUESTS_SUPPRESSED.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Returns the number of ResyncRequest emissions suppressed since last reset.
+    pub fn resync_requests_suppressed() -> u64 {
+        GLOBAL_RESYNC_REQUESTS_SUPPRESSED.with(|c| c.get())
+    }
+
+    /// Records that a `ResyncResponse` send was suppressed by the
+    /// per-(peer, contract) responder rate limiter (#4861).
+    pub fn record_resync_response_suppressed() {
+        GLOBAL_RESYNC_RESPONSES_SUPPRESSED.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Returns the number of ResyncResponse sends suppressed since last reset.
+    pub fn resync_responses_suppressed() -> u64 {
+        GLOBAL_RESYNC_RESPONSES_SUPPRESSED.with(|c| c.get())
     }
 
     /// Records that a delta was sent in a state change broadcast.
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -543,8 +543,28 @@ impl ConfigArgs {
                 .get_or_insert(cfg.per_user_inactive_ttl_secs);
             self.inactive_user_sweep_interval_secs
                 .get_or_insert(cfg.inactive_user_sweep_interval_secs);
-            self.module_cache_budget_bytes
-                .get_or_insert(cfg.module_cache_budget_bytes);
+            // #4864 upgrade migration: an existing node whose config.toml was
+            // auto-written on a >12 GiB box carries the OLD auto-derived clamp
+            // `module-cache-budget-bytes = 1610612736` (1.5 GiB, the previous
+            // MAX_DEFAULT_MODULE_CACHE_BUDGET_BYTES, which only ever appeared as
+            // an auto-derived value on boxes with >12 GiB RAM). A plain
+            // get_or_insert would merge that stale 1.5 GiB back and pin the node
+            // to it forever, so the exact large gateways this change targets
+            // would NEVER see the new 4 GiB default. That exact value is the only
+            // distinguishable "this was auto-derived, not operator-chosen" signal
+            // we have, so treat it as auto: skip the merge, leave self None, and
+            // let build() RE-DERIVE it via default_module_cache_budget_bytes()
+            // below (yielding 4 GiB on a large box). CLI/env explicit values are
+            // parsed into `self` BEFORE this file merge, so they still win. Any
+            // OTHER persisted value is an explicit operator choice and is
+            // preserved unchanged. Accepted, release-notes-worthy edge: an
+            // operator who EXPLICITLY set exactly 1.5 GiB is indistinguishable
+            // from the old auto-default and will be re-derived.
+            const OLD_AUTO_MODULE_CACHE_BUDGET_SENTINEL: usize = 1_610_612_736;
+            if cfg.module_cache_budget_bytes != OLD_AUTO_MODULE_CACHE_BUDGET_SENTINEL {
+                self.module_cache_budget_bytes
+                    .get_or_insert(cfg.module_cache_budget_bytes);
+            }
             self.shutdown_drain_secs
                 .get_or_insert(cfg.shutdown_drain_secs);
             self.max_blocking_threads
@@ -4916,6 +4936,81 @@ mod tests {
         assert_eq!(
             iface_tx_enabled, seed.telemetry.iface_tx_enabled,
             "iface_tx_enabled"
+        );
+    }
+
+    #[tokio::test]
+    async fn module_cache_budget_old_auto_sentinel_re_derives_but_explicit_values_persist() {
+        // #4864: an existing node whose config.toml was auto-written on a >12 GiB
+        // box carries the OLD auto-derived 1.5 GiB clamp
+        // (module-cache-budget-bytes = 1_610_612_736, the previous
+        // MAX_DEFAULT_MODULE_CACHE_BUDGET_BYTES). build()'s merge must treat that
+        // exact sentinel as "auto" and RE-DERIVE the current default (so large
+        // gateways pick up the new 4 GiB clamp instead of staying pinned to
+        // 1.5 GiB forever), while PRESERVING any other persisted value (a genuine
+        // operator choice) and DERIVING when the field is absent.
+
+        // Resolve the freshly-derived default the SAME way the production default
+        // fn does, so the assertions are host-independent (no hard-coded number).
+        // On a box where the derived default happens to equal the sentinel (exactly
+        // 12 GiB RAM) case 1 still holds: re-derivation is a no-op there.
+        let derived_default = crate::wasm_runtime::default_module_cache_budget_bytes();
+
+        // Seed config.toml with `module-cache-budget-bytes = seed` (or omit the key
+        // when `seed` is None), rebuild from clap-bare args, and return the resolved
+        // value. Mirrors all_persisted_config_fields_round_trip_through_build: a base
+        // build creates the on-disk secret files + a valid full Config to serialize.
+        async fn rebuilt_budget(seed: Option<usize>) -> usize {
+            let temp_dir = tempfile::tempdir().unwrap();
+            let mut base = clap_bare_args(temp_dir.path()).build().await.unwrap();
+            let toml_str = match seed {
+                Some(v) => {
+                    base.module_cache_budget_bytes = v;
+                    toml::to_string(&base).unwrap()
+                }
+                None => {
+                    // Drop the field so build() sees it as absent and the serde
+                    // default fills it — exercises the "absent -> derive" path.
+                    // The key is a top-level scalar, so removing its single line
+                    // keeps the TOML valid.
+                    toml::to_string(&base)
+                        .unwrap()
+                        .lines()
+                        .filter(|l| !l.trim_start().starts_with("module-cache-budget-bytes"))
+                        .collect::<Vec<_>>()
+                        .join("\n")
+                }
+            };
+            std::fs::write(temp_dir.path().join("config.toml"), toml_str).unwrap();
+            clap_bare_args(temp_dir.path())
+                .build()
+                .await
+                .unwrap()
+                .module_cache_budget_bytes
+        }
+
+        // 1. Exact old-auto sentinel -> RE-DERIVED to the fresh default (NOT the
+        //    stale 1.5 GiB value).
+        let from_sentinel = rebuilt_budget(Some(1_610_612_736)).await;
+        assert_eq!(
+            from_sentinel, derived_default,
+            "the old auto-derived 1.5 GiB sentinel must re-derive to the fresh \
+             default, not stay pinned at 1_610_612_736"
+        );
+
+        // 2. Some other explicit persisted value -> preserved exactly.
+        let explicit = 777_000_000usize;
+        let from_explicit = rebuilt_budget(Some(explicit)).await;
+        assert_eq!(
+            from_explicit, explicit,
+            "an explicit non-sentinel operator value must round-trip unchanged"
+        );
+
+        // 3. Field absent from config.toml -> resolves to the derived default.
+        let from_absent = rebuilt_budget(None).await;
+        assert_eq!(
+            from_absent, derived_default,
+            "an absent field must resolve to the derived default"
         );
     }
 
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1234,7 +1234,7 @@ pub struct Config {
     /// ~1.25× this value. Bounds the cache by total compiled bytes rather than
     /// entry count, so a node hosting many contracts doesn't thrash (issue
     /// #4441). When unset, the default scales with system RAM
-    /// (`clamp(total_ram / 8, 64 MiB, 1.5 GiB)`) so a small VPS doesn't OOM and
+    /// (`clamp(total_ram / 8, 64 MiB, 4 GiB)`) so a small VPS doesn't OOM and
     /// a big gateway still caches a large working set.
     #[serde(
         default = "default_module_cache_budget_bytes",
@@ -1354,7 +1354,7 @@ const fn default_inactive_user_sweep_interval_secs() -> u64 {
 }
 
 /// Default contract-module cache byte budget, scaled to system RAM
-/// (`clamp(total_ram / 8, 64 MiB, 1.5 GiB)`).
+/// (`clamp(total_ram / 8, 64 MiB, 4 GiB)`).
 ///
 /// Resolves to [`crate::wasm_runtime::default_module_cache_budget_bytes`], the
 /// single source of truth, so the operator-facing default and the in-code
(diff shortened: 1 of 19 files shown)


sanity (commented):
## Comprehensive PR Review: #4864

### Summary
- **PR Title:** fix(update): contain poison-contract CPU churn — enforced WASM preemption, merge-failure backoff, resync rate limits (#4861)
- **Type:** fix
- **CI Status:** running at review time (heavy suites triggered on ready-for-review)
- **Linked Issues:** #4861
- **Review tier:** Full (WASM engine, UPDATE op state machine, concurrency — all high-risk surfaces)
- **Reviewers run:** freenet:code-first-reviewer, freenet:testing-reviewer, freenet:skeptical-reviewer, freenet:big-picture-reviewer (parallel, independent) + external Codex pass (`codex review --base origin/main`). Freenet bug-pattern check run against the diff (clean: no `.send().await`, no wire-enum changes, no `biased;`, TTL-bound maps, jittered backoff, DST-compliant).

---

### Code-First Analysis
**Independent Understanding:** matches stated intent — epoch-enforced execution budgets, per-contract merge-failure backoff with strictly-delta-only reset, three-layer resync rate limiting, timeout classification, module-cache clamp raise.
**Alignment:** matches. Production epoch wiring verified real (shared backend registered via `create_backend_engine` → `register_engine_for_epoch`; not a silent no-op). Backoff-skip mirrors the dedup-skip completion path exactly; no handshake/cleanup skipped; RAII inflight guard clears tx state on suppressed early-returns (no stuck transactions).
**Gaps:** none blocking; see findings.

---

### Testing Assessment
**Coverage Level:** strong at unit level; wiring-level gaps

| Unit | ✅ | 38 new tests across merge_backoff/resync_rate_limit/executor/module_cache; jitter-robust; DST-clean (TimeSource + seeded GlobalRng) |
| Integration | ⚠️ | Real infinite-loop WAT preemption test (no wall-clock wrapper) — but bypasses the wrapper methods and `epoch_deadline_ticks` |
| Simulation | ⚠️ | No poison-storm sim; mitigant: gate has no feature flag, existing convergence sims route through it |
| E2E | N/A | |

**Regression Test:** present for mechanisms; the driver-level wiring (the actual #4861 fix surface) is pinned by source-scrape only, though a behavioral harness (`build_queue_full_test_node`) exists in the same file.

---

### Skeptical Findings
**Risk Level:** medium (one real false-positive-suppression concern; no correctness bug)

| Benign stale-version InvalidUpdate rejections trip the backoff | **Medium** | op_ctx_task.rs:1268, :2092 (verified) | `record_failure` gates on `is_contract_exec_rejection()` which includes benign stale-version rejects (#3914 documents these on every re-broadcast missed by dedup). One stale reject → 30s cooldown during which ALL merges are skipped, so a clean success cannot land to clear the entry (success starvation) → possible escalation. CRDT contracts unaffected (stale merges to no-change = success). Cannot exclude InvalidUpdate wholesale (fork poison presents identically); the discriminator is *consecutive-without-success*. |
| Client-local delta success does not clear backoff | **Medium** | op_ctx_task.rs:1257 is the only `record_success` site (verified) | [Codex P2] A contract recovering via `drive_client_update` delta stays suppressed for the remaining cooldown (up to 2h) for inbound broadcasts. |
| Epoch deadline can fire up to one tick early | **Medium** | wasmtime_engine.rs:397-401 (verified) | [Codex P2] Ticker phase not synchronized with arming: 50-tick deadline can trap at ~4.9s, nondeterministically killing guests that legitimately finish in [4.9s, 5.0s]. Needs +1 phase-safety tick. |
| Epoch trap escapes timeout classification on init/allocator paths | **Medium** | wasmtime_engine.rs:643 (`Instantiation`), :731 (`Runtime`) (verified) | [Codex P2 + skeptical, deduped] `instantiate_async`, `__frnt_set_id`, `initiate_buffer` map errors without `classify_runtime_error`, so a runaway start-function/allocator gets misclassified — not `is_wasm_timeout` (wrong cooldown class) and not even `is_contract_exec_rejection` (not recorded at all). |
| Responder (peer,contract) limiter map not pruned on disconnect | Low | resync_rate_limit.rs:146-151 | Idle-TTL sweep only; under extreme peer churn the 16384-cap vacant-at-capacity branch suppresses a new legit key until the ~5min sweep. Bounded (~1MB), self-healing, global cap still bounds storms. Matches UpdateRateLimiter precedent. |
| Verified-OK highlights | — | — | Full-state/streaming failures DO normalize through the `Update` error variant (pinned by executor test); no stuck-tx on suppression (RAII guard); permanent-stuck unreachable (cooldown/memo/reaper all self-expire); all 8 guest-entry paths re-arm the deadline; store-reuse-after-trap is pre-existing wasmtime behavior, not new. |

---

### Big Picture Assessment
**Goal Alignment:** yes — all three production poison classes + the module-cache amplifier addressed; defense-in-depth verified (no unbounded resync path remains).
**Anti-Patterns Detected:** none (bug-pattern check clean).
**Removed Code Concerns:** none — zero tests removed/weakened; existing dedup, counters, and #4857 heal-path throttle preserved.
**Scope Assessment:** focused; module-cache clamp is the most separable piece but maps directly to the issue's documented third mechanism — acceptable in-PR.
**CONTRIBUTING gate:** cleared — #4861 is the maintainer-authored issue with this exact fix plan.

---

### Documentation
- Code docs: good (module docs carry the three poison classes + containment philosophy, citing #4861)
- Architecture docs: n/a
- Needs: (a) one sentence at the ResyncResponse-apply merge noting it is bounded by the emit limiter + epoch preemption, not the backoff; (b) release-note items: default module-cache budget rises 1.5→4 GiB on 32+ GiB hosts; one-time full recompile on first start (epoch_interruption changes the compile-cache key); (c) parent-repo hosting-invariants: consider adding "merge-failure cooldown" to invariant 1's permitted transients (owner decision, out of this repo).

---

### Recommendations

#### Must Fix (Blocking)
1. **Invalid-class trip threshold**: require N=3 consecutive Invalid-class failures before the first cooldown (Timeout stays trip-at-1 — each timeout is a full 5s burn). Eliminates single-benign-stale-reject blackouts and success starvation; the fork storm (~2 failures/min, zero clean deltas) still trips within ~90s. Add a regression test: healthy busy contract with interleaved benign rejects + successes never enters backoff. (Skeptical M1)
2. **Clear backoff on successful client-local DELTA merge** in `drive_client_update` (delta-only, same rule as the relay driver). (Codex P2)
3. **+1 phase-safety tick** in `epoch_deadline_ticks` so the enforced budget is ≥ configured timeout (trap window becomes [T, T+100ms] instead of [T−100ms, T]); unit-test the function. (Codex P2; also closes testing gap #3)
4. **Classify epoch interrupts on all guest-entry paths** — route `instantiate_async`/`__frnt_set_id`/`initiate_buffer` errors through `classify_runtime_error` (or an explicit `Trap::Interrupt` check) so init-phase runaways get the Timeout quarantine. (Codex P2 + skeptical L3)
5. **Behavioral driver test**: using the existing `build_queue_full_test_node` harness with an exec-rejection stub, drive `drive_relay_broadcast_to` twice and assert the second merge is skipped (no second UpdateQuery; suppressed-counter == 1). The load-bearing #4861 behavior must not be source-scrape-only. (Testing H1)

#### Should Fix (Important)
6. Behavioral driver test that a full-state apply does NOT reset the backoff (same harness). (Testing #2)
7. Source-scrape pin: every guest-call site is preceded by `arm_epoch_deadline` (guards the unbounded-guest bug against future refactors). (Testing #4)
8. Split the responder suppression metric into per-peer vs global counters (currently indistinguishable). (Code-first nit)
9. The ResyncResponse-apply bounding comment (Documentation (a)).

#### Consider (Suggestions)
10. Follow-up sim test: merge-always-fails stub contract, assert contained storm (positive containment claim currently unproven at sim level).
11. Lazy payload hashing in the Allow path (micro-optimization; eager hash is negligible vs the merge it avoids — fine to decline with a comment).
12. Release-note items from Documentation (b).

---

### Verdict
**HEAD SHA reviewed:** 935518a2

Five Must Fix items (1–5). None is architectural — the containment design is validated by all five reviewers; these are calibration (trip threshold, phase tick), coverage (client-delta reset, init-path classification), and test-hardening fixes. Re-review required on the updated HEAD because Must Fix items exist and the fix diff will exceed the trivial threshold.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4864 — Round 2 (re-review after fixes)

### Summary
- **Review tier:** Full (re-review; prior round on 935518a2 had 5 Must Fix findings — all addressed in 935518a2..02a40d37)
- **HEAD SHA reviewed:** 02a40d37
- **Reviewers run:** freenet:code-first-reviewer, freenet:skeptical-reviewer, freenet:testing-reviewer, freenet:big-picture-reviewer (fresh full-diff passes) + external Codex (`codex review --base origin/main`)
- **CI:** rule-review warning from round 1 (fire-and-forget ticker) resolved — Claude Rule Review now passes; Clippy/Fmt/NAT/Windows green; Simulation/Unit&Integration running at review time

### Round-1 findings: all 10 verified fixed
All four Claude reviewers independently confirmed the round-1 fixes are correctly implemented: N=3 Invalid trip threshold (no off-by-one; first suppressing cooldown = base), client-local delta reset (delta-gated, not a malicious-client hole — a local client can only clear backoff by producing a genuinely successful merge), +1 epoch phase margin (never-early property proven), guest-entry interrupt classification (no non-interrupt-trap misclassification), behavioral driver tests (drive the REAL driver; injected error classification traced end-to-end; assertions non-vacuous), ticker catch_unwind + heartbeat (panic-sound, no poisoned-lock corruption), and the CI test-model fix (assertion relation unchanged; the case that left is covered by a pre-existing companion test — confirmed not added to paper over the change).

### New findings (this round)

#### Must Fix (Blocking) — both from the external Codex pass, both verified in code
1. **Peer-triggerable contract-wide blackout** (`op_ctx_task.rs:1246-1247`): the Invalid-class backoff is keyed by contract only. Any single connected peer can send 3 distinct invalid deltas (>100ms apart, passing the update rate limiter) and place a healthy contract into contract-wide cooldown; honest peers' valid deltas are then dropped unexecuted and cannot land the success that clears the entry; re-poisoning each expiry escalates toward 30min. **Fix direction:** scope Invalid-class per-(contract, sender) — matches the production fork evidence (each diverged sender trips its own channel) and reduces a lone attacker to blacking out only their own channel. Timeout-class stays contract-wide (a 5s burn is contract-intrinsic real cost). Payload memo stays contract-wide (content-addressed).
2. **Persisted config defeats the cache raise** (`module_cache.rs:1130` + `config.rs:546-547,1223`): `module-cache-budget-bytes` is persisted with the derived default and merged back unconditionally, so existing >12GiB nodes (including the target gateways) stay pinned at 1610612736 and never see the 4GiB default. **Fix direction:** sentinel migration — treat the exact old auto-default value as "auto" and re-derive; preserve other explicit values; release-note the explicit-1.5GiB edge case.

#### Should Fix
3. Class-escalation cooldown inflation: `Invalid,Invalid,Timeout` yields a 480s first Timeout cooldown instead of 120s (exponent counts pre-escalation failures). Floor the exponent on class escalation. (skeptical O1)
4. Stale-ticker detector's first warning can lag up to ~60s after an early ticker death (`last_warn` initialized 0). Treat 0 as never-warned. (skeptical O2)
5. Confirm the deserialization-poison error path (host-side deser failure of a malformed delta) actually satisfies the `is_contract_exec_rejection` recording gate; extend the gate if it surfaces as a different variant. (code-first B)
6. Behavioral test for the client-delta `record_success` path (currently source-scrape-pinned only). (testing M-L)
7. Test additions: pre-trip payload memoization with distinct payloads; ≥3 queue-full errors do not trip the backoff. (testing L)

#### Consider / noted without action
Heartbeat tracks thread-liveness not increment-liveness (documented; per-iteration ERROR is loud); guest-entry pin body over-extension (works correctly today, noted as a sharp edge); streaming behavioral backoff test (redundant with unit + pins); PR description synced to the N=3 threshold (done this round).

### Verdict
Findings 1-2 are adversarial/deployment-reach defects that materially affect the fix's production value; both have clear, scoped fix directions and are in progress. The containment architecture itself is validated by all reviewers across both rounds.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4864 — Round 3 (re-review after per-sender rework)

### Summary
- **HEAD SHA reviewed:** 14213a94 (round-3 commits: config sentinel migration, stale-ticker first-warn, escalation restart, deser classification pin, per-sender Invalid backoff rework with any-sender memo hard-skip)
- **Review tier:** Full — freenet:code-first, testing, skeptical, big-picture reviewers + external Codex
- **Round-2 P1s:** both verified fixed (per-sender scoping eliminates the single-peer blackout — the adversarial regression test proves a second sender's deltas still execute while an attacker's channel is tripped; the config sentinel migration reaches existing >12GiB nodes, with CLI/env precedence intact)

### Verified clean this round
No tests removed or weakened (replaced escalation tests are pinned more strongly by the structural two-channel split); all source-scrape pins anchor real production code after the renames; size-cap accounting balanced; sender-addr provenance is the transport immediate-hop (not payload-spoofable); memo hard-skip is zero-cost in both drivers; behavioral tests non-vacuous (verified line-by-line, including that the key adversarial test would fail on a contract-wide regression).

### New findings (converged from Codex + skeptical, independently)

#### Must Fix
1. **No-change delta "success" clears the contract-side quarantine** (op_ctx_task.rs:1292, :302, :397): `record_success_from_sender`/`_local` fire on any `Ok`, including `changed == false` no-op merges — but a no-change success means the state did NOT advance, so the memoized failures are not stale. On a busy contract with a co-circulating timeout payload, ordinary delta traffic wipes the 120s→2h Timeout quarantine, leaving only the 60s dedup floor (≈one 5s burn/min — ~120× the churn this PR removes), no attacker required. Fix: gate the contract-side clear on `changed == true`; the per-sender Invalid clear stays unconditional (channel-trust vs state-generation signals).
2. **Responder limiter-slot exhaustion** (node.rs:2957): limiter buckets allocate before any existence check — spraying bogus contract keys fills the capped maps and denies responses for legitimate contracts (fails closed). Fix: cheap on-disk existence check before the limiters.
3. **Queue-full resync path bypasses the global emit cap** (op_ctx_task.rs:1356): the pre-existing #4857 helper only has per-(contract,sender) throttling — route it through the per-contract limiter.
4. **Memo staleness after state-advancing full-state applies**: a changed streaming/resync apply invalidates the memo's premise, but memoized payloads stay suppressed ≤10min. Fix: `invalidate_payload_memo` (memo only — cooldowns and the strictly-delta-only no-reset doctrine untouched) at both changed==true full-state sites.
5. **`__frnt_set_id` inherits the instantiation epoch deadline's remainder** (wasmtime_engine.rs:744): a start function finishing near budget leaves ~zero ticks for the next guest call — spurious trap. Re-arm before every guest entry; extend the pin to all call sites per method.

#### Should Fix
Per-class idle-grace for the Invalid map (30min, not 2h — shrinks port-flood fill persistence); stale trip-gated clause in `check()`'s docstring (contradicts the implemented hard-skip); stale 1.5 GiB doc in config.rs; driver-level memo-replay regression test.

#### Declined with justification
First-warn unit test (process-global static, telemetry-only); gauge cross-map dedup test (HashSet by construction); direct CLI-beats-sentinel config test (structurally guaranteed by parse-before-merge + get_or_insert).

PR description updated this round with the per-sender scoping, config migration, and one-time recompile notes.

### Verdict
The containment architecture is validated three rounds running; the remaining defects are state-generation calibration (1, 4) and two amplification leftovers (2, 3) plus a spurious-trap edge (5). Fixes in progress.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4864 — Round 4 (re-review after round-3 fixes)

### Summary
- **HEAD SHA reviewed:** d8989b29 (round-4 commits: changed==true gating, responder existence gate + queue-full global cap, epoch re-arm before `__frnt_set_id`, per-class idle grace + docs + memo-replay test, client full-state no-reset test)
- **Review tier:** Full — all four freenet reviewer perspectives + external Codex
- **Round-3 findings:** all 5 Must-Fix verified correctly implemented end-to-end (changed-flag threading confirmed real at every call site; responder gate ordering pinned; queue-full flood behaviorally bounded to the burst allowance; the strengthened guest-entry pin genuinely detects a missing re-arm; per-class sweep correct)

### New findings

#### Must Fix (round 5, in progress)
1. **Guest-entry epoch timeouts misclassify as Invalid** (Codex P1): the new classifier's `WasmError::Timeout` formats as "execution error: execution timeout", missing the `is_wasm_timeout` needle — a runaway start function gets per-sender Invalid instead of the contract-wide Timeout quarantine. Fix: normalize to `MaxComputeTimeExceeded` at the source.
2. **Existence gate is a no-op on SQLite builds** (Codex P1): `contract_state_present` conservatively returns true without the redb sync probe, reopening the limiter-map key-flood on a supported backend. Fix: real sqlite existence probe (or per-peer allocation budget).
3. **node.rs resync-apply memo-clear ignores `state_changed`** (skeptical Medium + code-first, converged): the match arm discards the flag, so a no-change (CurrentWon) resync apply wipes the contract memo — the live protection for Invalid-class poison when no Timeout is active — and logs `changed=true` inaccurately. Fix: gate on `state_changed: true`, mirroring the streaming site.

#### Should Fix (round 5)
Coverage: `changed==false` not drivable end-to-end (add `SuccessNoChange` harness variant + a pin that call sites pass the real flag); positive control for the bogus-keys responder test; pins for the two `invalidate_payload_memo` call sites. Queue-full double-gate burns the per-sender window on a global-cap suppression (record-on-emit split, or document). Docs: two remaining stale 1.5 GiB docstrings; one-line note that the `__frnt_set_id` re-arm makes create_instance's worst case ~2× the per-call budget (correct, bounded at exactly two guest entries); module-doc §64 changed-qualifier.

#### Verified clean
Changed-flag threading (all four sites real), responder existence probe layering (same store, fails open, no new timing leak), both new behavioral tests non-vacuous (memo-replay distinguishes memo-skips from dedup-skips; client full-state asserts before AND after), per-class sweep accounting, queue-full pin restructure strengthened not diluted, zero tests removed across the round.

PR description updated through round 4 (state-advance gating, responder existence-gate, queue-full cap).

### Verdict
**State:** Needs Changes — Re-review Required After Fix (round 5: 10 scoped items, 3 blocking)
Four rounds in, the architecture is untouched by findings; remaining defects are classification normalization, backend parity, and one gating consistency hole — plus regression-detection coverage. Round 5 is the closing batch.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4864 — Closing (round 9)

### Summary
- **HEAD reviewed:** 7b6a9acd (final: round-9 typed-provenance/store-recovery/memo/PUT fixes + doc footgun + pin-window widening)
- **Review tier:** Full (WASM engine, UPDATE state machine, concurrency, protocol — all high-risk)
- **Rounds:** 9 review rounds. Each surfaced real defects; the architecture has been stable since round 1 (every finding was a localized fix, never a redesign).

### External-model substitution (this round)
Codex — the external non-Claude reviewer used in rounds 1–8 — hit its usage quota this round (resets 2026-07-25). Waiting to that reset was impractical for a critical fix, so per the multi-model-review fallback I ran a **diverse-Claude-lens** independent pass instead of skipping it: three lenses, each blind to the others —
- **skeptical bug-hunt** (races, edge cases, failure modes),
- **provenance / type-flow** (aimed at the load-bearing unforgeability property — Codex's round-9 specialty),
- **resource-leak / concurrency** (store arena, limiter counters, ticker state machine).

All three returned **no blocking findings**. This is recorded so the weaker-signal decision is auditable; the real Codex pass should be re-run before merge if convenient after 07-25, given the high-risk surface.

### Round-9 findings — all resolved
1. **Typed timeout provenance (P1, was latent since round 2).** `is_wasm_timeout`/`is_scheduler_timeout` had matched a *contract-controllable* error string, letting a contract forge the contract-wide trip-at-one Timeout quarantine and suppress honest peers up to 2h. Now gated on an unforgeable typed `host_timeout` field set only on the host `classify_result` path (modeled on the already-typed `is_contract_queue_full`). All three lenses independently verified it is airtight by the type system, survives every error transform, and that a forged cause string classifies false. This is the fix that satisfies the "a malicious contract cannot cause user-noticeable harm" requirement.
2. **Store-arena leak on `__frnt_set_id` init timeout (P1).** Recovery now fires on every guest-entry error path (verified exhaustive), also plugging a pre-existing `ensure_memory`-failure leak.
3. **Memo-invalidation consistency (P2).** Non-streaming full-state + client full-state success now invalidate the payload memo (memo-only, no cooldown reset — no-reset doctrine intact), matching the streaming/resync paths.
4. **PUT-returns-UPDATE-error semantics (P2).** Fresh-PUT validation errors now classify as `Put`, not `Update`; verified no UPDATE-path behavior is suppressed (a fresh PUT never reaches the predicate consumers).
5. **Doc footgun** — a stale comment claiming the scheduler-timeout *string* was load-bearing (the inverse of the new invariant) is corrected with a do-not-refactor warning.

### Deferred with justification (follow-up issues filed)
- **#4871** — `execute_wasm_blocking` scheduler/guest classification race: a just-started guest can run detached one epoch-bounded budget while excluded from backoff. Epoch-bounded and self-correcting; the complete fix rewrites the load-bearing blocking primitive, split out of the final round for focused review.
- **#4872** (P-high) — the resource lens's sharpened caveat on #4871: "epoch-bounded" holds only while the ticker is alive; a dead ticker leaves detached runaway guests unpreemptable, and inline `create_instance` entries have no wall-clock backstop. Pre-existing; round-8 improved ticker resilience (retryable + loud detection) but does not make the dead case *safe*.
- **#4869** — SQLite state-store backend doesn't compile (pre-existing ~15-method drift); the round-5 SQLite existence probe is correct-but-dormant until it's repaired.
- Contract-layer root causes of the actual production poison contracts: **river#422** (room merge exceeds the 5s budget), **river#423** (fork-reconciliation), and the aggregate-CPU **#4865** (PR-2 duty-cycle budget).
- Two pre-existing sim-timing flakies observed during full-suite runs: **#4867**, **#4870**.

### Verdict
**State:** Ready to Merge (pending CI green on 7b6a9acd)
The containment architecture has been validated across nine rounds by four Claude perspectives plus (rounds 1–8) the external Codex model, with every finding fixed or explicitly deferred with a filed issue. Full CI-profile nextest is green (4513 passed / 0 failed, the 2 leaky + sim flakies are pre-existing and tracked). Merge remains a human call — this is a high-risk change and it is being landed via the merge queue; the deploy/release step should be taken deliberately (see the follow-up issues for the residual hardening).

[AI-assisted - Claude]

Issue #3109, opened 2026-02-18
fix: corrupted_state_recovery replaces newer state with older incoming state


sanity:
## Problem

The `upsert_contract_state` corrupted state recovery logic can cause **state regression** — replacing a node's newer contract state with an older version received from the network.

### Observed behavior

On a peer running v0.1.133, the River UI contract (`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) regressed from a recent build (version 29521693, 1.2MB) to a weeks-old build (version 29510075, 922KB):

**Feb 17 04:28** — State successfully updated to version 29521693 (1,246,053 bytes):
```
Contract state updated contract=raAqMhMG7... new_size_bytes=1246053 phase="update_complete"
```

**Feb 18 02:21** — An UPDATE arrives with **older** version 29510075 (922,884 bytes). The merge correctly rejects it:
```
invalid contract update, reason: New state version 29510075 must be higher than current version 29521693
```

But then the corrupted_state_recovery path fires:
```
Merge failed with validated incoming state - local state likely corrupted, 
replacing with incoming state
contract=raAqMhMG7... 
error=New state version 29510075 must be higher than current version 29521693
incoming_state_size=922863
event="corrupted_state_recovery"
```

**Result:** The node replaced its newer state (v29521693) with the older incoming state (v29510075), causing the River UI to regress by several weeks.

### Root cause

The recovery logic in `upsert_contract_state` assumes that when a merge fails on "validated" incoming state, the **local** state must be corrupted. But the merge can fail for a legitimate reason: the incoming state is genuinely older than the local state. In this case, the correct action is to **reject the incoming state**, not replace local state with it.

### Expected behavior

When merge fails because the incoming state version is **lower** than the local version, the node should:
1. Reject the incoming update (the current state is correct and newer)
2. NOT trigger corrupted_state_recovery
3. Optionally log that a stale update was received

The corrupted_state_recovery path should only trigger when there's evidence that the **local** state is actually corrupted (e.g., incoming version is higher but merge still fails).

### Impact

- Any node can have its contract state regressed to an older version
- This is triggered by any peer broadcasting stale state (common after restarts with empty caches)
- Republishing the contract is only a temporary fix — the regression can repeat
- For River, this means users suddenly see an old UI with incompatible delegates, breaking the chat functionality

### Reproduction

Occurs naturally when:
1. A node has up-to-date contract state (version N)
[...]

PR #4904, opened 2026-07-21
fix(update): stop resending doomed deltas to delta-incapable contracts (HQk7 resync loop)

sanity:
## Problem

The production storm around contract `HQk7zGRQagL9AVUMM39C3W3nRVv81xk9BCAeY9MikirK` (an early stdlib-0.1.35 web-container) exposed a sender-side gap in the resync machinery.

HQk7's `update_state` accepts only `UpdateData::State` and rejects every delta with `"Invalid update"`. On 0.2.103 the summary-diff fan-out keeps computing deltas for it, so every broadcast becomes a self-sustaining loop:

```
delta send → "Invalid update" → delta_apply_failed → ResyncRequest →
full-state ResyncResponse → state changes → next delta → repeat
```

Measured: 3,102 `delta_apply_failed` events for this one contract in a 2h window; on the nova gateway alone, 59 `delta_apply_failed`/hr and 311 `resync_request_received`/hr for HQk7. The existing defenses cannot break the cycle:

- `ring/merge_backoff.rs` is RECEIVER-side — the sender keeps computing fresh doomed deltas.
- The #4861 failed-payload memo never matches — every state change produces fresh delta bytes.
- The resync rate limiters bound the full-state amplification, not the doomed delta sends that trigger it.

## Approach

A sender-side, TTL-bounded **delta-incompatibility memo** (`crates/core/src/ring/delta_incompat.rs`), analogous in discipline to `merge_backoff` / `resync_rate_limit`:

- **Arm signals** (counted per contract, `INCOMPAT_TRIP_THRESHOLD = 3` consecutive, mirroring `merge_backoff::INVALID_TRIP_THRESHOLD`'s rationale so one benign stale rejection can't flip a healthy contract):
  1. a `ResyncRequest` arriving within `DELTA_ATTRIBUTION_WINDOW` (60s) of a delta we delivered to that peer (the wire-visible form of the peer's `delta_apply_failed`) — armed in the `ResyncRequest` arm in `node.rs`, BEFORE the rate limiters so a suppressed response still counts;
  2. our own `Invalid`-class DELTA apply failure in the BroadcastTo driver (we fan the same contract to our own downstreams).
- **Effect**: `broadcast_to_single_peer` consults the memo before `compute_delta` and sends FULL STATE directly while armed.
- **Self-healing**: a successful delta apply clears the memo + counter; otherwise it expires after `DELTA_INCOMPAT_TTL` (10 min).
- **Bounded**: strict size caps at insert (8,192 contracts / 16,384 attributions; vacant-at-capacity skipped so churned keys cannot grow the maps) + Ring-reaper sweep.

### The suspected version-gate bypass (investigated, no core fix needed)

The same storm showed HQk7's stored state flip-flopping `2180 ↔ 1952` bytes (fork oscillation), suspected to be a resync/upsert path installing full state WITHOUT re-running the contract's version gate. Investigation outcome — **there is no core-side bypass**:

- Gateway logs show the flips happen inside `upsert_contract_state_deferrable` with `changed=true` and ZERO `corrupted_state_recovery` / `initial_state_installed` events — i.e. the apply went through `attempt_state_update` → the contract's own `update_state`, exactly like a normal update.
- Extracted the actual production WASM + params + state from the gateway store (probe reconstructs the exact key `HQk7zGRQ…`, proving identity) and ran it in the real runtime: its `update_state` ACCEPTS any full state over any current state (v1-over-v2, v2-over-v2, v3-over-v2 all accepted; deltas rejected with `"Invalid update"`). This build simply predates the `version <= current` gate the later web-container has — the flip-flop is the contract accepting both forks, which core cannot second-guess (the contract's `update_state` IS the only version oracle).
- The only two paths that install full state without a WASM merge are the fresh-install branch (gated on `is_new_contract`) and the #3109 corrupted-state recovery (gated on the LOCAL state failing `validate_state`, one-shot guard). Both gates are now pinned by source-scrape tests (`full_state_version_gate_pins` in `executor_impl.rs`) so a future refactor cannot add a blind `state_store` write for the resync path.

Residual (out of scope here, flagging for design discussion): with deltas suppressed, full-state sends to a gate-less contract still "succeed" on both forks, so fork oscillation can continue at the (much lower) full-state broadcast cadence without the resync round-trips. Damping that class for merge-ACCEPTING contracts (merge_backoff only trips on failing merges) is a separate design question.

## Testing

- `ring::delta_incompat::tests` (9 behavioral tests, mock time source): arm-after-threshold via attributed resyncs, unattributed/stale resyncs don't count, receiver-side arm, clear-on-success (memo AND counter), TTL expiry, immediate re-arm after expiry, strict capacity caps, reaper sweep keeps armed entries.
- Source-scrape pins (house style): `broadcast_to_single_peer_gates_deltas_on_incompat_memo` (gate before `compute_delta`, suppression → FullState, attribution gated on `sent_delta`), `resync_request_arm_feeds_delta_incompat_memo_before_limiters` (node.rs), `broadcast_to_feeds_delta_incompat_memo` (driver arms on `is_delta && Invalid` only, clears inside the `if is_delta` success arm), `full_state_version_gate_pins::*` (executor_impl.rs).
- Full `cargo test -p freenet --lib`: 4153 passed / 0 failed. `cargo fmt --check` clean; `cargo clippy -p freenet --lib` clean (the 11 `--tests` errors are pre-existing on the 0.2.103 baseline, count unchanged).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime/executor_impl.rs
+++ b/crates/core/src/contract/executor/runtime/executor_impl.rs
@@ -2621,3 +2621,105 @@ where
         Ok(())
     }
 }
+
+/// Source-scrape pins for the full-state version-gate invariant
+/// (HQk7 fork investigation).
+///
+/// Production observed a fork-oscillation storm where a contract's stored
+/// state flip-flopped between two divergent full states (2180 ↔ 1952 bytes)
+/// on every resync apply. The investigation verified that the CORE side is
+/// correct: every full-state install over EXISTING state routes through the
+/// contract's own `update_state` (via `attempt_state_update`), so a contract
+/// WITH a version gate keeps its higher-version state — the observed flips
+/// were the contract itself accepting both forks (its `update_state` predates
+/// the version gate). These pins keep that invariant true: a future refactor
+/// must not add a blind `state_store` write for the resync/upsert path that
+/// would bypass a well-behaved contract's version acceptance.
+#[cfg(test)]
+mod full_state_version_gate_pins {
+    /// Slice `bridged_upsert_contract_state_inner`'s body (its signature to
+    /// the next method's signature).
+    fn upsert_body() -> &'static str {
+        let src = include_str!("executor_impl.rs");
+        let start = src
+            .find("pub(in crate::contract::executor) async fn bridged_upsert_contract_state_inner(")
+            .expect("bridged_upsert_contract_state_inner not found");
+        let after = &src[start..];
+        let end = after
+            .find("pub(in crate::contract::executor) async fn bridged_summarize_contract_state(")
+            .expect("next method after bridged_upsert_contract_state_inner not found");
+        &after[..end]
+    }
+
+    /// A full state over EXISTING state must reach the contract's
+    /// `update_state` (the contract's version acceptance is the ONLY version
+    /// oracle core has), and the only WASM-merge bypass writing an initial
+    /// state must be gated on the contract being NEW.
+    #[test]
+    fn full_state_over_existing_state_runs_contract_update_state() {
+        let body = upsert_body();
+
+        // The incoming full state becomes an UpdateData::State merge input…
+        let updates_pos = body
+            .find("vec![UpdateData::State(incoming_state.clone().into())]")
+            .expect("full state must be wrapped as UpdateData::State for the WASM merge");
+        // …consumed by attempt_state_update (which invokes update_state).
+        let merge_pos = body
+            .find(".attempt_state_update(&params, &current_state, &key, &updates)")
+            .expect("bridged upsert must merge via attempt_state_update");
+        assert!(
+            updates_pos < merge_pos,
+            "the full-state update input must feed attempt_state_update \
+             (updates {updates_pos} < merge {merge_pos})"
+        );
+
+        // The direct initial-state store must be inside the is_new_contract
+        // branch: the gate check must precede the single `.store(` call.
+        let is_new_pos = body
+            .find("if is_new_contract {")
+            .expect("is_new_contract gate not found");
+        let store_pos = body
+            .find(".store(key, state_to_store, params.clone())")
+            .expect("initial-state store call not found");
+        assert!(
+            is_new_pos < store_pos,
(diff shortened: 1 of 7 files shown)


sanity:
Review hardenings landed in 35eda7b8 (all four items from the consolidated review):

1. **≥2-distinct-peer corroboration for resync-driven arming** (MED). The wire `ResyncRequest` is ambiguous — receivers also emit it on queue-full drops (#4857) and Timeout-class WASM failures, i.e. pure load. Resync strikes now count toward arming only once they have come from at least two DISTINCT peer addresses (attribution was already per-`(contract, peer)`), so a single overloaded or malicious peer can never false-arm a healthy contract into full-state fan-out. Local `Invalid`-class delta-apply failures remain contract-wide (genuine per-node evidence). New test: `single_peer_resyncs_do_not_arm_but_two_distinct_peers_do` (single-peer strikes at 3× the threshold never arm; one strike from a second distinct peer corroborates and arms).
2. **PUT summary-first gate**: `try_summary_first_put` now consults `suppress_deltas` before computing the ProbeReconcile delta and falls through to the full-state driver for an armed contract. Pinned by `summary_first_put_gates_delta_on_incompat_memo`.
3. **Bug-1 pin gap closed**: new node.rs source-scrape pin `resync_response_arm_applies_via_update_query_not_blind_store` asserts the ResyncResponse arm applies state via `ContractHandlerEvent::UpdateQuery` (the contract's own validate/update path) and contains no `PutQuery`/direct-store — covering the likelier blind-store regression site that the executor-body pins alone did not.
4. **Threshold off-by-one fixed**: the arm check now runs uniformly on every strike (including the entry-creating one), so `INCOMPAT_TRIP_THRESHOLD` means exactly "arm at N effective strikes" for any N; the constant's doc states the semantics.

Verification on 35eda7b8: `cargo test -p freenet --lib` 4156 passed / 0 failed; `cargo fmt --check` clean; `cargo clippy -p freenet --lib` clean. Diff vs main: 7 files, +1,034 (delta_incompat.rs 609 of those).

[AI-assisted - Claude]

PR #4905, opened 2026-07-21
ci: skip binstall smoke test while a release is mid-publish

sanity:
## Problem

The nightly binstall smoke test (05:00 UTC cron) hard-failed on 2026-07-21 ([run 29805611911](https://github.com/freenet/freenet-core/actions/runs/29805611911)) and paged both the public River room and the dev room — but nothing was actually broken. `cargo binstall freenet@0.2.103` failed with `Fallback to cargo-install is disabled` because no prebuilt binary was reachable.

The cause is a **structural release-ordering race**, not a broken release:

1. `release.yml` job `publish_crates` runs `cargo publish` **first** — crates.io immediately reports the new version as latest.
2. The GitHub release is created **as a draft**.
3. `cross-compile.yml` builds every target, attaches the binaries, and only at the very end runs `gh release edit --draft=false` (~30 min later).

For that whole cross-compile window, crates.io is ahead of the *published* GitHub release, so `cargo binstall` with compile-fallback disabled correctly can't find the asset. `v0.2.103`: draft created `05:51Z`, published `06:20Z`, nightly ran `05:59Z` — squarely inside the window. A `workflow_dispatch` re-run after publish ([run 29846279099](https://github.com/freenet/freenet-core/actions/runs/29846279099)) passed with no code change, confirming the failure was transient. This will recur whenever a release's cross-compile overlaps the nightly cron.

## Solution

Gate the binstall assertions on the GitHub release for the resolved crates.io version actually being **published**. Before installing, look up the release for tag `v{freenet_version}` (both `freenet` and `fdev` binaries ship in that single release):

- **published** (not draft) → run binstall as before;
- **draft, or no release tag yet, younger than 90 min** → a release *in progress*: skip with a `::notice::`, job stays green, next nightly re-checks;
- **draft, or crates.io ahead with no release, older than 90 min** → genuinely stuck: fail loudly (the permanent breakage this test exists to catch — assets that never attached / a release stuck in draft).

This keeps the test's real purpose intact (a genuinely stuck release is still caught once it ages past the threshold, and always by the next nightly) while removing the transient false alarm. It mirrors the transient-hardening this workflow already carries for the 403 rate-limit race (run 28280841654).

## Testing

- YAML parses; every `run:` block passes `bash -n`.
- Guard logic exercised against live data:
  - published `v0.2.103` → **proceed**;
  - missing release published 10 min ago → **skip** (exit 0);
  - missing release published 200 min ago → **error** (exit 1).
  - The draft branch uses identical age arithmetic; against the real event (`v0.2.103` draft 8 min old at nightly time) it skips instead of paging.

No production code touched — CI workflow only.

https://claude.ai/code/session_01QFpcaBmrQ7kfUpRAcK6Feg

### DIFF
--- a/.github/workflows/binstall-smoke-test.yml
+++ b/.github/workflows/binstall-smoke-test.yml
@@ -54,12 +54,89 @@ jobs:
           # under bash), so it's portable across the matrix.
           UA="freenet-core binstall-smoke-test (https://github.com/freenet/freenet-core)"
           for crate in freenet fdev; do
-            version=$(curl -sSf -H "User-Agent: $UA" \
-              "https://crates.io/api/v1/crates/${crate}" \
-              | jq -er '.crate.max_stable_version')
+            json=$(curl -sSf -H "User-Agent: $UA" \
+              "https://crates.io/api/v1/crates/${crate}")
+            version=$(echo "$json" | jq -er '.crate.max_stable_version')
             echo "Latest ${crate}: ${version}"
             echo "${crate}=${version}" >> "$GITHUB_OUTPUT"
+            # Also capture when the freenet version landed on crates.io, so the
+            # readiness guard below can tell an in-progress release (crate just
+            # published, GitHub release not un-drafted yet) from a genuinely
+            # stuck one. Both freenet and fdev binaries ship in the single
+            # `v{freenet_version}` GitHub release, so the freenet timestamp is
+            # the one that matters.
+            if [ "$crate" = "freenet" ]; then
+              published_at=$(echo "$json" \
+                | jq -er --arg v "$version" \
+                    '.versions[] | select(.num == $v) | .created_at')
+              echo "freenet_published_at=${published_at}" >> "$GITHUB_OUTPUT"
+            fi
           done
+
+      # The release pipeline publishes to crates.io FIRST (release.yml
+      # `publish_crates`), then creates the GitHub release as a *draft*, and
+      # cross-compile.yml attaches the binaries and only un-drafts it at the
+      # very end (`gh release edit --draft=false`) ~30 min later. For that
+      # whole cross-compile window crates.io reports the new version as latest
+      # while the GitHub release binaries are still a draft binstall can't see,
+      # so `cargo binstall` (compile fallback disabled) correctly fails. That is
+      # a release *in progress*, not the permanent breakage this smoke test
+      # exists to catch (assets that never attached / a release stuck in draft).
+      # If the nightly cron happens to run inside that window it hard-fails and
+      # pages the dev + River rooms with a false alarm (run 29805611911,
+      # 2026-07-21: v0.2.103 draft 05:51Z, published 06:20Z, nightly ran 05:59Z).
+      # So gate the binstall assertion on the release actually being published;
+      # skip (green, defer to the next nightly) while it is mid-publish, and
+      # only fail if it has been unpublished long enough to be genuinely stuck.
+      - name: Verify release is published (skip if mid-publish)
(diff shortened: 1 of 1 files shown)


sanity:
## Review + one fix (`d89de00f`); rebased onto `main` and marked ready

### This does NOT leave the test permanently skipped

Worth stating plainly, since "skip a smoke test" is the shape of a silent hole. It isn't one here:

| Situation | Behaviour |
|---|---|
| Release published (every normal night) | **runs** the binstall assertions, unchanged |
| Draft / no tag, younger than 90 min | skips green — a release is genuinely in flight |
| Draft older than 90 min | **fails loudly** |
| crates.io ahead, no release, older than 90 min | **fails loudly** |

So the permanent-breakage cases this test exists to catch (assets that never attached, a release stuck in draft) still fail, just once they age past the threshold — and always by the next nightly.

### The premise still holds; this is structural, not a stale one-off

Checked rather than assumed. The test **currently runs nightly on `main` and is green** — 07-22 through 07-27 all succeeded; the single failure was 2026-07-21, the release-window race. The guard is not on `main`.

The race is structural and recurring, not tied to one release: `release.yml` publishes to crates.io **first**, the GitHub release is created as a **draft**, and `cross-compile.yml` un-drafts it only at the very end. Recent draft windows:

```
v0.2.109  created 16:25:00Z  published 16:40:32Z  (~15 min)
v0.2.110  created 20:35:19Z  published 20:50:59Z  (~16 min)
v0.2.111  created 00:51:37Z  published 01:07:57Z  (~16 min)
v0.2.112  created 14:39:41Z  published 15:01:05Z  (~21 min)
```

Any release whose cross-compile spans the 05:00 UTC cron reproduces it. `STALE_MIN=90` leaves roughly 4x headroom over the observed 15-21 min windows.

### The bug I found and fixed

The guard collapsed **every** `gh release view` failure into "no release exists":

```bash
release_json=$(gh release view "$TAG" ... 2>/dev/null) || release_json=""
```

On a normal night the release is published and long-established, so a transient failure (rate limit, 5xx, network) fell through to the staleness branch, computed an age far past `STALE_MIN`, and **exited 1** — paging the dev and River rooms.

That is exactly the false alarm this PR exists to remove, so as written the guard would have introduced a fresh instance of the bug it fixes. Especially unwelcome here: this workflow already carries hardening for a 403 rate-limit race (run 28280841654), so transient GitHub API failures on this path are a known event, not a hypothetical.

Now a genuine "release not found" (a real signal) is distinguished from any other `gh` failure. A transient failure is not evidence about the release, so it skips green and defers to the next nightly; a genuinely stuck release is still caught either by the not-found path once it ages past the threshold, or the following night.

### Verification

Replayed the guard logic verbatim against an injectable fake `gh`:

```
published release (normal night)           -> PROCEED                  exit=0
draft 10min old (release in progress)      -> SKIP-draft-in-progress   exit=0
draft 200min old (genuinely stuck)         -> ERROR-draft-stale        exit=1
no release, crates.io 10min (racing)       -> SKIP-no-release-yet      exit=0
no release, crates.io 500min (stuck)       -> ERROR-no-release-stale   exit=1
transient gh 503, long-published version   -> TRANSIENT-SKIP           exit=0   <- was ERROR
```

The last row is the regression fixed; the two ERROR rows confirm the loud-failure purpose is intact.

YAML parses; the guard passes `bash -n`. Rebased onto current `main` (`6bd75e2d`), no conflicts.

Marked **ready for review** deliberately: freenet-core drafts skip the heavy CI jobs, so a draft's green is not a real signal.

[AI-assisted - Claude]

Issue #2776, opened 2026-01-20
Tracking: Graceful upgrades for delegates and contracts


sanity:
## Summary

When delegate or contract code changes, the `DelegateKey` or `ContractKey` changes because they are derived from the code hash. This means:
- Old data stored under the previous key becomes orphaned
- Applications need to implement their own migration logic
- There's no standard mechanism for upgrade coordination

## Recent Experience: River Chat Delegate Upgrade

When adding the signing API to River's chat delegate, the delegate key changed from:
```
[26, 147, 48, 130, ...] (code_hash: "8n6hw3vmym1qrvpbaunfnn5t8v1xzdmuiyaprtckwbpz")
```
to:
```
[106, 146, 176, 14, ...] (code_hash: "e7nsoec3be367u1sjvfbeeseudl4rkxesrbz6qmfrcfk")
```

### Workaround Implemented

1. **Stored legacy delegate key as constant** in UI code
2. **On startup**: Fire migration request to legacy delegate
3. **Response handler**: If data received from legacy delegate key, save to new delegate
4. **localStorage flag**: Track migration completion to avoid repeated attempts
5. **TODO comment**: Remove migration code after 2026-03-01

### Challenges Encountered

- Had to manually extract legacy delegate key bytes before deployment
- Migration code is application-specific and error-prone
- No way to know if legacy delegate code still exists on peer
- Response from legacy delegate returned 0 values (unclear if no data or code not present)
- Timing: Must deploy new code while old code is still running to capture the migration window

### 1. Delegate/Contract Versioning
- Built-in version field that doesn't affect key derivation
- Query delegate/contract by base key + version range
- Automatic routing to latest compatible version

### 2. Data Migration Support
- Standard `migrate()` callback when new version registered
- Access to previous version's data during migration
- Atomic migration with rollback on failure

### 3. Upgrade Coordination
- Notify subscribers when new version available
- Grace period before old version deprecated
- Ability to query which versions are available on network

### 4. Backwards Compatibility
- New delegate can declare it handles messages for old versions
- Message routing based on semantic version compatibility
- Clear deprecation signals

## Questions to Consider

1. Should key derivation exclude code hash entirely for upgradeable delegates?
2. How to handle breaking changes in state format?
3. Should there be a "migrate all data" admin operation?
4. How does this interact with contract sharding/replication?

## Related

- Delegate storage and secrets management
- Contract state format evolution
- Network-wide upgrade coordination

[AI-assisted - Claude]

[...]

PR #4908, opened 2026-07-22
feat(delegates): node-side one-shot secret copy-forward on registration with predecessors (#4117)

### LINKED ISSUE #4117: Delegate secrets are wiped on every code change, with no migration path
# Delegate secrets are wiped on every code change, with no migration path

## Summary

Delegate secret storage is partitioned by a key that is itself a hash
of `(params, code_wasm_bytes)`. Any change to the delegate's
compiled WASM — including unrelated source-level changes elsewhere in
the workspace that just happen to recompile the delegate (a new
field on a shared type, a `cargo update`, a different `rustc`
version) — produces a new key, a new on-disk namespace, and an
"empty new install" experience for every existing user. There is no
operator-facing or in-protocol mechanism to migrate or carry over
secrets across this transition.

For applications that use delegates to persist anything cosmetic or
configuration-like (display name, theme,
remembered settings), this means every delegate rebuild silently
resets the user's experience.

## Repro

A minimal delegate that stores one value in the secret store via
`ctx.set_secret(SECRET_ID, bytes)` and reads it back via
`ctx.get_secret(SECRET_ID)`:

1. `fdev publish delegate --code identity_delegate delegate` → delegate
   registered at key `K1`. Files appear under
   `~/.local/share/freenet/secrets/<K1_b58>/`.
2. Webapp writes a secret via the delegate. File appears at
   `~/.local/share/freenet/secrets/<K1_b58>/<SECRET_ID_b58>`.
3. Add a trivial change anywhere in the delegate's source tree (e.g.,
   add an `Option<String>` field to a `serde`-friendly struct in a
   workspace-shared crate consumed by the delegate). Rebuild +
   re-publish:
   `fdev publish delegate --code identity_delegate delegate` →
   delegate now registered at key `K2 ≠ K1`.
4. Webapp re-issues the same secret-read RPC. The delegate's
   `ctx.get_secret(SECRET_ID)` returns `None` (path
   `~/.local/share/freenet/secrets/<K2_b58>/<SECRET_ID_b58>` does
   not exist). The webapp falls back to its default value.

Observed on a live `freenet-core` 0.2.56 gateway during three
successive delegate rebuilds; each rebuild produced a new key and a
fresh empty secrets namespace.

## Root cause

`crates/freenet-stdlib/src/delegate_interface.rs:195-231`:

```rust
pub struct DelegateKey {
    key: [u8; DELEGATE_HASH_LENGTH],
    code_hash: CodeHash,
}

impl DelegateKey {
    fn from_params_and_code<'a>(
        params: impl Borrow<Parameters<'a>>,
        wasm_code: impl Borrow<DelegateCode<'a>>,
    ) -> Self {
        let code = wasm_code.borrow();
        let key = generate_id(params.borrow(), code);   // hash(params, code)
        Self { key, code_hash: *code.hash() }
    }

    pub fn encode(&self) -> String {
        bs58::encode(self.key)              // only `self.key`, not `code_hash`
            .with_alphabet(bs58::Alphabet::BITCOIN)
            .into_string()
    }
}
```

`crates/core/src/wasm_runtime/secrets_store.rs:108-167`:

```rust
pub fn store_secret(
    &mut self,
    delegate: &DelegateKey,
    key: &SecretsId,
    plaintext: Vec<u8>,
) -> RuntimeResult<()> {
    let delegate_path = self.base_path.join(delegate.encode());
    let secret_file_path = delegate_path.join(key.encode());
    ...
}
```

So the on-disk path is `base/<encode(self.key)>/<secret>`, and
`self.key = hash(params, code_wasm)`. The wasm bytes are part of the
hash → any rebuild that changes wasm bytes changes the path. The
`code_hash` field exists on the struct but is not used to construct
the storage path; it is only carried for `code_hash()` accessor
calls.

The corresponding cipher entry (`SecretsStore::ciphers`) is also
keyed by `DelegateKey`, but `fdev publish delegate` uses the same
default cipher every time, so the encryption key stays the same
across rebuilds — files moved between directories would decrypt
cleanly. The blocker is the path, not the crypto.

## Why this matters

Delegate-backed persistence is the only reliable per-user store
available to sandboxed webapps right now: the iframe runs at a "null"
origin in the default Tier-1 sandbox, so `localStorage` and
`indexedDB` don't survive a reload, and the manifest-driven
`allow-same-origin` opt-in (PR #4086) was reverted shortly after
landing.

The result is that an application developer has a binary choice:

- Use delegate secrets → every iterate-publish-deploy of the delegate
  wipes every existing user's profile.
- Use plain contracts → state survives code changes, but it's public,
  costs DHT bandwidth on every update, and the data needs to be
  signed/encrypted separately (which lands back in the delegate for
  key management).

There is no path that preserves user state across a delegate code
change.

## What "control over what is wiped" would look like

A few shapes worth considering:

### A. Decouple the storage path from `code_hash`

Make `DelegateKey::encode()` include both the params-derived id
and the explicit `code_hash`, but use a *separate* stable component
for the secrets path. For example:

```rust
pub struct DelegateKey {
    /// content-addressed id, function of (params, code)
    key: [u8; DELEGATE_HASH_LENGTH],
    /// explicit code hash, copy of code's hash() for routing
    code_hash: CodeHash,
    /// optional stable namespace; if set, used for the secrets path
    /// in place of `key`. Carried in `RegisterDelegate` payload so
    /// operators can keep a logical identity across rebuilds.
    stable_namespace: Option<[u8; 32]>,
}
```

`secrets_store.rs` would then use `stable_namespace.unwrap_or(self.key)`
when constructing the path. Operators that don't set it get the
current behavior; operators that do set it explicitly opt into
across-rebuild persistence.

Concern: lets two different code blobs claim the same secret namespace
on the same node. A malicious republish of an unrelated delegate
with the same namespace would access prior secrets. Mitigations:
require the namespace to be derived from a stable per-operator
keypair (e.g., the registration signing key the operator already
owns), or accept that the trust boundary on a single node is the
operator — the same operator pointing at the same namespace
deliberately is reasonable.

### B. Two-tier path with a separate migration RPC

Path becomes `base/<key_b58>/<code_hash_b58>/<secret>`. On
`RegisterDelegate` of a new code_hash, the runtime walks the
parent dir, finds a sibling code_hash dir, and either:

- automatically moves it (operator-opt-in via registration flag), or
- exposes a `MigrateSecrets { from_code_hash, to_code_hash }` RPC
  the operator can call from `fdev`.

Concern: more moving parts, but no breaking changes to user-facing
crypto.

## Suggested next step

Triage and decide the right shape. The application side is willing
to encode whatever the protocol prefers (an explicit
`stable_namespace` argument to `fdev publish delegate`; a flag at
`RegisterDelegate` time; running a separate `fdev migrate` step on
every deploy). What is missing today is *any* operator-facing
mechanism beyond manually `mv`-ing directories on the node.

## Notes

- Affected version: `freenet-core` 0.2.56 (current main), and every
  earlier version that shipped the secrets-store path layout.
- Companion context: PR #4086 (manifest-driven webapp permissions)
  was reverted on `upstream/revert-4086-webapp-permissions`, so the
  `allow-same-origin` opt-in that would have made `localStorage` a
  viable fallback is no longer available. This issue is more
  load-bearing now than it would have been with #4086 in place.
- The `code_hash` field already on `DelegateKey` carries the right
  primitive for option (B) — the runtime just doesn't surface it in
  the storage path or in any migration API today.

sanity:
## Problem

A delegate's on-disk key is `BLAKE3(code_hash || params)`, so ANY change to the delegate's compiled WASM mints a new key, a new secret namespace, and an "empty new install" for every existing user (#4117). Delegate-backed secrets are the only reliable per-user store a sandboxed web app has, so every delegate rebuild silently wipes profiles, room keys, and settings. Interim node-side path for the graceful-upgrades workstream (#2776).

## Solution

Node-side primitive behind `DelegateRequest::RegisterDelegateWithPredecessors { delegate, cipher, nonce, predecessors: Vec<DelegateKey> }` (freenet-stdlib 0.8.4, bincode tag 3). On registration the node runs a one-shot, idempotent, **Local-scope** copy-forward of each predecessor's secret VALUES into the successor's namespace — at-rest decrypt (predecessor Local DEK) → re-encrypt (successor Local DEK). Local-only by construction: the Local DEK is `HKDF(node_KEK, delegate_key)` (at rest); the per-user DEK needs the client's `dek_secret` (not at rest).

### Security — H1 same-origin gate (first-writer-wins, no grandfather, None never privileged)

A predecessor delegate key is public-derivable and a malicious web-app IS the client, so client consent cannot gate copy-forward. The node records each delegate's registration origin on its FIRST registration only — IMMUTABLE — and copies a predecessor's secrets ONLY IF the registering request's web-app origin equals that predecessor's recorded first origin. An attacker who re-registers a victim's public WASM later records nothing and can never satisfy the gate. Refusals (report `origin_refused`; not a registration failure):
- registering origin ≠ record (mismatch);
- registering origin is `None` / token-less (never privileged in v1 — nothing about a token-less connection is verifiable as locally trusted);
- no record (`no_provenance`) — such legacy data migrates via the app-side `freenet-migrate` path, whose predecessor-delegate OriginPolicy is the sound legacy authorizer.

An origin-LOOKUP failure fails the copy CLOSED (no copy, no seal). A first-writer origin-RECORD write failure FAILS THE ENTIRE REGISTRATION (persistence-succeeds-before-usable): the record is written BEFORE the delegate is registered, and if it cannot be durably persisted nothing is registered — for BOTH `RegisterDelegate` and `RegisterDelegateWithPredecessors`. A registered-but-recordless delegate would leave a claimable first-writer slot an attacker could later name as its own and migrate the accumulated secrets through; registration already depends on delegate-store disk health, so a failing origin-record DB is a visibly sick node and the app simply retries.

### Correctness / seal discipline

Seal only on a fully-clean copy (any read/import/enumeration/marker/reserved-lookup failure leaves the pair UNSEALED so a retry or the app-side path completes it). A collision counts as present only if the existing successor blob decrypts. The shared crate-format marker (`\0freenet-migrate/v1/pred-done:`+key, `b"1"/b"0"`, byte-pinned both repos) is written + verified FIRST, then the redb audit row; any marker failure fully un-seals. An absent predecessor (no on-disk dir) writes nothing. Order: newest-first, first-writer-wins (union-with-newest-precedence); delete-by-absence is app-side only.

### Reserved namespace + amplification bounds

- The whole reserved `\0freenet-migrate/` namespace is excluded from copy-forward via a durable, INDIVIDUALLY-KEYED, capped (256/delegate) redb reserved-hashes table, recorded at `store_secret` time BEFORE the value blob's rename (P1c) — the exclusion lookup fails CLOSED (aborts copy + un-seals; export aborts). `store_secret` REJECTS any reserved-namespace key that is not one of the FOUR pinned marker shapes (P2a): v1 `…/v1/pred-done:` / `…/v1/pred-wip:` + a 32-byte delegate key, or legacy `…/done:` / `…/wip:` + a CANONICAL `u32` decimal generation (non-empty, digits only, no leading zeros except `"0"`, ≤10 digits, ≤ `u32::MAX` — exactly what the crate's `generation.to_string()` produces; non-canonical duplicates like `"00"` and overflow forms like `"4294967296"` are refused). The four-shape matrix is confirmed against freenet-migrate 0.4.0 (no fifth `migrated:` shape) and byte-pinned in both repos (`reserved_marker_shape_matrix_is_stable` node-side; the crate mirror), so any drift trips CI on both sides.
- The client predecessor list is deduped silently and the cap is enforced on the UNIQUE count (matching the docstrings): a duplicate-heavy list whose unique count is within the cap is accepted, while an over-cap UNIQUE list (>64) is REJECTED whole before registration (P2b), never truncated (truncation would strand older generations). A separate raw-length DoS sanity bound (16× the cap = 1024) rejects a giant list up front so the dedupe work itself stays bounded.
- Enumeration carry-over registers each copied secret's raw key under the successor so `list_secret_keys` (River `room:<vk>` discovery) survives a rebuild; `list_secret_keys` + export filter the reserved namespace (L2).

### Match sites

Explicit `RegisterDelegateWithPredecessors` handling at all FOUR `DelegateRequest` match sites (register+migrate handler; `client_events.rs` app-routing no-op; `node.rs` op_name; `contract.rs` initial-params).

## Deploy ordering (LAW)

stdlib publish (done: 0.8.4) → **core merge** → **core release** → only THEN may any client emit tag 3.

**Marker SHAPE-EVOLUTION corollary.** The reserved-shape whitelist (above) is an ENFORCEMENT list, not just documentation — a foreign reserved key is refused rather than silently absorbed — so it inherits the same deploy-ordering obligation as the wire variant: any FUTURE marker shape (e.g. a hypothetical `\0freenet-migrate/v2/…`) MUST ship node-side acceptance in this matrix, in a RELEASED node, BEFORE any crate version writes it. If that ordering is ever violated the failure mode is SAFE and LOUD, never silent: an old node rejects the unknown marker write with a store error, degrading that one migration to a visible Incomplete-retry — the crate re-attempts once the node is upgraded — rather than a silent divergence. This is why the matrix is byte-pinned in both repos.

## Testing

Store-level: copy readable under successor DEK; one-shot/anti-resurrection; overwrite=false; user-scope skipped; enumeration carry-over; first-writer-wins order; crate-format marker + cross-path gating; empty-but-held NoData; absent-predecessor; whole-reserved-namespace exclusion; markers not chain-copied; corrupt-collision unsealed; seal-only-on-clean; H1 (match/mismatch/no-provenance, first-writer immutability, None never privileged); origin-record-failure ABORTS registration (redb-level fault-injected backend failure surfaces as Err not a silent Ok; handler-level: both variants abort under a failing origin-record DB, register nothing, copy nothing, predecessor intact, and a healthy retry after recovery registers + copies); P2a reserved-shape rejection (all four legal shapes accepted; foreign reserved key rejected); canonical-u32 legacy-gen boundaries (`"0"` accepted; `"00"`/`"4294967296"`/11-digit/legacy-u64 rejected); M2 resurrection known-limitation pin; marker-format byte-pin; redb round-trips. Unit: `dedupe_predecessors`. Handler-level (injects a matching origin) + a real-node freenet-ping e2e demonstrating the gate REFUSING a token-less copy end-to-end. `cargo fmt` + whole-workspace `cargo clippy --locked -- -D warnings` clean.

## Known limitation / follow-ups

- **M2 (#4909)**: a permanently-corrupt blob keeps a pair unsealed, so a user-deleted healthy co-resident secret resurrects on re-registration. Deliberate v1 pair-level tradeoff (per-secret sealing needs a crate-contract evolution), symmetric with the app-side path; pinned by a known-limitation test.
- Hosted/User-scope carry-over (re-wrap-on-login) — the at-rest DEK carve-out.
- An authenticated admin migration mechanism (None-class is unprivileged in v1).
- Crate-side byte-pin of the identical four-shape matrix — deferred to the next freenet-migrate PR (0.4.0 already published), tracked with the maintainer.

Refs #2776. Closes #4117.

[AI-assisted - Claude]

https://claude.ai/code/session_014wJpXbgCrnkSKgkpmnLGwQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2161,6 +2161,8 @@ dependencies = [
 [[package]]
 name = "freenet-macros"
 version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d26e5c6bee9ffa0e2683bc8b41a7bd9f1e57e7fd49ac4f6ae433e5a82e9006df"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2268,6 +2270,8 @@ dependencies = [
 [[package]]
 name = "freenet-stdlib"
 version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ffa7063130085b2319850c4407b6ae76de7b5a38cdd12e36a4684ff562cd1ceb"
 dependencies = [
  "arbitrary",
  "bincode",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,7 +149,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -160,7 +160,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1590,7 +1590,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1800,7 +1800,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1866,7 +1866,7 @@ dependencies = [
  "ed25519-dalek",
  "either",
  "freenet",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "futures 0.3.32",
  "glob",
  "hex",
@@ -2061,7 +2061,7 @@ dependencies = [
  "flatbuffers 25.12.19",
  "flate2",
  "freenet-macros 0.1.0",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "freenet-test-network",
  "futures 0.3.32",
  "headers",
@@ -2161,8 +2161,6 @@ dependencies = [
 [[package]]
 name = "freenet-macros"
 version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d26e5c6bee9ffa0e2683bc8b41a7bd9f1e57e7fd49ac4f6ae433e5a82e9006df"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2179,7 +2177,7 @@ dependencies = [
  "clap",
  "freenet",
  "freenet-ping-types",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "freenet-test-network",
  "futures 0.3.32",
  "rand 0.9.4",
@@ -2199,7 +2197,7 @@ name = "freenet-ping-contract"
 version = "0.1.11"
 dependencies = [
  "freenet-ping-types",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "serde_json",
 ]
 
@@ -2209,7 +2207,7 @@ version = "0.1.11"
 dependencies = [
  "chrono",
  "clap",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "humantime",
  "humantime-serde",
  "serde",
@@ -2269,9 +2267,7 @@ dependencies = [
 
 [[package]]
 name = "freenet-stdlib"
-version = "0.8.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ce69f7b87ca5fedb784da0194e26a3315bc23b11e346f7a8413c48abdcac73f"
+version = "0.8.4"
 dependencies = [
  "arbitrary",
  "bincode",
@@ -2333,7 +2329,7 @@ dependencies = [
  "byteorder",
  "ciborium",
  "ed25519-dalek",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "rand 0.9.4",
  "serde",
 ]
@@ -3503,7 +3499,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4065,7 +4061,7 @@ dependencies = [
  "once_cell",
  "png",
  "thiserror 2.0.18",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4197,7 +4193,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -5761,7 +5757,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6265,7 +6261,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "52d1cfed4120b4d927bf7c0f86d2087a4a7d6027c906d9f9d525a80573b9be51"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6667,7 +6663,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6704,7 +6700,7 @@ dependencies = [
 name = "test-contract-integration"
 version = "0.1.11"
 dependencies = [
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "serde",
  "serde_json",
 ]
@@ -6714,14 +6710,14 @@ name = "test-contract-mock-aligned"
 version = "0.1.0"
 dependencies = [
  "blake3",
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
 ]
 
 [[package]]
 name = "test-contract-update-nochange"
 version = "0.1.0"
 dependencies = [
- "freenet-stdlib 0.8.3",
+ "freenet-stdlib 0.8.4",
  "serde",
  "serde_json",
 ]
@@ -7420,7 +7416,7 @@ dependencies = [
  "once_cell",
  "png",
  "thiserror 2.0.18",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -8310,7 +8306,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -8736,7 +8732,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d6f32a0ff4a9f6f01231eb2059cc85479330739333e0e58cadf03b6af2cca10"
 dependencies = [
  "cfg-if",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


sanity:
## Consolidated review record (Full tier)

This PR touches wire format (new `ClientRequest` variant), secrets storage/crypto, and durable state — Full-tier per the multi-model review policy. Review ran per-code-content across every head; each substantive push was re-reviewed.

### Review rounds

**Claude multi-lens fallback** (Codex was quota-blocked for the first rounds; the policy's diverse-lens fallback ran instead, then Codex returned and ran on later heads):

- *Skeptical security lens* — found the three P1s that drove the trust-model redesign: self-authorizing origin gate (any registration appended its origin to the allow-set), unbounded grandfather window with silent origin-write failures creating grandfathered delegates, and fail-open reserved-marker exclusion.
- *Code-first lens* — intent/impl mismatches: seal-before-verify orderings, stale docstrings, the 4th `DelegateRequest` match site left implicit.
- *Testing lens* — drove the failure-injection tests (origin-write failure fails registration, registry-copy failure gates sealing, unreadable collision blocks seal) and the end-to-end WebSocket authenticated-origin test.
- *Big-picture lens* — amplification bounds (predecessor cap, reserved-marker row cap), deploy-ordering law documented in the PR body.

**External Codex passes** (once quota returned): two full passes plus per-delta checks on subsequent heads. Final Codex delta-check on `5952c48d` came back clean — no findings.

**Rule-review bot** (three rounds):
- First run on `5952c48d` was degenerate (warning count 1, no findings text — verified via workflow logs); rerun produced 2 real warnings, both fixed in `b6d37e5`:
  1. Predecessor cap checked on the RAW list before dedupe, contradicting documented deduped semantics → now two-tier: raw sanity bound 1024 → dedupe → cap 64 on UNIQUE count, reject-whole (never truncate).
  2. Over-cap rejection untested end-to-end → added `register_with_predecessors_cap_is_on_unique_count_end_to_end` through `Executor::delegate_request` (65-unique rejected with nothing registered; 120-raw/3-unique accepted; past-sanity-bound rejected).
- Round 2 on `b6d37e5`: 1 warning (boundary-value coverage: empty predecessor list and exactly-at-cap untested) + 1 info (`# Errors` docs missing on the six new pub redb methods). Both fixed in `df43de0`: at-cap 64-unique accepted, empty list accepted as plain registration (behavior confirmed before pinning, not changed), raw-list-at-exactly-1024 accepted, and `# Errors` sections added. That delta is purely additive tests + doc comments (verified: zero deletions, redb additions 100% doc comments, runtime.rs hunk inside the test module).

Each rule-review delta was directly re-verified hunk-by-hunk before the next run.

### Trust-model iterations (each a full re-review)

1. **Consent-primary** → rejected: successor-side consent can't authorize reading a predecessor's secrets.
2. **Append-origin gate** → rejected (self-authorizing, see P1 above).
3. **Final: first-writer-wins immutable origin record**, written durably BEFORE registration; origin-write failure fails the entire registration; no grandfather clause (recordless predecessor → `NoProvenance`, refuse); token-less (`None`) origin never privileged; only `Some(A) == Some(A)` copies; every lookup fail-closed.

### Seal discipline (4 fixes through review)

Seal only on a fully-clean copy: enumeration succeeded, every read/import succeeded, enumeration-registry copy succeeded, shared marker write-verified BEFORE the redb audit row. A collision counts as present only if it decrypts under the successor DEK; an absent predecessor writes nothing (retryable), while empty-but-held seals `NoData`.

### Smoke tests (branch binaries)

- Binary identity: `freenet`/`fdev` built from the branch, `strings` confirms freenet-stdlib 0.8.4.
- Old-client wire compat: riverctl 0.1.77 (stdlib 0.8.3) against an isolated branch node — clean PUT + GET, no deserialization errors. New variant is at bincode tag 3 (appended, not a new field), so old clients are unaffected until they opt in.

### Dismissed findings (with justification)

- *redb `.unwrap()` after length check* (early Claude pass): replaced with individually-keyed rows in the reserved-marker redesign; the flagged code no longer exists.
- *"Cap truncation not in the report"*: superseded — over-cap now rejects whole, so there is no truncation to report.
- *Concurrent first-writer regression test*: redb serializes writers in a single transaction (lookup + insert atomic); the insert-if-absent path is covered by unit tests, and a true multi-process race is not constructible in-process. Accepted as verified-by-construction.

### Deploy ordering

stdlib 0.8.4 is published and freenet-core is on it. Per the deploy-ordering law, no client emits tag 3 until a freenet-core release containing this PR is live on the network. The client-side (freenet-migrate crate) emission lands in a follow-up after that release.

[AI-assisted - Claude]

Issue #4912, opened 2026-07-22
nova gateway (0.2.104): sustained broadcast-queue-full churn, 100%+ CPU, client PUT pipeline delayed ~2min, webapp contract 503


sanity:
Observed 2026-07-22 ~15:30 UTC while publishing River UI through the nova gateway (5.9.111.215:31337, HTTP 7509), running 0.2.104 (cd5523d76e94, built 2026-07-21T19:15:50Z, up ~20h since auto-update).

## Symptoms

1. **Sustained broadcast-queue-full churn**: `Broadcast queue full, evicted oldest entry ... queue_depth=256` at **300-700 events/minute continuously for at least the whole 15:00 UTC hour**, spread across many contracts (top offenders in that hour: `FX2jB61N…` 3205, `HRnUWLpQ…` 2086, `82ZZ1G7H…` 2017, `FujmqVsU…` 1689, `HQk7zGRQ…` 1311). Log volume ~10-18 MB/hour all day.
2. **CPU saturation**: the freenet process at ~103% CPU.
3. **Client PUT pipeline delayed ~2 minutes**: a `fdev publish` PUT for the River webapp contract (`raAqMhMG…`, tx `01KY57BFRD5HT36XSD2PY54N01`) was registered at 15:33:20, but `PUT relay: processing Request` only fired at **15:35:20** — 2 minutes later — and the attempt then timed out at 15:35:50 (`outcome="timeout" timeout_secs=117`). The client got "put timed out after 1 peer attempt(s)". Notably the PUT **actually landed on-network anyway** (the network moved to the new state version; a later attempt got "New state version X must be higher than current version X") — so the gateway reported a false-negative failure to the client while the operation succeeded.
4. **Gateway doesn't hold/serve the River webapp contract**: `GET /v1/contract/web/raAqMhMG…/` returns 503 "Loading contract…" persistently, while the isolated hosted node on the same machine (try.freenet.org, also current main-ish) serves it fine at HTTP 200. Repeated requests over ~15 min never resolved to the app.

`contract_handling` error log concurrently shows repeated `Failed to send queue-full response (client may have disconnected) error=no response received from handler`.

## Why this seems worth a look

- The per-(contract,peer) keep-latest broadcast queue is supposed to absorb fan-out pressure by coalescing; queue_depth=256 staying pegged with constant evictions across *many* contracts for hours looks like the drain side (send workers / cwnd) can't keep up, or something upstream is generating update work faster than coalescing can absorb.
[...]

PR #4916, opened 2026-07-24
fix(operations): report which timeout condition actually fired

sanity:
## Problem

`drive_retry_loop` collapses three distinct attempt-timeout conditions into a bare `Err(())`, so the shared `outcome="timeout"` warning logs `RetryDriver::attempt_timeout` unconditionally:

```rust
Err(_) => {
    tracing::warn!(..., timeout_secs = attempt_timeout.as_secs(), ...);
```

But `attempt_timeout` only governs the **non-streaming** path. A streaming PUT is bounded by `STREAM_OP_INACTIVITY_TIMEOUT` (30s) or `STREAMING_ATTEMPT_TIMEOUT_CAP` (600s) instead, so for those the logged number is not the budget that fired.

This is not hypothetical. It sent the investigation of #4912 after the wrong mechanism. Two real PUTs on the nova gateway:

| tx | logged | attempt actually ran |
|---|---|---|
| `01KY57BFRD5HT36XSD2PY54N01` (in the issue) | `timeout_secs=117` | 30.001s |
| `01KY82FMT0QWMXG7Y2V19DGER1` (post-0.2.105) | `timeout_secs=96` | 30.5s |

Both were abandoned by the 30s stream-inactivity check while reporting a ~2-minute budget. Measurement basis, since it matters: the attempt span is `PUT relay: processing Request` to `attempt timed out`. The first transaction's client-visible span was ~150s, but the extra ~120s was contract-handler queue starvation *before* the attempt began, so 117 described neither number.

## Approach

Replace the untyped `Err(())` with a `TimeoutCause` enum naming the condition (`Deadline` / `StreamStall` / `StreamCeiling`). The warning now carries:

- `timeout_kind` - which condition fired
- `timeout_secs` - the configured bound for that condition
- `elapsed_secs` - how long the attempt actually ran

`StreamStall` carries the measured silence rather than echoing the nominal window. An atomic-tiebreak `continue` restarts a full window from the current instant rather than from the last fragment, so a confirmed stall is declared somewhere in `[window, 2*window)`. Reporting a flat 30 would understate the wait by up to 2x, which is a smaller instance of the same misreport.

Purely diagnostic: no control flow, retry, or timeout behavior changes. All three variants still advance to the next peer exactly as before, and the arm order in the match is byte-identical to main.

## Testing

- `attempt_timeout_budget_matches_the_condition_that_fired` - each variant maps to its real bound; asserts `StreamStall`'s budget is *not* `attempt_timeout` (the exact misreport), that `observed` diverges from `budget` for a stall, and pins the three label string values, which are an operator-facing interface for log filters.
- `streaming_stall_and_ceiling_return_distinct_causes` - arm-aware source pin: checks the ceiling arm and the inactivity arm *separately*, so swapping the two variants cannot pass.
- `timeout_warning_reports_condition_specific_budget` - source pin that the warning derives its fields from the cause, never a bare `attempt_timeout.as_secs()`. Anchored on the warning's message literal rather than a byte count, so it cannot panic on a non-char boundary.
- Both existing streaming tests now assert the CAUSE rather than `is_err()`.

Verified by mutation, not just by passing: reverting the warning line to its pre-fix form fails the pin, and swapping the two streaming variants between arms fails **three** tests.

`cargo test -p freenet --lib`: 4259 passed. The one failure, `test_set_own_addr_atomic_with_network_status`, is a pre-existing flake unrelated to this diff (it races unlocked `set_own_addr` test callers); diagnosed and filed as #4918. `cargo clippy`: clean.

## Scope

Refs #4912 but does not close it. #4912's four symptoms were storm fallout resolved by 0.2.105; the validation is posted [on the issue](https://github.com/freenet/freenet-core/issues/4912#issuecomment-5065527228) and it is now closed. This fixes the diagnostic defect that made it hard to read.

Related follow-ups found during this work: #4917 (fair-queue has no depth instrumentation), #4918 (the flaky test above).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/op_ctx.rs
+++ b/crates/core/src/operations/op_ctx.rs
@@ -2409,8 +2409,15 @@ mod tests {
         let warn_pos = body
             .find("outcome = \"timeout\"")
             .expect("drive_retry_loop must emit the outcome=timeout warning");
-        // Bound the window to the warning's own field list.
-        let window = &body[warn_pos..warn_pos + 400];
+        // Bound the window to the warning's own field list by anchoring on its
+        // closing message rather than a byte count. A fixed-width slice would
+        // panic if it ever landed mid-character — this file contains multibyte
+        // punctuation — and would silently widen or narrow as fields are added.
+        let msg_anchor = "\"{op_label}: attempt timed out; advancing\"";
+        let warn_len = body[warn_pos..]
+            .find(msg_anchor)
+            .expect("the timeout warning must retain its message literal");
+        let window = &body[warn_pos..warn_pos + warn_len];
 
         assert!(
             window.contains("timeout_kind = cause.label()"),
--- a/crates/core/src/operations/op_ctx.rs
+++ b/crates/core/src/operations/op_ctx.rs
@@ -526,14 +526,7 @@ pub(crate) enum TimeoutCause {
     Deadline,
     /// Streaming PUT: no fragment arrived for a whole
     /// [`STREAM_OP_INACTIVITY_TIMEOUT`] window (a confirmed stall).
-    ///
-    /// Carries the silence actually observed. This is NOT always the nominal
-    /// window: an atomic-tiebreak `continue` restarts a full window from the
-    /// current instant rather than from the last fragment, so a confirmed
-    /// stall is declared somewhere in `[window, 2 * window)`. Logging the flat
-    /// nominal value would understate the wait by up to 2x — a smaller version
-    /// of the very misreport this enum exists to fix.
-    StreamStall { silent_for: std::time::Duration },
+    StreamStall,
     /// Streaming PUT: the hard [`STREAMING_ATTEMPT_TIMEOUT_CAP`] ceiling was
     /// reached while fragments were still dribbling in.
     StreamCeiling,
@@ -546,41 +539,28 @@ impl TimeoutCause {
     /// queries). Treat a rename as a breaking change; the values are pinned by
     /// `attempt_timeout_budget_matches_the_condition_that_fired`.
     pub(crate) fn label(self) -> &'static str {
+        // Exhaustive on purpose: a catch-all would silently give a future
+        // variant the wrong label, which is the class of bug this whole type
+        // exists to prevent (and the codebase bans wildcard arms in
+        // classification code for exactly that reason).
         match self {
             Self::Deadline => "attempt_deadline",
-            Self::StreamStall { .. } => "stream_stall",
+            Self::StreamStall => "stream_stall",
             Self::StreamCeiling => "stream_ceiling",
         }
     }
 
     /// The configured bound that governed this attempt.
     ///
-    /// Only [`Self::Deadline`] is bounded by the driver's `attempt_timeout`;
-    /// the two streaming variants are bounded by their own constants. Logging
+    /// This is the RULE that fired, not a measurement: how long the attempt
+    /// actually ran is logged separately as `elapsed_secs`. Only
+    /// [`Self::Deadline`] is bounded by the driver's `attempt_timeout`; the two
+    /// streaming variants are bounded by their own constants. Logging
     /// `attempt_timeout` for those is what made #4912 hard to read.
     fn budget(self, attempt_timeout: std::time::Duration) -> std::time::Duration {
         match self {
             Self::Deadline => attempt_timeout,
-            Self::StreamStall { .. } => {
-                crate::operations::stream_progress::STREAM_OP_INACTIVITY_TIMEOUT
-            }
-            Self::StreamCeiling => crate::operations::STREAMING_ATTEMPT_TIMEOUT_CAP,
-        }
-    }
-
-    /// How long the attempt actually ran before being abandoned.
-    ///
-    /// `Deadline` and `StreamCeiling` fire exactly at their bound, so observed
-    /// equals budget. A stall is the one case where they diverge (see
-    /// [`Self::StreamStall`]), which is why it carries a measured value.
-    fn observed(self, attempt_timeout: std::time::Duration) -> std::time::Duration {
-        // Exhaustive on purpose: a catch-all would silently give a future
-        // variant the wrong observed value, which is the class of bug this
-        // whole type exists to prevent (and the codebase bans wildcard arms
-        // in classification code for exactly that reason).
-        match self {
-            Self::StreamStall { silent_for } => silent_for,
-            Self::Deadline => attempt_timeout,
+            Self::StreamStall => crate::operations::stream_progress::STREAM_OP_INACTIVITY_TIMEOUT,
             Self::StreamCeiling => crate::operations::STREAMING_ATTEMPT_TIMEOUT_CAP,
         }
     }
@@ -652,11 +632,7 @@ async fn await_streaming_attempt(
                 if elapsed < STREAM_OP_INACTIVITY_TIMEOUT {
                     continue;
                 }
-                // Report the silence we actually measured, not the nominal
-                // window: a tiebreak `continue` restarts a full window from
-                // now rather than from the last fragment, so `elapsed` can be
-                // up to 2x the window by the time a stall is confirmed.
-                return Err(TimeoutCause::StreamStall { silent_for: elapsed });
+                return Err(TimeoutCause::StreamStall);
             }
         }
     }
@@ -697,6 +673,15 @@ pub(crate) async fn drive_retry_loop<D: RetryDriver>(
         let request = driver.build_request(attempt_tx);
 
         let attempt_timeout = driver.attempt_timeout();
+        // Measure how long the attempt really runs, so the log carries an
+        // observation and not just the rule that fired. Deriving elapsed from
+        // the cause would repeat #4912's mistake in miniature: a stall's
+        // inactivity window is not the attempt's duration (a stream can make
+        // progress for 40 s and then stall for 30 s, a 70 s attempt).
+        // `tokio::time::Instant` is auto-paused under `start_paused(true)`, so
+        // simulation tests see virtual elapsed time — same reasoning as
+        // `subscribe/op_ctx_task.rs`'s `request_sent_at`.
+        let attempt_started = tokio::time::Instant::now();
         let mut ctx = op_manager.op_ctx(attempt_tx);
 
         // `round_trip: Result<Result<NetMessage, OpError>, TimeoutCause>` —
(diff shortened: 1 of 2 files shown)


sanity:
## Review round 1: all findings addressed

Ran the Full tier (this touches `operations/`, an always-Full surface): external Codex pass plus four Claude lenses (code-first, testing, skeptical, big-picture). Codex found nothing. The four lenses found no blocking *behavioral* bug and independently confirmed the arm order is byte-identical to main, `budget()` matches the real constants, and nothing leaks into a public or wire surface. Findings and disposition:

### Fixed

| # | Finding | What changed |
|---|---|---|
| 1 | **Swapping `StreamStall`/`StreamCeiling` between arms passed every test.** Both behavioral tests asserted only `is_err()`, and the source pin checked that both names appeared *somewhere* in the function. Raised independently by two reviewers, and it is exactly this PR's own bug class. | Both streaming tests now assert the cause; the pin is arm-aware. Verified by actually swapping the variants: **3 tests fail**. |
| 2 | **The measured stall time was computed and thrown away.** `handle.since_last()` is already in hand, and a tiebreak `continue` restarts a full window from now rather than from the last fragment, so true silence is in `[30,60)s`. Reporting a flat 30 understates by up to 2x. | `StreamStall` now carries `silent_for`; the warning gained `elapsed_secs`. This also retires the PR's original "we can't log elapsed without threading a clock" rationale, which was wrong. |
| 3 | **Doc/PR claimed "~30s in" and "4x" against the issue's ~150s span.** | Stated the measurement basis explicitly (attempt span vs client-visible span, the difference being pre-attempt queue starvation). Corrected in both the doc comment and the PR body. |
| 4 | **`timeout_kind` missing from the subscribe timeout**, so filtering on it silently drops subscribe/renewal timeouts from the `outcome="timeout"` stream. | Added, using the shared `label()` rather than a duplicated literal so the two cannot drift. |
| 5 | **Label values untested** (only uniqueness). They are an operator-facing interface. | Pinned all three string values. |
| 6 | **`AttemptTimeout` collided with the `attempt_timeout()` method**, forcing two meanings of "attempt timeout" on the reader. | Renamed to `TimeoutCause`. |
| 7 | **Brittle 400-byte pin window**: could panic on a non-char boundary (this file has multibyte punctuation) and silently rescope as fields are added. | Anchored on the warning's message literal. Fixed in `7dc22b4b`, before the reviews landed. |
| 8 | **Brittle end anchor** on a doc-comment line; **one stale `Err(())` doc reference** missed. | Anchored on the `fn` signature; doc updated. |

Also caught by review: my `observed()` used a catch-all arm, which trips `clippy::wildcard_enum_match_arm` and violates the codebase's no-catch-all rule for classification code. Now exhaustive, so a future variant fails to compile until handled.

### Not changed, with reasons

- **A behavioral test asserting the emitted log record** (via `install_log_capture`, precedent at `contract/executor/pool_tests/subscriber_limit_tests.rs:282`). Fair suggestion, and I agree it would be stronger in principle. But it needs `drive_retry_loop`, which needs a full `OpManager`; this file's own tests say that cost is why the source-pin convention exists here. With finding 1 fixed, variant attribution is now covered *behaviorally* at `await_streaming_attempt`, and the pin covers the warning's use of the cause. That combination gets most of the value. Worth revisiting if a cheap harness appears.
- **Should `StreamCeiling` terminate the op at all**, given the stream was still progressing? Genuine design question, plausibly feeding #3465. Out of scope for a diagnostic-only change, and note `MAX_PEER_ADVANCEMENTS_STREAMING = 0` means streaming PUT gets exactly one attempt today regardless. The enum is the right foundation if someone wants to differentiate later.
- **`timeout_kind` on the two relay-downstream sites** (`subscribe:2098`, `get:3635`). Those don't carry `outcome="timeout"`; they're a different event shape, so adding the field would imply membership in a stream they aren't part of.
- **`pub(crate)` on `TimeoutCause`.** Was flagged as broader than needed when it had no cross-module use; finding 4 gave it one.

Re-ran Codex against the updated content, since a changed diff is an unreviewed diff.

[AI-assisted - Claude]

sanity:
## Review round 2: Codex caught the field I added in round 1

Re-ran the external pass after the round-1 changes (a changed diff is an unreviewed diff). Codex found a real problem with `elapsed_secs`, which I had *added* in round 1 to address the "measured stall time is thrown away" finding:

> When a stream makes progress for 40 seconds and then stalls for 30 seconds, this returns only `silent_for`, so `elapsed_secs` reports 30 even though the attempt ran about 70 seconds.

Correct, and worth being blunt about: that is the same defect this PR exists to fix, reproduced one level down. I had named a field `elapsed_secs` and populated it with something that is not elapsed time. My own test data shows it: `streaming_put_retries_on_true_stall` asserts the attempt runs 60-120s while the carried silence is ~30s.

### Fix

Measure the attempt instead of deriving it. `tokio::time::Instant::now()` at attempt start, `.elapsed()` at the timeout, following the documented precedent at `subscribe/op_ctx_task.rs:866` (auto-paused under `start_paused(true)`, so simulation tests see virtual time; `drive_retry_loop` already uses tokio's clock for the non-streaming deadline).

The three fields now mean exactly what they say:

| field | meaning |
|---|---|
| `timeout_kind` | which rule fired |
| `timeout_secs` | the configured bound for that rule |
| `elapsed_secs` | how long the attempt actually ran (measured) |

This also let `StreamStall` go back to a unit variant and removed the `observed()` helper entirely, since a real clock covers what the carried value was approximating. Net simplification: the round-1 payload was the wrong shape for the problem.

Codex's alternative suggestion was to rename the field rather than measure. Measuring is better here: the true attempt duration is the number that would have made #4912 readable at a glance ("logged 117, ran 30"), and a renamed condition-specific duration would not have provided it.

### Mutation-verified

Both guards re-checked against the final content by deliberately breaking them:

- swap `StreamStall`/`StreamCeiling` between arms → **3 tests fail**
- derive `elapsed_secs` from the cause again → **1 test fails**

`cargo test -p freenet --lib`: 4260 passed, 0 failed. `cargo clippy`: clean.

Codex re-run against this final content is in flight; will report.

[AI-assisted - Claude]

sanity:
## Review round 3: elapsed was measuring cleanup too

Codex again, on the round-2 content:

> `release_pending_op_slot(attempt_tx).await` above can block for up to `NOTIFICATION_SEND_TIMEOUT` (30 seconds), so evaluating `attempt_started.elapsed()` here includes cleanup delay and can substantially overstate the attempt duration.

Verified and correct. `OpManager::NOTIFICATION_SEND_TIMEOUT` is 30s (`node/op_state_manager.rs:783`) and `release_pending_op_slot` awaits on it (`:1503-1510`), and that call sits between the attempt resolving and the warning. So under notification backpressure `elapsed_secs` could be inflated by up to 30s, which is exactly the size of the stall window the field exists to reveal. It would have been a plausible-looking wrong number, the worst kind for this particular PR.

Fix: stop the clock the moment the attempt resolves, before the cleanup await, and log that value.

Pinned by ordering rather than by string alone: the test now asserts the capture appears textually before the `release_pending_op_slot` await. Moving it after fails the test (verified by doing so).

`cargo test -p freenet --lib`: 4260 passed, 0 failed. `cargo clippy`: clean.

---

### Round tally

Three external rounds, each finding something real, and rounds 2 and 3 both found defects in the *fix for the previous round* rather than in the original change:

1. Variant swap undetectable by tests; measured stall time discarded; name collision; brittle pins. (Claude lenses + Codex)
2. `elapsed_secs` did not contain elapsed time. (Codex)
3. `elapsed_secs` included cleanup backpressure. (Codex)

That pattern is worth noting on a PR whose whole subject is "the log reported a number that wasn't the number that mattered." Each round the fix was still reporting a not-quite-honest duration. Current state measures a real interval, bounded to the attempt itself, with all four properties pinned against regression:

- swapping the streaming variants fails 3 tests
- deriving `elapsed_secs` from the cause fails 1
- moving the capture after cleanup fails 1
- reverting the warning to `attempt_timeout.as_secs()` fails 1

Codex re-run against this content is in flight.

[AI-assisted - Claude]

sanity:
## Review round 4: `elapsed_secs` withdrawn

Codex, third finding on the same field:

> When `stream_progress()` uses `VirtualTime`, the stall or ceiling advances through that injected `TimeSource`, while `tokio::time::Instant` does not advance with it. The timeout warning can therefore report `elapsed_secs=0` for a 30- or 600-second virtual attempt.

Verified: `VirtualTime::now_nanos()` reads a scheduler-advanced atomic (`simulation/time.rs:489`), entirely decoupled from tokio's clock, and `stream_progress.rs`'s own module docs say as much ("`tokio::time::sleep`/`timeout` would not advance under VirtualTime"). Under a non-paused runtime it would also leak wall-clock into simulation traces, which is a determinism-test hazard.

### Decision: remove the field

I could have threaded a clock through, but doing it correctly needs a clock matching the path being timed, for **both** branches. The streaming path's `TimeSource` is reachable only via `StreamProgress`'s `sleep` closure (no `now()`), and the non-streaming branch has no `TimeSource` at all. `OpManager` holds a `DynTimeSource` but exposes no getter. That is a design change, not a logging tweak.

Weighed against that: `elapsed_secs` produced **three** consecutive review findings, each a different way of being subtly wrong (derived-not-measured, then contaminated by cleanup backpressure, now wrong clock). On a log line whose entire purpose is trustworthy timing, a plausible-but-wrong duration is worse than no duration. So it comes out, and the work is tracked properly in #4919.

The #4912 fix stands without it. `timeout_kind=stream_stall timeout_secs=30` no longer misattributes a 30s stall to a 117s deadline, which was the whole complaint.

A pin asserts `elapsed_secs` cannot be reintroduced from a wall clock without failing a test.

### Final state

The diff is now smaller than at round 1 and does one thing:

```
outcome = "timeout",
timeout_kind = cause.label(),                          // which rule fired
timeout_secs = cause.budget(attempt_timeout).as_secs() // that rule's real bound
```

Pinned by mutation, each verified to fail:
- swap `StreamStall`/`StreamCeiling` between arms → 3 tests
- revert to `timeout_secs = attempt_timeout.as_secs()` → 1 test
- reintroduce an `elapsed_secs` field → 1 test

`cargo test -p freenet --lib`: 4260 passed, 0 failed. `cargo clippy`: clean.

### On the review process here

Four external rounds, each finding something real, and three of them found defects in the *fix for the previous round* rather than in the original change. Worth recording on a PR about a log that reported the wrong number: my own attempts to add a "correct" duration were wrong three times in three different ways. The change that survived is the one that only claims what it can actually substantiate.

Follow-ups opened along the way: #4917 (fair-queue depth instrumentation), #4918 (pre-existing flaky test), #4919 (measured attempt duration, done properly).

[AI-assisted - Claude]

PR #4920, opened 2026-07-24
fix(test): stop unlocked test writes to the external_address global

### LINKED ISSUE #4918: Flaky: test_set_own_addr_atomic_with_network_status races unlocked set_own_addr callers under cargo test
## Symptom

`ring::connection_manager::tests::test_set_own_addr_atomic_with_network_status` fails intermittently under `cargo test -p freenet --lib`. Measured on a clean tree: **1 failure in 3 full-suite runs**, and **0 failures in 11 isolated runs** (6 default, 5 with `--test-threads=1`). So it only fails when run alongside the rest of the suite, which points at cross-test interference rather than anything in the test's own logic.

Not caused by any recent change: it reproduces on a branch whose only diff is `crates/core/src/operations/op_ctx.rs`, which never references `network_status`, `set_own_addr`, or `external_address`.

## Diagnosis

The test writes and asserts on the process-global `network_status::external_address`. It is careful about this and documents a "GLOBAL-STATE CONTRACT" (`crates/core/src/ring/connection_manager.rs:3638-3654`), acquiring the crate-wide `TEST_GLOBAL_STATE_LOCK` for its whole body. Its comment claims:

> `connection_manager`'s `test_try_set_own_addr_*` tests and every test in `node::network_status` also touch this global. They are all serialized by the crate-wide `TEST_GLOBAL_STATE_LOCK`.

That inventory is incomplete. `ConnectionManager::set_own_addr` writes the global (`connection_manager.rs:1640` calls `network_status::set_external_address`), and there are test call sites of `set_own_addr` **outside** those two modules that never take the lock:

- `crates/core/src/node.rs:4487` (in `resync_request_for_bogus_keys_does_not_consume_limiter_slots`)
- `crates/core/src/node.rs:4592`
- `crates/core/src/node.rs:4735`
- `crates/core/src/node.rs:4864`
- `crates/core/src/ring.rs:8500`

`grep -rn "TEST_GLOBAL_STATE_LOCK" crates/core/src/` returns hits in only two files (`node/network_status.rs`, `ring/connection_manager.rs`), confirming none of the five sites above participate.

Under `cargo test` the whole lib-test binary is one process with tests running concurrently, so any of those five can land a write between the atomic test's write and its exact-equality read. The test runs 4,000 rounds x 8 threads, so it has a wide window to get unlucky. CI uses `cargo nextest` (per-process isolation), which masks this, which is why it shows up locally rather than in CI.

## Suggested fix

Either:

1. Have the five unlocked sites acquire `TEST_GLOBAL_STATE_LOCK` for the duration of their use, matching the documented contract; or
2. Better, remove the ability to get this wrong: make the global write opt-in for tests, or give these tests a `ConnectionManager` constructor that does not publish to the process-global address at all (none of the five appear to care about the global; they just need a local address set).

Option 2 is preferable because option 1 leaves a hazard that any new `set_own_addr` test call site re-introduces. If option 1 is taken, the contract comment at `connection_manager.rs:3646` should be corrected, since it currently asserts an invariant the code does not hold, and a future reader will trust it.

## Repro

```bash
# fails roughly 1 in 3
for i in 1 2 3 4 5 6; do cargo test -p freenet --lib 2>&1 | grep -E "^test result"; done

# passes consistently in isolation
cargo test -p freenet --lib ring::connection_manager::tests::test_set_own_addr_atomic_with_network_status
```

Found while running the pre-commit suite for #4916. Not fixing it there to keep that PR's diff scoped to its own subject; happy to take this separately.

[AI-assisted - Claude]

sanity:
## Problem

`ring::connection_manager::tests::test_set_own_addr_atomic_with_network_status` failed roughly **1 run in 3** under `cargo test -p freenet --lib`, and **0 times in 11 isolated runs** — a cross-test interference flake, not a bug in the test's own logic.

`ConnectionManager::set_own_addr` mirrors the address into the process-global `network_status::external_address`. `own_addr` is per-manager but that global is process-wide, so the atomic test's `own_addr == external_address` assertion only holds while no *other* manager writes the global. `NETWORK_STATUS` is a `OnceLock`, so once any test calls `init()` every later `set_own_addr` in the binary hits the live tracker.

Seven test call sites were writing it without holding `TEST_GLOBAL_STATE_LOCK`:

| File | Sites |
|---|---|
| `node.rs` | 4 resync tests |
| `ring.rs` | 1 storm test |
| `node/op_state_manager.rs` | `seed_location_and_one_connection` (a shared helper) + 1 test |

The test documents a "GLOBAL-STATE CONTRACT" and takes the lock, but its inventory of other writers was incomplete, so the contract read as satisfied when it wasn't.

## Approach

None of the seven reads `external_address` — they set an address only so routing knows it. The sole production reader of that global is the home-page dashboard card. So they now use a new `#[cfg(test)] set_own_addr_local_for_test`, which writes only the local field and never touches the global.

**Why not just take the lock at each site?** That was the obvious fix and I rejected it: the initial triage of this bug (in #4918) found only 5 of the 7 callers. If a careful pass misses two, the "remember to lock" contract will keep being broken by future callers. Removing the unnecessary global write is not a workaround — these tests genuinely have no business writing it — and it leaves nothing to remember.

## Testing

`test_no_test_calls_global_set_own_addr` scrapes the three files and fails if any calls the mirroring `set_own_addr`, with a message pointing at the lock requirement and this issue. It also asserts the safe helper *is* present, so an edit that silently drops address-setting can't pass vacuously.

**It is a source scrape on purpose.** The race only manifests under `cargo test`'s shared-process model; CI runs `cargo nextest`, which gives every test its own process and masks it completely. A behavioural guard therefore could never fail in CI, which is exactly why this flake survived. A scrape runs deterministically under both runners.

Verified two ways:
- **12 consecutive full-suite runs, 0 failures.** At the old ~1-in-3 rate the odds of that are under 1%.
- **Mutation:** reverting one caller to `set_own_addr` fails the guard.

Also corrected the atomic test's stale contract comment, which listed an incomplete set of writers.

Closes #4918

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -4484,7 +4484,7 @@ mod tests {
         op_manager
             .ring
             .connection_manager
-            .set_own_addr("127.0.0.1:14000".parse().unwrap());
+            .set_own_addr_local_for_test("127.0.0.1:14000".parse().unwrap());
 
         // Attach an EMPTY redb hosting store so `contract_state_present`
         // returns false for unknown keys (a fresh store creates the STATE
@@ -4589,7 +4589,7 @@ mod tests {
         op_manager
             .ring
             .connection_manager
-            .set_own_addr("127.0.0.1:14100".parse().unwrap());
+            .set_own_addr_local_for_test("127.0.0.1:14100".parse().unwrap());
 
         // The one contract we DO hold.
         let key = freenet_stdlib::prelude::ContractKey::from_id_and_code(
@@ -4732,7 +4732,7 @@ mod tests {
         op_manager
             .ring
             .connection_manager
-            .set_own_addr("127.0.0.1:14200".parse().unwrap());
+            .set_own_addr_local_for_test("127.0.0.1:14200".parse().unwrap());
 
         // Stand-in contract handler: flips `update_query_seen` to true if it EVER
         // receives an UpdateQuery (proof the WASM-apply path was entered),
@@ -4861,7 +4861,7 @@ mod tests {
         op_manager
             .ring
             .connection_manager
-            .set_own_addr("127.0.0.1:14300".parse().unwrap());
+            .set_own_addr_local_for_test("127.0.0.1:14300".parse().unwrap());
 
         // Stand-in contract handler: answers UpdateQuery with a SUCCESSFUL,
         // state-changed merge (mirrors the ResyncResponse arm's
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -2953,7 +2953,10 @@ mod tests {
     /// anyone's computed upstream.
     fn seed_location_and_one_connection(op_manager: &OpManager) {
         let self_addr: std::net::SocketAddr = "127.0.0.1:12000".parse().unwrap();
-        op_manager.ring.connection_manager.set_own_addr(self_addr);
+        op_manager
+            .ring
+            .connection_manager
+            .set_own_addr_local_for_test(self_addr);
         assert!(
             op_manager
                 .ring
@@ -3032,7 +3035,10 @@ mod tests {
         // Own address. The UPDATE sender is a DIFFERENT peer, so the self/sender
         // echo-skip never applies to A or B below.
         let self_addr: std::net::SocketAddr = "127.0.0.1:12000".parse().unwrap();
-        op_manager.ring.connection_manager.set_own_addr(self_addr);
+        op_manager
+            .ring
+            .connection_manager
+            .set_own_addr_local_for_test(self_addr);
         let sender_addr: std::net::SocketAddr = "127.0.0.1:20000".parse().unwrap();
 
         let key = reroot_contract_key(0x5A);
--- a/crates/core/src/operations/connect/op_ctx_task.rs
+++ b/crates/core/src/operations/connect/op_ctx_task.rs
@@ -531,6 +531,9 @@ async fn drive_client_connect_inner(
                 // diagnostics) and `own_location` (for routing).
                 // Idempotent: subsequent observations overwrite.
                 let location = Location::from_address(&address);
+                // GLOBAL-ADDR-WRITE-OK: production. A real node learning its
+                // peer-observed address is exactly what the external_address
+                // mirror is for (#4918).
                 op_manager.ring.connection_manager.set_own_addr(address);
                 op_manager
                     .ring
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -3750,7 +3750,10 @@ mod tests {
         );
         op_manager.ring.attach_op_manager(&op_manager);
         let self_addr: SocketAddr = "127.0.0.1:12000".parse().unwrap();
-        op_manager.ring.connection_manager.set_own_addr(self_addr);
+        op_manager
+            .ring
+            .connection_manager
+            .set_own_addr_local_for_test(self_addr);
 
         let handler = tokio::spawn(async move {
             while let Ok((id, ev, _priority)) = ch_channel.recv_from_sender().await {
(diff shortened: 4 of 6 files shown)


sanity:
## Review round 1: found two more unlocked writers, and a hole in my own guard

Codex, P1:

> `build_queue_full_test_node_with_clock` and `build_broadcast_test_node` still call the mirroring setter without `TEST_GLOBAL_STATE_LOCK` at `operations/update/op_ctx_task.rs:3753` and `:4022`. Because this guard scans only these three files, it passes while those calls can overwrite `external_address` during the atomicity assertion, leaving the flake unresolved.

Verified and correct. I had classified those two as production code; both are inside `#[cfg(test)]` modules (at 3553 and 3941) and are test node builders. `connect/op_ctx_task.rs:534` is the only genuine production caller.

So the caller inventory has now been wrong **three times running**: 5 sites in the original issue triage, 7 when I started the fix, 9 after this review. That is exactly the fragility I cited as the reason not to just add the lock everywhere — and I then reproduced it in the guard itself, which named three files and therefore could not see the two it did not name.

### Fix

Both sites migrated to the local-only setter, and the guard rewritten to **walk the entire crate source** with an allowlist of the two legitimate global writers (production connect, and this module's own atomicity test which holds the lock). It now fails closed on a new caller anywhere, rather than only where someone remembered to look.

Verified by mutation: reintroducing a bare `set_own_addr` in `update/op_ctx_task.rs` — the file the old guard could not see — now fails the guard, naming the file.

### State

- 7 consecutive full-suite runs clean after the additional migrations (12 before them, on the smaller change).
- `cargo fmt` clean. Clippy: the only errors locally are 10 pre-existing `let_underscore_must_use` lints in `client_events/websocket.rs`, which reproduce on pristine `origin/main` and are not in this diff — the known clippy-1.94 vs CI-1.93 pin gap.

Codex re-run against the updated content is in flight.

[AI-assisted - Claude]

sanity:
## Review round 2: the guard was still exempting whole files

Codex, P2:

> The allowlist operates at file granularity, so any future unlocked `.set_own_addr(...)` added to a test in either `connection_manager.rs` or `operations/connect/op_ctx_task.rs` is silently accepted. Both files contain test modules, making this a realistic recurrence of the exact race this guard claims to catch.

Correct, and it is the same mistake one level down: I replaced a hardcoded *file list* with a hardcoded *file allowlist*, which still trusts a human to notice that a test module lives inside an exempted file.

### Fix

Each call site is now checked individually. Any write to the global mirror must carry a `GLOBAL-ADDR-WRITE-OK` annotation with a reason, either on the call line or in the contiguous comment block directly above it. The walk-back stops at the first non-comment line, so a marker on an unrelated earlier statement cannot vouch for a later call.

The two legitimate writers are annotated in place:
- `connect/op_ctx_task.rs` — production; a node learning its peer-observed address is what the mirror is *for*.
- the atomicity test's own call — it holds `TEST_GLOBAL_STATE_LOCK` for its whole body.

Also: the search token is built at runtime (`format!(".set_own_addr{}", "(")`) so the guard's source does not contain the literal it searches for, which would otherwise match itself.

Verified by mutation: adding an unannotated caller **inside `connection_manager.rs`** — the file the previous version exempted wholesale — now fails the guard and names the file, line, and offending source line. 5 more full-suite runs clean.

### Round tally

Both review rounds found the same class of defect in successively narrower form: an inventory or exemption that depends on someone remembering to update it.

1. Two unlocked writers misclassified as production; guard scanned a fixed 3-file list.
2. Guard exempted whole files, including two containing test modules.

The current version has no list to keep current: it fails closed on every call site in the crate unless that site is explicitly justified.

Codex re-run against this content is in flight.

[AI-assisted - Claude]

sanity:
## Review round 3: clean, queued to merge

Codex on the final content:

> The patch isolates test-only address setup from the process-global network status while preserving production mirroring behavior. The source guard covers current call sites, and no functional regressions were identified.

No findings. CI green on `ff807233` (14 success, 2 skipped), merge state CLEAN. Queued.

### Summary of what changed across the rounds

The bug was one flaky test; every round found the same underlying shape — an inventory or exemption that depends on a human remembering to update it.

| Round | Found | Fix |
|---|---|---|
| initial triage | 5 unlocked writers | — |
| implementation | 7 (2 more found) | local-only setter |
| review 1 | 9 (2 misclassified as production) | migrate; guard scans whole crate |
| review 2 | guard exempted whole files, incl. two with test modules | per-call-site annotation |
| review 3 | clean | — |

Final state: 9 test call sites use a local-only setter that never touches the process-global mirror, and the two legitimate writers each carry a `GLOBAL-ADDR-WRITE-OK` annotation with a reason. The guard fails closed on any unannotated call site anywhere in the crate, so there is no list left to keep current.

Verification: 12 consecutive clean full-suite runs on the smaller change, 5 more after the extra migrations, against a ~1-in-3 pre-fix failure rate. Every guard property mutation-tested.

[AI-assisted - Claude]

Issue #4919, opened 2026-07-24
Log measured attempt duration alongside timeout_kind (needs a clock-correct source for both retry paths)


sanity:
## Context

#4916 made the `outcome="timeout"` warning in `drive_retry_loop` report **which** condition fired (`timeout_kind`) and **its configured bound** (`timeout_secs`), fixing the #4912 misattribution where a 30s stream stall was logged as a 117s per-attempt deadline.

What it deliberately does **not** add is the measured attempt duration. That would be genuinely useful: it turns "logged 117" into "logged bound 30, actually ran 70" at a glance, which is what made #4912 take log archaeology to unpick.

## Why it was dropped rather than shipped

It was attempted in #4916 and withdrawn after three successive review findings, each a different way of being subtly wrong:

1. Derived from the timeout cause, so a stream that made progress for 40s then stalled for 30s reported 30 for a 70s attempt.
2. Measured across `release_pending_op_slot(...).await`, which can block up to `NOTIFICATION_SEND_TIMEOUT` (30s) under notification backpressure, inflating the value by as much as the stall window it was meant to expose.
3. Measured with `tokio::time::Instant`, which does not track the injected `TimeSource`. The streaming path sleeps on that `TimeSource` (`VirtualTime` under DST, whose `now_nanos()` reads a scheduler-advanced atomic at `simulation/time.rs:489`), so a 30s *virtual* stall would log `elapsed_secs=0`. Under a non-paused runtime it would also leak wall-clock into simulation traces, which is a determinism-test hazard.

A plausible-but-wrong duration is worse than no duration on a log line whose entire purpose is trustworthy timing, so the field came out.

## What a correct implementation needs

A clock that matches the path being timed, for **both** branches of `drive_retry_loop`:

- **Streaming**: the driver's `TimeSource`, reachable today only through `StreamProgress`'s `sleep` closure. `StreamProgress` would need to expose `now()` (or an `elapsed_since`).
- **Non-streaming**: has no `TimeSource` at all; it uses `tokio::time::timeout` directly, which per `stream_progress.rs`'s own module docs is already not VirtualTime-correct.

So either `OpManager` grows a `TimeSource` accessor usable by the generic retry loop (it holds a `DynTimeSource` for the interest manager at `op_state_manager.rs:472` but exposes no getter), or `RetryDriver` gains a clock. Both are design changes rather than a logging tweak, which is why this is separate.

[...]

PR #4921, opened 2026-07-24
feat(contract): instrument fair-queue depth and admission outcomes

### LINKED ISSUE #4917: Observability: the contract-handler fair queue emits no depth or wait-time signal
## Problem

`crates/core/src/contract/fair_queue.rs` contains no `tracing::` call and no telemetry emission of any kind. The queue that gates **every** client operation reaching the WASM executor is invisible: you can see events being *rejected* (via `send_queue_full_response` in `contract.rs`), but never how deep the queue is, how long events wait in it, or which priority tier is being starved.

That means the queue is only observable once it is already saturated and shedding, which is exactly too late.

## Why this is worth fixing

This directly cost time diagnosing #4912. A client PUT sat 120.6 seconds between registration and relay processing on the nova gateway. Attributing that to fair-queue starvation required inferring from rejection *rates* over a whole hour rather than reading a depth anywhere, because no such signal exists. From the #4912 validation:

> There is no instrumented queue-depth on the client-dispatch path. No "queue depth=N" line exists to prove causation. The inference rests on the congestion being sustained network-wide for the whole hour at a rate proportionate to what fell inside the transaction's gap.

A depth or wait-time gauge would have made that a one-line answer.

## Suggested shape

Follow the existing pattern in `transport/shadow_demand.rs`: a relaxed-atomic gauge written on the hot path, sampled by a periodic aggregator that emits one rolled-up event per window. Explicitly **not** per-event logging, which would add noise to an error log that already runs several MB/hour.

Worth capturing per priority tier (`ClientLocal` / `NetworkRelay` / `Background`), since the interesting failure is a tier-specific starvation rather than a global total. Enqueue-to-dequeue wait time is arguably more useful than instantaneous depth, because it is the quantity a stalled client actually experiences.

Note `shadow_demand.rs` itself is transport-layer and carries a "never read from the production data path" rule, so the gauge belongs near the contract handler rather than being bolted onto that module.

## Not urgent

The saturation that motivated this is resolved (0.2.105 cost-aware eviction), so this is diagnostic hardening for the next occurrence, not a live problem.

[AI-assisted - Claude]

sanity:
## Problem

`crates/core/src/contract/fair_queue.rs` contained no `tracing::` call and no telemetry of any kind. Every client operation passes through this queue on its way to the single-threaded WASM executor, so a backlog here is what a user experiences as a slow or timed-out request — and it was completely invisible. Rejections could be inferred from their side effect (`send_queue_full_response`), but depth never could.

That cost real time on #4912: a client PUT sat 120.6s between registration and processing, and confirming the queue as the cause meant reasoning from rejection *rates* across a whole hour, because no depth signal existed to look at.

## What this adds

A gauge covering:

| Signal | Why |
|---|---|
| Depth, per priority tier | Is the executor backed up *now*, and in which lane |
| **High-water mark** (monotonic) | Was it *ever* backed up — the dashboard is polled, and a burst between polls leaves no trace in instantaneous depth |
| Rejections, split by cause | Global-capacity (node-wide backpressure) vs per-contract-cap (one noisy contract). Conflating them misdiagnoses a busy neighbour as a saturated node |
| Background shed | How much best-effort work is being dropped to protect client requests |

Surfaced two ways: a status-card row on the dashboard (hidden entirely while the queue has never been used, so an idle node stays uncluttered), and one WARN as high-water crosses 50% and 90% of capacity.

The WARN is **edge-triggered against the monotonic mark**, so each band logs at most once per process. This loop's log already runs to several MB/hour; a level-triggered or per-sample line would be pure noise. A single line saying the queue got deep is what #4912's logs could not produce.

## Design notes

- **Instance-scoped counters with a global mirror, not global-only.** Tests assert on `stats()`, which reads a single queue. A purely process-global counter would make every gauge assertion race other tests sharing the binary — that is exactly the #4918 flake I just fixed, and I did not want to reintroduce its shape here.
- **The snapshot reads the gauge at render time** rather than mirroring counters into `NetworkStatus`. `bug-prevention-patterns.md` records the rot mode for hand-maintained mirrors (#4009 / #4010) and prefers reading canonical state; this follows that.
- **Hot-path cost** is a few relaxed atomic stores per push/pop, the same shape as the transport layer's `shadow_demand` gauges. Kept in the contract layer rather than added to `shadow_demand`, which is transport-scoped and carries a "never read from the production data path" rule that does not apply here.
- **Depth only, no wait-time.** Queue latency would arguably be the more useful number, but measuring it needs a clock the contract loop does not have, and `crates/core/` bans wall-clock reads in favour of `TimeSource`. That is the same constraint #4919 is already tracking, so I did not want to half-solve it here.

## Testing

- `stats_track_depth_per_tier` — depth follows pushes and pops per tier, and high-water survives the queue draining (the property the whole metric exists for).
- `stats_distinguish_rejection_reasons` — a contract hitting its own cap must NOT register as global backpressure.
- `stats_count_global_capacity_rejections_and_background_sheds` — soft-cap rejection and shed accounting at capacity.
- `fair_queue_stats_hidden_when_queue_never_used` / `..._rendered_after_backlog_even_once_drained` — the dashboard row stays hidden when idle, and a drained burst still shows its peak.
- `every_total_queued_mutation_publishes` — anti-rot pin: every method changing `total_queued` must publish. Verified by mutation (removing the call from `pop` fails it).

`cargo test -p freenet --lib`: 4266 passed, 0 failed. `cargo clippy`: no new lints (the only local errors are the pre-existing `let_underscore_must_use` set in `client_events/websocket.rs`, which reproduces on pristine `main`).

Closes #4917

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -1801,6 +1801,8 @@ async fn push_with_background_eviction<CH>(
         match fair_queue.try_push(rejected.id, rejected.event, rejected.priority) {
             Ok(()) => return,
             Err(still_rejected) => {
+                // Terminal: eviction did not free enough room (#4917).
+                fair_queue.record_terminal_rejection(still_rejected.reason);
                 track_pending_reclamation_if_evict(contract_handler, &still_rejected);
                 send_queue_full_response(contract_handler.channel(), still_rejected).await;
                 return;
@@ -1808,6 +1810,9 @@ async fn push_with_background_eviction<CH>(
         }
     }
 
+    // Terminal: eviction could not help (wrong priority, wrong reason, or no
+    // Background queued), so this event is refused (#4917).
+    fair_queue.record_terminal_rejection(rejected.reason);
     track_pending_reclamation_if_evict(contract_handler, &rejected);
     send_queue_full_response(contract_handler.channel(), rejected).await;
 }
@@ -3603,6 +3608,56 @@ mod tests {
                 other => panic!("expected UpdateResponse for {label}, got {other}"),
             }
         }
+    }
+
+    /// Anti-rot pin for #4917: every TERMINAL admission rejection in
+    /// `push_with_background_eviction` must record itself.
+    ///
+    /// The counter deliberately does not live in `try_push`, because a
+    /// capacity failure there is often recovered by shedding Background and
+    /// retrying — counting it would report a client-visible reject for an
+    /// operation that succeeded. That correctness depends on the terminal arms
+    /// remembering to call `record_terminal_rejection`, so pin it: a new
+    /// terminal path that forgets would silently under-report queue-full
+    /// failures on the dashboard.
+    #[test]
+    fn terminal_rejection_paths_record_the_rejection() {
+        let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("src/contract.rs");
+        let source = std::fs::read_to_string(&path).expect("read own source");
+        let start = source
+            .find("async fn push_with_background_eviction")
+            .expect("push_with_background_eviction must exist");
+        let rest = &source[start..];
+        let end = rest
+            .find("\n/// ")
+            .map(|o| start + o)
+            .unwrap_or(source.len());
+        let body = &source[start..end];
+
+        let records = body.matches("record_terminal_rejection(").count();
+        assert_eq!(
+            records, 2,
+            "expected exactly two terminal rejection sites (eviction retry \
+             exhausted, and eviction-cannot-help), found {records}. If you \
+             added a terminal path, record the rejection there too; if you \
+             removed one, update this pin deliberately. See #4917."
+        );
+
+        // The shed loop also calls send_queue_full_response, for Background
+        // work that is re-emitted on its next cycle. That is a shed, not an
+        // admission rejection, and must NOT be counted as one.
+        let shed_loop = body
+            .find("for shed in evicted {")
+            .expect("the Background shed loop must exist");
+        let shed_end = body[shed_loop..]
+            .find('}')
+            .map(|o| shed_loop + o)
+            .expect("shed loop must close");
+        assert!(
+            !body[shed_loop..shed_end].contains("record_terminal_rejection"),
+            "a shed Background event is not an admission rejection; counting \
+             it would inflate the client-visible failure count (#4917)"
+        );
     }
 }
 
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -11,6 +11,10 @@ pub(crate) mod delegate_app_registry;
 mod executor;
 mod fair_queue;
 pub(crate) use fair_queue::Priority;
+/// Fair-queue occupancy for the status snapshot (#4917). Re-exported rather
+/// than widening `fair_queue` itself, which is otherwise private to this
+/// module.
+pub(crate) use fair_queue::{FairQueueStats, global_stats as fair_queue_stats};
 pub(crate) mod governance;
 mod handler;
 pub mod storages;
--- a/crates/core/src/contract/fair_queue.rs
+++ b/crates/core/src/contract/fair_queue.rs
@@ -385,6 +385,33 @@ impl FairEventQueue {
         }
     }
 
+    /// Count an admission failure that the caller could NOT recover from, i.e.
+    /// one that actually sent the client a queue-full response.
+    ///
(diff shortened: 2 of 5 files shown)


sanity:
## Review round 1: the dashboard reported zero for rejections that happened

Codex, P2:

> When one hot contract reaches `MAX_QUEUED_PER_CONTRACT` below the global cap, this row is rendered because `rejected_per_contract > 0`, but "Queue-full rejects" displays only `rejected_global_capacity`, so it reports zero and hides the actual rejection.

Correct, and a genuine bug in the rendering rather than a nit. The card's visibility condition and its contents disagreed: it appeared *because* rejections had happened, then displayed `0` for them. A card that shows up on an event and reports none of it is worse than not rendering at all — an operator would reasonably conclude the queue was fine.

It also contradicted this PR's own argument. I separated the two rejection causes precisely because node-wide backpressure and one contract hitting its cap call for different responses, then rendered only one of them.

### Fix

Per-contract rejections get their own tile. Deliberately **not** summed into the existing total — folding them together would restore exactly the misdiagnosis the split exists to prevent.

`fair_queue_per_contract_rejects_are_not_reported_as_zero` covers the specific scenario (per-contract rejections with global at zero). Verified by mutation: removing the tile fails it.

`cargo test -p freenet --lib`: 4267 passed, 0 failed. Clippy: no new lints.

Codex re-run against the updated content is in flight.

[AI-assisted - Claude]

sanity:
## Review round 2: the counter was counting recovered failures

Codex, P2:

> When a foreground push encounters the global cap while Background work is queued, `try_push` increments `rejected_global_capacity` before `push_with_background_eviction` sheds Background work and retries successfully. This tile therefore reports a "Queue-full reject" even though the foreground event was admitted and no queue-full response was sent.

Verified and correct. `push_with_background_eviction` owns a try → shed → retry sequence, and the retry usually succeeds. I was counting the first, recoverable failure, so the dashboard would report client-visible rejects for operations that completed fine — inflating exactly the number an operator reads that tile for.

Related to round 1's finding, and arguably the same underlying mistake: I put the accounting where the *event* was convenient rather than where the *meaning* is decided.

### Fix

`try_push` no longer counts. Only the caller knows whether a failure was terminal, so it says so via `record_terminal_rejection`, called from the two arms that actually send a queue-full response. The shed loop deliberately does not call it — a shed Background event re-emits on its next cycle and is already counted separately as `background_shed`.

Two tests:
- `recovered_capacity_hit_is_not_counted_as_a_rejection` — drives the full recovery path (fill to soft cap → foreground push fails → shed → retry succeeds) and asserts zero rejections but a non-zero shed count.
- `terminal_rejection_paths_record_the_rejection` — anti-rot pin on both terminal sites, plus a pin that the shed loop stays uncounted. Verified by mutation: dropping either record call fails it.

`cargo test -p freenet --lib`: 4269 passed, 0 failed. Clippy: no new lints.

Codex re-run in flight.

[AI-assisted - Claude]

sanity:
## Review round 3: clean, queued to merge

Codex on the final content:

> No actionable correctness issues were found in the fair-queue instrumentation, rejection accounting, snapshot integration, or dashboard rendering.

CI green on `f333efc8` (14 success, 2 skipped), merge state CLEAN.

### CI note

`macOS Service Unit` failed once on this commit and I re-ran it rather than treating it as broken code, after checking what actually failed:

```
error: failed to run custom build command for `zmij v1.0.21`
build-script-build: cannot execute binary file (exit status 126)
```

Exit 126 / "cannot execute binary file" on a *third-party* build script is a cached build-script binary from the wrong architecture, not a compile error. Supporting evidence: this diff touches five `.rs` files and no `Cargo.toml`, `Cargo.lock`, or `build.rs`; `zmij` is transitive; Linux Clippy/Fmt and Windows Check passed on the same commit; and the failure took 2m40s against the job's normal ~19m. The re-run passed in 18m55s.

### Both review rounds found the same mistake

Round 1: the card rendered because rejections had happened, then displayed `0` for them. Round 2: the counter incremented on capacity failures the caller went on to recover from, so admitted operations were reported as client-visible rejects.

Both were a case of putting the accounting where the *event* was convenient rather than where the *meaning* is decided — and both would have produced a dashboard that reads plausibly while being wrong, which is the specific failure mode this PR exists to prevent. Worth noting on a change whose entire purpose is to make an operator's picture of the queue trustworthy.

Final semantics: `try_push` counts nothing; the caller records terminal rejections at the two sites that actually send a queue-full response; sheds are counted separately under their own name; and the two rejection causes are displayed as separate tiles rather than summed.

[AI-assisted - Claude]

PR #4922, opened 2026-07-24
feat(broadcast): instrument which fan-out arm sends full state (#3335)

sanity:
## Problem

A production telemetry measurement on 2026-07-24 ([posted to #3335](https://github.com/freenet/freenet-core/issues/3335#issuecomment-5074264001)), prompted by a user reporting 4 GB of traffic in 90 minutes. Measuring one full telemetry rotation (~3 h, ~521 reporting peers, 93 % on 0.2.105) confirmed the report and showed the fleet baseline is high:

| percentile | combined rate | per day |
|---|---|---|
| median | 72 KB/s | ~6 GB |
| p90 | 255 KB/s | ~22 GB |
| p99 | 606 KB/s | ~52 GB |

81 % of outbound is bulk contract-state transfer. ~85 % of byte-weighted broadcast work is on contracts holding >= 500 KB of state, and **only 36 of 2,164 distinct contracts ever produced a real state change**.

What the measurement could **not** determine is *why* those large states go out whole. `broadcast_to_single_peer` has four distinct paths that send FULL STATE instead of a delta, none separately instrumented:

1. the `delta_incompat` memo is armed (#4904);
2. the wire-efficiency gate (`is_delta_efficient`) refused *before* computing anything, because the peer's summary is >= 50 % of our state size;
3. delta computation was attempted and failed;
4. no cached summary for one side of the pair.

Those four have different remedies, so without the split any fix is a guess. This PR closes that gap and nothing else — it is observation-only and changes no network behavior.

## Approach

Each arm is tagged where the payload is selected and recorded at the **two real-delivery sites**, the same gate that charges `BroadcastFanoutCost`, so the mix and the cost axis agree on what "sent" means and a dropped send never counts phantom bytes.

`compute_delta` returns a typed `DeltaUnavailable` instead of a bare `String`, so "the gate refused, no contract code ran" is distinguishable from "the contract failed". Its `Display` keeps the historical message prefix and adds the sizes.

A per-minute `broadcast_payload_mix` rollup reports per-arm sends and bytes, the full-state byte share, and the top contracts by full-state bytes.

**All window state lives behind a single `parking_lot::Mutex`** and is taken with one `mem::take`. This is deliberate and is the documented `.claude/rules/code-style.md` exception to preferring `DashMap`: the rollup needs an atomic read-modify-write across multiple keys (all arms plus the contract map) in one transaction, otherwise the aggregate and per-contract fields describe different sets of broadcasts. Cost is one short uncontended lock per *delivered broadcast* (not per packet), on a path that has already done WASM delta computation and a network write. `record_delivered` is a sync fn with no `.await` inside, so the guard cannot span an await point.

The published schema is designed to **add up**:

```
sum(top_contracts) + other_contracts_bytes + attribution_dropped_bytes == full_state_bytes
```

`other_contracts_bytes` covers contracts tracked but outside the reported top-N; `attribution_dropped_bytes` covers contracts the 256-entry cap refused to track at all. These are different omissions and conflating them hid a real gap (see review log below).

The simulation-only `broadcast_state_to_peers` is deliberately not instrumented; the production fan-out is the single site.

## Review log

Five rounds of external (Codex) review. **Four found real defects, all fixed:**

1. **Non-atomic window rollover.** Arm counters and the contract map drained at different instants, so an increment landing between sampling a map entry and `clear()` was lost, and one broadcast could be counted in one window for the aggregate and the next for per-contract fields. Fixed by the single-lock design above.
2. **`contracts_dropped` counted sends, not contracts.** One over-cap contract broadcasting 1,000 times read as "1,000 contracts dropped". Renamed `attribution_dropped_sends` and paired with `attribution_dropped_bytes`.
3. **The `deltas_suppressed` guard used a `_` wildcard**, so a missing-summary send was labelled memo-suppressed even though a delta was impossible anyway — overstating the memo, undercounting `FullNoSummary`. Scoped to `(Some(_), Some(_))`. Behavior unchanged; `compute_delta` lives only in the `(Some, Some)` arm, so the delta-incompat safety property holds on both paths.
4. **`window_secs` reported the nominal 60** even though `MissedTickBehavior::Delay` lets a saturated runtime stretch the real window, inflating every derived rate — worst exactly when the node is busiest. Now reports actual elapsed.
5. **The published schema didn't reconcile.** With 11..=256 contracts the top-N truncation dropped bytes no field accounted for. Added `other_contracts_bytes` + `contracts_tracked`.

Finding 5 is worth calling out: the existing reconciliation test *passed* while the bug was live, because it summed the in-memory map rather than the JSON a consumer receives. The test now asserts through `payload_mix_json`.

## Testing

- 14 unit tests: arm separation, window reset, delta-vs-full share (incl. the empty-window NaN boundary), bounded attribution with overflow reporting, deterministic tie-breaking, delta bytes never leaking into full-state attribution, actual-vs-nominal window duration, and schema reconciliation across truncation, over-cap, and both together.
- A multi-threaded `concurrent_records_racing_a_rollup_conserve_bytes` test: writer threads against a concurrent drain loop, asserting no byte is lost or double-counted across a rollover.
- Source-scrape pins: every arm is constructed, the mix is recorded at exactly the two delivery sites, the NotEfficient/ComputeFailed split survives refactors, and the suppression guard stays scoped.
- Full lib suite green (4285 passed, 0 failed). `cargo fmt` clean on the CI-pinned 1.93.0. `cargo clippy` reports nothing in any touched file. (Pre-existing clippy-1.94 lints in untouched files are a known separate follow-up.)

## Known gaps

- **The four-perspective Claude review panel did not run.** Six subagents were spawned across two attempts; every one went idle without producing a report. This PR therefore had one external model plus direct checks, not the Full-tier panel the review rule asks for on a broadcast-path change.
- **One unidentified test failure.** A single test failed once during a run sharing the machine with a concurrent clippy compile and a hot-spin loop in the new concurrency test. It did not reproduce across nine subsequent full-suite runs; the hot spin was removed. The failing test was never identified.
- **`broadcast_payload_mix` is not in the durable shadow-telemetry extractor's `EVENT_TYPES` allowlist** on nova. Raw telemetry retains ~3 days, so without that entry the data ages out before much accumulates. That script lives in the admin repo and deploys to nova, so it is not part of this PR.

Refs #3335. Overlaps #3798 (assigned to @iduartgomez), which closed the fan-out *width* gap in #4520/#4596; this is the payload *size* half. Follow-up hypothesis and the heavy contracts are in #4923.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge.rs
+++ b/crates/core/src/node/network_bridge.rs
@@ -16,6 +16,7 @@ use tokio::sync::mpsc::{self, Receiver, Sender};
 
 use crate::message::{NetMessage, NodeEvent};
 
+pub(crate) mod broadcast_payload_mix;
 pub(crate) mod broadcast_queue;
 mod handshake;
 pub(crate) mod in_memory;
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -8,7 +8,7 @@
 //! on contracts holding >= 500 KB of state, and that 97.9 % of it lands on
 //! contracts whose applies change nothing. What it could NOT determine is
 //! *why* those large states go out whole, because
-//! [`broadcast_to_single_peer`] has **three distinct paths that send FULL
+//! [`broadcast_to_single_peer`] has **four distinct paths that send FULL
 //! STATE** and none of them were separately instrumented:
 //!
 //! 1. [`PayloadArm::FullDeltaSuppressed`] — the `delta_incompat` memo is armed
@@ -42,19 +42,20 @@
 //!
 //! ## Cost
 //!
-//! One relaxed atomic add per delivered broadcast on the hot path, plus at
-//! most one bounded `DashMap` touch for per-contract attribution. Everything
-//! else happens in the aggregator task. Same shape as
-//! [`crate::transport::shadow_demand`]; the hot path only ever WRITES.
+//! One short uncontended mutex acquire per **delivered broadcast** (not per
+//! packet), covering a handful of integer adds and at most one bounded
+//! `HashMap` touch. Everything else happens in the aggregator task, and the
+//! hot path only ever WRITES. The lock is what makes a rollup a consistent
+//! snapshot; see [`PayloadMix`] for why per-field atomics were not enough.
 //!
 //! [`broadcast_to_single_peer`]: super::broadcast_queue::broadcast_to_single_peer
 
+use std::collections::HashMap;
 use std::sync::LazyLock;
-use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
-use dashmap::DashMap;
 use freenet_stdlib::prelude::ContractInstanceId;
+use parking_lot::Mutex;
 
 use crate::node::background_task_monitor::BackgroundTaskMonitor;
 
@@ -70,8 +71,10 @@ const ROLLUP_WINDOW: Duration = Duration::from_secs(60);
 /// contract we host, so an unbounded map here would be an amplification
 /// vector (see `.claude/rules/code-style.md`, "per-key collections"). Entries
 /// beyond the cap are still counted in the per-arm totals; only their
-/// per-contract attribution is dropped, and the count of dropped contracts is
-/// reported so a truncated window is never mistaken for a complete one.
+/// per-contract attribution is dropped, and both the count and the BYTE total
+/// of those unattributed sends are reported (`attribution_dropped_sends` /
+/// `attribution_dropped_bytes`) so a truncated window is never mistaken for a
+/// complete one.
 const MAX_TRACKED_CONTRACTS: usize = 256;
 
 /// How many contracts the emitted rollup names.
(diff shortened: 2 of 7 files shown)


sanity:
## Review round 5 — one further finding, NOT yet fixed

Codex round 5 on `3ef6a3fc`:

> **[P2] Keep the drained payload window scoped to one node** — `broadcast_payload_mix.rs:390`
> When multiple `NodeP2P` instances share a process, each starts its own aggregator, but every broadcast writes to the same process-global `BROADCAST_PAYLOAD_MIX`. Whichever node's ticker runs first drains all nodes' records and publishes them under its own `local_peer_id`, while the other nodes emit empty windows.

**Assessment — valid, production-benign, but a real defect:**

- **Production is unaffected.** One `NodeP2P` per process, so there is exactly one aggregator and one writer set.
- **Multi-node processes (the simulation harness) are affected**, and misattributed rather than merely duplicated.
- It is **worse than the established convention it copied**. `shadow_demand`'s aggregators read process-global counters *non-destructively* (`load` + locally-tracked delta), so N aggregators would each see the process total and double-report. This module's `take_window()` is destructive, so the first ticker to fire steals every other node's records.

**Fix (not applied):** give each node its own `PayloadMix` — hang it off the node (reachable from `op_manager` at both record sites, which already have it) instead of the `LazyLock` static, and hand the same instance to that node's aggregator. Small in principle, but it changes the hot-path record signature and therefore wants its own review round.

**Why this is not fixed here:** five consecutive external review rounds have each surfaced a genuine defect (non-atomic rollover, mislabelled counter, mis-scoped guard, wrong window duration, and now this). That is not a converging signal, and a hot-path change made late without a following review round is exactly how the previous four got in. Deferring to a fresh session rather than compounding.

Everything from rounds 1-4 IS fixed and covered by tests; see the updated PR description.

[AI-assisted - Claude]

PR #4926, opened 2026-07-25
fix: skip redundant queue-full resync retry when the first request landed (#4863)

### LINKED ISSUE #4863: Optional: conditional queue-full resync retry (skip redundant full-state resend when the first landed)
## Status update

The **aggregate-fan-out** half of this issue is now **addressed in PR #4862**: a node-wide cap (`MAX_OUTSTANDING_QUEUE_FULL_RESYNC_RETRIES`, `InterestManager::try_reserve_resync_retry_slot`) bounds the number of concurrently-outstanding queue-full-resync **retry tasks** regardless of throttle-LRU churn, which bounds the aggregate retry-induced full-state `ResyncResponse` fan-out per node. So the earlier MEDIUM (uncapped full-state fan-out under a broad drop storm) is subsumed for the retry path.

What remains is the **optional refinement** below. This is NOT a correctness gap — retries are bounded per-`(contract, peer)` (≤ `1 + QUEUE_FULL_RESYNC_MAX_RETRIES`), bounded node-wide (the cap), and each only fires after a real drop.

## Remaining: conditional retry

The retry is still **blind** — it re-dispatches regardless of whether the first `ResyncRequest` already landed. When the first landed, each extra retry makes the sender clear its summary and re-send full state again, so up to `QUEUE_FULL_RESYNC_MAX_RETRIES` redundant full-state `ResyncResponse`s land on a receiver whose queue was recently full. Making the retry **conditional** — re-dispatch only when the previous attempt was observed dropped — would eliminate the redundant resends in the common (first-attempt-landed) case.

This needs an **observed-drop signal**: the `P2pBridge::try_send` result at `broadcast.rs` (the actual drop point) fed back to the retry, or a "fresh state / ResyncResponse received since" check on the receiver side. Either is a non-trivial cross-layer change, hence deferred.

## Acceptance

- No redundant full-state `ResyncResponse` when the first `ResyncRequest` already landed.
- Preserve the #4857 heal (a rarely-changing field still re-syncs before the 5-min heartbeat), the #4251 per-`(contract, peer)` storm bound, and the #4862 node-wide cap.

Refs #4857, #4858, #4862.

[AI-assisted - Claude]

sanity:
## Problem

Follow-up to #4862 (the bounded trailing re-dispatch for the #4857/#4858 queue-full resync).

On a `ContractQueueFull` broadcast drop we send a rate-limited `ResyncRequest` to the sender so it clears its poisoned summary cache and re-sends. #4862 added a bounded trailing re-dispatch so that request still lands if the lossy `notify_node_event` → event loop → `P2pBridge::try_send` path drops it.

That retry is **blind**: it re-dispatches regardless of whether the first request already landed. Every `ResyncRequest` makes the sender clear its cached summary of us and re-send **full contract state**, so when the first request *did* land, each extra retry costs another full-state `ResyncResponse` — up to `QUEUE_FULL_RESYNC_MAX_RETRIES` (2) of them — landing on a receiver whose queue was full moments ago. That is amplification onto a just-saturated queue, the exact shape the #4251 storm bound exists to prevent.

## Solution

Make each retry attempt **conditional** on an observed-landed signal.

**Signal chosen: the #4864 round-8 correlation record** (`OutstandingResyncRequests`). `send_queue_full_resync_request` records `(contract, sender)` when it emits; the `ResyncResponse` receive arm in `node.rs` require-and-**consumes** that entry *before* applying the state. So the entry's absence is an **end-to-end** confirmation: the request landed, the sender answered, and we authorized the answer. Before each re-dispatch the retry asks `is_outstanding(contract, sender)`; if false it logs `queue_full_resync_retry_skipped_landed` and returns.

New read-only accessor `OutstandingResyncRequests::is_outstanding` — **never `consume`**: that one-shot authorization belongs to the receive arm, and consuming it in the retry would make the genuine `ResyncResponse` look unsolicited and be dropped without applying. Pinned by the storm-bound pin test.

`is_outstanding` is TTL-aware, matching `consume`'s acceptance predicate. Every other reason the entry can be absent leads to the same conclusion:

| Entry absent because | Would a re-dispatch help? |
|---|---|
| response arrived + applied (the common case) | No — heal is done |
| response arrived but the apply FAILED | No — the receive arm consumes before applying, and re-pulling a full state that just failed to merge is the #4861 poison-contract storm the merge backoff exists to stop |
| never recorded (brand-new key dropped at the strict `MAX_TRACKED_KEYS` cap) | No — the eventual response would be dropped as unsolicited anyway |
| TTL-expired (unreachable inside the ≤30s retry window; TTL is 60s) | No — same rejection |

**Why not the `P2pBridge::try_send` result** (the issue's option (a)): `try_send` succeeding does **not** mean the request reached the peer — the transport is UDP and the remote side can drop it too. So it is a negative-only signal, and gating retries on locally-observed drops alone would silently lose the #4857 heal in exactly the cases it exists for (sent fine locally, lost on the wire). It would also thread new cross-layer state (a bounded, TTL'd drop map) through the high-risk broadcast dispatch path. Option (b) is both simpler and strictly safer: it can only skip a retry once the heal has demonstrably completed.

**Ordering change:** the correlation record now happens **above** the retry spawn (it was between the spawn and the immediate enqueue). The retry reads that record, so a task spawned before it was written could observe "absent", conclude "landed", and skip the heal. Recording is a non-blocking map insert, so #4862 P2's requirement — spawn *before* the blocking `notify_node_event(...).await` — still holds. Pinned.

### Acceptance criteria

- **No redundant full-state `ResyncResponse` when the first request landed** — covered, with one honest residual (below).
- **#4857 heal preserved** — a genuinely dropped request leaves the record outstanding, so retries fire exactly as before (`queue_full_resync_retry_redispatches_within_reservation` still passes unchanged).
- **#4251 per-`(contract, peer)` storm bound preserved** — the retry still creates no reservation, consumes no global emit token, re-records nothing; the throttle window is untouched (it must not be released — that is the storm bound).
- **#4862 node-wide cap preserved** — and on a skip the retry slot is **released**, not burned: the task returns, dropping its RAII guard, instead of holding the slot for the rest of the window.

### Honest residual

The signal observes the **response**, not the request. A request that landed but whose `ResyncResponse` is still in flight (large state, slow WASM summarize) can still draw a retry. That window is bounded by the same `1 + MAX` per-`(contract, peer)` per-30s cap as before, and each subsequent attempt re-checks — so this is strictly fewer redundant full-state resends than the blind retry, never more. Closing it completely would need a request-level ack, which does not exist on this path.

Also unchanged (pre-existing behavior): if the *delta-failure* resync path records a new request for the same `(contract, sender)` mid-burst, the gate reads "outstanding" and a retry fires. That is the old blind behavior in a rare interleaving, not a regression.

## Testing

**Regression tests written FIRST and verified to FAIL without the fix** (both reported `Got [127.0.0.1:…, 127.0.0.1:…]` — the 2 blind retries — before the production change; both pass after):

- `queue_full_resync_retry_skipped_once_response_landed` (behavioral, `start_paused`) — one queue-full drop, then the response lands exactly the way `node.rs` does it (consume the correlation record), then the clock advances across the whole 20s burst window. **Zero** retries, and `outstanding_resync_retries() == 0` (slot released, not burned).
- `queue_full_resync_retry_stops_at_the_attempt_after_response_lands` (behavioral, edge case) — nothing landed yet, so retry 1 **must still fire** (heal preserved); the response then lands mid-burst and retry 2 must be skipped. Proves the gate is re-evaluated per attempt, not just on the first.

Unit tests for the new accessor:

- `outstanding_resync_is_outstanding_is_read_only_and_ttl_aware` — absent → false; recorded → true; scoped per `(contract, peer)`; repeated checks do **not** remove the entry (a later `consume` still succeeds); consumed → false; just-under-TTL → true, just-past-TTL → false.
- `outstanding_resync_is_strictly_capped` extended: an over-cap (never-recorded) key reads as not-outstanding, so the retry skips it.

Pin updates:

- `queue_full_resync_retry_is_bounded_and_reservation_scoped` — now also pins (a) the record precedes the retry spawn, (b) the retry contains no `.record(` **and no `.consume(`**, (c) the `is_outstanding` gate is checked **before** `try_notify_node_event`.
- `every_production_resync_request_emit_records_outstanding` — the fixed 600-byte lookback becomes an enclosing-fn lookback. The byte distance was always an arbitrary proxy; the reorder pushed the record past it. Same guarantee (an emit is preceded by its record, on the same code path), no brittle distance.
- Two pre-existing tests that call `resend_queue_full_resync_request` **directly** now record the outstanding request first, the way the real caller does. Without that, `…_terminates_when_injected_clock_frozen` would have passed for the wrong reason (skipped by the new gate rather than stopped by the liveness backstop) — a premise fix, not a weakened assertion.

`TimeSource` (`interest_manager.now()`, the injected throttle clock) and `GlobalRng` are used throughout; no new `Instant::now()`, no new `.send().await`, no new channel.

Commands run (worktree at `078d3459`):

```
cargo fmt                                    # clean
cargo clippy -- -D warnings                  # clean (the CI-exact invocation)
cargo test -p freenet --lib resync           # ok. 54 passed; 0 failed
cargo test -p freenet --lib                  # ok. 4273 passed; 0 failed; 25 ignored
cargo test -p freenet update                 # 250 passed after re-run (see note)
```

Note on `cargo test -p freenet update`: one run failed on `contract::executor::pool_tests::subscriber_limit_tests::send_update_notification_empty_local_snapshot_emits_warn`, which is unrelated to this change (executor subscriber-notification log capture, not the UPDATE resync path). It passed alone and on 6/6 subsequent identical runs, so it is order-dependent flake — filed separately as #4927 rather than papered over.

`cargo clippy --all-targets` reports 24 pre-existing `let_underscore_must_use` errors in test code across the crate (e.g. `wasm_runtime/contract_store.rs:1230`, untouched here). CI runs `cargo clippy --locked -- -D warnings` without `--all-targets`, so these are local-toolchain drift, not introduced by this PR.

Closes #4863

> Draft: concurrency + protocol-adjacent surface (Full tier). Needs a full multi-model review before ready/merge. Note that freenet-core DRAFT PRs skip the heavy Simulation and Unit&Integration CI jobs, so draft-green is a false signal — mark ready for review to get real CI.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -4972,10 +4972,13 @@ mod tests {
 
         // Seed the outstanding entry: WE sent a ResyncRequest to `source` for
         // `key`, so the incoming response is solicited.
-        op_manager
-            .ring
-            .outstanding_resync_requests
-            .record(*key.id(), source);
+        assert!(
+            op_manager
+                .ring
+                .outstanding_resync_requests
+                .record(*key.id(), source),
+            "seeding the outstanding entry must correlate (map is empty here)"
+        );
         assert_eq!(
             op_manager.ring.outstanding_resync_requests.len(),
             1,
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -1206,6 +1206,22 @@ async fn send_queue_full_resync_request(
         reason = "queue_full",
         "UPDATE relay: sending rate-limited ResyncRequest after queue-full drop (#4857)"
     );
+    // #4864 round-8: record the outstanding request so the matching ResyncResponse
+    // from this peer is authorized ONCE at the receive arm; an unsolicited or
+    // replayed response finds no entry and is dropped without running WASM.
+    //
+    // #4863: this MUST stay ABOVE the retry spawn below. The retry's conditional
+    // gate READS this record to decide whether the first request already landed
+    // (the receive arm consumes it on a matching ResyncResponse), so a retry task
+    // that could observe a not-yet-written record would conclude "landed" and skip
+    // the heal. Recording first makes that race structurally impossible; pinned by
+    // `queue_full_resync_retry_is_bounded_and_reservation_scoped`. Recording is a
+    // non-blocking map insert, so it does not delay the spawn (#4862 P2).
+    op_manager
+        .ring
+        .outstanding_resync_requests
+        .record(*key.id(), sender_addr);
+
     // #4862 (P2): the immediate ResyncRequest below rides a LOSSY path
     // (`notify_node_event` → event loop → `P2pBridge::try_send`), which can drop
     // it under the SAME backpressure that caused the queue-full drop. If dropped,
@@ -1222,8 +1238,11 @@ async fn send_queue_full_resync_request(
     // throttle's own clock) so it can NEVER spill into a new reservation. The
     // retries re-deliver the ONE already-authorized emit, so they do NOT
     // re-consult the per-sender throttle or the global emit cap and do NOT
-    // re-record the outstanding request (a redundant response for an
-    // already-consumed record is correctly dropped by #4864 before WASM).
+    // re-record the correlation entry recorded above.
+    //
+    // #4863: each attempt is CONDITIONAL on that correlation entry still being
+    // outstanding, so a retry fires only while the first request is NOT observed
+    // to have landed.
     //
     // #4862 P1: gate the spawn on a node-wide retry-task slot so the throttle LRU
     // evicting active reservations under key churn cannot spawn unbounded
@@ -1244,18 +1263,10 @@ async fn send_queue_full_resync_request(
         });
     }
 
-    // #4864 round-8: record the outstanding request (immediately before the emit
-    // so the round-8 pin's proximity check holds) so the matching ResyncResponse
-    // from this peer is authorized ONCE at the receive arm; an unsolicited or
-    // replayed response finds no entry and is dropped without running WASM. The
-    // #4862 trailing retry above re-sends the SAME request and relies on THIS one
-    // record — see `resend_queue_full_resync_request`.
-    op_manager
-        .ring
-        .outstanding_resync_requests
-        .record(*key.id(), sender_addr);
-
-    // Immediate (first) ResyncRequest — blocking best-effort enqueue.
+    // Immediate (first) ResyncRequest — blocking best-effort enqueue. The
+    // #4862 trailing retry above re-sends the SAME request and relies on the ONE
+    // correlation record made before the spawn — see
+    // `resend_queue_full_resync_request`.
     if let Err(e) = op_manager
         .notify_node_event(NodeEvent::SendInterestMessage {
             target: sender_addr,
@@ -1348,12 +1359,22 @@ fn jittered_resync_retry_delay(attempt: u32) -> Duration {
 /// clocks coincide and the injected stop fires first, so the floor is a no-op;
 /// it only takes effect when the injected clock stalls.
 ///
+/// Conditional (#4863): each attempt re-dispatches ONLY while the caller's
+/// request is still outstanding in the correlation map. The `ResyncResponse`
+/// receive arm consumes that entry before applying, so its absence means the
+/// heal already landed end-to-end — and re-dispatching then would make the sender
+/// clear its summary and ship FULL STATE again onto a receiver whose queue was
+/// just full (up to [`QUEUE_FULL_RESYNC_MAX_RETRIES`] redundant full-state
+/// responses per drop). The check is READ-ONLY (`is_outstanding`, never
+/// `consume`): consuming the entry here would make the genuine response look
+/// unsolicited and be dropped without applying.
+///
 /// Storm bound (#4251 / #4864 cap): this fn RE-DELIVERS the one already-authorized
 /// emit — it deliberately does NOT call `begin_resync_request` (no new
 /// reservation), does NOT consume a `resync_emit_limiter` token (no new global
-/// emit), and does NOT re-record `outstanding_resync_requests` (the initial
-/// record authorizes the eventual response once; a redundant response is dropped
-/// by #4864 before WASM). All its sends belong to the single reservation already
+/// emit), and does NOT re-record the correlation entry (the caller's one record
+/// authorizes the eventual response once; a redundant response is dropped by
+/// #4864 before WASM). All its sends belong to the single reservation already
 /// granted by the caller. Combined with the deadline stop, steady-state stays
 /// capped at one reservation (≤ `1 + MAX` sends) per (contract, peer) per window.
 ///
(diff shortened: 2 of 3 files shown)


sanity:
## Independent review (2 Claude lenses: skeptical + testing) — findings addressed in `e24fed6d`

Full tier per the concurrency/protocol-adjacent surface. Codex was not used (quota-blocked until ~2026-08-01); both lenses were Claude, run blind and independently.

**Neither lens found a blocking bug.** The core mechanism holds: the record/spawn reordering is correct (spawn establishes happens-before, so the task can never observe the pre-record state), `is_outstanding` and `consume` share the same clock and the same `<=` TTL predicate with no drift, the retry genuinely never consumes, and the residual is bounded — the gate can only suppress sends, never add one.

Five findings were worth fixing, and they are fixed:

### 1. Cap hole — the gate silently disabled itself node-wide (the real one)

`record()` dropped a brand-new key at `MAX_TRACKED_KEYS` and returned `()`. The gate then read `is_outstanding == false` and concluded "landed", so **while the map was full every queue-full retry was skipped node-wide, silently, with no signal.**

The accessor doc justified that with:

> the request was never correlated ... then the eventual response would be rejected as unsolicited anyway, so re-sending cannot heal anything either

That is **factually wrong**, and it is the load-bearing claim. The `ResyncRequest` heal has two independent channels and only one needs correlation:

- the responder clears its cached summary of us on receiving the **request itself** (`node.rs:3163-3167`, `update_peer_summary(.., None)`) — this is the #4857 heal, and it does not involve our record at all;
- only the full-state `ResyncResponse` needs the correlation entry.

So a re-dispatch **does** still heal when uncorrelated. `record` now returns whether the request is correlated, and the retry gates on it — uncorrelated requests fall back to unconditional re-dispatch, which is the correct behavior there. This also lets the accessor's contract drop the "absent for an unexpected reason" disjunction instead of arguing each case away.

### 2. The emit pin went vacuous for exactly its own regression class

This PR introduced the first **non-recording** use of `outstanding_resync_requests` (`is_outstanding`), while `every_production_resync_request_emit_records_outstanding` only grepped for the bare identifier. A **read** therefore satisfied a pin whose entire purpose is to require a **write**.

Verified empirically rather than by argument — replacing the helper's `.record(` with `.is_outstanding(`:

- **old assertion: PASSES** (vacuous)
- **new assertion: FAILS** with the intended message

It now matches `.record(` against whitespace-stripped source, so rustfmt re-wrapping the chain cannot break it. Also fixed `fn_start`, which matched only `"\nasync fn "` / `"\nfn "` and so skipped past `pub(crate) async fn`, anchoring to an *earlier* function whose window could contain an unrelated record — a false pass.

### 3. The change was unfalsifiable in production

The only skip signal was `tracing::debug!`, and `crates/core/Cargo.toml` sets `release_max_level_info`, so it is **compiled out of release builds**. Both the benefit and the main risk (a gate that wrongly skips silently loses the #4857 heal) were invisible in the field. Promoted to `info!`; the rate is reservation-bounded, so it cannot become a log storm.

### 4. Nothing pinned TTL > retry span

`OUTSTANDING_RESYNC_TTL` (60s) and the retry window were uncoupled. A TTL at or below the span would make later attempts read *their own expired entry* as "landed" and silently skip the heal — with entirely green CI. Now pinned next to the existing throttle-window assertion.

### 5. Comments overstated what the signal proves

- Absence means the response was **authorized**, not applied: the receive arm consumes *before* applying, and the apply can still fail (including with another `ContractQueueFull`). "End-to-end confirmation" was too strong.
- The amplification premise is **version-dependent**: against peers >= v0.2.102 the responder's own `resync_response_limiter` already caps full-state replies at one per `(peer, contract)` per 30s, and the whole retry burst fits inside one such window. The gate is still worth having (it also stops the redundant *requests*, and it is the only bound not dependent on the remote peer's version) — but the comments asserted it unconditionally.
- Corrected a false rationale: the comment on `queue_full_resync_retry_terminates_when_injected_clock_frozen` claimed that without the added record the test would pass for the wrong reason. It would not — the injected clock is frozen, `now >= target` never holds, and the gate is **unreachable** on that path. The record is inert there and that test does **not** cover the gate. Saying otherwise gave false confidence about DST coverage.

### Verification

- Both regression tests re-verified **non-vacuous by execution**, not reasoning: disabling the gate fails `queue_full_resync_retry_skipped_once_response_landed` and `queue_full_resync_retry_stops_at_the_attempt_after_response_lands` (6 passed / 2 failed), and both pass with it restored.
- `cargo test -p freenet --lib resync` — 54/54.
- `cargo fmt --all -- --check` and `cargo clippy --locked -p freenet -- -D warnings` (CI's exact invocation) clean.
- Rebased onto current `main` (`6bd75e2d`), no conflicts.

### Deliberately not changed

- **Cross-clock coupling** (the gate's TTL runs on `Ring`'s `InstantTimeSrc`, the retry paces on the injectable `interest_manager.now()`). Not exploitable today — the retry is hard-bounded by one reservation interval against a 60s TTL stamped microseconds before the spawn, so expiry inside the window is unreachable. Finding 4's pin is what keeps that margin honest if either constant is retuned.
- **Delta-failure interleaving** — pre-existing behavior, and the PR body already documents it. It can only reproduce the old blind behavior, never something worse.

Three unrelated tests (`test_slow_start_exit_schedules_slowdown`, `test_send_stream_without_bandwidth_limit`, `test_timeout_metering`) failed in a full local `--lib` run under heavy parallel load on this machine and all pass in isolation; they are transport/wasm-metering, not on any path this PR touches, and CI is the arbiter.

[AI-assisted - Claude]

Issue #4877, opened 2026-07-20
vega (7GB) module-cache thrash: undersized cache on small-RAM gateway → sustained recompile CPU, intensified once 0.2.102 unblocked the poison-timeout stall


sanity:
## Problem

After the 0.2.102 rollout (which contains the #4864 poison-contract CPU-churn fix), the **vega** gateway shows sustained high CPU (~100–170% of a core) driven almost entirely by **module-cache thrash**, while **nova** on the identical binary is healthy (~12% CPU). This is a RAM-size effect, not a code regression:

| Gateway | RAM | Module-cache budget (RAM/8, clamp 4 GiB) | Module-cache misses / 3000 log lines | CPU |
|---|---|---|---|---|
| nova | 125 GB | 4 GiB (clamp binds) — holds working set | ~15 | ~12% |
| vega | 7 GB | ~0.9 GiB (divisor binds, below even old 1.5 GiB clamp) | ~838 | ~100–170% |

vega's contract working set exceeds its ~0.9 GB in-memory module cache, so it constantly reloads/recompiles WASM modules. #4864's module-cache clamp raise (1.5→4 GiB) does **not** help vega because the `RAM/8` divisor binds far below the clamp on a 7 GB box.

## Why it intensified with 0.2.102 (mechanism)

On 0.2.101, vega's contract-handling loop was frequently **blocked** on poison-contract executions hitting the 5 s timeout (159 WASM timeouts/hr), so it churned through comparatively **few** distinct contracts per unit time → low cache-miss rate (~31/3000 lines). 0.2.102's fix (epoch preemption + merge-failure backoff) **unblocks** the loop — poison contracts no longer stall it for 5 s each — so it now processes the **full** contract set, which thrashes vega's undersized cache (~838/3000 lines, WASM timeouts down to near-zero). The bottleneck shifted from timeout-blocking to cache-thrash; net CPU stayed high on this specific under-resourced box.

This is a **net win for the network** (poison-contract CPU burn contained; nova 82%→12%; no crashes/panics; vega memory stable at 2.3 GB / 7 GB, no OOM), but it surfaces vega's pre-existing under-resourcing.

## Not a rollback situation

[...]

PR #4928, opened 2026-07-25
feat: surface cost-pressure eviction decisions for field diagnosis

sanity:
## Problem

Cost-pressure eviction (#4861, PR #4903, shipped in 0.2.105) is effectively
**write-only in the field**. The only log on the path fires on a *successful*
eviction, so the case that actually needs diagnosing — a sweep that shed
**nothing** — is completely silent. There is no log, metric, or counter when an
axis is under its floor, when no contract clears `COST_SHARE_THRESHOLD`, or when
a would-be offender is rejected by the zero-demand candidacy gate.

The gap is structural, and holds independently of any particular incident: on
a sweep that sheds nothing, these causes are indistinguishable from outside the
process even though each needs a different fix.

- the axis is below its floor;
- an offender exists but is immune **by design** (hosting-invariant 3:
  subscribed, or recently read);
- the load is diffuse — no single contract is individually above the axis
  floor, so the axis armed only on the sum;
- a real offender exists but has not yet SUSTAINED an above-floor rate;
- the work is attributed, but to contracts this node does not host.

The original author reports this blocking a diagnosis on the `vega` gateway
around 2026-07-25 (a sustained storm during which cost eviction never fired).
I have not independently corroborated that from logs, so the case above rests
on the structural argument rather than on that incident. Relatedly, #4877
(open) lists reducing vega's hosted contract count via demand-driven eviction
as one of its options, which is a question this line would help answer.

## Solution

Make the **decision** observable. No semantic change to eviction.

**1. A pure reject-reason classifier.** `classify_cost_rejection(..)` returns
`None` under exactly the condition `evict_cost_pressure` sheds a contract, and
otherwise names the gate that saved it (`zero_rate`, `under_share`,
`has_local_subs`, `has_downstream_subs`, `recently_accessed`).

**2. A rate-limited operator summary.** One `info!` per sweep carrying, per
axis: name, `total_rate`, `floor`, whether it **armed**, how many contracts the
meter attributed sustained work to, and every hosted contract over the share
cutoff with its rate, share, and outcome.

One line, wrapped here for readability. `axes` is recorded with `%`, so its
value is rendered unquoted, and `build_cost_axes` always returns all three axes:

```
cost-pressure eviction decision (#4861) axes_armed=1 hosted_contracts=214 cost_evictions_total=0 \
  axes=exec_cpu_micros_per_sec: total=1203.400 floor=50000.000 armed=false; \
       broadcast_fanout_bytes_per_sec: total=4102.900 floor=131072.000 armed=false; \
       broadcast_messages_per_sec: total=44.100 floor=10.000 armed=true attributed=2 \
         top=<key> rate=25.600 share=0.581 outcome=has_downstream_subs \
         also=<key> rate=11.200 share=0.254 outcome=recently_accessed
```

`axes` is deliberately the LAST field — its value contains spaces and `=`, so
the record only parses as key=value because nothing follows it. That ordering is
pinned by a test, not just a comment.

`attributed` is the field that makes the "nothing fired" case diagnosable,
because `total_rate` and the per-contract rates come from **different filters**
(`Meter::contract_cost_rates_multi` sums every positive rate into the total but
inserts only the sustained subset into `rates`, where "sustained" is a run
*above the axis floor*, not sample continuity). The decision table is on
`CostAxisDecision::attributed`.

`info!`, not `debug!` — release builds compile out `debug!` via
`max_level_info`, which is why the existing `queue_full_resync_throttled` debug
lines were useless during the incident.

**3. Counter export.** `cost_evictions_total` already reaches central telemetry
via `RouterSnapshot::hosting_cost_evictions_total` (#4903). The gap was the
**local dashboard** — the only hosting-counter view an operator gets on the box
with no collector in the loop. Added as its own tile, appended after "Evicted
w/ demand" so the #4338 miscalibration alarm keeps its first-row slot in the
5-column grid.

## Telemetry load at scale

Ian's constraint: telemetry changes must not overload the collector even at 10x
the current fleet (~1,700 peers today, so budget ~17,000).

**This PR adds zero records to the central collector.** Verified two ways:

- The tracing-subscriber stack (`tracing/tracer.rs`) wires only
  `fmt::layer()` sinks — rolling files, stderr, console. There is no
  `tracing_opentelemetry` layer, no `OpenTelemetryTracingBridge`, no
  `opentelemetry_sdk` anywhere in `crates/core/src`. A plain `tracing::info!`
  therefore cannot reach the collector.
- The collector's inputs are an explicitly enumerated set: `EventKind` via
  `register_events`, plus `send_standalone_event*`. `to_otlp_logs`
  (`tracing/telemetry.rs`) builds the OTLP body from `TelemetryEvent` structs
  only, stamping `severityNumber: 9 // INFO` on each. This PR touches neither
  `router.rs` nor `telemetry.rs`.
- `HostingSnapshot` has no `Serialize` derive; it reaches only the local HTTP
  dashboard, pull-based per page load. `RouterSnapshot::hosting_cost_evictions_total`
  already existed at 1 record/300s/node and is unchanged.

Arithmetic: **0 new records/hour/node x 17,000 peers = 0 new bytes to nova.**
Strictly better than the `broadcast_payload_mix` / `outbound_message_mix` 60s
rollup pattern, which at 17,000 peers is ~567 records/s arriving at the
collector.

Residual cost is **local disk only**, hard-bounded by the 60s sweep at one line
per sweep:

| Case | lines/h | bytes/line | bytes/h/node |
|---|---|---|---|
| Steady, quiet peer (all axes unarmed) | 12 | ~370 | ~4.5 KB |
| Steady, busy gateway (one axis armed) | 12 | ~465 | ~5.6 KB |
| Worst case (digest flips every sweep) | 60 | ~1.2 KB | ~70 KB |

Against a **measured** 35.15 MB/hour main-log rate on nova (12-hour mean) that
is 0.016%-0.2%. Both tracing rate limiters (1000/s global, 30/s per-callsite)
are untouched by orders of magnitude at this callsite's 0.0167/s peak.

## Hosting invariants

`.claude/rules/hosting-invariants.md` requires any change to the eviction path
to state which invariants it touches and how it preserves them.

This touches **invariant 3** (eviction is one demand-ordered decision) only by
reading it. The ranking, the candidacy predicate, the local/downstream
subscriber tiers, the recency terms, the share threshold, the axis floors and
the cost-eviction candidacy rule are byte-identical; no eviction input is added,
removed, or reordered. What is new is an explanation derived from that same
decision pass. No new mechanism is introduced, so the "test for a new mechanism"
gate does not apply.

## Scope (CONTRIBUTING)

Flagging rather than assuming: this is titled `feat:` and adds three observable
surfaces — an operator log line, a dashboard tile, and crate-internal API
(`CostAxisDecision`, `CostTopOffender`, `CostRejectReason`,
`classify_cost_rejection`, `evict_cost_pressure_observed`). It changes no
behaviour, but it is not a bug fix either, and neither referenced issue approves
it: #4861 is closed and #4877 is a different problem. Per
`.claude/rules/`, hosting/eviction changes are Ian's sign-off. Happy to open a
tracking issue mirroring #4658's shape if a maintainer wants one on record.

## Why eviction outcomes are unchanged

- The gates, their order, the candidacy predicate, the descending-rate sort and
  its key-byte tiebreak, and the removal loop are unchanged.
- The `filter_map` became an explicit loop only so the reporting scan can reuse
  the same pass. The explanation is derived **from** the decision, never from a
  second read of the same state that could race to a different conclusion.
- The one semantic tightening is defensive: a NaN attributed rate is now
  rejected explicitly in the share pre-filter instead of falling through to
  `cost_eviction_candidate` (which also rejects it), so the victim set is
  identical. It matters because the reporting pass sorts `probed`, where a NaN
  makes the comparator a non-total order and defeats the share-based bound that
  keeps the vector — and the log line — small.
- The `#4903` review's perf property holds: the cheap `rate <= share_cutoff`
  gate still short-circuits before the two DashMap `subscriber_counts` reads and
  the recency check.
- No new time source: `TimeSource` throughout.

## Testing

15 new tests in `crates/core/src/ring/hosting/cache.rs`, one in
`server/home_page.rs`, and one source-scrape pin in `ring.rs`. Highlights:

- `cost_reject_classification_matches_eviction_decision` — grid over 11 rates
  (incl. NaN, +/-inf, MIN_POSITIVE, both cutoff boundaries) x local x downstream
  x recency, asserting the classifier agrees with the sweep's own condition.
- `cost_reject_classification_matches_the_real_sweep` — the same grid driven
  through `evict_cost_pressure_observed`, because the test above compares the
  classifier against a hand-transcribed copy of the share gate and both could
  drift together away from the code.
- `cost_decision_logging_state_does_not_change_evictions` — two identical
  caches, one with the rate limiter primed to the digest the sweep actually
  produces (so it genuinely stays silent) and one fresh, shed identical victims.
- `cost_decision_top_rank_may_be_immune_while_a_lower_rank_is_shed` — pins that
  `top=` is the highest-rate probe, not necessarily the victim, and that
  `evicted` is a per-axis count.
- `cost_decision_digest_tracks_reasons_not_rates` and
  `..._ignores_offender_rank_swaps` — the change-trigger fires on a real gate
  change and stays quiet on rate wobble, which is what the volume figure rests
  on.
- `cost_decision_nan_rate_never_enters_the_report`.
- `log_cost_decision_still_emits_an_info_event_with_its_fields` — source-scrape
  pin on the emit. A subscriber-capture test was written first and rejected as
  inherently flaky: `set_default` is thread-local while tracing's callsite
  `Interest` cache is global, so sibling tests poison it (measured: 1 of 2
  events captured under `cargo test --lib cost`, passing when run alone).
- `dashboard_hosting_counter_source_tests` — pins that each dashboard eviction
  counter is mapped from its identically-named `stats` field (the #4009/#4010
  silent-rot pattern; `dashboard_hosting_snapshot` needs a live `Ring`).

```
cargo fmt --all --check                  # clean
cargo clippy --locked -- -D warnings     # clean (CI-equivalent invocation)
cargo test -p freenet --lib              # 4347 passed, 0 failed, 25 ignored
```

Behaviour-pinning tests were mutation-verified: reverting the NaN guard, the
`rate > 0.0` top-offender gate, the multi-offender reporting, the `attributed`
field, the per-axis eviction count, the digest sort, the rate limiter, and the
emit itself (deletion and `info!`->`debug!`) each fail a specific test.

## Review

Full-tier (eviction is a high-risk surface per `.claude/rules/pr-quality.md`).
Four independent Claude lenses: skeptical/adversarial (twice — once on the
original, once on the reworked code), telemetry-load-at-scale, and testing.
Codex was not attempted: it is quota-blocked until ~2026-08-01, and per the
current review policy external models are opt-in rather than a required pass.

The second round found four diagnostic-correctness defects — all fixed here,
because a summary that misdirects an operator is worse than no summary:
`zero_rate` conflating the meter's sustained gate with "nothing attributed"; an
all-zero axis naming an arbitrary innocent contract as the top offender; only
the top offender being explained when a storm contract ranked second; and a
decision-table row describing a state that cannot occur.

Refs #4861, #4903, #4877

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -1612,6 +1612,18 @@ pub struct HostingSnapshot {
     /// times). Should stay near zero; a rising value means the demand estimate
     /// is mis-ordering the working set (the #4338 miscalibration symptom).
     pub evictions_of_recently_read_total: u64,
+    /// Monotonic count of COST-PRESSURE evictions (cost-aware eviction,
+    /// #4861): zero-demand contracts shed because their attributed update-work
+    /// (WASM CPU / broadcast fan-out / broadcast message count) dominated the
+    /// node's total on a cost axis, independently of the byte budget. Disjoint
+    /// from [`Self::budget_evictions_total`].
+    ///
+    /// Surfaced here because the local dashboard is the only hosting-counter
+    /// view an operator gets ON the box, with no telemetry collector in the
+    /// loop — the gap that left the 2026-07-25 vega storm undiagnosable. The
+    /// same counter reaches central telemetry as
+    /// `RouterSnapshot::hosting_cost_evictions_total`.
+    pub cost_evictions_total: u64,
     /// Per-contract Greedy-Dual rows in EVICTION order (the next victim under
     /// budget pressure is first). May be truncated by the renderer; the full
     /// count is `contract_count`.
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -4234,6 +4234,7 @@ impl Ring {
             contract_count: stats.contract_count,
             budget_evictions_total: stats.budget_evictions_total,
             evictions_of_recently_read_total: stats.evictions_of_recently_read_total,
+            cost_evictions_total: stats.cost_evictions_total,
             contracts,
             disk_state_bytes,
             disk_wasm_bytes,
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -8710,3 +8710,71 @@ pub(crate) enum RingError {
     #[error("Peer has not joined the network yet (no ring location established)")]
     PeerNotJoined,
 }
+
+#[cfg(test)]
+mod dashboard_hosting_counter_source_tests {
+    //! Source-scrape pin for the dashboard hosting-counter mapping.
+    //!
+    //! `.claude/rules/bug-prevention-patterns.md` calls out "manually-mirrored
+    //! telemetry counters / dashboard providers" as a recurring silent-rot
+    //! pattern (#4009, #4010): a counter is wired to a card once, a later
+    //! refactor re-points or drops the mapping line, and the tile keeps
+    //! rendering a plausible-looking number from the wrong source with every
+    //! test still green.
+    //!
+    //! `dashboard_hosting_snapshot` is only reachable from `p2p_impl.rs`'s
+    //! provider registration and needs a live `Ring` to exercise, which this
+    //! suite has no scaffolding for. So — mirroring the other `*_source_tests`
+    //! modules in this file — pin the mapping in source instead.
+
+    /// Production source only: everything before the first top-level test
+    /// module. This module is itself `#[cfg(test)]` and declared last, so the
+    /// needles below cannot match this file's own test text.
+    fn production_source() -> &'static str {
+        const FULL: &str = include_str!("ring.rs");
+        let cutoff = FULL
+            .find("\n#[cfg(test)]\nmod ")
+            .expect("ring.rs must have a top-level #[cfg(test)] mod section");
+        &FULL[..cutoff]
+    }
+
+    #[test]
+    fn dashboard_hosting_snapshot_maps_each_eviction_counter_from_its_own_source() {
+        // Whitespace-stripped so a rustfmt reflow of the struct literal cannot
+        // silently vacate the pin.
+        let src: String = production_source()
+            .chars()
+            .filter(|c| !c.is_whitespace())
+            .collect();
+
+        // Each counter must come from the IDENTICALLY-named field on `stats`.
+        // Cross-wiring any two (the #4009/#4010 failure) breaks the matching
+        // needle. Needles are split so they do not appear contiguously in this
+        // file's own source.
+        for (counter, needle) in [
+            (
+                "cost_evictions_total",
+                concat!("cost_evictions_total:stats.", "cost_evictions_total,"),
+            ),
+            (
+                "budget_evictions_total",
+                concat!("budget_evictions_total:stats.", "budget_evictions_total,"),
+            ),
+            (
+                "evictions_of_recently_read_total",
+                concat!(
+                    "evictions_of_recently_read_total:stats.",
+                    "evictions_of_recently_read_total,"
+                ),
+            ),
+        ] {
+            assert!(
+                src.contains(needle),
+                "`dashboard_hosting_snapshot` must map {counter} from \
+                 `stats.{counter}` — the local dashboard is the only \
+                 hosting-counter view an operator gets with no telemetry \
+                 collector in the loop, so a cross-wired tile is silent"
+            );
+        }
+    }
+}
--- a/crates/core/src/ring/hosting/cache.rs
+++ b/crates/core/src/ring/hosting/cache.rs
@@ -761,6 +761,14 @@ pub struct HostingCache<T: TimeSource> {
     /// #4861). Only [`Self::evict_cost_pressure`] increments it. See
     /// [`HostingCacheStats::cost_evictions_total`].
     cost_evictions_total: u64,
+    /// Time of the last cost-pressure decision summary, for the rate limit in
+    /// [`Self::log_cost_decision`]. Observability only — never read by the
+    /// eviction decision.
+    last_cost_decision_log_at: Option<Instant>,
+    /// Digest of the last logged cost-pressure decision, so a CHANGE in the
+    /// decision surfaces within one sweep instead of waiting out
+    /// [`COST_DECISION_LOG_INTERVAL`]. Observability only.
+    last_cost_decision_digest: Option<u64>,
     /// Time source for testability
     time_source: T,
 }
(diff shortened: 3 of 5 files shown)


sanity:
## Consolidated review

Full tier (eviction is a high-risk surface per `.claude/rules/pr-quality.md`). Four independent blind Claude lenses, each reading the checked-out code:

| Lens | Focus |
|---|---|
| Skeptical / adversarial (round 1) | eviction-outcome equivalence, classifier faithfulness, bounds |
| Telemetry load at scale | collector impact at 1,700 and 17,000 peers |
| Testing | vacuous premises, mutation sensitivity, coverage gaps |
| Skeptical (round 2) + code-first | the reworked code, and intent-vs-implementation |

Codex was not run: it is quota-blocked until ~2026-08-01, and per the current review policy external models are opt-in rather than a required pass.

### The scale question

The governing constraint was that no telemetry change may overload the collector even at 10x the fleet. **This PR adds zero records to it.** The tracing-subscriber stack wires only `fmt::layer()` sinks; the collector's inputs are the enumerated `EventKind` / `send_standalone_event*` set; `HostingSnapshot` has no `Serialize` derive and reaches only the local pull-based dashboard. `RouterSnapshot::hosting_cost_evictions_total` already existed at 1 record/300s/node and is untouched. Arithmetic and citations are in the PR body.

Residual cost is local disk, hard-bounded at one line per 60s sweep: ~5 KB/hour/node steady state, ~70 KB/hour worst case, against a measured 35 MB/hour main log. For comparison, the existing `broadcast_payload_mix` / `outbound_message_mix` rollups would put ~567 records/s on the collector at 17,000 peers; this puts zero.

### Findings fixed

Round 2 found four defects in the *diagnoses themselves*, which matter more than usual here — a summary that misdirects an operator is worse than no summary:

- `zero_rate` conflated the meter's sustained gate with "nothing attributed". `total_rate` sums every positive rate while `rates` holds only the sustained subset, where "sustained" is a run above the **axis floor**, not sample continuity. Added `attributed` with a decision table.
- An all-zero armed axis named an arbitrary innocent contract as top offender — the same key, every heartbeat, deterministically. Now reports `top=none`.
- Only the top offender was explained. A storm contract ranked second behind a legitimately-expensive one had its reason silently dropped. All over-cutoff offenders are now reported.
- A decision-table row described a state that cannot occur.

Round 3 (code-first) found more:

- The emit source-scrape pin scraped to the end of the production region, so a `debug!` added to any later function would false-red, and another function's `info!` could false-green it. Window is now bounded to the function.
- `axes` must be the last field (its value contains spaces and `=`); this is now pinned by a test rather than a comment.
- `attributed` was documented as "network-wide" when it is meter-wide — an operator reading it as fleet-wide draws the opposite conclusion.
- Row 1 of the table conflated "diffuse load" with "a real offender not yet sustained for 150s". Different fixes; now distinguished.
- The change-digest now includes `attributed > 0`, so the row-1-to-row-2 crossing surfaces within a sweep instead of waiting out the 300s heartbeat.

Testing lens found, and this is fixed:

- Nothing asserted the log line is emitted at all — deleting the `tracing::info!` left every test green. A subscriber-capture test was written first and **rejected as inherently flaky**: `set_default` is thread-local while tracing's callsite `Interest` cache is global, so sibling tests poison it (measured: 1 of 2 events captured under `cargo test --lib cost`, passing when run alone). Replaced with a source-scrape pin.
- One test primed a rate limiter with an arbitrary digest that never matched, so the branch it claimed to exercise never ran and both scenarios logged. It now primes with the real digest and asserts one emitted and the other did not.
- A NaN rate could enter the reporting vector, where it makes the sort comparator a non-total order and defeats the share-based size bound. Rejected explicitly; victim set unchanged either way.
- Per-axis `evicted` attribution, digest component sensitivity, and the dashboard counter mapping (`#4009`/`#4010` silent-rot pattern) were all unpinned. Now pinned.

Every behaviour-pinning test was mutation-verified: reverting each fix fails a specific test, including outright deletion of the emit and `info!` to `debug!`.

### Findings deliberately not taken

- **Rename `offenders` / `CostTopOffender` to `probes` / `CostProbe`.** Fair point — the vector holds probes, and an entry with `rejected: Some(..)` is exonerated. But the operator-facing surface is the `top=` / `also=` tokens, which are unaffected, and a rename late in a high-risk file is churn without behavioural benefit. The rustdoc states the distinction explicitly.
- **Extract the probe scan out of `evict_cost_pressure_observed`.** The function grew, mostly in comments. Extracting would preserve the same-pass guarantee, but restructuring the eviction function for readability carries more risk than it buys on this surface.
- **Suppress the heartbeat on idle nodes.** The PR's thesis is that silence is ambiguous; "the code ran and no axis armed" is exactly the answer that was missing. Costs ~4.5 KB/hour on a quiet peer.

### Needs a maintainer

Flagging rather than assuming, per CONTRIBUTING: this is `feat:`-titled and adds observable surfaces, and neither referenced issue approves it (#4861 is closed; #4877 is a different problem). Hosting/eviction changes are @sanity's sign-off. The invariant-3 preservation statement `hosting-invariants.md` requires is now in the PR body.

Also worth a second look: the motivating vega incident is the original author's report and I could not corroborate it from logs, so the PR body now rests the case on the structural argument and labels the incident as uncorroborated.

[AI-assisted - Claude]

sanity:
## Closing: the motivating problem could not be corroborated

Closing per @sanity's decision. Recording the full state here so nothing is lost and whoever revisits starts from the corrected version rather than the original.

### What this was for

Cost-pressure eviction (#4861, PR #4903, shipped in 0.2.105) is effectively write-only in the field. The only log on the path fires on a *successful* eviction, so a sweep that shed **nothing** is silent — and that is the case that needs diagnosing. When cost eviction does not fire, five distinct causes are indistinguishable from outside the process, and each needs a different fix:

- the axis is below its floor;
- an offender exists but is immune **by design** (hosting-invariant 3: subscribed, or recently read);
- the load is diffuse — no single contract is individually above the axis floor, so the axis armed only on the sum;
- a real offender exists but has not yet **sustained** an above-floor rate (the meter's gate is a run above the axis floor for `min_window / 2` = 150s, not sample continuity);
- the work is attributed, but to contracts this node does not host.

This PR added a rate-limited per-sweep `info!` naming which of those held, plus a local dashboard tile for the existing `cost_evictions_total` counter.

### Why it's closed

The PR is `feat:`-titled and adds observable surfaces (an operator log line, a dashboard tile, and crate-internal API). Neither referenced issue approves it — **#4861 is closed** and **#4877 is a different problem** (module-cache thrash on an under-RAM'd box). Hosting/eviction changes are @sanity's sign-off per project rules.

The deciding factor is that **the motivating incident could not be corroborated.** The original premise was a ~7-hour storm on the `vega` gateway around 2026-07-25 during which cost-pressure eviction never fired once. I could not find evidence for that in logs or issues, and it sits in tension with `.claude/rules/hosting-invariants.md`, which records the #4861 storm as shed within ~15 min of the 0.2.105 upgrade and field-validated. I rewrote the PR body to rest on the structural argument above and to label the incident unverified — which was the honest thing to do, but it left the PR without the concrete problem it was built to solve.

### What was measured (the scale question)

The review was driven by the constraint that no telemetry change may overload the collector even at 10x the fleet. **This adds zero records to the central collector**, verified from source three ways:

- `tracing/tracer.rs` (`init_tracer` / `init_stdout_tracer`) registers only `tracing_subscriber::fmt::layer()` sinks — rolling file, error file at WARN, TTY console. There is no `tracing_opentelemetry`, no `OpenTelemetryTracingBridge`, no `opentelemetry_sdk` anywhere in `crates/core/src`; `init_tracer`'s endpoint parameter is literally `_endpoint`, unused.
- The collector's producers are exactly `TelemetryReporter::register_events` (`EventKind`), `send_standalone_event*`, `notify_of_time_out`, and two worker-internal generators. `to_otlp_logs` builds the POST body from `TelemetryEvent` structs only.
- The diff touches neither `router.rs` nor `telemetry.rs`. `HostingSnapshot` has no `Serialize` derive — it reaches only the local pull-based dashboard.

So a plain `tracing::info!` never enters `admit_event` and cannot occupy a slot in a saturated `(peer, second)` bucket against `MAX_EVENTS_PER_SECOND = 10`.

Decision frequency, for anyone sizing a future version of this: `evict_cost_pressure` has one production call site (`ring/hosting.rs`), driven by `sweep_get_subscription_cache` on `GET_SUBSCRIPTION_SWEEP_INTERVAL = 60s`. That is **1 decision/node/60s**, or **283 decisions/s network-wide at 17,000 peers**. Against a collector measured at ~94 GB/day, ~1,526 records/s, ~713 B/record:

| Shape | records/s @ 17k peers | GB/day | vs 94 GB/day |
|---|---|---|---|
| Per-decision telemetry event (every sweep) | 283 | 17.4 | **+18.5%** |
| Rate-limited to a 300s heartbeat | 57 | 3.5 | +3.7% |
| **What this PR actually did (local log only)** | **0** | **0** | **0%** |

**If this is ever revived as a telemetry event rather than a local log, the first row is the number to beat** — a per-event stream at this cadence would be a significant addition and would displace other telemetry on exactly the bursting peers you would most want to observe. A field on an existing rollup window is the shape to prefer.

Local-disk cost was the only cost: hard ceiling of one line per 60s sweep, ~5 KB/hour/node steady state and ~70 KB/hour worst case, against a measured ~35 MB/hour main log (0.016%–0.2%).

### What is unaffected by closing

`RouterSnapshot::hosting_cost_evictions_total` **already shipped in #4903** and is untouched by this PR. The counter that reaches central telemetry (1 record/300s/node) is unaffected. Only the local dashboard card goes away.

### Where the work is

Branch **`feat/cost-eviction-observability`**, head **`ce2f2de1e8816d343c201919c941314c92246623`** (also reachable as `refs/pull/4928/head`). Six commits on top of `main`. Not deleted.

CI: 13 of 15 checks green on `9658dc3c`, the immediately preceding SHA whose only difference from `ce2f2de1` is one corrected comment — including Simulation, Unit & Integration, NAT Validation and Windows Check. The two outstanding were macOS jobs queued on runner capacity, not failures. Locally: `cargo test -p freenet --lib` = 4347 passed / 0 failed, clippy and fmt clean.

### Review findings already fixed on that branch

Four blind Claude lenses ran (adversarial x2, telemetry-scale, testing, code-first). None found a behaviour change to eviction — the gates, ordering, candidacy predicate, sort and removal loop are unchanged, and the pinning tests were mutation-verified. But four defects were found in the **diagnoses themselves**, which for this PR is the whole failure mode, since a summary that misdirects an operator is worse than no summary. **All four are fixed on the branch — start from `ce2f2de1`, not from the original commit:**

1. **`zero_rate` conflated the meter's sustained gate with "nothing attributed".** `Meter::contract_cost_rates_multi` sums *every* positive per-contract rate into `total_rate` but inserts only the **sustained** subset into `rates`, where sustained means a run above the **axis floor**. So an axis can arm on genuinely contract-attributed work while every hosted contract reads 0.0. Fixed by adding `CostAxisDecision::attributed` with a decision table.
2. **An all-zero armed axis named an arbitrary innocent contract as top offender** — the same key, every heartbeat, deterministically, because the key-byte tiebreak resolved a tie at 0.0. Fixed by gating on `rate > 0.0` and reporting `top=none`.
3. **Only the top offender was explained.** A storm contract ranked second behind a legitimately expensive one had its reason silently dropped while the line reported the legitimate contract's. Fixed by reporting every over-cutoff offender (bounded at ~3 by construction).
4. **A decision-table row described a state that cannot occur** ("attributed but no offender, or none clears 25%" — a positive-but-under-cutoff hosted contract takes the fallback and renders `under_share`, so the list is non-empty).

Also worth carrying forward if this is revived:

- A **NaN attributed rate** could enter the reporting vector, where it makes the `partial_cmp` sort a non-total order and defeats the share-based bound that keeps the log line small. Rejected explicitly; the victim set is identical either way.
- **Nothing asserted the log line was emitted at all** — deleting the `tracing::info!` left every test green. A subscriber-capture test was written first and **rejected as inherently flaky**: `tracing::subscriber::set_default` is thread-local while tracing's callsite `Interest` cache is global, so sibling tests hitting the same callsite with no subscriber cache it as `never` process-wide (measured: 1 of 2 events captured under `cargo test --lib cost`, passing when run alone). Replaced with a source-scrape pin.
- One pre-existing test primed a rate limiter with an arbitrary digest that never matched, so the branch it claimed to exercise never ran and both scenarios logged.

### Note for a possible future issue

The **local dashboard tile** may be worth having on its own merits, independent of the log line. It is a few lines (`HostingSnapshot` field + `build_hosting_card` tile), adds nothing to the collector, and closes a real asymmetry: the hosting card shows `budget_evictions_total` and `evictions_of_recently_read_total` but not the cost counter, so an operator on the box cannot tell "the cost trigger is firing" from "it has never fired". If someone wants that, it should go through an issue with approved scope rather than ride along here. Note the 5-column `.g-norms` grid is shared by five cards, so a sixth tile wraps — it should be appended after "Evicted w/ demand" so the #4338 miscalibration alarm keeps its first-row slot.

[AI-assisted - Claude]

PR #4937, opened 2026-07-25
feat(server): show notifications via service worker so mobile browsers work

### LINKED ISSUE #4936: Gateway notifications don't display on mobile browsers (page-level Notification unsupported)
## Problem

Web notifications from any gateway-hosted webapp (River on try.freenet.org is the motivating case) never display on **mobile browsers** (Chrome / Firefox on Android).

The gateway shell (`shell_bridge.js`) shows notifications on behalf of the sandboxed contract iframe via the page-level `new Notification(title, opts)` constructor. Mobile browsers **reject that constructor** — it throws `Illegal constructor. Use ServiceWorkerRegistration.showNotification() instead`. The shell already catches the throw and reports `undeliverable`, so the user gets nothing (falling back to the in-app unread badge only). Desktop works; mobile has never worked.

Since notification *display* lives in the freenet-core shell (the contract app runs in an opaque-origin sandbox and cannot use the Notifications API or register a service worker itself), the fix must live here, not in the contract/app.

## Scope

**Foreground display only** — make notifications actually show on mobile while a shell page is open (and briefly backgrounded), which is the irreducible prerequisite for *any* mobile notification. This is what this issue covers.

**Out of scope:** background push when the app is fully closed. That needs a Web Push relay (VAPID keys + a service subscribing to rooms on Freenet), a separate, larger effort with decentralization/privacy implications. Tracked separately.

## Approach

Register a tiny same-origin service worker (`/freenet-notify-sw.js`, root scope) from the shell and prefer `registration.showNotification()`, keeping the page-level constructor as the desktop fallback. The worker has **no `fetch` handler**, so it never intercepts any request; it exists only to own `showNotification()` and route notification clicks back to the shell (which forwards them to the iframe as the existing `notification_click`). Requires a `worker-src 'self'` addition to the shell CSP.

Approved by @sanity (maintainer) for implementation.

[AI-assisted - Claude]

sanity:
## Problem

Web notifications from any gateway-hosted webapp never display on **mobile browsers** (Chrome / Firefox on Android). River on try.freenet.org is the motivating case: on a phone you never get a new-message notification, even after granting permission.

The gateway shell (`shell_bridge.js`) shows notifications on behalf of the sandboxed contract iframe (the app runs at an opaque origin and can't use the Notifications API itself, so it proxies over `postMessage`). The shell displays them with the **page-level `new Notification(title, opts)` constructor** — which **mobile browsers reject**: it throws `Illegal constructor. Use ServiceWorkerRegistration.showNotification() instead`. The shell already catches the throw and reports `undeliverable`, so mobile users get nothing but the in-app unread badge. Desktop works; mobile never has.

Because display lives in the freenet-core shell (the contract app is sandboxed at an opaque origin and cannot register a service worker itself), the fix has to live here, not in the contract/app.

## Solution

Register a tiny same-origin service worker and show notifications through it on mobile, keeping desktop exactly as it is.

- **New asset `notify_sw.js`**, served at **`/freenet-notify-sw.js`** (origin root). A service worker's scope defaults to its script's directory, so serving it at `/` gives it scope `/`, covering every `/v{1,2}/contract/web/<key>/` shell page with a single registration. The worker has **no `fetch` handler**, so it never intercepts, caches, or alters any request (pages, WASM, the node WebSocket are all untouched). It exists only to own `showNotification()` and to route a notification click back to the shell.
- **`shell_bridge.js`** now tries the page-level `new Notification()` **first** — so **desktop's notification display and click path are unchanged** — and only falls back to the service worker's `showNotification()` **when the constructor throws** (mobile). Registration is lazy (only notification-opted-in users register it) and pre-warmed on opt-in so the first message isn't missed; on desktop that registration stays **dormant** (the constructor succeeds first, so the worker is never used to display). A notification click fires in the worker (not the page), so the worker posts it to the shell, which forwards it to the iframe as the **existing** `notification_click` message. The worker routes the click **only to the originating contract's shell window** (matched by the `/v[12]/contract/web/<key>/` path carried in the notification data), so with multiple contract tabs open a click never focuses the wrong tab or leaks its room tag into another contract's iframe; if that contract has no open tab, the worker reopens its shell rather than the gateway dashboard.
- **CSP**: the shell CSP gains `worker-src 'self'`. Without it, `worker-src` falls back through `child-src` to `script-src 'unsafe-inline'`, which forbids a script URL, so `navigator.serviceWorker.register('/freenet-notify-sw.js')` would be CSP-blocked.

Scope is **foreground display only** — notifications while a shell page is open (or briefly backgrounded), which is the irreducible prerequisite for any mobile notification. Background push when the app is fully closed needs a Web Push relay (VAPID + a service that subscribes to rooms on Freenet) and is deliberately out of scope; it's a separate, larger effort with decentralization/privacy implications.

No wire-format or protocol change; no change to the sandboxed-iframe security model (the app still can't touch the Notifications API, still proxies through the shell, and the same per-contract consent + rate-limit + length-cap gates apply).

**Deployment note:** the worker is served at the origin root `/freenet-notify-sw.js`. Any TLS-terminating proxy in front of a hosted node (e.g. try.freenet.org) must forward that root path to the node, not only `/v{1,2}/contract/...`, or the registration 404s and the fix silently won't engage there. The node itself serves it correctly (covered by a real-router wiring test); this is a proxy-config check, which the manual Android smoke test against try.freenet.org will surface.

## Testing

- `notify_service_worker_route_serves_js` — the handler serves JS and the worker has **no** `fetch` handler (hard invariant).
- `notify_service_worker_route_is_wired` — drives the real `as_router` router with `GET /freenet-notify-sw.js` and asserts 200 + JS content type (catches a mis-registered/shadowed route).
- `shell_page_csp_allows_service_worker` — the shell CSP grants `worker-src 'self'`.
- `bridge_js_registers_notification_service_worker` — source pins: the shell registers the worker, falls back to `showNotification()` on mobile, **tries the page-level constructor BEFORE the worker fallback** (ordering pin, so a refactor can't silently switch desktop to SW-shown notifications), forwards the worker click to the iframe, writes the routing tag as `fnTag`, and gates registration on a secure context.
- `notify_sw.test.mjs` (new) — Node unit test extracting the worker's `pickNotifyClient` client-selection helper via markers: it routes a click **only** to the originating contract's window and returns null (→ open a fresh window) rather than posting the tag to a different contract's tab or the dashboard.
- `notify_service_worker_route_serves_js` — also pins the worker has no `fetch`/`onfetch`/`importScripts`, and reads `fnTag`/`fnUrl` + uses `pickNotifyClient`.
- Existing notification pins (`bridge_js_notification_proxy_invariants`, flood-cap) still pass — consent-key derivation, per-contract consent gate, snooze, rate-limit, and length-cap are unchanged.
- Asset toolchain green: prettier, ESLint (the worker uses only `self.*` worker globals), and the Node asset unit tests (rate-limiter markers intact).

Manual: verify on Android Chrome against try.freenet.org after the next release — send a first message to trigger the "Get notified?" bar, tap Enable, background the tab, and confirm the OS notification appears and tapping it opens the right room.

## Fixes

Fixes #4936


sanity (commented):
## Comprehensive PR Review: #4937

### Summary
- **PR Title:** feat(server): show notifications via service worker so mobile browsers work
- **Type:** feat (fixes broken mobile notification display)
- **CI Status:** green on the reviewed HEAD (see verdict SHA)
- **Linked Issues:** Fixes #4936; follow-up #4939 (cold-start room routing)
- **Review tier:** Full (HTTP server route + CSP — security-relevant surface)
- **Reviewers run:** 3 rounds. R1: code-first + testing + skeptical + big-picture + external Codex. R2 (after fixes): skeptical + code-first + Codex. R3 (final verify): Codex.

---

### Code-First Analysis
**Independent understanding:** the gateway shell displays notifications for the sandboxed contract iframe; it used the page-level `new Notification()` constructor, which mobile browsers reject. The PR adds a root-scoped service worker (`/freenet-notify-sw.js`) and falls back to `showNotification()` when the constructor throws, routing the click back to the originating contract's shell.
**Stated intent:** foreground mobile notification display; desktop unchanged; no request interception; background push out of scope.
**Alignment:** matches. Desktop display+click path is genuinely unchanged (constructor-first with early return; the worker is dormant on desktop). No wire/protocol change; the sandboxed-iframe security model and the existing per-contract consent + rate-limit + length-cap gates are untouched.

---

### Testing Assessment
**Coverage Level:** proportionate to risk.

| Unit (Rust) | ✅ | Route serves JS + no `fetch`/`onfetch`/`importScripts`; real-router wiring test; CSP `worker-src 'self'`; constructor-before-worker ordering pin; `fnTag` contract pin; eager-listener pin |
| Unit (Node) | ✅ | `notify_sw.test.mjs` unit-tests `contractPrefixOf` + `pickNotifyClient`, including the cross-contract substring-bypass regression |
| Integration | N/A | No new integration surface |
| Simulation | N/A | Not a routing/topology change |
| E2E | ⚠️ | Shell Playwright E2E runs but is Desktop-Chrome-only and can't reach the mobile branch; a browser test that stubs `Notification` to throw is a larger lift, deferred |

**Regression test:** present — the substring-bypass case and the per-contract routing are unit-tested.
**Missing tests:** a full browser-level notification E2E (needs a browser + mobile-branch stub) — deferred, noted.

---

### Skeptical Findings
**Risk Level:** low after fixes.

| Multi-tab click misroute + cross-contract tag leak (`clientList[0]`) | Medium | notify_sw.js | FIXED — route only to the originating contract's window |
| Substring `indexOf` bypass via crafted same-contract subpath | Medium | notify_sw.js | FIXED — anchored first-segment equality (`contractPrefixOf`) + regression test |
| Transient SW-registration failure cached for page lifetime | Low | shell_bridge.js | FIXED — clear memo + retry; listener guarded once |
| SW click listener installed lazily → click lost after reload | Low | shell_bridge.js | FIXED — listener installed eagerly at startup |
| `fnTag` uncapped | Low | shell_bridge.js | FIXED — capped to 64 |
| No-fetch pin missed `onfetch` | Low | client_api.rs | FIXED — also forbids `onfetch`/`importScripts` |
| SW response carried no CSP | Low | client_api.rs | FIXED — `default-src 'none'` (safe: worker makes no fetches) |

---

### Big Picture Assessment
**Goal Alignment:** yes — fixes the stated bug at the correct layer (the shell, since the app is sandboxed at an opaque origin), no scope creep, background push not attempted.
**Anti-Patterns Detected:** none. Tests added not weakened; no removed tests/fixes; existing notification security gates all intact.
**Removed Code Concerns:** none.
**Scope Assessment:** focused.

---

### Documentation
- Code docs: complete. The `worker-src` CSP-fallback comment was corrected to the spec-accurate chain (`child-src` absent → `script-src 'unsafe-inline'`, which forbids an external script URL). Dormant-desktop-worker wording is precise.
- Architecture/user docs: n/a. No `.claude/rules` or docs reference the shell/CSP; big-picture reviewer confirmed nothing goes stale.

---

### Recommendations

#### Must Fix (Blocking)
- None outstanding. The two Medium cross-contract-routing findings are fixed and regression-tested.

#### Should Fix (Important)
- None outstanding.

#### Consider (Suggestions)
1. **Cold-start room routing** (Codex P2): a notification clicked after its tab is closed reopens the correct app but not the specific room (the room isn't in the URL for River). Deferred to **#4939** — needs a coordinated freenet-core + River startup handoff. Accepted out-of-scope for this foreground-display PR.
2. **Deployment:** the worker is served at origin root `/freenet-notify-sw.js`; a TLS proxy in front of a hosted node must forward that root path (not only `/v{1,2}/contract/...`) or registration 404s. Verify against try.freenet.org; the manual Android smoke test surfaces it.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** 8cf51b2a8afed8361a218afc54cba58787ce9f65

Every finding from all three review rounds is fixed and regression-tested, or (cold-start room routing) explicitly deferred to #4939 with justification. Desktop behavior is unchanged; the sandboxed-iframe security model and existing notification gates are preserved. Merge once CI is green on the HEAD SHA above.

[AI-assisted - Claude]

sanity:
Applying `review-override` for the rule-review warning on `ensureNotifyServiceWorker`/`notifyRegistrationReady` (async retry + timeout-race branches untested), which the bot explicitly flagged "for reviewer judgment given the explicit rule vs. the precedent for non-pure browser-API code."

Justification:
- **Codebase convention.** These two functions are browser-API-bound (`navigator.serviceWorker`, `window.isSecureContext`, `setTimeout`, `navigator.serviceWorker.ready`). This repo's established pattern (see `shell_bridge_notifications.test.mjs`, `shell_bridge_reload.test.mjs`) is to extract only **pure** helpers for executable Node tests and **source-pin** everything that touches browser/SW globals. Every other browser-API function in `shell_bridge.js` (WS proxy, download, clipboard, navigation, recovery-reload) is guarded the same way. Testing these two would require a full `navigator.serviceWorker` mock harness the repo does not have.
- **Pure logic IS executably tested.** The routing decision (`contractPrefixOf`/`pickNotifyClient`), including the cross-contract substring-spoof case, is unit-tested in `notify_sw.test.mjs`.
- **The flagged branches were adversarially verified.** The multi-model review's skeptical pass explicitly confirmed both: the retry path clears the memo without racing a concurrent caller and never double-adds the listener; the timeout race is `settled`-guarded (no double-resolve), `clearTimeout`s on ready, and backstops a hanging `serviceWorker.ready` with no timer leak.

This is not cherry-picking the easy findings — it's the same source-pin-for-browser-API-code judgment the rest of this file already uses. [AI-assisted - Claude]

Issue #4923, opened 2026-07-24
perf(broadcast): is_delta_efficient gate refuses a delta and sends full state instead — inverted for large contracts


sanity:
## Summary

The fan-out's wire-efficiency gate refuses to compute a delta when the peer's summary is >= 50 % of our state size — and the refusal **falls back to sending the entire state**. Since full state is never smaller than the delta the gate declined to compute, the gate spends bandwidth to save the CPU of computing a delta, and the incentive runs backwards: the larger the state, the more a delta is worth computing, but the more likely a large summary trips the gate.

This is the leading hypothesis for the byte-side findings in the [2026-07-24 production measurement on #3335](https://github.com/freenet/freenet-core/issues/3335#issuecomment-5074264001) — median peer ~6 GB/day, p90 ~22 GB/day, 85 % of byte-weighted broadcast work on contracts with >= 500 KB of state, and only 36 of 2,164 contracts ever changing state.

It is a **hypothesis, not a confirmed root cause**: which of the four full-state paths actually emits the bytes is not yet measured. #4922 adds that instrumentation and should land and report before any fix here is sized.

## The code

`crates/core/src/ring/interest.rs`:

```rust
pub fn is_delta_efficient(summary_size: usize, state_size: usize) -> bool {
    if state_size == 0 { return false; }
    summary_size * 2 < state_size
}
```

The call site returns `Err(DeltaUnavailable::NotEfficient { .. })` when this is false, and `broadcast_queue.rs::broadcast_to_single_peer` answers that `Err` with:

```rust
Err(err) => { /* "Delta computation failed, falling back to full state" */
    (DeltaOrFullState::FullState(new_state.as_ref().to_vec()), false, arm)
}
```

The #4894 / #4901 semantic skip cannot rescue this path: it short-circuits only on byte-equal summaries or an available delta/staleness verdict, and for a contract refused at the gate no delta is ever computed, so there is no verdict to consult. With fan-out to ~19-20 co-hosts, one update to a 2.5 MB contract is ~50 MB of upload from a single peer.

## Candidate fixes

These compose; the first is small and safe, the second needs #4922's data to size.

1. **Consult the staleness verdict on the `Err` path before falling back.** If the peer is semantically converged, skip rather than flood. This reuses the existing #4894 machinery and does not change the gate.
[...]

PR #4938, opened 2026-07-25
feat(broadcast): split the two mystery full-state arms so they name a cause

sanity:
## Problem

#4922 established that **95.6 % of Freenet's wire bytes are whole contract states rather than deltas**, and attributed each byte to one of five fan-out arms. Segmenting strictly on `service.version == 0.2.106`, two arms carry essentially all of it — and neither is actionable as currently reported.

Independently re-derived on this branch from `/mnt/media/freenet-telemetry/logs.jsonl` (1,008 rollup windows, 333 peers, 0.2.106 only):

| arm | % of wire bytes |
|---|---|
| `full_no_summary` | **62.2 %** |
| `full_not_efficient` | **31.9 %** |
| `delta` | 4.5 % |

### `full_no_summary` — the largest single consumer of wire bytes on the network, and three unrelated bugs in one counter

The arm fires whenever *either* side of the summary pair is missing, so it silently merges:

1. **Our own summary missing.** `get_contract_summary` returned `None` — the contract handler timed out at `BROADCAST_CH_TIMEOUT` (10 s), errored, or answered unexpectedly. A **load** problem, and self-reinforcing: more full state → more load → more timeouts.
2. **The peer's summary missing, peer UNTRACKED.** `update_peer_summary` writes through `InterestManager::interested_peers` and is a **silent no-op** for a peer with no entry there. Since #4642 step 9 the fan-out resolves targets from `neighbor_hosting` (advertised co-hosts), while the summary cache still lives in `InterestManager` — the two are maintained by **independent mechanisms** (`NeighborHostingManager::handle_message` never touches `InterestManager`). A target in that state can never cache a summary from a delivery, so **every broadcast to it is full state, permanently**. This is #4442's chicken-and-egg re-opened for a different peer population.
3. **The peer's summary missing, peer TRACKED.** Ordinary cold start; repairs itself on the next delivery.

Those have nothing in common but the symptom, and need three different fixes. One bucket could not drive any of them.

### `full_not_efficient` — a measured contradiction with the evidence compiled out

The gate refuses a delta only when `summary * 2 >= state`. Yet the contract measured tripping it hardest has a 33,804 B summary against a 251,887 B state — **13.4 %**, which should pass comfortably. For it to refuse on a ~254 KB state, the summary would have to be ~127 KB.

`DeltaUnavailable::NotEfficient` has carried `summary_size` and `state_size` since #4922 and renders both in its `Display` impl — but its only reader is a `tracing::debug!`, and `debug!` is compiled out in release builds (`max_level_info`). So in production the gate refuses deltas and nobody can see on what.

## Solution

Observability only. **No broadcast behaviour changes** — every arm sends exactly what it sent before. Deliberately so: this is a live production bandwidth incident, and the root cause is not yet known well enough to safely change the gate.

- Split `FullNoSummary` into `FullNoOurSummary`, `FullNoTheirSummaryUntracked`, `FullNoTheirSummaryTracked`. The untracked/tracked split is decided by an in-memory `get_peer_interest` read — no contract-handler round-trip, on a path that has already committed to putting a whole state on the wire.
- Report the gate's real inputs for `full_not_efficient`: `not_efficient_{summary,state}_bytes_{sum,max}` plus `not_efficient_mean_summary_state_ratio` (null, not 0.0, when the gate never fired).
- **Back-compat:** `full_no_summary_sends` / `_bytes` are republished as the sum of the three new arms, derived from the same window, so existing dashboards and analysis scripts keep working rather than silently reading zero.

### What one release then answers

- `full_no_our_summary` dominant → a contract-handler saturation problem; fix the summarize path, not the gate.
- `full_no_their_summary_untracked` dominant → confirms the structural trap; the fix is to make the summary cache reachable for advertised co-hosts.
- `full_no_their_summary_tracked` dominant → cold start, and the question becomes why peers keep re-entering it.
- `not_efficient_mean_summary_state_ratio` ≈ 0.5+ → the gate is firing on genuinely summary-heavy contracts and the contracts are the problem; well below 0.5 → the gate's inputs are not what its callers assume, which is the real bug.

## Testing

- `no_summary_split_reports_each_cause_and_the_legacy_aggregate` — the split reports separately **and** the pre-split aggregate still adds up.
- `not_efficient_reports_the_gate_inputs_it_refused_on` — sums/maxima/ratio, including that a non-`NotEfficient` arm cannot contribute even if a caller passes sizes.
- `not_efficient_ratio_is_null_when_the_gate_never_fired`.
- `broadcast_to_single_peer_tags_every_payload_arm_pin` extended to require branching on *which* side is missing and to require the `get_peer_interest` consult. **Mutation-verified:** collapsing the split back to a catch-all `_` arm makes the pin FAIL; restoring makes it pass.
- `update_peer_summary_is_a_silent_noop_for_an_untracked_peer` (`ring/interest.rs`) pins the underlying mechanism, contrasts it against the tracked case, and is written to fail loudly with an explanatory message if the trap is ever closed. The pre-existing broadcast-path tests all `register_peer_interest` first, so none of them exercised this state.
- Pin needles are matched against a whitespace-stripped body: rustfmt re-wraps this call whenever its argument list changes width, which broke the original needle during development — exactly the kind of silent pin rot that would otherwise produce a false pass.
- `cargo test -p freenet --lib -- broadcast interest`: 168 passed, 0 failed. `cargo fmt` + `cargo clippy` clean on the touched files.

Also ruled out during this investigation, so it is not re-tried: the InterestSync summary exchange keys contracts by a 32-bit FNV-1a hash, which would poison a contract's cached summary with another's on a collision. Checked across **3,136 distinct contract IDs** observed in production telemetry — **zero collisions**.

Refs #3335, #4922, #4923

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -486,8 +486,13 @@ fn payload_mix_json(
         "not_efficient_state_bytes_max".into(),
         gate.state_bytes_max.into(),
     );
+    // Ratio-of-SUMS (aggregate summary bytes / aggregate state bytes), NOT a
+    // mean of per-send ratios — the two diverge under mixed contract sizes,
+    // and the aggregate form is the one that answers "were the refused bytes
+    // refused on oversized inputs" (poisoned pairs read ~1.0; an honest fleet
+    // reads well under 0.5). Named for what it computes.
     obj.insert(
-        "not_efficient_mean_summary_state_ratio".into(),
+        "not_efficient_summary_to_state_bytes_ratio".into(),
         if gate.state_bytes_sum == 0 {
             serde_json::Value::Null
         } else {
@@ -1012,7 +1017,7 @@ mod tests {
         assert_eq!(json["not_efficient_state_bytes_sum"], 3000);
         assert_eq!(json["not_efficient_summary_bytes_max"], 1400);
         assert_eq!(
-            json["not_efficient_mean_summary_state_ratio"],
+            json["not_efficient_summary_to_state_bytes_ratio"],
             2000.0 / 3000.0
         );
     }
@@ -1034,7 +1039,7 @@ mod tests {
             window.gate_stats(),
             60,
         );
-        assert!(json["not_efficient_mean_summary_state_ratio"].is_null());
+        assert!(json["not_efficient_summary_to_state_bytes_ratio"].is_null());
     }
 
     /// A window with no traffic must emit 0.0, not NaN — `Number::from_f64`
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -8,7 +8,7 @@
 //! on contracts holding >= 500 KB of state, and that 97.9 % of it lands on
 //! contracts whose applies change nothing. What it could NOT determine is
 //! *why* those large states go out whole, because
-//! [`broadcast_to_single_peer`] has **four distinct paths that send FULL
+//! [`broadcast_to_single_peer`] has **six distinct paths that send FULL
 //! STATE** and none of them were separately instrumented:
 //!
 //! 1. [`PayloadArm::FullDeltaSuppressed`] — the `delta_incompat` memo is armed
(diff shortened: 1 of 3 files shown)


sanity:
## Review gate record (incident release train)

**External model pass:** Codex unavailable (out of credits, confirmed by owner). Per the multi-model fallback, three diverse Claude review lenses were spawned (skeptical / code-first / telemetry-consumer). All three failed to deliver reports despite four prompts each — a session-infrastructure failure, recorded honestly rather than papered over.

**Substituted with an independent non-author review** (this session did not write this PR). Verified directly against the committed diff:

- **Arm selection** (`broadcast_queue.rs`): `(None,_)` → `FullNoOurSummary` before `(Some,None)` → tracked/untracked split — ordering convention documented in-code; both-missing resolves to ours-missing deliberately. The untracked probe (`get_peer_interest`) reads the same structure `update_peer_summary` writes through, so the classification matches the mechanism it names. In-memory DashMap read, no handler round-trip, on a path already committed to a full-state send. ✔
- **No behavior change:** every arm sends exactly what it sent before; the suppression guard remains scoped `(Some(_),Some(_))`. ✔
- **Window atomicity:** the gate-input sums/maxima live on the same mutex-guarded `Window` as the arm counters; the legacy `full_no_summary_*` aggregate is derived from the same drained window, so it cannot diverge from its parts. ✔
- **Null-vs-zero:** ratio publishes `Null` when the gate never fired; pinned by test. ✔
- **Pins:** needles match a whitespace-collapsed function body (rustfmt-proof); the arm-split pin fails if the split collapses to a catch-all. ✔

**One finding, fixed in `061c0efa`:** the ratio field was named `mean_summary_state_ratio` but computes a **ratio-of-sums** (aggregate summary bytes / aggregate state bytes). The two diverge under mixed contract sizes; an analyst trusting the name would mis-model the field. Renamed to `not_efficient_summary_to_state_bytes_ratio` with semantics documented at the construction site. Unreleased field, so the rename is free now and breaking later.

**Known non-blocking follow-up** (pre-existing, documented in the PR's own comments): the `FullNoOurSummary` path still sends `sender_summary_bytes = unwrap_or_default()` — an empty summary the receiver caches indistinguishably from a real one. The state-as-summary root-cause fix in flight adds a receiver-side floor; absent-vs-empty disambiguation remains open after it.

Merging on green CI at `061c0efa` as the measurement car of the 0.2.107 incident train (root cause: #3335 state-as-summary self-poisoning, fix in flight).

[AI-assisted - Claude]

PR #4940, opened 2026-07-25
feat(telemetry): emit BroadcastComplete with the per-fan-out delta/full-state split

sanity:
## Problem

`UpdateEvent::BroadcastComplete` is defined (`tracing/event_kind.rs:969`), matched in four places, and serialized by `tracing/telemetry.rs:1713` into a `{"type":"broadcast_complete", ..., "delta_ratio": ...}` event — but **nothing ever constructed it**, so that arm never fired. The same was true of the counters #4922 added to `InterestManager`: `delta_sends` / `full_state_sends` / `delta_bytes_saved` are incremented on every send, but `stats()` has no caller anywhere in `crates/`, so they were tallied in memory and never read.

Both read as live telemetry to anyone grepping for them. Neither produced a byte.

This matters because `state_size` — the only size the per-fan-out update telemetry carried — reports the post-apply **full** state whether a 15 KB delta or a 340 KB whole state crossed the wire. Nothing in the per-fan-out event stream could tell those apart, which is what blocked explaining a River room reporting ~370 KB broadcast payloads against a ~320-339 KB state.

## Solution

Each fan-out builds one `FanoutTally`; its per-target sends record into it; the fan-out emits one `BroadcastComplete` carrying the **same transaction** as its `BroadcastEmitted`, so the two join on `id` — `BroadcastEmitted` gives targets enqueued, `BroadcastComplete` gives the arms actually delivered.

`bytes_saved` is what closes the ambiguity: `state_size * (delta_sends + full_state_sends) - bytes_saved` is the real payload total the fan-out put on the wire.

**Per-fan-out, not cumulative.** The `InterestManagerStats` counters are process-lifetime totals across every contract and peer, so differencing them around a fan-out would attribute concurrent fan-outs to whichever sampled last (production runs one detached task per (contract, peer)). A per-fan-out accumulator is the only way to get correct attribution — so this deliberately does **not** read `stats()`, and `stats()` remains uncalled.

**Why a second tally is defensible here.** It duplicates information the canonical counters hold, which is the `.claude/rules/bug-prevention-patterns.md` "manually-mirrored telemetry counters" failure (#4009 / #4010). The defence is that every tally write sits immediately beside the canonical `record_delta_send` / `record_full_state_send` it shadows, at both recording sites, and pins assert that adjacency. There is no third write site.

**Retirement is a `Drop` obligation** (`FanoutTarget`), not an explicit call: a queue entry leaves three ways — sent, superseded by replace-on-dedup, evicted at capacity — and `broadcast_to_single_peer` has several early returns. A forgotten explicit call would strand the fan-out so its event never fired, silently and with no behavioral symptom.

**Cannot stall a broadcast.** `register_events` is `try_send` internally and drops on a full channel (#4466/#4467), so the four `.send().await`-in-a-fan-out incidents are not reachable. Superseded/evicted targets are retired *after* the queue lock is released, so completing a fan-out never spawns under the lock every broadcast contends on.

Observability only — no change to what is broadcast, to whom, or in what form.

### Relationship to #4922 (please read before reviewing scope)

#4922 is **not** just a counting half; it shipped complete in v0.2.106. Its `broadcast_payload_mix.rs` already reports per-arm sends and **real wire bytes** on a 60s per-node rollup, naming which of the four full-state fallbacks fired (`full_delta_suppressed` / `full_not_efficient` / `full_compute_failed` / `full_no_summary`) plus `top_contracts_by_full_state_bytes`.

That is the better tool for "what is this node's byte mix, and why", and **it is the measurement #4923 is blocked on** — it should be read before this PR is treated as unblocking anything. This PR adds a strictly narrower, complementary view: per-fan-out identity, which the rollup deliberately does not carry, so it cannot answer "this particular update to this contract went out as N deltas and M full states". Neither replaces the other, and the module docs say so.

## Testing

`FanoutTally` unit tests (creation guard, exactly-once emission, zero-target and zero-send fan-outs, saturating `bytes_saved`, 32-thread concurrent retirement); the tally rides the real `record_streaming_delivery` gate in the existing #4235 drop test, asserting a dropped stream is **not** counted as a send; source-scrape pins for co-location and both emission paths; and an end-to-end assertion in `test_subscription_broadcast_propagation` requiring one `BroadcastComplete` per `BroadcastEmitted` **and** a non-zero recorded send total.

Every test was mutation-verified — the production change it pins was reverted and the test confirmed to fail:

| # | Mutation | Caught by |
|---|---|---|
| M1 | Drop tally delta write in `record_delivery_to_interest` | co-location pin |
| M2 | Drop tally full-state write | co-location pin |
| M3 | `BroadcastEntry` holds `Arc<FanoutTally>` instead of `FanoutTarget` | compiler (not the pin — noted below) |
| M4 | Retire `FanoutTarget`s *under* the queue lock | queue pin |
| M5 | Replace-on-dedup keeps the stale target | queue pin |
| M6 | Remove the creation guard (emit before all targets dispatched) | 5 tally unit tests |
| M7 | Production fan-out builds the tally but never emits | emission pin |
| M8 | Simulation fan-out never emits | **simulation integration test** |
| M9 | Simulation fan-out stops writing the tally | sim pin **and** integration test |
| M10 | `BroadcastEmitted` mints its own tx (breaks the join) | emission pin |

M3 is caught by the type system rather than the pin; that assertion is belt-and-braces over a compiler-enforced property, which is stated honestly rather than claimed as pin coverage.

**Mutation testing found two real defects in my own pins**, both fixed in the second commit: two needles matched their own source (`include_str!` pulls the test module back in), so M5 initially passed against the bug. They are now split with `concat!`, matching the existing `record_delivered` precedent.

### One pre-existing test repaired

`egress_paths_gated_on_is_contract_broken` anchored on `broadcast_queue.enqueue` as one contiguous string; adding an argument made rustfmt split the call, so the needle stopped matching and the pin failed for a formatting reason rather than a real one. Re-anchored on `.enqueue(` and verified it resolves to the identical statement (gate offset 687 < enqueue offset 2651), so the ordering assertion it exists to make is intact.

### Checks

- `cargo fmt -- --check` — clean
- `cargo clippy --locked -- -D warnings` — clean
- `cargo clippy --locked -p freenet --features trace-ot -- -D warnings` — clean
- `cargo nextest run -p freenet --lib --features "simulation_tests,testing"` — 4298 passed, 0 failed
- `cargo test -p freenet --lib` — 4296 passed, 0 failed
- `test_subscription_broadcast_propagation` — passes

Note: under `cargo test --test-threads 8` with the simulation feature, three tests in files this PR does not touch (`connection_manager`, `home_page`, `rolling_rtt_stats`) fail on shared global state; they pass in isolation and under nextest, which is the runner CI uses and which `.claude/rules/testing.md` requires for exactly this reason. Flagging rather than folding a fix into this PR.

## Scope

Parent measurement issue #3335. Related: #4922 (merged, v0.2.106), #4923 (blocked on #4922's data).

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/pool_tests/non_idempotent_detector_tests.rs
+++ b/crates/core/src/contract/executor/pool_tests/non_idempotent_detector_tests.rs
@@ -213,8 +213,13 @@ fn egress_paths_gated_on_is_contract_broken() {
     let heal_gate = body
         .find("is_contract_broken")
         .expect("handle_sync_state_to_peer must gate on is_contract_broken");
+    // Anchor on the call, not the receiver: rustfmt splits
+    // `self.broadcast_queue.enqueue(..)` across lines once the argument list
+    // grows, so a receiver-shaped needle silently stops matching and this pin
+    // fails for a formatting reason rather than a real one. Same precedent as
+    // the `record_delivered` needle in broadcast_queue.rs.
     let heal_send = body
-        .find("broadcast_queue.enqueue")
+        .find(".enqueue(")
         .expect("handle_sync_state_to_peer must still enqueue the heal send");
     assert!(
         heal_gate < heal_send,
--- a/crates/core/src/node/network_bridge.rs
+++ b/crates/core/src/node/network_bridge.rs
@@ -16,6 +16,7 @@ use tokio::sync::mpsc::{self, Receiver, Sender};
 
 use crate::message::{NetMessage, NodeEvent};
 
+pub(crate) mod broadcast_fanout_tally;
 pub(crate) mod broadcast_payload_mix;
 pub(crate) mod broadcast_queue;
 mod handshake;
--- a/crates/core/src/node/network_bridge/broadcast_fanout_tally.rs
+++ b/crates/core/src/node/network_bridge/broadcast_fanout_tally.rs
@@ -1,0 +1,417 @@
+//! Per-fan-out delta-vs-full-state tally behind `UpdateEvent::BroadcastComplete`.
+//!
+//! ## Why
+//!
+//! `UpdateEvent::BroadcastComplete` has existed since #3622-era telemetry work
+//! and `telemetry.rs` has always known how to serialize it (`"type":
+//! "broadcast_complete"`, including a derived `delta_ratio`), but **nothing ever
+//! constructed it**, so the arm never fired. The same was true of
+//! [`InterestManagerStats`]: #4922 added `delta_sends` / `full_state_sends` /
+//! `delta_bytes_saved` counters to `InterestManager`, but `stats()` has no
+//! caller, so those counters were tallied in memory and never read. This module
+//! is the missing reporting half for the per-fan-out view (#4923).
+//!
+//! The gap mattered because `state_size` — the only size the update telemetry
+//! carried — logs the post-apply FULL state regardless of whether a small delta
+//! or the whole state crossed the wire. A River room reporting ~370 KB
+//! broadcast payloads against a ~320-339 KB state and a ~15 KB delta could not
+//! be explained from telemetry alone: nothing distinguished "we sent a delta"
+//! from "we sent the entire state".
+//!
+//! ## Relationship to the #4922 payload mix
+//!
+//! [`PayloadMix`] answers the same question at a coarser grain and with a finer
+//! *cause* breakdown: a 60-second per-node rollup of per-arm sends and real wire
+//! bytes, naming which of the four full-state fallbacks fired. It is the right
+//! tool for "what is this node's byte mix, and why". It deliberately does NOT
+//! carry per-fan-out identity, so it cannot answer "this particular update to
+//! this contract went out as N deltas and M full states".
+//!
+//! This tally is that second view, and only that: one event per fan-out,
+//! carrying the fan-out's transaction so it joins directly against the
+//! `BroadcastEmitted` emitted for the same fan-out. The two are complementary;
+//! neither replaces the other.
+//!
+//! ## Why a tally rather than reading the counters
+//!
+//! [`InterestManagerStats`] counters are process-lifetime monotonic totals
+//! across every contract and peer. Differencing them around a fan-out would
+//! attribute concurrently-running fan-outs (other contracts, other peers) to
+//! whichever one happened to sample last, because the production path runs one
+//! detached task per (contract, peer) with bounded concurrency. A per-fan-out
+//! accumulator is the only way to get correct attribution.
+//!
+//! To keep that from becoming a second, independently-rotting mirror — the
+//! failure mode in the "Manually-mirrored telemetry counters" row of
+//! `.claude/rules/bug-prevention-patterns.md`, and the #4009 / #4010 precedent —
+//! every tally write sits immediately next to the canonical
+//! `InterestManager::record_delta_send` / `record_full_state_send` call it
+//! shadows, and source-scrape pins in `broadcast_queue.rs` and
+//! `p2p_protoc/broadcast.rs` assert that co-location at both recording sites.
+//!
+//! [`InterestManagerStats`]: crate::ring::interest::InterestManagerStats
+//! [`PayloadMix`]: super::broadcast_payload_mix::PayloadMix
+
+use std::sync::Arc;
+
+use freenet_stdlib::prelude::ContractKey;
+use parking_lot::Mutex;
+
+use crate::message::Transaction;
+use crate::node::OpManager;
+use crate::tracing::NetEventLog;
+
+/// One fan-out's accumulated delta-vs-full-state split.
+///
+/// Returned by [`FanoutTally::finish_one`] to exactly one caller — the one that
+/// retires the last outstanding target — so a fan-out can never emit twice.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct FanoutSummary {
+    pub tx: Transaction,
+    pub key: ContractKey,
+    pub delta_sends: usize,
+    pub full_state_sends: usize,
+    pub bytes_saved: u64,
+    pub state_size: usize,
+}
+
+/// Mutable half of [`FanoutTally`], behind one lock.
+///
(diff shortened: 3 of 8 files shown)


sanity:
## Closing: scale-safe, but the question it answers is answered, and the join it is built around breaks on exactly the peers that matter

Branch preserved as **`telemetry-broadcast-complete`** (head `484cfc08`). It still merges cleanly onto `main` — this is not a rebase-cost decision.

### What it was for

`UpdateEvent::BroadcastComplete` was defined and serialized but never constructed, and `state_size` reported the post-apply *full* state whether a 15 KB delta or a 340 KB whole state crossed the wire. Nothing in the per-fan-out stream could tell those apart, which blocked explaining a River room reporting ~370 KB broadcast payloads against a ~320-339 KB state.

### Scale, measured (this was the gating question)

Ian's constraint: telemetry must not overload the collector even at 10x the current fleet. I measured a 200 MB slice of a live rotation of `/mnt/media/freenet-telemetry/logs.jsonl` — 280,944 records, 841 distinct peers.

**Collector today** (~1,700 peers): 1,405 records/MB, 5.24 GB per rotation every ~85 min → **~7.37M records/rotation, ~1,445 records/s, ~88.8 GB/day**.

**This PR's addition.** `FanoutTally::new` is called once per fan-out, outside the per-target loop (`p2p_protoc/broadcast.rs:531` vs `:539`); `finish_one` returns `Some` exactly once (`outstanding` starts at `targets + 1`, plus a `finished` flag); `FanoutTarget` is not `Clone` and has one construction site. So emission is **1:1 with the existing `update_broadcast_emitted`** and, given restart/queue-drop paths that strand targets, structurally **≤ 1:1** — it cannot amplify. Zero-target fan-outs return early at `broadcast.rs:270` before a tally exists. Only one of the two fan-out paths compiles into a shipped binary (`simulation_tests` is not a default feature).

Measured against the live sample, and synthesizing `broadcast_complete` records through the exact serializer at `telemetry.rs:1713`:

| | value |
|---|---|
| `update_broadcast_emitted` today | 1.54% of records, 1.57% of bytes, mean 723 B |
| `broadcast_complete` synthesized | mean **650 B** (0.90x — it drops the two `PeerKeyLocation` strings) |
| **Added load** | **+1.54% records, +1.41% bytes** |
| At ~1,700 peers | +1.26 GB/day on top of 88.8 GB/day |
| At ~17,000 peers | +12.6 GB/day on top of ~888 GB/day |

**So it passes the scale test.** It is a fixed proportional addition, not the traffic x peers amplification shape. Scale is not why this is closed.

### Why it is closed

**1. The motivating question is answered by the 60s rollup.** After #4922 + #4938 + #4956 + #4962, `broadcast_payload_mix` publishes per-arm sends and *real wire bytes* across seven arms, `full_state_byte_share`, `top_contracts_by_full_state_bytes` and `top_contracts_by_not_efficient_bytes` with reconciliation residuals, the gate's inputs, and #4962's `tracked_missing_<reason>` cause tags. "Deltas or full state, how many bytes, for which contracts, and why full state" is answered without this event.

**2. The join is worth close to nothing.** The PR's unique claim is per-fan-out identity via a shared transaction. But that `update_tx` is minted in the same block that emits both events, and the pre-existing comment at `p2p_protoc/broadcast.rs:527-529` is still accurate: it is synthetic and cannot be correlated with downstream `BroadcastReceived`/`BroadcastApplied`. The join therefore means "these two events came from the same twenty lines" — identical information to adding `broadcasted_to` as a field on `BroadcastComplete`. The join that would matter, to the originating update op, still does not exist and this PR does not create it.

**3. On busy peers it displaces rather than adds, and the join breaks asymmetrically.** `MAX_EVENTS_PER_SECOND = 10` per node (`tracing/telemetry.rs:173`), and 3.11% of records in the sample already land in `(peer, second)` buckets at or over that cap. Both events are `EventPriority::Operational` and compete for the same budget, so on a saturated peer the new event silently drops other operational telemetry. Worse: `BroadcastComplete` fires from the last `FanoutTarget::drop`, which can be up to `STREAM_COMPLETION_TIMEOUT = 120s` (`broadcast_queue.rs:31`) after its partner. The pair therefore gets independent admission decisions in different rate-limit windows — so the join degrades most on peers in a broadcast storm, which are exactly the peers whose delta/full-state split you want to read.

**4. Semantics drifted under #4957.** The gate is now post-compute (`interest.rs:1817`), so a `full_state_send` is frequently the genuinely optimal payload rather than a fallback failure. `delta_ratio` no longer reads as a waste metric, and the event's own rustdoc at `event_kind.rs:969` ("no cached summary or delta failed") is stale.

**5. The 1:1 property is untested where it ships.** `crates/core/tests/simulation_integration.rs` is `#![cfg(feature = "simulation_tests")]`, so the `broadcast_complete == broadcast_emitted` assertion exercises only the inline `broadcast_state_to_peers` path, which never ships. The production `BroadcastQueue` + `FanoutTarget` + `Drop` + spawn path — where dedup-swap, capacity eviction and RAII retirement actually live — is covered only by `FanoutTally` unit tests that never touch the queue, plus source-scrape pins.

For the record, the engineering in the branch held up under an adversarial audit: no double-emission, no memory leak on strand (the guard `Arc` drops unconditionally), no spawn under either lock, no `.await` across the `parking_lot` lock, and streaming vs inline delivery are mutually exclusive so nothing double-counts.

### Two real gaps this PR would have filled — kept, in #4979

The audit found two numbers `main` genuinely does not have, and they survive this closure:

1. **No per-contract *delta* accounting.** `contract_full_state_bytes` is gated on `arm.is_full_state()`, so there is a numerator with no denominator — you cannot compute any contract's delta:full ratio or its `bytes_saved`.
2. **Converged skips and dropped sends are measured nowhere.** The rollup counts delivered sends only, so targets skipped as already-converged (`broadcast_queue.rs:912`) are invisible. Given #4956's finding that 67% of applies are no-ops carrying 80% of bytes, "how many targets did this fan-out skip" is a directly relevant number nobody has.

Both are cheap counters on the existing fixed-cadence rollup — no new event stream, no RAII retirement machinery, and no exposure to the per-second cap. Filed as #4979.

[AI-assisted - Claude]

PR #4941, opened 2026-07-25
feat(telemetry): publish the efficiency gate's inputs on the NotEfficient arm

sanity:
## Problem

The 2026-07-25 measurement on v0.2.106 (`broadcast_payload_mix`, 1,319 peers / 291 emitting / 7,621 windows, segmented on `service.version == 0.2.106`) put **full state at 95.58% of fleet wire bytes**. Within that, `full_not_efficient` is **38.1% fleet-wide** and **83.1% of the River room's** full-state bytes.

That last number does not reconcile with the room itself. `is_delta_efficient` is `summary * 2 < state` — it refuses when the summary is >= 50% of state. The room measures **33,804 B of summary against a 251,887 B state = 13.4%**, far under the threshold, and structurally so: the per-message marginal is +67 B of summary per +521 B of state, so growth moves the ratio *away* from the threshold. For the gate to refuse on a ~254 KB state, `their_summary` would have to be ~127 KB — roughly 4x anything reachable. No constructible state of this contract trips this gate, yet that is where its bytes are attributed.

Both measurements are solid, so **the gate's inputs are not what anyone assumes** — and today they cannot be checked. `DeltaUnavailable::NotEfficient` has always carried `summary_size` and `state_size` and renders them in its `Display`, but its only consumer is a `tracing::debug!` in `broadcast_queue.rs`, and **`debug!` is compiled out of release builds** (`max_level_info`). Hence zero occurrences in production telemetry. This repo has been bitten by that before.

## Solution

Record the gate's two inputs into the existing 60s `broadcast_payload_mix` rollup as fixed-cardinality aggregates: **count, sum, min and max for both sizes**, plus a separate count of `state_size == 0` refusals.

**Why min/max and not just a mean** — the extremes are what separate the candidate explanations, and a mean collapses all of them into one number:

| observation | implicates |
|---|---|
| `summary_min == summary_max` at a fixed value | a defaulted/empty summary being sized wrong |
| `summary_max` far above anything reachable | a stale or wrong-peer summary in the cache |
| `state_min` at/near 0 | `is_delta_efficient`'s explicit `state_size == 0` early return |
| sizes matching the measured 13.4% ratio | the gate is **not** refusing on these inputs — the arm is mis-attributed at the classification site |

The zero-state case gets its own counter because the ratio cannot express it: anything over zero state is trivially ">= 50%", so it lands on this arm even though the ratio is not the reason.

**A mis-attribution tripwire, `not_efficient_ratio_would_have_passed`.** The min/max fields are computed *independently* across the two dimensions, so a large `summary_max` and a small `state_min` may come from different sends and must not be combined into a ratio. That leaves the fourth row above — "the arm is mis-attributed" — unsettled, and it is the row a fix would most easily be built in the wrong direction against.

`is_delta_efficient` returns `true` for `state != 0 && summary * 2 < state`, so a send with those sizes should have produced a **delta** and can never legitimately reach this arm. Counting them is therefore a direct test of that hypothesis: **the counter must be 0**, and any other value proves the sizes reaching this arm are not the sizes the gate evaluated. It is evaluated per sample, where the pair is still together, which is exactly what the independent extremes cannot do.

Chosen over rate-limited raw exemplars, which were the other way to close the same pairing gap: exemplars answer it indirectly and cost serialization surface, while this answers it outright in one number. The predicate deliberately mirrors the gate rather than calling it, so it stays a check on the *recorded* sizes even if the gate's own definition later changes.

**Where it is recorded.** Sizes are captured at the refusal and carried to the delivery site, not recorded at the refusal, so the distribution is gated on the same real-delivery condition as the arm's existing send and byte counts — the three always describe one set of sends, which is the #4903 round-3 accounting rule this module already follows.

**An idle window publishes `null`, not `0`.** A zero minimum summary is a specific and alarming claim ("the gate saw an empty summary"), and emitting it for a window with no refusals would manufacture exactly the finding this instrumentation exists to test for.

**Cardinality:** eight scalars per window, keyed by nothing. Flat regardless of traffic — this is per-send data feeding a per-node rollup, so anything keyed by contract or peer would be an amplification vector.

**Concurrency:** no new channel, no new lock, no send of any kind. The write rides the existing per-window `parking_lot::Mutex` that `record_delivered` already takes, inside the same critical section.

Observability only. **Nothing about what is sent changes, and the gate itself is untouched** — the mechanism is not yet understood, and changing broadcast behaviour on a hypothesis during a live bandwidth incident is the wrong move.

### Why not promote the `debug!` to `info!`

That was the stated fallback, and I deliberately did not do it. The call site fires **once per refused delta send** — 84,394 sends in the measured sample — on every peer in the fleet, on exactly the path that is already saturating uplinks. An unthrottled per-send `info!` risks making the incident worse in order to observe it. The aggregate answers the same question with better statistics and lands in the event the analysis already parses. If raw exemplars are wanted too, a rate-limited sample would be the safe shape; say the word.

## Testing

Eight unit tests over the aggregate (inputs reach the rollup; bimodal extremes survive; zero-state counted separately; idle window publishes null; sizes ignored for other arms; window drains between rollups) plus a source-scrape pin that the `NotEfficient` arm destructures the sizes rather than matching with `..`.

Every one mutation-verified (13 mutations) — production change reverted, test confirmed to FAIL, restored:

| # | Mutation | Caught by |
|---|---|---|
| N1 | Recorder drops the sizes entirely | 4 aggregate tests |
| N2 | Arm guard removed, any arm pollutes the distribution | `gate_sizes_are_ignored_for_other_arms` |
| N3 | Publish `0` instead of `null` on an idle window | `window_with_no_refusals_publishes_null_not_zero` |
| N4 | Window not drained, outlier smears across windows | `gate_input_distribution_resets_between_windows` |
| N5 | `min` seeded at 0 (fabricates an empty-summary finding) | 2 aggregate tests |
| N6 | Zero-state refusals no longer counted separately | `zero_state_refusals_are_counted_separately` |
| N7 | `NotEfficient` arm back to matching with `..` | `not_efficient_arm_captures_the_gate_sizes_pin` |
| N8 | One of the two delivery sites silently drops `gate_sizes` | `broadcast_to_single_peer_tags_every_payload_arm_pin` |
| N9 | Recorder moved to a process-global static | compiler + same pin |
| N10 | Tripwire never increments | `sizes_that_should_have_passed_the_gate_trip_the_misattribution_counter` |
| N11 | Tripwire predicate inverted (fires on genuine refusals) | `genuine_refusals_including_the_boundary_do_not_trip_the_counter` |
| N12 | Strict `<` relaxed to `<=` (boundary misread as mis-attribution) | same |
| N13 | Zero-state refusals also counted as would-have-passed | `zero_state_refusals_are_counted_separately` |

### A pin I broke and repaired

Growing `record_delivered`'s argument list pushed rustfmt to split the call across five lines, which broke `broadcast_to_single_peer_tags_every_payload_arm_pin` — it matched a contiguous single-line needle, and its sibling assertion matched a literal newline plus an exact indentation width. Both now match a whitespace-stripped copy of the function body, so they are independent of rustfmt's layout choices, and both were mutation-verified to still catch what they exist to catch (N8, N9). A pin that fails on reformatting is worse than no pin: it trains the next person to loosen the assertion.

### Checks

- `cargo fmt -- --check` — clean
- `cargo clippy --locked -- -D warnings` — clean
- Full lib suites under nextest — see CI

## Scope

Parent measurement issue #3335. Builds directly on #4922 (merged, v0.2.106), whose rollup this extends. Provides the data #4923 is blocked on — it is explicit that "#4922 adds that instrumentation and should land and report before any fix here is sized", and this closes the one input #4922 does not capture.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -164,6 +164,9 @@ struct Window {
     /// an 11-contract window look perfectly reconciled while 10 % of its bytes
     /// were unaccounted for.
     attribution_dropped_bytes: u64,
+    /// Distribution of the efficiency gate's two INPUTS on the refusals it
+    /// caused. See [`NotEfficientSizes`].
+    not_efficient: NotEfficientSizes,
 }
 
 impl Default for Window {
@@ -174,6 +177,103 @@ impl Default for Window {
             contract_full_state_bytes: HashMap::new(),
             attribution_dropped_sends: 0,
             attribution_dropped_bytes: 0,
+            not_efficient: NotEfficientSizes::default(),
+        }
+    }
+}
+
+/// The two sizes [`is_delta_efficient`][ide] compared when it refused to
+/// compute a delta.
+///
+/// [ide]: crate::ring::interest::is_delta_efficient
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct GateSizes {
+    /// `their_summary.as_ref().len()` — the cached summary for the TARGET peer.
+    pub summary_size: usize,
+    /// The `our_state_size` argument the fan-out passed to `compute_delta`.
+    pub state_size: usize,
+}
+
+/// Aggregated gate inputs across one window's `FullNotEfficient` refusals.
+///
+/// ## Why this exists
+///
+/// The gate is `summary * 2 < state`, so it refuses when the summary is >= 50 %
+/// of the state. The 2026-07-25 measurement found `full_not_efficient` at 38 %
+/// of fleet wire bytes and 83 % of the River room's — against a room whose
+/// summary measures 33,804 B on a 251,887 B state (13.4 %), well under the
+/// threshold, and structurally so: the per-message marginal is +67 B of summary
+/// per +521 B of state, so growth moves the ratio further from the threshold,
+/// not toward it. No reachable state of that contract should trip this gate,
+/// yet the gate is where its bytes are attributed.
+///
+/// So the gate's inputs are not what anyone assumes, and this records what they
+/// actually were. `DeltaUnavailable::NotEfficient` has always carried both
+/// sizes and renders them in its `Display`, but its only consumer is a
+/// `tracing::debug!`, and `debug!` is compiled out of release builds
+/// (`max_level_info`) — which is exactly why there are zero occurrences of it
+/// in production telemetry.
+///
+/// ## Why min and max, not just a mean
+///
+/// A mean cannot separate the candidate explanations; the extremes can:
+///
+/// * `summary_min == summary_max` at some fixed value suggests a defaulted or
+///   empty summary being sized wrong, not a real one.
+/// * A `summary_max` far above any reachable summary suggests a stale or
+///   wrong-peer summary in the cache.
+/// * A `state_min` at or near zero implicates [`is_delta_efficient`][ide]'s
+///   explicit `state_size == 0` early return, which also lands on this arm even
+///   though "summary >= 50 % of state" is not the reason —
+///   [`zero_state_samples`](Self::zero_state_samples) counts that case outright.
+/// * Sizes consistent with the measured 13.4 % ratio would mean the gate is NOT
+///   refusing on these inputs and the arm is mis-attributed at the
+///   classification site instead.
+///
+/// Fixed cardinality: eight scalars per window, keyed by nothing. This is
+/// per-send data feeding a per-node rollup, so anything keyed by contract or
+/// peer would be an amplification vector.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+struct NotEfficientSizes {
+    /// Refusals recorded this window. Zero means the other fields are
+    /// meaningless and they are published as `null`.
+    samples: u64,
+    summary_sum: u64,
+    summary_min: u64,
+    summary_max: u64,
+    state_sum: u64,
+    state_min: u64,
+    state_max: u64,
+    /// Refusals where `state_size == 0`, i.e. the gate's explicit zero-state
+    /// early return rather than the ratio test. Counted separately because the
+    /// two have completely different causes and the ratio cannot distinguish
+    /// them (anything over zero state is ">= 50 %").
+    zero_state_samples: u64,
+}
+
+impl NotEfficientSizes {
+    fn record(&mut self, sizes: GateSizes) {
+        let summary = sizes.summary_size as u64;
+        let state = sizes.state_size as u64;
+        if self.samples == 0 {
+            // First sample seeds both extremes; seeding min at 0 instead would
+            // pin it there forever and report a zero-byte summary that never
+            // happened.
+            self.summary_min = summary;
+            self.summary_max = summary;
+            self.state_min = state;
+            self.state_max = state;
+        } else {
+            self.summary_min = self.summary_min.min(summary);
+            self.summary_max = self.summary_max.max(summary);
+            self.state_min = self.state_min.min(state);
+            self.state_max = self.state_max.max(state);
+        }
+        self.samples = self.samples.saturating_add(1);
+        self.summary_sum = self.summary_sum.saturating_add(summary);
+        self.state_sum = self.state_sum.saturating_add(state);
+        if state == 0 {
+            self.zero_state_samples = self.zero_state_samples.saturating_add(1);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


sanity:
## Closing: superseded by #4938, and the tripwire is now arithmetically invalid against `main`

Branch preserved as **`telemetry-notefficient-sizes`** (head `8df9f5d2`) — nothing here is lost, and the reasoning below is the reason not to rebase it.

### What this PR was for

`full_not_efficient` was 38.1% of fleet wire bytes and 83.1% of the River room's, against a room measuring 33,804 B of summary on a 251,887 B state (13.4%) — far under the gate's `summary * 2 < state` threshold. The gate's inputs were invisible because `DeltaUnavailable::NotEfficient`'s only reader was a `tracing::debug!`, compiled out in release. This PR published those inputs as per-window aggregates.

### That instrumentation landed 70 minutes later, in #4938

#4938 (`1912fd4a`, merged 2026-07-25T23:12Z) added to the same rollup: `not_efficient_summary_bytes_sum`, `not_efficient_state_bytes_sum`, `not_efficient_summary_bytes_max`, `not_efficient_state_bytes_max`, `not_efficient_summary_to_state_bytes_ratio`, and `top_contracts_by_not_efficient_bytes`.

`main` lacks exactly seven of this branch's fields: `not_efficient_samples`, the two `_min`s, the two `_mean`s, `zero_state_sends`, and `ratio_would_have_passed`. Two of those are dead on arrival — `samples` is identically `full_not_efficient_sends`, so both means are already computable from published fields.

Meanwhile **this branch lacks two of `main`'s fields** (the ratio-of-sums and the per-arm per-contract attribution added by #4956). Rebased as-is it would be a net regression.

### The question it was built to answer is answered

The ratio-of-sums settled it empirically ([#4956 comment](https://github.com/freenet/freenet-core/issues/4956#issuecomment-5084560592)):

| version | not_efficient sends | mean state | mean summary | ratio |
|---|---|---|---|---|
| 0.2.107 | 21 | 831.0 KB | 830.9 KB | **1.000** |
| 0.2.108 | 2,459 | 618.1 KB | 618.0 KB | **1.000** |
| 0.2.109 | 281 | 551.5 KB | 551.4 KB | **1.000** |

with `not_efficient_summary_bytes_max == not_efficient_state_bytes_max == 839 KB` exactly. Some contract's `summarize_state` returns something state-sized, which permanently disabled delta computation for it. Fixed on both sides: #4943 (stop relabeling our full state as our summary) and #4957 (judge the artifact actually shipped, not a contract-controlled proxy). The arm went 31.9% of wire bytes on 0.2.106 → 16.0% on 0.2.109 → absent from 0.2.110+ top arms.

Notably `ratio_would_have_passed` would have read **0** throughout — mis-attribution was falsified by the ratio, not by the tripwire.

### The blocking reason: the tripwire is now wrong, not merely redundant

#4957 moved the gate **post-compute**. `is_delta_efficient` has no production caller — `#[cfg_attr(not(test), allow(dead_code))]` at `crates/core/src/ring/interest.rs:439`, with rustdoc saying "do not re-wire it as a pre-compute refusal." The live predicate is `gate_delta_size` (`interest.rs:1817`):

```rust
if delta.as_ref().len() >= our_state_size.saturating_add(MIN_FULL_STATE_SAVING_BYTES)
```

`summary_size` is carried on the variant as context and is **not an input to the decision**.

So this PR's counter — `state != 0 && summary * 2 < state`, documented as "**proof** that the sizes reaching this arm are not the sizes the gate evaluated" — now fires on every *honest* post-compute refusal that happens to have a small summary. The room's own 33,804 B / 251,887 B pair trips it. It would ship a false-alarm generator whose rustdoc asserts it is proof of a bug.

The `record()` comment ("deliberately mirrors the gate's predicate verbatim rather than calling it, so it stays a check on the RECORDED sizes even if the gate's own definition later changes") is precisely the decision that made it rot within a day.

Neither `main` nor this branch records what today's gate actually weighs: `delta.len()`.

### Rebase cost

`git merge-tree` against current `main`: **14 conflicting hunks** in `broadcast_payload_mix.rs` (`Window`, its `Default`, `record_delivered`, the whole `payload_mix_json` gate block, the test module) and **7** in `broadcast_queue.rs` (the gate-inputs local, the arm-classification match, both `record_delivered` call sites, two source-scrape pins). Every one sits in code #4938 / #4956 / #4962 rewrote. That is a rewrite, not a rebase — and the thing being rewritten is a predicate that is now invalid.

### Honest counter-argument, recorded

A ratio-of-sums is **byte-weighted**. A numerous population of *small* mis-attributed refusals would be invisible against a few large ones; this PR's per-sample counter is count-weighted and would catch that. That is a real methodological gap in `main`'s field. It does not rescue this branch: `main` publishes `full_not_efficient_sends` beside `_bytes`, so such a population would show as a collapsed mean payload (it did not — means were 551-831 KB), the arm has since gone to ~zero, and salvaging the idea means writing a new predicate against `delta.len()`, not rebasing this one.

Closing. Reopen or cherry-pick from `telemetry-notefficient-sizes` if the post-compute gate ever needs the same treatment — but write the predicate against `delta.len()`.

[AI-assisted - Claude]

PR #4943, opened 2026-07-25
fix(update): stop broadcasting our full state relabeled as our summary

sanity:
## Problem

Since 2026-03-03 (`e11cb4bbf`), every applied UPDATE has broadcast the node's **full post-update state, relabeled as its "summary"**, to every interested peer. `update_contract` built `UpdateExecution.summary` as `StateSummary::from(<state bytes>)`, and `send_proactive_summary_notification` shipped that value as our `SummaryEntry` — an optimization explicitly to "avoid an extra WASM `summarize_state` call".

Receivers cache it as our summary. The delta-efficiency gate then computes `is_delta_efficient(state_size, state_size)` → refuses → the next broadcast to that peer ships **full state** — which re-poisons them. A self-sustaining loop on every actively-updated contract, out-racing both repair mechanisms (post-delivery cache, ~5-min heartbeat) on any busy contract.

Production measurement (via #4938's instrumentation, 2026-07-25): the refused-delta arm is **41% of ALL network wire bytes** (flat over time — structural); for the hottest contract, **87.4%**, with per-send bytes equal to state size **byte-for-byte**. User-visible: 100 GB in hours; p90 peers above 10 GB/hour. Refs #3335, #4923.

## Solution — sender-side only, deliberately

1. **`send_proactive_summary_notification`** no longer accepts a precomputed summary; after its 100 ms throttle it fetches the contract's real summary (`get_contract_summary`), like its sibling `send_summary_back_on_rejection`. None → nothing sent (silence beats poison; the throttle token is deliberately consumed either way).
2. **`UpdateExecution` is now `{ value, changed }`** — the summary field is gone, and `update_contract` does zero summary work. The only genuine consumers — the two client-facing `UpdateResponse { summary }` sites in `drive_client_update` — fetch the real summary themselves post-commit (`contract_summary_or_empty`: real summarize via the handler; **empty** on failure, documented, never state bytes). The four relay destructure sites bind nothing: **zero added handler round-trips on the relay/no-change storm path.**
3. **Cost note (corrected per load review):** a post-commit summarize after a *changed* apply is always a memo miss (`summary_cache` invalidates on every state write) — the real argument is that broadcast fan-out already issues the identical `get_contract_summary` on every state change, so whichever runs first pays the single WASM call; the dominant no-change case is a pure memo hit.
4. **Scope boundary, stated plainly:** this PR stops the poison at its source and fixes the client API. It adds **no receiver-side protection** — an upgraded receiver still caches whatever summaries unupgraded senders emit until those senders upgrade (relief scales linearly with sender upgrade fraction; monotone at every mix, per load review). A receiver-side floor was built, reviewed, found traffic-neutral-and-priority-inverting as designed, and moved to `fix/summary-floor-v2` for a redesigned follow-up (keep-on-refuse; Background-priority budgeted size check; the same credibility test gating the staleness heal).

Wire format unchanged: no enum/derive/serde touched; `SummaryEntry`/`sender_summary_bytes` remain opaque bytes whose *values* become honest. Interoperable with unupgraded peers.

## Testing

- `client_facing_summary_is_the_contract_summary_not_the_state` — the incident pin; **mutation-verified** (discarding the contract's summary in the helper fails it: "must be the contract's summarize_state output (33800 bytes), not the merged state (253200 bytes)")
- `contract_summary_or_empty_error_path_returns_empty_not_state` + a scrape pin allowing `StateSummary::from(` in the helper **only** as the empty fallback — **second mutation verified** (flipping the fallback to non-empty bytes fails both)
- `update_contract_never_builds_a_summary_from_state_bytes` — `update_contract` contains no summary construction at all (whitespace-collapsed, reflow-proof needles)
- `honest_summary_passes_the_delta_gate_where_poisoned_refused`
- Full lib suite: **4297 passed, 0 failed**; fmt clean; clippy (CI gate) clean.

## Review record

Full-tier surface. Codex unavailable (out of credits, owner-confirmed) → fallback per rule: two independent Claude lenses (skeptical-correctness; load/rollout) + independent non-author verification by the coordinating session. Round 1 on `91c73ad2` produced 4 majors + 2 HIGHs — all in the receiver floor and the per-update summary cost — plus a GO on the sender half. This head (`8af256f5`) is the prescribed revision: subtractive (floor removed), the MAJOR-4/MEDIUM cost fix, and MINOR-6 test hardening. Delta re-verified hunk-by-hunk by the coordinator against the reviews' prescriptions; full CI re-runs on this head. Post-release falsifiers (via #4938 fields): `full_not_efficient` must collapse with sender upgrades; `full_no_our_summary` rising instead = handler-load loop, cut the four discarded-summary relay sites first; `full_no_their_summary_tracked` absorbing the drop = relief not reaching.

Fixes the root cause of #3335. Refs #4923.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/update.rs
+++ b/crates/core/src/operations/update.rs
@@ -130,9 +130,20 @@ pub(crate) struct BroadcastTargetResult {
     pub skipped_sender: usize,
 }
 
+/// Result of [`update_contract`]: the merged state and whether it changed.
+///
+/// Deliberately carries NO summary (#4923): an earlier version bundled
+/// `StateSummary::from(<state bytes>)` here — the state relabeled as a
+/// summary — which `send_proactive_summary_notification` then broadcast,
+/// poisoning every receiver's delta-efficiency gate into permanent
+/// full-state fallbacks. The consumers that genuinely need a summary (the
+/// two client-facing `UpdateResponse` sites in `drive_client_update`) fetch
+/// the contract's real one via [`contract_summary_or_empty`] themselves, so
+/// the relay/no-change hot paths pay zero extra contract-handler
+/// round-trips. Pinned by
+/// `update_contract_never_builds_a_summary_from_state_bytes`.
 pub(crate) struct UpdateExecution {
     pub(crate) value: WrappedState,
-    pub(crate) summary: StateSummary<'static>,
     pub(crate) changed: bool,
 }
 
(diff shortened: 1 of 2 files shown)


sanity:
## Review round 1 — skeptical lens: 4 majors, revision in progress

Full findings verified against the diff. Disposition:

- **MAJOR 1 (floor clears honest cache, same bytes on wire):** ACCEPTED — receiver floor **stripped from this PR**; preserved on `fix/summary-floor-v2` for a redesigned follow-up (keep-on-refuse + TTL-only, per the review's suggestion).
- **MAJOR 2 (unbudgeted NetworkRelay-priority whole-state size fetch on the interest-sync arm):** ACCEPTED — removed with the floor; v2 requires a Background-priority, budgeted size source before it returns.
- **MAJOR 3 (staleness heal ungated against poisoned summaries — the bigger emitter):** ACCEPTED as pre-existing; the sender fix closes it for upgraded senders. The v2 follow-up will gate the heal on the same credibility test. Not expanded tonight.
- **MAJOR 4 (+2 handler round-trips per applied relay update, value discarded by 4/6 consumers):** ACCEPTED — `UpdateExecution.summary` dropped/lazy; only the two client-facing drivers fetch it, relay/no-change paths back to zero added round-trips.
- **MINOR 5 (empty summary passes floor):** moot tonight (floor stripped); folded into v2's design (refuse len==0).
- **MINOR 6 (pin test can't catch helper-fallback reversion; error path unexercised):** ACCEPTED — scrape extended to the helper body + error-path stand-in case + second mutation transcript required.
- **MINOR 7 (arm re-attribution vs #4938 field data):** doc note added to PR body: post-fix, poison signatures leave `full_not_efficient`; analysts should read the arm split accordingly.
- **MINOR 8 + nits:** docs/rustdoc updates as suggested; global warn counter note moot with floor removal.

Net effect: tonight's diff contains **no receiver-side behavior changes** — sender-side de-poisoning + honest client summaries + tests only.

sanity:
## Review round 1 — load/rollout lens: GO, findings dispositioned

Verdict: **GO with HIGH-1/HIGH-2 fixed first** — both live in the receiver floor, which is already stripped from this PR per the skeptical lens (preserved on `fix/summary-floor-v2`; the v2 design requirements now include: Background-priority budgeted size source [HIGH-1] and floor-before-staleness ordering [HIGH-2]).

- **MEDIUM (4/6 relay sites discard the fetched summary; +50% loop turns on the storm path):** converges with skeptical MAJOR 4 — being fixed tonight (lazy summary; relay paths back to zero added round-trips).
- **Memoization reasoning corrected** in-code per this review: changed applies always miss the memo (state_hash_cache invalidated on every write); the fetch is ~free because the broadcast fan-out already forces the identical summarize, and the dominant no-change case is a pure memo hit.
- **Warn sizing (Q4):** moot tonight (floor stripped); v2 carries WARN_EVERY=1000 and the per-peer-sampling caveat.
- **Post-release falsifiers adopted** (release-validation plan): `full_not_efficient` bytes must **collapse** as senders upgrade; if `full_no_our_summary` rises to meet it, the load loop is real and the four discarded-summary relay sites are the first cut; if `full_no_their_summary_tracked` rises by the amount `full_not_efficient` fell, relief isn't reaching and the sender-fix coverage needs rechecking. FairQueue high-water via per-node /v1/node/status polling as the load cross-check.

Independent convergence note: two lenses, spawned blind to each other, isolated the same weak component (the floor) and the same strong component (the sender fix) with compatible mechanisms. The revision ships the strong half tonight.

PR #4949, opened 2026-07-26
feat: add netcheck for live-network put, get and retention (#4665)

netsirius:
## Problem

Findability and retention regressions (content that cannot be routed to, or that stops being retrievable after a few days) surface only through production telemetry, weeks after the change that caused them. PR CI and the nightly simulations cannot see them: a simulated network has none of the storage pressure, topology or churn that make those properties interesting.

PR #4848 attempted this as a synthetic CI test. The conclusion there was that a test which never touches the real network cannot answer the question. This PR is the agreed replacement, and supersedes it.

## Solution

`crates/netcheck` is a plain network client: it speaks only the freenet-stdlib WebSocket API and never links `freenet-core`, so it measures what a real user experiences and survives core refactors.

A `put-get` run PUTs 3 small contracts plus one ~1 MB contract through one gateway, boots an ephemeral peer with an empty data dir that joins through a *different* gateway, and GETs them back verifying byte-identity. It then re-GETs contracts from previous runs (24 h / 48 h / 7 d windows, tracked in a persistent manifest) against their blake3 hashes. One JSON line per operation on stdout, non-zero exit on any failure, and no retries by design: an operation that only succeeds on retry is the regression this exists to surface.

**What each check actually proves.** `--gateway-spec` pins the join, not the steady-state connection set: ring maintenance opens further connections on its own, and the gateway that stored the PUTs can be among them. The same-day GET is therefore a liveness and transfer check, while the findability signal lives in the multi-day re-GETs, whose contracts were published by a run that is long gone against a topology that has since changed. The run report records which peers the ephemeral node actually connected to, so a degraded run is visible in the data instead of silently passing.

**Safe beside a production node**, which is the intended deployment: the ephemeral node gets its own dirs and ports, child processes are killed by handle and never by name (no `pkill` anywhere in the crate), and `--disable-auto-update` stops a release published mid-run from making it exit 42.

A second subcommand, `netcheck emit`, reads a run's report on stdin and publishes it to an OTLP collector. It belongs in this crate rather than in a script because the translation between the report and the published schema is coupling that nothing external can verify: `emit` reads the report back with `OpReport` and `RunMeta`, the very structs that wrote it, so renaming a field becomes a compile error instead of a silently missing column downstream. A test writes a report through `Report::push` and parses it back field by field to keep that guarantee honest.

The division of labour is deliberate. The check measures and does not judge; `emit` decides which failures are severe, because that is deployment policy rather than a property of the network; the collector transports. What counts as severe is per scenario (`put_get`'s retention windows are soft, since the hosting cache is a byte-budget LRU and evicting an unwanted contract is correct behaviour rather than an outage), and a scenario nobody has classified has no soft dimensions, so any failure of a new check is hard by default.

### Running it

A check nothing runs measures nothing, so the nightly ships here too. It cannot run on a cloud runner: it needs the gateway's loopback WebSocket API and the UDP port opened on that host for its ephemeral peer.

`.github/workflows/netcheck-nightly.yml` runs on the self-hosted runner at 04:00 UTC, an hour after the simulation nightly so the morning's results arrive in a predictable order. Failures notify River through the same `riverctl` and `RIVER_SIGNING_KEY` transport the simulation nightly already uses, rather than a second standalone credential.

The job body is three steps. Everything else is in `scripts/netcheck-nightly.sh`, and that split is the point: **a workflow file cannot be exercised before it reaches the default branch**, since GitHub only dispatches workflows present there. Putting the logic in a script means it can be run by hand over ssh on the same host, with the same paths and the same flags, before and after this merges.

The script also carries the two things that matter when a check shares a machine with a production node:

- It refuses to run when the disk is below a threshold. Skipping a night costs nothing; filling the gateway's disk does.
- It removes the ephemeral node's working directory before and after each run. A killed run would otherwise leave it behind, and that node's hosting cache is budgeted at RAM/8 up to 1 GiB.

The getter gateway is resolved at runtime from the public index rather than hardcoded, so a rotated address or key does not silently leave the check reading through the same gateway it wrote to. Publishing failures are logged but never change the verdict, because a collector that is down is not a network regression. Every path and port is overridable by environment variable, so a second vantage point needs no edit to the script.

## Testing

Validated end to end on two linked local gateways reproducing the production nova/vega shape, PUT through one and the ephemeral getter joined through the other: 8/8 operations, then 12/12 on a second run with the manifest aged 24 h to exercise the retention path. `cargo fmt`, `clippy -D warnings` and `cargo test` clean.

15 unit tests. Seven cover retention-window selection: inclusive bounds, newest run wins inside a window, an empty window does not borrow from its neighbour, and the prune horizon outlives the widest window (a shorter horizon would make the 7 d check silently pass forever with nothing to ask for). Eight cover `emit`: the report round trip in both directions, the four verdicts, that an unclassified scenario inherits no leniency, and dimension parsing.

The publishing path was exercised end to end against a local OTLP collector configured identically to the production one, using the output of a real netcheck run: the records arrive with every field intact and land in the dashboard's tables (companion PR in freenet-telemetry-dashboard).

Chaining two runs also exposed a real harness bug, fixed here: a late `PutResponse` from an already-timed-out operation was consumed by the next PUT's wait loop and failed it on a key mismatch, turning one slow PUT into a cascade of false failures.

The workflow file itself is unverifiable until it is on `main`, which is why the logic is in a script. The intended first step after merging is a manual run over ssh before waiting for a cron.

## Notes for the reviewer

Two things that contradict the CLI documentation but are verified by working runs:

- `--is-gateway` does accept `--gateway` entries, although the `--skip-load-from-network` help says a gateway "always runs isolated" and mentions only `--gateways` JSON.
- `--disable-auto-update` is documented for from-source deployments running ahead of a release. Used here for a node that lives for minutes, where exiting to update is pointless and would fail the check spuriously.

Today any failed run notifies River. Distinguishing a genuine regression from a retention window that degraded (expected under storage pressure, since the hosting cache is a byte-budget LRU) from an infrastructure error is threshold work that belongs with the dashboard that can show the trend, not with a boolean job status. `netcheck emit` already computes the four-way verdict; only the alerting reads it as a single bit.

The dashboard panel that renders these results is a follow-up, with its own PR in freenet-telemetry-dashboard.

## Fixes

Part of #4665. Replaces #4848.

### DIFF
--- a/.github/workflows/netcheck-nightly.yml
+++ b/.github/workflows/netcheck-nightly.yml
@@ -1,0 +1,119 @@
+name: Netcheck (Nightly)
+
+# Runs netcheck against the LIVE network from the gateway host (#4665).
+#
+# Self-hosted on purpose: the check needs the gateway's loopback WebSocket API
+# and the UDP port opened there for its ephemeral peer, neither of which is
+# reachable from a cloud runner.
+#
+# The job body is deliberately thin. All the logic is in
+# scripts/netcheck-nightly.sh, which can be run by hand over ssh on the same
+# host; a workflow file cannot be tested before it reaches the default branch.
+
+on:
+  schedule:
+    # 04:00 UTC, an hour after the simulation nightly so a human reading the
+    # morning results sees them in a predictable order.
+    - cron: "0 4 * * *"
+  workflow_dispatch:
+
+# Never run two at once: they would contend for the ephemeral peer's UDP port
+# and both write the same manifest.
+concurrency:
+  group: netcheck-nightly
+  cancel-in-progress: false
+
+jobs:
+  netcheck:
+    name: Live network check
+    runs-on: self-hosted
+    # A normal run is minutes. The cap is generous because the first PUT after
+    # a release pays for WASM validation, and joining the ring is not instant.
+    # If it ever trips, something is genuinely stuck: investigate rather than
+    # raise it.
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Ensure the wasm target is present
+        # netcheck compiles the probe contract at startup. Idempotent.
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: Run netcheck against the live network
+        env:
+          NETCHECK_OTLP_ENDPOINT: http://127.0.0.1:4318
+          NETCHECK_VANTAGE: nova
+        run: bash scripts/netcheck-nightly.sh
+
+      - name: Upload the run report
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: netcheck-report
+          path: ~/netcheck/last-run.jsonl
+          if-no-files-found: warn
+          retention-days: 30
+
+  # Same transport as the simulation nightly: riverctl against the Freenet
+  # gateway with the shared RIVER_SIGNING_KEY, so there is one set of
+  # credentials to maintain rather than a second standalone token.
+  #
+  # Today any failure notifies. Distinguishing a real regression from a
+  # degraded retention window or an infrastructure error is the threshold
+  # work tracked in #4665, and belongs with the dashboard that can show it.
+  notify-failure:
+    name: Notify River on Failure
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    needs: netcheck
+    if: failure() && needs.netcheck.result == 'failure'
+    continue-on-error: true
+    steps:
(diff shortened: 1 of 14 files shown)


iduartgomez (approved):

iduartgomez:
@netsirius I guess we are missing chan ges in CI to run this?

netsirius:
@iduartgomez yes, but I was planning a separate branch for that. But it's probably better to keep it together. Just pushed the CI changes

PR #4953, opened 2026-07-26
fix(interest): upsert delivered summaries so untracked co-hosts escape full-state broadcasts

### LINKED ISSUE #4952: Untracked co-hosts are a full-state broadcast fixed point (58% of fleet broadcast bytes)
Post-0.2.108 fleet telemetry (broadcast_payload_mix, ~6h window, 666 reporting peers): `full_no_their_summary_untracked` is 4.84GB = 58% of all broadcast bytes; `full_no_their_summary_tracked` another 15%. Deltas are 70% of sends but only 17% of bytes.

Mechanism: the UPDATE fan-out target list is built from advertised co-hosts (NeighborHosting), which never enter `InterestManager::interested_peers`. After a delivered broadcast we cache "what we just sent them" via `update_peer_summary` (broadcast_queue.rs::record_delivery_to_interest) so the next send can be a delta — but that method is deliberately get_mut-only and silently no-ops for a peer with no interest entry. So an untracked co-host receives FULL STATE on every update, forever: send full → cache write no-ops → still no summary → send full again. The ~5-min InterestSync `Summaries` handler has the same no-op shape, so the anti-entropy backstop can't seed it either. The tracked arm self-heals after one full send, which is why it's small.

Fix: an upsert variant (`upsert_peer_summary`) used at the two call sites that know the peer holds the summarized state — the post-delivery cache and the `Summaries` handler — capped by MAX_INTERESTED_PEERS_PER_CONTRACT, not touching the demand counters that feed eviction ranking, and never feeding the state-broadcast target list (Source-2 stays removed).

[AI-assisted - Claude]

sanity:
## Problem

Post-0.2.108 fleet telemetry: `full_no_their_summary_untracked` is **58% of all broadcast bytes** (4.84GB in a ~6h window), plus 15% for the tracked variant. Deltas are 70% of sends but only 17% of bytes. Users report 20GB/6h nodes (#4952).

Mechanism: UPDATE fan-out targets come from NeighborHosting advertisements — a population that never enters `InterestManager::interested_peers`. The post-delivery summary cache (`record_delivery_to_interest`, the #4442 fix) and the InterestSync `Summaries` handler both write through `update_peer_summary`, which is deliberately get_mut-only and **silently no-ops for an untracked peer**. So an advertised co-host receives full state on every update, forever: send full → cache write no-ops → still no summary → send full again. A fixed point, not a cold start (pinned by the pre-existing `update_peer_summary_is_a_silent_noop_for_an_untracked_peer` test, which documents exactly this trap).

## Solution

New `InterestManager::upsert_peer_summary` — like `update_peer_summary` but creates the entry when absent — used at the two call sites that KNOW the peer holds the summarized state: the post-delivery cache (we just delivered them exactly that state) and the `Summaries` handler (the peer itself reported it). One delivered full state now seeds the summary; subsequent broadcasts collapse to deltas.

Safety properties, each deliberate:
- **Capped**: rejects a NEW peer at `MAX_INTERESTED_PEERS_PER_CONTRACT` with no side writes (no amplification vector, no reverse-index zombie).
- **No demand pollution**: never touches `downstream_subscriber_count` / `local_client_count` (eviction's invariant-3 demand signal is a separate, explicitly-incremented counter — verified, not derived from map membership).
- **No fan-out regrowth**: upserted entries never feed the state-broadcast target list (Source-2 removal stands, pinned in op_state_manager.rs). They do join the small `Summaries`-notification fan-out — which is beneficial (the co-host learns OUR summary and can send US deltas) and summary-sized.
- **Guard discipline**: holds the `interested_peers` shard guard across the reverse-index/hash-index writes (#4129/#4171).
- **`None` reports keep clear-only semantics** in the Summaries handler — no entry is created just to hold `None`.
- `update_peer_summary` itself is unchanged (its no-op pin still passes): writes of unknown provenance still can't grow the map.

## Testing

- 3 unit tests: seeding for an untracked peer (+ reverse-index maintenance + no local-interest side effects), cap rejection with no zombies, update-path preservation of `is_upstream`.
- 2 source-scrape pins (whitespace-stripped, rustfmt-proof): the delivery cache routes through the upsert (mutation-verified — reverting the call site fails the pin), and the Summaries arm upserts `Some`/clears `None`.
- Full interest + broadcast_queue suites: 105 passed. `cargo fmt` + `clippy -D warnings` clean.

## Expected field effect

`full_no_their_summary_untracked` should collapse to near-first-send-only levels as the fleet upgrades; fleet broadcast bytes should drop several-fold on busy contracts. Falsifier: if the arm does NOT collapse for upgraded emitter pairs, the untracked population is not delivery-reachable and this diagnosis is wrong.

Fixes #4952

[AI-assisted - Claude]

https://claude.ai/code/session_01F2CFGizKUBc5hewJ5qzwJb

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2810,11 +2810,23 @@ async fn handle_interest_sync_message(
                             _ => false,
                         };
 
-                        op_manager.interest_manager.update_peer_summary(
-                            &contract,
-                            &pk,
-                            their_summary,
-                        );
+                        // #4952: upsert (not update) when the peer reported a
+                        // real summary, so the ~5-min anti-entropy exchange can
+                        // seed a summary for an advertised co-host we don't
+                        // interest-track — otherwise those peers stay full-state
+                        // broadcast targets forever. A `None` report keeps the
+                        // old clear-only semantics (no entry is created just to
+                        // hold `None`).
+                        match their_summary {
+                            Some(theirs) => {
+                                op_manager
+                                    .interest_manager
+                                    .upsert_peer_summary(&contract, &pk, theirs);
+                            }
+                            None => op_manager
+                                .interest_manager
+                                .update_peer_summary(&contract, &pk, None),
+                        }
 
                         if is_stale && !stale_contracts.contains(&contract) {
                             stale_contracts.push(contract);
@@ -3946,6 +3958,37 @@ mod tests {
     /// `ResyncRequest` arm, which is already state-gated (returns early when
     /// `get_contract_state` is `None`) and is not heartbeat-driven, so it is
     /// excluded by slicing up to that arm.
+    /// #4952 pin: the `Summaries` handler must UPSERT a reported summary so
+    /// the ~5-min anti-entropy exchange can seed one for an advertised
+    /// co-host we don't interest-track (`update_peer_summary` no-ops for
+    /// untracked peers — the full-state fixed point). A `None` report keeps
+    /// the clear-only `update_` semantics: no entry is created just to hold
+    /// `None`. Matches whitespace-stripped source (rustfmt-proof).
+    #[test]
+    fn summaries_arm_upserts_reported_summary_pin() {
+        let src = include_str!("node.rs");
+        let handler_start = src
+            .find("async fn handle_interest_sync_message")
+            .expect("handle_interest_sync_message not found");
+        let handler_end = handler_start
+            + src[handler_start..]
+                .find("\nmod tests {")
+                .or_else(|| src[handler_start..].find("\n#[cfg(test)]"))
+                .expect("end of handler region not found");
+        let body: String = src[handler_start..handler_end].split_whitespace().collect();
+        assert!(
+            body.contains("upsert_peer_summary(&contract,&pk,theirs)"),
+            "Summaries arm must upsert a Some(summary) report (seeds untracked \
+             co-hosts, #4952)"
+        );
+        assert!(
+            body.contains("update_peer_summary(&contract,&pk,None"),
+            "Summaries arm must keep clear-only update semantics for a None \
+             report — creating an entry just to hold None would relabel \
+             untracked traffic as tracked without enabling deltas"
+        );
+    }
+
     #[test]
     fn interest_sync_periodic_arms_summarize_only_hosted_or_in_use_pin() {
         let src = include_str!("node.rs");
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -3418,9 +3418,12 @@ async fn handle_interest_sync_message(
             let peer_key = get_peer_key_from_addr(op_manager, source);
             if let Some(pk) = peer_key {
                 let summary = freenet_stdlib::prelude::StateSummary::from(summary_bytes);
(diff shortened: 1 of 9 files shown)


sanity:
## Review record (Full tier)

**External model**: Codex attempted (`codex exec`, gpt-5.6-sol) — hard usage-limit failure, resets Aug 1. Per the review policy, fell back to a 4-lens Claude panel with perspective diversity: skeptical (guard discipline/races/caps), code-first (independent model vs claims), testing (coverage + pin robustness), and a hosting-invariants/amplification lens. No blocking findings; every substantive finding addressed in f4e331cf:

| Finding | Disposition |
|---|---|
| Demand-counter increment keyed on interest-map newness — upsert-seeded entries make a later genuine subscriber skip the increment + #4359 flush (code-first MAJOR, sharpened during synthesis) | **Fixed**: keyed on hosting-map `NewAdd` (enum already existed; un-flattened the ring wrapper) |
| Unsubscribe decrement gated on `\|\| was_interested` — interest-only peers steal decrements; double-decrement after lease sweep | **Fixed**: gated on `was_downstream` alone; superseded pin test re-pinned to new semantics with rationale |
| Lease renewal clobbers cached summary with `None` every ~8 min (found during the fix — likely a chunk of the *tracked* arm's 15%) | **Fixed**: register-or-refresh preserves existing summaries |
| Sim fan-out path still wrote through the no-op (2 lenses) | **Fixed**: sim path upserts (prod parity) |
| 3 sibling self-reported-summary sites forfeit free seeding | **Fixed**: BroadcastTo inline/streaming + ResyncResponse now upsert |
| Silent at-cap rejection (3 lenses) | **Fixed**: debug-level diagnostic (rationale in comment) |
| Stale docs claim untracked arm is permanent; "never registers" causal overstatement (testing + code-first) | **Fixed**: telemetry-interpretation docs updated to post-fix reality |
| Test gaps: at-cap existing peer, TTL sweep, behavioral test through real delivery gate, untracked drop-guard, pin negative needle | **Fixed**: all added |
| Skeptical M1: Interests full-replace can wipe the pair for a co-host heartbeating a partial interest set that lacks the contract | **Accepted + documented**: worst case one full state per ~5-min cycle per such pair (the next delivery re-seeds); the arm's falsifier now includes the sawtooth signature. Empty-interest-set co-hosts (no heartbeat sent) keep the fix fully. |
| Skeptical M2: 120s stream completion vs 90s disconnect grace can resurrect a just-removed peer's entry | **Accepted**: bounded by the 20-min TTL sweep (verified no GC exemption; new sweep test), occupies one cap slot at most, proactive-Summaries skip unresolvable peers |
| Streaming tail-loss disclosure: an untracked peer with a lost tail now gets a delta-on-wrong-base healed by ResyncRequest/heartbeat instead of full-state-by-construction | **Accepted**: same tradeoff already accepted for tracked peers (#4145/#4235 backstops) |

Verification: full lib suite 4306/4306 green, clippy -D warnings clean, fmt clean; the delivery-site pin was mutation-verified (reverting the call site fails it). Note: `clippy --all-targets` surfaces ~10 pre-existing lints in test-target code outside CI's gate — untouched by this PR.

[AI-assisted - Claude]

PR #4955, opened 2026-07-26
chore: point telemetry dashboard references at telemetry.freenet.org

sanity:
## Problem

The network telemetry dashboard now serves over HTTPS at https://telemetry.freenet.org (nova, Caddy + Let's Encrypt). The old `http://nova.locut.us:3133/` URL 301-redirects there.

Four references in this repo still pointed at the old URL. One of them matters functionally:

`scripts/check-endpoints.sh` probed `http://nova.locut.us:3133`. That now answers **301**, which still falls inside the script's `200 <= code < 400` success band — so it would have gone on reporting the dashboard green while actually testing nothing but the redirect stub.

The other three are comments, and two of them had gone false rather than merely stale: they justify accepting `http:` URLs in the shell `open_url` handler by citing the dashboard as a plain-HTTP service with "no TLS configured". Anyone re-deriving that security decision now reads a rationale whose only cited example contradicts itself.

## Solution

- `scripts/check-endpoints.sh` — probe `https://telemetry.freenet.org` (now returns a real 200).
- `shell_bridge.js` / `path_handlers.rs` — keep the `http:`-acceptance rationale intact (it is still correct for other plain-HTTP self-hosted services) and reword the dashboard mention as the historical trigger for freenet/river#231 rather than a present-tense example.
- `home_page/cards.rs` — comment referencing the dashboard's ring visualization.

No behavior change: the Rust edits are comments only, and the JS edit is a comment inside the `open_url` block.

## Testing

- `cargo test -p freenet --lib shell_open_url` — 4/4 pass, including `shell_open_url_handler_accepts_http_and_https_but_blocks_localhost`, which source-scrapes the exact JS block whose comment changed.
- `node --check shell_bridge.js` — syntax clean.
- `cargo fmt --all` — clean.
- `bash scripts/check-endpoints.sh` — dashboard row now reports `200` (previously would report `301`). The other DOWN rows are pre-existing and unrelated (UDP gateway ports that never served HTTP).

Not opened as a draft deliberately: freenet-core drafts skip the heavy CI jobs, so a draft's green is not a real signal.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/home_page/cards.rs
+++ b/crates/core/src/server/home_page/cards.rs
@@ -644,7 +644,7 @@ pub fn build_ring_svg(
     // (x2,y2) with the control point pulled toward the SVG centre.
     // `bend_base` is the maximum bend for antipodal points; chord
     // length scales bend so short hops look flat and long hops arc
-    // through the interior. Matches the nova.locut.us:3133 ring
+    // through the interior. Matches the telemetry.freenet.org ring
     // routing visualization.
     let curve_path = |x1: f64, y1: f64, x2: f64, y2: f64, bend_base: f64| -> String {
         let dx = x2 - x1;
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -2601,7 +2601,11 @@ mod tests {
         );
         assert!(
             !html.contains("freenet.org"),
-            "shell page must not reference external origins (CORS)"
+            "shell page must not reference external origins (CORS). This is a \
+             plain substring check over the whole rendered page, including the \
+             inlined shell_bridge.js — so a COMMENT that merely mentions a \
+             freenet.org host trips it too. If that is what you hit, drop the \
+             hostname from the comment rather than loosening this assertion."
         );
         // Shell message handler must be present in bridge JS
         assert!(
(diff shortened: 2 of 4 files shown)


sanity:
## CI was red: `shell_page_contains_iframe_and_bridge` — caused by this PR, now fixed

The failing check was **Unit & Integration**, one test:

```
server::path_handlers::tests::shell_page_contains_iframe_and_bridge
panicked at crates/core/src/server/path_handlers.rs:2602:
shell page must not reference external origins (CORS)
```

**Root cause, and it was this PR's doing** — not pre-existing, not flaky. `main` has zero
occurrences of the string `freenet.org` in `shell_bridge.js`; this branch added exactly one,
in the reworded `open_url` comment:

```js
// moved to https://telemetry.freenet.org/), and an https-only
```

`shell_bridge.js` is inlined verbatim into the rendered shell page, and the guard is a plain
substring assertion (`!html.contains("freenet.org")`) over the whole page. It does not
distinguish code from comments, so *mentioning* the host in a comment trips it just as a real
external-origin reference would.

**Fix:** keep the corrected rationale, drop the hostname. The comment's job was to stop
citing the dashboard as a present-tense "no TLS configured" example, and it still does that —
it now says the dashboard was plain HTTP when freenet/river#231 was filed and has since moved
to HTTPS. The concrete URL stays where it is actually exercised, in `scripts/check-endpoints.sh`.

The guard itself is untouched. I deliberately did not loosen it to comment-aware matching:
that is a security-adjacent assertion and this is a `chore:` PR. Two things were added instead
so the next person does not lose time to the same trap:

- a note at the comment site saying why the hostname must not come back
- the cause spelled out in the assertion's own failure message

**Verification**

- `cargo test -p freenet --lib shell_` — 32/32 pass, including the previously-failing test.
- Mutation-checked the guard is not vacuous: re-inserting a `telemetry.freenet.org` mention
  into the JS comment makes it fail again with the new message; reverted.
- `cargo fmt --all -- --check` clean; `cargo clippy --locked -- -D warnings` (CI's exact
  invocation) clean.
- `node --check shell_bridge.js` clean.
- URLs verified live rather than assumed: `https://telemetry.freenet.org` → **200**, and
  `http://nova.locut.us:3133` → **301** to it, which is exactly the false-green the script
  change fixes (301 sits inside the script's `200 <= code < 400` success band).
- `bash scripts/check-endpoints.sh` — the dashboard row now reports 200.

Rebased onto current `main` (`7ef31542`, 0.2.112) — the branch was ~3 releases behind.

### Unrelated finding, not fixed here

While running the script: the `river.freenet.org` row reports DOWN because the hostname has
**no DNS record** (`dig +short river.freenet.org` returns nothing; `freenet.org` and
`telemetry.freenet.org` both resolve). That is a second stale row in the same file, but I have
not touched it — I do not know the correct replacement and guessing a URL is worse than leaving
an honest DOWN. Flagging it rather than silently changing it.

[AI-assisted - Claude]

Issue #4956, opened 2026-07-26
Redundant broadcast fan-out is the dominant bandwidth consumer: 67% of applies are no-ops carrying 80% of bytes


sanity:
## Bandwidth measurement: where the bytes actually go (0.2.108 fleet)

Measured from `/mnt/media/freenet-telemetry/logs.jsonl`, a ~15 min window (634-678 peers reporting on 0.2.108), taken just before 0.2.109 went public. Method notes at the bottom. This changes the priority order in the hand-off doc, so the corrections come first.

### 1. GET-path initial syncs are NOT a major consumer

The hand-off listed "GET-path initial syncs, UNMEASURED" as the top remaining suspect, on the theory that serving full state to new joiners was invisible to `broadcast_payload_mix` and could be a large share of per-node totals. It is measurable today, and it is small.

`transfer_completed` carries `bytes_transferred`, and `StreamId` has an `OPS_STREAM_BIT` that separates operations-level streams from transport-level ones. Splitting outbound stream bytes on that bit:

| bucket | bytes | streams | mean |
|---|---|---|---|
| ops-level (`OPS_STREAM_BIT` set) | 8.02 GB (57%) | 7,944 | 985 KB |
| transport-level | 6.04 GB (43%) | 140,306 | 42 KB |

The ops bucket is almost exactly the streamed full-state broadcast reported by the payload mix (7.45 GB of state bytes), leaving ~0.57 GB for `sender_summary_bytes` plus bincode framing. The transport bucket is mostly delta broadcasts (corr 0.69 between delta send count and transport stream count; peers with >50 transport streams but <10 delta sends hold only 3.4% of transport bytes).

**Update broadcast fan-out is roughly 85-90% of outbound bytes.** GET/PUT/subscribe serving is a minor contributor. I do not think building GET-path byte instrumentation is the best next investment.

Also worth recording since it was an open worry: **`sender_summary_bytes` overhead is ~7%, not a doubling.** The `not_efficient` arm reports a summary/state ratio of 0.999, but that arm is *defined* by summary >= 50% of state, so the ratio is selection bias and does not generalize.

### 2. The dominant waste is redundant fan-out: 67% of applies change nothing, and they carry 80% of the bytes

`update_broadcast_applied.changed` is the contract merge's own `state_changed` signal (traced to `UpdateExecution.changed` <- `ContractHandlerEvent::UpdateResponse.state_changed`), so it is a direct read of "did this delivery alter the receiver's state".

[...]

PR #4957, opened 2026-07-26
fix(broadcast): post-compute delta gate, plus an outbound-byte census to aim the next fix

sanity:
## Problem

`is_delta_efficient` (`summary_size * 2 < state_size`) runs as a **pre-compute** gate inside `compute_delta`: when a peer's cached summary is >= 50% of our state size, we refuse to ask the contract for a delta at all and send **full state** instead.

That trade is backwards, and the existing code comment on `DeltaUnavailable` already says so:

> Every caller falls back to sending FULL STATE, which is never smaller than the delta that was declined.

On the broadcast path the peer's summary is already sitting in our local interest cache. Its size is not a wire cost at fan-out time. The only wire question is "delta or full state", and refusing to compute the delta answers that question by sending the larger of the two. A large summary is a poor proxy for a large delta: a 500 KB room with a 500 KB summary still has a ~200 byte delta when one message is appended.

### Measured cost

From a ~15 min fleet window on 0.2.108 (634 peers, `broadcast_payload_mix` rollups, see #4956 for full method):

| arm | sends | bytes | share of broadcast bytes | mean |
|---|---|---|---|---|
| `full_not_efficient` | 2,120 | 0.81 GB | 6.8% | 372 KB |

Those 2,120 sends put 372 KB on the wire each because the gate declined to compute a delta. The rollup also reports the gate's own inputs, and the aggregate summary/state ratio for this arm is **0.999** — these are contracts whose summary is essentially the same size as their state, so the pre-compute proxy refuses ~100% of the time for them, permanently.

Share rises to roughly **11%** of broadcast bytes once 0.2.109 (#4953) collapses the `full_no_their_summary_untracked` arm.

## Approach

Move the gate to **after** computation and judge the real artifact:

- `compute_delta` no longer consults `is_delta_efficient` before calling the contract.
- A new `gate_delta_size` returns `DeltaUnavailable::NotEfficient` only when `delta.len() >= our_state_size`, so the caller's full-state fallback is taken exactly when it is genuinely the smaller payload.
- The gate is applied on the **cache-hit path too**, so a delta cached by the staleness probe (which never gates) cannot hand a caller a payload larger than the full state it exists to avoid.
- Oversized deltas are still **cached** on purpose: `cached_staleness_verdict` maps any non-empty cached delta to "peer is stale", which stays correct (the peer really has diverged, we just answer with full state), and it memoizes the refusal so the next fan-out target for the same summary pair costs no further WASM.

`is_delta_efficient` is kept, documented as a heuristic with no production caller, with its unit tests intact and a `do not re-wire this as a pre-compute refusal` note. `DeltaUnavailable::NotEfficient` keeps its name and fields because the #4938 payload-mix telemetry keys off them — the arm now means "computed, and not smaller than full state" rather than "refused before computing".

### Cost

Removing a pre-compute refusal means the contract's `get_state_delta` now runs for that population. That is **+0.76%** more delta computations (2,120 against the 279,864 the `delta` arm already performs in the same window), bounded by `BROADCAST_CH_TIMEOUT` and memoized by the existing delta cache. Given #4861 made WASM CPU a real eviction axis this was worth checking; it is noise.

## Testing

- Full `cargo test -p freenet --lib` suite.
- `interest.rs` gains direct coverage for the post-compute gate: delta smaller than state passes through, delta >= state refuses, and the cache-hit path applies the same verdict as a fresh compute.
- `is_delta_efficient`'s own unit tests are retained to pin the heuristic's behavior even though it has no production caller.

## Notes

This branch originated in the 0.2.107-0.2.109 bandwidth session as `fix/delta-gate-post-compute` and was set aside as marginal after being weighed against `full_delta_suppressed` (0.5% of bytes). That was the wrong arm — the change targets `full_not_efficient`. Rebased onto 0.2.109 `main` and re-measured here.

Do not release this until 0.2.109's field validation has a clean window, so the `untracked` collapse can be measured without a second broadcast change confounding it. The arms are instrumented separately, so review and merge can proceed.

Refs #4956

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/mock_runtime.rs
+++ b/crates/core/src/contract/executor/mock_runtime.rs
@@ -784,8 +784,12 @@ where
             );
         }
 
-        // Default mode: hash-based summary (32 bytes) to enable delta efficiency.
-        // With hash summaries, is_delta_efficient(32, state_size) = true when state > 64 bytes.
+        // Default mode: hash-based summary (32 bytes) — small, cheap to ship in
+        // interest syncs. (Historical note: while compute_delta still had its
+        // pre-compute is_delta_efficient gate, the 32-byte summary also kept
+        // that gate open for any state > 64 bytes; since #4923 the gate runs
+        // post-compute on the actual delta size, so summary size no longer
+        // decides delta viability.)
         let hash = blake3::hash(state.as_ref());
         Ok(StateSummary::from(hash.as_bytes().to_vec()))
     }
--- a/crates/core/src/node/network_bridge.rs
+++ b/crates/core/src/node/network_bridge.rs
@@ -20,6 +20,7 @@ pub(crate) mod broadcast_payload_mix;
 pub(crate) mod broadcast_queue;
 mod handshake;
 pub(crate) mod in_memory;
+pub(crate) mod outbound_message_mix;
 pub(crate) mod p2p_protoc;
 pub(crate) mod priority_select;
 
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -210,6 +210,15 @@ struct Window {
     /// Per-contract full-state byte attribution, bounded at
     /// [`MAX_TRACKED_CONTRACTS`].
     contract_full_state_bytes: HashMap<ContractInstanceId, u64>,
+    /// Per-contract bytes for the [`PayloadArm::FullNotEfficient`] arm ONLY.
+    ///
+    /// #4956: the aggregate gate-input ratio came back at 1.000 (summary size
+    /// == state size, max 839 KB each), which means some contract is feeding
+    /// state-sized bytes in as a "summary". `contract_full_state_bytes` mixes
+    /// every full-state arm together, so it cannot say WHICH contract, and the
+    /// culprit stayed unidentifiable. This narrows it to the one arm that
+    /// matters.
+    contract_not_efficient_bytes: HashMap<ContractInstanceId, u64>,
     /// Full-state sends that could not be attributed to a contract because
     /// the cap was already reached. This counts SENDS, not distinct
     /// contracts: one over-cap contract broadcasting 1,000 times contributes
@@ -259,6 +268,7 @@ impl Default for Window {
             sends: [0; PayloadArm::COUNT],
             bytes: [0; PayloadArm::COUNT],
             contract_full_state_bytes: HashMap::new(),
+            contract_not_efficient_bytes: HashMap::new(),
             attribution_dropped_sends: 0,
             attribution_dropped_bytes: 0,
             not_efficient_summary_bytes_sum: 0,
@@ -360,6 +370,16 @@ impl PayloadMix {
         // what this measurement is for.
         w.sends[idx] = w.sends[idx].saturating_add(1);
         w.bytes[idx] = w.bytes[idx].saturating_add(bytes);
+        if arm == PayloadArm::FullNotEfficient {
+            // Same cap discipline as the wider map: the key is
+            // contract-controlled, so it must not grow unbounded. Overflow is
+            // covered by the existing attribution_dropped_* counters below.
+            if let Some(tally) = w.contract_not_efficient_bytes.get_mut(contract) {
+                *tally = tally.saturating_add(bytes);
+            } else if w.contract_not_efficient_bytes.len() < MAX_TRACKED_CONTRACTS {
+                w.contract_not_efficient_bytes.insert(*contract, bytes);
+            }
+        }
         if arm.is_full_state() {
             if let Some(tally) = w.contract_full_state_bytes.get_mut(contract) {
                 *tally = tally.saturating_add(bytes);
@@ -403,6 +423,23 @@ impl Window {
         }
     }
 
+    /// The top [`TOP_CONTRACTS_REPORTED`] contracts in the
+    /// [`PayloadArm::FullNotEfficient`] arm, i.e. the ones whose delta the
+    /// gate refused. Same stable ordering as [`Self::top_contracts`].
+    fn top_not_efficient_contracts(&self) -> Vec<(ContractInstanceId, u64)> {
+        let mut tallies: Vec<(ContractInstanceId, u64)> = self
+            .contract_not_efficient_bytes
+            .iter()
+            .map(|(k, v)| (*k, *v))
+            .collect();
+        tallies.sort_by(|a, b| {
+            b.1.cmp(&a.1)
+                .then_with(|| a.0.as_bytes().cmp(b.0.as_bytes()))
+        });
+        tallies.truncate(TOP_CONTRACTS_REPORTED);
+        tallies
+    }
+
     /// The top [`TOP_CONTRACTS_REPORTED`] contracts by full-state bytes.
     fn top_contracts(&self) -> Vec<(ContractInstanceId, u64)> {
         let mut tallies: Vec<(ContractInstanceId, u64)> = self
@@ -444,6 +481,7 @@ struct NotEfficientGateStats {
 fn payload_mix_json(
     arms: &[(PayloadArm, u64, u64)],
     contracts: &[(ContractInstanceId, u64)],
+    not_efficient_contracts: &[(ContractInstanceId, u64)],
     tracked_full_state_bytes: u64,
     contracts_tracked: u64,
     attribution_dropped_sends: u64,
@@ -560,6 +598,19 @@ fn payload_mix_json(
     //     because the cap was already full.
     // Reporting only the second made an 11-contract window look perfectly
     // reconciled while 10 % of its bytes were unaccounted for.
+    // #4956: which contracts the gate actually refused on. The aggregate
+    // ratio says a state-sized "summary" is being fed in; this says by whom.
+    obj.insert(
+        "top_contracts_by_not_efficient_bytes".into(),
+        serde_json::Value::Array(
+            not_efficient_contracts
+                .iter()
+                .map(
+                    |(id, bytes)| serde_json::json!({ "contract": id.to_string(), "bytes": bytes }),
+                )
+                .collect(),
+        ),
+    );
     let top_sum: u64 = contracts.iter().map(|(_, b)| *b).sum();
     let other_contracts_bytes = tracked_full_state_bytes.saturating_sub(top_sum);
     obj.insert("other_contracts_bytes".into(), other_contracts_bytes.into());
@@ -593,6 +644,7 @@ pub(crate) fn emit_payload_mix_rollup(
     let payload = payload_mix_json(
         &window.arms(),
         &window.top_contracts(),
+        &window.top_not_efficient_contracts(),
         window.contract_full_state_bytes.values().sum(),
         window.contract_full_state_bytes.len() as u64,
         window.attribution_dropped_sends,
@@ -659,6 +711,42 @@ mod tests {
 
     fn contract(byte: u8) -> ContractInstanceId {
         ContractInstanceId::new([byte; 32])
+    }
+
+    /// #4956: the refused-delta arm must be attributable to a CONTRACT, not
+    /// just counted in aggregate. The aggregate gate ratio proved a
+    /// state-sized "summary" is being fed in; without this the culprit stays
+    /// anonymous. Only `FullNotEfficient` may land in the narrow map — a
+    /// different full-state arm sharing it would re-create the ambiguity.
+    #[test]
+    fn not_efficient_bytes_are_attributed_to_their_contract() {
+        let mix = PayloadMix::new();
+        mix.record_delivered(
+            PayloadArm::FullNotEfficient,
+            &contract(7),
+            600,
+            Some((600, 600)),
+        );
+        mix.record_delivered(
+            PayloadArm::FullNotEfficient,
+            &contract(7),
+            400,
+            Some((400, 400)),
+        );
+        // A different full-state arm must NOT pollute the narrow map.
+        mix.record_delivered(
+            PayloadArm::FullNoTheirSummaryUntracked,
+            &contract(8),
+            999,
+            None,
+        );
+        let w = mix.take_window();
+        let top = w.top_not_efficient_contracts();
+        assert_eq!(top.len(), 1, "only the refused arm belongs here: {top:?}");
+        assert_eq!(top[0].0, contract(7));
+        assert_eq!(top[0].1, 1000, "per-contract bytes must accumulate");
+        // The wider map still sees both, so the two views stay consistent.
+        assert_eq!(w.contract_full_state_bytes[&contract(8)], 999);
     }
 
     /// Taking the window leaves the accumulator empty so consecutive rollups
@@ -751,6 +839,7 @@ mod tests {
         let json = payload_mix_json(
             &window.arms(),
             &window.top_contracts(),
+            &window.top_not_efficient_contracts(),
             window.contract_full_state_bytes.values().sum(),
             window.contract_full_state_bytes.len() as u64,
             window.attribution_dropped_sends,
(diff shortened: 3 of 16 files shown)


sanity:
## CI caught a real design flaw in the first cut — fixed

The Simulation job failed three tests, and none was a fixture problem. Recording the diagnosis because the failure mode is not obvious.

The gate compared `delta.len() >= our_state_size` directly. Tracing the failure gave the actual pair: **delta_size=144, state_size=136**. So the gate was abandoning a delta to save **8 bytes**, and that is wrong twice over:

1. `test_sustained_update_fanout_no_full_state_storm` pins `delta_sends > full_state_sends`. Flipping every small contract to full state re-creates exactly the #4233 full-state fan-out shape that pin exists to catch. A byte-optimal rule that reintroduces the storm shape is not optimal.
2. The two summary-first PUT tests failed for a different and more serious reason: **that leg's fallback is not full state**. `reverse_delta_from_compute_result` ships NOTHING on a refusal, so the originator never reconciles and has to heal later via GET/anti-entropy. Refusing a slightly-oversized delta there is strictly worse than sending it. My original PR text asserted the full-state fallback is "genuinely optimal" — that is true on the broadcast path and false here.

Deltas also keep the receiver's peer-summary cache warm, which is the thing this whole issue is about, so trading one away for a rounding error is a bad deal independent of the pins.

**Fix:** the gate now requires full state to win by `MIN_FULL_STATE_SAVING_BYTES` (1 KiB) before abandoning a computed delta. The population this gate exists for — the #4956 poisoned-summary contracts whose delta is state-sized at 550-840 KB — clears a 1 KiB margin by orders of magnitude, so the intent is fully intact.

Boundary tests now cover both directions (one byte under the margin ships, exactly at the margin refuses, plus the real 144-vs-136 pair). All three previously-failing simulation tests pass locally, along with 56 `ring::interest` unit tests.

The asymmetry in point 2 is worth noting separately: a refusal means the reverse leg sends nothing at all. That predates this PR (the old pre-compute gate refused the same way) and the margin makes it much rarer, but it is a latent sharp edge for large contracts and probably deserves its own fix.

[AI-assisted - Claude]

sanity:
## Scope note: this PR now also carries the #4956 instrumentation

Adding it here rather than in a separate PR because both touch `broadcast_payload_mix.rs`, so splitting them would manufacture a conflict for no review benefit. They are separate commits and can be read independently. Both are wanted in the same release.

### Part 1 (unchanged): the post-compute delta gate

Already described above, plus the margin fix from the CI round.

### Part 2 (new): measure where the bytes actually go

The motivation is uncomfortable but worth stating plainly: **every bandwidth fix through 0.2.109 was aimed by an instrument that sees about a quarter of the traffic.** Pairing `broadcast_payload_mix` against `resource_utilization`'s `cumulative_bytes_sent` across 544 peers on 0.2.108:

| | |
|---|---|
| real bytes sent | median 299 MB/h, mean 707 MB/h per node |
| broadcast payload | median 61 MB/h, mean 189 MB/h per node |
| payload as share of real | median 14.7 %, **aggregate 26.7 %** |

The candidates for the missing ~73 % — GET/PUT/SUBSCRIBE payloads, the ~5-min InterestSync `Summaries` exchange (which ships a full `StateSummary` per shared contract to every connected peer, every cycle), CONNECT/NAT traffic, and transport framing plus retransmits — differ by orders of magnitude in cost and have completely different remedies. I have an estimate that points at InterestSync, but an estimate is what mis-prioritised the last round, so this measures instead of arguing.

Three additions, all observation-only, no behavior change:

1. **`outbound_message_mix`** — per-message-kind byte census at the single point every non-stream `NetMessage` is written to a connection. The arms *partition* the traffic, so they sum to the node's message total and the residual against `cumulative_bytes_sent` **names the transport overhead rather than hiding it**. `PeerConnection::send` now returns the serialized length it already computed, so nothing is serialized a second time just to measure it. `classify()` is exhaustive with no catch-all, so a future protocol message breaks the build instead of quietly joining `Other`.
2. **`peer_id` on `interest_heartbeat_cycle`** — one line. Its absence is precisely why the InterestSync hypothesis could not be confirmed by correlation.
3. **`top_contracts_by_not_efficient_bytes`** — the aggregate gate-input ratio is **1.000** (summary size == state size, max 839 KB each) on 0.2.107/108/109, which proves some contract is feeding state-sized bytes in as a "summary". The existing per-contract map mixes every full-state arm and cannot say which contract. This narrows it to the one arm that matters.

Stream bytes are deliberately **excluded** — they do not pass this call site and already have accounting via `transfer_completed`. Counting them here would break the partition property the census depends on.

### Hot-path cost

This records on every message, not every delivered broadcast, so it is a hotter path than the payload mix. It is one uncontended `parking_lot` lock acquire plus three integer updates, with no allocation, no map insert and no formatting — everything else happens in the aggregator task. I flagged this as the main risk before writing it and kept it to integer work deliberately.

### Testing

`cargo test -p freenet --lib`: 4317 passed, 0 failed. New unit tests cover the partition property (every byte lands in exactly one arm), max-byte tracking, unique arm indices/stems, window reset, window-seconds clamping, and that only the refused arm lands in the narrow contract map. Clippy clean on the touched files.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4957

### Summary
- **PR Title:** fix(broadcast): post-compute delta gate, plus an outbound-byte census to aim the next fix
- **Type:** fix + feat(telemetry)
- **CI Status:** running on `de2cd1bc` at time of writing; merge is gated on it being green
- **Linked Issues:** #4956
- **Review tier:** **Full** — touches the transport send path, a `PeerConnectionApi` trait signature, and broadcast payload selection. Three of those are named high-risk surfaces.
- **Reviewers run:** Codex attempted and **genuinely unavailable**; diverse-lens fallback run serially (see Independent Review Substitution).

### Independent Review Substitution (required disclosure)

`codex review --base origin/main` was invoked against the PR worktree and failed with a hard usage-limit error: *"You've hit your usage limit... try again at Aug 1st, 2026 2:42 PM."* Gemini CLI remains dead (`IneligibleTierError`, deprecated free tier), so it was not attempted. This was a real invocation, not an assumption.

Waiting was considered and rejected: the reset is ~6 days out, and this PR carries the instrumentation needed to diagnose an active user-facing bandwidth problem, so the change is time-sensitive. Per `~/.claude/rules/multi-model-review.md` I fell back to diverse adversarial lenses, run **serially in-context** (the skill's prescribed fallback when subagents are unavailable):

1. *skeptical bug-hunt* — races, edge cases, failure modes on the new hot path
2. *code-first* — read the implementation before the description, hunt intent/impl mismatches
3. *failure mode specific to this change* — measurement CORRECTNESS: does the instrument mean what it claims, and can it be misread?

This is a weaker signal than a true external model, whose blind spots are uncorrelated with the author's. **Recommend re-running the external pass after Aug 1 if this has not yet shipped.**

---

### Code-First Analysis
**Independent Understanding:** Part 1 stops refusing a delta based on a summary-size proxy and instead compares the computed delta against full state with a 1 KiB switch margin. Part 2 adds a per-message-kind outbound byte census, a `peer_id` on the interest heartbeat, and per-contract attribution for the refused-delta arm.
**Stated Intent:** matches.
**Alignment:** aligned. One earlier misalignment was found and corrected during review (below).

---

### Testing Assessment
**Coverage Level:** adequate

| Unit | ✅ | 4317 pass. New: margin boundary in both directions, real 144-vs-136 pair, census partition property, arm index/stem uniqueness, per-arm contract attribution |
| Integration | ✅ | 3 previously-failing summary-first PUT / fan-out sim tests now pass |
| Simulation | ⚠️ | CI Simulation job pending on current HEAD — **blocking** |
| E2E | N/A | no user-facing surface |

**Regression Test:** present and mutation-verified — flipping the gate operator `>=`→`>` fails the boundary test, so it is not vacuous.

---

### Skeptical Findings
**Risk Level:** low-medium (was medium-high before the two fixes below)

| Gate abandoned deltas to save 8 bytes | **High** | `ring/interest.rs` gate | FOUND BY CI, FIXED. Broke the #4233 storm pin and, worse, the summary-first PUT reverse leg whose fallback is *sending nothing*, not full state. Fixed with a 1 KiB margin |
| Census/`transfer_completed` overlap misdocumented | **Medium** | `outbound_message_mix.rs` module docs | FOUND IN THIS REVIEW, FIXED (`de2cd1bc`). Docs claimed stream bytes bypass the call site; only OPS-level streams do. A `NetMessage` >1.2 KB is fragmented by the transport-level `outbound_stream` and IS counted here *and* by `transfer_completed`. Summing them would double-count — exactly the misreading this instrument exists to prevent |
| Census completeness | Low | verified | No `NetMessage` path bypasses `send_message`; all direct `conn.send()` callers are tests; handshake uses the same route. The partition claim holds |
| Hot-path cost | Low | `record_sent` | One uncontended `parking_lot` acquire + 3 integer ops per message, no alloc/map/format. At observed rates (~155 msg/s/node) this is noise |
| Recording on failure | Low | `handle_peer_channel_message` | Recorded only on `Ok`, matching the payload mix's real-delivery rule, so a failed send cannot inflate the census |

---

### Big Picture Assessment
**Goal Alignment:** yes — Part 2 directly addresses the finding that prior fixes were aimed by an instrument covering ~27% of traffic.
**Anti-Patterns Detected:** none. Notably it does NOT reintroduce a `_ =>` catch-all in classification (code-style rule); `classify()` is exhaustive so a new protocol message breaks the build.
**Removed Code Concerns:** none. `is_delta_efficient` is retained with its tests and a "do not re-wire as a pre-compute refusal" note rather than deleted.
**Scope Assessment:** two concerns in one PR. Justified: both touch `broadcast_payload_mix.rs`, so splitting would manufacture a conflict. They are separate commits and independently readable.

---

### Documentation
- Code docs: complete; the gate's margin rationale and the census's overlap semantics are both documented with the reasoning, not just the behaviour.
- Architecture docs: n/a.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
1. **CI must be green on `de2cd1bc`** before merge — Simulation in particular, since it is what caught the original design flaw.

#### Should Fix (Important)
1. None outstanding — both findings above are fixed in-branch.

#### Consider (Suggestions)
1. The summary-first PUT reverse leg shipping *nothing* on a gate refusal is a latent sharp edge for large contracts. It predates this PR and the margin makes it rarer, but it deserves its own fix.
2. Re-run the external Codex pass after Aug 1 if this has not shipped, given the high-risk surfaces.

---

### Verdict
**State:** **Needs Changes — Blocked on CI only.** No outstanding code findings.
**HEAD SHA reviewed:** `de2cd1bc`

Both findings surfaced during review are already fixed on this HEAD. Note the honest caveat: this review's independent pass was a Claude-lens substitute, not a true external model, and the most serious defect in this PR was caught by **CI**, not by review — which is itself a reminder not to over-trust the fallback.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #4957

### Summary
- **PR Title:** fix(broadcast): post-compute delta gate, plus an outbound-byte census to aim the next fix
- **Type:** fix + feat(telemetry)
- **CI Status:** running on `de2cd1bc` at time of writing; merge is gated on it being green
- **Linked Issues:** #4956
- **Review tier:** **Full** — touches the transport send path, a `PeerConnectionApi` trait signature, and broadcast payload selection. Three of those are named high-risk surfaces.
- **Reviewers run:** Codex attempted and **genuinely unavailable**; diverse-lens fallback run serially (see Independent Review Substitution).

### Independent Review Substitution (required disclosure)

`codex review --base origin/main` was invoked against the PR worktree and failed with a hard usage-limit error: *"You've hit your usage limit... try again at Aug 1st, 2026 2:42 PM."* Gemini CLI remains dead (`IneligibleTierError`, deprecated free tier), so it was not attempted. This was a real invocation, not an assumption.

Waiting was considered and rejected: the reset is ~6 days out, and this PR carries the instrumentation needed to diagnose an active user-facing bandwidth problem, so the change is time-sensitive. Per `~/.claude/rules/multi-model-review.md` I fell back to diverse adversarial lenses, run **serially in-context** (the skill's prescribed fallback when subagents are unavailable):

1. *skeptical bug-hunt* — races, edge cases, failure modes on the new hot path
2. *code-first* — read the implementation before the description, hunt intent/impl mismatches
3. *failure mode specific to this change* — measurement CORRECTNESS: does the instrument mean what it claims, and can it be misread?

This is a weaker signal than a true external model, whose blind spots are uncorrelated with the author's. **Recommend re-running the external pass after Aug 1 if this has not yet shipped.**

---

### Code-First Analysis
**Independent Understanding:** Part 1 stops refusing a delta based on a summary-size proxy and instead compares the computed delta against full state with a 1 KiB switch margin. Part 2 adds a per-message-kind outbound byte census, a `peer_id` on the interest heartbeat, and per-contract attribution for the refused-delta arm.
**Stated Intent:** matches.
**Alignment:** aligned. One earlier misalignment was found and corrected during review (below).

---

### Testing Assessment
**Coverage Level:** adequate

| Unit | ✅ | 4317 pass. New: margin boundary in both directions, real 144-vs-136 pair, census partition property, arm index/stem uniqueness, per-arm contract attribution |
| Integration | ✅ | 3 previously-failing summary-first PUT / fan-out sim tests now pass |
| Simulation | ⚠️ | CI Simulation job pending on current HEAD — **blocking** |
| E2E | N/A | no user-facing surface |

**Regression Test:** present and mutation-verified — flipping the gate operator `>=`→`>` fails the boundary test, so it is not vacuous.

---

### Skeptical Findings
**Risk Level:** low-medium (was medium-high before the two fixes below)

| Gate abandoned deltas to save 8 bytes | **High** | `ring/interest.rs` gate | FOUND BY CI, FIXED. Broke the #4233 storm pin and, worse, the summary-first PUT reverse leg whose fallback is *sending nothing*, not full state. Fixed with a 1 KiB margin |
| Census/`transfer_completed` overlap misdocumented | **Medium** | `outbound_message_mix.rs` module docs | FOUND IN THIS REVIEW, FIXED (`de2cd1bc`). Docs claimed stream bytes bypass the call site; only OPS-level streams do. A `NetMessage` >1.2 KB is fragmented by the transport-level `outbound_stream` and IS counted here *and* by `transfer_completed`. Summing them would double-count — exactly the misreading this instrument exists to prevent |
| Census completeness | Low | verified | No `NetMessage` path bypasses `send_message`; all direct `conn.send()` callers are tests; handshake uses the same route. The partition claim holds |
| Hot-path cost | Low | `record_sent` | One uncontended `parking_lot` acquire + 3 integer ops per message, no alloc/map/format. At observed rates (~155 msg/s/node) this is noise |
| Recording on failure | Low | `handle_peer_channel_message` | Recorded only on `Ok`, matching the payload mix's real-delivery rule, so a failed send cannot inflate the census |

---

### Big Picture Assessment
**Goal Alignment:** yes — Part 2 directly addresses the finding that prior fixes were aimed by an instrument covering ~27% of traffic.
**Anti-Patterns Detected:** none. Notably it does NOT reintroduce a `_ =>` catch-all in classification (code-style rule); `classify()` is exhaustive so a new protocol message breaks the build.
**Removed Code Concerns:** none. `is_delta_efficient` is retained with its tests and a "do not re-wire as a pre-compute refusal" note rather than deleted.
**Scope Assessment:** two concerns in one PR. Justified: both touch `broadcast_payload_mix.rs`, so splitting would manufacture a conflict. They are separate commits and independently readable.

---

### Documentation
- Code docs: complete; the gate's margin rationale and the census's overlap semantics are both documented with the reasoning, not just the behaviour.
- Architecture docs: n/a.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
1. **CI must be green on `de2cd1bc`** before merge — Simulation in particular, since it is what caught the original design flaw.

#### Should Fix (Important)
1. None outstanding — both findings above are fixed in-branch.

#### Consider (Suggestions)
1. The summary-first PUT reverse leg shipping *nothing* on a gate refusal is a latent sharp edge for large contracts. It predates this PR and the margin makes it rarer, but it deserves its own fix.
2. Re-run the external Codex pass after Aug 1 if this has not shipped, given the high-risk surfaces.

---

### Verdict
**State:** **Needs Changes — Blocked on CI only.** No outstanding code findings.
**HEAD SHA reviewed:** `de2cd1bc`

Both findings surfaced during review are already fixed on this HEAD. Note the honest caveat: this review's independent pass was a Claude-lens substitute, not a true external model, and the most serious defect in this PR was caught by **CI**, not by review — which is itself a reminder not to over-trust the fallback.

[AI-assisted - Claude]

Issue #4924, opened 2026-07-25
Linux installer creates system service for home-directory binary, breaking on Fedora SELinux


finalbillybong:
## Summary

The Linux installer installs `freenet` at `~/.local/bin/freenet` but creates a **system-level** unit at `/etc/systemd/system/freenet.service`. On Fedora 44 with SELinux enforcing, this prevents the service from starting initially and, after correcting the file label, runs Freenet in the sensitive `init_t` domain. Normal Freenet behavior then triggers SELinux denials for outbound telemetry and WebAssembly executable memory.

This appears related to the design in #1490 and #2371, which proposed a **user-level** systemd unit for a binary installed under `~/.local/bin`.

## Environment

- Freenet `0.2.105` (`3e4e27b6990b`)
- Build timestamp: `2026-07-22T17:42:54Z`
- Fedora Linux 44 KDE, SELinux targeted/enforcing
- Kernel `7.0.14-201.fc44.x86_64`
- SELinux policy `selinux-policy-targeted-44.3-1.fc44.noarch`

## Reproduction

1. Install Freenet using the installer linked from freenet.org.
2. Reboot, or attempt to start `freenet.service`.
3. Inspect `systemctl status freenet.service` and SELinux alerts.

The generated service has approximately:

```text
FragmentPath=/etc/systemd/system/freenet.service
User=<current user>
ExecStart=/home/<user>/.local/bin/freenet network
```

### 1. Service fails to execute

The downloaded/replaced executable was labeled `user_tmp_t`:

```text
/home/<user>/.local/bin/freenet  unconfined_u:object_r:user_tmp_t:s0
```

At boot, systemd failed with:

```text
freenet.service: Unable to locate executable '/home/<user>/.local/bin/freenet': Permission denied
freenet.service: Failed at step EXEC spawning /home/<user>/.local/bin/freenet: Permission denied
status=203/EXEC
```

Running `restorecon ~/.local/bin/freenet` changed it to `home_bin_t`, allowing execution.

### 2. Freenet runs in `init_t`

With the normal home executable label, the system service runs as:

```text
system_u:system_r:init_t:s0
```

This produced AVCs during normal operation.

Telemetry connection (default alpha telemetry endpoint `http://nova.locut.us:4318`):

```text
avc: denied { name_connect } for comm="freenet-main" dest=4318 \
  scontext=system_u:system_r:init_t:s0 \
  tcontext=system_u:object_r:unreserved_port_t:s0 \
[...]

PR #4958, opened 2026-07-26
fix: restore SELinux file context after installing binaries, create user level service if using user level binary.

Cohors1316:
SELinux blocks freenet execution when mv from /tmp inherits an incorrect file context (gconf_home_t instead of bin_t), makes sure user level services are created for user level installations.


sanity:
@Cohors1316 — apologies for the month of silence on this. It's a well-put-together fix: the `restorecon` diagnosis is right, the routing change is the correct companion to it, and the test coverage (including the `/home/aliceproject` prefix-match case) is better than most of what lands here. `scripts/test-install-sh.sh` runs in CI, so those tests are real coverage.

I've also just approved the workflow runs on this branch — outside-contributor CI needs manual approval and nobody had done it, which is part of why this stalled.

One substantive thing needs a decision before it can merge, and it's ours to make rather than yours to fix.

## The `.local` rule changes every default install, not just SELinux ones

```sh
# scripts/install.sh:596
install_dir="${FREENET_INSTALL_DIR:-$HOME/.local/bin}"
```

`$HOME/.local/bin` is the default for **every** install, including when the script is run as root. So the new

```sh
case "$3" in
    */.local/*) echo "user"; return ;;
esac
```

makes every fresh Linux install choose a user service, on every distro, whether or not SELinux is involved. That's a much wider change than the title suggests, and it reverses a policy the old comment stated deliberately:

> A system service is preferred whenever we can elevate (it runs at boot, survives logout, and is the most reliable choice on headless servers)

Two mitigations, in fairness: existing installs are unaffected, because the `has_system_unit` check in `resolve_service_action` runs first and keeps whatever's already there; and anyone who wants a system service can set `FREENET_INSTALL_DIR=/usr/local/bin`. So this isn't a regression for current operators, only a change for new ones.

There's also a decent argument the broad behaviour is simply *correct* — a system unit whose `ExecStart` points into `/home` is fragile for reasons beyond SELinux (an unmounted `/home` at boot, a deleted user). I'm genuinely unsure which we want, which is why I'm not asking you to change it unilaterally.

**The two options:**

1. **Keep it broad** (what the PR does) — every `$HOME` install becomes a user service. Needs an explicit sign-off, plus the PR title/description updated to say so, since right now it reads as a targeted SELinux fix.
2. **Narrow it to SELinux** — gate the `.local` / `$HOME` branch on SELinux actually being enforcing (`selinuxenabled`, or `getenforce` returning `Enforcing`), so behaviour is unchanged everywhere else and the fix goes exactly where #4924 hurts.

My own lean is (2), because it matches how the PR is framed and has a blast radius equal to the bug. But I'd like @sanity or @iduartgomez to call it — I'll ping them, and I'll come back to you with an answer rather than leaving this open again.

## Small notes, none blocking

- `*/.local/*` is checked before the `$HOME` test, so a system path that happens to contain a `.local` component (`/opt/.local/bin`) also routes to a user service. Contrived, but if you narrow the rule anyway it's free to tighten.
- `_check_restore_skip` in `scripts/test-install-sh.sh` uses a fixed `/tmp/.freenet-restore-test`. A `mktemp` would be tidier; CI jobs get their own runner so it isn't actually a problem today.
- `restorecon -v ... 2>/dev/null || true` correctly no-ops on non-SELinux systems and when `fdev` is absent. Placement is right too — before the `--version` verification, which is what would otherwise fail.

Thanks again, and sorry it took this long to get you a real answer.

[AI-assisted - Claude]

sanity:
Follow-up on the CI now that the runs are unblocked: **`Rule Lint` is red, and it is our bug, not yours.** Do not reshape your tests to satisfy it.

The `fix:`-needs-a-regression-test gate counts added shell assertions with this diff scope (`.github/workflows/ci.yml:~536`):

```sh
NEW_SHELL_TESTS=$(git diff "$BASE"..."$HEAD" -- '**/*_test.sh' ...)
```

Your tests are in `scripts/test-install-sh.sh` — `test-*.sh`, not `*_test.sh` — so the glob never sees them and the counter scores zero. That file *is* run by CI (`ci.yml:278`, `sh scripts/test-install-sh.sh`), and your added `check_eq` calls are exactly the convention the counter recognises; they're just in a file it doesn't look at.

The gate's own comment already documents this as a known naming accident, and the counter's self-test (`scripts/rule_lint_shell_assertion_counter_test.sh:359-361`) explicitly excludes both `test-install-sh.sh` and `test-uninstall-sh.sh` for the same reason. So a `fix:` PR that adds real coverage to either of those two files cannot pass this check today. Yours is the first one to hit it.

The gate's error text names the intended escape hatch for exactly this case:

> If you have a real test the counter cannot see, say so on the PR rather than reshaping the test to please this grep.

So: I've applied the `test-exempt` label, with this comment as the justification. Your regression tests exist, are real, and run in CI on every push — the counter just cannot see the filename.

I'm raising the one-line glob fix separately so the next contributor doesn't hit it.

The rest of the suite is still running; the `Rule Lint` failure was the only immediate red. The service-mode routing question from my previous comment is still the thing that decides whether this merges as-is.

[AI-assisted - Claude]

sanity:
CI results are in. Two reds, both small, and one of them is genuinely yours:

**1. `Fmt` — shellcheck, `scripts/test-install-sh.sh:98`.** This one is real:

```
In scripts/test-install-sh.sh line 98:
    "$(HOME= FREENET_INSTALL_SH_LIB=1 INSTALL="$INSTALL" sh -c '
            ^-- SC1007 (warning): Remove space after = if trying to assign a value
                                  (for empty string, use var='' ... ).
```

`shellcheck -x` runs with warnings fatal on this file list (`ci.yml:276`). `HOME=''` fixes it. Though the subshell immediately does `unset HOME` on the next line, so dropping the `HOME=` prefix entirely would work too and reads better — the `unset` is what the test is actually exercising.

That's the whole failure; the step aborts on it, so `Self-test install.sh service-mode decision` and the uninstall self-test were both **skipped** and have not actually run yet. Worth knowing: your new assertions haven't been executed by CI even once. Clippy passed.

**2. `Rule Lint`** is still showing the stale result from before I applied `test-exempt` — the label only takes effect on a re-run. It'll clear on your next push.

So: one-line shellcheck fix, push, and we get a first real run of your tests. The service-mode policy question from my earlier comment is still the thing that decides whether this merges as-is — I'll come back with an answer on that.

[AI-assisted - Claude]

sanity:
Decision on the service-mode question, so you're not left waiting on it: **please narrow it to the SELinux condition.**

The reasoning isn't that your instinct is wrong — it's scope. A PR titled "restore SELinux file context" shouldn't also flip every default fresh Linux install from a system service to a user service, because `install_dir` defaults to `$HOME/.local/bin` for everyone. That's a behaviour change riding inside a bug fix, and the repo asks for one logical change per PR with behaviour changes agreed up front.

Concretely: gate the `*/.local/*` and `"${HOME}/"*` branches in `decide_linux_service_mode` on SELinux actually being in play (`selinuxenabled`, or `getenforce` returning `Enforcing`), so the PR does what its title says and the fix lands where #4924 hurts. Your existing test table covers this cleanly — the cases just gain a "SELinux on/off" dimension.

**Your idea was kept, not rejected.** I filed #5434 with the full analysis — the `install.sh:596` default, the four independent reasons a system unit pointing into `$HOME` is fragile beyond SELinux, three options including one you didn't raise (change the default install dir for root instead), and the measured blast radius. It's credited to you and unassigned, for @sanity / @iduartgomez to decide. If it goes your way, the broad version lands as its own change with the discussion behind it.

So the remaining work here is: the shellcheck one-liner at `test-install-sh.sh:98`, and the SELinux gate. Both small. Once those are in, this should be straightforward to land.

[AI-assisted - Claude]

sanity:
@Cohors1316 — rather than leave this sitting on you, I've pushed the narrowing I asked for so the PR can move. Shout if you'd rather I backed it out and let you do it.

**What changed:** the `*/.local/*` and `"${HOME}/"*` branches are now gated on SELinux actually being in force. Off SELinux, the routing decision is byte-for-byte what it was before your branch — which was the whole concern, since `install_dir` defaults to `$HOME/.local/bin` for everyone and the ungated rule would have flipped every ordinary Linux install to a user service.

I kept it in your style rather than reshaping it:

- The flag is passed in as `$4`, not probed inside `decide_linux_service_mode`, so the function stays pure and unit-testable alongside `am_root` and `can_elevate`. `resolve_service_action` computes it once via a new `selinux_active` helper.
- Permissive counts as enabled. An install that works only until someone runs `setenforce 1` isn't an install that works.
- Your `"${HOME}/"*` separator handling and its `/home/aliceproject` regression case are untouched — that detail was right and I left it exactly as you wrote it.

**Tests.** Your existing cases are relabelled to say which are SELinux cases, and the off-SELinux behaviour now has its own regression cases at both the `decide` and `resolve` levels. The `resolve`-level one matters: it pins that `resolve_service_action` genuinely consults `selinux_active`, so the gate can't be bypassed somewhere on the way through. I also added a case for an existing system unit still winning on an SELinux box, so a re-run refreshes rather than leaving two units of different kinds.

I verified the new guard by breaking it rather than by watching it pass: with the gate made unconditional, 4 of the new tests fail; with it in place all 23 pass. `shellcheck -x -S warning` is clean on both scripts.

The `restorecon` half of your PR — which is the actual #4924 diagnosis — is unchanged. That part was right from the start.

[AI-assisted - Claude]

sanity:
## Review findings applied — and CI on this branch has never actually run

@Cohors1316 — an independent reviewer went over the branch including my own commit, and found one thing in my change that needed fixing plus several silent-failure paths. All applied in `bd0369db8`. Also an important process note at the end.

### The fix to my fix: a root install must not take the user branch

`curl | sudo sh` resets `HOME` to `/root`, so `install_dir` becomes `/root/.local/bin` — which matches the `*/.local/*` arm. That routed a root install into `setup_service`'s **user** branch, which has none of the four root guards the system branch has, and `install_user_service` (`crates/core/src/bin/commands/service/linux.rs:229`) has none either. It would either fail obscurely for uid 0 (no `XDG_RUNTIME_DIR` under sudo) or install a root-owned user service running the node as root out of `/root` — exactly what `linux.rs:339` refuses on the system path.

This was inherited rather than introduced — your ungated version did it for *every* Linux install, so the SELinux gate strictly narrowed it — but it was still live, and it is now gated on non-root as well.

### Three silent failures, which matter more than they look

- **`selinux_active` no longer depends on `PATH`.** `selinuxenabled` lives in `/usr/sbin`, which a `curl | sh` from cron or a non-login RHEL shell need not have — so the whole fix would have been a silent no-op on exactly some of the affected machines. It now checks `/sys/fs/selinux/enforce` first, which is what `selinuxenabled` tests internally.
- **Its stdout is redirected too, not just stderr.** `resolve_service_action`'s stdout *is* its return value, and `setup_service` matches it with a `case` that had no default arm — so one stray line from a chatty `selinuxenabled` would have installed no service and said nothing. That `case` now has a `*)` arm as well, because "no service, no message" is the one outcome this installer must never produce.
- **A relabel refusal is now reported.** `|| true` is still required, but the user gets the manual `restorecon` command instead of meeting the failure later as a bare `203/EXEC` — the very symptom in #4924.

`${4:-0}` so an omitted flag means "no SELinux" rather than aborting under `set -u`. This file is hand-mirrored into freenet/web and a previous fix was already lost to a bulk resync, so it defends toward pre-#4924 behaviour.

### What this still does not fix, said out loud

`resolve_service_action` returns early when a system unit already exists. That is right — creating a second unit would orphan the first one's data — but it means **the people who reported #4924 get only half the fix**: `restore_install_context` runs unconditionally so their `203/EXEC` is fixed, but the `init_t` AVCs need the user-service routing that early return skips. Only fresh installs are fully fixed.

Rather than leave them with silence, `setup_service` now warns when SELinux + an existing system unit + a home-directory binary coincide, and names the one command that switches it. (The warning lives there, not in `resolve_service_action`, because `warn` writes to stdout — the test suite caught me putting it in the wrong place.)

### Verification

Every new guard was checked by breaking it, not by watching it pass: removing the root guard fails 4 tests; removing the `${4:-0}` default fails 1; restoring the swallow-everything `restorecon` form fails 1. All 31 pass with them in place. `shellcheck -s sh -x` clean on both scripts.

### The blocker, and it is not code

**CI has never run on this branch's current head.** The runs sit at `action_required` — GitHub holds an outside contributor's workflows until a maintainer approves them, and nobody has. The last real CI run here was 2026-08-24 and it failed. Both gates that matter for this PR (`ci.yml:314` shellcheck, `:324` `test-install-sh.sh`) live in workflows that have not executed on `bd0369db8`.

I ran both locally and they pass, but that is not the merge gate, and I deliberately did not approve the runs myself — that hold exists so a person decides before a contributor's code runs in our CI. @sanity needs to approve them. Flagging it because it is the only thing standing between this and merge.

### One companion still needed

Users install via `curl https://freenet.org/install.sh` — the copy in **freenet/web**, `hugo-site/static/install.sh`, which is byte-identical to freenet-core `main` and is hand-synced with nothing automating it. Until a companion PR lands there, this fixes #4924 for nobody who uses the documented install path. Happy to open it once this merges.

Thanks again for the original diagnosis — `mv(1)` preserving the SELinux xattr while the auto-updater's `fs::copy` does not is exactly right, and it is why `install.sh` really is the culprit.

[AI-assisted - Claude]

sanity:
Merging. @sanity approved the workflow runs and confirmed the service-type change is the intended default, so both halves land together.

Recording the scope question explicitly, since it was raised and is the thing a future reader will want: this PR contains **two independent changes**, and only one of them is a pure bug fix.

**The `restorecon` half is the bug fix.** @Cohors1316's diagnosis is exactly right and worth restating: `mv(1)` preserves the SELinux xattr while the auto-updater's `fs::copy` does not, which is why `install.sh` is the culprit and the update path is not. It runs unconditionally, so it fixes the reported `203/EXEC` for **everyone, including existing installs**.

**The routing half is a policy decision** — on SELinux, with a home-directory binary, a fresh install now gets a *user* service rather than a *system* service. It is gated on SELinux actually being in force and on the installer not running as root, so off SELinux the decision is byte-for-byte what it was before (verified across 132 inputs). Its blast radius is fresh SELinux installs only: `resolve_service_action` returns early when a system unit already exists, so an upgrading user keeps theirs.

**That means the people who reported #4924 get the relabel but not the routing.** `setup_service` now warns when SELinux + an existing system unit + a home-directory binary coincide, and names the one command that switches it — so they are told, rather than left with a service that still gets denied. #4924 should stay open for that half.

**Two companion items that are not in this PR:**

1. Users install via `curl https://freenet.org/install.sh` — the copy in **freenet/web**, `hugo-site/static/install.sh`, which is hand-synced with nothing automating it. Until that lands, this fixes #4924 for nobody using the documented path. Happy to open it.
2. The SELinux probe itself (`selinux_active`) is exercised only through stubs in the suite; the gate is mutation-verified, the probe is not.

Verification on this branch: all 31 install-script tests pass, `shellcheck -s sh -x` clean on both scripts, and every guard was checked by breaking it — removing the root guard fails 4 tests, removing the `${4:-0}` default fails 1, restoring the swallow-everything `restorecon` form fails 1.

Thanks @Cohors1316 — the original diagnosis was the hard part, and it was right.

[AI-assisted - Claude]

Issue #4961, opened 2026-07-26
Official River room updates take 4-20s to propagate: full-state sends hit the 2-stream large-payload limit


sanity:
## Why updates to the official River room take 4-20 s to propagate

Reported symptom: messages in the Freenet Official room (`4XwkMCQ4g1goJUzqVwdGgqEu2xgKHj7nge1GEXDDv4Yq`) take 4-20 s, sometimes longer, to appear. Individual chat messages are small, so this should be milliseconds.

It is not a routing or subscription problem. It is the full-state bandwidth bug (#4956) plus a concurrency limit that multiplies it.

### The chain

**1. The room's updates go out as FULL STATE, not deltas.**

The contract is confirmed in the full-state population — 0.57 GB of full-state broadcast bytes in a recent window (11.8 % of all fleet full-state bytes, 2nd highest contract). Fleet arm mix on 0.2.109 (3,257 node-minutes):

| arm | share of bytes | sends | mean payload |
|---|---|---|---|
| `full_no_their_summary_tracked` | 31.4 % | 3,262 | **455 KB** |
| `delta` | 27.0 % | 178,524 | 7.2 KB |
| `full_no_their_summary_untracked` | 21.3 % | 9,197 | 109 KB |
| `full_not_efficient` | 16.0 % | 1,521 | 498 KB |

A chat message that should be a ~5-7 KB delta is instead a **380-500 KB full state**, because the sender has no cached summary for the receiver.

**2. Anything >= 64 KB is classed "large" and rate-limited to 2 concurrent streams.**

`broadcast_queue.rs`:
```
DEFAULT_SMALL_PAYLOAD_CONCURRENCY: usize = 12;   // deltas, chat messages
DEFAULT_LARGE_PAYLOAD_CONCURRENCY: usize = 2;    // full state
```
The limit is reasonable on its own terms (don't saturate the uplink with whole states). It becomes a latency multiplier only because normal chat traffic is wrongly taking the large path.

**3. Large streams are slow per hop.** Measured from `transfer_completed` (same-peer timing, so clock skew does not apply):

| payload | median | p90 | p99 | max |
|---|---|---|---|---|
| < 64 KB | **0 ms** | 9 ms | 2.9 s | — |
| 64-256 KB | 2 ms | 65 ms | 1.7 s | — |
| ops 256 KB-1 MB | **378 ms** | 1.4 s | 8.2 s | 27 s |
[...]

PR #4962, opened 2026-07-26
feat(telemetry): tag why a tracked peer has no cached summary

sanity:
## Problem

`full_no_their_summary_tracked` — a full-state broadcast to a peer we DO track but hold no cached summary for — is **26.9% of broadcast bytes at a 357 KB mean** on the aged 0.2.109 fleet (605 peers, 3,772 node-minutes). That makes it the single largest remaining bandwidth arm, and the leading suspect for the 4-20s propagation latency on the official River room (#4961).

The rollup cannot say *why* the summary is absent, and the three possible causes have three completely different fixes:

| cause | fix shape |
|---|---|
| entry created summaryless, never seeded | the registration/heartbeat chain isn't firing, or an `Interests` full-replace wipes it every ~5 min |
| the peer reported `None` and we discarded a summary we had seeded from a real delivery | stop discarding on an unconfirmed `None` report |
| a resync or delta-apply failure cleared it | expected; self-heals on next delivery |

Only the third is measurable today, at **~8%** of tracked sends (381 resync requests vs ~4,600 tracked sends). The other ~92% splits between the first two by an unknown ratio, so a fix aimed at either one is a guess.

That matters here specifically. In this investigation, the two times mechanism was guessed at (summary doubling; River summaries being oversized) the guess was wrong; the two times it was measured (the ops/transport stream split; gate inputs at ratio 1.000) it produced a real finding. This PR buys the measurement before spending a release on the fix.

## Approach

Tag the absence at its source rather than infer it at the send.

`PeerInterest` gains a `summary_missing_reason`, and clearing a summary now goes through `clear_peer_summary(contract, peer, reason)`. The reason is a by-value parameter, so an untagged clear is **unrepresentable** — a future clear site cannot silently land in the wrong bucket. `update_peer_summary` is correspondingly narrowed to a non-`Option` summary, so caching and clearing are distinct at the type level (the `.claude/rules/bug-prevention-patterns.md` "paired fields → let the type system forbid it" shape). The three production clear sites each name their own cause.

The reason is read at the one site that classifies the arm and reported as `tracked_missing_<reason>_{sends,bytes}` on `broadcast_payload_mix`, alongside a `tracked_missing_unattributed_*` residual.

The residual is deliberate, not defensive padding. `their_summary` is read at the top of `broadcast_to_single_peer` and the reason is read much later, past an `.await`, so a concurrent upsert between the two yields no reason. A future untagged clear would do the same. Both surface as an explicit residual instead of inflating a cause that did not produce them.

**No behavior change.** `clear_peer_summary` does exactly what `update_peer_summary(.., None)` did; everything else is counters.

## Testing

Unit tests:
- the accessor returns no reason while a summary IS cached — `summary_absence` keeps its last value once a summary is re-cached, so a raw field read would attribute a live peer to whichever path last cleared it
- each of the four clear paths is distinguishable, and re-caching hides the reason again
- indices and labels are distinct (a collision would silently merge two causes into one bucket)
- per-reason bytes reconcile exactly against the arm total
- an untagged tracked send reports as unattributed, not folded into a reason
- a reason paired with a non-tracked arm is ignored (same discipline `gate_inputs` follows)

Source pins — the three existing ones now also require the reason tag, plus a new one:
- `broadcast_to_single_peer_tags_every_payload_arm_pin` now pins that the arm **reads** the reason. Mutation-testing found this gap: replacing the assignment with a no-op left every other test green while leaving the reason at `None` forever, so all tracked sends would land in `unattributed` — the split would emit, reconcile, and answer nothing.

Mutation-verified (each mutation confirmed to fail the intended test):
1. dropping the accessor's `is_none()` guard → fails the mis-attribution test
2. folding untagged sends into `NeverPopulated` → fails the residual test
3. dropping the reason read at the classification site → fails the new pin (this one passed everything before the pin was added)

Full lib suite: 4323 passed, 0 failed. CI-shaped `cargo clippy --all-features -- -D warnings`: clean.

## What this answers

One 60s rollup window after adoption tells us which of the three paths owns the ~92%, and the fix goes into the next release aimed rather than guessed.

Refs #4961, #4952

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2823,9 +2823,11 @@ async fn handle_interest_sync_message(
                                     .interest_manager
                                     .upsert_peer_summary(&contract, &pk, theirs);
                             }
-                            None => op_manager
-                                .interest_manager
-                                .update_peer_summary(&contract, &pk, None),
+                            None => op_manager.interest_manager.clear_peer_summary(
+                                &contract,
+                                &pk,
+                                crate::ring::interest::SummaryMissingReason::ClearedByNoneReport,
+                            ),
                         }
 
                         if is_stale && !stale_contracts.contains(&contract) {
@@ -3173,9 +3175,11 @@ async fn handle_interest_sync_message(
             // Clear cached summary for this peer
             let peer_key = get_peer_key_from_addr(op_manager, source);
             if let Some(ref pk) = peer_key {
-                op_manager
-                    .interest_manager
-                    .update_peer_summary(&key, pk, None);
+                op_manager.interest_manager.clear_peer_summary(
+                    &key,
+                    pk,
+                    crate::ring::interest::SummaryMissingReason::ClearedByResync,
+                );
             }
 
             // Get PeerKeyLocation for telemetry
@@ -3990,10 +3994,13 @@ mod tests {
              untracked-no-op update path (#4952)"
         );
         assert!(
-            body.contains("update_peer_summary(&contract,&pk,None"),
-            "Summaries arm must keep clear-only update semantics for a None \
-             report — creating an entry just to hold None would relabel \
-             untracked traffic as tracked without enabling deltas"
+            body.contains("clear_peer_summary(&contract,&pk,")
+                && body.contains("SummaryMissingReason::ClearedByNoneReport"),
+            "Summaries arm must keep clear-only semantics for a None report \
+             (creating an entry just to hold None would relabel untracked \
+             traffic as tracked without enabling deltas), and must tag the \
+             clear ClearedByNoneReport so #4961 can attribute the \
+             full_no_their_summary_tracked arm to this path"
         );
     }
 
@@ -6632,22 +6639,27 @@ mod tests {
         fn resync_request_handler_clears_cached_peer_summary() {
             let arm = resync_request_arm();
             assert!(
-                arm.contains("update_peer_summary"),
-                "ResyncRequest handler no longer calls update_peer_summary. \
+                arm.contains("clear_peer_summary"),
+                "ResyncRequest handler no longer calls clear_peer_summary. \
                  #4145 caching relies on this handler clearing the sender's \
                  cached summary so a delta-apply failure forces a fresh \
                  full-state resend instead of looping on unappliable deltas."
             );
-            // The clear MUST pass `None` (clear), not a `Some(summary)` cache.
-            // Strip whitespace so the multi-line call (`op_manager\n
-            // .interest_manager\n .update_peer_summary(&key, pk, None);`)
-            // matches regardless of formatting.
+            // The clear MUST go through the tagged clear API, not a
+            // `update_peer_summary` cache-write. Strip whitespace so the
+            // multi-line call matches regardless of formatting.
             let collapsed: String = arm.chars().filter(|c| !c.is_whitespace()).collect();
             assert!(
-                collapsed.contains("update_peer_summary(&key,pk,None)"),
-                "ResyncRequest handler must clear the cached summary with \
-                 `update_peer_summary(&key, pk, None)` (the `None` clears it). \
-                 Caching a summary here instead would defeat the #4145 backstop."
+                collapsed.contains("clear_peer_summary(&key,pk,"),
+                "ResyncRequest handler must clear the cached summary via \
+                 `clear_peer_summary`. Caching a summary here instead would \
+                 defeat the #4145 backstop."
+            );
+            assert!(
+                collapsed.contains("SummaryMissingReason::ClearedByResync"),
+                "the clear must be tagged ClearedByResync — #4961 needs the \
+                 full_no_their_summary_tracked arm attributed per clear path, \
+                 and an untagged clear would silently land in another bucket"
             );
         }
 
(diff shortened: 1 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #4962

### Summary
- **PR Title:** feat(telemetry): tag why a tracked peer has no cached summary
- **Type:** feat (telemetry / instrumentation)
- **Review tier:** **Full** — the diff touches `ring/interest.rs` (subscription/interest state) and the broadcast send path, and narrows a manager API across many call sites. No high-risk *semantic* surface actually changes, so Light was arguable; per the review rule, when torn, take Full.
- **Reviewers run:** subagent spawning is unavailable in this session, so the skill's prescribed fallback was used — distinct adversarial lenses run **serially in-context**, each committed to in writing before the next: skeptical bug-hunt, code-first, change-specific (telemetry mis-attribution), big-picture, testing. External (Codex) pass **not** attempted: per `~/.claude/rules/multi-model-review.md` as amended 2026-07-26, external models are opt-in only and were not requested.

### Hosting-invariants statement

Required by `.claude/rules/hosting-invariants.md` for anything touching subscription state. **No invariant is touched.** `clear_peer_summary` does exactly what `update_peer_summary(.., None)` did; the rest is counters. Retention, eviction ranking, placement, and freshness are all untouched, and there is no wire change.

---

### Code-First Analysis
**Independent understanding:** `PeerInterest` gains a private `summary_absence` tag; clearing a summary moves to a `clear_peer_summary(contract, peer, reason)` API while `update_peer_summary` narrows to a non-`Option` summary. Three production clear sites name their cause. The tag is read where the `FullNoTheirSummaryTracked` arm is decided and reported per-reason on `broadcast_payload_mix`, with an unattributed residual.

**Alignment with stated intent:** matches. The type-level split (clear takes a by-value reason; cache takes a non-`Option` summary) genuinely makes an untagged clear unrepresentable, which is what the description claims.

---

### Skeptical Findings

| Concern | Severity | Location | Outcome |
|---|---|---|---|
| Out-of-range reason index panics inside a held mutex on the broadcast send path | **High** | `broadcast_payload_mix.rs` `record_delivered` | **FIXED** (79093fa6) |
| Inline comment claims the reason is "always readable" at the classification site | Low | `broadcast_queue.rs` | **FIXED** (79093fa6) |
| Module docs describe the tracked arm without the new split | Low | `broadcast_payload_mix.rs` | **FIXED** (4b3cfa92) |
| `register_peer_interest` full-replaces the entry — a re-registration with `None` would wipe a good summary and be mis-tagged `NeverPopulated` | — | `interest.rs` | **Not a defect.** Both production callers (`node.rs:2663`, `node.rs:3031`) check `get_peer_interest(..).is_some()` first and call `refresh_peer_interest` when present; `register_peer_interest` runs only for a genuinely absent entry. The remaining `(.., None, true)` callers are inside `mod tests`. A pin at `node.rs:4226` already guards this ordering. |
| A direct `interest.summary = None` write would bypass the tag (`summary` is `pub`) | — | crate-wide | **Not present.** `grep '\.summary = '` across `crates/core/src/` returns only the two accessor bodies (`set_summary`, `clear_summary`). No bypass exists today. |

**On the high-severity finding:** adding a variant to `SummaryMissingReason` forces `index()` and `as_str()` to be updated (both exhaustive matches) but does **not** force `ALL` to be extended — and `ALL.len()` sizes the counter arrays. The sequence "add variant, fix the two compile errors, forget `ALL`" yields an out-of-range index, which as `[r]` panics. Now bounds-checked: the miss degrades to the unattributed residual (visible) instead of killing a send path.

---

### Testing Assessment

| Unit | ✅ | accessor guard; all four clear paths distinguishable; index/label distinctness; per-reason reconciliation; untagged→residual; reason ignored on non-tracked arms |
| Source pins | ✅ | three existing pins strengthened to require the reason tag; one new pin that the arm **reads** it |
| Simulation | ✅ | `simulation_integration` passed locally (exit 0) |
| Mutation | ✅ | see below |

**Mutation-verified** — each mutation confirmed to fail the intended test, so none of these pins is vacuous:
1. drop the accessor's `is_none()` guard → fails the mis-attribution test
2. fold untagged sends into `NeverPopulated` → fails the residual test
3. drop the reason read at the classification site → **passed everything**, which is what prompted adding the fourth pin; now fails it

Full lib suite: **4323 passed, 0 failed**. `cargo clippy --all-features -- -D warnings`: clean.

**Coverage gaps (accepted, not blocking):**
- No end-to-end test that the three *production* clear sites emit the right reason through a live InterestSync handler — covered by source pins, which is the established pattern in this file for exactly this shape.
- No test for the out-of-range degradation; the variant cannot be constructed from outside. Documented at the call site instead.

---

### Big Picture Assessment
**Goal alignment:** yes. `full_no_their_summary_tracked` is 26.9% of broadcast bytes on the aged fleet with three candidate causes and only ~8% currently attributable; this converts the remaining ~92% into a measurement.

**Scope:** the API narrowing is wider than "add a field", but it is what makes the tag unforgeable rather than advisory — the `.claude/rules/bug-prevention-patterns.md` "paired fields → let the type system forbid it" shape. Kept to that; no unrelated cleanup rode along.

**Removed code:** none. Three pin tests were strengthened, not weakened.

**Behavior change:** none. No wire change, no protocol enum change, so no mixed-version concern.

---

### Recommendations

#### Must Fix (Blocking)
None outstanding — the one high-severity finding was fixed in this PR.

#### Consider
1. If `NeverPopulated` comes back dominant, the follow-up will still need to separate "seeding never fired" from "an `Interests` full-replace wiped it". The guard at `node.rs:2663` argues the latter shouldn't happen; worth confirming against the data rather than assuming.

---

### Verdict
**State:** Ready to Merge, pending CI green on the current HEAD.
**HEAD SHA reviewed:** `4b3cfa920ed16b29cdffbd5fcb6fdc599b5c187a`

All findings from this pass are fixed. Re-verifying CI on the final SHA before merge, per the per-commit-SHA rule.

[AI-assisted - Claude]

PR #4969, opened 2026-07-27
fix(tracing): make the local event log opt-in on network nodes and stop fsyncing it

### LINKED ISSUE #4968: perf: _EVENT_LOG fsyncs every 5 events — 95% of all fsyncs, 4-6x write amplification, on by default for every user
A normal Freenet peer writes far more to disk than it needs to, and almost all of it is a local diagnostic log nothing currently reads.

## Measurements

`/proc/<pid>/io` on real nodes (v0.2.111), `write_bytes` is the block-layer counter GNOME System Monitor reports:

| node | uptime | written | rate |
|---|---|---|---|
| Framework laptop (normal desktop peer, real NAT) | 1.73 h | 3.83 GB | ~2.2 GB/hr |
| nova hosted node | 3.03 h | 4.53 GB | ~1.5 GB/hr |
| nova gateway | 3.30 h | 2.30 GB | ~0.70 GB/hr |

The rate is bursty: sampled windows on the Framework peer ranged from 264 MiB/hr (quiet) to 4.4 GiB/hr (busy).

**Write amplification is consistent and large.** Block-layer bytes divided by bytes the process actually issued via `write()` measured 2.5x, 4.4x, 5.2x and 6.2x across separate windows. In one 300s window 698 MiB/hr of logical writes became 4361 MiB/hr on disk.

## Cause

`strace -f -e trace=fsync,fdatasync,write,pwrite64` on the Framework peer over 60s:

```
fsync calls      : 55 (with fd captured)
  fd 15 -> ~/.local/share/freenet/_EVENT_LOG.0000012424   52  (95%)
  fd 142 -> /memfd:wasm-memory-image (deleted)             2
  fd 13 -> ~/.local/share/freenet/db/db                    1
block written    : 6.0 MiB
syscall written  : 2.4 MiB
```

**52 of 55 fsyncs go to the event log.** `LogFile::write_all` (`crates/core/src/tracing/aof.rs`) calls `sync_all()` after every `BATCH_SIZE` (5) events. `sync_all` flushes data *and* metadata, and since an append also changes the file size, each one forces a full filesystem journal commit. Roughly 20 KiB of logical append costs ~112 KiB of real block writes.

Independently, the log itself runs ~61 MiB/hour of appends (measured from segment mtime span over 1.57 h: 21 segments, 100.8 MB). Plain text logs are only ~3.3 MiB/hour, so they are not the problem.

## Nothing reads it

Traced every consumer of the network-mode `_EVENT_LOG`:

- `freenet service report` / NodeDiagnostics — does not read it
- `fdev verify-state` — reads the Local-mode `_EVENT_LOG_LOCAL`, not this one
- `event_aggregator` (`read_all_events`) — only reachable from `test_utils.rs`
- `examples/analyze_event_logs.rs` — a manual dev tool

It is also **not** the telemetry behind telemetry.freenet.org. `TelemetryReporter` is a separate sink fed in-memory off the same event stream (`node.rs` builds a `DynamicRegister` over both), so the dashboard does not depend on the disk log at all.

So every user peer pays ~61 MiB/hour of appends and 95% of its fsyncs for a forensic record we have never harvested.

## Fix

1. Default the log OFF in network mode (what end users run), ON in local mode (a dev mode where the log is the point and `fdev verify-state` consumes it). Add `--enable-event-log` for nodes we operate.
2. Drop the per-batch fsync so it is cheap even where enabled. `rotate_segment` already fsyncs a segment before closing it, and a torn tail is already truncated at the last good record offset on load, so durability that matters is unaffected.

Reported by a user asking whether 1.5 GB of disk writes over a few hours was normal. It is — our own peers write more.

[AI-assisted - Claude]

sanity:
## Problem

A normal Freenet peer writes far more to disk than it needs to. Measured on a live v0.2.111 peer (Framework laptop, real NAT), `/proc/<pid>/io` showed **3.83 GB written in 1.73 h**, with block-layer writes running **4-6x** the bytes the process actually issued via `write()`.

`strace` over 60s found **52 of 55 fsyncs (95%) going to `_EVENT_LOG`** — one to redb, two to a wasm memory image. `LogFile::write_all` called `sync_all()` after every `BATCH_SIZE` (5) events; `sync_all` flushes data *and* metadata, and since an append also changes file size, each one forced a full filesystem journal commit. ~20 KiB of logical append cost ~112 KiB of real block writes.

The log also runs ~61 MiB/hour of appends on every peer. Plain text logs are only ~3.3 MiB/hour, so they are not the problem.

Full measurements in #4968.

## Approach

The event log is a purely **local** diagnostic record. It is **not** the telemetry behind telemetry.freenet.org — `TelemetryReporter` is a separate sink fed in-memory off the same event stream (`node.rs` builds a `DynamicRegister` over both), so nothing here changes what the dashboard sees.

Traced every consumer of the network-mode log before touching it:

| consumer | reads `_EVENT_LOG`? |
|---|---|
| `freenet service report` / NodeDiagnostics | no |
| `fdev verify-state` | no — reads Local-mode `_EVENT_LOG_LOCAL` |
| `event_aggregator` (`read_all_events`) | only from `test_utils.rs` |
| `examples/analyze_event_logs.rs` | manual dev tool |

So every user peer paid ~61 MiB/hour and 95% of its fsyncs for a forensic record we have never harvested.

**1. Default it OFF in network mode, ON in local mode.** Network mode is what end users run. Local mode is a single-node dev mode where the log is the point and `fdev verify-state` consumes it. `--enable-event-log` / `FREENET_ENABLE_EVENT_LOG` / `enable-event-log` turns it on for nodes we operate.

**2. Drop the per-batch fsync** so it is cheap even where enabled. Durability that matters is untouched: `rotate_segment` still fsyncs a segment before closing it, and a torn tail is already truncated at the last good record offset on load.

`write_all` still calls `flush()`. That is not a durability barrier — `tokio::fs::File` buffers internally and hands writes to a background pool, so without it the bytes may never reach the OS and a reader can observe a zero-length segment. This was caught by the existing `zero_byte_stray_segment_is_recovered` test, not by inspection.

### The one subtle bit

`Config::enable_event_log` is `Option<bool>`, not `bool`, deliberately. A `config.toml` from an older release has no such key; a `#[serde(default)] bool` would deserialize that absence to `false`, indistinguishable from an operator explicit `false`, and the merge would pin it — silently stripping the log from every upgrading **local**-mode node and breaking `fdev verify-state`. That is the #3890/#4275 silent-revert class this repo has been bitten by twice.

## Testing

New tests, each **mutation-verified** (I reintroduced the bug and confirmed the test goes red, then reverted):

| test | mutation that kills it |
|---|---|
| `event_log_defaults_off_in_network_mode` | flip the mode default to `true` |
| `event_log_defaults_on_in_local_mode` | — |
| `event_log_explicit_setting_overrides_mode_default` | — |
| `event_log_absent_config_key_does_not_pin_local_mode_off` | treat absent key as explicit `false` |
| `event_log_persisted_true_is_honored_on_reboot` | — |
| `write_all_does_not_fsync_per_batch` (source pin) | restore `sync_all()` |

Each mutation killed **exactly** its target test and nothing else.

The exhaustive `all_persisted_config_fields_round_trip_through_build` guard seeds `Some(false)` in Local mode — a genuinely non-default value, so a dropped field fails the assert rather than passing by coincidence.

The source pin exists because a re-added fsync changes only I/O scheduling, not observable output, so no behavioral test can catch it. It cuts at `mod tests` (not `#[cfg(test)]`, which would truncate before most production code and go vacuous) and asserts the extracted slice really is the function body before checking it.

- `config::` suite: 103 passed
- `tracing::` suite: 116 passed
- CI-style `cargo clippy --locked -p freenet -- -D warnings`: clean

## Risk

Behavioral change to a default that affects every user, but no wire-format or protocol change. The failure mode to watch is a node we operate silently losing its log; `--enable-event-log` is the mitigation and the round-trip test covers the persistence path.

Closes #4968

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -197,7 +197,7 @@ pub struct ConfigArgs {
     ///
     /// Measured on a live 0.2.111 peer, writing it costs ~61 MiB/hour of
     /// appends and accounted for 95% of every fsync the process issued
-    /// (#4966). Enable it on nodes you operate and want to post-mortem.
+    /// (#4968). Enable it on nodes you operate and want to post-mortem.
     #[arg(
         long = "enable-event-log",
         env = "FREENET_ENABLE_EVENT_LOG",
@@ -563,7 +563,7 @@ impl ConfigArgs {
             // these are new fields, so a plain get_or_insert is correct.
             self.hosting_disk_pct.get_or_insert(cfg.hosting_disk_pct);
             self.max_hosting_disk.get_or_insert(cfg.max_hosting_disk);
-            // #4966. `cfg.enable_event_log` is itself an Option, so an older
+            // #4968. `cfg.enable_event_log` is itself an Option, so an older
             // config.toml with no such key merges as `None` and leaves the
             // mode-dependent default intact rather than pinning `false`.
             if let Some(persisted) = cfg.enable_event_log {
@@ -1296,7 +1296,7 @@ pub struct Config {
     /// mode-dependent default.
     ///
     /// NOT related to the telemetry that feeds telemetry.freenet.org; see the
-    /// `ConfigArgs::enable_event_log` docs (#4966).
+    /// `ConfigArgs::enable_event_log` docs (#4968).
     #[serde(
         default,
         rename = "enable-event-log",
@@ -2776,7 +2776,7 @@ impl Config {
     ///
     /// This does NOT gate the telemetry that feeds telemetry.freenet.org —
     /// that is a separate `TelemetryReporter` sink fed in-memory off the same
-    /// event stream (#4966).
+    /// event stream (#4968).
     pub fn event_log_enabled(&self) -> bool {
         self.enable_event_log
             .unwrap_or(matches!(self.mode, OperationMode::Local))
@@ -4019,7 +4019,7 @@ mod tests {
     }
 
     /// Build a `ConfigArgs` rooted at `dir` in the given mode. Shared by the
-    /// #4966 event-log default tests so each case differs only in what it sets.
+    /// #4968 event-log default tests so each case differs only in what it sets.
     fn event_log_args(dir: &std::path::Path, mode: OperationMode) -> ConfigArgs {
         ConfigArgs {
             mode: Some(mode),
@@ -4047,7 +4047,7 @@ mod tests {
         }
     }
 
-    /// #4966: a network-mode node (what end users run) must NOT write the local
+    /// #4968: a network-mode node (what end users run) must NOT write the local
     /// diagnostic event log by default. On a live 0.2.111 peer that log was
     /// ~61 MiB/hour of appends and 95% of every fsync the process issued.
     #[tokio::test]
@@ -4063,7 +4063,7 @@ mod tests {
         );
     }
 
-    /// #4966: local mode is a single-node dev mode where the log is the point,
+    /// #4968: local mode is a single-node dev mode where the log is the point,
     /// and `fdev verify-state` consumes `_EVENT_LOG_LOCAL`. It stays ON.
     #[tokio::test]
     async fn event_log_defaults_on_in_local_mode() {
@@ -4099,12 +4099,12 @@ mod tests {
         );
     }
 
-    /// Regression (#4966, the #3890/#4275 silent-revert class): building twice
+    /// Regression (#4968, the #3890/#4275 silent-revert class): building twice
     /// against the SAME config dir must not flip local mode's default off.
     ///
     /// The first build persists a `config.toml`. Because `enable_event_log` is
     /// `None` it is `skip_serializing_if`-omitted, so that file is byte-identical
-    /// in this respect to one written by a pre-#4966 release. If the merge step
+    /// in this respect to one written by a pre-#4968 release. If the merge step
     /// treated the absent key as an explicit `false`, the second build would
     /// silently strip the event log from every upgrading local-mode node and
     /// break `fdev verify-state`. It must still resolve to ON.
@@ -4124,7 +4124,7 @@ mod tests {
         assert!(
             !persisted.contains("enable-event-log"),
             "precondition: an unset event-log flag must be omitted from config.toml, \
-             otherwise this test is not exercising the pre-#4966 upgrade shape. Got:\n{persisted}"
+             otherwise this test is not exercising the pre-#4968 upgrade shape. Got:\n{persisted}"
         );
 
         let second = event_log_args(temp_dir.path(), OperationMode::Local)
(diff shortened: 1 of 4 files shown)


sanity:
## Post-merge review (deploy gate for 0.2.112)

This PR merged with no review posted, and I'm cutting 0.2.112 from it, so I ran the review against the deploy gate rather than ship a fleet-wide default change unreviewed. Lenses run sequentially in-session (skeptical / code-first / change-specific-failure-mode); Codex not attempted per the current multi-model-review rule.

**Verdict: no blocking findings.** Release proceeding. One non-blocking finding below, which I'm fixing now.

### Verified correct

- **`write_all` flush vs. fsync.** The obvious hazard here would be `file.get_mut()` reaching *through* a buffered writer, leaving bytes stranded in a userspace buffer that `flush()` never touches. It doesn't: `LogFile::file` is `Option<BufReader<File>>` (`aof.rs:123`), a **BufReader**, and both the `write_all` and the `flush` go through `get_mut()` to the same inner `tokio::fs::File`. So the flush acts on exactly the object the write went to. Correct as written.
- **`EventFlushHandle::noop()` cannot hang.** This one matters because every network-mode node now takes this path on shutdown. `mpsc::channel(1)` with the receiver dropped immediately leaves the channel *closed*, and tokio's `send` on a closed channel returns `Err` immediately rather than blocking. The `if ... .is_ok()` guard (`register.rs`) then skips the `rx` await entirely. No shutdown stall, and the reply wait is timeout-bounded at 2s regardless.
- **Config back-compat.** `enable_event_log: Option<bool>` with `unwrap_or(matches!(self.mode, OperationMode::Local))` resolves the per-mode default, and an older persisted config with no key reads as `None` and re-derives rather than pinning a stale value. Explicit flag/env/key wins.
- **Source pin quality.** `write_all_does_not_fsync_per_batch` cuts at `mod tests` (not at the first `#[cfg(test)]`, which would have made it vacuous), builds its needle with `concat!` so it can't match its own text, and carries an explicit vacuity guard asserting the extracted slice really contains the append call. That is the right shape.

### Non-blocking finding: integration-test failure diagnostics are now vacuous

`TestHarness::event_log_path` (`crates/core/src/test_utils.rs:1523-1526`) hardcodes `_EVENT_LOG` with the comment *"Nodes run in Network mode, so they create _EVENT_LOG not _EVENT_LOG_LOCAL"*. That comment is now false: network mode is exactly the mode where the log is off.

Nothing in `test_utils.rs` or `crates/freenet-macros/` sets `enable_event_log` — the identifier appears only in `config.rs` and `node.rs`. So network-mode test nodes write no event log, and the chain `generate_failure_report` → `aggregate_events` → `event_log_path` now aggregates an absent file. The `#[freenet_test]` macro runs this on failure by default (`AggregateEventsMode::OnFailure`), so **the event-log section of every integration-test failure report is now empty.**

No production impact, which is why it isn't blocking. But it silently removes the diagnostics the next person gets when an integration test fails, and it fails *quietly* — an empty report reads like "nothing interesting happened," not "this feature is off." That's the vacuous-fixture shape.

Fix is small: have the test harness opt in explicitly (`enable_event_log: Some(true)`), which also restores the comment's truth. Opening that now.

[AI-assisted - Claude]

PR #4973, opened 2026-07-27
fix(test): re-enable the event log for harness nodes so failure reports aren't empty

### LINKED ISSUE #4972: Integration-test failure reports lost their event-log section after #4969
## Problem

#4969 made the local `_EVENT_LOG` opt-in on network-mode nodes. That is correct for real peers — measured on a live 0.2.111 peer it was ~61 MiB/hour of appends and **95% of every fsync the process issued**, for a forensic record nothing harvests.

Harness nodes also run in Network mode, and `#[freenet_test]` reads that log back to build its failure report:

```
generate_failure_report -> aggregate_events -> event_log_path
```

Nothing in `crates/core/src/test_utils.rs` or `crates/freenet-macros/` set `enable_event_log` — the identifier appeared only in `config.rs` and `node.rs`. So harness nodes silently stopped writing the log, and **the event section of every failing integration test's report is now empty**.

`TestHarness::event_log_path` (`crates/core/src/test_utils.rs:1523-1526`) still carried the comment *"Nodes run in Network mode, so they create _EVENT_LOG not _EVENT_LOG_LOCAL"*. Network mode became precisely the reason it wouldn't.

## Why it matters more than the size of the diff suggests

The failure mode is the quiet kind. Nothing errors; the report just stops saying anything. An empty event summary reads like "nothing interesting happened" rather than "this diagnostic is switched off" — so the next person debugging a flaky or failing integration test loses the evidence *without being told they lost it*. Same shape as a vacuous test whose premise is unreachable.

`AggregateEventsMode::OnFailure` is the default, so this affects every `#[freenet_test]` unless a test opts out.

## Scope

Diagnostics only — no production behavior change, and no impact on `telemetry.freenet.org` (a separate in-memory `TelemetryReporter` sink, unaffected by the gate). 0.2.112 ships the #4969 fix as-is; this is a test-harness gap alongside it.

## Fix

Have `#[freenet_test]` opt in explicitly at both node-config sites (gateway and peer). A temp-dir node that lives for one test pays none of the disk cost that justified making the log opt-in for real peers.

Found while reviewing #4969 against the deploy gate for 0.2.112.

[AI-assisted - Claude]

sanity:
## Problem

#4969 made the local `_EVENT_LOG` opt-in on network-mode nodes. That is right for real peers — measured on a live 0.2.111 peer it was ~61 MiB/hour of appends and 95% of every fsync the process issued, for a forensic record nothing harvests.

Harness nodes also run in Network mode, and `#[freenet_test]` reads that log back to build its failure report:

```
generate_failure_report -> aggregate_events -> event_log_path
```

Nothing in `test_utils.rs` or `crates/freenet-macros/` set `enable_event_log` — the identifier appeared only in `config.rs` and `node.rs`. So every harness node silently stopped writing the log and **the event section of every failing integration test's report went empty**. `AggregateEventsMode::OnFailure` is the default, so this affects every `#[freenet_test]`.

The failure mode is the quiet kind: nothing errors, the report just stops saying anything. An empty event summary reads like "nothing interesting happened" rather than "this diagnostic is switched off", so the next person debugging a failing integration test loses the evidence without being told they lost it.

## Approach

`#[freenet_test]` opts in explicitly at both node-config sites (gateway and peer). A temp-dir node that lives for a single test pays none of the disk cost that justified making the log opt-in for real peers, so there's no reason to inherit that default here.

Also corrects the comment on `event_log_path`, which asserted that Network mode is *why* `_EVENT_LOG` exists. Network mode had become precisely why it wouldn't.

Considered and rejected: flipping the default back for all nodes (throws away the real per-peer win), and having the harness read `_EVENT_LOG_LOCAL` instead (that's the Local-mode path; harness nodes genuinely are Network-mode, so it would just move the mismatch).

## Testing

New `every_harness_node_enables_the_event_log` in `codegen.rs`. Three deliberate properties:

- **Asserts on the generated tokens, not source text**, so reformatting can't quietly make it vacuous.
- **Counts both nodes** (`== 2`), so dropping the opt-in from only the gateway arm or only the peer arm still trips it — not just wholesale removal.
- **Carries a vacuity guard** asserting the generated output really contains `ConfigArgs`, so a refactor that renamed or emptied `generate_node_setup` fails loudly instead of passing against an empty string.

Mutation-verified: removing the opt-in from **just the peer arm** fails the test with `left: 1, right: 2`; reverted, green again. `cargo fmt` and `cargo clippy -p freenet-macros -- -D warnings` clean.

## Scope

Diagnostics only. No production behavior change, and no effect on telemetry.freenet.org (a separate in-memory `TelemetryReporter` sink, unaffected by the gate). 0.2.112 ships #4969 as-is; this closes the harness gap alongside it.

Found while reviewing #4969 against the deploy gate for 0.2.112.

Closes #4972

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/test_utils.rs
+++ b/crates/core/src/test_utils.rs
@@ -1522,7 +1522,13 @@ impl TestContext {
     /// Get the path to a node's event log.
     pub fn event_log_path(&self, node_label: &str) -> anyhow::Result<PathBuf> {
         let node = self.node(node_label)?;
-        // Nodes run in Network mode, so they create _EVENT_LOG not _EVENT_LOG_LOCAL
+        // Nodes run in Network mode, so they create _EVENT_LOG not _EVENT_LOG_LOCAL.
+        // Network mode is also the mode where the event log is OFF by default
+        // (#4968), so the harness must opt in explicitly for this path to
+        // resolve to a file with anything in it — `#[freenet_test]` does that
+        // via `enable_event_log: Some(true)`. If that opt-in is ever dropped,
+        // this returns a path to an absent file and every failure report goes
+        // silently empty rather than failing loudly. See #4972.
         Ok(node.temp_dir_path.join("_EVENT_LOG"))
     }
 
--- a/crates/core/src/test_utils.rs
+++ b/crates/core/src/test_utils.rs
@@ -1587,26 +1587,7 @@ impl TestContext {
                     Ok(events) => {
                         writeln!(&mut report, "\nTotal events: {}", events.len()).unwrap();
 
-                        // An empty aggregation is almost never "nothing
-                        // happened" — it means the evidence source is missing,
-                        // and it reads identically to a genuinely quiet run.
-                        // Say so, or the reader silently draws the wrong
-                        // conclusion (#4972). A warning rather than an error on
-                        // purpose: `record_logs` sleeps briefly before creating
-                        // segment 0, so a fast failure legitimately has none.
-                        if events.is_empty() {
-                            writeln!(
-                                &mut report,
-                                "\n  ⚠️  NO EVENTS AGGREGATED across all nodes. This is very \
-                                 likely a missing evidence source rather than a quiet run — \
-                                 check that the event log is enabled for harness nodes \
-                                 (`enable_event_log: Some(true)`, #4972), that the test \
-                                 didn't fail before any events were written, and that \
-                                 `record_logs` could open the log. Treat assert-absence \
-                                 checks in this test as unproven."
-                            )
-                            .unwrap();
-                        }
+                        Self::warn_if_no_events(&mut report, events.len());
 
                         // Group by peer_id
                         let mut by_peer: HashMap<String, Vec<_>> = HashMap::new();
@@ -1820,6 +1801,36 @@ impl TestContext {
         Ok(report_dir)
     }
 
+    /// Append a loud warning when event aggregation came back empty.
+    ///
+    /// An empty aggregation is almost never "nothing happened" — it means the
+    /// evidence source is missing, and it reads identically to a genuinely
+    /// quiet run. Say so, or the reader silently draws the wrong conclusion
+    /// (#4972).
+    ///
+    /// Shared by the failure report AND the success summary on purpose: the
+    /// failure this guards is a test that PASSES vacuously, which never reaches
+    /// the failure report.
+    ///
+    /// A warning rather than an error deliberately: `record_logs` sleeps
+    /// briefly before creating segment 0, so a very fast test legitimately has
+    /// no segments and a hard error would be flaky.
+    fn warn_if_no_events(report: &mut String, event_count: usize) {
+        use std::fmt::Write;
+        if event_count == 0 {
+            writeln!(
+                report,
+                "\n  ⚠️  NO EVENTS AGGREGATED across all nodes. This is very likely a \
+                 missing evidence source rather than a quiet run — check that the event \
+                 log is enabled for harness nodes (`enable_event_log: Some(true)`, \
+                 #4972), that the test didn't fail before any events were written, and \
+                 that `record_logs` could open the log. Treat any assert-absence check \
+                 in this test as UNPROVEN."
+            )
+            .unwrap();
+        }
+    }
+
     pub async fn generate_success_summary(&self) -> String {
         use std::fmt::Write;
 
@@ -1834,6 +1845,14 @@ impl TestContext {
                 Ok(mut events) => {
                     writeln!(&mut report, "\n📊 Event Statistics:").unwrap();
                     writeln!(&mut report, "  Total events: {}", events.len()).unwrap();
+
+                    // This is the path that matters most for #4972: the failure
+                    // being guarded is a test that PASSES vacuously, and a
+                    // passing test with `aggregate_events = "always"` comes
+                    // through here, not through the failure report. (The macro
+                    // also does not catch panics, so an `assert!` failure skips
+                    // the failure report entirely.)
+                    Self::warn_if_no_events(&mut report, events.len());
 
                     // Count by event type
                     let mut by_type: HashMap<String, usize> = HashMap::new();
--- a/crates/core/src/test_utils.rs
+++ b/crates/core/src/test_utils.rs
@@ -1519,16 +1519,27 @@ impl TestContext {
             .collect()
     }
 
-    /// Get the path to a node's event log.
+    /// Base path for a node's event log. This is a segment *prefix*, not a file:
+    /// records live in `_EVENT_LOG.NNNNNNNNNN` segments, which is what
+    /// `read_all_events` scans for. Do not `.exists()`-check the returned path.
+    ///
+    /// Nodes run in Network mode, so the base name is `_EVENT_LOG`, not
+    /// `_EVENT_LOG_LOCAL`. Network mode is also the mode where the event log is
+    /// OFF by default (#4968), so the harness must opt in explicitly for these
+    /// segments to exist at all — `#[freenet_test]` does that via
+    /// `enable_event_log: Some(true)`.
+    ///
+    /// If that opt-in is ever dropped, this does NOT start returning a missing
+    /// path: `ConfigPathsArgs::build` writes a zero-length `_EVENT_LOG` stub
+    /// unconditionally, so the path resolves to a file that exists and is
+    /// empty. That is exactly why the breakage is silent — aggregation yields
+    /// zero events and every consumer reads that as "nothing happened" rather
+    /// than "the source is switched off". (With the opt-in present the stub is
+    /// deleted by `LogFile::migrate_legacy`, so the base path is absent while
+    /// the segments beside it hold the records — the inverse of what an
+    /// existence check would suggest.) See #4972.
     pub fn event_log_path(&self, node_label: &str) -> anyhow::Result<PathBuf> {
         let node = self.node(node_label)?;
-        // Nodes run in Network mode, so they create _EVENT_LOG not _EVENT_LOG_LOCAL.
-        // Network mode is also the mode where the event log is OFF by default
-        // (#4968), so the harness must opt in explicitly for this path to
-        // resolve to a file with anything in it — `#[freenet_test]` does that
-        // via `enable_event_log: Some(true)`. If that opt-in is ever dropped,
-        // this returns a path to an absent file and every failure report goes
-        // silently empty rather than failing loudly. See #4972.
         Ok(node.temp_dir_path.join("_EVENT_LOG"))
     }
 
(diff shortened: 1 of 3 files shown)


sanity:
## Review round applied — and the impact is worse than this PR originally claimed

An independent adversarial review (a reviewer that did not write this code) surfaced four findings. All four are addressed in `a0da534a`. Rebased onto current `main` (`7ef31542`, the 0.2.112 release bump).

### The finding that changes the framing

I described this as "failure reports go empty". That understated it. There is a **real assertion** that has been passing vacuously since #4969 landed this morning.

`crates/core/tests/operations.rs`, in `test_client_disconnect_does_not_immediately_unsubscribe_with_recent_read`:

```rust
let events = aggregator.get_all_events().await?;
let unsubscribe_sent_count = events.iter().filter(...).count();      // 0 when events is empty
let unsubscribe_received_count = events.iter().filter(...).count();  // 0 when events is empty
assert!(unsubscribe_sent_count == 0 && unsubscribe_received_count == 0, ...)
```

This asserts the **absence** of an event. With the log off, `get_all_events()` returns nothing, both counts are 0, and the assertion passes unconditionally — so the #4642/#4738 demand-gated-collapse guard has been showing green while proving nothing.

Added a non-vacuity guard that fails loudly when the evidence source is empty. **If this test goes red, treat it as a genuine collapse-gate signal, not a flake introduced by this PR.**

### My own comment was wrong in both directions

The explanatory comment I added to `event_log_path` claimed the path would resolve to "an absent file". Both halves are false, and I verified each against the code:

- **Opt-in dropped:** `ConfigPathsArgs::build` writes a zero-length `_EVENT_LOG` stub *unconditionally* (`config.rs:2554-2560`), so the path resolves to a file that **exists and is empty**. That is exactly why the breakage is silent — an absent file would have been detectable.
- **Opt-in present:** `migrate_legacy` **deletes** that stub (`aof.rs:283-285`) and records go to `_EVENT_LOG.NNNNNNNNNN` segments, so the base path is a segment *prefix* that does not exist as a file.

So anyone acting on my comment by adding an `.exists()` check would have got precisely backwards behavior. Rewritten to say what the code actually does, including "do not `.exists()`-check this path".

### Also fixed

- **Expected count derived from `args.nodes.len()`** rather than hardcoded `2`. The invariant is "one opt-in per node". With a literal, someone adding a third node-config arm that the 2-node fixture never reaches keeps a green test while shipping an arm without the opt-in — the same silent blind spot as the bug being fixed. Added a companion guard asserting one `ConfigArgs` per node, so the count keeps meaning what it claims.
- **`generate_failure_report` now warns when aggregation returns zero events** across all nodes. An empty report currently reads identically to a quiet run. Deliberately a warning and not a hard error: `record_logs` sleeps briefly before creating segment 0, so a fast failure legitimately has no segments and an error would be flaky.

### Verification

- End-to-end restoration was proven empirically by the reviewer, not just by reading: `test_always_aggregate` at the prior head gives `Total events: 1`; with the two opt-in lines removed, `Total events: 0`.
- No other harness site was missed — `TestContext` is the only Network-mode AOF consumer and `codegen.rs` holds its only constructor. Other hand-rolled Network-mode `ConfigArgs` never read an event log, and the DST harness runs `OperationMode::Local` (log on by default) with an in-memory listener.
- No measurable CI cost: strictly cheaper than the pre-#4969 baseline, which had the same log on *plus* a `sync_all()` per 5-event batch. Timed runs were within noise.
- Production-confined: `freenet-macros` is a `[dev-dependencies]` entry only, so the codegen change cannot reach a production build; the `test_utils.rs` change is comments plus one report line.
- Re-mutation-tested after the rewrite: dropping the opt-in from only the peer arm still fails (`left: 1, right: 2`); reverted, green. `cargo fmt --check`, `cargo clippy -p freenet-macros --all-targets -- -D warnings`, and `cargo check -p freenet --tests` all clean.

[AI-assisted - Claude]

sanity:
## Round 2 applied — including a claim this PR made that turned out to be false

All four findings from the re-review are addressed in `dedc4a31`.

### The important one: my own comment was wrong again

I wrote that deriving the expected count from `args.nodes.len()` prevents *"a fixture that never reaches a newly-added config arm keeps passing while that arm ships without the opt-in"*. **It doesn't.** The reviewer disproved it by mutation rather than by reading: adding a third node-config arm without the opt-in leaves the token-count test green, because both the node count and both match counts stay at 2 — identical to the hardcoded literal it replaced.

I reproduced that independently. Adding an unreached `relay-`-gated arm inside `generate_node_setup`:

```
every_harness_node_enables_the_event_log ................ ok      <- blind to it
every_config_args_literal_in_the_source_sets_the_opt_in . FAILED  <- catches it
  found 3 literal(s) but 2 opt-in(s)
```

Unreached-arm coverage is a property of the *fixture*; no token count against a single fixture can establish it. So the comment is corrected to say what the derivation actually buys (fixture node-count edits track automatically, no manual literal bump), and the hole is genuinely closed by a new source scrape over `generate_node_setup` that asserts every `ConfigArgs {` literal carries the opt-in — including arms no fixture reaches.

This is the second comment in this PR asserting a protection the code didn't provide (the first was the `event_log_path` "absent file" claim). Both were caught in review, neither by me. So the new test documents its own scope boundary explicitly: it covers only `generate_node_setup`, and a `ConfigArgs` literal in a different function is invisible to it — confirmed by mutation, and recorded as an accepted bound rather than left for someone to discover.

### Also fixed

- **The NO-EVENTS warning was on the wrong path.** It lived only in `generate_failure_report`, but what this PR guards against is a test that **PASSES** vacuously — and a passing test with `aggregate_events = "always"` goes through `generate_success_summary`. Compounding it, the macro deliberately doesn't catch panics, so the failure report only runs on the `Err` path and an `assert!` skips it entirely. Hoisted into `warn_if_no_events` and called from both.
- **The `operations.rs` guard was satisfiable by startup records alone.** `Lifecycle(Startup)` is unconditional per node, so `!events.is_empty()` held with 3 events and nothing subscription-related. It now also requires a `Subscribe` event — this test's subject is the subscription lifecycle, so without one an Unsubscribe wouldn't have been captured either. Healthy runs measure ~140-190 events with `Subscribe: 5`, against the 1 required.
- **`assert!` → `ensure!`** at that guard, so it returns `Err`, routes through `generate_failure_report`, and prints the NO-EVENTS warning alongside its own message instead of bypassing the report this PR just improved. Matches the macro's own documented guidance.

### The result that matters most

The reviewer ran the formerly-vacuous test three times: **passes in 64s with 140 / 151 / 192 events** at the guard point. So the #4642/#4738 demand-gated-collapse behavior is genuinely **correct**, not merely untested — and the guard has roughly two orders of magnitude of margin over zero, so it is not a flake risk. That was the open question from round 1 and it is now closed with measurements rather than argument.

`cargo fmt`, `cargo check -p freenet --tests`, and the macro suite (6/6) are clean. Full CI running.

[AI-assisted - Claude]

sanity:
## Round 3 applied — two more holes in my own scrape, both found by mutation

Applied in `aaffab27`. Both findings were in the guard I added last round, and both were *demonstrated* rather than argued — which is the only reason they were caught, since both looked fine on reading.

### 1. A `//` comment satisfied the opt-in count

A third `ConfigArgs` literal with **no** `enable_event_log` field, whose only mention of it lived in a comment, left both tests green (`literals=3, opt_ins=3`). That is precisely the hole the scrape exists to close.

Not hypothetical: `crates/core/src/test_utils.rs` already carries the prose *"via `enable_event_log: Some(true)`"*, so a developer adding an arm and bringing the surrounding comment along reproduces it exactly.

Fixed by stripping `//` lines before counting. That also closes the inverse cry-wolf direction, where a comment containing `ConfigArgs {` would inflate the literal count on correct code.

### 2. The vacuity guard didn't prove what it claimed

`body.contains("NetworkArgs")` was meant to prove the slice really was `generate_node_setup`. But `NetworkArgs` appears in **both** config arms, so a slice truncated after the first still passed its own guard with `literals=1, opt_ins=1` — green, while the peer arm sat entirely outside the scrape. And truncation was reachable: `split("mod tests")` cuts at the **first** occurrence, so that string appearing in any production comment would truncate mid-function.

Fixed by widening the scrape from `generate_node_setup` to the whole pre-`mod tests` region, as the reviewer recommended and verified. That deletes all three fragile mechanisms at once — the function-name needle, the `\nfn ` end boundary, and the truncation-blind vacuity guard — and it also covers a `ConfigArgs` added in a *different* function.

That last part matters: it's the gap my own bad mutation exposed in round 2, which I then documented as an "accepted bound". It wasn't one. It's now closed, and the caveat is deleted rather than maintained.

The surviving vacuity guard (`literals > 0`) is the one that actually works — a truncated or mis-cut slice yields zero literals and fails loudly.

### Verified in both directions

| mutation | token test | source scrape |
|---|---|---|
| unreached arm, opt-in only in a comment | ok | **FAILED** (3 literals, 2 opt-ins) |
| `ConfigArgs` in a different function | ok | **FAILED** (3 literals, 2 opt-ins) |
| reverted | ok | ok (6/6) |

The token test being blind in both rows is the point — it structurally cannot see arms a two-node fixture never generates, which is why the scrape exists.

### Also corrected: my justification for the Subscribe requirement

I wrote that it holds because "this test's subject is the subscription lifecycle". The real reason is stronger and I verified it: `UnsubscribeSent` and `UnsubscribeReceived` are variants of `SubscribeEvent`, wrapped as `EventKind::Subscribe(..)` (`crates/core/src/tracing/event_kind.rs:503,514`). So the guard demands an event from the **exact enum arm** the two counted assertions look inside — if the pipeline that would carry an `UnsubscribeSent` is dead, `EventKind::Subscribe` is absent too. The margin is structural rather than incidental as well: the test asserts `subscribed == true` before reaching the dwell loop, so a successful SUBSCRIBE has definitely emitted `SubscribeSuccess`. Comment updated to say that.

### Confirmed correct, no change needed

`warn_if_no_events` at both call sites (takes `usize`, so no borrow conflict with the later `events.iter()`); `ensure!` expands to an early `return Err`, giving identical pass/fail semantics to the previous `assert!` while routing through the failure report; and rustfmt cannot wrap either needle at their indentation levels.

`cargo fmt`, `cargo check -p freenet --tests`, macro suite 6/6 clean. Full CI running on `aaffab27`.

[AI-assisted - Claude]

sanity:
## Round 4: the source scrape is deleted

I added that guard in round 2, patched it in round 3, and it is gone in `bccaaa94`. Round 4 defeated it three more ways, and I reproduced the worst one before acting rather than deleting on report.

### Why

Five distinct mutations defeat it, each visible only by mutation:

| # | mutation | token test | scrape |
|---|---|---|---|
| A | opt-in named in a **trailing** comment on a code line | ok | ok ← blind |
| B | opt-in named in a `/* block comment */` | ok | ok ← blind |
| C | `ConfigArgs` literal in a different function | ok | *(fixed round 3)* |
| D | a production comment containing the words **`mod tests`** | ok | ok ← blind |
| E | opt-in named in a `///` doc comment | ok | ok ← blind even in a lexer-based version |

**D settles it.** `split("mod tests")` cuts at the first occurrence, so one innocuous comment truncates the scraped region before the peer arm — and both tests go green on precisely the bug the scrape exists to catch. I ran it myself:

```
// Kept in sync with the fixture in mod tests below.
   + an unreached arm with no opt-in

every_config_args_literal_in_the_source_sets_the_opt_in ... ok   ← should have failed
every_harness_node_enables_the_event_log ................ ok
```

The `//`-line strip from round 3 closes shape A-adjacent cases and nothing more. The root problem isn't the patch, it's the approach: **text matching cannot decide whether a token is a field or prose.** A lexer version (`proc_macro2`, which discards comments) kills A, B and D — and still loses to E, because `///` becomes `#[doc = "..."]` and the phrase survives inside a string literal.

### Why deleting is consistent with the rest of this PR, not a retreat from it

A guard that looks like protection but isn't is worse than no guard, because it stops the next person from adding real protection. That is exactly the failure mode of the two comments in this PR that claimed protection the code didn't provide — the `event_log_path` "absent file" claim and the `args.nodes.len()` unreached-arm claim. Keeping a guard with the same property while fixing those would have been incoherent.

It also never caught a real bug, needed three rounds of fixes, and guards a hypothetical (a future unreached arm) that it did not reliably cover anyway.

### What still covers the actual damage — unchanged

- **the token test** — both reachable config arms, mutation-verified to fail on a single-arm drop
- **`warn_if_no_events` in both report paths** — any runtime empty aggregation, whatever the cause
- **the `operations.rs` Subscribe guard** — the vacuous-assertion class, which was the real harm from #4972

The token test's docs now state plainly that unreached arms are **not** covered, tell you to add a fixture node that reaches any new arm, and record the two ways to enforce it for real if it ever matters: parse instead of match (`syn::parse_file` walking for an `ExprStruct` whose path ends in `ConfigArgs` — fields and doc attributes are structurally distinct there), or remove the duplication so both arms build their `ConfigArgs` through one shared helper and there is only one place to forget.

I did not take the shared-helper refactor in this PR: it changes generated code structure and wants its own verification, which is scope creep on a test-diagnostics fix.

Macro suite 5/5, `cargo fmt`, `cargo clippy -p freenet-macros --all-targets -- -D warnings` clean. CI running on `bccaaa94`.

[AI-assisted - Claude]

Issue #4980, opened 2026-07-27
Telemetry collector load at 10x fleet size: ~2.9 TB/day ceiling, 73% of bytes in six per-event streams


sanity:
## Summary

Measured the central telemetry collector's current load to answer "is this telemetry change safe if the network grows 10x". The answer for the two PRs in question was yes, but the measurement turned up a structural exposure worth recording separately: **at 17,000 peers the collector's hard ceiling is ~2.9 TB/day, and today's volume is dominated by six per-event streams that scale linearly with peer count.**

This is not urgent and nothing is currently broken. Filing it so the number exists before someone needs it.

## Measurement

200 MB slice of a live rotation of `/mnt/media/freenet-telemetry/logs.jsonl` on nova, 2026-07-27, mostly `service.version == 0.2.111`. 280,944 records, 841 distinct peers in the slice.

| | today (~1,700 peers) |
|---|---|
| records per MB | 1,405 (mean 712 B/record) |
| rotation | 5.24 GB every ~80-95 min |
| records/rotation | ~7.37M |
| **rate** | **~1,445 records/s, ~1.03 MB/s** |
| **daily** | **~88.8 GB/day** |

Straight 10x on per-peer rate: **~888 GB/day** at 17,000 peers.

## The composition is the part that matters

Six per-event streams are **72.9% of all collector bytes**:

| event | % of bytes | mean size |
|---|---|---|
| `update_broadcast_received` | 18.4% | 684 B |
| `connect_request_received` | 16.1% | 729 B |
| `transfer_completed` | 13.5% | 743 B |
| `transfer_started` | 10.0% | 503 B |
| `connect_connected` | 8.4% | 903 B |
| `connect_rejected` | 6.5% | 474 B |

Every one is per-event and scales with traffic x peers. By contrast the fixed-cadence rollups are cheap: `broadcast_payload_mix` is 1.5% of bytes, `outbound_message_mix` 0.75%, `shadow_*` under 1.5% combined — and each is flat per node per window regardless of how much traffic that node carries.

Two of the six look like straightforward wins if this ever needs attention:

- **`connect_request_received` + `connect_rejected` = 22.6% of all collector bytes.** In the slice, `connect_rejected` (26,921) is 62% of `connect_request_received` (43,441). That is a lot of bytes spent recording relay-forwarding churn one event at a time. A per-node rollup of connect outcomes would collapse almost all of it.
[...]

Issue #4965, opened 2026-07-27
perf: InterestSync Summaries exchange is 53.7% of all outbound bytes (26.5 KB mean per heartbeat message)


sanity:
## MEASURED: InterestSync is 53.7% of all outbound bytes — more than actual contract updates

First field data from the 0.2.110 `outbound_message_mix` census (626 peers, 3,265 node-minutes). This is the ~73% of outbound bytes that broadcast payload never accounted for, and it has a single dominant owner.

| message kind | share of outbound bytes | bytes | msgs | mean |
|---|---|---|---|---|
| **`interest_sync`** | **53.7%** | 4,766 MB | 180,018 | **26.5 KB/msg** |
| `update` | 41.5% | 3,685 MB | 99,358 | 37.1 KB |
| `neighbor_hosting` | 4.4% | 389 MB | 109,623 | 3.5 KB |
| `connect` | 0.3% | 25 MB | 214,610 | 0.1 KB |
| `get` / `subscribe` / `put` / other | <0.1% each | — | — | — |

**The anti-entropy heartbeat costs more than the payload it exists to keep fresh.** A `Summaries` message averages **26.5 KB**, which is not a heartbeat-shaped number. The wire type explains it — `Summaries { entries: Vec<SummaryEntry> }` where each entry is `(u32 hash, Option<Vec<u8>> summary_bytes)`, i.e. **full summary bytes for every shared contract, to every co-host neighbor, every cycle**. With River room summaries measured at ~33 KB (see #4956 / bandwidth notes), a handful of shared contracts per neighbor lands exactly here.

Cross-checked per version, so this isn't one bad build: 0.2.110 = 26.4 KB/msg over 171,612 msgs; 0.2.111 = 20.8 KB/msg over 17,786 msgs.

### It also appears not to be doing its job

The companion 0.2.111 measurement (#4961) says the tracked-but-summaryless broadcast arm — 27% of broadcast bytes at a 327 KB mean — is **99.8% `never_populated`**: the peer-interest entry was created with no summary and never had one written. Seeding it is exactly what the `Summaries` exchange is for.

So the current reading is: we spend 4.6 GB of fleet bandwidth on an exchange whose stated purpose is seeding peer summaries, while 99.8% of the peers that force full-state broadcasts have never been seeded. Either those are two disjoint peer populations, or the exchange is not landing where it matters. Both are worth confirming before touching the wire format — I have not established the mechanism yet, only the two numbers.

### Not proposing a fix in this comment

[...]

PR #4982, opened 2026-07-27
feat(telemetry): split the interest_sync census arm and measure summary-identical rate

sanity:
## Problem

`interest_sync` is 53-75% of all outbound bytes fleet-wide (~26 KB mean per heartbeat, ~3x actual contract updates) — the biggest known remaining bandwidth lever, larger than the 2.6-2.7x disk-write cut that 0.2.112 shipped. The obvious remedy is hash-first: exchange summary digests and ship `summary_bytes` only on mismatch.

Two things block deciding that, and both are measurement gaps rather than design questions.

**We can't attribute the bytes.** The census lumps the whole InterestSync family into one arm (`OutboundKind::InterestSync`), but the legs differ by orders of magnitude and have different remedies:

- `Interests` / `ChangeInterests` — 4 bytes per interest, just `u32` contract-id hashes.
- `Summaries` — a full `StateSummary` per shared contract, to every connected peer, every cycle.
- `ResyncRequest` / `ResyncResponse` — full contract **state**, i.e. heal traffic, not heartbeat.

Attributing the 53-75% to `Summaries` is currently inference from payload shapes, not measurement. And folding resync into the reply arm would attribute heal traffic to the heartbeat, overstating exactly the thing being measured.

**We can't tell whether hash-first would help at all.** It saves bytes precisely on the fraction where the two sides' summaries are byte-identical. That fraction is unmeasured, and there is real reason to think it might be low: `node.rs` carries a load-bearing comment (#4857, the 2.56M-`summarize_contract_state` storm) explaining that summaries which serialize non-deterministically produce different bytes for the *same* logical state. For any such contract a digest mismatches every time, ships the bytes anyway, and adds a round trip — strictly worse than today.

## Solution

Both measurements, as **fields on the existing per-minute `outbound_message_mix` rollup**.

Not new event streams, deliberately: there is a hard `MAX_EVENTS_PER_SECOND` per node, five aligned rollups already contend for the shadow sub-budget, and the cap leaves room for exactly one more stream (`tracing/telemetry.rs`). A new operational event would compete for a fixed per-second budget on precisely the busiest peers whose interest_sync behavior we want to read. A field on an existing rollup costs zero additional records. See #4980.

1. **Split the census arm** into `interest_sync_interests`, `interest_sync_summaries`, `interest_sync_resync`. The classify match stays exhaustive, so a new `InterestMessage` variant fails to compile rather than silently joining whichever arm it resembles.

2. **Count summary comparisons** as identical vs differing at the one place both sides' summaries exist (`node.rs::handle_interest_sync_message`), plus bounded per-contract attribution of the differing side. That last part is what makes a low rate readable: "the design is wrong" and "these three contracts serialize non-deterministically" produce the same aggregate ratio and have completely different fixes.

The per-contract map is bounded at 256 (mirroring `broadcast_payload_mix`) because the key is contract-controlled; an already-tracked contract keeps accruing past the cap, so the top-N ranks by who diverges most rather than by who arrived first.

## Scope

Observation only. Nothing reads these counters to make a decision, and there is no wire-format change — the wire change is step 2, gated on what this measures.

## Testing

- `interest_sync_legs_classify_into_separate_arms` — each of the five `InterestMessage` variants lands in its intended arm.
- `summary_comparisons_split_identical_from_differing` — the falsifier itself; an inverted boolean would still produce a plausible-looking ratio and send the design decision the wrong way with no visible symptom.
- `differing_attribution_is_bounded_but_keeps_accruing_known_contracts` — both halves: the bound, and that a known contract still accrues past it.

Mutation-verified (committed first): inverting the identical/differing branch fails 2 tests; lumping `Summaries` back into the request arm fails 1; removing the map bound fails 1.

`cargo fmt`, `cargo clippy --locked -- -D warnings` (the CI invocation) and the lib suite are clean.

## Follow-up

Step 2 — `Option<summary_digest>` on `SummaryEntry`, shipping bytes only on mismatch — is gated on this coming back with a high identical rate. Note for whoever builds it: `SummaryEntry.hash` is **not** a summary digest, it is FNV-1a of the contract instance id (`ring/interest.rs`), so hash-first needs a genuinely new field. And filtering sends against our cached `PeerInterest.summary` would be unsafe — it suppresses the heal in exactly the case the anti-entropy exists for, because every failure it heals is "our belief about a peer is wrong". A digest exchange stays safe because both sides compute it from actual state.

Refs #4965

[AI-assisted - Claude]

---

### The old `interest_sync_*` fields are RETIRED, not renamed

`interest_sync_msgs`, `interest_sync_bytes` and `interest_sync_max_bytes` **stop being emitted**. They are replaced by three separate stems, not aliased:

| retired | replaced by |
|---|---|
| `interest_sync_msgs` / `_bytes` / `_max_bytes` | `interest_sync_interests_*` (the `Interests` / `ChangeInterests` request leg) |
| | `interest_sync_summaries_*` (the `Summaries` reply — the suspected cost) |
| | `interest_sync_resync_*` (`ResyncRequest` / `ResyncResponse`, full state) |

**This matters for anyone re-running the #4965 baseline across the deploy boundary.** A saved query against `interest_sync_bytes` returns nothing after the upgrade, and nothing looks exactly like a fixed problem — the traffic appears to vanish when it has only been re-labelled. Sum the three new stems to compare against a pre-upgrade figure.

No in-repo consumer reads the old names (grep clean, and the live dashboard does not reference `interest_sync` at all), and the rollup itself is only about a day old from #4957, so the exposure is ad-hoc queries rather than anything checked in.

### A LOW identical rate probably means specific contracts, not a dead design

The obvious reading of a low `summary_entries_identical` ratio is "hash-first would not help, the summaries genuinely differ". Before concluding that, check `summary_differing_contracts`.

River's contract summaries have already been made deterministic — every summary type uses `BTreeSet` / `BTreeMap` rather than `HashSet` / `HashMap`: `MessagesSummary.message_ids`, `SecretsSummary.version_ids`, `DirectMessagesSummary.message_signatures`, plus `member_info`, `member` and `ban`. Each carries an explicit comment naming freenet-core#4857 and there is a dedicated `contract-summary-determinism.md` rule. That fix exists because `HashMap` iteration order made two identical logical states summarize to different bytes, causing spurious anti-entropy heals.

So the expectation is a **high** identical rate for the contracts that dominate our traffic, and a low rate only for third-party contracts that have not had the same fix.

- **Low aggregate, and `summary_differing_contracts` names a handful of non-River keys** → the protocol design is fine and hash-first is still worth building; the fix is upstream, in those contracts' summary types.
- **Low aggregate, and the differing set is broad and includes River keys** → that would falsify the determinism assumption and is the case where hash-first genuinely buys little.
- **`differing_attribution_dropped` non-zero** → the named list is a biased sample (slots are first-come, and entries arrive in contract-id order), so rank from the aggregate, not from the list.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2741,9 +2741,12 @@ async fn handle_interest_sync_message(
                 // `MAX_STALENESS_PROBES_PER_SUMMARIES` / `plan_staleness_probe`.
                 let mut staleness_probes_used = 0usize;
                 // Contracts already counted by the #4965 summary-comparison
-                // measurement in THIS message. Bounded by the message's entry
-                // count, which the transport already bounds; it holds contract
-                // ids only, no state.
+                // measurement in THIS message. Scoped to one message, so it
+                // cannot accumulate across a connection. Its size is bounded by
+                // the number of DISTINCT locally-known contracts the entries
+                // resolve to — `lookup_by_hash` only yields contracts this node
+                // already tracks, so a peer cannot grow it with unknown ids —
+                // and it holds contract ids only, no state.
                 let mut compared_contracts: std::collections::HashSet<
                     freenet_stdlib::prelude::ContractKey,
                 > = std::collections::HashSet::new();
@@ -4149,11 +4152,23 @@ mod tests {
              measurement; without it the identical/differing rollup is \
              silently always zero"
         );
+        // Structural, not presence: assert the call sits INSIDE the dedup
+        // guard's block. Two independent `contains` checks would pass a
+        // regression that kept both identifiers but moved the call out from
+        // behind the `if` — which is precisely the mutation that silently
+        // re-opens the skew. Slice from the guard to the end of its block.
+        let guard_off = summaries_arm
+            .find(concat!("compared_", "contracts.insert("))
+            .expect("the per-message dedup guard must exist in the Summaries arm");
+        let guard_block_end = summaries_arm[guard_off..]
+            .find("\n                                }")
+            .expect("dedup guard block end not found");
+        let guard_block = &summaries_arm[guard_off..guard_off + guard_block_end];
         assert!(
-            summaries_arm.contains(concat!("compared_", "contracts")),
-            "the measurement call must stay behind the per-message dedup set, \
+            guard_block.contains(concat!("record_summary", "_comparison")),
+            "the measurement call must sit INSIDE the per-message dedup guard, \
              or a peer repeating a hash can inflate the ratio that gates the \
-             wire-format decision"
+             wire-format decision. Guard block was:\n{guard_block}"
         );
 
         assert!(
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2740,6 +2740,13 @@ async fn handle_interest_sync_message(
                 // sending novel summary bytes for every hosted contract. See
                 // `MAX_STALENESS_PROBES_PER_SUMMARIES` / `plan_staleness_probe`.
                 let mut staleness_probes_used = 0usize;
+                // Contracts already counted by the #4965 summary-comparison
+                // measurement in THIS message. Bounded by the message's entry
+                // count, which the transport already bounds; it holds contract
+                // ids only, no state.
+                let mut compared_contracts: std::collections::HashSet<
+                    freenet_stdlib::prelude::ContractKey,
+                > = std::collections::HashSet::new();
                 for entry in entries {
                     for contract in op_manager.interest_manager.lookup_by_hash(entry.hash) {
                         if !op_manager.interest_manager.has_local_interest(&contract) {
@@ -2784,9 +2791,22 @@ async fn handle_interest_sync_message(
                                 // at the send site because only the receiver can
                                 // compare, and only in this arm because a `None` on
                                 // either side is not a comparison at all.
-                                op_manager
-                                    .outbound_mix
-                                    .record_summary_comparison(contract.id(), identical);
+                                //
+                                // Counted once per contract per message: `entries`
+                                // is peer-supplied and may repeat a hash, and
+                                // without this a peer could inflate either bucket
+                                // at will — corrupting the very ratio that decides
+                                // whether the wire change is worth building. The
+                                // dedup guards ONLY the measurement; the staleness
+                                // logic below still runs per entry exactly as
+                                // before, so this changes no behavior.
+                                if compared_contracts.insert(contract) {
+                                    op_manager.outbound_mix.record_summary_comparison(
+                                        contract.id(),
+                                        ours.as_ref(),
+                                        theirs.as_ref(),
+                                    );
+                                }
                                 let delta_verdict = if identical {
                                     // Byte-identical => converged; skip the probe.
                                     None
@@ -4116,6 +4136,25 @@ mod tests {
             .find("InterestMessage::ChangeInterests")
             .expect("ChangeInterests arm not found");
         let summaries_arm = &src[handler_start + summaries_off..handler_start + change_off];
+
+        // #4965: the measurement call must stay wired into this arm. It is
+        // pure observation, so deleting it breaks no test and no behavior —
+        // the rollup would simply report zero comparisons forever, which reads
+        // as "nothing to save here" and would retire the hash-first redesign
+        // for the wrong reason. Needle split so this assertion's own source
+        // cannot satisfy the scrape.
+        assert!(
+            summaries_arm.contains(concat!("record_summary", "_comparison")),
+            "the Summaries arm must record the #4965 summary-comparison \
+             measurement; without it the identical/differing rollup is \
+             silently always zero"
+        );
+        assert!(
+            summaries_arm.contains(concat!("compared_", "contracts")),
+            "the measurement call must stay behind the per-message dedup set, \
+             or a peer repeating a hash can inflate the ratio that gates the \
+             wire-format decision"
+        );
 
         assert!(
             summaries_arm.contains("peer_summary_has_pending_state"),
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2775,7 +2775,19 @@ async fn handle_interest_sync_message(
                         // `summary_indicates_stale_peer`.
                         let is_stale = match (our_summary.as_ref(), their_summary.as_ref()) {
                             (Some(ours), Some(theirs)) => {
-                                let delta_verdict = if ours.as_ref() == theirs.as_ref() {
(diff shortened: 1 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #4982

### Summary
- **PR Title:** feat(telemetry): split the interest_sync census arm and measure summary-identical rate
- **Type:** feat (telemetry / observation-only)
- **Review tier:** **Full** — the change adds a `parking_lot::Mutex` acquisition on an async receive path, which is a concurrency surface; when torn, take the heavier tier.
- **Reviewers run:** `code-first`, `testing`, `skeptical`, `big-picture` (round 1 on `6487955f`); `skeptical`, `testing` re-review (round 2 on `330d4d42`).
- **External model:** Codex is quota-blocked, so per `multi-model-review.md` this used the diverse-Claude-lens substitution — four distinct blind lenses rather than one extra reviewer, since a single extra Claude shares the author model's blind spots.

---

### Round 1 findings (all addressed in `330d4d42`)

| # | Finding | Severity | Location | Resolution |
|---|---|---|---|---|
| 1 | The doc comment claimed the cap kept "the top offenders accurate". **It does not** — slots are first-come and arrival order is not random, because `get_matching_contracts` sorts by contract id ascending (`ring/interest.rs:1533`). Under sustained cap pressure the same low-id contracts hold the slots every window and a higher-diverging high-id contract stays invisible, in exactly the storm the attribution exists to diagnose. | **Must Fix** (false claim) | `outbound_message_mix.rs` | Comment now states the bias plainly and says to read the aggregate |
| 2 | A capped window was indistinguishable from a complete one. The sibling `broadcast_payload_mix` reports `attribution_dropped_*` (`:102-103,236,244`) — this module claimed to mirror it and omitted exactly that. | Should Fix | `outbound_message_mix.rs` | Added `differing_attribution_dropped` + test |
| 3 | `record_summary_comparison` took a pre-computed `identical: bool` from an untested call site — the one bit the whole measurement rests on, where an inverted operand compiles and passes everything. | Should Fix | `node.rs` | Now takes both summaries and compares internally: the failure is **unrepresentable**, not merely covered |
| 4 | No per-message cap. `entries` is peer-supplied and may repeat a hash; unlike the WASM probe (`MAX_STALENESS_PROBES_PER_SUMMARIES`) nothing rationed the measurement, so a peer could skew the ratio gating the wire decision. | Should Fix | `node.rs` | Deduped per contract per message; guards only the measurement, staleness logic unchanged |
| 5 | The rollup JSON body had no test seam — inlined and returning `()`, unlike `broadcast_payload_mix::payload_mix_json` (`:549`) with 8 test call sites. An inverted sort comparator would report the LEAST-diverging contracts as "top offenders" and ship silently. | Should Fix | `outbound_message_mix.rs` | Extracted `outbound_mix_json`; sort + truncation now tested |

### Round 2 findings (addressed in `579883cc`)

| # | Finding | Severity | Resolution |
|---|---|---|---|
| 6 | `summary_entries_identical` / `summary_entries_differing` — the two headline numbers the PR exists to produce — were asserted only on the `Window` struct, never through `outbound_mix_json`. A swapped key would emit an exactly inverted ratio. | Should Fix | Round-trip test with distinguishable counts (3 vs 5) |
| 7 | The extended source pin checked that `record_summary_comparison` and `compared_contracts` each appeared *somewhere* in the arm — it would pass a regression that kept both identifiers but moved the call outside the dedup guard. | Should Fix | See below |
| 8 | Comment overstated that the transport bounds `Summaries.entries.len()`; no such cap was found. | Consider | Reworded to the bound that actually holds (distinct locally-known contracts, since `lookup_by_hash` only yields tracked ids) |

**On #7, worth reading:** I first replaced the presence check with a "structural" pin that sliced the guard's block and asserted the call was inside it. **Mutation testing showed it green** when the call was moved out from behind the `if` — the sliced block ran on to a later brace that still contained the call. I deleted it rather than patching it. Text matching cannot decide nesting, and a guard that looks like protection but isn't is worse than none, because it stops the next person adding real protection. The dedup **bypass** therefore has no coverage; that is stated in the code rather than papered over. The presence pin on the measurement call itself is kept and is mutation-verified effective — deleting the call fails it, which is the realistic regression here (pure observation, so deletion breaks nothing else and the rollup silently reports zero forever).

---

### Testing Assessment

| Unit | ✅ | 6 tests: leg classification, identical/differing split, cap + keep-accruing, drop reporting, sort/truncation, headline round-trip |
| Integration | ⚠️ | No test drives `handle_interest_sync_message` with a real `Summaries` message; the dedup bypass is the specific uncovered case |
| Simulation | N/A | Observation-only, no behavioral surface |

Mutation-verified across both rounds: inverting identical/differing fails 2 tests; lumping `Summaries` back into the request arm fails 1; removing the map bound fails 1; deleting the measurement call fails the pin.

---

### Big Picture

**Goal alignment:** yes — genuinely scoped to measurement. The `node.rs` change adds one non-branching call; `delta_verdict` / `is_stale` control flow is byte-for-byte unchanged. No wire-format change.

**Telemetry field rename:** `interest_sync_{msgs,bytes,max_bytes}` become three stems. This is a breaking rename, not additive. No in-repo consumer and no dashboard reference was found (`grep` clean; `/var/www/freenet-dashboard` does not read `interest_sync`), and the rollup itself is ~1 day old (#4957), so the practical risk is a saved ad-hoc query going quietly empty rather than erroring.

**Docs/rules:** `.claude/rules/ring.md`, `channel-safety.md` and `docs/architecture/event-dispatch.md` reference the handler and message semantics, none of which changed. Not stale.

---

### Verified, not findings

- **Not a blocking-in-async hazard.** `record_summary_comparison` is a sync `fn`, never holds the mutex across an `.await`, critical section is O(1), multi-threaded runtime.
- **Arm indices/arrays consistent.** `ALL` (10), `index()`, and the `[u64; 10]` arrays agree 1:1; `classify` is exhaustive over all 9 `NetMessageV1` and all 5 `InterestMessage` variants with no catch-all.
- **Dedup scoping correct.** `compared_contracts` is declared inside the `Summaries` arm, so it is per-message and cannot accumulate across a connection. It keys on the resolved `ContractKey`, which also closes a hash-collision variant.
- **Cap enforcement is race-free** — the `len` check and the insert are under the same lock.

---

### Verdict
**State:** Ready to Merge (pending CI green on the current HEAD)
**HEAD SHA reviewed:** `579883cc`

Residual gap, accepted and documented in code: the per-message dedup bypass has no executable coverage. Closing it needs a handler-level test with hosted state and an executor, which is disproportionate for observation-only telemetry — but it is a real gap, not a solved one.

[AI-assisted - Claude]

PR #4997, opened 2026-07-27
fix: evict oldest pairs instead of starving new ones in the UPDATE rate limiter

### LINKED ISSUE #4981: UPDATE rate limiter permanently starves new (sender, contract) pairs once 16,384 slots are held by busy peers — silent drops in release builds
## Symptom that led here

A user on 0.2.111 reported a capacity-reject counter climbing **22K → 36K in a few minutes** (~50-100/sec) on 2026-07-27 ~04:23 CDT.

The obvious suspect — the contract ban list's `capacity_rejected_total` — is ruled out. Governance defaults to `GovernanceMode::Off` (`crates/core/src/contract/governance.rs:434`, unchanged at the v0.2.111 tag); with Off the reaper returns zero decisions (`governance.rs:685`), and `apply_ban_decisions` is the only production caller of `ban()` (`ring.rs:1378`). `BanReason::Operator` has no production construction site. The rate arithmetic rules it out independently: the reaper ticks every 60s (`ring.rs:172`) and a contract already `Banned` cannot re-emit (`governance.rs:882`).

The dashboard renders a **"Capacity-dropped"** tile beside "UPDATEs relayed" and "Rate-limited" (`crates/core/src/server/home_page/cards.rs:114-116`), fed by an **identically named counter on a different struct**: `UpdateRateLimiter::capacity_rejected_total` (`crates/core/src/ring/update_rate_limit.rs:276`). That one sits on the UPDATE hot path, where 50-100/sec is unremarkable.

## The bug

The limiter keys on `(sender_addr, contract_instance_id)` with `MAX_TRACKED_PAIRS = 16_384` (`update_rate_limit.rs:127`). At capacity, a **new** pair is rejected and **never inserted** (`update_rate_limit.rs:270-277`), and the caller drops the UPDATE (`node.rs:1489-1499`). Because the pair was never inserted it remains "new" forever, so every subsequent UPDATE from it is dropped and counted again.

**The starvation is structural, and the module's own doc comment is where it hides.** `update_rate_limit.rs:125-126` reasons:

> *"only NEW pairs hit the cap. Combined with the TTL sweep this gives smooth recovery as idle pairs roll off."*

That holds only for idle pairs. An **accepted** UPDATE refreshes the entry timestamp (`update_rate_limit.rs:263`, `*entry.get_mut() = now`), so an actively-used pair can never age out under `CLEANUP_AGE` (5 min, swept on the 60s tick at `ring.rs:1323`). The 16,384 slots are therefore held by whoever got in first *and stays busy*, and newly-arriving pairs are locked out indefinitely with no recovery path.

This is precisely the pattern AGENTS.md warns about:

> *"Any condition that exempts an entry from garbage collection MUST either: 1. Expire via TTL, OR 2. Be overridden by an absolute age threshold."*

There is no absolute age cap here — only a refreshable TTL, which a busy pair refreshes forever.

## Why it is invisible

The drop logs at `tracing::debug!` (`node.rs:1490`), and crates/core builds with `release_max_level_info` (`crates/core/Cargo.toml:117`), so it is **compiled out of release binaries**. On a production node the only evidence that legitimate relayed UPDATEs are being discarded is a dashboard tile. That is why this surfaced as a user asking about a counter rather than as an error anyone could grep for.

## Saturation may be entirely benign

Tracked pairs = distinct sender addresses × distinct contracts relayed from them. **50 peers × ~330 contracts ≈ 16,500** — over the cap with no attacker involved. The module doc justifies the bound purely as an anti-attacker measure (`update_rate_limit.rs:116-125`) and never reckons with a legitimate node simply outgrowing 16,384 pairs. Given the fleet is growing, this gets more likely, not less.

## Severity

**Silent data loss on the UPDATE path**, not merely a noisy counter. Legitimate relayed UPDATEs are dropped, invisibly in release builds, with no path for a locked-out pair to recover while the map stays hot.

## Options (deliberately not a one-liner — this is a design choice)

- **(a) Raise the cap.** Cheapest; treats the symptom. The bound exists for a real reason, and a bigger number still starves eventually.
- **(b) Fail open at capacity** — admit the UPDATE untracked instead of dropping it. Preserves the memory bound and stops the data loss, but an attacker churning fresh contract ids then bypasses rate limiting entirely. Trades an availability bug for a security one.
- **(c) LRU / oldest-entry eviction on admission** — preferred. Removes the starvation, keeps the bound, needs no cap change, matches the ban list's existing `evict_one_operator` precedent and the project's general preference for demand-ordered eviction over "reject once full", and directly closes the AGENTS.md permanently-refreshable-exemption violation.

Whatever the fix, the drop should be visible in release — an `info!`/`warn!` (rate-limited) or a counter surfaced somewhere greppable, so this class stops being invisible.

## Confidence

Mechanism: **high**, established by code inspection at the v0.2.111 tag and at HEAD; I have not observed a live node. Attribution of the user's specific number to this counter: **moderate** — governance being Off makes the ban list essentially impossible, but which dashboard tile they read is inferred from a paraphrase.

Cheapest things that would settle attribution:
1. **Ask which card.** If it is the ban-list card (*"N bans rejected — list at capacity"*, `cards.rs:1409`), that is a separate and more urgent finding: it would mean governance was flipped on, which matters given #4338's calibration history.
2. **The neighbouring tiles.** A high "UPDATEs relayed" alongside it confirms real traffic being partially starved; near-zero would point elsewhere.
3. **Peers × hosted contracts on that node** — near or above 16,384 means benign saturation and a sizing bug; far below means genuine churn (the #4861 storm class, or an attacker).
4. `freenet service report` carries all three counters in its network_status snapshot.

**Nothing since 0.2.111 changes this.** `update_rate_limit.rs`, `contract_ban_list.rs` and `governance.rs` are byte-identical between the tag and HEAD; the only `node.rs` change is the 0.2.112 event-log opt-in, which does not touch the limiter.

[AI-assisted - Claude]

sanity:
## Problem

At `MAX_TRACKED_PAIRS` (16,384) the per-`(sender, contract)` UPDATE rate limiter refused a new pair and **never inserted it**, so the pair stayed "new" forever and every subsequent UPDATE from it was dropped and re-counted.

The module's own doc comment is where the reasoning went wrong:

> *"only NEW pairs hit the cap. Combined with the TTL sweep this gives smooth recovery as idle pairs roll off."*

That holds only for **idle** pairs. An accepted UPDATE restamps its entry (`*entry.get_mut() = now`), so a busy pair never ages out under `CLEANUP_AGE`. The 16,384 slots were held permanently by whoever got in first *and stayed active*, and newcomers were locked out with no recovery path.

This is the permanently-refreshable GC exemption `AGENTS.md` forbids — *"any condition that exempts an entry from garbage collection MUST either expire via TTL, or be overridden by an absolute age threshold"*. There was only a refreshable TTL, and a busy pair refreshes it forever.

**Saturation is not evidence of an attack.** Tracked pairs are distinct senders × distinct contracts relayed from them, so 50 peers × ~330 contracts ≈ 16,500 already exceeds the cap with no attacker involved. The reported symptom was a capacity counter climbing 22K → 36K in a few minutes on an ordinary node. The module justified the bound purely as an anti-attacker measure and never reckoned with a healthy node simply outgrowing it.

Severity is silent data loss on the UPDATE path, not a noisy counter.

## Approach

This PR is **two coupled changes**, and the second is not optional — shipping the first alone would trade an availability bug for a security regression.

### 1. Evict LRU at capacity instead of refusing

The memory bound is unchanged: the map still never exceeds the cap, one in one out.

**Why batched (cap/64, so 256 at the default).** Under the saturation this targets the map is *persistently* full, so evicting a single entry per admission would walk all 16,384 entries on **every** new pair — at the reported 50-100 new pairs/sec, a continuous full scan several times a second on the UPDATE receive path. Batching amortises the scan to ~0.39/sec at 100 new pairs/sec, without changing the policy: oldest first either way. `select_nth_unstable_by_key` partitions rather than sorts, so the pass stays linear.

The shard guard is released before the walk. `evict_oldest` touches every shard, and holding a guard across that deadlocks — the same hazard already documented on this module's `size` counter. The retry after eviction is bounded (`MAX_ADMISSION_ATTEMPTS`) because the freed slot is not reserved: another thread may take it, and an unbounded loop would let contention spin on the receive path.

### 2. A per-sender new-pair budget, to replace the ceiling eviction removes

Refusing at capacity was *incidentally* cutting off an attacker churning fresh contract ids — after 16,384 of them, everything was refused. That ceiling was never designed (the cap exists to bound memory), and it denied every peer rather than just the flooder, but it was real, and eviction removes it. So it is **replaced deliberately rather than dropped**:

- A token bucket keyed by `sender_addr` (`NEW_PAIR_BURST` = 1024, `NEW_PAIR_REFILL_INTERVAL` = 5 ms → 200/s sustained), consulted only on the new-pair branch.
- A new `RateLimitDecision::SenderNewPairBudget`, so a budget refusal is distinguishable from a capacity drop.
- **Charged before a slot is reserved or anything is evicted**, so a sender past its budget cannot push other peers' entries out on its way to being refused.
- `TokenBucketLimiter` gains `BucketOutcome` + `check_and_record_detailed` so the budget can tell "over its rate" from "no room to track it". `check_and_record` keeps the collapsed `bool`, so the three resync limiters are behaviourally unchanged by this PR.
- `Ring::new` now takes `max_connections` to size the budget map (`max_connections × SENDER_TRACKING_HEADROOM`, floored at `MIN_TRACKED_SENDERS`). Note the coupling: raising the connection cap also enlarges this map.

**Honest accounting of what this trades.** The old aggregate ceiling was ~16,384 fresh pairs and then nothing, for everyone. The new one is 200/s **per sender** with no node-wide aggregate — at a 200-connection cap, a 40,000/s worst case of admitted relay work. That is a much higher attacker ceiling, and it is the right trade only because the old ceiling denied the victims too. The binding condition for legitimate traffic is generous: one peer must sustain ~210 UPDATE/s to this node continuously before anything drops, against real gateway load of ~3.15/s node-wide.

**Residual risks, stated rather than glossed** (all recorded in the module docs):
- A sender *within* its budget does evict other peers' entries, at up to 200/s. What makes that harmless is LRU plus the window it protects: evicting a pair touched within `MIN_UPDATE_INTERVAL` needs the whole map turned over inside 100 ms, ~164,000 evictions/s.
- An attacker churning at the budget rate can push the node into a high-miss-rate regime deliberately, converting every peer's per-pair limit into a per-peer 200/s aggregate.
- The budget's own map fails **open** at capacity (counted by `new_pair_budget_untracked_total`), and `Bucket::refill` restamps on every check, so a full sender map disables the budget for senders not already in it. Same shape as the bug this PR fixes, one map over; tracked in #5000 with the sibling instance in the resync limiters.
- The eviction scan holds a blocking `std::sync::Mutex` from inside an async task. At realistic load that is a 0.03% duty cycle; at a full-scale coordinated flood it approaches 10%. The `MAX_TRACKED_PAIRS` docs now carry that table so nobody raises the constant without redoing it.

Rejected alternatives: raising the cap treats the symptom and still starves eventually; failing open on the pair map lets a fresh-id attacker bypass rate limiting entirely.

## Visibility

The other half of why this went unnoticed: the drop logged at `debug!`, which `release_max_level_info` compiles out, so a production node discarding legitimate relayed UPDATEs left **no greppable evidence** and a dashboard tile was the only signal.

- Saturation now logs at `info!`, throttled to one line a minute, on eviction rather than per drop.
- `capacity_evicted_total`, `new_pair_budget_rejected_total` and `new_pair_budget_untracked_total` are surfaced through the **existing** `RingStatsSnapshot` rollup and three dashboard tiles — no new telemetry stream.
- The residual `CapacityExceeded` drop, now reachable only when the map is full *and* every admission attempt lost its slot to a concurrent caller, logs at `info!` too.

A healthy busy node now reads as evictions climbing with capacity drops flat, which is a sharper signal than the old counter gave. Note this **changes what the existing "Capacity-dropped" tile means**: it is no longer the saturation signal, and near-unreachable.

## Testing

~20 tests. The ones that carry the argument:

- `at_capacity_new_pair_is_admitted_and_the_map_stays_bounded` — the newcomer gets in and the cap still binds.
- `at_capacity_evicts_the_oldest_pairs_not_arbitrary_ones` — LRU order specifically. Membership is observed through the limiter's own behaviour (a tracked pair inside `min_interval` answers `Rejected`; an evicted one reads as new and answers `Allowed`) rather than a test-only accessor.
- `busy_pairs_cannot_hold_slots_against_newcomers` — the starvation regression, with every incumbent restamping itself each round and a sanity assertion that the fixture stays inside `CLEANUP_AGE` so the TTL sweep cannot be what admitted the newcomers.
- `eviction_counts_actual_removals_not_the_selected_batch` — deterministically reproduces a concurrent reaper taking a victim first. Over-counting drives `size` **below** the map's true length and lets the map grow past the cap.
- `a_throttled_sender_cannot_evict_other_peers_entries` — the security-relevant ordering, at a cap small enough for eviction to be reachable.
- `eviction_is_batched_at_larger_caps` — the divisor arithmetic, which the cap-8 fixtures clamp to 1 and never exercise.
- `capacity_signals_are_logged_above_debug_level` and `log_due_fires_once_per_interval` — a log level is a behavioural change no behavioural test can see. The source-scrape pin is bounded to the enclosing function body, because an unbounded backwards search resolves an imported `info!` to a *preceding* site's macro and lets a downgrade pass.

**Changed tests, and why** — stated here rather than left for a reader to discover:

- `capacity_exceeded_when_cap_reached` → `at_capacity_new_pair_is_admitted_and_the_map_stays_bounded`. The removed assertion pinned the starvation *as the contract*.
- `concurrent_distinct_keys_do_not_overshoot_cap`: `len() == CAP` relaxed to `<= CAP`. Exactly-the-cap became unachievable once a caller can free a slot and lose it (measured 44/800 flakes before the eviction scan was serialised, 7/800 after). Compensated by an exact conservation assertion (`len() == allowed - evicted`), plus new assertions that evictions actually happened and that `size` settles back onto `len` — without those the fixture passed under the *reverted* implementation.
- `MAX_DROPS_PER_THOUSAND` widened 1 → 2 in `concurrent_admission_at_capacity_almost_never_drops`, against a measured worst case of 7 drops across 540 trials, with the bug producing ~51. A stochastic bound on a scheduler-dependent quantity; the deterministic half of what it pins is covered by `an_empty_map_at_a_nonzero_cap_retries_rather_than_dropping`.
- `cleanup_decrements_size_counter` — forced change only, invariants intact.

`cargo clippy --locked -- -D warnings` and `cargo fmt` clean.

## Agent rules updated

`AGENTS.md` requires stale rules to be fixed in the PR that finds them, and this PR found one that instructs the exact mistake it fixes. `.claude/rules/code-style.md` said, unqualified, *"Reject new entries when the limit is reached"* — right for age-out-only collections, wrong for refresh-on-use ones, and the module cites it by name. It now carries the distinction, and `.claude/rules/bug-prevention-patterns.md` gains a section for the pattern.

## Not fixed here

- **#5000** — the same starvation shape in `TokenBucketLimiter`'s own map, affecting the three resync limiters. This PR adds the mechanism (`BucketOutcome`/`check_and_record_detailed`) but deliberately leaves the resync call sites on the collapsed `bool`, so their behaviour is unchanged. **PR #5027 changes those call sites**; it should land after this one and rebase onto it, since the enum and method it adds are duplicated here.

Closes #4981

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/bug-prevention-patterns.md
+++ b/.claude/rules/bug-prevention-patterns.md
@@ -478,6 +478,72 @@ reporting a healthy service as unhealthy — which is precisely the direction
 that survives unnoticed.
 
 
+## A count cap enforced by REFUSAL, over entries that ordinary use refreshes
+
+**A bound enforced by refusing newcomers is only safe when incumbents age out on
+their own.** If ordinary traffic restamps an entry's TTL, nothing ever rolls off:
+the cap is held permanently by whoever got in first and stayed active, and every
+newcomer is refused **forever**, with no recovery path. That is the
+permanently-refreshable GC exemption `AGENTS.md` forbids — wearing a memory bound
+as a disguise, which is why it reads as correct in review.
+
+#4981: the UPDATE limiter's 16,384 `(sender, contract)` slots. `*entry.get_mut() =
+now` on every accepted UPDATE meant a busy pair refreshed its own TTL indefinitely,
+so a new pair was refused, **never inserted**, and therefore stayed "new" forever —
+every subsequent UPDATE from it dropped and re-counted. Silent data loss on the
+UPDATE propagation path.
+
+Two things made it survive for months:
+
+- **Saturation was read as an attack signal, so exceeding the cap looked like the
+  bound working.** It is not. Tracked pairs are distinct senders x distinct
+  contracts, so 50 peers x ~330 contracts ~= 16,500 already exceeds the cap with
+  nobody malicious. The module justified its bound purely as an anti-attacker
+  measure and never reckoned with a healthy node simply outgrowing it.
+- **The drop logged at `debug!`**, which `release_max_level_info` compiles out, so
+  a production node discarding legitimate relayed UPDATEs left no greppable
+  evidence and a dashboard tile was the only signal.
+
+### Fix shape (#4997)
+
+**Evict least-recently-used instead of refusing**, and:
+
+- **Evict a BATCH** (`cap / 64`), not one entry. The victim scan is linear in the
+  cap and cannot run while holding a shard guard, so under the saturation this
+  targets — where the map is *persistently* full — one-per-admission means a
+  continuous full scan on the receive path.
+- **Release the shard guard before the walk**, and bound the post-eviction retry:
+  the freed slot is not reserved, so another caller may take it.
+- **Replace any incidental ceiling eviction removes.** Refuse-at-cap was also
+  (accidentally) throttling attacker-chosen keys. Add the replacement explicitly
+  and charge it **before** a slot is reserved, so a throttled peer cannot evict
+  anyone on its way to being refused.
+- **Count what you ACTUALLY removed, not the batch you selected.** A concurrent
+  reaper can take a victim first; over-counting drives the size counter **below**
+  the map's true length, letting the map grow past the cap — the one bound the
+  whole mechanism exists to enforce. Note this is unobservable unless a test
+  interleaves the reaper with an admission: with no such test, `remove()` always
+  returns `Some` and the mutation is behaviourally invisible.
+- **Make saturation visible in RELEASE builds**: `info!` plus a counter, never
+  `debug!` alone.
+
+### Where the wrong instruction came from
+
+`.claude/rules/code-style.md` said, without qualification, *"Reject new entries when
+the limit is reached"*. That is right for age-out-only collections and wrong for
+refresh-on-use ones. It now carries the distinction; keep the two files in step.
+
+Audit question for any bounded per-key map: **does ordinary use restamp this
+entry's TTL?** If yes, reject-at-cap is a starvation bug, not a bound.
+
+```bash
+# Bounded per-key maps whose entries are restamped by ordinary use:
+grep -rnE "get_mut\(\) = now|last_seen = now|last_refill = now" crates/core/src/
+# ...cross-check each against how its cap is enforced:
+grep -rnE "max_tracked|MAX_TRACKED" crates/core/src/
+```
+
+
 ## A refusal that is not counted renders as a clean zero
 
 **Any code path that DISCARDS an input must count the discard.** Three bare
--- a/.claude/rules/code-style.md
+++ b/.claude/rules/code-style.md
@@ -88,8 +88,43 @@ actors (clients, network peers) can influence.
 
 1. Per-key collections (subscribers per contract, peers per resource)
    MUST have a maximum size enforced at insertion time.
-   → Reject new entries when the limit is reached
    → Return an error or false so callers know registration was rejected
+
+   BUT: how you enforce it depends on whether entries are REFRESHED BY
+   ORDINARY USE. Get this wrong and the bound starves newcomers forever.
+
+   → Entries that only AGE OUT (a TTL nothing resets): reject new
+     entries at the cap. Incumbents roll off on their own, so a
+     newcomer's wait is bounded.
+
+   → Entries REFRESHED ON EVERY USE (a `last_seen`/`last_refill` stamp
+     that ordinary traffic restamps): reject-at-cap is WRONG — you MUST
+     EVICT the least-recently-used entry instead. A busy entry refreshes
+     its own TTL forever, so the cap is held permanently by whoever got
+     in first and stayed active, and every newcomer is refused with no
+     recovery path. That is the permanently-refreshable GC exemption
+     AGENTS.md forbids, and this rule's earlier "reject new entries when
+     the limit is reached" wording is what produced it.
+
+     #4981: the UPDATE limiter's 16,384 `(sender, contract)` slots were
+     held by whoever got in first, and every new pair's UPDATE was
+     silently dropped — and dropped at `debug!`, which
+     `release_max_level_info` compiles out, so a production node
+     discarding legitimate relayed traffic left no greppable evidence.
+     Fixed in #4997 by LRU eviction.
+
+     Two things that make eviction safe rather than a new hole:
+       a. Eviction can remove an incidental ceiling. If refusing-at-cap
+          was also (accidentally) throttling attacker-chosen keys, you
+          need an explicit replacement — #4997 added a per-sender
+          new-pair token bucket, charged BEFORE any slot is reserved so
+          a throttled peer cannot evict anyone on its way to refusal.
+       b. The scan is linear in the cap and cannot run under a shard
+          guard. Evict a BATCH, not one entry, or a persistently-full
+          map means a full scan per admission on the receive path.
+
+   → Whichever you pick, saturation must be visible in RELEASE builds:
+     `info!` or a counter, never `debug!` alone.
 
 2. Per-client/per-peer resource counts MUST be bounded.
    → A single client must not hold unbounded subscriptions across all keys
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1598,35 +1598,55 @@ where
                     .update_rate_limiter
                     .check_and_record(sender_addr, *key.id());
                 if !rate_decision.is_allowed() {
-                    if matches!(
-                        rate_decision,
-                        crate::ring::update_rate_limit::RateLimitDecision::CapacityExceeded
-                    ) {
-                        // `info!` because `debug!` is compiled out of release
-                        // builds by `release_max_level_info`, which is how this
-                        // path came to drop legitimate relayed UPDATEs with no
-                        // greppable evidence on a production node (#4981).
-                        // Since that fix the limiter evicts instead of refusing,
-                        // so reaching here means the map was full AND every
-                        // admission attempt lost its freed slot to a concurrent
-                        // caller — rare, and worth seeing when it happens.
-                        tracing::info!(
-                            tx = %op.id(),
-                            %key,
-                            %sender_addr,
-                            phase = "update_dispatch_capacity_dropped",
-                            "UPDATE dispatch: dropped, rate-limiter map at capacity and \
-                             contended (see ring::update_rate_limit)"
-                        );
-                    } else {
-                        tracing::debug!(
-                            tx = %op.id(),
-                            %key,
-                            %sender_addr,
-                            ?rate_decision,
-                            phase = "update_dispatch_rate_limited",
-                            "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
-                        );
+                    use crate::ring::update_rate_limit::RateLimitDecision;
+                    // Matched exhaustively on purpose: a new drop reason
+                    // that fell into a catch-all would be logged at
+                    // `debug!` and therefore be invisible in release
+                    // builds, which is exactly how #4981 stayed hidden.
+                    match rate_decision {
+                        RateLimitDecision::CapacityExceeded => {
+                            // `info!` because `debug!` is compiled out of release
+                            // builds by `release_max_level_info`, which is how this
+                            // path came to drop legitimate relayed UPDATEs with no
+                            // greppable evidence on a production node (#4981).
+                            // Since that fix the limiter evicts instead of refusing,
+                            // so reaching here means the map was full AND every
+                            // admission attempt lost its freed slot to a concurrent
+                            // caller — rare, and worth seeing when it happens.
+                            tracing::info!(
+                                tx = %op.id(),
+                                %key,
+                                %sender_addr,
+                                phase = "update_dispatch_capacity_dropped",
+                                "UPDATE dispatch: dropped, rate-limiter map at capacity and \
+                                 contended (see ring::update_rate_limit)"
+                            );
+                        }
+                        RateLimitDecision::SenderNewPairBudget => {
+                            // Deliberately NOT logged per drop: a sender
+                            // over this budget is over it on every
+                            // message, so a line here would be a steady
+                            // stream. The limiter logs it once a minute
+                            // with the sender, and the drop is counted in
+                            // `new_pair_budget_rejected_total`.
+                            tracing::debug!(
+                                tx = %op.id(),
+                                %key,
+                                %sender_addr,
+                                phase = "update_dispatch_fresh_id_dropped",
+                                "UPDATE dispatch: dropped, sender over its new-pair budget"
+                            );
+                        }
+                        RateLimitDecision::Rejected { .. } | RateLimitDecision::Allowed => {
+                            tracing::debug!(
+                                tx = %op.id(),
+                                %key,
+                                %sender_addr,
+                                ?rate_decision,
+                                phase = "update_dispatch_rate_limited",
+                                "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
+                            );
+                        }
                     }
                     return Ok(());
                 }
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1598,14 +1598,36 @@ where
                     .update_rate_limiter
                     .check_and_record(sender_addr, *key.id());
                 if !rate_decision.is_allowed() {
-                    tracing::debug!(
-                        tx = %op.id(),
-                        %key,
-                        %sender_addr,
-                        ?rate_decision,
-                        phase = "update_dispatch_rate_limited",
-                        "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
-                    );
+                    if matches!(
+                        rate_decision,
+                        crate::ring::update_rate_limit::RateLimitDecision::CapacityExceeded
+                    ) {
+                        // `info!` because `debug!` is compiled out of release
+                        // builds by `release_max_level_info`, which is how this
+                        // path came to drop legitimate relayed UPDATEs with no
+                        // greppable evidence on a production node (#4981).
+                        // Since that fix the limiter evicts instead of refusing,
+                        // so reaching here means the map was full AND every
+                        // admission attempt lost its freed slot to a concurrent
+                        // caller — rare, and worth seeing when it happens.
+                        tracing::info!(
+                            tx = %op.id(),
+                            %key,
+                            %sender_addr,
+                            phase = "update_dispatch_capacity_dropped",
+                            "UPDATE dispatch: dropped, rate-limiter map at capacity and \
+                             contended (see ring::update_rate_limit)"
+                        );
+                    } else {
+                        tracing::debug!(
+                            tx = %op.id(),
+                            %key,
+                            %sender_addr,
+                            ?rate_decision,
+                            phase = "update_dispatch_rate_limited",
+                            "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
+                        );
+                    }
                     return Ok(());
                 }
 
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1487,35 +1487,55 @@ where
                     .update_rate_limiter
                     .check_and_record(sender_addr, *key.id());
                 if !rate_decision.is_allowed() {
-                    if matches!(
-                        rate_decision,
-                        crate::ring::update_rate_limit::RateLimitDecision::CapacityExceeded
-                    ) {
-                        // `info!` because `debug!` is compiled out of release
-                        // builds by `release_max_level_info`, which is how this
-                        // path came to drop legitimate relayed UPDATEs with no
-                        // greppable evidence on a production node (#4981).
-                        // Since that fix the limiter evicts instead of refusing,
-                        // so reaching here means the map was full AND every
-                        // admission attempt lost its freed slot to a concurrent
-                        // caller — rare, and worth seeing when it happens.
-                        tracing::info!(
-                            tx = %op.id(),
-                            %key,
-                            %sender_addr,
-                            phase = "update_dispatch_capacity_dropped",
-                            "UPDATE dispatch: dropped, rate-limiter map at capacity and \
-                             contended (see ring::update_rate_limit)"
-                        );
-                    } else {
-                        tracing::debug!(
-                            tx = %op.id(),
-                            %key,
-                            %sender_addr,
-                            ?rate_decision,
-                            phase = "update_dispatch_rate_limited",
-                            "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
-                        );
+                    use crate::ring::update_rate_limit::RateLimitDecision;
+                    // Matched exhaustively on purpose: a new drop reason
+                    // that fell into a catch-all would be logged at
+                    // `debug!` and therefore be invisible in release
+                    // builds, which is exactly how #4981 stayed hidden.
+                    match rate_decision {
+                        RateLimitDecision::CapacityExceeded => {
+                            // `info!` because `debug!` is compiled out of release
+                            // builds by `release_max_level_info`, which is how this
+                            // path came to drop legitimate relayed UPDATEs with no
+                            // greppable evidence on a production node (#4981).
+                            // Since that fix the limiter evicts instead of refusing,
+                            // so reaching here means the map was full AND every
+                            // admission attempt lost its freed slot to a concurrent
+                            // caller — rare, and worth seeing when it happens.
+                            tracing::info!(
+                                tx = %op.id(),
+                                %key,
+                                %sender_addr,
+                                phase = "update_dispatch_capacity_dropped",
+                                "UPDATE dispatch: dropped, rate-limiter map at capacity and \
+                                 contended (see ring::update_rate_limit)"
+                            );
+                        }
+                        RateLimitDecision::SenderNewPairBudget => {
+                            // Deliberately NOT logged per drop: a sender
+                            // over this budget is over it on every
+                            // message, so a line here would be a steady
+                            // stream. The limiter logs it once a minute
+                            // with the sender, and the drop is counted in
+                            // `new_pair_budget_rejected_total`.
+                            tracing::debug!(
+                                tx = %op.id(),
+                                %key,
+                                %sender_addr,
+                                phase = "update_dispatch_fresh_id_dropped",
+                                "UPDATE dispatch: dropped, sender over its new-pair budget"
+                            );
+                        }
+                        RateLimitDecision::Rejected { .. } | RateLimitDecision::Allowed => {
+                            tracing::debug!(
+                                tx = %op.id(),
+                                %key,
+                                %sender_addr,
+                                ?rate_decision,
+                                phase = "update_dispatch_rate_limited",
+                                "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
+                            );
+                        }
                     }
                     return Ok(());
                 }
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1487,14 +1487,36 @@ where
                     .update_rate_limiter
                     .check_and_record(sender_addr, *key.id());
                 if !rate_decision.is_allowed() {
-                    tracing::debug!(
-                        tx = %op.id(),
-                        %key,
-                        %sender_addr,
-                        ?rate_decision,
-                        phase = "update_dispatch_rate_limited",
-                        "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
-                    );
+                    if matches!(
+                        rate_decision,
+                        crate::ring::update_rate_limit::RateLimitDecision::CapacityExceeded
+                    ) {
+                        // `info!` because `debug!` is compiled out of release
+                        // builds by `release_max_level_info`, which is how this
+                        // path came to drop legitimate relayed UPDATEs with no
+                        // greppable evidence on a production node (#4981).
+                        // Since that fix the limiter evicts instead of refusing,
+                        // so reaching here means the map was full AND every
+                        // admission attempt lost its freed slot to a concurrent
+                        // caller — rare, and worth seeing when it happens.
+                        tracing::info!(
+                            tx = %op.id(),
+                            %key,
+                            %sender_addr,
+                            phase = "update_dispatch_capacity_dropped",
+                            "UPDATE dispatch: dropped, rate-limiter map at capacity and \
+                             contended (see ring::update_rate_limit)"
+                        );
+                    } else {
+                        tracing::debug!(
+                            tx = %op.id(),
+                            %key,
+                            %sender_addr,
+                            ?rate_decision,
+                            phase = "update_dispatch_rate_limited",
+                            "UPDATE dispatch: rejected by per-(sender, contract) rate limit"
+                        );
+                    }
                     return Ok(());
                 }
 
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -79,9 +79,16 @@ pub struct RingStatsSnapshot {
     /// may be getting dropped — operators should watch this.
     pub updates_rate_limited: u64,
     /// Total relayed UPDATEs dropped because the limiter's tracking map
-    /// was at capacity (`MAX_TRACKED_PAIRS`). A non-zero value suggests
-    /// identity churn / admission pressure, distinct from per-pair rate.
+    /// was at capacity (`MAX_TRACKED_PAIRS`) and eviction could not free
+    /// a slot. Since #4981 this means the map is full *and* contended;
+    /// ordinary saturation shows up in `updates_capacity_evicted`.
     pub updates_capacity_dropped: u64,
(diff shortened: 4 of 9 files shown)


sanity:
## Review findings addressed

All four blocking findings plus the three follow-ups. Every number below is measured on this branch, before and after.

### 1. An eviction that removed nothing was read as "the cap is 0" (blocking)

Confirmed and reproduced. At the production cap, 16 threads x 2000 new pairs, ten trials: **2699 drops in 320 000 admissions (0.84%), 10/10 trials affected**. Instrumenting the two branches showed **100%** came from `evicted == 0` with a non-empty map; the documented "map is empty" case never fired and the retry budget was never reached — the drop did not even spend it.

`evict_oldest` now returns an enum, so the three ways a pass can remove nothing have to be told apart at the call site instead of collapsing into a count. Two of them mean a slot is or may already be free and the admission retries; only an empty map is terminal.

**0 drops in 320 000** after.

### 2. Concurrent evictors were colliding on the same victims

Found while chasing the residual after the first fix, and it is the mechanism behind it: concurrent evictors partition the *same* map, so they select nearly the same oldest entries and then race to remove what each other already took. Measured: **44% of eviction passes removed nothing**, and the average pass removed **24 entries for a full 16 384-entry scan** against a 256-entry batch — the batching was not amortising anything.

The scan is now serialised, with a lock-free check before it and a re-check under the lock so a caller arriving after the batch is freed does no scan at all. Evictions per trial went from ~656 ragged scans to **exactly 61 full batches of 256**.

A caller that frees room now also keeps one of the slots it freed. Whether it won the slot back otherwise depended on batch size versus caller count — fine at the production cap (batch 256), but at a 1024 cap the batch is 16 and ~2 in 1000 admissions still dropped. The strict cap is unaffected (`size` counts entries plus outstanding reservations) and so is batching (`removed - 1` slots of headroom remain).

`MAX_ADMISSION_ATTEMPTS` 3 -> 6, measured on that worst ratio: three attempts dropped 7-26 of 6400 across a dozen runs, four 0-3, five 0-2, six none.

### 3. `concurrent_distinct_keys_do_not_overshoot_cap` was flaky (blocking)

Confirmed: **7 failures in 800 runs** on this branch (the review measured 44/800 before the serialisation reduced the race), always `len() == 7` against an expected 8.

The assertion was wrong, not the timing. Admissions are not conserved under eviction — a caller can free a slot and then lose it on every remaining attempt — so a run can legitimately end below the cap. Now an upper bound, plus the conservation identity that IS exact: `len() == allowed - evicted`.

**0 failures in 1000 runs.**

### 4. The LRU test did not pin eviction order (blocking)

Confirmed: deleting `select_nth_unstable_by_key` survived **349 of 400 runs** (12.75% kill — 1/8, the chance an arbitrary victim is the right one). The final assertion was vacuous exactly as described: by then the clock had run 108ms, so an evicted pair and a surviving one both answered `Allowed`.

Replaced with two fixtures at a 128 cap (batch 2) that keep every pair inside `min_interval` — stamping a microsecond apart and asserting the run stays inside the window — and assert the exact victim set across three batches. An arbitrary-victim implementation would have to pick the exact oldest pairs out of 128 candidates three rounds running.

The "a refresh moves a pair to the back" property got its own fixture, since the refresh is what destroys the discriminator: it refreshes all but six pairs, so the only entries outside the window are the six that must be evicted and every pair asserted on is observable.

Both mutants die **150/150**. The refresh fixture also kills a second mutant (dropping the restamp) that the arbitrary-victim fixture does not.

### 5. Stale `capacity_rejected_total` doc (blocking)

Fixed. It no longer says a non-zero value suggests an attacker churning identities — it contradicted the struct-field and variant docs in the same file, and that role now belongs to the fresh-id counter below.

### 6. The weakened rate limit — compensating control added

Checked the claim independently and it holds: nothing else in the node bounds a sender's fresh-contract-id churn. The ban list is keyed by contract; the MAD detector is `GovernanceMode::Off` by hardcoded default with no config toggle, and its 15-minute ramp-up would exclude fresh ids anyway; the transport rate-limits handshake packets and paces outbound bytes, with no inbound limit on an established connection. And a node does forward UPDATEs for contracts it does not host (`operations/update/op_ctx_task.rs`), so an unbounded fresh-id stream makes it a relay amplifier.

Added a token bucket keyed by `sender_addr` alone, bounding how fast one peer may introduce pairs the limiter has never seen (512 burst, then one per `MIN_UPDATE_INTERVAL`). Two properties matter:

- **Charged before the capacity path**, so a sender churning fresh ids cannot evict other peers' entries on its way to being throttled.
- **Never charged for an already-tracked pair**, so a peer's established traffic is bounded by `MIN_UPDATE_INTERVAL` per pair and by nothing else — the availability this PR exists to restore is not clawed back. Pinned by a test that runs 200 rounds over established pairs after the budget is spent and asserts not one is throttled.

One deviation from the suggestion, flagged deliberately: this bounds **new-pair introduction rate** rather than total UPDATE volume per peer. That is the property the old refuse-at-cap behaviour actually provided, and it carries no false-positive risk for a busy peer's established traffic, which a volume cap would. A volume cap would additionally bound a peer hammering many *established* pairs — but that is unchanged by this PR, not a regression it introduces. Happy to add the volume cap too if you would rather have both.

Its map is bounded by a multiple of the connection cap (`sender` is the immediate upstream hop, so only a connected peer can occupy an entry) and swept by the existing reaper.

### 7. Unthrottled `info!` at the dispatch site

Resolves as predicted: with `CapacityExceeded` now at **0 in 320 000** admissions at the production cap, that line does not fire. The new fresh-id drop is deliberately NOT logged per message at the dispatch site — a sender over budget is over it on every message — so it logs once a minute inside the limiter, with the sender, and the drop is counted.

### Also

- `a_zero_cap_refuses_every_pair_without_spinning` — the terminal `CapacityExceeded` path had no test at all, and it is the one path the enum's `MapEmpty` variant exists for.
- The release-visibility pin now covers the fresh-id line too, since that is its only release-visible evidence. Mutation-checked: downgrading it to `debug!` fails the pin.
- New counter surfaced through the existing `RingStatsSnapshot` rollup and its dashboard tile, same path as `capacity_evicted_total`.

### Verification

- `cargo clippy --locked -- -D warnings` and `cargo fmt --all --check` clean (CI's exact commands)
- 4344 lib tests pass
- The five concurrency tests: **0 failures in 400 runs each**
- The new regression test fails at **671 drops in 6400** with the original bug reintroduced

[AI-assisted - Claude]

sanity:
## Second review round (`ae7f4301`)

I ran an adversarial re-read over my own first commit. It found the same mistake I had just fixed, surviving one branch over, and a false claim in the docs of the control I added. Both are fixed.

### An empty map was still read as "the cap must be zero", and dropped

The first commit stopped treating "this eviction pass removed nothing" as terminal — but `evict_oldest` still returned a terminal outcome when the map came back *empty*, inferring the cap must be zero. Two windows put `size` at the cap while the map is not:

- `cleanup` sweeping — it decremented `size` once at the end, so `size` read full for the whole sweep while the map emptied.
- Every caller that has won a slot sitting between its reservation and its insert (reachable at small caps under contention: 8 slots, 64 callers).

In both, slots are about to exist. The only genuinely terminal condition is a cap of zero, which is a property of the configuration — so it is now read from the configuration rather than inferred. `MapEmpty` became `CapIsZero`, and `cleanup` decrements per removal.

Pinned by `an_empty_map_at_a_nonzero_cap_retries_rather_than_dropping`, which forces the condition rather than racing for it and fails deterministically with the fix reverted.

Worth stating plainly: **I could not reproduce the contention window empirically** — 0 occurrences in 300 runs of the 8-cap/64-thread fixture, with and without the fix. So that fixture deliberately does *not* assert on it; the pin would have been vacuous. The fix rests on the `cleanup` window, which is clear by inspection, and on the white-box test.

### "Established traffic never touches the budget" was false

This was the sharpest finding, and it undercut the control I added last round. A token is spent whenever the pair is **not currently in the map** — and eviction is what takes pairs out of it. So on a node whose working set exceeds the cap, ordinary relayed traffic keeps re-entering the new-pair path and being charged. The limiter cannot tell a returning pair from a genuinely fresh one without remembering what it evicted, which is exactly the unbounded memory the cap exists to refuse.

At the #4981 working set (~16 500 pairs vs a 16 384 cap) the miss rate is under 1% and it does not bind. On a larger gateway it would have — turning a control meant to bound fresh-id churn into a per-peer aggregate ceiling of **10 UPDATEs/sec**, dropping legitimate relayed traffic on exactly the busy node this PR exists to keep serving.

I corrected the claim rather than papering over it, and recalibrated for what the control actually is: **200/s sustained (was 10/s), 1024 burst** — twenty times the per-pair ceiling, so a peer must sustain 200 UPDATEs/s to this node continuously before anything is dropped, while a fresh-id flood is still bounded rather than unlimited. Mis-calibration is visible as `new_pair_budget_rejected_total` climbing on a node with no attacker, which is why the counter is on the dashboard.

The saturated case now has the fixture it was missing (a 64-entry cap cycling 128 contracts from one sender, so every check is a re-admission after eviction): at the sustained rate, twenty full cycles, nothing is dropped — and with the clock stopped the same peer is still cut off after its burst, so the bound is real.

**This is the one I would most want a second opinion on.** The rate is set by reasoning, not production data, and the honest position is that under heavy saturation this is a per-peer aggregate ceiling rather than the surgical fresh-id bound I described last round.

### The budget map could starve newcomers once full

`TokenBucketLimiter` returns the same `false` for "over its rate" and "no room for this key", and `Bucket::refill` restamps `last_refill` on every check including denied ones — so an active sender's entry never ages out and a newcomer arriving at a full map would be refused indefinitely. That is the permanently-refreshable GC exemption `AGENTS.md` forbids, and the #4981 starvation shape, one map over.

It now reports the two separately (`check_and_record` keeps the collapsed form, so the resync limiters are untouched — their 16 tests pass), and the UPDATE limiter **fails open** on a full map: `sender` is the immediate upstream hop rather than an attacker-chosen key, so a full map is a sizing accident, and refusing on it would starve a peer for someone else's behaviour. Counted separately so an undersized map is visible rather than silent.

The map is now sized from the node's configured `max_connections` rather than `DEFAULT_MAX_CONNECTIONS`, per the derive-from-config rule in `.claude/rules/ring.md`.

### Smaller

- The eviction log moved outside the eviction lock.
- The log-level pin searches back to the nearest `tracing::` instead of a fixed 200-char window — the eviction call site was already within ~19 characters of outgrowing it, and one more structured field would have broken the pin. Mutation-checked at all three sites.
- `cleanup_keeps_the_size_counter_in_step_as_it_sweeps` renamed to `cleanup_leaves_the_size_counter_equal_to_the_map_length`: it passed with the batched decrement, so it never tested the mid-sweep property its name claimed. It is non-vacuous for what it does pin (verified against a no-decrement mutant).
- A stale test-name reference and an overstated "order of magnitude in both directions" margin claim fixed.

### Verification

- `cargo clippy --locked -- -D warnings` and `cargo fmt --all --check` clean
- 4347 lib tests pass, including the 16 on the shared token-bucket primitive
- Six concurrency fixtures: 0 failures in 300 runs each

[AI-assisted - Claude]

sanity:
## Re-review fixes (`a4727bbb`)

Both fixes in, plus the optional comment corrections. Follow-up filed as #5000.

### Finding B — the sender map now pins fail-open

You were right and this was the important one. `the_new_pair_budget_map_is_bounded` asserted only that the map stays bounded, never that a sender arriving at a **full** map is admitted. I confirmed your mutant: fail-closed passes the whole suite. It now fails.

Renamed to `the_new_pair_budget_map_is_bounded_and_fails_open_when_full`, asserting the admission, the untracked count, and that nothing is charged as a rejection (a full map is a sizing accident, not evidence about any sender).

While there: `new_pair_budget_untracked_total` was written and never read, despite its doc calling it an operator signal — a safety valve nobody can see firing is the #4981 failure mode. Now surfaced through the existing `RingStatsSnapshot` rollup and a dashboard tile, same path as the others. That's the one non-test change in this commit; it's additive and matches the existing counter pattern exactly, so flagging it in case you want it looked at.

### Finding A — docstring corrected, and the exit now has a pin that reaches it

Verified your claim rather than taking it: the narrow `removed == 0` mutant survives **15/15** on that fixture and **29/29** across the suite. Your explanation is right — with the scan serialised, concurrent evictors no longer steal each other's victims, so `removed == 0` never occurs there.

The 671 figure isn't wrong, it's attributed wrong: that's what the **caller's** `Retry` arm produces, which covers the early-out and the under-lock re-check — where the measured drops actually came from. Docstring now says which, and explicitly says what it does *not* pin.

Added `evict_oldest_has_exactly_one_terminal_exit`, a source scrape asserting exactly one terminal exit guarded by the cap being zero. Your `cleanup` reasoning is the whole justification and it's recorded in the test: `cleanup` doesn't take the eviction lock and removes entries older than `CLEANUP_AGE` — precisely the oldest, precisely what `select_nth_unstable_by_key` selects — so a sweep landing between the collect and the removes takes the batch out from under the evictor, once a minute on every node. No fixture can schedule that, so the pin is on the shape. **Both terminal-exit mutants now die**, including the one that survived 15/15.

### Comment corrections

- **The "<1% miss rate" claim was wrong**, and worse than wrong-in-degree — replaced with your numbers and the binding condition. 1.33% uniform-random but **96.77% round-robin** at ~16,500 (LRU's worst case: every entry evicted exactly before its next use). The doc no longer rests the safety argument on the miss rate being small, because it isn't; it rests on one peer needing to sustain ~210 UPDATE/s continuously, against ~3.15/s node-wide on nova and ~2.84/s on vega.
- **Reservation transfer:** ~2 in 1000 → **~0.19 in 1000**. Both figures predated the eviction lock. Noted that what the reservation actually buys is independence from the batch/caller ratio, not the residual itself.
- **`MAX_TRACKED_PAIRS` now documents the scaling trap:** hold time is linear in the cap, 41 ns/entry — 672 µs at 16,384, **10.8 ms at 163,840** — with an explicit re-measure-before-raising note, since this PR installs the "evictions climbing = saturation" reading and the obvious response is to raise the cap.
- Stale `EvictionOutcome::MapEmpty` reference fixed.

### Follow-up filed: #5000

`TokenBucketLimiter` can permanently refuse new keys once its map is full — `Bucket::refill` restamps on denials, `cleanup` only drops `recovered && idle`, so an active key never ages out and a newcomer at the cap never recovers. The UPDATE limiter is unaffected (it fails open via `check_and_record_detailed`); the three resync limiters still use the collapsed form. Not reachable today (nova: 1,962 contracts vs a 16,384 cap) and it degrades resync rather than dropping UPDATEs, but it shouldn't be left looking fixed.

### Verification

`cargo clippy --locked -- -D warnings` and `cargo fmt --all --check` clean; 4348 lib tests pass; the four affected fixtures 0 failures in 200 runs each.

[AI-assisted - Claude]

sanity:
## Multi-perspective review — Full tier

A rate limiter is a DoS surface and the diff is large, so this got the Full treatment: **four independent blind reviewers** (code-first, testing-adequacy, skeptical/adversarial, big-picture) run in parallel, each told not to rubber-stamp, plus synthesis. Codex was not run; four diverse Claude lenses were substituted, recorded here per the review rule.

The branch has been **rebased onto current `main`** (clean), because it was a month behind and this PR edits `.claude/rules/*` files that `main` had moved a long way — editing the stale copies would have silently reverted them.

### Verdict

The eviction algorithm is **correct**. I could not break the cap or the size counter, and neither could the adversarial lens. The substantive findings were in the PR's *description*, in claims the module docs made, and in two tests that pinned nothing. All are addressed below.

### What was verified, and found safe

- **The map cannot exceed the cap under concurrency.** `size.fetch_add` is the single serialization point; a caller that reserves during another's scan sees `size` still at the cap, blocks on `eviction_lock`, re-checks, and takes the cheap `Retry`. No path admits past the cap.
- **`size` cannot drift from the map length.** Every decrement pairs with a real removal, and the reservation is handed back on every exit including the "pair appeared while we were evicting" branch. Underflow is impossible for the same reason.
- **Eviction is global, not per-shard.** `iter()` walks every shard; `select_nth_unstable_by_key(batch - 1)` guarantees the `batch` smallest by timestamp; `batch` is clamped to `entries.len()` and floored at 1, so the index is always in range. **No key pattern lets an attacker concentrate victims into a favoured shard.**
- **The starvation inversion does not work.** Accepted UPDATEs restamp, so a flooding pair is always the newest and the *last* evicted; victims are quiet pairs that would have aged out under `CLEANUP_AGE` anyway. Evicting your own entry to reset your rate window costs ~16,384 fresh-pair tokens to buy one extra UPDATE.
- **No lock cycle.** `eviction_lock` is outermost and taken with no shard guard held. The one nested acquisition has no inverse. The retry loop is bounded and cannot spin hot; a zero cap terminates via the config read rather than by inferring emptiness from a full map.
- **`bug-prevention-patterns.md`'s "cache bounded by entry COUNT" row does not apply, and it is worth saying why explicitly.** Both maps hold **fixed-size values** — `DashMap<(SocketAddr, ContractInstanceId), Instant>` and `Bucket { f64, Instant }`. Neither holds a `StateDelta`/`StateSummary`/`WrappedState`/`Vec<u8>`, so a count cap **is** a memory bound here and `ByteBoundedLruCache` is not needed. No new byte budget is introduced, so there is nothing to register in the hosting residual.

### Findings addressed

**1. The PR description described a different PR — all four lenses flagged this independently.** The body did not mention the **per-sender new-pair token bucket** at all: a new security control with its own constants, a new `RateLimitDecision` variant, a new `BucketOutcome`/`check_and_record_detailed` on the shared `TokenBucketLimiter`, a `Ring::new` signature change, and three dashboard tiles. Worse, its Approach section argued the *opposite* position — that "per-`(sender, contract)` rate limiting cannot stop a fresh-id attacker regardless" — while the code implements exactly such a bound. Since squash-merge uses the body as the commit message, `main`'s history would have permanently recorded a false account of what shipped. **Body rewritten** to describe what is actually here, including the honest accounting of what the budget trades (the old ceiling was ~16,384 fresh pairs then nothing for everyone; the new one is 200/s *per sender* with no node-wide aggregate) and four residual risks that were previously unstated.

**2. `.claude/rules/code-style.md` still instructed the exact mistake this PR fixes.** It said, unqualified, *"Reject new entries when the limit is reached"* — right for age-out-only collections, wrong for refresh-on-use ones, and the module cites it by name. `AGENTS.md` requires stale rules to be fixed in the PR that finds them. It now carries the distinction, and **`bug-prevention-patterns.md` gains a section** for the pattern: the fix shape, the two things that made #4981 survive for months, and the two traps inside the fix itself.

**3. The "counts attempted rather than actual removals" mutant was dead at HEAD.** No test interleaved `cleanup()` with `check_and_record()`, so `remove()` always returned `Some` and `removed == batch` unconditionally — mutating the count to `removed = batch` was behaviourally invisible across the whole suite. This is the DoS-relevant one: over-counting drives `size` **below** the map's true length and lets the map grow past the cap. Fixed by splitting `remove_victims` out of `evict_oldest` (inside it, the victim list is collected and consumed in the same breath, so no fixture can reach the window without racing a real reaper) and pinning it with a deterministic test.

**4. `a_sender_churning_fresh_contract_ids_is_cut_off_after_its_burst` asserted an ordering its fixture could not exercise.** It runs at `MAX_TRACKED_PAIRS` with 40 pairs, where nothing can evict whatever the ordering is — so `capacity_evicted_total() == 0` passes even with the budget check moved *after* the reserve/evict block. That ordering is what makes the budget a security control rather than a counter. Added `a_throttled_sender_cannot_evict_other_peers_entries` at a cap of 8, which asserts the fixture actually evicts before relying on the discrimination.

**Both new tests mutation-verified, each mutant applied in isolation:** counting the selected batch kills the first while the second still passes; skipping the budget check at capacity kills the second (and `under_saturation_re_admitted_pairs_are_not_throttled_at_the_sustained_rate`) while the first still passes.

**5. The log-level pin could be defeated silently.** `level_of` searched backwards for `tracing::` across the **whole file**, so a site written with an imported `info!`/`debug!` resolves to whatever macro precedes it — for `log_new_pair_budget` that is the eviction site's `tracing::info!` a few lines up, and a downgrade would pass the pin. `AGENTS.md` names this failure shape and says it has shipped twice here. Now bounded to the enclosing function body via an `fn_body` helper, with a panic message that tells the next person to restore the qualified form rather than widen the search.

**6. `log_due` gated both release-visible signals and had no test at all.** Muting it or unthrottling it were both invisible to the suite. Pinned directly, which also sidesteps the `#5314` process-global `Interest` trap a subscriber-capture test would have walked into.

**7. `concurrent_distinct_keys_do_not_overshoot_cap` passed under the reverted implementation.** `allowed == CAP` satisfies `>= CAP`, `len() == CAP` satisfies `<= CAP`, and conservation holds trivially at `evicted == 0` — so the fixture's comment about threads "racing to evict and insert" was aspirational. Added `capacity_evicted_total() > 0` and a deterministic `size == len()` check after all threads join. (The `len() == CAP` → `<= CAP` relaxation itself is **justified**, not a weakening: equality became unachievable once a caller can free a slot and lose it, and the overshoot class it guarded is still caught by `<=`.)

**8. Documentation corrections.** The batch amortisation figure was ~4× pessimistic — it divided the arrival rate by the *divisor* rather than by the *batch*; 100 new pairs/s costs ≈0.39 scans/s, not 1.5, and that figure is what a future reader uses when sizing `MAX_TRACKED_PAIRS`. The `[MAX_TRACKED_SENDERS]` intra-doc link pointed at a constant that does not exist. The cross-peer isolation claim was unconditional but only holds for churn *past* the budget — a sender inside its allowance does evict other peers' entries, and the LRU-turnover argument that makes that harmless was nowhere written down. The eviction log's `tracked` field counts reservations too, so it can read above the map's true length.

**9. `MAX_TRACKED_PAIRS`'s escalation note was calibrated on the wrong number.** It priced the eviction lock's hold against benign gateway load (~3.15 UPDATE/s node-wide) when the rate that matters is the one this PR's own budget *permits* — 200/s per sender with no aggregate. The doc now carries a duty-cycle table across benign, sybil, and full-flood arrival rates, and states plainly that `eviction_lock` is a blocking `std::sync::Mutex` taken from inside an async task, so at the top row it can park every tokio worker ~10% of the time. That row needs a full-scale coordinated flood to reach, so it is recorded as a hazard to track rather than rearchitected here — restructuring a sync hot-path function around `try_lock` or `spawn_blocking` is a larger and riskier change than the one under review, and a naive `try_lock` would convert contention into exactly the drops this PR exists to remove.

**10. `max_connections.saturating_mul(SENDER_TRACKING_HEADROOM)` had no floor.** `max_connections` is operator-settable; a configured `0` sized the sender map at 0, made every sender read `Untracked`, and silently disabled the fresh-id budget node-wide. A fail-open valve a config typo can pin open is not a valve. Added `MIN_TRACKED_SENDERS`.

### Deferred, with reasons

- **Re-measure the budget's calibration against live traffic.** Its headroom argument rests on measurements from 2026-07-27, and `main` has since reworked UPDATE broadcast fan-out (#5147/#5179) — which alters who sends UPDATEs to whom, i.e. the very quantity per-sender UPDATE rate is made of. I cannot take that measurement from here. It does not block the eviction fix, which is structural, but **it is worth doing before trusting the 200/s figure**, and if the margin has narrowed the refill rate should move. Flagging rather than silently assuming the old numbers hold.
- **Dashboard tile boundary tests** for the three new OR-guard terms, matching the existing per-term pattern in `home_page.rs`. Real gap, small, and no risk to the fix; not done here to keep this round reviewable.
- **The sender map's own fail-open-at-cap** is the same shape as the bug this PR fixes, one map over — `Bucket::refill` restamps on every check, so an active entry never ages out, and a full map disables the budget for everyone not already in it. Now documented in the module docs and linked to #5000 rather than re-architected here.

### On #5027

Since both touch `resync_rate_limit.rs` and neither references the other: **#4997 subsumes #5027's plumbing but not its behavioural fix.** The `BucketOutcome` enum, `check_and_record_detailed`, and the `check_and_record` delegate are near-duplicates and become redundant once this lands. But #5027's actual point — making the four resync call sites **fail open** on `Untracked` — is untouched here: this PR does not modify `op_ctx_task.rs` at all and deliberately keeps `check_and_record` collapsed so existing callers are unchanged. After this lands, the resync limiters still fail closed at a full map and #5000's resync-path starvation is still live. Recommend landing this first, then rebasing #5027 down to the call-site change alone — judged per-limiter rather than blanket, since two of the three are keyed by `ContractInstanceId`, which is attacker-choosable.

Lenses run: code-first, testing-adequacy, skeptical/adversarial, big-picture (4 independent Claude lenses; Codex not run, substitution recorded).

[AI-assisted - Claude]

Issue #4979, opened 2026-07-27
feat(telemetry): per-contract delta accounting and converged-skip counters on the broadcast rollup


sanity:
## Problem

Two numbers the bandwidth investigation keeps wanting are not measured anywhere. Both are cheap counters on the existing `broadcast_payload_mix` 60-second rollup.

These surfaced while closing #4940, which would have provided both as a side effect of a new per-fan-out event stream. That PR was closed for other reasons (its motivating question is answered by the rollup; its transaction join breaks on saturated peers). The two gaps are real and outlive it.

### 1. No per-contract *delta* accounting — a numerator with no denominator

`Window::contract_full_state_bytes` is only written under `if arm.is_full_state()` (`crates/core/src/node/network_bridge/broadcast_payload_mix.rs:435`), and `top_contracts_by_full_state_bytes` / `top_contracts_by_not_efficient_bytes` are both derived from full-state maps.

So for any given contract we can say how many full-state bytes it emitted, but not what fraction of its fan-out that was. There is no per-contract delta send count, no per-contract delta bytes, and therefore no per-contract delta:full ratio and no `bytes_saved`. A contract sending 400 MB/h of full state because it is busy and a contract sending 400 MB/h because its deltas never fire look identical in the rollup, and those want opposite fixes.

**Fix shape:** extend the existing per-contract map to carry `(full_state_bytes, delta_bytes, delta_sends, full_state_sends)` rather than a single `u64`. Same `MAX_TRACKED_CONTRACTS` cap, same top-10 truncation, same `attribution_dropped_*` reconciliation. No new event, no new cardinality key.

### 2. Converged skips and dropped sends are measured nowhere

The rollup records at delivered-send sites only. A fan-out target that is skipped as already-converged (`broadcast_queue.rs:903-912`) is counted in no arm and no residual — it is simply absent. The same is true of not-hosted early returns and dropped or timed-out streams.

That matters directly for #4956, which measured **67% of applies as no-ops carrying 80% of bytes**. The obvious question — how many targets did the fan-out already skip, versus how many redundant sends still went out — cannot be answered from telemetry today. Without it there is no way to tell whether a convergence fix is working: the sends it eliminates vanish from the numerator with nothing recording that they were eliminated on purpose.

[...]

PR #4925, opened 2026-07-25
chore: fix 26 clippy lints under Rust 1.94 --all-targets

m4nojTudu:
## Summary

- Fix all 26 latent clippy errors that fire under `cargo clippy --all-targets` with the repo's pinned 1.94 toolchain
- All lints are in test code (`#[cfg(test)]` / test binaries)
- Unblocks the CI toolchain bump from 1.93→1.94 and `--all-targets` adoption (#4818)

### Fixes by category

| Lint | Count | Fix |
|---|---|---|
| `let_underscore_must_use` | 10 | `let _ = expr` → `_ = expr` (assignment-discard) |
| `wildcard_enum_match_arm` | 9 | List variants explicitly for small enums; `#[allow]` for 20+ variant `ContractHandlerEvent` test stubs |
| `assertions_on_constants` | 3 | Wrap in `const { assert!() }` |
| `type_complexity` | 1 | Introduce `ProbeCase` type alias |
| `ptr_arg` | 1 | `#[allow]` — fn pointer signature must match generic `V` |
| `int_plus_one` | 1 | `x + 1 <= y` → `x < y` |
| `redundant_locals` | 1 | Remove redundant `let user = user` rebinding |

Closes #4914

## Test plan

- [x] `cargo clippy --all-targets --features trace,websocket,redb,wasmtime-backend,testing -- -D warnings` passes with zero errors
- [ ] CI passes (no behavioral changes — all edits are in test code)

PR #5003, opened 2026-07-28
perf(update): stop re-sending the summary to peers the broadcast already covered

sanity:
## Problem

`outbound_message_mix` (the 0.2.110 census, #4957) now has field data on 0.2.112, and **InterestSync is 52-55% of all outbound message bytes — more than the contract updates it exists to support.**

Re-derived on two non-overlapping rotated log files, 0.2.112 only:

| file | node-h | isync MB/h/node | share | B/msg | update MB/h/node | total MB/h/node |
|---|---|---|---|---|---|---|
| `…07-28T11-32` | 3,042 | 83.0 | 54.5% | 24,215 | 61.1 | 152.4 |
| `…07-28T06-22` | 2,752 | 105.6 | 52.2% | 28,143 | 88.4 | 202.4 |

#4965 filed this measurement but attributed it to the ~5-min InterestSync heartbeat. **It is not the heartbeat.** 10.9M InterestSync messages over 3,042 node-hours is 3,594 per node-hour — roughly 15x what a 300 s cadence over ~20 connections can emit.

Bucketing 60 s windows by their UPDATE message count measures the heartbeat floor directly rather than assuming a connection count. Windows with **zero** UPDATE messages are pure heartbeat:

| updates in window | windows | upd msgs/h | isync msgs/h | isync MB/h | isync per upd |
|---|---|---|---|---|---|
| 0 | 61,451 | 0 | 482 | 16.3 | — |
| 1–5 | 30,431 | 124 | 1,322 | 54.2 | 10.63 |
| 6–20 | 19,405 | 729 | 2,104 | 62.0 | 2.89 |
| 21–100 | 39,764 | 3,030 | 4,720 | 108.0 | 1.56 |
| >100 | 22,934 | 12,894 | 15,248 | 293.1 | 1.18 |

**Pearson r(update_msgs, interest_sync_msgs) per 60 s window = 0.924** (N=173,985; 0.880 on the second file). The heartbeat is the 16.3 MB/h floor, about a fifth. The other ~80% scales with the update rate.

## Approach

The emitter is `send_proactive_summary_notification`, which fires on every successful state change and fans our full summary out to **every** peer in `get_interested_peers`, excluding only the sender and self. Its only limit is a 100 ms per-contract throttle (up to 10/s/contract), which is not a bound at fleet update rates.

Most of those recipients are advertised co-hosts — i.e. exactly the broadcast fan-out targets, which `get_broadcast_targets_update` sources from `neighbor_hosting::neighbors_with_contract`. **The broadcast already carries this same summary** in `sender_summary_bytes` (`broadcast_queue.rs:1043` streaming, `:1207` non-streaming, both from `our_summary`), and the receiver upserts it into its cache of us on arrival — `drive_relay_broadcast_to` step 1 and `apply_streaming_broadcast` step 4, the #4952 upsert. Both the delta and full-state arms carry it.

So for every co-host we were paying ~24 KB to re-send, as a standalone message, a summary that peer had received moments earlier inside the broadcast.

This PR skips advertised co-hosts in the notification fan-out. **No wire-format change, no new message.**

The notification keeps doing its real job for the population it is actually needed for: interested peers we do **not** advertise-host toward, whose only other path to our summary is the heartbeat. Those two target sets are maintained by independent mechanisms (`neighbor_hosting` vs `interest_manager`), but not disjointly in the direction that matters — `record_delivery_to_interest` upserts every successfully-delivered broadcast target into `interested_peers`, so broadcast targets are a subset of notification targets.

**Recipient set** (per `.claude/rules/operations.md`, "Event emission review"): interested peers, minus the update's sender, minus self, minus every advertised co-host. The event is `NodeEvent::SendInterestMessage { target }` — targeted, one per remaining peer, never a fan-out variant.

### Failure mode considered

If a broadcast is dropped rather than delivered, the excluded co-host misses this summary — but it also missed the state itself, and the ~5-min heartbeat anti-entropy is the designated backstop for exactly that case. This mirrors `record_delivery_to_interest`, which likewise only seeds on a real delivery (#4235).

## Expected effect

The redundant share is the ratio between the two target sets. The fleet-wide 1.5 isync-per-update *message* ratio is consistent with roughly two-thirds redundancy, but that maps UPDATE messages to fan-out targets loosely, so I am **not** claiming a precise number. A per-window counter of "notification targets that were also broadcast targets" would pin it exactly and is worth adding alongside #4979. The 0.2.110 census will measure the actual delta on the next release regardless — `interest_sync_bytes` is already reported per node-minute.

## Testing

The recipient-set decision is split into a pure `proactive_summary_targets` so the emitted event's targets are asserted **directly** rather than inferred from the data layer that feeds it — the #3791 failure mode the operations rule calls out, where a data-layer-only test stays green after the dispatch site is reverted.

Six unit tests: co-host exclusion, the non-co-host population being preserved, sender/self exclusion surviving the change, the all-co-hosts collapse to zero messages, the empty set, and that exclusion is by `pub_key` not address (matching how `get_broadcast_targets_update` sources targets).

Plus a source-scrape pin, `proactive_notification_excludes_advertised_cohosts_in_production`, holding the production emitter to consulting the advertisement layer and routing through the helper. It cuts at `mod tests` (not `#[cfg(test)]`) and matches whitespace-stripped source so it survives rustfmt and cannot match itself.

**Mutation-tested, both directions:**
- removing the `neighbors_with_contract` lookup → pin fails ("MUST consult the advertisement layer")
- keeping the lookup but bypassing the helper → pin fails ("MUST route its recipient set through proactive_summary_targets")

(The first attempt at the second mutation silently failed to apply because rustfmt had reflowed the anchor; re-run against the real source. A mutation that doesn't apply looks exactly like a passing pin.)

Full `cargo test -p freenet --lib`: **4,348 passed, 0 failed**. `cargo clippy -p freenet --lib` clean; the 10 `--all-targets` clippy errors are pre-existing on the base commit (verified) in `websocket.rs`, `op_ctx_task.rs`, `contract_store.rs` — the backlog PR #4925 covers them.

## Reviewer notes

Risk tier: **Full** — this touches the broadcast/hosting surface. Review has not been run yet; this is a draft pending that.

Hosting invariants touched: **invariant 1** (freshness as a two-layer live-fan-out + anti-entropy mechanism). The change removes a redundant metadata message only; the state fan-out is untouched and still carries the summary, so no copy becomes less fresh and no read is served from a staler copy.

**Separately noted, not fixed here:** when `our_summary` is `None`, both broadcast paths send `unwrap_or_default()` — an empty `sender_summary_bytes` — which the receiver upserts as our summary, poisoning its cache of us with an empty `StateSummary`. That is pre-existing and orthogonal (the notification path returns early in the same condition, so this PR doesn't change its exposure), but it looks worth its own issue.

Refs #4965

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3612,6 +3612,9 @@ std::thread_local! {
     static GLOBAL_PENDING_OP_REMOVES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PENDING_OP_HWM: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_NEIGHBOR_HOSTING_UPDATES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    /// Recipients dropped from a proactive summary notification because they
+    /// are advertised co-hosts the broadcast already covered (#4965).
+    static GLOBAL_NOTIFICATION_COHOSTS_SKIPPED: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     // Hosting advertisement retractions emitted on stop-hosting (eviction).
     // Advertisement-layer reliability + retraction, #4642 spec step 1.
     static GLOBAL_NEIGHBOR_HOSTING_RETRACTIONS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
@@ -3702,6 +3705,7 @@ impl GlobalTestMetrics {
         GLOBAL_PENDING_OP_REMOVES.with(|c| c.set(0));
         GLOBAL_PENDING_OP_HWM.with(|c| c.set(0));
         GLOBAL_NEIGHBOR_HOSTING_UPDATES.with(|c| c.set(0));
+        GLOBAL_NOTIFICATION_COHOSTS_SKIPPED.with(|c| c.set(0));
         GLOBAL_NEIGHBOR_HOSTING_RETRACTIONS.with(|c| c.set(0));
         GLOBAL_ANTI_STARVATION_TRIGGERS.with(|c| c.set(0));
         GLOBAL_TERMINAL_CONSULT_ATTEMPTS.with(|c| c.set(0));
@@ -3859,6 +3863,26 @@ impl GlobalTestMetrics {
 
     pub fn neighbor_hosting_updates() -> u64 {
         GLOBAL_NEIGHBOR_HOSTING_UPDATES.with(|c| c.get())
+    }
+
+    /// A proactive summary notification skipped `n` advertised co-hosts,
+    /// because the broadcast to them already carried the summary in
+    /// `sender_summary_bytes` (#4965).
+    ///
+    /// Exists because the simulation-level effect of that exclusion is NOT
+    /// visible in `delta_sends` / `full_state_sends`: measured on the co-host
+    /// mesh scenario, reverting the exclusion left both metrics bit-identical
+    /// (140 / 6), so a test asserting on them alone would pass whether or not
+    /// the change was present. This counter is the one signal that actually
+    /// moves, which is what makes
+    /// `test_cohost_mesh_update_fanout_stays_delta_dominated` a test OF the
+    /// change rather than merely a test that runs alongside it.
+    pub fn record_notification_cohosts_skipped(n: u64) {
+        GLOBAL_NOTIFICATION_COHOSTS_SKIPPED.with(|c| c.set(c.get() + n));
+    }
+
+    pub fn notification_cohosts_skipped() -> u64 {
+        GLOBAL_NOTIFICATION_COHOSTS_SKIPPED.with(|c| c.get())
     }
 
     /// A hosting advertisement retraction was emitted because this node stopped
--- a/crates/core/src/node/network_bridge/outbound_message_mix.rs
+++ b/crates/core/src/node/network_bridge/outbound_message_mix.rs
@@ -440,6 +440,17 @@ struct Window {
     /// attributed — the identical side needs no diagnosis, and tracking it
     /// would double the map for no decision.
     differing_by_contract: HashMap<ContractInstanceId, u64>,
+    /// Recipients the proactive summary notification actually sent to, summed
+    /// over the window's notifications. See
+    /// [`OutboundMix::record_notification_recipients`].
+    notification_targets_sent: u64,
+    /// Recipients it SKIPPED because they are advertised co-hosts the
+    /// broadcast already covered (#4965). Paired with
+    /// [`Window::notification_targets_sent`] deliberately: the skipped count
+    /// alone cannot say whether the exclusion is doing much or nothing, since
+    /// a window with 100 skips out of 100 and one with 100 out of 10,000 are
+    /// completely different findings.
+    notification_cohosts_skipped: u64,
     /// Differing comparisons that could NOT be attributed because
     /// [`Window::differing_by_contract`] was already at
     /// [`MAX_TRACKED_CONTRACTS`].
@@ -630,6 +641,32 @@ impl OutboundMix {
         w.summary_entries_one_sided = w.summary_entries_one_sided.saturating_add(1);
     }
 
+    /// Record one proactive summary notification's recipient split (#4965).
+    ///
+    /// `sent` is how many standalone `Summaries` messages went out; `skipped`
+    /// is how many resolved interested peers were dropped because they are
+    /// advertised co-hosts the broadcast already covered.
+    ///
+    /// This is the RELEASE-VISIBLE measurement of the #4965 exclusion. The
+    /// emitter also logs the same pair at `debug!`, which
+    /// `release_max_level_info` compiles out — so in production that log
+    /// measures nothing, and a change judged on "how many sends did we avoid"
+    /// needs a counter that survives the release build.
+    ///
+    /// Lives on the outbound rollup rather than a new event stream for the
+    /// reason the module docs give: the per-node budget is 10 events/s and
+    /// this fires on every state change. Two integers per node-minute.
+    ///
+    /// Recorded even when both are zero — an early return before the fan-out
+    /// (throttled, gated, no summary) does NOT call this, so a zero pair means
+    /// "a notification ran and found nobody", which is a different fact from
+    /// "no notification ran" and the two must stay distinguishable.
+    pub(crate) fn record_notification_recipients(&self, sent: u64, skipped: u64) {
+        let mut w = self.window.lock();
+        w.notification_targets_sent = w.notification_targets_sent.saturating_add(sent);
+        w.notification_cohosts_skipped = w.notification_cohosts_skipped.saturating_add(skipped);
+    }
+
     /// Atomically take the current window, leaving a fresh empty one.
     fn take_window(&self) -> Window {
         std::mem::take(&mut *self.window.lock())
@@ -715,6 +752,20 @@ fn outbound_mix_json(w: &Window, window_secs: u64) -> serde_json::Value {
     body.insert(
         "differing_attribution_dropped".into(),
         w.differing_attribution_dropped.into(),
+    );
+
+    // #4965 co-host exclusion, measured in the release build. Emitted
+    // unconditionally as a PAIR: `skipped` alone cannot be read (100 of 100 and
+    // 100 of 10,000 are different findings), and a dropped field must not look
+    // like a quiet window. `interest_sync_summaries_notification_bytes` above
+    // shows the EFFECT; these two show the CAUSE, on the same node-minute row.
+    body.insert(
+        "notification_targets_sent".into(),
+        w.notification_targets_sent.into(),
+    );
+    body.insert(
+        "notification_cohosts_skipped".into(),
+        w.notification_cohosts_skipped.into(),
     );
     // Top differing contracts by count, so a low identical rate can be read as
     // "specific contracts are non-deterministic" vs "the design is wrong".
@@ -1090,6 +1141,55 @@ mod tests {
                 .and_then(|v| v.as_u64()),
             Some(5),
             "differing count must reach the body under its own key"
+        );
+    }
+
+    /// The #4965 notification split accumulates and reaches the body under
+    /// its own keys.
+    ///
+    /// Distinguishable counts (4 sent vs 9 skipped, and asymmetric per call)
+    /// so a swapped key or a `sent`/`skipped` transposition in
+    /// `record_notification_recipients` cannot pass. That transposition is the
+    /// realistic mistake: the two arguments are both `u64` and the pair is the
+    /// number the change is judged on, so an inverted report would claim a
+    /// saving that never happened while every other test stayed green.
+    #[test]
+    fn notification_recipient_split_reaches_the_rollup_body_under_the_right_keys() {
+        let mix = OutboundMix::new();
+        mix.record_notification_recipients(1, 2);
+        mix.record_notification_recipients(3, 7);
+        let body = outbound_mix_json(&mix.take_window(), 60);
+        assert_eq!(
+            body.get("notification_targets_sent")
+                .and_then(|v| v.as_u64()),
+            Some(4),
+            "sent count must accumulate and reach the body under its own key"
+        );
+        assert_eq!(
+            body.get("notification_cohosts_skipped")
+                .and_then(|v| v.as_u64()),
+            Some(9),
+            "skipped count must accumulate and reach the body under its own key"
+        );
+    }
+
+    /// An idle window still reports the pair, as zeros.
+    ///
+    /// "No notification ran this window" and "this build does not report the
+    /// counter" must not look the same — a dashboard reading a missing field
+    /// as zero would conclude the exclusion is doing nothing.
+    #[test]
+    fn notification_recipient_split_is_emitted_even_when_idle() {
+        let body = outbound_mix_json(&OutboundMix::new().take_window(), 60);
+        assert_eq!(
+            body.get("notification_targets_sent")
+                .and_then(|v| v.as_u64()),
+            Some(0)
+        );
+        assert_eq!(
+            body.get("notification_cohosts_skipped")
+                .and_then(|v| v.as_u64()),
+            Some(0)
         );
     }
 
--- a/crates/core/src/operations/update.rs
+++ b/crates/core/src/operations/update.rs
@@ -21,7 +21,8 @@ use crate::contract::{ContractHandlerEvent, ExecutorError, StoreResponse};
 use crate::message::{NodeEvent, Transaction};
 use crate::node::OpManager;
 use crate::ring::PeerKeyLocation;
-use std::collections::VecDeque;
+use crate::transport::TransportPublicKey;
+use std::collections::{HashSet, VecDeque};
 use std::net::SocketAddr;
 
 use dashmap::DashMap;
@@ -283,6 +284,32 @@ impl OpManager {
         let _tx = super::get::op_ctx_task::start_targeted_sub_op_get(self, instance_id, sender_pkl);
     }
 
+    /// The advertised co-hosts of a contract: peers that announced, via the
+    /// advertisement layer, that they host it.
+    ///
+    /// SINGLE SOURCE OF TRUTH for that population, deliberately shared by the
+    /// two call sites that must agree on it:
+    ///
+    /// - [`OpManager::get_broadcast_targets_update`], which broadcasts the
+    ///   state (carrying our summary in `sender_summary_bytes`) TO them, and
+    /// - [`send_proactive_summary_notification`], which EXCLUDES them from the
+    ///   standalone `Summaries` fan-out precisely because that broadcast
+    ///   already carried the summary (#4965).
+    ///
+    /// The exclusion is only sound while the excluded set is a subset of the
+    /// broadcast target set. If the two sites read different sources, or one
+    /// narrows its set and the other does not, the notification over-excludes:
+    /// a peer gets neither the broadcast nor the standalone summary, and every
+    /// unit test stays green because each site is individually correct. Sharing
+    /// the source is what makes a future narrowing apply to both.
+    ///
+    /// **If you add a filter, add it HERE, not at a call site.** That coupling
+    /// is the point of this function, and it is pinned by
+    /// `cohost_source_is_shared_between_broadcast_and_notification`.
+    pub(crate) fn advertised_cohost_pub_keys(&self, key: &ContractKey) -> Vec<TransportPublicKey> {
+        self.neighbor_hosting.neighbors_with_contract(key)
+    }
+
     /// Resolve the set of peers to broadcast a state update to.
     ///
     /// Targets are the advertised co-hosts from the proximity neighbor cache
@@ -348,7 +375,12 @@ impl OpManager {
         // advertised co-hosts — peers who announced they host this contract via
         // the advertisement layer. The former interest-manager arm (Source 2)
         // was removed in #4642 step 9; see this function's rustdoc.
-        let proximity_pub_keys = self.neighbor_hosting.neighbors_with_contract(key);
+        //
+        // Read through `advertised_cohost_pub_keys` rather than
+        // `neighbor_hosting` directly: `send_proactive_summary_notification`
+        // EXCLUDES this same set, and the shared accessor is what keeps a
+        // future narrowing from over-excluding there. See its rustdoc.
+        let proximity_pub_keys = self.advertised_cohost_pub_keys(key);
         let proximity_found = proximity_pub_keys.len();
 
         for pub_key in proximity_pub_keys {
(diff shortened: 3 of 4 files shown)


sanity:
## Full-tier review: holding this PR, two blocking findings

Five independent blind reviewers (code-first, skeptical, testing, project-invariants, convergence/summary-cache). Converting back to draft. Two blocking findings, one of which I verified myself and which invalidates the measurement this change rests on.

### Blocking 1: the 0.2.112 census cannot make the attribution I claimed

I verified this directly against the released tag:

```
$ git show v0.2.112:crates/core/src/node/network_bridge/outbound_message_mix.rs
NetMessageV1::InterestSync { .. } => OutboundKind::InterestSync,
```

On 0.2.112 that is a **single combined arm** covering every `InterestMessage` variant, including `ResyncResponse`, which carries **full contract state**. Both resync emitters (queue-full and delta-apply-failure) are themselves update-rate-correlated, so they would produce the same r≈0.92 signature I attributed to the proactive notification. The split into `InterestSyncInterests` / `InterestSyncSummaries` / `InterestSyncResync` is #4982 (`65b47c79`), which is on main but **in no tag**.

My heartbeat-floor argument was also wrong. I inferred the heartbeat share from windows containing zero UPDATE messages (482 msgs/h, ~13%), but that is a biased sample: idle windows sit on idle, less-connected nodes. Measuring it directly from `interest_heartbeat_cycle` instead:

| | |
|---|---|
| mean `peers_sent` per cycle | 50.8 (not the ~20 I assumed) |
| mean `interest_hashes` | 526 |
| implied heartbeat msgs/node-hour | ~1,220 (Interests + replies) |
| measured total InterestSync msgs/node-hour | 3,594 |
| **heartbeat share of messages** | **~34%**, not ~13% |

So what stands: InterestSync as a whole really is 52-55% of outbound message bytes, and it really does scale with the update rate. What does **not** stand: that `send_proactive_summary_notification` is the dominant emitter. That was an inference, and the instrument that settles it ships with the next release.

### Blocking 2: the source pin does not guard the fix

Found by the testing lens. This mutation passes all three needles, all six unit tests, clippy, and fmt, while completely defeating the change:

```rust
let targets = proactive_summary_targets(&resolved, &advertised_cohosts, sender_addr, self_addr);
for (_, peer_addr) in &resolved {   // was `&targets`
```

`targets.len()` is still read by the `debug!`, so there is no unused-variable warning. My own mutation testing missed this because I mutated the helper call and the co-host lookup, never the loop's iterand. Fix: a fourth needle on `for peer_addr in &targets` plus a byte-position assertion that it follows the helper call, matching the sibling idiom at `update.rs:1725-1731`.

### Major: the removed message is not pure duplication

Three reviewers converged on this independently, and I confirmed it by reading the code. A `Summaries` receiver does two things the broadcast receiver does not:

- runs the semantic staleness check (`summary_indicates_stale_peer` -> `peer_summary_has_pending_state`) and, when it holds state the sender lacks, emits a targeted `SyncStateToPeer` **back** to the sender (`node.rs:2784-2919`, capped by `MAX_STALE_SYNCS_PER_SUMMARIES`);
- records `record_summary_comparison`, the #4982 falsifier for whether a hash-first exchange is worth building.

The broadcast receive path is a bare `upsert_peer_summary` and nothing else. So every state change used to be an anti-entropy tick that could pull missing state back from a co-host. Excluding co-hosts drops that to the 300 s heartbeat. That lands on the same surface as #4961 (River room updates taking 4-20 s to propagate), which is exactly the wrong direction.

`send_summary_back_on_rejection` does not cover the gap: it is gated on the summaries being **equal**, and deliberately defers the diverged case to the heartbeat.

### Major: several paths emit no broadcast at all

The premise "the broadcast already carries the summary" fails outright where no broadcast is sent, and the notification was the only carrier:

- empty-delta converged skip (`broadcast_queue.rs:903-921`)
- contract-ban and broken-invariant egress gates (`broadcast.rs:211-249`)
- `try_notify_node_event(BroadcastStateChange)` dropped on a full channel (`executor_impl.rs:1982`), which correlates with exactly the high-update-rate load this PR targets
- `should_broadcast_contract` false (`broadcast_queue.rs:746`)
- streaming metadata-send / dispatch / timeout drops
- `upsert_peer_summary` refused at `MAX_INTERESTED_PEERS_PER_CONTRACT`

In all of these the driver still reaches the notification, because it gates only on `changed`.

### Minor

- `cohosts_skipped = advertised_cohosts.len()` logs the whole co-host set, not the number actually removed. That number *is* the size of the win, so it would overstate the fix's effect in the field. It is also `debug!`, compiled out under `release_max_level_info`.
- Source-pin off-by-one: `fn_src[1..].find(..)` indexes the `[1..]` slice but is used to slice `fn_src`, so the pinned window includes the next function's rustdoc.
- Pin needles anchor on local variable names rather than the call shape, against the guidance in `hosting-invariants.md`.
- The `impostor` half of the pub_key-vs-address test pins a state unreachable via `get_peer_by_pub_key`.
- New module-level `HashSet` import shadows the function-local one in `get_broadcast_targets_update`.
- PR body overstates: the notification fires on the two inbound-broadcast drivers only, not on "every successful state change" (a local client UPDATE never emits it).

### Disposition

Holding as draft. The bandwidth opportunity is real, but this specific change is not justified by the data I used and costs convergence responsiveness. Plan:

1. Ship the next release, which carries #4982's split arms.
2. Take one cycle of `interest_sync_summaries` vs `interest_sync_resync` data.
3. If the proactive notification really is dominant, redo this in a form that keeps the stale-peer heal, either by moving the staleness check onto the broadcast receive path or by going hash-first on the wire (which also addresses the heartbeat share, now measured at ~34%).

The empty-summary poisoning that two reviewers flagged here is genuinely independent and is going out as its own PR.

[AI-assisted - Claude]

sanity:
## The split-arm data is in, and it restores this PR's original premise

The retraction on #4965 (2026-07-28) withdrew the attribution to `send_proactive_summary_notification` on two grounds. Both are now resolved by the #4982 split arm, which shipped in 0.2.113 and is live on ~1,200 peers.

**Ground 1: `ResyncResponse` was hidden in the combined `InterestSync` arm and carries full state.** Measured on 0.2.115 (1,174 peers, 2,689 node-hours):

| arm | share of ALL outbound bytes | bytes | msgs | mean |
|---|---|---|---|---|
| **`interest_sync_summaries`** | **49.8%** | 511 GB | 12,726,443 | **40.2 KB** |
| `update` | 46.4% | 476 GB | 13,167,652 | 36.1 KB |
| `interest_sync_resync` | 0.9% | 8.8 GB | 60,123 | 145.8 KB |
| `interest_sync_interests` | 0.3% | 3.5 GB | 2,552,182 | 1.4 KB |

Resync is **0.9%**. It has the largest mean of any arm (145.8 KB, as expected for full state) but only 60,123 messages. It is not the driver, so it was not masking the attribution.

**Ground 2: the heartbeat-floor argument.** The corrected figure was that the heartbeat is ~34% of InterestSync *messages*, not 13%. That is true and it does not matter here, because **the heartbeat does not emit `Summaries` at all.** `interest_heartbeat` builds `InterestMessage::Interests { hashes }` (`crates/core/src/ring.rs:1517`), which is the 1.4 KB / 0.3% arm. The heartbeat is cheap in bytes precisely because it ships hashes.

So the 49.8% arm is not the heartbeat and not resync. The remaining emitters of `InterestMessage::Summaries` are `send_proactive_summary_notification` (`operations/update.rs:877`), the summary-back-on-rejection path (`:1024`, single-entry and gated on summaries matching, so narrow), and two receive-side reply paths (`node.rs:2700`, `:3107`).

**The message size independently corroborates it.** The notification sends `entries: vec![SummaryEntry::from_summary(hash, Some(&our_summary))]` — exactly one entry carrying one full summary. River room summaries measure ~33 KB. A single-entry `Summaries` message therefore lands right at the observed 40.2 KB mean. The heartbeat's ~526 hashes at 4 bytes each lands right at the observed 1.4 KB mean. Both arms match their construction.

## What this means for expected effect

This PR's body declines to claim a precise reduction, which was the right call then and I am not going to overclaim now either. But the target is bigger than the body assumed: this is the largest single consumer of outbound bytes on the network, ahead of contract updates themselves, and the mean notification (40.2 KB) is now **larger than the mean delta it exists to enable** (23.1 KB, from `broadcast_payload_mix` on the same fleet).

The redundancy is the overlap between `get_interested_peers` and the advertised co-host set. `interest_heartbeat_cycle` reports mean `peers_sent` = 50.8, so a single update to a busy contract can fan ~50 × 40 KB of standalone summary messages on top of the broadcast that already carried that same summary in `sender_summary_bytes`.

## Suggested framing for the merge

Worth updating the PR body before merge so it does not carry the superseded "it is not the heartbeat / 15x the cadence" arithmetic, which the author correctly retracted. The conclusion survives; that particular derivation of it does not. The clean statement is: the heartbeat emits `Interests` (1.4 KB, 0.3%), the per-update notification emits `Summaries` (40.2 KB, 49.8%), measured directly rather than inferred from a cadence estimate.

Also worth pinning as a follow-up: nothing currently counts "notification targets that were also broadcast targets", so the actual redundancy ratio stays unmeasured and the post-merge effect will only be visible as a total-bytes change in the census. The PR body already suggests this alongside #4979.

Related: #4965 (attribution, now settled), #5049 (the seeding trap — its post-apply summary ack would supply the digest that hash-first `Summaries` needs), #5050.

[AI-assisted - Claude]

sanity:
Correction to my comment above, posted in full at #4965 (comment 5131329015).

I claimed the heartbeat does not emit `Summaries`. That is misleading: `interest_heartbeat` sends `Interests { hashes }` (the cheap 1.4 KB arm, `ring.rs:1510-1517`), but the **receiver replies with a multi-entry `Summaries`** (`node.rs:2696-2700`, via `summary_if_hosted_or_in_use`). So heartbeat-driven traffic does land in the expensive arm, on the reply leg.

Net effect on this PR:

- **The resync elimination stands.** `interest_sync_resync` is 0.9%, so the specific ground for the 2026-07-28 retraction is resolved.
- **The attribution is narrowed but not settled.** `Summaries` at 49.8% has two significant emitters — this PR's per-update notification and the heartbeat reply — and the #4982 split arm cannot separate them. Your original r=0.92 / zero-update-floor evidence remains the strongest signal that the notification dominates, but it was derived pre-split.
- **Your backstop argument is strengthened.** The PR docs cite the ~5-min heartbeat as the recovery path for a co-host whose broadcast was dropped. Because the `Interests` reply carries summaries, that backstop is real rather than nominal.

So I would not treat this PR as "expected to remove ~50% of outbound bytes". It removes a message that is genuinely duplicated, which justifies itself, and the measured effect will tell us how the two emitters split. Suggest a follow-up to split `interest_sync_summaries` by emitter so that result is attributable rather than ambiguous.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #5003

### Summary
- **PR Title:** perf(update): stop re-sending the summary to peers the broadcast already covered
- **Type:** perf
- **CI Status:** passing — all heavy jobs green on `de12ef9a8` (Unit & Integration, Simulation, NAT Validation, Windows, macOS Service Unit, Clippy, Fmt)
- **Linked Issues:** #4965
- **Review tier:** **Full** (broadcast/hosting surface, update operation state machine)
- **Reviewers run:** Four independent lenses were spawned (code-first, skeptical, testing, big-picture). They did not return usable reports, so per `~/.claude/rules/multi-model-review.md` the lenses were re-run **serially in-session** against the checked-out PR code. Every finding below was verified by opening the cited file. No external (non-Claude) model was run: per the global rule as amended 2026-07-26, Codex is opt-in only and was not requested for this review.

---

### Code-First Analysis

**Independent understanding:** `send_proactive_summary_notification` fanned a standalone single-entry `InterestMessage::Summaries` to every interested peer on every state change. The PR resolves interested peers to `(TransportPublicKey, SocketAddr)`, builds an `advertised_cohosts` set, and filters them out via a new pure helper `proactive_summary_targets`.

**Stated intent:** those co-hosts are the broadcast fan-out targets and already received this summary inside the broadcast's `sender_summary_bytes`, so the standalone message was duplication.

**Alignment: matches.** Four claims verified:

1. **The two peer sets come from the same call.** `get_broadcast_targets_update` (`crates/core/src/operations/update.rs:352`) sources its only live fan-out arm from `self.neighbor_hosting.neighbors_with_contract(key)` — the identical call the PR excludes on (`update.rs` ~:930). This is the strongest possible form of the claim: the sets cannot structurally drift, because they are the same source. Source 2 (the interest-manager arm) was removed in #4642 step 9, so there is no second fan-out path that could reach a peer this exclusion misses.

2. **`sender_summary_bytes` is carried on both broadcast paths.** Streaming at `broadcast_queue.rs:1043`, non-streaming at `:1207`, both from `our_summary`. Note the streaming branch is full-state only (`_ => unreachable!("checked above")`), so deltas always take the non-streaming path; both carry the summary, so the claim holds for all four path/arm combinations.

3. **Identity handling is consistent.** The co-host exclusion keys on `TransportPublicKey` while sender/self exclusion keys on `SocketAddr`. That mirrors `get_broadcast_targets_update`, which resolves pub keys through `get_peer_by_pub_key` and then filters sender/self by address. The mixing is deliberate and matches the function it is mirroring, and the PR has a dedicated test for it.

4. **The heartbeat backstop is real, not nominal.** The PR relies on the ~5-min anti-entropy to recover a co-host whose broadcast was dropped. Verified: `interest_heartbeat` (`ring.rs:1510-1517`) sends `InterestMessage::Interests { hashes }`, and the receiver of `Interests` replies with a **multi-entry `Summaries`** built via `summary_if_hosted_or_in_use` (`node.rs:2696-2700`). So a peer that missed a broadcast is re-seeded on the next heartbeat exchange.

**Gaps:** none between intent and implementation.

---

### Testing Assessment

**Coverage level:** good at the unit level, with one real gap at the behavioural level.

| Unit | ✅ | 6 tests on the pure helper, plus a 3-assertion source-scrape pin |
| Integration | ⚠️ | none for the two-path interaction |
| Simulation | ❌ | none; arguably warranted (see below) |
| E2E | N/A | |

**Revert sensitivity** (would the test fail if the co-host exclusion were removed?):

| test | fails on revert? |
|---|---|
| `proactive_summary_targets_excludes_advertised_cohosts` | ✅ yes |
| `proactive_summary_targets_is_empty_when_every_peer_is_a_cohost` | ✅ yes |
| `proactive_summary_targets_cohost_exclusion_is_by_pub_key_not_addr` | ✅ yes |
| `proactive_summary_targets_keeps_interested_non_cohosts` | no — guards over-exclusion |
| `proactive_summary_targets_still_excludes_sender_and_self` | no — guards prior behaviour |
| `proactive_summary_targets_handles_empty_interest_set` | no — edge case |

Three of six bite directly on the change and the other three are complementary guards rather than filler. That is a healthy mix, not vacuous coverage.

**Source-scrape pin — all three claims confirmed by reading it:**
- Cuts at `"\nmod tests {"`, not `#[cfg(test)]`, so a `#[cfg(test)]`-level cut cannot disarm it ✅
- Cannot self-match: the needles live after the cut point ✅
- Whitespace-stripped, so it survives rustfmt reflowing the call ✅

**Missing test (the central gap):** nothing verifies the two paths *together*. The load-bearing behavioural claim is "the excluded peer already got this summary via the broadcast," and no test asserts that a suppressed co-host actually ends up with a correct cached summary after an update. Per `.claude/rules/test-philosophy.md`, update-propagation behaviour belongs in `run_simulation_direct`. This matters more than usual because the failure mode is **invisible to functional tests**: a wrongly-excluded peer still converges, it just converges via full-state sends. It would show up only as a bandwidth regression, and `full_no_their_summary_tracked` is already 15.1% of broadcast bytes with 89.1% tagged `never_populated`.

---

### Skeptical Findings

**Risk level: low-medium.**

| Dropped broadcast strands a co-host until the next heartbeat | Medium | `broadcast_queue.rs:1250` + NOTE `:1276` | A peer in `neighbors_with_contract` whose broadcast is dropped (queue full, or a failed stream transfer) is now excluded from both paths, so its summary is stale for up to ~5 min. The PR acknowledges this and the backstop is verified real. What the PR body understates is the **frequency**: fleet-wide stream transfer failure is currently **22.2%** (`transfer_completed` 249,534 vs `transfer_failed` 71,166), overwhelmingly `cwnd wait timeout`. So this is a regularly-exercised path, not a rare one. Still acceptable — the peer missed the *state* too, so it needs a heal regardless — but the PR body should say so with the real number. |
| Heartbeat reply can carry a `None` entry | Low | `node.rs:2661` | `summary_if_hosted_or_in_use` returns `None` for a contract we do not host or actively serve, in which case the reply entry carries no summary and does not re-seed. Narrow, and correct behaviour in itself (#4473), but it means the backstop is not unconditional. |
| Off-by-one in the pin's slice bound | Nit | pin test, `fn_end` | `fn_src[1..].find("\npub(crate) fn ")` returns an index relative to `fn_src[1..]` but is then used as `fn_src[..fn_end]`. The body slice therefore ends one byte early. Harmless (it only drops the preceding `\n`) but the two indices should share a base. |
| Empty-summary poisoning | Low (pre-existing, **not** worsened) | `broadcast_queue.rs:1043`, `:1207` | Both paths send `our_summary...unwrap_or_default()`, so when we have no summary an **empty** `sender_summary_bytes` goes out and the receiver upserts it as our summary. Verified this PR does **not** worsen it: `send_proactive_summary_notification` already returns early when there is no summary, so that path emitted nothing in this condition before the change either. It also self-corrects on the next broadcast that carries a real summary. Should be filed separately, as the PR body itself suggests. |

Edge sets (all peers are co-hosts, empty interest set, exclusion by pub key vs address) are covered by tests and behave correctly.

Checked against `.claude/rules/bug-prevention-patterns.md`: no `.send().await` in an event/recv loop introduced (the emit loop uses the existing `notify_node_event`), no protocol-enum change, no new spawn, no telemetry-counter mirroring. Nothing in that registry applies.

---

### Big Picture Assessment

**Goal alignment: yes.** Removing a genuinely duplicated message needs no bandwidth-share argument to justify it, and the change is correct on its own merits.

**Anti-patterns detected:** none. **Removed code concerns:** none — no test, comment, or safety check is weakened. **Scope:** focused; one file, one behaviour, plus tests.

**Hosting invariants:** the PR claims it touches invariant 1 (freshness as live fan-out plus anti-entropy) and preserves it. Verified. The state fan-out is untouched and still carries the summary; only a redundant metadata message is removed, and the anti-entropy layer that invariant 1 names as the second mechanism is exactly the backstop relied on.

**One documentation concern worth fixing before merge.** The new rustdoc (`update.rs` ~:855-862) permanently embeds:

> `outbound_message_mix` measured InterestSync at 52-55% of all outbound message bytes on 0.2.112, at a ~24 KB mean per message, correlating with the update rate at r=0.92.

The PR body can carry a superseded derivation; a rustdoc cannot, because it outlives the PR and will be read as current. Since that text was written, the #4982 split arm has shipped and shows `InterestMessage::Summaries` has at least two significant emitters — the per-update notification this PR changes, and the reply to a heartbeat `Interests` (`node.rs:2700`) which it does not. A rate argument puts the notification at roughly 88% and the heartbeat reply at 12%, but the **byte** split is still unmeasured (#5052 adds the counter). The current wording implies this change addresses the whole 52-55%.

Suggested replacement:

> Sending it a second time as a standalone `Summaries` message is pure duplication, and an expensive one: `interest_sync_summaries` is the largest single consumer of outbound bytes on the network (49.8% on 0.2.115, ~40 KB mean). That arm has more than one emitter — this notification and the reply to a heartbeat `Interests` (`node.rs`) — so this change addresses part of it, not all. See #4965 and #5052.

---

### Documentation
- Code docs: **complete** — the new rustdoc is unusually thorough, with the one accuracy caveat above.
- Architecture docs: n/a.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
1. **Tone down the rustdoc's bandwidth claim** (`update.rs` ~:855-862) so the permanent record does not imply this change addresses the whole InterestSync arm. Wording suggested above.
2. **Add a simulation test** covering both paths together: after an update, a suppressed co-host holds a correct cached summary, and subsequent broadcasts to it are deltas rather than full state. This is the only test that would catch the failure mode, and the failure mode is a silent bandwidth regression rather than a visible break.

#### Consider (Suggestions)
3. State the real transfer-failure rate (22.2%) in the PR body where it discusses the dropped-broadcast case, so the frequency of the fallback path is not understated.
4. File the `our_summary == None` empty-`sender_summary_bytes` poisoning as its own issue, as the PR body proposes.
5. Fix the pin's `fn_end` off-by-one so both indices share a base.
6. Land #5052 (emitter-level byte counter) before or with this, so the field effect is attributable rather than ambiguous.

---

### Verdict
**HEAD SHA reviewed:** `de12ef9a8a859afe528de68045f0c1764ab8bdfc`

The implementation is correct and well-tested at the unit level, and the central correctness question — whether the exclusion set can diverge from the broadcast target set — resolves cleanly, because both derive from the same `neighbors_with_contract` call. The two Should Fix items are a docs accuracy fix and a test addition; neither changes behaviour, so a diff-of-the-diff re-check is sufficient rather than a full re-review, provided no production code changes.

If the simulation test is deferred, that should be a deliberate decision recorded on the PR rather than an omission, given it is the only check that would catch a silent regression here.

[AI-assisted - Claude]

sanity (commented):
## Superseding review — my earlier verdict was wrong, do not merge as-is

My review above landed at "Needs Changes — Light Re-Check Sufficient" with no blocking findings. Four independent lenses have since completed and overturned that. Correcting it on the record rather than quietly revising: **the verdict is now Needs Changes — Re-review Required After Fix.**

Two things I asserted above are false, and I am withdrawing both:

- I wrote that the empty-summary poisoning is "**not** worsened" by this PR. **Wrong.** This PR removes the only corrective, on exactly the population the bug poisons.
- I treated "the broadcast already carried the summary" as established because both sets derive from `neighbors_with_contract`. Set membership is not delivery. There are pre-send exits with no notification-side counterpart.

---

### B1 (Blocking) — "pure duplication" is factually false: this removes a reverse-heal probe

`update.rs:857`. The standalone `Summaries` does **two** things at the receiver; the broadcast's `sender_summary_bytes` does one.

- Broadcast receive path: `op_ctx_task.rs:1590-1603` and `:2537-2550` do **only** `upsert_peer_summary`. No comparison, no heal.
- `Summaries` handler: `node.rs:2704-2880` upserts **and** runs `summary_indicates_stale_peer` (`ring/interest.rs:523`), and on `is_stale` emits a targeted `SyncStateToPeer` back at us (`node.rs:2895-2920`).

So the message was a per-update **reverse-heal probe**: "here is my summary, tell me if I am behind." It was the *only* per-update one in that direction — a co-host that merges our broadcast skips us in its own fan-out (`update.rs:361-368`, sender skip) and excludes us as `sender_addr` in its own notification. Post-PR that reverse path drops to the ~5-min heartbeat, so concurrent-update reconvergence between co-hosts goes from ~1 RTT to up to 5 minutes.

This stays *inside* hosting-invariant 1, which names only live fan-out and heartbeat anti-entropy as freshness layers, so it is a latency regression within the model rather than a violation. But the word "duplication" must not ship.

### B2 (Blocking) — `should_broadcast_contract` has no notification-side counterpart

`broadcast_to_single_peer` returns without sending at four points, none of them a "dropped" broadcast:

| exit | condition |
|---|---|
| `broadcast_queue.rs:738-740` | no socket addr |
| **`broadcast_queue.rs:746-753`** | **`!should_broadcast_contract`** |
| `broadcast_queue.rs:816-837` | `fanout_send_needed == false` (peer believed converged) |
| `broadcast_queue.rs:901-919` | empty delta (converged) |

`should_broadcast_contract` is a whole-contract gate: `(is_hosting_contract || contract_in_use) && contract_state_present` (`ring.rs:2249-2252`). `send_proactive_summary_notification` has **no** equivalent — it calls `get_contract_summary` unconditionally at `update.rs:891`.

So for a contract with state present but not hosted/in-use: zero broadcasts attempted to anyone, the emitter still obtains a summary, and every co-host is now excluded. That is stranding with no broadcast counterpart at all, and it is structural rather than a race.

The two converged exits are milder but directly contradict the docstring at `update.rs:867-871`: the peer did **not** miss the state, only our new summary — which is the input to *its* delta computation back toward us.

### B3 (Blocking) — a never-recovers case from a gating asymmetry

The notification is **ungated** (`update.rs:891-894`). The heartbeat reply is **gated** on the same `should_summarize_or_broadcast` predicate as the broadcast (`node.rs:3614-3625`, used at `:2661-2662`). And on a `None` report the receiver does not merely fail to seed — `node.rs:2867-2878` calls `clear_peer_summary(..., ClearedByNoneReport)`, actively **wiping** what it had.

For a contract where `should_summarize_or_broadcast == false` but state is retrievable: broadcast suppressed, heartbeat reply reports `None` and wipes, and the ungated standalone notification was the last path that could still deliver our summary. This PR removes it for co-hosts, so the peer's cache of us goes empty and stays empty. The heartbeat cannot repair it because the heartbeat is gated by the predicate that caused the problem.

Reachable population: state on disk, evicted from the hosting cache, no local client and no downstream subscriber, still receiving broadcasts because a `neighbor_contracts` retraction was dropped or has not fired — the #4440/#4473/#4610 population. Gate definitions verified; **frequency is inference, not measured.**

### B4 (Blocking unless quantified) — empty-summary poisoning is made strictly worse

Withdrawing my earlier "not worsened".

- `broadcast_queue.rs:1043` / `:1207` (and `p2p_protoc/broadcast.rs:791` / `:887`) send `unwrap_or_default()` — an empty summary — when `our_summary` is `None`.
- The receiver upserts it verbatim as `Some(empty)`, no emptiness check (`op_ctx_task.rs:1588`+`:1600`, `:2538`+`:2549`).
- The standalone notification could **never** poison: it early-returns on `None` (`update.rs:891-901`). It was therefore the corrective, overwriting the empty entry with a real summary on the next state change within the 100 ms throttle. The two `get_contract_summary` calls are separate contract-handler round-trips under `BROADCAST_CH_TIMEOUT`, so "broadcast had no summary, notification did" is reachable precisely under the load where the full-state fallback costs most.

Post-PR the corrective is gone for co-hosts. Two aggravators: `Some(empty)` is not `None`, so it clears the `_ => false` guard at `node.rs:2853` and enters the probe path, looking stale every cycle and firing heals; and the resulting waste is attributed to `Delta`/`FullNotEfficient` rather than `FullNoTheirSummary`, so it is **invisible to the very telemetry this PR is judged on**.

**This is cheap to decide.** The rate is already exported by name: `full_no_our_summary` (`broadcast_payload_mix.rs:199`). Pull it from production. If ~0, merge and state the number. If non-trivial, fix here.

### B5 (Blocking) — the net bandwidth direction is unproven and plausibly negative

One suppressed `Summaries` saves ~24-40 KB. One extra full-state broadcast costs ~355 KB (the measured `full_no_their_summary_tracked` mean), so **one avoidable full-state send cancels roughly 9-15 suppressed notifications**. A ~7% rise in that arm makes this PR net-negative on the exact metric it targets. Given B1-B4 each push in that direction, this needs a mesh simulation before merge rather than a field measurement after.

---

### Non-blocking, all verified

- **The pin's third assertion is dead.** `update.rs:2325`: needle `"forpeer_addrin&interested"`. The pre-PR loop was `for (peer_key, _interest) in &interested {` → strips to `for(peer_key,_interest)in&interested{`. It never matched the code it claims to guard. Fix to `in&interested` or drop it. (The other two assertions are sound and do catch the revert.)
- **The pin pins a weaker property than the one that must hold.** It asserts the emitter *mentions* `neighbors_with_contract`, never that the excluded set *equals* the delivered set. Add a filter to `get_broadcast_targets_update` and the exclusion silently over-excludes with every test green. Cheap fix: a shared `advertised_cohosts_for(key)` helper pinned at both call sites.
- **`body` overruns the emitter.** `fn_end` anchors on `"\npub(crate) fn "`, so `body` swallows `proactive_summary_targets`'s rustdoc (`update.rs:962-974`). A future doc edit quoting the call form could satisfy the pin with no production code behind it. Anchor on `"\npub(crate) fn proactive_summary_targets("`. (The off-by-one at `:2303` is real but harmless.)
- **`cohosts_skipped` overstates the saving.** `update.rs:956` reports `advertised_cohosts.len()` — all co-hosts, including those not interested and those that failed to resolve. Actual is `resolved.len() - targets.len()` less sender/self. This is the metric the fix is judged on.
- **No release-visible falsifier.** The only new signal is `tracing::debug!`, and `crates/core/Cargo.toml:117` sets `release_max_level_info`, so it does not exist in release builds.
- **Perturbs the #4965 falsifier mid-collection.** `record_summary_comparison` fires only inside the `Summaries` handler (`node.rs:2804`), so excluding co-hosts removes a large slice of that comparison population and changes the `summary_entries_identical`/`_differing` ratio's meaning across the release boundary. Either land after that reads out, or note the discontinuity.
- **The existing simulation guard structurally cannot catch this.** `test_sustained_update_fanout_no_full_state_storm` (`simulation_integration.rs:3190`) is a deliberate 1-gateway/2-subscriber star, so each subscriber's only co-host is the gateway, which is already excluded as sender. The exclusion is a no-op there. The emitter is reachable only from `drive_relay_broadcast_to` (`op_ctx_task.rs:1990`) and `apply_streaming_broadcast` (`:2752`), never from `drive_client_update`.
- **Throttle token and WASM summarize burned for zero targets.** `update.rs:881-901` consumes the 100 ms throttle and calls `get_contract_summary` before computing targets at `:934`. In the steady state this PR optimises for, `targets` is empty, so both are waste and the burned token blocks a genuinely-new non-co-host peer. Free fix: compute targets first, early-return on empty.

### Corrected in flight

One reviewer initially reported that the `Interests` full-replace wipes and re-registers with `summary: None` every cycle, then corrected it: `node.rs:2669-2683` takes a **refresh** branch for existing entries and preserves the cached summary. Only `remove_peer_interest` (`node.rs:2631-2639`) is destructive, and only for contracts the peer dropped. Recording the correction so it does not propagate.

---

### Recommendation

The goal is right and the premise is *mostly* right. The problem is that "already covered by the broadcast" is a structural assumption, and there are verified paths where it does not hold.

The suggested reshape is to make the exclusion **evidence-based rather than structural**: exclude a co-host only when `get_peer_summary(key, peer) == our_summary`, i.e. when we have fresh-ish evidence the peer already holds this exact summary. That naturally re-includes every peer in the B2/B3 paths, still collapses to zero messages in the steady state the PR targets, and avoids filtering on a belief that could be wrong — which is the failure shape the #4965 design note warns against for this layer.

Whatever lands needs a **relay-tree** simulation (1 gateway + 3 mutually-co-hosting peers) asserting `delta_sends > full_state_sends`, with the premise checks: assert `neighbor_hosting_updates() > 0` to confirm the peers really registered as co-hosts, and revert `update.rs:990` once to confirm the numbers move.

**Verdict: Needs Changes — Re-review Required After Fix.**
**HEAD SHA reviewed:** `de12ef9a8a859afe528de68045f0c1764ab8bdfc`

Reviewers: four independent lenses (code-first, skeptical, testing, big-picture), each blind to the others, run against the checked-out PR code. Every finding above was reported with a `file:line` its reviewer verified by opening the file. No external non-Claude model was run; per the global review rule as amended 2026-07-26, Codex is opt-in only and was not requested.

[AI-assisted - Claude]

sanity:
## Correction: the 22.2% transfer-failure rate I cited is wrong — it is ~1.66%

My review above uses "22.2% of stream transfers fail" as supporting context. That number is an artifact of my own measurement and I am withdrawing it.

**The error:** `transfer_failed` is emitted unsampled, while `transfer_completed` is sampled. I compared an unsampled numerator against a sampled denominator.

**Corrected, two independent ways:**
- Among the 65 peers with visible completions: 4,519 failed / 267,629 completed = **1.66% aggregate**, per-peer p25 0.14%, p50 1.15%, p95 6.08%, with 12 of 65 peers at zero failures.
- Among the 41 peers emitting both event types in a separate extraction: **2.1%**.

Honest limit: 94% of observed failures come from peers with no completion visibility at all, so a fleet-wide rate is not knowable from this telemetry — only that where we can see both, it is low single digits rather than 22%.

**What this changes in my review:** the dropped-broadcast path (B-series finding on `broadcast_queue.rs:1145-1165`) is exercised far less often than I implied. That *weakens* the frequency argument for the delivered-broadcast case specifically.

**What it does not change:** B2 and B3 do not depend on transfer failure at all. `should_broadcast_contract` and the `fanout_send_needed` converged-skip are deliberate no-send decisions taken before any transfer is attempted, and the gating asymmetry in B3 is structural. Those remain the blocking findings, and B5's arithmetic (one avoidable full-state send ≈ 9-15 suppressed notifications) is unaffected.

Also worth recording, since it points at a separate issue: the emit-path asymmetry itself. 636 of 686 peers emit only `transfer_failed` and never a completion. If that is sampling configuration, any dashboard comparing the two is misleading; if it is real behaviour, it is a much bigger finding. Neither this PR nor #5049 depends on the answer, but somebody should establish it.

Same correction applies to #5049, where I quoted the same figure.

[AI-assisted - Claude]

sanity:
## Branch corrected — the PR was pointing at superseded content

Force-pushed `3a4d1fd20` → `00fbf978d` (squash of the reviewed `fix/5003-review` branch onto current main).

**What was wrong:** the PR branch still carried `cohosts_skipped = resolved.len().saturating_sub(targets.len())`. That is the second of the three counter shapes this change went through, and it is the one [`bug-prevention-patterns.md`](.claude/rules/bug-prevention-patterns.md) singles out as the dangerous one: it counts the sender and self alongside the co-host drops, so it **overstates the saving**, and it stays non-zero when the filter is deleted, so it **silently defeats the simulation test whose only job is to catch that deletion**. CI was fully green on it — the green proved nothing, because every check passes on the wrong counter.

The reviewed and mutation-verified work (including `fix(update): count only cohost drops in cohosts_skipped`) lived on `fix/5003-review`, which was pushed but never had a PR of its own, so it was never on this one.

**Now at `00fbf978d`:** `cohosts_skipped` is produced by the filter itself via `ProactiveSummaryRecipients`, never re-derived at the call site; the bad shape is verifiably absent from the diff; the two-interested-co-host fixture is retained (with one, the target-narrowing mutation survives on ordering luck); and the rustdocs record *why* these tests could be vacuous rather than only that they were fixed. `cargo check -p freenet --tests` clean locally.

CI re-runs from scratch on the new SHA. Not merging until it is green on `00fbf978d`.

[AI-assisted - Claude]

Issue #3791, opened 2026-04-07
fix: summary mismatch broadcasts to all peers instead of just the stale peer


sanity:
## Problem

When a peer reports a stale summary during interest sync, the node emits `BroadcastStateChange` which broadcasts the full state to **all** subscribers/hosts of the contract (~28 peers), not just the specific peer that reported the mismatch.

With N peers exchanging summaries every 5 minutes, each mismatch triggers N×28 peer-transmissions when at most N targeted sends would suffice.

### Production evidence (gateway, 2026-04-07 hour 18 UTC)

- **4,820 UPDATE_PROPAGATION events**, of which **4,257 (88%) were triggered by summary mismatches**
- Average fan-out: 28.7 peers per broadcast
- Peak: 144 broadcasts/minute for the River room contract alone
- Total peer-transmissions: ~138,000/hour from the gateway
- Technic node: 581 MB uploaded vs 31 MB downloaded in ~1 hour (19:1 ratio)
- User report: 163:1 upload/download ratio, 1.1 GB upload in 45 minutes

### Root cause

`node.rs` line ~1418: the summary mismatch handler emits `NodeEvent::BroadcastStateChange { key, new_state }`, which triggers `handle_broadcast_state_change` → `get_broadcast_targets_update` → broadcast to ALL subscribers. The comment claims "sends state only to peers with stale summaries" but in the production path the broadcast targets all peers, not just the stale one.

## Fix

Replace the `BroadcastStateChange` emission with a targeted send directly to the peer that reported the stale summary. This changes the mismatch handler from "broadcast to all" to "unicast to the stale peer."

[AI-assisted - Claude]

PR #5004, opened 2026-07-28
fix(update): don't cache an empty summary from a sender that couldn't summarize

sanity:
## Problem

`sender_summary_bytes` is a bare `Vec<u8>` on the UPDATE broadcast wire, and every producer builds it as `our_summary…unwrap_or_default()`:

- `broadcast_queue.rs` streaming arm and inline arm
- the sim-path twins in `p2p_protoc/broadcast.rs`

So a sender whose `get_contract_summary` returned `None` (a WASM `summarize_state` error, or a contract-handler timeout at `BROADCAST_CH_TIMEOUT` under load) ships **empty bytes** rather than omitting the field, and the receiver cannot tell that apart from a genuinely empty summary.

Both broadcast receive paths then upserted it unconditionally, caching `Some(empty)` as that peer's summary. That is strictly worse than caching nothing:

- `summary_missing_reason()` reports no reason for a *populated* entry, so the cache looks healthy while being wrong
- it is therefore invisible to the `tracked_missing_*` telemetry built to find exactly this class of staleness
- our next broadcast to that peer computes a delta against a summary the peer never had

This was already documented but never fixed. `PayloadArm::FullNoOurSummary`'s rustdoc says it outright: "it also poisons the *receiver*: the wire field `sender_summary_bytes` is `unwrap_or_default()`, so the peer caches an EMPTY summary as ours and cannot tell it from a real one."

The sender side is already careful. `record_delivery_to_interest` gates its own write on `if let Some(summary)`. This closes the receiving half of the same asymmetry.

## Approach

Guard at the two **receive** sites rather than the four producers. One guard covers every producer, including peers still running older builds during a rollout, and it needs no wire change.

**Skip rather than clear.** An empty broadcast field is not a statement about what the peer holds; it is the sender failing to summarize *itself*. Discarding a good cached summary on the strength of it would throw away information for nothing. This is deliberately different from `SummaryMissingReason::ClearedByNoneReport`, which belongs to a peer that explicitly reported `None` in an InterestSync `Summaries` message, where the wire type is `Option<Vec<u8>>` and the `None` is a real claim about that peer. The new helper only ever adds information; it never removes any.

## Scope: this is correctness hygiene, not a bandwidth fix

Measured rather than assumed. The `full_no_our_summary` arm fired **109 times out of 7.09M broadcast sends** across ~3,000 node-hours of 0.2.112 telemetry, i.e. 0.0015%. Real bug, real fix, but it should not be sold as a bandwidth improvement and I am not claiming one.

## Testing

Four behavioral tests drive the guard against a **real `OpManager`** (via the existing `build_queue_full_test_node_with_clock` fixture) and assert on `get_peer_summary`, rather than testing pure logic that production could drift away from:

- empty bytes are not cached
- a real summary still is, so #4952's seeding is preserved
- an empty field does **not** evict a previously cached summary (the load-bearing "skip, don't clear" case)
- an unresolvable sender is a no-op on both the empty and non-empty branches

Plus a source pin that both receive paths route through the guard and neither re-inlines a bare `upsert_peer_summary`. A data-layer-only test would stay green after such a revert, which is the #3791 failure mode `.claude/rules/operations.md` calls out for event/dispatch changes.

**Mutation-tested both sites independently:** reverting either `drive_relay_broadcast_to` or `apply_streaming_broadcast` to the old unconditional upsert fails the pin, and the failure message names the specific function.

Full `cargo test -p freenet --lib`: **4,346 passed, 0 failed**. `cargo clippy -p freenet --lib` clean.

## Provenance

Found by the multi-reviewer pass on #5003, where two independent reviewers flagged it. It is genuinely orthogonal to that change, so it ships separately. #5003 itself is held as draft; see its review comment.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -127,9 +127,13 @@ pub(crate) enum PayloadArm {
     /// handler timed out (`BROADCAST_CH_TIMEOUT`, 10 s), errored, or answered
     /// unexpectedly. A load/contract problem, not a peer-tracking one.
     ///
-    /// Note this arm also poisons the *receiver*: the wire field
-    /// `sender_summary_bytes` is `unwrap_or_default()`, so the peer caches an
-    /// EMPTY summary as ours and cannot tell it from a real one.
+    /// This arm still ships an EMPTY `sender_summary_bytes` on the wire (the
+    /// field is `unwrap_or_default()` and cannot express "absent"), but that
+    /// no longer poisons the receiver: both broadcast receive paths route
+    /// through `update::op_ctx_task::seed_sender_summary_from_broadcast`,
+    /// which refuses to cache an empty summary. The peer therefore keeps a
+    /// truthful `None` (visible to `tracked_missing_*`) instead of a
+    /// `Some(empty)` that looked healthy while being wrong.
     FullNoOurSummary,
     /// Full state: we have our summary but none for the peer, **and the peer
     /// has no `PeerInterest` entry for this contract at all**.
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -1565,6 +1565,77 @@ async fn resend_queue_full_resync_request(
     }
 }
 
+/// Seed our cache of a broadcast sender's summary from the wire field.
+///
+/// # Why this is not an unconditional upsert
+///
+/// `sender_summary_bytes` is a bare `Vec<u8>` on the wire, and every producer
+/// builds it as `our_summary…unwrap_or_default()` (`broadcast_queue.rs` for
+/// both the streaming and inline arms, and the sim-path twins in
+/// `p2p_protoc/broadcast.rs`). A sender whose `get_contract_summary` returned
+/// `None` — a WASM `summarize_state` error, or a contract-handler timeout at
+/// `BROADCAST_CH_TIMEOUT` under load — therefore ships an EMPTY summary
+/// rather than omitting the field, and the receiver cannot distinguish that
+/// from a genuinely empty summary. That is the `PayloadArm::FullNoOurSummary`
+/// arm, whose rustdoc in `broadcast_payload_mix.rs` already records the
+/// consequence: "it also poisons the receiver … the peer caches an EMPTY
+/// summary as ours and cannot tell it from a real one."
+///
+/// Caching `Some(empty)` is strictly worse than caching nothing.
+/// [`PeerInterest::summary_missing_reason`] reports no reason for a populated
+/// entry, so the cache looks healthy while being wrong, and it is invisible to
+/// the `tracked_missing_*` telemetry that exists to find exactly this class of
+/// staleness. The sender side already takes care here —
+/// `record_delivery_to_interest` gates its own write on `if let Some(summary)`
+/// — so this closes the receiving half of the same asymmetry.
+///
+/// # Why skip rather than clear
+///
+/// An empty broadcast field is not a statement about what the peer holds; it
+/// is the sender failing to summarize *itself*. Clearing on it would discard a
+/// good cached summary for no information gain. That is deliberately different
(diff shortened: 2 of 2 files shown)


sanity:
## Full-tier review: 3 blind reviewers, no blocking findings, substantive fixes pushed

Lenses: skeptical adversarial, testing rigor, convergence/delta-efficiency + invariants. All findings addressed in `269024d7`.

### The premise was too narrow, and fixing it makes the case stronger

The skeptical lens found that empty `sender_summary_bytes` has **two** causes, not the one I documented. I verified each:

- **the sender could not summarize itself** (`unwrap_or_default()` of a `None`) — what I described
- **the contract's summary genuinely is empty** — live in this tree: `tests/test-app-1`'s container and deps return `StateSummary::from(vec![])` *unconditionally*; `website-contract` and `freenet-ping` return it for empty state

The guard is still correct, for a reason that covers both: **an empty summary is not usable evidence about the peer.** Under cause 1 it is simply false. Under cause 2 it is true but zero-information, and since the fan-out path compares cached summaries **by bytes**, two peers that both summarize to empty compare equal and `plan_fanout_send` **skips the broadcast entirely**. Matching zero-information summaries prove nothing about convergence, so that skip is wrong. Declining to cache converts a silent update blackout into a correct full-state send.

Rustdoc rewritten to state both causes and this unified justification, and to record that the `ResyncResponse` sibling site is correct as-is (its producer bails when it cannot summarize, so empty there is unambiguous) so nobody "fixes" it to match.

### A side effect I had dropped, now restored

On main the unconditional upsert also refreshed the interest entry's TTL. My guard skipped that, so an entry kept alive *solely* by empty-summary broadcasts would age out at `INTEREST_TTL`. The empty path now calls `refresh_peer_interest`: a broadcast arriving is liveness evidence even when its summary is not. It refreshes only an **existing** entry and never manufactures one. Both halves are now pinned by tests.

### The source pin's window was over-reading

Bounded with `find("\nasync fn ").or_else(|| find("\nfn "))`, which is not a `min` and does not match `pub(crate) async fn`. I replicated the slicing and confirmed the `drive_relay_broadcast_to` window ran **187 lines past** the function, swallowing `start_relay_request_update_streaming` and `start_relay_broadcast_to_streaming`. A sibling's body could have satisfied the positive needle. Now uses `extract_fn_body`, the brace-matching helper already sitting in this test module.

### Verification of the crux

The convergence lens independently traced `Some(empty)` vs `None` through `broadcast_to_single_peer` and confirmed the change is better on every axis, not a different wrong:

- **bytes**: equal in the realistic case (full state either way); strictly worse for `Some(empty)` when the contract returns a state-sized delta computed against a false base, which can then fail to apply and drive a resync
- **WASM**: strictly fewer with `None` — `Some(empty)` burns a `fanout_send_needed` probe *and* a `compute_delta`, two `GetDeltaQuery` round-trips on the serial contract-handling loop, only to fall through to full state
- **convergence**: identical in the realistic case, strictly better in the wrongful-skip case
- `Some(empty)` would also have **destroyed a previously good cached summary** on the first empty broadcast; the guard preserves it

### Testing

Eight tests, all against a real `OpManager`. Mutation-verified:

- removing the emptiness guard fails 2 behavioral tests, so these are genuine regression tests rather than pin-only coverage
- reverting **either** receive path to the old unconditional upsert fails the pin, with the message naming the specific function

`cargo test -p freenet --lib`: 4,349 passed, 0 failed. `cargo clippy -p freenet --lib` clean.

### Deferred, with reasons

- **No test drives the real drivers end-to-end.** The behavioral tests exercise the helper and the pin holds both drivers to routing through it. A driver-level test needs stream plumbing and a contract-handler fixture; worth doing, but it is coverage for the drivers generally rather than for this guard.
- **No production counter.** `debug!` is compiled out under `release_max_level_info`, so this ships without its own field signal. The arm it affects (`full_no_our_summary`) fired 109 times in ~3,000 node-hours, so a dedicated counter is hard to justify against the telemetry-volume budget in #4980.
- **Sibling sites still cache `Some(empty)`** from the InterestSync `Summaries` handler and `record_delivery_to_interest`. Those are reachable only with a genuinely-empty summary, not the failure case, and changing them is a broader behavioural call than this fix.

[AI-assisted - Claude]

Issue #4217, opened 2026-05-22
Add end-to-end #[freenet_test] for hosting-cache eviction reclaiming disk


sanity:
## Problem

Spun out of #4212 (testing reviewer). #4212 wires hosting-cache eviction to on-disk reclamation. The storage primitives, the refcount-safe removal, the eviction predicates, and the executor `reclaim_contract_storage` path all have unit tests. What is NOT covered end-to-end:

- The `RuntimePool::remove_contract` re-check guard (skip reclamation when a contract was re-hosted / re-subscribed since eviction) — building a `RuntimePool` requires a full `OpManager`, too heavy for a focused unit test.
- The `EvictContract` handler arm actually invoking reclaim on receipt.

## Ask

Add a `#[freenet_test]` that:
1. Configures a small `max-hosting-storage` budget.
2. Stores enough contracts to drive the hosting cache over budget.
3. Lets eviction run and asserts the evicted contracts' on-disk state/code is actually reclaimed.
4. Exercises the re-host race: re-access a contract in the window between eviction and `EvictContract` processing, and assert its state survives.

This is the integration-level proof that complements the unit coverage in #4212.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T15:32:54.491468798+00:00
Issue hash: `5a7628b4`

## Value Assessment

Issue #4217 is still relevant. The current code still has the exact gap described: `RuntimePool::remove_contract` contains the deletion-time guards for re-hosting, active use, and generation mismatch before calling `Executor::reclaim_contract_storage` in [runtime.rs](/home/ian/code/freenet/freenet-core/main/crates/core/src/contract/executor/runtime.rs:528), and the `EvictContract` handler still routes through `executor().remove_contract(...)` in [contract.rs](/home/ian/code/freenet/freenet-core/main/crates/core/src/contract.rs:1449). The unit-test module explicitly says end-to-end coverage of the guarded eviction path is deferred to a `#[freenet_test]` follow-up in [runtime.rs](/home/ian/code/freenet/freenet-core/main/crates/core/src/contract/executor/runtime.rs:4330). I did not find an integration test in `crates/core/tests` covering hosting-cache eviction reclaiming on-disk state/code through `EvictContract`.

[...]

PR #5005, opened 2026-07-28
fix(storage): reclaim redb free pages at startup

### LINKED ISSUE #5006: Contract database file never shrinks: 46-84% dead space on production peers
## Problem

The contract-state database file grows without bound relative to its contents and never shrinks.

redb reuses freed pages for later writes, but the file keeps its all-time high-water mark. Only *trailing* free space is truncated, and only on a clean `Database::drop` — which this binary never runs, since it exits via `std::process::exit`. The *interior* dead space a long-running peer accumulates is unreclaimable by any normal operation.

Measured on production peers (2026-07) by opening the real database files:

| Peer | File | Pages in use | Dead space |
|---|---|---|---|
| laptop peer | 2.68 GB | 255 MB of live rows across 747 contracts | **84.2%** |
| nova gateway | 2.65 GB | 1.38 GB | **46.7%** |

On the laptop this made Freenet's entire on-disk footprint 2.68 GB (3.92 GB on a filesystem without transparent compression) for roughly 300 MB of genuinely needed data.

## Why it is invisible

`DiskUsageTracker` sums logical state bytes and explicitly documents that database overhead is not counted, so neither the disk budget (#4683) nor the pre-write admission gate (#4702) nor the telemetry can see it. The laptop's dashboard reported 583 MB while the real footprint was 2.68 GB — a ~10x under-count, meaning both the budget and the `min(ram, disk)` eviction floor were deciding on a badly wrong number.

## Fix

Compact at startup, gated so it only runs when it is worth the rewrite. Verified: 2.65 GB -> 1.44 GB on nova's real database, after which the gate correctly declines and stays declined.

Follow-up (separate): feed the real file length into `DiskUsageTracker` so the budget stops under-counting between restarts.

[AI-assisted - Claude]

sanity:
## Problem

redb never returns freed pages to the OS. Freed pages get reused by later writes, but the file keeps its all-time high-water mark forever. Only **trailing** free space is truncated on commit; the interior dead space a long-running peer actually accumulates is unreclaimable by any normal operation.

Measured on production peers (2026-07) by opening the real database files:

| Peer | File | Live rows | Dead space |
|---|---|---|---|
| laptop peer | 2.68 GB | 255 MB across 747 contracts | **84.2%** |
| nova gateway | 2.59 GB | 766 MB across 2030 contracts | **51.7%** |

On the laptop that made Freenet's whole on-disk footprint 2.68 GB (3.92 GB on a filesystem without transparent compression) for ~300 MB of genuinely needed data.

The node cannot see this itself. `disk_usage.rs` counts logical state bytes and explicitly documents that "ReDb/SQLite database overhead are additional and not counted against it, so this is not a hard bound on total on-disk usage" — so both the disk budget and its telemetry under-report the real footprint several-fold. The laptop's dashboard reported 583 MB while the real footprint was 2.68 GB.

## Solution

Call `Database::compact()` at open. Startup is the one point where no transaction is live (`compact` returns `TransactionInProgress` otherwise) and the store is not serving anything yet.

Two gates, both of which must pass, so a node in a restart loop never rewrites a multi-GB file for a trivial gain:
- reclaimable >= 64 MiB (absolute floor)
- reclaimable >= 25% of the file (proportional)

Waste is measured from `DatabaseStats::allocated_pages() * page_size()` against the file length, which is cheap.

Compaction never fails startup. On error the database is left valid but uncompacted, which is exactly the current behaviour. Both gates live in a pure `compaction_is_worthwhile` so the boundary math is unit-testable, following the same split as `budget_for_ram` / `disk_budget_for_clamped`.

Measured effect: **2.68 GB -> 608 MB in 0.5s** on the laptop's database, **2.59 GB -> 1.28 GB in 3.2s** on nova's. Every row intact, and a second pass is a no-op (`compact()` returns `false`).

### Risk

`compact()` is not resumable: killing the process mid-compaction forces a repair on next open (redb uses two-phase commit, so no data is lost). The window is the 0.5-3.2s measured above, and both sides are logged so an interrupted run is diagnosable.

## Testing

- `startup_compaction_reclaims_dead_pages_and_preserves_rows` — the regression test. Verified red without the fix (`404754432 -> 404754432`, no change) and green with it, and it asserts every live row survives.
  - The fixture deletes a **scattered** subset rather than a contiguous tail. This matters: redb truncates trailing free space on commit, so a tail-delete fixture would shrink on its own and prove nothing. Interior dead space is the real production shape.
- `startup_compaction_skips_database_without_meaningful_dead_space` — a healthy database is byte-for-byte untouched, so the gate genuinely prevents pointless rewrites.
- `compaction_gate_requires_both_floor_and_fraction` — boundary math: floor-only, fraction-only, both, exact boundaries either side, and degenerate inputs (zero-length file, `in_use > file`) that must not divide by zero or panic.

### Why CI didn't catch this

There was no test asserting anything about the database file's size, only about its contents. #4217 ("Add end-to-end test for hosting-cache eviction reclaiming disk") is open and tracks the adjacent gap. These tests close the specific hole for the store's own file.

[AI-assisted - Claude]

## Fixes

Closes #5006

Follow-up filed as #5007 (disk accounting under-counts the real footprint ~10x; out of scope here).

### DIFF
--- a/crates/core/src/contract/storages/redb.rs
+++ b/crates/core/src/contract/storages/redb.rs
@@ -409,8 +409,8 @@ impl ReDb {
         );
 
         match Database::create(&db_path) {
-            Ok(mut db) => {
-                Self::reclaim_free_pages(&mut db, &db_path);
+            Ok(db) => {
+                let db = Self::reclaim_free_pages(db, &db_path)?;
                 Self::initialize_database(db)
             }
             Err(e) if Self::is_version_mismatch(&e) => {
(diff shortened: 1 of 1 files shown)


sanity:
## Review round 1 — consolidated

Four independent blind lenses (code-first, testing, skeptical/adversarial, big-picture/operational). External non-Claude models were not used, per the 2026-07-26 change to the multi-model-review rule; four diverse Claude lenses were substituted.

Every claimed defect was checked against **direct measurement on the real production databases** rather than accepted or dismissed on reasoning. That mattered: three of the most serious claims did not survive measurement, and three did.

### Fixed

| Finding | Severity | Status |
|---|---|---|
| A compaction I/O error made the node **fail to start** | major | fixed |
| Trailing growth slack counted as dead space → false-positive compaction on every unclean restart | major | fixed |
| All skip-path diagnostics compiled out of release builds | major | fixed |
| `stats()` read even when the file is below the absolute floor | minor | fixed |
| Tests: row-count-only integrity, loose assertion, non-isolating gate boundaries, no restart-idempotence | major | fixed |

The startup-failure one was the most serious and was not in the original design's threat model. redb latches an `io_failed` flag on any backend error, and every subsequent `begin_write` returns `PreviousIo`. The original code logged a warning and carried on, handing that poisoned handle to `initialize_database`, whose `begin_write` then returned `Err` out of `ReDb::new`. A survivable compaction failure became a node that will not boot — a regression against current behaviour, most plausibly triggered on a COW filesystem where a full-file rewrite can hit ENOSPC. The handle is now reopened on that path.

The slack finding is measured, not theoretical: a healthy database is **640 MiB with the handle open and 386 MiB after a clean drop**. redb only truncates trailing free space in `Database::drop`, and this binary exits via `std::process::exit`, so that destructor never runs in production. An unclean exit therefore presents ~40% slack with zero fragmentation, and the original gate fired on it. Compaction now runs cheapest-first: skip below the floor without measuring, then measure, then close/reopen to trim slack for free and re-measure before committing to a rewrite.

### Claimed but disproven by measurement

Recording these so they are not re-litigated.

- **"The gate re-fires forever, rewriting the database on every start."** On nova's real database: 2652 MiB → 1440 MiB, after which reclaimable is 1.9% and the gate declines, stable across repeat runs. The supporting arithmetic assumed `allocated_pages` survives compaction unchanged; it does not.
- **"`stats()` is a full btree walk costing hundreds of MB of reads at every start."** Measured at **0.1 ms** on a 386 MiB database.
- **"A healthy database gets compacted."** After a clean drop, 0.5% reclaimable; the gate declines. (The *unclean* variant of this claim was correct and is fixed above.)

### Checked and safe, with evidence

- **No 2x disk requirement.** `compact_pages` relocates into already-allocated low slots via `allocate_lowest` and then truncates; there is no temp copy.
- **No data loss on kill mid-compaction.** Two-phase commits throughout, fsync either side of the slot swap, truncation only after the shrunk layout is durable. redb's open path recalculates layout when the file is longer than recorded. A kill degrades to a repair, not corruption.
- **Cross-process safety.** redb takes an exclusive `flock` at open; a second node or the secrets CLI fails cleanly rather than racing.
- **No hosting-invariant interaction.** This changes only the physical layout of already-retained rows. No durable copy added, no change to eviction ordering or its inputs.

### Deferred, with reasons

- **`spawn_blocking`.** `ReDb::new` is `async` but synchronous throughout, and `Database::create` already blocks. Worth doing, but it is a pre-existing property of this constructor rather than something this PR introduces, and doing it properly means touching the repair path too.
- **Feeding the real file length into `DiskUsageTracker`.** This is the deeper defect the PR describes: the node's disk accounting counts logical state bytes and excludes database overhead, which was a ~10x under-count on the measured laptop, so the #4683/#4702 budget and eviction floor were deciding on a badly wrong number. Out of scope here (one logical change), but it should be filed — this PR shrinks the gap at startup and does nothing at runtime.
- **Operator kill switch.** Low risk given the change is gated, non-fatal and measured in seconds, and project rules discourage speculative config.

[AI-assisted - Claude]

sanity:
## Review round 2 — re-review of the restructured code

A changed diff is an unreviewed diff, so the fixes from round 1 got a fresh adversarial pass. It found a real defect in my own fix, and mutation testing confirmed it.

### The skip tests were vacuous

Round 1 correctly told me to stop asserting row counts. In fixing that I weakened the size assertions to `after <= before`, which passes whether or not compaction ran. Mutation testing made it unambiguous:

```
gate guards disabled (always compact)  ->  all three tests PASS
```

The gate *constants* were pinned by the unit test, but the gate's **wiring into `reclaim_free_pages`** was pinned by nothing. Asserting on file size cannot work here: compacting a healthy database barely changes its length, and re-compacting an already-compacted one leaves it byte-identical.

Fixed by making the decision observable. `reclaim_free_pages` now returns a `ReclaimOutcome` (`BelowFloor` / `NotWorthwhile` / `TrimSufficed` / `Compacted` / `Undetermined`), recorded via a test-only thread-local. Verified in **both** directions:

```
gate forced off (always compact)  ->  both skip tests FAIL
gate forced on  (never compact)   ->  the reclaims test FAILS
```

### Also fixed

- **Stage-3 reopen could stop the node booting.** It dropped the only working handle and propagated any failure straight out of `ReDb::new`, so losing a microseconds-wide race for the file lock was fatal. Now retries with backoff, and logs a phase before giving up.
- **Stat failure after trimming fell back to the pre-trim size**, overstating the file and biasing toward the expensive rewrite. Every other error arm skips; this one now does too.
- **Stage-1 below-floor skip logged nothing**, the same gap round 1 closed for the stage-2 skip. Now `info!`, and the start/done pair no longer reports two different "before" values under two names.

### Accepted and documented rather than fixed

The reviewer correctly noted the compaction-failure reopen **does not rescue a persistent cause**: with a full disk, the reopen's own repair commit fails too and the node still will not start. That is now stated plainly at the call site rather than implied away. It converts the transient case from fatal to survivable and does nothing more. There is also a comment explaining why this path deliberately does not route through `abort_process_on_redb_poison` like the store's other I/O errors — aborting at startup would just loop the supervisor.

### Verified safe, with reasons

- `io_failed` is a per-instance `AtomicBool` on `PagedCachedFile`, not persisted, so a fresh `Database::create` genuinely clears it. The reopen does what it claims.
- The flock is released synchronously by `drop(db)` (handle owned by value, no live transactions), so the reopen cannot self-deadlock.
- `pages_in_use_bytes` is correct: `WriteTransaction::new` performs no disk writes, `abort()` consumes the transaction and releases the write slot even on error, and all three `None` arms skip — a bad measurement can never cause a spurious rewrite.
- The stage-1 floor short-circuit is sound: `reclaimable <= file`, so `file < FLOOR` implies the floor gate fails.

### Known and not addressed

redb's `compact()` contains internal `assert!`s that would panic rather than return `Err`. They guard redb invariants that hold given an exclusive `&mut Database` with no live transactions, so I don't believe they're reachable, but the function's `Result` contract doesn't cover them and I'd rather say so than not.

[AI-assisted - Claude]

PR #5010, opened 2026-07-29
fix(tracing): never prune a log file an appender has open

sanity:
## Problem

The log pruner could delete files the tracing appenders currently have open. On Linux that leaves the appender writing to an unlinked inode: the disk space is not reclaimed, and everything logged for the rest of the hour is invisible to `freenet service report`. On Windows the deletion simply fails.

Three distinct ways in:

**1. The size pass spared only one file.** It popped the newest file by mtime as "the one the appender has open". There are **two** appenders over the same directory, and their open files do not advance in lockstep — whenever the main log has been written more recently, the live error log is an ordinary deletion candidate.

**2. The age pass spared nothing at all.** It applied a plain `modified < cutoff` test with no live check. This is reachable because **rotation is lazy**: an appender that is not written to never rotates, so its open file's mtime stays frozen at creation. A node that goes a full retention window without a warning therefore has a *live* error log that mtime alone cannot distinguish from an abandoned one.

**3. `max_log_files` pruned across both families and into the systemd-held files.** `tracing-appender`'s `prune_old_logs` selects victims with a bare `filename.starts_with(prefix)` test (verified against `tracing-appender-0.2.5/src/rolling.rs:655-724`). The main appender's prefix is `freenet`, which also matches `freenet.error.*` **and** the systemd/launchd-held bare `freenet.log` — the very file the name filter exists to protect. So the main appender's budget of 72 was spent across *both* families, halving each one's real retention to ~36h, and once over budget it deleted oldest-first into the legacy files. Whether a given file survived as far as the sort depended on `metadata().created()` support, making the behaviour filesystem-dependent.

Additionally, the prune sort was **not a total order**: `sort_by_key(|f| f.modified)` is stable, so an mtime tie *within a family* fell back to `read_dir` order (arbitrary on ext4). Ties are reachable at the rotation instant on a coarse-granularity filesystem, where the outgoing file's last write and the incoming file's creation land in the same tick — and the live-file pick then takes the closed sibling, leaving the genuinely-open file collectable.

## Solution

One pruner, one live set, honoured by both passes.

- `prune_log_files` computes the live set **once** over the full file list, before either pass runs, and both the age and size passes skip it.
- `live_file_indices` derives the live set **from the data**, not by iterating a hand-written list of `LogFamily` variants. A third appender would otherwise need someone to remember to extend that list, and forgetting silently re-exposes its open file — the exact bug being fixed.
- The sort is now `(modified, path)`. These names embed a zero-padded `YYYY-MM-DD-HH` stamp under a fixed per-family prefix, so within a family lexicographic order *is* chronological order. Uses `sort_by` rather than a `sort_by_key` returning a tuple, to avoid cloning a `PathBuf` per comparison.
- `LogFile` carries its family, classified once at read time. `rotating_log_family` documents why the filter must not be loosened to a bare `freenet` prefix: on Linux the auto-updater's state dir **is** the log dir, so `known_good_binary` and the probation marker live alongside the logs.
- `max_log_files` is dropped from both appenders, making `cleanup_old_logs` the single pruning authority. A source pin stops it being re-added.

### The live-file exemption has no TTL, deliberately

AGENTS.md requires GC exemptions to expire via TTL or be overridden by an absolute age threshold. This one has neither, and the argument is made in the rustdoc on `prune_log_files` rather than left for the next reader to reconstruct.

The obvious backstop — "delete it anyway past N days" — is not available: the appender still holds the descriptor, so unlinking reclaims no space until it rotates anyway, and the node loses the rest of that hour's logs. It would reintroduce exactly the bug this PR fixes.

The exemption is instead bounded by **rotation** rather than by time, and cannot become the growing blind spot the rule protects against:

- **Bounded in cardinality.** `live_file_indices` yields at most one index per distinct `LogFamily`, so the exempt set is at most two files, structurally. It cannot grow with uptime, file count, or log volume.
- **Bounded in bytes.** An exempt file cannot grow once its rotation period passes: `RollingFileAppender::write` calls `should_rollover` and swaps in a fresh file *before* writing the buffer (`tracing-appender-0.2.5/src/rolling.rs:227-236`), so the first write after the boundary lands in a new file, never the stale one. Each exempt file is frozen at what its family wrote during a single rotation period — ~21 MB/hour across both families on the measured gateway, against a 512 MiB budget.
- **Self-clearing.** The only event that could make an exempt file grow is a write, and a write is exactly what rotates it away: the exemption transfers to the new file and the superseded one becomes an ordinary candidate that the very next prune sweeps.

Unbounded in *time*, bounded in *bytes* and *count*. Pinned by `live_file_exemption_clears_once_the_appender_rotates`, which drives two prunes across a simulated rotation and asserts the once-exempt file is collected on the second.

Two limits of that argument are stated in the code rather than glossed: `live_file_indices` *infers* which file is open (there is no portable way to ask) and a backward clock step mid-hour can defeat the proxy; and the bytes leg rests on an **external** property that no test here pins, so `Cargo.toml` carries a comment on the `tracing-appender` line prompting a re-read of `rolling.rs` on any bump.

### No change to the retention defaults

`LOG_DIR_MAX_BYTES` stays 512 MiB and `LOG_RETENTION_HOURS` stays 72. This branch originally tried to cut the budget; the measurement said it could not be cut.

Measured on the production gateway (nova): **54 rotating files, 518.9 MiB, spanning 25.97 hours = 20.95 MB/hour** — about 20x a quiet laptop peer. 512 MiB buys ~25.6 hours there; 96 MiB would buy **4.8 hours**, so an 18:00 incident would be gone by 09:00. `default_budget_holds_a_day_of_a_busy_gateways_logs` now encodes that rate and asserts >= 24h, so re-sizing from a quiet peer fails CI. Raising `LOG_RETENTION_HOURS` was equally wrong in the mirror direction: quiet peers are bounded by age, not bytes, so it would have *increased* their disk.

Reducing the budget is blocked on **#5015**: `freenet.error.*` is currently a byte-for-byte duplicate of the main log, because `EnvFilter`'s `with_default_directive` is a fallback rather than a floor and any `RUST_LOG` value discards the error layer's `WARN` default (`tracing-subscriber-0.3.23/src/filter/env/builder.rs:334-338`). Half of that 20.95 MB/h is redundant. Fixing it halves the rate and lets the budget come down to ~256 MiB for the same history — the single largest lever on log disk, and a separate logical change (what gets *written*, not what gets *deleted*).

## Testing

Every behavioural change is mutation-verified — mutation applied, target test observed red, then reverted; tree confirmed clean and the suite re-run green afterwards.

| Test | Mutation | Result |
|---|---|---|
| `live_file_exemption_clears_once_the_appender_rotates` | exemption made family-wide (sticky) | FAILED |
| `time_pass_spares_a_live_file_whose_mtime_has_aged_out` | drop the live guard from the age pass | FAILED |
| `size_cap_spares_the_live_file_of_both_appenders` | spare only `LogFamily::Main` | FAILED |
| `prune_log_files_sorts_before_choosing_victims` | delete the sort entirely | FAILED |
| `live_pick_breaks_mtime_ties_by_name` | revert to `sort_by_key(\|f\| f.modified)` — the exact pre-fix line | FAILED |
| `default_budget_holds_a_day_of_a_busy_gateways_logs` | re-size budget to 96 MiB | FAILED |
| `live_file_indices_is_the_newest_of_every_family_present` | keep oldest per family instead of newest | FAILED |
| `error_logs_classify_as_their_own_family` | swap the prefix tests so `freenet.` is checked first | FAILED |
| `appenders_must_not_delegate_pruning_to_max_log_files` | re-add `.max_log_files(..)` to the main appender | FAILED |

The age-pass and rotation tests each keep a genuinely superseded sibling in the same family, so neither exemption can silently become "never delete error logs". The sort test passes a deliberately shuffled vector — before it, nothing fed `prune_log_files` out-of-order input, so deleting the sort passed or failed depending on how the filesystem enumerated a temp directory. The source pin splits its needle with `concat!` so it cannot match its own text, and the re-add mutation confirms it fires on a real builder call rather than on a comment.

`cargo fmt --check`, `cargo clippy -p freenet --lib --all-features`, and all 18 `tracing::tracer` tests are clean.

## Related

- #5015 — `freenet.error.*` duplicates the main log. Prerequisite for reducing the byte budget.
- #5019 — log-pruner follow-ups (hand-maintained prefix chain, uncounted delete failures, tiny-budget footgun).
- #5021 — make the byte budget configurable without recompiling. Split out of this PR as unrelated new configurability.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,12 @@ opentelemetry-jaeger = "0.22"
 opentelemetry-otlp = "0.32.0"
 opentelemetry_sdk = { version = "0.32", features = ["rt-tokio"] }
 tracing = "0.1"
+# Bumping this? Re-read `rolling.rs`'s `impl io::Write for
+# RollingFileAppender` first. The log pruner's GC-exemption argument
+# (crates/core/src/tracing/tracer.rs::prune_log_files) depends on the
+# rollover check running BEFORE the buffer is written — audited at 0.2.5.
+# If that order ever flips, the exemption stops being bounded in bytes
+# and nothing in our test suite will notice.
 tracing-appender = "0.2"
 tracing-core = "0.1"
 tracing-opentelemetry = "0.33.0"
--- a/crates/core/src/tracing/tracer.rs
+++ b/crates/core/src/tracing/tracer.rs
@@ -242,14 +242,25 @@ fn cleanup_old_logs(log_dir: &std::path::Path) {
     prune_log_files(files, cutoff, log_dir_max_bytes());
 }
 
-/// The indices in a **mtime-ascending** `files` of the newest file of
-/// each family — i.e. the file each appender currently has open.
+/// The indices, in a `files` sorted ascending by `(modified, path)`, of
+/// the newest file of each family.
 ///
 /// Discovered from the data rather than by walking a hand-written list
 /// of [`LogFamily`] variants: adding a third appender must not require
 /// remembering to extend a list here, because forgetting would silently
 /// re-expose that appender's open file to deletion — precisely the bug
 /// this function exists to prevent.
+///
+/// **This INFERS which file each appender has open; it does not know.**
+/// There is no portable way to ask whether some other part of the
+/// process holds a descriptor, so "newest of its family" is a proxy. It
+/// is right whenever time moves forward, which is why the caller sorts
+/// by `(modified, path)` — see [`prune_log_files`]. A backward clock step
+/// mid-hour can defeat it: the appender does not roll (rotation is keyed
+/// on the clock too), so its open file's mtime can end up older than a
+/// closed sibling's, and the proxy picks the wrong file. The consequence
+/// is bounded — it is the same exposure this code removed in the common
+/// case, not a new one — and no cheaper signal is available.
 fn live_file_indices(files: &[LogFile]) -> Vec<usize> {
     let mut newest: Vec<(LogFamily, usize)> = Vec::new();
     for (idx, file) in files.iter().enumerate() {
@@ -303,6 +314,16 @@ fn live_file_indices(files: &[LogFile]) -> Vec<usize> {
 ///   one. Each exempt file is therefore frozen at whatever its family
 ///   wrote during a single rotation period — on the measured gateway,
 ///   ~21 MB/hour across both families, against a 512 MiB budget.
+///
+///   **This leg rests on an EXTERNAL property that no test here pins.**
+///   Nothing in this crate would fail if a `tracing-appender` bump moved
+///   the rollover check after the write; the byte bound would silently
+///   become false while all these tests stayed green. It is audited at
+///   0.2.5, and `Cargo.toml` carries a pointer back here so a bump is
+///   prompted to re-read `rolling.rs`. A local test cannot cover it:
+///   forcing a rotation needs control of the appender's clock, which
+///   `tracing-appender` exposes only to its own `cfg(test)` builds.
+///   If you bump the dependency, re-check that ordering by hand.
 /// * **Self-clearing.** The only event that could make an exempt file
 ///   grow is a write, and a write is exactly what rotates it away: the
 ///   exemption transfers to the newly-created file and the superseded one
@@ -320,7 +341,24 @@ fn live_file_indices(files: &[LogFile]) -> Vec<usize> {
 /// It would reintroduce precisely the bug this function exists to fix.
 /// Do not add one.
 fn prune_log_files(mut files: Vec<LogFile>, cutoff: std::time::SystemTime, max_bytes: u64) {
-    files.sort_by_key(|file| file.modified);
+    // Sorting is load-bearing, not cosmetic: `live_file_indices` reads
+    // the live file off the END of each family's run, and the size pass
+    // walks this order to delete oldest-first. `cleanup_old_logs` feeds
+    // us `read_dir` order, which is arbitrary (hash order on ext4).
+    //
+    // The path is a tiebreak rather than mtime alone, because mtime is
+    // NOT a total order here: on a coarse-granularity filesystem the old
+    // file's final write and the new file's creation can land in the same
+    // tick, and a stable sort would then fall back to `read_dir` order —
+    // leaving the genuinely-open file looking like the older of the two
+    // and thus collectable. These file names embed a zero-padded
+    // `YYYY-MM-DD-HH` stamp under a fixed per-family prefix, so within a
+    // family lexicographic order IS chronological order.
+    files.sort_by(|a, b| {
+        a.modified
+            .cmp(&b.modified)
+            .then_with(|| a.path.cmp(&b.path))
+    });
 
     // Computed once, on the full set, and honoured by both passes below.
     let live = live_file_indices(&files);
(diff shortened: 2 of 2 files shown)


sanity:
Findings 1 and 4 addressed in 79381ab9, plus both flagged documentation gaps.

**Finding 1 — sort is now a total order.** `sort_by(|a, b| a.modified.cmp(&b.modified).then_with(|| a.path.cmp(&b.path)))`. Used `sort_by` rather than `sort_by_key` returning a tuple, to avoid cloning a `PathBuf` per comparison.

**Finding 4 — the sort is now pinned.** `prune_log_files_sorts_before_choosing_victims` passes a deliberately shuffled vector (newest first, oldest in the middle) and asserts the newest of each family survives. Deleting the sort line is now deterministically red rather than filesystem-dependent.

Separately, `live_pick_breaks_mtime_ties_by_name` covers the tie itself: two same-family files with identical mtimes, ordered so that mtime-only sorting mistakes the older for the live one.

**Mutation evidence:**

| Test | Mutation | Result |
|---|---|---|
| `prune_log_files_sorts_before_choosing_victims` | delete the sort entirely | FAILED |
| `live_pick_breaks_mtime_ties_by_name` | revert to `sort_by_key(\|file\| file.modified)` — the exact pre-fix line | FAILED |

Worth noting the first attempt at the sort-removal mutation **silently failed to apply** — I used `s\|...\|...\|` in perl and the `\|` delimiter collided with Rust's closure syntax, so the test passed against unmutated code and looked like a clean pin. Caught it on a `grep -c` sanity check. The mutation harness now aborts if `git diff` shows the file unchanged, so a no-op mutation can't masquerade as a passing test.

**Both flagged items taken:**

- `live_file_indices` no longer states as fact that it knows which file is open. It now says it *infers* it, names the case that defeats the proxy (a backward clock step mid-hour: the appender doesn't roll, so its open file's mtime can fall behind a closed sibling's), and notes the consequence is bounded and no cheaper signal exists.
- The `bounded in bytes` leg now says outright that it rests on an external property no test here pins, that a `tracing-appender` bump moving rollover after the write would silently invalidate it with all tests green, and that a local test can't cover it (forcing rotation needs the appender's clock, exposed only to its own `cfg(test)` builds). `Cargo.toml` carries a comment on the `tracing-appender` line pointing back to re-read `rolling.rs` on a bump — that's where a bump actually happens, so it's more likely to be seen than a comment in the tracer.

The three remaining items are filed together as #5019 (hand-maintained prefix chain, age-pass delete failures not counted in the byte accounting, `FREENET_LOG_DIR_MAX_BYTES=1` being as destructive as `0`).

19/19 `tracing::tracer` tests green, `cargo fmt --check` and `cargo clippy -p freenet --lib --all-features` clean at 79381ab9.

[AI-assisted - Claude]

sanity:
`FREENET_LOG_DIR_MAX_BYTES` removed in ce0de260 — the constant, `log_dir_max_bytes`, `parse_log_dir_max_bytes`, its test, and all doc text. `cleanup_old_logs` reads `LOG_DIR_MAX_BYTES` directly again. Verified no references remain anywhere in the tree.

Agreed on the scope call. The override was load-bearing while the plan was to *lower* the default — a gateway would have needed a way to claw back the history a smaller default took away. Once nova's 20.95 MB/hour showed the default could not come down at all, it stopped being part of what made this change safe and became unrelated configurability in a `fix:` PR.

Filed separately as **#5021** rather than folding into #5019: it is new configurability rather than a code-quality item, so it needs maintainer sign-off on its own issue per CONTRIBUTING.md. The issue carries the design notes from the prototype (degrade-safe fallback, pure parse helper so tests don't mutate process-global env state, and the low-end guard), and cross-references item 3 of #5019 since both concern the same tiny-budget footgun — whichever lands second should reuse the other's floor rather than inventing a second one.

One caveat kept, reworded: the "do not lower this so far that the current-hour files alone can fill it" note on the constant. Without the override it is guidance for whoever edits the constant rather than for an operator, but it is still the reason the budget has a floor worth respecting.

This PR is now purely: spare open log files in both prune passes, and make the prune sort a total order.

18/18 `tracing::tracer` tests green, `cargo fmt --check` and `cargo clippy -p freenet --lib --all-features` clean at ce0de260.

[AI-assisted - Claude]

Issue #3945, opened 2026-04-22
Web contract subresource fetch widens DoS amplification surface


sanity:
## Context

Flagged during review of PR #3942. That PR fixed #3940 by making \`variable_content\` trigger a GET + unpack when the webapp cache is cold. A necessary fix, but it widens the attack surface.

## Concern

Before #3942, an unauthenticated HTTP client requesting \`/v1/contract/web/<KEY>/anything\` for a random \`<KEY>\` got an immediate 404 from the local cache check. After #3942, that same request triggers a full network GET (fanning out to remote peers) plus an unpack attempt.

An attacker can now enumerate random 32-byte keys and force the local node to issue outbound GETs it would otherwise never issue. The per-key rate is bounded by the 30s fetch timeout, but parallel fan-out is not bounded.

The analogous attack via \`contract_home\` has always existed, but top-level page navigations are typically human-paced; subresource URLs are machine-fetchable and easier to spray.

## Possible mitigations

1. Rate-limit by source IP for \`variable_content\` cold-cache paths.
2. Only trigger the cold-cache fetch when the contract instance is already known to the local contract store (pinned / subscribed). Unknown keys get the 404 they got before.
3. Require \`Sec-Fetch-Site\` to indicate a same-origin or cross-origin-from-a-loaded-contract referrer.

Option 2 is the narrowest fix and preserves the user-facing scenario from #3940 (Delta loads an \`<img>\` from a known contract — both contracts end up visited by the user, so both should be in the local store or becoming so via subscription).

## Priority

Low — no known abuse in the wild; the analogous \`contract_home\` DoS vector has not been exploited. Filing so it doesn't get lost.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T16:03:39.582828400+00:00
Issue hash: `9f8d7404`

## Value Assessment

The issue remains fully relevant and unaddressed. The `variable_content` handler in `crates/core/src/server/path_handlers.rs` still unconditionally calls `ensure_contract_cached()` on any cache miss, regardless of whether the contract key is known to the local store. There is no source-IP rate limiting and no "is this contract subscribed/pinned?" gate before issuing the outbound network GET.

[...]

PR #5012, opened 2026-07-29
fix(server): bound the webapp cache with LRU eviction

sanity:
## Problem

The unpacked webapp cache (`~/.cache/freenet/webapp_cache/<instance_id>/`) has
per-contract staleness replacement but **no global bound**. It gains one entry
for every web contract the node has ever served and never sheds any, so it grows
without limit for the life of the install.

Measured on two real peers:

| peer | size | entries | oldest entry |
|---|---|---|---|
| A | 325 MB | 61 | 2026-03-14 (4.5 months untouched) |
| B (dev workstation) | 1236 MiB | 82 | 2026-01-30 (179 days untouched) |

These bytes are also **invisible to the node's own disk accounting**. The webapp
cache lives under the XDG *cache* dir, while `ring/hosting/disk_usage.rs` seeds
`disk_total_bytes` from `contracts_dir` (WASM blobs) + the redb state rows +
`wasmtime_cache_dir` — see `contract/handler.rs:130`, which is the only caller of
`set_hosting_disk_paths`. So nothing in the node bounded, measured, or reported
this directory, and it sat entirely outside the `#4683` disk budget.

Nothing depends on an entry existing: the cache is a pure, recomputable artifact
that `unpack_if_stale` re-extracts (`remove_dir_all` + `create_dir_all` +
`unpack`) whenever the contract's state hash changes. Evicting an entry costs one
re-unpack of state the node already holds.

## Solution

A 64 MiB total-size bound with LRU eviction, in `server/path_handlers.rs`.

**Access signal — mtime of the `{key}.hash` sentinel.** The sentinel already
exists (it is the "cache is populated" marker and holds the state hash), so this
adds no files and no sidecar index, and it survives restarts, which an in-memory
table would not. It is refreshed two ways, so it tracks last **use**, not last
creation:

- `unpack_if_stale` rewrites it on every re-extraction (already did).
- `refresh_cache_if_due` — the single point both cache-reading handlers
  (`variable_content`, `serve_sandbox_content`) pass through, plus `contract_home`
  after a successful fetch — refreshes it via `utimensat` (`filetime`) while the
  entry is being served.

The refresh is throttled to one per contract per 5 minutes
(`WEBAPP_CACHE_ACCESS_TOUCH_INTERVAL`), because a single page load fans out many
subresource requests and none of them need their own syscall. That makes the
on-disk ranking accurate to within 5 minutes — far finer than the horizon
eviction actually discriminates on (hours to months). `utimensat` rather than
rewriting the file, because the sentinel's *contents* are the state hash
`unpack_if_stale` compares against and a rewrite would race a concurrent unpack.

**The cache root comes from the node's config, not a global.** `WebappCache`
carries the root, the budget and the sweep bookkeeping. The router builds one per
server from `WebsocketApiConfig::webapp_cache_dir` (a derived, `#[serde(skip)]`
field, exactly like the existing `secrets_dir`) and threads it through the axum
`State` into every web handler, and on down through `refresh_cache_if_due` →
`ensure_contract_cached` → `handle_get_response` → `unpack_if_stale`.

That is deliberately a *required parameter with no fallback in the chain*, and it
is the second attempt at this isolation. The first gated a temp-dir redirect on
`#[cfg(test)]` — which is **false** in an integration test, because the lib is
linked there as an ordinary dependency. `tests/playwright_shell.rs` boots a real
node and fetches a shell page on a plain `cargo test`, so it still swept the
developer's real cache. Threading has no such blind spot: a caller that does not
supply a root does not compile. `#[freenet_test]` points each node at its own
temp dir, so every integration-test node is isolated for free.

The default root is unchanged (`~/.cache/freenet/webapp_cache`) — relocating it
would strand every existing install's directory with nothing left to sweep it,
which is the opposite of the point. `FREENET_WEBAPP_CACHE_DIR` overrides it, which
also gives operators running several nodes as one user a way to stop one node's
sweep reaching another's entries.

**Sweep trigger — after an unpack, plus a debounced reconcile sweep.** An unpack
is the only event that grows the cache, so it always sweeps, and the directory
walk it costs is small change next to the `remove_dir_all` + `unpack` it just
performed. No request that reads from a warm cache pays anything. The
`!needs_update` exit of `unpack_if_stale` also gets a sweep debounced to 10
minutes (`WEBAPP_CACHE_SWEEP_INTERVAL`), because a node that upgrades with an
already-oversized cache may keep serving contracts whose state hash never changes
and would otherwise never sweep at all. There is no `du` on the request path.

**Never evicting something in use** — three layers:

1. The contract whose request triggered the sweep is never its own victim.
2. Any entry served within `WEBAPP_CACHE_EVICTION_MIN_IDLE` (120s) is skipped.
   120s comfortably exceeds the 30s network-fetch timeout in
   `ensure_contract_cached`, so a request merely waiting on the node cannot have
   its files deleted underneath it. Per the GC-exemption rule in `AGENTS.md` this
   exemption is time-bounded, and a test pins that it expires.
3. An entry whose `CONTRACT_CACHE_LOCKS` mutex is held (an unpack is in flight)
   is skipped via `try_lock`.

These are a strong preference, not an interlock. The check and the
`remove_dir_all` are not atomic, and both guards are per-process while the
directory is per-user, so an eviction racing a request stays possible in
principle. What that costs is bounded: the cache is recomputable, and on Unix an
already-opened file survives unlinking (`ServeFile` opens the descriptor before
streaming), so an in-flight download completes and the worst case is a 404 or a
refetch. The docs and commit message say exactly this rather than claiming
nothing in use is ever evicted.

Concurrent sweeps are gated to one at a time. Each takes its own `live` snapshot
and deletes until *it* has freed the deficit, so N simultaneous unpacks would
each evict a full deficit's worth, drive the cache below budget, and over-report
`bytes_freed`.

The bound is best-effort in the other direction too: if every oversized entry is
protected, or a single webapp is itself larger than the budget, the sweep leaves
the cache over budget and the next one retries rather than deleting something
live.

**Resilience.** Per-entry removal failures are logged at `warn!` and the sweep
continues; nothing propagates to the request. Removal order is sentinel first,
then tree — a directory with no `{key}.hash` reads as a *cold* cache and is
simply refetched, whereas a `{key}.hash` with no directory reads as *warm* and
would serve 404s until the contract's state happened to change. An entry whose
sentinel cannot be removed is left fully intact rather than half-deleted into
that shape. Eviction also drops the entry's `CONTRACT_CACHE_REFRESH` timer:
`refresh_cache_if_due` short-circuits on a fresh timer alone, so an evicted
contract that kept its timer would 404 from the emptied directory for the rest of
the TTL window.

**Bounding.** The in-memory access table is only written where the entry is known
to exist (a warm sentinel, or a fetch that just populated one) and the sweep drops
the record on eviction, so it is bounded by the on-disk entry count — which the
budget bounds in turn. That gating is deliberate: `variable_content` is reachable
unauthenticated with an arbitrary key, so recording an access *before* the #3945
presence gate has run would hand an attacker an unbounded per-key map.

**Cross-process safety.** The cache directory is per-*user* but the guards are
per-process, and the documented multi-peer setup (`peer-manager.sh`) runs several
nodes as one user. When another process evicts an entry, this process's reconcile
timer is still fresh and knows nothing about it, so returning on the timer alone
served 404s out of the emptied directory for the rest of the TTL window.
`refresh_cache_if_due` now re-stats the sentinel under the refresh lock and
requires warm **and** fresh before short-circuiting.

**Sweep ownership.** `from_base58` is not a strict filter — stdlib zero-pads a
short decode rather than erroring, so ordinary names like `tmp`, `data` and
`assets` parse into well-formed but *wrong* ids. The scan therefore requires the
name to round-trip (parse, re-encode, compare). Without that the sweep would
charge a stray directory's bytes to a phantom entry, delete a path that does not
exist, treat the resulting `NotFound` as success, and count bytes it never
freed — reporting evictions while staying over budget.

**The one-off cost.** The sweep is awaited inline, which is right in steady
state (a directory walk plus at most one deletion, next to the `remove_dir_all` +
`unpack` that triggered it). The expensive case is the *first* sweep after
upgrading a node with an unbounded legacy cache: ~78 `remove_dir_all`s before the
shell page returns. Deletions are therefore capped per sweep
(`WEBAPP_CACHE_MAX_EVICTIONS_PER_SWEEP`, 8), so that backlog drains over the next
handful of unpacks rather than landing on one request; steady state never reaches
the cap. The walk itself is not capped — it is the price of knowing the size at
all — but it runs on `spawn_blocking` and shrinks with the cache. Capping rather
than detaching keeps the wiring tests synchronous and deterministic.

**The number.** 64 MiB against the observed per-entry distribution (a typical
unpacked webapp is a few hundred KB to a few MB) keeps roughly the last 15-30
distinct webapps actually browsed. Replaying peer B's real cache through the
ranking: 1236 MiB → 57.5 MiB across the 15 most recently used entries, freeing
1178 MiB (95.4%). Peer A's 325 MB would fall to roughly 64 MB, freeing ~260 MB
(~80%). `filetime` is a new *direct* dependency but not a new build unit — `tar`
already pulls it in, so the lockfile delta is one line.

## Testing

29 new tests, every one of them against a `TempDir` root. They cover the LRU ordering, each of the three in-use guards, the
exemption expiring, sentinel + refresh-timer cleanup, removal-failure resilience,
the touch throttle, the marker not disturbing the state hash, the concurrent-sweep
gate, the zero-padding filter, the cross-process refetch, and the boundary/edge
cases (exactly-at-budget, empty root, missing root, all-entries-protected,
unrecognized paths).

Nine exercise the **wiring** rather than the sweep in isolation, driving the real
handler path end to end: both sweep call sites in `unpack_if_stale`, all three
`note_cache_access` call sites (including the cold-fetch one), the reconcile
debounce as actually wired, the per-sweep cap, the router honouring the configured
cache dir, and — in `freenet-macros` — a pin that every harness node redirects its
cache. Without those the feature could be deleted from `unpack_if_stale` /
`refresh_cache_if_due` with a green suite.

**Mutation-tested** — every test was proven load-bearing by deliberately breaking
the code and confirming it goes red. 23 mutations across three rounds, each
applied with a uniqueness assertion on the anchor and a read-back check that the
edit reached disk (a mutation that silently fails to apply is indistinguishable
from a passing test).

Round 3 mutates the **call sites**, which is what the first two rounds missed:
they covered `enforce_webapp_cache_budget`'s internals only, so the review found
three deletions that removed the feature from production with a green suite. All
three now fail, and each was also verified individually rather than only as a
group:

| # | deletion | test(s) that go red |
|---|---|---|
| D1 | both `maybe_enforce_webapp_cache_budget` calls in `unpack_if_stale` | `reconcile_path_enforces_the_webapp_cache_budget`, `unpack_enforces_the_webapp_cache_budget` |
| D1a | only the `Reconcile` call site | `reconcile_path_enforces_the_webapp_cache_budget` |
| D1b | only the `Unpack` call site | `unpack_enforces_the_webapp_cache_budget` |
| D2 | all three `note_cache_access` calls | `serving_a_warm_entry_marks_it_in_use`, `contract_home_marks_the_entry_in_use` |
| D2a | only the `refresh_cache_if_due` warm-path call | `serving_a_warm_entry_marks_it_in_use` |
| D2b | only the `contract_home` call | `contract_home_marks_the_entry_in_use` |
| D3 | the `Reconcile` debounce condition | `webapp_cache_sweep_is_due_debounces_only_reconciles` |
| M14 | revert the cross-process re-stat (trust the timer alone) | `eviction_by_another_process_forces_a_refetch_despite_a_fresh_timer` |
| M15 | drop the base58 round-trip check | `webapp_cache_sweep_ignores_names_that_zero_pad_into_valid_ids` |
| M16 | drop the concurrent-sweep gate | `webapp_cache_concurrent_sweeps_do_not_over_evict` |

Rounds 1-2 covered the sweep's own logic:

| # | mutation | test(s) that went red |
|---|---|---|
| M1 | reverse the LRU ordering (evict most-recently-used first) | `evicts_least_recently_used_first`, `access_marker_makes_an_old_entry_most_recently_used` |
| M2 | drop the "entry that triggered this sweep" guard | `never_evicts_the_entry_in_use` |
| M3 | drop the recent-access (in-flight request) guard | `skips_recently_accessed_entry`, `stays_over_budget_when_all_entries_protected`, `access_exemption_expires` |
| M4 | ignore the in-flight unpack lock | `skips_entry_with_unpack_in_flight` |
| M5 | keep the reconcile timer for an evicted entry | `clears_refresh_timer_for_evicted_entry` |
| M6 | leave the `{key}.hash` sentinel behind | `removes_sentinel_with_directory`, `continues_after_entry_removal_failure`, `access_exemption_expires` |
| M7 | abort the sweep on the first removal failure | `continues_after_entry_removal_failure` |
| M8b | off-by-one at the budget boundary (both guards) | `is_noop_at_or_under_budget`, `evicts_least_recently_used_first`, `access_marker_makes_an_old_entry_most_recently_used` |
| M9 | remove the marker-refresh throttle | `access_marker_refresh_is_throttled` |
| M10 | make the on-disk marker refresh a no-op | `access_marker_updates_mtime_without_touching_contents`, `access_marker_makes_an_old_entry_most_recently_used` |
| M11 | make the access exemption effectively permanent | `access_exemption_expires` |
| M12 | account for / evict paths that are not contract entries | `ignores_unrecognized_paths` |
| M13 | panic instead of tolerating an unreadable cache root | `scan_of_missing_root_is_empty_not_a_panic` |

Three of these initially **survived**, and each survival was a real finding rather
than something to wave through:

- **M8** (flipping only the early-return `total <= max_bytes`) is a genuinely
  *equivalent* mutant: the eviction loop's own `live <= max_bytes` guard fires
  immediately at `live == total == max_bytes`, so the observable behaviour is
  identical (one wasted sort). M8b flips both guards and is killed.
- **M11** exposed a self-referential test: it advanced time by
  `WEBAPP_CACHE_EVICTION_MIN_IDLE` itself, so the assertion moved with the
  constant and could not notice the window becoming effectively permanent. The
  test now also asserts the constant is a sane finite window (longer than the 30s
  fetch it must cover, well short of a browsing session).
- **M13** exposed a test that could not distinguish graceful handling from a
  *swallowed* panic — a panic inside the sweep's `spawn_blocking` is caught by the
  `JoinError` arm and surfaces as an empty sweep, which is exactly what the test
  asserted. Added a direct unit test on `scan_webapp_cache` for a missing root.

Also run: `cargo fmt`, `cargo clippy -p freenet --lib --all-features` clean, and
the full `server::path_handlers` module green.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2058,6 +2058,7 @@ dependencies = [
  "ed25519-dalek",
  "either",
  "event-listener",
+ "filetime",
  "flatbuffers 25.12.19",
  "flate2",
  "freenet-macros 0.1.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ moka = { version = "0.12", features = ["sync"] }
 # File I/O
 directories = "6"
 dirs = "6"
+filetime = "0.2"
 flate2 = "1"
 notify = "8"
 tar = "0.4"
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -60,9 +60,12 @@ directories = { workspace = true }
 dirs = { workspace = true }
 either = { features = ["serde"], workspace = true }
 event-listener = { workspace = true }
-# `utimensat`-backed mtime refresh for the webapp cache's LRU marker
-# (server/path_handlers.rs). Already in the tree as a `tar` dependency, so this
-# is a re-export of an already-compiled crate rather than a new build unit.
+# Timestamp-only mtime refresh for the webapp cache's LRU marker
+# (server/path_handlers.rs) — 0.2.29 opens the file and calls `futimens`. Used
+# instead of rewriting the file because the marker IS the state-hash sentinel,
+# and a rewrite would race a concurrent unpack. Already in the tree as a `tar`
+# dependency, so this is a re-export of an already-compiled crate rather than a
+# new build unit.
 filetime = { workspace = true }
 flate2 = { workspace = true }
 flatbuffers = { workspace = true }
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -60,6 +60,10 @@ directories = { workspace = true }
 dirs = { workspace = true }
 either = { features = ["serde"], workspace = true }
 event-listener = { workspace = true }
+# `utimensat`-backed mtime refresh for the webapp cache's LRU marker
+# (server/path_handlers.rs). Already in the tree as a `tar` dependency, so this
+# is a re-export of an already-compiled crate rather than a new build unit.
+filetime = { workspace = true }
 flate2 = { workspace = true }
 flatbuffers = { workspace = true }
 futures = { workspace = true }
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1036,6 +1036,9 @@ impl ConfigArgs {
                 // Runtime-only: resolve the secrets dir for this mode so the WS
                 // serve layer can stamp per-user activity markers (#4561).
                 secrets_dir: config_paths.secrets_dir(mode),
+                // Runtime-only: resolve the unpacked-webapp cache so the HTTP
+                // layer knows which directory its LRU sweep may delete from.
+                webapp_cache_dir: default_webapp_cache_dir(),
             },
             secrets,
             log_level: self.log_level.unwrap_or(tracing::log::LevelFilter::Info),
@@ -2329,6 +2332,42 @@ pub struct WebsocketApiConfig {
     /// has no secrets tree to mark.
     #[serde(skip)]
     pub secrets_dir: std::path::PathBuf,
+
+    /// Directory holding unpacked web-contract bundles, derived in `build()`
+    /// (see [`default_webapp_cache_dir`]). Runtime-only for the same reason as
+    /// `secrets_dir` above: it is a resolved path, not operator-authored TOML.
+    ///
+    /// The HTTP layer builds its `WebappCache` from this and threads it to the
+    /// web handlers. It is a config value rather than a process global on
+    /// purpose — the cache is size-bounded by LRU EVICTION, so whoever owns the
+    /// path owns a directory something will delete from. Two consequences:
+    /// a node pointed at a temp data dir (every `#[freenet_test]` node) gets an
+    /// isolated cache instead of sweeping the developer's real one, and two
+    /// nodes run by the same user can be given separate caches instead of
+    /// silently sharing one.
+    #[serde(skip)]
+    pub webapp_cache_dir: std::path::PathBuf,
+}
+
+/// Default directory for unpacked web-contract bundles.
+///
+/// The XDG cache dir (`~/.cache/freenet/webapp_cache` on Linux), which is where
+/// this cache has always lived — deliberately unchanged, because relocating it
+/// would strand every existing installation's directory with nothing left to
+/// sweep it, which is the opposite of what the size bound is for.
+///
+/// `FREENET_WEBAPP_CACHE_DIR` overrides it. That exists for operators running
+/// several nodes as one user: the cache is per-user by default but its eviction
+/// guards are per-process, so pointing each node at its own directory is the
+/// clean way to keep one node's sweep away from another's entries.
+pub fn default_webapp_cache_dir() -> std::path::PathBuf {
+    if let Some(dir) = std::env::var_os("FREENET_WEBAPP_CACHE_DIR") {
+        return std::path::PathBuf::from(dir);
+    }
+    directories::ProjectDirs::from("", "The Freenet Project Inc", "freenet")
+        .map(|dirs| dirs.cache_dir().to_path_buf())
+        .unwrap_or_else(|| std::env::temp_dir().join("freenet"))
+        .join("webapp_cache")
 }
 
 #[inline]
@@ -2375,6 +2414,7 @@ impl From<SocketAddr> for WebsocketApiConfig {
             per_user_op_burst: default_per_user_op_burst(),
             per_user_export_min_interval_secs: default_per_user_export_min_interval_secs(),
             secrets_dir: std::path::PathBuf::new(),
+            webapp_cache_dir: default_webapp_cache_dir(),
         }
     }
 }
@@ -2394,6 +2434,7 @@ impl Default for WebsocketApiConfig {
             per_user_op_burst: default_per_user_op_burst(),
             per_user_export_min_interval_secs: default_per_user_export_min_interval_secs(),
             secrets_dir: std::path::PathBuf::new(),
+            webapp_cache_dir: default_webapp_cache_dir(),
         }
     }
 }
@@ -4999,6 +5040,7 @@ mod tests {
                 // serde-skip runtime field; repopulated by build() and not
                 // asserted in the round-trip (bound to `_` in the destructure).
                 secrets_dir: std::path::PathBuf::new(),
+                webapp_cache_dir: default_webapp_cache_dir(),
             },
             secrets: base.secrets.clone(),
             log_level: tracing::log::LevelFilter::Debug,
@@ -5205,6 +5247,9 @@ mod tests {
             per_user_op_burst,
             per_user_export_min_interval_secs,
             secrets_dir: _, // serde-skip runtime field, repopulated by build()
+            // serde-skip runtime field, repopulated by build() from
+            // `default_webapp_cache_dir()` (env-overridable).
+            webapp_cache_dir: _,
         } = ws_api;
         assert_eq!(ws_address, seed.ws_api.address, "ws_api.address");
         assert_eq!(ws_port, seed.ws_api.port, "ws_api.port");
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -2345,6 +2345,43 @@ pub struct WebsocketApiConfig {
     /// isolated cache instead of sweeping the developer's real one, and two
     /// nodes run by the same user can be given separate caches instead of
     /// silently sharing one.
+    ///
+    /// # What isolation this actually guarantees
+    ///
+    /// Precisely: **every node built through [`ConfigArgs::build`] serves from
+    /// the directory its config names, and every `#[freenet_test]` node has that
+    /// config pointed at its own temp dir** (the harness assigns it; pinned by
+    /// `every_node_isolates_its_webapp_cache` in `freenet-macros`). That covers
+    /// production and every integration test that goes through the harness,
+    /// including `tests/playwright_shell.rs`, the one test that actually fetches
+    /// `/v1/contract/web/` and therefore the one that actually sweeps.
+    ///
+    /// It is NOT "no test can ever reach the real cache". Two standalone
+    /// composition paths deliberately resolve [`default_webapp_cache_dir`]
+    /// themselves, because both are real user-facing modes rather than test
+    /// scaffolding:
+    ///
+    /// - `HttpClientApi::as_router`, which `server::local_node::run_local_node`
+    ///   serves through. Its signature is public API and takes no cache root.
+    /// - `WebsocketApiConfig::default()` / `From<SocketAddr>`, the fallback for
+    ///   any serving config not produced by `build()`.
+    ///
+    /// Leaving those resolved is the deliberate choice (see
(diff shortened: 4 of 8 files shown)


sanity:
Addressed the three findings in ad25f408.

**1. The call-site pin (`server.rs:655`).** Agreed, and this was the real gap: the threading is a compile-time guarantee against *omission* only, and the mutation that matters is *substitution*, which was free. Added `serve_client_api_builds_the_router_with_the_configured_webapp_cache_dir` in `server.rs`'s test module. It cuts the file at `mod tests {` (not `#[cfg(test)]` — `path_handlers.rs` has three of those above its test module, which is exactly how that cut goes vacuous), strips line comments, collapses whitespace so rustfmt re-wrapping the call cannot rot the needle, and splits every needle with `concat!`. Two assertions: the call's balanced-paren argument list must contain `config.webapp_cache_dir`, and `default_webapp_cache_dir` must not appear anywhere in `server.rs`'s production source.

Mutation result, verified with `git diff` that each edit landed on disk before trusting the outcome:

| mutation | result |
|---|---|
| `config.webapp_cache_dir.clone()` → `crate::config::default_webapp_cache_dir()` | new pin RED; pre-existing `router_config_is_rooted_at_the_configured_webapp_cache_dir` still green, confirming the gap was real |
| revert | green |
| keep the argument but add a fallback: `config.webapp_cache_dir.clone().canonicalize().unwrap_or_else(...default_webapp_cache_dir())` | RED on the second assertion |

Also probed the layer below: dropping the parameter in `as_router_with_origin_contracts` produces `unused variable: webapp_cache_dir`, and CI builds with `-D warnings`, so that hop is already guarded by the compiler. `server.rs:655` was the only silent one.

**2. `Default` / `From<SocketAddr>`: stated the narrower guarantee, did not make them empty.** Reasoning, spelled out in the field's rustdoc and pinned by `standalone_websocket_api_config_resolves_the_real_webapp_cache_dir`:

- Empty is not benign *here*. It is benign for `secrets_dir` only because that field's consumer reads empty as "stamping disabled". This field has no such consumer rule, so an empty root writes cache entries under the process's working directory and skips the sweep entirely, i.e. it installs finding 3's pathology as the default for standalone composition. Copying the value without the consumer semantics copies the look of the precedent, not its safety.
- It would not close the door it aims at. `as_router` is the direct router-composition entry a test would reach for, and it resolves `default_webapp_cache_dir()` itself because `local_node::run_local_node` serves through it; its signature is public API and carries no cache root. So one of the two doors stays open regardless.
- These constructors are the fallback for any future serving path that is not `ConfigArgs::build()`, where an unbounded cache under an arbitrary working directory is worse than a shared but bounded one in the canonical place.

The field's rustdoc now states exactly what is guaranteed (every `build()` node serves from the directory its config names; every `#[freenet_test]` node has that config pointed at its own temp dir), what is not, and what a new test composing a server or router directly has to do. The residual risk is made audible rather than silent by the startup log below. `WebsocketApiConfig::default().webapp_cache_dir = PathBuf::new()` trips the pin.

**3. Empty override, plus the missing log.** `FREENET_WEBAPP_CACHE_DIR=` now reads as unset, with a warning. The environment read is hoisted into a parameter (`resolve_webapp_cache_dir`) so the case is tested without `set_var`, which is `unsafe` in edition 2024 precisely because a racing writer in another test would be a flake rather than a finding; a source pin keeps `default_webapp_cache_dir` feeding the real environment read through that resolver. `WebappCache::with_root` now creates the root and logs it once with the budget, which is also what turns the other two shapes you named (root exists as a file, root cannot be created) into a startup warning instead of a cache that silently never populates and a sweep that silently never runs.

Mutations for 2 and 3, each verified on disk: accept an empty override → RED; `default_webapp_cache_dir` stops reading the environment → RED; drop the `create_dir_all` → RED; `.expect()` on it instead of warning → RED (a stray file at the root path would otherwise take the node down); empty `Default` → RED.

Also fixed three stale comments in `path_handlers.rs`'s test module that still referred to `PRODUCTION_WEBAPP_CACHE` and `webapp_cache_dir()`, neither of which exists any more.

`cargo fmt` clean, `cargo clippy -p freenet --lib --all-features` clean, full `cargo test -p freenet --lib` green (4387 passed). One unrelated flake surfaced during the repeat runs (`wasm_runtime::delegate::test::test_v2_delegate_update_existing_state`, 2 failures in 7 full-suite runs, never under a narrower filter); it is the #5023 family and the data point is recorded there rather than in a duplicate issue.

Not merging: the code changed, so the earlier review is stale per the per-content rule, and CI needs to go green on ad25f408.

[AI-assisted - Claude]

PR #5018, opened 2026-07-29
perf: reclaim the 58 MB known-good rollback binary at probation commit

sanity:
## Problem

Every Freenet node carries `~/.local/state/freenet/known_good_binary` — an
**uncompressed 58 MB copy of the executable** (measured on a real peer) — for
the entire days-to-weeks gap between releases, and nothing ever deletes it.

The blob is the crash-loop auto-rollback target (#4073): `freenet update`
snapshots the live binary just before overwriting it, then arms a probation
marker. It is reachable **only** through that marker:

* `handle_post_stop_at` returns `PostStopOutcome::Proceed` and never touches
  the blob when `read_probation_at` is `None`.
* `prepare_known_good_for_install_at` only consults the existing blob inside
  `if let Some(existing) = read_probation_at(dir)`; with no marker it
  re-captures fresh from the live binary.

Probation commits after 60 s of healthy uptime (`commit_probation`), which
removes the marker. From that moment the 58 MB is unreachable by any code
path — pure dead weight until the next update overwrites it.

Traced consumers (repo-wide sweep for `known_good` / `rollback_binary` across
Rust, shell, CI workflows, packaging, systemd units, the uninstaller and
`freenet service report`): `rollback.rs` only, plus `update.rs` calling
`prepare_known_good_for_install`. `service/purge.rs` deletes the whole state
directory on uninstall but never reads the blob.

## Solution

Reclaim the blob when the probation marker that made it reachable is retired.
This *removes* a resident cost rather than adding a code path — the alternative
(compressing the blob) was rejected: it would put a decode step on the one path
whose failure bricks a node, and the recorded SHA-256 would cover the stored
bytes rather than the bytes actually installed.

`commit_probation_at` now calls `discard_known_good_at` in both marker-removing
arms (`Committed` and `ClearedStale`). Because this is brick-recovery state, the
reclaim only unlinks a file it can positively identify as the dead rollback
target:

1. **Re-reads the marker and does nothing if one exists.** This makes the
   fail-safe ordering structural rather than documented — see below.
2. **Hashes the candidate; deletes only on a size + SHA-256 match** against the
   meta recorded in the marker being retired. A snapshot a concurrent
   `freenet update` captured for the *next* probation generation has different
   content and survives.
3. **Only ever considers `dir/known_good_binary`**, never the `rollback_binary`
   path carried in the marker JSON, so a corrupt marker cannot make us unlink an
   arbitrary file.

It returns `()`; every failure is logged (warn on a failed unlink or an
unverifiable blob) and swallowed, so reclamation can never fail a probation
commit or an update. A blob we fail to delete is simply overwritten by the next
install's capture — exactly today's steady state.

`CommitOutcome::Nothing` (no marker) deliberately does **not** delete: without a
marker there is no recorded size/SHA to identify the blob against, and deleting
on a guess is the wrong trade in this module. The post-stop paths in
`handle_post_stop_at` that remove the marker themselves (rollback succeeded,
stale marker, aged-out marker) also leave their blob for the next capture to
overwrite — they are the brick-recovery paths and are left untouched.

### Crash-ordering (requirement 4)

Remove the marker **first**, then reclaim the blob. The two orderings are not
symmetric:

| Crash point | Result |
|---|---|
| **Chosen:** after marker removal, before unlink | Marker gone (rollback correctly disarmed), blob orphaned. Identical to today's steady state; the next install's capture overwrites it. |
| **Rejected:** after unlink, before marker removal | An **armed** marker advertising a rollback target that no longer exists. A subsequent crash-loop burns to `RollbackUnavailable` instead of restoring. |

The guard in (1) enforces this rather than relying on comments: a caller that
reclaims while still armed gets a no-op, so the unsafe order cannot silently
ship.

### The tradeoff, stated plainly

**After probation commits, a node that degrades later has no local binary to
revert to and must recover by re-installing from GitHub.** Weigh it as:

* The crash-loop class this module exists for — a bad release that wedges at
  boot — is *unaffected*: it fires inside the 60 s probation window while the
  blob is still present.
* Post-commit, the blob was already unusable (no marker ⇒ no reader), so
  nothing that worked before stops working. This change makes the resource cost
  match the actual reachability.
* It leaves a committed node in the same position as one whose known-good
  capture failed — already a supported, non-fatal state that `commands::update`
  handles with a warning.
* Maintainer-approved tradeoff (behavior change, so flagging it explicitly per
  CONTRIBUTING.md).

### Concurrency (revised after review — the first version of this section was wrong)

The original PR claimed the concurrent-install path was "unreachable on the
auto-update path". That is true for systemd/post-stop, but **false in general**:
`service/wrapper.rs:909` runs a full `freenet update --quiet` *install* (not
`--check`) on the tray's "Check for Updates" while the `freenet network` child is
still running — it is not killed until `:917`. So on Windows/macOS a live node's
60 s commit timer and a real install genuinely overlap. Both windows are also
**second-scale, not microsecond-scale**: `replace_binary` copies 58 MB between
`prepare` and `begin_probation`, and the reclaim side is a cold 58 MB read + hash.

Two changes narrow it from both ends:

* **Re-read the marker after hashing, immediately before `remove_file`.** The
  first read is a cheap early-out; the read that the delete decision actually
  rests on is now one syscall from the unlink instead of a whole hash away.
* **`begin_probation_at` verifies the blob after writing the marker** and
  disarms (returns `Err`, which `update.rs` already surfaces as a warning) if it
  has gone. This closes the race from the installer's side regardless of
  interleaving. An armed marker naming a missing binary is *strictly worse than
  no marker*: it burns the crash budget to `RollbackUnavailable` **and**
  suppresses the ordinary forward-update self-heal (#4549) that a missing marker
  would have allowed.

What remains: both sides are still check-then-act on an unlocked state
directory, so an installer that writes its marker in the sub-syscall window
between the re-check and the unlink is not seen. Only a lock on the state
directory would close that. The rustdoc on `discard_known_good_at` says exactly
this rather than claiming the concurrent case is handled — on brick-recovery
code, a comment that overstates coverage is itself a hazard.

### Blocking I/O

`commit_probation` is blocking file I/O and now includes a cold 58 MB read +
SHA-256. It ran on a tokio worker (`GlobalExecutor::spawn` → plain
`handle.spawn`) at T+60 s of the first boot after every auto-update — on every
peer, every release, during ring bootstrap. It is now wrapped in
`tokio::task::spawn_blocking`, pinned by a whitespace-stripped source scrape of
`bin/freenet.rs`.

### What the blob was, beyond the automated readers

The module docs no longer say the blob is "reachable ONLY through a probation
marker". That holds for automated readers; it is false for an **operator with
shell access**, for whom it doubled as an offline recovery artifact for the
post-commit degradation class — a use no code path expresses and this module
never promised, but a real one. The docs now name it, and also note that the
commit trigger is a bare 60 s `sleep` with no health check beyond "the process
is still alive", so "committed" means less than it sounds like. Whether that
warrants keeping an offline copy or a stronger commit condition is flagged for
the maintainer, not decided here.

## Testing

New tests in `crates/core/src/bin/commands/rollback.rs`:

* `commit_reclaims_the_known_good_blob` — the main behavior.
* `commit_clears_stale_marker_for_other_version` (extended) — the
  `ClearedStale` arm reclaims too.
* `commit_with_no_marker_leaves_an_unidentifiable_blob` — `Nothing` never
  deletes on a guess.
* `commit_leaves_a_blob_that_is_not_the_retired_rollback_target` — the
  concurrent-fresh-capture race: content mismatch ⇒ blob survives.
* `discard_known_good_refuses_while_a_probation_marker_exists` — pins the
  fail-safe ordering, then proves the still-armed marker *still rolls back*.
* `commit_succeeds_even_when_the_blob_cannot_be_verified` — best-effort: an
  unhashable blob is left alone and the commit outcome is unchanged.
* `next_install_recaptures_a_fresh_blob_after_commit_reclaimed_it` —
  requirement 3, end-to-end from a state where the blob is **absent** (not
  merely stale): commit reclaims → the next `prepare_known_good_for_install_at`
  captures the live binary fresh and labels it correctly → arm the next
  probation → three crashes → the fresh target genuinely restores.

Added in review: `commit_leaves_a_same_size_blob_whose_content_differs`
(an 11-byte impostor against an 11-byte expected size — the original test used
38 bytes and so only exercised the *size* check),
`begin_probation_disarms_when_the_known_good_blob_is_missing`,
`begin_probation_arms_normally_when_the_blob_is_present`, and two source pins:
`discard_rechecks_the_marker_after_hashing_and_before_unlinking` and
`commit_probation_runs_on_a_blocking_thread`. The ordering pin exists because a
behavioural test for it would need a sleep-based race, i.e. a flaky test; its
needles are `concat!`-split so it cannot match its own text.

`pinned_version_is_not_reapplied_but_newer_is_allowed` armed probation from a
fabricated `KnownGoodMeta` with no blob on disk — something no production caller
does. It now captures a real blob first.

**Mutation testing — call sites and every new guard, each verified to fail
independently:**

| Mutation | Red |
|---|---|
| Delete both `discard_known_good_at(...)` calls from `commit_probation_at` | `commit_reclaims_the_known_good_blob`, `commit_clears_stale_marker_for_other_version`, `next_install_recaptures_a_fresh_blob_after_commit_reclaimed_it` |
| Remove the armed-marker early-out | `discard_known_good_refuses_while_a_probation_marker_exists` |
| Remove the post-hash re-check | `discard_rechecks_the_marker_after_hashing_and_before_unlinking` |
| **Move** the post-hash re-check above the hash (present, wrong place) | same pin — it asserts position, not existence |
| Weaken size+SHA to size-only | `commit_leaves_a_same_size_blob_whose_content_differs` |
| Drop the identity check entirely | `commit_leaves_a_blob_that_is_not_the_retired_rollback_target` |
| Remove `begin_probation_at`'s blob verification | `begin_probation_disarms_when_the_known_good_blob_is_missing` |
| Revert `spawn_blocking` to a direct call | `commit_probation_runs_on_a_blocking_thread` |

339/339 binary tests pass; `cargo fmt --check`, `clippy -p freenet --lib
--all-features` and `clippy -p freenet --bins --all-features -- -D warnings` all
clean.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/rollback.rs
+++ b/crates/core/src/bin/commands/rollback.rs
@@ -62,8 +62,9 @@
 //!    version as on-probation ([`begin_probation`]).
 //! 2. **Commit**: once the new version has run healthily for
 //!    [`COMMIT_HEALTHY_UPTIME_SECS`] the node clears the probation marker
-//!    ([`commit_probation`]). After that, ordinary later crashes never trigger
-//!    a rollback.
+//!    ([`commit_probation`]) and reclaims the known-good snapshot
+//!    ([`discard_known_good_at`]). After that, ordinary later crashes never
+//!    trigger a rollback.
 //! 3. **Detect + revert**: a crash during probation increments the marker's
 //!    crash counter ([`handle_post_stop`]); on the
 //!    [`ROLLBACK_CRASH_THRESHOLD`]th crash the previous binary is restored
@@ -74,6 +75,30 @@
 //!    (`commands::auto_update`) so we never loop update -> crash -> revert ->
 //!    re-update. A later, strictly-newer release (a fix) is NOT pinned and is
 //!    applied normally.
+//!
+//! ## The known-good blob lives exactly as long as the marker
+//!
+//! The retained snapshot is an UNCOMPRESSED copy of the executable (~58 MB
+//! measured on a real peer), and it is reachable ONLY through a probation
+//! marker: [`handle_post_stop_at`] returns [`PostStopOutcome::Proceed`] before
+//! looking at it when [`read_probation_at`] is `None`, and
+//! [`prepare_known_good_for_install_at`] re-captures fresh from the live binary
+//! rather than reading it in that case. Keeping it past the commit therefore
+//! bought nothing — it was dead weight for the days-to-weeks until the next
+//! update — so the commit reclaims it.
+//!
+//! The tradeoff this accepts: once probation has committed, a node that
+//! degrades LATER has no local binary to revert to and recovers by
+//! re-installing from GitHub. That is the same position as a node whose
+//! known-good capture failed (already a supported, non-fatal state — see
+//! `commands::update`), and the crash-loop class this module exists for (a bad
+//! release that wedges at boot) is unaffected, because it fires inside the
+//! probation window while the blob is still present.
+//!
+//! Not compression: adding a decode step to the one path whose failure bricks a
+//! node would weaken the integrity guarantee (the recorded SHA would cover the
+//! stored bytes, not the installed ones) for a saving that deleting the file
+//! beats outright.
 //!
 //! ## Bounded by construction (no new flap/loop)
 //!
@@ -173,6 +198,10 @@ const PROBATION_FILE: &str = "update_probation.json";
 const KNOWN_BAD_FILE: &str = "known_bad_version";
 
 /// Snapshot of the previous, known-good binary kept as the rollback target.
+///
+/// Captured by [`capture_known_good_at`] just before an install and reclaimed by
+/// [`discard_known_good_at`] when the probation marker that made it reachable is
+/// retired — its lifetime is exactly the marker's.
 const KNOWN_GOOD_BINARY_FILE: &str = "known_good_binary";
 
 /// Persisted post-update probation record.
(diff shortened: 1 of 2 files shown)


sanity:
Closing without merging. The maintainer's call is to **keep the rollback blob**, and investigation while deciding showed the trade was worse than this PR's framing (including mine when I proposed it).

## Why

The 58 MB was described here as dead weight once probation commits. That is true of *automated* readers and false of an operator at a shell.

Auto-rollback covers **boot-crash loops only**. The commit trigger is a bare `sleep(COMMIT_HEALTHY_UPTIME_SECS)` in `bin/freenet.rs:353` with no health check — the in-code comment says a node surviving the window "has cleared the same fast-crash boundary". So the OOM class this module's own DEPLOY WATCH note names (a memory-capped peer dying after minutes) was never covered automatically. Past the window, the blob's entire remaining value is manual offline recovery, and that is precisely the case this PR removed.

## Not wasted

The work here stands on its own and the findings are worth keeping:

- The check-then-act race across a 58 MB hash was real, and the "unreachable on the auto-update path" claim was wrong — `service/wrapper.rs:909` runs a full install while the node child is still alive, so a live node's commit timer and a real install genuinely overlap on Windows/macOS.
- `begin_probation_at` arming a marker whose blob is missing is strictly worse than no marker: it burns the crash budget to `RollbackUnavailable` *and* suppresses the forward self-heal a missing marker would permit.
- The commit path did a synchronous cold 58 MB read + SHA-256 on a tokio worker thread, not `spawn_blocking`, at T+60s of every post-update boot.

The last two are latent issues in code that stays. Filed as follow-ups rather than lost.

[AI-assisted - Claude]

PR #5020, opened 2026-07-29
fix(tracing): floor the error log at WARN regardless of RUST_LOG

### LINKED ISSUE #5015: freenet.error.* is a byte-for-byte duplicate of the main log, doubling log disk
## Problem

`freenet.error.*` is supposed to be a `WARN`+ subset of the main log. On any node with `RUST_LOG` set it is instead a **byte-for-byte duplicate of the main log**, so the log directory costs exactly twice what it should and half the retention budget buys nothing.

### Mechanism

`init_tracer` builds the error layer's filter like this (`crates/core/src/tracing/tracer.rs`):

```rust
let error_filter = tracing_subscriber::EnvFilter::builder()
    .with_default_directive(LevelFilter::WARN.into())
    .from_env_lossy();
```

`with_default_directive` is **not** a floor — it is a fallback that applies only when the parsed environment string produces no directives at all. From `tracing-subscriber-0.3.23/src/filter/env/builder.rs:334-338`:

```rust
if !has_dynamics && filter.statics.is_empty() {
    if let Some(ref default) = self.default_directive {
        filter = filter.add_directive(default.clone());
    }
}
```

So with `RUST_LOG` unset the error layer is `WARN`+ as intended, and with `RUST_LOG` set to anything at all the `WARN` default is discarded and the error layer inherits `RUST_LOG`'s level — the same level the main layer uses. The two layers then write identical content to two files.

### Evidence (nova, production gateway)

- First 20 000 lines of `freenet.error.*`: **19 011 INFO / 776 WARN / 93 ERROR**. An error log should contain no INFO at all; this is the main log's histogram.
- All 27 hourly `freenet.*` / `freenet.error.*` pairs are **identical in size**.
- Directory total: 54 rotating files, 518.9 MiB over 25.97 hours = **20.95 MB/hour**, of which ~half is this duplication.

## Why this matters

This is the single largest lever on log disk for a Freenet node. Fixing it:

- **halves log bytes outright**, on every node, with no loss of information;
- **makes any retention budget go twice as far** — the same disk buys twice the history;
- restores the error log's actual purpose. Right now `freenet.error.*` cannot be grepped as a high-signal view of what went wrong, because it contains everything.

It also blocks a disk reduction that is otherwise unavailable. In #5010 the log directory budget was investigated with a view to cutting it. At the measured gateway rate, 24 hours of history needs ~500 MiB, so the budget cannot come down from 512 MiB without costing a gateway real incident history. With the duplication fixed the rate halves and the same 24 hours fits in ~256 MiB.

## Suggested fix

Make the error layer's level an actual floor rather than a default, so it is `max(WARN, RUST_LOG)` — the error log should never contain anything below `WARN` regardless of `RUST_LOG`. Options include composing the env filter with a `LevelFilter::WARN` layer filter, or building the error filter's directives explicitly instead of relying on `from_env_lossy`'s fallback.

Worth checking whether any consumer (`freenet service report`, support workflows, dashboards) has come to depend on the error log carrying full INFO content, since fixing this will shrink it substantially.

## Scope note

Deliberately **not** fixed in #5010. That PR is log GC — which files get deleted; this is what gets written in the first place. Different logical change, different risk, separate PR.

[AI-assisted - Claude]

sanity:
## Problem

`freenet.error.*` is supposed to be a `WARN`+ subset of the main log. **On a node whose operator sets `RUST_LOG`** it is instead a byte-for-byte duplicate of the main log, so the log directory costs twice what it should and half of any retention budget buys nothing.

`init_tracer` built the error layer's filter like this:

```rust
let error_filter = tracing_subscriber::EnvFilter::builder()
    .with_default_directive(LevelFilter::WARN.into())
    .from_env_lossy();
```

`with_default_directive` is **not** a floor — it is a fallback applied only when the parsed environment string yields no directives at all. From `tracing-subscriber-0.3.23`, `src/filter/env/builder.rs`, the end of `Builder::from_directives`:

```rust
if !has_dynamics && filter.statics.is_empty() {
    if let Some(ref default) = self.default_directive {
        filter = filter.add_directive(default.clone());
    }
}
```

So with `RUST_LOG` unset the error layer was `WARN`+ as intended, and with `RUST_LOG` set to anything the `WARN` default was discarded and the error layer inherited `RUST_LOG`'s level — the same level the main layer uses. The two layers then wrote identical content to two files.

### Who is actually affected

**Not every node — this is a no-op on a stock install.** No install template sets `RUST_LOG`:

| install path | environment it sets |
|---|---|
| `service/linux.rs` (systemd, user + system units) | `FREENET_SUPERVISED`, the fast-crash marker, `HOME` |
| `service/macos.rs` (launchd plist) | none — the plist has no `EnvironmentVariables` key |
| `service/windows.rs` | none |

On any of those, `RUST_LOG` is unset, the error layer was already `WARN`+, and this change alters nothing.

The affected population is nodes whose operator sets it:

- gateways provisioned by the repo's own `scripts/init-gateway.sh`, which exports `RUST_LOG="info,freenet=debug,freenet-stdlib=debug,fdev=debug"` at line 3;
- the ten `freenet-peer-*.service` units on nova (plus `river-moderator`, `freenet-report-server`);
- anyone who exports it by hand for debugging.

The measured gateway is one of these — its running process has `RUST_LOG=info,freenet=info,freenet-stdlib=debug,fdev=debug` (nova's deployed copy of the script has drifted from the repo's `freenet=debug`; immaterial here, since `release_max_level_info` compiles DEBUG out of release builds anyway).

### Evidence (nova gateway, measured 2026-07-29)

| family | rotating files | bytes |
|---|---:|---:|
| `freenet.*` | 27 | 267,603,367 (255.2 MiB) |
| `freenet.error.*` | 28 | 277,106,637 (264.3 MiB) |
| **directory** | 55 | **519.5 MiB over ~26 h ≈ 20.9 MB/h** |

Per-level breakdown of five consecutive sealed hourly `freenet.error.*` files — an error log carrying ~35 000 INFO lines an hour is not an error log:

| file | total bytes | INFO | WARN | ERROR | WARN+ bytes | WARN+ share |
|---|---:|---:|---:|---:|---:|---:|
| `2026-07-28-21` | 9,744,405 | 34,851 | 1,381 | 178 | 483,270 | 4.96 % |
| `2026-07-28-22` | 9,824,488 | 35,042 | 1,488 | 176 | 516,093 | 5.25 % |
| `2026-07-28-23` | 9,777,026 | 34,736 | 1,618 | 188 | 557,450 | 5.70 % |
| `2026-07-29-00` | 12,645,271 | 46,874 | 2,022 | 276 | 625,552 | 4.95 % |
| `2026-07-29-01` | 10,967,387 | 39,663 | 1,367 | 165 | 456,561 | 4.16 % |

### Expected saving, scoped

**On a `RUST_LOG`-setting node** the error family keeps ~5 % of its bytes: **264.3 MiB → ~13 MiB**, the directory goes **519.5 MiB → ~268 MiB (48 % cut)**, and the rate goes **~20.9 → ~10.8 MB/h**. 24 h of history drops from ~500 MiB to ~260 MiB, so #5010's 512 MiB budget buys that gateway roughly two days instead of one.

**On a stock install: no change.** Such a node was never paying the duplication and was always at roughly the lower rate.

This matters for whoever tunes the budget next, so #5010's `LOG_DIR_MAX_BYTES` rustdoc is updated here rather than left saying the duplication blocks lowering it. The constant itself is deliberately **not** changed — how much history a gateway should keep is a separate judgement call, not a mechanical consequence of this fix. #5010's `default_budget_holds_a_day_of_a_busy_gateways_logs` keeps its pre-fix 20.95 MB/h constant on purpose (annotated): the higher rate makes that guard strictly more conservative, and "refreshing" it to 10.8 would weaken it.

## Solution

Extract `build_error_filter(directives)` and AND the env filter with a hard `LevelFilter::WARN`, so the error layer is `max(WARN, RUST_LOG)` **evaluated per target**:

```rust
tracing_subscriber::EnvFilter::builder()
    .with_default_directive(LevelFilter::WARN.into())
    .parse_lossy(directives)
    .and(LevelFilter::WARN)
```

| `RUST_LOG` | error log before | error log after |
|---|---|---|
| unset | `WARN`+ | `WARN`+ (unchanged) |
| `info` / `debug` / `trace` | everything | `WARN`+ |
| `error` | `ERROR` only | `ERROR` only |
| `freenet::ring=debug` | that target from DEBUG up | that target at `WARN`+ |

**Why AND and not a replacement.** `max(WARN, env)` rather than "always at least WARN" is deliberate: an operator who sets a *more* restrictive `RUST_LOG` must not suddenly get **more** in the error log than they asked for. Replacing the filter with a bare `LevelFilter::WARN` would resurrect records `RUST_LOG=error` explicitly suppressed, and would ignore per-target directives entirely. AND-ing keeps the operator's directives in force and makes the error log a `WARN`+ subset of what those directives already admit.

**Nothing is lost from the main log or the journal.** Only the error layer's filter changes; `build_filter` (main log + interactive console) and the stdout/stderr fallback are untouched. Per-layer filtering is what makes this safe — `Filtered::register_callsite` accumulates each layer's `Interest` and returns `Interest::always()` rather than short-circuiting, and `Layered::max_level_hint` takes the *maximum* across layers — so the error layer's WARN floor cannot suppress INFO on the main layer.

**One narrow place records leave every sink.** `build_filter` adds `moka=off` / `sqlx=error`; the error layer deliberately does not inherit them (adopting them would drop records the error log captures today). The consequence: under `RUST_LOG=info`, moka's INFO and sqlx's INFO previously reached the error log **and nowhere else**, and now reach neither. That is the intended `WARN`+ semantics rather than an oversight, and it is narrow — moka is the contract cache, and moka/sqlx `WARN`+ is still retained. Documented on `build_error_filter`.

**OpenTelemetry layers do not share this bug.** The `trace-ot` feature builds no `tracing_subscriber` layer at all — it plugs `OTEventRegister` (`tracing/metrics_client.rs`) into the crate's own `NetEventRegister` trait and drives the OTel SDK directly, with no `EnvFilter` anywhere. `init_tracer`'s `_endpoint` parameter is unused for the same reason. The only other `EnvFilter` sites in the workspace are `test_utils.rs` (`EnvFilter::new`, deliberately independent of `RUST_LOG`) and `crates/release-agent` (a single-destination subscriber, where fallback semantics are correct). Neither has a main/error split, so neither can exhibit this failure.

**Consumers checked.** `freenet service report` collects the two families separately and size-caps each, so the main log still carries full INFO and the report spends less of its budget on duplicates; `freenet service logs --err` tails `freenet.error.*` and becomes useful for the first time; the telemetry dashboard ingests the OTel collector's `logs.jsonl`, not the text logs. Nothing depends on the error log carrying INFO.

## Testing

Nine new tests in `tracing::tracer::error_filter_tests`. No environment variable is mutated (the helper takes the directive string, so `RUST_LOG` is never poked process-globally), no wall-clock or sleeps are involved, and the capture tests install their subscriber with the **thread-local** `tracing::subscriber::with_default` plus a synchronous in-memory writer, so parallel tests can neither observe nor perturb them. Ran 20× consecutively with identical results.

- `env_filter_default_directive_is_a_fallback_not_a_floor` — pins the upstream behavior being worked around, and is the non-vacuity control for the floor test (the two differ only by the code under test).
- `error_filter_is_warn_when_rust_log_unset` — the case that was already correct stays correct.
- `error_filter_floors_verbose_rust_log_at_warn` — `info` / `debug` / `trace` / `freenet=info` / `freenet=trace`.
- `error_filter_floors_per_target_directives_at_warn` — no INFO/DEBUG leak via a target-scoped directive.
- `error_filter_honors_a_more_restrictive_rust_log` — `error`, `off`, `freenet=error` are not widened.
- `error_log_directives_reads_rust_log` — the env-reading half matches `from_env_lossy()`. Near-tautological by construction; kept as documentation of that equivalence, **not** counted as coverage.
- `error_layer_writes_only_warn_and_above_under_verbose_rust_log` — end-to-end through `Filter::enabled`, not just the level hint.
- `main_layer_still_receives_info_alongside_the_floored_error_layer` — the hard constraint, plain subscriber shape.
- `main_layer_still_receives_info_under_the_release_rate_limited_shape` — **the same constraint for the shape release builds actually run.** `init_tracer` builds two shapes: the plain `Registry.with(main).with(error)` when rate limiting is off (`!cfg!(any(test, debug_assertions))`, i.e. debug/test builds) and `Registry.with(<global DynFilterFn>).with(main).with(error)` in release. The global filter layer composes `max_level_hint` differently, so both are now exercised.
- `init_tracer_wires_every_error_layer_through_build_error_filter` — call-site pin.

### Call-site mutation evidence

A correct filter helper is worthless if `init_tracer` stops calling it, and the behavioural tests exercise `build_error_filter` directly, so deleting the wiring leaves them green. Each mutation was applied to `tracer.rs`, verified to have actually changed the file (`diff` non-empty), and the suite re-run. **Re-run in full after the rebase onto #5010:**

| mutation | tests that went red |
|---|---|
| **M1** — delete `.and(LevelFilter::WARN)` (revert the fix itself) | 6 red: both floor tests, all three capture tests, and the pin |
| **M2** — revert one call site to the pre-fix inline `EnvFilter…from_env_lossy()` | **pin only** — the other 27 stayed green |
| **M3** — call site passes `""` instead of `&error_log_directives()` | **pin only** |
| **M4** — one error layer loses its `.with_filter(...)` entirely | **pin only** |
| **M5** — rename the module the pin anchors on | **pin only**, via its `expect` — the pin fails **closed** |

M1 is the explicit "revert the filter change and watch a test go red" check, and it now reddens 6 rather than 5, confirming the new release-shape test is not vacuous. M2–M4 are why the pin exists: each re-breaks #5015 in production while every internals test stays green. M5 checks the pin's own anchor — `concat!("mod error_filter_", "tests {")`, an exact anchor that can never match its own source and that panics on a rename rather than silently sliding the cut point down and swallowing test functions into the "production" slice.

Baseline restored and re-verified green after each run. `cargo fmt --check`, `cargo clippy -p freenet --lib --all-features`, and `cargo test -p freenet --lib` (4387 passed, 0 failed) are clean on the rebased head.

### Rebase note

Rebased onto `main` after #5010 merged. One conflict in `tracer.rs`: #5010 replaced `enforce_log_dir_size_cap` with `prune_log_files` (plus `live_file_indices`, `rotating_log_family`, `LogFile`), while my commit carried the old function's text adjacent to the two new helpers. Resolved by taking **main's deletion** of `enforce_log_dir_size_cap` and keeping only the two new helpers. Nothing in the error-layer wiring conflicted, and the pin's counts were re-verified post-rebase (2 error layers / 2 floored filters / 1 `.and(LevelFilter::WARN)` / 1 `from_env_lossy`). Because conflict resolution restales review, the full mutation battery and test suite above were re-run on the rebased head rather than carried over.

Closes #5015

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/tracing/tracer.rs
+++ b/crates/core/src/tracing/tracer.rs
@@ -23,12 +23,22 @@ use tracing_subscriber::{Layer, Registry};
 /// would be a serious mistake: at 96 MiB a gateway retains 4.8 hours, so
 /// an 18:00 incident is gone by 09:00.
 ///
-/// Note that ~half of that 20.95 MB/h is redundant: `freenet.error.*` is
-/// currently a byte-for-byte duplicate of the main log (see issue #5015),
-/// because `RUST_LOG` overrides the error layer's WARN default. Fixing
-/// that halves the rate outright and would let this default come down to
-/// ~256 MiB for the same history. Until it lands, this budget cannot be
-/// reduced without costing a gateway real incident history.
+/// That 20.95 MB/h is a **pre-#5015 measurement**, and ~half of it was
+/// redundant: `freenet.error.*` duplicated the main log byte for byte
+/// because `RUST_LOG` overrode the error layer's WARN default. #5015
+/// has since landed ([`build_error_filter`]), so on nodes that set
+/// `RUST_LOG` the error family drops to its WARN+ share (~5 % of its
+/// former bytes) and the same gateway now runs at ~10.8 MB/h — 512 MiB
+/// buys it roughly two days rather than one.
+///
+/// Scope that carefully before retuning this: the duplication only ever
+/// affected nodes whose operator sets `RUST_LOG`. No install template
+/// does, so a stock install was always at the lower rate and sees no
+/// change. The halving therefore applies to gateways provisioned by
+/// `scripts/init-gateway.sh` and similar, not fleet-wide. This default
+/// is deliberately left at 512 MiB here: lowering it is a separate
+/// judgement call about how much history a gateway should keep, not a
+/// mechanical consequence of #5015.
 ///
 /// Do NOT lower this so far that the current-hour files alone can
 /// approach it. The size pass never deletes a file an appender has open,
@@ -389,6 +399,85 @@ async fn periodic_log_prune(log_dir: PathBuf) {
     }
 }
 
+/// The `RUST_LOG` directive string exactly as `EnvFilter` itself would
+/// read it: the variable's value, or the empty string when unset.
+///
+/// Split out from [`build_error_filter`] so that the filter construction
+/// stays a pure function of its input and can be tested without mutating
+/// process-global environment state (which would race across the test
+/// binary's threads).
+fn error_log_directives() -> String {
+    std::env::var(tracing_subscriber::EnvFilter::DEFAULT_ENV).unwrap_or_default()
+}
+
+/// Build the filter for the `freenet.error.*` layer: `max(WARN, RUST_LOG)`,
+/// evaluated per target.
+///
+/// The error log is meant to be a WARN-and-above **subset** of the main
+/// log, so it can be grepped as a high-signal view of what went wrong.
+/// That needs a genuine floor, and `EnvFilter::with_default_directive`
+/// is not one — it is a *fallback* that is applied only when the parsed
+/// directive string produced no directives at all. See
+/// `tracing-subscriber-0.3.23`, `src/filter/env/builder.rs`, the
+/// `if !has_dynamics && filter.statics.is_empty()` branch at the end of
+/// `Builder::from_directives`.
+///
+/// So with `RUST_LOG` set — which production sets — the `WARN` default
+/// was silently discarded and the error layer inherited `RUST_LOG`'s
+/// level: the very same level the main layer uses. The two layers then
+/// wrote identical content to two files, doubling log disk on every node
+/// (issue #5015; the nova gateway's `freenet.error.*` was 19011 INFO /
+/// 776 WARN / 93 ERROR over its first 20000 lines, and every hourly
+/// `freenet.*` / `freenet.error.*` pair was byte-for-byte identical).
+///
+/// AND-ing the env filter with a hard [`LevelFilter::WARN`] makes the
+/// level an actual floor, per target:
+///
+/// - `RUST_LOG` unset → `WARN`+, unchanged from the intended behavior;
+/// - `RUST_LOG=info` / `debug` / `trace` → still only `WARN`+;
+/// - `RUST_LOG=error` → only `ERROR`. A deliberately *more* restrictive
+///   `RUST_LOG` is honored rather than widened — this is `max(WARN, env)`,
+///   not "always at least WARN", so the floor never resurrects records the
+///   operator explicitly asked to suppress;
+/// - per-target directives (`RUST_LOG=freenet::ring=debug`) → that target
+///   at `WARN`+ only, so no INFO/DEBUG can leak into the error log through
+///   a target-scoped directive either.
+///
+/// AND-ing (rather than replacing the env filter outright with a bare
+/// `LevelFilter::WARN`) is what keeps the operator's `RUST_LOG` in force:
+/// a target they silenced stays silent in the error log too.
+///
+/// This layer deliberately does NOT inherit `build_filter`'s `moka=off` /
+/// `sqlx=error` additions. Those exist to suppress third-party INFO/DEBUG
+/// chatter from the main log; at WARN+ they are not a volume concern, and
+/// adopting them here would remove records the error log captures today.
+/// So the error log is a WARN+ subset of the main log for every freenet
+/// target, and additionally keeps moka/sqlx WARN+ that the main log
+/// suppresses.
+///
+/// The one place this change removes records from EVERY sink: because the
+/// main layer sets `moka=off` / `sqlx=error`, moka's INFO and sqlx's INFO
+/// under `RUST_LOG=info` used to reach the error log **and nowhere else**,
+/// and now reach neither. That is the intended WARN+ semantics rather than
+/// an oversight, and it is narrow — moka is the contract cache, and moka
+/// and sqlx WARN/ERROR are still retained here. Anything at WARN or above
+/// from any target still lands in this log.
+///
+/// Only the error layer is affected. The main log's filter (`build_filter`
+/// inside [`init_tracer`]) and the console layer are untouched, so nothing
+/// is lost from the main log or the journal.
+fn build_error_filter<S>(directives: &str) -> impl tracing_subscriber::layer::Filter<S> + 'static
+where
+    S: 'static,
+{
+    use tracing_subscriber::filter::FilterExt;
+
+    tracing_subscriber::EnvFilter::builder()
+        .with_default_directive(LevelFilter::WARN.into())
+        .parse_lossy(directives)
+        .and(LevelFilter::WARN)
+}
+
 pub fn init_tracer(
     level: Option<LevelFilter>,
     _endpoint: Option<String>,
@@ -583,15 +672,11 @@ pub fn init_tracer(
                     .with_writer(main_writer.clone())
                     .with_filter(filter_layer);
 
-                let error_filter = tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(LevelFilter::WARN.into())
-                    .from_env_lossy();
-
                 let error_layer = tracing_subscriber::fmt::layer()
                     .with_level(true)
                     .with_ansi(false)
                     .with_writer(error_writer.clone())
-                    .with_filter(error_filter);
+                    .with_filter(build_error_filter(&error_log_directives()));
 
                 // Add console layer if running interactively
                 if also_log_to_console {
@@ -617,15 +702,11 @@ pub fn init_tracer(
                     .with_writer(main_writer)
                     .with_filter(filter_layer);
 
-                let error_filter = tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(LevelFilter::WARN.into())
-                    .from_env_lossy();
-
                 let error_layer = tracing_subscriber::fmt::layer()
                     .with_level(true)
                     .with_ansi(false)
                     .with_writer(error_writer)
-                    .with_filter(error_filter);
+                    .with_filter(build_error_filter(&error_log_directives()));
 
                 // Add console layer if running interactively
                 if also_log_to_console {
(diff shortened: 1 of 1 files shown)


sanity:
Review findings addressed. **The code changed, so the previous review is stale** — per the per-code-content rule this needs a fresh pass on `7a36125` before merge. Summary of what moved:

**1 — blast radius was overstated; corrected throughout.** Verified independently: `service/linux.rs` sets only `FREENET_SUPERVISED`, the fast-crash marker and `HOME`; the macOS plist has no `EnvironmentVariables` key; the Windows registration sets none. So **on a stock install `RUST_LOG` is unset, the error layer was already `WARN`+, and this fix is a no-op.** The affected population is nodes whose operator sets `RUST_LOG` — notably gateways provisioned by the repo's own `scripts/init-gateway.sh` (line 3 exports it), the ten `freenet-peer-*` units on nova, and hand-set debugging sessions. Confirmed the measured gateway is one of them by reading its live process environment: `RUST_LOG=info,freenet=info,freenet-stdlib=debug,fdev=debug`.

One correction to the finding as filed: the directive string quoted was nova's deployed `/usr/local/bin/init-gateway.sh`, which has drifted from the repo's `scripts/init-gateway.sh` (`freenet=debug` there, `freenet=info` deployed). Immaterial to the numbers — `release_max_level_info` compiles DEBUG out of release builds — but it means this is *not* nova-specific: any gateway provisioned by the checked-in script is affected.

The PR body, the commit message, and the `LOG_DIR_MAX_BYTES` rustdoc now all say "nodes that set `RUST_LOG`", never "every node".

**2 — release subscriber shape now covered.** Added `main_layer_still_receives_info_under_the_release_rate_limited_shape`. The hard-constraint test is now a shared helper driven by a `with_global_rate_filter` flag, so both shapes run: the plain registry (debug/test builds) and `Registry.with(<global DynFilterFn>).with(main).with(error)` (release). Non-vacuous — M1 now reddens **6** tests instead of 5.

**3 — the every-sink loss is documented.** `build_error_filter`'s rustdoc now states that because the main layer sets `moka=off` / `sqlx=error`, moka INFO and sqlx INFO under `RUST_LOG=info` previously reached the error log *and nowhere else* and now reach neither; their `WARN`+ is still retained. Called out as intended `WARN`+ semantics, not an oversight.

**4 — pin anchor hardened.** Now `concat!("mod error_filter_", "tests {")` — exact, still unable to match its own source, and it fails **closed**: added mutation **M5** renames the module and confirms the pin panics via its `expect` rather than sliding the cut point into the test module. Previously a rename could have swallowed test functions into the "production" slice.

Noted on `error_log_directives_reads_rust_log`: agreed it is near-tautological. Kept as documentation of the `from_env_lossy()` equivalence and explicitly **not** counted as coverage in the PR body.

**Rebased onto #5010** (merged since the review). One conflict in `tracer.rs`: #5010 replaced `enforce_log_dir_size_cap` with `prune_log_files`, while this commit carried the old function's text adjacent to the new helpers. Resolved by taking **main's deletion** and keeping only the two helpers. Nothing in the error-layer wiring conflicted; pin counts re-verified post-rebase (2/2/1/1). Also updated #5010's now-stale `LOG_DIR_MAX_BYTES` rustdoc — it said the duplication blocked lowering the budget — and annotated `default_budget_holds_a_day_of_a_busy_gateways_logs` to keep its 20.95 MB/h constant deliberately at the *pre-fix* rate, since the higher rate makes that guard strictly more conservative. **The 512 MiB constant itself is unchanged**; retuning it is a separate call.

Full mutation battery (M1–M5) and `cargo test -p freenet --lib` (4387 passed, 0 failed) re-run on the rebased head rather than carried over; `cargo fmt --check` and `cargo clippy -p freenet --lib --all-features` clean.

[AI-assisted - Claude]

kivancgnlp:
Nice — I converged on the same shape independently (posted a claim on #5015 not realizing this PR was already open, sorry for the noise). Yours is cleaner: taking `directives: &str` as a parameter keeps `build_error_filter` a pure function, so the tests run in parallel without the env-var mutex I ended up needing. Filing that pattern away.

PR #5027, opened 2026-07-29
fix: TokenBucketLimiter permanently refuses new keys when map is full

### LINKED ISSUE #5000: TokenBucketLimiter can permanently refuse new keys once its map is full (resync limiters)
## Problem

`TokenBucketLimiter` (`crates/core/src/ring/resync_rate_limit.rs`) has a bounded tracking map that can permanently refuse newcomers once full, with no path to recovery. This is the permanently-refreshable GC exemption `AGENTS.md` forbids, and the same shape as #4981.

Two mechanics combine:

1. **`check_and_record` conflates two outcomes.** It returns `false` both when the key is over its rate and when the key is *absent and the map is at `max_tracked`*. The caller cannot tell "this key misbehaved" from "we had no room to find out".

2. **`Bucket::refill` restamps `last_refill` on every check, including denied ones**, and `cleanup` only drops entries that are `recovered && idle` (`elapsed > CLEANUP_AGE`). So a key checked more often than every 5 minutes never ages out, regardless of whether its checks are being allowed or denied.

Net effect: once the map fills with continuously-active keys, a new key is refused **forever**. Nothing evicts to make room, and the incumbents cannot expire while they stay active.

## Scope

This was found while reviewing #4997, which added a per-`sender_addr` limiter built on this type. **That consumer is not affected** — it uses the new `check_and_record_detailed` and fails open on `Untracked`, so a full map degrades the control to off rather than refusing a peer (see `update_rate_limit.rs`, `BucketOutcome::Untracked` arm, and `the_new_pair_budget_map_is_bounded_and_fails_open_when_full`).

The three **resync** limiters still use the collapsed `check_and_record` and are untouched:

- `ResyncEmitLimiter` — keyed by `ContractInstanceId`
- `ResyncResponseLimiter` — keyed by `(SocketAddr, ContractInstanceId)`
- `ResyncResponseGlobalLimiter` — keyed by `ContractInstanceId`

## Severity: low, but it shouldn't be left looking fixed

Not reachable today. `MAX_TRACKED_KEYS` is 16,384 and nova hosts 1,962 contracts, so the per-contract maps are ~12% full; the `(peer, contract)` map is bounded by peers x contracts and is likewise far from the cap.

And the consequence is milder than #4981's: a refused key degrades resync (a delta-apply-failure recovery path) rather than dropping UPDATEs. But the failure mode is silent and permanent if it ever is reached, and the fix that landed for the UPDATE limiter makes this look addressed when it isn't.

## Suggested fix

`check_and_record_detailed` already exists and distinguishes the cases; the resync callers just need to decide what `Untracked` should mean for them. Either:

- **fail open** like the UPDATE limiter (a full map is a sizing accident, not evidence about the key), or
- **evict** to make room, which needs a recency signal the bucket doesn't currently keep honestly — `last_refill` is restamped on denials, so it tracks "last checked", not "last allowed".

If the second is chosen, note that fixing `refill` to stop restamping on denial is itself a behaviour change to the resync limiters and wants its own test coverage.

## Not blocking

Filed from the #4997 review rather than fixed there, to keep that PR to the UPDATE path. #4997 is the last gate on 0.2.113.

[AI-assisted - Claude]

waterWang:
## Problem

`TokenBucketLimiter` has a bounded tracking map that can permanently refuse newcomers once full, with no path to recovery. `check_and_record` conflates two outcomes: it returns `false` both when the key is over its rate and when the key is absent and the map is at `max_tracked`. The caller cannot tell "this key misbehaved" from "we had no room to find out".

Additionally, `Bucket::refill` restamps `last_refill` on every check including denied ones, and `cleanup` only drops entries that are `recovered && idle`. So a key checked more often than every 5 minutes never ages out.

Net effect: once the map fills with continuously-active keys, a new key is refused **forever**.

## Fix

1. **Add `BucketOutcome` enum** — three variants: `Allowed`, `RateLimited`, `Untracked` (map at capacity).
2. **Add `check_and_record_detailed` method** — returns `BucketOutcome` instead of a collapsed bool.
3. **Update all 4 resync limiter call sites** to use `check_and_record_detailed` and fail open on `Untracked` (a full map is a sizing accident, not evidence about the key).
4. **Add tests** — verify `Allowed`, `RateLimited`, `Untracked` outcomes, fail-open semantics, and backward compatibility of `check_and_record`.

## Scope

The three resync limiters are updated:
- `ResyncEmitLimiter` (2 call sites in `op_ctx_task.rs`)
- `ResyncResponseLimiter` (1 call site in `node.rs`)
- `ResyncResponseGlobalLimiter` (1 call site in `node.rs`)

The UPDATE limiter (`UpdateRateLimiter`) already has its own `RateLimitDecision` enum and is not affected.

Fixes #5000

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -3183,11 +3183,16 @@ async fn handle_interest_sync_message(
             // This is the mixed-version-rollout guard: a not-yet-upgraded peer
             // that still emits unlimited ResyncRequests must not be able to make
             // this (upgraded) node full-state-reply in a loop. When suppressed,
-            // simply don't respond — the requester will retry later.
-            if !op_manager
-                .ring
-                .resync_response_limiter
-                .check_and_record((source, *key.id()))
+            // simply don't respond — the requester will retry later. A full map
+            // (Untracked) is a sizing accident, not evidence about this key, so
+            // fail open (#5000).
+            if matches!(
+                op_manager
+                    .ring
+                    .resync_response_limiter
+                    .check_and_record_detailed((source, *key.id())),
+                crate::ring::resync_rate_limit::BucketOutcome::RateLimited
+            )
             {
                 crate::config::GlobalTestMetrics::record_resync_response_suppressed_per_peer();
                 tracing::debug!(
@@ -3204,10 +3209,14 @@ async fn handle_interest_sync_message(
             // distinct requester IPs drive ~9,733 full-state responses/day for
             // one forked contract. This bounds a single contract's total
             // resync-response cost (~12/min) regardless of requester count.
-            if !op_manager
-                .ring
-                .resync_response_global_limiter
-                .check_and_record(*key.id())
+            // A full map (Untracked) is a sizing accident, so fail open (#5000).
+            if matches!(
+                op_manager
+                    .ring
+                    .resync_response_global_limiter
+                    .check_and_record_detailed(*key.id()),
+                crate::ring::resync_rate_limit::BucketOutcome::RateLimited
+            )
             {
                 crate::config::GlobalTestMetrics::record_resync_response_suppressed_global();
                 tracing::debug!(
--- a/crates/core/src/operations/update/op_ctx_task.rs
+++ b/crates/core/src/operations/update/op_ctx_task.rs
@@ -1177,11 +1177,15 @@ async fn send_queue_full_resync_request(
     // global-cap-only, with no per-sender throttle. The global suppression records
     // the same `record_resync_request_suppressed` metric node.rs does.) On
     // suppression, CANCEL the reservation so the per-sender window is released and
-    // the sender retries as soon as global tokens refill.
-    if !op_manager
-        .ring
-        .resync_emit_limiter
-        .check_and_record(*key.id())
+    // the sender retries as soon as global tokens refill. A full map (Untracked)
+    // is a sizing accident, so fail open (#5000).
+    if matches!(
+        op_manager
+            .ring
+            .resync_emit_limiter
+            .check_and_record_detailed(*key.id()),
+        crate::ring::resync_rate_limit::BucketOutcome::RateLimited
+    )
     {
         op_manager
             .interest_manager
@@ -1930,11 +1934,15 @@ async fn drive_relay_broadcast_to(
                 // full-state resync storm. When the emit is suppressed, skip the
                 // sender summary-clear too — it is part of the resync handshake,
                 // so clearing it without emitting would just force the sender to
-                // full-state us on the next interest cycle.
-                if op_manager
-                    .ring
-                    .resync_emit_limiter
-                    .check_and_record(*key.id())
+                // full-state us on the next interest cycle. A full map
+                // (Untracked) is a sizing accident, so fail open (#5000).
+                if !matches!(
+                    op_manager
+                        .ring
+                        .resync_emit_limiter
+                        .check_and_record_detailed(*key.id()),
+                    crate::ring::resync_rate_limit::BucketOutcome::RateLimited
+                )
                 {
                     tracing::warn!(
                         tx = %incoming_tx,
--- a/crates/core/src/ring/resync_rate_limit.rs
+++ b/crates/core/src/ring/resync_rate_limit.rs
@@ -99,6 +99,25 @@ impl Bucket {
     }
 }
 
+/// Outcome of a token-bucket rate-limit check. Distinguishes a genuine
+/// rate-limit hit from a capacity rejection so callers can decide whether
+/// to fail open or fail closed.
+///
+/// - [`Allowed`](BucketOutcome::Allowed): a token was consumed — proceed.
+/// - [`RateLimited`](BucketOutcome::RateLimited): the key is over its rate
+///   — the caller should throttle.
+/// - [`Untracked`](BucketOutcome::Untracked): the key was absent and the
+///   tracking map is at its [`MAX_TRACKED_KEYS`] cap. A full map is a
+///   sizing accident, not evidence about this key. Callers should **fail
+///   open** (treat as allowed) unless there is a specific security reason
+///   to drop traffic from unknown keys.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum BucketOutcome {
+    Allowed,
+    RateLimited,
+    Untracked,
+}
+
 /// A token-bucket rate limiter keyed by `K`. One token per key is consumed per
 /// allowed event; the bucket holds up to `capacity` and refills one token per
 /// `refill_interval`.
@@ -166,6 +185,45 @@ impl<K: Eq + Hash + Clone> TokenBucketLimiter<K> {
                 });
                 self.allowed_total.fetch_add(1, Ordering::Relaxed);
                 true
+            }
+        }
+    }
+
+    /// Check-and-consume one token, returning a [`BucketOutcome`] that
+    /// distinguishes rate limiting from capacity rejection.
+    ///
+    /// Callers that want to fail open on a full map should gate on
+    /// `matches!(outcome, BucketOutcome::RateLimited)` rather than
+    /// inverting the bool from [`check_and_record`](Self::check_and_record).
+    pub fn check_and_record_detailed(&self, key: K) -> BucketOutcome {
+        let now = self.time_source.now();
+        use dashmap::mapref::entry::Entry;
+        match self.buckets.entry(key) {
+            Entry::Occupied(mut occ) => {
+                let bucket = occ.get_mut();
+                bucket.refill(now, self.capacity, self.refill_interval);
+                if bucket.tokens >= 1.0 {
+                    bucket.tokens -= 1.0;
+                    self.allowed_total.fetch_add(1, Ordering::Relaxed);
+                    BucketOutcome::Allowed
+                } else {
+                    self.suppressed_total.fetch_add(1, Ordering::Relaxed);
+                    BucketOutcome::RateLimited
+                }
+            }
+            Entry::Vacant(vac) => {
+                let prev = self.size.fetch_add(1, Ordering::Relaxed);
+                if prev >= self.max_tracked {
+                    self.size.fetch_sub(1, Ordering::Relaxed);
+                    self.suppressed_total.fetch_add(1, Ordering::Relaxed);
+                    return BucketOutcome::Untracked;
+                }
+                vac.insert(Bucket {
+                    tokens: self.capacity - 1.0,
+                    last_refill: now,
+                });
+                self.allowed_total.fetch_add(1, Ordering::Relaxed);
+                BucketOutcome::Allowed
             }
         }
     }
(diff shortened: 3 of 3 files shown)


sanity:
Thanks for this, and apologies — four weeks with no reply on a good-faith fix to a real bug is not acceptable and I'm sorry it happened.

Reviewed properly now. The diagnosis in #5000 is right and worth fixing, but I don't think fail-open is the right resolution, and there are a couple of concrete breakages. Details below.

## 1. Fail-open converts "map full" into a total bypass of the storm guard

`crates/core/src/ring/resync_rate_limit.rs:38-46` records the current behaviour as a deliberate decision from the #4864 review, not an oversight:

> Under extreme `(peer, contract)` churn a NEW key at capacity is denied (`check_and_record` returns `false`), which for the responder means "don't send this response" — a safe, conservative outcome, not a correctness bug.

These limiters exist to cap the #4861 full-state storm, where ~45 distinct requester IPs drove ~9,733 full-state responses in a day for one forked contract. `resync_response_limiter` is keyed by `(SocketAddr, ContractInstanceId)`, and either dimension is cheap for a remote peer to vary. Once 16,384 slots are occupied, every subsequent *new* key is `Untracked` and, under this PR, allowed — so filling the map turns the responder-side guard off for everyone new. That trades a liveness problem (newcomers refused) for an availability problem (the amplification cap the limiter was built for stops applying), which I think is the worse of the two.

The module doc is also left contradicting the code, which the rule review flagged as a warning.

## 2. The root cause named in the description isn't actually fixed

The PR body identifies it correctly:

> `Bucket::refill` restamps `last_refill` on every check including denied ones, and `cleanup` only drops entries that are `recovered && idle`. So a key checked more often than every 5 minutes never ages out.

Neither is touched by the diff. So the map still fills permanently with live keys; the PR only changes what happens once it *is* full. "No path to recovery" from #5000 is still there.

Worth saying: fixing the aging directly would itself be unsafe. Dropping an actively-denied key hands it a fresh full bucket on its next check, which is exactly what a rate limiter must not do. So the recovery path genuinely has to come from somewhere other than the TTL sweep.

## 3. Suggested direction — an overflow bucket

Keep the strict `MAX_TRACKED_KEYS` cap, but on the vacant-at-capacity path charge the event against a **single shared token bucket per limiter** instead of allowing it unconditionally. Memory stays bounded (one extra bucket), the rate stays bounded, and no amount of key churn can disable the guard — newcomers get throttled as a group rather than refused forever. `BucketOutcome` is still the right enum for it; only the meaning of the third variant changes, from "caller decides" to "charged to the overflow bucket".

Two cheaper options if that's more than you want to build:

- `MAX_TRACKED_KEYS = 16_384` is ~1 MB by the module's own estimate (`resync_rate_limit.rs:81-82`). Raising it is nearly free and pushes the corner far out without changing any semantics.
- On a full map, evict the entry closest to fully refilled rather than refusing the newcomer — that entry is the one whose removal loses the least information.

## 4. This breaks an existing pin test

`crates/core/src/operations/update/op_ctx_task.rs:4011`:

```rust
let global_cap = helper.find(".check_and_record(*key.id())").expect(
    "helper must ALSO gate on the global per-contract emit cap \
     (resync_emit_limiter.check_and_record, #4864 round-4)",
);
```

The diff renames that call site to `.check_and_record_detailed(*key.id())`, so the `find` returns `None` and the test panics. The pin is load-bearing — it asserts the global emit cap is ordered *before* the emit so an unthrottled emission can't reopen the #4251 storm — so it needs updating to the new literal, not removing.

## 5. `check_and_record` is now a second copy of the bucket accounting

`check_and_record_detailed` duplicates the whole body rather than the old method delegating to the new one. Two hand-maintained copies of token accounting will drift. Suggest:

```rust
pub fn check_and_record(&self, key: K) -> bool {
    matches!(self.check_and_record_detailed(key), BucketOutcome::Allowed)
}
```

## 6. `suppressed_total` now counts events that were sent

The vacant-at-capacity arm bumps `suppressed_total` and returns `Untracked`, but every caller treats `Untracked` as allowed. So the counter stops meaning "events we suppressed". `Untracked` should get its own counter. (There's a repo rule about exactly this shape — a metric describing a filtering decision has to be produced by the code that makes the decision, not inferred alongside it.)

## Two process notes

- The CLA check is red: *"waterWang seems not to be a GitHub user"* — the commit email isn't linked to a GitHub account. That needs resolving before anything can merge, independently of the code.
- The main CI suite has never run on this branch (outside-contributor runs need maintainer approval), which is part of why #4 above went unnoticed. Once the branch is updated we'll get the runs approved.

Genuinely a real bug you found, and the `BucketOutcome` split is the right primitive — it's the fail-open default at the four call sites that I'd like to change before this lands. Happy to keep iterating.

[AI-assisted - Claude]

sanity:
@waterWang — I need to correct part of my review. I checked the wrong thing first, and two of my points were unfair to you.

**Your PR overlaps an existing unmerged PR: #4997**, opened 2026-07-27 by @sanity, two days before yours. It touches the same file and adds the same primitives — `BucketOutcome` (with `Untracked`), `check_and_record_detailed`, and `check_and_record` reduced to `matches!(self.check_and_record_detailed(key), BucketOutcome::Allowed)`. You had no way to know; it isn't linked from #5000 and it has sat untouched since 2026-07-28 with a clean merge state.

That changes three things I said:

- **My "make `check_and_record` delegate" point was already done in #4997.** Not an omission of yours to fix.
- **My "`Untracked` needs its own counter" point is also already done there** — `new_pair_budget_untracked_total`, a dedicated counter rather than folding into `suppressed_total`.
- **I framed fail-open as your own questionable choice. That was wrong of me.** Issue #5000's "Suggested fix" explicitly offers it as one of two options ("fail open like the UPDATE limiter... or evict to make room"), and #4997's UPDATE consumer does exactly that. You followed a documented suggestion.

I should also retract the shape of my "you didn't fix the root cause" point. It's true of this PR in isolation, but #4997 *does* fix the `Bucket::refill` restamp — its own comment names it: *"`Bucket::refill` restamps on every check, so an active sender's entry never ages out and the newcomer would never recover."* So the fix exists; it just hasn't landed.

## What actually survives

Two things, and they're the real content:

1. **What `Untracked` should mean for the three *resync* callers specifically** — which #5000 deliberately leaves open. My concern isn't fail-open as a principle; it's fail-open on `resync_response_limiter` in particular. That one is keyed by `(SocketAddr, ContractInstanceId)` and is the #4861 amplification guard (~45 requester IPs, ~9,733 full-state responses/day for one forked contract). A full map there means the guard is off for every new key, which is a different risk profile from #4997's new-pair budget. That's worth deciding deliberately rather than inheriting.
2. **The pin test at `op_ctx_task.rs:4011` still breaks** on the rename. That one is unchanged.

## Sequencing

The sensible order is #4997 first — it delivers the enum, the delegation, the dedicated counter and the refill fix — after which your PR reduces to the small, genuinely-open decision about the three resync call sites. That is a better PR than this one and a much smaller one.

I'm leaving this open rather than closing it. The CLA still blocks merging either way, but I'd rather not close a PR that found a real problem. **Thank you** — the `Untracked` distinction is the right primitive, you identified the `refill`/`cleanup` interaction precisely in the description, and the fact that the project lead independently reached the same design two days earlier is a point in your favour, not against you. I've updated #5000 with all of the above so this doesn't get lost, and flagged #4997's four-week stall internally.

Sorry for the four weeks of silence, and sorry for the parts of my first review that put this on you.

[AI-assisted - Claude]

sanity:
@waterWang — the answer on #4997 is in, and I owe you one more correction plus a clearer verdict. Both checked against the code this time rather than inferred.

## Correction: I was wrong that #4997 fixes the `refill` restamp

In my last comment I said #4997 fixes the `Bucket::refill` restamp, so "the fix exists; it just hasn't landed". **That is wrong.** I over-read a comment in its `update_rate_limit.rs` hunk that describes the starvation it removes *there* (by LRU-evicting oldest pairs), and took it for a fix to `refill` in `resync_rate_limit.rs`.

Verified: #4997's diff touches neither `fn refill`, nor `fn cleanup`, nor the `recovered && idle` predicate. **The restamp you identified in your Problem section is still open and still unfixed by anything.** Your description of it stands as the accurate account. Sorry — that is twice now I have corrected myself at your expense, and both times because I described a diff I had not read closely enough.

## The verdict: #4997 subsumes your plumbing, but NOT your actual fix

**Redundant once #4997 lands:** `BucketOutcome{Allowed, RateLimited, Untracked}`, `check_and_record_detailed`, and `check_and_record` reduced to a delegate. Near-identical implementations to yours.

**Not subsumed, and this is the part that matters:** #4997 does not touch `op_ctx_task.rs` at all, and its `node.rs` hunk is the UPDATE decision match, not the resync limiters — I checked, it changes zero resync call sites. It says so on purpose: *"check_and_record keeps the collapsed form so existing callers are unchanged."*

**So after #4997 lands, all three resync limiters still fail closed at a full map, and #5000's resync-path starvation is still live.** Converting those four call sites is your change and nothing else does it.

## The thing worth doing beyond a rebase: argue fail-open per limiter

My earlier objection was too blunt — "fail-open is wrong" — and #4997's own rationale gives the sharper version. It distinguishes callers "whose keys are NOT attacker-chosen". Applying that test to the three (`crates/core/src/ring/resync_rate_limit.rs:221-225`):

```rust
pub(crate) type ResyncEmitLimiter           = TokenBucketLimiter<ContractInstanceId>;
pub(crate) type ResyncResponseLimiter       = TokenBucketLimiter<(SocketAddr, ContractInstanceId)>;
pub(crate) type ResyncResponseGlobalLimiter = TokenBucketLimiter<ContractInstanceId>;
```

A `ContractInstanceId` **is** attacker-choosable — anyone can mint contracts. So for the emit limiter and the global responder limiter, failing open means an attacker churning contract ids bypasses the guard entirely, which is the #4861-class amplification those two exist to bound. `ResyncResponseLimiter` is the strongest case for fail-open, since the `SocketAddr` half is not freely chosen — though the contract half still is.

That is a per-limiter argument, and it may well land differently for each of the three.

## Concrete ask

1. **#4997 lands first** (it is @sanity's, clean, and has been open since 27 July — we are unsticking it).
2. **Rebase this down to the call-site change only** — drop the duplicated `BucketOutcome`, `check_and_record_detailed`, and the delegate, since #4997 brings them.
3. **Justify `Untracked` per limiter rather than blanket.** The emit and global limiters need a different argument than the per-peer one, or a different mechanism (the shared overflow bucket I suggested earlier is one option that bounds rate without a bypass).
4. Update the module doc at `resync_rate_limit.rs:38-46`, which still records the fail-closed choice as deliberate, and fix the pin at `op_ctx_task.rs:4011`.

That is a much smaller and better-argued diff than the current one, and it fixes something #4997 provably does not.

## To be clear about what was yours

Your diagnosis was right, the `refill`/`cleanup` interaction in your Problem section is an accurate description of an *unfixed* bug, and fail-open was never your invention — #5000's own "Suggested fix" offers it as one of two options. The remaining gap is real and it is yours to close if you want it. The CLA is still the one hard blocker on merging anything here.

[AI-assisted - Claude]

sanity:
Status update, so this isn't left ambiguous again.

**#4997 merged on 2026-08-25.** That is why this branch is now conflicted. As predicted in the earlier comment, it took `BucketOutcome`, `check_and_record_detailed`, and the `check_and_record` delegate with it — so that plumbing is already on `main` and no longer needs this PR.

**What is still unique to your work, and still unfixed on main:** the `Bucket::refill` restamp. `refill` restamps `last_refill` on every check including denied ones, and `cleanup` only drops entries that are `recovered && idle`, so a key checked more often than every 5 minutes never ages out and the map fills permanently. #4997 touched neither `refill` nor `cleanup` nor the `recovered && idle` predicate. Your Problem section remains the accurate description of that defect, and it is the part worth landing.

**The blocker is not technical.** `license/cla` is pending: the CLA assistant reports the commit email on this branch isn't associated with a GitHub account, so it cannot match a signature to it. Nothing can merge until that is resolved — please either sign at the link in the CLA comment above, or add the commit's email address to your GitHub account and re-run the check.

If you'd rather not carry this further, say so and I'll pick the `refill`/`cleanup` fix up separately with credit to you and to #5000 — no hard feelings either way, and the analysis was good.

To be concrete about what a rebase would need to look like: drop the `BucketOutcome` / `check_and_record_detailed` additions (now on main), keep the `refill` restamp and `cleanup` predicate fix, and re-examine the fail-open decision separately — it remains genuinely contested, since `resync_response_limiter` is keyed by `(SocketAddr, ContractInstanceId)` and both dimensions are cheap for a remote peer to vary, so failing open at capacity is a bypass of the #4861 storm guard rather than a neutral default. That question does not need to block the `refill` fix, which is uncontroversial.

[AI-assisted - Claude]

PR #5029, opened 2026-07-29
fix: bound wasmtime compile cache by the disk budget it is charged to

### LINKED ISSUE #5014: wasmtime compile cache is charged against the DISK budget but bounded by a RAM-derived limit — disk-tight/RAM-rich hosts are unprotected and can wedge admission
## Problem

The wasmtime on-disk compile cache is **charged against the aggregate disk
budget**, but the soft limit that bounds it is derived from **RAM**. Those are
different resources, so on a disk-tight-but-RAM-rich host the bound provides no
protection at all.

### The charging path (disk)

`DiskUsageTracker::total_bytes()` (`crates/core/src/ring/hosting/disk_usage.rs:161-166`)
sums three consumers:

```rust
self.state_bytes.load(Ordering::Relaxed)
    .saturating_add(self.wasm_bytes.load(Ordering::Relaxed))
    .saturating_add(self.compile_cache_bytes.load(Ordering::Relaxed))
```

`compile_cache_bytes` is a `du`-walk of the cache directory, seeded at
`disk_usage.rs:255` and refreshed by `refresh_compile_cache()` at
`disk_usage.rs:528`.

That total is what gates admission:

- `admit_state_write` (`disk_usage.rs:424-447`) — `projected = total_bytes() − old + new`, rejects when `projected > budget_bytes`
- `admit_wasm_write` (`disk_usage.rs:501-515`) — `projected = total_bytes() + blob_len`

and it feeds the eviction floor `effective_budget = min(ram_budget, disk_budget)`
(`ring/hosting.rs::recompute_effective_budget`).

### The bounding path (RAM)

`Config::wasmtime_cache_size_bytes` resolves from
`wasm_runtime::runtime::default_wasmtime_cache_size_bytes()`, which is
`clamp(read_total_ram_bytes() / 8, 128 MiB, 512 MiB)` — a **memory** signal.
(Before #5011 it was a flat 512 MiB, which had the same mismatch and no scaling
at all.)

### Worked example

A 16 GiB VM with a small data disk — say 400 MiB free on the data-dir mount:

1. RAM is ample, so the compile cache resolves to its 512 MiB ceiling.
2. The disk budget is `clamp(0.5 × (freenet_used + available), 128 MiB, 32 GiB)`.
   With freenet occupying ~512 MiB (mostly compile cache) and 400 MiB available,
   that is `0.5 × 912 MiB ≈ 456 MiB`.
3. `total_bytes()` already includes the ~512 MiB of compile cache, so
   `total_bytes() > budget_bytes` before a single byte of new state is
   considered.
4. **Every** `admit_state_write` and `admit_wasm_write` now rejects. The node
   cannot accept a PUT, cannot store a WASM blob, and cannot make progress.
5. Hosting eviction cannot dig out: the sweep sheds **contract state**, and
   `compile_cache_bytes` is not something it can reclaim. Wasmtime prunes the
   cache on its own ~1h cleanup schedule, to 70% of the soft limit
   (`default_files_total_size_limit_percent_if_deleting() == 70`,
   `wasmtime-internal-cache/src/config.rs:219`) — which is still ~358 MiB and
   still over budget.

The node wedges on admission while the thing consuming the budget is outside the
eviction loop's reach.

## Why the current bound does not help

A RAM-derived clamp only tracks disk pressure by coincidence. The three shapes:

- **RAM-poor, disk-rich** — bound is tight, disk has room. Over-conservative but safe.
- **RAM-rich, disk-rich** — bound at ceiling, plenty of disk. Fine.
- **RAM-rich, disk-tight** — bound at ceiling, no disk. **Unprotected.** ← the failure

## Suggested direction

Size the compile-cache soft limit against the **disk** signal, or against
`min(ram_signal, disk_signal)`. `disk_budget_for_clamped(freenet_used, available,
pct, min, max)` already exists in `ring/hosting/cache.rs:210-240` and already does
the `used + available` basis, saturating conversion, and clamp handling.

Complications worth thinking through before implementing:

- **Ordering / bootstrapping.** The runtime config (and therefore the wasmtime
  `Cache`) is built in `Executor::from_config_with_shared_modules`, before the
  `DiskUsageTracker` is seeded. A disk-derived limit needs either a pre-seed
  `statvfs` read of the data-dir mount or a deferred re-application of the limit.
- **Wasmtime applies the limit once at `Cache::new`.** There is no live
  re-application path today, so a limit that should track a changing disk budget
  needs a mechanism (rebuild the `Cache`, or accept a start-time-only value).
- **Double-counting.** If the compile cache is bounded by a fraction of the disk
  budget AND charged against it, the interaction should be stated explicitly so
  the two do not fight (e.g. carve the compile-cache allowance out of the budget
  before the state gate sees it).
- **The 70% prune factor** means the steady-state footprint is `soft_limit × 0.7`,
  not `soft_limit`. Any budget arithmetic should use the steady-state figure, and
  the ~1h cleanup cadence means the cache can sit at the full soft limit for up to
  an hour after a burst.
- **Interaction with `min(ram_budget, disk_budget)`.** Shrinking the compile cache
  raises the disk budget, which raises the effective hosting budget — a feedback
  loop worth checking for oscillation.

## Evidence (production gateway, nova, 2026-07-28)

Measured directly from `/home/freenet/.local/share/freenet/wasmtime-cache`:

- **418 cache artifacts**, 198.8 MiB total (plus 417 tiny `.stats` sidecars, which
  the `du`-walk also counts)
- mean 487 KiB, p50 397 KiB, p90 811 KiB, max 1.70 MiB
- 185 live `*.wasm` blobs in the contracts dir — entries are keyed per distinct
  WASM blob (engine config + bytes hash), and outlive blob deletion, so entry
  count exceeds live-blob count

So on a busy gateway the compile cache is a ~200 MiB line item in
`total_bytes()` — not a rounding error against a disk budget that floors at
128 MiB.

## Context

Split out of #5011 (which made the limit node-relative rather than flat). #5011
narrows the exposure on RAM-poor hosts but does not address the RAM-vs-disk
mismatch, which is the more fundamental problem.

Related: #4683 (aggregate disk budget), #4702 (live recompute + admission gate).

[AI-assisted - Claude]

sanity:
## Problem

The wasmtime on-disk compile cache is **charged against the aggregate disk
budget**, but the soft limit that bounded it was derived from **RAM**.

`DiskUsageTracker::total_bytes()` sums state + wasm + `compile_cache_bytes`, and
that total gates `admit_state_write` / `admit_state_update` / `admit_wasm_write`
and feeds the eviction floor `min(ram_budget, disk_budget)`. The limit came from
`default_wasmtime_cache_size_bytes()` = `clamp(total_ram / 8, 128 MiB, 512 MiB)`
(#5011) — a memory signal — so a RAM-rich but disk-tight host got no protection
at all.

That mismatch is worse than a merely loose bound, because the compile cache is
the one consumer of the disk budget the hosting sweep **cannot reclaim**: the
sweep sheds contract *state*, and wasmtime owns its cache directory. Once the
cache alone pushes `total_bytes()` past `budget_bytes`, every admission rejects
and eviction cannot dig out no matter how much state it sheds. Wasmtime's own
cleanup does not rescue it either — it prunes only to 70 % of the soft limit.

The worked example from #5014: a 16 GiB VM with 400 MiB free on the data mount
resolved the cache to its 512 MiB ceiling while the disk budget was
`0.5 × (512 MiB + 400 MiB) ≈ 456 MiB`. `total_bytes() > budget_bytes` before a
single new byte was considered, so the node could not accept a PUT, could not
store a WASM blob, and could not recover.

## Solution

Two halves. The bound stops the wedge from forming; the startup prune is what
makes it reach a node that is already in it.

### 1. Bound the limit by the budget it is charged to

Resolve the soft limit as **`min(ram_term, disk_budget / 4)`**.

- **`ram_term`** is unchanged (`clamp(total_ram / 8, 128 MiB, 512 MiB)`) and
  survives as the **upper** bound — a large disk must not license an unbounded
  cache, and the useful size of this cache is set by the distinct-contract-blob
  population the RAM-scaled hosting budget already bounds.
- **`disk_budget / 4`** is the new **binding** term on a disk-tight host.

A quarter is a *structural share*, not a fitted constant: a divisor of 1 lets the
cache claim the entire budget, which is the defect; any divisor ≥ 2 buys the
property that it never holds the majority. 4 picks a comfortable margin inside
that, and with wasmtime's 70 % prune the steady state is 17.5 %.

**What the cap does and does not guarantee.** It bounds *this cache's*
contribution, so `total_bytes() > budget` can no longer be true before a single
new byte is considered, and eviction always has something it can shed. It is
**not** a bound on the cache's instantaneous footprint (wasmtime writes the
artifact first and only then posts the cleanup notification, behind an hourly
lock, so the real bound is `0.7 × L + bytes written since the last cleanup`), and
it is **not** a guarantee that the other 75 % of the budget is reclaimable —
redb dead space inside the storage term is reclaimable only by the restart
compaction, not by eviction. An earlier revision of this description claimed
both; neither is true, and the rustdoc now says so.

### 2. Prune an oversized cache at startup

Lowering the soft limit does not shrink an existing cache. Wasmtime reads the
limit in exactly one place — its cleanup pass — and that pass is reachable **only
from the cache-WRITE path** (`handle_on_cache_update`), gated behind a
once-per-`cleanup_interval` (1 h default) filesystem lock. `handle_on_cache_get`
never cleans up. A node with a stable contract-blob set performs only cache hits
after a restart, so a cache written under the old, larger limit persists
indefinitely.

On the wedged node that is self-sustaining: its oversized cache keeps
`total_bytes()` over budget, `admit_wasm_write` rejects, so it cannot take on the
new contracts whose compiles would produce the cache misses that would trigger a
cleanup. Without a prune this PR would prevent the wedge on future starts and
leave an already-wedged node wedged — i.e. not fix the node in the issue.

`ring::prune_compile_cache_to_limit` runs in `RuntimePool::new`, before the first
`Cache::new`, and deletes oldest-first down to wasmtime's own 70 %-of-limit
target. Safe by construction: the cache is a pure derived artifact, entries are
read with `fs::read` into a `Vec` (never mmap'd), the tree is `create_dir_all`'d
again on the next write, and wasmtime's own worker already does `remove_file` /
`remove_dir_all` on the same tree. Oldest-first rather than a wipe because a wipe
is equally safe but pays a whole recompile wave on every tight-budget restart.

This is also what makes "restart the node" a real remedy for the residual hole
below, rather than a hope that wasmtime's write-path cleanup happens to fire.

### Nothing that works today gets smaller

The disk term only becomes the smaller of the two when
`disk_budget < 4 × ram_term`, i.e. Freenet-reachable capacity on the data mount
under ~4 GiB at the 512 MiB ceiling. Resolved values (using the real artifact
sizes from #5011's review: 418 artifacts, 198.8 MiB, mean 487 KiB, p50 397 KiB,
p90 811 KiB; steady state is 0.7 × the limit):

| Host | disk budget | limit | vs. #5011 |
|---|---|---|---|
| 2 GiB-cgroup laptop, large disk | 32 GiB (capped) | **256 MiB** | unchanged |
| 16 GiB VM, 400 MiB free, cache-dominated usage | ~456 MiB | **114 MiB** (~80 MiB steady, ~100 artifacts at p90) | 512 MiB → 114 MiB |
| 125 GiB server, 3.7 TB disk | 32 GiB (capped) | **512 MiB** | unchanged |

Only the disk-starved shape moves — the one that could not accept writes at all.
The middle row reproduces at 114 MiB because the issue's node had essentially all
of its ~512 MiB of Freenet usage in the compile cache; a node with the same free
space but that usage split as contract state resolves lower, because the start-
time estimate omits state (see below).

The floor of the whole composition is 32 MiB (a quarter of the disk budget's own
128 MiB MIN clamp), roughly 40 artifacts at the measured p90 and 67 at the mean.
It binds when `wasm + compile_cache + free ≤ ~256 MiB` at the default pct — note
that condition is about the terms the estimate *measures*, not about total
Freenet-reachable disk, so a node holding a lot of state can reach the floor with
more capacity than that.

### The three complications #5014 named

**Ordering.** Wasmtime fixes the soft limit once, at `Cache::new`, reached from
`from_config_with_shared_modules` — before `Ring::set_hosting_disk_paths`
configures the `DiskUsageTracker` and long before the first sweep seeds it. So
the budget cannot be read from the tracker. `ring::startup_disk_budget_estimate`
re-derives it from the filesystem, reusing the live budget's own
`disk_budget_for_clamped` math, the same `used + available` basis, the same MIN
floor, and the operator's `--hosting-disk-pct` / `--max-hosting-disk`. Its
`used` term omits persisted contract-state bytes (unreadable before storage
exists), and the math is monotone non-decreasing in that term, so the estimate is
always `≤` the budget the first 60 s recompute installs — the cache therefore
claims at most a quarter of the *live* budget too, never more.

That omission is **not immaterial**, and it bites hardest on exactly the
disk-tight hosts the disk term exists for. A 16 GiB VM holding 1 GiB of state,
50 MiB wasm, a 50 MiB cache and 200 MiB free has a live budget of ~662 MiB
(live-consistent disk term ~165 MiB), while the estimate sees `50 + 50 + 200 =
300 MiB` and resolves ~37 MiB — about 4.5× under. The direction is always safe
(bounded more tightly than the live budget licenses, never less), but the
rustdoc no longer claims the two agree. The gap is the redb *row* total, not the
database *file*, whose size is a plain `metadata()` read; folding it in is left
to the follow-up that lands `du_walk_shallow` (#5033), which is where that walker
comes from.

The limit is resolved **once per pool** in `RuntimePool::new` and threaded into
every executor, so the data-mount walk does not repeat `pool_size` times and does
not re-sample a cache dir the first executor's engine has begun writing into. The
walks and the prune run on `spawn_blocking`, matching the discipline the
identical walks follow in the 60 s sweep.

**Double-counting / feedback.** There is no loop, and the reason is worth stating
because it is not obvious. The budget's basis is `freenet_used + available`, not
free space: every byte the cache writes moves one byte from `available` into
`freenet_used` on the same mount, so the basis is the mount's Freenet-reachable
capacity and is **invariant under the cache's own size**. This is exactly why
the start-time estimate must include the compile-cache `du` walk — an estimate
built on bare free space would shrink when the cache is full and grow when it is
empty, re-deriving a different limit on every restart and oscillating
(50 → 109 → 100 MiB on the worked example).

For the same reason, the interaction with `min(ram_budget, disk_budget)` is
stable: shrinking the cache does **not** raise the disk budget (the basis does
not move) — it lowers `total_bytes()`, creating headroom under an unchanged
budget. Nothing oscillates.

**No live re-application — the residual hole, stated plainly.** Wasmtime offers
no way to change the limit on a live `Cache`; rebuilding one would mean tearing
down the backend engine every pool executor shares and discarding its compiled
modules. So the limit `L` is fixed for the life of the process while the live
budget is recomputed every 60 s. Work the algebra at the margin: after the sweep
has shed everything it can, the node is over budget iff `L > pct × (L +
available)`, i.e. at the default pct = 0.5 iff `L > available`. **Any consumer
outside Freenet that drives free space on the data mount below the cache's size
re-opens the #5014 wedge for the rest of that process's life.** An earlier
revision described the 60 s recompute as covering this case; the recompute
shrinking the budget *is* the mechanism of the hole, not a mitigation of it.

Three things bound it, and they are why this is deferred rather than fixed here:
Freenet's own writes cannot reach it (the basis is invariant under them, so it
takes an external mount-filler); `HostingManager::recompute_effective_budget` now
warns, edge-triggered, the first time `compile_cache_bytes` exceeds a quarter of
the live budget, so the condition is no longer silent; and a restart is now a
real remedy, because it re-derives the limit *and* prunes the existing cache down
to it.

### Layering

`default_wasmtime_cache_size_bytes` now takes the resolved `disk_budget` as one
`u64` instead of four hosting knobs, and `RuntimePool::new` calls
`ring::startup_disk_budget_estimate` itself. `ring` already depends on
`wasm_runtime` (`read_total_ram_bytes`, `ModuleCacheMetrics`), so the reverse
edge closed a module cycle; and the pool already passes those same four values to
`Ring::set_hosting_disk_paths` a few lines later, so the knowledge was already
sitting there.

### Hosting invariants

Invariant 3 (`.claude/rules/hosting-invariants.md`) is the only one touched, and
in the direction of restoring it. The eviction ranking, the demand signal, and
the budget formula are all unchanged. What changes is that a **non-reclaimable**
consumer of that budget is now bounded by it, so the budget the sweep defends
always has reclaimable headroom to work with. Before this change the sweep could
shed every contract it held and still be over budget — its constraint was
unsatisfiable. Invariants 1, 2, 4 and 5 are untouched (no placement, freshness,
or routing change).

Note the framing correction the reviews forced: the compile cache is **not**
genuinely non-reclaimable — `rm -rf` on that tree is safe at any time. Its
non-reclaimability was a property of the code (the sweep does not touch it), not
of the artifact, which is what makes the startup prune available and cheap.

## Testing

`wasm_runtime::runtime::wasmtime_disk_cache_sizing_tests`
- `resolved_limit_never_claims_more_than_its_share_of_the_disk_budget` — the
  invariant, over the RAM × disk-budget cross product including `u64::MAX`. The
  `× 4` is a literal, not the divisor constant, so it still fails if the divisor
  is mutated to 1.
- `disk_term_is_inert_on_hosts_with_room` — the three machine shapes, two of
  which must be bit-identical to #5011, plus the exact crossover point.
- `disk_term_may_fall_below_the_ram_floor` — the RAM floor must NOT be
  re-imposed after the `min`; re-imposing it restores the defect.
- `disk_term_boundaries_and_monotonicity`, `resolved_limit_is_the_smaller_of_the_two_terms`
  — 0, 3, 4, `u64::MAX`, monotonicity, no overflow.
- `default_soft_limit_matches_the_pure_clamp_of_this_hosts_ram_signal` — now
  pure (the resolver no longer touches the filesystem), so its concrete
  assertion is exact on every host.

`ring::hosting::disk_usage::startup_estimate_tests`
- `estimate_is_invariant_under_cache_growth_on_the_same_mount` (anti-feedback),
  `estimate_is_monotone_in_measured_used` (why omitting state is safe),
  `unreadable_free_space_clamps_to_the_cap_not_to_zero`,
  `estimate_boundaries_stay_inside_the_clamp`,
  `measured_used_counts_wasm_blobs_and_the_whole_compile_cache`,
  `measured_used_is_zero_when_nothing_exists_yet`.
- `prune_deletes_oldest_first_down_to_the_target`,
  `prune_leaves_a_cache_that_already_fits_untouched`,
  `prune_boundaries_are_safe` (missing dir, zero limit, `u64::MAX` target
  overflow), `prune_target_matches_wasmtimes_own_cleanup_target`. mtimes are set
  explicitly so the outcome cannot depend on how fast the host writes files or on
  directory iteration order.
- **Removed:** `estimate_composes_the_du_walk_and_the_clamp_on_a_real_dir` was
  vacuous. It ran at `pct = 0.0`, which multiplies the du-walk result by zero, so
  deleting both `write_file` calls changed nothing; and its second assertion was
  a containment check on a function that clamps into exactly that interval by
  construction. This PR calls out that same anti-pattern elsewhere and then
  reproduced it. Replaced by
  `estimate_composes_the_measured_walk_and_the_free_space_read` (below) plus a
  slim `estimate_on_real_dirs_resolves_the_min_floor_at_pct_zero` that claims
  only what it tests.

`ring::hosting::tests`
- `compile_cache_over_share_warning_is_edge_triggered` — the boundary does not
  warn, crossing up warns once, staying over does not re-warn, recovery re-arms.
  Asserts the latch transitions rather than captured log lines (subscriber-
  capture tests are flaky here); the latch *is* the emit decision.
- `recompute_effective_budget_runs_the_over_share_check`.

Call-site pins (the wiring is where this fix actually lives, and it is invisible
to every test of the pure math):
- `runtime_pool_resolves_compile_cache_limit_from_the_disk_signal` — the disk
  knob bindings, the argument ORDER into the estimate (both dirs are `&Path`, so
  a swap compiles), the resolver applied to that budget, the prune call, one call
  site, the resolution POSITION relative to the first executor, and — the edge
  the first round missed — that **both** `from_config_with_shared_modules` call
  sites receive the resolved value.
- `estimate_composes_the_measured_walk_and_the_free_space_read` — the other edge
  the first round missed: that `startup_disk_budget_estimate` actually composes
  `measure_startup_disk_used(contracts_dir, compile_cache_dir)` and
  `available_bytes(contracts_dir)`.
- `from_config_with_shared_modules_wires_offload_and_budget`,
  `default_soft_limit_reader_derives_from_ram_and_disk_signals` (extended to pin
  that `disk_budget` reaches the pure math and that the resolver does not reach
  into `ring`).

All region delimiters in the pins this PR touches now use
`split_once(..).expect(..)` rather than `split(..).next().expect(..)`, whose
`expect` is unreachable — a reworded end marker used to widen the region silently
to end-of-file, `mod tests` included, while the assertions kept passing for the
wrong reason. `runtime_pool_sizes_caches_by_byte_budget`'s fixed 110-chunk window
is likewise replaced by the function's end marker; the count version stopped
covering the cache construction as soon as anything was added to the preamble.

### Mutation testing

Each mutation was applied by exact string replacement, independently confirmed to
have landed (`git diff --numstat` non-empty, diff printed), the guard run, then
reverted.

| Mutation | Site | Guard | Result |
|---|---|---|---|
| first executor gets `512 MiB` instead of the resolved limit (the engine-creating one) | `RuntimePool::new` | `runtime_pool_resolves_compile_cache_limit_from_the_disk_signal` | **red** |
| loop executors get `512 MiB` instead of the resolved limit | `RuntimePool::new` | same | **red** |
| resolution moved AFTER the first executor (textual "exactly once" count stays at 1) | `RuntimePool::new` | same, positional assertion | **red** |
| the two `&Path` arguments to the estimate swapped (compiles) | `RuntimePool::new` | same | **red** |
| the startup prune removed | `RuntimePool::new` | same | **red** |
| `available_bytes(contracts_dir)` -> `None` (disables the disk bound on every host) | `startup_disk_budget_estimate` | `estimate_composes_the_measured_walk_and_the_free_space_read` | **red** |
| `measure_startup_disk_used(..)` -> `0` (the bare-free-space oscillation) | same | same | **red** |
| the two `&Path` arguments swapped inside the estimate | same | same | **red** |
| resolver drops the disk term (`disk_budget` -> `u64::MAX`), i.e. the #5014 defect itself | `default_wasmtime_cache_size_bytes` | `default_soft_limit_reader_derives_from_ram_and_disk_signals` + `default_soft_limit_matches_the_pure_clamp_of_this_hosts_ram_signal` | **red** |
| prune ignores the already-fits early return (deletes under the limit) | `prune_compile_cache_to_limit` | `prune_leaves_a_cache_that_already_fits_untouched` | **red** |
| prune target 70 -> 100 (diverges from wasmtime's own cleanup target) | same | `prune_deletes_oldest_first_down_to_the_target` | **red** |
| prune deletes NEWEST first | same | same | **red** |
| the over-share check dropped from the 60s recompute | `recompute_effective_budget` | `recompute_effective_budget_runs_the_over_share_check` | **red** |
| the over-share check uses its own divisor instead of the sizing rule's | `warn_if_compile_cache_over_share` | `compile_cache_over_share_warning_is_edge_triggered` | **red** |

14/14 caught. Nine of the fourteen are pure *wiring* mutations that every test of
the sizing math passes cleanly — which is the point: the fix lives in the call
sites. The first round of this PR shipped with the first eight rows of that list
uncovered.

Retained from the first round (re-run on this head): divisor 4 -> 1, RAM floor
re-imposed after the `min`, measured-used stops counting the compile cache,
executor re-hardcodes the limit, replacement executor re-derives instead of
reusing — all still **red**.

`cargo fmt` clean; `cargo clippy -p freenet --lib --all-features` clean; full
`cargo test -p freenet --lib --all-features` passes apart from four failures that
reproduce identically on the unmodified base commit under the same parallel load
(`test_set_own_addr_atomic_with_network_status`, `homepage_renders_dynamic_title`,
`cross_connection_median_returns_some_when_a_peer_has_inflation`,
`validate_host_function_delegate`) and all pass in isolation.

Closes #5014

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -1286,13 +1286,16 @@ mod executor_pin_tests {
     #[test]
     fn from_config_with_shared_modules_wires_offload_and_budget() {
         let src = include_str!("runtime.rs");
+        // `split_once` (not `split(..).next()`, whose `expect` can never fire) so
+        // a reworded end marker fails loudly instead of silently widening the
+        // region to end-of-file, `mod tests` included.
         let body = src
-            .split("pub(crate) async fn from_config_with_shared_modules(")
-            .nth(1)
+            .split_once("pub(crate) async fn from_config_with_shared_modules(")
             .expect("from_config_with_shared_modules must exist")
-            .split("\n    pub async fn preload(")
-            .next()
-            .expect("end of from_config_with_shared_modules");
+            .1
+            .split_once("\n    pub async fn preload(")
+            .expect("end marker of from_config_with_shared_modules must exist")
+            .0;
         assert!(
             body.contains("offload_compilation: production_offload_compilation()"),
             "must set offload_compilation from the production gate"
@@ -1331,7 +1334,8 @@ mod executor_pin_tests {
     }
 
     /// Pin: `RuntimePool::new` MUST resolve the wasmtime on-disk compile-cache
-    /// soft limit from the DISK signal, and must resolve it exactly once.
+    /// soft limit from the DISK signal, prune an already-oversized cache down to
+    /// it, and thread the resolved value into every executor it builds.
     ///
     /// This is the wiring the #5014 fix lives in, and it is invisible to every
     /// unit test of the pure math: `default_wasmtime_cache_size_bytes` could be
@@ -1339,55 +1343,140 @@ mod executor_pin_tests {
     /// wrong knobs, and nothing else in the tree would notice. The needles are
     /// assembled from `concat!` fragments that appear nowhere verbatim in
     /// `pool.rs`, so the pin cannot match its own source through `include_str!`.
+    ///
+    /// Region delimiters use `split_once(..).expect(..)` rather than
+    /// `split(..).next()`: the latter is `Some` unconditionally, so its `expect`
+    /// can never fire and a reworded end marker would silently widen the region
+    /// to end-of-file (including `mod tests`) while the assertions kept passing
+    /// for the wrong reason.
     #[test]
     fn runtime_pool_resolves_compile_cache_limit_from_the_disk_signal() {
         let src = include_str!("runtime/pool.rs");
         let body = src
-            .split("pub async fn new(")
-            .nth(1)
+            .split_once("pub async fn new(")
             .expect("RuntimePool::new must exist")
-            .split("\n    /// Get the current health status")
-            .next()
-            .expect("end of RuntimePool::new");
+            .1
+            .split_once("\n    /// Get the current health status")
+            .expect("end marker of RuntimePool::new must exist")
+            .0;
         // Whitespace-collapsed so a rustfmt re-wrap of the (multi-argument) call
         // cannot silently vacate the pin.
         let collapsed = body.split_whitespace().collect::<Vec<_>>().join(" ");
 
         assert!(
-            collapsed.contains(concat!(
-                "let wasmtime_cache_size_bytes = ",
-                "crate::wasm_runtime::default_wasmtime_",
-                "cache_size_bytes("
-            )),
-            "the pool must resolve the compile-cache soft limit through the \
-             node-relative resolver"
+            collapsed.contains(concat!("crate::ring::startup_disk_", "budget_estimate(")),
+            "the pool must derive the aggregate disk budget from the filesystem \
+             before any executor exists — the tracker does not exist yet (#5014)"
         );
-        // The disk knobs are what make the bound a DISK bound. Passing the
-        // directories but not the operator's pct/cap (or vice versa) would
-        // silently fall back to a budget unrelated to this node's disk.
+        assert!(
+            collapsed.contains(concat!(
+                "let limit = crate::wasm_runtime::default_wasmtime_",
+                "cache_size_bytes(disk_budget)"
+            )),
+            "the compile-cache soft limit must be the resolver applied to THAT \
+             disk budget — passing anything else (a constant, the RAM term alone) \
+             restores the #5014 defect"
+        );
+        // The disk knobs are what make the bound a DISK bound. Binding the wrong
+        // config value to a name (or swapping the two dirs, which both type as
+        // `&Path` and so compile happily) would silently size the cache against a
+        // budget unrelated to this node's disk, so pin the bindings themselves and
+        // then the ORDER in which they reach the estimate.
         for needle in [
-            concat!("&config.contracts", "_dir()"),
-            concat!("&config.wasmtime_", "cache_dir()"),
-            concat!("config.hosting_", "disk_pct"),
-            concat!("config.max_hosting", "_disk"),
+            concat!("let contracts_dir = config.contracts", "_dir();"),
+            concat!("let compile_cache_dir = config.wasmtime_", "cache_dir();"),
+            concat!("let hosting_disk_pct = config.hosting_", "disk_pct;"),
+            concat!("let max_hosting_disk = config.max_hosting", "_disk;"),
         ] {
             assert!(
                 collapsed.contains(needle),
-                "RuntimePool::new must pass `{needle}` to the compile-cache \
+                "RuntimePool::new must bind `{needle}` for the compile-cache \
                  resolver — without it the limit stops tracking this node's disk \
                  budget (#5014)"
             );
         }
-        // Exactly once: the resolver walks the contracts dir, walks the compile
-        // cache dir, and statvfs's the mount. A second call inside `new` would
-        // mean re-measuring per executor, which both wastes startup I/O and
-        // samples a cache dir the first executor has begun writing into.
+        for needle in [
+            "&contracts_dir, &compile_cache_dir",
+            "hosting_disk_pct, max_hosting_disk",
+        ] {
+            assert!(
+                collapsed.contains(needle),
+                "the estimate's arguments must stay in order (`{needle}`): both \
+                 dirs are `&Path`, so a swap compiles and silently measures the \
+                 wrong trees"
+            );
+        }
+        // Lowering the limit does not shrink an existing cache — wasmtime reads it
+        // only on its write-path cleanup, behind an hourly lock — so without this
+        // prune the fix bounds future starts and leaves an already-wedged node
+        // wedged (#5014). It must run against the cache dir and the freshly
+        // resolved limit.
+        assert!(
+            collapsed.contains(concat!(
+                "crate::ring::prune_compile_",
+                "cache_to_limit( &compile_cache_dir, limit )"
+            )) || collapsed.contains(concat!(
+                "crate::ring::prune_compile_",
+                "cache_to_limit(&compile_cache_dir, limit)"
+            )),
+            "the pool must prune an already-oversized compile cache down to the \
+             freshly resolved limit, before the first `Cache::new`"
+        );
+        // Exactly one CALL SITE inside `new`. This is a textual count, so it does
+        // not by itself prove a single runtime invocation — the positional
+        // assertion below is what rules out the case that actually matters
+        // (re-resolving per executor). What the count buys is that a second,
+        // divergent resolution cannot be added without the pin noticing.
         assert_eq!(
             collapsed
                 .matches(concat!("default_wasmtime_", "cache_size_bytes("))
                 .count(),
             1,
-            "the compile-cache soft limit must be resolved exactly once per pool"
+            "the compile-cache soft limit must have exactly one call site in `new`"
+        );
+        // Position: the resolution must precede the FIRST executor construction.
+        // Moving it inside the `for i in 1..pool_size` loop keeps the textual
+        // count at 1 while re-walking the mount once per executor and re-sampling
+        // a cache dir the first engine has begun writing into — the very defect
+        // the count assertion reads as if it prevents.
+        let resolved_at = collapsed
+            .find(concat!("crate::ring::startup_disk_", "budget_estimate("))
+            .expect("estimate call must exist");
+        let first_executor_at = collapsed
+            .find("from_config_with_shared_modules(")
+            .expect("RuntimePool::new must build executors");
+        assert!(
+            resolved_at < first_executor_at,
+            "the disk budget must be resolved BEFORE the first executor is built: \
+             wasmtime fixes the soft limit at `Cache::new` inside that call, and a \
+             later resolution would be both inert and measured against a cache dir \
+             the engine has started writing into"
+        );
+
+        // THE load-bearing wiring edge: the resolved value must actually reach the
+        // executor constructor. Everything above can hold while this argument is a
+        // constant — the value would still be resolved, still logged, still stored
+        // on the pool struct, and the whole #5014 fix inert on every host with a
+        // fully green suite. The first call site is the one that creates the
+        // engine (and so fixes wasmtime's soft limit); the second must agree with
+        // it, since a divergence there is a silent per-executor difference.
+        let threaded = "shared_contract_index.clone(), wasmtime_cache_size_bytes";
+        assert_eq!(
+            collapsed
+                .matches("from_config_with_shared_modules(")
+                .count(),
+            2,
+            "RuntimePool::new is expected to build exactly two executor shapes \
+             (the engine-creating first one and the shared-backend rest); a third \
+             call site would be unpinned by the assertion below"
+        );
+        assert_eq!(
+            collapsed.matches(threaded).count(),
+            2,
+            "every executor `RuntimePool::new` builds must receive the RESOLVED \
+             compile-cache soft limit as its last argument — a constant there \
+             leaves the value resolved, logged and stored while the bound never \
+             reaches wasmtime (#5014)"
         );
 
         // Third call site: a replacement executor must reuse the pool's resolved
@@ -1395,12 +1484,12 @@ mod executor_pin_tests {
         // engine has been filling, and the engine's soft limit was fixed at
         // startup anyway — so a fresh measurement would be both wrong and inert.
         let replacement = src
-            .split("async fn create_replacement_executor(")
-            .nth(1)
+            .split_once("async fn create_replacement_executor(")
             .expect("create_replacement_executor must exist")
-            .split("\n    /// Return an executor to the pool, replacing it")
-            .next()
-            .expect("end of create_replacement_executor");
+            .1
+            .split_once("\n    /// Return an executor to the pool, replacing it")
+            .expect("end marker of create_replacement_executor must exist")
+            .0;
         let replacement = replacement.split_whitespace().collect::<Vec<_>>().join(" ");
         assert!(
             replacement.contains(concat!("self.wasmtime_", "cache_size_bytes")),
@@ -1419,17 +1508,22 @@ mod executor_pin_tests {
     fn runtime_pool_sizes_caches_by_byte_budget() {
         // RuntimePool::new lives in runtime/pool.rs after the split.
         let src = include_str!("runtime/pool.rs");
+        // Bounded by `new`'s end marker rather than by a fixed chunk count: the
+        // count version silently stopped covering the cache construction as soon
+        // as anything was added to the preamble, which is a pin that fails for a
+        // reason unrelated to what it guards. `split_once` (not `split(..).next()`,
+        // whose `expect` can never fire) so a reworded marker fails loudly instead
+        // of widening the region to end-of-file, `mod tests` included.
         let body = src
-            .split("pub async fn new(")
-            .nth(1)
+            .split_once("pub async fn new(")
             .expect("RuntimePool::new must exist")
-            // Take the first chunk of the function body. Whitespace is collapsed
-            // so the assertions below survive line-wrapping / reformatting of the
-            // (now multi-line, metrics-threaded, interest-predicate-threaded)
-            // cache construction.
-            .split("\n    ")
-            .take(110)
-            .collect::<String>()
+            .1
+            .split_once("\n    /// Get the current health status")
+            .expect("end marker of RuntimePool::new must exist")
+            .0
+            // Whitespace is collapsed so the assertions below survive
+            // line-wrapping / reformatting of the (now multi-line,
+            // metrics-threaded, interest-predicate-threaded) cache construction.
             .split_whitespace()
             .collect::<Vec<_>>()
             .join(" ");
--- a/crates/core/src/contract/executor/runtime.rs
+++ b/crates/core/src/contract/executor/runtime.rs
@@ -9,7 +9,6 @@ use super::{
     ContractExecutor, ContractRequest, ContractResponse, ExecutorError, InitCheckResult,
     RequestError, Response, SLOW_INIT_THRESHOLD, STALE_INIT_THRESHOLD, StateStoreError, now_nanos,
 };
-use crate::wasm_runtime::default_wasmtime_cache_size_bytes;
 pub(crate) use contract_ops::ReclaimOutcome;
 pub use pool::RuntimePool;
 pub(crate) use pool::{ExportAdmission, ExportDone, MAX_CONCURRENT_EXPORTS};
@@ -438,6 +437,7 @@ impl Executor<Runtime> {
         inherited_origins: crate::wasm_runtime::SharedInheritedOrigins,
         shared_backend: Option<BackendEngine>,
         shared_contract_index: SharedContractIndex,
+        wasmtime_cache_size_bytes: u64,
     ) -> anyhow::Result<Self> {
         let db = shared_state_store.storage();
         // Pool executors all share ONE contract instance index (#4218), so a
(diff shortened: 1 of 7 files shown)


sanity:
## Consolidated review — Full tier

Four independent lenses ran blind against the checked-out code at `14d20095`: **code-first**, **testing/test-integrity**, **skeptical/adversarial**, **big-picture/design**, plus a **cross-PR interaction** review against #5033.

**Verdict: SHIP-WITH-FIXES**, with test integrity coming back **WEAK**.

The pins that exist are genuinely well built — correct regions, no self-match through `include_str!`, no doc-comment or string-literal matches, no test-source contamination, and the arithmetic tests correctly avoid writing assertions against the constants they guard. All eight claimed mutations really are caught, and every concrete value re-derives correctly by hand. The problem is what the mutation table does *not* cover.

### Blocking

**1. Two load-bearing wiring edges are unpinned, each with a one-token mutation that makes the entire fix inert with a fully green suite.**

- `pool.rs:464` — `RuntimePool::new` passes `wasmtime_cache_size_bytes` to `Executor::from_config_with_shared_modules`. Nothing tests either argument site. Changing it to `512 * 1024 * 1024` leaves the value still resolved, still logged, still stored in the struct — and #5014 unfixed on every host in the world. Green.
- `disk_usage.rs:766-767` — `startup_disk_budget_estimate`'s composition is unpinned. Three surviving mutations: `available_bytes(contracts_dir)` → `None` makes the basis `u64::MAX` and **disables the disk bound on every host**; `measure_startup_disk_used(...)` → `0` reintroduces exactly the bare-free-space oscillation the rustdoc says must not happen; and swapping the two `&Path` arguments at `wasm_runtime/runtime.rs:488-489` compiles and silently measures the wrong directories. All green.

This is the same helper-internals-tested / call-site-untested shape as the six prior PRs in this series. Both fixes are a few lines of source-scrape in the idiom this PR already uses.

Relatedly, the comment at `disk_usage.rs:1262-1266` claims "an estimate built on bare free space fails every line here" — it does not; that test drives the *pure* function, so the wiring is uncovered.

**2. The fix never shrinks an already-oversized cache, so it does not un-wedge the node #5014 describes.** Found independently by two reviewers. Lowering the soft limit evicts nothing until wasmtime next compiles a new distinct blob and runs its ~1h cleanup. On the wedged node — which by definition cannot accept new contracts — that trigger may never fire. The PR prevents the wedge on future starts while leaving an existing wedged node wedged.

**3. The framing is wrong in a way that points at a better fix.** The PR is built on the compile cache being a *non-reclaimable* consumer. It isn't: `rm -rf` on that directory is safe at any time on Unix and self-healing — reads are `fs::read` into a `Vec` (no mmap), the directory is recreated on the next write, and wasmtime's own worker already deletes files there. Reclaim-under-pressure is available and cheap. That directly solves #2: **prune the cache at startup when it exceeds the resolved limit.**

The `/4` cap is still worth having, but it bounds *the cache's contribution* — it does not guarantee "75% of the aggregate budget is always reclaimable". That guarantee is false once #5033's `max(state, db_file)` includes redb dead space, which eviction cannot reclaim either.

### Also worth closing

- `estimate_composes_the_du_walk_and_the_clamp_on_a_real_dir` is **vacuous**: it runs at `pct = 0.0`, multiplying the du-walk result by zero, and its second assertion is a containment check on a function that clamps into exactly that interval by construction. Deleting both `write_file` calls changes nothing. This PR calls out that exact anti-pattern at `wasm_runtime/runtime.rs:1411-1413` and then reproduces it.
- "Resolves exactly once" is a textual occurrence count, not an invocation count — moving the call inside the executor loop keeps the count at 1 while re-walking the mount per executor, which is the defect the comment claims to guard.
- Region delimiters use `.split(marker).next()`, which is `Some` unconditionally, so the trailing `.expect(...)` can never fire; if a marker is reworded the region silently widens to end-of-file, including `mod tests`.
- Layering: `wasm_runtime` → `ring` closes a module cycle; there is a cleaner seam.
- The `du` walks run inline on the async runtime, contradicting the discipline the identical walks follow elsewhere (`spawn_blocking`).
- Stale rustdoc contradicting the new behavior, and two doc claims the code does not support (`disk_usage.rs:693`, and the PR body's "under ~256 MiB" / 114 MiB table row).

### Interaction with #5033 (merging first)

No wedge, no collapse, no double-counting — the `estimate <= live budget` invariant survives and gets *stronger*. But after #5033 lands, `measure_startup_disk_used` still omits the single largest term, so the compile-cache limit lands **4-9x below its documented intent** on an aged peer (~102 MB where a quarter of the live budget would be ~437 MB). #5033 introduces `du_walk_shallow`, which needs no storage layer — so the stated reason for the omission ("redb rows are not readable before the storage layer exists") no longer covers `db_file_bytes`. Worth fixing during the rebase; the ordering is favorable because `RuntimePool::new` runs the #5005 startup compaction *before* resolving the estimate.

Fixes in progress on this branch; re-review will run on the updated diff.

[AI-assisted - Claude]

sanity:
Closing as superseded by #5328 ("fix(wasm): bound the wasmtime compile cache by disk headroom, not just RAM"), which merged 2026-08-15 and closed the shared issue #5014.

Verified against `origin/main`: `crates/core/src/wasm_runtime/runtime.rs` now carries the disk-side term this PR was written to add — `wasmtime_cache_size_for_disk`, `combine_wasmtime_cache_size` and `default_wasmtime_cache_size_bytes_for_dir`, with the rustdoc citing #5014 directly ("the disk-tight case is closed by composing this RAM term with a disk-derived term via min()"). So the RAM-only clamp that could wedge admission on a disk-tight, RAM-rich host is gone from main.

This branch has been conflicting since, and its approach (thread a pool-resolved `wasmtime_cache_size_bytes` into `from_config_with_shared_modules`) is the same fix landed earlier. Nothing here is lost — the branch and diff stay accessible, and this can be reopened if the read above is wrong.

[AI-assisted - Claude]

PR #5033, opened 2026-07-29
fix: count database file and webapp cache in disk accounting

### LINKED ISSUE #5007: Disk accounting under-counts real footprint ~10x (excludes database overhead)
## Problem

`DiskUsageTracker` sums **logical** contract-state bytes and excludes the database file's own overhead. The caveat is documented (`config.rs`, `cache.rs`: "ReDb/SQLite database overhead are additional and not counted against it, so this is not a hard bound on total on-disk usage"), but the size of the gap in practice was not appreciated.

Measured on a production peer (2026-07): the node reported **583 MB** of disk usage while its real footprint was **2.68 GB**, a ~10x under-count. The gap is the redb file's dead space.

## Why it matters

Two mechanisms consume this number and both were therefore deciding on badly wrong input:

- the aggregate disk budget and its eviction floor, `min(ram_budget, disk_budget)` (#4683)
- the pre-write admission gate (#4702)

A peer can be far over its real disk allowance while the budget believes it has tens of GB of headroom.

## Proposal

Feed the actual `db` file length into `DiskUsageTracker`. A `std::fs::metadata(db_path).len()` on the existing 60s sweep is cheap and exact, and would replace an estimate that is currently an order of magnitude out.

Startup compaction (PR #5005) shrinks the gap once per start but does nothing at runtime, so a long-lived peer re-accumulates dead space between restarts and the accounting drifts again.

## Related

- #5005 — startup compaction
- #4683 — aggregate disk budget
- #4702 — pre-write admission gate
- #4217 — end-to-end test for eviction reclaiming disk

[AI-assisted - Claude]

sanity:
## Problem

`DiskUsageTracker` summed **logical** contract-state rows and walked only `contracts_dir` + `wasmtime_cache_dir`. Two blind spots followed, both measured on production peers:

- **Database dead space.** redb is copy-on-write and only truncates *trailing* free space, so the interior dead space a busy node accumulates is invisible to any row-level accounting. A peer reported **583 MB** while occupying **2.68 GB** — a ~10x under-count.
- **The webapp cache.** It lives under the XDG *cache* dir, outside the walked set: **1236 MiB / 82 entries** on one machine, 325 MB / 61 on another. Nothing measured or reported those bytes.

Both feed the `min(ram, disk)` eviction floor (#4683) and the pre-write admission gate (#4702), so both have been deciding on a number an order of magnitude wrong.

This is the structural half of the disk-reduction effort. The six preceding PRs are one-time reclaims and they do not hold: the nova gateway compacted 2.59 GB → 1.41 GB and was back at **2.81 GB ten hours later**, because nothing could see it happening.

## Solution

**The storage term becomes `max(state_bytes, db_file_bytes)`**, with the database directory measured on the existing 60s sweep.

*Why `max` and not `state + (file − state)`.* Freeing rows returns pages to redb for **reuse** without shrinking the file, so a derived "overhead" term would inflate by exactly what eviction just freed — the aggregate would stay pinned above budget however much the node shed, an eviction treadmill that never converges. Under `max` the aggregate rests at the measured file size, which is the truth and a fixed point. Two more properties fall out: an unmeasured database degrades to the old behavior (`max(state, 0) == state`), and a within-window write burst that has outrun the last measurement is still visible to the gate.

*Why periodic and not delta-tracked.* A file length is not computable at a write chokepoint: redb extends the file when it needs a page, not per row. A per-write `metadata()` would be a syscall on the PUT hot path — under the `state_sizes` lock, which `admit_state_write` holds across `total_bytes()` — in exchange for no extra accuracy, since the live `state_bytes` delta already captures the marginal cost of an in-flight write. Staleness costs at most one sweep of growth in the admitting direction, which is the same exposure the wasm gauge already accepts.

**The webapp cache is measured and reported, but not budgeted.** #5012 already caps it at a hard 64 MiB; charging that against a hosting budget whose own floor is 128 MiB would spend up to half a small node's allowance re-bounding something already bounded. Hosting eviction has no lever on it either, so charging it reproduces the treadmill shape above. And it need not share the mount whose free space the budget's basis measures (XDG cache vs XDG data). Visibility was the actual defect for these bytes; double-bounding is not the fix.

### Measurement is allocation, not apparent length

The walks charge **allocated blocks** (`st_blocks * 512`, `du` semantics), not `metadata().len()`.

`len()` is what `ls` reports, and for a sparse file it over-states consumption. redb's file **is** sparse on some peers — measured on the nova gateway (2026-07-29, stable across repeated samples):

```
apparent (metadata().len) = 2,749,370,368   (2.56 GiB)
allocated (st_blocks*512) = 1,839,210,496   (1.71 GiB)
```

A **910 MB / 49% over-charge** on the term this PR makes dominant. The same file on the Framework laptop is not sparse at all (`size == allocated`), so this varies with each peer's compaction history and no constant correction exists. Over-charging makes the admission gate refuse writes the disk can still take: the mirror image of the under-count this PR fixes, and no more honest.

Allocation is also the only unit that composes with the budget's other term. The budget is `pct * (used + available)` where `available` is `statvfs`'s `f_bavail * f_frsize` — real free blocks. `used` has to be the same currency, so all four walks were converted rather than just the database one: block slack on a small file **is** charged (the filesystem spent that block, and `available` already reflects it) and a hole is not. Non-unix has no `st_blocks` in `std` and falls back to `len()`.

### A failed measurement is not a measurement of zero

Every walk used to swallow its `read_dir` error and return 0, and the refresh stored that unconditionally. One EACCES (a packaging or user change), a transient EMFILE, or the directory being replaced set `db_file_bytes = 0`; `max(state, 0) == state` reverted the whole aggregate to exactly the pre-#5007 under-count; and telemetry showed a plausible `hosting_disk_db_bytes: 0` with no log to contradict it. Under a persistent EACCES it was permanent and indistinguishable from a healthy small node — silently reintroducing the bug this PR exists to fix, on the term it exists to make visible.

A failed walk now keeps the gauge's **previous** value and warns, edge-triggered per gauge so a permanently-unreadable directory logs once rather than once a minute. This is the same fail-loud discipline the state seed already follows. A directory that does not **exist** is still a legitimate zero (a store not yet created), not a failure.

### An UPDATE absorbed by dead space is not a disk cost

The growth-only gate refused *any* growing UPDATE on an over-budget node. That was wrong in the case that matters: redb reuses freed pages, so a write that keeps live state at or under the measured file size does not grow the file by a byte. Refusing it aborted the merge while the node kept the contract, kept serving it, and kept advertising as a host — a copy that has dropped out of the update mesh but still answers reads, which `hosting-invariants.md` invariant 1 says must be impossible by construction. For a relayed UPDATE the rejection is fire-and-forget, so no peer learns. Growth is the normal case for the flagship contract class (a River room appends messages), and making `used` honest widens the at-risk population.

The projection is now `max(db_file_bytes, state_bytes − old + new)` — the honest storage term — instead of `max(db_file_bytes, state_bytes) − old + new`, which charged the delta on top of a file that will not move. Growth that fits inside the dead space is admitted; growth that genuinely enlarges the footprint is bounded exactly as before; a fresh PUT is still refused, because admitting new tenants is the decision the budget exists to make and a PUT has no already-hosted state whose merge would stall.

Two bounded imprecisions, both self-correcting within one sweep: `state_bytes` counts state payload only, so `db_file − state` also includes live non-state rows and slightly over-states what is reclaimable; and `db_file` is up to one sweep window stale.

**Residual, stated plainly:** growth that *exceeds* the dead space on an over-budget node is still refused, and the stale-host window stays open for that case. That is genuine disk exhaustion, where admitting would overflow the budget. Closing it means de-registering hosting and re-homing the contract's subscribers — epic #4642 piece D — not widening this gate. The edge-triggered over-budget warning now names the consequence, since it is not visible in the byte counts.

### Rollout: this cannot trigger mass eviction

The obvious fear — 10x the usage against an unchanged budget evicts everything — points the wrong way here. `used` sits in the budget's **own basis** (`clamp(pct * (used + available), MIN, cap)`), not on the other side of a comparison. That is monotonically non-decreasing in `used`, and `min(ram, ·)` preserves it, so the effective budget installed on the cache can only ever **rise**. No ramp is needed for the eviction floor, and `effective_budget_never_falls_as_measured_usage_rises` pins it across the whole shape of the clamp.

The tightening lands on the aggregate **admission gate**, which is where it belongs: it compares `used` against the budget, so each newly-visible byte costs `(1 − pct)` of headroom. That bounds what a full node **accepts**; it does not bound the footprint. Nothing here shrinks a database file — dead space is reclaimed by the startup compaction (#5005) and by nothing else — so a node already past its budget refuses new tenants at that size rather than returning to a smaller one. Making the figure honest is what lets an operator see that; a runtime compaction trigger is what would fix it.

When does a node land over budget? Whichever of the three clamp regions binds:

- in the linear region at `pct = 0.5`, when its own footprint exceeds its remaining free space;
- at the **32 GiB cap** (`DEFAULT_MAX_HOSTING_DISK_BYTES`), whenever its footprint exceeds 32 GiB regardless of how much disk is free — a node with 40 GB used and 500 GB free is over budget;
- the **128 MiB floor** only ever raises the budget, so it cannot put a node over.

Refusing new contracts there is a hard **storage ceiling**, which is the clause `hosting-invariants.md` sanctions ("storage is a minor term plus a hard ceiling"). It is **not** invariant 3's emergent refusal — that means a newcomer entering the eviction ranking and losing, whereas this is an at-capacity refusal on a resource axis and the contract never enters the ranking. The earlier draft of this description cited the wrong clause.

### Hosting invariants touched

- **Invariant 1** (a served copy is fresh). The dead-space exemption above is what keeps this PR from widening the stale-host failure class; the residual case is stated rather than papered over.
- **Invariant 3** (eviction is one demand-ordered decision, budget sized by the scarcest local resource). Preserved: the ordering is untouched, no new eviction input, no pin, no valve. Only the *measurement* of the disk resource changes, and the monotonicity property above proves the floor cannot tighten. The admission gate is justified by the hard-ceiling clause, not by invariant 3's emergent refusal.
- **Byte-only-eviction anti-pattern.** Not re-introduced: this makes an existing *capacity* measurement honest; it does not make bytes an eviction signal.

## Testing

27 new tests. The properties, not just the arithmetic:

- `evicting_state_cannot_shrink_the_aggregate_below_the_measured_file` — the fixed-point property that rules out the eviction treadmill.
- `within_window_state_growth_exceeds_a_stale_db_measurement` — the other half of the `max`.
- `db_measurement_is_shallow_and_skips_the_other_mode_split` — `db_dir` in Network mode contains `local/`, whose bytes belong to a different node instance; siblings of the database (sqlite `-wal`, redb `.backup`) still count.
- `absent_db_measurement_degrades_to_the_logical_row_sum` — backward-compat floor.
- `admission_gate_charges_the_database_file_not_just_the_rows` — the consumer that made the under-count load-bearing.
- `sparse_file_is_charged_by_allocation_not_apparent_length` / `block_slack_on_a_small_file_is_charged` — both directions of the measurement unit.
- `an_unreadable_database_directory_keeps_the_previous_measurement`, `an_unreadable_webapp_cache_keeps_the_previous_measurement`, `measurement_resumes_after_a_transient_failure` — a failed walk is not a zero, and does not poison the gauge afterwards.
- `a_missing_directory_measures_zero_rather_than_retaining` — and the over-correction is not made either.
- `admit_state_update_growth_absorbed_by_dead_space_is_admitted_over_budget`, `..._growth_beyond_dead_space_is_still_bounded`, `..._does_not_charge_growth_twice_against_the_file`, `dead_space_does_not_exempt_a_fresh_put` — the four corners of the growth gate.
- `over_budget_from_dead_space_still_merges_hosted_updates`, `honest_over_budget_usage_does_not_tighten_the_eviction_floor`, `over_budget_clears_when_free_space_grows` — what the node actually **does** in the adverse state, which was previously asserted only in prose and pinned only by a `bool`.
- `webapp_cache_never_enters_the_budgeted_aggregate` + `webapp_cache_is_measured_and_reported` — the reported-not-budgeted split, both directions.
- `all_four_measured_dirs_are_read_from_their_own_path` — the transposition hazard `HostingDiskPaths` exists to make unlikely.
- `effective_budget_never_falls_as_measured_usage_rises` — the no-mass-eviction property.
- `over_disk_budget_is_logged_on_the_edge_only` — both ways of losing the operable signal.

**Call sites were mutation-tested, not just internals.** Each mutation confirmed red on the tests listed:

| # | Mutation | Caught by |
|---|---|---|
| 1 | drop both new refreshes from the sweep | `refresh_disk_usage_measures_database_and_webapp_cache` |
| 1b | drop only the webapp refresh | same |
| 2 | `db_dir: config.contracts_dir()` (wrong accessor) | `hosting_disk_paths_are_installed_from_the_node_config` |
| 3 | `webapp_cache_dir: default_webapp_cache_dir()` (process-wide) | same, via the accessor assert (see note below) |
| 4 | delete the whole install call | same (pin guard) |
| 5 | revert `total_bytes` to the row sum | 5 disk_usage tests |
| 6 | charge the webapp cache to the aggregate | `webapp_cache_never_enters_the_budgeted_aggregate` |
| 7 | budget deducts non-state footprint (the rejected design) | `effective_budget_never_falls_as_measured_usage_rises` |
| 8 | drop the over-budget signal call | `-D warnings` dead-code gate |
| 9 | `file_disk_bytes` returns `len()` (apparent size) | the two measurement tests + 9 walk tests |
| 10 | a failed walk stores 0 (the old behavior) | the three retention tests |
| 11 | growth gate reverts to `total − old + new`, no exemption | the three growth-gate tests + `over_budget_from_dead_space_still_merges_hosted_updates` |

A precision note on mutation 3, since the table used to leave it ambiguous. Two assertions in that test would trip under it: the loop's accessor assert (`config.ws_api.webapp_cache_dir` is no longer in the extracted arguments) and the negative guard (`!src.contains("default_webapp_cache_dir")`, since the substituted symbol would now appear in the production slice). The loop runs first, so the accessor assert is what actually reports; the guard is never reached. The guard on its own **cannot** go red from production code being *deleted* — it is a forward guard against reintroducing the process-wide default, which is what its own code comment says. Credit for catching mutation 3 belongs to the accessor assert.

Mutation 7 initially **passed** — the monotonicity fixture drove `used` purely through state rows, leaving the non-state footprint at zero, so the rejected deduct design was invisible to it. That is the exact inversion the test exists to forbid. Fixed in `cc6069a2`, and re-confirmed red after the fixture was rewritten (see below). Committed before mutating so nothing was lost to the revert.

### The measurement fix silently un-armed two existing tests

Worth stating on its own, because it is the failure mode this series keeps hitting: **a change that makes measurement more correct disarmed the tests guarding it, and every one of them stayed green.**

Switching the walks from apparent length to allocated blocks invalidated two fixtures that had been written in bytes:

- `effective_budget_never_falls_as_measured_usage_rises` drove `used` through a **sparse** `set_len` file. A hole allocates nothing, so every entry in its sweep — `0`, `128 MiB`, `583 MiB`, `2680 MiB`, `40 GiB` — collapsed to the same measured `used` of 0. The "never falls" assertion then compared a constant against itself. It passed, asserting nothing, and would have gone on passing.
- `refresh_disk_usage_measures_database_and_webapp_cache` — the **call-site pin**, the one test standing between this PR and a silently-dropped `refresh_db_file` — grew its fixtures from 11 to 1100 bytes. Neither size moves an allocation-based measurement on a 4 KiB-block filesystem, so the pin would have passed while measuring nothing at all.

Both are fixed and both were re-confirmed red under their mutations (7 and 1 in the table above). The monotonicity fixture now injects the database size through a `#[cfg(test)]` setter, which also lets it reach the 32 GiB cap that no on-disk fixture could materialize; the call-site pin grows by whole 64 KiB chunks and reads the post-growth sizes back from the filesystem rather than assuming them.

**Rule for anyone adding a fixture to this module: think in blocks, not bytes.** A file's contribution is `st_blocks * 512`. Sub-block sizes are indistinguishable from each other, sparse files contribute nothing regardless of length, and a filesystem with transparent compression can collapse a run of zeros — which is why the test helper writes a pseudo-random stream. Assert against a size read back from the filesystem, not against the byte count you wrote.

Four of the new tests pass **before** their fix as well as after, and are stated as such rather than as regression pins: `a_missing_directory_measures_zero_rather_than_retaining` and `dead_space_does_not_exempt_a_fresh_put` guard against over-correcting in the opposite direction, and `honest_over_budget_usage_does_not_tighten_the_eviction_floor` / `over_budget_clears_when_free_space_grows` cover behavior that was correct but unpinned.

The source pin reads a comment-stripped, whitespace-collapsed slice of the production half of `handler.rs` (so a rustfmt re-wrap cannot rot a needle) and assembles every needle from `concat!` fragments appearing nowhere verbatim, so it cannot match its own source through `include_str!`.

`cargo fmt` clean; `cargo clippy -p freenet --lib --all-features` clean; 4478 lib tests pass.

## Also in this PR

- `DiskUsageStats::total_bytes` delegates to `total_bytes()` instead of re-deriving the aggregate, so the budgeted formula has one definition.
- The home-page tooltip says its parts do not sum to the tile above them, because state bytes live *inside* the database file and the aggregate takes the larger of the two — an operator who adds them up and gets a different number had no way to tell a display bug from the `max` rule.
- Two comments this work invalidated: `path_handlers.rs` still said the accounting walks only `contracts_dir` + `wasmtime_cache_dir`, and the sweep comment in `ring.rs` still said the gauges were "observational only" (untrue since #4702).
- The over-budget warning's remediation advice named only the startup compaction. That is wrong for one case the shallow walk deliberately counts: redb's schema migration renames the old database to `db.backup.<timestamp>` (`redb.rs::backup_and_remove_database`), nothing ever deletes it, and no restart reclaims it. The warning now names the file and the remedy. This is complementary to reporting true dead space from the backend's own in-use byte count, not a substitute for it: in-use bytes tell an operator how much the *live database* would give back to a compaction, whereas an orphaned migration backup is a separate file that no compaction can ever touch and only `rm` reclaims. Both land in the same `log_disk_budget_transition` warning, so if the final diff shows them together they are additive edits and neither supersedes the other: the figure should come from `db_file - db_in_use`, and the migration-backup clause still has to be there, because that backup file is counted by the shallow walk but is not part of the live database's in-use bytes.
- `seed()`'s rustdoc no longer claims every gauge is populated the instant `is_seeded` flips. It is not — the walks run just after the flag — and the doc now says why that window is permissive rather than blocking, and why moving the walks ahead of the flag would trade it for a worse one.

Closes #5007

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -116,6 +116,9 @@ pub struct ConfigArgs {
     /// are evicted (least-valuable-first) and their on-disk state reclaimed.
     /// This bounds tracked contract state only — WASM code blobs and ReDb/
     /// SQLite database overhead are additional and not counted against it.
+    /// The separate aggregate DISK budget (`hosting-disk-pct` /
+    /// `max-hosting-disk`) does count them, including the database file's own
+    /// footprint since #5007.
     /// Default: 1 GiB.
     #[arg(long, env = "MAX_HOSTING_STORAGE")]
     pub max_hosting_storage: Option<u64>,
@@ -1206,6 +1209,9 @@ pub struct Config {
     /// are evicted (least-valuable-first) and their on-disk state reclaimed.
     /// This bounds tracked contract state only — WASM code blobs and ReDb/
     /// SQLite database overhead are additional and not counted against it.
+    /// The separate aggregate DISK budget (`hosting-disk-pct` /
+    /// `max-hosting-disk`) does count them, including the database file's own
+    /// footprint since #5007.
     ///
     /// The default is capability-relative (RAM-scaled): `clamp(total_ram / 8,
     /// 128 MiB, 1 GiB)`, so a memory-constrained host gets a proportionally
--- a/crates/core/src/contract/handler.rs
+++ b/crates/core/src/contract/handler.rs
@@ -124,12 +124,20 @@ impl ContractHandler for NetworkContractHandler {
         // This must be done before loading the cache so evictions work correctly
         let storage = executor.state_store().inner().clone();
         op_manager.ring.set_hosting_storage(storage.clone());
-        // Aggregate disk-usage tracker paths (#4683): the mode-resolved
-        // contracts dir (WASM blobs) and the relocated wasmtime compile-cache
-        // dir. Seeded lazily on the first sweep tick.
+        // Aggregate disk-usage tracker paths (#4683, #5007). This is the ONLY
+        // place the tracker learns where the node's bytes live, so a directory
+        // missing here is a directory nothing measures, bounds, or reports —
+        // which is exactly how the database file's dead space and the whole
+        // webapp cache stayed invisible until #5007. Seeded lazily on the first
+        // sweep tick. Pinned by
+        // `hosting_disk_paths_are_installed_from_the_node_config`.
         op_manager.ring.set_hosting_disk_paths(
-            config.contracts_dir(),
-            config.wasmtime_cache_dir(),
+            crate::ring::HostingDiskPaths {
+                contracts_dir: config.contracts_dir(),
+                wasmtime_cache_dir: config.wasmtime_cache_dir(),
+                db_dir: config.db_dir(),
+                webapp_cache_dir: config.ws_api.webapp_cache_dir.clone(),
+            },
             config.hosting_disk_pct,
             config.max_hosting_disk,
         );
@@ -1442,6 +1450,135 @@ pub mod test {
             "Expected NoEvHandlerResponse, got {:?}",
             result
         );
+    }
+
+    /// Call-site pin: the disk-usage tracker must be installed with all four of
+    /// the node's own directories (#4683, #5007).
+    ///
+    /// This one call is the entire boundary between "measured" and "invisible".
+    /// The tracker cannot walk a directory it was never given, so a field
+    /// dropped here — or wired to the wrong accessor — silently reproduces
+    /// exactly the two blind spots #5007 exists to close: a database whose dead
+    /// space nothing sees (~10x under-count on a production peer) and a webapp
+    /// cache nothing measures at all (1236 MiB on one machine). Both consumers
+    /// of the resulting figure, the `min(ram, disk)` eviction floor (#4683) and
+    /// the pre-write admission gate (#4702), then decide on a number that is an
+    /// order of magnitude wrong, and every other test in the tree stays green.
+    ///
+    /// Nothing else can catch it. The `HostingDiskPaths` fields are all
+    /// `PathBuf`, so the type system permits both omission (via struct-update
+    /// syntax) and transposition; and a behavioral test would have to boot a
+    /// real node with a real store, which is what this whole file's `build` path
+    /// does only in integration tests that do not assert on disk accounting.
+    ///
+    /// Read against the comment-stripped, whitespace-collapsed production half
+    /// of this file so a rustfmt re-wrap of the call cannot rot the needles, and
+    /// the needles are assembled from fragments so this test cannot match its
+    /// own source through `include_str!`.
+    #[test]
+    fn hosting_disk_paths_are_installed_from_the_node_config() {
+        let src = production_source();
+
+        let call = concat!("set_hosting_disk", "_paths(");
+        assert_eq!(
+            src.matches(call).count(),
+            1,
+            "pin guard: expected exactly one disk-paths install in this file, so \
+             this test is asserting against the wrong thing"
+        );
+        let args = call_arguments(&src, call);
+
+        for (field, accessor) in [
+            // Pre-#5007 pair. Present so the pin covers the whole struct, not
+            // only the fields the issue added.
+            (
+                concat!("contracts", "_dir:"),
+                concat!("config.contracts", "_dir()"),
+            ),
+            (
+                concat!("wasmtime_cache", "_dir:"),
+                concat!("config.wasmtime_cache", "_dir()"),
+            ),
+            // #5007 blind spot 1: the storage backend's database directory. Its
+            // measured size is the only term that can see redb's dead space.
+            (concat!("db", "_dir:"), concat!("config.db", "_dir()")),
+            // #5007 blind spot 2: the unpacked-webapp cache. Reported, not
+            // budgeted — but it has to be measured to be reported at all, and it
+            // must come from the node's OWN config, never the process-wide
+            // default, for the same reason `server.rs` threads it.
+            (
+                concat!("webapp_cache", "_dir:"),
+                concat!("config.ws_api.webapp_cache", "_dir"),
+            ),
+        ] {
+            assert!(
+                args.contains(field),
+                "the disk tracker must be installed with `{field}`; found \
+                 arguments `{args}`"
+            );
+            assert!(
+                args.contains(accessor),
+                "`{field}` must be resolved from the node's config via \
+                 `{accessor}`; found arguments `{args}`"
+            );
+        }
+
+        // The substitution template must not exist on this path at all: the
+        // process-wide webapp-cache default would measure some other node's
+        // directory (or the developer's) instead of this one's.
+        assert!(
+            !src.contains(concat!("default_webapp", "_cache_dir")),
+            "the node's disk accounting must never resolve the process-wide \
+             webapp cache default — it has to measure the directory this node \
+             actually serves from"
+        );
+    }
+
+    /// Production half of this file, comment-stripped and whitespace-collapsed,
+    /// for the source pin above.
+    ///
+    /// Cut at the test module's `mod` line, NOT at `#[cfg(test)]`: that
+    /// attribute also sits on individual test-only items earlier in this file,
+    /// so cutting there truncates the slice the pin reads and turns it vacuous
+    /// rather than red. Comments are dropped before collapsing so a `//` line
+    /// cannot run into the code below it, and whitespace is collapsed so a
+    /// rustfmt re-wrap of a call's arguments cannot rot a needle. The needle for
+    /// the cut is split so it cannot match its own source.
+    fn production_source() -> String {
+        const FULL: &str = include_str!("handler.rs");
+        let cutoff = FULL
+            .find(concat!("pub mod ", "test {"))
+            .expect("handler.rs must have a test module");
+        FULL[..cutoff]
+            .lines()
+            .map(|line| line.split("//").next().unwrap_or(""))
+            .collect::<String>()
+            .chars()
+            .filter(|c| !c.is_whitespace())
+            .collect()
+    }
+
+    /// The balanced-parenthesis argument list of the (single) call to
+    /// `call_prefix` within `src`.
+    fn call_arguments(src: &str, call_prefix: &str) -> String {
+        let start = src
+            .find(call_prefix)
+            .unwrap_or_else(|| panic!("could not find {call_prefix}"))
+            + call_prefix.len();
+        let mut depth: i32 = 1;
+        for (offset, ch) in src[start..].char_indices() {
+            match ch {
+                '(' => depth += 1,
+                ')' => {
+                    depth -= 1;
+                    if depth == 0 {
+                        return src[start..start + offset].to_string();
+                    }
+                }
+                _ => {}
+            }
+        }
+        panic!("unterminated call to {call_prefix}");
     }
 }
 
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -1621,14 +1621,24 @@ pub struct HostingSnapshot {
     /// disk tracker is configured and seeded (early startup) — distinct from
     /// `Some(0)`, which means seeded-and-empty.
     pub disk_state_bytes: Option<u64>,
+    /// Measured byte size of the storage backend's database directory (#5007).
+    /// `disk_db_bytes − disk_state_bytes` is the database's dead space: bytes
+    /// the file holds that no live row accounts for, reclaimable only by the
+    /// startup compaction. Same seeded-gate semantics as `disk_state_bytes`.
+    pub disk_db_bytes: Option<u64>,
     /// Aggregate on-disk bytes used by `*.wasm` code blobs. Same
     /// seeded-gate semantics as `disk_state_bytes`.
     pub disk_wasm_bytes: Option<u64>,
     /// Aggregate on-disk bytes used by the wasmtime compile cache. Same
     /// seeded-gate semantics as `disk_state_bytes`.
     pub disk_compile_cache_bytes: Option<u64>,
-    /// Sum of `disk_state_bytes` + `disk_wasm_bytes` + `disk_compile_cache_bytes`
-    /// — the aggregate the disk budget bounds. Same seeded-gate semantics.
+    /// Measured byte size of the unpacked-webapp cache (#5007). Reported for
+    /// visibility but NOT part of `disk_total_bytes`: #5012 caps it separately
+    /// and hosting eviction has no lever on it. Same seeded-gate semantics.
+    pub disk_webapp_cache_bytes: Option<u64>,
+    /// `max(disk_state_bytes, disk_db_bytes)` + `disk_wasm_bytes` +
+    /// `disk_compile_cache_bytes` — the aggregate the disk budget bounds. Same
+    /// seeded-gate semantics.
     pub disk_total_bytes: Option<u64>,
     /// The aggregate disk budget the admission gate checks projected writes
     /// against (`HostingManager::disk_budget_bytes`, #4702). `None` while it
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -100,10 +100,6 @@ pub use connection_manager::{set_nn_lattice_enabled, set_nn_lattice_force_active
 mod connection;
 mod hosting;
 pub(crate) use broken_invariants::{BrokenInvariant, BrokenInvariantsTracker};
-/// Pre-write admission-gate rejection type (#4683, PR 3). Surfaced to the
-/// executor chokepoints so a write that would overflow the aggregate disk
-/// budget is refused before any bytes land.
-pub(crate) use hosting::DiskBudgetExceeded;
 /// The pre-A2 flat 1 GiB budget, used as the upgrade-migration sentinel in
 /// `config::ConfigArgs::build` so an upgraded node re-derives its hosting budget
 /// instead of keeping the historically-pinned default (#4565).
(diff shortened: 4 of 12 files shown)


sanity:
## Consolidated review — Full tier

Five independent lenses ran blind against the checked-out code at `f24ba620`: **code-first**, **testing/test-integrity**, **skeptical/adversarial**, **big-picture/hosting-invariants**, and a **cross-PR interaction** review against #5029. Plus direct measurement on two production peers.

**Verdict: SHIP-WITH-FIXES.** The `max(state_bytes, db_file_bytes)` design is right and was independently verified by three reviewers — including the point that it *kills a real downward spiral that exists on main today* (with `avail ≈ 0`, `used = state` → budget halves → shed → budget halves → down to the 128 MiB floor). The rejection of `state + (file − state)` is correct. Test integrity on the accounting paths came back **SOUND** — the six-PR vacuity pattern does not repeat there.

The blockers are around the edges of that core, not in it.

### Blocking

**1. The measurement over-counts a sparse database file.** Not from code review — from measuring live peers. `du_walk_shallow` uses `metadata().len()` (apparent size). On the nova gateway, stable across repeated samples:

```
size (metadata().len)     = 2,749,370,368   (2.56 GiB)
allocated (st_blocks*512) = 1,839,210,496   (1.71 GiB)
```

A **910 MB / 49% over-charge**. The framework laptop's file is *not* sparse, so this varies by peer and compaction history. It also mixes dimensions: the budget basis is `used + available`, and `available` comes from `statvfs`, which reports allocated blocks — so charging apparent bytes against allocated free bytes is not comparing like with like. Should be `st_blocks * 512` on Unix, `len()` fallback elsewhere.

**2. `refresh_db_file` fails silently to zero on the now-dominant term.** `disk_usage.rs:687-690` → `:723-727`. One `read_dir` failure (EACCES, transient EMFILE, dir replaced) sets `db_file_bytes = 0`, `max(state, 0)` reverts the aggregate to exactly the pre-#5007 under-count, with no log and a plausible `hosting_disk_db_bytes: 0` in telemetry. This silently reintroduces the bug the PR exists to fix, and contradicts the module's own fail-loud seeding discipline at `:114-120`.

**3. Growing UPDATEs are refused while the peer keeps advertising as a fresh host.** Found independently by two reviewers. The gate admits only `new_size <= old` (`:618`); a growing UPDATE on an over-budget node is rejected at `executor_impl.rs:1912-1921`, nothing de-registers hosting or tears down subscriptions, and the rejection is fire-and-forget with no `UpdateMsg::Error` so no peer learns. Growth is the normal case for River rooms. That is a copy which has dropped out of the update mesh while still answering reads as authoritative — what invariant 1 says must be impossible by construction. The mechanism predates this PR (#4702), but this PR makes it ~10x more reachable, and **the PR body's "UPDATEs still flow" is materially wrong.**

**4. Two unpinned call sites**, both the "helper internals tested, wiring untested" shape from the prior six PRs:
- `ring.rs:1788/:4230` — the Ring→snapshot wiring for both new gauges. Deleting the `hosting_disk_db_bytes` assignment leaves the field `None` in every telemetry event forever, suite and clippy green. Mapping `webapp_cache_bytes` into `db_bytes` makes the dashboard report nonsense. Green either way.
- `handler.rs:136-139` (pin at `:1484`) — asserts field names and accessors *independently*, so it catches omission and wrong-accessor but **not transposition**, which is the exact hazard `HostingDiskPaths` was introduced to prevent. `contracts_dir: config.db_dir(), db_dir: config.contracts_dir()` passes the pin.

### Also worth closing

- **The wedge is invisible to central telemetry.** `RouterSnapshotInfo` has no `disk_budget_bytes`, the crossing signal is a `warn!` (only enumerated `EventKind`s reach the collector), and `disk_headroom` clamps at 0 so it cannot distinguish "full" from "5 GB over". A node refusing every PUT is undetectable fleet-wide — and this PR is what makes that state reachable.
- **"Dead space" is mislabelled.** `db_file_bytes − state_bytes` is "everything that isn't state payload", which structurally includes live index rows (`contract_params`, `hosting_metadata`, the various indices). The warn tells operators to restart, which is wrong for the live portion. `Database::stats()` already exposes in-use bytes and `RedbStorage` already consults it for `compaction_is_worthwhile` (`redb.rs:610`) — plumbing that gives the figure compaction can actually reclaim.
- **The over-budget state has no behavioral coverage** — only the log flag is tested.
- Doc accuracy: "18 new tests" (the diff adds 14), and the over-budget claim ignores both clamps (with the 32 GiB cap, 40 GB used / 500 GB free is over budget).
- Stale comments this PR invalidates: `path_handlers.rs:155-157`, `ring.rs:2993-2994`.

### Verified correct

Path wiring (no transposition at the install), no double-counting across the four disjoint dirs, admission-gate direction conservative for growth, saturating arithmetic throughout, safe uninitialized/cold-start behavior, `spawn_blocking` for the walks, correct lock ordering with no lock-across-await, and no hits against `bug-prevention-patterns.md`. **No eviction treadmill** — traced across sweeps and it is a genuine fixed point.

Fixes in progress on this branch; re-review will run on the updated diff.

[AI-assisted - Claude]

sanity:
## This PR no longer merges onto current `origin/main`

Found while screening stale PRs for an unrelated CI issue (freenet/freenet-core#5452), but this is a bigger and prior problem: attempting a real `git merge origin/main` on this branch produces **18 conflict hunks across 5 files** — `crates/core/src/config.rs`, `crates/core/src/ring.rs`, `crates/core/src/ring/hosting.rs` (one hunk alone spans lines 8142-8548, 406 lines), `crates/core/src/ring/hosting/disk_usage.rs`, and `crates/core/src/server/home_page.rs`.

These aren't drive-by textual collisions. `crates/core/src/ring/hosting.rs` has grown from 7003 to 8502 lines on `main` since this PR's merge-base (2026-07-29) via 9 intervening commits touching that same file, including:

- #5333 — replaced the fixed 1 GiB resident-overhead ceiling with a live-memory-aware budget
- #5328 — bound the wasmtime compile cache by disk headroom, not just RAM
- #5324 — bound peer memory at the 2 GiB cap (module-cache keying, Store arena, cache composition)

This PR is itself a disk-accounting/budget change touching the same subsystem, so `main` has independently reworked the exact problem this PR addresses while it sat idle.

I aborted the merge rather than resolving it. Resolving 400+ lines of conflicting logic in a subsystem under active, invariant-governed redesign (see `.claude/rules/hosting-invariants.md`) requires a judgment call about which approach is correct where the two diverge — that's not something a screening pass should decide, and any clippy or test result obtained from a self-resolved merge would be verifying code nobody actually wrote.

Question for whoever owns this next: **is this PR still wanted given #5333/#5328/#5324, or has the underlying problem already been addressed by the new budget model?** If it's still wanted, it likely needs reworking onto the new model rather than a mechanical rebase.

[AI-assisted - Claude]

sanity:
Disposition, so this stops sitting in the stale queue without a decision.

**Recommend closing and redoing rather than rebasing.** The prior comment measured 18 conflict hunks across 5 files, with one hunk in `crates/core/src/ring/hosting.rs` spanning 406 lines — and that file has grown from 7,003 to 8,502 lines on `main` since this branch's merge-base (2026-07-29) across 9 intervening commits.

That is past the point where a rebase is a mechanical operation. Resolving a 406-line conflict inside the hosting logic is authoring new code under the label of conflict resolution, and it lands with no review of the resolution itself — which is exactly the failure mode the per-code-content review rule exists to prevent. A fresh implementation against today's `hosting.rs` would be smaller, honestly reviewable, and would not carry three months of assumptions about a file that has since changed underneath it.

The *intent* is still good and should not be lost with the branch: the database file and webapp cache are real disk consumers that the accounting does not see, so a node's reported usage understates reality and the eviction thresholds are keyed off a number that is wrong in a known direction.

Not closing it myself, because this is a judgement call about work someone may still want and there is no CI or contributor blocker forcing the issue — flagging it for @sanity to close or re-target. If it is closed, the disk-accounting gap deserves an issue so the reasoning above survives the branch.

Reviewed as part of a sweep of PRs untouched for 24h+.

[AI-assisted - Claude]

PR #5037, opened 2026-07-29
fix(tracing): reclaim orphaned _EVENT_LOG segments when log is off

### LINKED ISSUE #5035: Orphaned _EVENT_LOG segments are never reclaimed after #4968 disabled the log — 224 MB on a user peer, 412 MB on nova
## Problem

#4968 (shipped v0.2.112) correctly flipped the `_EVENT_LOG` default to **OFF in network mode**, so new peers no longer pay for it. But nothing ever removes the segments a peer already wrote. Every peer that ran a pre-0.2.112 build is carrying its entire accumulated event log as **dead weight, forever** — never read, never written, never pruned.

## Measured (2026-07-29)

| Peer | `_EVENT_LOG*` | files | share of data dir | last write |
|---|---|---|---|---|
| framework (real-NAT user peer, 1.3 GB total) | **224 MB** | 23 | **17%** | Jul 27 |
| nova gateway (2.5 GB total) | **412 MB** | 24 | **16%** | Jul 27 |

On both, the active `_EVENT_LOG` is 0 bytes and the newest segment's mtime is the day 0.2.112 was deployed — writing genuinely stopped, and the accumulated segments simply stayed.

This is the second-largest consumer on framework after the database, and it is larger than `contracts/` by 7x.

## Why it is safe to delete

Per the flag's own documentation in `config.rs`:

> This log is a PURELY LOCAL forensic record. It is NOT the telemetry that feeds telemetry.freenet.org [...] Nothing in the node reads this log back to make decisions, and `freenet service report` does not include it.

So when `event_log_enabled() == false` the segments have no consumer at all.

## Proposed fix

On startup, when `event_log_enabled()` is false, delete orphaned `_EVENT_LOG.<index>` segments and log the reclaimed byte total at INFO.

Constraints:
- **Only when the log is disabled.** An operator who deliberately enables it must keep their history.
- Do not touch `_EVENT_LOG_LOCAL` in local mode — `fdev verify-state` consumes it (and local mode has the log enabled anyway, so the guard already covers this).
- Match the segment-name grammar `aof.rs` already defines (`SEGMENT_INDEX_DIGITS` = 10 ASCII digits) rather than a loose glob, so an unrelated file starting with `_EVENT_LOG` is never removed.
- Deleting is one-time per peer; after the first disabled startup there is nothing left to remove.

Accepted trade-off: a user who disables the log, restarts, then re-enables it loses prior history. That is the right call for a diagnostic log nothing reads, and the reclaim is logged.

## Related

- The event log is bounded by **record count** (`MAX_LOG_RECORDS` = 100,000 over 20 segments of 5,000), never by bytes. While it is enabled its byte footprint is a free variable of record size — 412 MB on nova at ~4.1 KB/record. Worth a byte cap as a separate follow-up for operators who run with it on.
- `_EVENT_LOG*` is also outside the set of directories #5033 measures, as is `secrets/` (126 MB on framework). Filed separately.

[AI-assisted - Claude]

sanity:
## Problem

#4968 (shipped v0.2.112) flipped the local `_EVENT_LOG` default to **OFF in network mode**, so new peers no longer pay for it. But nothing ever removed the segments a peer had already written. Every peer that ran a pre-0.2.112 build carries its accumulated event log as dead weight, forever.

Measured 2026-07-29:

| Peer | `_EVENT_LOG*` | files | share of data dir | last write |
|---|---|---|---|---|
| framework (real-NAT user peer, 1.3 GB total) | **224 MB** | 23 | **17%** | Jul 27 |
| nova gateway (2.5 GB total) | **412 MB** | 24 | **16%** | Jul 27 |

On both, the active `_EVENT_LOG` base file is 0 bytes and the newest segment's mtime is the day 0.2.112 deployed — writing genuinely stopped and the segments simply stayed. On framework this is the second-largest consumer after the database, and 7x larger than `contracts/`.

## Approach

`tracing/aof.rs::LogFile::reclaim_segments` (mechanism) iterates `scan_segment_indices` — the same function `LogFile::open` uses — so the segment-name grammar (`<base>.` + exactly `SEGMENT_INDEX_DIGITS` = 10 ASCII digits) is reused rather than re-invented as a glob. `_EVENT_LOG.backup`, `_EVENT_LOG.0000000001.tmp`, `_EVENT_LOG_notes.txt`, `_EVENT_LOG.00000000ab`, 9- and 11-digit suffixes, `_EVENT_LOG_LOCAL*`, and the `_EVENT_LOG` base file itself are all untouched.

`tracing/aof.rs::reclaim_orphaned_event_log` (policy) returns `None` — no sweep attempted — unless **both** guards pass:

1. `Config::event_log_enabled()` is false. An operator who deliberately enabled the log keeps every segment.
2. The node is in Network mode. In Local mode `Config::event_log()` resolves to `_EVENT_LOG_LOCAL`, which `fdev verify-state` discovers.

Call site: `NodeConfig::new` (`node.rs`), what `run_network` calls with the fully-resolved `Config`. Unlike a full node build it is cheap enough to drive from a unit test, which is what makes the wiring pinnable by **behaviour** rather than by a source scrape.

Failures never fail startup: an undeletable file, a directory occupying a segment name, an unreadable directory — each is counted and logged, and startup continues.

### What this deletes, stated plainly

This removes files, irreversibly, on the first restart after upgrading. Two things follow that are easy to miss:

- **The bytes are not unreadable, only un-read.** Nothing consumes these segments *automatically* — no reader selects `_EVENT_LOG` on its own, and `freenet service report` excludes it (nothing under `bin/commands/` references `_EVENT_LOG` at all). But every reader that a human or a test can aim at a path **will** pick these up, and more of them than an earlier draft of this section claimed. `fdev verify-state --log-file <path>` takes an arbitrary path, and `--log-directory` walks a tree accepting any file name *containing* `EVENT_LOG` (`verify_state.rs:180-190`; only its *directory* branch at `verify_state.rs:172` is `_EVENT_LOG_LOCAL`-specific), so pointing it at a network node's data dir does read these segments. `EventLogAggregator::from_aof_files` likewise takes caller-supplied paths, and its in-tree caller aims it at `_EVENT_LOG`, **not** `_EVENT_LOG_LOCAL` (`test_utils.rs:1558`). So the claim this rests on is "nothing reads these on its own", NOT "no possible reader" and NOT "the readers look elsewhere". The rustdoc at `tracing/aof.rs` states it in the same terms.
- **Turning the log OFF now discards existing history at the next restart.** An operator who has been collecting forensic data and then disables the log to stop growth loses what they already collected. To keep it, keep the log enabled, or copy the segments elsewhere before restarting.

## Testing

**`tracing::aof::tests`:**
- `reclaim_segments_removes_only_exact_segment_names` — asserts the returned `EventLogReclaim` equals `{ files_removed: 3, bytes_reclaimed: <exact seeded total>, failures: 0 }`, and that nine near-miss names survive.
- `reclaim_segments_counts_failures_without_panicking` — a *directory* occupying a segment name makes `remove_file` fail deterministically with `EISDIR` on every platform and regardless of whether the test runs as root (a `chmod` would not, since root ignores it).
- `reclaim_segments_reports_nothing_when_there_is_nothing_to_do`, including a data dir that does not exist.
- `reclaim_refuses_while_the_event_log_is_enabled` — **the critical guard**: asserts the decision is `None`, not merely that files survived.
- `reclaim_sweeps_segments_when_the_log_is_disabled` — exact byte total, decoys survive, and a second call reports `default()` (idempotent).
- `reclaim_never_touches_the_local_mode_event_log`.

**`node::tests::event_log_reclaim_wiring`** drives `NodeConfig::new` itself — the real startup entry point — so deleting the call site turns it red.

### Call-site mutation results

| Mutation | Result |
|---|---|
| Delete `reclaim_orphaned_event_log(&config)` from `NodeConfig::new` | **RED** — `node_startup_reclaims_orphaned_segments_when_log_disabled` |
| Delete the `event_log_enabled()` guard | **RED** — 2 tests (aof policy + node wiring) |
| Delete the Network-mode guard | **RED** — `reclaim_never_touches_the_local_mode_event_log` |

Local gates on the head commit: `cargo check -p freenet --all-targets` = 0, `cargo test` = 28 passed / 0 failed, `cargo clippy -p freenet --all-targets` = 0, `cargo fmt --check` clean.

## Review findings addressed

Full-tier review (skeptical + code-first lenses, blind, read-only). No blocking findings. Four fixed:

1. **The rustdoc claimed the segments "have no consumer at all"** — false, as above. Both lenses flagged it independently; a load-bearing justification for an irreversible delete now states the true, checkable claim.
2. **A sweep that removed nothing logged success.** `is_empty()` was false whenever `failures > 0`, so an all-failures sweep emitted `info!("reclaimed orphaned event-log segments", files_removed = 0)` — reading as success exactly when an operator needs to know bytes are still there. Now a reclaim logs `info!` and an all-failures sweep logs a distinct `warn!`. This left `is_empty` dead, so it is removed.
3. **The sole real-startup caller is now recorded at the call site.** `bin/freenet.rs:244` (`run_network`) is the only caller of `NodeConfig::new` on the real-node startup path, so it is the only way the sweep reaches a real data dir. An earlier draft of this note gave the wrong reason for why the rest are safe — it said they are all `#[cfg(test)]` fixtures or the `#[freenet_test]` harness and that all of them get a temp data dir from `ConfigPathsArgs::default_dirs`. Neither half holds: `node/testing_impl.rs` and `test_utils` are not cfg-gated and compile into release builds, the `crates/core/tests/*` integration tests are neither fixtures nor the harness, and none of those reach `default_dirs`. All 40+ call sites were re-checked; the invariant that actually holds is that each non-`run_network` caller sets an explicit `tempfile` data dir (the `#[freenet_test]` harness and every `crates/core/tests/*` test), **or** uses `mode: Local` which the Network-mode guard refuses (`node/testing_impl.rs`), **or** sets `enable_event_log: Some(true)` which the enabled guard refuses (also the harness). The call-site comment now states that invariant instead.
4. **The unlink site names its round-trip invariant.** `scan_segment_indices` matches a *name* and returns a parsed index; `reclaim_segments` rebuilds the name from that index. That is safe only because the grammar is exactly 10 ASCII digits and `segment_path` formats to the same width. Loosening the grammar without carrying the matched `OsString` would delete a different file from the one that matched — now stated where someone would make that edit.

## Scope

This PR previously bundled a **webapp-cache startup sweep** as well, which the rule-review flagged (`git-workflow.md`, one logical change per PR). That warning was correct. The webapp-cache half has been split out and will follow as its own PR; this PR is the event-log reclaim only.

### Known limitation: two nodes sharing one data dir

There is no lock on the data dir, and the sweep does not take one. `check_for_existing_process` (`bin/freenet.rs:1010-1038`) guards only the WS API port, so a node started on an already-in-use data dir with a *different* `--ws-api-port` will sweep the segments the first node has open and is still appending to.

This is not a new hazard class — two nodes on one data dir already corrupt the redb state store, and the log must have been deliberately enabled for there to be anything to lose — but it is worth stating that this change makes that configuration destructive rather than merely broken. Not fixed here; a cheap future guard would be to skip any segment whose mtime is within the last minute, using the `metadata` call the sweep already makes.

### Not covered by any test: does this fire in the field?

No CI test can verify that a real 0.2.11x peer's orphaned segments actually live at `<data_dir>/_EVENT_LOG.NNNNNNNNNN`. If they do not, the reclaim is a silent no-op in production and every test still passes — a reclaim that quietly reclaims nothing is indistinguishable from one with nothing to do.

The `info!` at `tracing/aof.rs::reclaim_orphaned_event_log` is the instrument for this, and it is visible in a release build (`info!`, not `debug!`, and the logger is installed before `run_network`). **After this ships, check a gateway's log on its first restart for `files_removed > 0`.** If the line never appears on a peer known to be carrying segments, the path assumption is wrong.

### Known limit (deliberate)

A *legacy monolithic* `_EVENT_LOG` base file — pre-#3992 format, folded into segment 0 by `LogFile::migrate_legacy` on the first open with the log enabled — is not reclaimed, because this change never deletes the base file. `migrate_legacy` runs on the *open* path, which never executes while the log is disabled, so on a node old enough to have one it survives. It would only be non-empty on a peer that skipped the entire April–July window and upgraded straight to 0.2.112+ with the log already off; on both measured peers the base file is 0 bytes.

Closes #5035

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -170,6 +170,21 @@ async fn run(config: Config) -> anyhow::Result<()> {
         "Freenet node starting"
     );
 
+    // Reclaim the unpacked-webapp cache before this node serves anything
+    // (#5035). #5012 bounded that cache at 64 MiB but wired enforcement only to
+    // the web handlers, so a node that serves no webapp requests — a gateway,
+    // normally — never enforced the bound and kept its whole legacy cache (1236
+    // MiB measured) forever. This is the missing trigger.
+    //
+    // It lives HERE, in the binary, rather than beside the cache in
+    // `serve_client_api_in_impl`, because a dozen in-process integration tests
+    // call that with `webapp_cache_dir` resolved to the developer's REAL
+    // `~/.cache/freenet/webapp_cache`; sweeping there would make `cargo test`
+    // evict it, the exact regression #5012 fixed twice. See the rustdoc on
+    // `reclaim_webapp_cache_at_startup`. Runs for both modes: `freenet local`
+    // shares the same per-user cache directory.
+    freenet::server::reclaim_webapp_cache_at_startup(config.ws_api.webapp_cache_dir.clone()).await;
+
     match config.mode {
         OperationMode::Local => run_local(config).await,
         OperationMode::Network => run_network(config).await,
@@ -1295,6 +1310,84 @@ fn main() {
 mod tests {
     #[cfg(target_os = "linux")]
     use super::parse_listening_inode;
+
+    /// Production half of this file, comment-stripped and whitespace-collapsed,
+    /// for the source pins below.
+    ///
+    /// Cut at `mod tests {`, NOT at `#[cfg(test)]`, which also sits on
+    /// individual test-only items. Only lines whose TRIMMED prefix is `//` are
+    /// dropped, so a `//` inside a string literal (this file has URLs) cannot
+    /// truncate a real line of code — the #5026 item-2 trap. Whitespace is then
+    /// collapsed so rustfmt re-wrapping a call cannot rot a needle. The cut
+    /// needle is split so it cannot match its own source through `include_str!`.
+    fn production_source() -> String {
+        const FULL: &str = include_str!("freenet.rs");
+        let cutoff = FULL
+            .find(concat!("mod ", "tests {"))
+            .expect("freenet.rs must have a test module");
+        FULL[..cutoff]
+            .lines()
+            .filter(|line| !line.trim_start().starts_with("//"))
+            .collect::<String>()
+            .chars()
+            .filter(|c| !c.is_whitespace())
+            .collect()
+    }
+
+    /// #5035: `run` must reclaim the webapp cache before it starts a node.
+    ///
+    /// This is a SOURCE pin because the call site cannot be reached from a
+    /// test: `run` starts a real node, and the reason the call lives in the
+    /// binary at all is that no in-process test may reach it (an in-process
+    /// caller resolves the developer's real `~/.cache/freenet/webapp_cache` and
+    /// would have `cargo test` evict it — see the rustdoc on
+    /// `reclaim_webapp_cache_at_startup`). The sweep's own behaviour is covered
+    /// by the `server::path_handlers` tests; what is unpinnable by behaviour,
+    /// and pinned here, is that `run` still calls it.
+    ///
+    /// Comments are stripped before scanning on purpose: the call site's own
+    /// comment names the function in prose, so an un-stripped scan would stay
+    /// green with the call deleted.
+    #[test]
+    fn run_reclaims_the_webapp_cache_before_starting_a_node() {
+        let src = production_source();
+        let sig = concat!("asyncfn", "run(config:Config)");
+        let start = src
+            .find(sig)
+            .expect("pin guard: `run` not found in the production slice");
+        let after = &src[start + sig.len()..];
+        let end = after
+            .find(concat!("asyncfn", "run_local("))
+            .expect("pin guard: `run_local` follows `run`; slice extraction is wrong");
+        let body = &after[..end];
+
+        // Vacuity guard: prove the slice is the real body of `run` rather than
+        // an empty or mis-sliced range, which would pass no matter what.
+        assert!(
+            body.contains(concat!("OperationMode::Local=>", "run_local(config)")),
+            "pin guard: the extracted slice is not `run`'s body, so this pin \
+             proves nothing. Slice:\n{body}"
+        );
+
+        let call = concat!(
+            "reclaim_webapp",
+            "_cache_at_startup(config.ws_api.webapp_cache_dir"
+        );
+        assert!(
+            body.contains(call),
+            "`run` must reclaim the webapp cache before starting a node (#5035). \
+             Without this call the 64 MiB bound from #5012 is enforced only by \
+             web requests, so a node that serves none — a gateway — never \
+             enforces it at all. Slice:\n{body}"
+        );
+        // Ordering: the sweep must precede the mode dispatch, so it completes
+        // before anything can serve from (or write to) the cache.
+        assert!(
+            body.find(call) < body.find("matchconfig.mode"),
+            "the reclaim must run BEFORE the mode dispatch starts a node. \
+             Slice:\n{body}"
+        );
+    }
 
     #[test]
     fn auto_update_default_stays_enabled_flag_and_dirty_disable() {
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -5303,6 +5303,41 @@ async fn load_gateways_from_index(url: &str, pub_keys_dir: &Path) -> anyhow::Res
     Ok(gateways)
 }
 
+/// Test-only: build a `ConfigArgs` rooted at `dir` in the given mode, ready to
+/// `build()` into a real `Config` whose data dir is `dir`.
+///
+/// Lives at module level rather than inside `mod tests` so the event-log tests
+/// in `tracing::aof` and `node` can share one definition of "a config that
+/// builds" with the `#[cfg(test)]` config tests here — the three modules must
+/// agree on the shape or they stop testing the same thing.
+#[cfg(test)]
+pub(crate) fn event_log_test_args(dir: &std::path::Path, mode: OperationMode) -> ConfigArgs {
+    ConfigArgs {
+        mode: Some(mode),
+        // A non-gateway network node with no gateways is rejected by
+        // `build()`, so the network-mode cases build as a gateway. That is
+        // the realistic shape anyway: a gateway IS a network-mode node, and
+        // it is exactly the kind of node we operate and want the log on.
+        network_api: {
+            let is_network = matches!(mode, OperationMode::Network);
+            NetworkArgs {
+                is_gateway: is_network,
+                // A gateway must declare a public address.
+                public_address: is_network.then(|| "203.0.113.1".parse().unwrap()),
+                public_port: is_network.then_some(31337),
+                skip_load_from_network: true,
+                ..Default::default()
+            }
+        },
+        config_paths: ConfigPathsArgs {
+            config_dir: Some(dir.to_path_buf()),
+            data_dir: Some(dir.to_path_buf()),
+            log_dir: Some(dir.to_path_buf()),
+        },
+        ..Default::default()
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use httptest::{Expectation, Server, matchers::*, responders::*};
@@ -6617,30 +6652,7 @@ shutdown-drain-secs = 42
     /// Build a `ConfigArgs` rooted at `dir` in the given mode. Shared by the
     /// #4968 event-log default tests so each case differs only in what it sets.
     fn event_log_args(dir: &std::path::Path, mode: OperationMode) -> ConfigArgs {
-        ConfigArgs {
-            mode: Some(mode),
-            // A non-gateway network node with no gateways is rejected by
-            // `build()`, so the network-mode cases build as a gateway. That is
-            // the realistic shape anyway: a gateway IS a network-mode node, and
-            // it is exactly the kind of node we operate and want the log on.
-            network_api: {
-                let is_network = matches!(mode, OperationMode::Network);
-                NetworkArgs {
-                    is_gateway: is_network,
-                    // A gateway must declare a public address.
-                    public_address: is_network.then(|| "203.0.113.1".parse().unwrap()),
-                    public_port: is_network.then_some(31337),
-                    skip_load_from_network: true,
-                    ..Default::default()
-                }
-            },
-            config_paths: ConfigPathsArgs {
-                config_dir: Some(dir.to_path_buf()),
-                data_dir: Some(dir.to_path_buf()),
-                log_dir: Some(dir.to_path_buf()),
-            },
-            ..Default::default()
-        }
+        super::event_log_test_args(dir, mode)
     }
 
     /// #4968: a network-mode node (what end users run) must NOT write the local
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -4045,6 +4045,41 @@ async fn load_gateways_from_index(url: &str, pub_keys_dir: &Path) -> anyhow::Res
     Ok(gateways)
 }
 
+/// Test-only: build a `ConfigArgs` rooted at `dir` in the given mode, ready to
+/// `build()` into a real `Config` whose data dir is `dir`.
+///
+/// Lives at module level rather than inside `mod tests` so the event-log tests
+/// in `tracing::aof` and `node` can share one definition of "a config that
+/// builds" with the `#[cfg(test)]` config tests here — the three modules must
+/// agree on the shape or they stop testing the same thing.
+#[cfg(test)]
+pub(crate) fn event_log_test_args(dir: &std::path::Path, mode: OperationMode) -> ConfigArgs {
+    ConfigArgs {
+        mode: Some(mode),
+        // A non-gateway network node with no gateways is rejected by
+        // `build()`, so the network-mode cases build as a gateway. That is
+        // the realistic shape anyway: a gateway IS a network-mode node, and
+        // it is exactly the kind of node we operate and want the log on.
+        network_api: {
+            let is_network = matches!(mode, OperationMode::Network);
+            NetworkArgs {
+                is_gateway: is_network,
+                // A gateway must declare a public address.
+                public_address: is_network.then(|| "203.0.113.1".parse().unwrap()),
+                public_port: is_network.then_some(31337),
+                skip_load_from_network: true,
+                ..Default::default()
+            }
+        },
+        config_paths: ConfigPathsArgs {
+            config_dir: Some(dir.to_path_buf()),
+            data_dir: Some(dir.to_path_buf()),
+            log_dir: Some(dir.to_path_buf()),
+        },
+        ..Default::default()
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use httptest::{Expectation, Server, matchers::*, responders::*};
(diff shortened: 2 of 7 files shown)


sanity (commented):

Reviewed at head `3533e4c83` in a dedicated read-only worktree. Three blind, independent lenses ran in parallel, each told not to rubber-stamp: **skeptical** (assume bugs exist), **code-first** (read code before description, flag intent/impl mismatch), **testing** (coverage gaps and vacuous tests). No external non-Claude pass was run; three diverse Claude lenses were substituted, which is recorded here per the review rule. Every `file:line` below was re-verified by the synthesiser against the checked-out code; findings that did not check out are listed under "Dismissed".

**Verdict: the deletion itself is sound.** The predicate is tight, the match→rebuild round-trip is byte-exact, both guards are load-bearing, and the wiring test would go red if the call site were removed. The two MAJORs are both *description* accuracy on the justification for an irreversible delete, not defects in the delete.

---

### What I verified closed (the hazards this PR had to answer)

**Matching predicate.** `scan_segment_indices` (`crates/core/src/tracing/aof.rs:227-258`) requires `strip_prefix("_EVENT_LOG.")` **and** `rest.len() == 10` **and** every byte `is_ascii_digit()` **and** `parse::<u32>()` succeeding (`aof.rs:248-252`). It is the same function `LogFile::open` calls (`aof.rs:146`) — the grammar is genuinely reused, not re-invented as a glob.

**Match-then-rebuild round-trip is safe, including the non-UTF-8 angle.** `scan_segment_indices` matches on a lossy `String` (`aof.rs:247`) while `segment_path` (`aof.rs:214-225`) rebuilds from the base's raw `OsString` plus `format!(".{:010}")`. `to_string_lossy` only ever *inserts* U+FFFD — it never synthesises ASCII — so a lossy name of exactly `_EVENT_LOG.` + 10 ASCII digits proves the raw bytes were exactly that. No aliasing. The comment at `aof.rs:671-678` names this invariant at the unlink site, which is the right place for it.

**Path traversal is unreachable.** `ConfigPathsArgs::build` unconditionally re-derives `event_log = app_data_dir.join("_EVENT_LOG")` (`config.rs:3567`), and `ConfigPathsArgs::merge` (`config.rs:3444-3448`) carries only `config_dir`/`data_dir`/`log_dir` — a persisted `event_log` key is not honoured. So `base.file_name()` is always literally `_EVENT_LOG`, the sweep is confined to the data dir, and the `.expect(...)` at `aof.rs:221`/`aof.rs:234` cannot fire.

**Both guards are load-bearing.** `aof.rs:772` (`event_log_enabled()`) and `aof.rs:775` (`OperationMode::Network`). The second is not decorative: in Local mode `Config::event_log()` resolves to `_EVENT_LOG_LOCAL` (`config.rs:3784-3786`), so without it an explicit `--enable-event-log false` on a local node would sweep exactly what `fdev verify-state` consumes.

**Malformed config fails closed.** `read_config` maps a toml error to `io::ErrorKind::InvalidData` (`config.rs:582-605`) and `build()` propagates it — the node never starts, so nothing is deleted. A missing `enable_event_log` field does not default to "enabled": it is `Option<bool>` merged only when `Some` (`config.rs:924-926`, `config.rs:3800-3803`).

**Silent-no-op is closed at the wiring level.** `node.rs:14167-14192` drives the real entry point `NodeConfig::new` and asserts the **observable outcome** (files gone from disk), with a real precondition at `node.rs:14181`. Deleting the call at `node.rs:538`, inverting the enabled guard, or no-opping `reclaim_segments` each turns a test red. `bin/freenet.rs:244` is inside `run_network` and is the only real-node-startup caller (I enumerated all 40+ `NodeConfig::new` call sites; every other one uses an explicit `tempfile` dir, Local mode, or `enable_event_log: Some(true)` in the `#[freenet_test]` harness at `freenet-macros/src/codegen.rs:326`).

**No new source-scrape tests.** The only `include_str!` in the touched files is the pre-existing `write_all_does_not_fsync_per_batch` (`aof.rs:1090-1123`), which this PR does not disturb — its slice ends at `"\n    pub fn "`, and `reclaim_segments` is `fn`, not `pub fn`, and sits after `encode_batch`. It also carries its own vacuity guard (`aof.rs:1111`). No #5450-shaped self-satisfying pin was introduced.

**`config.rs` is a pure `#[cfg(test)]` move.** The 24 deletions are the move-out of `event_log_args`' body to a module-level `event_log_test_args` (`config.rs:5314`); I diffed the bodies and they are byte-identical. No production default, field, persisted key, or `build()` behaviour changes.

---

### MAJOR — description, not code

**1. The PR body's reader claim is false, and the code's own rustdoc says so.**

The body states *"both in-tree readers (`fdev verify-state --log-directory`, `EventLogAggregator`) discover the Local-mode `_EVENT_LOG_LOCAL` instead."* Neither half holds:

- `crates/fdev/src/verify_state.rs:180-190` — the **file** branch accepts any name whose `to_string_lossy().contains("EVENT_LOG")`. `_EVENT_LOG.0000000000` matches. Only the *directory* branch (`verify_state.rs:172`) is `_EVENT_LOG_LOCAL`-specific.
- `crates/core/src/test_utils.rs:1558` — `event_log_path` joins `"_EVENT_LOG"`, not `_EVENT_LOG_LOCAL`, and `test_utils.rs:2175` feeds it to the aggregator.

The rustdoc at `aof.rs:730-745` states this **correctly** ("does pick these up"; "aims it at `_EVENT_LOG` deliberately"). Only the PR body regressed. This is the paragraph that justifies an irreversible delete, and the body is the durable record, so it is worth correcting before merge. Not blocking: the code is right and the sweep's safety does not depend on the claim.

**2. The caller audit is wrong on both halves, in the body and in the code comment.**

Body item 3 and `node.rs:527-530` both say *"every other caller is a `#[cfg(test)]` fixture or the `#[freenet_test]` harness, and both get a temp data dir from `ConfigPathsArgs::default_dirs`."*

- `crates/core/src/node/testing_impl.rs:2785` and `:2888` are **not** `#[cfg(test)]` — the first `#[cfg(test)]` in that file is line 6593, and `node.rs` declares `pub(crate) mod testing_impl;` unconditionally. They compile into release builds. (Safe: `id: Some(label)` plus `mode: Some(OperationMode::Local)` at `testing_impl.rs:2774-2775` trips guard 2.)
- Eight files under `crates/core/tests/` call `NodeConfig::new` directly (`hosted_mode_{migrate,export,import,isolation}.rs`, `operations_before_join.rs`, `in_process_restart.rs`, `error_notification.rs`, `origin_attestation_scope.rs`). They are neither `#[cfg(test)]` fixtures nor the harness, and none reach `default_dirs` — each sets an explicit `tempfile` data dir.
- The harness (`freenet-macros/src/codegen.rs:309-311`) sets explicit temp dirs and never reaches `default_dirs`; it also sets `enable_event_log: Some(true)` (`codegen.rs:326`), a stronger protection than the one claimed and unmentioned.

The comment contradicts itself four lines later at `node.rs:532-535`, which correctly notes that `test_utils` and `testing_impl` are not cfg-gated. **The safety conclusion is unaffected** — I checked every caller — but the stated *reason* is not the real invariant, and this comment is exactly the kind of load-bearing justification that rots silently. Suggested replacement: *every non-`run_network` caller either sets an explicit temp `data_dir`, or passes `id: Some(..)` with `mode: Local`, and the `#[freenet_test]` harness additionally sets `enable_event_log: Some(true)`.*

---

### MINOR

**3. Two nodes on one data dir: the second destroys the first's live event log.** `crates/core/src/bin/freenet.rs:1010-1038` — `check_for_existing_process` is a TCP connect to the WS API port, not a lock on the data dir, and `reclaim_segments` (`aof.rs:656`) takes no `FILE_LOCK`. Node A running with `--enable-event-log`, node B started on the same `--data-dir` with a different `--ws-api-port`, and B unlinks A's active segments at `aof.rs:684` while A keeps appending to an unlinked inode. Mitigating: sharing a data dir already corrupts the redb, and the log must have been deliberately enabled — but this converts "two instances is broken" into "two instances is destructive". A cheap guard is to skip any segment whose mtime is within, say, 60s, using the `metadata` call already made at `aof.rs:683`.

**4. `std::fs::metadata` follows symlinks; `remove_file` does not.** `aof.rs:683` vs `aof.rs:684`. Unlinking a segment-named symlink is the *correct* behaviour (the target is untouched), but `bytes_reclaimed` would report the target's size when zero bytes were freed. `symlink_metadata` is the right call. Cosmetic today, but this repo has a standing rule that a number describing an action must be produced by the action.

**5. `failures` conflates two different outcomes.** `aof.rs:666` (directory scan failed, nothing attempted) and `aof.rs:691` (one file resisted) increment the same counter. The two `warn!` messages differ so logs are distinguishable, but the struct field a future metric export would read is not. `aof.rs:711-715` documents the ambiguity rather than removing it.

**6. Blocking `std::fs` I/O inside `async fn NodeConfig::new`.** `node.rs:538` → `aof.rs:658`/`683`/`684` run `read_dir` plus N `metadata`/`remove_file` syscalls inline on a tokio worker; the rest of `aof.rs` uses `tokio::fs` deliberately. Startup-only and ~24 unlinks in the measured case, so the practical cost is small, but `spawn_blocking` costs nothing here.

**7. No test asserts that non-`_EVENT_LOG` files survive.** `aof.rs:871-889` lists nine decoys, every one prefixed `_EVENT_LOG`. `reclaim_sweeps_segments_when_the_log_is_disabled` (`aof.rs:1014`) runs against a dir that `build()` has populated with `config.toml`, `db/`, `secrets/`, `contracts/` — and asserts survival of none of them. A mutation of the shape "delete every entry that does *not* match the grammar" passes the whole suite. Contrived, but three lines to close, and this is an irreversible delete inside a user's data dir.

**8. The legacy-monolithic miss is silent.** `reclaim_orphaned_event_log` logs nothing when `files_removed == 0 && failures == 0` (`aof.rs:791-808`), which is the correct default — but it is also the case for a node carrying a non-empty pre-segmentation `_EVENT_LOG` base file, where the PR's own goal is missed and the operator gets no explanation. Documented as a deliberate known limit; noting it because "no log" and "nothing to do" are indistinguishable to an operator running `du`.

**9. Return value discarded.** `node.rs:538` drops the `Option<EventLogReclaim>`; the struct is not `#[must_use]`, so a future caller silently ignoring a non-zero `failures` would not be caught.

---

### NIT

- `aof.rs:1065` — `assert!(bytes > 0);` where `bytes` is `1_000 + 1_000`. Unfalsifiable; it reads like a precondition and establishes nothing. The honest precondition is asserting the two `_EVENT_LOG_LOCAL.*` files exist before the sweep.
- `aof.rs:639` — the rustdoc summary "Unlink every `<base>.NNNNNNNNNN` segment" is slightly overbroad: `aof.rs:250` also requires `parse::<u32>()`, so `_EVENT_LOG.9999999999` matches the stated grammar but is not swept. Consistent with `open` and unproducible by the writer, so this is wording only.
- `node.rs:14195` — `node_startup_keeps_segments_when_log_enabled` omits the `segments_present == 2` precondition its sibling has at `node.rs:14181`. Not vacuous, but the asymmetry is gratuitous.
- Nothing pins the `info!`/`warn!` split at `aof.rs:791`/`aof.rs:801`, which was review round 2's finding. A regression to "log success on an all-failures sweep" would not be caught.
- `node/testing_impl/network.rs:76` and `:82` construct `EventRegister::new(...)` unconditionally, unlike the gated production site at `node.rs:822`. Harmless (SimNetwork peers are Local mode with temp dirs), but it is the shape where a guard stops being universal.

---

### Dismissed, with reasons

- **Non-UTF-8 / WTF-8 filenames manufacturing a false match** — impossible; `to_string_lossy` only inserts U+FFFD, never ASCII digits.
- **Path traversal via a crafted `event_log` base name** — unreachable; the path is re-derived at `config.rs:3567` and not merged from persisted config.
- **A malformed or partially-written `config.toml` bypassing a guard** — fails closed at `config.rs:582-605`; the node does not start.
- **A missing `enable_event_log` field defaulting toward deletion** — `Option<bool>`, merged only when `Some` (`config.rs:924-926`).
- **`config.rs`'s 24 deletions changing a production default** — verified byte-identical `#[cfg(test)]` move.
- **A request for a cross-file source pin asserting `run_network` still calls `NodeConfig::new`** — downgraded and not adopted. If that call vanished the node would not start at all, so this is not a silent-rot class; and a source scrape here is precisely the #5450 shape the repo is trying to stop adding.
- **Symlink loops / unbounded recursion in the sweep** — there is no recursion; `scan_segment_indices` reads one directory level.
- **Deleting a segment out from under a concurrent in-process reader** — the sweep runs at `node.rs:538`, strictly before `EventRegister::new` at `node.rs:822`, and that path only exists when the log is enabled, in which case the sweep declined.

---

### One thing worth doing after merge

No CI test can verify that a real 0.2.11x peer's orphaned segments actually live at `<data_dir>/_EVENT_LOG.NNNNNNNNNN`. If they do not, the reclaim is a silent no-op in production and every test still passes. The `info!` at `aof.rs:792` is the right instrument and will be visible in a release build — **check a real gateway's log for `files_removed > 0` on the first restart after this ships.**

Lenses: skeptical, code-first, testing (three diverse Claude lenses; no external model pass, substituted per the review rule). All cited locations re-verified against `3533e4c83`.

[AI-assisted - Claude]

PR #5044, opened 2026-07-29
fix(server): block web-asset path traversal and stop spurious CORS errors on decoded filenames

### LINKED ISSUE #2: NAT traversal

sanity:
## Problem

Two bugs in the web-contract asset serving path (`variable_content` in `crates/core/src/server/path_handlers.rs`), both rooted in how the already-percent-**decoded** `{*path}` wildcard segment is handled.

**1. Spurious CORS error on filenames with spaces (user report).** cirro / SUB0PT1MAL reported on Matrix (2026-07-29) that Delta assets whose names contain a space (e.g. `my image.png`) fail with a CORS error. The handler reconstructs a URL string from the decoded segment and **re-parses it as an `axum::http::Uri`**. A space (also `<`, `>`, backticks) is a valid filename byte but an invalid raw-URI character, so the parse fails with a `400`. Because that error response carried no `Access-Control-Allow-Origin` header, the sandboxed iframe's null-origin subresource fetch surfaced it to the app as an opaque "CORS error" that masked the real `400`.

**2. Unauthenticated path traversal / arbitrary local-file read (found while investigating #1).** `variable_content` joined the decoded relative path onto the contract cache dir with **no containment check** — unlike its HTML sibling `sandbox_content_body`, which has one. `..%2f..%2fetc%2fpasswd` decodes to `../../etc/passwd`; `%2fetc%2fpasswd` decodes to the absolute `/etc/passwd` (and `Path::join` with an absolute path replaces the base entirely). Both escape the cache and serve the file.

Confirmed against a local loopback node:

| Request | Before this PR |
|---|---|
| `.../my%20image.png` | `400`, no `Access-Control-Allow-Origin` → opaque CORS error in-app |
| `.../..%2f..%2f..%2fetc%2fhostname` | `200 OK` serving `/etc/hostname`, with `Access-Control-Allow-Origin: *` |

**Severity: critical, and remotely exploitable on hosted/proxied nodes** (corrected during review — the original framing understated this). Two exposure paths:

- **Remote (hosted nodes):** the HTTP API supports wildcard binds (`0.0.0.0` / `::`, `server.rs`), and hosted mode is documented as reachable over the internet via the node's reverse proxy (`hosted_migrate.rs`). The web-asset route is on that public surface, so `https://<hosted-node>/v1/contract/web/<anykey>/..%2f..%2fetc%2fpasswd` was a straight **remote unauthenticated arbitrary-file-read**. try.freenet.org and hosted gateways are exactly such nodes.
- **Client-side (any node):** this path's response carries `Access-Control-Allow-Origin: *`, so a **malicious web contract's sandboxed iframe JS can read arbitrary local files off a visitor's machine cross-origin**.

Given the severity, this likely warrants a GitHub security advisory (GHSA), matching the project's prior pattern, and prioritizing the hosted-gateway upgrade after release.

## Approach

- **Extract the asset path textually** (`relative_asset_path`) instead of round-tripping through `Uri`, so decoded filenames with URI-invalid bytes survive. Fixes #1.
- **Add `resolve_web_asset_path`**, a traversal guard that rejects `..`, root, and drive-prefix components before joining, plus a canonicalize/containment symlink (TOCTOU) guard when the target exists — mirroring `sandbox_content_body`. A genuinely-missing asset still falls through to a normal `404`. Fixes #2.
- **Attach the sandbox CORS headers to both success and error responses** in `web_subpages`, so any error on the subresource path surfaces as a real status to the app rather than an opaque CORS failure.

The now-redundant `get_file_path` wrapper is removed; its tests point at `relative_asset_path`.

## Testing

New regression tests, **each verified to fail without its corresponding fix** (reverted each mechanism, confirmed the test fails, restored):

- `variable_content_serves_asset_with_space_in_filename`
- `variable_content_rejects_parent_dir_traversal`
- `variable_content_rejects_absolute_path_escape`
- `web_subpages_error_response_carries_cors_header`
- `relative_asset_path_preserves_space_in_filename`

`cargo fmt` clean; clippy clean on the touched files; all 152 server-module unit tests pass.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -586,6 +586,11 @@ async fn web_subpages(
     // (user report: SUB0PT1MAL / cirro, 2026-07-29). Building the response from
     // the error here, and then adding the headers, lets the app see the true
     // status with CORS allowed.
+    //
+    // SECURITY: error bodies on this route are now cross-origin-readable by a
+    // malicious contract's iframe JS. Keep them non-sensitive — reflected
+    // request path and generic io/parse messages only. Any new error path here
+    // MUST NOT embed internal filesystem paths, config, or secrets.
     let mut response = match result {
         Ok(r) => r.into_response(),
         Err(e) => e.into_response(),
@@ -740,15 +745,33 @@ async fn serve_sandbox_response(
         return redirect_to_shell_root(&key, api_version, None);
     }
 
-    let contract_response = path_handlers::serve_sandbox_content(
+    let contract_response = match path_handlers::serve_sandbox_content(
         key,
         api_version,
         sub_path,
         request_sender,
         webapp_cache,
     )
-    .await?;
-    let mut response = contract_response.into_response();
+    .await
+    {
+        Ok(r) => r.into_response(),
+        Err(e) => {
+            // Same null-origin reasoning as `web_subpages`: an error subresource
+            // response without CORS is surfaced by the browser as an opaque
+            // "CORS error" inside the iframe, masking the real status. Attach the
+            // sandbox CORS headers so the app sees the true 4xx/5xx. (CSP is only
+            // meaningful on served content, so it is skipped for the error.)
+            //
+            // SECURITY: as in `web_subpages`, these error bodies are now
+            // cross-origin-readable by a malicious contract's iframe JS. Keep
+            // them non-sensitive — any new error path here MUST NOT embed
+            // internal filesystem paths, config, or secrets.
+            let mut response = e.into_response();
+            add_sandbox_cors_headers(&mut response);
+            return Ok(response);
+        }
+    };
+    let mut response = contract_response;
     add_sandbox_cors_headers(&mut response);
     // See `sandbox_csp_for_origin` for why we interpolate a concrete origin
     // rather than using `'self'`, and `sandbox_origin_from_headers` for why we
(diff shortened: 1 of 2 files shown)


sanity:
## Multi-perspective review — complete

Ran a Full-tier review (this touches an auth/security surface): **4 independent blind Claude reviewers** with distinct lenses — security-bypass (adversarial), testing, code-first correctness, big-picture/CORS-model — plus a **focused adversarial re-review of the review-response delta**. Verdict: **no blocking findings; the core traversal fix is correct and the threat model is complete** (only two filesystem sinks exist — `variable_content`, now guarded, and `sandbox_content_body`, already guarded).

### Findings addressed in `ad1377f1`

- **TOCTOU parity (medium; flagged by all 4):** `resolve_web_asset_path` now returns the **canonical** path when the target exists, so the caller opens the already-resolved path — closing the check→open symlink window, matching `sandbox_content_body`. (Verified: no added canonicalize cost — the base already canonicalized both sides.)
- **CORS-on-error consistency (low):** `serve_sandbox_response` now attaches the sandbox CORS headers to its **error** path too, so an HTML sandbox subresource error is a real status, not an opaque CORS failure — the same bug's sibling.
- **Test coverage (high/medium):** added `variable_content_rejects_symlink_escape` (pins the canonicalize half, which had **zero** coverage — verified it fails if that half is removed) and `resolve_web_asset_path_boundary_table` (mid-`..`, absolute, and **positive** nested-asset coverage, so a hardening regression can't silently 400 every real webapp). Fixed a process-global cache-key collision in the CORS test.
- **Guard comments:** cross-referenced the two containment guards; documented that CORS-decorated error bodies are now cross-origin-readable and must stay non-sensitive.

### Severity correction

The original "loopback-only → client-side" framing understated this. The HTTP API supports wildcard binds and hosted mode is internet-reachable via the reverse proxy, so this was a **remote unauthenticated arbitrary-file-read on hosted/proxied nodes** (try.freenet.org, hosted gateways). PR description updated. Recommend a GHSA and prioritizing the hosted-gateway upgrade after release.

### Non-blocking, deliberately not changed (to avoid re-staling CI/review for very-low items)

- Boundary-table "accept" assertions are trivially-true for non-existent targets — real containment-on-existing-file is covered by the symlink (reject) and space-in-filename (accept-and-serve) tests, so there's no coverage gap.
- Dangling-symlink (`x -> /outside/notyet`) returns the unresolved path — **pre-existing** (not introduced here), and not exploitable: a contract author has no write primitive to create the outside target in the check→open window, and any pre-existing target is caught by canonicalize.
- The traversal tests use `Err(_) => false` (leak-focused) rather than asserting the exact error variant — optional hardening; the fixtures ensure the guard is actually reached today.

Lenses run: security-bypass, testing, code-first, big-picture, + delta re-review.

[AI-assisted - Claude]

Issue #3046, opened 2026-02-16
Investigate: 49% of room subscribers miss update broadcasts


sanity:
## Problem

Funnel analysis of River telemetry (10h window, ~850 peers) shows that **77 out of 158 peers (49%)** who successfully subscribed to the official room contract never received update broadcasts that happened while they were online.

### Funnel data

| Stage | Count | % |
|-------|-------|---|
| Peer connected | 843 | 100% |
| River container loaded | 307 | 36.4% |
| Official room joined | 211 | 25.0% |
| Received room updates | 93 | 11.0% |
| Sent room updates (active chat) | 5 | 0.6% |

### What the data shows

Of the 77 peers who subscribed but missed broadcasts:

- **30 peers**: Last room-related activity was >5 min before the update — consistent with Interest TTL expiry (5 min TTL in `interest.rs:53`)
- **25 peers**: Last room-related activity was <5 min before the update — Interest TTL does NOT explain these
- **22 peers**: Insufficient timestamp data to determine

For comparison, among the 81 peers who DID receive broadcasts, only 5 had a >5 min gap — suggesting TTL expiry is real but not the only cause.

### Current diagnostic gap

We can see `update_broadcast_emitted` (with target list) and `update_broadcast_received` on individual peers, but we **cannot see why a specific peer was NOT in the broadcast target list**. The broadcasting node's `handle_broadcast_state_change()` makes several filtering decisions that are invisible:

1. Proximity cache peer → address resolution failed (peer reconnected/NAT changed)
2. Interest manager peer → address resolution failed  
3. Peer skipped due to summary comparison (thought they already had the state)
4. Interest expired by TTL sweep before broadcast
5. Peer's subscription registered on a different node than the one broadcasting

### Step 1: Add broadcast delivery telemetry (PR forthcoming)

Add a `update_broadcast_delivery_summary` telemetry event emitted by the broadcasting node with per-broadcast breakdown:

- `proximity_found` / `proximity_resolve_failed`
- `interest_found` / `interest_resolve_failed`  
- `skipped_summary_match`
- `targets_sent` / `send_failed`
- Contract key

This lets us answer: when a broadcast happens, how many subscribers exist, and how many are lost at each stage?

### Step 2: Collect data (1-2 days after deployment)

Re-run the funnel analysis script (`~/river-funnel.py`) with the new telemetry data to identify the dominant failure mode.

### Step 3: Fix based on confirmed cause

Possible fixes depending on findings:
[...]

PR #5055, opened 2026-07-30
fix(interest): stop clobbering cached summaries; add per-contract broadcast byte attribution

### LINKED ISSUE #4672: is_upstream clobber: bundle register_downstream_subscriber + GET into a merge, not a bare register(false)
## Problem

Two production `register_peer_interest(.., is_upstream = false)` sites can overwrite an existing `is_upstream = true` entry (and its cached delta-sync summary) in a **cyclic** topology — a peer that is simultaneously our **upstream** host *and* our **downstream** subscriber/GETter for the same contract:

- `crates/core/src/operations/subscribe.rs:643` — `register_downstream_subscriber` registers the requesting peer as a downstream party.
- `crates/core/src/operations/get/op_ctx_task.rs:2650` — registers a GETting peer as a downstream party.

`register_peer_interest` overwrites the **whole** `PeerInterest`, so registering an existing upstream peer as `is_upstream = false` flips the flag true → false and wipes the summary. That defeats the `send_unsubscribe_upstream` lookup (`get_interested_peers().find(|i| i.is_upstream)`) the same way the `ChangeInterests` arm did before #4671.

Unlike the `ChangeInterests` arm, these two sites register a **genuine downstream party** as `is_upstream = false` — that is the intended semantic on the normal (acyclic) chain, so they were left unchanged in #4671 to keep that fix focused. They only clobber when the same peer is both upstream and downstream for one contract, which does not form under normal chain routing and self-corrects on the next subscribe. So this is lower-severity and **not required for #4671**.

## Proposed fix (clean, class-wide)

Rather than mirror the `get_peer_interest → refresh` guard at each call site (which would drop the legitimate downstream registration), make the **primitive** monotonic in `is_upstream`: on re-registration of an existing entry, `is_upstream = existing.is_upstream || new_is_upstream` (never downgrade true → false), and preserve the cached summary unless a fresh one is supplied. That fixes all current and future `register_peer_interest(.., false)` sites at once — including the two above and the `ChangeInterests` arm — without each caller having to know about the upstream/downstream distinction.

Needs a unit test in `ring/interest.rs` (register(false) over an existing upstream entry must keep `is_upstream = true` + summary) and a check that the downstream-registration paths still register a genuinely new peer.

Related: #4671 (the `ChangeInterests` instance, fixed there), piece-D interest-gated renewal (#4642) which makes the upstream edge routine.

[AI-assisted - Claude]

sanity:
## Problem

Three production sites replaced a peer's `PeerInterest` with a bare `register_peer_interest(.., summary: None)`, wiping the cached delta-sync summary of an already-tracked peer. The entry then reports `NeverPopulated`, which is both wrong (it HAD been populated) and expensive: every subsequent broadcast to that peer falls back to full state until the summary is re-seeded.

This fires at renewal cadence. `SUBSCRIPTION_RENEWAL_INTERVAL` is 120s against an 8-minute lease, and a renewal re-registers through the same outbound-SUBSCRIBE machinery as a client request, so an unguarded site clobbers roughly 30 times per subscribed contract per hour.

**Measured impact.** On the full retained collector corpus (0.2.111–0.2.115, ~52,350 node-hours, exact 60s-rollup counters, no sampling, residual exactly zero), `never_populated` is **89.4% of the tracked-missing arm's sends and 93.4% of its bytes — 16.46% of ALL broadcast bytes fleet-wide**, stable across every version in the window (89.8–96.2%).

Caveat, stated because it bounds the claim: `PeerInterest::new` unconditionally stamps `NeverPopulated` and `register_peer_interest` unconditionally inserts fresh, so the tag conflates "created and never seeded" with "seeded then wiped by re-registration". Both are this PR's domain, so the target holds, but the tag does not separate them and this PR does not claim it does.

This PR targets that cause only. It does **not** address the `full_no_their_summary_untracked` arm (13.8% of broadcast bytes) — the #4953 upsert that covers untracked co-hosts shipped in v0.2.113 and is already live on the fleet.

## Approach

Guard all three sites with refresh-if-present / register-if-absent:

- `subscribe.rs` — both finalizers, via `refresh_peer_interest_with_upstream` (asserts upstream-ness on an existing entry without wiping the summary).
- `subscribe.rs::register_downstream_subscriber` and `get/op_ctx_task.rs` (#4672) — via the plain refresh, deliberately NOT the `_with_upstream` variant: those pass `is_upstream = false`, and asserting that on an existing entry would downgrade a legitimate upstream edge and break `Unsubscribe` routing.

`refresh_peer_interest` now returns whether an entry existed, so the guard is one clone-free atomic call. The `get_peer_interest(..).is_some()` form it replaces deep-copied the cached `StateSummary` (up to ~840 KB for exactly the state-sized-summary contracts that make this path expensive) purely to test presence, and was not atomic with the refresh.

Two further fixes found while validating the above:

- **Converged-skip TTL refresh** (#3046/#3093). Preserving the summary makes the converged-skip reachable for the first time, and that path returned without refreshing the peer's interest TTL. Fixed in BOTH fan-out implementations — the production one and the `simulation_tests`-gated duplicate, which is the body the simulation tests actually execute.
- **Per-contract TOTAL broadcast attribution** (#4979/#5057), with its own cap-overflow counters so a `Delta`-only contract's drop cannot vanish from the schema.

## Testing

`cargo test -p freenet --lib`: 4495 passed, 0 failed.

Source pins, each mutation-tested (mutation applied, anchor verified, test confirmed red, restored):

| pin | mutation | result |
|---|---|---|
| `broadcast_to_single_peer_refreshes_interest_on_every_skip_pin` | drop either skip's refresh | red |
| `broadcast_state_to_peers_uses_semantic_delta_skip` | drop either sim skip's refresh; also drop the production one (anti-drift) | red |
| `change_interests_arm_guards_register_with_refresh_pin` | bare register / unbound guard / revert to two-lookup form | red |
| `subscribe_finalizers_do_not_clobber_cached_summaries` | unbind a finalizer's guard; revert `register_downstream_subscriber` | red |
| `get_interest_registration_does_not_clobber_or_downgrade` | bare register; second unguarded register | red |
| `summaries_arm_writes_summary_outside_staleness_branch_pin` | move the summary write inside `if is_stale`; delete it | red |
| `total_map_cap_overflow_is_reported_separately_from_full_state` | drop the overflow counters | red |

`test_interest_ttl_refresh_on_broadcast` was rewritten. It asserted `late_broadcasts > 0`, using broadcast VOLUME as a proxy for interest liveness — invalid here, because suppressing redundant sends to converged peers is the intended win, so the proxy moves in the same direction as the improvement. Bisect showed the first commit alone flips it (main 218 events / late 30 → 127 / late 0, identical for every later commit). Evidence the quiet is correct, not lost delivery: the #3046 delivery breakdown reads `targets_sent=0, send_failed=0, skipped_summary_match=66` for the late phase; ground-truth convergence is 3/3 with zero diverged on both branch and main; and main shows the same shape at a lower skip ratio. The rewrite asserts those properties directly and its scope is stated in the rustdoc — it is a correctness check, not a regression gate, and the source pins carry both #3093 and this PR's own regression.

## Fixes

Closes #4672

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2870,6 +2870,25 @@ async fn handle_interest_sync_message(
                         // broadcast targets forever. A `None` report keeps the
                         // old clear-only semantics (no entry is created just to
                         // hold `None`).
+                        //
+                        // LOAD-BEARING POSITION, not just a write: this runs
+                        // OUTSIDE the `is_stale` branch, on converged and stale
+                        // peers alike, and it is what refreshes the peer's
+                        // interest TTL here — both `set_summary` and
+                        // `clear_summary` end in `PeerInterest::refresh`. There
+                        // is no explicit `refresh_peer_interest` on this path;
+                        // the refresh is a CONSEQUENCE of the write.
+                        //
+                        // So moving this inside the staleness branch — which
+                        // reads as a pure optimisation ("why rewrite a summary
+                        // we just proved unchanged?") — silently stops
+                        // refreshing every converged peer, and they age out at
+                        // `INTEREST_TTL` while every test stays green. That is
+                        // the #3046/#3093 bug class, which has already appeared
+                        // at two other sites in this same fan-out logic (the
+                        // production and sim-only converged skips, both fixed in
+                        // #5055). Pinned by
+                        // `summaries_arm_writes_summary_outside_staleness_branch_pin`.
                         match their_summary {
                             Some(theirs) => {
                                 op_manager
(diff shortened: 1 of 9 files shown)


sanity:
Expanded to a **third** site after a history review surfaced #4672, which tracks this same clobber class and names a site my issue missed: the remote-GET requester registration at `get/op_ctx_task.rs:3197`.

It needed a *different* guard, which is worth calling out because getting it wrong would be a silent routing bug:

| site | `is_upstream` | guard |
|---|---|---|
| `subscribe.rs:528`, `:829` | `true` | `refresh_peer_interest_with_upstream(.., true)` — must ASSERT the flag, since the bare register it replaces did |
| `get/op_ctx_task.rs:3197` | `false` | plain `refresh_peer_interest` — must NOT assert, or it DOWNGRADES a legitimate upstream edge and breaks `Unsubscribe` routing |

Both directions are now pinned: the guard must be present, and the GET site must not use the `_with_upstream` form.

Also noting for whoever reviews: #4672 proposes a class-wide fix — make the primitive monotonic in `is_upstream` and preserve the cached summary unless a fresh one is supplied — which would cover all five sites at once and remove the possibility of this drifting again. That is a better shape than five hand-guarded call sites. I did not take it here because it changes the primitive's semantics on a routing-critical flag, which deserves its own review rather than riding along with a bug fix. Worth doing as a follow-up.

[AI-assisted - Claude]

sanity:
Third commit added, and it is a **separable** concern from the first two — flagging so a reviewer can take them independently:

| commit | concern | risk |
|---|---|---|
| 1-2 | the three unguarded `register_peer_interest` sites (behaviour) | low, but it is behaviour |
| 3 | per-contract total broadcast byte attribution (telemetry only) | none — fields on an existing 60s rollup |

They ride together because they are one Phase-1 unit of the bandwidth work rather than because they interact. Happy to split if that is easier to review.

**On the telemetry commit.** `contract_full_state_bytes` is written only under `arm.is_full_state()`, making it a numerator with no denominator — and structurally blind to a contract whose entire cost is in the delta arm. That is exactly #5056: a contract at 55.6% of all broadcast sends whose "deltas" are full-state-sized. Attributing it needed a natural experiment over peers that happened to receive only that contract, because no counter could answer directly.

The new `top_contracts_by_total_bytes` closes that, and is the input a per-contract outbound budget (#5057) would be sized from — the point being to choose that ceiling from the observed distribution rather than guessing.

No new event stream, per the constraint that killed #4940 (`MAX_EVENTS_PER_SECOND = 10`/node, 3.11% of records already colliding, collector at ~88.8 GB/day).

Its test asserts both directions — the delta-only contract appears in the total map and outranks a full-state contract, and does *not* appear in the full-state map so the two stay unconflated. Mutation-verified: re-gating the write on `is_full_state()` makes it fail with 'this is the #5056 blind spot'.

[AI-assisted - Claude]

PR #5061, opened 2026-07-30
feat(telemetry): split interest_sync_summaries by emitter

### LINKED ISSUE #5052: feat(telemetry): split interest_sync_summaries by emitter — 49.8% of outbound bytes is currently unattributable
## Summary

`interest_sync_summaries` is **49.8% of all outbound bytes** on the live fleet — the single largest consumer, ahead of contract updates. #4982 split the parent `InterestSync` arm into Interests / Summaries / Resync, which was enough to eliminate resync (0.9%) and the Interests exchange (0.3%) as suspects. It is not enough to attribute the remaining 49.8%, because `InterestMessage::Summaries` has multiple distinct emitters and the arm counts them together.

## Measured

v0.2.115, 1,174 peers, 2,689 node-hours (`outbound_message_mix`):

| arm | share of all outbound bytes | bytes | msgs | mean |
|---|---|---|---|---|
| **`interest_sync_summaries`** | **49.8%** | 511 GB | 12,726,443 | 40.2 KB |
| `update` | 46.4% | 476 GB | 13,167,652 | 36.1 KB |
| `neighbor_hosting` | 2.5% | 26 GB | 5,049,937 | 5.2 KB |
| `interest_sync_resync` | 0.9% | 8.8 GB | 60,123 | 145.8 KB |
| `interest_sync_interests` | 0.3% | 3.5 GB | 2,552,182 | 1.4 KB |

## The emitters that are currently indistinguishable

| # | site | shape | trigger |
|---|---|---|---|
| 1 | `operations/update.rs:877` (`send_proactive_summary_notification`) | **single**-entry | every state change, fanned across interested peers |
| 2 | `node.rs:2700` (reply to `InterestMessage::Interests`) | **multi**-entry, one per advertised hash via `summary_if_hosted_or_in_use` | every heartbeat received |
| 3 | `operations/update.rs:1024` (summary-back on rejection) | single-entry, gated on summaries matching | rejected broadcast, narrow |
| 4 | `node.rs:3107` | — | — |

Emitters 1 and 2 are both plausibly large and have **opposite** fixes:

- If **1** dominates, #5003 (skip co-hosts already covered by the broadcast, no wire change) is most of the answer.
- If **2** dominates, the fix is hash-first (#4965): exchange digests and ship summary bytes only on mismatch, which is a wire-format change. Note emitter 2 is multi-entry, so a single reply can be much larger than a notification.

The best current evidence is #5003's own: r=0.92 between UPDATE messages and InterestSync messages per 60s window, plus a ~1/5 floor in zero-UPDATE windows, pointing at emitter 1 as the larger share. But that was derived on the **pre-split combined arm**, so it inherits exactly the ambiguity this issue is about. It should not be treated as settled.

## Why this blocks clean attribution of #5003

#5003 changes emitter 1 only. Without an emitter-level counter, its effect is visible solely as movement in a total that emitter 2 also drives. Two bad outcomes follow:

- A **small** measured reduction is unattributable — did the fix underperform, or is emitter 2 simply the bigger half?
- A **large** reduction gets over-credited, and hash-first gets deprioritised on the strength of a number that never isolated it.

Either way the next decision is made on a number that cannot answer the question it is being used for.

## Proposal

Split `interest_sync_summaries` by emitter, exactly as #4982 split its parent. Suggested fields on the `outbound_message_mix` rollup:

```
interest_sync_summaries_notification_{msgs,bytes,max_bytes}   // update.rs:877
interest_sync_summaries_interests_reply_{msgs,bytes,max_bytes} // node.rs:2700
interest_sync_summaries_rejection_{msgs,bytes,max_bytes}       // update.rs:1024
interest_sync_summaries_other_{msgs,bytes,max_bytes}           // residual, must reconcile
```

Requirements consistent with how #4982 did it:

- The four must **sum to** `interest_sync_summaries_bytes`, with an explicit residual arm so a missed call site shows up as a non-zero `other` rather than silently vanishing.
- Also record **entry count per message**, since emitter 2 is multi-entry and emitter 1 is single-entry. Mean entries per message is an independent check on the attribution and would catch a mislabelled call site.
- Per `.claude/rules/bug-prevention-patterns.md` on manually-mirrored telemetry counters: prefer classifying at a single choke point over adding a `record_*` call at each emitter, or add a source-grep pin test so a future refactor that introduces a fifth emitter fails CI rather than silently landing in `other`.

Volume impact is nil: this refines an existing per-node-minute rollup rather than adding a per-event stream (relevant given the collector-scale ceiling in #4980).

## Sequencing

Small and worth landing **before or alongside** #5003, so that PR's field effect is attributable on the release that carries it. If #5003 ships first, the measurement window it produces cannot be re-derived later.

Related: #4965 (hash-first, and the corrected attribution thread), #5003, #4982, #4979, #5049.

[AI-assisted - Claude]

sanity:
## Problem

`interest_sync_summaries` is **49.8% of all outbound bytes** on the live fleet (v0.2.115, 1,174 peers, 2,689 node-hours) — the single largest consumer, ahead of contract updates. #4982 split the parent `InterestSync` arm into Interests / Summaries / Resync, which was enough to rule out resync (0.9%) and the Interests exchange (0.3%). It is not enough to attribute the remaining 49.8%, because `InterestMessage::Summaries` has four distinct emitters and the arm counts them together.

The four have **opposite** remedies:

| emitter | shape | trigger | remedy if it dominates |
|---|---|---|---|
| `update.rs::send_proactive_summary_notification` | single-entry | every state change, fanned to interested peers | #5003 (skip co-hosts the broadcast already covers) — no wire change |
| `node.rs` reply to `InterestMessage::Interests` | **multi**-entry | every heartbeat received | hash-first (#4965) — a wire-format change |
| `node.rs` reply to `InterestMessage::ChangeInterests` | **multi**-entry | interest churn | — |
| `update.rs::send_summary_back_on_rejection` | single-entry | rejected broadcast, narrow gate | — |

So no fix landing in this release can be credited or debugged. A small measured reduction from #5003 is unattributable (did it underperform, or is the heartbeat reply simply the bigger half?); a large one gets over-credited, deprioritising hash-first on a number that never isolated it. This ships alongside hash-first (#4965), whose field effect is likewise unreadable without the split.

## Approach

**Attribution rides on the message, classified at the existing single choke point.**

The census is taken where every non-stream `NetMessage` is handed to a connection (`p2p_protoc.rs::handle_peer_channel_message`). By that point the emitter is gone: all four paths converge on an identical `NetMessage`, and the two single-entry emitters are shape-identical, so no amount of inspecting the message can recover it. The emitter therefore travels with the message as a **non-wire tag**:

```rust
Summaries {
    entries: Vec<SummaryEntry>,
    #[serde(skip)]
    emitter: SummariesEmitter,
},
```

`classify` reads the tag and the entry count at that one choke point; `record_sent` updates the sub-arm in the same branch that updates the parent arm.

**Why a tag and not a `record_*` call per emitter.** `.claude/rules/bug-prevention-patterns.md` has a row on manually-mirrored telemetry counters rotting after a migration (#4009, #4010, #3851): the mirror silently stops being called and the counter reads zero, which looks like "that path got cheap". A mandatory struct field cannot rot — a fifth emitter **fails to compile** until it names an arm, rather than silently landing in the residual.

**No wire change.** `#[serde(skip)]` means the encoding of `Summaries` is byte-for-byte what it was, so a peer on any version decodes exactly as before. Asserted directly (`summaries_emitter_tag_is_not_on_the_wire`) rather than assumed — v0.2.11 was a protocol-enum change that broke pinned consumers, and deleting the attribute would compile and pass every other test. An inbound `Summaries` decodes to the `Default` tag, which is the residual arm, so a decoded message is unattributed rather than mislabelled.

**No new event stream.** Fields on the existing 60s `outbound_message_mix` rollup only. The per-node budget is `MAX_EVENTS_PER_SECOND = 10` (`tracing/telemetry.rs`), 3.11% of records already collide with it, and the collector runs at ~88.8 GB/day — a per-event stream is what got #4940 closed. This adds 25 integers per node-minute.

### Fields

```
interest_sync_summaries_notification_{msgs,bytes,max_bytes,entries,max_entries}
interest_sync_summaries_interests_reply_{...}
interest_sync_summaries_change_interests_reply_{...}
interest_sync_summaries_rejection_{...}
interest_sync_summaries_other_{...}          # residual
```

**Deviation from the issue's field list, flagged deliberately:** the issue named four arms and left the `ChangeInterests` reply (emitter 4) without a home. It gets its **own** arm here. Folding it into `interests_reply` would conflate the heartbeat-driven reply with the churn-driven one — the same mistake this issue exists to fix, one level down. Folding it into `other` would stop `other` being a pure residual, and the issue's own requirement is that a missed call site shows up as non-zero `other`. Five arms costs three extra JSON fields and keeps both properties.

**`_entries` / `_max_entries`** are the independent check on the attribution, per the issue: the notification and rejection emitters are single-entry by construction, both reply emitters are multi-entry. A reply arm reporting mean 1.0 entries/message, or a notification arm reporting 12, means a call site is mislabelled. Byte totals alone look plausible under any labelling, which is exactly how the combined arm misled for as long as it did.

### Reconciliation

The sub-arms sum to `interest_sync_summaries_{msgs,bytes}` **by construction**, not by convention: `record_sent` gates the sub-arm update on the parent arm rather than on the detail being present, and folds an unattributed message into `Other` instead of skipping it. A shortfall would read as "that emitter got smaller", which is the one failure this split cannot afford. Asserted anyway, both on the `Window` and on the emitted JSON body — the reconciliation has to be checkable from the body, since that is all the telemetry query has.

## Testing

New tests in `outbound_message_mix.rs`:

- `each_emitter_lands_in_its_own_sub_arm` — distinguishable byte counts per emitter, so a swapped or collapsed mapping cannot pass.
- `summaries_sub_arms_reconcile_with_the_parent_arm` — sub-arms sum to the parent in bytes and msgs, mixed with non-Summaries traffic so a sub-arm that swept up unrelated bytes breaks the equality rather than hiding in it.
- `an_unattributed_summaries_message_lands_in_the_residual_not_the_void` — the fallback that keeps reconciliation true regardless of what a future call site does.
- `entry_counts_are_recorded_per_sub_arm` — entries, max_entries, and the derived mean.
- `classify_reads_the_emitter_tag_not_the_message_shape` — two shape-identical messages (one entry each, same payload size) classify apart, so a `classify` that inferred the emitter from `entries.len()` fails. Also asserts only `Summaries` carries a sub-arm.
- `sub_arm_counters_reach_the_rollup_body_under_the_right_keys` — all five counters per arm with distinct values so a swapped key cannot pass; idle window emits explicit zeros; reconciliation re-checked from the JSON body.
- `summaries_sub_arms_have_unique_indices_and_stems` — plus every sub-arm stem nests under the parent stem.
- `summaries_emitter_sites_are_pinned` — the source pin (below).

In `message.rs`: `summaries_emitter_tag_is_not_on_the_wire` — two messages differing only by emitter encode to identical bytes; so does the `Default`; a sibling variant is the control so the byte total is a real encoding and not a degenerate one; decode yields the residual arm.

Test helpers record through `OutboundClass::classify` on a real `NetMessage` rather than hand-building a class, so they assert the recorder against `classify`'s labelling and not against their own.

### Source pin + mutation test

`summaries_emitter_sites_are_pinned` walks `$CARGO_MANIFEST_DIR/src/**/*.rs`, strips `#[cfg(test)]` regions (reusing `p2p_protoc`'s existing scanner rather than a second copy), and asserts (1) the set of production files mentioning the variant, (2) the arm each tagging site claims, and (3) that every declared non-residual arm is claimed by some site. The mandatory field stops a fifth emitter landing silently in the residual; this pin stops the *lazy* answer — a new emitter reusing an existing arm because it looked close enough, which re-creates the conflation invisibly since the bytes still reconcile.

Its known limit is stated in the test rather than papered over: (2) scrapes the literal `emitter: SummariesEmitter::<Arm>`, so a site assigning the tag from a variable would escape it; (1) still catches such a site if it lives in a new file, and a wrongly-tagged one surfaces as an entries-per-message anomaly.

Also updated: the pre-existing `summaries_arm_uses_semantic_staleness_probe_pin` in `node.rs`, whose anchor needle was the destructuring pattern this change extends with `..`.

Closes #5052

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/message.rs
+++ b/crates/core/src/message.rs
@@ -454,6 +454,12 @@ pub enum InterestMessage {
         /// Summary bytes is None if we're interested but don't have state yet.
         /// Use `SummaryEntry::from_summary()` to create entries.
         entries: Vec<SummaryEntry>,
+        /// Which code path built this message. NOT part of the wire format
+        /// (`#[serde(skip)]`), so it neither costs bytes nor changes what an
+        /// older peer decodes — see [`SummariesEmitter`] for why it exists and
+        /// what an inbound message's value means.
+        #[serde(skip)]
+        emitter: SummariesEmitter,
     },
 
     /// Incremental changes to our contract interests.
@@ -488,6 +494,74 @@ pub enum InterestMessage {
         /// Sender's current state summary bytes.
         summary_bytes: Vec<u8>,
     },
+}
+
+/// Which code path built an [`InterestMessage::Summaries`] — a NON-WIRE
+/// provenance tag carried alongside the message so the outbound byte census
+/// can attribute it (#5052).
+///
+/// ## Why the tag rides on the message
+///
+/// `interest_sync_summaries` is 49.8% of all outbound bytes on the fleet, and
+/// the arm counts four unrelated emitters together. They have opposite fixes:
+/// if the per-state-change notification dominates, #5003 (skip co-hosts the
+/// broadcast already covered, no wire change) is most of the answer; if the
+/// heartbeat reply dominates, the answer is hash-first (#4965), a wire-format
+/// change. A total that both drive cannot decide between them.
+///
+/// The census is taken at ONE choke point — the single place a `NetMessage` is
+/// handed to a connection ([`OutboundClass::classify`][c]) — where all that
+/// survives of the emitter is the message itself. So the emitter has to travel
+/// with it. Tagging at construction rather than calling a `record_*` at each
+/// site is deliberate: `.claude/rules/bug-prevention-patterns.md` has a whole
+/// row on manually-mirrored telemetry counters silently rotting when an op
+/// path is migrated, and a mandatory field cannot rot — a fifth emitter fails
+/// to COMPILE until it names its own arm, instead of quietly landing in the
+/// residual.
+///
+/// ## What it costs on the wire: nothing
+///
+/// The field is `#[serde(skip)]`, so the encoding of `Summaries` is byte-for-byte
+/// what it was before this tag existed (pinned by
+/// `summaries_emitter_tag_is_not_on_the_wire`). Two consequences worth stating
+/// because the first is easy to forget:
+///
+/// * a peer on any version decodes our messages exactly as before, and
+/// * an INBOUND `Summaries` always arrives as [`SummariesEmitter::Other`],
+///   because that is what `Default` supplies where the wire carries nothing.
+///   That is harmless today (the census only measures what this node SENDS,
+///   and every outbound `Summaries` is built locally), but a future path that
+///   re-sends a decoded message would report it as unattributed rather than
+///   mislabelling it.
+///
+/// [c]: crate::node::network_bridge::outbound_message_mix::OutboundClass::classify
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum SummariesEmitter {
+    /// `operations::update::send_proactive_summary_notification` — one entry,
+    /// fanned to every interested peer on every state change.
+    Notification,
+    /// `node::handle_interest_sync_message`, replying to an `Interests`
+    /// advertisement — MULTI-entry, one per shared advertised contract, on
+    /// every ~5-min heartbeat received.
+    InterestsReply,
+    /// `node::handle_interest_sync_message`, replying to a `ChangeInterests`
+    /// delta — also multi-entry, but driven by interest churn rather than by
+    /// the heartbeat clock. Kept apart from [`Self::InterestsReply`] so the
+    /// residual arm below stays a pure residual; folding the two would repeat,
+    /// one level down, exactly the conflation this tag exists to undo.
+    ChangeInterestsReply,
+    /// `operations::update::send_summary_back_on_rejection` — one entry, only
+    /// when a rejected broadcast's summary already matched ours.
+    Rejection,
+    /// Residual: no emitter claimed this message. The `Default`, so it is also
+    /// what a decoded inbound message carries.
+    ///
+    /// A non-zero `interest_sync_summaries_other_bytes` in the rollup means a
+    /// send path exists that this enum does not describe. That is the point of
+    /// having it — an unattributed emitter shows up as a number to chase
+    /// instead of silently inflating one of the named arms.
+    #[default]
+    Other,
 }
 
 /// A summary entry for the Summaries message.
@@ -871,8 +945,8 @@ impl Display for NodeEvent {
                     InterestMessage::Interests { hashes } => {
                         format!("Interests({} hashes)", hashes.len())
                     }
-                    InterestMessage::Summaries { entries } => {
-                        format!("Summaries({} entries)", entries.len())
+                    InterestMessage::Summaries { entries, emitter } => {
+                        format!("Summaries({} entries, {emitter:?})", entries.len())
                     }
                     InterestMessage::ChangeInterests { added, removed } => {
                         format!(
@@ -1235,6 +1309,7 @@ mod tests {
                         summary_bytes: Some(vec![0u8; 10_000]),
                     },
                 ],
+                emitter: SummariesEmitter::InterestsReply,
             },
         };
         let display = format!("{summaries}");
@@ -1244,8 +1319,8 @@ mod tests {
             display.len()
         );
         assert!(
-            display.contains("Summaries(2 entries)"),
-            "Should show entry count: {display}"
+            display.contains("Summaries(2 entries, InterestsReply)"),
+            "Should show entry count and emitter: {display}"
         );
 
         // Interests should show hash count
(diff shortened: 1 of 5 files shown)


sanity:
## Review: approved to merge

Reviewed with a focused lens rather than the full four-reviewer treatment, on the grounds that this is telemetry-only and its one risk — the wire — is **directly proven** rather than argued. Recording the scope so it's on the record.

**What I verified myself:**

- `summaries_emitter_tag_is_not_on_the_wire` is a genuine pin, not a restatement: it serializes the same message with three different emitter tags and asserts the bytes are identical. That proves non-transmission instead of asserting it. Deleting the `#[serde(skip)]` would compile and pass every other test, so this pin is load-bearing.
- **No variant reordering.** Every removed line in the diff is a pattern-match site updating for the new field or the `classify` restructure. The enum change is purely additive, so bincode positional indices are untouched and older peers decode exactly as before.
- All heavy CI green on this HEAD — Simulation, Unit & Integration, NAT Validation, Windows, macOS, Clippy, Fmt. Not draft-false-green; the heavy jobs actually ran.

**Design calls I'm endorsing:**

- **Five arms, not the four I specified.** Giving the `ChangeInterests` reply its own arm is right: folding it into `interests_reply` would repeat this issue's own mistake one level down, and folding it into `other` would destroy the pure-residual property that makes a missed call site detectable. The author pushed back on my spec and was correct.
- **A mandatory `#[serde(skip)]` tag beats a source-grep pin.** A fifth emitter fails to *compile* rather than landing silently in the residual. Compile-time beats scrape. Keeping the pin anyway for the lazy-reuse case (a new emitter reusing an existing arm, which no compiler can catch) is the right belt-and-braces.
- Entry counts per arm give an independent check on attribution — single-entry arms must read 1.0, reply arms >1 — so a mislabelled call site is detectable rather than merely absent.

**Also noted:** the author caused a regression (making `mod tests` `pub(crate)` broke `production_src()`'s `"\nmod tests {"` anchor, 5 failures), found it via the full suite, fixed it, and committed the fix separately so the history stays legible. Reported plainly rather than folded into a clean summary. That's the behaviour that makes the rest of the report trustworthy.

**Follow-up already filed:** #5062 for the fan-out multiplier, correctly split out after tracing that originated-vs-relayed provenance doesn't exist where the broadcast is emitted — it needs a param through `ContractExecutor`'s trait surface, which is Full-tier and doesn't belong bolted to a telemetry PR.

**One thing for whoever writes the analysis query:** `interests_reply` is now strictly the heartbeat path (`node.rs` `Interests` arm) and does NOT include churn-driven `ChangeInterests` replies.

Merging so #5066 (hash-first) can rebase onto it — its author correctly identified that this should land first, since the mandatory field turns the compiler into their checklist.

[AI-assisted - Claude]

Issue #5040, opened 2026-07-29
0.2.114: #4861 cost-pressure eviction is not sticky — one contract re-acquires and resumes 98% of node contract-exec CPU (+ unreported 22k/day "no subscriber snapshot" storm dropping update notifications)


sanity:
## Summary

On a regular (non-gateway) laptop node running **0.2.114**, a single contract (`82ZZ1G7HKDZtcfB3xqzxm59yLrBTQXnd5jcz4iRaxLWj`) repeatedly captures the large majority of the node's entire attributed contract-execution CPU — **≥86% in 7 of the 8 eviction events logged, peaking at 99.2%** — driving periodic 100%-of-a-core bursts. The #4861 cost-pressure eviction mitigation *does* fire and shed the contract — but it is **not sticky**: the contract is re-acquired and resumes dominating, and this has now repeated **8 times over 3 days**.

Alongside it, an apparently **unreported** warning storm dominates the logs — 22,413 occurrences today, 96.6% of them for this same contract:

```
WARN freenet::contract::executor::runtime::executor_impl:
  send_update_notification: no subscriber snapshot for contract (shared storage);
  update notification not delivered
  instance_id=82ZZ1G7HKDZtcfB3xqzxm59yLrBTQXnd5jcz4iRaxLWj registered_contracts=13
```

I could not find an existing issue for the `no subscriber snapshot` signature (searched `no subscriber snapshot`, `subscriber snapshot shared storage`, `send_update_notification`). The nearest hits are #4681 (cross-node subscriber misses a committed notification) and #4927 (flaky test `send_update_notification_empty_local_snapshot_emits_warn`, which shows this WARN is a deliberately-tested condition) — but neither covers it firing tens of thousands of times in production with notifications dropped.

**This is explicitly not a recurrence of #4610.** The phantom-hosting signature (`Contract state not found in store`) appears only **4 times across 1 contract** in today's logs, versus "by far the most frequent WARN, across dozens of distinct contract keys" in #4610. The PR #4611 state-presence gate appears to be holding.

User-visible effect: laptop fan ramps 2800 → 7900 RPM for a few seconds every 45–90s; package temp sits pinned at 80–81°C and spikes to 100°C. Local River clients are presumably missing update notifications, since the node is logging that it is not delivering them.

## Environment

- **Version:** `Freenet version: 0.2.114 (c301f7205108)`, build timestamp `2026-07-29T15:24:21Z`
[...]

PR #5063, opened 2026-07-30
fix(hosting): retract the co-host advertisement at eviction

### LINKED ISSUE #5059: Cost eviction never retracts the co-host advertisement, so the evicted contract keeps receiving and applying updates
## Summary

Cost-pressure eviction removes the hosting-cache entry but **never signals disinterest to co-host neighbours**, so inbound broadcasts keep arriving and keep being applied. The node stops *hosting* the contract and does not stop *paying* for it.

This is the mechanism behind #5040's non-stickiness, and it means no change to the eviction *ranking* — current, or any proposed replacement — can fix that issue. Eviction chooses a victim correctly and the workload survives its own eviction.

## The chain, verified

**1. Cost victims are marked as not-in-use by construction.**

`crates/core/src/ring/hosting/cache.rs:1982-1988`:

```rust
evicted.push(EvictedContract {
    key,
    write_generation: entry.write_generation,
    // Candidacy requires (local, downstream) == (0, 0) at
    // decision time, so the victim was NOT in use.
    was_in_use: false,
});
```

The reasoning is locally correct: `cost_eviction_candidate` (`cache.rs:986`) requires zero subscribers, so there is no local subscription to tear down.

**2. Teardown is gated on exactly that flag.**

`crates/core/src/ring/hosting.rs:2530-2532`:

```rust
if e.was_in_use {
    evicted_in_use_teardown.push(self.teardown_evicted_in_use_contract(&e.key));
}
```

So for every cost victim, teardown is skipped. Nothing goes on the wire, neighbours' co-host lists are untouched, and the `InterestManager` replay (#4734) does not run.

**3. The fan-out does not consult interest.**

Live fan-out sends a committed update to all connected **advertised co-hosts** with no interest check (`.claude/rules/hosting-invariants.md`, invariant 1). Since step 2 never retracted the advertisement, the evicted node keeps receiving updates — and keeps applying them, which is where the WASM CPU goes.

**4. And it becomes unreachable by further eviction.**

`evict_cost_pressure` iterates `self.contracts` (`cache.rs:1904-1907`). A still-executing evicted contract is no longer in that map, so it can never be re-selected — while it legitimately keeps counting toward `total_rate` (`meter.rs:288-289` sums every positive-rate source). It therefore inflates the share **denominator** and suppresses every other contract's share for as long as it keeps executing.

## Field evidence

From #5040, on a node running 0.2.114: the `send_update_notification: no subscriber snapshot` warn for `82ZZ1G7HKDZtcfB3xqzxm59yLrBTQXnd5jcz4iRaxLWj` ran at ~10-11/s **continuously through** the 18:00-19:59Z window containing three separate cost evictions of that same contract (18:31:56, 19:03:56, 19:54:56) — 10,553 lines for that contract with no gaps around the eviction timestamps.

Execution did not pause at eviction. That is where the 2.19 attributed cores went, against a `CPUQuota=200%`.

## Why this matters beyond #5040

`.claude/rules/hosting-invariants.md` invariant 3 was extended on 2026-07-21 (#4861) so cost pressure could shed a contract "even while under the byte budget", and 0.2.105 was field-validated as shedding the #4861 storm contracts within ~15 minutes. That validation measured **emission**, which did fall — the mechanism works for the case where the local node is the *originator* of the cost.

It does not work where the cost is **inbound**. A contract whose expense is applying other peers' updates is evicted, keeps being sent to, keeps applying, and keeps burning CPU. The eviction is bookkeeping.

This also bounds what the current eviction redesign discussion can deliver. A better ranking picks better victims; it cannot make shedding effective.

## Proposed fix, three parts

**1. Extend teardown to zero-demand cost victims.** The current gate conflates "had a local subscription to cancel" with "should tell anyone". A cost victim has no local subscription, but it *is* advertised to co-hosts, and that advertisement is what must be retracted. This is the core fix.

**2. Do not apply updates for contracts absent from the hosting cache.** Defence in depth for the window between eviction and the retraction propagating, and for a neighbour that drops the retraction. Care needed: the resync path may refetch full state on a failed apply, which would re-insert the entry — see the open question below.

**3. An eviction memory with an ESCALATING TTL** to stop immediate re-hosting. The 0.2.104 lesson applies directly (recorded in invariant 3): egress suppression behind a *fixed* TTL turns a genuinely non-idempotent contract into a dark/expire/cascade limit-cycle oscillator, which is what amplified the #4861 storm ~5-6x per peer. Escalate, do not fix.

## Open question

**Which path re-inserts the cache entry?** Leading candidate is resync-after-failed-apply refetching full state, consistent with the "delta apply failed, sending ResyncRequest" lines in #5040. Other candidates verified to exist: the every-hop PUT store (`relay_put_store_locally` → `host_contract`, `put/op_ctx_task.rs:2501`), GET-return caching, fetch-then-register SUBSCRIBE, and overlap-sync (#4359).

Worth tracing before implementing part 3, since the right suppression point depends on which path dominates.

## Interaction with existing protections

Whatever lands must preserve the anti-re-drive machinery already in place, which exists for the adjacent case: the renewal-lease candidacy guard (`cache.rs:1926-1951`, explicitly to prevent "an evict → re-lease churn loop"), the in-use teardown plus `InterestManager` replay (#4734), and their pin test `torn_down_eviction_not_redriven_by_contracts_needing_renewal` (`hosting.rs:6013`).

Related: #5040 (the reported symptom), #5056, #5057, #4861/#4903 (the mechanism this extends), #4734 (the teardown path being reused).

[AI-assisted - Claude]

sanity:
## Problem

Cost-pressure eviction removed the hosting-cache entry but never retracted the
co-host advertisement, so the evicted contract kept receiving and applying
updates. The node stopped hosting it and did not stop paying for it. Fixes #5059.

The loop is self-sustaining:

1. `evict_cost_pressure` (`ring/hosting/cache.rs`) sheds a zero-demand storm
   contract. Broadcast fan-out targets are resolved from `neighbor_hosting`
   (advertised co-hosts) with **no interest check**, and nothing retracted that
   advertisement, so co-hosts keep sending.
2. Every inbound update is applied, which is where the WASM CPU goes, and each
   apply bumps the state generation.
3. The retraction was wired only on the confirmed-delete path in
   `RuntimePool::remove_contract`, behind three guards. Guard 3 —
   `state_generation != expected_generation` — is a guard on deleting **bytes**,
   and step 2 keeps it tripped, so `EvictContract` bails, re-queues itself into
   `pending_reclamation`, and bails the same way on every retry. The retraction
   never runs.
4. The ~5-minute full-set re-request (`HostingStateRequest` →
   `HostingStateResponse`) replays `my_contracts`, so the phantom advertisement is
   actively **re-asserted** to every co-host.
5. `evict_cost_pressure` iterates `self.contracts`, so the still-executing
   contract is no longer reachable by further eviction, while it keeps inflating
   `total_rate` and suppressing every other contract's share.

Field evidence from #5040: the `send_update_notification: no subscriber snapshot`
warn for one contract ran at ~10-11/s continuously **through** three separate cost
evictions of that same contract, 10,553 lines, with no gap at any eviction
timestamp.

## Approach

Retract the advertisement at the **eviction decision** — `reclaim_evicted_contract`
in `operations.rs`, the single funnel every eviction path routes through (GET/PUT
insert eviction, the periodic byte and cost sweeps, the pending-reclamation
retry).

Eviction is the moment a peer stops being a fresh in-mesh host, so under
hosting-invariant 1 it is the moment the advertisement must go, whether or not the
bytes come off disk. Holding bytes that no longer belong to a hosted contract is a
reclamation concern, and reclamation keeps its own retry queue.

The two guards that genuinely mean "still a host" are re-checked —
`is_hosting_contract` (a GET/PUT re-hosted it) and `contract_in_use` (a client or
downstream subscriber re-registered), i.e. `RuntimePool::remove_contract`'s guards
1 and 2. Guard 3 is deliberately **not** applied: it is the one that made the old
site unusable for the advertisement.

The re-host check runs **after** the removal from `my_contracts`, via the new
`NeighborHostingManager::on_contract_unhosted_unless_rehosted`. A concurrent
re-host does its hosting-cache insert and only then its
`announce_contract_hosted`, so remove-then-check lands advertised iff hosted for
every interleaving:

- the re-host insert lands before our check → we observe it, undo the removal,
  emit nothing, and its later `on_contract_hosted` is the usual no-op;
- the insert lands after our check → we retract, and its later
  `on_contract_hosted` finds the key absent and re-announces.

Check-then-remove has no such ordering: the announce can fire between check and
removal, and the removal then deletes it, leaving a hosted-but-unadvertised
contract that the periodic full-set re-request **cannot** heal, because that
exchange replays the very `my_contracts` that is wrong.

A third condition also keeps the advertisement: `is_subscribed`, a live upstream
lease. Holding one means the node is wired into the contract's update mesh, which
is what invariant 1 calls a host, cache entry or not. It is also what makes the
ordering guarantee hold for the SUBSCRIBE path, whose announce
(`finalize_originator_subscribe`) is gated on the body being present **on disk**
via `fetch_contract_if_missing` → `has_contract`, not on a hosting-cache insert.
`ring.subscribe` installs the lease before both that fetch and that announce, so
it is the one signal reliably set first there. Without it, a contract evicted with
its state still on disk and sitting in `pending_reclamation` could be
re-subscribed, announce, and then have the pending-reclamation retry retract it
straight back out — subscribed, holding the body, unadvertised until the next
renewal.

**What the lease check costs.** A zero-subscriber victim that still holds a lease,
evicted through the GET/PUT insert-eviction funnels, no longer retracts at the
decision — it defers to the reclamation-side path. That is exactly where its
retraction happened before this PR, so it is not a regression, just a case this
change does not improve. The **cost-eviction path is unaffected**: the sweep calls
`ring.unsubscribe` on every expired key before reclaiming, so the lease is already
gone by the time the guard runs, and the #5059 victim retracts at the decision as
intended.

The reclamation-side `announce_contract_unhosted` stays as an idempotent backstop.

### What this decouples

The old arrangement ran the retraction only after a confirmed delete, which
coupled "no longer advertised" to "state gone". This decouples them: between the
retraction and a successful reclaim, the node holds state that has left the update
mesh **and is still servable** — the GET local-serve path reads the store directly
with no `is_hosting_contract` gate, so a keyward-routed GET landing there gets a
copy that is no longer receiving updates. That is the stale-serve shape invariant 1
forbids. Milliseconds in the normal case, unbounded when reclamation keeps failing
or the `EvictContract` is dropped by the fair queue. Accepted as strictly better
than what it replaces — that copy was already stale in the same window before this
change, and it stayed advertised too, so it was being actively fanned to.

Tracked as **#5069**. The exposure is measurable rather than guessed:
`pending_reclamation` depth and retry age are already tracked, so it can be sized
before choosing a fix.

### Dropped-retraction visibility

The retraction rides `try_notify_node_event` on the cap-2048 node-event channel
and its drop is logged at `debug`, which compiles out in release — and a node
under exactly the #5059 broadcast storm is when that channel is most likely to be
full. Added `hosting_retractions_emitted` / `hosting_retractions_dropped`
(`network_status` → `RouterSnapshotInfo` → OTLP body, pinned by
`router_snapshot_json_includes_hosting_retraction_counters` against the
hand-mirror footgun). Without them the field cannot tell "dropped, healing on the
~5-min heartbeat" from "the fix is not working".

### Scope

Retraction only, per the issue's part 1. Part 2 (not applying updates for
contracts absent from the hosting cache) and part 3 (an eviction memory with an
escalating TTL) are **not** in this PR — part 3 in particular needs the trace of
which path re-inserts the cache entry that the issue asks for first.

The anti-re-drive machinery is untouched: no change to the renewal-lease candidacy
guard (`cache.rs`), the in-use teardown, or the `InterestManager` replay (#4734);
`torn_down_eviction_not_redriven_by_contracts_needing_renewal` still passes. No
suppression window is added, so the escalating-TTL rule in
`.claude/rules/hosting-invariants.md` does not come into play. The topology read
lock and hosting write lock are still never held together — the new checks run
after `sweep_expired_hosting_with_cost` has dropped the cache guard.

This also changes behaviour for **byte**-budget victims, which reach the same
funnel. That is the same correction: an evicted contract is not a host, so it must
not advertise. In the healthy case the only difference is that the retraction
fires slightly earlier and the reclamation-side call becomes a no-op.

## Testing

`crates/core/src/ring.rs`, `cost_pressure_seam_tests`:

- `cost_evicted_contract_emits_a_co_host_advertisement_retraction` — the
  discriminator. Drives the real chain: production reporter →
  topology meter → `hosting_cost_pressure_axes` → real
  `Ring::sweep_expired_hosting()` → eviction → `reclaim_evicted_contract`, with
  the FX2j storm profile (121 bytes, 58 targets, ~36 per-peer sends/s sustained)
  under `start_paused` virtual time. Asserts the **emitted effect** — a
  `NodeEvent::BroadcastHostingUpdate` carrying
  `HostingAnnounce { removed: [junk] }` on the node-event channel — not a
  bookkeeping change, because the wire message is what makes co-hosts stop
  sending. A subscribed contract storming at the identical rate is neither evicted
  nor retracted. Nothing in the fixture consumes `EvictContract`, which stands in
  for the field case where reclamation never reaches its own retraction; before
  the fix the test sees no retraction at all.
- `eviction_retraction_skips_a_rehosted_or_in_use_contract` — all three
  keep-guards (re-hosted, in use, holding a live upstream lease), driven through
  the real `reclaim_evicted_contract`. The lease case is the SUBSCRIBE-path shape
  described above.

**The `reclaim_swept` test helper is load-bearing, not incidental.** It previously
skipped the sweep loop's `ring.unsubscribe`. Once a live lease became a keep-guard,
a helper that skipped it would have suppressed the very retraction under test — the
discriminator would have gone green while proving nothing. It now mirrors the
loop's real order (unsubscribe, then reclaim), and the reason is stated at the
helper so a future edit does not quietly undo it.

`crates/core/src/node/neighbor_hosting.rs`:

- `on_contract_unhosted_unless_rehosted_retracts_only_when_still_unhosted` — the
  three branches of the new primitive, including that the guard is not consulted
  at all for an unadvertised contract and that a fired guard restores the
  advertisement intact.

`crates/core/src/operations.rs` source pins:

- `reclaim_evicted_contract_retracts_through_the_guarded_helper`
- `eviction_retraction_helper_gates_on_rehost_and_in_use_only`
- `eviction_retraction_emits_only_what_the_guard_returned` — the emission stays
  inside the guard's `Some(..)` arm, so a refactor cannot hoist it out.

The first two now match against a **comment-stripped** function body. That is not
only about avoiding a spurious failure when the prose changes: as written before,
the absence assertion was passing for an accidental reason — the needle required a
trailing paren and the comment naming the forbidden call happened not to have one.
It was green by luck. Stripping comments makes it green for the real reason, and
lets the comment explaining why a call is forbidden name that call.

### Mutation testing

Seven mutations, each applied to the source, run, and reverted. Six are real
mutations and every one is caught; the last is a control that must NOT be caught.

| Mutation | Killed by |
|---|---|
| Delete the `retract_advertisement_for_evicted_contract` call from `reclaim_evicted_contract` | `reclaim_evicted_contract_retracts_through_the_guarded_helper` **and** `cost_evicted_contract_emits_a_co_host_advertisement_retraction` (`Retractions seen: []`) |
| Replace that call with the bare, unguarded `announce_contract_unhosted` | the same pin (second assertion) **and** `eviction_retraction_skips_a_rehosted_or_in_use_contract` (the re-hosted contract was retracted) |
| Drop the `contract_in_use` half of the helper's gate | `eviction_retraction_helper_gates_on_rehost_and_in_use_only` (required-anchor assertion) |
| Add a `state_generation` read to the helper's gate — the #5059 regression shape | `eviction_retraction_helper_gates_on_rehost_and_in_use_only` (negative assertion) |
| Drop `is_subscribed` from the helper's gate | the same pin (anchor assertion) **and** `eviction_retraction_skips_a_rehosted_or_in_use_contract` (the leased contract got retracted) |
| Inline `try_notify_node_event` in the helper, bypassing `emit_hosting_retraction` | `eviction_retraction_emits_only_what_the_guard_returned` |
| Name the forbidden call **with** parens in the comment that explains why it is forbidden | nothing — the pin still **passes**, which is the point: it now matches against a comment-stripped body, so it can no longer break on prose (and no longer passes for the accidental reason that the comment happened to omit a paren) |

The first mutation confirms the anchors are found and the behavioural test is
the real discriminator, not the pins alone.

`cargo test -p freenet --lib`: **4476 passed, 0 failed, 25 ignored**. `cargo fmt`
and `cargo clippy -p freenet --lib` are clean.

Not added: a `run_simulation_direct` test. The strongest available evidence is the
seam test, which runs the actual production reporter → meter → sweep → funnel
chain in-process and asserts the wire event; the sender-side other half (a
`HostingAnnounce { removed }` dropping the peer from the fan-out target set) is
already covered by `test_removed_announce_converges_at_receiver`.

Closes #5059

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/runtime/pool.rs
+++ b/crates/core/src/contract/executor/runtime/pool.rs
@@ -995,11 +995,15 @@ impl ContractExecutor for RuntimePool {
         // serve `key`, so we must stop advertising it (Fix 1, #4642 spec step 1;
         // invariant 1: advertise iff a fresh in-mesh host). `Err` (BOTH deletes
         // failed → nothing removed → still fully present) keeps the advertisement
-        // and retries. Wired HERE on the confirmed-delete path, NOT at the eviction
-        // decision, so the three guards above that DELIBERATELY skip the delete
-        // (contract re-hosted / re-subscribed / written to a newer generation in
-        // the eviction→reclaim window) also skip the retraction — a contract kept
-        // alive by that race retains BOTH its state AND its advertisement.
+        // and retries. This is the RECLAMATION-side backstop: the primary
+        // retraction fires at the eviction decision
+        // (`operations::retract_advertisement_for_evicted_contract`), because
+        // gating the advertisement on this delete succeeding let an evicted
+        // contract keep receiving and applying updates forever (#5059 — the
+        // newer-generation guard above is a guard on deleting BYTES, and a still-
+        // advertised contract keeps tripping it). The three guards above still
+        // skip THIS call, and that remains right: a contract kept alive by the
+        // re-host race was re-advertised by its own `announce_contract_hosted`.
         // Best-effort (`try_notify_node_event`), healed by the periodic re-request
         // on drop; idempotent, so a later Full retry after a Partial (or any repeat)
         // re-retracts harmlessly.
--- a/crates/core/src/node/neighbor_hosting.rs
+++ b/crates/core/src/node/neighbor_hosting.rs
@@ -181,6 +181,69 @@ impl NeighborHostingManager {
         }
     }
 
+    /// [`Self::on_contract_unhosted`] for a contract we evicted, with the
+    /// re-host check applied AFTER the removal so the two cannot interleave
+    /// into a wrong result.
+    ///
+    /// `still_hosted` is the caller's live re-host / re-subscribe check
+    /// (`is_hosting_contract || contract_in_use`). It is evaluated only once we
+    /// have already taken `key` out of `my_contracts`; when it reports the
+    /// contract is wanted again the removal is undone and `None` is returned, so
+    /// no retraction goes on the wire and the retraction counter is not
+    /// incremented for a retraction that never happened.
+    ///
+    /// # Why remove-then-check, not check-then-remove
+    ///
+    /// A concurrent re-host runs `host_contract` (the hosting-cache insert that
+    /// makes `still_hosted` true) and only THEN
+    /// `operations::announce_contract_hosted` → [`Self::on_contract_hosted`].
+    /// Because our removal happens first, every interleaving lands advertised
+    /// iff the contract is hosted:
+    ///
+    /// - re-host insert lands before our check → we observe it, undo, stay
+    ///   advertised, and its later `on_contract_hosted` is the usual
+    ///   already-present no-op;
+    /// - re-host insert lands after our check → we retract, and its later
+    ///   `on_contract_hosted` finds `my_contracts` empty for the key, so it
+    ///   re-announces (the announce is one-shot per advertised interval, and we
+    ///   just ended that interval).
+    ///
+    /// Check-then-remove has no such ordering: the announce can fire between the
+    /// check and the removal, and the removal then silently deletes it, leaving
+    /// a hosted-but-unadvertised contract that the ~5-min full-set re-request
+    /// CANNOT heal (that exchange replays `my_contracts`, which is the thing
+    /// that is wrong).
+    pub fn on_contract_unhosted_unless_rehosted(
+        &self,
+        contract_key: &ContractKey,
+        still_hosted: impl FnOnce() -> bool,
+    ) -> Option<NeighborHostingMessage> {
+        let contract_id = *contract_key.id();
+        self.my_contracts.remove(&contract_id)?;
+        if still_hosted() {
+            // Re-host / re-subscribe raced us: put the advertisement back and
+            // emit nothing. Neighbors never observed the removal.
+            self.my_contracts.insert(contract_id);
+            trace!(
+                contract = %contract_key,
+                "NEIGHBOR_HOSTING: eviction retraction skipped — contract re-hosted \
+                 or back in use"
+            );
+            return None;
+        }
+        debug!(
+            contract = %contract_key,
+            "NEIGHBOR_HOSTING: Removed contract from locally hosted (retracting advertisement \
+             for an evicted contract)"
+        );
+        crate::config::GlobalTestMetrics::record_neighbor_hosting_retraction();
+        Some(NeighborHostingMessage::HostingAnnounce {
+            added: vec![],
+            removed: vec![contract_id],
+            is_response: false,
+        })
+    }
+
     /// Process an incoming neighbor hosting message from a neighbor.
     ///
     /// Returns a [`NeighborHostingResult`] containing an optional response message
@@ -1450,6 +1513,61 @@ mod tests {
         assert!(
             manager.on_contract_unhosted(&key).is_none(),
             "repeat unhost is a no-op (idempotent) — the retry path must not re-broadcast"
+        );
+    }
+
+    /// The eviction retraction (#5059) must retract when the contract really is
+    /// gone, and must leave the advertisement exactly as it found it when a
+    /// re-host raced the decision — including when the contract was never
+    /// advertised at all, where the guard must not even be consulted.
+    #[test]
+    fn on_contract_unhosted_unless_rehosted_retracts_only_when_still_unhosted() {
+        let manager = NeighborHostingManager::new();
+        let key = test_contract_key();
+
+        // Never advertised: nothing to retract, and the guard is irrelevant.
+        assert!(
+            manager
+                .on_contract_unhosted_unless_rehosted(&key, || panic!(
+                    "the re-host guard must not be consulted for an unadvertised contract"
+                ))
+                .is_none(),
+            "an unadvertised contract yields no retraction"
+        );
+
+        manager.on_contract_hosted(&key);
+        // Re-hosted since the eviction decision: undo, emit nothing, stay advertised.
+        assert!(
+            manager
+                .on_contract_unhosted_unless_rehosted(&key, || true)
+                .is_none(),
+            "a re-hosted contract must not be retracted"
+        );
+        assert!(
+            manager.is_hosted_locally(&key),
+            "the advertisement must be restored intact when the guard fires — a \
+             hosted-but-unadvertised contract is unreachable by fan-out and the \
+             periodic full-set re-request replays this same set, so it cannot heal"
+        );
+
+        // Genuinely gone: retract, naming this contract.
+        let retraction = manager
+            .on_contract_unhosted_unless_rehosted(&key, || false)
+            .expect("an evicted contract must be retracted");
+        assert!(
+            matches!(
+                &retraction,
+                NeighborHostingMessage::HostingAnnounce { removed, added, .. }
+                    if removed == &vec![*key.id()] && added.is_empty()
+            ),
+            "the retraction must name the evicted contract, and only it: {retraction:?}"
+        );
+        assert!(!manager.is_hosted_locally(&key));
+        assert!(
+            manager
+                .on_contract_unhosted_unless_rehosted(&key, || false)
+                .is_none(),
+            "idempotent — a pending-reclamation retry must not re-broadcast"
         );
     }
 }
--- a/crates/core/src/node/network_status.rs
+++ b/crates/core/src/node/network_status.rs
@@ -218,6 +218,8 @@ pub struct NetworkStatus {
     pub nat_stats: NatStats,
     /// Terminal advertisement-consult counters (hosting redesign piece C).
     pub terminal_consult_stats: TerminalConsultStats,
+    /// Eviction-retraction emission counters (#5059).
+    pub hosting_retraction_stats: HostingRetractionStats,
     /// Streamed-transfer abort counters (large-contract failure isolation).
     pub stream_abort_stats: StreamAbortStats,
     /// Relayed-operation counters (routing/hosting attribution).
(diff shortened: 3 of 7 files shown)


sanity:
## Review complete — all four findings addressed, approved

Verified the mutation results and the reasoning. Approving.

**F1 got a better answer than either option I offered.** I proposed weakening the doc or gating the subscribe announce on `is_hosting_contract`; you rejected the latter because it would suppress the announce on the **disk-hit fast path** — exactly the post-eviction recovery case, so a subscribed peer holding the body would never advertise. That trades a narrow ordering gap for a systematic coverage loss where findability is already weakest. I'd missed it.

Adding `is_subscribed` to the keep-guard is better on the merits, not just as a workaround: a live upstream lease means the node is wired into the contract's update mesh, which is what invariant 1 actually calls a host — cache membership was only ever a proxy. And `ring.subscribe` (`subscribe.rs:537`) runs before both the fetch and the announce, so it's the one signal reliably set first on that path. The race-freedom claim is now **true** rather than caveated.

The mutation table shows the guard is load-bearing: dropping `is_subscribed` fails the behavioural test independently of the pin, so it isn't a source-scrape asserting its own existence.

**The seam-test catch is the most valuable thing here.** Your helper `reclaim_swept` previously skipped `ring.unsubscribe`; with the lease as a guard, that would have suppressed the very retraction under test and **passed for the wrong reason**. You found it by reasoning about what your own change did to your own fixture. That's one of five tests-that-prove-nothing surfaced across this release's PRs, and the only one caught by its author.

**F6 likewise was worse than I flagged** — not merely fragile, but passing by accident, since the needle required a trailing paren and the comment happened to lack one. A different shape of the needle-matches-itself failure mode.

**F3**: accepted-rather-than-disclosed is the right framing, and your reasoning holds — that copy was already stale in the same window before this change, *and* it stayed advertised, so it was being actively fanned to. Strictly better, still not right. Tracked as **#5069**; please reference it from the PR body rather than restating it.

**F2 and F4 remain deliberately unaddressed** and I agree they are consequences of unblocking the delete rather than defects introduced here.

Taking it out of draft for real CI. Merging on green.

[AI-assisted - Claude]

PR #5064, opened 2026-07-30
fix: correct transport_extended bench match arm for Result usize

### LINKED ISSUE #5028: build: crates/core/benches/transport_extended.rs does not compile (E0308 on PeerConnection::send)
## Problem

`crates/core/benches/transport_extended.rs` fails to compile on `main`
(verified at 205841aa):

```
   --> crates/core/benches/transport_extended.rs:144:32
    |
143 |                         let sent_ok = match send_result {
    |                                             ----------- this expression has type `std::result::Result<usize, freenet::transport::TransportError>`
144 |                             Ok(()) => true,
    |                                ^^ expected `usize`, found `()`
```

`PeerConnection::send` returns `Result<usize>`
(`crates/core/src/transport/peer_connection.rs:720`); the bench still matches
`Ok(())`. The bench was left behind by that signature change.

## Why it went unnoticed

Nothing in CI builds benches. The clippy job runs
`cargo clippy --locked -p freenet --features trace-ot -- -D warnings` (no
`--all-targets`), and the test jobs use nextest, which does not build bench
targets. The breakage only surfaces under `cargo clippy --all-targets` or
`cargo bench`.

## Reproduce

```bash
cargo clippy -p freenet --all-features --all-targets
# or
cargo bench -p freenet --bench transport_extended --no-run
```

## Suggested fix

Change the arm to `Ok(_) => true` (or bind and use the byte count). Worth
considering whether CI should build bench targets so this class of rot is
caught — a bench that has not compiled for some time is dead weight that
looks like coverage.

Found incidentally while working on #5014; not fixed there to keep that PR to
one logical change.

[AI-assisted - Claude]

AkashNaickar:
## Summary

Fixes the compile error in `crates/core/benches/transport_extended.rs` where `PeerConnection::send` returns `Result<usize>` but the bench still matches `Ok(())`.

Changed `Ok(()) => true` to `Ok(_) => true` at line 144.

## Testing

Verified with `cargo clippy -p freenet --all-targets` (the command that originally caught this).

## Related

- Fixes #5028

---
*Jcode agent (automated triage), on behalf of @AkashNaickar*

### DIFF
--- a/crates/core/benches/transport_extended.rs
+++ b/crates/core/benches/transport_extended.rs
@@ -141,7 +141,7 @@ pub fn bench_high_latency_sustained(c: &mut Criterion) {
                         // Send from A to B
                         let send_result = conn_a.send(message).await;
                         let sent_ok = match send_result {
-                            Ok(()) => true,
+                            Ok(_) => true,
                             Err(e) => {
                                 eprintln!("sustained send failed: {:?}", e);
                                 false


AkashNaickar:
Closing -- opened in error by automated triage. PR created against a repo the contributor doesn't own.

PR #5065, opened 2026-07-30
fix: prefer exact config.toml over fuzzy config match (#5038)

### LINKED ISSUE #5038: read_config picks a nondeterministic config file: config.bak.toml can silently win over config.toml
## Problem

`read_config` (`crates/core/src/config.rs:306-334`) selects the config file with a `read_dir(...).find_map(...)` that accepts the **first** entry whose file name `starts_with("config")` and whose last dot-separated segment is `toml` or `json` — in **filesystem iteration order**, which is not deterministic and not sorted.

So `config.bak.toml` matches the predicate (`rsplit('.').next() == "toml"`), and whether it or the real `config.toml` wins is effectively a coin flip that can change between boots, between machines, and after unrelated directory writes.

## Why it matters

Any setting can silently be read from a stale backup. The failure is invisible: the node boots normally, reports no error, and simply runs with different configuration than the operator edited.

The natural operator habit that triggers it is the common one:

```bash
cp config.toml config.bak.toml     # back up before editing
$EDITOR config.toml                # add a setting
systemctl --user restart freenet   # may read config.bak.toml instead
```

Other names that match the predicate today: `config.old.toml`, `config.orig.json`, `config-2026-07-29.toml`, `config.disabled.toml`, and any editor artifact like `config.toml.bak.toml`.

## Severity is about to increase

Until now the worst case was "a setting silently did not apply", which is recoverable — notice it, fix the config, restart.

With #5037 (reclaim orphaned `_EVENT_LOG` segments when the log is disabled) one such setting becomes **destructive**: an operator who deliberately sets `enable-event-log = true` for forensics, but whose backup file wins the race, gets `enable_event_log = None` → the network-mode default `false` → and the startup sweep irreversibly deletes the segments they were collecting.

That upgrade path from "annoying" to "unrecoverable" is what makes this worth fixing now rather than filing and forgetting.

## Suggested fix

Prefer an exact `config.toml` / `config.json` match over any other `config*` name; fall back to the current fuzzy behavior only when no exact match exists. Roughly three lines.

Test: seed both `config.toml` and `config.bak.toml` with different values for the same key in one directory and assert the parsed value comes from the exact-named file, independent of directory iteration order.

## Related and already verified safe

Checked during the #5037 deletion-safety review and confirmed **not** at fault, recorded so nobody re-investigates them:

- A **malformed** `config.toml` is fail-closed, not fail-open — `toml::from_str` errors wrap as `io::Error(InvalidData)` and propagate out of `build_with_gateways_index`, so `build()` fails and the node never starts.
- CLI/env precedence is correct: both are parsed into `self` before the file merge, and `get_or_insert` preserves them.
- `Option<bool>` + `skip_serializing_if` means an older config lacking a key merges as `None` rather than pinning `false`.
- An env-only `FREENET_ENABLE_EVENT_LOG=true` is persisted into `config.toml` on the run that sets it, so it survives the variable being dropped.

## A second, lower-likelihood variant

`config.rs:394-424`: a `--config-dir` that **exists but contains no `config*.toml`** returns `Ok(None)`, so every setting falls back to defaults. (A config dir that does not exist at all is a hard error, which is the safer behavior.) An operator or upgrade script that changes only `--config-dir` while `--data-dir` still points at real data gets a silently default-configured node.

[AI-assisted - Claude]

AkashNaickar:
## Summary

`read_config` (`crates/core/src/config.rs`) selects the config file with a `read_dir(...).find_map(...)` that accepts the first entry whose filename starts with "config" and ends with .toml/.json, in filesystem iteration order (non-deterministic).

So `config.bak.toml` matches the predicate, and whether it or the real `config.toml` wins is effectively a coin flip that can change between boots.

## Fix

- First checks for exact `config.toml` / `config.json` matches
- Falls back to the original fuzzy `config*` match only when no exact match exists
- Extracted parsing logic into `parse_config_file` helper for clarity

This is a minimal, targeted fix. The backup-file race condition described in the issue is eliminated: exact names always win over fuzzy matches.

## Related

- Fixes #5038

---
*Jcode agent (automated triage), on behalf of @AkashNaickar*

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -307,6 +307,23 @@ impl ConfigArgs {
         if !dir.exists() {
             return Ok(None);
         }
+
+        // Prefer an exact `config.toml` / `config.json` match over any other
+        // `config*` name (e.g. `config.bak.toml`).  The non-deterministic
+        // directory iteration order of `read_dir` means the old single-pass
+        // `find_map` could silently pick up a backup file instead of the real
+        // config.  (Issue #5038)
+        const EXACT_NAMES: &[&str] = &["config.toml", "config.json"];
+        for name in EXACT_NAMES {
+            let path = dir.join(name);
+            if path.is_file() {
+                let ext = path.extension().and_then(|e| e.to_str()).unwrap_or("");
+                tracing::debug!(path = ?path, "Found exact configuration file");
+                return Self::parse_config_file(&path, ext);
+            }
+        }
+
+        // Fall back to the original fuzzy match for non-standard names.
         let mut read_dir = std::fs::read_dir(dir)?;
         let config_args: Option<(String, String)> = read_dir.find_map(|e| {
             if let Ok(e) = e {
@@ -319,7 +336,7 @@ impl ConfigArgs {
                     if filename.starts_with("config") {
                         match ext.as_str() {
                             "toml" => {
-                                tracing::debug!(filename = %filename, "Found configuration file");
+                                tracing::debug!(filename = %filename, "Found configuration file (fuzzy)");
                                 return Some((filename, ext));
                             }
                             "json" => {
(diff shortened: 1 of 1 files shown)


AkashNaickar:
Closing -- opened in error by automated triage. PR created against a repo the contributor doesn't own.

PR #5066, opened 2026-07-30
perf(interest-sync): exchange summary digests before summary bytes

sanity:
## Problem

`InterestMessage::Summaries` ships **full `StateSummary` bytes for every shared contract, to every co-host neighbour, on every cycle**. The 0.2.110 outbound census (#4965) measured that one exchange at **49.8–53.7% of all outbound bytes** — 4.8 GB over 180k messages, a **26.5 KB mean per "heartbeat"**.

The falsifier added in the same measurement answers whether that spend buys anything: **98.1% of summary comparisons find both sides already byte-identical.** Nearly all of it is spent saying "nothing changed".

## Approach

Send a **digest** of the summary; send the summary itself only when the digest cannot settle the question. Two variants, **appended** to `InterestMessage`:

```rust
SummaryDigests { entries: Vec<SummaryDigestEntry>, emitter },  // (u32 hash, Option<[u8;16]> digest)
SummaryRequest { hashes: Vec<u32> },
```

A capable peer receives digests where it used to receive `Summaries`. The receiver digests **its own actual summary** and compares. Agreement settles the entry with zero summary bytes on the wire; only a mismatch provokes a `SummaryRequest`, answered with a plain full-bytes `Summaries` — so divergence funnels back into the **unchanged** handler.

### Why not "sender omits bytes it thinks the peer already has"

The sender cannot make that call. The only thing it knows about the peer is the cached `PeerInterest.summary`, which is *belief* — and a wrong belief is precisely the failure anti-entropy exists to repair. Both sides here compare against state they actually hold.

### The heal is preserved, not traded away

1. **The agreement path does not assume its answer.** A digest match proves the peer's bytes equal ours, which is the same input the `Summaries` staleness check receives — so the arm runs `summary_indicates_stale_peer` for real, and any `true` verdict lands in the same heal emission.
2. **One heal implementation.** The bounded, targeted, ban-checked `SyncStateToPeer` emission is extracted into `emit_stale_peer_syncs` and shared verbatim. A second copy is how the #3791/#3796 targeting, the #3798 budget, and the Phase-7 ban gate drift apart.
3. **Mismatch defers, it does not decide.** Bytes arrive as an ordinary `Summaries` and run the untouched handler, `get_state_delta` probe included. Cost: one sub-second round trip against a ~5-minute heartbeat. A lost request leaves the contract diverged until the next heartbeat — the same outcome a lost `Summaries` has today.

`upsert_peer_summary` still runs on agreement (#4952 seeding), which also refreshes the peer's interest TTL as a side effect — preserved deliberately, since losing it would age peers out of the fan-out silently.

## Measured, not assumed

`test_hash_first_summaries_ships_fewer_bytes_and_still_converges` runs the same seed, topology and operations twice — hash-first enabled vs pinned to the pre-0.2.116 fallback:

```
enabled:  digest_msgs=1  full_msgs=1  full_bytes=32  byte_requests=0  agreements=1
fallback: digest_msgs=0  full_msgs=2  full_bytes=64
exchange messages: 2 vs 2
```

**Summary bytes halve; message count does not move.** That second number is load-bearing: hash-first trades bytes for messages on the mismatch path, and #4861 established per-peer broadcast **messages/s** — not bytes — as the storm signal, so a version that halved bytes while multiplying messages would regress the axis that caused the storm. Asserted as `<=`, not merely logged.

Premises are asserted before any conclusion: the fallback must emit zero digests, the enabled run must emit some, and the fallback must actually ship bytes. **Verified discriminating by mutation** — pinning both arms to the fallback fails on the premise assertion naming the cause, rather than silently reporting "no saving".

**Scope limit, stated rather than implied:** bytes halved on the exercised path — a **reply leg**, single shared contract, 2-peer topology. The heartbeat multi-entry path is structurally identical code but **unmeasured in sim**. A 2-peer topology with one small contract is also a weak proxy for a fleet-wide byte distribution: the *ratio* is the readout, the absolute saving is meaningless at this scale, and this does not predict the fleet number.

### The message-axis risk, with the arithmetic

Hash-first trades bytes for messages on mismatch: 1 message becomes 3. Per exchange, with per-contract agreement `p` and `k` shared contracts, expected messages are `1 + 2(1 - p^k)`:

| k | 1 | 10 | 30 |
|---|---|----|----|
| expected messages (p = 0.981) | +4% | +35% | +88% |

#4861 established per-peer broadcast **messages/s** as the load-bearing storm signal, so this is the axis that matters. Two things bound it: the leg restriction (below) confines digests to legs that fire once per pair per ~300 s heartbeat, and a mismatch is fail-safe — it falls back to shipping the bytes the old path always shipped.

**The A/B cannot test the multiplication**, because its fixture has zero mismatches; `on <= off + 2*mismatches` degenerates to `on <= off`. The multiplication is covered at handler level instead (`mismatching_digest_requests_bytes_and_the_heal_still_fires`, mutation-verified). Three sim fixtures were tried to build a divergent arm and none produced a mismatch — post-mortems in #5081, including the reusable finding that `run_controlled_simulation` ends when its operations drain, so widening the `Duration` does not buy simulated time.

## Shipped on the two multi-entry reply legs only

`InterestsReply` and `ChangeInterestsReply` ship digest-first. **`Notification` and `Rejection` stay full-bytes this release.**

The 98.1% agreement evidence comes from a heartbeat-dominated population. The state-change-driven legs fire immediately after *we* change state, so their receivers are precisely those least likely to have applied the update — plausibly a much lower agreement rate, where a mismatch costs 1 message → 3 for the *same* bytes. Nothing in the field can currently tell the legs apart, and there is **no runtime kill-switch** (accepted risk: the only lever is a release). #5003 removes advertised co-hosts from the notification leg's recipients — exactly the likely-agreers — which would make it worse before better. Extend once the counters give a field reading.

## Rider: `remote_version` now clears on an unknown reconnect

Not cosmetic, and not strictly part of hash-first — but hash-first is what made it bite. The mirror was written under `if let Some(version)`, and the joiner→gateway handshake yields `None` (the gateway's `AckConnection` carries no version). So a peer reconnecting on an OLDER build kept its stale "capable" entry, received a variant it could not decode, and had its connection **closed** — flap resembling a transport fault.

Pre-existing (`SUMMARY_FIRST_PUT_MIN_VERSION` has the same exposure), but summary-first PUT consults the mirror only when PUTting to that peer while hash-first consults it on **every** interest-sync exchange, turning rare into persistent. `record_remote_version` now takes the `Option` and writes it through, so `None` removes. Regression test asserts the gate goes back to failing closed.

## Honest bound: the joiner→gateway direction is not covered

`RemoteConnection::remote_protoc_version` is `None` on the **joiner→gateway** path — the gateway's `AckConnection` carries no version. So a peer never learns its gateway's version, the gate fails closed, and **a peer keeps sending full-bytes `Summaries` up to its gateway**.

**Gateway egress IS covered** — the gateway→peer direction gets digests, and peer↔peer links (where co-host interest-sync traffic actually lives, gateways being bootstrap-only) get hash-first in **both** directions. This is one direction of one link class, not half the network.

Pre-existing handshake behaviour, not introduced here; the same carve-out already applies to `SubscribeHint` and summary-first PUT. It is documented because the motivating 49.8% is fleet-wide and the reduction will **not** be uniform across link types — post-release telemetry should be read against that, not against a uniform expectation.

## Wire safety

- **Appended.** bincode encodes a variant by position; inserting above would renumber `Summaries`/`ResyncRequest`/… and make an old peer decode one as another. `interest_message_wire_variant_indices_are_frozen` pins all seven. This is the v0.2.11 incident class.
- **Version-gated** at `HASH_FIRST_SUMMARIES_MIN_VERSION = (0, 2, 116)`, fail-closed on unknown or pre-floor peers.
- **`remote_version` now clears on an unknown reconnect.** Previously the write was guarded by `if let Some(version)`, so a peer reconnecting on an older build kept a stale entry, received a variant it could not decode, and the connection was **closed** — flap resembling a transport fault. Pre-existing, but hash-first consults the mirror on *every* interest-sync exchange rather than only when PUTting, turning rare into persistent.
- **Not a stdlib type.** `InterestMessage` is core-to-core and appears nowhere outside `crates/core/src`; riverctl/fdev speak the stdlib client API, untouched.
- **Deterministic digest**: truncated BLAKE3 (128-bit), pinned to a literal independently reproduced with `b3sum`. Never a `Hasher` — a per-node digest would make every comparison mismatch and re-create the #4857 shape.

## Telemetry

All three legs classify into the existing `interest_sync_summaries` arm, so that field stays a like-for-like before/after total — if it does not shrink, this did not work. The digest form carries the **same #5061 emitter tag** as the `Summaries` it replaces; without that, a path would lose attribution the moment its peer upgraded, showing a named arm shrinking and the residual growing — indistinguishable from the saving being demonstrated. `SummaryRequestReply` and `SummaryRequest` get their own arms since they are what hash-first *adds*.

Agreement rate is split by **message shape** (single-entry ≈ the state-change-driven sites). A true per-emitter split is not implementable: the tag is `#[serde(skip)]`, so an inbound digest always decodes as `Other` and the receiver — the only side that can judge agreement — cannot know the send site. **Stated confound:** a heartbeat reply for a pair sharing one contract is also single-entry, so this is directional evidence, not attribution.

The comparison counter also gains a third **`one_sided`** bucket. The 98.1% headline was computed over `(Some, Some)` comparisons only, structurally excluding the case where we hold nothing and the peer does — which under hash-first classifies as `NeedBytes`, costing +2 messages for bytes the old path delivered immediately *and* used to seed our cache. Its size was never measured; now it will be.

### Field falsifier — load-bearing, not belt-and-braces

The sim proves the mechanism on the agree path only (see the scope limit above), so **these counters are the primary check at fleet scale**, not a confirmation of something already proven:

| what to read | expectation | what a miss means |
|---|---|---|
| `interest_sync_summaries_bytes` (parent arm) | falls between upgraded pairs | the saving is not landing |
| `interest_sync_summaries_request_leg` / `_request_reply` **message** counts | small relative to `_interests_reply` | mismatch tail is multiplying messages — the #4861 axis |
| single-entry agreement rate | ~98%+ | the heartbeat population does not agree as measured |
| **multi-entry** agreement rate | ~98%+ | **low here is the spurious-mismatch signature** — savings evaporate while correctness holds |
| `summary_entries_one_sided` | small | the phantom-interest population is bigger than assumed |

Read the **parent arm** for the before/after, not the per-emitter split: during adoption the split is not like-for-like, because a pair only uses digests once *both* sides are upgraded.

Note `interest_sync_summaries_request_leg` is deliberately not `..._request` — the latter is a strict prefix of `..._request_reply`, so a dashboard glob would double-count.

## Reviewer note: two pre-existing pins were tightened, and one of mine was vacuous

A PR that tightens pins it did not write reads as scope creep unless the reason is explicit. **These pins were weakened *by* this PR in a way that stayed green** — that is the failure mode worth your attention:

- `summaries_arm_uses_semantic_staleness_probe_pin` anchors on "the next arm". Inserting two arms widened its window to span them; since the digest arm calls the same functions, its assertions would have kept passing **with those calls deleted from `Summaries` itself**.
- `summaries_emitter_sites_are_pinned` scoped files by whether they construct the variant. Centralising construction meant `operations/update.rs` no longer did, so it **dropped out of scope entirely and its two emitters stopped being pinned** — the pin went quiet exactly when it became load-bearing.
- My own `digest_arm_shares_the_single_heal_path` was **vacuous on arrival**: the identifiers it asserts are also in the arm's comments, so deleting both calls left it green. Fixed by stripping comments from the scanned window; verified by mutation.

A **fourth** turned up in review, in a pin I had already "fixed" once: `summaries_arm_uses_semantic_staleness_probe_pin`'s window anchor was `InterestMessage::SummaryDigests { entries }` — a form appearing nowhere in `node.rs` except the pin's own line, which `include_str!` includes. It self-matched, widening the window ~8x. Correcting the needle alone was **still** insufficient: the correct window's own comments satisfied two assertions on prose. Both halves were required.

Filed as **#5076** — now four instances of one mechanism, with "a scrape matching zero things must fail" plus "a needle must resolve to a KNOWN LINE, not merely somewhere" (matched-itself defeats matched-zero), and a note that other scrape pins may already be silently green.

## Release-timing guard

`HASH_FIRST_SUMMARIES_MIN_VERSION = (0, 2, 116)` is only correct if this ships in 0.2.116. The old guard compared against a frozen literal, which catches the floor moving *down* but not the *release* moving up past a stationary floor — the real risk at five releases in four days. If 0.2.116 shipped without this feature, peers on the real 0.2.116 would satisfy the floor while carrying no variant index, fail to decode, and lose their connections during the staggered rollout.

`HASH_FIRST_SHIPPED_IN: Option<(u8,u8,u16)> = None` plus a test coupling the floor to `CARGO_PKG_VERSION` makes that state fail CI instead. Verified by simulating the bump: marker `None` with the crate at the floor fails and names both remedies; `Some(floor)` passes; `Some(mismatch)` fails.

## The per-message cap was measured, not guessed

The receive-side cap started at 256 on the assumption that "a healthy pair shares a handful of contracts". Telemetry says otherwise — `hosting_contract_count` over 16,578 deployed samples:

| p50 | p75 | p90 | p95 | p99 | max |
|-----|-----|-----|-----|-----|-----|
| 417 | 698 | 825 | 976 | 2463 | 2814 |

**73% exceed 256**, so the old value would have bound on the common case rather than the abuse case. Byte metrics could not have revealed this: a `None` entry (a contract the responder doesn't host) costs ~5 bytes, so entry count is decoupled from message size and a cheap 400-entry message is invisible in every byte arm. Now 4096, above the observed maximum. `summaries_entries` (#5061, same release) is the ongoing check.

## Simulation coverage

Sim nodes report `CARGO_PKG_VERSION`, so at 0.2.115 every sim exercised the *fallback* and the first integration run of this change would have been the release PR — ambiguous between the feature and the version bump.

Closed by making the floor overridable. **Sims default it ON**, unlike the two existing floor overrides which fail closed: those gate behavioural *cascades* that load unrelated sims, whereas this changes only the *encoding* of an exchange every sim already runs, with identical convergence semantics.

Full suite on this branch: **137/137 passed** (nextest, hash-first ON).

*Note for anyone reading a red `cargo test` on this suite:* it needs process-per-test isolation (`test_strict_determinism_exact_event_equality`: *"Process isolation via nextest eliminates inter-test DashMap state leakage (#3051)"*). Under `cargo test` this suite produced 10 failures that are entirely absent under nextest. The isolation problem is a property of **the suite sharing a process**, not of the individual tests that happen to mention it — tests that never mention nextest can still be corrupted by a neighbour.

## Testing

- `cargo test -p freenet --lib` — **4508 passed, 0 failed**
- `cargo nextest run -p freenet --test simulation_integration` — **137 passed, 0 failed**
- `cargo clippy --locked -- -D warnings` and the `--features trace-ot` pass — clean
- 22 new unit tests; the A/B simulation; 7 mutations on the core logic, all caught

Refs #4965. Related: #4961, #4956, #4952, #4857, #4861, #5061.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/ring.md
+++ b/.claude/rules/ring.md
@@ -215,7 +215,8 @@ WHEN exchanging peer lists in sync protocols (e.g., interest sync):
   → Stale peer entries cause operations to be sent to dead nodes
 
 WHEN summarizing contracts in the InterestSync heartbeat handlers
-(handle_interest_sync_message: Interests / Summaries / ChangeInterests):
+(handle_interest_sync_message: Interests / Summaries / SummaryDigests /
+SummaryRequest / ChangeInterests):
   → ONLY call get_contract_summary for contracts we host OR actively serve
     AND for which we actually hold state:
     (is_hosting_contract || contract_in_use) && contract_state_present — go
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3659,6 +3659,11 @@ std::thread_local! {
     // for a peer pair sharing exactly ONE contract is also single-entry, so the
     // single bucket is "state-change-driven sites PLUS narrow heartbeats", not
     // a clean partition. It is directional evidence, not attribution.
+    /// Peak size of the digest arm's per-hash local-summary cache (#4965).
+    /// The observable for the RETENTION bound: the cache holds owned summary
+    /// clones, so its peak entry count is what decides whether a hostile
+    /// message can accumulate hundreds of MB.
+    static GLOBAL_SUMMARY_CACHE_PEAK: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_SUMMARY_AGREE_SINGLE: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_SUMMARY_AGREE_MULTI: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_SUMMARY_MISMATCH_SINGLE: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
@@ -3715,6 +3720,7 @@ impl GlobalTestMetrics {
         GLOBAL_SUMMARY_DIGEST_MSGS.with(|c| c.set(0));
         GLOBAL_SUMMARY_FULL_MSGS.with(|c| c.set(0));
         GLOBAL_SUMMARY_FULL_BYTES.with(|c| c.set(0));
+        GLOBAL_SUMMARY_CACHE_PEAK.with(|c| c.set(0));
         GLOBAL_SUMMARY_AGREE_SINGLE.with(|c| c.set(0));
         GLOBAL_SUMMARY_AGREE_MULTI.with(|c| c.set(0));
         GLOBAL_SUMMARY_MISMATCH_SINGLE.with(|c| c.set(0));
@@ -3992,6 +3998,20 @@ impl GlobalTestMetrics {
     pub fn record_summary_full_msg(bytes: u64) {
         GLOBAL_SUMMARY_FULL_MSGS.with(|c| c.set(c.get() + 1));
         GLOBAL_SUMMARY_FULL_BYTES.with(|c| c.set(c.get() + bytes));
+    }
+
+    /// Observe the digest arm's local-summary cache size, keeping the peak.
+    ///
+    /// Records the RETENTION bound directly rather than through a proxy: the
+    /// cache holds owned summary clones, so a peak proportional to the number
+    /// of hashes a peer named — rather than to ONE hash's contract set — is
+    /// the accumulation this is here to catch.
+    pub fn note_summary_cache_size(len: usize) {
+        GLOBAL_SUMMARY_CACHE_PEAK.with(|c| c.set(c.get().max(len as u64)));
+    }
+
+    pub fn summary_cache_peak() -> u64 {
+        GLOBAL_SUMMARY_CACHE_PEAK.with(|c| c.get())
     }
 
     /// One advertised digest matched our own summary, settling that contract
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3646,7 +3646,23 @@ std::thread_local! {
     static GLOBAL_SUMMARY_DIGEST_MSGS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_SUMMARY_FULL_MSGS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_SUMMARY_FULL_BYTES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
-    static GLOBAL_SUMMARY_DIGEST_AGREEMENTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    // #4965 agreement rate, split by MESSAGE SHAPE rather than by emitter.
+    //
+    // The emitter tag (#5052) is `#[serde(skip)]`, so an INBOUND SummaryDigests
+    // always decodes as `Other` — the receiver, which is the only side that can
+    // judge agreement, cannot know which send site produced it. Entry count is
+    // the best available proxy and needs no wire change: the notification and
+    // rejection emitters are single-entry BY CONSTRUCTION while both reply
+    // emitters are multi-entry (see `outbound_message_mix::SummariesDetail`).
+    //
+    // Known ambiguity, stated so the number is not over-read: a heartbeat reply
+    // for a peer pair sharing exactly ONE contract is also single-entry, so the
+    // single bucket is "state-change-driven sites PLUS narrow heartbeats", not
+    // a clean partition. It is directional evidence, not attribution.
+    static GLOBAL_SUMMARY_AGREE_SINGLE: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_SUMMARY_AGREE_MULTI: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_SUMMARY_MISMATCH_SINGLE: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_SUMMARY_MISMATCH_MULTI: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_SUMMARY_BYTE_REQUESTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
 }
 
@@ -3699,7 +3715,10 @@ impl GlobalTestMetrics {
         GLOBAL_SUMMARY_DIGEST_MSGS.with(|c| c.set(0));
         GLOBAL_SUMMARY_FULL_MSGS.with(|c| c.set(0));
         GLOBAL_SUMMARY_FULL_BYTES.with(|c| c.set(0));
-        GLOBAL_SUMMARY_DIGEST_AGREEMENTS.with(|c| c.set(0));
+        GLOBAL_SUMMARY_AGREE_SINGLE.with(|c| c.set(0));
+        GLOBAL_SUMMARY_AGREE_MULTI.with(|c| c.set(0));
+        GLOBAL_SUMMARY_MISMATCH_SINGLE.with(|c| c.set(0));
+        GLOBAL_SUMMARY_MISMATCH_MULTI.with(|c| c.set(0));
         GLOBAL_SUMMARY_BYTE_REQUESTS.with(|c| c.set(0));
     }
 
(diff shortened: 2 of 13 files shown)


sanity (commented):
## Comprehensive review: #5066 (head `94be6fa4b`)

**Review tier: Full** (wire-adjacent protocol surface, centerpiece of a bandwidth release). **Reviewers run:** four independent Claude lenses, blind to each other — big-picture, code-first, testing, skeptical/wire — each reading the checked-out code in its own worktree. CI at this SHA: all 12 checks green.

**Verdict: Needs Changes — Re-review Required After Fix.** Two blocking findings. The mechanism itself survived all four hostile reads: every lens's completeness sweep came back clean (send/receive site inventory complete, version gating a single fail-closed predicate with no drift, wire variants appended and pinned against the real encoder, no new bounded-channel sends, attribution arithmetic correct, loss self-heals within one heartbeat). The defects are in the verification layer and the release mechanics, not the design.

---

### BLOCKING 1 — `summaries_arm_uses_semantic_staleness_probe_pin` is fully disarmed (found independently by all four lenses)

`node.rs:4626-4629`: the end-marker needle omits `, ..`, matches nothing in production code, and self-matches its own string literal at :4627 via `include_str!`. The scan window silently widens from the intended ~208 lines to 1,632 — swallowing the digest arms, later handler arms, and the pin's own rustdoc, which between them satisfy all four assertions with the true `Summaries` arm's calls deleted.

Proven by mutation (testing + skeptical lenses, independently): reverting the arm to the pre-#4857 bare byte compare — the exact ~20M-bad-heal regression this pin exists to catch — leaves the pin GREEN on this branch. The identical mutation is RED on main, so this PR regressed a working pin while its own comment claims to have tightened it.

Fix requires BOTH halves (code-first lens proved the needle alone is insufficient — two assertions remain satisfiable on comments at :3074/:3032 inside the corrected window): needle → `{ entries, .. } => {`, plus routing the window through `code_only()`, which must be hoisted out of `mod hash_first_summaries` where it is currently unreachable. Then re-run the bare-byte-compare mutation and watch all four assertions fail. Add `assert!(next_off > summaries_off)` and a window-size bound. Also file this on #5076 as instance three, with the additional rule both lenses derived: **matched-zero-must-fail does not catch matched-itself — a scrape needle must be verified to resolve to its intended line.**

### BLOCKING 2 — the version floor is a release-timing hazard with no automated guard

`HASH_FIRST_SUMMARIES_MIN_VERSION = (0,2,116)` is correct only if this PR ships in 0.2.116. The guard test pins a hard-coded `LAST_RELEASE_WITHOUT_VARIANTS = (0,2,115)`; if 0.2.116/117 ship without this PR (five releases in the last four days), the floor is stale, the test stays green, and the failure mode is fleet-wide connection churn during the staggered rollout: capable-looking peers send `SummaryDigests`, pre-variant receivers fail bincode decode, and the decode path closes the connection. RELEASING.md's manual check is insufficient at this cadence. Fix: a shipped-marker (`Option<(u16,u16,u16)> = None`) with a unit test asserting `None ⇒ floor > crate version`, so a release bump to the floor's value forces an explicit flip-or-raise decision; RELEASING.md updated to match.

### Design decision (made, two lenses independently recommended it): digest-first ships on the multi-entry REPLY legs only

The 98.1% agreement evidence is from the heartbeat-dominated population. The proactive-notification leg fires immediately after a state change at peers least likely to have applied it; a mismatch there is 1 message → 3 carrying the same bytes anyway — the #4861 messages/s axis on the highest-frequency path — and the per-leg instrumentation is test-only, so the fleet could not tell the legs apart post-release. `Notification` and `Rejection` stay full-bytes this release; extend with field data. The skeptical lens's arithmetic goes in the PR body: expected messages per exchange = 1 + 2(1−0.981^k) — +35% at k=10 shared contracts — bounded on the reply legs by the ~300s heartbeat cadence.

### Significant

- **The 256-entry receive cap contradicts fleet data**: `hosting_contract_count` p50=417, p90=825, p99=2463, 73% of peers above 256 — surfaced by the author's own scheduled telemetry check (F4). Byte metrics could never have shown this: a non-hosted entry costs ~5 B, so high-entry messages are byte-invisible. Raised to 4096 (above max observed 2814), kept as an explicit backstop, distribution written into the comment.
- **The 98.1% denominator excludes the one-sided `(None, Some)` case** — under hash-first that's `NeedBytes` (+2 messages for bytes the old path shipped immediately and cache-seeded), the #4473 phantom-interest shape, unmeasured. A third comparison bucket (`one_sided`) is being added so post-deploy data can size it.
- **Two more unfailable tests found by the testing lens**: `digest_entries_are_deduplicated_and_capped` asserts nothing in ~94% of runs (unseeded `GlobalRng` rotation must land the one known hash in the window; the miss arm passes unconditionally), and the A/B's message assertion `on <= off` algebraically reduces to `mismatches == 0`, which its fixture guarantees — and which any added disagree coverage would trip. Both being fixed with the A/B rework (bounded form + a deliberately-divergent arm asserting a non-empty mismatch population).

### Should fix (batched)

Doc contradicting code at `p2p_protoc.rs:965-969` (`SummaryRequest` listed as consulting the gate; it must not — the chooser there loops the exchange, and a pin enforces its absence); `config.rs:3978` citing a nonexistent test; `.claude/rules/ring.md` still enumerating three summarizing arms of five (the canonical #4473 storm-guard rule); the orphan-arm telemetry guard neutralized by registering the rollup file (re-scope); stem rename (`interest_sync_summaries_request` is a glob-prefix of `_request_reply`); mixed-version node-level sim (the override is already per-node; the current comment overclaims one exists); `SummaryDigestEntry` only-constructor claim unpinned; +1 RTT heal latency on divergence stated as accepted cost; `NeedBytes` deferring the TTL-refresh side effect qualified in body; `remote_version` unbounded growth on disconnect filed separately (pre-existing); the `record_remote_version(Option)` rider named in the Problem section.

### Checked and clean — the short version of four long lists

Send/receive inventory complete crate-wide (no missed constructor, no unhandled arm); exactly one version predicate, fail-closed, mirror-race checked (no await between writes); `SummaryRequest` correctly needs no gate (only a variant-carrying peer can provoke it); wire indices frozen with the new arms pinned against the real bincode config; no `.send().await` on bounded channels in new code; no new per-event telemetry stream; heal path single-implementation (extracted body diffed line-by-line against main — identical); BLAKE3 vector reproduced externally via `b3sum`; four of the new pins mutation-verified as genuinely discriminating; loss of a request/reply self-heals within one heartbeat; A/B premise assertions mutation-verified.

### Verdict

**Needs Changes — Re-review Required After Fix.** HEAD reviewed: `94be6fa4b`. The fix batch (both blockers, the leg restriction, the cap raise, the counters, the doc set) lands as one coherent set; the after-fixes review round covers it on the new SHA. CI must be green on that SHA before merge.

[AI-assisted - Claude]

PR #5067, opened 2026-07-30
fix: correct transport_extended bench match arm for Result<usize>

AkashNaickar:
## Summary

`PeerConnection::send` returns `Result<usize>` but the bench at line 144 still matches `Ok(())`. Changed to `Ok(_)`.

## Testing

Verified with `cargo clippy -p freenet --all-targets`.

### DIFF
--- a/crates/core/benches/transport_extended.rs
+++ b/crates/core/benches/transport_extended.rs
@@ -141,7 +141,7 @@ pub fn bench_high_latency_sustained(c: &mut Criterion) {
                         // Send from A to B
                         let send_result = conn_a.send(message).await;
                         let sent_ok = match send_result {
-                            Ok(()) => true,
+                            Ok(_) => true,
                             Err(e) => {
                                 eprintln!("sustained send failed: {:?}", e);
                                 false


AkashNaickar:
Hi @sanity, both this PR and #5068 pass all CI checks (CLA, rule review, security). Let me know if you'd like any changes.

sanity:
Thank you for this fix, and I'm sorry it took four weeks to get you an answer — that's on us, not on you.

Closing it, because the same fix has already landed by another route: commit `643ee5d35` (PR #5279, merged 2026-08-12) changed the same match arm, and `main` now reads `Ok(_bytes_sent) => true` at line 144. I checked what would happen if we resolved the conflicts on this branch rather than closing it, and the result is a **zero-line diff** — there is genuinely nothing left for it to change.

To be clear about what that means: your diagnosis was correct and the bug was real. It got fixed while your PR sat waiting, which is a failure of our process rather than a judgement on your patch. You found it independently and you were right.

I'm also closing #5028, since the underlying issue is resolved in `main`.

If you want another one, issues labelled `good first issue` are a reasonable place to start, and I'd suggest commenting on the issue before opening a PR so we can confirm nobody else is already on it — that would have caught this collision early. We've just finished a backlog sweep and put automation in place so PRs stop sitting unanswered like this one did.

[AI-assisted - Claude]

Issue #2531, opened 2026-01-01
perf: Consider replacing RSA-2048 with X25519 for transport key exchange


sanity:
## Summary

The transport layer currently uses RSA-2048 for key exchange during connection establishment. X25519 (Curve25519 Diffie-Hellman) would be significantly faster while providing equivalent security.

## Performance Comparison

| Operation | RSA-2048 | X25519 |
|-----------|----------|--------|
| Key generation | ~50-100ms | ~0.05ms |
| Key exchange | ~1-2ms | ~0.1ms |
| Key size | 256 bytes | 32 bytes |

X25519 is roughly **100-1000x faster** for these operations.

## Current Usage

From `crates/core/src/transport/crypto.rs`:
```rust
const BITS: usize = 2048;
let priv_key = RsaPrivateKey::new(rng, BITS).expect("failed to generate a key");
```

RSA is used to encrypt symmetric keys during connection handshake (`Pkcs1v15Encrypt`).

## Proposed Change

Replace RSA with X25519 key exchange + symmetric encryption:

1. Each peer generates an X25519 keypair (ephemeral or long-term)
2. Use ECDH to derive shared secret
3. Derive symmetric key from shared secret (HKDF)
4. Use AES-GCM or ChaCha20-Poly1305 for message encryption

This is the standard approach used by TLS 1.3, WireGuard, Noise Protocol, etc.

## Benefits

- **Faster connection establishment** - especially important for mesh networks with many peers
- **Smaller key material** - 32 bytes vs 256 bytes reduces packet sizes
- **Modern cryptography** - X25519 is the current best practice for key exchange
- **Constant-time operations** - X25519 implementations are inherently constant-time

## Migration Path

1. Add X25519 support alongside RSA
2. Negotiate protocol version during handshake
3. Deprecate RSA after transition period
4. Remove RSA support in future major version

## Relevant Crates

- `x25519-dalek` - Pure Rust X25519 implementation
- `chacha20poly1305` - Already used in the codebase

## Questions to Consider

- Is there a specific reason RSA was chosen over ECDH?
- Are there any compatibility requirements with external systems?

[AI-assisted - Claude]

PR #5068, opened 2026-07-30
fix: prefer exact config.toml over fuzzy config file match

### LINKED ISSUE #5038: read_config picks a nondeterministic config file: config.bak.toml can silently win over config.toml

AkashNaickar:
## Summary

`read_config` uses `read_dir().find_map()` which accepts any `config*.toml` in filesystem iteration order. This means `config.bak.toml` can silently win over `config.toml`.

Now checks for exact `config.toml` / `config.json` first, falling back to fuzzy match only when no exact match exists. Extracted parsing logic into `parse_config_file` helper.

Fixes #5038.


AkashNaickar:
Hi @sanity, both this PR and #5067 pass all CI checks (CLA, rule review, security). Let me know if you'd like any changes.

iduartgomez (approved):

sanity:
@AkashNaickar — an apology first, because this one is genuinely our fault rather than the ordinary kind of delay.

**@iduartgomez approved this PR on 2026-08-07, at the head commit it still has, and then nobody merged it.** It has sat approved for over two weeks. You did everything right and the process dropped it. That's the worst outcome in a backlog sweep we've just run across 59 open PRs, and it's the one I most wanted to find and fix.

**The bug is still real in `main`.** `config.rs:546` still does the fuzzy `find_map`, so a stray `config.bak.toml` can still win over the exact `config.toml`. This is worth landing.

**What it needs to land — a rebase, and a smaller one than the conflict count suggests.** `main` moved underneath it:

1. **Drop the `parse_config_file` extraction.** #5124 rewrote both parse bodies, so that part of your diff is now obsolete — it's the source of most of the conflict noise.
2. **Keep the ~18-line `EXACT_NAMES` preloop** — that is the actual fix and it still applies.
3. **Keep both regression tests.** They're the reason this is mergeable rather than a judgement call.

That should be around thirty minutes of work, and the result is a much smaller diff than what's here now — which will also make the re-review quick.

If you'd rather not spend more time on a PR that's already been through review once, say so and we'll carry it over the line ourselves with credit to you. That's a completely reasonable answer given how long we left it sitting, and no hard feelings either way.

One process note so this doesn't recur: an approved-and-unmerged PR was invisible to us because nothing watched for it. We're adding an automated sweep that reports exactly this state — green, approved, and unmerged past 48 hours — precisely because your PR proved nothing was catching it.

Separately: your other PR #5067 was superseded by a fix that landed while it waited, and I've closed it with an explanation. Your diagnosis there was correct too.

[AI-assisted - Claude]

sanity:
Follow-up with something concrete that should make this PR stronger, plus full disclosure of something we found on our side.

## Disclosure first

While draining this backlog we found an **independently-written fix for the same bug** sitting uncommitted in one of our own worktrees (`config.rs`, +251/-40 — same `EXACT_NAMES` idea, same `config.toml` vs `config.bak.toml` problem, same issue #5038). It was written without knowledge of your PR, which is our failure: your PR was open and approved the whole time.

**We are not landing it.** Your PR is the one that goes in. It was approved on 2026-08-07 and it should have merged then. What follows is the useful part of ours, offered to you.

## One thing worth checking in your test

`read_config_prefers_exact_config_toml_over_backup` writes both `config.toml` and `config.bak.toml` into a temp dir, then asserts that `config.toml` won.

The old code was a single `find_map` over `read_dir`, and **`read_dir` makes no ordering guarantee**. So whether that test can fail without your fix depends on whether the filesystem happens to yield `config.bak.toml` first. If it yields `config.toml` first, the assertion passes with the bug still present.

I want to be precise about what I did and didn't establish: I have **not** run your test against unpatched code, so I am not claiming it is vacuous — only that its ability to fail is not guaranteed by its own construction. This repo's contribution rules ask for exactly this check anyway (*write test → verify it fails → apply fix → verify it passes*), so it is worth confirming: **stash the `EXACT_NAMES` preloop, run the test, and see whether it goes red.** If it does, you're done. If it passes, the test isn't yet pinning the bug.

## What our version did differently, if it's useful

It extracted the selection into a **pure function** (`pick_config_file(names) -> Option<ConfigFilePick>`) taking an iterator of filenames rather than reading a directory. That makes the ordering property directly testable without a filesystem at all, and the test it enables is:

- `an_exactly_named_config_beats_a_config_prefixed_backup_in_any_order` — feeds both orderings explicitly, so it fails without the fix regardless of what any filesystem does
- `a_config_prefixed_file_is_still_accepted_when_nothing_is_exactly_named` — your fallback case, order-independent
- `config_file_selection_prefers_toml_and_ignores_non_config_names`

Take the idea, take none of it, or take it as a follow-up — entirely your call. **The smallest version that resolves the concern is just asserting both orderings**, which doesn't require restructuring anything.

One note on the rest of our diff: it's larger partly because it carries an unrelated change that belongs elsewhere, so `+251/-40` isn't a fair comparison against your `+105/-34`. Yours is the more focused change.

## Still needed to land

The rebase described in the previous comment: drop the now-obsolete `parse_config_file` extraction (#5124 rewrote both parse bodies), keep the `EXACT_NAMES` preloop and the tests. Offer stands — if you'd rather not spend more time on this, say so and we'll carry it over the line with credit to you.

[AI-assisted - Claude]

sanity:
Rebased onto current `main` and pushed to this branch, per @sanity's offer above. Credit stays with @AkashNaickar: both commits are unchanged in authorship, only their content was re-resolved. Head is now `c595553`.

## What conflicted, and how it was resolved

The conflict was entirely with #5124, which rewrote both parse bodies inside `read_config` to add key-spelling-alias normalization (`CONFIG_KEY_SPELLINGS` / `redundant_key_spellings`).

* **Dropped the `parse_config_file` extraction.** It was a faithful copy of the parse bodies as they existed in July, and #5124 has since rewritten both of them. Keeping the extraction meant re-applying the old bodies over the new ones, which was most of the conflict noise and would have silently reverted #5124.
* **Kept the `EXACT_NAMES` preloop.** That is the actual fix and it still applies unchanged in substance. It now assigns into `config_args` and lets the existing `match config_args` do the parsing, rather than returning early through a helper.
* **The #5124 parse arms are untouched.** `git diff origin/main..HEAD` is one file, and the two hunks are the head of `read_config` and the new tests. Nothing inside the `toml` / `json` arms changed.

## Test changes

The original tests used a two-line document (`mode = "local"` + `log_level = "debug"`). That no longer deserializes: `ConfigPaths` is flattened into `Config` and its `PathBuf` fields (`contracts_dir`, `delegates_dir`, `secrets_dir`, `db_dir`, `event_log`, `data_dir`, `config_dir`) have no `#[serde(default)]`, so a minimal document fails to parse before it ever reaches the priority logic. Both tests now build on `released_config_toml_without_secret_paths()`, the helper the neighbouring #5124 tests use, and vary `log_level` to identify which file won.

## Verification, per CONTRIBUTING's write-test-then-verify-it-fails rule

@sanity flagged that the original test's ability to fail depends on what order `read_dir` happens to yield. That turned out to be worth checking, because **as originally written the test could not fail on this machine at all.**

Measured directly against `readdir(2)` on ext4, 2000 trials per arm:

* Iteration order is **completely independent of the order the files were written.** Writing the backup first, or the real config first, produced identical orderings in every one of 2000 trials. So the "write the backup first to bias against the fix" idea does not do anything on ext4.
* With `config.bak.toml` as the only decoy, the **unfixed** code picked `config.toml` anyway in **2000/2000** runs. The test as originally written would have passed against the bug it was written to catch.

ext4 orders directory entries by a per-filesystem hash seed, so which `config*.toml` comes first is deterministic for a given filesystem but arbitrary across filesystems. Two changes came out of that:

1. **`read_config_prefers_exact_config_toml_over_backup` now writes five decoys** rather than one. More competing names means a smaller chance the exact one happens to sort first. Verified: against the unfixed code this now fails **5/5**. That lowers the odds of a vacuous pass; it does not eliminate them, and the test's doc comment says so rather than implying the coverage is stronger than it is.

2. **Added a third test, `read_config_finds_exact_config_without_listing_the_directory`,** which pins the guarantee deterministically and with no dependence on `read_dir` order at all. It makes the directory execute-only (`--x--x--x`), which permits resolving and opening a path through it while denying `readdir`. The unfixed implementation's first act is `std::fs::read_dir(dir)?`, so it fails there with `PermissionDenied` regardless of what order any filesystem would have returned; the exact-match pass only calls `Path::is_file` on the two candidate names, so it is unaffected. This is the test that actually establishes the property the fix claims. It is `#[cfg(unix)]` gated since the mechanism is POSIX permissions and this crate also builds for Windows, and it asserts that the permission denial really happened, so a root CI runner fails loudly instead of going quietly vacuous.

Empirical results, running each test five times against the unfixed `read_config` with the new tests in place:

| test | unfixed code |
|---|---|
| `read_config_prefers_exact_config_toml_over_backup` | FAILED 5/5 (`left: Trace, right: Debug`) |
| `read_config_finds_exact_config_without_listing_the_directory` | FAILED 5/5 (`PermissionDenied`) |
| `read_config_fuzzy_fallback_when_no_exact_match` | PASSED 5/5 (does not depend on the fix, as intended) |

With the fix restored, all three pass on every run (9/9 across three repeats).

## Local checks

`cargo fmt --check` clean, `cargo clippy -p freenet --all-targets -- -D warnings` clean, `cargo test -p freenet --lib config::` 160 passed / 0 failed / 1 ignored (the pre-existing #2531 gateway-key test).

Not merging this myself; it needs the review pass.

[AI-assisted - Claude]

sanity:
## Automated review, and a note on why this has been stuck

Apologies for how long this has sat. A Torvan review session went over this PR at full depth today. Posting its findings here, along with a correction about the approval already on this PR. **This comment is not an approval and does not clear the PR for merge.**

### The issue is real and has become more consequential

`read_config` in `crates/core/src/config.rs` still does an unordered `read_dir().find_map()` fuzzy match on `config*` filenames, so a stray backup file such as `config.bak.toml` can nondeterministically win over `config.toml`. Since #5037 (event-log segment reclamation) merged on 2026-08-30, that nondeterminism can flip `enable-event-log` and trigger irreversible segment deletion on startup, so the impact is worse now than when the issue was filed.

### Review of the current head (`c5955531`)

Four independent reviewer lenses ran against the rebased head. **No blocking findings.**

- The rebase correctly preserved #5127's key-spelling normalization in both the TOML and JSON parse arms, verified byte for byte.
- The three regression tests are non-vacuous. The directory-ordering claim was independently reproduced: the one-decoy test passes 200/200 against unfixed code, the five-decoy test fails 200/200 against unfixed code, matching the empirical verification in the PR description.
- Non-blocking: the fix introduces a new deterministic but undocumented precedence, where an exact `config.toml` always beats an exact `config.json`. That behaviour has no dedicated test.
- Non-blocking: the PR description still refers to a `parse_config_file` extraction that was dropped during the rebase.
- No supply-chain files are touched. The diff is confined to `crates/core/src/config.rs`.

### Two things that block merging right now

**1. The existing approval is against a superseded commit.** The APPROVED review from @iduartgomez is against `5a7c6ee2` from 7 August. This branch was later rebased onto current `main` to resolve a conflict with #5127, and now heads at `c5955531`. That rebase was done by automation on our side, not by @AkashNaickar, so the stale approval is our doing rather than anything the contributor did. It needs a fresh look at the current head before it can count.

**2. CI is not green.** At `c5955531`, NAT Validation is failing and Windows Service Unit is still pending. Everything else passes. That needs resolving before merge regardless of review state.

### Process change

Merging pull requests from outside the project now requires explicit approval from a project developer. Our automation can review and report, as it has here, but it does not merge third-party changes on its own judgement. That change was made today, and it is the reason this comment ends in a handoff rather than a merge.

@iduartgomez when you have a moment, this needs your look at the current head rather than the August commit.

[AI-assisted - Claude]

sanity:
@AkashNaickar — status, and a heads-up that this is blocked on us rather than on you.

**CI has not actually run on the current head.** I updated this branch onto current `main` earlier, which created a new head, and GitHub holds an outside contributor's workflow runs at `action_required` until a maintainer approves them. Only the incidental checks (CLA, Snyk, rule review) have executed. None of the seven required ones — Clippy, Fmt, Unit & Integration, Simulation, NAT Validation, Rule Lint, macOS bundle-swap E2E — has run on `af7a9efb9`.

Worth saying plainly because the PR page reads as though everything is fine: a filter that only shows failures reports "all green" when nothing ran at all. @sanity needs to approve the held runs. I did not approve them myself — that hold exists so a person decides before contributor code runs in our CI.

**On the change itself:** the exact-match-then-fuzzy-fallback shape is right, and the test design is better than most of what lands here. `read_config_finds_exact_config_without_listing_the_directory` making the directory execute-only so `read_dir` fails with `PermissionDenied` — while `Path::is_file` on the two candidate names still resolves — pins the guarantee without depending on filesystem iteration order, and the doc comment is honest that the other test would have passed against the unfixed code. That candour is the right instinct; keeping the fuzzy fallback so nothing that used to load stops loading is the other half.

An independent review pass is running against it now and I will post anything it turns up. Nothing so far suggests a change is needed from you.

[AI-assisted - Claude]

sanity:
## Review complete — findings applied in `fdcadafd0`

@AkashNaickar — reviewed properly now. The fix is right and I have pushed three changes on top rather than round-tripping them to you. Two came out of the review; one is a sharp edge in the surrounding code that your change makes permanent.

### Your central guard really does work — verified, not assumed

Transcribed all three new tests onto our own `main` and ran them against the **unfixed** code (the third-party-code hook correctly prevents building your fork's branch here, so nothing of yours was executed on this machine):

| test | unfixed | with fix |
|---|---|---|
| `read_config_finds_exact_config_without_listing_the_directory` | **FAILED** — `PermissionDenied` | ok |
| `read_config_prefers_exact_config_toml_over_backup` | **FAILED** — `left: Trace, right: Debug` | ok |
| `read_config_fuzzy_fallback_when_no_exact_match` | passed | ok |

The third passing either way is correct and your own naming says so — it is a non-regression guard, not a regression test.

Your doc comment's candour about the second test's limited power is exactly the right instinct, and it is what led to the first fix below.

### 1. The main guard was vacuous on tmpfs

Its detection power is a property of the filesystem rather than of the code. On ext4 (nova) the unfixed code picks `config.aaa.toml` 2000/2000 with your five decoys — matching your numbers. On **tmpfs** (`/dev/shm`) it picks `config.toml` **400/400 even with all five decoys**: tmpfs is insertion-ordered and the test wrote `config.toml` last, so the test passed against the bug.

`tempfile::tempdir()` honours `TMPDIR`, so a runner with `TMPDIR` on tmpfs would silently neuter it. Writing the exact name **first** makes the unfixed code pick a decoy on both filesystems. One-line reorder.

### 2. The root assertion broke a supported local workflow

`docker/test-runner/Dockerfile` is `FROM rust:latest` with no `USER`, and `run.sh` runs `docker run` with no `--user`, with `CMD ["cargo","test","--workspace"]` — that is what `/freenet:linux-test` drives. Root bypasses POSIX permission checks, so the `--x--x--x` setup cannot produce its fault and your `assert_eq!(listing, Err(PermissionDenied))` fires.

GitHub CI is unaffected (no `container:` in any workflow; `test_unit` runs as the non-root runner user). But the obvious "fix" for a test that fails locally is to delete the assertion — which is the one real regression guard in this PR.

It now **skips loudly** under root, detecting root by the readdir succeeding when it should not, and says in the skip message that the guard is exercised by the non-root CI run. Your instinct to fail rather than skip silently was right; this keeps the loudness without breaking the container path.

### 3. `config.toml` now permanently shadows `config.json` — and the node creates the shadow

This one is not really about your diff, but your diff makes it deterministic and permanent, so it is worth closing here.

`build()` persists unconditionally (`config.rs:1534-1537`), so on a first run it writes `config.toml`. A `config.json` operator therefore boots once, gets a `config.toml` created beside theirs, and from boot two the exact-match pass reads the toml and never their json again — later edits silently discarded. That is the same invisible-config failure class #5038 is about. Previously it was an order-dependent coin flip; now it is settled, in toml's favour, forever.

Narrowing the fix would be worse, so instead it is no longer silent: loading `config.toml` while a `config.json` exists now warns and names which one won. Same reasoning applied to the fuzzy fallback, which was whispering at `debug` — an operator running on `config.bak.toml` is exactly the person #5038 is about, and `debug` means nobody sees it. That branch now warns too.

### Still blocked, and not on you

**CI has never run on this head.** The workflow runs are `action_required` — GitHub holds an outside contributor's workflows until a maintainer approves them. The four green checks on the PR page are only `Claude Rule Review`, CLA, `rule-review/warnings` and Snyk; none of them compiles or tests anything. The last run that actually executed was on a superseded commit and failed on NAT Validation. @sanity needs to release the runs; I deliberately did not approve them myself, since that hold exists so a person decides before contributor code runs in our CI.

### Smaller things not changed

- `Fixes #5038` will auto-close an issue whose second variant (a `--config-dir` that exists but holds no `config*` returns `Ok(None)` → silent defaults) is untouched. Worth either handling it or downgrading to `Refs #5038`.
- The PR description still claims a `parse_config_file` extraction that is not in the diff — dropped in a rebase.
- On case-insensitive filesystems `dir.join("config.toml").is_file()` matches `Config.TOML`, which the old case-sensitive scan ignored. Behaviour change, almost certainly a harmless one.

For the record, several things I went looking for are genuinely fine: `is_file()` is correct over `exists()` (a *directory* named `config.toml` would otherwise reach `File::open` and fail EISDIR); error handling is not weakened, since both call sites still propagate; and no second reader diverges — `read_config` holds the only fuzzy `config*` scan in the tree, and the writer and the diagnostic report already use the exact name, so your fix makes reader and writer agree.

[AI-assisted - Claude]

sanity:
Rule review's warning taken, in `45d78fba2` — and it was right that this is not a peripheral branch.

**The one architectural decision in this PR, stated plainly so it is a decision rather than a side effect:** `config.toml` now beats `config.json` *permanently*. Before, the fuzzy scan took whichever entry `read_dir` yielded first, so which format won was arbitrary. Now it is settled. And because `build()` writes `config.toml` unconditionally, a `config.json` operator gets one created beside theirs on first boot, after which their json is never read again.

That is small in practice — json config is likely rare — but it is a real semantic change that nobody explicitly chose, so it now has a test recording it. If the precedence is ever revisited, `read_config_prefers_config_toml_over_a_valid_config_json` is where the decision lives. The shadowing case also warns and names which file won, so an affected operator is told rather than left wondering.

**A note on how the test nearly went wrong,** because it is the more useful half. My first version wrote the decoy by hand as `{"log_level": "error"}`. The test failed — not because precedence was wrong, but because that json is missing required fields (`missing field 'mode'`). Had it loaded at all, the test would have "passed" for entirely the wrong reason: json failing to parse, rather than toml being preferred.

The decoy is now derived from the same fixture as the toml, round-tripped through `toml::Value` into JSON with one value flipped, so it is a genuinely loadable config. The assertion that loads the json **alone** first exists precisely to keep that failure mode caught rather than assumed — and it is what caught it.

Mutation-verified: reordering `EXACT_NAMES` so json is probed first fails the test (`left: Error, right: Debug`). Restored, all four config tests pass.

Also took the Info item — the fuzzy fallback's `json` arm still returned silently while its `toml` sibling warned. Same reasoning applies (a fuzzy-matched `config*.json` is just as invisibly-not-the-file-you-edited), so it warns too and the asymmetry I introduced is gone.

@AkashNaickar — for the record, the core of this is yours and it is good: exact-match-then-fuzzy-fallback is the right shape, keeping the fallback means nothing that used to load stops loading, and making the directory execute-only so `read_dir` fails while `Path::is_file` still resolves is a genuinely clever way to pin the guarantee without depending on filesystem iteration order.

[AI-assisted - Claude]

PR #5071, opened 2026-07-30
test(wasm_runtime): offline self-delta / summarize-determinism probe harness

sanity:
## Problem

We want to enforce a staleness-probe invariant: *if a peer's summary equals my state's summary, they hold what I hold, so `get_state_delta(S, summarize_state(S))` must be empty.* The only thing blocking it is fear of false positives (#4295 froze the ping contract permanently that way).

That fear was never converted into data. There was no way to answer "does this hold for the contracts the network actually hosts?" without a harness.

## Approach

A test-gated analysis harness that runs the **real** runtime (`Runtime` + wasmtime + `ContractStore`, so Cranelift `OptLevel::None` exactly as production) over (contract WASM, real state) pairs and records, per pair:

- `summarize_state` called **three times**, each on a fresh WASM instance, with full byte comparison — so a non-deterministic summarize (the #4857 HashMap-iteration-order class) is reported *separately* from a broken delta. Those two failures need different fixes and must not be conflated.
- `get_state_delta(S, first_summary)` length, plus the **first 32 bytes of the delta**. The head is the load-bearing part: it distinguishes a semantically-empty encoding (CBOR `null` = `f6`, or a map of empty collections) from a byte-identical whole-state dump. `ring/interest.rs:1856` and `:1944` judge peer staleness by raw `!delta.as_ref().is_empty()`, so those two cases look identical to core but are completely different bugs.
- `summarize_state(S).len()` vs `S.len()` — a summary at or above state size is a bug in every case and is nearly free to collect.

It also records the instance id derived from (params, code) alongside the stored one, which is a free check that the harness is running the right code with the right params.

Driven entirely by `PROBE_DUMP` / `PROBE_WASM` / `PROBE_OUT`. With those unset it returns immediately, so it costs CI nothing.

## Testing

Run against the full contract set the nova gateway hosts: **2470 of 2489 stored states, covering all 190 distinct code hashes** (the other 19 rows have no `contract_index` entry, so their code hash is unknown). The derived instance id matched the stored one for all 2470.

Findings are reported separately; the short version is that the invariant does **not** hold universally — 818 of 2470 states return a non-empty self-delta — and that several of the failures are semantically-empty encodings rather than broken delta mechanisms, which is exactly the distinction a naive `is_empty()` enforcing would get wrong.

## Note

This is an analysis harness, not a product change. It is offered because it is the fixture you would want to re-run before enabling any enforcing probe, and because the same shape could become a CI gate for the invariant. If the team would rather not carry it in-tree, closing this is fine — the findings stand on their own.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/wasm_runtime/tests.rs
+++ b/crates/core/src/wasm_runtime/tests.rs
@@ -7,6 +7,7 @@ mod cache;
 mod contract;
 mod contract_metering;
 mod execution_handling;
+mod self_delta_probe;
 mod time;
 
 pub(crate) fn get_test_module(name: &str) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
--- a/crates/core/src/wasm_runtime/tests/self_delta_probe.rs
+++ b/crates/core/src/wasm_runtime/tests/self_delta_probe.rs
@@ -1,0 +1,187 @@
+//! Offline probe: does `get_state_delta(S, summarize_state(S))` return EMPTY?
+//!
+//! Analysis harness (not a CI test). Driven entirely by environment variables;
+//! with none set it is a no-op, so it costs CI nothing.
+//!
+//! Usage:
+//!   PROBE_DUMP=/path/to/dump PROBE_WASM=/path/to/contracts \
+//!   PROBE_OUT=/path/to/results.jsonl \
+//!   cargo test -p freenet --lib self_delta_probe -- --nocapture --test-threads=1
+//!
+//! `PROBE_DUMP` must contain `manifest.json` plus `state/<instance>.bin` and
+//! `params/<instance>.bin`, as produced by the redb dumper.
+
+use std::sync::Arc;
+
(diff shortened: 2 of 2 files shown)


sanity:
Closing during a backlog sweep, taking the PR up on its own offer: "If the team would rather not carry it in-tree, closing this is fine — the findings stand on their own."

It has sat as a draft since 2026-07-30 with no further pushes, no comments and no reviews, and no linked issue, so nothing is waiting on it. The findings it produced (2470 of 2489 nova-hosted states probed, 818 returning a non-empty self-delta, and the semantically-empty-vs-whole-state-dump distinction that a naive `is_empty()` staleness probe would get wrong) are the durable part, and they are recorded independently of this branch.

Worth reopening if anyone moves to enable an enforcing staleness probe — this is the fixture you would want to re-run first. The branch is untouched.

[AI-assisted - Claude]

PR #2954, opened 2026-02-10
feat: nix flake support

tt0fu:
Added a way for nix users to run freenet with `nix run` and add install freenet via flake inputs.
Be aware that the version number in the flake must be updated with the version of the crate.

PR #5073, opened 2026-07-30
build: update and fix the nix flake

Skyb0rg007:
This implements a bunch of changes to the nix `flake.nix`. The commit messages likely need to be reworded as I'm not sure what some of the prefixes should be.

## Summary of the changes

1. The nix build now works. The flake.lock will now be automatically updated via dependabot so hopefully there won't be any issues with the Rust compiler lagging behind. There are options for converting the toolchain.toml file into Nix, but doing so would require a third-party dependency.

2. The freenet-autoupdate script now uses git tags. Previously every invocation would use the latest commit on `main`

3. The freenet build is now reproducible! Doing so requires the definition of `SOURCE_DATE_EPOCH` (nix does this for you), but its easy to do this for the other CI workflows too. This helps with security as anyone could verify that the binaries on the GitHub releases page were not tampered with.

4. A basic Nix `devShell` was added, to make it easier for nix-users to contribute

5. Allow specifying git revision information via environment variables (`FREENET_GIT_COMMIT_HASH` and `FREENET_GIT_IS_DIRTY`). This was necessary for nix, but can also be useful for building packages for other systems: you can now download a GitHub archive tarball and still embed the proper git commit information instead of needing to depend on the full git commit history in `.git`.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,81 @@
+name: Nix
+
+# Nothing else in CI evaluates flake.nix, package.nix or the devShell, so the
+# weekly nixpkgs-unstable bump from dependabot would otherwise land with no
+# build signal behind it.
+#
+# NOT a required check: path-filtered workflows do not reliably produce a check
+# context for branch protection (see the note in playwright-shell.yml).
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'flake.nix'
+      - 'flake.lock'
+      - 'package.nix'
+      - 'rust-toolchain.toml'
+      - 'Cargo.lock'
+      - 'crates/core/build.rs'
+      - '.github/workflows/nix.yml'
+  pull_request:
+    paths:
+      - 'flake.nix'
+      - 'flake.lock'
+      - 'package.nix'
+      - 'rust-toolchain.toml'
+      - 'Cargo.lock'
+      - 'crates/core/build.rs'
+      - '.github/workflows/nix.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  nix-build:
+    name: Nix flake build
+    runs-on: ubicloud-standard-8
+    # A cold build compiles the whole dependency graph; no binary cache exists.
+    timeout-minutes: 90
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: cachix/install-nix-action@v31
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+
+      - name: Check flake outputs evaluate
+        run: nix flake check --no-build
+
+      - name: Build the freenet package
+        run: nix build .#freenet -L
+
+      # The devShell breaks independently of the package. `cargo metadata` runs
+      # cargo against the real workspace; a bare `cargo --version` would not.
+      - name: Check the devShell opens and cargo works inside it
+        run: |
+          nix develop --command rustc --version
+          nix develop --command cargo metadata --format-version 1 > /dev/null
+          # Asserted directly because `cargo metadata` runs no build script, so
+          # it passes in a shell where `cargo clippy --workspace` cannot compile.
+          nix develop --command pkg-config --exists openssl
+
+      # A shell whose rustc differs from the pin gives clippy and rustfmt results
+      # that disagree with CI.
+      - name: Verify the devShell matches rust-toolchain.toml
+        run: |
+          pinned="$(grep -oP 'channel\s*=\s*"\K[^"]+' rust-toolchain.toml)"
+          actual="$(nix develop --command rustc --version | awk '{print $2}')"
+          if [ "$pinned" != "$actual" ]; then
+              echo "devShell rustc is $actual but rust-toolchain.toml pins $pinned" >&2
+              exit 1
+          fi
+          echo "devShell rustc matches the pin: $actual"
+
+      # writeShellApplication runs shellcheck at build time, so building the
+      # generated wrapper is the check.
+      - name: Build the auto-update wrapper
+        run: nix build .#freenet-autoupdate -L
--- a/flake.nix
+++ b/flake.nix
@@ -2,6 +2,10 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
     flake-utils.url = "github:numtide/flake-utils";
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
   };
 
   outputs =
@@ -9,6 +13,7 @@
       self,
       nixpkgs,
       flake-utils,
+      rust-overlay,
       ...
     }:
     let
@@ -114,13 +119,46 @@
       let
         pkgs = import nixpkgs {
           inherit system;
-          overlays = [ self.overlays.default ];
+          overlays = [
+            rust-overlay.overlays.default
+            self.overlays.default
+          ];
         };
+
+        # The repo's pinned toolchain, not whatever nixpkgs-unstable ships:
+        # clippy and rustfmt only match CI when the compiler version does.
+        toolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+        rustPlatform = pkgs.makeRustPlatform {
+          cargo = toolchain;
+          rustc = toolchain;
+        };
+
+        # Applied here rather than in the overlay, so downstream consumers of
+        # the overlay still get a plain-nixpkgs package.
+        freenet = pkgs.freenet.override { inherit rustPlatform; };
       in
       {
         packages = rec {
-          inherit (pkgs) freenet freenet-autoupdate;
+          inherit freenet;
+          inherit (pkgs) freenet-autoupdate;
           default = freenet-autoupdate;
+        };
+        devShells.default = pkgs.mkShell {
+          inputsFrom = [ freenet ];
+          # Not needed by the freenet binary, but release-agent's openssl-sys
+          # fails to build without them, so `cargo clippy --workspace` in this
+          # shell would not compile.
+          nativeBuildInputs = [ pkgs.pkg-config ];
+          buildInputs = [ pkgs.openssl ];
+          packages = [
+            toolchain
+            pkgs.cargo-nextest
+            pkgs.jq
+            pkgs.nixfmt
+            pkgs.pre-commit
+            pkgs.python3
+            pkgs.shellcheck
+          ];
         };
       }
     );
--- a/flake.nix
+++ b/flake.nix
@@ -14,21 +14,27 @@
     {
       overlays.default = _: pkgs: {
         freenet = pkgs.callPackage (import ./package.nix) { };
-        freenet-autoupdate = pkgs.writeShellScriptBin "freenet-autoupdate" ''
-          trap "" INT
+        freenet-autoupdate = pkgs.writeShellApplication {
+          name = "freenet-autoupdate";
+          runtimeInputs = [ pkgs.gh pkgs.nix ];
+          text = ''
+            while true; do
+                vsn="$(gh release view --repo freenet/freenet-core \
+                  --json tagName --jq .tagName)"
+                echo "Running freenet $vsn"
+                exit_code=0
+                nix --option experimental-features 'nix-command flakes' \
+                  run "github:freenet/freenet-core/$vsn#freenet" || exit_code=$?
 
-          while true; do
-              nix run github:freenet/freenet-core#freenet
-              exit_code=$?
-
-              if [ $exit_code -eq 42 ]; then
-                  echo "Autoupdate triggered. Restarting..."
-              else
-                  echo "Non-autoupdate exit code: $exit_code. Stopping."
-                  break
-              fi
-          done
-        '';
+                if [ $exit_code -eq 42 ]; then
+                    echo "Autoupdate triggered. Restarting..."
+                else
+                    echo "Non-autoupdate exit code: $exit_code. Stopping."
+                    exit $exit_code
+                fi
+            done
+          '';
+        };
       };
     }
     // flake-utils.lib.eachDefaultSystem (
--- a/flake.nix
+++ b/flake.nix
@@ -14,21 +14,78 @@
     {
       overlays.default = _: pkgs: {
         freenet = pkgs.callPackage (import ./package.nix) { };
-        freenet-autoupdate = pkgs.writeShellScriptBin "freenet-autoupdate" ''
-          trap "" INT
+        freenet-autoupdate = pkgs.writeShellApplication {
+          name = "freenet-autoupdate";
+          runtimeInputs = [
+            pkgs.curl
+            pkgs.nix
+          ];
+          text = ''
+            # So the node exits 42 for us instead of logging that it will not update.
+            export FREENET_SUPERVISED=1
 
-          while true; do
-              nix run github:freenet/freenet-core#freenet
-              exit_code=$?
+            # Mirrors MAX_UPDATE_FAILURES (crates/core/src/bin/commands/update.rs):
+            # stop if the node keeps asking to update but the tag never moves.
+            max_failures=3
+            failures=0
+            last_vsn=""
 
-              if [ $exit_code -eq 42 ]; then
-                  echo "Autoupdate triggered. Restarting..."
-              else
-                  echo "Non-autoupdate exit code: $exit_code. Stopping."
-                  break
-              fi
-          done
-        '';
+            # ±20% jitter, mirroring UPDATE_REPOLL_JITTER_FRACTION
+            # (crates/core/src/bin/commands/auto_update.rs). Every instance
+            # polls the same releases/latest redirect and is handed exit 42 at
+            # roughly the same wall-clock moment when a release lands, so a
+            # fixed sleep makes them all retry in lockstep.
+            jittered() {
+                printf '%s' "$(( $1 * 8 / 10 + RANDOM % ($1 * 4 / 10 + 1) ))"
+            }
+
+            while true; do
+                # The auth-free redirect the node itself uses
+                # (GITHUB_LATEST_REDIRECT_URL in .../commands/auto_update.rs).
+                # Not api.github.com: that spends REST quota shared per source
+                # IP (#5102), and needs a token this wrapper should not require.
+                vsn="$(curl -fsSL -o /dev/null -w '%{url_effective}' \
+                  -A freenet-updater \
+                  https://github.com/freenet/freenet-core/releases/latest \
+                  | sed -n 's|.*/releases/tag/||p')" || vsn=""
+
+                if [ -z "$vsn" ]; then
+                    # A network blip must not kill the supervisor.
+                    delay="$(jittered 60)"
+                    echo "Could not resolve the latest freenet release; retrying in $delay seconds." >&2
+                    sleep "$delay"
+                    continue
+                fi
+
+                echo "Running freenet $vsn"
+                exit_code=0
+                nix --extra-experimental-features 'nix-command flakes' \
+                  run "github:freenet/freenet-core/$vsn#freenet" || exit_code=$?
+
+                if [ "$exit_code" -eq 42 ]; then
+                    if [ "$vsn" = "$last_vsn" ]; then
+                        failures=$((failures + 1))
+                    else
+                        failures=0
+                    fi
+                    last_vsn="$vsn"
+
+                    if [ "$failures" -ge "$max_failures" ]; then
+                        echo "freenet $vsn requested an update $max_failures times without the" >&2
+                        echo "released version changing. Giving up." >&2
+                        exit 1
+                    fi
+
+                    delay="$(jittered $((60 * (failures + 1))))"
+                    echo "Autoupdate triggered. Restarting in $delay seconds..."
+                    sleep "$delay"
+                else
+                    echo "Non-autoupdate exit code: $exit_code. Stopping."
+                    exit "$exit_code"
+                fi
+            done
+          '';
+        };
       };
     }
     // flake-utils.lib.eachDefaultSystem (
--- a/flake.nix
+++ b/flake.nix
@@ -11,14 +11,35 @@
       flake-utils,
       ...
     }:
+    let
+      inherit (nixpkgs) lib;
+
+      # `rev` exists only for a clean git source, `dirtyRev` only for a dirty one,
+      # and neither for a tarball, `path:` or non-git source drop — where an
+      # unguarded dereference is a hard eval failure, not a fallback.
+      gitCommitHash =
+        if self ? rev then
+          lib.substring 0 12 self.rev
+        else if self ? dirtyRev then
+          lib.substring 0 12 (lib.removeSuffix "-dirty" self.dirtyRev)
+        else
+          "unknown";
+
+      # Dirty only when nix says so: treating "no VCS metadata" as dirty would
+      # disable auto-update for every source-drop build (see crates/core/build.rs).
+      gitDirty = self ? dirtyRev;
+    in
     {
-      overlays.default = _: pkgs: {
-        freenet = pkgs.callPackage (import ./package.nix) { };
-        freenet-autoupdate = pkgs.writeShellApplication {
+      overlays.default = final: prev: {
+        freenet = final.callPackage (import ./package.nix) {
+          inherit gitCommitHash gitDirty;
+          sourceDateEpoch = toString self.lastModified;
+        };
+        freenet-autoupdate = final.writeShellApplication {
           name = "freenet-autoupdate";
           runtimeInputs = [
-            pkgs.curl
-            pkgs.nix
+            final.curl
+            final.nix
           ];
           text = ''
             # So the node exits 42 for us instead of logging that it will not update.
--- a/flake.nix
+++ b/flake.nix
@@ -63,6 +63,19 @@
           inherit (pkgs) freenet freenet-autoupdate;
           default = freenet-autoupdate;
         };
+        devShells.default = pkgs.mkShell {
+          inputsFrom = [ self.packages.${system}.freenet ];
+          packages = [
+            pkgs.cargo-nextest
+            pkgs.clippy
+            pkgs.jq
+            pkgs.nixfmt
+            pkgs.pre-commit
+            pkgs.python3
+            pkgs.rustfmt
+            pkgs.shellcheck
+          ];
+        };
       }
     );
 }
--- a/flake.nix
+++ b/flake.nix
@@ -11,12 +11,25 @@
       flake-utils,
       ...
     }:
+    let
+      inherit (nixpkgs) lib;
+    in
     {
       overlays.default = _: pkgs: {
-        freenet = pkgs.callPackage (import ./package.nix) { };
+        freenet = pkgs.callPackage (import ./package.nix) {
+          gitDirty = !(self ? rev);
+          gitCommitHash =
+            if self ? rev then
+              lib.substring 0 12 self.rev
+            else
+              lib.substring 0 12 (lib.removeSuffix "-dirty" self.dirtyRev);
+        };
         freenet-autoupdate = pkgs.writeShellApplication {
           name = "freenet-autoupdate";
-          runtimeInputs = [ pkgs.gh pkgs.nix ];
+          runtimeInputs = [
+            pkgs.gh
+            pkgs.nix
+          ];
           text = ''
             while true; do
                 vsn="$(gh release view --repo freenet/freenet-core \
--- a/package.nix
+++ b/package.nix
@@ -1,28 +1,47 @@
 {
-  rustPlatform,
   lib,
+  rustPlatform,
+  gitCommitHash ? "unknown",
+  gitDirty ? false,
+  # stdenv's default is 1980-01-01: reproducible, but useless in a bug report.
+  # The flake passes `self.lastModified`, which is both.
+  sourceDateEpoch ? null,
 }:
+assert lib.assertMsg (
+  gitCommitHash == "unknown"
+  || (
+    builtins.stringLength gitCommitHash <= 40 && builtins.match "[0-9a-fA-F]+" gitCommitHash != null
+  )
+) "gitCommitHash (${gitCommitHash}) must be <= 40 hex characters";
 rustPlatform.buildRustPackage {
   pname = "freenet";
-  version = (fromTOML (builtins.readFile ./crates/core/Cargo.toml)).package.version;
+  version = (builtins.fromTOML (builtins.readFile ./crates/core/Cargo.toml)).package.version;
 
   src = ./.;
 
-  cargoLock = {
-    lockFile = ./Cargo.lock;
+  cargoLock.lockFile = ./Cargo.lock;
+
+  env = {
+    FREENET_GIT_COMMIT_HASH = gitCommitHash;
+    # Not a Nix bool: `false` coerces to "", which build.rs reads as "no override".
+    FREENET_GIT_IS_DIRTY = if gitDirty then "1" else "0";
+  }
+  // lib.optionalAttrs (sourceDateEpoch != null) {
+    SOURCE_DATE_EPOCH = sourceDateEpoch;
   };
 
   cargoBuildFlags = [
-    "--package"
-    "freenet"
-    "--bin"
-    "freenet"
+    "--package=freenet"
+    "--bin=freenet"
   ];
 
   doCheck = false;
 
-  meta = with lib; {
+  meta = {
+    description = "Peer-to-peer platform for decentralized applications";
     homepage = "https://github.com/freenet/freenet-core";
-    license = licenses.agpl3Only;
+    donationPage = "https://freenet.org/donate/";
+    license = lib.licenses.agpl3Only;
+    mainProgram = "freenet";
   };
 }
--- a/package.nix
+++ b/package.nix
@@ -1,28 +1,34 @@
 {
-  rustPlatform,
   lib,
+  rustPlatform,
+  gitCommitHash ? "unknown",
+  gitDirty ? false,
 }:
 rustPlatform.buildRustPackage {
   pname = "freenet";
-  version = (fromTOML (builtins.readFile ./crates/core/Cargo.toml)).package.version;
+  version = (builtins.fromTOML (builtins.readFile ./crates/core/Cargo.toml)).package.version;
 
   src = ./.;
 
-  cargoLock = {
-    lockFile = ./Cargo.lock;
+  cargoLock.lockFile = ./Cargo.lock;
+
+  env = {
+    FREENET_GIT_COMMIT_HASH = gitCommitHash;
+    FREENET_GIT_IS_DIRTY = gitDirty;
   };
 
   cargoBuildFlags = [
-    "--package"
-    "freenet"
-    "--bin"
-    "freenet"
+    "--package=freenet"
+    "--bin=freenet"
   ];
 
   doCheck = false;
 
-  meta = with lib; {
+  meta = {
+    description = "Peer-to-peer platform for decentralized applications";
     homepage = "https://github.com/freenet/freenet-core";
-    license = licenses.agpl3Only;
+    donationPage = "https://freenet.org/donate/";
+    license = lib.licenses.agpl3Only;
+    mainProgram = "freenet";
   };
 }
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -81,6 +81,16 @@ updates:
           - "minor"
           - "patch"
 
+  - package-ecosystem: "nix"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    # Put all updates in one PR
+    groups:
+      nix:
+        patterns:
+          - "*"
+
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,13 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every weekday
       interval: "daily"
+
+  - package-ecosystem: "nix"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Put all updates in one PR
+    groups:
+      nix:
+        patterns:
+          - "*"
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,7 @@ logs/
 
 # Local-only: release-agent binary staged for install.sh (built per-machine)
 scripts/release-agent/freenet-release-agent
+
+# Nix
+/result
+/result-*
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,7 @@ logs/
 
 # Local-only: release-agent binary staged for install.sh (built per-machine)
 scripts/release-agent/freenet-release-agent
+
+# Nix
+result
+result-*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,27 @@ When in doubt about which bucket you're in, file an issue first. It costs you a 
 - **Read the codebase conventions.** See [AGENTS.md](AGENTS.md) for project structure, coding standards, and testing requirements.
 - **Ask questions.** If something is unclear, ask on the issue or in our [Matrix channel](https://matrix.to/#/#freenet:matrix.org) before writing code.
 
+## Building with Nix (optional)
+
+An alternative to a local rustup install, not a requirement.
+
+```bash
+nix develop                  # dev shell: pinned toolchain, nextest, shellcheck, …
+nix build .#freenet          # build the node binary (result/bin/freenet)
+nix run .                    # run the auto-updating wrapper
+```
+
+The shell's toolchain is pinned to [`rust-toolchain.toml`](rust-toolchain.toml),
+so `cargo fmt` and `cargo clippy -- -D warnings` inside it agree with CI. Builds
+are reproducible: the same source gives a bit-identical binary, since
+`BUILD_TIMESTAMP` comes from `SOURCE_DATE_EPOCH` and the git provenance from the
+flake rather than a `.git` directory.
+
+Packagers building outside Nix can supply that provenance with
+`FREENET_GIT_COMMIT_HASH` (≤40 hex characters) and `FREENET_GIT_IS_DIRTY`
+(`1`/`true`, `0`/`false`, or empty to detect it with `git`). The latter gates
+auto-update — a build marked dirty will never update itself.
+
 ## Quality Standards
 
 - PR titles must follow [Conventional Commits](https://www.conventionalcommits.org/) format (`feat:`, `fix:`, `docs:`, etc.) — CI enforces this.
--- a/crates/core/build.rs
+++ b/crates/core/build.rs
@@ -1,5 +1,5 @@
-use std::process::Command;
 use chrono::TimeZone;
+use std::process::Command;
 
 fn main() {
     // Emit build metadata for startup logging
@@ -110,23 +110,47 @@ fn read_min_compatible_from_cargo_toml() -> Option<String> {
 }
 
 fn emit_build_metadata() {
-    // Git commit hash
-    let git_hash = Command::new("git")
-        .args(["rev-parse", "--short=12", "HEAD"])
-        .output()
-        .ok()
-        .and_then(|o| String::from_utf8(o.stdout).ok())
-        .map(|s| s.trim().to_string())
-        .unwrap_or_else(|| "unknown".to_string());
+    // Git commit hash. FREENET_GIT_COMMIT_HASH supplies it for source drops with
+    // no `.git`; empty means "no override", as for FREENET_GIT_IS_DIRTY below.
+    let git_hash = match std::env::var("FREENET_GIT_COMMIT_HASH") {
+        Ok(v) if !v.trim().is_empty() => {
+            let v = v.trim().to_string();
+            // Validated because an embedded newline in a `cargo:` directive would
+            // emit further directives of the caller's choosing.
+            if v.len() > 40 || !v.chars().all(|c| c.is_ascii_hexdigit()) {
+                panic!("FREENET_GIT_COMMIT_HASH ({v}) must be <= 40 hex characters");
+            }
+            v
+        }
+        _ => Command::new("git")
+            .args(["rev-parse", "--short=12", "HEAD"])
+            .output()
+            .ok()
+            .and_then(|o| String::from_utf8(o.stdout).ok())
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty())
+            .unwrap_or_else(|| "unknown".to_string()),
+    };
     println!("cargo:rustc-env=GIT_COMMIT_HASH={git_hash}");
 
-    // Git dirty flag
-    let git_dirty = Command::new("git")
-        .args(["status", "--porcelain"])
-        .output()
-        .ok()
-        .map(|o| !o.stdout.is_empty())
-        .unwrap_or(false);
+    // Git dirty flag. Parsed strictly: GIT_DIRTY is the auto-update kill switch
+    // (`auto_update_is_disabled` in src/bin/freenet.rs), so a lenient
+    // "any non-empty value is dirty" rule would let FREENET_GIT_IS_DIRTY=false
+    // ship a release binary that never updates itself.
+    let git_dirty = match std::env::var("FREENET_GIT_IS_DIRTY")
+        .as_deref()
+        .map(str::trim)
+    {
+        Ok("1") | Ok("true") => true,
+        Ok("0") | Ok("false") => false,
+        Ok("") | Err(_) => Command::new("git")
+            .args(["status", "--porcelain"])
+            .output()
+            .ok()
+            .map(|o| !o.stdout.is_empty())
+            .unwrap_or(false),
+        Ok(other) => panic!("FREENET_GIT_IS_DIRTY ({other}) must be 1, true, 0, false, or empty"),
+    };
     let dirty_suffix = if git_dirty { "-dirty" } else { "" };
     println!("cargo:rustc-env=GIT_DIRTY={dirty_suffix}");
 
@@ -146,6 +170,11 @@ fn emit_build_metadata() {
     let timestamp = now.format("%Y-%m-%dT%H:%M:%SZ").to_string();
     println!("cargo:rustc-env=BUILD_TIMESTAMP={timestamp}");
     println!("cargo:rerun-if-env-changed=SOURCE_DATE_EPOCH");
+    // Required: emitting any rerun-if-* directive makes Cargo honor only the
+    // declared set, so without these an unset var leaves stale provenance —
+    // and a stale GIT_DIRTY disables auto-update.
+    println!("cargo:rerun-if-env-changed=FREENET_GIT_COMMIT_HASH");
+    println!("cargo:rerun-if-env-changed=FREENET_GIT_IS_DIRTY");
 
     // Rebuild if git HEAD changes
     println!("cargo:rerun-if-changed=.git/HEAD");
--- a/crates/core/build.rs
+++ b/crates/core/build.rs
@@ -1,5 +1,5 @@
-use std::process::Command;
 use chrono::TimeZone;
+use std::process::Command;
 
 fn main() {
     // Emit build metadata for startup logging
@@ -111,22 +111,27 @@ fn read_min_compatible_from_cargo_toml() -> Option<String> {
 
 fn emit_build_metadata() {
     // Git commit hash
-    let git_hash = Command::new("git")
-        .args(["rev-parse", "--short=12", "HEAD"])
-        .output()
-        .ok()
-        .and_then(|o| String::from_utf8(o.stdout).ok())
-        .map(|s| s.trim().to_string())
-        .unwrap_or_else(|| "unknown".to_string());
+    let git_hash = std::env::var("FREENET_GIT_COMMIT_HASH").unwrap_or_else(|_| {
+        Command::new("git")
+            .args(["rev-parse", "--short=12", "HEAD"])
+            .output()
+            .ok()
+            .and_then(|o| String::from_utf8(o.stdout).ok())
+            .map(|s| s.trim().to_string())
+            .unwrap_or_else(|| "unknown".to_string())
+    });
     println!("cargo:rustc-env=GIT_COMMIT_HASH={git_hash}");
 
     // Git dirty flag
-    let git_dirty = Command::new("git")
-        .args(["status", "--porcelain"])
-        .output()
-        .ok()
-        .map(|o| !o.stdout.is_empty())
-        .unwrap_or(false);
+    let git_dirty = match std::env::var("FREENET_GIT_IS_DIRTY") {
+        Ok(v) if !v.is_empty() => true,
+        _ => Command::new("git")
+            .args(["status", "--porcelain"])
+            .output()
+            .ok()
+            .map(|o| !o.stdout.is_empty())
+            .unwrap_or(false),
+    };
     let dirty_suffix = if git_dirty { "-dirty" } else { "" };
     println!("cargo:rustc-env=GIT_DIRTY={dirty_suffix}");
 
--- a/crates/core/build.rs
+++ b/crates/core/build.rs
@@ -1,4 +1,5 @@
 use std::process::Command;
+use chrono::TimeZone;
 
 fn main() {
     // Emit build metadata for startup logging
@@ -130,8 +131,21 @@ fn emit_build_metadata() {
     println!("cargo:rustc-env=GIT_DIRTY={dirty_suffix}");
 
     // Build timestamp (ISO 8601)
-    let timestamp = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string();
+    let now = match std::env::var("SOURCE_DATE_EPOCH") {
+        Ok(val) => {
+            let epoch: i64 = val
+                .parse()
+                .unwrap_or_else(|_| panic!("SOURCE_DATE_EPOCH ({val}) is not a valid integer"));
+            chrono::Utc
+                .timestamp_opt(epoch, 0)
+                .single()
+                .unwrap_or_else(|| panic!("SOURCE_DATE_EPOCH ({val}) is not a valid timestamp"))
+        }
+        Err(_) => chrono::Utc::now(),
+    };
+    let timestamp = now.format("%Y-%m-%dT%H:%M:%SZ").to_string();
     println!("cargo:rustc-env=BUILD_TIMESTAMP={timestamp}");
+    println!("cargo:rerun-if-env-changed=SOURCE_DATE_EPOCH");
 
     // Rebuild if git HEAD changes
     println!("cargo:rerun-if-changed=.git/HEAD");
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -3609,4 +3609,60 @@ mod tests {
         let vague = GithubRateLimitedError { retry_after: None }.user_message();
         assert!(vague.contains("within the hour"), "{vague}");
     }
+
+    /// [`fn_body`] for a file with no test module. It requires one, to prove the
+    /// anchor did not fall through into the pin's own literals (#5102/#5103);
+    /// `build.rs` is a different file from this one, which gives that by
+    /// construction. The `\n}\n` bound still matters — without it a moved anchor
+    /// matches later in `build.rs` and the assertions pass vacuously.
+    fn build_rs_fn_body<'a>(src: &'a str, signature: &str) -> &'a str {
+        let at = src
+            .find(signature)
+            .unwrap_or_else(|| panic!("definition not found in build.rs: {signature}"));
+        assert!(
+            at == 0 || src.as_bytes()[at - 1] == b'\n',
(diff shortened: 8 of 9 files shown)


Skyb0rg007:
* `crates/core/build.rs:69-77` — The `FREENET_GIT_IS_DIRTY` override is silently ignored when Nix passes an explicit "false".`package.nix` sets `env.FREENET_GIT_IS_DIRTY = gitDirty;` (a Nix bool); Nix's derivation-env coercion turns `false` into an **empty string**, not an absent var. The match `Ok(v) if !v.is_empty() => true` only recognizes the non-empty ("dirty") case — an empty-but-set value falls through to the same `_` arm as "unset", which then shells out to `git status --porcelain`. In the sandboxed Nix build this happens to still resolve to `false` only because `git`/`.git` aren't available there (an accidental result, not a designed one); outside that specific sandbox context (e.g. impure builds, or reusing this code path elsewhere) an explicit `false` would no longer be honored and live repo state would leak in instead. Since the entire point of this change is deterministic/reproducible build metadata, the override should be read as a tri-state (unset → fallback; `"1"` → true; `""` → explicit false) rather than collapsing "unset" and "" into the same branch.

This behavior is intentional in order to let users disable the environment variable-based dirty check by simply setting it to the empty string. If this is odd, I can make the build.rs logic check for "1" and "0" explicitly, and modify the nix derivation to use `if gitDirty then "1" else "0"`.

* The `fix(nix): use most recent tag, not main branch` commit changes `freenet-autoupdate`'s update-selection logic (now resolves via `gh release view --json tagName` instead of running `main`), but no test accompanies it that would catch a regression to the old "run main" behavior. Per the review criteria for `fix:` PRs, a regression test is expected. (Caveat: this is Nix/shell logic with no existing shell-test harness in this repo, so a traditional `#[test]` isn't a natural fit here — flagging per the letter of the rule, but this may warrant a review-override rather than blocking.)

I don't think a whole new test infrastructure is required for something that couldn't cause any actual problems. The previous update script didn't have a test either.

 * `crates/core/build.rs` — the new `SOURCE_DATE_EPOCH` parse/fallback branch (valid epoch, invalid epoch, unset) has no test coverage exercising its parsing logic.
 
 This code is taken verbatim from the https://reproducible-builds.org/docs/source-date-epoch/ site, so I don't think it needs a test.

iduartgomez (commented):
## Comprehensive PR Review: #5073

### Summary
- **PR Title:** build: update and fix the nix flake
- **Type:** build (with one `fix(nix):` commit)
- **CI Status:** partial — `Fmt`, `Conventional Commits`, `Rule Lint`, `Claude Rule Review`, `CLA`, `Snyk`, `macOS bundle-swap E2E` pass; `Clippy`, `Unit & Integration`, `Simulation`, `NAT Validation`, `Windows Check`, `Windows/macOS Service Unit` still pending at review time
- **Linked Issues:** none
- **Review tier:** Full (touches release/CI config — dependabot, the auto-update wrapper, and build-provenance metadata)
- **Reviewers run:** code-first, testing, skeptical, big-picture + external model (`codex review --base origin/main`)

Thanks for this — the flake genuinely needed the attention, and most of the execution is good: `writeShellApplication` (which gets you shellcheck at build time), `mainProgram`, `cargoLock.lockFile`, `meta` cleanup, and `SOURCE_DATE_EPOCH` support are all the right calls, and pinning the updater to release tags instead of arbitrary `main` commits is a real improvement. The findings below concentrate on the paths outside the happy Nix case.

---

### Code-First Analysis
**Independent Understanding:** `build.rs` gains env-var overrides (`FREENET_GIT_COMMIT_HASH`, `FREENET_GIT_IS_DIRTY`, `SOURCE_DATE_EPOCH`) that take precedence over shelling out to `git`; `package.nix` plumbs the first two from flake-level `self.rev`/`self.dirtyRev`; `flake.nix` rewrites `freenet-autoupdate` to resolve the latest release tag via `gh` and adds a devShell; dependabot gains a `nix` ecosystem entry.

**Stated Intent:** the 5-point summary in the PR description.

**Alignment:** matches — all five items are present in the diff.

**Gaps:** two behavior changes ride along undescribed:
- `trap "" INT` is removed from `freenet-autoupdate` without a note.
- The non-42 path changes from `break` (loop exits, script then exits 0 — the child's failure code was swallowed) to `exit $exit_code`. This is a genuine bug fix in exit-code propagation, but it's also a silent breaking change for supervisors. See SF4.

---

### Testing Assessment
**Coverage Level:** insufficient at the CI level — but your two specific pushbacks in-thread are both reasonable and I'm not asking for those tests.

| Unit | N/A | `build.rs` is not part of any `cargo test` target |
| Integration | ❌ | No CI job evaluates or builds the flake at all |
| Simulation | N/A | — |
| E2E | N/A | — |

**On your replies:**
- The `SOURCE_DATE_EPOCH` block fails loudly via `panic!` rather than silently miscompiling — the right failure mode for a build script. Agreed, no dedicated test needed.
- There is no shellcheck/bats harness anywhere in this repo for embedded shell, and building one for a single wrapper would be disproportionate. Agreed. (For the record, I extracted the generated script with its injected `set -euo pipefail` and ran `shellcheck -x` — clean, including the unquoted `[ $exit_code -eq 42 ]`.)

**Missing coverage — the real gap:** nothing in CI evaluates the flake. `grep -rn "nix flake\|nix build" .github/workflows/` finds no such job. Every claim this PR makes — that `package.nix` evaluates, that the derivation builds, that the `dirtyRev` branch works, that `devShells.default` opens — is unverified. And this same diff adds weekly `nixpkgs-unstable` dependabot bumps (`.github/dependabot.yml:15-23`), which will now land with zero build signal. A `paths:`-filtered job on `flake.nix`/`flake.lock`/`package.nix`/`crates/core/build.rs` running `nix flake check` + `nix build .#freenet` is the prerequisite for that dependabot entry to be net-positive. Reasonable as a fast follow-up rather than a hard blocker on this PR.

---

### Skeptical Findings
**Risk Level:** medium-high

| `freenet-autoupdate` hard-requires `gh` auth; any `gh` blip is fatal | high | `flake.nix:31-33`, `:65` | MF1 |
| `FREENET_GIT_IS_DIRTY=false` means *dirty* → silently disables auto-update | high | `crates/core/build.rs:126-134` | MF2 |
| Missing `rerun-if-env-changed` for both new git vars → stale provenance | high | `crates/core/build.rs:152-153` | MF3 |
| `self.dirtyRev` dereferenced unguarded → hard eval failure | medium | `flake.nix:24-25` | SF1 |
| Re-introduces the `api.github.com` dependency #5102 removed | medium | `flake.nix:31-32` | SF2 |
| No sleep/backoff, and bypasses the `MAX_UPDATE_FAILURES` loop breaker | medium | `flake.nix:29-45` | SF3 |
| Exit-code semantics changed from "always 0" to propagating | medium | `flake.nix:46` | SF4 |
| devShell toolchain won't match the `1.94.0` pin | medium | `flake.nix:66-78` | SF5 |
| `.gitignore` `result` is unanchored | low | `.gitignore:84` | C5 |
| `BUILD_TIMESTAMP` reads `1980-01-01` on Nix builds | low | `crates/core/build.rs:139-152` | C6 |

**MF1 — `gh release view` needs auth even for a public repo, and `set -e` makes that immediately fatal.** (`flake.nix:31-33`)

`writeShellApplication` injects `set -euo pipefail`. I verified both halves:

```
$ env -u GH_TOKEN -u GITHUB_TOKEN GH_CONFIG_DIR=/tmp/empty \
    gh release view --repo freenet/freenet-core --json tagName --jq .tagName
To get started with GitHub CLI, please run:  gh auth login
Alternatively, populate the GH_TOKEN environment variable with a GitHub API authentication token.

$ bash -c 'set -euo pipefail; v="$(false)"; echo "unreachable"'; echo "exit=$?"
exit=1
```

`runtimeInputs` supplies the `gh` binary but not a credential. Since `packages.default = freenet-autoupdate` (`flake.nix:65`), plain `nix run github:freenet/freenet-core` — the advertised entrypoint — now dies before ever launching a node, for every user who hasn't run `gh auth login`. The old script needed no credentials at all.

The same construct has a second effect even for authenticated users: any transient `gh` failure (network blip, GitHub 5xx, expired token, secondary rate limit) permanently kills the supervisor loop. Previously the loop was fully resilient to GitHub being unreachable. For a supervisor, exiting permanently on a transient upstream failure is the wrong failure mode.

**MF2 — `FREENET_GIT_IS_DIRTY` truthiness is "any non-empty string", so `=false` marks the build dirty and silently disables auto-update.** (`crates/core/build.rs:126-134`)

```rust
let git_dirty = match std::env::var("FREENET_GIT_IS_DIRTY") {
    Ok(v) if !v.is_empty() => true,
    _ => /* fall back to `git status --porcelain` */
};
```

`false`, `0`, `no`, `off` all mean **dirty**. `GIT_DIRTY` is the auto-update kill switch — `auto_update_is_disabled(git_dirty, flag) = !git_dirty.is_empty() || flag` (`crates/core/src/bin/freenet.rs:205-207`, whose doc comment stresses the default is load-bearing). And `crates/core/src/config.rs:227-231` explains why `--disable-auto-update` deliberately has **no** env binding:

> *"a truthy env value is easy to leave set by accident, and silently disabling auto-update fleet-wide is the exact failure this must avoid."*

This PR creates exactly that env binding, at build time, with inverted-value semantics for the most natural way to write "false". Your in-thread reasoning holds for the Nix sandbox specifically, and I'd accept that narrow argument — but this is the broader hazard, and the explicit `"1"`/`"0"` form you already offered fixes it. Please take that option: parse `"1"|"true"` → dirty, `"0"|"false"|""` → clean, anything else → `panic!`, and set `FREENET_GIT_IS_DIRTY = if gitDirty then "1" else "0"` in `package.nix`. Worth a comment on the match arm too, since the semantics already caused confusion in this thread.

Related asymmetry while you're there: `FREENET_GIT_COMMIT_HASH=""` yields `Ok("")` → an *empty* commit hash (logged at `freenet.rs:168`), not the git fallback and not `"unknown"`. Two new vars, two different empty-string rules.

**MF3 — `cargo:rerun-if-env-changed` is registered for `SOURCE_DATE_EPOCH` but not the two git vars.** (`crates/core/build.rs:152-153`)

Emitting any `cargo:rerun-if-*` directive switches off Cargo's "always rerun" default and honors only the declared set. Concretely:

1. `FREENET_GIT_IS_DIRTY=1 cargo build` → binary baked with `GIT_DIRTY="-dirty"`.
2. Edit any `src/**.rs` (touches neither `.git/HEAD` nor `.git/index`, the only two `rerun-if-changed` paths at lines 156-157).
3. `cargo build` with the var now unset → crate recompiles, build script does **not** rerun, `GIT_DIRTY` stays `"-dirty"`.

Result: a release binary that silently never auto-updates and reports a fabricated commit hash. `.claude/rules/deployment.md` ("WHEN implementing auto-update") makes the `GIT_DIRTY` gate mandatory, so this is a correctness issue, not a cosmetic one. Unlikely inside the Nix sandbox (always fresh), but this PR explicitly advertises the vars for non-Nix packagers, where shared target dirs are normal.

```rust
println!("cargo:rerun-if-env-changed=FREENET_GIT_COMMIT_HASH");
println!("cargo:rerun-if-env-changed=FREENET_GIT_IS_DIRTY");
```

(Found independently by two reviewers including the external model.)

**SF1 — `self.dirtyRev` is dereferenced unguarded.** (`flake.nix:20-25`)

```nix
gitDirty = !(self ? rev);
gitCommitHash =
  if self ? rev then lib.substring 0 12 self.rev
  else lib.substring 0 12 (lib.removeSuffix "-dirty" self.dirtyRev);
```

`self ? rev` is guarded; `self.dirtyRev` is not. `dirtyRev` exists only when the source is a git tree that is dirty *and* has at least one commit. It's absent — and evaluation dies with `attribute 'dirtyRev' missing` — for `nix build path:/path/to/checkout`, tarball flakes, vendored source drops, and a freshly `git init`'d tree. That last set includes the "download a GitHub archive tarball" case item 5 of your description is meant to enable. And because `devShells.default` uses `inputsFrom = [ self.packages.${system}.freenet ]` (`flake.nix:67`), `nix develop` forces the same expression.

```nix
gitCommitHash =
  if self ? rev then lib.substring 0 12 self.rev
  else if self ? dirtyRev then lib.substring 0 12 (lib.removeSuffix "-dirty" self.dirtyRev)
  else "unknown";
```

Note also that `gitDirty = !(self ? rev)` treats "no VCS metadata" as "dirty", which per MF2 flips auto-update off for any non-git source drop. (Also flagged by the external model.)

**SF2 — the `gh` call re-introduces the `api.github.com` dependency that #5102 deliberately removed.** (`flake.nix:31-32`)

`crates/core/src/bin/commands/auto_update.rs:123-141` carries an unusually emphatic comment:

> *"This is deliberately **not** `api.github.com`. GitHub's REST API allows only 60 unauthenticated requests per hour per source IP, and that budget is collective… Do NOT 'simplify' this back to `api.github.com`: the JSON body is not needed to learn the tag, and paying REST quota for it is precisely the bug."*

`gh release view --json tagName` is precisely that REST call. Being authenticated (5000/hr) means #5102's exact exhaustion doesn't reproduce — but the wrapper now *requires* a token to perform the one lookup the core went out of its way to make token-free, and it resolves the version by a different mechanism than the node itself does.

This is also the clean fix for MF1. Use the same auth-free redirect the core uses, and make failure non-fatal:

```sh
vsn="$(curl -fsSLI -o /dev/null -w '%{url_effective}' \
       https://github.com/freenet/freenet-core/releases/latest \
       | sed 's|.*/releases/tag/||')" || vsn=""
if [ -z "$vsn" ]; then
    echo "Could not resolve latest release; retrying in 60s." >&2
    sleep 60
    continue
fi
```

No `gh`, no auth, no REST quota, matches the node's own resolver, and a blip no longer kills the daemon.

**SF3 — no sleep/backoff, and the wrapper bypasses the `MAX_UPDATE_FAILURES` loop breaker.** (`flake.nix:29-45`)

`crates/core/src/bin/commands/update.rs` defines `MAX_UPDATE_FAILURES: u32 = 3` specifically so the exit-42 restart loop terminates (the #3934 runaway-restart protection). That counter is maintained around the `freenet update` subprocess dispatch; the Nix wrapper never invokes it — it just re-resolves a tag and re-execs. Combined with no `sleep`, no backoff and no jitter, a node that misdetects an update spins `gh` + `nix run` + exit-42 as fast as the loop turns. That's the backoff-without-jitter class in `.claude/rules/bug-prevention-patterns.md`.

Worth noting this got *harder* to self-heal, not easier: under the old `main`-tracking behavior a bad build could be fixed by pushing to `main`; now only cutting a new release breaks the loop. At minimum add a `sleep` on the 42 path.

**SF4 — exit-code semantics changed from "always 0" to "propagate the child's code".** (`flake.nix:46`)

`break` → `exit $exit_code`. Previously the wrapper always exited 0. Now Ctrl-C yields 130, a crash yields 1, `EXIT_CODE_ALREADY_UP_TO_DATE = 2` yields 2, and "already running" yields 43. Anyone running this under systemd with `Restart=on-failure` flips from "clean stop" to "restart loop" for all of those, without having been told to add a matching `SuccessExitStatus=`. The new behavior is arguably more correct — but it should be called out in the description, and 43 (already running) probably still ought to map to 0.

**SF5 — the devShell toolchain won't match the repo's pin.** (`flake.nix:66-78`)

`rust-toolchain.toml` pins `channel = "1.94.0"`; the devShell takes `rustc`/`cargo` from `inputsFrom` (nixpkgs-unstable's default) plus `pkgs.clippy`/`pkgs.rustfmt`. AGENTS.md says `cargo fmt && cargo clippy -- -D warnings` "must match CI" — from this shell it won't, since lint sets and rustfmt output drift between versions. Either pin the Rust toolchain to match, or document the devShell as build-only. Separately, `inputsFrom` on a `buildRustPackage` drags in `cargoSetupHook`, which sets `CARGO_HOME`/vendor-dir variables that commonly break a plain `cargo build` against crates.io — worth actually opening the shell and running `cargo clippy` before merging.

**Checked and found NOT to be problems** (recorded so they don't get re-raised):
- `package-ecosystem: "nix"` is valid — GitHub shipped Dependabot Nix support in April 2026; `groups` and `directory: "/"` are correct here.
- `meta.donationPage` is a real attribute (`check-meta.nix` declares `donationPage = str;`).
- `env` accepting a Nix bool is legal (`make-derivation.nix`: "derivation, string, boolean or integer"), and `false` → `""` as you describe.
- The generated script passes `shellcheck`, so `writeShellApplication` won't fail the build.
- `fromTOML` → `builtins.fromTOML` is cosmetic, not a fix — `fromTOML` is in the global scope, so the old form evaluated fine. The v0.2.120 flake is evaluable, so the wrapper isn't dead on arrival.
- Removing `trap "" INT` is largely benign: bash defers default-disposition SIGINT until the foreground child completes, and `nix run` execs the target so `freenet` still gets its graceful shutdown. The old script also stopped the loop on Ctrl-C. The only real delta is the exit code (SF4) — but please confirm the removal was deliberate.
- `chrono` is already a legitimate build-dependency, and `TimeZone::timestamp_opt` isn't deprecated, so clippy won't trip on it.
- `inherit (nixpkgs) lib;` is correct (system-independent, exposed at the flake's top level).

---

### Big Picture Assessment
**Goal Alignment:** yes. Nix flake support (#2954, #3014) was added without a linked issue too, so no-issue is consistent with precedent, and the `build.rs` changes are inert unless the env vars are set — non-Nix builds are behaviorally unchanged.

**Anti-Patterns Detected:** SF3 (missing backoff) is the one repo-catalogued pattern hit. No tests removed, no assertions weakened, no error-swallowing; the `panic!`s are appropriately fail-loud.

**Removed Code Concerns:** the `trap "" INT` removal — benign on analysis, but please confirm it was intentional.

**Scope Assessment:** focused enough. Five items, but all nix-flake-adjacent and already split into clean per-topic commits. I would not ask for a split.

**One correction to the PR description (item 3):** the claim that reproducibility lets "anyone verify that the binaries on the GitHub releases page were not tampered with" does not hold today. Release artifacts are built by `.github/workflows/cross-compile.yml` with plain `cargo build --release` — I confirmed `SOURCE_DATE_EPOCH` appears nowhere under `.github/`, and nix isn't involved in that pipeline. What this actually achieves is that two people building via `nix build` from the same source get bit-identical output — real and worthwhile, just not GitHub-release verifiability. Worth rewording, or wiring `SOURCE_DATE_EPOCH` into `cross-compile.yml` in a follow-up (your description already flags this as easy-but-not-done).

**On trust model (context, not a request):** the nix path builds from tagged source over HTTPS and never touches the signed `SHA256SUMS.txt`/`.sig` that `update.rs` verifies against the baked-in `FREENET_RELEASE_PUBKEY`. That divergence predates this PR, and pinning to release tags strictly improves it. Noted for maintainer awareness only.

---

### Documentation
- Code docs: incomplete — `FREENET_GIT_IS_DIRTY`'s semantics need a comment at `crates/core/build.rs:126` (see MF2).
- Architecture docs: n/a.
- User docs: needs-update — `nix develop`, `nix build`, and `flake.nix` appear nowhere in `README.md`, `CONTRIBUTING.md`, or `docs/` (verified by grep). The new devShell is item 4's whole point but is undiscoverable without reading `flake.nix`.

---

### Recommendations

#### Must Fix (Blocking)
1. **MF1** — `freenet-autoupdate` requires `gh` auth and dies on any `gh` failure (`flake.nix:31-33`). This breaks the flake's default package for unauthenticated users. Fixing via SF2's auth-free redirect resolves both at once; make a failed lookup retry rather than exit.
2. **MF2** — make `FREENET_GIT_IS_DIRTY` an explicit tri-state (`crates/core/build.rs:126-134`). As written, `=false` disables auto-update — the precise hazard `config.rs:227-231` documents. Take the `"1"`/`"0"` option you proposed.
3. **MF3** — add `cargo:rerun-if-env-changed` for `FREENET_GIT_COMMIT_HASH` and `FREENET_GIT_IS_DIRTY` (`crates/core/build.rs:152-153`); stale `GIT_DIRTY` can flip the auto-update gate.

#### Should Fix (Important)
1. **SF1** — guard `self.dirtyRev` (`flake.nix:24-25`); currently a hard eval failure for non-git trees, including the tarball case item 5 advertises.
2. **SF2** — use the `releases/latest` redirect rather than `gh`/`api.github.com`, per the explicit warning in `auto_update.rs:123-141`.
3. **SF3** — add a `sleep` (and ideally a consecutive-restart cap) to the loop; the wrapper has no equivalent of `MAX_UPDATE_FAILURES`.
4. **SF4** — call out the `break` → `exit $exit_code` change in the description, and confirm the `trap "" INT` removal was intentional. Consider mapping 43 to 0.
5. **SF5** — align the devShell toolchain with `rust-toolchain.toml`'s 1.94.0, or document it as build-only.
6. Reword the reproducible-builds claim (see Big Picture).

#### Consider (Suggestions)
1. Validate `FREENET_GIT_COMMIT_HASH` (`is_ascii_hexdigit`, length ≤ 40) before interpolating it into a `cargo:` directive — an embedded newline would emit additional directives. Low real-world risk since it requires build-env control, but it's a two-line guard.
2. Export `FREENET_SUPERVISED=1` in the wrapper. Without it, `supervisor_status()` returns `Unsupervised` and `freenet.rs:1290-1308` logs an error claiming the process "will now EXIT WITHOUT UPDATING and will not be restarted" — false under this wrapper, and operators will chase it.
3. Follow-up CI job running `nix flake check` + `nix build .#freenet`, `paths:`-filtered. Prerequisite for the dependabot entry to be net-positive.
4. Use `--extra-experimental-features` rather than `--option experimental-features` (`flake.nix:39`) — the latter replaces the user's `nix.conf` value instead of extending it.
5. Overlay convention: `flake.nix:18` is `_: pkgs:`, so `pkgs` is `prev` — `pkgs.callPackage` resolves `rustPlatform` pre-overlay, so downstream Rust overrides don't compose. `final: prev: { freenet = final.callPackage …; }` is the usual form.
6. Anchor the gitignore entries as `/result` and `/result-*`. Verified today: `git check-ignore -v crates/core/src/result` → `.gitignore:84:result`. Nothing collides now; future footgun only.
7. `BUILD_TIMESTAMP` becomes `1980-01-01T00:00:00Z` on Nix builds (nixpkgs stdenv sets `SOURCE_DATE_EPOCH=315532800`). Correct for reproducibility but uninformative in bug reports — passing `self.lastModified` would be both reproducible and meaningful.

#### Questions
1. Was `nix build .#freenet` run to completion on this branch? `package.nix` declares no `nativeBuildInputs`/`buildInputs`, yet `Cargo.lock` carries `openssl-sys` (4570), `pkg-config` (4959), and `libsqlite3-sys` (3815). If any resolve into the `freenet` binary's feature graph, the derivation needs `pkg-config` + `openssl`. With no CI job there's no evidence either way.
2. Was `nix develop` opened and `cargo clippy -- -D warnings` run inside it? See SF5.
3. What's the intended `gitDirty` for a clean release build? Today a clean `nix run github:.../vX.Y.Z#freenet` gives `GIT_DIRTY=""` → in-binary auto-update **enabled** → the node exits 42 and relies on this wrapper. Intentional, or should the Nix path disable in-binary auto-update and let the wrapper own version selection exclusively? The current arrangement has both the node and the wrapper independently resolving "latest", which is the setup for SF3's loop.

---

### Verdict
**HEAD SHA reviewed:** 15372fe91

The direction is right and nothing here is hard to fix. The three Must-Fix items are the `gh`-auth regression (breaks the flake's default entrypoint for unauthenticated users) and the two build-metadata issues that can silently ship a release binary with auto-update disabled. MF3 and SF1 were each found independently by more than one reviewer, so I'd treat them as solid.

Because blocking findings are involved, a fresh review pass on the updated HEAD is warranted before merge — specifically re-confirming MF1, MF2, MF3, SF1.

Also note unit/integration CI was still pending when this review ran; please confirm green on the final HEAD.

[AI-assisted - Claude]

Skyb0rg007:
@iduartgomez please try to avoid pasting AI-generated walls of text, as it makes human review of the content very difficult. If you want to make a ton of AI-generated comments, use GitHub's comment system so that I can respond to each point in a targeted manner, rather than using a single comment chain. I'd like to respond to each point individually.

Skyb0rg007:
Because I didn't get a response I'm just dumping my claude output verbatim.

# Reply

All findings addressed. Verified locally with nix 2.34.8.

**MF1 / SF2 — `gh` auth and `api.github.com`.** Fixed together: the wrapper now
resolves the tag with the same auth-free `releases/latest` redirect the node uses
(`GITHUB_LATEST_REDIRECT_URL`), `pkgs.gh` → `pkgs.curl`, and a failed lookup logs,
sleeps 60s and retries instead of dying. Verified with no credentials in the
environment (`v0.2.127`), and verified that the failure path survives
`set -euo pipefail` rather than exiting.

**MF2 — `FREENET_GIT_IS_DIRTY`.** Took the explicit option: `"1"`/`"true"` → dirty,
`"0"`/`"false"` → clean, unrecognised → `panic!`. `package.nix` now sets `"1"`/`"0"`
rather than relying on Nix's `false → ""` coercion. One deviation from your
suggestion: empty still means "no override, ask `git`" rather than "clean". A
stray empty value in a real checkout should fall back to detection, not assert
cleanliness and re-enable auto-update on a dirty dev build. `FREENET_GIT_COMMIT_HASH`
now follows the same empty-means-fallback rule instead of baking in an empty hash.

**MF3 — `rerun-if-env-changed`.** Added for both vars.

MF2 and MF3 are pinned by a new test,
`build_script_git_dirty_override_cannot_silently_disable_auto_update`. It is a
cross-file scrape (`auto_update.rs` reading `build.rs`), bounded to
`emit_build_metadata`. Mutation-tested three ways: reverting the `"0"`/`"false"`
arm, deleting a `rerun-if-env-changed` line, and renaming the anchored function —
all three make it fail, the last with `definition not found in build.rs` rather
than a vacuous pass.

**SF1 — `self.dirtyRev`.** Guarded, with `"unknown"` as the third case, and
`gitDirty` is now `self ? dirtyRev` so a source drop is not treated as dirty.
Mutation-tested: the previous flake dies with `attribute 'dirtyRev' missing` on a
`git archive` tarball extracted to a non-git directory; the new one evaluates to
`FREENET_GIT_COMMIT_HASH=unknown`, `FREENET_GIT_IS_DIRTY=0`.

**SF3 — backoff.** Added a `sleep` on the 42 path plus a consecutive-restart cap of
3 against an unchanged tag, mirroring `MAX_UPDATE_FAILURES`.

**SF4 — exit codes / `trap`.** Both were deliberate; documented in the description
rather than changed, including the `SuccessExitStatus=` note for systemd users.

**SF5 — devShell toolchain.** Added `rust-overlay` and pinned to
`rust-toolchain.toml`. `nix develop --command rustc --version` now reports 1.94.0
(nixpkgs-unstable currently ships 1.97.0). `clippy`/`rustfmt` come from the
toolchain file's `components` rather than nixpkgs.

**Consider items.** Taken: hash validation (in both `build.rs` and `package.nix`),
`FREENET_SUPERVISED=1`, the CI job, `--extra-experimental-features`,
`final: prev:`, anchored `/result`, and `SOURCE_DATE_EPOCH` from
`self.lastModified` (`Build timestamp: 2026-08-17T03:17:43Z` rather than 1980).

## Your three questions

**1. Was `nix build .#freenet` run to completion?** It is now, and it needs no
`pkg-config`/`openssl`: `cargoBuildFlags` restricts the build to
`--package=freenet --bin=freenet`, and none of `openssl-sys`, `pkg-config` or
`libsqlite3-sys` resolve into that binary's feature graph.

**2. Was `cargo clippy -- -D warnings` run in `nix develop`?** Yes — and it found
something. `cargoSetupHook` does not break plain cargo (`cargo fmt --check` and
`cargo metadata` are fine), but `cargo clippy --workspace --all-targets` failed to
compile: `release-agent` pulls in `openssl-sys`, which the shell could not satisfy.
Fixed by adding `pkg-config` + `openssl` to the devShell. Clippy is now clean
workspace-wide with `-D warnings`. The CI job asserts `pkg-config --exists openssl`
directly, since `cargo metadata` runs no build scripts and cannot see this class.

**3. Intended `gitDirty` for a clean release build?** Auto-update stays enabled, and
I think that is right: the node and the wrapper now resolve "latest" through the
same redirect rather than two different mechanisms, so they agree on the target,
and the wrapper's restart cap bounds the pathological case you describe. Happy to
flip the Nix path to `--disable-auto-update` and let the wrapper own version
selection exclusively if you would rather have one authority.

## Also

The reproducibility claim in item 3 was overstated and is reworded — it does not
give GitHub-release verifiability, since `cross-compile.yml` builds those with a
plain `cargo build --release`. What it does give is checked:
`nix build .#freenet --rebuild` rebuilds from scratch and reports the outputs
identical.

`CONTRIBUTING.md` gained a "Building with Nix" section covering `nix develop`,
`nix build`, the toolchain pin, and the two provenance env vars for non-Nix
packagers.

sanity:
@Skyb0rg007 — sorry for the week of silence after you addressed everything. Point taken on the wall-of-text; keeping this short.

I read the current state. The build.rs tri-state, the `self ? dirtyRev` guard, the toolchain pin via rust-overlay, and the CI job all look right, and the `--rebuild` check is the honest version of the reproducibility claim.

Two things stand between this and merge:

**1. Jitter — one line, and I do want it.** Both retry sleeps in `freenet-autoupdate` are fixed (`sleep 60` on the lookup failure, `sleep $((60 * (failures + 1)))` on the 42 path). Every instance polls the same `releases/latest` redirect and gets exit 42 at roughly the same wall-clock moment on a new release, so they retry in lockstep. The node's own poller already jitters for this reason (`UPDATE_REPOLL_JITTER_FRACTION = 0.25`). Something like:

```sh
jittered() { printf '%s' $(( $1 * 8 / 10 + RANDOM % ($1 * 4 / 10 + 1) )); }
```

or any ±20% equivalent — `writeShellApplication` uses bash, so `$RANDOM` is available.

**2. The other rule-review warning is ours, not yours.** The "`fix:` commit with no regression test" one — you're right that a shell/Nix path has no natural `#[test]` here and that the old wrapper had no test either. I'm applying `review-override` for it rather than asking you to invent a harness.

**On CI:** the runs on this branch had been sitting unapproved since 2026-08-22 — outside-contributor workflows need a manual approval and nobody gave it. I've approved them, so you should see the full suite (including your new `Nix` job) run now.

**On your question 3** (auto-update stays enabled for a clean Nix build): agreed, leave it as is. The node and the wrapper resolving "latest" through the same redirect is the property that matters, and your restart cap bounds the bad case.

Push the jitter change and I'll take it from there.

[AI-assisted - Claude]

sanity:
Justification for the `review-override` label, so it isn't a bare bypass: the warning is the `fix:`-needs-a-regression-test rule firing on a Nix/shell change that has no Rust surface a `#[test]` could cover, and the repo has no shell-test harness for `flake.nix`. The jitter warning is *not* covered by the override — that one's still the ask.

[AI-assisted - Claude]

Skyb0rg007:
Jitter added and force-pushed (squashed into `fix(nix): run the latest release tag, not main`, and rebased onto the current branch tip).

Both sleeps now go through one helper, ±20% either side, matching `UPDATE_REPOLL_JITTER_FRACTION`:

```sh
jittered() {
    printf '%s' "$(( $1 * 8 / 10 + RANDOM % ($1 * 4 / 10 + 1) ))"
}
```

The lookup-failure path uses `jittered 60`; the exit-42 path uses `jittered $((60 * (failures + 1)))`. Both now print the delay they actually picked instead of claiming a fixed 60s.

Checked the bounds empirically over 500 draws per base — 60 → 48..72, 120 → 96..144, 180 → 144..216, so it's a real ±20% spread and never zero or negative. `nix build .#freenet-autoupdate` passes, which includes the shellcheck pass `writeShellApplication` runs, and `nixfmt --check` is clean.

Thanks for the `review-override` on the other warning, and for approving the workflow runs.

package.nix
```diff
@@ -1,28 +1,47 @@
...
+  version = (builtins.fromTOML (builtins.readFile ./crates/core/Cargo.toml)).package.version;
 
   src = ./.;
 
-  cargoLock = {
-    lockFile = ./Cargo.lock;
+  cargoLock.lockFile = ./Cargo.lock;
+
+  env = {
+    FREENET_GIT_COMMIT_HASH = gitCommitHash;
```
sanity, line 25: **Blocking.** This passes `"unknown"` through unconditionally, but `build.rs` validates it as hex and panics:

```
FREENET_GIT_COMMIT_HASH (unknown) must be <= 40 hex characters
```

`"unknown"` is non-empty so it reaches the validator, length 7 passes the `<= 40` check, and then `chars().all(is_ascii_hexdigit)` fails — none of `u n k n o w n` is a hex digit. Hard build failure for `nix build path:/some/tarball#freenet`, a `tarball+https://` input, or a plain `pkgs.callPackage ./package.nix { }` (the nixpkgs-upstreaming path the new `mainProgram`/`description` are clearly aiming at).

Note the assert just above deliberately whitelists `"unknown"` while `build.rs` rejects it — the two validators disagree, and the assert's own message describes a condition its first disjunct exists to violate.

Suggest omitting the attr instead of inventing a sentinel, mirroring what you already do for `sourceDateEpoch`. With it unset, `build.rs` falls to its own git probe, which yields `"unknown"` in the sandbox anyway — same value, no panic:

```nix
  env = {
    # Not a Nix bool: `false` coerces to "", which build.rs reads as "no override".
    FREENET_GIT_IS_DIRTY = if gitDirty then "1" else "0";
  }
  # Omitted rather than passed as "unknown": build.rs validates this var as hex
  # and panics on anything else.
  // lib.optionalAttrs (gitCommitHash != "unknown") {
    FREENET_GIT_COMMIT_HASH = gitCommitHash;
  }
  // lib.optionalAttrs (sourceDateEpoch != null) {
    SOURCE_DATE_EPOCH = sourceDateEpoch;
  };
```

flake.nix
```diff
@@ -2,48 +2,164 @@
...
+                    continue
+                fi
+
+                echo "Running freenet $vsn"
+                exit_code=0
+                nix --extra-experimental-features 'nix-command flakes' \
+                  run "github:freenet/freenet-core/$vsn#freenet" || exit_code=$?
+
+                if [ "$exit_code" -eq 42 ]; then
+                    if [ "$vsn" = "$last_vsn" ]; then
```
sanity, line 92: **Should fix before merge.** This counts every exit 42 for the lifetime of the wrapper, but 42 doesn't only mean "update available":

- `EXIT_CODE_UPDATE_NEEDED = 42` (auto_update.rs:30)
- `FATAL_LISTENER_EXIT_CODE = 42` (p2p_impl.rs:48)

The distinct fast-crash code 45 is emitted *only* when `enable_fast_crash_exit_code()` is turned on, and that is gated on a systemd marker env var (freenet.rs:341) which this wrapper doesn't set. So under this wrapper an ordinary fatal exit — including a boot wedge and the redb-poison restart at freenet.rs:352 — also arrives as 42.

Concretely: a gateway runs 8 hours, hits a fatal listener exit → `failures=1`. Runs 8 more hours, same → 2. Once more → `exit 1`, and the peer is off the network for good, with a message claiming it "requested an update 3 times" when it never requested anything.

That inverts the design: p2p_impl.rs documents 42 as burst-**exempt** (systemd whitelists it via `SuccessExitStatus=42 43`) precisely so it does *not* count toward a failure limit, and systemd's limiter is a rate, not a total. The node's own `MAX_UPDATE_FAILURES` lockout also keeps the node *running* while disabling auto-update — the opposite of stopping it.

Making the cap a rate keeps the bound you were asked for while fixing this. Add near the other counters:

```sh
# Only a run shorter than this counts as a failure. Matches
# MIN_HEALTHY_UPTIME_FOR_UPDATE_EXIT (crates/core/src/node/p2p_impl.rs).
min_healthy_uptime=60
```

time the run:

```sh
started=$SECONDS
nix --extra-experimental-features 'nix-command flakes' \
  run "github:freenet/freenet-core/$vsn#freenet" || exit_code=$?
ran_for=$((SECONDS - started))
```

and gate the increment on it:

```sh
if [ "$vsn" = "$last_vsn" ] && [ "$ran_for" -lt "$min_healthy_uptime" ]; then
    failures=$((failures + 1))
else
    failures=0
fi
```

I checked this against six sequences: 3 crashes-after-8h now survives; a same-tag instant-death loop still gives up on the 4th; a tag change resets; one healthy run mid-loop resets; and the 59s/60s boundary behaves. shellcheck clean at default severity (what `writeShellApplication` runs), and `$SECONDS` avoids adding a `date` dependency.

flake.nix
```diff
@@ -2,48 +2,164 @@
...
+            jittered() {
+                printf '%s' "$(( $1 * 8 / 10 + RANDOM % ($1 * 4 / 10 + 1) ))"
+            }
+
+            while true; do
+                # The auth-free redirect the node itself uses
+                # (GITHUB_LATEST_REDIRECT_URL in .../commands/auto_update.rs).
+                # Not api.github.com: that spends REST quota shared per source
+                # IP (#5102), and needs a token this wrapper should not require.
+                vsn="$(curl -fsSL -o /dev/null -w '%{url_effective}' \
```
sanity, line 73: **Should fix before merge.** No `--connect-timeout`/`--max-time`, so a blackholed :443 (a dropped-SYN network, a captive portal) stalls the supervisor indefinitely — no log line, no node, no retry. The node's own probe of this exact URL is bounded at 10s with a long rationale (`PROBE_CHAIN_TIMEOUT`, auto_update.rs:732).

```sh
vsn="$(curl -fsSL --connect-timeout 5 --max-time 10 \
  -o /dev/null -w '%{url_effective}' \
  -A freenet-updater \
  https://github.com/freenet/freenet-core/releases/latest \
  | sed -n 's|.*/releases/tag/||p')" || vsn=""
```

The existing `|| vsn=""` then routes a timeout into the retry path you already wrote, which is the behaviour you intended.

flake.nix
```diff
@@ -2,48 +2,164 @@
...
-        '';
+      overlays.default = final: prev: {
+        freenet = final.callPackage (import ./package.nix) {
+          inherit gitCommitHash gitDirty;
+          sourceDateEpoch = toString self.lastModified;
+        };
+        freenet-autoupdate = final.writeShellApplication {
+          name = "freenet-autoupdate";
+          runtimeInputs = [
+            final.curl
```
sanity, line 46: Minor: `runtimeInputs` omits `coreutils` and `gnused`, but the script uses `sleep` (coreutils) and `sed`. `writeShellApplication` *prepends* `runtimeInputs` to `$PATH` rather than replacing it, so these resolve from the ambient PATH in a normal environment — this is hardening, not a live bug. But under a supervisor with a stripped `PATH` a missing `sed` gives 127 → `pipefail` → `|| vsn=""` → the retry branch, i.e. an infinite loop logging "Could not resolve the latest freenet release" and blaming the network for a missing binary.

```nix
runtimeInputs = [
  final.coreutils
  final.curl
  final.gnused
  final.nix
];
```

flake.nix
```diff
@@ -2,48 +2,164 @@
...
+            final.nix
+          ];
+          text = ''
+            # So the node exits 42 for us instead of logging that it will not update.
+            export FREENET_SUPERVISED=1
+
+            # Mirrors MAX_UPDATE_FAILURES (crates/core/src/bin/commands/update.rs):
```
sanity, line 53: Nit: `MAX_UPDATE_FAILURES` is at `crates/core/src/bin/commands/auto_update.rs:121`, not `update.rs`. Worth fixing because this repo pins cross-file references in tests, so a stale path tends to outlive the person who can spot it.

flake.nix
```diff
@@ -2,48 +2,164 @@
...
+            # stop if the node keeps asking to update but the tag never moves.
+            max_failures=3
+            failures=0
+            last_vsn=""
+
+            # ±20% jitter, mirroring UPDATE_REPOLL_JITTER_FRACTION
```
sanity, line 59: Nit: the comment says the jitter mirrors `UPDATE_REPOLL_JITTER_FRACTION`, but that constant is `0.25` (auto_update.rs:1612) and the code does ±20%. ±20% is what was actually asked for, so the code is right and the comment is wrong — worth rewording so nobody later "resyncs" the code to 0.25 on the strength of it.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,81 @@
...
+    paths:
+      - 'flake.nix'
+      - 'flake.lock'
+      - 'package.nix'
+      - 'rust-toolchain.toml'
+      - 'Cargo.lock'
+      - 'crates/core/build.rs'
+      - '.github/workflows/nix.yml'
+
+concurrency:
```
sanity, line 31: Suggestion: add an explicit read-only token.

```yaml
permissions:
  contents: read
```

The repository default is `write` (`default_workflow_permissions: write`), and while fork PRs get a read-only token regardless, the `push:` trigger doesn't — so on main this compiles nixpkgs and rust-overlay while holding a write-capable `GITHUB_TOKEN`. 18 of 24 workflows here also omit a block, so this is convention rather than something you introduced; it's just that a job whose entire purpose is executing third-party build inputs is the one worth the two lines, and this PR is also what starts the automated monthly bumps.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,81 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  nix-build:
+    name: Nix flake build
+    runs-on: ubicloud-standard-8
+    # A cold build compiles the whole dependency graph; no binary cache exists.
+    timeout-minutes: 90
```
sanity, line 40: Minor: the comment ("a cold build compiles the whole dependency graph; no binary cache exists") is a fair worry but measurably not what happens — the run on this HEAD was genuinely cold and took **4m40s** end to end (418 `Compiling` lines, `Finished release profile in 3m 03s`), because nixpkgs comes from cache.nixos.org and only the workspace builds from source. 90 is ~19x that; 30 would still be ~6x headroom and matches `playwright-shell.yml`, without tying up a paid 8-vCPU runner for an hour when something wedges.

The same measurement retires the cost objection to this job generally: ~23% of merged commits touch `Cargo.lock`, so it fires on roughly 1 PR in 4 at ~5 minutes.

sanity (commented):
Re-reviewed at `82ebec3` (Full tier: five independent lenses — build-metadata, shell, CI/infra, supply-chain, testing — each reading blind. The external non-Claude pass could not run, Codex is out of quota until tomorrow, so a fifth Claude lens was substituted; recording that rather than quietly dropping it).

The jitter is right, and I checked it rather than taking it on trust: 60 → 48-72s, 120 → 96-144s, 180 → 144-216s, never zero or negative, and the `$(( ))` sits inside `printf`'s arguments so it dodges the `set -e` "arithmetic evaluating to 0 returns 1" trap. All required CI is green.

**But please don't merge this HEAD — there's one blocking bug, and it's in the fix for an earlier review finding.** Details inline; short version:

1. **Blocking.** `package.nix` passes `"unknown"` into `FREENET_GIT_COMMIT_HASH`, and `build.rs` panics on it (`u n k n o w n` contains no hex digit). So every source without VCS metadata fails to build — a release tarball, a `path:` drop, a bare `callPackage ./package.nix { }`. That's precisely the case the `else "unknown"` fallback was added to support when we asked you to guard `self.dirtyRev`, so that fix can never actually be exercised. CI can't see it: `actions/checkout` leaves a clean `.git`, so `self.rev` is always a real sha.

2. **Should fix before merge.** The restart cap counts *every* exit 42 for the wrapper's lifetime. 42 is overloaded — `EXIT_CODE_UPDATE_NEEDED` (auto_update.rs:30) and `FATAL_LISTENER_EXIT_CODE` (p2p_impl.rs:48) — and the distinct fast-crash code 45 is only emitted under a systemd unit advertising it, which this wrapper isn't. So an ordinary crash, or a redb-poison restart (freenet.rs:352), also arrives as 42. Three of those across a node's lifetime, however far apart, take it off the network permanently. 42 is deliberately *burst-exempt* under the supervisors the node ships with; systemd's own limiter is a rate (`StartLimitBurst=5`/`StartLimitIntervalSec=120`), not a total.

3. **Should fix before merge.** The release probe has no timeout — a blackholed :443 stalls the supervisor forever, no log line, no node.

I wrote and tested the patch for all three, but I can't push it: this is a fork and the API gives me `push: false`, and I'm deliberately not fetching or building a contributor's branch locally. So it's yours to apply or improve — the full diff is in the inline comments, and the counting change is verified against six sequences (healthy-node-crashes-3-times-far-apart now survives; a genuine boot wedge is still bounded at 3, so the cap you were asked for is preserved).

Two things in the description still need a fix, and both end up in the permanent commit message since this repo squash-merges:
- Item 3 still claims reproducibility lets "anyone verify that the binaries on the GitHub releases page were not tampered with". It doesn't — `cross-compile.yml` builds those with plain `cargo build --release` and `SOURCE_DATE_EPOCH` appears nowhere under `.github/`. You said this was reworded; it wasn't.
- The `SuccessExitStatus=` note for systemd users that you said you'd documented isn't in the description either. The `break` → `exit "$exit_code"` change is real and matters: under `Restart=on-failure`, 130 (Ctrl-C), 2 and 43 flip from clean stop to restart loop.

Everything else I'd take as follow-up issues rather than hold this on: `$vsn` isn't validated before going into the flakeref (bounded — the repo is a fixed literal, so worst case is ref selection within this repo, and it's still strictly better than main's current unpinned default-branch `nix run`); the `paths:` filter keys on `Cargo.lock` and so misses a feature flag pulling `openssl-sys` into the graph; and `SOURCE_DATE_EPOCH` is parsed strictly while the two Freenet vars treat empty as "no override".

Sorry this took another round — the two real bugs were both in the fixes rather than the original, which is the case for re-reviewing rather than trusting the previous pass. The overall change is good and I'd like to land it.

[AI-assisted - Claude]

sanity:
Three small additions from the remaining lenses, then I'll stop adding to this.

**The new test pins the parse but not the use.** `build_script_git_dirty_override_cannot_silently_disable_auto_update` asserts on the four match arms and the two `rerun-if-env-changed` lines, all of which is good. But the line the parse feeds is unpinned:

```rust
let dirty_suffix = if git_dirty { "-dirty" } else { "" };
println!("cargo:rustc-env=GIT_DIRTY={dirty_suffix}");
```

Invert that ternary and all five assertions stay green, while every binary ships with auto-update disabled — the exact outcome the test is named after. One more `body.contains(...)` on the `dirty_suffix` line would close it. Worth doing precisely because the test is otherwise well built: I checked the vacuity cases (signature appearing twice, `\n}\n` landing early, empty body, renamed function) and they all fail loudly, and `include_str!("../../../build.rs")` does resolve correctly to `crates/core/build.rs` — including under `cargo package`, since `crates/core` sets no `include`/`exclude`. The cross-file scrape is the right mechanism here and I'm not asking you to replace it.

**Worth adding to the curl fix** alongside the timeout, since it's the same line: `--proto '=https' --proto-redir '=https' --max-redirs 3`. The node's own resolver refuses any scheme/host/port change while following this redirect and caps the chain at `MAX_PROBE_REDIRECTS = 3` (auto_update.rs:824-878), with an explicit threat model in the comment. `curl -L` by default will follow off-origin and will downgrade to plain http. It isn't a code-execution path — the flakeref pins the repo, so the worst case is ref selection within freenet-core — but it's free to match the policy the Rust side already committed to.

**One thing I couldn't verify** and you can in ten seconds: `sourceDateEpoch = toString self.lastModified;` dereferences `self.lastModified` unguarded, on the same source types that produced the `"unknown"` bug. `rev` and `dirtyRev` both got `self ? …` guards; this one didn't. I believe Nix always populates `lastModified`, but I have no nix here to confirm it — if it's always present, a one-line comment saying so would stop the next reviewer asking again.

[AI-assisted - Claude]

sanity:
Last one, and it's the useful half: two ways to stop this class of bug rather than just this instance. Both belong with the fix.

**One CI step would have caught the blocking bug.** The non-git eval path has no coverage *by construction*: `nix.yml` uses `actions/checkout`, which always leaves a clean tree, so `self ? rev` is true and the job only ever exercises the `rev` branch. `nix flake check --no-build` never reaches `build.rs` at all. That's why `"unknown"` could reach a hex validator and nobody noticed. Adding this to `nix.yml` closes it:

```yaml
      # The `rev` branch is the only one a git checkout can reach, so the
      # no-provenance fallback (gitCommitHash = "unknown") is otherwise
      # untested — which is how it shipped a build-time panic.
      - name: Build from a source drop with no VCS metadata
        run: nix build 'path:.' -L
```

Cheap, and it fails today against `82ebec3`, which is the property you want from a regression test.

**And the new supervisor isn't in the pin that enumerates supervisors.** `supervisors_set_supervised_marker` (crates/core/src/bin/freenet.rs:2275) already pins the in-process wrapper, the macOS launchd script and both systemd unit templates, each asserted against the Rust source, all from #4580. This PR adds a fourth supervisor that correctly exports `FREENET_SUPERVISED=1` but doesn't extend that test, so the next person to touch the marker has three of four call sites pinned and no signal about the fourth.

The same gap is what let the exit-42 conflation through. No unit test could have caught it, but a pin would have forced a look at both meanings of 42 while the code was being written — and note the wrapper's constants have *already* drifted on arrival: the comment claims it mirrors `UPDATE_REPOLL_JITTER_FRACTION` when that constant is 0.25 and the code does ±20%. Two hand-copied constants, one already wrong, none pinned. Since `flake.nix` sits outside `crates/core`, `include_str!` can't reach it without breaking `cargo package`, so a `grep` step in `nix.yml` is the right home for that one.

Neither of these blocks the merge on its own — but the first is what turns "we fixed it" into "it can't come back", and it's three lines.

[AI-assisted - Claude]

sanity:
## Review: one blocking build defect, and a scope question for the maintainers

Thanks @Skyb0rg007 — and sorry for the slow round-trip. Reviewed by reading only (the repo's hook deliberately blocks building outside-contributor code on this machine, which is expected and not a comment on your PR). The flake genuinely is broken on `main` today — `package.nix:7` uses bare `fromTOML`, which is not in scope in a `callPackage` file — so the direction here is worth having, and the toolchain pin, devShell, `SOURCE_DATE_EPOCH` handling and git-provenance env vars are all clean work. The mutation-tested source pin is a nicer test than most in-house PRs get.

I checked the 2026-09-06 force-push separately: it is a clean rebase onto newer `main`, all 9 files byte-identical except a `docker/freenet-node` block that came *from* main. No earlier fix was dropped.

### BLOCKING — the no-VCS fallback panics at build time

`package.nix:4` defaults `gitCommitHash ? "unknown"`, `:11` whitelists it, `:25` passes it into `FREENET_GIT_COMMIT_HASH`, and `crates/core/build.rs:120` then rejects any non-hex value with `panic!`.

So any source without VCS metadata fails to build: a release tarball, `nix build 'path:.'`, and `pkgs.callPackage ./package.nix { }` with default arguments — which is exactly how a downstream nixpkgs packaging would call it. CI cannot see this, because `actions/checkout` always leaves a clean `.git`, so `self ? rev` is always true and the `else "unknown"` branch is never exercised.

Fix is one line: `FREENET_GIT_COMMIT_HASH = if gitCommitHash == "unknown" then "" else gitCommitHash;` — empty means "no override" per `build.rs:116`, and the `git` fallback at `build.rs:125-132` then correctly yields `unknown` in the sandbox. Worth adding `nix build 'path:.' -L` to the workflow so it can never regress.

### IMPORTANT

**A GitHub outage takes every nix-supervised node offline, and the lookup can hang forever** (`flake.nix:73-84`). The loop resolves `releases/latest` before *every* start and on failure only sleeps and retries — it never falls back to the last version it successfully ran, though that store path is already built locally. `curl -fsSL` has no `--connect-timeout`/`--max-time`, and curl has no default timeout, so a black-holed connection wedges the supervisor with the node down. `--proto '=https' --proto-redir '=https' --max-redirs 3` is also missing; `curl -L` will follow off-origin and downgrade to http. The Rust side already commits to all of this at `auto_update.rs:824-878`. An unreachable github.com is squarely in Freenet's threat model, so "node keeps running what it has" is the behaviour we want.

**Exit 42 is overloaded** (`flake.nix:91-111`). `p2p_impl.rs:48` uses 42 as `FATAL_LISTENER_EXIT_CODE` (#4549) — a healthy-uptime fatal exit that systemd treats as burst-exempt and always restarts. The wrapper counts those toward `max_failures` and then `exit 1`s permanently, so three such exits between releases leaves the node down until a human notices. At minimum replace the terminal `exit 1` with a capped indefinite backoff.

**The devShell will stamp every `cargo build` as 1980-01-01** (`flake.nix:146-162`). nixpkgs' stdenv exports `SOURCE_DATE_EPOCH=315532800`, `nix develop` sources that setup, and `build.rs:158` now honours it — so binaries built in the shell report that date in `freenet --version`, the field ops uses to correlate a log with an artifact. `package.nix:6-8` shows you already handled this for the package; the devShell needs `shellHook = "unset SOURCE_DATE_EPOCH";`. This one only exists because two independently-good parts of the PR met.

**The reproducibility claim has no gate.** `CONTRIBUTING.md` states "the same source gives a bit-identical binary", and `rebuild` appears zero times in the diff. Either add `nix build .#freenet --rebuild -L` to the workflow or drop the sentence — a documented property with no check drifts silently the first time a dependency embeds a path.

Related, and this one matters beyond the PR: the description says this lets anyone verify the GitHub release binaries were not tampered with. It cannot. Releases are static **musl** cross-builds from `cross-compile.yml`; the nix build is glibc-dynamic against nixpkgs-unstable, and nothing compares them. Same version string, materially different artifact. Please drop that claim rather than carrying it into release notes.

### Minor

- `.github/dependabot.yml` nix block omits `cooldown: default-days: 7` and `open-pull-requests-limit`, which all seven other ecosystems set; the file's own header argues explicitly for setting the limit. Without the cooldown, a freshly-broken nixpkgs-unstable rev is proposed immediately, which is the case the cooldown exists for.
- `nix.yml:4` says "the weekly nixpkgs-unstable bump"; the block this same PR adds schedules `monthly`.
- `flake.nix:59` says "±20% jitter, mirroring `UPDATE_REPOLL_JITTER_FRACTION`"; that constant is `0.25`.
- `flake.nix:43-114` never forwards `"$@"`, so `nix run github:freenet/freenet-core -- --config-dir ...` silently discards arguments — worth fixing now that `CONTRIBUTING.md` documents `nix run .` as an entry point.

### Two repo-side things this needs, which are ours not yours

`.claude/rules/deployment.md:24-42` documents exactly two ways the auto-update gate is controlled; this PR adds a third (`FREENET_GIT_IS_DIRTY` as a build-time kill switch), so that rule will read as complete while being wrong. And `freenet.rs:2275`'s `supervisors_set_supervised_marker` pins three supervisors against source text (#4580); this adds a correct fourth that the pin cannot see, since `flake.nix` is outside `crates/core` and `include_str!` cannot reach it.

### The scope question — and why I am not merging on my own judgment

The thing I want to flag plainly, because it is a decision rather than a defect: **this is not only a build-system fix. It adds a fourth production supervisor and a second software-update channel**, and `packages.default` makes it what `nix run github:freenet/freenet-core` hands you. That channel picks a tag over TLS and runs it, with no verification against the release signing key the normal path uses (`update.rs:43-120`), no known-bad pin, and no crash-loop rollback — while exporting `FREENET_SUPERVISED=1`, which the node reads as a promise that its supervisor *will* apply updates and restart, so it stops warning on the strength of it.

That is a maintainer call, not something to accept silently in review, so I have put it to @sanity: is nix a supported deployment path (in which case the wrapper needs signing and rollback parity, and we need a binary cache — without one every nix user recompiles the whole dependency graph on each release with the node down throughout), or a developer convenience (in which case dropping `packages.default = freenet-autoupdate` makes most of the above moot and the build-only half lands quickly)?

**The build-only half of this PR is good and nearly independent of the supervisor half**, so if the answer is "dev convenience", this gets much shorter.

Nothing here is a reason to be discouraged — the blocking item is one line, and the rest is mostly about a decision that was never yours to make.

[AI-assisted - Claude]

sanity:
## Decision on scope, and what happens to this PR

@Skyb0rg007 — Ian has made the call I raised in my last review: **nix is a supported deployment path**, not a developer convenience. His reasoning is that nix is popular and getting more so, and that it deserves a first-class story rather than a partial one.

That is the answer you were owed weeks ago, and it should not have taken this long to get to you. Thank you for staying with it through several rounds — the jitter change, the `else "unknown"` fix, the mutation-tested pin. None of that was wasted.

### What we are doing, and why it differs from this PR

Because nix is now a *deployment* path, the governing constraint became Ian's: **a peer that does not auto-update is a liability for the network.** That turned out to force a different design, for a reason neither of us had on the table.

`freenet update` calls `std::env::current_exe()` and replaces the running binary **in place** (`crates/core/src/bin/commands/update.rs`), after verifying `SHA256SUMS.txt` against the ed25519 key baked into the binary. `/nix/store` is read-only. **So the node's own updater physically cannot run from a store path.**

Your wrapper solves that by re-resolving the flake and running `nix run github:freenet/freenet-core/$vsn#freenet`. That works, and it is a reasonable reading of the problem — but as a deployment path it has costs we cannot take:

- it **rebuilds from source**, and no binary cache exists, so every nix peer recompiles the whole dependency graph on every release with the node down throughout;
- it trusts GitHub plus the flake hash rather than `FREENET_RELEASE_PUBKEY`, so it cannot participate in the signed-floor work;
- it never consults the known-bad pin a crash-loop rollback writes, and has no rollback of its own.

The alternative we are taking: **nix bootstraps the binary into a mutable state directory, and the node self-updates from there**, exactly as on every other platform. The supervisor becomes a faithful port of the systemd unit — and the thing that makes that work is something your version reasonably assumed otherwise: **systemd makes no update decisions at all.** It does not resolve tags, poll GitHub, or count failures. It restarts the node and runs `freenet update`; the updater disambiguates exit 42 (which is also `FATAL_LISTENER_EXIT_CODE`), verifies signatures, snapshots for rollback, and counts crash probation.

Ported faithfully, the wrapper *loses* its update block — no curl, no jitter, no `max_failures`, no tag parsing — and nix peers then auto-update identically to every other peer, covered by the same Gate A / Gate B canaries we already run at release time.

The honest cost is that the running binary drifts from the store path over time, which is not nix-idiomatic. So we will ship both: a pinned, declarative, non-self-updating `packages.freenet` for people who want `nixos-rebuild` to control the version, and the self-updating node as `packages.default`.

### Your work is being carried over, with credit

The flake fix itself (`package.nix`'s bare `fromTOML` — the repo genuinely does not build under nix today), the toolchain pin to `rust-toolchain.toml`, the devShell, `SOURCE_DATE_EPOCH`, and the git-provenance env vars are all your design and are being kept. The replacement PR credits you and links here.

One blocking defect carries over and is fixed there: `package.nix` defaults `gitCommitHash ? "unknown"` and passes it to `FREENET_GIT_COMMIT_HASH`, which `build.rs` requires to be all-hex, so it panics. That breaks a release tarball, `nix build 'path:.'`, and any downstream `pkgs.callPackage`. CI could never catch it — `actions/checkout` always leaves a clean `.git`, so the no-VCS branch has no coverage by construction. We are adding `nix build 'path:.'` so it cannot regress.

Also dropping one claim rather than carrying it: reproducibility here does not let anyone verify the published release binaries. Releases are static **musl** cross-builds; the nix build is glibc-dynamic, and nothing compares the two.

**I am leaving this PR open rather than closing it** — that is Ian's call, not mine. If you would rather take the redesign yourself, say so and I will hand it over; you have earned the option.

[AI-assisted - Claude]

sanity:
## Closing — we are taking this work over, with credit

@Skyb0rg007 — closing this now that the redesign is underway, and Ian has confirmed nix is a **supported deployment path**. That is the decision you were waiting on, and you waited far too long for it. Thank you for the persistence through several review rounds.

To be clear about why this is closed rather than merged: **it is not a rejection of the work.** The parts that were yours are being carried into the replacement, and the PR credits you:

- the flake fix itself — `package.nix`'s bare `fromTOML` means the repo genuinely does not build under nix today, and you found it;
- pinning the toolchain to `rust-toolchain.toml`, so clippy and rustfmt match CI;
- the devShell;
- `SOURCE_DATE_EPOCH` support and the git-provenance env vars;
- the mutation-tested source pin, which was better than most in-house tests get.

The one thing that changed is the update mechanism, and only because of a constraint neither of us had surfaced when you designed it: `freenet update` calls `std::env::current_exe()` and replaces the running binary **in place**, after verifying `SHA256SUMS.txt` against the ed25519 key baked into the binary. `/nix/store` is read-only, so the node's own updater cannot run from a store path at all.

Your wrapper answered that by re-resolving the flake per release. For a dev convenience that is the right shape. As a deployment path it costs too much: it rebuilds from source with no binary cache (node down for a full dependency-graph compile, and Ian ships **multiple releases per day**), it trusts GitHub plus the flake hash rather than the release signing key, and it never consults the known-bad pin a crash-loop rollback writes.

The replacement has nix bootstrap the binary into a **mutable state directory**, after which the node self-updates exactly as on every other platform. The supervisor becomes a port of the systemd unit — and the thing that makes it small is that **systemd makes no update decisions at all**: it restarts the node and runs `freenet update`, and the updater handles tag resolution, exit-42 disambiguation (42 is also `FATAL_LISTENER_EXIT_CODE`), signature verification, rollback and crash probation. So the wrapper loses its whole update block, and nix peers fall under the same Gate A / Gate B release canaries as everyone else.

Two things from your PR are being fixed rather than carried:

- `package.nix` defaults `gitCommitHash ? "unknown"` and passes it to `FREENET_GIT_COMMIT_HASH`, which `build.rs` requires to be all-hex — so it panics on any source without VCS metadata (a release tarball, `nix build 'path:.'`, downstream `pkgs.callPackage`). CI could never have caught it: `actions/checkout` always leaves a clean `.git`, so the no-VCS branch has no coverage by construction. We are adding `nix build 'path:.'` so it cannot come back.
- The reproducibility claim that this lets anyone verify the published release binaries. It cannot — releases are static **musl** cross-builds, the nix build is glibc-dynamic, and nothing compares them.

**If you would rather own the redesign, say so and I will hand it back** — the offer stands and is not a formality. Otherwise, review is very welcome on the replacement PR when it goes up; you know this ground better than we do, and the drift tradeoff (a self-updating binary diverging from its store path is not nix-idiomatic) is exactly the kind of thing worth your eyes.

[AI-assisted - Claude]

Skyb0rg007:
Hmm, based on this description I'll also close my nixpkgs PR, since that could result in out-of-date clients.

PR #5094, opened 2026-08-01
fix(shell): reply with a notification_status on every notification path

### LINKED ISSUE #5043: fix(shell): notification_status is never sent on success or recovery, so a transient `undeliverable` sticks for the session
Found while closing River's half of freenet/river#510 (River now consumes
`notification_status` and surfaces it). Distinct from #4966, which is about
`requestPermission` never settling; this is about statuses the shell never sends
at all.

All line references are `crates/core/src/server/path_handlers/assets/shell_bridge.js`
on `main` as of 2026-07-29, and each was verified by reading the source rather
than inferred.

## 1. A transient `undeliverable` is never retracted

`notifyRegistrationReady(timeoutMs)` resolves `null` when the service worker is
registered but not yet active — the comment there says "Registered but not yet
active (first visit): wait briefly for it", and the `setTimeout` resolves `null`
on expiry. `showAppNotification` calls it with `1500` and, on `null`, posts
`notifyStatusToIframe('undeliverable')`.

So on mobile Chrome/Firefox the **first** notification of a session can race the
worker's activation, report `undeliverable`, and then start working normally.
There is no `notifyStatusToIframe` on the success path and no "recovered"
status, so nothing ever retracts it.

Consequence for consumers: a client that trusts `undeliverable` shows a
permanent problem indicator for a system that is working. River is now such a
client — as of freenet/river#542 plus its follow-up it puts a persistent marker
on the notification bell — so this shell behaviour is what promotes a 1.5s race
into a session-long false warning. That is why it is worth fixing rather than
documenting.

## 2. A per-message delivery drop is never reported at all

`showAppNotification` returns silently in three cases, none of which calls
`notifyStatusToIframe`:

- `typeof Notification === 'undefined'`
- `Notification.permission !== 'granted'`
- `!contractHasConsent()`

So the app is never told that a specific notification was dropped.

The interesting combination: a `granted` reported once is the last thing the app
hears. If the user then revokes the permission in site settings, every
subsequent notification is dropped at the `permission !== 'granted'` check while
the app keeps displaying "notifications are enabled" indefinitely. A framed app
**cannot** detect this itself — its origin is opaque, so `Notification.permission`
is unreadable, which is the whole reason the status bridge exists. Only the shell
can close it.

(River fixed the equivalent staleness on its *top-level* path by always
preferring the browser's live answer over a remembered one. That remedy is
unavailable framed.)

## 3. Minor: the affordance-already-showing path returns with no status

`if (notifyAffordanceShown) return;` in `maybeOfferNotifications` returns without
posting a status. Benign today — the bar is on screen, so the user is not left
without feedback — but it means a `notification_enable_prompt` can be answered
with silence, which a client cannot distinguish from the message being lost.

## Suggested direction

Not prescriptive about the shape, but the useful invariant is: **every
`notification_enable_prompt` and every `notification` gets exactly one status
reply.** Concretely that means a success status (or a `granted` re-affirmation)
on the delivery path, and reporting the three silent-return cases in
`showAppNotification` rather than dropping them. A retractable
`undeliverable` — or simply not reporting it until the worker has had a real
chance to activate — would fix (1).

Wire compatibility: these are new values on an existing string field, and River
ignores unrecognised statuses rather than guessing (it logs at `warn!` and keeps
the status it has), so adding statuses will not break the current client.

[AI-assisted - Claude]

skandragon:
## Problem

`shell_bridge.js` never sends a `notification_status` on several paths, so a framed app's view of notification health goes stale and can't recover (#5043):

1. **A transient `undeliverable` is never retracted.** `notifyRegistrationReady(1500)` resolves `null` when the notification service worker is registered but not yet active. On mobile Chrome/Firefox the *first* notification of a session can lose that 1.5s race, report `undeliverable`, then work fine — with no success status, nothing ever retracts it. River puts a persistent marker on the notification bell for it, so a 1.5s race becomes a session-long false warning.

2. **Per-message drops are never reported.** `showAppNotification` returned silently for `typeof Notification === 'undefined'`, `permission !== 'granted'`, and `!contractHasConsent()`. Worst case: the user revokes the permission in site settings after a `granted`; every notification is then dropped at the permission check while the app still displays "notifications enabled". The app **cannot** detect this itself — its origin is opaque, so `Notification.permission` is unreadable. Only the shell can close it.

3. **The affordance-already-showing path returned with no status**, so a `notification_enable_prompt` could be answered with silence — indistinguishable from a lost message.

## Solution

The invariant: **every `notification` and every `notification_enable_prompt` gets exactly one status reply.**

- Each of the three silent returns in `showAppNotification` now reports why: `unsupported`, `denied`/`default`, `default`.
- Both delivery paths (desktop page-level constructor, mobile service-worker `showNotification`) re-affirm `granted`. That re-affirmation *is* the retraction of a stale `undeliverable` — no new "recovered" status needed, and no change to the wire shape.
- `maybeOfferNotifications` replies `default` when the bar is already on screen.

The constructor-path `granted` is posted **outside** the `try`, guarded by a flag: a throw from the status post must not read as a constructor failure and fall through to a second, duplicate delivery via the service worker.

Rate-limiter drops still return silently — reporting a status on every throttled notification would be noise, and the throttle is working-as-intended rather than a health signal.

Wire compatibility: these are new values on an existing string field. River ignores unrecognised statuses (logs at `warn!` and keeps the status it has), so no client breaks.

## Testing

`shell_bridge_notifications.test.mjs` gains case 9, which extracts `showAppNotification` verbatim from the asset (between new `notify-show:BEGIN`/`:END` markers — the same technique the existing rate-limiter cases use) and drives it with every collaborator injected. It asserts *exactly one* status per call across all nine branches: the four drop paths, both delivery paths, both genuine-failure paths, and a retraction sequence where an `undeliverable` is followed by a delivery and the last status the client sees is `granted`. Verified failing against the pre-fix asset.

`bridge_js_notification_proxy_invariants` in `path_handlers.rs` pinned the combined `permission !== 'granted' || !contractHasConsent()` expression; the gates are now two statements so each can report its own status, so the pin asserts both substrings. Same invariant, same strength.

`npm test` and `npm run lint` in `crates/core/src/server` pass; `cargo test -p freenet --lib server::` passes (436). Pre-existing clippy lints elsewhere in the crate are untouched.

## Fixes

Closes #5043

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -5602,11 +5602,6 @@ mod tests {
                 && show_slice.contains("!contractHasConsent()"),
             "showAppNotification must gate on browser permission AND per-contract consent"
         );
-        // Note: "exactly one status per message" is asserted BEHAVIORALLY in
-        // shell_bridge_notifications.test.mjs (cases 9 and 10), which drives the
-        // extracted functions and counts the posts. A substring count here would
-        // be a weaker restatement of that, so it deliberately isn't duplicated.
-        //
         // "Not now" must be durable so a contract that re-sends the enable prompt
         // can't re-pin the host-owned bar over the app.
         assert!(
@@ -5669,6 +5664,80 @@ mod tests {
         assert!(
             !SHELL_BRIDGE_JS.contains("'serviceWorker' in navigator"),
             "feature-detect by attempting the read (serviceWorkerOrNull), not via `in`"
+        );
+    }
+
+    /// #5043: a framed app can't read `Notification.permission` (opaque origin),
+    /// so the shell reporting a status is its ONLY way to learn a notification
+    /// was dropped. Regression pins for the two paths that were silent and were
+    /// re-broken during review of the first fix — the rate-limiter drop, and the
+    /// async service-worker chain, whose only rejection handler covered
+    /// `showNotification` and left a rejected registration lookup (or a
+    /// synchronous throw) with no reply at all.
+    ///
+    /// The exactly-one-status-per-message behavior is verified by driving the
+    /// real extracted functions in `shell_bridge_notifications.test.mjs` (cases
+    /// 9 and 10, run by the lint-assets CI job). These are source pins for the
+    /// two specific silent-return shapes, in the same discipline as the
+    /// `SHELL_BRIDGE_JS.contains` guards above, so a refactor that reintroduces
+    /// either shape fails here too.
+    #[test]
+    fn bridge_js_notification_drops_are_never_silent() {
+        let start = SHELL_BRIDGE_JS
+            .find("notify-show:BEGIN")
+            .expect("notify-show:BEGIN marker must bracket showAppNotification");
+        let end = SHELL_BRIDGE_JS[start..]
+            .find("notify-show:END")
+            .expect("notify-show:END marker must bracket showAppNotification");
+        let show = &SHELL_BRIDGE_JS[start..start + end];
+
+        // The rate-limiter drop is the most frequent one (a busy room hits the
+        // 3s per-tag throttle constantly). `if (...) return;` on one line is the
+        // exact shape it had while it was silent.
+        assert!(
+            !show.contains("if (!notifyLimiter.ok(opts.tag, Date.now())) return;"),
+            "the rate-limited drop must post a status, not return silently (#5043)"
+        );
+        assert!(
+            show.contains("notifyLimiter.ok(") && show.contains("notifyStatusToIframe('granted')"),
+            "the rate-limited drop must report 'granted' — permission and consent \
+             are intact and the shell merely coalesced the message"
+        );
+
+        // The service-worker chain needs a terminal .catch: without it a rejected
+        // notifyRegistrationReady, a synchronous throw from showNotification, or
+        // a non-thenable return all leave the app with no reply.
+        let sw_start = show
+            .find("notifyRegistrationReady(")
+            .expect("the mobile fallback must go through notifyRegistrationReady");
+        let sw_chain = &show[sw_start..];
+        assert!(
+            sw_chain.contains(".catch("),
+            "the service-worker delivery chain must end in a .catch so a rejected \
+             registration lookup or a throwing showNotification still replies (#5043)"
+        );
+        // ...and that backstop must not turn one reply into two: every post from
+        // the chain goes through the post-at-most-once helper, never through
+        // notifyStatusToIframe directly.
+        assert!(
+            show.contains("var swReplied = false;") && sw_chain.contains("swReply("),
+            "the service-worker chain's .catch backstop must be guarded so a throw \
+             from the success-path status post can't produce a second reply"
+        );
+        assert!(
+            !sw_chain.contains("notifyStatusToIframe("),
+            "the service-worker chain must post via swReply(), which is what bounds \
+             it to one reply — a direct notifyStatusToIframe call bypasses that"
+        );
+
+        // The constructor path's status post must sit OUTSIDE the try: inside, a
+        // throw from it reads as "constructor unsupported" and the worker path
+        // displays the SAME notification a second time.
+        assert!(
+            show.contains("shownByConstructor = true;")
+                && show.contains("if (shownByConstructor) {"),
+            "the constructor path must record delivery in a flag and report \
+             outside the try, so a throwing status post can't double-deliver"
         );
     }
 
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -5575,9 +5575,11 @@ mod tests {
         // Every notification is gated on BOTH the browser permission AND this
         // contract's own consent, so one contract's gateway-wide browser grant
         // can't notify the user on behalf of a different contract.
+        // (Two separate gates since #5043, so each drop can report its own
+        // `notification_status` back to the app instead of returning silently.)
         assert!(
-            SHELL_BRIDGE_JS
-                .contains("Notification.permission !== 'granted' || !contractHasConsent()"),
+            SHELL_BRIDGE_JS.contains("Notification.permission !== 'granted'")
+                && SHELL_BRIDGE_JS.contains("!contractHasConsent()"),
             "showAppNotification must gate on browser permission AND per-contract consent"
         );
         // "Not now" must be durable so a contract that re-sends the enable prompt
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -5572,16 +5572,41 @@ mod tests {
             SHELL_BRIDGE_JS.contains(r"/\/v[12]\/contract\/web\/([^/?#]+)/"),
             "notification consent key must derive from the /v[12]/contract/web/<key> path"
         );
+        // The markers bracket showAppNotification so shell_bridge_notifications
+        // .test.mjs can extract and drive it. Pin them here: without this,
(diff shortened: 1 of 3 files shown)


skandragon:
Filed #5095 for the one remaining silent drop (rate-limiter), left out here to keep this to one logical change.

skandragon:
Pushed d0486f7 after a multi-agent review pass. The reviewers were right that the first commit asserted an invariant it didn't hold — four more silent paths, all inside the code the commit rewrote:

- **Rate-limiter drop** posted nothing, and it's the most frequent drop in practice. Now reports `granted` (coalesced by shell policy, permission and consent intact) so a throttled message can't read as a delivery problem — and can still retract a stale `undeliverable`. Closes #5095, which I'd filed to defer this.
- **Service-worker chain** handled only `showNotification`'s rejection. A synchronous throw, a non-thenable return, or a rejection from the registration lookup itself produced zero statuses. One terminal `.catch` backstops the chain; a `swReplied` flag keeps it to exactly one post even if a status post throws.
- **`shownByConstructor`** was set after the `onclick` assignment, so a throw from that assignment could fall through to the worker path and display the same notification twice. Set immediately after the constructor returns.
- **Missing contract key** reported `default`, sending the app into a re-prompt loop it can never win. `maybeOfferNotifications` already calls that same permanent condition `undeliverable`; they now agree.

Also: the constructor `catch` logs at `console.debug` (on desktop that throw is a real bug, previously indistinguishable from the expected mobile path), and the header comment now states the two real bounds on the guarantee rather than an unconditional claim.

Tests grew from 9 checks to 20. The ones worth calling out: the `shownByConstructor` hoist is now pinned by a throwing status post (reverting the hoist fails only that check — before, the revert passed the whole suite), the retraction check drives one harness across two messages and asserts the exact `['undeliverable', 'granted']` sequence instead of re-asserting what the test itself pushed, and new `notify-offer` markers give the enable-prompt status ladder its first coverage at all.

The Rust pin now scopes its gate assertion to the marker-bounded region and pins the markers themselves — deleting the extraction region together with its tests no longer leaves CI green.

Not changed: `sendToIframe` no-ops when the iframe is gone, so "exactly one reply" is bounded by best-effort delivery, and a `notification` with a non-string `title` is dropped by the dispatcher's type check before reaching `showAppNotification`. Both are now stated in the header comment rather than papered over.

skandragon:
Pushed c0f18e2 to satisfy the `fix:`-needs-a-regression-test gate.

The gate counts Rust `#[test]`s plus `check` lines in added `*_test.sh` files. This fix's regression tests are Node self-tests in `shell_bridge_notifications.test.mjs`, which the lint-assets job runs on every PR — so they score zero and the fix is rejected. That's the same gap the shell carve-out was added for; the comment above it says as much ("a genuine shell regression test scores zero against TEST_ATTR_RE (Rust-only) and the fix is wrongly rejected"), one language over.

I did not touch `ci.yml` — changing the gate inside the PR the gate is blocking is the CI-chasing pattern this repo watches for, and widening a test-enforcement rule is a maintainer's call. I also didn't take `test-exempt`, since the fix plainly can be tested and is.

Instead the new Rust test pins the two silent-return *shapes* by source, which is the discipline the surrounding `bridge_js_*` guards already use for this asset — so it's in-convention rather than a box-tick. It's mutation-checked: reverting the limiter drop to `if (...) return;`, deleting the terminal `.catch`, swapping `swReply` for a direct `notifyStatusToIframe`, or undoing the `shownByConstructor` hoist each fail it. The behavioral proof (exactly one status per message, 20 checks across both extracted functions) stays in the .mjs suite.

**Suggested follow-up for a maintainer:** extend the `NEW_SHELL_TESTS` carve-out in `ci.yml` to count added `check(` lines in `**/*.test.mjs`. Any future fix to the browser assets hits this same wall. Happy to open that as its own PR if you want it — it seemed wrong to bundle a CI policy change into a bug fix.

skandragon:
Filed #5096 for the CI gate gap (extend the `fix:`-needs-a-test carve-out to `*.test.mjs`), so it's tracked independently of this PR.

sanity:
Reviewed and merging. Apologies for the three-week silence, especially given you did three rounds of self-review on this and filed two follow-ups (#5095, #5096) while waiting.

## What I checked

**Risk tier: Light-plus.** No transport, crypto, or wire-format surface, but the `notification_status` string field is a cross-repo contract with River, and this repo has a documented bug class for exactly that (protocol/enum changes breaking downstream consumers). So I did the diff read plus a downstream-consumer pass rather than the diff alone.

**The wire-compatibility claim holds, and I verified it against River rather than taking it from the description.** `NotificationStatus::from_shell_status` (`river/ui/src/components/app/notifications.rs:219-229`) parses exactly `granted` / `denied` / `dismissed` / `default` / `undeliverable` / `unsupported` and returns `None` for anything else, which the caller at `:581-596` ignores with a `warn!`. This PR introduces no new status string: every value it adds on a previously-silent path is one of those six. So there is no drift to absorb.

**The semantic mapping is right too, per River's own definitions of the variants:**
- `!contractConsentKey()` reporting `undeliverable` matches River's `Undeliverable` doc, which literally names "no contract key to gate consent on".
- `!contractHasConsent()` reporting `default` maps to `Undecided` ("nothing is blocked, so asking again is still worthwhile"), which is accurate: the next enable prompt can fix it.
- The already-showing path replying `default` cannot clobber a good state. `maybeOfferNotifications` returns `granted` early at `shell_bridge.js:568-575` when permission and consent are both present, so the new reply at `:582-587` is only reachable when the state genuinely is undecided.

**I traced the exactly-one-status invariant by hand across all nine branches** and it holds, including the two subtle ones: the `shownByConstructor` hoist means a throw from the `onclick` assignment still reports `granted` and returns rather than falling through to a duplicate service-worker delivery; and `swReplied` is set *before* `notifyStatusToIframe` is called, so a throw from the success-path status post yields zero replies rather than two. That is the right trade given `sendToIframe` is best-effort anyway, and your header comment already states that bound.

## Two observations, neither blocking, neither asking for a change here

**1. The rate-limited `granted` can transiently mask a genuinely undeliverable device.** The limiter runs before any delivery attempt, so on a mobile device where the service worker never activates the sequence is: first message to a tag reports `undeliverable`, a second within the 3s throttle reports `granted` with nothing displayed, and the next unthrottled message reports `undeliverable` again. River flips to `Granted` for that window. It self-corrects on the next delivery attempt and it is strictly better than the pre-fix behavior (where `undeliverable` stuck for the whole session with nothing able to retract it), so this is a limitation rather than a regression. Worth a sentence in the comment at `shell_bridge.js:858-864`, which currently argues only the flood-policy half.

**2. Each throttled drop now costs a River re-render.** `record_notification_status` writes the `NOTIFICATION_STATUS` `GlobalSignal` unconditionally (`notifications.rs:450-457`), with no equality check, so a burst of throttled messages in a busy room schedules one deferred signal write per drop. A cheap shell-side bound would be to skip the re-affirmation when the last posted status was already `granted`, which keeps the retraction property (the first throttled message after an `undeliverable` still retracts it) while dropping the steady-state chatter. Equally fixable on the River side. Either way it is a UI cost, not a correctness problem.

## On the CI gate

Your read on `NEW_SHELL_TESTS` is right, and declining to widen the gate inside the PR it was blocking was the correct call. #5096 tracks it. The Rust source pins you added instead are in-convention with the surrounding `bridge_js_*` guards, and bounding them to the `notify-show` / `notify-offer` marker regions (plus pinning the markers themselves, so deleting the region with its tests cannot leave CI green) is a genuinely stronger shape than the file-wide `contains` it replaced.

Thanks for the care here, and for the mutation checks.

[AI-assisted - Claude]

Issue #5087, opened 2026-07-31
Same-origin target="_blank" opens a sandboxed shell: cross-contract links land on a blank tab


skandragon:
## Summary

A same-origin `<a target="_blank">` link inside a contract app opens a **sandboxed** top-level shell. That shell's own `frame-src 'self'` can never match (its origin is opaque), so its app iframe is refused, no request is ever dispatched, and the new tab renders blank.

`navigation_interceptor.js` already prevents this for cross-origin links (freenet/river#208), but deliberately exempts same-origin links with an explicit non-`_self` target. Cross-contract links are same-origin, so they take the exemption and hit the dead end.

Node: 0.2.116 (also reproduced against a node built from `7f1f8387`). Seen in Safari 26.x; not browser-specific — it is sandbox flag inheritance, which all engines implement.

## Reproduce

1. Open any contract app **by pasting its URL** into a new tab: `http://<node>:7509/v1/contract/web/<KEY_A>/`. Works.
2. Inside that app, click a link of the form `<a href="/v1/contract/web/<KEY_B>/" target="_blank">`. Any app with such a link will do; I hit it with calliope's lightbox "open page" button (`href={`/v1/contract/web/${r}/`} target="_blank" rel="noopener"`), pointing at an image contract.
3. The new tab is blank.

Paste the *same* URL into a fresh tab and it loads correctly. Arrival path is the only variable — this is what makes it look like a browser/profile problem.

### Observed in the failing tab

The top-level document is healthy — the shell itself loads and sets the right `src`:

```json
{ "title": "Freenet",
  "origin": "http://10.44.101.1:7509",
  "len": 92487,
  "app": true,
  "src": "/v1/contract/web/DPZS3nmaS8XRqLufy3cq4t2DWkfG8k22gi8jcbykRzAH/?__sandbox=1#/b/freenet",
  "hash": "#/b/freenet" }
```

`document.baseURI` is correct, there is no `<base>` element and no `<meta http-equiv>` CSP. Yet the frame never loads:

```
Refused to load http://10.44.101.1:7509/v1/contract/web/DPZS…/?__sandbox=1#/b/freenet
  because it does not appear in the frame-src directive of the Content Security Policy.

Sandbox access violation: Blocked a frame at "http://10.44.101.1:7509" from accessing a
  frame at "null".  Both frames are sandboxed and lack the "allow-same-origin" flag.
```

[...]

PR #5100, opened 2026-08-02
fix(shell): open new tabs natively so target="_blank" works in Firefox

skandragon:
## Problem

Firefox users must right-click → "Open in new tab" for links that work on a
plain left-click in Chrome and Safari. Every cross-CONTRACT link is
same-ORIGIN (one gateway serves them all), so this hits every cross-app link
and thumbnail.

#5089 fixed a blank tab on same-origin `target="_blank"` (#5087) by routing the
click through the shell's `open_url` bridge. That put `window.open` inside a
`message` handler. Firefox's popup blocker gates on the dispatching event type
(`dom.popup_allowed_events`: click, auxclick, … never `message`), so the open is
refused and the click does nothing. Chrome and Safari keep working only because
they propagate user activation across the frame tree. Right-click bypasses JS
entirely, hence the workaround users found.

## Solution

Only a real user gesture, in the frame that received the click, opens a tab in
every browser. So stop intercepting new-window activations and let the browser
do it.

That is sound because the shell iframe now carries
`allow-popups-to-escape-sandbox`. The popup becomes a normal top-level document
at the node's real origin:

- the shell loads — its `frame-src 'self'` matches, where an opaque origin never
  could (the blank tab #5087 was about)
- `localStorage` works, so the hosted per-user access key resolves (#4645)
- a cross-origin destination sees a real `Origin`, not `null` (river#208) — the
  sandbox-inheriting popup is what broke CORS there in the first place

Modifier and middle-clicks are native again too, restoring background-tab and
new-window placement that the postMessage route collapsed into a plain
foreground tab. The `window.open` override is deleted: a native `window.open`
from a gesture now produces a working tab by itself.

**Still intercepted:** same-origin in-contract links (the `navigate` hop), and
cross-origin links with *no* target — those would navigate the app frame to a
foreign origin, which `frame-src 'self'` refuses, making the click silently
dead. They open via `window.open` from inside the click handler, where the
gesture is live.

**Companion guard (required, not optional).** Escaping the sandbox means a
contract can navigate a tab to anything the node serves. HTML sub-paths and
`?__sandbox=1` are already forced back through the shell, but a non-HTML asset
falls through to `variable_content`, and a contract-authored `evil.svg` served
as `image/svg+xml` executes script when it *is* the document — at the node's
origin, where it could read the hosted access key from `localStorage` or scrape
an auth token from a fetched shell page. `nosniff` does not help; the type is
genuinely scriptable. Top-level document loads of contract assets now carry
`Content-Security-Policy: sandbox`. Subresource fetches are untouched.

This reverses #1499's decision to withhold `allow-popups-to-escape-sandbox`.
That decision was correct when the only guard was the sandbox attribute; the
three server-side guards above are what make it safe now. **A reviewer other
than the author should confirm that reasoning** — it is the one part of this PR
where being wrong is expensive.

## Testing

- 435 unit tests pass. New: `web_subpages_sandboxes_top_level_asset_documents`
  pins the CSP guard on both the document and subresource paths.
- Five new browser specs, one per new-tab path, including that the same-origin
  new tab renders its own shell + app frame rather than a blank one.
- **Counter-check:** restoring the two asset files from `main` fails all five,
  in chromium, firefox and webkit. They are real regression tests, not
  assertions that happen to hold.
- Rust source-pin tests rewritten to the new contract; `window-open.spec.ts`
  deleted along with the override it tested. Net −450 lines.

### Known limitation

The suite is chromium-only, which is how #5087 shipped green in the first place.
I ran all three engines locally (30/30) and they pass, but the config is left as
chromium-only per maintainer preference.

More importantly: **the popup-blocker mechanism is not reproduced by the
harness.** With the pre-fix assets, Playwright's Firefox *does* open the popup
from the `message` handler, even with `dom.disable_open_during_load` restored to
its production default. Playwright's Firefox is a patched headless build. The
diagnosis above is the best explanation for the reported symptom — the timing
matches #5089 exactly, and right-click-bypasses-JS is the signature — but
confirming it needs a real Firefox. The fix removes the shell round-trip
entirely, so it holds regardless of which mechanism caused the failure.

## Fixes

Follow-up to #5087 / #5089. Needs a maintainer decision on the #1499 reversal.

### DIFF
--- a/.github/workflows/playwright-shell.yml
+++ b/.github/workflows/playwright-shell.yml
@@ -110,9 +110,16 @@ jobs:
           path: ~/.cache/ms-playwright
           key: playwright-${{ runner.os }}-${{ steps.pw.outputs.version }}
 
+      # All three engines, because this suite's subject is per-engine
+      # behaviour. #5087 (a blank new tab) reproduces only in WebKit, #5106 (a
+      # dead click) only in Firefox, and the #3818 escape guard passes
+      # VACUOUSLY in Chromium — there the escape dies a step earlier, so the
+      # assertion runs against an empty set and the header it exists to protect
+      # could be deleted with the suite still green. A chromium-only matrix is
+      # how #5087 shipped green in the first place.
       - name: Install Playwright browsers
         working-directory: crates/core/tests/playwright
-        run: npx playwright install --with-deps chromium
+        run: npx playwright install --with-deps chromium firefox webkit
 
       - name: Clean test directories
         run: rm -rf /tmp/freenet /tmp/freenet-* 2>/dev/null || true
--- a/crates/core/src/server/client_api.rs
+++ b/crates/core/src/server/client_api.rs
@@ -796,12 +796,46 @@ fn redirect_to_shell_sub_page(
 /// into the shell. `__sandbox` is a server-interpreted routing flag;
 /// `authToken` is the shell's auth credential and must only come from
 /// `AuthToken::generate()`, never from an attacker-controlled URL.
-fn is_sensitive_query_param(param: &str) -> bool {
+///
+/// The NAME is percent-decoded before the check. A raw prefix match is
+/// bypassable by encoding one character of the name — `authT%6Fken=evil`
+/// survives it, and `new URLSearchParams(...).get("authToken")` in the iframe
+/// then returns `evil`, which is exactly the webapp-reads-its-credential-from-
+/// `location.search` case this exists to prevent. Only the name is decoded; the
+/// value is forwarded byte-for-byte, since re-encoding it could break a signed
+/// or opaque app parameter.
+pub(super) fn is_sensitive_query_param(param: &str) -> bool {
     // Prefix-match so variants like `__sandbox_debug` or `authTokenExtra`
     // (from a future refactor or an adversarial URL) are also stripped.
-    // Matches the filter in `path_handlers::shell_page` that forwards
-    // query params into the iframe.
-    param.starts_with("__sandbox") || param.starts_with("authToken")
+    // Shared with `path_handlers::shell_page`, which forwards query params into
+    // the iframe — the two filters used to be separate copies of this rule.
+    let name = param.split('=').next().unwrap_or(param);
+    let decoded = percent_decode_ascii(name);
+    decoded.starts_with("__sandbox") || decoded.starts_with("authToken")
+}
+
+/// Percent-decodes the ASCII escapes in a query-parameter NAME so it can be
+/// compared against a literal. Deliberately minimal: invalid escapes are left
+/// as-is (they cannot form the names we are looking for), and non-ASCII bytes
+/// are passed through, because the only decision this feeds is a prefix match
+/// against two ASCII literals.
+fn percent_decode_ascii(s: &str) -> String {
+    let bytes = s.as_bytes();
+    let mut out = String::with_capacity(bytes.len());
+    let mut i = 0;
+    while i < bytes.len() {
+        if bytes[i] == b'%' && i + 2 < bytes.len() {
+            let hex = &s[i + 1..i + 3];
+            if let Ok(byte) = u8::from_str_radix(hex, 16) {
+                out.push(byte as char);
+                i += 3;
+                continue;
+            }
+        }
+        out.push(bytes[i] as char);
+        i += 1;
+    }
+    out
 }
 
 /// Returns true if a contract sub-path request is a top-level HTML
@@ -865,7 +899,7 @@ async fn serve_sandbox_response(
         // sandbox content off a top-level document, and that job is done by
         // redirecting at all, whatever the target.
         return redirect_to_shell_sub_page(&key, api_version, sub_path, query_string)
-            .or_else(|_| redirect_to_shell_root(&key, api_version, None));
+            .or_else(|_| redirect_to_shell_root(&key, api_version, query_string));
     }
 
     let contract_response = match path_handlers::serve_sandbox_content(
@@ -1919,6 +1953,40 @@ mod tests {
             Some(CONTRACT_CONTENT_SANDBOX_CSP),
             "the header is unconditional on this route"
         );
+    }
+
+    /// A percent-encoded parameter NAME must not slip the sensitive-param
+    /// filter. `authT%6Fken=evil` reads back as `authToken` from
+    /// `new URLSearchParams(location.search)` inside the iframe, which is
+    /// precisely the "webapp reads its credential from `location.search`" case
+    /// the filter exists for; a raw `starts_with` never sees it.
+    #[test]
+    fn sensitive_query_params_are_matched_after_decoding_the_name() {
+        for evil in [
+            "authT%6Fken=evil",
+            "%61uthToken=evil",
+            "%5F%5Fsandbox=1",
+            "__sandbo%78=1",
+            "%61uthTokenExtra=evil",
+        ] {
+            assert!(
+                is_sensitive_query_param(evil),
+                "{evil} must be stripped: the browser decodes the name before a \
+                 webapp reads it back"
+            );
+        }
+        // Ordinary app params are untouched, including ones that merely
+        // CONTAIN an escape in their value.
+        for ok in [
+            "invitation=abc",
+            "room=%2Fpath",
+            "q=authToken",
+            "myauthToken=x",
+        ] {
+            assert!(!is_sensitive_query_param(ok), "{ok} must be preserved");
+        }
+        // A malformed escape is not a decode, and must not become one.
+        assert!(!is_sensitive_query_param("auth%zzToken=x"));
     }
 
     /// Regression test for the sub-page loss that removing the `window.open`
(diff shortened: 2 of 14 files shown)


sanity:
Thanks for this — the diagnosis is right and you got there first. The conflicts are ours: #5107 (the revert of #5089) landed about three hours after you opened this, on the same files. The branch is rebased and force-pushed.

You asked for "a reviewer other than the author" on the #1499/#3818 reversal. That was the right thing to ask, and the answer is that the reversal is sound but the companion guard as written did not close the hole. Details below; everything is fixed on the branch.

## The rebase

Your three commits are squashed into one (still authored by you) so the conflicts could be resolved once against the final state rather than three times over. Nothing of yours was dropped. Two things of **ours** were replaced, and one comment of yours was corrected:

- **`interceptor_loopback_list_matches_shell_open_url_refusal`** (added by #5107) compared the interceptor's `isLoopbackHost` against the shell's `open_url` refusal. You delete that helper with the `window.open` override, so the parity half has no left-hand side. The shell-side floor survives as `shell_open_url_refusal_names_every_loopback_host` — that refusal is still a live control, since any contract can post `open_url` directly whether or not the injected interceptor ever does.
- **`navigation_interceptor_leaves_same_origin_target_blank_to_the_browser`** (also #5107) keeps both branch assertions and loses its whole-handler forward COUNT, which required exactly one `open_url` forward. There are now zero, and your file-wide "no `type: 'open_url'` anywhere" assertion subsumes it — mutation-confirmed during review.
- The `shell_page_*` comment: #5107 corrected the "#1499" citation — the flag was removed by **#3818** (`ec140e09c`); #1499 is the motivation, not the change. That correction is kept inside your inverted assertion.
- The interceptor's head comment gained one line from #5107's cost analysis, because your change fixes it too: an opaque-origin tab sends `Origin: null`, and `/permission/pending` answers 200 with an *empty list* rather than an error, so the permission prompt silently never appeared and nothing was logged.

Your commit message said four interceptor tests were `#[ignore]`d; it's three. The fourth (`..._routes_same_origin_target_blank_through_open_url`) pinned #5089, which main had already reverted, so it went away with the rebase rather than being retained. Message corrected.

## The security finding

`Sec-Fetch-Dest: document` is the wrong gate. It covers the pasted-URL shape, but not the way a contract actually reaches the node's origin once popups can escape:

1. From a click, `window.open('about:blank')`. The popup escapes, so it is a **top-level context with no sandboxing flags**, and its `about:blank` document inherits the opener's origin — so the contract can script it.
2. In that popup, `document.write` an `<iframe src=".../contract/web/KEY/…">`. That is a *nested navigable*: the request carries `Sec-Fetch-Dest: iframe`, so it skips the guard, and its parent has no sandboxing flags for it to inherit.
3. The contract's own bytes then execute at the node's **real origin** — `localStorage` (the hosted per-user access key) and same-origin `fetch` of any node route, another app's shell page and auth token included.

No scriptable SVG is needed: the contract's ordinary HTML is enough, so `?__sandbox=1` content was equally exposed — and that route had no guard at all, since it relied entirely on the iframe `sandbox` attribute for its opaque origin. That is precisely the escape #3818 removed the flag to prevent. Reproduced against a real node in **chromium, firefox and webkit**, each reporting the node's own origin with the access key readable.

**The fix**, and the smallest one I could find: serve the sandbox as a *header* on every response carrying contract bytes, so the opaque origin is decided by the server rather than by whichever context embeds them. `CONTRACT_CONTENT_SANDBOX_CSP` mirrors the iframe's own token list, so in-frame behaviour is unchanged — the effective policy is the intersection of the two, and a test keeps them literally equal. Top-level documents keep your stricter no-`allow-scripts` policy. A request with **no** `Sec-Fetch-Dest` now fails closed, which also covers browsers predating Fetch Metadata; before, `Sec-Fetch-Dest: image` from a non-browser client was enough to skip the header entirely.

The new spec — `a contract cannot reach a node-origin context by escaping the sandbox (#3818)` — fails in all three engines without that commit and passes with it. Chromium and WebKit additionally refuse the cross-document write once the opener is itself CSP-sandboxed, so the escape dies a step earlier there; Firefox allows the write and is stopped by the nested frame's opaque origin.

## What review found on top

Four independent blind reviewers ran over this. Two blocking findings, both in the new early-return logic, both reproduced in Chromium and Firefox against the real interceptor:

- **Right-click was intercepted.** `auxclick` fires for the secondary button as well as the middle one, and `preventDefault` there does not suppress the context menu — that comes from `mousedown`. So `e.button === 1` left right-click going through the handler: the user got the menu **and** a stray tab (cross-origin) or an app-frame navigation (same-origin). Right-clicking a link is exactly the workaround users adopted while `target="_blank"` was broken, so it is the last click that should misbehave. Now `e.button` truthiness, which also covers back/forward buttons — and `altKey`, since save-link is the same class of thing.
- **`target="_top"` and `_parent` became dead clicks**, and `target="_SELF"` sent a cross-origin click into the app frame. Hoisting the target check above origin classification assumed every non-`_self` target is a new-window activation. `_top`/`_parent` name an *ancestor*, which the sandbox forbids navigating, so the browser did nothing at all — where main opened a tab. And browsers match the reserved keywords ASCII-case-insensitively, so `_SELF` was read as a new-window request and fell through to a frame navigation `frame-src 'self'` then refused (Chromium replaces the app frame with its error page). Both now fall through to classification, and both have end-to-end specs.

And, from the same round:

- **`window.open(location.href)` lost its sub-path and query.** Deleting the override removed the client-side `__sandbox` strip, so such an open now reaches the server as a top-level document load of a `?__sandbox=1` URL — which redirected to the contract *root*, dropping the page and the app's own params (an invitation, say). It now redirects to the shell for the requested page, preserving the query minus `__sandbox`/`authToken`. Hosted-node-only, which is why it wasn't obvious: on loopback the deleted override fell back to native anyway.
- **Two injection gaps**, both pre-existing but now load-bearing. `sanitize_shell_sub_path` saw the path once-decoded (axum decodes a path param exactly once) while the browser decodes what we hand back again — so `%252e%252e` arrived as the literal `%2e%2e`, passed the dot-segment check, and normalized to `..` in the browser. Measured end to end: `/contract/web/KEYA/%252e%252e/KEYB/index.html` produced an iframe `data-src` the browser resolved to KEYB's page, so KEYB's app ran inside a shell holding KEYA's auth token. And `is_sensitive_query_param` prefix-matched raw text, so `authT%6Fken=evil` survived and `URLSearchParams.get("authToken")` in the iframe returned `evil`.
- **`serve_sandbox_response`'s CSP was the one built at runtime** rather than from a static string, and was applied with `if let Ok(...)` — silently skipped on a parse failure, leaving the app frame sandboxed by the attribute alone, i.e. the exact state in which the escape works again. Now falls back to the static directive and logs.
- **Error bodies are `text/html` at the node's origin** and reflect request-derived text ("Page not found: {page}"). Escaped now, with the retry page explicitly exempted so its meta-refresh tag still works — and `From<WebSocketApiError> for Response`, a second rendering path that skipped all of it, now delegates instead.
- One of my own fixes shipped a **panic**: the percent-decode helper sliced a `&str` by byte index, so `?%€=1` from any unauthenticated GET landed mid-character. Caught by the repo's rule-review job and by a reviewer independently, on the same commit.

## The Playwright config is no longer chromium-only

You flagged this as a known limitation, and it turned out to matter more than a limitation. The #3818 escape guard passes **vacuously** in Chromium — the escape is refused a step earlier there, so the assertion runs against an empty set and the header it exists to protect could be deleted with the suite still green. Only Firefox exercises it. Since #5087 reproduces only in WebKit and #5106 only in Firefox, a single-engine matrix was never going to hold this file honest — which is how #5087 shipped green in the first place. The config and the workflow now run all three.

## Verified

Against a real node in chromium, firefox and webkit: **48 browser specs (16 × 3 engines), 0 failed**, plus 445 unit tests. Counter-checked, not just asserted — reverting the sandbox header fails the escape spec in all three engines; reverting only `sandbox_csp_for_origin`'s prefix fails it on the `?__sandbox=1` probe alone; reverting the two interceptor fixes fails the right-click and `_top` specs in Chromium and Firefox.

Everything else you claimed holds: same-origin `target="_blank"` renders a working shell + app frame in WebKit (#5087), the programmatic `window.open` tab has a real origin and usable `localStorage` (#4645), and the cross-origin tab sees a real `Origin` (river#208). The harness still can't reproduce Firefox's popup blocker — but since the mechanism is disputed (#5107 measured the bridge's loopback refusal instead) and your fix removes the round-trip entirely, the comments now carry that uncertainty rather than picking a winner.

## Not done here

Three pre-existing interceptor edge cases the review surfaced are filed as #5123 rather than fixed here, since none is caused by this change: a bare `data-freenet-no-intercept` attribute has a falsy dataset value so it still intercepts; `tagName !== 'A'` misses SVG `<a>`; and WebKit does nothing for a middle-click on an untargeted cross-origin link. That issue also carries two cleanups worth a deliberate decision — the shell's `open_url` handler now has no producer in shipped code, and the "Open this app in a normal tab" panel is unreachable now that popups escape the sandbox. I corrected its rationale and kept it as a fail-safe rather than deleting ~60 lines of UI in this PR.

Fixes #5087.

[AI-assisted - Claude]

PR #5101, opened 2026-08-02
test: isolate playwright fdev config dir on macOS/Windows

skandragon:
## Problem

`fdev website` resolves its signing-key store through `dirs::config_dir()`,
which honours `XDG_CONFIG_HOME` on Linux only. On macOS it is
`$HOME/Library/Application Support`; on Windows, `%APPDATA%`.

`playwright_shell.rs` sets only `XDG_CONFIG_HOME`, so on macOS it:

1. wrote `smoke-fixture.toml` into the developer's **real** config dir — the
   exact thing its own comment promises not to touch ("so the signing key never
   touches the developer's real config dir"), and
2. failed on every subsequent run with `Error: Key 'smoke-fixture' already
   exists`, because `fdev website init` refuses to overwrite a signing key.

CI never saw either: it runs on Linux, where `XDG_CONFIG_HOME` is honoured.

## Solution

An `isolated_fdev()` helper that redirects `HOME` and `APPDATA` alongside
`XDG_CONFIG_HOME`, used by both `fdev` call sites.

Redirecting `HOME` wholesale is safe here specifically because `fdev website`
shells out to nothing — no cargo, no toolchain lookup — so a synthetic `HOME`
cannot break a child build.

## Testing

The harness itself is the test. Before: the second local run on macOS fails at
`website init`. After: `FREENET_PLAYWRIGHT=1 cargo test --release -p freenet
--features testing --test playwright_shell` passes repeatedly, and
`~/Library/Application Support/freenet/website-keys/` stays clean.

No behavior change on Linux, where `XDG_CONFIG_HOME` was already doing the work.

Split out of #5100 to keep one logical change per PR.

### DIFF
--- a/crates/core/tests/playwright_shell.rs
+++ b/crates/core/tests/playwright_shell.rs
@@ -142,12 +142,34 @@ fn strip_ansi(s: &str) -> String {
 // fdev website publish helpers
 // ---------------------------------------------------------------------------
 
+/// An `fdev` command whose signing-key store is redirected into `config_home`.
+///
+/// `fdev website` resolves its key directory through `dirs::config_dir()`, which
+/// honours `XDG_CONFIG_HOME` only on Linux — on macOS it is `$HOME/Library/
+/// Application Support`, and on Windows `%APPDATA%`. Setting `XDG_CONFIG_HOME`
+/// alone therefore left this test writing `smoke-fixture.toml` into the
+/// developer's REAL config dir on macOS: the run polluted a directory this
+/// harness promises not to touch, and the NEXT run failed with "Key
+/// 'smoke-fixture' already exists" because `website init` refuses to overwrite a
+/// signing key. Redirect the home/appdata vars too, so the isolation holds
+/// wherever the test runs and the suite stays re-runnable.
+///
+/// Safe to redirect wholesale because `fdev website` shells out to nothing —
+/// no cargo, no toolchain lookup — so a synthetic HOME can't break a child
+/// build.
+fn isolated_fdev(config_home: &Path) -> Command {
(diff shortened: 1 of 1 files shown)


sanity:
Reviewed at Light tier (test-only change, no production surface). Apologies for the delay getting to this; three weeks is longer than a one-file test-isolation fix should sit.

The fix is correct and I verified the load-bearing claim rather than taking it from the doc comment: `fdev website` resolves its key directory through `dirs::config_dir()` at `crates/fdev/src/website.rs:88`, and `grep 'Command::new\|std::process'` over `website.rs` returns nothing, so it genuinely spawns no child. Redirecting `HOME` wholesale for that child cannot break a toolchain lookup that never happens. The macOS half of this is a real fix for a real annoyance.

One caveat worth recording so nobody later reads the title as a guarantee.

**The Windows leg is very likely a no-op.** `dirs` 6.0.0 resolves `config_dir()` on Windows through `dirs-sys` 0.5.0's `known_folder(FOLDERID_RoamingAppData)`, which calls `SHGetKnownFolderPath`. That reads the per-user Shell Folders registry value, not the calling process's environment block, so `.env("APPDATA", config_home)` at `playwright_shell.rs:163` does not redirect it. The Unix backend does consult `$HOME` first, which is why the macOS half works. Setting the variable is harmless and is right if `fdev` ever reads `%APPDATA%` directly, so I am not asking you to change anything. It just means a local Windows run would still write into the real `%APPDATA%\freenet\website-keys\`. Not exercised in CI today, since `playwright-shell.yml` only runs on Linux.

Merging. Thanks for catching this, and for the unusually clear WHY in the helper's doc comment.

[AI-assisted - Claude]

Issue #4914, opened 2026-07-22
chore: 25 latent clippy lints under rust 1.94 (--all-targets) will block CI on the next toolchain bump


sanity:
## Problem

`cargo clippy --all-targets --features trace,websocket,redb,wasmtime-backend,testing -- -D warnings` fails on `main` under the toolchain the repo itself pins — **25 errors** across 11 files. CI is green today only because it doesn't run this command:

```yaml
# .github/workflows/ci.yml:407-410, 419
- uses: dtolnay/rust-toolchain@stable
  with:
    toolchain: 1.93.0
...
run: cargo clippy --locked -- -D warnings
```

CI's `clippy_check` job pins **1.93.0** and never passes `--all-targets`. But `rust-toolchain.toml` (repo root) pins:

```toml
[toolchain]
channel = "1.94.0"
```

So any contributor or agent running the ambient `cargo clippy --all-targets` (which uses the 1.94.0 toolchain via the pin) hits a wall of pre-existing failures unrelated to their change — and the moment CI's `clippy_check` toolchain is bumped to match `rust-toolchain.toml` (1.94.0), or `--all-targets` is added to close the gap `#4818` describes, all 25 go red and block every in-flight PR at once.

Related: **#4818** already tracks the narrower "CI never lints test code" gap with one example (`contract_store.rs:1230`). That same line is one of the 25 here — this issue supersedes it with the full current inventory plus the toolchain-version-pin risk, reproduced fresh under 1.94.0 today.

## Reproduction

```bash
rustup install 1.94.0   # or just build in a worktree, which picks up rust-toolchain.toml
cargo clippy --all-targets --features trace,websocket,redb,wasmtime-backend,testing -- -D warnings
```

Confirmed with `rustc 1.94.0 (4a4ef493e 2026-03-02)`.

## Full inventory (25 errors, 11 files)

| File | Count | Lints |
|---|---|---|
| `crates/core/src/operations/update/op_ctx_task.rs` | 8 | 6× `let_underscore_must_use` (lines 4190, 4220, 4272, 4349, 4397, 5383), 2× `wildcard_enum_match_arm`-style "wildcard match will also match any future added variants" (3773, 4076) |
| `crates/core/src/client_events/websocket.rs` | 4 | 3× `let_underscore_must_use` (2702, 2715, 2738), 1× wildcard match (2754) |
| `crates/core/src/node.rs` | 3 | 3× wildcard match (4640, 4755, 4884) |
[...]

PR #5103, opened 2026-08-02
fix(update): stop spending GitHub REST quota to check for releases

### LINKED ISSUE #5102: Auto-update polls api.github.com on a timer and exhausts the shared 60 req/hr/IP budget; never backs off on 403/429
## Problem

A user reported (Matrix, 2026-08-02) that auto-update failed with "too many requests" — GitHub was rate-limiting them — and they had to update by hand.

Every Freenet node's update path polls **`https://api.github.com/repos/freenet/freenet-core/releases/latest`**. That endpoint allows **60 unauthenticated requests per hour per source IP**, and the budget is **collective**: it is shared by every Freenet node behind the same NAT/CGNAT/VPN egress *and* by every unrelated tool on that IP.

Verified against live GitHub:

```
$ curl -sI -H 'User-Agent: freenet-updater' https://api.github.com/repos/freenet/freenet-core/releases/latest
x-ratelimit-limit: 60
x-ratelimit-remaining: 59
x-ratelimit-resource: core
```

### Where the requests come from

| # | Call site | Cadence |
|---|---|---|
| 1 | `auto_update::startup_update_check` | once per node process start — explicitly *"decoupled from the backoff / failure-count state"*, so it is unconditional |
| 2 | version-mismatch loop (`check_if_update_available`) | 60s tick, persisted exponential backoff 1 min → 1 h |
| 3 | periodic re-poll (`freenet.rs` priority 4) | every 6 h ± 25 % |
| 4 | `freenet update` → `get_latest_release` | systemd `ExecStopPost` runs it on **every** non-clean exit: `case "$EXIT_STATUS" in 0\|43) ;; *) freenet update` — so once per crash and once per exit-42 update cycle |
| 5 | dashboard JS (`dashboard.js::checkForUpdate`) | 12 h localStorage TTL on success — but **only success is cached**, so once GitHub starts refusing, every page reload issues another request |

### Why "1 poll/hour/node" still exhausted a shared 60/hour

We cut **1–4 releases per day**, so a peer is in version-mismatch (trigger 2) more or less permanently, sitting at its 1 h max-backoff floor. Two on-disk token buckets (`GITHUB_POLL_BUCKET_CAPACITY = 8`, `GITHUB_POLL_REFILL_SECS = 600`, one for the node path and one for the installer) bound a *single* node to **16 requests burst + 12/hour sustained** — i.e. one node may consume up to **20 % of an entire IP's hourly GitHub budget**, and 27 % of it instantly. Five nodes behind one CGNAT exhausts it; fewer than that if anything else on the IP touches the GitHub API.

The reasoning that missed this is still in the source. `freenet.rs`:

> At 6h ± 25% the minimum interval is ~4.5h, far under GitHub's unauthenticated 60 req/hr/IP limit.

and `auto_update.rs`'s own test:

> We assert the minimum interval is >= 1 hour, i.e. <= 1 periodic request per hour — a >=60x safety margin.

Both compute a **per-node** margin against a **per-IP, shared** budget. A per-node margin says nothing about how much of a shared budget the fleet consumes.

### We never back off when told to

Nothing in the update path reads `x-ratelimit-remaining`, `x-ratelimit-reset`, or `Retry-After`. On a 403/429:

- the node treats it as a generic network error and merely grows its own backoff;
- the installer does `anyhow::bail!("GitHub API returned error: {} {}", status, body)`.

So we keep knocking through the rate-limit window — which is what escalates a primary (hourly) limit toward a longer secondary block, and is the specific escalation-to-IP-block risk here. The installer's bail is also the likely source of the user-visible string: `reqwest`'s `StatusCode` renders 429 as `429 Too Many Requests`.

**Honesty note:** I could not retrieve the reporter's exact error text or status code (their message is not in the last 400 messages of #freenet-locutus, so it was a DM or another room). The 403-vs-429 attribution above is inferred from the reported wording plus the code paths that can emit it, not confirmed from their logs.

## Fix

`https://github.com/freenet/freenet-core/releases/latest` answers with a `302` whose `Location` carries the tag, is served by the web front end, and **does not touch the REST budget**. Measured:

```
OLD endpoint (api.github.com/releases/latest): core.used 1 -> 2  (delta=1)
NEW endpoint (github.com/releases/latest):     core.used 2 -> 2  (delta=0)
```

1. Detection (triggers 1–3) resolves the tag from that redirect: **zero REST quota**.
2. `freenet update` probes the tag the same way and only spends a REST request — now by exact tag, not `latest` — when an install is actually going ahead. "Already up to date" (every `ExecStopPost` after a crash) and `--check` become free.
3. Honour `Retry-After` / `x-ratelimit-reset` on 403/429 via a cooldown persisted in `state_dir()` and **shared by the node and installer paths**, so a limited IP goes quiet instead of continuing to knock.
4. Replace the bare "429 Too Many Requests" with a message naming the shared-IP cause, saying nothing is broken, and giving the manual download route.
5. Negative-cache the dashboard's check so a rate-limited browser stops re-requesting on every reload. (It must stay on `api.github.com` — it runs in a browser and the redirect endpoint sends no CORS headers.)

[AI-assisted - Claude]

sanity:
## Problem

A user reported (Matrix, 2026-08-02) that auto-update failed with "too many requests" and they had to update by hand.

Every node's update path polled **`api.github.com/repos/freenet/freenet-core/releases/latest`**, which allows **60 unauthenticated requests per hour per source IP**. That budget is **collective** — shared by every Freenet node behind the same NAT/CGNAT/VPN egress *and* by everything else on that IP.

Five call sites reach it: the startup check (unconditional, explicitly *"decoupled from the backoff"*), the version-mismatch loop (60s tick, backoff 1 min → 1 h), the 6 h periodic re-poll, `freenet update` (which systemd's `ExecStopPost` runs on **every** non-clean exit — so once per crash), and the dashboard JS.

The bound we had was reasoned **per node**, against a budget that is **per IP**. The source says so in as many words:

> At 6h ± 25% the minimum interval is ~4.5h, far under GitHub's unauthenticated 60 req/hr/IP limit.

and its test:

> a >=60x safety margin

The two on-disk token buckets cap one node at **16 burst + 12/hour sustained** — up to **20 % of an entire IP's hourly allowance from a single peer**. We cut 1–4 releases/day, so the mismatch trigger is armed almost permanently and peers sat at that ceiling continuously. A handful behind one CGNAT exhausts 60/hr.

Nothing read `Retry-After` / `x-ratelimit-reset` either, so on 403/429 we kept knocking through the window — the specific behaviour that escalates a primary limit toward a longer secondary block, which is the IP-block risk. The installer's `anyhow::bail!` is the likely source of the reported string (`reqwest` renders 429 as `429 Too Many Requests`).

## Approach

`https://github.com/<repo>/releases/latest` returns a `302` whose `Location` carries the tag, is served by the web front end, and does **not** touch the REST budget. Measured against live GitHub:

```
OLD (api.github.com/releases/latest): core.used 1 -> 2  (delta=1)
NEW (github.com/releases/latest):     core.used 2 -> 2  (delta=0)
```

1. **Detection** reads the tag from that redirect — zero REST quota on the timer-driven path.
2. **`freenet update`** probes the same way and spends a REST request only when an install is actually going ahead, requested **by exact tag** rather than `latest` (so the assets we install match the version we decided on). "Already up to date" and `--check` are now free. Verified end-to-end: `freenet update --check` consumes 0 REST requests.
3. **Cooldown on 403/429**, persisted in `state_dir()` from `Retry-After` / `x-ratelimit-reset`. Deliberately **shared** by the node and installer paths — unlike the token buckets, which are separate on purpose — because a rate limit is a property of the IP, not of a code path. Never shortened by a later writer; clamped so a skewed clock or torn write cannot disable updates indefinitely.
4. **Legible failure**: names the shared-IP cause, says nothing is broken, gives the manual download route.
5. **Dashboard** negative-caches failures so a limited browser stops re-requesting on every reload. It must stay on `api.github.com` (browser + no CORS headers on the redirect), so backing off is the only lever there.

Deliberately **not** changed: the poll cadence. The endpoint switch removes the quota cost entirely; changing intervals too would be an unevidenced behaviour change on top of a bug fix.

## Testing

13 new tests. The two source pins were **mutation-tested** — each confirmed to fail when the fix is reverted:

- reverting the const to `api.github.com` → `detection_endpoint_does_not_use_the_rest_api` FAILS
- restoring the eager asset fetch → `already_up_to_date_and_check_never_reach_the_rest_api` FAILS

Covered: `Location` parsing (absolute/relative/trailing-slash/query/no-`v`, and rejection of non-release redirects so a bogus tag can't read as "newer" and trigger a pointless exit-42); 403 **and** 429 both recognised; `Retry-After` preferred over `x-ratelimit-reset`; absolute-reset → delta conversion (treating it as a delta would give a ~56-year cooldown); clamping at both ends; cooldown expiry, never-shortened, bounded default, and corrupt/skewed input failing open.

Also updated the stale `test_get_latest_release_consults_install_bucket` pin for the rename, and corrected the "60x safety margin" comment that encoded the wrong denominator.

`cargo fmt`, `cargo clippy -p freenet --bins --lib -- -D warnings`, `cargo test -p freenet --bin freenet` (343 pass), home_page lib tests, and the asset `npm run lint` + `npm test` gates all clean. The 10 pre-existing `--all-targets` clippy errors are #4914 and reproduce identically on `main`.

Closes #5102

[AI-assisted - Claude]

---

## Review round (commits 2–3)

Full-tier review surfaced three findings, all fixed here:

1. **`--force` was refused by our own cached cooldown** (`0f7a9a1`). The check ran unconditionally, so an operator running `freenet update --force` after a rate-limit episode was blocked by our stale marker for up to 6 h — even once GitHub had recovered — and told "already up to date". That silently broke the escape hatch two operator-facing messages point at ("Run `freenet update --force` to override", and the crash-loop rollback's advice). `--force` now waives the *cache*; a **live** 403/429 still refuses and still re-arms the cooldown.
2. **Response handling was inline on the network path, so untestable** (`0f7a9a1`). Extracted as a pure `classify_probe_response`. This also hardened two cases: the tag is now read from `Location` for *any* non-rate-limited status rather than only `302` (every redirect code carries the same header; pinning to one would turn a harmless GitHub-side change into a fleet-wide update outage), and 403/429 classify as rate-limited *before* `Location` is consulted.
3. **Nothing prevented a supervisor gaining `--force`** (`7693440`). Since `--force` now waives both local limiters, that is only safe while it stays operator-only — a supervisor with `--force` would let every crash in a crash-loop knock at GitHub unbounded. Added a mutation-tested pin over the systemd unit, macOS wrapper, and `spawn_update_command`.

Also verified empirically during review: **prerelease semantics are unchanged**. On `electron/electron`, whose newest release *is* a prerelease, both the REST and web `releases/latest` endpoints return the same stable tag.

The full consolidated review is posted as a PR comment, including an honest note on how it was run: six independent reviewers were spawned and none returned, so the lenses were run serially in-session (the documented fallback). That is a weaker signal than blind independent review — an external pass before merge would still add value.

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -570,16 +570,27 @@ pub async fn check_if_update_available(current_version: &str) -> UpdateCheckResu
 ///   asked for, or when this response is itself a `403`/`429`. Callers must
 ///   treat this as "we did not learn anything", never as "no update exists".
 /// * Any transport / parse failure otherwise.
-pub(crate) async fn fetch_latest_release_tag() -> Result<String> {
+pub(crate) async fn fetch_latest_release_tag(bypass_cached_cooldown: bool) -> Result<String> {
     // Honour a cooldown GitHub asked for earlier (possibly in a previous
     // process). Checked BEFORE the request so a limited IP goes quiet instead of
     // continuing to knock — continuing to knock while limited is what escalates a
     // soft per-hour limit into a longer secondary block.
-    if let Some(remaining) = github_cooldown_remaining() {
-        return Err(GithubRateLimitedError {
-            retry_after: Some(remaining),
+    //
+    // `bypass_cached_cooldown` is for `freenet update --force` ONLY. The stored
+    // deadline is our *cached belief* about GitHub, and it can be stale — GitHub
+    // may have reset early, or the limit may have been another client's doing.
+    // `--force` is the documented operator escape hatch (both the token-bucket
+    // message and the crash-loop rollback advice tell users to run it), so it
+    // must not be blocked by our own cache. It does NOT bypass a LIVE 403/429:
+    // the request below still runs, and a real refusal is still honoured and
+    // still re-records the cooldown.
+    if !bypass_cached_cooldown {
+        if let Some(remaining) = github_cooldown_remaining() {
+            return Err(GithubRateLimitedError {
+                retry_after: Some(remaining),
+            }
+            .into());
         }
-        .into());
     }
 
     let client = reqwest::Client::builder()
@@ -593,36 +604,69 @@ pub(crate) async fn fetch_latest_release_tag() -> Result<String> {
 
     let response = client.get(GITHUB_LATEST_REDIRECT_URL).send().await?;
     let status = response.status();
-
-    if is_rate_limited_status(status) {
-        // GitHub told us to stop. Persist the reset instant it handed back so
-        // every poll path on this machine — including the short-lived `freenet
-        // update` processes the supervisor spawns — stays quiet until then.
-        return Err(note_rate_limited_response(|name| {
-            response
-                .headers()
-                .get(name)
-                .and_then(|v| v.to_str().ok())
-                .map(str::to_owned)
-        })
-        .into());
-    }
-
     let location = response
         .headers()
         .get(reqwest::header::LOCATION)
         .and_then(|v| v.to_str().ok())
         .map(str::to_owned);
 
-    match location
-        .as_deref()
-        .and_then(parse_tag_from_release_location)
-    {
-        Some(tag) => Ok(tag),
-        None => anyhow::bail!(
+    match classify_probe_response(status, location.as_deref()) {
+        ProbeOutcome::Tag(tag) => Ok(tag),
+        ProbeOutcome::RateLimited => {
+            // GitHub told us to stop. Persist the reset instant it handed back so
+            // every poll path on this machine — including the short-lived
+            // `freenet update` processes the supervisor spawns — stays quiet
+            // until then.
+            Err(note_rate_limited_response(|name| {
+                response
+                    .headers()
+                    .get(name)
+                    .and_then(|v| v.to_str().ok())
+                    .map(str::to_owned)
+            })
+            .into())
+        }
+        ProbeOutcome::Unusable => anyhow::bail!(
             "GitHub returned {status} with no parseable release tag in Location ({})",
             location.as_deref().unwrap_or("<absent>")
         ),
+    }
+}
+
+/// What a release-probe response means.
+#[derive(Debug, PartialEq)]
+pub(crate) enum ProbeOutcome {
+    /// A usable release tag (leading `v` already stripped).
+    Tag(String),
+    /// GitHub is rate-limiting us; the caller records the cooldown.
+    RateLimited,
+    /// Anything else — no usable tag. Deliberately NOT a version, so a captive
+    /// portal, proxy interstitial, or an unexpected GitHub response can never be
+    /// mistaken for a release and trigger a pointless exit-42 update cycle.
+    Unusable,
+}
+
+/// Classify a `releases/latest` probe response from its status and `Location`.
+///
+/// Pure, so the whole decision is unit-testable without a network round trip —
+/// the status/redirect handling is the part most likely to break silently if
+/// GitHub's response shape ever shifts.
+///
+/// The tag is read from `Location` for ANY non-rate-limited status that supplies
+/// a parseable one, rather than only for `302`. GitHub sends `302` today, but
+/// `301`/`303`/`307`/`308` are all legitimate redirect responses and all carry
+/// the same header; pinning to one code would turn a harmless server-side change
+/// into a fleet-wide update outage.
+pub(crate) fn classify_probe_response(
+    status: reqwest::StatusCode,
+    location: Option<&str>,
+) -> ProbeOutcome {
+    if is_rate_limited_status(status) {
+        return ProbeOutcome::RateLimited;
+    }
+    match location.and_then(parse_tag_from_release_location) {
+        Some(tag) => ProbeOutcome::Tag(tag),
+        None => ProbeOutcome::Unusable,
     }
 }
 
@@ -640,7 +684,9 @@ async fn get_latest_version() -> Result<String> {
         return Err(RateLimitedError.into());
     }
 
-    fetch_latest_release_tag().await
+    // Never bypasses the cooldown: this is the automated path, and the whole
+    // point of the cooldown is to keep it quiet while the IP is limited.
+    fetch_latest_release_tag(false).await
 }
 
 /// Get the state directory for update tracking files.
@@ -2036,7 +2082,7 @@ mod tests {
             .split_once("async fn probe_latest_tag(")
             .expect("probe_latest_tag definition not found");
         let (head, _) = body
-            .split_once("fetch_latest_release_tag()")
+            .split_once("fetch_latest_release_tag(")
             .expect("tag fetch call not found");
         assert!(
             head.contains("try_consume_install_poll()"),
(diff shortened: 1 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #5103

### Summary
- **PR Title:** fix(update): stop spending GitHub REST quota to check for releases
- **Type:** fix
- **CI Status:** was fully green (16 success / 2 skipped) on `e01b1226`; re-running on the current head
- **Linked Issues:** #5102
- **Review tier:** **Full** — touches deploy/release/update infrastructure, a named high-risk surface in `.claude/rules/pr-quality.md`
- **HEAD SHA reviewed:** `76934401b874e23113d65d74a44a3d4a6928ffef`

**How this review was run — please read.** Six independent reviewers were spawned (the four `freenet:*` lenses plus two change-specific ones: an update-liveness/brick-risk lens and an HTTP/external-contract lens). **None of them returned within ~50 minutes**, including after an explicit status request. Rather than block or skip the independent pass, the lenses were run **serially in-session**, which is the documented fallback in both the `pr-review` skill and `~/.claude/rules/multi-model-review.md`. No external (non-Claude) model was run: per the 2026-07-26 change to the multi-model rule, Codex is opt-in only and was not requested.

That is a genuinely weaker signal than blind, independent reviewers — the author model reviewed its own work, so blind spots are correlated. **Findings were nonetheless real and material**, and are recorded below with the fixes. A human or external pass before merge would still add value.

---

### Code-First Analysis
**Independent Understanding:** Detection of "is there a newer release" moves from `api.github.com/repos/.../releases/latest` (REST JSON, 60 req/hr/IP) to `github.com/.../releases/latest`, parsing the tag out of the `302`'s `Location`. `freenet update` probes the same way and only spends a REST request — by exact tag — once an install is committed to. 403/429 persist a cooldown shared by node and installer.

**Stated Intent:** Matches.

**Gaps found:** One, now fixed — see Must Fix #1. The description claimed `--force` bypasses "our limiter", but the cooldown check ran unconditionally, so `--force` was *also* blocked by it.

---

### Testing Assessment
**Coverage Level:** adequate after fixes (17 new tests; 348 pass in the binary suite)

| Unit | ✅ | tag parsing, status classification, header parsing, cooldown arithmetic |
| Integration | ⚠️ | no test drives a real HTTP response; mitigated by extracting `classify_probe_response` as a pure function (Must Fix #2) |
| Simulation | N/A | no routing/topology surface |
| E2E | ✅ | verified manually against live GitHub, see below |

**Regression tests:** present and **mutation-tested** — each confirmed to fail when the fix is reverted:
- revert the endpoint const → `detection_endpoint_does_not_use_the_rest_api` FAILS
- restore the eager asset fetch → `already_up_to_date_and_check_never_reach_the_rest_api` FAILS
- add `--force` to `spawn_update_command` → `no_supervisor_path_invokes_update_with_force` FAILS

**Empirical verification (live GitHub):**
- `freenet update --check` consumes **0** REST requests (`core.used` unchanged); the old endpoint consumes 1.
- The binary correctly resolves `0.2.118` from the `Location` header end to end.
- **Prerelease semantics are identical.** Tested on `electron/electron`, whose newest release *is* a prerelease (`v44.0.0-alpha.8`): both the REST and web `releases/latest` return `v43.2.0`. So the endpoint switch cannot make nodes chase a prerelease. (freenet-core has 0 prereleases/drafts in its last 60 releases regardless.)

---

### Skeptical Findings
**Risk Level:** medium before fixes, low after

| `--force` blocked by our own cached cooldown | **High** | `update.rs` probe path | Fixed in `0f7a9a1` |
| Response handling inline on the network path, untestable | Medium | `auto_update.rs` | Fixed in `0f7a9a1` — extracted `classify_probe_response` |
| Only `302` would have been handled | Medium | `auto_update.rs` | Fixed — tag now read from `Location` for any non-rate-limited status |
| Nothing prevented a supervisor gaining `--force` | Medium | supervisors | Fixed in `7693440` — pin test |

**Considered and deliberately kept:**
- **A REST 403 on the asset fetch also suppresses the (now free) detection poll.** Over-conservative, but correct: we cannot install while REST is limited, so suppressing detection prevents a detect → exit-42 → install-fails restart cycle.
- **Cooldown fails *open* when the state dir is unresolvable**, while the token buckets fail *closed*. Consistent overall: on such a machine detection is already disabled by the bucket, so failing closed here too would only add a second permanent wedge.
- **`Retry-After` is parsed as delta-seconds only**, not the RFC 9110 HTTP-date form. GitHub documents delta-seconds for secondary limits; an unparseable value degrades to `x-ratelimit-reset` and then to a bounded default, so the worst case is a slightly less precise cooldown.
- **Cadence is unchanged.** The endpoint switch removes the quota cost entirely; changing intervals as well would be an unevidenced behaviour change riding along on a bug fix.

**Bug-pattern check** (`.claude/rules/bug-prevention-patterns.md`): no `biased;` select, no `GlobalExecutor::spawn`, no bounded-channel `.send().await`, no protocol-enum or wire-format change, no non-string map keys, no telemetry counter mirroring. Cleanup/GC exemptions are time-bounded (the cooldown is clamped to 6 h and re-clamped on read, so a skewed clock or torn write cannot create a permanent GC-style blind spot) — this is the rule that most directly applies here, and it is satisfied and tested.

---

### Big Picture Assessment
**Goal Alignment:** yes — addresses the reported symptom and its root cause.
**Anti-Patterns Detected:** none.
**Removed Code Concerns:** none. The #4073 token buckets and the #3934 lockout are preserved; the stale `test_get_latest_release_consults_install_bucket` pin was updated for the rename rather than deleted.
**Scope Assessment:** focused. The dashboard JS change is the same defect class (a poll that kept knocking while refused) and is ~30 lines; the by-exact-tag switch is required by the two-stage split and removes a latent race where `latest` could move between probe and install.
**Other instances of the pattern:** `scripts/install.sh` (one-shot manual install), `scripts/gateway-auto-update.sh` (prefers the authenticated `gh` CLI, 5000/hr), and `crates/release-agent` (server-side, on-demand) also use the REST API. None is timer-driven on end-user machines, so none is left in the state this PR fixes.

**Improvement worth noting:** a GitHub rate limit previously propagated as a generic error out of `freenet update`; it now exits `ALREADY_UP_TO_DATE`, so an external rate limit no longer counts toward the `MAX_UPDATE_FAILURES` lockout.

---

### Documentation
- Code docs: complete — every new public item has a rustdoc explaining *why*.
- Architecture/user docs: no stale references found (`grep api.github.com` over `docs/`, `AGENTS.md`, `.claude/rules/` returns only an unrelated `api.github.com/meta` CIDR note).
- Corrected in this PR: the "`>=60x safety margin`" test comment, which measured a per-node margin against a per-IP shared budget — the reasoning error behind the bug.

---

### Recommendations

#### Must Fix (Blocking)
1. ~~`--force` was refused by our own cached cooldown, silently breaking the escape hatch two operator messages point at.~~ **Fixed — `0f7a9a1`.**
2. ~~Status/redirect handling was inline on the network path and untestable.~~ **Fixed — `0f7a9a1`, extracted as pure `classify_probe_response` with tests.**

#### Should Fix (Important)
3. ~~Nothing prevented a future change from giving a supervisor `--force`, which would remove every local bound.~~ **Fixed — `7693440`, mutation-tested pin.**

#### Consider (Suggestions)
4. The `github.com` web endpoint may have its own abuse limits, undocumented and not signalled via `x-ratelimit-*`. Our cadence (≤1/hr/node worst case) is far below anything plausible, and a 403/429 from it is handled identically — but it is an assumption about an unpublished limit, worth remembering if update reliability ever regresses.
5. Once this ships, it would be worth confirming from telemetry that update-check failures actually drop, rather than assuming.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `76934401b874e23113d65d74a44a3d4a6928ffef`

All three findings are fixed and mutation-tested. The caveat above stands: the independent reviewers did not return, so this is a self-review with distinct serial lenses, not blind independent review.

[AI-assisted - Claude]

Issue #5096, opened 2026-08-02
ci: fix:-needs-a-test gate can't see .test.mjs suites, so browser-asset fixes are wrongly rejected


skandragon:
Hit while fixing #5043 (PR #5094). Filing rather than fixing inline, because changing a test-enforcement gate inside the PR that gate is blocking is the CI-chasing pattern this repo watches for.

## Problem

`ci.yml`'s rule 5 rejects a `fix:` PR that adds no new test. It counts two things:

- Rust test attributes (`TEST_ATTR_RE` — `#[test]`, `#[tokio::test]`, `#[freenet_test]`, …)
- `check ` assertion lines in added `**/*_test.sh` files (`NEW_SHELL_TESTS`)

The browser assets under `crates/core/src/server/**/assets/` are tested by neither. Their regression tests are Node self-tests — six suites today, 145 `check(` assertions total:

```
base58_encode.test.mjs               15
connecting_recovery.test.mjs         12
dashboard_refresh.test.mjs           32
notify_sw.test.mjs                   16
shell_bridge_notifications.test.mjs  49
shell_bridge_reload.test.mjs         21
```

These are real tests and CI really runs them — `lint-assets.yml` fires on `pull_request` and runs `npm test`, which exits non-zero on any failed assertion. But they score zero against rule 5, so **any `fix:` PR whose regression test is a `.test.mjs` case is rejected** and has to either claim `test-exempt` (false — the fix is tested) or add a Rust test purely to satisfy the counter.

This is precisely the gap the shell carve-out was added for, one language over. Its own comment says so:

> Some fixes are to shell tooling (e.g. scripts/release-agent/*.sh), whose regression tests are shell `*_test.sh` self-tests run in CI, not Rust `#[test]`s. […] Without this, a genuine shell regression test scores zero against TEST_ATTR_RE (Rust-only) and the fix is wrongly rejected — the gap that hid the announce-to-river fix's own test.

## Suggested fix

Mirror `NEW_SHELL_TESTS` for the Node suites — the assertion helper is `check(...)` rather than `check ...`:

```bash
NEW_MJS_TESTS=$(git diff "$BASE"..."$HEAD" -- '**/*.test.mjs' \
  | grep -E '^\+' | grep -vF '+++' \
  | grep -cE '^\+[[:space:]]*check\(' || true)
```

then include it in the final condition alongside `NEW_TESTS` and `NEW_SHELL_TESTS`.

[...]

Issue #5076, opened 2026-07-30
Source-scrape pins can silently stop guarding anything while staying green (two instances in #5066)


sanity:
## The hazard

This codebase relies heavily on **source-scrape pins** — tests that `include_str!` a source file, slice a window between two anchors, and assert the window contains (or does not contain) an identifier. They guard wiring that no type or unit test can: "this arm delegates to the shared helper", "this call site routes through the instrumented constructor", "this emitter names its own arm".

They have a failure mode that is invisible by construction: **a pin can stop guarding anything while staying green.** Two independent instances surfaced within a few hours on one PR (#5066).

## Instance 1 — the scanned window silently widened

`node.rs::summaries_arm_uses_semantic_staleness_probe_pin` scopes itself as "the `Summaries` arm" by slicing from `InterestMessage::Summaries` to `InterestMessage::ChangeInterests` — i.e. anchored on *whatever arm comes next*.

Inserting two new arms between them widened the window to span the new code. Because the new arms happen to call the same functions, the pin's positive assertions would have kept passing **with those calls deleted from the arm it was written to guard**.

Nothing failed. The pin reported success about a property it no longer tested.

## Instance 2 — the scrape's scope silently emptied

`outbound_message_mix.rs::summaries_emitter_sites_are_pinned` finds emitter sites by matching `emitter: SummariesEmitter::` and scopes candidate files to those constructing `InterestMessage::Summaries { .. }`.

A refactor moved construction behind a helper, so the arm name became a call *argument* rather than a struct *field*, and `operations/update.rs` no longer constructed the variant at all. That file dropped out of the pin's scope entirely and its two emitters stopped being pinned — at exactly the moment the refactor made the attribution more fragile, not less.

Again: green.

## A third, related instance

The same PR had a pin whose needle matched its own *production comment*: the arm's explanatory comment named the identifier the pin searched for, so deleting the real call left the pin satisfied by prose. That one is guarded against for the assertion's own source (`concat!("record_summary", "_comparison")`) but nobody had applied the same thought to the **scanned window**.

## Proposed mitigation

**A scrape that matches zero — or an unexpected number of — things must FAIL, not pass.** This single property catches all three instances:

1. **Assert scope is non-empty.** A file walk that finds no files, a window slice that yields no anchor, a needle that matches nothing: all should be hard errors. Instance 2 was precisely "the scan found nothing and reported success".
[...]

Issue #4846, opened 2026-07-16
Harden shell open_url handler: complete loopback allow-list + rate-limit


sanity:
## Problem

The shell bridge's `open_url` handler (crates/core/src/server/path_handlers/assets/shell_bridge.js) is the primary gate for URLs a sandboxed contract asks the shell to open (a malicious contract can post `open_url` directly). Two pre-existing LOW-severity gaps were surfaced by the multi-model review of PR #4844:

1. **Incomplete loopback allow-list.** It blocks `localhost` / `127.0.0.1` / `::1` / `0.0.0.0`. Still reachable: other 127.0.0.0/8 (`127.0.0.2`), and IPv4-mapped IPv6 (`[::ffff:127.0.0.1]`, which serializes as `[::ffff:7f00:1]`). (Decimal/hex/octal IPv4 like `2130706433` / `0x7f000001` are normalized to dotted-decimal by `URL.hostname`, so those are already covered.) The bracketed-`[::1]` case was fixed in #4645/#4844; this is the remaining tail. Impact is low: `window.open` produces a **visible** top-level tab with `noopener`, so it is not a blind-read SSRF primitive — but a forged link to a loopback service is still worth closing.
2. **No rate-limit.** Unlike the clipboard (1/s) and download (1/2s) handlers, `open_url` has no throttle, so a contract could tab-bomb. The browser popup-blocker mitigates the `window.open`-triggered route (no user gesture), but a direct-post loop is unthrottled.

## Proposed fix

- Broaden the loopback check to the full 127.0.0.0/8 range and IPv4-mapped IPv6 loopback (parse/normalize the host rather than string-matching a fixed list).
- Add a small rate-limit to `open_url` matching the clipboard/download pattern.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-07-24T12:35:43.104398515+00:00
Issue hash: `c42f14d2`

## Value Assessment

[...]

PR #5107, opened 2026-08-03
fix(shell): stop routing same-origin target="_blank" through open_url

### LINKED ISSUE #5106: Regression from #5089: same-origin target="_blank" is a dead click on a loopback node; #5087 was WebKit-only
## Summary

PR #5089 made same-origin `<a target="_blank">` clicks route through the shell's
`open_url` bridge instead of falling through to the browser. That fixed a blank
tab in Safari (#5087), but it regressed the common case: **on a loopback node the
click now does nothing at all**, and the bug it fixed turns out to be WebKit-only,
so Firefox and Chrome users were traded a working tab for a dead click.

Reported against Firefox; measured below as engine-independent for the loopback half.

## Cause 1 — `open_url` refuses loopback, so the click is swallowed

`navigation_interceptor.js` now does, for a same-origin non-`_self` target:

```js
if (target.target && target.target !== '_self') {
  e.preventDefault();
  window.parent.postMessage({ __freenet_shell__: true, type: 'open_url', ... }, '*');
  return;
}
```

The shell's `open_url` handler in `shell_bridge.js` refuses loopback hosts:

```js
if (h === 'localhost' || h === '127.0.0.1' || h === '::1' || h === '0.0.0.0') return;
```

A local node is served from loopback, which is how most desktop users browse. So
the interceptor cancels the browser's navigation and the shell then drops the
message on the floor. No tab, no error, no feedback.

The `window.open` override further down the same file never had this hole — it
checks `isLoopbackHost` and falls back to native precisely so a forwarded open
can't be silently dropped. The anchor branch #5089 added has no such fallback.

## Cause 2 — the blank tab being fixed was WebKit-only

#5089's premise is that the natively-opened popup inherits the sandbox, so the
top-level shell has an opaque origin and its own `frame-src 'self'` can't match.
The first half is true in every engine; the second is not.

I drove the real `navigation_interceptor.js` inside a harness that reproduces the
shell faithfully (same iframe `sandbox` attribute, `frame-src 'self'`, one origin
serving both "contracts", a top-level app-document load redirected to the shell),
with each browser's popup blocker left ON — Playwright disables popup blocking by
default, which masks this class of bug.

Clicking a same-origin `target="_blank"` link in the app frame:

| engine | interceptor | host | popup? | popup origin | app iframe loaded? |
|---|---|---|---|---|---|
| Firefox  | pre-#5089 | routable | 1 | `null` (opaque) | **YES** |
| Firefox  | #5089     | routable | 1 | real origin     | YES |
| Firefox  | pre-#5089 | loopback | 1 | `null` (opaque) | **YES** |
| Firefox  | #5089     | loopback | **0** | — | — |
| Chromium | pre-#5089 | routable | 1 | `null` (opaque) | **YES** |
| Chromium | #5089     | routable | 1 | real origin     | YES |
| Chromium | pre-#5089 | loopback | 1 | `null` (opaque) | **YES** |
| Chromium | #5089     | loopback | **0** | — | — |
| WebKit   | pre-#5089 | routable | 1 | `null` (opaque) | **NO** |
| WebKit   | #5089     | routable | 1 | real origin     | YES |
| WebKit   | pre-#5089 | loopback | 1 | `null` (opaque) | **NO** |
| WebKit   | #5089     | loopback | **0** | — | — |

In the WebKit rows the popup document reports a CSP violation
`frame-src <- http://<host>` and the app frame stays empty; Gecko and Blink report
no violation and load the frame from the same opaque origin. So #5087 was a
Safari-specific bug, and the `0 popups` rows are the regression #5089 introduced
for everyone.

The rows marked `0` are the dead click: the bridge logged `REFUSED_LOOPBACK` and
no window was ever opened.

Note for anyone re-testing: #5089 states the failure did not reproduce in headless
WebKit/Chromium and that popups there "did not inherit the sandbox flags". They do —
the harness above shows `origin: null` in all three engines. Whatever masked it in
the original harness also hid the loopback dead-end.

## Resolution

Reverting the #5089 hunk (PR to follow). That restores the pre-#5089 behavior in
every engine and reopens #5087 for Safari.

The intended real fix is `allow-popups-to-escape-sandbox` on the app iframe, which
would make the natively-opened popup non-sandboxed and fix #5087 without routing
anything through a bridge that can refuse the URL. That needs its own security
review against #1499, which deliberately removed that flag.

If instead the bridge route is kept, it must mirror the `window.open` override's
`isLoopbackHost` fallback so a refused URL falls back to native rather than
vanishing.

[AI-assisted - Claude]

sanity:
Fixes #5106. Reverts the JS change from #5089; reopens #5087.

Reverting at @sanity's direction (raised offline, not on an issue): more Firefox users than Safari users are likely affected, so undo #5089 for now while a safer fix is worked out.

## Why

#5089 stopped same-origin `<a target="_blank">` from falling through to the browser and routed it through the shell's `open_url` bridge instead, to fix a blank tab (#5087). On a **loopback** node that turns the click into nothing at all. The shell's `open_url` handler refuses loopback hosts:

```js
if (h === 'localhost' || h === '127.0.0.1' || h === '::1' || h === '0.0.0.0') return;
```

`127.0.0.1:7509` is the documented local path — all three platforms print `Open http://127.0.0.1:7509/ in your browser` on service start (`service/linux.rs:809`, `service/macos.rs:753`, `service/windows.rs:170`), and it is what the tray and setup wizard open (`DASHBOARD_URL`, `service.rs:199`). So the interceptor `preventDefault`s the navigation and the shell then drops the message: no tab, no error, no feedback.

The `window.open` override further down **`navigation_interceptor.js`** never had this hole — it checks `isLoopbackHost` and falls back to native precisely so a forwarded open can't be silently swallowed. The anchor branch #5089 added has no such fallback.

## Evidence

I drove the real `navigation_interceptor.js` in a harness reproducing the shell faithfully: real `SHELL_PAGE_CSP` and `sandbox_csp_for_origin` served as headers, interceptor inlined as `path_handlers.rs:1746` does, one origin serving both "contracts", top-level app-document load redirected to the shell.

| engine | interceptor | host | popup? | popup origin | app iframe loaded? |
|---|---|---|---|---|---|
| Firefox  | pre-#5089 | routable | 1 | `null` (opaque) | **YES** |
| Firefox  | #5089     | routable | 1 | real origin     | YES |
| Firefox  | pre-#5089 | loopback | 1 | `null` (opaque) | **YES** |
| Firefox  | #5089     | loopback | **0** | — | — |
| Chromium | pre-#5089 | routable | 1 | `null` (opaque) | **YES** |
| Chromium | #5089     | routable | 1 | real origin     | YES |
| Chromium | pre-#5089 | loopback | 1 | `null` (opaque) | **YES** |
| Chromium | #5089     | loopback | **0** | — | — |
| WebKit   | pre-#5089 | routable | 1 | `null` (opaque) | **NO** |
| WebKit   | #5089     | routable | 1 | real origin     | YES |
| WebKit   | pre-#5089 | loopback | 1 | `null` (opaque) | **NO** |
| WebKit   | #5089     | loopback | **0** | — | — |

The `0` rows are the regression: the bridge logged `REFUSED_LOOPBACK` and no window was ever opened. The WebKit rows record `frame-src <- http://<host>` in the popup and an empty app frame; Gecko and Blink record no violation and load the frame from the same opaque origin. All twelve rows were independently reproduced by a second, higher-fidelity harness during review, including on a non-loopback, non-secure-context address (`172.17.0.1`), so the WebKit result is not a 127/8 secure-context artifact.

**Correction to an earlier version of this description.** I originally attributed #5089's clean harness result to Playwright disabling popup blocking by default. Review re-ran the full matrix with blockers on *and* off: all twelve cells are identical, and the `0` rows are `REFUSED_LOOPBACK`, not blocked popups. Popup blocking is irrelevant to everything measured here, and I have no evidence for why the original harness missed it. (Also: my harness's launcher does not apply the blocker flag to WebKit at all, so "blockers on in all three engines" was wrong as stated — it was on in Firefox and Chromium.)

## What the revert costs

Stating this plainly, because my first draft priced it as "a WebKit-only blank tab" and that was wrong on three counts. Blind review caught all three.

The tab the browser opens inherits the app frame's sandbox, so it has an opaque origin. That has **three** consequences:

1. **WebKit renders it blank** (#5087) — for every Safari user, on every host, not just loopback.
2. **On a hosted node it dead-ends in every engine** (#4645). An opaque origin can't read the per-user access key from localStorage, so the tab lands on the "Open this app in a normal tab" panel. `shell_bridge.js:80` names the trigger explicitly — "window.open, or a `target=_blank` link" — and `localStorage` throws `SecurityError` in an opaque origin in Gecko, Blink and WebKit alike (measured in all three during review). #4645 has been annotated.
3. **The permission surface is dead in that tab, in every engine — and this is the one that bites the users this PR is *for*.** From an opaque origin the tab sends `Origin: null` (and `Sec-Fetch-Site: cross-site`), and all three permission endpoints deny it: `/permission/events` via `is_origin_trusted` (`permission_prompts.rs:586-590`), `/permission/{nonce}/respond` with a 403, and `/permission/pending` — the polling fallback — with **HTTP 200 and an empty list**. That last one is why it is invisible: no console error, no failed request, just an overlay that never appears. #5087's trace shows the endpoints being refused.

   The sharp version: on a *hosted* node the app frame never loads anyway, so there is no app to request a permission. On a **local node in Firefox or Chrome** — precisely the users the revert buys a working tab for — the app frame loads fine, the delegate requests permission, and the prompt silently never appears. That is what "app iframe loaded? YES" in the table above is hiding.

   Two honest qualifiers: this is deliberate and fail-closed (a sandboxed iframe on any site can present `Origin: null`, so the permission surface always requires a real one — `permission_prompts.rs:578-590` says so). And it is not *introduced* by this PR: the `window.open` override's own loopback fallback already produces this same opaque-origin tab on a local node today. The revert restores the cost for anchors too.

Also: `target="_top"` and `target="_parent"` anchors match this branch too. The sandbox denies the in-place top-navigation they ask for (no `allow-top-navigation`) in every engine, and post-revert Gecko and Blink then do nothing at all — a dead click, by this PR's own standard. **WebKit is different**: it converts the denied navigation into a popup, so those clicks still land somewhere there. (Measured across 5 runs; an earlier draft said "blocked outright, both engines", which was true of the two I had tested and false in the third. **If you re-measure, wait >2.5s** — the WebKit popup does not register within ~2s, and a short wait reports zero, which is how the wrong version got written.)

Under #5089 all three opened a new tab. To be precise about what that was and wasn't: no top-level navigation happened in any engine or either version, so **no sandbox restriction was bypassed** — `allow-popups` permits the tab. #5089 simply made `_top`/`_parent` behave like `_blank`, silently changing the semantics. An earlier draft of this section called it converting a sandbox-denied navigation into an action the sandbox had refused; that read as a sandbox escape and was wrong.

#5089 fixed 1, 2 and 3 for the anchor path as a side effect. Reverting restores all three.

## The alternative this PR declines, and why you may want it instead

Two independent reviewers converged on the same point, so it belongs in the description rather than a comment: **mirroring `isLoopbackHost` in the anchor branch is strictly better than either #5089 or this revert.** Return natively on a loopback host, forward otherwise. That keeps every one of #5089's wins, removes the dead click entirely, and is no worse than the revert anywhere (WebKit-on-loopback stays blank either way). `isLoopbackHost` is a hoisted declaration in the same IIFE, already in scope for `handleAnchorClick` — it is a few lines inside the hunk this PR is already editing.

It matters that #5087's actual reporter was **not** on loopback: their trace reads `"origin": "http://10.44.101.1:7509"`, and #5089 records them confirming the fix worked. So #5089's benefit landed in exactly the configuration the revert takes away, and the loopback dead click never affected them. That also means the claim "loopback is how most desktop users browse" — which I asserted in #5106 — is unsupported; `127.0.0.1` is the *documented* path, but the ws-API default bind is `::` (all interfaces, `config.rs:2510`), so LAN access works out of the box and at least one real user was using it.

This PR reverts anyway because @sanity asked for the revert first and is working on the real fix. Recording it here so the choice is visible as **declined for sequencing**, not unavailable. Reverting leaves both #5087 and #5106 open.

## Change

- `navigation_interceptor.js`: the same-origin non-`_self` branch is `return` again — byte-identical to pre-#5089 apart from the comment, which now records the costs above and the two constraints on any future fix.
- `path_handlers.rs`: #5089's regression test is replaced by `navigation_interceptor_leaves_same_origin_target_blank_to_the_browser`. Besides pinning the branch from both sides, it asserts `handleAnchorClick` forwards `open_url` from exactly **one** place — without that, a re-route added as a separate earlier branch (`if (target.target === '_blank') { … }`) evades both branch-scoped assertions while the bug is fully present. Mutation-verified. That whole-handler count is also what makes the anchor-vs-`window.open` asymmetry deliberate rather than an accident of two reverts.
- `shell.spec.ts` + fixture: a behavioral regression test in a real browser. It asserts **both** that the interceptor leaves the click to the browser (a popup is created, at the link's own href) and that no interception postMessage is sent — neither subsumes the other. Note the first half pins "the interceptor did not cancel it", not "a real user gets a tab": Playwright leaves Chromium's popup blocker off, so it cannot speak to what a blocker-on user sees. Asserting only the postMessage (my first version) stays green on `{ e.preventDefault(); return; }`, which *is* #5106's dead click; asserting only the popup stays green under #5089, because this harness's node binds `127.x.y.1` and so is not in the refusal list (#4846). Both clicks now happen in one document, so the control proves the listener was live in the document the negative assertion is about.

## About the second commit

Being explicit, because the framing in its commit message was too generous and review called it: `interceptor_loopback_list_matches_shell_open_url_refusal` is **independent hardening, not a regression test for this bug**. It passes identically on `origin/main` — this PR does not modify either loopback list. The behavioral regression test for #5106 is the Playwright one.

It is here partly because the `fix:`-needs-a-test CI gate counts `#[test]` attributes in added `.rs` lines only, and the first commit's `#[test]` line survived as diff context, so the gate failed with a genuine new test present. The gate also cannot see the `.test.mjs` / `.ts` suites (#5096). That said, it earns its place on merit: `isLoopbackHost`'s own comment claims the two lists are "kept in sync" with nothing enforcing it, #4846 proposes changing one of them, and a one-sided change reproduces the silent-drop failure this PR is about. It lifts out cleanly as its own PR if a maintainer prefers.

Per review it now also pins a **floor** on the shell-side list rather than bare equality: that list is a security control (it is what stops a contract forging an `open_url` at the viewer's own loopback services), while `isLoopbackHost` blocks nothing and is only a prediction of it. A bare `assert_eq!` is satisfiable by shrinking *both*, which would be green while opening a real hole. The failure messages now name the correct repair direction.

Writing it also surfaced a pre-existing defect: the sibling `shell_open_url_handler_*` tests bound their scraped region with the next `} else if`, but `open_url` is the **last** arm of that chain, so the bound falls through to `unwrap_or(rest.len())` and their region is the entire rest of the file — 32499 bytes (my commit message says 32469, which is the *character* count; `str::len()` is bytes, and the region contains 15 non-ASCII chars). A stray `h === 'string'` lives inside it. My first draft of the new test inherited the bug. Reported on #5076 rather than fixed here.

## Tests

`Shell Playwright E2E` was green on `3227b5610` and `3c2903def`, with the new test running in both (`✓ 8 … same-origin link with target=_blank is NOT intercepted (#5106)`, 19 passed). It must be re-confirmed on the current head before merge, per the per-SHA rule — the head has moved since, and the later commits are not comment-only.

Locally: all `navigation_interceptor_*`, `interceptor_loopback_*`, `shell_open_url_*` and `shell_page_*` tests pass. Prettier + ESLint clean on the served assets.

Non-vacuity, verified rather than assumed: under #5089's JS the Rust test's first assertion finds `!== '_self') {` instead of `return` and its second finds `type: 'open_url'` in the branch, so both fail; the Playwright test's `toEqual([])` sees `["open_url"]`.

## What this does not change

- The iframe `sandbox` attribute is untouched; `allow-popups-to-escape-sandbox` stays absent and the pin on its absence stays green. Note for anyone reading the comments: that flag was *deliberately removed* by PR #3818 (`ec140e09c`), because an escaped popup gains the node's real origin and can then reach other apps' data and bypass permission prompts. Any fix down that road has to answer #3818 — the code comments now say so, since comments outlive PR descriptions.
- The cross-origin branch (freenet/river#208) and the `window.open` override are untouched. Review confirmed by measurement that the revert does not weaken #3818's hardening, river#208/#3852, #3853/#3854, or GHSA-824h-7x5x-wfmf, and that no contract-side trick (`<base href>`, `document.domain`, credentialed URLs, exotic schemes) can steer a cross-origin URL into the restored branch — `<base href>` moves links the *safe* way, into the cross-origin path.

## One citation corrected beyond the revert

`allow-popups-to-escape-sandbox` was removed by **PR #3818** (`ec140e09c`, "browser-based permission prompts with iframe security hardening"). The existing comments cite **#1499**, which is the *delegate-user-interaction feature request* that motivated it — it says nothing about sandbox flags, and there is no PR #1499. #3818 is also the commit that wrote the original "#1499" pointer, so the error is as old as the code.

I corrected the two pre-existing sites (`path_handlers.rs` assertion + comment, `shell_bridge.js` open_url header) rather than adding a correct citation next to broken ones, since this PR's whole argument leans on that reasoning and sends readers to it twice. Comment-only, and each notes what it used to say so the change is auditable.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -7130,6 +7130,95 @@ mod tests {
         );
     }
 
+    /// The interceptor's `isLoopbackHost` and the shell `open_url` handler's
+    /// loopback refusal must name the SAME hosts. `isLoopbackHost` exists only
+    /// to predict what the shell will refuse, so that `window.open` falls back
+    /// to native instead of forwarding an open that will be dropped — its own
+    /// comment says "Kept in sync with the loopback block in shell_bridge.js's
+    /// open_url handler", but nothing enforced that until now.
+    ///
+    /// Drift in either direction is a silent user-visible bug:
+    ///   - shell refuses a host the interceptor does not list → the forwarded
+    ///     open vanishes and the click does nothing. That is exactly the
+    ///     failure #5106 hit through the anchor path, which had no such list
+    ///     at all.
+    ///   - interceptor lists a host the shell would open → the open needlessly
+    ///     falls back to a sandbox-inheriting native popup (#4645).
+    ///
+    /// #4846 proposes completing this allow-list (it is exact-match today, so
+    /// e.g. `127.0.0.2` is not covered). Whoever does that must change both
+    /// sides; this test is what tells them so.
+    #[test]
+    fn interceptor_loopback_list_matches_shell_open_url_refusal() {
+        // Collect the host literals from each `h === '<host>'` comparison in a
+        // region. Both sides normalize into a local `h` before comparing, so
+        // this picks up the comparison chain and nothing else (scheme checks
+        // use `u.protocol`, not `h`).
+        fn loopback_hosts(region: &str) -> Vec<String> {
+            const NEEDLE: &str = "h === '";
+            let mut out = Vec::new();
+            let mut rest = region;
+            while let Some(i) = rest.find(NEEDLE) {
+                let after = &rest[i + NEEDLE.len()..];
+                match after.find('\'') {
+                    Some(end) => {
+                        out.push(after[..end].to_string());
+                        rest = &after[end..];
+                    }
+                    None => break,
+                }
+            }
+            out.sort();
+            out.dedup();
+            out
+        }
+
+        let interceptor_fn_idx = NAVIGATION_INTERCEPTOR_JS
+            .find("function isLoopbackHost")
+            .expect("interceptor isLoopbackHost helper present");
+        let interceptor_region = &NAVIGATION_INTERCEPTOR_JS[interceptor_fn_idx..];
+        // The helper is short; bound the region at its closing brace so a later
+        // `h === ...` elsewhere in the file can't leak in.
+        let interceptor_end = interceptor_region
+            .find("\n  }")
+            .expect("isLoopbackHost helper is brace-terminated");
+        let interceptor_hosts = loopback_hosts(&interceptor_region[..interceptor_end]);
+
+        // Anchor tightly on the refusal itself — `var h = u.hostname` up to the
+        // `return;` that drops the message. The sibling tests here bound the
+        // open_url branch with the next `} else if`, but open_url is the LAST
+        // arm of that chain, so that bound silently runs to end-of-file; a
+        // stray `h === 'string'` further down would then be read as a loopback
+        // host. Slicing to the `return;` keeps this to the comparison chain.
+        let refusal_idx = SHELL_BRIDGE_JS
+            .find("var h = u.hostname")
+            .expect("shell open_url loopback normalization present");
+        let rest = &SHELL_BRIDGE_JS[refusal_idx..];
+        let refusal_end = rest
+            .find("return;")
+            .expect("shell open_url loopback refusal returns");
+        let shell_hosts = loopback_hosts(&rest[..refusal_end]);
+
+        // Guard against the comparison going vacuous: if a refactor renames the
+        // local or restructures the chain, both sides would extract nothing and
+        // the equality below would pass while guarding nothing (#5076).
+        assert!(
+            shell_hosts.contains(&"localhost".to_string()),
+            "extracted no loopback hosts from the shell open_url handler — the \
+             `h === '...'` chain moved or was renamed, so this test is no longer \
+             comparing anything. Re-anchor it. Got: {shell_hosts:?}"
+        );
+        assert_eq!(
+            interceptor_hosts, shell_hosts,
+            "isLoopbackHost and the shell's open_url loopback refusal have \
+             drifted. A host the shell refuses but the interceptor does not \
+             list makes window.open forward an open the shell then drops \
+             silently (#5106); the reverse forwards to a sandbox-inheriting \
+             native popup (#4645). Update both, or #4846 when completing the \
+             allow-list."
+        );
+    }
+
     /// Direct postMessages from a malicious iframe can synthesize an
     /// `open_url` payload without going through the upstream
     /// `NAVIGATION_INTERCEPTOR_JS` scheme filter, so the shell-side
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -6830,12 +6830,28 @@ mod tests {
     ///     frame; Gecko and Blink loaded it. Firefox and Chrome users were
     ///     therefore traded a working tab for a dead click.
     ///
-    /// So: keep this branch a plain `return` — hand the click back to the
-    /// browser. Re-routing it needs either `allow-popups-to-escape-sandbox`
-    /// on the app iframe (so the popup is not sandboxed in the first place)
-    /// or a loopback fallback mirroring `window.open`'s. Both assertions
-    /// below must stay: the first fails if the `return` is replaced, the
-    /// second if a forward is added alongside it.
+    /// The revert is not free, and this test should not be read as saying the
+    /// current behaviour is good: the tab the browser opens inherits the
+    /// sandbox, so on a HOSTED node it can't read the per-user access key and
+    /// dead-ends on the "Open this app in a normal tab" panel in every engine
+    /// (#4645), and on WebKit it renders blank (#5087). #5089 fixed both for
+    /// this path incidentally. This test pins the lesser failure, not a good
+    /// one, until a fix lands that has neither.
+    ///
+    /// Two constraints on that fix. `allow-popups-to-escape-sandbox` on the
+    /// app iframe was deliberately REMOVED by #1499 — an escaped popup gains
+    /// the node's real origin, letting a malicious app reach other apps' data
+    /// and bypass permission prompts (pinned in `shell_page_*` above) — so
+    /// that route has to answer #1499 first. A bridge re-route needs a
+    /// loopback fallback mirroring `window.open`'s `isLoopbackHost`.
+    ///
+    /// Note that such a fix WILL fail both assertions below, and that is
+    /// correct — they pin today's behaviour, not a permanent invariant.
+    /// Whoever lands it should replace this test, not weaken it. The first
+    /// assertion is the load-bearing one; the second is a cheap belt-and-
+    /// braces check that a forward was not added ahead of the `return`
+    /// (anything added after it would be unreachable, so it catches less
+    /// than the first).
     #[test]
     fn navigation_interceptor_leaves_same_origin_target_blank_to_the_browser() {
         let js = NAVIGATION_INTERCEPTOR_JS;
@@ -6860,8 +6876,11 @@ mod tests {
         );
         assert!(
             !block.contains("type: 'open_url'"),
-            "the same-origin new-window branch must not forward to open_url without a \
-             loopback fallback like the window.open override's isLoopbackHost check (#5106)"
+            "no open_url forward may be added to the same-origin new-window branch ahead \
+             of the `return` (#5106). If you are DELIBERATELY re-routing this branch — \
+             with a loopback fallback like the window.open override's isLoopbackHost \
+             check, or after removing the sandbox inheritance — this test is pinning the \
+             old behaviour and should be replaced, not deleted piecemeal"
         );
     }
 
@@ -7146,27 +7165,38 @@ mod tests {
     ///     falls back to a sandbox-inheriting native popup (#4645).
     ///
     /// #4846 proposes completing this allow-list (it is exact-match today, so
-    /// e.g. `127.0.0.2` is not covered). Whoever does that must change both
-    /// sides; this test is what tells them so.
+    /// e.g. `127.0.0.2` is not covered). Whoever does that one-sidedly will
+    /// trip this test. Note the limit, though: it compares LITERALS, so if
+    /// #4846 is implemented as a rule (`h.startsWith('127.')`) on both sides,
+    /// both literal sets shrink together and this test goes green without
+    /// having checked the new rule. It catches one-sided literal drift — the
+    /// common case — not a symmetric refactor.
     #[test]
     fn interceptor_loopback_list_matches_shell_open_url_refusal() {
         // Collect the host literals from each `h === '<host>'` comparison in a
-        // region. Both sides normalize into a local `h` before comparing, so
-        // this picks up the comparison chain and nothing else (scheme checks
-        // use `u.protocol`, not `h`).
+        // region.
+        //
+        // The needle is NOT self-limiting: it also matches the tail of
+        // `hash === '` and `iframePath === '`, which occur elsewhere in
+        // shell_bridge.js. What makes this sound is purely the REGION BOUNDS
+        // chosen below — do not loosen them on the theory that `h` is unique.
+        // That mistake is live: see the note on the shell anchor.
         fn loopback_hosts(region: &str) -> Vec<String> {
             const NEEDLE: &str = "h === '";
             let mut out = Vec::new();
             let mut rest = region;
             while let Some(i) = rest.find(NEEDLE) {
                 let after = &rest[i + NEEDLE.len()..];
-                match after.find('\'') {
-                    Some(end) => {
-                        out.push(after[..end].to_string());
-                        rest = &after[end..];
-                    }
-                    None => break,
-                }
+                let end = after.find('\'').expect(
+                    "unterminated host literal in a loopback comparison — the region \
+                     bounds or the JS shape changed; failing loudly rather than \
+                     comparing a silently truncated list",
+                );
+                out.push(after[..end].to_string());
+                // Advances by at least NEEDLE.len() each iteration, so this
+                // terminates; all indices are ASCII-needle finds, so every
+                // slice lands on a char boundary.
+                rest = &after[end..];
             }
             out.sort();
             out.dedup();
--- a/crates/core/src/server/path_handlers.rs
+++ b/crates/core/src/server/path_handlers.rs
@@ -4724,10 +4724,17 @@ mod tests {
         // allow-popups-to-escape-sandbox must NOT be present. It was removed because
         // escaped popups gain localhost:7509 origin, allowing malicious web apps to
         // access other apps' data and bypass permission prompts. External links are
-        // now opened via the open_url shell bridge message instead. See #1499.
+        // now opened via the open_url shell bridge message instead.
+        //
+        // Citation corrected in #5107: this said "#1499", but that is the
+        // delegate-user-interaction feature request. The flag was removed by
+        // PR #3818 (`ec140e09c`, "browser-based permission prompts with iframe
+        // security hardening"), which is also the commit that wrote the
+        // original "#1499" pointer here. #1499 is the motivation; #3818 is the
+        // change and the place to argue with.
         assert!(
             !html.contains("allow-popups-to-escape-sandbox"),
-            "allow-popups-to-escape-sandbox must not be set (security: #1499)"
+            "allow-popups-to-escape-sandbox must not be set (security: #3818)"
         );
         // open_url handler must be present in shell bridge JS for external links
         assert!(
@@ -6812,8 +6819,13 @@ mod tests {
             .expect("same-origin target check present");
         let block = &js[cross_origin_idx..target_attr_idx];
 
+        // Match the payload shape, not a bare `shiftKey`. This region is the
+        // one #5106 grew by ~30 comment lines, and a bare needle is satisfiable
+        // by prose that merely discusses shift-clicking — the same way a bare
+        // `preventDefault` needle in the sibling test above was left matching a
+        // comment, mutation-confirmed, until it was tightened.
         assert!(
-            block.contains("shiftKey"),
+            block.contains("shiftKey: !!e.shiftKey"),
             "cross-origin open_url postMessage must include shiftKey to honour \
              shift-click as a new-window request (#3853); got block: {block}"
         );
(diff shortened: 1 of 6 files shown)


sanity:
Local test output promised in the PR body (`cargo test -p freenet --lib -- navigation_interceptor interceptor_loopback shell_open_url shell_page`):

```
running 25 tests
test server::path_handlers::tests::navigation_interceptor_leaves_same_origin_target_blank_to_the_browser ... ok
test server::path_handlers::tests::navigation_interceptor_handles_cross_origin_target_blank ... ok
test server::path_handlers::tests::navigation_interceptor_forwards_shift_key_for_open_url ... ok
test server::path_handlers::tests::navigation_interceptor_listens_on_click_and_auxclick ... ok
test server::path_handlers::tests::navigation_interceptor_js_intercepts_clicks ... ok
test server::path_handlers::tests::navigation_interceptor_overrides_window_open ... ok
test server::path_handlers::tests::interceptor_loopback_list_matches_shell_open_url_refusal ... ok
test server::path_handlers::tests::shell_open_url_handler_accepts_http_and_https_but_blocks_localhost ... ok
test server::path_handlers::tests::shell_open_url_handler_blocks_ipv6_loopback ... ok
test server::path_handlers::tests::shell_open_url_handler_rejects_dangerous_schemes ... ok
test server::path_handlers::tests::shell_open_url_handler_honours_shift_key ... ok
test server::client_api::tests::shell_page_csp_allows_service_worker ... ok
test server::client_api::tests::shell_page_csp_allows_same_origin_and_websocket_connect ... ok
... (14 shell_page_* tests) ...
test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 4567 filtered out
```

And the behavioural test in CI on this branch:

```
✓ 8 [chromium] › tests/shell.spec.ts:274:5 › same-origin link with target=_blank is NOT intercepted (#5106) (797ms)
  19 passed (6.3s)
```

[AI-assisted - Claude]

sanity:
Updated after blind review (lens: code-first). Findings and what I did:

**Important — the PR mispriced its own cost.** I framed the revert as costing "a WebKit-only blank tab". Incomplete: the sandbox-inheriting tab also dead-ends on a **hosted** node in **every** engine (#4645), because an opaque origin cannot read the per-user access key. `shell_bridge.js` names the trigger explicitly ("window.open, or a target=_blank link"). #5089 fixed both for the anchor path incidentally; reverting restores both. The PR body now has a "What the revert costs" section stating this, the code comments say it, and I posted a correction on #5106. The trade is still net-favourable — a dead click is worse than a tab that explains itself — but it is not free.

**Important — comments recommended `allow-popups-to-escape-sandbox` with no caveat.** #1499 removed that flag deliberately (escaped popup gains the node's real origin → can reach other apps' data and bypass permission prompts), and there is a test pinning its absence. The PR description hedged; the comments did not, and comments outlive PR descriptions. Both now carry the caveat.

**Minor — the revert test's doc contradicted itself**, calling a loopback-fallback re-route a valid future fix while saying "both assertions must stay" (such a fix fails both). It now says a correct fix *will* fail them and should replace the test rather than weaken it. Assertion 2's message no longer implies a fallback would satisfy it, and the doc no longer overstates what it catches — anything added after the `return` is unreachable, so it catches less than assertion 1.

**Minor — `loopback_hosts` gave a false safety reason.** I wrote that the `h === '` needle picks up "the comparison chain and nothing else". It is not self-limiting: it matches the tail of `hash === '` and `iframePath === '` elsewhere in `shell_bridge.js`. Soundness comes entirely from the region bounds — which is exactly the belief that would justify loosening them, i.e. the sibling tests' bug. Corrected.

**Minor — the #4846 claim was too strong.** If #4846 is implemented as `h.startsWith('127.')` on both sides, both literal sets shrink together and the test goes green having checked nothing. It catches one-sided literal drift, not a symmetric refactor. Now stated.

**Nit — silent truncation.** An unterminated host literal now `.expect()`s instead of returning a partial list.

Not changed, with reasons: the 89-line test addition stays (thematically the PR's central argument; lifts out cleanly as its own commit if a maintainer prefers). The extra comment bytes ship uncompressed to every client — consistent with the file's existing 35-line block, so noted not objected to. The new fixture link spawns a real popup in the E2E suite; harmless today, but worth knowing if a console/CSP assertion is ever added after that test.

Verified after the change: all 11 `navigation_interceptor_*` / `interceptor_loopback_*` / `shell_open_url_*` tests pass, Prettier + ESLint clean, and I re-checked that the enlarged comment introduces no false positive into any scraped region (it sits before `target_attr_idx`; the shift-key slice that does contain it has no `shiftKey` text in the comment).

[AI-assisted - Claude]

sanity:
Round 3 of blind review (lenses: testing, security, adversarial claim-check). Every finding taken; head is now `4814c744a`. Three of these were things I should not have shipped.

**Testing — the behavioural test did not assert the behaviour.** It checked only that no interception postMessage was sent. But `{ e.preventDefault(); return; }` — cancel, forward nothing — sends no message either, and that *is* #5106's dead click. The test was green on the exact symptom it was added for. It now asserts a tab actually opens as well. Neither assertion subsumes the other: popup-only stays green under #5089, because this harness binds `127.x.y.1` and so misses the refusal list (#4846). Both clicks also moved into one document — `target="_blank"` opens a new browsing context so the fixture's listener survives, and the control now proves it was live in the document the negative assertion is about. That retired a `waitForTimeout(300)` which failed *open* across an OOPIF boundary.

**Testing — anchoring hole.** Both assertions scope to the same-origin branch, so a re-route added as a separate *earlier* branch evades them entirely while the bug is fully present. Mutation-verified both ways. Now also pins that `handleAnchorClick` forwards `open_url` from exactly one place.

**Security — the parity test coupled a security control to a prediction.** The shell's loopback refusal is what stops a contract forging an `open_url` at the viewer's own loopback services, including the node's API. `isLoopbackHost` blocks nothing; it only predicts that refusal. A bare `assert_eq!` is satisfiable by shrinking **both** — green, while `open_url` would happily open `http://127.0.0.1:7509/` for anyone who asks. Now floors the shell side first and names the repair direction, since someone red on #4846 otherwise has an easier wrong fix available than the right one.

**Self-inflicted, worth stating.** My own comment prose made a pin vacuous: "preventDefault-ed" in the new same-origin comment sits inside the cross-origin block, so `navigation_interceptor_handles_cross_origin_target_blank`'s bare `preventDefault` needle matched prose — deleting the real call left that river#208 pin green. Mutation-confirmed, tightened to `e.preventDefault()`. I then audited every needle in these tests against comment-only text; no other prose-only match.

**Claim-check refuted three of my own claims**, all now retracted in the description and on #5106:
- The popup-blocker explanation was unsupported. Re-running the matrix with blockers on *and* off gives identical results in all twelve cells; the zero rows are `REFUSED_LOOPBACK`, not blocked popups. I have no evidence for why #5089's harness came back clean.
- "Loopback is how most desktop users browse" was an unsupported universal.
- **#5087's reporter was never on loopback** (`origin: http://10.44.101.1:7509`) and confirmed #5089 fixed it for them. #5089's benefit landed in exactly the configuration this revert removes.
- Also corrected: `32469` was a character count; `str::len()` is bytes, so the sibling region is `32499` bytes. The commit message carries the wrong figure; I am not force-pushing to fix a number in prose.

The measurements themselves were independently reproduced on a second, higher-fidelity harness (real CSP as headers, interceptor inlined as the server does), including on a non-loopback non-secure-context address — so the WebKit-only result is not a 127/8 artifact.

**The description now leads with the alternative this PR declines.** Two reviewers independently concluded that mirroring `isLoopbackHost` in the anchor branch is strictly better than either #5089 or this revert. It is recorded as declined-for-sequencing, not unavailable — @sanity asked for the revert first — but reverting leaves both #5087 and #5106 open.

[AI-assisted - Claude]

sanity:
Round 4, head is now `e000fba0c`. The adversarial claim-check lens found a citation error that predates this PR and two of my own retracted claims that had survived in code comments.

**`#1499` is the wrong issue, and always has been.** The comments say `allow-popups-to-escape-sandbox` was removed "See #1499". Issue #1499 is *"Delegates should be able to interact with users"* — a feature request about signing with a private key — and there is no PR #1499. The flag was removed by **PR #3818** (`ec140e09c`, "browser-based permission prompts with iframe security hardening"), which is also the commit that wrote the "#1499" pointer. I verified all of this before changing anything.

So the *rationale* in the comments is correct and faithfully repeats the repo's; only the pointer is wrong, and it sends a maintainer to a feature request with nothing to answer. I corrected my own two uses plus the two pre-existing sites (`path_handlers.rs` comment + assertion message, `shell_bridge.js:1200`), because this PR leans on that reasoning twice and leaving a corrected citation next to broken ones would be worse than either. Each site records what it used to say.

**Two claims I had retracted everywhere else were still sitting in the Rust doc comment** — "which is how most desktop users browse" and "(popup blockers on)". That is precisely the failure mode the "comments outlive PR descriptions" principle warns about, so both are now fixed *and* carry an explicit "do not repeat this from earlier drafts" note, so the correction is what survives rather than just the absence.

**The `_top`/`_parent` paragraph was wrong in WebKit.** I wrote "the sandbox blocks them outright — measured, both engines". True of the two I tested; WebKit converts the denied navigation into a popup. Rewritten. I also withdrew "#5089 converted a sandbox-denied top-navigation into an action the sandbox had refused" — no top-level navigation happened in any engine or version, `allow-popups` permits the tab, so nothing was bypassed; #5089 just made `_top`/`_parent` behave like `_blank`. That phrasing read as a sandbox escape, which is not a claim to leave lying around in a security-adjacent PR.

**Cost #3 reached the code comments**, where it belonged all along.

**And one I inflicted on myself, which is a neat inversion of this PR's theme.** Naming the flag in full inside `shell_bridge.js` broke `shell_page_contains_iframe_and_bridge` — that file is inlined verbatim into the shell page, and the test greps the rendered page for the flag name to prove it is not set. So after a whole review cycle spent fixing pins that comment prose could satisfy *vacuously*, I tripped the same coupling in the opposite direction. Reworded, with a comment explaining why not to name it there. Worth stating the asymmetry: prose creating a **false alarm** is fail-safe; prose creating a **silent pass** is what actually bites, and that one bit this PR earlier via `preventDefault`.

Local: 25/25 including all `shell_page_*`. Prettier + ESLint clean. CI was fully green on `4814c744a` (Playwright, Unit & Integration, Simulation, NAT, Windows, Clippy, Fmt, Rule Lint) and is re-running here.

[AI-assisted - Claude]

PR #5108, opened 2026-08-03
fix(broadcast): schedule by wire payload, one drain per lane

### LINKED ISSUE #4961: Official River room updates take 4-20s to propagate: full-state sends hit the 2-stream large-payload limit

sanity:
## Problem

`broadcast_queue.rs` schedules every outbound state-change broadcast through one
of two concurrency pools — 12 permits for small payloads, 2 for large ones,
split at 64 KiB. Production telemetry from the 0.2.118 fleet (the
`network_efficiency_v1` queue block added in #5091, 1,096 reporting peers) shows
the scheduler making the wrong decision on both axes:

**1. The lane was chosen from the contract STATE size, not the wire payload.**
`enqueue()` set `payload_size = new_state.size()`, but the real payload is
picked later inside `broadcast_to_single_peer` — usually a small delta, because
the target's summary is cached. Per node-day:

| counter | per node-day |
|---|---|
| `scheduled_large` | 45,982 |
| `queued_large_actual_small` | 29,661 (**64.5%**), avg ~950 B |
| `queued_small_actual_large` | 0 fleet-wide |

So roughly two thirds of everything squeezed through the 2-permit pool had no
business being there, and the error was entirely one-directional: any contract
whose state exceeds 64 KiB sent even its tiniest deltas through the large lane.

**2. One FIFO worker blocked on the large semaphore, so small entries behind a
large one could not be scheduled at all.** The drain loop did
`sem.acquire_owned().await` INLINE, before `tokio::spawn`. While the worker
waited for one of the 2 large permits it could not `pop_front` anything —
including small entries whose 12 permits were entirely free. Per node-day:
1,887 head-of-line incidents, 237 s of worker block time, and a 7,260 s
small-entry wait integral (~61 ms added latency per small broadcast).

Together these are very likely the mechanism behind #4961 (Official River room
updates taking 4-20 s to propagate).

**3.** Minor, same area: the dedup-replacement path overwrote `existing.new_state`
but left `existing.payload_size` at the superseded state's value, so the lane
classification (and the `scheduled_*_state_bytes` attribution) went stale the
moment a state crossed the threshold.

## Approach

**Per-lane FIFOs and a drain worker per lane.** `QueueState` now holds two
order maps keyed by a global enqueue sequence, and `start_worker` spawns
`drain_lane` once per lane. A queued send waiting for large-lane capacity can no
longer stall small entries. The alternative — moving the `acquire` inside the
spawned task — also unblocks the stall but unbounds spawned tasks and gives up
the replace-on-dedup that stops a hot contract from queueing one send per
update. Ordering by a shared sequence keeps capacity eviction dropping the
*globally* oldest entry and keeps a re-laned entry in its original FIFO
position; `entries` and the order maps are always mutated together, so there are
no stale keys to skip.

**Lane chosen from the predicted wire payload.** At enqueue time we cannot know
the payload, but we can know whether a delta is *likely*: a delta is sent when we
hold the target's cached summary and the delta-incompat memo is not armed for the
contract. Both are in-memory reads at the enqueue site, which already has
`op_manager` (`InterestManager::has_peer_summary`, new, avoids
`get_peer_summary`'s clone; `DeltaIncompat::deltas_suppressed_peek`, new, avoids
inflating `suppressed_total` with predictions that never became sends).

**The prediction decides scheduling only; the uplink bound is enforced against
the real payload.** A prediction can be wrong in the expensive direction — the
contract refuses the delta as not smaller than the state and sends full state
anyway — and 12 concurrent full-state streams saturate a residential uplink,
which is the exact failure this queue exists to prevent. So the permit travels
with the send as a `SendLanePermit`, and at the point the real `payload_size`
becomes known `ensure_capacity_for` takes a large-lane permit before any bytes
go out.

That correction is the one remaining cross-lane coupling, and it is **bounded,
not absent**. Mispredictions are not independent events: a `get_contract_summary`
timeout, an `Interests` full-replace wiping cached summaries, or the
delta-incompat memo arming mispredicts *every peer of a contract at once*, and a
contract whose deltas are never smaller than its state mispredicts on every send
to every peer *permanently*. So the wait is staged:

1. For `UPGRADE_SLOT_HOLD_WINDOW` (250 ms) the send keeps its small-lane slot.
   However many mispredictions arrive together, no unrelated small broadcast is
   stalled longer than that.
2. Past that it moves to a parking area of `MAX_PARKED_LANE_UPGRADES` slots
   (one small-pool turnover), giving up its small-lane slot only once a parking
   slot is in hand. Parked sends hold no pool capacity, so without this cap the
   drain would keep dispatching replacements and the parked set — one serialized
   payload each — would grow at the enqueue rate while draining at the large
   lane's.
3. With parking exhausted the send simply keeps its small-lane slot and waits
   under it. The small lane then applies backpressure, which is the correct
   response to an uplink genuinely saturated with full-state sends, and is
   bounded work rather than unbounded memory.

In-flight sends are therefore bounded at 12 + 2 + 12, and no large payload
reaches the wire outside the large-payload limit.

`queued_class` and the permit are handed to `broadcast_to_single_peer` bundled as
`QueueScheduling` rather than as two `Option`s, per
`.claude/rules/bug-prevention-patterns.md` — a class without a permit would put
bytes on the wire outside the limit, and a permit without the class would
silently stop feeding the classification counters.

## Telemetry

No counter changes definition, recording site, or position in the
`network_efficiency_v1` `queue` array. Several change what they MEAN, because
the small/large split moved from state size to predicted payload and because the
lanes now drain concurrently. The full list is on the `queue` field in
`router.rs`; in short:

- `scheduled_large` / `scheduled_large_state_bytes` should drop by roughly the
  64.5% that was misrouted; `scheduled_small` picks it up. Small-lane state
  bytes are no longer bounded by the 64 KiB threshold.
- `queued_large_actual_small` should collapse toward zero — that IS the
  classification fix landing.
- `queued_small_actual_large` (0 fleet-wide today, and near-structurally so —
  a delta may exceed the state by up to `MIN_FULL_STATE_SAVING_BYTES`, and the
  stale-dedup defect fixed here could also fire it) is now the misprediction
  COUNT and is **expected to become non-zero**. Each of those sends still waits
  for a large-lane permit before hitting the wire, so a non-zero value is a
  latency cost, not an uplink breach. It is the leading indicator for the
  correction path: if it is material, the wait and the parking area deserve
  their own duration/occupancy counters — which means growing the positional
  `queue: [u64; 15]` in `NetworkEfficiencyV1`, a wire-schema change this bug
  fix deliberately does not carry.
- `large_head_blocking_incidents` / `large_head_blocked_millis` keep their exact
  trigger, but the large pool now has a second consumer (a mispredicted send
  correcting its lane), so an incident no longer implies a large-lane *entry*
  caused the exhaustion.
- `small_entry_millis_blocked` keeps its exact recorded quantity but shifts from
  **blockage** to **overlap**, because those entries are now drained
  concurrently by their own worker; its population also widened with the lane
  definition. Collapsing toward zero is the primary before/after signal for
  defect 2.

## Testing

New regression tests in `broadcast_queue.rs`, most of them driving the real
`drain_lane` loop through a test dispatch (the production pop/permit/spawn
ordering is what is under test):

- `small_entry_is_scheduled_while_the_large_lane_is_saturated`
- `large_state_with_an_expected_delta_takes_the_small_lane`
- `delta_send_expected_requires_a_cached_summary_and_an_unarmed_memo`
- `mispredicted_large_payload_waits_for_large_lane_capacity` (incl. the
  threshold boundary and both phases of the wait)
- `parking_area_is_bounded_and_falls_back_to_holding_the_slot`
- `cancelling_a_waiting_send_releases_its_permit_in_either_phase`
- `lane_for_routes_a_large_state_delta_send_to_the_small_lane`
- `each_lane_is_woken_by_its_own_enqueue`
- `large_lane_send_does_not_upgrade_against_itself`
- `closed_large_pool_does_not_strand_the_send`
- `dedup_replacement_relanes_and_refreshes_state_size`
- `capacity_eviction_drops_the_globally_oldest_across_lanes`
- `broadcast_to_single_peer_gates_wire_payload_on_lane_capacity_pin` and
  `start_worker_spawns_one_drain_per_lane`, plus the existing HOL-wiring pin
  retargeted at `drain_lane`
- `delta_incompat::peek_does_not_count_a_suppression`,
  `interest::has_peer_summary_agrees_with_get_peer_summary`

Every one was verified to FAIL against a deliberate reversion of the defect it
covers — fourteen mutations, fourteen detections, no collateral failures. The
matrix is in the review comments below.

Local: `cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`,
`cargo check -p freenet --all-targets`,
`cargo check -p freenet --features simulation_tests --lib`, and
`cargo test -p freenet --lib` (4,583 tests) all clean.

Fixes #4961. Measurement source: #5091.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/pool_tests/non_idempotent_detector_tests.rs
+++ b/crates/core/src/contract/executor/pool_tests/non_idempotent_detector_tests.rs
@@ -213,8 +213,10 @@ fn egress_paths_gated_on_is_contract_broken() {
     let heal_gate = body
         .find("is_contract_broken")
         .expect("handle_sync_state_to_peer must gate on is_contract_broken");
+    // Anchor on the method, not `broadcast_queue.enqueue`: rustfmt breaks the
+    // receiver onto its own line once the argument list grows.
     let heal_send = body
-        .find("broadcast_queue.enqueue")
+        .find(".enqueue(")
         .expect("handle_sync_state_to_peer must still enqueue the heal send");
     assert!(
         heal_gate < heal_send,
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -9,6 +9,15 @@
 //! `BroadcastQueue` limits the number of concurrent outbound broadcast streams
 //! via a semaphore, deduplicates entries per (contract, peer), and replaces
 //! older entries with newer state when a duplicate is enqueued.
+//!
+//! The limit is two pools — 12 concurrent small payloads, 2 concurrent large
+//! ones — and each pool has its OWN FIFO and its OWN drain worker, so a send
+//! waiting for large-payload capacity cannot delay small sends behind it. The
+//! pool is chosen from the PREDICTED wire payload (a large-state contract sends
+//! a small delta to any peer whose summary we hold), and corrected against the
+//! real payload before any bytes go out. See #4961 for the measurements that
+//! drove all three of those; before it, one worker drained one FIFO and picked
+//! the pool from the full contract STATE size.
 
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
@@ -30,9 +39,14 @@ use super::p2p_protoc::P2pBridge;
 /// `queue` submodule.
 const STREAM_COMPLETION_TIMEOUT: Duration = Duration::from_secs(120);
 
-/// The queued state size used to choose the small or large concurrency pool.
-/// The actual delta/full payload may be much smaller; telemetry records that
-/// mismatch at the point where the wire payload is known.
+/// Payload size, in bytes, at or above which a send is rate-limited by the
+/// narrow large-payload pool instead of the wide small-payload pool.
+///
+/// Applied to the PREDICTED wire payload at enqueue time (see
+/// `queue::classify_payload_lane`) and again to the ACTUAL wire payload once it
+/// is known (see [`SendLanePermit::ensure_capacity_for`]). It used to be
+/// applied to the full contract STATE size, which is not what goes on the wire
+/// whenever a delta is sent — the classification half of #4961.
 const BROADCAST_QUEUE_PAYLOAD_SIZE_THRESHOLD: usize = 64 * 1024;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -42,6 +56,111 @@ const BROADCAST_QUEUE_PAYLOAD_SIZE_THRESHOLD: usize = 64 * 1024;
 pub(super) enum QueuedPayloadClass {
     Small,
     Large,
+}
+
+/// The concurrency permit a dispatched broadcast holds while its payload is on
+/// the wire, together with the lane that permit came from.
+///
+/// The queue picks a lane from a PREDICTION of the wire payload (a large-state
+/// contract whose target has a cached summary sends a small delta, not the
+/// state). A prediction can be wrong in the expensive direction — the contract
+/// may refuse a delta (`DeltaUnavailable::NotEfficient`) and send full state
+/// after all — and 12 concurrent full-state streams saturate a residential
+/// uplink, which is the exact failure this queue exists to prevent. So the
+/// permit is upgraded to the large lane at the point the real payload size
+/// becomes known; see [`Self::ensure_capacity_for`].
+// Constructed only by the production queue, which is cfg'd out under
+// `simulation_tests`; the type stays compiled because `broadcast_to_single_peer`
+// (shared with the sim fan-out) takes it.
+#[cfg_attr(feature = "simulation_tests", allow(dead_code))]
+pub(super) struct SendLanePermit {
+    /// The permit currently held. Dropping `self` releases it.
+    permit: tokio::sync::OwnedSemaphorePermit,
+    /// Which pool `permit` came from.
+    lane: QueuedPayloadClass,
+    /// The large-payload pool, for the upgrade path.
+    large_pool: Arc<tokio::sync::Semaphore>,
+}
+
+#[cfg_attr(feature = "simulation_tests", allow(dead_code))]
+impl SendLanePermit {
+    fn new(
+        lane: QueuedPayloadClass,
+        permit: tokio::sync::OwnedSemaphorePermit,
+        large_pool: Arc<tokio::sync::Semaphore>,
+    ) -> Self {
+        Self {
+            permit,
+            lane,
+            large_pool,
+        }
+    }
+
+    /// Make sure this send holds a LARGE-lane permit before putting
+    /// `payload_size` bytes on the wire, if the payload turned out to be large
+    /// while the queue had predicted small.
+    ///
+    /// This is what keeps the uplink guarantee independent of the prediction's
+    /// accuracy: the small pool's 12 permits may only ever admit small
+    /// payloads, so a mispredicted send blocks here until large-lane capacity
+    /// exists, exactly as if it had been queued as large.
+    ///
+    /// The small permit is held ACROSS the wait and released only once the
+    /// large permit is in hand (assigning `self.permit` drops the old one).
+    /// Releasing first would let the worker dispatch further entries while this
+    /// task parks, so the number of in-flight sends — each holding a clone of a
+    /// contract state — would be bounded only by the queue depth (256) instead
+    /// of by the pools (14). Holding it costs one small-lane slot for the
+    /// duration of the wait, which is bounded by large-lane drain rate.
+    ///
+    /// Cancellation: this is awaited inline by the send that OWNS the
+    /// `SendLanePermit`, so dropping that send drops the permit too — mid-wait
+    /// or mid-send, whichever permit is currently held is released exactly
+    /// once. Neither pool can leak a slot or be double-charged.
+    async fn ensure_capacity_for(&mut self, payload_size: usize) {
+        if payload_size < BROADCAST_QUEUE_PAYLOAD_SIZE_THRESHOLD
+            || matches!(self.lane, QueuedPayloadClass::Large)
+        {
+            return;
+        }
+        match self.large_pool.clone().acquire_owned().await {
+            Ok(large) => {
+                // Swap, then drop: the small permit is released by this
+                // assignment, after the large one is already held.
+                self.permit = large;
+                self.lane = QueuedPayloadClass::Large;
+            }
+            Err(_) => {
+                // The pool is closed, which only happens at shutdown. Keep the
+                // small permit and let the send proceed rather than stalling
+                // forever; the worker logs the closure separately.
+                tracing::debug!(
+                    payload_size,
+                    "Large broadcast pool closed; sending mispredicted payload \
+                     under the small-lane permit"
+                );
+            }
+        }
+    }
+}
+
+/// Everything the production queue hands to one dispatched send.
+///
+/// The two fields must co-occur: a queued class without a permit would put
+/// bytes on the wire outside the concurrency limit, and a permit without the
+/// class would silently stop feeding the classification-accuracy counters
+/// (`queued_large_actual_small` / `queued_small_actual_large`). Bundling them
+/// is the `.claude/rules/bug-prevention-patterns.md` "paired `Option` fields
+/// that must co-occur" fix shape — the sim fan-out passes `None` for the whole
+/// bundle, so it cannot pass one half.
+#[cfg_attr(feature = "simulation_tests", allow(dead_code))]
+pub(super) struct QueueScheduling {
+    /// The lane the entry was CLASSIFIED into at enqueue time. Immutable — the
+    /// mismatch counters compare it against the real payload, so it must not
+    /// track [`SendLanePermit::lane`] across an upgrade.
+    queued_class: QueuedPayloadClass,
+    /// The concurrency permit gating this send's bytes onto the wire.
+    permit: SendLanePermit,
 }
 
 /// Process-global UPDATE-broadcast stream-assembly telemetry (#4440).
@@ -337,7 +456,8 @@ pub(super) async fn fanout_send_needed(
 // `cargo clippy --features simulation_tests -- -D warnings` clean.
 #[cfg(not(feature = "simulation_tests"))]
 mod queue {
-    use std::collections::{HashMap, VecDeque};
+    use std::collections::{BTreeMap, HashMap};
+    use std::future::Future;
     use std::sync::Arc;
     use std::sync::atomic::{AtomicBool, Ordering};
     use std::time::Instant;
@@ -346,10 +466,10 @@ mod queue {
     use tokio::sync::{Mutex, Notify, Semaphore};
 
     use crate::node::OpManager;
-    use crate::ring::PeerKeyLocation;
+    use crate::ring::{PeerKey, PeerKeyLocation};
 
     use super::super::p2p_protoc::P2pBridge;
-    use super::{QueuedPayloadClass, broadcast_to_single_peer};
+    use super::{QueueScheduling, QueuedPayloadClass, SendLanePermit, broadcast_to_single_peer};
 
     /// Maximum concurrent outbound broadcast streams for small payloads (< 64KB).
     /// Small payloads (deltas, chat messages) can fan out aggressively without
@@ -366,19 +486,93 @@ mod queue {
     /// Key for deduplicating broadcast entries: (contract, peer identity).
     type DedupeKey = (ContractKey, PeerKeyLocation);
 
+    /// Choose the concurrency pool for one queued broadcast.
+    ///
+    /// `delta_expected` says whether the send is expected to put a DELTA on the
+    /// wire rather than the whole state — see [`delta_send_expected`]. That is
+    /// the whole point of this function: the lane must be chosen from the
+    /// predicted WIRE payload, not from the contract state, because a contract
+    /// whose state exceeds the threshold still sends ~1 KB deltas to every peer
+    /// whose summary we hold. Classifying those as large squeezed two thirds of
+    /// all broadcast traffic through a 2-permit pool (#4961; measured on the
+    /// 0.2.118 fleet as 29,661 of 45,982 large-lane items per node-day having an
+    /// actual payload below the threshold, averaging ~950 bytes).
+    ///
+    /// A prediction is not a guarantee — the contract can refuse the delta and
+    /// send full state anyway — so it decides SCHEDULING only. The uplink bound
+    /// is enforced against the real payload by
+    /// [`SendLanePermit::ensure_capacity_for`].
+    pub(super) fn classify_payload_lane(
+        state_size: usize,
+        delta_expected: bool,
+    ) -> QueuedPayloadClass {
+        if state_size < super::BROADCAST_QUEUE_PAYLOAD_SIZE_THRESHOLD || delta_expected {
+            QueuedPayloadClass::Small
+        } else {
+            QueuedPayloadClass::Large
+        }
+    }
+
+    /// Whether the send for `(key, target)` is expected to carry a delta.
+    ///
+    /// Mirrors the two conditions `broadcast_to_single_peer` itself applies
+    /// before it can compute a delta: we must hold a cached summary for the
+    /// target, and deltas must not be suppressed for the contract by the
+    /// delta-incompat memo. Both are in-memory reads (a DashMap lookup with no
+    /// summary clone, and a non-counting peek at the memo — using the counting
+    /// `suppress_deltas` here would double-count every suppressed send).
+    fn delta_send_expected(
+        op_manager: &OpManager,
+        key: &ContractKey,
+        target: &PeerKeyLocation,
+    ) -> bool {
+        let peer_key = PeerKey::from(target.pub_key().clone());
+        op_manager.interest_manager.has_peer_summary(key, &peer_key)
+            && !op_manager
+                .ring
+                .delta_incompat
+                .deltas_suppressed_peek(key.id())
+    }
+
     /// A pending broadcast entry in the queue.
     struct BroadcastEntry {
         key: ContractKey,
         target: PeerKeyLocation,
         new_state: WrappedState,
-        /// Payload size in bytes, used to select concurrency pool.
-        payload_size: usize,
+        /// Contract STATE size in bytes. Feeds the `scheduled_*_state_bytes`
+        /// counters only — it is deliberately NOT what picks the lane (see
+        /// [`classify_payload_lane`]).
+        state_size: usize,
+        /// The pool this entry is queued for. Recomputed on dedup replacement,
+        /// so a state that crosses the threshold (or a peer whose summary
+        /// appears/disappears) re-lanes the entry instead of going stale.
+        lane: QueuedPayloadClass,
+        /// Global enqueue order, and this entry's key in its lane's order map.
+        seq: u64,
     }
 
-    /// Internal queue state: FIFO ordering via VecDeque + HashMap for dedup lookup.
+    /// Internal queue state: per-lane FIFO ordering + HashMap for dedup lookup.
+    ///
+    /// The lanes are separate FIFOs so a large entry waiting for large-lane
+    /// capacity cannot stall small entries behind it: each lane is drained by
+    /// its own worker, and a blocked worker only ever blocks its own lane. With
+    /// one shared FIFO the single worker awaited the large semaphore INLINE, so
+    /// it could not `pop` anything at all while waiting — the head-of-line half
+    /// of #4961 (1,887 incidents and a 7,260 s small-entry wait integral per
+    /// node-day on the 0.2.118 fleet).
+    ///
+    /// Ordering is by a global monotonic sequence number so the two FIFOs stay
+    /// comparable: capacity eviction still drops the globally oldest entry, and
+    /// re-laning an entry on dedup replacement preserves its original position.
+    /// `entries` and the two order maps are always mutated together, so an order
+    /// map never names a key `entries` lacks.
     struct QueueState {
-        /// FIFO order of dedup keys. Entries may be stale if replaced by dedup.
-        order: VecDeque<DedupeKey>,
+        /// Small-lane order: enqueue sequence -> dedup key.
+        small_order: BTreeMap<u64, DedupeKey>,
+        /// Large-lane order: enqueue sequence -> dedup key.
+        large_order: BTreeMap<u64, DedupeKey>,
+        /// Sequence assigned to the next NEW entry.
+        next_seq: u64,
         /// Actual entries, keyed by (contract, peer). Dedup replaces the state in-place.
         entries: HashMap<DedupeKey, BroadcastEntry>,
         /// Pairs already removed from the queue but waiting for a permit or
@@ -386,15 +580,28 @@ mod queue {
         /// as before, while telemetry can identify duplicates that queued
         /// deduplication cannot see.
         active: HashMap<DedupeKey, u64>,
-        /// Number of nominal-small entries still queued. Maintained under the
+        /// Number of entries queued in the small lane. Maintained under the
         /// queue lock so head-of-line observation is O(1).
         small_queued: usize,
-        /// Time integral for the one FIFO worker waiting at a large-payload
+        /// Time integral for the LARGE-lane worker waiting at a large-payload
         /// permit. Updated on every small-queue population change, so arrivals
         /// during the wait are included rather than sampled away.
+        ///
+        /// Recorded at exactly the pre-#4961 condition, so the counters it
+        /// feeds keep their definitions across the fix. What changes is the
+        /// consequence, not the measurement: with a drain per lane those small
+        /// entries are being scheduled concurrently, so
+        /// `small_entry_millis_blocked` becomes an OVERLAP integral rather than
+        /// a blockage, and collapsing toward zero is the fix landing.
         hol_block: Option<HolBlockHandle>,
     }
 
+    /// Lock ordering, now that TWO workers can touch an observation
+    /// concurrently (the large lane owns the handle, the small lane mutates the
+    /// queue population it integrates over): the tokio queue mutex is always
+    /// taken BEFORE this inner `std::sync::Mutex`, and never the other way —
+    /// `finish_now` holds only the inner one, and nothing reaches for the queue
+    /// lock while holding it. So the two lanes cannot invert.
     #[derive(Clone)]
     struct HolBlockHandle(Arc<HolBlockState>);
 
@@ -487,7 +694,9 @@ mod queue {
     impl QueueState {
         fn new() -> Self {
             Self {
-                order: VecDeque::new(),
+                small_order: BTreeMap::new(),
+                large_order: BTreeMap::new(),
+                next_seq: 0,
                 entries: HashMap::new(),
                 active: HashMap::new(),
                 small_queued: 0,
@@ -499,18 +708,55 @@ mod queue {
             self.entries.len()
         }
 
-        /// Pop the oldest entry. Skips stale keys (removed by eviction or dedup).
-        fn pop_front(&mut self, now: Instant) -> Option<BroadcastEntry> {
-            while let Some(key) = self.order.pop_front() {
-                if let Some(entry) = self.entries.remove(&key) {
-                    if entry.payload_size < super::BROADCAST_QUEUE_PAYLOAD_SIZE_THRESHOLD {
-                        self.set_small_queued(self.small_queued.saturating_sub(1), now);
-                    }
-                    return Some(entry);
-                }
-                // Stale key (was evicted or already popped), skip
+        fn order_mut(&mut self, lane: QueuedPayloadClass) -> &mut BTreeMap<u64, DedupeKey> {
+            match lane {
+                QueuedPayloadClass::Small => &mut self.small_order,
+                QueuedPayloadClass::Large => &mut self.large_order,
             }
-            None
+        }
+
+        /// Pop the oldest entry queued for `lane`, or `None` if that lane is
+        /// empty. A busy OTHER lane is invisible here — that is the point.
+        fn pop_lane(&mut self, lane: QueuedPayloadClass, now: Instant) -> Option<BroadcastEntry> {
+            let (_, key) = self.order_mut(lane).pop_first()?;
+            let entry = self
+                .entries
+                .remove(&key)
+                .expect("order maps and entries are mutated together");
+            debug_assert_eq!(entry.lane, lane, "entry popped from the wrong lane");
+            if matches!(lane, QueuedPayloadClass::Small) {
+                self.set_small_queued(self.small_queued.saturating_sub(1), now);
+            }
+            Some(entry)
+        }
+
+        /// Drop the globally oldest entry across both lanes (capacity eviction).
+        fn evict_oldest(&mut self, now: Instant) -> Option<BroadcastEntry> {
+            let oldest_small = self.small_order.keys().next().copied();
+            let oldest_large = self.large_order.keys().next().copied();
+            let lane = match (oldest_small, oldest_large) {
+                (Some(small), Some(large)) if large < small => QueuedPayloadClass::Large,
+                (Some(_), _) => QueuedPayloadClass::Small,
+                (None, Some(_)) => QueuedPayloadClass::Large,
+                (None, None) => return None,
+            };
+            self.pop_lane(lane, now)
+        }
+
+        /// Insert a brand-new entry at the tail of its lane.
+        fn push_new(&mut self, key: DedupeKey, entry: BroadcastEntry) {
+            self.order_mut(entry.lane).insert(entry.seq, key.clone());
+            self.entries.insert(key, entry);
+        }
+
+        /// Move an already-queued entry to a different lane, keeping its
+        /// original sequence (and therefore its FIFO position).
+        fn relane(&mut self, key: &DedupeKey, from: QueuedPayloadClass, to: QueuedPayloadClass) {
+            let Some(seq) = self.entries.get(key).map(|entry| entry.seq) else {
+                return;
+            };
+            self.order_mut(from).remove(&seq);
+            self.order_mut(to).insert(seq, key.clone());
         }
 
         fn set_small_queued(&mut self, next: usize, now: Instant) {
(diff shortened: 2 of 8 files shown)


sanity:
## Multi-perspective review (Full tier — concurrency/scheduling change)

Four independent Claude reviewers, each blind to the others, each reading the
checked-out code in a separate detached worktree at the PR head (not the diff
text), each told not to rubber-stamp. Lenses:

1. **code-first** — read the code before the description; flag intent/impl mismatches
2. **testing** — coverage gaps, vacuous tests, would each regression test fail without its fix
3. **skeptical adversarial** — deadlock, lost wakeup, permit leak, starvation, unbounded growth, panics
4. **concurrency/scheduling + telemetry truth** — permit lifecycle, cancellation, lock ordering, and whether all 15 counters still mean what `router.rs` claims

Every finding below was fixed or is dismissed with a specific reason. Three
reviewers independently found the same two P1s, which is what a diverse review
is for — I had reasoned about the second one while writing it and got the
tradeoff wrong.

### P1 — fixed: the lane-correction wait was charged to the contract as WASM CPU

Found independently by lenses 1, 3, and 4. `report_send_cpu` measures WALL time
since the summarize/delta work started and reports it on `ExecCpuMicros` — the
axis cost-pressure eviction reads (#4861), whose per-node floor is 50 ms/s. I
had placed `ensure_capacity_for` (an unbounded semaphore wait) *before* it, so a
5 s wait for uplink capacity would report 5,000,000 µs of "contract CPU", 100×
the floor, and could nominate a zero-demand contract for eviction for work it
never did. Pre-change there was no `.await` between the delta computation and
the report.

Fixed by moving `report_send_cpu` above the correction, and pinned:
`broadcast_to_single_peer_gates_wire_payload_on_lane_capacity_pin` now asserts
`report_send_cpu` precedes `ensure_capacity_for`, which in turn precedes both
send sites. The pre-existing `..._reports_send_cost_pin` only counts occurrences,
so it could not have caught either direction.

### P1 — fixed: the correction could starve the small lane it was protecting

Found independently by lenses 1, 3, and 4. The upgrade held the small-lane
permit across the wait for a large one. With 12 small permits and 2 large, twelve
simultaneous mispredictions occupy the whole small pool, the small drain then
blocks on `acquire_owned`, and cross-lane head-of-line blocking is back —
through the correction path, in the lane that now carries nearly all traffic.
Lens 3 and 4 both pointed out the part I had underweighted: **mispredictions are
correlated bursts, not independent events** (a `get_contract_summary` timeout, an
`Interests` full-replace, the memo arming, or a NotEfficient contract each
mispredict every peer at once), and a large permit can be held for
`STREAM_COMPLETION_TIMEOUT` = 120 s.

I had considered releasing the permit before waiting and rejected it because it
unbounds in-flight sends (each holding a serialized payload) at the queue depth
rather than the pools. Both objections are real, so the wait is now two-phase:
hold the slot for `UPGRADE_SLOT_HOLD_WINDOW` (250 ms) — long enough for the
common case of a large lane about to free a permit — then hand it back and keep
waiting without it. Small-lane exposure to a misprediction burst is bounded at
250 ms; the in-flight count only exceeds the pools under sustained large-lane
congestion, when the alternative is a stalled small lane. Nothing is on the wire
while no permit is held. Covered by
`mispredicted_large_payload_waits_for_large_lane_capacity`, which now asserts
both phases.

The four "impossible by construction" doc claims were also corrected — the
guarantee is *bounded*, not absent, and saying otherwise is exactly the
intent/impl mismatch lens 1 exists to catch.

### P1 — fixed: new tests introduced a flake in an existing test

Lens 2. `shadow_demand::tests::broadcast_queue_depth_gauge_round_trips` asserts
an exact value of a process-global gauge, justified by a comment saying no test
instantiates the production queue. My scheduling tests do, and both
`enqueue_in_lane` and `drain_lane` publish that gauge, so a store could land
between its store and its load. Both sides now carry `#[serial_test::serial]`
and the stale comment is corrected.

### P1 — fixed: the production classifier input was untested

Lens 2. Every scheduling test called `enqueue_in_lane` with the lane passed in
by hand, so `delta_send_expected` — the predicate `enqueue` actually calls — had
no coverage: dropping the `!`, swapping `&&` for `||`, or hard-coding either
term left the whole suite green. It now takes `&InterestManager` and
`&DeltaIncompat` instead of `&OpManager` (both trivially constructible in a unit
test) and has a truth-table test.

### P1 — fixed: the correction's call site and its guard were unpinned

Lens 2, two findings. Nothing in the crate calls `broadcast_to_single_peer`, so
deleting the `ensure_capacity_for` call or sliding it past a send left every test
green — now pinned (above). And `ensure_capacity_for`'s already-Large early
return was untested: removing it would make a large-lane send wait for a *second*
large permit while holding one, wedging full-state broadcasting permanently once
two are in flight. Now covered by
`large_lane_send_does_not_upgrade_against_itself`, plus
`closed_large_pool_does_not_strand_the_send` for the shutdown bypass and the
threshold boundary in the misprediction test.

### P2 — fixed

- **`scheduled_*` meaning change undocumented** (lenses 1, 3, 4): the partition
  moved from state size to predicted payload, so small-lane state bytes are no
  longer threshold-bounded. The `router.rs` `queue` doc now lists all the
  meaning changes, not two of them, and the PR body no longer opens with a line
  that contradicts its own bullets.
- **`start_worker`'s lane→pool wiring pinned only by an occurrence count**
  (lens 2): handing the small lane the large semaphore is a one-token slip that
  reinstates #4961 at full strength while the pin still counts 2. Now pins the
  exact wiring.
- **A vacuous assertion** (lens 2): `large_permits.available_permits() == 0`
  could not fail — the test's own blocker holds both. Replaced with the
  assertion that discriminates (the dispatch took a *small* permit).
- **Doc corrections** (lens 1): `delta_send_expected` mirrors two of the three
  delta preconditions, not all of them (our own summary is unknowable at enqueue
  time — now stated, with the reason); `suppressed_total` counts attempts that
  reached payload selection while armed, not "sends the memo forced to full
  state"; the `has_peer_summary` rationale no longer conflates summary size with
  state size; the over-count a peek would cause is worse than 2×, not exactly 2×.
- **Unmonitored worker handles** (lens 1): pre-existing, but there are two tasks
  now and one dying would silently kill its lane. Noted at the call site;
  reachable only via a closed semaphore, i.e. shutdown.

### P2 — acknowledged, not fixed here

- **No counter for small-lane permit starvation or upgrade-wait duration**
  (lenses 3, 4). Correct, and it is the residual I would most want measured. It
  needs a 16th element in the positional `queue: [u64; 15]` array, i.e. a
  `network_efficiency_v1` schema bump and a collector/dashboard change — out of
  scope for a bug fix, and the fix caps the exposure at 250 ms per misprediction
  rather than leaving it unbounded. `queued_small_actual_large` (0 fleet-wide
  today) is the leading indicator: if it is material after rollout, the duration
  axis is the follow-up. Called out in the PR body.
- **`available_permits() == 0` sampling race at the HOL trigger** (lens 4): an
  upgrader can take the last large permit between the sample and the acquire, so
  `large_head_blocking_incidents` can miss a wait. Pre-existing race shape, now
  slightly more likely; documented on the counter rather than changed, since
  changing the trigger would break the pre/post comparison this counter exists
  for.
- **Two skip exits hold a lane permit across a contract-handler round trip for a
  send that ships zero bytes** (lens 4). Unchanged from `main`; genuinely
  pre-existing and out of scope, though now concentrated in the small pool.

### Verified sound (not merely skipped)

The reviewers explicitly cleared, with reasoning: no lock inversion (the tokio
queue mutex is always taken before the inner observation mutex, and `finish_now`
never reaches for the queue lock); no lock held across an await; no lost wakeup
on any of the enqueue / dedup / re-lane / eviction paths (the notify targets the
*post*-relane lane); no permit leak or double-release, including on all six
early-return exits of `broadcast_to_single_peer` and under cancellation; no
deadlock (nothing holding a large permit ever waits for one); `entries` and both
order maps are mutated together on every path, so the `.expect` and the
`debug_assert_eq!` cannot fire; `small_queued` accounting exact across all five
mutation sites; array order and all 15 counter recording sites unchanged; no
double count or lost count from the `suppress_deltas` / `deltas_suppressed_peek`
split; the retargeted pins still bind the regions they intend to; and the new
tests are deterministic under `start_paused` (no wall-clock assertions).

### Mutation matrix

Each defect reverted in isolation (or in an independent batch), rebuilt, and the
suite re-run. Nine mutations, nine detections, no collateral failures.

| # | mutation | test that failed |
|---|---|---|
| 1 | lane ignores the delta prediction (pre-fix: state size only) | `large_state_with_an_expected_delta_takes_the_small_lane` |
| 2 | dedup replacement leaves `state_size`/`lane` at the superseded values | `dedup_replacement_relanes_and_refreshes_state_size` |
| 3 | `ensure_capacity_for` is a no-op | `mispredicted_large_payload_waits_for_large_lane_capacity` |
| 4 | one worker drains a single global FIFO, acquiring inline (the pre-fix scheduler) | `small_entry_is_scheduled_while_the_large_lane_is_saturated` |
| 5 | `report_send_cpu` moved back below the capacity wait | `broadcast_to_single_peer_gates_wire_payload_on_lane_capacity_pin` |
| 6 | drop the already-Large guard in `ensure_capacity_for` | `large_lane_send_does_not_upgrade_against_itself` |
| 7 | never hand the small-lane slot back (phase 2 removed) | `mispredicted_large_payload_waits_for_large_lane_capacity` |
| 8 | prediction ignores the delta-incompat memo | `delta_send_expected_requires_a_cached_summary_and_an_unarmed_memo` |
| 9 | small lane wired to the large pool in `start_worker` | `start_worker_spawns_one_drain_per_lane` |

[AI-assisted - Claude]

sanity:
## Review round 2 — re-run on the revised content

Round 1's fixes changed the code materially, so that review was stale. Four
independent Claude reviewers again (code-first, testing, skeptical concurrency,
telemetry truth), blind to each other and to round 1's findings, reading commit
`8e5ca8792` in a separate checkout. It found a **P0 I introduced while fixing
round 1**, which is the argument for re-reviewing rather than assuming a fix is
an improvement.

### P0 — fixed: the guard test I added in round 1 was vacuous by round 2

`large_lane_send_does_not_upgrade_against_itself` asserted that an
already-large-lane send returns instead of waiting for a second permit from its
own pool. Round 2's two-phase wait then made the *reverted* code pass too:
without the guard, phase 2 drops the send's own permit back into the pool and
immediately re-acquires it, so it completes in 250 ms — inside the test's 5 s
bound. The test now asserts the virtual clock did not advance at all. Verified
failing against the reverted guard.

Worth stating plainly: the fix for one round's finding silently disarmed that
same round's test. Only re-running the review caught it.

### P1 — fixed: I documented an in-flight bound that does not exist

Found independently by three of the four lenses. Round 2's phase-2 comment said
the in-flight send count is "bounded by the `active` tracking cap rather than by
the pools". It is not: `track_active` is observation-only and `drain_lane`
dispatches regardless of what it returns. A parked send holds no permit, so the
drain immediately dispatches a replacement, and the parked set grows at the
enqueue rate while draining at the large lane's — each parked send retaining a
serialized payload that is ≥64 KiB by construction and ~1 MB for the large-state
contracts this path exists for.

The skeptical lens also corrected my mental model of *when*: I had described
mispredictions as bursty, but a contract whose deltas are never smaller than its
state mispredicts on every send to every peer **forever**, so this is a steady
state, not a transient.

Fixed structurally: phase 2 now takes a parking permit from a pool of
`MAX_PARKED_LANE_UPGRADES` (one small-pool turnover) BEFORE giving up its
small-lane slot. With parking exhausted the send keeps its small-lane slot and
waits under it, so the small lane applies backpressure — bounded work — instead
of the node accumulating unbounded payloads. Backpressure is the right answer
for an uplink genuinely saturated with full-state sends. New test
`parking_area_is_bounded_and_falls_back_to_holding_the_slot`; the constant is
tied to the pool size by `parking_matches_one_small_pool_turnover`.

### P1 — fixed: the enqueue-time composition was still untested

Round 1 made `delta_send_expected` unit-testable and `classify_payload_lane` was
already pure, but nothing covered *wiring them together* — a dropped `!` in
`enqueue` restores the entire classification defect with every test green.
Extracted as `lane_for` with a test that walks the actual #4961 case.

### P2 — fixed

- **Uncovered paths** (testing lens): the symmetric `evict_oldest` direction (a
  version that always picks one lane passed the old test), the per-lane `Notify`
  split (only incidentally covered), and the cancellation-safety claim in the
  docs. All three now have directed tests; the first two were verified failing
  against their reverts.
- **Six doc claims corrected** (code-first + telemetry lenses), each of which
  would have misled the next reader:
  - `UPGRADE_SLOT_HOLD_WINDOW`'s "covers the common case" was unsupported — a
    full-state stream at the ~1.25 MB/s this module cites takes ~0.5-0.7 s for
    the 550-840 KB states involved, so phase 2 is a *normal* outcome. The
    window's real job is capping small-pool occupancy, and it now says so.
  - `large_pool_closed`'s "happens only at shutdown" — nothing closes these
    pools at all today. Now marked unreachable, and identified as the one
    documented way a large payload can reach the wire unmetered.
  - "two thirds of all broadcast traffic" was two thirds of *large-lane*
    traffic; the evidence cited does not support the stronger claim.
  - The threshold's "ACTUAL wire payload" — the sender's summary and framing
    ride along uncounted (pre-existing blind spot, but the exactness claim was
    new).
  - `queued_small_actual_large` "structurally zero" pre-fix: a delta may exceed
    the state by up to `MIN_FULL_STATE_SAVING_BYTES`, and the stale-dedup defect
    this PR fixes could also fire it. Also relabelled a count, not a rate —
    nothing counts correct predictions and `scheduled_small` is a biased
    denominator.
  - Two more counters whose meaning shifted (`active_tracking_overflow`, which
    went from structurally unreachable to "the parking area piled up", and
    `enqueues_while_pair_active`, whose sampling window now includes the
    correction wait) added to the `router.rs` list, plus a note that the
    HOL counters' gating population widened too.
- **`pop_lane`'s `.expect`** is a second way a lane worker can die; noted at
  `start_worker` alongside the closed-semaphore exit.

### P2 — acknowledged, not fixed

- **No counter for the correction's wait or the parking area.** Still the
  residual I would most want measured, and round 2 sharpened why: under the
  steady-state misprediction case the only existing counter that moves is
  `active_tracking_overflow`, and only at the extreme. Adding
  `lane_upgrade_slot_released` / `lane_upgrade_wait_millis` means growing the
  positional `queue: [u64; 15]` in `NetworkEfficiencyV1` — a wire-schema change
  emitted by every peer, needing a collector/dashboard change, which is not
  something a bug fix should carry. `queued_small_actual_large` (0 fleet-wide
  today) is the leading indicator; if it is material after rollout the duration
  axis is the follow-up. Documented in the PR body and in `router.rs`.
- **`has_peer_summary` ↔ `begin_peer_summary_broadcast::Known` is unpinned**
  (telemetry lens). The predictor and the send path read the same field via
  different accessors; the new test pins `has_peer_summary` ↔ `get_peer_summary`,
  but the send path calls neither. They agree today (verified by reading both);
  it is a rot vector, not a live bug, and its only symptom is a rising
  misprediction count that the docs now tell analysts to expect. Worth a pin,
  but it belongs with the interest-manager API rather than in this PR.
- **A single lane worker dying is now quieter than before** (one lane stops
  instead of all broadcasting). Unreachable today (nothing closes the pools);
  noted in the code.
- **Two skip exits hold a lane permit across a contract-handler round trip.**
  Unchanged from `main`, pre-existing.

### Verified sound in round 2

Both concurrency lenses independently cleared, with reasoning: the
timeout/acquire tie at the phase boundary (a ready acquire wins; a dropped
`acquire_owned` returns any assigned permit, so nothing leaks across it); no
double-release (`take_large` assigns before dropping); no deadlock; lock
ordering; lost wakeups on every enqueue/dedup/re-lane/eviction path; `entries` ↔
order-map consistency on every path, so `pop_lane`'s `.expect` and
`debug_assert_eq!` cannot fire; `small_queued` exact across all five mutation
sites; `active` bounded; the `broadcast_queue_depth_gauge` serial group complete
(and `BROADCAST_QUEUE_EFFICIENCY_METRICS` needing no protection, since the only
test asserting on it uses a local instance); all pin slice boundaries correct
including the self-match hazard; array order and all 15 recording sites intact;
`suppress_deltas`/`deltas_suppressed_peek` with no double or lost count; and the
determinism of every `start_paused` assertion traced explicitly through the
virtual clock.

### Mutation matrix (cumulative)

Fourteen mutations across three batches, each reverted in isolation or in an
independent batch, rebuilt, suite re-run. **Fourteen detections, no collateral
failures.**

| # | mutation | test that failed |
|---|---|---|
| 1 | lane ignores the delta prediction (pre-fix: state size only) | `large_state_with_an_expected_delta_takes_the_small_lane` |
| 2 | dedup replacement leaves `state_size`/`lane` stale | `dedup_replacement_relanes_and_refreshes_state_size` |
| 3 | `ensure_capacity_for` is a no-op | `mispredicted_large_payload_waits_for_large_lane_capacity` |
| 4 | one worker drains a single global FIFO, acquiring inline (the pre-fix scheduler) | `small_entry_is_scheduled_while_the_large_lane_is_saturated` |
| 5 | `report_send_cpu` moved back below the capacity wait | `broadcast_to_single_peer_gates_wire_payload_on_lane_capacity_pin` |
| 6 | drop the already-Large guard | `large_lane_send_does_not_upgrade_against_itself` |
| 7 | never hand the small-lane slot back | `mispredicted_large_payload_waits_for_large_lane_capacity` |
| 8 | prediction ignores the delta-incompat memo | `delta_send_expected_requires_a_cached_summary_and_an_unarmed_memo` |
| 9 | small lane wired to the large pool | `start_worker_spawns_one_drain_per_lane` |
| 10 | park without taking a parking slot (unbounded) | `parking_area_is_bounded_and_falls_back_to_holding_the_slot` |
| 11 | `lane_for` negates the prediction | `lane_for_routes_a_large_state_delta_send_to_the_small_lane` |
| 12 | `evict_oldest` always picks the large lane | `capacity_eviction_drops_the_globally_oldest_across_lanes` |
| 13 | one shared `Notify` for both lanes | `each_lane_is_woken_by_its_own_enqueue` |
| 14 | (batch A/B/D re-verification of 1-9 above) | as listed |

### Validation on the final content

`cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`,
`cargo check -p freenet --all-targets`,
`cargo check -p freenet --features simulation_tests --lib` — all clean.
`cargo test -p freenet --lib` — 4,583 passed, 0 failed, run twice.

[AI-assisted - Claude]

sanity:
## Review round 3 — re-run on the round-2 content

Round 2's fixes changed the code again, so that review was stale too. Two lenses
this time (adversarial concurrency, testing + truth-of-claims), independent and
blind, reading `eb3435cd8`. Both converged on the same three P1s, and one of them
is **the same failure shape round 2 found** — a round's own fix disarming a test
that same round had added. That is now twice, which is the whole argument for
re-reviewing after every fix rather than assuming a fix is an improvement.

### P1 — fixed: the cancellation test I added in round 2 never cancelled anything

`cancelling_a_waiting_send_releases_its_permit_in_either_phase` moved the
`JoinHandle` into `tokio::time::timeout`. Dropping a `JoinHandle` **detaches** a
task; it does not abort it — so despite a comment saying "abort it the way a
shutdown would", nothing was aborted. Both branches then passed for reasons
unrelated to cancellation: the phase-1 branch because the send *completed
normally* once the blocker was dropped, the phase-2 branch because parking had
already released the slot before the assertion ran. The test would have stayed
green with the permit leaked on drop.

It now uses `&mut`, actually calls `abort()`, asserts the abort took effect, and
checks all three pools **before** unblocking anything. Verified failing against a
`forget()`-the-permit mutation.

### P1 — fixed: phase 2 sampled parking availability once, so a send could be pinned for its whole wait

`try_acquire_owned()` on the parking area meant a send that merely *arrived*
while parking was full kept its small-lane slot for the entire remaining wait —
even if a slot freed a millisecond later. Both lenses derived the same reachable
steady state: 12 sends park, the drain dispatches 12 replacements, those find
parking full, and all 12 pin small-lane permits against a 2-permit pool. That is
the #4961 symptom, one layer down.

Phase 2 now **waits** for a parking slot. It cannot deadlock: slots are released
only by sends that already hold a large permit and are waiting on nothing else,
so the area always drains at the large lane's rate. The concurrency lens checked
this for cycles independently and agreed.

The behavioural difference is now pinned: when the first parked send leaves, its
slot goes straight to the send that was waiting for one, which then releases
*its* small-lane slot. Under the old sampling code that second send kept pinning
a small-lane slot even though the area had room.

### P1 — fixed: the parking bound was untestable, and its guarantee was overstated in five places

The test helper built a fresh parking `Semaphore` **per permit**, so no test ever
had two upgraders contend for one pool — and the bound exists *only* because
production shares a single pool. Moving the constructor one line, into the
per-dispatch call, would have restored the unbounded parked set with every test
green. Tests now share a pool and assert contention, hand-off and slot return;
`start_worker_spawns_one_drain_per_lane` additionally pins that the pool is
constructed exactly once and only cloned per lane.

Separately, five doc sites still claimed a stall bounded by
`UPGRADE_SLOT_HOLD_WINDOW` "however many mispredictions arrive at once". That
bound ends where the parking area does. All five now say so, and the module
header states the guarantee this design deliberately gives up: past parking
capacity the small lane applies backpressure at the large lane's drain rate,
because the alternative is an unbounded set of parked sends each retaining a
serialized payload — bounded work beats unbounded memory.

### P2 — fixed

- **`router.rs`'s active-key-overflow note was wrong** — I had written that the
  counter becomes reachable via the parking area. It does not: the parking bound
  is exactly what keeps worst-case in-flight (12 + 12 + 2, plus 2 popped) an
  order of magnitude under the 256-key cap. It stays a pure invariant check, and
  a non-zero value means a leaked tracking guard, not a backlog. Saying otherwise
  would have misdirected the first operator to see one.
- **The hold-window rationale cited the boundary case**, 64 KiB at ~1.25 MB/s ≈
  52 ms — which argues the *opposite* of the conclusion it supported. Uses the
  ~1 MB figure the path actually exists for.
- **Two pins matched their own mutations.** `enqueue` must bind its lane FROM
  `lane_for` (a body that merely mentions it is not enough), and the parking pool
  pin needed both halves (constructed once, cloned per lane). Both were caught by
  running the mutations rather than by reading the pins — they looked fine.
- **Closed-pool-while-parked** now covered: the one path where a send proceeds
  holding no permit at all, which the docs call out as the single way a large
  payload can reach the wire unmetered (unreachable today — nothing closes these
  pools).
- **`each_lane_is_woken_by_its_own_enqueue`'s ordering assumption** made explicit;
  its failure mode under a reverted shared `Notify` depends on spawn order.
- "seven other pins" → there are ten.

### P2 — acknowledged, not fixed

- **Still no counter for the correction path's wait or the parking area's
  occupancy.** Raised in all three rounds, and I agree it is the residual worth
  measuring — under sustained misprediction the only existing counter that would
  move is `active_tracking_overflow`, and the parking bound now keeps even that
  unreachable. Adding one means growing the positional `queue: [u64; 15]` in
  `NetworkEfficiencyV1`, which every peer emits and the collector parses — a wire
  schema change that does not belong in a bug fix. `queued_small_actual_large`
  (0 fleet-wide today) is the leading indicator: it counts every entry into the
  correction path, so if it is material after rollout, the duration/occupancy
  axes are the follow-up. This is stated in the PR body and in `router.rs`.
- **`has_peer_summary` ↔ `begin_peer_summary_broadcast::Known` remains unpinned.**
  Verified equivalent by reading both; a rot vector, not a live bug, and the pin
  belongs with the interest-manager API.
- **A single lane worker dying is quieter than pre-change** (one lane stops
  rather than all broadcasting). Unreachable today; noted in the code, along with
  the second way it could happen (`pop_lane`'s `.expect`).

### Verified sound in round 3

Both lenses independently cleared: no cycle across the three semaphores
(acquisition is always small → parking → large, and nothing holding a large
permit ever waits for either of the others); parking slots cannot leak or be
released early (`_parked` is a named binding living to the end of the function,
past both exits and cancellation); no path releases the small permit without
cover; both pools FIFO so neither drain worker can be starved by upgraders; lock
ordering; lost wakeups; `QueueState` invariants, so `pop_lane`'s `.expect` and
`debug_assert_eq!` are unreachable; the memory bound of the parked set is real
(`WrappedState` is `Arc`-backed, but `payload` is a genuine copy — ~26 retained
worst case); every `start_paused` assertion forced by the virtual clock rather
than poll order; and all eleven source-scrape pins bounding the region they
intend, including the self-match hazard.

### Mutation matrix (cumulative)

**Eighteen mutations, eighteen detections, no collateral failures.** Rounds 3-4
added:

| # | mutation | test that failed |
|---|---|---|
| 15 | parking slot released as soon as it is taken (bound becomes a no-op) | `parking_area_is_bounded_and_returns_its_slots` |
| 16 | `take_large` leaks the small permit instead of releasing it | `cancelling_a_waiting_send_releases_its_permit_in_either_phase` |
| 17 | `enqueue` derives its lane from a constant, keeping a stray `lane_for` mention | `start_worker_spawns_one_drain_per_lane` |
| 18 | parking pool constructed per dispatch instead of shared | `start_worker_spawns_one_drain_per_lane` |

Mutations 17 and 18 initially passed — the pins matched the mutated code. Both
pins were tightened and re-verified individually.

### Validation on the final content

`cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`,
`cargo check -p freenet --all-targets`,
`cargo check -p freenet --features simulation_tests --lib` — clean.
`cargo test -p freenet --lib` — **4,584 passed, 0 failed**, run twice.

[AI-assisted - Claude]

sanity:
## Review round 4 — no P0/P1; seven P2s fixed

Re-run on the round-3 content (`053d05a32`) with a focused adversarial lens,
told explicitly that rounds 2 and 3 had each found a defect introduced by the
previous round's fix, and to assume this round did it again.

It did not. What it verified, in detail rather than by assertion:

- **The simpler shape is safe.** Round 3's reviewers suggested a `tokio::select!`
  between the large-pool acquire and the parking acquire; I implemented an
  unconditional wait for parking instead. The reviewer proved these equivalent:
  a task takes the parking slot, releases the small permit, and registers on the
  large pool **all within one poll**, with no `.await` between — so at every
  suspension point every parking-slot holder is already a registered `large_pool`
  waiter, and parking being full implies `large_pool.available_permits() == 0`.
  A send waiting for a parking slot therefore cannot miss an idle large permit.
  Ordering is in fact slightly better than the `select!`: parking-denied sends
  enter the large FIFO *behind* the parked ones, so genuine large-lane entries
  queue behind at most 12 upgraders instead of up to 24.
- **No cycle** across the three semaphores, with the release chain traced.
- **Both rewritten tests are armed**, with the specific mutation that fails each
  assertion named, and their determinism under `start_paused` re-derived.
- Permit lifecycle on every exit, `QueueState` invariants, lock ordering, and all
  eleven source-scrape pins (including self-match and sibling-slice hazards).

### The seven P2s, all fixed

1. **Two doc sites still described the behaviour this round removed** — "past the
   parking area's capacity the small-lane slot IS held for the whole wait". That
   is the pre-fix code. Left standing, it is precisely what a future round would
   read as the contract and "restore". Both now say the slot is held only until a
   parking slot frees.
2. **A vacuous assertion in the cancellation test**: `large.available_permits() == 0`
   could not fail, because the test itself held the pool's only permit. The pool
   now has a spare so the assertion can fail.
3. **The same test never checked the slot was still HELD at abort time** in the
   phase-1 case — only that it came back. It now asserts the precondition too.
4. **Cancellation coverage gap created by this round**: phase 2 now has two
   distinct waits, and the new one (queued for a parking slot while still holding
   a small-lane slot) was untested. The test now covers all three states a send
   can be aborted in, driven by an explicit `Phase` enum.
5. **The `enqueue` pin bound the wrong half.** It asserted the lane is *derived*
   from `lane_for`, not that it is the lane actually enqueued with — so deriving
   it and passing a constant would have passed. Both halves pinned now.
6. **Nothing pinned that `drain_lane` passes the SHARED parking area** into each
   send. Constructing one per dispatch there restores the unbounded parked set —
   the exact slip the previous commit's own message claimed would go unnoticed,
   and it would have. Pinned, and verified against that mutation.
7. `.close()` grep note stale (two test hits, not one).

### Mutation matrix (cumulative)

**Twenty-one mutations, twenty-one detections, no collateral failures.** Round 5
added:

| # | mutation | test that failed |
|---|---|---|
| 19 | `drain_lane` builds its own parking pool instead of sharing | `worker_wires_hol_boundaries_and_direct_normal_cleanup` |
| 20 | `enqueue` derives the lane, then enqueues with a constant | `start_worker_spawns_one_drain_per_lane` |
| 21 | the parking slot is forgotten instead of released | `cancelling_...`, `parking_area_is_bounded_and_returns_its_slots`, `large_pool_closing_mid_park_...` |

Two of the pins added in round 4 (17 and 18 in the matrix) initially passed their
own mutations and had to be tightened — which is why every pin here is verified
by running the mutation rather than by reading the assertion.

### Validation on the final content

`cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`,
`cargo check -p freenet --all-targets`,
`cargo check -p freenet --features simulation_tests --lib` — clean.
`cargo test -p freenet --lib` — **4,584 passed, 0 failed**, run twice.

Four review rounds, sixteen lens-passes in total (4 + 4 + 2 + 1, each blind and
independent, each reading the checked-out code rather than the diff). Every
finding is fixed above or dismissed with a stated reason. The one thing I would
still want and did not add is a counter for the correction path's wait and the
parking area's occupancy — it needs a `NetworkEfficiencyV1` wire-schema change,
and `queued_small_actual_large` is its leading indicator in the meantime.

[AI-assisted - Claude]

Issue #5049, opened 2026-07-30
never_populated (89% of the tracked no-summary arm) conflates cold entries with self-inflicted clobbers; two unguarded register_peer_interest sites


sanity:
## Summary

A peer can only send a **delta** if it holds a cached summary for the recipient. That cache is seeded from a successful delivery. But 22.2% of stream transfers currently fail, and a failed send is precisely the case that does **not** seed the cache. So a contract whose state is large enough to stream can get permanently stuck sending full state, and the mechanism that would rescue it is the mechanism that keeps failing.

This is currently **15.1% of all broadcast bytes** (45.8 GB in a ~5h window), of which **89.1% is tagged `never_populated`**.

## Measured

Fleet, v0.2.115, 1,197 peers, 2,864 node-hours, 303 GB of broadcast payload (`broadcast_payload_mix`):

| arm | share of broadcast bytes | bytes | sends | mean |
|---|---|---|---|---|
| `delta` | 68.4% | 207.3 GB | 8,958,373 | 23.1 KB |
| **`full_no_their_summary_tracked`** | **15.1%** | **45.8 GB** | 128,825 | **355.6 KB** |
| `full_no_their_summary_untracked` | 11.2% | 33.9 GB | 259,942 | 130.5 KB |
| `full_delta_suppressed` | 4.5% | 13.6 GB | 4,454,087 | 3.1 KB |
| `full_compute_failed` | 0.5% | 1.4 GB | 1,564 | 918.1 KB |
| `full_no_our_summary` | 0.3% | 0.8 GB | 944 | 815.8 KB |
| `full_not_efficient` | 0.0% | 1.4 MB | 12 | 114.4 KB |

Why the tracked arm has no summary (`tracked_missing_*`, total 45.8 GB):

| reason | share | bytes | sends |
|---|---|---|---|
| **`never_populated`** | **89.1%** | **40.8 GB** | 109,806 |
| `none_report` | 10.3% | 4.7 GB | 8,350 |
| `resync` | 0.6% | 256.7 MB | 3,542 |
| `delta_apply_failed` | 0.1% | 41.8 MB | 7,108 |
| `unattributed` | 0.0% | 3.4 MB | 19 |

Transport, same window: `transfer_completed` 249,534 vs `transfer_failed` 71,166 = **22.2% failure**, essentially all `cwnd wait timeout after 3s`.

[...]

PR #5109, opened 2026-08-03
fix(ring): retain peer summaries across disconnect-grace interest removal

sanity:
## Problem

A dropped connection schedules a deferred interest removal
(`INTEREST_DISCONNECT_GRACE_PERIOD`, 90s). When the grace lapses,
`remove_peer_interest_for` deletes the whole `PeerInterest` — **cached summary
included**. The peer's identity is its transport public key
(`PeerKey(TransportPublicKey)`), which is stable across a reconnect from a new
address and across a restart, so the belief we held is still about the same peer
when it comes back. But the recreated entry starts summary-less, and a
summary-less entry is a **missing input** to `get_state_delta`, so the next
broadcast to that peer ships FULL STATE.

## Evidence (0.2.118 fleet, 1,414 reporting peers, `network_efficiency_v1`)

Removal causes — `DisconnectGrace` is not a contributor, it is the mechanism:

| cause | removals | later recreated (≤20 min) |
|---|---|---|
| **DisconnectGrace** | **19,271,751 (82%)** | **2,645,774** |
| InterestsReplace | 2,896,191 | 1,316,229 |
| ChangeInterests | 684,897 | 665,794 |
| TtlExpiry | 498,714 | 58,229 |
| Unsubscribe | 216,855 | 107,509 |

Missing-summary full-state sends (cumulative, latest snapshot per peer):

| class | sends | bytes | mean |
|---|---|---|---|
| UntrackedFirstObserved | 771,400 | 78.20 GB | 101 KiB |
| TrackedFirstNew | 258,021 | 77.35 GB | 300 KiB |
| TrackedFirstRecreated | 25,704 | 7.72 GB | 300 KiB |
| UntrackedFirstRecreated | 21,150 | 6.53 GB | 309 KiB |
| every `*Repeat*` class | 186 | 31 MB | — |

**Timing** — `ms_age`, the age of the summary-less entry at its first send:
164,867 under 1s (58%), 54,655 at 1–9s (19%), 13,770 at 10–59s, 13,707 at
1–5 min, 35,728 at ≥5 min. Nearly four in five first-sends beat the peer's
`Summaries` reply, which is the window a retained summary short-circuits.

The recreation counters are a **floor**: `recent_removal` is only honoured inside
a 20-minute window, and the bounded history LRU overflowed 1,597,674 times
fleet-wide. Total removals (23.6M) and total new-summary-less registrations
(28.6M) are in the same range, so most creations must be recreations — only 4.79M
of which are detected as such.

## The fix

A bounded LRU preserves the cached summary when an interest entry is torn down,
and seeds the recreated entry from it. Retention requires **two** conditions.

**The cause must be `DisconnectGrace`.** That is the one removal where the peer
never withdrew its interest — the transport dropped and nothing the peer said
invalidated our belief. The withdrawal causes are removals where the peer told us
it stopped tracking the contract, so a re-registration may follow a re-fetch that
moved its state somewhere our belief does not describe.

**The provenance must be peer-asserted.** A summary written by
`record_delivery_to_interest` — our own summary cached as theirs on *sender-side*
send completion — is wrong whenever a stream tail is lost, and transport loss is
exactly what puts us on the retention path, so the disconnecting population is
enriched in such beliefs. Today's unconditional teardown is a real repair for
them: `broadcast_to_single_peer`'s converged-skip sends **nothing** when the
cached belief matches our own summary, so a wrong-and-matching belief is silent,
and the teardown is what breaks that silence. Retention keeps that repair intact
and costs almost nothing: `Delivery` is 23.8M of 585M summary writes fleet-wide;
`DigestAgreement`, `InterestSummary` and `InboundBroadcast` — all peer-asserted —
are the rest.

Every removal that does *not* store **discards** instead, and
`upsert_peer_summary_from` discards when it recreates an entry. That makes the
scope exact: a resident retained entry always comes from the pair's most recent
teardown, and that teardown was a disconnect.
`remove_peer_interest_for` is the single removal chokepoint —
`sweep_expired_interests`, `remove_evicted_in_use` and
`remove_all_peer_interests_for` all delegate to it — so no path drops an entry
without passing through this decision.

**Bounded on entries and bytes.** The key is peer-controlled, so an unbounded map
here would be an amplification vector (`.claude/rules/code-style.md`).
`RETAINED_SUMMARY_CACHE_SIZE = 16384` is sized from the fleet — the
`interest_sync_summaries_*` rollup measures a **262-byte mean per `Summaries`
entry** over 228M entries, so the cap costs ~4.3 MB of payload plus ~3 MB of keys
and LRU nodes — and `RETAINED_SUMMARY_CACHE_MAX_BYTES = 8 MiB` bounds the case
where a node's summaries run large. Eviction fails **safe**: forgetting a retained
summary only restores today's behaviour for that pair.

## Why a stale summary is safe here

The retained summary is always one the **peer itself asserted** about its own
state, merely older. A cached peer summary is a belief rather than live evidence
everywhere in this module, and two backstops correct wrong ones: the InterestSync
summary/digest exchange, and the delta-apply-failure → `ResyncRequest` path.
`RETAINED_SUMMARY_TTL` (= `INTEREST_TTL`) bounds how much staleness this *adds* —
a live entry's summary already has no age bound at all, because the converged-skip
refreshes the interest TTL without refreshing the summary.

The two directions of wrongness are not symmetric:

- **Peer moved ahead** (the common case — it kept syncing with others while
  disconnected from us): the delta is a superset of what it needs, and if it is
  not smaller than our state the post-compute efficiency gate (#4923) discards it
  and we send full state exactly as today. On the send itself this can never cost
  more bytes than current behaviour.
- **Peer moved back** (lost, evicted, or re-fetched state — the transport keypair
  survives a db-only wipe, so this *is* reachable): we send too small a delta. If
  the contract rejects it, the total cost is delta + `ResyncRequest` +
  full-state `ResyncResponse` — two round trips **more** than sending full state
  up front — and repeated `Invalid`-class rejections feed the per-*contract*
  `delta_incompat` memo, which then forces full state to every peer of that
  contract for its TTL. That is the honest cost ceiling of being wrong.
  Production says it is rarely paid: `ClearedByDeltaApplyFailure` currently holds
  **32** entries fleet-wide and `ClearedByResync` **413**, against 12.84M known
  summaries — and restricting retention to peer-asserted summaries keeps it that
  way.

## The first broadcast after a restore, when the retained summary is stale

This is the question the whole change turns on, because a retained-but-wrong
summary is worse than no summary if recovery is not guaranteed — no summary at
least produces a correct full-state send. **It costs at most one wasted delta and
then self-heals; it cannot wedge the pair.** Three properties make that true, and
each is a specific code path rather than an argument:

1. **The restore is single-use.** `take_retained_summary` *pops* the entry
   (`interest.rs`), so one retained summary seeds exactly one recreated entry.
   There is no path that re-applies the same belief on a later broadcast.

2. **A rejected delta clears the belief and forces full state.** The receiver's
   `Invalid`-class rejection drives
   `operations/update/op_ctx_task.rs` → `clear_peer_summary(ClearedByDeltaApplyFailure)`
   + `ResyncRequest` → full-state `ResyncResponse`. Cost: one delta + two round
   trips more than sending full state up front.

3. **The recovery cannot re-arm the bad belief.** The full-state resend that
   follows is cached by `record_delivery_to_interest` with
   `SummaryPopulationSource::Delivery` → `SummaryProvenance::SelfAssumed`, which
   this change refuses to retain. So a stale belief that failed once cannot be
   preserved across the next disconnect and replayed. That closes the only shape
   that could have looped.

The silent-divergence variant — the contract *accepts* a too-small delta and ends
up missing state, with no rejection to trigger (2) — is bounded by the
InterestSync exchange, and load-bearingly so: the `Summaries` arm in `node.rs`
overwrites our cached belief with the peer's own report **unconditionally**,
outside the staleness branch (pinned by
`summaries_arm_writes_summary_outside_staleness_branch_pin`), and only then emits
the targeted `SyncStateToPeer` heals. So the correction does not depend on the
stale detection agreeing; ground truth replaces the belief on the next exchange
either way, and the reconnect itself triggers one.

Net versus the full-state send it replaces: strictly better when the peer moved
ahead (the common case), and bounded by one wasted delta plus two round trips
when it moved back — which is why retention is restricted to summaries the peer
itself asserted, since that is what keeps the "moved back" population small.

## Secondary fix: the untracked staleness reset hid recreation

`begin_peer_summary_broadcast`'s untracked arm reset `recent_removal` whenever
`last_observed` was older than `INTEREST_TTL`. Those are different clocks —
`last_observed` is stamped *only* on that untracked path, so it says nothing about
when the interest entry was removed, and the removal's own freshness is already
enforced by the `removed_at` filter one line below. A pair observed untracked at
T0, tracked, removed at T1, and broadcast to again at T2 > T0+20min lost its T1
removal and reported `UntrackedFirstObserved`.

That is why `UntrackedFirstObserved` is the largest class on the fleet (771K
sends / 78.2 GB) while `UntrackedRepeatSequential` is ~zero: churn cycles longer
than `INTEREST_TTL` all landed in the "never seen before" bucket. The counter
could not answer "genuinely never tracked, or tracked-then-dropped?"; now it can.
No routing changes.

## Telemetry — read this before interpreting the rollup

`NetworkEfficiencyV1.retain: [u64; 6]` —
`[stored, skipped_oversized, restored, expired, evicted, discarded]`.
`Restored + Expired + Discarded + Evicted` accounts for **every** stored entry
that has left the cache, including one replaced by a later store for the same
pair, so `stored − departures` is exactly what is resident. `restored` counts
pairs *seeded* from the cache (whether the resulting broadcast was a delta is
decided later and reported by `ms_s`); `evicted` is the resize signal for
`RETAINED_SUMMARY_CACHE_SIZE`.

`recreated_after_removal` shifts for **two** independent reasons in this PR, and
both are intentional: (1) it now also counts recreations that arrive *with* a
summary, so the series keeps measuring the same population instead of appearing
to collapse exactly as retention starts succeeding; and (2) the staleness-reset
fix above means a pair whose removal is recent but whose last untracked
observation is stale now counts as recreated where it previously did not. Expect a
level shift, not a trend. `reg_new_m` deliberately still describes what the caller
supplied, so it stays comparable with pre-retention field data.

## Tests

19 new unit tests in `ring::interest::tests`, covering: the reconnect round trip
asserted on the production read path (`begin_peer_summary_broadcast`, not just the
accessor); a restored belief surviving a second disconnect; self-assumed summaries
*not* retained; every population source's retention eligibility; within-grace
reconnect still owned by the cancel path; the scope cut iterated over
`InterestRemovalCause::ALL`; the interleaving case (disconnect-store → upsert
recreate → withdrawal removal → re-register must **not** restore); TTL at the
boundary and one tick past; caller-supplied supersession; existing-entry
supersession; `recreated_after_removal` continuity; entry bound at, over, and
under capacity; byte bound; oversized rejection at exactly the bound and one byte
over; empty summaries; peer-identity and per-contract keying; a 4-thread
store/take/discard race; the telemetry array-order pin; and the untracked
staleness-reset regression. A shared `assert_retention_bookkeeping` helper pins
the counter-vs-residency identity and `total_bytes` against a recomputed sum
after every scenario that churns the cache — an under-counted `total_bytes` is
the dangerous direction, since the cache would then exceed its memory budget
while reporting compliance.

### Why a `test(priority-select)` commit is riding in a `fix(ring)` PR

Scope creep is a fair thing to flag here, so: it was blocking the full
`cargo test -p freenet --lib` run on this branch, which is one of the gates for
this change, and it is unrelated to the change itself.
`priority_select::tests::test_priority_select_event_loop_simulation` bounds each
poll at 50ms of REAL time while its producer task sleeps 10ms before its first
send, leaving iteration 0 a ~40ms scheduling margin. It failed on the dev box at
load average ~70 (four parallel release builds, not this branch) and passes 3/3
in isolation, so it is latently broken rather than newly broken, and is unlikely
to be failing on GitHub's runners.

The bound is a **hang guard**, not a latency assertion: a waker-registration
regression makes the poll hang forever, so a generous bound fails identically to
a tight one. Widening it therefore costs no real coverage. It is now a 10s guard
with the reasoning recorded inline, in its own commit so it can be reverted
independently. The same file has ~20 more wall-clock timeouts of this shape —
filed as **#5110** with the per-site analysis rather than swept in here, because
several of them deliberately assert "nothing is ready yet" and want a different
fix (`tokio::time::pause`), not a wider bound.

## What this does NOT fix

- `UntrackedFirstObserved` / `UntrackedFirstRecreated` (84.7 GB): those sends
  happen *before* any registration, so the restore hook is not on that path.
  Whether the same cache should also be consulted on the untracked broadcast arm
  is a separate change with its own safety question.
- The non-disconnect removal causes (4.3M removals, 2.15M recreations).
- The transfer-failure loop in #5049, which is about summaries that were never
  populated rather than ones that were destroyed.
- Benefit is bounded by a race: on reconnect the entry is recreated either by the
  `Interests` registration (restore fires) or by the peer's `Summaries` reply via
  `upsert_peer_summary_from` (ground truth, strictly better, restore discarded).
  Retention only pays when registration wins, which is the ~1 RTT window the
  `ms_age` data says four in five first-sends fall inside.

## Note on #5049

The "self-inflicted clobber" half of #5049 is **not supported by production data**.
`reg_ow_k` — registrations that overwrote an already-KNOWN summary — totals **613**
fleet-wide against 28.6M new-summary-less registrations, and every production
`register_peer_interest` site goes through `register_peer_interest_from` behind a
refresh-if-present guard (the bare-call sites are all `#[cfg(test)]`, pinned by
`p2p_protoc.rs`). The `never_populated` population is churn, not clobbering.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -1215,13 +1215,23 @@ pub(super) async fn broadcast_to_single_peer(
             // peer (`interested_peers`), not evidence from it. So a wedged but
             // still-connected peer whose cached summary happens to match ours
             // gets refreshed indefinitely, where the TTL previously reaped it.
-            // Two backstops bound that: disconnect drops the entry outright
-            // (`InterestManager::remove_peer`), and the peer's own ~5-min
-            // `Interests` heartbeat is a full REPLACE (node.rs), so an entry the
-            // peer no longer claims is removed regardless of how recently we
-            // refreshed it. TTL expiry is therefore not the mechanism that
-            // reaps a live-but-wedged peer, and using it as one costs every
+            // Two backstops bound that: the disconnect-grace teardown drops the
+            // entry (`InterestManager::execute_pending_removals`), and the peer's
+            // own ~5-min `Interests` heartbeat is a full REPLACE (node.rs), so an
+            // entry the peer no longer claims is removed regardless of how
+            // recently we refreshed it. TTL expiry is therefore not the mechanism
+            // that reaps a live-but-wedged peer, and using it as one costs every
             // converged peer its updates.
+            //
+            // The first backstop is now PARTIAL, and deliberately so. The
+            // teardown preserves the cached summary across the disconnect when
+            // the PEER asserted it (`retain_summary_for_disconnect`), so a
+            // reconnecting peer can be sent a delta instead of a full state. It
+            // still drops a `SelfAssumed` belief — the one this comment is about,
+            // written by `record_delivery_to_interest` below on sender-side send
+            // completion and therefore wrong exactly when a stream tail is lost.
+            // That is the case where a matching-but-wrong belief makes this skip
+            // silent, so that case keeps its unconditional repair.
             //
             // Refresh ONLY. Do not cache the summary or record delivery
             // telemetry here: nothing was delivered, and `sent_delta` has no
--- a/crates/core/src/ring.rs
+++ b/crates/core/src/ring.rs
@@ -2202,6 +2202,7 @@ impl Ring {
                     tel,
                     shadow,
                     eff: telemetry.network_efficiency_delivery,
+                    retain: lifecycle.retained_summaries,
                 });
             }
 
--- a/crates/core/src/ring/interest.rs
+++ b/crates/core/src/ring/interest.rs
@@ -169,6 +169,54 @@ const MISSING_SUMMARY_HISTORY_SIZE: usize = 65536;
 /// speculatively.
 const MISSING_SUMMARY_ACTIVE_SIZE: usize = 256;
 
+/// How long a summary preserved across a disconnect-grace removal stays usable.
+///
+/// Deliberately equal to [`INTEREST_TTL`]: a LIVE interest entry can already
+/// carry a summary this old before the sweep expires it (nothing refreshes the
+/// cached summary itself — `refresh_peer_interest` only bumps the TTL), so a
+/// retained summary is never staler than a belief the broadcast path already
+/// acts on today. Bounding it makes the added staleness explicit rather than
+/// open-ended.
+const RETAINED_SUMMARY_TTL: Duration = INTEREST_TTL;
+
+/// Max (contract, peer) pairs whose summary survives a disconnect-grace removal.
+///
+/// The key is peer-controlled (a remote peer picks which contracts it registers
+/// interest in, and disconnecting is free), so this MUST be bounded — see the
+/// per-key-collection rule in `.claude/rules/code-style.md`. LRU eviction fails
+/// SAFE: forgetting a retained summary only restores today's behaviour for that
+/// pair (one full-state send), it never produces a wrong delta.
+///
+/// Sized against the fleet rather than guessed: the 0.2.118 outbound rollup
+/// (`interest_sync_summaries_*`, 1,414 peers, 228M entries) measured a **262-byte
+/// mean per `Summaries` entry**, so this cap costs ~4.3 MB at the typical size
+/// and the byte budget below bounds the atypical one. As with
+/// [`MISSING_SUMMARY_HISTORY_SIZE`], the resize signal is the visible overflow
+/// counter (`NetworkEfficiencyV1::retain`, `Evicted` slot), not a re-derivation
+/// from first principles: a saturated LRU of this shape evicts continuously, so
+/// that counter reports steady-state pressure, never reaching zero at any finite
+/// cap.
+const RETAINED_SUMMARY_CACHE_SIZE: usize = 16384;
+
+/// Total payload budget for retained summaries.
+///
+/// An entry count alone does not bound memory here: summary bytes are contract-
+/// defined and, for entries seeded from a peer's own `Summaries` report,
+/// remote-influenced. The cache trims least-recently-used entries until it is
+/// under this budget, so a few large summaries cannot inflate node memory. At the
+/// measured 262-byte mean this never binds; it binds only when a node's summaries
+/// average >512 B, which is exactly the case an entry-only cap would miss.
+const RETAINED_SUMMARY_CACHE_MAX_BYTES: usize = 8 * 1024 * 1024;
+
+/// Largest single summary worth retaining.
+///
+/// Above this, one entry would monopolise
+/// [`RETAINED_SUMMARY_CACHE_MAX_BYTES`], and a summary that large is unlikely
+/// to yield a delta smaller than the state anyway (the post-compute efficiency
+/// gate would discard it — see [`is_delta_efficient`] and #4923). Skipping is
+/// the same outcome as today: full state on the next send.
+const MAX_RETAINED_SUMMARY_BYTES: usize = 64 * 1024;
+
 /// Telemetry field order for the first missing-summary send age histogram:
 /// <1s, 1-9s, 10-59s, 60-299s, and >=300s.
 pub(crate) const FIRST_MISSING_SUMMARY_SEND_AGE_LABELS: [&str; 5] =
@@ -506,6 +554,144 @@ struct MissingPairHistory {
     recent_removal: Option<(InterestRemovalCause, Instant)>,
 }
 
+/// Fate of one entry in the across-disconnect summary retention cache.
+///
+/// `Stored + SkippedOversized` accounts for every disconnect-grace removal that
+/// carried a summary; `Restored + Expired + Discarded + Evicted` accounts for
+/// every stored entry that has left the cache, so the residual is what is still
+/// resident. `Restored` is the counter that measures whether the retention is
+/// actually buying full-state sends back.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum RetainedSummaryOutcome {
+    /// A disconnect-grace removal's cached summary was preserved.
+    Stored,
+    /// The removed entry's summary exceeded [`MAX_RETAINED_SUMMARY_BYTES`].
+    SkippedOversized,
+    /// A recreated pair was seeded from its retained summary.
+    Restored,
+    /// A retained summary was found but was older than [`RETAINED_SUMMARY_TTL`].
+    Expired,
+    /// Dropped under entry-count or byte-budget pressure before it was claimed.
+    Evicted,
+    /// Popped without being used: the registration brought its own summary, or
+    /// it targeted an entry that already existed.
+    Discarded,
+}
+
+impl RetainedSummaryOutcome {
+    pub(crate) const COUNT: usize = 6;
+    pub(crate) const ALL: [Self; Self::COUNT] = [
+        Self::Stored,
+        Self::SkippedOversized,
+        Self::Restored,
+        Self::Expired,
+        Self::Evicted,
+        Self::Discarded,
+    ];
+
+    pub(crate) const fn index(self) -> usize {
+        self as usize
+    }
+
+    pub(crate) const fn as_str(self) -> &'static str {
+        match self {
+            Self::Stored => "stored",
+            Self::SkippedOversized => "skipped_oversized",
+            Self::Restored => "restored",
+            Self::Expired => "expired",
+            Self::Evicted => "evicted",
+            Self::Discarded => "discarded",
+        }
+    }
+}
+
+/// A peer summary preserved across a disconnect-grace interest removal.
+#[derive(Clone, Debug)]
+struct RetainedSummary {
+    summary: StateSummary<'static>,
+    retained_at: Instant,
+}
+
+/// Bounded LRU of summaries preserved across disconnects, capped on BOTH entry
+/// count and total payload bytes.
+///
+/// The count cap alone would not bound memory: summary bytes are contract-
+/// defined and partly remote-influenced (a peer's own `Summaries` report seeds
+/// most entries), which is exactly the amplification shape
+/// `.claude/rules/code-style.md` bans for peer-controlled keys.
+struct RetainedSummaryCache {
+    entries: LruCache<(ContractKey, PeerKey), RetainedSummary>,
+    total_bytes: usize,
+}
+
+impl RetainedSummaryCache {
+    fn new() -> Self {
+        Self {
+            // Unbounded LRU + explicit trimming: `LruCache`'s own capacity
+            // eviction is silent, and a silent eviction would desynchronise
+            // `total_bytes` from the map AND lose the `Evicted` counter that
+            // tells us whether the cache is too small in production.
+            entries: LruCache::unbounded(),
+            total_bytes: 0,
+        }
+    }
+
+    /// Store `summary` for `key`, returning how many entries were evicted to
+    /// make room.
+    fn store(
+        &mut self,
+        key: (ContractKey, PeerKey),
+        summary: StateSummary<'static>,
+        now: Instant,
+    ) -> u64 {
+        let bytes = summary.as_ref().len();
+        if let Some(previous) = self.entries.put(
+            key,
+            RetainedSummary {
+                summary,
+                retained_at: now,
+            },
+        ) {
+            // Same key replaced: not an eviction, just a byte-count fixup.
+            self.total_bytes = self
+                .total_bytes
+                .saturating_sub(previous.summary.as_ref().len());
+        }
+        self.total_bytes = self.total_bytes.saturating_add(bytes);
+        self.trim()
+    }
+
+    /// Evict least-recently-used entries until both caps hold. Never evicts the
+    /// entry just inserted below one entry, so a single oversized summary
+    /// cannot spin here (callers reject those before calling `store`).
+    fn trim(&mut self) -> u64 {
+        let mut evicted = 0;
+        while self.entries.len() > RETAINED_SUMMARY_CACHE_SIZE
+            || (self.total_bytes > RETAINED_SUMMARY_CACHE_MAX_BYTES && self.entries.len() > 1)
+        {
+            match self.entries.pop_lru() {
+                Some((_, dropped)) => {
+                    self.total_bytes = self
+                        .total_bytes
+                        .saturating_sub(dropped.summary.as_ref().len());
+                    evicted += 1;
+                }
+                None => break,
+            }
+        }
+        evicted
+    }
+
+    /// Remove and return the retained entry for `key`, if any.
+    fn take(&mut self, key: &(ContractKey, PeerKey)) -> Option<RetainedSummary> {
+        let retained = self.entries.pop(key)?;
+        self.total_bytes = self
+            .total_bytes
+            .saturating_sub(retained.summary.as_ref().len());
+        Some(retained)
+    }
+}
+
 /// Fixed-cardinality lifecycle counters copied into telemetry snapshots.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct InterestLifecycleSnapshot {
@@ -524,6 +710,9 @@ pub(crate) struct InterestLifecycleSnapshot {
     pub(crate) current_summary_state: [u64; SummaryMissingReason::COUNT + 1],
     pub(crate) history_overflow: u64,
     pub(crate) active_overflow: u64,
+    /// Across-disconnect summary retention outcomes, in
+    /// `RetainedSummaryOutcome::ALL` order.
+    pub(crate) retained_summaries: [u64; RetainedSummaryOutcome::COUNT],
 }
 
 struct InterestLifecycleMetrics {
@@ -540,6 +729,7 @@ struct InterestLifecycleMetrics {
     registration_cap_rejected: [AtomicU64; InterestRegistrationSource::COUNT],
     history_overflow: AtomicU64,
     active_overflow: AtomicU64,
+    retained_summaries: [AtomicU64; RetainedSummaryOutcome::COUNT],
 }
 
 pub(crate) struct MissingSummaryAttempt {
@@ -574,6 +764,7 @@ impl InterestLifecycleMetrics {
             registration_cap_rejected: std::array::from_fn(|_| AtomicU64::new(0)),
             history_overflow: AtomicU64::new(0),
             active_overflow: AtomicU64::new(0),
+            retained_summaries: std::array::from_fn(|_| AtomicU64::new(0)),
         }
     }
 }
@@ -1038,6 +1229,21 @@ pub struct InterestManager<T: TimeSource> {
     /// can race past `MISSING_SUMMARY_ACTIVE_SIZE` (bounded by the number of
     /// concurrent racers, not fixed at one).
     missing_summary_active: DashMap<(ContractKey, PeerKey), u16>,
+
+    /// Peer summaries preserved across a `DisconnectGrace` interest removal.
+    ///
+    /// A dropped connection destroys the whole `PeerInterest`, summary included
+    /// (`remove_peer_interest_for`). The peer's IDENTITY is its transport public
+    /// key ([`PeerKey`]), which is stable across a reconnect from a new address
+    /// and across a restart (the keypair is persisted alongside the state store,
+    /// see `config::secret`), so the belief we held about that peer is still
+    /// about the same peer when it comes back — but a summary-less recreated
+    /// entry forces a FULL STATE on the next broadcast. This cache carries the
+    /// belief across the gap; see [`Self::take_retained_summary`] for why that
+    /// is safe.
+    ///
+    /// Bounded on entries AND bytes; see [`RetainedSummaryCache`].
+    retained_summaries: Mutex<RetainedSummaryCache>,
     interest_lifecycle_metrics: InterestLifecycleMetrics,
 }
 
@@ -1092,6 +1298,7 @@ impl<T: TimeSource + Sync> InterestManager<T> {
                     .expect("MISSING_SUMMARY_HISTORY_SIZE must be > 0"),
             )),
             missing_summary_active: DashMap::new(),
+            retained_summaries: Mutex::new(RetainedSummaryCache::new()),
             interest_lifecycle_metrics: InterestLifecycleMetrics::new(),
         }
     }
@@ -1224,8 +1431,21 @@ impl<T: TimeSource + Sync> InterestManager<T> {
             .last_observed
             .is_some_and(|observed| now.saturating_duration_since(observed) > INTEREST_TTL)
         {
+            // Reset the SEND counter only. `recent_removal` must NOT be wiped
+            // here: `last_observed` is stamped exclusively on this untracked
+            // path, so it says nothing about when the pair's interest entry was
+            // removed, and the removal's own freshness is already enforced by
+            // the `removed_at` filter below.
+            //
+            // Wiping it conflated the two clocks and systematically hid
+            // recreation: a pair observed untracked at T0, then tracked, then
+            // removed at T1, then broadcast to again at T2 > T0 + INTEREST_TTL
+            // lost its T1 removal and was reported `UntrackedFirstObserved`
+            // instead of `UntrackedFirstRecreated` — which is why that class is
+            // the largest on the fleet (771K sends / 78.2 GB on 0.2.118) while
+            // `UntrackedRepeatSequential` is ~zero: churn cycles longer than
+            // INTEREST_TTL all landed in the "never seen before" bucket.
             record.send_starts = 0;
-            record.recent_removal = None;
         }
         let first = record.send_starts == 0;
         let recreated = record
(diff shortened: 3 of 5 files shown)


sanity:
## Multi-perspective review (FULL tier) — 4 independent blind reviewers

This change touches state authorization and protocol behaviour, so it took the Full tier: four subagent reviewers, each reading the checked-out code with no sight of the others, with distinct lenses — **code-first** (form a model from the code, then diff it against the description), **testing** (coverage gaps and vacuous tests), **skeptical** (adversarial bug hunt), and **stale-summary correctness** (the failure mode specific to this change: can a retained belief produce an unappliable delta, silent divergence, or an unbounded/poisonable cache). No external-model pass; per `multi-model-review.md` the Claude lenses are the review.

Three of the four independently found the same design defect, which was real. Every finding below is either fixed in `a717adfb` or dismissed with a reason. Every cited line was re-read before acting.

### Blocking — fixed

**B1. The restore was not scoped, only the store was.** `restore_wanted = is_new && summary.is_none()` never consulted the pair's removal cause. On the real reconnect path the entry is usually recreated by `upsert_peer_summary_from` (the peer's `Summaries` reply / digest agree / delivery) rather than by registration — every production registration site is refresh-if-present — so the retained entry was **orphaned**. A later withdrawal removal (`Unsubscribe` / `InterestsReplace` / `ChangeInterests`), which deliberately retains nothing, could then be followed by a registration that restored the orphan. That resurrects a belief across exactly the causes the scope exists to refuse, and `retained_summary_is_not_applied_to_an_existing_entry` was vacuous against it because it called `register_peer_interest_from` on an existing entry — a sequence production never performs.

Fixed by making the invariant exact rather than incidental: every removal that does not store now **discards**, and `upsert_peer_summary_from` discards when it recreates an entry. A resident retained entry therefore always comes from the pair's most recent teardown, and that teardown was a disconnect. `remove_peer_interest_for` is the single removal chokepoint (`sweep_expired_interests`, `remove_evicted_in_use`, `remove_all_peer_interests_for` all delegate to it), so no path drops an entry without passing through the decision. New test `a_withdrawal_removal_cannot_resurrect_a_belief_retained_at_an_earlier_disconnect` drives the full interleaving for all four withdrawal causes.

**B2. Retention silently deleted a backstop another module's comment depends on.** `broadcast_queue.rs`'s converged-skip sends **nothing** when the cached belief matches our summary, and its comment names "disconnect drops the entry outright" as one of two backstops for a wrong-but-matching belief. `record_delivery_to_interest` writes our own summary as the peer's on *sender-side* completion — wrong whenever a stream tail is lost — and transport loss is precisely what puts us on the retention path, so the disconnecting population is enriched in exactly those beliefs.

Fixed by restricting retention to **peer-asserted** provenance (`SummaryProvenance::PeerReported`: `InterestSummary`, `DigestAgreement`, `InboundBroadcast`, `ResyncResponse`; anything else, including a future source added without thought, defaults to `SelfAssumed`). A `SelfAssumed` belief still gets today's unconditional teardown, so the repair is intact where it matters. Cost is ~4% of the population: `Delivery` is 23.8M of 585M summary writes fleet-wide. The `broadcast_queue.rs` comment is updated to say what is now true. This also materially narrows the "restored belief suppresses the reconnect-time `SyncStateToPeer` heal" finding — the skip now fires only on evidence the peer gave us, never on our own assumption.

### Major — fixed

**M1. `MAX_RETAINED_SUMMARY_BYTES = 64 KiB` excluded the population the change targets, on a justification the codebase forbids.** The doc cited `is_delta_efficient` / the post-compute gate, but that function's own rustdoc records that the summary-size proxy was removed as a pre-compute refusal in #4923 (it made full-state sends 41% of all wire bytes) and "do not re-wire it as a pre-compute refusal"; and the post-compute gate keys on the **delta's** size, not the summary's — #4923's own pin test shows a state-sized summary yielding a 3-byte delta. Raised to 1 MiB, above the 550-840 KB large-state population this change exists for, and the justification rewritten as what it actually is: a per-entry share of the byte budget.

**M2. Deep clone before the eligibility check.** `interest.summary.clone()` on a `Cow<[u8]>` ran unconditionally, then the size was measured and the copy thrown away — per contract of every departing peer, under the `interested_peers` shard write guard. Eligibility (cause, provenance, size) is now decided on `&` before any clone.

**M3. The documented counter identity was false.** A same-key `store` replaced an entry with no departure counted, so `stored − departures` over-reported residency. `store` now reports `replaced` and the caller counts it `Discarded`. Pinned by `assert_retention_bookkeeping`, which also checks `total_bytes` against a recomputed sum — a mutation test confirmed that deleting the `total_bytes` decrement in `take()` previously left all 87 tests green, which would have silently disabled retention fleet-wide while `Evicted` climbed.

**M4. `recreated_after_removal` moves for two reasons, only one documented.** The `else if is_new` arm is one; the staleness-reset fix is the other (it feeds `recent_removal_cause`). Both are intentional; the PR body now states the combined effect so the field series is read as a level shift, not a trend.

### Minor — fixed

- Restored provenance is preserved, so a peer that reconnects and disconnects again before reporting a fresh summary does not silently lose retention eligibility (`a_restored_summary_survives_a_second_disconnect`).
- `RETAINED_SUMMARY_TTL`'s "never staler than a live entry" claim was wrong — the converged-skip refreshes the interest TTL without refreshing the summary, so a live entry's summary age is already unbounded. The doc now says the TTL bounds the staleness this **adds**, and states that enforcement is lazy with no sweep.
- `take_retained_summary`'s "can never cost more bytes than today" is now scoped to the send itself, with the real ceiling stated: on an `Invalid`-class rejection the cost is delta + `ResyncRequest` + full-state `ResyncResponse` (two round trips **more** than full state up front), and repeated rejections feed the per-**contract** `delta_incompat` memo that forces full state to every peer of that contract for its TTL.
- `trim`'s "cannot spin" rationale replaced with the real one (`pop_lru` strictly decreases `len`; the `len() > 1` term is what stops it evicting the entry just inserted).
- `router.rs`'s `restored` description no longer claims a full-state send was skipped — that is decided later by the efficiency gate and reported by `ms_s`.
- Memory estimate now counts keys and LRU nodes (~3 MB), not only payload.
- "keypair persisted alongside the state store" corrected — it is a sibling directory, so a db-only wipe or a contract eviction returns the same `PeerKey` with different state. That case is now named explicitly in the safety argument rather than dismissed.
- `RetainedSummaryOutcome::ALL` / `as_str()` had no callers, so nothing pinned that the telemetry array's positional decoding matched `ALL`. Added `retained_summary_outcome_indices_and_labels_are_distinct`, matching the existing `summary_missing_reason_*` guard.

### Test gaps — fixed

The scope test now iterates `InterestRemovalCause::ALL` instead of a hand-written list, so a future variant is covered when added. Added: exact-TTL boundary vs one tick past; exactly-at-capacity (evicts nothing) alongside over-capacity; oversized at exactly the bound vs one byte over; empty summaries; per-source retention eligibility; the reachable orphan/interleaving shape; a 4-thread store/take/discard race asserting the bookkeeping invariant; and the core regression now asserts on the production read path (`begin_peer_summary_broadcast` returning `Known`) rather than only the `get_peer_summary` accessor.

**Verified against a disabled-fix build: 17 of the 21 new tests fail without the fix.** The four that pass are pure scope/invariant guards (within-grace reconnect, non-disconnect causes, self-assumed not retained, the array-order pin) and pass by construction.

### Dismissed, with reasons

- **"One global mutex on the interest hot path, taken under a shard guard."** Kept. Lock order is consistent everywhere (`interested_peers` shard → `retained_summaries` → `missing_summary_history`, never inverted — all four reviewers checked this independently and found no inversion), and `remove_peer_interest_for` already takes `missing_summary_history.lock()` on every removal, so this is a proportionate second acquisition on the same path, not a new class of contention.
- **"No sweep; expired entries hold slots until LRU pressure evicts them."** Accepted as designed and now documented. The cache is bounded on both entries and bytes, so dead weight costs a slot, never unbounded memory, and a sweep would add a timer for no correctness gain.
- **"Suspend/resume defeats the TTL."** `Instant` is `CLOCK_MONOTONIC`, so a suspended laptop's retained entries look fresh on resume. This is the same exposure a live entry already has (`is_expired_at` uses the same clock), so it is not a regression — noted, not fixed here.
- **"The cap-rejection path (`MAX_INTERESTED_PEERS_PER_CONTRACT`) returns before claiming the retained entry, and is unpinned."** Correct and deliberate: a rejected registration made no state change, so claiming the entry would discard a belief for a pair that was never recreated. Left unpinned — the branch returns before any retention code, so there is nothing behavioural to guard.
- **"Use `#[tokio::test(start_paused = true)]` for the `priority_select` flake instead of a larger hang guard."** Agreed as the better fix, and out of scope here: that file has ~20 more wall-clock timeouts of the same shape, several of which deliberately assert "nothing is ready yet" and want different treatment. Filed as #5110 with the per-site analysis; this PR changes only the one test that actually failed CI-adjacent runs.
- **"Poisoning: a hostile peer controls the retained bytes."** Confirmed safe by the stale-summary reviewer at code level, not taken on faith: the key is `(ContractKey, PeerKey)` and the restored value is only ever installed into that same peer's entry, so garbage bytes can only produce a bad delta for their own author. The one cross-peer lever is LRU eviction of legitimate entries, which fails safe (an evicted pair reverts to one full-state send, never a wrong delta). Retention now also requires the peer to have asserted the summary while connected — which it could already feed to `get_state_delta` directly.

### Also verified clean by more than one reviewer

Byte accounting in `store`/`trim`/`take` (no drift, `saturating_*` throughout, `LruCache::unbounded` never evicts silently); the 8 MiB + entry-cap memory bound enforced on every insertion path; cross-peer and cross-contract key leakage (types differ, so a transpose will not compile); the grace-sweep vs reconnect TOCTOU (`execute_pending_removals` re-removes from `pending_removals` before acting); eviction CPU (O(1) pops, amortized bounded); and that the `priority_select` hang-guard reasoning is accurate (iteration 0 is the only one that genuinely awaits a wake; the rest read buffered values).

Gates on the current HEAD: `cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`, `cargo test -p freenet --lib` (4588 passed, 0 failed), `cargo check -p freenet --all-targets`, `cargo check -p freenet --features simulation_tests --lib`.

[AI-assisted - Claude]

sanity:
## Split out the unrelated commit (scope rule-review)

The `rule-review/warnings` check was correctly failing: this PR bundled a `test(priority-select)` commit into a `fix(ring)` change. I did **not** use the `review-override` label — the exemption for "a tiny follow-up needed to make the PR mergeable" does not apply, because that test flaked under load average ~70 on the build host (four concurrent release builds), not on CI. This PR's 12 checks were green without it.

While splitting it, the commit turned out to contain **two** unrelated changes under a message that described only one:

- 12 lines in `priority_select/tests.rs` — the hang-guard fix the message is about. Now **#5114**, standing alone.
- 30 lines in `ring/interest.rs` — tightening the retention over-cap test from `Evicted >= 64` to an exact count. Genuinely part of this change, and unmentioned in the commit message. Kept here, as its own commit with an honest description.

So the split is on file boundaries rather than commit boundaries, and the branch is now three `ring`-scoped commits.

### The reviewed code is unchanged

`git diff` between the reviewed head (`a717adfb`) and the new head (`13340c3f`) touches **only** `priority_select/tests.rs` — 1 insertion, 11 deletions, i.e. exactly the reversion of the extracted hunk. Every line of ring/interest/telemetry code the four reviewers read is byte-identical, so the review stands on its own terms rather than by assertion. Per the per-code-content rule this is a re-verify-CI, not a re-review: CI must go green on `13340c3f` before this merges, and it will not be merged on the strength of the old SHA's green.

Merge order also changed: **#5108 goes first** (it was green and reviewed while this was blocked), so this will rebase onto it. Its `interest.rs` footprint is a small `has_peer_summary` helper, so the rebase should be clean — but if resolving it takes any judgment, the review gets re-run rather than assumed.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #5109 — **re-review, round 2**

### Summary
- **PR Title:** fix(ring): retain peer summaries across disconnect-grace interest removal
- **Type:** fix
- **CI Status:** all 14 checks green on `a717adfba`
- **Linked Issues:** context #5049, #4961, #4952; spun-off #5110
- **Review tier:** Full — state authorization + protocol behaviour, 1,536-line diff
- **Reviewers run:** code-first, testing, skeptical, big-picture — 4 independent, blind, against an isolated worktree at `a717adfba`. External non-Claude pass **not run**: codex is routinely quota-exhausted and the Gemini CLI has been unavailable since 2026-06-24, so the documented diverse-Claude-lens fallback was used. This is round 2; round 1 ran on `1e55df908` and its findings are in the comment above.

### Verdict

**State: Needs Discussion — recommend NOT landing the retention mechanism.**
**HEAD SHA reviewed:** `a717adfba`

Round 1 found a blocking scope defect: the restore was unscoped, so a belief retained at a `DisconnectGrace` could be resurrected after a withdrawal. I fixed it and asserted the invariant was now exact. **Round 2 found the same defect class again, via two new routes, plus a falsified TTL claim.** That is the second time I have asserted an exactness that review disproved, on a state-sync path. The pattern says the invariant is hard to hold, not that I keep making careless slips — so I am stopping rather than patching a third time.

---

### Must Fix (blocking) — verified against the code

**M1 — the discard is gated on `if removed`, so any withdrawal arriving while the pair is untracked is a no-op.**
`crates/core/src/ring/interest.rs:2056`. The store/discard decision sits inside `if removed { … }`. A retained belief exists *only* while the pair is untracked — that is the state a disconnect teardown leaves it in — so every withdrawal signal in that window silently fails to invalidate it. Found independently by the code-first and skeptical reviewers; I verified both routes:

- **`Unsubscribe`.** `crates/core/src/operations/subscribe.rs:946` calls `remove_peer_interest_for(&key, peer, Unsubscribe)` **unconditionally**, and it returns `false` for an untracked pair. Sequence: peer reports `S1` → transport drops → grace teardown retains `S1` → peer reconnects and unsubscribes (its client dropped the subscription while offline) → **no discard** → peer re-fetches the contract from elsewhere → re-registers within the TTL → `S1` restored, provenance re-stamped `PeerReported`. A belief survives an explicit unsubscribe — precisely what the round-1 fix was supposed to make impossible.
- **`Interests` full-replace omission.** `crates/core/src/node.rs:2963-2975` iterates `get_contracts_for_peer(pk)`. After the teardown the pair is not in `peer_contracts`, so a peer that reconnects and *omits* the contract from its full-replace set produces no removal at all. This half is **not** reachable by moving the discard out of `if removed` — the loop structurally cannot see it without a peer→retained reverse index.

The rustdoc at `interest.rs:1786-1800` is literally true ("a resident entry always comes from the pair's most recent *teardown*") but is offered as evidence for a different claim ("nothing the peer said invalidated what we believed"), which is false. Under this repo's own standard that gap is a review failure, not a nit.

**M2 — `RETAINED_SUMMARY_TTL` does not bound belief age, and one of my own tests blesses the behaviour that breaks it.**
`interest.rs:172-188` (the claim), `:708-724` (`store` stamps `retained_at: now` on *every* store), `:1984-1989` (the restore re-stamps provenance to `PeerReported`, making the belief eligible to be stored again). The documented formula is "live age + grace + **at most** one TTL"; the code permits N of each. A peer that disconnects, reconnects, registers (restoring the belief), and disconnects again before its `Summaries` reply carries a belief from hour zero indefinitely — at a cost of 90s per cycle — while `RetainedSummaryOutcome::Expired` never fires, so telemetry reports the bound as honoured. `a_restored_summary_survives_a_second_disconnect` (`interest.rs:6126`) pins exactly this cycle as desired behaviour. The PR body repeats the claim verbatim.

---

### Should Fix — the benefit is smaller than the PR argues

**S1 — the headline is ~20x above what the change can reach.** (big-picture) The mechanism only helps a recreation that is *simultaneously* `DisconnectGrace`-caused, held a `PeerReported` summary at teardown, recreated via registration rather than `upsert_peer_summary_from`, and broadcast to before the `Summaries` reply lands. The only cause-attributed measurement of that population is `TrackedFirstRecreated` = 25,704 sends / **7.72 GB** — and that row spans *all* removal causes, so it is an upper bound. The bridge to the 77 GB `TrackedFirstNew` row is a population-level ratio ("most creations must be recreations"), which does not establish that those particular full-state sends were disconnect recreations holding a retained peer-asserted belief. The body does label the counters a floor, so this is calibration rather than misrepresentation — but a 7.72 GB upper bound against two rounds of scope defects is a poor trade.

**S2 — mutation testing shows several new guards are not actually pinned.** (testing, with a validated harness: positive controls produced the expected failures, so a green result is a real hole)
- `let restore_wanted = is_new && summary.is_none()` → dropping `is_new` leaves **all 4,588 lib tests green**. `retained_summary_is_not_applied_to_an_existing_entry` (`interest.rs:6459`) does not pin it — by the time its final registration runs the cache is already empty. The premise is in fact unreachable: a resident retained entry and a live `PeerInterest` for the same pair cannot coexist, because the store happens inside `remove_peer_interest_for` and both recreate paths drain the cache under the same shard guard.
- Deleting `discard_retained_summary` from the removal path (`interest.rs:2095`) → all tests green. The two tests whose docstrings claim to cover it do not.
- Deleting the provenance assignment in `upsert_peer_summary_from` (`interest.rs:2268`) → all tests green.
- `concurrent_store_take_and_discard_keep_the_bookkeeping_consistent` (`interest.rs:6760`) cannot contend: all four workers use one contract, and every cache mutation is under that contract's DashMap shard write guard, so they fully serialize. A serial stress test wearing a concurrency label.

My "17 of 21 fail without the fix" claim holds only in the coarse sense that reverting the feature breaks those tests; it does not survive per-decision mutation. Correcting it here because the claim is in the PR body.

**S3 — the converged-skip residual is real for `PeerReported` beliefs too.** A restored belief that byte-matches our summary makes `broadcast_to_single_peer` send **nothing** (`broadcast_queue.rs:1181-1244`) and refresh the TTL, so there is no delta to reject and no `ResyncRequest` — recovery falls entirely to the next `Summaries` round. Combined with M2, a flapping peer never reaches it. My round-1 mitigation argued only `SelfAssumed` beliefs can be wrong-and-matching; that is too strong.

*Partial dismissal, with reason:* the skeptical reviewer offered `DigestAgreement` as a counterexample — a `PeerReported` belief byte-identical to ours *by construction*. I checked `node.rs:3550-3585`: the digest match **proves** the peer holds those bytes (BLAKE3-128 over the summary), and the code says so ("Fact, not belief"). The `PeerReported` classification is correct there, and the scenario still requires the peer to lose state afterwards — the already-acknowledged "moved back" case, not a hole in the provenance logic. The residual stands; that particular framing does not.

---

### Consider

- `SkippedOversized` (`interest.rs:2086-2094`) fires without testing provenance, inflating the only tuning signal for `MAX_RETAINED_SUMMARY_BYTES` with entries that were ineligible anyway. (3 reviewers)
- `Evicted` conflates "cache too small" with "expired dead weight" — the TTL is enforced only on `take`, so an aged-out entry leaves via `trim` counted `Evicted`, which is advertised as *the* resize signal.
- No per-peer fairness bound: one peer can occupy all 16,384 slots with a single `Summaries` message after connecting, because `on_ring_connection_established` volunteers every contract hash we index and the upsert creates entries. Fails safe and memory stays bounded, but it denies the feature to honest peers and pins `Evicted` high.
- Both at-cap early returns (`interest.rs:1946`, `:2248`) skip the claim/discard, so a fresher peer-reported summary can be dropped while the older retained belief survives.
- `interest.rs:1750-1754` derives "~95% of the population" from a fleet-wide write mix, but provenance is last-writer-wins and `record_delivery_to_interest` writes `SelfAssumed` on every delivered broadcast — for exactly the large-state pairs this targets, the last write before a transport loss is likely `Delivery`. The retained *fraction* cannot be predicted from that ratio.
- `PeerInterest::new`'s comment names one overriding caller; there are two (`interest.rs:888-891` vs `:2267`). Stale `.claude/rules/flaky-tests.md` citation at `priority_select/tests.rs:438` (that is a global rule, not in-repo). PR body says 19 new tests; there are 22.

---

### What held up under scrutiny

Worth recording, because these were the parts under most suspicion and they are sound:

- **The removal chokepoint claim is true.** `sweep_expired_interests`, `remove_evicted_in_use` and `remove_all_peer_interests_for` all delegate to `remove_peer_interest_for`; nothing else mutates `interested_peers` destructively. (code-first and big-picture, independently)
- **Lock ordering is correct** — always `interested_peers` shard → `retained_summaries` → `missing_summary_history`, never inverted; no deadlock, and `total_bytes` cannot drift because every mutation is a single acquisition. (3 reviewers)
- **Bounds hold on both axes**; `trim` terminates and never evicts the just-inserted entry; the counter identity `Stored − departures == residency` is correct across replace+evict.
- **No cross-peer or cross-contract leakage** — the key types differ, so a transpose will not compile.
- **`SummaryProvenance::of` classifies every source correctly** against how each is actually produced at its call sites.
- **The `recent_removal` classifier fix is correct and independently valuable.** Both reviewers who checked it confirmed it cannot resurrect an arbitrarily old removal (every reader still filters on `removed_at <= INTEREST_TTL`), it changes no routing, its test is non-vacuous, and the wipe it removes was introduced two commits ago (`bf4568c40`, #5091) with no rationale and no pin — so nothing is contradicted.

---

### Recommendation — split three ways

1. **Land the `recent_removal` classifier fix on its own.** ~15 lines, verified correct by two reviewers, non-vacuous test, no routing change. It is the better explanation for `UntrackedFirstObserved` being far larger than #4952 should allow — currently attributed on #5049 to the seeding/heartbeat chain.
2. **Move the `priority_select` timeout fix to its own PR** under #5110. `.claude/rules/git-workflow.md:51` treats unrelated cleanup riding along as grounds for closing, and #5110 already exists to carry it.
3. **Convert the retention mechanism to an issue** with the round-1 and round-2 findings, the 7.72 GB upper bound, and the two open design questions: how to invalidate a retained belief on a withdrawal that arrives while the pair is untracked (needs a peer→retained index, or moving the decision from removal-cause to a recorded withdrawal signal), and how to bound belief age across reconnect cycles (carry the original assertion timestamp rather than re-stamping on store).

The mechanism may well be worth building — the churn evidence is solid and the direction is right. It should not land at the end of a session in which two consecutive review rounds each found the invariant leaking in a new place.

[AI-assisted - Claude]

sanity:
Withdrawn pending redesign — see #5115, which carries the churn evidence, the honest 7.72 GB bound on the reachable benefit, the two open design problems, and the parts of this implementation that were reviewed as sound and are reusable.

Keeping this PR as a draft rather than closing it so the diff stays available as the reference implementation for a second attempt.

[AI-assisted - Claude]

sanity:
**Status marker: deliberately parked — not stale, not awaiting review.**

Recording this so backlog sweeps stop re-litigating it. This PR was withdrawn by its author on 2026-08-03 pending redesign, with an explicit decision to keep it open as a draft so the diff stays available as the reference implementation for a second attempt. That decision stands and is being respected.

Where the work actually lives now:

- **#5115** (open) is the replacement record. It carries the churn evidence, the honest 7.72 GB bound on the reachable benefit (not the 78 GB the original argued from), and the two design problems that blocked the first attempt: invalidation while the pair is untracked, and belief-age vs slot-age TTL.
- **#5117** (merged 2026-08-03, `cd74b1372`) landed the secondary half. `ring/interest.rs` now carries the "Reset the SEND counter only, `recent_removal` must NOT be wiped here" guard plus the `untracked_staleness_reset_preserves_a_recent_removal` test.

The primary retention mechanism is still absent from main — no `RetainedSummaryCache`, no `SummaryProvenance`, no `take_retained_summary` anywhere in `crates/core/src` — so this is genuinely unfinished work, not a superseded one. It is parked on a design question, and the branch is now well behind main and conflicting, so a second attempt should expect to re-derive the diff rather than rebase it.

Verified against `origin/main` @ `9196e4be6`.

[AI-assisted - Claude]

sanity:
**Status: deliberately parked, not stalled.** Recording this so backlog sweeps stop re-triaging it.

The author withdrew this on 2026-08-03 in favour of issue #5115, and the secondary half landed separately as merged #5117. It is being kept open as a draft — the author's stated reason: *"Keeping this PR as a draft rather than closing it so the diff stays available."*

Leaving that decision intact. Noting for anyone who reaches this from a staleness report: this is not an unowned PR that needs reviving, and the live thread is #5115.

(Found during a sweep of 59 open PRs on 2026-08-24. Triage confirmed the withdrawal and the #5117 split independently before writing this.)

[AI-assisted - Claude]

PR #5112, opened 2026-08-03
feat(telemetry): measure the fan-out multiplier via originated-vs-relayed applies (#5062)

### LINKED ISSUE #5062: feat(telemetry): measure the fan-out multiplier (sends per originated update, per contract)
## Summary

Measure the **fan-out multiplier** — sends per *originated* update, per contract — so cross-hop broadcast amplification becomes an observable number rather than the leading untested hypothesis.

## Why this matters

After hash-first (#4965) and the one pathological contract (#5056) are removed, roughly **400 MB/day/node of genuine delta traffic** remains against a zero-based budget of **~8 MB/day** for actual chat messages. That is about **50x unexplained**.

The leading hypothesis is fan-out redundancy. Measured room fan-out is **17.1 co-hosts (p90 38)**, so if each hop re-broadcasts to its own co-hosts without effective dedup, that alone is a ~17x multiplier and would account for most of the gap. If it holds, the next 10x is a **propagation-topology** problem, not a payload problem — a completely different remedy from everything currently queued.

Nothing measures it today.

## Why `broadcasted_to` is not enough

`update_broadcast_emitted` carries `broadcasted_to` (`tracing/telemetry.rs:1221`, `:1695`), which is the number of targets for **one** emission — i.e. a single node's fan-out *degree*. The multiplier that matters is total sends divided by updates **originated**, which captures re-broadcast amplification **across hops**.

Total sends alone cannot separate the two cases that matter:

- one node fans out to 17, and
- 17 nodes each fan out to 17.

Those differ by 17x in cost and have different fixes. Only an originated-vs-relayed split distinguishes them.

## What is needed

Fields on the existing 60s `outbound_message_mix` / `broadcast_payload_mix` rollups (no new event stream, fixed cardinality, bounded per-contract maps as those files already do at `MAX_TRACKED_CONTRACTS = 256`):

1. per-contract **sends**
2. per-contract count of updates **originated locally** vs **relayed / re-broadcast**

so `sends / originated` is computable fleet-wide.

## START HERE: check the transaction-identity shortcut first

**Do not begin with the trait surgery below.** Check this first, because it may make all of it unnecessary:

`broadcast_queue.rs:1025` mints a **fresh** `Transaction` per broadcast invocation:

```rust
let update_tx = crate::message::Transaction::new::<crate::operations::update::UpdateMsg>();
```

- If a relayed re-broadcast carries the **inbound** transaction id anywhere reachable at that point, then counting distinct originated-vs-seen tx ids per contract gives the multiplier directly, and none of the plumbing below is needed.
- If both paths mint fresh, transaction identity cannot distinguish them and the plumbing is unavoidable.

Either way, nobody should discover which by doing the surgery first.

## The plumbing, if the shortcut does not work

Traced against `fcd943cb0`. The cheap half and the expensive half are very different:

**Cheap half — per-contract sends.** `broadcast_payload_mix::record_delivered` already fires once per target peer with the contract id, from `broadcast_queue.rs:1178` (streaming) and `broadcast_queue.rs:1228` (inline). A bounded per-contract counter there is roughly 20 lines.

**Expensive half — originated vs relayed. The provenance does not exist where the broadcast is emitted.**

- `NodeEvent::BroadcastStateChange { key, new_state, is_retry, is_reemit }` has no origin field. `is_retry` / `is_reemit` are about *this node's* re-emission bookkeeping (`p2p_protoc/broadcast.rs:179`), not about who originated the update.
- The production emit sites are inside the executor — `contract/executor/runtime/executor_impl.rs:1991` and `:2411` — and fire after a **local apply**. A local apply happens identically for a client-originated UPDATE and for an inbound relayed `UpdateMsg::BroadcastTo`. That is exactly the ambiguity to resolve, and it is resolved nowhere on this path.
- The apply entry point is the `ContractExecutor` trait's `upsert_contract_state` (`contract/executor.rs:859`) and `upsert_contract_state_deferrable` (`contract/executor.rs:880`), whose signature is `(key, update, related_contracts, code)` — no provenance parameter.

So the change is:

1. new provenance parameter on both trait methods
2. every impl: `Executor<Runtime>`, `RuntimePool`, `MockRuntime`, `MockWasmRuntime`
3. ~20 call sites including tests
4. new field on `NodeEvent::BroadcastStateChange` (4 production emit sites)
5. threading from both UPDATE entry points (client request vs inbound relay)
6. then the cheap rollup part

## Risk and ordering

- Steps 1-5 touch the **UPDATE operation state machine** and the **contract executor**. Both are always-Full-tier review surfaces per `.claude/rules/pr-quality.md`, and both have prior production incidents.
- `broadcast_payload_mix.rs` is also touched by open PR #5055 (`top_contracts_by_total_bytes`). This should land **after** #5055 or expect a conflict.
- Deliberately kept out of #5052 / PR #5061 (the `interest_sync_summaries` emitter split), which gates hash-first attribution and must ship this release. This is next-release work.

Related: #5052, #5056, #4965, #5055, #5003.

[AI-assisted - Claude]

sanity:
Closes #5062.

## What the shortcut check established

#5062 says to check the transaction-identity shortcut before doing anything else. **It does not work**, definitively:

- The inbound `Transaction` is dropped at `update/op_ctx_task.rs:1796` (relay `BroadcastTo`) and `:293` / `:411` (client), because `update_contract` (`update.rs:682`) has no `Transaction` parameter.
- From there down, the client-originated and relay-originated paths are byte-identical: same `UpdateQuery` variant, same fair queue, same `maybe_defer_upsert`, same executor call, same emit.
- `NodeEvent::BroadcastStateChange` carries no tx and no origin; `BroadcastEntry` carries none; `broadcast_queue.rs:1432` mints a fresh one. This is already documented in-tree at `p2p_protoc/broadcast.rs:522-530`.

So transaction identity cannot distinguish them.

**But the issue's ~160-call-site trait surgery is not needed either.** `update_contract` already receives the caller's provenance intent and already learns `state_changed` from the handler's reply — and `bridged_upsert_contract_state_inner` calls `commit_state_update` (which emits the fan-out) *only* inside the `updated_state != current_state` branch. So a changed apply is what causes a broadcast, and it can be attributed at that one function. The whole measurement is one parameter and two counter calls, with no `ContractExecutor` trait change.

## What this ships

Four fields on the existing 60s `broadcast_payload_mix` rollup:

```
applies_client_local_changed / applies_client_local_total
applies_network_relay_changed / applies_network_relay_total
```

Fixed cardinality (2 origins x 2 outcomes), no per-contract map, no peer IDs, recorded into the same window and drained by the same atomic take as the existing `sends` counters.

**The number to read** is the ratio of sums over the fleet:

```
sum(applies_network_relay_changed) / sum(applies_client_local_changed)
```

— distinct peer-side state transitions per originated update. Against the measured ~17.1 co-host degree: ≈17 means the update reaches its co-host set once and stops (payload problem); ≫17 means state keeps genuinely moving hop after hop (propagation-topology problem). That is the #5062 question.

Note the discriminant is **not** "did a re-broadcast happen" — one happens in both topologies, since the executor emits on every changed apply. It is whether the second hop's recipients *also* change state. Review caught this; the first draft of the docs had it wrong.

No derived ratio is published. Most nodes serve no local clients, so a per-node per-window ratio is `0/0` far more often than not, and publishing one would invite exactly the mean-of-ratios aggregation the module's existing `not_efficient_summary_to_state_bytes_ratio` is named to steer people away from. The division belongs in the query.

### R/C is biased UPWARD — read it as an upper bound

An earlier revision of this PR called R/C "sound" because both halves come from the same function. Review showed that is wrong, and wrong in the direction that costs money: same function is not the same population, and the residuals do not cancel — they all push R/C toward the ≫17 reading whose remedy is the expensive one.

Numerator gets relayed `changed` applies with no origination behind them: anti-entropy heals (`SyncStateToPeer`) share the broadcast queue and arrive as the same `UpdateMsg::BroadcastTo` wire shape, so the receiver cannot tell them from a fan-out leg — and heal volume rises with churn, i.e. it is loudest under the conditions being measured. Denominator misses originations entirely: a client PUT never calls `update_contract`, yet its fan-out produces relayed changed applies fleet-wide; same for delegate-driven work and for ResyncResponse heals.

Pulling the other way, and smaller: the broadcast dedup cache and the merge-failure backoff gate short-circuit before `update_contract`, and fair-queue shedding drops sub-`ClientLocal` work first.

**Practical reading: R/C ≈ 17 is strong evidence of one-hop reach**, since the biases can only push it up. **R/C ≫ 17 is suggestive, not conclusive** — bound the heal and PUT-origination contribution before spending on propagation topology. This is all stated in the module docs, not just here.

## Correction to the issue text

#5062 says "nothing measures it today" — written before #5091 shipped. The `network_efficiency_v1` receiver-side counters now measure the **effect** (fleet-wide: 11.4M byte-identical delta duplicates, 5.2M full-state duplicates, 17.5 payloads delivered per one that changes anything, 97.2% of received contract-payload bytes changing nothing). What was still missing, and is what this PR adds, is the **attribution** — and that is the difference between a payload fix and a propagation-topology fix. Nobody should re-derive this.

## Honest limits, stated in the module docs

- `total_sends / applies_client_local_changed` is the more direct statement of "sends per originated update", and the one #5062 asks for, but it is **looser still** than R/C and is demoted to a coarse cross-check. Its numerator additionally carries every heal delivery, every PUT/GET-cache/Resync/first-install/delegate-driven send, and no-target retries plus `PendingBroadcastStore` re-emits of already-counted applies. Against that, `record_delivered` is delivery-gated while `record_apply` is not, and the executor's emit is best-effort under channel pressure.
- Telemetry coverage must be uniform across relay-heavy and client-serving nodes for either ratio to be unbiased. Gateways relay much and serve few local clients, so a reporting population skewed toward them under-samples the denominator — same upward direction as everything else.
- Residuals where `changed` and "a broadcast happened" diverge are enumerated in `ApplyOrigin`'s docs rather than claimed absent. The load-correlated one (`try_notify_node_event` dropping silently when the node-event channel is full) matters most, since it biases downward exactly when the signal is strongest.
- These counters will never reconcile against the existing `ReceiverApplyStats`: those are cumulative, not windowed, and fed by only the two receive drivers that build a `ReceiverTerminalGuard`. Documented so nobody wastes an afternoon diffing them.

## Provenance is explicit, not derived from `Priority`

`contract::Priority` is defined on nearly this axis and deriving from it would have been free. Review found a live counter-example: `put/op_ctx_task.rs`'s summary-first reverse-delta merge tags `Priority::ClientLocal` while applying content the **remote holder** sent back — its own comment says the tag is chosen to match the originator-loopback store's scheduling lane. Deriving there would have booked a relayed apply into the originated bucket, deflating both ratios in the flattering direction.

So origin is an explicit argument the compiler demands at all 8 production call sites (`ClientLocal` x2, `NetworkRelay` x6 including the corrected one), and a source-scrape pin holds those counts — flipping one tag was, per review, *the* production mutation for this feature and previously passed the entire suite. Scheduling class and data provenance are free to diverge, and a future re-tag for queue-fairness reasons must not silently move a measurement's denominator.

## Telemetry volume

Measured on the actual field names, not estimated:

| | bytes/rollup | KB/peer/day | MB/day @ 1,100 peers | share of ~30 GB/day |
|---|---|---|---|---|
| all-zero window | 130 | 183 | 196 | 0.64% |
| busy node | 134 | 188 | 202 | 0.66% |
| 6-digit worst case | 150 | 211 | 227 | 0.74% |

## Deliberately left undone

**Per-contract origin split.** #5062 asks for it; it is not here. A top-N per-contract array roughly doubles this event's contribution, to refine a number nobody has read yet. Per-contract *sends* already ship (`top_contracts_by_total_bytes`, #4979/#5055), so the moment the node-level ratio says re-broadcast amplification is real, the per-contract denominator becomes a small follow-up against a known-useful measurement instead of a speculative one. If the ratio comes back ≈17, it is not needed at all.

**Separating heal sends from fan-out sends** in `record_delivered`, and **a PUT-path apply counter**. Together these are what would turn R/C from an upper bound into a tight number. The first needs a flag on `BroadcastEntry` and `broadcast_to_single_peer`; the second needs instrumenting `put_contract`'s `PutQuery` path. Both are more than a telemetry-only change should carry the night before a release, and neither is needed if R/C comes back ≈17.

## Scope and risk

Telemetry only. No propagation behaviour, fan-out policy, or wire-format change. The one signature change is `update_contract`'s new parameter, which the compiler enforces at every call site.

## Validation

`cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`, `cargo check -p freenet --all-targets`, `cargo check -p freenet --features simulation_tests --lib`, and `cargo test -p freenet --lib` (4,576 passed, 0 failed) all green.

Four tests are mutation-verified rather than asserted to work:

| test | mutation | result |
|---|---|---|
| `update_contract_records_the_apply_against_the_origin_it_was_given` | record commented out | fails (`left: 0, right: 1`) |
| same | origin hardcoded | fails |
| `update_contract_call_sites_tag_their_origin_pin` | a driver's `ClientLocal` flipped to `NetworkRelay` | fails (`(1,5)` vs `(2,4)`) |
| `record_apply_indexes_the_origin_arm_fallibly_pin` | `get_mut` replaced with `[idx]` | fails |
| `concurrent_records_racing_a_rollup_conserve_bytes` | always write the unchanged slot | fails (`left: 0, right: 668`) |

Two earlier tests were replaced because review showed they did **not** test what their names said — one checked `ALL` against itself, the other asserted a compile-time truth about fixed-size arrays. See the review comment.

Refs #5062, #5091, #4956, #5055.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -84,6 +84,11 @@ pub(crate) use op_state_manager::OpManager;
 
 mod network_bridge;
 
+/// Where an applied contract update came from (#5062). Re-exported because
+/// `operations::update` has to name it to declare its provenance, and
+/// `network_bridge` is private outside this module.
+pub(crate) use network_bridge::broadcast_payload_mix::ApplyOrigin;
+
 // Re-export fault injection types for test infrastructure.
 // No cfg gate: underlying items are unconditionally compiled and integration
 // tests compile the lib without cfg(test).
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -83,13 +83,55 @@
 //! that separates them is
 //!
 //! ```text
-//! sum(total_sends) / sum(applies_client_local_changed)
+//! sum(applies_network_relay_changed) / sum(applies_client_local_changed)
 //! ```
 //!
-//! summed over the fleet as a RATIO OF SUMS, never a mean of per-window
-//! ratios — most nodes serve no local clients, so a per-node ratio is `0/0`
-//! most windows. `applies_network_relay_changed / applies_client_local_changed`
-//! reads the re-broadcast amplification directly.
+//! — how many distinct peer-side state transitions one originated update
+//! causes. Against the measured ~17.1 co-host degree, ≈17 says the update
+//! reaches its co-host set once and stops; ≫17 says state keeps genuinely
+//! moving hop after hop. See [`ApplyOrigin`] for why "did a re-broadcast
+//! happen" is NOT the discriminant (one happens either way) and whether its
+//! recipients then change state is.
+//!
+//! Always a RATIO OF SUMS over the fleet, never a mean of per-window ratios:
+//! most nodes serve no local clients, so a per-node per-window ratio is `0/0`
+//! far more often than not. Note also that the sum is unbiased only if
+//! telemetry coverage is uniform across relay-heavy and client-serving nodes;
+//! gateways relay much and serve few local clients, so a reporting population
+//! skewed toward them under-samples the denominator.
+//!
+//! ### `total_sends / applies_client_local_changed` is an UPPER BOUND only
+//!
+//! Tempting, and the more direct statement of "sends per originated update",
+//! but the two are not drawn from the same population. `total_sends` counts
+//! every delivery `record_delivered` sees, and the broadcast queue carries
+//! traffic that no `applies_*` counter can ever match:
+//!
+//!   * targeted anti-entropy heals (`NodeEvent::SyncStateToPeer`) enqueue into
+//!     the SAME queue, with no apply behind them at all — and heal volume
+//!     rises with churn, i.e. loudest under the conditions being measured;
+//!   * PUT-, GET-cache-, ResyncResponse-, first-install- and delegate-driven
+//!     applies broadcast without passing through `update_contract`, so they
+//!     add sends with no denominator entry;
+//!   * no-target retries and `PendingBroadcastStore` re-emits re-send an apply
+//!     that was counted once, or never.
+//!
+//! All of those inflate it. Pulling the other way, `record_delivered` is
+//! delivery-gated while `record_apply` is not, and the executor's broadcast
+//! emit is best-effort under channel pressure. So the net direction is not
+//! known, which is why the relay/client ratio above — same function, same drop
+//! semantics on both halves — is the one to read for a spend/don't-spend
+//! decision. Separating heal sends from fan-out sends would make this bound
+//! tight and is the natural follow-up.
+//!
+//! ### Not the same thing as the receiver-apply counters
+//!
+//! [`ReceiverApplyStats`] above also counts changed-vs-no-op applies, and the
+//! two will never reconcile — by construction, not by bug. Those are
+//! cumulative rather than windowed, and are fed only by the two broadcast-
+//! receive drivers that build a [`ReceiverTerminalGuard`], whereas
+//! `applies_network_relay_*` covers every relayed `update_contract` call. The
+//! relay arm is therefore a strict superset. Do not diff them.
 //!
 //! ### What is deliberately NOT here
 //!
@@ -106,11 +148,20 @@
 //!
 //! One short uncontended mutex acquire per **delivered broadcast** (not per
 //! packet), covering a handful of integer adds and at most one bounded
-//! `HashMap` touch, plus one more per completed update APPLY (strictly rarer
-//! than sends, since one apply is what fans out to N peers). Everything else
-//! happens in the aggregator task, and the hot path only ever WRITES. The lock
-//! is what makes a rollup a consistent snapshot; see [`PayloadMix`] for why
-//! per-field atomics were not enough.
+//! `HashMap` touch, plus one more per completed update APPLY.
+//!
+//! That second acquisition is NOT rare: every RECEIVED broadcast runs
+//! `update_contract` too, and by the #5091 figure above only ~1 in 17.5 of
+//! those changes state and fans out. So on a relay-heavy node the apply and
+//! send rates are comparable and this roughly DOUBLES the acquisition rate on
+//! `window` — worth stating plainly, since "one more, strictly rarer" would
+//! license the next apply-path counter on reasoning that is off by ~17x. It is
+//! still cheap in absolute terms: the apply critical section is two array
+//! writes, against up to three bounded `HashMap` touches for a send.
+//!
+//! Everything else happens in the aggregator task, and the hot path only ever
+//! WRITES. The lock is what makes a rollup a consistent snapshot; see
+//! [`PayloadMix`] for why per-field atomics were not enough.
 //!
 //! [`broadcast_to_single_peer`]: super::broadcast_queue::broadcast_to_single_peer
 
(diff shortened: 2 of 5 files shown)


sanity:
## Consolidated review

Two rounds, six independent blind sub-agent reviewers, none of which wrote the code. Full tier, because the change touches the UPDATE apply path.

**Round 1** (lenses: code-first, testing, skeptical-adversarial, telemetry-schema/volume/cardinality) ran against `48a64f6`.
**Round 2** (lenses: skeptical-adversarial, testing-rigour + documentation-truthfulness) ran against `970b186`, after the round-1 fixes, and was told to verify the fixes rather than re-report the findings.

Every finding is listed below with its disposition. Nothing was dismissed as "won't fix".

### BLOCKER / MAJOR — fixed

| # | Finding | Fix |
|---|---|---|
| 1 | Provenance derived from `contract::Priority`, a **scheduling** class. Live counter-example: `put/op_ctx_task.rs`'s summary-first reverse-delta merge tags `Priority::ClientLocal` while applying content the *remote holder* sent back — its own comment says the tag matches the originator-loopback store's lane. This booked a relayed apply into the originated bucket and falsified the "can only ever look worse, never better" claim the docs leaned on. | `ApplyOrigin` is now an explicit parameter on `update_contract`, compiler-enforced at all 8 sites; that site is tagged `NetworkRelay` with the rationale inline. `from_priority` deleted. |
| 2 | **The documented topology discriminant was wrong.** "Under the first topology essentially every broadcast follows a locally-originated apply" is false — the executor emits on *every* changed apply, so a re-broadcast happens in both topologies. | Rewritten. What differs is whether the second hop's *recipients* change state: under one-hop reach their applies are no-ops (counted in `_total`, not `_changed`). The discriminant logic was independently re-verified as correct in round 2. |
| 3 | **`relay_changed / client_changed` was called "sound"** on a same-function argument. Same function ≠ same population, and the residuals don't cancel — they all push the ratio **up**, toward the ≫17 reading whose remedy is the expensive one. Heals (`SyncStateToPeer`) share the broadcast queue and arrive as the same `UpdateMsg::BroadcastTo` shape, so receivers book them as relayed applies with no origination behind them; client PUTs and delegate work never call `update_contract`, so their originations are missing from the denominator while their fan-out feeds the numerator. | Claim retracted. Docs now enumerate the biases and state the practical reading: ≈17 is strong evidence of one-hop reach (biases can only push up); ≫17 is suggestive and wants the heal/PUT contribution bounded first. `record_apply` no longer claims the biases point "toward no problem". |
| 4 | `total_sends` is not a clean numerator — heals, PUT/GET-cache/Resync/first-install/delegate-driven sends, retries and re-emits. | Demoted from headline to a coarse cross-check, with the contaminants named. |
| 5 | The "1:1 correlate" claim was overstated: `try_notify_node_event` is best-effort and drops silently under channel pressure (**load-correlated** — biases down exactly when the signal is strongest); the first-install branch uses a second emitter; the replay loop can emit several per apply; a mid-init UPDATE returns `NoChange` and is replayed later. | All four residuals enumerated in `ApplyOrigin`'s docs rather than claimed absent. Round 2 independently verified each against the executor. |
| 6 | **Nothing pinned the origin tag at any call site.** Flipping `ClientLocal`→`NetworkRelay` in a driver passed the entire suite while collapsing the denominator. Both round-2 reviewers called this *the* production mutation for the feature. | New pin `update_contract_call_sites_tag_their_origin_pin`; mutation-verified (fails `(1,5)` vs `(2,4)`). |
| 7 | **The production wiring had no behavioural coverage**, and the pin guarding it passed when the calls were commented out — a reviewer demonstrated this empirically. | New `update_contract_records_the_apply_against_the_origin_it_was_given` drives the real `update_contract` against a real `OpManager`. Mutation-verified against a commented-out record (`left: 0, right: 1`) and a hardcoded origin. Priority is deliberately held constant while origin varies, so a regression to deriving from `Priority` fails it. |
| 8 | Two tests did not test what their names said. The `ALL`-exhaustiveness test derived its variant list *from* `ALL`, so it checked `ALL` against itself and stayed green in exactly the "added to `index()`, forgotten in `ALL`" case it claimed to catch. The out-of-range test asserted `get_mut(COUNT).is_none()` — a compile-time truth about any fixed-size array — and stayed green when the guard was replaced with `[idx]`. | First is now an exhaustive `match` in the test body, which fails to compile until a new variant is listed in both places. Second replaced with a source-scrape pin; mutation-verified. |
| 9 | Stale cross-references survived the round-1 fixes: `update.rs` still said "`priority` is a sound provenance source" and "1:1 correlate"; the emit site still named the demoted ratio as "the quantity that matters" — the comment a dashboard author reads. | Both rewritten. |

### MINOR — fixed

- Infallible `w.applies[idx]` inside the held mutex, against `record_delivered`'s own documented anti-panic precedent → fallible `get_mut`, plus the pin above.
- A doc block had landed on the wrong function, leaving `record_apply` undocumented and an orphaned sentence fragment on the wrapper → repaired.
- The pin joined on a single space instead of removing whitespace, so a rustfmt reflow would have reddened it on a pure reformat — the failure that teaches the next person to delete the pin. Now matches the sibling pin's actual convention (and the comment claiming that sibling was "in this same file" is corrected; it is in `operations/update.rs`).
- The concurrency test conserved only the *total*, so a mutation writing only the unchanged slot passed while zeroing the measurement → conserves `changed` separately; mutation-verified (`left: 0, right: 668`).
- Cost claim said applies are "strictly rarer than sends, since one apply fans out to N peers" — wrong, since every *received* broadcast also runs `update_contract`, so the rates are comparable and the lock rate roughly doubles. Corrected, because the old justification was off by ~17x and would license the next apply-path counter.
- Stale "3 x 2 `u64`s" (it is 2x2), "at most one `HashMap` touch" (up to three), "two array writes" (one), "`ReceiverApplyStats` above" (below).
- Added a note that these counters will never reconcile against `ReceiverApplyStats` — cumulative not windowed, and fed by only the two receive drivers that build a `ReceiverTerminalGuard`, so the relay arm is a strict superset. Documented so nobody wastes an afternoon diffing them.
- Dropped the unused `Background` arm (4 published fields, not 6), which also cut the volume.
- Documented the relay-side suppressions that sit *before* `update_contract` (broadcast dedup cache, merge-failure backoff — the latter concentrated on poison contracts) and the fair-queue shedding asymmetry.

### Accepted, not fixed — with reasons

- **Per-contract origin split** (asked for by #5062). Deferred on volume: a top-N array roughly doubles this event's contribution to refine a number nobody has read yet. Per-contract *sends* already ship. If the ratio comes back ≈17 it is not needed at all. Stated in the PR body and module docs.
- **Separating heal sends from fan-out sends** at `record_delivered`, which would make `total_sends / originated` tight rather than loose. Needs a flag on `BroadcastEntry` and `broadcast_to_single_peer` — more than a telemetry-only change should carry the night before a release. Named as the follow-up.
- **A PUT-path apply counter**, which would close the denominator asymmetry. Same reasoning; named as the follow-up.
- **The pin can still pass on commented-out code** (whitespace collapse doesn't strip comments). Not closed at the pin, because the behavioural test now covers both arms — which is the right instrument for that. Stated rather than left implied.
- **The no-change behavioural test uses the `UpdateNoChange` reply**, which in production is the *exceptional* shape (`UpsertResult::NoChange`-with-successful-fetch yields `UpdateResponse { state_changed: false }` instead). Both shapes reach the same `record_apply(origin, false)`; the more common one is covered by the changed-arm test's `state_changed` plumbing. Worth a follow-up, not a blocker.
- **The ordering assertion is textual, not executional** — the harness's stand-in always serves the fallback's `GetQuery`, so that exit is unreachable in test. Now said explicitly at the pin.
- **Long field names cost ~half the added volume.** Kept: at 0.66% of the collector, legibility of a metric whose entire failure mode is misinterpretation is worth more than 0.3%.
- **Rollup drops correlate with busy windows** and `broadcast_payload_mix` is not in `KnownShadowRollup::ALL`. Pre-existing property of the whole event, not introduced here; a drop loses numerator and denominator together.
- `#[cfg(test)] applies_snapshot` verified to compile under `--features simulation_tests`.

### Verified clean by reviewers

All 8 call-site tags checked individually against surrounding behaviour; no missed or addable-untagged call site (the parameter has no default). `counts[changed_idx]` cannot panic (`usize::from(bool)` against `[u64; 2]`). No lock held across an `.await`, no re-entrancy, no deadlock. Fixed cardinality genuinely fixed — no map, no contract keys, no peer IDs. Volume arithmetic independently recomputed and confirmed. Saturating arithmetic and the single-window/single-drain design confirmed correct. No existing test was weakened; the nine modified call sites are pure argument additions.

### CI

`cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`, `cargo check -p freenet --all-targets`, `cargo check -p freenet --features simulation_tests --lib`, `cargo test -p freenet --lib` (4,576 passed, 0 failed) — all green on `949c05b3a`.

[AI-assisted - Claude]

sanity:
## Post-merge verification of the sole-emitter claim vs #5108

This merged ~6 minutes after #5108 without a recorded re-verification against the merged tree, so I did it after the fact. Recording the result here rather than leaving the claim resting on a pre-merge check.

**The claim holds.** Production emitters of `NodeEvent::BroadcastStateChange` on merged `main` (`7f5a0d14b`) are exactly two, both in `contract/executor/runtime/executor_impl.rs` (`:1999`, `:2419`), plus `mock_runtime.rs:298` which is not production. Every other match in the tree is a comment, a source-scrape pin, or the consumer side in `p2p_protoc/broadcast.rs`.

**#5108 modified none of them.** Its diff covered `broadcast_queue.rs`, `p2p_protoc.rs`, `p2p_protoc/broadcast.rs`, `delta_incompat.rs`, `interest.rs`, `router.rs` and one test file — all on the scheduling/receiving side of the event. It changed how a fan-out is scheduled, not whether the event is emitted, so it adds no fifth residual to the four this PR documents on `ApplyOrigin` (best-effort `try_notify_node_event` drops, the first-install second emitter, the replay loop, and mid-init `NoChange`).

**One interaction worth naming for whoever reads these counters.** #5108 gives each payload lane its own FIFO with its own capacity eviction, so the rate at which *queued* sends are dropped before reaching the wire can differ from pre-#5108. That feeds `total_sends`, the numerator — which this PR's own review already demoted from headline to coarse cross-check with its contaminants enumerated (heals, PUT/GET-cache/Resync/first-install/delegate sends, retries, re-emits). So it is a known-soft number that got marginally softer, not a newly broken one. The denominator (`applies_client_local_changed`, counted in `update_contract` at apply time) is untouched by queue scheduling.

Net: no action needed, and the biases still point the way the docs say — all upward, so ≈17 remains strong evidence of one-hop reach while ≫17 stays merely suggestive.

[AI-assisted - Claude]

Issue #5110, opened 2026-08-03
priority_select tests use tight wall-clock timeouts as assertions; ~20 sites flake under load


sanity:
## Summary

`crates/core/src/node/network_bridge/priority_select/tests.rs` uses ~20 tight
wall-clock `tokio::time::timeout` bounds (5ms–300ms) as if they were assertions.
They are not: in every case the property under test is "the stream yields the
event" (waker registration, priority ordering, cancellation safety), and a
regression makes the poll hang **forever**. The timeout is a hang guard. Sized as
a latency budget, it instead measures whether the producer task got scheduled
promptly — which fails on a loaded machine while the behaviour under test is fine.

## Observed

`test_priority_select_event_loop_simulation` failed with
`Iteration 0 should complete` during a full `cargo test -p freenet --lib` run on
nova at load average ~70 (many concurrent agent builds). It passes 3/3 in
isolation on the same checkout. Its producer task sleeps 10ms of real time before
its first send, and the consumer bounded each poll at 50ms — so iteration 0 had a
~40ms scheduling margin.

That one instance is fixed in #5109 (10s hang guard, reasoning recorded
inline), because it was blocking an unrelated PR's CI. The rest of the file is
untouched.

## The remaining sites

```
$ rg -n "timeout\(Duration::from_millis" crates/core/src/node/network_bridge/priority_select/tests.rs
```

At the time of filing: lines 118, 183, 232, 277, 334, 761, 1075, 1173, 1257,
1367, 1534, 1697, 1808, 1863, 1878, 1928, 1940, 1952, 1994, 3166. Two 5ms bounds
(334, 1863) are the most exposed; several 50ms and 100ms bounds sit in loops
behind spawned producers with real `sleep`s, the same shape as the one that
failed.

## Suggested fix

Per-site judgement, not a blanket bump:

- **Hang guards** (most of them): a single named `const` with a generous bound
  (seconds) and a comment saying it is a hang guard, not a latency assertion.
- **Sites that deliberately assert "nothing is ready yet"** (the 5ms polls at 334
  and 1863 look like this): a wall-clock bound is the wrong primitive entirely —
  these want `tokio::time::pause()` + `advance()`, or an explicit readiness
  handshake, so "not ready" is a fact rather than a race.
- Producer-side `sleep(Duration::from_millis(10))` used to force the consumer to
  park first should become an explicit synchronization point where the test
  depends on the ordering.

[...]

PR #5114, opened 2026-08-03
test(priority-select): size the event-loop poll bound as a hang guard

sanity:
Split out of #5109, where it rode along inside a `fix(ring)` PR and tripped the scope rule-review check.

## Why it is separate

It has no dependency on the retention change. The flake was surfaced by load average ~70 on the build host (four concurrent release builds), not by the ring work, and #5109's 12 CI checks were green without it. So it is not covered by the "tiny follow-up needed to make the PR mergeable" exemption — it is a genuinely independent fix.

## The change

`test_priority_select_event_loop_simulation` bounded each poll at 50 ms of REAL time while its producer task sleeps 10 ms before its first send, leaving iteration 0 a ~40 ms scheduling margin. It failed on a loaded machine and passed 3/3 in isolation.

The bound is a **hang guard, not a latency assertion**: the property under test is that the stream keeps its waker registration across `.next().await` calls, and a regression makes iteration 0 hang forever — so any generous bound fails identically. Widening it therefore costs no coverage. Wall-clock durations are not a synchronization primitive (`.claude/rules/flaky-tests.md`).

## Scope

The same file has ~20 more timeouts of this shape. Several deliberately assert "nothing is ready yet" and want different treatment (`start_paused`) rather than a bigger number, so they are analysed per-site in #5110 rather than swept in here. This changes only the one test that actually failed.

Refs #5110

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/priority_select/tests.rs
+++ b/crates/core/src/node/network_bridge/priority_select/tests.rs
@@ -428,13 +428,23 @@ async fn test_priority_select_event_loop_simulation() {
 
     let mut received_events = Vec::new();
 
+    // Hang guard, NOT a latency budget. The property under test is that the
+    // stream keeps its waker registration across `.next().await` calls; a
+    // regression makes iteration 0 hang FOREVER, so any generous bound fails the
+    // same way. The previous 50ms bound read like an assertion but only measured
+    // whether the producer task — which sleeps 10ms of REAL time before its first
+    // send — got scheduled promptly, so it failed on a loaded machine while the
+    // behaviour under test was fine. Wall-clock durations are not a
+    // synchronization primitive (`.claude/rules/flaky-tests.md`).
+    const WAKER_REGISTRATION_HANG_GUARD: Duration = Duration::from_secs(10);
+
     // Simulate event loop: poll stream until we've received all expected messages (3+2+2+1 = 8)
     let expected_count = 8;
     for iteration in 0..expected_count {
         tracing::info!("Event loop iteration {}", iteration);
 
         // Poll the SAME stream on each iteration - waker registration is maintained
-        let result = timeout(Duration::from_millis(50), stream.as_mut().next()).await;
+        let result = timeout(WAKER_REGISTRATION_HANG_GUARD, stream.as_mut().next()).await;
         assert!(result.is_ok(), "Iteration {} should complete", iteration);
 
         let event = result.unwrap().expect("Stream should yield value");


sanity:
## Adversarial review — no blocking findings

One independent reviewer (Light tier: single-file, test-only change), which did not write the code and was told to assume a bug exists. Method: read the full diff, the full test function, the `PrioritySelectStream::poll_next` implementation it exercises, the mock stream types, the nextest CI config, and grepped for `start_paused` precedent.

### The central claim was checked, not accepted

The PR rests on "this bound is a hang guard, not a latency assertion, so widening it costs no coverage." That is only true if no waker regression can produce a *slow but finite* poll — otherwise the old 50 ms bound was catching something real. Verified against `poll_next`: it is pure synchronous poll code with no internal sleeps or timing logic. Every branch either returns `Poll::Ready` immediately when a channel already has data, or falls through registering `cx`'s waker on the real mpsc-backed streams. Failure is therefore structurally binary — resolve near-instantly once data is enqueued, or never wake at all. The only source of delay is OS scheduling of the producer's `sleep(10 ms)`, which is the flake this PR describes rather than a regression in the code under test.

Also confirmed: the meaningful assertions are untouched (exact per-channel event counts and priority ordering), so the test is not left vacuous; and 10 s is safe against the harness — CI runs `cargo nextest run --profile ci` with a 120 s slow-timeout and `terminate-after = 2` (240 s cap), no per-test override, and the assertion fires on the first timeout rather than accumulating across all 8 iterations, so the pathological case is 80 s.

### Two P3 notes, both routed to #5110 rather than fixed here

**`start_paused` would have been the better fix.** This codebase already uses `#[tokio::test(start_paused = true)]` with `GlobalExecutor::spawn` producer tasks that sleep (`transport/peer_connection/outbound_stream.rs:1222`, `operations/op_ctx.rs:1077`). Virtual time removes the wall-clock race at its root instead of widening the margin, and this test has no real-duration assertions that would complicate it. Not a blocker for a deliberately narrow PR, but it means the shipped fix is a margin widening where a root fix was available.

**A pre-existing test-design weakness, neither introduced nor worsened here.** The mock streams never wake the task themselves, and all 8 real messages land in one scheduling burst after a single shared `sleep(10 ms)` with no yields between sends. So the wake-rescue path is effectively exercised once, on the empty→burst transition at iteration 0. A regression isolated to *one* channel's waker registration could be masked, because the other three real channels register correctly in the same `poll_next` call and share the same `Context` waker.

Both added to #5110 so the per-site analysis there accounts for them.

[AI-assisted - Claude]

PR #5116, opened 2026-08-03
docs(telemetry): complete the fan-out residual list; fix a shipped contradiction

sanity:
Follow-up to #5112. **Docs only** — no counter, schema, or behaviour change.

## Why

@team-lead asked for the sole-emitter claim in #5112 to be re-verified by exhaustive search against the tree that now also contains #5108, rather than by re-reading my own description of it. #5112 merged through the queue before that re-verification happened, so this closes it after the fact.

## The claim holds

#5108 modifies `broadcast_to_single_peer` in exactly two places and adds **zero** new early returns — all five existing returns are byte-identical to pre-#5108 — and leaves both `record_delivered` call sites untouched. Nothing it changed is upstream of `record_apply`. The shipped headline ratio is apply-side on *both* halves (`applies_network_relay_changed / applies_client_local_changed`), so #5108's emit- and delivery-side changes cannot touch it directly. All four documented residuals still hold exactly as written.

## What it turned up

**1. #5112 shipped a self-contradiction.** `ApplyOrigin`'s rustdoc still said the ratio was "SOUND … both halves come from the same instrumented function" — the exact sentence retracted in the module docs 270 lines above, *in the same commit*. The module doc was revised in review round 2; the rustdoc was not updated in lockstep. Neither round of reviewers caught it.

**2. The residual list was incomplete**, in ways predating #5108:

- It attributed ">1 broadcast per changed apply" solely to first-install + queued-op replay. Three re-emission funnels also re-fire an already-counted apply: the no-target retry loop (up to `MAX_BROADCAST_RETRIES` further events), the stash-recheck re-emit, and the `PendingBroadcastStore` interest flush.
- It omitted three paths where a changed apply yields **zero** fan-out: the contract-ban egress gate (reachable as a race — every `update_contract` call site is ban-gated upstream and the executor has none, and the widest window is the streaming pair, which gates on the header then awaits `assemble()` for seconds); a **third** broken-invariant check at the handler, which swallows the fan-out *after* a successful commit — a different mechanism from the one listed; and the no-targets give-up/stash path.

**3. A downward bias, which the previous framing excluded by assertion.** `BroadcastQueue`'s small-to-large permit upgrade (`ensure_capacity_for` phase 2) releases its permit and awaits the large pool unboundedly. The small drain worker reclaims that permit and dispatches another entry, which may also mispredict and park. Nothing bounds the parked count — `track_active`'s cap is a telemetry counter and the drain loop proceeds when it reports full — so parked upgraders queue ahead of the large worker on the same fair semaphore, the large lane backs up against the shared cap, and `evict_oldest` drops entries. Mispredictions are bursty and correlated across every peer of a contract, so **whole fan-outs** are dropped, receivers never apply, and `applies_network_relay_changed` never fires.

The trap this sets, and the reason it is worth documenting rather than filing: **a fall in R/C toward ~17 reads as "one-hop reach confirmed, we are done" when it can equally mean the broadcast queue started evicting fan-outs.** R/C is coupled to broadcast-queue scheduling health. The docs now say to check the queue's eviction and HOL counters for the same window before concluding anything from a *drop* in this number.

So the docs no longer claim a single bias direction. They said "all push the same way", then over-corrected to "can only push it up". Both came from wanting one arrow, and there isn't one.

## Not fixed here

The unbounded parked-upgrader behaviour itself, and the claim at `broadcast_queue.rs` that in-flight sends are "bounded by the `active` tracking cap" (false for the reason above). Both belong to #5108's author, and I am not claiming a regression — only that the behaviour is unbounded and undocumented. Raised with @team-lead separately.

## Validation

`cargo fmt --all --check` and `cargo test -p freenet --lib broadcast_payload_mix` (36 passed) green. Doc-only, so no counter or schema change to re-verify.

Refs #5062, #5091, #5108, #5112.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -130,7 +130,7 @@
 //! relay much and serve few local clients, so a population skewed toward them
 //! under-samples the denominator.
 //!
-//! #### Downward (fewer, but one of them is the dangerous one)
+//! #### Downward (including the one that can look like success)
 //!
 //! Small and diffuse: the broadcast dedup cache and the merge-failure backoff
 //! gate both short-circuit BEFORE `update_contract`, so some relayed applies
@@ -138,9 +138,14 @@
 //! — again the churn being measured); and fair-queue shedding drops
 //! sub-`ClientLocal` work first.
 //!
-//! Then the one that can manufacture a clean-looking result. **Introduced by
-//! #5108**, so R/C measured on 0.2.118 and earlier does not carry it, and a
-//! comparison spanning that boundary is confounded:
+//! Then the one that can manufacture a clean-looking result. The MECHANISM
+//! below is new in #5108, so a comparison spanning that boundary is
+//! confounded — but do not read that as "0.2.118 R/C is clean". Pre-#5108 the
+//! queue had a single drain worker that blocked inline on the large semaphore
+//! and evicted strict-FIFO from one 256-deep queue, which drops whole fan-outs
+//! more cleanly than the two-lane version; see the head-of-line figures
+//! measured on that build in `broadcast_queue`. Both builds deflate R/C; they
+//! do it by different routes:
 //!
 //! `BroadcastQueue`'s small-to-large permit upgrade (`ensure_capacity_for`
 //! phase 2) releases its permit and awaits the large pool with no timeout and
@@ -151,9 +156,11 @@
 //! `tracked: false` for untrack bookkeeping; it is not an admission gate).
 //! Parked upgraders queue ahead of the large drain worker on the same
 //! FIFO-fair semaphore, the large lane backs up against the shared depth cap,
-//! and `evict_oldest` drops entries. Since a fan-out's per-peer entries carry
-//! contiguous `seq` and eviction walks seq order, what is dropped is typically
-//! a WHOLE FAN-OUT rather than scattered targets. See #5118.
+//! and `evict_oldest` drops entries. Eviction walks `seq` order, and a
+//! fan-out's per-peer entries are enqueued in a tight loop, so they form a
+//! temporal cluster — concurrent fan-outs interleave their `seq` allocations,
+//! so this is a tendency rather than a guarantee, but what gets dropped is
+//! typically most of a FAN-OUT rather than scattered targets. See #5118.
 //!
 //! A dropped fan-out means the receiving peers never apply, so
 //! `applies_network_relay_changed` never fires: **R/C is biased DOWN, and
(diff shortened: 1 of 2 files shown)


sanity:
## Consolidated review

LIGHT tier by diff shape (doc-only), but run as three full rounds because the docs *are* the product here — a wrong claim sends someone into a multi-week topology rewrite. Six independent blind sub-agent reviewers across three rounds, two lenses each (factual accuracy against the code; reader-outcome/usability). None wrote the text.

Rounds ran on `7fdc96378`, `bb306b7c6`, and `7ac6d3da0`; the PR was re-reviewed after every content change per the per-code-content rule.

### Findings that changed the work

| Round | Finding | Disposition |
|---|---|---|
| 1 | **BLOCKER** — the JSON emit-site comment, the closest thing here to a dashboard consumer's docstring, still said "both are biased UPWARD" — the exact claim this PR exists to retract. Fixed in two places, missed in the third. | Fixed |
| 1 | The instruction to "check the queue's eviction and HOL counters for the same window" was unfollowable: counters unnamed, in a different event, and **there is no same window** (60 s rollup vs lifetime-cumulative counters exported every 30 min). | Fixed — named, located, cadence stated, and "flat" defined as the delta between snapshots |
| 1 | The reader was left with two hedged readings and no way to choose. | Fixed — replaced with an explicit procedure and bands |
| 1 | The parked-upgrader finding had no durable home ("raised with the team lead"), while every other follow-up in this file cites an issue. | Filed #5118 and cited it |
| 1 | A known-false claim was being left in `broadcast_queue.rs` on ownership grounds. Reviewer: ownership is a reason to loop in the author, not to ship a statement you know to be untrue. | Fixed in place, twice — the same false bound appeared again 45 lines above in `UPGRADE_SLOT_HOLD_WINDOW`'s rustdoc |
| 2 | **The ">1 broadcast per apply" bullet was wrong, and I introduced it in round 1.** The replay loop actually *inflates* R/C (each replayed op commits a genuinely different state → receivers' applies are `changed`), and the three re-emission funnels fire only when nothing was sent, so they belong to the zero-delivery bullet directly above — whose sign is the opposite. Two consecutive bullets asserting opposite signs for one mechanism. | Fixed |
| 2 | The procedure's central conditional overclaimed: flat queue counters ⇒ "downward bias excluded" ⇒ upper bound. Those counters see one of ≥5 downward effects; `try_notify_node_event`'s drop is load-correlated the same way and has **no counter**. | Fixed — flat counters now license "the queue did not eat fan-outs", under its own heading so it can't be quietly upgraded |
| 2 | The ≫17 branch said to bound the heal contribution first — impossible, no heal counter exists. | Fixed — stated plainly, with the actionable version |
| 2 | The ≈17/≫17 boundary was undefined; a real reading of 41 fell in the gap and would have been read as "≫17 → topology". | Fixed — tens/ambiguous/hundreds bands, plus the note that R/C well *below* 17 is a third failure with a third remedy |
| 3 | **The de-duplication deleted the metric's core derivation while asserting it was preserved** — the module doc still pointed at `ApplyOrigin` for reasoning that now existed in neither place, behind a pointer that still resolved. | Fixed — restored into the single surviving copy |
| 3 | "All three re-emission funnels sit inside `targets.is_empty()`" was false for the `PendingBroadcastStore` flush. | Fixed, then refined again after final review named the load-bearing driver |

Smaller fixes: "gated in FOUR places" (grep finds 7-9, and the pre-commit gate rarely fires on the dominant UPDATE path); "contiguous `seq`" softened to a temporal cluster, since concurrent fan-outs interleave allocation; "0.2.118 does not carry it" corrected (the *mechanism* is new in #5108, the bias is not — the old single-worker queue evicted strict-FIFO); `capacity_evictions` named as the counter that licenses the conclusion rather than `active_tracking_overflow`, which only fires at the depth cap; the band edges demoted to an explicitly-unverified working prior; retraction narration cut from three tellings to one.

### Accepted, not fixed

- Two LOW imprecisions the final round called non-blocking were fixed anyway in `d776c30d0`, since both were the same species as the defect they replaced.
- The 200-300 stretch between bands is unassigned. Both edges are explicitly fuzzy and the surrounding text asks only which order of magnitude, so sharpening it would imply precision the data doesn't support.
- The unbounded parked-upgrader behaviour itself is **not** fixed here — #5118, for #5108's author.

### Validation

`cargo fmt --all --check`, `cargo clippy -p freenet --lib -- -D warnings`, `cargo check -p freenet --all-targets`, `cargo test -p freenet --lib` (4,589 passed, 0 failed), and `cargo doc -p freenet --no-deps` (no warnings in either touched file) all green locally. Doc-comment-only — every changed line is `//!` or `///`, and no source-scrape pin reads the changed prose.

[AI-assisted - Claude]

Issue #5115, opened 2026-08-03
Disconnect-grace teardown destroys the peer summary cache: 82% of interest removals, and two design problems that blocked the first attempt


sanity:
## Summary

A disconnect-grace interest teardown destroys the peer's cached `StateSummary` along with the `PeerInterest`, so the same peer reconnecting re-registers summary-less and the next broadcast to it ships FULL STATE. `DisconnectGrace` is **82% of all interest removals fleet-wide**, which makes this the dominant churn path.

An implementation was attempted in #5109 and **withdrawn after two rounds of FULL-tier review**, each of which found the same class of defect — a retained belief outliving a signal that should have invalidated it — in a new place. This issue records the evidence (which is solid and worth keeping) and the two design problems that must be solved before another attempt.

## Measured (0.2.118 fleet, 1,414 reporting peers, `network_efficiency_v1`)

Removal causes, and how many of those pairs were later recreated within the 20-minute correlation window:

| cause | removals | recreated |
|---|---|---|
| **DisconnectGrace** | **19,271,751 (82%)** | **2,645,774** |
| InterestsReplace | 2,896,191 | 1,316,229 |
| ChangeInterests | 684,897 | 665,794 |
| TtlExpiry | 498,714 | 58,229 |
| Unsubscribe | 216,855 | 107,509 |

Missing-summary full-state sends (cumulative, latest snapshot per peer):

| class | sends | bytes | mean |
|---|---|---|---|
| UntrackedFirstObserved | 771,400 | 78.20 GB | 101 KiB |
| TrackedFirstNew | 258,021 | 77.35 GB | 300 KiB |
| **TrackedFirstRecreated** | **25,704** | **7.72 GB** | 300 KiB |
| UntrackedFirstRecreated | 21,150 | 6.53 GB | 309 KiB |
| every `*Repeat*` class | 186 | 31 MB | — |

`ms_age` — the age of the summary-less entry at its first send: **164,867 under 1s (58%)**, 54,655 at 1–9s, 13,770 at 10–59s, 13,707 at 1–5 min, 35,728 at ≥5 min. So the broadcast usually beats the reconnecting peer's `Summaries` reply, which is the window retention would close.

Summary sizes, for anyone sizing a cache: the `interest_sync_summaries_*` outbound rollup measures a **262-byte mean per `Summaries` entry** across 228M entries.

## Honest bound on the reachable benefit

[...]

PR #5117, opened 2026-08-03
fix(telemetry): stop the untracked staleness reset from hiding recreation

sanity:
## Problem

`begin_peer_summary_broadcast`'s untracked arm resets a pair's missing-summary history when it goes stale, and the reset wiped `recent_removal` alongside the send counter:

```rust
if record.last_observed.is_some_and(|observed| now - observed > INTEREST_TTL) {
    record.send_starts = 0;
    record.recent_removal = None;   // <- removed by this PR
}
```

Those are **different clocks**. `last_observed` is stamped *only* on this untracked broadcast path, so it says nothing about when the pair's interest entry was removed. The removal's own freshness is already enforced two lines below, by the `removed_at` filter that survives the reset:

```rust
let recreated = record.recent_removal
    .filter(|(_, removed_at)| now.saturating_duration_since(*removed_at) <= INTEREST_TTL)
    .is_some();
```

So the wipe added nothing and cost the signal. A pair observed untracked at T0, tracked, removed at T1, and broadcast to again at T2 > T0 + `INTEREST_TTL` lost its T1 removal and was reported `UntrackedFirstObserved` — "never seen before" — no matter how recent T1 was.

## Why it matters

This is the field the fleet's missing-summary attribution reads from. On the 0.2.118 fleet (1,414 reporting peers), `UntrackedFirstObserved` is the single largest missing-summary class at 771,400 sends / 78.20 GB against `UntrackedFirstRecreated`'s 21,150 / 6.53 GB — and the wipe means an unknown share of the former is actually the latter. That is the discrepancy previously attributed to the interest seeding / heartbeat chain when the class looked far larger than #4952 should allow.

### What this does and does not claim

Being precise here, because an earlier draft of this PR overstated it and the review caught it:

- It moves sends **between** `UntrackedFirstObserved` and `UntrackedFirstRecreated`, and — via a second reader, below — between `TrackedFirstNew` and `TrackedFirstRecreated`. Totals are unchanged.
- It does **not** explain why `UntrackedRepeat*` is ~zero. First-vs-Repeat is decided by `send_starts`, whose reset this PR deliberately keeps. `UntrackedRepeat*` is structurally near-zero for an unrelated reason: `record_delivery_to_interest` upserts an interest entry on delivery (#4952), so a delivered untracked send makes the pair *tracked* and the next broadcast takes the tracked arm. An earlier draft called the `UntrackedRepeatSequential ≈ 0` reading a "fingerprint" of this bug; it isn't, and it will still read ~zero after this lands.
- It does **not** bound the effect size. The staleness branch only fires when `last_observed.is_some()`, so the affected population is the intersection of: previously broadcast to while untracked >20 min ago, AND removed within the last 20 min, AND the history entry surviving the LRU across both. A pair with `last_observed == None` never tripped the wipe and already classified correctly. I have not measured what fraction of the 771K moves, and this PR does not assert one.
- It does **not** make the counter correct. `recent_removal` is honoured only within `INTEREST_TTL` and lives in a bounded LRU (raised to 65,536 by #5097, still overflowing 1,597,674 times fleet-wide on 0.2.118), so recreation stays undercounted. This removes a systematic mis-attribution; it does not remove the undercount.

## Two readers, not one

`recent_removal` has a second reader in `register_peer_interest_from` (`interest.rs:1498-1512`), which sets `NeverPopulatedOrigin::New { recreated }` — feeding the tracked-arm `TrackedFirstNew`/`TrackedFirstRecreated` split — and increments `recreated_after_removal`. It applies the same `removed_at` filter, so the deletion is equally safe there, but it means **four** series step at this change, not two: the untracked split, the tracked split, and `recreated_after_removal`.

That discontinuity is recorded on the wire-schema field docs (`router.rs`, on `ms_s` and `recreated`), following the precedent set by the `queue` field's #4961 note — so it survives where a PR body would not.

## Safety

Behaviourally this changes telemetry classification only — no routing, no payload selection, no broadcast targeting, no state authorization. Both reviewers traced the full consumer chain independently: `attempt.class` is read at exactly one non-test site to index `delivered_sends`/`delivered_bytes`; `NeverPopulatedOrigin` is read at exactly one site to pick a class; and `broadcast_to_single_peer`'s payload selection keys off `their_summary` and `tracked_missing_reason`, neither derived from `recent_removal`.

It cannot resurrect an arbitrarily old removal — both readers filter on `removed_at <= INTEREST_TTL`, which is what the wipe was redundant with, so it could only ever suppress true positives.

It cannot double-count. On the untracked path, honouring removal T1 at broadcast T2 requires `T2 − last_observed > TTL` and `T2 − T1 <= TTL`; `last_observed` is then unconditionally re-stamped to T2, so a second reset needs `T3 − T2 > TTL`, whence `T3 − T1 > TTL` and the filter rejects the same removal. Between resets `send_starts > 0` forces the Repeat arm, which never consults `recreated`. On the registration path the increment requires the peer to be absent from `interested_peers`, and the only removal site overwrites `recent_removal` with a fresh stamp — so one removal yields at most one recreation count.

No growth or LRU-profile change: `MissingPairHistory` is `Copy` with the removal stored inline, and `history.put` runs on every untracked broadcast regardless, so key set, entry count and recency order are byte-identical pre- and post-fix.

The wiped line arrived in `bf4568c40` (#5091) with no rationale comment and no pin test (`git log -S` confirms), and nothing in the repo asserted the old behaviour — so removing it contradicts no documented invariant and breaks no test.

## Test

One regression test pinning three things, each guarding a different regression route:

- **T2** — a removal one minute old classifies as a recreation even though the last untracked observation is 22 minutes old. Verified to fail without the fix: `left: UntrackedFirstObserved, right: UntrackedFirstRecreated`.
- **T3** — a removal 22 minutes old is still refused, so the fix cannot drift into honouring a stale removal. Independently load-bearing: replacing the surviving `removed_at` filter with `recent_removal.is_some()` fails here.
- **T4** — the second reader sees it too, asserted on `recreated_after_removal[DisconnectGrace]` rather than on a class. This counter reads 0 pre-fix and 1 post-fix, so it is a real pin, and it was the series most at risk of stepping unpinned.

## Provenance

Split out of #5109, which attempted to retain peer summaries across a disconnect teardown. That mechanism was withdrawn after two rounds of full-tier review (see #5115); this classifier fix was reviewed as correct and independently valuable by two of the four reviewers, so it is landing on its own.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/ring/interest.rs
+++ b/crates/core/src/ring/interest.rs
@@ -1227,18 +1227,30 @@ impl<T: TimeSource + Sync> InterestManager<T> {
             // Reset the SEND counter only. `recent_removal` must NOT be wiped
             // here: `last_observed` is stamped exclusively on this untracked
             // path, so it says nothing about when the pair's interest entry was
-            // removed, and the removal's own freshness is already enforced by
-            // the `removed_at` filter immediately below.
+            // removed. Its freshness is enforced independently, by the
+            // `removed_at` filter in BOTH readers — the one immediately below,
+            // and `register_peer_interest_from`'s, which drives
+            // `NeverPopulatedOrigin::New { recreated }` and
+            // `recreated_after_removal`. So the wipe was redundant with those
+            // filters and could only ever suppress true positives.
             //
-            // Wiping it conflated the two clocks and systematically hid
-            // recreation: a pair observed untracked at T0, then tracked, then
-            // removed at T1, then broadcast to again at T2 > T0 + INTEREST_TTL
-            // lost its T1 removal and reported `UntrackedFirstObserved` instead
-            // of `UntrackedFirstRecreated`. That is why the former is the
-            // largest missing-summary class on the fleet (771K sends / 78.2 GB
-            // on 0.2.118) while `UntrackedRepeatSequential` is ~zero: every
-            // churn cycle longer than INTEREST_TTL landed in the "never seen
-            // before" bucket.
+            // Wiping it conflated the two clocks and hid recreation: a pair
+            // observed untracked at T0, tracked, removed at T1, then broadcast
+            // to again at T2 > T0 + INTEREST_TTL lost its T1 removal and
+            // reported `UntrackedFirstObserved` however recent T1 was.
+            //
+            // Scope of the correction, stated precisely because it is easy to
+            // overclaim: this moves sends between `UntrackedFirstObserved` and
+            // `UntrackedFirstRecreated` (and, via the second reader, between
+            // `TrackedFirstNew` and `TrackedFirstRecreated`). It does NOT
+            // change the First-vs-Repeat split, which is decided by
+            // `send_starts` — still reset on the line below — and it does not
+            // change the total number of missing-summary sends or bytes. The
+            // affected population is only pairs that were broadcast to while
+            // untracked more than INTEREST_TTL ago AND removed within the last
+            // INTEREST_TTL AND whose history entry survived the LRU across
+            // both; a pair with `last_observed == None` never tripped the wipe
+            // and already classified correctly.
             record.send_starts = 0;
         }
         let first = record.send_starts == 0;
@@ -3151,15 +3163,22 @@ mod tests {
     ///
     /// `last_observed` is stamped only when a pair is broadcast to while
     /// UNTRACKED, so it measures a different clock from the removal. Wiping the
-    /// removal alongside the send counter made every churn cycle longer than
-    /// `INTEREST_TTL` report as `UntrackedFirstObserved` ("never seen before")
-    /// rather than `UntrackedFirstRecreated` — which is why the former is the
-    /// largest missing-summary class on the fleet while `UntrackedRepeat*` is
-    /// ~zero. Without the fix this test observes `UntrackedFirstObserved`.
+    /// removal alongside the send counter made a pair whose removal was recent,
+    /// but whose last untracked observation was stale, report
+    /// `UntrackedFirstObserved` ("never seen before") rather than
+    /// `UntrackedFirstRecreated`. Without the fix, step T2 below observes
+    /// `UntrackedFirstObserved`.
     ///
-    /// The removal's own freshness is still enforced, by the `removed_at`
-    /// filter that survives the reset — pinned by the second half of this test,
-    /// so the fix cannot drift into honouring an arbitrarily old removal.
+    /// Three things are pinned, because each guards a different way this can
+    /// regress:
+    /// - T2: a recent removal survives a stale-`last_observed` reset.
+    /// - T3: a removal older than `INTEREST_TTL` is still refused — the
+    ///   `removed_at` filter is the remaining guard, so the fix cannot drift
+    ///   into honouring an arbitrarily old removal.
+    /// - T4: the SECOND reader (`register_peer_interest_from`) sees it too.
+    ///   That reader drives `recreated_after_removal` and the tracked-arm
+    ///   `TrackedFirstNew`/`TrackedFirstRecreated` split, so it steps on this
+    ///   change as well and would otherwise be unpinned.
     #[test]
     fn untracked_staleness_reset_preserves_a_recent_removal() {
         let (manager, time) = make_manager();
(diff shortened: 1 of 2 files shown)


sanity (commented):
## Light-tier review — 2 independent blind lenses

Tier: **Light** per `multi-model-review.md` — one line of production code, telemetry classification only, no high-risk surface. It removes code, so the skill's Light set is skeptical + big-picture. Both read the checked-out branch, not just the diff. No external non-Claude pass (codex quota / Gemini CLI unavailable since 2026-06-24); recorded per the fallback rule.

**Both reviewers cleared the mechanism and found the same class of problem in the narrative** — which mattered, because that narrative was going into permanent source comments. All findings fixed in `32b4ada5`.

### Fixed

**1. The `UntrackedRepeatSequential ≈ 0` attribution was wrong.** (big-picture) I called it "the fingerprint of this bug". It isn't: `recent_removal` only chooses between `UntrackedFirstObserved` and `UntrackedFirstRecreated`; First-vs-Repeat is decided by `send_starts`, whose reset this PR deliberately keeps. And the class is structurally near-zero for an unrelated reason — `record_delivery_to_interest` upserts on delivery (#4952), so a delivered untracked send makes the pair tracked and the next broadcast takes the tracked arm. It will still read ~zero after this lands, and an operator reading my comment would have concluded the fix didn't work. Removed from the comment, the test doc, and the PR body.

**2. Effect size was overstated with no bound.** (big-picture) "Every churn cycle longer than INTEREST_TTL" — but the staleness branch only fires when `last_observed.is_some()`, and `last_observed` is stamped only by a *previous untracked broadcast*. A pair that never had one already classified correctly. The real population is an intersection, and I have not measured what fraction of the 771K moves. The comment now states that explicitly, and the PR body concedes it rather than asserting a number. This is the same inference shape #5115 criticised in #5109; fair to hold this PR to it.

**3. `recent_removal` has a SECOND reader, unnamed and unpinned.** (both reviewers, independently) `register_peer_interest_from` reads the same field to set `NeverPopulatedOrigin::New { recreated }` — feeding the tracked-arm `TrackedFirstNew`/`TrackedFirstRecreated` split — and to increment `recreated_after_removal`. So **four** series step here, not two. My comment said "the `removed_at` filter immediately below", naming only one. Both readers apply the same filter so the deletion is equally safe, but an auditor couldn't establish that from the comment. Comment now names both; added a T4 step to the test asserting `recreated_after_removal[DisconnectGrace]`, verified independently load-bearing (`left: 0, right: 1` with the fix reverted and the T2 assertion neutralised, so T4 fails on its own rather than behind T2).

**4. The discontinuity belonged on the wire-schema field, not the PR body.** (big-picture) `NetworkEfficiencyV1.queue` carries a 35-line "changed MEANING at #4961 — read pre- and post-fix series separately" note; that is this repo's durable mechanism for exactly this, and `v` was not bumped for #4961 either. Added equivalent notes to `ms_s` and `recreated` covering all four affected rows.

**5. T3 leaked an in-flight attempt.** (skeptical) `MissingSummaryAttempt` has no `Drop` — only the guard does — so the final step left `missing_summary_active` at 1. Harmless today, but a trap: anyone appending a step would get `inflight = true` and a class unrelated to what is pinned. Now dropped through the guard like the earlier steps.

**6. Hard-coded fleet figures in a permanent comment**, stated present-tense in text whose whole purpose is to invalidate them. Removed.

**7. LRU figure cited without its cap.** (big-picture) `MISSING_SUMMARY_HISTORY_SIZE` was raised 4,096 → 65,536 by #5097, which is in 0.2.118. Load-bearing rather than cosmetic, since this fix's effect size depends on entries surviving the LRU for >20 min. Now stated.

### Verified sound by both lenses

Full census of `recent_removal`: definition, one writer, two readers, both `removed_at`-filtered. `attempt.class` is read at exactly one non-test site to index `delivered_sends`/`delivered_bytes`; `NeverPopulatedOrigin` at exactly one site to pick a class; `broadcast_to_single_peer`'s payload selection keys off `their_summary` and `tracked_missing_reason`, neither derived from this field. **No routing, payload selection, broadcast targeting, or authorization input.**

No growth or LRU-profile change — `MissingPairHistory` is `Copy` with the removal inline, and `history.put` runs on every untracked broadcast regardless, so key set, entry count and recency order are identical pre- and post-fix.

No double-count, proved both ways. Untracked path: honouring T1 at T2 requires `T2 − last_observed > TTL` and `T2 − T1 <= TTL`; `last_observed` is re-stamped to T2, so a second reset needs `T3 − T2 > TTL`, whence `T3 − T1 > TTL` and the filter rejects it — and between resets `send_starts > 0` forces the Repeat arm, which never consults `recreated`. Registration path: the increment requires the peer absent from `interested_peers`, and the only per-peer removal site overwrites `recent_removal` with a fresh stamp, with every other removal helper delegating to it.

The deleted line arrived in `bf4568c40` (#5091) with no comment and no test (`git log -S` confirms) — the "no rationale, no pin" claim checks out.

Mutation-tested both original halves: restoring the wipe fails T2 (`UntrackedFirstObserved` vs `UntrackedFirstRecreated`); replacing the surviving `removed_at` filter with `recent_removal.is_some()` fails T3. Neither is redundant.

Gates on `32b4ada5`: fmt, clippy `-D warnings`, 4,590 lib tests, all-targets, simulation_tests — all green.

[AI-assisted - Claude]

PR #5119, opened 2026-08-03
fix(broadcast): bound the lane-correction parking area (fixes unbounded growth shipped in #5108)

sanity:
## RELEASE BLOCKER — fixes a bug already on `main`

#5108 shipped an **unbounded parking area** in the lane-correction path. This fixes it. Do not cut a release from `main` without this.

## The bug

`ensure_capacity_for` phase 2 drops its small-lane permit and awaits the large pool unboundedly. `drain_lane` immediately reclaims that permit and dispatches another entry, which may also mispredict and park. Nothing bounds the parked count — `track_active` returns `false` at the 256 cap and the drain loop proceeds anyway (`broadcast_queue.rs:1182-1189`); it is a telemetry counter, not admission control. Each parked send holds a serialized payload that is >= 64 KiB by construction.

Reachable in **steady state**, not only in bursts: any contract whose deltas never beat its state mispredicts on every send to every peer, indefinitely. `delta_incompat` does not help — it arms on contracts that *reject* deltas, not ones whose deltas are merely never smaller.

### It is worse than memory growth

Parked upgraders queue ahead of the large worker on the same FIFO-fair semaphore, so `large_order` backs up against the shared queue cap and `evict_oldest` drops the **globally oldest** entry. Mispredictions are bursty and correlated across every peer of a contract, so this discards **whole fan-outs**, not stray targets — an update-propagation failure.

Found independently by two reviewers from different starting points (a round-3 concurrency review of #5108, and a re-verification of #5112's telemetry correspondence against the merged tree).

Also corrects `broadcast_queue.rs:149-151`, which claimed in-flight sends are "bounded by the `active` tracking cap". That claim was false, for the reason above.

## It is live in production now

Framework, an ordinary node hosting 864 contracts, reports `queued_small_actual_large` = **240 per node-day carrying 61.3 MB/day** at 1.0h uptime. The parking path is entered in normal operation — roughly 10 mispredictions an hour, each parking a ~255 KB payload. This is not a theoretical edge case.

## The fix

Phase 2 takes a **parking permit before** giving up its small-lane slot, from a pool of `MAX_PARKED_LANE_UPGRADES`. When parking is exhausted the send keeps its small-lane slot and waits under it, so the small lane applies backpressure — bounded work — instead of the node accumulating unbounded payloads. That is the correct behaviour for an uplink genuinely saturated with full-state sends. Nothing is on the wire while no permit is held.

## Review

Four rounds. Round 3 found the bug (plus a P0: a test that had gone vacuous in the same commit that added it — the two-phase wait let a send re-acquire its own permit inside the time bound). Round 4 found no P0/P1, verified the simpler wait-for-a-parking-slot shape is equivalent to and better ordered than the `select!` alternative, confirmed both rewritten tests are genuinely armed, and fixed seven P2s — including two doc sites still describing behaviour round 4 had removed, and a cancellation test whose "took no large permit" assertion could not fail because it held the pool's only permit itself.

## Note on scheduling validation

The #5108 scheduling change this sits on top of is measuring well in a soak on framework. Matched-uptime, control vs fixed: `scheduled_large` 29,664 -> 2,880 per node-day, `queued_large_actual_small` 17,976 -> 0, lane misclassification 60.6% -> 0.0%. `capacity_evictions`, `dedup_replacements` and `enqueues_while_pair_active` all fell to zero, so the fix also removes duplicate scheduling work rather than only redistributing it. That validation is the reason this parking fix matters — the scheduling work is worth shipping, and cannot ship without it.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -51,10 +51,16 @@ const STREAM_COMPLETION_TIMEOUT: Duration = Duration::from_secs(120);
 /// narrow large-payload pool instead of the wide small-payload pool.
 ///
 /// Applied to the PREDICTED wire payload at enqueue time (see
-/// `queue::classify_payload_lane`) and again to the ACTUAL wire payload once it
-/// is known (see [`SendLanePermit::ensure_capacity_for`]). It used to be
-/// applied to the full contract STATE size, which is not what goes on the wire
-/// whenever a delta is sent — the classification half of #4961.
+/// `queue::classify_payload_lane`) and again to the selected payload — the
+/// delta or the full state — once it is known (see
+/// [`SendLanePermit::ensure_capacity_for`]). It used to be applied to the full
+/// contract STATE size, which is not what goes on the wire whenever a delta is
+/// sent — the classification half of #4961.
+///
+/// "Selected payload" is not quite the byte count on the wire: the sender's own
+/// summary and the message framing ride along with it, so a small payload with
+/// a large summary is under-counted here. That blind spot is pre-existing (the
+/// old state-size classification had it too) and is not what #4961 was about.
 const BROADCAST_QUEUE_PAYLOAD_SIZE_THRESHOLD: usize = 64 * 1024;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


sanity:
Closing as a duplicate of #5120, which was opened first and is the one to review.

My error: I concluded the owning agent had stalled (no branch pushed under the name I had suggested, no activity in its worktree for 12 minutes) and opened this from its local commits. It had in fact already pushed the same work under a different branch name — `fix/broadcast-lane-upgrade-parking` — before my message suggesting a name reached it. Both PRs point at the identical head `201a98017`, so no content is lost by closing this one.

Everything in this description applies to #5120. Carrying over the one thing that PR may not have, since it is production evidence gathered separately: framework, an ordinary node hosting 864 contracts, reports `queued_small_actual_large` = **240 per node-day carrying 61.3 MB/day** at 1.0h uptime. The parking path is entered in normal operation, so this is not a theoretical edge case.

Also deleting the duplicate branch `fix/bound-lane-upgrade-parking`.

Issue #5133, opened 2026-08-03
freenet service (disable/enable/report) resolves Roaming config dir on Windows, diverging from network/local


sanity:
## Summary

Found while investigating #5132 (config.toml gets scrubbed on launch). Not itself a cause of that issue's symptom (which is fixed in #5132's PR), but a related, real bug in the same area worth tracking separately.

On Windows, `freenet service <subcommand>` (e.g. `disable`, `enable`, `report`) resolves a **different** config directory than `freenet network` / `freenet local`.

## Root cause

`ConfigArgs::build_with_gateways_index()` (`crates/core/src/config.rs`) explicitly sets `self.config_paths.config_dir = Some(defaults.config_local_dir()...)` before building — i.e. **Local** AppData (`%LOCALAPPDATA%`), per the #3739 migration (Roaming syncs across domain machines and is wrong for node data).

`freenet_main()`'s `Command::Service` arm (`crates/core/src/bin/freenet.rs`, around line 1083) instead calls `cli.config.config_paths.build(cli.config.id.as_deref())?` **directly**, bypassing that step. Since `--config-dir` isn't passed on the CLI, `ConfigPathsArgs::build()` falls through to its own fallback (`crates/core/src/config.rs`, around line 2684-2692), which resolves `defaults.config_dir()` — **Roaming** AppData (`%APPDATA%`) — instead of `config_local_dir()`.

## Impact

- `freenet service disable` / `enable` write and check the daemon-disabled marker via this Roaming-resolved dir (`service.rs`, `daemon_control.rs`), while `freenet network`'s startup check reads the marker via the Local-resolved dir (`freenet.rs`). These never agree on Windows, so `freenet service disable` is effectively a no-op there.
- `freenet service report`'s config-file collector (`report.rs`) also scans the Roaming-derived dir first, so diagnostics can't find the real (Local) `config.toml`.

This is the same Roaming-vs-Local divergence class already fixed once for `uninstall --purge` (#3969) and documented at length in `purge.rs`'s doc comments — the `service` subcommand is one of the call sites that migration didn't reach.

## Suggested fix

Route the `Command::Service` arm through the same `config_local_dir()`-setting logic `build_with_gateways_index()` uses, or extract that logic into a shared helper both paths call.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-08-14T01:13:13.377182196+00:00
Issue hash: `578c2fe6`

## Value Assessment

[...]

PR #5135, opened 2026-08-03
fix(windows): add wrapper single-instance guard, make config.toml writes atomic

### LINKED ISSUE #5132: config.toml gets scrubbed on launch
running freenet on windows 11. every time I edit the config.toml (%localappdata%/The Freenet Project Inc/Freenet/config/config.toml) and relaunch freenet (whether via the downloaded exe or %localappdata%/Freenet/bin/freenet.exe), any config modifications disappear and the config.toml reverts to its default. 

I assume this is because running freenet (from any method) effectively "reinstalls" certain files?

sanity:
## Summary

Fixes #5132 — on Windows, editing `config.toml` and relaunching Freenet reverted the edits to defaults.

**Root cause:** the Windows wrapper had no single-instance guard at all. Every relaunch (double-clicking the downloaded exe, or launching the installed `%localappdata%\Freenet\bin\freenet.exe`, while Freenet was already running via autostart or a prior launch) fell straight through to `kill_stale_freenet_processes`, which unconditionally kills every `freenet network` process for the current user — "stale" in name only, it can't distinguish an orphan from the process a live wrapper is actively supervising — then spawned a fresh node. When that kill raced the first wrapper's own respawn of its just-killed child closely enough, two node processes could concurrently read and (pre-fix) non-atomically rewrite `config.toml`, which is how a hand-edit could come back as defaults.

macOS already has the exact fix for this class of bug — a flock-based single-instance guard, added for a different trigger (launchd RunAtLoad racing an open-launch) — but it was `#[cfg(target_os = "macos")]`-gated and never extended to Windows; a stale comment on the guard's call site even asserted "Windows has install-time guards" as the reason, which isn't accurate (the install-time detection only checks *where* the binary lives, not whether an instance is currently running).

## Changes

1. **`single_instance.rs`** — implements the Windows analogue of the macOS flock guard: a named kernel mutex (`CreateMutexW`), acquired in `run_wrapper` before `kill_stale_freenet_processes`. `CreateMutexW`'s "already exists" signal fires even for a second handle from the same process, so it can be exercised directly in a `#[test]` rather than needing a simplified stand-in.
2. **`wrapper.rs`** — widens the guard's call site (and its already-cfg-gated import) from `macos` to `any(macos, windows)`; corrects the stale "Windows has install-time guards" comment; platform-specific log detail on the "another wrapper is running" exit path.
3. **`config.rs`** — makes the `config.toml` persist path atomic: write to a sibling temp file, then `rename` into place, instead of `File::create` truncating the destination directly. Closes the window where a concurrent reader (e.g. a second `freenet` process racing this one) could observe a truncated/partial file. The temp file's extension (`toml.tmp`) keeps `read_config`'s directory scan from ever picking it up as a candidate config file.
4. **`ci.yml`** — adds a `windows_unit` job (mirrors the existing `macos_unit`) running the service-command test suite on `windows-latest`. `windows_check` only `cargo check`s, which never executes tests — without this, the new Windows-only tests would compile but never actually run, matching `.claude/rules/deployment.md`'s "compilation is not verification" rule for platform-gated code.

## Tests

- `wrapper_single_instance_lock_is_exclusive` (Windows-only, `commands/service.rs`) — first acquire succeeds, second acquire in-process reports `AnotherWrapperRunning`, dropping the first allows a new acquire to succeed. Exercises the real production function directly (not a simplified stand-in), since `CreateMutexW`'s already-exists signal fires for a second handle from the same process.
- `persisting_config_never_truncates_existing_file_on_temp_write_failure` (`config.rs`) — forces the temp-file write step to fail (by pre-occupying its path with a directory) and asserts the destination `config.toml` survives byte-for-byte untouched. Fails against the pre-fix code (which writes directly to `config.toml` with no temp file, so nothing in this scenario fails and the file gets overwritten instead of preserved).

## Verification

- Windows target (`x86_64-pc-windows-gnu`) cross-compiles cleanly (`cargo check`), including the new tests.
- `cargo clippy -p freenet --bin freenet -- -D warnings` clean on the native target.
- All 108 `config::` tests and all 107 `commands::service::` tests pass on Linux.
- The new Windows-only test cannot be exercised on this (Linux) dev machine; the new `windows_unit` CI job is what will actually execute it — will confirm green there before merging.

## Out of scope (filed separately)

While investigating, found that `freenet service` (disable/enable/report) resolves a *different* config directory (Roaming, not Local AppData) than `freenet network`/`local` on Windows — a related but distinct bug (breaks `service disable`, not `config.toml` content). Filed as #5133 rather than folding into this PR.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -608,7 +608,15 @@ jobs:
   windows_unit:
     name: Windows Service Unit
     runs-on: windows-latest
-    timeout-minutes: 30
+    # This job is new (introduced alongside the #5132 fix), so unlike
+    # macos_unit/windows_check it has no warm Swatinem cache from prior
+    # main-branch runs yet — a stone-cold Windows compile of the full
+    # freenet bin + test harness (wasmtime, tao, tray-icon, criterion,
+    # proptest, ...) does not finish in 30 minutes. 60 minutes gives this
+    # one-time cold start headroom; revisit downward once main has run
+    # this job at least once (cache only saves on `github.ref ==
+    # refs/heads/main`, see the cache step below).
+    timeout-minutes: 60
     needs: fmt_check
     # Same skip conditions as test_unit (see test_unit comment for details).
     if: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -598,6 +598,52 @@ jobs:
       - name: Test service commands
         run: cargo nextest run -p freenet --bin freenet --profile ci -E 'test(commands::service)'
 
+  # Windows unit tests for the wrapper's Windows-only code paths (named-mutex
+  # single-instance guard, process classification, etc). `windows_check` above
+  # only `cargo check`s — compilation is not verification of a `#[cfg(target_os
+  # = "windows")]` path (see .claude/rules/deployment.md) — so without this job
+  # the Windows-specific tests in commands::service would never actually run.
+  # Added alongside the #5132 fix (Windows wrapper had no single-instance
+  # guard at all), mirroring macos_unit above.
+  windows_unit:
+    name: Windows Service Unit
+    runs-on: windows-latest
+    timeout-minutes: 30
+    needs: fmt_check
+    # Same skip conditions as test_unit (see test_unit comment for details).
+    if: |
+      (github.event_name == 'pull_request' || github.event_name == 'merge_group') &&
+      !startsWith(github.head_ref, 'release/v')
+
+    env:
+      RUST_LOG: error
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.93.0
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Install nextest
+        run: |
+          $tmp = New-TemporaryFile | Rename-Item -NewName { $_ -replace 'tmp$', 'zip' } -PassThru
+          Invoke-WebRequest -OutFile $tmp https://get.nexte.st/latest/windows
+          $outputDir = if ($Env:CARGO_HOME) { Join-Path $Env:CARGO_HOME "bin" } else { "$Env:USERPROFILE\.cargo\bin" }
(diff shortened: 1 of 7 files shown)


sanity:
## Review summary

Ran a Full-tier multi-perspective review per this repo's review policy (concurrency + CI-config are both high-risk surfaces): 5 independent, blind Claude reviewers, each with a distinct lens — code-first, adversarial/skeptical, testing-coverage, unsafe-FFI/concurrency, and CI-configuration.

**Findings, synthesized (deduplicated across reviewers) and addressed:**

1. **(Confirmed by 3/5 reviewers) Atomic config write wasn't atomic under *concurrent* writers.** The temp file used a fixed shared name (`config.toml.tmp`), so two processes racing `ConfigArgs::build()` could interleave writes to the same temp file before either renamed. Fixed: the temp filename now embeds the writing process's PID, plus added `fsync` before rename. Verified: traced all `ConfigArgs::build()` call sites to confirm it's never invoked concurrently within one process, so PID-uniqueness is sufficient.

2. **(High severity, adversarial lens) Self-relaunch races the old wrapper's own mutex.** Verified directly by tracing the code: the tray's shutdown handling on the "UpdatedRestarting" terminal status includes a hard-coded 1-second sleep before releasing the lock, which reliably outlasts how fast a freshly-spawned wrapper reaches the same acquire call. Without a fix, auto-update and crash-loop-rollback self-relaunch would near-certainly (not rarely) fail silently on both Windows and macOS. Fixed: bounded retry (~5s) around the acquire call before concluding a second, unrelated instance is running.

3. Added a source-scrape pin test verifying the guard acquire call precedes `kill_stale_freenet_processes` textually (a call-order invariant no behavioral test can exercise), mirroring this file's existing `windows_first_run_wiring_is_present` pattern.

4. Added `RUST_MIN_STACK` to the new `windows_unit` CI job for parity with `macos_unit`/`test_unit` (prevents a plausible SIGSEGV during LLVM optimization on this workspace's largest crate).

5. Documented a known, out-of-scope limitation: the mutex is session-scoped, not cross-session for the same Windows user (e.g. RDP).

A follow-up verification pass (adversarial re-check of just the new diff) confirmed both fixes are correct and complete, with no new issues introduced.

Non-blocking items intentionally left as-is (with rationale): a Windows-specific `rename` sharing-violation retry (fails loud, not silently — acceptable trade-off, deferred); two concurrent windows-latest CI jobs per PR (cost/queue-time only, not correctness).

[AI-assisted - Claude]

PR #5145, opened 2026-08-04
fix(fdev): resolve the build target dir from the crate being built, not fdev's own compile-time path

### LINKED ISSUE #5144: fdev build panics (exit 101, "Could not find workspace root") without CARGO_TARGET_DIR — the fallback resolves fdev's own compile-time path, not yours
## Summary

On a released `fdev` binary, `fdev build` panics when `CARGO_TARGET_DIR` is unset, with `Could not find workspace root` and exit 101. It panics **even when run from a crate that is a member of a real Cargo workspace**, because the fallback does not look at the user's directory at all — it resolves `fdev`'s *own* compile-time path.

Reproduced on `fdev 0.3.280` (release binary from the `v0.2.118` assets).

## Reproduction

```bash
T=$(mktemp -d); mkdir -p "$T/src"; cd "$T"
cat > Cargo.toml <<'EOF'
[package]
name = "panic-probe"
version = "0.1.0"
edition = "2021"
[workspace]
[lib]
crate-type = ["cdylib"]
[features]
default = ["freenet-main-contract"]
freenet-main-contract = []
EOF
echo 'pub fn nothing() {}' > src/lib.rs
printf '[contract]\nlang = "rust"\n' > freenet.toml

env -u CARGO_TARGET_DIR fdev build   # exit 101
CARGO_TARGET_DIR="$T/target" fdev build   # exit 0, "Contract compiled"
```

```
 INFO fdev::build: Compiling contract with rust, args: ["build", "--lib", "--target",
      "wasm32-unknown-unknown", "--features", "freenet-main-contract", "--release"]

thread 'main' (1740809) panicked at crates/fdev/src/util.rs:103:10:
Could not find workspace root
```

## The probe that names the cause

Run the same thing from a crate that **is** a member of a workspace whose root is one directory up:

```bash
T=$(mktemp -d); mkdir -p "$T/ws/member/src"
printf '[workspace]\nmembers = ["member"]\nresolver = "2"\n' > "$T/ws/Cargo.toml"
# ... member/Cargo.toml + src/lib.rs + freenet.toml as above ...
cd "$T/ws/member" && env -u CARGO_TARGET_DIR fdev build    # still exit 101, same panic
```

If the fallback searched upward from the working directory it would find `$T/ws` immediately. It does not:

```rust
// crates/fdev/src/util.rs
pub fn get_workspace_target_dir() -> PathBuf {
    std::env::var("CARGO_TARGET_DIR").map(PathBuf::from).unwrap_or_else(|_| {
        let manifest_dir = env!("CARGO_MANIFEST_DIR");        // fdev's OWN build-time path
        let workspace_root = find_workspace_root_from(manifest_dir);
        workspace_root.join("target")
    })
}

fn find_workspace_root_from(start_path: &str) -> PathBuf {
    PathBuf::from(start_path).ancestors()
        .find(|p| { p.join("Cargo.toml").exists() && /* contains [workspace] */ })
        .expect("Could not find workspace root")              // panic, not an error
        .to_path_buf()
}
```

`env!("CARGO_MANIFEST_DIR")` is expanded when **fdev** is compiled. On a release binary that is the build machine's path, which does not exist on the user's machine, so `ancestors()` finds nothing and `.expect` panics.

Called from `build::compile_rust_wasm_lib` (to set the child cargo's `CARGO_TARGET_DIR`) and again from `build::get_out_lib` (for the output path).

## Why it matters

- It affects the **prebuilt binaries**, which is the documented install path. A locally-built `fdev` will not panic, because its compile-time path exists on that machine — so this is invisible to anyone developing on a from-source build.
- The failure is a bare panic (exit 101) with a message that names nothing the user controls. Nothing points at `CARGO_TARGET_DIR`.
- The workaround is already known to the project: `freenet-agent-skills`' `dapp-builder/references/build-system.md` tells you to pin `CARGO_TARGET_DIR` per crate and mentions that fdev "panics". But it explains the cause as *"`fdev build` walks up to find `Cargo.toml`, then uses cargo's default `target/` (which is the workspace's)"* — which the workspace-member probe above disproves. The workaround is documented; the diagnosis beside it is not right.

## Suggested fix

Two independent parts, both small:

1. **Search from the working directory**, then fall back to `env!("CARGO_MANIFEST_DIR")` (harmless on a from-source build). That makes the documented behaviour true and fixes the workspace-member case.
2. **Return an error rather than panicking.** `Could not find workspace root` with no context, at exit 101, is not actionable; a plain error naming `CARGO_TARGET_DIR` as the remedy is.

Happy to open a PR for this if welcome.

## Environment

```
Freenet Development Tool 0.3.280
freenet 0.2.118 (738f99f0e2e7), prebuilt x86_64-unknown-linux-musl release assets
Linux, rustc 1.94.0
```

[AI-assisted - Claude]

magks:
Fixes #5144.

## Why

`fdev build` panics on a released binary when `CARGO_TARGET_DIR` is unset — exit 101, `Could not find workspace root` — and it panics **even from inside an ordinary Cargo workspace**, because the fallback never looks at the user's tree:

```rust
let manifest_dir = env!("CARGO_MANIFEST_DIR");        // fdev's OWN compile-time path
let workspace_root = find_workspace_root_from(manifest_dir);
```

`env!("CARGO_MANIFEST_DIR")` is expanded when *fdev* is compiled, so on a release binary it is the path on the machine that built the release. Nothing above it exists for the user, `ancestors()` matches nothing, and `.expect("Could not find workspace root")` panics. A from-source `fdev` doesn't panic — its compile-time path exists — it silently resolves a user's contract to **freenet-core's own** `target/` instead. Same root cause, two faces, and the source-build face is why this is invisible while developing on the repo and fatal on the published binaries the install docs point at.

The workaround (set `CARGO_TARGET_DIR` per crate) is already in `freenet-agent-skills`' `dapp-builder/references/build-system.md`, but the cause recorded there — *"`fdev build` walks up to find `Cargo.toml`, then uses cargo's default `target/`"* — is not what the code does. It never consults the user's directory at all. Hence the second probe in the issue: a crate that **is** a member of a workspace whose root is one level up still panics.

## What this changes

`get_workspace_target_dir` now takes the crate being built and resolves from there:

1. `CARGO_TARGET_DIR` when set — unchanged, and still the first thing checked.
2. Otherwise the `target/` of the workspace that crate belongs to, found by walking up **from the crate**.
3. Otherwise the crate's own `target/` — a crate in no workspace is an ordinary thing to build, and that is where cargo puts it.

`find_workspace_root_from` returns `Option<PathBuf>` instead of `.expect`-ing. Nothing on the `fdev build` path can act on a panic here.

Both call sites in `build.rs` already have the crate directory to hand (`compile_rust_wasm_lib`'s `work_dir`, and `get_out_lib`'s), so they pass it. They have to agree on the value or packaging looks where the build didn't put anything, so resolving both from the same `work_dir` also removes a way for them to drift.

The env read is split from the resolution (`resolve_target_dir(env_target_dir, crate_dir)`) purely so the fallback is testable — see below. Same injected-getter shape as `auto_update::detect_supervisor_status`.

**Behaviour change worth naming:** a from-source `fdev` used to put a user's contract artefacts under *freenet-core's* target dir. It now puts them under the user's. That is what the function name and doc comment always claimed, and it is the fix rather than a side effect — but it is a change, so it should be a deliberate one.

**Deliberately not touched:** `crates/core/src/util/workspace.rs` has the same shape, but it is only reachable from core's own test utilities, which by definition run from a source tree where `env!("CARGO_MANIFEST_DIR")` is valid. Different blast radius, different change; this PR is the fdev bug.

## Tests

Four in `crates/fdev/src/util.rs`. The two that matter fail without the fix — verified by reverting just the function body and re-running:

```
resolves_the_workspace_of_the_crate_being_built_not_fdevs_own ... FAILED
  left:  "/…/freenet-core-pr/target"        <- fdev's own workspace
  right: "/tmp/.tmpPSv6hQ/ws/target"        <- the workspace under build
a_crate_in_no_workspace_gets_its_own_target_dir_and_does_not_panic ... FAILED
```

- `resolves_the_workspace_of_the_crate_being_built_not_fdevs_own` — the regression test proper: a member crate in a temp workspace must resolve to *that* workspace's `target/`.
- `a_crate_in_no_workspace_gets_its_own_target_dir_and_does_not_panic` — the arm that used to be an `.expect`.
- `an_explicit_target_dir_wins_over_the_search` — `CARGO_TARGET_DIR` still short-circuits.
- `find_workspace_root_from_returns_none_when_there_is_no_workspace`.

One trap worth flagging for anyone testing this area: a first version of these tests read `CARGO_TARGET_DIR` from the real environment and skipped when it was set. `cargo test` sets it for its children, so all of them took the skip and **passed against the unfixed code**. That is why the resolution takes the environment as an argument rather than reading it — mutating the process environment is not an alternative, since `set_var` is unsound with threads and `unsafe` from Rust 2024.

`cargo fmt --all --check`, `cargo clippy -p fdev --all-targets -- -D warnings` and `cargo test -p fdev` are clean.

[AI-assisted - Claude]


sanity:
Reviewed this at Light tier (two independent blind lenses plus my own read). The fix is correct and I found no blocking issues. Notes below, starting with the CI failure since it is not yours.

## The NAT Validation failure is a stale base, not your change

`NAT Validation` failed twice on `de76ad6c` with `ConnectivityFailed("Network did not reach 100% connectivity within 120s")`. It is not caused by this PR:

* The NAT/ping tests never call fdev. `apps/freenet-ping/app/tests/common/mod.rs:515` carries its own private `find_workspace_root()`, so nothing in `crates/fdev/` can reach them. (That is a third copy of the same pattern you fixed, alongside the `crates/core/src/util/workspace.rs` one you correctly scoped out.)
* The failure is a peer-connectivity timeout that happens during network setup, before any contract artefact is touched.

The likely cause is that this branch is based on `77b3dad3` (Aug 3), which pins `freenet-test-network = "0.1.23"`, the crate that builds the Docker NAT network. `main` bumped it to `0.1.24` in #5220 on Aug 7. Every branch that passed NAT concurrently with your two failures is on 0.1.24. A rebase onto current `main` should clear it, and this PR needs one anyway (`BEHIND`). I have not proven this by rebasing, so treat it as high confidence rather than established.

## Two test findings

**1. The no-workspace fallback branch is never exercised.** In `a_crate_in_no_workspace_gets_its_own_target_dir_and_does_not_panic`, the fixture writes `[workspace]` into `solo`'s own `Cargo.toml`. `Path::ancestors()` yields the path itself first, so `find_workspace_root_from(&solo)` matches on `solo` and returns `Some(solo)`. That means `resolve_target_dir`'s `.unwrap_or_else(|| crate_dir.to_path_buf())` never runs in any test. Replacing that fallback with `panic!()` leaves all four tests green.

Both outcomes produce `solo/target`, so the assertion cannot distinguish "found itself as workspace root" from "found nothing and fell back". A fixture whose manifest has no `[workspace]` at all, under a tempdir with no `Cargo.toml` above it, would reach the branch the test name describes.

**2. `find_workspace_root_from_returns_none_when_there_is_no_workspace` cannot fail.** No `Cargo.toml` is ever written under `tmp`, so a match rooted inside `tmp` is unreachable by construction, and the `|| !found...starts_with(tmp.path())` clause tolerates any `Some(p)` from outside. The only value that would fail the assertion cannot occur. I understand the OR clause is defending against a real workspace above the tempdir on some machine, but as written the test does not assert the property in its name.

## Confirmed working

* Your claim that the two regression tests fail without the fix holds. Under the mutation you described (reverting the body to `env!("CARGO_MANIFEST_DIR")`) both do fail, verified by building the branch and running them.
* The two `build.rs` call sites provably cannot disagree: both receive the same `work_dir` at `build.rs:486-487` and `770-771`, and `resolve_target_dir` is a pure function of `(env, crate_dir)`.
* Net panic reduction from `.expect` to `Option`. No newly reachable unwrap.

## Two things worth a line in the description

* Several test fixtures (`test-app-1`, `test-contract-1/2`, `test-delegate-*`, `test-contract-metering`) are each their own nested workspace, so their fdev-built artefacts now land in per-fixture `tests/test-*/target/` instead of the shared `<repo>/target/`. Self-consistent, but it means more disk and less cache reuse, and the description only makes the general claim.
* `content.contains("[workspace]")` is unchanged logic, but this PR widens its exposure considerably: it previously only ever walked ancestors of fdev's own fixed compile path, and now walks ancestors of arbitrary user crates. It is sound against `[workspace.metadata]`, `[workspace.dependencies]` and `workspace = true`, though it will still match the string inside a comment.

None of this blocks. A rebase plus the two test tweaks and this is ready from my side.

[AI-assisted - Claude]

sanity:
Merging this. Apologies for leaving it sitting after the review on 8 Aug with no follow-up; three weeks of silence on a correct fix is on us, not on you.

## Confirming the fix, and its blast radius

I re-verified the two findings from the earlier review against the code rather than taking them on trust, and both still stand as written (details below). Neither is blocking, so they should not hold the fix.

I also checked the one thing that would make this risky, which is the deliberate behaviour change: a from-source `fdev` used to write a user's contract artefacts into freenet-core's own `target/`, and now writes into the user's. Nothing in this repo depends on the old placement:

- `crates/fdev/src/util.rs::get_workspace_target_dir` has exactly two callers, both in `crates/fdev/src/build.rs` (`:111` and `:161`), and this PR changes both. There is no third consumer.
- `crates/core/src/util/workspace.rs` carries its own separate copy for core's test utilities. You were right to scope it out, and it is genuinely unreachable from here.
- Every workflow that builds contracts sets `CARGO_TARGET_DIR` explicitly (`ci.yml:744`, `:927`, `:999`, `playwright-shell.yml:56`, `benchmarks*.yml`, `simulation-nightly.yml:49`), as does `apps/freenet-ping/run-ping.mk:92`. All of those take the env short-circuit, so CI behaviour is unchanged by construction.

So the entire reachable surface of this change is a user running `fdev build` with `CARGO_TARGET_DIR` unset, which today either panics with exit 101 on a released binary or silently writes into the wrong tree from source. Both faces are the bug.

The diagnosis in the description is also worth calling out as correct in a way that is easy to get wrong: `env!("CARGO_MANIFEST_DIR")` expanding at *fdev's* compile time is exactly why this is invisible while developing on the repo and fatal on the binaries the install docs point at. The note in `freenet-agent-skills` recorded a different cause, and yours is the right one.

## The CI failure is not yours

`NAT Validation` failed on `de76ad6c` with a peer-connectivity timeout during network setup. The NAT/ping tests never call fdev (`apps/freenet-ping/app/tests/common/mod.rs:515` carries its own private `find_workspace_root`), so nothing in `crates/fdev/` can reach them. The branch is based on `77b3dad3` from 3 Aug, which pins `freenet-test-network 0.1.23`; `main` moved to `0.1.24` in #5220 on 7 Aug. The merge queue rebases onto current `main` and re-runs the required checks before merging, so it will pick that up. If it does not go green on the rebased head the queue will eject rather than merge, and I will come back to it.

## The two test findings, re-verified

Not asking you to act on these here; I am filing them as a follow-up so they are not lost.

1. **The no-workspace fallback branch is still never exercised.** In `a_crate_in_no_workspace_gets_its_own_target_dir_and_does_not_panic` the fixture writes `[workspace]` into `solo`'s own manifest. `Path::ancestors()` yields the path itself first, so `find_workspace_root_from(&solo)` matches on `solo` and returns `Some(solo)`, and `resolve_target_dir`'s `.unwrap_or_else(|| crate_dir.to_path_buf())` never runs in any of the four tests. Both routes produce `solo/target`, so the assertion cannot tell "found itself as the root" from "found nothing and fell back". Replacing that fallback with `panic!()` leaves the suite green. A fixture whose manifest has no `[workspace]` at all, under a temp dir with no `Cargo.toml` above it, would reach the branch the test name describes.

2. **`find_workspace_root_from_returns_none_when_there_is_no_workspace` cannot fail.** No `Cargo.toml` is written anywhere under `tmp`, so a match rooted inside `tmp` is unreachable by construction, and the `|| !found...starts_with(tmp.path())` clause tolerates any `Some(p)` from outside it. The only value that would fail the assertion cannot occur. The OR clause is defending against a stray `Cargo.toml` above the temp dir on the test machine, which is a real hazard; pointing `find_workspace_root_from` at a path under a directory you control end to end is the cleaner way to get that.

Worth saying plainly that the trap you documented is the more valuable half of this PR: a first version of these tests read `CARGO_TARGET_DIR` from the real environment, `cargo test` sets it for its children, so every test took the skip and passed against the unfixed code. Catching that yourself and restructuring the function to take the environment as an argument is why the two tests that matter actually fail without the fix.

Thanks for this, and for the patience.

[AI-assisted - Claude]

sanity:
Update: I hit "Update branch" on this rather than asking you for another round, so it is now merged up to current `main` at `019a8cd3` and CI is re-running on that head. That is an additive merge commit, not a force-push, so nothing of yours was rewritten. Auto-merge is armed, so once the checks are green on `019a8cd3` it goes into the merge queue on its own. If NAT still fails there, the stale-`freenet-test-network` theory was wrong and I will dig in properly rather than leaving it on you.

[AI-assisted - Claude]

Issue #3332, opened 2026-02-28
fix: node WebSocket API returns 'node not available' — event loop stuck, router has 0 success events


iduartgomez:
## Problem

The local Freenet node's WebSocket API is rejecting all client connections with `client error: node not available`, making it impossible to connect any application (River, riverctl, etc.) to the node.

### User Impact
- River application cannot connect at all
- `riverctl debug websocket` fails: `Failed to send test request: unhandled error: client error: comm channel between client/host closed`
- `riverctl invite accept <code>` fails: `Failed to send GET request: unhandled error: client error: comm channel between client/host closed`
- All WebSocket clients are rejected

### Environment
- Freenet v0.1.153 (23c70a9dfb65), built 2026-02-27T06:22:33Z
- macOS Darwin 25.3.0
- riverctl v0.1.38 (also tested with v0.1.22)

### Symptom 1: WebSocket "node not available" errors (ongoing)
Error log is flooded with:
```
ERROR freenet::client_events::websocket: WebSocket protocol error: client error: node not available
```
321 occurrences in the 09:00 hour alone, every 1-3 seconds.

The error originates from `ErrorKind::NodeUnavailable` in `crates/core/src/client_events/websocket.rs:781` — the internal `request_sender` channel to the node's event loop is failing (either closed or the receiver side isn't consuming).

### Symptom 2: Router has ZERO success events
```
router_snapshot failure_events=2954 success_events=0 prediction_active=true consider_n_closest_peers=5
```
The failure count has been climbing steadily (~10 per 5 minutes) since at least Feb 26 17:00, but success_events has **never been non-zero** in any log file. This means the router's prediction model is operating purely on failure data.

### Symptom 3: Subscribe operations continuously timing out
Every ~30 seconds:
```
Transaction timed out tx=... tx_type=Subscribe elapsed_ms=~61000 ttl_ms=60000
Reported operation timeout as routing failure peer=Some(73.198.74.90:60404) contract_location=0.636997145516876
```
All subscribe attempts to this specific contract/peer are failing, feeding the router failure counter.

### Symptom 4: Notification channel was full (earlier logs)
The earliest logs (Feb 26 18:12) show the event loop was stuck:
```
ERROR freenet::node::op_state_manager: notify_op_change: Notification channel full for too long, event loop may be stuck
  tx=... timeout_secs=30 channel_pending=2048 channel_remaining=0
```
```
[...]

Issue #3305, opened 2026-02-27
fix: missing delegates queried at high frequency causing WebSocket "node not available" flood


iduartgomez:
## Problem

When a client requests execution of delegates that are not present in the local node's store, the node returns errors but the client retries at very high frequency (~3 Hz). This floods both the error log and the WebSocket layer with errors, ultimately causing all WebSocket clients to see \"node not available\".

## Observed Behavior

In production logs (2026-02-27 09:59):

```
ERROR contract_handling: freenet::contract::executor::runtime: Failed executing delegate delegate_key=2njmy16nWGcC12ehugTN2m66tM2kFhQXKQUL14KeYHob error=delegate ... not found in store
ERROR contract_handling: freenet::contract: Failed executing delegate request ... error=missing delegate ...
```

This repeats for **two delegate keys** at ~3 Hz for over 10 seconds (50+ error pairs in the log), before:

```
WARN client_event_handling: freenet::client_events::combinator: Client transport disconnected, cleaning up client slot proxy_idx=1
```

During and after this period, all WebSocket responses return:
```
ERROR freenet::client_events::websocket: WebSocket protocol error: client error: node not available
```

at a rate of ~3–4 per second continuously.

## Root Cause

Two issues compound:

1. **No rate limiting on delegate-not-found errors**: When a delegate is missing, there's no backoff or circuit-breaker on the client retry path. The client retries immediately on every failure.

2. **\"node not available\" is too broad**: A delegate execution failure for one contract should not make the entire node appear unavailable to all WebSocket clients.

## Expected Behavior

1. Delegate-not-found errors should be returned to the requesting client as a proper error response (not \"node not available\"), allowing other clients to continue operating normally
2. The node should optionally attempt to fetch missing delegates from the network before returning an error, similar to how contracts are fetched on demand
3. If a client is hammering with failing requests, the server should apply backpressure (rate limit or disconnect after N consecutive failures)

## Affected Components

- `crates/core/src/contract/executor/runtime.rs` — delegate lookup
- `crates/core/src/client_events/websocket.rs` — error propagation
- `crates/core/src/client_events/combinator.rs` — client slot management

PR #5146, opened 2026-08-04
fix(ws): stop reporting a throttled delegate request as a missing delegate

sanity:
> Requested for today's release — it makes a live third-party breakage diagnosable.

## The problem

`DelegateError::Missing` carries two incompatible meanings, and a client cannot tell them apart:

| Where | Meaning |
|---|---|
| `executor.rs` — real `DelegateNotFound` | this delegate is **not registered** on this node |
| `websocket.rs` — synthesised on backoff (#3305, #3332) | you are being **throttled** after repeated failures |

Those call for opposite responses. *Not registered* means the client is built against a delegate key that no longer exists — the app needs updating and its developer needs to hear about it. *Throttled* is transient and wants a retry.

## Why this is not hypothetical

The ghostkeys delegate was re-keyed, and every third-party integration silently broke. An app's delegate key is a build-time constant; user secrets migrate forward automatically, but an app's *reference* does not. After a re-key it addresses an empty namespace.

The first anyone knew was a user reporting their Ghost Key **"works in the vault but says no key found"** elsewhere — the app had told them to go and buy a Ghost Key they already owned.

With `Missing` overloaded, **no amount of care in the app could have produced a better message.** Background: [freenet/ghostkeys#21](https://github.com/freenet/ghostkeys/issues/21).

## The change

One call site. The backoff rejection now returns `ExecutionError` naming the cause and the retry delay, so `Missing` means only what it says.

`ExecutionError` carries no key — fine here, since the client sent the request and knows which delegate. It also means `delegate_error_key` returns `None` for it, which is *correct*: a throttle rejection isn't a per-delegate failure to record.

**Verified this cannot affect the rate limiter.** The backoff branch returns immediately (`return Ok(Some(Message::Binary(...)))`) and never reaches the failure-recording call at the node-response site. I checked this specifically because breaking the limiter would reintroduce #3332.

## Why not a dedicated variant

`RateLimited { key, retry_after }` would be better, but `DelegateError` lives in `freenet-stdlib` — that needs a stdlib release plus a version bump here, too long a chain for today. This fix needs neither and doesn't foreclose the better one.

Fully additive for clients: `DelegateError` is `#[non_exhaustive]`, and nothing branched on `Missing` for the throttle case because nothing could.

## Tests

Two, pinning both directions:

- a throttle rejection must **not** be `Missing`, must be `ExecutionError`, must name its cause, and must yield no key to the failure tracker
- an unregistered delegate must **still** report `Missing` carrying its key

`cargo test -p freenet --lib client_events::websocket::tests::` — **71 passed**, including the existing rate-limiter suite unchanged. `cargo fmt --check` and `cargo clippy -p freenet --lib` clean.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events/websocket.rs
+++ b/crates/core/src/client_events/websocket.rs
@@ -3473,6 +3473,57 @@ mod tests {
         assert!(delegate_error_key(&err).is_none());
     }
 
+    /// Pin the WIRING, not just the shape.
+    ///
+    /// The tests above construct the error by hand, so they prove what a
+    /// throttle rejection should look like but not that the backoff branch
+    /// actually builds one. A future edit could reinstate
+    /// `DelegateError::Missing` there and they would all still pass.
+    ///
+    /// Source-scrape instead (same idiom as `server.rs` and
+    /// `ring/update_rate_limit.rs`): the backoff branch must not name
+    /// `Missing`. Anchored on `check_backoff`, which is the branch's own
+    /// condition, so this survives the surrounding code being reorganised.
+    #[test]
+    fn backoff_branch_does_not_construct_a_missing_delegate_error() {
+        const SRC: &str = include_str!("websocket.rs");
+
+        let (body, _) = SRC
+            .split_once("\nmod tests {")
+            .expect("websocket.rs should have a tests module");
+
+        let idx = body
+            .find("rate_limiter.check_backoff(key_bytes)")
+            .expect("the delegate backoff branch should still exist");
+
+        // The branch is short; take a generous window rather than trying to
+        // brace-match.
+        let window = &body[idx..(idx + 2000).min(body.len())];
+
+        // Strip line comments before asserting. The branch carries a comment
+        // explaining why it does NOT use `Missing`, and a naive scrape cannot
+        // tell that apart from code that does -- which is exactly what this
+        // test caught on its first run.
+        let branch: String = window
+            .lines()
+            .filter(|l| !l.trim_start().starts_with("//"))
+            .collect::<Vec<_>>()
+            .join("\n");
+
+        assert!(
+            !branch.contains("DelegateError::Missing"),
+            "the backoff branch must not report a throttled request as a \
+             missing delegate -- `Missing` means the delegate is not \
+             registered, and conflating the two leaves clients unable to \
+             tell an app that needs updating from one that should retry \
+             (freenet/ghostkeys#21)"
+        );
+        assert!(
+            branch.contains("DelegateError::ExecutionError"),
+            "the backoff branch should report ExecutionError naming the cause"
+        );
+    }
+
     /// The counterpart: a genuinely unregistered delegate still reports
     /// `Missing`, and still carries its key for failure tracking.
     #[test]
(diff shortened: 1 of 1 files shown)


sanity (commented):
## Review: #5146

**Tier: Full.** Changes which error variant reaches every websocket client — a client-facing API surface, and one where the failure mode is silent misdiagnosis rather than a crash.

**How it ran:** diverse-Claude-lens substitute (external models are opt-in per `multi-model-review.md`), as distinct serial adversarial passes: *code-first*, *skeptical*, *testing*, *cross-repo consumer impact*, and *wire compatibility*.

### Findings

**1. The tests didn't pin the wiring** — addressed in `10b53eb`.

The original two construct the error by hand. They prove the *shape* of a throttle rejection but not that the backoff branch builds one; reinstating `DelegateError::Missing` there would leave them all green. Added a source-scrape pin anchored on `check_backoff`, matching the idiom already in `server.rs` and `ring/update_rate_limit.rs`.

It strips line comments first — the branch carries a comment explaining why it does *not* use `Missing`, and the first run of the test flagged that comment as if it were code. Worth noting the mechanism caught a real ambiguity immediately.

Mutation-verified: reverting the branch fails it, restoring passes.

### Cross-repo consumer impact — checked, and better than expected

Grepped every client for a branch on `DelegateError::Missing`. River: none. fdev: none. **The ghostkeys vault: yes**, mapping it to `DelegateCallError::NotRegistered`.

Tracing what that drives turned up a latent bug this fix also closes:

| | throttled probe → |
|---|---|
| **before** | `Missing` → `NotRegistered` → `ProbeVerdict::Skipped` |
| **after** | `ExecutionError` → `ProbeVerdict::AnsweredWithError` |

`Skipped` is counted as neither an error nor undetermined. So during a vault migration sweep, a legacy-delegate probe that merely hit the backoff throttle was **silently indistinguishable from a delegate that does not exist** — and the sweep moved on. Keys under that delegate could be missed with no trace.

The vault's own doc comment cites this websocket synthesis as reason #1 that `NotRegistered` is not proof of absence. After this lands, that reason is gone (reason #2, `UnregisterDelegate` leaving secrets behind, remains — so the vault's caution is still correct, just for one reason instead of two). I'll update that comment in ghostkeys.

### Wire compatibility — no risk

No variant is added; an existing one is emitted. Old clients decode `ExecutionError` today. `DelegateError` is `#[non_exhaustive]` regardless, so every client already has a catch-all. This is strictly safer than the `RateLimited` variant it defers.

### Verified rather than assumed

- **The rate limiter cannot break.** The backoff branch returns immediately and never reaches the failure-recording call at the node-response site — checked specifically, since breaking the limiter would reintroduce #3332.
- `delegate_error_key` returning `None` for `ExecutionError` is correct, not a regression: a throttle rejection is not a per-delegate failure to record.
- `remaining` is still consumed (now in the message as well as the log), so no unused-variable drift.

### Verification on this HEAD

- `cargo test -p freenet --lib client_events::websocket::tests::` — **72 passed**, including the full pre-existing rate-limiter suite unchanged
- `cargo fmt --all --check` clean; `cargo clippy -p freenet --lib` zero warnings

### Verdict

**Ready to merge.** The change is one expression; the case for it is that `Missing` was load-bearing for two incompatible meanings and clients had no way to separate them — demonstrated by a real user being told to buy a Ghost Key they already owned.

[AI-assisted - Claude]

Issue #5161, opened 2026-08-04
Nodes never learn a gateway's protocol version, so every version-gated feature fails closed on gateway links


sanity:
## Problem

**A node never learns the protocol version of any gateway it connects through, no matter how new that gateway is.** As a result it permanently treats the newest, highest-degree nodes on the network as "unknown version" and falls back to legacy wire formats when talking to them.

The version travels in the asymmetrically-encrypted intro packet (`transport/connection_handler.rs:2127`, which packs `PROTOC_VERSION` alongside the symmetric key). Only the **receiving** side of an intro learns the sender's version — recorded at `:1831` on the inbound path and `:2060` on the outbound one.

- **Peer to peer:** NAT traversal means both sides send intro packets, so both learn. Version known in both directions.
- **Joiner to gateway:** the joiner sends its intro, so the gateway learns the joiner's version. The gateway does **not** send an intro back — it replies with `AckConnection { result: Ok(OutboundConnection { key, remote_addr }) }`, and that struct (`transport/symmetric_message.rs:261`) has two fields, neither of them a version.

So `connection_handler.rs:2313` hardcodes it, with the comment stating the cause outright:

```rust
// AckConnection (joiner->gateway) payload carries no protocol version.
remote_protoc_version: None,
```

### It is worse than passive: reconnecting erases what we knew

`ConnectionManager::record_remote_version` (`ring/connection_manager.rs:1384`) writes the `Option` **through**:

```rust
match version {
    Some(v) => { self.remote_version.write().insert(addr, v); }
    None    => { self.remote_version.write().remove(&addr); }
}
```

A reconnect through a gateway therefore actively deletes a version we had previously learned for that address.

## Consequences

Every version-gated wire feature fails closed on these links. Currently that means `supports_hash_first_summaries` and `supports_summary_first_put`, and it will mean every future one by default.

Concretely for the summary exchange (#5155, #5153): the gate declining sends the **uncapped** legacy `Summaries` message — one full summary per shared contract, measured at ~266 contracts x 16,675 B, with a largest observed single message of **1.26 MB**. Gateways are the highest-degree nodes on the network, so these are exactly the links carrying the most anti-entropy traffic.

**The expense is one-directional, which is a useful diagnostic.** The gate keys on the peer's recorded version at whichever node is emitting (`node.rs:2787`). On a joiner-gateway link the gateway holds the joiner's version and replies with cheap digests, while the joiner holds `None` and replies with full bytes. So:

[...]

Issue #5156, opened 2026-08-04
obs: split the hash-first version-gate refusal into unknown-version vs pre-floor


sanity:
## Problem

`version_supports_hash_first_summaries` fails closed (`p2p_protoc.rs:1054`), dropping to the full-bytes summary fallback. That fallback is where 85.3% of the summary arm's bytes are going (companion issue). But **we cannot tell which of two causes is responsible, and they have opposite implications.**

1. **Pre-floor peers.** 2.8% of nodes are below the `(0,2,116)` floor. At a measured 67.9 peers/node that is ~1.9 links each. **This self-heals as the fleet upgrades.**
2. **Unknown remote version.** `PeerConnection::remote_version()` is documented "*None on joiner->gateway connections*" (`peer_connection.rs:1575`), and the `AckConnection` path constructs `remote_protoc_version: None` (`connection_handler.rs:2312-2314`). Each node has ~1-2 gateway links of 67.9. **This never self-heals.**

Both predict the same order of magnitude (~2% of links), both produce the same observable signature, and no current telemetry separates them. If it is cause 2, waiting for fleet upgrade fixes nothing and the gate itself needs repair.

## Fix

Split the `false` decision into two counters — refused-by-unknown-version and refused-by-pre-floor — and export them on the existing rollup. Apply the same split to `supports_summary_first_put`, which has identical structure and the same failure mode.

Follow the existing counter discipline in this codebase (see the "Metric describing a filtering decision, re-derived at the call site" row in `.claude/rules/bug-prevention-patterns.md`): **the gate itself must emit the counts**, not a caller reconstructing them by subtraction. A subtraction across two collections silently absorbs every other filter applied between them and keeps reporting a plausible non-zero value after the filter it claims to measure is deleted. Test shape that catches it: delete the filter and assert the count goes to exactly 0.

## Why this is a prerequisite, not a nice-to-have

It decides whether several downstream items in #5153 are worth building. If cause 1 dominates, the gap closes on its own and effort should go elsewhere; if cause 2 dominates, gateway links are permanently degraded and that is a bug with its own fix. One release of data settles it.

It also generalises: any future digest-first mechanism in this codebase (the NeighborHosting digest is already queued) needs the same instrumentation from day one, because #4965 shows a fail-closed gate plus an unbounded fallback silently reverts to the old behaviour without any signal that it has.

## Testing

Unit tests asserting each counter increments for its own cause and not the other, plus the delete-the-filter mutation check described above.

[...]

PR #5164, opened 2026-08-04
perf: bound the full-bytes summary fallback (#5155)

### LINKED ISSUE #5155: perf: bound the full-bytes summary fallback — 85% of the summary arm bypasses hash-first (~9% of egress)
## Problem

Hash-first summaries (#4965) is deployed and firing, but **85.3% of the InterestSync summary arm is still full summary bytes** — because the fallback taken when the version gate declines is unbounded.

The summary arm is 23.7% of all outbound bytes. Within it:

| sub-arm | GB | msgs | B/msg | entries/msg | B/entry |
|---|---:|---:|---:|---:|---:|
| interests_reply | 44.03 | 1,158,785 | 37,998 | 265.91 | 143 |
| notification | 39.51 | 2,369,170 | 16,675 | 1.00 | 16,675 |
| request_reply (digest path) | 3.18 | 269,245 | 11,806 | 1.31 | 9,019 |
| request_leg (digest path) | 0.007 | 271,240 | 25 | 1.31 | 19 |

A `SummaryDigestEntry` is **21 bytes** by construction (4 B hash + 1 B Option tag + 16 B truncated BLAKE3 — `message.rs:666-690`, `ring/interest.rs:776`). Bucketing windows by `max_bytes` (a single-message quantity that within-window averaging cannot blur) separates the populations cleanly: windows that are pure-digest measure **20-22 B/entry against the predicted 21**, validating the model — and they are only **1.7% of the arm's bytes**. Windows whose largest single message exceeds 200 KB hold 75% of the bytes, and cannot be digest-form at any entry count the fleet produces (4096 cap x 21 B = 86 KB ceiling).

Digest floor 6.76 GB vs 46.01 GB observed.

## Cause

`get_matching_contracts` (`ring/interest.rs:2468-2480`) returns one **full** summary per shared contract with **no send-side cap**. The 4096 cap at `node.rs:2322` is receive-side only. The largest single message observed is 1.26 MB.

The gate (`version_supports_hash_first_summaries`, `p2p_protoc.rs:1054`) fails closed, dropping to that unbounded fallback whenever the remote version is pre-floor **or unknown**. At a measured 265.91 shared contracts x 16,675 B per summary, ~2.8% of peers below the floor plausibly account for ~96% of the arm's remaining bytes: 1.9 peers x 266 x 16,675 B ~= 8.4 MB per heartbeat, against 66 digest replies x 266 x 21 B ~= 0.37 MB.

## Fix

Bound the full-bytes fallback: a rotating, bounded subset of entries per heartbeat rather than the whole shared set. Convergence is unaffected — the rotation covers the full set over several cycles, and this is a backstop layer with a 5-minute cadence, not a delivery path. `emit_stale_peer_syncs` is the in-tree precedent for this shape.

This is safe regardless of which cause dominates (pre-floor peers, which self-heal as the fleet upgrades, or unknown-version links, which do not), which is why it should land before the diagnostic in the companion issue resolves that question.

## Impact

Recovers up to **37.6 GB — 39.5% of the arm, ~9.4% of all outbound bytes.** The single largest safe byte win currently identified, with no wire-format change.

## Testing

A test asserting the fallback reply is bounded regardless of shared-contract count, and a convergence test confirming the rotation still covers the full set within a bounded number of cycles.

Part of #5153.

[AI-assisted - Claude]

sanity:
## Problem

The hash-first summary exchange (#4965) picks a wire form per peer: 21-byte digests for peers at or above 0.2.116, full summary bytes otherwise. The fallback had **no send-side cap** — `get_matching_contracts` (`ring/interest.rs`) returns one complete summary per shared contract, and every one of them goes out in a single message. The 4096 cap at `node.rs` is receive-side only.

At the measured 265.91 shared contracts and 16,675 B per summary that is a multi-megabyte message every 5 minutes per such peer. The largest single message observed in the field is **1.26 MB** — a minority of links that cannot negotiate digests costs more than every link that can, combined. The summary arm is 23.7% of all outbound bytes.

To be precise about the share, since an earlier draft of this description overstated it: 85.3% of the arm's bytes are full summaries, but roughly half of that is the `Notification` leg (39.51 GB), which is single-entry by construction and never consults the version gate at all — `operations/update.rs` calls `full_summaries_message` directly, by the explicit #4965 review decision. This PR touches the unbounded fallback only, which is the **~39.5% of the arm** #5155 attributes to it, or ~9% of all outbound bytes.

Worth noting for reviewers, because it changes how you should weigh this PR: the gate declines on unknown version as well as on old version, and unknown is not rare. The joiner-to-gateway handshake carries no protocol version at all (`transport/connection_handler.rs` hardcodes `remote_protoc_version: None` on the `AckConnection` path), so every connection a node opens through a gateway reports an unknown version no matter how new that gateway is. Which of the two causes dominates is not yet measured — that is #5156 — and repairing the handshake is #5161. **This PR is the guard that makes the gate's failure cheap, not the repair.**

## Solution

Bound the fallback to a rotating window: at most **64 entries** and **9 KiB of summary bytes** per reply, resuming where the previous reply to that peer stopped, so a stable shared set is still covered in `ceil(n / 64)` rounds. `emit_stale_peer_syncs` is the in-tree precedent for the shape.

64 comes from tying the fallback's cost to the digest message it stands in for: a digest reply over the same set is 265.91 × 21 B ≈ 5.6 KB and a full-bytes entry averages 143 B, so ~39 entries is break-even; 64 spends ~9.2 KB to halve the rotation period.

Three decisions worth reviewing on their own terms:

**The byte budget is not redundant with the entry cap.** Summaries are not uniform — a contract we do not host contributes no bytes, a River room contributes ~16.7 KB — so 64 entries can still be ~1.07 MB, which is the message class this PR exists to remove, reproduced under a cap that looks like it forbids it. The 143 B/entry the entry cap derives from is a fleet *average*; it describes the typical mix, and the worst case is where the bytes are. The budget is checked before each entry and never blocks the first, so a reply is at most `budget + one summary` and an oversized summary cannot stall the rotation behind it.

**The cursor stores the last contract id SENT, not an index.** An index names a position, and a removal below it shifts every later contract down by one, so the resume steps over one. Wrapping means that contract is revisited eventually, so this is *not* permanent starvation — it is the loss of the `ceil(n / limit)` bound, which is the entire basis on which bounding the reply is defensible. `rotation_does_not_lose_the_coverage_bound_when_contracts_are_removed` runs both designs over one removal schedule and asserts the index one misses contracts the key one covers.

**Cycle boundaries start at a random offset, not at 0.** Mid-cycle the cursor resumes deterministically; at a boundary — no cursor, cursor evicted, or the previous window ending on the highest id — the next cycle begins at a random offset. A fixed restart at 0 is a tail-starvation failure this codebase has already rejected twice for the same reason, and it is reachable here two ways, neither needing an attacker: the cursor is in-memory and keyed by address, so it is lost on our restart, on eviction, and whenever a peer reconnects from a new port; and `sorted` is the intersection with the hash list the *peer* advertised, so alternating a full `Interests` with a single high-id one parks the cursor at the end and forces a restart every other round. A reviewer demonstrated 7 of 12 contracts never advertised over 50 cycles under that pattern. Randomising the boundary costs nothing — a cycle from any offset still covers the set in `ceil(n / limit)` rounds because the window wraps — and removes the fixed prefix.

**The version gate is read once.** The size decision and the encoding decision both need it, and two reads open a window where a reconnect changes the answer between them. The dangerous direction is known → unknown (`record_remote_version` writes `None` through, clearing the entry), which would build the full entry set believing it was headed for digests and then encode it as full bytes — the unbounded message, reassembled from two correct-looking halves. `summary_reply_form` resolves it once and the form is threaded through.

**Not changed:** digest-capable peers keep receiving the whole shared set every round. The `SummaryRequest` reply leg is untouched — it is demand-gated (it answers an explicit request for bytes, and replying with digests would loop the exchange). Note "demand-gated", not "bounded": its only hard limit is `MAX_SUMMARY_HASHES_PER_MESSAGE`, and only a digest-capable peer can reach it. `ChangeInterestsReply` is untouched: it is event-driven rather than periodic, so there is no guaranteed next round to rotate into, and a bound there could skip a newly-added interest indefinitely. Interest *registration* also stays unbounded — the rotation bounds what we advertise, never who counts as a broadcast target, since rotating that would convert a bandwidth bound into missed updates.

## Cost, stated honestly — and corrected

An earlier draft of this description put the cost at "~25 minutes for a typical 266-contract set", from `ceil(266 / 64)` heartbeats. **That number is wrong for the population this change actually affects, by roughly an order of magnitude**, and the review caught it.

`ceil(n / 64)` holds only when the entry cap is what binds. On a real fallback link it never is. The 143 B/entry the entry cap is derived from is a fleet average dominated by 21-byte digest entries and by entries for contracts the sender does not host, which carry no summary at all. A hosted River room carries ~16.7 KB — larger than the whole 9 KiB budget — so the budget binds after the first hosted contract in a round. The honest cycle length is:

```
rounds per cycle ~= total summary bytes for the shared set / 9 KiB
```

For the heaviest links, the ~1.16 MB reply behind the observed 1.26 MB maximum, that is ~130 rounds at a 5-minute heartbeat: **on the order of ten hours**, not 25 minutes. Worst case, every summary over budget, it degenerates to one contract per round.

So the real trade is: on links we cannot negotiate digests with, noticing a **pre-existing** divergence on a contract **nobody touches** can take hours instead of 5 minutes. The event-driven paths — push-on-update, request-full-state-when-too-far-behind, resend-on-failed-patch — are untouched and still bound detection for any contract anybody actually touches, and the peer's own summaries still reach us unbounded every round so *its* staleness is still detected immediately. Only the "we are behind and they must tell us" direction is delayed, on ~2.8% of links.

I think that is still the right trade, because the alternative on those same links is a multi-megabyte message every 5 minutes and there is no setting of these constants that is both cheap and fast. But it is an argument for repairing the version gate (#5156 / #5161) rather than tuning here, and it should not be sold as minutes. `heavy_summaries_make_the_cycle_far_longer_than_the_entry_cap_suggests` pins the long cycle as a test so the comfortable number cannot quietly return.

One second-order cost the review surfaced, not otherwise obvious: the peer-summary cache the broadcast path diffs against is populated from this exchange on fallback links (`DigestAgreement` never fires there), so it warms more slowly, and a `(contract, peer)` pair with no cached summary is what ships full state instead of a delta. `Delivery` covers actively-updated contracts and #5153 finds 57.5% of cold-start sends fire before any heartbeat could have helped, so the marginal cost is probably small — but it is a bytes-for-bytes trade on the same links, not purely a latency one.

## Testing

Six new behavioural tests at the handler level and six at the rotation core, all mutation-tested (they fail against the un-bounded and un-rotated versions of the code, verified by reverting each mechanism in turn):

- `fallback_reply_is_bounded_at_the_widest_observed_shared_set` — 2,448 shared contracts, the widest case seen in the field.
- `fallback_byte_budget_binds_before_the_entry_cap` — 5 KB summaries; the entry cap alone would not bound this.
- `oversized_summary_still_sends_and_the_rotation_advances` — a summary larger than the whole budget; asserts progress rather than a stall.
- `digest_peer_still_receives_every_shared_contract` — the no-change property for the 98%.
- `fallback_rotation_covers_the_whole_shared_set` — full coverage within `ceil(n / cap)` rounds.
- `bounding_the_reply_does_not_bound_interest_registration` — peer interest registered for contracts this round did not advertise.
- Rotation core: bounds and wrapping, coverage for n ∈ {1, 5, 63, 64, 65, 128, 266, 2448}, insertion mid-cycle, removal mid-cycle (with the index-cursor control), a removed cursor id, and per-peer cursor isolation.

Plus, from review: `heavy_summaries_make_the_cycle_far_longer_than_the_entry_cap_suggests` (pins the honest cycle length), `fallback_window_start_randomises_the_cycle_boundary` and a GlobalRng source pin (pin the boundary randomisation), coverage tests sweeping starting offsets over set sizes that do not divide evenly by the limit, and absolute literals in place of assertions written against the constants that produced them.

`cargo fmt`, `cargo clippy --locked -- -D warnings` (the CI invocation), and `cargo test -p freenet` are all clean.

## Review

Five independent blind reviewers (code-first, testing, adversarial, big-picture, and a convergence-specific lens). Consolidated findings and their dispositions are in a comment below; every finding was fixed or has a stated reason. Two were substantive — the cycle-boundary starvation and the order-of-magnitude cost error — and both are addressed above.

Closes #5155

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -4347,11 +4347,12 @@ impl GlobalTestMetrics {
     // === Hash-first summary exchange falsifiers (#4965) ===
     //
     // The claim under test is "the common case stops shipping summary bytes".
-    // These counters are fed from the TWO constructor functions that can build
-    // these messages — `node::summaries_reply_for_peer` (which chooses the
-    // encoding) and `node::full_summaries_message` (which every full-bytes
-    // path routes through, including the `operations::update` helpers, which
-    // are CALLERS rather than construction sites). So
+    // These counters are fed from the constructor functions that can build
+    // these messages — `node::summaries_reply_in_form` (which applies the
+    // chosen encoding, and which `node::summaries_reply_for_peer` delegates
+    // to) and `node::full_summaries_message` (which every full-bytes path
+    // routes through, including the `operations::update` helpers, which are
+    // CALLERS rather than construction sites). So
     // `summary_full_bytes() == 0` means no summary byte was put on the wire by
     // any path, not merely by the one a test happened to exercise.
     // `no_uninstrumented_full_summaries_construction` pins that no production
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -9455,6 +9455,123 @@ mod tests {
             );
         }
 
+        /// Two `Interests` from the same peer handled concurrently cost at
+        /// most one round of rotation progress, and cannot wedge or corrupt it.
+        ///
+        /// The handler reads the cursor, awaits up to 64 summary fetches, then
+        /// writes the cursor back, so two overlapping invocations for one peer
+        /// can both read the same start and build the same window. Inbound
+        /// messages are dispatched one task per message, so this is reachable
+        /// in production and a peer can provoke it deliberately. The comment at
+        /// the cursor-advance site accepts that cost as "one round"; this test
+        /// is that claim's evidence rather than its restatement.
+        ///
+        /// The load-bearing assertion is the progress floor: after `PAIRS`
+        /// concurrent pairs run back to back with no sequential round between
+        /// them, the union of what was advertised must be at least
+        /// `PAIRS * 64` contracts — i.e. every pair advanced the rotation by at
+        /// least one window even in the worst interleaving. Losing MORE than
+        /// one round per pair, or losing the cursor entirely and re-sending the
+        /// same window forever, both fail here. The bound holds for any
+        /// interleaving, so the test does not depend on the scheduler
+        /// reproducing a particular one: a genuinely concurrent run and a
+        /// serialised run both satisfy it, and only a broken rotation does not.
+        ///
+        /// Multi-threaded flavour with `spawn` on purpose: an earlier version
+        /// of this test used `join!` on a current-thread runtime and the two
+        /// futures did NOT overlap — it passed while exercising nothing.
+        #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+        async fn concurrent_interests_from_one_peer_cost_at_most_one_round() {
+            const PAIRS: usize = 3;
+            let h = build_harness("hf-bound-concurrent", 17170, vec![7u8; 64]).await;
+            let keys = host_many(&h, 400);
+            let hashes = distinct_hashes(&keys);
+            let expected: HashSet<u32> = hashes.iter().copied().collect();
+            let ids: HashSet<ContractInstanceId> = keys.iter().map(|k| *k.id()).collect();
+
+            let mut covered: HashSet<u32> = HashSet::new();
+            for pair in 0..PAIRS {
+                let spawn_one = || {
+                    let op_manager = Arc::clone(&h.op_manager);
+                    let hashes = hashes.clone();
+                    let peer = h.old_peer;
+                    tokio::spawn(async move {
+                        handle_interest_sync_message(
+                            &op_manager,
+                            peer,
+                            InterestMessage::Interests { hashes },
+                        )
+                        .await
+                    })
+                };
+                let (first, second) = tokio::join!(spawn_one(), spawn_one());
+
+                for (which, joined) in [("first", first), ("second", second)] {
+                    match joined.expect("handler task panicked") {
+                        Some(InterestMessage::Summaries { entries, .. }) => {
+                            assert!(
+                                entries.len() <= 64,
+                                "pair {pair} {which} reply carried {} entries; the \
+                                 per-reply bound must hold regardless of overlap",
+                                entries.len()
+                            );
+                            assert!(summary_bytes_of(&entries) <= 9 * 1024 + h.our_summary.len());
+                            covered.extend(entries.iter().map(|e| e.hash));
+                        }
+                        other => panic!("pair {pair} {which} reply was {other:?}"),
+                    }
+                }
+
+                // The cursor must still name a contract that exists, not a
+                // value torn between the two writers.
+                let cursor = h
+                    .op_manager
+                    .interest_manager
+                    .peek_fallback_cursor(h.old_peer)
+                    .expect("a fallback reply must leave a cursor");
+                assert!(
+                    ids.contains(&cursor),
+                    "after pair {pair} the cursor is no longer a member of the \
+                     shared set — concurrent writers corrupted the rotation \
+                     rather than merely duplicating a window"
+                );
+            }
+
+            assert!(
+                covered.len() >= PAIRS * 64,
+                "{PAIRS} concurrent pairs advertised only {} distinct contracts; \
+                 each pair must advance the rotation by at least one 64-entry \
+                 window, so the overlap cost more than the one round the \
+                 cursor-advance comment claims",
+                covered.len()
+            );
+
+            // And the rotation still completes from there.
+            for _ in 0..hashes.len().div_ceil(64) {
+                match handle_interest_sync_message(
+                    &h.op_manager,
+                    h.old_peer,
+                    InterestMessage::Interests {
+                        hashes: hashes.clone(),
+                    },
+                )
+                .await
+                {
+                    Some(InterestMessage::Summaries { entries, .. }) => {
+                        covered.extend(entries.iter().map(|e| e.hash));
+                    }
+                    other => panic!("expected full bytes, got {other:?}"),
+                }
+            }
+            assert_eq!(
+                covered.len(),
+                expected.len(),
+                "{} contracts still unadvertised after the concurrent pairs plus \
+                 a full sequential cycle",
+                expected.difference(&covered).count()
+            );
+        }
+
         /// The rotation bounds what we ADVERTISE, never who we consider a
         /// broadcast target.
         ///
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2861,6 +2861,15 @@ const MAX_FALLBACK_SUMMARIES_PER_REPLY: usize = 64;
 /// same links is a multi-megabyte message every 5 minutes; there is no setting
 /// of this constant that is both cheap and fast, which is the argument for
 /// repairing the version gate (#5156 / #5161) rather than tuning here.
+///
+/// # Before retuning either constant, get the field reading
+///
+/// Everything above is a MODEL, derived from fleet averages. That is how the
+/// entry cap came to be sized off a mean of 143 B/entry when the population it
+/// governs is nothing like the mean — the error survived review of the
+/// arithmetic because no counter contradicted it. **#5169** adds
+/// entries-per-reply and bytes-per-reply on this path. Do not move either
+/// constant on the strength of another average; read the distribution.
 const MAX_FALLBACK_SUMMARY_BYTES_PER_REPLY: usize = 9 * 1024;
 
 /// Which wire form a multi-entry `Summaries` reply to a peer takes.
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -2867,9 +2867,13 @@ const MAX_FALLBACK_SUMMARIES_PER_REPLY: usize = 64;
 /// Everything above is a MODEL, derived from fleet averages. That is how the
 /// entry cap came to be sized off a mean of 143 B/entry when the population it
 /// governs is nothing like the mean — the error survived review of the
-/// arithmetic because no counter contradicted it. **#5169** adds
+/// arithmetic because no counter contradicted it. **#5168** adds
 /// entries-per-reply and bytes-per-reply on this path. Do not move either
 /// constant on the strength of another average; read the distribution.
+///
+/// Those counters are also the cleanest read on #5161: as gateway links learn
+/// peer versions and stop taking this path at all, they should fall toward
+/// zero.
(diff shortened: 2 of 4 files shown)


sanity:
## Multi-perspective review (Full tier)

Five independent reviewers, each blind to the others, each with a distinct lens: **code-first** (read the code before the description, flag intent/implementation mismatches), **testing** (vacuous assertions, coverage gaps — this one mutation-tested 11 variants of the production code), **adversarial** (bug hunt, remote-influenced input, concurrency), **big-picture** (does it achieve the goal, are the exclusions sound, what did it make stale), and a fifth aimed at **the failure mode specific to this change** (can a bounded anti-entropy exchange break state convergence).

Findings and dispositions below. Fixed in 4a51f77.

### Substantive — fixed

**1. Cycle boundaries restarted at index 0, which can starve the tail.** Raised independently by the adversarial and convergence lenses. Two reachable paths, neither needing an attacker: the cursor is in-memory and keyed by address, so it is lost on our own restart, on LRU eviction, and whenever a peer reconnects from a new source port — a peer returning to a boundary more often than a cycle completes would only ever hear about the head of its set. And `sorted` is the intersection with the hash list the *peer* advertised, so alternating a full `Interests` with a single high-id one parks the cursor at the end and forces a restart every other round. The adversarial reviewer proved it with a throwaway test over the shipped functions: **7 of 12 contracts never advertised across 50 cycles.**

This is the failure `emit_stale_peer_syncs` and the `SummaryDigests` arm both already rotate to avoid, with the same argument in their comments, and I did not carry it over. Cycle boundaries now draw a random offset from `GlobalRng`. It costs nothing in the ordinary case — a cycle beginning at any offset still covers the set in `ceil(n / limit)` rounds because the window wraps — and it removes the fixed prefix. Mid-cycle resumption is unchanged and still deterministic.

**2. The stated cost was wrong by roughly an order of magnitude.** Raised by three lenses. `ceil(n / 64)` rounds holds only when the entry cap binds, and on a real fallback link it never does: 143 B/entry is a fleet average dominated by digest entries and by contracts the sender does not host, while a hosted River room is ~16.7 KB, larger than the whole 9 KiB budget. The honest cycle length is `total summary bytes / 9 KiB` rounds — on the order of ten hours for the heaviest links, not ~25 minutes. Constants' rustdoc corrected, the entry cap re-documented as the guard for the opposite extreme rather than the governing bound, PR description rewritten, and `heavy_summaries_make_the_cycle_far_longer_than_the_entry_cap_suggests` added so the comfortable number cannot quietly return.

**3. The PR body attributed the whole 85.3% to this fallback.** Roughly half of that is the `Notification` leg, which is single-entry and never consults the version gate. #5155 itself says 39.5%; my description had regressed it. Corrected.

**4. A source-scrape pin I silently weakened.** `summary_request_reply_is_always_full_bytes` banned only the string `summaries_reply_for_peer`; this PR split that into `summary_reply_form` + `summaries_reply_in_form`, neither a substring of the old name, so routing the `SummaryRequest` arm through the split pair would have reintroduced the digests→request→digests loop with the pin still green. Now bans all three.

**5. Four doc sites this PR made wrong.** `summaries_reply_for_peer`'s "both multi-entry legs route through here" (`InterestsReply` no longer does), the gate constant's emission inventory in `p2p_protoc.rs`, the counter provenance in `config.rs`, and the `SummaryRequest` arm's "runs the identical loop" rate-limit argument (the loops are no longer identical; the conclusion survives but now rests on who can reach that arm). All corrected.

### Test strength — fixed

**6. The reply-bound test asserted against the constant that produced it.** Raised by the testing lens, which verified that raising the cap to 100,000 left it green. Now asserts absolute literals, and pins which of the two limits binds in that fixture so a regression to either is caught rather than masked by the other.

**7. Wrapping was pinned only by index literals.** Both coverage tests started from offset 0 and used sizes divisible by the limit, so neither would have failed if the window truncated instead of wrapping. Coverage tests now sweep starting offsets and include sizes that do not divide evenly.

**8. At-capacity cursor-LRU behaviour untested, and the eviction comment overstated its guarantee.** The comment claimed eviction "can never skip" a contract; true per round, false under sustained eviction with the old fixed restart. Both the behaviour and the comment are fixed by finding 1; the comment now says what eviction actually costs.

### Accepted with reasons recorded, not changed

**9. The cursor advances on send attempt, not delivery.** A dropped reply costs those contracts a cycle. Making it delivery-driven needs an ack this exchange does not have. Recorded in a comment.

**10. Two concurrent `Interests` from one peer can duplicate a window,** losing a round of progress. Reserving the window at read time would trade this for a worse failure: a budget-cut round would then skip everything it did not reach. Recorded in a comment.

**11. The cursor is keyed by `SocketAddr` while sibling interest state uses `PeerKey`,** which `.claude/rules/ring.md` warns against. Kept deliberately: the version gate this rotation is subordinate to is itself keyed by `SocketAddr` (`supports_hash_first_summaries`), and keying the two differently would be the real inconsistency. Consequences are bounded — an inherited cursor only changes where in the cycle a reconnected peer resumes, and randomised boundaries make that harmless.

**12. The entry cap is charged against entries that carry no summary,** so on phantom-heavy sets the window fills with filler and real advertisements rotate more slowly than the entry count suggests. Real, but the byte budget is what governs cost, and dropping `None` entries would change behaviour on the receive side (they clear a stale cached summary). Noted rather than redesigned.

### Not adopted, flagged instead

**13. A field counter for entries-per-fallback-reply.** The big-picture lens is right that the cost model here was argued from a fleet average and turned out to be an order of magnitude off, and that #5153's own discipline says to ship a counter with a change like this. I have not added one, because it means touching the telemetry rollup and this PR is already at the edge of "one logical change". Recommending it as an immediate follow-up rather than dropping it silently.

### Verified clean by more than one reviewer

The `partition_point` premise (`get_matching_contracts` really does sort ascending by id bytes, and ids in `matching` are unique); no panics, no overflow, all indices in range; no lock held across an `.await`; the cursor map is bounded; the receiver never treats an absent contract as information, so a subset is a latency trade and not a correctness break; the skipped summary fetch has no side effect the system depends on; interest registration and broadcast targeting still cover every shared contract; the digest path is genuinely untouched; and no test was weakened or deleted.

[AI-assisted - Claude]

Issue #5166, opened 2026-08-04
gateway_version_probe never learned the probed gateway's version (fixed by #5161, verify after rollout)


sanity:
## Problem

`gateway_version_probe` (`crates/core/src/operations/connect.rs:1613`) exists to discover newer releases so a node knows to auto-update. Its doc comment states the mechanism:

> The transport handshake exchanges version info via `report_peer_version()` / `signal_urgent_update()`.

On the path the probe actually takes, that was not true for the gateway it probes.

`report_peer_version` is called from exactly two places, both in intro-packet parsing (`transport/connection_handler.rs`, the inbound `gateway_connection` path and `decrypt_asym` on the outbound NAT-traversal path). The probe calls `start_client_connect` against a gateway, which is the joiner->gateway handshake — the joiner sends an intro and receives an `AckConnection`, and never parses an intro from the gateway. So `report_peer_version` was never reached for the gateway's own version.

## Impact

The probe was not fully inert. The CONNECT it starts is forwarded by the gateway, and the peers that hole-punch back to the node DO send intros, so their versions are reported. But:

- The **gateway's own** version was never learned, and gateways are typically the first nodes on a new release (the release workflow drives `gateway-update.yml` directly). That is the earliest available signal and the probe could not see it.
- A node already at `max_connections`, whose forwarded connects are rejected, established only the gateway link and therefore learned nothing from the probe at all.

Net effect: nodes noticed a release later than the mechanism intended, with the delay worst in exactly the case the probe was written for.

## Status

Confirmed, and **fixed as a side effect of #5161**. That change has the acceptor include its protocol version in the connection ack (`AckConnectionV2`, version-gated on the joiner), and the joiner calls `report_peer_version` on what it reads out of that ack. So after #5161 the probe does learn the gateway's version, and the comment quoted above becomes accurate.

Filed separately rather than folded into that PR, per review scoping.

## What remains

- Verify against production telemetry after 0.2.120 has rolled out that version discovery through a gateway-only probe now fires — the fix is only effective on links where BOTH ends are at or above the 0.2.120 floor, so it ramps with fleet upgrade rather than switching on at once.
- Decide whether the probe needs anything beyond the transport fix. It currently relies entirely on the handshake side effect; there is no explicit assertion anywhere that a probe produced a version observation, so a future regression in the same place would again be silent. A test or counter that pins "a gateway-only probe yields a version observation" would close that.

Found while working #5161.

[AI-assisted - Claude]

[...]

PR #5167, opened 2026-08-04
fix(transport): carry the acceptor's version in the connection ack

### LINKED ISSUE #5161: Nodes never learn a gateway's protocol version, so every version-gated feature fails closed on gateway links

sanity:
## Problem

**A node never learned the protocol version of any gateway it connected through, no matter how new that gateway was.** It therefore treated the newest, highest-degree nodes on the network as unknown-version permanently, and fell back to legacy wire formats when talking to them.

The version rides in the asymmetrically-encrypted intro packet, and only the **receiving** side of an intro learns it. Peer to peer, NAT traversal means both sides send one, so both learn. A gateway does not send an intro back — it replies with `AckConnection { result: Ok(OutboundConnection { key, remote_addr }) }`, a struct with no version field. So `connection_handler.rs` hardcoded `remote_protoc_version: None` for every joiner→gateway link, with a comment stating the cause outright.

Every version-gated wire feature failed closed on those links: today `supports_hash_first_summaries` and `supports_summary_first_put`, and every future one by default. For the summary exchange (#5153, #5155) that means the uncapped legacy `Summaries` message — one full summary per shared contract — on exactly the links carrying the most anti-entropy traffic.

**This is not only a gateway bug.** The joiner's `AckConnection` arm is also reachable peer-to-peer whenever one side of the hole-punch races ahead to the ack before parsing the other's intro. The existing `simulate_nat_traversal` test documented this by asserting only that *at least one* of the two sides captured a version.

## Solution

Add `SymmetricMessagePayload::AckConnectionV2 { connection: OutboundConnectionV2 { key, remote_addr, protoc_version: [u8; 8] } }`, appended **last** so no existing variant index moves, and emit it in place of the legacy success ack when the peer is at or above a version floor.

**Why this needs no bootstrapping round-trip.** The usual chicken-and-egg with a new wire field — you cannot send it until you know the peer decodes it, and you cannot know that without asking — does not arise. The acceptor has already decrypted and parsed the joiner's intro packet, which carries the joiner's version, *before* it builds the ack. The gate reads a version already in hand, and a pre-floor joiner receives byte-identical traffic.

Design choices worth calling out:

- **`ack_ok` is left untouched** rather than generalised to take an `Option<[u8; 8]>`. The pre-floor path must keep producing exactly the bytes it produces today, and calling the same unmodified function is a stronger guarantee than any amount of branch review. A golden-bytes test pins it regardless.
- **A parallel struct, not a new field on `OutboundConnection`.** The payload is bincode, whose layout is not forward-tolerant, so the version can only go to a peer known to decode it — which means it must be omitted conditionally, and bincode cannot express that without a hand-written `Serialize`. An `Option` would still emit its discriminant byte and change what a pre-floor peer sees.
- **Gated at both `ack_ok` send sites**, the gateway path and the peer-to-peer `RemoteInbound` path, so the hole-punch race above is fixed too. Both sites have already parsed the remote's intro, so both know its version.
- **The failure ack has no V2 form.** An incompatible peer is still rejected with the legacy `AckConnection { result: Err(..) }`, which is correct precisely because such a peer may not carry the new index.
- The joiner also calls **`report_peer_version`** on what it reads from the ack. That is what makes the fix reach auto-update discovery, not just the local gate — see #5166.

Floor: `GATEWAY_ACK_VERSION_MIN_VERSION = (0, 2, 120)`, the release this ships in, with a `HASH_FIRST_SHIPPED_IN`-style guard so a slipped or renumbered release cannot leave it stale silently. A test-only override is threaded `NodeConfig → P2pConnManager → create_connection_handler → config_listener → UdpPacketsListener`; unlike the three existing floor overrides this has to reach the transport layer, because the gate decides how a handshake ack is encoded and the handshake runs below the node layer.

### The erasure (`record_remote_version`)

The issue also asked to stop a `None` from deleting a previously known version. **Kept as is, deliberately**, and the reasoning is now recorded in the code.

The write-through is not an oversight: it was added specifically because this path yields `None`, and a stale entry can name a version *newer* than the peer actually runs — a peer restarting on a downgraded binary at a reused address. We would then send it a variant it cannot decode, which closes the connection as `ConnectionError::Serialization` and presents as a transport fault. Relaxing it would re-open exactly that.

What this PR changes is what `None` **means**. Enumerating where one can still arise: the joiner→gateway ack and the peer-to-peer ack race now both yield `Some` for any remote at or above the floor, and the intro-parsing path always did. So a remaining `None` is not "we failed to learn it" — it is the positive fact that the remote is below the floor, and clearing a stale newer entry is the accurate write rather than a lossy one. Both readings fail closed identically at every `version_supports_*` gate.

### Rollout watch item

From 0.2.120, node→gateway links become eligible for the other version-gated features at once, on the highest-degree nodes on the network. Concretely that is `supports_hash_first_summaries` (the intended benefit, #5153) and `supports_summary_first_put`. Both floors are below 0.2.120, so a gateway that sends `AckConnectionV2` provably supports them and the transition is safe by construction — but it is a real change in traffic shape on the busiest links and worth watching during the staggered rollout. SubscribeHint is *not* affected: it is held off by the `PLACEMENT_MIGRATION_ENABLED` compile-time kill switch, independent of any version gate.

### Simulation default is OFF, opt in per sim

`SimNetwork::enable_gateway_ack_version()`. This deviates from `hash_first_summaries_floor_override`, which defaults ON so a version-gated wire change gets integration coverage before its release, and matches the two cascade gates instead.

The gate is encoding-only where it fires, but the version it teaches is the **input to every other `version_supports_*` gate**, so enabling it network-wide makes node→gateway links newly eligible for summary-first PUT probes and hash-first digests. That is the cascade shape, not the encoding shape. Measured rather than assumed: with the default ON, `test_summary_first_put_holder_found_ships_delta` and `test_summary_first_put_reverse_delta_converges_originator` fail deterministically, because the regular node's own PUT starts passing the emission gate.

The cost — the sim suite keeps a topology asymmetry production no longer has, from 0.2.120 onward — is stated in the code and bought back with a dedicated A/B sim test rather than a suite-wide flip. The rustdocs on those two summary-first tests, which asserted the asymmetry was permanent and "not a simulation-harness gap", are corrected either way.

## Testing

Every claim above has a test, and the ones that could pass vacuously were mutation-checked.

- `ack_ok_produces_identical_bytes_to_pre_5161_encoding` — **the compatibility guarantee, asserted on bytes.** A golden literal captured from the pre-change encoding, not a re-serialization of the same struct: the latter would compare the encoder against itself and pass even if the layout moved wholesale. Fails on a field added to `OutboundConnection`, a reordering of `SymmetricMessage`, a bincode config change, or a variant inserted before `AckConnection`.
- `appending_ack_connection_v2_did_not_renumber_existing_variants` — reads variant indices out of the serialized bytes. Byte-identity for the ack alone would be worth little if every other message silently changed meaning on the wire.
- `ack_ok_with_version_round_trips_and_differs_from_legacy` — the gate chooses between genuinely different bytes, not a decorated identical message.
- `joiner_learns_gateway_version_when_at_or_above_the_floor` / `pre_floor_joiner_gets_the_legacy_ack_and_still_connects` — both sides of the gate through the real mock transport. Against **explicit** floors rather than the crate version, so neither flips at the 0.2.120 release. The pre-floor case asserts the handshake still *completes*, not merely that no version was learned. Mutation-checked: forcing `version_supports_ack_version` to always-false makes the first fail.
- `ack_version_gate_is_fail_closed_and_inclusive_at_the_floor` — unknown version must not receive the new encoding; the release exactly at the floor must.
- `ack_version_floor_tracks_the_shipping_release` — `ACK_VERSION_SHIPPED_IN` is `None`, so this fails the moment a release bump lifts `PCK_VERSION` to 0.2.120 without someone consciously deciding.
- `test_joiner_records_gateway_version_through_sim_handshake` — end-to-end through the real simulated handshake, **with a control arm**: the same topology with the gate off, asserting `remote_version == None` and `supports_hash_first_summaries == false`. Without that arm the enabled arm could pass on a harness that populated versions by some other route, and would never have failed against the bug it describes.
- `record_remote_version_none_clears_a_stale_entry` and `record_remote_version_overwrites_downwards_at_a_reused_address` — pin the anti-downgrade behaviour, including that "keep the highest version seen" would be the same bug in a different shape.

Added after the Full-tier review (see the review comment for the full list and the mutations that drove it):

- `nat_traversal_ack_racer_learns_version_from_the_ack` — the peer-to-peer emit site, which had **no coverage at all** until a reviewer probed it with a `panic!` and the whole suite still passed. Forcing the race is the difficulty: with no packet loss both peers learn versions from intros the old way, so the first version of this test was itself vacuous. It now drops peer B's intro so A stays in `StartOutbound` and receives B's ack instead.
- `test_gateway_ack_version_unlocks_node_originated_summary_first_put` — the cascade the simulation default otherwise excludes: the reversed-roles mirror of the existing summary-first test, 1 delta send with the ack gate on and 0 without.
- `duplicate_v2_ack_on_established_connection_is_answered_with_legacy_ack` — pins the deliberate choice to answer a retransmitted V2 ack in the old encoding.
- A third arm on the A/B sim test, because two of its four assertions were algebraically implied by the other two.
- `simulate_gateway_connection`'s assertion de-circularized. As first written it computed its expected value with the predicate under test, and a reviewer's mutation — a gateway emitting V2 to every peer regardless of version — left it green.

Verification: `cargo fmt`; `cargo clippy --locked -- -D warnings` (the CI invocation) clean; `cargo test -p freenet --lib` 4648 passed; and the simulation tests under `cargo nextest run -p freenet --features "simulation_tests,testing" --test simulation_integration`, which the lib run does not reach.

Deliberately **not** touched: the summary fallback itself, which is #5155.

### On the issue's third acceptance criterion

#5161 asks that `record_remote_version(addr, None)` stop erasing a previously recorded version. This PR deliberately does the opposite (argued above) and pins the erasure with tests. `Closes #5161` will therefore auto-close an issue one of whose three stated criteria was answered rather than implemented — flagging it explicitly so that is a decision rather than an accident.

Closes #5161

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -406,6 +406,31 @@ pub struct NodeConfig {
     /// `#[serde(skip)]`; never serialized.
     #[serde(skip)]
     pub(crate) hash_first_summaries_floor_override: Option<(u8, u8, u16)>,
+    /// Test-only override for the version-carrying-ack floor
+    /// (`GATEWAY_ACK_VERSION_MIN_VERSION`, #5161). Threaded down into the
+    /// TRANSPORT layer (`create_connection_handler`), unlike the three
+    /// overrides above, which are read at the `ConnectionManager`: the gate it
+    /// controls decides how a handshake ack is ENCODED, and the handshake runs
+    /// below the node layer entirely.
+    ///
+    /// In production this is `None` → the real `(0, 2, 120)` floor (untouched).
+    ///
+    /// **Simulations default this ON**, like `hash_first_summaries_floor_override`
+    /// and unlike the two cascade gates. Same reasoning: it changes only the
+    /// ENCODING of an ack every sim already exchanges, with identical handshake
+    /// semantics, so defaulting it on costs unrelated sims nothing and buys the
+    /// one thing a version-gated wire change otherwise cannot get — the whole
+    /// simulation suite exercising the new path BEFORE the release that lifts
+    /// the crate version over the floor. It also un-breaks the sim topology
+    /// asymmetry that several existing tests document and work around (a
+    /// regular node could not learn its gateway's version, so gateway-directed
+    /// PUTs had to originate from the gateway).
+    ///
+    /// Not cfg-gated for the same reason as `subscribe_hint_floor_override`:
+    /// `node::testing_impl` sets it and is compiled unconditionally.
+    /// `#[serde(skip)]`; never serialized.
+    #[serde(skip)]
+    pub(crate) ack_version_floor_override: Option<(u8, u8, u16)>,
     /// Test-only harness flag: when set, a startup-hosted contract
     /// (`SeedHostedContract`, i.e. `append_contracts` with `subscription =
     /// true`) is registered in the neighbor-hosting advertised set so the
@@ -573,6 +598,7 @@ impl NodeConfig {
             subscribe_hint_floor_override: None,
             summary_first_put_floor_override: None,
             hash_first_summaries_floor_override: None,
+            ack_version_floor_override: None,
             advertise_seeded_hosts: false,
             hosting_time_source_override: None,
         })
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -415,16 +415,16 @@ pub struct NodeConfig {
     ///
     /// In production this is `None` → the real `(0, 2, 120)` floor (untouched).
     ///
-    /// **Simulations default this ON**, like `hash_first_summaries_floor_override`
-    /// and unlike the two cascade gates. Same reasoning: it changes only the
-    /// ENCODING of an ack every sim already exchanges, with identical handshake
-    /// semantics, so defaulting it on costs unrelated sims nothing and buys the
-    /// one thing a version-gated wire change otherwise cannot get — the whole
-    /// simulation suite exercising the new path BEFORE the release that lifts
-    /// the crate version over the floor. It also un-breaks the sim topology
-    /// asymmetry that several existing tests document and work around (a
-    /// regular node could not learn its gateway's version, so gateway-directed
-    /// PUTs had to originate from the gateway).
+    /// **Simulations default this OFF**, like the two cascade gates and unlike
+    /// `hash_first_summaries_floor_override`. The gate is encoding-only where it
+    /// fires, but the version it teaches is the INPUT to every other
+    /// `version_supports_*` gate, so enabling it network-wide makes
+    /// node->gateway links newly eligible for those features — a cascade in
+    /// effect. Measured rather than assumed: defaulting it ON changes the
+    /// outcome of the summary-first PUT sims.
+    ///
+    /// A sim that wants it calls `SimNetwork::enable_gateway_ack_version`; see
+    /// that method for the full argument and for what the OFF default costs.
     ///
     /// Not cfg-gated for the same reason as `subscribe_hint_floor_override`:
     /// `node::testing_impl` sets it and is compiled unconditionally.
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -712,6 +712,10 @@ pub(in crate::node) struct P2pConnManager {
     ledbat_min_ssthresh: Option<usize>,
     /// Congestion control configuration.
     congestion_config: CongestionControlConfig,
+    /// Test-only override for the #5161 version-carrying-ack floor, threaded
+    /// into the transport handler. `None` in production.
+    /// See `NodeConfig::ack_version_floor_override`.
+    ack_version_floor_override: Option<(u8, u8, u16)>,
     blocked_addresses: Option<HashSet<SocketAddr>>,
     /// Per-contract retry count for broadcasts that found no targets yet.
     broadcast_retries: HashMap<freenet_stdlib::prelude::ContractKey, u8>,
@@ -1195,6 +1199,7 @@ impl P2pConnManager {
             is_gateway: config.is_gateway,
             this_location: config.location,
             check_version: !config.config.network_api.ignore_protocol_version,
+            ack_version_floor_override: config.ack_version_floor_override,
             bandwidth_limit: config.config.network_api.bandwidth_limit,
             global_bandwidth: config
                 .config
@@ -1279,6 +1284,7 @@ impl P2pConnManager {
             global_bandwidth,
             ledbat_min_ssthresh,
             congestion_config,
+            ack_version_floor_override,
             blocked_addresses,
             broadcast_retries,
             broadcast_no_target_streak,
@@ -1296,6 +1302,7 @@ impl P2pConnManager {
                 global_bandwidth,
                 ledbat_min_ssthresh,
                 Some(congestion_config.clone()),
+                ack_version_floor_override,
             )
             .await?;
 
@@ -1365,6 +1372,9 @@ impl P2pConnManager {
             global_bandwidth: None, // Already used for connection handler, not needed in ctx
             ledbat_min_ssthresh,
             congestion_config, // Already used for connection handler, kept for struct completeness
+            // Already consumed by create_connection_handler above; kept for
+            // struct completeness, same as the two fields around it.
+            ack_version_floor_override,
             blocked_addresses,
             broadcast_retries,
             broadcast_no_target_streak,
--- a/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc/connection_lifecycle.rs
@@ -1053,9 +1053,11 @@ impl P2pConnManager {
         // `supports_summary_first_put` check. Keep this write paired with
         // the `self.connections.insert` above if that call site changes.
         // Write the Option THROUGH — unconditionally, so a reconnection whose
-        // version is unknown (`None`, the joiner->gateway path) CLEARS any
-        // stale entry rather than leaving us believing a downgraded peer is
-        // still current. See `ConnectionManager::record_remote_version`.
+        // version is unknown CLEARS any stale entry rather than leaving us
+        // believing a downgraded peer is still current. Since #5161 a `None`
+        // here means the remote is genuinely below the version-carrying-ack
+        // floor, not that this path cannot observe it.
+        // See `ConnectionManager::record_remote_version`.
         self.bridge
             .op_manager
             .ring
--- a/crates/core/src/node/testing_impl.rs
+++ b/crates/core/src/node/testing_impl.rs
@@ -543,6 +543,35 @@ impl ControlledSimulationResult {
             .is_some_and(|ring| ring.is_hosting_contract(key))
     }
 
+    /// The protocol version `label`'s node had recorded for the peer at `addr`
+    /// at the end of the run, or `None` if it never learned one (#5161).
+    ///
+    /// Reads the same `ConnectionManager` mirror the production emission gates
+    /// read, so it answers the question those gates ask rather than a proxy for
+    /// it. `None` when the node never published its Ring.
+    pub fn node_recorded_remote_version(
+        &self,
+        label: &NodeLabel,
+        addr: SocketAddr,
+    ) -> Option<(u8, u8, u16)> {
+        self.node_rings
+            .get(label)
+            .and_then(|ring| ring.connection_manager.remote_version(addr))
+    }
+
+    /// Whether `label`'s node would emit the hash-first summary encoding to the
+    /// peer at `addr` — i.e. whether that link passes the version gate (#4965).
+    ///
+    /// The end-to-end consequence of #5161 on a node->gateway link: before the
+    /// version-carrying ack this could never be true from the node's side, so
+    /// the node answered every anti-entropy exchange with full summary bytes.
+    /// `false` when the node never published its Ring.
+    pub fn node_supports_hash_first_summaries(&self, label: &NodeLabel, addr: SocketAddr) -> bool {
+        self.node_rings
+            .get(label)
+            .is_some_and(|ring| ring.connection_manager.supports_hash_first_summaries(addr))
+    }
+
     /// Number of contracts `label`'s node held in its hosting cache at the end
     /// of the run (its "cache size"). Returns 0 if the node never published its
     /// Ring. See [`Ring::hosting_contracts_count`](crate::ring::Ring::hosting_contracts_count).
@@ -1407,6 +1436,21 @@ pub struct SimNetwork {
     /// release PR, where a red sim could not be attributed between the
     /// feature and the version bump.
     hash_first_summaries_floor_override: Option<(u8, u8, u16)>,
+    /// Per-node version-carrying-ack version-floor override (#5161, see
+    /// [`SimNetwork::enable_gateway_ack_version`]).
+    ///
+    /// **Defaults to `Some(SIM_MIGRATION_DISABLED_FLOOR)` — OFF**, like the two
+    /// cascade gates and unlike `hash_first_summaries_floor_override`. The gate
+    /// is encoding-only where it fires, but the version it teaches is the INPUT
+    /// to every other `version_supports_*` gate, so enabling it network-wide
+    /// makes node->gateway links newly eligible for those features. Measured,
+    /// not assumed: defaulting it ON changed the outcome of unrelated sims.
+    ///
+    /// The cost of OFF, stated plainly: sims keep the pre-0.2.120 topology
+    /// asymmetry (a regular node never learns its gateway's version) even after
+    /// production stops having it. `enable_gateway_ack_version` is how a test
+    /// that cares opts in.
+    ack_version_floor_override: Option<(u8, u8, u16)>,
     /// Optional controllable hosting clock injected into every node's
     /// `HostingManager` (via `NodeConfig::hosting_time_source_override`). When
     /// set, hosting-cache TTL and subscription-lease eviction advance ONLY when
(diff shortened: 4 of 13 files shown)


sanity:
## Full-tier review

This touches the transport handshake and the wire format, both of which `~/.claude/rules/multi-model-review.md` names as always-Full. Four independent reviewers read the checked-out code blind, each with a distinct lens, none told the others' findings:

- **code-first** — read the code before the description, flag intent/impl divergence
- **testing** — attack the test suite; prove vacuity by mutation where possible
- **wire compatibility / mixed-version rollout** — the failure mode specific to this change
- **skeptical** — adversarial bug hunt in concurrency, lifetimes, control flow, error paths

**No reviewer found a blocking defect in the wire change.** Each independently confirmed the parts that matter most: `AckConnectionV2` is appended last so no bincode variant index moves; the restructured `if let`/`match` in `traverse_nat` is semantics-preserving (every path in the `if let` body diverges, and the compiler enforces it — `sent_tracker` is moved inside a loop, so any fall-through would be a use-after-move); the gate reads only the version parsed during *this* handshake, never a cached one, so it cannot go stale against a peer that downgraded at a reused address; and there is no wildcard `match` anywhere on `SymmetricMessagePayload` that could silently misclassify the new variant.

The rollout reviewer walked the full version-pair matrix, joiner→gateway and peer-to-peer, and found no cell that breaks. Two details worth recording: the `RemoteInbound` receive arm never deserializes the packet at all (it only checks `is_intro_packet()`), so a V2 ack arriving there is structurally invisible; and an old peer receiving V2 gets a clean serde `invalid variant index` error rather than a silent misparse, because the new form uses a new index rather than growing an existing struct — which is the trap `bincode`'s `allow_trailing_bytes` would otherwise have set.

The findings were in the tests and the surrounding procedure. All are now fixed in 509ad5d.

### Two tests were vacuous, both proven by mutation

**`simulate_gateway_connection` was tautological, and my change had made it weaker than what it replaced.** It computed its expected value by calling `version_supports_ack_version` — the predicate under test — with the production constant the production code reads. The reviewer mutated the gate to ignore the floor entirely (a 0.2.120 gateway sending `AckConnectionV2` to every 0.2.119 peer, which would break their handshakes outright and, since the release cascade upgrades gateways first, lock the fleet out of bootstrap) and the test stayed green. The line I had replaced, a plain `assert_eq!(joiner_view, None)`, would have caught it. The expectation is now derived by comparing the crate version to the floor directly, and I re-ran that mutation against the fix: it fails.

**The peer-to-peer emit site had no coverage whatsoever.** A reviewer installed `panic!()` in it; the whole `mock_transport` suite and the sim test all passed. My first attempt at a test for it was *itself* vacuous, and I only caught that by mutating: with no packet loss both peers parse each other's intro on the `StartOutbound` path and learn versions the old way, so the assertion held with the branch reverted. Forcing the race is the whole difficulty. `nat_traversal_ack_racer_learns_version_from_the_ack` drops peer B's first packet — its intro — so peer A never parses one, stays in `StartOutbound`, and receives B's ack instead. That is exactly the arm that used to yield `None`, and reverting the branch now fails the test.

### Two of the sim A/B test's four assertions carried no information

With the sim's default hash-first floor of `(0,0,0)`, `supports_hash_first_summaries` reduces algebraically to `remote_version.is_some()` — so asserting it alongside `is_some()` restated the premise and could not fail on a hash-first-gate defect. A third arm (ack gate on, hash-first floor unreachable) makes the two disagree: version known, gate still closed. The version is also now asserted as an exact value rather than `is_some`, since a regression recording the *wrong* version would have satisfied the old assertion.

### The cascade the simulation default excludes is now exercised

The code-first reviewer's main objection was that keeping the ack gate off suite-wide permanently excludes from simulation exactly what production starts doing at 0.2.120: node→gateway links become eligible for every other `version_supports_*` feature at once, on the highest-degree nodes on the network. That is fair, and the two summary-first tests that detected the difference had been left asserting the old topology.

I kept the OFF default — the argument in the PR body stands, and the reviewer's own "minimum fix" was to exercise the cascade somewhere rather than to flip it — but `test_gateway_ack_version_unlocks_node_originated_summary_first_put` now covers it. It is the reversed-roles mirror of `test_summary_first_put_holder_found_ships_delta`, whose rustdoc records that with the roles reversed the node's PUT "never even logs a `PUT summary-first:` line". With the ack gate on it does: 1 delta send with the gate, 0 without, mutation-confirmed.

### Procedure and clarity

- **`docs/RELEASING.md` had no entry for this floor**, and its framing ("features that add a new `NetMessageV1` wire variant") did not even cover a `SymmetricMessagePayload` variant, so a releaser scanning for applicable floors would have skipped it. Added, with the framing widened and the severity difference called out: the other three floors gate application messages, where a peer that cannot decode drops the connection; this one gates the ack itself, where a peer that cannot decode never completes the handshake. Also corrected a stale line claiming `HASH_FIRST_SHIPPED_IN` is "currently `None`" when it is `Some((0,2,116))`.
- **The `NodeConfig` rustdoc said "Simulations default this ON"** — a leftover from an earlier iteration, and the copy a maintainer is most likely to read first. It is OFF.
- **The gate was duplicated verbatim at two emit sites**; now one `build_success_ack`, so they cannot drift and the reasoning lives in one place.
- **The unreachable success-ack arms were hiding in a catch-all-by-enumeration** whose comment claimed the opposite of what it delivered: a future `AckConnectionV3` added there but forgotten in `ack_ok_parts` would compile, satisfy exhaustiveness, and present as a handshake that silently never completes. They are now their own arm, logged at ERROR.
- **`create_mock_peer_internal` silently dropped `bandwidth_limit`** when a floor override was passed. One constructor now takes both.
- **The duplicate-V2-ack-answered-with-legacy-ack decision was load-bearing and unpinned** — nothing drove `process_inbound` with either success ack. Now tested.
- **The golden literal cannot vouch for itself**, as the testing reviewer put it: a wrong literal would have been silently "fixed" by copying whatever the encoder emits. Its provenance and derivation are now recorded next to it, with an explicit instruction not to re-derive it from the current code if it ever fails.
- **`report_peer_version` is called here without `is_compatible`**, unlike the four intro-parsing sites. Deliberate — the acceptor already ran the bidirectional check before choosing to send V2, and would have replied `ack_error` had it required a version newer than ours, so the "we must update" branch is unreachable — but the asymmetry now says so in the code.

### Findings addressed by explanation rather than change

**The `version_discovery` global-state race.** The skeptical reviewer noted that the new joiner-side `report_peer_version` widens a pre-existing window in which a concurrent handshake test could corrupt `version_discovery_tests`' reads of `HIGHEST_SEEN_VERSION`. They could not reproduce it in 12 iterations. It is also not reachable under CI: the sim and lib suites run under `cargo nextest`, which gives every test its own process, so the process-global cannot leak between tests there. Serializing seventeen transport tests to close a window that only exists under plain `cargo test` is not a trade worth making inside a wire-format PR.

**The issue's third acceptance criterion.** #5161 asks that `record_remote_version(addr, None)` stop erasing a previously recorded version. This PR does the opposite and pins the erasure, for the reason argued in the PR body — the reviewer independently enumerated every producer of `remote_protoc_version` and confirmed the argument holds. But they were right that `Closes #5161` will read as though all three criteria were met. Flagged for @sanity: the erasure decision should be confirmed or overridden explicitly rather than absorbed by the auto-close.

Lenses run: code-first, testing, wire-compat/rollout, skeptical. External non-Claude models were not invoked, per the current review rule.

[AI-assisted - Claude]

sanity:
**Erasure decision: confirmed as an override, not an oversight.** The PR is right to pin the erasure and `Closes #5161` stands.

The issue's item 2 was already withdrawn publicly on #5161 before the review ran — I had proposed that change, an agent pushed back on it, and I verified in the code that the pushback was correct. The rustdoc at `ring/connection_manager.rs:466-486` records that the guard once existed, why its justification was wrong, and that it was removed because of this exact path: a peer reconnects on an *older* build at the same address, the joiner→gateway handshake yields `None`, and a guard would retain the stale newer version — so we would keep speaking a wire format that peer cannot decode. Retaining a version we can no longer vouch for is worse than admitting we do not know it, because `None` fails closed and a stale `Some` fails open.

Two of #5161's three acceptance criteria are met and the third is retracted at source. Nothing outstanding for me here.

Thanks for surfacing it rather than letting the auto-close absorb it — that was the right call.

[AI-assisted - Claude]

PR #5170, opened 2026-08-04
fix(ci): stop merge-queue entries cancelling each other's rule review

sanity:
## The bug

`claude-pr-review.yml` triggers on `merge_group` because branch protection requires its check to report there. Its concurrency key is:

```yaml
group: claude-pr-review-${{ github.event.pull_request.number }}
cancel-in-progress: true
```

`github.event.pull_request` **does not exist on a `merge_group` event.** The key therefore collapses to the constant `claude-pr-review-` for every queued PR, and with `cancel-in-progress: true` each new queue entry cancels the previous entry's run of a **required** check. A cancelled check is never reported, so the older entry goes `UNMERGEABLE` and stalls in the queue.

It only bites when two or more PRs are queued at once, which is presumably why it has gone unnoticed.

## Observed

2026-08-04, with #5160, #5163 and #5167 queued together for the 0.2.120 release:

```
1 #5160 UNMERGEABLE
2 #5163 AWAITING_CHECKS
3 #5167 QUEUED
```

#5160's rule-review run (`30959723455`, event `merge_group`) was **cancelled 2 seconds after it started** — created 23:21:38Z, cancelled 23:21:40Z — by #5163's entry. Meanwhile #5160 itself reported `CLEAN` / `MERGEABLE`, so nothing on the PR page indicated a problem. The only symptom is a queue entry that never advances.

## The fix

Fall back to `merge_group.head_sha`, which is unique per queue entry. That is the correct isolation unit: queue entries are independent, so none of them should cancel any other. `run_id` is a further fallback so a future trigger carrying neither field can never share a group.

## Similar-bug audit

The other two workflows triggering on `merge_group` — `ci.yml` and `macos-dmg-swap-e2e.yml` — both key on `${{ github.workflow }}-${{ github.ref }}`. On a `merge_group` event `github.ref` is the per-entry `gh-readonly-queue/...` branch, so both are already unique per entry and need no change. This workflow was the only one affected.

## Verification

This cannot be fully verified before merge: the failure mode requires two entries in the real merge queue, and the fixed key only takes effect once the workflow is on `main`. What is verifiable now is that the expression is valid and that the PR-event path is unchanged — `github.event.pull_request.number` is still the first operand, so nothing about per-PR behaviour moves. The proof will be the next multi-PR queue: entries should advance independently rather than the leader stalling.

[AI-assisted - Claude]

### DIFF
--- a/.github/scripts/check_merge_group_concurrency.py
+++ b/.github/scripts/check_merge_group_concurrency.py
@@ -1,0 +1,249 @@
+#!/usr/bin/env python3
+"""Guard against merge-queue entries cancelling each other's required checks.
+
+A workflow that triggers on ``merge_group`` runs once per merge-queue entry.
+Its ``concurrency.group`` key must therefore contain something that VARIES per
+entry. If it does not, every entry lands in the same group, and with
+``cancel-in-progress: true`` each new entry cancels the previous entry's run.
+When the workflow provides a REQUIRED check, a cancelled run never reports, so
+the older queue entry goes ``UNMERGEABLE`` and stalls the queue behind it.
+
+The trap is that this looks fine on a PR. ``github.event.pull_request`` simply
+does not exist on a ``merge_group`` event, so a key like::
+
+    group: claude-pr-review-${{ github.event.pull_request.number }}
+
+silently collapses to the constant ``claude-pr-review-`` -- GitHub's own
+cancellation message names that empty group verbatim. Nothing fails, nothing is
+logged on the PR, and the only symptom is a queue entry that never advances.
+
+Shipped in ``claude-pr-review.yml`` and found 2026-08-04 when three PRs were
+queued together for the 0.2.120 release. Fixed by #5170.
+
+A key is accepted when it references any token that differs between queue
+entries: ``merge_group`` (``merge_group.head_sha`` is the queue entry's own
+commit), ``github.ref`` (on a merge_group event this is the per-entry
+``gh-readonly-queue/...`` branch), ``github.sha``, or ``github.run_id``.
+
+Usage:
+    check_merge_group_concurrency.py [WORKFLOW_DIR]
+    check_merge_group_concurrency.py --self-test
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+# A `group:` key is per-entry-safe if it mentions any of these. `github.ref` and
+# `github.sha` both resolve to per-entry values on a merge_group event, and
+# `run_id` is unique per run, so any of them breaks the collapse.
+PER_ENTRY_TOKENS = ("merge_group", "github.ref", "github.sha", "github.run_id")
+
+# `  merge_group:` at exactly two-space indent is a trigger under `on:`. Deeper
+# indents are payload references (`github.event.merge_group.head_sha`), not
+# triggers, so anchoring the indent avoids matching a workflow that merely reads
+# the payload without being triggered by the event.
+MERGE_GROUP_TRIGGER_RE = re.compile(r"^  merge_group:\s*(?:#.*)?$", re.M)
+
+# Capture each `concurrency:` block's `group:` line. GitHub allows `concurrency`
+# at workflow level and per job, so scan for the key rather than assuming
+# top-level, and keep the indent so the message can point at the right one.
+CONCURRENCY_GROUP_RE = re.compile(
+    r"^(?P<indent>[ ]*)concurrency:\s*(?:#.*)?\n"
+    r"(?:^(?P=indent)[ ]+.*\n)*?"  # any keys before `group:` (e.g. `cancel-in-progress`)
+    r"^(?P=indent)[ ]+group:[ ]*(?P<group>.*)$",
+    re.M,
+)
+
+
+def _line_of(text: str, offset: int) -> int:
+    return text.count("\n", 0, offset) + 1
+
+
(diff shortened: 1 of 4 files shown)


sanity:
## Full-tier review

CI configuration is a high-risk surface under `~/.claude/rules/multi-model-review.md`, always Full regardless of diff size. Three independent reviewers read the checked-out workflow blind, each with a distinct lens, none told the others' findings:

- **Actions expression semantics** — verify the mechanics from documentation and live evidence, not intuition
- **regression** — the change *loosens* cancellation; what breaks or gets more expensive
- **skeptical, aimed at the diagnosis rather than the patch** — is the causal story established, or merely plausible

**No reviewer found a defect at any severity.** More usefully, two of them turned up evidence stronger than what the PR body originally claimed.

### The diagnosis is confirmed by GitHub, not inferred from timing

The skeptical reviewer pulled the cancelled run's check-run annotation:

```
Canceling since a higher priority waiting request for claude-pr-review- exists
```

Note the trailing dash and absent PR number. That is GitHub Actions stating that the concurrency group collapsed to the constant string — the mechanism, in GitHub's own words, not a correlation. Fetch it with `gh api repos/freenet/freenet-core/check-runs/92160747188/annotations`.

They also closed the gap between "check was cancelled" and "entry went UNMERGEABLE", which the PR body asserted without support: the repo ruleset (`4028534`) lists `Claude Rule Review` in `required_status_checks` with `strict_required_status_checks_policy: true`, so the check must report a fresh success against the queue's synthetic merge commit. A cancelled run never does. They separately ruled out a conflict or a moved base (`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`).

### The fix is GitHub's own documented idiom

The semantics reviewer found that GitHub's concurrency documentation recommends this exact pattern for this exact problem:

```yaml
concurrency:
  group: ${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

with the rationale "if your workflow responds to other events in addition to `pull_request` events, you will need to provide a fallback". They confirmed `merge_group.head_sha` against the machine-readable `octokit/webhooks` schema (top-level string, no null case), and confirmed `||` returns the operand rather than a coerced boolean, with `null` and `''` both in the documented falsy set. On `pull_request_target` the expression is byte-identical to the old one, since a PR number is never falsy — so the working per-PR path does not move.

### Loosening cancellation costs almost nothing here

The regression reviewer checked the thing worth worrying about — that fewer cancellations means concurrent runs racing the sticky comment and the commit status — and found it foreclosed by guards already in the file: every step that posts a comment or writes a status is gated `if: github.event_name != 'merge_group'`, and the Claude API call itself is skipped in merge_group context. So N queued PRs now produce N trivial jobs rather than N reviews. They also checked whether the repo's `github.ref` idiom (used by `ci.yml` and `macos-dmg-swap-e2e.yml`) would be more consistent here, and concluded it would **reintroduce this bug on the other half** — `github.ref` on `pull_request_target` is the base branch, so all PRs would collapse into one group.

### Still not verifiable before merge

As the PR body says: the failure needs two real queue entries, and the fixed key only takes effect once the workflow is on `main`. The proof is the next multi-PR queue advancing independently. Everything checkable without that has been checked.

Lenses run: Actions expression semantics, regression, skeptical/diagnosis. External non-Claude models were not invoked, per the current review rule.

[AI-assisted - Claude]

sanity:
## Addressed the rule-review warning: added the regression guard

The warning was fair and I have taken it rather than overridden it. It also caught a second thing on its own terms — this PR was titled `fix:` and added no test, which Rule Lint blocks. Both are closed by the same work.

`.github/scripts/check_merge_group_concurrency.py` asserts that any workflow triggering on `merge_group` has a concurrency group that varies per queue entry: `merge_group.head_sha`, `github.ref` (the per-entry `gh-readonly-queue/...` branch), `github.sha`, or `run_id`. Unlike the other rule-lint linters it is **not diff-scoped** — the invariant is a property of each workflow as it stands, and scanning all of them takes milliseconds.

**Confirmed non-vacuous rather than assumed.** Run against the real pre-fix `claude-pr-review.yml`, reconstructed from `HEAD~1`, it exits 1 and names the collapsing key:

```
claude-pr-review.yml:98: workflow triggers on `merge_group` but its concurrency
group does not vary per queue entry:
      group: claude-pr-review-${{ github.event.pull_request.number }}
```

The seven fixtures deliberately cover the cases where a naive matcher goes wrong:

- the `github.ref` form used by `ci.yml` and `macos-dmg-swap-e2e.yml` must **not** be flagged, or the linter would demand pointless changes to two working workflows;
- `merge_group` appearing only as a payload reference (`github.event.merge_group.head_commit.message`, which `ci.yml` really does read) is not a trigger;
- the group key is found regardless of its order inside the `concurrency:` block, or whether it sits at workflow or job level.

`scripts/merge_group_concurrency_test.sh` covers what a fixture suite structurally cannot: that the linter fires on the exact key that shipped, that this repo satisfies the invariant today, and that the linter is **still wired into `rule_lint`**. That last check matters — deleting the CI step would otherwise retire the guard while every other test kept passing, which is the same shape of silent failure as the bug itself.

Both are run in CI: the self-test and the repo scan in `rule_lint`, the shell test alongside the other `*_test.sh` self-tests.

The Full-tier review above stands — this commit is additive test and lint infrastructure and does not touch the one-line fix it guards.

[AI-assisted - Claude]

Issue #5046, opened 2026-07-30
feat(telemetry): actual Otel metrics export


skandragon:
I would like to be able to monitor my node for longer periods of time than a snapshot, and see internals like where CPU time is going for what is running, memory usage for the subsystems, etc.  I see that there is a custom log exporter that apparently drives the public dashboard, so would not want to break that, but I would like logs for the Freenet server as well as metrics to go to my collector.

Is there interest in this beyond my preference to make sure I monitor all the things and want to see more details on what my node is doing?

sanity:
# Issue Priority Assessment

Generated: 2026-08-14T01:20:56.135883065+00:00
Issue hash: `cc4719d3`

## Value Assessment

This confirms the picture the issue describes: `trace-ot` feature wires up `tracing-opentelemetry`/`opentelemetry-otlp` crates for distributed *tracing* (spans/Jaeger export), and separately there's a custom binary WebSocket protocol (`metrics_client.rs`, `NetLogMessage`/`PeerChange`) that feeds the public telemetry dashboard. Neither path exports OTel **metrics** (counters/gauges/histograms via an OTLP metrics pipeline) that a user could point their own collector at — `opentelemetry_sdk` is pulled in with `rt-tokio` but nothing in `crates/core/src` constructs a `MeterProvider` or OTLP metrics exporter. So the issue is accurate and still unresolved.

## Assessment

**Relevance**: Still fully applicable. The dependency scaffolding for OpenTelemetry already exists (`opentelemetry`, `opentelemetry-otlp`, `opentelemetry_sdk`, `tracing-opentelemetry` in `Cargo.toml`, gated behind the `trace-ot` feature), but it's used only for span/trace export to Jaeger-style backends, not metrics. The actual "monitor my node over time" data path is the bespoke WebSocket protocol in `crates/core/src/tracing/metrics_client.rs` that pushes `NetLogMessage`/`PeerChange` events to a hardcoded local port (`FDEV_NETWORK_METRICS_SERVER_PORT`, default 55010) feeding the public dashboard — not a standards-based collector endpoint a user could point Prometheus/Grafana/any OTel collector at. No code changes since this issue was filed have addressed it.

[...]

PR #5178, opened 2026-08-05
feat(otel): export node metrics via the OpenTelemetry SDK

skandragon:
## Problem

Node operators have no way to monitor a node over time — memory, ring health, transport throughput, queue depths — beyond point-in-time snapshots. The existing telemetry pipeline (`telemetry-enabled` / `telemetry-endpoint`) is a hand-rolled OTLP-JSON log POST that feeds the project's central dashboard; it isn't a general-purpose metrics exporter and operators shouldn't have to piggyback on it.

## Approach

Adds a second, fully independent pipeline built on the OpenTelemetry SDK (`tracing/otel.rs`), gated by `otel-telemetry-enabled` / `otel-endpoint`. The two pipelines share no config, no endpoint, and no fallback — enabling one has no effect on the other, and `otel-endpoint` never defaults to the dashboard collector.

Key decisions:

- **Standard OTEL env vars win over config** (`OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_SERVICE_NAME`, etc.); without any, exports go to `http://localhost:4318`.
- **Counters are observable, never mirrored.** Every counter is an observable callback reading a cumulative atomic the measured code already keeps, so there is no second call site to forget when someone adds one — a hand-placed `otel::record_*` next to an existing `fetch_add` is the #4009/#4010 rot class. Where only a period counter existed (`transfers_*`, `nat_traversal_*`), the observable reads a **new never-reset twin** incremented in the same function: `take_snapshot` swaps the period counters to zero for the legacy telemetry worker, so observing those directly would go non-monotonic on any node running both pipelines. Only the two histograms are synchronous — a sum-and-count atomic cannot rebuild a distribution — and each of those two call sites is pinned to its own function.
- **All histograms are base-2 exponential** via a single `with_view` — no per-instrument bucket boundaries.
- **Collector auth** (`otel-auth-mode`, default `disabled`): in `freenet` mode, a per-request `Authorization: Bearer freenet/<pubkey>/<audience>/<timestamp>/<signature>` — an XEdDSA signature over the token prefix, signed with the node's x25519 transport secret. The collector verifies with stock Ed25519 after Montgomery→Edwards conversion; no shared secret, no second key. `<audience>` is base58 of the first 16 bytes of `SHA-256("{host}:{port}{path}")` for the target collector — a hash rather than the URL because a URL contains the token's `/` separator; the canonicalization rules are in both docs so a collector can reproduce it. A token cannot be replayed at a different collector.

  Default is `disabled` (no header) so that pointing the exporter at your own collector never ships a signed identity assertion unasked; an `Authorization` supplied via `OTEL_EXPORTER_OTLP_HEADERS` is never overwritten in either mode — which is how static header auth, the only other scheme in common use with OTLP, is configured today.

  To be precise about what `disabled` withholds: the **signature**, not the identity. `freenet.node.pubkey` / `.fingerprint` ride every export batch in both modes, because metrics with no node id to group by aren't useful. Both docs now say this plainly.
- **No per-connection identifying attributes** (series-cost multiplication); the node identifies itself via two resource attributes, `freenet.node.pubkey` (verifiable against the bearer signature) and `freenet.node.fingerprint` (dashboard cross-reference). Never a `PeerId`, which would leak the socket address.

### Build cost

- **No new native build dependency**: the exporter always installs its own `HttpClient` (workspace reqwest 0.12), so none of `opentelemetry-otlp`'s `reqwest-*`/TLS features are enabled. Net new packages in `Cargo.lock`: `xeddsa` + `convert_case`.
- **`opentelemetry-otlp` and `opentelemetry_sdk` move from `optional = true` to non-optional**, so *every* build and cross-compile target now links the OTLP exporter, `prost` and `opentelemetry-proto` — including builds where `otel-telemetry-enabled` is off. Measured on this branch's release binary vs `upstream/main` (aarch64-apple-darwin, `cargo build --release --bin freenet`):

  | | bytes |
  |---|---|
  | `upstream/main` (7c5cc4be) | 47,067,488 |
  | this branch | 48,469,168 |
  | **delta** | **+1,401,680 (+1.34 MiB, +3.0%)** |

  The disabled path is inert at runtime (`init` returns before building anything), so this is link cost only.

Design doc: `docs/design/otel-metrics-exporter.md`. Operator docs: `docs/otel-metrics.md`.

## Testing

- Unit tests covering config parsing/precedence (flat keys, env-var priority, endpoint isolation), provider construction inside a tokio runtime, bearer-token format and XEdDSA sign/verify round-trip, audience canonicalization, suppression when disabled, and resource-attribute derivation.
- **Instruments are actually executed.** `instrument_callbacks_export_named_datapoints` builds a provider over `InMemoryMetricExporter`, registers every instrument against it, collects once, and asserts the exported datapoints **by name and by attribute**. `init` returns early under `cfg(test)`, so before this no callback, name, unit or attribute ran in CI — and a panic in any callback kills the `PeriodicReader` thread and stops all metrics permanently, with no export-side signal to report it. It uses a **local** provider: neither `global::set_meter_provider` nor the `INSTRUMENTS` OnceLock is touched, since both are process-global and would leak across tests in one `cargo test` process.
- **The source-scrape pin was mutation-tested.** It is bounded to the calling function now (a `method_body` helper, the indented sibling of `fn_body` in `bin/commands/auto_update.rs`). Verified by moving `record_rtt_ms` out of `record_rtt_sample` into `record_transfer_failed` — the bounded pin fails, the previous whole-file `contains` stayed green.
- `cumulative_outcome_counters_survive_a_snapshot_reset` pins the non-monotonicity hazard directly.
- **Run under plain `cargo test`, not only nextest** (per `.claude/rules/testing.md` — CI's process-per-test cannot see interference through `INSTRUMENTS` / `EXPORT_FAILING`): `cargo test --release -p freenet --lib -- otel:: config:: ring::hosting:: transport::metrics:: network_status` → **471 passed, 0 failed**. Full `--lib` run: 4956 passed, 7 failed — all 7 pre-existing, reproduced identically on a clean `upstream/main` worktree (log-capture/tracing-interest and temp-dir cross-test interference in `wasm_runtime::migration_tests`, `operations::update::tests::log_severity`, `subscriber_limit_tests`).
- Zero `unwrap`/`expect`/`panic!`/indexing in production paths, still.
- Verified end-to-end against a local OTLP collector on :4318.

Note: `freenet.process.memory.rss` is Linux-only by design; on macOS/Windows the gauge registers but emits no datapoints.

## Fixes

Refs #5046 — that issue also asks for log export, which is an explicit non-goal here, so it stays open.

[AI-assisted - Claude]

---

## Maintainer changes (2026-09-03)

Pushed onto this branch by the maintainers rather than sent back as another review round. The design is unchanged and the auth stays — see below.

**On the auth-mode split.** Resolved: `otel-auth-mode` stays in this PR. The reviewer's case for splitting rested on there being no collector that ingests from nodes it does not operate; @skandragon pointed out that is exactly what he is offering (a no-signup shared collector, org owned by the project), and that a static token cannot serve it because issuing per-node tokens *is* signup. That is the case node-identity signing exists for, so the premise was wrong and the split is withdrawn.

That decision makes the three security items previously parked "for PR B" blocking here, since they would now ship. All three are addressed:

- **The signature now covers the request body.** The signing input is the token prefix plus `/<base58 SHA-256 of the body>`; the hash is not transmitted, and the collector recomputes it from what it received. Without this a token authenticated only "this node addressed this collector at this second", so anyone holding one could attach it to metrics of their own invention — the spoofing the scheme exists to prevent. Token stays five fields.
- **A replay bound is specified.** `REPLAY_WINDOW` (300s) is declared as a wire contract, and `docs/otel-metrics.md` now carries a single normative list of all five collector obligations — verify signature, verify body hash, verify audience, enforce the clock window, refuse a repeated `(pubkey, timestamp, body hash)` — in that order, with a note on what a replay actually gains an attacker given the body is bound.
- **Plaintext `http` is no longer silently accepted under `auth-mode = freenet`**, and no credential of any kind goes out in cleartext to a non-loopback collector. It fails the export with a `WARN` naming the reason, and says so at startup rather than one export interval later.

Two ambient mechanisms would have defeated that last check, since it can only inspect the URI the exporter aimed at, so both are disabled on the client and pinned by a test: reqwest replays a redirect with the original headers, and defaults to honouring `HTTP_PROXY` with no loopback exemption of its own — so a proxy in the environment would take an export aimed at `http://localhost:4318` off the machine, credential attached.

The cleartext guard is not `Authorization`-only. Every header declared through `OTEL_EXPORTER_OTLP_HEADERS` counts, because that is where OTLP credentials live and `Authorization` is merely the most common spelling — `x-honeycomb-team`, `api-key` and `dd-api-key` are all in ordinary use. Those values are also redacted out of any logged error body by value rather than by keyword, and before truncation rather than after; endpoints are stripped of userinfo in all five log sites that carry the raw operator string.

**One real bug fixed.** `HostingReason::Restored` classified on `!seeded_this_run` alone, which no code path clears, so after any restart the bucket absorbed the entire persisted hosted set permanently and `reason="routed"` would have read near zero fleet-wide. It now also requires `read_count == 0`, which the reload does reset, with a regression test that drives a routed access through a restored contract. `HostingReason`'s array indexing moved from `reason as usize` to a compiler-checked exhaustive match, since an out-of-range index there panics inside an observable callback and would silently kill every metric for the process lifetime.

**Test coverage.** The nine ring/status-sourced callbacks are guarded by `if let Some(x) = <source>()`, so in a fresh test process their bodies never ran and half the instruments asserted nothing. `RingSources` puts those accessors behind function pointers so the export test injects fixtures; all twenty instruments and every `reason=` / `op=` / `result=` / `position=` attribute are now asserted, including all seven `HostingReason` strings. The f64 gauge's attributes were structurally unassertable and now are not.

**Also.** `freenet.connect.attempts` renamed to `freenet.connect.gateway_failures` — its only writer is inside `record_gateway_failure`, and an instrument name is an external contract operators alert on. The error-body read is capped, generously enough not to truncate a legitimate `partial_success`. A test pins that the exporter stays opt-in: off by default, auth off by default, no endpoint baked in, default resolving to loopback — a privacy-focused network must not report to a third party unless asked. Plus the doc-drift fixes from review.

Reviewed at Full tier: four independent Claude lenses plus `codex review`, findings synthesised and each fixed or justified.


sanity:
## Review

Thanks for this. The craft here is high, and a few things in it are genuinely careful in ways that are easy to get wrong: the `OnceLock` reasoning about instruments binding to the no-op provider, the `std::thread::scope` hop to build the exporter off the async runtime, dodging the clap `SetTrue` trap on `--otel-telemetry-enabled`, and adding *new* cumulative packet counters rather than observing the period accumulators that `take_snapshot` zeroes. That last one is exactly the trap the AGENTS.md note warns about, and you avoided it deliberately.

The problems below are mostly about **packaging**, not correctness. The biggest one has a small fix.

Review ran five independent lenses (supply chain / security, cryptography, code-first correctness, test quality, and architecture), plus direct verification of the load-bearing claims against crate sources.

---

### Blocking

**1. The OTel stack is now unconditional, and it pulls a C build dependency into every build.**

`crates/core/Cargo.toml` moves `opentelemetry-otlp` and `opentelemetry_sdk` out of the optional `trace-ot` feature, so every freenet binary compiles them whether or not the operator ever sets `otel-telemetry-enabled`. Via `reqwest-rustls` that pulls `rustls-platform-verifier`, which pulls `aws-lc-rs` and `aws-lc-sys`. `aws-lc-sys` builds AWS-LC from C and assembly through `cc` + `cmake`.

Eleven new lockfile packages, ten of them downstream of that chain, plus a second reqwest major (0.13 alongside the workspace's 0.12) and a second TLS root store.

Suggested fix: put the pipeline behind a default-off cargo feature, or switch `reqwest-rustls` to `reqwest-rustls-webpki-roots`, which skips the platform verifier and therefore aws-lc entirely. Worth noting the default `auth_mode = "freenet"` path uses the workspace reqwest 0.12 (which already has TLS) through `FreenetAuthClient`, so the 0.13 TLS stack is dead weight in the default configuration.

**2. The release build path was never exercised, and it is where this is most likely to break.**

`.github/workflows/cross-compile.yml` explicitly skips PRs ("Skip PRs - only build on main and releases to reduce CI costs"). It builds `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` (the arm64 job installs only `musl-tools musl-dev`), a macOS cross-compile, and Windows MSVC. No workflow in the repo installs `cmake` or `nasm`.

So `aws-lc-sys` can pass every PR check and fail on `main`, on the release path, after merge. `docs/design/otel-metrics-exporter.md` lines 246-251 identifies this risk and says to verify with a `workflow_dispatch` run on the branch before merging. Please do that and post the result.

**3. Dual rustls crypto provider.**

On `main`, `rustls 0.23.40` resolves with `ring` only. With this PR it carries both `ring` and `aws-lc-rs`. In `rustls-0.23.40/src/crypto/mod.rs:265-286`, `from_crate_features()` has one cfg block per provider and each requires `not(feature = <the other>)`, so with both enabled it returns `None`, and `get_default_or_install_from_crate_features()` panics with "Could not automatically determine the process-level CryptoProvider". There is no `CryptoProvider::install_default()` anywhere in `crates/`.

To be fair about severity: this does **not** panic today. Every reachable caller passes a provider explicitly (`reqwest-0.12.28/src/async_impl/client.rs:770` hardcodes `ring`, reqwest 0.13 hardcodes `aws_lc_rs`, sqlx uses `tls-rustls-ring`). What the PR removes is the property that made it impossible. The bare `ClientConfig::builder()` path does exist in `tokio-tungstenite-0.27.0/src/tls.rs:126` for `wss://` connects, and that crate also backs freenet-stdlib's `WebApi`, so the exposure reaches downstream consumers. This codebase has hit the dual-provider panic before, and CI did not catch it that time either. Fixing (1) removes this.

**4. `rule-review/warnings` is red** on the untested `OtelAuthMode::Disabled` path. That gate blocks merge. (The `macOS Service Unit` failure was an unrelated rustc SIGSEGV compiling `wasmprinter`; it passed on re-run.)

---

### Functional bugs worth fixing before merge

**5. `OTEL_SERVICE_NAME` is silently ignored,** which contradicts the PR description's "Standard OTEL env vars win over config".

`Resource::builder()` seeds from `EnvResourceDetector` into `self` (`opentelemetry_sdk-0.32.1/src/resource/mod.rs:62-68`). `ResourceBuilder::with_attributes` then does `self.resource.merge(&Resource::new(kvs))`, and `merge` inserts `other` over `self` (`mod.rs:179-182`). So `.with_service_name("freenet-node")` overwrites the env value. Same shape for `service.version`, `os.type`, and `host.arch` against `OTEL_RESOURCE_ATTRIBUTES`.

Effect: two nodes on one host with distinct `OTEL_SERVICE_NAME` values both export `service.name=freenet-node`, and any dashboard keyed on it merges them. Fix: apply the literal only when the env var is absent.

**6. The bearer token clobbers an operator's own `Authorization` header.**

`tracing/otel.rs:142` does an unconditional `headers_mut().insert(AUTHORIZATION, ...)`, and the exporter applies `OTEL_EXPORTER_OTLP_HEADERS` before calling `send_bytes`. An operator pointing at a hosted collector that requires `Authorization: Basic ...` gets a 401 on every export, and the only workaround (`otel-auth-mode = disabled`) is not documented anywhere operator-facing.

This one bites the use case in #5046 directly, since the ask was to export to the operator's own collector. Fix: `if !request.headers().contains_key(AUTHORIZATION)`.

Related: `FreenetAuthClient` uses `reqwest::blocking::Client::new()`, where the client it replaces is built with `.timeout(resolve_timeout(...))`, so `OTEL_EXPORTER_OTLP_TIMEOUT` is ignored.

**7. The bearer token has no audience binding.**

Commit `7035ae4a4` removed the nonce, so the signed payload is `freenet/<pubkey>/<timestamp>` and does not name the collector it is for. A collector you export to can replay your token to any other collector accepting this scheme and impersonate your node, bounded only by clock skew. The default endpoint is plain `http`, so a passive observer on that path can do the same.

Two thoughts. First, if the auth scheme stays, sign the target endpoint into the payload. Second, and worth considering more seriously: `auth_mode` defaults to `freenet`, so an operator pointing at their own collector ships a signed assertion of their node identity there every 60 seconds without asking for it, and there is no collector today that verifies these tokens. Defaulting to `disabled` and letting `OTEL_EXPORTER_OTLP_HEADERS` carry auth would cover #5046's actual request. The XEdDSA scheme could then land separately, alongside the collector that consumes it.

To be clear, the cryptography itself checks out. I read `xeddsa 1.1.0`'s source: it clamps the scalar, forces the Edwards sign bit to zero by negating when needed, and derives the nonce as `H(padding || a || M || Z)`, so even a failing RNG cannot leak the key. `auth_token_signer()` passes the x25519 secret correctly, and the interop test verifies with genuinely independent `curve25519-dalek` + `ed25519-dalek` rather than the code that signed. The concern is protocol design and defaults, not the primitive.

**8. `AGENTS.md` describes the wrong key type.** It says `freenet.node.pubkey` is "the base58 ed25519 verifying key derived from the transport keypair". The code emits the **x25519 Montgomery** public key (`transport/crypto.rs:117-119`). Those are different byte strings, and a collector implementer following that line would call `VerifyingKey::from_bytes` on it and fail. The module rustdoc in `tracing/otel.rs` gets it right. Same paragraph calls `freenet.node.fingerprint` "unverifiable by the collector", while `otel.rs` and its test say the collector recomputes it.

**9. The startup log hides the destination in the one case it matters.** When the endpoint comes from `OTEL_EXPORTER_OTLP_*`, the log prints the literal `<resolved by OTEL_* env or SDK default>` instead of the URL. Env-over-config is the documented and intended precedence, so that part is fine, but an operator whose config endpoint was overridden by an inherited env var cannot tell from the logs where their node's signed identity is being sent. Logging the resolved value, and warning when env overrides a config endpoint, is cheap.

---

### Tests

Several of the new tests cannot fail for the reason they name:

- Deleting the entire suppression block in `init()` leaves every test passing. `otel_suppression_reason` is well covered but `init()` is its only caller and is unreachable under `cfg(test)`, so a `--id` test network would ship to a collector with no test going red. A source-scrape pin that `init` calls it and returns before `build_provider` would close this.
- The identity-attribute guards build their expected strings in the test body and never call `init` or `build_provider`. Changing `init` to use `peer_id.to_string()`, the exact regression the comment names, leaves them green. Extracting an `identity_attributes(keypair)` helper and asserting on its output would fix it.
- `OtelConfig` is not destructured in the config round-trip guard, unlike `TelemetryConfig` and its other siblings. Per `.claude/rules/code-style.md` the guard's value is that a new field fails to *compile* until classified; a fourth `OtelConfig` field will compile clean and silently never merge. This is the #3890 / #4275 class.
- `DEFAULT_OTEL_ENDPOINT` has no production reader; its only consumers are the two assertions in the test that pins it.
- The eight new `record_*` mirrors in `transport/metrics.rs` have no pin. Delete any one and the counter reports zero forever with nothing failing. This is verbatim the "manually-mirrored telemetry counters" row in `.claude/rules/bug-prevention-patterns.md` (#4009 / #4010 precedents).
- `--otel-telemetry-enabled=false` cannot override a config-file `true` (`config.rs:922` is one-directional, and clap's `default_value = "false"` makes "unset" and "explicitly false" indistinguishable). The test that looks like it covers this asserts at parse level, before the merge. Making the field `Option<bool>` with `get_or_insert`, matching `endpoint` and `auth_mode`, would fix both. I realize this matches the existing `reference-ping` / `iface-tx` pattern, so it is not a regression, but "the off switch does not work" reads differently for a network export than for a ping toggle.

---

### Smaller

- `docs/superpowers/` is a new directory whose only content is an 869-line agent work plan, already stale against what shipped (it describes a three-field `OtelConfig`; the shipped struct has four). `docs/design/otel-metrics-exporter.md` links to it as the canonical plan, so a reader gets pointed at the stale copy. Suggest deleting it.
- `docs/design/otel-metrics-exporter.md` still says "Status: proposed" and disagrees with the code in several places: it documents `otel::init(&self.config.otel, self.local_peer_id_string())` where the code passes `&self.key_pair`, documents `service.instance.id` and `peer.id` attributes that are never set, and its configuration table omits `otel-auth-mode` entirely.
- The `ponytail:` comment convention is defined only in this PR's own docs. Future readers of `crates/core/` will not know what it means. A plain `NOTE:` with the same text, or a tracked issue reference, would carry better.
- "Closes #5046" is a little strong given log export is an explicit non-goal; the issue asks for "logs for the Freenet server as well as metrics". Maybe leave it open.
- No operator-facing docs for `otel-telemetry-enabled` / `otel-endpoint` / `otel-auth-mode`. The requester is a node operator, and these currently appear only in `AGENTS.md` and `docs/design/`. Compare `docs/secrets-at-rest.md`.
- A poisoned `NETWORK_STATUS` lock takes down ten of the nineteen instruments permanently, including the fair-queue stats, which are pure atomics and do not need that lock at all.
- `freenet.contract.queue.depth` emits `queue="total"` alongside the three tiers, so `sum by (queue)` double-counts.
- `error_for_status()` discards the collector's response body, which is where OTLP partial-success and error detail live.
- `AGENTS.md` gains 72 lines for an off-by-default exporter. That file is loaded into every agent session in this repo. The genuinely rule-shaped parts (do not add per-instrument bucket boundaries, never export a `PeerId`, read cumulative rather than period counters) are about eight lines plus a link to the design doc.

---

### Checked and clean

Recording these so nobody re-runs them: no CI, workflow, build script, or `.claude/rules` changes. The transport secret is never logged or serialized (the manual `Debug` impl on `FreenetAuthClient` emits nothing), and no attacker-influenced bytes reach `sign()`. Every exported attribute is either a node-derived constant or a low-cardinality literal; no `PeerId`, socket address, IP, or contract key appears anywhere, and the `Resource` builder attaches no hostname or process detector. No deadlock or hot-path stall from the observable callbacks: `fair_queue_stats()` and the transport counters are lock-free atomic loads, and `otel_metrics_snapshot()` takes only short read locks that already have the homepage handler as a caller. No task or thread leak from the missing shutdown hook, and it cannot block process exit. Endpoint precedence matches `resolve_http_endpoint` exactly, including the `/v1/metrics` asymmetry. Isolation from the existing dashboard pipeline is structurally enforced rather than asserted.

---

### Suggested path

Items 1 through 4 are what I would want resolved before merge. Feature-gating (1) default-off takes care of (3) as a side effect and shrinks (2) to a formality. Items 5, 6, and 8 are small and worth doing in the same pass. Item 7 is the one real design question, and splitting the auth scheme into its own change would let the metrics exporter land sooner.

Happy to re-review once those are in.

[AI-assisted - Claude]

skandragon:
Thanks — most of this landed. Pushed as `faa6a1e8`.

### 1–3 (deps, release build, dual crypto provider): fixed, but not the suggested way

The `reqwest-rustls` → `reqwest-rustls-webpki-roots` swap does **not** drop aws-lc:

```
opentelemetry-http/reqwest-rustls-webpki-roots = ["reqwest/default-tls", "reqwest/webpki-roots"]
reqwest 0.13: default-tls = ["rustls"]; rustls = ["__rustls-aws-lc-rs", "dep:rustls-platform-verifier", …]
```

It adds webpki roots on top of the same stack.

What actually removes it: `FreenetAuthClient` (now `OtlpHttpClient`) is installed in **every** auth mode, not just when a signer exists. `opentelemetry-otlp` builds a client only when none was supplied (`exporter/http/mod.rs:223-267`), so its `reqwest-blocking-client` / `reqwest-rustls` features are now off entirely. Our client rides the workspace reqwest 0.12, which already has `rustls-tls`, so `https://` endpoints still work.

`Cargo.lock` delta versus `main` went from 11 new packages to **2** — `xeddsa` and its `convert_case`. No reqwest 0.13, no `rustls-platform-verifier`, no `aws-lc-rs`/`aws-lc-sys`, no `cmake`, and rustls resolves with `ring` alone again. That also means (2) no longer has a native-toolchain risk to verify, and (3) is structurally back to impossible.

I preferred this to a default-off cargo feature: feature-gating would mean an operator can't enable metrics without building their own binary, which defeats the point of the issue.

### 4 (rule-review warning): fixed — `provider_builds_with_auth_disabled`.

### 5, 6, 8, 9: fixed

- Resource literals are applied only for keys `OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES` did not declare.
- `Authorization` is set only when absent, so `OTEL_EXPORTER_OTLP_HEADERS` wins. Also honors `OTEL_EXPORTER_OTLP_{METRICS_,}TIMEOUT` (the SDK only resolves that for a client it builds itself), and no longer calls `error_for_status()`, so the body survives for the SDK's `HttpClient.StatusError` log.
- `AGENTS.md` corrected (x25519, not ed25519) and cut from +72 lines to +29: the rule-shaped parts plus links.
- Startup log names the resolved endpoint and warns when an `OTEL_*` variable overrode a configured one.

### 7 (audience binding + default): both done

Default `otel-auth-mode` is now `disabled`. Tokens are `freenet/<pubkey>/<audience>/<timestamp>/<signature>` where `<audience>` is the target collector's `host:port`, taken from the request URI — a token sent to one collector no longer verifies at another (`a_token_for_one_collector_does_not_verify_at_another`).

One correction on the premise, though: **a collector that verifies these tokens does exist and works.** It's an OTel collector extension — <https://github.com/cardinalhq/cardinalhq-otel-collector/tree/main/extension/fbnauthextension> — which is what the scheme was built against, and it verified end-to-end before this PR opened. I'll update it to match whatever format we settle on here, so treat the wire format as ours to choose. That's also why I'd rather keep the scheme in this PR than split it out.

### Tests

Fixed the vacuous ones you named:

- `init` now returns its suppression reason, so `init_refuses_to_start_from_a_test_process` asserts on `init`'s own behavior rather than on `otel_suppression_reason` in isolation. Mutation-tested: deleting the suppression block makes it fail.
- Identity attributes come from an `identity_attributes(keypair)` helper that production calls; the guards assert on its output.
- `OtelConfig` is destructured in the config round-trip guard.
- Cross-file scrape pins all nine hot-path `record_*` mirrors (cross-file, so it can't be satisfied by its own literal). Mutation-tested: deleting one fails.
- `DEFAULT_OTEL_ENDPOINT` deleted — you're right that it had no production reader.
- `otel-telemetry-enabled` is `Option<bool>` now, and `otel_cli_false_overrides_a_config_file_that_says_true` asserts through `build()`, not at parse level.

### Smaller

`docs/superpowers/` deleted. Design doc updated to match what shipped (status, auth-mode row, wire-up signature, dependency section). `ponytail:` → `NOTE:`. `queue="total"` dropped. `Closes #5046` → `Refs`, since log export is a non-goal. New operator page at `docs/otel-metrics.md`.

Not changed: the poisoned-`NETWORK_STATUS`-lock point is fair but it's a pre-existing property of `otel_metrics_snapshot`'s callers, and splitting the fair-queue reads out of it is a separate change.

sanity:
## Review

Thanks for this — it's a strong piece of work and I'd like to land it. Some specifics worth naming, because they're things that usually go wrong and didn't: zero `unwrap`/`expect`/`panic!`/indexing across all 1083 production lines, the dependency reasoning at the `opentelemetry-otlp` dep site (which not only avoids the second reqwest major and the aws-lc/rustls double-provider panic, but actually *removes* `reqwest 0.13.3` from the lock), the `std::thread::scope` hop off the async runtime, and verifying the XEdDSA pubkey against stock `ed25519-dalek` rather than round-tripping through the same crate that produced the signature. The docs are genuinely good.

I ran this through a multi-perspective review (code-first, testing, skeptical, big-picture, a crypto/privacy pass, plus an external non-Claude model). A few things I want to flag as *checked and fine*, so nobody re-litigates them: no deadlock or lock inversion in the collection path; the blocking reqwest call inside `send_bytes` is safe because `PeriodicReader` runs on its own OS thread (`opentelemetry_sdk-0.32.1/src/metrics/periodic_reader.rs:170`), so export backpressure can't reach node operations; the disabled path is genuinely inert; and no tests were removed or weakened.

## Proposal: split this into two PRs

The metrics exporter and the XEdDSA collector authentication are separable, and I think splitting gets the exporter merged much sooner. The exporter needs a fixed list of mechanical changes (below). The auth scheme needs a design decision from me that I haven't made yet — specifically whether Freenet is heading toward a federated/shared-collector model, which is the only thing that makes proving node identity to a collector load-bearing. I don't want that question to hold up the metrics work, and I don't want you iterating on the auth details before I've answered it.

So: **PR A = metrics exporter, `otel-auth-mode` omitted entirely** (or hard-wired to `disabled` with no `freenet` variant). **PR B = collector auth**, opened after we've settled the collector question. I've listed the auth findings at the bottom so none of it is lost.

More on that below — first, the one thing blocking CI right now.

## Blocking CI

**1. `crates/core/src/ring/hosting.rs:4608` doesn't compile.** The merge of `main` in 9bfa4ae4 was a clean textual merge but a semantic conflict: #5279 added a fourth `HostingCause` argument to `record_contract_access`, and this PR's new `hosted_by_reason_partitions_the_hosting_cache` test still calls the three-arg form. The library compiles (Windows Check passed) but the lib-test target doesn't, which is why Clippy, Simulation, and Unit & Integration all fail identically. `HostingCause::Other` is the right value here. Nothing else in CI is red.

Worth knowing: because the test target never built, **no test job has actually executed on this branch**, so CI has told us nothing about behaviour yet.

## Required for PR A

### Correctness — the hosted-set classifier

**2. `local_access` reads a sticky flag, not the signal the policy uses.** `hosted_by_reason` classifies on `entry.local_client_access` (`ring/hosting.rs:1108`), which is set at `cache.rs:2128` and never cleared. The hosting policy consults the age-gated `has_recent_local_client_access` (`cache.rs:2143`) instead. `cache.rs:4319-4340` is an existing test that proves these two diverge over time. As written, `local_access` monotonically absorbs contracts across a node's uptime while `routed` drains into it, so the gauge stops meaning what its own doc says and disagrees with the policy it's meant to describe. Please switch to the recency-gated signal.

**3. A restart silently zeroes the `abandoned` bucket.** The restore path sets `abandoned_at: None` (`cache.rs:2888`), so previously-abandoned contracts come back classified as `Routed` — whose doc says they "arrived through a routed GET/PUT" (`hosting.rs:259-261`), which is false for a restored contract. `HostingCause` has a `StartupRestore` variant precisely so a bulk reload can't masquerade as live demand. Either add the equivalent bucket or drop that claim from `Routed`'s doc; I'd prefer the bucket.

**4. Cross-reference the two taxonomies.** `HostingCause` (`cache.rs:729`) and `HostingReason` (`hosting.rs:243`) are genuinely different things — provenance-at-admission versus current-demand — so this isn't duplication. But `host_begin`'s doc at `router.rs:301` describes itself as answering "why is this peer hosting this contract" in aggregate, which is verbatim what `HostingReason` claims. Please cross-reference both rustdocs in both directions, and consider a stock-flavoured name (`HeldFor`, `HostingDemandClass`) so the next reader doesn't conflate them.

### Performance

**5. The hosted-set walk runs twelve times per collection cycle.** `otel_metrics_snapshot()` (`node/network_status.rs:146-149`) invokes the hosting-reason provider unconditionally. Twelve observable callbacks call `snapshot()`; only two (`otel.rs:898`, `otel.rs:918`) read `hosting_reasons`. Each walk holds the `hosting_cache` read lock doing three DashMap lookups per contract, so a collection blocks cache writers six times more than it needs to. This is also what the comment at `p2p_impl.rs:527-530` says it's avoiding. Either memoize the snapshot for one collection pass (the callbacks fire back-to-back on the reader thread) or give `hosting_reasons` its own accessor used only by the two gauges that need it.

**6. Queue metrics are suppressed by unrelated providers.** `otel_metrics_snapshot()` `?`-chains `NETWORK_STATUS`, `RING_STATS_PROVIDER` and `HOSTING_REASON_PROVIDER` before reading `fair_queue_stats()`, which is a free function over always-available statics. Any path where the ring provider isn't registered silently zeroes `freenet.contract.queue.*` too. Make the fields individually optional.

### Duplication

**7. Seven of the nine hot-path mirrors shadow counters that already exist.** In `transport/metrics.rs`, each new `otel::record_*` call sits directly above a `fetch_add` on a counter that's already there:

```rust
pub fn record_nat_traversal_attempt(&self) {
    crate::tracing::otel::record_nat_traversal("attempt");        // added
    self.nat_traversal_attempts.fetch_add(1, Ordering::Relaxed);  // already exists
}
```

An observable counter reading `nat_traversal_attempts` needs no call site at all and can never drift. This is exactly the reasoning your own comment gives for the new `cumulative_packets_*` counters — please apply it to the four NAT variants, both transfer outcomes, and `record_op_result` too. Keep `record_rtt_ms` and `record_cwnd` as sync calls; those are histograms and can't be reconstructed from a sum-and-count atomic.

This matters beyond tidiness: manually-mirrored telemetry counters are a documented repeat offender in this repo (`.claude/rules/bug-prevention-patterns.md`, incidents #4009 and #4010, where mirrors rotted silently and counters read zero for months). Removing the mirrors removes the whole failure class, and it makes item 9 mostly moot.

### Tests

**8. Nothing ever exercises an instrument.** `register_metrics()`'s only caller is `otel.rs:438` inside `init()`, and `init()` always returns early under `cfg(test)` — your own `init_refuses_to_start_from_a_test_process` asserts exactly that. So none of the twelve callbacks, no instrument name, no unit, and no attribute is ever executed in CI. That's load-bearing rather than cosmetic: a panic in any callback kills the `PeriodicReader` thread and stops *all* metrics permanently, and `report_export_failure` won't fire because that's an export-side signal.

The seam already exists — `build_provider` is `pub(crate)` and already has two `#[tokio::test]`s. Please add a test that builds a provider with `ManualReader` + `InMemoryMetricExporter`, calls `register_metrics()`, collects once, and asserts the expected instrument names and attributes appear.

**9. The pin test can't catch what it's for.** `every_sync_instrument_still_has_its_hot_path_mirror` (`otel.rs:1499`) does `source.contains(call)` against the whole file. `AGENTS.md:81-96` requires source-scrape pins to be bounded to the calling function via the `fn_body()` helper, and to be mutation-tested when written. Concretely: swap the literals between `record_transfer("completed")` (`metrics.rs:199`) and `record_transfer("failed")` (`metrics.rs:284`) and the pin stays green while success and failure invert on the wire. The mapping is correct today, so this is latent rather than an active bug. If item 7 lands, most of this test disappears; whatever remains needs bounding and a mutation check.

### Robustness

**10. A poisoned lock silently kills all twelve metrics forever.** `network_status.rs:141` does `NETWORK_STATUS.get()?.read().ok()?`. That's a poisonable `std::sync::RwLock`, and the existing writers already tolerate poison field-by-field — but here the `?` makes every ring and queue metric vanish permanently with no log at all. `unwrap_or_else(|e| e.into_inner())` matches what the SDK itself does; at minimum log once on that branch.

**11. The shutdown comment points at something that panics.** `otel.rs:431-436` suggests keeping the provider in a `OnceLock` and calling `shutdown()` from the graceful shutdown path in `bin/freenet.rs`. Done the obvious way from an async fn that panics — as your own tests demonstrate, which is why they wrap it in `spawn_blocking` (`otel.rs:1771`, `otel.rs:1799`). Please carry that caveat into the comment.

### Docs

**12.** The PR description says `<audience>` is the collector's `host:port`; the code hashes `{host}:{port}{path}` and base58s the first 16 bytes (`otel.rs:207-219`). Both docs are correct — only the PR body is stale.

**13.** The description doesn't mention that `opentelemetry-otlp` and `opentelemetry_sdk` move from `optional = true` to non-optional, so every build and cross-compile target now links the OTLP exporter, `prost` and `opentelemetry-proto` even with the feature off. The design doc covers it honestly; please state it in the PR body too, with the measured binary-size delta.

## Gating criteria for re-review

Re-request review once all of these hold. If they do, I expect one round rather than several:

- [ ] All CI green on the exact head SHA (not an earlier one — any rebase or amend invalidates a prior run).
- [ ] `cargo test -p freenet --lib otel:: config:: ring::hosting::` green under **plain `cargo test`**, not only `nextest`. CI runs process-per-test, so it structurally cannot see cross-test interference through process-global state, and this PR has two such statics (`INSTRUMENTS`, `EXPORT_FAILING`). Please say in the PR that you ran it this way.
- [ ] The new collection test asserts at least one real exported datapoint, by name and attribute.
- [ ] Any remaining source-scrape pin has been mutation-tested: apply the exact regression it names, confirm it fails, revert.
- [ ] Still zero `unwrap`/`expect`/`panic!` in production paths. You're at zero now — worth keeping.
- [ ] PR body updated for items 12 and 13.

## Tooling

You may already be using these, but in case not — the review agents and skills we use on this repo are public:

```
/plugin marketplace add freenet/freenet-agent-skills
/plugin install freenet@freenet-agent-skills
```

`/freenet:pr-review` is the one worth running before you re-request: it triages the risk tier and runs the same code-first / testing / skeptical / big-picture lenses I used here, so most of the above would surface locally first. `/freenet:systematic-debugging` is useful if any of the classifier changes turn out to be less mechanical than they look. Repo: https://github.com/freenet/freenet-agent-skills

## For PR B (collector auth), so none of it is lost

Not asking you to act on these yet — they'll need revisiting once I've answered the collector question, and some may become moot.

- The token authenticates the *sender*, not the *request*. The signed payload is `freenet/{pubkey}/{audience}/{timestamp}` (`otel.rs:159`); the body isn't covered. With the nonce removed, freshness rests entirely on a collector-side clock check whose tolerance isn't specified in code or either doc, and no replay cache is described. Since the body is unsigned, a replayed token can carry a *different* `freenet.node.pubkey` than the one it authenticates.
- `endpoint_problem` (`otel.rs:116-127`) accepts plaintext `http://` silently, and the SDK default is `http://localhost:4318`. Worth warning or refusing when `auth-mode = freenet` resolves to a non-loopback plaintext endpoint.
- `audience_of` (`otel.rs:217`) omits the scheme, so on the standard OTLP port `https://` and `http://` hash identically — and reqwest's `cross_host` check ignores the scheme too, so an https→http redirect retains the `Authorization` header. Also worth setting `redirect::Policy::none()`; a single-target exporter has no use for ten hops.
- `truncated_body` (`otel.rs:279-283`) names "a collector may echo back our bearer token" as the thing it mitigates, but a token is ~182 chars including `Bearer `, comfortably inside the 256-byte limit. `freenet service report` uploads logs wholesale, so please redact on `Bearer freenet/` rather than truncate.
- `response.bytes()` (`otel.rs:330`) buffers the whole response with no size cap. Time-bounded by the client timeout, not size-bounded.
- The construction itself checked out, for what it's worth: no chosen-message oracle (every signed field is node-determined), nothing in `transport/` verifies with that key so a signature can't be replayed as a transport message, clamping is genuinely consistent between `xeddsa` and `x25519_dalek`, and the nonce uses OS entropy. The separate question I need to answer is whether the transport identity key should be signing assertions at all.
- One thing to flag regardless of the split: `freenet.node.pubkey` and `freenet.node.fingerprint` ship in every export batch even when `auth-mode = disabled`, because `identity_attributes` is called before the auth-mode match (`otel.rs:423`). That's deliberate and documented at `docs/otel-metrics.md:125-133`, but the justification for the disabled default — "must not ship a signed assertion of this node's identity" — reads as though disabled withholds the identity, when it withholds only the signature. Worth saying plainly in both docs.

Happy to answer questions on any of the above. The exporter itself is in good shape and I don't expect PR A to take many rounds.

[AI-assisted - Claude]

sanity:
## Review

Ran this at Full tier on head `60d16a40c`: four independent blind lenses (code-first correctness, testing adequacy, adversarial bug-hunt, goal alignment), each reading the checked-out code rather than the diff text, then I verified every finding below against the source myself before repeating it. Apologies for the week of silence since the 17 Aug review; you turned that round quickly and it should not have sat.

**First, the good news, because it is most of the news.** All thirteen items from the 17 Aug review are addressed at head. I checked each one in the code rather than trusting the description:

- The CI-blocking semantic conflict is fixed (`ring/hosting.rs:4655` now passes `HostingCause::Other`), and CI is green across the board.
- `hosted_by_reason` now classifies on the recency-gated signal: `cache.rs:2607-2612` derives `recent_local_client_access` from `local_client_last_access` aged against `SUBSCRIPTION_LEASE_DURATION`, the same window `hosting.rs:2825-2829` uses.
- `HostingCause` and `HostingReason` are cross-referenced in both directions (`hosting.rs:226-234`, `cache.rs:755-765`, `router.rs:297-306`).
- `otel_hosting_reasons()` / `otel_ring_stats()` / `otel_status_scalars()` are three independent accessors, so an unregistered ring provider no longer zeroes the queue metrics.
- Only two hand-placed sync call sites remain, both histograms (`transport/metrics.rs:354`, `:386`), and grep confirms no other `tracing::otel::` call site exists in the tree. The seven mirrors are gone.
- The poison path is `unwrap_or_else(|p| p.into_inner())` with a `Once`-guarded WARN (`network_status.rs:143-158`).
- `method_body` is genuinely bounding, and I checked the mutation reasoning independently: it anchors on the four-space `\n    }\n`, both pinned bodies contain only deeper-indented closers, and the cross-file `include_str!` means the pin's own literal cannot satisfy it. Moving `record_rtt_ms` to another function does fail it.

The dependency work is better than the description claims. Net lock delta versus the merge base is **+`xeddsa`, +`convert_case`, and `-reqwest 0.13.3`**: `reqwest` goes from two majors to one, and `aws-lc-sys` / `aws-lc-rs` / `cmake` / `nasm` are absent on both sides. No tests were removed or weakened anywhere in the diff.

---

## One real bug

### `HostingReason::Restored` is sticky for the process lifetime, which inverts what the gauge reports

Two lenses found this independently and I confirmed it by grep.

`hosting.rs:1152-1153` classifies on `!seeded_this_run`. That flag has exactly two writers in the whole cache:

- `cache.rs:2116` — the not-cached insert branch of `record_access`, sets `true`
- `cache.rs:2974` — `load_persisted_entry_with_demand`, sets `false`

The existing-entry refresh branch (`cache.rs:2041-2085`) updates `last_accessed`, `recency_seq`, `read_count`, `abandoned_at` and `keep_score`, and never touches `seeded_this_run`. So a contract reloaded at startup that subsequently serves any amount of routed GET/PUT traffic stays in `reason="restored"` until the process exits.

The rustdoc at `hosting.rs:280-281` says the bucket means "Reloaded from persisted hosting metadata at startup **and not touched since** (`!seeded_this_run`)". The second half is not what the flag means.

This matters operationally rather than cosmetically. Every production node restarts, and after a restart `restored` absorbs the entire persisted hosted set permanently while `routed` only ever counts contracts admitted since boot. `freenet.node.contracts.hosted{reason="routed"}` therefore systematically under-reports transit hosting, and the bucket's own stated justification ("A bulk reload must not read as live demand") comes out backwards in practice: live demand reads as bulk reload.

`hosted_by_reason_partitions_the_hosting_cache` passes because it loads the persisted entry at `hosting.rs:4682-4692` and asserts immediately, never issuing a `record_access` against it afterwards. A test that loads a persisted entry, then drives one routed access through it, then asserts it has moved to `Routed`, would fail against the current code and is the regression test for this.

Either fix the predicate (`!seeded_this_run && read_count == 0` works, since `read_count` *is* reset on reload at `cache.rs:2969`) or drop the "not touched since" claim and rename the bucket to say it is provenance rather than recency. I would prefer the predicate, because provenance is already what `HostingCause::StartupRestore` records.

---

## Should fix

### `freenet.connect.attempts` does not count connection attempts

`otel.rs:938` describes it as "Connection attempts made since startup". `NetworkStatus::connection_attempts` has exactly one writer in the tree, inside `record_gateway_failure` (`network_status.rs:715-718`). Nothing increments it on an initiated or successful connection. The series is gateway connection *failures*.

The local dashboard mislabels it the same way, so this is inherited rather than introduced. But an OTel instrument name and description is an external contract that operators will build alerts on, and "attempts near zero" versus "failures near zero" are opposite readings of node health. Rename to `freenet.connect.gateway_failures`, or fix the description to say what it counts.

### The export test asserts 10 of the 20 instruments, and its closing comment overstates what it establishes

I counted: 22 `freenet.*` string literals, of which two are resource attributes, so 20 instruments (18 `with_callback` plus the 2 sync histograms). `instrument_callbacks_export_named_datapoints` asserts 10 names (`otel.rs:1876-1887`) and 8 attribute pairs.

Never asserted: `freenet.process.memory.rss`, `.ring.connections`, `.node.contracts.hosted`, `.node.contracts.hosted.bytes`, `.connect.attempts`, `.operation.results`, `.ring.lattice.neighbor`, `.ring.lattice.neighbor.distance`, `.ring.lattice.probes`, `.contract.updates`. Rename any of them, or drop `reason=` from the two hosted gauges or `op=`/`result=` from `freenet.operation.results`, and nothing fails. That includes the `HostingReason::as_str` values, whose own rustdoc calls them "a metrics contract" where "renaming one silently empties a panel".

The part worth flagging is the comment at `otel.rs:1926-1929`, which says what the test *does* guarantee about those instruments is "their callbacks RAN, without panicking, which is what keeps the PeriodicReader thread alive". It does not. Each of those bodies is `if let Some(x) = otel_ring_stats() { ... }` (`otel.rs:889`, `:905`, `:925`, `:940`, `:955`, `:983`, `:1000`, `:1019`, `:1036`). Under nextest's process-per-test the providers are unregistered, the closure returns `None`, and the guarded body — which is where the indexing and the `as u64` casts live — never executes. A panic in `reasons.count(reason)` would not be caught.

The reasoning for *why* those instruments are excluded is correct and I would not want it dropped. The fix is to close the gap rather than restate it: give `register_ring_metrics` and the status-sourced observables a source closure parameter the way `meter` is already passed, hand the test a fixture, and then every name and attribute is assertable deterministically without touching a global.

### The auth-mode to signer mapping is executed by no test

`otel.rs:424-427` maps `OtelAuthMode::Freenet` to a signer and `Disabled` to `None`. `init()` returns at `otel.rs:386` under `cfg(test)`, so this match is unreachable from the test binary, and `disabled_auth_mode_sends_no_authorization_header` builds `OtlpHttpClient` directly via `test_client(&keypair, false)`, bypassing it. Invert the two arms and the entire suite still passes, while the privacy default silently becomes its opposite. Extracting `fn auth_signer_for(config, keypair) -> Option<_>` and asserting both arms is a two-line test.

### `truncated_body`'s 256-byte window is not the mitigation it claims

`otel.rs:278-283` gives the rationale that a collector "may echo back headers (including our bearer token), so it is truncated rather than logged whole". A `freenet` token is roughly 175 characters, about 182 with `Bearer `, so it fits inside 256 with room to spare. `report_export_failure` logs that body at WARN (`otel.rs:336`), and `freenet service report` uploads node logs wholesale. Upstream deliberately keeps the same body at DEBUG for exactly this reason.

Redact on `Bearer ` / `freenet/` rather than truncate, or drop the stated rationale if the window was only ever a log-volume bound. This was on the 17 Aug list under "PR B"; see the scope note below for why it is worth resolving either way.

### Unbounded response body read

`otel.rs:330` calls `response.bytes()` with no size cap, bounded only by the 10s client timeout. Only 256 bytes are ever logged and the SDK only needs enough to decode a partial-success message. A 64 KiB cap costs nothing.

---

## Nits

- **`ring/hosting.rs:1122`** says "six buckets"; `HostingReason::ALL` has seven.
- **`otel.rs:1811`** says "the twelve callbacks" and **`network_status.rs:175`** says "twelve times a cycle"; `grep -c with_callback` returns 18. The rationale for splitting the accessors is stronger than the number a reader would reason from.
- **`network_status.rs:185`** has a dangling intra-doc link to `OtelMetricsSnapshot`, which no longer exists (renamed `OtelStatusScalars` in this PR). Rustdoc will warn.
- **`otel.rs:118-121`**'s "the SDK will silently fall back to `http://localhost:4318`" is true for an unparseable *env-var* endpoint, but a bad config-file endpoint returns `InvalidUri` and the exporter fails to start. `otel.rs:417-421` applies the message to the config endpoint first, so the common case gets the wrong remediation.
- **`fair_queue.rs:432-434`** says these "four are exported as OTel observable counters"; `HIGH_WATER` is deliberately a gauge, as `otel.rs:1085-1087` says. Three and one.
- The `store`-to-`fetch_max` change at `fair_queue.rs:441-443` is behaviour-equivalent today (the sources are only ever `saturating_add`-ed) and I have no objection, but it touches a path the local dashboard reads and is unmentioned in the description. One sentence would do.
- The measured binary-size delta in the description is against `7c5cc4be1`, roughly 30 commits before this PR's merge base, on `aarch64-apple-darwin`. Nodes run musl. Either re-measure and name the target or drop the exact figure.
- The whole hosting-reason feature — a new public enum, `Ring::hosted_by_reason`, a provider registration, and an O(hosted) walk under the cache read lock twice per collection cycle — is in the design doc but absent from the PR description. A reader of the description alone would not know this PR touches `ring/hosting` at all.

A note on something I checked and am **not** billing to you: one lens flagged the `.claude/rules/bug-prevention-patterns.md` citations in your comments as pointing at content the in-repo file does not contain. That is true of the file, but the same citation style is used at 40-plus pre-existing sites across the crate. It is repo convention and a housekeeping problem on our side, not a finding against this PR.

---

## On the scope question, which is not mine to settle

The 17 Aug review proposed splitting the collector auth into a separate PR, pending a design decision Ian has not made yet about whether Freenet is heading toward a federated or shared-collector model. Your reply in the PR body argues the split does not fit what `otel-auth-mode` is. That is a fair argument and it is squarely a maintainer call, so I am flagging it for him rather than deciding it by merging.

What I can add is the measurement, so the decision is made on numbers instead of impressions. The auth code is additive and self-contained: about 656 of 3,633 non-lockfile added lines (roughly 18%), 13 of 31 new tests, and zero whole files. Removing it would not change the node wiring, since `node.rs:831` still needs the keypair for the identity resource attributes, and `OtlpHttpClient` stays either way because it is what trims the dependency graph and adds the export-failure WARN the SDK does not emit. Only the `signer`/`pubkey_b58` fields and the 13-line header block at `otel.rs:292-300` are auth. Notably, the entire net dependency delta is auth: without it, `xeddsa` and `convert_case` go away and the lock delta becomes purely the `-reqwest 0.13.3` improvement.

The other half worth naming: three of the security items parked under "For PR B" on 17 Aug are unaddressed at head while the auth code is still in this PR. The signed payload does not cover the body and no clock tolerance or replay bound is specified in code or either doc; `endpoint_problem` (`otel.rs:116-127`) accepts plaintext `http://` silently even when `auth-mode = freenet` resolves a non-loopback endpoint; and no `redirect::Policy` is set on the client (`otel.rs:660-665`) while `audience_of` omits the scheme, so an https-to-http redirect retains the `Authorization` header. If the auth stays here those need resolving here, because the PR they were deferred to would not exist.

## What would land the exporter

Independent of the scope question: the `Restored` classifier is the one thing I would call a genuine bug rather than a polish item, and it has a small fix and an obvious regression test. The two test gaps (instrument coverage, auth-mode mapping) are each a handful of lines. The nits are nits.

The branch is also 9 commits behind `main` now, so it needs a rebase and a fresh CI confirmation on the new head before anything merges.

Thank you for the persistence on this. The instinct that produced the observable-counters-over-mirrors design is the right one and it retires a failure class this repo has been bitten by twice, and the honesty in the design doc about its own risks is genuinely unusual.

[AI-assisted - Claude]

sanity:
@skandragon — an answer on the split, and an apology for how long it took to arrive.

**Short version: please do split it, and you can start now without waiting on the design question.**

The split was proposed on 2026-08-17 pending a decision about whether Freenet is heading toward a federated/shared-collector model. You pushed back in the PR body rather than in a comment, so the thread never came back around, and it has sat since. That's our failure to close a loop, not yours.

**Here is why the split stands on its own, independent of that design question** — which means you don't have to wait for it:

- **Reviewability.** The auth work is ~656 of 3,633 non-lockfile added lines (~18%), 13 of 31 new tests, and **zero whole files** — so it cuts cleanly. It also carries the **entire net dependency delta** (+xeddsa, +convert_case; the exporter half is net-negative, −reqwest 0.13.3). A reviewer can judge "export node metrics via OTel" on its merits far more easily when it isn't also "add a new authentication scheme".
- **Three security items are unaddressed while the auth code stays in PR A** — unsigned body / no replay bound, plaintext `http://` accepted under `auth-mode=freenet`, and no `redirect::Policy` set. Those were parked "for PR B". If the auth stays in PR A, they stop being parked and become blocking, because they'd be shipping.
- **PR A can land now.** The exporter half is in good shape (details below). The design question only gates the auth half, so splitting converts an indefinite block into one shipped PR plus one waiting on a decision.

**On the exporter half — all 13 of the 2026-08-17 items are addressed at head**, verified independently by two reviewers, including the CI-blocking `record_contract_access` arity fix. No tests were removed. The dependency delta is better than the PR body claims.

**One real bug to fix before PR A lands:** `HostingReason::Restored` is sticky for the process lifetime. `seeded_this_run` is written at only two sites — fresh insert (true) and reload (false) — and the existing-entry refresh branch never touches it. We checked `main` and it uses a `read_count == 0 && seeded_this_run` conjunction that this PR drops. Since every node restarts, `restored` would absorb the transit population and `freenet.node.contracts.hosted{reason="routed"}` would read near-zero fleet-wide. The rustdoc claim "and not touched since" isn't true as written. The partition test passes only because it never accesses `restored` after loading it.

**Smaller items for PR A**, none blocking on their own:
- `freenet.connect.attempts` counts gateway **failures**, not attempts — `connection_attempts` has exactly one writer, inside `record_gateway_failure`. Pre-existing, but this PR promotes it to an external metrics contract, which is the moment to fix or rename it.
- The export test asserts 10 of 20 instruments; renaming any of the others, or dropping a `reason=`/`op=`/`result=` attribute, fails nothing. Its closing comment says the callbacks "RAN, without panicking" — they don't: the bodies are `if let Some(x) = otel_hosting_reasons()`, and in a fresh test process no provider is registered, so the guarded code never executes.
- The auth-mode → signer mapping is executed by no test (`init()` returns early under `cfg(test)`); inverting the two arms passes the whole suite.
- `truncated_body`'s 256-byte window is described as a bearer-token-leak mitigation, but a Freenet token is ~182 chars including `Bearer ` — it fits inside the window.
- Doc drift: "six buckets" vs the 7-variant `ALL`; "twelve callbacks" vs 18; a dangling intra-doc link to `OtelMetricsSnapshot` (renamed to `OtelStatusScalars`).

**To be explicit about authority:** the *reviewability* case above is ours to make and we're making it. The **federated/shared-collector design question is @sanity's**, and it still gates PR B — nothing here pre-empts it. If he'd rather keep it as one PR, that overrides this.

Thank you for your patience, and for pushing back rather than silently complying — that's what surfaced that the thread had stalled. We've just finished a sweep of a 59-PR backlog and found that outside-contributor CI had never been approved (58 workflow runs pending), so several of your PRs were not merely unreviewed but unrunnable. That's fixed, and automation is going in so it doesn't recur.

[AI-assisted - Claude]

sanity:
@skandragon — pushed onto your branch rather than sending you round again. Two commits on top of `f483eee7f`; nothing of yours was reverted and the design is unchanged.

**The auth stays.** Your pushback was right and the reasoning behind dropping it was wrong. The case for removal rested on there being no collector that ingests from nodes it doesn't operate — and that's precisely what you're offering. A static token can't serve it, because issuing one per node *is* the signup you're trying to avoid. Node-identity signing is the answer to that problem, so `otel-auth-mode` stays and the split is withdrawn. Sorry for the round trip; the missing fact was ours to ask for, not yours to volunteer.

Keeping it makes the three items parked "for PR B" blocking here, since they'd now ship. All three are done:

- **The signature covers the request body**, as a hash that isn't transmitted and which the collector recomputes from what it received. Without it a token authenticated only "this node, this collector, this second", so anyone holding one could attach it to metrics of their own invention — the spoofing the scheme exists to stop. Token stays five fields.
- **A replay bound is specified**: `REPLAY_WINDOW`, 300s. Worth a look from you specifically, since you'll be on the collector side: `docs/otel-metrics.md` now carries a single normative list of all five obligations in order — verify signature, verify body hash, verify audience, enforce the clock window, refuse a repeated `(pubkey, timestamp, body hash)`. It previously listed only the last two under "must enforce", so an implementer skimming would have built a collector that accepts any body from anyone holding a fresh token.
- **No credential goes out in cleartext**, yours or an operator's, to a non-loopback collector. Fails the export with a `WARN` naming the reason, and says so at startup rather than one interval later.

Two ambient mechanisms would have defeated that last check, since it can only inspect the URI we aimed at, so both are now off and pinned by a test: reqwest replays redirects with the original headers, and honours `HTTP_PROXY` by default with no loopback exemption — a corporate proxy would have taken an export aimed at `http://localhost:4318` off the machine with the token attached.

The guard also isn't `Authorization`-only any more. Every header from `OTEL_EXPORTER_OTLP_HEADERS` counts, since `x-honeycomb-team` / `api-key` / `dd-api-key` are all credentials too, and those values get redacted out of logged error bodies by value rather than keyword.

**One real bug.** `HostingReason::Restored` classified on `!seeded_this_run`, which nothing ever clears, so after any restart it absorbed the whole persisted hosted set permanently and `reason="routed"` would have read near zero fleet-wide. Now also requires `read_count == 0`, with a regression test that drives a routed access through a restored contract.

**Test coverage.** Your nine ring-sourced callbacks are guarded by `if let Some(x) = <source>()`, so in a fresh test process the bodies never ran and half the instruments asserted nothing. `RingSources` puts those accessors behind function pointers so the export test injects fixtures — all twenty instruments and every attribute value are asserted now.

Reviewed at Full tier on the new head: four independent Claude lenses plus `codex review`. Codex caught a real one I'd introduced, and the adversarial lens caught the `HTTP_PROXY` hole.

**Yes please on the docs** — especially the privacy section, which is the part we'd do worst. And thank you for keeping this rebased for two weeks while we sat on it.

[AI-assisted - Claude]

PR #5179, opened 2026-08-05
feat(update): attach the originator's target list to broadcasts (#5147)

### LINKED ISSUE #5147: Broadcast is mesh re-fan-out, not tree forwarding: each node receives every update ~18x. This is the dominant bandwidth cost.
## The answer, from telemetry we already have

Two existing measurements settle the question #5062 was opened to answer, without any new instrumentation:

| measurement | source | value |
|---|---|---|
| fan-out degree (`broadcasted_to`) | `update_broadcast_emitted`, pre-existing | median **17**, mean 24.7, p90 58 |
| duplicate deltas per state-changing delta | `network_efficiency_v1` receiver dedup counters (#5091, 0.2.117) | **18.3** |
| payloads delivered per state-changing one | same | **22.4** |
| share of received contract bytes that change nothing | same | **97.0%** |

Measured 2026-08-04 across 860 nodes on 0.2.119 with >=2h uptime, 11,283 node-hours.

#5062 framed the two possibilities as: **(A)** one node fans out to its ~17 co-hosts, or **(B)** ~17 nodes each re-fan-out to their own co-hosts. Those predict very different receiver-side duplication:

- Under (A) each recipient receives **one** copy. Dedup would be ~0.
- Under (B) each recipient receives **~17** copies, one from each co-host that got it and re-sent. Dedup ~16-17.

**Measured dedup is 18.3 against a median fan-out of 17.** That is (B).

## What this means

The broadcast mesh re-fans-out rather than forwarding along a tree. Every co-host that receives an update re-broadcasts it to every other co-host, so each node receives essentially the same payload from all of its co-hosts. The receiver's dedup cache catches them — but **after the bytes are already on the wire**, which is why this shows up as ~1 GB/node/day of byte-identical duplicate deltas.

This is the dominant bandwidth cost on the network. For scale, the queue-scheduling fix in 0.2.119 (#5108/#5120) was a real win on its own axis (lane misclassification 59.5% -> 0.0%, head-of-line incidents down 86% per node-hour) and moved the wasted-byte share only from 99.0% to 97.0%.

## Caveat, stated properly

Three re-emission funnels inflate the dedup count beyond distinct senders: the retry path at `p2p_protoc/broadcast.rs:330` (up to 4 events per apply), the stash-recheck at `:388`, and the interest-flush at `op_state_manager.rs:2183`. So some of the 18.3 is one sender retrying rather than 18 senders.

Even discounting 4x for retries, ~14 independent senders remain. The conclusion is (B) either way — but anyone building the fix should measure the retry contribution rather than assume all 18.3 are distinct peers, because it changes how much a suppression fix can actually recover.

## Proposed direction

Suppress re-broadcast to peers the broadcast has already reached. #5003 established exactly this pattern for the *summary* ("stop re-sending the summary to peers the broadcast already covered"); the same needs to apply to the payload.

Options worth weighing, roughly in order of increasing invasiveness:

1. **Carry covered-peer information with the broadcast** so a relaying node can skip peers already served. Bounded-size, so it degrades gracefully on large co-host sets.
2. **Suppress by recency**: do not re-broadcast a payload to a peer we have seen the same payload from, or sent it to, within a short window. Cheap, local, no wire change; the receiver-side dedup cache already proves the payload-identity check is workable.
3. **Spanning-tree / gossip suppression** over the co-host set. Most effective, most invasive, and needs care around partition healing.

Note #3911 ("UPDATE relay forward path lacks HTL / loop avoidance") is adjacent and may share a mechanism.

## Measurement notes for whoever picks this up

Getting these wrong cost several rounds of corrections on 2026-08-02/03:

- **uptime = `5 + 30 * eff[0]` minutes.** The block snapshots its own delivery counters before enqueueing itself, so `eff[0]` counts only PRIOR exports; and the first export lands at t=5min because `ring.rs:1636-1637` consumes tokio's immediate first tick while `:1651` initialises the tick counter to `5u8`. Neither `eff[0]*30` nor `(eff[0]+1)*30` is right.
- **Key nodes by the peer-id prefix before `@`.** The full `peer_id` embeds `addr:port`, so a reconnect on a new port looks like a new node (~15% inflation).
- **Skip nodes with <2h uptime.** A restart resets summary caches and moves the headline metric ~5pp on its own; an upgrade is a restart.
- **Compare same-window, not against an older baseline.** Comparing 0.2.119 against a pre-release number suggested duplication had got *worse* (17.5 -> 22.4); the same-window control on 0.2.118 nodes read 24.7, i.e. 0.2.119 is better. Cross-window comparison would have produced a false regression report.
- Ratios are reliable; absolute per-node-day rates depend on the denominator above.

A reusable analysis script lives at `~/code/admin/nova/freenet/post-release-telemetry.py` on nova (telemetry is at `/mnt/media/freenet-telemetry/logs.jsonl`, needs sudo, purged after ~3 days).

Refs #5062, #5091, #4956, #5003, #3911

[AI-assisted - Claude]

sanity:
## Problem

Contract broadcast is a mesh re-fan-out. `A` applies an update and sends it to its advertised co-hosts `B`, `C`, `D`. `B` applies it and re-broadcasts to ITS co-hosts — which include `C` and `D`, who already have it. Every node does this simultaneously, so on the 0.2.119 fleet each node received **~18.6 byte-identical copies of every update, 74.6% of all received contract bytes** (#5147).

`B` cannot currently tell that `C` already has it. The information that would say so (`sender_summary_bytes`) rides ON the duplicate itself, so it always arrives too late to prevent it.

## Approach

The originator attaches to its broadcast the list of peers it is sending to; a relayer excludes listed peers from its own re-fan-out.

**Exact list, not a Bloom filter.** This supersedes the covered-set-as-Bloom direction (commits `8afecb599`, `2c8b070de`, retired — not built on). At the median fan-out of 17 an exact list of 8-byte hashes is no larger than the Bloom, and it has no false positives, so the salt / saturation guard / union semantics that existed solely to manage those false positives are all unnecessary. A false positive here means a peer that genuinely needed the update silently does not get it until the ~5-minute heartbeat, so removing that failure mode outright is worth more than the bytes.

**One hop only**, which is also the security rule. A relayer does not forward the list and does not union its own targets into it (measured hop-2 suppression is already 0.99, so multi-hop buys nothing). The list is honored ONLY from the message that delivered the payload, so a malicious relayer cannot attach a fabricated "everyone is covered" claim to suppress third-party fan-out. That is enforced structurally rather than by a runtime check: **no other variant has the field**, and `only_payload_bearing_variants_carry_a_target_list` parses the enum to keep it that way.

**Transaction-seeded hashing.** Hashes are `blake3(tx_id ‖ pub_key)[..8]`, following `VisitedPeers`, whose rustdoc documents the topology-inference rationale for tx-seeding. It buys a second property free: a list lifted from one broadcast and replayed onto another resolves to nothing.

### Provenance plumbing — the choice, and why

`NodeEvent::BroadcastStateChange` is emitted by the executor on ANY committed write and carries no idea whether the write came from a local client or an inbound broadcast; nothing between the receive driver and the fan-out carries networking context.

Threading it explicitly through `ContractHandlerEvent::UpdateQuery` is ~10 signatures and 2 trait methods × 3 impls, **plus three things worth naming because they are easy to miss**: the deferral path re-runs the upsert (so the value must live in `DeferralCtx` and be replayed), there is a second `BroadcastStateChange` emitter at `executor_impl.rs` that bypasses `commit_state_update`, and there is a second `UpdateQuery` producer at `node.rs` (the `ResyncResponse` arm) that bypasses `update_contract` entirely. That is a large diff through the WASM layer with three places where getting it wrong is silent.

Instead the apply parks its claim on a short-TTL store keyed by contract, and the fan-out takes it. Ordering is sound by construction: the write happens before the `UpdateQuery` enters the contract handler, and the read can only happen after that query committed.

**The race, and why it can only under-suppress.** Two different updates to the same contract can be in flight, and the entries are keyed by contract. `register` therefore **narrows** rather than replaces: coverage intersects, and a disagreeing sender is dropped. Every peer in the intersection was covered by *both* live applies, so whichever apply the fan-out belongs to, each suppressed peer genuinely had the state. A mixed-up entry can only suppress fewer peers — today's behaviour — never withhold an update. Three things keep that from being merely decorative: the fair queue serialises WASM merges per contract key; duplicates are dropped by `broadcast_dedup_cache` *before* `update_contract`, so ~18 copies of one update register once between them; and an apply that changes nothing discards its entry rather than leaving it to expire (no-change is the common case — 97% of received contract bytes change nothing).

**Residual, stated rather than hidden:** commit paths that emit `BroadcastStateChange` without going through `update_contract` (the executor-internal replay, the second emitter) can take an entry belonging to a concurrent relay apply of the same contract. That one IS over-suppression. It needs a same-contract PUT-or-replay concurrent with a relayed update inside one contract-handler round trip; it is TTL-bounded and healed by the heartbeat.

### The dead sender-exclusion, revived

`get_broadcast_targets_update` has had sender- and self-exclusion filters for a long time and **both were structurally dead**: the re-fan-out call site had no sender to pass and handed in its own address, so `is_local_update_initiator` was always true there. A relayer echoed every broadcast straight back to the peer that had just sent it — a 100%-wasted send on every relayed broadcast, and one the target list cannot fix, since a peer never names itself on its own list. Replacing the `&SocketAddr` sentinel with a `BroadcastOrigin` makes both filters live. Self-exclusion is now unconditional; broadcasting to ourselves is never correct.

Called out separately so its share of the measured effect can be attributed. Say the word if you would rather it were its own PR.

## Testing

**The simulation earned its keep — it failed first, and the failure was a real bug.**

First run: the feature made traffic *worse* (treatment sends 528 vs control 460, redundant deliveries 400 vs 321). A control-vs-control run gave byte-identical numbers, clearing the rig. Cause: a fully-suppressed fan-out has an empty target set, and `handle_broadcast_state_change` read empty as the no-subscriber FAILURE case — three retries with backoff, then a stash. Each retry re-resolved targets after the coverage claim had been consumed, so it suppressed nothing and re-broadcast the whole co-host set one backoff later. **The mechanism defeated itself and would have shipped looking plausible**: every unit test passed, and the only symptom was a bandwidth number nobody had a baseline for. Second commit fixes it; `fully_covered_fanout_returns_before_the_no_target_retry` pins the ordering.

After the fix, 1 gateway + 6 subscribing peers, 20 CRDT updates, identical seed, arms differing by exactly one `enable_broadcast_target_list()` call:

```
control:   sends=460 summary_skips=317 suppressed=0   redundant=321 converged=1/1
treatment: sends=256 summary_skips=21  suppressed=500 redundant=136 converged=1/1
```

**44% fewer broadcast legs, 58% fewer redundant deliveries, convergence unchanged.** `sends + summary_skips + suppressed` is **777 in both arms** — the two runs did identical fan-out work and the treatment merely reclassified 500 legs, which rules out a topology difference as the explanation. The test asserts that identity.

Two honest caveats on that number:

- **The mechanisms overlap.** Of the 500 suppressions, roughly 300 were legs the pre-existing summary-match gate would have skipped for free, which is why the send reduction (44%) is smaller than the suppression count suggests. On the fleet the overlap is much smaller — its dedup ratio is 18.6 against this simulation's 3.7 — so the marginal value there should be larger, not smaller.
- **No suppression fraction is asserted.** The measured fleet figures (0.647 at degree 5 rising to 0.924 at 17) are degree-dependent, and this simulation's achieved degree is far below the fleet's. A fleet-derived threshold here would be measuring the wrong graph.

Unit and pin coverage, against the requirements:

| Requirement | Test |
|---|---|
| Excludes exactly the listed peers, no others | `a_relayer_excludes_exactly_the_listed_peers_and_no_others` — asserts unlisted co-hosts are still PRESENT, which is the half that matters |
| Does not forward the list onward (one hop) | `the_list_a_relayer_emits_describes_only_its_own_sends` |
| List on a non-payload-bearing message ignored | `only_payload_bearing_variants_carry_a_target_list` (structural — no other variant has the field) |
| Over-cap truncates, never omits or over-suppresses | `over_cap_truncation_degrades_to_partial_suppression`, `covered_peers_over_cap_truncates_rather_than_omitting` |
| Pre-floor peer gets byte-identical traffic | `legacy_broadcast_to_encoding_is_unchanged_by_the_v2_variants` (bytes, pinned against a hand-built literal — comparing to a re-serialization would compare the encoder to itself) + `a_pre_floor_peer_receives_byte_identical_broadcast_traffic` (routing, and asserts the gate FLIPS at floor so it cannot pass against an always-shut gate) |
| End-to-end duplicate reduction with a control arm | `test_5147_originator_target_list_cuts_duplicate_deliveries` |

Plus: variant-index freeze with a fixed-length table (adding a variant without extending it is a compile error), gate-discriminates-by-version, floor-tracks-shipping-release, floor-stays-above-0.2.119, replay-under-another-transaction, and eight store unit tests covering the intersection property directly.

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and `cargo test -p freenet` (4662) all clean. (`clippy --all-targets` has 16 errors on `main` unrelated to this change.)

## Rollout

`BroadcastToV2` / `BroadcastToStreamingV2` are appended last and gated per-recipient on `BROADCAST_TARGET_LIST_MIN_VERSION = (0,2,120)`, failing closed. **Until #5167 deploys, a node never learns its gateway's version, so the gate stays shut on gateway links regardless of what the gateway runs** — a suppression shortfall there, not a correctness problem, and it resolves itself once #5167 ships. Not fixed here.

Sim default is OFF, unlike hash-first: `node.rs` documents that encoding changes default ON in sims while behavioural gates default OFF, and this one removes peers from a fan-out.

## Not in scope

The size-thresholded probe is a separate PR, and its threshold should be set from the residual measured after this lands.

Closes #5147

[AI-assisted - Claude]

---

# UPDATE — read this before reviewing

Two design-level findings landed after this PR was opened. The measured numbers below are real, but **the design is not settled**, and one blocking item is unimplemented.

## 1. BLOCKING and unimplemented: merge semantics

Found independently by two blind review lenses. **"A delivered this payload to C" is not "C has the state B is about to send."** B sends `diff(B_post_merge_state, C_cached_summary)`, and B's post-merge state is `B_prior ⊔ payload`. Whenever `B_prior` is not a subset of A's state, B's fan-out to C carries content A never sent.

Sharpest form: **suppression happens before the queue, so it bypasses the `fanout_send_needed` summary comparison that would otherwise have caught the divergence.** That is a weakening of an existing safety net. Nothing compensates faster than the 5-minute heartbeat — `proactive_summary_targets` excludes both the sender and every advertised co-host, exactly the suppressed population.

Narrowing that is real but does not close it: content ORIGINATING at B propagates via B's own client-update fan-out, which registers empty coverage and is never suppressed. So the exposure is relayed content that reached B by a path that did not reach C.

**Agreed direction (not yet implemented): suppress only when our post-merge summary equals the sender's**, compared at receive time where the driver already holds the merged value. That makes the invariant exact rather than bounded.

## 2. The seeding interaction, and what this PR's numbers actually mean

The suppressed legs also carry `sender_summary_bytes`, which seeds the receiver's cached summary and feeds `fanout_send_needed`. Suppress the leg and you suppress its passenger. Full write-up, with the regression table and the rig check that ruled out a topology artefact: <https://github.com/freenet/freenet-core/issues/5147#issuecomment-5187248908>.

The simulation caught this by failing first — the feature made traffic **worse** (sends 528 vs 460, redundant 400 vs 321) because a fully-suppressed fan-out has an empty target set, which the code read as the no-subscriber failure case and retried, re-broadcasting everything one backoff later after the coverage claim was consumed. **The mechanism defeated itself and would have shipped looking plausible: every unit test passed.**

After that fix:

```
                sends   summary_skips   suppressed   redundant   converged
control           460             317            0         321         1/1
treatment         256              21          500         136         1/1
```

44% fewer legs, 58% fewer redundant deliveries, convergence unchanged. `sends + summary_skips + suppressed` is **777 in both arms**, which is what rules out a topology difference — asserted by the test.

## 3. Caveats a reviewer must not have to derive

- **This simulation cannot detect the bug class in item 1.** One contract, one writer, so `converged` reduces to `1 >= 1`. A reviewer proved it by mutation: making `covers()` suppress *every* co-host gave byte-identical results and still reported convergence. Future sim evidence needs multiple writers and multiple contracts.
- **The end-to-end numbers come from the `simulation_tests` inline fan-out, not the production queue** — a separate copy of the same logic. `production_send_site_gates_both_v2_variants_on_the_version_check` and `the_queue_carries_the_target_list_and_refreshes_it_on_replacement` (both mutation-verified) are what cover the production path; before they existed, forcing both production `target_list_for` matches to the legacy arm left every test green.
- **Of the 500 suppressions, ~300 were legs the summary gate would have skipped for free.** The fleet overlap is smaller (dedup ratio 18.6 vs this sim's 3.7), so marginal value there should be larger — but the seeding starvation is masked, not absent.
- **The revived sender exclusion currently ships UNGATED** — no version floor, no sim override — while having the same unsafe shape as the target list. Team-lead ruling: gate it like the target list or drop it from this PR. Not yet done.
- **`FullNoTheirSummaryTracked` is the fleet metric to watch.** A peer with no cached summary gets full state (357 KB mean, 26.9% of broadcast bytes on 0.2.109). The sim's 16 full-state sends, identical in both arms, say nothing either way.

## Reproducing

```
cargo test -p freenet --features "simulation_tests,testing" \
  --test simulation_integration -- --test-threads=1 --nocapture \
  test_5147_originator_target_list_cuts_duplicate_deliveries
```

Arms differ by one `enable_broadcast_target_list()` call, need different network names, same seed. Revert the `fully_covered` early return in `p2p_protoc/broadcast.rs` to reproduce the original regression.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3958,6 +3958,7 @@ std::thread_local! {
     /// ours (the pre-existing mechanism, counted for #5147 diagnosis).
     static GLOBAL_FANOUT_SUMMARY_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     /// Fan-out targets dropped because the originator named them (#5147).
+    static GLOBAL_BROADCAST_SENDER_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_BROADCAST_TARGETS_SUPPRESSED: std::cell::Cell<u64> =
         const { std::cell::Cell::new(0) };
     /// Inbound broadcast payloads that reached a terminal classification
@@ -4062,6 +4063,7 @@ impl GlobalTestMetrics {
         GLOBAL_DELTA_SENDS.with(|c| c.set(0));
         GLOBAL_FANOUT_SUMMARY_SKIPS.with(|c| c.set(0));
         GLOBAL_BROADCAST_TARGETS_SUPPRESSED.with(|c| c.set(0));
+        GLOBAL_BROADCAST_SENDER_SKIPS.with(|c| c.set(0));
         GLOBAL_BROADCAST_DELIVERIES.with(|c| c.set(0));
         GLOBAL_REDUNDANT_BROADCAST_DELIVERIES.with(|c| c.set(0));
         GLOBAL_FULL_STATE_SENDS.with(|c| c.set(0));
@@ -4196,6 +4198,25 @@ impl GlobalTestMetrics {
     /// exactly 0 when the feature is off, which a derived count would not do.
     pub fn record_broadcast_target_suppressed() {
         GLOBAL_BROADCAST_TARGETS_SUPPRESSED.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Records that the peer which delivered this update was dropped from our
+    /// own fan-out (#5147 sender exclusion).
+    ///
+    /// A FOURTH terminal outcome for an offered leg, alongside sent /
+    /// summary-skipped / list-suppressed. It needs its own counter for the same
+    /// reason the others do — it is incremented by the filter that makes the
+    /// decision, not derived at a call site — and because without it the
+    /// simulation's leg-accounting identity is short by exactly the number of
+    /// sender exclusions, which reads as the two arms having done different
+    /// amounts of work when they did not.
+    pub fn record_broadcast_sender_skipped() {
+        GLOBAL_BROADCAST_SENDER_SKIPS.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Fan-out legs dropped because the target was the delivering peer.
+    pub fn broadcast_sender_skips() -> u64 {
+        GLOBAL_BROADCAST_SENDER_SKIPS.with(|c| c.get())
     }
 
     /// Fan-out targets suppressed by the originator target list since reset.
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3958,9 +3958,10 @@ std::thread_local! {
     /// ours (the pre-existing mechanism, counted for #5147 diagnosis).
     static GLOBAL_FANOUT_SUMMARY_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     /// Fan-out targets dropped because the originator named them (#5147).
-    static GLOBAL_BROADCAST_SENDER_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_BROADCAST_TARGETS_SUPPRESSED: std::cell::Cell<u64> =
         const { std::cell::Cell::new(0) };
+    /// Fan-out legs dropped because the target was the delivering peer (#5147).
+    static GLOBAL_BROADCAST_SENDER_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     /// Inbound broadcast payloads that reached a terminal classification
     /// (#5147). Denominator for the duplicate-delivery ratio.
     static GLOBAL_BROADCAST_DELIVERIES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3954,6 +3954,19 @@ impl SimulationIdleTimeout {
 std::thread_local! {
     static GLOBAL_RESYNC_REQUESTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_DELTA_SENDS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    /// Fan-out legs skipped because the peer's cached summary already matched
+    /// ours (the pre-existing mechanism, counted for #5147 diagnosis).
+    static GLOBAL_FANOUT_SUMMARY_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    /// Fan-out targets dropped because the originator named them (#5147).
+    static GLOBAL_BROADCAST_TARGETS_SUPPRESSED: std::cell::Cell<u64> =
+        const { std::cell::Cell::new(0) };
+    /// Inbound broadcast payloads that reached a terminal classification
+    /// (#5147). Denominator for the duplicate-delivery ratio.
+    static GLOBAL_BROADCAST_DELIVERIES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    /// The subset of those that carried nothing new — a deduped duplicate, or
+    /// a merge that moved no state (#5147).
+    static GLOBAL_REDUNDANT_BROADCAST_DELIVERIES: std::cell::Cell<u64> =
+        const { std::cell::Cell::new(0) };
     static GLOBAL_FULL_STATE_SENDS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PENDING_OP_INSERTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PENDING_OP_REMOVES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
@@ -4047,6 +4060,10 @@ impl GlobalTestMetrics {
     pub fn reset() {
         GLOBAL_RESYNC_REQUESTS.with(|c| c.set(0));
         GLOBAL_DELTA_SENDS.with(|c| c.set(0));
+        GLOBAL_FANOUT_SUMMARY_SKIPS.with(|c| c.set(0));
+        GLOBAL_BROADCAST_TARGETS_SUPPRESSED.with(|c| c.set(0));
+        GLOBAL_BROADCAST_DELIVERIES.with(|c| c.set(0));
+        GLOBAL_REDUNDANT_BROADCAST_DELIVERIES.with(|c| c.set(0));
         GLOBAL_FULL_STATE_SENDS.with(|c| c.set(0));
         GLOBAL_PENDING_OP_INSERTS.with(|c| c.set(0));
         GLOBAL_PENDING_OP_REMOVES.with(|c| c.set(0));
@@ -4155,6 +4172,68 @@ impl GlobalTestMetrics {
     /// last reset.
     pub fn resync_responses_suppressed() -> u64 {
         Self::resync_responses_suppressed_per_peer() + Self::resync_responses_suppressed_global()
+    }
+
+    /// Records a fan-out leg skipped by the summary-match gate.
+    pub fn record_fanout_summary_skip() {
+        GLOBAL_FANOUT_SUMMARY_SKIPS.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Fan-out legs skipped by the summary-match gate since reset.
+    pub fn fanout_summary_skips() -> u64 {
+        GLOBAL_FANOUT_SUMMARY_SKIPS.with(|c| c.get())
+    }
+
+    /// Records one fan-out target dropped because the originator's list named
+    /// it (#5147).
+    ///
+    /// Incremented BY the filter in `get_broadcast_targets_update`, alongside
+    /// its own `skipped_covered` field, never re-derived from the difference
+    /// between the resolved co-host set and the final target set — that
+    /// subtraction also absorbs the sender, self, and resolve-failure filters,
+    /// so it would keep reporting a plausible number after this filter is
+    /// deleted. The simulation's discriminator rests on this counter going to
+    /// exactly 0 when the feature is off, which a derived count would not do.
+    pub fn record_broadcast_target_suppressed() {
+        GLOBAL_BROADCAST_TARGETS_SUPPRESSED.with(|c| c.set(c.get() + 1));
+    }
+
+    /// Fan-out targets suppressed by the originator target list since reset.
+    pub fn broadcast_targets_suppressed() -> u64 {
+        GLOBAL_BROADCAST_TARGETS_SUPPRESSED.with(|c| c.get())
+    }
+
+    /// Records one inbound broadcast payload reaching a terminal outcome, and
+    /// whether it carried anything new (#5147).
+    ///
+    /// Called from `PayloadMix::record_receiver_terminal` — the ONE place that
+    /// classifies an inbound broadcast — so the redundancy count is produced by
+    /// the code making the decision rather than re-derived from set sizes
+    /// elsewhere. See `.claude/rules/bug-prevention-patterns.md`, "Metric
+    /// describing a filtering decision, re-derived at the call site": a
+    /// subtraction across two collections silently absorbs every other filter
+    /// between them and keeps reporting a plausible number after the filter it
+    /// claims to measure is gone.
+    ///
+    /// A `Failed` outcome is counted as a delivery but NOT as redundant: the
+    /// payload may well have carried something new and the merge simply broke.
+    /// Calling a failure redundant would flatter any change that increased
+    /// merge failures.
+    pub fn record_broadcast_delivery(redundant: bool) {
+        GLOBAL_BROADCAST_DELIVERIES.with(|c| c.set(c.get() + 1));
+        if redundant {
+            GLOBAL_REDUNDANT_BROADCAST_DELIVERIES.with(|c| c.set(c.get() + 1));
+        }
+    }
+
+    /// Inbound broadcast payloads that reached a terminal outcome since reset.
+    pub fn broadcast_deliveries() -> u64 {
+        GLOBAL_BROADCAST_DELIVERIES.with(|c| c.get())
+    }
+
+    /// Those of [`Self::broadcast_deliveries`] that changed nothing.
+    pub fn redundant_broadcast_deliveries() -> u64 {
+        GLOBAL_REDUNDANT_BROADCAST_DELIVERIES.with(|c| c.get())
     }
 
     /// Records that a delta was sent in a state change broadcast.
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -65,12 +65,16 @@ pub(crate) use network_bridge::broadcast_queue_metrics::BROADCAST_QUEUE_EFFICIEN
 // re-export rather than a path through `network_bridge` directly. Mirrors
 // `BROADCAST_STREAM_METRICS` above.
 pub(crate) use network_bridge::p2p_protoc::{
-    HASH_FIRST_SUMMARIES_MIN_VERSION, SUMMARY_FIRST_PUT_MIN_VERSION,
+    BROADCAST_TARGET_LIST_MIN_VERSION, HASH_FIRST_SUMMARIES_MIN_VERSION,
+    SUMMARY_FIRST_PUT_MIN_VERSION, version_supports_broadcast_target_list,
     version_supports_hash_first_summaries, version_supports_summary_first_put,
 };
 // Test-only: the release-timing marker has no runtime reader by design (see
 // its rustdoc), so re-exporting it unconditionally is an unused import in the
 // non-test build.
+#[cfg(test)]
+#[allow(unused_imports)]
+pub(crate) use network_bridge::p2p_protoc::BROADCAST_TARGET_LIST_SHIPPED_IN;
 #[cfg(test)]
 pub(crate) use network_bridge::p2p_protoc::HASH_FIRST_SHIPPED_IN;
 #[cfg(test)]
@@ -431,6 +435,30 @@ pub struct NodeConfig {
     /// `#[serde(skip)]`; never serialized.
     #[serde(skip)]
     pub(crate) ack_version_floor_override: Option<(u8, u8, u16)>,
+    /// Test-only override for the originator-target-list version floor
+    /// ([`BROADCAST_TARGET_LIST_MIN_VERSION`], #5147). Threaded exactly like
+    /// the three overrides above and consulted as
+    /// `broadcast_target_list_floor_override.unwrap_or(BROADCAST_TARGET_LIST_MIN_VERSION)`
+    /// by `ConnectionManager::supports_broadcast_target_list`.
+    ///
+    /// In production this is `None` → the real `(0, 2, 120)` floor.
+    ///
+    /// **Simulations default this to `SIM_MIGRATION_DISABLED_FLOOR` — OFF —
+    /// like `subscribe_hint_floor_override` and `summary_first_put_floor_override`,
+    /// and unlike `hash_first_summaries_floor_override`.** The rule stated at
+    /// that field applies: hash-first defaults ON because it changes only the
+    /// ENCODING of an exchange with identical semantics, whereas this gate
+    /// changes BEHAVIOUR — it removes peers from a fan-out. Every sim that
+    /// asserts on delivery or convergence would silently be exercising a
+    /// different delivery graph, and a failure could not be attributed between
+    /// the sim's own subject and this suppression. Tests that want it opt in
+    /// via `SimNetwork::enable_broadcast_target_list`.
+    ///
+    /// Not cfg-gated, for the same reason as the others: `node::testing_impl`
+    /// sets it and is compiled unconditionally. `#[serde(skip)]`; never
+    /// serialized.
+    #[serde(skip)]
+    pub(crate) broadcast_target_list_floor_override: Option<(u8, u8, u16)>,
     /// Test-only harness flag: when set, a startup-hosted contract
     /// (`SeedHostedContract`, i.e. `append_contracts` with `subscription =
     /// true`) is registered in the neighbor-hosting advertised set so the
@@ -599,6 +627,7 @@ impl NodeConfig {
             summary_first_put_floor_override: None,
             hash_first_summaries_floor_override: None,
             ack_version_floor_override: None,
+            broadcast_target_list_floor_override: None,
             advertise_seeded_hosts: false,
             hosting_time_source_override: None,
         })
@@ -1535,7 +1564,9 @@ where
                     update::UpdateMsg::RequestUpdate { key, .. }
                     | update::UpdateMsg::BroadcastTo { key, .. }
                     | update::UpdateMsg::RequestUpdateStreaming { key, .. }
-                    | update::UpdateMsg::BroadcastToStreaming { key, .. } => *key,
+                    | update::UpdateMsg::BroadcastToStreaming { key, .. }
+                    | update::UpdateMsg::BroadcastToV2 { key, .. }
+                    | update::UpdateMsg::BroadcastToStreamingV2 { key, .. } => *key,
                 };
 
                 // Phase 7 ban-list gate. Runs BEFORE the rate limiter
@@ -1596,6 +1627,10 @@ where
                         }
                         return Ok(());
                     }
+                    // The legacy variant carries no target list, so the
+                    // relayer suppresses nothing and behaves exactly as it does
+                    // today. `CoveredPeers::empty()` is that "nobody is
+                    // covered" statement, not a missing value.
                     update::UpdateMsg::BroadcastTo {
                         id,
                         key,
@@ -1609,6 +1644,7 @@ where
                             payload.clone(),
                             sender_summary_bytes.clone(),
                             sender_addr,
+                            crate::ring::broadcast_coverage::CoveredPeers::empty(),
                         )
                         .await
                         {
@@ -1661,6 +1697,7 @@ where
                             *stream_id,
                             *total_size,
                             sender_addr,
+                            crate::ring::broadcast_coverage::CoveredPeers::empty(),
                         )
                         .await
                         {
@@ -1669,6 +1706,69 @@ where
                                 %key,
                                 error = %err,
                                 "UPDATE relay dispatch: start_relay_broadcast_to_streaming failed"
+                            );
+                        }
+                        return Ok(());
+                    }
+                    // #5147. These two arms are the ONLY places a target list
+                    // is honored, and that is the security rule rather than an
+                    // implementation detail: the list is trusted solely because
+                    // it arrived on the message that delivered the payload, so
+                    // its author is attesting to sends it actually made. A
+                    // relayer therefore cannot fabricate an "everyone is
+                    // covered" list to suppress third-party fan-out, because
+                    // there is no other message type that carries one. Do NOT
+                    // add a `covered` field to a non-payload-bearing variant.
+                    update::UpdateMsg::BroadcastToV2 {
+                        id,
+                        key,
+                        payload,
+                        sender_summary_bytes,
+                        covered,
+                    } => {
+                        if let Err(err) = update::op_ctx_task::start_relay_broadcast_to(
+                            op_manager.clone(),
+                            *id,
+                            *key,
+                            payload.clone(),
+                            sender_summary_bytes.clone(),
+                            sender_addr,
+                            covered.clone(),
+                        )
+                        .await
+                        {
+                            tracing::error!(
+                                tx = %id,
+                                %key,
+                                error = %err,
+                                "UPDATE relay dispatch: start_relay_broadcast_to (v2) failed"
+                            );
+                        }
+                        return Ok(());
+                    }
+                    update::UpdateMsg::BroadcastToStreamingV2 {
+                        id,
+                        key,
+                        stream_id,
+                        total_size,
+                        covered,
+                    } => {
+                        if let Err(err) = update::op_ctx_task::start_relay_broadcast_to_streaming(
+                            op_manager.clone(),
+                            *id,
+                            *key,
+                            *stream_id,
+                            *total_size,
+                            sender_addr,
+                            covered.clone(),
+                        )
+                        .await
+                        {
+                            tracing::error!(
+                                tx = %id,
+                                %key,
+                                error = %err,
+                                "UPDATE relay dispatch: start_relay_broadcast_to_streaming (v2) failed"
                             );
                         }
                         return Ok(());
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -926,6 +926,12 @@ impl PayloadMix {
         outcome: ReceiverTerminalOutcome,
         payload_bytes: usize,
     ) {
+        // #5147 measurement, emitted from the classification itself so the
+        // redundancy count cannot drift from the decision it describes.
+        crate::config::GlobalTestMetrics::record_broadcast_delivery(matches!(
+            outcome,
+            ReceiverTerminalOutcome::Dedup | ReceiverTerminalOutcome::NoOp
+        ));
         let outcome_index = outcome.index();
         let kind_index = usize::from(!is_delta) * 5 + outcome_index;
         let bucket = state_size_bucket(payload_bytes as u64);
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -611,7 +611,16 @@ pub(super) async fn fanout_send_needed(
 ) -> bool {
     match plan_fanout_send(&op_manager.interest_manager, key, summaries, *probes_used) {
         FanoutSendPlan::Send => true,
-        FanoutSendPlan::Skip => false,
+        FanoutSendPlan::Skip => {
+            // #5147 diagnostic: the pre-existing summary-match skip is the
+            // OTHER mechanism suppressing fan-out legs, and it is fed by the
+            // `sender_summary_bytes` that rides on the very sends the target
+            // list removes. Counting it is what makes an interaction between
+            // the two visible instead of showing up as an unexplained rise in
+            // total sends.
+            crate::config::GlobalTestMetrics::record_fanout_summary_skip();
+            false
+        }
         FanoutSendPlan::Probe => {
             *probes_used += 1;
             let SummaryPair { ours, theirs } = summaries;
--- a/crates/core/src/node/network_bridge/broadcast_queue.rs
+++ b/crates/core/src/node/network_bridge/broadcast_queue.rs
@@ -1831,6 +1831,168 @@ mod queue {
             }
         }
 
+        /// #5147: the PRODUCTION send site must decide the wire variant through
+        /// the version gate, and must build the legacy arm unchanged.
+        ///
+        /// `broadcast_to_single_peer` is the production message-construction
+        /// site and is exercised by no behavioural test — it needs a live
+        /// `P2pBridge` and `OpManager`, and the end-to-end simulation runs a
+        /// SEPARATE inline copy in `p2p_protoc/broadcast.rs`. A reviewer forced
+        /// both `match target_list_for(..)` expressions to their `None` arm —
+        /// making the feature entirely inert in production — and every lib test
+        /// still passed. This pin is what closes that hole.
+        #[test]
+        fn production_send_site_gates_both_v2_variants_on_the_version_check() {
+            const SOURCE: &str = include_str!("broadcast_queue.rs");
+
+            // Assembled with `concat!` rather than written whole, and NOT
+            // spelled out in any comment here either.
+            //
+            // This test module sits BEFORE the function in the file, and the
+            // pre-existing pins further down slice the function's body by
+            // searching for its signature from the top of the file. Any earlier
+            // occurrence — in a literal OR in a comment — makes those pins slice
+            // from here instead and assert against this test's text. Writing it
+            // out once broke two of them; writing it out in the comment
+            // explaining that fix broke them again.
+            let fn_anchor = concat!("pub(super) async fn ", "broadcast_to_single_", "peer(");
+            let start = SOURCE
+                .find(fn_anchor)
+                .expect("broadcast_to_single_peer renamed or removed");
+            let after = &SOURCE[start + fn_anchor.len()..];
+            let end = after
+                .find("\n#[cfg(test)]")
+                .or_else(|| after.find("\nmod "))
+                .map(|p| start + fn_anchor.len() + p)
+                .unwrap_or(SOURCE.len());
+            let body = &SOURCE[start..end];
+
+            for (variant, legacy) in [
+                ("UpdateMsg::BroadcastToV2 {", "UpdateMsg::BroadcastTo {"),
+                (
+                    "UpdateMsg::BroadcastToStreamingV2 {",
+                    "UpdateMsg::BroadcastToStreaming {",
+                ),
+            ] {
+                assert!(
+                    body.contains(variant),
+                    "the production send site must be able to emit `{variant}`. \
+                     Without it the whole feature is inert in production while \
+                     the simulation — which runs a SEPARATE inline copy of this \
+                     logic — still reports it working."
+                );
+                assert!(
+                    body.contains(legacy),
+                    "the production send site must keep emitting `{legacy}` for \
+                     pre-floor peers, byte-identically to today"
+                );
+            }
+
+            let gate = body
+                .find("target_list_for(")
+                .expect("the production send site must consult the version gate");
+            let first_v2 = body
+                .find("UpdateMsg::BroadcastToStreamingV2 {")
+                .expect("checked above");
+            assert!(
+                gate < first_v2,
+                "the version gate (offset {gate}) must be consulted BEFORE any \
+                 V2 variant is constructed (offset {first_v2}); a pre-floor \
+                 peer that receives one cannot decode it and the connection is \
+                 CLOSED"
+            );
+        }
+
+        /// #5147: the target list must survive the queue, and a dedup
+        /// replacement must REFRESH it.
+        ///
+        /// The full fan-out set exists only at `handle_broadcast_state_change`;
+        /// it is immediately split into one queue entry per target, and the
+        /// send site sees only its own peer. So the list rides on
+        /// `BroadcastEntry`, and the replace-on-dedup path has to update it in
+        /// lockstep with `new_state` — a superseded entry's list names peers a
+        /// DIFFERENT fan-out targeted, which is over-suppression, the one
+        /// direction this design refuses.
+        ///
+        /// This is also the only test that drives the PRODUCTION queue path:
+        /// the end-to-end simulation runs the `simulation_tests` inline branch,
+        /// a separate copy, so nothing else would notice if the queue dropped
+        /// the list entirely. Mutation-verified — replacing the refresh with a
+        /// no-op fails this test.
+        #[tokio::test(start_paused = true)]
+        #[serial_test::serial(broadcast_queue_depth_gauge)]
+        async fn the_queue_carries_the_target_list_and_refreshes_it_on_replacement() {
+            use crate::transport::TransportKeypair;
+
+            let queue = BroadcastQueue::new();
+            let group = LaneGroup {
+                small_pool: Arc::new(Semaphore::new(4)),
+                large_pool: Arc::new(Semaphore::new(4)),
+                upgrade_slots: Arc::new(Semaphore::new(super::super::MAX_PARKED_LANE_UPGRADES)),
+                small_notify: queue.small_notify.clone(),
+                large_notify: queue.large_notify.clone(),
+            };
+            let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+            let worker = tokio::spawn(drain_lane(
+                QueuedPayloadClass::Small,
+                queue.queue.clone(),
+                group,
+                move |entry: BroadcastEntry, scheduling: QueueScheduling| {
+                    let tx = tx.clone();
+                    async move {
+                        let _scheduling = scheduling;
+                        let _ = tx.send(entry.fanout);
+                    }
+                },
+            ));
+
+            let key = contract(77);
+            let target = PeerKeyLocation::random();
+            let superseded: FanoutTargets =
+                Arc::new(vec![TransportKeypair::new().public().clone()]);
+            let current: FanoutTargets = Arc::new(vec![
+                TransportKeypair::new().public().clone(),
+                TransportKeypair::new().public().clone(),
+            ]);
+
+            // Two enqueues for the same (contract, peer): the second dedups
+            // onto the first and must win on BOTH state and target list.
+            queue
+                .enqueue_in_lane(
+                    QueuedPayloadClass::Small,
+                    key,
+                    target.clone(),
+                    state(16),
+                    superseded.clone(),
+                )
+                .await;
+            queue
+                .enqueue_in_lane(
+                    QueuedPayloadClass::Small,
+                    key,
+                    target,
+                    state(32),
+                    current.clone(),
+                )
+                .await;
+
+            let dispatched = tokio::time::timeout(Duration::from_secs(5), rx.recv())
+                .await
+                .expect("the drain must dispatch")
+                .expect("the drain must send the entry's fanout");
+
+            assert_eq!(
+                *dispatched, *current,
+                "the replacement's target list must reach the send site; a \
+                 superseded list names peers a different fan-out targeted, so \
+                 the recipient would suppress peers this broadcast never \
+                 delivered to"
+            );
+            assert_ne!(*dispatched, *superseded);
+
+            worker.abort();
+        }
+
         /// A lane drain can only exit on a closed pool, which nothing does in
         /// production. But there are TWO of them now, and the pre-#4961 failure
         /// — one worker exiting, so all broadcast draining stops — was at least
@@ -3205,19 +3367,6 @@ fn record_delivery_to_interest<T: crate::util::time_source::TimeSource + Sync>(
     }
 }
 
-/// Send a state change broadcast to a single peer.
-///
-/// This is the per-target body extracted from `broadcast_state_to_peers`.
-/// It handles delta computation, streaming vs inline decision, and telemetry.
-///
-/// For streaming sends, a completion oneshot is created internally and threaded
-/// through the stream send path. The function awaits it (with timeout) so the
-/// semaphore permit — owned by `scheduling`, dropped when this function returns
-/// — is held until the actual stream transfer finishes.
-///
-/// `scheduling` is `Some` exactly when the production queue dispatched this
-/// send, and carries both the lane it was classified into and that lane's
-/// permit. The sim fan-out calls in-line with `None` (no queue, no permits).
 /// Build the originator target list to attach to one leg of a fan-out (#5147),
 /// or `None` when the recipient is on a pre-floor release.
 ///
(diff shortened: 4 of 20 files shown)


sanity:
## Full-tier review — 5 independent blind lenses + synthesis

Risk tier: **Full** (wire-format/protocol change on the contract-broadcast path, which carries prior P0 and P1 history). Five reviewers ran blind and in parallel — *code-first*, *skeptical*, *wire-format/version/security*, *big-picture*, *testing* — each reading the checked-out code rather than the diff text, none seeing the others.

Reviewed at `086f3e71d`, rebased onto `8600d5c35`. **The branch was behind main and conflicting; I rebased it.** The conflicts were all additive (this branch's floor override vs the merged #5167 ack-version override), but two things needed real fixes and are in the rebase commits: the branch had inserted its two new sim helpers *inside* `disable_hash_first_summaries`'s doc comment (hijacking it and leaving that function undocumented), and the sim-test file needed reconstructing as main's file plus the #5147 block.

**Recommendation: do not merge yet, and let 0.2.120 wait for it rather than compressing this review.** The five merged PRs do not depend on this one, so a slip costs only time. The design is sound and unusually well-argued — what follows is mostly about evidence not covering what it claims to cover.

Findings marked ✅ I verified myself against the source rather than relaying.

---

## BLOCKING

### B1. The headline simulation does not exercise the production send path ✅
`p2p_protoc/broadcast.rs:741`, `broadcast_queue.rs:1197`, `:3711`, `:3856`
*(raised independently by the testing, code-first and wire lenses)*

`broadcast_state_to_peers` — the function the simulation drives — is `#[cfg(feature = "simulation_tests")]` and is **never compiled into a production binary**. The code says so itself at `broadcast.rs:776-779`: *"production splits the fan-out across a queue and rebuilds the list per leg, this one holds the whole set in scope. A test that only exercises this branch proves nothing about production."*

`broadcast.rs:595` is the only call site in the repo passing a non-empty `FanoutTargets`; all 19 others, **including every test**, pass `no_fanout()`. So the entire evidentiary basis for 44% fewer legs / 58% fewer redundant deliveries comes from a path that is not the one shipping.

Mutations that should be red and would all be green:
- replace `fanout.clone()` with `no_fanout()` at `broadcast.rs:600` — the feature becomes **entirely inert in production**, and the sim still reports `suppressed=500` and passes;
- delete `existing.fanout = fanout;` at `broadcast_queue.rs:1197` — the line whose own comment calls it the guard against "over-suppression, the one direction this design refuses". Every queue test replaces an empty fanout with an empty fanout, so the two are indistinguishable.

*Wanted:* a queue test enqueuing two entries with **distinct non-empty** fanouts, asserting the replaced entry carries the new one; plus a source-scrape pin that the production enqueue loop passes the built `fanout` rather than `no_fanout()`, and that the `None` arm of both send sites names the legacy spellings.

### B2. The fully-covered branch increments the `no_targets` failure counter its own comment says it avoids ✅
`p2p_protoc/broadcast.rs:320-328` + `operations/update/propagation_stats.rs:253-258`
*(skeptical + big-picture, independently)*

The comment: *"…but NOT through the `no_targets` counter below — a fully-covered fan-out is a success and must not read as a propagation failure to an operator."* Then it calls `record_broadcast(*key.id(), 0, …)`, and `record_broadcast` is unconditional:

```rust
entry.updates += 1;
entry.targets_total += targets as u64;
if targets == 0 { entry.no_targets += 1; }
```

So the operator-facing `update_propagation_summary` reports rising `no_targets` **in direct proportion to how well #5147 works** — the exact false alarm the comment names, on the telemetry this rollout will be judged by. The pin at `propagation_stats.rs:744` asserts only the call *count* (3), so it passes while restating the false claim.

### B3. `fully_covered` fires on `skipped_sender > 0` alone, cancelling the retry and discarding the #4359 stash ✅
`p2p_protoc/broadcast.rs:304`
*(all four of skeptical, code-first, wire, big-picture)*

```rust
let fully_covered = target_result.targets.is_empty()
    && (target_result.skipped_covered > 0 || target_result.skipped_sender > 0);
```

`skipped_sender > 0` only means "the peer that delivered this is one of my advertised co-hosts" — true on essentially every relayed broadcast. It says nothing about peers **not yet** in the co-host set, which is precisely the population the retry/stash branch exists for (its own comment: *"a race between an update and subscription establishment"*).

Scenario: B hosts K, its only advertised co-host is A. C has just subscribed to B but its `HostingAnnounce` has not landed. A broadcasts to B → `targets=[]`, `skipped_sender=1` → `fully_covered` → `broadcast_retries.remove`, `broadcast_no_target_streak.remove`, and `pending_broadcasts.take(key.id())` **discards the stash** (`broadcast.rs:319`) — having sent to nobody. Previously B retried 3× over ~6 s then stashed for the interest flush, so C got the state when its advertisement arrived. Now C waits for the ~5-minute heartbeat.

The stash drop is the sharp part: the targets-found path justifies the same `take` as *"this fan-out is reaching targets now, so a previously-abandoned state is superseded"* (`:534`) — which does not transfer to a branch that reaches nobody. It arguably should **re-stash** `new_state`. And per B6 this is reachable with the version gate fully shut, so it is fleet-wide from day one.

### B4. The assertion carrying the byte-identity test's name is a tautology ✅
`node/op_state_manager.rs:3329-3339`
*(testing + wire, independently)*

```rust
let legacy = UpdateMsg::BroadcastTo { id: tx, key, payload: payload.clone(), sender_summary_bytes: summary.clone() };
assert_eq!(
    bincode::serialize(&legacy)…,
    bincode::serialize(&UpdateMsg::BroadcastTo { id: tx, key, payload: payload.clone(), sender_summary_bytes: summary.clone() })…,
    "a pre-floor peer's bytes must be exactly today's BroadcastTo"
);
```

Both sides are the same expression over the same four variables. **No input makes this red**; it proves bincode is deterministic. The test's genuinely load-bearing lines are `target_list_for(…).is_none()` at `:3320` and the gate-flip `assert_ne!` at `:3363` — those are real and worth keeping. Delete the tautology; the actual byte pin lives at `update.rs:1782`.

### B5. A bandwidth change that asserts no bytes, with the field revealing its known worst case captured, printed, and never checked ✅
`tests/simulation_integration.rs:17321-17323`, logged at `:17510-17530`

Every end-to-end assertion counts **messages** (`sends`, `redundant`, `suppressed`). Two mechanisms in this very design let total **bytes** rise while all three counts fall:

1. the list costs bytes on every surviving leg — up to 64 × 8 + 8 = **520 bytes** per `BroadcastToV2`, and nothing weighs added against saved;
2. suppressing a leg removes the `sender_summary_bytes` piggyback that keeps peer-summary caches warm, and a peer with no cached summary of us gets **FULL STATE** (`FullNoTheirSummaryTracked`: 26.9% of broadcast bytes at a 357 KB mean in the 0.2.109 profile) instead of a few-KB delta.

The test's own docstring at `:17352-17360` states hazard 2 outright. A change cutting `sends` 44% while raising `full_state_sends` 10× makes the fleet strictly worse on bytes and passes every assertion here. `resync_suppressed` has the same shape.

*Minimum:* `assert!(treatment.full_state_sends <= control.full_state_sends)`. *Better:* `PayloadMix` already tracks payload bytes — capture a byte total per arm and assert it fell. The issue's claim is about bytes; the test's is about legs.

---

## MAJOR

### M1. The revived sender/self exclusion is not version-gated and not covered by the sim opt-out
`operations/update.rs:427`, `:436`; `update/op_ctx_task.rs:775`; `node.rs:1647`; `testing_impl.rs:1451`
*(skeptical, code-first, wire, and the synthesis independently)*

`resolve_covered_peers` always builds `BroadcastOrigin::relayed(sender_addr, …)` — including the legacy arm, which passes `CoveredPeers::empty()` rather than "no origin". So `sender` is always `Some`, and both filters fire on every relayed apply regardless of anyone's version. `supports_broadcast_target_list` has exactly one consumer, `target_list_for` on the **send** side.

Consequences: `disable_broadcast_target_list()` does not restore pre-#5147 fan-out, so **both sim arms carry it** and the control arm is not main's behaviour (the reported 44% is the covered-list half only — conservative, but the arm's described role is inaccurate); it ships as an unconditional behaviour change at 0.2.120 with no floor and no runtime switch; and it is what makes B3 reachable on links where the target list never activates, including every gateway link until #5167 propagates.

Worth noting the flip side, since it is a genuine positive the PR does not claim: this half works against legacy peers too, so unlike the target list it delivers savings **unilaterally** on 0.2.120 without waiting for fleet adoption.

Big-picture raises a substantive design objection here too: *"sending it back to the sender is a guaranteed duplicate"* is not true under CRDT merge — the relayer's post-merge state can be strictly ahead. Previously `plan_fanout_send` made that call semantically (Skip on matching summaries, Send when genuinely ahead); the blunt address filter now drops the sender before that gate runs, and `proactive_summary_targets` (`update.rs:1281`) also excludes the sender, so there is no second path short of the heartbeat. No test covers the ahead-sender case.

### M2. `BroadcastCoverageStore` has no sweeper; orphaned entries are never reclaimed
`ring/broadcast_coverage.rs:281`, `:404`, `:415` *(skeptical, code-first, big-picture)*

Entries leave only via `take`, `discard`, or same-key replacement. `expires_at` governs **validity**, never **residency**. Two systematic producers: the banned-contract (`broadcast.rs:211`) and broken-invariant (`:238`) early returns run **before** `take` at `:274`, so every `keep()`-ing apply on such a contract leaks an entry; and `try_notify_node_event(BroadcastStateChange)` is best-effort by design (#4145), so a dropped event leaves a kept entry with no consumer. Both fire hardest under exactly the storm conditions (#4861/#4903 shape) where the node is already under pressure. Sweep on `register`, or `entries.retain(…)` periodically.

### M3. The stale-entry replacement branch is dead in effect ✅
`ring/broadcast_coverage.rs:380-391` *(skeptical, testing)*

```rust
if existing.expires_at <= now { existing.origin = origin; }   // "Replace rather than narrow"
else { existing.origin.narrow(&origin); }
existing.expires_at = existing.expires_at.min(expires_at);    // the PAST value wins
```
In the stale branch `existing.expires_at <= now < expires_at`, so `min` keeps the expired deadline; the replacement is born expired and `take` discards it. Direction is safe (under-suppression), but the orphan backstop does the opposite of its rustdoc — and **no test anywhere exercises `COVERAGE_TTL`**, so neither this nor a future inversion would be caught. Use `expires_at` in the stale arm, and add a `#[cfg(test)]` constructor that can set the deadline.

### M4. The security rustdoc states a guarantee the design does not provide
`ring/broadcast_coverage.rs:52-72` *(skeptical, wire)*

The doc argues a malicious originator is a non-attack ("it could simply not send") and that honouring a list only from the payload-bearing message closes the relayer case. That rule closes fabrication by a peer that delivers **nothing** — but a malicious relayer *is* sending a payload-bearing message, mints the `update_tx` that seeds the hashes, and `resolve` verifies nothing about the claim. M, a genuine co-host, forwards a real update to B naming all of B's co-hosts; B suppresses its entire re-fan-out, and where B is a cut vertex propagation stops for the cycle. This is an **amplification** over withholding: withholding costs M its own links, lying costs M's links *plus* B's whole fan-out. Bounded (one hop, one cycle, heartbeat heals, attacker must be in the delivery path) and probably acceptable — but state the residual rather than presenting the hole as closed.

### M5. Release sequencing: the floor may name a release that does not carry the code
`p2p_protoc.rs:1089`, `:1078`; `docs/RELEASING.md`

`BROADCAST_TARGET_LIST_MIN_VERSION = (0,2,120)` with `SHIPPED_IN = None` while the crate is 0.2.119 and 0.2.120 is the release being cut. If 0.2.120 tags without this PR, peers on the real 0.2.120 satisfy the floor and receive a `BroadcastToV2` whose variant index they lack — and decode failure **closes the connection** (`p2p_protoc.rs:3483-3489`), so across the 0-4h staggered rollout that is fleet-wide churn presenting as a transport fault. The tripwire at `connection_manager.rs:3359` fires correctly in both directions at the next bump, and the lower-bound guard at `:3419` catches a lowering. Two residual gaps: `docs/RELEASING.md` documents the four sibling constant pairs and not this one, so the releaser's checklist never mentions it; and resolving the red test by setting `SHIPPED_IN = Some((0,2,120))` when the code did **not** land goes green, since the `Some` arm only checks `shipped == floor && floor <= current`. **Either confirm it lands in 0.2.120, or raise the floor to (0,2,121) — and add the constant pair to `docs/RELEASING.md` in this PR.**

### M6. `assert_eq!(control_legs, treatment_legs)` pins an exact equality on a quantity that is not conserved
`tests/simulation_integration.rs:17594-17601` *(code-first, skeptical, testing)*

The three-way partition is a genuinely good control against a topology confound, but at least four outcomes fall outside all three buckets: `sends` counts only **successful** sends (both increment sites sit inside the `Ok` arm), the `compute_delta` `Ok(None)` empty-delta return skips without incrementing `record_fanout_summary_skip`, the `should_broadcast_contract` early return, and queue-capacity eviction. The treatment arm also adds up to 520 bytes per broadcast, so packetisation and timing differ between arms. One extra send failure flips 777 ≠ 776 and the test fails blaming "a topology difference" that did not occur. Under this repo's flaky-tests-are-blockers rule that is a merge blocker as written; a tolerance band, or counting the missing buckets, preserves the intent.

### M7. A named peer that never receives the payload is still published as covered
`broadcast_queue.rs:3235-3245` *(skeptical, code-first, big-picture)*

The `target_list_for` rustdoc names queue eviction as the only non-benign enqueue-vs-delivery gap. Two more are at least as likely: a non-streaming `bridge.send` error only logs a WARN (`:3931`) with no retry and no re-stash, and a dropped stream deliberately does not record delivery (`:3945`, #4235) — stream-transfer failure being a known recurring class here. A fails to C; B and D were both told C is covered, so neither re-fans-out to C, and under a half-broken-but-not-yet-pruned link this repeats every broadcast, leaving C persistently a cycle behind — indistinguishable from the stale-host class invariant 1 forbids. Gate the claim on the same `res.is_ok()` the cost axis already uses, or at minimum document these as residuals. Related: `fanout_send_needed` uses the peer's **cached** summary, so one node's stale belief is now published as a covered claim affecting every relayer.

### M8. The simulation's only safety assertion is close to a one-bit check
`tests/simulation_integration.rs:17631-17640` *(testing, big-picture)*

`treatment.converged.0 >= control.converged.0` compares a count of **contracts**, and there is one contract — so it is `1 >= 1`. There is no premise that the control arm converged at all (`0 >= 0` passes if a future change breaks both), `replica_count` is never captured, and `check_convergence_from_logs` skips contracts with `peer_states.len() < 2` and counts only peers that logged a state hash — so a peer that stops receiving, and therefore stops logging, is removed from the denominator rather than counted as diverged. Every sibling test in this file uses the stronger `convergence.is_converged()`. Capture `replica_count` and `diverged`; assert `treatment.diverged.is_empty()` and `treatment_replicas >= control_replicas`.

### M9. `update_msg_wire_variant_indices_are_frozen`'s stated compile-time guarantee does not exist ✅
`operations/update.rs:1677-1678`

The rustdoc: *"The table is a FIXED-LENGTH array, so adding a variant without extending it is a compile error rather than a silently-passing test."* `let expected: [(u32, UpdateMsg); 6] = [ … ]` is an array literal with no exhaustiveness relationship to the enum; a 7th variant compiles and the test passes. It *does* catch insertion and reordering — the dangerous operation today, so the test earns its keep — but the justification is false and a maintainer will trust it. Worth flagging that **two of five lenses repeated this claim rather than testing it**, which is a fair measure of how convincing the comment is. Derive the table from a wildcard-free `match`, or fix the comment.

### M10. No receive-side bound on `CoveredPeers.hashes`
`ring/broadcast_coverage.rs:119`, `:200` *(code-first, big-picture)*

`MAX_COVERED_PEERS` is enforced only in `from_targets`; `resolve` accepts any length off the wire. The wire does bound it in practice (a short message caps at ~170 hashes, and `resolve` is O(candidates) not O(named), so there is no amplification) — but the cap's rustdoc frames it as an invariant when it is sender self-discipline.

---

## MINOR / NIT

- **The rate-limiter and ban-list source pins were not extended to the new variants** ✅ (`update_rate_limit.rs:711-717`, `contract_ban_list.rs:1107-1113`). Both enumerate "all four wire variants" and say *"If a new UPDATE wire variant was added … update this list."* Two were added. **Not a live bypass** — I checked: the key is extracted from all six variants at `node.rs:1565-1569` *before* both gates — but the guard written to catch an ungated variant failed to fire on the change that added two.
- **`origin.covers()` short-circuits before connection resolution** (`update.rs:411-415`), so a covered-but-unresolvable peer counts as `skipped_covered` rather than `proximity_resolve_failed`, skips the stale-neighbour reap, and can make `fully_covered` true when the remaining candidates all failed to resolve — an operator then sees success where the real story is a stale cache.
- **Retry and #4359 re-emit paths consume coverage they do not own** (`broadcast.rs:274`): `take` runs unconditionally at the top, including for `is_retry` / `is_reemit` re-entries which register nothing. The store's intersection argument protects concurrent *registrations*, not a stale *consumer*. The residual section names only the executor replay and the second emitter; `contract_ops.rs:126`/`:147` belong to the same class.
- **`the_list_a_relayer_emits_describes_only_its_own_sends`** (`op_state_manager.rs:3407-3410`) builds the outbound list itself instead of calling `target_list_for`. Changing `target_list_for` to union `origin.covered` — the exact attack the security rule forbids — leaves it green.
- **Tautological `all(...)`** at `broadcast_coverage.rs:576`: `resolved` is mapped from `candidates`, which *is* `peers`, so no implementation makes it red.
- **`legacy_broadcast_to_encoding_is_unchanged_by_the_v2_variants` is a partial oracle** (`update.rs:1795-1806`): only the variant index and length prefix are literal; `id`, `key`, `payload` are re-serialized with the same encoder, so reordering `DeltaOrFullState`'s variants is a genuine wire break that stays green.
- **`deliveries` captured and never asserted** — its own rustdoc calls it "denominator for the duplicate-delivery ratio" and no ratio is computed, so an arm that halves redundant deliveries *because* it halved total deliveries passes discriminator B.
- **Privacy claim overstated** (`broadcast_coverage.rs:60-70`): tx-seeding defeats cross-transaction correlation, but the recipient holds the seed in the clear and can test membership for any public key it knows — including peers that advertised co-hosting to the *sender* but not to it. One sentence fixes it.
- **Doc inconsistencies**: `:74` says "48-byte hashes" against `PeerHash = [u8; 8]`; `:113` says the cap prevents "a half-kilobyte list" while 64 × 8 + 8 = 520 bytes; `update.rs:333` rustdoc still describes the removed `is_local_update_initiator`; `update.rs:1769` references a test name that does not exist (`a_pre_floor_peer_gets_the_legacy_variant`); a blank line was deleted at `update.rs:328-329`, gluing a doc comment to the previous body.

---

## Verified sound (so the next reader need not redo it)

Wire compatibility is genuinely solid on the substance: both V2 variants are appended last under an explicit APPEND-ONLY marker, no existing variant **or field** moves, and no other enum's discriminants change. The gate fails closed on unknown *and* pre-floor, and `supports_broadcast_target_list_gate_discriminates_by_version` probes four points so neither a stuck-open nor a stuck-closed gate passes. Version lying is self-harming in both directions. One-hop non-propagation holds mechanically — the outgoing list derives solely from the relayer's own `targets`, never unioning or forwarding what arrived. Transaction seeding round-trips correctly on **both** the inline and streaming paths, and the streaming case honours the list only after successful assembly. `only_payload_bearing_variants_carry_a_target_list` was checked for the self-match trap and is clean: its enum-body slice ends before the `Display` impl and the test's own assertion strings. Collision-grinding buys an attacker nothing (it can name the victim's real hash directly), so truncation risk is purely accidental loss, which is under-suppression. `register` intersects rather than replaces and every non-broadcast apply site registers an explicit empty set, so the failure direction is under-suppression throughout. `skipped_covered` and `record_broadcast_delivery` are emitted **by the deciding code**, not re-derived by subtraction at a call site — correct per the "metric describing a filtering decision" row in `bug-prevention-patterns.md`. The dedup-before-`update_contract` ordering holds. The sim's three premises are the right ones and defend against the thread-local counter trap, the default-flip trap and the empty-mesh trap. `fully_covered_fanout_returns_before_the_no_target_retry` is a real ordering pin on real anchors.

## Suggested order of work

1. **B1** — a production-path test, because until it exists none of the other evidence is about the shipped code.
2. **B2, B3, M3** — three small, self-contained code fixes.
3. **B4, B5, M6, M8** — the test repairs, which are what make a re-run of the measurement mean something.
4. **M5** — decide the floor question, which gates the release either way.
5. The rest can follow, but M1's design objection deserves an explicit answer rather than being carried silently.

[AI-assisted - Claude]

sanity:
## Review findings addressed — `de60c2b1a`

Branch rebased onto main and pushed. Every fix below was **mutation-verified**: I reverted the fix, confirmed the intended guard went red, and confirmed nothing else did. A guard I have not watched fail is not a guard I know works, and three of these were originally protected by tests that could not fail.

### BLOCKING

**B1 — production fan-out path had no coverage.** `production_fanout_passes_its_target_list_to_the_queue` pins both anchors, bounded to the non-simulation function so the `cfg(simulation_tests)` twin cannot satisfy it. Both named mutations were applied and confirmed red: passing `no_fanout()` at the enqueue loop (feature inert on the fleet), and deleting `existing.fanout = fanout` on dedup replacement.

**B2 — `no_targets` false alarm.** Added `record_fully_covered_broadcast`, a separate entry point with its own `fully_covered` counter, so the outcome is attributed by the code that decides it. The old pin asserted a call *count* and passed either way; it now pins the separate entry point, and `a_fully_covered_fanout_is_not_counted_as_a_propagation_failure` asserts the behaviour plus the counterfactual (that `record_broadcast(c, 0, _)` really is the failure path — so if that ever stops being true, the two entry points get re-examined).

**B3 (partial) — the stash discard.** The branch no longer destroys an earlier give-up's stash. It refreshes instead: only when something was already stashed does it put the *current* state back. Leaving the old stash would queue stale state; stashing unconditionally would add a write to every fully-covered fan-out, which is the common outcome in the target regime. The retry cancellation is unchanged and remains correct.

**B4 — the tautological assertion.** Deleted. Replaced with a check that the pre-floor branch still encodes as bincode variant index 1 — the property worth asserting next to the gate decision, since a variant an old peer cannot decode does not degrade the feature, it closes the connection.

**B5 — a bandwidth change asserting no bytes.** The simulation now asserts `full_state_sends` and the duplicate *fraction*. Both thresholds come from measurement, not guesswork (below).

### MAJOR

**M3 — the stale-replacement branch was inert.** `expires_at` is replaced rather than `min`-ed, so the fresh entry is no longer born expired. There was **no test exercising `COVERAGE_TTL` at all**, because `Instant` has no seam here — added `insert_with_deadline` plus tests for both the replace and the expiry path.

**M2 — coverage entries were never reclaimed.** `register` now sweeps expired entries above `SWEEP_THRESHOLD`, placed **after** the shard guard drops; doing it inside the `Vacant` arm is the DashMap re-entrancy deadlock this codebase has hit before.

**M5 — release sequencing.** `docs/RELEASING.md` documented the four sibling constant pairs and not this one. Added, including the way to resolve the tripwire *wrongly*: the `Some(..)` arm asserts `shipped == floor && floor <= current`, so setting the marker to `Some((0,2,120))` when the code did not land goes green and records a claim nothing can check. If this slips, **raise the floor**.

### Measurement, reproduced independently on the rebased branch

```
control:   deliveries=440 redundant=321 sends=460 (delta=444 full=16) suppressed=0   summary_skips=317 converged=(1,1)
treatment: deliveries=236 redundant=136 sends=256 (delta=240 full=16) suppressed=500 summary_skips=21  converged=(1,1)
```

Matches the PR body exactly. Two things this settles:

- **The feared byte inversion does not occur in this regime.** `full_state_sends` is **16 in both arms**; the saving is entirely in deltas (444 → 240). So B5 was a coverage gap rather than a live bug — and the new assertion holds with margin rather than by a hair, which is why it is safe to add now instead of speculative.
- **The redundancy win is not a denominator effect.** Deliveries fell 46% while redundant deliveries fell 58%, so the duplicate fraction dropped 73.0% → 57.6%. The old assertion could not tell those apart; the new ratio assertion can.

### Still open — not patched, and one needs a decision rather than a patch

- **M1 (decision needed).** The revived sender/self exclusion is **not** version-gated and **not** covered by the sim opt-out, so `disable_broadcast_target_list()` does not restore pre-#5147 behaviour and *both* sim arms carry it — the control arm is therefore not main's behaviour, and the reported 44% is the covered-list half only (conservative, but the arm's described role is inaccurate). It ships as an unconditional fleet-wide change with no floor and no runtime switch. The big-picture lens also argues "sending it back to the sender is a guaranteed duplicate" is false under CRDT merge, since the relayer can be strictly ahead and `proactive_summary_targets` also excludes the sender. **Worth weighing on the other side:** this half works against legacy peers, making it the one part of #5147 that saves bandwidth unilaterally at 0.2.120 rather than waiting for fleet adoption.
- **B3 remainder.** `fully_covered` still fires on `skipped_sender > 0` alone, which is true on essentially every relayed broadcast.
- **M6** (`assert_eq!(control_legs, treatment_legs)` is an exact equality on a quantity that is not conserved — a flake risk under this repo's own flaky-tests-are-blockers rule), **M7** (a failed send still publishes that peer as covered), **M8** (the convergence assertion reduces to `1 >= 1`), **M4**, **M10**, and the MINOR list.

`cargo fmt` clean, `cargo clippy --locked -p freenet` clean (the `--all-targets` errors are pre-existing on main — same `let _ =` pattern present at the base commit), full lib suite 4700 passed / 0 failed, and the #5147 simulation green with the new assertions.

One unrelated thing found and fixed in passing: the shared `EMISSION_SITES` allowlist walker scrapes for `notify_node_event(` textually without skipping comments, so a new *doc comment* tripped it. Allowlisting the file would have told future readers it emits gossip; reworded the prose instead and left a note.

[AI-assisted - Claude]

sanity:
## Which transaction the hash comparison uses (asked explicitly, answering explicitly)

**It uses the transaction of the message that delivered the payload, never the transaction of the re-broadcast.** This matters because if the two ever diverged, every comparison would silently resolve to no match, the feature would be inert, and every test that pins the *gate* would still pass.

Where it is enforced:

- `resolve_covered_peers(op_manager, &key, &incoming_tx, sender_addr, &covered)` — `operations/update/op_ctx_task.rs:1859` (non-streaming) and `:2788` (streaming). Both pass `incoming_tx`, which is the `id` off the inbound `BroadcastToV2` / `BroadcastToStreamingV2`.
- Resolution happens at **receive** time, inside the relay driver, before the apply. The re-broadcast's own transaction is minted later and per-leg, in `broadcast_to_single_peer`, and is structurally unreachable from the resolution site — there is no code path by which the wrong one could be substituted.
- The outgoing list is a separate construction from the relayer's own targets under its own per-leg tx (`target_list_for`), so the inbound and outbound hash spaces never meet.

Behavioural coverage, not just structural: `a_list_replayed_onto_another_transaction_suppresses_nothing` (`op_state_manager.rs`) and `a_list_does_not_resolve_under_a_different_transaction` (`ring/broadcast_coverage.rs`) both assert that a list resolved under any other transaction yields the empty set. So the "silently inert" failure mode fails a test rather than passing quietly.

This also gives the replay defence mechanically rather than by rule: a list lifted from one broadcast and replayed onto another cannot match, because the seed differs.

## Full-tier review status

Five blind lenses ran (code-first, skeptical, testing, big-picture, and one aimed at the change-specific failure — a relayer wrongly suppressing a peer that needed the update). Findings fixed in `bb3722397` and the commits before it; the two design-level findings are in the PR body above, one of them **unimplemented and blocking**.

**The review is stale as of now** and must be re-run before merge: the branch was rebased and had commits pushed to it by a second agent while the review was running, so no lens has read the current content. Per `~/.claude/rules/multi-model-review.md` that is not a formality — the conflict resolutions themselves were unreviewed judgement calls, and two of them chose between competing fixes for the same finding.

[AI-assisted - Claude]

sanity:
## Branch ownership — single writer from here

**I am the sole writer to `fix/5147-originator-target-list` from now on.** If you have changes for this PR, please post them as review comments rather than pushing — I will apply them and attribute them.

This is not territorial. Two of us were pushing to this branch concurrently: we independently duplicated four review fixes, diverged twice, and the branch moved underneath a rebase that was already in progress. The concrete cost is that **a Full-tier review is per-code-content, so with two writers every review is stale the moment it finishes** — and on a protocol change with a P0 and a P1 in its history, that is the thing we cannot afford to be sloppy about.

Cross-account agent messaging is scoped per session, so this PR is the only shared surface. Hence a comment rather than a DM.

### Your work is preserved — nothing was force-pushed

Head is now `bb3722397`, rebased onto `d7f21690c` with all six of your commits intact beneath it. Where we had both fixed the same finding differently, I took whichever was better rather than defending mine:

- **Yours won on the #4359 stash.** Refresh-if-present beats my leave-it-alone: mine would have kept STALE state queued for re-emission, since this apply's state is newer. I updated my two pin tests to assert *your* behaviour.
- **Yours won on the pre-floor assertion.** Replacing the tautological byte comparison with a real one (bincode variant index == 1) is better than my deleting it and pointing elsewhere.
- **Yours won on the propagation counter.** A dedicated `record_fully_covered_broadcast` entry point is cleaner than my `record_broadcast(key, 0, ..)`, which landed in the operator-facing `no_targets` failure bucket.

### Two things from my side that must not be lost in any future rebase

Both are the same failure mode as the blocking design finding on this PR: **a path that reports success while having delivered nothing.**

1. **`fully_covered` requires `proximity_resolve_failed == 0`.** Without it, a fan-out that is empty *only because every co-host lookup failed* is reported as a completed broadcast — genuine unreachability recorded as success, with the retry that would have healed it skipped.
2. **`fully_covered` must not simply discard the #4359 stash.** It sends zero bytes, so the stashed state reached nobody, and the peers it suppressed hold the *originator's* payload, which does not contain that state. Discarding it deletes the only non-heartbeat recovery route for content that has never propagated. Your refresh-if-present is the right shape; the thing to preserve is that it is not a bare `take`.

### Still open before this can merge

- **BLOCKING:** the merge-semantics gap (see the PR body). Agreed fix is to suppress only when our post-merge summary equals the sender's, compared at receive time. Unimplemented.
- The revived sender exclusion ships ungated and must be gated like the target list or dropped.
- The Full review is stale and will be re-run on final content once the branch is single-writer.

[AI-assisted - Claude]

sanity:
## Option 3 does not drop in at the receive site — and the reason points at where it does belong

Agreed direction was: *suppress only when our post-merge summary equals the sender's, compared at receive time where the driver already holds the merged value.* I went to implement that and it does not work at that site. Writing up why, and the placement that does work, so nobody re-derives this.

### Why receive-time comparison cannot gate the claim

The coverage claim is registered inside `update_contract`, **before** the `UpdateQuery` enters the contract handler. That ordering is load-bearing — it is what guarantees the claim is present when its own fan-out reads it.

But the fan-out event is emitted from **inside** that same round trip: `executor_impl.rs:1999`, in `commit_state_update`, while the handler is still processing the query. So by the time `update_contract` returns to the relay driver — the first moment the merged value exists — `NodeEvent::BroadcastStateChange` is already on the channel and `handle_broadcast_state_change` may already have consumed the claim.

Concretely, all three shapes fail:

- **Compare after `update_contract` returns, then `discard()` on mismatch.** Racy: the fan-out may have taken the claim already. The window is the same one the store's `narrow()` was built to survive, except here losing the race means honoring a claim we have just decided is unsound — over-suppression, the exact failure the design refuses.
- **Compare before the apply.** Our state is pre-merge; the sender's summary describes their post-merge state. Not comparable.
- **Compare at the fan-out site.** `handle_broadcast_state_change` has no local summary, and fetching one is a contract-handler round trip **on the p2p_protoc event loop** — the #4145 wedge, and squarely what `.claude/rules/channel-safety.md` forbids.

### Where it does belong: the send site

`broadcast_to_single_peer` **already computes our summary**, at `broadcast_queue.rs:3530`, with no extra round trip and off the event loop. That is the natural home, and it is also where the existing `fanout_send_needed` summary gate lives — which is not a coincidence. Part of the seeding interaction reported earlier is that suppression currently happens *before* the queue and therefore bypasses that gate; moving the decision to the send site puts the two mechanisms back in the same place.

Shape:

1. Carry `sender_summary_bytes` in `BroadcastOrigin` alongside `sender` and `covered` (it is already on the wire, already parsed by the relay driver).
2. Thread it to the send site the same way `fanout` is threaded — `handle_broadcast_state_change` → `enqueue` → `BroadcastEntry` → `broadcast_to_single_peer`. That plumbing exists and is tested (`the_queue_carries_the_target_list_and_refreshes_it_on_replacement`).
3. At the send site, honor the target list **only if** `our_summary == sender_summary`. Otherwise ignore it and send as today.

That makes the invariant exact rather than bounded: if we hold precisely what the sender holds, then everyone the sender covered is genuinely covered by us. It is a proof, not a heuristic — the same shape as choosing intersection over latest in the store.

Two consequences worth stating up front:

- **Suppression moves from target-resolution to send time.** `skipped_covered` stops being a `get_broadcast_targets_update` counter. The `fully_covered` early return, which exists because a fully-suppressed fan-out has an empty target set, will need rethinking — at send time the target set is no longer empty, the legs are just dropped later. **Do not delete `fully_covered` without re-deriving that path**; the retry-and-re-broadcast regression it fixes is documented above and cost a full measurement cycle to find.
- **It is a real restructure, not a patch**, on a protocol path with a P0 and a P1 in its history. It needs its own Full review, and the control/treatment run repeated — the numbers in this PR describe suppression at target-resolution time and do not carry over unchanged.

### Status

**Not implemented.** I am stopping here rather than starting a restructure of this size with limited capacity remaining; a half-applied change to the suppression site would be worse than none, and this write-up is worth more than a partial diff. The PR remains blocked on it, as it should be.

[AI-assisted - Claude]

sanity:
## Status update — one blocking item closed, one remains

Head is `31729f41a`. `cargo fmt`, `cargo clippy --locked -- -D warnings`, and 4707 lib tests all clean.

### Closed: the sender exclusion is no longer ungated

It needed no wire change of its own, so it was shipping unconditionally to every peer on the floor release while the target list beside it was gated. That was backwards — it has the **same** unsafe shape as the list, since excluding the peer that delivered a payload assumes it holds what we are about to send, and on a merge path our post-merge state can be strictly newer than what it sent.

Now gated on the **sender's** version: if the peer that delivered this payload predates the feature, we are in the legacy world with it and behave toward it exactly as today. Unknown fails closed, which is the common case on gateway links until #5161 deploys. `the_sender_exclusion_is_version_gated_and_fails_closed` asserts both directions — the at-floor half is load-bearing, because without it an always-closed gate would pass while the feature sat inert.

### Still blocking

The merge-semantics gap. See the previous comment for why option 3 does not drop in at the receive site and where it does belong (the send site, where `our_summary` already exists). **Unimplemented, and this PR should not merge without it.**

### Also outstanding

- The Full review must be re-run on final content. It is stale: the branch was rebased and had commits from a second agent land mid-review. Two of the conflict resolutions chose between competing fixes for the same finding and are themselves unreviewed.
- The heartbeat-spaced simulation run was never completed — the recipe is in the issue comment. It answers whether the seeding interaction is regime-dependent, which matters more now that the plan is to ship rather than defer.
- The simulation needs multiple writers and multiple contracts before its numbers can speak to lost updates at all.

[AI-assisted - Claude]

sanity:
## Caught before merge: the version gate had made this feature completely inert

Pushed as `628688eca`. Flagging prominently because the failure mode is the one that looks like success.

### What happened

`31729f41a` gated the sender exclusion behind the #5147 version floor. The intent is right — the exclusion has the same unsafe shape as the list, so it should not inherit the list's caution without its gate. But it routed **both** halves of a relayed coverage claim through one check:

```rust
if !supports_broadcast_target_list(sender_addr) {
    return BroadcastOrigin::local();   // discards the list too
}
```

That reads as symmetric caution. It is not. **A node does not learn its gateway's version until #5167 propagates**, and simulations default that ack gate OFF. So the send side opened — a diagnostic run emitted `BroadcastToV2` 1242 times — while the receive side failed closed on the same links, discarded every list, and suppressed nobody.

On the merged branch both arms came out byte-identical:

```
control:   sends=460 redundant=321 suppressed=0
treatment: sends=460 redundant=321 suppressed=0
```

The feature did nothing at all. **Only the end-to-end simulation caught it.** Every unit test stayed green, including `the_sender_exclusion_is_version_gated_and_fails_closed` — it drives the gate directly with a recorded version, so it never sees the case where no version was ever recorded.

Worth noting this is the same shape as the BLOCKING finding the review already raised: the production path had no coverage, and a one-token change could make the feature inert with everything green. This was that, arriving through a different door.

### The fix

The two halves carry different evidence, so they get different gates:

* **The covered LIST is self-gating.** It rides only on `BroadcastToV2` / `BroadcastToStreamingV2`, so a pre-floor sender hands us an empty one and suppresses nothing whatever we do. Holding a populated list *is* proof the sender supports the feature — the sender's own act, which is strictly stronger evidence than our record of its version.
* **The sender EXCLUSION keeps the version gate.** It has no wire signal of its own. `BroadcastOrigin::relayed_list_only` is the new origin: a live list, no sender to exclude.

The gating commit's intent is preserved exactly — a pre-floor sender still gets today's behaviour, because it cannot send us a list in the first place. What changes is that we stop discarding claims the message in hand has already proven honourable.

### Measured after the fix

Now genuinely apples-to-apples: with the sender exclusion gated, the control arm is main's behaviour rather than main-plus-an-ungated-change.

```
single-writer  control:   sends=460 redundant=321 suppressed=0   replicas=7 diverged=0
single-writer  treatment: sends=288 redundant=160 suppressed=428 replicas=7 diverged=0

multi-writer   control:   sends=459 redundant=319 suppressed=0   replicas=7 diverged=0
multi-writer   treatment: sends=420 redundant=288 suppressed=229 replicas=7 diverged=0
```

**−37% sends / −50% redundant deliveries single-writer; −8.5% / −10% multi-writer.**

### The multi-writer arm, and what it settles

New in this push, and it answers the design question that was left open for a decision: the list attests DELIVERY of the originator's payload, but a relayer re-broadcasts its own POST-MERGE state, which under concurrent writers can be strictly newer. Suppressing on delivery could therefore skip a leg carrying something the recipient genuinely lacked.

The existing measurement could not see this — **every one of its 20 updates originates from the gateway**, so all peers merge the same payload into the same prior state and suppression is correct by construction. That is the regime most favourable to the feature.

Rotating the writer across peers makes relayer-ahead-of-recipient reachable every round. Result: **convergence and replica count are unchanged** (replicas=7, diverged=0 in both arms). The concern is real in principle and does not bite in practice here.

It also shows something the single-writer number alone would have overstated: **the benefit is strongly regime-dependent**, roughly 37% versus 8.5% on the same topology. Real River traffic is multi-writer, so the lower figure is likely the better guide to fleet impact — though this 7-peer graph's overlap is not the fleet's degree-16-17 overlap either, and the 0.894-0.924 suppression measured offline is a topology property rather than a workload one.

The new test asserts SAFETY only (divergence, replica count), not bandwidth. The single-writer test owns the bandwidth claim; a second sample of the same mechanism would not add evidence.

### Two smaller corrections in the same push

* **`record_broadcast_sender_skipped`.** The sender exclusion is a FOURTH terminal outcome for an offered leg and was counted nowhere. Recorded at the decision site, per the "metric must come from the decision" rule.
* **The leg-conservation equality was never a conservation law**, and the code-first lens said so before the numbers did. Sender skips were one leak; legs also exit via a failed send (`sends` counts only successes), the empty-delta return, `should_broadcast_contract`, and queue eviction. With all four buckets the arms sit at 877 vs 869 — under 1% — so it is now a 5% bound. Still a real discriminator against a topology difference, without being a tripwire that fires on one extra send failure.

`cargo fmt` clean, `cargo clippy --locked` clean, full lib suite 4707 passed / 0 failed, both simulation tests green.

[AI-assisted - Claude]

sanity:
## Re-review of the unreviewed delta, and a retraction

The Full-tier review covered `086f3e71d`. Ten commits landed after it, including a behavioural change to version gating that only its author had read — so the review was stale in exactly the way `multi-model-review.md` warns about. Re-reviewed with two fresh blind lenses. Both found real defects. Pushed as `53f80886d`.

### Retraction first

I reported earlier in this thread that the multi-writer arm **answers** the open design question — that suppressing on delivery is safe when a relayer's post-merge state can be newer than the payload it received. **It does not, and I should not have said it did.**

The testing lens named the mutation that settles it. I ran it: make suppression **unconditional** — drop every advertised co-host on any relayed fan-out, ignoring the list. 480 suppressions, sends 459 → 317, and the test stays **GREEN** at `diverged=0`, `replicas=7`.

Three properties of the setup defeat it, each sufficient alone:

- **Topology** — 7 peers at `max_connections: 12` is near-clique, so no peer depends on a relayer and over-suppressing relay legs costs nobody state.
- **Merge semantics** — the CRDT fixture is an LWW register with strictly increasing versions, so missing an *older* update is harmless by construction; and `apply_crdt_full_state` returning `CurrentWon` re-broadcasts, actively healing the case being probed.
- **Settle window** — convergence is checked after 90 simulated seconds of quiet.

That is the same "verification that cannot fail" pattern this review has been enforcing elsewhere, in the test written to close the review's own hardest finding. The test is renamed `test_5147_multi_writer_suppression_is_regime_dependent` and its rustdoc now states what it does establish and what it does not, with the vacuity mutation recorded so nobody re-derives it. **The design question is open: #5186**, with the three changes that would close it.

### The other substantive finding

`resolve_covered_peers` had exactly two callers in the entire test tree, and both passed an **empty** `CoveredPeers` — so "is the list honored when the sender's version is unknown?" was untested, and every other #5147 test constructs a `BroadcastOrigin` directly, bypassing the function. Re-applying the inertness regression left the whole unit suite green; only a 40-second 7-node simulation caught it.

`a_covered_list_is_honored_even_when_the_sender_version_is_unknown` now covers it in **0.02s**. The regression was re-applied and confirmed to redden it.

### Also fixed

- **A fourth doc splice.** Same mistake as the three already fixed: an item inserted directly above another silently steals its doc comment. This one severed `fully_covered_fanout_returns_before_the_no_target_retry`'s rustdoc mid-sentence, leaving it documented as the fragment "this test FAILS." and moving its mutation-check instruction onto an unrelated pin. Rustc accepts this and nothing catches it.
- **The receive-side `take(MAX_COVERED_PEERS)` had zero coverage** — every test builds `CoveredPeers` through `from_targets`, which truncates first, so deleting the cap made nothing red. The new test builds the struct directly, which is the real threat model (an oversized list off the wire). Mutation-verified.
- **Stale numbers in two discriminator comments.** DISCRIMINATOR E claimed 16 == 16 held "with margin" — under a `<=` comparison equal counts pass by exactly one step, and more suppression is precisely the direction that pushes full-state up. It now says so and names the first hypothesis if it ever reddens.
- **The leg tolerance passed vacuously** if both arms did zero fan-out (`0 <= 0`); premise added.
- `a_pre_floor_peer_receives_byte_identical_broadcast_traffic` no longer asserts byte-identity, so it is renamed to `..._receives_the_legacy_variant_index`.
- The `#4359` stash recheck re-passed the consumed claim, firing both suppression counters twice for one fan-out while `sends` counted once — a one-sided inflation of the treatment arm only, inside the very comparison meant to detect arm asymmetry. Probes with a local origin now.
- The rationale for keeping the sender exclusion gated overclaimed: not sending a peer a message cannot break its decoding, and the semantic worry is version-independent. It is a rollout switch, and the comment now says so.

### Where this leaves the PR

`cargo fmt` clean, `cargo clippy --locked` clean, 4709 lib tests pass, both simulations green.

**Current measurements, apples-to-apples now that the sender exclusion is gated:**

```
single-writer: sends 460 -> 288 (-37%), redundant 321 -> 160 (-50%)
multi-writer:  sends 459 -> 420 (-8.5%), redundant 319 -> 288 (-10%)
```

**I am not merging this on my own judgement, and the reason is not a red signal.** CI was green through every discovery above, including the several commits where the feature did nothing at all. In the final stretch, adversarial reading found a live regression *and* that my own verification of the fix was incapable of failing. Immediately after that is a poor moment to self-approve a protocol change carrying a permanent wire surface — particularly with #5186 now marking the central safety question as explicitly unmeasured rather than answered.

The judgement call for @sanity is narrow: **is −8.5% on realistic multi-writer traffic worth the permanent wire-format surface now, or should #5186 close first?** Everything else is done, and the release is prepared either way — including the detail that 0.2.120 needs **two** version markers flipped rather than the one the release notes previously recorded.

[AI-assisted - Claude]

PR #5183, opened 2026-08-05
fix(put): never let an op-execution payload displace a live reply waiter

### LINKED ISSUE #5182: A remote WS client receives `failed notifying, channel closed` for a PUT that succeeded
## Summary

When the WebSocket client runs on a **different machine** from the node and the node's
event loop is contended, a PUT can return

```
put failed after 1 peer attempt(s) (3 infra-retries on same peer;
last error: failed notifying, channel closed)
```

for an operation that **completed successfully** — the state is stored on the node and
replicated to its peers, byte-identical, confirmed by an independent read-back. In every
sample taken, *every* failing PUT had stored correctly. Not one was an accurate report of a
failure.

`"failed notifying, channel closed"` is the string `crates/core/tests/error_notification.rs`
names `FORBIDDEN_MARKER`, with the comment: *"a response containing this string means the
client got an infrastructure leak rather than the contract's real rejection reason."*

**This is not #4111.** That issue is closed and its fix is present at the commit tested; it
concerns a *real* `PutMsg::Error` being dropped at dispatch so the client saw the synthesised
marker instead of the contract's actual reason. Here there is **no error at all** — the
operation succeeds. Same user-visible string, different route.

## Mechanism

`handle_op_execution` installs an op-execution payload's reply callback into
`pending_op_results[tx]`, guarded only on the **incoming** callback:

```rust
if callback.is_closed() { /* skip */ }
else { state.pending_op_results.insert(*msg.id(), callback); }
```

It never asks whether a **live waiter is already registered for that tx**.

Producers such as `OpCtx::send_local_loopback` and `OpCtx::send_fire_and_forget` construct a
throwaway channel and drop its receiver:

```rust
let (response_sender, _response_receiver) = mpsc::channel::<WaiterReply>(1);
self.op_execution_sender.send((response_sender, msg, None)).await
```

`_response_receiver` is a **named binding**, so it drops when the producer's poll finishes —
*after* the `.await` hands the sender over. The event loop runs on another thread and can
dequeue the payload inside that enqueue→drop window and observe `is_closed() == false`.

The insert then **displaces the originator's live waiter**. Its sender drops, the driver's
`recv()` in `recv_waiter_reply` yields `None`, which is `OpError::NotificationError`; the
three `MAX_INFRA_RETRIES` attempts each re-lose the same race; and the client is told the PUT
failed — after it had already succeeded.

## Reproduction (deterministic, ~2 commands)

The attached PR adds a unit test with **no timing and no threads**. From the PR branch:

```bash
gh pr checkout 5183

# GREEN with the fix:
cargo test -p freenet --lib op_execution_payload_must_not_clobber_live_waiter

# Undo only the fix commit (leaves the test in place) -> RED:
git revert --no-commit HEAD
cargo test -p freenet --lib op_execution_payload_must_not_clobber_live_waiter
```

RED output on `main`:

```
panicked at crates/core/src/node/network_bridge/p2p_protoc.rs:
originator's waiter channel was closed by the dummy-callback install:
the driver's recv() yields None -> NotificationError -> the client sees
'failed notifying, channel closed' for a PUT that succeeded
test result: FAILED. 0 passed; 1 failed
```

The test models the race by holding the dummy receiver alive across the install call, then
dropping it, and asserts (a) the originator's channel is still connected and (b) the terminal
reply actually reaches it.

## How it was found in the field

Not reproducible on one machine. Setup was a gateway on a VPS and a client on a workstation
~130 ms apart over WireGuard, with the node's own logs instrumented (the no-waiter branch of
`try_forward_driver_reply` returns `false` with **no log statement at any level**, so no log
level could surface it — the node had to be patched).

One failing PUT, traced:

```
22.987881  PUT relay: no next hop, finalizing at this node   <- the PUT SUCCEEDS
22.988056  releasing pending_op_results slot  attempt=1
22.988118  try_send to waiter FAILED  err=channel closed
23.417     attempt=4 -> send_and_await failed  outcome="wire_error"
23.417651  Delivering result to 1 clients                    <- client told it FAILED
```

The reply arrives **62 µs** after its waiter is torn down; the client hears "failed"
**430 ms** after the operation succeeded.

Supporting observations:

| observation | result |
|---|---|
| `try_send` failure kind | **`err=channel closed` in 118 of 118** — never `Full` |
| "no callback registered" branch | fires **0 times** — the waiter is registered; its *receiver* is gone |
| client co-located with node | **0 / 12** failures |
| client remote, node unloaded | **0 / 25** failures (2 drops) |
| same arrangement, node pinned to 1 core | **9-19 / 50** failures across runs (82-100 drops) |
| PUT that merges to a no-op | ~**1 / 20** |

That 9-19 spread includes two runs of **identical** code that returned 9 and 19. At n=50 the
rate is not resolvable, which is why the only stable artefact offered here is the unit test.

The drop is normally survivable — in one control run 100 `try_send` drops produced 20
`NotificationError`s and 10 client-visible failures; the infra-retries usually recover.

## Why the existing guards do not catch it

Every test guarding `FORBIDDEN_MARKER` builds its client against a socket bound to
`127.0.0.1:0` (`error_notification.rs`). A co-located client keeps the enqueue→drop window
shut. The guards are sound; the scenario is out of their reach.

## Suggested fix

Do not let an op-execution payload displace a waiter that is still live. The incoming
callback's own `is_closed()` cannot be the test, because it is read inside the very window
that makes it unreliable. The attached PR checks the incumbent instead; a payload whose own
callback is closed is still skipped, so the #3100 bound on `pending_op_results` is unaffected.

Other shapes are possible (fixing each producer to drop its receiver before the send;
re-tuning `MAX_INFRA_RETRIES`), and the maintainers may prefer one of those — a producer-side
fix was tried here first and is both whack-a-mole (more than one producer) and untestable
without timing.

## Not claimed

- That this is the only producer of the marker. #4111 was another, already fixed.
- That any *rate* here characterises anything beyond this hardware and this link. The counts
  above are what was observed, not a measurement: identical code returned 9 and 19 failures
  per 50 on consecutive runs, so nothing at this sample size separates signal from noise.
  Only the unit test is stable.
- That latency as such is the trigger. A `netem` experiment silently failed to apply and was
  not retried, so RTT is not separated from event-loop contention.

## Environment

```
freenet 0.2.118 (738f99f0e2e7) — field observation
freenet main @ 998b730          — mechanism confirmed present, test RED, fix GREEN
full suite with fix: cargo test -p freenet --lib  ->  4671 passed; 0 failed; 28 ignored
cargo fmt --all --check: clean;  cargo clippy -p freenet --lib -D warnings: clean
```

magks:
Fixes #5182.

## What breaks today

`handle_op_execution` installs a payload's reply callback into `pending_op_results[tx]`
guarded only on the **incoming** callback — it never checks whether a **live waiter is
already registered** for that tx:

```rust
if callback.is_closed() { /* skip */ }
else { state.pending_op_results.insert(*msg.id(), callback); }
```

`send_local_loopback` / `send_fire_and_forget` build a throwaway channel and drop its
receiver, but `_response_receiver` is a **named binding**: it drops when the producer's poll
finishes, *after* the `.await` that hands the sender over. The event loop, on another thread,
can dequeue the payload inside that window and see `is_closed() == false`. The insert then
displaces the originator's live waiter — its sender drops, `recv_waiter_reply` gets `None`
→ `OpError::NotificationError`, the three infra-retries re-lose the same race, and the client
receives `FORBIDDEN_MARKER` for a PUT that **stored and replicated correctly**.

## The change

1. **`test(put)`** — a deterministic reproduction, no timing and no threads. It models the
   window by holding the dummy receiver alive across the install, then dropping it, and
   asserts the originator is still connected *and* actually receives the terminal reply.
2. **`fix(put)`** — check the incumbent: if `pending_op_results[tx]` holds a waiter that is
   still live, do not replace it. A payload whose own callback is closed is still skipped, so
   the #3100 bound on the map is unaffected.

The fix commit also extracts `install_pending_op_callback` out of `handle_op_execution` so
the decision is unit-testable without a full `P2pConnManager`, following the shape of the
existing `fail_awaiting_connection` helper.

## Verifying RED → GREEN

```bash
gh pr checkout 5183
cargo test -p freenet --lib op_execution_payload_must_not_clobber_live_waiter   # GREEN
git revert --no-commit HEAD                                                    # undo the fix only
cargo test -p freenet --lib op_execution_payload_must_not_clobber_live_waiter   # RED
```

## Checks

```
cargo test -p freenet --lib            4671 passed; 0 failed; 28 ignored
cargo fmt --all -- --check             clean
cargo clippy -p freenet --lib -D warnings   clean
```

Verified against `main` @ 998b730, not only against the 0.2.118 tag where it was first
observed.

## Notes for review

- **This is not #4111.** That fix is present and is about a *real* `PutMsg::Error` being lost
  at dispatch; here the operation succeeds and there is no error to forward. Two routes to
  the same string.
- **Why not fix the producers?** Dropping `_response_receiver` before the send was tried
  first. It is whack-a-mole — there is more than one producer — and it could not be shown to
  help: 9 and 19 failures per 50 against a control of 10, with identical code varying 9→19
  between consecutive runs. The consumer-side guard closes every producer at once and is
  testable without timing.
- **Why the existing guards miss it:** every `FORBIDDEN_MARKER` test binds `127.0.0.1:0`, and
  a co-located client keeps the window shut. Field observation needed a client on a separate
  machine from the node.
- The `try_forward_driver_reply` no-waiter branch returns `false` with **no log at any
  level**, so this is invisible to log-level tuning; the node had to be patched to see it.
  Happy to add a `debug!` there in a follow-up if wanted.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -4457,6 +4457,7 @@ std::thread_local! {
     static GLOBAL_PENDING_OP_INSERTS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PENDING_OP_REMOVES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_PENDING_OP_HWM: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
+    static GLOBAL_PENDING_OP_SKIPS: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     static GLOBAL_NEIGHBOR_HOSTING_UPDATES: std::cell::Cell<u64> = const { std::cell::Cell::new(0) };
     /// Recipients dropped from a proactive summary notification because they
     /// are advertised co-hosts the broadcast already covered (#4965).
@@ -4556,6 +4557,7 @@ impl GlobalTestMetrics {
         GLOBAL_REDUNDANT_BROADCAST_DELIVERIES.with(|c| c.set(0));
         GLOBAL_FULL_STATE_SENDS.with(|c| c.set(0));
         GLOBAL_PENDING_OP_INSERTS.with(|c| c.set(0));
+        GLOBAL_PENDING_OP_SKIPS.with(|c| c.set(0));
         GLOBAL_PENDING_OP_REMOVES.with(|c| c.set(0));
         GLOBAL_PENDING_OP_HWM.with(|c| c.set(0));
         GLOBAL_NEIGHBOR_HOSTING_UPDATES.with(|c| c.set(0));
@@ -4782,6 +4784,20 @@ impl GlobalTestMetrics {
 
     pub fn pending_op_removes() -> u64 {
         GLOBAL_PENDING_OP_REMOVES.with(|c| c.get())
+    }
+
+    /// A waiter install was refused because a LIVE incumbent already held the tx.
+    ///
+    /// Expected to stay at zero: the invariant is one live waiter per tx per
+    /// node, so a non-zero count means either a genuine collision or that the
+    /// guard is refusing a legitimate waiter — the latter otherwise surfaces
+    /// only as a driver retry with nothing pointing back here.
+    pub fn record_pending_op_skip() {
+        GLOBAL_PENDING_OP_SKIPS.with(|c| c.set(c.get() + 1));
+    }
+
+    pub fn pending_op_skips() -> u64 {
+        GLOBAL_PENDING_OP_SKIPS.with(|c| c.get())
     }
 
     /// Track high-water mark for pending_op_results size.
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -1087,6 +1087,14 @@ fn try_forward_driver_reply(
     op_label: &'static str,
 ) -> bool {
     let Some(callback) = pending_op_result else {
+        // Was silent at every level. Finding the clobber bug required patching
+        // a node to see this branch at all, because "no waiter is registered"
+        // is exactly the symptom a displaced waiter produces.
+        tracing::debug!(
+            tx_id = %reply.id(),
+            %op_label,
+            "try_forward_driver_reply: no waiter registered for this tx; dropping the reply"
+        );
         return false;
     };
     let tx_id = *reply.id();
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -6387,4 +6387,94 @@ pub(crate) mod tests {
              just its call site. Block:\n{body}"
         );
     }
+    /// Deterministic reproduction of the "failed notifying, channel closed"
+    /// leak on originator-loopback PUT (the #4111 forbidden marker,
+    /// v0.2.118 field reports).
+    ///
+    /// Sequence modelled (no timing, no threads):
+    ///   1. The originator's `send_and_await` registers a LIVE waiter under
+    ///      `tx` (`install_pending_op_callback` observes a live callback and
+    ///      inserts it).
+    ///   2. The relay driver — running on the SAME node (originator
+    ///      loopback) — emits a payload for the SAME `tx` through the
+    ///      op-execution channel (`send_fire_and_forget` forward, or
+    ///      `send_local_loopback` Response). Its dummy callback is designed
+    ///      to be observed closed, but the receiver half only drops when the
+    ///      producer's poll finishes; the event loop runs on a different OS
+    ///      thread and can dequeue the payload inside the enqueue→drop
+    ///      window (observed in the field on a starved 1-CPU gateway:
+    ///      `is_closed()` returned false). That state is modelled here by
+    ///      holding the dummy receiver alive across the install call.
+    ///   3. The producer's poll finishes: the dummy receiver drops.
+    ///   4. The terminal Response arrives; the bypass looks up
+    ///      `pending_op_results[tx]` and forwards via
+    ///      `try_forward_driver_reply`.
+    ///
+    /// Pre-fix behaviour: step 2's insert CLOBBERS the live waiter, dropping
+    /// the originator's sender — the driver's `recv()` yields `None`
+    /// (surfaced as `OpError::NotificationError`, i.e. "failed notifying,
+    /// channel closed"), and step 4's `try_send` hits the closed dummy
+    /// (`err=channel closed` with a callback still registered). The PUT
+    /// itself succeeded — an infrastructure leak misreported as a failure.
+    ///
+    /// Contract pinned: an op-execution payload must NEVER displace a LIVE
+    /// `pending_op_results` waiter.
+    #[tokio::test]
+    async fn op_execution_payload_must_not_clobber_live_waiter() {
+        use crate::message::MessageStats; // for `msg.id()`
+        use crate::transport::ExpectedInboundTracker;
+        use tokio::sync::mpsc::error::TryRecvError;
+
+        let mut state = super::EventListenerState::new(ExpectedInboundTracker::empty_for_test());
+        let tx = crate::dev_tool::Transaction::new::<crate::operations::put::PutMsg>();
+
+        // Step 1: the originator's live waiter (send_and_await's callback).
+        let (originator_sender, mut originator_rx) = mpsc::channel::<crate::node::WaiterReply>(1);
+        super::dispatch::install_pending_op_callback(&mut state, tx, originator_sender);
+        assert!(
+            state.pending_op_results.contains_key(&tx),
+            "precondition: the originator's live waiter must be installed"
+        );
+
+        // Step 2: the relay's dummy callback for the SAME tx, observed while
+        // its receiver is still alive.
+        let (dummy_sender, dummy_receiver) = mpsc::channel::<crate::node::WaiterReply>(1);
+        super::dispatch::install_pending_op_callback(&mut state, tx, dummy_sender);
+
+        // Step 3: the producer's poll finishes — the dummy receiver drops.
+        drop(dummy_receiver);
+
+        // The originator's driver must still be wired: its `recv()` must NOT
+        // resolve to `None` (that is `OpError::NotificationError` — the
+        // forbidden "failed notifying, channel closed" leak).
+        match originator_rx.try_recv() {
+            Err(TryRecvError::Empty) => {} // still connected, still waiting — correct
+            Err(TryRecvError::Disconnected) => panic!(
+                "originator's waiter channel was closed by the dummy-callback \
+                 install: the driver's recv() yields None -> NotificationError \
+                 -> the client sees 'failed notifying, channel closed' for a \
+                 PUT that succeeded (the #4111 forbidden marker)"
+            ),
+            Ok(other) => panic!("no reply was sent yet, got {other:?}"),
+        }
+
+        // Step 4: the Response arrives and the bypass forwards it to the
+        // registered callback (mirrors `process_message`'s clone-then-forward).
+        let reply = crate::message::NetMessage::V1(crate::message::NetMessageV1::Aborted(tx));
+        let cb = state.pending_op_results.get(&tx).cloned();
+        let handled = crate::node::try_forward_driver_reply(cb.as_ref(), reply, "put");
+        assert!(handled, "a callback must be registered for the tx");
+
+        // The originator must actually RECEIVE the terminal reply.
+        match originator_rx.try_recv() {
+            Ok(crate::node::WaiterReply::Reply(msg)) => {
+                assert_eq!(msg.id(), &tx, "forwarded reply must carry the tx");
+            }
+            other => panic!(
+                "originator never received the terminal reply (got {other:?}): \
+                 the reply was forwarded into the clobbering dummy callback \
+                 whose receiver is gone (try_send err=channel closed)"
+            ),
+        }
+    }
 }
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -6477,4 +6477,164 @@ pub(crate) mod tests {
             ),
         }
     }
+
+    /// A REFUSED newcomer's sender must be DROPPED, not parked.
+    ///
+    /// The guard's contract is not only "the incumbent survives" — it is also
+    /// that the refused caller learns it was refused. Dropping the sender makes
+    /// the newcomer's `recv()` yield `None`, which is what drives the ~360ms
+    /// infra-retry in `op_ctx.rs`. A refactor that stashed skipped senders
+    /// anywhere — a retry queue, a side map, even a stray clone held in a local
+    /// — would silently convert every skip into a full-`OPERATION_TTL` hang,
+    /// and every other test in this module would still pass: they all assert
+    /// about the INCUMBENT, and none observes what became of the newcomer.
+    #[tokio::test]
+    async fn install_pending_op_callback_drops_a_refused_newcomer() {
+        use crate::transport::ExpectedInboundTracker;
+
+        let mut state = super::EventListenerState::new(ExpectedInboundTracker::empty_for_test());
+        let tx = crate::dev_tool::Transaction::new::<crate::operations::put::PutMsg>();
+
+        // A live incumbent holds the slot.
+        let (incumbent_tx_, _incumbent_rx) = mpsc::channel::<crate::node::WaiterReply>(1);
+        state.pending_op_results.insert(tx, incumbent_tx_);
+
+        // The newcomer is refused. Keep its RECEIVER so we can observe the
+        // sender's fate; the function must not retain the sender anywhere.
+        use tokio::sync::mpsc::error::TryRecvError;
+
+        let (newcomer_tx, mut newcomer_rx) = mpsc::channel::<crate::node::WaiterReply>(1);
+        super::dispatch::install_pending_op_callback(&mut state, tx, newcomer_tx);
+
+        assert!(
+            !state
+                .pending_op_results
+                .get(&tx)
+                .expect("incumbent must still hold the slot")
+                .is_closed(),
+            "the live incumbent must survive the refusal"
+        );
+
+        // `None`, not `Empty`: every clone of the sender is gone, so the
+        // newcomer's driver fails fast and retries instead of hanging to TTL.
+        assert!(
+            matches!(newcomer_rx.try_recv(), Err(TryRecvError::Disconnected)),
+            "a refused newcomer's sender must be dropped, not parked — otherwise \
+             its recv() blocks until OPERATION_TTL instead of failing fast"
+        );
+    }
+
+    /// Branch 1 of `install_pending_op_callback`: an incoming callback that is
+    /// ALREADY closed is never inserted.
+    ///
+    /// This is the pre-existing #3100 bound — the driver task can be cancelled
+    /// between `op_execution_sender.send()` and the event loop reaching the
+    /// payload (simulation teardown drops driver futures), and inserting a dead
+    /// sender would hold the slot until the 60s sweep. Guarded end-to-end by
+    /// `test_pending_op_results_bounded`, but only through a full simulation;
+    /// this pins the decision itself.
+    #[tokio::test]
+    async fn install_pending_op_callback_skips_a_closed_incoming_callback() {
+        use crate::transport::ExpectedInboundTracker;
+
+        let mut state = super::EventListenerState::new(ExpectedInboundTracker::empty_for_test());
+        let tx = crate::dev_tool::Transaction::new::<crate::operations::put::PutMsg>();
+
+        let (sender, receiver) = mpsc::channel::<crate::node::WaiterReply>(1);
+        drop(receiver); // the driver was cancelled before we got here
+        assert!(
+            sender.is_closed(),
+            "precondition: the callback must be closed"
+        );
+
+        super::dispatch::install_pending_op_callback(&mut state, tx, sender);
+
+        assert!(
+            !state.pending_op_results.contains_key(&tx),
+            "a closed callback must not occupy a pending_op_results slot: it \
+             would be held until the 60s sweep, which is the #3100 unbounded \
+             growth this skip exists to prevent"
+        );
+    }
+
+    /// Branch 3 of `install_pending_op_callback`, the sub-case the live-waiter
+    /// guard puts at risk: a CLOSED incumbent must still be replaced.
+    ///
+    /// The guard refuses to displace a live waiter only. If it were ever
+    /// loosened to `pending_op_results.get(&tx).is_some()` — dropping the
+    /// `!existing.is_closed()` predicate — dead entries would accumulate in the
+    /// map, reintroducing #3100 by the other door. That mutation passes
+    /// `op_execution_payload_must_not_clobber_live_waiter`, whose incumbent is
+    /// live; this test is what fails it.
+    #[tokio::test]
+    async fn install_pending_op_callback_replaces_a_closed_incumbent() {
+        use crate::transport::ExpectedInboundTracker;
+        use tokio::sync::mpsc::error::TryRecvError;
+
+        let mut state = super::EventListenerState::new(ExpectedInboundTracker::empty_for_test());
+        let tx = crate::dev_tool::Transaction::new::<crate::operations::put::PutMsg>();
+
+        // A stale incumbent whose receiver is already gone.
+        let (stale_sender, stale_receiver) = mpsc::channel::<crate::node::WaiterReply>(1);
+        super::dispatch::install_pending_op_callback(&mut state, tx, stale_sender);
+        drop(stale_receiver);
+        assert!(
+            state.pending_op_results[&tx].is_closed(),
+            "precondition: the incumbent must be closed"
+        );
+
+        // A live callback for the same tx must take the slot.
+        let (fresh_sender, mut fresh_rx) = mpsc::channel::<crate::node::WaiterReply>(1);
+        super::dispatch::install_pending_op_callback(&mut state, tx, fresh_sender);
+
+        assert!(
+            !state.pending_op_results[&tx].is_closed(),
+            "a closed incumbent must be replaced, not preserved: keeping it \
+             strands the slot and strands this waiter"
+        );
+
+        // And the slot must be wired to the NEW waiter, not merely non-closed.
+        let reply = crate::message::NetMessage::V1(crate::message::NetMessageV1::Aborted(tx));
+        use crate::message::MessageStats; // for `msg.id()`
+
+        let cb = state.pending_op_results.get(&tx).cloned();
+        crate::node::try_forward_driver_reply(cb.as_ref(), reply, "put");
+        // Assert the reply ARRIVED, not merely that the map had an entry.
+        // `try_forward_driver_reply` returns true for any Some(callback) — a
+        // try_send that fails Closed or Full is swallowed into a debug! — and
+        // `!matches!(.., Err(Empty))` also passes on Err(Disconnected) and on
+        // Ok(PeerDisconnected). A refactor that dropped `fresh_sender` instead
+        // of installing it would satisfy both and still be wrong.
+        match fresh_rx.try_recv() {
+            Ok(crate::node::WaiterReply::Reply(msg)) => {
+                assert_eq!(msg.id(), &tx, "the replacing waiter got the wrong tx");
+            }
+            other => {
+                panic!("the replacing waiter never received the terminal reply (got {other:?})")
+            }
+        }
+    }
+
+    /// Branch 3 of `install_pending_op_callback`, plain case: no existing
+    /// entry and a live callback inserts.
+    ///
+    /// Asserted as a precondition inside
+    /// `op_execution_payload_must_not_clobber_live_waiter`; stated separately
+    /// so the extracted function's branch coverage is legible on its own.
+    #[tokio::test]
+    async fn install_pending_op_callback_inserts_when_the_slot_is_empty() {
+        use crate::transport::ExpectedInboundTracker;
+
+        let mut state = super::EventListenerState::new(ExpectedInboundTracker::empty_for_test());
+        let tx = crate::dev_tool::Transaction::new::<crate::operations::put::PutMsg>();
+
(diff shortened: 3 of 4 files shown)


magks:
Addressed the rule-review Info (`dispatch.rs:450`) in 161c7ee — with one deviation worth flagging, because taking the note literally would not have been enough.

The note asked for two branches: the closed-new-callback skip, and the plain insert-with-no-existing-entry. I added both, plus a third the note didn't name — and the third is the only one that carries risk.

The guard I added reads:

```rust
.is_some_and(|existing| !existing.is_closed())
```

so a **closed** incumbent must still be replaced. If that predicate were ever loosened to a plain `.is_some()`, dead entries would accumulate in `pending_op_results` and #3100's unbounded growth would return through the door this fix opened.

I ran that mutation to see what would catch it:

| test | guard mutated to `.is_some()` |
|---|---|
| `install_pending_op_callback_skips_a_closed_incoming_callback` | ok |
| `install_pending_op_callback_inserts_when_the_slot_is_empty` | ok |
| `op_execution_payload_must_not_clobber_live_waiter` | ok |
| `install_pending_op_callback_replaces_a_closed_incumbent` | **FAILED** |

Both branches the note asked for survive it, and so does the race test in this PR — its incumbent is live, so it is structurally blind to this. Only the closed-incumbent case fails. The guard was restored after the run.

The empty-slot case was already asserted as a precondition inside `op_execution_payload_must_not_clobber_live_waiter`; I wrote it out separately anyway so the extracted function's branch coverage is legible in one place rather than implied by another test's setup.

`cargo fmt --all -- --check` and `cargo clippy -p freenet --lib -- -D warnings` are both clean. For completeness: `clippy --all-targets` does fail on this crate, but identically with and without this commit (5 + 45 errors), and none of them cite `p2p_protoc.rs`.

iduartgomez (approved):

iduartgomez (commented):
## Comprehensive PR Review: #5183

### Summary
- **PR Title:** fix(put): never let an op-execution payload displace a live reply waiter
- **Type:** fix
- **CI Status:** green — Clippy, Fmt, Simulation (17m), Unit & Integration (12m), NAT Validation, Windows Check, macOS Service Unit, Rule Lint, Conventional Commits, CLA, Snyk all pass; only Windows Service Unit still pending
- **Linked Issues:** Fixes #5182 (references #4111, #3100)
- **Review tier:** Full — the change sits on the node event-loop dispatch path and is concurrency/async in nature
- **Reviewers run:** code-first, testing, skeptical, big-picture, plus an external non-Claude model (`codex review --base origin/main`)

Every `file:line` below was opened and confirmed against `161c7ee86` before inclusion. Findings from the subagent pass that did not survive that check were dropped.

The external model returned **no findings**: *"The callback-install guard preserves existing live waiters while retaining closed-callback cleanup behavior. The changed tests and targeted p2p protocol tests pass."* — it compiled and ran the suite itself, independently corroborating the Unit & Integration result.

---

### Code-First Analysis

**Independent understanding:** `install_pending_op_callback` (`crates/core/src/node/network_bridge/p2p_protoc/dispatch.rs:450-484`) is extracted from `handle_op_execution` and becomes a three-way decision: (1) incoming callback closed → skip (pre-existing #3100 bound); (2) incoming live but incumbent live → skip (new); (3) otherwise insert/replace. `handle_op_execution` now delegates at `dispatch.rs:238`.

**Stated intent:** matches. The PR body's account of the enqueue→drop window — `send_local_loopback` / `send_fire_and_forget` (`crates/core/src/operations/op_ctx.rs:163`, `:192`) bind `_response_receiver` as a *named* binding that only drops when the producer's poll completes, so the event loop on another thread can observe `is_closed() == false` — is accurate to the code.

**Alignment:** matches. No gap between description and implementation.

---

### Testing Assessment

**Coverage level:** adequate for the fix; one weak assertion and one uncovered invariant.

| Unit | ✅ | 4 new `#[tokio::test]`s, `p2p_protoc.rs:5974-6171`; deterministic, no timing, no threads |
| Integration | ⚠️ | No integration-level repro; existing `FORBIDDEN_MARKER` tests all bind `127.0.0.1:0` (`crates/core/tests/error_notification.rs:558-560`), which the PR correctly identifies as why they miss this |
| Simulation | ✅ | `test_pending_op_results_bounded` (#3100 guard) untouched and still valid — see metrics note below |
| E2E | N/A | |

**Regression test:** present and valid — **mutation-verified**. Replacing the new incumbent predicate with `else if false` makes `op_execution_payload_must_not_clobber_live_waiter` fail with its intended message; restoring it passes. The third test's own documented mutation (`.is_some_and(!closed)` → `.is_some()`) does fail `install_pending_op_callback_replaces_a_closed_incumbent` at `p2p_protoc.rs:6131`, so that doc comment's claim is accurate too.

**Metrics ledger is safe:** removes are recorded only under `if ... .remove(&tx).is_some()` (`p2p_protoc.rs:2647`, `:2655`, `:2678`) and the sweep counts actual removals (`p2p_protoc.rs:1574-1579`), while `test_pending_op_results_bounded` uses `inserts.saturating_sub(removes)` (`crates/core/tests/simulation_integration.rs:6473`). Skipping inserts cannot underflow or inflate the leak metric.

---

### Skeptical Findings

**Risk level:** low. I specifically hunted for the fix's own new failure mode — a *legitimate* live waiter being refused — and could not construct a reachable path in today's code:

- `drive_retry_loop` allocates a fresh `attempt_tx` per retry (`crates/core/src/operations/op_ctx.rs:665-670`, `:448`), so retries never collide with an incumbent.
- Sequential same-tx reuse (GET relay `get/op_ctx_task.rs:3567-3577`; CONNECT's `send_fire_and_forget(ObservedAddress)` at `connect/op_ctx_task.rs:825` immediately before `send_to_and_collect_replies` at `:855`) drops the previous receiver synchronously in the producer task *before* the next payload is enqueued, and the single event loop drains FIFO — so the incumbent is always observed closed and branch 3 replaces it correctly.
- `live_tx_tracker` registration (`dispatch.rs:271`) gates on `contains_key`, which is still true when the skip is caused by a live incumbent — unchanged behavior.

| Weak terminal assertion | med | `p2p_protoc.rs:6136-6146` | See below |
| New skip branch is unobservable | med | `dispatch.rs:476-480` | See below |
| `Transaction::NULL` as a shared key | low | `message.rs:43`, `:1135-1138` | Latent, not live |
| Commit `fbac38e42` does not compile | low | git history | See below |

**1. Weak terminal assertion in `install_pending_op_callback_replaces_a_closed_incumbent`** — `p2p_protoc.rs:6136-6146`:

```rust
assert!(
    crate::node::try_forward_driver_reply(cb.as_ref(), reply, "put"),
    "a callback must be registered for the tx"
);
assert!(
    !matches!(fresh_rx.try_recv(), Err(TryRecvError::Empty)),
    "the replacing waiter must receive the terminal reply"
);
```

`try_forward_driver_reply` returns `true` for *any* `Some(callback)` — including one whose `try_send` failed Closed **or** Full; the error is swallowed into a `debug!` and `true` is returned unconditionally (`crates/core/src/node.rs:1055-1077`, verified). So the first assert only restates "the map has an entry", which the assert six lines above already covers. The second passes on `Err(Disconnected)` and on `Ok(WaiterReply::PeerDisconnected{..})`, not just on the reply actually arriving.

Concretely: a refactor that dropped `fresh_sender` instead of installing it would yield `Err(Disconnected)` here and this assertion would still pass. The two `is_closed()` asserts at `:6122-6135` happen to catch that specific bug today, so nothing live is masked — but this is a needless soft spot in a test whose stated purpose is precision, and the PR **already uses the correct stronger form two tests earlier** at `p2p_protoc.rs:6056-6059`. Given `crates/core/src/node.rs:9506` / #5177 just documented a vacuous-concurrency-test incident in this same file family, please match the stronger shape:

```rust
assert!(matches!(fresh_rx.try_recv(), Ok(WaiterReply::Reply(msg)) if msg.id() == &tx));
```

**2. The new skip branch is invisible in production** — `dispatch.rs:476-480` logs at `debug!` with no metric. The PR body itself makes this exact complaint about `try_forward_driver_reply`'s no-waiter branch ("returns `false` with **no log at any level**... the node had to be patched to see it"), and then adds a second branch with nearly the same property. `GlobalTestMetrics::record_pending_op_insert()` sits three lines below at `dispatch.rs:483` and the module already has `record_pending_op_insert` / `record_pending_op_remove` / `record_pending_op_size` (`crates/core/src/config.rs:4182-4199`) — a `record_pending_op_skip()` sibling would let a sim assertion pin `skips == 0` on a healthy run. Without it, if this guard ever *does* refuse a legitimate waiter in the field, the only symptom is a driver failing and retrying, with nothing correlating it back to the guard.

**3. The guard depends on an invariant that is load-bearing but unstated:** "no two live callbacks are ever installed for the same tx on one node." I verified it holds today (see the three bullets above), but nothing pins it. `Transaction::NULL = Ulid(0)` (`message.rs:43`) is returned as `msg.id()` for `NeighborHosting`, `InterestSync`, `ReadyState`, and `SubscribeHint` (`message.rs:1135-1138`) — those four route via `network_bridge.send()` / `NodeEvent::SendNetMessage` rather than `op_execution_sender`, so this is **latent, not live**. But the consequence changed shape: pre-fix a NULL-keyed collision was last-writer-wins (churn); post-fix the first live NULL waiter *sticks* and blocks every subsequent one until the 60s sweep. Cheap insurance: state the invariant in `install_pending_op_callback`'s rustdoc and add `debug_assert!(&tx != Transaction::NULL, ...)`.

**4. Commit `fbac38e42` does not compile.** The test-first commit references `super::dispatch::install_pending_op_callback` three times, but that function does not exist until `f915b17a2` (verified via `git show`). The PR head is fine and the RED→GREEN recipe in the description works, but the repo's test-first workflow (AGENTS.md: "Write a test that reproduces the bug FIRST / Verify it FAILS without the fix") is not actually demonstrated by the history, and `git bisect` will hit a build failure. Squashing the first two commits, or reordering so the extraction lands first, fixes both.

**Investigated and cleared (recorded so it isn't re-litigated):**
- The outbound dispatch at `dispatch.rs:278` fires regardless of whether the install was skipped. This is **pre-existing, documented, intended** behavior — the original comment ("the outbound request below still dispatches because the receiving peer's view of the operation is independent of the local driver's lifecycle") is preserved, and `put/op_ctx_task.rs:1977-1987` documents that fire-and-forget forward + bypass-to-originator *is* the loopback design. A stray-reply-evicts-the-real-reply chain on the capacity-1 waiter channel (`op_ctx.rs:241`, `:354`) would need a live incumbent and a live reply-expecting newcomer simultaneously, which the reachability analysis above rules out.
- No `.claude/rules/bug-prevention-patterns.md` pattern is tripped: the file currently holds only the Windows `Command::spawn`-after-`FreeConsole` pattern and the self-satisfying `include_str!` pin pattern; this PR adds no source-scrape pins.
- No hot-spin from a refused waiter: both CONNECT drain loops terminate on `Ok(None)` (`connect/op_ctx_task.rs:302`, `:962`) rather than `continue`.

---

### Big Picture Assessment

**Goal alignment:** yes. This fixes the invariant violation at the single choke point every producer funnels through, rather than patching a symptom.

**On "why not fix the producers":** the PR's *empirical* argument is weak and should not be leaned on — 9 vs 19 failures per 50 on identical code means the harness has no discriminating power at n=50, which cuts both ways (it also can't show the consumer-side fix helps). The *architectural* argument stands on its own, though: there is more than one producer, a producer-side fix needs a separate race-free ordering proof per site, and only the consumer-side guard is deterministically testable. That is the right call — just don't cite the numbers as evidence.

**Anti-patterns detected:** none. No `#[ignore]`, no weakened assertions, no swallowed errors, no compat shims.

**Removed code concerns:** none. The `is_closed()` skip, its full comment, and both `record_pending_op_insert()` / `record_pending_op_size()` calls moved verbatim into the new function; the comment was expanded, not trimmed. The #3100 bound is preserved and now additionally pinned at unit level.

**Scope assessment:** focused. The extraction of `install_pending_op_callback` is the mechanism that makes the fix unit-testable without a full `P2pConnManager`, follows the existing `fail_awaiting_connection` pattern, and stays inside the two files the fix touches. It satisfies the one-logical-change policy.

**One prior mitigation is now arguably redundant, and this is worth an explicit answer.** `NodeEvent::SendNetMessage` exists *solely* to dodge this same clobber — its rustdoc at `crates/core/src/message.rs:911-926` says routing CONNECT's `ConnectFailed` through `op_execution_sender` "would overwrite the existing `pending_op_results` slot for the same tx, tearing down the multi-reply receiver." That is precisely what `install_pending_op_callback` now prevents generically. Same question for the hand-written originator-loopback branches in `put/op_ctx_task.rs:1977-1987` and `get/op_ctx_task.rs:3586-3623`. Not a merge blocker, but AGENTS.md flags "fix-then-fix-the-fix" accretion as a known meta-pattern, and this PR makes four overlapping mitigations for one bug class without stating the relationship between them. A sentence in the rustdoc (or a follow-up issue) saying whether the guard subsumes them would save the next maintainer that trace.

---

### Documentation

- Code docs: complete. `install_pending_op_callback`'s rustdoc (`dispatch.rs:431-449`) explains the enqueue→drop window well.
- `dispatch.rs:4` still reads `//! Behavior-preserving extraction from p2p_protoc.rs`. Pre-existing, but this PR is the first change in that file to actually alter behavior, so the module doc is now inaccurate. One-line fix.
- `.claude/rules/operations.md` documents `pending_op_results` *removal* semantics (#4313 `TransactionOrphaned` / send-before-drop) but not the insertion side. A short addendum noting the "never displace a live incumbent" invariant would keep that file the canonical map. Optional.
- User/architecture docs: no change needed.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
1. **Strengthen the terminal assertion** at `p2p_protoc.rs:6144-6147` to `matches!(fresh_rx.try_recv(), Ok(WaiterReply::Reply(msg)) if msg.id() == &tx)`, matching the correct form the PR already uses at `:6056-6059`. Optionally drop the redundant `assert!(try_forward_driver_reply(...))` above it, since it cannot fail once the entry exists.
2. **Add a `GlobalTestMetrics::record_pending_op_skip()`** in the new live-incumbent branch (`dispatch.rs:476-480`), so the guard's own failure mode is observable and pinnable in simulation — the PR's own diagnosis of why this bug took field patching to find applies to the branch it adds.
3. **Squash `fbac38e42` into `f915b17a2`** (or reorder the extraction first) so every commit builds and the test-first workflow is actually demonstrated by the history.

#### Consider (Suggestions)
4. Document the "one live waiter per tx per node" invariant in `install_pending_op_callback`'s rustdoc, with a `debug_assert!(&tx != Transaction::NULL, ...)` and a pointer to `RetryDriver::new_attempt_tx` (`op_ctx.rs:448`) as the reason retries can't collide.
5. Add a unit test pinning that a refused newcomer's sender is **dropped** (its `recv()` yields `None`, driving the ~360ms infra-retry at `op_ctx.rs:713-745`) rather than parked. A future refactor that stashed skipped senders anywhere would silently convert every skip into a full-`OPERATION_TTL` hang, and all four current tests would still pass.
6. State whether `NodeEvent::SendNetMessage` (`message.rs:911-926`) and the two hand-written loopback branches are now subsumed; file the follow-up if so.
7. Fix the stale `//! Behavior-preserving extraction` module doc at `dispatch.rs:4`.
8. Take up your own offer of the `debug!` in `try_forward_driver_reply`'s no-waiter branch, and consider noting in the PR body that this touches the shared dispatch path (CONNECT / GET / UPDATE / SUBSCRIBE), not only PUT — the `fix(put):` prefix understates the blast radius even though PUT is where it was observed.
9. Track an integration-level repro as a separate issue. The "all existing guards bind `127.0.0.1`" gap the PR identifies still exists after this PR; only the specific bug is now covered at unit level. Not worth blocking on — making that race deterministic is genuinely hard.

---

### Verdict
**HEAD SHA reviewed:** `161c7ee864652599aee1aea235796f675af5dba1`

The fix is correct, well-reasoned, minimally scoped, and its regression test is mutation-verified to genuinely fail without the fix. CI is green (only Windows Service Unit outstanding). No blocking findings. The three Should Fix items are a test-assertion tightening, an observability counter, and a commit squash — all mechanical, none altering the fix logic, so a diff-of-the-diff pass on the fix commits is sufficient rather than a full re-review. Nice work on the deterministic reproduction; modelling the enqueue→drop window by holding the dummy receiver alive across the install is a clean way to make a thread-timing bug testable without threads.

[AI-assisted - Claude]

iduartgomez (dismissed):

magks:
Two corrections to my own PR, both verified rather than reasoned about.

### 1. The RED → GREEN recipe above does not work

`HEAD` on this branch is `161c7ee8`, which is a **test** commit. Reverting it leaves the
fix (`f915b17a`) in place, so the reproduction stays green:

```
git revert --no-commit HEAD
cargo test -p freenet --lib op_execution_payload_must_not_clobber_live_waiter
→ test result: ok. 1 passed          # the recipe promises RED
```

Reverting `f915b17a` directly does not work either — it removes
`install_pending_op_callback`, which the tests call, so you get `E0425` instead of a red
test.

**What does work.** Keep the extraction and the `is_closed()` guard, and delete only the
incumbent-waiter branch, which leaves upstream's exact semantics:

```rust
// crates/core/src/node/network_bridge/p2p_protoc/dispatch.rs
 pub(super) fn install_pending_op_callback(...) {
     if callback.is_closed() {
         ...
-    } else if state
-        .pending_op_results
-        .get(&tx)
-        .is_some_and(|existing| !existing.is_closed())
-    {
-        // A LIVE waiter is already registered for this tx — never displace it.
-        ...
     } else {
         state.pending_op_results.insert(tx, callback);
     }
 }
```

```
→ test result: FAILED. 0 passed; 1 failed     # RED, 0 compile errors
panicked: originator's waiter channel was closed by the dummy-callback install:
the driver's recv() yields None -> NotificationError -> the client sees
'failed notifying, channel closed' for a PUT that succeeded (the #4111 forbidden marker)
```

Both directions were run on this branch: fix present → `1 passed`; incumbent branch
removed → `1 failed`.

### 2. Still reproduces on v0.2.122

Re-checked today against `48c166c4`. `handle_op_execution` is unchanged — it still guards
only on the incoming `callback.is_closed()` and never asks whether a live waiter is
already registered for that `tx`. Extracting that logic verbatim into a callable helper
and running the same test gives the same failure and the same marker.

So the defect is live four releases on from where it was found, and this fix still
applies.

*Scope of the above: measured on this branch and on a v0.2.122 worktree, not on a merge
of this branch with current `main`.*

[AI-assisted - Claude]

magks:
Rebased onto current `main` (`48c166c4`, v0.2.122) and re-verified there, which
supersedes the scope note on my previous comment — that hedge was weaker than the
evidence warranted.

Same three commits, replayed with no content change (`git range-diff` reports all three
as `=`). `dispatch.rs`, the only file the fix touches, is byte-identical between the old
branch base and current `main` — upstream has not modified it in the 23 commits since.

Measured on the rebased branch:

```
fix present                       test result: ok.     1 passed
incumbent-waiter branch removed   test result: FAILED. 0 passed; 1 failed   (0 compile errors)
cargo test -p freenet --lib       test result: ok.     4766 passed; 0 failed; 28 ignored
```

The RED direction is the one from my previous comment: keep the extraction and the
`is_closed()` guard, delete only the incumbent-waiter branch. That leaves upstream's exact
semantics, so the failure is against real behaviour rather than a strawman — it panics
with the `#4111` forbidden marker.

Net: the defect is present in v0.2.122, the fix applies to current `main` unchanged, and
the full lib suite is clean with it.

[AI-assisted - Claude]

magks:
Thanks — that review was more useful than the PR deserved, particularly the mutation check on my own regression test and the external-model cross-run.

All three Should Fix are in, plus four of the suggestions. Now at `401a9bf9`, two commits, rebased on current `main` (`48c166c4`).

### Should Fix

**1. Terminal assertion** — `p2p_protoc.rs:6605-6614`. You were right that it only restated "the map has an entry": `try_forward_driver_reply` returns `true` for any `Some(callback)`, and `!matches!(.., Err(Empty))` also passes on `Err(Disconnected)` and `Ok(PeerDisconnected)`. Now matches the stronger form the PR already used, and the redundant assert above it is gone:

```rust
match fresh_rx.try_recv() {
    Ok(WaiterReply::Reply(msg)) => assert_eq!(msg.id(), &tx, "the replacing waiter got the wrong tx"),
    other => panic!("the replacing waiter never received the terminal reply (got {other:?})"),
}
```

**2. `record_pending_op_skip()`** — `config.rs` alongside its insert/remove/size siblings (counter, reset, accessor), called at `dispatch.rs:504`. Its doc says the expectation directly: this should stay at zero, and a non-zero count means either a genuine collision or the guard refusing a legitimate waiter — which otherwise surfaces only as a driver retry with nothing pointing back here. That makes `skips == 0` pinnable in simulation.

**3. Squashed.** `fbac38e42` is folded into the fix commit, so every commit in the series builds and `git bisect` is safe.

Worth noting I hit this independently from the other direction before your review landed: the standalone reproduction patch in my own notes wouldn't compile against a stock checkout, for exactly this reason — it calls `install_pending_op_callback`, which the fix creates. Two routes to the same defect. The squash commit message records why the test now ships inside the fix rather than ahead of it.

### Suggestions taken

**4.** Invariant documented in `install_pending_op_callback` (`dispatch.rs:456-480`) with `debug_assert!(tx != *Transaction::NULL, ...)`, pointing at `RetryDriver`'s per-attempt channel as the reason retries can't collide, and noting that the consequence of a NULL-keyed collision changed shape with this guard — last-writer-wins before, first-waiter-sticks after.

**5.** Added `install_pending_op_callback_drops_a_refused_newcomer` (`p2p_protoc.rs:6492`). Your reasoning holds: every existing test asserts about the *incumbent*, so a refactor that parked skipped senders would pass all four while converting each skip into a full-`OPERATION_TTL` hang. It asserts `Err(Disconnected)` on the newcomer's receiver — sender dropped, driver fails fast and retries.

**7.** `dispatch.rs:4-5` no longer claims to be a behavior-preserving extraction.

**8.** Took the `debug!` offer — `node.rs:1092-1097`. The comment says why it matters: finding this bug needed a patched node precisely because "no waiter is registered" is the symptom a displaced waiter produces.

### Deferred

**6** and **9** as follow-ups rather than scope creep here — the `SendNetMessage` / loopback subsumption question, and the integration-level repro. Agreed the second is genuinely hard to make deterministic; the `127.0.0.1` gap is unchanged by this PR.

### Verification

```
install_pending_op_callback tests    4 passed
op_execution_payload_must_not_clobber_live_waiter   1 passed
cargo clippy -p freenet --lib -D warnings   clean
cargo fmt --all -- --check                  clean
cargo test -p freenet --lib          4767 passed; 0 failed; 28 ignored
```

Also re-confirmed the defect on **v0.2.122** before touching any of this: `handle_op_execution` was unchanged, and the reproduction still failed with the `#4111` marker.

One note on `fix(put):` — you're right that it understates the blast radius, since this is the shared dispatch path. I left the title alone to avoid invalidating the review you just did against it; happy to retitle to `fix(node):` if you'd prefer, either here or at squash-merge.

[AI-assisted - Claude]

sanity:
Ran the light re-check on `401a9bf9` that the previous review said would be sufficient. It passes. Independent verification of each item, against the code rather than the summary:

**Guard logic unchanged.** The predicate at `dispatch.rs:483-487` is character-for-character identical to the reviewed version. The only other deltas in that file are the module doc line, the `debug_assert!` block, and the `record_pending_op_skip()` call. That is what keeps this a light re-check rather than a new review.

**Should Fix 1** — `p2p_protoc.rs:6592-6612`. Assertion is now the strong form, and the redundant `assert!(try_forward_driver_reply(..))` is reduced to a bare call.

**Should Fix 2** — `record_pending_op_skip` / `pending_op_skips` at `config.rs:4795,4799`, with the thread_local at `:4460` and the reset at `:4560`, sitting between the remove and high-water-mark siblings. Called at `dispatch.rs:504`.

**Should Fix 3** — `fbac38e4` is gone. Neither remaining commit forward-references a symbol introduced by a later one, so every commit builds and `git bisect` is safe.

**New test is real.** `install_pending_op_callback_drops_a_refused_newcomer` (`p2p_protoc.rs:6489-6531`) holds the newcomer's receiver and asserts `Err(TryRecvError::Disconnected)`. If the function stashed or cloned the sender instead of dropping it, `try_recv` returns `Err(Empty)` and the test fails. It has a red input, which was the point of asking for it.

**`debug_assert!` is latent, not live.** `Transaction::NULL` is produced only at `message.rs:1135-1138` for `NeighborHosting` / `InterestSync` / `ReadyState` / `SubscribeHint`, and all four route through `network_bridge.send()` (`node.rs:1894/2051/2072/2092`), never `op_execution_sender`. The single production caller is `dispatch.rs:239` with `*msg.id()` from an `OpCtx`, and every `OpCtx` send already asserts `msg.id() == &self.tx` against a tx from `Transaction::new::<T>()`. No path reaches it with a NULL tx, and no test constructs one.

**Rebase confirmed.** `998b730` to `48c166c4`. Upstream did not touch `dispatch.rs` at all across those 23 commits, so the fix file had no rebase-conflict exposure. One correction to the summary above: `range-diff` does not report the fix commits as unchanged, since commit 1 is a squash that gained real code. The substance is only the three review fixes, but the literal claim overstates it slightly.

Nothing blocking from this pass. The remaining gate is CI going green on `401a9bf9` (Unit & Integration, Simulation, NAT, Windows and macOS Service Unit still running at the time of writing), and clearing the `CHANGES_REQUESTED` that is still recorded against the superseded `161c7ee8`.

Separately: apologies that the workflow approvals have been the slow part of this on both PRs. That is on us, not on you, and the turnaround on the review feedback here was faster and more complete than we had any right to expect from a first contribution.

[AI-assisted - Claude]

magks:
Filed suggestion 9 as #5253 — the `FORBIDDEN_MARKER` guards all bind `127.0.0.1:0`, so they structurally cannot catch a failure that needs a client remote from the node, which is why this bug reached a deployment before a test.

It includes a concrete route using the `turmoil` runner already in the tree, and is honest that the seed-sweep may not converge, since Turmoil schedules tasks rather than the mid-future point this window lives in. Even a comment on the three bind sites would beat the status quo.

Suggestion 6 (whether `NodeEvent::SendNetMessage` and the loopback branches are now subsumed) I'm still checking — its doc comment says it exists precisely to avoid overwriting `pending_op_results` for a live tx, so the guard in this PR may well make it redundant. I'd rather verify that than assert it.

[AI-assisted - Claude]

PR #5188, opened 2026-08-05
fix(service): reap same-version init-adopted orphans

### LINKED ISSUE #5187: systemd stale-orphan pre-flight (#3967) spares same-version orphans, permanently wedging the unit into unsupervised mode
## Symptom

On a Linux host running the user unit, `freenet.service` was **inactive (dead)** while a healthy `freenet network` process (PPID=1, current binary version) kept running unsupervised. Every start attempt logged:

```
systemd[1570]: Started freenet.service - Freenet Node.
freenet[60363]: Another freenet instance is already running. Exiting without error to avoid restart loop.
```

and the unit went back to dead — but "successfully" (exit 43 is in `SuccessExitStatus`), so nothing looked failed. The node ran for ~12h with no `Restart=` supervision, no `systemctl stop/restart` control, and no auto-update ExecStopPost path.

## Cause

The #3967 `ExecStartPre` pre-flight only kills an init-adopted (PPID==1) orphan when its `Freenet version:` line **differs** from the on-disk binary (or can't be read). A **same-version** orphan is deliberately spared ("as is a current-version orphan"). But any PPID==1 holder at ExecStartPre time wedges the unit exactly the same way regardless of version: ExecStart exits 43, `RestartPreventExitStatus=43` stands the unit down, and the orphan runs unsupervised forever. Version parity doesn't make the orphan legitimate — when the operator starts the unit, a PPID==1 holder is always wrong.

(How the orphan arose: the node outlived a dead user session and was re-parented to init — the same shape #3967 fixed for the old-binary case.)

## Secondary issue

journald logs `Ignoring unknown escape sequences` for the unit line on every daemon-reload, caused by the `\$$2` inside the embedded awk. Cosmetic, but it makes the unit look broken.

## Proposed fix

Kill **any** init-adopted orphan in the pre-flight, version notwithstanding — this is a strict superset of the current version-mismatch behavior, and the version-compare machinery (and its awk backslash) can be deleted. A user-run instance parented by a live shell (PPID!=1) stays protected. PR incoming.

Verified on the affected host: an equivalent unconditional PPID==1 reaper installed as ExecStartPre cleanly reaped the same-version orphan and the unit has run supervised since.

skandragon:
Fixes #5187.

The #3967 `ExecStartPre` pre-flight only killed an init-adopted (PPID==1) orphan whose `Freenet version:` differed from the on-disk binary, deliberately sparing same-version orphans. But a same-version orphan wedges the unit identically: ExecStart exits 43, `RestartPreventExitStatus=43` stands the unit down (reported as success), and the node runs unsupervised forever — no `Restart=`, no `systemctl` control, no ExecStopPost auto-update path. Hit in production; repro logs in #5187.

Changes (both Linux unit templates, user + system):

- The pre-flight now kills **any** PPID==1 `freenet network` orphan, version notwithstanding — a strict superset of the old behavior, so the #3967 old-binary case is still covered. A user-run instance parented by a live shell (PPID!=1) is still left alone and wins the port via exit 43 as designed.
- The version-compare machinery (`ondisk`/`exe`/`hv`/`mismatch`) is deleted with its gate. Dropping the embedded awk (`\$$2` → `cut -d" " -f2`) also silences journald's "Ignoring unknown escape sequences" warning on every daemon-reload.
- Tests updated: the $$-doubling needles track the surviving identifiers, the PPID==1 gate assertion matches the new form, and a new assertion pins the *absence* of a version gate so the spare-same-version behavior can't come back.

The macOS wrapper's exit-43 self-heal has the same version-gated shape and is deliberately out of scope here.

Verified: `sh -n` on the post-$$-collapse body (existing render test emulates this; also run by hand), and an equivalent unconditional PPID==1 reaper deployed as ExecStartPre on the affected host cleanly reaped a same-version orphan, with the unit running supervised since.

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -1482,16 +1482,15 @@ mod tests {
         // Verify exit code 43 prevents restart (another instance already running)
         assert!(service_content.contains("RestartPreventExitStatus=43"));
 
-        // Regression for #3967: stale-orphan self-heal pre-flight. Because
+        // Regression for #3967: orphan self-heal pre-flight. Because
         // RestartPreventExitStatus=43 blocks restart on exit 43, the
-        // version-compare-and-kill must run as an ExecStartPre before the main
-        // ExecStart so a stale orphan can't hold the port forever.
+        // orphan-reap must run as an ExecStartPre before the main
+        // ExecStart so an init-adopted orphan can't hold the port forever.
         assert!(
             service_content.contains("ExecStartPre=")
-                && service_content.contains("Freenet version:")
                 && service_content.contains("kill -TERM")
                 && service_content.contains("kill -KILL"),
-            "systemd unit must run a stale-orphan version-compare-and-kill pre-flight (#3967)"
+            "systemd unit must run an orphan-reap pre-flight (#3967)"
         );
 
         // Verify graceful shutdown timeout is set. 45s = 30s drain
@@ -1583,13 +1582,12 @@ mod tests {
         // Verify exit code 43 prevents restart (another instance already running)
         assert!(service_content.contains("RestartPreventExitStatus=43"));
 
-        // Regression for #3967: stale-orphan self-heal pre-flight (system unit).
+        // Regression for #3967: orphan self-heal pre-flight (system unit).
         assert!(
             service_content.contains("ExecStartPre=")
-                && service_content.contains("Freenet version:")
                 && service_content.contains("kill -TERM")
                 && service_content.contains("kill -KILL"),
-            "system systemd unit must run a stale-orphan version-compare-and-kill pre-flight (#3967)"
+            "system systemd unit must run an orphan-reap pre-flight (#3967)"
         );
 
         // Logging routes to journal (same reasoning as the user-unit test).
@@ -1902,7 +1900,7 @@ mod tests {
 
             // Post-render the file on disk MUST carry the systemd-level `$$`
             // form for the shell identifiers the pre-flight relies on.
-            for needle in ["$$pid", "$$exe", "$$ondisk", "$$self", "$$(id -u)"] {
+            for needle in ["$$pid", "$$ppid", "$$self", "$$(id -u)"] {
                 assert!(
                     pre.contains(needle),
                     "{which} ExecStartPre must double shell dollars (missing `{needle}`); \
@@ -1917,13 +1915,7 @@ mod tests {
             // NB: we cannot blanket-assert absence of `$(id -u)` because the
             // VALID doubled form `$$(id -u)` contains it as a substring; the
             // `$$(id -u)` presence check above covers that identifier instead.
-            for bad in [
-                "/proc/$pid/",
-                "\"$pid\"",
-                "\"$exe\"",
-                "\"$ondisk\"",
-                "\"$self\"",
-            ] {
+            for bad in ["/proc/$pid/", "\"$pid\"", "\"$ppid\"", "\"$self\""] {
                 assert!(
                     !pre.contains(bad),
                     "{which} ExecStartPre contains bare single-`$` `{bad}` that systemd \
@@ -1967,11 +1959,18 @@ mod tests {
                  (#3967 round-2 RE-REVIEW)"
             );
 
-            // MUST-FIX #3/#4: kill is gated on PPID==1 AND (readable-version
-            // mismatch OR unreadable holder version). Render-level check.
+            // MUST-FIX #3: kill is gated on PPID==1 (init-adopted orphan).
+            // The former version-compare gate is deliberately GONE: sparing a
+            // current-version orphan wedges the unit into success-but-dead
+            // with the node unsupervised. Render-level check.
             assert!(
-                pre.contains("[ \"$$ppid\" = \"1\" ] &&"),
+                pre.contains("if [ \"$$ppid\" = \"1\" ]; then"),
                 "{which} ExecStartPre must require PPID==1 before killing (#3967 MUST-FIX 3)"
+            );
+            assert!(
+                !pre.contains("$$ondisk") && !pre.contains("$$hv"),
+                "{which} ExecStartPre must not version-gate the orphan kill — a \
+                 same-version orphan wedges the unit into unsupervised mode"
             );
 
             // Now emulate systemd's `$$` -> `$` collapse and confirm the body
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -2231,30 +2231,31 @@ echo "RC=$?"
         );
     }
 
-    /// Behavioral regression for the SYSTEMD inline self-heal DECISION (#3967,
-    /// round-2 RE-REVIEW). The render + `sh -n` test proves the inline body
-    /// PARSES; this proves it DECIDES correctly. Crucially it locks the round-2
-    /// fix: the round-1 unit anchored the holder loop on
-    /// `[ "$exe" != "$bin" ] && continue`, which skipped every holder whose exe
-    /// differed from the unit's on-disk binary — i.e. exactly the stale orphan
-    /// (running an OLD binary) the self-heal exists to kill. Here the fake
-    /// holder's exe path is DIFFERENT from the on-disk binary, and we assert it
-    /// IS killed.
+    /// Shared harness for the systemd ExecStartPre self-heal DECISION tests
+    /// (#3967). The render + `sh -n` test proves the inline body PARSES; the
+    /// tests built on this harness prove it DECIDES correctly.
     ///
-    /// We extract the `ExecStartPre` `/bin/sh -c '...'` body from the rendered
-    /// unit, collapse systemd's `$$` -> `$`, and run it under `sh` with
-    /// PATH-shimmed `pgrep`/`readlink`/`timeout` and a `kill` shell function
-    /// (functions beat the builtin) that records targets to a marker.
+    /// Extracts the `ExecStartPre` `/bin/sh -c '...'` body from the rendered
+    /// user unit, collapses systemd's `$$` -> `$`, and returns the on-disk
+    /// binary path plus a `run(holder_version, holder_ppid, pgrep_pid)`
+    /// closure that executes the body under `sh` with PATH-shimmed
+    /// `pgrep`/`readlink`/`timeout` and a `kill` shell function (functions
+    /// beat the builtin) recording kills to a marker; it returns whether the
+    /// holder was killed. The version machinery is deliberately kept even
+    /// though the current pre-flight is version-blind: if a version gate is
+    /// ever reintroduced, the same-version regression test below fails.
     ///
     /// Linux-only because it exercises the `/proc`-based systemd inline path
     /// and the unit generators are `#[cfg(target_os = "linux")]`.
-    #[test]
     #[cfg(target_os = "linux")]
-    fn test_systemd_execstartpre_heal_decision_behavior() {
+    fn systemd_heal_harness(
+        dir: &std::path::Path,
+    ) -> (
+        std::path::PathBuf,
+        impl Fn(&str, &str, Option<&str>) -> bool,
+    ) {
         use std::os::unix::fs::PermissionsExt;
 
-        let tmp = tempfile::tempdir().unwrap();
-        let dir = tmp.path();
         let chmod_x = |p: &std::path::Path| {
             std::fs::set_permissions(p, std::fs::Permissions::from_mode(0o755)).unwrap();
         };
@@ -2332,11 +2333,12 @@ echo "RC=$?"
         // shim's */proc/*/exe arm fires.
 
         let killed_marker = dir.join("killed");
+        let harness_path = dir.join("harness.sh");
         let path_env = format!("{}:{}", bin.display(), std::env::var("PATH").unwrap());
 
         // Returns (holder_was_killed). HOLDER_PPID drives the fake stat file's
         // PPID field; HOLDER_VER drives the holder binary's version.
-        let run = |holder_ver: &str, holder_ppid: &str, pgrep_pid: Option<&str>| -> bool {
+        let run = move |holder_ver: &str, holder_ppid: &str, pgrep_pid: Option<&str>| -> bool {
             std::fs::remove_file(&killed_marker).ok();
             std::fs::write(
                 &stat_file,
@@ -2351,7 +2353,6 @@ echo "RC=$?"
                 marker = killed_marker.display(),
                 body = body,
             );
-            let harness_path = dir.join("harness.sh");
             std::fs::write(&harness_path, harness).unwrap();
             let mut cmd = std::process::Command::new("sh");
             cmd.arg(&harness_path)
@@ -2363,6 +2364,19 @@ echo "RC=$?"
             cmd.output().expect("failed to run systemd heal harness");
             killed_marker.exists()
         };
+        (ondisk_bin, run)
+    }
+
+    /// Behavioral regression for the systemd self-heal decision (#3967,
+    /// round-2 RE-REVIEW): an init-adopted orphan (PPID==1) is killed, a
+    /// user-parented holder (PPID!=1) is deferred to.
+    #[test]
+    #[cfg(target_os = "linux")]
+    fn test_systemd_execstartpre_heal_decision_behavior() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let tmp = tempfile::tempdir().unwrap();
+        let (ondisk_bin, run) = systemd_heal_harness(tmp.path());
 
         // (a) ROUND-2 CORE: orphan (PPID==1) running an OLD version at a
         // DIFFERENT exe path than the on-disk binary -> MUST be KILLED. The
@@ -2374,11 +2388,8 @@ echo "RC=$?"
              exe==bin anchor wrongly skipped it)"
         );
 
-        // (b) orphan (PPID==1) running the SAME current version -> DEFER.
-        assert!(
-            !run("1.0.0 (current)", "1", None),
-            "systemd self-heal must DEFER to a same-version holder (polite path)"
-        );
+        // (b) same-version PPID==1 orphan -> KILLED; covered by the dedicated
+        // regression test test_systemd_execstartpre_kills_same_version_orphan.
 
         // (c) version-mismatched but USER-parented (PPID!=1) -> DEFER.
         assert!(
@@ -2386,21 +2397,33 @@ echo "RC=$?"
             "systemd self-heal must DEFER to a user-parented (PPID!=1) holder (#3967 MUST-FIX 3)"
         );
 
-        // (d) on-disk binary version UNREADABLE (`$ondisk` empty) -> DEFER even for
-        // a PPID==1 orphan whose own version IS readable. With no trustworthy
-        // baseline we must not version-compare; killing here would reap a healthy
-        // current node during an update window when the binary is momentarily
-        // unreadable. Mirrors the macOS test's chmod-644 case (#3967 MUST-FIX 4),
-        // which the systemd path's `[ -n "$$ondisk" ]` guard implements but had no
-        // behavioral coverage on Linux (rule-review Info on #4408).
+        // (d) on-disk binary UNREADABLE -> still KILLED. The pre-flight no
+        // longer reads any version (the version gate was exactly the #3967
+        // same-version wedge), so an unreadable on-disk binary cannot spare a
+        // PPID==1 orphan.
         std::fs::set_permissions(&ondisk_bin, std::fs::Permissions::from_mode(0o644)).unwrap();
         assert!(
-            !run("0.9.0 (old)", "1", None),
-            "systemd self-heal must DEFER when the on-disk version is unreadable, even \
-             for a readable-version PPID==1 orphan (#3967 MUST-FIX 4: empty $ondisk \
-             must not drive a kill)"
+            run("0.9.0 (old)", "1", None),
+            "systemd self-heal must KILL a PPID==1 orphan even when the on-disk \
+             version is unreadable — the pre-flight is version-blind"
         );
-        chmod_x(&ondisk_bin); // restore for any later use
+    }
+
+    /// Regression for the same-version orphan wedge (#3967 follow-up): a
+    /// SAME-VERSION init-adopted orphan (PPID==1) MUST be killed. The earlier
+    /// pre-flight version-compared and deferred to a same-version holder,
+    /// which wedged the unit into success-but-dead (ExecStart exits 43, no
+    /// restart) with the node running unsupervised forever.
+    #[test]
+    #[cfg(target_os = "linux")]
+    fn test_systemd_execstartpre_kills_same_version_orphan() {
+        let tmp = tempfile::tempdir().unwrap();
+        let (_ondisk_bin, run) = systemd_heal_harness(tmp.path());
+        assert!(
+            run("1.0.0 (current)", "1", None),
+            "systemd self-heal must KILL a same-version PPID==1 orphan — sparing \
+             it wedges the unit into success-but-dead / unsupervised mode (#3967)"
+        );
     }
 
     #[test]
--- a/crates/core/src/bin/commands/service/linux.rs
+++ b/crates/core/src/bin/commands/service/linux.rs
@@ -450,31 +450,31 @@ Environment=FREENET_SUPERVISED=1
 # (auto-updated but not reinstalled) or the mac/Windows wrapper keeps emitting the
 # self-healing exit 42 instead of a 45 the supervisor would mishandle.
 Environment={fast_crash_marker}=1
-# Stale-orphan self-heal (issue #3967): RestartPreventExitStatus=43 below
-# means an exit 43 ("another instance already running") never restarts the
-# unit. That is correct for a legitimate second instance, but if the port
-# holder is an ORPHANED `freenet network` (PPID=1) still running an OLD
-# binary, the unit would stand down and the orphan would serve stale assets
-# forever. This pre-flight runs before every start: it finds the port
-# holder, and kills it ONLY when it is an init-adopted orphan (PPID==1) whose
-# `Freenet version:` line differs from the binary this unit would launch (or
-# whose version can't be read). A user-run `freenet network` (parented by a
-# shell, PPID!=1) is always left alone, as is a current-version orphan.
+# Orphan self-heal (issue #3967): RestartPreventExitStatus=43 below means an
+# exit 43 ("another instance already running") never restarts the unit. That is
+# correct for a legitimate second instance, but an init-adopted ORPHANED
+# `freenet network` (PPID==1) holding the port wedges the unit permanently:
+# ExecStart exits 43, the unit reports success-but-dead, and the orphan runs
+# UNSUPERVISED forever — no Restart= on crash, no systemctl control, and (if it
+# is an old binary) stale assets served indefinitely. This pre-flight runs
+# before every start and kills ANY init-adopted orphan, regardless of version:
+# an earlier revision compared `Freenet version:` lines and spared a
+# current-version orphan, but that just wedges the unit into the unsupervised
+# state — when the operator starts this unit, a PPID==1 holder is always wrong.
+# A user-run `freenet network` (parented by a live shell, PPID!=1) is left
+# alone; it wins the port and the unit stands down via exit 43 as designed.
 #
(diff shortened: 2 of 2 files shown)


iduartgomez (commented):
## Comprehensive PR Review: #5188

### Summary
- **PR Title:** fix(service): reap same-version init-adopted orphans
- **Type:** fix
- **CI Status:** all green (15 checks, incl. Rule Review with 0 warnings)
- **Linked Issues:** Fixes #5187; follow-up to #3967 / #4408
- **Review tier:** Full (deploy/service-unit surface + unconditional process kill)
- **Reviewers run:** code-first, testing, skeptical, big-picture, plus an external non-Claude model (`codex review --base origin/main`). Every finding below was re-verified against the checked-out code before inclusion; several subagent findings were dropped or downgraded as false positives (noted inline).

The core diagnosis in #5187 is correct and well-argued, and the write-up is unusually accurate about its own change. The concern below is that the chosen discriminator — `PPID == 1` alone — is sound for the **user** unit and unsound for the **system** unit, where systemd *is* PID 1.

---

### Code-First Analysis

**Independent understanding:** Both systemd templates (`service/linux.rs:477` user, `:633` system) replace the version-compare orphan reaper with a version-blind one. New predicate is `PPID == 1` and nothing else; `readlink /proc/PID/exe`, both `timeout 5 … --version` probes, and the `$ondisk`/`$hv`/`mismatch` machinery are gone. PPID extraction moves from `awk "{{print \$$2}}"` to `cut -d" " -f2`.

**Stated intent:** matches. No false claims found in the description.

**Alignment:** matches. No scope creep — the diff touches only the two templates and their tests.

**Gaps (both minor, both worth a line in the PR body):**
- The pre-flight no longer execs any binary, removing up to two 5s `timeout` calls from every service start. That is a real start-path latency and failure-mode improvement, unmentioned.
- The set of newly-killed processes is *not* limited to same-version orphans. A non-freenet `pgrep -f` match with PPID==1 was already killed before (the old `[ -z "$hv" ]` disjunct covered it), so blast radius against non-freenet processes is genuinely unchanged — but a **readable, same-version, deliberately-detached** node is a new kill class. See Should Fix #3/#4.

---

### Testing Assessment

**Coverage level:** adequate for the fix itself; gaps around the system unit and the secondary journald fix.

| Unit (render pins) | ✅ | `service.rs:1897-1997`; anchored via `.lines().find(…)` + panic, correctly bounded per the AGENTS.md source-scrape-pin rule — these scrape a rendered `String`, not `include_str!`, and a moved anchor fails loudly |
| Unit (behavioral, user unit) | ✅ | `systemd_heal_harness` + two tests |
| Unit (behavioral, system unit) | ❌ | harness hardcodes `generate_user_service_file` (`service.rs:2288`); system template has render-level coverage only |
| Integration / Simulation / E2E | N/A | template generation only |

**Regression test: present and valid.** The testing pass reproduced the harness exactly (same shims, same `$$`→`$` collapse, same fake `/proc/PID/stat`) and drove both the `origin/main` and PR-head templates through it:

```
                                  pre-fix   post-fix
(a) old-version PPID=1 orphan      killed    killed
(b) SAME-version PPID=1 orphan     spared    killed
(c) old-version PPID=4321          spared    spared
(d) ondisk unreadable, orphan      spared    killed
```

`test_systemd_execstartpre_kills_same_version_orphan` genuinely fails against the pre-fix template. **This resolves a disagreement between reviewers:** one pass called the new test tautological. Empirically it is not — it is a valid guard against reverting the fix. What *is* true is the weaker claim that, given the current body, cases (a), (b) and (d) all assert the single fact "PPID==1 ⇒ killed", so (d) in particular is now inert (the `chmod 644` at `service.rs:2405` cannot influence a script that never reads the binary).

**Missing tests:**
- The journald/backslash half of #5187 is entirely unpinned (Should Fix #5).
- System unit has no behavioral coverage, and nothing pins that the two `ExecStartPre` bodies stay identical.
- The self-PID skip (MUST-FIX #5) has render-level coverage only. This matters *more* after this PR under a system unit, where the pre-flight `sh` itself has PPID==1 and its argv contains the literal `freenet network`. The harness already carries an unused `pgrep_pid: Option<&str>` hook (`service.rs:2341`) that no Linux caller ever passes `Some` to; wiring it is ~5 lines.
- `$$w` is absent from the `$$`-doubling needle list (`service.rs:1903`) even though the body uses `[ $$w -lt 12 ]` and `w=$$((w+1))`. A regression to bare `$w` survives the `sh -n` check (it still parses) and would silently break the TERM→KILL escalation. Pre-existing, not introduced here.

---

### Skeptical Findings

**Risk level:** medium-high — the failure mode is "SIGKILL a healthy production node", but it requires a specific install state.

| # | Concern | Severity | Location (verified) | Details |
|---|---------|----------|---------------------|---------|
| 1 | `PPID==1` is the *healthy* signature for a system unit | **High** | `linux.rs:633`, `:610-612`, `:63` | See below |
| 2 | 12s TERM→KILL is shorter than the node's own 30s drain | Medium | `linux.rs:477`/`:633`, `config.rs:1663`, `linux.rs:516` | `TimeoutStopSec=45`; drain default 30; reaper escalates at 12 |
| 3 | SIGKILL is classified as a crash | Medium | `linux.rs:539`, `rollback.rs:716` | `ExecStopPost` forwards any `$EXIT_STATUS` not in `0\|43`; `classify_stop` maps `KILL` via `_ => StopClass::Crash` → counts toward crash-loop probation / auto-rollback |
| 4 | No settle wait after `kill -KILL` before ExecStart | Medium | `linux.rs:477`/`:633` | Loop ends immediately after SIGKILL; ExecStart binds within ms |
| 5 | No identity check on the victim at all | Medium | `linux.rs:477`/`:633` | `pgrep -f` unanchored substring match, `--version` probe removed |
| 6 | Documented "user-run instance is left alone" now false | Medium | `linux.rs:463-465`, `:632` | `nohup`/`screen -dmS`/`disown` + logout all reparent to PID 1 |
| 7 | Unbounded serial reap loop in a start-gating script | Low | `linux.rs:477`/`:633` | ≥8 victims × 12s exceeds the default `TimeoutStartSec=90`; the `-` prefix suppresses exit status, not a start-job timeout |
| 8 | PID-reuse TOCTOU between TERM and the final KILL | Low | `linux.rs:477`/`:633` | Pre-existing; frequency rises with the unconditional reap |

**On #1, the one I'd call blocking.** For a `Type=simple` **system** unit (`linux.rs:610-611`), PID 1 forks the service directly, so a correctly-supervised node's PPID is 1 by construction. For a **user** unit the parent is `systemd --user`, so PPID != 1 and the heuristic is sound. The templates and their comments do not distinguish the two.

This is only reachable when both units exist for one UID — but that state is explicitly contemplated by the code:

- `use_system_mode()` (`linux.rs:63`) is `system_flag || (has_system_service() && !has_user_service())` — written specifically to resolve the both-installed case, and it resolves it to *user* mode.
- The system unit's `User={username}` (`linux.rs:612`) is derived from `SUDO_USER`/`USER` (`linux.rs:331`) and root is explicitly refused (`linux.rs:341`), so the system unit runs as **the same UID** that owns the user unit.
- Neither `install_user_service` (`linux.rs:228`) nor `install_system_service` (`linux.rs:324`) checks for or removes the other, and both `systemctl enable`.

So: user unit starts → its `ExecStartPre` runs as that UID → `pgrep -f -u $(id -u) "freenet network"` finds the system unit's healthy main process → PPID is 1 → SIGTERM, then SIGKILL 12s later (mid-drain, per #2) → the system unit sees `KILL`, which is a `Crash` for rollback purposes (#3) and outside `SuccessExitStatus=42 43`. The previous version gate spared it in the overwhelmingly common same-version case. It does settle rather than loop (the system unit's own pre-flight spares the user node, PPID != 1) — but it settles by killing a healthy supervised node and wedging its unit into exactly the `success-but-dead` state this PR exists to eliminate.

Note #2/#3 are consequences that only bite on this path: a genuine unmanaged orphan has no `ExecStopPost`, so no spurious rollback. Scoped to #1, but they make its blast radius worse than "one restart".

**On #5, which the external model raised independently as its single P1.** With the `--version` probe gone, nothing verifies the victim is even a freenet binary. Concretely, same-UID + PPID==1 collateral now includes `screen -dmS fn freenet network` (the *screen* daemon's argv contains the pattern), a disowned wrapper `sh -c 'while :; do freenet network; done'`, and — the external model's framing — a second deliberately-detached node on a different `--ws-api-port`/`--config-dir`, which the reaper kills without ever checking whether it holds *this* unit's port. Anchoring on argv[0] basename from `/proc/$pid/cmdline` (first NUL-delimited token is `freenet` or ends in `/freenet`) would cost nothing and is **not** the `exe == on-disk binary` anchor the tests rightly forbid at `service.rs:1954` — that anchor fails because a #3967 orphan runs a *different* binary; a basename check does not.

**Verified and found safe (dropped from the findings above):**
- `cut -d" " -f2` ≡ `awk '{print $2}'` after `sed "s/.*) //"`. Tested across comm values containing spaces, `") "`, `"x)"`, exactly `")"`, and a crafted fake-stat prefix — all identical. `sed`'s greedy `.*) ` always lands on the real closing paren (no later stat field contains `)`), and the remainder is single-space separated, so `cut`'s non-collapsing behavior is irrelevant. The only divergence (`cut` passes a delimiter-less line through whole) is unreachable for a real stat file. **No bug.**
- `$$` escaping in the shortened line, and `sh -n` on the collapsed body: clean.
- Self-exclusion and the PID-1 skip: no self-kill, no systemd-kill. The `$( )` subshell running `pgrep` has PPID = the sh, not 1.
- **Single-unit lifecycle paths are all safe**, which is worth stating explicitly: `systemctl start` on an active unit is a no-op (no ExecStartPre re-run); `restart`/`try-restart` stop first, and default `KillMode=control-group` + `TimeoutStopSec=45` guarantee the cgroup is dead before ExecStartPre; the `RestartSec` window has no live main process; there is no socket activation; `ExecStopPost` runs `{binary} update`, whose cmdline does not match `freenet network`. The hazard is specifically multi-unit.
- Bad-needle pins have no false positives: `"$$ppid"` does not contain `"$ppid"`, and `/proc/$$pid/` does not contain `/proc/$pid/`.
- Graceful TERM is not a rollback hazard — `freenet.rs` SIGTERM handler exits 0, and `classify_stop("0") = NotCrash`. Only the SIGKILL escalation is.

---

### Big Picture Assessment

**Goal alignment:** yes, both the primary fix and the secondary journald ask.

**Anti-patterns detected:** none of the classic ones. No `#[ignore]`, no commented-out tests, no swallowed errors, no weakened tolerances. The test refactor preserves rather than launders coverage — deleted case (b) comes back inverted as a dedicated, better-named test.

**Removed code concerns:** the deletion of MUST-FIX #4 ("defer when the on-disk version is unreadable") is **not** a re-introduced bug, and I want to be explicit about why, since it reads like one. That guard existed solely to stop an empty `$ondisk` baseline from making every holder look mismatched and thereby *driving* a kill. With the version comparison gone entirely, it has no subject left to protect — there is no baseline to be untrustworthy about. MUST-FIX #3 (PPID gate) and #6 (stat parse) both survive intact and remain pinned. The removal is coherent.

**Scope assessment:** focused.

**macOS divergence.** `service/macos.rs:272` still carries the byte-identical version gate (`[ "$ppid" = "1" ] && { [ "$version_mismatch" = "0" ] || [ -z "$holder_ver" ]; }`), and the defer path exits 0 under `KeepAlive.SuccessfulExit=false` — the same wedge shape. Deferring it is defensible, but for a reason the PR body does not give: `macos.rs:298` begins every wrapper launch with an *unconditional* `pkill -f -u "$(id -u)" "freenet network"`, so a `launchctl kickstart` heals a wedged macOS install, whereas `systemctl start` was permanently ineffective pre-fix. That asymmetry is the actual justification and belongs in the PR body. After this PR the macOS exit-43 path is the only remaining version-gated killer of four code paths (`wrapper.rs:563` and `macos.rs:298` are both unconditional). The two test suites now assert **opposite policies for identical inputs** (`service.rs:2171` asserts macOS DEFERS on same-version; `service.rs:2419` asserts Linux KILLS), and nothing will fail CI. `gh issue list` finds no tracking issue for the macOS half — only umbrella #3967 (open) and closed #4382/#4410.

**Delivery gap worth noting for operators:** a host already in the #5187 wedge has a dead unit, so no `ExecStopPost` runs and no `freenet update` fires. It keeps its old version-gated unit until someone manually runs `freenet update` or `freenet service install`. The fix is preventative, not self-delivering to the population it was written for. (Separately, `update.rs:806` and `:2565` hardcode `systemctl --user`, so a system-mode install is never detected or restarted by `freenet update` — pre-existing, but it compounds the dual-install state in Must Fix #1.)

---

### Documentation
- **Code docs:** incomplete. `linux.rs:463-465` and `:632` still claim a user-run instance is protected by shell parentage — false after this change for `nohup`/`screen`/`disown`. Neither comment mentions that a system-unit-supervised node is also PPID==1.
- **Stale comments in the test harness:** `service.rs:2330-2331` still says "leave [the readlink call] so the shim's `*/proc/*/exe` arm fires" and `:2323`/`:2328` still describe `awk` — the body no longer calls `readlink`, `timeout`, `grep`, or `awk`. `service.rs:2378`'s case-(a) rationale ("OLD version at a DIFFERENT exe path", citing the round-1 `exe==bin` anchor) is now vacuous.
- **Architecture / rules docs:** up to date. Grepped `.claude/rules/`, `AGENTS.md`, `CONTRIBUTING.md`, and `docs/` for `orphan`, `3967`, `ExecStartPre`, `Freenet version:` — nothing describes the version-gated behavior. No agents/skills/hooks affected. `update.rs:1925` decides unit staleness by byte-for-byte template diff (not marker greps, per #4287), so existing installs pick the new unit up without test churn.
- **PR description:** content is all there but lacks the `## Problem / ## Solution / ## Testing / ## Fixes` headings required by `.claude/rules/git-workflow.md`.

---

### Recommendations

#### Must Fix (Blocking)
1. **Handle the system-unit case.** `PPID==1` is the healthy-state signature for `Type=simple` under the system manager, and `use_system_mode()` (`linux.rs:63`) is written for the both-units-installed state that makes this reachable with the same UID. Options, in rough preference order:
   - Skip any pid whose cgroup is a live systemd unit — `grep -q '/system.slice/' /proc/$pid/cgroup && continue` distinguishes "supervised by systemd" from "adopted by init" far more precisely than PPID, and a genuine #3967 orphan (leftover/root cgroup) is still reaped. This is the cheapest fix and needs no install-time changes.
   - Or refuse to reap a pid that is the `MainPID` of an active freenet unit (`systemctl show -p MainPID --value`, both system and `--user`).
   - Or scope the version-blind reap to the user template only, and make the installers refuse/clean up the other flavor.

   Worth answering in the PR thread regardless: **was the wedge observed on a user unit or a system unit?** If it was a user unit, scoping the change to that template is a small, safe patch.

#### Should Fix (Important)
2. **Raise the TERM→KILL escalation to ≥30s** (drain default, `config.rs:1663`) or ideally 45 (`TimeoutStopSec`, `linux.rs:516`). 12s was defensible when only stale/unresponsive orphans were targets; once a healthy node can be a target, the normal outcome is SIGKILL mid-drain with in-flight PUT/GET/UPDATE aborted — and `KILL` reaches `classify_stop` as a `Crash` (`rollback.rs:716`), counting toward crash-loop probation and potentially triggering a spurious rollback to an older binary.
3. **Poll after `kill -KILL` before falling through to ExecStart.** The SIGKILL branch is by definition only reached for a process that ignored SIGTERM for 12s — exactly the process whose socket teardown is not instantaneous. ExecStart then fails to bind → exit 43 → `RestartPreventExitStatus=43` → the unit stands down, which is the wedge this PR fixes.
4. **Add a cheap identity check.** Match argv[0] basename from `/proc/$pid/cmdline` against `freenet`. Without it the reaper kills `screen -dmS fn freenet network`, disowned wrapper scripts, and second detached nodes on unrelated ports. (This was the external model's sole P1.)
5. **Pin the journald/backslash fix** — it is half of #5187 and currently nothing guards it. One line in the existing loop at `service.rs:1898`:
   ```rust
   assert!(!pre.contains('\\'),
       "{which} ExecStartPre must contain no backslash — systemd C-unescapes Exec \
        args and journald warns 'Ignoring unknown escape sequences' (#5187)");
   ```
6. **File a tracking issue for the macOS version-gated exit-43 path** (`macos.rs:272`), reference it from a `// see #NNNN` comment there, and record in the PR body *why* macOS is less urgent (the unconditional `pkill` at `macos.rs:298` means kickstart heals it).
7. **Fix the comments that are now false** — `linux.rs:463-465` and `:632` on shell parentage, and state explicitly that detached-but-current instances are now reaped. Call the behavior change out in the PR body / release notes: this is a user-visible semantic widening shipped under a `fix:` prefix.

#### Consider (Suggestions)
8. **Strengthen the absence pin.** `!pre.contains("$$ondisk") && !pre.contains("$$hv")` (`service.rs:1970`) pins two variable *names*; a gate reintroduced as `$$want`/`$$got` passes silently. Assert the mechanism instead: `!pre.contains("--version")` and `!pre.contains("/exe")`.
9. **Add system-unit behavioral coverage**, or at minimum pin that the two `ExecStartPre` bodies are identical after normalizing `{binary}`. They are byte-identical today; nothing enforces it, and the system template is a hand-duplicated literal.
10. **Wire the dead `pgrep_pid` hook** (`service.rs:2341`) to cover the self-PID skip behaviorally, or drop the parameter.
11. **Retire or repurpose case (d)** (`service.rs:2400`) — it is inert. The genuinely uncovered fail-safe is `/proc/$pid/stat` missing/unreadable → empty `$ppid` → no kill.
12. **Add `$$w` to the doubling needles** (`service.rs:1903`) and `[ $w -lt` to the bad list.
13. Refresh the stale harness comments (`service.rs:2323`, `:2328`, `:2330`, `:2378`).
14. Cap victims per run (`linux.rs:477`/`:633`) so the serial loop can't exceed the default `TimeoutStartSec=90`.
15. Add the operator recovery note (already-wedged hosts need a manual `freenet update` / `freenet service install`) to the PR body or #5187.
16. Restructure the PR description under the `## Problem / ## Solution / ## Testing / ## Fixes` headings.
17. Mention the removed `timeout 5 … --version` calls as a start-latency win.

---

### What's Done Well

Worth saying plainly: the diagnosis in #5187 is correct and precisely argued, the "strict superset" framing is accurate for every case the #3967 round-2 review cared about, the `sed`-to-final-`)` parse (MUST-FIX #6) is preserved, the instinct to add a *negative* pin against the old gate is exactly right, and the test refactor strengthens rather than launders coverage. The `cut`/`awk` swap is genuinely equivalent — I tested it against adversarial `comm` values specifically expecting to find a hole and did not. The PR description is honest about what it does, including pre-emptively scoping out macOS.

---

### Verdict
**HEAD SHA reviewed:** `7ca5cb86501009ebfeab99279a3d0e347b0ed3e2`

Must Fix #1 is the blocker: the discriminator needs to distinguish "adopted by init" from "supervised by systemd", which `PPID==1` cannot do under a system unit. A cgroup check is a small patch. Should Fix #2–#4 are on the kill path and go from rarely-reached to routinely-reached with this change, so they matter more now than they did under #3967. #5 (journald pin) is one line and closes the untested half of the issue.

Because #1–#4 are all in the concurrency/process-lifecycle and deploy-config high-risk surfaces, a fresh full review on the updated HEAD is required before merge rather than a diff-of-the-diff check.

[AI-assisted - Claude]

iduartgomez (changes requested):

sanity:
Status update, since this has sat since 7 Aug with a `CHANGES_REQUESTED` on it and no word from us. Sorry for the silence. Nothing here is a rejection.

**Are the requested changes addressed? No.** The head commit is unchanged since the review, so the one blocking finding is still open, and that is the only thing gating this. CI is green, the diagnosis in #5187 is correct and precisely argued, and the review was largely positive. This is a fix we want.

One practical note before the substance: I am describing this behaviourally rather than pointing at line numbers on purpose. Both files this PR touches are being rewritten right now by #5242 (+492/-14 across the same two files), landing shortly, so any line reference I gave you today would be stale before you read it. You will need a rebase regardless, and #5242 is worth reading first: it changes how a deliberate systemd stop gets classified, which overlaps one of the review's secondary findings below.

## The blocker

The reaper's discriminator is "the parent is PID 1". For the **user** unit that correctly means "adopted by init", because a supervised process there is parented by `systemd --user`. For the **system** unit it means the opposite: a `Type=simple` service under the system manager is forked directly by PID 1, so a perfectly healthy supervised node has PPID 1 by construction. Both unit templates get the same version-blind reaper, and neither the code nor its comments distinguish the two cases.

That is only reachable when both units exist for the same user, but the code explicitly contemplates that state. The mode-selection helper exists specifically to resolve the both-installed case, the system unit runs as the same UID that owns the user unit, and neither installer checks for or removes the other. So the user unit's pre-flight can find the system unit's healthy main process, SIGTERM it, then SIGKILL it, which systemd classifies as a crash for rollback purposes. The result is that unit wedged into exactly the success-but-dead state this PR exists to eliminate.

## Next step

**One question first, because the answer may shrink this to a very small patch: was the wedge you hit in #5187 on a user unit or a system unit?** If it was a user unit, scoping the version-blind reap to the user template and leaving the system template as it is would be small and obviously safe. That is where I would start.

If it has to cover both, the behaviour you want is: reap a process that init adopted, and never one that systemd is actively supervising. The cheapest way to tell those apart is the process's cgroup rather than its parent. A systemd-supervised process sits inside its unit's own cgroup slice, whereas a genuine #3967 orphan has been reparented out of one. That asks a far more precise question than PPID can, still reaps the orphans #3967 cared about, and needs no install-time changes.

Two others on the same kill path are worth folding in while you are there, because they go from rarely-reached to routinely-reached the moment a healthy node can be a target:

- **Give the process longer to drain before escalating to SIGKILL.** The current escalation window is shorter than the node's own shutdown drain default, so once a live node is a possible target the normal outcome is a hard kill mid-drain with in-flight operations aborted. A SIGKILL also reads as a crash to the rollback machinery, which can count toward crash-loop probation.
- **Check the victim's identity before killing it.** With the `--version` probe gone, nothing verifies the target is even a freenet binary. Matching argv[0]'s basename from the process's cmdline is enough, and it is deliberately *not* the "exe equals the on-disk binary" anchor the existing tests rightly forbid. That anchor fails because a #3967 orphan runs a different binary; a basename check still passes there. Without it the reaper will also kill a `screen -dmS` session whose argv happens to contain the pattern, a disowned wrapper script, and a second deliberately-detached node on an unrelated port.

Everything else in the review is "should fix" or "consider" rather than gating. The one-line journald pin is worth taking, since it guards the currently-untested half of #5187.

## After that

Push a fix and re-request review, and we will turn it around promptly this time rather than in three weeks. Because the blocker sits in the process-lifecycle and deploy-config surfaces, it gets a fresh full review on the new head rather than a diff-of-the-diff check.

If you would rather not carry this further, say so and we will pick it up from here. The analysis in #5187 is the hard part and it is already done.

[AI-assisted - Claude]

sanity:
Following up on the 24 Aug status update. We ran a fresh full-tier review on the current head (`7ca5cb8`) rather than leaving you waiting on the last one, and it turned up enough new material to be worth writing down — including a consequence that makes the blocker more serious than that comment made it sound. None of this is a rejection; the diagnosis in #5187 is still correct and this is still a fix we want.

First, a correction to that comment: **#5242 has not landed.** It is still open, so "landing shortly, read it first" was wrong and you should not be waiting on it. You do still need a rebase — the branch is 155 commits behind main — but main has touched `service.rs` / `service/linux.rs` only twice since your merge base (#5460, help text; #5289, macOS PATH), and neither goes near the pre-flight. The version-gated reaper is still on main verbatim, so nothing has superseded this.

## The blocker, and why it is worse than we said

The 24 Aug comment described the system-unit case as "a healthy node gets killed". Verified, and the mechanism is as described: the system unit is `Type=simple` (`service/linux.rs:611`) under the system manager, so its main process is a child of PID 1 for its entire healthy life, and both templates now carry the identical version-blind body (`:477` user, `:633` system). `use_system_mode` (`:62-65`) exists specifically to arbitrate the both-installed case and neither installer removes the other, so the configuration is contemplated rather than excluded.

What we had not traced is where that kill *goes*:

- `rollback.rs::classify_stop` (`:701-718`) returns `NotCrash` only for `0`, `2`, `42`, `43`, `44`. Everything else — including `KILL`/137 from the reaper's SIGKILL — falls to `_ => StopClass::Crash`. The module docs at `:20-23` say signal deaths are counted deliberately.
- `ROLLBACK_CRASH_THRESHOLD = 3` (`rollback.rs:126`). Three crashes inside the probation window restore the previous binary **and** pin the current version as known-bad so it is refused on re-update.

Both units run as the same user and share one state directory, so a reaper that repeatedly kills a healthy supervised node does not merely take it down — it can force a version downgrade and blacklist a good release. Ian's note on #5187 today, "be extremely careful not to break auto-update when fixing this", is exactly this hazard. That is what moves this from "an unlikely misconfiguration" to something we have to get right before merging.

The fix direction from the 24 Aug comment stands and is still the cheap one: discriminate on the **cgroup** rather than the parent. A systemd-supervised process sits in its unit's own cgroup slice; a genuine #3967 orphan has been reparented out of one. `/proc/$pid/cgroup` answers that question precisely, reaps everything #3967 cared about, and needs no install-time change.

**And the question from that comment is still the one that could shrink all of this to a small patch: was the wedge you hit in #5187 on a user unit or a system unit?** If it was a user unit, scoping the version-blind reap to the user template alone is small and obviously safe, and the cgroup work can be a separate change.

## Also worth folding in

Four smaller things, all verified against your head:

1. **The escalation window is outside a lockstep the unit itself documents.** The reaper waits 12 seconds before SIGKILL (`linux.rs:477`, `:633`), but the node's shutdown drain defaults to 30s (`config.rs:1663-1664`), and `TimeoutStopSec=45` carries the comment "If you raise `shutdown-drain-secs`, raise this in lockstep" (`:510-516`). The reaper was never brought into that lockstep, so the unit gives itself 45s to stop this binary and gives a stranger 12s. For a true orphan that is defensible; combined with the blocker it is the step that turns a wrong target into a `KILL` status and therefore into a rollback crash count.

2. **macOS has the same wedge, still live.** `heal_stale_orphan_or_defer` (`service/macos.rs:254-292`) has the identical `ppid==1 && (mismatch || empty)` gate, and on the defer path the wrapper logs "Another instance (current version) is already running" and `exit 0` (`macos.rs:383-384`) — with `KeepAlive.SuccessfulExit=false`, launchd then never respawns. That is the same success-but-dead state this PR exists to eliminate. "Deliberately out of scope" is a reasonable scoping call, but without a tracking issue it reads as "not a problem", and the test at `service.rs:2171-2176` will then assert same-version holders must be *deferred* while the new Linux test asserts they must be *killed*. Either fix it here or file it and link it from the body.

3. **Test case (d) is now vacuous.** `service.rs:2400-2409` chmods the on-disk binary to `0644` and asserts the orphan is still killed, described as covering the unreadable-on-disk-version path. The pre-flight no longer reads the on-disk binary at all, so that chmod cannot change the result: the call is identical in effect to case (a) eight lines above. It reads as coverage and provides none.

4. **The absence-pin is spelling-specific.** `service.rs:1970-1974` asserts `!pre.contains("$$ondisk") && !pre.contains("$$hv")`. A reintroduced gate spelled `$$ver` or `$$cur` passes it. The behavioural test is the real guard here; the pin mainly adds confidence it does not earn.

## Where this leaves us

The honest summary is that the hard part — the diagnosis in #5187 — is done and correct, and what remains is a discriminator change plus a decision about scope. You have been waiting a long time for that to be said clearly, and we are sorry it took this long twice.

Two ways forward, both fine by us:

- **You carry it.** Answer the user-vs-system question, push the change, and we will review on the new head promptly rather than in weeks. Because the blocker sits in process lifecycle and deploy config, it gets a fresh full review rather than a diff-of-the-diff.
- **We carry it.** If you would rather not spend more time on this, say so — or just say nothing for a week and we will take that as the answer — and we will open a follow-up PR that keeps your commits and credits you as author, with this PR closed pointing at it. The analysis is the part that was hard to get, and it is yours either way.

What we will not do is leave it sitting silently again.

[AI-assisted - Claude]

sanity:
One thing off your plate: I filed the macOS half as #5577 rather than leaving it as an action for you. It records the mechanism (`heal_stale_orphan_or_defer` at `service/macos.rs:254-292`, then `exit 0` at `:383-384` against `KeepAlive.SuccessfulExit=false`), notes the two platforms' tests currently assert opposite invariants, and carries forward the warning about not copying this PR's current shape.

So point 2 in my previous comment is done — scoping macOS out of this PR is fine as it stands, and nothing further is needed from you there.

[AI-assisted - Claude]

Issue #5191, opened 2026-08-05
tracking: 0.2.121 bandwidth follow-ups deferred from 0.2.120 (#5186, #5189, #5190, telemetry)


sanity:
## Purpose

0.2.120 ships the #5147 originator target list. Four things were deliberately **not** fixed in it, each for a stated reason. This issue exists so none of them is lost, and so the next person picks them up in an order that makes sense rather than one at a time.

Ordering matters here: three of the four touch the same call path (how summaries reach peers), so doing them separately means designing it three times.

---

## 1. #5190 — suppression over-excludes peers from the summary notification `T-bug`

A peer suppressed by the target list gets **neither** the broadcast **nor** the standalone summary, so its cached summary of us goes stale until the ~5-minute heartbeat. #5147 added a filter at a call site that `advertised_cohost_pub_keys`'s rustdoc says in bold must be added in the accessor, precisely to prevent this.

**Why it did not block 0.2.120:** the target list is version-gated at (0,2,120), so impact ramps from zero as the fleet upgrades. And the measured benefit already includes this defect — every #5147 simulation ran with it present — so it is upside to fix, not a correction to a claim.

**Do not take the "cheap" fix.** Simply not excluding co-hosts from the notification ships ~33 KB of summary per contract to every co-host on every state change; at degree 16-17 that plausibly costs more than the target list saves.

The pin also needs strengthening: it currently asserts both sites read the same *source*, which cannot see a call-site narrowing. It should assert the exclusion set is a **subset of the post-filter broadcast set**.

## 2. #5189 — piece 2, as a trigger + one reply arm (NOT a new protocol)

Read the correction comment on that issue before anything else. The short version: the inverted protocol already exists (`SummaryDigests` → `DigestVerdict` → `Summaries` → `SyncStateToPeer` heal), deployed since #4965. Piece 2 is (a) a commit-time trigger that announces a digest instead of the payload for the lazy class, and (b) the one missing reply arm — today the exchange terminates at `Summaries`, so a digest receiver that finds *itself* stale does nothing.

Zero new wire variants, and it inherits the deployed version floor, the resync throttles, the emit budget and the `broken_invariants` egress gate. A parallel protocol re-opens each of those storm classes.

**Split on `delta_send_expected`, not on size:** eager-push iff (cached summary held AND payload small), announce otherwise.

**This is where #5190 should be fixed**, since both concern how summaries reach peers.

## 3. #5186 — over-suppression under concurrent writers is unmeasured

[...]

PR #5192, opened 2026-08-05
chore(release): flip both version markers for 0.2.120

sanity:
## Problem

0.2.120 carries two wire-gated features whose version floors were written as *predictions* about the next release:

- `GATEWAY_ACK_VERSION_MIN_VERSION = (0, 2, 120)` (#5167, version-carrying connection ack)
- `BROADCAST_TARGET_LIST_MIN_VERSION = (0, 2, 120)` (#5147 / #5179, originator target list on broadcast)

Each has a companion `*_SHIPPED_IN` marker, currently `None`, meaning "not yet shipped, so the floor must stay strictly above the crate version". Cutting 0.2.120 makes both predictions true, so both markers have to move in the same commit that precedes the cut.

These gate **wire variants whose failure mode is a closed connection, not a degraded feature**. A peer at or above the floor without the code receives a `BroadcastToV2` variant index it has no arm for, `decode_msg` fails, and the link drops. The release cascade upgrades gateways first, so getting this wrong is fleet-wide churn.

## Approach

Both markers set to a **literal** `Some((0, 2, 120))`.

Deliberately **not** `Some(<FLOOR_CONSTANT>)`. That form compiles, passes, and reads as tidier, but it makes each guard's `shipped == floor` assertion compare the floor against itself, so that arm can never fail again. The `BROADCAST_TARGET_LIST_SHIPPED_IN` rustdoc previously instructed exactly that wrong resolution, so this PR corrects the doc rather than leaving it to mislead the next release.

## Timing constraint

Both markers **disarm their own tripwire the moment they are flipped** — the `Some(_)` arm only asserts `shipped == floor` and cannot tell whether the code actually shipped.

- Merge and cut in the same sitting.
- **If 0.2.120 slips, revert this commit.** A standing flip means a renumbered release leaves the floors naming 0.2.120, and a peer on the real 0.2.120 then reads as at-floor, receives a variant it cannot decode, and drops the connection.

## Testing

Guard tests `ack_version_floor_tracks_the_shipping_release` and `broadcast_target_list_floor_tracks_the_shipping_release` cover the constants directly.

Pre-flight soak gates, all green on **framework** (NATed residential peer, 2h on merged main `795ab3317`):

| Gate | Result |
|---|---|
| Outbound bandwidth | **−25.3%** vs matched time-of-day 0.2.118 window (2.38 vs 3.19 GB/day), inside #5164's predicted −20–40% |
| Stability | 0 restarts, 0 panics; every error class decayed into 0.2.118's own steady-state hourly range |
| River smoke test | HEALTHY (6/6 checks, 0 console/page errors), with live message delivery after the swap |

Note the soak **cannot** validate #5147 itself: the target list is version-gated at (0,2,120) and framework's peers are all below it, so the gate stays shut. #5147's real-world effect is a post-release telemetry question as the fleet adopts 0.2.120, tracked in #5191.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -1076,9 +1076,14 @@ pub(crate) fn version_supports_hash_first_summaries(
 /// must equal the floor.
 ///
 /// RELEASE-TIME ACTION: when the release carrying this feature is cut, set this
-/// to `Some(BROADCAST_TARGET_LIST_MIN_VERSION)` and freeze both.
+/// to a LITERAL `Some((major, minor, patch))` of that release and freeze both.
+///
+/// Write the literal, NOT `Some(BROADCAST_TARGET_LIST_MIN_VERSION)`. Naming the
+/// constant compiles, passes, and reads as tidier, but it makes the guard's
+/// `shipped == floor` assertion compare the floor against itself, so that arm
+/// can never fail again, which is the entire point of the guard.
 #[cfg_attr(not(test), allow(dead_code))]
-pub(crate) const BROADCAST_TARGET_LIST_SHIPPED_IN: Option<(u8, u8, u16)> = None;
+pub(crate) const BROADCAST_TARGET_LIST_SHIPPED_IN: Option<(u8, u8, u16)> = Some((0, 2, 120));
 
 /// Version floor for the originator target list on broadcast (#5147).
 ///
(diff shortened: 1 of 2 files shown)


sanity:
Closing: right change, wrong branch. The rule review caught a real ordering bug that the release checklist I was following had wrong.

`broadcast_target_list_floor_tracks_the_shipping_release`'s `Some` arm asserts `floor <= current` against `env!("CARGO_PKG_VERSION")`. On main that is still `0.2.119`, so `(0,2,120) <= (0,2,119)` fails and this PR would have gone red in Unit & Integration. But leaving the markers `None` while the release bump raises Cargo.toml to 0.2.120 fails the *other* arm (`floor > current`). Neither ordering works on its own.

The only commit where both guards pass is one where Cargo.toml is already 0.2.120 AND both markers are `Some`. That is the release PR itself. Precedent: 0.2.116 (#5083) — the bot's bump commit (5bd80b7ba) touched only version files, and the `HASH_FIRST_SHIPPED_IN` flip was pushed onto `release/v0.2.116` before the squash-merge landed both together as 1b3bf6cab.

So the flip moves onto `release/v0.2.120` once release.yml generates it. Commit 09da58d11 on this branch is preserved for that cherry-pick; it also carries a rustdoc fix, since `BROADCAST_TARGET_LIST_SHIPPED_IN`'s docs instructed the reader to write `Some(BROADCAST_TARGET_LIST_MIN_VERSION)` — the form that makes the `shipped == floor` arm compare the floor against itself.

[AI-assisted - Claude]

PR #5195, opened 2026-08-05
fix(telemetry): phase mix rollups off the shadow-stat boundary

sanity:
> **Corrected after production measurement.** The first version of this description claimed this bug drops one rollup per minute on every gateway. That is wrong, and the correction is in "Measured impact" below. The code change is unchanged; only the justification is.

## Problem

There are seven shadow rollup streams (`KnownShadowRollup::ALL`) but the shadow sub-budget `MAX_SHADOW_EVENTS_PER_SECOND` is 6, and admission uses a per-second tumbling window.

Every shadow aggregator is spawned from the same block in `p2p_impl.rs`, so their tickers start within microseconds of each other. The five shadow-stat streams close a 30 s window; the two mix rollups (`outbound_message_mix`, `broadcast_payload_mix`) tick every 60 s. So every 60 s all seven land in the same second and the seventh exceeds the sub-budget.

`MAX_SHADOW_EVENTS_PER_SECOND`'s docstring still described "five shadow streams" and "one slot of headroom for a sixth". Two streams were added since it was written, and nothing tied the constant to the stream count — that is the actual defect being fixed, and it is the reason an eighth stream would land silently too.

## Measured impact: small, and NOT the reason to care

Before writing this up I measured it against production telemetry (1,382 nodes, per-stream counters from `NetworkEfficiencyV1.shadow[i][5]`, which is `dropped_shadow_limit_total`):

**The sub-budget has dropped 54 events fleet-wide, all-time — 0.0005% of shadow rollups.**

The collision is real but nearly unreachable, because `admit_event` (telemetry.rs) checks the **aggregate** cap *before* the shadow sub-budget and returns early on failure:

```rust
if self.events_this_second >= MAX_EVENTS_PER_SECOND {   // 10
    ...; return false;                                  // <-- shadow dies here
}
if priority == EventPriority::Shadow {
    if self.shadow_events_this_second >= MAX_SHADOW_EVENTS_PER_SECOND { ... }
}
```

On any node already emitting 10 events in that second, all seven rollups are charged to the aggregate cap and the sub-budget branch is never reached. Fleet-wide the aggregate cap drops **77% of operational** and **31% of shadow** telemetry; per node, operational loss is p25 46% / median 64% / p90 82%.

So the shadow sub-budget is close to dead code in production, and **the real telemetry-loss problem is the 10/s aggregate cap.** That is a separate and much larger issue, filed as #5197 — it is not what this PR fixes, and this PR should not be read as addressing it.

## Why land this anyway

Because the two interact, and in an order that matters:

1. It is a genuine latent defect. Seven aligned streams against a cap of six is wrong regardless of what currently masks it, and nothing ties the cap to the stream count, so an eighth stream lands silently.
2. **It is a prerequisite for fixing the aggregate cap.** Whatever relieves the aggregate starvation (a shadow floor, a burst allowance, a higher cap) makes the sub-budget the binding constraint for the first time. Fixing that one without this one would relocate the drops onto the sub-budget rather than eliminate them, and the alignment would start firing exactly when someone is relying on the instrument.
3. It slightly relieves aggregate pressure today: the collision second carries 5 shadow events instead of 7.

If reviewers disagree that (2) justifies landing it ahead of the aggregate fix, closing this and folding the offset into that work is a reasonable alternative. I would rather land the prerequisite first than carry an unexplained alignment into the change that makes it matter.

## Approach

Offset both minute rollups by 15 s rather than raising the cap to 7. Raising the cap would leave only 3 of the 10 aggregate slots per second for operational telemetry, which — given operational telemetry is already losing 77% — is the wrong direction. The offset costs no telemetry volume.

15 s is half the shadow window, maximising distance to the nearest 30 s boundary and so tolerating spawn-time skew best. Both mix rollups get the same offset, preserving the shared cadence that lets them be joined per node-minute without interpolation. `outbound_message_mix` now aliases both constants from `broadcast_payload_mix` rather than restating the cadence, so the value the join depends on cannot drift.

Starting each ticker with `interval_at` at the first real emit also replaces the skip-the-immediate-tick dance, and makes the first rollup's reported window match the period it actually accumulated over.

The docstring is rewritten to state the invariant it actually carries: **not** "cap >= number of streams" but "cap >= the most streams that can share one second".

## Testing

`shadow_rollup_emit_schedule_fits_the_shadow_sub_budget` derives the emit schedule from the production cadence constants and drives the **real `admit_event`**, so a change to the admission policy is caught too, not just the arithmetic.

- Asserts no stream is dropped, naming the stream and the second.
- A vacuity guard asserts the model actually aligns at least two streams in some second — a schedule that never collides would otherwise pass while testing nothing.
- The schedule array is sized by `KnownShadowRollup::ALL.len()`, so an eighth stream **fails to compile** until its cadence is declared.

**Mutation-checked**: zeroing `ROLLUP_PHASE_OFFSET` fails with `broadcast_payload_mix was dropped at t=60s: 7 shadow rollups share that second, against a sub-budget of 6`.

**Known limitation, stated plainly:** that test exercises shadow events in isolation, with no operational traffic in the window. That is precisely why it passes while the production path is dominated by the aggregate cap — the test cannot see the interaction that makes this bug nearly unreachable. It guards the alignment invariant, not the production drop rate.

Also run: 126 tests across the three touched modules; `cargo fmt --check` and both of CI's clippy invocations clean.

## Note for reviewers

`cargo clippy --all-targets` reports 46 pre-existing errors in untouched files. CI runs clippy without `--all-targets`. Unrelated; see #5196.

Refs #5191. Real telemetry-loss problem: #5197.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -93,6 +93,15 @@ mod network_bridge;
 /// `network_bridge` is private outside this module.
 pub(crate) use network_bridge::broadcast_payload_mix::ApplyOrigin;
 
+/// Re-exported for `tracing::telemetry`'s shadow sub-budget guard, which has
+/// to name the minute rollups' cadence and phase to derive the worst-case
+/// number of shadow events in one second. Test-only, and `network_bridge` is
+/// private outside this module.
+#[cfg(test)]
+pub(crate) use network_bridge::broadcast_payload_mix::{
+    ROLLUP_PHASE_OFFSET as MIX_ROLLUP_PHASE_OFFSET, ROLLUP_WINDOW as MIX_ROLLUP_WINDOW,
+};
+
 // Re-export fault injection types for test infrastructure.
 // No cfg gate: underlying items are unconditionally compiled and integration
 // tests compile the lib without cfg(test).
--- a/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
+++ b/crates/core/src/node/network_bridge/broadcast_payload_mix.rs
@@ -323,21 +323,42 @@ pub(crate) const ROLLUP_WINDOW: Duration = Duration::from_secs(60);
 /// each other. The five shadow-stat streams close a
 /// [`crate::transport::shadow_stats::SHADOW_ROLLUP_WINDOW_SECS`] (30 s) window
 /// and the two mix streams a 60 s one, so without an offset all seven emit in
-/// the SAME second every 60 s. The telemetry worker admits at most
-/// `MAX_SHADOW_EVENTS_PER_SECOND` (6) shadow events per second, so on a
-/// gateway — the only place all seven run, since `shadow_reference_ping` and
-/// `shadow_iface_tx` are gateway opt-ins — exactly one rollup was dropped
-/// every minute, and WHICH one depended on task-wakeup order. Gateways are
-/// precisely where these rollups get read.
+/// the SAME second every 60 s, against a `MAX_SHADOW_EVENTS_PER_SECOND` of 6.
 ///
-/// Offsetting is preferred over raising the shadow sub-budget to 7: it adds
-/// no telemetry volume, whereas a cap of 7 would leave only 3 of the 10
-/// aggregate slots per second for operational telemetry.
+/// This is a LATENT defect, not a measured production loss. Do not restate it
+/// as one: an earlier version of this comment claimed gateways lost a rollup
+/// every minute, and production telemetry says otherwise on both counts.
+/// Measured over 1382 nodes, the shadow sub-budget has dropped 54 events
+/// fleet-wide in total, because `admit_event` tests the aggregate 10/s cap
+/// FIRST and returns early — that cap discards ~31 % of shadow and ~77 % of
+/// operational telemetry, and is the real bottleneck (see #5197). And the two
+/// opt-in streams are not in fact co-enabled on the production gateways, which
+/// run six streams against a cap of six and so cannot drop at all; only 2 of
+/// 1382 nodes run all seven.
+///
+/// The offset is kept anyway because the invariant is worth holding
+/// independently of today's numbers: nothing else ties the sub-budget to the
+/// emit schedule. The cap was chosen against a hand-counted five streams, two
+/// were added without revisiting it, and whoever changes the aggregate cap
+/// makes the sub-budget binding for the first time. The guard is the schedule
+/// test named below, not this constant.
+///
+/// Offsetting is preferred over raising the sub-budget to 7: it adds no
+/// telemetry volume, whereas a cap of 7 would leave only 3 of the 10 aggregate
+/// slots per second for operational telemetry that is already losing 77 %.
 ///
 /// 15 s is half the shadow window, which maximises the distance to the
 /// nearest 30 s boundary and so is the most tolerant of spawn-time skew.
 /// Applied to BOTH mix rollups equally, preserving the shared cadence that
 /// lets them be joined per node-minute without interpolation.
+///
+/// Known limit: the separation is probabilistic, not structural. Both cadences
+/// use `MissedTickBehavior::Delay`, so relative phase drifts on a node whose
+/// runtime stalls, and 15 s of margin is eventually consumable. Accumulated
+/// drift, not spawn-time skew, is the mode to watch. Relatedly, production
+/// shows sub-budget drops on nodes running only FIVE streams, which this
+/// schedule model says is impossible — so some source of emission bunching
+/// remains unidentified (see #5195 discussion).
 ///
 /// `telemetry::tests::shadow_rollup_emit_schedule_fits_the_shadow_sub_budget`
 /// derives the collision schedule from these constants and fails if this is
@@ -1637,6 +1658,13 @@ pub(crate) fn spawn_payload_mix_aggregator(
         // replaces the previous skip-the-immediate-tick dance, and makes the
         // reported window of that first rollup match the period it actually
         // accumulated over.
+        //
+        // Consequence worth knowing downstream: the FIRST window after every
+        // node start is 75 s, not 60 s. `rollup_window_secs` derives
+        // `window_secs` from real elapsed time, so the emitted record is
+        // self-describing and correct — but an analysis that hardcodes 60
+        // under-reports that one record by 20 %. Gateways restart on every
+        // auto-update, so this is not rare. Normalise by `window_secs`.
         let mut ticker = tokio::time::interval_at(
             last_rollup + ROLLUP_WINDOW + ROLLUP_PHASE_OFFSET,
             ROLLUP_WINDOW,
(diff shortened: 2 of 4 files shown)


sanity:
## Holding this: production contradicts the model my own test asserts

Measuring rather than trusting the reasoning has now falsified my justification twice, and turned up a third thing that argues against merging yet.

### The two falsifications (already folded into the description)

1. The shadow sub-budget drops **54 events fleet-wide, all-time** (0.0005%). `admit_event` checks the aggregate 10/s cap first and returns early, so the sub-budget is nearly unreachable. The real loss is the aggregate cap: 77% of operational, 31% of shadow. That is #5197.
2. **nova and vega run SIX shadow streams, not seven.** `shadow_iface_tx` has zero attempts on both, so they align 6 against a cap of 6 and cannot drop by construction. Fleet-wide only **2 of 1382** nodes run all seven, and those two drop about **1 per 480 minutes**, not the 1 per minute I claimed.

### The third thing, which is a finding against this PR's test

16 nodes fleet-wide show any sub-budget drop. **14 of them run only FIVE active streams**, all on 0.2.119 (so not an old-cap artifact — every one is on the current cap of 6):

| version | drops | active streams | uptime |
|---|---|---|---|
| 0.2.119 | 8 | 5 | 275 min |
| 0.2.119 | 6 | 5 | 2675 min |
| 0.2.119 | 4 | 5 | 1175 min |
| 0.2.119 | 2-3 | 5 | various |
| 0.2.119 | 6-7 | **7** | 365 / 2885 min |

Five streams against a sub-budget of six should be structurally impossible to exceed. Reaching the sub-budget branch at all requires **six** shadow events in one window, and a five-stream node should never produce more than five.

So there is an unmodelled source of shadow-event concurrency, and **the schedule model in `shadow_rollup_emit_schedule_fits_the_shadow_sub_budget` is contradicted by production.** The test is internally consistent and correctly mutation-checked, but a model that says "max 5 in a second" while production shows at least 6 is not faithful, and a fixed phase offset may be the wrong *shape* of fix if emissions can bunch for a reason I have not found.

Hypotheses I checked and **eliminated**, so nobody re-treads them:

- **Catch-up bursts.** Every aggregator sets `MissedTickBehavior::Delay` (`rolling_rtt_stats.rs:316`, `shadow_demand.rs:350,468`, `reference_ping.rs:130`), so missed ticks are skipped, not replayed. A stall yields one tick per stream, not several.
- **Unknown shadow emitters.** `send_standalone_shadow_event_with_peer_id` has exactly 7 call sites, matching the 7 known streams. No shadow-priority event escapes `KnownShadowRollup`.
- **Per-peer emission.** `spawn_aggregator` (`rolling_rtt_stats.rs:307-335`) emits once per window per node, inside the `window.ticks >= SHADOW_ROLLUP_WINDOW_SECS` branch — not once per peer.
- **Old-cap version skew.** All 16 droppers are on 0.2.119.

Still open: the lazily-reset rate-limit window (`admit_event` anchors the second to the first arriving event, not to a wall-clock boundary), and ticker drift under `Delay`. Neither obviously produces six emissions from five streams. **I could not determine the mechanism from static reading.**

### What I am doing about it

Not merging. The code is correct as far as it goes and CI can run, but I am not landing a fix whose stated justification has been wrong twice while an unexplained counterexample to its own test model sits in production data.

Reviewers: the useful question is no longer "is this diff correct" — I believe it is — but **"is a fixed phase offset the right shape of fix, given emissions demonstrably bunch beyond the modelled cadence?"** If the answer is no, this should be closed and the offset folded into #5197.

Raw per-node data behind the tables is in the working notes; the counters are `NetworkEfficiencyV1.shadow[i][5]` (`router.rs:275`, populated `ring.rs:2150`), which ride `router_snapshot` at `EventPriority::NetworkEfficiency` and therefore bypass the very limiter they report on.

[AI-assisted - Claude]

sanity:
## Big-picture lens findings addressed (`f612ea918`)

All BLOCKING and MAJOR items from the big-picture review are fixed. That lens's own verdict was **land, not close**, but on a rewritten rationale — and it explicitly retracted a prior argument of its own (that the offset meaningfully relieves aggregate-cap pressure; on nova that is 6 slots to 4 in a system dropping ~1400 events/min, which is noise). I have not used that argument.

**BLOCKING — false production claim written into two docstrings.** This was the important catch: I corrected the PR description after the measurement but left the falsified claim in `broadcast_payload_mix.rs` and `telemetry.rs`, where the next person to touch the sub-budget would inherit it as fact. Both now state the measured truth (54 sub-budget drops fleet-wide; aggregate cap is the real bottleneck; gateways run six streams, not seven) and explicitly warn against restating it as a measured loss.

**MAJOR — stale `SHADOW_STREAMS = 5` test contradicted the new invariant.** `test_all_aligned_shadow_rollups_are_admitted` asserted `MAX_SHADOW_EVENTS_PER_SECOND > 5` "with headroom" — a hardcoded stream count drifting from reality, which is the exact defect this PR exists to close, sitting in the more discoverable location. Renamed to `test_all_aligned_30s_shadow_rollups_are_admitted`, scoped to the 30 s group, contradicting const-assert removed, and it now points at the schedule test as the real guard. Not deleted, per repo rules.

**MAJOR — the spawn sites were unpinned.** The mutation check covered the constant, which its own docstring already guards; the ticker lines were the unguarded half. Confirmed: no test drove either spawn function, so reverting `interval_at(...)` to `interval(ROLLUP_WINDOW)` left the whole suite green. Added bounded `include_str!` pins to both (AGENTS.md `fn_body` form, so they cannot self-match).

Mutation-verified, both directions:
```
outbound_mix_aggregator_applies_the_phase_offset_pin ... FAILED
payload_mix_aggregator_applies_the_phase_offset_pin ... FAILED
```

**MAJOR — the schedule test modelled the mechanism but not the condition.** It runs with zero operational load, so the aggregate branch is unreachable for the whole modelled hour and it proves a property of a system that does not exist in production. Rather than change that test (its terms are the alignment invariant's own), added `the_aggregate_cap_not_the_shadow_subbudget_is_what_binds_under_load`, which fills the aggregate cap with operational traffic and asserts every rollup is refused **and charged to the aggregate counter with the sub-budget counter left at zero**.

That test is worth more than the offset. It pins the *check ordering* — the fact that falsified my original diagnosis — and #5197 will change how the aggregate cap works. If it reorders those two checks, the drops silently move between counters and every reading built on them shifts.

**MINOR — first window is 75 s, not 60 s.** Now documented at the spawn site. `window_secs` is derived from real elapsed time so the record is self-describing, but an analysis hardcoding 60 under-reports that one record by 20%, and gateways restart on every auto-update. Flagged for whoever writes the 0.2.121 analysis: normalise by `window_secs`.

**MINOR — stale "room for exactly one more aligned rollup stream"** in `outbound_message_mix.rs` corrected: whether a stream fits is decided by the emit schedule, not a spare slot.

**Also recorded**: the offset's separation is probabilistic, not structural — both cadences use `MissedTickBehavior::Delay`, so relative phase drifts on a stalling node and 15 s of margin is eventually consumable. Accumulated drift, not spawn-time skew, is the mode to watch.

## Still open, still in draft

The unexplained counterexample from the previous comment stands: 14 nodes running **five** streams show sub-budget drops, which this PR's schedule model says is impossible. That is now noted in the `ROLLUP_PHASE_OFFSET` docstring rather than left only in this thread. Four hypotheses eliminated (catch-up bursts, hidden emitters, per-peer emission, old-cap skew); mechanism undetermined.

Staying in draft until the remaining review lenses report, since the open question is whether a fixed phase offset is the right *shape* of fix given emissions demonstrably bunch beyond the modelled cadence.

Verification: 129 lib tests pass across the three touched modules, `cargo fmt --check` clean, both CI clippy invocations clean.

[AI-assisted - Claude]

sanity:
**Backlog triage: recommending this be folded into #5197 rather than revived on its own.**

Re-verified the defect against `origin/main` @ `9196e4be6` — it is still real and still present:

- `tracing/telemetry.rs:459` — `MAX_SHADOW_EVENTS_PER_SECOND: usize = 6`, while `KnownShadowRollup::ALL` (`:71`) is still **7** streams.
- `broadcast_payload_mix.rs:316` and `outbound_message_mix.rs:197` both still `ROLLUP_WINDOW = Duration::from_secs(60)`, both still spawned with a plain `tokio::time::interval`. No `ROLLUP_PHASE_OFFSET` exists in main.
- The stale docstring this PR set out to fix is still at `telemetry.rs:440` ("There are five shadow..."), as is the drifting `SHADOW_STREAMS = 5` test constant.

So the PR is not superseded. The reason not to revive it as-is is the one its own author raised on 2026-08-05 and never resolved: the schedule model in `shadow_rollup_emit_schedule_fits_the_shadow_sub_budget` is **contradicted by production**. 14 of the 16 nodes showing sub-budget drops run only five active streams, which should be structurally impossible against a cap of 6. Catch-up bursts and unknown emitters were both ruled out, so there is an unmodelled source of shadow-event concurrency — and a fixed phase offset may simply be the wrong shape of fix for it.

The measured benefit also does not carry a revival on its own: 54 dropped events fleet-wide all-time (~0.0005%), and only 2 of 1382 nodes run all seven streams at once.

Meanwhile the problem this PR's own body identifies as the real one — **#5197**, the aggregate cap dropping 77% of operational and 31% of shadow telemetry — is still open and is the larger signal loss by three orders of magnitude.

Recommendation: leave this parked, and pick the phase-offset question up inside #5197's work, where the unmodelled-concurrency anomaly has to be understood anyway. Whoever takes #5197 should treat the five-streams-yet-six-drops contradiction as evidence to explain, not noise to route around — the docstring and `SHADOW_STREAMS` drift are worth fixing there regardless of what happens to the offset itself.

[AI-assisted - Claude]

Issue #5189, opened 2026-08-05
design(update): architect #5147 piece 2 as receiver-driven pull, not a probe grafted onto the send path


sanity:
## Summary

Piece 2 of the #5147 design ("size-thresholded probe") is deferred to 0.2.121. This issue is about **not grafting it onto the existing send path**, because the obvious implementation is both unsafe and, more importantly, points at a structural problem worth fixing instead.

## Why the obvious implementation is wrong

The probe applies to payloads ≥ `streaming_threshold` (64 KiB). Those go through the broadcast queue's **large lane**:

```rust
const DEFAULT_LARGE_PAYLOAD_CONCURRENCY: usize = 2;
```

`broadcast_to_single_peer` runs holding one of those two permits. Inserting "send probe → await reply → send payload" there means each large broadcast occupies **half the node's large-payload capacity for a full round trip before a single byte moves**. At the design's own 80 ms one-way figure that is ~160 ms of dead permit time per leg.

This is the code path behind #4145, #4231, #3519 and #3309 — every one of them a backpressure stall — and `.claude/rules/channel-safety.md` exists because of exactly this shape. A two-phase queue (probe stage that does not hold the send permit, payload re-enqueued on reply) would be *safe*, but it is a lot of new machinery: a parked-leg store keyed by (contract, peer, tx), a reply router, timeouts, and keep-latest semantics extended across both phases so a newer state arriving while parked replaces the parked leg.

Before building that, it is worth asking why we need it at all.

## The structural problem underneath

**The sender maintains a model of the receiver's state, and every defect on #5147 traces back to that model being wrong, missing, or stale.**

- **Missing** → `FullNoTheirSummaryTracked`: no cached summary, so we send blind full state. 26.9% of broadcast bytes at a 357 KB mean.
- **Stale** → `fanout_send_needed` skips on a *cached* summary, i.e. what we *believe* the peer holds. Before #5147 a wrong belief cost one leg; with the target list it is published one hop and suppresses that peer across every relayer.
- **Second-hand** → the target list is itself a model, of what a *different* sender did. It is right most of the time and the failure mode is the one #5186 is open about.

The probe as specified adds a fourth model-repair mechanism: ask the peer to confirm the model before acting on it. That is treating the symptom.

**The receiver always knows its own state exactly.** Nothing needs to model it.

## Proposed shape: announce → request → exact delta

For payloads above the threshold, invert the direction:

[...]

PR #5200, opened 2026-08-05
fix(update): narrow the proactive summary exclusion with the broadcast target set

### LINKED ISSUE #5190: fix(update): #5147 suppression over-excludes peers from the proactive summary notification — they get neither the broadcast nor the summary
## Summary

#5147 added a filter at a call site that `advertised_cohost_pub_keys`'s rustdoc explicitly says must not be added at a call site. The result: a peer suppressed by the target list gets **neither the broadcast nor the standalone summary notification**, so its cached summary of us goes stale until the ~5-minute InterestSync heartbeat.

The codebase predicted this precisely (`operations/update.rs:316-325`):

> The exclusion is only sound while the excluded set is a subset of the broadcast target set. If the two sites read different sources, **or one narrows its set and the other does not**, the notification over-excludes: a peer gets neither the broadcast nor the standalone summary, and every unit test stays green because each site is individually correct.
>
> **If you add a filter, add it HERE, not at a call site.**

## Mechanism

Two sites are meant to partition the co-host set:

- `get_broadcast_targets_update` broadcasts to advertised co-hosts, carrying our summary in `sender_summary_bytes`.
- `send_proactive_summary_notification` **excludes** advertised co-hosts from the standalone `Summaries` fan-out, on the reasoning that the broadcast already carried the summary (#4965).

#5147 narrowed the first set only. `update.rs:414` drops peers named by the originator's list:

```rust
if origin.covers(&pub_key) { skipped_covered += 1; ...; continue; }
```

while `update.rs:1166` still builds the exclusion set from the **unfiltered** accessor:

```rust
let advertised_cohosts: HashSet<TransportPublicKey> = op_manager
    .advertised_cohost_pub_keys(key)          // <- NOT narrowed
    .into_iter().collect();
```

and `proactive_summary_targets` skips every member of it (`update.rs:1291`).

So for a covered peer: suppressed from the broadcast, and skipped by the notification because it is still "an advertised co-host". It receives nothing.

## Why the existing pin does not catch it

`cohost_source_is_shared_between_broadcast_and_notification` checks that the two sites read the same *source*. They still do — the narrowing happens **after** the shared read, at the call site. The pin cannot see call-site filtering, which is exactly why the rustdoc asks for filters to live in the accessor.

## Severity

Not a correctness bug: convergence still heals via the ~5-minute heartbeat, and the direction is staleness, not corruption.

But it is a **third summary-starvation channel** on this feature, after (a) the target list removing the `sender_summary_bytes` piggyback it depends on, and (b) the version gate that made the whole feature inert. It matters because the summary-match gate does most of the real deduplication work — in the #5147 simulation `summary_skips` fell from 417 to 146 when the list engaged — so starving summary refresh partially undercuts the feature that caused the starvation.

Pre-#5147 the "gets neither" case was rare (a failed send). Post-#5147 it is every covered co-host on every relayed broadcast.

**Ramps with adoption:** the target list is version-gated at (0,2,120), so this only bites as the fleet upgrades. It does not affect nodes below the floor.

## Fix options

1. **Pass the actually-broadcast set as the exclusion set** (preferred — restores the stated invariant exactly). The exclusion must be "peers we really sent to", not "peers we might have sent to". Needs the notification path to see the post-filter target set.
2. **Subtract the covered set from the exclusion set.** Equivalent outcome, smaller change, but leaves two sources that must be kept in step — the thing the rustdoc is trying to prevent.
3. **Stop excluding co-hosts from the notification.** Trivially safe in direction (send more, not less) but discards #4965's measured saving.

Whichever is chosen, the pin needs strengthening: it should assert the exclusion set is a **subset of the post-filter broadcast set**, not merely that both derive from the same accessor. A source-identity check cannot see a call-site narrowing, and that is the whole failure mode here.

## Provenance

Found by an independent architecture review of the #5147 follow-on work, not by a test. Related: #5186 (over-suppression unmeasured), #5189 (piece 2 architecture). The fix likely belongs with #5189, since both concern how summaries reach peers.

[AI-assisted - Claude]

sanity:
## Problem

#5147 added a filter at a call site that `advertised_cohost_pub_keys`'s rustdoc says in bold must not be added at a call site. The result: a peer suppressed by the originator's target list gets **neither the broadcast nor the standalone summary notification**, so its cached summary of us goes stale until the ~5-minute InterestSync heartbeat.

Two sites are meant to partition the co-host set. `get_broadcast_targets_update` broadcasts to advertised co-hosts, carrying our summary in `sender_summary_bytes`. `send_proactive_summary_notification` **excludes** advertised co-hosts from the standalone `Summaries` fan-out precisely because the broadcast already carried it (#4965).

#5147 narrowed the first set only. The broadcast dropped peers named by the originator's list, while the notification still built its exclusion set from the unfiltered accessor. A covered peer was therefore suppressed from the broadcast *and* skipped by the notification.

The codebase predicted this exactly:

> The exclusion is only sound while the excluded set is a subset of the broadcast target set. If the two sites read different sources, **or one narrows its set and the other does not**, the notification over-excludes: a peer gets neither the broadcast nor the standalone summary, and every unit test stays green because each site is individually correct.
>
> **If you add a filter, add it HERE, not at a call site.**

Direction is staleness, not corruption, and it heals on the heartbeat. But it is a third summary-starvation channel on this feature, and the summary-match gate does most of the real deduplication work — in the #5147 simulation `summary_skips` fell from 417 to 146 when the list engaged — so starving summary refresh partially undercuts the feature that caused the starvation. Pre-#5147 the "gets neither" case needed a failed send; post-#5147 it is every covered co-host on every relayed broadcast. It ramps with adoption, since the target list is gated at (0,2,120).

## Approach

**Fix option 1 from the issue** (restore the stated invariant), implemented so that the narrowing lives where the rustdoc demands.

New `OpManager::broadcast_cohost_targets(key, origin)` returns the co-hosts this fan-out will *actually* send to, applying the #5147 narrowing inside the accessor. The broadcast sends to `targets`; the notification excludes `targets`. The excluded set is a subset of the broadcast set **by construction** rather than by two call sites agreeing to stay in step.

The raw `advertised_cohost_pub_keys` survives for the one caller that legitimately wants the un-narrowed population — `resolve_covered_peers`, which resolves an originator's list against everyone we could have targeted, where narrowing would be circular.

**The origin is threaded, not re-derived.** `drive_relay_broadcast_to` and its streaming sibling now hoist `resolve_covered_peers(...)` into a local and hand the same `BroadcastOrigin` to both the fan-out and the notification. Re-resolving at the notification would re-read the co-host population at a later instant and could disagree; defaulting it to `local()` would silently restore this bug.

`covered_skipped` is returned rather than recorded inside the accessor: two callers now make the same call and only one is performing a broadcast suppression, so recording inside would double-count. The count is still produced by the filter, per the "a filtering metric must come from the filter" rule — the caller only decides whether this invocation represents a suppressed send.

## Testing

**The existing behavioural guard passed with the bug present.** `notification_exclusion_is_a_subset_of_broadcast_targets` only ever used an **empty** covered set, and with nothing covered the broadcast set and the raw co-host set are identical, so the two sites cannot disagree. The bug lives entirely in the case that test does not construct.

New `covered_peer_suppressed_from_broadcast_still_gets_the_summary_notification` constructs it: peer A is an advertised co-host, interested, and named by the originator's covered list; peer B is a co-host that is *not* covered. It asserts both premises (A really is suppressed from the broadcast, B really is targeted), then that **A is notified** and **B is not** — the second half being what stops it passing under a mutation that simply notifies everyone and discards #4965's saving.

**Verified to fail without the fix.** Reverting the notification to the raw accessor:

```
covered_peer_suppressed_from_broadcast_still_gets_the_summary_notification ... FAILED
  #5190: A was suppressed from the broadcast by the #5147 target list, so nothing
  carried our summary to it — the standalone notification MUST reach it. notified={}
```

Empty: A received nothing.

**The structural pin is strengthened, because the old one could not have caught this.** `cohost_source_is_shared_between_broadcast_and_notification` asserted both sites read the same *source* — which remained true, since the narrowing happened after the shared read. A source-identity check cannot see a call-site narrowing, which is the whole failure mode. It now pins that the raw accessor is consumed exactly once in the file, by the narrowing accessor, and that both consumers go through that accessor with an origin. Both pins also fail under the revert.

Full suite: 4710 passed, 0 failed. `cargo fmt --check` and both CI clippy invocations clean.

## Known residual, deliberately not fixed here

The exclusion set is still read at a slightly later instant than the broadcast set (the notification is spawned as a background task). A co-host advertised in between would be excluded without having been broadcast to. That race is **pre-existing** — both sites always read independently — and this change does not widen it. Fixing it means passing the resolved target list across the spawn boundary, which is a larger change than the regression warrants.

Fixes #5190
Refs #5191

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -4190,8 +4190,14 @@ impl GlobalTestMetrics {
     /// Records one fan-out target dropped because the originator's list named
     /// it (#5147).
     ///
-    /// Incremented BY the filter in `get_broadcast_targets_update`, alongside
-    /// its own `skipped_covered` field, never re-derived from the difference
+    /// The count is PRODUCED by the filter — `OpManager::broadcast_cohost_targets`,
+    /// which returns it as `covered_skipped` — and incremented by its
+    /// broadcast-side caller `get_broadcast_targets_update`, alongside that
+    /// function's own `skipped_covered` field. The increment sits at the caller
+    /// rather than in the filter because #5190 gave the filter a SECOND caller
+    /// (the proactive summary notification), and only one of the two represents
+    /// a suppressed send; incrementing inside would double-count. The count is
+    /// still never re-derived from the difference
     /// between the resolved co-host set and the final target set — that
     /// subtraction also absorbs the sender, self, and resolve-failure filters,
     /// so it would keep reporting a plausible number after this filter is
--- a/crates/core/src/node/op_state_manager.rs
+++ b/crates/core/src/node/op_state_manager.rs
@@ -3158,6 +3158,19 @@ mod tests {
         }
         assert_eq!(result.skipped_covered, 2, "the filter counts its own drops");
         assert_eq!(got.len(), 3);
+        // `proximity_found` is the population BEFORE any narrowing, and it
+        // feeds a telemetry denominator (`tracing/register.rs`). #5190 moved
+        // the covered-peer filter into `broadcast_cohost_targets`, so the
+        // count is now `targets.len() + covered_skipped` rather than the raw
+        // list length. Asserted here — with a NON-EMPTY covered set, which the
+        // other two `proximity_found` assertions lack — because dropping the
+        // `+ covered_skipped` term silently halves that denominator with the
+        // whole suite green.
+        assert_eq!(
+            result.proximity_found, 5,
+            "proximity_found must still count every advertised co-host, \
+             including the ones the #5147 list suppressed"
+        );
     }
 
     /// A local apply carries no claim, so nothing is suppressed: byte-for-byte
--- a/crates/core/src/operations/update.rs
+++ b/crates/core/src/operations/update.rs
@@ -323,8 +323,58 @@ impl OpManager {
     /// **If you add a filter, add it HERE, not at a call site.** That coupling
     /// is the point of this function, and it is pinned by
     /// `cohost_source_is_shared_between_broadcast_and_notification`.
+    ///
+    /// That warning was ignored once and produced #5190: #5147 filtered the
+    /// originator's covered set at the BROADCAST call site only, so a covered
+    /// peer was suppressed from the broadcast and still excluded from the
+    /// notification, receiving neither. The narrowing now lives in
+    /// [`Self::broadcast_cohost_targets`] below, which BOTH sites go through.
+    ///
+    /// This raw accessor survives for one caller that legitimately wants the
+    /// UNnarrowed population: `resolve_covered_peers`, which resolves an
+    /// originator's list against everyone we could have targeted. Narrowing
+    /// there would be circular. Any site that participates in the
+    /// broadcast/notification partition must use `broadcast_cohost_targets`.
     pub(crate) fn advertised_cohost_pub_keys(&self, key: &ContractKey) -> Vec<TransportPublicKey> {
         self.neighbor_hosting.neighbors_with_contract(key)
+    }
+
+    /// The advertised co-hosts this fan-out will actually broadcast to, after
+    /// the #5147 target-list narrowing.
+    ///
+    /// This is the population the broadcast/notification partition is defined
+    /// over, and the ONLY thing either side of it may read. The broadcast
+    /// sends to `targets`; the notification excludes `targets` and therefore
+    /// still notifies the peers this narrowing dropped, which is the whole
+    /// point — they did not get the summary in a broadcast, because they did
+    /// not get a broadcast.
+    ///
+    /// `covered_skipped` is returned rather than recorded here because two
+    /// callers make this same call and only one of them is performing a
+    /// broadcast suppression. Recording inside would double-count the metric
+    /// (see the "a filtering metric must come from the filter" rule — the
+    /// count IS produced by the filter, the caller only decides whether this
+    /// invocation is the one that represents a suppressed send).
+    pub(crate) fn broadcast_cohost_targets(
+        &self,
+        key: &ContractKey,
+        origin: &BroadcastOrigin,
+    ) -> CohostBroadcastTargets {
+        let mut targets = Vec::new();
+        let mut covered_skipped = 0usize;
+        for pub_key in self.advertised_cohost_pub_keys(key) {
+            // #5147: the originator told us it already delivered to this peer,
+            // so re-sending is a byte-identical duplicate.
+            if origin.covers(&pub_key) {
+                covered_skipped += 1;
+                continue;
+            }
+            targets.push(pub_key);
+        }
+        CohostBroadcastTargets {
+            targets,
+            covered_skipped,
+        }
     }
 
     /// Resolve the set of peers to broadcast a state update to.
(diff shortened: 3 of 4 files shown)


sanity:
## Holding this: the simulation A/B shows the fix costs more traffic than it saves

CI agrees — **Simulation: FAILURE**, all 15 other checks green on `3d496d41b`. I reproduced it locally and ran a baseline to attribute it.

### The measurement

`test_5147_originator_target_list_cuts_duplicate_deliveries`, same seed, clean `origin/main` vs this branch:

| arm | sends | suppressed | summary_skips | accounted legs |
|---|---|---|---|---|
| control (#5147 off) | 460 | 0 | 417 | 877 |
| **main** (#5147 on) | 288 | 428 | 146 | ~877 |
| **this branch** | 275 | 429 | **334** | **1045** |

Clean `origin/main` **passes**; this branch **fails** the leg-accounting identity (gap 168 against a 43 tolerance). Not flaky, and not the newly-landed #5199 — the baseline run was on current main.

### The fix works, and that is exactly the problem

Restoring summaries to covered peers re-seeds their cached summary of us, so the summary-match gate revives: `summary_skips` climbs 146 → 334, back toward control's 417, and real sends fall 288 → 275. That is the intended effect, confirmed.

The extra legs come from the same mechanism. A `Summaries` recipient calls `upsert_peer_summary_from`, which **inserts the sender into its interested-peers map if absent** (#4952, `or_default()` under a cap). So the notifications keep more peers mutually tracked, enlarging *their* fan-out target sets on later broadcasts. The +168 legs are real work this change creates elsewhere in the network, not accounting noise.

### The trade, on the axis this workstream exists to shrink

- #5147 alone: saves **172 sends**, adds nothing, covered peers go stale for ~5 min.
- #5147 + this fix: saves **185 sends**, adds **up to 429 standalone `Summaries`** (one per suppressed peer).

Net roughly **+250 messages to buy 13 fewer sends**.

`.claude/rules/hosting-invariants.md` invariant 3 is explicit that per-peer broadcast **messages/s** is "the load-bearing storm signal, since a tiny-payload high-frequency storm is invisible in bytes" — the #4861 lesson. This lands squarely there, in the clique regime #5147 was built for.

**Measured vs inferred, stated plainly:** the +168 legs and +188 `summary_skips` are measured and prove the notifications land and change network state. The ~429 added messages is an **upper bound, not a measurement** — the 100 ms/contract throttle may collapse many of them, and `SuppressionArm` has no notification counter. That missing counter was called out in review before any of this was run, and it is the first thing to add.

### Why I am not just widening the tolerance

The assertion is doing its job. Its stated purpose is "a gap this size means the two runs did genuinely different work, so the reduction is not attributable to #5147" — and the two arms now genuinely do different work, because of this change. Relaxing the bound would convert a true finding into silence, which is the failure mode the surrounding comment block was written to prevent.

### Where this leaves the fix

The bug is real: a covered co-host receives neither the broadcast nor the summary. But it self-heals on the ~5-minute heartbeat, while this cure adds permanent message-axis load.

The better shape is the one review and #5189 independently point at: a peer that receives a covered list already knows those peers just applied the same update, so it could **upsert the same summary for every peer named on the list** — fixing the staleness locally, at zero traffic, instead of fanning out N notifications. That belongs with #5189 (piece 2), which is the natural home for "how summaries reach peers" and is where #5190 was originally scoped to land.

Recommendation: hold this branch, add the notification counter to `SuppressionArm` so the trade is measurable rather than bounded, and fold the fix into #5189 in the upsert-from-list shape.

Review findings from this round are addressed in `a2daf5293` regardless (non-blocking notification send, pin slice bound, counter rustdoc), so the branch is in a reviewable state if the trade is later judged acceptable.

[AI-assisted - Claude]

sanity:
Superseded by #5203, which fixes the same bug by DERIVING the covered peers' summaries instead of transmitting ours.

The measurement on this branch is what motivated the replacement: restoring the notification cost ~429 extra messages to save 13 sends (+250 net), on the messages/s axis hosting-invariants calls the load-bearing storm signal, and it failed the #5147 leg-accounting identity. A digest would not have helped either — the cost is one message per destination regardless of size.

#5203 costs zero messages and zero bytes, is identical to main on the single-writer arm, and measures **15 fewer sends and 22 fewer redundant deliveries** on the multi-writer arm.

The review findings raised here were all addressed before this branch was retired, and two of them carried over into #5203: the origin threading (now used to derive rather than to narrow) and the lesson that a pin must cover the CALL SITE, not just the helper — #5203's test failed exactly that mutation on its first draft and was rewritten to run through the production entry point.

The notification cost counter that made this measurable is split out as #5202.

[AI-assisted - Claude]

Issue #5186, opened 2026-08-05
test(update): #5147 over-suppression under concurrent writers is unmeasured — the multi-writer test cannot detect it


sanity:
## Summary

#5147's target list attests **delivery**: "A sent this payload to C". But a relayer re-broadcasts its own **post-merge** state, which under concurrent writers can be strictly newer than the payload it received. Suppressing C on A's say-so can therefore skip a leg carrying state C genuinely lacks.

The mechanism is bounded (one hop, one cycle, healed by the ~5-minute anti-entropy heartbeat) and the failure direction is under-delivery rather than corruption. **But it is currently unmeasured**, and this issue exists so that is not mistaken for measured.

## Why the existing tests do not answer it

`test_5147_originator_target_list_cuts_duplicate_deliveries` cannot reach the condition at all: **every one of its 20 updates originates at the gateway**, so all peers merge the same payload into the same prior state, the relayer's state and the originator's payload always agree, and suppression is correct by construction. That is the regime most favourable to the feature.

`test_5147_multi_writer_suppression_is_regime_dependent` rotates the writer, which makes relayer-ahead-of-recipient *reachable*. Its convergence assertions still do not constrain the hazard, and this was demonstrated rather than suspected:

> Mutation: make suppression **unconditional** — drop every advertised co-host on any relayed fan-out, ignoring the list. Result: 480 suppressions, sends 459 → 317, and the test stays **GREEN** with `diverged=0` and `replicas=7`.

Three independent properties of the setup explain that, each sufficient on its own:

1. **Topology.** 7 peers at `max_connections: 12` is near-clique, so no peer depends on a relayer for any update. Over-suppressing relay legs costs nobody state. The hazard needs a recipient whose *only* path to the newer state runs through the suppressing relayer.
2. **Merge semantics.** The CRDT fixture is an LWW register with strictly increasing versions (`mock_runtime.rs`), so missing an *older* update is harmless by construction. Worse, `apply_crdt_full_state` returns `CurrentWon` and re-broadcasts, which actively heals the exact case being probed.
3. **Settle window.** Convergence is checked after 90 simulated seconds of quiet — ample for that healing.

## What would actually close it

Any one of these makes the hazard observable; the first is probably the cheapest:

- **Sparser graph.** Lower `max_connections` so some peer's only route to an originator is through a relayer that will suppress it.
[...]

PR #5203, opened 2026-08-06
fix(update): derive covered peers' summaries instead of transmitting ours

### LINKED ISSUE #5190: fix(update): #5147 suppression over-excludes peers from the proactive summary notification — they get neither the broadcast nor the summary

sanity:
> **HELD — see the correction comment below.** Three reviewers found that (a) an earlier revision of this description made a mutation-verification claim that was not true of the shipped tree (now fixed and re-verified), and (b) the inference's first premise is contradicted by `target_list_for`'s own rustdoc: the covered list records who we ENQUEUED to, not who received. That turns a one-cycle wrong belief into durable local state that causes a silent `Skip`. The traffic measurement below stands; the safety argument does not, as written.

## Problem

#5147 lets a relayer skip fan-out legs the originator already covered. But a fan-out leg is also how a peer **learns our summary** — it rides along as `sender_summary_bytes`. Suppressing the leg starved the covered peer's cached summary of us, and #4965 separately excludes advertised co-hosts from the standalone `Summaries` fan-out. So a covered peer received **neither**, and stayed stale until the ~5-minute InterestSync heartbeat.

The accessor's rustdoc predicted the shape exactly: *"a peer gets neither the broadcast nor the standalone summary, and every unit test stays green because each site is individually correct."*

## Why not the obvious fix

I built the obvious fix first (restore the standalone notification to covered peers) and measured it. It costs more than it saves:

| #5147 A/B, single-writer | sends | suppressed | summary_skips | notif sent |
|---|---|---|---|---|
| control (#5147 off) | 460 | 0 | 417 | 0 |
| main (#5147 on) | 288 | 428 | 146 | 0 |
| notify-based fix | 275 | 429 | 334 | **429** |

Roughly **+250 messages to buy 13 fewer sends**, on the per-peer messages/s axis `.claude/rules/hosting-invariants.md` invariant 3 calls "the load-bearing storm signal, since a tiny-payload high-frequency storm is invisible in bytes" (#4861). It also failed the leg-accounting identity, because a `Summaries` recipient calls `upsert_peer_summary_from`, which inserts the sender into its interested-peers map — so notifying enlarges *other* peers' fan-out sets. Sending a summary is not a read-only act.

A digest instead of full bytes would not have helped: the cost is one message **per destination** regardless of size.

## Approach: derive it, do not send it

Being on the originator's covered list **means** that peer received and applied this same update from the same originator. If its prior state matched ours, its post-merge state equals ours, so its summary equals **our own post-merge summary** — which we are already holding.

So the relayer writes that into its own model of each covered peer. No message, no bytes.

The originator sends the same list to everyone on it, so this is **symmetric**: every covered peer derives our summary from its own copy at the same moment. That is what repairs the direction #5190 is actually about (their model of us), even though each node only ever writes its own map.

The #4965 co-host exclusion is therefore **unchanged** — covered peers stay excluded from the notification, and are no longer starved by it.

## Measured

Single-writer arm: **identical to main** (sends 288, suppressed 428, summary_skips 145 vs 146). Costs nothing. It cannot show a benefit either, because in that regime the same peers are covered every time, so their summaries are never consulted.

Multi-writer arm, where the originator rotates so a peer covered on one update is a fan-out target on the next:

| multi-writer treatment | main | this PR |
|---|---|---|
| sends | 420 | **405** |
| redundant deliveries | 288 | **266** |
| summary_skips | 178 | 181 |

**15 fewer sends and 22 fewer redundant deliveries, at zero added messages.** The fix pays for itself in the regime where the staleness it removes actually mattered.

`test_5147_originator_target_list_cuts_duplicate_deliveries`, `test_5147_multi_writer_suppression_is_regime_dependent` and `test_cohost_mesh_update_fanout_stays_delta_dominated` all pass.

## When the derivation is wrong

If a covered peer's prior state had diverged from ours, the derived summary is wrong and we may skip a send it needed, leaving it stale until anti-entropy corrects it.

That is the **same optimistic-cache trust model already in use**: #4442 caches a summary on `Delivered`, which is sender-side completion rather than a receiver ack, and relies on the same heartbeat and the delta-apply-failure resync to correct it. This adds a new *population* to that model, not a new assumption about it. The over-suppression risk it shares with #5147 is what **#5186** exists to measure, and this PR does not close that.

## Testing

`covered_peer_summary_is_derived_without_sending_anything` asserts both halves: the covered peer's model **is** populated, and **no message was sent** to populate it. Asserting only the first would pass equally for the notify-based fix.

Read-back goes through `begin_peer_summary_broadcast`, the same accessor the broadcast path uses, so the value is asserted visible to the code that consumes it rather than to a test-only peek. An uncovered co-host (B) is the control against a mutation that seeds everyone.

**Mutation-verified as of `c2d11cbb7`** — see the correction comment below. An earlier revision of this description claimed this before it was true of the tree: the test called the helper directly and the call site was unguarded. It now runs through `send_proactive_summary_notification`, and deleting the call site fails with `covered peer A was not seeded ... the derivation did not happen, so #5190 is unfixed`.

Full suite: 4712 lib tests, 3 simulation tests, `cargo fmt --check` and both CI clippy invocations clean.

Fixes #5190
Refs #5147, #5186, #5189

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/operations/update.rs
+++ b/crates/core/src/operations/update.rs
@@ -3465,6 +3465,99 @@ mod tests {
     /// the weakest statement that still rules out vacuity. The strict form
     /// ("A specifically is a target") would fire first under a narrowing
     /// mutation and hide the fact that the subset property is what detects it.
+    /// #5190: a covered peer's summary is DERIVED, not requested.
+    ///
+    /// #5147 suppresses the fan-out leg to a covered peer, and that leg is how
+    /// the peer would have learned our summary; #4965 separately excludes
+    /// advertised co-hosts from the standalone notification. So it received
+    /// neither and stayed stale until the ~5-minute heartbeat.
+    ///
+    /// What this test must prove is BOTH halves: the model is populated, AND
+    /// no message was sent to populate it. Asserting only the first would pass
+    /// equally for the transmit-based fix that was measured at 429 extra
+    /// messages to save 13 sends.
+    ///
+    /// The read-back goes through `begin_peer_summary_broadcast`, the same
+    /// accessor the broadcast path uses, so this asserts the value is visible
+    /// to the code that actually consumes it rather than to a test-only peek.
+    #[tokio::test]
+    async fn covered_peer_summary_is_derived_without_sending_anything() {
+        use crate::ring::interest::PeerSummaryForBroadcast;
+
+        let (op_manager, mut rx, _guard) =
+            build_notification_test_node("notif-covered-inference-5190").await;
+        let key = crate::operations::test_utils::make_contract_key(12);
+        op_manager
+            .ring
+            .host_contract(key, 1024, crate::ring::AccessType::Put);
+
+        // A: advertised co-host the originator says it already delivered to.
+        // B: advertised co-host NOT on the list. B is the control — it stops
+        //    this passing under a mutation that seeds every peer regardless.
+        let (a_pk, _a_addr) = connect_peer(&op_manager, 19301, 0.31);
+        let (b_pk, _b_addr) = connect_peer(&op_manager, 19302, 0.32);
+        let (_d_pk, sender_addr) = connect_peer(&op_manager, 19304, 0.34);
+        advertise_cohost(&op_manager, &a_pk, &key);
+        advertise_cohost(&op_manager, &b_pk, &key);
+
+        let a_key = crate::ring::PeerKey::from(a_pk.clone());
+        let b_key = crate::ring::PeerKey::from(b_pk.clone());
+        let origin = BroadcastOrigin::relayed(
+            sender_addr,
+            [a_key.clone()].into_iter().collect::<HashSet<_>>(),
+        );
+
+        // Premise: A has no cached summary yet, so a Known result below is
+        // attributable to this call rather than to prior state.
+        assert!(
+            matches!(
+                op_manager
+                    .interest_manager
+                    .begin_peer_summary_broadcast(&key, &a_key),
+                PeerSummaryForBroadcast::Missing { .. }
+            ),
+            "premise: A must start with no cached summary"
+        );
+
+        let our_summary = StateSummary::from(vec![7u8, 7, 7]);
+        let seeded = seed_covered_peer_summaries(&op_manager, &key, &origin, &our_summary);
+        assert_eq!(seeded, 1, "exactly the one covered peer should be seeded");
+
+        match op_manager
+            .interest_manager
+            .begin_peer_summary_broadcast(&key, &a_key)
+        {
+            PeerSummaryForBroadcast::Known(got) => assert_eq!(
+                &got[..],
+                &our_summary[..],
+                "#5190: the covered peer's summary must be derived from our own \
+                 post-merge summary — it applied the same update from the same \
+                 originator, so its summary equals ours"
+            ),
+            PeerSummaryForBroadcast::Missing { reason, .. } => panic!(
+                "covered peer A was not seeded (missing reason {reason:?}); the \
+                 derivation did not happen, so #5190 is unfixed"
+            ),
+        }
+
+        assert!(
+            matches!(
+                op_manager
+                    .interest_manager
+                    .begin_peer_summary_broadcast(&key, &b_key),
+                PeerSummaryForBroadcast::Missing { .. }
+            ),
+            "B was NOT on the originator's list, so nothing is known about what \
+             it holds and nothing may be assumed"
+        );
+
+        // The whole point: none of that cost a message.
+        assert!(
+            drain_interest_targets(&mut rx).is_empty(),
+            "deriving the summary must cost ZERO messages"
+        );
+    }
+
     #[tokio::test]
     async fn notification_exclusion_is_a_subset_of_broadcast_targets() {
         let (op_manager, mut rx, _guard) =
--- a/crates/core/src/operations/update.rs
+++ b/crates/core/src/operations/update.rs
@@ -1069,10 +1069,16 @@ pub(crate) async fn update_contract(
 /// — the fan-out and heartbeat already behaved this way — and this gate makes
 /// the third path match them rather than introducing it. Tracked with the rest
 /// of the #4440 advertisement-hygiene work; deliberately not papered over here.
+///
+/// `origin` MUST be the one this fan-out registered. It is not used to choose
+/// recipients — the #4965 co-host exclusion below is unchanged — but to DERIVE
+/// the covered peers' summaries without sending them anything. See
+/// [`seed_covered_peer_summaries`].
 pub(crate) async fn send_proactive_summary_notification(
     op_manager: &OpManager,
     key: &ContractKey,
     sender_addr: SocketAddr,
+    origin: &BroadcastOrigin,
 ) {
     use crate::message::{SummariesEmitter, SummaryEntry};
     use crate::ring::interest::contract_hash;
@@ -1142,6 +1148,13 @@ pub(crate) async fn send_proactive_summary_notification(
     // reply in the rollup.
     let hash = contract_hash(key);
     let full_entry = SummaryEntry::from_summary(hash, Some(&summary));
+
+    // #5190, and the reason this function takes an `origin` at all.
+    //
+    // Derive the covered peers' summaries rather than mailing ours to them.
+    // This costs nothing and is what makes the co-host exclusion below sound
+    // again for peers the #5147 target list suppressed.
+    seed_covered_peer_summaries(op_manager, key, origin, &summary);
 
     // Get interested peers and resolve them to addresses. Peers that don't
     // resolve to a live socket address are dropped here, as before.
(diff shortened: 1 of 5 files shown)


sanity:
## Correction, and holding this PR

Three independent reviewers found the same BLOCKING issue and they are right. Two things need saying plainly before anything else.

### 1. A claim in the PR description was false

The description said the regression test "now runs through `send_proactive_summary_notification`" and that I had mutation-verified it. **That was not what was in the tree.** The shipped test called `seed_covered_peer_summaries` directly, so deleting the production call site left the whole suite green.

What happened: I did write the entry-point version and did verify the mutation against it. Then I ran `git checkout -- crates/core/src/operations/update.rs` to revert the mutation, which also reverted the still-uncommitted rewrite. I then wrote the description from what I had verified rather than from what was committed. My own notes warn against exactly that `git checkout` pattern; I walked into it anyway.

Fixed in `c2d11cbb7`, committed **before** mutating this time. Actual output with the call site deleted:

```
covered_peer_summary_is_derived_without_sending_anything ... FAILED
covered peer A was not seeded (missing reason None); the derivation did not happen, so #5190 is unfixed
```

Reviewers were also right that the "zero messages" assertion was vacuous as shipped: it asserted no sends from a function containing no send path. Through the entry point it now has something to say, though see below for why I am not claiming it is sufficient.

### 2. The inference's first premise is false, and that is a design problem

`target_list_for`'s own rustdoc (`broadcast_queue.rs:3384-3407`) says the covered list describes who we **ENQUEUED** to, not who received bytes, and enumerates queue eviction at capacity, send failure, `should_broadcast_contract` false, and unresolvable address — noting these **cluster under congestion**.

So "on the covered list ⇒ received and applied this update" is wrong, and this PR makes the consequence worse rather than better:

- **#5147 alone:** a wrong belief costs one suppressed leg, re-evaluated from scratch next update.
- **This PR:** the wrong belief is written into `interested_peers[key][peer].summary` as durable local state, where `plan_fanout_send` byte-compares it and **Skips**. Because we then send nothing, no delta fails to apply, so no ResyncRequest fires. The delta-apply-failure corrector my rustdoc leans on **cannot fire in the Skip case**. Only the ~5-minute heartbeat corrects it.

That inverts the safety direction: a **missing** summary makes us send full state (expensive, safe); a **wrong** one makes us skip (cheap, unsafe). Trading the first for the second is not obviously an improvement, and my "same trust model as #4442" defence does not hold — #4442 caches the summary of state it actually attempted to deliver, on its own send path. This caches a third party's pre-dispatch claim.

Related, independently confirmed: the seeded summary is fetched by a **fresh `GetSummaryQuery` inside the spawned task**, so it is not pinned to this transaction's merge. On a hot contract a second update can commit in that window and we attribute its summary to the earlier update's covered list. The 100ms throttle makes this the *dominant* pairing under bursts rather than a rare race.

### Where that leaves it

The mechanism is sound, the security bound is real (`covered.resolve` filters against our own advertised co-host list under tx-seeded hashes, so a malicious originator cannot inject arbitrary peers), and the traffic measurement stands: 15 fewer sends and 22 fewer redundant deliveries at zero added messages on the multi-writer arm.

But it is not mergeable on a premise the codebase's own documentation contradicts. The candidate fixes, in increasing cost:

1. **Do not let a derived summary cause a `Skip`** — use it only as a delta base, so a wrong derivation costs one unappliable delta plus a resync instead of silent starvation. This preserves most of the saving and restores the safe failure direction. Needs a trust tier on the cached summary, plumbed into `plan_fanout_send`.
2. Seed from the driver's `updated_value` rather than re-querying, closing the summary/transaction pairing gap.
3. Gate on **#5186** measuring over-suppression first, which this PR does not close and which the derived population makes more load-bearing.

I am not attempting (1) tonight: it is a protocol-core change to the broadcast gate, and this project's guard rightly blocks that after 9 PM. Converting to draft.

Remaining review findings (seed count re-derived rather than produced by the decision; the seed running *before* the recipient set is read so it can add notification targets; the misplaced rustdoc block; the per-peer summary clone on large-summary contracts; missing index/label guard on the widened enum) are all valid and tracked, but secondary to the two above.

**The measurement counter that made this whole comparison possible is split out as #5202 and is independently mergeable.**

[AI-assisted - Claude]

sanity:
## Backlog review — staying a draft, and why

Picked up during a PR-backlog sweep. **This should stay a draft.** Two findings, one of which corrects an assumption about this PR that was circulating elsewhere.

### 1. The hold is correct, and I verified the premise defect independently

The approach rests on: *"Being on the originator's covered list **means** that peer received and applied this same update from the same originator."*

`target_list_for`'s own rustdoc (`broadcast_queue.rs`) says the opposite, in terms:

> The list describes who we ENQUEUED to, which is not identical to who receives bytes.

and then enumerates four gaps it explicitly calls **not benign** — queue eviction at `max_queue_depth`, `bridge.send` failure or a stream dropped before `Delivered`, `should_broadcast_contract` returning false, and `fanout_send_needed` skipping on a cached summary. Three of the four **cluster under congestion**, which is exactly when fan-out suppression matters most. The rustdoc also states the bound: "None of these are corrected for here... The bound on all of them is the interest heartbeat."

So the derivation is unsound precisely in the regime it targets, and the consequence is worse than the status quo rather than equal to it. Today a wrongly-cached summary costs one leg and is corrected by the next comparison. This PR would take a belief that is wrong for a whole *population* of peers at once, write it into durable local state via `upsert_peer_summary_from`, and turn it into a silent `Skip` that heals only on the ~5-minute heartbeat. That is a different risk class from the #4442 optimistic cache it is compared to, not another instance of it.

The traffic measurement stands and is worth keeping: **15 fewer sends and 22 fewer redundant deliveries at zero added messages** in the multi-writer arm, plus the negative result that the obvious notify-based fix costs ~250 messages to save 13 sends. That is real work and the reason not to close this.

**What would unblock it:** a soundness argument that survives the enqueue-vs-delivered gap. Either derive only from legs actually confirmed `Delivered` (which the message cannot wait for, per the rustdoc — so it likely means deriving on a later cycle rather than inline), or keep the derivation but make it self-correcting so a wrong derived summary cannot cause a durable skip. Absent one of those, the measured saving is not worth the divergence it buys.

### 2. The "orphan commit" is not orphaned — do not cherry-pick it

A commit `b1d7747a7` (*"test(update): parameterise the #5147 arm and record what it cannot see"*) was found off this branch and preserved at `origin/rescue/5203-b1d7747a7`, on the assumption it was unmerged work at risk of being lost.

It is **already on `main`**, as `28b4fa0dd` — *"test(update): parameterise the #5147 sim arm and record what it cannot see (#5216)"* — merged via #5216. `ArmTopology` and `explore_5147_sparse_full_state_sensitivity` are both present on `main` today.

`b1d7747a7` is a **superseded pre-merge revision**, not the landed one. Diffing it against `main` shows applying it would **remove 56 lines and re-add 31 older ones**, dropping (among other things) the note recording that the null result was replicated on clean main with no #5190 partition in tree. Cherry-picking it onto this branch would regress the file.

Nothing needs rescuing here. The rescue ref can be deleted once someone confirms this independently.

### Not done

I did not push a rebase. This branch is BEHIND, but it is a held draft that is not landing on its current premise, and a rebase would consume a full CI run for no decision. Whoever resumes it should rebase then.

Lens: single adversarial read plus independent verification of the two factual claims above against the code and against `main`. No full multi-lens review — the design premise fails before the implementation is worth reviewing.

[AI-assisted - Claude]

PR #3648, opened 2026-03-25
feat: ipv6 dual-stack support for all sockets

sanity:
## Problem

Freenet defaults to binding on `0.0.0.0` (IPv4-only), which means nodes on IPv6-only networks (increasingly common on mobile carriers) cannot connect. Users have reported connectivity issues that trace back to this limitation. Additionally, running separate IPv4 and IPv6 sockets isn't supported — there's no way to serve both address families from a single node instance.

## Approach

Change the default bind address from `0.0.0.0` to `::` (IPv6 unspecified) and set `IPV6_V6ONLY=false` on all IPv6 sockets. This creates **dual-stack sockets** that accept both IPv4 (via IPv4-mapped addresses `::ffff:x.x.x.x`) and IPv6 connections on a single socket. This is the standard approach used by most modern network services.

Key changes:
- **Default bind address** → `::` instead of `0.0.0.0` (backward compatible: existing configs with `"0.0.0.0"` deserialize as `IpAddr::V4` and work as before)
- **UDP transport** (`Socket::bind` for `UdpSocket`) → uses `socket2` to create the socket with `set_only_v6(false)` before binding, then converts to tokio `UdpSocket`
- **TCP WebSocket listener** (`server.rs`) → sets `set_only_v6(false)` on IPv6 sockets
- **`is_private_ip`** → handles IPv4-mapped IPv6 addresses (`::ffff:127.0.0.1` etc.) that appear when IPv4 clients connect to dual-stack sockets. Without this, local IPv4 clients would be rejected as non-private.
- **Default local address** → `::1` instead of `127.0.0.1` for local mode

Cross-platform: `set_only_v6(false)` is the default on Linux but must be explicitly set on macOS, Windows, and some BSDs. Using `socket2` ensures consistent behavior across all platforms.

## Testing

- Added `udp_dual_stack_accepts_ipv4` test: binds to `[::]:0`, sends from `127.0.0.1`, verifies receipt
- Added `udp_dual_stack_accepts_ipv6` test: binds to `[::]:0`, sends from `[::1]`, verifies receipt
- Added IPv4-mapped address tests for `is_private_ip`: `::ffff:127.0.0.1` (private), `::ffff:10.0.0.1` (private), `::ffff:8.8.8.8` (public)
- All 2098 existing tests pass

[AI-assisted - Claude]

PR #5210, opened 2026-08-06
fix(config): default the client API to loopback in both operation modes

sanity:
## Problem

The node's client API — the local HTTP/WebSocket control surface — took its
default bind address from `OperationMode`: loopback in local mode, all
interfaces (`::`) in network mode.

That coupling is wrong. `OperationMode` is a statement about the overlay:
whether this node participates in the ring. It is not a statement about who may
drive the node's control API, which is a separate and much more privileged
surface. A default install of a network peer therefore listened on every
interface without the operator ever asking for it.

The repository documents no intent behind that default. The only prose on the
subject is the rustdoc on `--allowed-source-cidrs`, which is a warning attached
to the opt-in flag for *extending* reach.

Refs: private advisory GHSA-824h-7x5x-wfmf.

Scope note for CONTRIBUTING: this is a user-visible behavior change, which
normally needs an approved public issue first. It is tracked in the private
advisory instead, at Ian's direction — a public issue is the wrong vehicle here.

## Approach

**1. Loopback by default in both modes.** The no-flags branch of
`resolve_ws_api_address` is mode-independent. `OperationMode` keeps governing
ring participation and `secrets_dir(mode)`, and it still scopes the
compatibility auto-widen in point 2; what it no longer does is decide the
*default* bind of the client API. Loopback here is
`::1`, which also binds `127.0.0.1` through the existing companion-bind
mechanism (#4330), so `ws://127.0.0.1:7509` and `ws://[::1]:7509` both keep
working unchanged.

Three places baked the wildcard in, and all three had to move for the default to
actually be loopback: the mode-keyed resolution in `build()`,
`WebsocketApiConfig::default()`, and `ConfigArgs::default()` — which pinned
`Some(default_listening_address())`, i.e. an *explicit* address that takes
precedence over any default, so changing only the first two left every
programmatic composition on the wildcard. That third one is what made the first
run of the new tests fail, which is the argument for having written them.

**2. Auto-widen for backward compatibility — `--allowed-source-cidrs` only,
network mode only, this-boot flags only.** Before this change a network-mode
node bound `::` by default, so an operator who set `--allowed-source-cidrs` and
nothing else was relying on that wide default; the flag is genuinely inert on a
loopback socket, since a non-private source can never reach `::1`, so setting it
is unambiguous intent to serve non-local clients. If it is set, `--ws-api-address`
is not, and the mode is Network, the bind widens and says so at INFO.

Three scoping rules, each of which was wrong in an earlier cut of this PR:

- **`--allowed-host` is NOT a trigger.** It is a Host-header allowlist that is
  fully functional on loopback, the same-host reverse proxy is its primary
  documented use, and that is the *only* shape in which hosted mode's
  `userToken` is honoured at all (`decide_user_token` requires a loopback
  source). Widening for it would bind every interface in the commonest proxy
  deployment for no functional gain. A proxy on a different host is a deliberate
  multi-machine choice whose operator passes `--ws-api-address`.
- **Network mode only.** Local mode has always bound loopback, so there is no
  wide default to preserve and widening there would mean this hardening *opened*
  a socket that used to be closed.
- **This-boot grants only.** The widen decision reads the CLI/env value captured
  *before* the `config.toml` merge. Reading the merged value would make the
  widen sticky: one boot with the flag would pin the node to the wildcard
  forever, and removing the flag would never narrow it again — the hardening
  would permanently miss every node that ever set an allow-list.

**3. Upgrade migration — what makes this reach existing installs.** `build()`
persists the *resolved* config, so every node that has ever booted in network
mode already carries a wildcard `ws-api-address` in its `config.toml`. Merging
that value back would pin the existing fleet to the old bind and leave the
change reaching fresh installs only.

The sentinel is "a value this code could itself have written"
(`is_auto_derivable_ws_api_address`): the two wildcards (`0.0.0.0` before #3648,
`::` after) **and the two loopbacks**. Including loopback is load-bearing, not
tidiness: the migration boot persists a loopback address, and reading that back
as an operator choice on the next boot would take the Explicit short-circuit and
permanently disable the auto-widen remedy this PR's own release note points at.
The sentinel enumerates the four exactly rather than calling `is_loopback()`, so
a deliberately-pinned loopback alias like `127.0.0.5` stays an operator choice.
Any other persisted value — a specific interface IP — is preserved untouched.
CLI/env values are parsed before the merge, so `--ws-api-address ::` still wins.

**Re-derivation preserves the address family, and this is the safety-critical
part.** The primary bind is fatal while only the companion is best-effort
(`serve_dual_stack`). A node installed before #3648 carries `0.0.0.0`; on a host
with IPv6 disabled, re-deriving that to `::1` fails the bind with EAFNOSUPPORT —
and `build()` rewrites `config.toml` *before* the bind is attempted, while
`rollback.rs` does not snapshot config. So the crash-loop rollback restores the
previous binary onto a config that now names an unbindable family, it dies
identically, and rollback does not fire twice: the node is down permanently.
That is `code-style.md`'s "the brick-safety mechanism becomes the brick",
reached through a *value* change rather than the key change the existing rule
guards. Persisted `0.0.0.0` therefore becomes `127.0.0.1`, and an IPv4-family
node widens to `0.0.0.0` rather than `::`.

**4. Startup messages.** A re-derivation that MOVED the bind is reported in
whichever direction it moved, both at `warn!`. Narrowing is the one notice an
operator gets that clients on other machines just lost access. Widening matters
just as much and is easy to overlook: a node whose `config.toml` named a
loopback address, with `--allowed-source-cidrs` in its unit, now binds the
wildcard — because loopback is a value this code writes itself and so cannot be
told from a deliberate pin. On `main` that node stayed loopback. It is symmetric
with the pre-existing `::1` case and unavoidable given family preservation, but a
hardening PR that widens a node someone pinned has to say so loudly rather than
at `info!`.

Then the exposure warning, with two triggers:

- a **non-loopback bind** warns regardless of hosted mode. Hosted mode does not
  make a wide bind safe: `decide_user_token` returns the shared single-user
  context whenever a connection *omits* `userToken`, so hosted mode adds a
  per-user namespace for well-behaved clients without removing the shared one.
- **`--allowed-host` on a loopback bind with hosted mode off** warns
  unconditionally. A reverse proxy terminates the connection itself, so every
  visitor arrives wearing the proxy's source address and the node's own
  source-IP filters cannot tell them apart. Not gated on anything, because with
  hosted mode off there is no isolation at all.
- **loopback + hosted mode** — try.freenet.org's shape — stays quiet, and that
  is a **known gap, not a safety claim**. An untokened connection reads the
  shared namespace there too, and behind a public proxy that is any visitor;
  containment today is that the shared namespace is empty (measured on the live
  hosted node: zero files outside `*/users/*` across 28 delegates), which is a
  fact about current state rather than a structural guarantee. Making this
  branch conditional on that requires probing the secrets tree, which is its own
  change with its own failure modes — a probe that is wrong in either direction
  is worse than no warning, since one that is too eager fires on every boot of
  the flagship and trains operators to ignore the only signal there is. Tracked
  separately rather than bolted onto a default-hardening PR.

The messages are emitted from `Config::log_client_api_exposure()`, called from
`main` **after** `set_logger`. They cannot live in `ConfigArgs::build()`: it runs
before the global tracing subscriber is installed, so anything emitted there is
silently dropped — and this log is the only place an operator learns why their
LAN clients stopped connecting. `build()` records the decision on a runtime-only
`WebsocketApiConfig::exposure` field and `main` replays it.

Note this warning is the one behavior here that is *not* a bug fix in the
strict sense — it is a diagnostic on an existing configuration shape. It is in
this PR rather than a follow-up because the default change is what makes the
remaining exposed shapes worth naming.

## Release note

**Who is affected:** anyone who reaches a self-hosted node's UI or API **from a
different machine** — a phone or laptop browsing the node's dashboard, or
`fdev` / `riverctl` pointed at another host. After this change the node answers
on loopback only unless told otherwise, so those connections are refused. Access
from the machine running the node (`localhost` / `127.0.0.1` / `::1`) is
unchanged.

This applies **on upgrade**, not only to fresh installs: a wildcard
`ws-api-address` that was previously written into `config.toml` automatically is
re-derived to loopback. A persisted `127.0.0.1` or specific interface address is
left alone.

**Remedy — one flag.** Pick the narrowest that fits:

| Want | Add |
| --- | --- |
| Serve clients on other machines | `--ws-api-address ::` (or `WS_API_ADDRESS=::`) |
| One interface only | `--ws-api-address 192.0.2.10` |
| A VPN overlay (Tailscale/WireGuard) | `--allowed-source-cidrs 100.64.0.0/10` — widens the bind on its own, network mode |
| Behind a reverse proxy on another host | `--ws-api-address ::` plus `--allowed-host your.domain` |

**The flag has to live in the node's invocation and stay there** (the systemd
unit, docker-compose `environment:`, whatever launches it) — running it once is
not enough, and neither is leaving it in `config.toml`. Both the address and the
CIDR grant are re-decided from the invocation on every boot, precisely so that
removing a flag narrows the socket again. A specific address (`192.0.2.10`) is
distinguishable from this code's own output and does persist.

**On a systemd install, use `systemctl edit` to add a drop-in** overriding
`ExecStart`. Do not hand-edit the generated unit: that trips the installer's
sidecar hash, permanently marks the unit `UserModified`, and silently opts the
node out of every future unit-template change (including future hardening).

**Three environment variables were renamed.** `WS_API_ADDRESS`, `ALLOWED_HOST`
and `ALLOWED_SOURCE_CIDRS` are now `FREENET_WS_API_ADDRESS`,
`FREENET_ALLOWED_HOST` and `FREENET_ALLOWED_SOURCE_CIDRS`. Unnamespaced names
were survivable while a network-mode node bound `::` anyway and these only
shaped which `Host` headers it accepted. They are not now that
`--ws-api-address` decides the socket directly at the highest precedence and
`--allowed-source-cidrs` decides whether it is wide — a stray value in a shared
container, CI runner or systemd environment could open the API. A leftover
old-style variable is reported at startup rather than silently ignored.

**Also in this PR, and a widening rather than a narrowing:** `--ws-api-address`,
`--allowed-host` and `--allowed-source-cidrs` placed *before* the subcommand
(`freenet --allowed-source-cidrs … network`) were previously discarded. They are
now honoured, so a node whose operator wrote them in that position starts
admitting the CGNAT sources they intended all along. That is the fix for a
silently-ignored flag, but it is a filter widening riding inside a hardening
change and is called out here rather than buried.

Accepted edge, following directly from that: an operator who hand-edited
`ws-api-address = "::"` into `config.toml`, with no flag and no allow-list, is
indistinguishable from the old auto-default and will be re-derived to loopback.

**Rollback note:** after one boot on this version, `config.toml` holds a
concrete loopback address. A crash-loop auto-rollback to the previous binary
reads that as an explicit choice, so the rolled-back node comes back
loopback-only — and the old binary has none of these messages, so nothing
explains why. No key spelling changes and every value the new binary writes
parses and binds on the old one, so there is no parse-brick; but the previous
*reachability* is not restored automatically. Re-add `--ws-api-address ::`.

**Operator documentation:** `docs/client-api-exposure.md`, covering the default,
the narrowest flag for each situation, why the flag has to stay in the
invocation, the systemd drop-in, the renamed environment variables, and what the
three startup messages mean.

## Testing

New config-resolution tests in `crates/core/src/config.rs`:

- loopback default in **both** operation modes;
- auto-widen on `--allowed-source-cidrs`, and on `--allowed-host`;
- **no** auto-widen when `--ws-api-address` is given (try.freenet.org's exact
  shape: explicit loopback + hosted mode + proxy host);
- empty allow-lists (`[]`, which every already-persisted `config.toml` carries)
  do not count as widening intent;
- re-derivation preserves the address family, and the auto-widen does too —
  both written first and confirmed RED against the pre-fix code, since the test
  they replace pinned the unsafe behaviour;
- `--allowed-host` alone does not widen, while its allowlist still applies;
- a persisted CIDR grant does not widen a later flag-less boot (the sticky-widen
  case), while a grant still in the invocation does;
- Local mode with a widening flag still binds loopback, at both the pure
  resolver and the full-`build()` level;
- the documented remedy still works on an ALREADY-MIGRATED node: persisted
  loopback plus a newly-added CIDR grant widens the bind (it silently would not
  if the sentinel treated loopback as an operator choice);
- a re-derivation that widens past a persisted loopback address is REPORTED
  rather than silent;
- the loss-of-access notice fires on the migration boot and is absent on the
  next one (the half of its filter that was previously unverified);
- blank allow-list entries (`ALLOWED_HOST=` with no value, which parses as
  `Some(vec![""])`) do not count as widening intent;
- the remedy still works on the *second* boot: upgrade → persisted `::1` → add
  `--allowed-source-cidrs` → widens (this fails if the sentinel omits loopback);
- `build()` records the exposure decision on the resolved `Config`, since that
  is the only route by which it can reach a log;
- the warning's trigger matrix, asserting the *reason* it reports, including
  that a wildcard bind warns even in hosted mode and that
  proxied-with-hosted-mode-off fires it;
- migration: persisted `::` and persisted `0.0.0.0` are both re-derived to
  loopback; persisted `127.0.0.1` and a specific interface IP survive; an
  upgrading Tailscale-style node gets its wildcard bind straight back from
  auto-widening with no operator action;
- a source pin asserting the path stays wired end to end, now reaching ACROSS
  files: `build()` resolves through the helper and records the decision but does
  not emit the warning itself (that would be dropped);
  `log_client_api_exposure` consults the predicate and warns; and
  `bin/freenet.rs` still calls both `log_client_api_exposure` and
  `merge_pre_subcommand_ws_api_args` (the latter from *both* subcommand arms).
  Each helper has exactly one call site in a different file, so without the
  cross-file scrape deleting one left every test green.

Every pin above was mutation-tested: the mutation was applied, the test confirmed
red, and the mutation reverted.

Also verified empirically against a node built from this branch, reading the
actual listening sockets from the kernel (`ss -ltnp`, filtered to the node's own
PID) rather than from precedence code — results in a comment below.

Docker NAT simulation was checked and is unaffected: `freenet-test-network`'s
Docker path passes `--ws-api-address 0.0.0.0` explicitly, and its local-backend
path connects over `127.0.0.1`, which the companion bind still serves. Every
integration test under `crates/core/tests/` sets `ws_api.address` explicitly.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -1058,6 +1058,12 @@ fn run_node(config_args: ConfigArgs) -> anyhow::Result<()> {
     rt.block_on(async move {
         let config = config_args.build().await?;
         freenet::config::set_logger(None, None, config.paths().log_dir());
+        // Same reason as the info! below: `ConfigArgs::build()` resolves the
+        // client API's bind address, but it runs with no subscriber installed,
+        // so it records the decision instead of logging it. Replay it here —
+        // this is the only place an operator learns that their node stopped
+        // answering clients on other machines, and why.
+        config.log_client_api_exposure();
         // The logger is needed before this info which is why it's here instead of above
         tracing::info!(
             max_blocking_threads,
@@ -1070,6 +1076,34 @@ fn run_node(config_args: ConfigArgs) -> anyhow::Result<()> {
     })?;
 
     Ok(())
+}
+
+/// Carry the client-API exposure flags across the subcommand boundary.
+///
+/// `--ws-api-address` / `--allowed-host` / `--allowed-source-cidrs` placed
+/// BEFORE the subcommand (`freenet --ws-api-address :: network`) land in the
+/// top-level `ConfigArgs`, which the `Network`/`Local` arms otherwise discard —
+/// the same trap `--disable-auto-update` hit in #4690.
+///
+/// This matters more since the client API started defaulting to loopback
+/// (GHSA-824h-7x5x-wfmf): these three flags are exactly how an operator asks
+/// for a wider bind, so silently dropping one leaves the node loopback-only
+/// while the operator believes they opted out. Previously the same typo was
+/// harmless, because network mode bound `::` either way.
+///
+/// The subcommand's own value wins; the top-level one only fills a gap.
+fn merge_pre_subcommand_ws_api_args(config: &mut ConfigArgs, top_level: &ConfigArgs) {
+    config.ws_api.address = config.ws_api.address.or(top_level.ws_api.address);
+    config.ws_api.allowed_host = config
+        .ws_api
+        .allowed_host
+        .take()
+        .or_else(|| top_level.ws_api.allowed_host.clone());
+    config.ws_api.allowed_source_cidrs = config
+        .ws_api
+        .allowed_source_cidrs
+        .take()
+        .or_else(|| top_level.ws_api.allowed_source_cidrs.clone());
 }
 
 fn freenet_main() -> anyhow::Result<()> {
@@ -1111,11 +1145,13 @@ fn freenet_main() -> anyhow::Result<()> {
             // opt-out leaves the from-source node auto-updating (the exact loop
             // we are preventing).
             config.disable_auto_update |= cli.config.disable_auto_update;
+            merge_pre_subcommand_ws_api_args(&mut config, &cli.config);
             run_node(config)
         }
         Some(Command::Local { mut config }) => {
             config.mode = Some(OperationMode::Local);
             config.disable_auto_update |= cli.config.disable_auto_update;
+            merge_pre_subcommand_ws_api_args(&mut config, &cli.config);
             run_node(config)
         }
         None => {
@@ -1340,6 +1376,82 @@ mod tests {
                 "flag must be honored in either position: {argv:?}"
             );
         }
+    }
+
+    /// Same trap, for the flags that now decide the client API's bind: placed
+    /// before the subcommand they land in the top-level `ConfigArgs`, which the
+    /// `Network`/`Local` arms discard. Dropping one leaves the node
+    /// loopback-only while the operator believes they widened it.
+    #[test]
+    fn ws_api_exposure_flags_are_honored_before_the_subcommand() {
+        use super::{Cli, Command, merge_pre_subcommand_ws_api_args};
+        use clap::Parser;
+        for argv in [
+            vec![
+                "freenet",
+                "--ws-api-address",
+                "::",
+                "--allowed-host",
+                "node.example",
+                "--allowed-source-cidrs",
+                "100.64.0.0/10",
+                "network",
+            ],
+            vec![
+                "freenet",
+                "network",
+                "--ws-api-address",
+                "::",
+                "--allowed-host",
+                "node.example",
+                "--allowed-source-cidrs",
+                "100.64.0.0/10",
+            ],
+        ] {
+            let cli = Cli::try_parse_from(&argv).expect("parse");
+            let mut config = match cli.command {
+                Some(Command::Network { config }) => config,
+                _ => panic!("expected network subcommand"),
+            };
+            merge_pre_subcommand_ws_api_args(&mut config, &cli.config);
+            assert_eq!(
+                config.ws_api.address,
+                Some("::".parse().unwrap()),
+                "--ws-api-address must survive in either position: {argv:?}"
+            );
+            assert_eq!(
+                config.ws_api.allowed_host.as_deref(),
+                Some(&["node.example".to_string()][..]),
+                "--allowed-host must survive in either position: {argv:?}"
+            );
+            assert_eq!(
+                config.ws_api.allowed_source_cidrs.as_deref(),
+                Some(&["100.64.0.0/10".to_string()][..]),
+                "--allowed-source-cidrs must survive in either position: {argv:?}"
+            );
+        }
+    }
+
+    /// The subcommand's own value wins; the top-level one only fills a gap.
+    #[test]
+    fn subcommand_ws_api_address_beats_the_pre_subcommand_one() {
+        use super::{Cli, Command, merge_pre_subcommand_ws_api_args};
+        use clap::Parser;
+        let cli = Cli::try_parse_from([
+            "freenet",
+            "--ws-api-address",
+            "::",
+            "network",
+            "--ws-api-address",
+            "127.0.0.1",
+        ])
+        .expect("parse");
+        let mut config = match cli.command {
+            Some(Command::Network { config }) => config,
+            _ => panic!("expected network subcommand"),
+        };
+        merge_pre_subcommand_ws_api_args(&mut config, &cli.config);
+        assert_eq!(config.ws_api.address, Some("127.0.0.1".parse().unwrap()));
     }
 
     #[test]
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1271,8 +1271,13 @@ impl ConfigArgs {
         // reports it after the subscriber exists. See `WsApiExposure`.
         let ws_api_exposure = WsApiExposure {
             source: ws_api_address_source,
+            // Report the discard ONLY when the bind actually got narrower.
+            // A node whose persisted `::1` is dropped and then auto-widened to
+            // `::` is getting MORE exposed, so "clients on other machines can
+            // no longer reach this node" would be flatly untrue — and would
+            // print directly above the auto-widen line saying the opposite.
             dropped_persisted_address: dropped_persisted_wildcard
-                .filter(|persisted| *persisted != ws_api_address),
+                .filter(|persisted| !persisted.is_loopback() && ws_api_address.is_loopback()),
         };
         let this = Config {
             mode,
@@ -9679,6 +9684,24 @@ shutdown-drain-secs = 42
         assert_eq!(
             migrated.ws_api.exposure.source,
             WsApiAddressSource::DefaultLoopback
+        );
+
+        // Boot 3 in the same dir: the operator adds a CIDR grant, so the
+        // persisted `::1` is dropped and immediately auto-widened back to `::`.
+        // The bind got WIDER, so the "clients on other machines can no longer
+        // reach this node" notice must NOT fire — it would print directly above
+        // the auto-widen line and say the opposite of it.
+        let mut widened_later = ws_api_test_args(OperationMode::Network, upgrade_dir.path());
+        widened_later.ws_api.allowed_source_cidrs = Some(vec!["100.64.0.0/10".to_string()]);
+        let cfg = widened_later
+            .build_with_gateways_index(&index_url)
+            .await
+            .expect("third build should succeed");
+        assert_eq!(cfg.ws_api.address, default_listening_address());
+        assert_eq!(cfg.ws_api.exposure.source, WsApiAddressSource::AutoWidened);
+        assert_eq!(
+            cfg.ws_api.exposure.dropped_persisted_address, None,
+            "a widening re-derivation must not report itself as a loss of access"
         );
     }
 
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -3119,7 +3119,7 @@ const fn default_local_address() -> IpAddr {
 /// See [`WsApiExposure`].
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub enum WsApiAddressSource {
-    /// The operator named an address (`--ws-api-address`, `WS_API_ADDRESS`, or
+    /// The operator named an address (`--ws-api-address`, `FREENET_WS_API_ADDRESS`, or
     /// a `ws-api-address` key in `config.toml` that this code could not itself
     /// have written). Used verbatim.
     ///
@@ -3151,8 +3151,14 @@ pub struct WsApiExposure {
     /// How the bind address was chosen.
     pub source: WsApiAddressSource,
     /// A `ws-api-address` this code could itself have auto-written, discarded
-    /// from `config.toml` during the merge so it could be re-derived. `Some`
-    /// only when the operator supplied no address by flag or env.
+    /// from `config.toml` during the merge so it could be re-derived.
+    ///
+    /// `Some` only when the operator supplied no address by flag or env AND the
+    /// re-derivation actually MOVED the bind. A value that re-derives to itself
+    /// — the steady state from the second boot onward — is not recorded, because
+    /// re-announcing an unchanged bind every boot is how a log line gets tuned
+    /// out. Which direction it moved selects the message; see
+    /// `Config::log_client_api_exposure`.
     pub dropped_persisted_address: Option<IpAddr>,
 }
 
@@ -3302,7 +3308,8 @@ fn resolve_ws_api_address(
 /// current state rather than a structural guarantee. Making this branch fire
 /// only when the shared namespace actually holds something needs a probe of the
 /// secrets tree, which is a separate change with its own failure modes — it is
-/// tracked separately rather than bolted onto a default-hardening PR, because a
+/// tracked in advisory GHSA-824h-7x5x-wfmf §8, with the measured tree shape and
+/// the requirements, rather than bolted onto a default-hardening PR, because a
 /// probe that is wrong in either direction is worse than no warning: too eager
 /// and it fires on every boot of the flagship and trains operators to ignore the
 /// one signal there is.
@@ -9495,7 +9502,7 @@ shutdown-drain-secs = 42
         }
     }
 
-    /// `ALLOWED_HOST=` declared with no value — routine in a docker-compose
+    /// `FREENET_ALLOWED_HOST=` declared with no value — routine in a docker-compose
     /// `.env`, a k8s ConfigMap, or a systemd `Environment=` line — parses as
     /// `Some(vec![""])`. Reading that as intent would widen the bind on a node
     /// whose operator granted nothing.
@@ -9588,7 +9595,9 @@ shutdown-drain-secs = 42
         }
     }
 
-    /// The exposure warning fires on hosted-mode-off AND (non-loopback bind OR
+    /// The exposure warning fires on a non-loopback bind REGARDLESS of hosted
+    /// mode (an untokened connection reaches the shared namespace either way),
+    /// and additionally on (non-loopback bind OR
     /// an `--allowed-host` reverse-proxy grant). Both triggers matter: a proxy
     /// terminates the connection itself, so every visitor looks local to the
     /// node's own source-IP filters even when the bind is `127.0.0.1`.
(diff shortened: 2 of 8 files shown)


sanity:
## Empirical verification (not a review — self-check)

Ran a node built from this branch per scenario and read the **actual listening
sockets from the kernel** (`ss -ltnp`, filtered to that node's own PID) plus its
own log, rather than reasoning from the config-precedence code.

| Invocation | Bound sockets | auto-widen INFO | exposure WARN |
| --- | --- | --- | --- |
| `freenet network` (no flags) | `127.0.0.1`, `[::1]` | 0 | 0 |
| `network --hosted-mode --ws-api-address 127.0.0.1 --allowed-host try.freenet.org` (try.freenet.org's exact shape) | `127.0.0.1`, `[::1]` | 0 | **0** |
| `network --ws-api-address 127.0.0.1 --allowed-host node.example.org` (proxied, hosted **off**) | `127.0.0.1`, `[::1]` | 0 | 1 |
| `network --allowed-source-cidrs 100.64.0.0/10` (Tailscale shape) | `0.0.0.0`, `[::]` | 1 | 1 |
| `local --allowed-host proxy.example --allowed-source-cidrs 100.64.0.0/10` | `127.0.0.1`, `[::1]` | 0 | 1 |

So: **try.freenet.org is unaffected** — it passes `--ws-api-address 127.0.0.1`
explicitly (`/etc/systemd/system/freenet-hosted.service`), takes the Explicit
branch, does not auto-widen, and does not trip the warning. The genuinely
dangerous shape (proxied with hosted mode **off**) does warn, and names the
proxy as the reason rather than the bind.

The nova gateway and the ten local test peers are also unaffected: their
`config.toml` files already carry `ws-api-address = "127.0.0.1"`, which is not
an auto-derivable value and so is preserved by the migration.

### Upgrade path, end to end

```
boot 1 (pre-upgrade shape)   ws-api-address = "::"     persisted
boot 2 (this branch, no flag) sockets: 127.0.0.1, [::1]
                              ws-api-address = "::1"    persisted
                              INFO "The client API now defaults to loopback, so the
                              `ws-api-address` previously auto-written to config.toml
                              has been re-derived. Clients on OTHER machines can no
                              longer reach this node's API. Add --ws-api-address :: …"
```

### Docker NAT simulation

Unaffected. `freenet-test-network`'s Docker path passes `--ws-api-address 0.0.0.0`
explicitly for both gateway and peer containers; its local-backend path connects
over `127.0.0.1`, which the `::1` bind still serves through the #4330 companion
bind (pinned by `client_api_reachable_on_both_ipv4_and_ipv6_loopback`). Every
integration test under `crates/core/tests/` and the `#[freenet_test]` codegen set
`ws_api.address` explicitly.

`docker/freenet-node/docker-compose.yml` is the one in-repo deployment that
relied on the implicit wide default (it uses `network_mode: host`); it gains a
commented `WS_API_ADDRESS=::` with an explanation. No CI job exercises it.

[AI-assisted - Claude]

sanity:
## CI red is a GitHub Actions outage, not this change

Every failing job on all three pushes fails identically, before running any
project code:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]Failed to resolve action download info.
```

That includes `Fmt`, `Rule Lint` and `Conventional Commits`, which then cancel
the downstream jobs. githubstatus.com currently reports **Actions:
partial_outage**. Nothing here has produced a genuine CI failure yet, so this PR
has **no verified green CI** and must not be merged until a run completes
cleanly on the current head.

Locally, on `a7d8811`:

- `cargo test -p freenet --lib config::` — 154 passed, 0 failed
- `cargo test -p freenet --bin freenet ws_api` — 2 passed
- full `cargo test -p freenet --lib` — 4730 passed, 1 failed:
  `wasm_runtime::tests::contract_metering::test_timeout_metering` ("Took 7.35s,
  should timeout before 7s"). It passes in isolation at 5.43s of its 7s budget
  and failed only under three concurrent cargo builds on the same box; it is a
  wall-clock assertion and touches nothing this PR modifies. Flagging rather
  than dismissing — if it recurs on idle CI it is a real problem in that test's
  timing margin, independent of this change.
- `cargo fmt --check` clean; `cargo clippy -p freenet -- -D warnings` clean.

Needs a re-run once Actions recovers.

[AI-assisted - Claude]

sanity:
## Review round addressed (`c17d0b5`)

**B1 (blocking) — fixed, and it was a brick.** Re-derivation was family-blind:
`0.0.0.0` was auto-derivable but always re-derived to `::1`. On an IPv6-disabled
host the primary bind fails with EAFNOSUPPORT, and the primary bind is fatal
while only the companion is best-effort. Because `build()` rewrites `config.toml`
before the bind and `rollback.rs` does not snapshot config, the crash-loop
rollback restores the old binary onto a config naming an unbindable family, it
dies the same way, and rollback does not fire twice. Re-derivation now stays
within the persisted family.

Following the theme note, both new family tests were written first and confirmed
**red** against the pre-fix code:

```
re_derivation_preserves_the_address_family   left: ::1   right: 127.0.0.1
auto_widen_preserves_the_address_family      left: ::    right: 0.0.0.0
```

The test that pinned the unsafe behaviour is gone.

**Also fixed:** I1 (grant captured before the merge, so the widen is no longer
sticky), I2 (`FREENET_ALLOWED_*`, with a startup warning naming any leftover
old-style variable so the rename is not silent), I3 (`--allowed-host` is no
longer a widening trigger), I4 (`systemctl edit` drop-in, not hand-editing the
unit), I5 (`warn!`), I7 (cross-file pin over `bin/freenet.rs`, asserting *both*
subcommand arms), I8, M1, M3, M4, M5.

### I6 — I chose (a), the accurate warning

Loopback + hosted mode now warns **only when the shared delegate-secret
namespace actually holds something**. I re-measured on the live hosted node
before implementing it and got the same result you did:

```
$ find /home/ian/freenet-hosted/data/secrets -mindepth 2 -type f -not -path "*/users/*" | wc -l
0
```

The probe stops at the first file it finds, skips `users/` components, ignores
node key material at depth 1, caps its own recursion depth, and returns
"occupied" if it cannot read the tree — so a permissions bug cannot silence it.
It was cheap enough that (b) was not needed. try.freenet.org stays quiet today
and starts warning the moment that stops being true.

**Not gated on occupancy:** the hosted-mode-off proxy case. With no isolation at
all the namespace fills the instant anyone uses it, so a point-in-time emptiness
check would be meaningless there.

### Skipped, with reasons

- **M2** — not skipped, but it is a filter *widening*, so it is now called out
  explicitly in the release note rather than left in the diff.
- **M6** (`config_dir` not mode-split, so alternating `freenet local` and
  `freenet network --ws-api-address ::` in one directory flaps the persisted
  value) — real, but pre-existing and not caused by this change; every write is
  correct for its own invocation. Fixing it means mode-splitting `config_dir`,
  which is a much larger blast radius than this PR should carry. Left alone
  deliberately rather than half-fixed.

### Re-verified empirically against a rebuilt binary

| Invocation | Bound sockets | Note |
| --- | --- | --- |
| `network --allowed-host node.example.org` | `127.0.0.1`, `[::1]` | **I3**: no longer widens |
| `network` with `0.0.0.0` persisted, flagless | `127.0.0.1`, `[::1]` | **B1**: re-derives `0.0.0.0` → `127.0.0.1`, stays IPv4-bindable |
| `network --allowed-source-cidrs 100.64.0.0/10` | `0.0.0.0`, `[::]` | widens, both log lines |
| try.freenet.org's exact invocation | `127.0.0.1`, `[::1]` | silent |
| `local` with both allow-lists | `127.0.0.1`, `[::1]` | never widens |

Local: 157 config tests, 365 bin tests, `cargo fmt --check` and
`cargo clippy -p freenet -- -D warnings` clean.

**CI is still not green and I am not claiming otherwise.** GitHub Actions has
escalated from partial to **major_outage**; every job on the previous head failed
with `Failed to resolve action download info`, and even the job logs return
`BlobNotFound`. This needs a genuinely clean run before any merge decision.

[AI-assisted - Claude]

sanity:
## Occupancy probe split out; D1–D5 fixed (`3236174`)

**Probe removed from this PR.** Agreed on both counts, and P1 was worse than
described. I measured the live hosted tree before touching anything:

```
<secrets_dir>/delegate_cipher                        node key material (depth 1)
<secrets_dir>/users/<user_id>/<file>                 12,396 files  ← the killer
<secrets_dir>/<delegate_key>/users/<user_id>/<file>  50,324 files
files outside every users/ component:                0
```

There is a `users/` directory at **two** levels, and my fixture had only the
deeper one. The probe would have reported "occupied" on try.freenet.org on every
boot — the exact desensitisation the design existed to prevent, which is a
sharper version of your reflection: the fixture encoded what I expected the tree
to look like, so it could only confirm what I already believed.

The hosted-plus-loopback branch returns to **quiet**, and the predicate now
documents that as a *known gap*, not a safety claim, with the reason the probe
is deferred. P1–P9 carry to the follow-up PR, and its fixture gets built from the
shape above.

**D1** — `--ws-api-address`'s clap help still described the pre-reversal
behaviour, claiming `--allowed-host` widens the bind "since neither flag does
anything on a loopback listener". Both halves false, on `freenet network --help`.
Rewritten, and it now says a same-host proxy needs no address flag while an
off-host one does.

**D2** — you're right that this needed to be deliberate. A re-derivation that
widens is now reported at `warn!` with its own message naming what happened:
config said loopback, that value is one this code writes itself so it cannot be
told from a pin, and the CIDR grant widened past it. The previous round had
*suppressed* this line to avoid contradicting the auto-widen message; the fix is
to make it directional, not silent.

**D3** — `FREENET_WS_API_ADDRESS`, with the same legacy-name warning. It was the
one I left, and it is the one with the most direct effect.

**D4** — `remedy_still_works_after_the_upgrade_persisted_loopback` restored, and
the stale rustdoc cross-reference fixed. You were right that the deletion wasn't
tracked; I replaced it with a test of a different property and did not notice the
gap.

**D5** — `docs/client-api-exposure.md`.

**Minors** — gateway compose comment corrected (it is the *host's* loopback under
`network_mode: host`), the `127.0.0.1` assertion that passed for the wrong reason
replaced with `127.0.0.5` plus an explanation, and the rollback note now says the
old binary has no message to explain the narrowing.

Local on `3236174`: 156 config tests, 365 bin tests, fmt and clippy clean. CI
still needs a genuinely green run — Actions has been in outage all afternoon and
every job on the previous heads failed on `Failed to resolve action download
info`, with job logs returning `BlobNotFound`.

[AI-assisted - Claude]

sanity:
## G1–G6 fixed (`ff06b69`)

**G1** — fair hit, and worth naming for what it is: the document I wrote to
explain the rename was the last place I applied the rename. It told Docker
operators to set `WS_API_ADDRESS` three paragraphs above the section explaining
that `WS_API_ADDRESS` is no longer read. Fixed, then swept the whole tree for the
unnamespaced spellings — the only remaining hits are the deprecation notice
itself, its pin, and the doc sentence that names the old spellings on purpose.

**G2** — and the irony is deserved. The record-site filter was pinned, so
re-applying the old suppression there fails CI; the *branch selection* was not,
so deleting the widening `else` achieved the same silence with every test green.
That is the cannot-fail shape sitting inside the function whose comment block
warns about it. Now pinned by message stem, along with the legacy-env array where
dropping a pair silently removes a deprecation notice. Both mutation-tested:

```
delete the widening else  → "no longer reports the widening re-derivation…"  FAILED
drop WS_API_ADDRESS pair  → "no longer reports a leftover `WS_API_ADDRESS`…"  FAILED
```

**G3** — three stale rustdocs, including one claiming the exposure warning needs
hosted-mode-off while the body of the very test it documented asserted the
opposite. **G4** — the missing "and only when the bind actually moved" half of
the filter's contract. **G5** — the deferred probe now names advisory
GHSA-824h-7x5x-wfmf §8 rather than saying "tracked separately" with nothing to
follow. **G6** — the operator doc now states that no warning fires for loopback
plus hosted plus proxy, and that the silence means "not yet reported" rather than
"checked and safe".

Local on `ff06b69`: 156 config tests, 365 bin tests, `cargo fmt --check` and
`cargo clippy -p freenet -- -D warnings` clean.

CI: the gate jobs (Fmt, Rule Lint, Conventional Commits) and macOS bundle-swap
went green on the previous head before this push, which is the first real signal
today — Actions still reports major_outage but is evidently servicing jobs again.
This push restarts the run on `ff06b69`; that is the head that needs to be green.

[AI-assisted - Claude]

PR #5218, opened 2026-08-07
feat(dashboard): added bandwidth graph

### LINKED ISSUE #5152: Bandwidth usage graph in local dashboard
It would be nice to have a graph for the current bandwidth usage in the dashboard

Je3r:
Fix #5152

Added a graph showing the last minute of upload and download bandwidth in the dashboard.

The code was written by GPT 5.6 Sol (Codex)

### DIFF
--- a/crates/core/src/server/bandwidth_chart.test.mjs
+++ b/crates/core/src/server/bandwidth_chart.test.mjs
@@ -1,0 +1,679 @@
+// Behavioral tests for the browser-local bandwidth chart in dashboard.js.
+//
+// The production factory is extracted between exact, unique sentinels. Browser
+// dependencies are injected, so sampling, rendering, cleanup, and edge cases
+// run deterministically under Node without external packages.
+
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+import assert from 'node:assert/strict';
+
+const here = dirname(fileURLToPath(import.meta.url));
+const assetPath = join(here, 'home_page/assets/dashboard.js');
+const src = readFileSync(assetPath, 'utf8');
+
+function exactRegion(begin, end) {
+  const start = src.indexOf(begin);
+  const finish = src.indexOf(end);
+  assert.notEqual(start, -1, 'missing source sentinel: ' + begin);
+  assert.notEqual(finish, -1, 'missing source sentinel: ' + end);
+  assert.equal(src.lastIndexOf(begin), start, 'duplicate source sentinel: ' + begin);
+  assert.equal(src.lastIndexOf(end), finish, 'duplicate source sentinel: ' + end);
+  assert.ok(finish > start, 'source sentinels are out of order');
+  return src.slice(start + begin.length, finish);
+}
+
+const factorySource = exactRegion(
+  '/* bandwidth-chart-factory:BEGIN */',
+  '/* bandwidth-chart-factory:END */',
+);
+const createBandwidthChart = new Function(
+  factorySource + '\nreturn createBandwidthChart;',
+)();
+
+function hasClass(element, className) {
+  return (element.getAttribute('class') || '').split(/\s+/).includes(className);
+}
+
+function matches(element, selector) {
+  if (selector === '[data-bandwidth-chart]') {
+    return Object.hasOwn(element.attrs, 'data-bandwidth-chart');
+  }
+  if (selector === 'svg.bw-chart') {
+    return element.getAttribute('tag') === 'svg' && hasClass(element, 'bw-chart');
+  }
+  if (selector.startsWith('.')) return hasClass(element, selector.slice(1));
+  return false;
+}
+
+function fakeEl(attrs = {}) {
+  return {
+    attrs: { ...attrs },
+    style: {},
+    innerHTML: '',
+    children: [],
+    parentNode: null,
+    setAttribute(name, value) {
+      this.attrs[name] = String(value);
+    },
+    getAttribute(name) {
+      return this.attrs[name] ?? null;
+    },
+    appendChild(child) {
+      if (child.parentNode) child.parentNode.removeChild(child);
+      child.parentNode = this;
+      this.children.push(child);
+    },
+    removeChild(child) {
+      const index = this.children.indexOf(child);
+      if (index >= 0) {
+        this.children.splice(index, 1);
+        child.parentNode = null;
+      }
+    },
+    querySelector(selector) {
+      for (const child of this.children) {
+        if (matches(child, selector)) return child;
+        const nested = child.querySelector(selector);
+        if (nested) return nested;
+      }
+      return null;
+    },
+    querySelectorAll(selector) {
+      const found = [];
+      for (const child of this.children) {
+        if (matches(child, selector)) found.push(child);
+        found.push(...child.querySelectorAll(selector));
+      }
+      return found;
+    },
+    closest(selector) {
+      let current = this;
+      while (current) {
+        if (matches(current, selector)) return current;
+        current = current.parentNode;
+      }
+      return null;
+    },
+    contains(element) {
+      if (element === this) return true;
+      return this.children.some((child) => child.contains(element));
+    },
+    focus() {
+      this.focused = true;
+    },
+  };
+}
+
+function makeChart(options = {}) {
+  const appended = [];
+  const createdNS = [];
+  let now = options.now ?? 0;
+  const deps = {
+    now: () => now,
+    getRect: () => ({
+      left: options.left ?? 0,
+      top: options.top ?? 0,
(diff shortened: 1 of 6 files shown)


sanity:
@Je3r — thank you for this, and apologies for the three weeks of silence. You deserved an answer much sooner, and the answer isn't "no".

**The feature is wanted.** There's no bandwidth chart anywhere in `main`, and this is the only work cross-referenced on open issue #5152. The conflicts are also minor — I ran the three-way merges and there are exactly three, all mechanical: one hunk in `dashboard.js`, one in `style.css`, one line in `package.json`.

**But please don't rebase yet, because there's a design question to settle first — and it would be unfair to have you resolve conflicts on an approach we then ask you to change.**

**The architecture question.** `main` renders its charts **server-side as SVG in Rust** (see `peer_detail.rs` and `estimator.rs`). This PR renders **client-side in JavaScript** (+576 lines). Both are legitimate, but they're different house styles, and having two charting approaches in one dashboard is a maintenance cost that outlives whichever is nicer to write. This is a maintainer call rather than a code-review nit, so I'm flagging it to @sanity / @iduartgomez rather than deciding it at you.

Two concrete options:

- **Port to the server-side SVG pattern**, matching `peer_detail.rs`/`estimator.rs`. More work, consistent with the rest of the dashboard.
- **Keep it client-side**, and accept JS charting as a second sanctioned pattern in the dashboard — which should be a deliberate decision, not a side effect of merging.

**Two concrete bugs regardless of which way that goes:**

1. **`chart-section` is used but never defined.** The markup applies that class and no CSS rule for it exists, so the styling silently does nothing.
2. **The chart can fail to mount.** `build_transfer_card` early-returns empty for roughly the first 10 seconds, so the mount element doesn't exist yet when the script looks for it — the chart then appears only on a later refresh, which reads as a flaky feature to a user.

**What I'd suggest:** hold off on the rebase until we answer the server-vs-client question, and I'll chase that answer rather than leaving you waiting again. If the verdict is "port it", I'm happy to point you at the exact pattern to follow.

For transparency about the delay: we've just swept a 59-PR backlog and found that outside-contributor CI had never been approved — 58 workflow runs sitting unapproved, four of them yours. So this PR wasn't ignored on the merits; it was never actually runnable for anyone to judge. We're fixing that with automation so it doesn't recur.

[AI-assisted - Claude]

sanity:
@Je3r — the answer on the architecture question, so you're not left waiting again.

**Please port it to the server-side SVG pattern.** The dashboard renders its charts server-side in Rust (`peer_detail.rs`, `estimator.rs`), and this PR renders client-side in JS. Both approaches are reasonable in isolation; carrying two of them in one dashboard is a maintenance cost that outlives whichever is nicer to write, and consistency is the right default when there's no specific reason to diverge.

Being straight about the authority here: this is a house-style call, and **@sanity or @iduartgomez can overrule it** — if either says client-side JS is welcome, that wins and this comment is void. But you shouldn't have to wait on that to make progress, and "match the existing pattern" is the answer that's safe to act on now.

**Porting also resolves both bugs for free**, which is part of why it's the cheaper path:

1. `chart-section` is applied in the markup but no CSS rule for it exists, so that styling silently does nothing today.
2. `build_transfer_card` early-returns empty for roughly the first 10 seconds, so the mount element doesn't exist when the script looks for it — the chart appears only on a later refresh, which reads as a flaky feature rather than a loading state.

Both are artifacts of mounting client-side into a server-rendered card. Rendering the chart in the same pass that builds the card makes them not arise.

**The feature itself is wanted.** There's no bandwidth chart anywhere in `main`, and this is the only work cross-referenced on open issue #5152. Please don't read this as a rejection — it's a request to change the rendering approach, not the idea.

**On the conflicts: don't rebase yet.** There are only three and all are mechanical (one hunk in `dashboard.js`, one in `style.css`, one line in `package.json`), but if you're porting to server-side rendering the JS-side conflicts stop mattering. Port first, then rebase once — no point resolving conflicts in code you're about to replace.

If you'd like a concrete starting point, `estimator.rs` is the closest analogue to what this PR does and is a reasonable template to follow.

For transparency about the three-week delay: we found during a backlog sweep that outside-contributor CI had never been approved — 58 workflow runs sitting unapproved, four of them yours. So this PR wasn't judged and set aside; it was never actually runnable. That's fixed, your runs are approved, and automation is going in so it doesn't recur.

[AI-assisted - Claude]

sanity:
@Je3r — coming back to this, and I owe you a correction rather than another nudge.

I re-reviewed this PR from scratch. **One of the two bugs I cited when I asked you to port to server-side SVG does not exist, and the argument I used to justify the port was backwards.** You've been sitting on an instruction that was partly wrong, so let me fix that before anything else.

### Retracting bug #1 — it isn't a bug

I said "`chart-section` is applied in the markup but no CSS rule for it exists." That's wrong. Your PR adds it, in the same diff:

```css
.chart-section { ... }
.chart-section h3 { ... }
```

I grepped `main`'s `style.css`, where it is indeed absent, instead of yours. My mistake, and I'm sorry — being told to fix a bug that isn't there is a genuinely annoying way to lose a week.

### Correcting the "porting is the cheaper path" claim

I wrote that porting to server-side SVG "resolves both bugs for free, which is part of why it's the cheaper path." That was wrong in a way that matters for how much work I was actually asking of you.

A server-side rate chart needs the **server** to hold a time series. It doesn't have one. `network_status.rs:2333-2334` collects only cumulative totals:

```rust
let bytes_uploaded = TRANSPORT_METRICS.cumulative_bytes_sent();
let bytes_downloaded = TRANSPORT_METRICS.cumulative_bytes_received();
```

There is no ring buffer, no sample history, nothing to difference into a rate. The two server-side SVGs I pointed you at (`cards.rs::build_ring_svg`, `estimator.rs`) both render state that **already exists** — which is exactly why they're cheap to render server-side and a bandwidth *rate* chart is not.

So porting isn't a restyling of your work. It means first adding server-side bandwidth sampling, then rendering from it. That's a substantially **larger** change than I represented, and your design note says as much in your own Rust test:

> *The browser owns bandwidth history so HTTP snapshot reads remain side-effect free.*

That's a real argument I talked past. Server-held history does have a genuine advantage — it survives a page reload, where yours doesn't — so the port may still be the right call. But it needs to be asked for honestly, as "please also build the sampling layer", not as a free consequence.

### Bug #2 stands, but it's milder than I said

`build_transfer_card` early-returns empty only when **both** counters are 0 **and** uptime < 10s (`cards.rs:522`). Since `sampleAndRender` re-queries the mount on every refresh, it self-heals within one cycle. That's a one-cycle-late mount in a narrow window, not "a flaky feature". I overstated it.

### One real bug I missed

`sampleAndRender` wipes accumulated history whenever the mount is absent:

```js
var host = root && root.querySelector('[data-bandwidth-chart]');
if (!host) {
  clearOverlay();
  clearHistory();
  ...
```

If the transfer card disappears transiently — a snapshot going `None` during a reconnect — your accumulated minute of history is discarded rather than preserved across the gap. Worth guarding whichever way the architecture question lands.

### Your CI has never actually run

Worth knowing, because it isn't your fault and it changes what "green" means here. There are four workflow runs against your head commit `3f3f571` — `CI`, `Shell Playwright E2E`, `Lint server web assets`, `macOS DMG-swap E2E` — and all four are recorded as `failure` with **zero jobs executed**. That's a startup failure from the fork-approval flow, not your code failing a test. The four checks showing green on the PR are only the CLA, Snyk and the rule review.

So there is no real build, test or lint signal on this branch, and there never has been.

One consequence to plan for: `main` bumped **eslint 9.39 → 10.9** (a major) in early September, and your 575 new JS lines have never been linted under it.

### Where this leaves the architecture question

Genuinely still open, and it's mine to settle rather than yours. What I can tell you now:

- **The feature is wanted.** Issue #5152 is still open and unassigned, and there is still no bandwidth chart anywhere in `main` — I re-checked. Nothing has superseded your work.
- **Your conflicts are still mechanical**, and still only three: the `oldMain.innerHTML` swap line in `dashboard.js` (nest inside main's new `captureFilterFocus`/blur), the `:root` block in `style.css` (union, plus values for main's new `prefers-color-scheme: light` block, since the dark values wash out on `#f7f5f2`), and the `"test"` script in `package.json` (union).
- **The port instruction is suspended**, not confirmed, pending the corrected cost above. I'm putting it back to @sanity / @iduartgomez with a recommendation that we accept client-side JS for *time-series* charts specifically — the server holds no history, and that's a principled line rather than an exception.

**Please don't do any work on this until that's settled** — you shouldn't rebase or rewrite twice on my account. I'll follow up here with a decision either way.

Thank you for your patience on this; five weeks and a wrong instruction is not the experience a contribution this well-tested deserved. The `.mjs` test suite in particular — counter resets, backward clocks, BigInt overflow, window gaps — is more thorough than a lot of what we merge.

[AI-assisted - Claude]

PR #5224, opened 2026-08-08
fix(diagnostics): report real node uptime instead of a hardcoded zero

### LINKED ISSUE #5223: Node diagnostics: uptime_seconds is always 0 (hardcoded literal)
## Problem

`NodeInfo::uptime_seconds` in the node-diagnostics response is always `0`.

The field is a hardcoded literal in the `QueryNodeDiagnostics` handler:

```rust
// crates/core/src/node/network_bridge/p2p_protoc.rs
uptime_seconds: 0, // TODO: implement actual uptime tracking
```

The `TODO` has been there for the field's whole lifetime, so every
`freenet service report` upload has claimed the node had zero uptime.

## Impact

This is the field a human reaches for when diagnosing a restart loop or a
wedged node, so a silently-dead value is actively misleading rather than
merely absent — someone debugging a node that keeps restarting would read
`uptime_seconds = 0` as confirmation of their hypothesis.

Same failure class as #3987, where a `HashMap` with non-string keys made the
whole `network_status` field upload empty for roughly ten months unnoticed.

## Evidence

Three independent user diagnostic reports, all reading `uptime_seconds = 0`
while every sibling field in the same response carries real data:

| Report | Connected peers | Hosting contracts | `uptime_seconds` |
|---|---|---|---|
| `R7JSRK` | 18 | 933 | `0` |
| `WEWYWY` | 169 | 1813 | `0` |
| `R7W5NC` | 83 | 706 | `0` |

`R7JSRK` came from a node that had been running for roughly seven hours at
report time. In all three, `location`, `listening_address`, `is_gateway`,
`active_connections` and `system_metrics` are populated correctly — the zero
is specific to this one field, not a symptom of a broken response.

## Scope

Only `uptime_seconds` is dead. Every other field in `NodeDiagnosticsResponse`
was checked against the three reports and carries real values.

One adjacent oddity, noted but deliberately not changed here: the handler
computes `let _network_subs = op_manager.get_network_subscriptions();` and
discards it, so `response.subscriptions` contains application subscriptions
only. That is a separate question about intended semantics, not a dead field.

## Fix

Record the node's start instant and compute elapsed seconds when a
diagnostics query arrives, drawing the clock from the node's existing
injectable time source so it honours the DST rule against `Instant::now()`.

[AI-assisted - Claude]

sanity:
## Problem

`NodeInfo::uptime_seconds` in the node-diagnostics response is always `0`. It
was a hardcoded literal in the `QueryNodeDiagnostics` handler:

```rust
uptime_seconds: 0, // TODO: implement actual uptime tracking
```

The `TODO` had been there for the field's whole lifetime, so every
`freenet service report` upload has claimed the node had zero uptime.

This is the field a human reaches for when diagnosing a restart loop or a
wedged node, so a silently-dead value is actively misleading rather than
merely absent — someone debugging a node that keeps restarting would read
`uptime_seconds = 0` as confirmation of their hypothesis. Same failure class
as #3987, where a `HashMap` with non-string keys made the whole
`network_status` field upload empty for about ten months unnoticed.

### Evidence

Three independent user diagnostic reports, all reading `uptime_seconds = 0`
while every sibling field in the same response carries real data:

| Report | Connected peers | Hosting contracts | `uptime_seconds` |
|---|---|---|---|
| `R7JSRK` | 18 | 933 | `0` |
| `WEWYWY` | 169 | 1813 | `0` |
| `R7W5NC` | 83 | 706 | `0` |

`R7JSRK` came from a node that had been running roughly seven hours at report
time. In all three, `location`, `listening_address`, `is_gateway`,
`active_connections` and `system_metrics` are populated correctly — the zero
is specific to this one field, not a symptom of a broken response.

## Solution

`P2pConnManager` records the instant it was built and computes elapsed seconds
when a diagnostics query arrives.

Two details worth a reviewer's attention:

- **The clock is the node's injectable time source**, taken from
  `NodeConfig::hosting_time_source_override` and falling back to
  `InstantTimeSrc`. This mirrors how `OpManager` sources the interest
  manager's clock (`op_state_manager.rs`), honours the DST rule against bare
  `Instant::now()` in `crates/core/`, and lets a simulation advance uptime
  deterministically.
- **The start instant is carried across the destructure/rebuild** at
  event-loop entry rather than re-read from the clock. Re-reading there would
  restart the measurement at event-loop entry and silently under-report uptime
  for the life of the process. There is an inline comment at that site saying
  so, because it is the obvious thing for a future edit to get wrong.

The helper saturates rather than panicking if the clock reports an instant
before the start. That is unreachable with the production monotonic clock, but
a diagnostics query must never be able to take the node down.

### Scope

Only `uptime_seconds` was dead. Every other field in `NodeDiagnosticsResponse`
was checked against the three reports above and carries real values.

One adjacent oddity, noted but deliberately **not** changed here to keep this
to one logical change: the handler computes
`let _network_subs = op_manager.get_network_subscriptions();` and discards it,
so `response.subscriptions` contains application subscriptions only. That is a
question about intended semantics rather than a dead field, and is filed
separately if it turns out to be wrong.

No wire-format or stdlib change: `uptime_seconds` already exists on
`freenet_stdlib::client_api::NodeInfo`.

## Testing

Two tests reproduce the bug. Both were verified by re-applying the exact
`uptime_seconds: 0` literal and re-running:

- `test_node_diagnostics_reports_nonzero_increasing_uptime`
  (`crates/core/tests/node_diagnostics_uptime.rs`) boots a gateway and queries
  diagnostics over the real WebSocket client path twice, asserting uptime is
  non-zero **and strictly increasing**. Under the mutation it fails, with the
  node's own log showing the wire response carrying `uptime_seconds: 0`. The
  increasing assertion is what stops a future hardcoded constant from passing
  a bare `> 0` check.
- `node_diagnostics_computes_uptime_rather_than_hardcoding_it` is a
  source pin bounded to the `NodeInfo { .. }` literal in the handler, so
  re-inlining a value at the call site fails CI. It fails under the mutation
  with the offending literal printed.

Two further unit tests cover the helper's boundaries on a mock clock:
sub-second truncation, the 1-second boundary, the seven-hour `R7JSRK` case,
and a clock behind the start instant. **These pass under the mutation by
design** — they exercise the helper directly, so they are coverage for the
computation, not reproductions of the call-site bug.

Also verified end to end: the e2e test asserts against a real node's real
diagnostics reply rather than a unit-level stub, which is what makes it catch
the handler dropping the helper call.

Closes #5223

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/node/network_bridge/p2p_protoc.rs
+++ b/crates/core/src/node/network_bridge/p2p_protoc.rs
@@ -35,6 +35,7 @@ use crate::transport::{
     Socket, TransportError, TransportKeypair, TransportPublicKey, create_connection_handler,
     global_bandwidth::GlobalBandwidthManager, peer_connection::StreamId,
 };
+use crate::util::time_source::{DynTimeSource, InstantTimeSrc};
 use crate::{
     client_events::ClientId,
     config::GlobalExecutor,
@@ -678,6 +679,26 @@ fn collect_contract_states<O: ContractPresenceOracle + ?Sized>(
     states
 }
 
+/// Whole seconds elapsed between `started_at` and `clock`'s current instant.
+///
+/// Backs `NodeInfo::uptime_seconds` in the node-diagnostics reply. Truncates
+/// toward zero, so a node younger than one second reports `0` — that is the
+/// only case in which a healthy node legitimately reports zero uptime.
+///
+/// Saturating rather than panicking: `Instant` subtraction underflows if the
+/// clock reports an instant before `started_at`. Wall-clock `InstantTimeSrc` is
+/// monotonic and cannot do that, but an injected test clock can be constructed
+/// at an earlier epoch, and a diagnostics query must never take the node down.
+///
+/// Measures MONOTONIC time, which does not advance while the host is
+/// suspended. A laptop peer resumed after an overnight suspend therefore
+/// reports the time it was actually running, not wall-clock time since start.
+/// That is the more useful number for "how long has this node been serving",
+/// which is what the field is read for.
+fn uptime_seconds(clock: &DynTimeSource, started_at: Instant) -> u64 {
+    clock.now().saturating_duration_since(started_at).as_secs()
+}
+
 pub(in crate::node) struct P2pConnManager {
     pub(in crate::node) gateways: Vec<PeerKeyLocation>,
     pub(in crate::node) bridge: P2pBridge,
@@ -697,6 +718,17 @@ pub(in crate::node) struct P2pConnManager {
     listening_ip: IpAddr,
     listening_port: u16,
     is_gateway: bool,
+    /// Clock backing `NodeInfo::uptime_seconds` in the node-diagnostics reply.
+    ///
+    /// Always a wall-clock [`InstantTimeSrc`], including under simulation — see
+    /// the rationale at its construction in `build`. Held as a
+    /// [`DynTimeSource`] rather than a bare `Instant` so the computation goes
+    /// through the `TimeSource` abstraction and stays unit-testable.
+    uptime_clock: DynTimeSource,
+    /// Instant this node's network event loop was constructed, read from
+    /// [`Self::uptime_clock`]. `NodeInfo::uptime_seconds` is the elapsed time
+    /// since this point.
+    started_at: Instant,
     /// If set, will sent the location over network messages.
     ///
     /// It will also determine whether to trust the location of peers sent in network messages or derive them from IP.
@@ -1236,6 +1268,22 @@ impl P2pConnManager {
             .connection_manager
             .try_set_own_addr(advertised_addr);
 
+        // Uptime is measured from here — the last step of node construction,
+        // immediately before the event loop starts serving. Reading `now()` once
+        // and storing the instant (rather than re-deriving a start time later)
+        // keeps `uptime_seconds` monotonic for the life of the process.
+        //
+        // Deliberately NOT `config.hosting_time_source_override`, even though
+        // that is the node's injectable clock and `OpManager` sources the
+        // interest manager from it. Simulations freeze and jump that clock to
+        // drive hosting TTL and eviction, so borrowing it here would make a sim
+        // node report either a frozen 0 or a 24h uptime seconds after boot.
+        // Uptime and hosting TTL are unrelated concerns; injectability lives at
+        // the `uptime_seconds` helper boundary, which is where the unit tests
+        // drive a mock clock.
+        let uptime_clock: DynTimeSource = Arc::new(InstantTimeSrc::new());
+        let started_at = uptime_clock.now();
+
         Ok(P2pConnManager {
             gateways,
             bridge,
@@ -1248,6 +1296,8 @@ impl P2pConnManager {
             listening_ip: listener_ip,
             listening_port: listen_port,
             is_gateway: config.is_gateway,
+            uptime_clock,
+            started_at,
             this_location: config.location,
             check_version: !config.config.network_api.ignore_protocol_version,
             ack_version_floor_override: config.ack_version_floor_override,
@@ -1329,6 +1379,8 @@ impl P2pConnManager {
             listening_ip,
             listening_port,
             is_gateway,
+            uptime_clock,
+            started_at,
             this_location,
             check_version,
             bandwidth_limit,
@@ -1417,6 +1469,11 @@ impl P2pConnManager {
             listening_ip,
             listening_port,
             is_gateway,
+            // Carried across the destructure/rebuild, NOT re-read from the clock:
+            // re-reading here would restart the uptime measurement at event-loop
+            // entry and silently under-report it for the life of the process.
+            uptime_clock,
+            started_at,
             this_location,
             check_version,
             bandwidth_limit,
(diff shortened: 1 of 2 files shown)


sanity:
## Review

Light tier per the risk table (small, self-contained diagnostic correction; no
wire-format, protocol, routing or auth surface). Run with two independent
Claude reviewers, each blind to the other and to the author's reasoning:

- **correctness** — clock source, capture point, the destructure/rebuild,
  event-loop safety, and whether the fix actually reaches the uploaded report.
- **test quality** — source-pin vacuity, e2e flakiness, whether the test
  exercises the production path, coverage gaps, and whether CI runs it at all.

Both reviewers independently landed on the same top finding, which was a real
bug in the first version of this PR.

### Fixed

**Sourcing the clock from `hosting_time_source_override` (SHOULD-FIX, both reviewers).**
The original commit drew the uptime clock from the node's injectable time
source, mirroring how `OpManager` sources the interest manager. That is wrong:
simulations deliberately freeze and jump that clock to drive hosting TTL and
eviction, so a sim node would have reported either a frozen `0` forever or a
24h uptime seconds after boot. The reviewers also noted the branch was
untested and its stated justification unrealised — nothing anywhere advanced
uptime through it. Now an unconditional `InstantTimeSrc`, with the reasoning
recorded at the construction site so the next author does not re-derive it.
Injectability lives at the `uptime_seconds` helper boundary, which is where
the unit tests drive a mock clock, so the DST rule is still satisfied.

**Nothing pinned the carry-across-destructure (NIT, testing reviewer).**
Correct and worth acting on: re-reading the clock at the ctx rebuild would
restart the measurement, and *no existing test would catch it* — uptime from
event-loop entry is still non-zero and still increasing, so the e2e test and
the other pin both stay green. Added
`event_loop_ctx_carries_start_instant_rather_than_re_reading_it`, and
mutation-tested it (applying `started_at: uptime_clock.now()` at the rebuild
makes it fail).

**`MAX_PLAUSIBLE_TEST_UPTIME_SECS` was an assertion with no red state (NIT, both).**
The comment claimed it caught a UNIX-epoch mix-up, which is not representable
through `Instant`, so no input could fail it. Tightened from 3600 to 300 —
now genuinely fails if `started_at` were measured from machine boot or a
process-wide instant shared across tests in the binary — and the comment now
says what it actually buys.

**Monotonic-vs-wall-clock semantics undocumented (NIT, correctness reviewer).**
`Instant` does not advance across host suspend, which matters because
diagnostic reports come disproportionately from laptop peers. Documented in
the helper's rustdoc, with the rationale for preferring "time actually
serving" over wall-clock elapsed.

**`self.is_gateway` beside `ctx.uptime_clock` (NIT).** Now reads `ctx` for the
whole literal.

### Acknowledged, deliberately not done here

**A third "uptime" origin (SHOULD-FIX, correctness reviewer).** The tree
already has `p2p_impl.rs:455` (`start_time`, feeding the shutdown telemetry
`uptime_secs` and the fatal-listener exit-code classification) and
`REDB_POISON_PROCESS_START`. This adds a third, captured at
`P2pConnManager::build`, so the diagnostics uptime and the telemetry shutdown
uptime for one node will differ slightly, and neither is process start. The
reviewer's suggested fix — promote the poison instant to a general
`PROCESS_START` read by all three — is sound but reaches into the crash-loop
exit-code and auto-rollback machinery, which is a materially riskier change
than the diagnostic correction this PR is scoped to. Filed as **#5226**.

**Capture point excludes startup work (SHOULD-FIX, same reviewer).** `build`
runs after the redb open/migration, so a node that spent minutes wedged in
migration reports uptime from after that. Real, and the same follow-up covers
it. It does not affect the reported bug: a node up seven hours now reports
~seven hours instead of `0`, which is the entire user-visible symptom. Called
out explicitly rather than left implicit, since "uptime" now has a specific
meaning worth knowing — time the network event loop has been serving.

**nextest `retries = 2` could mask e2e flakiness.** The testing reviewer
confirmed `second > first` is deterministic (`#[freenet_test]` expands to a
plain multi-thread `tokio::test` with no paused/virtual time, and
`tokio::time::sleep` never returns early, so `second >= first + 2`), so a
`retries = 0` override would be guarding a failure mode that cannot occur.
Left alone rather than touching CI config.

### Verified clean by review

- Exactly three `P2pConnManager { .. }` sites tree-wide; no fourth
  initialisation that could diverge. `build` is not called in any retry loop,
  and the event loop cannot be restarted in place.
- The source pin is **not** vacuous: `production_src()` cuts at
  `\npub(crate) mod tests {`, so the pin's own needle strings are excluded —
  the exact self-match failure this repo has shipped before. Every way the
  anchor can break fails loudly rather than passing.
- End-to-end path traced: `freenet service report` → `report.rs` →
  `client_events.rs` → this handler → `diagnostics_to_json`. No second
  producer of `uptime_seconds` (`report.rs:1030`'s `3600` is a serialisation
  fixture).
- CI does run the new test: no `required-features`, no `cfg` guard, and the
  only nextest exclusion is `blocked_peers`.
- No event-loop blocking: the helper is O(1), no await, no allocation, no lock
  in production.

[AI-assisted - Claude]

sanity:
## End-to-end verification against a real node

Beyond the automated tests, I built the binary from this branch and ran the
actual user-facing command against a real running node. `--local` writes the
report to disk and never uploads, so nothing was sent to the production
endpoint.

Isolated gateway node, then:

```
$ freenet --config-dir $D --data-dir $D service report --local $D/report2.json --no-message
  Querying local node diagnostics (up to 60s)... ok
  - Version: 0.2.121 (86714bf6505b)
  - Node status: running
```

Reading the generated report the same way the three broken production reports
were read (`report.network_status` → `node_info`):

```
network_status keys = ['connected_peers_detailed', 'contract_states',
                       'network_info', 'node_info', 'subscriptions',
                       'system_metrics']
uptime_seconds      = 208
is_gateway          = True
listening_address   = 127.0.0.1:53510
location            = 0.970421
```

The node had been up roughly three and a half minutes, so `208` is the correct
value rather than merely a non-zero one. Same report structure as R7JSRK /
WEWYWY / R7W5NC, which read `0` in that field.

Worth noting for anyone reproducing this: the field was already present in the
uploaded reports (as `0`), so the JSON serialisation path was never the
problem — only the value. That is what distinguishes this from #3987, where
serialisation dropped the whole `network_status` object.

Unrelated snag hit while doing the above: in a **debug** build,
`freenet service report` panics on a reached `unreachable!()` in
`ConfigPathsBuilder::build` unless both `--config-dir` and `--data-dir` are
passed. Release builds are unaffected, which is why real users' reports upload
fine. Filed separately as #5229; not touched here.

[AI-assisted - Claude]

Issue #5222, opened 2026-08-07
Post-mortem: why the fleet-wide auto-update break passed CI, and the process changes it implies


sanity:
## Why the fleet-wide auto-update break (#5221) passed CI, and what to change

The bug: from v0.2.120, the node's update check fed the raw GitHub tag `"v0.2.121"` to `semver::Version::parse`, which rejects a leading `v`. Every update was silently dropped. Two releases shipped with it. It was found only because adoption telemetry looked wrong and someone chased it for six hours.

Notably, **the test suite was green throughout, and so were three separate human-facing surfaces.**

### Five whys

**1. Why did the fleet stop auto-updating?**
The node's in-process check parsed a `v`-prefixed tag as semver, failed, and returned "no update" — silently, at `auto_update.rs:1524` (startup path) and `:462` (the urgent/discovery/mismatch path).

**2. Why was a `v`-prefixed tag reaching a semver parser?**
PR #5104 changed `parse_tag_from_release_location` to return the tag **verbatim** and introduced `version_from_tag()` to strip it. The stripper was wired into the **installer** path only (`update.rs:389`). The node's path was never updated.

**3. Why did that change land without the node path being updated?**
The change served a real installer need: the raw tag is required to build asset URLs. The producer's contract was changed for one consumer's benefit, and the other consumers were not audited. Nothing forced that audit, because **both the raw tag and the version are `String`**. There is no type distinction between "a tag that addresses a release" and "a version you can compare", so the compiler could not object.

**4. Why did the tests not catch it?**
Both halves were tested — against **contradictory contracts**:
- The producer test (`auto_update.rs:2608-2613`) asserts the tag comes back **verbatim**, comment: *"the tag is returned VERBATIM"*.
- The consumer tests (`startup_update_check_with_fetcher`, ~:1703-1759) inject **already-stripped** literals: `Ok("0.1.75".to_string())`.

Each passed. Neither tested the seam. The suite asserted a contract on one side and its opposite on the other, and reported success.

**5. Why could the test double diverge from the real producer?**
Because the fake is typed `Result<String>` — identical to the real fetcher — so nothing constrains it to emit what the real one emits. The double was free to drift, and did.

**Root cause:** a stringly-typed boundary carrying two different contracts, with test doubles that were not required to honour the producer's actual output format.

### Why it stayed invisible in production

Three surfaces normalise **independently** and all reported correctly while the deciding path failed:
[...]

PR #5225, opened 2026-08-08
fix(update): normalise the release tag on the node's detection path

### LINKED ISSUE #5221: Fleet cannot auto-update: #5104 made the node parse the raw 'v0.2.121' tag as semver and silently drop every update
## Summary

0.2.121 is propagating roughly **10x slower** than 0.2.120 did, and the difference is not explained by the documented pacing. Six and a half hours after publish it sits at ~9.6% of the fleet, where 0.2.121's predecessor reached ~85% in under five hours.

This matters beyond one release: **0.2.121 is itself a security fix** (client API loopback binding + withholding delegate notifications from non-local connections, closing GHSA-824h-7x5x-wfmf). If most peers cannot take an update, the network cannot be patched.

## Measured adoption, same method, same instrument

Version share of telemetry records, sampled from rotated collector logs. Each release's own pre-release window is included as a control and reads **0%**, which validates the method.

```
0.2.120  (published 2026-08-05 17:07Z)      0.2.121  (published 2026-08-07 16:29Z)
  pre-release:  0.0%                          pre-release:  0.0%
  +1.2h:       31.4%                          +2.6h:        2.0%
  +4.6h:       84.7%                          +4.0h:        6.8%
  +9.0h:       88.4%                          +6.1h:       ~9.0%
                                              +6.7h:        9.6%
```

Caveat stated up front: this is **record share, not distinct-peer share**. A freshly-updated node emits fewer records (fewer contracts, and the node-side telemetry limiter discards load-proportionally), so this metric *understates* adoption. The same bias applies to both releases, so the comparison holds, but the absolute figures are soft.

Second caveat: this is an **n=1 comparison**. 0.2.119 published 2026-08-03 15:45Z and telemetry retains only ~3.2 days, so its early curve has aged out. It is therefore possible that 0.2.120 was anomalously *fast* rather than 0.2.121 being slow.

## Ruled out, with evidence

| Hypothesis | Evidence against |
|---|---|
| Updater broken | `freenet update` verified end-to-end on a real NATed peer: check → download → signature verify → install → restart, exit 0 |
| Missing or unsigned artifacts | Asset set byte-for-byte comparable to 0.2.120, including `SHA256SUMS.txt.sig` |
| Gateways not updated | nova and vega both verified **at process level** running 0.2.121, started within minutes of publish |
| Breaking-change fast path not firing | `min-compatible-version = 0.2.64` for **both** releases, so neither trips `signal_urgent_update()` |
| Crash-loop auto-rollback reverting nodes | Refuted twice: (a) 0.2.121 nodes observed with **5.1h continuous uptime**, impossible under rollback churn; (b) no `known_bad_version` marker anywhere on an inspected host — it persists after a rollback, so its absence means none ever completed |
| GitHub outage | The 2026-08-06 incident was **Actions/Pages only**, no API Requests component, and resolved 02:04Z — 14h before publish |
| GitHub rate limiting | 60/60 unauthenticated requests available; `freenet update --check` correctly reports 0.2.121 |
| Discovery threshold unmet | Affected nodes have **169 and 83 connected peers**; at ~9% adoption they should each observe far more than `MIN_VERSION_REPORTERS = 2` peers already on 0.2.121 |

## The mechanism, and where it can break

Auto-update is **supervised-install-only** (`auto_update.rs`). The running node never replaces its own binary:

> it exits 42 and relies on its supervisor to apply the update and restart it

So the chain is: detect → **exit 42** → supervisor `ExecStopPost` runs `freenet update` → restart.

The load-bearing consequence is that **the systemd unit is not updated when the binary auto-updates**. A node whose unit predates the exit-42 hook detects the update, exits, and is restarted unchanged — indefinitely, across any number of releases. `auto_update.rs` names this case explicitly:

> an OLD systemd unit (e.g. a node auto-updated to a 45-aware binary but whose unit file was never regenerated)

This is consistent with every check passing: detection works, the API works, `freenet update` works. The untested link is the **handoff to the supervisor**, which cannot be exercised from outside an affected node.

## Field evidence

Three diagnostic reports from affected users:

- Node on **0.2.120**, 169 peers, running, healthy. `~/.local/state/freenet/install_failures.json` **does not exist** — so it has never attempted and failed an install. The other updater state files appear empty/absent too, i.e. no auto-update state has ever been written.
- Node on **0.2.120**, 83 peers, running, healthy.
- Node on **0.2.119** — it missed 0.2.120 as well, despite that release reaching ~85%. Evidence of a **chronic non-updating tail** that predates 0.2.121.

The chronic tail plausibly explains the individual user reports. It does **not** explain the aggregate difference, since the same tail existed during 0.2.120's rollout.

## The open question

**Either** 0.2.120 introduced a regression in the auto-updater — note that `auto_update.rs` gained **1,102 lines** between v0.2.119 and v0.2.120 (#5104, #5150, #5151, #5159), meaning the code now responsible for the entire fleet's updates **had never handled a live release before 0.2.121** — **or** the two-release comparison is misleading for a reason not yet identified.

The first possibility is the concerning one, because it is self-masking: the release that breaks updating is the release everyone is stuck on.

## Suggested next steps

1. **End-to-end reproduction**: take a supervised node with a known-good modern unit, downgrade it to 0.2.120, enable auto-update, and observe whether it detects, exits 42, and is healed by `ExecStopPost`. This directly exercises the one untested link.
2. **Instrument the handoff**: exit-42 emission and `ExecStopPost` invocation are currently invisible in telemetry. A counter for "update detected" versus "update applied" would have made this diagnosable in minutes instead of hours.
3. **Ask affected operators** for `systemctl cat freenet.service | grep -E "ExecStopPost|SuccessExitStatus"`. Absence of the `freenet update` hook means that node structurally cannot auto-update and needs its unit regenerated.
4. **Re-check the n=1 caveat** once 0.2.122 ships, which gives a third curve and settles whether 0.2.120 or 0.2.121 is the outlier.

[AI-assisted - Claude]

sanity:
## Problem

**The fleet cannot auto-update.** Since v0.2.120 the node fetches the GitHub release tag and hands it to `semver::Version::parse` **verbatim**, so `"v0.2.121"` fails to parse and every update is silently dropped.

Confirmed on a production node (not inferred):

```
INFO  Startup update check against GitHub current="0.2.120" jitter_secs=20
WARN  Startup update check: failed to parse latest version 'v0.2.121':
      unexpected character 'v' while parsing major version number
```

#5104 changed `parse_tag_from_release_location` to return the tag verbatim — correct for the installer, which builds asset URLs from it — and added `version_from_tag()` to strip. But the stripper was wired into the **installer** path only (`update.rs`). The node's detection path was never updated.

Measured impact: 0.2.120 reached ~85% fleet adoption in 4.6h; 0.2.121 has been flat at ~9-11% for over seven hours. The bug is present in **both** 0.2.120 and 0.2.121, so a fix cannot deliver itself — every operator has to run `freenet update` by hand once.

Tracking: #5221. Post-mortem: #5222.

## Solution

Normalise at the **producer**, in `get_latest_version`, not at each call site.

That function is the node's *detection* path — its own rustdoc says so. Every consumer of it wants a comparable version, never a string that addresses a release. One line at the boundary fixes:

- `compare_versions_for_startup` (the startup check — the site in the log above)
- `check_if_update_available` (the urgent / discovery / version-mismatch triggers, a **separate** broken site that the startup path's tests do not cover)
- the #4073 gates `is_version_pinned_bad` / `is_version_install_gated`

That last one is why the per-call-site fix is wrong. Those gates compare by **exact string equality** against values their writers store **stripped** (`update.rs` `begin_probation` / `record_install_failure`). Fixing only the parses would leave them comparing `"v0.2.121"` against a stored `"0.2.121"`, never matching — silently disabling the crash-loop-rollback pin and the install-failure gate, restoring the restart loop those gates exist to prevent, with no log line.

The raw tag is untouched where a release is **addressed**: `update.rs` still builds asset URLs from `latest_tag`.

## Testing

`get_latest_version_normalises_the_tag` — a **bounded** source pin (the function does network I/O with no injection seam), asserting the normalisation is present. **Verified to FAIL with the normalisation removed**, then pass once restored:

```
test get_latest_version_normalises_the_tag ... FAILED     # fix removed
test get_latest_version_normalises_the_tag ... ok         # fix restored
```

`raw_tag_is_rejected_by_the_consumers` — documents the real failure: a `v`-prefixed tag is **not** detected as an update, while the normalised form is. This is the behaviour that makes upstream normalisation necessary rather than optional.

Full binary suite: 367 passed, 0 failed.

## Why CI did not catch this

The producer's contract and the consumers' contract were each tested, and they **contradicted each other**:

- the producer test asserts the tag comes back **verbatim** (`"v1.2.3"`, with a comment saying so)
- every consumer test injects an **already-stripped** literal (`Ok("0.1.75".to_string())`)

Both passed. Nothing tested the seam. Because both the tag and the version are `String`, the compiler could not object, and the test double was free to drift from the real producer — and did.

#5222 tracks the structural follow-ups: typing the boundary so a release tag and a version cannot be confused (as `release-agent/src/github.rs` already does by returning `semver::Version`), and a release canary that actually exercises **auto**-update — the designated pre-release peer has been running `--disable-auto-update` since 2026-07-29, so the one machine positioned to catch this was configured not to.

## Scope

Deliberately minimal — this normalisation and its tests, nothing else. Every operator must install this release by hand, so a defect means asking the whole network to repeat that. The graceful-shutdown exit-code fix and the `uptime_seconds` fix are ready but held for the next release.

Note on ordering: the exit-code fix must land **after** this one. While this bug is live, graceful-shutdown-exits-1 accidentally fires the unit's `ExecStopPost` hook on every restart, which runs `freenet update` — whose installer path strips correctly and succeeds. That accident is currently the fleet's only working update route, and is the likeliest explanation for the ~10% that have moved.

Fixes #5221

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/auto_update.rs
+++ b/crates/core/src/bin/commands/auto_update.rs
@@ -953,7 +953,25 @@ async fn get_latest_version() -> Result<String> {
     // Never bypasses the cooldown: this is the automated path, and the whole
     // point of the cooldown is to keep it quiet while the IP is limited. (The
     // check above already returned; the flag covers the re-check inside.)
-    fetch_latest_release_tag(false).await
+    //
+    // Normalise HERE, at the producer, not at the call sites. This function is
+    // the node's DETECTION path (see the rustdoc above): every consumer of it
+    // wants a comparable version, never a string that addresses the release.
+    // `fetch_latest_release_tag` returns the tag VERBATIM (`v0.2.121`) since
+    // #5104, which is correct for the INSTALLER — `update.rs` builds asset URLs
+    // from it — but every consumer on this path either parses it as semver or
+    // compares it against a stored version, and both reject the leading `v`.
+    //
+    // #5104 stripped at one consumer (`update.rs:389`) and left this path raw,
+    // which silently broke auto-update fleet-wide for v0.2.120 and v0.2.121:
+    // `semver::Version::parse("v0.2.121")` fails, the update is dropped with a
+    // `warn!`, and nothing else notices. Stripping per-call-site would fix the
+    // parses and leave the #4073 gates below (`is_version_pinned_bad`,
+    // `is_version_install_gated`) comparing a `v`-prefixed string against the
+    // stripped values their writers store — silently disabling the
+    // crash-loop-rollback pin. One normalisation at the boundary fixes both.
+    let tag = fetch_latest_release_tag(false).await?;
+    Ok(version_from_tag(&tag).to_string())
 }
 
 /// Get the state directory for update tracking files.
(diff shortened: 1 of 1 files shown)


sanity:
Merging with Windows/macOS Service Unit still pending. Ian explicitly authorised this narrower exception: the four core checks (Clippy, Fmt, Unit & Integration, Simulation) are green on this exact HEAD, and the two platform service-unit checks take 33 min / 18 min respectively. Justification for the exception: this is an emergency fix for a fleet-wide auto-update failure (#5221) where every hour of delay leaves ~1,080 nodes unable to receive updates, and the change is a single-line normalisation in the node's update-detection path with no platform-specific surface.

Late-night merge restrictions also explicitly overridden by Ian for this action.

[AI-assisted - Claude]

Issue #5228, opened 2026-08-08
macOS/Windows wrappers cannot tell a service stop from a child-only SIGTERM, so a manual `pkill freenet` now leaves the node down


sanity:
Follow-up to #5227 / the PR that makes a graceful shutdown exit 0.

## Problem

The macOS launchd wrapper script (`service/macos.rs`) and the in-process wrapper used by the Windows/macOS tray (`service/wrapper.rs`) both decide what to do from the node child's exit code alone:

- `macos.rs`: `elif [ $EXIT_CODE -eq 0 ]; then log_event "Normal shutdown"; exit 0`
- `wrapper.rs::next_wrapper_action`: `0 => { state.reset_failure_streak(); WrapperAction::Exit }`

Neither can tell **"the service is being stopped"** from **"someone SIGTERMed the node child only"**. Both cases now present as child exit 0.

Before #5227 this was moot, because `freenet network` could not exit 0 — the branch was unreachable dead code. Now it is the branch a graceful stop takes.

## Consequences

Correct, and an improvement, when launchd/the wrapper is the one stopping: `launchctl stop`, `freenet service stop`, reboot. (Previously the macOS wrapper *fought* the stop — child exit 1 fell into the `else` branch, ran `freenet update --quiet`, and relaunched the node; launchd then SIGKILLed the wrapper at the end of its grace period, leaving a PPID-1 orphan holding the port, i.e. the #3967 stale-orphan signature.)

Wrong when the SIGTERM reached the child only. Real triggers:

- an operator running `pkill -f "freenet network"` to bounce the node — which the wrapper's own log line suggests ("Manual intervention may be required ('pkill freenet')");
- the startup/port-conflict sweeps in `macos.rs` and `wrapper.rs::kill_stale_freenet_processes`, both of which `pkill` with the default SIGTERM. If a user has both the LaunchAgent and the Freenet.app tray, whichever starts second silently terminates the other's node.

In those cases the node goes down, the wrapper exits, and the macOS plist's `KeepAlive.SuccessfulExit=false` means launchd does not respawn it — the node stays down until logout/reboot/`launchctl kickstart`.

Linux is unaffected: systemd has this information out of band (an explicit stop job suppresses `Restart=`), and `Restart=always` restarts after an exit 0 that was not a stop job.

## Suggested fix

Give each wrapper the same "was *I* asked to stop?" discriminator the node itself now has.

Shell wrapper (`macos.rs`) — the TERM trap already exists:

```sh
[...]

Issue #5227, opened 2026-08-08
Graceful shutdown exits 1, so a clean `systemctl stop` is counted as a crash and can trigger auto-rollback


sanity:
## Problem

A **graceful** shutdown of `freenet network` exits with status **1**, so a deliberate `systemctl stop` is reported to systemd as a failure and counted as a crash by the post-update crash-loop rollback (#4073).

Observed directly on a production node running 0.2.121 (a deliberate `systemctl --user stop`; nothing crashed):

```
18:15:14  CRITICAL: Network event listener exited: Graceful shutdown
18:15:16  Error: Graceful shutdown
18:15:16  systemd: freenet.service: Main process exited, code=exited, status=1/FAILURE
18:15:16  Freenet 0.2.121: crash 1/3 during post-update probation (node stop status 1);
          not updating, will auto-roll-back if it keeps crashing.
```

## Root cause

`crates/core/src/bin/freenet.rs::run_network_node_with_signals` returns `result` directly. On SIGTERM that `result` is `Err(EventLoopExitReason::GracefulShutdown)` — a typed sentinel, not a real error — so `main` falls through to its `eprintln!("Error: {e:?}") + std::process::exit(1)` fallback.

`rollback::classify_stop` (`crates/core/src/bin/commands/rollback.rs`) treats any status outside `{0, 2, 42, 43, 44}` as a crash. Three clean stops of a freshly updated node inside its probation window are therefore enough to auto-roll-back a perfectly healthy release.

Two independent pieces of evidence that exit 0 was the intent:

- the adjacent `if result.is_ok() { tracing::info!("Graceful shutdown complete") }` in the same function can never fire;
- both service wrappers already have a written-but-unreachable branch for it — `macos.rs` (`elif [ $EXIT_CODE -eq 0 ]` → `log_event "Normal shutdown"`) and `service/wrapper.rs` (`0 => WrapperAction::Exit`).

The classification predicate `p2p_impl::listener_exit_is_graceful` was already correct and already pinned by `listener_exit_graceful_classification`; it just did not reach the exit code. The bug lives entirely in that gap, which is why no in-process test caught it.

## Complication

[...]

PR #5230, opened 2026-08-08
fix(node): exit 0 on a requested graceful shutdown

sanity:
## Problem

A **graceful** shutdown of `freenet network` exited with status **1**. Observed directly on a production node running 0.2.121 — a deliberate `systemctl --user stop`, nothing crashed:

```
18:15:14  CRITICAL: Network event listener exited: Graceful shutdown
18:15:16  Error: Graceful shutdown
18:15:16  systemd: freenet.service: Main process exited, code=exited, status=1/FAILURE
18:15:16  Freenet 0.2.121: crash 1/3 during post-update probation (node stop status 1);
          not updating, will auto-roll-back if it keeps crashing.
```

`run_network_node` reports a clean SIGTERM stop as `Err(EventLoopExitReason::GracefulShutdown)` — a typed sentinel, not a real error — and `run_network_node_with_signals` returned it verbatim, so `main` fell through to its `eprintln!("Error: {e:?}")` + `std::process::exit(1)` fallback.

That matters because `rollback::classify_stop` treats any status outside `{0, 2, 42, 43, 44}` as a crash. Three clean stops of a freshly updated node inside its post-update probation window were enough to auto-roll-back a perfectly healthy release (#4073).

Two independent signals that exit 0 was always the intent:

- the adjacent `if result.is_ok() { tracing::info!("Graceful shutdown complete") }` in the same function could never fire;
- both service wrappers already contain a written-but-unreachable branch for it — `macos.rs` (`elif [ $EXIT_CODE -eq 0 ]` → `log_event "Normal shutdown"`) and `service/wrapper.rs` (`0 => WrapperAction::Exit`).

## Solution

`finish_run(result, shutdown_requested)` in `bin/freenet.rs` maps the sentinel to `Ok(())`, but only when **both** conditions hold:

1. `p2p_impl::listener_exit_is_graceful(&err)` — the same predicate the #4549 fatal-listener force-exit already uses (made `pub` and re-exported, so the two classifications cannot disagree about which errors are candidates);
2. **this process asked the node to stop** — an `AtomicBool` set by the signal task immediately before it calls `shutdown_handle.shutdown()`.

Condition 2 is the load-bearing half, and the reason this is not a one-line change. `EventLoopExitReason::GracefulShutdown` is **not** exclusive to a requested stop:

- `p2p_protoc.rs` raises the same value when a critical internal channel dies (`ChannelCloseReason::{Bridge, Controller, Notification, OpExecution}` — the path that logs `CRITICAL: Channel closed …`);
- `client_events.rs` sends the identical `NodeEvent::Disconnect` when every client transport has been lost, a degraded state its own comments call "restart recommended".

Neither is distinguishable from an operator stop at the p2p layer, and both must keep exiting non-zero: on Linux that is what still fires the unit's `ExecStopPost` self-heal and counts the crash toward probation, and on the macOS/Windows wrappers an exit 0 is read as "normal shutdown" and would leave the node **down** instead of relaunching it. Keying on the sentinel alone would have converted those faults into silent successes.

Nothing else changes. Every other error still exits non-zero, and the auto-update exit still surfaces as `UpdateNeededError` (exit 42).

## ⚠️ Merge ordering: this MUST land after the `v`-prefix update fix

**Do not merge this until the fix for the `v`-prefix parse bug (#5104 regression) is on `main`.** This is not a stylistic preference; merging it first makes the fleet strictly worse.

PR #5104 (shipped in 0.2.120) made the node's in-process update check hand the raw GitHub tag to `semver::Version::parse`. `compare_versions_for_startup` in `auto_update.rs` parses `latest` directly, and semver rejects a leading `v`, so on 0.2.120 and 0.2.121 every node logs

```
WARN Startup update check: failed to parse latest version 'v0.2.121':
     unexpected character 'v' while parsing major version number
```

and fails open — returning `None`, never exiting 42. The node's own update path is dead fleet-wide.

I verified the consequence rather than assuming it. Because a graceful shutdown currently exits **1**, it does not match the unit's `0|43` skip list, so `ExecStopPost` runs `freenet update --quiet` on **every** restart. That path goes through `update.rs::run_async`, which calls `version_from_tag()` (`tag.strip_prefix('v')`) *before* `Version::parse` — so it parses correctly and installs. Reachability confirmed end to end: exit 1 → `*)` arm → `FREENET_POST_STOP_EXIT_CODE=1 freenet update --quiet` → `handle_post_stop("1", …)` → `PostStopOutcome::Proceed` (no probation marker) → normal update flow → install.

So while #5104 is live, that accident is the only working auto-update route, and this PR removes it.

One caveat on the accidental path that the two bugs make worse together: it only installs when the node is **not** in post-update probation. A node freshly updated to 0.2.121 that is restarted three times inside its probation window hits `PostStopOutcome::CrashRecorded` instead and gets **rolled back** — because this very bug reports each clean stop as a crash. The accidental path therefore both delivers updates and, on a recently-updated node, can undo them.

## Behaviour change to be aware of: `ExecStopPost` no longer fires on a graceful stop

The generated systemd unit has:

```
ExecStopPost=-/bin/sh -c 'case "$EXIT_STATUS" in 0|43) ;; *) … freenet update --quiet ;; esac'
SuccessExitStatus=42 43
```

Today's exit 1 is neither 0 nor 43, so **`freenet update` currently runs on every graceful stop**. After this change a graceful stop exits 0 and that hook no longer fires. This is deliberate; it is safe because:

- **The intended trigger is unaffected.** A node that detects an update exits **42**, which still matches the `*)` arm and still runs `freenet update`. Verified against the template text in `service/linux.rs` (both the user and system units) — the skip list is only `0|43`.
- **Nothing depends on the update running at stop time.** Update discovery is the node's own boot-time `startup_update_check` plus the ~6h jittered re-poll, both of which end in exit 42. So `systemctl restart`, a reboot, and `freenet service restart` still pick up a new release — one boot later at worst.
- **`Restart=always` still restarts after an exit 0**, so a graceful stop that was not an explicit stop job still brings the node back on Linux.
- It also removes a race during release rollouts on units generated by `freenet service install`: previously every `systemctl stop` of such a unit spawned a `freenet update --quiet` that could install the GitHub latest underneath a deploy script's own binary swap. Note this never applied to the gateway fleet, which runs `scripts/freenet-gateway.service` — that unit has no `ExecStopPost` at all (only `Restart=always`), so no such race existed there.

Also checked and unaffected: `RestartPreventExitStatus=43`, the `StartLimit*` counters (they count starts, not statuses), `crates/release-agent/` and `gateway-update.yml` (both judge convergence by `/version` + `systemctl is-active`, never by the node's exit code), and the Windows service stop path (a hard `taskkill`, which never produces exit 0).

**Known consequence on the macOS/Windows wrappers**, called out rather than hidden: those wrappers cannot tell "the service is being stopped" from "someone SIGTERMed the node child only" (e.g. a manual `pkill -f "freenet network"`). With exit 0 they now take their `"Normal shutdown"` branch and stand down, where previously exit 1 made them relaunch. In the common case this is a **fix** — on macOS a `launchctl stop` previously had the wrapper fight the stop, relaunch the node, and get SIGKILLed, leaving a PPID-1 orphan holding the port (the #3967 stale-orphan signature). The manual-`pkill` case is a genuine regression on desktop; it needs a `WRAPPER_TERMINATING`-style flag in both wrappers and cannot be verified from CI (no runner exercises those paths), so it is filed as #5228 rather than smuggled in here with an untestable diff.

## Testing

**End-to-end on the process exit status**, which is the whole point — the classification predicate was already correct and already pinned by `listener_exit_graceful_classification`; the bug lived in the gap between that predicate and the exit code, where no in-process test can see it.

- `crates/core/tests/graceful_shutdown_exit_code.rs::graceful_shutdown_on_sigterm_exits_zero` spawns the real `freenet` binary (via `CARGO_BIN_EXE_freenet`, so it can never assert against a stale build), waits for a readiness marker that is only logged *after* the SIGTERM handler is installed, sends SIGTERM, and asserts `ExitStatus::code() == Some(0)` — plus that `signal()` is `None` (proving the handler ran, not the default disposition) and that the signal path actually executed (so a 0 cannot be credited to some other clean-exit path).
- `finish_run_maps_only_a_requested_graceful_stop_to_success` pins the truth table, including the case a live node cannot easily be driven into: an **unrequested** `GracefulShutdown` must stay non-zero.
- `graceful_classification_is_by_type_not_by_message` pins that the predicate keys on the error type, so an error whose `Display` merely reads "Graceful shutdown" cannot be laundered into a success exit.

**Revert-verified**: with `finish_run` restored to the pre-fix `result`, the end-to-end test fails with `Some(1)` (the exact production symptom) and the unit truth table fails on its first assertion. Both pass with the fix.

The node is spawned fully offline (`--skip-load-from-network`, `--disable-auto-update`), with `RUST_LOG` pinned on the child rather than inherited — CI's `test_unit` job exports `RUST_LOG=error`, which would otherwise filter out the readiness marker. A nextest override sets `retries = 0` for it: the subject of the test is a signal race, so a retry would be indistinguishable from the regression it exists to catch.

Reviewed by three independent blind lenses (correctness; supervisor/service-manager interaction; test quality). The correctness lens found the `ChannelCloseReason` conflation described above, which is why the `shutdown_requested` gate exists.

Addresses the main path of #5227, but does **not** fully close it — deliberately not using a closing keyword.

Adversarial review found a second, pre-existing path to the same symptom that this PR does not cover: the SIGTERM handler is installed at `bin/freenet.rs:366`, which `run_network` only reaches after `serve_client_api`, `NodeConfig::new` and `node_config.build()` (the redb open + contract-store load). A stop *during that startup window* kills the process under the default signal disposition; systemd then sets `$EXIT_STATUS` to the signal NAME, `service/linux.rs:539` matches its `*)` arm, and `rollback.rs` has no arm for signal names, so it falls through to `_ => StopClass::Crash`.

That window matters precisely because it coincides with probation: a node inside its post-update probation window has just restarted, so it is also near the front of its boot. The new end-to-end test structurally cannot see this, since it waits for the readiness marker specifically to get past that point.

#5227 stays open for that remainder.

[AI-assisted - Claude]

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -58,6 +58,18 @@ retries = 0
 filter = "test(test_ping_blocked_peers)"
 retries = 0
 
+# #5227 graceful-shutdown exit code. The whole subject of this test is a signal
+# race (SIGTERM must reach an installed handler and the process must exit 0), so
+# a retry would be indistinguishable from the regression it exists to catch —
+# exactly the flake-laundering the repo's flaky-tests rule forbids. It also boots
+# a real `freenet network` subprocess, so its budget (120s readiness + 45s exit)
+# needs more headroom than the 240s default; without this the harness SIGKILLs it
+# instead of letting the test emit its own log-attached diagnostic.
+[[profile.ci.overrides]]
+filter = "test(graceful_shutdown_on_sigterm_exits_zero)"
+retries = 0
+slow-timeout = { period = "300s", terminate-after = 2 }
+
 [profile.nightly]
 # Nightly tests run longer by design; no retries since failures need investigation.
 retries = 0
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -366,9 +366,18 @@ async fn run_network_node_with_signals(
     let mut sigterm = signal::unix::signal(signal::unix::SignalKind::terminate())
         .context("failed to install SIGTERM handler")?;
 
+    // Set when THIS process asks the node to stop because an operator signalled
+    // it (SIGTERM from `systemctl stop`, SIGINT from Ctrl+C). Load-bearing for
+    // the exit code: see `finish_run` — an "operator asked us to stop" exit is a
+    // SUCCESS, but the same `EventLoopExitReason::GracefulShutdown` value is
+    // ALSO produced by a fault (a critical internal channel dying, see
+    // `p2p_protoc::ChannelCloseReason`), which must keep failing loudly.
+    let shutdown_requested = Arc::new(std::sync::atomic::AtomicBool::new(false));
+
     // Spawn a task to listen for shutdown signals and trigger graceful shutdown
     let signal_task = {
         let shutdown_handle = shutdown_handle.clone();
+        let shutdown_requested = Arc::clone(&shutdown_requested);
         GlobalExecutor::spawn(async move {
             #[cfg(unix)]
             let shutdown_reason = tokio::select! {
@@ -383,6 +392,9 @@ async fn run_network_node_with_signals(
             };
 
             tracing::info!(reason = shutdown_reason, "Initiating graceful shutdown");
+            // Record the request BEFORE making it, so the flag is always visible
+            // by the time the event loop can observe the shutdown and return.
+            shutdown_requested.store(true, std::sync::atomic::Ordering::SeqCst);
             shutdown_handle.shutdown().await;
         })
     };
@@ -899,11 +911,62 @@ async fn run_network_node_with_signals(
     // and complete their cleanup without being forcefully killed by SIGKILL.
     tokio::time::sleep(std::time::Duration::from_secs(2)).await;
 
+    let result = finish_run(
+        result,
+        shutdown_requested.load(std::sync::atomic::Ordering::SeqCst),
+    );
+
     if result.is_ok() {
         tracing::info!("Graceful shutdown complete");
     }
 
     result
+}
+
+/// Decide what a finished node run means for the PROCESS EXIT CODE (#5227).
+///
+/// `run_network_node` reports a clean SIGTERM/SIGINT stop as
+/// `Err(EventLoopExitReason::GracefulShutdown)` — a typed sentinel, not a real
+/// error. `main` used to hand that straight to its `eprintln!("Error: …")` +
+/// `std::process::exit(1)` fallback, so a deliberate `systemctl stop` was logged
+/// as `status=1/FAILURE`; and because `rollback::classify_stop` treats any
+/// status outside {0, 2, 42, 43, 44} as a crash, three clean stops of a freshly
+/// updated node inside its post-update probation window were enough to trigger
+/// an auto-rollback of a perfectly healthy release (#4073).
+///
+/// Two conditions are required to map that to a success exit, and the second is
+/// the load-bearing one:
+///
+/// 1. `listener_exit_is_graceful(&err)` — the SAME predicate the #4549
+///    fatal-listener force-exit path uses, so the two cannot disagree about
+///    which errors are candidates.
+/// 2. `shutdown_requested` — THIS process asked the node to stop, because an
+///    operator signalled it.
+///
+/// (1) alone is NOT sufficient, and this is the whole reason the flag exists.
+/// `p2p_protoc` raises the very same `GracefulShutdown` when a critical internal
+/// channel dies (`ChannelCloseReason::{Bridge, Controller, Notification,
+/// OpExecution}`) — a node-fatal fault that logs `CRITICAL: Channel closed …`.
+/// `client_events.rs` likewise sends the identical `NodeEvent::Disconnect` when
+/// every client transport has died, a degraded state its own comments call
+/// "restart recommended". Neither is distinguishable from an operator stop at the
+/// p2p layer, so this flag is the sole discriminator. Both must keep a non-zero
+/// exit: on Linux that is what still fires the unit's `ExecStopPost` self-heal
+/// and counts the crash toward probation, and on the macOS/Windows wrappers an
+/// exit 0 is read as "normal shutdown" and would leave the node DOWN rather than
+/// relaunching it.
+///
+/// The auto-update exit is untouched: it surfaces as `UpdateNeededError`, not an
+/// `EventLoopExitReason`, so it still exits 42 and still fires the supervisor's
+/// `freenet update` hook. In the rare race where an update is detected in the
+/// same instant as a SIGTERM, the unbiased `select!` above may resolve the node
+/// arm and the exit is a clean 0 — the operator asked us to stop, so stopping
+/// wins, and the update is picked up by `startup_update_check` on the next boot.
+fn finish_run(result: anyhow::Result<()>, shutdown_requested: bool) -> anyhow::Result<()> {
+    match result {
+        Err(e) if shutdown_requested && freenet::listener_exit_is_graceful(&e) => Ok(()),
+        other => other,
+    }
 }
 
 /// Exit code when another freenet instance is already running.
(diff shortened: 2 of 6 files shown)


sanity:
## Review summary

Reviewed at `e922bb45c` with three independent lenses, each blind to the others. All 13 CI checks green at that SHA, including Windows Service Unit and macOS Service Unit.

### Testing lens — adequate

Both halves of the two-sided contract are covered, which was my main worry going in:

- **Clean stop exits 0**: `tests/graceful_shutdown_exit_code.rs` spawns the real binary, waits on a readiness marker, sends SIGTERM, asserts `code() == Some(0)` and `signal() == None`.
- **Fault still exits non-zero**: `finish_run_maps_only_a_requested_graceful_stop_to_success` asserts an *unrequested* `GracefulShutdown` stays `Err`. This is the half that keeps a genuinely dead node from looking healthy, and it is the case a live node cannot easily be driven into, so a unit test on the decision function is the right instrument.

Readiness was verified against the real call graph rather than the test's own comment: the marker is logged from `run_network_node`, which is only reached after the SIGTERM handler is installed at `freenet.rs:366`, so the test cannot signal before the handler exists. Three plausible weakened-fix mutations (identity, broadening to `Err(_) => Ok(())`, dropping the `shutdown_requested` conjunct) are each caught by a distinct assertion. `retries = 0` is correct and correctly scoped to the one signal-race test.

### Deployment lens — safe to ship, one tracked desktop regression

Walked every consumer of the exit code:

- **`freenet service stop`, sleep/wake, service-manager stop**: no regression on any platform. macOS `launchctl bootout` and Windows `taskkill /f` both take the wrapper down with the node.
- **The updater path is untouched**: the node exits 42 by itself; `finish_run` only rewrites `GracefulShutdown`, and `UpdateNeededError` is a different type that falls through.
- **systemd**: `Restart=always` restarts on exit 0 too, `RestartPreventExitStatus=43` does not list 0, and `StartLimit*` counts start attempts rather than exit statuses. So neither restart behaviour nor the start-limit accounting changes.
- **`classify_stop`**: `"0" | "2" | "43" | "44" => NotCrash` confirmed.
- Gateway units, release-agent, `gateway-update.yml`, Docker, and deploy scripts all unaffected (none read the node's exit status).

One real regression, desktop-only: a SIGTERM aimed at the node child while the wrapper survives (`pkill -f "freenet network"`) previously relaunched and now exits. This needs an unsupported stop path, and is the gap already tracked as **#5228** — worth noting that this PR makes #5228 more load-bearing, since it introduces the first routine way for the wrapper to reach its terminal exit arm. On the modern macOS tray install there is no launchd safety net (`RunAtLoad=true` with `KeepAlive=false`), which is worth adding to #5228.

Also identified a narrow laundering window: a SIGTERM arriving while the node is already unwinding from a channel-death shutdown would be scored as a requested stop. Cost is one skipped self-heal and one uncounted probation crash, in a case where the operator did ask to stop. Not worth a code change.

### Adversarial lens — did not report

Commissioned and did not return findings before merge. Recording it rather than treating silence as a clean result. The two questions it was asked (whether the `shutdown_requested` flag can launder a fault into exit 0 in a wider window, and whether any producer of `GracefulShutdown` exits the wrong way) were both covered by the deployment lens, which enumerated the producers and found the one narrow window noted above.

### Corrections made

- Fixed a factual error in the PR description: it claimed this removes an `ExecStopPost` race for the gateway auto-update path. The gateway fleet runs `scripts/freenet-gateway.service`, which has **no `ExecStopPost`** (only `Restart=always`), so that race never existed there. The claim holds only for units generated by `freenet service install`.

### Field validation

Verified the mechanism on a real NATed peer using the genuine installer path, not synthetic files:

```
install (freenet update --force)  -> probation armed, crash_count=1
clean stop inside the 60s window  -> crash_count=2
restart, wait 90s                 -> committed and cleared
```

This surfaced a detail the issue did not capture: **an install begins probation with a strike already spent**, because the installer's own service restart exits 1 and is counted. So an updated node started at 1 of 3, and only two further restarts were needed to roll it back. This PR removes that.

It also falsified #5232, which I had filed claiming probation is never committed. It commits correctly, at 60s, including across a restart. #5232 has been corrected and re-scoped; 0.2.123 ships this PR alone.

[AI-assisted - Claude]

sanity:
## Adversarial lens — reported after the summary above; no blocker, ship it

Correcting my earlier note that this lens did not report: it did, and it found the most important thing in the review. Recording it here in full.

### The core is right, and for a verifiable reason

The store at `freenet.rs:397` is SeqCst and strictly precedes `shutdown_handle.shutdown()` at `:398`, and that shutdown is *what causes* the event loop to return. So store → send Disconnect → recv → break → return → load at `:916` is a genuine happens-before chain: the flag can never be stale-false on a real stop, and `abort()` cannot un-store it.

Every producer of `GracefulShutdown` was enumerated against the code rather than the PR text. `EventLoopExitReason` has exactly two variants, all producers funnel through `p2p_protoc.rs:2897`, and the flag is set in exactly two places:

| Producer | Sets flag | Exit | Correct |
|---|---|---|---|
| `ShutdownHandle::shutdown` from the signal task (SIGTERM/SIGINT) | yes | 0 | yes — the bug being fixed |
| `ChannelCloseReason::{Bridge,Controller,Notification,OpExecution}` (all four variants) | no | non-zero | yes — node-fatal |
| `client_events.rs:304`, every client transport lost | no | non-zero | yes |
| Auto-update path, returns `UpdateNeededError` | no | 42 | yes — sentinel never surfaces |

One widening worth recording: `ShutdownHandle::shutdown` (`node.rs:166-192`) sets its flag and then drains for up to `drain_timeout` (default **30s**) before sending Disconnect. So the false-true window after a SIGTERM is seconds, not microseconds. The conclusion is unchanged — the operator asked us to stop, and we are stopping — but the duration is worth knowing.

The `listener_exit_is_graceful` coupling is safe because it is pinned from **both** sides (`p2p_impl.rs:1064` for the #4549 force-exit reading, `freenet.rs:1412` for the exit-code reading), so a future narrowing edit breaks a test rather than silently changing the other consumer.

### Finding: #5227 is not fully closed (see updated PR description)

Detailed in the PR body above. Summary: the SIGTERM handler is installed after the heavy startup work, so a stop during startup dies under the default disposition and is classified as a crash from the signal name. Pre-existing and unchanged by this PR, which strictly improves matters — but the PR previously read as if #5227 were closed. The closing keyword has been removed and #5227 stays open.

Two candidate follow-ups, cheap and independent:
1. Install the signal handler before the heavy build.
2. Gate the unit's `ExecStopPost` on systemd's **`$SERVICE_RESULT`** (which is `success` for a stop job however the process died) rather than on `$EXIT_STATUS` alone. This would have fixed the original incident with no node change at all, and covers the startup window for free.

### Follow-ups filed separately

- **Unpinned invariant.** The entire fault/stop discrimination rests on a single `store` at `freenet.rs:397`, and both new tests take `shutdown_requested` as a *parameter*. An edit that also sets it elsewhere — the tempting one being the auto-update arm at `:891`, which likewise calls `shutdown()` — would reopen fault-laundering with a fully green suite. A source-scrape pin asserting exactly one `shutdown_requested.store(`, inside the signal task, is three lines and is the highest-value follow-up here.
- **CI flake vector.** `graceful_shutdown_exit_code.rs:76-79` reserves ports bind-then-drop while `test_unit` runs the workspace in parallel; a collision kills the child before the readiness marker, which is a hard failure under `retries = 0`. The comparable `test_ping_blocked_peers` override is only safe without retries because it has an in-process retry loop with jitter, which this test lacks. Fix shape: bounded retry of the *spawn* with fresh ports when the child exits pre-readiness, keeping `retries = 0` for everything after SIGTERM is sent.
- **#5228 additions.** `service/wrapper.rs:1284-1294` returns on `WrapperAction::Exit` without sending `status_tx`, so on the new exit-0 path the tray icon vanishes rather than showing a stopped state. Also `launch_at_login.rs:160-163` sets `RunAtLoad=true` with `KeepAlive=false`, so the modern macOS tray install has no launchd safety net at all.
- **Informational.** `scripts/update-remote-gws.sh:406`'s unconditional `systemctl reset-failed` existed to clear the failed state this bug produced; it becomes a no-op and its comment is now misleading.

### On revert

`Err(GracefulShutdown)` downcasts to neither `UpdateNeededError` nor `AlreadyRunningError`, so `main` falls to `freenet.rs:1387-1388` and `status.code()` is `Some(1)` — the e2e test does fail without the fix. Note this lens did not run cargo (read-only), so that remains static reasoning.

[AI-assisted - Claude]

Issue #4118, opened 2026-05-13
release.yml: use RELEASE_PAT so release.published triggers downstream workflows


sanity:
## Problem

The v0.2.57 release cascade required manual intervention because GitHub Actions' GITHUB_TOKEN suppresses downstream workflow triggers (anti-recursion safeguard). Two concrete failures:

1. **The bump PR (#4116) had 0 check-runs.** `release.yml` creates the PR via `gh pr create` using `GITHUB_TOKEN`. No CI fired. Had to `gh pr close && gh pr reopen` to wake the workflows — and that broke `wait_for_pr` polling (see #2 below).

2. **`release.published` event didn't fire `gateway-update.yml` or `release-announce.yml`** after `cross-compile.yml` undrafted the release. The undraft was performed by `gh release edit --draft=false` using `GITHUB_TOKEN`, so the resulting event was suppressed.

Workaround for v0.2.57: I triggered both workflows via `workflow_dispatch` manually. That's not zero-touch.

## Fix

Add a `RELEASE_PAT` repo secret (a PAT with `repo` + `workflow` scopes) and use it in:

- `release.yml` line 80, 142, 167 (already coalesces: `${{ secrets.RELEASE_PAT || secrets.GITHUB_TOKEN }}` — just needs the secret set)
- `cross-compile.yml` `attach-to-release` job (currently uses GITHUB_TOKEN for `gh release edit --draft=false`)

## Acceptance

Next release cuts via `gh workflow run release.yml --field version=X.Y.Z` and runs to completion without close+reopen or manual workflow_dispatch.

[AI-assisted - Claude]

PR #5235, opened 2026-08-08
ci: cut the release from one validated commit, not from moving main

### LINKED ISSUE #5233: release.yml: publish and tag check out 'ref: main', so releases ship whatever wins a race with the merge queue
## Problem

`release.yml` does not cut the release from the commit the workflow was launched
against. The `publish` and `tag` jobs both check out a *moving* reference:

```yaml
# .github/workflows/release.yml:250-252  (Publish to crates.io)
- uses: actions/checkout@v7
  with:
    ref: main
- name: Pull latest changes
  run: git pull origin main
```

```yaml
# .github/workflows/release.yml:305-307  (tag job)
- uses: actions/checkout@v7
  with:
    ref: main
```

The tag is then created on that checkout (`release.yml:347-357`,
`git tag -a "v$VERSION"`).

Between `workflow_dispatch` and those checkouts sits the `wait_for_pr` job,
which blocks until the version-bump PR merges — typically ~20 minutes. Anything
that lands on `main` during that window silently becomes part of the release.

So the released artifact is **whatever wins a race with the merge queue**, not
the validated SHA the run started from.

## It happened tonight (2026-08-07/08, release 0.2.122)

Run [31231116377](https://github.com/freenet/freenet-core/actions/runs/31231116377),
launched from `242affb1e`:

| Time (UTC) | Event |
|---|---|
| 00:49:06 | Release run launched, `headSha=242affb1e` |
| 00:49:10–00:49:16 | `Validate release inputs` |
| 00:49:24–00:49:58 | `Update versions and create PR` (opens the bump PR) |
| 00:50:00–01:09:18 | `Wait for PR to merge` (blocks ~19 min) |
| **01:08:45** | **Unrelated PR #5224 merges to main via the merge queue** |
| 01:09:20 | `Publish to crates.io` starts → `checkout ref: main` + `git pull origin main` |

Resulting `main`:

```
48c166c46  fix(diagnostics): report real node uptime ... (#5224)   <- HEAD at checkout
9c12b7961  build: release 0.2.122 (#5231)
242affb1e  fix(update): normalise the release tag ... (#5225)      <- what the run validated
```

0.2.122 was deliberately scoped to be **minimal** — the v-prefix updater fix and
nothing else — because 0.2.120/0.2.121 cannot auto-update (#5221), so every
operator must install this release by hand and a second manual install would be
very costly. #5224 was held back for exactly that reason. The hold could not be
enforced: the merge queue landed it before the release's checkout, and the
workflow picked it up with no signal that the release contents had changed.

The outcome was accepted this time (the change is a read-only diagnostic field
with no wire or protocol surface, and cancelling mid-publish would have burned
the version number and forced the second manual install the scope decision
existed to prevent). The mechanism is the problem, not this particular commit.

## Why this is worth fixing beyond the scope question

1. **"Cut a minimal release" is not achievable in principle.** No amount of
   coordination closes the window, because the window is defined by a job that
   waits on a human/queue-paced event.
2. **What was tested is decoupled from what ships.** `validate` runs against the
   launch SHA; publish and tag run against a later tree. The CI signal the
   release decision was made on does not describe the artifact.
3. **It is silent.** Nothing in the run log says "main moved under us". The
   release notes and the tag simply point at a different tree than intended.

## Suggested fix

Resolve `main` to a SHA once, in `validate` (after the bump PR merges, in
`wait_for_pr`), expose it as a job output, and have `publish` and `tag` check out
that exact SHA rather than `ref: main`. Drop the `git pull origin main` step,
which exists only to re-widen the same window.

A cheaper interim guard: after checkout, fail the job if `git rev-parse HEAD`
does not match the expected SHA, so a moved `main` aborts the release loudly
instead of shipping silently.

Related: #5222 (post-mortem for tonight), #5221 (the updater bug 0.2.122 fixes).

[AI-assisted - Claude]

sanity:
## Problem

`release.yml`'s `publish_crates` and `create_release` jobs checked out `ref: main` and then `git pull origin main`. Both run **after** `wait_for_pr`, which blocks ~20 minutes waiting for the version-bump PR to clear the merge queue. `main` is a moving reference, so anything that landed during that window silently became part of the published crates and of the release tag.

This is observed, not hypothetical. Release v0.2.122 ([run 31231116377](https://github.com/freenet/freenet-core/actions/runs/31231116377)) launched from `242affb1e`; PR #5224 merged at 01:08:45Z; the publish job checked out at 01:09:20Z and got `48c166c46`. The tag `v0.2.122` dereferences to `48c166c46`, while the version-bump PR #5231 merged as `9c12b7961` — the tag is one commit past the tree that carries the bump, and `cross-compile.yml` built the shipped binaries from it. That release was deliberately scoped to be minimal, because 0.2.120/0.2.121 cannot auto-update (#5221) and every operator has to install it by hand. The scope decision was load-bearing and the mechanism defeated it.

Three consequences: a minimal release is unachievable in principle, the CI signal describes a different tree than the one that ships, and it is completely silent.

## Solution

Resolve `main` to **one immutable SHA, once**, and have everything downstream check out that SHA explicitly.

`wait_for_pr` now ends by resolving the version-bump PR's **merge commit** (`scripts/resolve_release_sha.sh`) and exporting it as a job output. That commit is the right anchor rather than the launch SHA: it is the tree the merge queue's full-suite release gate actually validated, it contains the version bump, and later merges move `main` past it without changing it.

- The resolver **fails closed** — it never falls back to a moving reference — and **retries** (6 × 10s), because it runs *after* the bump PR has merged, so a single API blip would otherwise strand a half-done release (bump on main, nothing published or tagged). It distinguishes "the API could not be queried" from "the PR has no merge commit" so the error does not blame the PR for an API fault.
- It **warns loudly**, listing the commits, when `main` moved between the launch commit and the release commit. That is the "silent" consequence above. It is a warning rather than a hard failure because failing after the bump PR has already merged leaves an awkward recovery state — happy to make it hard-fail if you'd prefer. If the compare call itself fails it says so, rather than printing a false all-clear.
- `publish_crates` and `create_release` check out that SHA; both `git pull origin main` steps are gone. The release notes' `target_commitish` is pinned too — the tag does not exist yet at that point, so `main` there meant the notes could describe commits not in the release.
- `scripts/verify_release_checkout.sh` runs after each checkout and before anything irreversible. It checks HEAD against the pinned SHA **and** that the checked-out `crates/core/Cargo.toml` declares the release version. The second check is the load-bearing one: the SHA comparison is nearly a tautology (the same expression feeds `ref:`), whereas the version is derived independently, so it catches any wrong tree — including the `|| github.sha` fallback landing on the launch commit, which would publish a crate whose version disagrees with the tag.

`create_release` gains `wait_for_pr` in `needs` purely so the output is in scope; the ordering it implies was already enforced through `publish_crates`.

### What this does not fix

Pinning does **not** exclude the commits that raced in. The merge queue builds the version bump on top of whatever landed first, so those commits are ancestors of the pinned SHA either way. What changes is narrower: publish, tag and the release notes now agree on **one immutable commit** instead of re-resolving a moving branch three times, and the divergence is **reported** instead of silent — the run log names every commit that landed between launch and release. Genuinely excluding racing commits would need a queue-level hold, which is a different change.

### Bundled: the #4118 token-coalesce guard has never worked

`check-token-coalesce.yml` filtered diff headers with `grep -v '^\+\+\+'`. In GNU BRE `\+` is the one-or-more operator, so that matches **every** line starting with `+` — the guard's input was always empty and it has flagged nothing since it was written. Verified against GNU grep 3.11 (the ubuntu-latest version): the old filter misses a bare `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` line, the fixed one catches it.

This is bundled rather than split because this PR adds two new bare-`GITHUB_TOKEN` lines and rests on that guard to police them; claiming the guard passes while knowing it cannot fail would be asserting a green signal already shown to be meaningless. The job now self-tests before checking the real diff, using the same two detector functions the real check uses — sharing only the line-filter was not enough, because duplicated regexes meant breaking the real one still left the self-test green. The self-test asserts both detectors fire on bare tokens *and* stay quiet on coalesced and `coalesce-exempt` lines; five mutations of the guard's own logic, including the original BRE bug, all turn it red.

Worth recording: **the self-test failed on its first CI run, correctly.** `${{ secrets.GITHUB_TOKEN }}` written literally inside a `run:` block is not a literal — Actions evaluates it before the shell sees it, so the fixture became the real token value and stopped resembling the pattern. The guard refused to certify itself rather than reporting a meaningless pass. Fixed by assembling the fixture so the raw YAML carries no expression marker.

Arming it also required exempting two pre-existing read-only lookups (`binstall-smoke-test.yml`, and the guard's own doc comment) that would otherwise false-positive on any future PR touching those lines.

## What is and is not verified

Being explicit, because this workflow is how we ship.

**Verified:**
- `scripts/release_mergequeue_test.sh` (already run by `ci.yml`) goes from 22 to 101 checks and passes. Job-scoped pins use `sed` ranges that now assert they terminated on their closing anchor, so renaming a job cannot silently widen a range to EOF.
- **All 22 mutations tested turn the suite red**, including every one two review rounds found it had missed: commenting the guard out rather than deleting it, *moving* the guard to after `cargo publish` / after the tag push, and re-pointing the checkout with `git reset --hard` / `git checkout main` / `git merge` rather than `git pull`. The harness asserts each mutation actually changed the tree, so a `sed` that fails to apply is reported rather than counted as a pass.
- Behavioural tests drive both new scripts with a mocked `gh` that asserts the real invocation shape (`--json mergeCommit`, the jq filter) and models real `gh` faithfully — an unmerged PR exits 0 printing nothing, which I verified against a live PR; modelling it as an error had left the "PR has no merge commit" branch with no coverage.
- `actionlint` clean, and shown **non-vacuous**: pointing `needs.wait_for_pr.outputs.release_sha` at a nonexistent output name makes it error at all four sites, which is what actually proves `create_release` can read the new output.
- `shellcheck` clean on all three scripts.
- The `RELEASE_PAT` coalesce is unchanged on every checkout and `gh` step; both new `gh` steps are read-only and carry `# coalesce-exempt:` markers. I simulated the (fixed) guard against this PR's own diff: it passes.

**Not verified:**
- **This cannot be exercised end to end without cutting a real release, and I did not cut one.** The job wiring is reasoned and statically checked, not run. Specifically unexercised: that `actions/checkout` resolves a bare merge-commit SHA at default `fetch-depth` on the first attempt; that `gh pr view --json mergeCommit` is populated promptly after a *merge-queue* merge rather than lagging (the retry loop exists for exactly this and is itself untested against real lag); and that `generate-notes` accepts a full SHA as `target_commitish`.
- The shipped retry budget (6 × 10s) is overridden in every test, so the real wait is untested. It fails closed.
- The `|| github.sha` fallback is reachable only on the dry-run-with-failed-`wait_for_pr` path. It now fails there with an explicit "wait_for_pr did not resolve a release commit" error, where previously that dry run passed. That is a deliberate behaviour change: a dry run sitting on the pre-bump commit is not dry-running the release.

## Other moving-reference patterns

Surveyed the rest of the cascade; details in a comment below. `cross-compile.yml`, `gateway-update.yml`, `release-announce.yml` and `binstall-smoke-test.yml` are **not** affected. `scripts/release.sh` (the legacy local path) has the same bug and is **not** fixed here — see the comment for why and the follow-up issue.

Closes #5233

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/binstall-smoke-test.yml
+++ b/.github/workflows/binstall-smoke-test.yml
@@ -92,7 +92,7 @@ jobs:
         id: release_ready
         shell: bash
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # coalesce-exempt: read-only release lookup, emits no event
         run: |
           set -euo pipefail
           TAG="v${{ steps.versions.outputs.freenet }}"
--- a/.github/workflows/check-token-coalesce.yml
+++ b/.github/workflows/check-token-coalesce.yml
@@ -72,7 +72,17 @@ jobs:
 
           # Self-test: a known-bad line MUST be flagged. Runs the same
           # function the real check uses, so the two cannot drift.
-          SELF_TEST_BAD='+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}' # coalesce-exempt: this is the fixture the self-test feeds in, not a real token usage
+          #
+          # The fixture is assembled from a variable holding a bare `$` on
+          # purpose. Writing the two-brace expression literally here does NOT
+          # produce a literal: Actions evaluates expressions inside `run:`
+          # blocks before the shell sees them, so the fixture would become the
+          # real token value and the self-test would fail — it did exactly
+          # that on the first CI run of this change, which is the self-test
+          # doing its job. Splitting the `$` from the braces keeps the raw
+          # text free of the expression marker.
+          D='$'
+          SELF_TEST_BAD="+          GH_TOKEN: ${D}{{ secrets.GITHUB_TOKEN }}"
           SELF_TEST_OUT=$(printf '%s\n+++ b/.github/workflows/x.yml\n' "$SELF_TEST_BAD" | added_lines)
           if ! echo "$SELF_TEST_OUT" | grep -qE 'GH_TOKEN:\s*\$\{\{\s*secrets\.GITHUB_TOKEN'; then
             echo "::error::Self-test failed: this guard cannot detect a bare GITHUB_TOKEN line, so its result is meaningless. Fix added_lines() before trusting this job."
--- a/.github/workflows/check-token-coalesce.yml
+++ b/.github/workflows/check-token-coalesce.yml
@@ -70,29 +70,57 @@ jobs:
             grep -E '^\+' | grep -vE '^\+\+\+' | grep -vE 'coalesce-exempt' || true
           }
 
-          # Self-test: a known-bad line MUST be flagged. Runs the same
-          # function the real check uses, so the two cannot drift.
-          #
-          # The fixture is assembled from a variable holding a bare `$` on
+          # The two detectors, defined ONCE and used by both the self-test and
+          # the real check below. Sharing only `added_lines` was not enough:
+          # while the regexes were duplicated, breaking the real one left the
+          # self-test green — the same "the check cannot fail" bug this guard
+          # exists to catch, one layer up.
+          flags_bare_gh_token() {
+            grep -qE 'GH_TOKEN:\s*\$\{\{\s*secrets\.GITHUB_TOKEN'
+          }
+          flags_bare_checkout_token() {
+            grep -qE 'token:\s*\$\{\{\s*github\.token'
+          }
+
+          # Fixtures are assembled from a variable holding a bare `$` on
           # purpose. Writing the two-brace expression literally here does NOT
           # produce a literal: Actions evaluates expressions inside `run:`
-          # blocks before the shell sees them, so the fixture would become the
+          # blocks before the shell sees them, so a fixture would become the
           # real token value and the self-test would fail — it did exactly
           # that on the first CI run of this change, which is the self-test
           # doing its job. Splitting the `$` from the braces keeps the raw
           # text free of the expression marker.
           D='$'
-          SELF_TEST_BAD="+          GH_TOKEN: ${D}{{ secrets.GITHUB_TOKEN }}"
-          SELF_TEST_OUT=$(printf '%s\n+++ b/.github/workflows/x.yml\n' "$SELF_TEST_BAD" | added_lines)
-          if ! echo "$SELF_TEST_OUT" | grep -qE 'GH_TOKEN:\s*\$\{\{\s*secrets\.GITHUB_TOKEN'; then
-            echo "::error::Self-test failed: this guard cannot detect a bare GITHUB_TOKEN line, so its result is meaningless. Fix added_lines() before trusting this job."
+          BAD_GH="+          GH_TOKEN: ${D}{{ secrets.GITHUB_TOKEN }}"
+          BAD_CHECKOUT="+          token: ${D}{{ github.token }}"
+          GOOD_GH="+          GH_TOKEN: ${D}{{ secrets.RELEASE_PAT || secrets.GITHUB_TOKEN }}"
+          EXEMPT_GH="${BAD_GH} # coalesce-exempt: self-test fixture"
+
+          self_test_diff() { printf '%s\n+++ b/.github/workflows/x.yml\n' "$1"; }
+          fail_self_test() {
+            echo "::error::Self-test failed: $1 This job's result would be meaningless — fix it before trusting a pass."
             exit 1
+          }
+
+          # Must FIRE on the bad forms...
+          if ! self_test_diff "$BAD_GH" | added_lines | flags_bare_gh_token; then
+            fail_self_test "a bare secrets.GITHUB_TOKEN line is not detected."
+          fi
+          if ! self_test_diff "$BAD_CHECKOUT" | added_lines | flags_bare_checkout_token; then
+            fail_self_test "a bare github.token checkout line is not detected."
+          fi
+          # ...and stay QUIET on the good and the exempted ones, or every PR
+          # that touches a workflow fails for no reason.
+          if self_test_diff "$GOOD_GH" | added_lines | flags_bare_gh_token; then
+            fail_self_test "a correctly coalesced RELEASE_PAT line is reported as bad."
+          fi
+          if self_test_diff "$EXEMPT_GH" | added_lines | flags_bare_gh_token; then
+            fail_self_test "the '# coalesce-exempt:' marker is not honoured."
           fi
           if printf '+++ b/.github/workflows/x.yml\n' | added_lines | grep -q .; then
-            echo "::error::Self-test failed: this guard is not stripping the '+++' diff header."
-            exit 1
+            fail_self_test "the '+++' diff header is not being stripped."
           fi
-          echo "Self-test passed: the guard detects a bare GITHUB_TOKEN line."
+          echo "Self-test passed: both detectors fire on bare tokens and stay quiet on coalesced and exempted lines."
 
           ERRORS=()
           for f in $CHANGED; do
@@ -105,10 +133,10 @@ jobs:
             # workflow file is flagged for review. Suppress via inline comment
             # `# coalesce-exempt: <reason>` on the same line.
             ADDED=$(git diff "$BASE_SHA"..."$HEAD_SHA" -- "$f" | added_lines)
-            if echo "$ADDED" | grep -qE 'GH_TOKEN:\s*\$\{\{\s*secrets\.GITHUB_TOKEN'; then
+            if echo "$ADDED" | flags_bare_gh_token; then
               ERRORS+=("$f: new step uses bare \`secrets.GITHUB_TOKEN\` — should coalesce on \`RELEASE_PAT\` (see AGENTS.md → Release Workflow). Add `# coalesce-exempt: <reason>` on the line if intentionally read-only and event-quiet.")
             fi
-            if echo "$ADDED" | grep -qE 'token:\s*\$\{\{\s*github\.token'; then
+            if echo "$ADDED" | flags_bare_checkout_token; then
               ERRORS+=("$f: new actions/checkout uses bare \`github.token\` — same coalesce rule applies for any subsequent \`git push\`.")
             fi
           done
--- a/.github/workflows/check-token-coalesce.yml
+++ b/.github/workflows/check-token-coalesce.yml
@@ -56,6 +56,34 @@ jobs:
           # of a tag, `gh pr create`, `gh pr reopen`, or `gh workflow run` —
           # in jobs/steps that reference `${{ secrets.GITHUB_TOKEN }}` or
           # `${{ github.token }}` without the coalesce.
+          # Added lines of a diff, minus the `+++` file header, minus anything
+          # explicitly exempted.
+          #
+          # The `+++` filter MUST be an ERE (`grep -vE`). As a BRE, `\+` is
+          # GNU's one-or-more operator, so `^\+\+\+` matches EVERY line
+          # starting with `+` and this function returned nothing at all —
+          # which meant this guard silently flagged nothing from the day it
+          # was written until 2026-08-07. Verified against GNU grep 3.11, the
+          # version on ubuntu-latest. The self-test below now fails if that
+          # ever regresses, so the guard can never go quietly blind again.
+          added_lines() {
+            grep -E '^\+' | grep -vE '^\+\+\+' | grep -vE 'coalesce-exempt' || true
+          }
+
+          # Self-test: a known-bad line MUST be flagged. Runs the same
+          # function the real check uses, so the two cannot drift.
+          SELF_TEST_BAD='+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}' # coalesce-exempt: this is the fixture the self-test feeds in, not a real token usage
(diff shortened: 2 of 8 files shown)


sanity:
## Review summary

Three independent blind lenses were run against the checked-out branch, plus a survey of the rest of the release cascade. All findings are addressed below; none were dismissed.

**Lenses:** (1) *does this break the release cascade* — the failure mode specific to this change; (2) *testing / vacuity* — assume a test cannot fail and find it; (3) *skeptical bug hunt* — read the code before the description, assume a bug exists. Plus a read-only survey of `cross-compile.yml`, `gateway-update.yml`, `binstall-smoke-test.yml`, `release-announce.yml` and `scripts/release.sh`.

### Blocking findings, all fixed

Lenses 2 and 3 independently concluded the **pins were vacuous**, and proved it by mutation:

| Regression the pin names | Suite before | Now |
|---|---|---|
| Guard **commented out** rather than deleted (`# TODO: re-enable bash scripts/verify_release_checkout.sh`) | green | red |
| Resolver step commented out | green | red |
| `target_commitish` reverted with the old line left as a comment | green | red |
| Guard **moved to after `cargo publish`** | green | red |
| Guard **moved to after the tag push** | green | red |
| `git fetch origin main && git reset --hard origin/main` | green | red |
| `git checkout main` | green | red |
| `git merge origin/main` | green | red |
| `ref: 'main'` / `ref: refs/heads/main` | green | red |
| `publish_crates` renamed, widening `WAIT_JOB`'s sed range to EOF | green | red |

Every needle is now anchored to an executable line, both jobs assert the guard's position *relative to the checkout and to that job's irreversible step*, and the moving-reference check covers the re-pointing forms rather than one spelling. **All 20 mutations now turn the suite red**, verified in a copied tree.

Lens 1 also found the guard itself was **a tautology**: `ref:` and `RELEASE_SHA` came from the same expression, so `actions/checkout` guaranteed the match. It now also asserts the checked-out manifest declares `needs.validate.outputs.version` — an oracle derived independently of the SHA, which is the only thing that can catch the `|| github.sha` fallback landing on the pre-bump launch commit.

### Behaviour bugs found and fixed

- **Single-shot resolve could strand a half-done release.** The step runs *after* the bump PR has merged; one transient API failure meant version bump on main, nothing published or tagged. Now retries, and distinguishes "API unreachable" from "PR has no merge commit" so the error stops blaming the PR for an API fault.
- **A failed `compare` call printed "✅ main did not move".** #5233's complaint was a silent divergence; a *false* all-clear is worse because it ends the investigation. Now reports the check as inconclusive.
- **`grep -m1 … | cut` aborted under `pipefail`** when the manifest had no readable version, killing the script before the message explaining why — a blank failed step mid-release.
- **The warning counted the version bump as an extra**, so two raced-in commits read as "3 commits".
- **`RESOLVE_ATTEMPTS=0`/non-numeric** ran the loop zero times and blamed the PR for a config error.

### Test-fidelity findings

- The `gh` mock modelled an unmerged PR as a **non-zero exit**; real `gh` exits 0 printing nothing (verified against a live PR). Every test for the "PR has no merge commit" branch was therefore exercising the *API-failure* branch, leaving the branch the retry loop exists for with no coverage. Fixed, and the two reasons are now asserted to stay distinct.
- The mock now asserts the real invocation shape (`--json mergeCommit`, the jq filter), so a typo in the field name turns the suite red instead of only failing closed mid-release.
- The verifier's version grep is now checked against the **real** `crates/core/Cargo.toml`, not only a synthetic fixture — a move to `version.workspace = true` would otherwise break every release at the guard with no CI warning.

### The bundled `check-token-coalesce.yml` fix, and what it caught

Lens 1 found the #4118 guard has **never worked**: `grep -v '^\+\+\+'` is a BRE where `\+` is the one-or-more operator, so it stripped every added line and the guard's input was always empty. Verified against GNU grep 3.11.

Worth recording: after fixing it, **the new self-test failed on its first CI run** — correctly. `${{ secrets.GITHUB_TOKEN }}` written literally inside a `run:` block is not a literal; Actions evaluates it before the shell sees it, so the fixture became the real token value. The guard refused to certify itself rather than reporting a meaningless pass. Fixed by assembling the fixture so the raw YAML carries no expression marker.

Arming the guard also required exempting two pre-existing read-only lookups that would otherwise false-positive on any future PR touching those lines.

### Survey of the rest of the cascade

- `cross-compile.yml` — **not affected.** Every build job uses bare `actions/checkout@v7` on a tag push, which resolves to the immutable tag SHA and is fixed for the whole matrix run. `attach-to-release` and `verify-signing-key` have no checkout at all and scope everything to `${GITHUB_REF#refs/tags/}`.
- `gateway-update.yml` — **not affected.** Its one checkout is only used to source a decision-logic library; the version shipped comes from `github.event.release.tag_name`. Its `releases/latest` query is a deliberate correctness gate, not a race.
- `binstall-smoke-test.yml` — **not affected.** It reads crates.io "latest" on purpose; that is what the test is for.
- `release-announce.yml` — **not affected.** No checkout anywhere; works off the immutable tag name.
- `scripts/release.sh` — **same bug**, at `:705-707`, `:733-736`, `:914-915`, tagging at `:1093-1106`. **Not fixed here**: it is a stateful script with resumable state across invocations, so the SHA must be persisted to its state file rather than threaded through a job output — not the same-shape change, and outside this PR's scope. Filed as #5237. It is the legacy path (`docs/RELEASING.md:445`); the primary path is the workflow this PR fixes.

### Not fixed, by choice

The `|| github.sha` fallback is kept rather than removed: without it, an empty `ref:` makes `actions/checkout` default to `main`, which would reinstate the exact bug. It is now unreachable in a non-dry-run release by mechanism as well as by the two `if:` conditions, and fails with an explicit "wait_for_pr did not resolve a release commit" message on the one path that reaches it.

[AI-assisted - Claude]

sanity:
## Final review round

A fourth lens read the complete final content (the earlier three ran against code that has since changed substantially, so those reviews were stale). Verdict: **no blocking findings, safe to merge.** It independently re-ran the suite (101 checks), shellcheck, six mutations of the release pins and three of the coalesce guard, and confirmed the original BRE bug empirically on GNU grep 3.11.

Everything it raised is now fixed:

**`scripts/RELEASE_RECOVERY.md` still documented the bug.** `docs/RELEASING.md` was corrected earlier in this PR; the recovery doc was not — and it is the one an operator reaches for at exactly the moment this matters, when the workflow has failed mid-release and `main` has been moving for 20+ minutes. It said `git pull origin main` then `git tag`, plus `target_commitish=main` and a `git pull` before `cargo publish`. A human following it would have reintroduced #5233 by hand, with the two docs contradicting each other. All four sites now resolve the bump PR's merge commit, with a manifest sanity check. A scan confirms neither doc contains a moving-`main` recipe any more.

**The coalesce guard's error message deleted its own remediation.** The backticks around `` `# coalesce-exempt: <reason>` `` were unescaped inside a double-quoted string, so bash ran them as command substitution and the message rendered as `Add  on the line`. Nobody had noticed because the BRE bug meant that line had never once executed — this PR makes it reachable, so it is the first thing a blocked author will ever see. Fixed, and the guard's second prose line is now exempted too; a repo-wide scan shows **no unexempted hits remain**, so arming the guard cannot false-positive on an existing line.

**A zero-commit compare printed an all-clear.** `.commits == 0` means the launch commit is not an ancestor of the release commit (force-push, rewritten base), not "the bump was the only change" — a small false all-clear in the section whose entire purpose is avoiding those. Now reported as inconclusive. The compare call keeps `gh`'s stderr as well, matching the `pr view` call.

**A comment overstated what the merge queue validated.** The full release suite is gated on the merge-group head message starting with `build: release`, so it runs only when the bump is the last entry in its group. Reworded to "the commit the queue merged" rather than a guarantee about which suite ran.

**Three `grep -n` assignments could abort the test run.** Under `set -e` an empty result killed the script mid-suite, so 56 of 101 checks never ran and the "irreversible step was located" check could never print FAIL. The suite still went red, so this was misleading diagnostics rather than a false green, but it is fixed.

### One framing correction, worth stating plainly

Pinning does **not** exclude the commits that raced in. The merge queue builds the version bump on top of whatever landed first, so those commits are ancestors of the pinned SHA either way. What this change actually delivers is narrower than "a minimal release is now achievable":

1. publish, tag, and the release notes all agree on **one immutable commit**, instead of three independent re-resolutions of a moving branch; and
2. the divergence is **reported** instead of silent — the run log now names every commit that landed between launch and release.

Genuinely excluding racing commits would need a queue-level hold, which is a different change. The resolver's own warning text says this correctly ("a scope notice, not a correctness failure"); flagging it here so nobody expects #5233's *symptom* to be prevented outright.

### Deliberately not done

- **Ancestry check** (`git merge-base --is-ancestor "$RELEASE_SHA" origin/main`) before tagging. It would close the last theoretical "tag a commit not on main" path, but only `create_release` has the `fetch-depth: 0` to run it, so it would need guarding, and `mergeCommit` was empirically confirmed to be the base-branch commit across four real merge-queue merges in this repo. Adding a new failure mode to the release path for a path I could not make occur did not seem like a good trade.
- **`scripts/release.sh`** has the same defect at `:705-707`, `:733-736`, `:914-915`, tagging at `:1093-1106`. Filed as #5237 rather than fixed here: it is a stateful script whose resume paths mean the SHA must be persisted to its state file, so it is not the same-shape change. It is the legacy local path.

[AI-assisted - Claude]

Issue #5232, opened 2026-08-08
Post-update probation commit is invisible in the journal, and has no regression test


sanity:
## Summary

Post-update probation appears never to be committed, so a node stays probationary indefinitely after an update. Combined with #5227 (a clean stop exits 1 and is scored as a crash), every ordinary restart burns one of three strikes and is refused an update — and the third rolls the node back.

## Evidence

**Probation still active ~1 hour after install.** On a production node updated at ~17:00, a deliberate `systemctl --user stop` at 18:15 produced:

```
Freenet 0.2.121: crash 1/3 during post-update probation (node stop status 1);
not updating, will auto-roll-back if it keeps crashing.
```

`COMMIT_HEALTHY_UPTIME_SECS` is **60**, and the node had been up and healthy for roughly an hour. It should have committed long before.

**It fired more than once.** The same node's journal over an 8-hour window:

```
2  crash 1/3 during post-update probation
2  not updating
```

Both from ordinary clean stops. Nothing crashed. Note both are `1/3` — consistent with probation being re-entered rather than a counter advancing, which is itself worth understanding.

**Rollback is happening in the wild.** Independent telemetry analysis found 3 of 200 peers ever seen on 0.2.121 showing a version regression, two with identity verified (same peer_id, IP:port and ring location under both versions). One — `5Jdk42gGP9MTQVG2L` — started 0.2.121 at 17:14 and then emitted **only 0.2.120 from 18:30 through 23:30, 7,000+ records**. See freenet/freenet-core#5221 for the full analysis.

## Suspected cause (not investigated)

`commit_probation_task` in `crates/core/src/bin/freenet.rs` is a fire-and-forget `GlobalExecutor::spawn` that sleeps `COMMIT_HEALTHY_UPTIME_SECS` and then calls `commit_probation(&version)`. If that task never lands — dropped handle, cancelled at shutdown, or the version it commits not matching the marker's — probation is never cleared.

Worth checking specifically:

1. Does the spawned task actually run to completion on a healthy node? Its `JoinHandle` is aborted during shutdown (`commit_probation_task.abort()`), which is correct, but a node restarted inside the first 60s would never commit — and with #5227 that restart also counts as a crash.
2. Does `commit_probation(&version)` match the marker written by the installer? The installer stores a **stripped** version (`begin_probation(release.tag_name.trim_start_matches('v'), …)`), so a mismatch in format would leave the marker in place forever. This is adjacent to the #5104 class already fixed in #5225.
[...]

PR #5240, opened 2026-08-08
fix(rollback): announce the post-update probation commit on stderr (#5232)

sanity:
## Problem

#5232 reported that post-update probation is never committed — so a node stays probationary indefinitely after an update, every later clean stop is scored as a crash, and the third one rolls the node back off the release it just installed and pins that release as known-bad. A real user hit exactly that outcome on 0.2.122.

**`commit_probation` was never broken.** It fires at exactly 60s on every healthy boot, and this PR changes no rollback decision, threshold, or state transition. What was broken is that you could not see it from where the report was written.

Under the shipped systemd deployment the node's `tracing` output goes to the rolling log files under the log dir (the console layer is added only when stdout is a terminal, which it is not under a service manager), while the unit records `StandardOutput=journal` / `StandardError=journal`. The three other decisions in this flow — the crash count, the rollback, and rollback-unavailable — are all `eprintln!` in `commands::update`, so they reach the journal. The commit that disarms them did not. `journalctl -u freenet` therefore showed strikes accumulating against a probationary version and never once showed a commit.

Evidence that the commit does work, from the same node the issue was filed against, in `~/.local/state/freenet/freenet.*.log` and absent from that host's journal:

```
2026-08-05T13:53:29Z INFO freenet::commands::rollback: Auto-update probation passed ... version="0.2.119"
2026-08-05T17:50:40Z INFO ... version="0.2.120"
2026-08-07T23:25:54Z INFO ... version="0.2.121"
2026-08-08T02:43:00Z INFO ... version="0.2.122"
```

Four commits, one per version, each exactly 60s after that boot's start banner. Confirmed independently by running the real binary with an armed marker in a scratch `HOME`: the marker is removed at ~65s.

The issue's other two observations resolve the same way. The `1/3` seen twice was two different nodes: that host's boot banners show `23:15:18Z version="0.2.120"` and `23:24:54Z version="0.2.121"`, so the node that stopped was 0.2.120 while the `ExecStopPost` binary matching the marker was 0.2.121 — a hand-driven test bench with the binary swapped underneath, not a counter that resets. And the marker and `build_info::VERSION` strings match exactly (`0.2.121` both sides), so the `v`-prefix concern does not apply here.

## Solution

Announce both marker-clearing outcomes on stderr, alongside the decisions they cancel. One line per node start at most, and only when a probation marker exists. The wording lives in `commit_announcement`, a pure function over `CommitOutcome`, so both branches are unit-testable — `commit_probation` itself reads the process-global `$HOME` and cannot be driven from a test.

**Scope note, stated plainly because it goes beyond the reported symptom:** the `ClearedStale` branch is also escalated, from `debug!` (compiled out of release builds by `release_max_level_info`, so it was completely silent in the field) to `warn!` plus the stderr line. That branch *discards* a marker belonging to a different version **without committing anything**, so rollback protection for that version disappears. It is indistinguishable from a healthy commit by the state directory alone — both just delete the marker — and it is the branch a version-comparison regression would take, the #5104 `v`-prefix class.

One case is deliberately left alone and now documented rather than changed: a node parked by `freenet service disable` (`run_disabled_idle`) returns before the commit timer is spawned and keeps its marker until the 1h TTL retires it. Committing there would disarm rollback on evidence the mechanism does not have — that process never joined the network, so it has not demonstrated the health probation is testing for.

## Testing

`crates/core/tests/post_update_probation_commit.rs` spawns the real `freenet network` binary with its own `HOME` and covers both halves of the contract:

| Test | Asserts | Runtime |
|---|---|---|
| `probation_is_committed_after_a_healthy_uptime_window` | up past the window ⇒ marker gone, commit announced on stderr, and the following post-stop `freenet update` records no crash | ~65s |
| `probation_survives_a_stop_inside_the_commit_window` | stopped inside the window ⇒ marker retained, the stop IS counted `1/3`, and `crash_count` is persisted | ~15s |

Three deliberate choices, each load-bearing:

- **It drives the real binary.** The commit is not a property of `commit_probation_at`, which is already unit-tested and correct. It is a property of its only caller — a `GlobalExecutor::spawn` in `run_network_node_with_signals` that nothing awaits or monitors. A unit test on `commit_probation_at` passes happily while that task never runs. `commands::rollback` also lives in the bin crate behind a `$HOME`-derived state dir, so no in-process harness can reach it.
- **It asserts stderr, not the log files.** Asserting the log file would pass on a build where the journal is silent again, i.e. it would not hold this fix. (CI also sets `RUST_LOG=error`, which filters the `tracing::info!` out entirely.)
- **The second test is the positive control for the first.** "No crash was recorded" is a negative assertion on a subprocess's stderr, and the preceding assertion has already established the marker is gone — so on its own it could not fail, and would equally have passed with a misspelled `FREENET_POST_STOP_EXIT_CODE`, a `HOME` that did not relocate the state dir, an unparseable fixture, or a reworded needle. The in-window test forces the same fixture, env var and needle to produce a crash line. It is also the half that produced the user-visible rollback in #5232 (three fast restarts), and it makes no network call — the crash branch exits before `freenet update` probes GitHub.

Mutation-tested; each of these fails the suite:

| Mutation | Result |
|---|---|
| Delete the `commit_probation` call (the failure #5232 hypothesised) | FAILED |
| `v`-prefix version mismatch — marker removed as stale, never committed | FAILED |
| Revert this PR's `eprintln` (tracing-only, pre-fix state) | FAILED |

The second is the one that matters for test design: the marker *is* removed, so a file-existence check alone would have passed it while rollback protection was silently gone.

`rollback.rs::tests::both_marker_clearing_outcomes_are_announced` pins both announcement strings (and the silent `Nothing` case) in microseconds, so a reword fails there rather than after the 60s window.

nextest override: 300s period, `retries = 0`. Retrying brick-safety equipment would turn "the commit sometimes does not happen" — the exact bug class #5232 was filed about — back into a green run.

## A worked example of the failure mode this area keeps producing

Worth reading even if you skim the rest, because it happened *inside the fix*.

`commit_probation` announces "committed, auto-rollback disarmed". `remove_probation_at` discarded its `remove_file` result, so that line printed whether or not the marker was actually gone — and a failed unlink leaves rollback armed inside its 1h TTL. The fix for a misleading journal was itself capable of misleading the journal.

Fixing that introduced a second instance. The new `ClearFailed` message began *"post-update probation passed, but the marker could NOT be removed…"* — which **contains** the exact substring the end-to-end test greps for to conclude the node committed. A failed clear would have passed the guard while rollback was still armed: the same false-green, one level up, inside the test written to prevent it.

Both are now closed: `commit_probation_at` reports `ClearFailed` instead of a false `Committed`, the e2e keys on `"committed, auto-rollback disarmed"` which only the success branch can produce, and a unit test pins the two disjoint in both directions.

The pattern to watch for: **a success signal and a failure signal that share a substring, or a message emitted before the state change it describes.** Both produce a green check over a broken invariant.

## Review

Three independent lenses (skeptical, testing, code-first). No blocking findings; every should-fix is addressed in the second commit, including several assertions that could not fail, three comments that described things that were not true, a poll that raced the announcement, and the missing `ClearedStale` coverage.

## Not in scope

- **#5227** (a graceful shutdown exits 1, so it is scored as a crash) is the real cause of the user-visible rollback and is fixed separately in #5230. Rolling back needs three crashes recorded while the marker is alive, and the marker dies 60s into the first healthy boot — so it takes three stops each inside a 60s window, which is a fast restart loop, which is what that user's `StartLimitBurst` trip says they had.
- Nodes on ≤0.2.121 log `Startup update check: failed to parse latest version 'v0.2.121'`, so update detection is dead there and a node already rolled back onto one cannot auto-update out of it. Separate issue, separately owned. (The known-bad pin is not the trap: it is exact-version, so a newer release is never blocked by it.)

Refs #5232

[AI-assisted - Claude]

---

## Addendum: ejected by the merge queue, and the lint that caught this PR with its own construct

This PR sat green and `mergeable=CLEAN` for 27 days, was enqueued, and was **ejected** on a clippy error its branch CI never reported:

```
error: wildcard match will also match any future added variants
    --> crates/core/src/bin/commands/rollback.rs:1267:13
1267 |             other => panic!(
     = note: `-D clippy::wildcard-enum-match-arm` implied by `-D warnings`
```

Ruled out first: not a conflict (`git merge-tree --write-tree` against current main is clean, so no rebase was needed), not a toolchain bump (1.94.0 both sides), not a new lint (already configured at the merge-base). What changed is the lint's **reach** — #5292 extended lint coverage to test code after this branch last ran CI. The diff was never wrong; only the merged combination fails, so only a merge-group run or a local build with main merged in could have surfaced it.

**The irony is worth stating, because it is the argument for the lint.** `wildcard_enum_match_arm` exists in this repo so that adding an enum variant forces a compile error at every match site rather than silently falling through a wildcard. This is the PR that **adds** `CommitOutcome::ClearFailed` — and its own new test matched on `CommitOutcome` with `other =>`, the exact construct that would swallow the next variant someone adds. The lint caught the PR introducing a variant using the thing that would have hidden the following one.

Fixed by enumerating the three arms, not with `#[allow]`. An `#[allow]` would have silenced the lint and disarmed that guarantee for precisely the enum this PR extends, inside the test whose job is to tell the outcomes apart. With the arms enumerated, a fourth `CommitOutcome` variant breaks this match — which is the property the lint is there to preserve.

Verified with clippy against the **merged** tree, not the branch alone: a stale branch still carries old copies of files main has since rewritten, so a branch-only run tests a tree that will never exist after merge.

[AI-assisted - Claude]

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -80,6 +80,21 @@ slow-timeout = { period = "300s", terminate-after = 2 }
 filter = "test(test_placement_migration_at_scale_renewal_load_stays_bounded) | test(test_connection_growth_plateau_diagnostic) | test(test_connection_growth_stall_regression)"
 slow-timeout = { period = "150s", terminate-after = 2 }
 
+# The post-update probation tests (#5232) spawn the real `freenet` binary and
+# wait out the rollback machinery's own 60s commit window, which is a hard-coded
+# constant in the binary under test — so ~65s of the runtime is a deliberate
+# sleep, not slack, and the default 120s period leaves little room for a cold
+# debug-build node start on a loaded runner.
+#
+# retries = 0 for the same reason as the canary above: this is brick-safety
+# equipment, and an intermittent failure here is a finding, not noise to absorb.
+# A retry would turn "the commit sometimes does not happen" — the exact bug class
+# #5232 was filed about — back into a green run.
+[[profile.ci.overrides]]
+filter = "test(/probation_(is_committed|survives)/)"
+slow-timeout = { period = "300s", terminate-after = 2 }
+retries = 0
+
 [profile.nightly]
 # Nightly tests run longer by design; no retries since failures need investigation.
 retries = 0
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -70,6 +70,21 @@ filter = "test(graceful_shutdown_on_sigterm_exits_zero)"
 retries = 0
 slow-timeout = { period = "300s", terminate-after = 2 }
 
+# The post-update probation tests (#5232) spawn the real `freenet` binary and
+# wait out the rollback machinery's own 60s commit window, which is a hard-coded
+# constant in the binary under test — so ~65s of the runtime is a deliberate
+# sleep, not slack, and the default 120s period leaves little room for a cold
+# debug-build node start on a loaded runner.
+#
+# retries = 0 for the same reason as the canary above: this is brick-safety
+# equipment, and an intermittent failure here is a finding, not noise to absorb.
+# A retry would turn "the commit sometimes does not happen" — the exact bug class
+# #5232 was filed about — back into a green run.
+[[profile.ci.overrides]]
+filter = "test(/probation_(is_committed|survives)/)"
+slow-timeout = { period = "300s", terminate-after = 2 }
+retries = 0
+
 [profile.nightly]
 # Nightly tests run longer by design; no retries since failures need investigation.
 retries = 0
--- a/crates/core/src/bin/commands/rollback.rs
+++ b/crates/core/src/bin/commands/rollback.rs
@@ -486,19 +486,53 @@ fn remove_probation_at(dir: &Path) {
 /// Clear probation if the running version matches the marker (it has proven
 /// healthy). A marker for a DIFFERENT version is stale (e.g. left over after a
 /// rollback or external install) and is also removed.
+///
+/// Both outcomes DISARM rollback for the marker's version, so both are reported
+/// on stderr as well as through `tracing` (#5232). The node's `tracing` output
+/// goes to the rolling log files under the log dir, which systemd does NOT
+/// capture — only the process's stdout/stderr reach the journal. An operator
+/// (or an agent) reading `journalctl -u freenet` therefore saw the crash-count
+/// and rollback lines, which the installer prints on stderr, but never the
+/// commit that disarms them. #5232 was filed on exactly that asymmetry: the
+/// commit had been working all along and looked, from the journal, as though it
+/// had never once run. The other two decisions in this module are already on
+/// stderr; this puts the third one where the first two are.
+///
+/// The volume is bounded: at most one line per node start, and only when a
+/// probation marker actually exists.
 pub fn commit_probation(current_version: &str) {
     if let Some(dir) = state_dir() {
         match commit_probation_at(dir.as_path(), current_version) {
-            CommitOutcome::Committed => tracing::info!(
-                version = current_version,
-                "Auto-update probation passed: new version ran healthily for \
-                 {COMMIT_HEALTHY_UPTIME_SECS}s; committing (rollback disarmed)."
-            ),
-            CommitOutcome::ClearedStale { marker_version } => tracing::debug!(
-                running = current_version,
-                marker = %marker_version,
-                "Cleared stale auto-update probation marker for a different version."
-            ),
+            CommitOutcome::Committed => {
+                eprintln!(
+                    "Freenet {current_version}: post-update probation passed (ran healthily for \
+                     {COMMIT_HEALTHY_UPTIME_SECS}s); committed, auto-rollback disarmed."
+                );
+                tracing::info!(
+                    version = current_version,
+                    "Auto-update probation passed: new version ran healthily for \
+                     {COMMIT_HEALTHY_UPTIME_SECS}s; committing (rollback disarmed)."
+                );
+            }
+            CommitOutcome::ClearedStale { marker_version } => {
+                // Deliberately louder than the `debug!` this used to be. A
+                // release build compiles `debug!` out entirely, so this branch
+                // — which drops rollback protection for `marker_version`
+                // without committing anything — was completely silent in the
+                // field. It is the branch a version-comparison regression would
+                // take, and it is indistinguishable from a healthy commit by
+                // the state directory alone (both just delete the marker).
+                eprintln!(
+                    "Freenet {current_version}: discarded a post-update probation marker \
+                     belonging to {marker_version} (not the running version); auto-rollback \
+                     protection for {marker_version} is gone."
+                );
+                tracing::warn!(
+                    running = current_version,
+                    marker = %marker_version,
+                    "Cleared stale auto-update probation marker for a different version."
+                );
+            }
             CommitOutcome::Nothing => {}
         }
     }
--- a/crates/core/src/bin/commands/rollback.rs
+++ b/crates/core/src/bin/commands/rollback.rs
@@ -574,15 +574,25 @@ pub(crate) fn commit_announcement(
 /// evidence the mechanism does not have.
 pub fn commit_probation(current_version: &str) {
     if let Some(dir) = state_dir() {
-        // Announce AFTER the state transition, never before. `eprintln!` panics
-        // on a write error (EPIPE once the supervisor's reader is gone), and a
-        // panic in front of the removal would leave a healthy node's marker
-        // armed — turning an observability line into a rollback bug. In this
-        // order the worst case is a lost announcement.
+        // `eprintln!` is the ONLY fallible step here, so it goes last.
+        //
+        // It panics on a write error (EPIPE once the supervisor's reader is
+        // gone; a Windows child that inherited invalid standard handles), and a
+        // panic unwinds this whole task. So anything sequenced after it is lost
+        // when stderr is broken:
+        //
+        //  - Before the removal, a panic would leave a healthy node's marker
+        //    armed — an observability line turned into a rollback bug.
+        //  - Before the `tracing` record, a panic would take that record with
+        //    it, leaving the operator with NEITHER channel on exactly the node
+        //    whose stderr is already broken. That is strictly worse than before
+        //    this announcement existed, when the record always got written.
+        //
+        // Hence: transition, then log, then announce. The message is built up
+        // front because the `match` below consumes `outcome`; building a string
+        // cannot fail, so doing it early costs nothing.
         let outcome = commit_probation_at(dir.as_path(), current_version);
-        if let Some(line) = commit_announcement(&outcome, current_version) {
-            eprintln!("{line}");
-        }
+        let announcement = commit_announcement(&outcome, current_version);
         match outcome {
             CommitOutcome::Committed => tracing::info!(
                 version = current_version,
@@ -607,6 +617,9 @@ pub fn commit_probation(current_version: &str) {
                 "Probation passed but the marker could not be removed; rollback is STILL armed."
             ),
             CommitOutcome::Nothing => {}
+        }
+        if let Some(line) = announcement {
+            eprintln!("{line}");
         }
     }
 }
--- a/crates/core/src/bin/commands/rollback.rs
+++ b/crates/core/src/bin/commands/rollback.rs
@@ -483,56 +483,95 @@ fn remove_probation_at(dir: &Path) {
     let _rm = std::fs::remove_file(dir.join(PROBATION_FILE));
 }
 
+/// The operator-facing line for a marker-clearing outcome, or `None` when
+/// nothing happened. Split out from [`commit_probation`] so both branches are
+/// unit-testable: `commit_probation` itself reads the process-global `$HOME`
+/// through [`state_dir`] and so cannot be driven from a test.
+///
+/// ## Why these go to stderr and not only to `tracing` (#5232)
+///
+/// **This is the canonical explanation; the other comments point here.** Under
+/// the shipped systemd deployment the node's `tracing` output goes to the
+/// rolling log files under the log dir (`tracing::tracer::init_tracer` adds the
+/// console layer only when stdout is a terminal, which it is not under a service
+/// manager), while the unit records `StandardOutput=journal` /
+/// `StandardError=journal`. So `tracing` does not reach the journal and
+/// stdout/stderr does. The three other decisions in this flow — the crash count,
+/// the rollback, and rollback-unavailable — are all `eprintln!` in
+/// [`super::update`], so `journalctl -u freenet` showed strikes accumulating
+/// against a probationary version and never once showed the commit that disarms
+/// them. #5232 was filed reading exactly that asymmetry. This puts the
+/// marker-clearing decisions on the same channel as the decisions they cancel.
+///
+/// (The fallback tracer — used when no log dir resolves, or when
+/// `FREENET_LOG_TO_STDERR` is set — does write to stdout, so on those paths the
+/// `tracing` line reaches the journal too and the operator sees both.)
+///
+/// Volume is bounded: at most one line per node start, and only when a probation
+/// marker actually exists.
+pub(crate) fn commit_announcement(
+    outcome: &CommitOutcome,
+    current_version: &str,
+) -> Option<String> {
+    match outcome {
+        CommitOutcome::Committed => Some(format!(
+            "Freenet {current_version}: post-update probation passed (ran healthily for \
+             {COMMIT_HEALTHY_UPTIME_SECS}s); committed, auto-rollback disarmed."
+        )),
+        // Deliberately as loud as the commit. This branch drops rollback
+        // protection for `marker_version` WITHOUT committing anything, it is the
+        // branch a version-comparison regression would take, and it is
+        // indistinguishable from a healthy commit by the state directory alone
+        // (both just delete the marker).
+        CommitOutcome::ClearedStale { marker_version } => Some(format!(
+            "Freenet {current_version}: discarded a post-update probation marker belonging to \
+             {marker_version} (not the running version); auto-rollback protection for \
+             {marker_version} is gone."
+        )),
+        CommitOutcome::Nothing => None,
+    }
+}
+
 /// Clear probation if the running version matches the marker (it has proven
-/// healthy). A marker for a DIFFERENT version is stale (e.g. left over after a
-/// rollback or external install) and is also removed.
+/// healthy). A marker for a DIFFERENT version is stale (e.g. an external install
+/// replaced the binary, or two binaries share a `$HOME`) and is also removed.
 ///
-/// Both outcomes DISARM rollback for the marker's version, so both are reported
-/// on stderr as well as through `tracing` (#5232). The node's `tracing` output
-/// goes to the rolling log files under the log dir, which systemd does NOT
-/// capture — only the process's stdout/stderr reach the journal. An operator
-/// (or an agent) reading `journalctl -u freenet` therefore saw the crash-count
-/// and rollback lines, which the installer prints on stderr, but never the
-/// commit that disarms them. #5232 was filed on exactly that asymmetry: the
-/// commit had been working all along and looked, from the journal, as though it
-/// had never once run. The other two decisions in this module are already on
-/// stderr; this puts the third one where the first two are.
+/// Both outcomes DISARM rollback for the marker's version, so both are announced
+/// on stderr as well as through `tracing` — see [`commit_announcement`] for why
+/// stderr specifically.
 ///
-/// The volume is bounded: at most one line per node start, and only when a
-/// probation marker actually exists.
+/// Reached only from the node's commit timer, so "survived the window ⇒
+/// committed" holds for a process that actually ran the network. A node parked
+/// by `freenet service disable` (`run_disabled_idle`) never reaches it and keeps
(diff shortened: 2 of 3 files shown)


sanity:
## Review: three independent lenses, all findings addressed

Ran three blind reviewers on `ac704ebc0` — skeptical (adversarial bug hunt), testing (coverage and vacuity), and code-first (code before description, to catch claim/implementation drift). **No blocking findings.** Every should-fix is addressed in `9aa196af8`; the dismissals are below with reasons.

### Fixed

**Assertions that could not fail.** All three lenses independently flagged that "no crash was recorded" was unfalsifiable: it is a negative on a subprocess's stderr, and the preceding assertion had already established the marker was gone. It would equally have passed with a misspelled `FREENET_POST_STOP_EXIT_CODE`, a `HOME` that did not relocate the state dir, an unparseable fixture, or a reworded needle. `probation_survives_a_stop_inside_the_commit_window` is now the positive control, and it also covers the untested half of the contract — a stop INSIDE the window keeps rollback armed, which is the half that produced the user-visible rollback. ~15s, no network call.

**`ClearedStale` had no coverage**, despite this PR arguing it is the more dangerous branch. The wording moved into `commit_announcement`, a pure function over `CommitOutcome`, with a unit test over both branches plus the silent `Nothing`. Microseconds instead of a second 60s node boot.

**A race between the commit poll and the announcement.** `commit_probation_at` deletes the marker before `commit_probation` prints, so polling the file could observe "gone" microseconds before the line landed, then read stderr exactly once. It now polls the announcement itself. The skeptical lens added the corollary, which is now a comment in the source: do NOT "fix" this by printing before the removal — `eprintln!` panics on EPIPE, and a panic in front of the state transition would leave a healthy node's marker armed, turning an observability line into a rollback bug.

**Three comments that described things that are not true.** The other stderr prints are in `commands::update` and there are three, not two in this module; the commit task's handle is retained and aborted, not dropped; and `arm_probation` claimed a fixture re-read that does not exist.

**An overstated claim.** "tracing goes to the log files" holds under the shipped systemd deployment but not for the fallback tracer (no log dir resolvable, or `FREENET_LOG_TO_STDERR`), which writes to stdout. Since this PR is meant to be the permanent record of why #5232's evidence was misread, that is now qualified.

**`run_disabled_idle` never commits** — a genuine gap in the original "every healthy boot" claim, found by the skeptical lens. Documented rather than changed, deliberately: a node parked by `freenet service disable` never joined the network, so it has not demonstrated the health probation tests for, and committing there would disarm rollback on evidence the mechanism does not have. The 1h TTL retires the marker instead. Bounded in the meantime — an idle node's `systemctl stop` exits 0, which is `NotCrash`.

**Diagnostics and robustness.** The commit deadline now starts when the WS API answers rather than at spawn, so a slow cold start cannot be reported as "the marker is still armed". Failure messages carry the node's exit status and stderr — a port collision or a rejected flag reports itself there, often before `tracing` is initialised, so the previous diagnostic printed the one file guaranteed to be empty. A stale `target/debug/freenet` predating a version bump now fails a version precondition instead of taking the ClearedStale branch and looking like a rollback bug. `--log-dir` points at the state dir, matching production on Linux where the log pruner and the marker share a directory. The shutdown wait is bounded.

**nextest override**: 300s period, `retries = 0`. Retrying brick-safety equipment would turn "the commit sometimes does not happen" back into a green run.

### Re-mutated after the rework

The suite changed substantially, so the mutation testing was redone. The two tests now fail on **disjoint** mutations, which is the property worth having — neither is redundant, and the positive control genuinely can fail:

| Mutation | committed test | in-window test |
|---|---|---|
| Remove the stderr announcement (pre-fix state) | FAILED | ok |
| `classify_stop`: status 1 no longer a crash | ok | FAILED |

Plus the two from the first round, both still fatal: deleting the `commit_probation` call, and a `v`-prefix mismatch where the marker is removed as stale and never committed.

### Not taken

- **`env!("CARGO_BIN_EXE_freenet")` instead of the `target_dir()` helper** (two lenses, as a nit). Correct in principle, but `persistence_roundtrip.rs` and `fdev_publish_e2e.rs` use the same helper, and changing one of three is worse than changing none. The concrete hazard it guards — a stale binary — is now caught by the version precondition. Worth a follow-up across all three.
- **Parameterising the 60s window so the test runs in ~10s** (testing lens, raised as a fork in the road rather than a finding). Declining: it puts a new env knob on a brick-safety path and stops exercising the constant at its production value. The cheap test now carries the fixture-canary role that motivated most of the speed argument.
- **TCP-reserved port handed to a UDP listener** — pre-existing house pattern; the improved diagnostics mean a collision now reports itself as a collision.

[AI-assisted - Claude]

sanity:
## Rebased onto 0.2.123, re-reviewed, re-verified

Force-pushed `9aa196af8` → `c6dcd3418`. The branch was `CONFLICTING` after `#5230` and the 0.2.123 release landed on main. Because a rebase with a conflict resolution restales both the review and CI, here is what was re-done rather than assumed.

**The conflict** was in `.config/nextest.toml`: #5230 added a `retries = 0` override for `graceful_shutdown_on_sigterm_exits_zero` at the same anchor where this branch adds one for the probation tests. Resolved by keeping both — they are independent, non-overlapping filters. A reviewer confirmed both blocks are byte-identical to their respective sides, no markers remain, the TOML is valid, and the two filter expressions neither overlap nor shadow (and the probation regex does not accidentally match `probation_roundtrip_and_commit` in `rollback.rs`).

**#5230 changes the ground this test stood on**, so that got the real scrutiny. A graceful shutdown now exits **0**, and `classify_stop` short-circuits `"0"` to `NotCrash` *before* `handle_post_stop_at` ever reads the marker. Both tests forward `FREENET_POST_STOP_EXIT_CODE=1` explicitly rather than the observed status, which was written as future-proofing and turns out to be load-bearing: forwarding the real status would now mean neither test touched the probation logic at all, and the in-window test would lose its ability to fail entirely.

The assertions were already right. Three pieces of prose were not, and `c6dcd3418` fixes them — they had been written against the old behaviour and stated its opposite:

- `post_stop_update`'s rationale said 1 is "what a graceful shutdown exits with today" and hedged about #5227's fix landing. It has landed.
- The module doc credited the user-visible rollback to "three fast restarts". Post-#5230 ordinary restarts no longer accumulate strikes, and this PR's own "Not in scope" section already said #5227 was the cause. They now agree.
- "the stop IS counted" merged two separable claims. The test asserts both and now distinguishes them: marker retention is demonstrated by a **real** graceful stop, the `1/3` counting by a forwarded crash status.

Also checked: #5230 does not touch `rollback.rs`, so there was no semantic collision for the rebase to reconcile silently; `classify_stop` is unchanged by both sides; and `MIN_HEALTHY_UPTIME_FOR_UPDATE_EXIT`, which `COMMIT_HEALTHY_UPTIME_SECS` mirrors, was not modified.

**The 0.2.123 bump** lines up: `NODE_VERSION`, `build_info::VERSION`, and the planted marker all resolve to the same `env!("CARGO_PKG_VERSION")` of the same package, so `Committed` stays reachable and `ClearedStale` is not spuriously triggered. `assert_binary_matches_crate_version()` is doing real work on a just-bumped base — a `target/debug/freenet` built before the bump reports 0.2.122, would take the ClearedStale branch, and would otherwise fail as "rollback regression" instead of "rebuild your binary".

**Re-verified locally on the rebased head**: `cargo fmt --check` clean, the 31 `rollback::` unit tests pass, and both integration tests pass against a freshly built 0.2.123 binary (65s).

[AI-assisted - Claude]

sanity:
## Rebased, plus: the new message could lie, and now cannot

Force-pushed `c6dcd3418` → `8cd407b3e`. Rebase onto main was a clean fast-forward (no conflicts, no mid-sequence edits), so the earlier review stands on the replayed commits; the new commit is reviewed below on its own terms.

### The problem with the message this PR adds

`remove_probation_at` discarded its `remove_file` result, so `commit_probation` printed **"committed, auto-rollback disarmed"** whether or not the marker was actually gone. A failed unlink leaves the marker armed and inside its 1h TTL, so an ordinary crash in the next hour can still roll the node back off a version it was just told had passed — and the commit timer fires once per process, so nothing retries.

That would have reintroduced the misleading-journal problem at the exact moment of fixing it. The trigger is a read-only or full state directory, which is the same fleet condition behind #5244.

`remove_probation_at` now returns `io::Result` — "already absent" counts as success, since the post-condition is "no marker", not "I deleted something" — and `commit_probation_at` reports a new `ClearFailed` outcome instead of a false `Committed`. Its announcement says the marker survived, names the cause, and points at the state directory.

The other removal sites still drop the result, but now explicitly and with the reason bound in the name: the stale-version and TTL branches are not announced and re-attempt the removal on the next stop, and the post-rollback one self-heals because the restored binary's next stop sees a version mismatch and takes the stale branch. Surfacing those properly needs an output channel this process does not have — that is #5244.

### A false-green this nearly shipped

The `ClearFailed` line originally began "post-update probation passed, but the marker … could NOT be removed". That **contains** the substring the end-to-end test greps for to conclude the node committed. A failed clear would have sailed through that guard while rollback was still armed — the precise false-green this PR exists to prevent, one level up.

The e2e now keys on `"committed, auto-rollback disarmed"`, which only the success branch can produce, and `a_failed_clear_is_never_announced_as_a_pass` pins the two disjoint in both directions so neither can drift into the other.

### Tests

- `a_failed_clear_is_never_announced_as_a_pass` — the line says STILL ARMED, names the cause, never says "disarmed", and does not contain the e2e's success needle.
- `an_unremovable_marker_reports_clear_failed_not_committed` — drives `commit_probation_at` against a marker in a read-only directory and asserts the outcome is not `Committed`. Skips when running as root, which bypasses the directory permission check; skipping is honest where the test would otherwise be vacuous rather than passing.
- `removing_an_absent_marker_succeeds` — an idempotent re-commit must not raise a spurious `ClearFailed`.

34 `rollback::` unit tests and both integration tests pass locally; `cargo fmt --check` and `clippy` clean.

Related: #5244, filed for the structural cause — the `freenet update` process installs no tracing subscriber at all, so every `tracing::*` call in it is a no-op, and an update can install with rollback protection never armed and no trace anywhere.

[AI-assisted - Claude]

sanity:

Two independent blind lenses (skeptical, code-first), plus my own verification of every cited location. Reviewed at head `8cd407b3e8c7ac466abdbdc6586d60ed64dee07f`.

### Verified

- **This really is announcement-only.** `handle_post_stop_at` is behaviourally identical: the only change through it is `let _rm = remove_probation_at(dir)` becoming `let _retried_next_stop = ...` at `rollback.rs:881, :890, :926, :944, :977`, with the `Result` discarded exactly as before. `should_rollback`, `ROLLBACK_CRASH_THRESHOLD`, `PROBATION_MAX_AGE_SECS` and `COMMIT_HEALTHY_UPTIME_SECS` are untouched, and the set of sites that remove the marker — and their order relative to every other side effect — is unchanged.
- **Nothing parses this stream.** The only consumers of the node's stderr in the tree are `service/wrapper.rs:1098-1102` and the embedded launchd script at `service/macos.rs:390`, both matching solely on `"already in use"`. Neither new line can produce that substring. The `ClearFailed` line does interpolate an arbitrary `io::Error` into stderr, which is the one piece of text this code does not fully control — but `remove_file` cannot plausibly produce an "already in use" message on any platform, so it is a coupling to know about rather than a live bug.
- **The premise holds.** With file logging and a non-terminal stdout no console layer is installed (`tracer.rs`, gated on `stdout().is_terminal()`), so `tracing` genuinely does not reach the journal under a service manager. The `debug!` → `warn!` escalation is justified for the same reason: `release_max_level_info` compiled that branch out of release builds entirely.
- **The tests are not vacuous.** `probation_survives_a_stop_inside_the_commit_window` is a genuine positive control for the negative assertion in the other test — it forces the same fixture, env var and needle to produce a crash line, so a misspelled `FREENET_POST_STOP_EXIT_CODE`, a `HOME` that failed to relocate the state dir, or a reworded needle fails loudly instead of passing silently. The two needles are provably disjoint and pinned as such at `rollback.rs:1318-1325`, so the e2e cannot report a commit while rollback is still armed. `serve_client_api` binds and listens before `node_config.build()` and before the commit task is spawned, so `wait_until_ready` returning genuinely leaves the whole 60s window ahead — the in-window test is not racing the timer.
- **`.config/nextest.toml:83-86` weakens nothing** — it raises the slow timeout and sets `retries = 0`, which is stricter than the ci profile's `retries = 2`.
- **The body is accurate and current**, including the `ClearFailed` half added by the final commit. I checked specifically because the last commit headline differs from the PR title; the body covers it (and the false-green substring collision it introduced and closed) properly. No rewrite needed before squash.

### Non-blocking — filed as #5574 rather than re-spinning a 27-day-green PR

1. **`rollback.rs:583-608` — `eprintln!` runs before the `match` that emits `tracing`.** The comment at `:577-581` reasons carefully about ordering against the *state transition* and concludes "the worst case is a lost announcement", but not about ordering against the tracing emission. `eprintln!` panics on a write error, and the panic unwinds the spawned task before the `match` below it runs — so on a node whose stderr is broken (journald restart → EPIPE, or a Windows child that inherited invalid handles) the operator now loses the `tracing` record they *did* get before this PR. Narrow, and it cannot take the node down (`panic = "abort"` is off and the handle is only ever aborted), but the stated worst case is not the actual worst case. `let _ = writeln!(std::io::stderr(), "{line}")`, or emitting the tracing record first, closes it.
2. **`rollback.rs:547-555` — the `ClearFailed` message is wrong for the stale sub-case.** `commit_probation_at:643-651` routes a marker for a *different* version whose unlink failed into the same variant, so the operator is told "auto-rollback is STILL ARMED and a crash within the next hour can still roll this node back". It cannot: `handle_post_stop_at:872-883` takes the version-mismatch branch and returns `Proceed`. It errs in the safe direction, but a PR whose thesis is "never announce a disarm that did not happen" should equally not announce a danger that does not exist.
3. **`post_update_probation_commit.rs` — the committed test makes a live GitHub call.** With the marker gone, the post-stop `freenet update --check` falls through to `probe_latest_tag`. It cannot install and cannot make the assertion pass wrongly, and the probe timeouts are 10s, but the module doc advertises "makes no network call" about the *other* test only, so this is easy to miss.
4. **`reserve_port()` bind-then-drop race under `retries = 0`.** The override's justification points at `test_ping_blocked_peers`, but that test's own note says `retries = 0` is safe there *because* it has an in-process port-retry loop. This one has none, so a lost race surfaces as "the node exited before its WS API came up" — environmental, but indistinguishable at CI-status level from the regression the test exists to catch.
5. Nits: `env!("CARGO_BIN_EXE_freenet")` would be more robust than the hand-built `target/debug` path (which breaks under `--target <triple>` and can prefer a stale debug binary); the `ClearFailed` unit test silently no-ops as root, which is common in containers; and the `#5244` cross-reference at `rollback.rs:877-880` is loosely worded — that issue is about the missing tracing subscriber, whereas `freenet update` does have stderr.

Two things the reviewers raised that I checked and dropped: the claim that the body is stale (it is not — it covers `ClearFailed`), and the concern that `XDG_STATE_HOME` could redirect the spawned node at the real state dir (`auto_update::state_dir()` is `dirs::home_dir()`-derived on every platform, so `HOME` isolation is sufficient).

Lenses: freenet skeptical-reviewer, freenet code-first-reviewer, plus synthesis. No external (Codex) pass was run; findings were verified against the checked-out code at the head SHA instead.

[AI-assisted - Claude]

sanity:
Sequencing note, so this isn't picked up out of order.

This is `DIRTY` and needs a rebase, but it should wait for **#5247** rather than being rebased now. Both came out of the same #5232/#5244 investigation and both touch the CLI/update output plumbing, so rebasing this first would just guarantee a second conflict resolution against #5247 a few hours later.

#5247 is currently in review with its blocking `rule-review` warning resolved; once it lands, this rebases onto a base that already carries the tracing-subscriber fix it depends on for its stderr output to go anywhere at all.

Not otherwise reviewed yet — flagging the order rather than the content.

[AI-assisted - Claude]

sanity:
Review findings for this PR arrived after it merged, so they are in a follow-up rather than on this branch: **#5649**.

The substantive one is that `ClearFailed` is produced by two branches of `commit_probation_at` and worded for only one. Reached from the version-MISMATCH (stale) arm, both halves of the line are false: the running version was never on probation, so "ran healthily for 60s" describes a test that was not performed, and "auto-rollback is STILL ARMED and a crash within the next hour can still roll this node back" cannot happen, because `handle_post_stop_at` tests `state.new_version != current_version` first, drops the marker and returns `Proceed`. The operator is told brick-safety equipment is live when it is not — the same misleading-journal failure #5232 was filed on, inside the branch this PR added to remove it. Split into `CommitFailed` / `StaleClearFailed` with separate wording, and the new test asserts that premise against `handle_post_stop_at` itself rather than in prose.

Three smaller ones: `eprintln!` on a tokio worker could panic on EPIPE (with `panic = "abort"` one commented-out line away in `Cargo.toml`, that path runs abort → non-zero exit → scored as a probation crash → rollback, so the announcement of a successful commit would have caused the rollback it reports disarming); the `#5104 v-prefix` assertion in the integration test was unreachable, since `commit_probation` emits at most one line per process and it sat behind the generic assertion; and `retries = 0` in `nextest.toml` cited the canary's justification without the in-test port-retry loop that justification depends on.

Note for anyone reading this branch: it was recreated on origin by mistake while applying the fixes, and has been deleted again. The work is on `fix/5240-followup-stale-clear-wording`.

Nothing here changes a rollback decision, threshold or state transition — same as this PR.

[AI-assisted - Claude]

PR #5241, opened 2026-08-08
test(node): pin shutdown_requested to a single signal-handler store

sanity:
## Problem

#5230 (shipped as 0.2.123) makes a requested graceful shutdown exit 0. It does that only when an `AtomicBool` confirms **this** process asked the node to stop:

```rust
Err(e) if shutdown_requested && freenet::listener_exit_is_graceful(&e) => Ok(()),
```

The second conjunct is load-bearing. The same `EventLoopExitReason::GracefulShutdown` sentinel is also raised by genuine faults — a critical `p2p_protoc` channel dying (`ChannelCloseReason::{Bridge, Controller, Notification, OpExecution}`), and `client_events.rs` losing every client transport. Those must keep exiting non-zero, or systemd skips the unit's `ExecStopPost` self-heal and does not count the crash toward probation, and the macOS/Windows wrappers read exit 0 as "normal shutdown" and stand down, leaving the node dead.

So the whole fault-vs-operator-stop discrimination rests on `shutdown_requested` meaning exactly "an operator asked us to stop".

Nothing tested that. `finish_run_maps_only_a_requested_graceful_stop_to_success` takes the flag as a parameter and `tests/graceful_shutdown_exit_code.rs` drives it with a real signal, so neither can see how the flag comes to be set, nor what is fed to `finish_run`. Several one-line edits reopen fault-laundering — the node exits 0, looks healthy, and stays dead — with a completely green test suite. The most tempting site is the auto-update arm of the `select!` in `run_network_node_with_signals`, which also calls `shutdown_handle.shutdown()` but is not an operator stop and must keep exiting 42.

Found by adversarial review of #5230, which rated it the highest-value follow-up.

## Approach

Test-only; no production logic is touched. Two pins.

**`shutdown_requested_is_stored_only_by_the_signal_handler`** — the flag is raised in exactly one place, in the signal-handler task, only after a signal has actually arrived.

The core of it is **closed-world**: an enumeration of every permitted mention of `shutdown_requested` in production, in source order, identified by what follows it. A blocklist of forbidden spellings was the first attempt and it leaked badly — `(*shutdown_requested).store(..)`, `shutdown_requested.as_ref().store(..)`, and handing `&shutdown_requested` to a helper that stores through it all raise the flag while matching no banned string and leaving the single-store count at 1. Enumerating what is forbidden needs foresight about how the next author will spell it; enumerating what is allowed does not. Writing the check immediately found two mentions I had not accounted for (`finish_run`'s own same-named parameter), which is the check working.

On top of that: the one `.store(` must sit inside the brace-matched `let signal_task = { … }` block, and **after** the signal await — a store hoisted to the top of that block stays inside it and keeps the count at 1, but raises the flag unconditionally at startup so every fault exits 0.

**`finish_run_is_fed_the_live_shutdown_flag`** — the write site is only half of it. `finish_run(result, true)` reopens the identical hole without touching the store. So does widening the argument (`shutdown_requested.load(..) || transports_lost`) — not hypothetical, since lost client transports are one of the two faults that must keep exiting non-zero. The whole paren-matched argument list is matched and `||`, `&&`, `true`, `false` rejected, rather than asserting a substring.

This repo has shipped two broken source-scrape pins (#5102), so the mechanics follow `.claude/rules/bug-prevention-patterns.md` deliberately:

- **The scanned region is bounded to the source before `#[cfg(test)] mod`** (`production_region`). A pin scanning the whole file would match its own assertion literals — `include_str!` pulls in the test module. Cutting the region is a structural fix, unlike the `concat!("a", "b")` needle-splitting used by the neighbouring pins in this file, which only holds until someone writes the contiguous string. The anchor is asserted **unique** rather than taking the first match: an earlier `#[cfg(test)] mod` would truncate the region and hide production code, fail-open in the one direction a bounded region exists to prevent.
- **Blocks are brace/paren-matched** (`braced_block`, `delimited_block`), not `split_once`'d. A bare `split_once` does not fail when its anchor moves; it matches a later occurrence and the region balloons silently. A missing anchor or unbalanced delimiters panics.
- **Anti-vacuity guards run both ways.** If brace matching escaped the signal-task block, one of `update_tx`, `startup_update_check`, `run_network_node(`, `UpdateNeededError`, `finish_run(` would be inside it — the first two bound the near side of an over-run, the rest the far side. Conversely the block must contain a signal await; that marker is a *set*, so extracting the cfg-gated wait into a helper (a legitimate refactor) fails informatively rather than panicking that the block is not the signal handler. The marker also had to be one the auto-update arm does not share: the first draft used `shutdown_handle.shutdown().await`, which appears in both, so it could not discriminate.
- **All matching is whitespace-stripped** (`squeeze`) so the pins survive `rustfmt` re-wrapping a call across lines.

## Testing

Eleven mutations. Each is applied to production code and its diff verified non-empty before running, so none is a mutation that failed to mutate; the driver reverts between runs and re-checks the baseline at the end.

| Mutation | Result | Assertion that fired |
|---|---|---|
| second `store(true, …)` in the auto-update arm | FAILED | closed-world mention list |
| **move** the store to the auto-update arm | FAILED | store not in the signal task |
| **hoist** the store above the signal await | FAILED | store-after-signal-await ordering |
| `shutdown_requested.swap(true, …)` in the arm | FAILED | closed-world mention list |
| `let sr = Arc::clone(&…); sr.store(…)` in the arm | FAILED | closed-world mention list |
| `(*shutdown_requested).store(true, …)` in the arm | FAILED | closed-world mention list |
| `shutdown_requested.as_ref().store(true, …)` in the arm | FAILED | closed-world mention list |
| helper `fn mark(flag: &AtomicBool)` + `mark(&shutdown_requested)` | FAILED | closed-world mention list |
| `finish_run(result, true)` | FAILED | both pins |
| `finish_run(result, …load(..) \|\| true)` | FAILED | argument-widening ban |
| `finish_run(result, …load(..) \|\| client_transports_lost)` | FAILED | argument-widening ban |

The last four in that list are the ones the *earlier* blocklist version of this pin let through; they are why it was rewritten closed-world.

Representative output:

```
MUTATION: helper-borrow-store-in-auto-update-arm
+fn mark_stop_requested(flag: &std::sync::atomic::AtomicBool) {
+    flag.store(true, std::sync::atomic::Ordering::SeqCst);
+}
+                    mark_stop_requested(&shutdown_requested);

test tests::shutdown_requested_is_stored_only_by_the_signal_handler ... FAILED
panicked at crates/core/src/bin/freenet.rs:1622:9:
production mentions `shutdown_requested` in a way this pin does not recognise. ...
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 368 filtered out
```

```
MUTATION: finish_run-arg-widened-or-variable
+    let client_transports_lost = false;
-        shutdown_requested.load(std::sync::atomic::Ordering::SeqCst),
+        shutdown_requested.load(std::sync::atomic::Ordering::SeqCst) || client_transports_lost,

test tests::finish_run_is_fed_the_live_shutdown_flag ... FAILED
panicked at crates/core/src/bin/freenet.rs:1673:13:
`finish_run`'s argument must be the flag ALONE: `||` widens the operator-stop test to
cover something else, which is the #5230 fault-laundering hole in a new spelling; got
`result,shutdown_requested.load(std::sync::atomic::Ordering::SeqCst)||client_transports_lost,`
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 368 filtered out
```

Reverted baseline, full `freenet` bin test module:

```
test result: ok. 370 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
git status: (clean)
```

## Known limits

These pins are source scrapes and inherit the class's limits: cutting the whole `#[cfg(test)] mod tests` disarms them, and they reason about text rather than semantics. Specific gaps left open, stated rather than papered over:

- The pins prove the correct `finish_run` call *exists*; they do not prove it *dominates* the exit code. Re-mapping `result` after the call would not be caught.
- **The flag's initial value is not pinned.** The enumeration records only the first 13 characters after each mention, so the declaration's window is `"=Arc::new(std"` and the initialiser sits past the cut. `AtomicBool::new(true)` therefore passes both pins while laundering every fault into a clean exit 0 — the #5230 hole itself. Nothing else in the tree catches it. Tracked in #5575.
- **Only the first `= finish_run(` call site is inspected.** `delimited_block` takes the first match with no uniqueness assertion, unlike `production_region`. A later `return finish_run(result, true);` passes both pins. Distinct from the re-mapping limit above, because it is a `finish_run` call fed a constant. Tracked in #5575.
- Lesser, also in #5575: the widening check is a blocklist (leaks single `|`, `^`, `.max()`); the anti-vacuity escape markers are never asserted to exist, so renaming one silently disarms that guard; and `shutdown_signal(` matches nothing in the file, so the claim that the marker set lets a helper extraction "fail informatively" holds only if the helper is named with that substring. The "all matching is whitespace-stripped" note applies to block *contents*, not to the three anchors, which are matched against unsqueezed text.
- Review raised a stronger alternative: replace the `Arc<AtomicBool>` with a `oneshot::Sender`, which is `!Clone` and consumed by `send`, making "exactly one raiser, exactly once" a compile-time property and deleting the need for the scrape. That is a production-logic change and out of scope for this test-only PR.

Follow-up to #5230.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/freenet.rs
+++ b/crates/core/src/bin/freenet.rs
@@ -1472,9 +1472,22 @@ mod tests {
     /// region would widen to the whole file and the pin would pass vacuously.
     /// Refuse instead. (Same fail-closed reasoning as `commands::auto_update`'s
     /// `fn_body`; see `.claude/rules/bug-prevention-patterns.md`.)
+    ///
+    /// Uniqueness is asserted rather than taking the first match: a second
+    /// `#[cfg(test)] mod` added ABOVE this one would truncate the region early
+    /// and hide any production code below it — fail-OPEN in exactly the
+    /// direction a bounded region exists to protect.
     fn production_region(src: &str) -> &str {
+        let anchor = "\n#[cfg(test)]\nmod ";
+        assert_eq!(
+            src.matches(anchor).count(),
+            1,
+            "expected exactly one `#[cfg(test)] mod` in this file — the region this \
+             pin counts over is bounded by it, and an earlier one would truncate \
+             the region and hide production code"
+        );
         let tests_at = src
-            .find("\n#[cfg(test)]\nmod ")
+            .find(anchor)
             .expect("test module not located — this pin cannot bound anything");
         &src[..tests_at]
     }
@@ -1528,32 +1541,37 @@ mod tests {
     /// `run_network_node_with_signals`, which also calls `shutdown_handle
     /// .shutdown()` but is NOT an operator stop (it must keep exiting 42).
     ///
-    /// So: exactly one store in production code, inside the signal-handler task.
+    /// So: exactly one way to raise the flag, in the signal-handler task, only
+    /// after a signal has actually arrived — and `finish_run` must be fed the
+    /// flag rather than a constant.
     #[test]
     fn shutdown_requested_is_stored_only_by_the_signal_handler() {
         let src = strip_line_comments(include_str!("freenet.rs"));
         let prod = production_region(&src);
-        let signal_task = braced_block(prod, "let signal_task = {");
+        let flat = squeeze(prod);
+        let signal_task = squeeze(braced_block(prod, "let signal_task = {"));
 
         // Anti-vacuity: if brace matching ran past the block, these markers from
         // later in `run_network_node_with_signals` would be inside it, and the
         // containment assertion below would prove nothing.
         for escaped in ["run_network_node(", "UpdateNeededError", "finish_run("] {
             assert!(
-                !signal_task.contains(escaped),
+                !signal_task.contains(&squeeze(escaped)),
                 "the scoped signal-task block escaped its braces (found `{escaped}`) — \
                  this pin would pass vacuously"
             );
         }
-        // Anti-vacuity, other direction: confirm we scoped the right block.
-        assert!(
-            squeeze(signal_task).contains(&squeeze("shutdown_handle.shutdown().await")),
-            "scoped block is not the signal-handler task"
-        );
+        // Anti-vacuity, other direction: confirm we scoped the right block. The
+        // marker must be one the TEMPTING block does not share — awaiting a
+        // signal is unique to this task, whereas `shutdown_handle.shutdown()`
+        // appears in the auto-update arm too and so would not discriminate.
+        let awaits_a_signal = signal_task
+            .rfind(&squeeze("ctrl_c()"))
+            .expect("scoped block is not the signal-handler task: it awaits no signal");
 
         let needle = squeeze("shutdown_requested.store(");
         assert_eq!(
-            squeeze(prod).matches(&needle).count(),
+            flat.matches(&needle).count(),
             1,
             "production code must set `shutdown_requested` in exactly ONE place. A \
              second store — most temptingly in the auto-update arm, which also calls \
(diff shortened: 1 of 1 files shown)


sanity:
## Full-tier review — merging, with two verified gaps recorded in #5575

Two independent blind lenses (skeptical, code-first), both of which reconstructed the pins' matching logic and drove ~20 mutations through them, plus my own verification of every cited location. Reviewed at head `9b991a7ff8534dc8f0c8684efcc1605c59cac86e`.

### The mutation table is accurate, and the pins are not vacuous

All eleven claimed mutations reproduce, and each fires on the assertion the table names — including the two subtlest: moving the store into the auto-update arm (caught by containment, not by the enumeration, which is position-blind between the clone and the load) and hoisting the store above the signal await (caught by the ordering check, which a location-only pin would miss).

Non-vacuity confirmed directly: every literal the pins search for exists in the real source, the seven-element `follows` vector computed from the file matches `expected` byte-for-byte, and the production region is ~1400 lines rather than an empty or truncated slice.

**The #5450 self-satisfaction trap is genuinely avoided.** The region anchor is written `"\n#[cfg(test)]\nmod "`, which in the file's own bytes is the two-character escape `\n` and not a newline, so the assertion literal cannot match itself. The anchor count is asserted to be exactly 1, so a missing anchor fails with `left: 0, right: 1` and a duplicated one fails too — fail-closed in both directions, which is the property a bounded region exists to provide. Even if the region cut vanished entirely, the enumeration is length- and order-sensitive, so the test module's own mentions would bloat the vector and fail rather than pass.

### Two gaps that survive, both verified

Neither blocks — the pin is strictly better than the nothing that is on `main` today — but both are places where the property can be broken with the suite green, so they are recorded in #5575 rather than left implicit.

1. **`AtomicBool::new(false)` → `new(true)` passes both pins.** The enumeration records only the first 13 characters after each mention, so the declaration's window is `"=Arc::new(std"` and the initialiser sits about thirty characters past the cut. Flipping it leaves all seven windows byte-identical and both pins green, while laundering every fault into a clean exit 0 — the exact #5230 hole. Nothing else in the tree catches it: `finish_run_maps_only_a_requested_graceful_stop_to_success` takes the bool as a parameter, and `tests/graceful_shutdown_exit_code.rs` asserts SIGTERM → exit 0, which a stuck-true flag still satisfies.
2. **`delimited_block` does not assert its anchor is unique.** It takes the first `= finish_run(` with no count check, unlike `production_region`, which asserts uniqueness and documents exactly this reasoning. A later `return finish_run(result, true);` therefore passes both pins. This is adjacent to but distinct from the acknowledged "re-mapping `result` after the call" limit, because it is a `finish_run` call fed a constant.

Both are a couple of lines each. #5575 also records three lesser items: the widening check is a blocklist (leaks single `|`, `^`, `.max()`) in a PR that argues at length against blocklists; the anti-vacuity escape markers are never asserted to exist, so renaming `update_tx` silently disarms that guard; and `shutdown_signal(` matches nothing in the file, so the "a set, so extracting the wait into a helper fails informatively" rationale is a name gamble rather than a design property — any other helper name gets the panic the body says is avoided.

### Maintainability

Several of the seven pinned windows encode incidental source text rather than the invariant. Importing `Ordering`, inserting a statement between the `Arc::clone` and the `GlobalExecutor::spawn`, adding a parameter to `finish_run`, or renaming `finish_run`'s parameter all fail the pin — each with a message that talks exclusively about fault-laundering and a node that stays dead. Someone who only reordered an import is told they may have opened a security hole. The message does close with "if the change is legitimate, extend the list deliberately", and `assert_eq!` dumps both vectors, so it is recoverable rather than confusing — but leading with the benign interpretation would reduce the chance a future author deletes the pin instead of updating it. Two of the seven entries are `finish_run`'s own `bool` parameter, in the list only through a name collision; scoping the scan to `run_network_node_with_signals` would drop both without weakening anything.

Rebase risk is low: current `origin/main` has the same production shapes and a single `#[cfg(test)] mod tests`, so `expected` still holds.

Lenses: freenet skeptical-reviewer, freenet code-first-reviewer, plus synthesis. No external (Codex) pass was run; both lenses independently re-implemented the pin logic and executed mutations against the real file, and I verified the two gaps above directly against the source.

[AI-assisted - Claude]

PR #5242, opened 2026-08-08
fix(service): don't count a deliberate systemd stop as a crash

sanity:
## Problem

#5230 (0.2.123) closed the main path: a *requested* graceful shutdown now exits 0, so an ordinary `systemctl stop` of a **running** node is no longer scored as a post-update-probation crash.

A second path to the same symptom remained, and it is the one that overlaps the rollout scenario.

The node installs its SIGTERM handler in `run_network_node_with_signals` (`crates/core/src/bin/freenet.rs:367`), which `run_network` only reaches after `serve_client_api` (`:227`), `NodeConfig::new` (`:236`) and `node_config.build()` — the redb open plus contract-store load. A SIGTERM arriving before that point kills the process under the **default disposition**. systemd then reports `$EXIT_STATUS` as the signal *name* `TERM`; the unit's `ExecStopPost` `case` has no arm for names, so it fell through to `*)` and ran `freenet update --quiet`; and `rollback.rs::classify_stop` likewise has no arm for names, so it fell through to `_ => StopClass::Crash`.

The window coincides with probation by construction: a node inside its post-update probation window has *just* restarted, so it is also near the front of its boot. On a gateway with a large contract store, `build()` is not brief. That is exactly when an operator or a deploy script is most likely to stop it.

## Solution

Gate the hook on systemd's `$SERVICE_RESULT` in addition to `$EXIT_STATUS`. The hook is skipped only when the result is `success` **and** `$EXIT_CODE` is `killed` — which `man systemd.exec` documents as precisely "killed by HUP/INT/TERM/PIPE under the default disposition".

```diff
-ExecStopPost=-/bin/sh -c 'case "$$EXIT_STATUS" in 0|43) ;; *) …update --quiet ;; esac'
+ExecStopPost=-/bin/sh -c 'case "$$SERVICE_RESULT $$EXIT_CODE" in "success killed") exit 0 ;; esac; case "$$EXIT_STATUS" in 0|43) ;; *) …update --quiet ;; esac'
```

**The guard is narrow on purpose.** This hook does two jobs: it is the #4549 self-heal *and* it is what **counts crashes** for the #4073 crash-loop rollback. An over-broad skip would silently disarm rollback — a node crash-looping on a bad release would have no way back. That failure is silent and strictly worse than the bug being fixed.

The brief for this change proposed that `$SERVICE_RESULT` is `success` for a stop job *however* the process died. **That is not what systemd does**, and gating on `$SERVICE_RESULT` alone would have been a serious regression. Two concrete traps the narrow form avoids:

- `SuccessExitStatus=42 43` makes the **voluntary update exit 42** report `$SERVICE_RESULT=success`. Skipping on the result alone would have disabled auto-update entirely. (Confirmed against real systemd, not just the man page.)
- `oom-kill` and `timeout` also report `$EXIT_CODE=killed`. Skipping on the disposition alone would have stopped counting OOM kills and hung shutdowns.

## systemd `$SERVICE_RESULT` semantics

Source: `man systemd.exec`, "$SERVICE_RESULT" (Table 5 and the `$SERVICE_RESULT`/`$EXIT_CODE`/`$EXIT_STATUS` combination table). Rows marked **probed** were additionally reproduced against real systemd 255 (255.4-1ubuntu8.16) using transient `systemd-run --user` units.

| Scenario | `$SERVICE_RESULT` | `$EXIT_CODE` | `$EXIT_STATUS` | Hook | Verified |
|---|---|---|---|---|---|
| `systemctl stop`, node handled SIGTERM, exit 0 | `success` | `exited` | `0` | **skip** | probed |
| `systemctl stop` during startup, killed by SIGTERM (**the #5227 window**) | `success` | `killed` | `TERM` | **skip** | probed |
| Ctrl+C / SIGINT before the handler is installed | `success` | `killed` | `INT` | **skip** | documented |
| Killed by SIGHUP under the default disposition | `success` | `killed` | `HUP` | **skip** | documented |
| Killed by SIGPIPE under the default disposition | `success` | `killed` | `PIPE` | **skip** | documented |
| Another instance holds the port (exit 43) | `success` | `exited` | `43` | **skip** | probed |
| Voluntary update-needed exit 42 (`SuccessExitStatus=42`) | `success` | `exited` | `42` | **run** | probed |
| Panic (exit 101) | `exit-code` | `exited` | `101` | **run** | probed |
| Fast crash / boot wedge (exit 45, #4551) | `exit-code` | `exited` | `45` | **run** | probed |
| Early startup error (exit 1) | `exit-code` | `exited` | `1` | **run** | probed |
| SIGABRT | `signal` | `killed` | `ABRT` | **run** | probed |
| SIGSEGV with core dump | `core-dump` | `dumped` | `SEGV` | **run** | documented |
| SIGKILL from outside (OOM-shaped) | `signal` | `killed` | `KILL` | **run** | probed |
| OOM killer | `oom-kill` | `killed` | `TERM`/`KILL` | **run** | documented only |
| Stop hung past `TimeoutStopSec`, escalated | `timeout` | `killed` | `KILL` | **run** | probed |
| `ExecCondition=` failed | `exec-condition` | `exited` | `1`…`254` | **run** | documented |
| Protocol violation / start limit hit | `protocol`, `start-limit-hit` | *unset* | *unset* | **run** (forwards empty) | documented |
| Watchdog deadline missed | `watchdog` | `killed`/`dumped`/`exited` | various | **run** | **unreachable** — neither unit sets `WatchdogSec=` |
| systemd < 232 (`$SERVICE_RESULT` unset) | *unset* | *unset* | as before | unchanged | documented (added in v232) |

Notes on cases the brief asked about specifically:

- **A stop job while the process is mid-startup** — this is the bug. `success`/`killed`/`TERM`; now skipped. Probed.
- **SIGKILL during a stop job** — probed via a unit that traps and ignores SIGTERM with `TimeoutStopSec=2`. systemd escalates and reports `timeout`, not `success`, so the hook still runs and the crash is still counted. Behaviour unchanged from today; a shutdown that hangs past its timeout remains classified as a fault.
- **A crash with `Restart=always` in effect** — `Restart=` does not enter into `$SERVICE_RESULT`; it selects what happens *after* `ExecStopPost`. Probed exit 101 / 45 / 1 with the units' real `Restart=always` semantics unchanged; all still run the hook.
- **A failed start** — an `ExecStart` that exits non-zero immediately is `exit-code`/`exited`/`N` and still runs the hook, preserving the #4549 self-heal. Probed via exit 1.
- **`systemctl restart`** — the stop half behaves exactly like `systemctl stop` (both rows above). The start half is a normal start.

**What I could not verify by probe:** a genuine kernel OOM kill (`oom-kill`), a real `core-dump`, `exec-condition`, and the `protocol`/`start-limit-hit` family. These rest on the documented table only. All are `$SERVICE_RESULT` values other than `success`, so they take the unchanged `$EXIT_STATUS` path regardless. The `oom-kill` row is the one worth calling out, because it is the case where `$EXIT_CODE=killed` and the guard must *not* fire; the externally-SIGKILLed probe (`signal`/`killed`/`KILL`) exercises the same shape.

## Two consequences, stated explicitly

Both are recorded in a comment at the hook itself, not just here.

**1. The node must never die by TERM/INT/HUP/PIPE to report a *fault*.** systemd cannot tell who sent a signal, so all four members of the `success`/`killed` cell are now indistinguishable from an operator's stop and are uncounted. This is a genuine widening beyond the TERM case that motivated the change. It is safe **today**: I grepped the crate and there is no self-signal path — the fatal-listener and redb-poison aborts use `process::exit`, and Rust's runtime sets SIGPIPE to `SIG_IGN`. If anyone later adds a `libc::raise`, restores the default SIGPIPE disposition, or routes an abort through TERM, that fault becomes permanently invisible to rollback. The guard comment says so.

**2. A release that *hangs* in the pre-handler window no longer produces any rollback signal.** Previously, an operator's `systemctl stop` of a wedged node was counted as a crash, so three of them triggered a rollback. That is gone. Being honest about it: **that rescue path was an accident of the exact ambiguity that caused #5227** — it is the same event, and it cannot be kept without reinstating the bug it caused. A release that *crashes* in that window is unaffected and still counts (probed). Detecting a startup **hang** needs a real mechanism — `WatchdogSec=`, or installing the signal handler before the heavy startup work — which is follow-up work, not something this hook can do. Raised by adversarial review; I judged it non-blocking for this PR but it should not be lost.

## Testing

Two layers.

**1. In CI — `exec_stop_post_counts_crashes_and_skips_deliberate_stops`** (`service/linux.rs`). Rather than string-matching the template, it extracts the script systemd would hand to `/bin/sh`, undoes the `$$` escaping, and **runs it under a real shell** with the environment systemd documents for each scenario, for both the user and the system unit. 19 rows; every row asserts the decision in **both** directions, and each run that fires also asserts the status was forwarded through `POST_STOP_EXIT_CODE_ENV_VAR` (referenced via the constant, so a rename cannot leave the test checking a variable nothing sets).

A test that pinned only the skip would let someone disable crash counting entirely and stay green, so the crash rows are the load-bearing half. Mutation-tested — each mutation applied to the committed tree, the pin re-run, then reverted:

| Mutation | Result |
|---|---|
| **M1** revert the guard (restore the old `ExecStopPost`) | **FAILS** on `"systemctl stop during startup, killed by SIGTERM"` |
| **M2** over-broad: skip on `$SERVICE_RESULT=success` alone | **FAILS** on `"voluntary update-needed exit 42"` |
| **M3** over-broad: skip on `$EXIT_CODE=killed` alone | **FAILS** on `"SIGABRT"` (and 4 more rows) |
| **M4** single-`$` revert of the two new variables | **FAILS** in `systemd_units_pass_node_exit_code_to_post_stop_update` |

M4 exists because the decision-matrix test un-doubles `$$` before running the script, so `$SERVICE_RESULT` and `$$SERVICE_RESULT` are byte-identical to it. A single-`$` revert would make the guard compare empty strings and silently restore the bug, so the doubling is pinned on the raw template alongside the pre-existing `$$EXIT_STATUS` pin.

**2. Manually, against real systemd 255** — not checked in, because GitHub runners have no systemd user session. I generated the unit with a stand-in binary, took the `ExecStopPost` line **the template actually emits**, passed it to `systemd-run --user` as a transient unit property, and checked whether the stand-in `freenet update` really ran for each of 10 scenarios. All 10 matched the table above. Re-running that harness under mutation **M2** correctly failed the exit-42 row, so it is not vacuous either. This is what closes the seam between "the shell script decides correctly" and "systemd sets the variables I think it does".

Also corrected in this PR: three pre-existing comments and assert messages claimed `ExecStopPost` "skips only 0 and 43". There is now a third exemption, so they had become false.

## Scope limits — please read, these are real

- **Linux/systemd only.** The macOS (`launchd` + wrapper script) and Windows wrappers do not use these units. Their equivalent of this startup window **stays open**; this PR does not touch them.
- **The deployed fleet is not fixed the moment this merges.** An existing install keeps its current unit until the unit is re-templated. Three routes, in order of speed:
  - `freenet service doctor` re-templates immediately (`service/purge.rs::service_doctor` → `install_service`), as does a reinstall.
  - **Automatically, with roughly a one-release lag**: `freenet update` calls `ensure_service_file_updated` (`update.rs:789`), which re-renders the unit from the template compiled into the **running** binary. So a node updating *away from* the first release containing this change writes the new unit. Correcting an earlier version of this section, which implied only the manual routes existed.
  - A plain `systemctl restart` does **not** re-template.
  - Caveat on the automatic route: `#4287`'s sidecar-hash check deliberately leaves a **hand-edited** unit alone, so an operator who customised their unit keeps the old hook until they reconcile it.
- **The node-side window itself is untouched.** Installing the signal handler before the heavy startup work (candidate 1 in the issue) remains unbuilt; this is candidate 2. They are complementary. This one fixes the class rather than one instance and has a much smaller blast radius, but a node that dies to an *unhandled* signal in that window still exits under the default disposition — the change here is only that systemd and the updater no longer misread that as a crash.
- **Deployed nodes running the old unit still mis-classify a signal-name stop**, because `rollback.rs::classify_stop` has no arm for signal names either. Adding one would cover the fleet without a re-template, but it also carries a real under-counting risk (a start/stop-timeout death would stop counting), so I did not smuggle it into this PR. Worth a separate decision.
- **A stop whose graceful shutdown itself errors** still exits non-zero, so it is still counted as a crash. Unchanged, and out of scope here.

## Required pre-release smoke test (framework machine)

This change alters the **unit template**, so `cargo test` and CI cannot see the thing that actually matters: whether a real `systemctl stop` of a real node still gets scored as a crash. The general release smoke test does not cover it. Run this specifically:

1. Let the framework node auto-update onto a build containing this change, **and confirm its unit was re-templated** — the fix does nothing until it is:
   ```
   grep -c 'success killed' ~/.config/systemd/user/freenet.service   # expect 1
   ```
   If it is 0, the node updated but is still on the old unit; force it with `freenet service doctor` before continuing, otherwise the test is vacuous.
2. Note the probation state before the stop:
   ```
   cat ~/.local/state/freenet/probation.json 2>/dev/null
   ```
3. `systemctl --user stop freenet`, then start it again.
4. **Assert no probation crash was recorded** — the crash count must be unchanged from step 2, and the journal must NOT contain the `crash N/3 during post-update probation` line:
   ```
   journalctl --user -u freenet --since '5 min ago' | grep -i 'probation'
   ```
5. Repeat once with the stop issued **during startup** (stop within a second or two of starting, before the node is ready). This is the actual #5227 window — step 3 alone exercises only the path #5230 already fixed, so **skipping step 5 makes the whole test vacuous for this change.**

Negative control, so the test can fail: `freenet update` must still be invoked for a genuine crash. `kill -ABRT` the node and confirm the probation counter *does* increment.

## Suggested release-note wording

> A deliberate `systemctl stop` is no longer counted as a crash by the post-update rollback, even when the stop arrives while the node is still starting up. New and repaired systemd installs get this immediately; existing installs pick it up automatically about one release later (when the node next updates), or right away via `freenet service doctor`. A hand-customised unit is left alone and keeps the old behaviour until reconciled. macOS and Windows are unaffected.

Refs #5227

[AI-assisted - Claude]

---

## Addendum: ejected by the merge queue, and why the branch's green could not see it

This PR sat green and `mergeable=CLEAN` for 27 days, was enqueued, and was **ejected** by the merge queue on a clippy error that its own branch CI never reported:

```
error: non-binding `let` on an expression with `#[must_use]` type
  --> crates/core/src/bin/commands/service/linux.rs:1238:9
1238 |         let _ = std::fs::remove_file(&marker);
   = note: requested on the command line with `-D clippy::let-underscore-must-use`
```

Three explanations ruled out before landing on the real one: not a conflict (`git merge-tree --write-tree` against current main is clean, so no rebase was needed and none was done), not a toolchain bump (`rust-toolchain.toml` pins 1.94.0 at the merge-base and on main), and not a newly-added lint (already `deny` at the merge-base). What changed on main is the lint's **reach** — #5292 extended lint coverage to test code after this branch last ran CI, and the offending line is in test code this PR adds.

**The diff was never wrong; only the merged combination fails.** No amount of re-reviewing this PR would have found it, and the branch's own checks are structurally incapable of reporting it.

The fix handles the value rather than using the `let _rm = ...` named-binding idiom found elsewhere in this tree. That idiom silences the lint and would have left a real defect: `run_hook` decides whether the hook ran by testing for this marker *after* running the script, so a marker left behind by the previous scenario makes the next one report a **skipped** hook as having run — every "must skip" row in the decision matrix would pass while measuring the wrong thing. Only `NotFound` is tolerated; anything else panics, because at that point the matrix can no longer measure what it claims to.

Verified with clippy against the **merged** tree (this branch merged onto current main), not the branch alone — a stale branch still carries old copies of files main has since rewritten, so a branch-only run reports unrelated errors and tests a tree that will never exist after merge.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -1446,8 +1446,10 @@ mod tests {
         );
         // #4073 / #4551: ExecStopPost must fire `freenet update` for ANY
         // non-graceful exit — every status except 0 (graceful) and 43 (another
-        // instance) — so it covers the voluntary update-needed 42, the fast-crash
-        // 45, AND the panic/signal/early-error codes the watchdog never produces.
+        // instance), and, since #5227, except a deliberate stop identified by
+        // $SERVICE_RESULT ahead of this case — so it covers the voluntary
+        // update-needed 42, the fast-crash 45, AND the panic/signal/early-error
+        // codes the watchdog never produces.
         // Firing broadly preserves the #4549 self-heal AND enables crash-loop
         // rollback (#4073); 45 and the crash codes staying out of
         // SuccessExitStatus still count toward StartLimitBurst. A `case` (not
@@ -1456,7 +1458,10 @@ mod tests {
             service_content.contains("case \"$$EXIT_STATUS\" in 0|43) ;; *)")
                 && service_content.contains("update --quiet"),
             "ExecStopPost must run `freenet update` for any non-graceful exit via a \
-             case that skips only 0 and 43 (#4073 broadened crash coverage; #4551)"
+             case that skips 0 and 43 (#4073 broadened crash coverage; #4551). Since \
+             #5227 a deliberate stop is skipped as well, by a separate \
+             $SERVICE_RESULT guard ahead of this case — see \
+             service::linux::tests::exec_stop_post_counts_crashes_and_skips_deliberate_stops"
         );
         assert!(
             !service_content.contains("42|45)"),
@@ -1555,12 +1560,16 @@ mod tests {
             "ExecStopPost must use the doubled $$EXIT_STATUS (single-$ revert breaks auto-update)"
         );
         // #4073 / #4551: ExecStopPost fires `freenet update` for ANY non-graceful
-        // exit (every status except 0 and 43) — system unit too.
+        // exit (every status except 0 and 43, and since #5227 except a deliberate
+        // stop identified by $SERVICE_RESULT) — system unit too.
         assert!(
             service_content.contains("case \"$$EXIT_STATUS\" in 0|43) ;; *)")
                 && service_content.contains("update --quiet"),
             "ExecStopPost must run `freenet update` for any non-graceful exit via a \
-             case that skips only 0 and 43 (#4073 broadened crash coverage; #4551)"
+             case that skips 0 and 43 (#4073 broadened crash coverage; #4551). Since \
+             #5227 a deliberate stop is skipped as well, by a separate \
+             $SERVICE_RESULT guard ahead of this case — see \
+             service::linux::tests::exec_stop_post_counts_crashes_and_skips_deliberate_stops"
         );
         assert!(
             !service_content.contains("42|45)"),
--- a/crates/core/src/bin/commands/service/linux.rs
+++ b/crates/core/src/bin/commands/service/linux.rs
@@ -536,7 +536,32 @@ TimeoutStopSec=45
 # `freenet update` so #4073 can tell a post-stop restart from a manual update
 # and classify the crash; an OLD binary (e.g. one we rolled back TO) ignores the
 # unknown env var.
-ExecStopPost=-/bin/sh -c 'case "$$EXIT_STATUS" in 0|43) ;; *) {post_stop_env}="$$EXIT_STATUS" {binary} update --quiet ;; esac'
+#
+# A DELIBERATE STOP IS NOT A CRASH (#5227). The leading guard skips the hook when
+# systemd reports the unit's $SERVICE_RESULT as "success" AND $EXIT_CODE as
+# "killed" — i.e. systemd itself considers this a clean stop and the process died
+# from a signal rather than by exiting. Per `man systemd.exec` ($SERVICE_RESULT
+# table) that combination is precisely "killed by HUP/INT/TERM/PIPE under the
+# default disposition", which is what happens when a `systemctl stop` lands
+# BEFORE the node has installed its own SIGTERM handler (it is installed after
+# the redb open and contract-store load, which on a large gateway is not brief).
+# systemd then passes $EXIT_STATUS as the signal NAME "TERM"; the $EXIT_STATUS
+# case below has no arm for names, so it fell through to `*)` and rollback.rs
+# scored a clean stop as a post-update probation crash.
+#
+# The guard is narrow ON PURPOSE. This hook is not only the self-heal — it is
+# also what COUNTS crashes for the #4073 rollback, so an over-broad skip would
+# silently disarm rollback, which is worse than the bug it fixes. Still counted:
+#   "exit-code"  panic 101, fast-crash 45, early-startup error 1 (EXIT_CODE=exited)
+#   "signal" / "core-dump"   SEGV, ABRT
+#   "oom-kill"   EXIT_CODE IS "killed", but the result is not "success"
+#   "timeout"    a stop that hung past TimeoutStopSec and was escalated
+#   "watchdog"   keep-alive deadline missed
+# The voluntary update exit 42 also still runs the updater: SuccessExitStatus=42
+# does make its $SERVICE_RESULT "success", but its $EXIT_CODE is "exited", not
+# "killed", so the guard does not match. On systemd < 232 $SERVICE_RESULT is unset,
+# the guard can never match, and the unit degrades to exactly the old behaviour.
+ExecStopPost=-/bin/sh -c 'case "$$SERVICE_RESULT $$EXIT_CODE" in "success killed") exit 0 ;; esac; case "$$EXIT_STATUS" in 0|43) ;; *) {post_stop_env}="$$EXIT_STATUS" {binary} update --quiet ;; esac'
 # Exit 42 (auto-update) and 43 (another instance) are clean exits, so they are
 # not counted as failures — without this, rapid update cycles (exit 42 →
 # ExecStopPost → restart) could exhaust the burst limit and kill the service.
@@ -694,7 +719,32 @@ TimeoutStopSec=45
 # `freenet update` so #4073 can tell a post-stop restart from a manual update
 # and classify the crash; an OLD binary (e.g. one we rolled back TO) ignores the
 # unknown env var.
-ExecStopPost=-/bin/sh -c 'case "$$EXIT_STATUS" in 0|43) ;; *) {post_stop_env}="$$EXIT_STATUS" {binary} update --quiet ;; esac'
+#
+# A DELIBERATE STOP IS NOT A CRASH (#5227). The leading guard skips the hook when
+# systemd reports the unit's $SERVICE_RESULT as "success" AND $EXIT_CODE as
+# "killed" — i.e. systemd itself considers this a clean stop and the process died
+# from a signal rather than by exiting. Per `man systemd.exec` ($SERVICE_RESULT
+# table) that combination is precisely "killed by HUP/INT/TERM/PIPE under the
+# default disposition", which is what happens when a `systemctl stop` lands
+# BEFORE the node has installed its own SIGTERM handler (it is installed after
+# the redb open and contract-store load, which on a large gateway is not brief).
+# systemd then passes $EXIT_STATUS as the signal NAME "TERM"; the $EXIT_STATUS
+# case below has no arm for names, so it fell through to `*)` and rollback.rs
+# scored a clean stop as a post-update probation crash.
+#
+# The guard is narrow ON PURPOSE. This hook is not only the self-heal — it is
+# also what COUNTS crashes for the #4073 rollback, so an over-broad skip would
+# silently disarm rollback, which is worse than the bug it fixes. Still counted:
+#   "exit-code"  panic 101, fast-crash 45, early-startup error 1 (EXIT_CODE=exited)
+#   "signal" / "core-dump"   SEGV, ABRT
+#   "oom-kill"   EXIT_CODE IS "killed", but the result is not "success"
+#   "timeout"    a stop that hung past TimeoutStopSec and was escalated
+#   "watchdog"   keep-alive deadline missed
+# The voluntary update exit 42 also still runs the updater: SuccessExitStatus=42
+# does make its $SERVICE_RESULT "success", but its $EXIT_CODE is "exited", not
+# "killed", so the guard does not match. On systemd < 232 $SERVICE_RESULT is unset,
+# the guard can never match, and the unit degrades to exactly the old behaviour.
+ExecStopPost=-/bin/sh -c 'case "$$SERVICE_RESULT $$EXIT_CODE" in "success killed") exit 0 ;; esac; case "$$EXIT_STATUS" in 0|43) ;; *) {post_stop_env}="$$EXIT_STATUS" {binary} update --quiet ;; esac'
 # Exit 42 (auto-update) and 43 (another instance) are clean exits, so they are
 # not counted as failures — without this, rapid update cycles (exit 42 →
 # ExecStopPost → restart) could exhaust the burst limit and kill the service.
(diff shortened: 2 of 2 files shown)


sanity:
## Review: 2 independent blind lenses + a real-systemd probe

Both reviewers read the checked-out code without seeing each other's findings. One was pointed specifically at the failure mode that matters here — *does genuine crash counting still work* — because silently disarming the #4073 rollback would be worse than the bug being fixed. Both independently reproduced the mutation battery and both probed real systemd 255 rather than trusting the man page alone.

**Nothing blocking on the implementation.** Both lenses tried to break the shell logic, the systemd semantics and the test, and neither could. Verified between them: systemd's parsing of the doubled `$$` and the nested quotes inside the single-quoted argument; `exit 0` mid-`sh -c`; the full decision matrix under `dash`; no interaction with `SuccessExitStatus`, `RestartPreventExitStatus`, or the #4551 `StartLimitBurst` limiter; `classify_stop` still correct for everything that now reaches it; both templates changed identically; `systemd-analyze verify` accepts the emitted line.

### Findings addressed in `e99d0ccbf`

| # | Finding | Fix |
|---|---|---|
| 1 | `success/killed/HUP` and `success/killed/PIPE` are also skipped — an unstated widening beyond TERM/INT | Both now have test rows and a table row, with the reason the widening is currently safe |
| 2 | The comment cited `"watchdog"` as evidence the guard is narrow, but neither unit sets `WatchdogSec=`, so that result is unreachable | Labelled unreachable; still-counted list completed with `exec-condition` / `protocol` / `start-limit-hit` / `resources` |
| 3 | The harness could not express "`$EXIT_STATUS` unset", so the whole no-main-process family was untestable | `exit_status` is now `Option`; `protocol` and `start-limit-hit` rows added (hook fires, forwards empty, `post_stop_status_from_env` maps to `None` — pre-existing, now pinned) |
| 4 | Nothing pinned the `$$` doubling of the two *new* variables; the decision-matrix test un-doubles `$$`, so a single-`$` revert is invisible to it | Pinned on the raw template next to the existing `$$EXIT_STATUS` pin, and added as mutation **M4** |
| 5 | A latent self-signal path (`libc::raise`, restored SIGPIPE disposition, abort-via-TERM) would become permanently uncounted | Guard comment at the hook. Grepped: no such path exists today |
| 6 | Three pre-existing comments/assert messages claimed `ExecStopPost` "skips only 0 and 43" — now false | Corrected, pointing at the decision-matrix test |
| 7 | The exec assertion blamed "syntax error" for a status-126 `noexec $TMPDIR` failure | Message now names that case |
| 8 | PR body understated fleet coverage: `ensure_service_file_updated` (`update.rs:789`) re-templates on the next update | Verified in source and corrected — automatic with a ~one-release lag, with the #4287 hand-edited-unit caveat |

### Findings recorded, deliberately not fixed here

**A release that HANGS in the pre-handler window loses its only rollback signal.** It never exits, so `StartLimitBurst` never trips; previously an operator's `systemctl stop` counted as a crash and three of them rolled back. That is gone.

I judged this non-blocking, and want the reasoning on the record rather than buried: **that rescue path was an accident of the exact ambiguity that caused #5227.** A wedged node's stop and a healthy node's stop are the same event to systemd — indistinguishable by construction — so it cannot be kept without reinstating the bug. A release that *crashes* in that window is unaffected and still counts (probed). Detecting a startup **hang** needs `WatchdogSec=` or an earlier signal-handler install; both are follow-up, and both are noted in the guard comment so the next person does not rediscover this the hard way.

The reviewer's preferred fix — install the handler before the heavy startup work — is candidate 1 from #5227 and remains the better root fix. This PR is candidate 2; they are complementary, and #5227 stays open for candidate 1.

Also recorded, unchanged and out of scope: a `systemctl stop` whose graceful shutdown itself errors still exits non-zero and is still counted; and deployed nodes on the old unit still mis-classify a signal-name stop, because `classify_stop` has no arm for signal names (fixing that would cover the fleet without a re-template but carries its own under-counting risk — worth a separate decision, not a smuggled one).

### Unrelated, found in passing

`cargo clippy --all-features` fails on `crates/core/src/transport/connection_handler.rs:384` (`new_test_with_bandwidth` is dead — only the definition and a doc-comment reference exist). Pre-existing on `main`, not touched by this PR, and CI's clippy invocation does not use `--all-features` so it is not currently red. Flagging rather than expanding this PR's scope.

[AI-assisted - Claude]

sanity:

Two independent blind lenses (skeptical, code-first), plus my own verification of every cited location and a row-by-row check of the body's systemd table. Reviewed at head `e99d0ccbf048e29ea795e46c4987404f9a49ded4`.

### The dangerous direction is closed

The question that matters for this change is not "is a deliberate stop skipped" but "can a real crash be misclassified as deliberate", because this hook is what counts crashes for the #4073 post-update rollback — an over-broad skip silently disarms rollback on a node crash-looping off a bad release.

`$SERVICE_RESULT=success` together with `$EXIT_CODE=killed` is reachable only when the main process died by a signal that systemd's `is_clean_exit` treats as clean: HUP/INT/TERM/PIPE under the default disposition, or a signal named in the unit's own `SuccessExitStatus=`. Neither unit lists a signal there (`linux.rs:599`, `:806`) and both are `Type=simple`, so the second door is shut today. Everything genuinely faulty was traced to a result the guard does not match: kernel/cgroup OOM and `systemd-oomd` (SIGKILL → `oom-kill`/`signal`), stop-timeout escalation (`timeout`), SEGV/ABRT (`core-dump`/`signal`), and panic 101 / fast-crash 45 / early-error 1 (`exit-code`). Exit 42 is the interesting near-miss — `SuccessExitStatus=42 43` really does make it report `success` — and the guard cannot match it because `$EXIT_CODE` is `exited`, which is exactly why keying on `$SERVICE_RESULT` alone would have disabled auto-update entirely.

### Verified

- **Both units changed identically** — `linux.rs:588` (user) and `linux.rs:795` (system) are byte-identical `ExecStopPost=` lines, with the rationale block duplicated verbatim at `:539-587` and `:746-794`.
- **Shell semantics.** The two `case`s compose as independent statements. `"success killed"` is a quoted literal with globbing disabled. Unset or empty variables yield `" "`, `"success "` or `" killed"` — none match — so on systemd < 232 the unit degrades to exactly the previous behaviour, as claimed.
- **The tests execute the real script**, not a template string: `exec_stop_post_script` (`linux.rs:1203`) extracts from the *generated* unit and undoes systemd's `$$` escaping, and `run_hook` (`linux.rs:1266`) clears all three variables first so an inherited value cannot mask a missing guard.
- **The matrix fails in both directions** — deleting the guard fails the five `success`/`killed` rows (`linux.rs:1342-1378`); breaking crash counting fails the twelve fault rows. That is the right shape for a hook with two jobs, and it is the reason this is mergeable.
- **The body's scenario table is accurate.** I checked every row against the guard the diff actually installs; all of them produce the outcome claimed, including the `unset` legacy-systemd row and the `protocol`/`start-limit-hit` row that forwards an empty status.

### Non-blocking, worth a follow-up rather than a re-spin

1. **The skip is broader than "a `systemctl stop`".** systemd cannot attribute the signal, so an external `kill -TERM`, a config-management run, or a userspace OOM daemon such as `earlyoom`/`nohang` (which sends SIGTERM before escalating) lands in the same uncounted bucket. `linux.rs:544-547` frames the combination as "what happens when a `systemctl stop` lands before the handler is installed", which understates the scope; `linux.rs:572-579` does state the underlying fact. Worth rewording where a reader first meets it.
2. **`SuccessExitStatus=` is a second, unguarded input.** It accepts signal names as well as numbers, so a future `SuccessExitStatus=42 43 SIGABRT` would make an abort report `success`/`killed` and this guard would silently stop counting aborts, with nothing failing. A digits-and-spaces assertion on that field, in the spirit of the existing exit-45 pins, would close it.
3. **The residual is `SERVICE_RESULT=timeout`.** A legitimate but slow `systemctl stop` that exceeds `TimeoutStopSec=45` is still counted as a probation crash. The tests pin that as intended (`linux.rs:1411-1427`) and it is defensible, but it is the same user-visible symptom reached by a different result and belongs in the body as a known residual.
4. **`classify_stop` is untouched**, so this fixes systemd >= 232 only; `rollback.rs:700-718` still maps `"TERM"` to `Crash` for the macOS/Windows wrapper paths.
5. Nits: the doubling pin at `linux.rs:1173-1181` matches bare substrings anywhere in the unit, and a comment written in the established `# ... $$EXIT_STATUS is doubled ...` style would satisfy it — pin the full comparison, as the sibling check at `service.rs:1443` already does. `exec_stop_post_script` (`linux.rs:1203-1216`) takes the first `ExecStopPost=` and slices first-quote to last-quote without asserting either is unique.

The lead paragraphs at `linux.rs:518-524` and `:725-731` still state the pre-#5227 rule flatly ("every `$EXIT_STATUS` except 0 and 43") and are corrected only twenty lines later.

None of the above changes the decision. Merging.

Lenses: freenet skeptical-reviewer, freenet code-first-reviewer, plus synthesis. No external (Codex) pass was run; findings were verified against the checked-out code at the head SHA instead.

[AI-assisted - Claude]

sanity:
## Full-tier review complete — two independent lenses, both `merge`

Risk-tiered as Full (auto-update / rollback surface). Lenses: an adversarial skeptical read that **built and ran** code, and a big-picture goal-alignment read. `Claude Rule Review` is green. The external non-Claude pass was attempted (`codex review --base origin/main`) and failed with **"You've hit your usage limit... try again at Sep 7th, 2026 8:14 AM"**; retried once, same result — so per the review rule it was substituted with the extra Claude lens, and I am recording that here rather than letting it pass unnoticed.

### The discriminator, checked against real systemd rather than against the docs

`case "$SERVICE_RESULT $EXIT_CODE" in "success killed") exit 0` at `service/linux.rs:588` and `:795`. The reviewer probed **systemd 255.4-1ubuntu8.17** with transient user units carrying the exact `ExecStopPost` line the template emits:

| event | RESULT | EXIT_CODE | STATUS | hook |
|---|---|---|---|---|
| stop during startup | success | killed | TERM | **skipped** — the fix |
| voluntary update exit 42 | success | exited | 42 | ran — updater intact |
| panic exit 101 | exit-code | exited | 101 | ran |
| external SIGKILL (OOM-shaped) | signal | killed | KILL | ran |
| stop hung past TimeoutStopSec, escalated | timeout | killed | KILL | ran |
| crash (101) **during** a pending stop | exit-code | exited | 101 | ran |
| SIGABRT during a pending stop | core-dump | dumped | ABRT | ran |

The under-counting direction — the dangerous one, since a crash loop misread as a deliberate stop means the rollback never fires — holds up:

- **A pending stop job does not launder a crash.** systemd derives `$SERVICE_RESULT` from how the process actually died, not from whether a stop was requested. That answers the race question directly.
- **A hung shutdown escalated to SIGKILL reports `timeout`, not `success`**, so it is still counted.
- **OOM still counts**: the units set `MemoryMax=2G`, so a memory-regression release dies by cgroup SIGKILL → `signal`/`oom-kill`.
- **No state is introduced at all** — no file, marker, flag or env var of its own. The guard reads two variables systemd sets fresh per stop. The sticky-flag failure mode I was most worried about cannot occur because there is nothing that can go stale.

The one real widening is a fault delivered as TERM/INT/HUP/PIPE, confined to the pre-handler startup window — once the handler is installed, SIGTERM is a graceful exit 0 that was already uncounted pre-PR. It extends an existing blind spot slightly rather than creating a new class.

### Tests

12 of 19 rows assert **crash-still-counted**, which is the load-bearing direction. Verified not vacuous by mutation, run against the tree: reverting the guard fails 2 tests; skipping on `$SERVICE_RESULT` alone fails on the exit-42 row; skipping on `$EXIT_CODE=killed` fails on SIGABRT. The reviewer also wrote an independent 22-row shell replication and got 0 failures on the committed script, 4/1/5 on the three mutants. No test was removed or weakened — the only functional deletions in the diff are the two replaced `ExecStopPost` lines, and the previously-stale "skips only 0 and 43" comments were *strengthened* rather than loosened.

### The accepted regression, judged rather than restated

A release that **hangs** during startup now has no rollback path: pre-PR an operator's `systemctl stop` of a wedged node scored a crash, and three inside probation rolled it back.

Both lenses agree this trade is right, and the second put it better than "the loss is small": **the removed rescue path was never a mechanism, it was an accident of the same ambiguity that caused #5227.** A wedged node's stop and a healthy node's stop are one indistinguishable systemd event. There is no version of "keep counting operator stops as crashes" that does not also reinstate the false-rollback bug. Keeping the old behaviour does not buy hang detection — it buys back false rollbacks of every healthy release an operator stops during startup, which is the worse failure given that a bad release failing to roll back is this project's worst outcome.

#5227 is open, correctly scoped to candidate 1 (install the signal handler before the heavy startup work), and its text already records exactly what this PR does and does not close. **It is unassigned**, and that is the one thing worth acting on: once the miscounting symptom stops being visible it is easy to acquire a false sense that the crash-loop story is finished.

### Also worth knowing

- **The deployed fleet is not fixed at merge.** `ensure_service_file_updated` renders from the template compiled into the *running* binary, so a node gets the new unit when it updates *away from* the first release containing this change — roughly one release of lag. The PR body states this correctly.
- **An unmentioned upside:** pre-PR a machine reboot landing in a node's startup window scored a crash, and three during probation would have rolled back a healthy release. Also fixed.
- No conflict or ordering dependency with the sibling PRs #5240 and #5247 — disjoint file sets, complementary facets of the same subsystem.

No BLOCKING or IMPORTANT code findings from either lens. Merging.

[AI-assisted - Claude]

Issue #5168, opened 2026-08-04
obs: count entries and bytes per fallback summary reply


sanity:
## Problem

#5164 bounds the full-bytes summary fallback with a byte budget (9 KiB primary) and an entry cap (64 secondary). Both constants were sized from fleet measurements — and the first attempt at sizing them was **an order of magnitude off**, because it used the fleet mean bytes-per-entry (143 B) for a distribution whose tail (16.7 KB single-contract summaries) is the entire problem. The error was caught in review, not by data.

That is precisely the situation #5153's discipline calls for: a change whose cost model was argued from an average must ship with a counter that measures the real distribution, so the next sizing decision is made from data.

## Fix

Two fields on the existing outbound rollup, recorded by the fallback send path itself (per the "metric describing a filtering decision" rule — the code that makes the decision emits the count, never a caller re-deriving it):

- entries per fallback reply
- summary bytes per fallback reply

These directly answer: how often does the byte budget bind vs the entry cap, and what is the real per-entry size distribution on fallback links. They also give the before/after for #5161 — as gateway links learn versions and leave the fallback path, these counters should fall toward zero, which is the cleanest possible confirmation that the repair worked.

Raised by the big-picture reviewer on #5164; kept out of that PR to preserve one-logical-change.

Part of #5153. Refs #5164, #5155, #5156, #5161.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-08-14T01:11:18.167622126+00:00
Issue hash: `5095be1e`

## Value Assessment

**Relevance:** Fully live. #5164 (the PR this issue is a follow-up to) is merged; `MAX_FALLBACK_SUMMARY_BYTES_PER_REPLY = 9 * 1024` and the corresponding entry-cap logic sit in `crates/core/src/node.rs:3024` and `:3379-3412`. That loop already computes `entries.len()` and `summary_bytes_used` per reply but discards both — nothing in `tracing/telemetry.rs` records a fallback-specific counter today. The gap described in the issue is real and unfixed.

**User impact:** Indirect but real — this is an observability gap, not a user-facing bug, so no one hits it directly. Its value is entirely in de-risking future sizing decisions. The issue itself documents that the first attempt at these two constants was off by an order of magnitude (using fleet-mean bytes/entry against a fat-tailed distribution) and was only caught by a human reviewer, not data. Without these counters, the next person who needs to retune the byte budget or entry cap is back to guessing from anecdote.

[...]

PR #5243, opened 2026-08-08
fix(interest-sync): bound the digest reply's summarize loop (#5238)

### LINKED ISSUE #5238: Recurrence on 0.2.122: unbounded summarize_contract_state storm (~50-95/sec) with RSS at the 2G cap
## Summary

The `summarize_contract_state` storm from #4610 is back, on 0.2.122, with the same signature: sustained ~50-95 calls/sec against a 30/sec per-callsite cap, and RSS parked at the 2G cgroup cap.

This is the third occurrence in the family: #4473 → #4610 (recurrence on 0.2.84) → now 0.2.122. #4611 gated the symptom, #4612 was filed as the root cause (phantom-hosting: `contract_in_use` true without state) and closed 2026-07-12.

**Not a 0.2.122 regression.** Evidence spans Aug 05 through Aug 08 across versions.

## Evidence

Node: `framework`, a real NATed peer. 0.2.122 (48c166c46ef1), user service, `MemoryMax=2G`, `CPUQuota=200%`, **933 hosted contracts**, 18 stable connections.

The rate-limiter lines are **stderr-only**, so they appear in `journalctl --user -u freenet` and NOT in `~/.local/state/freenet/freenet.*.log`. A clean tracing-log tail is fully consistent with the storm running, which is part of why this went unnoticed.

Long-running process (started Aug 07 21:41 CDT), cumulative counter is per-process and monotonic:

```
Aug 08 08:10:27  summarize_contract_state: dropped  500 in last 30s (cumulative: 132084)
Aug 08 08:17:08  summarize_contract_state: dropped 1655 in last 30s (cumulative: 140121)
Aug 08 08:19:41  summarize_contract_state: dropped 1797 in last 30s (cumulative: 144906)
```

132,084 events accumulated over ~10.5 hours before observation began.

**It resumes immediately after restart.** A fresh process (pid 302455) measured minutes after start:

```
Aug 08 08:27:54  dropped  632 in last 30s (cumulative: 1277)
Aug 08 08:28:25  dropped  767 in last 30s (cumulative: 2044)
Aug 08 08:28:55  dropped 1960 in last 30s (cumulative: 4004)
Aug 08 08:29:26  dropped  149 in last 30s (cumulative: 4153)
```

Dropped 9-60/sec on top of the 30/sec allowed, so ~50-95 attempted calls/sec, sustained rather than bursty. It is user-mode compute: 4.6s utime in 7s wall during a burst, stime moved 10 ticks.

Memory, from systemd's own per-run accounting against `MemoryMax=2G`: Aug 05 hit **2G exactly and swapped**; subsequent runs peaked 1.4G, 1.8G, 1.6G, 1.7G, 1.3G, 1.9G, 1.9G. VIRT climbs 7.9g → 22.8g over minutes, with `Refreshing engine store to reclaim virtual memory lifetime_instances=500` firing every 10-40 min.

## Why this looks like phantom-hosting again

Unverified, from log-timing correlation only, no source trace:

- Full neighbour hosting-state dumps arrive 3-5x/min carrying 74-1046 contracts each, correlating with bursts of ~20 distinct `Module cache miss — compiling contract=X` inside ~0.5s.
- `NEIGHBOR_HOSTING ... overlapping=0` in nearly every sync — only 1 of ~15+ sampled was non-zero (at 118) — across many peers each carrying hundreds of contracts, on a node hosting 933 itself.

That `overlapping=0` pattern is what a phantom population would look like: peers advertising interest in contracts they hold no state for, so overlap computes to nothing while summarize is still driven against stateless contracts. It is consistent with #4612's description of `hosted/in-use` being set without state, and with the known neighbour-overlap churn that re-drives `fetch_contract` on phantom contracts.

Load is flat across dozens of keys, so this is **not** the 0.2.105 single-storm-contract shape that cost-pressure eviction (#4903) addresses.

## Open questions

1. Did #4612's root fix land, and does it cover the path producing these? #4611 explicitly gated the symptom without fixing the phantom-hosting.
2. Is 933 hosted contracts on a laptop itself the anomaly? Eviction is demand-driven and capability-relative, so a memory-bound peer sitting at its cgroup cap suggests the budget is not binding on the resource that is actually scarce here.
3. Is the state-presence gate from #4611 still effective, or is there a second path to summarize that bypasses it?

## Caveats

Single-node observation, no comparison peer (nova's unit returned `MainPID=0` for the query used). Whether the cause is the binary or framework's local state was not determined. Not proposed as a release blocker.

Investigation by a background session looking into an unrelated question (laptop fan noise, which turned out not to be freenet: ~15% of one core on a 20-core machine). Full write-up with raw log lines retained at `/tmp/freenet-framework-findings-2026-08-08.md` on nova.

Related: #4610, #4611, #4612, #4473, #5040, #4565, #4392.

[AI-assisted - Claude]

sanity:
## Problem

The periodic InterestSync heartbeat computed **one contract summary per shared contract, per message, in both directions**, with no bound on the hash-first digest path.

#5155 (PR #5164) bounded the full-bytes fallback and deliberately left the digest path alone, on the reasoning that digests are cheap to *send* — they are, 21 bytes each. But an entry of **either** form costs one `summary_if_hosted_or_in_use` call to *produce*, and that call is a contract-handler round trip that re-enters WASM whenever the memoised state-hash misses. Bounding bytes never bounded that. Two in-tree comments already conceded the gap in as many words:

- `node.rs` — *"Peer is at or above the hash-first floor: 21-byte digests, whole shared set every round. Unchanged by #5155."*
- `node.rs` — *"a digest-capable peer still gets the unwindowed loop on `Interests`."*

Once digest-capability became the common case (#5066 merged 2026-07-31, floor 0.2.116, immediately before the evidence window) a real NATed peer hosting 933 contracts across 18 connections sustained **~55 summarize calls/sec** against the 30/sec per-callsite log limiter, resuming within seconds of every restart, with RSS parked at its 2 GB cgroup cap.

**This is not the #4610/#4612 phantom-hosting recurrence it was filed as.** Every call in the loop is a correctly-gated summarize of a genuinely hosted contract — `should_summarize_or_broadcast` holds — and #4610's `Contract state not found in store` WARN has **zero** occurrences on the affected node, in either journald or the log files. The defect is that nothing bounded *how many*.

## Solution

Bound all four summarize loops in `handle_interest_sync_message` to 64 entries, reusing the rotation #5164 already built rather than inventing a mechanism: random cycle-boundary offset, id-keyed cursor (not an index, so it survives churn), wrapping window, so coverage stays `ceil(n / 64)` rounds.

| Loop | Before | After |
|---|---|---|
| `Interests` reply, digest form | unbounded | 64, rotating |
| `Interests` reply, full-bytes form | 64 / 9 KiB (#5155) | unchanged |
| inbound `SummaryDigests` comparison | 4096 pairs | 64, random offset |
| `SummaryRequest` reply | unbounded | 64, random offset |
| inbound `Summaries` comparison | unbounded | 64, random offset |

The last two were not in the original scope and are consequences of the first two. **The recorded justification for leaving `SummaryRequest` unbounded was that a peer could already force the same work by spamming `Interests`, which ran the same loop.** That was true; it stopped being true the moment both forms of `Interests` were windowed, so leaving it would have left an amplification path *easier* than the one being closed. The full-bytes `Summaries` arm is the same argument and worse — it has neither the digest arm's per-message memo nor its pair dedup.

`ChangeInterests` is deliberately **not** bounded: it is driven by interest churn rather than a clock, so there is no guaranteed next round to rotate into and a window there could defer a newly-added interest indefinitely rather than by a bounded number of heartbeats. That was #5155's reason too and was never written down; it is now.

The rotation helpers lose their `fallback_` prefix, since they serve both forms now.

**This does not touch hosting, placement, or eviction.** It is anti-entropy pacing. Against `.claude/rules/hosting-invariants.md` it touches invariant 1's second freshness layer only; invariant 1's text is updated in this PR, because it described the heartbeat as "unfiltered" and that is no longer true for any peer.

## Convergence

The cost is real and is the thing to review.

**What is delayed.** A stable shared set of `n` contracts is now covered in `ceil(n / 64)` heartbeats instead of one. At the ~450-contract shared sets implied by the field measurement that is 8 rounds, **~40 minutes**; at the worst case of a peer sharing the whole 933-contract hosted set it is 15 rounds, **~75 minutes**. Both numbers are in the tree; the second is on the regression test so the friendlier one cannot be quoted alone.

**Why that is affordable.**

1. **This is the backstop layer, not the delivery path.** A committed update goes to every connected advertised co-host immediately via live fan-out, with no interest check, untouched here. The event-driven repairs — resend-on-failed-patch, request-full-state-when-too-far-behind, `ResyncRequest`, `ChangeInterests` — are all untouched. Anti-entropy only has to catch a divergence that *all* of those missed.
2. **It is an order of magnitude tighter than the cost already accepted on the sibling path.** `MAX_FALLBACK_SUMMARY_BYTES_PER_REPLY` documents cycles **on the order of ten hours** for the heaviest full-bytes links, shipped in #5164.
3. **The digest path cannot degenerate the way the fallback does.** The byte budget is gated off for digests, so every selected entry is always sent and the cursor always advances a full 64. The fallback's "worst case, one contract per round" cannot occur here.
4. **Both directions are independently windowed and heal opposite directions**, so `ceil(n/64)` per direction is conservative — the union covers faster.

**The second-order cost, stated because it is not obvious and it lands on the majority of links.** `SummaryPopulationSource::DigestAgreement` is the only thing that warms our cached belief about a peer's summary for a *quiescent* contract on a digest link, and the broadcast path sends a delta only to a peer whose summary we hold. So for up to `ceil(n/64)` heartbeats after a connection forms, the first update to a not-yet-covered contract **ships full state instead of a delta**. The direction is safe (a missing cached summary makes us send more, never skip, so no update is lost) and `Delivery` seeds the cache after any successful send so only the first update per (contract, peer) pays — but it is a bytes-for-bytes trade against a CPU saving, not purely a latency one. #5164 flagged this for the ~2.8% of links on the fallback; here it applies to the ~98%.

**What truncation costs where it is not cursor-rotated.** The three receive-side caps use a fresh random offset rather than a cursor, so coverage of an over-cap sender's set is coupon-collector — about `(n/64)·H_n` rounds, ~43 for n=450 — not `ceil(n/64)`. Only a pre-#5238 sender emits an over-cap message, so this binds during a rollout and then stops, and in that window the degraded direction is only "we notice that *it* is stale": the old peer's own view is repaired by our cursor-rotated send leg, which it compares in full because it has no cap. Documented on the constant.

**On 64.** Sized against a CPU budget, not a message size: `2 × cap × peers / heartbeat` gives ~7.7 calls/sec at 18 connections, about a quarter of the 30/sec limiter. **128 is the obvious retune candidate and is recorded in-tree as such** — it halves both the convergence and the cache-warm latency at ~15.4 calls/sec, still half the limiter. 64 ships first because the measured node is pinned at its 2 GB cap and the margin is worth more than the latency right now; the equality with `MAX_FALLBACK_SUMMARIES_PER_REPLY` is a symmetry argument, not a budget one, and should not hold the value down once #5168's distribution lands.

**Remaining gap, not closed here.** This is a per-*reply* cap, so the per-*node* rate still scales with connection count; a 200-connection gateway sharing 64+ contracts with every peer models to ~85 calls/sec even at this cap. Capping is monotone — it can only lower what a reply computes — so no peer is made worse off, but an aggregate node-wide budget would need #5168's entries-per-reply distribution first, not another fleet average.

## Measurement

To be validated on `framework`, a real NATed peer carrying the live reproduction, by swapping the binary and comparing like-for-like windows.

### Before (recorded; the reproduction is live)

`framework`, a real NATed peer, running the **same release this would be measured against** — `0.2.123 (cd8f5dfbffd0)`, pid 690799, one continuous process. Window 15:15:24 → 17:54:30 UTC, **2.7 hours**:

| | |
|---|---|
| dropped `summarize_contract_state` log events | 14,854 → 253,691 cumulative |
| rate above the 30/sec per-callsite cap | **25.0 /sec** |
| implied attempted calls | **~55 /sec, sustained** |
| CPU | 842 s over 9,546 s = **8.8% of one core** |
| RSS | peaks at **2,031 MB** against a `MemoryMax=2G` cgroup cap |

The rate limiter drops *log events*, not work, so this counter is a faithful proxy for the call rate rather than an artefact of logging.

### After

**NOT YET RUN.** The binary is built and staged (`0.2.123 (995beade5765)` — the same release as what is on framework, differing only by `cc70301ec`, an unrelated PUT reply-waiter fix, plus this change), the swap/measure/restore procedure is scripted with positive controls, and the whole thing takes ~25 minutes. It is waiting on coordination for the machine, which is in use for release validation.

**Do not read the tests as a substitute for this.** They prove the loops are bounded; they do not prove that these loops are what produced the 55 calls/sec. That inference is from static analysis and a timeline match, and #5238 is the third recurrence in this family — the two prior fixes each looked conclusive too. If the after-rate does not fall substantially, the diagnosis is wrong and this PR should not merge.

One detail that matters for reading the result: if the fix works, the drop counter **stops emitting lines entirely** (the rate falls under the 30/sec cap), and silence is indistinguishable from a dead node. The procedure therefore also samples other rate-limited callsites and recent journal output as positive controls, and records CPU time as an uncensored continuous measure.

## Testing

Six new behavioural tests, all asserting the **expensive operation** (`GetSummaryQuery` round trips) rather than the message, because an entry-count assertion alone stays green under a regression that trims the reply *after* summarizing everything — which is the cheap regression to write and reproduces the storm in full:

- `digest_reply_bounds_the_summarize_calls_not_just_the_entries` — 933 shared contracts, the field case.
- `digest_rotation_covers_the_whole_shared_set` — full coverage in `ceil(n/64)` rounds.
- `digest_reply_records_the_rotation_cursor` — direct, rather than inferred from coverage.
- `digest_receive_leg_bounds_its_summarize_calls` — deliberately disagreeing digests, so nothing short-circuits.
- `summaries_receive_leg_bounds_its_summarize_calls`
- `summary_request_reply_bounds_its_summarize_calls`

**Mutation-tested, six mutations, each killed by the intended test** (committed before mutating): unbound the digest window; bound the *message* but not the *loop*; revert the receive-side cap to 4096; delete the `SummaryRequest` truncation; delete the `Summaries` truncation; restore `if form == FullBytes` on the cursor record. The second is the one that matters — it is the regression an entry-count-only test would miss, and it fails on the `fetches` assertion with the message written for it.

Four pre-existing tests had fixtures shrunk (512 pairs → 63; 32 contracts → 21; 64 repeats → 20) because the new cap would otherwise bind and make their stated premises false. Each was re-checked to still discriminate the mechanism it was written for; the reasoning is in each test's doc comment.

One test was **deleted**: `digest_peer_still_receives_every_shared_contract`, which asserted precisely the behaviour this PR changes.

`cargo fmt`, `cargo clippy --locked -- -D warnings`, and the full `cargo test -p freenet --lib` (4,767 tests) are clean.

### A flaky test, found and fixed

`digest_rotation_covers_the_whole_shared_set` as first written failed **~1% of runs**. A round whose window ends on the highest id leaves the cursor past the end, which `summary_window_start` answers with a fresh random offset — correct anti-starvation behaviour that breaks the exact-tiling assertion — and `GlobalRng` falls back to `rand::rng()` in a plain `#[tokio::test]`, so it was non-reproducible. A reviewer predicted it analytically and the next full-suite run reproduced it. Fixed by seeding the *cursor*, which removes the boundary re-draw from the schedule entirely, so the test pins the tiling property rather than one lucky seed.

**`fallback_rotation_covers_the_whole_shared_set` has carried the identical flake latently since #5164** and is fixed the same way here.

## Review

Two independent blind reviewers, one lens specifically on convergence and anti-entropy behaviour. No blocking defect. Everything they found is fixed in the second commit: the flaky test, the fourth unbounded loop, the peer-summary-cache cost above, and four in-tree comments that asserted the model this change disproves — including one on the very field being renamed, and `MAX_SUMMARY_HASHES_PER_MESSAGE`'s claim that its sites "cannot be raised on one side only" when they are now independently tunable.

That constant's *"per-entry work is a cheap DashMap get"* reasoning is kept **verbatim** and marked as the error that produced #5238, rather than quietly corrected. "A peer cannot make us summarize something we do not have" bounds *which* contracts, not *how many*, and a future editor is better served seeing the mistake than seeing a tidy version that looks like nobody ever made it.

## Follow-up branch

`fix/5238-tests-followup` carries one further commit closing two more review gaps: the interest-registration test now runs against both wire forms (it only covered full-bytes, which was the only windowed form before this change), and `over_cap_digest_message_stays_bounded` gets a known hash so its `SummaryRequest` branch is actually reachable rather than being a no-panic smoke test. It sits on a rebase of this branch onto current main; it is a separate ref only because folding it in would need a force-push. Fold it in however you prefer before merging.

Closes #5238

[AI-assisted - Claude]

---

## Rebase reconciliation (2026-08-16)

This branch had diverged: a local commit, `test(interest-sync): close two review gaps in the #5238 bound tests`, was committed on 2026-08-08 but never pushed, so it was not on this PR. It is now included (it is the third commit). It closes two tests that had quietly stopped exercising what they claim:

- `bounding_the_reply_does_not_bound_interest_registration` ran only against the full-bytes peer. Under #5155 that was the only windowed form; now both are, so it is parameterised over both.
- `over_cap_digest_message_stays_bounded` built its 6,096-entry fixture entirely from hashes this node does not track, so nothing resolved, the `SummaryRequest` arm was unreachable, and it was a no-panic smoke test wearing a bound assertion. It now interleaves a known hash at a period below the cap — appending it as a block does *not* work, because pair dedup collapses the copies to a single pair and the 64-*distinct*-pair scan fills its quota before reaching the block — and the assertion is tightened from "None is acceptable" to "the request must fire and must name exactly that one contract".

The branch was then rebased onto current `main` (`ca5912215`). Two conflicts in `node.rs` were resolved by hand, both additive-vs-additive against main's #5153 telemetry work:

- `Summaries` arm: main's `single_entry` shape classifier is kept and is computed **before** this PR's truncation, so it classifies the length of the message the peer actually sent rather than the windowed remainder.
- `SummaryDigests` arm: main's `peer_key.clone()` is kept (it is needed by the later `emit_stale_peer_syncs` call); only the doc reference to the renamed constant is taken from this branch.

Both resolutions are unreviewed judgment calls by construction, so they were called out specifically to the reviewers.

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -3002,14 +3002,109 @@ const MAX_FALLBACK_SUMMARIES_PER_REPLY: usize = 64;
 /// zero.
 const MAX_FALLBACK_SUMMARY_BYTES_PER_REPLY: usize = 9 * 1024;
 
+/// Maximum number of contract summaries one periodic `Interests` reply computes
+/// on the hash-first DIGEST path (#5238).
+///
+/// #5155 bounded the full-bytes fallback and deliberately left this path alone,
+/// on the reasoning that digests are cheap to SEND. They are — 21 bytes each.
+/// But an entry of either form costs one `summary_if_hosted_or_in_use` call to
+/// PRODUCE, and that call is a contract-handler round trip that re-enters WASM
+/// whenever the memoised state-hash misses. Bounding bytes never bounded that,
+/// so once digest-capability became the common case (#5066 merged 2026-07-31,
+/// floor 0.2.116) the periodic heartbeat was left looping over the WHOLE shared
+/// interest set, on every message, in both directions.
+///
+/// That produced the third `summarize_contract_state` storm in the family
+/// (#4473 -> #4610 -> #5238): a real NATed peer hosting 933 contracts over 18
+/// connections sustained ~54 calls/sec against the 30/sec per-callsite log
+/// limiter, resuming within seconds of every restart. It is NOT a
+/// phantom-hosting recurrence — every call in that loop is a correctly-gated
+/// summarize of a genuinely hosted contract (`should_summarize_or_broadcast`
+/// holds), and the `Contract state not found in store` WARN that was #4610's
+/// signature does not appear. The defect is that nothing bounded HOW MANY.
+///
+/// # Where 64 comes from
+///
+/// The cost bounded here is CPU, so it is sized against a CPU budget rather
+/// than a message size. Per heartbeat and per peer we pay two summarize-heavy
+/// legs: the reply we build here, and the peer's reply that we compare against
+/// our own state in the `SummaryDigests` arm (bounded by
+/// [`MAX_SUMMARY_COMPARISONS_PER_MESSAGE`]). So
+///
+/// ```text
+/// summarize calls/sec ~= 2 x cap x connected_peers / INTEREST_HEARTBEAT_INTERVAL
+/// ```
+///
+/// At 64 and the observed 18 connections that is ~7.7/sec — about a quarter of
+/// the 30/sec per-callsite limiter these storms are measured against, and an
+/// ~7x reduction from the measured 54/sec. It is also the same number as
+/// [`MAX_FALLBACK_SUMMARIES_PER_REPLY`], which restores the property #5155's
+/// split lost: one periodic reply costs the same whichever wire form it takes.
+///
+/// # What it costs in convergence
+///
+/// A stable shared set of `n` contracts is covered in `ceil(n / 64)` heartbeats
+/// (the window rotates and wraps — see `rotation_window_indices`). At the
+/// ~450-contract shared sets implied by the field measurement that is 8 rounds,
+/// or **~40 minutes** for the anti-entropy backstop to come back round to a
+/// given contract, against 5 minutes before. Two reasons that is affordable,
+/// and one caveat:
+///
+/// - This is the BACKSTOP layer, not the delivery path. A committed update goes
+///   to every connected advertised co-host immediately via live fan-out, with
+///   no interest check and untouched by this change; the event-driven repairs
+///   (resend-on-failed-patch, request-full-state-when-too-far-behind) are also
+///   untouched. Anti-entropy only has to catch a divergence that ALL of those
+///   missed.
+/// - It is far tighter than the cost already accepted on the sibling path:
+///   [`MAX_FALLBACK_SUMMARY_BYTES_PER_REPLY`] documents cycles on the order of
+///   ten HOURS for the heaviest full-bytes links, shipped in #5155.
+///
+/// The caveat, stated rather than buried: this is a per-REPLY cap, so the
+/// per-NODE rate still scales with connection count. A 200-connection gateway
+/// sharing 64+ contracts with every peer would model to ~85 calls/sec even at
+/// this cap. Capping is monotone — it can only lower what a reply computes,
+/// never raise it — so no peer is made worse off, but an aggregate node-wide
+/// budget is a real remaining gap. Sizing one needs the entries-per-reply
+/// distribution (#5168), not another fleet average; see
+/// [`MAX_FALLBACK_SUMMARIES_PER_REPLY`] for how retuning off a mean went wrong
+/// last time.
+const MAX_DIGEST_SUMMARIES_PER_REPLY: usize = 64;
+
+/// Maximum number of digest entries one inbound `SummaryDigests` message may
+/// COMPARE against local state (#5238).
+///
+/// The digest exchange has two summarize-heavy legs, not one, and they are
+/// independent: the sender pays [`MAX_DIGEST_SUMMARIES_PER_REPLY`] calls to
+/// build the message, and the receiver pays one call per entry to compute its
+/// OWN summary for the comparison. Bounding only the send side would move half
+/// the storm rather than remove it, and would leave the receiver exposed to any
+/// peer that has not upgraded (or is not honest) sending a full set.
+///
+/// Equal to [`MAX_DIGEST_SUMMARIES_PER_REPLY`] on purpose: a sender running
+/// this release never emits more than that, so between two upgraded peers this
+/// cap never truncates and the two legs cover exactly the same contracts each
+/// round. It binds only on a message from a peer that predates this change.
+///
+/// This REPLACES [`MAX_SUMMARY_HASHES_PER_MESSAGE`] (4096) as the WORK bound on
+/// this arm. That constant is unchanged and still bounds the `SummaryRequest`
+/// input, where it remains the right anti-abuse ceiling.
+///
+/// Over-cap messages are still processed from a random offset (the existing
+/// `rotate_left`), so truncating never starves the tail of a set the sender
+/// happens to emit in contract-id order.
+const MAX_SUMMARY_COMPARISONS_PER_MESSAGE: usize = MAX_DIGEST_SUMMARIES_PER_REPLY;
+
 /// Which wire form a multi-entry `Summaries` reply to a peer takes.
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub(crate) enum SummaryReplyForm {
-    /// Peer is at or above the hash-first floor: 21-byte digests, whole shared
-    /// set every round. Unchanged by #5155.
+    /// Peer is at or above the hash-first floor: 21-byte digests, bounded and
+    /// rotated by entry COUNT (#5238 — #5155 left this form unbounded because
+    /// it was bounding bytes, and digests are 21 bytes each; what stormed was
+    /// the summarize call each entry costs to produce).
     Digests,
     /// Peer is pre-floor, or its version is unknown: full summary bytes,
-    /// bounded and rotated (#5155).
+    /// bounded and rotated by entry count AND by bytes (#5155).
     FullBytes,
 }
 
@@ -3262,38 +3357,36 @@ async fn handle_interest_sync_message(
 
             // #5155: resolve the wire form BEFORE building anything, and take
             // exactly one reading of the version gate (see
-            // `summary_reply_form`). A digest-capable peer keeps getting the
-            // whole shared set every round; only the full-bytes fallback is
-            // bounded, because only it pays per-contract summary bytes.
+            // `summary_reply_form`). The form still decides the ENCODING and
+            // whether a byte budget applies; since #5238 it no longer decides
+            // WHETHER the reply is bounded — both forms are.
             let form = summary_reply_form(op_manager, source);
 
-            // Positions in `matching` whose summaries this reply carries, in the
-            // order they are charged against the byte budget.
+            // Positions in `matching` whose summaries this reply carries, in
+            // the order they are charged against the caps.
             //
-            // `None` means every contract in the set's own order: the digest
-            // path, unchanged, and deliberately WITHOUT materialising an
-            // identity permutation — that would put an allocation the size of
-            // the shared set on the hot path this change claims not to touch.
+            // A bounded window that rotates from where the last reply to this
+            // peer stopped, so successive rounds cover the whole set instead of
+            // re-sending the same prefix forever. The resume point is derived
+            // from the last contract id SENT rather than a stored index, which
+            // is what makes it survive contracts being added and removed
+            // between rounds — see `first_index_after`.
             //
-            // `Some(window)` is the full-bytes fallback: a bounded window that
-            // rotates from where the last reply to this peer stopped, so
-            // successive rounds cover the whole set instead of re-sending the
-            // same prefix forever. The resume point is derived from the last
-            // contract id SENT rather than a stored index, which is what makes
-            // it survive contracts being added and removed between rounds — see
-            // `first_index_after`.
-            let window: Option<Vec<usize>> = match form {
-                SummaryReplyForm::Digests => None,
-                SummaryReplyForm::FullBytes => {
-                    let start = op_manager
-                        .interest_manager
-                        .fallback_window_start(source, &matching);
-                    Some(crate::ring::interest::rotation_window_indices(
-                        matching.len(),
-                        start,
-                        MAX_FALLBACK_SUMMARIES_PER_REPLY,
-                    ))
-                }
+            // #5238: the digest path is windowed too. #5155 left it whole
+            // because it was bounding BYTES and a digest is 21 of them, but
+            // every entry of either form costs one `summary_if_hosted_or_in_use`
+            // call to produce, and that call is what stormed. Only the LIMIT
+            // differs by form, and only the full-bytes path additionally
+            // charges a byte budget on top (see the loop below).
+            let entry_cap = match form {
+                SummaryReplyForm::Digests => MAX_DIGEST_SUMMARIES_PER_REPLY,
+                SummaryReplyForm::FullBytes => MAX_FALLBACK_SUMMARIES_PER_REPLY,
+            };
+            let window = {
+                let start = op_manager
+                    .interest_manager
+                    .summary_window_start(source, &matching);
+                crate::ring::interest::rotation_window_indices(matching.len(), start, entry_cap)
             };
 
             // Register/refresh peer interest for EVERY shared contract, not just
@@ -3338,12 +3431,11 @@ async fn handle_interest_sync_message(
             }
 
             // Build the summaries this reply carries, in rotation order.
-            let reply_len = window.as_ref().map_or(matching.len(), |w| w.len());
-            let mut entries = Vec::with_capacity(reply_len);
+            let mut entries = Vec::with_capacity(window.len());
             let mut summary_bytes_used = 0usize;
             let mut last_included: Option<ContractInstanceId> = None;
-            for pos in 0..reply_len {
-                let contract = matching[window.as_ref().map_or(pos, |w| w[pos])];
+            for &index in &window {
+                let contract = matching[index];
                 // #5155 byte budget. The check is RETROSPECTIVE — it asks
                 // whether the budget is already spent, not whether this entry
                 // would breach it — which is what guarantees the first entry
@@ -3390,12 +3482,13 @@ async fn handle_interest_sync_message(
             //   round of progress. Reserving the window at read time instead
             //   would trade that for a worse failure: a budget-cut round would
             //   then skip everything it did not reach.
-            if form == SummaryReplyForm::FullBytes {
-                if let Some(last) = last_included {
-                    op_manager
-                        .interest_manager
-                        .record_fallback_cursor(source, last);
-                }
+            //
+            // #5238: recorded for BOTH forms now. Under #5155 only the
+            // full-bytes path rotated, so only it had a cursor to advance.
+            if let Some(last) = last_included {
+                op_manager
+                    .interest_manager
+                    .record_summary_cursor(source, last);
             }
 
             if entries.is_empty() {
@@ -3718,15 +3811,23 @@ async fn handle_interest_sync_message(
             let mut request_hashes: Vec<u32> = Vec::new();
 
             if let Some(pk) = peer_key {
-                // See `MAX_SUMMARY_HASHES_PER_MESSAGE`: entries are
+                // See `MAX_SUMMARY_COMPARISONS_PER_MESSAGE`: entries are
                 // peer-supplied and cheap for the peer to fabricate, so
                 // deduplicate by hash and process a bounded window. The window
                 // starts at a random offset when the cap binds, for the same
                 // starvation reason `emit_stale_peer_syncs` rotates — a
                 // sender's entries arrive in contract-id order, so a fixed
                 // prefix would starve the tail on every cycle forever.
+                //
+                // #5238: this cap was `MAX_SUMMARY_HASHES_PER_MESSAGE` (4096),
+                // which bounds the MESSAGE but not the WORK — each entry costs
+                // a `summary_if_hosted_or_in_use` call below, so a routine
+                // heartbeat from a peer sharing hundreds of contracts ran
+                // hundreds of them. This is the receiving half of the same
+                // storm the send-side window closes; both legs of one
+                // round-trip pay it independently.
                 let mut entries = entries;
-                if entries.len() > MAX_SUMMARY_HASHES_PER_MESSAGE {
+                if entries.len() > MAX_SUMMARY_COMPARISONS_PER_MESSAGE {
                     let start = crate::config::GlobalRng::random_range(0..entries.len());
                     entries.rotate_left(start);
                 }
@@ -3837,7 +3938,7 @@ async fn handle_interest_sync_message(
                     // The cap counts distinct PAIRS, matching what is actually
                     // processed — counting distinct hashes would no longer
                     // bound the work now that several pairs can share a hash.
-                    if seen_pairs.len() >= MAX_SUMMARY_HASHES_PER_MESSAGE {
+                    if seen_pairs.len() >= MAX_SUMMARY_COMPARISONS_PER_MESSAGE {
                         break;
                     }
                     if !seen_pairs.insert((entry.hash, entry.summary_digest)) {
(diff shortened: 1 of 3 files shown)


sanity (commented):
## Comprehensive PR Review: #5243

### Summary

- **PR Title:** fix(interest-sync): bound the digest reply's summarize loop (#5238)
- **Type:** fix
- **CI Status:** all checks green on `f433581ae` (14 pass, 2 skipped)
- **Linked Issue:** #5238
- **Review tier:** **Full** — anti-entropy protocol path, concurrency state (per-peer LRU cursors), 875 insertions, and it changes convergence timing
- **Reviewers run:** five independent blind lenses — code-first, testing, skeptical, big-picture, plus a convergence/coverage lens targeted at this change's specific failure mode. External (non-Claude) pass not run; per the standing review rule external models are an opt-in extra, and five diverse lenses were run in its place.

This review is on **reconciled content**. The branch had diverged from origin: a local commit closing two review gaps in the tests was never pushed, and the branch was two commits behind. It has been reconciled, rebased onto `ca5912215`, and two `node.rs` conflicts were resolved by hand. Hand-resolved conflicts are unreviewed judgment calls, so they were called out to every reviewer specifically.

---

### Verdict up front

**No correctness defect found in the production paths.** No reviewer could construct a permanent coverage hole from non-adversarial input. Both hand-resolved conflicts are correct — verified independently by three lenses.

Findings were addressed in the commits pushed after this review; details below.

---

### The two hand-resolved conflicts — both correct

**`single_entry` computed before the truncation** (`node.rs:3700` vs `:3716-3721`). Correct, and provably unobservable either way: the truncation only fires above 64 entries and truncates *to* 64, so `len() == 1` evaluates false under both orderings. It is also the better of two identical choices, since it classifies the length the peer actually sent — matching the sibling digest arm, which reads `entries.len()` after a rotate that reorders but never shortens.

**`peer_key.clone()` retained** (`node.rs:3664`, used at `:3957`). Correct and load-bearing: `peer_key` is consumed 293 lines later by `emit_stale_peer_syncs`. Removing the clone is a compile error, not a silent bug.

---

### The measurement gate — and why it could not be met as designed

The PR set itself a gate: *"if the after-rate does not fall substantially, the diagnosis is wrong and this PR should not merge."* That gate was unmet, and one reviewer initially called it blocking. Investigating it produced the most useful result of this review.

**The instrument the PR designed around cannot produce a rate.** `summarize_contract_state` is an `info_span!` (`contract.rs:2770`), not an event, and the fmt layer has no `with_span_events`. The string reaches a log file *only* via the per-callsite rate-limiter's overflow summary, which fires at 30 events/sec. **It is a threshold detector, not a rate meter — it cannot distinguish 29/sec from 0/sec.** I confirmed zero occurrences on both `framework` and the nova gateway and nearly reported that as "the storm is gone"; it means nothing of the kind.

**There is a continuous instrument, already deployed fleet-wide**, emitted once per node-minute including zeros: the `outbound_message_mix` shadow event. Measured over **~27,500 node-minutes**:

| leg | entries / node-minute | summarize calls/sec |
|---|---|---|
| send (`interest_sync_summaries_interests_reply_entries`) | mean 1859, p50 1021, p90 4825, p99 12603, max 22987 | mean **31.0**, p99 **210**, max **383** |
| receive (identical + differing + one-sided) | mean 959 | **16.0** |
| **combined** | **2818** | **≈47/sec** |

Two conclusions:

1. **The storm is fleet-wide and worse than the filed anecdote.** The mean sits exactly at the 30/sec limiter — which is why the log line appears only intermittently — and p99 is 210/sec against the 130/sec the issue was filed on. This is not one odd laptop.
2. **99.83% of the work finds nothing.** 26,457,275 identical comparisons against 44,305 differing. Anti-entropy burns ~47 summarize calls/sec per node to discover a divergence roughly one time in six hundred.

That second number is the real justification for accepting slower convergence and is stronger than the argument the PR makes. It has been added to the cap's rustdoc.

**Recommendation on the gate:** it was designed on the wrong instrument, and an unrunnable gate cannot block a merge. The diagnosis does not depend on field data anyway — for `SummaryReplyForm::Digests` the window was literally `None` and the loop ran `0..matching.len()`, which is read off the source, not inferred. A follow-up re-measurement on `interest_sync_summaries_interests_reply_entries` would settle the effect size from data already being collected.

---

### Findings addressed in this PR

**1. `ChangeInterests` was left as the cheapest amplification path — MEDIUM, flagged independently by three lenses.** The `added` loop (`node.rs:4585`) iterates a peer-supplied list with no dedup and no length cap, one `summary_if_hosted_or_in_use` round trip each, sequentially on the handler. The PR's own argument for bounding `SummaryRequest` — that bounding `Interests` made the unbounded sibling *easier* to drive than the path being closed — applies here verbatim and was not followed through. Leaving it repeats #5155's mistake one level up: bound what you measured, leave the equivalent you did not.

Fixed by **deduplicating** the hash list. Deliberately dedup and not a window: the recorded reason for not windowing this arm is that it is churn-driven with no guaranteed next round, so a window could defer a genuinely new interest — but a repeated hash carries nothing the first copy did not, so dedup cannot drop anything. Regression test added and **mutation-verified**: with the dedup removed it fails with *"500 copies of one hash made 500 summarize round trips."*

An additional input clip on *distinct* hashes was suggested and **deliberately not taken** — that one can drop a real new interest, which is the failure mode the arm's design avoids. Left to maintainer judgment.

**2. A cross-constant invariant with no guard — LOW-MEDIUM.** The "never binds against a #5155+ peer" claim rests on `MAX_FALLBACK_SUMMARIES_PER_REPLY <= MAX_SUMMARY_COMPARISONS_PER_MESSAGE`, a relationship between two independently-motivated constants. Raising the fallback one alone would silently truncate legitimate replies from upgraded peers with no test failing. Now a `const _: () = assert!(...)`.

**3. A residual argument elsewhere silently recalibrated — MEDIUM.** `broadcast_queue.rs:3381` justified caching a sender-side `Delivered` on the grounds that "the periodic InterestSync summary exchange (~5 min)" bounds the window in which a lost stream tail leaves the cached summary wrong. That number is now ~40 minutes. The second backstop cannot cover it: it fires on a delta that *fails to apply*, and in this case `plan_fanout_send` returns `Skip` on a byte-equal cached belief, so nothing is sent and no delta exists to fail. Comment corrected at the cache site, and the wrong-cached-summary case added to the cap's cost section.

**4. The affordability argument stepped around its own mechanism — MEDIUM.** "Live fan-out, with no interest check and untouched by this change" is true about *interest* but live fan-out is not unconditional — it has a summary gate fed by the cache this change warms more slowly. Reworded to distinguish the missing-belief case (safe: sends more, never skips) from the wrong-belief case (the real cost).

**5. Test-comment corrections.** `over_cap_digest_message_stays_bounded` carried three factually wrong numbers (claimed 2,000 unknowns / ~25% of message / ~1e-8 omission probability; the fixture is 6,096 unknowns with structural interleaving) and framed as probabilistic something now guaranteed — contradicting a correct block 30 lines above. Also noted explicitly that this test is **not** a regression pin for this fix: under a full revert to the 4,096 cap it still passes. It is a repair of a previously vacuous test, worth having on its own terms, but the bound is pinned by the five tests that do fail on a targeted revert.

`bounding_the_reply_does_not_bound_interest_registration` overstated what running both wire forms buys — there is one shared registration loop today, so either case alone would catch a move inside the window. Corrected to say it is insurance against a future per-form split. Its importance is also now stated: `INTEREST_TTL` is 20 minutes against a 40-75 minute coverage cycle, so had that loop been inside the window this PR would have caused steady-state interest expiry and broken live fan-out. **That is the single most important thing the PR gets right**, and it was under-sold.

**6. Seed guards** on the two new `set_seed` calls, so a panicking assertion cannot leave the seed and `THREAD_INDEX` pinned.

---

### Findings deferred, with tracking issues

- **#5338** — the rotation window is narrower than advertised: the cursor is keyed by `SocketAddr` rather than `PeerKey`, so a peer reconnecting from a new source port loses it and degrades to coupon-collector coverage (~90 min at n=450) — and that is precisely the NATed-peer population #5238 was measured on. Plus: non-hosted entries return `None` through a pure in-memory gate yet still consume a slot of a budget that exists to bound CPU.
- **#5339** — no simulation can exercise the cap. Sims run 3-15 contracts against a cap of 64, and the constant is a `const`, so the interest-sync sims produce identical traces before and after this change.
- **#5238 stays open.** At `DEFAULT_MAX_CONNECTIONS = 200` (`ring.rs:555`) the per-reply cap models to ~85 calls/sec, still above the 30/sec limiter. The PR documents this. Sizing a node-wide budget needs #5168's entries-per-reply distribution.

---

### Testing assessment

|---|---|---|
| Unit | ✅ | Five tests fail on a targeted revert, across all four bounded legs plus the cursor-recording line; assertions are on `summary_queries` (the expensive operation), not entry counts |
| Convergence property | ✅ | `ceil(n/64)` pinned across n ∈ {1,5,63,64,65,100,128,266,2448} × five offsets, plus an index-cursor control that provably *does* miss contracts |
| Integration | ✅ | Handler-level end-to-end at n=200 and n=933 |
| Simulation | ❌ | Structurally impossible — see #5339 |

The four shrunk fixtures were checked individually: all retain a premise assertion pinning them under the cap, and none lost a load-bearing assertion. Local run on the final content: 42 passed in `hash_first_summaries`, 141 in the wider `interest` filter.

---

### Verdict

**State:** Ready to Merge
**HEAD SHA reviewed:** `f433581ae`, plus the fix commits pushed after it (re-verified locally: fmt, clippy, tests green; the new dedup test mutation-verified).

The change is monotone — capping can only lower what a reply computes, so no peer is made worse off than `main` — and it addresses a fleet-wide storm measured at ~47 summarize calls/sec per node, of which 99.83% finds nothing. `Closes #5238` is downgraded to a partial fix, since the residual at 200 connections is real and documented.

[AI-assisted - Claude]

sanity (commented):
## Re-review after fixes — PR #5243

The earlier five-lens review's findings were addressed in `21b7b65c1`. That changed 203 lines, which makes the prior review stale by this repo's own rule, so a three-lens re-review ran on the new content: an adversarial pass on the one behavioural change, a doc-accuracy pass, and a test pass. Two further commits came out of it.

**Head now: `5dd3ae9a8`.**

### The dedup is lossless — settled structurally, not assumed

The one behavioural change in the fix commit was deduplicating `ChangeInterests`'s peer-supplied hash list. The obvious worry is that this repo already shipped a bug of exactly that shape: `colliding_contract_hashes_do_not_drop_the_second_entry` exists because hash-only dedup on the `SummaryDigests` arm destroyed real information.

It is the opposite shape here, and the difference is structural rather than a judgement call:

- `ChangeInterests` carries `added: Vec<u32>` (`message.rs:469-474`) — a bare hash list with **no per-entry payload**. Two occurrences of the same `u32` are the same value. `SummaryDigests` carries `(hash, digest)` **pairs**, so two colliding contracts produce entries with the same hash and *different* digests — which is why the fix there deduped on the pair (`node.rs:4233`).
- The first occurrence already covers every colliding contract: `lookup_by_hash` (`interest.rs:2723-2729`) returns a clone of the *entire* `Vec<ContractKey>` for that hash, and the inner loop iterates all of it.
- That Vec cannot contain duplicates — `index_contract_hash` guards with `if !entry.contains(contract)` (`interest.rs:2687-2694`).

So occurrence #2 would re-iterate an identical contract set doing idempotent work. Nothing is dropped.

The `HashSet` is a genuinely new memory term (the old loop was O(1) for a flood of *unknown* hashes). It is bounded at ~2.85 bytes per wire byte against a `Vec<u32>` and a reassembly buffer already paid, so it never becomes the dominant allocation, and it is per-message. Accepted.

### Two commits the re-review produced

**`b39bec471` — the dedup test's assertion was far too weak.** I first wrote `fetches < repeats`. That passes under an **inverted** dedup condition (`if seen_added.insert(hash)` skips only the *first* occurrence): 499 of 500 round trips survive, the amplification path is essentially wide open, and the assertion is satisfied. Any batched dedup collapsing to `repeats/K` passes too.

The fixture advertises one hash resolving to one contract, so correct dedup costs exactly one round trip. Now pinned as `assert_eq!(fetches, 1)`, and **mutation-verified in both directions**: inverted condition fails at 499 (the old assertion passed this), dedup removed entirely fails at 500.

That my own first attempt was the weak one is the argument for the re-review, not against it.

**`5dd3ae9a8` — a false premise and the last unguarded seed.** The test asserted "premise: fixture hosts exactly one contract"; the node hosts two, because `build_harness` hosts its own `h.key` before `host_many` adds one. The conclusion held but the stated reason did not — only `host_many`'s hash reaches `added`. Reworded to what it actually guards. Separately, `over_cap_digest_message_stays_bounded` still carried the last bare `set_seed` in the file, pinning THREAD_SEED/THREAD_RNG/THREAD_INDEX and never clearing on either path — the #5314 cross-test-interference class that plain `cargo test` can see and `cargo nextest` structurally cannot. Now guarded.

### Verified clean

- **All eight new documentation claims** checked against the code at that exact blob, including that `plan_fanout_send` really does short-circuit to `Skip` on a byte-equal cached belief so the ResyncRequest backstop genuinely cannot fire for that case (`broadcast_queue.rs:565-566`), and that there is literally one interest-registration loop shared by both wire forms (`node.rs:3549-3551`).
- The `const _: () = assert!` is the right inequality in the right direction, trips in both directions, and uses a form already in-tree (`transport/fixed_rate/controller.rs:48-51`).
- The `seed_guard` conversions are `let _seed`, not `let _` (which would drop immediately and clear the seed before the body ran), and are declared before the harness so the seed clears last.

### Residual, not blocking

The "never binds against a #5155+ peer" note enumerates two emitters; `ChangeInterestsReply` also produces a `Summaries` and its entry count is peer-driven. Theoretical — production emits one contract per `ChangeInterests` (measured mean 1.000) — and the new dedup strictly *reduces* that exposure, since entries are now bounded by distinct hashes where repeats previously inflated them.

Unrelated and pre-existing, flagged because it came up while bounding the memory question: `InboundStream::new` does `Vec::with_capacity` from a peer-declared `u64` with no cap found under `crates/core/src/transport/` (`inbound_stream.rs:36-42`). Not this PR's business, and it means the `HashSet` is nowhere near the cheapest allocation lever a peer already has.

### Verdict

**Ready to merge** once CI is green on `5dd3ae9a8`. No correctness defect found across eight lens-passes over two rounds; every finding is fixed, and the two behavioural pins are mutation-verified rather than merely passing.

[AI-assisted - Claude]

sanity:
**External model pass (Codex) — clean.**

Run per the Full-tier requirement, `codex review --base origin/main` against `5dd3ae9a8` in an isolated worktree:

> The new per-message bounds, rotating summary cursor, receive-side truncation, and ChangeInterests deduplication are internally consistent and covered by targeted regression tests. No actionable correctness regressions were identified.

One caveat worth recording rather than glossing: its attempt to run `cargo test -p freenet --lib node::tests` was blocked on the cargo artifact lock (a concurrent build of mine held it), so its pass was static analysis only. The test evidence comes from the local runs instead — 42 in `hash_first_summaries`, 141 in the wider `interest` filter, both dedup mutations confirmed red.

Full review record: five blind Claude lenses on the original content, three on the fix commit, plus this external pass. Every finding fixed or explicitly deferred with a tracking issue (#5338, #5339). CI green on `5dd3ae9a8`.

[AI-assisted - Claude]

PR #5247, opened 2026-08-08
fix(update): give the mute update process a voice, and stop two safety states going unrecorded (#5244)

### LINKED ISSUE #5244: brick-safety: `freenet update` installs no tracing subscriber, so rollback can fail to arm with no trace anywhere
## Summary

The `freenet update` process **never installs a tracing subscriber**, so every `tracing::warn!` / `error!` / `info!` inside it is a no-op — not just `debug!`. The supervisor invokes it as `freenet update --quiet`, so any site whose only output is a `tracing::warn!` plus a `!self.quiet`-gated `eprintln!` produces **no output anywhere on the automated path**.

The consequence that matters: **an update can install with crash-loop rollback protection never armed, and nothing records it.** The trigger is an unwritable or full state directory — a fleet condition, not a freak event.

This is not a logging gap. #4073's entire purpose is to stop a bad release bricking the fleet, and this is a path where that machinery is off and invisible. Breaking auto-update is close to the worst failure this project can have, precisely because the fix cannot deliver itself; a silent failure to arm rollback is the same class one layer down.

Found while investigating #5232, which turned out to be a misreading of the journal caused by this same visibility gap.

## The structural cause

`freenet::config::set_logger` is called exactly once in the binary — `crates/core/src/bin/freenet.rs:1123`, inside `run_node`, which is reached only for `Network` / `Local` / no-subcommand. `freenet_main` (line 1172) dispatches:

```rust
Some(Command::Update(cmd)) => cmd.run(build_info::VERSION),   // freenet.rs:1191
```

with no subscriber installed, and there is no other subscriber initialisation anywhere under `crates/core/src/bin/`. With no global default, every `tracing::*` macro in that process is a no-op at every level.

Now add the supervisor's invocation, from the generated unit (`crates/core/src/bin/commands/service/linux.rs:539`):

```
ExecStopPost=-/bin/sh -c 'case "$EXIT_STATUS" in 0|43) ;; *) FREENET_POST_STOP_EXIT_CODE="$EXIT_STATUS" <binary> update --quiet ;; esac'
```

So on the path that actually runs in production, both channels are closed.

This also explains why the #5232 investigation was misled: `handle_post_stop`'s outcome arms use **unconditional** `eprintln!` ("Critical brick-safety signal: emit on stderr even under `--quiet`"), so crash counts and rollbacks *do* reach the journal — while the decisions that cancel them do not. The journal showed strikes accumulating and nothing else, which reads as a broken mechanism.

## Verified findings

I read each of these directly.

### 1. An update can install with no rollback protection, silently

`crates/core/src/bin/commands/update.rs:703-718`

```rust
let known_good = match rollback::prepare_known_good_for_install(current_version, &current_exe) {
    Ok(prepared) => Some(prepared),
    Err(e) => {
        if !self.quiet {
            eprintln!("Warning: failed to snapshot known-good binary for crash-loop \
                       rollback: {e}. Proceeding without rollback protection for this update.");
        }
        tracing::warn!(error = %e, "Failed to capture known-good rollback binary (#4073)");
        None
    }
};
```

`known_good = None` means the `if let Some(...)` guard at line 745 is skipped, so `begin_probation` is never even attempted. The update lands, brick-safety is never armed, and on the automated path neither the `eprintln!` (gated) nor the `warn!` (no subscriber) emits anything.

The sibling case at `update.rs:746-759` — `begin_probation` itself failing — has the identical shape and the identical silence.

### 2. A permanently locked-out node says nothing, forever

`crates/core/src/bin/freenet.rs:872-877`

```rust
} else {
    tracing::debug!(
        "Periodic re-poll: skipped — auto-update locked out after repeated \
         failed installs (#3934); run `freenet update` to recover"
    );
}
```

This one runs in the **node** process, where tracing works — but `debug!` is compiled out by `release_max_level_info` (`crates/core/Cargo.toml:124`). The loud once-per-process `LOCKOUT_WARNED` warning (`auto_update.rs:403-412`) is only reachable from the peer-signal triggers, so on a node whose peers are all the same version this `debug!` is the *only* trace that auto-update is permanently off. Every 6 hours, forever, silently.

### 3. The update-failure lockout fails OPEN on an unreadable counter

`crates/core/src/bin/commands/auto_update.rs:1054-1058`

The doc comment immediately above it states the intended contract:

> * Missing file → `0` (legitimate "no failures yet").
> * Present but unparseable → `MAX_UPDATE_FAILURES` (defensive: if the counter file has been truncated or corrupted we must NOT silently reset the lockout — that would be an amplification vector for any process that can partially overwrite the file, defeating the #3934 fix.)

The code defends only the *parse* path:

```rust
match fs::read_to_string(dir.join("update_failures")) {
    Ok(s) => s.trim().parse().unwrap_or(MAX_UPDATE_FAILURES),
    Err(_) => 0,
}
```

`Err(_)` catches every read error, not just `NotFound` — EACCES, EIO, a directory in its place. All of them read as "no failures" and silently reset the lockout. Making the file unreadable defeats #3934 more easily than partially overwriting it, which is the attack the comment names.

Twenty lines away in the same file, `read_github_poll_bucket_at` (`auto_update.rs:1239-1242`) gets this exactly right, and says why:

```rust
Err(e) if e.kind() == std::io::ErrorKind::NotFound => BucketRead::Missing,
// Any other read error (permissions, etc.) is treated as corrupt =>
// deny, conservatively, rather than granting a free poll.
Err(_) => BucketRead::Corrupt,
```

Same file, same intent, opposite behaviour.

## Why these three compound

They share a trigger. If the state directory becomes unwritable or full:

1. the known-good capture fails → the update installs with no rollback protection (finding 1), silently;
2. installs keep failing → but the failure counter cannot persist, and an unreadable counter reads as zero (finding 3), so the #3934 lockout never engages;
3. the node keeps exiting 42 in a loop → and the one line that would say so is compiled out (finding 2).

Every guard in the chain is disabled by the same condition, and the whole sequence runs without a single line in the journal.

## Unverified — reported by an audit pass, not individually spot-checked

Same class, same file, listed so they are not lost. Each needs confirming before acting.

- `update.rs:441-462` — installer refuses a known-bad-pinned or install-gated version and exits "already up to date". Silent on the automated path. This is the terminal state of the whole #4073 machine.
- `update.rs:1254-1265` and `auto_update.rs:1291-1296` — the GitHub poll rate-limit bucket denies a poll. Silent. An unwritable state dir means the node never polls again and is indistinguishable from a healthy up-to-date node.
- `auto_update.rs:1076-1080` — `let _write = fs::write(...)` discards the lockout counter's write result. Note `record_install_failure_at` (`rollback.rs:711-720`) *does* `warn!` on the same failure; the sibling counters disagree.
- `update.rs:1476-1490` — a release published without `SHA256SUMS.txt.sig` installs anyway under the `REQUIRE_RELEASE_SIGNATURE = false` transition, and the notice is `!quiet`-gated. A waived signature check should never be quiet, and the silence also destroys the evidence needed to decide when it is safe to flip that flag to `true`.
- `update.rs:493-505` — nothing marks the moment a version becomes install-gated, which is the single most operator-relevant event in this subsystem.
- `auto_update.rs:1260-1283`, `1415-1421`, `994-1029` — corrupt-bucket deny, cooldown persist failure, and poll-cadence state, all `debug!` or swallowed.
- `service.rs` is a fourth writer of the `update_failures` counter and was outside the audit's scope; worth a pass with the same lens.

## Suggested fix

Two candidate shapes, and they are not mutually exclusive:

1. **Install a tracing subscriber in the `freenet update` process.** Fixes the class rather than the instances we happen to have noticed, and cannot be forgotten by the next person who adds a `warn!`. Needs a decision on where its output goes, since the point is to reach the journal.
2. **Drop the `!self.quiet` guard on brick-safety-relevant messages in `update.rs`**, following the `handle_post_stop` precedent that already works and is already documented as deliberate.

Whichever is chosen, **"installed without rollback protection" and "locked out of auto-update permanently" must never be quiet, on any path, under any flag.** Finding 3 is a separate one-line fix: match `NotFound` explicitly and treat other read errors as `MAX_UPDATE_FAILURES`, matching both the doc comment and the sibling reader.

## Related

- #5232 — the misreading this came out of. Its stderr-visibility fix is in #5240, which addresses one instance (the `ClearedStale` branch silently disarming rollback) but not this structural cause.
- #3934 — the lockout that finding 3 defeats.
- #4073 — the crash-loop auto-rollback epic.

[AI-assisted - Claude]

sanity:
## Problem

`freenet update` ran with **no tracing subscriber at all**. `set_logger` is called only on the node path (`run_node`), so `freenet_main`'s `Update` arm installed nothing and every `tracing::warn!` / `error!` in the installer was a no-op — not just `debug!`. The supervisor invokes it as `freenet update --quiet`, so any site whose only output was a `warn!` plus a `!quiet`-gated `eprintln!` produced **nothing anywhere** on the automated path.

The consequence that matters: **an update can install with crash-loop rollback protection never armed, and nothing records it.** The trigger is an unwritable or full state directory, which is a fleet condition rather than a freak event. #4073 exists to stop a bad release bricking the fleet, and this is a path where it is off and invisible.

Found while investigating #5232, which turned out to be a misreading of the journal caused by this same gap. Full analysis in #5244.

### 1. Give the process a voice

`set_cli_logger` installs a stderr-only subscriber at WARN for the `Update` arm.

**Two traps decided its shape, and both produce a fix that looks done and changes nothing:**

- Passing a log dir — the natural copy-paste from `run_node` — sets `use_file_logging`, routing everything into the rolling log files. systemd captures stdout/stderr, **not** those files; that asymmetry is the whole of #5232. Every `warn!` would start "working" and the journal would stay exactly as blind.
- With no log dir, `init_tracer` still falls through to *stdout* unless `FREENET_LOG_TO_STDERR` happens to be set.

So `init_cli_stderr_tracer` is explicit rather than configuration-dependent: stderr, always.

It is **additive** rather than a flag on `set_logger`. That function serves the node, where output must keep reaching the log files (`freenet service report` collects them, and on Windows nothing captures stdout). Adding a "log to stderr instead" switch to the node's only logging call would put a mis-settable lever on the one path that must not lose files.

WARN, not INFO: this runs on every non-clean stop, so an INFO default would flood the journal of exactly the crash-looping node whose journal we need to read.

**This fixes the class — on systemd.** Every other `tracing::warn!` in the update path (the blocked-version refusal, the rate-limit exits, the install-gate counter) becomes visible without touching its call site.

**It does NOT reach the field on Windows or the macOS tray path.** `spawn_update_command` nulls stdio on every platform (load-bearing on Windows: the wrapper has called `FreeConsole()`, and inheriting invalid handles fails `spawn()` with os error 6). So under those supervisors the new subscriber output and both unconditional messages go to the bit bucket. Its doc comment also claimed nulling was harmless because "`--quiet` already suppresses all output", which this PR makes false; that comment is corrected here with the shape of the real fix recorded. The fix itself is deliberately NOT in this PR: it needs a log dir threaded through five call sites plus a stdio change on the code path behind three prior os-error-6 incidents, and I cannot exercise Windows or macOS from here. Tracked in #5244.

### 2. Two messages that must never be quiet

Both are states where the machinery designed to stop a bad release bricking the fleet is silently OFF, so both also get an **unconditional** `eprintln!` — belt-and-braces, surviving any later refactor that drops the subscriber. This follows `handle_post_stop`'s existing precedent, which is why its crash and rollback lines were visible in the field while the rest of the file was not.

- "failed to snapshot the known-good binary" — the update is about to land with no rollback target.
- "installed the update but FAILED TO ARM crash-loop rollback" — it landed and the marker is absent, so the post-stop hook finds nothing to count.

Deliberately **only those two**. A general de-quieting would leave `--quiet` meaning nothing. The separation this restores: `--quiet` controls chattiness, never whether a safety event is recorded.

A third, "auto-update is LOCKED OUT after repeated failed installs", was `debug!` — compiled out of release builds — and the loud once-per-process warning in `check_if_update_available` is reachable only from the peer-signal triggers, so on a node whose peers all share its version nothing said it would never update again. Now warn + stderr, once per process, mirroring that existing `LOCKOUT_WARNED` pattern.

### 3. A fail-open on the lockout (second commit, reviewable on its own)

`get_update_failure_count_at` matched `Err(_) => 0`, catching every read error and not just `NotFound`. A counter made unreadable — EACCES, EIO, a directory in its place — read as "no failures" and silently reset the #3934 lockout.

The doc comment directly above the function already promised the opposite and named the threat ("an amplification vector for any process that can partially overwrite the file"). Only the parse path was defended; making the file unreadable defeats it more easily than partially overwriting it. `read_github_poll_bucket_at`, twenty lines away in the same file, already makes this distinction and explains why. Same file, same intent, opposite behaviour.

Note the states that reach this arm are anomalous ones — a chmod'd or wrong-owner file, a directory in its place, a failing mount — not the read-only/full state dir case, where reads still succeed or report `NotFound`. Treating them as fully-failed is self-healing: the gate re-reads every time and recovers the moment the file is readable again.

## Why these compound

One bad state directory disabled all three guards at once: rollback never arms (silent), the failure counter can neither persist nor read back (silent, and fails open), and the line that would say the node is locked out was compiled out. The whole sequence ran without one line in the journal.

## Testing

All deterministic and offline. The runtime tests seed the installer's GitHub cooldown file, so `probe_latest_tag` defers with exactly one known WARN before any HTTP — no network dependency, no quota spent, and the assertion is on a WARN rather than a TRACE so it survives `release_max_level_info` (an earlier revision asserted TRACE and would have passed in CI's debug build while failing on the build we ship).

| Test | Holds |
|---|---|
| `a_warning_from_the_update_process_reaches_stderr` | A WARN reaches stderr and stdout stays clean. Before the fix both streams were empty whatever `RUST_LOG` said. |
| `the_update_process_does_not_write_ansi_escapes_to_a_pipe` | journald stores bytes verbatim, so escapes must not reach a non-tty. |
| `the_update_arm_installs_the_cli_logger_at_warn_and_not_a_file_logger` | The wiring AND the level — differences no runtime test can see, since a log-dir subscriber still shows output on a terminal, and nothing on this path logs at INFO so a "quiet at default" test would pass under a level promotion. |
| `an_unreadable_failure_counter_does_not_reset_the_lockout` | The fail-open. Uses a directory rather than a chmod so it fails for root too and cannot degrade into a no-op in a container. |
| `an_unreadable_counter_is_never_persisted_as_a_count` | The regression review caught: an unreadable counter must write nothing, or one transient EIO locks the node out permanently. |
| `an_absent_failure_counter_reads_as_no_failures` | The legitimate case still reads zero, so a fresh node is not born locked out. |

Mutation-tested; each kills the guard that names it:

| Mutation | Result |
|---|---|
| Remove the subscriber from the `Update` arm | runtime test FAILED, pin FAILED |
| `set_logger` instead of `set_cli_logger` (**the trap**) | runtime test FAILED, pin FAILED |
| Restore `Err(_) => 0` on the counter | unit test FAILED |

The second is the one worth noting: it is the implementation a reasonable person would write, it compiles, it makes `warn!` "work" — and both guards reject it, because the runtime test finds the output on stdout instead of stderr and the pin sees the wrong call.

The source pin is scraped **across files** (test in `tests/`, source in `src/bin/`) so it cannot be satisfied by its own assertion strings — the self-matching failure mode documented in `.claude/rules/bug-prevention-patterns.md`.

Also verified by hand: `RUST_LOG=trace freenet update --check --quiet` now writes 24 lines to stderr and 0 to stdout, and is silent at the default level.

## Not in scope

#5244 lists further sites in the same class that this PR does not touch — the install-gate threshold moment, the rate-limit denies with no log at any level, the swallowed counter write, and the `!quiet`-gated unsigned-release allowance. Most become visible through the subscriber; the ones with no logging at all still need their own lines.

Closes #5244

[AI-assisted - Claude]


sanity:
## Review response — two lenses, one real regression, two tests that could not fail

Skeptical and code-first, both blind. Between them they found something I had introduced, two assertions that could not fail, and a claim in the body that was true only on Linux. All addressed in `d45fe5a2f`; the body is corrected too.

### The regression I introduced

Making the counter read fail-closed was right for the **gate** — an unreadable counter suppresses updates while the condition lasts, and heals by itself because the gate re-reads every time. But `record_update_failure_at` fed that same defensive answer back through `+ 1`, so **one transient EIO would have written 4 to disk and locked auto-update off permanently.** Worse than the fail-open it replaced.

The decision is now `next_failure_count`, a pure function that writes nothing when the counter is unreadable. Extracting it is what makes it testable at all — the alternative is contriving a filesystem that reads one way and writes another.

I also had an unsupported claim in the doc comment and the body: "an unwritable state dir is what makes installs fail, so the counter is unreadable exactly when it is needed." Not true — EROFS/ENOSPC break *writes*; reads still succeed or report `NotFound`. The states that actually reach that arm are anomalous ones (chmod'd file, wrong owner, a directory in its place, a failing mount). Corrected in both places.

### Two tests that could not fail

- `the_update_process_is_quiet_when_nothing_is_wrong` claimed to pin WARN-vs-INFO, but **nothing on the `--check` path logs at INFO**, so promoting the level would have left it green. Deleted; the level is pinned in the source scrape instead, where it is a fact about the code rather than about which paths happen to log.
- The source pin scraped a region that included the arm's own comment block — one editorial pass from being satisfied by a comment *about* `set_cli_logger`. Comments are now stripped before matching.

### Two tests that would have gone red on a network blip

Both ran `update --check`, which reaches api.github.com — so they depended on the network being up **and** GitHub not throttling a shared CI egress IP, and spent real quota twice per run, which #5102 went to some trouble to conserve. They now seed the installer's GitHub cooldown file, so `probe_latest_tag` defers with exactly one known WARN before any HTTP. Deterministic, offline, and the whole file now runs in 0.01s.

That fixture fixed a subtler problem too: the old assertion looked for TRACE output from the HTTP stack, which `release_max_level_info` compiles out entirely. It passed only because CI builds debug, and would have failed on the build we ship. A WARN survives both.

A third test in that revision, on `RUST_LOG` escalation, rested on a premise that the cooldown fixture invalidates (no HTTP means DEBUG adds nothing). Removed rather than contorted.

### A claim that was true only on Linux

"This fixes the class" is systemd-only. `spawn_update_command` nulls stdio on every platform, so under the Windows wrapper and the macOS tray path the new subscriber output and both unconditional messages go to the bit bucket. Its doc comment also said nulling was harmless because "`--quiet` already suppresses all output" — which this PR makes false. Comment corrected with the shape of the real fix recorded.

Not fixed here, deliberately: it needs a log dir threaded through five call sites plus a stdio change on the code path behind three prior os-error-6 incidents, and I cannot exercise Windows or macOS from here. Shipping that untested is exactly the kind of change that has bitten this file before. Body now states the limitation; #5244 tracks it.

### Also fixed

- **ANSI into the journal.** The CLI tracer went through `init_stdout_tracer`'s `pretty` layer, which colours unconditionally and splits each event over several lines. journald stores bytes verbatim. It now builds its own compact layer with ANSI gated on `stderr().is_terminal()`, with a test asserting no escapes reach a pipe.
- **A comment promising something the code did not do.** `set_cli_logger` said a failed subscriber install must not fail the update, while the only realistic failure panicked inside `init_stdout_tracer`. The new layer returns the error instead, so the promise holds.
- **Operator advice that does not work.** The lockout message said to run `freenet update` to clear the counter — which cannot clear what it cannot read. It now names the file and `--force` as well.
- `CARGO_BIN_EXE_freenet` instead of a hand-rolled `target/debug` lookup that would happily assert against a stale binary — the failure mode this test exists to prevent. `LOCKOUT_REPORTED` moved next to the branch it guards. Test env hygiene: `FREENET_DISABLE_LOGS` and friends are now removed so a developer's environment cannot produce a vacuous pass.

### Not taken

- **Rate-limiting the CLI subscriber.** Raised because `RUST_LOG=debug` in the unit would produce unbounded output. An operator who explicitly asks for debug on a specific node should get it; the default is WARN and the process is short-lived.
- **`clear_update_failures_at`'s swallowed `remove_file`.** Real, and it is why the lockout message now names the file directly — but it is one of the several swallowed writes #5244 lists as unfixed, and fixing them piecemeal inside this PR would blur what it is.

[AI-assisted - Claude]

sanity:
Resolved the blocking `rule-review` warning — but not the way it asked, and the disagreement is the substance.

## The flagged behaviour is right; the docs were the bug

The warning was that `next_failure_count` rewrites an unparseable counter as `1` while `get_update_failure_count_at` reads the same file as `MAX_UPDATE_FAILURES`, and proposed persisting `MAX + 1` to make them agree.

That would be worse, for two reasons.

**`fs::write` is not atomic.** A power cut during `record_update_failure_at` leaves a truncated — hence unparseable — counter. Persisting `MAX + 1` for that turns one ill-timed crash into a permanent auto-update lockout, clearable only by deleting the file by hand. That is the failure this whole subsystem exists to prevent, and it is strictly worse than the two extra attempts the current behaviour costs.

**It is not the amplification vector the read-side doc warns about.** Reaching the rewrite requires a recorded failure, which requires an update attempt — and while the counter is unparseable the gate is shut, so no automatic attempt happens. Meanwhile a process that can *choose* the file's contents does not need any of this: it writes `0`. Corruption alone only ever shuts the gate.

So the asymmetry stands. What was genuinely wrong is that the two functions' doc comments contradicted each other — one promising that corruption must never reset the lockout, the other resetting it — with nothing explaining why. That is now written down on both sides, since it was easy enough to read as an inconsistency that it cost a reviewer a finding.

The new test pins the property that actually makes it safe, which is that reopening is **bounded**: the gate shuts on a damaged counter, recording rewrites it as `1` rather than `MAX + 1`, and `MAX` further failures put the lockout back. Without that assertion, "resets to 1" really would be a way to get unlimited attempts out of a corrupted file.

## The `Unit & Integration` red was real, and its guard was too blunt

`init_tracer_wires_every_error_layer_through_build_error_filter` failed because it pins `.from_env_lossy()` to exactly one caller, and this PR's `init_cli_stderr_tracer` is a second one.

That caller is legitimate — it is a console filter, and `RUST_LOG` is the operator's console knob, which is exactly what the guard means to permit. But simply bumping the count to 2 would leave a weaker guard than before, since two callers could be the wrong two. So each permitted caller is now pinned to its own function body (`build_filter` for the node's console layers, `init_cli_stderr_tracer` for the CLI's), and the total is pinned at 2 on top, so a third caller anywhere still fails rather than sliding in unnoticed. What the guard protects against — an error layer inheriting `RUST_LOG`'s level and losing the WARN floor (#5015) — is unchanged.

## Not changed

The `compare_exchange` `Release`/`SeqCst` ordering flagged as Info. The adjacent, pre-existing `set_logger` does exactly the same thing, so changing only the new one would create the inconsistency rather than remove it — and a failure ordering stronger than success has been permitted since Rust 1.64. Not worth the churn here; a follow-up touching both would be fine.

## Verification

Honestly: **against CI, not locally.** This machine is at load 74 with several agents compiling and a local `cargo test` made no measurable progress in ten minutes, so I pushed and am relying on the runners. I will not merge until the required checks are green on this head.

[AI-assisted - Claude]

sanity:
Correcting my previous comment, and fixing the failure it caused.

**The regression test I described was not actually added.** The command that would have written it was rejected by a pre-execution hook, and that rejects the whole command — so only the doc reconciliation landed, while my commit message and comment both described a test that did not exist. That is my error, and worth stating plainly rather than quietly adding it now.

It is there in `b0f2c5032`, and it is mutation-verified rather than merely passing: applying the `unwrap_or(MAX_UPDATE_FAILURES)` change the rule review asked for turns it red along with `an_unreadable_counter_is_never_persisted_as_a_count` (`left: Some(4), right: Some(1)`), and both pass with the committed code. So the test genuinely discriminates the behaviour under discussion rather than pinning whatever is there.

### The `Unit & Integration` red was my pin, and the rule review was right about why

`init_tracer_wires_every_error_layer_through_build_error_filter` failed on `build_filter is the node's console filter and must read RUST_LOG`. The cause is exactly what the rule review identified: my `body_of` helper ended its slice at the next `fn`/`pub fn` instead of at the closing brace.

Measured against the real file, that is not a marginal difference:

| | body size | `.from_env_lossy()` found |
|---|---|---|
| old next-`fn` heuristic | **14,944 chars** (the whole rest of `init_tracer`) | 2 → assertion fails |
| brace-matched | 283 chars | 1 |
| `init_cli_stderr_tracer`, brace-matched | 2,067 chars | 1 |

AGENTS.md names this shape specifically, and the file's own `fn_body` helper cannot be reused here because it refuses indented definitions — `build_filter` is nested inside `init_tracer`. So brace matching, plus two over-swallow detectors that reject a body containing the test-module attribute or a following `pub fn`, so a later edit cannot quietly widen it back.

The weaker guarantee the rule review objected to — "somewhere between this function and the next there is one call", which can pass vacuously after an unrelated edit — is gone. The production slice holds exactly 2 calls in total, so all three assertions hold.

Verification, honestly: the counter test was built and run locally, including its mutation. The tracer guard was verified by replaying the extraction logic against the actual file (the numbers above) rather than by a local compile — the machine is at load ~60 with several agents building and the lib build made no progress in twenty minutes. CI is the check on that one, and I will not merge until it is green.

[AI-assisted - Claude]

PR #5249, opened 2026-08-08
feat(obs): count contract-exec WASM calls apart from the cache hits

sanity:
## Problem

Contract-exec WASM volume is not measurable in production, so none of the last month's storm work can be evaluated.

The only production signal for `summarize_contract_state` is a span opened at **handler entry** (`contract.rs`, `info_span!("summarize_contract_state", %key)`). It fires identically for a state-hash cache HIT and for a real WASM `summarize_state` invocation. The cache is the entire reason the observed cadence is survivable, so the two differ by orders of magnitude in cost — and the span cannot tell them apart.

The published "rates" are not even that span's count. Every figure quoted across #4473 (~40/sec), #4610 (~70-80/sec), #5040 (~49-57/sec) and #5238 (~50-95/sec) comes from the **dropped-line counter of a tracing rate limiter** (`util/rate_limit_layer.rs`), which caps LOG EVENTS at 30/sec per callsite and has no effect on work done.

So nobody can currently say whether a peer at ~50/sec is doing 50 WASM invocations or 50 cache hits. Five PRs in five weeks bounded or gated call sites against that undifferentiated number.

A counter for exactly this **already sat on the WASM slow path** (`executor_impl.rs`, with a comment describing the invariant) and was a no-op in the field: it writes only to `topology_registry`, which keys on a thread-local that `SimNetwork` sets and a production node never does. The simulation asserted the invariant held, the field suggested it did not, and the counter that would adjudicate was switched off precisely where it mattered.

## Approach

`ring::contract_exec_metrics::ContractExecMetrics` — eight monotonic per-node counters, each recorded **by the code that makes the decision** rather than re-derived by a caller (the "metric describing a filtering decision" rule in `bug-prevention-patterns.md`), partitioning each cached path exactly so nothing ever has to be subtracted:

```
calls to bridged_summarize_contract_state
  = summarize_fast_hits + summarize_reload_hits + summarize_wasm_calls
calls to bridged_get_contract_state_delta
  = delta_fast_hits     + delta_reload_hits     + delta_wasm_calls

total WASM summarize_state = summarize_wasm_calls + summarize_wasm_uncached
total WASM get_state_delta = delta_wasm_calls     + delta_wasm_uncached
```

| arm | meaning |
|---|---|
| `*_fast_hits` | detector hash present and matching a cached summary/delta. No state load, no WASM. The arm that makes a 50/sec anti-entropy cadence cheap. |
| `*_reload_hits` | detector cold (restart, eviction) so the state was loaded and hashed, but the cache still matched. Real I/O, no WASM. |
| `*_wasm_calls` | WASM ran, on the cached path. The expensive work every storm fix was trying to bound. |
| `*_wasm_uncached` | WASM ran at a site with no cache in front of it (PUT/UPDATE response summaries in `contract_ops`, the per-subscriber client-notification delta fan-out). Separate so the cached partition stays exact while the WASM TOTAL stays honest. |

The delta path is included because it is the same seam with the same ambiguity, and it runs **per-subscriber** during broadcast fan-out — the hotter of the two.

`*_reload_hits` is the arm the existing tests could not see at all. The cache suite's oracle is `MockStateStorage::get_count` (no state reload ⇒ no WASM call), which scores a reload hit as a miss. Separating it says whether a storm is a cache-sizing problem or a state-churn problem.

### Why this surface, and not a new one

Two existing readouts, both free:

- **`router_snapshot`** (5-min cadence, `EventKind::RouterSnapshot`) already carries this class of per-node monotonic counter — `contract_module_cache_evictions_total`, `broadcast_stream_*_total`. Adding fields to it costs **no shadow-rollup budget slot** (`MAX_SHADOW_EVENTS_PER_SECOND` is 6, with one slot of headroom that the two mix rollups already crowd).
- **the `tracing::info!` line that event already emits**, so a local operator reads it from `journalctl` with no collector at all. `info!` survives `release_max_level_info`; a `debug!` would be compiled out of release builds.

Both lifetime totals (for collector-side differencing) and **per-window deltas** for the four headline arms are emitted, so a single snapshot answers the question without a stateful reader.

### How an operator reads it

One `router_snapshot` line per node every 5 minutes:

```
journalctl -u freenet -o cat | grep router_snapshot | tail -1
```

```
contract_exec_summarize_fast_hits_last_snapshot=14903
contract_exec_summarize_wasm_calls_last_snapshot=31
contract_exec_summarize_reload_hits_total=…
contract_exec_summarize_wasm_uncached_total=…
contract_exec_delta_fast_hits_last_snapshot=…
contract_exec_delta_wasm_calls_last_snapshot=…
```

Divide by 300 s. That example reads as ~49.7 summarize calls/sec of which **0.1/sec is real WASM** — the cache is doing its job and the storm rate is not CPU. The two rates converging is the signal that the change detector has stopped covering the load.

The same field names appear in the `router_snapshot` OTLP body for the collector.

## Cost

**One `Relaxed` `fetch_add` on a `u64` per contract-handler call** — a few ns, uncontended (the contract-handling loop has concurrency exactly 1, enforced by `&mut self`). The fast path takes exactly one increment; the slow paths take one more, alongside a state load and a WASM invocation that dwarf it. Nothing allocates and nothing locks.

Two details keep it there:

- the counters hang off `Ring` via `Arc` (not a process global) so unit tests stay isolated, per the #4488 rationale that moved `ModuleCacheMetrics` off a global — and the hot accessor returns a **borrow**, so a bump costs no `Arc` refcount RMW on top of the counter's own atomic;
- the client-notification fan-out clones the `Arc` **once per fan-out**, not per subscriber, because both fan-out arms hold two executor maps mutably for the whole loop.

The two cache-hit sites each moved one already-existing `.clone()` of the cached value ahead of the borrow's end (`LruCache::get` takes `&mut self`). Same clone, same count, different line.

## Behaviour

Pure instrumentation. Nothing a node *does* changes; no hosting, placement, or eviction path is touched (`hosting-invariants.md` reviewed — no invariant is in scope).

## Testing

Six behavioural tests driving a real `OpManager`/`Ring`-backed executor (`summarize_delta_cache_tests.rs`):

- repeated summarize of unchanged state counts a fast hit and **no** WASM call
- summarize after a write counts a WASM call and **not** a hit (without this, a mutation counting everything as a fast hit would still pass the first test)
- cold detector over unchanged state counts a **reload hit**, not a fast hit and not a WASM call
- delta sibling, plus: an unseen peer summary against unchanged state is a genuine miss (pins that the fast-hit count is not simply "the state didn't change")
- the three cached arms **partition every call exactly once**, with non-vacuity assertions that all three actually fired

Plus four pins:

- `summarize_wasm_call_records_both_sinks` — the production record must **not** be nested inside the `get_own_addr()` guard, which is the exact shape that made the original counter sim-only. A runtime test cannot cover this: the unit fixtures have no bound listener, so `get_own_addr()` returns `None` and the sim sink never fires there.
- `summarize_wasm_counter_sits_after_both_cache_hit_returns` — the counter must sit on the slow path, after both early returns, immediately before the call it describes.
- `contract_exec_export_maps_each_field_to_its_own_counter` — the `ring.rs` → `RouterSnapshotInfo` mirror seam, in the shape of `reconcile_shadow_export_maps_each_field_to_its_own_site`. A field swap there compiles cleanly and reports one arm's count under another's name.
- `router_snapshot_json_includes_contract_exec_counters` — the hand-mirrored OTLP `json!` block, asserting the **full** set with distinct per-field values so a partially-mirrored or cross-wired addition fails rather than shipping half-visible. That block has silently dropped fields before.

Both source-scrape pins strip comment lines before matching. The first draft of `summarize_wasm_call_records_both_sinks` failed because `get_own_addr()` appears in the block comment above the call as well as in the call, and the comment came first — a pin that matches its own explanation is not pinning anything.

### Mutation-tested (each applied to a committed tree, then reverted)

| mutation | caught by |
|---|---|
| nest the production record inside the `get_own_addr()` guard | `summarize_wasm_call_records_both_sinks` |
| record a reload hit as a fast hit (copy-paste slip) | `cold_detector_over_unchanged_state_counts_a_reload_hit` + the partition test's non-vacuity |
| drop one field from the OTLP `json!` mirror | `router_snapshot_json_includes_contract_exec_counters` |
| feed `_wasm_calls_total` from `fast_hits` in the export | `contract_exec_export_maps_each_field_to_its_own_counter` |

## Scope note

#5168 asks for entries-and-bytes per fallback summary reply. That is a different measurement (reply sizing on the `outbound_message_mix` rollup) and is left open; this PR is the sibling it names in spirit — the counter that says whether the work being bounded is expensive at all. `Refs`, not `Closes`.

Refs #5168, #5238, #4440.

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/contract/executor/pool_tests/summarize_delta_cache_tests.rs
+++ b/crates/core/src/contract/executor/pool_tests/summarize_delta_cache_tests.rs
@@ -810,6 +810,132 @@ async fn repeated_delta_counts_a_fast_hit_and_no_wasm_call() {
     );
 }
 
+/// Delta twin of `cold_detector_over_unchanged_state_counts_a_reload_hit`.
+/// The delta cache is keyed on (state-hash, peer-summary-hash), so a cold
+/// detector re-derives the state hash and finds the SAME key — a reload hit,
+/// not a WASM call.
+#[tokio::test(flavor = "current_thread")]
+async fn delta_cold_detector_over_unchanged_state_counts_a_reload_hit() {
+    let storage = MockStateStorage::new();
+    let (op_manager, _guards) = build_op_manager("exec_counts_delta_reload").await;
+    let mut exec = Executor::new_mock_wasm_uncached_with_op_manager(
+        "exec_counts_delta_reload",
+        storage.clone(),
+        op_manager.clone(),
+    )
+    .await
+    .expect("create uncached executor with op_manager");
+
+    let contract = test_contract(b"exec_counts_delta_reload");
+    let key = contract.key();
+    let peer_summary = StateSummary::from(vec![3u8; 4]);
+    exec.upsert_contract_state(
+        key,
+        Either::Left(WrappedState::new(vec![8, 8, 8, 1])),
+        RelatedContracts::default(),
+        Some(contract.clone()),
+    )
+    .await
+    .expect("PUT");
+    let _ = exec
+        .get_contract_state_delta(key, peer_summary.clone())
+        .await
+        .expect("warm the delta cache");
+
+    exec.state_store.cache_invalidator().invalidate(&key);
+
+    let before = exec_counts(&op_manager);
+    let _ = exec
+        .get_contract_state_delta(key, peer_summary)
+        .await
+        .expect("delta with cold detector");
+    let after = exec_counts(&op_manager);
+    assert_eq!(
+        after.delta_reload_hits - before.delta_reload_hits,
+        1,
+        "a cold detector over unchanged state must be counted as a delta reload hit"
+    );
+    assert_eq!(
+        after.delta_wasm_calls, before.delta_wasm_calls,
+        "the cached delta still matched, so no WASM ran and none may be counted"
+    );
+    assert_eq!(
+        after.delta_fast_hits, before.delta_fast_hits,
+        "the detector was cold, so this is not a FAST hit"
+    );
+}
+
+/// The client-notification fan-out computes a per-subscriber delta with NO
+/// cache in front of it. That work must land on the `uncached` arm — not on
+/// `delta_wasm_calls`, which would inflate the cached path's miss count and
+/// make a healthy cache look like it was thrashing.
+///
+/// This arm is the one most likely to dominate `get_state_delta` volume on a
+/// client-facing node, so it gets its own coverage rather than riding on the
+/// cached-path tests.
+#[tokio::test(flavor = "current_thread")]
+async fn client_notification_fanout_delta_counts_as_uncached() {
+    use crate::contract::executor::ClientId;
+
+    let storage = MockStateStorage::new();
+    let (op_manager, _guards) = build_op_manager("exec_counts_fanout").await;
+    let mut exec = Executor::new_mock_wasm_uncached_with_op_manager(
+        "exec_counts_fanout",
+        storage.clone(),
+        op_manager.clone(),
+    )
+    .await
+    .expect("create uncached executor with op_manager");
+
+    let contract = test_contract(b"exec_counts_fanout");
+    let key = contract.key();
+    exec.upsert_contract_state(
+        key,
+        Either::Left(WrappedState::new(vec![1, 1, 1, 1])),
+        RelatedContracts::default(),
+        Some(contract.clone()),
+    )
+    .await
+    .expect("PUT");
+
+    // A subscriber WITH a cached summary is what makes the fan-out take the
+    // delta arm rather than shipping full state.
+    let (tx, _rx) = tokio::sync::mpsc::channel(16);
+    exec.register_contract_notifier(
+        *key.id(),
+        ClientId::FIRST,
+        tx,
+        Some(StateSummary::from(vec![2u8; 4])),
+    )
+    .expect("register subscriber");
+
+    let before = exec_counts(&op_manager);
+    exec.upsert_contract_state(
+        key,
+        Either::Left(WrappedState::new(vec![5, 5, 5, 5, 5])),
+        RelatedContracts::default(),
+        None,
+    )
+    .await
+    .expect("UPDATE drives the fan-out");
+    let after = exec_counts(&op_manager);
+
+    assert_eq!(
+        after.delta_wasm_uncached - before.delta_wasm_uncached,
+        1,
+        "the per-subscriber fan-out delta must be counted on the UNCACHED arm"
+    );
+    assert_eq!(
+        after.delta_wasm_calls, before.delta_wasm_calls,
+        "the fan-out delta has no cache in front of it, so it must NOT inflate \
+         the cached path's miss count"
+    );
+    assert_eq!(
+        after.delta_fast_hits, before.delta_fast_hits,
+        "the fan-out delta is not a cache hit"
+    );
+}
+
 /// The cached path's three arms must PARTITION its calls exactly: every call to
 /// `summarize_contract_state` lands on exactly one of fast hit / reload hit /
 /// WASM call, so no consumer ever has to subtract one counter from another to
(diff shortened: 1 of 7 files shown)


sanity:
## Rebased onto current main, and re-reviewed on the rebased head

This had gone stale (72 commits behind, CONFLICTING). Rebased to `ed728df4`, now MERGEABLE.

### What conflicted, and how it was resolved

Exactly one file: `crates/core/src/contract/executor/runtime/executor_impl.rs`. Both sides append a new `#[cfg(test)] mod` at end-of-file — main added `conformance_capture_pins` (RFC #5320), this branch adds `contract_exec_counter_pins` — and git folded their shared trailing `);\n }\n}` together, truncating both sides. Resolved by keeping both blocks, each given its own copy of that tail.

Because a conflict resolution is itself an unreviewed judgement call, the result was checked for silent loss rather than just for absent markers:

- diffstat against main after the rebase is identical to the pre-rebase diffstat against the old merge base — 7 files, 1560 insertions, 14 deletions, same per-file numbers;
- the added/removed lines of `git diff origin/main...HEAD` for that file are byte-identical to the pre-rebase delta;
- line counts reconcile: 3037 (merge base) + 214 (this branch's net) + 405 (main's net) = 3656;
- the load-bearing call site was checked directly rather than inferred: at `executor_impl.rs:1382-1385` the production record still sits **outside** the `get_own_addr()` guard with the simulation record inside it, which is the ordering this PR exists to protect.

### Review

Full tier (bumped from Light because the rebase required a judgement call inside the contract executor). Two independent blind lenses: skeptical and code-first. Code-first found no divergence between intent and implementation. Four findings were confirmed and are fixed in `ed728df4`:

1. **Both fan-out arms record an uncached delta, but only one is reachable from any single test fixture**, and nothing pinned the count — so deleting either recorder left CI green. That is precisely the "counter is a no-op on the path that matters, with a passing test standing over it" failure this module exists to remove. Added a source-level count pin that also checks each recorder binds to its own `get_state_delta` call.
2. **`summarize_body()`'s end-delimiter `.expect()` could never fire.** The signature literal it searches for also occurs in the helper's own source, which `include_str!` pulls in, so a rename of the delta method would silently widen the scrape region to the whole production file instead of panicking. Live inside the module whose own doc claims to have closed that shape. Now truncates at the first test module, matching the correct sibling in `contract_ops.rs`.
3. **The partition identity had become false during the rebase.** `conformance::runtime_oracle` calls `Runtime::summarize_state` / `get_state_delta` directly rather than through the executor, records nothing, and runs on a real node via the conformance shadow probe — the "mirrored counter rots when a new terminal path is added" pattern. Narrowed the claim to executor-mediated invocations and documented the oracle as a fifth exception. An overstated completeness claim is worse than a missing one on a counter whose whole point is that the previous number was undifferentiated.
4. **A ninth counter would have recorded but never exported**, since the export was sixteen hand-written assignments guarded by a scrape hardcoding today's eight names. `ContractExecSnapshot` is now destructured exhaustively with no `..`, so a ninth arm fails to compile until exported.

### One reviewer finding was refuted rather than fixed

The skeptical lens raised a **blocking** claim that the 35-field `router_snapshot` event exceeds `tracing`'s 32-field cap and would not compile, citing `tracing` 0.1.44 and tokio-rs/tracing#782.

The cap lives in `tracing-core`, not `tracing`, and the locked `tracing-core` 0.1.36 implements `ValidLen` for **all** N via const generics:

```rust
impl<'a, const N: usize> ValidLen<'a> for [(&'a Field, Option<&'a (dyn Value + 'a)>); N] {}
```

The "up to 32 key-value fields" line in `tracing`'s macro docs is stale relative to that impl. Settled empirically rather than by argument: a throwaway crate pinned to the locked `tracing` 0.1.44 / `tracing-core` 0.1.36 with 34 named fields plus a message compiles clean, and `cargo check --locked --all-targets` on this tree is green.

Worth recording because the reasoning was sound and the conclusion still wrong — the version that carries the constraint was not the version that was checked.

[AI-assisted - Claude]

Issue #5039, opened 2026-07-29
Flaky: three lib tests fail only under full parallel `cargo test --lib` load (rtt registry cross-talk + two others)


sanity:
Found while verifying #5029. These are **pre-existing** — they reproduce identically on the unmodified base commit `14d20095`, and all pass in isolation.

## Observed

Four consecutive full runs of `cargo test -p freenet --lib --all-features` on nova (load average ~33, many parallel agents) produced a rotating failure set:

| run | failures |
|---|---|
| 1 | `test_set_own_addr_atomic_with_network_status`, `homepage_renders_dynamic_title`, `cross_connection_median_returns_some_when_a_peer_has_inflation` |
| 2 | `test_set_own_addr_atomic_with_network_status`, `cross_connection_median_...`, `validate_host_function_delegate` |
| 3 | `send_update_notification_empty_local_snapshot_emits_warn`, `test_set_own_addr_atomic_with_network_status`, `cross_connection_median_...`, `validate_host_function_delegate` |
| 4 | `test_set_own_addr_atomic_with_network_status`, `cross_connection_median_...` |

Every one of them passes when run alone.

Already tracked, listed only so the set is complete: `test_set_own_addr_atomic_with_network_status` (#4819) and `send_update_notification_empty_local_snapshot_emits_warn` (#4927). This issue covers the remaining three.

## 1. `transport::rolling_rtt_stats::tests::cross_connection_median_returns_some_when_a_peer_has_inflation` — root cause identified

```
crates/core/src/transport/rolling_rtt_stats.rs:776:
assertion failed: median > Duration::ZERO
```

`cross_connection_median_inflation()` reads a **process-global registry**. The test registers one peer with a known inflation and then asserts the median *across every peer any concurrently-running test has registered* is non-zero. The test's own comment acknowledges the cross-talk ("other tests may contribute peers") but the assertion is not actually robust to it: once enough concurrently-registered peers have zero inflation, the median lands on a zero entry and the assertion fails.

The neighbouring test `cross_connection_median_is_robust_to_a_single_outlier` already does the right thing — it filters to its own peers and open-codes the median. The fix is to give this test the same treatment (or scope the registry per test) rather than to loosen the assertion, since the assertion is the only coverage of the public function.

## 3. `wasm_runtime::delegate::test::validate_host_function_delegate`

[...]

PR #5251, opened 2026-08-08
fix(wasm_runtime): issue WASM instance ids from one process-global allocator

### LINKED ISSUE #4213: test: test_large_secret_data flaky again under heavy parallel lib-test load

### LINKED ISSUE #5023: test: delegate secret tests flake ~1/8 under full wasm_runtime suite (#3248 class)
## Symptom

Two `wasm_runtime::delegate::test` tests fail intermittently when the whole
`wasm_runtime::` lib-test set runs in one process:

- `wasm_runtime::delegate::test::test_large_secret_data`
- `wasm_runtime::delegate::test::test_store_and_retrieve_secret`

Captured panic (the only one I caught in full):

```
---- wasm_runtime::delegate::test::test_store_and_retrieve_secret stdout ----
thread 'wasm_runtime::delegate::test::test_store_and_retrieve_secret' panicked at
crates/core/src/wasm_runtime/delegate/test.rs:1335:13:
Expected SecretResult(Some(...)), got SecretResult(None)
```

Both tests store a secret through the delegate WASM boundary and then read it
back; the failure is that the read-back misses. Only one of the two fails in
any given run, never both.

## Reproduction

```bash
for i in $(seq 1 12); do
  cargo test -p freenet --lib --all-features wasm_runtime:: 2>&1 | grep -E "^test result:|^    wasm_runtime"
done
```

Observed rate ~1 in 8 runs on a heavily loaded machine (load average 20-30,
several concurrent cargo builds). It does **not** reproduce with the narrower
`wasm_runtime::delegate` filter — 24 consecutive clean runs — so it appears to
need the parallelism / memory pressure of the full `wasm_runtime::` set (~485
tests in one process).

## This is pre-existing, not a regression

Verified by running the identical command on `43eb1ea9` (`fix(storage):
reclaim redb free pages at startup (#5005)`, main at the time of testing) with
no other changes in the tree:

| Tree | Runs | Failures |
|---|---|---|
| `43eb1ea9` (main) | 12 | 1 (`test_large_secret_data`, run 6) |
| PR #5017 branch | 12 | 1 (`test_store_and_retrieve_secret`, run 2) |

Same tests, same rate, so #5017 is not implicated. Noting it here rather than
fixing it in that PR, which is a different subsystem.

## Hypothesis (unverified)

Both tests are in the family `test_large_secret_data`'s own doc comment
describes:

> Indirect regression test for #3248 (stale WASM memory base pointer). ...
> Under full parallel test suite runs (~1600 tests) the relocation is more
> likely due to memory pressure.

So the natural first suspicion is that the `refresh_mem_addr_from_caller`
mitigation for #3248 does not cover every host-function path reachable from
these delegates, or that the `MEM_ADDR` map has a race under
`#[tokio::test(flavor = "multi_thread")]`. That is a hypothesis from reading
the test's own comment, **not** something I traced — a real diagnosis needs
someone to instrument `MEM_ADDR` and confirm whether the pointer is stale at
the failing read, or whether the miss comes from somewhere else entirely
(e.g. the secret path/key differing, which would point at a different cause).

Worth checking first, because it is cheap and would change the diagnosis:
whether the read is missing because the *key* is garbage (stale pointer while
reading the key out of WASM memory) or because the value genuinely was not
persisted.

## Why it matters

Per the project's flaky-test policy a flaky test is a broken test, and this
one is guarding a real past production bug (#3248) — if the mitigation is
incomplete, the flake is the symptom of a live defect in the host-function
memory handling rather than test noise.

[AI-assisted - Claude]

sanity:
## Problem

`wasm_runtime::delegate::test::test_large_secret_data` and its siblings fail intermittently under the full parallel lib suite and never in isolation. Tracked as #4213 and #5023, with two distinct symptom shapes:

- `test_large_secret_data` / `test_store_and_retrieve_secret`: `Expected SecretResult(Some(...)), got SecretResult(None)`
- `test_v2_delegate_update_existing_state`: `error_code: -1` / `ContractNotFound`

Both have the same cause, and it is provable by reading the code, independent of any observed flake rate.

`MEM_ADDR`, `DELEGATE_ENV` and `CONTRACT_IO` are process-**global** DashMaps keyed by WASM instance id, so the id namespace is process-global too. But `WasmEngine::create_instance` took a caller-supplied `id: i64`, and only `RunningInstance::new` drew one from the shared `INSTANCE_ID` counter. The engine unit tests in `wasmtime_engine.rs` picked their own ids:

- `0..10_001` in `test_instance_limit_override_allows_many_instances`
- `0..STORE_REFRESH_THRESHOLD` (500) in the three store-refresh tests
- literals such as `999`, `0`, `1`

Those tests' `drop_instance` calls do `MEM_ADDR.remove(&id)`. Running in the same test binary as the delegate tests, they removed the entry belonging to a **live** delegate or contract instance in a concurrently-running test that had been issued the same id from the counter. Every host function on the victim instance then took its "no MEM_ADDR entry" branch and returned `ERR_NOT_IN_PROCESS`, which the stdlib's `DelegateCtx` collapses into "not found" (`result < 0` → `None`).

That is why the flake needs the `wasm_runtime::` or full-lib filter and has never reproduced under `wasm_runtime::delegate` alone (#5023 reports 24 consecutive clean runs there): the offending engine tests are filtered out.

**Scope:** the collision needs two tests to share a process, so it bites `cargo test`, which is what AGENTS.md tells contributors to run and what both issues reported. CI runs `cargo nextest` (`ci.yml:893-903`), which gives each test its own process. **A green CI tick is therefore not coverage for this defect class** — nextest cannot observe it at any repeat count, before or after the fix.

### Evidence

Captured with temporary instrumentation on the host-function early-return paths, which are otherwise `tracing::warn!`-only and invisible in a test run:

```
---- wasm_runtime::delegate::test::test_large_secret_data stdout ----
set_secret id=77 delegate=9XS1At... val_len=1048576 ok=true
refresh_mem_addr NO-MEM_ADDR-ENTRY id=89
get_secret EARLY no MEM_ADDR id=89
panicked: Expected SecretResult(Some(...)), got SecretResult(None)

---- wasm_runtime::delegate::test::test_v2_delegate_update_existing_state stdout ----
refresh_mem_addr NO-MEM_ADDR-ENTRY id=91
panicked: Expected ContractState, got ContractNotFound { ..., error_code: -1 }
```

`get_secret_len` for id 89 had already succeeded (no miss logged), so the entry existed and then vanished between two host calls on the same live instance. Both observed ids fall inside the `0..500` and `0..10_001` ranges the engine tests iterate.

### On the failure rate — read this before citing a number

The original author measured 3 failures in 9 `cargo test -p freenet --lib wasm_runtime::` runs on one machine, consistent with the ~1/8 in #5023.

**That rate has NOT been independently reproduced.** A later attempt saw 0 failures in 20 runs, but correctly established that its own measurement could not have detected the defect: it ran 538 of 5267 lib tests in ~20s, which strips most of the parallel load #4213 names as the trigger. So neither the 3/9 nor the 0/20 should be treated as a measurement of this bug's rate, and no before/after rate comparison in this PR should be read as validation.

**This change stands on the mechanism, not on a rate.** Two independent issuers writing into one process-global keyspace is visible by reading `create_instance` against `MEM_ADDR`; the fix removes the second issuer.

## Approach

Renumbering the offending tests would fix today's collision and leave the hazard for the next test that picks an id. Instead the collision is made unrepresentable:

- `native_api.rs` owns the single `NEXT_INSTANCE_ID` allocator, sited next to the global maps whose keyspace it defines, exposed as `next_instance_id()`.
- `WasmEngine::create_instance` no longer takes an `id`. It allocates one and returns it in the `InstanceHandle`, so no caller of `create_instance` can supply one.
- `RunningInstance::new` reads `handle.id`; the local `INSTANCE_ID` static is gone.
- Engine test call sites updated (mechanical: drop the id argument), 25 in total.

Production behaviour is unchanged: the same counter semantics, the same monotonic ids, the same single production caller (`runtime.rs:304`, still wrapped in `classify_result` per the #4864 invariant, and still counted 1/1 by the `guest_entry_calls_route_through_classify_result` pin in `contract/executor.rs`).

**What this does and does not close**, stated precisely because the docs previously overstated it:

- It closes the `create_instance` parameter, which is where every id that actually reached these maps came from.
- It is **not** a type-level guarantee. `InstanceHandle.id` is `pub(super)`, so code inside `wasm_runtime` can still construct a handle with a chosen id (`delegate/test.rs` does, twice, inertly — `process_outbound` binds it as `_handle` and never touches the maps). Both the allocator rustdoc and the regression test's rustdoc now say so.
- The WASM ABI is a separate id surface this does not validate: four host functions take a guest-supplied instance id. Called out in the `NEXT_INSTANCE_ID` docs so the invariant is not over-read.

## Testing

Two new tests:

- `instance_ids_are_globally_unique_across_engines`: engine A holds a live instance with a recorded `MEM_ADDR` entry while engine B churns 64 instances the way the store-refresh and instance-limit tests do. Asserts B is never issued A's id and that A's entry survives. Mutation-tested: with `create_instance` reverted to a per-engine id (`self.lifetime_instances as i64`) it fails with `left: 0, right: 0`.
- `create_instance_allocates_its_own_instance_id`: bounded-region source pin. The signature stops a *caller* passing an id, but nothing stops `create_instance` itself from reverting to a per-engine counter. Mutation-tested against exactly that change. It cuts the file at the `#[cfg(test)]` module before scraping, so it cannot match its own assertion message and pass while guarding nothing.

One pre-existing pin hardened in passing: `create_instance_recovers_store_on_guest_entry_failure` scraped the whole file with no test-module cutoff, so a rename of `create_instance` would have let its `find` fall through into the test module instead of panicking. It fails closed today only by the accident of which string literal happens to sit earliest. Same cutoff applied.

Because `cargo nextest` cannot see this defect class, the meaningful signal is plain `cargo test` in one process: `cargo test -p freenet --lib wasm_runtime` is green on this head, as is `cargo check --all-targets` and `cargo fmt`.

Closes #4213
Closes #5023

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/wasm_runtime/engine.rs
+++ b/crates/core/src/wasm_runtime/engine.rs
@@ -165,6 +165,10 @@ pub(crate) trait WasmEngine: Send {
     /// `CONTRACT_IO` maps, so a caller-chosen id could collide with a LIVE
     /// instance belonging to another engine in the same process and make
     /// [`Self::drop_instance`] evict that instance's entry (#4213 / #5023).
+    ///
+    /// Any future backend implementing this trait MUST allocate the same way;
+    /// `create_instance_allocates_its_own_instance_id` in the wasmtime backend
+    /// pins that for the one implementation that exists today.
     fn create_instance(
         &mut self,
         module: &Self::Module,
--- a/crates/core/src/wasm_runtime/engine.rs
+++ b/crates/core/src/wasm_runtime/engine.rs
@@ -155,6 +155,10 @@ pub(crate) trait WasmEngine: Send {
     /// `CONTRACT_IO` maps, so a caller-chosen id could collide with a LIVE
     /// instance belonging to another engine in the same process and make
     /// [`Self::drop_instance`] evict that instance's entry (#4213 / #5023).
+    ///
+    /// Any future backend implementing this trait MUST allocate the same way;
+    /// `create_instance_allocates_its_own_instance_id` in the wasmtime backend
+    /// pins that for the one implementation that exists today.
     fn create_instance(
         &mut self,
         module: &Self::Module,
--- a/crates/core/src/wasm_runtime/engine.rs
+++ b/crates/core/src/wasm_runtime/engine.rs
@@ -148,10 +148,16 @@ pub(crate) trait WasmEngine: Send {
     ///
     /// Sets up imports, calls `__frnt_set_id`, records memory address,
     /// and ensures sufficient memory for `req_bytes`.
+    ///
+    /// The instance id is NOT a parameter: it is issued by
+    /// `native_api::next_instance_id` and returned in the [`InstanceHandle`].
+    /// Instance ids key the process-global `MEM_ADDR` / `DELEGATE_ENV` /
+    /// `CONTRACT_IO` maps, so a caller-chosen id could collide with a LIVE
+    /// instance belonging to another engine in the same process and make
+    /// [`Self::drop_instance`] evict that instance's entry (#4213 / #5023).
     fn create_instance(
         &mut self,
         module: &Self::Module,
-        id: i64,
         req_bytes: usize,
     ) -> Result<InstanceHandle, WasmError>;
 
--- a/crates/core/src/wasm_runtime/engine.rs
+++ b/crates/core/src/wasm_runtime/engine.rs
@@ -158,10 +158,16 @@ pub(crate) trait WasmEngine: Send {
     ///
     /// Sets up imports, calls `__frnt_set_id`, records memory address,
     /// and ensures sufficient memory for `req_bytes`.
+    ///
+    /// The instance id is NOT a parameter: it is issued by
+    /// `native_api::next_instance_id` and returned in the [`InstanceHandle`].
+    /// Instance ids key the process-global `MEM_ADDR` / `DELEGATE_ENV` /
+    /// `CONTRACT_IO` maps, so a caller-chosen id could collide with a LIVE
+    /// instance belonging to another engine in the same process and make
+    /// [`Self::drop_instance`] evict that instance's entry (#4213 / #5023).
     fn create_instance(
         &mut self,
         module: &Self::Module,
-        id: i64,
         req_bytes: usize,
     ) -> Result<InstanceHandle, WasmError>;
 
--- a/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
+++ b/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
@@ -2910,8 +2910,18 @@ mod tests {
             .find("    fn create_instance(")
             .expect("create_instance not found");
         let body = &src[start..];
+        // `+ 1` because this `find` searches `body[1..]`, so its index is one
+        // short of the position in `body`. Without it the slice drops the byte
+        // before the next method -- harmless today (it is the ASCII newline
+        // closing this method), but `&body[..end]` would then slice at an
+        // arbitrary byte, and this file is full of multi-byte em dashes: one
+        // landing there turns the pin into a "byte index is not a char
+        // boundary" panic that says nothing about what it guards. Matches
+        // `create_instance_recovers_store_on_guest_entry_failure` below, which
+        // this test's rustdoc claims to follow.
         let end = body[1..]
             .find("\n    fn ")
+            .map(|i| i + 1)
             .expect("create_instance body must end at the next method");
         let body = &body[..end];
         assert!(
--- a/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
+++ b/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
@@ -2893,7 +2893,19 @@ mod tests {
     /// `create_instance_recovers_store_on_guest_entry_failure` below.
     #[test]
     fn create_instance_allocates_its_own_instance_id() {
-        let src = include_str!("wasmtime_engine.rs");
+        // Cut the test module off BEFORE scraping. `include_str!` pulls in the
+        // whole file, this module included, and the needle below occurs here as
+        // a string literal. Without the cut, renaming `create_instance` would
+        // not panic: `find` would fall through to this function's own source,
+        // and the region would then be this test's tail -- whose assertion
+        // message contains `native_api::next_instance_id()`, so the pin would
+        // scrape its own error string and PASS while guarding nothing.
+        // Same remedy as `contract_ops.rs::production_source`; see #5450.
+        let full = include_str!("wasmtime_engine.rs");
+        let cutoff = full
+            .find("\n#[cfg(test)]\nmod tests {")
+            .expect("wasmtime_engine.rs must have a top-level #[cfg(test)] mod tests");
+        let src = &full[..cutoff];
         let start = src
             .find("    fn create_instance(")
             .expect("create_instance not found");
@@ -3843,7 +3855,7 @@ mod tests {
 
         // Charge one instance to learn what the arena actually retains per
         // instance, then set a budget only a few instances wide.
-        let handle = engine.create_instance(&module, 0, 1024).unwrap();
+        let handle = engine.create_instance(&module, 1024).unwrap();
         engine.drop_instance(&handle);
         let per_instance = engine.retired_instance_bytes;
         assert!(
@@ -3859,8 +3871,8 @@ mod tests {
         // here is attributable to the byte bound alone.
         let creations = 40;
         assert!(creations < STORE_REFRESH_THRESHOLD);
-        for i in 1..=creations {
-            let handle = engine.create_instance(&module, i as i64, 1024).unwrap();
+        for _ in 1..=creations {
+            let handle = engine.create_instance(&module, 1024).unwrap();
             engine.drop_instance(&handle);
             if engine.lifetime_instances <= previous {
                 refreshes += 1;
@@ -3886,7 +3898,7 @@ mod tests {
             "engine must stay healthy across byte-budget refreshes"
         );
         let handle = engine
-            .create_instance(&module, 999_999, 1024)
+            .create_instance(&module, 1024)
             .expect("should create instance after byte-budget refresh");
         engine.drop_instance(&handle);
     }
@@ -3919,7 +3931,7 @@ mod tests {
 
         for i in 0..12 {
             let handle = engine
-                .create_instance(&module, i as i64, 1024)
+                .create_instance(&module, 1024)
                 .unwrap_or_else(|e| panic!("instance {i} must still be creatable: {e}"));
             assert_eq!(
                 engine.lifetime_instances, 1,
--- a/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
+++ b/crates/core/src/wasm_runtime/engine/wasmtime_engine.rs
@@ -790,9 +790,11 @@ impl WasmEngine for WasmtimeEngine {
     fn create_instance(
         &mut self,
         module: &Module,
-        id: i64,
         req_bytes: usize,
     ) -> Result<InstanceHandle, WasmError> {
+        // Ids come from the one process-global allocator, never from the
+        // caller. See `native_api::NEXT_INSTANCE_ID` for why (#4213 / #5023).
+        let id = native_api::next_instance_id();
         {
             let store = self
                 .store
@@ -2273,7 +2275,7 @@ mod tests {
         let module = engine
             .compile(SIMPLE_WASM)
             .expect("offloaded compile should succeed");
-        let handle = engine.create_instance(&module, 0, 1024).unwrap();
+        let handle = engine.create_instance(&module, 1024).unwrap();
         engine.drop_instance(&handle);
         assert!(engine.module_compiled_size(&module) > 0);
     }
@@ -2292,7 +2294,7 @@ mod tests {
         let module = engine
             .compile(SIMPLE_WASM)
             .expect("inline compile should succeed");
-        let handle = engine.create_instance(&module, 0, 1024).unwrap();
+        let handle = engine.create_instance(&module, 1024).unwrap();
         engine.drop_instance(&handle);
         assert!(engine.module_compiled_size(&module) > 0);
     }
@@ -2509,7 +2511,7 @@ mod tests {
         let start = std::time::Instant::now();
         // create_instance runs the start function via instantiate_async; the
         // global epoch ticker preempts it once the armed 3-tick deadline elapses.
-        let result = engine.create_instance(&module, 1, 0);
+        let result = engine.create_instance(&module, 0);
         let elapsed = start.elapsed();
 
         // `InstanceHandle` is not Debug, so match rather than `{result:?}`.
@@ -2754,7 +2756,7 @@ mod tests {
             .compile(SIMPLE_WASM)
             .expect("compile under current_thread+offload must succeed (no panic)");
         let handle = engine
-            .create_instance(&module, 0, 1024)
+            .create_instance(&module, 1024)
             .expect("module must be instantiable");
         engine.drop_instance(&handle);
         assert!(engine.module_compiled_size(&module) > 0);
@@ -2804,7 +2806,7 @@ mod tests {
             .compile(SIMPLE_WASM)
             .expect("compile with no runtime + offload must succeed inline");
         let handle = engine
-            .create_instance(&module, 0, 1024)
+            .create_instance(&module, 1024)
             .expect("module must be instantiable");
         engine.drop_instance(&handle);
         assert!(engine.module_compiled_size(&module) > 0);
@@ -2974,7 +2976,7 @@ mod tests {
         // default 10,000 limit. Without our ResourceLimiter override this would fail.
         for i in 0..10_001 {
             let handle = engine
-                .create_instance(&module, i, 1024)
+                .create_instance(&module, 1024)
                 .unwrap_or_else(|e| panic!("instance {i} should succeed: {e}"));
             engine.drop_instance(&handle);
         }
@@ -3325,7 +3327,7 @@ mod tests {
         );
 
         let handle = engine
-            .create_instance(&module, 999, 1024)
+            .create_instance(&module, 1024)
             .expect("create instance");
 
         let result = engine.call_3i64_async_imports(&handle, "process", 0, 0, 0);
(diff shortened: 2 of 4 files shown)


sanity:
## Review (Full tier: WASM runtime is a high-risk surface)

Two independent blind Claude reviewers read the checked-out code: a **skeptical** lens (adversarial bug hunt) and a **code-first** lens (read the code before the description, flag intent/implementation mismatches). No external model pass, per the current review policy.

The code-first reviewer found no discrepancies and independently confirmed the pin test in `contract/executor.rs:2357` still holds after the signature change (`create_calls == create_wrapped == 1`). The skeptical reviewer raised six findings; all are addressed below.

### Fixed in c986d9a3d

**1. (High) The doc comment recorded the cause as more general than it is.** The collision needs the two tests to share a process. CI runs `cargo nextest`, which gives each test its own process (`.github/workflows/ci.yml:546-556`), so **CI was never affected**. The flake bites `cargo test`, which is what AGENTS.md tells contributors to run and what both #4213 and #5023 actually reported (`cargo test -p freenet --lib ...` in each). Left unqualified, the rustdoc would have sent the next investigator down the wrong path if a delegate test ever flakes in CI. The doc now names the runner explicitly.

**2. (Medium) "no caller can supply an id of its own" was too strong.** Four host functions take an instance id as a guest-supplied WASM parameter and look it up in the same global maps: `__frnt__logger__info`, `__frnt__rand__rand_bytes`, `__frnt__time__utc_now`, `__frnt__fill_buffer`. That is a separate, pre-existing id surface this PR does not validate. The claim is now scoped to `create_instance` callers, with an explicit note that the WASM ABI is a distinct surface. The surface itself is out of scope here and has been reported through a non-public channel.

**3. (Medium) The regression test's rustdoc oversold it.** A caller-supplied id is now a compile error, not something a runtime test can catch. The rustdoc now claims only what the test pins: the allocator is process-global, not per-engine.

**5. (Low) Enforcement moved from one site to every backend, with no pin.** Added `create_instance_allocates_its_own_instance_id`, a bounded-region source scrape following the `create_instance_recovers_store_on_guest_entry_failure` convention already in this file. Mutation-tested: with `create_instance` reverted to `self.lifetime_instances as i64` it fails with the intended message; with the fix it passes.

### Not changed, with reasons

**4. (Low) Gate `drop_instance`'s `MEM_ADDR.remove` on `self.instances.remove(...).is_some()`.** Declined. With ids now process-unique, `drop_instance` provably cannot reach another engine's entry, so the gate adds no safety. It does change behaviour on the `replace_store()` path, where `self.instances` is cleared while instances are still live: gating would skip the removal for those orphans and lean entirely on `RunningInstance::drop`. That is a real behavioural change on a high-risk surface in exchange for no demonstrated bug, which is the wrong trade in this PR.

**6. (Low) `replace_store()` clears `self.instances` without removing those `MEM_ADDR` entries.** Pre-existing, unreachable as a use-after-free (every host function refreshes the base pointer from its own `Caller` first), and out of scope for a flake fix. Noted so it is not lost.

### Also verified by the reviewers

- All 20 call-site edits are mechanically correct, including the two that could have transposed arguments: `(&module, 1, 0)` → `(&module, 0)` and `(&module, STORE_REFRESH_THRESHOLD as i64, 1024)` → `(&module, 1024)`.
- No remaining id source outside the allocator. `create_instance` is called only from `runtime.rs:73` and this file's test module; nothing in `crates/core/tests/` or `simulation/`.
- No test depended on the id it passed. `SIMPLE_WASM` makes no host calls, so `999`, `999_999` and `-1` were inert; the store-refresh tests assert on `lifetime_instances`, not ids.
- The `-1` sentinel used by `CURRENT_DELEGATE_INSTANCE` is still unreachable: the allocator starts at 0 and only increments, and the one call site that used `-1` is gone.
- Ordering is preserved: `next_instance_id()` is the first statement of `create_instance`, before `__frnt_set_id` tells the guest its id. A failed create burns an id exactly as the old `INSTANCE_ID.fetch_add` did.

### Verification

14 consecutive full `cargo test -p freenet --lib` runs on `f222c81b9`: **zero** `wasm_runtime` failures. Two runs hit unrelated pre-existing flakes, both already tracked or being tracked: `transport::rolling_rtt_stats::tests::cross_connection_median_returns_some_when_a_peer_has_inflation` (#5039, root cause already identified there) and `util::rate_limit_layer::tests::per_callsite_concurrent_writers_and_summary_no_deadlock`.

For contrast, before the fix the narrower `cargo test -p freenet --lib wasm_runtime::` loop reproduced the delegate flake 3 times in 9 runs on the same machine.

A re-verification campaign is running on `c986d9a3d`; the tally will be posted here.

[AI-assisted - Claude]

sanity:
## Re-verification on `c986d9a3d`

12 consecutive full `cargo test -p freenet --lib` runs on the final commit, `--test-threads` at the default (16 cores, machine load average 20-30 throughout from other concurrent work):

| runs | `wasm_runtime` failures | other failures |
|---|---|---|
| 12 | **0** | 1 (`topology::small_world_rand::tests::chi_squared_test`, a statistical test unrelated to this change) |

Combined with the 14 runs on `f222c81b9` reported above: **26 full-suite runs, zero occurrences of the flake**. For contrast, before the fix the narrower `cargo test -p freenet --lib wasm_runtime::` loop reproduced it 3 times in 9 runs on the same machine.

The other failures seen across the 26 runs were all pre-existing and unrelated: `cross_connection_median_returns_some_when_a_peer_has_inflation` (#5039, root cause already identified there), `per_callsite_concurrent_writers_and_summary_no_deadlock` (newly diagnosed in a comment on #5039), and the chi-squared test above.

[AI-assisted - Claude]

sanity:
## Rebased onto current main, reviewed, and pushed at `fbb2ad9a`

This had gone stale and CONFLICTING. Rebased over ~70 commits of drift; three things surfaced that the original branch could not have known about.

### CI's green tick does not examine this defect. Please don't read it as coverage.

This is the point worth carrying forward from the PR, and it survives whatever anyone concludes about the flake rate.

The bug needs **two tests alive in one process**: an engine test picks an id, `drop_instance` removes that id's `MEM_ADDR` entry, and the entry it removes belongs to a *live* delegate instance that the allocator issued the same id to. CI runs `cargo nextest`, which gives every test its own process. **There is no repeat count at which nextest can observe this**, because the second participant is never in the room. It is structural blindness, not luck.

`AGENTS.md` tells contributors to run plain `cargo test`, which is the only runner that can see it, and that is what both #4213 and #5023 were reported against. `.claude/rules/testing.md` documents the general class.

### The measurement: I could not independently reproduce it, and I am not dressing that up

The author measured **3 failures in 9 runs**. I did not reproduce it:

| setup | runs | failures |
|---|---|---|
| `wasm_runtime::` filter (538 of 5267 lib tests) | 20 | 0 |
| full lib suite, no filter (5241 tests) | 2 | 0 |

What that does and does not establish:

- **It is not evidence the fix is unnecessary.** A race that reproduces 1-in-3 for one person and 0-in-22 for another is a race whose window depends on scheduling, and this machine was running several concurrent build lanes throughout — an uncontrolled confound I can't characterise in either direction.
- **My filtered setup was probably too quiet.** #4213's title is "flaky again under heavy parallel *lib-test* load", and the 20-run arm exercised 538 of 5267 tests in ~20s each. That removes most of the concurrent load the issue names. It is a milder cousin of the mistake that produces #5023's "24 consecutive clean runs" under `wasm_runtime::delegate` — that filter removes the colliding engine tests outright; mine kept the competitor but dropped the pressure.
- The full-lib arm is only 2 runs, which is too few to say anything. It was cut short deliberately: the host was at 96% disk with under an hour of headroom, and accumulating more null data was not worth the risk to concurrent work.

I verified the measurement could have failed rather than assuming it: both sides of the race genuinely ran (118 `wasm_runtime::delegate` and 46 `wasmtime_engine` test lines per run), and the binary under test was confirmed unfixed **at the artifact level** — 0 `next_instance_id` symbols in the compiled binary, with an mtime predating every source change. That is stronger than grepping the tree, which is only a proxy for what was actually compiled.

**The case for merging rests on the mechanism, which is provable by reading and does not depend on a rate.** `MEM_ADDR` / `DELEGATE_ENV` / `CONTRACT_IO` are process-global maps keyed by instance id, so the id keyspace is process-global — while `create_instance` accepted a caller-supplied id, giving two independent issuers into one keyspace. That is a correctness bug in the allocator's contract. How often it bites and whether it is wrong are different questions, and only the second one gates the fix.

### What the rebase surfaced

**Four stale call sites.** `wasmtime_engine.rs:3835, 3852, 3878, 3911` still passed an id — tests main added *after* this branch was written, so the branch's mechanical argument-drop never reached them. Worth noting: two pick `i as i64` and one picks `999_999`, i.e. main kept adding fresh instances of the exact hazard this PR removes, which suggests the parameter would have gone on attracting collisions for as long as it existed.

**The new pin could not fail.** `create_instance_allocates_its_own_instance_id` scraped `include_str!("wasmtime_engine.rs")` for `"    fn create_instance("` without cutting off the test module, and that needle occurs in the file as the pin's own string literal. While the real method exists it sorts first and the pin works. On the rename the pin exists to catch, `find` falls through to the pin's own source rather than returning `None`; the region becomes this test's tail — and that tail's assertion message contains the literal `native_api::next_instance_id()`. The assertion would find its needle inside its own error string and pass while guarding nothing. Now truncates before the test module, matching `contract_ops.rs::production_source`. Filed **#5450**, which surveys the rest of the crate for the same shape.

**The conflict nearly cost a production function.** Both hunks in `runtime.rs` are this branch deleting something main had meanwhile added to. In the second, main put `wasm_code_hash` — a production function with a live caller on the PUT path at `runtime.rs:1352` — into the very region this branch deletes the `INSTANCE_ID` static from. Taking the incoming side wholesale would have deleted it. Resolved by keeping main's content minus only the two symbols this branch removes, after checking that `AtomicI64` had no other use, `Path` had four, and `wasm_code_hash` had a caller. The rebased diff against main matches the original against its old base exactly: 4 files, 189 insertions, 37 deletions.

[AI-assisted - Claude]

sanity:
## Consolidated review — rebased onto current main, re-reviewed

**This review covers `da0d9cdd7`**, the rebased content — not the previous head `678e8821c`, which the earlier lens comments on this PR were written against. The branch is rebased onto `origin/main` (`af7c52e7f`).

The previous review was stale, so it was re-run: two independent blind lenses, **skeptical** and **code-first**, each reading the checked-out code rather than the diff text, each told not to rubber-stamp. No external (non-Claude) pass ran; per the repo's `pr-quality.md` the external pass is opt-in and a Claude lens was substituted.

### What the conflict resolution actually did

There have been two rebases. The first (onto `3a03ad07a`) hit one real conflict, in `crates/core/src/wasm_runtime/runtime.rs`. Main had inserted roughly 200 lines of host-clock-deprecation code — `decide_host_clock_warning`, sitting directly below `wasm_code_hash` — into the exact region this PR deletes `static INSTANCE_ID` from. Taking "theirs" wholesale would have deleted that production code and still compiled cleanly.

The resolution keeps the entire main-side block and deletes only the single `static INSTANCE_ID` line. The branch's whole delta on that file is three hunks:

1. drop the now-unused `use std::sync::atomic::AtomicI64` (nothing else in the file references it);
2. drop `static INSTANCE_ID`;
3. `RunningInstance::new` calls the two-argument `create_instance` and reads `handle.id`.

No hunk deletes main-side content. Both main-side functions survive with live call sites: `wasm_code_hash` at `runtime.rs:63`, called from `runtime.rs:1552`; `decide_host_clock_warning` at `runtime.rs:238`, called from `runtime.rs:205`.

The second rebase (onto `af7c52e7f`, 21 further commits) applied with no conflict. The strongest available check was run on it: diffing the pre-rebase PR diff against the post-rebase PR diff gives **zero content-line differences** — the only differences are one blob hash and one hunk-header offset (`@@ -290` to `@@ -291`), caused by `cf7b200be` shifting `native_api.rs` by a line. The rebase changed nothing about this PR's own content.

### Verification

- **Arity sweep.** A whole-repo grep for any three-argument `create_instance(` returns nothing. One production call site (`runtime.rs:304`), still wrapped in `classify_result`, so the `guest_entry_calls_route_through_classify_result` pin in `contract/executor.rs` still measures 1/1.
- **Source-scrape pins.** Both `create_instance_allocates_its_own_instance_id` and `create_instance_recovers_store_on_guest_entry_failure` were re-checked against the current file: both bracket lines 928-991, the production `create_instance` body, and both find their needles.
- `cargo check --all-targets` clean at `da0d9cdd7`; `cargo fmt` clean; working tree clean at that SHA.

### Findings addressed

**Both lenses independently flagged the same overclaim, and it is the most useful thing they found.** The rustdoc said issuing ids inside `create_instance` "makes a caller-supplied id a compile error". It does not. `InstanceHandle.id` is `pub(super)`, so anything inside `wasm_runtime` can still construct a handle with a chosen id and hand it to `drop_instance` — `delegate/test.rs:4285` and `:4346` build `InstanceHandle { id: 0 }` today, inert only because `process_outbound` binds it as `_handle` and never touches the global maps. The property is enforced by nobody having tried, not by the type system.

Fixed by correcting the claim in both rustdocs and in the PR body rather than by restructuring the type. A PR that overstates its own guarantee is worse than one that understates it: the next reader stops checking. What the signature genuinely closes is the `create_instance` parameter, which is where every id that actually reached these maps came from.

**Hardened in passing:** `create_instance_recovers_store_on_guest_entry_failure`, the pin this PR's new pin names as its model, scraped the whole file with no test-module cutoff. A rename of `create_instance` would have let its `find` fall through into the test module instead of panicking; it failed closed only by the accident of which string literal sits earliest. Same cutoff applied.

### Findings considered and not changed

- `SeqCst` on the allocator is stronger than uniqueness requires. It is exactly what the `INSTANCE_ID` static it replaces used; changing it would be an unrelated behavioural edit.
- `i64` wraparound after 2^63 allocations could eventually reach the `-1` sentinel. Identical to the pre-change static, and unreachable.
- One lens read the `+ 1` comment on the region-end index as describing the wrong byte. Checked against the file: the text at the match is `}\n\n    fn drop_instance`, and the byte the `+ 1` preserves is the newline closing the method, which is what the comment says. No change.

### On CI, and what a green tick here does and does not mean

CI runs `cargo nextest` (`ci.yml:893-903`), which gives every test its own process. **This defect class is invisible to it at any repeat count**, before or after the fix, because the collision requires two tests in one process. A green CI tick on this PR is not evidence the bug existed or that it is fixed.

The meaningful signal is plain `cargo test` in one process. That run is in progress on `da0d9cdd7` and its result will be posted here; it was green on the pre-rebase content, and the rebase is proven above to have changed none of this PR's own content.

Equally, no rate in this PR should be read as validation. The originally reported 3-in-9 has **not** been independently reproduced, and a later attempt's 0-in-20 could not have detected the defect either — it ran 538 of 5267 lib tests in about 20 seconds, stripping most of the parallel load that #4213 names as the trigger. This change stands on the mechanism: two independent issuers writing into one process-global keyspace is visible by reading `create_instance` against `MEM_ADDR`, and the fix removes the second issuer.

Lenses run: skeptical, code-first.

[AI-assisted - Claude]

Issue #4963, opened 2026-07-27
feat(packaging): Create maintainable freenet packages


mattgeddes:
I watched this YouTube video introducing Freenet (https://youtu.be/3RBNboYUlVI?si=OpvL0vmF9fdgUY7r) and noticed a comment from a viewer about needing proper packages for distribution package management systems. I agree and said that in downloading/running Freenet to check it out, I'd consider taking a look at what it would take to package it suitably for most distribution package managers.

I have a small prototype changeset that's purely additive (ie, doesn't touch anything existing so shouldn't cause any kind of regression). So I figured I'd raise a Github issue, confirm some details and then raise a pull request once I've cleaned up the prototype and have done some more testing. I hope this is OK.

What I feel I can add in under this specific issue is:
- For any Linux (presumably statically linked MUSL libc) binary on any supported architecture, generate a package for `freenet` (possibly including `freenet-ping`) and a package for `fdev`, similar to the tarballs shipped today.
- These packages would be:
   - Debian-style .deb packages (will also work on *buntu)
   - RedHat-style .rpm packages (will likely work on *SuSE, and I may be able to expend some time testing that if necessary)
   - ArchLinux-style packages.

I have .deb and .rpm working now (needs some tidying) and have a plan for Arch that I'm confident ought to work.

The result would be pretty straight forward to add to your existing CI/CD pipeline. You'd just get extra artifacts out. I'd provide docs for it too obviously.

@sanity , let me know if this appeals to you and I'll get it finished and a PR raised hopefully within a few days.

I don't plan to explicitly build an apt etc repository for automatic updates, but do have some ideas around that that I had planned to test with some of my own Rust code first.

I also have some thoughts on making this more convenient for Desktop Linux users, but I'll raise separate issues/PRs for that if the packaging stuff works out.

sanity:
# Issue Priority Assessment

Generated: 2026-07-29T16:01:45.745263049+00:00
Issue hash: `4e4698aa`

## Value Assessment

This confirms the release pipeline currently only ships raw tarballs (via `cross-compile.yml`, `actions/upload-artifact` per-platform binaries), no `.deb`/`.rpm`/AUR packaging exists anywhere in the repo — so the issue is fully relevant and not obsolete.

**Assessment**

[...]

PR #5267, opened 2026-08-11
build(packaging): initial .deb and .rpm packages for #4963

mattgeddes:
I didn't hear back on my [proposed changes](https://github.com/freenet/freenet-core/issues/4963) for adding .rpm and .deb package builds (per a YouTube comment request on a Freenet intro video).  So I've started out just by adding a basic .deb and .rpm package.

These don't change the current build at all -- just add to it. There are no additional _package_ dependencies introduced either -- just a couple of tool dependencies that can be installed with:
- `cargo install cargo-deb`, and
- `cargo install generate-rpm`

With both of those installed, and building on, say, aarch64 Linux, after, or instead of, `cargo build --release`, running `cargo deb` will result in a debian package being built with for aarch64 and including the freenet binary.

The same applies for creating the RPM package. `cargo generate-rpm -p crates/core` will package up the binary into an aarch64 package. 

In both cases, a `freenet` user is created, there is a systemd unit file included that starts freenet as a service. It is disabled by default, but can be enabled with `sudo systemctl enable freenet` or just started once with `sudo systemctl start freenet`. All of Freenet's configuration, state and data is placed/stored in the freenet user's home directory, meaning that it is cleaned up on package removal. 

I've done the bulk of my testing on Debian and Ubuntu. I haven't tried cross compiling and building and suspect I'll have some questions about your CI/CD pipeline if you're mainly cross compiling. I've done some minimal testing of the RPM -- I'm currently travelling. I didn't (yet) create ArchLinux packages, but the approach seems the same. I also haven't created a separate package for `fdev` yet, as planned, but would be happy to if the current work is met with joy.

Feel free to reach out via the PR comments or Github issue with any questions or comments. Just be sure to tag me -- I don't think I'm automatically CCd on Issues being outside the org.

I do have quite a few technical questions about the freenet approach, and can also see myself making a few more contributions over time.

I hope this helps.


sanity:
Thanks @mattgeddes — and sorry for the silence, both here and on #4963. Distro packages are something we've wanted and nobody has done, so this is genuinely welcome. Reviewed it properly now.

The Cargo.toml metadata approach is the right one, and I like that it's purely additive to the existing build. Two things need fixing before it can land, and a handful of smaller notes.

## Blocking 1 — the systemd unit doesn't implement Freenet's exit-code contract

This is the one that matters. `freenet network` uses exit codes to talk to its supervisor:

- **42** — "a new release exists, please update me". The node does *not* update itself; the supervisor is expected to run `freenet update` and restart.
- **43** — "another instance is already running on this port". Must not restart.
- **45** — fast-crash / boot-wedge, must count toward the crash-loop limiter.

The unit generated by `freenet service install` (`crates/core/src/bin/commands/service/linux.rs:435-560`) encodes all of that:

```ini
Environment=FREENET_SUPERVISED=1
ExecStart=/usr/bin/freenet network
Restart=always
RestartSec=10
SuccessExitStatus=42 43
RestartPreventExitStatus=43
TimeoutStopSec=45
ExecStopPost=-/bin/sh -c 'case "$EXIT_STATUS" in 0|43) ;; *) ... freenet update --quiet ;; esac'
```

The packaged unit has `ExecStart` and `Restart=always` and none of the rest. Concretely, on a packaged node:

- The first time a new release ships, the node exits 42. Nothing runs `freenet update`, and `RestartSec` is unset (systemd default 100 ms), so systemd restarts immediately, the node re-detects the same update, exits 42 again — and within a second or so trips the default `StartLimitBurst`, at which point the unit enters `failed` and **the node stays down**. The package would work fine right up until the next release and then break, on every installed machine at once.
- Without `FREENET_SUPERVISED=1` the node also logs a loud "no supervisor" error on that exit, which is the correct complaint.
- Exit 43 restart-loops for the same reason.

The safest fix is not to hand-write a second unit at all, but to render the same text `service install` produces — the directives above have a lot of history behind them (#4073, #4549, #4551, #4580, #3967) and a duplicate will drift away from them silently, and won't be covered by the pin tests in `service.rs` that guard the generated one. If factoring that template out is more than you want to take on, we're happy to do that part — say the word.

There's a design question underneath it that's ours to answer, not yours: whether a distro package should manage the service itself or defer to `freenet service install`. I'd like @sanity / @iduartgomez to weigh in before you spend more time on the unit.

## Blocking 2 — `userdel freenet` runs on upgrade, not just removal

```toml
pre_uninstall_script = """
userdel freenet
"""
```

`pre_uninstall_script` maps to RPM's `%preun`, which runs on **upgrade** as well as erase (`$1` is the count of instances that will remain: `1` on upgrade, `0` on erase). So `rpm -U` deletes the service account out from under the running service. It also fails non-zero if the user doesn't exist or still owns a running process, which aborts the transaction. Needs the standard guard:

```sh
[ "$1" = "0" ] || exit 0
userdel freenet 2>/dev/null || :
```

Though I'd lean toward not deleting the user at all — most distro policy is to leave system users behind, precisely because their files outlive the package.

Relatedly, the description says *"All of Freenet's configuration, state and data is placed/stored in the freenet user's home directory, meaning that it is cleaned up on package removal."* That isn't what happens in either package: `userdel` without `-r` leaves `/var/lib/freenet` in place, and the `.deb` has no `prerm`/`postrm` at all, so nothing is removed there even on `purge`. Worth either implementing or dropping the claim — I'd suggest dropping it and leaving node data behind on uninstall, which is the less surprising behaviour for a node with peer state and keys.

## Smaller notes

- **`license = "AGPL"`** in the `generate-rpm` table — the SPDX identifier is `AGPL-3.0-only` (see `LICENSE.md`); bare `AGPL` isn't valid SPDX and `rpmlint` will complain. `crates/core/Cargo.toml` itself uses `license-file`, so there's nothing to inherit from.
- **`freenet.service` installed with `mode = "0755"`** in the RPM assets — unit files should be `0644`.
- **Unit install path** `/lib/systemd/system` — `/usr/lib/systemd/system` is the correct location on modern RPM distros. It works today via the merged-`/usr` symlink but rpm is unhappy about writing through one.
- **`.deb`/`.rpm` asymmetry**: the RPM ships `LICENSE.md` and `README.md` under `/usr/share/doc/freenet/`, the deb doesn't. The deb creates the user with `adduser --system`, the RPM with `useradd -r -m`. Neither is wrong, but they'll be easier to keep honest if they agree.
- **No CI coverage.** Nothing runs `cargo deb` / `cargo generate-rpm`, so a broken asset path or a malformed table won't be noticed until someone tries a release build — and the metadata will rot. Not a blocker for the first cut, but if you're willing, a job running both in check mode would make this durable. (We can add it if you'd rather not.)
- **`assets` hardcodes `target/release/freenet`** — fine for a native build; worth a note in the docs for the cross-compile case, since our release pipeline cross-compiles.

## On your questions

Yes, our release binaries are cross-compiled (`.github/workflows/cross-compile.yml`), so if you do get to `fdev` and to CI wiring, that's the constraint to design around. And separate `fdev` packages would be welcome — but let's get this one landed first rather than growing the PR.

Marking this as needing changes rather than merging, but the direction is right and I'd like to see it land. Thanks again for picking it up off a YouTube comment, of all places.

[AI-assisted - Claude]

sanity:
Two housekeeping notes now that CI is actually running on this (the workflow runs had been sitting unapproved — outside-contributor runs need a manual approval and nobody had given one, which is a large part of why this stalled):

- **I retitled the PR.** `Conventional Commits` was failing because the title was missing the colon after the scope: `build(packaging) Initial ...` → `build(packaging): initial ...`. Nothing else changed; revert it if you'd rather word it differently.
- **`rule-review/warnings` is red** with the "no CI runs `cargo deb` / `cargo generate-rpm`" warning from my earlier review. That one resolves either by adding a check-mode job or by a maintainer applying `review-override`. I'd rather not override it — the metadata will silently rot without a build behind it, and packaging config is exactly the kind of thing nobody notices is broken until a release.

The rest of the suite is running now; I'll follow up if anything else goes red that isn't already covered above.

[AI-assisted - Claude]

sanity:
Correction to one line in my review, because it would send you looking for a log message that never appears.

I wrote that without `FREENET_SUPERVISED=1` the node "logs a loud 'no supervisor' error". Under your unit it will not. systemd sets `INVOCATION_ID` for every service, and `detect_supervisor_status` (`crates/core/src/bin/commands/auto_update.rs:95-107`) treats that as `SupervisedUnverified` — a soft warning, not the loud `Unsupervised` error.

That makes the situation slightly worse rather than better, and it is the sharper version of the point: under the packaged unit the node believes a supervisor is present, so it exits 42 confidently, while nothing on the system ever runs `freenet update`. The failure is quiet. A packaged node would stay pinned to its install-time version indefinitely and take a restart burst on each new release, with no obviously-wrong log line pointing at the cause.

Everything else in the review stands — including that the exit-42 behaviour is not conditional on supervision (`crates/core/src/bin/freenet.rs:331-342`: an unsupervised or custom-unit node "must keep exiting 42").

[AI-assisted - Claude]

mattgeddes:
> Thanks @mattgeddes — and sorry for the silence, both here and on #4963. Distro packages are something we've wanted and nobody has done, so this is genuinely welcome. Reviewed it properly now.
> 
> The Cargo.toml metadata approach is the right one, and I like that it's purely additive to the existing build. Two things need fixing before it can land, and a handful of smaller notes.

Great. Thanks for the feedback. Coming into the project fresh, there was quite a bit I missed (like the return code semantics). All of the comments seem reasonable and I'll try and knock them off this week, addressing them in the review comments as well as additional commits.

We're on the same page about tackling a separate `fdev` package once we have this PR in a state you're happy with. Another follow-up PR would be to add ArchLinux package builds to Cargo.toml too. But I'll do that separately.

One suggestion I have around whether freenet manages its own systemd unit file... we're running as a non-root user (on purpose, because there's nothing the service does yet that needs to be privileged). We ought to be able to keep your return-code semantics for signalling updates, and have a static systemd unit file without breaking the package management system. If folks are looking to use their preferred package management system to install a package, having the package then try to update files tracked (and hashed) by the package management system will send alarm bells ringing.  As you say, it's for you guys to ultimately decide, but feel free to loop me into the discussion (via a github issue or whatever you prefer) and I'll gladly offer suggestions/input for trying to get the best of both worlds.

Thanks again. Also, once the dust settles a little, I'd love to chat about the wider vision and approach.

sanity:
## The packaged unit doesn't just miss auto-update — it would brick the node

@mattgeddes — following up with the specifics, because my earlier note said the packaged node would sit "quietly pinned" to its install-time version. Having now compared the shipped unit against the one `freenet service` generates, that was too kind. It is worse than pinned.

`crates/core/scripts/debian/freenet.service` is:

```
[Unit]
Description=Freenet Gateway Service
After=network.target

[Service]
ExecStart=/usr/bin/freenet
User=freenet
WorkingDirectory=/var/lib/freenet
Restart=always
```

The unit the node's own installer writes carries, at minimum, `SuccessExitStatus=42 43`, `RestartPreventExitStatus=43`, an `ExecStopPost` that runs `freenet update` on a non-graceful exit, `Environment=FREENET_SUPERVISED=1`, and `StartLimitAction=none` / `StartLimitBurst=5` / `StartLimitIntervalSec=120` in `[Unit]` (the placement matters — systemd silently ignores those in `[Service]`, see #4551). See the assertions in `crates/core/src/bin/commands/service.rs:1436-1569` for the full contract.

**What happens on the first release that triggers a version mismatch:** the node exits 42. Because 42 is not in `SuccessExitStatus`, systemd counts it as a failure. Because there is no `ExecStopPost`, nothing ever runs `freenet update`, so the restart comes back on the same binary and exits 42 again. With the default start limit, systemd gives up and leaves the unit failed. So a packaged node doesn't stay on an old version — it stops, and the journal blames a crash loop rather than a missing updater.

That is squarely the failure mode #4073 exists to prevent, so it needs to be right before packages ship.

**Two ways to fix it, and I'd take the second.** You can copy the directives above into the shipped unit — but then there are two definitions of the unit in the tree that must be kept in step, and the one under `scripts/debian/` has no test behind it. Better: have the package's maintainer script call the node's own service installer so the unit is generated from the single source that already has `service.rs`'s tests guarding it.

### Other things I found while in there

- **The unit has no `[Install]` section.** `systemctl enable freenet` fails with "no installation config" — and the deb metadata says `systemd-units = { enable = false }` precisely so the user enables it themselves, so this blocks the documented path. Needs `[Install]` / `WantedBy=multi-user.target`.
- **`pre_uninstall_script` deletes the user on UPGRADE, not only on removal.** RPM runs `%preun` for the outgoing package during an upgrade with `$1 == 1`, so `userdel freenet` fires mid-upgrade and the new package's `%post` will not recreate it (`id -u` still succeeds only if the delete failed). Guard it: `[ "$1" = 0 ] || exit 0`.
- **The `.deb` never removes the user** it creates in `postinst`, while the RPM does. Asymmetric — pick one policy. (Leaving the user in place on purge is a defensible choice; the asymmetry is the problem.)
- **`freenet.service` is installed `mode = "0755"`** in the RPM asset list. Unit files should be `0644`.
- **Neither package ships `fdev`**, only `freenet`. Intentional? Worth stating either way.
- **The RPM installs the unit by hand** into `/lib/systemd/system` without a `daemon-reload`, while the deb path goes through `systemd-units`. Those two will drift.

### On the `rule-review/warnings` red

Still standing, and I'd rather not override it: nothing in CI runs `cargo deb` or `cargo generate-rpm`, so a broken asset path or malformed table stays invisible until someone tries a real release build. A check-mode job would also have caught the missing `[Install]` section. I'm happy to add that job myself if you'd like — say the word and I'll push it to this branch rather than making it another round trip for you.

Nothing here is a criticism of the diagnosis: packaging is genuinely wanted (#4963) and the metadata is close. It's the unit that needs to come from the same place as the installer's.

[AI-assisted - Claude]

mattgeddes:
Hi @sanity. 

Many apologies for the delay in finishing this off. I got caught up in my own decentralised web project (not competing with freenet) and a few prototypes around securing AI to try to show people we have solutions to a number of problems. Long story short, I've gone through each of your recommendations about and have fixes for all (except the CI/CD one) as two commits -- one for the systemd unit file changes and one for the Cargo.toml rpm/deb changes.

In case I've missed any, I'll lay out here what I've fixed between the two commits and we can haggle over anything I've misunderstood or missed :)

1. systemd exit code contract -- I believe this is fixed in the second commit
2. userdel on remove -- You're right and I've removed that.
3. AGPL string -- Fixed
4. Executable unit file in RPM -- Fixed
5. RPM prefers /usr for service unit files -- Fixed
6. Debian package doesn't include docs and differences in useradd/adduser invocation -- I went with the more LSB-compliant useradd approach and made them both the same. Debian automatically includes the license-file file, so I added a note and explicitly included the README.md file, so both RPM and DEB packages have the two docs included.
7. CI/CD coverage -- _I didn't make this change_. I'm more of a developer with pretty intimate knowledge of the operating system, but I haven't done much with sophisticated CI/CD pipelines and wasn't sure of myself with this one. Happy to assist anyone working on it with what I know about the package building stuff if it'll help.
8. Hard-coded `target/release` path in binary artifacts. I didn't like this either. Both cargo-deb and cargo-generate-rpm claim that they use `target/release` as a placeholder and explicitly replace it. I added comments to this effect, and did a quick test with the RPM, where I did a clean build (so no `target/release` on the filesystem), and build with `--profile dev` and built the package with `--profile dev` too, and it build the package.

I've also checked the formatting of the metadata you're putting in the commit messages and I think I've got that right this time too.

I hope that all of this helps. Happy to address any other concerns or comments or just to chat about the project, its goals and future.

Issue #5264, opened 2026-08-11
Client API has no authentication and issues app identity on request


sanity:
## Problem

The node's WebSocket client API has no authentication, and app identity is **issued on request** rather than proven. A token can be supplied but is treated as metadata, never as a gate — a connection with no token proceeds fully (`crates/core/src/client_events/websocket.rs`).

Those two together mean any process that can reach the API can present itself as any application, and thereby reach that application's delegate secrets. The delegate sandbox exists precisely so an app's secrets are not merely readable local state, so this undercuts a boundary we advertise.

**Scope, stated accurately:** this is a *local* exposure. The API defaults to loopback — `resolve_ws_api_address` returns `WsApiAddressSource::DefaultLoopback` unless an operator sets an explicit address or grants CIDRs, and a persisted wildcard is treated as auto-derivable and dropped. An earlier framing of this as a LAN-reachable issue is stale; that vector is closed. What remains is same-machine.

This is hardening, not a fix for a remotely-exploitable hole. Filing it because it is currently tracked nowhere.

## Constraint: friction is the binding requirement

Whatever we build has to be near-invisible. A node that demands credential setup before anything works will push users toward disabling it, which is worse than not having it. The design target is **setup on first connection to the web interface** and nothing thereafter.

That means trust-on-first-use: the first client to connect establishes the credential. It is a real hole if an attacker gets there first, and it is very likely an acceptable one — it is the SSH host-key model, and the window is a local process racing the user to a node they just started. Worth stating explicitly in the design rather than leaving implicit.

## Why passkeys look like the right primitive

A node-wide shared secret would gate *access* to the API but would not address the second half of the problem: anything holding the secret could still claim any app's identity.

A **passkey is origin-bound**. If the credential lives in the browser, the app's own origin is what holds it, and separation between apps falls out of WebAuthn rather than requiring a separate per-app identity scheme layered on top. That maps onto the property we actually want — an app can prove it is itself — using a primitive users already have zero-friction flows for.

Open questions this raises rather than settles:

- Non-browser clients (`fdev`, `riverctl`, tooling) need a path that isn't WebAuthn. A local file credential is the obvious answer; how it is scoped is the question.
- How an origin maps to the app identity the delegate layer authorizes against, given delegates are keyed by hash today.
- Whether existing app identity issuance stays as-is for unauthenticated clients during a transition, or is gated from the start.
- Recovery when the credential is lost, without recreating the hole it closes.

## Related

[...]

Issue #4945, opened 2026-07-26
0.2.107: shell_bridge SecurityError in sandboxed iframe kills freenetBridge — all locally-served web apps hang


sanity:
Reading `navigator.serviceWorker` throws a SecurityError in a sandboxed document without `allow-same-origin`: the property exists on Navigator (so the `'serviceWorker' in navigator` feature-check passes) but its getter throws. #4937 added an eager `installNotifyClickListener()` call at the tail of `freenetBridge()` that reads it unguarded, so in 0.2.107 the uncaught throw kills the bridge before its message handlers install. Result: every web app served by a local node (River, Delta, Atlas) hangs at loading / WS connect timeout. Observed on a real peer immediately after upgrading to 0.2.107 (console: `Uncaught SecurityError ... at installNotifyClickListener`). Fix: feature-detect by attempting the read in try/catch and route all serviceWorker access through the accessor.

[AI-assisted - Claude]

sanity:
Fixed in v0.2.108, now released. The 0.2.107 regression was an unguarded read of navigator.serviceWorker in the served shell bridge — in the sandboxed app iframe that read throws a SecurityError, which killed the bridge before its message handlers installed, so every locally-served web app (River/Delta/Atlas) hung at loading or WS timeout. All serviceWorker access now goes through a try/catch accessor (PR #4946).

Verified on a real NATed peer: browser-level Playwright check renders River rooms with live content, zero console errors, on the official 0.2.108 binary. Auto-update will roll peers forward over the next 0-4h stagger window; a manual `freenet update` (or reinstall) fixes it immediately.

[AI-assisted - Claude]

PR #5269, opened 2026-08-11
fix(server): carry permission events over a WebSocket, not SSE (#5213)

### LINKED ISSUE #5213: Web apps hang on "loading..." once ~6 tabs are open: per-tab SSE stream exhausts the browser's per-origin HTTP/1.1 connection limit
## Problem

A user reported that after opening several Freenet apps in new browser tabs, an app tab got stuck permanently showing "loading...". It recovered on its own after they closed an unrelated tab.

That symptom is consistent with exhausting the browser's per-origin HTTP/1.1 connection budget, and the code paths line up:

1. **The local server is HTTP/1.1 only.** `axum` is pinned with `default-features = false` and the feature list does not include `http2` (`crates/core/Cargo.toml:41`, workspace pin `Cargo.toml:35`), and the server is plain `http://` (`crates/core/src/server.rs:210-217`), so no browser would negotiate h2c anyway. There is no multiplexing.

2. **Every web app shares one origin.** All contracts are served path-namespaced from the same `http://[::1]:7509` / `127.0.0.1:7509` (`server/client_api/v1.rs:26-28`, `v2.rs:18-20`; port default `config.rs:2861`). There are no per-contract subdomains or ports, so the browser's 6-connections-per-origin cap is global across *all* Freenet tabs, not per app.

3. **Every open tab holds a long-lived HTTP connection open.** The shell bridge unconditionally opens a Server-Sent Events stream per tab: `var es = new EventSource('/permission/events');` (`server/path_handlers/assets/shell_bridge.js:1879`), served by `Sse::new(...).keep_alive(...)` at `server/client_api/permission_prompts.rs:104,714,822-823`. This is documented in the bridge itself at `shell_bridge.js:1476-1478` ("Every open Freenet tab subscribes to /permission/events"). It was added in `12ccd8272` (#4023).

So each open tab permanently occupies one of the ~6 sockets the browser will allow to that origin. At around six open tabs the budget is gone, and the next tab's shell HTML, iframe document, wasm/JS assets and `/permission/pending` fetches all queue behind the held-open SSE responses and never complete.

The WebSocket itself is *not* the constraint — browsers pool WebSockets separately (Chrome ~255/profile, Firefox 200) — and neither is the server: `MAX_SSE_CONNECTIONS = 64` (`permission_prompts.rs:138`) is an order of magnitude above where the browser blocks.

## Why it looks like a hang rather than an error

The block is silent. The queued requests just sit in the browser's connection queue; nothing errors, so `es.onerror` never fires and the 3s `/permission/pending` fallback at `shell_bridge.js:1870` never kicks in either. The app UIs have no timeout on this path:

- Atlas renders "Loading…" whenever `STATE` is `None` (`ui/src/main.rs:222-225`) and only leaves it after a `GetResponse` for the index contract arrives (`main.rs:368-374`). If the wasm bundle never loads or the WebSocket never opens, it stays there indefinitely.
- River shows `SynchronizerStatus::Connecting` (`ui/src/components/app.rs:56-57`) and "Loading your rooms…" (`ui/src/components/room_list.rs:53-55,83`) under the same conditions.

Closing any tab frees a socket and the pending tab suddenly completes, which matches the reported "it started working again after closing another tab".

## Impact

Anyone browsing more than a handful of Freenet apps at once hits this, and the failure presents as "the app is broken" with no diagnostic. Tabs 5 and 6 are already degraded (1-2 free sockets for asset loading) before the hard wall at 7.

## Possible directions (not yet decided)

- **Enable HTTP/2.** Would multiplex everything over one connection and remove the cap entirely, but h2 requires TLS in browsers, and h2c (cleartext) is not supported by any major browser, so this needs a TLS story for the local server first.
- **Stop opening an SSE stream per tab.** Elect one tab as the permission-event listener via `BroadcastChannel` or a `SharedWorker`, and have it fan out to the others. Removes the per-tab long-lived HTTP connection entirely and is probably the smallest fix.
- **Push permission events over the existing WebSocket** instead of a separate SSE channel — the shell already holds a WebSocket per tab, and those don't count against the 6-per-origin pool.
- **Surface the condition.** At minimum, a timeout in the shell bridge that reports "too many Freenet tabs open" rather than an indefinite "loading…".

## Verification status

The mechanism is confirmed statically (HTTP/1.1-only server + one `EventSource` per tab + single shared origin). It has **not** been reproduced in a browser, so the exact tab count at which it breaks is unconfirmed and will vary by browser and by whether a tab finished loading its assets before its EventSource was established. A repro would be: open N Freenet app tabs, watch devtools Network for requests stuck in "Queueing"/"Stalled", and confirm N.

[AI-assisted - Claude]

sanity:
## Problem

Freenet web apps hang permanently on "Loading…" once about six browser tabs are open, and recover the moment an unrelated tab is closed. Reported by Ian and others; tracked as #5213.

Three things combine:

1. **The local server is HTTP/1.1 only.** `axum` is pinned with `default-features = false` and no `http2` feature (`crates/core/Cargo.toml:41`), and the listener is plain `http://`, so there is no multiplexing.
2. **Every web app shares one origin.** All contracts are served path-namespaced from `127.0.0.1:7509`, so the browser's ~6-connections-per-origin budget is global across *all* Freenet tabs, not per app. Mixing River + Delta + Atlas hits the wall exactly as fast as six tabs of one app.
3. **Every tab permanently held one of those connections.** `shell_bridge.js` opened `new EventSource('/permission/events')` unconditionally per tab, served as a keep-alive SSE stream. That connection lived for the life of the tab.

Six tabs, six connections, budget gone. The seventh tab's shell document, wasm, JS and `/permission/pending` fetches all queue in the browser and never go out. Nothing errors, so `es.onerror` never fires, the 3s polling fallback never engages, and the app UIs (which have no timeout on this path) sit on "Loading…" forever.

### Reproduced

Headless Chromium against a live 0.2.123 node, opening real contract app tabs one at a time:

```
tab 1-6 |  28-36ms | doc=200   | ok
tab 7   |  15001ms | doc=NEVER | STALLED (15s navigation timeout)
```

`doc=NEVER` is the shell document request never receiving a response — a browser-queue stall, not an application state. Closing the stalled tab plus two others and retrying the same URL: **31ms**. Control against the node homepage (same origin, same server, no `EventSource`): **12 tabs, no stall**, which isolates the SSE stream as the cause rather than anything general about the listener. Details in [the issue](https://github.com/freenet/freenet-core/issues/5213#issuecomment-5257919161).

## Approach

Move the permission-event channel from SSE onto a WebSocket the shell opens for itself: new `GET /permission/events/ws`, and `shell_bridge.js` uses it instead of `EventSource`.

Browsers pool WebSockets separately from the per-origin HTTP limit and far more generously (~255 per profile in Chrome, 200 in Firefox), so the channel stops competing with page and asset loads. The per-tab connection still exists; it just comes from a pool ~40x larger.

**Why not the alternatives.** Electing one SSE listener per browser via `BroadcastChannel` would also work but needs leader-election and failover logic for when the listener tab closes. Pushing these events over the app's existing client-API WebSocket would mix trusted shell control-plane traffic into the sandboxed app's data channel (exactly the separation #3836 established), requires a client-API wire change with the pinned-consumer compat risk from the v0.2.11 incident, and does not even work when the app has not opened a socket — the shell's WebSocket is created on demand from a postMessage, so it may not exist. HTTP/2 removes the limit wholesale but needs TLS on the local listener first, since no major browser speaks h2c.

**What is deliberately unchanged.** The #3836 trust boundary: prompts still render in the shell's DOM outside the sandbox, over a channel the sandboxed app cannot see or touch. Event names and `data` payloads are byte-identical to the SSE stream (the name simply rides inside a JSON envelope, since a WebSocket frame has no `event:` slot), which is what lets the bridge share one set of handlers. The loopback gate (#3819), the origin/`Sec-Fetch-Site` gate, the subscriber cap, and the subscribe-then-snapshot race avoidance all carry over unchanged.

**The SSE route is kept and still served.** A tab opened before a node upgrade still has the old bridge running. Being precise about what that buys, because the first draft of this paragraph overstated it: the pre-#5213 client registers an error listener on the `EventSource` and falls back to polling `/permission/pending` every 3s, so dropping the route would degrade those tabs to that fallback (a prompt takes up to ~3s to appear) rather than break them. Keeping it costs nothing and spares them the latency. Retirement is tracked in #5272. The client-side pin therefore asserts the *shell* no longer opens one, not that the route is gone.

### Security note

This is the part most worth reviewing, and the review caught a real hole in the first version of it — details under "Review findings" below.

Browsers do **not** apply the same-origin policy to WebSocket handshakes: any page anywhere can open a socket to `ws://127.0.0.1:7509/...` with no CORS preflight, and no CORS response header can withhold the payload afterwards. Browsers *are* required to send `Origin` on every WebSocket handshake, including same-origin ones, so the origin header is the only thing available to gate on — and it therefore has to be checked strictly.

The route requires, in order and all **before** the upgrade and before a subscriber slot is claimed: a loopback peer address (#3819), a caller that passes the existing allowed-host/CIDR check, **and** an `Origin` that is present and whose authority equals the `Host` header exactly. A hostile page can neither read the stream nor exhaust the cap by looping handshakes.

Getting this wrong would leak raw permission nonces to any page the user visits, and a nonce is a capability — `/permission/{nonce}/respond` approves a delegate action with it. `permission_ws_rejects_cross_origin_handshake` covers it (four hostile origins), together with `permission_ws_rejects_other_localhost_ports` and `permission_ws_rejects_handshake_without_origin`, all paired with `permission_ws_accepts_same_origin_loopback_handshake` so the rejection tests cannot pass vacuously on a malformed request.

The subscriber cap is now shared across both transports rather than per-transport, since two independent caps would silently allow twice the intended ceiling during an upgrade when both are in use. `MAX_SSE_CONNECTIONS` / `SSE_CONNECTIONS` are renamed to `MAX_PERMISSION_SUBSCRIBERS` / `PERMISSION_SUBSCRIBERS` to match.

## Testing

**New server tests** (`permission_prompts.rs`):

- `permission_ws_accepts_same_origin_loopback_handshake` — the control that keeps the rejection tests below non-vacuous.
- `permission_ws_rejects_cross_origin_handshake` — CSWSH, over four hostile origins including `null` and a `127.0.0.1.evil.example` prefix trap.
- `permission_ws_rejects_lan_peer` — #3819 parity.
- `permission_ws_rejects_missing_connect_info` — the fail-closed half of #3819, which also proves the handler consults `peer_is_loopback` rather than trusting the transport.
- `permission_ws_honours_shared_subscriber_cap` — fills every slot through the same helper the SSE path uses (so a separate counter would fail here), asserts 503, then releases one slot and asserts the next handshake is admitted, pinning that a cap rejection is transient.
- `permission_ws_delivers_added_and_removed_end_to_end` — real socket against a real served router.
- `permission_ws_replays_pending_on_connect` — bootstrap replay, so a tab loading while a prompt is already pending still renders it.
- `ws_envelope_wraps_data_unchanged` — the wire contract between handler and bridge.
- `permission_ws_slot_is_held_while_live_and_released_on_disconnect` — the cap is only meaningful if slots come back.
- `permission_ws_pump_emits_resync_when_the_broadcast_laps_it` — a lagging subscriber must be told to reconcile, not silently miss prompts.
- `permission_ws_pump_gives_up_on_a_client_that_stops_reading` — a stalled reader must not pin a slot forever.

**New browser-side unit tests** (`shell_bridge_permission_ws.test.mjs`, marker-bounded extraction of the real shipped source): URL scheme derivation, the backoff curve and its jitter band, envelope dispatch including every malformed shape, and a stateful machine suite driving a stubbed `WebSocket` — poll-before-handshake, stop-on-open, reconcile-on-open, poll-and-single-reconnect on close, compound backoff growth, stable-socket reset, and no double-schedule on a duplicate close.

Plus a suite for the prompt-loss race specifically: a prompt raised while the snapshot was in flight is not destroyed; a card shown in the *same millisecond* the request was issued survives (the tie resolves toward keeping it, since a wrongly-kept card is corrected by the next reconcile while a wrongly-hidden one is gone for good); a card the snapshot genuinely could have seen and did not is still hidden, so cross-tab dismissal keeps working; and a source-bounded pin that `showCard` still records the timestamp the comparison depends on, since the other three stub `showCard` and would all pass without it.

**Browser E2E for the bug class itself** (`crates/core/tests/playwright/tests/connection-exhaustion.spec.ts`, new). This is the only test that exercises the actual failure: a per-origin connection budget exists solely inside a browser, which is why the entire Rust suite stayed green while every app hung. It runs in the existing `playwright-shell.yml` harness across chromium, firefox and webkit. Two tests: one loads the shell, waits, and fails if **any** HTTP request is still unfinished (the real invariant — a streamed `fetch()` or long-poll would break #5213 the same way an `EventSource` did, and no substring pin can see that); the other opens 8 tabs in one browser context plus a 9th, asserting every navigation resolves *and* that every tab actually opened a permission WebSocket, so the run cannot pass by subscribing to nothing.

Both were mutation-verified rather than assumed: reintroducing a single held-open `EventSource` into `shell_bridge.js` and re-running the harness turns them red, which is what establishes that they are watching the real defect and not just a warm-cache page load.

**Regression pin** (`path_handlers.rs`): `shell_page_permission_overlay_present_and_safe` now asserts the shell JS contains **no** `EventSource(` — reintroducing a held-open HTTP request per tab is exactly the bug — plus that it subscribes to `/permission/events/ws` and owns its own reconnect (a WebSocket has no auto-reconnect, so without that a single node restart would strand every tab on the polling fallback). The existing `#4849 F2` marker-bounded Notification guard still covers the new handlers.

**Browser verification.** `shell_bridge.js` is the file that broke in #4945, where every `riverctl` check passed while every locally-served web app was broken, because riverctl bypasses the served shell. So this was verified by driving real Chromium against a node built from this branch: tabs well past the old wall all load, every tab opens a `/permission/events/ws` socket, no tab issues an SSE request, and there are no uncaught page or console errors from the bridge. Numbers in a comment below.

## Review findings

A Full-tier multi-perspective review (four specialist lenses plus a WebSocket-security lens; Codex is opt-in-only under the project's review policy and was not run) found two blocking issues. Both are fixed, and each fix was mutation-verified by reverting it and confirming the new test fails.

1. **A security regression this PR introduced.** The WebSocket route initially reused `is_caller_trusted`, which accepts any localhost origin on **any port**. Under SSE that was survivable: the trusted branch sets no `Access-Control-Allow-Origin`, so the browser withheld the body from a cross-origin reader. WebSockets have no CORS, so a page served by any other local listener could have read raw permission nonces. Found independently by two reviewers. Fixed by requiring `Origin` to be present and its authority to equal `Host`.
2. **A prompt-loss race.** `reconcileFromPending` hid every overlay card missing from the `/permission/pending` response, so a prompt arriving while that request was in flight was permanently destroyed — silently dropping a live security prompt. Cards now record when they were shown and are only hidden by a response issued after that.

Also addressed from the review: one deadline for the whole replay phase instead of one per message (which could stack to 32 x 10s against a stalled client); frame and message size caps; the fallback poll now starts before the handshake resolves, so a *hanging* handshake also degrades to polling; the reconnect backoff resets only after the socket has been stable for 10s, so an accept-then-close server cannot pin reconnects at 1Hz; all `onclose` side effects moved inside the socket-identity guard; a console warning after five consecutive failed handshakes; and a quiesce loop in the cap test, since WebSocket slots release asynchronously where SSE's `GuardedStream` released synchronously.

## Second review round

A fresh Full review (five blind lenses) ran on the updated HEAD, since the first round's fixes touched security and concurrency. It found no surviving Must Fix in shipped behaviour, but substantial **test vacuity** — and two of the claims in the section above turned out to be wrong. Correcting them is the main content of this round.

**Two retractions.**

1. The send timeout was reported as mutation-verified. It was not. The test supplied a replay frame, so the pump exited via the *replay* deadline and never reached the steady-state `send_bounded` path; the earlier mutation went red for a different mechanism than claimed. The test is now split so each deadline is exercised separately, and stripping the timeout makes the reworked one fail.
2. The Playwright non-vacuity control only proved a socket was *created*. Playwright fires `websocket` before the handshake resolves, so a 403 on every tab would have kept the spec green — and the `EventSource` mutation could not have caught it, since the socket is still created under that mutation. It now asserts the socket is live and error-free, which is what makes the new same-origin gate actually covered in a browser.

**Five regressions that previously passed every check now fail**: inverting `prompt_added` to hide cards, disabling `resync`, dropping the fallback poll's idempotency guard, dropping the socket-identity guard, and dropping the reconnect dedup. The last needed a different shape than suggested — the identity guard short-circuits before the dedup is reached, so it is isolated via the constructor-throw path instead.

**Security hardening**, all in code this PR added:

- The same-origin gate compared authority only, so `http://h` and `https://h` were the same origin. Behind a TLS-terminating proxy that is reachable. Now the scheme is compared too — but only when `X-Forwarded-Proto` is present. My first version treated its absence as "therefore http", which would have silently refused a legitimate TLS proxy that omits it, recreating the exact silent-degradation class this PR removes.
- `origin_authority` now **rejects** `@`, `?` and `#` rather than stripping them, so it cannot disagree with `is_localhost_origin`'s prefix matching. Under the old behaviour `http://127.0.0.1:7509@evil.example` was localhost to one parser and `evil.example` to the other, and both gates passed together.
- The legacy SSE path gets a lower ceiling than the shared cap, so a stalled SSE stream (which has no send timeout) cannot occupy all 64 slots and 503 every new tab into the polling fallback. One suggested fix here was **declined**: requiring `Origin` on the SSE route would break the pre-upgrade tabs the route exists to serve, because browsers omit `Origin` on same-origin `EventSource`.
- The loopback doc claim is qualified. It is a boundary against a direct off-box connection only; behind a proxy every client looks loopback. Pre-existing and tracked in #5264, but this PR is what wrote the claim down.

**Client fixes**: the `reconcileFromPending` ADD pass had the mirror of the bug fixed in round 1 and could resurrect an answered prompt (single timestamp, not a per-nonce map, which would grow unbounded); the reconnect stable-timer leaked in the two-socket case the identity guard exists for; the failure counter reset on open rather than with the backoff, so accept-then-drop never reached the warn threshold; and the causal comparison now uses a monotonic clock, since `Date.now()` steps backwards on an NTP correction.

One finding was left unfixed deliberately: there is no inbound rate limit on the pump. Only a same-origin local client can reach it, frames are capped at 1 KB, and `select!` still services the other arms, so it would be a counter with no demonstrated need.

## Third review round

Three lenses (security, testing, skeptical) on the round-2 HEAD, because that round changed security code substantially. The testing lens re-ran mutations rather than taking the previous round's claims on trust, which is what surfaced most of this.

**The round-2 ADD-pass guard introduced a denial-of-prompt, and my reasoning about it was wrong.** I had argued it was safe because additions arrive over the socket. That misses `resync` — which exists *because* the server dropped `prompt_added` frames and will never resend them, and where the 3s poll is stopped precisely because the socket is healthy. A removal of any unrelated nonce landing inside that reconcile's fetch window buried the dropped prompt until the server auto-denied it. A backgrounded tab draining a churn burst is both what causes the lag and what delivers such a removal, so it is the expected shape of the recovery rather than a contrived race. The guard is now keyed per nonce; restoring the shared-timestamp semantics fails the new cross-nonce test and nothing else.

**The `#4849 F2` non-vacuity pin matched its own signpost comment.** A comment added in round 2 spelled the very literals the guard searches for, so the pin passed even under the refactor it exists to catch. This is the self-match class `AGENTS.md` records as having shipped twice. De-quoted, and mutation-verified that the pin now fails when the literals are hoisted out of the region.

**The `X-Forwarded-Proto` check's claimed coverage was overstated.** The attack it describes arrives on the plain-HTTP vhost — which is exactly the listener likely to be the one missing that header. So the check narrows the attack rather than closing it. The doc now says so, and a one-shot warning fires when forwarding headers arrive without it, because an unobservable degraded gate is this PR's whole subject. Also: a present-but-blank header is now treated as absent (it would otherwise refuse every handshake on that deployment), the ambiguous-authority reject set gained `\`, space and tab, and the SSE sub-cap became a small absolute 16 rather than three quarters of the budget, with refusals logging the ceiling actually enforced.

**Test coverage**: 5 HIGH, 8 MEDIUM and 7 LOW gaps closed, each mutation-verified from MISSED to CAUGHT. The most consequential were a fixture defect — section 7 froze the clock across the fetch, so the two readings the whole section exists to distinguish were numerically identical — and the absence of any test for the commonest reconcile case (a card on screen *and* in the snapshot). On the Rust side, a three-frame slow-sink test now pins that the replay deadline bounds the whole phase rather than each message; the single-frame test could not distinguish the two designs.

Deliberately not fixed, both stated rather than quietly dropped: the pump has no inbound rate limit (reachable only by a same-origin local client, frames capped at 1 KB), and the Playwright workflow that now serves as the primary guard for this invariant is not a required check — a pre-existing structural gap this PR leans on more heavily, filed as #5275 rather than restructured here.

Closes #5213

[AI-assisted - Claude]


sanity:
## Browser verification (the #4945 lesson)

`shell_bridge.js` is the file that broke in #4945, where a SecurityError took out every locally-served web app while every `riverctl` check passed, because riverctl bypasses the served shell and the sandboxed iframe. So this was verified by driving real headless Chromium against a served app, not by unit tests alone.

**Controlled A/B.** Same contract app, same browser, same machine, 10 tabs opened one at a time. The only difference is the node binary: the running v0.2.123 node versus one built from this branch. The branch node ran against isolated copies of the contract store and webapp cache on a spare port, with `FREENET_WEBAPP_CACHE_DIR` pointed at the copy, so it could not touch the real node's state.

**Before — v0.2.123:**

```
tab 1-6  | doc=200   | permWS=0 | ok
tab 7    | doc=NEVER | permWS=0 | STALLED
tab 8    | doc=NEVER | permWS=0 | STALLED
tab 9    | doc=NEVER | permWS=0 | STALLED
tab 10   | doc=NEVER | permWS=0 | STALLED
```

**After — this branch:**

```
tab 1-10 | doc=200   | permWS=1 | ok     (every tab)
```

```
PASS  all 10 tabs loaded
PASS  every tab opened a permission WebSocket
PASS  no client SSE requests to /permission/events
PASS  no uncaught page errors
PASS  no console errors
```

`doc=NEVER` is the shell document request never receiving a response — a browser-queue stall rather than an application state.

One detail worth reading closely: the before run shows **exactly six** SSE requests, one per tab that got through. Tabs 7-10 never issued theirs at all, because they were already queued behind the six held-open streams. That is the mechanism visible directly in the numbers.

The last two checks are the #4945 class specifically: no uncaught page errors and no console errors from the bridge, on every tab.

## Test-integrity notes

Two things worth flagging for review, because both are cases where the tests initially looked fine and were not.

**The rejection tests were vacuous at first.** My first version drove the gates through `tower::ServiceExt::oneshot`. A WebSocket upgrade needs a connection hyper can hand off, and `oneshot` with an empty body has none, so axum's `WebSocketUpgrade` extractor returned **426 Upgrade Required before any gate ran**. Every rejection test would have passed a loose "not 200" assertion while proving nothing about security. This was caught only because `permission_ws_accepts_same_origin_loopback_handshake` — a deliberate control asserting the happy path reaches 101 — failed alongside them. All the WebSocket tests now drive real TCP, with the peer address injected via an `Extension` layer so a test can still present a LAN peer or none at all, and each asserts the *specific* status (403 for a gate, 503 for the cap) so a 426-style false pass cannot recur.

**Both gates were mutation-tested.** Deleting the origin gate fails `permission_ws_rejects_cross_origin_handshake` and nothing else. Deleting the loopback gate fails `permission_ws_rejects_lan_peer` and `permission_ws_rejects_missing_connect_info`, and nothing else. So the security tests catch the removal they claim to.

## Local gate

`cargo fmt --all --check`, `cargo clippy --locked -- -D warnings` (CI's exact invocation), `cargo test -p freenet --lib` for the permission and shell suites (89 passed), plus the asset Prettier/ESLint job and the Node asset tests — all green on the pushed commit.

Note the pre-existing clippy state: `cargo clippy --all-targets` reports 54 errors across 21 files on `main`, all in test code, because CI deliberately does not lint tests. Not touched here beyond the one in code I added.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #5269

### Summary
- **PR Title:** fix(server): carry permission events over a WebSocket, not SSE (#5213)
- **Type:** fix
- **CI Status:** green except `rule-review/warnings` (2 warnings, both since fixed — see below)
- **Linked Issues:** #5213
- **Review tier:** **Full** — touches a permission/origin-gating surface (auth) and adds an endpoint
- **Reviewers run:** `code-first`, `testing`, `skeptical`, `big-picture`, plus a fifth **WebSocket-security / rollout** lens standing in for the external model pass. Per `.claude/rules/pr-quality.md` and the global multi-model rule (changed 2026-07-26), Codex is opt-in only and was not attempted; the substitute lens is recorded here so a reader can see independent review happened and in what form.
- **HEAD SHA reviewed:** cd6ae02addc4d5aa19fad0a1dfacbfa6d37cea32

**Headline: two blocking findings.** The change does what it claims and the server half held up to hard adversarial probing — but it widens who can read the permission stream, and it leaves a path where a live security prompt is silently destroyed. Both are fixable without redesign.

---

### Code-First Analysis

**Independent understanding:** adds `GET /permission/events/ws` carrying the same `prompt_added` / `prompt_removed` / `resync` events in a `{event, data}` envelope; generalises `SSE_CONNECTIONS` into a shared `PERMISSION_SUBSCRIBERS` + `try_claim_subscriber_slot()`; keeps the SSE route registered; swaps the shell from `EventSource` to a self-managed WebSocket with 1s→30s backoff plus ±20% jitter.

**Alignment:** matches. Every checkable claim in the description holds — gate ordering (`permission_prompts.rs:915-940`), subscribe-then-snapshot preserved (`:947-960`), event payloads identical to SSE, SSE route still served (`:105`). No intent/implementation discrepancy. The findings below are consequences the description did not account for, not misdescriptions of the diff.

---

### Testing Assessment

| Unit | ✅ | Gate tests verified non-vacuous; each asserts a *specific* status (403/503), so a 426 from an unreached handler fails rather than satisfies |
| Integration | ⚠️ | No test proves a live socket holds a slot, or that disconnect releases one |
| Simulation | N/A | |
| E2E | ❌ | `crates/core/tests/playwright/` exists, is path-filtered to `crates/core/src/server/**` so it runs on this PR, and is the exact harness the manual repro used — but no spec was added |

**Regression test:** present but narrower than its own comment. The pins assert `!contains("new EventSource(")`; the stated invariant is "must not ride *any* held-open HTTP request". A streamed `fetch()` or long-poll reintroduces #5213 verbatim and passes both pins.

---

### Skeptical Findings

**Risk level: medium.** The server half is solid — I could not wedge the pump, leak a slot, miss an event through the subscribe/snapshot ordering, or bypass the gates, and the cancel-safety claim checks out arm by arm. The problems are on the client state machine.

|---|---|---|---|
| Stale `/permission/pending` response permanently hides a live prompt | **High** | `shell_bridge.js:1849-1851` | See Must Fix 2 |
| Cross-port localhost origin can read the stream | **High** | `permission_prompts.rs:921-928` | See Must Fix 1 |
| WS cap test lacks the quiesce loop the SSE cap test has | Medium | `permission_prompts.rs:3283-3291` | WS guards release in a *spawned task*, not synchronously like `GuardedStream`; `serial_test`'s lock releases before the runtime is torn down, so a prior test's pump can still be unwinding. Flaky by construction |
| Backoff resets on `onopen` | Medium | `shell_bridge.js:1987` | Accept-then-immediately-close (a proxy that accepts the upgrade then drops it) pins retries at ~1 Hz forever with no growth, each cycle firing two `/permission/pending` fetches. Same class of silent unbounded loop as #5213 itself |
| No fallback poll while the *first* handshake hangs | Medium | `shell_bridge.js:2006-2013` | If the upgrade hangs rather than fails — the common WS-through-proxy mode — the tab shows no prompts and runs no poll until the browser's own timeout |
| Send timeout can cause a spurious browser popup | Low-Med | `permission_prompts.rs:870` | Dropping a frozen tab's socket decrements `receiver_count`, which feeds `maybe_alert_no_tab` (`user_input.rs:422`) — a user with tabs open can get an unexpected window. SSE had no such mechanism |
| No `max_message_size` on the upgrade | Low | `permission_prompts.rs:962` | Defaults are 64 MiB message / 16 MiB frame, for a channel that discards all inbound |
| `onclose` side effects outside the `permSocket === sock` guard | Low | `shell_bridge.js:1999-2003` | Latent: a future `visibilitychange`/`online` trigger would let a stale socket's close orphan the live one |
| No per-iteration cap on the inbound drain arm | Low | `permission_prompts.rs:1012-1021` | CPU burn only; no starvation (`select!` without `biased` randomises) |

**`bug-prevention-patterns.md`: no pattern tripped.** The self-satisfying-pin pattern is in scope in spirit and the PR *satisfies* it — both pins are cross-file scrapes (assertion literals live in the test, scanned content is `shell_bridge.js`), so they cannot be satisfied by their own literals.

---

### Big Picture Assessment

**Goal alignment:** yes, and it is a real fix rather than a mitigation. Grepping every served asset confirms that after this change **no tab holds any long-lived HTTP request**; remaining hits are `new WebSocket` (pooled separately) and short `fetch` calls.

**Guard test — clean and strengthened.** Verified mechanically rather than by reading: assertion messages diffed between `HEAD~1` and `HEAD` give **3 reworded, 6 added, 0 removed**. The #3836 Notification guards are untouched. The #4849 F2 marker region was re-executed against the new file — markers bracket 23,684 bytes; `'prompt_added'`, `'prompt_removed'`, `showCard(`, `new WebSocket(` are inside; `innerHTML`, `Notification.requestPermission`, `new Notification(`, `visibilityState` are absent. The negative asserts are live.

**Scope creep: none.** The cap rename is *forced* by the change — two independent 64-slot caps would allow 128 live receivers during exactly the upgrade window this PR creates; leaving them separate would have been the defect. The JS helper extraction is load-bearing for the Node test.

**Anti-patterns:** none detected.

---

### Documentation

Stale in several security-relevant places (no `.claude/rules`, `AGENTS.md` or `docs/` content is affected — this is all in-file):

- `permission_prompts.rs:523-526` — the origin-check doc says *"EventSource is the wire that matters here."* This is the security-critical function and, per this PR's own analysis, the WebSocket handshake now matters **more**. Worst possible place to understate the threat model.
- `shell_bridge.js:1476-1489` — the overlay region header still describes SSE throughout; the correct explanation exists 380 lines lower.
- `permission_prompts.rs:5`, `:12` — module doc still says the shell subscribes to SSE and calls `/permission/pending` a fallback "for environments without `EventSource`".
- `config.rs:2539` — `--allowed-host` help enumerates the gated permission endpoints and omits the new one.
- `user_input.rs:213-215`, `path_handlers.rs:5003-5005`, `shell_bridge.js:41` — misc SSE references.

---

### Recommendations

#### Must Fix (Blocking)

1. **The WebSocket is reachable from any other port on the same loopback host; SSE effectively was not.** `is_origin_trusted` accepts *any* localhost origin on *any* port. On SSE that leniency never mattered: the trusted branch sets no `Access-Control-Allow-Origin` — deliberately, per the rationale at `permission_prompts.rs:283-290` — so a page at `http://localhost:3000` passed the server gate and had the body withheld **by the browser**. The browser was the enforcement. WebSockets have no CORS, so that backstop is gone, and any local page (dev server, notebook, another local app's UI, anything with an XSS on `127.0.0.1:*`) can now read raw nonces, delegate-authored prompt text, delegate key and caller identity — and can hold cap slots, which a CORS-torn-down EventSource could not.

   Impact is disclosure plus a social-engineering setup rather than silent auto-approval: `/permission/{nonce}/respond` uses the `Json` extractor, so a cross-origin POST needs a preflight that 405s. Both reviewers who found this independently rated it Should Fix on that basis; I am treating it as **blocking** because it is a security regression this PR introduces, and the fix is small and precise.

   **Fix:** at the permission-WS gate only, additionally require Origin authority == `Host` — the same equality `origin_matches_allowed_host` (`:655-673`) already applies for branch 3. The real client always satisfies it: `permSocketUrl` builds from `loc.host`. Apply at the gate, **not** inside `is_origin_trusted`, which is shared with `hosted_import` and `pending_prompts`. Add the missing `http://localhost:3000` test case — the current set is all cross-*site*, which is why the widening is invisible in CI.

2. **A stale `/permission/pending` response silently destroys a live prompt card, permanently.** `reconcileFromPending` (`shell_bridge.js:1835-1854`) hides every card absent from the fetched list; the fetch is async and nothing versions it, so any card added by the socket between dispatch and resolution is hidden and never re-shown. Reachable on an ordinary node restart: `onclose` fires poll F1 → socket reconnects → `onopen` calls `stopFallbackPoll` (which clears the *interval*, it cannot cancel in-flight F1) → prompt X arrives over the socket and renders → F1 resolves with the pre-restart list and hides X. The prompt is invisible with a healthy socket and no poll running, and auto-denies at the server timeout, with nothing surfaced.

   The race shape predates this PR, but the PR makes overlapping in-flight reconciles the *normal* case: `reconcileFromPending` is now reachable from four sites and every close/reopen cycle issues two. **Fix:** gate the hide on causal ordering, not recency — stamp `card._fnShownAt` in `showCard` and only hide when `_fnShownAt <= issuedAt`. A generation counter alone does not fix the step-4→5 ordering.

#### Should Fix (Important)

3. Copy the SSE cap test's quiesce loop (`:2174-2183`) into `permission_ws_honours_shared_subscriber_cap` — WS slots release asynchronously, so without it the test is flaky, and a flaky test here would mask a real cap regression.
4. Start the fallback poll at the top of `openPermSocket` and let `onopen` stop it, so a hanging first handshake is covered. This makes the poll the always-on safety net and the socket the accelerator — which is what the design comment already claims.
5. Reset the reconnect backoff only after the socket has been up some minimum (~10s), not on `onopen`.
6. Add a slot-lifecycle test: connect → count is baseline+1 → drop the client → count returns to baseline. Nothing currently proves the guard survives into `pump_permission_events`, nor that a disconnect releases the slot; a leak there wedges the channel at 64 for the node's lifetime.
7. Add a JS reconnect test with stubbed `WebSocket`/`setTimeout` (precedent: `shell_bridge_reload.test.mjs`). Today, deleting `permReconnectDelay = 1000` passes every test while making every tab take up to 30s to recover from every node restart. This also retires the weak `html.contains("schedulePermReconnect")` name pin, which passes if the body is gutted.
8. Add a Playwright spec opening N tabs and asserting each `page.goto` resolves. The harness exists, runs on this PR's paths, and its README says it exists precisely because Rust substring assertions let four regressions ship. This is the only test that would cover the actual bug class.
9. Fix the stale docs above, prioritising `permission_prompts.rs:523-526`.
10. Set `ws.max_message_size(...)`/`max_frame_size(...)` — one line, zero risk.
11. Bound the whole replay phase with one deadline: with `MAX_PENDING_PROMPTS = 32` and a 10s per-message timeout, a client that stops reading holds a slot ~320s, not the 10s the constant implies.
12. Rename `ws_envelope_matches_the_sse_event_contract` — it never reads the SSE path, so it proves self-consistency, not cross-transport agreement. `ws_envelope_wraps_data_unchanged` is honest.

#### Consider

13. File a tracker for SSE retirement and reference it at the route registration. The stated justification is also overstated: the 0.2.123 client registers `es.addEventListener('error', startFallbackPoll)`, so a 404 drops it to polling — removal costs pre-upgrade tabs latency, not correctness.
14. `console.warn` after N consecutive failed handshakes. A tab permanently on the poll is currently indistinguishable from a healthy one — the same silence that made #5213 hard to find.
15. Restate or raise the 64-slot cap. Its "generous headroom" rationale was calibrated against SSE, where the browser's own 6-per-origin limit made it unreachable; at ~255 WebSockets per profile it now binds.
16. Add an `AGENTS.md` rules row for `server/path_handlers/assets/`. This is the second incident in this one file (#4945, now #5213) that no Rust or `riverctl` test could catch, and the generalisable lesson — *a long-lived HTTP request per tab against the single shared local origin permanently consumes one of ~6 browser connections, and fails silently* — is recorded nowhere an agent will read it.
17. Rename the `serial_test` key from `sse_global_counter` to match `PERMISSION_SUBSCRIBERS`.
18. Note in the `perm-ws-decisions` comment that the block must stay nested inside `perm-overlay-flow` — the F2 anchor literals now all live in `permEventAction`, and hoisting those pure helpers to module scope is the obvious next refactor.

---

### Already addressed since the reviewed SHA

The `rule-review/warnings` CI check flagged two gaps, both now fixed with mutation-verified tests (`pump_permission_events` was made generic over its sink/stream to make them reachable):

- `permission_ws_pump_emits_resync_when_the_broadcast_laps_it` — drives the real pump with a receiver lapped by 3. Replacing the resync branch with a silent `return` fails it.
- `permission_ws_pump_gives_up_on_a_client_that_stops_reading` — drives the real pump with a sink that never accepts; asserts the pump returns *and* the slot is released. Removing the timeout fails it with `Elapsed`.

These close testing findings F6 and F7.

---

### Verdict

Two blocking findings (1 and 2), both introduced or materially worsened by this PR, both with small precise fixes. The server-side design is sound and survived hard probing; the gate ordering, slot accounting, cancel safety and #3836 boundary all check out. Work is on the client state machine and on the origin gate's newly-load-bearing role.

A fresh Full review must run on the updated HEAD before merge — findings 1, 2, 3, 5 and 7 all touch security or concurrency-adjacent behaviour, so the diff-of-the-diff shortcut does not apply.

[AI-assisted - Claude]

sanity:
Pausing here at a clean stopping point. Everything is committed and pushed; HEAD is `4d1233347`.

**State:** local gate green (fmt, clippy, 100 Rust tests, asset lint, 92 JS checks). CI on this exact SHA was 12 pass / 0 fail with 6 heavy jobs still running when I stopped watching; all six passed on the two previous SHAs, and `rule-review/warnings` is green so no override label is needed. **Not merged** — that is the remaining step, after confirming CI on the exact HEAD SHA.

**Five review rounds.** The #5213 fix itself has been stable since the first commit and passes in Chromium, Firefox and WebKit locally and in CI. Everything since has hardened the origin gate reviewers asked for, and fixed the tests around it. Rounds 3 and 4 each found a real defect in the previous round's fix, and round 5 found that the tests for both of those rounds were validating a harness *duplicate* rather than the shipped guard — four mutations, including a wholesale revert of the previous commit, all passed green. That is fixed: the real block is extracted behind markers and executed, and every guard mutation is now caught.

Three claims I made earlier and later retracted, since they are in the commit history: the send timeout was not actually mutation-verified (the test exited via a different deadline); the Playwright control only proved a socket was *created*, not accepted; and I argued the add-guard could not cause a denial-of-prompt, which was wrong on the `resync` path.

Three reviewer suggestions were declined with reasons recorded in the code — requiring `Origin` on the legacy SSE route would break the pre-upgrade tabs it exists to serve; clearing a pending reconnect strands a tab on polling after a hung upgrade; and an inbound pump rate limit is unreachable except by a same-origin local client.

Follow-ups filed rather than smuggled in: #5272 (retire the legacy SSE route) and #5275 (the Playwright suite is now the primary guard for several browser-only invariants but cannot block a merge).

Hand-off notes at `~/code/tmp/handoff-5269-permission-websocket.md`, with the full finding-by-finding record at `~/code/tmp/REVIEW-TODO-5269-detail.md`.

[AI-assisted - Claude]

Issue #5266, opened 2026-08-11
netcheck's own reporting cannot support the diagnosis it exists to enable


sanity:
## Summary

While using the nightly Netcheck run to diagnose live-network GET failures (13 consecutive nightly failures, 2026-07-30 through 2026-08-11), we found three defects in `crates/netcheck/` itself that undermine the diagnosis the tool exists to produce. None of these are about the underlying network failures — they're about netcheck's own reporting being unable to support conclusions drawn from it. Filing together since they were found in one pass over the same investigation and are small, independent, mechanical fixes.

## 1. Deterministic op ordering confounds contract age with position-in-run

`crates/netcheck/src/manifest.rs:66-79` (`retention_targets`) iterates a fixed-order `WINDOWS` const:

```rust
const WINDOWS: &[(&str, u64, u64)] = &[
    ("24h", 20 * 3600, 30 * 3600),
    ("48h", 44 * 3600, 54 * 3600),
    ("7d", 156 * 3600, 180 * 3600),
];
```

`crates/netcheck/src/scenarios/put_get.rs` then processes this run's own "0h" contracts first (lines 140-166), followed by the retention windows in the order `retention_targets` returned them (lines 169-193). The result is deterministic, every single run: 0h GETs are always issued first, 7d GETs are always issued last, with 24h/48h always in between in the same order.

This means "contract age" and "how late in the run's op sequence this GET was issued" are the same variable by construction — netcheck cannot statistically distinguish a genuine age/retention effect (e.g. an older, less-recently-touched contract being harder to find) from a session/connection effect (e.g. something about the ephemeral peer's state degrading the longer a run goes on). We observed exactly this ambiguity: failure rate rose sharply for the 7d bucket across 13 runs' worth of data, but per-run failure indices also skewed toward the back half of the op sequence in most runs, and the two explanations cannot be separated on the existing data.

**Fix:** shuffle or round-robin the retention windows, and interleave them with the 0h ops, each run — rather than a fixed 0h-then-24h-then-48h-then-7d order every time.

## 2. `latency_ms` is a hardcoded constant on failure, not a measurement

`crates/netcheck/src/scenarios/put_get.rs::push_get_report` (and the equivalent inline arm for PUT in `run()`):

```rust
Err(e) => report.push(OpReport {
    ...
    latency_ms: op_timeout.as_millis(),  // always the configured timeout, not actual elapsed time
    ...
}),
```

[...]

Issue #5256, opened 2026-08-09
fix(transport): four defects and a coverage gap in the streaming/fragment path


sanity:
Found during a source audit of the streaming/fragment path (investigating state corruption, freenet-core#5153). None of these explains that corruption — the fragment path was structurally excluded — but all four are real and were found en route. Filed together because they are one subsystem and one audit; split if that is easier to action.

## 1. `piped_stream.rs:259` — fragment-size off-by-one (latent)

```rust
const FRAGMENT_PAYLOAD_SIZE: usize = crate::transport::packet_data::MAX_DATA_SIZE - 40;
```

Every other site subtracts **41**, not 40 — see `peer_connection.rs:63`, whose rustdoc records the measured overhead as 41 bytes ("the extra byte vs. the original 40 comes from the Option discriminant of `metadata_bytes`"). This makes `total_fragments` wrong for `PipedStream`.

**Latent**: `PipedStream` has no production callers (only a doc reference at `peer_connection.rs:2095`). Worth fixing before it acquires one.

## 2. `outbound_stream.rs:190` — wire-vs-payload over-charge (#5054 class, second site)

```rust
let packet_size = stream_to_send.len().min(MAX_DATA_SIZE);
```

`packet_size` is computed *before* `available_payload` (`:296`), so the first fragment — which carries embedded metadata and therefore less payload — charges the congestion window and the token bucket a full 1130 bytes while carrying fewer.

Same class as **#5054** ("LOSS_PAUSE_MARGIN is sized in wire bytes but flight size is charged in payload bytes"), at a second location. Two independent instances suggests the wire-vs-payload distinction wants a type rather than care.

## 3. `streaming.rs:539` — `Stream` impl can end early and silently truncate

```rust
if buffer.is_complete() && buffer.get(next_idx).is_none() { Poll::Ready(None) }
```

`is_complete()` is `contiguous >= min_complete_fragments` (`streaming_buffer.rs:308`), which can be true *before* the metadata-overflow fragment arrives. A consumer polling at that moment gets `None` and the stream ends early — silent truncation, no error.

`assemble()` guards exactly this case at `streaming.rs:386-395`; the `Stream` impl does not. The two paths should agree.

## 4. `streaming_buffer.rs:388` — `take()` does not advance `consumed_frontier`

`take()` swaps the slot to null but never touches `consumed_frontier`, so a retransmission can refill a slot a reclaiming consumer has already passed.

[...]

Issue #5054, opened 2026-07-30
cwnd-wait stall: LOSS_PAUSE_MARGIN is sized in wire bytes but flight size is charged in payload bytes, capping the window at 53 fragments


sanity:
## Summary

A stream aborts after waiting 3s for congestion-window space. The modal failure sends exactly **53 fragments**, and 53 falls out of a **unit mismatch between two constants**: the loss-pause margin is sized in *wire* bytes while flight size is charged in *payload* bytes.

Data: one complete archived rotation, `logs-2026-07-30T13-15-16.598.jsonl`, 5.242 GB, 5,049,060 records, 0 parse failures, spanning ~28.9h (2026-07-29T11:41 → 2026-07-30T16:36 UTC). 77,258 `transfer_failed`, of which **n=73,046 (94.55%)** are the `cwnd wait timeout ... packets` variant. All figures below are exact counts over that file.

## First, a correction to the rate

An earlier reading of this data reported a 22.2% transfer failure rate. **That is wrong; the real rate is 1.66%.** The two events are sampled differently *downstream of the node*: `/etc/otelcol-contrib/config.yaml` routes `transfer_failed` to the unsampled `logs` pipeline (lines 198-208) and `transfer_completed` to `logs/transfer` (lines 222-230) through `probabilistic_sampler/transfer` at `sampling_percentage: 5`, peer-clustered on `peer_id`. freenet-core emits both identically (`transport/metrics.rs:817` and `:858` — same channel, same `EventPriority::Operational`, no emit-site sampling).

Measured directly on the 65 peers inside the sampled cohort, where numerator and denominator are both complete (sampling is peer-clustered, so a sampled peer keeps 100% of its completions):

- completions 267,629, failures 4,519 → **1.6605% aggregate**
- per-peer: min 0.000%, p25 0.14%, **p50 1.15%**, p75 2.02%, p95 6.08%, max 17.39%; twelve peers at literally 0%

Still ~2,670 real failures/hour fleet-wide. A real bug, not an emergency. **Any dashboard comparing these two events is currently misleading by ~20x** and that is worth fixing on its own.

## The mechanism (confirmed, not hypothesised)

[...]

PR #5270, opened 2026-08-11
fix(transport): stop poll_next ending a relayed stream one fragment early

sanity:
## Problem

`StreamingInboundStream::poll_next` ended a stream as soon as
`buffer.is_complete()` was true and the next fragment slot was empty.

But `is_complete()` is `contiguous >= min_complete_fragments` — the **base**
fragment count. When the sender embeds metadata in fragment #1 (#2757), that
fragment carries a reduced payload, so the sender emits `base + 1` fragments.
`is_complete()` therefore fires **one fragment early**, while the final
overflow fragment is still in flight.

`pipe_stream` — the relay forwarding path in `outbound_stream.rs` — drives this
`Stream`. So a relay stopped forwarding short of the `total_length_bytes` it had
itself advertised. The downstream peer then waited in `assemble()` for a fragment
that would never be sent, and failed with `StreamError::InactivityTimeout`:

```
stream assembly: no fragments received within inactivity timeout
```

which is verbatim the error string appearing in nightly netcheck runs.

**`assemble()` already had this guard** (it keeps waiting when `is_complete()` is
true but the assembled length is short of `total`) — added after the same
truncation caused blank pages in the River web app. `poll_next` never received
the equivalent. This PR closes that asymmetry.

Why this path is reached in practice: the GET **originator** uses `assemble()`,
but the GET **relay** pipes via `poll_next`
(`get/op_ctx_task.rs` → `pipe_stream` → `peer_connection.rs` →
`outbound_stream.rs`, `inbound_handle.stream()`). PUT relaying has the same
shape.

Precisely how often, corrected: **this needs two or more relay hops on the response
path, not one.** A single-relay GET response goes `relay_send_found` ->
`conn_manager.send_stream`, and the originator consumes it with `assemble()`, which was
already guarded. `pipe_stream` is reached only from the "Forward path (remote
upstream)" branch at `get/op_ctx_task.rs:4048`, which runs when a peer forwards an
inbound stream it itself received. An earlier revision of this description said "any GET
that is relayed at least once exercises it", which is wrong.

## Approach

Make the byte count the discriminator, exactly as `assemble()` does: only end the
stream when every advertised byte has actually been yielded.

```rust
if self.handle.buffer.is_complete()
    && self.handle.buffer.get(next_idx).is_none()
    && self.bytes_read >= self.handle.buffer.total_bytes()
```

The two paths consume the same buffer and must reach the same conclusion about
when a stream has ended; they now do. The existing behaviour is preserved for the
common case where fragment #1 has full payload capacity and the overflow slot is
legitimately never used.

## Testing

Adds `test_stream_waits_for_overflow_fragment`. It constructs the exact shape the
sender produces (fragment #1 with reduced payload, so `base + 1` fragments needed),
asserts the buffer reports `is_complete()` while `try_assemble()` still returns
`None`, then drives `poll_next` manually with a no-op waker so the assertion lands
on the precise poll inside that window. **No sleeps and no timing race.**

## Scope — please read before assuming this fixes the nightly

This removes **one contributing failure mode**. It will **not** by itself turn the
nightly netcheck green.

- It requires two conditions, not one: a payload-length condition **and** the
  relay's reader reaching the last fragment before it arrives.

  Calling the second a coin-flip race, as an earlier revision did, understates it in
  the other direction. Once the length condition holds, truncation is closer to the
  DEFAULT than to 50/50: `is_complete()` being true implies slots `1..base` are
  already filled, so the only slot `poll_next` can find empty in that state is exactly
  `base + 1` -- the last one. A reader that is not backlogged reaches it, and
  not-backlogged is the normal steady state for a relay whose downstream keeps up. The
  bug is *avoided* mainly when the relay is downstream-bottlenecked enough that the
  final fragment has already landed. So the expected rate is nearer "the payload-length
  fraction of >=2-hop relayed streams" than "that fraction times a small race
  probability" -- which argues this fix is worth somewhat more than the rest of this
  section claims, not less.
- An earlier revision of this description put the payload-length condition at
  "roughly 15% of payloads". That number is not supported by anything in the code
  and has been removed. The window is `m / FRAGMENT_PAYLOAD_SIZE` where `m` is the
  embedded metadata size, so with `FRAGMENT_PAYLOAD_SIZE = 1130` (`1200 - 1 - 12 -
  16 - 41`) 15% would imply ~161 bytes of metadata. A GET relay's
  `GetMsg::ResponseStreaming` is on the order of ~110 bytes, i.e. nearer 10%; PUT's
  `PutMsg::RequestStreaming` carries a `skip_list` that grows per hop, so it is
  larger and variable. Order of magnitude right, precise figure unfounded.
- It produces a ~5s failure signature — that is `STREAM_INACTIVITY_TIMEOUT`, the
  downstream peer's wait in `assemble()`. Treat it as a floor rather than a
  fingerprint: 5s is what several unrelated stream failures also cost, so timing
  alone does not attribute a failure to this defect. Of 38 observed netcheck
  failures it is *consistent with* the four at 6.5–7.6s. It does **not** explain the
  60s-multiple or 120.2s populations.

Related but explicitly **not** addressed here: #5054 (cwnd-wait stall) remains the
best-supported hypothesis for the bulk of the failures and is untouched by this
change, and #5266 records three defects in netcheck's own reporting that mean some
residual signal may be measurement artefact rather than network fault.

## Relationship to #5256

This is item 3 of #5256 ("four defects and a coverage gap in the streaming/fragment path"),
which names this defect at this line with this reasoning. Referenced as a **partial** fix, not
`Closes`: item 2 (`packet_size` wire-vs-payload over-charge) and item 4 (`take()` never advances
`consumed_frontier`) are still open. Items 1 and 5 have since been fixed on main independently.

## Review

Full-tier review (transport is a named high-risk surface here), four independent blind Claude
lenses: code-first, testing, skeptical, big-picture. No BLOCKER from any lens. Changes made:

- **CI was red and is now green.** The pre-existing `test_streaming_inbound_stream_basic` HUNG
  under this change — 240s timeout on all three nextest retries. It advertised
  `StreamHandle::new(id, 15)` while pushing a single 5-byte fragment. That data is
  self-inconsistent rather than a real scenario: senders set `total_length_bytes` exactly
  (`send_stream` from `stream_to_send.len()`, `pipe_stream` from `inbound_handle.total_bytes()`),
  and `assemble()` already rejected that same buffer, so the old `poll_next` was the path that
  disagreed with the rest of the module. It now advertises `payload.len()`, and its terminal
  await is bounded by a 5s timeout so a future regression reports itself instead of hanging
  until nextest's kill with no diagnosis.
- **A second test carried the same landmine.** `test_bytes_read_tracking` advertised 15 while
  pushing 14 bytes; it survives only because it polls once and never reaches the terminal poll.
  Fixed the same way, so the next person to add an end-of-stream assertion there does not
  rediscover the hang instead of the bug they were chasing.
- **The added comment overclaimed and has been scoped.** It said the two paths "must reach the
  same conclusion about when a stream has ended". The `next_idx > total_fragments` exit is still
  unguarded, so that is not universally true; the comment now says which exit it fixes and points
  at #5445 for the rest.
- **Two impact claims were wrong in opposite directions** and are corrected above:
  reachability needs two or more relay hops rather than any relayed GET, and the second
  condition is nearer the default outcome than a coin flip.
- **The RED-without-fix claim is now discharged.** The description flagged it as the authoring
  agent's unverified claim. Reverting the production hunk locally and re-running:
  `test_stream_waits_for_overflow_fragment` fails with *"stream ended early: forwarded 2825 of
  3390 bytes"*. Restoring the hunk turns it green. All 43 streaming tests pass.

Filed rather than folded in, both found by the `is_complete()` call-site audit this review ran:

- **#5440** — `PipedStream` computes `total_fragments` as the base count with no allowance for
  the overflow fragment, so it would declare completion early *and* reject the genuine final
  fragment. Currently dead code (`#![allow(dead_code)]`, "not yet wired into the main forwarding
  path"), so latent rather than live — but it is the third implementation of this predicate and
  the second to get it wrong.
- **#5445** — `pipe_stream` reports a successful transfer after forwarding fewer bytes than it
  advertised, because the loop's other exit is unguarded. Not reachable from a conforming sender.
  Deliberately not fixed here: it is a production behaviour change on a high-risk path and wants
  its own review rather than a late addition to this one.

`.claude/rules/bug-prevention-patterns.md` gains the generalisable lesson: the discriminator is
the byte total, never the fragment count, and when a predicate this load-bearing has three
implementations the duplication is the bug.

## Provenance

Authored by an agent that terminated on a quota limit before it could open this PR;
its work was committed and pushed intact rather than lost. The RED-without-fix
reproduction is the authoring agent's claim and is **flagged in the commit message as
not independently re-verified** — a reviewer should confirm the test fails with the
production hunk reverted.

Refs #5256

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/bug-prevention-patterns.md
+++ b/.claude/rules/bug-prevention-patterns.md
@@ -158,6 +158,52 @@ arms, the INFO-level checks on `MARKER_CHECK_COMPLETE` and
 canary still runs, and still runs before `--draft=false` — is pinned by
 `scripts/release_canary_wiring_test.sh`.
 
+## A completion predicate that is a proxy for the real termination condition
+
+**A cheap proxy that *usually* agrees with the real condition reads as
+equivalent, gets copied to each new consumer, and diverges in the one
+case nobody tests.** freenet-core has three implementations of "is this
+stream finished" over the same fragment buffer, and two of them shipped
+the same off-by-one.
+
+The proxy here is a FRAGMENT COUNT; the real condition is
+`bytes_delivered >= advertised_bytes`. Two properties make this genre
+expensive out of proportion to the bug:
+
+- **The failure is silent at the site that causes it.** The truncating
+  peer reports success. The victim is the *downstream* peer, which dies
+  on an inactivity timeout naming neither the cause nor the culprit — so
+  the error surfaces in a different process from the defect, and the
+  telemetry that would prompt an audit says everything was fine (#5445).
+- **Each copy looks locally reasonable**, so fixing one consumer leaves
+  the others and nothing flags the asymmetry.
+
+The buffer allocates `base + 1` fragment slots, where
+`base = ceil(total_bytes / FRAGMENT_PAYLOAD_SIZE)`, because embedding
+metadata in fragment #1 (#2757) reduces that fragment's payload and makes
+the sender emit one extra fragment. `is_complete()` is
+`contiguous >= base` — the BASE count — so it goes true **one fragment
+early** on exactly those streams.
+
+| Consumer | State |
+|---|---|
+| `streaming_buffer::assemble()` | Correct: discriminates on assembled length vs `total_size`, keeps waiting when `is_complete()` is true but bytes are short. |
+| `StreamingInboundStream::poll_next` | Was wrong; fixed in #5270. Ended on `is_complete()` alone, so `pipe_stream` forwarded short and the downstream peer sat in `assemble()` until its 5 s inactivity timeout. |
+| `PipedStream` | Still wrong, currently dead code — #5440. Computes `total_fragments` as the base count, so it declares completion early AND then rejects the genuine final fragment as out of range. |
+
+**The rule:** the discriminator is the BYTE total, never the fragment
+count. A fragment count is an estimate derived from an assumption about
+payload sizes; `total_bytes` is what the sender actually advertised.
+
+**And the transferable half:** when a predicate this load-bearing has more
+than one implementation, the bug is the duplication, not any one copy.
+Before adding a fourth, ask whether it can call the existing one. When
+auditing a fix to one, grep for the others — #5440 was found only because
+the #5270 review enumerated every `is_complete()` call site, and it had
+been wrong since it was written. A truncation that reports success is
+also invisible in telemetry (#5445), so nothing prompts the audit on its
+own.
+
 ## Cross-test interference through process-global state: CI cannot see it
 
 A guard whose two inputs rot from one cause. Every CI job runs `cargo
@@ -228,6 +274,49 @@ hand-rolling a `split_once`. It:
 **Prefer a cross-file scrape.** A pin that lives in `auto_update.rs` and
 scrapes `update.rs` cannot be satisfied by its own assertion literal at
 all — a structural guarantee rather than a check someone must remember.
+
+### The variant `fn_body()` does not catch: the anchor survives the deletion
+
+`fn_body()` fixes a *moved* anchor. It does nothing about an anchor that
+is still exactly where the pin expects because the call was **commented
+out**. `// interleave(&mut ops, seed);` contains the string
+`interleave(&mut ops, seed);`, so `split_once` finds it, the region
+splits where it always did, the before/after assertions still hold, and
+the pin stays green over a shuffle that no longer runs. Deleting the line
+outright fails loudly; disabling it does not — and disabling it is what a
+person actually does while debugging, which is precisely when the pin is
+the only thing left watching.
+
+Found in review of #5271, where the PR's own new pin AND the pre-existing
+one it was modelled on both survived commenting out the call they guard.
+Verified by doing it: both stayed green before the fix, both go red after.
+
+**The rule:** a pin that guards a *call* must require the call at
+statement position — the call text preceded on its line by nothing but
+whitespace — not merely present in the region. A few lines:
+
+```rust
+fn split_at_call<'a>(body: &'a str, call: &str) -> (&'a str, &'a str) {
+    let at = body
+        .match_indices(call)
+        .find(|(i, _)| {
+            let line_start = body[..*i].rfind('\n').map_or(0, |n| n + 1);
+            body[line_start..*i].trim().is_empty()
+        })
+        .map(|(i, _)| i)
+        .unwrap_or_else(|| panic!("`{call}` is not called at statement position"));
+    (&body[..at], &body[at + call.len()..])
+}
+```
+
+**The generalisation, which is the part worth carrying:** a source-scrape
+pin asserts that *text exists*, while the property it is standing in for
+is that *code runs*. Every gap between those two is a way for the pin to
+pass vacuously — a moved anchor, a self-matched literal, a commented-out
+call, a call moved inside a branch that is never taken. When writing one,
+ask what the cheapest edit is that keeps the text and removes the
+behaviour, then make that edit fail. **And test the pin by performing that
+edit**, because a pin nobody mutated is a pin nobody has evidence about.
 
 ### The same class without `include_str!`: an expectation stored as a copy
 
--- a/crates/core/src/transport/peer_connection.rs
+++ b/crates/core/src/transport/peer_connection.rs
@@ -1602,11 +1602,26 @@ impl<S: super::Socket, T: TimeSource> PeerConnection<S, T> {
     /// ```ignore
     /// if let Some(handle) = peer_conn.recv_stream_handle(stream_id) {
     ///     let mut stream = handle.stream();
-    ///     while let Some(chunk) = stream.next().await {
+    ///     loop {
+    ///         // Bound the wait. See the note below on why `stream.next()`
+    ///         // must never be awaited unbounded.
+    ///         let next = tokio::select! {
+    ///             r = stream.next() => r,
+    ///             _ = time_source.sleep(STREAM_INACTIVITY_TIMEOUT) => break,
+    ///         };
+    ///         let Some(chunk) = next else { break };
     ///         process_chunk(chunk?);
     ///     }
     /// }
     /// ```
+    ///
+    /// **Always bound the wait.** `StreamingInboundStream::poll_next` yields
+    /// `Pending` until every advertised byte has been delivered, so a peer that
+    /// advertises a `total_length_bytes` it never sends, or a lost final
+    /// fragment, leaves an unbounded `stream.next().await` hanging forever. The
+    /// one production consumer, `pipe_stream`, wraps it in exactly this
+    /// `select!` against `STREAM_INACTIVITY_TIMEOUT` and reports a diagnostic
+    /// failure; anything new must do the same.
     #[allow(dead_code)]
     pub(crate) fn recv_stream_handle(
         &self,
--- a/crates/core/src/transport/peer_connection/streaming.rs
+++ b/crates/core/src/transport/peer_connection/streaming.rs
@@ -1860,7 +1860,9 @@ mod tests {
                     assert_eq!(data.len(), expected);
                     yielded += data.len();
                 }
-                other => panic!("expected fragment of {expected} bytes, got {other:?}"),
+                other @ (Poll::Ready(_) | Poll::Pending) => {
+                    panic!("expected fragment of {expected} bytes, got {other:?}")
+                }
             }
         }
 
@@ -1889,7 +1891,9 @@ mod tests {
                 assert_eq!(data.len(), overflow_size);
                 yielded += data.len();
             }
-            other => panic!("expected the overflow fragment, got {other:?}"),
+            other @ (Poll::Ready(_) | Poll::Pending) => {
+                panic!("expected the overflow fragment, got {other:?}")
+            }
         }
 
         assert_eq!(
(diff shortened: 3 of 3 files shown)


sanity:
## Revert-and-run check: independently verified

The PR body flagged the RED-without-fix reproduction as the authoring agent's
claim, not independently re-verified. I have now run it.

**With the fix (`6ad28417a`):** 16/16 pass in
`transport::peer_connection::streaming::tests`.

**With the production hunk reverted** (dropping only
`&& self.bytes_read >= self.handle.buffer.total_bytes()`, leaving the tests
untouched), exactly one test fails:

```
test test_stream_waits_for_overflow_fragment ... FAILED

panicked at streaming.rs:1838:
stream ended early: forwarded 2825 of 3390 bytes. `is_complete()` fires at
the base fragment count, so the overflow fragment was still in flight;
ending here truncates the relayed stream and strands the downstream peer in
assemble() until its inactivity timeout.
```

Two things worth noting:

- The failure is **specific**: 15 of 16 still pass under the revert, so the
  test is not a blanket assertion that would fail on any change.
- `test_stream_ends_when_overflow_slot_unused` **passes in both directions**,
  which is what you want — it guards the opposite case (the overflow slot
  legitimately unused), so the fix cannot be "make poll_next never terminate".

Also pushed `6ad28417a` to fix the `Fmt` check, which was failing on the array
literal in the new test and had skipped Unit & Integration, Simulation, and
the rest of the CI matrix.

[AI-assisted - Claude]

sanity:
## Full-tier review — 4 independent blind lenses

Risk tier: **Full**, not negotiable down. Transport/NAT is a named high-risk surface in this repo
with production incident history, and "the diff is only three lines" is not a reason to review it
as though it were — those three lines sit on the forwarding path of every relayed GET and PUT.

Lenses run, each blind to the others, each reading the checked-out code rather than the diff text:
**code-first**, **testing**, **skeptical/adversarial**, **big-picture**. No external (Codex) pass;
per this repo's `pr-quality.md` the external pass is opt-in, and Claude lenses were substituted.
Every `file:line` repeated below was re-verified by hand first.

**No BLOCKER from any lens.** The fix is correct and its premise checks out: `assemble()` already
discriminated on the byte total (returning `None` when `is_complete()` is true but the assembled
length is short), so `poll_next` genuinely was the path that disagreed with the rest of the module.

### CI was red. It is fixed, and the cause was not the production change

`test_streaming_inbound_stream_basic` **hung** — 240s timeout on all three nextest retries. It is a
pre-existing test the PR does not touch, and it advertised `StreamHandle::new(id, 15)` while
pushing a single 5-byte fragment, so the new `bytes_read >= total_bytes()` conjunct left it
`Pending` forever.

That data is self-inconsistent rather than a real scenario, and I checked rather than assumed:
senders set `total_length_bytes` exactly (`send_stream` from `stream_to_send.len()`, `pipe_stream`
forwarding `inbound_handle.total_bytes()`), and `assemble()` already rejected that same buffer. The
`15` dates from the original streaming PR #2443, where the comment says "Push all fragments" above
a single push. It now advertises `payload.len()`, and its terminal await is bounded by a 5s timeout
so a future regression reports itself rather than hanging until nextest's kill with no diagnosis.

Two lenses were asked independently whether that counts as weakening a test; both said no, and the
evidence agrees: **with the production hunk reverted, the corrected test still passes.** It is a
correction, not a change that masks the bug.

A second test carried the same landmine — `test_bytes_read_tracking` advertised 15 while pushing 14
— surviving only because it polls once and never reaches the terminal poll. Fixed the same way.

### The unverified claim is now discharged

The description flagged the RED-without-fix reproduction as the authoring agent's claim, not
independently checked. I reverted the production hunk locally and rebuilt:

```
test transport::peer_connection::streaming::tests::test_stream_waits_for_overflow_fragment ... FAILED
stream ended early: forwarded 2825 of 3390 bytes.
```

Restoring the hunk turns it green; all 43 streaming tests pass.

### Also changed under review

- **The added comment overclaimed.** It said the two paths "must reach the same conclusion about
  when a stream has ended". The `next_idx > total_fragments` exit is still unguarded, so on a short
  stream `assemble()` returns `None` (a failure) while `poll_next` returns `Ready(None)` (a
  success). The comment now states which exit it fixes and points at the issue for the rest. A
  confidently wrong comment on this path is worse than none.
- **The description's "roughly 15% of payloads" is unfounded** and has been removed. Deriving it
  from the constants, 15% implies ~161 bytes of embedded metadata; a GET relay's
  `ResponseStreaming` is nearer ~110 bytes, so ~10%, and PUT's carries a per-hop `skip_list` that
  makes it variable. Order of magnitude right, precise figure unsupported.
- **The ~5s signature is a floor, not a fingerprint.** It is `STREAM_INACTIVITY_TIMEOUT`, which
  several unrelated stream failures also cost, so timing alone cannot attribute a failure here. The
  description now says so instead of implying the 6.5–7.6s cluster is diagnostic.
- **#5256 is now referenced.** This is item 3 of that open issue, at that line, with that
  reasoning — as a partial fix, since items 2 and 4 remain open (items 1 and 5 have since been
  fixed on main independently). Without the reference the issue keeps carrying a fixed item and
  the next agent redoes this work.

### Found by the `is_complete()` call-site audit, filed rather than folded in

- **#5440 — `PipedStream` has the same off-by-one and is the third implementation of this
  predicate.** It computes `total_fragments` as the base count with no allowance for the overflow
  fragment, so it would declare completion early *and then reject the genuine final fragment as out
  of range*. Currently dead code (`#![allow(dead_code)]`; its module doc says "not yet wired into
  the main forwarding path"), so latent, not live — but it has been wrong since it was written and
  was found only because this review enumerated every call site.
- **#5445 — `pipe_stream` reports a successful transfer after forwarding fewer bytes than it
  advertised.** Not reachable from a conforming sender. Deliberately not fixed here: it is a
  production behaviour change on a high-risk path, and adding an unreviewed one to a transport PR
  at the end of a review is exactly what this process exists to prevent. It is also part of why
  the original truncation was hard to find — the relay's telemetry said it succeeded.

### Verified clean

No test removed, ignored, or weakened. The commit message does carry the "not independently
re-verified" flag the description claims. No TODOs, debug prints, or half-applied renames. Blast
radius checked: the only production consumer of the `Stream` impl is `pipe_stream`, whose
`stream.next()` is already inside a `select!` against a 5s inactivity timeout with a diagnostic
`emit_transfer_failed` — so the new `Poll::Pending` cannot hang a relay indefinitely; worst case it
converts a silent truncation into a diagnosed 5s pipe stall. `stream_with_reclaim` has no
production caller. The non-relayed originator path uses `assemble()` and is untouched. The fix also
cannot read one fragment too *many*: senders fragment `stream_to_send` exactly, so the payload sum
equals `total` exactly and `bytes_read` cannot overshoot.

`.claude/rules/bug-prevention-patterns.md` gains the generalisable lesson: the discriminator is the
byte total, never the fragment count, and when a predicate this load-bearing has three
implementations, the duplication is the bug.

**Rebase note:** this branch was BEHIND and has been rebased onto current main. The rebase was a
clean fast-forward with no conflicts, and content parity was verified by diffing the pre- and
post-rebase trees rather than by checking for leftover conflict markers. CI and this review are
both on the current head.

[AI-assisted - Claude]

sanity:
## Follow-up after the review comment above

Two more changes landed since I posted, plus the disposition of the remaining test-lens findings.

**CI went red on Clippy after the rebase, and the cause is worth naming.** Three
`clippy::wildcard_enum_match_arm` errors on `other =>` arms in this PR's own tests. Neither the
code nor the lint changed: `wildcard_enum_match_arm = "warn"` has been in `crates/core/Cargo.toml`
since #3194 in February, and CI's `-D warnings` has always promoted it. What changed is the gate's
*scope* — **#5292, "ci(clippy): bring cfg(test) and bench targets under the clippy gate"**, landed
on main during the twelve days this PR sat unreviewed, and test code was previously outside the
clippy gate entirely.

The original run on this branch did show Clippy passing, and that record was not wrong. It was a
statement about a *(commit, gate)* pair, and the rebase re-paired the same commit with a wider
gate. Fixed as `other @ (Poll::Ready(_) | Poll::Pending)`. A fourth `other` arm is deliberately
left alone — both `Poll` variants are already matched explicitly there, so the wildcard covers only
the inner `Option` and clippy does not flag it; CI reported exactly three.

**The doc example on `recv_stream_handle` showed a pattern that now hangs.** It had
`while let Some(chunk) = stream.next().await` with no bound. Since `poll_next` yields `Pending`
until every advertised byte arrives, a peer advertising a `total_length_bytes` it never sends —
or a lost final fragment — leaves that await hanging forever. `pipe_stream` already wraps it in a
`select!` against `STREAM_INACTIVITY_TIMEOUT`; the example now shows that shape and says why, so
the next consumer copies the bounded version rather than the one that reads more naturally.

### Remaining test-lens findings, dismissed with reasons

- **Dropping either of the other two conjuncts is an uncaught mutation.** Dropping `is_complete()`
  is behaviourally equivalent — `bytes_read >= total_bytes` already implies enough fragments — so
  no test *can* catch it, which is the same observation as "that conjunct is nearly redundant".
  Dropping `get(next_idx).is_none()` only matters once every advertised byte is already delivered,
  so what it would discard is beyond `total_bytes`. Neither is worth a guard.
- **The short-final-fragment shape is untested by the new tests**, and one assertion exercises the
  `next_idx > total_fragments` guard rather than the changed predicate. Both are test-precision
  points on a three-line fix that now has a regression test verified RED by reverting the
  production hunk. Adding more shapes here is churn against a well-covered predicate.
- **No test drives `pipe_stream` end to end over a multi-fragment stream.** Real, and the most
  valuable missing coverage — but it belongs with #5445, where the guard it would exercise lives.
  Recorded there rather than left implicit.

Local verification on the current head: `cargo fmt --all --check` clean, `cargo clippy -p freenet
--all-targets` clean, 43/43 streaming tests pass.

[AI-assisted - Claude]

PR #5271, opened 2026-08-12
fix(netcheck): make the report able to support the diagnosis it exists for

### LINKED ISSUE #5266: netcheck's own reporting cannot support the diagnosis it exists to enable

sanity:
## Problem

`crates/netcheck` is the nightly synthetic check against the live network. It has been failing every night since 2026-07-30, and while using its report to diagnose those failures we found three defects in the harness's own reporting. They are not network bugs — they are the reason the report cannot support conclusions drawn from it.

**1. Deterministic op ordering confounded contract age with position-in-run.** `manifest.rs::retention_targets` walks a fixed `WINDOWS` const, and `put_get.rs` processed this run's 0h contracts first and then the windows in that order. Every run, deterministically: 0h GETs first, 7d GETs last. Observed failures cluster toward the end of each run's op sequence, so "old contracts fail" and "later ops fail" were the *same variable* and could not be told apart — which is precisely the question the harness is used to answer.

**2. `latency_ms` was a constant on failure.** Every failed op reported exactly the configured `op_timeout` (120000 ms), including ops whose `error` field shows the node returned a specific terminal error well before the client's deadline. This is worse than an absent field: it looks like a measurement. It made "the client gave up waiting" and "the server failed fast" indistinguishable from the field designed to draw that distinction.

**3. The error arm was not key-filtered.** `client.rs` correctly skips *successful* responses carrying another contract's key, but the `Err` arm bailed on the first error seen regardless of provenance. A late error from an already-timed-out, already-reported op got attributed to whatever op was next in flight, with the wrong contract's key baked into the reported error text. Confirmed in GH Actions runs 31073282100 (2026-08-06) and 31148314225 (2026-08-07), where a `7d` FAIL line carries the key of the *preceding* FAIL line. Scope of the *defect*, precisely: it corrupts error **text** only. It does not affect `ok`/fail classification or op counts by size/age/label. **The fix does** — see "What this changes about the nightly's numbers" below. An earlier revision of this description carried the defect's scope sentence over to the fix, which was wrong.

## Approach

**Ordering.** The GET phase now builds one `Vec<GetOp>` covering this run's contracts and every retention window (each carrying how to verify what comes back — byte-identity for this run's own, recorded blake3 for previous runs'), then shuffles it before issuing anything.

Shuffle rather than round-robin, because round-robin over windows of unequal length still leaves systematic structure (the longest window trails the tail of every run). The shuffle is seeded from `blake3(run_id)` and the seed is logged next to the order it produced: randomness nobody can replay would trade one diagnostic problem for another, and a run needs to be reproducible from its own report.

**Latency.** `client::put`/`client::get` now return an `Attempt<T>` — the outcome, the measured elapsed time, and the count of errors skipped by the key filter, travelling together — measured on both arms, and both report sites use it. Returning the duration outside the `Result` is deliberate: it removes the code path where a constant could be substituted for a measurement, rather than relying on a future editor remembering to measure the error path.

**Key filter.** The recv-loop match body is extracted into `classify_put` / `classify_get`, returning `Done` / `Ignore` / `Fatal`. Errors that name a contract (`ContractError::{Get,Put,Update,Subscribe,...}`, `ErrorKind::IncorrectState`) are now filtered by key exactly as responses are.

Structured key fields alone would not have made the filter fire in production, though. The shape a terminal *network* failure actually arrives in is `ErrorKind::OperationError { cause }`, which has no key field and writes the id into the prose. So the filter also extracts contract ids from free text, guarded so an English word cannot become an id: a token must be 32..=44 base58 characters **and** round-trip (decode then re-encode must reproduce it exactly, since `from_base58` zero-pads short input). Anything failing those guards is simply not found, which leaves the error unattributed and therefore fatal — the conservative direction. Errors that name **no** contract keep the old conservative behaviour and still fail the waiting op — silently ignoring a disconnect would hang the op until its deadline. Extracting the classifier also makes the decision testable without a live node.

## Testing

Every fix has a test that was verified RED by mutating the fix away, not merely observed green:

| Mutation applied | Tests that went RED |
|---|---|
| `interleave` made a no-op | `the_get_order_is_not_the_fixed_age_sequence` ("must not issue 0h, then 24h, then 48h, then 7d"), `the_oldest_contracts_are_not_always_last` (`the last GET of a run was always the {"7d"} bucket across 20 runs`), `different_runs_get_different_orders` |
| `push_get_report` hardcodes 120 s on failure | `a_failed_get_reports_the_time_it_actually_took` (left: 120000, right: 37) |
| `client::get` returns `timeout` on the error arm | `a_fast_failure_reports_the_time_it_took_not_the_op_timeout` |
| `classify_get`'s `Err` arm loses its key check | `an_error_naming_another_contract_is_ignored_not_attributed`, `a_late_error_during_a_get_does_not_fail_that_get` — both fail with the wrong contract's key in the message, reproducing the exact log signature from runs 31073282100 / 31148314225 |
| `classify_put`'s `Err` arm loses its key check | `put_errors_are_key_filtered_the_same_way` |

Two of the key-filter tests run end to end against an in-process fake node (a localhost websocket speaking the same bincode `HostResult` frames the real node does), so the race is exercised through `client::get`'s actual loop: a stale error for another contract arrives inside the wait window, ahead of this op's own response, and the GET still succeeds.

The ordering tests also pin the properties a shuffle could quietly break — `interleaving_keeps_every_op_exactly_once` (no dropped or duplicated ops) and `the_order_is_reproducible_from_the_run_id`. Neither goes red under the no-op mutation, correctly: they guard the fix, they do not detect the bug.

27 tests pass; `cargo fmt --all --check` and `cargo clippy -p netcheck --all-targets -- -D warnings` are clean. No existing test was changed or removed.

## What this changes about the nightly's numbers

Read this before interpreting the first post-merge run.

The key filter changes `ok` classification, not just error text. Where a stale error from an
already-timed-out op previously killed the *next* op too, that op now keeps waiting and can
succeed. `a_late_error_during_a_get_does_not_fail_that_get` exists to assert exactly that.

So **the failed-op count can drop, and that drop is not a network improvement.** It is main
having double-counted a single genuine failure as two: the op that actually timed out, plus
the innocent op that inherited its error. The new count is the honest one. On a check that
has been red every night since 2026-07-30, a number moving in the good direction for a
reporting reason has to be declared, or it will be read as a recovery.

Second-order effect: an op that a stale error used to kill immediately now runs to its full
`--op-timeout-secs` (120 s default). Runs on a bad night get slower and will sit nearer the
workflow's 45-minute ceiling, which was already sized for 4 PUTs plus up to 16 GETs.

## Not covered

The issue's fix note for defect 3 mentions correlating on a transaction id as an alternative. The client API surfaces no tx id on errors, so this filters on the contract key. Errors that name no contract at all (`ChannelClosed`, `Disconnect`, `FailedOperation`, `NodeUnavailable`) remain unattributable and still fail whichever op is waiting — that residue is unchanged and cannot be closed without a protocol change.

`errors_ignored` counts only what the key filter skipped inside an op's own wait window. Errors arriving in the gap *between* ops are discarded by `drain_stray_responses` before the window opens; they are logged to stderr but not counted, because an unfiltered pre-op drain is a different event from a filtering decision an op made. The field's rustdoc says so, so a zero is not read as "no stale error existed anywhere near this op".

**`seq` and `errors_ignored` do not reach the telemetry dashboard yet.** `netcheck emit` publishes both, but the ingest writes a fixed column list (`insert_check_op`) that has no column for either, so it drops them without erroring. The jsonl artifact carries them, and that is where run order and skipped-error counts have to be read until the `check_ops` table gains the columns. Worth stating plainly because `seq` exists precisely so the shuffle would not destroy ordering information: in the jsonl it does its job, in the DB the ordering that used to be recoverable from the fixed `age` sequence is now recoverable from nothing. Noted in the README too, and filed as freenet/freenet-telemetry-dashboard#21.

None of this explains the live-network GET failures netcheck has been reporting nightly. That investigation is separate; this only makes the report trustworthy enough to run it.

## Review

Full-tier review (transport/diagnostics surface), four independent blind Claude lenses:
code-first, testing, skeptical, big-picture. Findings addressed in-branch:

- The description denied an `ok`-classification change the PR's own test pins — corrected
  above, no code change.
- `CONTRACT_ID_B58_LEN`'s upper bound was unpinned: every id in the tests encodes to 43
  base58 characters, so narrowing the range to `32..=43` left the whole suite green while
  silently dropping every id at or above 58^43 (~6% of the id space) out of attribution.
  Added `id_of(255)` — the all-ones id, 44 characters — as a positive case.
- The OTLP record hand-mirrors `seq` and `errors_ignored` from `OpReport` into `CheckOp`, and
  substituting a literal for either was invisible to every test. Added a source-scrape pin in
  `emit.rs`, the remedy `.claude/rules/bug-prevention-patterns.md` prescribes for
  manually-mirrored telemetry fields, with the same anti-vacuity guard the crate's `fn_body`
  helper uses.
- `errors_ignored`'s rustdoc overclaimed what a zero means, given `drain_stray_responses`
  discards pre-op errors uncounted. Tightened to state the window it actually measures.
- README documents `seq` and `errors_ignored`, no longer overstates replayability from the
  seed alone, and records the dashboard-ingest gap.
- A misleading assertion comment (a "one character longer" token that is in fact inside the
  length bound and rejected by the round-trip) corrected.

Closes #5266

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4231,6 +4231,7 @@ dependencies = [
  "futures 0.3.32",
  "hex",
  "rand 0.9.4",
+ "rand_chacha 0.9.0",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4228,6 +4228,7 @@ dependencies = [
  "chrono",
  "clap",
  "freenet-stdlib 0.8.5",
+ "futures 0.3.32",
  "hex",
  "rand 0.9.4",
  "reqwest 0.12.28",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ chrono = { version = "0.4", default-features = true }
 
 # Random
 rand = "0.9"
+rand_chacha = "0.9"
 
 # Observability
 opentelemetry = "0.32"
--- a/crates/netcheck/Cargo.toml
+++ b/crates/netcheck/Cargo.toml
@@ -14,6 +14,9 @@ clap = { workspace = true, features = ["derive"] }
 freenet-stdlib = { workspace = true, features = ["net"] }
 hex = { workspace = true }
 rand = { workspace = true }
+# Version-stable RNG for the run-order shuffle: StdRng's algorithm is
+# explicitly not reproducible across rand releases.
+rand_chacha = { workspace = true }
 reqwest = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/crates/netcheck/Cargo.toml
+++ b/crates/netcheck/Cargo.toml
@@ -21,3 +21,8 @@ tempfile = { workspace = true }
 tokio = { workspace = true }
 tokio-tungstenite = { workspace = true }
 x25519-dalek = { workspace = true }
+
+[dev-dependencies]
+# Driving the fake node in `client`'s tests: Sink/Stream extensions for the
+# server half of the websocket.
+futures = { workspace = true }
--- a/crates/netcheck/README.md
+++ b/crates/netcheck/README.md
@@ -20,7 +20,11 @@ Each `put-get` run:
    contract is" and "how late in the run the GET was issued" were the same
    variable, so an age effect could not be told apart from a
    within-run session effect. The order is shuffled from a seed derived
-   from the run id, and that seed is logged, so a run stays reproducible.
+   from the run id, and that seed is logged next to the order it produced,
+   so a run's order can be read back from its own report. (The seed alone
+   does not regenerate it: the permutation also depends on how many ops
+   the run had, which varies with which retention windows the manifest
+   had populated.)
 4. Prints one JSON line per operation on stdout and exits non-zero if
    anything failed. No retries by design: an operation that only succeeds
    on retry is the regression netcheck exists to surface.
@@ -69,7 +73,7 @@ One `"event":"run"` line with the conditions of the run, then one
  "freenet_version":"Freenet version: 0.2.106 (c707af0f786e)",
  "pinned_gateways":["100.27.151.80:31337,c28123df…"],
  "ephemeral_peers":["100.27.151.80:31337"]}
-{"event":"op","op":"get","age":"24h","label":"20260725-030000/small-0",…,"ok":true,"latency_ms":412}
+{"event":"op","seq":7,"op":"get","age":"24h","label":"20260725-030000/small-0",…,"ok":true,"latency_ms":412,"errors_ignored":0}
 ```
 
 `freenet_version` is what separates "the network broke" from "the release
@@ -81,6 +85,31 @@ long it actually took, not the configured `--op-timeout-secs`: reporting
 the deadline for every failure hid the difference between a fast terminal
 error from the node and the client giving up waiting, which is the
 distinction the field exists to draw.
+
+`seq` is the operation's 0-based position in the order the run actually
+executed. It has to be recorded because nothing else preserves it: every
+record of a run is published with the same timestamp, and line order is
+lost once the report is parsed. Before the GET order was shuffled,
+position could be recovered from `age`; now it cannot, so without `seq`
+the shuffle would destroy the very information it was introduced to
+disentangle.
+
+`errors_ignored` counts incoming errors this operation attributed to
+another contract and skipped. A server-side error for an op that already
+hit its own deadline arrives during the *next* op's wait window; charging
+it to that op stamps the wrong contract's key into the reported error.
+Zero is emitted too — "no stale error arrived during this op" is a
+measurement, and an absent field would be indistinguishable from a run
+that never counted. Errors that name no contract stay unattributable and
+still fail whichever op is waiting, so this can never swallow a real
+failure silently.
+
+Both fields reach the jsonl report. Neither reaches the telemetry
+dashboard yet: `insert_check_op` writes a fixed column list that has no
+`seq` or `errors_ignored` column, so `netcheck emit` publishes them and
+the ingest drops them without erroring. Read run order and skipped-error
+counts from the jsonl artifact until that table gains the columns
+(freenet/freenet-telemetry-dashboard#21).
 
 ## Production use (nova)
 
--- a/crates/netcheck/README.md
+++ b/crates/netcheck/README.md
@@ -98,11 +98,14 @@ disentangle.
 another contract and skipped. A server-side error for an op that already
 hit its own deadline arrives during the *next* op's wait window; charging
 it to that op stamps the wrong contract's key into the reported error.
-Zero is emitted too — "no stale error arrived during this op" is a
-measurement, and an absent field would be indistinguishable from a run
-that never counted. Errors that name no contract stay unattributable and
-still fail whichever op is waiting, so this can never swallow a real
-failure silently.
+Zero is emitted too, because an absent field would be indistinguishable
+from a run that never counted. Read it precisely: it counts errors the
+key filter skipped inside an op'''s own wait window. Errors arriving in the
+gap *between* ops are discarded by the pre-op drain, unfiltered and
+uncounted, so a zero means "the filter did not fire during this op", not
+"no stale error existed anywhere near it". Errors that name no contract
+stay unattributable and still fail whichever op is waiting, so this can
+never swallow a real failure silently.
 
 Both fields reach the jsonl report. Neither reaches the telemetry
 dashboard yet: `insert_check_op` writes a fixed column list that has no
--- a/crates/netcheck/README.md
+++ b/crates/netcheck/README.md
@@ -12,11 +12,16 @@ Each `put-get` run:
 2. Boots an **ephemeral peer** (`freenet network` child process with an
    empty data dir) that joins the network **through a different gateway**.
    Having no replicas of its own, it cannot answer any of its own GETs.
-3. GETs this run's contracts through that peer and verifies byte-identity.
-4. Re-GETs contracts published by previous runs (24 h / 48 h / 7 d windows,
-   tracked in a persistent JSON manifest) and verifies their blake3 state
-   hashes.
-5. Prints one JSON line per operation on stdout and exits non-zero if
+3. GETs this run's contracts through that peer (verifying byte-identity)
+   together with contracts published by previous runs (24 h / 48 h / 7 d
+   windows, tracked in a persistent JSON manifest, verified against their
+   blake3 state hashes). The two are issued as **one interleaved
+   sequence**: with a fixed 0 h → 24 h → 48 h → 7 d order, "how old the
+   contract is" and "how late in the run the GET was issued" were the same
+   variable, so an age effect could not be told apart from a
+   within-run session effect. The order is shuffled from a seed derived
+   from the run id, and that seed is logged, so a run stays reproducible.
+4. Prints one JSON line per operation on stdout and exits non-zero if
    anything failed. No retries by design: an operation that only succeeds
    on retry is the regression netcheck exists to surface.
 
@@ -70,6 +75,12 @@ One `"event":"run"` line with the conditions of the run, then one
 `freenet_version` is what separates "the network broke" from "the release
 that landed last night broke". Without it a failure cannot be attributed
 after the fact.
+
+`latency_ms` is a measurement on **both** arms. A failed op reports how
+long it actually took, not the configured `--op-timeout-secs`: reporting
+the deadline for every failure hid the difference between a fast terminal
+error from the node and the client giving up waiting, which is the
+distinction the field exists to draw.
 
 ## Production use (nova)
 
--- a/crates/netcheck/src/client.rs
+++ b/crates/netcheck/src/client.rs
@@ -146,7 +146,15 @@ fn error_contract_ids(err: &ClientError) -> Vec<ContractInstanceId> {
         ErrorKind::IncorrectState(key) => vec![*key.id()],
         // The shape terminal network failures actually arrive in: no key
         // field, the id written into the prose. See `contract_ids_in_text`.
-        ErrorKind::OperationError { cause } => contract_ids_in_text(cause.as_ref()),
+        //
+        // `Unhandled` is the same shape and matters for the same reason: it is
+        // what `ClientError::from(String)` produces, which is the sink stdlib's
+        // own stream-assembly failure path uses -- and stream assembly is the
+        // motivating failure class here, so leaving it unscanned would exempt
+        // the very errors this filter exists to attribute.
+        ErrorKind::OperationError { cause } | ErrorKind::Unhandled { cause } => {
+            contract_ids_in_text(cause.as_ref())
+        }
         // `ErrorKind` is `#[non_exhaustive]` too, and the same reasoning
         // applies: unrecognised means unattributed means fatal.
         _ => Vec::new(),
@@ -233,6 +241,30 @@ fn classify_get(
             match contract {
                 Some(contract) => Reaction::Done((contract, state)),
                 None => Reaction::Fatal(anyhow!("GET {label} ({id}) returned no contract")),
+            }
+        }
+        // A dead-ended GET does NOT arrive as `Err`. The node answers
+        // `Ok(NotFound)` on purpose, so a client can tell "contract genuinely
+        // absent" from "operation failed" -- see the comment on the arm that
+        // publishes it in `operations/get/op_ctx_task.rs`, which calls this
+        // "the dominant production `not_found` mode".
+        //
+        // Falling through to `Ignore` here is what made defect 2 survive its
+        // own fix: netcheck would discard the node's terminal answer, keep
+        // waiting, and report `latency_ms: 120000` -- the configured deadline
+        // again, for the single most common failure, now dressed as a
+        // measurement instead of an acknowledged constant. `NotFound` carries
+        // `instance_id`, so it is key-filtered exactly like every other
+        // terminal reply.
+        Ok(HostResponse::ContractResponse(ContractResponse::NotFound { instance_id })) => {
+            if instance_id != id {
+                Reaction::Ignore(format!(
+                    "GET {label}: NotFound for different key {instance_id}, ignoring"
+                ))
+            } else {
+                Reaction::Fatal(anyhow!(
+                    "GET {label} ({id}): node answered NotFound (contract not located)"
+                ))
             }
         }
         Ok(other) => Reaction::Ignore(format!(
@@ -459,6 +491,15 @@ mod tests {
 
     fn id_of(byte: u8) -> ContractInstanceId {
         ContractInstanceId::new([byte; 32])
+    }
+
+    fn reaction_name<T>(r: &Reaction<T>) -> &'static str {
+        match r {
+            Reaction::Done(_) => "Done",
+            Reaction::Ignore(_) => "Ignore",
+            Reaction::IgnoredError(_) => "IgnoredError",
+            Reaction::Fatal(_) => "Fatal",
+        }
     }
 
     /// A structured terminal GET failure: the API names the contract in a key
@@ -762,6 +803,54 @@ mod tests {
                 Reaction::Fatal(_) => {}
                 _ => panic!("a related-contract key must not be read as the error's subject"),
             }
+        }
+    }
+
+    #[test]
+    fn a_notfound_for_this_contract_fails_the_get_immediately() {
+        // The dominant production failure shape. Before this arm existed it
+        // fell through to `Ok(other) => Ignore`, so netcheck discarded the
+        // node's terminal answer and burned the full op timeout — reporting
+        // `latency_ms: 120000` for the most common failure there is, which is
+        // exactly the constant-latency defect this PR set out to remove.
+        let id = id_of(3);
+        match classify_get(
+            id,
+            "0h small-0",
+            Ok(HostResponse::ContractResponse(ContractResponse::NotFound {
+                instance_id: id,
+            })),
+        ) {
+            Reaction::Fatal(e) => {
+                let msg = e.to_string();
+                assert!(
+                    msg.contains("NotFound"),
+                    "the reported cause must say the node answered NotFound, got: {msg}"
+                );
+            }
+            other => panic!(
+                "a NotFound naming this contract is a terminal answer and must fail the GET                  fast, not be ignored into a synthetic timeout: {}",
+                reaction_name(&other)
+            ),
+        }
+    }
+
+    #[test]
+    fn a_notfound_for_another_contract_is_ignored() {
+        // Key-filtered exactly like every other terminal reply: a NotFound
+        // answering an op that already gave up must not kill this one.
+        match classify_get(
+            id_of(3),
+            "0h small-0",
+            Ok(HostResponse::ContractResponse(ContractResponse::NotFound {
+                instance_id: id_of(9),
+            })),
+        ) {
+            Reaction::Ignore(_) => {}
+            other => panic!(
+                "a NotFound naming a different contract must be ignored: {}",
+                reaction_name(&other)
+            ),
         }
     }
 
--- a/crates/netcheck/src/client.rs
+++ b/crates/netcheck/src/client.rs
@@ -5,8 +5,8 @@ use std::time::{Duration, Instant};
 
 use anyhow::{Result, anyhow, bail, ensure};
 use freenet_stdlib::client_api::{
-    ClientRequest, ContractRequest, ContractResponse, HostResponse, NodeQuery, QueryResponse,
-    WebApi,
+    ClientError, ClientRequest, ContractError, ContractRequest, ContractResponse, ErrorKind,
+    HostResponse, NodeQuery, QueryResponse, RequestError, WebApi,
 };
 use freenet_stdlib::prelude::*;
 use tokio_tungstenite::connect_async;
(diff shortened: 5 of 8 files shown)


sanity:
## Consolidated review — 2 independent lenses (code-first, testing) + verification

Reviewed at `bccd18ed7`. Lenses ran blind to each other. The testing lens executed
nine mutations against a throwaway worktree; the code-first lens read the diff before
the description. I independently verified the central finding below before posting.

**Verdict: the ordering and latency work is correct and worth landing. The key-filter
fix does not reach the failure it was built from, and its tests are constructed so
that this cannot show up as red. I'd treat that as blocking**, because the PR's stated
purpose is making the instrument trustworthy.

---

### BLOCKING 1 — the key filter cannot see the key in the errors the network actually sends

`error_contract_id` (client.rs:79) matches only
`ErrorKind::RequestError(RequestError::ContractError(..))` and `ErrorKind::IncorrectState`.
Everything else returns `None` → `Fatal` → attributed to the waiting op.

But the failures netcheck actually reports are `ErrorKind::OperationError`, whose
definition is:

```rust
#[error("error while executing operation in the network: {cause}")]
OperationError { cause: Cow<'static, str> },
```

**It has no key field.** The contract id lives as free text inside `cause`, written by
`synthesized_get_error_cause` (`get/op_ctx_task.rs`). And `client_events.rs:228-247`
collapses essentially every `OpError` into `OperationError`, so this is the dominant
terminal-failure shape rather than an edge case.

I verified this three ways rather than taking it on report: read `error_contract_id`
at this PR's head; confirmed the variant definition and its missing key field in
freenet-stdlib 0.8.4; and sampled a recent nightly, where every failure is either the
plain client-side timeout string or exactly that `OperationError` shape — e.g.

```
"error":"GET .../large-1MB (5fSGSjA8...): websocket error: client error:
 error while executing operation in the network: GET response stream assembly failed
 for 5fSGSjA8... after exhausting retries: stream assembly: no fragments received..."
```

Both instances the PR cites as motivation (runs `31073282100`, `31148314225`) are this
shape, so **both would still be misattributed after this PR**. The "Not covered" section
lists ChannelClosed / Disconnect / FailedOperation / NodeUnavailable and omits
`OperationError`, which is the one that matters.

### BLOCKING 2 — the fixture asserts a shape the system under test cannot produce

client.rs:326 documents `contract_error_for` as *"the error shape the live network
actually produced in the runs that motivated the key filter"* and then builds
`ContractError::Get { key, cause }`. The **cause string** is copied verbatim from the
live logs; the **enum wrapper** is not. All five key-filter tests are built on it.

Reconciling the two lenses, since they appear to disagree here: the testing lens
confirmed by mutation that these tests DO discriminate (M5/M6 turn them red), and it is
right. The code-first lens is also right that the input is unrealistic. Both hold at
once — the tests are valid tests *of the filter as written*, but the filter as written
does not cover what production emits. So the coverage is real and the reassurance it
provides is not. Worth noting that
`an_error_naming_no_contract_still_fails_the_waiting_op` (client.rs:369) is the test
that actually describes today's production behaviour, and it asserts the **unfixed**
outcome.

At least one test must be driven by an `ErrorKind::OperationError` fixture carrying the
real cause string, or netcheck's regression coverage for this defect is fictional.

### BLOCKING 3 — landing the latency fix over an inert filter makes the residual misreport *more* convincing

Today a misattributed line reads `latency_ms: 120000`, and #5266 itself names that
constant as the tell that the field is fake. After this PR the same misattributed error
— arriving mid-window, well before the deadline — is still `Fatal`, but now carries a
genuinely measured short elapsed. It reads as a real, fast, terminal failure of a
contract that never failed.

For a measuring instrument that is a regression, not an improvement: an overstated or
plausible-looking wrong number terminates investigation, which is precisely what
happened twice already on this file. This is the argument for landing the ordering and
latency work together **with** a working filter, rather than shipping the latency fix
over an inert one.

---

### IMPORTANT — the ordering fix can be silently undone with tests *and* clippy green

Every ordering test drives `interleave` directly against a synthetic vector
(`fixed_sequence`, put_get.rs:314). Nothing asserts that `run()` applies the shuffle to
the real `ops` vector, or that it does so **after** both loops have filled it.

Mutation M1b, executed: move the `interleave` call three lines up, above the retention
loop. **Tests green, clippy green** — and the exact confound this PR exists to remove is
back, every night. That outright deletion (M1) is caught at all is incidental: clippy's
`dead_code` fires only because nothing else calls the private helper. That is a
byproduct of visibility, not coverage of the decision.

This is the mutate-the-helper-not-the-call-site gap. AGENTS.md prescribes the bounded
`fn_body()` source-scrape pin for exactly this; whatever pin is added should be
mutation-tested against M1b when written, not merely written.

### IMPORTANT — the PUT half of the latency fix is shipped but unguarded

The pre-fix PUT loop had the identical `latency_ms: op_timeout.as_millis()` bug, the PR
correctly fixes it, and PUT failures land in the same JSONL with the same field — but
mutations M4-PUT (client.rs:183) and M8 (put_get.rs:80) both stay **green**. Half the
fix is pinned. One test each closes it.

### IMPORTANT — position-in-run is still not recoverable, so half of #5266 remains open

`OpReport` (report.rs:55) carries no index or timestamp, and emit.rs:207 stamps every
record with the same `now_ns`. Before this PR, position was recoverable because it was a
deterministic function of `age`. After it, position is genuinely random and **nowhere in
the emitted data** — it survives only as line order in the raw workflow log. So the
harness can now test "is it age?" but not "is it position?", which was the other half of
what #5266 asked. A `seq: usize` threaded into `CheckOp` completes it.

### IMPORTANT — `MissingRelated` / `ContractStackOverflow` name a *different* contract

client.rs:87 treats these as "this error is about contract `key`". Per `contract.rs:421`
the key there is the **related/dependency** contract, not the failing op's. Under the new
code such an error is ignored, the op runs to its 120s deadline, and the real cause is
discarded to stderr. Low reachability today (netcheck PUTs with
`RelatedContracts::default()`), so not a live misreport — but the classifier's semantics
are wrong as written. Only Get/Put/Update/Subscribe/IncorrectState have a key that is the
subject.

---

### Minor

- **client.rs:93** — `ErrorKind::IncorrectState` is claimed in the PR body and has zero
  coverage; deleting the arm is green.
- **client.rs:90** — `_ => return None` is a wildcard over a `#[non_exhaustive]` enum. All
  seven current variants are handled, so it is unreachable today, but a future
  key-carrying variant silently restores mis-attribution. Degrades rather than breaks;
  worth a comment naming that.
- **put_get.rs:314** — `fixed_sequence()` hardcodes the window list instead of deriving
  from `manifest::WINDOWS`, so a new retention window leaves these tests exercising the
  old shape.
- **client.rs:214/:265** — ignored errors go to stderr only, so they land in the Actions
  log rather than `last-run.jsonl`, the artifact people actually analyze. "Was there a
  stale error this run?" can no longer be answered from the artifact. An
  `errors_ignored` count on `OpReport` keeps the filter auditable.
- **put_get.rs:401 / client.rs:472** — the `assert_ne!` companions are tied to hardcoded
  `120s` literals duplicating `main.rs:111`'s `default_value_t`. They were never carrying
  the discrimination (the `assert_eq!` against an injected value is), so this is
  tightening, not a correctness problem.
- **Naming** — put_get.rs:270 is a uniform shuffle, not an interleave; a permutation of
  ~16 items can still put all four 7d ops in the tail. The guard test only checks that
  the last element's bucket varies across 20 seeds. The reproducibility note at :248 is
  also slightly overstated: the permutation depends on `ops.len()`, so the same run id
  with different manifest state gives a different order.
- **Calibration** — 0h is the only non-soft dimension, so only a 0h failure pages. After
  the shuffle a 0h GET may be issued 15+ minutes later, behind up to twelve 120s
  timeouts. Probably the right trade, but it changes what triggers a page and nobody has
  calibrated it.

---

### Confirmed sound

Latency honesty on the GET path (clock starts after the drain, `elapsed()` taken outside
the `Result`; `push_get_report` no longer receives `op_timeout`, so no path can
synthesize a constant). Shuffle is seeded, reproducible, logged, and varies per run, with
no dropped or duplicated ops. The error arm's structure mirrors the success arm's, and
key-less errors correctly stay fatal rather than hanging the op. The adversarial
interleaving in `a_late_error_during_a_get_does_not_fail_that_get` is genuinely
constructed over a real bincode websocket with deterministic ordering, not a timing race.
Baseline 27 tests pass, 10/10 repeat runs, no flakiness. And netcheck **is** covered by
the PR-time `--workspace` nextest run (ci.yml:556) — this is not the nightly-only trap.

### Suggested path

Keep the ordering and latency work. For the filter, the honest options are: drain until
this op's own response or its deadline (the issue's own fallback — costs fast-fail
latency signal); or match `OperationError` and extract the id from the cause text,
bounded to "ignore only when the text carries a parseable id that is not ours", falling
back to today's fatal behaviour when nothing parses. The second preserves the fast-fail
signal the latency fix exists to expose, so it is the one I'd pick.

[AI-assisted - Claude]

sanity:
Reviewed this while diagnosing the chronically-red Netcheck (Nightly). The reporting fixes here are real and needed, but there is a gap that matters more than any of the three defects the PR sets out to fix, and this PR is the natural place to close it.

## The largest single failure class is still swallowed by the catch-all

`classify_get` handles `GetResponse` and sends everything else to `Reaction::Ignore`. `ContractResponse::NotFound` is a **terminal answer to the GET**, and it lands in that catch-all. The op then keeps waiting until the client's own 120s deadline expires and reports `timed out after 120s`.

Measured across all 14 nightly runs (229 ops, 41 failures): **25 of 41 failures are the bare 120s timeout, and 20 of those 25 have a `NotFound` for the requested key sitting in the log**, discarded as "unexpected". So roughly half of all netcheck failures are mislabeled.

Directly verified on the most recent run (`31563947324`, 7d/small-2, key `Gjr4vwDS…`):

```
GET 20260805-051003/small-2: skipping unexpected response: ContractResponse(NotFound { instance_id: ContractInstanceId("Gjr4vwDSHg2s…
{"event":"op","op":"get",…,"ok":false,"latency_ms":120000,"error":"GET 20260805-051003/small-2 (Gjr4vwDS…) timed out after 120s"}
```

The node answered. The harness threw the answer away and reported a timeout.

Two consequences, and the first is the serious one:

1. **It points the diagnosis at the wrong subsystem.** "Timed out" reads as a transport or streaming problem. `NotFound` is a findability/placement problem — a routed GET that dead-ended for a contract that exists. Those have different owners and different fixes. This PR's stated purpose is to make the report able to support the diagnosis it exists for, and this is the single biggest way it currently cannot.

2. **It wastes ~113s per occurrence.** Median `NotFound` arrival is 6.8s against a 120s deadline. That is most of the nightly's runtime spent waiting for a reply that already came.

Suggested shape: a `NotFound` arm returning `Reaction::Fatal` with its own error text (something that says dead-end, not timeout), key-filtered on `instance_id` the same way the `GetResponse` arm is. It should still fail the op — this is not about making the check greener.

Worth noting the interaction with the `latency_ms` fix: once failures report a measured elapsed, a mislabeled `NotFound` will read as a *genuine fast timeout*, which is more plausible and therefore more misleading than today's obviously-fake 120000 constant. The two changes are better landed together.

## On the residual key-filter concern

The filter added here matches `ContractError`/`IncorrectState`. Every failure the nightly actually produces is `ErrorKind::OperationError`, which carries no key field — the contract id is free text inside `cause`. So the filter does not engage on the traffic that motivated it. The tests pass because the fixture constructs an error variant the system does not emit in these cases. Flagging it rather than prescribing a fix, since a real `OperationError`-driven fixture is the thing that would make the test meaningful.

## Separately

I have opened #5276 for a transport-side cause of the *other* main failure class (the stream-assembly inactivity timeouts): retransmitted packets are never re-acknowledged, so one lost receipt strands a packet in the sender's congestion window until `CWND_WAIT_TIMEOUT` aborts the stream. Independent of this PR, different crate, no conflict.

Full investigation notes going onto #5256.

[AI-assisted - Claude]

sanity:
## Full-tier review — 4 independent blind lenses

Risk tier: **Full**. This is a diagnostic harness rather than a production path, but the diff is
large (+1196 originally) and the whole point of the change is whether a number can be trusted,
which is not something a skim establishes.

Lenses run, each blind to the others, each reading the checked-out code rather than the diff text:
**code-first**, **testing**, **skeptical/adversarial**, **big-picture**. No external (Codex) pass;
per this repo's `pr-quality.md` the external pass is opt-in and Claude lenses were substituted.
Every `file:line` below was re-verified by hand before being repeated — reviewers hallucinate
locations, and two cited lines did not survive that check and were dropped.

**No BLOCKER from any lens.** Findings and disposition:

### Fixed in-branch

| Finding | Why it mattered |
|---|---|
| **`classify_get` ignored `Ok(NotFound)`** | The one that undermines the PR's own thesis. A dead-ended GET is not an `Err` — the node publishes `Ok(HostResponse::…NotFound { instance_id })` on purpose, and freenet-core's own comment on that arm calls it "the dominant production `not_found` mode". Every non-`GetResponse` `Ok` went to `Reaction::Ignore`, so netcheck discarded the terminal answer and burned the full deadline. Defect 2 was therefore still producing `latency_ms: 120000` for the most common failure there is — and now dressed as a measurement rather than an acknowledged constant, which is worse than before. `NotFound` carries `instance_id`, so it is now key-filtered like every other terminal reply. |
| **`ErrorKind::Unhandled` was not scanned** | It is the sink for `ClientError::from(String)`, which is what stdlib's own stream-assembly failure path produces. Stream assembly is the motivating failure class, so the key filter was exempting exactly the errors it exists to attribute. |
| **The PUT phase kept the confound the GET phase lost** | `build_run_contracts` appends `large-1MB` after the smalls and the PUT loop walked the vector, so the only large contract went last every night: "the 1 MB PUT failed" and "the last PUT failed" were one variable. That is the age/position confound again, on the size axis. Both phases now shuffle from the same run-derived seed. |
| **`StdRng` is not reproducible across `rand` versions** | The rustdoc promised that re-deriving the seed "replays that run exactly"; `rand` documents `StdRng`'s algorithm as explicitly unstable across releases. A dependabot bump would have silently invalidated every historical replay while `the_order_is_reproducible_from_the_run_id` stayed green — it only compares one build against itself. Now `ChaCha8Rng`, which carries a portability guarantee. |
| **The seed reached stderr, not the artifact** | The uploaded artifact is stdout. The seed records a *methodology* change: before the shuffle every 0h GET ran in the first few slots after settle; now one can be the run's last op, minutes later. Success rates either side are not comparable. `RunMeta.order_seed` is `Option<u64>`, so "no seed recorded" means "run from before the shuffle". |
| **`CONTRACT_ID_B58_LEN`'s upper bound was unpinned** | Every id in the tests is `[n; 32]` for small n and encodes to 43 base58 characters; only the all-ones id reaches 44. Every negative case was rejected by the round-trip check independently of the length filter, so narrowing the range to `32..=43` passed all 40 tests while dropping ~6% of the id space out of attribution — the filter failing to fire on precisely the errors it exists for. Added `id_of(255)`. |
| **The OTLP mirror had no test** | `seq` and `errors_ignored` are hand-copied from `OpReport` into `CheckOp` in `emit::run`, and substituting a literal for either compiled, passed clippy, and was invisible to every test. `seq` has no second source — every record of a run carries the same timestamp — so a zeroed `seq` loses the run's order silently. This is the "manually-mirrored telemetry field" row in `.claude/rules/bug-prevention-patterns.md`; added the source-scrape pin that row prescribes. |
| **Both call-site pins were vacuous** | `split_once` on the call text matches a *commented-out* call, so disabling either shuffle left the pin green. Verified by doing it. This was reachable in the pre-existing GET pin, not only the PUT pin added here. `split_at_call` now requires the call to begin its line after nothing but whitespace; with it, commenting out either shuffle turns both pins red. |
| **`errors_ignored`'s rustdoc overclaimed** | It said a zero means "no stale error arrived during this op", which `drain_stray_responses` can falsify — it runs before the window opens and discards errors unfiltered and uncounted. The count is deliberately left alone (a pre-op drain is a different event from a filtering decision an op made); the doc now states the window it actually measures. |
| **The description denied a change its own test pins** | It carried the *defect's* scope sentence ("does not affect `ok`/fail classification") over to the *fix*. The fix does change it: `a_late_error_during_a_get_does_not_fail_that_get` asserts a GET now succeeds where main failed it. On a check red every night since 2026-07-30, a change that can only reduce the failed-op count has to be declared or the first post-merge run reads as a recovery. See "What this changes about the nightly's numbers". |
| README, and a wrong assertion comment | README documented neither new field, overstated replayability from the seed alone, and did not record the ingest gap. One assertion said a token was rejected for being "one character longer than the id"; it is 44 characters, inside the bound, and rejected by the round-trip instead. |

Every behaviour added under review has a test **verified RED by mutating the fix away**, not merely
observed green: making the `NotFound` key check unconditional, replacing `seq: op.seq` with `seq: 0`,
narrowing the base58 range to `32..=43`, and commenting out either shuffle each turn their pin red.

### Filed rather than grown into this PR

- **freenet/freenet-telemetry-dashboard#21** — `insert_check_op` writes a fixed column list with no
  `seq` or `errors_ignored` column, so `netcheck emit` publishes both and the ingest drops them
  without erroring. Net effect on that database is a regression, not a no-op: op ordering used to be
  recoverable from the fixed `age` sequence and now is recoverable from nothing. The jsonl artifact
  keeps `seq`, so nothing is lost at the source. Recorded in the README and the description.

### Dismissed, with reasons

- **An ignored error's *text* still reaches no JSON record** when it arrives after its own op gave up
  (only the count does). Real, and narrowed by the `NotFound` and `Unhandled` fixes above — but
  closing the remainder needs a new report field and a design choice about where to attach text that
  by definition belongs to an op that has already been written out. Noted as a follow-up in the
  description rather than designed at the end of a review.
- **No per-op timestamp.** `seq` plus each op's `latency_ms` reconstructs elapsed time to within the
  inter-op gaps, which is enough for the question the shuffle raises. A new field for exactness is a
  separate change.
- **`#[non_exhaustive]` catch-alls unpinned.** The failure direction is safe — an unrecognised variant
  is unattributed, therefore fatal, therefore today's behaviour — and both sites say so in a comment.
  Pinning a hypothetical future variant is speculative.
- **`fn_body` is now a third hand-rolled copy in the repo.** True, and worth consolidating, but not
  in a PR about report trustworthiness.
- **Latency can exceed `--op-timeout-secs`** (the pre-deadline `send` and the drain are outside the
  deadline). Pre-existing, and it errs toward reporting real elapsed time, which is the direction this
  PR is arguing for.
- **`fake_node` panics into a detached task**; **`seq` uses `#[serde(default)]` rather than `Option`.**
  Both cosmetic at the scale netcheck runs.

### Verified clean

No test removed, ignored, or weakened — `op_report_roundtrip` *gains* assertions. No file under
`.github/` touched, so this cannot change whether the nightly reports failure; `verdict()` and
`Report::all_ok` appear in the diff only as context. No new cache, no non-string map key on a
serialized type, no `.send().await` in a loop. One new dev-dependency (`futures`, already in the
workspace) plus `rand_chacha`, already vendored transitively.

**Note on the nightly:** netcheck-nightly is a known-never-green harness tracked separately. It is
not evidence for or against this PR in either direction, and this PR was judged on the required
checks and on reading the code.

[AI-assisted - Claude]

sanity:
## Follow-up: a delta review of my own review fixes

The changes I made in response to the review above were themselves unreviewed, and I wrote them,
so I spawned a fresh adversarial lens on just that delta. It found something I had missed, which
is the argument for doing this rather than trusting my own read.

**The reproducibility fix covered half the version-dependent path, and its commit message
overclaimed.** Swapping `StdRng` for `ChaCha8Rng` pins the RNG *stream*. It does not pin the
*permutation*, because `ops.shuffle(&mut rng)` is `rand::seq::SliceRandom::shuffle` — the algorithm
lives in `rand`, not in the generator, and it changed between 0.8 and 0.9 (reverse Fisher-Yates via
`gen_index`, versus a forward pass via `IncreasingUniform`). Same seed, same byte-identical ChaCha
stream, different order. The breakage I claimed to have fixed had simply moved from `rand-rngs` to
`rand-seq`, and `the_order_is_reproducible_from_the_run_id` is blind to it for exactly the reason
it was blind to the original: it only compares one build against itself.

`SeedableRng::seed_from_u64` is a third instance — `rand_core`'s default expansion, not a ChaCha
guarantee, and its own doc calls changing it "a value-breaking change".

`interleave` now pins all three: `ChaCha8Rng` from seed bytes assembled inline, and Fisher-Yates
written out. The rustdoc now says what makes the claim true instead of asserting it.

The generalisable part, and the reason this got its own commit: **fixing the component you thought
of does not pin a property that depends on three**, and the test meant to guard it could not see
any of them, because self-consistency within one build is not evidence about portability across
builds. That is the same trap the PR was already fixing an instance of.

Also from that review:

- **`order_seed` reached the jsonl artifact but not the published record.** Unlike `seq` and
  `errors_ignored` it is not blocked on a fixed column list — `CheckRun.conditions` is free-form
  JSON — so there was no reason for the dashboard's verdict series to carry no marker of the very
  methodology change the field exists to record. Now mirrored, and outside the scope of
  freenet/freenet-telemetry-dashboard#21, which names only the other two.
- **The emit pin sliced to the test module rather than to the end of `run`.** Tight only while
  `run` is the last item in the file; adding a function after it would silently widen the window,
  at which point the pin passes with the mirrors living in the new function while `CheckOp` is
  zeroed. Bounded to `run`'s body, re-verified RED by replacing `seq: op.seq` with `seq: 0`.
- A panic message had a line-continuation join leave eighteen spaces mid-sentence.

Dismissed, with reasons: a `NotFound` naming another contract stays an uncounted `Ignore`, because
it is an `Ok` and the pre-existing treatment of a `GetResponse` for another key is also an uncounted
`Ignore` — counting one and not the other would make the field mean two things. One seed driving
both shuffles is fine: they run over different-length vectors of different contents, and two seeds
would need two recorded fields to stay replayable. And
`a_notfound_for_another_contract_is_ignored` does survive deleting the whole `NotFound` arm; its
sibling catches that mutation, so the pair is sound — I corrected the commit message's per-test
claim rather than the test.

Local verification on the current head: `cargo fmt --all --check` clean, `cargo clippy -p netcheck
--all-targets -- -D warnings` clean, 44/44 tests pass.

**Lenses on this PR in total:** code-first, testing, skeptical, big-picture on the original, plus a
fifth adversarial pass on the delta.

[AI-assisted - Claude]

PR #5276, opened 2026-08-12
fix(transport): re-acknowledge retransmitted packets so streams stop stalling

sanity:
## Problem

A retransmitted packet was never acknowledged, so a receipt that is destroyed outright could never be repaired by the retransmit it caused.

`ReportResult::AlreadyReceived` documents that the packet "**will be re-acknowledged** but should otherwise be ignored" (`received_packet_tracker.rs:141-143`). The implementation did not do this: the `Occupied` branch returned the variant without queueing anything (`:67`), and the only caller logs at `trace!` and `continue`s (`peer_connection.rs:1119-1126`). `pending_receipts.push` existed at exactly one site — the `Vacant` branch — so nothing could re-acknowledge.

**A lost receipt is normally recoverable**, and an earlier version of this description wrongly said otherwise. Receipts ride on noop packets, but those go through `packet_sending` and are tracked and retransmitted like any other, so a single lost noop is repaired by its own resend.

The hole is narrower than that, and real. Both flush sites in the recv loop call `get_receipts()` — a `mem::take` — and move the list into `noop()`:

```rust
let receipts = self.received_tracker.get_receipts();
if !receipts.is_empty() {
    if let Err(e) = self.noop(receipts).await {
        if e.is_transient_send_failure() {
            tracing::warn!(..., "ACK noop send failed, will retry");
```

`peer_connection.rs:1104-1115` and `:1477-1493`. On a transient send failure those receipts are **already gone** — there is nothing left to retry with, and the log message says otherwise. Nothing regenerates them. The same applies when all 12 retransmits of the receipt-carrying noop are lost.

When that happens the sender retransmits into silence and its packet stays pinned in flight. If the congestion window fills before the retransmit budget runs out, the stream aborts at `CWND_WAIT_TIMEOUT` (3s) via `drop_stream`, which purges the stream's packets from the resend queue — and the receiver sees a transfer that simply stopped. Answering the retransmit is the only signal that can repair it.

(The stranding is bounded by the `Abandon` path from #4345, so "permanent" — also in an earlier version of this description — was wrong too. The harm is an aborted stream, not a permanently pinned window.)

### Evidence, and how far it actually goes

The receiver-side signature is real and common. From vega, 03:00–05:00 UTC across August: **1441 stream-assembly inactivity timeouts** and **145 cwnd-wait aborts**, e.g.

```
Stream assembly timed out — no fragments received within inactivity window
    stream_id=2196300212 received_fragments=46 total_fragments=187 timeout_secs=5
```

Stall points across those 1441 are spread throughout the transfer (2.4% at ≤2 fragments, 5.8% at 3–10, 21.7% at 11–60, 69.0% mid-to-late, 1.1% near-complete), which fits an any-time receipt-loss event and does not fit a fixed window cap.

**What this evidence does not establish:** an earlier version cited `flightsize == cwnd` on every cwnd-wait abort as the signature of a window that cannot drain. That is close to tautological — the log at `outbound_stream.rs:199-209` only fires from inside a loop entered when `flightsize + packet_size > cwnd`, so it prints those two values whatever the cause. It does not discriminate this mechanism from a slow peer or a congested reverse path.

So: this closes a genuine hole that produces exactly this failure signature. It is **not** established that it is the dominant contributor to those 1441 timeouts. Nothing currently counts duplicate re-acks, so that cannot be measured today — a counter paired with the existing `record_stream_send_abort_cwnd()` would settle it, and is worth doing before this is credited with any netcheck improvement.

## Approach

Queue the receipt in the `Occupied` branch — what the enum's documented contract already promised.

- **Bounded by `MAX_PENDING_RECEIPTS`.** That cap is load-bearing, not advisory: `send_packet`'s receipt chunker splits an oversized list only once (`split_off` returns the tail, `peer_connection.rs:2390-2412`), so a list longer than ~2×289 receipts serializes past `MAX_DATA_SIZE`, and that error is not a transient send failure, so it kills the connection. The `Vacant` arm holds this line with `QueueFull`; this arm must not route around it, or a peer replaying old ids becomes a remote connection-kill. Dropping the re-ack at capacity is safe — the list is about to flush, and an unrepaired receipt draws another retransmit.
- **Deduplicated**, so a retransmit burst for one id does not queue one receipt per copy.
- **Still returns `AlreadyReceived`, not `QueueFull`**, so the caller can tell a duplicate from a fresh packet. Note this does *not* by itself guarantee the payload is skipped — the caller's `(_, true)` arm is matched before `(AlreadyReceived, _)` and wins whenever the 600ms receipt timer trips. That is pre-existing; an earlier version of this description asserted the opposite. Filed separately as #5277.

Delivery is guaranteed by the unconditional drain on the background ACK tick (`peer_connection.rs:1476-1493`), so worst-case latency is one tick rather than unbounded.

Verified no double-release of flight size: `report_received_receipts` only emits an ack-info tuple when the id is still in `pending_receipts` (`sent_packet_tracker.rs:538`), so a re-ack for an already-released packet produces nothing and cannot double-decrement. This is the invariant `.claude/rules/transport.md:86-110` governs and was the highest-risk interaction.

## Testing

Three tests, each verified by mutation to fail without the specific line it pins.

- `retransmitted_packet_is_reacknowledged_after_its_receipt_was_sent` — the production sequence. The **drain between the two reports** is what makes it discriminating: without it the first report's receipt is still queued and the assertion passes regardless. That drain is the "receipt went out and was lost" step.
- `repeated_retransmits_queue_one_receipt_per_flush` — pins the dedup.
- `reacknowledging_duplicates_cannot_grow_the_list_past_capacity` — pins the bound. Written against **600 distinct** replayed ids, deliberately: dedup alone caps the list at the number of distinct ids, so an earlier version of this test that replayed one id repeatedly **passed with the bound removed** and pinned nothing. With 600 it fails at `grew the receipt list to 600, past the 20 cap`.

Revert-and-run: removing `pending_receipts.push` fails exactly the first two (6 others pass); removing the length check fails exactly the third (8 others pass). So each test discriminates its own line rather than blanket-failing.

- `transport::received_packet_tracker`: 9 passed
- full `transport::` suite: 697 passed, 0 failed
- `cargo fmt --check` clean; `cargo clippy` adds no new warnings

`test_report_receipt_already_received` still asserts `pending_receipts.len() == 1` after a duplicate — the duplicate must not queue a *second* copy of an unsent receipt — with a comment saying which of the two things it pins.

## Not fixed here

Found while reviewing this, filed rather than folded in, since both are pre-existing and in the recv loop rather than this tracker:

- **#5277** — `(AlreadyReceived, true)` falls through to `process_inbound`, re-delivering a duplicate payload, which can permanently wedge legacy `InboundStream` reassembly.
- **#5278** — the receipt chunker splits only once, so >2× the per-packet maximum kills the connection.

## Scope

One contributing cause, not the whole netcheck red. That check's largest failure class is separate and not a transport problem at all: the network answers `NotFound` and the harness discards it, then reports a 120s timeout. Details on #5256 and #5271.

[AI-assisted - Claude]


sanity:
## Review round 1 — findings and what changed

Ran two independent blind reviewers (skeptical, code-first) per the Full-tier rule for transport changes. Both independently found the same three problems, one of them blocking. Pushed `d1c26be5b` and rewrote the PR description.

**Blocking, fixed.** The first commit removed the only bound on `pending_receipts` — the `Occupied` arm pushed without the length check the `Vacant` arm enforces via `QueueFull`. My description had asserted the list was "bounded by `MAX_PENDING_RECEIPTS` (20)", which was true *before* the change and false after it. That matters more than it sounds: the receipt chunker at `peer_connection.rs:2390-2412` splits an oversized list only once, so >2×289 receipts serializes past `MAX_DATA_SIZE`, and that error is not transient, so it kills the connection. A peer replaying old packet ids could have triggered it. Now bounded, with a test.

**My stated mechanism was wrong, corrected.** I claimed receipts are never retransmitted, so a lost one is unrecoverable. Not true — `noop()` goes through `packet_sending` and is tracked and retransmitted like any other packet. The genuinely unrecoverable case is narrower: the recv loop's flush sites have already `mem::take`n the list into `noop()` when the send fails, so their "will retry" log has nothing left to retry with. That hole is real and the fix still closes it, but the description now says so accurately rather than overclaiming.

**Weak evidence withdrawn.** I cited `flightsize == cwnd` on every cwnd-wait abort as the signature of a window that cannot drain. It is near-tautological: the log only fires from inside a loop entered when `flightsize + packet_size > cwnd`. It does not discriminate this mechanism from a slow peer. The stall-point distribution is the honest evidence, and I have added an explicit statement that this is **not** established as the dominant contributor to the 1441 timeouts, plus a note that a duplicate-re-ack counter is what would settle it.

**A test that pinned nothing.** My first bound test replayed one id repeatedly and **passed with the bound removed** — dedup alone caps the list at the number of *distinct* ids. Rewritten against 600 distinct ids; it now fails with `grew the receipt list to 600, past the 20 cap`. Each of the three tests is individually mutation-verified to fail without the specific line it pins.

**Not folded in, filed instead** — both pre-existing, both in the recv loop rather than this tracker: #5277 (`(AlreadyReceived, true)` falls through to `process_inbound`, re-delivering a duplicate payload and permanently wedging legacy `InboundStream` reassembly) and #5278 (the chunker bug above).

**Confirmed clean** by both reviewers and re-verified: no double-release of flight size. `report_received_receipts` only emits an ack-info tuple when the id is still in `pending_receipts` (`sent_packet_tracker.rs:538`), so a re-ack for an already-released packet produces nothing. Also no starvation of the ACK tick (`deterministic_select!` rotates rather than being `biased`), and no concurrency hazard (the tracker is a plain field on the single connection task).

One reviewer suggestion I did **not** take, flagging it as a judgement call: fixing #5277's arm ordering here. It is a separate defect with a different failure mode, and it touches the recv loop, so it deserves its own diff and its own review rather than riding along.

Still open and worth a reviewer's opinion: whether this should land at all without the counter that would let us tell in production whether it helped.

[AI-assisted - Claude]

sanity:
## Multi-perspective review — Full tier

Transport is a named high-risk surface, so this got the Full treatment: **four independent blind reviewers** (code-first, testing-adequacy, skeptical/adversarial, big-picture) run in parallel on the rebased head, each told not to rubber-stamp, plus synthesis. Codex was not run; four diverse Claude lenses were substituted, recorded here per the review rule.

The branch was rebased onto current `main` first (clean, no conflicts), so the review ran on the code that would actually land.

### Verdict

**The fix is correct. Do not merge yet — Simulation is reproducibly red on this branch.** Details in the last section; it is not caused by the review commit.

### What was verified (not assumed)

Two lenses independently traced, and I re-derived, the highest-risk interaction: `report_received_receipts` only emits an ack-info tuple while the id is still in `pending_receipts` (`sent_packet_tracker.rs`), so a re-ack of an already-released packet produces nothing. No double-decrement of flight size, no RTT sample from a stale packet, no double-count for BBR. `mark_retransmitted` puts a re-acked resend on the Karn branch, so RTT is not sampled from it either.

On amplification: a duplicate can never return `QueueFull` (only the `Vacant` arm can), so it never reaches the per-packet flush arm. Extra outbound is amortised at **≤ 1 noop per `ACK_CHECK_INTERVAL` per connection**, and only while a connection is receiving duplicates and nothing else. A replaying peer costs one hashmap lookup plus a ≤20-element scan and gets a deamplifying reflection. No feedback loop.

### Findings addressed

**1. Two factually false claims in the new test rustdoc** — the most serious finding, caught by three lenses. `received_packet_tracker.rs` asserted that receipts "travel on noop / piggybacked packets, which are never retransmitted" and that a stranded packet has "no remaining way to clear them". Both are false; both are fossils of claims this PR's own description explicitly retracts; and the first **contradicted a correct comment 150 lines above it in the same file**. That comment is what a future reader will consult when deciding whether the `Occupied` push can be deleted. Rewritten.

**2. Comment accuracy.** `send_packet` → `packet_sending`; the chunker mechanism restated correctly (`split_off` returns the *tail*, so the failure threshold is 2× `max_num`, not `max_num`); all **four** receipt-destroying sites named, not just the two recv-loop ones (`outbound_short_message` and the fragment send have the identical take-then-fail shape and log nothing at all); the real worst-case re-ack latency stated — the 100 ms ACK tick is rebuilt inside every `recv()` call and can be starved indefinitely on a busy connection, so the actual bound is the 600 ms path; and the mixed dup-then-fresh path reaching `MAX_PENDING_RECEIPTS + 1` documented.

**3. `.claude/rules/transport.md` updated.** This change makes duplicate receipts for an already-released packet a **deliberate, routine wire event** rather than an accident, and duplicate ACK ids feed straight into the release-exactly-once invariant that rule records as the cause of #4345. That dependency existed only in this PR's description, which no future agent reads. It is now in the path-scoped rule, with an explicit "do not make an unknown receipt id an error, and do not hoist per-id work above the membership guard".

**4. New composition test — the gap that mattered most.** The three original tests pin the mechanism inside one struct; none pinned the **effect** the fix exists to produce. `reacknowledged_retransmit_drains_flightsize_exactly_once` composes `ReceivedPacketTracker` + `SentPacketTracker` + `BbrController` and drives the production sequence: ack → receipt destroyed by a `mem::take`'d failed send → retransmit → re-ack → release, asserting the packet leaves the congestion window with the retransmit budget untouched. Its second half pins the double-release invariant, which had **zero** coverage and was verified only by prose.

**Mutation-verified, run not assumed:** reverting the `Occupied` arm's push takes the suite from 718 passed / 0 failed to 714 / 4 failed, including this new test.

**5. Tightened the capacity test.** `reacknowledging_duplicates_cannot_grow_the_list_past_capacity` asserted `<=`, which also passes if the arm stops re-acknowledging altogether. Now `==`, and it fails under the revert mutation where before it passed vacuously. Added the interleaved dup-then-fresh case so the `cap + 1` ceiling is pinned rather than accidentally true.

**6. Two log lines that lied.** Both flush sites logged `"will retry"` / `"will retry next tick"` after `get_receipts()` had already `mem::take`n the list — there was nothing left to retry with, and the next tick starts from an empty list. This PR makes the situation *recoverable*, but the log stays actively misleading to anyone debugging from it. Replaced with what actually happens.

**7. A latent trap recorded.** `packet_loss_proportion`'s decay runs for every id in the receipt list **before** the `pending_receipts` membership guard, so deliberate duplicate receipts now bias the loss estimate toward "no loss" in exactly the lossy regime that produces them. Inert today — the only production caller discards the value and BBR derives loss from its own round accounting — so this is a comment at the decay site, not a change. It becomes a real bias the moment anyone consumes that value.

### Dismissed, with reasons

- **"Add a simulation test."** Dismissed on the testing lens's own evidence rather than convenience: the DST drop callback is `Fn(&str, SocketAddr, SocketAddr) -> PacketDeliveryDecision` — no access to packet *contents* — so it cannot selectively drop the receipt-carrying noop, which is the precise event this repairs. A blanket loss rate would exercise the path only stochastically and would pass with the fix reverted more often than not. `SIMULATION_ACK_CHECK_INTERVAL` is also relaxed on the stated assumption that this scenario does not arise in-memory. Unit + the new composition test is the right level under `test-philosophy.md`.
- **#5277 / #5278 deferral.** Legitimate. Both verified open and real; both are pre-existing, live in the recv loop rather than this tracker, and this PR worsens neither.
- **"Add the duplicate-re-ack counter before merge."** Deferred to a follow-up. It needs a new field on `StreamAbortSnapshot` plus the `router_snapshot` export path — a cross-cutting telemetry change with its own named bug-pattern row in this repo. This PR's own text scopes the counter as needed *"before this is credited with any netcheck improvement"*, i.e. before crediting, not before merging. The fix stands on its own: it implements a documented enum contract that was never implemented.

### Blocking: Simulation is reproducibly red on this branch

`test_5147_originator_target_list_cuts_duplicate_deliveries` fails after all 3 nextest attempts, at `crates/core/tests/simulation_integration.rs:17895`:

> `premise: the treatment arm never excluded a delivering sender, so sender_skips explains none of the leg-accounting gap it was added to explain`

This is **not** caused by the review commit. It fails identically on all three heads of this branch:

| head | date | result |
|---|---|---|
| `d1c26be5b` (original) | 2026-08-12 | same test, TRY 3 FAIL |
| `18c1085de` (rebased) | 2026-08-24 | same test, TRY 3 FAIL |
| `01b210973` (review fixes) | 2026-08-24 | same test, TRY 3 FAIL |

Meanwhile `main` and the merge queue are green across their recent runs. So it is consistent, branch-specific, and survived a rebase spanning ~250 commits of `main` — which rules out both a flake and a stale base.

Note the assertion that fires is a **premise guard**, not the property under test: it says the treatment arm never excluded a delivering sender, so `sender_skips` is zero and cannot account for the leg gap. A plausible reading is that this PR's added ACK traffic perturbs delivery ordering enough that the sender-exclusion filter never fires, making the premise unsatisfiable without the suppression logic itself being wrong. That is a **hypothesis, not a diagnosis** — I have not reproduced it locally, so it needs confirming before anyone acts on it.

Either way the rule is the rule: red CI on the head commit blocks the merge. The PR is otherwise rebased, reviewed, and ready.

Lenses run: code-first, testing-adequacy, skeptical/adversarial, big-picture (4 independent Claude lenses; Codex not run, substitution recorded).

[AI-assisted - Claude]

sanity:
### Follow-up: a clean control for the Simulation failure

Since posting the review I got a control worth recording, because it removes the main alternative explanation.

**PR #4997** — an unrelated branch (UPDATE rate limiter), rebased onto the same `main`, run in the same CI at the same time — has **Simulation passing in 13m16s**, along with all seven other required checks. So `test_5147_originator_target_list_cuts_duplicate_deliveries` passes on today's `main` plus an unrelated change, in this environment, right now.

Combined with the earlier evidence — the same test failing all 3 nextest attempts on all three heads of this branch, spanning 2026-08-12 to today and a rebase across ~250 commits — that leaves very little room for "brittle test" or "bad runner". The failure tracks this branch.

I have **not** identified the mechanism, and I want to be clear that the hypothesis in the review (that the added ACK traffic perturbs delivery ordering so the sender-exclusion filter never fires, making the `treatment.sender_skips > 0` premise unsatisfiable) is still a hypothesis. The assertion that fires is a premise guard rather than the property under test, so it is genuinely possible that the guard is over-tight and this PR is merely the thing that exposed it — that would make it a test bug rather than a product bug, but it still has to be resolved before merge either way.

One practical note for whoever picks this up: these tests are feature-gated, and a plain

```
cargo test -p freenet --test simulation_integration <name>
```

builds, runs, and reports `running 0 tests ... 0 filtered out` — it looks like a filter typo but the binary simply contains no tests. The features CI uses are `--features simulation_tests,testing`.

[AI-assisted - Claude]

sanity:
### Diagnosis: bisected to one line. Causation proven, mechanism still open.

Following up again — I reproduced the Simulation failure locally and bisected it. The hypothesis in my earlier comments can now be replaced with a measurement.

**Reproduction** (note the feature flags; without them the binary registers zero tests and reports `running 0 tests ... 0 filtered out`, which reads like a filter typo):

```
cargo test -p freenet --locked --test simulation_integration \
  --features simulation_tests,testing \
  test_5147_originator_target_list_cuts_duplicate_deliveries
```

**Result on this branch as it stands:** FAILED in 35.8s at `simulation_integration.rs:17895`, the `assert!(treatment.sender_skips > 0)` premise guard.

**Then I reverted exactly one thing** — the `pending_receipts.push(packet_id)` this PR adds to the `Occupied` arm of `report_received_packet`, leaving everything else on the branch in place — and re-ran the same command:

**PASSED, 1 passed / 0 failed, 75.5s.**

So the failure is caused by this PR's re-acknowledgement, isolated to that single line. Together with #4997's Simulation passing green on an unrelated branch off the same `main`, environment and test brittleness are both ruled out as the primary cause.

**What this does not settle: whether the product or the test is wrong.**

Counters from the failing run's two arms:

| arm | sends | suppressed | summary_skips |
|---|---|---|---|
| control | 460 | 0 | 409 |
| treatment | 256 | 500 | 121 |

The #5147 suppression itself is working hard in the treatment arm (500 suppressions). What goes to zero is specifically `sender_skips` — the count of times a *delivering sender* was excluded from fan-out.

There is a reading in which this PR is correct and the test premise is the casualty: unacknowledged retransmits produce duplicate payload deliveries (the `(_, true)` arm re-processing, #5277), duplicate deliveries are what create a "delivering sender" to exclude, and acknowledging the retransmit removes them. On that reading the assertion was resting on duplicate traffic that this PR legitimately eliminates, and the fix should land with the premise guard adjusted.

There is also a reading in which the ACK traffic genuinely perturbs the fan-out accounting that #5147 depends on, and that would be a real regression.

**I have not distinguished them, and I do not want to guess at it in a review comment** — `sender_skips` is not printed by the test, only named in the assertion message, so telling the two apart needs someone who knows what that counter is meant to represent. @iduartgomez / @netsirius, this is your call: the bisect above makes it a 75-second experiment either way.

Until it is resolved the PR cannot merge on the red-CI rule, but it is otherwise rebased, reviewed, and ready, and the cause is now a single known line rather than an open question.

[AI-assisted - Claude]

PR #5287, opened 2026-08-12
fix: propagate delegate execution failures to the client (2 of 3 bail-outs; third is an open question)

### LINKED ISSUE #5263: Delegate execution errors reach the client as an empty successful DelegateResponse
## Summary

When a delegate's `process` returns an error during `DelegateRequest::ApplicationMessages`, the client receives a **successful-looking** `HostResponse::DelegateResponse { values: [] }` instead of an error. Apps cannot distinguish "the delegate answered nothing" from "the delegate failed", so request/response-style delegates silently hang forever.

## Where

`crates/core/src/contract.rs`, `handle_delegate_with_contract_requests` (~line 570, `execute_delegate_request` error arm):

```rust
Err(err) => {
    ...
    tracing::error!(..., phase = "execution_failed", "Failed executing delegate request");
    // Return whatever we accumulated so far
    return accumulated_messages;   // ← empty on the common first-iteration failure
}
```

The error is logged node-side only; `client_events.rs` then wraps the empty list in `Ok(HostResponse::DelegateResponse { key, values })` and logs `outcome="executed"`.

## How this bites in practice (real user report)

A Freebird user's session ended up on an unattested WebSocket (stale auth token / node reached through a proxy, so `origin_contract = None`). Every delegate `ApplicationMessages` request then failed inside the delegate with `missing message origin` — but the browser console showed only:

```
host response: Delegate(key=FQmU…, 0 values)   × 5
```

No error ever reached the app, so its startup KV `Get` (stored posting key) never resolved and the UI spun on a loading screen indefinitely. The node log had the truth (`phase = "execution_failed"`), but nothing client-side did.

Notably, the pre-existing comment in `client_events/websocket.rs` about closing stale-token connections already recognizes this failure shape ("all subsequent messages would be processed with origin_contract=None … and fails with 'missing message origin'") — but when a connection does end up unattested for any other reason, the resulting delegate errors are invisible to the client.

## Expected

Delegate execution failure should propagate to the requesting client as a `ClientError` (like other request failures), or at minimum as a distinguishable response — never as an empty success. The mid-loop partial-accumulation case could still return accumulated messages plus an error indication, but the common case (failure on the first iteration, nothing accumulated) should be a hard error.

## Repro sketch

1. Register any request/response delegate that rejects `origin: None` (both River's chat delegate and Freebird's KV delegate do).
2. Open a WebSocket to `/v1/contract/command` without an auth token and send `DelegateRequest::ApplicationMessages`.
3. Observe: node logs `Failed executing delegate request`, client receives `DelegateResponse { values: [] }` with no error.

Node version: reproduced against v0.2.123 sources (behavior unchanged on current main, cfc3e656).

sanity:
## Problem

A delegate execution failure reached the client as an **empty successful** `DelegateResponse`. `handle_delegate_with_contract_requests` logged `phase = "execution_failed"` and then returned the messages it had accumulated so far; `process_open_request` mapped that straight to `HostResponse::DelegateResponse`. An app could not distinguish "the delegate answered nothing" from "the delegate failed", so it sat on a loading screen indefinitely.

## Scope — read this before assuming the class is closed

`handle_delegate_with_contract_requests` has **three** `return Ok(accumulated_messages)` bail-outs. This PR closes **two** of them, deliberately, and the third is left with a documented open question rather than a guess.

| site | disposition |
|---|---|
| execution failure (`Err` from `execute_delegate_request`) | **fixed** — now returns `Err`, the original #5263 case |
| unexpected response variant (`Ok(_other)`) | **fixed** — an internal invariant violation, nothing legitimate reaches it |
| iteration limit exceeded | **left as `Ok`** — see below |

**The iteration-limit arm is an open question, not deferred work.** Exhausting the budget may be a legitimate outcome for a delegate with more contract requests than the cap allows, in which case converting it to `Err` would turn working delegates into client-visible failures — a regression, not a fix. Whether the cap exists to bound work or to detect misbehaviour is author intent and is not recoverable from the code. Tracked as **#5454**; @iduartgomez is the likeliest person to know which was meant. The arm now carries a comment saying exactly this, marked open, so a future reader does not mistake the absence of a change for a considered "no".

One more deliberate carve-out, unchanged by this PR: `is_missing_delegate()` still returns an empty success for a delegate that is not installed on this node, so migration probes stay benign. It is a typed match on `RequestError::DelegateError(Missing(_))`, not a string sniff, so it cannot swallow genuine failures — but nothing pins it, and it is the one branch separating a true positive from a false positive here. Also in #5454.

## Solution

`ContractHandlerEvent::DelegateResponse` now carries `Result<Vec<OutboundDelegateMsg>, ExecutorError>`, and `client_events.rs` maps it through a new `delegate_request_outcome` so the classification lives in one place with its own tests rather than inline in `process_open_request`.

A genuine execution failure becomes `Err(Error::Executor(..))` and reaches the client as a `ClientError`. A delegate that legitimately produces no outbound messages still returns SUCCESS with an empty `values` — that path is unchanged and pinned by `successful_execution_returns_the_values`.

## Also fixed here: a regression this PR introduced

The early return added to `handle_delegate_notification` originally sat **above** `delegate_app_registry::sweep_expired()`, which is the registry's only production caller and the thing that time-bounds the delegate→apps map under the AGENTS.md GC-exemption rule. That made the TTL sweep conditional on the delegate having *succeeded*.

Not theoretical: the failure #5263 exists to fix is one a delegate repeats on **every** invocation (rejecting `origin: None`), so on an affected node the backstop would never run at all, and registrations for apps that vanished without a clean disconnect would pin their slots against `MAX_APPS_PER_DELEGATE` until restart. The bug fix would have created the exact condition the TTL was introduced to prevent. The sweep now runs before the return, pinned by `ttl_sweep_precedes_the_execution_failure_return`.

## Testing

- `delegate_request_surfaces_execution_failure_5263` — drives `handle_contract_event` over a real channel with a mock executor; verified red before the fix, green after.
- `delegate_request_outcome_tests` (4) — each asserts a specific `Error` variant against a specific input, not merely "is an error". `genuine_execution_failure_becomes_executor_error` is the one that pins #5263; the other three are the success guard and refactor-safety pins for the extraction.
- `ttl_sweep_precedes_the_execution_failure_return` and `unexpected_response_variant_does_not_become_a_fake_success` — source pins. Both mock executors return `Err` unconditionally, so neither class can be driven behaviourally today. Both truncate at the test module before scraping, because their needles occur in their own text (the vacuous-pin shape surveyed in **#5450**).

**Known gap, stated rather than papered over:** no test exercises a *real* WASM delegate failing end-to-end to a real WebSocket client. Coverage is two unit layers joined by a compiler-checked shared type, which is a reasonable argument that this PR does not need it — but it is not the same thing as having it.

Rebased onto current main over ~92 commits. The only conflict was in `client_events.rs`, where main's `unsupported_request_tests` (from merged #5392) and this branch's `delegate_request_outcome_tests` both append at end-of-file; both kept, verified by hunk enumeration (4 hunks, the last a pure append) with zero deletions touching #5392 content.

Closes #5263

[AI-assisted - Claude]

### DIFF
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -465,6 +465,73 @@ where
     }
 }
 
+/// Maps the contract handler's response to a client-driven `DelegateRequest`
+/// into either the values to hand back to the client, or the `Error` to
+/// propagate — logging exactly as the inline match this was extracted from
+/// did. Split out of `process_open_request`'s `DelegateOp` arm so the fix for
+/// #5263 (a genuine delegate execution failure must reach the client as
+/// `Err`, not an empty `Ok`) is unit-testable without constructing an
+/// `OpManager`.
+fn delegate_request_outcome(
+    res: Result<ContractHandlerEvent, crate::contract::ContractError>,
+    client_id: ClientId,
+    request_id: RequestId,
+    delegate_key: &DelegateKey,
+    msg_type: Option<&str>,
+) -> Result<Vec<OutboundDelegateMsg>, Error> {
+    match res {
+        Ok(ContractHandlerEvent::DelegateResponse(Ok(values))) => {
+            if let Some(mt) = msg_type {
+                tracing::info!(
+                    delegate = %delegate_key,
+                    msg_type = %mt,
+                    request_id = %request_id,
+                    outcome = "executed",
+                    "DelegateRequest dispatch"
+                );
+            }
+            Ok(values)
+        }
+        // Genuine delegate execution failure (#5263). Previously this arm
+        // didn't exist: a failure was indistinguishable from
+        // `Ok(ContractHandlerEvent::DelegateResponse(Ok(vec![])))`, so the
+        // client silently received an empty successful response instead of
+        // an error.
+        Ok(ContractHandlerEvent::DelegateResponse(Err(exec_err))) => {
+            tracing::error!(
+                client_id = %client_id,
+                request_id = %request_id,
+                delegate = %delegate_key,
+                error = %exec_err,
+                phase = "error",
+                "Delegate execution failed"
+            );
+            Err(Error::Executor(exec_err))
+        }
+        Err(err) => {
+            tracing::error!(
+                client_id = %client_id,
+                request_id = %request_id,
+                delegate = %delegate_key,
+                error = %err,
+                phase = "error",
+                "Delegate operation failed (contract error)"
+            );
+            Err(Error::Contract(err))
+        }
+        Ok(_) => {
+            tracing::error!(
+                client_id = %client_id,
+                request_id = %request_id,
+                delegate = %delegate_key,
+                phase = "error",
+                "Delegate operation failed (unexpected state)"
+            );
+            Err(Error::Op(OpError::UnexpectedOpState))
+        }
+    }
+}
+
 #[inline]
 async fn process_open_request(
     mut request: OpenRequest<'static>,
@@ -1672,7 +1739,7 @@ async fn process_open_request(
                 // it on a channel the delegate/client cannot reach.
                 let user_context = request.user_context;
 
-                let res = match op_manager
+                let handler_res = op_manager
                     .notify_contract_handler_prioritized(
                         ContractHandlerEvent::DelegateRequest {
                             req,
@@ -1682,58 +1749,14 @@ async fn process_open_request(
                         },
                         crate::contract::Priority::ClientLocal,
                     )
-                    .await
-                {
-                    Ok(ContractHandlerEvent::DelegateResponse(Ok(res))) => {
-                        if let Some(ref mt) = msg_type {
-                            tracing::info!(
-                                delegate = %delegate_key,
-                                msg_type = %mt,
-                                request_id = %request_id,
-                                outcome = "executed",
-                                "DelegateRequest dispatch"
-                            );
-                        }
-                        res
-                    }
-                    // Genuine delegate execution failure (#5263). Previously
-                    // this arm didn't exist: a failure was indistinguishable
-                    // from `Ok(ContractHandlerEvent::DelegateResponse(Ok(vec![])))`,
-                    // so the client silently received an empty successful
-                    // response instead of an error.
-                    Ok(ContractHandlerEvent::DelegateResponse(Err(exec_err))) => {
-                        tracing::error!(
-                            client_id = %client_id,
-                            request_id = %request_id,
-                            delegate = %delegate_key,
-                            error = %exec_err,
-                            phase = "error",
-                            "Delegate execution failed"
-                        );
-                        return Err(Error::Executor(exec_err));
-                    }
-                    Err(err) => {
-                        tracing::error!(
-                            client_id = %client_id,
-                            request_id = %request_id,
-                            delegate = %delegate_key,
-                            error = %err,
-                            phase = "error",
-                            "Delegate operation failed (contract error)"
-                        );
-                        return Err(Error::Contract(err));
-                    }
-                    Ok(_) => {
-                        tracing::error!(
-                            client_id = %client_id,
-                            request_id = %request_id,
-                            delegate = %delegate_key,
-                            phase = "error",
-                            "Delegate operation failed (unexpected state)"
-                        );
-                        return Err(Error::Op(OpError::UnexpectedOpState));
-                    }
-                };
+                    .await;
+                let res = delegate_request_outcome(
+                    handler_res,
+                    client_id,
+                    request_id,
+                    &delegate_key,
+                    msg_type.as_deref(),
+                )?;
 
                 let host_response = Ok(HostResponse::DelegateResponse {
                     key: delegate_key.clone(),
@@ -2454,6 +2477,89 @@ mod unsupported_request_tests {
              client, not merely log and fall through to `Ok(None)` — a \
              client whose request lands here would wait forever with no \
              response. Arm content: {arm:?}"
+        );
+    }
+}
+
+/// Unit tests for `delegate_request_outcome`, the mapping `process_open_request`
+/// uses to turn the contract handler's response to a client `DelegateRequest`
+/// into either the values to return or the `Error` to propagate. These drive
+/// the function directly — no `OpManager` needed — so the #5263 fix (a
+/// genuine delegate execution failure reaches the client as `Err`, not an
+/// empty `Ok`) is covered at the exact layer the client actually sees.
+#[cfg(test)]
+mod delegate_request_outcome_tests {
+    use super::*;
+
+    fn key() -> DelegateKey {
+        DelegateKey::new([3u8; 32], CodeHash::new([0u8; 32]))
+    }
+
+    #[test]
+    fn genuine_execution_failure_becomes_executor_error() {
+        let exec_err = crate::contract::ExecutorError::other(anyhow::anyhow!(
+            "missing message origin for message type: \"application message\""
+        ));
+        let res = delegate_request_outcome(
+            Ok(ContractHandlerEvent::DelegateResponse(Err(exec_err))),
+            ClientId::next(),
+            RequestId::new(),
+            &key(),
+            None,
+        );
+        assert!(
+            matches!(res, Err(Error::Executor(_))),
+            "a genuine delegate execution failure must reach the client as \
+             Err(Error::Executor(_)), not a fake success (#5263); got {res:?}"
+        );
+    }
+
+    #[test]
+    fn successful_execution_returns_the_values() {
+        let res = delegate_request_outcome(
+            Ok(ContractHandlerEvent::DelegateResponse(Ok(vec![]))),
+            ClientId::next(),
+            RequestId::new(),
+            &key(),
+            Some("ApplicationMessages"),
+        );
+        assert!(
+            res.expect("a successful response must not be an error")
+                .is_empty(),
+            "an empty Vec on a genuinely successful response is still correct — \
+             only a genuine failure should become Err"
+        );
+    }
+
+    #[test]
+    fn channel_level_contract_error_propagates_as_contract_error() {
+        let res = delegate_request_outcome(
+            Err(crate::contract::ContractError::NoEvHandlerResponse),
+            ClientId::next(),
+            RequestId::new(),
+            &key(),
+            None,
+        );
+        assert!(
+            matches!(res, Err(Error::Contract(_))),
+            "a channel-level failure (handler dropped, no response) must \
+             surface as Error::Contract; got {res:?}"
+        );
+    }
+
+    #[test]
+    fn unexpected_response_variant_is_reported_as_unexpected_op_state() {
+        let res = delegate_request_outcome(
+            Ok(ContractHandlerEvent::ExportUserSecretsResponse(Ok(vec![]))),
+            ClientId::next(),
+            RequestId::new(),
+            &key(),
+            None,
+        );
+        assert!(
+            matches!(res, Err(Error::Op(OpError::UnexpectedOpState))),
+            "a response of the wrong ContractHandlerEvent variant must be \
+             reported, not silently accepted; got {res:?}"
         );
     }
 }
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -1684,7 +1684,7 @@ async fn process_open_request(
                     )
                     .await
                 {
-                    Ok(ContractHandlerEvent::DelegateResponse(res)) => {
+                    Ok(ContractHandlerEvent::DelegateResponse(Ok(res))) => {
                         if let Some(ref mt) = msg_type {
                             tracing::info!(
                                 delegate = %delegate_key,
@@ -1695,6 +1695,22 @@ async fn process_open_request(
                             );
                         }
                         res
+                    }
+                    // Genuine delegate execution failure (#5263). Previously
+                    // this arm didn't exist: a failure was indistinguishable
+                    // from `Ok(ContractHandlerEvent::DelegateResponse(Ok(vec![])))`,
+                    // so the client silently received an empty successful
+                    // response instead of an error.
+                    Ok(ContractHandlerEvent::DelegateResponse(Err(exec_err))) => {
+                        tracing::error!(
+                            client_id = %client_id,
+                            request_id = %request_id,
+                            delegate = %delegate_key,
+                            error = %exec_err,
+                            phase = "error",
+                            "Delegate execution failed"
+                        );
+                        return Err(Error::Executor(exec_err));
                     }
                     Err(err) => {
                         tracing::error!(
--- a/crates/core/src/client_events.rs
+++ b/crates/core/src/client_events.rs
@@ -2490,6 +2490,7 @@ mod unsupported_request_tests {
 #[cfg(test)]
 mod delegate_request_outcome_tests {
     use super::*;
+    use freenet_stdlib::prelude::ApplicationMessage;
 
     fn key() -> DelegateKey {
         DelegateKey::new([3u8; 32], CodeHash::new([0u8; 32]))
@@ -2514,8 +2515,53 @@ mod delegate_request_outcome_tests {
         );
     }
 
+    /// The success arm must return the delegate's OWN values, not merely
+    /// something that is not an error.
+    ///
+    /// Passing only `vec![]` here would be vacuous: a mutation to
+    /// `Ok(_) => Ok(vec![])` discards every outbound message the delegate
+    /// produced and an emptiness-only assertion still passes. So this drives a
+    /// non-empty payload and asserts it arrives verbatim. `OutboundDelegateMsg`
+    /// is not `PartialEq`, hence the match on the payload rather than
+    /// `assert_eq!` on the Vec.
     #[test]
     fn successful_execution_returns_the_values() {
+        let payload = vec![7u8, 8, 9];
+        let res = delegate_request_outcome(
+            Ok(ContractHandlerEvent::DelegateResponse(Ok(vec![
+                OutboundDelegateMsg::ApplicationMessage(ApplicationMessage::new(payload.clone())),
+            ]))),
+            ClientId::next(),
+            RequestId::new(),
+            &key(),
+            Some("ApplicationMessages"),
+        );
+        let values = res.expect("a successful response must not be an error");
+        assert_eq!(
+            values.len(),
+            1,
+            "the delegate's outbound messages must be returned, not dropped"
+        );
+        // `let ... else` rather than a `match` with a catch-all: this crate
+        // denies `clippy::wildcard_enum_match_arm`, and spelling out all seven
+        // sibling variants here would need editing every time one is added,
+        // for no gain -- the test only cares about the one it asked for.
+        let OutboundDelegateMsg::ApplicationMessage(msg) = &values[0] else {
+            panic!("expected the ApplicationMessage back, got {:?}", &values[0]);
+        };
+        assert_eq!(
+            msg.payload, payload,
+            "the payload must arrive verbatim; a success arm that substitutes \
+             its own value is the mutation this pins"
+        );
+    }
+
+    /// The companion to the above, and the one closest to #5263: a delegate
+    /// that legitimately produces NO outbound messages is still a SUCCESS.
+    /// This is the regression this PR could most plausibly cause — turning a
+    /// legitimate no-op response into a client-visible error.
+    #[test]
+    fn genuinely_empty_success_stays_ok() {
         let res = delegate_request_outcome(
             Ok(ContractHandlerEvent::DelegateResponse(Ok(vec![]))),
             ClientId::next(),
@@ -2524,7 +2570,7 @@ mod delegate_request_outcome_tests {
             Some("ApplicationMessages"),
         );
         assert!(
-            res.expect("a successful response must not be an error")
+            res.expect("an empty but successful response must not be an error")
                 .is_empty(),
             "an empty Vec on a genuinely successful response is still correct — \
              only a genuine failure should become Err"
--- a/crates/core/src/contract.rs
+++ b/crates/core/src/contract.rs
@@ -596,7 +596,14 @@ where
                 iterations = iterations,
                 "Exceeded maximum contract request iterations, possible infinite loop"
             );
-            // Return whatever we accumulated so far
+            // Stays `Ok` DELIBERATELY, and this is an OPEN QUESTION, not a
+            // settled decision. Exhausting the iteration budget may be a
+            // legitimate outcome for a delegate that simply has more work than
+            // the cap allows, in which case converting it to `Err` would turn
+            // working delegates into client-visible failures. Nobody has
+            // established which it is, so #5263 deliberately did not touch it.
+            // Confirm the intent before changing this -- see the tracking
+            // issue linked from the PR.
             return Ok(accumulated_messages);
         }
 
@@ -622,8 +629,16 @@ where
                     phase = "unexpected_response",
                     "Unexpected response type from delegate request"
                 );
-                // Return whatever we accumulated so far
-                return Ok(accumulated_messages);
+                // An internal invariant violation, not a delegate outcome: the
+                // executor answered a delegate request with a variant that is
+                // not a delegate response at all. Reporting it as an empty
+                // success is the same lie #5263 removes from the failure arm
+                // below -- the client cannot tell "the delegate said nothing"
+                // from "this node is confused". Nothing legitimate reaches
+                // here, so there is no working behaviour to regress.
+                return Err(ExecutorError::other(anyhow::anyhow!(
+                    "unexpected response variant for a delegate request on {delegate_key}"
+                )));
             }
             Err(err) => {
                 // Downgrade "not found" to warn — expected during legacy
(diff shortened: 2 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #5287

### Summary
- **PR Title:** fix: propagate delegate execution failures to the client instead of a fake empty success
- **Type:** fix (+ test follow-up commit)
- **CI Status:** pending on latest push (6e9518d1b) — Conventional Commits / Fmt / Clippy / Claude Rule Review / macOS bundle-swap E2E in flight; Rule Lint, CLA, Snyk passing. No red checks at time of review.
- **Linked Issues:** Closes #5263
- **Review tier:** Full (touches contract/delegate execution — a high-risk surface per this repo's rules)
- **Reviewers run:** freenet:code-first-reviewer, freenet:testing-reviewer, freenet:skeptical-reviewer, freenet:big-picture-reviewer, all completed. A Fable 5 lens was also requested (per standing instruction to include at least one Fable 5 pass) and attempted 3 times; all 3 attempts failed with transient `529 Overloaded` server errors before producing any findings. No external non-Claude model was attempted, per this repo's `multi-model-review.md`, which makes that opt-in only — the 4 completed lenses satisfy the Full-tier "3-5 independent blind reviewers with distinct lenses" bar on their own.

---

### Code-First Analysis
**Independent Understanding:** `ContractHandlerEvent::DelegateResponse` changes from `Vec<OutboundDelegateMsg>` to `Result<Vec<OutboundDelegateMsg>, ExecutorError>`. `handle_delegate_with_contract_requests` (contract.rs) now returns `Err(err)` on a genuine delegate execution failure instead of silently returning whatever it had accumulated (often empty). The one exception, `is_missing_delegate()` (a legacy migration probe), stays a benign `Ok(accumulated_messages)`, unchanged. Both call sites (client-driven in `handle_contract_event`, notification-driven in `handle_delegate_notification`) and the `client_events.rs` consumer were updated to thread the `Result` through, so a genuine failure now reaches the client as `Err(Error::Executor(...))`. A follow-up commit extracts that `client_events.rs` mapping into a standalone `delegate_request_outcome` function with direct unit tests.

**Stated Intent:** Matches exactly — verified against the PR body and commit messages.

**Alignment:** Full match. The code-first reviewer independently traced every return path in the loop (5 sites) and confirmed each is correctly typed, and independently re-ran the mutation-test proof (revert `return Err(err)` → `return Ok(accumulated_messages)`, confirm the regression test fails, restore, confirm it passes).

**Gaps:** None of substance in the core logic. Two informational observations (see Big Picture / Documentation below).

---

### Testing Assessment
**Coverage Level:** Adequate, now — a gap identified mid-review has been closed in a follow-up commit.

| Unit (contract.rs) | ✅ | `contract::hol_4391_tests::delegate_request_surfaces_execution_failure_5263` drives a `DelegateRequest` through `handle_contract_event` against `MockWasmContractHandler` (whose mock executor unconditionally returns a genuine, non-`is_missing_delegate` error), asserting `DelegateResponse(Err(_))`. |
| Unit (client_events.rs) | ✅ (added this pass) | `client_events::delegate_request_outcome_tests` (4 tests) drives the extracted `delegate_request_outcome` function directly, covering: genuine execution failure → `Err(Error::Executor(_))`; success → values returned; channel-level `ContractError` → `Err(Error::Contract(_))`; wrong response variant → `Err(Error::Op(OpError::UnexpectedOpState))`. |
| Integration | N/A | Not added; judged unnecessary given the unit coverage now spans both layers (contract.rs's internal `Result` AND client_events.rs's client-facing mapping). |
| Simulation / E2E | N/A | Not applicable — this is a response-plumbing fix, not a network-behavior change. |

**Regression Test:** Present and valid. Both new/updated test suites were verified per this repo's rule (write test → confirm it fails without the fix → restore → confirm it passes), independently by the PR author and by the code-first reviewer for the original test, and by the author for the follow-up client_events.rs tests.

**Originally Missing, Now Fixed:** The testing reviewer's Important finding #1 — nothing proved `client_events.rs::process_open_request` itself propagates the error to the caller, only that `contract.rs`'s internal `Result` was correct, and the consuming match had a wildcard `Ok(_) => Err(Error::Op(OpError::UnexpectedOpState))` arm that meant deleting the new error arm would still compile and still return *an* error (just the wrong, uninformative one) — a regression only a test would catch. **Fixed**: the mapping was extracted into `delegate_request_outcome` and directly unit-tested (commit 6e9518d1b).

**Still Open, Lower Risk (Consider, not blocking):**
- Testing reviewer's finding #2: no test for the `is_missing_delegate()` probe path staying `Ok` (would require making `MockWasmRuntime`'s delegate error configurable, currently hardcoded to always error).
- Testing reviewer's finding #3: no test for the notification-driven call site's new early-return specifically (pre-existing coverage gap for `handle_delegate_notification` generally — the PR added a branch to it but didn't create the gap).

---

### Skeptical Findings
**Risk Level:** Low (no blocking bugs found across 4 independent adversarial/code-first passes plus author mutation-testing)

| Mid-loop `Err` discards already-accumulated messages | Medium (untested edge case, not a bug) | `crates/core/src/contract.rs:648` (the `Err(err) => ... return Err(err);` arm) | If a multi-round-trip delegate flow accumulates real `ApplicationMessage`/`ContextUpdated` output in an early iteration, then fails on a *later* iteration, the fix discards that partial data and returns only `Err`. Before the fix, the client got the partial data mislabeled as success — arguably worse. The linked issue #5263 floated preserving partial results as optional ("could"), so the current behavior isn't a violation of the issue's ask, but it's an unverified, undocumented broadening of the fix's scope beyond the single-iteration case the production incident and the regression test actually cover. Confirmed independently by both the code-first and skeptical reviewers. |
| Notification-path comment overstates behavioral equivalence | Medium (accuracy of a comment, not a functional bug) | `crates/core/src/contract.rs:2143-2148` (`let Ok(outbound) = outbound else { return; };`) | The inline comment claims this matches "the pre-existing behavior of falling through with nothing to route." True only when `accumulated_messages` was empty at the point of failure (the dominant case, including the production incident). In the multi-iteration-then-fail case, pre-fix code would still have reached `delegate_app_registry::route_to_apps` with the partial data; post-fix it's dropped before ever reaching that block. Same underlying tradeoff as the row above — not inaccurate about the common case, just imprecise about an untested edge case. |
| `ExecutorError`'s `Display` text now reaches the client, unredacted, for the first time on the delegate path | Low | `crates/core/src/client_events.rs` (`Error::Executor(exec_err)` → `ErrorKind::OperationError { cause: format!("{err}") }`) | Checked `freenet-stdlib` 0.8.5's `DelegateError` `Display` impls — they expose identifiers/keys and free-text execution-error strings, not raw secret material. This exact exposure shape is already precedented by the pre-existing `GetResponse` path (`client_events.rs:945`), so it's not a new *class* of exposure, just the first time it applies to delegates specifically — worth a second look given delegates hold per-user secrets, but not a regression this PR introduces. |
| CI checks (unrelated to diff content) | Informational | — | At initial review time, one CI check (`Windows Service Unit` install-nextest step) had failed and looked like an installer/infra flake unrelated to this diff (no Windows-specific code touched). After the follow-up push, CI is re-running fresh (see Summary). Per this repo's no-merge-with-red-CI rule, confirm all required checks are green on the current HEAD SHA before merging. |

**Open question investigated during synthesis — `DelegateError::MissingSecret`:** the skeptical reviewer asked whether a delegate legitimately probing for an optional secret on first run could now surface as a spurious hard client error. Traced the code: `MissingSecret` is constructed at `crates/core/src/contract/executor.rs:409-420`, reached only via `RuntimeInnerError::SecretStoreError(MissingSecret)` bubbling out of `ExecutorError::execution` at `delegates.rs:501` — i.e. only when a delegate's own WASM `process()` (V1 request/response API) fails to handle a missing secret gracefully inside its own logic. The newer V2 host-function secret API (`native_api.rs::get_secret`/`get_secret_len`) returns an error *code* directly to the guest, so a V2 delegate that checks for "not found" never reaches this path at all. `MissingSecret` was *already* being silently swallowed before this PR (same as every other non-`is_missing_delegate` failure) — this PR doesn't add or remove any special-casing for it, it just stops swallowing the general case. There's no existing precedent (test, comment, or prior carve-out) establishing `MissingSecret` as an expected/routine condition analogous to the `is_missing_delegate()` migration-probe carve-out. Given no evidence this is a routine occurrence in production, adding a speculative new carve-out now would be exactly the kind of unevidenced hardening this repo's own rules caution against. Recommend: **Consider** — worth a follow-up if a real diagnostic report ever surfaces this class of failure, same as how `is_missing_delegate()` itself was presumably added.

---

### Big Picture Assessment
**Goal Alignment:** Yes — the fix does exactly what #5263 asks, with the exempted migration-probe case preserved unchanged.
**Anti-Patterns Detected:** None. No `#[ignore]`, no weakened assertions, no removed tests, no CI-chasing.
**Removed Code Concerns:** None — purely additive/restructuring diff; every `_`-wildcard consumer of `ContractHandlerEvent::DelegateResponse` elsewhere in the codebase was confirmed unaffected by the type change (they don't inspect the payload).
**Scope Assessment:** Focused. The follow-up test-coverage commit is a proportionate, low-risk pure extraction (no behavior change), not scope creep.
**AI-instructions / rules file:** No `.claude/rules/bug-prevention-patterns.md` row is a close match. Given this is the second independently-reported real-world case of "a fallible call's error arm silently defaults to `Ok`/empty on a client-facing response path" (the PR body cites both the GHSA-824h Freebird case and a second independent Freebird report), a new row may be worth adding as a follow-up — not blocking this PR.

---

### Documentation
- Code docs: Minor gap — `handle_delegate_with_contract_requests`'s top-level doc comment doesn't mention it can now return `Err` on genuine failure, or that a later-iteration failure discards prior `accumulated_messages`. Flagged independently by both code-first and big-picture reviewers as a nice-to-have, not a blocker.
- Two pre-existing "fake success" paths remain out of scope and undocumented as known gaps: `contract.rs:592-600` (iteration-limit exceeded) and `contract.rs:618-626` (unexpected response variant) — both still silently return `Ok(accumulated_messages)`. Narrower/rarer than #5263's actual failure mode; worth naming as known gaps (in a comment or a tracked issue) so they aren't rediscovered later as "the same bug."
- Architecture / user docs: n/a, no public API or config surface changed.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None remaining — the one Important-severity gap (client_events.rs error-propagation coverage) was fixed during this review pass (commit 6e9518d1b).

#### Consider (Suggestions)
1. Add a doc-comment note on `handle_delegate_with_contract_requests` covering the new `Err` return and the later-iteration partial-accumulation-discard behavior.
2. Consider a follow-up test for the `is_missing_delegate()` probe path staying `Ok` (requires making `MockWasmRuntime`'s delegate error configurable).
3. Consider naming the two untouched "fake success" paths (iteration-limit, unexpected-variant) as known gaps, either inline or as a tracked follow-up issue.
4. Consider whether `DelegateError::MissingSecret` should get the same benign-`Ok` treatment as `is_missing_delegate()` — deferred pending real evidence of it being a routine, expected condition (see investigation above).
5. Consider adding a `.claude/rules/bug-prevention-patterns.md` row for "a fallible call's error arm defaults to `Ok`/empty on a client-facing response path" given two independent real-world reports of this exact shape.

---

### Verdict
**State:** Ready to Merge, contingent on CI going green on the current HEAD SHA (6e9518d1b) — no code-level blockers found across 4 independent review lenses plus author-performed mutation testing on both the original fix and the follow-up test-coverage commit.
**HEAD SHA reviewed:** 6e9518d1b (includes both the original fix commit and the follow-up client_events.rs test-coverage commit)

CI was re-triggered by the follow-up push and was still running at review time (Fmt/Clippy/Conventional-Commits/Claude-Rule-Review/macOS-E2E pending, Rule-Lint/CLA/Snyk passing, nothing red). Confirm all required checks pass on this exact SHA before merging, per this repo's no-merge-with-red-CI rule. The Consider-tier suggestions above are optional polish, not merge blockers.

[AI-assisted - Claude]

sanity (commented):
## Full-tier review — PR #5287

Reviewed at head `64fd1f5cc`. This PR had no Full-tier review (the lane driving it hit a quota limit before running one, and the automated rule-review reported "Failed to run"), so this is the first.

**Lenses run:** `skeptical` (adversarial bug hunt), `code-first` (read the code before the description, flag intent/impl mismatches), `testing` (coverage quality, vacuous-pin audit). Each ran blind to the others against a dedicated read-only worktree. No external model pass; three Claude lenses were substituted per the review rule's availability clause. Every `file:line` below was re-verified against the checked-out code by the synthesizer; findings that did not survive that check are listed at the end.

---

### BLOCKING

**1. `crates/core/src/contract.rs:3097-3125` — a rebase detached a security pin's doc block and stapled it onto the new test. This is also the red CI check.**

Two lenses found this independently, and it is invisible in the diff because both blocks are `+` lines adjacent to unchanged context.

On `origin/main`, the `/// H2 — the control that REPLACED the deleted registration-record gate` block documents `inter_delegate_hop_forwards_the_originating_scope`. The two new tests were inserted between that block and its function. Today:

- lines 3097-3110 (the H2 rationale — why the inter-delegate hop must forward the originating `connection_scope` rather than a literal) now document `unexpected_response_variant_does_not_become_a_fake_success` at `:3125`, running straight into that test's own doc at `:3111` with no separator;
- `inter_delegate_hop_forwards_the_originating_scope` at `:3212` is left with **no doc comment at all**, losing the recorded reasoning for a security pin.

The Clippy failure is the same defect: `error: doc list item without indentation` at `contract.rs:3111`, `clippy::doc_lazy_continuation` under `-D warnings`. Line 3111 is read as a lazy continuation of H2's numbered list item 2 at `:3109-3110`. This is a genuine lint in this PR's own added code, not staleness — the 14 commits main has moved since the run touch none of the four changed files.

Fix: move `:3097-3110` back down to immediately above `:3211`. That resolves the finding and the red check together.

---

### IMPORTANT

**2. `crates/core/src/contract.rs:650-655` — the `is_missing_delegate()` guard is untestable from any fixture in the tree, in both directions.**

This is the branch the fix newly introduced, and it is the one deciding whether an error becomes client-visible. Both `MockWasmRuntime` (`executor/mock_wasm_runtime.rs:402`) and `MockRuntime` (`executor/mock_runtime.rs:751`) return `ExecutorError::other(anyhow!(...))`, which is `Either::Right`. `is_missing_delegate()` (`executor.rs:725-732`) matches only `Either::Left(RequestError::DelegateError(Missing(_)))`, so it is unreachable from any fixture.

Consequence: **delete the guard and every test in this PR still passes**, while legacy migration probes start returning client-visible errors — the exact regression the guard's own comment says it exists to prevent. `MockWasmRuntime` already has an override mechanism (`ValidateOverride`, `UpdateOverride`); a `DelegateOverride` returning `ExecutorError::request(RequestError::DelegateError(Missing(..)))` closes this and finding 3 in one fixture.

**3. `crates/core/src/client_events.rs:2518` — `successful_execution_returns_the_values` does not test what its name says.**

It only ever passes `vec![]` and only asserts `.is_empty()`. Mutating the success arm to `Ok(_) => Ok(vec![])` passes it. Relatedly, `contract.rs:712` has no test proving a genuine empty success stays `Ok` at the *contract* layer at all, because both mocks fail unconditionally — so the regression this change could most plausibly cause (a legitimate no-op delegate response becoming an `Err`) has no coverage where the classification is actually made.

One-line fix: pass a non-empty `Vec<OutboundDelegateMsg>` and `assert_eq!` on it, keeping a separate empty-vec case.

**4. The failure is flattened to `ErrorKind::OperationError`, which the delegate rate limiter does not see, and which makes error text load-bearing across a repo boundary.**

`client_events.rs:509` returns `Err(Error::Executor(exec_err))`; the catch-all at `client_events.rs:348-355` renders it as `ErrorKind::OperationError { cause: format!("{err}") }`. But a delegate execution failure is usually already a *typed* request error — `executor.rs:433-435` maps `RuntimeInnerError::DelegateExecError` to `ExecutorError::request(DelegateError::ExecutionError(..))`, and `is_request()`/`unwrap_request()` exist to recover it (`node.rs:5878-5879` does exactly that).

Two consequences:

- The #3305 delegate rate limiter records only for `ErrorKind::RequestError(RequestError::DelegateError(_))` (`websocket.rs:2058-2075`); these failures land in `_ => {}`. To be fair to the change, this is **not a regression** — before the PR the fake success hit `record_success(key.bytes())` at `:2059`, which was worse. But #5263 is a failure a delegate repeats on *every* invocation, which is precisely what the backoff was built to damp, and this was the opportunity to connect it.
- Error **text** becomes the interface. River's synchronizer string-matches on it: `contains("delegate") && contains("not found")` triggers `request_legacy_seal_on_quiescence()`, a **permanent one-way** migration seal (`river/main/ui/.../freenet_synchronizer.rs:757-777`); `contains("not supported")` triggers a sticky compatibility error state. No currently-produced string trips either — `DelegateError::Missing` renders "missing delegate {key}", not "not found" — so this is a hazard, not a proven break. Nothing pins it, and a delegate author's own error text now flows straight into that matcher.

Suggested fix: mirror `node.rs:5878` — if `exec_err.is_request()`, return the typed `RequestError`.

**5. The disclosed "no end-to-end test" gap is weaker than the PR argues, because the harness already exists.**

The compiler-checked-shared-type argument is structurally sound for the layers it names, and I verified it: the `Result` in `contract/handler.rs:825` cannot be re-flattened without a type error, and `client_events.rs:2354` proves *some* `Err` from `process_open_request` reaches a proxy's `send()`. What defeats it is cost. `crates/core/tests/origin_attestation_scope.rs` boots a real node with the real executor and WASM runtime, publishes the real `test-delegate-attested` WASM, and drives it over a real WebSocket with `WebApi`; its helper `attested_origin_via` (`:206`) already sends `DelegateRequest::ApplicationMessages` and asserts on `HostResponse::DelegateResponse { values }` at `:255-256`. And that delegate returns `DelegateError::Other("deserialize inbound: …")` on a payload it cannot decode (`tests/test-delegate-attested/src/lib.rs:29-30`) — a genuine execution failure, *not* `is_missing_delegate`, i.e. exactly the #5263 shape.

So: send `ApplicationMessage::new(vec![0xff; 8])` and assert `client.recv()` yields `Err(_)` rather than `DelegateResponse { values: [] }`. Pre-fix that fails; post-fix it passes. One leg in an already-booted node, no new boot. A `run_simulation_direct` test is *not* warranted here — delegates are not exercised by that harness (same failing mocks); the real-node WS test is both cheaper and more direct.

---

### MINOR

**6. `contract.rs:663` and `:2173-2175` — partial delegate output is now discarded, undisclosed.** `accumulated_messages` accrues `ApplicationMessage`/`ContextUpdated` across loop iterations (`:696-700`) and the loop re-executes after servicing contract requests. A delegate that emits an `ApplicationMessage` on iteration 1 and fails on iteration 2 previously had that message delivered as a success; it is now dropped, and on the notification path it is no longer fanned out to registered apps. Defensible, but it is a second behavior change and the PR body describes only the first-iteration case. Note also that the error now implies "nothing happened" while earlier-iteration PUT/UPDATE/SUBSCRIBE side effects have already committed and are not rolled back.

**7. The blast radius is wider than "delegate execution failure".** The new `Err` channel also carries paths that previously reached the client as empty successes: `runtime/delegates.rs:355` and `:370` (over-cap predecessor list on `RegisterDelegateWithPredecessors` — a client previously saw registration *succeed*), `:323` (failing `RegisterDelegate`), and `:521` (`_ => Err(ExecutorError::other("not supported"))`, the catch-all for future variants). All improvements, but worth a line in the description given #5392 just did adjacent work on unsupported-operation errors.

**8. `contract.rs:2012` and `:2062-2066` — `send_queue_full_response`'s comments are now factually wrong.** They still list `DelegateRequest` under "Events without error response variants" and say "For fire-and-forget events (delegates, disconnects), no response is sent." After this PR `DelegateRequest` *does* have a typed error response variant, exactly like `ExportUserSecrets` at `:2055`. Behavior stays safe (`NoEvHandlerResponse` → an error), but the typed `ContractQueueFull` marker is discarded, so a delegate request shed by backpressure is indistinguishable from a genuine delegate failure — the very distinction this PR is about. Worth asking whether `DelegateResponse(Err(ContractQueueFull))` should now be sent.

**9. Three ERROR lines per failed request.** `contract.rs:657-662`, then `client_events.rs:501-508`, then `client_events.rs:349-353`. Before, only the first fired. The motivating failure recurs on every invocation and this crate builds with `release_max_level_info`, so all three ship. Suggest demoting `client_events.rs:501` to `debug!` or dropping the `contract.rs:657` line now that the error propagates.

**10. `contract.rs:524-535` — rustdoc not updated for the new return type.** Still ends "Returns the final response with contract request messages filtered out", with no `# Errors` section, while `:546` now returns `Result<_, ExecutorError>`.

**11. `contract.rs:3179` — the TTL pin asserts textual order, not conditionality.** It correctly fails for the named regression (sweep moved below the bail) and fails closed if either anchor is renamed. It does *not* fail if someone writes `if outbound.is_ok() { sweep_expired(); }` while leaving it textually above — which reintroduces the exact defect. A behavioural test is awkward (`delegate_app_registry` is process-global; see the cross-test-interference rule), so a source assertion is defensible, but the gap is worth naming in the rustdoc.

**12. NIT — `client_events.rs:470-471`** claims the extraction logs "exactly as the inline match this was extracted from did." True for three arms; the `DelegateResponse(Err(_))` arm at `:500-509` is new and had no predecessor. **`contract.rs:1060`** still has `Ok(_) => {}` on the inter-delegate hop, silently absorbing the wrong-variant response that `:638` now treats as an invariant violation — fire-and-forget, so arguably out of scope, but the asymmetry deserves a comment or a tracking issue.

---

### Checked and found clean (no action)

- **The iteration-limit arm is correctly framed.** `contract.rs:599-606` reads verbatim as an OPEN QUESTION, names the concrete cost of guessing wrong ("converting it to `Err` would turn working delegates into client-visible failures"), says who has to decide, and carries #5454. This is the right shape; not a finding. (Small wording nit: "#5263 deliberately did not touch it" points a reader at the bug report rather than the decision.)
- **Vacuous-pin audit (#5450): both source pins are sound**, verified mechanically against the file bytes rather than by reading their doc comments. `full.find("\nmod tests {")` lands at line 2977 — the only column-0 occurrence; the apparent hits at 3128/3182 are the escaped two-character `\n`. Both scrapes bound correctly to their target function bodies and fail under the deletion mutation they name.
- **No missed `ContractHandlerEvent::DelegateResponse` site.** All 16 occurrences repo-wide checked; the wildcard matches (`operations.rs:771`, `operations/update.rs:893`, `fair_queue.rs:652`, `contract.rs:2067`, `:2875`, `handler.rs:991`) are structurally payload-oblivious (queue routing, `Display`, `unreachable!`). No catch-all silently absorbs the new shape; nothing outside `crates/core/src` constructs the variant.
- **No new wire variant, and old clients survive.** `ErrorKind::OperationError` exists at the same variant index back to stdlib 0.6.1, so an old riverctl/River UI deserializes it. `should_rm` at `websocket.rs:2307-2311` is false for `OperationError`, so the connection is **not** dropped, and River's handler treats an unmatched API error as log-only — no panic, no reconnect storm. (Caveat under finding 4 about error *text*.)
- **`is_missing_delegate()` is genuinely typed and tight** (`executor.rs:725-732`), sole constructor at `:428-429` reached only from delegate-store lookup misses; the pooled executor passes it through untouched. A real execution failure cannot be misread as missing. (It is the *testing* of it that finding 2 objects to, not the matcher.)
- **The sweep really does precede the bail in production code** (`contract.rs:2171` vs `:2173`), not merely in the assertion.
- **#5392 survived the rebase intact** — `mod unsupported_request_tests` is byte-identical to `origin/main`; the only deletion in `client_events.rs` is the 45-line inline match being extracted.
- **No test weakened, deleted, relaxed, or `#[ignore]`d.** The one pre-existing test touched (`fair_queue.rs:1296`) is a mechanical fixture update in a routing table whose assertion does not read the payload. `handler.rs` is the variant type plus rustdoc.
- **The #5263 regression test bites the exact reverted line.** Reverting `contract.rs:663` to `return Ok(accumulated_messages)` makes `contract.rs:5318` fail, because the mock's `ExecutorError::other` is `Either::Right` and so skips the missing-delegate guard.

### Dropped during verification

- A claim that the H2 doc block spans `:3097-3123` — the block ends at `:3110`; the new test's own doc begins at `:3111`. Corrected above.
- A concern that `ErrorKind::OperationError` carries no request correlation, so an app with several in-flight delegate requests still cannot tell which failed. Verified true, but pre-existing and equally true of the success side — not a regression, so not a finding against this PR.

---

**Verdict.** The core change is sound and the PR's self-description is unusually honest — every claim it makes, I checked independently and it holds. Finding 1 must be fixed: it is a real Clippy failure and it silently deletes the rationale for an unrelated security pin. Findings 2, 3 and 5 are worth closing before merge because they are cheap and they target the branches this change itself created. Findings 4 and 6-12 are legitimately follow-up material.

[AI-assisted - Claude]

sanity:
## Review status after the rebase — what the lenses actually read

The Full-tier review above ran against `64fd1f5cc`. The branch has since been rebased onto current main and two review findings fixed, so per the reviews-are-per-code-content rule that CI signal and part of that review are stale. Recording precisely what changed, so nobody has to infer it.

**Base moved** `fc9bd228a` → `af7c52e7f` (35 commits of main). **Head** `64fd1f5cc` → ``fe2185914``.

**The rebase was clean** — "Successfully rebased", 6 commits replayed, no `CONFLICT`, no mid-sequence edits. Exactly one of the 35 new commits touches any file this PR changes: `e6155dac7` (#5178, OTel metrics export), and only `crates/core/src/contract/fair_queue.rs`. Zero hunks in `contract.rs` or `client_events.rs`, so nothing landed in the region the doc block moved through — checked directly rather than inferred from the absence of a conflict, since a clean rebase means git found no textual overlap, not that a move is still semantically a move.

The PR's own delta to `fair_queue.rs` is byte-identical before and after the rebase (the one-line `DelegateResponse(vec![])` → `DelegateResponse(Ok(vec![]))` fixture update). The file's *content* differs only because main's #5178 arrived in it.

### What changed since the lenses read it

Two findings from the review, both in the test surface:

1. **BLOCKING finding 1 (the detached doc block) is fixed.** `contract.rs:3097-3110` moved back above `inter_delegate_hop_forwards_the_originating_scope`. Verified a pure move rather than a rewrite: 14 insertions, 14 deletions, and `sort | md5sum` of the file is unchanged from `64fd1f5cc` — no line added, lost, or altered. Re-checked after the rebase and it still holds. The four `include_str!` source scrapes were re-audited; `mod tests {` remains the sole column-0 occurrence and every scrape bounds on production text, so relocating lines inside the test module cannot reach them.

   This also clears the red **Clippy** check. It was a genuine lint in this branch's code — `clippy::doc_lazy_continuation` at `contract.rs:3111`, the new test's doc reading as a lazy continuation of H2's numbered list — not staleness.

2. **IMPORTANT finding 3 (the vacuous success test) is fixed.** `successful_execution_returns_the_values` passed only `vec![]` and asserted only `.is_empty()`, so the `Ok(_) => Ok(vec![])` mutation passed it. It now drives a non-empty `ApplicationMessage` and asserts the payload arrives verbatim. The empty-vec case is split out as `genuinely_empty_success_stays_ok` rather than deleted — it guards the over-correction (a legitimate no-op response becoming a client-visible error), which is what #5454 asked for.

**No production code changed.** Both fixes are confined to `#[cfg(test)]` and doc comments; the diff to every non-test path is identical to what the three lenses read. So the review's substantive findings stand as written against the current head, and the parts of it that examined production behaviour do not need re-running.

The remaining findings (the unpinned `is_missing_delegate` carve-out, the cheaper end-to-end path through `origin_attestation_scope.rs`, and the `ErrorKind::OperationError` flattening with its River string-match coupling) are recorded on #5454 rather than grown into this PR.

[AI-assisted - Claude]

PR #5289, opened 2026-08-12
fix: put freenet/fdev on PATH after macOS DMG install

sanity:
## Problem

Installing Freenet via the DMG (the featured macOS method on freenet.org/quickstart) leaves `freenet` and `fdev` unreachable from Terminal. The DMG only ever produces a drag-to-Applications `.app` bundle — there's no `.pkg`/postinstall step to set up PATH — and `fdev` wasn't even bundled into the DMG at all. Only the separate `install.sh` script handled PATH, and even that only *prints* instructions rather than applying them.

## Approach

- **Bundle `fdev` into the DMG** alongside `freenet` (`package-macos.sh`, `cross-compile.yml`, `macos-dmg-test.yml`), gated on optional `FREENET_ARM64_FDEV_BIN`/`FREENET_X86_FDEV_BIN` so local `.app` smoke-tests without fdev binaries still work unchanged.
- **On first wrapper launch, automatically symlink `freenet`/`fdev` into `/usr/local/bin`** (on `PATH` by default for macOS login shells, unlike `~/.local/bin`) via a new `service::cli_symlinks` module:
  - Unprivileged symlink creation is tried first.
  - A single administrator-privileges `osascript` prompt (AppleScript's `quoted form of` performs the shell quoting, so there's no hand-rolled shell escaping) covers whatever still needs it.
  - One-shot, gated on its own marker — not the existing first-run/legacy-migration markers, for the same reason those are kept separate: an already-onboarded user still needs this to run once on their next launch.
  - Skips (without writing the marker) when launched from Gatekeeper's App Translocation mount — the common "double-click inside the DMG" first-launch pattern — since a symlink into that ephemeral path would dangle; retries on a later, non-translocated launch.
  - Never repoints a symlink/file it doesn't recognize as its own: requires both the app-bundle path shape and a matching tool basename before treating an existing link as "ours to repair."
  - A declined/failed elevated install surfaces as a desktop notification, not just a log line a GUI menu-bar app user would never see.
  - Runs on its own background thread so a slow or declined password prompt never delays the tray icon appearing.
- Split into a pure decision core (`decide_cli_symlink_action`, `build_elevated_symlink_script`, `is_translocated_path`) and a macOS-only execution wrapper, per `.claude/rules/deployment.md`'s cfg'd-platform-path pattern, so the logic is unit-tested on Linux CI rather than only being "verified" by compiling.

## Review

Reviewed with 5 independent lenses (code-first, testing, skeptical, big-picture, and a Fable 5 adversarial pass) before this PR's second commit. Findings addressed:
- **Must Fix**: App Translocation permanently breaking the fix (see above) — fixed.
- **Should Fix**: the "safe to repoint" check was too broad (matched any app bundle, not the same tool) — now also requires basename match. Residual risk (a same-named binary inside some *other* app's bundle) is accepted as bounded: the symlink *target* is always our own trusted binary, never attacker- or foreign-app-influenced, so at worst this reclaims a path claim, never substitutes code. Full bundle-identity verification via `Info.plist` is left as a follow-up.
- **Should Fix**: the one-shot marker was written even when `osascript` failed to spawn (not just when the user declined) — fixed, a spawn failure now retries next launch.
- **Should Fix**: declined/failed installs were invisible to a GUI user — added a desktop notification.
- Test gaps: added a marker-distinctness test, a real backslash-escaping test (the prior test's name promised this but its fixture had no literal backslash), and coverage for the two behavior changes above.

Deferred as follow-up (not blocking): an orphaned admin-privileges dialog if the app quits mid-prompt; a genuine E2E test for `macos_ensure_cli_symlinks` itself, matching the `macos-bundle-swap-e2e.yml` precedent for this exact class of macOS-only side effect; `package-macos.sh` shellcheck coverage and a `test-install-sh.sh`-style harness for its new branch.

## Testing

18 unit tests in `service/cli_symlinks.rs` plus a new marker-distinctness test in `service.rs`. Full `commands::service` suite (123 tests) green. `cargo fmt` + `cargo clippy -D warnings` clean.

The macOS-only execution path itself is not exercised by any CI in this PR — `macos-dmg-test.yml` verifies DMG *packaging* (that `fdev` lands correctly in the bundle), not the first-launch symlink *behavior*. Runtime verification is deferred to manual testing after the next release, same as the pre-existing `launch_at_login.rs` first-launch macOS-only paths this PR follows the pattern of.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/cross-compile.yml
+++ b/.github/workflows/cross-compile.yml
@@ -273,6 +273,18 @@ jobs:
           name: binaries-x86_64-macos-freenet
           path: bin-x86_64
 
+      - name: Download ARM64 macOS fdev binary
+        uses: actions/download-artifact@v8
+        with:
+          name: binaries-arm64-macos-fdev
+          path: bin-arm64
+
+      - name: Download x86_64 macOS fdev binary
+        uses: actions/download-artifact@v8
+        with:
+          name: binaries-x86_64-macos-fdev
+          path: bin-x86_64
+
       - name: Install packaging dependencies
         run: |
           brew install create-dmg librsvg
@@ -336,12 +348,14 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
           FREENET_ARM64_BIN: bin-arm64/freenet
           FREENET_X86_BIN: bin-x86_64/freenet
+          FREENET_ARM64_FDEV_BIN: bin-arm64/fdev
+          FREENET_X86_FDEV_BIN: bin-x86_64/fdev
           CODESIGN_IDENTITY: "Developer ID Application: Ian CLARKE (R55ZESJCXG)"
           ASC_API_KEY_ID: ${{ secrets.APPLE_ASC_API_KEY_ID }}
           ASC_API_ISSUER_ID: ${{ secrets.APPLE_ASC_API_KEY_ISSUER_ID }}
           # ASC_API_KEY_PATH set by the previous step via $GITHUB_ENV.
         run: |
-          chmod +x bin-arm64/freenet bin-x86_64/freenet
+          chmod +x bin-arm64/freenet bin-x86_64/freenet bin-arm64/fdev bin-x86_64/fdev
           ./scripts/package-macos.sh
 
       - name: Clean up secrets from runner
--- a/.github/workflows/macos-dmg-test.yml
+++ b/.github/workflows/macos-dmg-test.yml
@@ -50,12 +50,16 @@ jobs:
           gh release download "$TAG" --repo "${{ github.repository }}" \
             --pattern 'freenet-aarch64-apple-darwin.tar.gz' \
             --pattern 'freenet-x86_64-apple-darwin.tar.gz' \
+            --pattern 'fdev-aarch64-apple-darwin.tar.gz' \
+            --pattern 'fdev-x86_64-apple-darwin.tar.gz' \
             --dir download
           mkdir -p bin-arm64 bin-x86_64
           tar -xzf download/freenet-aarch64-apple-darwin.tar.gz -C bin-arm64
           tar -xzf download/freenet-x86_64-apple-darwin.tar.gz  -C bin-x86_64
-          chmod +x bin-arm64/freenet bin-x86_64/freenet
-          file bin-arm64/freenet bin-x86_64/freenet
+          tar -xzf download/fdev-aarch64-apple-darwin.tar.gz    -C bin-arm64
+          tar -xzf download/fdev-x86_64-apple-darwin.tar.gz     -C bin-x86_64
+          chmod +x bin-arm64/freenet bin-x86_64/freenet bin-arm64/fdev bin-x86_64/fdev
+          file bin-arm64/freenet bin-x86_64/freenet bin-arm64/fdev bin-x86_64/fdev
 
       - name: Install packaging dependencies
         run: brew install create-dmg librsvg
@@ -104,6 +108,8 @@ jobs:
           VERSION: ${{ steps.tag.outputs.tag }}
           FREENET_ARM64_BIN: bin-arm64/freenet
           FREENET_X86_BIN: bin-x86_64/freenet
+          FREENET_ARM64_FDEV_BIN: bin-arm64/fdev
+          FREENET_X86_FDEV_BIN: bin-x86_64/fdev
           CODESIGN_IDENTITY: "Developer ID Application: Ian CLARKE (R55ZESJCXG)"
           ASC_API_KEY_ID: ${{ secrets.APPLE_ASC_API_KEY_ID }}
           ASC_API_ISSUER_ID: ${{ secrets.APPLE_ASC_API_KEY_ISSUER_ID }}
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -1383,6 +1383,23 @@ mod tests {
     }
 
     #[test]
+    fn onboarding_markers_are_distinct_files() {
+        // The whole point of splitting cli_symlinks_marker_path (and
+        // legacy_migration_marker_path) out from first_run_marker_path is
+        // that an already-onboarded user — who has the first-run marker set
+        // — still needs the CLI-symlink one-shot (and the legacy migration)
+        // to run once on their next launch. If a future edit accidentally
+        // collapsed these onto the same path, that guarantee would silently
+        // break and this is the only thing that would catch it.
+        let first_run = first_run_marker_path().unwrap();
+        let legacy_migration = legacy_migration_marker_path().unwrap();
+        let cli_symlinks = cli_symlinks_marker_path().unwrap();
+        assert_ne!(first_run, legacy_migration);
+        assert_ne!(first_run, cli_symlinks);
+        assert_ne!(legacy_migration, cli_symlinks);
+    }
+
+    #[test]
     #[cfg(target_os = "linux")]
     fn test_systemd_user_service_file_generation() {
         let binary_path = PathBuf::from("/usr/local/bin/freenet");
--- a/crates/core/src/bin/commands/service.rs
+++ b/crates/core/src/bin/commands/service.rs
@@ -11,12 +11,14 @@
 //! - `log_utils` – log file rotation helpers
 //! - `single_instance` – macOS wrapper single-instance lock (flock-based)
 //! - `launch_at_login` – macOS Launch at Login, plist helpers, legacy migration
+//! - `cli_symlinks` – macOS first-launch `freenet`/`fdev` PATH symlink setup
 //! - `wrapper` – process wrapper loop, state machine, log events
 //! - `purge` – data purge, doctor, process-reaping
 //! - `linux` – Linux/systemd service management
 //! - `macos` – macOS/launchd service management
 //! - `windows` – Windows registry/task service management
 
+mod cli_symlinks;
 mod launch_at_login;
 mod linux;
 mod log_utils;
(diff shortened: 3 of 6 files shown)


sanity (commented):
## Comprehensive PR Review: #5289

### Summary
- **PR Title:** fix: put freenet/fdev on PATH after macOS DMG install
- **Type:** fix
- **CI Status:** was red on `Simulation`/`Unit & Integration` at commit 46292c44e (confirmed: transient `nextest` install network flake — `tar: Error is not recoverable` — at a step this PR's diff never touches; unrelated recent `main` runs were green). Confirmed green on retry at commit 14095ba5d. Final commit 99b6d4dce's CI run was in progress at review time.
- **Linked Issues:** none (author is the repo maintainer; not a CONTRIBUTING.md process gap, but noted below)
- **Review tier:** Full (deploy/release/CI-config surface, plus code that runs a privileged shell command)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all Claude), plus a Fable 5 adversarial pass — first round, on commit 46292c44e. Same three lenses (skeptical, Fable, testing) re-run on the fix commit 14095ba5d after Must/Should-Fix findings were addressed. External (Codex/Gemini) review not attempted, per this repo's `multi-model-review.md` (external models are opt-in, not default; the Fable 5 lens was run instead per explicit user request).

---

### Round 1 Findings (commit 46292c44e) — all addressed in commits 14095ba5d and 99b6d4dce

| Finding | Severity | Location | Status |
|---|---|---|---|
| **App Translocation**: launching Freenet.app directly from the mounted DMG (the common "double-click inside the DMG" first-launch pattern, before dragging to /Applications) runs it from Gatekeeper's randomized, ephemeral `AppTranslocation` mount. The old code symlinked into that ephemeral path, then wrote the one-shot marker unconditionally — permanently breaking the fix even after the user later did the right thing. | **Must Fix** | `cli_symlinks.rs` (`macos_ensure_cli_symlinks`) | **Fixed** — new `is_translocated_path` check, skips without writing the marker. Verified: check runs before any symlink/marker logic, no fall-through to the marker write. |
| `decide_cli_symlink_action`'s "safe to repoint an existing symlink" check matched any path under some `.app/Contents/MacOS/`, not specifically the same tool — a stale link happening to sit inside an unrelated app's bundle could be silently repointed. | Should Fix | `cli_symlinks.rs:112-120` | **Fixed** — now also requires basename match. Residual gap (same-basename binary inside a *different* app) explicitly accepted and now pinned by `same_tool_different_app_bundle_is_still_repaired_by_design`: bounded risk since the symlink *target* is always our own trusted binary, never attacker- or foreign-app-influenced — reclaims a path claim, never substitutes code. Full bundle-identity verification via `Info.plist` deferred as follow-up. |
| One-shot marker was written even when `osascript` itself failed to spawn (not just when the user declined the prompt) — a transient environment hiccup permanently disabled the feature. | Should Fix | `cli_symlinks.rs` (elevated-install branch) | **Fixed** — `osascript_spawn_failed` flag gates the marker write; traced exhaustively by two independent re-reviewers, no leak found. |
| A declined/failed elevated install was only logged to a file a GUI menu-bar-app user has no way to see. | Should Fix | `cli_symlinks.rs` | **Fixed** — added `notify_cli_symlink_declined`, a desktop notification via the same `osascript display notification` mechanism as the existing `notify_stuck_wrapper`. |
| PR description overstated `macos-dmg-test.yml`'s coverage (claimed it gives the new runtime behavior "real-world coverage" — it only verifies DMG *packaging*, never launches the app or checks PATH). | Should Fix (docs accuracy) | PR description | **Fixed** — description corrected. |
| Test gap: the "escapes quotes and backslashes" test's name promised backslash coverage but its only fixture had no literal backslash. | Should Fix | `cli_symlinks.rs` tests | **Fixed** — split into separate quote/backslash tests, then (round 2) a combined fixture added since the split tests still couldn't distinguish escape-*order* regressions. |
| No test protecting the invariant that `cli_symlinks_marker_path`, `first_run_marker_path`, and `legacy_migration_marker_path` are distinct files (the entire reason this PR's marker exists as a separate file). | Should Fix | `service.rs` tests | **Fixed** — `onboarding_markers_are_distinct_files` added. |

### Round 2 Findings (re-review of commit 14095ba5d) — addressed in 99b6d4dce

| Finding | Severity | Status |
|---|---|---|
| The split quote-only/backslash-only tests each contain only ONE of the two special characters `applescript_escape` handles, so **neither can catch a regression that swaps its escape order** — both orders happen to produce a passing result when only one character is present. | Should Fix | **Fixed** — combined fixture added, mutation-tested (swapped the real escape order, confirmed the new test fails, reverted). |
| The "same tool, different app bundle" residual-risk case (accepted by design, documented in the commit message) was not pinned by a test, so a future editor could mistake the basename check for full hardening or "fix" the accepted tradeoff inconsistently. | Should Fix | **Fixed** — `same_tool_different_app_bundle_is_still_repaired_by_design` added with an explanatory comment. |
| Skeptical re-review, code-first (round 1), and Fable (round 2) independently confirmed no new bugs were introduced by the fix commit, and that all four fixes are logically sound (exhaustive state-space traces of the translocation guard, the basename comparison, and the `osascript_spawn_failed` flag). | — | Verified, no action needed. |

### Minor / Not Blocking (accepted as-is, or deferred)
- `notify_cli_symlink_declined` uses `.spawn()` and drops the `Child` unreaped for the wrapper's lifetime (one benign zombie); its sibling `notify_stuck_wrapper` appears to wait. Cosmetic inconsistency, not a bug.
- `notify_cli_symlink_declined`'s own spawn failure is silently swallowed rather than logged, unlike `notify_stuck_wrapper`. Trivial.
- Non-UTF8 path degrades to an empty string in the elevated-script pair list rather than being explicitly filtered — practically unreachable on APFS, and fails safely (the `ln` command just fails).
- An orphaned admin-privileges dialog if the app quits while the password prompt is open (the spawned thread isn't joined, and `process::exit()` on Quit doesn't kill the child `osascript` process). Real but narrow — needs an exact-timing quit-during-prompt window. Left as a follow-up.
- A TOCTOU window during the (possibly long, human-timescale) password prompt if `/usr/local` itself gets redirected by a same-UID attacker between attempt 1's failure and the elevated attempt's execution. Narrow reachability, bounded impact (payload is always our own trusted binary path, not attacker-controlled — a stray root-owned symlink primitive, not code execution).
- No genuine E2E test exercises `macos_ensure_cli_symlinks` itself on a real macOS runner (only the pure decision core is unit-tested). This matches the codebase's existing precedent for macOS-only side-effecting code (`launch_at_login.rs`'s equivalent functions are also untested), but the repo already has a working template for exactly this class of gap (`macos-bundle-swap-e2e.yml`) that this PR doesn't extend. Recommended as a near-term follow-up, not a blocker for this PR.
- `package-macos.sh`'s new fdev-bundling branch has no CI exercise on this PR (`cross-compile.yml` only runs on push to main/tags; `macos-dmg-test.yml` is `workflow_dispatch`-only) and `package-macos.sh` isn't in `ci.yml`'s shellcheck list. A `test-install-sh.sh`-style harness would close this cheaply. Follow-up.
- No linked GitHub issue for this fix (author is the maintainer, so not a CONTRIBUTING.md violation in effect, but there's no paper trail an external contributor would otherwise be required to leave). Optional.
- The freenet.org quickstart page copy (separate `freenet/web` repo) isn't updated to reflect that PATH setup is now automatic. Acceptable gap; a follow-up issue there would be worthwhile.
- `.claude/rules/bug-prevention-patterns.md`'s Release Smoke-Test Checklist has no step for manually verifying `which freenet fdev` after a real DMG launch — worth a line given this PR ships a `#[cfg(target_os = "macos")]` path with no CI runtime coverage.

---

### Testing Assessment
**Coverage Level:** adequate for the pure decision logic; the macOS-only execution wrapper and the packaging script changes rely on manual trace + existing project precedent rather than CI-exercised tests (see "Not Blocking" above).

25 unit tests across `service/cli_symlinks.rs` (16, up from the original 11) and `service.rs` (1 new: marker distinctness), covering: missing/correct/stale/broken symlinks, moved and renamed app bundles, foreign occupants (regular file and unrelated symlink), App Translocation detection, the tightened basename check (both the fix and its accepted residual-risk boundary), the AppleScript command builder (multi-tool single-prompt, `quoted form of` usage, quote escaping, backslash escaping, and — after round 2 — combined-character escape-order regression detection). All were mutation-tested where it mattered (the escape-order test and the basename-tightening test were both confirmed to fail against the pre-fix code). 125 total `commands::service` tests pass.

### Documentation
- Code docs: complete — every new public item has a doc comment explaining the "why," including the accepted residual-risk tradeoffs.
- PR description: corrected to accurately describe test coverage and review findings.
- Follow-up docs (release smoke-test checklist, quickstart page copy): not updated in this PR; noted above as optional follow-ups.

---

### Verdict
**State:** Ready to Merge, pending final CI green on commit 99b6d4dce (the last push was still running checks at review time — confirm `Simulation`/`Unit & Integration`/`Clippy`/`Fmt` etc. are green on this exact SHA before merging, per this repo's no-merge-with-red-ci rule).

**HEAD SHA reviewed:** 99b6d4dce959dc706080e30c25af3369dc5d729e (branch `fix/macos-dmg-cli-path`)

This is a re-review: round 1 (5 lenses) on 46292c44e found one Must-Fix (App Translocation) and several Should-Fix issues; round 2 (3 lenses) on 14095ba5d confirmed all were correctly fixed with no new bugs, and surfaced two more cheap test gaps, closed in 99b6d4dce (a purely test-additive commit, self-verified via mutation testing rather than requiring a third full review round). No open Must-Fix or Should-Fix items remain.

[AI-assisted - Claude]

PR #5303, opened 2026-08-13
fix(canary): freeze the marker pair, drop the unsound version relation

sanity:
## Problem

Two bugs in the release auto-update canary, one of which **would have blocked the 0.2.126 release**.

`MARKER_LATEST_SEEN_SINCE` names the first release whose binary emits `Startup update check: GitHub reports latest release`. Gate B skips its positive equality check for any previous release below that value.

**Bug 1 — a self-detonating pin.** The self-test added in #5290 asserted the constant was *not below* the crate version, and stated its own premise in its comment: *"the marker is new in this tree, so no already-published release emits it."* That premise was true only while 0.2.125 was unpublished. It expired at 00:11Z on 2026-08-13 when 0.2.125 shipped. The constant correctly stays at `0.2.125` forever, so the pin goes red on the next bump and blocks a healthy release. Reproduced: set the crate version to 0.2.126 against `main` and the pin fails.

**Bug 2 — a silent one, found during review.** Reword the marker text in every place a developer must touch it (`auto_update.rs`, `freenet.rs`, `auto-update-canary.sh`, and the log fixtures in both test files) and leave the constant alone: **the entire suite passes.** Nothing prompts anyone. A stale constant is the *default* outcome of a reword, not an escape from a warning. The consequence lands post-publish, where Gate B arms, greps for text the previous release never emitted, and trips the release Matrix alarm with a message blaming the node.

## Solution

**Delete the version relation entirely, and freeze the pair.**

Three attempts at a relation between the constant and the crate version were considered, and the conclusion is that **no fixed relation between two quantities on different clocks survives both publication and a reword**:

| relation | correct when | wrong when |
|---|---|---|
| `constant >= crate` (shipped in #5290) | before publication | after publication — blocks 0.2.126 |
| `constant <= crate` (first attempt here) | after publication | during a reword — red for the *correct* value |
| `constant <= crate + 1` (proposed in review) | see correction below | — |

**Correction, found by a later reviewer:** an earlier revision of this table claimed `<= crate + 1` "also fires on a reword", and that is arithmetically false. Under a reword the correct constant *is* crate+1, and `crate+1 <= crate+1` passes. The claim entered this description because it was relayed from a reviewer who withdrew their own suggestion on that reasoning; nobody checked the arithmetic. A loose upper bound of that shape appears to be phase-independent — it is violated only when the constant is more than one release ahead, which is never legitimate — so it is being re-evaluated as plausibility bounding alongside the freeze rather than as a replacement for it.

The two *equality-style* relations are what cannot work: the constant tracks release history and the crate version tracks this tree, so no relation asserting they stay in step survives both publication and a reword. What remains is:

1. **The value frozen** against its historical literal `0.2.125`. This catches strictly more than the relation ever did, including the empty string and forward creep at release time.
2. **The marker text frozen alongside it**, so a reword forces the author to confront the version constant in the same edit. This is what closes Bug 2.
3. **A format guard on `version_at_least`**, which previously accepted garbage as a pass: `version_at_least "not-a-version" "0.2.125"` returned true.

### Why the marker freeze is base64-encoded

Written first as a plain literal, and it **did not work**. The mutation it exists to catch is a rename sweep — and the sweep rewrote the frozen expectation too, because the expectation was byte-identical to the value it guarded. The pin followed the rename and the suite stayed green.

Encoded, a text sweep cannot reach it, and a deliberate reword has to regenerate it, which is exactly the moment of attention the pin exists to create. The failure message prints the decoded old text, the new text, and the re-encode command; an empty decode fails loudly rather than making the comparison vacuous.

The version half stays plaintext deliberately, and the asymmetry is documented: a repo-wide `sed` on a version string is not how bumps happen here (they touch `Cargo.toml` and the lockfile), so the realistic wrong edit is a human raising it in one place, which a plain compare catches.

### Also

- `gate_b_arm_case "0.2.124" no` closes a verified vacuity hole: hardcoding the threshold inside the decision previously left the *whole suite green*, because the existing cases straddled the gap.
- The frozen pin's failure message no longer reads as instructions for the creep edit that disarms Gate B.
- The operator-facing comment regains its reword escape hatch, under a heading naming when the constant should and should not move.
- The retraction-valve reason is written into the constant's comment, so the next reader does not conclude the skip branch is dead code. `prev_version` is the newest *currently-published* release, so a retraction legitimately puts it back below the constant.

## Testing

Suite is 68 assertions, all green. Every pin was mutation-tested by applying the exact regression it names to a **committed** tree and confirming it goes red.

| Mutation | Result |
|---|---|
| crate → 0.2.126 (the landmine) | `main`'s pin FAILS, reproduced. This HEAD green. Defused. |
| marker reword sweep, constant left stale | **plaintext freeze: GREEN — the bug above.** base64 freeze: RED, naming both strings |
| constant → 0.2.126 | RED via the freeze (which catches strictly more than the deleted relation) |
| `version.workspace = true` in Cargo.toml | green — the second spurious fire of the deleted relation, now gone |
| format guard removed | 4 RED: `not-a-version >= 0.2.125`, `0.2.125 >= ''`, `'' >= ''`, `0.2.125-rc.1 >= 0.2.125` all wrongly reported yes |
| threshold hardcoded to 0.2.124 in the decision | RED (was green before the new case) |
| `prev_emits_latest_seen` made to skip on malformed input | 2 RED |

The refuse-vs-arm asymmetry is deliberate and both directions are pinned: `version_at_least` refuses malformed input because *"is a >= b"* has no answer for garbage, while `prev_emits_latest_seen` arms Gate B on it because *"should Gate B assert anything"* does, and skipping is the silent direction.

`shellcheck -x` clean; the exact CI shellcheck invocation rc=0; `cargo fmt --all --check` clean. No Rust files changed.

## The `ci(rule-lint)` commits

Scope beyond the fix, called out as a **deliberate exception** to one-logical-change: without it the Rule Lint check rejects this PR for having no regression test, and `test-exempt` would put a false statement in the record on a PR that demonstrably has tests.

`"ok` is tightened to `["']ok[[:space:]]+-` and `pass` is added. Measured on a fixture: `echo "okay, the suite is starting"` and bare `echo "ok"` are now **rejected**; single-quoted `echo 'ok - ...'` and `pass "..."` are now **accepted**.

An honest limit, now stated in the CI comment itself: this check has never recognised assertions. It recognises the *success-reporting line* by prefix. `check "x" "a" "a"` is a real call to a real helper that compares a value with itself and asserts nothing, and the old pattern already matched it. A shared assertion helper would make it tidier without making it non-decorative. Whether something is a real regression test belongs to review and mutation testing.

Two false negatives are left in deliberately and documented rather than chased with a fifth pattern: `printf`-style reporting, and an assertion whose only added line is its FAIL branch (counting FAIL would let a PR score by adding an error message alone).

## Corrections to earlier revisions of this description

Kept visible rather than silently edited:

- An earlier version said the suite was "32/32". It was 61 at that point and is 68 now.
- An earlier version, and the comment added to `ci.yml`, claimed `auto-update-canary_test.sh` "has no helper at all". False — `check()` is at line 55 with 12 call sites. It is specialised to log-fixture cases; the file's other assertions have nothing to fixture and are written as bare echoes.
- An earlier version proposed `constant <= crate version` as the fix. That is the approach this HEAD deletes, for the reason in the table above.

## Fixes

Refs #5290. Prevents a self-inflicted block on the 0.2.126 release, and closes the silent reword path.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/bug-prevention-patterns.md
+++ b/.claude/rules/bug-prevention-patterns.md
@@ -87,7 +87,7 @@ documentation. The source pin then tracks the comment, not the code.
 | `Startup update check complete` | Emitted at `debug!`, so absent from every release binary. The canary's "did the check finish?" assertion could never observe it. |
 | `failed to parse latest version` | Occurs twice in `auto_update.rs`: the production `tracing::warn!` (the format literal `Startup update check: failed to parse latest version '{}'`) and a prose comment inside that file's own `#[cfg(test)] mod tests` (`// WARN failed to parse latest version 'v0.2.121':`). Both are quoted rather than cited by line number on purpose — the first version of this row cited `:1546`/`:1757`, which this very commit's +23 lines had already shifted to `:1569`/`:1780`. A line number in a rule about stale pins rots faster than the thing it describes. Rewording the production line left all 22 assertions green — including `ok - source pin: parse-failure marker` — while a node carrying the #5221 bug then logged check-ran + reworded-warn + check-complete and the canary reported `OK: parsed GitHub's response`. An ordinary log reword deletes the gate, with CI green throughout. |
 | (no marker at all) | The gate had nothing to say WHICH release the node compared against, so its healthy verdict was byte-identical to a silently-wrong comparator's — see the positive-fact rule below. Closed by `MARKER_LATEST_SEEN`. |
-| `triggering auto-update` | A fixed string, so it never matched `freenet.rs:609`'s "triggering IMMEDIATE auto-update". A node that took the urgent path read as one that never decided to update, for as long as that site had existed. Fail-closed, hence unnoticed. Closed by `MARKER_TRIGGERED_RE` plus a count pin. |
+| `triggering auto-update` | A fixed string, so it never matched the urgent site's "triggering IMMEDIATE auto-update" (cited by phrase: the six line numbers this table and the canary once carried were all low by 12 within one release). A node that took the urgent path read as one that never decided to update, for as long as that site had existed. Fail-closed, hence unnoticed. Closed by `MARKER_TRIGGERED_RE` plus a count pin. |
 
 ### The rule
 
@@ -111,7 +111,7 @@ documentation. The source pin then tracks the comment, not the code.
   `api.github.com`): two sources that are allowed to disagree produce failures
   that are not bugs.
 - **Pin the COUNT when a marker is supposed to match a SET of call sites.** A
-  fixed-string `MARKER_TRIGGERED` missed `freenet.rs:609` ("triggering
+  fixed-string `MARKER_TRIGGERED` missed the urgent site ("triggering
   IMMEDIATE auto-update") for as long as that site existed, so a node taking
   the urgent path read as one that never decided to update. It failed CLOSED,
   which is exactly why nobody noticed — **fail-closed is not the same as
@@ -231,6 +231,49 @@ version constant untouched. Two adjacent assertions are not a pair.
 - **Mutation-test the REMEDIATION PATH, not just the regression.** Apply
   the mutation, then do exactly what your own failure message says, and
   check whether the suite goes green while the problem remains.
+
+### Load-bearing justifications rot fastest, and their staleness is self-concealing
+
+A systematic sweep of #5303 found eight stale comments. **Seven were not
+descriptions of code — they were justifications**: *"this is the only thing
+keeping X out"*, *"mutation-tested: deleting this left the suite green"*, *"the
+grep returns nothing, so that hazard is not real"*, *"the lifecycle test covers
+this against real boots"*. Every one was false at the head that carried it.
+
+Two properties make this class worse than an ordinary stale comment:
+
+- **They tell the next reader to stop checking.** That is their whole purpose.
+  So the comment most likely to be wrong is the one most likely to suppress the
+  verification that would catch it. One of them — *"the grep returns nothing"* —
+  was the sole stated reason for deleting a real entry from a
+  contributor-facing list. The grep returned two hits.
+- **They rot precisely when the thing they justify is STRENGTHENED.** "This is
+  the only guard" stops being true the moment you add a second guard, which is
+  the good outcome. Three of the seven rotted that way inside four commits.
+
+They also mislead the *careful* reviewer specifically: a comment that outlived
+its code is indistinguishable from the bug it describes, and the careless reader
+never gets that far. One such comment in #5303 ("recorded from the LAST attempt",
+left behind by the commit that replaced last-writer-wins with a latch) cost a
+full review round — the reviewer read the tree rather than the author's report,
+which was the correct instinct, and the tree lied.
+
+**The rule:**
+
+- **A count or a grep result does not belong in prose.** Compute it where it can
+  go red. #5303's `EXIT_UNVERIFIED_ENVIRONMENTAL` pin reads the constant out of
+  the script instead of restating `75`; its shell-assertion counter recomputes
+  "no `*_test.sh` is invisible to this grep" on every CI run instead of carrying
+  the numbers a reviewer had measured by hand. Both replaced sentences that had
+  already gone stale once.
+- **Prefer "the first of two guards" to "the only guard".** State what a check
+  does, not what nothing else does — the second claim is falsified by the next
+  improvement.
+- **When you change code, the nearby comment is part of the change.** Especially
+  when the change makes an old caveat unnecessary: that is when the sentence
+  survives, because nothing forces you to look at it.
+- **When a comment and the code disagree, verify against the code before
+  reporting** — then fix the comment as a defect in its own right.
 
 ### Relation pins between quantities on different clocks
 
--- a/.claude/rules/bug-prevention-patterns.md
+++ b/.claude/rules/bug-prevention-patterns.md
@@ -274,6 +274,25 @@ which was the correct instinct, and the tree lied.
   survives, because nothing forces you to look at it.
 - **When a comment and the code disagree, verify against the code before
   reporting** — then fix the comment as a defect in its own right.
+- **A defect found at one site is not fixed until you have ENUMERATED every
+  site of that shape, and "enumerate" means grep, not recall.** This appeared
+  four times in one night in #5303, and the fixes were the misses: the
+  comment-stripping filter was applied to one job-block extraction while three
+  siblings went without (twice, in successive commits); the SIGPIPE fix in
+  #5236 corrected two helpers and left four call sites inside the very function
+  those helpers serve; and a false "on all N attempts" claim was corrected in
+  one branch while its sibling twelve lines away kept it. Each time the author
+  had just understood the defect completely, which is exactly when the
+  remaining instances feel like they cannot be there.
+
+  Three of the four were caught only because a MUTATION FAILED TO APPLY —
+  `PATTERN NOT FOUND` with the suite green. **An unapplied mutation is not a
+  pass; it is an unexplained observation**, and here the explanation was always
+  that the string had moved or been half-fixed. Chase it.
+
+  Prefer removing the class over fixing the instances: three extractions became
+  one `yaml_job_block` helper, so a fourth caller cannot forget the filter
+  because there is nothing to forget.
 
 ### Relation pins between quantities on different clocks
 
--- a/.claude/rules/bug-prevention-patterns.md
+++ b/.claude/rules/bug-prevention-patterns.md
@@ -195,6 +195,73 @@ hand-rolling a `split_once`. It:
 scrapes `update.rs` cannot be satisfied by its own assertion literal at
 all — a structural guarantee rather than a check someone must remember.
 
+### The same class without `include_str!`: an expectation stored as a copy
+
+A pin does not need a sliding anchor to follow its subject. **An expected
+value stored as a plaintext copy of the value it guards is rewritten by
+the same edit that changes the value**, so the pin renames itself and
+stays green. Nothing self-references; the two literals are simply
+identical, and the realistic edit is a sweep over both.
+
+Seen in #5303. `MARKER_LATEST_SEEN_SINCE` names the first published
+release emitting a log marker; rewording the marker requires moving it,
+and nothing prompted anyone (the source pin interpolates the marker, so
+it follows a rename by construction). A freeze was added storing the
+expected marker text as a literal beside the constant. Its own mutation
+test killed it: a reword is performed as a `sed` sweep across the files
+mentioning the marker, that sweep rewrote the frozen expectation too, and
+the suite passed with the constant still stale.
+
+Then the fix's *fix* failed the same way one level up. With the text
+stored base64 the reword went red — but following the failure message's
+own instruction (regenerate the encoded blob) went green again with the
+version constant untouched. Two adjacent assertions are not a pair.
+
+**The rules:**
+
+- **Encode the expectation** (base64, hash, checksum) whenever the value
+  it guards is text a sweep could match. Leave it plaintext only when the
+  realistic wrong edit is a human typing one new value — a version
+  constant qualifies, since bumps touch `Cargo.toml` and the lockfile,
+  not test scripts. Document the asymmetry where you rely on it.
+- **A freeze forces a decision only if the remediation cannot be
+  performed without making that decision.** Freeze values that must move
+  together as ONE blob, so regenerating the expectation is impossible
+  without re-stating both.
+- **Mutation-test the REMEDIATION PATH, not just the regression.** Apply
+  the mutation, then do exactly what your own failure message says, and
+  check whether the suite goes green while the problem remains.
+
+### Relation pins between quantities on different clocks
+
+A pin asserting a fixed relation between two values that change on
+*different schedules* is correct in one phase and wrong in another. It
+looks anchored — the second quantity is one the author cannot influence,
+which is normally the point — but it encodes a phase assumption nobody
+writes down.
+
+#5290 pinned `MARKER_LATEST_SEEN_SINCE >= crate version`, on the premise
+"the marker is new in this tree, so no published release emits it". True
+until that release published; the pin then blocks the *next* release for
+holding the correct value. Reversing it to `<= crate version` is correct
+after publication and wrong during a marker reword, where the right
+constant is crate+1 and the pin goes red for it.
+
+The constant tracks **release history**; the crate version tracks **this
+tree**. No relation asserting they stay in step holds in both phases.
+
+- Ask **"what future state legitimately requires this to move?"** before
+  pinning a relation. If the answer is "a state where the relation is
+  violated", the relation is the wrong pin.
+- Prefer a **freeze on the value**, which is phase-independent and
+  catches strictly more (the empty string, and the most likely wrong
+  value, neither of which a one-sided relation sees).
+- A *loose plausibility bound* is not the same thing and is not refuted
+  by the above: `<= crate + 1` passes during a reword. Note that the
+  claim "it fires on a reword too" was repeated through two reviews
+  before anyone did the arithmetic — **check a refutation before
+  propagating it**, especially one that is a single comparison.
+
 ### Audit
 
 Every pin must be mutation-tested when written: apply the exact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -414,58 +414,37 @@ jobs:
 
           if [ "$IS_FIX" -gt 0 ] && [ "$HAS_EXEMPT_LABEL" -eq 0 ]; then
             NEW_TESTS=$(echo "$ALL_ADDED_STRIPPED" | grep -cE "$TEST_ATTR_RE" || true)
-            # Some fixes are to shell tooling (e.g. scripts/release-agent/*.sh),
-            # whose regression tests are shell *_test.sh self-tests run in CI,
-            # not Rust #[test]s. Count added assertion lines (`check ...`) in
-            # added *_test.sh files so such fixes aren't forced to claim
-            # test-exempt. Without this, a genuine shell regression test scores
-            # zero against TEST_ATTR_RE (Rust-only) and the fix is wrongly
-            # rejected — the gap that hid the announce-to-river fix's own test.
-            # `check` is one of several assertion helpers in this repo's shell
-            # self-tests; `auto-update-canary_lifecycle_test.sh` reports through
-            # `ok`/`bad` instead. Matching only `check` made this rule miss a
-            # genuine behavioural regression test (#5290 added a lifecycle case
-            # that reproduces the fault and goes red without the fix) and demand
-            # `test-exempt` for a PR that had exactly the test the rule wants —
-            # the same "gate cannot see its subject" shape the comment above
-            # describes. Match the helpers actually in use.
+            # Some fixes are to shell tooling, whose regression tests are shell
+            # *_test.sh self-tests, not Rust #[test]s. Without this they score
+            # zero against TEST_ATTR_RE and the fix is wrongly rejected — the gap
+            # that hid the announce-to-river fix's own test.
             #
-            # Third and fourth forms, same gap. `auto-update-canary_test.sh`
-            # DOES have a helper -- `check()` at the top of the file, 12 call
-            # sites -- but it takes a log fixture and drives
-            # `assert_detection_healthy`, so it can only express the log-fixture
-            # cases; that file's other ~49 assertions (source pins, the frozen
-            # constants, the trigger-site count) have nothing to fixture and are
-            # written as a bare `echo "ok   - ..."`. And four more self-tests
-            # report through `pass` (`release_canary_wiring_test.sh`,
-            # `release_state_restore_test.sh`, and two under `release-agent/`).
-            # A PR whose entire regression test is one of those scored ZERO and
-            # was told to add a test it had already added.
+            # WHAT THIS RECOGNISES, precisely: the SUCCESS-REPORTING LINE of an
+            # assertion, by prefix. Not assertions. A decorative
+            # `echo "ok - looks fine"` satisfies it, and so did `ok "..."` and a
+            # self-comparing `check "x" "a" "a"` before it was widened. What
+            # stops a decorative test is review and mutation-testing, not this
+            # grep; widening only adds spellings.
             #
-            # Be honest about what this can and cannot do. It recognises the
-            # SUCCESS-REPORTING LINE of an assertion, by prefix; it has never
-            # recognised assertions. A single `echo "ok - looks fine"` that
-            # tests nothing satisfies it -- but so does `ok "..."` and so does a
-            # self-comparing `check "x" "a" "a"`, both of which the pre-existing
-            # pattern already accepted. Widening adds spellings, not a weakness;
-            # what actually stops a decorative test is review and
-            # mutation-testing the pin, not this regex.
+            # The alternation was widened three times, each time stopping at the
+            # file in front of whoever was editing. It is now derived from an
+            # ENUMERATION of every `*_test.sh` under scripts/ rather than from
+            # the file at hand: `check` (4 files), `ok`/`bad` (lifecycle),
+            # `pass` (3 files), `echo "ok   - "` (34 sites, and all 34 use that
+            # exact separator, so anchoring on it costs nothing while rejecting
+            # `echo "okay, starting"`), and `echo "PASS [...]"`
+            # (release_state_restore_test.sh, the last file that still scored
+            # zero). Both quote styles. Re-run that enumeration before adding a
+            # sixth pattern; adding one because a file you happen to be editing
+            # scores zero is how the first three went in.
             #
-            # It still has FALSE NEGATIVES, left in deliberately rather than
-            # chased with a fifth pattern: `printf 'ok - ...'` scores zero, and
-            # so does an assertion whose only added line is its `echo "FAIL ..."`
-            # branch. Counting FAIL lines would also let a PR score by adding an
-            # error message alone. The durable fix is one shared assertion helper
-            # across these files, not more regex.
-            #
-            # `["']ok[[:space:]]+-` rather than a bare `"ok`: anchored on the
-            # `ok   - ` separator every one of the ~30 existing sites uses, so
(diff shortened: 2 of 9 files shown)


sanity (commented):
## Comprehensive PR Review: #5303

### Summary
- **PR Title:** fix(canary): pin MARKER_LATEST_SEEN_SINCE below the crate version, not above
- **Type:** fix
- **CI Status:** pending at time of review
- **Review tier:** **Full** (touches `.github/workflows/ci.yml` — deploy/release/CI config is a listed high-risk surface)
- **HEAD SHA reviewed:** `da1b3a44c851baaa118ac9ba78d2d5403dfda178`
- **Reviewers run:** code-first, testing, skeptical, big-picture, plus a fifth lens specific to this change's failure mode (pin correctness: can it fail, will it fire spuriously). External models not run — opt-in only per the current review rule, so a fifth Claude lens was substituted rather than waiting on quota.

**Verdict: Needs Changes — Re-review Required After Fix.**

The PR's central claim is **verified independently, by execution, by three separate reviewers**: with the crate at 0.2.126, `main`'s pin goes red (`MARKER_LATEST_SEEN_SINCE (0.2.125) is BELOW the crate`) and the PR's stays green. The landmine is real and the direction of the fix is right. The frozen fact was also checked against ground truth — the published v0.2.125 musl asset was downloaded and `strings` finds the marker.

But the review found that **the replacement pin reintroduces the same defect class with the sign flipped**, and that is blocking.

---

### The blocking finding: no static relation is correct in both phases

`main`'s pin (`constant >= crate`) was correct before publication and wrong after. This PR's pin (`constant <= crate`) is correct after publication and wrong during a **marker reword** — which the code explicitly anticipates (`auto_update.rs` and the source pin at `_test.sh:864` both say "do not reword without updating both").

On a normal dev tree the crate version equals the last published release; the bump happens inside the release commit. So a marker reworded on main today first ships in crate+1. Then:

- constant = crate+1 (**the correct value**) → relation pin RED, blocks a healthy release
- constant = crate (**green**) → Gate B demands the NEW text from the published previous binary, which emits the OLD text → Gate B goes spuriously red post-publish, on the Matrix channel `cross-compile.yml:764` alerts

Green test means broken canary; correct constant means red test. Verified by execution in an isolated worktree.

**The conclusion is not to tune the inequality** — it is that no static relation survives both phases. Two reviewers independently proposed relation tweaks (`<= crate+1`, and a conditional form); the lens that tested them **withdrew its own suggestion** after confirming it also fires on a reword.

**Fix: delete the relation pin, keep the frozen literal.** The frozen literal is phase-independent: it pins a value and forces a deliberate co-edit for any change, reword included. Deleting it also removes a second spurious fire found in testing: `version.workspace = true` in `crates/core/Cargo.toml` makes the relation pin fail with "could not read the crate version".

---

### Must Fix (Blocking)

1. **Delete the `constant <= crate version` relation pin** (`scripts/auto-update-canary_test.sh:775`), keep the frozen literal. Reason above.
2. **Add `gate_b_arm_case "0.2.124" no`** (`_test.sh:~479`). Verified vacuity hole: replacing the decision body with a hardcoded `version_at_least "$1" "0.2.124"` leaves the **entire suite green** — nothing pins that the decision reads the constant rather than a literal. The existing cases straddle the gap (0.2.123 no / 0.2.125 yes), so any threshold in {0.2.124, 0.2.125} satisfies all of them. Verified both ways: red against the hardcode, green against real code, and it survives a legitimate reword so it adds no spurious-fire surface. Also correct the annotation "the one release that must skip" — two releases skip now.
3. **Reword the frozen pin's failure message** (`_test.sh:817-819`). It currently instructs the reader to "change the constant and this expectation together, in one commit" — which is *exactly* the creep edit that disarms Gate B. Verified: constant + frozen literal + crate all moved to 0.2.126 leaves the suite fully green while Gate B skips its positive check (runtime-confirmed: `constant=0.2.126, prev=0.2.125 -> SKIP`). The failure message should not read as instructions for the failure mode.
4. **Restore the reword escape hatch** in the operator-facing comment (`scripts/auto-update-canary.sh:173-175`). This PR *narrowed* it, from "…or this constant is set one release too early (bump it)" to "…not a reason to touch this constant". Under a reword, touching the constant is the correct response, and an on-call reader debugging a red Gate B lands here, not in the test file.
5. **Fix a factual error in the CI comment** (`.github/workflows/ci.yml:436`): "`auto-update-canary_test.sh` has no helper at all" is false. `check()` is defined at `_test.sh:55` with 12 call sites. Accurate statement: `check()` is specialised to log-fixture cases and the file's other assertions are bare echoes. **Same error is in the PR body and will be corrected there too.**
6. **Update the user-facing error text** (`ci.yml:448`). It still says "an added `'check'` assertion", naming one of three accepted forms, so a developer who trips the gate is told to add the wrong thing. This is the same misleading message that sent this PR's author down the wrong path.
7. **Tighten the new regex** from `"ok` to `"ok[[:space:]]+-` (`ci.yml:443`). Unanchored, it counts `echo "okay, the suite is starting"` and bare `echo "ok"` as regression tests. Costs nothing: all 30 `echo "ok` sites in the repo use the `ok   - ` form.

### Should Fix

8. **Add `pass` to the regex alternation.** One word, and it closes four more shell self-tests that currently score zero (`release_state_restore_test.sh`, `release_canary_wiring_test.sh`, `release-agent/deploy-local-gateway_test.sh`, `release-agent/gateway-auto-update_test.sh`). The PR predicts a "fourth pattern" and defers it; this is the cheap part of that set.
9. **Reconcile contradictory adjacent sentences** in `auto-update-canary.sh:169-182`: "makes this self-retiring" now sits beside "FROZEN … release history does not change".
10. **PR body says "Full canary suite 32/32"**; the suite has 61 assertions. The body's weight rests on exact evidence, so the number should be right.

### Consider / Follow-ups

- **Reconciling two reviewers who disagreed.** Big-picture argues the version gate is now dead code (once 0.2.125 is published, `prev >= constant` always, so the skip branch is unreachable) and should be deleted, making Gate B's positive assertion unconditional. That is appealing but **not safe as stated**, and the blocking finding is why: the constant is precisely the mechanism that makes a marker reword survivable. Delete it and a reword release fails Gate B with no way to express "the previous binary legitimately emits different text". Recommend keeping the constant; if the gate is revisited, handle the reword case explicitly (e.g. the canary accepting the previous release's marker text).
- **One shared assertion helper or marker convention across all shell self-tests**, replacing the regex-widening cycle. This is the third widening; two further unmatched idioms already exist.
- **Add the self-detonating-pin class to `.claude/rules/bug-prevention-patterns.md`.** "A pin whose premise is *this release is not yet published* detonates the moment it publishes" generalises beyond this file, and right now the lesson lives only in a comment inside one test file. Also list the `MARKER_LATEST_SEEN_SINCE` pins in the enumeration at `:153-157`.
- **For whoever cuts 0.2.126:** Gate B's positive equality check has never armed in production (it was skipped for 0.2.125 because prev was 0.2.123). 0.2.126 is the first release where it runs for real, so a previously-unexercised path goes live at release time.

### What holds up

- The landmine is real, reproduced independently three times.
- The frozen literal does the load-bearing work: forward creep at release time is invisible to the relation pin and caught only by the freeze.
- The "two pins are complementary" claim survived a deliberate attempt to refute it (one reviewer reported the failed refutation, which is the right thing to do).
- Neighbouring untouched pins were audited and are live, each reddened by mutating its subject. Every whole-file-grep marker has its only hits in production code above the `#[cfg(test)]` boundary, so none is currently satisfiable by a comment or test block.
- `shellcheck -x` clean; suite 61/61 green at HEAD.

### Process note

The shared review worktree was briefly dirty with one lens's mutation edits while others were reading it — my error in co-locating them, corrected mid-review. Three reviewers noticed independently and re-ran against clean `git archive` exports, and both worktrees were verified clean afterwards. Findings above were taken from those clean runs. Flagging it because it could have produced phantom findings and did not only because the reviewers caught it.

[AI-assisted - Claude]

sanity:
## Review addendum: two corrections to my synthesis above, and one new finding

The fifth lens reported after I posted. It corrects the consolidated review in two places, both in the direction of "the problem is worse than stated", and adds a finding none of the other four had. Recording it here rather than editing the review above, so the correction is visible.

### Correction 1 — the reword failure is SILENT, not a "tempting escape"

My review described a developer facing a red pin and taking a bad escape. That framing is wrong, and the truth is worse.

Verified by rewording `MARKER_LATEST_SEEN` in the complete set of places a developer must touch (`auto_update.rs`, `freenet.rs`, `auto-update-canary.sh`, and the log fixtures in both test files) while leaving the constant alone: **the suite passes completely.** `All auto-update-canary assertions passed`, lifecycle green, zero red.

So nobody is tempted, because nobody is prompted. **A stale constant is the DEFAULT outcome of a reword**, not an escape from a warning. The source pin at `_test.sh:864` interpolates `$MARKER_LATEST_SEEN`, so it follows the rename by construction; the relation check and the freeze never look at the marker at all.

Severity also shifts. The old landmine fired in CI on the bump PR: cheap, pre-publish, with a message naming the constant. This one fires in the **post-publish** canary (`auto-update-selfupdate-canary` has `needs: attach-to-release`), trips the Matrix alarm at `cross-compile.yml:764`, and the alarm text blames the node — "a node on the previous release may not be able to auto-update to this one". Same defect class, strictly later and more expensive, and misdiagnosing.

**This changes the recommended fix.** In addition to deleting the relation pin, **freeze the PAIR**: add `MARKER_LATEST_SEEN_FROZEN='Startup update check: GitHub reports latest release'` alongside the version freeze at `_test.sh:809`. A reword then forces the author to confront the version constant in the same edit, which is the only thing that closes the silent path. This supersedes item 1 of my review as the highest-value change.

### Correction 2 — the "delete the version gate" argument is refuted, and my reason was the weaker one

I argued deletion was unsafe because the constant is what makes a reword survivable. That holds, but there is a stronger and independent reason I missed.

`prev_version` is not "the previous version". It is `.[0]` of `releases?per_page=30` filtered `draft==false and prerelease==false` (`cross-compile.yml:718-720`) — **the newest currently-published release**. Releases in this repo demonstrably do fail to publish: `v0.2.124` is `draft=true, published=null` on the live API right now, which is the entire reason this PR exists.

So if 0.2.125 were ever retracted or re-drafted, `prev_version` becomes `v0.2.123`, and the skip branch correctly fires because 0.2.123 genuinely does not emit the line. **The branch is a live safety valve for release retraction.** Deleting it converts a retraction into a false fleet alarm.

Big-picture is right that the branch is dormant and wrong that it is dead. Dormant-but-correct is what a valve looks like.

### New finding — `version_at_least` accepts garbage as a pass

`scripts/auto-update-canary.sh:780`, verified against GNU coreutils 9.4:

- `version_at_least "not-a-version" "0.2.125"` → **TRUE**. A non-empty malformed crate version silently passes; the `-z` guard at `:772` covers empty only.
- `version_at_least "0.2.125" ""` → **TRUE**. An emptied constant passes the relation check.
- `"0.2.125-rc.1" >= "0.2.125"` → TRUE, which contradicts the docstring's "semver" (pre-releases sort below in semver, above under `sort -V`). No practical impact; this repo cuts no rc tags.

A one-line format check on `crate_version` turns the first case from a silent pass into a loud failure. Worth doing regardless of the relation pin's fate, since `version_at_least` has other callers.

### Also worth adding to item 7 (the regex)

The false-**negative** half was not covered above. Verified: `echo 'ok - …'` (single quotes) → 0. `printf 'ok - …'` → 0. `echo "FAIL - …" >&2` → 0. So the matcher misses the branch that actually carries the assertion, and pins the spelling of one convention. That is the "pin the invariant, not the spelling of one violation" shape from `.claude/rules/browser-assets.md` rule 2 — in a gate whose entire history is that it cannot see its subject.

### Revised recommendation

Unchanged verdict (**Needs Changes — Re-review Required After Fix**), with the must-fix list amended:

- **1a (new, highest value).** Freeze the marker string alongside the version.
- **1b.** Delete the relation pin, keep the freeze. Reinforced: the relation catches only the strictly-above-crate class, while the freeze catches everything else including the empty string and the most-likely-wrong value. The comment block at `:750-771` presents this backwards.
- **New.** Add a format check on `crate_version`.
- Items 2-7 stand.
- The "delete the version gate" follow-up is **withdrawn**, on the retraction-valve ground.

[AI-assisted - Claude]

sanity (commented):
## Re-review: #5303 at `98196ec69`

Mandatory re-review after seven blocking findings were addressed. Four independent lenses, each in its own isolated worktree this time (last round's shared checkout was briefly dirty with one lens's mutations, which three reviewers noticed and worked around).

- **Review tier:** Full (CI config). External models opt-in only per the current rule; four Claude lenses with distinct adversarial angles ran instead.
- **HEAD reviewed:** `98196ec69`. `main` has not moved since the branch point, so this is the content that would merge.
- **Verdict: Needs Changes — Re-review Required After Fix.**

### What holds up, verified independently and repeatedly

Between them the lenses executed roughly forty mutations. The following were attacked and could not be broken:

- **The base64 marker freeze does what it claims.** A `sed` rename sweep across all four files holding the marker goes RED. A plaintext expectation would have been rewritten by the same sweep — it rewrote three other occurrences in that very file — so the encoding is doing real work.
- **Decode failure is loud in every mode:** empty, corrupt, deleted assignment, and a stubbed-out `base64` binary all produce a clean FAIL via the `-z` guard, never a vacuous comparison. No `set -e`, so no mid-suite abort.
- **Deleting the relation pin lost nothing.** The freeze reddens for constant `0.2.999`, `''`, `not-a-version`, `0.2.100`, `0.2.126` — every state the relation caught and several it did not. The self-detonation claim was ground-truthed: crate at 0.2.126 leaves this HEAD green and reddens `main`'s pin.
- **The frozen value is true.** The published v0.2.125 musl asset was downloaded and `strings`-grepped: the marker is present.
- **Every source-scrape pin passes the AGENTS.md bounded-region rule.** `preflight_body`/`selfupdate_body` are awk-bounded with an explicit empty-body FAIL; `pin_marker` binds to the emitting macro call, not a whole-file grep; every scrape is cross-file. **No bare unbounded-anchor pin exists in the file.** That is a clean bill on this repo's most common defect.
- All eight new assertions go red under the mutation each one names.

### Must Fix

**1. The "pair freeze" does not force the pair, and the escape is its own error message.** Found independently by three lenses. Two-step mutation: (a) reword sweep → RED, correctly; (b) then run only the copy-pasteable command the FAIL hands you, regenerating the encoded blob, and leave `MARKER_LATEST_SEEN_SINCE` alone → **suite fully green, constant stale.** That is exactly the state the pin exists to prevent. The two frozen values are independent assertions; the coupling is prose sitting directly above a one-liner that silences the alarm without honouring it.

The generalisation, which is the most useful thing this review produced: **a freeze forces a decision only if the remediation cannot be performed without making that decision.** Fix: encode both values as one blob, so regenerating the expectation is impossible without re-stating the version.

**2. The freeze protects one of six markers, and the unprotected set includes the silent direction.** Gate B's `assert_detection_healthy` runs over the *previous release's* binary, so the cross-version hazard belongs to Gate B, not to one constant. Verified: the identical sweep applied to `MARKER_PARSE_FAIL` or `MARKER_CHECK_RAN` leaves the suite green.

Directions differ and one is much worse. A `MARKER_CHECK_RAN` reword makes Gate B red against a healthy release — loud, same harm class as the bug being fixed. A **`MARKER_PARSE_FAIL` reword makes Gate B's negative check stop matching, so a genuinely broken previous release passes.** Silent false-pass, and that is the #5221 signature this canary exists to catch. Freeze that one here (it needs no version constant); scope the comment honestly for the rest; follow-up issue for the others.

**3. An unreadable constant now makes Gate B skip silently — a regression in the direction this PR is about.** `prev_emits_latest_seen` guards `$1` only, so an emptied or corrupt `MARKER_LATEST_SEEN_SINCE` falls through to `version_at_least`, which refuses, which skips. Verified: on `main` the same input *armed*. The comment's claim that an unusable version arms rather than skips is true only of `$1`. The resulting note reads "first emitted by v" with an empty version. Defence-in-depth only — CI's freeze catches a bad constant first — but it is the one input class the code claims to handle. One line: test both operands.

**4. The CI comment claims coverage that does not exist.** Measured: the widening fixes three of the four files it names. `release_state_restore_test.sh` reports `echo "PASS [$name]"` (uppercase) and scores **0 before and after**. A gate whose own documentation asserts coverage it lacks is the defect class this PR is about.

Worse, the *dominant* blind spot is undocumented while a hypothetical one is: `[[:space:]]` after the alternation excludes every underscore-suffixed helper (`gate_b_arm_case`, `version_ge_case`, `trigger_case`, `pin_marker`, …), 29 of 63 sites in the canary test alone. **4 of this PR's own 8 new assertions score zero**; its count of 2 comes entirely from the freeze block's echo lines. Meanwhile the documented `printf` false negative does not occur anywhere in the repo. Swap them, and add the helper-call-site pattern.

### Should Fix

**5. A load-bearing rationale claim is arithmetically false — and it was mine.** The description claimed `<= crate + 1` "also fires on a reword". It does not: under a reword the correct constant *is* crate+1, and `crate+1 <= crate+1` passes. The claim entered this PR because I relayed a reviewer's withdrawal of their own suggestion without checking the arithmetic behind it. Corrected in the description.

This matters because that claim is the stated reason no bound of any kind survives, which leaves the "wrong-but-consistent value" case unguarded — both frozen and live values are edited by the same hand in the same edit. A pure *upper* bound appears phase-independent (violated only when the constant runs more than one release ahead, which is never legitimate). Given that three relations in this file have now been wrong, it is to be **mutation-tested across four phases before being trusted**, not asserted on reasoning.

**6. `gate_b_arm_case "0.2.124"` does not pin what its comment says.** Verified: hardcoding `0.2.124` → red, `0.2.126` → red, **the correct literal `0.2.125` → green.** It pins "the threshold is exactly 0.2.125", not "the decision reads the constant". Two subshell cases fix it, or downgrade the claim.

**7. Three of five `is_dotted_version` refusal clauses are unpinned** — deleting `.*`, `*.`, `*..*` leaves the suite green.

### Follow-ups

- **The structural closure for finding 1:** a networked check that downloads the published `v$MARKER_LATEST_SEEN_SINCE` musl asset and greps it for `$MARKER_LATEST_SEEN`. This is the only assertion that **cannot be satisfied by editing the test file** — reword-and-regenerate leaves the published binary still emitting the old text. It was already performed by hand during this review; making it a gate is the structural version. Belongs beside the canary jobs in `cross-compile.yml`, since it needs network.
- Freeze the remaining Gate B markers.
- Add both lessons this PR earned to `.claude/rules/bug-prevention-patterns.md`: a pin on a relation between two quantities on different clocks is right in one phase and wrong in another; and an expectation stored as a plaintext copy of the value it guards is rewritten by the same edit that changes the value.
- `docs/RELEASING.md` "If Gate B fails" goes straight to "ship a fix release, roll nodes by hand". This PR's subject is a Gate B failure that is *not* a fleet problem, and that is the page an operator reads at 1am.

### Note on the process

This PR has now been wrong four times in the same way — the original pin, its first replacement, the plaintext freeze, and the pair coupling. Each was found by applying the mutation the previous round named, to the fix the previous round produced. That is the practice working rather than failing: two of the four were caught by the *author* mutation-testing their own new pin and reporting the failure unprompted. It is worth stating plainly that the code is a real improvement over `main` on every axis it targets at every one of those iterations.

[AI-assisted - Claude]

sanity (commented):
## Review round 3: #5303 at `f360228d8`

Three independent lenses, isolated worktrees. **Verdict: Needs Changes.** The round-2 defect is confirmed fixed; three new ones were found, and one behavioural harness closes most of them.

### Confirmed fixed, by a reviewer other than the author

The round-5 defect — a single-value freeze that could be silenced by following its own recipe — **holds now**. A reviewer swept `failed to parse` → `could not parse` across all four files, watched it go red, then did exactly what the failure message says and nothing else. **Still red**, on the behavioural backstop. That is the property the previous four rounds all failed. The backstop was also checked for faithfulness against `v0.2.121:auto_update.rs:1528` rather than assumed.

### Must fix

**1. The retry loop's `environmental` flag is last-writer-wins.** A hung updater on attempt 1 (a real finding, as the code's own comment says) followed by a port collision on attempt 2 yields exit 75 and the quiet "This is NOT a stranded fleet and needs no fleet action." A false-quiet on the exact class the gate exists to raise. Latch the flag: once any attempt is a non-environmental indeterminate, the verdict is `fault`.

**2. A real #5221 can reach the quiet message via a reorder nothing pins.** As shipped it cannot — a log holding both markers returns rc=1 because the parse-fail branch precedes the fetch-fail branch. That ordering is the only thing making the split safe, and no fixture in the suite holds both markers. Swap the branches (an innocent "infra check before product check" tidy) and the same log classifies environmental. One `check` case closes it.

**3. "Retry only the environmental case" is pinned by nothing.** Mutating `[ "$rc" -eq 2 ] || break` to also retry rc=1 leaves all three suites green. Deterministic faults still fail, but an *intermittent* real fault becomes a pass — a regression arriving as a flaky pass through the canary's own retry logic.

**4. The classifier is pinned by substring presence only.** Flipping `&&` to `||`, or keeping the call and ignoring its result, leaves everything green while every indeterminate becomes environmental. **This is the exact shape the round-2 `prev_emits_latest_seen` fix closed, reintroduced one function over** — and the pin's own failure message names this hazard and cannot see it.

**1–4 are all closed by one thing**, which both lenses proposed independently: a behavioural driver for `cmd_selfupdate` (stub `curl`/`tar`/`run_node_until_check`, plant fixture logs, assert attempt count *and* returned code). `release_wait_for_binaries_test.sh:211` is the in-repo precedent.

**5. A port collision is reported as "could not reach GitHub"** in both the `::error::` and the Matrix message. The node never started. Gate A kept a dedicated port-collision message; Gate B lost it this round. Carry the cause, not just the flag.

**6. The new self-test is invisible to the counter it guards.** 26 bare `expect` sites; the pattern requires `expect_`. So `ci.yml`'s "329 of 329, zero false positives" is false at this HEAD, and the discipline it prescribes ("re-run the enumeration before adding a pattern") was not followed by the commit that prescribes it.

**7. The Gate B wiring pin has a hole.** `exit "$rc"` is pinned; `|| rc=$?` is not. Changing the latter to `|| true` — the reflex fix — leaves the suite green while Gate B reports success on a broken updater.

### Should fix

- `CANARY_ATTEMPTS=00` passes the zero guard, the loop never runs, and `rc` is unbound → exit 1 → the loud alarm.
- The `75` constant and the workflow's literal `if [ "$rc" -eq 75 ]` are coupled and unpinned.
- The backstop's fixture yields rc=2, not the rc=0 its message claims, because it omits the completion line that a real #5221 log carries.
- `MARKER_FETCH_FAIL` now feeds a *classification decision*, not just a diagnosis, so a reword silently disables the environmental path and sends every blip back to the loud alarm. Comment corrected here; the freeze goes to #5309 as its top item. (Deliberately not frozen in this PR: its reword direction is a loud false alarm, not a silent pass. `MARKER_PARSE_FAIL` was frozen here precisely because its direction *is* silent.)
- The body describes roughly 40% less than the diff and is being rewritten.

### Deliberately NOT doing

No behavioural backstop for `MARKER_LATEST_SEEN`. After a legitimate reword the canary is *meant* to stop reading old binaries and skip, so such a pin would fire on the correct remediation. Raised unprompted by a reviewer; agreed. Worth recording so the lesson from `MARKER_PARSE_FAIL` is not over-applied.

### Verified sound — please do not re-litigate

The vacuity guard is layered and the layer that matters works (forcing the exact first-draft `FRAG=''` bug is caught by 11 negative cases). Zero false positives across all 54 new pattern hits repo-wide. The notify split survives step reorder, reformatting, condition widening, and output deletion; cancelled/empty/never-set all land loud. Exit 75 propagates end to end. Every version-gate pin reddens under its named mutation, and the 0.2.124 straddle is confirmed closed. A real #5221 cannot currently be misclassified (marker ordering verified with a log containing both).

### One thing nobody can test locally

Whether GitHub populates `needs.<job>.outputs` for a **failed** job. It is load-bearing for the whole notify split. If it comes back empty, everything takes the loud branch — so it degrades safely — but it needs watching on the first real tag run, and that belongs in the body.

---

Five of this PR's defects have now been found *inside the fixes for earlier ones*. That is worth stating plainly rather than hiding: each was caught by applying the mutation the previous round named, and two were caught by the author against their own work before anyone asked. The code is a real improvement over `main` at every iteration.

[AI-assisted - Claude]

sanity:
## On the scope-creep warnings

The rule-review check flags four of these, and a human-equivalent big-picture reviewer raised the same objection earlier in this PR's review. It deserves an answer on the record rather than a silent override label, because the objection is basically correct on its face: this PR contains four logical changes and its title names one.

**The counter-argument, and it is a causal one rather than a convenience one.**

The Gate B environmental classification (exit 75, the corroboration probe, the retry loop, the notify split) is not merely *bundled* with the marker freeze. It is **caused** by it.

Gate B's positive equality check has never executed in production — it was skipped for 0.2.125 because the previous release predated the marker. The constant frozen in this PR is precisely what makes it arm for the first time on 0.2.126. And the marker it then requires is emitted **only when the node's own GitHub fetch succeeds** — a fetch with no retry, which was observed failing twice in one day on a single real machine during this PR's review.

So shipping the marker half alone would ship the change that arms the check *without* the change that stops it false-alarming, and the first symptom would be a post-publish 🚨 telling the team the fleet may be stranded, on a healthy release. Splitting here does not reduce risk; it sequences the two halves in the order that produces the incident.

**The rule-lint half is load-bearing in a narrower and more boring way**, verified rather than asserted: on this branch's diff the pre-existing regex scores **0** added shell assertions, so without the widening, Rule Lint rejects this PR for having no regression test. `test-exempt` was rejected as an alternative because it would put a false statement in the record on a PR that demonstrably has tests.

**What is genuinely separable and is being kept separate:** the Gate A reordering — moving `cargo publish` downstream of the pre-publish canary so a blocked gate burns a deletable tag instead of a permanent crates.io version — is a distinct branch, already written, deliberately *not* in this PR.

**What this PR should have done differently:** the title and body named one change for most of the PR's life. Both now describe all four. That was the reviewer's actual condition for shipping it whole, and it was the fair criticism.

For anyone auditing later: this went through four review rounds (three full, one focused), which found **seven** instances of the same defect class — a gate that cannot see its subject — every one of them inside the fix for the previous one. The last of those was found because a mutation returned `PATTERN NOT FOUND` alongside a green suite and someone investigated instead of moving on.

[AI-assisted - Claude]

